Issue #2, opened 2015-11-14
need a C++ API


dneto0:
The SPIR-V Tools API is a C API to provide ABI stability and a lowest-common-denominator target for multiple languages.  However, it's begging for a more convenient C++ API.

antiagainst:
+1

antiagainst:
This is done via #410 & #411.

PR #28, opened 2015-11-24
Several unit tests for spvBinaryParse

dneto0:
Add a non-zero spv_result_t value SPV_REQUESTED_TERMINATION
which should be used to signal an ok result, but signals
early termination for a process, such as binary parsing.

Tests include:
 - correct contents sent to header and instruction callbacks
 - non-zero status from a callback should terminate parsing,
   but the parser should not generate its own diagnostic.

TODO: Check diagnostics generated by the parser itself.

### DIFF
--- a/test/BinaryParse.cpp
+++ b/test/BinaryParse.cpp
@@ -24,12 +24,429 @@
 // TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 // MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
+#include "UnitSPIRV.h"
+
+#include "TestFixture.h"
+#include "gmock/gmock.h"
+
+// Returns true if two spv_parsed_operand_t values are equal.
+// To use this operator, this definition must appear in the same namespace
+// as spv_parsed_operand_t.
+static bool operator==(const spv_parsed_operand_t& a,
+                       const spv_parsed_operand_t& b) {
+  return a.offset == b.offset && a.num_words == b.num_words &&
+         a.type == b.type && a.number_kind == b.number_kind &&
+         a.number_bit_width == b.number_bit_width;
+}
+
 namespace {
 
+using ::spvtest::Concatenate;
+using ::spvtest::MakeInstruction;
+using ::spvtest::MakeVector;
+using ::testing::_;
+using ::testing::AnyOf;
+using ::testing::ElementsAre;
+using ::testing::Eq;
+
+// An easily-constructible and comparable object for the contents of an
+// spv_parsed_instruction_t.
+struct ParsedInstruction {
+  ParsedInstruction(const spv_parsed_instruction_t& inst)
+      : words(inst.words, inst.words + inst.num_words),
+        opcode(inst.opcode),
+        ext_inst_type(inst.ext_inst_type),
+        type_id(inst.type_id),
+        result_id(inst.result_id),
+        operands(inst.operands, inst.operands + inst.num_operands) {}
+
+  std::vector<uint32_t> words;
+  SpvOp opcode;
+  spv_ext_inst_type_t ext_inst_type;
+  uint32_t type_id;
+  uint32_t result_id;
+  std::vector<spv_parsed_operand_t> operands;
+
+  bool operator==(const ParsedInstruction& b) const {
+    return words == b.words && opcode == b.opcode &&
+           ext_inst_type == b.ext_inst_type && type_id == b.type_id &&
+           result_id == b.result_id && operands == b.operands;
+  }
+};
+
+// Prints a ParsedInstruction object to the given output stream, and returns
+// the stream.
+std::ostream& operator<<(std::ostream& os, const ParsedInstruction& inst) {
+  os << "\nParsedInstruction( {";
+  spvtest::PrintTo(spvtest::WordVector(inst.words), &os);
+  os << "}, opcode: " << int(inst.opcode)
+     << " ext_inst_type: " << int(inst.ext_inst_type)
+     << " type_id: " << inst.type_id << " result_id: " << inst.result_id;
+  for (const auto& operand : inst.operands) {
+    os << " { offset: " << operand.offset << " num_words: " << operand.num_words
+       << " type: " << int(operand.type)
+       << " number_kind: " << int(operand.number_kind)
+       << " number_bit_width: " << int(operand.number_bit_width) << "}";
+  }
+  return os;
+}
+
+// Sanity check for the equality operator on ParsedInstruction.
+TEST(ParsedInstruction, ZeroInitializedAreEqual) {
+  spv_parsed_instruction_t pi = {};
+  ParsedInstruction a(pi);
+  ParsedInstruction b(pi);
+  EXPECT_THAT(a, ::testing::TypedEq<ParsedInstruction>(b));
+}
+
+// A template class with static member functions that forward to non-static
+// member functions corresponding to the header and instruction callbacks
+// for spvBinaryParse.
+template <typename Client, typename T>
+class BinaryParseClient : public T {
+ public:
+  static spv_result_t Header(void* user_data, spv_endianness_t endian,
+                             uint32_t magic, uint32_t version,
+                             uint32_t generator, uint32_t id_bound,
+                             uint32_t reserved) {
+    return static_cast<Client*>(user_data)
+        ->HandleHeader(endian, magic, version, generator, id_bound, reserved);
+  }
+
+  static spv_result_t Instruction(
+      void* user_data, const spv_parsed_instruction_t* parsed_instruction) {
+    return static_cast<Client*>(user_data)
+        ->HandleInstruction(parsed_instruction);
+  }
+};
+
+using Words = std::vector<uint32_t>;
+using Endians = std::vector<spv_endianness_t>;
+using Sentences = std::vector<Words>;  // Maybe this is too cute?
+using Instructions = std::vector<ParsedInstruction>;
+
+// A binary parse client that captures the results of parsing a binary,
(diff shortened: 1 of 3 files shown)


test/BinaryParse.cpp
```diff
@@ -24,12 +24,421 @@
...
 
+#include "UnitSPIRV.h"
+
+#include "TestFixture.h"
+#include "gmock/gmock.h"
+
+// Returns true if two spv_parsed_operand_t values are equal.
+// To use this operator, this definition must appear in the same namespace
+// as spv_parsed_operand_t.
+static bool operator==(const spv_parsed_operand_t& a,
```
dmircevski: Didn't we say we'll use anonymous namespace instead of `static`?

test/BinaryParse.cpp
```diff
@@ -24,12 +24,421 @@
...
+
+  bool operator==(const ParsedInstruction& b) const {
+    return words == b.words && opcode == b.opcode &&
+           ext_inst_type == b.ext_inst_type && type_id == b.type_id &&
+           result_id == b.result_id && operands == b.operands;
+  }
+};
+
+// Prints a ParsedInstruction object to the given output stream.
+void PrintTo(const ParsedInstruction& inst, std::ostream* os) {
```
dmircevski: Why not an overloaded `operator<<`?  That way EXPECT* macros can also stream it.

test/BinaryParse.cpp
```diff
@@ -24,12 +24,421 @@
...
+  static spv_result_t Instruction(
+      void* user_data, const spv_parsed_instruction_t* parsed_instruction) {
+    return static_cast<Client*>(user_data)
+        ->HandleInstruction(parsed_instruction);
+  }
+};
+
+using Words = std::vector<uint32_t>;
+using Endians = std::vector<spv_endianness_t>;
+using Sentences = std::vector<Words>;  // Maybe this is too cute?
```
dmircevski: No way!  It's just the right amount of cute. ;)

test/BinaryParse.cpp
```diff
@@ -24,12 +24,421 @@
...
+
+// A binary parse client that captures the results of parsing a binary,
+// and whose callbacks can be made to succeed for a specified number of
+// times, and then always fail with a given failure code.
+template <typename T>
+class CaptureParseResults
+    : public BinaryParseClient<CaptureParseResults<T>, T> {
+ public:
+  // Capture the header from the parser callback.
+  // If no more successes ar e
```
dmircevski: Que?

test/BinaryParse.cpp
AWoloszyn: IIRC in order to get EXPECT* to print using the << operator you have to define
::std::ostream& operator<<(::std::ostream&, const ParsedInstruction&) rather than
ParsedInstruction::operator<<(::std::ostream&)
which is not allowed.

test/BinaryParse.cpp
antiagainst: lol

test/BinaryParse.cpp
dneto0: gtest supports PrintTo directly.
https://code.google.com/p/googletest/wiki/AdvancedGuide#Teaching_Google_Test_How_to_Print_Your_Values

However the Google C++ style guide recommends using the stream shift operator when you can.  Will change to it.

test/BinaryParse.cpp
dneto0: Yes, but that doesn't compile, as alluded to in the blurb comment.

This is the error you get if you put the operator in the anonymous namespace instead:

stl_algobase.h: In instantiation of ‘static bool std::__equal<_BoolType>::equal(_II1, _II1, _II2) [with _II1 = const spv_parsed_operand_t*; _II2 = const spv_parsed_operand_t*; bool _BoolType = false]’:
...
/sr/include/c++/4.8/bits/stl_algobase.h:797:22: error: no match for ‘operator==’ (operand types are ‘const spv_parsed_operand_t’ and ‘const spv_parsed_operand_t’)

and about 630 more lines of errors showing you various things it tried.

The problem is that spv_parse_operand_t is declared outside of any namespace.  So an equality operator on it must be in the same namespace.

test/BinaryParse.cpp
dneto0: Done.  Now print via the stream operator.

test/BinaryParse.cpp
dneto0: Whoops. Fixed.

test/BinaryParse.cpp
dmircevski: Ah, right, that makes sense.  Hopefully the C++ interface from #2 will obviate the need for local operator== everywhere.

test/BinaryParse.cpp
```diff
@@ -24,12 +24,428 @@
...
+  // Otherwise returns the stored failure code.
+  virtual spv_result_t HandleHeader(spv_endianness_t endian, uint32_t magic,
+                                    uint32_t version, uint32_t generator,
+                                    uint32_t id_bound, uint32_t reserved) {
+    endians_.push_back(endian);
+    headers_.push_back({magic, version, generator, id_bound, reserved});
+    return ComputeResultCode();
+  }
+
+  // Capture the parsed instruction data from the parser callback.
```
dmircevski: Optional: consider using googlemock for this.  Simply declare `HandleInstruction` a `MOCK_METHOD1` and use `EXPECT_CALL` in the tests below.  The framework is powerful enough that you shouldn't need any of the rest of the class.

test/BinaryParse.cpp
```diff
@@ -24,12 +24,428 @@
...
+                                              SPV_EXT_INST_TYPE_NONE,
+                                              0,  // type id
+                                              result_id,
+                                              i32_operands.data(),
+                                              uint16_t(i32_operands.size())};
+  return ParsedInstruction(parsed_i32_inst);
+}
+
+using BinaryParseTest =
+    spvtest::TextToBinaryTestBase<CaptureParseResults<::testing::Test>>;
```
dmircevski: Commingling the capturer with the test class is gratuitous.  Perhaps the capturer would be more readable as its own object?

test/BinaryParse.cpp
```diff
@@ -24,12 +24,428 @@
...
+      SpvOpExtInst, {2, 3, 1, uint32_t(OpenCLLIB::Entrypoints::Sqrt), 4});
+  const spv_parsed_instruction_t parsed_inst = {instruction.data(),
+                                                uint16_t(instruction.size()),
+                                                SpvOpExtInst,
+                                                SPV_EXT_INST_TYPE_OPENCL_STD,
+                                                2,  // type id
+                                                3,  // result id
+                                                operands.data(),
+                                                uint16_t(operands.size())};
+  EXPECT_THAT(instructions().size(), Eq(2u));
```
dmircevski: Consider using `testing::ElementsAre` instead of this.

test/BinaryParse.cpp
dneto0: Yes, I expect that.

test/BinaryParse.cpp
dneto0: I believe the current way leads to less code and more readable tests, once you've digested this using definition.
Otherwise you have to carry around the capturer object and give it a name.

(And before it's suggested, every test explicitly writes out the spvBinaryParse call, as per advice I absorbed from
https://cppcon2015.sched.org/event/3vmd/your-tests-are-terrible-tales-from-the-trenches)

dneto0: If you think different, I look forward to an alternative suggestion via pull request.

test/BinaryParse.cpp
dneto0: But this test is focused only on the contents of the second element.  I don't want to retest what was already tested in the earlier tests.  That would be confusingly redundant.

test/BinaryParse.cpp
dneto0: I don't think so, since this class has special behaviours to check the parser's response to request for early termination. See ComputeResultCode().

test/BinaryParse.cpp
dmircevski: Then use `testing::_` for the first element in `ElementsAre`.

test/BinaryParse.cpp
dmircevski: Googlemock is very powerful; I've yet to see special behaviours that `EXPECT_CALL` can't mimic.  But this is non-essential; the current code works.

test/BinaryParse.cpp
dmircevski: Fair enough.

test/BinaryParse.cpp
```diff
@@ -24,12 +24,428 @@
...
+  EXPECT_THAT(instructions().at(1), ParsedInstruction(parsed_inst));
```
AWoloszyn: I found it a little jarring to see EXPECT_THAT without an operator would 

    EXPECT_THAT(instructions().at(1), Eq(ParsedInstruction(parsed_inst)));

work?

dneto0: This works:
 EXPECT_THAT(instructions(), ElementsAre(_,ParsedInstruction(parsed_inst)));

and I can switch to that. Coming right up.

test/BinaryParse.cpp
dneto0: The other way to frame this choice:  The TextToBinaryParseBase test fixture itself is a commingling of test running mechanics with shortcuts for driving the API under test.  My approach is just another way of extending the test fixture.
Both approaches: commingling vs. utility objects are legitimate.

dmircevski:
+2

dneto0:
Squashed and pushed into master.

test/BinaryParse.cpp
dmircevski: See #32 :)

Issue #29, opened 2015-11-25
buffer overflow when there's a string literal at end of input, but without terminating null char


dneto0:
Also, there's an assert failure if the last instruction has a word count that extends beyond the input.

I have fixes for both queued up, buliding on https://github.com/KhronosGroup/SPIRV-Tools/pull/28

PR #30, opened 2015-11-27
spvBinaryParse: fix overruns, handle image format

dneto0:
Add unit tests for all diagnostics issued by spvBinaryParse.
    
    Handle image format operands in the binary parser and the
    disassembler.
    
    Document that the callback function pointers can be null,
    in which case they are ignored.
    
    Detect exhaustion of input when parsing an operand,
    to avoid buffer overruns on some invalid input cases.
    
    Fix the description strings for some operand types.
    Make the diagnostic messages for those operand types
    consistent between the assembler and binary parser.
    
    Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/29


source/binary.cpp
```diff
@@ -414,7 +431,7 @@ spv_result_t Parser::parseInstruction() {
   return SPV_SUCCESS;
 }
 
-spv_result_t Parser::parseOperand(size_t inst_offset,
+spv_result_t Parser::parseOperand(const size_t inst_offset,
```
dmircevski: We don't usually do this for parameters passed by value.  Why is it necessary?

source/binary.cpp
```diff
@@ -571,11 +593,23 @@ spv_result_t Parser::parseOperand(size_t inst_offset,
       convert_operand_endianness = false;
       const char* string =
           reinterpret_cast<const char*>(_.words + _.word_index);
-      size_t string_num_words = (strlen(string) / 4) + 1;  // Account for null.
+      // Compute the length of the string, but make sure we don't run off the
+      // end of the input.
+      const size_t remaining_input_bytes =
+          sizeof(uint32_t) * (_.num_words - _.word_index);
+      const size_t string_num_bytes = strnlen(string, remaining_input_bytes);
```
dmircevski: The name is misleading; this will normally be one less than the string's number of bytes.

source/binary.cpp
dneto0: Not necessary.  I thought it was useful documentation for the implmemetation that the variable would not change.
Will remove, as it appears jarring to you.

source/binary.cpp
dneto0: Renamed to string_num_content_bytes

test/BinaryParse.cpp
```diff
@@ -446,7 +490,269 @@ TEST_F(BinaryParseTest, ExtendedInstruction) {
...
+
+        // (End of input exhaustion test cases.)
+
+        // In this case the instruction word count is too small, where
+        // it would truncate a multi-word operand to OpConstant.
+        {Concatenate({ExpectedHeaderForBound(3),
+                      {MakeInstruction(SpvOpTypeInt, {1, 64, 0})},
+                      {spvOpcodeMake(4, SpvOpConstant), 1, 2, 44, 44}}),
+         "Invalid word count: OpConstant starting at word 9 says it has 4"
+         " words, but found 5 words instead."},
```
dmircevski: Is this because the last `44` isn't recognized as an OpCode?

dmircevski:
Looks good.  I wonder if we should try to contrive some test case where the extra words might be valid OpCodes.  Also, did I miss a test for string being too short?

test/BinaryParse.cpp
dneto0: No. The error is issued before it hits that last 44.  The parser is working from the assumption that the word count (4) is valid.  But the type information says the constant is 64 bits wide.
The 1 is the result type, the 2 is the result id, and there should be two words in the literal. But the word count of 4 says we shouldn't go that far.
I will add a better comment, and add the case where the second 44 is absent.

dneto0: Actually, the case where the second literal word is absent is already covered at line 600.
"End of input reached .... truncated possibly multi-word literal number operand at word offset 3."

dmircevski:
+2

The string-too-short case will be handled by the operand-type pattern.

dneto0:
@dmircevski if the extra words are valid opcodes, then the parse continues.  The parser grammar sets the expectation about what words should come next.  There's a final check that the sum of the operand word widths plus the opcode/count word match the advertised word count for the instruction.  That's in fact the "invalid word count" diagnostic.

PR #32, opened 2015-11-28
Simplify binary parser tests using GoogleMock.

dmircevski:


test/BinaryParse.cpp
```diff
@@ -435,15 +341,14 @@ TEST_F(BinaryParseTest, ExtendedInstruction) {
...
-                                                uint16_t(instruction.size()),
-                                                SpvOpExtInst,
-                                                SPV_EXT_INST_TYPE_OPENCL_STD,
-                                                2,  // type id
-                                                3,  // result id
-                                                operands.data(),
-                                                uint16_t(operands.size())};
-  EXPECT_THAT(instructions(), ElementsAre(_, ParsedInstruction(parsed_inst)));
+  EXPECT_CALL(client_, Instruction(ParsedInstruction(spv_parsed_instruction_t{
+                           instruction.data(), uint16_t(instruction.size()),
```
antiagainst: C-style cast: `uint16_t(...)`.

test/BinaryParse.cpp
```diff
@@ -99,95 +103,37 @@ TEST(ParsedInstruction, ZeroInitializedAreEqual) {
...
-  }
-};
-
-using Words = std::vector<uint32_t>;
-using Endians = std::vector<spv_endianness_t>;
-using Sentences = std::vector<Words>;  // Maybe this is too cute?
-using Instructions = std::vector<ParsedInstruction>;
-
 // A binary parse client that captures the results of parsing a binary,
 // and whose callbacks can be made to succeed for a specified number of
```
antiagainst: Is this comment still useful?

dmircevski:
Addressed feedback so far.

test/BinaryParse.cpp
```diff
@@ -249,20 +195,32 @@ ParsedInstruction MakeParsedInt32TypeInstruction(uint32_t result_id) {
...
-            spvBinaryParse(context, this, binary.data(), binary.size(), Header,
-                           Instruction, &diagnostic));
-  EXPECT_EQ(nullptr, diagnostic);
-  EXPECT_THAT(endians(), AnyOf(Eq(Endians{SPV_ENDIANNESS_LITTLE}),
-                               Eq(Endians{SPV_ENDIANNESS_BIG})));
-  EXPECT_THAT(headers(), Eq(Sentences{ExpectedHeaderForBound(1)}));
-  EXPECT_THAT(instructions(), Eq(Instructions{}));
+  EXPECT_HEADER(1).WillOnce(Return(SPV_SUCCESS));
+  Parse(binary, SPV_SUCCESS);
+  EXPECT_EQ(nullptr, diagnostic_);
```
dneto0: The old version of the test also ensured that the instruction callback was never called.

So should add 
```
EXPECT_CALL(client_, Instruction(_)).Times(0)
```

I assume we can say Times(0).

test/BinaryParse.cpp
```diff
@@ -274,41 +232,27 @@ TEST_F(BinaryParseTest, TwoScalarTypesGenerateTwoInstructionCallbacks) {
...
-  num_passing_callbacks_ = 0;
-
-  spv_diagnostic diagnostic = nullptr;
-  EXPECT_EQ(SPV_ERROR_INVALID_BINARY,
-            spvBinaryParse(context, this, binary.data(), binary.size(), Header,
-                           Instruction, &diagnostic));
+  InSequence calls_expected_in_specific_order;
+  EXPECT_HEADER(3).WillOnce(Return(SPV_ERROR_INVALID_BINARY));
+  // Early exit means no calls to Instruction().
+  EXPECT_CALL(client_, Instruction(_)).Times(0);
```
dneto0: Right.

dmircevski:
Addressed the feedback so far.

test/BinaryParse.cpp
dneto0: ? I don't understand.
That's a constructor cast, which is in C++ but not in C.
(And Andrew found out last week the Google style guide prefers static_cast<uint16_t>(...) here, instead of the constructor cast.  But we have not been doing that very much in this codebase so far.)

test/UnitSPIRV.h
```diff
@@ -99,6 +99,7 @@ class WordVector {
 
 inline void PrintTo(const WordVector& words, ::std::ostream* os) {
   size_t count = 0;
+  const auto saved_flags = os->flags();
```
dneto0: The width is reset after operator<< on "value".
(See Notes in http://en.cppreference.com/w/cpp/io/manip/setw)
However, you're right that std::hex isn't, so it's useful to save and restore the flags.

However, my reading of the reference pages says the fill character is not reset automatically. So we have to save and restore that too.

But these are preexisting problems with the code, so this CL is a step forward.

test/BinaryParse.cpp
dmircevski: We've actually known for a while that the style guide discourages constructor casts (to set apart conversions from bitwise casting).  It's just the [rule](https://google.github.io/styleguide/cppguide.html#Casting)'s awkward wording that makes it sound like a C cast.

In the spirit of adopting the Google style guide over time, I've changed all the constructor casts to `static_cast`.

test/UnitSPIRV.h
dneto0: FWIW, the style guide says to use "Use explicit formatting functions rather than stream manipulators or formatting flags to control formatting details such as number base, precision, or padding."
However, what are the formatting functions of which they speak, if not sprintf-like behaviour which is rather low level, or the output stream calls ??  Count me mystified.

test/BinaryParse.cpp
```diff
@@ -384,25 +316,14 @@ TEST_F(BinaryParseTest, EarlyReturnWithOnePassingCallback) {
...
       "%3 = OpTypeFloat 32");
-
-  num_passing_callbacks_ = 1;
-  fail_code_ = SPV_REQUESTED_TERMINATION;
-  spv_diagnostic diagnostic = nullptr;
-  EXPECT_EQ(SPV_REQUESTED_TERMINATION,
-            spvBinaryParse(context, this, binary.data(), binary.size(), Header,
-                           Instruction, &diagnostic));
+  InSequence calls_expected_in_specific_order;
+  EXPECT_HEADER(4).WillOnce(Return(SPV_REQUESTED_TERMINATION));
```
dneto0: This test is OnePassingCallback, so the header call should return SPV_SUCCESS, and we should get exactly one call to the instruction callback, which should have the correct contents for the  OpTypeVoid, and then itself should return SPV_REQUESTED_TERMINATION.

dneto0:
There's still one regression with respect to the original form of the tests.  Once that's fixed, it's good to go.
For bonus points, save and restore the fill character in the PrintTo function.

test/BinaryParse.cpp
dmircevski: Done.

dneto0:
+2
And thanks.  I'll add more extensive mocking to my toolbox!

dmircevski:
Squashed & merged.

PR #35, opened 2015-11-30
test binary parsing where the string operand is shorter than implied by the word count

dneto0:
There are two commits in this PR.

The first adds a test case where the string operand occurs last in the instruction, but has fewer words than implied by the instruction word count.

The second adds short string test cases where the string operand is not the last operand in an instruction.  This only occurs with OpDecorate and OpMemberDecorate for LinkageAttributes decorations.  A LinkageAttributes decoration takes a literal string operand, followed by an attributes enum word.  It should take nothing else.  But to enforce that, we have to tweak the grammar to drop the OperandVariableLiterals at the end of OpDecorate and OpMemberDecorate when parsing opcode.inc.  So I did that and also wrote a few more failing assembler and disassembler cases for OpDecorate and OpMemberDecorate.  OpMemberDecorate did not have any test cases before; I cloned a big set from the OpDecorate set.  (I cloned rather them rather than further complexify the OpDecorate tests via parameterization.)  Yes, I made up the word "complexify".

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -99,9 +99,18 @@ spv_operand_type_t convertOperandClassToType(SpvOp opcode,
         break;
     }
   } else if (operandClass == OperandVariableLiterals) {
-    if (opcode == SpvOpConstant || opcode == SpvOpSpecConstant) {
-      // The number type is determined by the type Id operand.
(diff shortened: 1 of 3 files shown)


antiagainst:
LGTM.

PR #36, opened 2015-12-01
Add -frtti for tests that use mocking

dneto0:
The tests now use mocked methods.  This requires runtime type
information for those test classes.
https://github.com/google/googletest/issues/610

This has been fixed recently upstream in googletest.  Until we pick
up that fix, add -frtti for the the test executable only.


dmircevski:
+2

PR #38, opened 2015-12-01
Don't add -frtti for tests. Use newer googletest.

dneto0:
Using -frtti caused link failures when both of the following are
in effect:
	-DDISABLE_EXCEPTIONS=ON
	-DDISABLE_RTTI=ON

The correct fix is to use tip-of-tree googletest.
Specifically, we need a version of googletest with the fix in
https://github.com/google/googletest/pull/612
In particular, it must be later than googletest 1.7.0.


dneto0:
dup of 37 (!)

PR #40, opened 2015-12-02
Fix namespace on PrintTo

dneto0:


AWoloszyn:
LGTM

umar456:
Wouldn't this be handled by Koenig lookup?

dneto0:
I think it should.  But MSVC 2013 won't compile it, so we have to give it some help.  :-(

PR #42, opened 2015-12-04
Added Float16, and HexFloat conversions

AWoloszyn:


antiagainst:
Nice job before I even read it!

include/util/hex_float.h
```diff
@@ -39,6 +39,16 @@
 
 namespace spvutils {
 
+class Float16 {
+ public:
+  Float16(uint16_t v): val(v) {}
+  Float16() {}
```
dneto0: redundant?

include/util/hex_float.h
```diff
@@ -145,12 +171,36 @@ template <>
 struct HexFloatTraits<FloatProxy<double>> {
   typedef uint64_t uint_type;
   typedef int64_t int_type;
+  typedef FloatProxy<double> underlying_type;
```
dneto0: Could we make all these typedefs using declarations?

include/util/hex_float.h
AWoloszyn: The default constructor will not get created if we specify a constructor, so in order to have one, we have to write it.

include/util/hex_float.h
AWoloszyn: Done

include/util/hex_float.h
dneto0: ok

include/util/hex_float.h
```diff
@@ -39,19 +39,43 @@
 
 namespace spvutils {
 
+class Float16 {
+ public:
+  Float16(uint16_t v): val(v) {}
+  Float16() {}
+  static bool isNan(const Float16 val) {
+    return (val.val & 0x7FFF) == 0x7C00;
```
dneto0: I think that's the is-infinity test.
isnan should be 
(val.val & 0x7C00 == 0x7C00) && (val.val & 3FF != 0)

include/util/hex_float.h
```diff
@@ -111,9 +135,9 @@ std::istream& operator>>(std::istream& is, FloatProxy<T>& value) {
 template <typename T>
 struct HexFloatTraits {
   // Integer type that can store this hex-float.
-  typedef void uint_type;
+  using uint_type = void;
   // Signed integer type that can store this hex-float.
-  typedef void int_type;
+  using int_type = void;
   // The number of bits that are actually relevant in the uint_type.
```
dneto0: Optional:  Out of symmetry with the previous two, perhaps should also have, along with comments saying what they are.
  using underlying_type = void;
  using native_type = void;

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
   static const uint32_t fraction_right_shift =
       (sizeof(uint_type) * 8) - num_fraction_bits;
 
+  // The maximum representable normalized exponent.
+  static const int_type max_exponent =
```
dneto0: So for float 16, this is 31-15 = 16.   That's the exponent for infinity and NaN.
Maybe you need a different name or comment, since the example at https://en.wikipedia.org/wiki/Half-precision_floating-point_format#Exponent_encoding
shows Emin  as -14 and Emax as 15.  Those are for normal and subnormal numbers.

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+      (exponent_mask >> num_fraction_bits) - exponent_bias;
+  // The minimum representable exponent for normalized numbers.
+  static const int_type min_exponent = -exponent_bias;
```
dneto0: Min biased exponent for normal or subnormal numbers is 1, so 1-15 = -14 is min exponent.  Not -15.
Or maybe you need a different comment?

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+  // Returns the bits associated with the exponent, shifted to start at the
+  // lsb of the type.
+  const uint_type getExponentBits() const {
+    return (getBits() & exponent_mask) >> num_fraction_bits;
+  }
+
+  // Returns the exponent in unbiased form. This is the exponent in the
+  // human-friendly form.
+  const int_type getUnbiasedExponent() const {
+    return (getExponentBits() - exponent_bias);
```
dneto0: I'd probably  static_cast<int_type>(getExponentBits()) before doing the subtraction.

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+  // Returns just the significand bits from the value.
+  const uint_type getSignificandBits() const {
+    return getBits() & fraction_encode_mask;
+  }
+
+  // If the number was normalized, returns the unbiased exponent.
+  // If the number was denormal, normalize the exponent first.
+  const int_type getUnbiasedNormalizedExponent() const {
+    uint_type non_sign_bits = getBits() & ~sign_mask;
+    if ((non_sign_bits & ~sign_mask) == 0) {  // special case if everything is 0
```
dneto0: Apply absorption, to simplify to
if ( non_sign_bits == 0) { ...

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+      return 0;
+    }
+    int_type exp = getUnbiasedExponent();
+    if (exp == -exponent_bias &&
+        non_sign_bits != 0) {  // We are in denorm land.
```
dneto0: Already determined that non_sign_bits is not zero, so can skip the second clause.

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+
+  // If the number was normalized, returns the unbiased exponent.
+  // If the number was denormal, normalize the exponent first.
+  const int_type getUnbiasedNormalizedExponent() const {
+    uint_type non_sign_bits = getBits() & ~sign_mask;
+    if ((non_sign_bits & ~sign_mask) == 0) {  // special case if everything is 0
+      return 0;
+    }
+    int_type exp = getUnbiasedExponent();
+    if (exp == -exponent_bias &&
```
dneto0: simplify to   if (exp == min_exponent) {

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+    }
+    return exp;
+  }
+
+  // Returns the signficand after it has been normalized.
+  const uint_type getNormalizedSignificand() const {
+    int_type unbiased_exponent = getUnbiasedNormalizedExponent();
+    uint_type significand = getSignificandBits();
+    for (int_type i = unbiased_exponent;
+         i <= -static_cast<int_type>(exponent_bias); ++i) {
```
dneto0: simplify to i <= min_exponent

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+      significand <<= 1;
+    }
+    significand &= fraction_encode_mask;
+    return significand;
+  }
+
+  // Returns true if this number represents a negative value.
+  bool isNegative() const { return (getBits() & sign_mask) != 0; }
+
+  // Note this assumes EVERY significand is normalized, and has an implicit
```
dneto0: Blurb should start out with what the function does. Tricky details come after.

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+  // 0 actually means 0.
+  // If underflow_round_up is set, then on underflow, if a number is non-0
+  // and would underflow, we round up to the smallest denorm.
+  void setFromSignUnbiasedExponentAndNormalizedSignificand(
+      bool negative, int_type exponent, uint_type significand,
+      bool round_denorm_up) {
+    const int_type negative_bias = -static_cast<int_type>(exponent_bias);
+    bool significand_is_zero = significand == 0;
+
+    if (exponent <= negative_bias) {
```
dneto0: if ( exponent <= min_exponent )

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+      // If this was dernom, then we have to shift the bit on, meaning
```
dneto0: "denorm"

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+    }
+    significand &= fraction_encode_mask;
+    return significand;
+  }
+
+  // Returns true if this number represents a negative value.
+  bool isNegative() const { return (getBits() & sign_mask) != 0; }
+
+  // Note this assumes EVERY significand is normalized, and has an implicit
+  // leading one. This means that the only way that this method will set 0,
```
dneto0: Interesting.   So this method should not be used with raw bits extracted from a subnormal number, since only normal numbers have the implicit leading 1.  I'd add something to that effect after the "This means that".  E.g.
"Also, don't use this method with raw bits extracted from a subnormal number, since subnormals don't have leading implicit 1 in the significand."

 https://en.wikipedia.org/wiki/Half-precision_floating-point_format

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+
+    // put it all together
+    exponent = (exponent << exponent_left_shift) & exponent_mask;
+    significand &= fraction_encode_mask;
+    value |= exponent | significand;
+    value_ = BitwiseCast<T>(value);
+  }
+
+  // Increments the significand of this number by the given amount.
+  // If this would spill the significand into the implicit bit,
```
dneto0: Also state the assumption both significand and to_increment are within bounds of valid significands for the current type.
That assumption is used when overflow is checked by just looking at the first exponent bit.

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+      *carry = true;
+      // The implicit 1-bit will have carried, so we should zero-out the
+      // top bit and shift back.
+      significand &= ~first_exponent_bit;
+      significand >>= 1;
+    }
+    return significand;
+  }
+
+  // Returns the significand, rounded to fit in a significand of
```
dneto0: Blurb doesn't quite line up with the method signature.
The num_bits is the num_bits for the other_T ?

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+  template <typename other_T>
+  typename other_T::uint_type getRoundedNormalizedSignificand(
+      round_direction dir, bool* carry_bit) {
+    using other_uint_type = typename other_T::uint_type;
+    static const int_type num_throwaway_bits =
+        static_cast<int_type>(num_fraction_bits) -
+        static_cast<int_type>(other_T::num_fraction_bits);
+
+    static const uint_type last_significant_bit =
+        (num_throwaway_bits < 0) ? 0 : static_cast<uint_type>(1u)
```
dneto0: Strongly recommend parens around the shift expression.

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+      return out_val;
+    }
+
+    // If every non-representable bit is 0, then we don't have any casting to
+    // do.
+    if ((significand & throwaway_mask) == 0) {
+      return static_cast<other_uint_type>(significand >> num_throwaway_bits);
+    }
+
+    // We are down-casting here.
```
dneto0: Meaning we are narrowing the significand, with some form of rounding.

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+      }
+      return;
+    }
+    uint_type significand = getSignificandBits();
+    bool carried = false;
+    typename other_T::uint_type rounded_significand =
+        getRoundedNormalizedSignificand<other_T>(round_dir, &carried);
+
+    int_type exponent = getUnbiasedExponent();
+    if (exponent == -static_cast<int_type>(exponent_bias)) {
```
dneto0: clearer as if(exponent == min_exponent)

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
 
         })));
 
+int32_t unbiased_exponent(double f) {
+  return spvutils::HexFloat<spvutils::FloatProxy<float>>(
+      static_cast<float>(f)).getUnbiasedNormalizedExponent();
+}
+
+TEST(HexFloatOperationTest, UnbiasedExponent) {
```
dneto0: good

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+  EXPECT_EQ(-128, unbiased_exponent(ldexp(1.0f, -128)));
+  EXPECT_EQ(-129, unbiased_exponent(ldexp(1.0f, -129)));
+  EXPECT_EQ(-140, unbiased_exponent(ldexp(1.0f, -140)));
+  // Smallest representable number
+  EXPECT_EQ(-126 - 23, unbiased_exponent(ldexp(1.0f, -126 - 23)));
+  // Should get rounded to 0 first.
+  EXPECT_EQ(0, unbiased_exponent(ldexp(1.0f, -127 - 23)));
+}
+
+// Creates a float the is the sum of 1/(2 ^ fractions[i]) for i in factions
```
dneto0: "....a float that is the..."

test/HexFloat.cpp
dneto0: I wonder if it should be a template function over <float>, then you can test the float16 case?

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+  f.setFromSignUnbiasedExponentAndNormalizedSignificand(
+      negative, unbiased_exponent, significand, round_denorm_up);
+  return f.value().getAsFloat();
+}
+
+TEST(HexFloatOperationTests,
+     SetFromSignUnbiasedExponentAndNormalizedSignificand) {
+
+  EXPECT_EQ(1.f, set_from_sign(false, 0, 0, false));
+  EXPECT_EQ(static_cast<float>(ldexp(1.f, -149)), set_from_sign(false, -149, 0, false));
```
dneto0: Would be nice to say these next 4 cases show how rounding up denorms works.

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+  EXPECT_EQ(ldexp(1.0f, -127), set_from_sign(false, -127, 0, false));
+  EXPECT_EQ(ldexp(1.0f, -128), set_from_sign(false, -128, 0, false));
+  EXPECT_EQ(float_fractions({0, 1, 2, 5}),
+            set_from_sign(false, 0, bits_set({0, 1, 4}), false));
+  EXPECT_EQ(ldexp(float_fractions({0, 1, 2, 5}), -32),
+            set_from_sign(false, -32, bits_set({0, 1, 4}), false));
+  EXPECT_EQ(ldexp(float_fractions({0, 1, 2, 5}), -128),
+            set_from_sign(false, -128, bits_set({0, 1, 4}), false));
+
+  EXPECT_EQ(-1.f, set_from_sign(true, 0, 0, false));
```
dneto0: This group is all the negative sign cases.

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+  static uint_type incrementSignificand(uint_type significand,
+                                         uint_type to_increment, bool* carry) {
+    significand += to_increment;
+    if (significand & first_exponent_bit) {
+      *carry = true;
+      // The implicit 1-bit will have carried, so we should zero-out the
+      // top bit and shift back.
+      significand &= ~first_exponent_bit;
+      significand >>= 1;
+    }
```
dneto0: Should have else  { *carry = false; }

include/util/hex_float.h
```diff
@@ -213,12 +270,310 @@ class HexFloat {
...
+    exponent = (exponent << exponent_left_shift) & exponent_mask;
+    significand &= fraction_encode_mask;
+    value |= exponent | significand;
+    value_ = BitwiseCast<T>(value);
+  }
+
+  // Increments the significand of this number by the given amount.
+  // If this would spill the significand into the implicit bit,
+  // carry is set to true and the significand is shifted to fit into
+  // the correct location.
```
dneto0: carry should always be set.

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+};
+
+using HexFloatRoundTest =
+    ::testing::TestWithParam<RoundSignficandTest>;
+
+TEST_P(HexFloatRoundTest, RoundDownToFP16) {
+  using HF = spvutils::HexFloat<spvutils::FloatProxy<float>>;
+  using HF16 = spvutils::HexFloat<spvutils::FloatProxy<spvutils::Float16>>;
+
+  bool carry_bit = false;
```
dneto0: move  deffinition of carry_bit down, or eliminate the assignment at line 713

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+    EXPECT_EQ(
+        bits_set({0, 1, 4, 22}),
+        HF(static_cast<float>(float_fractions({0, 1, 2, 5, 23})))
+            .getRoundedNormalizedSignificand<HF>(round, &carry_bit));
+    EXPECT_FALSE(carry_bit);
+  }
+}
+
+using rd = spvutils::round_direction;
+struct RoundSignficandTest {
```
dneto0: Our convention seems to be this should be named RoundSignificandCase

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+
+    EXPECT_EQ(
+        bits_set({0, 1, 4, 22}),
+        HF(static_cast<float>(float_fractions({0, 1, 2, 5, 23})))
+            .getRoundedNormalizedSignificand<HF>(round, &carry_bit));
+    EXPECT_FALSE(carry_bit);
+  }
+}
+
+using rd = spvutils::round_direction;
```
dneto0: Recommend RD instead of rd.

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+    // Cases where original number was denorm. Note: this should have no effect
+    // the number is pre-normalized.
+    {static_cast<float>(ldexp(float_fractions({0, 1, 11, 13}), -128)), std::make_pair(half_bits_set({0}), false), rd::to_zero},
+    {static_cast<float>(ldexp(float_fractions({0, 1, 11, 13}), -129)), std::make_pair(half_bits_set({0, 9}), false), rd::to_positive_infinity},
+    {static_cast<float>(ldexp(float_fractions({0, 1, 11, 13}), -131)), std::make_pair(half_bits_set({0}), false), rd::to_negative_infinity},
+    {static_cast<float>(ldexp(float_fractions({0, 1, 11, 13}), -130)), std::make_pair(half_bits_set({0, 9}), false), rd::to_nearest_even},
+   })));
+// clang-format on
+
+struct UpCastSignficandTest {
```
dneto0: UpCastSignifcandCase

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+        << input_value.getRoundedNormalizedSignificand<HF>(round, &carry_bit)
+        << "  0x" << GetParam().expected_result;
+    EXPECT_FALSE(carry_bit);
+  }
+}
+
+INSTANTIATE_TEST_CASE_P(F16toF32, HexFloatRoundUpSignificandTest,
+  ::testing::ValuesIn(std::vector<UpCastSignficandTest>(
+  {
+    {0x3F00, 0x600000},
```
dneto0: Would be nice to have a comment saying that 0xFC00 of the source 16-bit hex float covers the sign and exponent, so those bits are ignored for this test.

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+    {static_cast<float>(ldexp(1.0f, -24)), 0x1, {rd::to_zero, rd::to_positive_infinity, rd::to_negative_infinity, rd::to_nearest_even}},
+
+    // Overflow
+    {static_cast<float>(ldexp(1.0f, 16)), positive_infinity, {rd::to_zero, rd::to_positive_infinity, rd::to_negative_infinity, rd::to_nearest_even}},
+    {static_cast<float>(ldexp(1.0f, 18)), positive_infinity, {rd::to_zero, rd::to_positive_infinity, rd::to_negative_infinity, rd::to_nearest_even}},
+    {static_cast<float>(ldexp(1.3f, 16)), positive_infinity, {rd::to_zero, rd::to_positive_infinity, rd::to_negative_infinity, rd::to_nearest_even}},
+    {static_cast<float>(-ldexp(1.0f, 16)), negative_infinity, {rd::to_zero, rd::to_positive_infinity, rd::to_negative_infinity, rd::to_nearest_even}},
+    {static_cast<float>(-ldexp(1.0f, 18)), negative_infinity, {rd::to_zero, rd::to_positive_infinity, rd::to_negative_infinity, rd::to_nearest_even}},
+    {static_cast<float>(-ldexp(1.3f, 16)), negative_infinity, {rd::to_zero, rd::to_positive_infinity, rd::to_negative_infinity, rd::to_nearest_even}},
+
```
dneto0: Missing transfer of NaNs

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+    // Transfer of Infinities
+    {std::numeric_limits<float>::infinity(), positive_infinity, {rd::to_zero, rd::to_positive_infinity, rd::to_negative_infinity, rd::to_nearest_even}},
+    {-std::numeric_limits<float>::infinity(), negative_infinity, {rd::to_zero, rd::to_positive_infinity, rd::to_negative_infinity, rd::to_nearest_even}},
+  })));
+
+struct UpCastTest{
```
dneto0: UpCastCase

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+    {0x3F01, float_fractions({0, 1, 2, 10})},
+    {0xBF01, -float_fractions({0, 1, 2, 10})},
+
+    // denorm
+    {0x0001, static_cast<float>(ldexp(1.0, -24))},
+    {0x0002, static_cast<float>(ldexp(1.0, -23))},
+    {0x8001, static_cast<float>(-ldexp(1.0, -24))},
+    {0x8011, static_cast<float>(-ldexp(1.0, -20) + -ldexp(1.0, -24))},
+
+    // inf
```
dneto0: NaNs?

dneto0:
Ok.  Two functional issues:
The float 16 isNaN test looks wrong.
And incrementSignificand should always set *carry to something so that a previous value doesn't persist.

AWoloszyn:
@dneto0  Addressed feedback so far.

include/util/hex_float.h
```diff
@@ -39,19 +39,37 @@
 
 namespace spvutils {
 
+class Float16 {
+ public:
+  Float16(uint16_t v) : val(v) {}
+  Float16() {}
```
antiagainst: `= default`?

include/util/hex_float.h
```diff
@@ -39,19 +39,37 @@
...
+  static bool isNan(const Float16 val) {
+    return ((val.val & 0x7C00) == 0x7C00) && ((val.val & 0x3FF) != 0);
+  }
+  uint16_t val;
```
antiagainst: `val` should be `private`.

AWoloszyn:
@antiagainst  Done

include/util/hex_float.h
```diff
@@ -213,12 +268,312 @@ class HexFloat {
...
+  // Returns just the significand bits from the value.
+  const uint_type getSignificandBits() const {
+    return getBits() & fraction_encode_mask;
+  }
+
+  // If the number was normalized, returns the unbiased exponent.
+  // If the number was denormal, normalize the exponent first.
+  const int_type getUnbiasedNormalizedExponent() const {
+    uint_type non_sign_bits = getBits() & ~sign_mask;
+    if (non_sign_bits == 0) {  // special case if everything is 0
```
antiagainst: `non_sign_bits` is only used here. Consider use `getBits() & ~sign_mask` directly?

AWoloszyn:
Rebased.

antiagainst:
consider putting `clang-format: off` at appropriate places and run `clang-format`?

test/HexFloat.cpp
```diff
@@ -528,5 +528,442 @@ INSTANTIATE_TEST_CASE_P(
...
+    HF16(0xFC01).castTo(f, round);
+    EXPECT_TRUE(f.value().isNan());
+    HF16(0x7C10).castTo(f, round);
+    EXPECT_TRUE(f.value().isNan());
+    HF16(0xFF00).castTo(f, round);
+    EXPECT_TRUE(f.value().isNan());
+  }
+}
+
 // TODO(awoloszyn): Add fp16 tests and HexFloatTraits.
```
antiagainst: Is this line still necessary?

include/util/hex_float.h
```diff
@@ -39,19 +39,44 @@
...
+  }
+  uint16_t get_value() const {
+    return val;
+  }
+private:
+  uint16_t val;
+};
+
 template <typename T>
 struct FloatProxyTraits {
```
dmircevski: This could use a blurb now, describing how exactly to specialize it.

include/util/hex_float.h
```diff
@@ -143,14 +174,38 @@ struct HexFloatTraits<FloatProxy<float>> {
...
 };
 
+// Traits for IEEE half.
+// 1 sign bit, 5 exponent bits, 10 fractional bits.
+template <>
+struct HexFloatTraits<FloatProxy<Float16>> {
+  using uint_type = uint16_t;
+  using int_type = int16_t;
+  using underlying_type = uint16_t;
+  using native_type = uint16_t;
```
dmircevski: This is surprising.  Why isn't it `FloatProxy<Float16>>` and `Float16`?

include/util/hex_float.h
```diff
@@ -143,14 +174,38 @@ struct HexFloatTraits<FloatProxy<float>> {
...
+  static const uint_type num_used_bits = 16;
+  static const uint_type num_exponent_bits = 5;
+  static const uint_type num_fraction_bits = 10;
+  static const uint_type exponent_bias = 15;
+};
+
+enum class round_direction {
+  to_zero,
```
dmircevski: Consider naming these as constants: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

include/util/hex_float.h
```diff
@@ -190,10 +247,15 @@ class HexFloat {
       spvutils::SetBits<uint_type, 0,
                         num_fraction_bits + num_overflow_bits>::get;
 
-  // The topmost bit in the fraction. (The first non-implicit bit).
+  // The topmost bit in the fraction in the nibble-aligned fraction.
```
dmircevski: Unclear: fraction in the fraction?

test/HexFloat.cpp
```diff
@@ -528,5 +528,468 @@ INSTANTIATE_TEST_CASE_P(
 
         })));
 
+int32_t unbiased_exponent(double f) {
```
dmircevski: Blurb needed, explaining why f is double and then cast to float.

include/util/hex_float.h
```diff
@@ -213,12 +275,312 @@ class HexFloat {
...
+  }
+
+  // Returns just the significand bits from the value.
+  const uint_type getSignificandBits() const {
+    return getBits() & fraction_encode_mask;
+  }
+
+  // If the number was normalized, returns the unbiased exponent.
+  // If the number was denormal, normalize the exponent first.
+  const int_type getUnbiasedNormalizedExponent() const {
```
dmircevski: This hardly belongs to the HexFloat class -- it's unrelated to hex serialization.

include/util/hex_float.h
AWoloszyn: This is because both underlying_type and native_type represent numerical types, which Float16 is not. IF Float16 was 16-bits then all of this would just work out, but Float16 is a 32-bit value containing a 16-bit value we care about.

include/util/hex_float.h
AWoloszyn: Will address in a future CL.

dneto0:
Changes are good, and I like the cleanups made.

Verified that all @dmircevski comments addressed.
The @antiagainst  still has one unaddressed, to remove the TODO. That's relatively minor though.

+2

AWoloszyn:
Rebased for merge.

PR #43, opened 2015-12-09
Fix setting of off-by-default cmake options

dneto0:
An enclosing CMake project should be able to set the
off-by-default CMake options in SPIRV-Tools by just doing:
  set(SPIRV_SKIP_EXECUTABLES ON)
instead of
  set(SPIRV_SKIP_EXECUTABLES ON CACHE BOOL "" FORCE)

Also, fix the SPIRV_WARN_EVERYTHING option processing so Cmake understands which options
to send to Clang vs. GCC.

Note: With SPIRV_WARN_EVERYTHING enabled, the code doesn't
compile with either Clang or GCC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,10 +47,17 @@ option(SPIRV_WERROR "Enable error on warning" ON)
 if(UNIX)
   set(SPIRV_WARNINGS -Wall -Wextra -Wno-missing-field-initializers)
 
-  option(SPIRV_WARN_EVERYTHING "Enable -Weverything" OFF)
+  option(SPIRV_WARN_EVERYTHING "Enable -Weverything" ${SPIRV_WARN_EVERYTHING})
   if(${SPIRV_WARN_EVERYTHING})
-    set(SPIRV_WARNINGS ${SPIRV_WARNINGS}
-      -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded)
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+      set(SPIRV_WARNINGS ${SPIRV_WARNINGS}
+        -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded)
+    elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+      set(SPIRV_WARNINGS ${SPIRV_WARNINGS} -Wpedantic -pedantic-errors)
+    else()
+      message(STATUS "Unknown compiler ${CMAKE_CXX_COMPILER_ID}, "
+                     "so SPIRV_WARN_EVERYTHING has no effect")
+    endif()
   endif()
 
   if(${SPIRV_WERROR})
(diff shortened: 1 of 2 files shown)


dneto0:
Rebased against master

CMakeLists.txt
```diff
@@ -47,10 +47,17 @@ option(SPIRV_WERROR "Enable error on warning" ON)
...
+  option(SPIRV_WARN_EVERYTHING "Enable -Weverything" ${SPIRV_WARN_EVERYTHING})
   if(${SPIRV_WARN_EVERYTHING})
-    set(SPIRV_WARNINGS ${SPIRV_WARNINGS}
-      -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded)
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+      set(SPIRV_WARNINGS ${SPIRV_WARNINGS}
+        -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-padded)
+    elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+      set(SPIRV_WARNINGS ${SPIRV_WARNINGS} -Wpedantic -pedantic-errors)
+    elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
```
antiagainst: I assume you mean `else()` here?

README.md
```diff
@@ -100,8 +100,10 @@ The following CMake options are supported:
   version of `clang` this option enables the use of the sanitizers documented
   [here](http://clang.llvm.org/docs/UsersManual.html#controlling-code-generation).
   This should only be used with a debug build. Disabled by default.
-* `SPIRV_WARN_EVERYTHING=OFF` - On UNIX platforms enable the `-Weverything`
-  compiler front end option, disabled by default.
+* `SPIRV_WARN_EVERYTHING=OFF` - On UNIX platforms enable more strict warnings,
```
antiagainst: Actually I'd like to see these options to be represented as `SPIRV_WARN_EVERYTHING={ON|OFF}` and then have `{dis|en}abled by default` at the end of the sentence. Right now it reads like we only support `OFF`.

antiagainst: BTW, on L97, `SPIRV_SKIP_EXECUTATBLES` is `OFF` by default.

CMakeLists.txt
dneto0: Yikes. you're right.

README.md
dneto0: Reworded the settings for each option, as suggested.

README.md
```diff
@@ -93,17 +93,21 @@ development environment.
...
+* `SPIRV_USE_SANITIZER=<sanitizer>`, default is no sanitizing - On UNIX platforms
+  with an appropriate version of `clang` this option enables the use of the
+  sanitizers documented
   [here](http://clang.llvm.org/docs/UsersManual.html#controlling-code-generation).
-  This should only be used with a debug build. Disabled by default.
-* `SPIRV_WARN_EVERYTHING=OFF` - On UNIX platforms enable the `-Weverything`
-  compiler front end option, disabled by default.
-* `SPIRV_WERROR=ON` - Forces a compilation error on any warnings encountered by
-  enabling the compiler-specific compiler front-end option, enabled by default.
+  This should only be used with a debug build.  Disabled by default.
```
antiagainst: One final thing: redundant `Disabled by default` at the end.

dneto0: D'oh. Sorry
Fixed

antiagainst:
+2

PR #46, opened 2015-12-14
Fix tests which rely on OpName; Use ASSERT_THAT

umar456:
All OpName calls should appear before annotation and type
declarations. This fixes the ordering of the `OpName` instructions
in the SSA unit tests

This PR also replaces the `ContainsString` function with the
google_mock's `HasSubstr` and `ASSERT_THAT` functions.

### DIFF
--- a/test/Validate.SSA.cpp
+++ b/test/Validate.SSA.cpp
@@ -28,23 +28,22 @@
 
 #include "UnitSPIRV.h"
 #include "ValidateFixtures.h"
+#include "gmock/gmock.h"
 
 #include <sstream>
 #include <string>
 #include <utility>
 
+using ::testing::HasSubstr;
+
 using std::string;
 using std::pair;
 using std::stringstream;
+
 namespace {
 
 using Validate =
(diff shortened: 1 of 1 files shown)


test/Validate.SSA.cpp
```diff
@@ -75,7 +74,7 @@ TEST_F(Validate, IdUndefinedBad) {
     )";
   CompileSuccessfully(str);
   ASSERT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
-  ASSERT_TRUE(ContainsString(getDiagnosticString(), "missing"));
+  ASSERT_THAT(getDiagnosticString(), HasSubstr("missing"));
```
AWoloszyn: I would switch these ASSERT_THAT to EXPECT_THAT since this is not something that would invalidate any new expectations we add to the test.

AWoloszyn:
Other than switching ASSERT_THAT to EXPECT_THAT, looks good.

umar456:
Addressed feedback from @AWoloszyn

dneto0:
+2.  I'll squash and merge

dneto0:
squashed and merged into master.  Closing.

PR #47, opened 2015-12-15
Logical layout validation

umar456:
* Validates module level instructions for logical layout
  conformance
* Does not validate:
  1. Function logical layout
  2. Minor cases with OpVariable
  3. Missing MemoryModel instruction in module
  4. Order of function definition and function declaration
* 782 unit tests for logical layout

### DIFF
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -436,6 +436,109 @@ void DebugInstructionPass(ValidationState_t& _,
   }
 }
 
+// TODO(umar): Check MemoryModel is in module
+// TODO(umar): Check OpVariable storage class is not function in module section
+// TODO(umar): Make sure function declarations appear before function
+// definitions
+// TODO(umar): Better error messages
+// NOTE: This function does not handle CFG related validation
+// Performs logical layout validation. See Section 2.4
+spv_result_t ModuleLayoutPass(ValidationState_t& _,
+                              const spv_parsed_instruction_t* inst) {
+  if (_.is_enabled(SPV_VALIDATE_LAYOUT_BIT)) {
+    SpvOp opcode = inst->opcode;
+
+    if (libspirv::ModuleLayoutSection::kModule == _.getLayoutStage()) {
+      // Module scoped instructions are processed by determining if the opcode
+      // is part of the current stage. If it is not then the next stage is
+      // checked.
+      while (_.isOpcodeInCurrentLayoutStage(opcode) == false) {
+        // TODO(umar): Check if the MemoryModel instruction has executed
+        _.progressToNextLayoutStageOrder();
+        if (_.getLayoutStage() == libspirv::ModuleLayoutSection::kFunction) {
+          // All module stages have been processed. Recursivly call
+          // ModuleLayoutPass
+          // to process the next section of the module
+          return ModuleLayoutPass(_, inst);
+        }
+      }
+    } else {
+      // Validate the function layout.
+      switch (opcode) {
+        case SpvOpCapability:
+        case SpvOpExtension:
+        case SpvOpExtInstImport:
+        case SpvOpMemoryModel:
+        case SpvOpEntryPoint:
+        case SpvOpExecutionMode:
+        case SpvOpSourceContinued:
+        case SpvOpSource:
+        case SpvOpSourceExtension:
+        case SpvOpString:
+        case SpvOpName:
+        case SpvOpMemberName:
+        case SpvOpDecorate:
+        case SpvOpMemberDecorate:
+        case SpvOpGroupDecorate:
+        case SpvOpGroupMemberDecorate:
+        case SpvOpDecorationGroup:
+        case SpvOpTypeVoid:
+        case SpvOpTypeBool:
+        case SpvOpTypeInt:
+        case SpvOpTypeFloat:
+        case SpvOpTypeVector:
+        case SpvOpTypeMatrix:
+        case SpvOpTypeImage:
+        case SpvOpTypeSampler:
+        case SpvOpTypeSampledImage:
+        case SpvOpTypeArray:
+        case SpvOpTypeRuntimeArray:
+        case SpvOpTypeStruct:
+        case SpvOpTypeOpaque:
+        case SpvOpTypePointer:
+        case SpvOpTypeFunction:
+        case SpvOpTypeEvent:
+        case SpvOpTypeDeviceEvent:
+        case SpvOpTypeReserveId:
+        case SpvOpTypeQueue:
+        case SpvOpTypePipe:
+        case SpvOpTypeForwardPointer:
+        case SpvOpConstantTrue:
+        case SpvOpConstantFalse:
+        case SpvOpConstant:
+        case SpvOpConstantComposite:
+        case SpvOpConstantSampler:
+        case SpvOpConstantNull:
+        case SpvOpSpecConstantTrue:
+        case SpvOpSpecConstantFalse:
+        case SpvOpSpecConstant:
+        case SpvOpSpecConstantComposite:
+        case SpvOpSpecConstantOp:
+          return _.diag(SPV_ERROR_INVALID_LAYOUT) << "Invalid Layout";
+        case SpvOpVariable: {
+          const uint32_t* storage_class =
+              inst->words + inst->operands[2].offset;
+          if (*storage_class != SpvStorageClassFunction)
+            return _.diag(SPV_ERROR_INVALID_LAYOUT)
+                   << "All OpVariable instructions in a function must have a "
+                      "storage class of Function[7]";
+          break;
+        }
+        default:
+          return SPV_SUCCESS;
+      }
+    }
+  }
+  return SPV_SUCCESS;
+}
+
+// Shame
+#define CHECK_RESULT(EXPRESSION)                \
+  do{                                           \
+    spv_result_t ret = EXPRESSION;              \
+    if(ret) return ret;                         \
+} while(false);
+
 spv_result_t ProcessInstructions(void* user_data,
                                  const spv_parsed_instruction_t* inst) {
   ValidationState_t& _ = *(reinterpret_cast<ValidationState_t*>(user_data));
@@ -447,10 +550,13 @@ spv_result_t ProcessInstructions(void* user_data,
   DebugInstructionPass(_, inst);
 
   // TODO(umar): Perform CFG pass
-  // TODO(umar): Perform logical layout validation pass
   // TODO(umar): Perform data rules pass
   // TODO(umar): Perform instruction validation pass
-  return SsaPass(_, can_have_forward_declared_ids, inst);
+  spv_result_t ret = SPV_SUCCESS;
+  CHECK_RESULT(ModuleLayoutPass(_, inst))
+  CHECK_RESULT(SsaPass(_, can_have_forward_declared_ids, inst))
+
+  return ret;
 }
 
 } // anonymous namespace
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -25,6 +25,7 @@
 // MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
 #include "validate.h"
+#include "validate_types.h"
 
 #include "binary.h"
 #include "diagnostic.h"
@@ -56,6 +57,7 @@ using std::transform;
 using std::unordered_set;
 using std::vector;
 
+using libspirv::ValidationState_t;
 
 #define spvCheckReturn(expression) \
   if (spv_result_t error = (expression)) return error;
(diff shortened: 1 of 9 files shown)


source/validate.cpp
```diff
@@ -518,6 +436,109 @@ void DebugInstructionPass(ValidationState_t& _,
   }
 }
 
+// Allows for piping of multiple functions which return spv_result_t
+spv_result_t operator||(spv_result_t lhs, spv_result_t rhs) {
```
dekimir: Not sure why this is required, given that `spv_result_t` implicitly converts to `bool`.  What happens if you delete this function?

umar456: You lose the type of error returned by the function. The validation step returns 0 or 1 for all errors.

dekimir: Ah, of course.  Thx. :)

dneto0: I'm ok with having an overload, but the ordinary || operator is short-circuiting, but this implementation isn't.  The RHS will always be evaluated.   Better to make it "operator|".
With that, the precedence changes too, but if that ever matters, use parens to group the expression.

source/validate.cpp
```diff
@@ -518,6 +436,109 @@ void DebugInstructionPass(ValidationState_t& _,
...
+  if (lhs) return lhs;
+  if (rhs) return rhs;
```
dneto0: Alternately ,replace the last two lines withe
```
return rhs;
```
Coder's choice.   Your code has a nicely suggestive parallel structure.

source/validate.cpp
```diff
@@ -518,6 +436,109 @@ void DebugInstructionPass(ValidationState_t& _,
...
+        if (_.getLayoutStage() == libspirv::ModuleLayoutSection::kFunction) {
+          // All module stages have been processed. Recursivly call
+          // ModuleLayoutPass
+          // to process the next section of the module
+          return ModuleLayoutPass(_, inst);
+        }
+      }
+    } else {
+      // Validate the function layout.
+      switch (opcode) {
```
dneto0: OpUndef is missing.  Should go in the function section.  It's not a "constant instruction", despite being listed in the constant-creation instructions section of the spec.

dneto0: My bad.  I didn't read far enough.  These are all the instructions *not* allowed in the function area.

source/validate_types.cpp
```diff
@@ -0,0 +1,204 @@
...
+int ValidationState_t::incrementInstructionCount() {
+  return instruction_counter_++;
+}
+
+ModuleLayoutSection ValidationState_t::getLayoutStage() const {
+  return current_layout_stage_;
+}
+
+void ValidationState_t::progressToNextLayoutStageOrder() {
+  if (++module_layout_order_stage_ >= GetModuleOrder().size()) {
```
dneto0: Quibble.   module_layout_order_stage_ is beyond size, continuing to call this function will continue to increment the value.  Eventually it will wrap around and get back to the start.
Better to increment the value only when it's less than the limit.  Yes, it will make slightly more code, but will be slightly more safe.

source/validate_types.h
```diff
@@ -0,0 +1,122 @@
...
+  std::unordered_set<uint32_t> unresolved_forward_ids_;
+
+  // Validation options to determine the passes to execute
+  uint32_t validation_flags_;
+
+  std::map<uint32_t, std::string> operand_names_;
+
+  // The stage which is being processed by the validation. Partially based on
+  // Section 2.4. Logical Layout of a Module
+  uint8_t module_layout_order_stage_;
```
dneto0: Coder's choice, but my take:   I'd also take a regular int or unsigned.
Google Style guide says to use int by default in most situations.   I personally like unsigned for things that are counters.

source/validate_types.cpp
```diff
@@ -0,0 +1,204 @@
...
+using std::string;
+using std::to_string;
+using std::unordered_set;
+using std::vector;
+
+namespace {
+const vector<vector<SpvOp>>& GetModuleOrder() {
+  // See Section 2.4
+  // clang-format off
+  static const vector<vector<SpvOp>> moduleOrder = {
```
dneto0: FYI:  Google style guide forbids program-duration class variables with static duration.  Reasons given here:

Two options:
1) The easiest thing to do is to make a regular member variable in the validation state to store the value, and initialize it in the constructor.
2) Second thing you could do is make this a static variable that is a plain array-of-arrays, and then use the function form of std::begin, std::end and std::find on the array.

test/Validate.Layout.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+// returns true if order is between MIN and MAX(inclusive)
+template <int MIN, int MAX>
+bool Range(int order) {
+  return order >= MIN && order <= MAX;
+};
+
+// clang-format off
+// SPIRV source used to test the logical layout
+vector<string> instructions = {
```
dneto0: static duration variable is a class.
This I care less about since it's in test code. Similarly, we don't care about performance so you could just make a function that returns this as a newly constructed (not static) value.
I bet you can make this Plain Old Data array and use std::begin and std::end (including std::end(a) - std::begin(a) for the size).

source/validate_types.cpp
umar456: I think that applies to non-local static objects. The order is not defined for objects in different translation units but it is well defined for static variables in function scope. I think the google style guide allows function local variables:

umar456: Google style guide:
"_(This prohibition does not apply to a static variable within function scope, since its initialization order is well-defined and does not occur until control passes through its declaration.)_"

test/Validate.Layout.cpp
dneto0: And now I see how this is used in the GenerateCode function.  I need to think more.

test/Validate.Layout.cpp
```diff
@@ -0,0 +1,199 @@
...
+                        , make_pair(string("OpTypeVector %intt 4"),     Range<16, 26>)
+                        , make_pair(string("OpTypeMatrix %vec4 4"),     Range<16, 26>)
+                        , make_pair(string("OpTypeStruct"),             Range<16, 26>)
+                        , make_pair(string("%vfunct = OpTypeFunction"), Range<16, 26>)
+                        )));
+// clang-format on
+
+// Creates a new vector which removes the string if the substr is found in the
+// instructions vector and reinserts it in the location specified by order.
+// NOTE: This will not work correctly if there are two instances of substr in
```
dneto0: This NOTE is ok.  Alternately you could just say that we do the special thing with the *first* string that has the given substring.
Coder's choice.

test/Validate.Layout.cpp
```diff
@@ -0,0 +1,199 @@
...
+  "OpSourceContinued \"void main(){return;}\"",
+  "OpSourceExtension \"Test extension\"",
+  "OpName %id \"MyID\"",
+  "OpMemberName %struct 1 \"my_member\"",
+  "OpDecorate %dgrp RowMajor",
+  "OpMemberDecorate %struct 1 RowMajor",
+  "%dgrp   = OpDecorationGroup",
+  "OpGroupDecorate %dgrp %mat33 %mat44",
+  "%boolt  =  OpTypeBool",
+  "%voidt  =  OpTypeVoid",
```
dneto0: Recommend adding the missing OpType* instructions, including
OpTypeFloat, OpTypeImage, OpTypeSmapler, OpTypeSampledImage, OpTypeArray, OpTypeRuntimeArray, OpTypeOpaque, ...
Certainly I'm interested in Float, Array

source/validate.cpp
```diff
@@ -518,6 +436,109 @@ void DebugInstructionPass(ValidationState_t& _,
...
+        case SpvOpSpecConstantTrue:
+        case SpvOpSpecConstantFalse:
+        case SpvOpSpecConstant:
+        case SpvOpSpecConstantComposite:
+        case SpvOpSpecConstantOp:
+          return _.diag(SPV_ERROR_INVALID_LAYOUT) << "Invalid Layout";
+        case SpvOpVariable: {
+          const uint32_t* storage_class =
+              inst->words + inst->operands[2].offset;
+          if (*storage_class != 7)
```
dneto0: Replace magic 7 with use of the enum, e.g. static_cast<uint32_t>(SpvStorageClassFunction)

source/validate.cpp
```diff
@@ -518,6 +436,109 @@ void DebugInstructionPass(ValidationState_t& _,
...
+        case SpvOpConstantSampler:
+        case SpvOpConstantNull:
+        case SpvOpSpecConstantTrue:
+        case SpvOpSpecConstantFalse:
+        case SpvOpSpecConstant:
+        case SpvOpSpecConstantComposite:
+        case SpvOpSpecConstantOp:
+          return _.diag(SPV_ERROR_INVALID_LAYOUT) << "Invalid Layout";
+        case SpvOpVariable: {
+          const uint32_t* storage_class =
```
dneto0: It's really a pointer to the storage class value. Could save yourself a few "*" here and there by just dereferencing the value in the initializer:
```
const uint32_t storage_class = inst->words[inst->operands[2].offset]];

File: source/validate_types.cpp (hunk 6 above)
Comment (dneto0) [inline source/validate_types.cpp]: Ok. I stand corrected!  And I'm really glad, because the code is more readable this way.

File: test/Validate.Layout.cpp [hunk 12]
```diff
@@ -0,0 +1,199 @@
...
+  "%viifunct =  OpTypeFunction %voidt %intt %intt",
+  // TODO(umar): OpConstant fails because the type is not defined
+  // TODO(umar): OpGroupMemberDecorate
+  "OpLine %str 3 4",
+  "%func   = OpFunction %voidt None %vfunct",
+  "OpFunctionEnd",
+  "%func2   = OpFunction %voidt None %viifunct",
+  "%funcp1 = OpFunctionParameter %intt",
+  "%funcp2 = OpFunctionParameter %intt",
+  "OpReturn",
```
dneto0: Recommend adding an OpLabel, a function-storage-class variable, a load, a store, and maybe one arithmetic operation.
I'm not sure how well that works with the all-combinations nature of the test.

test/Validate.Layout.cpp
dneto0: I'm fine with what you have.  It's single-threaded test code. (For now.)

dneto0:
Really nice set of tests.  Having all the combinations is really powerful.

* Please replace the magic 7 in the check for Function storage class.
* Tiny cleanups elsewhere as noted.
* A small number of additions to the test case: OpTypeFloat, OpTypeArray certainly.  Certainly a Label, and it would be nice to have load, store and one arithmetic.  The slight concern is that we aren't explicitly testing all the ops.  The risk is low, however.

umar456:
Ok Great. Have added additional functionality to logical layout since I submitted this. Should I wait to push those changes in a separate PR?

test/Validate.Layout.cpp
umar456: I can probably find a way to tests those. I didn't want to add to this because that would create a prohibitively large number of tests.

dneto0:
I prefer to get this one polished and merged before looking at the next one.  Thanks.  I expect we should get this turned around today.

source/validate.cpp
dneto0: Two more thoughts about this:
* I could see how bitwise-or operator| could be read as confusing, since bitwise-or'ing of the spv_result_t integer values doesn't make sense.  But I think that's a better tradeoff than using the operator|| since, in *my* mind anyway, the short-circuiting is a stronger signal than the confusion of bitwise or-ing.   Basically, no C operator fits exactly right.
* Using an operator overload means both the LHS and RHS are now function arguments.  And so the order of evaluation is not specified.  Since there is no certain strict ordering between them, they are concurrently evaluated.  (Depends on who you ask what "concurrent" means....) The issue I'm concerned about is that, depending on the implementation, a module that fails more than one kind of check can will have an indeterminate ordering of messages from the different checks.   Now, the tests are set up to check only one set of rules at a time, so it won't show up in the tests.  It just might confuse users who are on different implementations, trading validation messages back and forth.  Not a big deal, but the absolutely safest thing to do would be to force evaluation of one set of tests at time, by using separate statements.

I still recommend what I said yesterday: please switch to operator|.  Otherwise, coder's choice.

umar456:
Addressed feedback so far.
* Used macros instead of overloading the operator. Seems more readable this way.
* Additional tests should appear in a separate PR

source/validate_types.cpp
```diff
@@ -0,0 +1,206 @@
...
+  unresolved_forward_ids_.erase(id);
+  return SPV_SUCCESS;
+}
+
+void ValidationState_t::assignNameToId(uint32_t id, string name) {
+  operand_names_[id] = name;
+}
+
+string ValidationState_t::getIdName(uint32_t id) const {
+  string out = to_string(id);
```
dneto0: Sorry, std::to_string is not available on Android.  I patched your previous CL before merging into master.  Please pull that fix into this commit.

source/validate_types.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+bool ValidationState_t::isOpcodeInCurrentLayoutStage(SpvOp op) {
+  const vector<SpvOp>& currentStage =
+      GetModuleOrder()[module_layout_order_stage_];
+  return end(currentStage) != find(begin(currentStage), end(currentStage), op);
+}
+
+libspirv::DiagnosticStream ValidationState_t::diag(
+    spv_result_t error_code) const {
+  return libspirv::DiagnosticStream({0, 0, (uint64_t)(instruction_counter_)},
```
dneto0: This fails build for Android, with "narrowing conversion".
I fixed this in you previous CL before merging on master.   (Replaced the c-style cast to uint64_t with static_cast<size_t>)

dneto0:
-1. Sorry, fails build on Android. Please pull in the Android-build fixes I had applied to your previous CL before merging into master.

source/validate_types.cpp
dneto0: Hm.. didn't survive html formatting.
replace c-style cast with
```
static_cast<size_t>(instrucition_counter_)
```

umar456:
Ahh. Sorry about that. I updated master but I created the refactoring commit before you made your change. I have updated and pushed the changes.

dneto0:
There was still a stray "using std::to_string" but no uses.  I just removed it.  I kept the two separate commits, and pushed into master.

dneto0:
Please rebase your future work onto current master, so we don't have this bump in the road again.
Closing this MR.

PR #48, opened 2015-12-18
Add spirv.py from the SPIR-V 1.0.2 registry

dneto0:


dneto0:
FYI.  This is the first step in a sequence of commits to add example code for a SPIR-V coverage tool.
See prototype work at https://github.com/dneto0/SPIRV-Tools/commits/add-spirv-cov-example-tool

antiagainst:
LGTM

PR #49, opened 2015-12-23
Fix parser assert failure for a bad OpSwitch

dneto0:
Emit a diagnostic if the OpSwitch selector refers to an ID that
is valid but has no type.

Discovered by afl-fuzz.

### DIFF
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -555,8 +555,9 @@ spv_result_t Parser::parseOperand(size_t inst_offset,
         // The literal operands have the same type as the value
         // referenced by the selector Id.
         const uint32_t selector_id = peekAt(inst_offset + 1);
-        auto type_id_iter = _.id_to_type_id.find(selector_id);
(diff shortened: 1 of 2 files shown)


dneto0:
Rebased

source/binary.cpp
```diff
@@ -555,8 +555,9 @@ spv_result_t Parser::parseOperand(size_t inst_offset,
         // The literal operands have the same type as the value
         // referenced by the selector Id.
         const uint32_t selector_id = peekAt(inst_offset + 1);
-        auto type_id_iter = _.id_to_type_id.find(selector_id);
-        if (type_id_iter == _.id_to_type_id.end()) {
+        const auto type_id_iter = _.id_to_type_id.find(selector_id);
+        if (type_id_iter == _.id_to_type_id.end() ||
+            type_id_iter->second == 0) {
```
dneto0: See the convention described on line 243.

dekimir:
+2

dneto0:
Rebased

PR #50, opened 2015-12-24
Rename endian.h to spirv_endian.h

dneto0:
If we later add a source/ as an -I include directory,
then avoid confusing other headers that want to include the
standard "endian.h" from /usr/include.


umar456:
LGTM

dekimir:
+2

antiagainst:
It seems `endian.cpp` is not reamed as `spirv_endian.cpp`?

dneto0:
New version also moves source/endian.cpp to source/spirv_endian.cpp.

antiagainst:
LGTM

PR #51, opened 2015-12-26
Additional logical layout checks

umar456:
This PR should provide complete checks for logical layout of a module. 

Added:
* Missing memory model check
* Variable storage class checks
* Function definition and declaration ordering
* Label/branch instruction checks

### DIFF
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -473,9 +473,9 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
 
       if (opcode == SpvOpVariable) {
         const uint32_t* storage_class = inst->words + inst->operands[2].offset;
-        if (*storage_class == 7) {
+        if (*storage_class == SpvStorageClassFunction) {
           return _.diag(SPV_ERROR_INVALID_LAYOUT)
-                 << "Variables can not have a function[7] storage class "
+                 << "Variables cannot have a function[7] storage class "
                     "outside of a function";
         }
       }
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -291,12 +291,6 @@ spv_result_t spvValidateIDs(const spv_instruction_t* pInsts,
 }
 
 namespace {
-// Shame
-#define CHECK_RESULT(EXPRESSION)   \
-  do {                             \
-    spv_result_t ret = EXPRESSION; \
-    if (ret) return ret;           \
-  } while (false);
 
 // TODO(umar): Validate header
 // TODO(umar): The Id bound should be validated also. But you can only do that
@@ -557,7 +551,7 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
               return _.diag(SPV_ERROR_INVALID_LAYOUT)
                      << "Cannot declare a function in a function body";
             }
-            CHECK_RESULT(_.get_functions().RegisterFunction(
+            spvCheckReturn(_.get_functions().RegisterFunction(
                 inst->result_id, inst->type_id,
                 inst->words[inst->operands[2].offset],
                 inst->words[inst->operands[3].offset]));
@@ -568,7 +562,7 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
                                                          "instructions must be "
                                                          "in a function body";
             }
-            CHECK_RESULT(_.get_functions().RegisterFunctionParameter(
+            spvCheckReturn(_.get_functions().RegisterFunctionParameter(
                 inst->result_id, inst->type_id));
             break;
           case SpvOpLine:  // ??
@@ -579,7 +573,7 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
                      << "Label instructions must be in a function body";
             }
             _.progressToNextLayoutStageOrder();
-            CHECK_RESULT(_.get_functions().RegisterSetFunctionDeclType(
+            spvCheckReturn(_.get_functions().RegisterSetFunctionDeclType(
                 kFunctionDeclDefinition));
             break;
           case SpvOpFunctionEnd:
@@ -591,9 +585,9 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
               return _.diag(SPV_ERROR_INVALID_LAYOUT)
                      << "Function end instructions must be in a function body";
             }
-            CHECK_RESULT(_.get_functions().RegisterSetFunctionDeclType(
+            spvCheckReturn(_.get_functions().RegisterSetFunctionDeclType(
                 kFunctionDeclDeclaration));
-            CHECK_RESULT(_.get_functions().RegisterFunctionEnd());
+            spvCheckReturn(_.get_functions().RegisterFunctionEnd());
             break;
           default:
             return _.diag(SPV_ERROR_INVALID_LAYOUT)
@@ -615,15 +609,15 @@ spv_result_t CfgPass(ValidationState_t& _,
     SpvOp opcode = inst->opcode;
     switch (opcode) {
       case SpvOpFunction:
-        CHECK_RESULT(_.get_functions().RegisterFunction(
+        spvCheckReturn(_.get_functions().RegisterFunction(
             inst->result_id, inst->type_id,
             inst->words[inst->operands[2].offset],
             inst->words[inst->operands[3].offset]));
-        CHECK_RESULT(_.get_functions().RegisterSetFunctionDeclType(
+        spvCheckReturn(_.get_functions().RegisterSetFunctionDeclType(
             kFunctionDeclDefinition));
         break;
       case SpvOpFunctionParameter:
-        CHECK_RESULT(_.get_functions().RegisterFunctionParameter(
+        spvCheckReturn(_.get_functions().RegisterFunctionParameter(
             inst->result_id, inst->type_id));
         break;
       case SpvOpFunctionEnd:
@@ -631,10 +625,10 @@ spv_result_t CfgPass(ValidationState_t& _,
           return _.diag(SPV_ERROR_INVALID_LAYOUT) << "Function declarations "
                                                      "must appear before "
                                                      "function definitions.";
-        CHECK_RESULT(_.get_functions().RegisterFunctionEnd());
+        spvCheckReturn(_.get_functions().RegisterFunctionEnd());
         break;
       case SpvOpLabel:
-        CHECK_RESULT(_.get_functions().RegisterBlock(inst->result_id));
+        spvCheckReturn(_.get_functions().RegisterBlock(inst->result_id));
         break;
       case SpvOpBranch:
       case SpvOpBranchConditional:
@@ -643,7 +637,7 @@ spv_result_t CfgPass(ValidationState_t& _,
       case SpvOpReturn:
       case SpvOpReturnValue:
       case SpvOpUnreachable:
-        CHECK_RESULT(_.get_functions().RegisterBlockEnd());
+        spvCheckReturn(_.get_functions().RegisterBlockEnd());
         break;
       default:
         if (_.in_block() == false) {
@@ -666,12 +660,11 @@ spv_result_t ProcessInstructions(void* user_data,
 
   DebugInstructionPass(_, inst);
 
-  // TODO(umar): Perform CFG pass
   // TODO(umar): Perform data rules pass
   // TODO(umar): Perform instruction validation pass
-  CHECK_RESULT(ModuleLayoutPass(_, inst));
-  CHECK_RESULT(CfgPass(_, inst));
-  CHECK_RESULT(SsaPass(_, can_have_forward_declared_ids, inst));
+  spvCheckReturn(ModuleLayoutPass(_, inst));
+  spvCheckReturn(CfgPass(_, inst));
+  spvCheckReturn(SsaPass(_, can_have_forward_declared_ids, inst));
 
   return SPV_SUCCESS;
 }
@@ -699,12 +692,8 @@ spv_result_t spvValidate(const spv_const_context context,
   // NOTE: Parse the module and perform inline validation checks. These
   // checks do not require the the knowledge of the whole module.
   ValidationState_t vstate(pDiagnostic, options);
-  auto err = spvBinaryParse(context, &vstate, binary->code, binary->wordCount,
-                            setHeader, ProcessInstructions, pDiagnostic);
-
-  if (err) {
-    return err;
-  }
+  spvCheckReturn(spvBinaryParse(context, &vstate, binary->code, binary->wordCount,
+                                setHeader, ProcessInstructions, pDiagnostic));
 
   // TODO(umar): Add validation checks which require the parsing of the entire
   // module. Use the information from the processInstructions pass to make
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -61,8 +61,7 @@ using libspirv::ValidationState_t;
 using libspirv::kLayoutFunctionDeclarations;
 using libspirv::kLayoutFunctionDefinitions;
 using libspirv::kLayoutMemoryModel;
-using libspirv::FunctionDecl::kFunctionDeclDeclaration;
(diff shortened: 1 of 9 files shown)


source/validate.cpp
```diff
@@ -286,6 +291,12 @@ spv_result_t spvValidateIDs(const spv_instruction_t* pInsts,
 }
 
 namespace {
+// Shame
```
dekimir: :)

Instead of `CHECK_RESULT(expr)`, why can't you use, eg, `if (auto error = expr) return error;` ??

umar456: Clang complains about using assignment in a condition statement. I could have added `#pragma clang diagnostic ignore -Wparentheses` but that seemed uglier at the moment.

dekimir: That seems dubious.  How does `spvCheckReturn` work, then?  And all the places in the code where we already do exactly what I wrote above.

dekimir: Just in `binary.cpp`, I count six occurrences of `if (auto error = ...) ...`.

umar456: Hmm. it looks like it complains if the error variable is defined outside of the if statement but does not if you define it in the statement.

```
//warning
spv_result_t ret;
if(ret = exp) return ret;

// ok
if(spv_result_t ret = exp) return ret;
```
I will update the macro. Thanks :)

source/validate.cpp
```diff
@@ -287,6 +292,10 @@ spv_result_t spvValidateIDs(const spv_instruction_t* pInsts,
 
 namespace {
 
+// Shame
+#define CHECK_RESULT(EXPRESSION) \
```
dekimir: This is now no different than `spvCheckReturn`.  Also, it contradicts the style established in other code, where we don't use a macro at all, opting to inline the `if` instead.

umar456: Hmm. I don't know how I missed that.

I find that the macro form is more readable. I can expand it to an if statement if that is not a significant enough motivation.

source/validate_types.h
```diff
@@ -41,9 +41,107 @@ namespace libspirv {
 // This enum represents the sections of a SPIRV module. The MODULE section
```
dekimir: This comment is out of date now.

source/validate_types.h
```diff
@@ -41,9 +41,107 @@ namespace libspirv {
...
 // contains instructions who's scope spans the entire module. The FUNCTION
 // section includes SPIRV function and function definitions
-enum class ModuleLayoutSection {
-  kModule,    // < Module scope instructions are executed
-  kFunction,  // < Function scope instructions are executed
+// clang-format off
+enum ModuleLayoutSection {
+  kLayoutCapabilities         = 0,  // < Section 2.4 #1
```
dekimir: The initializers (`= 0`) are unnecessary here; the code effect would be identical without them.

dekimir: Also, if these must line up with the `GetModuleOrder()` result, the blurb should mention that.  Or, ideally, this would become a part of `GetModuleOrder`.

source/validate_types.cpp
```diff
@@ -185,23 +189,127 @@ ModuleLayoutSection ValidationState_t::getLayoutStage() const {
 }
 
 void ValidationState_t::progressToNextLayoutStageOrder() {
-  module_layout_order_stage_ +=
-      module_layout_order_stage_ < GetModuleOrder().size();
-  if (module_layout_order_stage_ >= GetModuleOrder().size()) {
-    current_layout_stage_ = libspirv::ModuleLayoutSection::kFunction;
+  if (current_layout_stage_ <= GetModuleOrder().size()) {
+    current_layout_stage_ = (ModuleLayoutSection)(current_layout_stage_ + 1);
```
dekimir: Use `static_cast`.

https://google.github.io/styleguide/cppguide.html#Casting

source/validate_types.cpp
```diff
@@ -185,23 +189,127 @@ ModuleLayoutSection ValidationState_t::getLayoutStage() const {
...
   }
 }
 
 bool ValidationState_t::isOpcodeInCurrentLayoutStage(SpvOp op) {
-  const vector<SpvOp>& currentStage =
-      GetModuleOrder()[module_layout_order_stage_];
+  const vector<SpvOp>& currentStage = GetModuleOrder()[current_layout_stage_];
   return end(currentStage) != find(begin(currentStage), end(currentStage), op);
```
dekimir: How do `begin` and `end` compile here without the `std::` prefix?

source/validate_types.h
```diff
@@ -110,13 +219,15 @@ class ValidationState_t {
...
-
   // The section of the code being processed
   ModuleLayoutSection current_layout_stage_;
+
+  Functions module_functions_;
+
+  std::vector<SpvCapability> module_capabilities_;
 };
+
+const std::vector<SpvOp>& getModuleOrderInstructions(uint8_t order);
```
dekimir: I can't see where this is used?

source/validate.cpp
```diff
@@ -520,24 +538,117 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
...
+                inst->words[inst->operands[2].offset],
+                inst->words[inst->operands[3].offset]));
+            break;
+          case SpvOpFunctionParameter:
+            if (_.in_function_body() == false) {
+              return _.diag(SPV_ERROR_INVALID_LAYOUT) << "Function parameter "
+                                                         "instructions must be "
+                                                         "in a function body";
+            }
+            spvCheckReturn(_.get_functions().RegisterFunctionParameter(
```
dekimir: Do we already check whether OpFunctionParameter immediately follows an OpFunction?  If not, add a TODO.

source/validate.cpp
```diff
@@ -448,22 +450,38 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
...
+          // ModuleLayoutPass to process the next section of the module
           return ModuleLayoutPass(_, inst);
         }
       }
+
+      if (opcode == SpvOpVariable) {
+        const uint32_t* storage_class = inst->words + inst->operands[2].offset;
+        if (*storage_class == 7) {
+          return _.diag(SPV_ERROR_INVALID_LAYOUT)
+                 << "Variables can not have a function[7] storage class "
```
dekimir: Nit: "cannot" or "can't".

source/validate_types.cpp
umar456: [Koenig lookup](https://en.wikipedia.org/wiki/Argument-dependent_name_lookup)

since the vector is part of the `std` namespace, the appropriate `std::begin` and `std::end` functions are called.

source/validate.cpp
umar456: This is now checked in the RegisterFunctionParameter function. See [implementation](https://github.com/KhronosGroup/SPIRV-Tools/pull/51/files#diff-84a5aa9a556fde9600c6f5583e34ed81R253) and [unit tests](https://github.com/KhronosGroup/SPIRV-Tools/pull/51/files#diff-6dc2086e9da925d662089f2876b41bb7R380).

umar456:
Addressed feedback and rebased

source/validate.cpp
```diff
@@ -520,24 +538,117 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
...
             return _.diag(SPV_ERROR_INVALID_LAYOUT)
-                   << "All OpVariable instructions in a function must have a "
-                      "storage class of Function[7]";
+                   << "All Variable instructions in a function must have a "
+                      "storage class of function[7]";
+        } break;
+        default:
           break;
+      }
+      if (_.getLayoutStage() == kLayoutFunctionDeclarations) {
```
dekimir: How can this ever be false?  Aren't we still in the `else` block started on line 482?

dekimir: Never mind, this is about declarations vs. definitions.  My bad...

source/validate.cpp
```diff
@@ -448,22 +450,38 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
...
+    if (_.getLayoutStage() < kLayoutFunctionDeclarations) {
       // Module scoped instructions are processed by determining if the opcode
       // is part of the current stage. If it is not then the next stage is
       // checked.
       while (_.isOpcodeInCurrentLayoutStage(opcode) == false) {
-        // TODO(umar): Check if the MemoryModel instruction has executed
         _.progressToNextLayoutStageOrder();
-        if (_.getLayoutStage() == libspirv::ModuleLayoutSection::kFunction) {
+
+        if (_.getLayoutStage() == kLayoutMemoryModel &&
```
dekimir: Why single out this stage to check the ordering?  Doesn't the same principle apply to every stage?  Eg, `OpMemoryModel` before `OpCapability` or `OpString` before `OpExecutionMode`.

umar456: Perhaps the error is misleading. This condition takes advantage of the fact that the memory model instruction is required. Therefore if the memory model instruction does not appear at the kLayoutMemoryModel stage then either the instruction does not exist or the layout of instruction is incorrect.

source/validate.cpp
```diff
@@ -520,24 +538,117 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
...
+                   << "All Variable instructions in a function must have a "
+                      "storage class of function[7]";
+        } break;
+        default:
           break;
+      }
+      if (_.getLayoutStage() == kLayoutFunctionDeclarations) {
+        switch (opcode) {
+          case SpvOpFunction:
+            if (_.in_function_body()) {
```
dekimir: Can this ever be true in the function-declarations stage?

dekimir: Ah, it looks like even a declaration has a function body, where parameters go.  Never mind...

source/validate.cpp
```diff
@@ -520,24 +538,117 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _,
...
+              return _.diag(SPV_ERROR_INVALID_LAYOUT) << "Function parameter "
+                                                         "instructions must be "
+                                                         "in a function body";
+            }
+            spvCheckReturn(_.get_functions().RegisterFunctionParameter(
+                inst->result_id, inst->type_id));
+            break;
+          case SpvOpLine:  // ??
+            break;
+          case SpvOpLabel:
```
dekimir: Isn't a label forbidden in function declarations?

umar456: This case is taken when you reach the first definition. We update the layout stage to kLayoutFunctionDefinition and treat the current function as a definition.

dekimir: Then I don't understand the comment on line 598, nor how we ensure that no declaration is accidentally intermingled with definitions.

umar456: The `CfgPass` function is defined in line 606. This function is only executed when the `getLayoutStage` returns `kFunctionDefinitions`. This stage switch case is skipped in that case (see line 547).

I decided to put those checks in the CfgPass because only function definitions can have blocks. It seemed to be the natural place for something like this. Although now that I think about it I can probably just as easily have that logic here. I can put the layout logic here if you feel this is a better location for something like that.

Anyway line 624 ensures you cannot have function declarations intermingled with function definitions by making sure that the function has at least one block.

source/validate_types.cpp
```diff
@@ -186,23 +186,133 @@ ModuleLayoutSection ValidationState_t::getLayoutStage() const {
...
+  (void)ret_type_id;
+  (void)function_control;
+
+  return SPV_SUCCESS;
+}
+
+spv_result_t Functions::RegisterFunctionParameter(uint32_t id,
+                                                  uint32_t type_id) {
+  assert(in_function_ == true &&
+         "Function parameter instructions can not be declared of a function");
```
dekimir: "cannot"
"_outside_ of"

source/validate_types.cpp
```diff
@@ -186,23 +186,133 @@ ModuleLayoutSection ValidationState_t::getLayoutStage() const {
...
+  (void)id;
+  (void)type_id;
```
dekimir: What are the above two lines for?

source/validate_types.cpp
```diff
@@ -186,23 +186,133 @@ ModuleLayoutSection ValidationState_t::getLayoutStage() const {
...
+    return module_.diag(SPV_ERROR_INVALID_LAYOUT)
+           << "Function parameters must only appear immediatly after the "
+              "function definition";
+  }
+  // TODO(umar): Validate function parameter type order and count
+  return SPV_SUCCESS;
+}
+
+spv_result_t Functions::RegisterSetFunctionDeclType(FunctionDecl type) {
+  assert(in_function_ == true && "Function can not be declared of a function");
```
dekimir: Confusing.  Do you mean "inside a function"?

source/validate_types.h
```diff
@@ -38,12 +38,108 @@
...
+
+enum class FunctionDecl {
+  kFunctionDeclUnknown,      // < Unknown function declaration
+  kFunctionDeclDeclaration,  // < Function declaration
+  kFunctionDeclDefinition    // < Function definition
+};
+
+class ValidationState_t;
+
+class Functions {
```
dekimir: Blurb needed.

test/Validate.Layout.cpp
```diff
@@ -46,30 +47,98 @@ using std::vector;
...
 using ::testing::HasSubstr;
 
-using pred_type = function<bool(int)>;
+using pred_type = function<spv_result_t(int)>;
 using ValidateLayout =
     spvtest::ValidateBase<tuple<int, tuple<string, pred_type, pred_type>>,
                           SPV_VALIDATE_LAYOUT_BIT>;
 
+// TODO(umar): probably should move this to a central location
+ostream& operator<<(ostream& os, const spv_result_t& err) {
```
dekimir: Why not something like `to_string(spv_result_t)`?  Then the result can be streamed natively, in addition to using it for other purposes.

source/validate_types.cpp
umar456: I am eventually going to check the parameter type and use the id in some way. This is done to avoid unused parameter warnings in clang.

dekimir: Then please add a comment explaining that.

test/Validate.Layout.cpp
umar456: I thought we couldn't use `to_string` because the Android compiler does not support it? Are you suggesting I should create a function like `to_string` for `spv_result_t` instead of a stream operator? perhaps a `spvResultToString`.

dekimir: Yes, I meant creating our own function.

umar456:
Addressed feedback

* Created libspirv::spvResultToString in diagnostic.h
* Documented various functions and classes
* Fixed error messages

source/validate.cpp
```diff
@@ -58,6 +58,11 @@ using std::unordered_set;
 using std::vector;
 
 using libspirv::ValidationState_t;
+using libspirv::kLayoutFunctionDeclarations;
+using libspirv::kLayoutFunctionDefinitions;
+using libspirv::kLayoutMemoryModel;
+using libspirv::FunctionDecl::kFunctionDeclDeclaration;
```
dekimir: This doesn't compile for me, as FunctionDecl is not a namespace.

umar456: Fixed

dekimir:
Squashed and merged as 42a2879.

umar456:
@dekimir The author  and the commiter of the merge is not listed as me. Is this a new policy or was this a mistake. I know when @dneto0 performed the merge this wasn't the case.

dekimir:
@umar456 this was a mistake -- I've squashed it in a different way than @dneto0 does.  My apologies for that.  The policy hasn't changed.

dneto0:
I plan to redo the commit with a force push.  Should take about 10 min.

umar456:
Awesome! Thanks

dneto0:
Ok. I squashed all of @umar456's commits into one with a hopefully sensible combined commit message. Then @dekimir's patch after that. 
Compiles and tests fine on our systems.
Required a force push onto master.  So beware if you've rebased in the last 10 minutes.

PR #55, opened 2016-01-05
Fix strict clang warnings

dneto0:
Avoid:
- variable or argument shadowing
- implicit narrowing conversions
- implicit conversions of character array to bool (that the compiler complains about)


AWoloszyn:
+2

dneto0:
Rebased

PR #56, opened 2016-01-05
Fix list of opcodes allowed for OpSpecConstantOp

dneto0:
Was missing BitcastXor, GenericCastToPtr.
Remove duplicate of SLessThanEqual


AWoloszyn:
+2

PR #57, opened 2016-01-06
Fix strict GCC warnings

dneto0:
Fixes GCC warnings for:
- variables shadowing methods in the same class
- implicit conversions that the compiler thinks could change values


antiagainst:
LGTM

PR #58, opened 2016-01-07
Add configuration for Travis CI.

antiagainst:


antiagainst:
This configuration works: https://travis-ci.org/antiagainst/SPIRV-Tools

.travis.yml
```diff
@@ -0,0 +1,68 @@
...
+  - if [ "$CC" = "clang" ]; then export CC=clang-3.7 CXX=clang++-3.7; fi # Make sure that clang-3.7 is selected.
+
+before_script:
+  - git clone https://github.com/google/googletest.git external/googletest
+
+script:
+  - mkdir build && cd build
+  - cmake -GNinja -DCMAKE_BUILD_TYPE=${BUILD_TYPE} ..
+  - ninja
+  - ./UnitSPIRV --gtest_break_on_failure
```
dneto0: Why should we use a non-default failure mode?

dneto0: It's worth a comment at least.

antiagainst: For me, this makes the failure much easier to locate (always at the end of the log). This helps because we have thousands of tests, the output volume is huge and I don't know how to suppress those successful ones.

antiagainst: But I don't care that much. Will change to `./UnitSPIRV` only. But I'll defer that until Travis CI is turned on for this project to verify that this configuration works.

dneto0:
+2

antiagainst:
@dneto0 Is Travis CI enabled for SPIRV-Tools now?

dneto0:
Yes.  I will rebase and merge.

antiagainst:
In as 866b6ab9daa9933bcb9aa95f99f595c6bff6d9b6.

PR #61, opened 2016-01-08
Hooked up 16-bit floating point numbers.

AWoloszyn:
This adds half-precision constants to spirv-tools.
16-bit floats are always disassembled into hex-float format,
but can be assembled from floating point or hex-float inputs.

### DIFF
--- a/include/util/hex_float.h
+++ b/include/util/hex_float.h
@@ -697,6 +697,10 @@ std::ostream& operator<<(std::ostream& os, const HexFloat<T, Traits>& value) {
   return os;
 }
 
+// Parses a floating point number from the given stream and stores it into the
+// value parameter.
+// If the negate_value parameter is true then the number is negated before
+// it is stored into the value parameter.
 template <typename T, typename Traits>
 inline std::istream& ParseNormalFloat(std::istream& is, bool negate_value,
                                       HexFloat<T, Traits>& value) {
@@ -706,6 +710,26 @@ inline std::istream& ParseNormalFloat(std::istream& is, bool negate_value,
     val = -val;
   }
   value.set_value(val);
+  return is;
+}
+
+// Specialization of ParseNormalFloat for FloatProxy<Float16> values.
+// This will parse the float as it were a 32-bit floating point number,
+// and then round it down to fit into a Float16 value. 
+// The number is rounded towards zero.
+// Any floating point number that is too large will be rounded to +- infinity.
+template <>
+inline std::istream&
+ParseNormalFloat<FloatProxy<Float16>, HexFloatTraits<FloatProxy<Float16>>>(
+    std::istream& is, bool negate_value,
+    HexFloat<FloatProxy<Float16>, HexFloatTraits<FloatProxy<Float16>>>& value) {
+  float f;
+  is >> f;
+  if (negate_value) {
+    f = -f;
+  }
+  HexFloat<FloatProxy<float>> float_val(f);
+  float_val.castTo(value, round_direction::kToZero);
   return is;
 }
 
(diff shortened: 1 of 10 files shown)


include/util/hex_float.h
```diff
@@ -709,6 +709,21 @@ inline std::istream& ParseNormalFloat(std::istream& is, bool negate_value,
   return is;
 }
 
+template <>
```
dneto0: Needs a blurb.  Actually, the previously existing function at line 700 is missing it, which could probably serve to document this one, since this one is just the specialization for the float16-based proxy.

include/util/hex_float.h
```diff
@@ -709,6 +709,21 @@ inline std::istream& ParseNormalFloat(std::istream& is, bool negate_value,
...
+ParseNormalFloat<FloatProxy<Float16>, HexFloatTraits<FloatProxy<Float16>>>(
+    std::istream& is, bool negate_value,
+    HexFloat<FloatProxy<Float16>, HexFloatTraits<FloatProxy<Float16>>>& value) {
+  float f;
+  is >> f;
+  if (negate_value) {
+    f = -f;
+  }
+  HexFloat<FloatProxy<float>> float_val(f);
+  float_val.castTo(value, round_direction::kToZero);
```
dneto0: This rounding behaviour should be described in the blurb.

source/disassemble.cpp
```diff
@@ -233,9 +233,12 @@ void Disassembler::EmitOperand(const spv_parsed_instruction_t& inst,
             stream_ << word;
             break;
           case SPV_NUMBER_FLOATING:
-            // Assume only 32-bit floats.
-            // TODO(dneto): Handle 16-bit floats also.
-            stream_ << spvutils::FloatProxy<float>(word);
+            if (operand.number_bit_width == 16) {
+              stream_ << spvutils::FloatProxy<spvutils::Float16>(uint16_t(word & 0xFFFF));
```
dneto0: Masking out the top bits is redundant, but I'm ok keeping it.

source/text_handler.cpp
```diff
@@ -387,9 +387,14 @@ spv_result_t AssemblyContext::binaryEncodeFloatingPointLiteral(
...
-    case 16:
-      return diagnostic(SPV_ERROR_INTERNAL)
-             << "Unsupported yet: 16-bit float constants.";
+    case 16: {
+      spvutils::HexFloat<FloatProxy<spvutils::Float16>> hVal(0);
+      if (auto error = parseNumber(val, error_code, &hVal,
+                                   "Invalid 16-bit float literal: "))
+        return error;
+      return binaryEncodeU32(
+          static_cast<uint32_t>(hVal.value().getAsFloat().get_value()), pInst);
```
dneto0: Might be worth mentioning that getAsFloat will return the spvutils::Float16, and in turn the get_value() will return a uint16_t, so the encoding is correct as per SPIR-V spec, i.e. the bits are in the least significant 16 bits, and the most significant 16 bits are all 0.

test/UnitSPIRV.cpp
```diff
@@ -26,6 +26,7 @@
 
 #include "UnitSPIRV.h"
 
+#include "TestFixture.h"
```
dneto0: should be the last include.

dneto0:
-1.  See inline comments.

First line of the commit comment could be clearer: "Support 16-bit float in assembler and disassembler"

Also, update project README to remove this item from Future Work for assembler and disassembler.

AWoloszyn:
Updated. Responded to all comments.

dneto0:
+2

dneto0:
This was merged into master with 43401d2ed0e4f815fcb19b8ec224c73749f44bf1

PR #63, opened 2016-01-11
Made compilation with gcc and clang more strict.

AWoloszyn:
Added additional compilation flags to gcc and clang builds.
Adds -Wall -Wextra -Wno-long-long -Wshadow -Wundef -Wconversion
-WNo-sign-conversion and -Wno-missing-field-initializers
where appropriate.

Does not add -Wundef to tests, because GTEST tests undefined
macros all over the place.


dneto0:
+2
Thanks!

Issue #62, opened 2016-01-11
spvDiagnosticDestroy should be safe to call on nullptr


dneto0:
That's a useful behaviour, to simplify client code.
The current implementation is already safe in this way.
Need to document it and write tests for it.

PR #64, opened 2016-01-11
spvDiagnosticDestroy is safe to call on nullptr

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/62

### DIFF
--- a/test/DiagnosticDestroy.cpp
+++ b/test/DiagnosticDestroy.cpp
@@ -1,0 +1,57 @@
+// Copyright (c) 2015-2016 The Khronos Group Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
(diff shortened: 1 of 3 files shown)


test/DiagnosticDestroy.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+TEST(DiagnosticDestroy, DestroyNull) { spvDiagnosticDestroy(nullptr); }
+
+TEST(DiagnosticDestroy, DestroyValidDiagnostic) {
+  spv_position_t position = {};
+  spv_diagnostic diagnostic = spvDiagnosticCreate(&position, "");
+  EXPECT_NE(nullptr, diagnostic);
+  spvDiagnosticDestroy(diagnostic);
+  // We aren't allowed to use the diagnostic pointer anymore.
+  // So we can't test its behaviour.
```
AWoloszyn: Optional, test re-assigning to an existing spv_diagnostic. Since this is a C api, and spv_diagnostic is a pointer probably fine to just ignore this case.

dneto0: Added a test case to that effect.

test/DiagnosticDestroy.cpp
```diff
@@ -0,0 +1,59 @@
...
+// Returns a newly created diagnostic value.
+spv_diagnostic MakeValidDiagnostic() {
+  spv_position_t position = {};
+  spv_diagnostic diagnostic = spvDiagnosticCreate(&position, "");
+  return diagnostic;
+}
+
+TEST(DiagnosticDestroy, DestroyValidDiagnostic) {
+  spv_diagnostic diagnostic = MakeValidDiagnostic();
+  EXPECT_NE(nullptr, diagnostic);
```
dneto0: This is silly. 
I can assert that check in the utility function.

AWoloszyn:
+2

PR #65, opened 2016-01-11
Fix endian-ness conversion in the binary parser.

AWoloszyn:
When we were swapping endian-ness in the binary parser, we would
write outside of the bounds of our vector.


dneto0:
+2
Wow, yay fuzzing.

PR #67, opened 2016-01-12
Fixed use-after-free in isStartOfNewInst() tests.

AWoloszyn:


dneto0:
+2
My bad, and thanks for fixing!

PR #68, opened 2016-01-13
Fix build by renaming ValidationState_t constructor argument.

dekimir:
The `diag` argument shadows the `diag` method, so the strict build fails.


dneto0:
We have shadowing warnings turned on, which turn into errors.
+2

PR #70, opened 2016-01-13
Remove static vector from GetModuleOrder

### LINKED ISSUE #53: GetModuleOrder() result has unspecified order of destruction
Currently, `GetModuleOrder()` returns a reference to a static `vector` object.  This is a violation of the style guide because other static objects may accidentally dereference this in their destructors, without any guarantee that the `GetModuleOrder` vector isn't already destroyed.  Instead of creating a static `vector`, the method should either use a C array or a static pointer to a dynamically allocated `vector`.

References:
https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2

umar456:
Fixes #53 

Not as clean but follows the Google style guide.

### DIFF
--- a/source/validate_types.cpp
+++ b/source/validate_types.cpp
@@ -38,85 +38,162 @@ using std::find;
 using std::string;
 using std::unordered_set;
 using std::vector;
+using namespace libspirv;
 
 namespace {
-const vector<vector<SpvOp>>& GetModuleOrder() {
+bool IsInstructionInLayoutSection(ModuleLayoutSection layout, SpvOp op) {
   // See Section 2.4
+  bool out = false;
   // clang-format off
-  static const vector<vector<SpvOp>> moduleOrder = {
-    {SpvOpCapability},
-    {SpvOpExtension},
-    {SpvOpExtInstImport},
-    {SpvOpMemoryModel},
-    {SpvOpEntryPoint},
-    {SpvOpExecutionMode},
-    {
-      // first set of debug instructions
-      SpvOpSourceContinued,
-      SpvOpSource,
-      SpvOpSourceExtension,
-      SpvOpString,
-    },
-    {
-      // second set of debug instructions
-      SpvOpName,
-      SpvOpMemberName
-    },
-    {
-      // annotation instructions
-      SpvOpDecorate,
-      SpvOpMemberDecorate,
-      SpvOpGroupDecorate,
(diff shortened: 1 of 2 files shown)


source/validate_types.cpp
```diff
@@ -38,85 +38,162 @@ using std::find;
...
-      SpvOpSpecConstantFalse,
-      SpvOpSpecConstant,
-      SpvOpSpecConstantComposite,
-      SpvOpSpecConstantOp,
-      SpvOpVariable,
-      SpvOpLine
-    }
-  };
+  switch (layout) {
+    case kLayoutCapabilities:  out = op == SpvOpCapability;    break;
```
dneto0: I tend to just do an early return in situations like this.  (Also, that's not forbidden by Google style guide, but it is by some.   You would also tend to add a dummy "return <something>" at the end of the function to avoid warnings in some compilers.)

Coder's choice.  (i.e. ok to keep this)

source/validate_types.cpp
```diff
@@ -186,15 +263,15 @@ ModuleLayoutSection ValidationState_t::getLayoutStage() const {
 }
 
 void ValidationState_t::progressToNextLayoutStageOrder() {
-  if (static_cast<size_t>(current_layout_stage_) <= GetModuleOrder().size()) {
+  // Guard against going past the last element(kLayoutFunctionDefinitions)
+  if (current_layout_stage_ <= kLayoutFunctionDefinitions) {
```
dneto0: Optionally, you could define  
  kLayoutLastSection = kLayoutFunctionDefinitions
and then use kLayoutLastSection here.  It would be slightly more readable.

Coder's choice.

dneto0: Also, in past we've had problems with different compilers complaining about the signed/unsigned comparison.  This version looks pristinely good to me.

dneto0:
For the record, I prefer the new form of this, i.e. switch statements over the use of containers.  Two reasons:
 - the STL containers version(s) of this induces concern for leaks, races, destructor order, etc. as evidenced on an earlier MR and in https://github.com/KhronosGroup/SPIRV-Tools/issues/53      I much prefer to KISS and have everyone move on to implementing new functionality.
 - the containers form creates extra failure modes, e.g. if allocation fails.  The switch form will fail only at compile or link time.

+2
I'll merge.

PR #72, opened 2016-01-14
Validation refactor

umar456:
* Organize passes into seperate files
* Remove module layout logic from Cfg pass
* Remove module layout logic from Functions class
* Refactor ModuleLayoutPass for readability
* Adapt consistent naming of layout sections (Stage/Section -> Section)

### DIFF
--- a/include/libspirv/libspirv.h
+++ b/include/libspirv/libspirv.h
@@ -260,9 +260,10 @@ typedef enum spv_validate_options_t {
   SPV_VALIDATE_ID_BIT = SPV_BIT(2),
   SPV_VALIDATE_RULES_BIT = SPV_BIT(3),
   SPV_VALIDATE_SSA_BIT = SPV_BIT(4),
+  SPV_VALIDATE_INSTRUCTION_BIT = SPV_BIT(5),
   SPV_VALIDATE_ALL = SPV_VALIDATE_BASIC_BIT | SPV_VALIDATE_LAYOUT_BIT |
                      SPV_VALIDATE_ID_BIT | SPV_VALIDATE_RULES_BIT |
-                     SPV_VALIDATE_SSA_BIT,
+                     SPV_VALIDATE_SSA_BIT | SPV_VALIDATE_INSTRUCTION_BIT ,
   SPV_FORCE_32_BIT_ENUM(spv_validation_options_t)
 } spv_validate_options_t;
 
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -26,6 +26,7 @@
 
 #include "validate.h"
 #include "validate_types.h"
+#include "validate_passes.h"
 
 #include "binary.h"
 #include "diagnostic.h"
@@ -41,30 +42,23 @@
 #include <cstdio>
 #include <functional>
 #include <iterator>
-#include <map>
 #include <sstream>
 #include <string>
-#include <unordered_set>
 #include <vector>
 
 using std::function;
-using std::map;
 using std::ostream_iterator;
 using std::placeholders::_1;
 using std::string;
 using std::stringstream;
 using std::transform;
-using std::unordered_set;
 using std::vector;
 
+using libspirv::CfgPass;
+using libspirv::InstructionPass;
+using libspirv::ModuleLayoutPass;
+using libspirv::SsaPass;
 using libspirv::ValidationState_t;
-using libspirv::kLayoutFunctionDeclarations;
-using libspirv::kLayoutFunctionDefinitions;
-using libspirv::kLayoutMemoryModel;
-using libspirv::FunctionDecl;
-
-#define spvCheckReturn(expression) \
-  if (spv_result_t error = (expression)) return error;
 
 #if 0
 spv_result_t spvValidateOperandsString(const uint32_t* words,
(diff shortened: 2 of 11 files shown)


include/libspirv/libspirv.h
```diff
@@ -260,6 +260,7 @@ typedef enum spv_validate_options_t {
   SPV_VALIDATE_ID_BIT = SPV_BIT(2),
   SPV_VALIDATE_RULES_BIT = SPV_BIT(3),
   SPV_VALIDATE_SSA_BIT = SPV_BIT(4),
+  SPV_VALIDATE_INSTRUCTION_BIT = SPV_BIT(5) | SPV_VALIDATE_LAYOUT_BIT,
```
dekimir: This is not a bit but two bits, it seems.  So the name is misleading.

Also, should this be in `SPV_VALIDATE_ALL`?

dneto0: Before getting to read the code, I assumed at this point that the layout checks passing allow the instruction tests to make simplifying assumptions.   However, if that's the case, then a comment to that effect should be added here.
(And now off to read the further down.)

source/validate.cpp
```diff
@@ -41,27 +42,24 @@
...
 
-using libspirv::ValidationState_t;
-using libspirv::kLayoutFunctionDeclarations;
-using libspirv::kLayoutFunctionDefinitions;
-using libspirv::kLayoutMemoryModel;
-using libspirv::FunctionDecl;
+using libspirv::ModuleLayoutPass;
+using libspirv::CfgPass;
+using libspirv::SsaPass;
+using libspirv::InstructionPass;
```
dneto0: These are out of order.  :-)

source/validate.cpp
```diff
@@ -41,27 +42,24 @@
...
+
+using namespace libspirv;
```
dneto0: Avoid using an entire namespace.

source/validate_instruction.cpp
```diff
@@ -0,0 +1,62 @@
...
+namespace libspirv {
+
+spv_result_t InstructionPass(ValidationState_t& _,
+                             const spv_parsed_instruction_t* inst) {
+  if (_.is_enabled(SPV_VALIDATE_INSTRUCTION_BIT)) {
+    SpvOp opcode = inst->opcode;
+    switch (opcode) {
+      case SpvOpVariable: {
+        const uint32_t storage_class = inst->words[inst->operands[2].offset];
+        if (_.getLayoutSection() > kLayoutFunctionDeclarations) {
```
dekimir: Will this also catch `OpVariable` between `OpEndFunction` and the next `OpFunction`?

source/validate_cfg.cpp
```diff
@@ -0,0 +1,61 @@
...
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#include "validate_passes.h"
+#include "validate_types.h"
+
+#define spvCheckReturn(expression)                      \
```
dekimir: DRY

source/validate_layout.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include "validate_passes.h"
+#include "libspirv/libspirv.h"
+
+#include "diagnostic.h"
+#include "opcode.h"
+#include "operand.h"
+
+#include <cassert>
+
+#define spvCheckReturn(expression)                      \
```
dekimir: DRY

source/validate_types.cpp
```diff
@@ -322,15 +322,8 @@ spv_result_t Functions::RegisterFunctionParameter(uint32_t id,
...
-  if (in_block()) {
-    return module_.diag(SPV_ERROR_INVALID_LAYOUT)
-           << "Function parameters cannot be called in blocks";
-  }
-  if (block_ids_.back().size() != 0) {
-    return module_.diag(SPV_ERROR_INVALID_LAYOUT)
-           << "Function parameters must only appear immediatly after the "
-              "function definition";
-  }
+  assert(in_block() == false &&
```
dneto0: There are asserts here in place of diagnostics.  ?!

source/validate_instruction.cpp
```diff
@@ -0,0 +1,62 @@
...
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+// Performs validation on instructions that appear inside of a SPIR-V block.
+
+#include "validate_passes.h"
+#include "validate_types.h"
+
+namespace libspirv {
+
+spv_result_t InstructionPass(ValidationState_t& _,
```
dneto0: This new functionality should not be in the same merge request as a pure refactor.

dneto0:
-1 I'm ok with rearranging code, but do it in different merge requests from functionality changes.
I see diagnostics replaced with asserts, and the addition of an Instruction pass.


Also, this needs to rebase against master.

source/validate_instruction.cpp
umar456: Those instructions are handled by the layout pass. Perhaps it would be a good idea to decouple the layout and the instruction and type checks. I might do that in a future PR.

source/validate_types.cpp
umar456: Those checks are performed in the validate_layout.cpp. These are sanity checks.

source/validate_instruction.cpp
umar456: This is the same functionality as the checks in line [473](https://github.com/KhronosGroup/SPIRV-Tools/pull/72/files#diff-b8a71d60452cadda1e5f190db6789d2dL473) and [483](https://github.com/KhronosGroup/SPIRV-Tools/pull/72/files#diff-b8a71d60452cadda1e5f190db6789d2dL483). I just did this separate these checks from the layout checks.

include/libspirv/libspirv.h
umar456: @dneto0 is correct, but I think it would be a good idea to isolate the instruction checks from the layout checks for unit testing purposes.

umar456:
Addressed feedback

include/libspirv/libspirv.h
```diff
@@ -260,6 +260,7 @@ typedef enum spv_validate_options_t {
   SPV_VALIDATE_ID_BIT = SPV_BIT(2),
   SPV_VALIDATE_RULES_BIT = SPV_BIT(3),
   SPV_VALIDATE_SSA_BIT = SPV_BIT(4),
+  SPV_VALIDATE_INSTRUCTION_BIT = SPV_BIT(5),
```
dekimir: I'm still unclear why this isn't in `_ALL`.

source/validate_cfg.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "validate_passes.h"
+#include "validate_types.h"
+
+namespace libspirv {
+
+// TODO(umar): Support for merge instructions
+// TODO(umar): Structured control flow checks
+spv_result_t CfgPass(ValidationState_t& _,
+                     const spv_parsed_instruction_t* inst) {
+  if (_.getLayoutSection() == kLayoutFunctionDefinitions) {
```
dekimir: This is unconditional?  I.e., no `SPV_VALIDATE` bit guards it?

include/libspirv/libspirv.h
umar456: Ahh my mistake. I misunderstood your previous comment.

source/validate_cfg.cpp
umar456: I am not sure how to handle this one because SPV_VALIDATE_CFG_BIT would require SPV_VALIDATE_LAYOUT_BIT to perform correctly (i.e. only check the instructions inside of functions etc.). Unless we repeat logic in each of the passes this would yield incorrect results. Unless these bit flags implicitly indicate that the incoming instructions can only contain instructions that are relevant to this pass.

umar456: I could run the layout checks if the CFG bit is passed into the validator.

dekimir: We could enforce the dependency here: if CFG_BIT is disabled, skip this pass.  Otherwise, check LAYOUT_BIT and proceed if it's enabled, skip with a warning if it isn't.

source/validate_instruction.cpp
```diff
@@ -0,0 +1,62 @@
...
+                             const spv_parsed_instruction_t* inst) {
+  if (_.is_enabled(SPV_VALIDATE_INSTRUCTION_BIT)) {
+    SpvOp opcode = inst->opcode;
+    switch (opcode) {
+      case SpvOpVariable: {
+        const uint32_t storage_class = inst->words[inst->operands[2].offset];
```
dekimir: Might be safer to rely on the operand_type == SPV_OPERAND_TYPE_STORAGE_CLASS than to hard-code 2 here.

source/validate_layout.cpp
```diff
@@ -0,0 +1,205 @@
...
+              FunctionDecl::kFunctionDeclDefinition));
+        break;
+
+      case SpvOpFunctionParameter:
+        if (_.in_function_body() == false) {
+          return _.diag(SPV_ERROR_INVALID_LAYOUT) << "Function parameter "
+                                                     "instructions must be in "
+                                                     "a function body";
+        }
+        if (_.get_functions().get_block_count() != 0) {
```
dekimir: This just checks that the instruction is in the first block, not that it's immediately after OpFunction, right?

source/validate_cfg.cpp
dekimir: After thinking about this some more, I'm beginning to agree with @umar456 that we should transitively turn on dependent passes, to maximize user convenience.  It may be tricky, though, propagating all the dependencies.

source/validate_layout.cpp
umar456: When a label is encountered a new block is added to the function increasing the count to 1. The only instruction which can appear between the function instruction and the first label is the function parameter and the line instruction. I don't know if the line instruction is valid between the function and function parameter instructions.

source/validate_layout.cpp
```diff
@@ -0,0 +1,205 @@
...
+                                                     "function definitions.";
+        }
+        spvCheckReturn(_.get_functions().RegisterFunctionEnd());
+        if (_.getLayoutSection() == kLayoutFunctionDeclarations) {
+          spvCheckReturn(_.get_functions().RegisterSetFunctionDeclType(
+              FunctionDecl::kFunctionDeclDeclaration));
+        }
+        break;
+
+      case SpvOpLine:  // ??
```
dekimir: My understanding is that OpLine may safely be ignored here, but in `ModuleScopedInstructions()` we should check that it doesn't appear until after annotations.  Or is that magically handled by `while isOpcodeInCurrentLayoutSection`?

And we shouldn't forget about OpNoLine, which is much the same.

umar456: Yes. `OpLine` is handled in `isOpcodeInCurrentLayoutSection` (see [122](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/validate_types.cpp#L122)) but it seems I have missed `OpNoLine`.

umar456: I can fix OpNoLine in a future PR

source/validate_instruction.cpp
umar456: I don't know how I could use that here. Go through all of the operands in a for loop or perform an assert perhaps?

source/validate_cfg.cpp
umar456: There are two use cases (which I see) for these flags.
1. Perform isolated unit tests on small snippets. For example:
```
OpLabel
OpNop
OpBranch %next
```
2. Only display and exit for errors of one type for a complete module

I don't think implicitly adding additional checks will solve either of these problems. I think it would be better(easier?) to target the first use case. The second use case will require additional machinery to work correctly

source/validate_instruction.cpp
dekimir: Not an assert, a comparison.  The one that equals STORAGE_CLASS is the one you want.  Today, that'll be 2, but tomorrow it may be something else and this code will still work.

source/validate_cfg.cpp
dekimir: I'd be OK with scrapping these options altogether, making usage simpler.  Use case 2 seems unlikely, and use case 1 can be accomplished by tests calling passes directly.  It's hard to argue that something is a unit test if it invokes the entire machine.  Also, expanding an API just for testing purposes is often a hint that design is suboptimal.

source/validate_types.cpp
dneto0: I see that now. Thanks.

source/validate_instruction.cpp
dneto0: SPIR-V spec specifies that it is 2.   We've done that in all manner of places. It's not the kind of change that is likely to occur.

dneto0:
I'm satisfied this is a refactor of the existing functionality.
Seems we have identified an issue with OpNoLine that can be fixed in a future PR.

+1

source/validate_cfg.cpp
umar456: That is probably a better approach. I don't think I want to do something like that in this PR.

umar456:
Squashed and rebased.

source/validate_layout.cpp
umar456: @dneto0 Is it possible for the OpLine instruction to appear between the OpFunction and OpFunctionParameter instructions?

dneto0: @umar456:  Yes OpLine can appear there.  The OpLine instruction description is the best place to look for that. 

Allowing it to appear between OpFunction and OpFunctionParameter is necessary to describe source line info for the function parameters themselves.
E.g.
```
  void MyFunc(
      int has_one_parmeter_on_this_line,
     char and_another_pamaeter_here
 ) { ...
```

PR #73, opened 2016-01-15
Disable warnings C4503 on MSVC for test code

dneto0:
C4503 is "decorated name length exceeded", triggered
by some heavily templated types.


dneto0:
I'm looking for confirmation that this fixes compilation problems on Windows.

dekimir:
+2

PR #74, opened 2016-01-15
Track uses and defs during parsing.

dekimir:
Replace two other, imperfect mechanisms for use-def tracking.

Use ValidationState_t::entry_points to track entry points.

Concentrate undefined-ID diagnostics in a single place.

Move validate_types.h content into validate.h due to increased
inter-dependency.


dekimir:
@umar456 PTAL

source/validate.h
```diff
@@ -36,33 +46,267 @@
...
+
+  // Retuns true if the called after a function instruction but before the
+  // function end instruction
+  bool in_function_body() const;
+
+  // Returns true if called after a label instruction but before a branch
+  // instruction
+  bool in_block() const;
+
+  // Keeps track of ID definitions and uses.
```
dneto0: For myself: This is new code.

source/validate.h
```diff
@@ -36,33 +46,267 @@
 typedef struct spv_id_info_t {
   uint32_t id;
   SpvOp opcode;
-  const spv_instruction_t* inst;
-  spv_position_t position;
+  std::vector<uint32_t> words;
```
dneto0: What does the field mean?
Actually, I'd like a blurb for each member. 
What is opcode?

source/validate.cpp
```diff
@@ -332,14 +270,34 @@ void DebugInstructionPass(ValidationState_t& _,
   }
 }
 
+void ProcessIds(ValidationState_t& _, const spv_parsed_instruction_t& inst) {
```
dneto0: Needs a blurb.

source/validate.cpp
```diff
@@ -332,14 +270,34 @@ void DebugInstructionPass(ValidationState_t& _,
...
+  if (inst.result_id) {
+    _.usedefs().AddDef(
+        {inst.result_id, inst.opcode,
+         std::vector<uint32_t>(inst.words, inst.words + inst.num_words)});
+  }
+  for (auto op = inst.operands; op != inst.operands + inst.num_operands; ++op) {
+    if (op->type == SPV_OPERAND_TYPE_ID)
```
dneto0: There are also SPV_OPERAND_TYPE_TYPE_ID, .... *_MEMORY_SEMANTICS_ID, *_SCOPE_ID.
Maybe we need a utility function bool spvOperandTypeIsID(spv_operand_type_t)

source/validate.cpp
```diff
@@ -332,14 +270,34 @@ void DebugInstructionPass(ValidationState_t& _,
...
+      _.usedefs().AddUse(inst.words[op->offset]);
+  }
+  if (inst.opcode == SpvOpEntryPoint) {
```
dneto0: This is not populating entry points correctly.   
The comment at the entry_points_ member says it's the list of IDs that are entry points.

So the right code here would be: 
```
if inst.opcode == SpvOpEntryPoint) { _.entry_points().push_back(inst.words[2]) }
```

The current code in the MR will add any IDs in the *interface* as entry points.

This is a case where the attempt to be general leads to confusing and incorrect code.  Just use the 2 as the index of the entry point ID.

source/validate.cpp
dneto0: Also, I'd argue this capturing of entry points is logically separate from collecting uses and defs.
It's arguable whether it even merits its own function, since the right solution is a single line of code.

source/validate.h
```diff
@@ -36,33 +46,267 @@
...
+  // function end instruction
+  bool in_function_body() const;
+
+  // Returns true if called after a label instruction but before a branch
+  // instruction
+  bool in_block() const;
+
+  // Keeps track of ID definitions and uses.
+  class UseDefTracker {
+   public:
```
dneto0: Could be a separate class (not nested) with unit tests.  But probably ok.
Coder's choice.

dneto0:
-1
- At least two bugs 
- Need blurbs on a struct and some methods
- Prefer to rename the ProcessIDs method, and probably break out the entry-point collection one-liner.

Also, I defer to @umar456 about the code motion.

source/validate.cpp
umar456: These types of checks should ideally be performed by the SsaPass and after the CfgPass. This will allow you to prove dominance. Some of this functionality is already implemented in the current version of the SsaPass.

source/validate_types.h
```diff
@@ -1,241 +0,0 @@
-// Copyright (c) 2015-2016 The Khronos Group Inc.
```
umar456: This file needs to be removed from CMakeLists.txt

source/validate.h
```diff
@@ -36,33 +46,267 @@
...
+        return std::make_pair(false, spv_id_info_t{});
+      } else {
+        return std::make_pair(true, *found);
+      }
+    }
+
+    // Returns uses of IDs lacking defs.
+    std::unordered_set<uint32_t> FindUsesWithoutDefs() const {
+      auto diff = uses_;
+      for (const auto d : defs_) diff.erase(d.id);
```
umar456: Copy would be more efficient then erase in this case. Also a vector<uint32_t> would be sufficient for what you are trying to do.

dekimir: More efficient than O(1) of this `erase()`?  I doubt it.  The all-vector solution is quadratic overall; this is linear.

source/validate.cpp
dekimir: I agree this should go into SsaPass, and I've tried to put it there in two different ways, but I keep getting lots of test failures because of interactions of `SPV_VALIDATE_ID_BIT` and `SPV_VALIDATE_SSA_BIT`.  I'd rather leave it for the next PR.

source/validate.h
umar456: You are also copying the whole set in the line above. Probably a good idea to perform one operation instead of 2.

dekimir:
Addressed feedback so far.

source/validate.h
dekimir: @umar456 can you sketch some code to illustrate your proposed alternative?  I don't think I understand it properly, unless you really are suggesting replacing a linear method with a quadratic one.

umar456: My bad. I misread the function. I was thinking about doing a copy_if operation but that is for feasible in this case, but Isn't this doing the same thing as 'unresolvedForwardIds'?

dekimir: No.  `unresolvedForwardIds` will not catch a forward-declared ID that was used but never defined.

umar456: Perhaps I missed a couple of cases but that is exactly I am checking [here](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/validate.cpp#L382). I have several test cases in Validate.SSA.cpp which test for this scenario( [65](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/Validate.SSA.cpp#L65), [123](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/Validate.SSA.cpp#L123), basically anything with the %missing ID). Did you run into a test case I missed?

umar456: It looks like the ValidateID.cpp tests require the SPV_VALIDATE_SSA_BIT flag to work correctly. These flags are getting unnecessary complicated. Perhaps we need to run the tests by directly calling the function instead of spvValidate + flag.

I should also further document the responsibilities of each of the passes.

dekimir: To amend my previous statement: `unresolvedForwardIds` _could_ catch undefined IDs if it were integrated with the validate-ids pass.  But like I said above, that'll take some non-trivial work.

As for flags, the validate-id tests currently seem to work fine (for me) without the *SSA_BIT, because `UseDefTracker` is calculated unconditionally.  But when I tried to move that calculation into the SsaPass, I couldn't get tests to pass easily because of the check interference.

umar456:
For some reason Github is not showing the `validate_id.cpp` file in the Files changed tab. I completely missed those changes.

dekimir:
@umar456 you're right, it just says "522 additions, 741 deletions not shown". :confounded:  But these are an important part of this PR.

umar456:
All the checks with the "is not defined" error messages are redundant. These checks are performed by the SSA pass. These should be asserts instead.

dekimir:
@umar456 wouldn't an assert failure terminate the validator without checking the rest of the instructions?  I don't think that's what we want.

Perhaps we can just return false and omit the redundant diagnostics?

source/validate.cpp
```diff
@@ -332,14 +270,28 @@ void DebugInstructionPass(ValidationState_t& _,
...
+// Collects use-def info about an instruction's IDs.
+void ProcessIds(ValidationState_t& _, const spv_parsed_instruction_t& inst) {
+  if (inst.result_id) {
+    _.usedefs().AddDef(
+        {inst.result_id, inst.opcode,
+         std::vector<uint32_t>(inst.words, inst.words + inst.num_words)});
+  }
+  for (auto op = inst.operands; op != inst.operands + inst.num_operands; ++op) {
+    if (spvIsIdType(op->type))
+      _.usedefs().AddUse(inst.words[op->offset]);
```
umar456: This will allow the multiple assignments to the same ID. It used to be checked in [defineId](https://github.com/KhronosGroup/SPIRV-Tools/pull/74/files#diff-84a5aa9a556fde9600c6f5583e34ed81L219) but that check is missing here.

umar456: It will also allow the use of IDs which can not be forward declared but this case will be check by the SsaPass so that check is not necessary

umar456:
@dekimir Sure that will work. That condition should never be triggered.

dneto0:
FYI. I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/75 for the Travis CI issue.
For today, please ignore that failure. :-(

dneto0:
My low level comments have been addressed.
Now you'll need to address @umar456's feedback.

source/validate.cpp
dekimir: The multiple-ID-definitions check is [already covered](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/binary.cpp#L478) by the underlying parser, so it's redundant anywhere in the validator.

dekimir:
Addressed feedback so far.

dneto0:
Looks linear to me.
+1
Just want @umar456 's feedback.

umar456:
LGTM

dneto0:
Pushed onto master as 961f5dc544
I'm not sure why it didn't keep the same commit hash.

PR #77, opened 2016-01-18
Add disassembler option to show byte offset

dneto0:
The option is disabled by default.

The offset is printed in hex, as a comment after each instruction.


dneto0:
Rebased

AWoloszyn:
+2

PR #78, opened 2016-01-19
Since g++-4.9 has an ICE occasionally in travis add a retry.

AWoloszyn:
Try to re-run the build with ninja -j 1 if the build fails,
this should reduce the number of build failures we see.


AWoloszyn:
@dneto0   @dekimir PTAL.

If you examine the Travis build log you can see that the ICE did in fact occur, but the second build succeeded.
https://travis-ci.org/KhronosGroup/SPIRV-Tools/builds/103450811

dekimir:
+2

PR #79, opened 2016-01-21
Fix OpLoad pointee type validation.

dekimir:


dekimir:
@umar456 PTAL

umar456:
LGTM

source/validate_id.cpp
```diff
@@ -779,14 +779,21 @@ bool idUsage::isValid<SpvOpLoad>(const spv_instruction_t* inst,
...
-  auto type = usedefs_.FindDef(pointer.second.words[1]);
-  assert(type.first);
-  spvCheck(resultType.second.id != type.second.id,
-           DIAG(resultTypeIndex)
-               << "OpLoad Result Type <id> '" << inst->words[resultTypeIndex]
-               << " does not match Pointer <id> '" << pointer.second.id
-               << "'s type.";
-           return false);
+  auto pointerType = usedefs_.FindDef(pointer.second.words[1]);
+  if (!pointerType.first || pointerType.second.opcode != SpvOpTypePointer) {
```
dneto0: As an aside, I'm finding the use of a std::pair return value from usedefs_.FindDef very hard to read.  I'd rather have a bespoke struct with nice field values.

dneto0: This check for existence of the def and whether it's a pointer type suggests  a short utility function. 
The pattern is also used within OpVariable, OpCopyMemorySized.

Maybe we can do that later, along with a use of a DiagnosticStream object to shorten the return sequences.  But let's wait so we avoid making challenging merges for collaborators.

test/ValidateID.cpp
```diff
@@ -709,7 +709,7 @@ TEST_F(ValidateID, OpLoadResultTypeBad) {
 %5 = OpVariable %3 UniformConstant
 %6 = OpFunction %1 None %4
 %7 = OpLabel
-%8 = OpLoad %2 %5
+%8 = OpLoad %3 %5
```
dneto0: Hm.  Should stay as %2  ?!

dneto0:
-1 I think one of the test changes is in correct.  Which begs the question of whether the code is correct.
Also, you're adding diagnostics.  No better time than the present to add tests for those diagnostics.

test/ValidateID.cpp
dekimir: Why?  That would make it identical to the preceding test.

source/validate_id.cpp
dekimir: I'm OK doing the proposed refactorings in a later PR.

test/ValidateID.cpp
dneto0: Whoops.  I didn't realize it was supposed to be a *BAD* result.
And the GitHub differ makes it very difficult to see beyond the end of this segment.

dneto0:
+2

dneto0:
Ok.  Your test changes are correct.
I can wait for the diagnostics tests.
+2

dekimir:
Merged via 58852aa012462a0a6da8391374943fd89bfc4151.

PR #80, opened 2016-01-21
Don't fail validation in unimplemented cases.

dekimir:


dekimir:
@umar456 PTAL

umar456:
The unimplemented instruction tests will incorrectly pass with this change. It would probably be a good idea to have a flag which can be flipped in the unit tests and checks if the instruction is implemented.

dneto0:
Applying this change will get us closer to the point where we can turn on the validator by default.  That would give us a big win.

There's lots of other checks that are unimplimented but which don't cause a test failure, so this CL is in line with existing behaviour.

The danger with just printing to stderr is that it's too easily ignored.  As a middle ground, I suggest moving the TODO cases into real CASEs but inside a DISABLED_ test case.  The number and nature of DISABLED_ test cases is something we're much more likely to monitor.

-1 to make the TODOs more prominent by moving them into DISABLED_ test cases.

dekimir:
I'm confused about which tests you guys are talking about.  This is not test code -- it's actual `spv-val` code.  Validating perfectly legal programs currently fails if they contain any of these TODO instructions.

Adding `DISABLED_` (to what?) is unlikely to change anything.  And no unit tests are affected by this change.

dneto0:
Yikes.  I must be having a bad code reading day.
+2

dneto0:
Rebased

PR #82, opened 2016-01-21
Fix spvOpcodeIsScalarType() to include Boolean.

dekimir:
Also remove redundant validations of `OpConstant` and `OpConstantComposite`. Binary parser already performs these checks, so the validations can never be triggered.

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -431,6 +431,7 @@ int32_t spvOpcodeIsScalarType(const SpvOp opcode) {
   switch (opcode) {
     case SpvOpTypeInt:
     case SpvOpTypeFloat:
+    case SpvOpTypeBool:
       return true;
     default:
       return false;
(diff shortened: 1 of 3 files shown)


dekimir:
@umar456 PTAL

dneto0:
The logic is good.
But it's an implementation detail that the validator uses the parser.  There are some validator tests for these cases. Two of them  are DISABLED_:  ValidateID/DISABLED_OpSpecConstantBad and ValidateID/DISABLED_OpConstantBad
Let's enable them, changing the expected non-success status to whatever the binary parser gives us.  And add a comment that the specific non-success status is an implementation detail and can change to just about any other non-success status.

-1

dekimir:
Done.

dneto0:
+2
I'll squash and merge

dneto0:
Pushed into master as 0846e1745f

PR #83, opened 2016-01-22
Fix validation of OpStore object

dekimir:
Remove `spvOpcodeIsObject()` and test the spec's definition of what an object is.

Add `type_id` to `spv_id_info_t` and use it instead of `words[1]`. This triggered some asserts on tests, where the code incorrectly assumed words[1] had a type.  Remove the asserts and handle gracefully.

Also add tests for `OpStore` of a label, a void, and a function.


dekimir:
@umar456 PTAL

test/ValidateID.cpp
```diff
@@ -821,6 +821,53 @@ TEST_F(ValidateID, OpStoreTypeBad) {
   CHECK(spirv, SPV_ERROR_INVALID_ID);
 }
 
+TEST_F(ValidateID, OpStoreVoid) {
```
umar456: This should follow the naming convention of other Tests by adding the "Bad" Suffix

source/validate_id.cpp
```diff
@@ -454,7 +454,7 @@ bool idUsage::isValid<SpvOpConstantComposite>(const spv_instruction_t* inst,
           return false;
         }
         auto constituentResultType =
-            usedefs_.FindDef(constituent.second.words[1]);
+            usedefs_.FindDef(constituent.second.type_id);
         if (!constituentResultType.first ||
             componentType.second.opcode !=
                 constituentResultType.second.opcode) {
```
dneto0: I think you missed a words[1] instance on line 497.  For vector member of OpTypeMatrix.
Test case would be a matrix with a label as argument, for example.

dneto0:
-1 
I think you're missing one more place where words[1] was used.

Also, looking at the tests a high level concern (in general, not with this CL in particular) is that only checking the error status code makes the tests quite fragile.  If you mis-type some of the assembly cases, then you could generate the same result status code but for the wrong reason.  It would be more robust to check the diagnostic string, which should be more specific to the particular ID that's invalid.  Yes, that's more typing, but probably worth it in the long run.
For example, the binary parser has a test for every diagnostic it can issue.  Making those tests flushed out a few real bugs.

test/ValidateID.cpp
dekimir: No, they shouldn't.  Not only is this convention sparsely used (less than half of the total test names, even counting those with the _prefix_ instead of the suffix), it adds no useful information and provides a false sense of coverage.

Before 276a724b2580dc66936b5235f672c03241937cff, `OpTypeVectorGood` didn't say whether it was testing an int, float, or bool vector -- it was just letting us know that it was, well, "good".  Similarly, before 961f5dc54408b1516b9e4d85ee1fb4891b903f61, `OpExecutionModeEntryPointBad` was, in fact, testing a _missing_ entry point, not a bad one.  So the "Good/Bad" suffix says nothing about what the test actually covers.  When such a test fails, the name doesn't suggest which functionality failed.  A good name would tell me immediately that, eg, a Boolean vector failed while the int one passed.

Most of the FooGood/FooBad tests leave out lots of coverage; they seem to forget there's more than one good or bad case.  Over the past week, I've been working hard to improve that.

Furthermore, once we move away from the `CHECK` macro to a more nuanced way of testing the validation outcome (hello, `ValidateFixtures` :), the distinguishing power of the "Bad" suffix will further erode: does "Bad" mean failed parsing, failed validation, one of several non-success statuses, a particular error message, or what?  It's not clear, and therefore not useful.

dekimir:
Replaced the `wods[1]` instance as suggested.

Agree that the tests should use better expectations than just success/failure.  Will change that wholesale over in another PR.

dneto0:
Functionally looks good.  I don't want to block this further pending naming discussion.
+2

dneto0:
Squashed and pushed into master as a4342f3f442178e

PR #85, opened 2016-01-26
Qualify CMake function names with "spvtools" to avoid conflictions.

antiagainst:
If a third-party project defines a CMake function with the same
name, by importing that project's CMake configuration, we may end
up overwriting our own copy. Qualify all defined functions to
reduce that probability.


antiagainst:
Counterpart for https://github.com/google/shaderc/pull/103

dekimir:
+2

AWoloszyn:
Thank you I had run into this issue the other day, just haven't had time to make an MR for it.

PR #86, opened 2016-01-27
Fix validation of return value.

dekimir:


dekimir:
@umar456 PTAL

source/validate_id.cpp
```diff
@@ -1671,13 +1671,17 @@ bool idUsage::isValid<SpvOpReturnValue>(const spv_instruction_t* inst,
...
   auto valueType = usedefs_.FindDef(value.second.type_id);
-  assert(valueType.first);
+  if (!valueType.first || SpvOpTypeVoid == valueType.second.opcode) {
+    DIAG(valueIndex) << "OpReturnValue value's type <id> '"
+                     << value.second.type_id << "' is missing or void.";
+    return false;
+  }
   // NOTE: Find OpFunction
   const spv_instruction_t* function = inst - 1;
   while (firstInst != function) {
```
dneto0: Not part of this CL, but this scan backward for the start of the function looks a little worrying.  Something should make sure that we don't run off the front of the module and get an invalid access.

test/ValidateID.cpp
```diff
@@ -1322,7 +1322,7 @@ TEST_F(ValidateID, OpReturnValueConstantGood) {
   const char* spirv = R"(
 %1 = OpTypeVoid
 %2 = OpTypeInt 32 0
-%3 = OpTypeFunction %2 %2
```
dneto0: Extraneous removal of the argument type?   But ok by me.   It's a simplification.

test/ValidateID.cpp
```diff
@@ -1344,19 +1345,71 @@ TEST_F(ValidateID, OpReturnValueVariableGood) {
      OpFunctionEnd)";
   CHECK(spirv, SPV_SUCCESS);
```
dneto0: This example looks wrong. Validation should fail. ?
  %7 is the result of an OpVariable, and hence is a pointer to 32-bit int.
But the function returns 32-bit int (%2), and the return value is a pointer-to-32-bit int in Function storage class.

test/ValidateID.cpp
```diff
@@ -1344,19 +1345,71 @@ TEST_F(ValidateID, OpReturnValueVariableGood) {
...
+}
+
+TEST_F(ValidateID, OpReturnValueIsVoid) {
+  const char* spirv = R"(
+%1 = OpTypeVoid
+%2 = OpTypeInt 32 0
+%3 = OpTypeFunction %1
+%5 = OpFunction %1 None %3
+%6 = OpLabel
+%7 = OpFunctionCall %1 %5
```
dneto0: Ha ha. Infinite recursion.  But works for this (static) validation case.

dneto0:
-1.  Seems like one test is wrong: the test that intends to return the value stored in a variable.  Needs a load.   Or did it intend to return the pointer that is the storage allocated for the variable.

test/ValidateID.cpp
dekimir: Not extraneous, since `OpFunctionParameter` is absent after `OpFunction`.  The old assembly was incorrect.

dneto0: Ah, good call.

dekimir:
Fixed the incorrect test.  Also fixed the case of truly returning a pointer.

source/validate_id.cpp
dekimir: Something _is_ making sure of that -- note the `firstInst` check. ;)

dneto0:
+2
@dekimir please squash and commit.  I'm not able to at the moment.

dekimir:
Merged with 1e157bc.

PR #87, opened 2016-01-28
Initialize vector<string>s on first use, not statically.

dekimir:


dekimir:
@umar456 sorry to mess up your pretty code; forced by style guide.

umar456:
Looks good to me.

dneto0:
-1

I still think this is forbidden by the style guide.

In particular: 
"Objects with static storage duration, including global variables, static variables, static class member variables, and function static variables, must be Plain Old Data (POD): only ints, chars, floats, or pointers, or arrays/structs of POD."
And then explains why: indeterminate destruction order in this case, and the potential for keeping references to data that has already been destroyed.

Yes, these have static storage duration, as per:
http://en.cppreference.com/w/cpp/language/storage_duration#Storage_duration
because the variables are marked "static", even though they are function static.

Sure, you could argue it's not a problem now in the (test) code. But same was true before this CL.

Seems a sideways move. Not really solving the problem.

dekimir:
@dneto0 Um, pointers? ;)

> must be Plain Old Data (POD): only ints, chars, floats, or pointers, or arrays/structs of POD.

dneto0:
Whoops.  I missed seeing that it's a "new", and hence my whole interpretation is wrong.
This leaks an object, but I'm ok with that in test code, and it's the recommended solution for this kind of thing.

+2

PR #91, opened 2016-01-29
fix the off-by-one bug, found in Windows Debug build tests

Qining:
validate_types.cpp:line 374 in ValidationState_t::registerCapability() triggers Debug assertion on Windows build, as we are dereferencing module_capabilities_[56] while the size of the vector is 56. 

Updating the size of the vector to the correct number of buckets required fixed the bug.

### DIFF
--- a/source/validate_types.cpp
+++ b/source/validate_types.cpp
@@ -204,8 +204,7 @@ bool IsInstructionInLayoutSection(ModuleLayoutSection layout, SpvOp op) {
   return out;
 }
 
-// NOTE: We are using vector to map the ID of the capability to its
-// availability. This variable is the maximum ID of capabilities.
+// This variable is the maximum ID of capabilities.
 static const size_t kCapabilitiesMaxValue =
     SpvCapabilityStorageImageWriteWithoutFormat;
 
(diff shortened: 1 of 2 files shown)


source/validate_types.cpp
```diff
@@ -209,7 +209,7 @@ bool IsInstructionInLayoutSection(ModuleLayoutSection layout, SpvOp op) {
 // entire range of the capability. Currently capability 26 is missing so the
 // count is actually 54
 // The maximum ID of capabilities plus one
-static const size_t kCapabilitiesCount = 56;
+static const size_t kCapabilitiesMaxValue = 56;
```
dekimir: So is the comment incorrect, then?

dekimir:
@umar456 what do you think?

umar456:
This commit looks correct based on the spirv.h version I have on my system, but the comment needs to be updated to reflect the new name.

source/validate_types.cpp
```diff
@@ -205,11 +205,9 @@ bool IsInstructionInLayoutSection(ModuleLayoutSection layout, SpvOp op) {
 }
 
 // NOTE: We are using vector to map the ID of the capability to its
```
dneto0: I agree with @dekimir's advice in general that the blurb for a construct should describe the intrinsic meaning of the object.   Don't comment on the use.  So the text about "We are using the vector to map ...." should really be a comment at the declaration of module_capabilities_.  There is where you describe how the vector is indexed.

dneto0:
-1

AWoloszyn:
I would ask that the commit message is a bit more descriptive than "fixed the bug"

dneto0:
+2

dneto0:
Squashed, rebased, and pushed into master as 7dbfdda6

PR #92, opened 2016-01-29
Forbid Generic variables.

dekimir:


dekimir:
@umar456 PTAL

umar456:
Looks good.

On a side note. I wanted to talk about how to represent the instructions names in error messages. I have been representing them by removing the Op prefix. What are your thoughts?

dekimir:
I don't have a strong opinion either way.  I've been using the Op prefix because the spec always uses it when setting the rules.  Some of my error messages are copied verbatim from the spec.

umar456:
Makes sense.

test/Validate.Storage.cpp
```diff
@@ -141,4 +141,34 @@ INSTANTIATE_TEST_CASE_P(MatrixOp, ValidateStorage,
...
+TEST_F(ValidateStorage, GenericVariableOutsideFunction) {
+  const auto str = R"(
+          OpCapability Kernel
+          OpMemoryModel Logical OpenCL
+%intt   = OpTypeInt 32 1
+%ptrt   = OpTypePointer Function %intt
+%var    = OpVariable %ptrt Generic
+)";
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_LAYOUT, ValidateInstructions());
```
umar456: I don't think this should be a layout error

dekimir: An instruction error, perhaps?

umar456: Sounds good to me.

dneto0: Or the catchall INVALID_BINARY could work too.  Unless we have something more specific that is still accurate.

dneto0:
LGTM

dekimir:
Changed to `INVALID_BINARY`.

PR #93, opened 2016-02-01
Use AssemblyGrammar for capability validation

dekimir:


dekimir:
@umar456 PTAL

test/Validate.Capability.cpp
```diff
@@ -522,28 +522,39 @@ make_pair("OpDecorate %intt BuiltIn PatchVertices\n",             TessellationDe
...
-make_pair("OpDecorate %intt BuiltIn NumEnqueuedSubgroups\n",      KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn SubgroupId\n",                KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn SubgroupLocalInvocationId\n", KernelDependencies()),
+make_pair("OpDecorate %intt BuiltIn NumWorkgroups\n",             ShaderDependencies()),
+make_pair("OpDecorate %intt BuiltIn WorkgroupSize\n",             ShaderDependencies()),
+make_pair("OpDecorate %intt BuiltIn WorkgroupId\n",               ShaderDependencies()),
+make_pair("OpDecorate %intt BuiltIn LocalInvocationId\n",         ShaderDependencies()),
+make_pair("OpDecorate %intt BuiltIn GlobalInvocationId\n",        ShaderDependencies()),
+make_pair("OpDecorate %intt BuiltIn LocalInvocationIndex",        ShaderDependencies()),
+make_pair("OpCapability Shader\n"
```
dekimir: So this may be a spec bug, but it appears that the `BuiltIn` keyword itself requires the `Shader` capability, even if the argument following it applies to the `Kernel` or `GLCompute` execution model. :confused: Our opcode tables reflect that, requiring the above line for the test to pass.

source/spirv_definition.h
```diff
@@ -39,6 +43,16 @@ typedef uint64_t spv_capability_mask_t;
 #define SPV_CAPABILITY_AS_MASK(capability) \
   (spv_capability_mask_t(1) << (capability))
 
+namespace libspirv {
+template <typename Functor>
```
dneto0: Need a method blurb.  :-)

source/spirv_definition.h
```diff
@@ -39,6 +43,16 @@ typedef uint64_t spv_capability_mask_t;
...
+inline void ForEach(spv_capability_mask_t capabilities, Functor f) {
+  for (int cap = SpvCapabilityMatrix;
+       cap <= SpvCapabilityStorageImageWriteWithoutFormat; ++cap)
```
dneto0: It would be nice if there were a "kCapabilityMaxValue" we could reuse.
There's a kCapabilitiesMaxValue static size_t variable in validate_types.cpp

Maybe it's time to put that definition in this header.  
I can wait for another CL, however.

Also, I'd much rather have that max value be defined by some file in the SPIR-V registry, either spirv.h or something generated from spirv.json.  So another reason this can wait if you want.

source/validate_instruction.cpp
```diff
@@ -27,474 +27,72 @@
...
+      !_.HasAnyOf(opcode_desc->capabilities))
     return _.diag(SPV_ERROR_INVALID_CAPABILITY)
-           << mode + " sample address mode requires Kernel capability";
+           << "Opcode " << spvOpcodeString(inst->opcode)
+           << " requires one of these capabilities: "
+           << ToString(opcode_desc->capabilities, _.grammar());
+  for (int i = 0; i < inst->num_operands; ++i) {
+    spv_operand_desc operand_desc;
+    if (SPV_SUCCESS ==
+            _.grammar().lookupOperand(inst->operands[i].type,
```
dneto0: This is correct for most operands.  It is incorrect when the operand is the bitwise-OR of masks.

See "Set 5" in the definition of spv_operand_type_t.

For example, SPV_OPERAND_TYPE_IMAGE.  The Bias bit requires Shader, but Lod depends on nothing, and MinLod depends on MinLod.

source/validate_types.cpp
```diff
@@ -301,85 +304,29 @@ bool ValidationState_t::in_block() const {
...
-  } while (getDependantCapability(capability) != capability);
-
-  // Set Base capability
-  module_capabilities_[capability] = true;
+  module_capabilities_[cap] = true;
+  spv_operand_desc desc;
+  if (SPV_SUCCESS ==
+      grammar_.lookupOperand(SPV_OPERAND_TYPE_CAPABILITY, cap, &desc))
+    libspirv::ForEach(desc->capabilities,
+                      [this](SpvCapability c) { registerCapability(c); });
```
dneto0: Recursion FTW!

source/validate_types.cpp
```diff
@@ -301,85 +304,29 @@ bool ValidationState_t::in_block() const {
...
+bool ValidationState_t::hasCapability(SpvCapability cap) const {
   return module_capabilities_[cap];
 }
 
+bool ValidationState_t::HasAnyOf(spv_capability_mask_t capabilities) const {
+  if (!capabilities)
+    return true;  // No capabilities requested: trivially satisfied.
+  bool found = false;
+  libspirv::ForEach(capabilities, [&found, this](SpvCapability c) {
+    found |= hasCapability(c);
```
dneto0: Alternate is to just return early.

Coder's choice.

source/validate_types.cpp
```diff
@@ -301,85 +304,29 @@ bool ValidationState_t::in_block() const {
...
+  spv_operand_desc desc;
+  if (SPV_SUCCESS ==
+      grammar_.lookupOperand(SPV_OPERAND_TYPE_CAPABILITY, cap, &desc))
+    libspirv::ForEach(desc->capabilities,
+                      [this](SpvCapability c) { registerCapability(c); });
 }
 
-bool ValidationState_t::hasCapability(SpvCapability cap) {
+bool ValidationState_t::hasCapability(SpvCapability cap) const {
   return module_capabilities_[cap];
```
dneto0: This begs the question: why not store module_capabilities_ as a bitmask in the first place.  Then HasAnyOf is just a bitwise-AND followed by test against zero.

I'm ok either way.

test/Validate.Capability.cpp
```diff
@@ -522,28 +522,39 @@ make_pair("OpDecorate %intt BuiltIn PatchVertices\n",             TessellationDe
...
-make_pair("OpDecorate %intt BuiltIn EnqueuedWorkgroupSize\n",     KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn GlobalOffset\n",              KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn GlobalLinearId\n",            KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn SubgroupSize\n",              KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn SubgroupMaxSize\n",           KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn NumSubgroups\n",              KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn NumEnqueuedSubgroups\n",      KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn SubgroupId\n",                KernelDependencies()),
-make_pair("OpDecorate %intt BuiltIn SubgroupLocalInvocationId\n", KernelDependencies()),
+make_pair("OpDecorate %intt BuiltIn NumWorkgroups\n",             ShaderDependencies()),
```
dneto0: Definitely worth a comment here to say that the only reason you need ShaderDependencies for NumWorkgroups through LocalInvocationIndex is that *BuiltIn* has a dependency on the Shader capability.
And that that seems funny in the first place, i.e. is likely a spec bug.

dneto0:
Really like seeing net -550 lines of code!

Just one functionality bug with mask operands.  Maybe we finally need an spvOperandIsMask utility function.  Other spots in the codebase  that need to know this (in binary.cpp and text.pp) process the difference in a switch statement, so there's an explicit list of cases there.  Not ideal to have the same knowledge spread in multiple places, but that's how it is for now.
This suggests a requirement for the next version of spirv.json to capture what enums are masks vs. not.  Actually, the spirv.json seems to have that, but it's not reflected in spirv.h

dneto0:
-1 
for the mask operand issue.

dekimir:
Addressed feedback.

dneto0:
+2
I'll squash and merge

dneto0:
Squashed and pushed into master as  0c8bdfe163d

PR #94, opened 2016-02-02
Validate mask operand capabilities in a more data driven way

dneto0:
Adds spvOperandIsConcreteMask.

Use spvOperandIsConcreteMask to determine if an operand is a bitmask.  Check its individual
bits in a generic way.


dekimir:
+2

PR #95, opened 2016-02-02
Enable ctest.

dekimir:


CMakeLists.txt
```diff
@@ -258,6 +259,9 @@ if (NOT ${SPIRV_SKIP_EXECUTABLES})
       ${CMAKE_CURRENT_SOURCE_DIR}
       ${gmock_SOURCE_DIR}/include ${gtest_SOURCE_DIR}/include)
     target_link_libraries(UnitSPIRV PRIVATE ${SPIRV_TOOLS} gmock)
+    add_test(NAME spirv-tools-testsuite
+      COMMAND UnitSPIRV
+      WORKING_DIRECTORY ${SPIRV-Tools_BINARY_DIR})
```
AWoloszyn: I think by convention if our current file is the root of the project we just use ${CMAKE_CURRENT_BINARY_DIR} instead of spelling out the project itself. We are at least doing that with ${CMAKE_CURRENT_SOURCE_DIR}

Not terribly important to me

AWoloszyn:
One comment from me, optional.
Otherwise +2.

dekimir:
Removed `WORKING_DIRECTORY` -- the default is what we want.

AWoloszyn:
+2

Issue #89, opened 2016-01-28
Get rid of SPV_VALIDATE_*_BITs


dekimir:
Let's discuss whether and how we can get rid of these options.  I suspect that the vast majority of users doesn't need sophisticated ways to turn off certain validations.  Furthermore, the options are intertwined in unexpected ways -- for instance, multiple validations need `_.current_layout_section_`, but this is only computed when `SPV_VALIDATE_LAYOUT_BIT` is set.

Unit tests also shouldn't need fine-grained controls of this kind -- those that do aren't really _unit_ tests.  Moreover, it's fishy design to add so much API just for testing purposes.

My opinion is that we should eliminate these controls altogether, perform full validation always, and fix unit tests to have minimally invalid assembly or invoke only specific validations directly.

@umar456 @dneto0 @AWoloszyn ^^

umar456:
I was planning on using spvBinaryParse like [this](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/validate.cpp#L325) and calling the passes individually. This should be easily generalizable in a fixture.

umar456:
I tried a couple of approaches to get around this. One approach was to pass the `*Pass` function pointers to the `ValidateInstructions` function. Another was to ignore the errors based on the type of flags passed into the `ValidateState_t::diag` function. I don't think either of these approaches solved the problem we have been experiencing.

The only solution would be to parse the entire binary for each of the passes. This will allow us to test for each type of pass independently. This will obviously be slower but might solve this issue.

Thoughts?

dekimir:
@umar456 I don't like the idea of needless repeated parsing.  Is it not possible to fix the test assemblies not to trigger unrelated validations?  A lot of the time it's something trivial, like missing `OpMemoryModel`.

dekimir:
@umar456 is it OK if we assign this to someone on our team?

umar456:
sure

antiagainst:
In `master` as b6ccd0d89114d94026871edd31cf886280dad20a.

PR #96, opened 2016-02-03
Eliminate parameter-order check from ID validation

dekimir:
Also ensure we don't get thrown off by too many parameters.

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -964,12 +964,6 @@ bool idUsage::isValid<SpvOpFunctionParameter>(const spv_instruction_t* inst,
   size_t paramIndex = 0;
   assert(firstInst < inst && "Invalid instruction pointer");
   while (firstInst != --inst) {
-    spvCheck(SpvOpFunction != inst->opcode &&
-                 SpvOpFunctionParameter != inst->opcode &&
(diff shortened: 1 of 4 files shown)


dekimir:
@umar456 PTAL

umar456:
I already perform these checks in the layout pass (see [unit test](https://github.com/KhronosGroup/SPIRV-Tools/pull/96/files#diff-6dc2086e9da925d662089f2876b41bb7L165)). A better approach would be to remove that check from validate_id.

dekimir:
@umar456 The "unit test" link just takes me to this PR's diff :question:

umar456:
Sorry, Here is the link https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/Validate.Layout.cpp#L165

dneto0:
Functionality looks good to me.
But agree deleting code would be better.  Suggest @dekimir try removing the check from validate_id but keep the tests, possibly adjusting the failure result code if it's different.

dekimir:
Eliminated the check, kept the tests.  The problem with Umar's test is that it doesn't exercise the code in question due to fewer validation bits turned on.  #89 strikes again. :(

dneto0:
+2

PR #97, opened 2016-02-03
Make BaseTypeNullable a transitive check.

dekimir:


dekimir:
@umar456 PTAL

source/validate_id.cpp
```diff
@@ -609,64 +609,54 @@ bool idUsage::isValid<SpvOpConstantSampler>(const spv_instruction_t* inst,
   return true;
 }
 
+bool IsBaseTypeNullable(const std::vector<uint32_t>& instruction,
```
dneto0: Missing a method blurb.
In particular, it's not immediately clear why you'd need a the usedefs.
Also, should say that you expect the instruction to be a type-defining instruction, and you're checking that the type being defined is what is nullable.

dneto0: Also, why is it "Base" type?

dneto0:
-1 for the missing blurb.  Potentially also for the "Base" part of the name, which doesn't seem necessary.

Optional: Add a negative test case for OpTypeRuntimeArray.

dekimir:
Renamed to `IsTypeNullable`, added a blurb, and added a neg test for `OpTypeRuntimeArray`.

dneto0:
+2
I'll squash and merge.

dekimir:
Also clarified the error message a bit.

PR #99, opened 2016-02-08
More parsing tests for literals with leading minus

dneto0:
Also checks some hex literal cases.

This addresses part of
https://github.com/KhronosGroup/SPIRV-Tools/issues/45
by removing the parseNumber case for "-0" on unsigned
integers.  We don't care about that platform difference
at the level of std::istringstream, since we reject it
at a higher parsing level.

### DIFF
--- a/test/TextToBinary.Constant.cpp
+++ b/test/TextToBinary.Constant.cpp
@@ -118,8 +118,9 @@ TEST_P(OpConstantValidTest, ValidTypes) {
                             "%2 = OpConstant %1 " +
                             GetParam().constant_value + "\n";
   std::vector<uint32_t> instructions;
-  EXPECT_THAT(CompiledInstructions(input),
-              Eq(GetParam().expected_instructions));
+  EXPECT_THAT(CompiledInstructions(input), Eq(GetParam().expected_instructions))
+      << " type: " << GetParam().constant_type
+      << " literal: " << GetParam().constant_value;
 }
 
 // clang-format off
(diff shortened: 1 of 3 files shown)


test/TextToBinary.cpp
```diff
@@ -319,15 +319,71 @@ TEST(AssemblyContextParseNarrowUnsignedIntegers, Sample) {
...
+  const spv_result_t ec = SPV_FAILED_MATCH;
+  uint32_t u32;
+
+  // Invalid parse.
+  EXPECT_EQ(SPV_FAILED_MATCH, context.parseNumber("", ec, &u32, ""));
+  EXPECT_EQ(SPV_FAILED_MATCH, context.parseNumber("0=", ec, &u32, ""));
+
+  // Valid values.
+  EXPECT_EQ(SPV_SUCCESS, context.parseNumber("0", ec, &u32, ""));
+  EXPECT_EQ(0, u32);
```
dekimir: Should be `0u` -- see Travis failures.

dneto0: Fixed in next commit

test/TextToBinary.Constant.cpp
```diff
@@ -133,12 +134,33 @@ INSTANTIATE_TEST_CASE_P(
...
+         MakeInstruction(SpvOpConstant, {1, 2, 0})})},
+      {"OpTypeInt 16 1", "-0",
+        Concatenate({MakeInstruction(SpvOpTypeInt, {1, 16, 1}),
+         MakeInstruction(SpvOpConstant, {1, 2, 0})})},
+      {"OpTypeInt 16 1", "-0x0",
+        Concatenate({MakeInstruction(SpvOpTypeInt, {1, 16, 1}),
+         MakeInstruction(SpvOpConstant, {1, 2, 0})})},
+      {"OpTypeInt 16 1", "-32768",
+        Concatenate({MakeInstruction(SpvOpTypeInt, {1, 16, 1}),
+         MakeInstruction(SpvOpConstant, {1, 2, uint32_t(-32768)})})},
```
dekimir: "Convert arithmetic types using brace initialization."

https://google.github.io/styleguide/cppguide.html#Casting

test/TextToBinary.Constant.cpp
```diff
@@ -181,12 +221,61 @@ INSTANTIATE_TEST_CASE_P(
...
+         MakeInstruction(SpvOpConstant, {1, 2, 0xffffffffu, 0x7fffffffu})})},
+      {"OpTypeInt 64 1", "0x7fffffff",
+        Concatenate({MakeInstruction(SpvOpTypeInt, {1, 64, 1}),
+         MakeInstruction(SpvOpConstant, {1, 2, 0x7fffffffu, 0})})},
+    }));
+// clang-format on
+
+// A test case for checking OpConstant with invalid literals with a leading minus.
+struct InvalidLeadingMinusCase {
+  std::string type;
```
dekimir: I wonder if this should just be `uint32_t uint_width`.  It currently allows the user to specify a signed type.

dekimir:
+2 despite nits.

test/TextToBinary.Constant.cpp
dneto0: The style guide has been changing in this regard.  That's new advice.  We never did go back and change *all* our existing uint32_t( .. ) to static_cast.  I'll use the brace initialization in future.

dneto0:
Squashed and pushed into master as feda995c841

test/TextToBinary.Constant.cpp
dneto0: Acknowledged.
Hopefully what I've got is not too confusing. It's more in line with the immediately previous test, which does use signed types as well. But I see your point.
I went with what I had since you gave it a +2. :-)

PR #100, opened 2016-02-10
Reject float literals with two leading signs

dneto0:
E.g. --1 should be rejected.

### DIFF
--- a/include/util/hex_float.h
+++ b/include/util/hex_float.h
@@ -364,11 +364,11 @@ class HexFloat {
       // the significand is not zero.
       significand_is_zero = false;
       significand |= first_exponent_bit;
-      significand  = static_cast<uint_type>(significand >> 1);
+      significand = static_cast<uint_type>(significand >> 1);
     }
 
     while (exponent < min_exponent) {
-      significand  = static_cast<uint_type>(significand >> 1);
+      significand = static_cast<uint_type>(significand >> 1);
       ++exponent;
     }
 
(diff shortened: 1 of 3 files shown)


include/util/hex_float.h
```diff
@@ -711,13 +711,37 @@ std::ostream& operator<<(std::ostream& os, const HexFloat<T, Traits>& value) {
   return os;
 }
 
+// Returns true if the negate_value is true and the next character on the
```
AWoloszyn: Returns true if the negate_value parameter is true 
or 
Returns true if negate_value is true.

AWoloszyn:
-1 for comment other than that looks good

Also I expect +1.0p3 will fail parse (I don't think we handle unary + in hex-float), but that is a separate issue.

dneto0:
All feedback addressed.  Amended the previous commit.

test/TextToBinary.Constant.cpp
```diff
@@ -279,6 +279,48 @@ INSTANTIATE_TEST_CASE_P(
...
+  uint32_t width;
+  std::string literal;
+};
+
+using OpConstantInvalidFloatConstant = spvtest::TextToBinaryTestBase<
+    ::testing::TestWithParam<InvalidFloatConstantCase>>;
+
+TEST_P(OpConstantInvalidFloatConstant, Samples) {
+  // Check both kinds of instructions that take literal floats.
+  for (const auto& instruction : {"OpSpecConstant", "OpSpecConstant"}) {
```
dekimir: Is it me, or are both elements of the initializer list the same??

dneto0: Whoops.  Yes, first one should be OpConstant.

dneto0:
All feedback addressed.

dekimir:
+2

dneto0:
Squashed and pushed into master via 000cad9cc6b

PR #101, opened 2016-02-11
Float literal parsing fixes

dneto0:
- For 32- and 64-bit floats, overflow is a parse error

  This works around a difference between Xcode's istringstream
  and other platforms.  Xcode's runtime library will happlily
  "round up" overflow values to infinity.  We want to make it fail.

- When parsing a float fails due to bad syntax, follow C++11
  behaviour for operator>> and set the value to zero.

- When parsing a 32-bit or 64-bit float overflows, follow C++11
  behaviour for operator>> and set the value to the nearest
  normal value: either max or lowest finite value for the type.

- Add FloatProxy<T>::max() and ::lowest()

- Make 16-bit overflow behaviour more consistent: we always get a
  16-bit infinity of the right sign, whether the original string
  is a normal value for 32-bit or an overflow value for 32-bit.
  That matches our earlier intent.
  Added TODO's to make 16-bit overflow always an error, just like
  for 32-bit and 64-bit.

- Simplify normal parsing of Float16 values by delegating to
  normal parsing of 32-bit floats.

### DIFF
--- a/include/util/hex_float.h
+++ b/include/util/hex_float.h
@@ -772,8 +772,11 @@ inline bool RejectParseDueToLeadingSign(std::istream& is, bool negate_value,
 // If negate_value is true then the number may not have a leading minus or
 // plus, and if it successfully parses, then the number is negated before
 // being stored into the value parameter.
-// If the value is an infinity, then set the fail bit on the stream, and set
-// the value to 0.
+// If the value cannot be correctly parsed, then set the fail bit on the
+// stream, and set the value to zero.
+// If the value overflows the target floating point type, then set the fail
+// bit on the stream and set the value to the nearest finite value for the
+// type, which can either be positive or negative.
 template <typename T, typename Traits>
 inline std::istream& ParseNormalFloat(std::istream& is, bool negate_value,
                                       HexFloat<T, Traits>& value) {
--- a/include/util/hex_float.h
+++ b/include/util/hex_float.h
@@ -32,7 +32,6 @@
 #include <cmath>
 #include <cstdint>
 #include <iomanip>
-#include <iostream>
 #include <limits>
 
 #include "bitutils.h"
@@ -46,8 +45,17 @@ class Float16 {
   static bool isNan(const Float16 val) {
     return ((val.val & 0x7C00) == 0x7C00) && ((val.val & 0x3FF) != 0);
   }
+  // Returns true if the given value is any kind of infinity.
+  static bool isInfinity(const Float16 val) {
+    return ((val.val & 0x7C00) == 0x7C00) && ((val.val & 0x3FF) == 0);
+  }
   Float16(const Float16& other) { val = other.val; }
   uint16_t get_value() const { return val; }
+
+  // Returns the maximum normal value.
+  static Float16 max() { return Float16(0x7bff); }
+  // Returns the lowest normal value.
+  static Float16 lowest() { return Float16(0xfbff); }
 
  private:
   uint16_t val;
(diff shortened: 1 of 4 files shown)


include/util/hex_float.h
```diff
@@ -722,7 +759,7 @@ inline bool RejectParseDueToLeadingSign(std::istream& is, bool negate_value,
     if (next_char == '-' || next_char == '+') {
       // Fail the parse.  Emulate standard behaviour by setting the value to
       // the zero value, and set the fail bit on the stream.
-      value = HexFloat<T, Traits>(typename HexFloat<T, Traits>::uint_type(0));
+      value = HexFloat<T, Traits>(typename HexFloat<T, Traits>::uint_type{0});
```
AWoloszyn: I am not 100% sure this won't explode on VS2013

AWoloszyn:
Other than watching out for the VS2013 issue, looks good to me.

include/util/hex_float.h
dekimir: Well, https://github.com/KhronosGroup/SPIRV-Tools/commit/000cad9cc6ba4d81f11893f25f676471e5e9fe86#diff-e1047a2b988b6e4fc371902d06c57a7bR1104 didn't explode, so there's hope.

include/util/hex_float.h
```diff
@@ -735,6 +772,8 @@ inline bool RejectParseDueToLeadingSign(std::istream& is, bool negate_value,
 // If negate_value is true then the number may not have a leading minus or
 // plus, and if it successfully parses, then the number is negated before
 // being stored into the value parameter.
+// If the value is an infinity, then set the fail bit on the stream, and set
```
dekimir: Infinity?  Or overflow?

test/HexFloat.cpp
```diff
@@ -1054,8 +1054,8 @@ FloatParseCase<T> GoodFloatParseCase(std::string literal, bool negate_value,
 INSTANTIATE_TEST_CASE_P(FloatParse, ParseNormalFloatTest,
                         ::testing::ValuesIn(std::vector<FloatParseCase<float>>{
                             // Failing cases due to trivially incorrect syntax.
-                            BadFloatParseCase<float>("abc", false),
-                            BadFloatParseCase<float>("abc", true),
+                            BadFloatParseCase<float>("abc", false, 0.0f),
```
dekimir: We no longer need to explicitly specify the parameter type for the `*FloatParseCase` methods -- the compiler should auto-resolve it based on the argument types.

test/HexFloat.cpp
```diff
@@ -1113,11 +1119,118 @@ INSTANTIATE_TEST_CASE_P(
...
-// TODO(awoloszyn): Add fp16 tests and HexFloatTraits.
+// A test case for detecting infinities.
+template <typename T>
+struct InfinityParseCase {
+  std::string input;
+  bool expect_success;
+  T expected_value;
+};
+
+using FloatProxyParseInfinityFloatTest =
```
dekimir: Are we testing infinity parsing or overflow parsing?  Seems like the latter, based on test cases.

include/util/hex_float.h
dneto0: Good point.  Actually, the comment is not quite right.  If it fails to parse as a number, the value is set to 0.  Will fix in followup commit.

test/HexFloat.cpp
dneto0: True for the float case, and I'll fix that.  Not so for the Float16 since I'm supplying uint16_t constructor arguments instead of Float16.  Doesn't seem much of a win to change

```BadFloatParseCase<Float16>("abc", false, uint16_t{0})```
to
```BadFloatParseCase("abc", false, Float16(uint16_t{0}))```

test/HexFloat.cpp
dneto0: Good point.  Changing Infinity to Overflow in this struct name and test classes and names.

dneto0:
Addressed all feedback, except for the caution about VS2013.

dekimir:
+2

dneto0:
Squashed and pushed into master as 6bad02c32

PR #102, opened 2016-02-12
Float overflow is fail and set max value

dneto0:


dneto0:
Hmm. didn't squash correctly.

PR #103, opened 2016-02-12
Parse overflow for fp16  like for 32- and 64-bit

dneto0:
In that case, set the stream fail bit, and set the maximum normal
value or lowest normal value, depending on the sign of the original
value.

### DIFF
--- a/test/TextToBinary.cpp
+++ b/test/TextToBinary.cpp
@@ -514,23 +514,31 @@ TEST(AssemblyContextParseFloat16, Infinities) {
   AssemblyContext context(AutoText(""), nullptr);
   const spv_result_t ec = SPV_FAILED_MATCH;
   spvutils::HexFloat<spvutils::FloatProxy<spvutils::Float16>> f(0.0f);
+  const uint32_t f16_max = uint32_t{0x7bff};
+  const uint32_t f16_low = uint32_t{0xfbff};
 
-  EXPECT_EQ(SPV_SUCCESS, context.parseNumber("0", ec, &f, ""));
-  EXPECT_TRUE(!f.value().isInfinity());
-  EXPECT_EQ(SPV_SUCCESS, context.parseNumber("1.5", ec, &f, ""));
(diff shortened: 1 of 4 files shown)


test/TextToBinary.cpp
```diff
@@ -514,23 +514,31 @@ TEST(AssemblyContextParseFloat16, Infinities) {
...
-  EXPECT_TRUE(f.value().isInfinity());
-  EXPECT_EQ(SPV_SUCCESS, context.parseNumber("-1e400", ec, &f, ""));
-  EXPECT_TRUE(f.value().isInfinity());
+  const uint32_t f16_max = uint32_t{0x7bff};
+  const uint32_t f16_low = uint32_t{0xfbff};
+
+  EXPECT_EQ(SPV_SUCCESS, context.parseNumber("-0.0", ec, &f, ""));
+  EXPECT_EQ(uint16_t{0x8000}, f.value().getAsFloat().get_value());
+  EXPECT_EQ(SPV_SUCCESS, context.parseNumber("1.0", ec, &f, ""));
+  EXPECT_EQ(uint16_t{0x3c00}, f.value().getAsFloat().get_value());
```
dekimir: Why are the above cases in the test called "Infinities"?

dneto0: You're right.  They are Overflow.

dneto0:
Addressed feedback, and expanded the test comments to say what value is set on overflow.  Actually test those values as well.

test/TextToBinary.cpp
dneto0: The *old* behaviour of 16-bit floats was that on overflow it would produce 16-bit infinite value.  That's what the old test proved.  Now this CL changed the behaviour of parse overflow on 16-bit values so it's the same as 32-bit and 64-bit values: you get a failbit set on the stream and the value stored is the nearest finite value.  That's what operator>> does targeting float values, in C++11.

dekimir:
+2

dneto0:
Squashed and pushed into master as c7a9a909a82

PR #105, opened 2016-02-14
Configure gmock locally only if not already configured.

dneto0:
Allows an enclosing project to have already configured gmock, even
if the developer has placed the googletest sources under this
project's external/googletest directory.

Describe how to enable the tests in the two essential scenarios:
where SPIR-V Tools is standalone, or where SPIR-V Tools is configured
as a part of a larger project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,13 @@ We intend to maintain a linear history on the GitHub `master` branch.
 The project contains a number of tests, used to drive development
 and ensure correctness.  The tests are written using the
 [googletest](https://github.com/google/googletest) framework.  The `googletest`
-source is not provided with this project.  Download the `googletest` source
-into the `<spirv-dir>/external/googletest` directory before configuring
-and building the project.
+source is not provided with this project.  There are two ways to enable
+tests:
+* If SPIR-V Tools is configured as part of an enclosing project, then the
+  enclosing project should configure `googletest` before configuring SPIR-V Tools.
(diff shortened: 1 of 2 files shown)


external/googletest
```diff
@@ -0,0 +1 @@
+../../../third_party/googletest/
```
dekimir: What does this do?

dneto0: This 'external/googletest' file was incorrectly added to the comit.
On my machine it's a symlink with those contents.

README.md
```diff
@@ -66,9 +66,13 @@ We intend to maintain a linear history on the GitHub `master` branch.
 The project contains a number of tests, used to drive development
 and ensure correctness.  The tests are written using the
 [googletest](https://github.com/google/googletest) framework.  The `googletest`
-source is not provided with this project.  Download the `googletest` source
-into the `<spirv-dir>/external/googletest` directory before configuring
-and building the project.
+source is not provided with this project.  There are two ways to enable
+tests:
+* If SPIR-V Tools is configured as part of an enclosing project, then the
+  enclosing project should configure `googletest` before configuring SPIR-V Tools.
```
dekimir: Well, the enclosing project doesn't _have_ to configure `googletest`; it's just that SPIR-V Tools will find it automatically if it's already configured and won't look for it under `external`.

dekimir:
+2 despite nits.

dneto0:
Pushed into master as 31c087036781

PR #107, opened 2016-02-16
Update capabilities for 1.0 Rev3

dneto0:


dneto0:
Rebased against master

source/operand.cpp
```diff
@@ -1011,102 +1011,73 @@ static const spv_operand_desc_t capabilityInfoEntries[] = {
...
 // If the source arrays were std::array, then we could have used
 // std::array::size.
-#define ARRAY_SIZE(A) (static_cast<uint32_t>(sizeof(A)/sizeof(A[0])))
+#define ARRAY_SIZE(A) (static_cast<uint32_t>(sizeof(A) / sizeof(A[0])))
 
 static const spv_operand_desc_group_t opcodeEntryTypes[] = {
-  // TODO(dneto): Reformat this table.
-    {SPV_OPERAND_TYPE_SOURCE_LANGUAGE,
-     ARRAY_SIZE(sourceLanguageEntries),
+    // TODO(dneto): Reformat this table.
```
dekimir: Is this now done?

test/Validate.Capability.cpp
```diff
@@ -514,43 +516,26 @@ make_pair("OpDecorate %intt BuiltIn PatchVertices\n",             TessellationDe
...
 make_pair("OpDecorate %intt BuiltIn VertexIndex\n",               ShaderDependencies()),
 make_pair("OpDecorate %intt BuiltIn InstanceIndex\n",             ShaderDependencies()),
-// Though the remaining builtins don't require Shader, the BuiltIn keyword
-// itself currently does require it.
-make_pair("OpDecorate %intt BuiltIn NumWorkgroups\n",             ShaderDependencies()),
-make_pair("OpDecorate %intt BuiltIn WorkgroupSize\n",             ShaderDependencies()),
-make_pair("OpDecorate %intt BuiltIn WorkgroupId\n",               ShaderDependencies()),
-make_pair("OpDecorate %intt BuiltIn LocalInvocationId\n",         ShaderDependencies()),
-make_pair("OpDecorate %intt BuiltIn GlobalInvocationId\n",        ShaderDependencies()),
-make_pair("OpDecorate %intt BuiltIn LocalInvocationIndex",        ShaderDependencies()),
```
dekimir: The above six cases disappeared, but I would have expected them to stay with `AllDependencies()` -- what am I missing?

dekimir:
-1 for deleted tests (though I could be convinced that's OK after a discussion).

test/Validate.Capability.cpp
dneto0: I had the sense backward.  You're right, we should test those with AllCapabilities().

dneto0:
All feedback addressed:
- re-added tests for BuiltIns without a capability dependency
- fix LocalInvocationIndex: It does not depend on a capability.

dekimir:
+2

dneto0:
Squashed and pushed into master as  dd8a661b

Issue #104, opened 2016-02-13
Float parsing fails tests on Windows.


dneto0:
Somewhat related to #45 . Fixes for that caused me to change the parser (to work around C++ library bugs on Mac).  I also added tests to validate expected behaviour when parsing overflow numbers and non-numbers.

But those tests are now failing on Windows. 
See AppVeyor results from github from testing github.com/google/shaderc
https://ci.appveyor.com/project/dneto0/shaderc/build/288/job/et8t59ppbyoe74n7
It appears that std::operator>>(float&) is broken on edge cases on that MSVC compiler and runtime.

```
[ RUN      ] FloatParse/ParseNormalFloatTest.Samples/1
C:\projects\shaderc\third_party\spirv-tools\test\HexFloat.cpp(1033): error: Value of: parsed_value.value()
Expected: is equal to 0
  Actual: -0x1p-149
 literal: abc negate: true
[  FAILED  ] FloatParse/ParseNormalFloatTest.Samples/1, where GetParam() = 32-byte object <61-62 63-00 6F-46 33-32 2F-48 65-78 46-6C 6F-61 03-00 00-00 0F-00 00-00 01-00 65-73 00-00 00-00> (0 ms)

...

 RUN      ] FloatParse/ParseNormalFloatTest.Samples/14
C:\projects\shaderc\third_party\spirv-tools\test\HexFloat.cpp(1033): error: Value of: parsed_value.value()
Expected: is equal to 3.40282e+38
  Actual: 0
 literal: 1e40 negate: false
[  FAILED  ] FloatParse/ParseNormalFloatTest.Samples/14, where GetParam() = 32-byte object <31-65 34-30 00-65 73-2F 31-33 65-78 46-6C 6F-61 04-00 00-00 0F-00 00-00 00-00 65-73 FF-FF 7F-7F> (0 ms)
[ RUN      ] FloatParse/ParseNormalFloatTest.Samples/15
C:\projects\shaderc\third_party\spirv-tools\test\HexFloat.cpp(1033): error: Value of: parsed_value.value()
Expected: is equal to -3.40282e+38
  Actual: -0x1p-149
 literal: 1e40 negate: true 
[...]

PR #108, opened 2016-02-16
Don't promise particular values when float parsing fails.

dneto0:
This should address problems on Windows:
https://github.com/KhronosGroup/SPIRV-Tools/issues/104


AWoloszyn:
+2

PR #109, opened 2016-02-16
Use syntax tables for SPIR-V 1.0 Rev3

dneto0:
- The SPIR-V spec generator has changed how it represents optional
  operands.  Now it tracks a separate boolean flag indicating optionality.
  However, SPIRV-Tools still wants to represent both operand class
  and optionality in the same enums space (SPV_OPERAND_TYPE_*).
  So there's extra work in the patch.

- In the spec generator, OperandImage is now OperandImageOperands.
  This affects enum translation in opcode.cpp.

- In the spec generator, image operands are explicitly followed by
  Id, and VariableIds.  However, SPIRV-Tools uses the bits set
  in the image operand bitmask to control the number and meaning
  of the Ids that follow.  So in writing the opcode.inc syntax
  table, drop all operands after OperandImageOperands.

- Some enums are now more explicitly represented in the generated
  opcode.inc:
    - AccessQualifier (e.g. on OpTypeImage), in both required and
      optional flavours.
    - MemoryAccess (e.g. on loads and stores)

- Add SPV_OPERAND_TYPE_OPTIONAL_ACCESS_QUALIFIER

- Add tests for the optional AccessQualifier operand on OpTypeImage.

- Update the AccessQualifier test for OpTypeImage so it's a round
  trip test through the disassembler as well.


source/syntax_tables.patch
```diff
@@ -137,8 +141,6 @@ index 0000000..85bca89
 +  const OperandParameters& operands = inst.operands;
 +  for (int i = 0; i < operands.getNum() ; i++) {
 +    switch (operands.getClass(i)) {
-+      case spv::OperandOptionalId:
-+      case spv::OperandOptionalImage:
```
dekimir: I don't understand why these two are removed but `OptionalLiteral*` cases remain here.

dekimir: Is it that `OperandOptionalId` and `OperandOptionalImage` no longer exist?

dneto0: Yes, they were removed from the doc.h in the SPIR-V spec generator.
So they didn't appear in the grammar and would be dead enums if not removed here.

dneto0: Actually, a *compilation failure* if not removed here.

dneto0:
All questions answered so far.

PR #111, opened 2016-02-17
Remove remaining float overflow value tests

dneto0:
Should fix tests on Windows.


dekimir:
+2

dneto0:
Rebased and pushed into master as  2732f57ed2

PR #112, opened 2016-02-17
Test OpImageSparseRead

dneto0:
Also tweak the image operands test with OpImageFetch so the input assembly
more closely matches the expected output.  But don't make it a round
trip test since we want to check the assembler's ability to read the
image operand mask with out-of-order mask names.

The OpImageSparseRead assembly test is also a round trip disassembler
test.


dekimir:
+2

dneto0:
Rebased and pushed into master as  4ddb431b

PR #113, opened 2016-02-17
Fixed internal compiler error in gcc-4.9.

AWoloszyn:
This showed up in mips and mips64 builds. A combination of templates
and the error reporting were causing gcc to crash. This splits up the
functionality in a way that now successfully compiles.


dneto0:
Testing locally.

dneto0:
+2
Rebased and pushed into master as 4ddb431b16

PR #116, opened 2016-02-17
Rearrange headers

dneto0:
Now we have public headers arranged as follows:
$SPIRV_TOOLS_ROOT/include/spirv-tools/libspirv.h
$SPIRV_TOOLS_ROOT/include/spirv/spirv.h
$SPIRV_TOOLS_ROOT/include/spirv/GLSL.std.450.h
$SPIRV_TOOLS_ROOT/include/spirv/OpenCL.std.h

A project should use -I$SPIRV_TOOLS_ROOT/include
and then #include "spirv-tools/libspirv.h"

The headers from the SPIR-V Registry can be accessed as "spirv/spirv."
for example.

The install target should also install the headers from the SPIR-V
Registry.  The libspirv.h header is broken otherwise.

The SPIRV-Tools library depends on the headers from the SPIR-V Registry.

The util/bitutils.h and util/hex_float.h are pulled into the internal
source tree.  Those are not part of the public API to SPIRV-Tools.


AWoloszyn:
+1 Looks good to me, will let @dekimir take a look.

PR #119, opened 2016-02-18
Fix the source organization section in README.md

dneto0:
It now reflects the new header file organization.
It also tells API clients what to add to the search
include path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,13 @@ We intend to maintain a linear history on the GitHub `master` branch.
 
 ### Source code organization
 
-* `external/headers`: Standard SPIR-V header files used by the implementation,
-   from the SPIR-V Registry
 * `external/googletest`: Intended location for the
-  [googletest](https://github.com/google/googletest) sources, not provided.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,11 +51,13 @@ We intend to maintain a linear history on the GitHub `master` branch.
...
 
-* `external/headers`: Standard SPIR-V header files used by the implementation,
-   from the SPIR-V Registry
 * `external/googletest`: Intended location for the
-  [googletest](https://github.com/google/googletest) sources, not provided.
-* `include/libspirv/libspirv.h`: C API public interface
+  [googletest](https://github.com/google/googletest) sources, not provided
+* `include/` : API clients should add this directory to the include search
+  path
+* `include/spirv-tools/libspirv.h`: C API public interface
```
dekimir: Why single C out?  Perhaps we should describe the whole directory.

dekimir: Hm, looks like that file is the only content of the directory.  Meh...

PR #121, opened 2016-02-18
Support using stdin to supply input in as and val.

antiagainst:
This allows getting output from other utilities using pipe.


dekimir:
+2

PR #124, opened 2016-02-18
Remove unnecessary null pointer checks

### LINKED ISSUE #110: Remove unnecessary null pointer checks
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for NULL before delete p?") like the following.
* [spvBinaryDestroy](https://github.com/KhronosGroup/SPIRV-Tools/blob/7ffd8ffeb2abb2314dd2666353ae4efed276e8fc/source/binary.cpp#L807 "spvBinaryDestroy function")
* [spvTextDestroy](https://github.com/KhronosGroup/SPIRV-Tools/blob/3b69d05086b991eefd585a6032659194e5c63000/source/text.cpp#L757 "spvTextDestroy function")

1ace:
Fixes #110

### DIFF
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -815,8 +815,6 @@ spv_result_t spvBinaryParse(const spv_const_context context, void* user_data,
 // that a spv_binary_t value is created.
 void spvBinaryDestroy(spv_binary binary) {
   if (!binary) return;
-  if (binary->code) {
-    delete[] binary->code;
-  }
+  delete[] binary->code;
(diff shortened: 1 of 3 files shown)


elfring:
Thanks for [your small source code improvement](https://github.com/1ace/SPIRV-Tools/commit/eb6ae976c6742849c80bea1726890fed53573f5e "Deletion of a few unnecessary checks").

PR #125, opened 2016-02-19
More modifications to support pipe in command line tools

antiagainst:
* Support outputting to stdout in `as`.
* Fix command line option errors in `val`.
    * There is no `-capability`.
    * `-rules` is missing.
    * `-id` by default is turned off.

### DIFF
--- a/tools/as/as.cpp
+++ b/tools/as/as.cpp
@@ -103,7 +103,7 @@ int main(int argc, char** argv) {
     char buf[1024];
     while (size_t len = fread(buf, 1, sizeof(buf), fp))
       contents.insert(contents.end(), buf, buf + len);
-    fclose(fp);
+    if (use_file) fclose(fp);
   } else {
     fprintf(stderr, "error: file does not exist '%s'\n", inFile);
     return 1;
@@ -129,7 +129,7 @@ int main(int argc, char** argv) {
       fprintf(stderr, "error: could not write to file '%s'\n", outFile);
       return 1;
     }
-    fclose(fp);
+    if (!use_stdout) fclose(fp);
   } else {
     fprintf(stderr, "error: could not open file '%s'\n", outFile);
     return 1;
--- a/tools/as/as.cpp
+++ b/tools/as/as.cpp
@@ -46,7 +46,7 @@ Options:
 
   -h              Print this help.
 
-  -o <filename>   Set the output filename.
+  -o <filename>   Set the output filename. Use '-' to mean stdout.
 )",
       argv0, argv0);
 }
@@ -121,13 +121,15 @@ int main(int argc, char** argv) {
     return error;
   }
 
-  if (FILE* fp = fopen(outFile, "wb")) {
+  const bool use_stdout = outFile[0] == '-' && outFile[1] == 0;
+  if (FILE* fp = (use_stdout ? stdout : fopen(outFile, "wb"))) {
     size_t written =
         fwrite(binary->code, sizeof(uint32_t), (size_t)binary->wordCount, fp);
     if (binary->wordCount != written) {
       fprintf(stderr, "error: could not write to file '%s'\n", outFile);
       return 1;
     }
+    fclose(fp);
   } else {
     fprintf(stderr, "error: could not open file '%s'\n", outFile);
     return 1;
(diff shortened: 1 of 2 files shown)


antiagainst:
Now we can do something like

```bash
echo "%4 = OpTypeFunction %3" | /path/to/spirv-as -o - | /path/to/spirv-val -id
```

tools/as/as.cpp
```diff
@@ -121,7 +121,8 @@ int main(int argc, char** argv) {
     return error;
   }
 
-  if (FILE* fp = fopen(outFile, "wb")) {
+  const bool use_stdout = *outFile == '-' && *(outFile + 1) == 0;
+  if (FILE* fp = (use_stdout ? stdout : fopen(outFile, "wb"))) {
```
dneto0: I guess we always leaked the output handle.  Yikes.

tools/as/as.cpp
```diff
@@ -121,7 +121,8 @@ int main(int argc, char** argv) {
     return error;
   }
 
-  if (FILE* fp = fopen(outFile, "wb")) {
+  const bool use_stdout = *outFile == '-' && *(outFile + 1) == 0;
```
dneto0: Alternate could have written outFile[0] and outFile[1].
But this is ok too.  Doesn't slow me down.

tools/val/val.cpp
```diff
@@ -55,19 +55,20 @@ int main(int argc, char** argv) {
...
-      } else if (!strcmp("id", argv[argi] + 1)) {
+      } else if (!strcmp("id", cur_arg + 1)) {
         options |= SPV_VALIDATE_ID_BIT;
-      } else if (!strcmp("rules", argv[argi] + 1)) {
+      } else if (!strcmp("layout", cur_arg + 1)) {
+        options |= SPV_VALIDATE_LAYOUT_BIT;
+      } else if (!strcmp("rules", cur_arg + 1)) {
         options |= SPV_VALIDATE_RULES_BIT;
-      } else if (0 == *(argv[argi] + 1)) {
+      } else if (0 == *(cur_arg + 1)) {
```
dneto0: I'd find cur_arg[1] easier to read.

dneto0:
I'm actually surprised there isn't an -all option on the validator executable.  It should be easy to say "get me everything!"

+1 
Suggest @umar456 take a second look.

dekimir:
Tee-hee, I have a `-all` implemented on a stash that I pull out when I need to debug. :)  I've never sent it up because it'll be better to actually fix #89 .

umar456:
I would remove the -[capability, rules, id] altogether from the command line tool. The only use case I see for this is in unit testing and we are already performing that through the `libspirv` API.

tools/as/as.cpp
antiagainst: Good catch! Didn't check enough lines to find that. Fixed.

tools/as/as.cpp
antiagainst: Done.

tools/val/val.cpp
antiagainst: Done.

antiagainst:
@dneto0 @dekimir @umar456:  There actually is a `ALL` bit in the enumeration. Added `-all`. I'll try to remove all these bits and let `spirv-val` to validate all by default as the next step.

tools/as/as.cpp
```diff
@@ -121,13 +121,15 @@ int main(int argc, char** argv) {
     return error;
   }
 
-  if (FILE* fp = fopen(outFile, "wb")) {
+  const bool use_stdout = outFile[0] == '-' && outFile[1] == 0;
+  if (FILE* fp = (use_stdout ? stdout : fopen(outFile, "wb"))) {
```
umar456: I would use a unique pointer for something like this.
``` 
auto file = unique_ptr<FILE, int (*)(FILE*)>(use_stdout ? stdout : fopen(outFile, "wb"), fclose);
```

umar456:
+2 Minor issue but since we are not using exceptions its not important.

tools/as/as.cpp
umar456: Ha. That will actually close stdout when it goes out of scope.

antiagainst: Good catch! I'm not aware of the problem previously. TIL. Thanks!

antiagainst:
@umar456 Does the code look good now? :)

umar456:
LGTM :+1:

PR #128, opened 2016-02-22
Remove spv_validate_options_t and change tests accrodingly.

antiagainst:
Users always want to run all the checks. The `spv_validate_options_t`
mechanism, which provides little benefits to users, complicates the
internal implementation and also makes the tests exercise different
paths as users do.

Right now the tests are more like integration tests instead of
unit tests, which should be our next refactoring aim.

Addresses #89.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -24,7 +24,7 @@
 // TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 // MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
-#ifndef SPIRV_TOOLS_LIBSPIRV_H_      
+#ifndef SPIRV_TOOLS_LIBSPIRV_H_
 #define SPIRV_TOOLS_LIBSPIRV_H_
 
 #include "spirv/GLSL.std.450.h"
@@ -48,7 +48,7 @@ extern "C" {
 
 // Helpers
 
-#define spvIsInBitfield(value, bitfield) ((value) == ((value) & bitfield))
+#define spvIsInBitfield(value, bitfield) ((value) == ((value)&bitfield))
 
 #define SPV_BIT(shift) (1 << (shift))
 
@@ -63,7 +63,7 @@ typedef enum spv_result_t {
   SPV_END_OF_STREAM = 2,
   SPV_WARNING = 3,
   SPV_FAILED_MATCH = 4,
-  SPV_REQUESTED_TERMINATION = 5, // Success, but signals early termination.
+  SPV_REQUESTED_TERMINATION = 5,  // Success, but signals early termination.
   SPV_ERROR_INTERNAL = -1,
   SPV_ERROR_OUT_OF_MEMORY = -2,
   SPV_ERROR_INVALID_POINTER = -3,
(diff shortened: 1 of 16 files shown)


dekimir:
+1

@umar456 PTAL

include/spirv-tools/libspirv.h
```diff
@@ -388,10 +375,9 @@ spv_result_t spvBinaryToText(const spv_const_context context,
 // pointer.
 void spvBinaryDestroy(spv_binary binary);
 
-// Validates a SPIR-V binary for correctness. The options parameter is a bit
-// field of spv_validation_options_t.
+// Validates a SPIR-V binary for correctness.
 spv_result_t spvValidate(const spv_const_context context,
-                         const spv_const_binary binary, const uint32_t options,
+                         const spv_const_binary binary,
```
dneto0: This breaks existing clients.  I'm ok with that since the validator is known to be work-in-progress, and clients who want compatibility should be pulling in known commits or tags.

source/validate_instruction.cpp
```diff
@@ -119,7 +119,6 @@ spv_result_t CapCheck(ValidationState_t& _,
 // clang-format off
 spv_result_t InstructionPass(ValidationState_t& _,
                              const spv_parsed_instruction_t* inst) {
-  if (_.is_enabled(SPV_VALIDATE_INSTRUCTION_BIT)) {
     if (inst->opcode == SpvOpCapability)
```
dneto0: I would expect this to be reformatted to the left since an indentation level has disappeared.

test/Validate.Capability.cpp
```diff
@@ -284,18 +283,48 @@ const vector<string>& SampledBufferDependencies() {
...
                             Values(
-make_pair("OpEntryPoint Vertex %func \"shader\" %var1 %var2\n",                 ShaderDependencies()),
-make_pair("OpEntryPoint TessellationControl %func \"shader\" %var1 %var2\n",    TessellationDependencies()),
-make_pair("OpEntryPoint TessellationEvaluation %func \"shader\" %var1 %var2\n", TessellationDependencies()),
-make_pair("OpEntryPoint Geometry %func \"shader\" %var1 %var2\n",               GeometryDependencies()),
-make_pair("OpEntryPoint Fragment %func \"shader\" %var1 %var2\n",               ShaderDependencies()),
-make_pair("OpEntryPoint GLCompute %func \"shader\" %var1 %var2\n",              ShaderDependencies()),
-make_pair("OpEntryPoint Kernel %func \"shader\" %var1 %var2\n",                 KernelDependencies())
-                                                           )));
+make_pair(string(kOpenCLMemoryModel) +
```
dneto0: It looks weird to have the OpenCL memory model applied to graphics shaders (vertex through GLCompute), and the GLSL450 memory model applied to a Kernel entry point.
Is that intentional?  Given a choice, I'd flip them around.

test/Validate.Capability.cpp
```diff
@@ -325,101 +354,226 @@ INSTANTIATE_TEST_CASE_P(ExecutionMode, ValidateCapability,
...
-make_pair("OpExecutionMode %func InputTrianglesAdjacency", GeometryDependencies()),
-make_pair("OpExecutionMode %func Quads",                   TessellationDependencies()),
-make_pair("OpExecutionMode %func Isolines",                TessellationDependencies()),
-make_pair("OpExecutionMode %func OutputVertices 42",       GeometryTessellationDependencies()),
-make_pair("OpExecutionMode %func OutputPoints",            GeometryDependencies()),
-make_pair("OpExecutionMode %func OutputLineStrip",         GeometryDependencies()),
-make_pair("OpExecutionMode %func OutputTriangleStrip",     GeometryDependencies()),
-make_pair("OpExecutionMode %func VecTypeHint 2",           KernelDependencies()),
-make_pair("OpExecutionMode %func ContractionOff",          KernelDependencies())
+make_pair(string(kOpenCLMemoryModel) +
```
dneto0: Similarly here.

test/Validate.Capability.cpp
dneto0: @dekimir set me straight.  The kOpenCLMemoryModel string already includes the Kernel capability.  So you have to do the opposite in order to exercise the test case.

Resolved.

test/Validate.Capability.cpp
dneto0: Resolved, as above.

test/Validate.SSA.cpp
```diff
@@ -191,7 +199,7 @@ TEST_F(Validate, ForwardDecorateInvalidIDBad) {
 
 TEST_F(Validate, ForwardMemberDecorateGood) {
   char str[] = R"(
-           OpCapability Matrix
+           OpCapability Shader
```
dneto0: Indeed!   Every module needs an OpMemoryModel instruction, and  Logical requires Shader capability.

Interesting. Good.

dneto0:
+1
This change makes it easier to use the validator correctly.

Optional: Fix potential formatting problem in one place.

Still would like @umar456 to review.

source/validate_instruction.cpp
antiagainst: Done. Previously it was wrapped by `clang-format : off`. ;)

AWoloszyn:
This LGTM! +2.

dneto0:
Rebased, resolved conflict with the --version patch, and pushed into master as b6ccd0d89

PR #129, opened 2016-02-22
Implement --version for spirv-tools

Qining:
'spirv-as --version', 'spirv-dis --version' and 'spirv-val --version'
will display version information.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,6 @@ spvtools_default_compile_options(${SPIRV_TOOLS})
 target_include_directories(${SPIRV_TOOLS} PUBLIC
   ${CMAKE_CURRENT_SOURCE_DIR}/include
   ${CMAKE_CURRENT_BINARY_DIR})
-add_dependencies(${SPIRV_TOOLS} spirv-tools-build-version)
 
 # Defaults to OFF if the user didn't set it.
 option(SPIRV_SKIP_EXECUTABLES
@@ -170,14 +169,21 @@ if (NOT ${SPIRV_SKIP_EXECUTABLES})
   add_executable(spirv-as ${CMAKE_CURRENT_SOURCE_DIR}/tools/as/as.cpp)
   spvtools_default_compile_options(spirv-as)
   target_link_libraries(spirv-as PRIVATE ${SPIRV_TOOLS})
+  target_include_directories(spirv-as PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+  add_dependencies(${SPIRV_TOOLS} spirv-tools-build-version)
+
 
   add_executable(spirv-dis ${CMAKE_CURRENT_SOURCE_DIR}/tools/dis/dis.cpp)
   spvtools_default_compile_options(spirv-dis)
   target_link_libraries(spirv-dis PRIVATE ${SPIRV_TOOLS})
+  target_include_directories(spirv-dis PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+  add_dependencies(${SPIRV_TOOLS} spirv-tools-build-version)
 
   add_executable(spirv-val ${CMAKE_CURRENT_SOURCE_DIR}/tools/val/val.cpp)
   spvtools_default_compile_options(spirv-val)
   target_link_libraries(spirv-val PRIVATE ${SPIRV_TOOLS})
+  target_include_directories(spirv-val PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+  add_dependencies(${SPIRV_TOOLS} spirv-tools-build-version)
 
   # Find gmock if we can. If it's not already configured, then try finding
   # it in external/googletest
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -148,7 +154,10 @@ set(SPIRV_SOURCES
 
 add_library(${SPIRV_TOOLS} ${SPIRV_SOURCES})
 spvtools_default_compile_options(${SPIRV_TOOLS})
-target_include_directories(${SPIRV_TOOLS} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
+target_include_directories(${SPIRV_TOOLS} PUBLIC
+  ${CMAKE_CURRENT_SOURCE_DIR}/include
+  ${CMAKE_CURRENT_BINARY_DIR})
+add_dependencies(${SPIRV_TOOLS} spirv-tools-build-version)
```
dekimir: Nit: this dependency should go on the executables, not the library.  I presume there's no way to ask the library for build-version?

tools/as/as.cpp
```diff
@@ -47,10 +47,15 @@ is used.
   -h              Print this help.
 
   -o <filename>   Set the output filename. Use '-' to mean stdout.
+  --version       Display assembler verision information.
```
dekimir: typo: "verision"

tools/as/as.cpp
```diff
@@ -79,6 +84,15 @@ int main(int argc, char** argv) {
             return 1;
           }
         } break;
+        case '-': {
+          // Long options
+          if (0 == strcmp(argv[argi], "--version")) {
+            printf("%s\r\n", kBuildVersion);
```
dekimir: Why `\r`?

tools/dis/dis.cpp
```diff
@@ -43,6 +43,7 @@ or if the filename is "-", then the binary is read from standard input.
 Options:
 
   -h, --help      Print this help.
+  --version       Display disassebler version information.
```
dekimir: Typo: "disassebler"

tools/val/val.cpp
```diff
@@ -47,10 +47,15 @@ or if the filename is "-", then the binary is read from standard input.
   -id        Perform id validation (default OFF)
   -layout    Perform layout validation (default OFF)
   -rules     Perform rules validation (default OFF)
+  --version  Display validator verision information
```
dekimir: typo: "verision"

dekimir:
-1 for typos

tools/as/as.cpp
Qining: I remember on Windows '\r\n' is the newline, but actually I'm not sure whether we need to do this here.

Another thing that I feel weird is that why are we using 'printf' instead of 'std::count' in these spirv-tools?

dekimir: Nope, `\n` is the platform-specific newline everywhere.  You only need `\r` for cross-platform reading and writing (eg, reading Windows text files on Unix).

I agree about `printf()`, but that's a separate story and can wait for now.

CMakeLists.txt
Qining: Moved dependencies() and the include directory down to executables.

Maybe I should leave the include directory here with INTERFACE scope, but I feel that makes this thing more confusing.

tools/as/as.cpp
Qining: Done.

tools/as/as.cpp
Qining: Changed to '\n', and many thanks for the clarification!

tools/dis/dis.cpp
Qining: Done

tools/val/val.cpp
Qining: Done

CMakeLists.txt
```diff
@@ -148,7 +154,9 @@ set(SPIRV_SOURCES
 
 add_library(${SPIRV_TOOLS} ${SPIRV_SOURCES})
 spvtools_default_compile_options(${SPIRV_TOOLS})
-target_include_directories(${SPIRV_TOOLS} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
+target_include_directories(${SPIRV_TOOLS} PUBLIC
```
dekimir: Is this still necessary?

tools/as/as.cpp
dekimir: I'm afraid I don't see the change yet.  Also make sure to remove all `\r`s added below.

CMakeLists.txt
Qining: No...I should have it removed already.

dekimir:
Getting better, just need to exorcise all the `\r`s and possibly remove `target_include_directories` for the library.

dekimir:
+2

dneto0:
Rebased, squashed, and pushed into master as 4e797f823

Issue #126, opened 2016-02-19
Parsing problem when file doesn't end in EOL


dekimir:
Running `spirv-as` on the attached file yields:

```
error: 21: 1: Invalid Opcode name 'OpCapability Shader
OpMemoryModel Logical GLSL450

%i32 = OpTypeInt 32 1
%f32 = OpTypeFloat 32
%v4i32 = OpTypeVector %i32 4
%timg = OpTypeImage %i32 2D 0 0 0 0 Unknown
%pimg = OpTypePointer Input %timg
%tfun = OpTypeFunction %i32

%vimg = OpVariable %pimg Input
%coord = OpConstant %i32 0
%bias = OpConstant %f32 0.

%main = OpFunction %i32 None %tfun
%lbl = OpLabel
%img = OpLoad %timg %vimg
%r1 = OpImageFetch %v4i32 %img %coord Offset|Bias %bias %bias
;%r2 = OpImageSampleExplicitLod %v4i32 %sample %coord Bias 0.2
OpReturnValue %coord
OpFunctionEndѷ'
```

IOW, it thinks the whole program is an opcode name. :scream_cat: 

[shader.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/138518/shader.txt)

dneto0:
I'm surprised it doesn't stop at the first space after OpCapbility.  ?!  Hmm.

dekimir:
Forgot to add: if I add a newline at the end of the file, it assembles just fine.

antiagainst:
Done via #131.

dneto0:
I verified it against the shader.txt file originally submitted, but with the trailing newline removed.  (Somehow it crept in?)

antiagainst:
@dneto0 Yeah, the attached file has EOL before EOF.

@dekimir kindly shows me how to use the power of `echo -n`.

PR #131, opened 2016-02-22
getWord() should not go beyond the end of the source text.

antiagainst:
This addresses Bug #126 where EOL is missing at the end of
source file.


dneto0:
+2

PR #133, opened 2016-02-24
[WIP] Add appveyor.xml to build on Windows.

sheredom:
I'm testing Appveyor integration (for Windows builds).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,6 +21,6 @@ configuration:
 build_script:
   - md build
   - cd build
-  - if "%PLATFORM%"=="x64" cmake -G "%VSVERSION% Win64" ../test
-  - if "%PLATFORM%"=="Win32" cmake -G "%VSVERSION%" ../test
+  - if "%PLATFORM%"=="x64" cmake -G "%VSVERSION% Win64" ..
+  - if "%PLATFORM%"=="Win32" cmake -G "%VSVERSION%" ..
   - msbuild /m /p:Configuration="%CONFIGURATION%" /p:Platform="%PLATFORM%" spirv-tools.sln
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,8 +18,10 @@ configuration:
   - Release
 
 build_script:
+  - git clone https://github.com/google/googletest.git external/googletest
   - md build
   - cd build
   - if "%PLATFORM%"=="x64" cmake -G "%VSVERSION% Win64" ..
   - if "%PLATFORM%"=="Win32" cmake -G "%VSVERSION%" ..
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -0,0 +1,27 @@
+version: '{build}'
+
+skip_tags: true
+
+install: []
+
+environment:
+  matrix:
+    - VSVERSION: Visual Studio 12
```
dneto0: I'd prefer to stay away from older MSVC compilers.  Their C++11 support is sometimes lacking.

appveyor.yml
```diff
@@ -0,0 +1,27 @@
...
+  - Win32
+  - x64
+
+configuration:
+  - Debug
+  - Release
+
+build_script:
+  - git clone https://github.com/google/googletest.git external/googletest
+  - md build
```
dneto0: Take a look at https://github.com/google/shaderc/blob/master/.appveyor.yml 
It uses cmake --build, which potentially frees up a dependency.  I also find it easier to read. :-)
Also it tests via ctest.  That will let us add other test executables without having to update this config.

source/util/bitutils.h
```diff
@@ -27,6 +27,8 @@
 #ifndef LIBSPIRV_UTIL_BITUTILS_H_
 #define LIBSPIRV_UTIL_BITUTILS_H_
 
+#include <stdint.h>
```
dneto0: I guess this is a symptom of VS 12?
In C++ code like this we'd prefer #include <cstdint>

dneto0:
See https://github.com/google/shaderc/blob/master/.appveyor.yml for another AppVeyor config that we've been using.  I prefer the use of cmake to execute the build, and ctest to run the tests.

sheredom:
Tbh now that I've seen the shaderc one? I'd be totally happy bringing something like that over. I just wanted some form of Windows testing in the project! :)

appveyor.yml
sheredom: This is the VS I personally use - 2015 is far too temperamental for my liking!

source/util/bitutils.h
sheredom: Yup, VS12 issue.

antiagainst:
@sheredom This is kinda tested under windows because it is pulled in as a dependency for shaderc. But, yeah, definitely worth to do that in its own repo.

But AFAIK, there are some permission issue with appveyor (it requires to access private repo, if I recall that correctly) which may need to be allowed by the org (KhronosGroup) first.

sheredom:
Righto, I'm gonna reject my own MR then! :)

PR #134, opened 2016-02-24
Actually find the python executable before use.

AWoloszyn:
Also change PYTHON_EXE to PYTHON_EXECUTABLE to match what
the cmake documentation says.


antiagainst:
+2.

PR #137, opened 2016-02-25
Fix issues reported by cppcheck.

antiagainst:
Remove code not being used, add explicit to constructors, and
add missing fields in constructors.


source/binary.cpp
```diff
@@ -62,47 +62,6 @@ spv_result_t spvBinaryHeaderGet(const spv_const_binary binary,
   return SPV_SUCCESS;
 }
 
-// TODO(dneto): This API is not powerful enough in the case that the
-// number and type of operands are not known until partway through parsing
-// the operation.  This happens when enum operands might have different number
-// of operands, or with extended instructions.
-spv_operand_type_t spvBinaryOperandInfo(const uint32_t word,
```
dneto0: I'm glad this API is going.  But we should also remove the code in validate.cpp between an #if 0 / #endif pair that references it, as well the prototype in binary.h

source/opcode.cpp
```diff
@@ -485,16 +478,6 @@ int32_t spvOpcodeIsPointer(const SpvOp opcode) {
   }
 }
 
-int32_t spvInstructionIsInBasicBlock(const spv_instruction_t* pFirstInst,
```
dneto0: Should also remove the prototype for this in opcode.h

source/text_handler.h
```diff
@@ -263,10 +263,9 @@ class AssemblyContext {
     // It also allows octal input, but we don't care about that case.
     text_stream >> std::setbase(0);
     text_stream >> *value_pointer;
-    bool ok = true;
 
     // We should have read something.
-    ok = (text[0] != 0) && !text_stream.bad();
+    bool ok = (text[0] != 0) && !text_stream.bad();
```
dneto0: Something complained about this?  Weird.  I think any decent front end would collapse this.  But ok.

dneto0:
Please delete even more code.
-1

source/binary.cpp
antiagainst: Done.

source/opcode.cpp
antiagainst: Done.

source/text_handler.h
antiagainst: Right. Kinda wired for me. But whatever.

dekimir: Not weird at all -- in fact, the style guide [requires](https://google.github.io/styleguide/cppguide.html#Local_Variables) it.

dekimir:
+1

dneto0:
+2

PR #139, opened 2016-03-01
Make libspirv.h more C friendly.

### LINKED ISSUE #138: "struct spv_context_t" misses a typedef for C compilation
Including `libspirv.h` in a C project causes an error at this line : 
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/include/spirv-tools/libspirv.h#L344
because `spv_context_t` isn't a known type, but `struct spv_context_t` is.

Since libspirv is intended as a C API, this should be fixed by adding
`typedef struct spv_context_t spv_context_t;` just here : 
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/include/spirv-tools/libspirv.h#L334

dekimir:
In C, `#include <stdbool.h>` so `bool` is recognized.

Add a `typedef` for `spv_context_t`.

Fixes #138.


include/spirv-tools/libspirv.h
```diff
@@ -33,6 +33,8 @@
 
 #ifdef __cplusplus
 extern "C" {
+#else
+#  include <stdbool.h>
```
antiagainst: Will this pass clang-format?

dekimir: Um, not so much. ;)  Fixed, thx.

antiagainst:
LGTM

PR #140, opened 2016-03-01
kernel related fixes

a2flo:


dekimir:
Thanks, Florian.  Could you please add a test that fails before this change?

a2flo:
Should this go into Validate.SSA.cpp or somewhere else?
Note though that I don't have anything that generates OpPtrAccessChain yet, but since it's along the same lines as OpInBoundsPtrAccessChain (and OpAccessChain) I added it to spvOpcodeIsPointer as well.

[simple_kernel.spvasm.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/153599/simple_kernel.spvasm.txt)

umar456:
You should put your tests in the [ValidateID.cpp](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/ValidateID.cpp) file.

dekimir:
As @umar456 said, tests for `validate_id.cpp` code go to `ValidateID.cpp`.  The program you posted is quite long and wouldn't make a great test -- I'm sure it can be reduced to a much smaller part to exercise the feature under test.

Regarding `OpPtrAccessChain` test: we encourage contributors to provide test coverage for everything they implement.  That way we all gain confidence that the code works as advertised; more importantly, that ensures that any future changes to the code will not accidentally drop the functionality you worked to provide.  It would be really great if you could develop a test for `OpPtrAccessChain`, too.

a2flo:
Added tests for all of them (needed to modify the IR to actually get a OpPtrAccessChain ...) + added kernel/OpenCL testing support to Validate.cpp.

source/opencl_std_ext_inst.inc
```diff
@@ -53,8 +53,8 @@ ExtInst(minmag, 44, List(OperandId, OperandId))
 ExtInst(modf, 45, List(OperandId, OperandId))
 ExtInst(nan, 46, List(OperandId))
 ExtInst(nextafter, 47, List(OperandId, OperandId))
-ExtInst(pow, 48, List(OperandId, OperandId, OperandId))
-ExtInst(pown, 49, List(OperandId))
+ExtInst(pow, 48, List(OperandId, OperandId))
+ExtInst(pown, 49, List(OperandId, OperandId))
```
dekimir: @dneto0 is this file generated by `syntax_tables.patch`?  If so, then manual modification won't work.

dekimir: Also, @a2flo: why does this seem to differ from `pow` and `pown` specification [here](https://www.khronos.org/registry/spir-v/specs/1.0/OpenCL.ExtendedInstructionSet.100.html)?

test/ExtInst.OpenCL.std.cpp
```diff
@@ -152,8 +152,8 @@ INSTANTIATE_TEST_CASE_P(
         CASE2(Modf, modf),
         CASE1(Nan, nan),
         CASE2(Nextafter, nextafter),
-        CASE3(Pow, pow),
-        CASE1(Pown, pown),
+        CASE2(Pow, pow),
+        CASE2(Pown, pown),
```
dekimir: Seems to violate the [spec](https://www.khronos.org/registry/spir-v/specs/1.0/OpenCL.ExtendedInstructionSet.100.html).

source/validate_id.cpp
```diff
@@ -188,14 +188,18 @@ bool idUsage::isValid<SpvOpEntryPoint>(const spv_instruction_t* inst,
...
   }
-  // TODO: Check the entry point signature is void main(void), may be subject
-  // to change
-  auto entryPointType = usedefs_.FindDef(entryPoint.second.words[4]);
-  if (!entryPointType.first || 3 != entryPointType.second.words.size()) {
-    DIAG(entryPointIndex) << "OpEntryPoint Entry Point <id> '"
-                          << inst->words[entryPointIndex]
-                          << "'s function parameter count is not zero.";
-    return false;
+  // don't check kernel function signatures
```
dekimir: Actually, I don't see why this check exists at all, even for graphics entry points.  I can't find this restriction anywhere in the SPIR-V spec.  @umar456 do you happen to know why an entry point function with parameters would be considered invalid?

test/ValidateID.cpp
```diff
@@ -1476,6 +1515,71 @@ TEST_F(ValidateID, UndefinedIdMemSem) {
...
+      OpDecorate %3 Constant
+      OpDecorate %4 FuncParamAttr NoCapture
+      OpDecorate %3 LinkageAttributes "__spirv_GlobalInvocationId" Import
+ %5 = OpTypeInt 32 0
+ %6 = OpTypeVector %5 3
+ %7 = OpTypePointer UniformConstant %6
+ %3 = OpVariable %7 UniformConstant
+ %8 = OpTypeVoid
+ %9 = OpTypeStruct %5
+%10 = OpTypePointer CrossWorkgroup %9
```
dekimir: Why isn't this an interface in `OpEntryPoint`?

source/validate_id.cpp
dekimir: Found it: https://www.khronos.org/registry/vulkan/specs/1.0/xhtml/vkspec.html#spirvenv-module-validation

dekimir:
I don't understand why you had to change `pow` and `pown` to make the tests work.  The tests never use those functions.  Can you help me understand?

source/validate_id.cpp
umar456: This seems like a Vulkan specific rule. This should not apply to all of SPIR-V modules. Is there a way to check if a module targets Vulkan shaders as opposed to OpenCL or GLSL?

dekimir: @umar456 Yes: it's exactly what this PR is doing. :)

test/ValidateID.cpp
dekimir: Never mind, this is actually in line with the OpenCL spec.

test/ValidateID.cpp
```diff
@@ -71,6 +88,28 @@ const char kGLSL450MemoryModel[] = R"(
   ASSERT_EQ(expected, result);                                                 \
   spvContextDestroy(context);
 
+#define CHECK_KERNEL(str, expected, bitness)                                   \
+  ASSERT_EQ(bitness == 32 || bitness == 64, true);                             \
```
dekimir: `ASSERT_TRUE` would express the contract much better here than `ASSERT_EQ`.

a2flo:
The spec is broken in regards to pow and pown, the bug has been reported at https://www.khronos.org/bugzilla/show_bug.cgi?id=1469
This fixes it to the correct way: both pow and pown taking 2 parameters (x and y). Note that SPIRV-LLVM is actually generating correct code here (2 parameters), hence not validating correctly without this fix.

I have no idea what the purpose of syntax_tables.patch is or why it's even in the repository. But opencl_std_ext_inst.inc seems to have been generated externally.

More kernel related fixes and tests coming in later ;)

dekimir:
`opencl_std_ext_inst.inc` is auto-generated from the spec itself.  When that bug is resolved and the spec updated, this project will reflect that without manual intervention.  Before that, we prefer to stick with the official spec as it stands.

I'm not sure I received an answer to my larger question: why are the `pow` and `pown` changes necessary for this PR?

dekimir:
In fact, when I pull this branch and back out the changes to `test/ExtInst.OpenCL.std.cpp` and `source/opencl_std_ext_inst.inc` (keeping everything else), the project builds and tests just fine.

@a2flo if you drop changes to those two files, this can go in immediately.

a2flo:
I'm just trying to get SPIR-V code that is generated by SPIRV-LLVM to validate correctly. I can split this up into multiple PRs if you want me to.

Concerning pow/pown: these are clearly and obviously wrong and should be fixed right away. Even more so because SPIRV-LLVM is the only OpenCL SPIR-V compiler right now (afaik) and it's doing it the correct way. And no, I haven't pushed a test for these yet. They test fine, because the macro generated tests assume the same broken pow/pown behavior.

dekimir:
Yes, please split the PR and leave the `pow`/`pown` changes for later.  I agree they should be fixed, but this project is not where you fix them -- we just reflect what the spec says.

a2flo:
*reverted

dekimir:
Wait, did you close the PR by mistake?  It says " @a2flo closed this 20 minutes ago". :confused:

a2flo:
GitHub PRs are weird, can't I finish / close up this PR without it automatically pulling in changes from future changes in master?

dekimir:
It'll close automatically when I merge the code in, which I'm about to do.

I've also asked the SPIR-V working group about the pow/pown bug and have been assured that it's being fixed as we speak.  I'll make sure it's on the agenda at the next working group meeting on Tuesday, if it's not fixed by then.

dekimir:
Squashed and merged into master via 680f9b7.

PR #142, opened 2016-03-02
bitcast fixes

a2flo:
-> commit message

I'm not sure about the spvOpcodeIsPointer/spvOpcodeIsPotentialPointer split, but putting OpBitcast into spvOpcodeIsPointer isn't correct either (imo). Since it's only used in OpLoad and OpStore it might make sense to put everything in spvOpcodeIsPotentialPointer and simply remove spvOpcodeIsPointer?
Note though that both OpLoad and OpStore now go on and check if the operand is actually a pointer type, making the spvOpcodeIsPointer/spvOpcodeIsPotentialPointer redundant -> could remove the check + spvOpcodeIsPointer/spvOpcodeIsPotentialPointer, but the error msg might be nice to have?

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -236,22 +236,10 @@ int32_t spvOpcodeIsPointer(const SpvOp opcode) {
   switch (opcode) {
     case SpvOpVariable:
     case SpvOpAccessChain:
-    case SpvOpPtrAccessChain:
     case SpvOpInBoundsAccessChain:
-    case SpvOpInBoundsPtrAccessChain:
     case SpvOpFunctionParameter:
-      return true;
-    default:
-      return false;
-  }
-}
-
-int32_t spvOpcodeIsPotentialPointer(const SpvOp opcode) {
-  if (spvOpcodeIsPointer(opcode))
-    return true;
-
-  switch (opcode) {
-    case SpvOpBitcast:
+    case SpvOpImageTexelPointer:
+    case SpvOpCopyObject:
       return true;
     default:
       return false;
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -246,6 +246,18 @@ int32_t spvOpcodeIsPointer(const SpvOp opcode) {
   }
 }
 
+int32_t spvOpcodeIsPotentialPointer(const SpvOp opcode) {
+  if (spvOpcodeIsPointer(opcode))
+    return true;
+
+  switch (opcode) {
+    case SpvOpBitcast:
+      return true;
+    default:
+      return false;
+  }
+}
+
 int32_t spvOpcodeGeneratesType(SpvOp op) {
   switch (op) {
     case SpvOpTypeVoid:
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -232,7 +232,7 @@ int32_t spvOpcodeIsComposite(const SpvOp opcode) {
   }
 }
 
-int32_t spvOpcodeIsPointer(const SpvOp opcode) {
+int32_t spvOpcodeReturnsLogicalPointer(const SpvOp opcode) {
   switch (opcode) {
     case SpvOpVariable:
     case SpvOpAccessChain:
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -763,7 +763,7 @@ bool idUsage::isValid<SpvOpLoad>(const spv_instruction_t* inst,
   auto pointer = usedefs_.FindDef(inst->words[pointerIndex]);
   if (!pointer.first ||
       (addressingModel == SpvAddressingModelLogical &&
-       !spvOpcodeIsPointer(pointer.second.opcode))) {
+       !spvOpcodeReturnsLogicalPointer(pointer.second.opcode))) {
     DIAG(pointerIndex) << "OpLoad Pointer <id> '" << inst->words[pointerIndex]
                        << "' is not a pointer.";
     return false;
@@ -793,7 +793,7 @@ bool idUsage::isValid<SpvOpStore>(const spv_instruction_t* inst,
   auto pointer = usedefs_.FindDef(inst->words[pointerIndex]);
   if (!pointer.first ||
       (addressingModel == SpvAddressingModelLogical &&
-       !spvOpcodeIsPointer(pointer.second.opcode))) {
+       !spvOpcodeReturnsLogicalPointer(pointer.second.opcode))) {
     DIAG(pointerIndex) << "OpStore Pointer <id> '" << inst->words[pointerIndex]
                        << "' is not a pointer.";
     return false;
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -755,7 +755,7 @@ bool idUsage::isValid<SpvOpLoad>(const spv_instruction_t* inst,
            return false);
   auto pointerIndex = 3;
   auto pointer = usedefs_.FindDef(inst->words[pointerIndex]);
-  if (!pointer.first || !spvOpcodeIsPointer(pointer.second.opcode)) {
+  if (!pointer.first || !spvOpcodeIsPotentialPointer(pointer.second.opcode)) {
     DIAG(pointerIndex) << "OpLoad Pointer <id> '" << inst->words[pointerIndex]
                        << "' is not a pointer.";
     return false;
@@ -783,13 +783,18 @@ bool idUsage::isValid<SpvOpStore>(const spv_instruction_t* inst,
                                   const spv_opcode_desc) {
   auto pointerIndex = 1;
   auto pointer = usedefs_.FindDef(inst->words[pointerIndex]);
-  if (!pointer.first || !spvOpcodeIsPointer(pointer.second.opcode)) {
+  if (!pointer.first || !spvOpcodeIsPotentialPointer(pointer.second.opcode)) {
     DIAG(pointerIndex) << "OpStore Pointer <id> '" << inst->words[pointerIndex]
                        << "' is not a pointer.";
     return false;
   }
   auto pointerType = usedefs_.FindDef(pointer.second.type_id);
-  assert(pointerType.first);
+  if (!pointer.first || pointerType.second.opcode != SpvOpTypePointer) {
+    DIAG(pointerIndex) << "OpStore type for pointer <id> '"
+                       << inst->words[pointerIndex]
+                       << "' is not a pointer type.";
+    return false;
+  }
   auto type = usedefs_.FindDef(pointerType.second.words[3]);
   assert(type.first);
   spvCheck(SpvOpTypeVoid == type.second.opcode, DIAG(pointerIndex)
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -50,6 +50,8 @@ class idUsage {
           const spv_operand_table operandTableArg,
           const spv_ext_inst_table extInstTableArg,
           const spv_instruction_t* pInsts, const uint64_t instCountArg,
+          const SpvMemoryModel memoryModelArg,
+          const SpvAddressingModel addressingModelArg,
           const UseDefTracker& usedefs,
           const std::vector<uint32_t>& entry_points, spv_position positionArg,
           spv_diagnostic* pDiagnosticArg)
@@ -58,6 +60,8 @@ class idUsage {
         extInstTable(extInstTableArg),
         firstInst(pInsts),
         instCount(instCountArg),
+        memoryModel(memoryModelArg),
+        addressingModel(addressingModelArg),
         position(positionArg),
         pDiagnostic(pDiagnosticArg),
         usedefs_(usedefs),
@@ -74,6 +78,8 @@ class idUsage {
   const spv_ext_inst_table extInstTable;
   const spv_instruction_t* const firstInst;
   const uint64_t instCount;
+  const SpvMemoryModel memoryModel;
+  const SpvAddressingModel addressingModel;
   spv_position position;
   spv_diagnostic* pDiagnostic;
   UseDefTracker usedefs_;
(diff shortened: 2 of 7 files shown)


a2flo:
On a related note: is there any restriction on which kind of pointer casts OpBitcast is allowed to do? I'd think that casts between different storage classes are forbidden, but it doesn't say so in the spec. Or is this wanted undefined behavior? Technically, I guess you could bitcast all you like as long as you don't actually use the pointer.

Which brings me to the next problem: OpStore doesn't actually check if the pointer operand can be written to. There are tests that perform stores to UniformConstant memory, which should not be valid as it is read-only memory.

Overall: would it make sense to forbid storage class casts in OpBitcast and/or should OpStore track down the original pointer and check if it can actually write to it?
OpStore should probably check the writability of the specified pointer operand right now anyways (w/o tracking down the original pointer).

source/validate_id.cpp
```diff
@@ -740,7 +740,7 @@ bool idUsage::isValid<SpvOpLoad>(const spv_instruction_t* inst,
            return false);
   auto pointerIndex = 3;
   auto pointer = usedefs_.FindDef(inst->words[pointerIndex]);
-  if (!pointer.first || !spvOpcodeIsPointer(pointer.second.opcode)) {
+  if (!pointer.first || !spvOpcodeIsPotentialPointer(pointer.second.opcode)) {
```
dekimir: This isn't quite right.  The check here should be whether the _type_ of the `pointer` ID is an `OpTypePointer` result.  That cannot be determined from the opcode alone.  Please consider something like this instead:

```
auto pointerType = usedefs_.FindDef(pointer.second.type_id);
if (!pointerType.first || pointerType.second.opcode != SpvOpTypePointer) {
...
```

test/ValidateID.cpp
```diff
@@ -823,6 +823,40 @@ TEST_F(ValidateID, OpLoadPointerBad) {
...
+%1  = OpTypeVoid
+%2  = OpTypeInt 32 1
+%9  = OpTypeFloat 32
+%3  = OpTypePointer UniformConstant %2
+%4  = OpTypeFunction %1
+%6  = OpFunction %1 None %4
+%7  = OpLabel
+%8  = OpLoad %2 %3
+%10 = OpBitcast %9 %8
+%8  = OpLoad %2 %10
```
dekimir: %8 cannot be defined twice.

test/ValidateID.cpp
```diff
@@ -823,6 +823,40 @@ TEST_F(ValidateID, OpLoadPointerBad) {
...
+TEST_F(ValidateID, OpLoadBitcastNonPointerBad) {
+  const char* spirv = R"(
+%1  = OpTypeVoid
+%2  = OpTypeInt 32 1
+%9  = OpTypeFloat 32
+%3  = OpTypePointer UniformConstant %2
+%4  = OpTypeFunction %1
+%6  = OpFunction %1 None %4
+%7  = OpLabel
+%8  = OpLoad %2 %3
```
dekimir: The pointer argument must be a variable, not a type.

test/ValidateID.cpp
```diff
@@ -823,6 +823,40 @@ TEST_F(ValidateID, OpLoadPointerBad) {
...
+  const char* spirv = R"(
+%1  = OpTypeVoid
+%2  = OpTypeInt 32 1
+%9  = OpTypeFloat 32
+%3  = OpTypePointer UniformConstant %2
+%10 = OpTypePointer UniformConstant %9
+%4  = OpTypeFunction %1
+%6  = OpFunction %1 None %4
+%7  = OpLabel
+%11 = OpBitcast %10 %3
```
dekimir: The second argument must be an object, not a type.

test/ValidateID.cpp
dekimir: To be more precise, the spec says in section 2.16.1:

> A pointer can only be created by the following instructions:
> - `OpVariable`
> - `OpAccessChain`
> - `OpInBoundsAccessChain`

It is unfortunate that the `OpLoad` blurb isn't clearer about it.

dekimir:
>Overall: would it make sense to forbid storage class casts in OpBitcast and/or should OpStore track down the original pointer and check if it can actually write to it?

These are excellent questions, though the first one is a question for the spec, not the validator.  As for `OpStore`, I think we should check the storage class of its pointer and forbid storing into read-only classes.

source/validate_id.cpp
a2flo: This is already checked after that check, hence the "this is redundant, but might still be nice to have for diagnostic reasons".

test/ValidateID.cpp
a2flo: You're right with that bug of course. But it seems like the spec is incomplete at that point, there are more opcodes that "create" pointers and others that (can) result in pointers.

a2flo: Oh wait, this only concerns Logical addressing, not Physical*. Even then, isn't OpFunctionParameter a pointer as well? And OpBitcast isn't restricted to Phyiscal, so it can "create" pointers as well. Really depends on how you define "create".

source/validate_id.cpp
dekimir: I guess `spvOpcodeIsPointer` is meant to enforce the spec section 2.16.1.  In that case, it shouldn't be extended to `OpBitCast`, as the spec seems to forbid that.  BUT it should be limited to the Logical addressing model.  I presume OpenCL implementations will use a physical addressing model and won't be hampered by this.

test/ValidateID.cpp
dekimir: My understanding is that `OpBitCast` was deliberately left out from that list.  `OpFunctionParameter` is interesting -- I'll ask about that...

dekimir: So `OpFunctionParameter` should be on the list of pointer-creating instructions.  I've filed a spec bug to add it.

source/validate_id.cpp
a2flo: Well, OpBitCast doesn't say it's limited to Kernel or Addresses and it clearly states that it can bitcast pointers, so it should be possible even for Logical/shaders.

Also, where exactly in the spec does it say that you have to use Logical for shaders? I can't seem to find this.

Either way, there are more instructions that can result in pointers even for Logical/shaders (2.16.1 is incomplete as I said).

So, it would be better to just remove spvOpcodeIsPointer/spvOpcodeIsPotentialPointer and just validate that the input pointer operand is actually a pointer.

dekimir: >Well, OpBitCast doesn't say it's limited to Kernel or Addresses and it clearly states that it can bitcast pointers, so it should be possible even for Logical/shaders.

That's not what the spec currently says.  Perhaps you can file a bug and get it changed?  My impression, from talking to the spec authors, is that `OpBitCast` was deliberately forbidden from creating a pointer in the Logical addressing model.  That's why I believe the `spvOpcodeIsPointer()` should be kept, but only applied in the Logical model.

> Also, where exactly in the spec does it say that you have to use Logical for shaders?

I don't think the spec says anything like that.  It merely says what's legal if you _do_ use Logical.

> (2.16.1 is incomplete as I said).

Agreed: `OpFunctionParameter` is missing.  The spec will be updated to reflect that.  You can even include it in `spvOpcodeIsPointer()` today.

> So, it would be better to just remove spvOpcodeIsPointer/spvOpcodeIsPotentialPointer

I agree about `IsPotentialPointer`, but I believe `IsPointer` should stay, as I said above.

dekimir: Actually, the Vulkan spec [does say](https://www.khronos.org/registry/vulkan/specs/1.0/xhtml/vkspec.html#spirvenv-module-validation) that Logical must be selected for graphics modules.

a2flo: What about OpImageTexelPointer, OpCopyObject, OpSelect or OpPhi?
The "A pointer can only be an operand to the following instructions" list is incomplete too. This is missing all of the atomic functions for example.

I think both of these lists shouldn't really exist and any restrictions should either be explicitly listed for each opcode or should be implied through capabilities.

Quite unfortunate about bitcast though.

> Actually, the Vulkan spec does say that Logical must be selected for graphics modules.

Ah, thanks.

dekimir: > I think both of these lists shouldn't really exist and any restrictions should either be explicitly listed for each opcode or should be implied through capabilities.

This is a valid opinion, but again, it concerns the spec and not this code. We just follow the spec.

How about simply restricting the `spvOpcodeIsPointer()` call to only the Logical address model?  That can be accepted today, and it should save OpenCL implementations from improperly restrictive validation.

dekimir: Meanwhile, I filed a [spec bug](https://www.khronos.org/bugzilla/show_bug.cgi?id=1471) for this and will ensure it's addressed pronto by the SPIR-V working group.

dekimir:
@a2flo: the spec rules have been updated recently to [allow more instructions](https://www.khronos.org/registry/spir-v/specs/1.1/SPIRV.html#_universal_validation_rules) to produce and consume pointers in the logical addressing model.  Would you like to modify this PR accordingly?  If not, I'll craft a separate PR that fixes the restricted-opcodes set and applies this verification only in the logical model.

a2flo:
Sure, will do. Remove the pointer check for physical/compute altogether + add new allowed logical ops + move the bitcast tests to the physical/compute part, because they make no sense in the logical part (mostly), right?
Also, is there going to be a better fix to this, since the bug report is still open? (no OpPhi/OpSelect is a major shortcoming IMO)

dekimir:
@a2flo: yes, that list ("Remove the pointer check...") is right.

There is ongoing spec work to address the larger issues; we just figured we'd deliver the expanded whitelists quickly to remove the immediate blockage.

a2flo:
Alright, fixed + rebased. I hope this is the best way to retrieve and store the addressing/memory model, this seemed to be the most reasonable way of doing this and it defaults to the least permissive models.

source/opcode.cpp
```diff
@@ -236,10 +236,10 @@ int32_t spvOpcodeIsPointer(const SpvOp opcode) {
   switch (opcode) {
```
dekimir: The name of this function is now misleading.  It should be more indicative of its function, which is whether the opcode may produce a pointer (in a particular addressing model).

dekimir: Moreover, since this is only used in validation, the method could move into `ValidationState_t`.

test/ValidateID.cpp
```diff
@@ -1694,6 +1694,76 @@ TEST_F(ValidateID, OpPtrAccessChainGood) {
...
+%3  = OpTypeInt 32 1
+%4  = OpTypeFloat 32
+%5  = OpTypePointer UniformConstant %3
+%6  = OpTypeFunction %2
+%7  = OpVariable %5 UniformConstant
+%8  = OpFunction %2 None %6
+%9  = OpLabel
+%10 = OpLoad %3 %7
+%11 = OpBitcast %4 %10
+%12 = OpLoad %3 %11
```
dekimir: You would think that this fails not because of `OpBitcast` but because of `%4`.  It doesn't seem related to the code being changed above, and it's likely already covered elsewhere but with a different opcode.  OTOH, another test isn't a bad thing. No change needed.

dekimir:
Looks pretty good, and the discovery/storing of models is reasonable.  Just needs a better name for the opcode-is-pointer function.

source/opcode.cpp
a2flo: spvOpcodeIsLogicalPointer maybe? or spvOpcodeReturnsLogicalPointer?
Not sure about the pulling into idUsage part though. It makes sense in that it could directly use the addressingModel variable, but on the other hand you still want a global function to test this (like the others in opcode.h), even if it isn't used anywhere else right now. An alternative could be to add an SpvAddressingModel argument to it (if the spvOpcodeIsPointer name is kept, otherwise it would be confusing).

test/ValidateID.cpp
a2flo: Well, yes and no. Since OpBitcast can both return a pointer or a non-pointer type, this is supposed to test OpLoad when it can't determine this from the opcode alone, but has to test the id type. Hence this fails at OpLoad, because %11 isn't a pointer.
The 32-bit int to float bitcast is however valid (even in logical mode).

a2flo: Oh, and this is related, because isValid<SpvOpStore> was missing that id pointer type check, so I added these tests for both of them, just to make sure they do the right thing when using OpBitcast (OpLoad from a non-pointer is of course already tested elsewhere).

source/opcode.cpp
dekimir: Oooh, I like passing in addressing model in.  Something like
```
bool spvMayReturnPointer(SpvOp, SpvAddressingModel);
```

test/ValidateID.cpp
dekimir: Doesn't OpLoad always look at the operand type first, before tracking the value back to OpBitcast?  And if that first check fails (b/c the type is not a pointer type), then the value producer doesn't matter.

If that's not how it happens, it should be.  Which in turn makes me think that the `spvMayReturnPointer` calls shouldn't happen in validating the load/store instructions but the instructions whose result type is a pointer.  But that's for another PR...

a2flo: Nope, OpLoad/OpStore first check if the id came from an opcode that can/does return a pointer, then they check if id actually is a pointer, then they check if the result type matches. I had originally hoped that the first check is redundant, but it isn't in logical addressing mode. It _is_ redundant in physical addressing mode however, because it doesn't have pointer opcode restrictions and the type check will make sure that it's actually using a pointer. So a `spvMayReturnPointer` is only necessary for the logical addressing model, which is way I'd rather have a "logical" in that name or always return true when using it with a `SpvAddressingModelPhysical*` argument.

And yes, for the time being it's better to do this check in OpLoad/OpStore (and actually all other pointer taking opcodes ...) rather checking it in all opcodes that can return a pointer (which includes the ones that aren't allowed in logical mode). Until validate_id.cpp is completely done this would probably break too much stuff.

source/opcode.cpp
a2flo: -> other comment below
(I'd rather not add checks for everything that could possibly ever return a pointer in physical mode, so either always return true in physical mode here, or have "logical" in the function name)

dekimir: I'm leaning towards "always return true in physical mode".  Seems more general.

test/ValidateID.cpp
dekimir: Agreed; let's keep this as it currently is.

source/opcode.cpp
dekimir: Actually, wait, I take that back.  It wouldn't be right to return true for, say, `OpTypeInt`.  Let's bake "logical" into the function name.

a2flo:
Okay, renamed it. This should be all now, right?

dekimir:
Squashed (keeping authorship) and merged to master via 66fcb45.

Issue #144, opened 2016-03-05
update_build_version script syntax and str object issues


bschindler:
I'm trying to build SPIRV-Tools under gentoo linux (with python-3.5) and I run into the following issues: 

[  3%] Update build-version.inc in the Spirv-tools build directory (if necessary).
/usr/bin/python3.5 /home/benjamin/Software/SPIRV-Tools/utils/update_build_version.py /home/benjamin/Software/SPIRV-Tools
  File "/home/benjamin/Software/SPIRV-Tools/utils/update_build_version.py", line 67
    print 'usage: {0} <spirv-tools_dir>'.format(sys.argv[0])
                                       ^
SyntaxError: invalid syntax

I commented out line 67 as it is likely not used during build but then I run into this issue: 

[  3%] Update build-version.inc in the Spirv-tools build directory (if necessary).
/usr/bin/python3.5 /home/benjamin/Software/SPIRV-Tools/utils/update_build_version.py /home/benjamin/Software/SPIRV-Tools
Traceback (most recent call last):
  File "/home/benjamin/Software/SPIRV-Tools/utils/update_build_version.py", line 76, in <module>
    main()
  File "/home/benjamin/Software/SPIRV-Tools/utils/update_build_version.py", line 70, in main
    new_content = ('"spirv-tools ' + describe(sys.argv[1]).replace('"', '\\"') + '\\n"\n')
TypeError: a bytes-like object is required, not 'str'

As my python experience is extremely limited, I thought I should just file this.

antiagainst:
@bschindler: The script is using python2 syntax right now. The `print` error results from [syntax difference between python2 vs. python3](https://docs.python.org/3.0/whatsnew/3.0.html#print-is-a-function).

If you have python2 installed, you may want to change the first line of `utils/update_build_version.py` to `#!/usr/bin/env python2` or symlink your system Python to python2.

Sorry for the inconvenience, we'll fix the issue.

bschindler:
That does explain it. Using python2 did work here. Thanks for the response

dekimir:
Let's keep this open and make the script work for both Python 2 and Python 3.

a2flo:
Note that changing the first line of utils/update_build_version.py to python2 doesn't work when cmake finds python3 and uses it to call the script:
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/CMakeLists.txt#L119 ->
`${PYTHON_EXECUTABLE}
[...]

PR #145, opened 2016-03-06
Improve Python3 Compatibility

antiagainst:
Addresses #144.

 * Use `print_function` from `__future__` to tackle the print syntax
   change.
 * In Python3 `str` and `bytes` are no longer the same thing. Need
   proper decode.

### DIFF
--- a/utils/update_build_version.py
+++ b/utils/update_build_version.py
@@ -71,6 +71,11 @@ def main():
         sys.exit(1)
 
     new_content = '"spirv-tools {}\\n"\n'.format(
+        # decode() is needed here for Python3 compatibility. In Python2,
+        # str and bytes are the same type, but not in Python3.
+        # Popen.communicate() returns a bytes instance, which needs to be
+        # decoded into text data first in Python3. And this decode() won't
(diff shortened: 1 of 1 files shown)


utils/update_build_version.py
```diff
@@ -54,20 +57,21 @@ def describe(dir):
...
 
 def main():
     if len(sys.argv) != 2:
-        print 'usage: {0} <spirv-tools_dir>'.format(sys.argv[0])
+        print('usage: {0} <spirv-tools_dir>'.format(sys.argv[0]))
         sys.exit(1)
 
-    new_content = ('"spirv-tools ' + describe(sys.argv[1]).replace('"', '\\"') + '\\n"\n')
+    new_content = '"spirv-tools {}\\n"\n'.format(
+        describe(sys.argv[1]).decode('ascii').replace('"', '\\"'))
```
dekimir: Needs a similar comment about `decode('ascii')`.

antiagainst: Done.

dekimir:
+2

PR #147, opened 2016-03-11
Fix the corner case in version generation script.

antiagainst:
When the given directory is not inside the SPIRV-Tools project,
`describe()` returns a `str` instance instead of `bytes` instance
in Python3, which will case problem when calling `decode()` on it.

### DIFF
--- a/utils/update_build_version.py
+++ b/utils/update_build_version.py
@@ -62,7 +62,8 @@ def describe(dir):
         try:
             return command_output(['git', 'rev-parse', 'HEAD'], dir).rstrip()
         except:
-            return 'unknown hash, ' + datetime.date.today().isoformat()
+            return 'unknown hash, {}'.format(
+                datetime.date.today().isoformat()).encode('ascii')
 
 
 def main():


antiagainst:
As suggested by https://github.com/KhronosGroup/SPIRV-Tools/commit/f7e63786a919040cb2e0e572d960a0650f2c2881#commitcomment-16653277.

utils/update_build_version.py
```diff
@@ -62,7 +62,8 @@ def describe(dir):
         try:
             return command_output(['git', 'rev-parse', 'HEAD'], dir).rstrip()
         except:
-            return 'unknown hash, ' + datetime.date.today().isoformat()
+            return 'unknown hash, {}'.format(
+                datetime.date.today().isoformat()).encode('ascii')
```
dekimir: I think it would be much more useful to leave this as is and add `decode` to lines 60 and 63.

But if you want this, then please change the blurb on line 54.

antiagainst: Sorry. See #148.

PR #150, opened 2016-03-14
Add patchlevel to CMake requirement

danlrobertson:
As far as I know, CMake <2.8.12 will not have `target_compile_options`. See the [CMake 2.8.12 ChangeLog](https://cmake.org/files/v2.8/CMakeChangeLog-2.8.12) for details.


dekimir:
Makes sense.  Thanks, @danlrobertson!

danlrobertson:
Np, thanks for the quick response!

Issue #152, opened 2016-03-15
spirv-as reports whole program as the invalid opcode


dekimir:
I assemble a simple program with an incorrect opcode at the end.  `spirv-as` should report just that opcode as invalid, but instead it reports the whole program as the opcode name:

```
deki@deki-macbookpro>cat bug.spvasm 
OpCapability Shader
OpMemoryModel Logical GLSL450
OpXYZ
deki@deki-macbookpro>./spirv-as bug.spvasm 
error: 3: 1: Invalid Opcode name 'OpCapability Shader
OpMemoryModel Logical GLSL450
OpXYZ'
```

antiagainst:
Interesting. The symptom is the same as #126. But this time EOL is there...

PR #154, opened 2016-03-15
Remove duplicate and buggy implementation of getWord().

antiagainst:
We already have a better `getWord()` implemented as a free function.
Use that instead of creating another one.

Addresses #152.

### DIFF
--- a/source/text_handler.cpp
+++ b/source/text_handler.cpp
@@ -143,6 +143,8 @@ spv_result_t getWord(spv_text text, spv_position position, std::string& word,
         case ' ':
         case ';':
         case '\t':
+        case '\v':
+        case '\r':
         case '\n':
           if (escaping || quoting) break;
         // Fall through.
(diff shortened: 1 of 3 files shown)


test/TextToBinary.cpp
```diff
@@ -62,6 +62,27 @@ TEST(GetWord, Simple) {
   EXPECT_EQ("abc", AssemblyContext(AutoText("abc\n"), nullptr).getWord());
 }
 
+TEST(GetWord, MultipleWords) {
```
dekimir: This tests more than just multiple words -- it tests various forms of word break.  I'd split it into multiple tests with separate responsibilities.

test/TextToBinary.cpp
```diff
@@ -62,6 +62,27 @@ TEST(GetWord, Simple) {
...
+  auto context = AssemblyContext(&text, nullptr);
+
+  EXPECT_EQ("abc", context.getWord());
+
+  spv_position_t position = {};
+  position = {1, 1, contents.find('d')};
+  context.setPosition(position);
+  EXPECT_EQ("def", context.getWord()) << position.index;
+
+  position = {1, 5, contents.find('g')};
```
dekimir: Why is `contents.find(...)` necessary if 5 is hard-coded?

antiagainst: Changed. Previously because of line endings, `index` is not the same as `column`.

test/TextToBinary.cpp
antiagainst: Reasonable. Removed different whitespaces since they are tested in the previous test.

source/text_handler.cpp
```diff
@@ -234,22 +236,11 @@ bool AssemblyContext::hasText() const {
...
-  uint64_t index = current_position_.index;
-  while (true) {
-    switch (text_->str[index]) {
-      case '\0':
-      case '\t':
-      case '\v':
-      case '\r':
-      case '\n':
-      case ' ':
-        return std::string(text_->str, text_->str + index);
```
antiagainst: This is the direct reason why #126 behaves like that. We should return `string(text_->str + current_position_.index, ...)`.

antiagainst: But I figure it's better to use the more robust version of `getWord()` anyway.

antiagainst:
A better way to fix this in done via #155.

PR #158, opened 2016-03-16
Travis CI: use trusty (Ubuntu 14.04) for Linux.

antiagainst:


antiagainst:
This is **much** nicer. Now I don't need to install the build essentials by myself.

dekimir:
+2

Issue #153, opened 2016-03-15
include/spirv/OpenCL.std.h:271:2: error: extra ‘;’ [-Werror=pedantic]


azur3:
Building on Ubuntu and getting this message.

Need to fix the code or remove pedantic?

Thanks

dekimir:
Hi, @azur3.  Sorry about the trouble.

Could you provide the whole build log?  How closely are you following the README instructions?

dneto0:
This is a problem with the OpenCL header as published in the SPIR-V Registry.

dneto0:
I've filed a bug on the Khronos public bugzilla to track this at the working group level.
https://www.khronos.org/bugzilla/show_bug.cgi?id=1476

dekimir:
@dneto0 agreed, but:

1. if @azur3 is getting this error by turning on `SPIRV_WARN_EVERYTHING`, they can work around the problem by turning it off
2. we can fix `SPIRV_WARN_EVERYTHING` by disabling certain warnings for files we can't control that have known problems

dneto0:
Correct.  I don't want to turn off things in SPIRV_WARN_EVERYTHING because that's how I noticed the pedantic problem in the first place.
Recommend action: don't use SPIRV_WARN_EVERYTHING or -pedantic right now.  Use those only after we get a fixed header file from upstream.

AWoloszyn:
This should now be fixed.

PR #162, opened 2016-03-18
Make everything compile cleanly with -Wpedantic.

AWoloszyn:
This makes SPIRV_WARN_EVERYTHING compile cleanly.

### DIFF
--- a/test/AssemblyContext.cpp
+++ b/test/AssemblyContext.cpp
@@ -82,7 +82,7 @@ INSTANTIATE_TEST_CASE_P(
       // A very long string, encoded after an initial word.
       // SPIR-V limits strings to 65535 characters.
       {std::string(65535, 'a'), {1}},
-    }));
+    }),);
 // clang-format on
 
 }  // anonymous namespace
--- a/test/Validate.Capability.cpp
+++ b/test/Validate.Capability.cpp
@@ -37,6 +37,7 @@
 
 namespace {
 
+
 using std::pair;
 using std::make_pair;
 using std::stringstream;
@@ -52,7 +53,7 @@ using ValidateCapability =
     spvtest::ValidateBase<tuple<string, pair<string, vector<string>>>>;
 
 TEST_F(ValidateCapability, Default) {
-  const char str[] = R"(
+    const char str[] = R"(
             OpCapability Kernel
             OpCapability Matrix
             OpMemoryModel Logical OpenCL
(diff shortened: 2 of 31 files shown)


AWoloszyn:
Fixes: https://github.com/KhronosGroup/SPIRV-Tools/issues/153

test/UnitSPIRV.h
```diff
@@ -215,5 +215,19 @@ inline std::string MakeLongUTF8String(size_t num_4_byte_chars) {
   return result;
 }
 
+template <typename T>
+std::string DefaultParameterPrinter(
```
dekimir: Needs blurb.

test/UnitSPIRV.h
```diff
@@ -215,5 +215,19 @@ inline std::string MakeLongUTF8String(size_t num_4_byte_chars) {
...
+    const typename testing::TestParamInfo<T>& instance) {
+  std::ostringstream ss;
+  ss << instance.index;
+  return ss.str();
+}
+
+template<typename T>
+std::string EnumParameterPrinter(
```
dekimir: Needs blurb.

test/Validate.Capability.cpp
```diff
@@ -51,8 +53,15 @@ using testing::ValuesIn;
 using ValidateCapability =
     spvtest::ValidateBase<tuple<string, pair<string, vector<string>>>>;
 
+std::string ValidateCapabilityPrinter(
```
dekimir: Needs blurb.

test/AssemblyContext.cpp
```diff
@@ -82,7 +83,8 @@ INSTANTIATE_TEST_CASE_P(
       // A very long string, encoded after an initial word.
       // SPIR-V limits strings to 65535 characters.
       {std::string(65535, 'a'), {1}},
-    }));
+    }),
+    &DefaultParameterPrinter<EncodeStringCase>);
```
dekimir: This is onerous (we must add junk to all INSTANTIATE calls now and forever) and surprising (we can't use INSTANTIATE the way Googletest doc suggest). There has to be a better way. :pray:

dekimir: Turns out the variable-arguments list may be empty.  This compiles cleanly with `--pedantic`:

```
#define foo(a, ...) a
int main() {
  return foo(0,);
}
```

AWoloszyn:
Replaced arguments with ,

dneto0:
+2

PR #163, opened 2016-03-29
Capability fix: ImageReadWrite & ImageMipmap require ImageBasic.

antiagainst:


dneto0:
+2
Verified against SPIR-V 1.0 Rev 4

PR #165, opened 2016-03-30
Change spvContextCreate() to take a target environment.

dekimir:
Add the spv_target_env enum.

Run clang-format.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -343,8 +343,13 @@ typedef spv_context_t* spv_context;
 
 // Platform API
 
-// Creates a context object.
-spv_context spvContextCreate();
+// Certain target environments impose additional restrictions on SPIR-V, so it's
+// often necessary to specify which one applies.  SPV_ENV_UNIVERSAL means
+// environment-agnostic SPIR-V.
+enum spv_target_env { SPV_ENV_UNIVERSAL, SPV_ENV_VULKAN };
+
+// Creates a context object.  Returns null if env is invalid.
+spv_context spvContextCreate(spv_target_env env);
 
 // Destroys the given context object.
 void spvContextDestroy(spv_context context);
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -346,7 +346,7 @@ typedef spv_context_t* spv_context;
 // Certain target environments impose additional restrictions on SPIR-V, so it's
 // often necessary to specify which one applies.  SPV_ENV_UNIVERSAL means
 // environment-agnostic SPIR-V.
-enum spv_target_env { SPV_ENV_UNIVERSAL, SPV_ENV_VULKAN };
+enum spv_target_env { SPV_ENV_UNIVERSAL_1_0_4, SPV_ENV_VULKAN_1_0_7 };
 
 // Creates a context object.  Returns null if env is invalid.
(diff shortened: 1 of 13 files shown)


dekimir:
@antiagainst @AWoloszyn @Qining @dneto0 PTAL

include/spirv-tools/libspirv.h
```diff
@@ -343,8 +343,13 @@ typedef spv_context_t* spv_context;
 
 // Platform API
 
-// Creates a context object.
-spv_context spvContextCreate();
+// Certain target environments impose additional restrictions on SPIR-V, so it's
+// often necessary to specify which one applies.  SPV_ENV_UNIVERSAL means
+// environment-agnostic SPIR-V.
+enum spv_target_env { SPV_ENV_UNIVERSAL, SPV_ENV_VULKAN };
```
dneto0: How do we accomodate evolution of the SPIR-V spec or the Vulkan spec?
It seems to me there be versions attached.
E.g. SPV_ENV_UNIVSERAL_1_0 and SPV_ENV_VULKAN_1_0
If we don't add the version distinction now, then the meaning of these non-versioned names will be ambiguous in the future.

test/TextToBinary.cpp
```diff
@@ -538,4 +540,21 @@ TEST(AssemblyContextParseMessages, Errors) {
   spvDiagnosticDestroy(diag);
 }
 
+TEST(CreateContext, ValidEnvironment) {
+  EXPECT_THAT(spvContextCreate(static_cast<spv_target_env>(99999999)),
```
dneto0: GCC fails to compile this since the 99999999 value is out of range of the underlying type.  GCC might be using an 8 bit or 16 bit int for the enum.
Suggest using -1 as the invalid value.

dneto0:
Overall, this looks good. 
-1 for compilation failure, and I think the enum values should have explicit versions in their names.

include/spirv-tools/libspirv.h
```diff
@@ -346,7 +346,7 @@ typedef spv_context_t* spv_context;
 // Certain target environments impose additional restrictions on SPIR-V, so it's
 // often necessary to specify which one applies.  SPV_ENV_UNIVERSAL means
 // environment-agnostic SPIR-V.
-enum spv_target_env { SPV_ENV_UNIVERSAL, SPV_ENV_VULKAN };
+enum spv_target_env { SPV_ENV_UNIVERSAL_1_0_4, SPV_ENV_VULKAN_1_0_7 };
```
dneto0: It's great having this level of control.  But right now Vulkan is rev'ing at a weekly pace.
Many users will only care about a more coarse grain target.  That is I expect uses that drop the revision number, e.g.
SPV_ENV_UNIVERSAL_1_0 and SPV_ENV_VULKAN_1_0 will be very common.

So lets add those *in addition* to the more specific ones.  The people who care about the tiniest changes can still get what they need.

Also the particular enum values should be stable over time.  So don't make the less specific ones equal to the more specific ones.

I think that also means that even when we retire support for a very specific target env enum, we have to keep its value slot occupied.   But we could change its name to break clients to indicate that the code is no longer supported. We could also cause the context creation to return nullptr when we don't support that value any more.

dekimir:
Feedback addressed.

dneto0:
+2

dekimir:
Squashed and merged to master via 3d27da4.

PR #167, opened 2016-03-31
Use the JSON grammar file to generate various info tables.

antiagainst:
This patch uses a Python script to parse the JSON grammar file to
generate the opcode table and operand kind tables.
    
Now we don't need to do the post-processing (from `OperandClass`
to `spv_operand_type_t`) and copying of the opcode info table is
not required anymore!


dneto0:
The Travis builds fail since simplejson is not available.  Why not just import it as "json"?

source/operand.cpp
```diff
@@ -29,1066 +29,21 @@
...
+// Pull in operand info tables automatically generated from JSON grammar.
+#include "operand.inc"
 
 spv_result_t spvOperandTableGet(spv_operand_table* pOperandTable) {
   if (!pOperandTable) return SPV_ERROR_INVALID_POINTER;
 
-  static const spv_operand_table_t table = {ARRAY_SIZE(opcodeEntryTypes),
-                                            opcodeEntryTypes};
+  static const spv_operand_table_t table = {
+      ARRAY_SIZE(_pygen_variable_OperandInfoTable),
```
dneto0: Global scope identifiers with a single leading underscore are reserved for system use.

utils/generate_grammar_tables.py
```diff
@@ -0,0 +1,329 @@
...
+"""Generates various info tables from SPIR-V JSON grammar."""
+
+from __future__ import print_function
+
+import functools
+import re
+import simplejson as json
+
+# Prefix for all C variables generated by this script.
+PYGEN_VARIABLE_PREFIX = '_pygen_variable'
```
dneto0: Avoid the leading underscore.

utils/generate_grammar_tables.py
```diff
@@ -0,0 +1,329 @@
+#!/usr/bin/env python
+"""Generates various info tables from SPIR-V JSON grammar."""
+
+from __future__ import print_function
+
+import functools
+import re
+import simplejson as json
```
dneto0: This causes failure on the Travis bots.  Can' we just import it as 'json'?

utils/generate_grammar_tables.py
```diff
@@ -0,0 +1,329 @@
...
+
+# Prefix for all C variables generated by this script.
+PYGEN_VARIABLE_PREFIX = '_pygen_variable'
+
+CAPABILITY_BIT_MAPPING = {
```
dneto0: Shouldn't this table come from the JSON file?   Otherwise this script is a second source of truth for the capabilities.

dneto0:
-1

source/operand.cpp
antiagainst: Hmm... http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier interesting

utils/generate_grammar_tables.py
antiagainst: Done

utils/generate_grammar_tables.py
antiagainst: `simplejson` has better error message. It helped me for developing. But yeah, `json` should be fine since the grammar file are syntactically correct now.

utils/generate_grammar_tables.py
antiagainst: Good catch! Done.

dneto0:
+2

PR #168, opened 2016-04-01
Handle CRLF in assembly text.

### LINKED ISSUE #166: spvTextToBinary cannot handle CR+LF unless fread is used by client
I built a small DLL using the SPIRV-Tools.lib library to allow another language to call into SPIRV-Tools using a foreign-function interface of that language to call into C libraries.
The function in the DLL accepted a single memory pointer to the assembly text, which was loaded by the client program before calling into the DLL and eventually into spvTextToBinary.
This assembly text was read from a file as is, and not via fread, which in "r" text mode apparently converts CR+LF to LF under msvc 14.
The function spvTextToBinary() seems to be unable to handle CR+LF linebreaks in the source assembly text.
The spirv-as assembly tool works flawlessly, though, because fread (in "r" text mode) converts CR+LF to LF (\n).
But it does not work when using other mechanism that read the file as is without any conversion of line-breaks.

So the current workaround is to force the client of spvTextToBinary() to convert CR+LF to LF before handing the assembly text to spvTextToBinary.

Error message I am getting for [the example](https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#_example) in the SPIR-V specification:
```
error: 6: 1: Expected <opcode> or <result-id> at the beginning of an instruction, found '
'.
```

dekimir:
Fixes #166.

@AWoloszyn @dneto0 @antiagainst @Qining PTAL.

### DIFF
--- a/source/text_handler.cpp
+++ b/source/text_handler.cpp
@@ -90,6 +90,7 @@ spv_result_t advance(spv_text text, spv_position position) {
       return advance(text, position);
     case ' ':
     case '\t':
+    case '\r':
       position->column++;
       position->index++;
       return advance(text, position);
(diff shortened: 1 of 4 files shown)


AWoloszyn:
+2 LGTM

PR #169, opened 2016-04-01
Fix validation of array length.

dekimir:
@dneto0 @AWoloszyn @Qining @antiagainst PTAL.

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -278,8 +278,7 @@ bool idUsage::isValid<SpvOpTypeArray>(const spv_instruction_t* inst,
   }
   auto lengthIndex = 3;
   auto length = usedefs_.FindDef(inst->words[lengthIndex]);
-  if (!length.first || (SpvOpConstant != length.second.opcode &&
-                        SpvOpSpecConstant != length.second.opcode)) {
+  if (!length.first || !spvOpcodeIsConstant(length.second.opcode)) {
     DIAG(lengthIndex) << "OpTypeArray Length <id> '" << inst->words[lengthIndex]
                       << "' is not a scalar constant type.";
     return false;
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -270,12 +270,12 @@ bool idUsage::isValid<SpvOpTypeSampler>(const spv_instruction_t*,
 // OpTypeInt.
 bool aboveZero(const std::vector<uint32_t>& constWords,
                const std::vector<uint32_t>& typeWords) {
-  const auto width = typeWords[2];
+  const uint32_t width = typeWords[2];
   const bool is_signed = typeWords[3];
-  const auto loWord = constWords[3];
+  const uint32_t loWord = constWords[3];
   if (width > 32) {
     // The spec currently doesn't allow integers wider than 64 bits.
-    const auto hiWord = constWords[4];  // Must exist, per spec.
+    const uint32_t hiWord = constWords[4];  // Must exist, per spec.
     if (is_signed && (hiWord >> 31)) return false;
     return loWord | hiWord;
   } else {
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -272,19 +272,15 @@ bool aboveZero(const std::vector<uint32_t>& constWords,
                const std::vector<uint32_t>& typeWords) {
   const auto width = typeWords[2];
   const bool is_signed = typeWords[3];
-  if (width == 64) {
-    if (is_signed) {
-      int64_t value = constWords[3] | (uint64_t{constWords[4]} << 32);
-      return value > 0;
-    } else {
-      uint64_t value = constWords[3] | (uint64_t{constWords[4]} << 32);
-      return value > 0;
-    }
-  } else {  // Per spec, must be 32 bits or less, sign-extended.
-    if (is_signed)
-      return (constWords[3] > 0) && !(constWords[3] >> 31);
-    else
-      return constWords[3] > 0;
+  const auto loWord = constWords[3];
+  if (width > 32) {
+    // The spec currently doesn't allow integers wider than 64 bits.
+    const auto hiWord = constWords[4];  // Must exist, per spec.
+    if (is_signed && (hiWord >> 31)) return false;
+    return loWord | hiWord;
+  } else {
+    if (is_signed && (loWord >> 31)) return false;
+    return loWord > 0;
   }
 }
 
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -264,6 +264,30 @@ bool idUsage::isValid<SpvOpTypeSampler>(const spv_instruction_t*,
   return true;
 }
 
+// True if the integer constant is > 0. constWords are words of the
+// constant-defining instruction (either OpConstant or
+// OpSpecConstant). typeWords are the words of the constant's-type-defining
+// OpTypeInt.
+bool aboveZero(const std::vector<uint32_t>& constWords,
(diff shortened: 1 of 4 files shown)


source/validate_id.cpp
```diff
@@ -278,8 +278,7 @@ bool idUsage::isValid<SpvOpTypeArray>(const spv_instruction_t* inst,
   }
   auto lengthIndex = 3;
   auto length = usedefs_.FindDef(inst->words[lengthIndex]);
-  if (!length.first || (SpvOpConstant != length.second.opcode &&
-                        SpvOpSpecConstant != length.second.opcode)) {
+  if (!length.first || !spvOpcodeIsConstant(length.second.opcode)) {
```
dneto0: This is too permissive given the code that later checks the value.
The later code uses the lengths of the instructions and maps that to an assumption about it being a SpvOpConstant or SpvOpSpecConstant.
Should have a TODO saying the remaining cases should be handled (or punted entirely)

Qining:
+1, but I don't quite understand the 'remaining cases' in David's comments :(

dekimir:
Feedback addressed.

source/validate_id.cpp
```diff
@@ -264,6 +264,30 @@ bool idUsage::isValid<SpvOpTypeSampler>(const spv_instruction_t*,
...
+// constant-defining instruction (either OpConstant or
+// OpSpecConstant). typeWords are the words of the constant's-type-defining
+// OpTypeInt.
+bool aboveZero(const std::vector<uint32_t>& constWords,
+               const std::vector<uint32_t>& typeWords) {
+  const auto width = typeWords[2];
+  const bool is_signed = typeWords[3];
+  if (width == 64) {
+    if (is_signed) {
+      int64_t value = constWords[3] | (int64_t{constWords[4]} << 32);
```
yavn: This doesn't compile with VS 2013:

```
error C2397: conversion from 'const unsigned int' to 'int64_t' requires a narrowing conversion
```

which I find a bit strange, since it's 32-bit to 64-bit, so there's clearly enough bits.
I'd change it to:

```c++
(uint64_t{constWords[4]} << 32)
```

We're shifting left, so signedness makes no difference.

dekimir: Thanks @yavn.  I changed it to `uint64_t`, as you suggest.  Aside from the VS2013 error, I've realized that shifting a signed value runs into undefined behaviour all too easily.

source/validate_id.cpp
```diff
@@ -264,6 +264,30 @@ bool idUsage::isValid<SpvOpTypeSampler>(const spv_instruction_t*,
...
+bool aboveZero(const std::vector<uint32_t>& constWords,
+               const std::vector<uint32_t>& typeWords) {
+  const auto width = typeWords[2];
+  const bool is_signed = typeWords[3];
+  if (width == 64) {
+    if (is_signed) {
+      int64_t value = constWords[3] | (uint64_t{constWords[4]} << 32);
+      return value > 0;
+    } else {
+      uint64_t value = constWords[3] | (uint64_t{constWords[4]} << 32);
```
dneto0: An unsigned value is non-zero if any of its bits are non-zero.  So you could have saved some work here.  Or you could have been a little more DRY by sharing this computation with the signed case, and then had the signed case do the constructor-cast.

source/validate_id.cpp
```diff
@@ -264,6 +264,30 @@ bool idUsage::isValid<SpvOpTypeSampler>(const spv_instruction_t*,
...
+      return value > 0;
+    }
+  } else {  // Per spec, must be 32 bits or less, sign-extended.
+    if (is_signed)
```
dneto0: The code could be more DRY by putting the constWords[3] value into a variable with a nicer name.

source/validate_id.cpp
```diff
@@ -264,6 +264,30 @@ bool idUsage::isValid<SpvOpTypeSampler>(const spv_instruction_t*,
...
 
+// True if the integer constant is > 0. constWords are words of the
+// constant-defining instruction (either OpConstant or
+// OpSpecConstant). typeWords are the words of the constant's-type-defining
+// OpTypeInt.
+bool aboveZero(const std::vector<uint32_t>& constWords,
+               const std::vector<uint32_t>& typeWords) {
+  const auto width = typeWords[2];
+  const bool is_signed = typeWords[3];
+  if (width == 64) {
```
dneto0: The equality test is not right.
The assembler supports bit widths up to and including 64 bits.
So this code is wrong when using 48 bits, for example.  The 48 bit case is tested elsewhere in the assembler and disassembler.  You may as well handle that case correctly here too.

source/validate_id.cpp
dneto0: That comment no longer applies now that the later code switches on opcode.

test/ValidateID.cpp
```diff
@@ -378,14 +383,93 @@ TEST_F(ValidateID, OpTypeArrayElementTypeBad) {
...
+enum Signed { kSigned, kUnsigned };
+
+// Creates an assembly snippet declaring OpTypeArray with the given length.
+string MakeArrayLength(const string& len, Signed isSigned, int width) {
+  ostringstream ss;
+  ss << " %t = OpTypeInt " << width << (isSigned == kSigned ? " 1" : " 0")
+     << " %l = OpConstant %t " << len << " %a = OpTypeArray %t %l";
+  return ss.str();
+}
+
```
dneto0: Test the 48 bit case.

test/ValidateID.cpp
```diff
@@ -378,14 +383,93 @@ TEST_F(ValidateID, OpTypeArrayElementTypeBad) {
...
+TEST_F(ValidateID, OpTypeArrayLength0) {
```
dneto0: This next run of tests are all invalid cases.
How about some positive test cases for lengths that are constant integers: signed case, and also signed/unsigned wider than 32 bits (e.g. 48 and 64)

dneto0:
-1  for invalid code for 48 bit widths, and missing test cases.

source/validate_id.cpp
dekimir: The SPIR-V spec doesn't allow 48-bit integers -- the "Data rules" part of section 2.16.1 allows only 32-bit integers plus whatever's enabled by capabilities.  And capabilities only exist for Int64, Int16, and Int8.

dekimir:
Made `aboveZero()` DRY, covered all widths between 32 and 64, and added many more tests. PTAL.

source/validate_id.cpp
```diff
@@ -272,19 +272,15 @@ bool aboveZero(const std::vector<uint32_t>& constWords,
...
-  } else {  // Per spec, must be 32 bits or less, sign-extended.
-    if (is_signed)
-      return (constWords[3] > 0) && !(constWords[3] >> 31);
-    else
-      return constWords[3] > 0;
+  const auto loWord = constWords[3];
+  if (width > 32) {
+    // The spec currently doesn't allow integers wider than 64 bits.
+    const auto hiWord = constWords[4];  // Must exist, per spec.
+    if (is_signed && (hiWord >> 31)) return false;
```
dneto0: This depends critically on the fact that hiWord is an unsigned type, and yet you declare it via auto.
Ok, I guess.

dneto0:
-1
Fine except it fails the GCC Travis builds.

dekimir:
Fixed `auto` and GCC failure.

dneto0:
+2 yay

dekimir:
Squashed and merged to master via 3fb2676.

PR #170, opened 2016-04-01
Remove dependency on SPIR-V headers in libspirv.h.

antiagainst:
For fulfilling this purpose, the |opcode| field in the
|spv_parsed_instruction_t| struct is changed to of type uint16_t.

Also add functions to query the information of a given SPIR-V
target environment.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -27,10 +27,6 @@
 #ifndef SPIRV_TOOLS_LIBSPIRV_H_
 #define SPIRV_TOOLS_LIBSPIRV_H_
 
-#include "spirv/GLSL.std.450.h"
-#include "spirv/OpenCL.std.h"
-#include "spirv/spirv.h"
-
 #ifdef __cplusplus
 extern "C" {
 #else
@@ -40,17 +36,7 @@ extern "C" {
 #include <stddef.h>
 #include <stdint.h>
 
-// Versions
-// This library is based on SPIR-V 1.0 Rev2
-// TODO(dneto): Use the values from the SPIR-V header, when it's updated for
-// SPIR-V 1.0 public release.
-#define SPV_SPIRV_VERSION_MAJOR (SPV_VERSION >> 16)
-#define SPV_SPIRV_VERSION_MINOR (SPV_VERSION & 0xffff)
-#define SPV_SPIRV_VERSION_REVISION (SPV_REVISION)
-
 // Helpers
-
-#define spvIsInBitfield(value, bitfield) ((value) == ((value)&bitfield))
 
 #define SPV_BIT(shift) (1 << (shift))
 
@@ -287,7 +273,7 @@ typedef struct spv_parsed_instruction_t {
   const uint32_t* words;
   // The number of words in this instruction.
   uint16_t num_words;
-  SpvOp opcode;
+  uint16_t opcode;
   // The extended instruction type, if opcode is OpExtInst.  Otherwise
   // this is the "none" value.
   spv_ext_inst_type_t ext_inst_type;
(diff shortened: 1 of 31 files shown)


include/spirv-tools/libspirv.h
```diff
@@ -353,6 +339,9 @@ typedef enum {
   SPV_ENV_VULKAN_1_0_7      // Vulkan 1.0 revision 7.
 } spv_target_env;
 
+// Returns a string describing the given SPIR-V target environment.
+const char* spvTargetEnv(spv_target_env env);
```
dneto0: I think I want the name of the function to indicate that it returns a name for the given enum.
Something like  spvTargetEnvDescription

dneto0:
-1  I think this just needs a tweak on the name of the spvTargetEnv function.

include/spirv-tools/libspirv.h
antiagainst: Done.

dneto0:
+2

tools/val/val.cpp
```diff
@@ -63,8 +63,8 @@ int main(int argc, char** argv) {
     if ('-' == cur_arg[0]) {
       if (0 == strcmp(cur_arg, "--version")) {
         printf("%s\n", kBuildVersion);
-        printf("Target: SPIR-V %d.%d rev %d\n", SPV_SPIRV_VERSION_MAJOR,
-               SPV_SPIRV_VERSION_MINOR, SPV_SPIRV_VERSION_REVISION);
+        printf("Target: %s\n",
+               spvTargetEnvDescription(SPV_ENV_UNIVERSAL_1_0_4));
```
dekimir: Might want to mention the default Vulkan version here, too.

source/text.cpp
```diff
@@ -664,7 +664,7 @@ spv_result_t SetHeader(uint32_t* words, const uint32_t bound) {
 
   words[SPV_INDEX_MAGIC_NUMBER] = SpvMagicNumber;
   words[SPV_INDEX_VERSION_NUMBER] =
-      SPV_SPIRV_VERSION_WORD(SPV_SPIRV_VERSION_MAJOR, SPV_SPIRV_VERSION_MINOR);
+      spvVersionForTargetEnv(SPV_ENV_UNIVERSAL_1_0_4);
```
dekimir: This will be easy to forget when upgrading to the next SPIR-V version -- hard for maintenance.

tools/dis/dis.cpp
```diff
@@ -31,6 +31,8 @@
 
 #include "spirv-tools/libspirv.h"
 
+#define spvIsInBitfield(value, bitfield) ((value) == ((value)&bitfield))
```
dekimir: I'm not crazy about the copy-paste of identical definition from `spirv_definition.h`, but I suppose it's OK because this is a `.cpp` file.

source/text.cpp
dekimir: Might be better to add something to `Version.cpp` or `libspirv.h` that sets the default binary-output version and just reference that here.

tools/dis/dis.cpp
antiagainst: To be honest, I don't like it, too. And I found actually we don't need it. :)

tools/val/val.cpp
antiagainst: Good point. And I assume the vulkan thing is not that relevant to as and dis. So I only added it for val.

source/text.cpp
antiagainst: Good catch. 3fd8d32 passes the target env specified when creating context object into the implementation internals. It should solve this.

dekimir:
+2

PR #173, opened 2016-04-03
Enable linking against libSPIRV-Tools from linux shared libraries

chrisforbes:
Required for linking against libSPIRV-Tools from LVL's
libVkLayer_core_validation.so.


antiagainst:
LTGM

PR #174, opened 2016-04-04
Use JSON grammar files to generate extended instruction sets.

antiagainst:


utils/generate_grammar_tables.py
```diff
@@ -134,32 +134,71 @@ def __str__(self):
             ref_type_id=(1 if self.ref_type_id else 0))
 
 
-def generate_instruction(inst):
+class ExtInstInitializer(object):
+    """Prints a SPIR-V instruction as the initializer for
```
dneto0: Nitpick: 
Prints is not right here, really.  This is a class definition. Instances convert to a string suitable for printing.

dneto0:
Overall this looks good.  
The one weird thing is that the OpenCL spellings in SPIR-V assembly used lower-case names for functions instead like in the spec, but the names in OpenCL.std.h file use upper case letters.  For example SMul_hi is what's in OpenCL.std.h but the spelling we've used is s_mul_hi.

This will be a problem if we ever use the JSON file to generate the header. But we can solve it at that point, since that's a different process starting from JSON.

Sticky +2 with minor suggestion to fix the nit.  (But maybe nobody else would notice it.)

utils/generate_grammar_tables.py
antiagainst: Changed the comments.

antiagainst:
@dneto0: yeah, I agree with your thoughts on the inconsistency between the header, the spec/grammar.

And I'd prefer either snake_case or CamelCase (or consistent with the core spec), but the header is using something hybrid, which is quite strange for me.

PR #175, opened 2016-04-04
Fix Windows error C2797.

dekimir:


dekimir:
@AWoloszyn @antiagainst @dneto0 @Qining PTAL

test/ValidateID.cpp
```diff
@@ -406,7 +406,7 @@ class OpTypeArrayLengthTest
     : public spvtest::TextToBinaryTestBase<::testing::TestWithParam<int>> {
  protected:
   OpTypeArrayLengthTest()
-      : position_{0, 0, 0}, diagnostic_(spvDiagnosticCreate(&position_, "")) {}
+      : position_(), diagnostic_(spvDiagnosticCreate(&position_, "")) {}
```
AWoloszyn: Should these not be initialized to something?

dneto0: I think we should leave it in the initializer list.  Otherwise the compiler might not initialize it.  In some flows that's considered a bug, and I'd agree with it.
I believe (should look it up), using () to initialize it will mean it's zero-initialized, which is the right thing.

dneto0:
+1
Only nit is that it's not a Windows error.  It's a Microsoft *compiler* error.

test/ValidateID.cpp
AWoloszyn: I was under the impression that this will NOT zero-initialize which is why I made the comment. iirc, it will call the default-constructor, which in this case does nothing.

dneto0: Ew. You're right!  It's a POD type, and so leaves values in an indeterminate state.
So good catch!    Best to explicitly set the values in the  actual code of this constructor, then?

AWoloszyn: And reading the spec, TIL it will be zero-initialized.
8.5 
10:  - An object whose initializer is an empty set of parentheses, i.e., (), shall be value-initialized.
7: To value-initialize an object of type T means: .... if T is a (possibly cv-qualified) non-union class type without a user-provided constructor, then the object is zero-initialized and, if T’s implicitly-declared default constructor is non-trivial, that constructor is called.

AWoloszyn:
+2

dneto0:
ok. LGTM.

dekimir:
Updated to avoid relying on obscure standard paragraphs that change between versions.  Leveraged a solution recommended by the [C2797 webpage](https://msdn.microsoft.com/en-us/library/dn793970.aspx). PTAL.

AWoloszyn:
+1 Much prefer this to having to read the spec :+1:

dneto0:
+2

dekimir:
Rebased and merged to master via 7649c73.

PR #176, opened 2016-04-07
Remove the outdated patch file.

antiagainst:
Now we use the JSON grammar and the `generate_grammar_tables.py`
script to maintain various info tables. That patch file will not
be needed anymore.


dekimir:
+2

PR #177, opened 2016-04-08
Allow to only build certain grammar tables.

antiagainst:
Now we can generate tables only for the core grammar or for a specific extended instruction set.

### DIFF
--- a/utils/generate_grammar_tables.py
+++ b/utils/generate_grammar_tables.py
@@ -309,26 +309,47 @@ def main():
                         help='input JSON grammar file for core SPIR-V '
                         'instructions')
     parser.add_argument('--extinst-glsl-grammar', metavar='<path>',
-                        type=str, required=True,
+                        type=str, required=False, default=None,
                         help='input JSON grammar file for GLSL extended '
                         'instruction set')
     parser.add_argument('--extinst-opencl-grammar', metavar='<path>',
-                        type=str, required=True,
+                        type=str, required=False, default=None,
                         help='input JSON grammar file for OpenGL extended '
                         'instruction set')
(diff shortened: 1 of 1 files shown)


utils/generate_grammar_tables.py
```diff
@@ -339,20 +348,26 @@ def main():
...
+            print(generate_instruction_table(grammar['instructions'], False),
+                  file=open(args.core_insts_output, 'w'))
+        if args.operand_kinds_output is not None:
+            print(generate_operand_kind_table(grammar['operand_kinds']),
+                  file=open(args.operand_kinds_output, 'w'))
+
+    if args.extinst_glsl_grammar is not None:
+        with open(args.extinst_glsl_grammar) as json_file:
+            grammar = json.loads(json_file.read())
+            if args.glsl_insts_output is not None:
```
dekimir: Would it make more sense to check this condition right on line 358?  What's the point of loading a grammar file if no output will be produced from it?

antiagainst: Good point. Addressed in the next commit.

utils/generate_grammar_tables.py
```diff
@@ -309,27 +309,48 @@ def main():
...
+        exit(1)
+    if (args.opencl_insts_output is None) != \
+            (args.extinst_opencl_grammar is None):
+        print('error: --opencl-insts-output and --extinst-opencl-grammar '
+              'should be specified together.')
+        exit(1)
+    if all([args.core_insts_output is None,
+            args.glsl_insts_output is None,
+            args.opencl_insts_output is None]):
+        print('error: at least one output should be speciffied.')
```
dekimir: Typo: "specified"

dekimir:
+2

PR #178, opened 2016-04-11
Remove revision from SPV_ENV_*.

dekimir:
Revisions proliferate quickly and we don't promise to retain extra
grammar copies for every historical one.

@antiagainst @AWoloszyn @dneto0 @Qining PTAL


antiagainst:
+2

PR #179, opened 2016-04-12
Make paths to grammar output files

dneto0:
The directory for a grammar output file will be created if
it doesn't already exist.

This is required to make the Android.mk build work.


antiagainst:
+2

PR #180, opened 2016-04-13
Grammar update

antiagainst:
* `IdType` is renamed to `IdResultType`.
* `version` is splitted into `major_version` and `minor_version`.
* Seperate `Scope` and `IdScope` operand kinds. Same for `MemorySemantics`.
* `OpPhi` now uses `PairIdRefIdRef` as the last operand.


dneto0:
+2

PR #183, opened 2016-04-15
Rework CMake rules for building grammar tables.

antiagainst:
Use `add_custom_command()` to generate all .inc files and manage
dependency with `set_source_files_properties()`.

Tested:
* `touch` the grammar json, then `ninja` once, rebuild the `.inc` files. `ninja` twice, clean.
* `touch` `generate_grammar_tables.py`, then `ninja` once, rebuild the `.inc` files. `ninja` twice, clean.
* `touch source/print.cpp`, then `ninja` once, `print.o` got rebuilt and all things depends on it. no rebuilding of `.inc` files. `ninja` twice, clean.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -24,6 +24,84 @@
 # TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 # MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
+set(GRAMMAR_PROCESSING_SCRIPT "${spirv-tools_SOURCE_DIR}/utils/generate_grammar_tables.py")
+
+# macro() definitions are used in the following because we need to append .inc
+# file paths into some global lists (*_CPP_DEPENDS). And those global lists are
+# later used by set_source_files_properties() calls.
+# function() definitions are not suitable because they create new scopes.
+macro(spvtools_core_tables VERSION)
+  set(GRAMMAR_JSON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/spirv-${VERSION}.core.grammar.json")
+  set(GRAMMAR_INSTS_INC_FILE "${spirv-tools_BINARY_DIR}/core.insts-${VERSION}.inc")
+  set(GRAMMAR_KINDS_INC_FILE "${spirv-tools_BINARY_DIR}/operand.kinds-${VERSION}.inc")
+  add_custom_command(OUTPUT ${GRAMMAR_INSTS_INC_FILE} ${GRAMMAR_KINDS_INC_FILE}
+    COMMAND ${PYTHON_EXECUTABLE} ${GRAMMAR_PROCESSING_SCRIPT}
+      --spirv-core-grammar=${GRAMMAR_JSON_FILE}
+      --core-insts-output=${GRAMMAR_INSTS_INC_FILE}
+      --operand-kinds-output=${GRAMMAR_KINDS_INC_FILE}
+    DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${GRAMMAR_JSON_FILE}
+    COMMENT "Generate info tables for SPIR-V v${VERSION} core instructions and operands.")
+  list(APPEND OPCODE_CPP_DEPENDS ${GRAMMAR_INSTS_INC_FILE})
+  list(APPEND OPERAND_CPP_DEPENDS ${GRAMMAR_KINDS_INC_FILE})
+endmacro(spvtools_core_tables)
+
+macro(spvtools_glsl_tables VERSION)
+  set(CORE_GRAMMAR_JSON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/spirv-${VERSION}.core.grammar.json")
+  set(GLSL_GRAMMAR_JSON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/extinst-${VERSION}.glsl.std.450.grammar.json")
+  set(GRAMMAR_INC_FILE "${spirv-tools_BINARY_DIR}/glsl.std.450.insts-${VERSION}.inc")
+  add_custom_command(OUTPUT ${GRAMMAR_INC_FILE}
+    COMMAND ${PYTHON_EXECUTABLE} ${GRAMMAR_PROCESSING_SCRIPT}
+      --spirv-core-grammar=${CORE_GRAMMAR_JSON_FILE}
+      --extinst-glsl-grammar=${GLSL_GRAMMAR_JSON_FILE}
+      --glsl-insts-output=${GRAMMAR_INC_FILE}
+    DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${CORE_GRAMMAR_JSON_FILE} ${GLSL_GRAMMAR_JSON_FILE}
+    COMMENT "Generate info tables for GLSL extended instructions and operands v${VERSION}.")
+  list(APPEND EXTINST_CPP_DEPENDS ${GRAMMAR_INC_FILE})
+endmacro(spvtools_glsl_tables)
+
+macro(spvtools_opencl_tables VERSION)
+  set(CORE_GRAMMAR_JSON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/spirv-${VERSION}.core.grammar.json")
+  set(OPENCL_GRAMMAR_JSON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/extinst-${VERSION}.opencl.std.grammar.json")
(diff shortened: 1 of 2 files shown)


source/CMakeLists.txt
```diff
@@ -24,6 +24,54 @@
...
+#   every time the grammar is updated; once for generating the .inc files, the
+#   second time is ninja find that .inc is updated and then rebuild .cpp files.
+# * If we use add_custom_command() with PRE_BUILD, then the grammar processing
+#   script will always run no matter whether the grammar is updated.
+# * add_dependencies() is used to add *target* dependencies to a target.
+# * The following solution only generates .inc files when the script or the
+#   grammar files is updated, and in a single ninja run.
+set_source_files_properties(
+  ${CMAKE_CURRENT_SOURCE_DIR}/opcode.cpp
+  PROPERTIES OBJECT_DEPENDS ${GRAMMAR_INC_FILES})
```
antiagainst: I know that `opcode.cpp` doesn't depends on all `.inc` files. But I don't care of putting two more dependencies here.

source/CMakeLists.txt
```diff
@@ -24,6 +24,54 @@
...
+    --extinst-opencl-grammar=${CMAKE_CURRENT_SOURCE_DIR}/extinst.opencl.std.grammar.json
+    --core-insts-output=${spirv-tools_BINARY_DIR}/core.insts.inc
+    --glsl-insts-output=${spirv-tools_BINARY_DIR}/glsl.std.450.insts.inc
+    --opencl-insts-output=${spirv-tools_BINARY_DIR}/opencl.std.insts.inc
+    --operand-kinds-output=${spirv-tools_BINARY_DIR}/operand.kinds.inc
+  DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${GRAMMAR_JSON_FILES}
+  COMMENT "Generate info tables for SPIR-V instructions and operands." VERBATIM)
+
+# The following .cpp files include the above generated .inc files.
+# Add those .inc files as their dependencies.
```
dekimir: Why doesn't CMake auto-detect this dependency, like it does for `.h` files included from `.cpp`??

source/CMakeLists.txt
```diff
@@ -24,6 +24,54 @@
...
+#
+# Why using such an awkward way?
+# * If we use add_custom_target() to define a target to generate all .inc files
+#   and let ${SPIRV_TOOLS} depends on that, then we need to run ninja twice
+#   every time the grammar is updated; once for generating the .inc files, the
+#   second time is ninja find that .inc is updated and then rebuild .cpp files.
```
dekimir: Reword "is ninja find".

source/CMakeLists.txt
```diff
@@ -24,6 +24,54 @@
...
+# * If we use add_custom_command() with PRE_BUILD, then the grammar processing
+#   script will always run no matter whether the grammar is updated.
+# * add_dependencies() is used to add *target* dependencies to a target.
+# * The following solution only generates .inc files when the script or the
+#   grammar files is updated, and in a single ninja run.
+set_source_files_properties(
+  ${CMAKE_CURRENT_SOURCE_DIR}/opcode.cpp
+  PROPERTIES OBJECT_DEPENDS ${GRAMMAR_INC_FILES})
+set_source_files_properties(
```
dekimir: DRY, eg, by using a `foreach`.

dekimir:
-1 for question around dependency auto-detection.

source/CMakeLists.txt
antiagainst: I'm not crystally clear about how CMake manages header dependencies. So please argue if the following doesn't make sense. And I'd like to poke @AWoloszyn for expertise. :)

From my understanding, you almost specify everything. You'll need to list source files for targets, and you'll need to specify how these targets involve with each other. But it do works if you omit some header files as dependencies for targets; they are not explicitly listed but get added in those `.d` files: I think that's because CMake leverages the power of GCC/Clang. (?)

For the `.inc` files here, they are generated by a custom Python script. Although they are included similarly as typical `.h` files, but apparently the compiler knows nothing about this extension and I doubt the compiler will give any sort of help here. So, what information we end up give CMake about these `.inc` files? CMake only sees that a custom command generates some `.inc` files that get no users. So it will not generate these `.inc` files by itself. And later when the compiler searches the `-I` directories, which is given by CMake, none of these `.inc` files will be found.

source/CMakeLists.txt
antiagainst: Done.

source/CMakeLists.txt
antiagainst: Done.

antiagainst:
Rebased. It looks more complicated than before, but, works much better.

Now if we touch `generate_grammar_tables.py`, all tables will rebuild; while if we touch GLSL ExtInstSet, only GLSL ExtInstSet related tables will rebuild.

source/CMakeLists.txt
```diff
@@ -24,6 +24,80 @@
 # TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 # MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
+set(GRAMMAR_PROCESSING_SCRIPT ${spirv-tools_SOURCE_DIR}/utils/generate_grammar_tables.py)
+
+macro(spvtools_core_tables VERSION)
```
AWoloszyn: Does this have to be a macro or can it be a function?

AWoloszyn: To answer my own question, Yes, things are being added to global scope.

antiagainst: Added comments about this issue.

source/CMakeLists.txt
```diff
@@ -24,6 +24,84 @@
...
 # MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
+set(GRAMMAR_PROCESSING_SCRIPT ${spirv-tools_SOURCE_DIR}/utils/generate_grammar_tables.py)
+
+# macro() definitions are used in the following because we need to append .inc
+# file paths into some global lists (*_CPP_DEPENDS). And those global lists are
+# later used by set_source_files_properties() calls.
+# function() definitions are not suitable because they create new scopes.
+macro(spvtools_core_tables VERSION)
+  set(GRAMMAR_JSON_FILE ${CMAKE_CURRENT_SOURCE_DIR}/spirv-${VERSION}.core.grammar.json)
```
dekimir: Which scope are these names in?  Can they impact other projects when spirv-tools is added to them?

source/CMakeLists.txt
```diff
@@ -24,6 +24,84 @@
...
+  set(GRAMMAR_INSTS_INC_FILE ${spirv-tools_BINARY_DIR}/core.insts-${VERSION}.inc)
+  set(GRAMMAR_KINDS_INC_FILE ${spirv-tools_BINARY_DIR}/operand.kinds-${VERSION}.inc)
+  add_custom_command(OUTPUT ${GRAMMAR_INSTS_INC_FILE} ${GRAMMAR_KINDS_INC_FILE}
+    COMMAND ${PYTHON_EXECUTABLE} ${GRAMMAR_PROCESSING_SCRIPT}
+      --spirv-core-grammar=${GRAMMAR_JSON_FILE}
+      --core-insts-output=${GRAMMAR_INSTS_INC_FILE}
+      --operand-kinds-output=${GRAMMAR_KINDS_INC_FILE}
+    DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${GRAMMAR_JSON_FILE}
+    COMMENT "Generate info tables for SPIR-V v${VERSION} core instructions and operands.")
+  list(APPEND OPCODE_CPP_DEPENDS "${GRAMMAR_INSTS_INC_FILE}")
```
dekimir: Perhaps it would be better to put the quotes in the `GRAMMAR_INSTS_INC_FILE` definition.  When passing an argument, the CMake convention is to put quotes around something if we want to treat it as a list.

dekimir:
+1

Looks better now.  Works better, too. :)

dneto0:
+2

source/CMakeLists.txt
AWoloszyn: "Each new directory or function creates a new scope. This command will set the value of a variable into the parent directory or calling function (whichever is applicable to the case at hand). PARENT_SCOPE cannot be combined with CACHE."

AWoloszyn: So that implies that these variables are in the scope of "source/CMakeLists.txt". Unfortunately CMake is a bit vague with its useof "directory". From what I can understand from this it means
"CMakeLists.txt file referenced by a use of add_subdirectory" not "the place in the file-system that this is sitting"

antiagainst: I believe we are defining variables in the `source/` directory names here. So, should not be a great problem unless somebody include `source/CMakeLists.txt` directory.

source/CMakeLists.txt
antiagainst: Done.

PR #185, opened 2016-04-18
Update cmake files to support Android as a build target.

drWulf:
This enables cross-compiling to Android using scripts like
cmake-android that update the system name to reflect the build
target.


AWoloszyn:
Rebased and committed as 009c4358b5a1c93203166b3ed60a548f63522e81

PR #188, opened 2016-04-19
First tests for SPIR-V v1.1.

dekimir:
Add test for named-barrier instructions and capability.

Add spv_target_env as an optional argument to CompileSuccessfully() and
CompileFailure().  Currently defaults to UNIVERSAL_1_0, though that
could change in the future.

Make spv_context a local variable in test methods instead of a
TextToBinaryTestBase member.  Introduce ScopedContext to make temp
contexts easier.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -334,8 +334,8 @@ typedef spv_context_t* spv_context;
 // environment-agnostic SPIR-V.
 typedef enum {
   SPV_ENV_UNIVERSAL_1_0,  // SPIR-V 1.0 latest revision, no other restrictions.
-  SPV_ENV_UNIVERSAL_1_1,  // SPIR-V 1.1 latest revision, no other restrictions.
   SPV_ENV_VULKAN_1_0,     // Vulkan 1.0 latest revision.
+  SPV_ENV_UNIVERSAL_1_1,  // SPIR-V 1.1 latest revision, no other restrictions.
 } spv_target_env;
 
 // Returns a string describing the given SPIR-V target environment.
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -334,8 +334,8 @@ typedef spv_context_t* spv_context;
 // environment-agnostic SPIR-V.
 typedef enum {
   SPV_ENV_UNIVERSAL_1_0,  // SPIR-V 1.0 latest revision, no other restrictions.
+  SPV_ENV_UNIVERSAL_1_1,  // SPIR-V 1.1 latest revision, no other restrictions.
   SPV_ENV_VULKAN_1_0,     // Vulkan 1.0 latest revision.
-  SPV_ENV_UNIVERSAL_1_1,  // SPIR-V 1.1 any revision, no other restrictions.
 } spv_target_env;
 
 // Returns a string describing the given SPIR-V target environment.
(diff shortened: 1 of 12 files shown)


dekimir:

test/TextToBinary.Barrier.cpp
```diff
@@ -86,4 +88,101 @@ TEST_F(OpMemoryBarrier, BadInvalidMemorySemanticsId) {
...
+                             SPV_ENV_UNIVERSAL_1_1),
+              Eq("Expected id to start with %."));
+  EXPECT_THAT(CompileFailure("OpMemoryNamedBarrier %bar %scope \"semantics\"",
+                             SPV_ENV_UNIVERSAL_1_1),
+              Eq("Expected id to start with %."));
+}
+
+using TypeNamedBarrierTest = spvtest::TextToBinaryTest;
+
+TEST_F(TypeNamedBarrierTest, OpcodeRecognizedInV11) {
```
antiagainst: Will this `OpcodeRecognizedInV11` test appear for each newly added instruction in 1.1? It would be nice to have some kind of fixture to wrap the `OpcodeRecognizedInV11` & `OpcodeUnRecognizedInV10` and supply a complete list using templating.

antiagainst:
It seems clang panics because of one numeric conversion.

test/TextToBinary.Barrier.cpp
dekimir: I'm not sure how common it will be; my crystal ball is cloudy on that.  I can DRY it here, or I can create unit tests for, eg, `AssemblyGrammar::lookupOpcode()` or `spvTextEncodeOpcode()`.  I spent a little time making `spvTextEncodeOpcode()` public so it can be tested, but I quickly decided against making it public.

dneto0:
-1 build failures on Clang.

include/spirv-tools/libspirv.h
```diff
@@ -334,8 +334,8 @@ typedef spv_context_t* spv_context;
 // environment-agnostic SPIR-V.
 typedef enum {
   SPV_ENV_UNIVERSAL_1_0,  // SPIR-V 1.0 latest revision, no other restrictions.
+  SPV_ENV_UNIVERSAL_1_1,  // SPIR-V 1.1 latest revision, no other restrictions.
```
dneto0: Reordering these breaks backward compatibility.

test/TextToBinary.Barrier.cpp
antiagainst: I'm happy with the current solution; we can always refactor later.

dneto0: I think we can delay generalizing.  The proposed tests are clear and explicit.

tools/val/val.cpp
```diff
@@ -63,10 +63,9 @@ int main(int argc, char** argv) {
     if ('-' == cur_arg[0]) {
       if (0 == strcmp(cur_arg, "--version")) {
         printf("%s\n", kBuildVersion);
-        printf("Targets:\n  %s\n  %s\n  %s\n",
-               spvTargetEnvDescription(SPV_ENV_UNIVERSAL_1_0),
-               spvTargetEnvDescription(SPV_ENV_VULKAN_1_0),
-               spvTargetEnvDescription(SPV_ENV_UNIVERSAL_1_1));
+        printf("Targets:\n  %s\n  %s\n",
```
dneto0: Why drop universal 1.0?  That's a possible target.

dneto0:
-1 for
 - clang breakage
 - unnecessary backward compatibility break on enum ordering. (In fact, we should probably explicitly assign values just to emphasize the point.)
 - removal of universal 1.0 as a possible target for spirv-val.

tools/val/val.cpp
dekimir: There's currently no way to select that target.

dekimir:
Addressed feedback so far.

tools/val/val.cpp
dneto0: I see.  Vulkan is supported via --vulkan. 
For another CL, I'd like to see a --target-env option to match what glslc does.  But that's another story.

dneto0:
+2

dekimir:
Squashed and merged to master via 7b33485.

PR #189, opened 2016-04-19
Mention 1.1 support in the README.

dekimir:

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,11 @@ headers, and XML registry.
 
 ### Assembler, binary parser, and disassembler
 
-* Based on SPIR-V 1.0 Revision 3.
-  * Supports GLSL std450 extended instructions.
-  * Supports OpenCL extended instructions.
+* Based on SPIR-V version 1.1 Rev 1
+* Support for extended instruction sets:
+  * GLSL std450 version 1.0 Rev 3
+  * OpenCL version 1.0 Rev 2
+* Support for SPIR-V 1.0 (with or without additional restrictions from Vulkan 1.0)
 * Assembler only does basic syntax checking.  No cross validation of
   IDs or types is performed, except to check literal arguments to
   `OpConstant`, `OpSpecConstant`, and `OpSwitch`.


README.md
```diff
@@ -22,9 +22,10 @@ headers, and XML registry.
 
 ### Assembler, binary parser, and disassembler
 
-* Based on SPIR-V 1.0 Revision 3.
-  * Supports GLSL std450 extended instructions.
-  * Supports OpenCL extended instructions.
+* Based on SPIR-V 1.1, including extended instruction sets:
```
dneto0: The SPIR-V version is decoupled from the extended instruction sets.
Better to list the extended instruction sets separately.

Also, I'd say SPIR-V 1.1 Rev 1  since in the future we might apply fixes from future revs, and someone should be able to tell whether those fixes are applied by seeing that we've rolled to a newer Rev.

dneto0:
-1

README.md
dekimir: Done.

README.md
```diff
@@ -22,9 +22,11 @@ headers, and XML registry.
 
 ### Assembler, binary parser, and disassembler
 
-* Based on SPIR-V 1.0 Revision 3.
-  * Supports GLSL std450 extended instructions.
-  * Supports OpenCL extended instructions.
+* Based on SPIR-V 1.1 Rev 1
+* Support for extended instruction sets:
+  * GLSL std450 (v1.0)
```
dneto0: The extended instruction instruction sets have versions to.  And why write the version differently from SPIRV?
GLSL std450 is at 1.0 Rev 3
OpenCL is at 1.0 Rev 2    (even though the header file doesn't advertise it.)
https://www.khronos.org/registry/spir-v/specs/1.0/OpenCL.ExtendedInstructionSet.100.html

dneto0:
-1

README.md
dekimir: Done.

dneto0:
+2

PR #190, opened 2016-04-21
Add tests for SubgroupDispatch

dekimir:
`SubgroupDispatch` is a new feature in v1.1 adding a new capability, new execution models, and new instructions.

Extended the testing framework to allow testing with different SPIR-V versions.

### DIFF
--- a/test/OpcodeRequiresCapabilities.cpp
+++ b/test/OpcodeRequiresCapabilities.cpp
@@ -135,6 +135,8 @@ INSTANTIATE_TEST_CASE_P(
         ExpectedOpCodeCapabilities{SpvOpEmitStreamVertex,
                                    mask(SpvCapabilityGeometryStreams)},
         ExpectedOpCodeCapabilities{SpvOpTypeNamedBarrier,
-                                   mask(SpvCapabilityNamedBarrier)}), );
+                                   mask(SpvCapabilityNamedBarrier)},
+        ExpectedOpCodeCapabilities{SpvOpGetKernelMaxNumSubgroups,
+                                   mask(SpvCapabilitySubgroupDispatch)}), );
 
 }  // anonymous namespace
--- a/test/OperandCapabilities.cpp
+++ b/test/OperandCapabilities.cpp
@@ -37,12 +37,19 @@ struct EnumCapabilityCase {
   uint64_t expected_mask;
 };
 
-using EnumCapabilityTest = ::testing::TestWithParam<EnumCapabilityCase>;
+// Test fixture for testing EnumCapabilityCases.
+template <spv_target_env env>
+class EnumCapabilityTest : public ::testing::TestWithParam<EnumCapabilityCase> {
+ protected:
+  spv_target_env env_ = env;  // Target environment to use in tests.
+};
 
-TEST_P(EnumCapabilityTest, Sample) {
+using EnumCapabilityTestV10 = EnumCapabilityTest<SPV_ENV_UNIVERSAL_1_0>;
+using EnumCapabilityTestV11 = EnumCapabilityTest<SPV_ENV_UNIVERSAL_1_1>;
+
+TEST_P(EnumCapabilityTestV10, Sample) {
   spv_operand_table operandTable;
-  ASSERT_EQ(SPV_SUCCESS,
-            spvOperandTableGet(&operandTable, SPV_ENV_UNIVERSAL_1_0));
+  ASSERT_EQ(SPV_SUCCESS, spvOperandTableGet(&operandTable, env_));
   spv_operand_desc entry;
   ASSERT_EQ(SPV_SUCCESS,
             spvOperandTableValueLookup(operandTable, GetParam().type,
@@ -66,7 +73,7 @@ TEST_P(EnumCapabilityTest, Sample) {
 
 // See SPIR-V Section 3.3 Execution Model
 INSTANTIATE_TEST_CASE_P(
-    ExecutionModel, EnumCapabilityTest,
+    ExecutionModel, EnumCapabilityTestV10,
     ::testing::ValuesIn(std::vector<EnumCapabilityCase>{
         CASE1(EXECUTION_MODEL, ExecutionModelVertex, Shader),
         CASE1(EXECUTION_MODEL, ExecutionModelTessellationControl, Tessellation),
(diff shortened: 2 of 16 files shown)


test/OpcodeRequiresCapabilities.cpp
```diff
@@ -135,6 +135,8 @@ INSTANTIATE_TEST_CASE_P(
         ExpectedOpCodeCapabilities{SpvOpEmitStreamVertex,
                                    mask(SpvCapabilityGeometryStreams)},
         ExpectedOpCodeCapabilities{SpvOpTypeNamedBarrier,
-                                   mask(SpvCapabilityNamedBarrier)}), );
+                                   mask(SpvCapabilityNamedBarrier)},
+        ExpectedOpCodeCapabilities{SpvOpGetKernelMaxNumSubgroups,
```
dneto0: Also need OpGetKernelLocalSizeForSubgroupCount

test/OperandCapabilities.cpp
```diff
@@ -37,12 +37,19 @@ struct EnumCapabilityCase {
   uint64_t expected_mask;
 };
 
-using EnumCapabilityTest = ::testing::TestWithParam<EnumCapabilityCase>;
+// Test fixture for testing EnumCapabilityCases.
+template <spv_target_env env>
+class EnumCapabilityTest : public ::testing::TestWithParam<EnumCapabilityCase> {
+ protected:
+  spv_target_env env_ = env;  // Target environment to use in tests.
```
dneto0: Should be const.  

Could also be static but I don't care about that part.

test/TextToBinary.ModeSetting.cpp
```diff
@@ -140,22 +140,30 @@ TEST_F(OpEntryPointTest, WrongModel) {
...
 // Test OpExecutionMode
 
-using OpExecutionModeTest = spvtest::TextToBinaryTestBase<
-    ::testing::TestWithParam<EnumCase<SpvExecutionMode>>>;
+template <spv_target_env env>
+class OpExecutionModeTest
+    : public spvtest::TextToBinaryTestBase<
+          ::testing::TestWithParam<EnumCase<SpvExecutionMode>>> {
+ protected:
+  spv_target_env env_ = env;
```
dneto0: Should be const

test/TextToBinary.ModeSetting.cpp
```diff
@@ -190,15 +198,26 @@ INSTANTIATE_TEST_CASE_P(
...
         {CASE(ContractionOff), {}},
-    }),);
+    }), );
+
+INSTANTIATE_TEST_CASE_P(
+    TextToBinaryExecutionMode, OpExecutionModeTestV11,
+    ::testing::ValuesIn(std::vector<EnumCase<SpvExecutionMode>>{
+        {CASE(SubgroupSize), {12}},
+        {CASE(SubgroupsPerWorkgroup), {64}},
+        {CASE(LocalSize), {64, 1, 2}},
```
dneto0: Might be worth saying that in this test, LocalSize and below are spot checking features available in 1.0.
Above that line are features new to 1.1

test/TextToBinary.SubgroupDispatch.cpp
```diff
@@ -0,0 +1,138 @@
...
+using ::testing::Eq;
+using ::testing::Matcher;
+using ::testing::Property;
+using ::testing::SizeIs;
+using std::vector;
+
+// Creates a matcher of SPIR-V word vectors that matches a single instruction
+// with the given opcode and length. TODO(dekimir): move this into TestFixture
+// and DRY other tests that perform this match in longhand.
+Matcher<vector<uint32_t>> IsOpcode(SpvOp opcode, uint16_t length) {
```
dneto0: It's more than an opcode check.  It's a check for the opcode word and the instruction length.  So at least the name is wrong.

I dislike the number of layers of abstraction that you have to plough through to know what the test is testing.  That's high complexity for low payoff.
We already have many tests that check the well-formedness of the instructions when the initial word (opcode,length) is specified.  So I think you gain very little over just explicitly matching against an explicitly constructed vector.

Please don't DRY the rest of the tests in this way.

dneto0:
-1
- Missing OpGetKernelLocalSizeForSubgroupCount in OpcodeRequiresCapabilities.cpp
- Two places where const should be used
- Please avoid increases in complexity in test code that have little to no benefit.

test/TextToBinary.SubgroupDispatch.cpp
dekimir: The number of layers is actually quite low, and the benefit substantial: this lets us check a `CompiledInstructions` result (with meaningful failure messages) in the shortest, most readable, least error-prone fashion.  Other tests either have multiple lines to assert the vector length and the first element (assuming they even test both), or they rely on `ElementsAre(spvOpcodeMake(...), _, _, _, _)`, which gets cumbersome with 7-argument instructions like the ones tested here.

I'll change the name to something like `IsSingleInstruction`.

test/OpcodeRequiresCapabilities.cpp
dekimir: See comment on line 114.

dekimir:
Addressed feedback so far.

test/TextToBinary.SubgroupDispatch.cpp
```diff
@@ -0,0 +1,138 @@
...
+namespace {
+
+using ::testing::AllOf;
+using ::testing::Eq;
+using ::testing::Matcher;
+using ::testing::Property;
+using ::testing::SizeIs;
+using std::vector;
+
+// Creates a matcher of SPIR-V word vectors that matches a single instruction
```
dneto0: No reason to make this new abstraction. Most of the other tests in the suite use a matcher like:
```
  Eq(MakeInstruction(the_opcode, { the, list, of, operands }))
```
which is fully explicit and easier to read.  There are about 250 such examples.

dneto0:
-1.  Please follow the predominant pattern in the rest of the tests.

test/TextToBinary.SubgroupDispatch.cpp
dekimir: K, will leverage `MakeInstruction()` instead.

dekimir: Done.

dneto0:
+2

dekimir:
Squashed and merged to master via 2ea54f5.

PR #191, opened 2016-04-21
Avoid going out of bound in advanceLine() and refine comments.

antiagainst:
This fixes a heap-buffer-overflow problem reported by clang asan.


dekimir:
+2

PR #192, opened 2016-04-21
Fix Windows build.

dekimir:
Replace `#ifdef 0` with comment inside a macro call.


dneto0:
+2

PR #194, opened 2016-04-22
Add CHANGES file with high level software history

dneto0:
Add a high level version number for SPIRV-Tools, beginning
with v2016.0wip.  The README describes the format of the
version number.

The high level version number is extracted from the CHANGES
file.  That works around:
 - stale-bait for when we don't add tags to the repository
 - our inability to add tags to the repository

Option --version causes spirv-as, spirv-dis, and spirv-val to
show the high level version number.

Add spvSoftwareVersionString to return the C-string for
the high level version number.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,9 +1,9 @@
 Revision history for SPIRV-Tools
 
-v2016.0wip 2016-04-21
+v2016.0-dev 2016-04-22
 
- - Adds v<year>.<index> versioning, with "wip" indicating
-   work-in-progress.  The intent is to more easly report
+ - Adds v<year>.<index> versioning, with "-dev" indicating
+   work in progress.  The intent is to more easly report
    and summarize functionality when SPIRV-Tools is incorporated
    in downstream projects.
 
--- a/CHANGES
+++ b/CHANGES
@@ -1,0 +1,19 @@
+Revision history for SPIRV-Tools
+
+v2016.0wip 2016-04-21
+
+ - Adds v<year>.<index> versioning, with "wip" indicating
+   work-in-progress.  The intent is to more easly report
+   and summarize functionality when SPIRV-Tools is incorporated
+   in downstream projects.
+
+ - Summary of functionality (See the README.md for more):
+   - Supports SPIR-V 1.1 Rev 1
+   - Supports SPIR-V 1.0 Rev 5
+   - Supports GLSL std450 extended instructions 1.0 Rev 3
+   - Supports OpenCL extended instructions 1.0 Rev 2
+   - Assembler, disassembler are complete
+     - Supports floating point widths of 16, 32, 64 bits
+     - Supports integer widths up to 64 bits
+   - Validator is incomplete
+   - Supported on Linux, OSX, Android, Windows
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,21 +133,6 @@ endif()
 
 find_host_package(PythonInterp)
 
-set(SPIRV_TOOLS_BUILD_VERSION_INC
-	${CMAKE_CURRENT_BINARY_DIR}/build-version.inc)
-set(SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR
-	${CMAKE_CURRENT_SOURCE_DIR}/utils/update_build_version.py)
-set(SPIRV_TOOLS_CHANGES_FILE
-	${CMAKE_CURRENT_SOURCE_DIR}/CHANGES)
-add_custom_command(OUTPUT ${SPIRV_TOOLS_BUILD_VERSION_INC}
-   COMMAND ${PYTHON_EXECUTABLE}
-	   ${SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR}
-           ${spirv-tools_SOURCE_DIR}
-   DEPENDS ${SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR}
-           ${SPIRV_TOOLS_CHANGES_FILE}
-   COMMENT "Update build-version.inc in the SPIRV-Tools build directory (if necessary).")
-add_custom_target(spirv-tools-build-version ALL
-	DEPENDS ${SPIRV_TOOLS_BUILD_VERSION_INC})
 
 # Defaults to OFF if the user didn't set it.
 option(SPIRV_SKIP_EXECUTABLES
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,15 +133,21 @@ endif()
 
 find_host_package(PythonInterp)
 
-set(SPIRV_TOOLS_BUILD_VERSION_INC ${spirv-tools_BINARY_DIR}/build-version.inc)
-set(SPIRV_TOOLS_CHANGES_FILE ${spirv-tools_SOURCE_DIR}/CHANGES)
+set(SPIRV_TOOLS_BUILD_VERSION_INC
+	${CMAKE_CURRENT_BINARY_DIR}/build-version.inc)
+set(SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR
+	${CMAKE_CURRENT_SOURCE_DIR}/utils/update_build_version.py)
+set(SPIRV_TOOLS_CHANGES_FILE
+	${CMAKE_CURRENT_SOURCE_DIR}/CHANGES)
 add_custom_command(OUTPUT ${SPIRV_TOOLS_BUILD_VERSION_INC}
    COMMAND ${PYTHON_EXECUTABLE}
-           ${spirv-tools_SOURCE_DIR}/utils/update_build_version.py
+	   ${SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR}
            ${spirv-tools_SOURCE_DIR}
-   DEPENDS ${SPIRV_TOOLS_CHANGES_FILE}
+   DEPENDS ${SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR}
+           ${SPIRV_TOOLS_CHANGES_FILE}
    COMMENT "Update build-version.inc in the SPIRV-Tools build directory (if necessary).")
-add_custom_target(spirv-tools-build-version ALL DEPENDS ${SPIRV_TOOLS_BUILD_VERSION_INC})
+add_custom_target(spirv-tools-build-version ALL
+	DEPENDS ${SPIRV_TOOLS_BUILD_VERSION_INC})
 
 # Defaults to OFF if the user didn't set it.
 option(SPIRV_SKIP_EXECUTABLES
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,11 +133,15 @@ endif()
 
 find_host_package(PythonInterp)
 
-add_custom_target(spirv-tools-build-version
-  ${PYTHON_EXECUTABLE}
-  ${CMAKE_CURRENT_SOURCE_DIR}/utils/update_build_version.py
-  ${spirv-tools_SOURCE_DIR}
-  COMMENT "Update build-version.inc in the Spirv-tools build directory (if necessary).")
+set(SPIRV_TOOLS_BUILD_VERSION_INC ${spirv-tools_BINARY_DIR}/build-version.inc)
+set(SPIRV_TOOLS_CHANGES_FILE ${spirv-tools_SOURCE_DIR}/CHANGES)
+add_custom_command(OUTPUT ${SPIRV_TOOLS_BUILD_VERSION_INC}
+   COMMAND ${PYTHON_EXECUTABLE}
+           ${spirv-tools_SOURCE_DIR}/utils/update_build_version.py
+           ${spirv-tools_SOURCE_DIR}
+   DEPENDS ${SPIRV_TOOLS_CHANGES_FILE}
+   COMMENT "Update build-version.inc in the SPIRV-Tools build directory (if necessary).")
+add_custom_target(spirv-tools-build-version ALL DEPENDS ${SPIRV_TOOLS_BUILD_VERSION_INC})
 
 # Defaults to OFF if the user didn't set it.
 option(SPIRV_SKIP_EXECUTABLES
--- a/README.md
+++ b/README.md
@@ -18,6 +18,23 @@ SPIR-V is defined by the Khronos Group Inc.
 See the [SPIR-V Registry][spirv-registry] for the SPIR-V specification,
 headers, and XML registry.
 
+## Verisoning SPIRV-Tools
+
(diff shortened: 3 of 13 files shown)


utils/update_build_version.py
```diff
@@ -50,6 +54,23 @@ def command_output(cmd, dir):
...
 
 
+def deduceSoftwareVersion(dir):
+    """Returns a software version number parsed from the CHANGES file
+    in the given dir.
+
+    The CHANGES file describes most recent versions first.
+    """
+
+    pattern = re.compile('(v\d+\.\d+(wip)?) ')
```
AWoloszyn: If we expect the entire WIP line to be of the form

v2016.0wip 2016-04-21

Then it might be nice to change the pattern to make sure that the entire line is correct.

$(v\d+\.\d+(wip)? \d\d\d\d-\d\d-\d\d^
Also if we expect that to always be the first line after "Revision history for SPIRV-Tools" then we could error out if the second non-empty line does not have that format. It feels like that would catch mistakes in the file more easily.

README.md
```diff
@@ -18,6 +18,23 @@ SPIR-V is defined by the Khronos Group Inc.
...
+## Verisoning SPIRV-Tools
+
+See [`CHANGES`](CHANGES) for a high level summary of recent changes, by version.
+
+SPIRV-Tools project version numbers are of the form `v`*year*`.`*index* and with
+an optional `wip` suffix to indicate work in progress.  For exampe, the
+following versions are ordered from oldest to newest:
+
+* `v2016.0`
+* `v2016.1wip`
```
antiagainst: I think the convention is to call a version in progress as `rc`. This is the first time I've seen `wip` for that purpose.I'd suggest we follow the typical naming, like `2016.1-rc`.

README.md
```diff
@@ -18,6 +18,23 @@ SPIR-V is defined by the Khronos Group Inc.
 See the [SPIR-V Registry][spirv-registry] for the SPIR-V specification,
 headers, and XML registry.
 
+## Verisoning SPIRV-Tools
+
+See [`CHANGES`](CHANGES) for a high level summary of recent changes, by version.
+
+SPIRV-Tools project version numbers are of the form `v`*year*`.`*index* and with
```
antiagainst: The `v` prefix is kinda redundant for me; it stands for `version` and should not be part of the version number. (We say `version 2016.0` or `v2016.0` instead of `version v2016.0` to mean version `2016.0`.)

CHANGES
```diff
@@ -0,0 +1,19 @@
+Revision history for SPIRV-Tools
+
+v2016.0wip 2016-04-21
```
antiagainst: I'd suggest to remove the time tag on this release notes; from my understanding, at any time, there is only one `wip` "version"---the `master` branch. It will change frequently, and eventually be in a state to be tagged as the next release. A release should have a time tag, but the ongoing `master` dev branch should not, I think. Having a time tag for it just gives promise to update the time tag every time we put something new in the `master` branch.

CMakeLists.txt
```diff
@@ -133,11 +133,15 @@ endif()
...
-add_custom_target(spirv-tools-build-version
-  ${PYTHON_EXECUTABLE}
-  ${CMAKE_CURRENT_SOURCE_DIR}/utils/update_build_version.py
-  ${spirv-tools_SOURCE_DIR}
-  COMMENT "Update build-version.inc in the Spirv-tools build directory (if necessary).")
+set(SPIRV_TOOLS_BUILD_VERSION_INC ${spirv-tools_BINARY_DIR}/build-version.inc)
+set(SPIRV_TOOLS_CHANGES_FILE ${spirv-tools_SOURCE_DIR}/CHANGES)
+add_custom_command(OUTPUT ${SPIRV_TOOLS_BUILD_VERSION_INC}
+   COMMAND ${PYTHON_EXECUTABLE}
+           ${spirv-tools_SOURCE_DIR}/utils/update_build_version.py
```
antiagainst: This can just be `${CMAKE_CURRENT_SOURCE_DIR}`.

utils/update_build_version.py
dneto0: +1 to the suggestion on matching the date as well.
I was working from the conventions recommended for Perl modules:

https://metacpan.org/pod/release/BRICAS/CPAN-Changes-0.23/lib/CPAN/Changes/Spec.pod

For the preamble, it says "Any amount of text before the first Version line will be considered part of the preamble."
I should document that in the script.

source/CMakeLists.txt
```diff
@@ -135,6 +135,7 @@ set(SPIRV_SOURCES
   ${CMAKE_CURRENT_SOURCE_DIR}/opcode.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/operand.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/print.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/software_version.cpp
```
antiagainst: `Android.mk` needs to be updated for new files. ;)

CMakeLists.txt
```diff
@@ -133,11 +133,15 @@ endif()
...
+           ${spirv-tools_SOURCE_DIR}
+   DEPENDS ${SPIRV_TOOLS_CHANGES_FILE}
```
antiagainst: It should also depends on `update_build_version.py`.

README.md
dneto0: I was following the "dotted integer" convention recommend for Perl modules:
https://metacpan.org/pod/CPAN::Meta::Spec#Version-Formats
It turns out that in Perl, tokens of the form  v2016.1  are a "v-string" that you can do sensible ordering comparisons on.

But I departed from the convention, since I added the "wip" to it to indicate work-in-progress.

I think I still prefer to keep the 'v'.

README.md
dneto0: With rc I usually see a number following it too, like rc1 rc2 rc3.  I don't to promise to update numbers all the time.
I'd like a third opinion.

antiagainst: `-dev` as discussed. :)

README.md
dneto0: We settled on a "-dev" suffix instead of "wip".  That's what LLVM and Clang do.

CMakeLists.txt
dneto0: Done

source/CMakeLists.txt
dneto0: I've posted https://github.com/google/shaderc/pull/209 for this in google/shaderc

CHANGES
dneto0: When we want to finalize a release, we check in a new CHANGES file that removes the -dev suffix, then release that commit, then check in a new CHANGES with a new version number with -dev.

There is no interaction with git repo tags.

dneto0:
Responded to feedback.  I still want to keep the "-dev" string in the CHANGES file, and explained why closer to @antiagainst 's comment.

include/spirv-tools/libspirv.h
```diff
@@ -329,6 +329,16 @@ typedef spv_context_t* spv_context;
 
 // Platform API
 
+// Returns the SPIRV-Tools software version as a null-terminated string.
+// The contents of the underlying storage is valid for the remainder of
+// the process.
+const char* spvSoftwareVersionString();
```
dekimir: Nit: add a newline here.

test/SoftwareVersion.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "gmock/gmock.h"
+
+using ::testing::AnyOf;
+using ::testing::Eq;
+using ::testing::Ge;
+using ::testing::StartsWith;
+
+namespace {
+
+void CheckFormOfHighLevelVersion(std::string version) {
```
dekimir: Wouldn't `const string&` be better?

test/SoftwareVersion.cpp
```diff
@@ -0,0 +1,73 @@
...
+  std::istringstream s(version);
```
dekimir: IWYU

test/SoftwareVersion.cpp
```diff
@@ -0,0 +1,73 @@
...
+  char v = 'x';
+  int year = -1;
+  char period = 'x';
+  int index = -1;
+  s >> v >> year >> period >> index;
+  EXPECT_THAT(v, Eq('v'));
```
dekimir: Add `<< version` to all these `EXPECT`s (or use `SCOPED_TRACE` in all callers :).

dekimir: Actually, not that important: this is called with only one parameter string ever, found in `build-version.inc`.

utils/update_build_version.py
```diff
@@ -19,14 +19,18 @@
 #
 # Args: <spirv-tools_dir>
 #
-# For each directory, there will be a line in build-version.inc containing that
-# directory's "git describe" output enclosed in double quotes and appropriately
-# escaped.
+# For each directory, there will be a line in build-version.inc containing two
+# strings:
```
dekimir: "separated by a comma"?

utils/update_build_version.py
```diff
@@ -19,14 +19,18 @@
 #
 # Args: <spirv-tools_dir>
 #
-# For each directory, there will be a line in build-version.inc containing that
-# directory's "git describe" output enclosed in double quotes and appropriately
-# escaped.
+# For each directory, there will be a line in build-version.inc containing two
```
dekimir: Unclear what "each directory" refers to if there's a single argument.  (Sorry, this was my fault in the original version.)

CMakeLists.txt
```diff
@@ -133,11 +133,21 @@ endif()
...
+set(SPIRV_TOOLS_CHANGES_FILE
+	${CMAKE_CURRENT_SOURCE_DIR}/CHANGES)
+add_custom_command(OUTPUT ${SPIRV_TOOLS_BUILD_VERSION_INC}
+   COMMAND ${PYTHON_EXECUTABLE}
+	   ${SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR}
+           ${spirv-tools_SOURCE_DIR}
+   DEPENDS ${SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR}
+           ${SPIRV_TOOLS_CHANGES_FILE}
+   COMMENT "Update build-version.inc in the SPIRV-Tools build directory (if necessary).")
+add_custom_target(spirv-tools-build-version ALL
```
dekimir: I'm confused because this target is never referenced.

dneto0:
Responded to feedback.

source/CMakeLists.txt
```diff
@@ -102,6 +102,23 @@ set_source_files_properties(
   ${CMAKE_CURRENT_SOURCE_DIR}/ext_inst.cpp
   PROPERTIES OBJECT_DEPENDS "${EXTINST_CPP_DEPENDS}")
 
+set(SPIRV_TOOLS_BUILD_VERSION_INC
+	${spirv-tools_BINARY_DIR}/build-version.inc)
+message(" inc is ${SPIRV_TOOLS_BUILD_VERSION_INC}")
```
dekimir: Is this for debugging?  Why is it necessary?

source/CMakeLists.txt
```diff
@@ -102,6 +102,23 @@ set_source_files_properties(
...
+set(SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR
+	${spirv-tools_SOURCE_DIR}/utils/update_build_version.py)
+set(SPIRV_TOOLS_CHANGES_FILE
+	${spirv-tools_SOURCE_DIR}/CHANGES)
+add_custom_command(OUTPUT ${SPIRV_TOOLS_BUILD_VERSION_INC}
+   COMMAND ${PYTHON_EXECUTABLE}
+	   ${SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR}
```
dekimir: Weird indentation.

source/CMakeLists.txt
```diff
@@ -102,6 +102,23 @@ set_source_files_properties(
...
+           ${spirv-tools_SOURCE_DIR} ${SPIRV_TOOLS_BUILD_VERSION_INC}
+   DEPENDS ${SPIRV_TOOLS_BUILD_VERSION_INC_GENERATOR}
+           ${SPIRV_TOOLS_CHANGES_FILE}
+   COMMENT "Update build-version.inc in the SPIRV-Tools build directory (if necessary).")
+add_custom_target(spirv-tools-build-version
```
dekimir: Add a comment explaining that this isn't a part of any dependence chain but a convenience for standalone building of the `.inc` file.

dekimir:
-1 Almost there.

source/CMakeLists.txt
dneto0: oops.  Will remove.

source/CMakeLists.txt
dneto0: Tabs FTW.  (I mean that sarcastically.)

source/CMakeLists.txt
dneto0: Will do.

dneto0:
Responded to feedback.

utils/update_build_version.py
```diff
@@ -14,24 +14,28 @@
...
+# Updates an output file with version info unless the new content is the same
+# as the existing content.
 #
-# Args: <spirv-tools_dir>
+# Args: <spirv-tools_dir> <output-file>
 #
-# For each directory, there will be a line in build-version.inc containing that
-# directory's "git describe" output enclosed in double quotes and appropriately
-# escaped.
+# For the given directory, there will be a line in build-version.inc containing
```
dekimir: Now this just reads weird. The directory is only used to run `git describe` in it.

utils/update_build_version.py
```diff
@@ -72,15 +93,19 @@ def describe(dir):
 
 
 def main():
-    if len(sys.argv) != 2:
+    if len(sys.argv) != 3:
         print('usage: {0} <spirv-tools_dir>'.format(sys.argv[0]))
```
dekimir: Needs an update.

dekimir:
-1 Needs updated printed usage and a comment fix.

dneto0:
Responded to comments.

source/CMakeLists.txt
```diff
@@ -102,6 +102,24 @@ set_source_files_properties(
   ${CMAKE_CURRENT_SOURCE_DIR}/ext_inst.cpp
   PROPERTIES OBJECT_DEPENDS "${EXTINST_CPP_DEPENDS}")
 
+set(SPIRV_TOOLS_BUILD_VERSION_INC
+	${spirv-tools_BINARY_DIR}/build-version.inc)
```
antiagainst: Nit: I'd say this indent is kinda strange; but I don't care that much.

dekimir:
+2

dneto0:
Rebased, squashed and pushed into master as  9166854ac9

PR #195, opened 2016-04-22
Eliminate kCapabilitiesMin/Max.

dekimir:
Turn module_capabilities_ into a spv_capability_mask_t.


antiagainst:
+1

dneto0:
+2

Qining:
So I guess we won't have more than 64 capabilities?

dekimir:
@Qining: I'm sure one day we'll have more than 64, but we decided to keep it simple for now (supporting different SPIR-V versions) and react to the limit busting in the future when it occurs.

PR #197, opened 2016-04-24
WIP: Control Flow Graph Validation

umar456:
## Control-Flow Graph
- [x] Blocks exist only within a function.
- [x] The first block in a function definition is the entry point of that function and cannot be the target of any branch.
- [X] The order of blocks in a function must satisfy the rule that blocks appear before all blocks they dominate.
- [X] Each block starts with a label.
  - [X] A label is made by OpLabel.
  - [X] This includes the first block of a function (OpFunction is not a label).
  - [X] Labels are used only to form blocks.
- [X] The last instruction of each block is a branch instruction.
- [X] Branch instructions can only appear as the last instruction in a block.
- [X] OpLabel instructions can only appear within a function.
- [X] All branches within a function must be to labels in that function.

## Control-Flow declarations
- [X] the merge block declared by a header block cannot be a merge block declared by any other header block
- [X] each header block must dominate (in the CFG) its merge block, unless the merge block is unreachable in the CFG
- [ ] all CFG back edges must branch to a loop header, with each loop header having exactly one back edge branching to it
- [ ] for a given loop, its back-edge block must post dominate the OpLoopMerge's Continue Target, and that Continue Target must dominate the back-edge block

- [X] Within a function definition:
  - [X] A block always starts with an *OpLabel* instruction. This may be immediately preceded by an *OpLine* instruction, but the *OpLabel* is considered as the beginning of the block.

  - [X] A block always ends with a branch instruction.

  - [X] All *OpVariable* instructions in a function must have a Storage Class of *Function*.

  - [X] All *OpVariable* instructions in a function must be in the first block in the function. These instructions, together with any immediately preceding *OpLine* instructions, must be the first instructions in that block. 

  - [X] A function definition can be immediately preceded by an *OpLine* instruction.

## Constructs must satisfy
- [ ] the only blocks in a construct that can branch outside the construct are
   - [ ] a block branching to the construct's merge block
   - [ ] a block branching from one _case construct_ to another, for the same *OpSwitch*
   - [ ] a continue block for the innermost loop it is nested inside of
   - [ ] a break block for the innermost loop it is nested inside of
   - [ ] a return block

### DIFF
--- a/source/validate.h
+++ b/source/validate.h
@@ -35,6 +35,7 @@
 #include <unordered_set>
 #include <utility>
 #include <vector>
+#include <list>
 
 #include "assembly_grammar.h"
 #include "binary.h"
@@ -160,7 +161,7 @@ class BasicBlock {
   ///
   /// This iterator will iterate to the the immediate dominators
   /// of the block
-  class DominatorIterator {
+  class DominatorIterator : public std::iterator<std::forward_iterator_tag, BasicBlock*> {
    public:
     /// @brief Constructs the end of dominator iterator
     ///
@@ -308,9 +309,9 @@ class Function {
   std::vector<BasicBlock*>& get_blocks();
 
   /// Returns a vector of all the cfg constructs in the function
-  const std::vector<CFConstruct>& get_constructs() const;
+  const std::list<CFConstruct>& get_constructs() const;
   /// Returns a vector of all the cfg constructs in the function
-  std::vector<CFConstruct>& get_constructs();
+  std::list<CFConstruct>& get_constructs();
 
   // Returns the number of blocks in the current function being parsed
   size_t get_block_count() const;
@@ -372,7 +373,7 @@ class Function {
   BasicBlock* current_block_;
 
   /// The constructs that are available in this function
-  std::vector<CFConstruct> cfg_constructs_;
+  std::list<CFConstruct> cfg_constructs_;
 
   /// The variable IDs of the functions
   std::vector<uint32_t> variable_ids_;
@@ -426,7 +427,7 @@ class ValidationState_t {
   libspirv::DiagnosticStream diag(spv_result_t error_code) const;
 
   // Returns the function states
-  std::vector<Function>& get_functions();
+  std::list<Function>& get_functions();
 
   // Returns the function states
   Function& get_current_function();
@@ -523,7 +524,7 @@ class ValidationState_t {
   // The section of the code being processed
   ModuleLayoutSection current_layout_section_;
 
-  std::vector<Function> module_functions_;
+  std::list<Function> module_functions_;
 
   spv_capability_mask_t
       module_capabilities_;  // Module's declared capabilities.
--- a/source/validate.h
+++ b/source/validate.h
@@ -111,11 +111,15 @@ public:
 
   uint32_t get_id() const {return id_; }
 
-  const std::vector<BasicBlock*>& get_dominators() const {return dominators_; }
-  std::vector<BasicBlock*>&       get_dominators()       {return dominators_; }
+  const std::vector<BasicBlock*>& get_predecessors() const {return predecessors_; }
+  std::vector<BasicBlock*>&       get_predecessors()       {return predecessors_; }
 
   const std::vector<BasicBlock*>& get_out_blocks() const {return out_blocks_; }
   std::vector<BasicBlock*>&       get_out_blocks()       {return out_blocks_; }
+
+  void SetImmediateDominator(BasicBlock* dom_block);
+  BasicBlock* GetImmediateDominator();
+  const BasicBlock *const GetImmediateDominator() const;
 
   void RegisterNext(BasicBlock& next);
   void RegisterNext(std::vector<BasicBlock*> next);
@@ -128,7 +132,6 @@ private:
   const uint32_t id_;
   BasicBlock* immediate_dominator_;
   std::vector<BasicBlock*> predecessors_;
-  std::vector<BasicBlock*> dominators_;
   std::vector<BasicBlock*> out_blocks_;
   ValidationState_t& module_;
 };
@@ -222,6 +225,8 @@ class Functions {
   bool IsMergeBlock(uint32_t merge_block_id) const;
 
   bool isFirstBlock(uint32_t id) const;
+
+  std::vector<BasicBlock*>& get_first_blocks() { return first_blocks_; }
 
   // Returns the number of blocks in the current function being parsed
   size_t get_block_count() const;
@@ -423,6 +428,10 @@ class ValidationState_t {
   SpvMemoryModel memory_model_;
 
 };
+
+
+std::unordered_map<BasicBlock*, uint32_t>
+CalculateDominators(BasicBlock &first_block);
 
 }  // namespace libspirv
 
--- a/source/validate.h
+++ b/source/validate.h
@@ -298,13 +298,10 @@ class Function {
   spv_result_t RegisterSelectionMerge(uint32_t merge_id);
 
   /// Registers the end of the block
-  void RegisterBlockEnd(SpvOp branch_instruction);
-
-  /// Registers the end of the block
-  void RegisterBlockEnd(uint32_t next_id, SpvOp branch_instruction);
-
-  /// Registers the end of the block
-  void RegisterBlockEnd(std::vector<uint32_t> next_list,
+  ///
+  /// @param[in] successors_list A list of ids to the blocks successors
+  /// @param[in] branch_instruction the branch instruction that ended the block
+  void RegisterBlockEnd(std::vector<uint32_t> successors_list,
                         SpvOp branch_instruction);
 
   /// Returns true if the \p merge_block_id is a merge block
--- a/source/validate.h
+++ b/source/validate.h
@@ -147,14 +147,11 @@ class BasicBlock {
   /// Returns the immedate dominator of this basic block
   const BasicBlock* GetImmediateDominator() const;
 
-  /// Adds @p next as a successor of this BasicBlock
-  void RegisterBranchWithoutSuccessor(SpvOp branch_instruction);
-
-  /// Adds @p next as a successor of this BasicBlock
-  void RegisterSuccessor(BasicBlock& next, SpvOp branch_instruction);
+  /// Ends the block without a successor
+  void RegisterBranchInstruction(SpvOp branch_instruction);
 
   /// Adds @p next BasicBlocks as successors of this BasicBlock
-  void RegisterSuccessor(std::vector<BasicBlock*> next, SpvOp branch_instruction);
+  void RegisterSuccessor(std::vector<BasicBlock*> next = {});
 
   /// Returns true if the id of the BasicBlock matches
   bool operator==(const BasicBlock& other) const { return other.id_ == id_; }
@@ -248,6 +245,10 @@ class CFConstruct {
   const BasicBlock* get_merge() const { return merge_block_; }
   const BasicBlock* get_continue() const { return continue_block_; }
 
+  BasicBlock* get_header() { return header_block_; }
+  BasicBlock* get_merge() { return merge_block_; }
+  BasicBlock* get_continue() { return continue_block_; }
+
  private:
   BasicBlock* header_block_;    ///< The header block of a loop or selection
   BasicBlock* merge_block_;     ///< The merge block of a loop or selection
@@ -337,7 +338,7 @@ class Function {
   size_t get_block_count() const;
 
   /// Returns the id of the funciton
-  uint32_t get_id() const { return id_; };
+  uint32_t get_id() const { return id_; }
 
   // Returns the number of blocks in the current function being parsed
   size_t get_undefined_block_count() const;
@@ -380,7 +381,7 @@ class Function {
   /// The type of declaration of each function
   FunctionDecl declaration_type_;
 
-  /// The beginning of the block of functions
+  /// The blocks in the function mapped by block ID
   std::unordered_map<uint32_t, BasicBlock> blocks_;
 
   /// A list of blocks in the order they appeared in the binary
--- a/source/validate.h
+++ b/source/validate.h
@@ -201,26 +201,17 @@ std::ostream& operator<<(std::ostream& os, const BasicBlock &other);
 // This class manages all function declaration and definitions in a module. It
 // handles the state and id information while parsing a function in the SPIR-V
 // binary.
-//
-// NOTE: This class is designed to be a Structure of Arrays. Therefore each
-// member variable is a vector whose elements represent the values for the
-// corresponding function in a SPIR-V module. Variables that are not vector
-// types are used to manage the state while parsing the function.
-class Functions {
+class Function {
  public:
-  explicit Functions(ValidationState_t& module);
+  Function(
+           uint32_t id, uint32_t result_type_id,
+           SpvFunctionControlMask function_control,
+           uint32_t function_type_id,
+           ValidationState_t& module);
 
-  // Registers the function in the module. Subsequent instructions will be
-  // called against this function
-  spv_result_t RegisterFunction(uint32_t id, uint32_t ret_type_id,
-                                uint32_t function_control,
-                                uint32_t function_type_id);
 
   // Registers a function parameter in the current function
   spv_result_t RegisterFunctionParameter(uint32_t id, uint32_t type_id);
-
-  // Register a function end instruction
-  spv_result_t RegisterFunctionEnd();
 
   // Sets the declaration type of the current function
   spv_result_t RegisterSetFunctionDeclType(FunctionDecl type);
@@ -251,14 +242,10 @@ class Functions {
 
   bool isFirstBlock(uint32_t id) const;
 
-  std::vector<BasicBlock*>& get_first_blocks() { return first_blocks_; }
+  BasicBlock* get_first_block() { return first_block_; }
 
   // Returns the number of blocks in the current function being parsed
   size_t get_block_count() const;
-
-  // Returns true if called after a function instruction but before the
-  // function end instruction
-  bool in_function_body() const;
 
   // Returns true if called after a label instruction but before a branch
   // instruction
@@ -276,35 +263,39 @@ class Functions {
   ValidationState_t& module_;
 
   // Function IDs in a module
-  std::vector<uint32_t> id_;
+  uint32_t id_;
 
-  // OpTypeFunction IDs of each of the id_ functions
-  std::vector<uint32_t> type_id_;
+  // The type of the function
+  uint32_t function_type_id_;
+
+  // The type of the return value
+  uint32_t result_type_id_;
+
+  SpvFunctionControlMask function_control_;
 
   // The type of declaration of each function
-  std::vector<FunctionDecl> declaration_type_;
+  FunctionDecl declaration_type_;
+
 
   // TODO(umar): Probably needs better abstractions
 
   // The first BasicBlock in a function
-  std::vector<BasicBlock*> first_blocks_;
+  BasicBlock* first_block_;
 
   // The beginning of the block of functions
-  std::vector<std::unordered_map<uint32_t, BasicBlock>> blocks_;
+  std::unordered_map<uint32_t, BasicBlock> blocks_;
 
   // The block that is currently being parsed
   BasicBlock* current_block_;
 
-  std::vector<CFConstructs> cfg_constructs_;
+  CFConstructs cfg_constructs_;
 
   // The variable IDs of the functions
-  std::vector<std::vector<uint32_t>> variable_ids_;
+  std::vector<uint32_t> variable_ids_;
 
   // The function parameter ids of the functions
-  std::vector<std::vector<uint32_t>> parameter_ids_;
+  std::vector<uint32_t> parameter_ids_;
 
-  // NOTE: See correspoding getter functions
-  bool in_function_;
 };
 
 class ValidationState_t {
@@ -352,7 +343,10 @@ class ValidationState_t {
   libspirv::DiagnosticStream diag(spv_result_t error_code) const;
 
   // Returns the function states
-  Functions& get_functions();
+  std::vector<Function>& get_functions();
+
+  // Returns the function states
+  Function& get_current_function();
 
   // Returns true if the called after a function instruction but before the
   // function end instruction
@@ -401,7 +395,17 @@ class ValidationState_t {
   const std::vector<uint32_t>& entry_points() const { return entry_points_; }
 
   // Registers the capability and its dependent capabilities
-  void registerCapability(SpvCapability cap);
+  void RegisterCapability(SpvCapability cap);
+
+  // Registers the function in the module. Subsequent instructions will be
+  // called against this function
+  spv_result_t RegisterFunction(uint32_t id, uint32_t ret_type_id,
+                                SpvFunctionControlMask function_control,
+                                uint32_t function_type_id);
+
+  // Register a function end instruction
+  spv_result_t RegisterFunctionEnd();
+
 
   // Returns true if the capability is enabled in the module.
   bool hasCapability(SpvCapability cap) const;
@@ -437,7 +441,7 @@ class ValidationState_t {
   // The section of the code being processed
   ModuleLayoutSection current_layout_section_;
 
-  Functions module_functions_;
+  std::vector<Function> module_functions_;
 
   spv_capability_mask_t module_capabilities_;  // Module's declared capabilities.
 
@@ -452,6 +456,8 @@ class ValidationState_t {
   SpvAddressingModel addressing_model_;
   SpvMemoryModel memory_model_;
 
+  // NOTE: See correspoding getter functions
+  bool in_function_;
 };
 
 
@@ -482,8 +488,6 @@ void
 printDominatorList(BasicBlock &block);
 
 }  // namespace libspirv
-
-// Functions
 
 /// @brief Validate the ID usage of the instruction stream
 ///
--- a/source/validate.h
+++ b/source/validate.h
@@ -28,6 +28,7 @@
 #define LIBSPIRV_VALIDATE_H_
 
 #include <algorithm>
+#include <array>
 #include <map>
 #include <string>
 #include <unordered_map>
@@ -42,6 +43,24 @@
 #include "spirv-tools/libspirv.h"
 #include "spirv_definition.h"
 #include "table.h"
+
+#define _printf_  printf
+#define MSG(msg,...) do {                       \
+    _printf_(__FILE__":%d: " msg "\n",          \
+             __LINE__, ##__VA_ARGS__);          \
+  } while (0)
+
+#ifdef DEBUG
+#define MSG_DBG      MSG
+#else
+#define MSG_DBG(...) do { /* nothing */ } while (0)
+#endif
+
+
+#define SHOW(exp)   do { MSG("%s  %ld",   #exp, (long)exp);   } while (0)
+#define SHOW_P(ptr) do { MSG("%s  %#zx",  #ptr, (size_t)ptr); } while (0)
+#define SHOW_F(exp) do { MSG("%s  %g",    #exp, (double)exp); } while (0)
+#define SHOW_S(exp) do { MSG("%s  %s",    #exp, exp);         } while (0)
 
 // Structures
 
@@ -85,6 +104,72 @@ enum class FunctionDecl {
 
 class ValidationState_t;
 
+class BasicBlock
+{
+public:
+  BasicBlock(uint32_t id, ValidationState_t& module);
+
+  uint32_t get_id() const {return id_; }
+
+  const std::vector<BasicBlock*>& get_dominators() const {return dominators_; }
+  std::vector<BasicBlock*>&       get_dominators()       {return dominators_; }
+
+  const std::vector<BasicBlock*>& get_out_blocks() const {return out_blocks_; }
+  std::vector<BasicBlock*>&       get_out_blocks()       {return out_blocks_; }
+
+  void RegisterNext(BasicBlock& next);
+  void RegisterNext(std::vector<BasicBlock*> next);
+
+  bool operator==(const BasicBlock &other) const { return other.id_ == id_; }
+  bool operator==(const uint32_t &id) const { return id == id_; }
+
+  friend std::ostream& operator<<(std::ostream& os, const BasicBlock &other);
+private:
+  const uint32_t id_;
+  BasicBlock* immediate_dominator_;
+  std::vector<BasicBlock*> predecessors_;
+  std::vector<BasicBlock*> dominators_;
+  std::vector<BasicBlock*> out_blocks_;
+  ValidationState_t& module_;
+};
+
+class CFConstructs {
+
+// Universal Limit of ResultID + 1
+static const uint32_t kInitialValue = 0x400000;
+
+public:
+  CFConstructs(const ValidationState_t &module)
+    : module_(module)
+    , merge_blocks_()
+    , selection_header_blocks_()
+    , loop_header_blocks_()
+    , continue_blocks_()
+    , back_edges_() {}
+
+  bool IsMergeBlock(uint32_t merge_block_id) const;
+
+  spv_result_t RegisterLoopMerge(uint32_t header_id, uint32_t merge_id, uint32_t continue_id);
+
+  spv_result_t RegisterSelectionMerge(uint32_t header_id, uint32_t merge_id);
+
+
+private:
+  libspirv::DiagnosticStream diag(spv_result_t error_code) const;
+
+  const ValidationState_t &module_;
+  std::unordered_set<uint32_t> merge_blocks_;
+  std::vector<uint32_t> selection_header_blocks_;
+
+  // Loop constructs
+  std::vector<uint32_t> loop_header_blocks_;
+  std::vector<uint32_t> continue_blocks_;
+  std::vector<std::pair<uint32_t, uint32_t> > back_edges_;
+
+};
+
+std::ostream& operator<<(std::ostream& os, const BasicBlock &other);
+
 // This class manages all function declaration and definitions in a module. It
 // handles the state and id information while parsing a function in the SPIR-V
 // binary.
@@ -121,14 +206,22 @@ class Functions {
   spv_result_t RegisterBlockVariable(uint32_t type_id, uint32_t id,
                                      SpvStorageClass storage, uint32_t init_id);
 
-  spv_result_t RegisterBlockLoopMerge(uint32_t merge_id, uint32_t continue_id,
-                                      SpvLoopControlMask control);
+  spv_result_t RegisterLoopMerge(uint32_t merge_id, uint32_t continue_id);
 
-  spv_result_t RegisterBlockSelectionMerge(uint32_t merge_id,
-                                           SpvSelectionControlMask control);
+  spv_result_t RegisterSelectionMerge(uint32_t merge_id);
 
   // Registers the end of the block
   spv_result_t RegisterBlockEnd();
+
+  // Registers the end of the block
+  spv_result_t RegisterBlockEnd(uint32_t next_id);
+
+  // Registers the end of the block
+  spv_result_t RegisterBlockEnd(std::vector<uint32_t> next_list);
+
+  bool IsMergeBlock(uint32_t merge_block_id) const;
+
+  bool isFirstBlock(uint32_t id) const;
 
   // Returns the number of blocks in the current function being parsed
   size_t get_block_count() const;
@@ -141,7 +234,12 @@ class Functions {
   // instruction
   bool in_block() const;
 
-  libspirv::DiagnosticStream diag(spv_result_t error_code) const;
+        BasicBlock& get_current_block()       { return *current_block_; }
+  const BasicBlock& get_current_block() const { return *current_block_; }
+
+  void printDotGraph() const;
+
+  void printBlocks() const;
 
  private:
   // Parent module
@@ -157,8 +255,17 @@ class Functions {
   std::vector<FunctionDecl> declaration_type_;
 
   // TODO(umar): Probably needs better abstractions
+
+  // The first BasicBlock in a function
+  std::vector<BasicBlock*> first_blocks_;
+
   // The beginning of the block of functions
-  std::vector<std::vector<uint32_t>> block_ids_;
+  std::vector<std::unordered_map<uint32_t, BasicBlock>> blocks_;
+
+  // The block that is currently being parsed
+  BasicBlock* current_block_;
+
+  std::vector<CFConstructs> cfg_constructs_;
 
   // The variable IDs of the functions
   std::vector<std::vector<uint32_t>> variable_ids_;
@@ -168,7 +275,6 @@ class Functions {
 
   // NOTE: See correspoding getter functions
   bool in_function_;
-  bool in_block_;
 };
 
 class ValidationState_t {
@@ -189,6 +295,8 @@ class ValidationState_t {
   // the <id> is the numeric valid of the id and the Name is a name assigned by
   // the OpName instruction
   std::string getIdName(uint32_t id) const;
+
+  std::string getIdOrName(uint32_t id) const;
 
   // Returns the number of ID which have been forward referenced but not defined
   size_t unresolvedForwardIdCount() const;
--- a/source/validate.h
+++ b/source/validate.h
@@ -111,7 +111,6 @@ class BasicBlock {
   /// Constructor for a BasicBlock
   ///
   /// @param[in] id The ID of the basic block
-  /// @param[in] module A reference of the module of the basic block
   explicit BasicBlock(uint32_t id);
 
   /// Returns the id of the BasicBlock
@@ -151,7 +150,7 @@ class BasicBlock {
   void RegisterBranchInstruction(SpvOp branch_instruction);
 
   /// Adds @p next BasicBlocks as successors of this BasicBlock
-  void RegisterSuccessor(std::vector<BasicBlock*> next = {});
+  void RegisterSuccessors(std::vector<BasicBlock*> next = {});
 
   /// Returns true if the id of the BasicBlock matches
   bool operator==(const BasicBlock& other) const { return other.id_ == id_; }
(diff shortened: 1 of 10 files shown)


umar456:
Fixing gcc build issues.

source/validate.h
```diff
@@ -84,91 +106,280 @@ enum class FunctionDecl {
...
+class Function;
+
+// This class represents a basic block in a SPIR-V module
+class BasicBlock {
+ public:
+  /// Constructor for a BasicBlock
+  ///
+  /// @param[in] id The ID of the basic block
+  /// @param[in] module A reference of the module of the basic block
+  BasicBlock(uint32_t id, ValidationState_t& module);
```
dekimir: I'd describe what happens to the `module` reference: is it captured in the `BasicBlock` object and why isn't it `const`?

source/validate.h
```diff
@@ -84,91 +106,280 @@ enum class FunctionDecl {
...
+  /// of the block
+  class DominatorIterator : public std::iterator<std::forward_iterator_tag, BasicBlock*> {
+   public:
+    /// @brief Constructs the end of dominator iterator
+    ///
+    /// This will create an iterator which will represent the element
+    /// before the root node of the dominator tree
+    DominatorIterator();
+
+    /// @brief Constructs an iterator for the given block which points
```
dekimir: Unfinished?

source/validate.h
```diff
@@ -84,91 +106,280 @@ enum class FunctionDecl {
...
+  // The set of predecessors of the BasicBlock
+  std::vector<BasicBlock*> predecessors_;
+
+  // The set of successors of the BasicBlock
+  std::vector<BasicBlock*> successors_;
+  ValidationState_t& module_;
+  Function* function_;
+};
+
+/// @brief Returns true if the iterators point to the same element
```
dekimir: Seems odd.  I'd expect it to return true if the arguments were the same iterator (including `dom_end()`, which presumably doesn't point to anything).  If I want to check the equality of the pointed-to elements, I can always dereference the iterators.

source/validate.h
umar456: Good point. I was only using this for printing. I don't need a reference to this in the block.

source/validate.h
```diff
@@ -43,6 +45,26 @@
...
+#else
+#define MSG_DBG(...) do { /* nothing */ } while (0)
+#endif
+
+
+#define SHOW(exp)   do { MSG("%s  %ld",   #exp, (long)exp);   } while (0)
+#define SHOW_P(ptr) do { MSG("%s  %#zx",  #ptr, (size_t)ptr); } while (0)
+#define SHOW_F(exp) do { MSG("%s  %g",    #exp, (double)exp); } while (0)
+#define SHOW_S(exp) do { MSG("%s  %s",    #exp, exp);         } while (0)
+// clang-format on
```
dekimir: Just a reminder to remove the above hunk once this is ready for submission. (I presume the macros are for debugging only.)

source/validate.h
umar456: The comparison function is written that way to satisfy the [Multi-Pass guarantee of the Forward Iterator.](http://en.cppreference.com/w/cpp/concept/ForwardIterator#Multipass_guarantee) notably:

> If a and b compare equal (a == b is contextually convertible to true) then either they are both non-dereferenceable or *a and *b are references bound to the same object

source/validate.h
umar456: Do you think it would be detrimental to leave these in the code? I found these to be useful for debugging in many situations. I would like to avoid adding these every time i want to make changes to the project.

source/validate.h
dekimir: Right, so the blurb is not quite correct -- it omits the non-dereferenceable case.

source/validate.h
dekimir: Not the end of the world if they're left in, but then I really wouldn't like them to be macros.  Much nicer to have inline functions in a namespace.  Also possibly replace `printf` with `operator<<`?

source/validate.h
umar456: I see what you mean. Fixed in the latest commit.

source/validate.h
umar456: The nice thing about these macros is that it prints the file name, line number and the name of the variable in a simple line like:
```
SHOW(block_id);
```
I don't think it is as convenient to do this in a function.

dekimir: So keep the top-level macros to capture `__FILE__` and `__LINE__`, but use functions below them.  I feel that the `MSG` macros are difficult to understand. Will the variadic versions ever be invoked with no arguments (which the standard prohibits)? Will the `MSG_DBG` replacement rescan correctly to become a function-like invocation of `MSG`? I'm sure there's more questions there, and I wouldn't want to debug through it on an unfamiliar compiler.  Not to mention that macros don't typecheck their arguments and pollute scopes surrounding their invocations.

umar456: Fixed via 463ba51

source/validate.h
```diff
@@ -84,91 +103,284 @@ enum class FunctionDecl {
...
+  /// @param[in] dom_block The dominator block
+  void SetImmediateDominator(BasicBlock* dom_block);
+
+  /// Returns the immedate dominator of this basic block
+  BasicBlock* GetImmediateDominator();
+
+  /// Returns the immedate dominator of this basic block
+  const BasicBlock* GetImmediateDominator() const;
+
+  /// Sets the @p next BasicBlock as the successor of this BasicBlock
```
dekimir: Better to say "a successor"?

source/validate.h
```diff
@@ -84,91 +103,284 @@ enum class FunctionDecl {
...
+  void RegisterSuccessor(BasicBlock& next);
```
dekimir: Does this add a successor or drop previously registered one(s)?  If adding, perhaps `Add` should be in the name.

source/validate.h
```diff
@@ -84,91 +103,284 @@ enum class FunctionDecl {
...
+
+  /// Sets the @p next BasicBlock as the successor of this BasicBlock
```
dekimir: This blurb should use plural, I think.

source/validate.h
```diff
@@ -84,91 +103,284 @@ enum class FunctionDecl {
...
+
+  /// Returns true if the id of the BasicBlock matches
+  bool operator==(const BasicBlock& other) const { return other.id_ == id_; }
+
+  /// Returns true if the id of the BasicBlock matches
+  bool operator==(const uint32_t& id) const { return id == id_; }
+
+  /// @brief A BasicBlock dominator iterator class
+  ///
+  /// This iterator will iterate to the the immediate dominators
```
dekimir: s/to the the/over/

source/validate.h
dekimir: I guess "RegisterFoo" is a widely used naming pattern here and elsewhere, so never mind.

source/validate.h
```diff
@@ -84,91 +103,284 @@ enum class FunctionDecl {
...
+  /// Returns the first block of the current function
+  BasicBlock* get_first_block();
+
+  /// Returns a vector of all the blocks in the function
+  const std::vector<BasicBlock*>& get_blocks() const;
+
+  /// Returns a vector of all the blocks in the function
+  std::vector<BasicBlock*>& get_blocks();
+
+  /// Returns a vector of all the cfg constructs in the function
```
dekimir: Doesn't return a vector. ;)

source/validate_cfg.cpp
```diff
@@ -24,32 +24,294 @@
...
+using stack_info = tuple<cbb_ptr, bb_iter, bb_iter>;
+
+stack_info CreateStack(const BasicBlock &block) {
+  return make_tuple(&block, begin(block.get_successors()),
+                    end(block.get_successors()));
+}
+
+template <typename T>
+bool Contains(const vector<T> &vec, T val) {
+  return find(begin(vec), end(vec), val) == end(vec);
```
dekimir: Am I missing something?  I would have expected `!= end`.

umar456: Ahh, yes indeed. I need to fix this.

source/validate_cfg.cpp
```diff
@@ -24,32 +24,294 @@
...
 
-// TODO(umar): Support for merge instructions
-// TODO(umar): Structured control flow checks
+namespace {
+
+using bb_ptr = BasicBlock *;
+using cbb_ptr = const BasicBlock *;
+using bb_iter = vector<BasicBlock *>::const_iterator;
+enum                    {kBlock, kIter  , kEnd };
+using stack_info = tuple<cbb_ptr, bb_iter, bb_iter>;
```
dekimir: I'd give this type a different name, describing what it _is_, rather than what one specific client uses it for.

source/validate_cfg.cpp
```diff
@@ -24,32 +24,294 @@
...
+
+stack_info CreateStack(const BasicBlock &block) {
```
dekimir: Misnamed: this creates a `stack_info`, not a stack.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,579 @@
...
+using ::testing::HasSubstr;
+
+using libspirv::BasicBlock;
+using libspirv::ValidationState_t;
+
+using ValidateCFG = spvtest::ValidateBase<bool>;
+using spvtest::ScopedContext;
+
+namespace libspirv {
+vector<const BasicBlock *> PostOrderSort(const BasicBlock &entry,
```
dekimir: Looks like this is testing non-public interfaces, a sign of strained design.  Why not promote `PostOrderSort()` to a reusable public method?  If not, then why bake this private detail into tests?

source/validate_cfg.cpp
```diff
@@ -24,32 +24,294 @@
...
+
+template <typename T>
+bool Contains(const vector<T> &vec, T val) {
+  return find(begin(vec), end(vec), val) == end(vec);
+}
+
+}
+
+vector<const BasicBlock*>
+PostOrderSort(const BasicBlock &entry, size_t size) {
```
dekimir: Blurb needed.

source/validate_cfg.cpp
```diff
@@ -24,32 +24,294 @@
...
+
+using bb_ptr = BasicBlock *;
+using cbb_ptr = const BasicBlock *;
+using bb_iter = vector<BasicBlock *>::const_iterator;
+enum                    {kBlock, kIter  , kEnd };
+using stack_info = tuple<cbb_ptr, bb_iter, bb_iter>;
+
+stack_info CreateStack(const BasicBlock &block) {
+  return make_tuple(&block, begin(block.get_successors()),
+                    end(block.get_successors()));
```
dekimir: Would it be much more expensive to store subscripts instead of iterators?  It would certainly be safer.

At least we should say in the tuple description that the `BasicBlock`s it captures mustn't change during the tuple's lifetime.

test/Validate.CFG.cpp
umar456: I created the tests when I was developing the PostOrderSort function. I didn't want to add PostOrderSort to the public API because I didn't think it belongs in the SPIRV validation library. I can remove the tests if you feel that we shouldn't do something like this.

source/validate_cfg.cpp
umar456: Wouldn't passing the block as a const reference imply that the successor iterators should be valid for the tuple's lifetime. That was my thought process anyway.

umar456: On second thought I guess it doesn't. I will note that in the function.

dekimir: Yeah, it's not the `const&` of the parameter as much as the fact that the tuple captures a pointer to it.  Speaking of, the Google style guide [suggests](https://google.github.io/styleguide/cppguide.html#Reference_Arguments) passing a `const*` in such situations, as an (imperfect) convention to signal that something unusual is going on.

But even that only suggests that the `BasicBlock`'s lifetime must outlast the tuple; we need an even stronger restriction here: the `BasicBlock` must remain alive _and_ its successors vector must not change during the tuple's lifetime.

test/Validate.CFG.cpp
dekimir: It's not the end of the world to do this, but it's a distinctly third option in order of preference:

1. Make `PostOrderSort` public, give it a proper blurb, and design it so that it's reusable in other contexts.  Keep the tests, as we're now committed to this (very reasonable) breakdown of functionality.

2. Keep `PostOrderSort` private and remove its tests, but ensure that it actually has 100% coverage via the public-interface test cases.  That way we can change the internal implementation tomorrow without invalidating any tests.

3. What's there now. :)

source/validate_cfg.cpp
dekimir: Before fixing it, please add a test case that fails because of this bug.  I'm rather surprised there isn't one currently.

umar456: The bug is really in the name and not in the logic. It should have been called DoesNotContain :)

source/validate_cfg.cpp
umar456: @dekimir Would removing the function and performing the tuple creation inline alleviate the iterator issues? I just created that function to make the code more readable, but as you suggested it's interface opens the door to incorrect usage.

source/validate.h
```diff
@@ -84,91 +103,293 @@ enum class FunctionDecl {
...
+  void printBlocks() const;
 
  private:
-  // Parent module
+  /// Parent module
   ValidationState_t& module_;
 
-  // Function IDs in a module
-  std::vector<uint32_t> id_;
+  /// Function ID of
```
dekimir: Dangling "of".

source/validate.h
```diff
@@ -84,91 +103,293 @@ enum class FunctionDecl {
...
+  uint32_t id_;
 
-  // OpTypeFunction IDs of each of the id_ functions
-  std::vector<uint32_t> type_id_;
+  /// The type of the function
```
dekimir: Does this refer to return type?  If so, add "return" to "type".

umar456: This refers to the OpTypeFunction id for this function. The result_type_id is the OpType* of the return type.

source/validate_cfg.cpp
```diff
@@ -24,32 +24,314 @@
...
+///
+/// Returns a vector of basic block pointers in a Control Flow Graph(CFG) which
+/// are sorted in the order they were accessed in a post order traversal.
+///
+/// @param[in] entry the first block of a CFG
+/// @param[in] depth_hint a hint about the depth of the CFG
+///
+/// @return A vector of pointers in the order they were access in a post order
+/// traversal
+vector<const BasicBlock*> PostOrderSort(const BasicBlock& entry, size_t size) {
```
dekimir: This implementation is way more complex than I'd expect.  Here's a working-code example of the post-order sorting logic:

```
struct Node {
  int val;
  vector<const Node*> children;
};

void append_in_post_order(const Node& node, vector<const Node*>& seq) {
  for (const auto& c : node.children) append_in_post_order(*c, seq);
  seq.push_back(&node);
}

vector<const Node*> post_order(const Node& root) {
  vector<const Node*> seq;
  append_in_post_order(root, seq);
  return seq;
}
```

Why can't we use the above?

dekimir: Minor correction: the above works in a tree, not a DAG.  But it's trivial to adapt to a DAG with a simple "visited" bit for each node.

source/validate_cfg.cpp
```diff
@@ -24,32 +24,314 @@
...
+
+vector<pair<BasicBlock*, BasicBlock*>> CalculateDominators(
+    const BasicBlock& first_block) {
+  struct block_detail {
+    size_t dominator;
+    size_t postorder_index;
+  };
+
+  vector<cbb_ptr> postorder = PostOrderSort(first_block, 10);
+  const size_t undefined_dom = static_cast<uint32_t>(postorder.size());
```
dekimir: Why cast to `uint32_t` and not to `size_t`?

source/validate_cfg.cpp
```diff
@@ -24,32 +24,314 @@
...
+  bool changed = true;
+  while (changed) {
+    changed = false;
+    for (auto b = postorder.rbegin() + 1; b != postorder.rend(); b++) {
+      // printf("processing: %d\n", (*b)->get_id());
+      size_t& b_dom = idoms[*b].dominator;
+      const vector<BasicBlock*>& predecessors = (*b)->get_predecessors();
+
+      // first processed predecessor
+      BasicBlock* idom =
```
dekimir: Is this guaranteed not to return `end(predecessors)`?  If so, there should be an assertion.

source/validate_cfg.cpp
```diff
@@ -24,32 +24,314 @@
...
+            }
+            while (finger2 < finger1) {
+              finger2 = idoms[postorder[finger2]].dominator;
+            }
+          }
+          idom_idx = finger1;
+        }
+      }
+      if (b_dom != idom_idx) {
+        b_dom = idom_idx;
```
dekimir: This assignment seems dead code.

source/validate_cfg.cpp
umar456: I was trying to avoid recursion as we don't know how deep the user's code could nest. I guess we can rely on the compiler to perform tail call optimization. I am not partial to either approach. Your call.

dekimir: I'd say use recursion, for readability.  If a user reports a problem with stack busting, then we can think about replacing it with iteration.  But I doubt that's likely.

(It won't be tail recursion, unfortunately, because post-order means more work after the recursive call has returned.)

umar456: Hmm, thats true. In that case would like to avoid the recursion version. I would rather avoid a problem now rather then wait for things to break. I don't think its significantly less readable then your version. I was able to simplify the algorithm further:

```cpp
vector<const BasicBlock*> PostOrderSort(const BasicBlock& entry, size_t size) {
  struct block_info {
    cbb_ptr block;
    bb_iter iter;
  };

  vector<cbb_ptr> out;
  vector<block_info> staged;
  unordered_set<uint32_t> processed;

  staged.reserve(size);
  staged.emplace_back(block_info{&entry, begin(entry.get_successors())});
  processed.insert(entry.get_id());

  while (!staged.empty()) {
    block_info& top = staged.back();
    if (top.iter == end(top.block->get_successors())) {
      out.push_back(top.block);
      staged.pop_back();
    } else {
      BasicBlock* child = *top.iter;
      if (processed.find(child->get_id()) == end(processed)) {
        staged.emplace_back(block_info{child, begin(child->get_successors())});
        processed.insert(child->get_id());
      }
      top.iter++;
    }
  }
  return out;
}
```
It should be slightly more readable then the previous version.

source/validate_cfg.cpp
umar456: This line updates the reference to idoms[*b].dominator. it is not an assignment to the local variable.

umar456:
Addressed feedback so far

source/validate.h
```diff
@@ -84,91 +103,303 @@ enum class FunctionDecl {
...
+  SpvFunctionControlMask function_control_;
 
-  // The variable IDs of the functions
-  std::vector<std::vector<uint32_t>> variable_ids_;
+  /// The type of declaration of each function
+  FunctionDecl declaration_type_;
 
-  // The function parameter ids of the functions
-  std::vector<std::vector<uint32_t>> parameter_ids_;
+  /// The beginning of the block of functions
```
dekimir: Not sure what this means.

source/validate.h
```diff
@@ -84,91 +103,303 @@ enum class FunctionDecl {
...
+  std::unordered_map<uint32_t, BasicBlock> blocks_;
 
-  // NOTE: See correspoding getter functions
-  bool in_function_;
-  bool in_block_;
+  /// A list of blocks in the order they appeared in the binary
+  std::vector<BasicBlock*> ordered_blocks_;
+
+  /// Blocks which are forward referenced by blocks but not defined
```
dekimir: Unless this is always up-to-date, I'd keep it out of the class and with the code that actually calculates it.

dekimir: Never mind, I see below that it is always kept up-to-date. :)

source/validate_cfg.cpp
```diff
@@ -24,32 +24,303 @@
...
+    cbb_ptr block;
+    bb_iter iter;
+  };
+
+  vector<cbb_ptr> out;
+  vector<block_info> staged;
+  unordered_set<uint32_t> processed;
+
+  staged.reserve(size);
+  staged.emplace_back(block_info{&entry, begin(entry.get_successors())});
```
dekimir: Why is `block_info{` necessary here?  Doesn't `emplace_back` do exactly that for you?

source/validate_cfg.cpp
```diff
@@ -24,32 +24,303 @@
...
+    }
+  }
+  return out;
+}
+}
+
+vector<pair<BasicBlock*, BasicBlock*>> CalculateDominators(
+    const BasicBlock& first_block) {
+  struct block_detail {
+    size_t dominator;
```
dekimir: I'd add a blurb describing what the `dominator` number means.

source/validate_cfg.cpp
```diff
@@ -24,32 +24,303 @@
...
+void printDominatorList(BasicBlock& b) {
+  std::cout << b.get_id() << " is dominated by: ";
+  const BasicBlock* bb = &b;
+  while (bb->GetImmediateDominator() != bb) {
+    bb = bb->GetImmediateDominator();
+    std::cout << bb->get_id() << " ";
+  }
+}
+
+#define CFG_ASSERT(ASSERT_FUNC, TARGET) \
```
dekimir: Should be `#undef`ed below after use.

source/validate_cfg.cpp
umar456: I thought so too but Apple clang was complaining about it. I think it needs a constructor to work correctly.

source/validate_types.cpp
```diff
@@ -335,36 +359,54 @@ SpvMemoryModel ValidationState_t::getMemoryModel() const {
...
 
-spv_result_t Functions::RegisterFunction(uint32_t id, uint32_t ret_type_id,
-                                         uint32_t function_control,
-                                         uint32_t function_type_id) {
+spv_result_t ValidationState_t::RegisterFunction(
+    uint32_t id, uint32_t ret_type_id, SpvFunctionControlMask function_control,
+    uint32_t function_type_id) {
+  assert(in_function_ == false &&
+         "Function instructions can not be declared in a function");
   assert(in_function_ == false &&
```
dekimir: Duplicate of the above?

source/validate_types.cpp
```diff
@@ -335,36 +359,54 @@ SpvMemoryModel ValidationState_t::getMemoryModel() const {
...
+}
 
-spv_result_t Functions::RegisterFunction(uint32_t id, uint32_t ret_type_id,
-                                         uint32_t function_control,
-                                         uint32_t function_type_id) {
+spv_result_t ValidationState_t::RegisterFunction(
+    uint32_t id, uint32_t ret_type_id, SpvFunctionControlMask function_control,
+    uint32_t function_type_id) {
+  assert(in_function_ == false &&
+         "Function instructions can not be declared in a function");
```
dekimir: This reads like a diagnostic, so either this shouldn't be an `assert` or the message should describe an internal error.

source/validate_types.cpp
```diff
@@ -335,36 +359,54 @@ SpvMemoryModel ValidationState_t::getMemoryModel() const {
...
   return SPV_SUCCESS;
 }
 
-spv_result_t Functions::RegisterFunctionParameter(uint32_t id,
-                                                  uint32_t type_id) {
-  assert(in_function_ == true &&
+spv_result_t ValidationState_t::RegisterFunctionEnd() {
+  assert(in_function_body() == true &&
+         "Function end can only be called in functions");
+  assert(in_block() == false && "Function end cannot be called inside a block");
```
dekimir: Either change the message to an internal-error indication or replace `assert` with a real check.

dekimir: On second thought, this could go into a separate PR that systematically addresses it everywhere.

source/validate_types.cpp
```diff
@@ -376,42 +418,261 @@ spv_result_t Functions::RegisterFunctionParameter(uint32_t id,
...
   assert(module_.getLayoutSection() !=
              ModuleLayoutSection::kLayoutFunctionDeclarations &&
          "Function declartions must appear before function definitions");
-  assert(declaration_type_.back() == FunctionDecl::kFunctionDeclDefinition &&
+  assert(declaration_type_ == FunctionDecl::kFunctionDeclDefinition &&
          "Function declaration type should have already been defined");
 
-  block_ids_.back().push_back(id);
-  in_block_ = true;
+  std::unordered_map<uint32_t, BasicBlock>::iterator tmp;
```
dekimir: Is it possible to have a more descriptive name than `tmp`?

source/validate_types.cpp
umar456: The asserts are sanity checks. 

I tried to design everything in the validate_type class to only track the state of the validation. The checks are being performed in the other validate_[cfg,layout,ssa].cpp files.These checks should be performed by the callers in those files.

dekimir: In that case, I find the assert messages confusing.  They read as if there's a problem with the SPIR-V module and not the code's own contract.

source/validate_cfg.cpp
umar456: this is in a cpp file. I will probably use this throughout the file so the undef will most likely be on the last line. I can always expand the macro but I though this was easier to read.

source/validate_types.cpp
```diff
@@ -376,42 +418,261 @@ spv_result_t Functions::RegisterFunctionParameter(uint32_t id,
...
+  bool success = false;
+  tie(tmp, success) = blocks_.insert({id, BasicBlock(id)});
+  if (is_definition) {  // new block definition
+    assert(in_block() == false && "Blocks cannot be nested");
+
+    undefined_blocks_.erase(id);
+    current_block_ = &tmp->second;
+    if(ordered_blocks_.empty()) current_block_->set_reachability(true);
```
dekimir: Isn't there something like `IsFirstBlock()`?

source/validate_types.cpp
umar456: I see what you mean. Can you suggest alternate wording? Here is what can think of:

1. RegisterFunction can only be called when parsing the binary outside of another function.
2. RegisterFunction expects to be called after the RegisterFunctionEnd or as the first function of a module

source/validate.h
```diff
@@ -84,91 +103,303 @@ enum class FunctionDecl {
...
+  /// @param[in] dom_block The dominator block
+  void SetImmediateDominator(BasicBlock* dom_block);
+
+  /// Returns the immedate dominator of this basic block
+  BasicBlock* GetImmediateDominator();
+
+  /// Returns the immedate dominator of this basic block
+  const BasicBlock* GetImmediateDominator() const;
+
+  /// Adds @p next as a successor of this BasicBlock
```
dekimir: Not sure what `next` refers to?  And why does the blurb mention a successor in a `WithoutSuccessor` method?

source/validate.h
```diff
@@ -84,91 +103,303 @@ enum class FunctionDecl {
...
+  void RegisterBranchWithoutSuccessor(SpvOp branch_instruction);
+
+  /// Adds @p next as a successor of this BasicBlock
+  void RegisterSuccessor(BasicBlock& next, SpvOp branch_instruction);
```
dekimir: I'd keep the branch-instruction registration separate, for a normalized interface.  That would also allow for DRYing `RegisterBlockEnd()` with a single overload like this:
```
void RegisterBlockEnd(SpvOp branch_instruction, std::vector<uint32_t> successors = {});
```

source/validate_types.cpp
dekimir: Number 1 sounds good to me.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  return "OpName %" + head.first + " \"" + head.second + "\"\n" +
+         nameOps(names...);
+}
+
+template <typename... Args>
+string nameOps(string head, Args... names) {
+  return "OpName %" + head + " \"" + head + "\"\n" + nameOps(names...);
+}
+
+class Block {
```
dekimir: Needs a blurb and a justification for existing besides `BasicBlock` above.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  lhs.successors_.push_back(successor);
+  return lhs;
+}
+
+string header =
+    "OpCapability Shader\n"
+    "OpMemoryModel Logical GLSL450\n";
+
+string types_consts =
+    nameOps("voidt", "boolt", "intt", "one", "two", "ptrt", "funct") +
```
dekimir: It's not obvious why this is necessary -- isn't `OpName` an instruction without any effects?

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+    }
+    out << "\n";
+
+    return out.str();
+  }
+  friend Block &operator>>(Block &curr, vector<Block> successors);
+  friend Block &operator>>(Block &lhs, Block &successor);
+};
+
+Block &operator>>(Block &lhs, vector<Block> successors) {
```
dekimir: Blurb needed.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  if (lhs.type_ == SpvOpBranchConditional) {
+    assert(successors.size() == 2);
+  } else if (lhs.type_ == SpvOpSwitch) {
+    assert(successors.size() > 1);
+  }
+  lhs.successors_ = successors;
+  return lhs;
+}
+
+Block &operator>>(Block &lhs, Block &successor) {
```
dekimir: Blurb needed.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+    nameOps("voidt", "boolt", "intt", "one", "two", "ptrt", "funct") +
+    "%voidt   = OpTypeVoid\n"
+    "%boolt   = OpTypeBool\n"
+    "%intt    = OpTypeInt 32 1\n"
+    "%one     = OpConstant %intt 1\n"
+    "%two     = OpConstant %intt 2\n"
+    "%ptrt    = OpTypePointer Function %intt\n"
+    "%funct   = OpTypeFunction %voidt\n";
+
+TEST_F(ValidateCFG, Default) {
```
dekimir: Not really sure what `Default` as a test-case name refers to.  Is there a more descriptive name?

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  Block cont("cont");
+  Block merge("merge");
+  Block end("end", SpvOpReturn);
+
+  // cannot share the same merge
+  loop.setBody(
+      " %cond2   = OpSLessThan %intt %one %two\n"
+      " OpLoopMerge %merge %cont None\n");
+  badhead.setBody(
+      " %cond1   = OpSLessThan %intt %one %two\n"
```
dekimir: Doesn't this leave `%cond` undefined when `OpBranchConditional` references it?

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  str += t >> merge;
+  str += f >> cont;
+  str += cont >> loop;
+  str += merge >> end;
+  str += end;
+  str += "OpFunctionEnd\n";
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_CFG, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(), HasSubstr("merge"));
```
dekimir: This substring is too vague.  Isn't there a way to assert with higher certainty that the validation fails specifically because of `%merge` being reused by multiple header blocks?

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  Block badhead("badhead", SpvOpBranchConditional);
+  Block t("t");
+  Block f("f");
+  Block cont("cont");
+  Block merge("merge");
+  Block end("end", SpvOpReturn);
+
+  // cannot share the same merge
+  loop.setBody(
+      " %cond2   = OpSLessThan %intt %one %two\n"
```
dekimir: I don't understand why this is necessary; isn't the this block's branch unconditional?

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+
+  string str = header + nameOps("merge", make_pair("func", "Main")) + types_consts +
+               "%func    = OpFunction %voidt None %funct\n";
+
+  str += entry >> badhead;
+  str += badhead >> vector<Block>({t, f});
+  str += t >> merge;
+  str += f >> cont;
+  str += cont >> loop;
+  str += loop >> merge;
```
dekimir: `loop` is misnamed, as there are no back edges here.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  str += bad >> entry;  // Cannot target entry block
+  str += end;
+  str += "OpFunctionEnd\n";
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_CFG, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(), HasSubstr("First block"));
+  EXPECT_THAT(getDiagnosticString(), HasSubstr("entry"));
+  EXPECT_THAT(getDiagnosticString(), HasSubstr("Main"));
+  EXPECT_THAT(getDiagnosticString(), HasSubstr("bad"));
```
dekimir: I'd use `MatchesRegex` instead of four (possibly disconnected) `HasSubstr`.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+}
+
+TEST_F(ValidateCFG, BranchConditionalTrueTargetFirstBlockBad) {
+  Block entry("entry");
+  Block bad("bad", SpvOpBranchConditional);
+  Block f("f");
+  Block merge("merge");
```
dekimir: This block seems superfluous for this test case.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+      " %cond    = OpSLessThan %intt %one %two\n"
+      "OpSelectionMerge %end None\n");
+
+  string str = header + nameOps("bad", "merge", make_pair("func", "Main")) +
+               types_consts + "%func    = OpFunction %voidt None %funct\n";
+
+  str += entry >> merge;
+  str += merge >> vector<Block>({bad, end});
+  str += bad >> vector<Block>({cont, f});
+  str += cont >> merge;
```
dekimir: This is a back edge not going into a loop header.  In the future, this (instead of merge domination) may trigger the validation failure and make this test fail despite the code working correctly.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  ASSERT_EQ(SPV_SUCCESS, ValidateInstructions());
+}
+
+TEST_F(ValidateCFG, UnreachableMergeInstruction) {
+  Block entry("entry");
+  Block branch("branch", SpvOpBranchConditional);
+  Block t("t", SpvOpReturn);
+  Block f("f", SpvOpReturn);
+  Block merge("merge", SpvOpUnreachable);
+  Block end("end", SpvOpReturn);
```
dekimir: This block seems redundant to this test case.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  str += f;
+  str += merge >> end;
+  str += end;
+  str += "OpFunctionEnd\n";
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_SUCCESS, ValidateInstructions());
+}
+
+TEST_F(ValidateCFG, UnreachableMergeInstruction) {
```
dekimir: The test name is not terribly suggestive of what's being tested, especially when compared to the previous one.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,601 @@
...
+  str += unreachablechildt >> merge;
+  str += unreachablechildf >> merge;
+  str += merge >> exit;
+  str += exit;
+  str += "OpFunctionEnd\n";
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_SUCCESS, ValidateInstructions());
+}
+
```
dekimir: Add (a TODO for?) more test cases: empty function, single-block loop, nested loops, nested selections.

umar456:
Addressed feedback so far

umar456:
@dekimir I am working on additional changes to the CFG validation. Should I add it to this branch or make a different PR for those?

dekimir:
@umar456 please don't add anything to this branch, as I'm about to approve and merge it.  Sorry for the delay, I've been swamped with other work. :(

umar456:
No problem. Just wanted to know how you wanted to handle additional changes.

source/validate.h
```diff
@@ -84,91 +103,302 @@ enum class FunctionDecl {
...
 class ValidationState_t;
+class Function;
+
+// This class represents a basic block in a SPIR-V module
+class BasicBlock {
+ public:
+  /// Constructor for a BasicBlock
+  ///
+  /// @param[in] id The ID of the basic block
+  /// @param[in] module A reference of the module of the basic block
```
dekimir: Outdated?

source/validate.h
```diff
@@ -84,91 +103,302 @@ enum class FunctionDecl {
...
+  BasicBlock* GetImmediateDominator();
+
+  /// Returns the immedate dominator of this basic block
+  const BasicBlock* GetImmediateDominator() const;
+
+  /// Ends the block without a successor
+  void RegisterBranchInstruction(SpvOp branch_instruction);
+
+  /// Adds @p next BasicBlocks as successors of this BasicBlock
+  void RegisterSuccessor(std::vector<BasicBlock*> next = {});
```
dekimir: Nitpick: consider plural RegisterSuccessor**s**.

source/validate.h
```diff
@@ -190,6 +420,8 @@ class ValidationState_t {
   // the OpName instruction
   std::string getIdName(uint32_t id) const;
 
+  std::string getIdOrName(uint32_t id) const;
```
dekimir: Blurb needed.

source/validate.h
```diff
@@ -84,91 +103,302 @@ enum class FunctionDecl {
...
+  /// @param[in] continue_id The continue block ID of the loop
+  ///
+  /// @return Returns SPV_SUCCESS if the call was successful
+  spv_result_t RegisterLoopMerge(uint32_t merge_id, uint32_t continue_id);
+
+  /// Registers a selection merge construct in the function
+  /// @return Returns SPV_SUCCESS if the call was successful
+  spv_result_t RegisterSelectionMerge(uint32_t merge_id);
+
+  /// Registers the end of the block
```
dekimir: The three overloads seem unnecessary; the most general case should be enough.

umar456: Agreed. Removed

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,629 @@
...
+  loop_head.setBody("OpLoopMerge %loop_merge %branch None\n");
+
+  string str = header + nameOps("loop_head", "loop_merge", make_pair("func", "Main")) + types_consts +
+               "%func    = OpFunction %voidt None %funct\n";
+
+  str += entry >> vector<Block>({loop_head, loop_merge});
+  str += loop_head >> branch;
+  str += branch >> vector<Block>({t, f});
+  str += t >> loop_head;
```
dekimir: Sigh, now the back-edge block is not post-dominating the continue target, violating another rule. :confounded: I'd go with something quite simple:
```
entry >> merge
header >> {merge, f}  // header is unreachable
```
I believe this violates only the rule about header dominating the merge.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,629 @@
...
+  str += "OpFunctionEnd\n";
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_CFG, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              MatchesRegex("Block [0-9]\\[merge\\] is already a merge block "
+                            "for another header"));
+}
+
+TEST_F(ValidateCFG, MergeBlockTargetedByMultipleHeaderBlocksSelectionBad) {
```
dekimir: This seems identical to the previous test?  If there's a subtle difference I'm missing, it should be called out clearly in a comment.

test/Validate.CFG.cpp
dekimir: Can't see where this has been addressed?

test/Validate.CFG.cpp
dekimir: Can't see where this has been addressed?

umar456: Hmm. It seems that I made these changes in a different branch. I have make the necessary changes in the latest commit

test/Validate.CFG.cpp
umar456: Added TODOs [here](https://github.com/KhronosGroup/SPIRV-Tools/pull/197/files#diff-be61021999392e983166ea28c9ca008bR607)

umar456:
Adressed feedback

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,599 @@
...
+      MatchesRegex(
+          "Block\\(s\\) \\{[0-9]\\[middle2\\] .\\} are referenced but not "
+          "defined in function [0-9]\\[Main\\]"));
+}
+
+TEST_F(ValidateCFG, HeaderDoesntDominatesMergeBad) {
+  Block entry("entry");
+  Block merge("merge");
+  Block head("head", SpvOpBranchConditional);
+  Block f("f");
```
dekimir: This block is unnecessary, though the test is still valid. OK to keep, in the interest of moving this PR along.

test/Validate.CFG.cpp
```diff
@@ -0,0 +1,599 @@
...
+          "[0-9]\\[merge\\]"));
+}
+
+TEST_F(ValidateCFG, UnreachableMerge) {
+  Block entry("entry");
+  Block branch("branch", SpvOpBranchConditional);
+  Block t("t", SpvOpReturn);
+  Block f("f", SpvOpReturn);
+  Block merge("merge");
+  Block end("end", SpvOpReturn);
```
dekimir: I still think this block is unnecessary (`merge` could return), though the test is still correct.  OK to keep, in the interest of moving this PR along.

dekimir:
+2

Merging now...

dekimir:
Squashed and merged to master via  0b9fd50.  Thanks, @umar456!

PR #198, opened 2016-04-25
Refactor v1.1 tests and turn them on.

dekimir:
Previously, the "V11" text fixture was empty.

### DIFF
--- a/test/OperandCapabilities.cpp
+++ b/test/OperandCapabilities.cpp
@@ -632,7 +632,7 @@ INSTANTIATE_TEST_CASE_P(
 
 // See SPIR-V Section 3.31 Capability
 INSTANTIATE_TEST_CASE_P(
-    Capability, EnumCapabilityTest,
+    CapabilityDependsOn, EnumCapabilityTest,
     Combine(
         Values(SPV_ENV_UNIVERSAL_1_0, SPV_ENV_UNIVERSAL_1_1),
         ValuesIn(std::vector<EnumCapabilityCase>{
@@ -697,6 +697,15 @@ INSTANTIATE_TEST_CASE_P(
             CASE1(CAPABILITY, CapabilityMultiViewport, Geometry),
             // clang-format on
         })), );
+
+INSTANTIATE_TEST_CASE_P(
+    CapabilityDependsOnV11, EnumCapabilityTest,
+    Combine(Values(SPV_ENV_UNIVERSAL_1_1),
+            ValuesIn(std::vector<EnumCapabilityCase>{
+                CASE1(CAPABILITY, CapabilitySubgroupDispatch, DeviceEnqueue),
+                CASE1(CAPABILITY, CapabilityNamedBarrier, Kernel),
+                CASE1(CAPABILITY, CapabilityPipeStorage, Pipes),
+            })), );
 
 #undef CASE0
 #undef CASE1
--- a/test/OperandCapabilities.cpp
+++ b/test/OperandCapabilities.cpp
@@ -26,9 +26,18 @@
 
 // Test capability dependencies for enums.
 
+#include <tuple>
+
 #include "UnitSPIRV.h"
 
 namespace {
+
+using ::testing::Combine;
+using ::testing::TestWithParam;
+using ::testing::Values;
+using ::testing::ValuesIn;
+using std::get;
+using std::tuple;
 
 // A test case for mapping an enum to a capability mask.
 struct EnumCapabilityCase {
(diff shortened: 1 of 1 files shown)


test/OperandCapabilities.cpp
```diff
@@ -38,23 +47,17 @@ struct EnumCapabilityCase {
...
- protected:
-  const spv_target_env env_ = env;  // Target environment to use in tests.
-};
-
-using EnumCapabilityTestV10 = EnumCapabilityTest<SPV_ENV_UNIVERSAL_1_0>;
-using EnumCapabilityTestV11 = EnumCapabilityTest<SPV_ENV_UNIVERSAL_1_1>;
+using EnumCapabilityTest =
+    TestWithParam<tuple<spv_target_env, EnumCapabilityCase>>;
 
-TEST_P(EnumCapabilityTestV10, Sample) {
```
dekimir: This was a test case only in the `V10` fixture, leaving the `V11` fixture empty.

test/OperandCapabilities.cpp
```diff
@@ -73,135 +76,156 @@ TEST_P(EnumCapabilityTestV10, Sample) {
...
+            CASE1(EXECUTION_MODE, ExecutionModeOutputPoints, Geometry),
+            CASE1(EXECUTION_MODE, ExecutionModeOutputLineStrip, Geometry),
+            CASE1(EXECUTION_MODE, ExecutionModeOutputTriangleStrip, Geometry),
+            CASE1(EXECUTION_MODE, ExecutionModeVecTypeHint, Kernel),
+            CASE1(EXECUTION_MODE, ExecutionModeContractionOff, Kernel),
+        })), );
+
+INSTANTIATE_TEST_CASE_P(
+    ExecutionModeV11, EnumCapabilityTest,
+    Combine(Values(SPV_ENV_UNIVERSAL_1_1),
```
antiagainst: Is `SPV_ENV_UNIVERSAL_1_0` missing here?

antiagainst: Never mind. I see these are new executions added in v1.1.

antiagainst: But it seems we should also have `Initializer` & `Finalizer` here?

test/OperandCapabilities.cpp
```diff
@@ -242,394 +266,437 @@ INSTANTIATE_TEST_CASE_P(
...
+                SPV_OPERAND_TYPE_DECORATION, uint32_t(SpvDecorationSpecId),
+                SPV_CAPABILITY_AS_MASK(SpvCapabilityShader)})));
+
+INSTANTIATE_TEST_CASE_P(
+    DecorationSpecIdV11, EnumCapabilityTest,
+    Combine(Values(SPV_ENV_UNIVERSAL_1_1),
+            Values(EnumCapabilityCase{
+                SPV_OPERAND_TYPE_DECORATION, uint32_t(SpvDecorationSpecId),
+                (SPV_CAPABILITY_AS_MASK(SpvCapabilityShader) |
+                 SPV_CAPABILITY_AS_MASK(SpvCapabilityKernel))})));
```
antiagainst: Version 1.1 also added `MaxByteOffset` as a new decoration which depends on `Addresses`.

test/OperandCapabilities.cpp
```diff
@@ -242,394 +266,437 @@ INSTANTIATE_TEST_CASE_P(
...
+            CASE1(CAPABILITY, CapabilityStorageImageExtendedFormats, Shader),
+            CASE1(CAPABILITY, CapabilityImageQuery, Shader),
+            CASE1(CAPABILITY, CapabilityDerivativeControl, Shader),
+            CASE1(CAPABILITY, CapabilityInterpolationFunction, Shader),
+            CASE1(CAPABILITY, CapabilityTransformFeedback, Shader),
+            CASE1(CAPABILITY, CapabilityGeometryStreams, Geometry),
+            CASE1(CAPABILITY, CapabilityStorageImageReadWithoutFormat, Shader),
+            CASE1(CAPABILITY, CapabilityStorageImageWriteWithoutFormat, Shader),
+            CASE1(CAPABILITY, CapabilityMultiViewport, Geometry),
+            // clang-format on
```
antiagainst: New capabilities---`SubgroupDispatch`, `NamedBarrier`, & `PipeStorage`---have their own capability dependencies.

test/OperandCapabilities.cpp
dekimir: Those tests have not yet been implemented.  This is a refactoring of existing tests.

test/OperandCapabilities.cpp
dekimir: To be added later...

test/OperandCapabilities.cpp
dekimir: `PipeStorage` to be added later.

antiagainst:
LGTM

test/OperandCapabilities.cpp
dekimir: Added `CapabilityDependsOnV11`, which tests new capabilities' dependencies.

antiagainst:
Since we added new capabilities in v1.1 in the previous commit, I'd think it's nice to add other new ones in this PR too. Will +2 after that.

dekimir:
@antiagainst: not quite sure what you mean.  There's a bunch more v1.1 features to add, and I'm going sequentially through the change notes (aside from one premature `PipeStorage` test in the last commit).  Once I go through them all, I'll know I've added everything that's needed.  But why can't I check in the intermediate steps along the way?

antiagainst:
@dekimir: I was not saying that we should have all the new shiny things in v1.1 added in this particular PR. This PR modifies the test cases for testing *capability dependencies of enumerants* of operand kinds. There is just a small number of such new enumerants that depends on some capabilities (as commented earlier). Since you've addressed some of them, it's better to address all in the same PR. It's more clear and easy to verify.

dekimir:
@antiagainst: so what's still missing?

antiagainst:
@dekimir:

* `Initializer` & `Finalizer` in the `ExecutionMode` operand kind: depending on the `Kernel` cap. Around line 160.
* `MaxByteOffset` in the `Decoration` operand kind: depending on the `Addresses` cap. Around line 468.

antiagainst:
Anyway, I don't care whether it's in the same PR anymore. I can always refer back to this PR to refresh my mind.

LGTM.

dekimir:
@antiagainst: as I explained above, those will be added when their turn comes in the change notes, which is after pipes, `OpSizeOf`, and loop ivdep.

dekimir:
Squashed and merged to master via 27d6008.

PR #205, opened 2016-04-29
Add v1.1 capabilities validation.

dekimir:
Explain the meaning of test parameters.  Factor out methods for convenience and readability.

### DIFF
--- a/test/Validate.Capability.cpp
+++ b/test/Validate.Capability.cpp
@@ -42,23 +42,49 @@ namespace {
 using spvtest::ScopedContext;
 using std::get;
 using std::make_pair;
+using std::ostringstream;
 using std::pair;
 using std::string;
-using std::ostringstream;
 using std::tuple;
 using std::vector;
 using testing::Combine;
 using testing::Values;
 using testing::ValuesIn;
 
-// Parameter for validation test fixtures. It's a tuple so Combine() can be used
-// to produce it.  The first string is a capability name, the second an assembly
-// text.  See below methods for how they're interpreted.
+// Parameter for validation test fixtures.  Captures assembly to test in two
+// parts: a variable top line and a fixed remainder.  The top line will be an
+// OpCapability instruction, while the remainder will be some assembly text that
+// succeeds or fails to assemble depending on which capability was chosen.  For
+// instance, the following will succeed:
+//
+// OpCapability Pipes ; implies Kernel
+// OpLifetimeStop %1 0 ; requires Kernel
+//
+// and the following will fail:
+//
+// OpCapability Kernel
+// %1 = OpTypeNamedBarrier ; requires NamedBarrier
+//
+// So how does the test parameter capture which capabilities should cause
+// success and which shouldn't?  The answer is in the last element: it's a
+// vector of capabilities that make the remainder assembly succeed.  So if the
+// first-line capability exists in that vector, success is expected; otherwise,
+// failure is expected in the tests.
+//
+// We will use testing::Combine() to vary the first line: when we combine
+// AllCapabilities() with a single remainder assembly, we generate enough test
+// cases to try the assembly with every possible capability that could be
+// declared. However, Combine() only produces tuples -- it cannot produce, say,
+// a struct.  Therefore, this type must be a tuple.
+//
+// The first string is a capability name for the first line, the second the
+// remainder assembly, and the vector at the end has all capabilities that must
+// succeed.  See below for convenience methods to access each one.
 using CapTestParameter = tuple<string, pair<string, vector<string>>>;
 
 const string& Capability(const CapTestParameter& p) { return get<0>(p); }
-const string& Assembly(const CapTestParameter& p) { return get<1>(p).first; }
-const vector<string>& RequiredCapabilities(const CapTestParameter& p) {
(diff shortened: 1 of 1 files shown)


test/Validate.Capability.cpp
```diff
@@ -40,18 +40,51 @@
 namespace {
 
 using spvtest::ScopedContext;
+using std::get;
 using std::make_pair;
 using std::pair;
 using std::string;
-using std::stringstream;
+using std::ostringstream;
```
antiagainst: ordering

test/Validate.Capability.cpp
```diff
@@ -40,18 +40,51 @@
...
 using testing::ValuesIn;
 
-using ValidateCapability =
-    spvtest::ValidateBase<tuple<string, pair<string, vector<string>>>>;
+// Parameter for validation test fixtures. It's a tuple so Combine() can be used
+// to produce it.  The first string is a capability name, the second an assembly
+// text.  See below methods for how they're interpreted.
+using CapTestParameter = tuple<string, pair<string, vector<string>>>;
+
+const string& Capability(const CapTestParameter& p) { return get<0>(p); }
```
antiagainst: Instead of defining `CapTestParameter` as a `tuple` and have multiple standalone functions specifically for it, why not have a `struct` or `class` to group all the data and methods together? I'd think the latter is more easier to read and maintain.

umar456: This is done to use the testing::Combine functionality of google test framework. The combine construct requires tuples as far as I understand.

antiagainst: Oh, nice to know.

antiagainst:
Discussed offline. It would be nice to add some high level comments to explain how the capability tests are structured to achieve their purposes. Now it takes quite some time to grasp by jumping in fresh or coming back after not touching this code for a while.

test/Validate.Capability.cpp
dekimir: This was a `tuple` before this PR and remains a `tuple` after it.  I'm just trying to clarify its use here.  Will add a better explanatory comment.

dekimir:
Feedback addressed.

test/Validate.Capability.cpp
```diff
@@ -40,18 +40,77 @@
...
 using std::vector;
 using testing::Combine;
 using testing::Values;
 using testing::ValuesIn;
 
-using ValidateCapability =
-    spvtest::ValidateBase<tuple<string, pair<string, vector<string>>>>;
+// Parameter for validation test fixtures.  Captures assembly to test in two
+// parts: a variable top line and a fixed remainder.  The top line will be an
+// OpCapability instruction, while the remainder will be some assembly text that
```
antiagainst: s/`OpCapability instruction`/`capability`/

test/Validate.Capability.cpp
```diff
@@ -40,18 +40,77 @@
...
+// first-line capability exists in that vector, success is expected; otherwise,
+// failure is expected in the tests.
+//
+// We will use testing::Combine() to vary the first line: when we combine
+// AllCapabilities() with a single remainder assembly, we generate enough test
+// cases to try the assembly with every possible capability that could be
+// declared. However, Combine() only produces tuples -- it cannot produce, say,
+// a struct.  Therefore, this type must be a tuple.
+//
+// The first string is a capability name for the first line, the second the
```
antiagainst: It seems to me that this paragraph should be at the very beginning.

antiagainst:
Much clear to me now, thanks!

Just two more nits.

test/Validate.Capability.cpp
dekimir: Why reduce precision?  And why confuse the capability name with the capability instruction?

test/Validate.Capability.cpp
dekimir: Wouldn't that be confusing, before we explained what "capability", "remainder", and "vector of must-pass" mean?

test/Validate.Capability.cpp
antiagainst: We are looping the cases in `AllCapabilities`, so each one is just a **capability name**, not an instruction.

dekimir: True, but this isn't talking about the parameter yet -- it's talking about the assembly to test.  Perhaps I can make that clearer?

test/Validate.Capability.cpp
antiagainst: But, by providing this paragraph here, we don't have an overall brief intro at the very beginning. (I assume comments should typically follow this pattern: brief intro and then detailed explanation.)

Besides, the above detailed explanation already depicts exactly what these components are; repeating at the very end seems like duplicating without more info.

test/Validate.Capability.cpp
antiagainst: Well, this is the comment for `CapTestParameter`, right? And the very beginning sentence reads like we are starting to explain what `Parameter for validation test fixtures`, right?

dekimir:
OK, reworded again.  PTAL.

antiagainst:
Cool! +2.

dekimir:
Squashed and merged to master via 3732d6c.

PR #206, opened 2016-04-29
Fix several warnings exposed in MSVS.

null77:
diagnostic.cpp:
- unreachable code

operand.cpp
- conversion between int and uint32_t
- unreachable code

text.cpp:
- assignment within conditional expression

hex_float.h:
- assignment within conditional expression
- conversion from 'const int' to 'unsigned int'
- unreachable code

validate_id.cpp
- forcing value to bool 'true' or 'false'

validate_types.cpp:
- forcing value to bool 'true' or 'false'

### DIFF
--- a/source/util/hex_float.h
+++ b/source/util/hex_float.h
@@ -515,7 +515,7 @@ class HexFloat {
     // If we are up-casting, then we just have to shift to the right location.
     if (num_throwaway_bits <= 0) {
       out_val = static_cast<other_uint_type>(significand);
-      uint_type shift_amount = -num_throwaway_bits;
+      uint_type shift_amount = static_cast<uint_type>(-num_throwaway_bits);
       out_val = static_cast<other_uint_type>(out_val << shift_amount);
       return out_val;
     }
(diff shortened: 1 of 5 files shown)


source/util/hex_float.h
```diff
@@ -662,12 +659,19 @@ inline uint8_t get_nibble_from_character(int character) {
   const char* dec = "0123456789";
   const char* lower = "abcdef";
   const char* upper = "ABCDEF";
-  const char* p = nullptr;
-  if ((p = strchr(dec, character))) {
+  const char* p = p = strchr(dec, character);
```
dekimir: Does 'p =' really have to be repeated?

dekimir:
BTW, which MSVS version is this?  We seem to have no trouble on 2013 and 2015 with the code as it currently stands.

Also, I would oppose a blanket ban on assignment within conditional expression going forward.

source/util/hex_float.h
dekimir: Interestingly, this line seems to break the gcc build (see "checks have failed" below).

null77:
Updated CL, reverted the assignment within conditional fix. The extra 'p' was a bug, and it's cool that the CI caught that. Updated commit message to show that I'm using VS2015.

dekimir:
Rebased and merged to master via 34cb003.

PR #208, opened 2016-05-09
Add --target-env.

dekimir:
Convert `--vulkan` to `--target-env` in spirv-val.  Add `--target-env` to spirv-as.

@dneto0 @Qining @AWoloszyn @antiagainst PTAL.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -351,10 +351,6 @@ typedef enum {
 // Returns a string describing the given SPIR-V target environment.
 const char* spvTargetEnvDescription(spv_target_env env);
 
-// Parses s into *env and returns true if successful.  If unparsable, returns
-// false and sets *env to SPV_ENV_UNIVERSAL_1_0.
-bool spvParseTargetEnv(const char* s, spv_target_env* env);
-
 // Creates a context object.  Returns null if env is invalid.
 spv_context spvContextCreate(spv_target_env env);
 
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -351,6 +351,10 @@ typedef enum {
 // Returns a string describing the given SPIR-V target environment.
 const char* spvTargetEnvDescription(spv_target_env env);
 
+// Parses s into *env and returns true if successful.  If unparsable, returns
+// false and sets *env to SPV_ENV_UNIVERSAL_1_0.
+bool spvParseTargetEnv(const char* s, spv_target_env* env);
+
 // Creates a context object.  Returns null if env is invalid.
 spv_context spvContextCreate(spv_target_env env);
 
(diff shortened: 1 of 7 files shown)


tools/val/val.cpp
```diff
@@ -46,6 +45,8 @@ NOTE: The validator is a work in progress.
 Options:
   -h, --help   Print this help.
   --version    Display validator version information.
+  --target-env {vulkan1.0|spv1.0|spv1.1}
+               Use Vulkan/SPIR-V1.0/SPIR-V1.1 validation rules.
```
antiagainst: Vulkan1.0

tools/as/as.cpp
```diff
@@ -92,6 +92,19 @@ int main(int argc, char** argv) {
             print_usage(argv[0]);
             return 0;
           }
+          if (0 == strcmp(argv[argi], "--target-env")) {
```
antiagainst: Help messages need to be updated too.

include/spirv-tools/libspirv.h
```diff
@@ -351,6 +351,10 @@ typedef enum {
 // Returns a string describing the given SPIR-V target environment.
 const char* spvTargetEnvDescription(spv_target_env env);
 
+// Parses s into *env and returns true if successful.  If unparsable, returns
+// false and sets *env to SPV_ENV_UNIVERSAL_1_0.
+bool spvParseTargetEnv(const char* s, spv_target_env* env);
```
antiagainst: Do we really need to expose this function in the public API? Other public APIs in `libspirv.h` use `spv_target_env` directly; I don't see the points to expose a function converting strings to `spv_target_env`s here. It's mainly for our internal use.

antiagainst:
-1.

I don't think we need to export the `spvParseTargetEnv`.

dekimir:
Feedback addressed.

source/spirv_target_env.h
```diff
@@ -0,0 +1,34 @@
...
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#ifndef LIBSPIRV_TARGET_ENV_H_
```
antiagainst: nit: this should match the file name.

dekimir: Done.

antiagainst:
LGTM

dekimir:
Rebased, squashed, and pushed to master via 2703039.

PR #209, opened 2016-05-10
Add tests for MaxByteOffset.

dekimir:
@antiagainst @AWoloszyn @dneto0 @Qining PTAL.

PR #210, opened 2016-05-10
Test SpecId validation.

dekimir:

Separate from #209, but using the same technique.

### DIFF
--- a/test/Validate.Capability.cpp
+++ b/test/Validate.Capability.cpp
@@ -194,6 +194,68 @@ const vector<string>& AllCapabilities() {
   return *r;
 }
 
+const vector<string>& AllV10Capabilities() {
+  static const auto r = new vector<string>{
+    "",
+    "Matrix",
+    "Shader",
+    "Geometry",
+    "Tessellation",
+    "Addresses",
+    "Linkage",
+    "Kernel",
+    "Vector16",
+    "Float16Buffer",
+    "Float16",
+    "Float64",
+    "Int64",
+    "Int64Atomics",
+    "ImageBasic",
+    "ImageReadWrite",
+    "ImageMipmap",
+    "Pipes",
+    "Groups",
+    "DeviceEnqueue",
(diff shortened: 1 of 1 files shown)


test/Validate.Capability.cpp
```diff
@@ -833,6 +891,31 @@ make_pair(string(kGLSL450MemoryModel) +
...
 )),);
 
+// clang-format on
+INSTANTIATE_TEST_CASE_P(
+    DecorationSpecId, ValidateCapability,
+    Combine(ValuesIn(AllV10Capabilities()),
+            Values(make_pair(string(kOpenCLMemoryModel) +
+                                 "OpDecorate %intt SpecId 1\n"
+                                 "%intt = OpTypeInt 32 1\n",
+                             ShaderDependencies()))));
```
antiagainst: I'm not particularly clear here: do we need a comma before exiting `Values`?

test/Validate.Capability.cpp
```diff
@@ -194,6 +194,68 @@ const vector<string>& AllCapabilities() {
   return *r;
 }
 
+const vector<string>& AllV10Capabilities() {
```
antiagainst: It would be nice to rewrite `AllCapabilities` to reuse this function.

dekimir: It would be, particularly if `AllCaps` could still rely on static initialization in that case.  Otherwise, there'll be overhead whenever it's used, and it's used in nearly every test...

We have a similar pattern in `MatrixDependencies` vs. `ShaderDependencies`.

test/Validate.Capability.cpp
dekimir: No, but we do need one before closing `INSTANTIATE`.  Fixed.

antiagainst:
LGTM

dekimir:
Rebased and pushed into master via a7c1265.

PR #211, opened 2016-05-11
Remove done TODOs in OpNameTest.

dekimir:
Simplify the instantiation by using Values() instead of ValuesIn().

Run clang-format.


Qining:
LGTM

PR #218, opened 2016-05-25
Create a function for adding unittests in CMake.

antiagainst:
Also remove unnecessary main() function for unittests.


antiagainst:
This is the first commit in #219.

test/CMakeLists.txt
```diff
@@ -24,6 +24,29 @@
...
+    target_compile_options(${target} PRIVATE /wd4503)
+  endif()
+  target_include_directories(${target} PRIVATE
+    ${spirv-tools_SOURCE_DIR}
+    ${gtest_SOURCE_DIR}/include
+    ${gmock_SOURCE_DIR}/include
+  )
+  target_link_libraries(${target} PRIVATE ${libs})
+  target_link_libraries(${target} PRIVATE gmock_main)
+  add_test(NAME spvtools_${target} COMMAND ${target})
```
dneto0: The test name is inconsistent with the old style. 
Before we had spirv-tools-... and now its spvtools_
I like the hyphen (I think we used that elsewhere too), and I like spelling out spirv-tools.

dneto0:
-1 just for the naming of the test.

test/CMakeLists.txt
antiagainst: Done.

dneto0:
+2

Issue #217, opened 2016-05-23
Validator does not consider type aliasing with structs in interface blocks


Themaister:
When writing structs to an SSBO this validation error triggers:
```
error: 196: OpStore Pointer <id> '10's type does not match Object <id> '11's type.
```

SPIR-V:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 1
; Bound: 25
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %4 "main"
               OpExecutionMode %4 LocalSize 1 1 1
               OpSource ESSL 310
               OpName %4 "main"
               OpName %8 "MyStruct"
               OpMemberName %8 0 "foo"
               OpName %10 "bar"
               OpName %11 "MyStruct"
               OpMemberName %11 0 "foo"
               OpName %12 "SSBO0"
               OpMemberName %12 0 "foo_std430"
               OpName %14 ""
               OpName %20 "SSBO1"
               OpMemberName %20 0 "foo_std430_out"
               OpName %22 ""
               OpMemberDecorate %11 0 Offset 0
               OpMemberDecorate %12 0 Offset 0
               OpDecorate %12 BufferBlock
               OpDecorate %14 DescriptorSet 0
               OpDecorate %14 Binding 0
               OpMemberDecorate %20 0 Offset 0
               OpDecorate %20 BufferBlock
               OpDecorate %22 DescriptorSet 0
               OpDecorate %22 Binding 1
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeFloat 32
          %7 = OpTypeVector %6 4
          %8 = OpTypeStruct %7
          %9 = OpTypePointer Function %8
         %11 = OpTypeStruct %7
         %12 = OpTypeStruct %11
         %13 = OpTypePointer Uniform %12
         %14 = OpVariable %13 Uniform
         %15 = OpTypeInt 32 1
         %16 = OpConstant %15 0
         %17 = OpTypePointer Uniform %11
         %20 = OpTypeStruct %11
         %21 = OpTypePointer Uniform %20
         %22 = OpVariable %21 Uniform
          %4 = OpFunction %2 None %3
          %5 = OpLabel
         %10 = OpVariable %9 Function
         %18 = OpAccessChain %17 %14 %16
[...]

PR #219, opened 2016-05-25
Obsolete: First drop of the optimization framework and two passes

antiagainst:
This includes:
* `PassManager` & `Pass`
* Class hierarchy for SPIR-V types
* Def-use analysis
* `StripDebugInfoPass` & `UnifyTypePass`

Things definitely need improvement:
* Tests, lots of tests
* Better error handling and status reporting (panicking all the time is not good)

@dneto0 @dekimir @AWoloszyn @Qining PTAL

Preferably focusing on high-level design issues first; expecting radical changes. :)

### DIFF
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -1,0 +1,10 @@
+add_library(SpvToolsOpt
+  ${CMAKE_CURRENT_SOURCE_DIR}/constructs.h
+
+  ${CMAKE_CURRENT_SOURCE_DIR}/constructs.cpp
+)
+
+spvtools_default_compile_options(SpvToolsOpt)
+target_include_directories(SpvToolsOpt
+  PRIVATE ${spirv-tools_SOURCE_DIR}/include
+)
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -2,10 +2,12 @@ add_library(SpvToolsOpt
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.h
   ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.h
   ${CMAKE_CURRENT_SOURCE_DIR}/types.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/type_builder.h
 
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/types.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/type_builder.cpp
 )
 
 spvtools_default_compile_options(SpvToolsOpt)
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -1,7 +1,9 @@
 add_library(SpvToolsOpt
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.h
 
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.cpp
 )
 
 spvtools_default_compile_options(SpvToolsOpt)
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -10,6 +10,7 @@ add_library(SpvToolsOpt
 
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/def_use_manager.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/passes.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/types.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/type_builder.cpp
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -1,19 +1,19 @@
 add_library(SpvToolsOpt
-  ${CMAKE_CURRENT_SOURCE_DIR}/constructs.h
-  ${CMAKE_CURRENT_SOURCE_DIR}/def_use_manager.h
-  ${CMAKE_CURRENT_SOURCE_DIR}/reflect.h
-  ${CMAKE_CURRENT_SOURCE_DIR}/passes.h
-  ${CMAKE_CURRENT_SOURCE_DIR}/pass_manager.h
-  ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.h
-  ${CMAKE_CURRENT_SOURCE_DIR}/types.h
-  ${CMAKE_CURRENT_SOURCE_DIR}/type_builder.h
+  constructs.h
+  def_use_manager.h
+  reflect.h
+  passes.h
+  pass_manager.h
+  spv_builder.h
+  types.h
+  type_builder.h
 
-  ${CMAKE_CURRENT_SOURCE_DIR}/constructs.cpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/def_use_manager.cpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/passes.cpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.cpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/types.cpp
-  ${CMAKE_CURRENT_SOURCE_DIR}/type_builder.cpp
+  constructs.cpp
+  def_use_manager.cpp
+  passes.cpp
+  spv_builder.cpp
+  types.cpp
+  type_builder.cpp
 )
 
 spvtools_default_compile_options(SpvToolsOpt)
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -1,6 +1,7 @@
 add_library(SpvToolsOpt
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.h
   ${CMAKE_CURRENT_SOURCE_DIR}/def_use_manager.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/reflect.h
   ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.h
   ${CMAKE_CURRENT_SOURCE_DIR}/types.h
   ${CMAKE_CURRENT_SOURCE_DIR}/type_builder.h
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -1,10 +1,12 @@
 add_library(SpvToolsOpt
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/def_use_manager.h
   ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.h
   ${CMAKE_CURRENT_SOURCE_DIR}/types.h
   ${CMAKE_CURRENT_SOURCE_DIR}/type_builder.h
 
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/def_use_manager.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/types.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/type_builder.cpp
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -1,9 +1,11 @@
 add_library(SpvToolsOpt
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.h
   ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/types.h
 
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/spv_builder.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/types.cpp
 )
 
 spvtools_default_compile_options(SpvToolsOpt)
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -2,6 +2,8 @@ add_library(SpvToolsOpt
   ${CMAKE_CURRENT_SOURCE_DIR}/constructs.h
   ${CMAKE_CURRENT_SOURCE_DIR}/def_use_manager.h
   ${CMAKE_CURRENT_SOURCE_DIR}/reflect.h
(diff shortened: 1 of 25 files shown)


antiagainst:
Running the compiled `spirv-opt` on the SPIR-V provided in https://github.com/KhronosGroup/SPIRV-Tools/issues/217, debug insts are successfully removed and `%20`, `%21` is replaced by `%12`, `%13`, respectively.

source/opt/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+add_library(SpvToolsOpt
+  ${CMAKE_CURRENT_SOURCE_DIR}/constructs.h
```
dekimir: Why is `CMAKE_CURRENT_SOURCE_DIR` necessary?

antiagainst: Good point. They are unnecessary. Removed.

antiagainst:
Discussed in the morning.

Right now this PR serves as a complementary piece for reviewing the design. So leave it as-is for now. Code in this PR will be merged in eventually (with reviews and fixes definitely), but that will happen in other PRs with smaller granularity.

source/opt/passes.cpp
```diff
@@ -0,0 +1,58 @@
...
+  def_use_manager_.AnalyzeDefUse(module);
+
+  for (const auto inst : module->types()) {
+    type_builder_.CreateType(*inst);
+  }
+  for (const auto& inst : module->annotations()) {
+    type_builder_.AttachDecoration(inst);
+  }
+
+  const auto groups = type_builder_.GroupSameTypes();
```
Qining: Shall we call the function: getTypeGroupsForSameTypes() or getTypeGroups()?

tools/opt/opt.cpp
```diff
@@ -0,0 +1,184 @@
...
+  if (error) {
+    spvDiagnosticPrint(diagnostic);
+    spvDiagnosticDestroy(diagnostic);
+    return error;
+  }
+#endif
+
+  ir::Module module;
+  ir::SpvBuilder spv_builder(&module);
+  spvBinaryParse(context, &spv_builder, source.data(), source.size(),
```
Qining: Maybe trivial, but I think it is better to move all the parsing related code, including the two call back functions, to SpvBuilder. Just leave a BuildSpv(&module) in here.

source/opt/passes.h
```diff
@@ -0,0 +1,67 @@
...
+class TypeUnificationPass : public Pass {
+ public:
+  TypeUnificationPass() : type_builder_(&type_map_) {}
+  const char* name() const override { return "TypeUnification"; }
+  bool process(ir::Module* module) override;
+
+ private:
+  type::IdToTypeMap type_map_;
+  type::TypeBuilder type_builder_;
+  analysis::DefUseManager def_use_manager_;
```
Qining: I think it is better to have the def/use info attached to each instruction. That way we don't need to rebuild the def-use manager again for each pass that may need def-use info. Anyway, may be trivial.

antiagainst: Yes, I am debating on this issue, too.

The consideration was that some passes may not need the def-use analysis at all, like debug-info-removal. So I pulled this analysis out as a separate component. But it is quite common for the other passes. So embedding it in the in-memory representation may be a good idea. Another way of doing this may be just have the def-use analysis result passed around to different passes needing it.

source/opt/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+add_library(SpvToolsOpt
```
dneto0: This library name is inconsistent with the name of the main library (SPIRV-Tools)
I'm tempted to say this library should be SPIRV-Tools-opt.

But users might reasonably expect that SPIRV-Tools contains everything from the SPIRV-Tools project, but really it's the "core" stuff like the binary parser, assembler, etc.  But maybe that's ok.   We have the same issue with Shaderc and solved that problem somewhat with a way to build a shaderc_combined library.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+class Module;
+
+// Operand is a overloaded term. For better understanding, in this file, we
+// use payload to mean what "operand" is supposed to mean in the SPIR-V spec.
+// Operand itself is reserved to mean the real operands to an instruction,
+// excluding type ids and result ids.
+
+// A payload to an Inst. It can be a type id, a result id, or a real logical
+// operand.
+struct Payload {
```
dneto0: I think Payload is a worse term than Operand.  The things you're trying to represent really are all logical operands in the SPIR-V sense, even if it's result ID and type ID.
Note that the SPIR-V spec uses the term "operand" to mean a single word in an instruction after the opcode/wordcount word.  Within SPIRV-Tools I've used "logical operand" to mean "a named parameter to an instruction, but which might take multiple words to express" (including the cases of result ID and type ID.)

I'd much prefer you call this class Operand and be clear in the blurb for the class that it's a logical operand.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+  Payload(spv_operand_type_t t, std::vector<uint32_t>&& w)
+      : type(t), words(std::move(w)) {}
+
+  spv_operand_type_t type;      // Type of this payload.
+  std::vector<uint32_t> words;  // Binary segments of this payload.
```
dneto0: We're probably long overdue defining the following:
```
using Words = std::vector<uint32_t>
```

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+
+  // TODO(antiagainst): create fields for literal number kind, width, etc.
+};
+
+class Inst {
```
dneto0: Need blurbs on all classes.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+#define LIBSPIRV_OPT_CONSTRUCTS_H_
+
+#include <functional>
+#include <vector>
+
+#include "spirv-tools/libspirv.h"
+#include "spirv/spirv.h"
+
+namespace spvtools {
+namespace opt {
```
dneto0: The data structures representing higher level constructs in the IR should probably be in the namespace spvtools::ir.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+  // If |keep_nop| is true and this is a OpNop, avoids pushing.
+  void ToBinary(std::vector<uint32_t>* binary, bool keep_nop) const;
+
+ private:
+  uint32_t TypeResultIdCount() const {
+    return (type_id_ != 0) + (result_id_ != 0);
+  }
+
+  SpvOp opcode_;
+  uint32_t type_id_;
```
dneto0: Describe the fields.  for type_id and result_id, you should say that 0 means the there is no type_id (or result_id respectively).

source/opt/constructs.h
dekimir: There's already `spvtest::TextToBinaryTest::SpirvVector`.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+#include "spirv/spirv.h"
+
+namespace spvtools {
+namespace opt {
+namespace ir {
+
+class Function;
+class Module;
+
+// Operand is a overloaded term. For better understanding, in this file, we
```
dneto0: This is really confusing.  I now don't know what you mean by the i'th operand to an instruction.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+class Inst {
+ public:
+  Inst() : opcode_(SpvOpNop), type_id_(0), result_id_(0) {}
+  Inst(SpvOp op) : opcode_(op), type_id_(0), result_id_(0) {}
+  Inst(const spv_parsed_instruction_t& inst, std::vector<Inst>&& dbg_line = {});
+
+  SpvOp opcode() const { return opcode_; }
+  uint32_t type_id() const { return type_id_; }
+  uint32_t result_id() const { return result_id_; }
+  // Returns the first word of |index|th operand.
```
dneto0: I no longer know what this means.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+  uint32_t NumPayloadWords() const {
+    return NumOperandWords() + TypeResultIdCount();
+  }
+  // Returns the |index|th payload.
+  const Payload& GetPayload(uint32_t index) const;
+  uint32_t NumPayloads() const {
+    return static_cast<uint32_t>(payloads_.size());
+  }
+  void SetPayload(uint32_t index, std::vector<uint32_t>&& data);
+  std::vector<Inst>& dbg_line_info() { return dbg_line_info_; }
```
dneto0: Is this intended to be non-const?

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+  Inst(const spv_parsed_instruction_t& inst, std::vector<Inst>&& dbg_line = {});
+
+  SpvOp opcode() const { return opcode_; }
+  uint32_t type_id() const { return type_id_; }
+  uint32_t result_id() const { return result_id_; }
+  // Returns the first word of |index|th operand.
+  uint32_t GetSingleWordOperand(uint32_t index) const {
+    return GetSingleWordPayload(index + TypeResultIdCount());
+  }
+  uint32_t NumOperandWords() const;
```
dneto0: Need blurbs on all methods.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+    return static_cast<uint32_t>(payloads_.size());
+  }
+  void SetPayload(uint32_t index, std::vector<uint32_t>&& data);
+  std::vector<Inst>& dbg_line_info() { return dbg_line_info_; }
+
+  bool IsNop() const {
+    return opcode_ == SpvOpNop && type_id_ == 0 && result_id_ == 0 &&
+           payloads_.empty();
+  }
+  void ToNop() {
```
dneto0: This is very much like assignment from a default-constructed Inst().   So it's not clear that a client needs this method.
Is it supposed to clear the debug_line_info_ also?

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+  void AddInstruction(Inst&& i) { insts_.push_back(std::move(i)); }
+
+  void ForEachInst(const std::function<void(Inst*)>& f);
+
+  // Pushes the binary segments for this instruction into the back of *|binary|.
+  // If |keep_nop| is true and this is a OpNop, avoids pushing.
+  void ToBinary(std::vector<uint32_t>* binary, bool keep_nop) const;
+
+ private:
+  Function* function_;
```
dneto0: Must document each member.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+  Inst label_;
+  std::vector<Inst> insts_;
```
dneto0: I guess this doesn't include the label instruction?

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+
+class Function {
+ public:
+  Function(Module* module, Inst&& def_inst)
+      : module_(module), def_inst_(def_inst), end_inst_(SpvOpFunctionEnd) {}
+
+  void AddParameter(Inst&& p) { params_.push_back(std::move(p)); }
+  void AddBasicBlock(BasicBlock&& b) { blocks_.push_back(std::move(b)); }
+
+  void ForEachInst(const std::function<void(Inst*)>& f);
```
dneto0: Why not just give const ref access to the underlying vector?  Are you wanting to act on the label as well?

needs blurbs.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+  Inst label_;
+  std::vector<Inst> insts_;
+};
+
+class Function {
+ public:
+  Function(Module* module, Inst&& def_inst)
+      : module_(module), def_inst_(def_inst), end_inst_(SpvOpFunctionEnd) {}
+
+  void AddParameter(Inst&& p) { params_.push_back(std::move(p)); }
```
dneto0: Blurbs.

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+
+  void ForEachInst(const std::function<void(Inst*)>& f);
+
+  // Pushes the binary segments for this instruction into the back of *|binary|.
+  // If |keep_nop| is true and this is a OpNop, avoids pushing.
+  void ToBinary(std::vector<uint32_t>* binary, bool keep_nop) const;
+
+ private:
+  Module* module_;
+  Inst def_inst_;
```
dneto0: Need descriptions.  What's def-inst and end-inst?

source/opt/constructs.h
```diff
@@ -0,0 +1,229 @@
...
+  uint32_t bound;
+  uint32_t reserved;
+};
+
+class Module {
+ public:
+  Module() : header_({}) {}
+
+  void SetHeader(const ModuleHeader& header) { header_ = header; }
+  void AddCapability(Inst&& c) { capabilities_.push_back(std::move(c)); }
```
dneto0: It seems fishy that we have lots of "add" methods and no removals.  What if an optimization finds a dead function and wants to remove it?

antiagainst:
Okay, back to work on this PR. Since this is so gigantic, I'll split it into multiple CLs:

* Classes for in-memory representation (IMR).
* Builder for creating IMR.
* Passes & Pass manager.
* StripDebugInfoPass
* Type hierarchy & UnifyTypePass

source/opt/constructs.h
antiagainst: This is addressed in #238. No more payload. Just operands now.

source/opt/constructs.h
antiagainst: Addressed in #238.

source/opt/constructs.h
antiagainst: Yeah, but I won't bother right now since it's only used in the optimization part now.

source/opt/constructs.h
antiagainst: Should be fine now since the terms are revised.

source/opt/constructs.h
antiagainst: Done.

source/opt/constructs.h
antiagainst: Yes, so the caller can modify it at its will.

source/opt/constructs.h
antiagainst: Done.

source/opt/CMakeLists.txt
antiagainst: Renamed.

Actually I'd think it would be nice to split `libSPIRV-Tools.so` into multiple smaller libraries via code reorganization (towards better structure). But that may break existing users and it's not a high priority task.

source/opt/constructs.h
antiagainst: Done.

source/opt/constructs.h
antiagainst: We'll need this for transformations indicating "discarding" this instruction.

No, I don't think "debug_line_info" should be cleared at the same time since those line-related debug instructions are just placed here. They are not a part of this instruction.

source/opt/constructs.h
antiagainst: Done.

source/opt/constructs.h
antiagainst: It actually includes.

source/opt/constructs.h
antiagainst: The caller should be able to modify the content. So I'd think it's better to provide non-const ref.

source/opt/constructs.h
antiagainst: Done.

source/opt/constructs.h
antiagainst: Done.

source/opt/constructs.h
antiagainst: Remove is "implemented" using `ToNop`.

PR #221, opened 2016-06-02
Additional CFG tests

umar456:


test/Validate.CFG.cpp
```diff
@@ -120,16 +120,19 @@ class Block {
...
+        assert(successors_.size() == 0);
         out << "OpUnreachable\n";
         break;
       case SpvOpBranch:
+        assert(successors_.size() == 1);
         out << "OpBranch %" + successors_.front().label_;
         break;
       default:
-        assert(1 != 1 && "Unhandled");
+        assert(1 != 0 && "Unhandled");
```
dekimir: What's the purpose of, in effect, `assert(true)`?

test/Validate.CFG.cpp
```diff
@@ -589,10 +588,100 @@ TEST_F(ValidateCFG, UnreachableBranch) {
...
+  loop2.setBody("OpLoopMerge %loop2_merge %loop2 None\n");
+
+  string str =
+      header + types_consts + "%func    = OpFunction %voidt None %funct\n";
+
+  str += entry >> loop1;
+  str += loop1 >> vector<Block>({loop2, loop1_merge});
+  str += loop2 >> vector<Block>({loop2, loop2_merge});
+  str += loop2_merge >> loop1_merge;
+  str += loop1_merge;
```
dekimir: Where is the back edge to `loop1`?

umar456:
Addressed feedback

test/Validate.CFG.cpp
```diff
@@ -638,7 +639,8 @@ TEST_F(ValidateCFG, NestedLoops) {
   str += loop1 >> vector<Block>({loop2, loop1_merge});
   str += loop2 >> vector<Block>({loop2, loop2_merge});
   str += loop2_merge >> loop1_merge;
-  str += loop1_merge;
+  str += loop1_merge >> vector<Block>({loop1, exit});
```
dekimir: Let me see if I got this right: loop2 is loop1's continue target, and loop1_merge is loop1's back-edge block?  But loop2 doesn't dominate loop1_merge (there's a path entry >> loop1 >> loop1_merge), violating the rule that the continue target must dominate the back-edge block.

dekimir: Which reminds me: we should add a TODO to test that the validator flags this violation. :)

umar456: Good find. I clearly need to work on my graphs.

test/Validate.CFG.cpp
```diff
@@ -581,10 +580,133 @@ TEST_F(ValidateCFG, UnreachableBranch) {
...
+  loop2.setBody("OpLoopMerge %loop2_merge %loop2 None\n");
+
+  string str =
+      header + types_consts + "%func    = OpFunction %voidt None %funct\n";
+
+  str += entry >> loop1;
+  str += loop1 >> loop2;
+  str += loop2 >> vector<Block>({loop2, loop2_merge});
+  str += loop2_merge >> loop1_merge;
+  str += loop1_merge >> vector<Block>({loop1, exit});
```
dekimir: I'm afraid this is subtly wrong: `loop1_merge` branches to `exit`, which is outside the continue construct.  But `loop1_merge` isn't allowed to branch outside because of [this restriction](https://www.khronos.org/registry/spir-v/specs/1.1/SPIRV.html#_a_id_structuredcontrolflow_a_structured_control_flow) (look for  "the only blocks in a construct that can branch outside the construct are").

I think a quick way to fix this is to modify `loop1`'s continue target to be itself, i.e.:
```
loop1.setBody(
       "%cond    = OpSLessThan %intt %one %two\n"
       "OpLoopMerge %loop1_merge %loop1 None\n");
```

umar456: Thanks for bearing with me on this. I think I have corrected the issue.

dekimir:
Squashed and pushed into master via f9903c8.

PR #222, opened 2016-06-02
Make googletest regexes portable.

dekimir:
On Windows, test::MatchesRegex() doesn't support brackets.  On Linux, \d
doesn't work. :(

@dneto0 @antiagainst @AWoloszyn @Qining @umar456 PTAL.


dneto0:
+2

PR #225, opened 2016-06-03
Split validation classes into multiple files

umar456:
This change splits the validation class implementation into separate source files.

Previously all of the validation types were implemented in the `source/validate_types.cpp` file and defined in the `source/validate.h` header.

This change moves all of the types associated with the validation into the `source/val` directory. The `validate_[cfg,id,instruction,layout,ssa].cpp` and `validate.cpp` files are still located in the `source` folder. I would leave it up to you guys to decide the location of these files.

This PR does not add or remove any logic from the validator. Its purely a refactoring change.

This change was inspired by the work in #219.

### DIFF
--- a/source/diagnostic.h
+++ b/source/diagnostic.h
@@ -31,10 +31,23 @@
 #include <sstream>
 #include <utility>
 
-#include "spirv-tools/libspirv.h"
+#include <spirv-tools/libspirv.h>
+
+/// For debugging purposes only
+/// Prints the string to stdout
+#define MSG(msg)                                        \
+  do {                                                  \
+    libspirv::message(__FILE__, size_t(__LINE__), msg); \
+  } while (0)
+
+/// For debugging purposes only
+/// prints the variable value/location/and name to stdout
+#define SHOW(exp)                                               \
+  do {                                                          \
+    libspirv::message(__FILE__, size_t(__LINE__), #exp, (exp)); \
+  } while (0)
 
 namespace libspirv {
-
 class diagnostic_helper {
  public:
   diagnostic_helper(spv_position_t& position, spv_diagnostic* diagnostic)
@@ -103,6 +116,14 @@ class DiagnosticStream {
   helper.stream()
 
 std::string spvResultToString(spv_result_t res);
+
+/// Helper functions for printing debugging information
+void message(std::string file, size_t line, std::string name);
+
+template <typename T>
+void message(std::string file, size_t line, std::string name, T val) {
+  std::cout << file << ":" << line << ": " << name << " " << val << std::endl;
+}
 
 }  // namespace libspirv
 
--- a/source/diagnostic.h
+++ b/source/diagnostic.h
@@ -31,7 +31,7 @@
 #include <sstream>
 #include <utility>
 
-#include <spirv-tools/libspirv.h>
+#include "spirv-tools/libspirv.h"
 
 /// For debugging purposes only
 /// Prints the string to stdout
--- a/source/val/Construct.h
+++ b/source/val/Construct.h
@@ -1,0 +1,58 @@
+// Copyright (c) 2015-2016 The Khronos Group Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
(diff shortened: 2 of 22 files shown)


umar456:
Rebased against master

source/diagnostic.h
```diff
@@ -31,10 +31,23 @@
...
+#define MSG(msg)                                        \
+  do {                                                  \
+    libspirv::message(__FILE__, size_t(__LINE__), msg); \
+  } while (0)
+
+/// For debugging purposes only
+/// prints the variable value/location/and name to stdout
+#define SHOW(exp)                                               \
+  do {                                                          \
+    libspirv::message(__FILE__, size_t(__LINE__), #exp, (exp)); \
```
dneto0: Does a use of this macro even compile?  The message method only takes 3 parameters.

source/val/Construct.h
```diff
@@ -0,0 +1,58 @@
...
+
+class BasicBlock;
+
+/// @brief This class tracks the CFG constructs as defined in the SPIR-V spec
+class Construct {
+ public:
+  Construct(BasicBlock* header_block, BasicBlock* merge_block,
+            BasicBlock* continue_block = nullptr);
+
+  const BasicBlock* get_header() const;
```
dneto0: The style guide allows (encourages) simple accessors like this to be named like the member but without the trailing underscore.
But this is ok too.

source/val/Construct.h
```diff
@@ -0,0 +1,58 @@
...
+  const BasicBlock* get_continue() const;
+
+  BasicBlock* get_header();
+  BasicBlock* get_merge();
+  BasicBlock* get_continue();
+
+ private:
+  BasicBlock* header_block_;    ///< The header block of a loop or selection
+  BasicBlock* merge_block_;     ///< The merge block of a loop or selection
+  BasicBlock* continue_block_;  ///< The continue block of a loop block
```
dneto0: I presume this is nullptr for a selection.  Should say so.

source/val/Construct.h
umar456: ahh that would be nice. I can do that in another PR. I have done this in several places.

source/val/Construct.h
umar456: This is going to be removed in the next PR. I can apply the changes here if you like.

source/val/Function.cpp
```diff
@@ -0,0 +1,232 @@
...
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#include <val/Function.h>
+
+#include <cassert>
+
+#include <algorithm>
+
+#include <val/BasicBlock.h>
```
dneto0: I really prefer to reserve include < ... > for system or compiler-supplied headers.

source/val/Function.cpp
```diff
@@ -0,0 +1,232 @@
...
+size_t Function::get_undefined_block_count() const {
+  return undefined_blocks_.size();
+}
+
+const vector<BasicBlock*>& Function::get_blocks() const {
+  return ordered_blocks_;
+}
+vector<BasicBlock*>& Function::get_blocks() { return ordered_blocks_; }
+
+const BasicBlock* Function::get_current_block() const { return current_block_; }
```
dneto0: The signature has changed. It used to return a reference instead of a pointer.

source/val/Construct.h
dneto0: If it's in the next PR, then don't bother with this one.  the constructor above indicates it for now.

dneto0:
-1 Please change the #include with angle brackets to local files back to #include with quotes.

Also normally you should keep a "moving files around" PR to just doing that.  Then do logic changes like the Function::get_current_block() change and renames (like CFConstruct to Construct) to a different PR.
This time around I saw the get_current_block change in the first commit, separately.

umar456:
Sorry about that. I will keep that in mind for future PR.

I have made the suggested changes in the latest commit. Are you ok with the directory structure described in the descriptions?

source/validate_layout.cpp
```diff
@@ -26,17 +26,17 @@
 
 // Source code for logical layout validation as described in section 2.4
 
-#include <spirv-tools/libspirv.h>
 #include <validate.h>
 
+#include <cassert>
+
 #include <diagnostic.h>
```
dneto0: These should be include "" as well.

dneto0:
Ok, I squashed these into two commits and pushed onto master as 90a4252aae7

PR #226, opened 2016-06-04
Avoid accessing invalid iterator.

dekimir:
@umar456 @antiagainst @AWoloszyn @dneto0 @Qining PTAL

### DIFF
--- a/source/validate_cfg.cpp
+++ b/source/validate_cfg.cpp
@@ -85,11 +85,11 @@ vector<const BasicBlock*> PostOrderSort(const BasicBlock& entry, size_t size) {
       staged.pop_back();
     } else {
       BasicBlock* child = *top.iter;
+      top.iter++;
       if (processed.find(child->get_id()) == end(processed)) {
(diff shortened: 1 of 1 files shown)


umar456:
The iterator should still be valid after the if statement. The iterator that is incremented is part of the block_info struct and not the staged vector.

dekimir:
It's actually `top` that's invalid, due to `staged.emplace_back()` on line 90.

umar456:
'top' is a reference to the last block_info element and not an iterator.

dekimir:
Yes, sorry -- wrong title.  But the bug is real, and this is the fix.

umar456:
I don't understand the bug. I don't create an iterator for any of the containers being updated in that if statement.

dekimir:
`emplace_back()` reallocates `staged` memory, invalidating the `top` reference.  Dereferencing `top` after that is invalid, and it raises an exception in the Windows debug build.

umar456:
Ahh gotcha. Makes sense. +2

PR #227, opened 2016-06-08
Refactor depth first traversal to be more generic

umar456:
Refactor the way the post order vector is created. This new method
will allow for the detection of back-edges and create the post order
vector in one pass.

### DIFF
--- a/source/validate_cfg.cpp
+++ b/source/validate_cfg.cpp
@@ -67,15 +67,15 @@ struct block_info {
   bb_iter iter;   ///< Iterator to the current child node being processed
 };
 
-/// Returns true if the edge from last element of the staged vector to \p id
-/// creates a back-edge
+/// Returns true if a block with @p id is found in the @p work_list vector
 ///
-/// @param[in] staged Set of blocks visited in the the depth first traversal
+/// @param[in] work_list Set of blocks visited in the the depth first traversal
 ///                   of the CFG
 /// @param[in] id The ID of the block being checked
-/// @return true if the edge staged.back().block->get_id() => id is a back-edge
-bool IsBackEdge(vector<block_info> staged, uint32_t id) {
-  for (auto b : staged) {
+/// @return true if the edge work_list.back().block->get_id() => id is a
+/// back-edge
+bool FindInWorkList(vector<block_info> work_list, uint32_t id) {
+  for (auto b : work_list) {
     if (b.block->get_id() == id) return true;
   }
   return false;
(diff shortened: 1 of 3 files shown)


source/validate_cfg.cpp
```diff
@@ -59,56 +60,72 @@ using bb_ptr = BasicBlock*;
...
-    bb_iter iter;
-  };
+using get_blocks_func = function<const vector<BasicBlock*>*(const BasicBlock*)>;
+
+struct block_info {
+  cbb_ptr block;
+  bb_iter iter;
+};
+
+bool IsBackEdge(vector<block_info> staged, uint32_t id) {
```
dneto0: Needs a blurb.

source/validate_cfg.cpp
```diff
@@ -212,8 +229,15 @@ spv_result_t MergeBlockAssert(ValidationState_t& _, uint32_t merge_block) {
 spv_result_t PerformCfgChecks(ValidationState_t& _) {
   for (auto& function : _.get_functions()) {
     // Updates each blocks immediate dominators
+    vector<const BasicBlock*> postorder;
+    vector<pair<uint32_t, uint32_t>> back_edges;
     if (auto* first_block = function.get_first_block()) {
-      auto edges = libspirv::CalculateDominators(*first_block);
+      DepthFirstTraversal(*first_block, successor, [](cbb_ptr) {},
+                          [&](cbb_ptr b) { postorder.push_back(b); },
+                          [&](cbb_ptr f, cbb_ptr t) {
```
dneto0: Suggest rename f -> from, t -> to

umar456:
Addressed feedback

source/validate_cfg.cpp
```diff
@@ -59,56 +60,72 @@ using bb_ptr = BasicBlock*;
...
+        preorder(child);
+        staged.push_back({child, begin(*successor_func(child))});
         processed.insert(child->get_id());
       }
     }
   }
-  return out;
 }
+
+const vector<BasicBlock*>* successor(const BasicBlock* b) {
```
dneto0: Needs a blurb.

source/validate_cfg.cpp
```diff
@@ -59,56 +60,72 @@ using bb_ptr = BasicBlock*;
...
+/// BasicBlock and calls the pre/postorder functions when it needs to process
+/// the node in pre order, post order. It also calls the backedge function
+/// when a back edge is encountered
+void DepthFirstTraversal(const BasicBlock& entry,
+                         get_blocks_func successor_func,
+                         function<void(cbb_ptr)> preorder,
+                         function<void(cbb_ptr)> postorder,
+                         function<void(cbb_ptr, cbb_ptr)> backedge) {
   vector<cbb_ptr> out;
   vector<block_info> staged;
```
dneto0: I personally have a mild preference to renaming "staged" -> "work_list".
It's also worth saying that the work list is used as a stack of nodes. At any point of execution, from right to left, it its the sequence of nodes from the current node back to the entry node.

source/validate_cfg.cpp
```diff
@@ -59,56 +60,72 @@ using bb_ptr = BasicBlock*;
...
+
+struct block_info {
+  cbb_ptr block;
+  bb_iter iter;
+};
+
+bool IsBackEdge(vector<block_info> staged, uint32_t id) {
+  for (auto b : staged) {
+    if (b.block->get_id() == id) return true;
+  }
```
dneto0: I see what this function replaced below.  Its implementation could be
return staged.find(id) != staged.end().

source/validate_cfg.cpp
```diff
@@ -59,56 +60,72 @@ using bb_ptr = BasicBlock*;
...
-      out.push_back(top.block);
+    if (top.iter == end(*successor_func(top.block))) {
+      postorder(top.block);
       staged.pop_back();
     } else {
       BasicBlock* child = *top.iter;
       top.iter++;
-      if (processed.find(child->get_id()) == end(processed)) {
-        staged.emplace_back(block_info{child, begin(child->get_successors())});
+      if (IsBackEdge(staged, child->get_id())) {
```
dneto0: I don't think separating out IsBackEdge into its own function provides enough value.  It's literally "is this id in this list".

source/validate_cfg.cpp
```diff
@@ -59,56 +60,72 @@ using bb_ptr = BasicBlock*;
...
+        backedge(top.block, child);
+      }
+      if (processed.count(child->get_id()) == 0) {
+        preorder(child);
+        staged.push_back({child, begin(*successor_func(child))});
```
dneto0: Suggest using emplace_back like you did before.  That would be more consistent.

source/validate_cfg.cpp
```diff
@@ -59,56 +60,72 @@ using bb_ptr = BasicBlock*;
...
+  return false;
+}
 
+/// @brief Depth first traversal starting from the \p entry BasicBlock
+///
+/// This function performs a depth first traversal from the \p entry
+/// BasicBlock and calls the pre/postorder functions when it needs to process
+/// the node in pre order, post order. It also calls the backedge function
+/// when a back edge is encountered
+void DepthFirstTraversal(const BasicBlock& entry,
```
dneto0: The blurb should say that this algorithm requires the successor_func return a pointer to a collection such that iterators to that collection remain valid for the lifetime of the algorithm.
That is, the begin iterator we capture on line 109 remains valid for the duration of the algorithm.

dneto0:
-1 for generally minor things:  Definitely need blurbs, and to document the constraints on the successor function passed into the traverser.

source/validate_cfg.cpp
umar456: I originally had it as this unwieldy if statement:
```cpp
if(find_if(begin(work_list), end(work_list), [=](const block_info& b) { return b.block->get_id() == child->get_id(); })) {
backedge(top.block, child);
}
```
I personally thought it was hard to read but I can put it back if you think it is a better approach.

source/validate_cfg.cpp
umar456: staged is a vector of the `block_info` struct where as id is an integer. In order for that to work i would need to add a comparator between block_info and an integer. I don't know if that will be less complex then what I have now

source/validate_cfg.cpp
dneto0: It's ok to replace with a function, as long as the function name (and blurb) describe exactly what the function does.  "IsBackEdge" as a name is poor in this case because that pulls in assumptions about the context in which it's called. It's really "is the given Id in this container?".

umar456:
Addressed feedback so far

source/validate_cfg.cpp
```diff
@@ -59,56 +60,97 @@ using bb_ptr = BasicBlock*;
...
+/// Returns true if a block with @p id is found in the @p work_list vector
 ///
-/// Returns a vector of basic block pointers in a Control Flow Graph(CFG) which
-/// are sorted in the order they were accessed in a post order traversal.
+/// @param[in] work_list Set of blocks visited in the the depth first traversal
+///                   of the CFG
+/// @param[in] id The ID of the block being checked
+/// @return true if the edge work_list.back().block->get_id() => id is a
+/// back-edge
+bool FindInWorkList(vector<block_info> work_list, uint32_t id) {
```
dneto0: Just noticed this makes a copy when it doesn't need to.
Should be a const ref to the vector.

dneto0:
I squashed and pushed into master as e3dcaf56100
Then I added a followup commit to avoid the copy on the work list arg: 9ddd54d16d1

PR #230, opened 2016-06-09
Fix diagnostic message for layout errors

### LINKED ISSUE #228: [SPIRV-Val][OpenCL] Wrong diagnostics of OpLabel layout in function definition/declaration.
I got following error message for the attached [SPIR-V file](https://github.com/KhronosGroup/SPIRV-Tools/files/307610/test-OpLabel.zip):

  error: 25: A function must begin with a label

The disassembled module part looks like this:

         %12 = OpFunction %6 None %11
         %13 = OpFunctionParameter %8
         %14 = OpFunctionParameter %10
         %15 = OpLabel

According to specification function definition layout is:
- Function definition, using OpFunction.
- Function parameter declarations, using OpFunctionParameter.
- Block
- Block
- …
- Function end, using OpFunctionEnd.

Block must start with OpLabel instruction.

It looks like SPIR-V validation tools doesn't expect OpFunctionParameter to follow OpFunction instruction.

Function declaration layout check has the same issue.

umar456:
Fixed invalid error message when an instruction that appears in a block is encountered before a function declaration.

Fixes #228

### DIFF
--- a/source/val/ValidationState.cpp
+++ b/source/val/ValidationState.cpp
@@ -288,7 +288,8 @@ Function& ValidationState_t::get_current_function() {
 bool ValidationState_t::in_function_body() const { return in_function_; }
 
 bool ValidationState_t::in_block() const {
-  return module_functions_.back().get_current_block() != nullptr;
+  return module_functions_.empty() == false &&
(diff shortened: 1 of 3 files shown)


dneto0:
+2

Issue #229, opened 2016-06-09
[SPIRV-Val][OpenCL] Incorrect check on using OpTypePointer.


bader:
I got following error for the attached [SPIR-V file](https://github.com/KhronosGroup/SPIRV-Tools/files/307621/test-OpTypePointer.zip):

    error: 350: OpReturnValue value's type <id> '10' is a pointer, but a pointer can only be an operand to OpLoad, OpStore, OpAccessChain, or OpInBoundsAccessChain.

The list of the instructions accepting pointer operands is too short. There are a lot more instructions (especially with Kernel capability) that have pointer operands.

PR #231, opened 2016-06-13
In physical addressing, functions can return pointers

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/229

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -1710,12 +1710,11 @@ bool idUsage::isValid<SpvOpReturnValue>(const spv_instruction_t* inst,
                      << value.second.type_id << "' is missing or void.";
     return false;
   }
-  if (SpvOpTypePointer == valueType.second.opcode) {
-    DIAG(valueIndex) << "OpReturnValue value's type <id> '"
-                     << value.second.type_id
-                     << "' is a pointer, but a pointer can only be an operand "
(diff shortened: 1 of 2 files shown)


source/validate_id.cpp
```diff
@@ -1710,10 +1710,12 @@ bool idUsage::isValid<SpvOpReturnValue>(const spv_instruction_t* inst,
...
-  if (SpvOpTypePointer == valueType.second.opcode) {
+  if (addressingModel == SpvAddressingModelLogical &&
+      SpvOpTypePointer == valueType.second.opcode) {
     DIAG(valueIndex) << "OpReturnValue value's type <id> '"
                      << value.second.type_id
-                     << "' is a pointer, but a pointer can only be an operand "
+                     << "' is a pointer, but in the Logical addressing model "
+                        "a pointer can only be an operand "
                         "to OpLoad, OpStore, OpAccessChain, or "
                         "OpInBoundsAccessChain.";
```
dekimir: This is no longer accurate: https://www.khronos.org/registry/spir-v/specs/1.1/SPIRV.html#_universal_validation_rules

dekimir:
-1 for a minor error-message inaccuracy, otherwise looks good.

dneto0:
Good catch on the message.
All feedback addressed.

dekimir:
+2

Issue #143, opened 2016-03-04
Fix OpenCL pow and pown arguments


dneto0:
See https://www.khronos.org/bugzilla/show_bug.cgi?id=1469

We'll wait for the fix in the next revision of SPIR-V 1.0, then implement it here in SPIR-V Tools.

PR #232, opened 2016-06-14
Fix OpenCL pow and pown arguments

dneto0:
Spec fix appeared in SPIR-V 1.0 Rev 2:
https://www.khronos.org/registry/spir-v/specs/1.0/OpenCL.ExtendedInstructionSet.100.html#_changes_from_version_1_0_revision_1

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/143


AWoloszyn:
LGTM +2

PR #234, opened 2016-06-14
Construct and back-edge validation

umar456:
- [x] all CFG back edges must branch to a loop header, with each loop header having exactly one back edge branching to it
- [x] for a given loop, its back-edge block must post dominate the OpLoopMerge's Continue Target, and that Continue Target must dominate the back-edge block

### DIFF
--- a/source/val/BasicBlock.cpp
+++ b/source/val/BasicBlock.cpp
@@ -35,9 +35,10 @@ namespace libspirv {
 BasicBlock::BasicBlock(uint32_t id)
     : id_(id),
       immediate_dominator_(nullptr),
+      immediate_post_dominator_(nullptr),
       predecessors_(),
       successors_(),
-      type_(1),
+      type_(0),
       reachable_(false) {}
 
 void BasicBlock::SetImmediateDominator(BasicBlock* dom_block) {
@@ -65,7 +66,7 @@ void BasicBlock::RegisterSuccessors(vector<BasicBlock*> next_blocks) {
   for (auto& block : next_blocks) {
     block->predecessors_.push_back(this);
     successors_.push_back(block);
-    if (block->reachable_ == false) block->set_reachability(reachable_);
+    if (block->reachable_ == false) block->set_reachable(reachable_);
   }
 }
 
--- a/source/val/BasicBlock.cpp
+++ b/source/val/BasicBlock.cpp
@@ -37,6 +37,7 @@ BasicBlock::BasicBlock(uint32_t id)
       immediate_dominator_(nullptr),
       predecessors_(),
       successors_(),
+      type_(1),
       reachable_(false) {}
 
 void BasicBlock::SetImmediateDominator(BasicBlock* dom_block) {
--- a/source/val/BasicBlock.cpp
+++ b/source/val/BasicBlock.cpp
@@ -62,7 +62,7 @@ BasicBlock* BasicBlock::GetImmediatePostDominator() {
   return immediate_post_dominator_;
 }
 
-void BasicBlock::RegisterSuccessors(vector<BasicBlock*> next_blocks) {
+void BasicBlock::RegisterSuccessors(const vector<BasicBlock*>& next_blocks) {
   for (auto& block : next_blocks) {
     block->predecessors_.push_back(this);
     successors_.push_back(block);
--- a/source/val/BasicBlock.cpp
+++ b/source/val/BasicBlock.cpp
@@ -44,11 +44,22 @@ void BasicBlock::SetImmediateDominator(BasicBlock* dom_block) {
   immediate_dominator_ = dom_block;
 }
 
+void BasicBlock::SetImmediatePostDominator(BasicBlock* pdom_block) {
+  immediate_post_dominator_ = pdom_block;
+}
+
 const BasicBlock* BasicBlock::GetImmediateDominator() const {
   return immediate_dominator_;
 }
 
+const BasicBlock* BasicBlock::GetImmediatePostDominator() const {
+  return immediate_post_dominator_;
+}
+
 BasicBlock* BasicBlock::GetImmediateDominator() { return immediate_dominator_; }
+BasicBlock* BasicBlock::GetImmediatePostDominator() {
+  return immediate_post_dominator_;
+}
 
 void BasicBlock::RegisterSuccessors(vector<BasicBlock*> next_blocks) {
   for (auto& block : next_blocks) {
@@ -64,24 +75,29 @@ void BasicBlock::RegisterBranchInstruction(SpvOp branch_instruction) {
 }
 
 BasicBlock::DominatorIterator::DominatorIterator() : current_(nullptr) {}
-BasicBlock::DominatorIterator::DominatorIterator(const BasicBlock* block)
-    : current_(block) {}
+
+BasicBlock::DominatorIterator::DominatorIterator(
+    const BasicBlock* block,
+    std::function<const BasicBlock*(const BasicBlock*)> dominator_func)
+    : current_(block), dom_func_(dominator_func) {}
 
 BasicBlock::DominatorIterator& BasicBlock::DominatorIterator::operator++() {
-  if (current_ == current_->GetImmediateDominator()) {
+  if (current_ == dom_func_(current_)) {
     current_ = nullptr;
   } else {
-    current_ = current_->GetImmediateDominator();
+    current_ = dom_func_(current_);
   }
   return *this;
 }
 
 const BasicBlock::DominatorIterator BasicBlock::dom_begin() const {
-  return DominatorIterator(this);
+  return DominatorIterator(
+      this, [](const BasicBlock* b) { return b->GetImmediateDominator(); });
 }
 
 BasicBlock::DominatorIterator BasicBlock::dom_begin() {
-  return DominatorIterator(this);
+  return DominatorIterator(
+      this, [](const BasicBlock* b) { return b->GetImmediateDominator(); });
 }
 
 const BasicBlock::DominatorIterator BasicBlock::dom_end() const {
@@ -89,6 +105,24 @@ const BasicBlock::DominatorIterator BasicBlock::dom_end() const {
 }
 
 BasicBlock::DominatorIterator BasicBlock::dom_end() {
+  return DominatorIterator();
+}
+
+const BasicBlock::DominatorIterator BasicBlock::pdom_begin() const {
+  return DominatorIterator(
+      this, [](const BasicBlock* b) { return b->GetImmediatePostDominator(); });
+}
+
+BasicBlock::DominatorIterator BasicBlock::pdom_begin() {
+  return DominatorIterator(
+    this, [](const BasicBlock* b) { return b->GetImmediatePostDominator(); });
+}
+
+const BasicBlock::DominatorIterator BasicBlock::pdom_end() const {
+  return DominatorIterator();
+}
+
+BasicBlock::DominatorIterator BasicBlock::pdom_end() {
   return DominatorIterator();
 }
 
--- a/source/val/BasicBlock.h
+++ b/source/val/BasicBlock.h
@@ -78,21 +78,26 @@ class BasicBlock {
   /// Returns true if the block is reachable in the CFG
   bool is_reachable() const { return reachable_; }
 
-  /// Returns the type of the BasicBlock
-  bool is_type(BlockType type) const { return type_.test(type); }
+  /// Returns true if BasicBlock is of the given type
+  bool is_type(BlockType type) const {
+    if (type == kBlockTypeUndefined) return type_.none();
+    return type_.test(type);
+  }
 
-  /// Sets the reacability of the basic block in the CFG
-  void set_reachability(bool reachability) { reachable_ = reachability; }
+  /// Sets the reachability of the basic block in the CFG
+  void set_reachable(bool reachability) { reachable_ = reachability; }
 
   /// Sets the type of the BasicBlock
-  void set_type(BlockType type) { type_.set(static_cast<int>(type)); }
+  void set_type(BlockType type) {
+    type_.set(type);
+  }
 
   /// Sets the immedate dominator of this basic block
   ///
   /// @param[in] dom_block The dominator block
   void SetImmediateDominator(BasicBlock* dom_block);
 
-  /// Sets the immedate dominator of this basic block
+  /// Sets the immedate post dominator of this basic block
   ///
   /// @param[in] pdom_block The post dominator block
   void SetImmediatePostDominator(BasicBlock* pdom_block);
@@ -137,8 +142,8 @@ class BasicBlock {
     ///        @p block
     ///
     /// @param block          The block which is referenced by the iterator
-    /// @param dominator_func This function will be called to get the dominator
-    ///                       of the current block
+    /// @param dominator_func This function will be called to get the immediate
+    ///                       (post)dominator of the current block
     DominatorIterator(
         const BasicBlock* block,
         std::function<const BasicBlock*(const BasicBlock*)> dominator_func);
@@ -157,17 +162,17 @@ class BasicBlock {
     std::function<const BasicBlock*(const BasicBlock*)> dom_func_;
   };
 
-  /// Returns an dominator iterator which points to the current block
+  /// Returns a dominator iterator which points to the current block
   const DominatorIterator dom_begin() const;
 
-  /// Returns an dominator iterator which points to the current block
+  /// Returns a dominator iterator which points to the current block
   DominatorIterator dom_begin();
 
-  /// Returns an dominator iterator which points to one element past the first
+  /// Returns a dominator iterator which points to one element past the first
   /// block
   const DominatorIterator dom_end() const;
 
-  /// Returns an dominator iterator which points to one element past the first
+  /// Returns a dominator iterator which points to one element past the first
   /// block
   DominatorIterator dom_end();
 
@@ -201,7 +206,7 @@ class BasicBlock {
   std::vector<BasicBlock*> successors_;
 
   /// The type of the block
-  std::bitset<kBlockTypeCOUNT> type_;
+  std::bitset<kBlockTypeCOUNT - 1> type_;
 
   /// True if the block is reachable in the CFG
   bool reachable_;
--- a/source/val/BasicBlock.h
+++ b/source/val/BasicBlock.h
@@ -30,9 +30,22 @@
 #include "spirv/1.1/spirv.h"
 
 #include <cstdint>
+
+#include <bitset>
 #include <vector>
 
 namespace libspirv {
+
+enum BlockType : uint32_t {
+  kBlockTypeUndefined,
+  kBlockTypeHeader,
+  kBlockTypeLoop,
+  kBlockTypeMerge,
+  kBlockTypeBreak,
+  kBlockTypeContinue,
+  kBlockTypeReturn,
+  kBlockTypeCOUNT  ///< Total number of block types. (must be the last element)
+};
 
 // This class represents a basic block in a SPIR-V module
 class BasicBlock {
@@ -61,10 +74,17 @@ class BasicBlock {
   /// Returns the successors of the BasicBlock
   std::vector<BasicBlock*>* get_successors() { return &successors_; }
 
-  /// Returns true if the  block should be reachable in the CFG
+  /// Returns true if the block is reachable in the CFG
   bool is_reachable() const { return reachable_; }
 
+  /// Returns the type of the BasicBlock
+  bool is_type(BlockType type) const { return type_.test(type); }
+
+  /// Sets the reacability of the basic block in the CFG
   void set_reachability(bool reachability) { reachable_ = reachability; }
+
+  /// Sets the type of the BasicBlock
+  void set_type(BlockType type) { type_.set(static_cast<int>(type)); }
 
   /// Sets the immedate dominator of this basic block
   ///
@@ -140,6 +160,9 @@ class BasicBlock {
 
   /// The set of successors of the BasicBlock
   std::vector<BasicBlock*> successors_;
+
+  /// The type of the block
+  std::bitset<kBlockTypeCOUNT> type_;
 
   bool reachable_;
 };
--- a/source/val/BasicBlock.h
+++ b/source/val/BasicBlock.h
@@ -121,7 +121,7 @@ class BasicBlock {
   void RegisterBranchInstruction(SpvOp branch_instruction);
 
   /// Adds @p next BasicBlocks as successors of this BasicBlock
-  void RegisterSuccessors(std::vector<BasicBlock*> next = {});
+  void RegisterSuccessors(const std::vector<BasicBlock*>& next = {});
 
   /// Returns true if the id of the BasicBlock matches
   bool operator==(const BasicBlock& other) const { return other.id_ == id_; }
--- a/source/val/BasicBlock.h
+++ b/source/val/BasicBlock.h
@@ -89,6 +89,7 @@ class BasicBlock {
 
   /// Sets the type of the BasicBlock
   void set_type(BlockType type) {
+    if (type == kBlockTypeUndefined) type_.reset();
     type_.set(type);
   }
 
--- a/source/val/BasicBlock.h
+++ b/source/val/BasicBlock.h
@@ -89,8 +89,10 @@ class BasicBlock {
 
   /// Sets the type of the BasicBlock
   void set_type(BlockType type) {
-    if (type == kBlockTypeUndefined) type_.reset();
-    type_.set(type);
+    if (type == kBlockTypeUndefined)
+      type_.reset();
+    else
+      type_.set(type);
   }
 
   /// Sets the immedate dominator of this basic block
--- a/source/val/BasicBlock.h
+++ b/source/val/BasicBlock.h
@@ -32,6 +32,7 @@
 #include <cstdint>
 
 #include <bitset>
+#include <functional>
 #include <vector>
 
 namespace libspirv {
@@ -91,11 +92,22 @@ class BasicBlock {
   /// @param[in] dom_block The dominator block
   void SetImmediateDominator(BasicBlock* dom_block);
 
+  /// Sets the immedate dominator of this basic block
+  ///
+  /// @param[in] pdom_block The post dominator block
+  void SetImmediatePostDominator(BasicBlock* pdom_block);
+
   /// Returns the immedate dominator of this basic block
   BasicBlock* GetImmediateDominator();
 
   /// Returns the immedate dominator of this basic block
   const BasicBlock* GetImmediateDominator() const;
+
+  /// Returns the immedate post dominator of this basic block
+  BasicBlock* GetImmediatePostDominator();
+
+  /// Returns the immedate post dominator of this basic block
+  const BasicBlock* GetImmediatePostDominator() const;
 
   /// Ends the block without a successor
   void RegisterBranchInstruction(SpvOp branch_instruction);
@@ -111,7 +123,7 @@ class BasicBlock {
 
   /// @brief A BasicBlock dominator iterator class
   ///
-  /// This iterator will iterate over the dominators of the block
+  /// This iterator will iterate over the (post)dominators of the block
   class DominatorIterator
       : public std::iterator<std::forward_iterator_tag, BasicBlock*> {
    public:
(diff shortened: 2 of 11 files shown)


source/val/BasicBlock.h
```diff
@@ -61,22 +75,40 @@ class BasicBlock {
...
   std::vector<BasicBlock*>* get_successors() { return &successors_; }
 
-  /// Returns true if the  block should be reachable in the CFG
+  /// Returns true if the block is reachable in the CFG
   bool is_reachable() const { return reachable_; }
 
+  /// Returns the type of the BasicBlock
+  bool is_type(BlockType type) const { return type_.test(type); }
+
+  /// Sets the reacability of the basic block in the CFG
```
dneto0: Minor typo in comment.  "reachability".
Minor:  I could also go with "set_reachable"  since it matches the member name.

source/val/BasicBlock.h
```diff
@@ -61,22 +75,40 @@ class BasicBlock {
...
   void set_reachability(bool reachability) { reachable_ = reachability; }
 
+  /// Sets the type of the BasicBlock
+  void set_type(BlockType type) { type_.set(static_cast<int>(type)); }
```
dneto0: Minor.  BlockType has underlying type uint32_t.   It would be slightly better to use uint32_t here instead of static cast to int.
And it's probably too baroque to use std::underlying_type.  Yes, too baroque.

(still reviewing...)

source/val/BasicBlock.h
```diff
@@ -61,22 +75,40 @@ class BasicBlock {
   /// Returns the successors of the BasicBlock
   std::vector<BasicBlock*>* get_successors() { return &successors_; }
 
-  /// Returns true if the  block should be reachable in the CFG
+  /// Returns true if the block is reachable in the CFG
   bool is_reachable() const { return reachable_; }
 
+  /// Returns the type of the BasicBlock
```
dneto0: Returns true if block is of the given type.

source/val/BasicBlock.h
```diff
@@ -61,22 +75,40 @@ class BasicBlock {
...
 
+  /// Sets the type of the BasicBlock
+  void set_type(BlockType type) { type_.set(static_cast<int>(type)); }
+
   /// Sets the immedate dominator of this basic block
   ///
   /// @param[in] dom_block The dominator block
   void SetImmediateDominator(BasicBlock* dom_block);
 
+  /// Sets the immedate dominator of this basic block
```
dneto0: Copypasta.
You mean "post-dominator".

source/val/BasicBlock.h
```diff
@@ -104,8 +136,12 @@ class BasicBlock {
     /// @brief Constructs an iterator for the given block which points to
     ///        @p block
     ///
-    /// @param block The block which is referenced by the iterator
-    explicit DominatorIterator(const BasicBlock* block);
+    /// @param block          The block which is referenced by the iterator
+    /// @param dominator_func This function will be called to get the dominator
```
dneto0: Do you mean it gets the *immediate* dominator?  Otherwise the choice is not necessarily unique.

source/val/BasicBlock.h
```diff
@@ -118,29 +154,56 @@ class BasicBlock {
...
 
-  /// Returns an iterator which points to the current block
+  /// Returns an dominator iterator which points to the current block
   const DominatorIterator dom_begin() const;
+
+  /// Returns an dominator iterator which points to the current block
   DominatorIterator dom_begin();
 
-  /// Returns an iterator which points to one element past the first block
+  /// Returns an dominator iterator which points to one element past the first
```
dneto0: "past the first" is right in the dominator case but not right in the post-dominator case.  It's really the one element past the last valid block (in either direction).

Typo "an" -> "a"

dneto0: Nope I was wrong about the comment.
Typo still stands though:  "an" -> "a"

source/val/BasicBlock.h
dneto0: Functional issue:  Isn't it true that if you setting any type other than Undefined should *clear* the undefined bit?

As I understand it, the Undefined bit is set in the constructor, and nothing ever resets it.

source/val/Function.cpp
```diff
@@ -93,15 +99,36 @@ spv_result_t Function::RegisterLoopMerge(uint32_t merge_id,
                                          uint32_t continue_id) {
   RegisterBlock(merge_id, false);
   RegisterBlock(continue_id, false);
-  cfg_constructs_.emplace_back(get_current_block(), &blocks_.at(merge_id),
-                               &blocks_.at(continue_id));
+  BasicBlock& merge_block = blocks_.at(merge_id);
+  BasicBlock& continue_block = blocks_.at(continue_id);
+  assert(merge_block.is_type(kBlockTypeUndefined));
```
dneto0: This is always going to be true since the Undefined type bit is never reset.  ?!

source/val/Construct.h
```diff
@@ -28,29 +28,88 @@
...
+  void set_exit(BasicBlock* exit_block);
 
  private:
-  BasicBlock* header_block_;    ///< The header block of a loop or selection
-  BasicBlock* merge_block_;     ///< The merge block of a loop or selection
-  BasicBlock* continue_block_;  ///< The continue block of a loop block
+  /// The type of the construct
+  ConstructType type_;
+
+  /// These are the constructs that are related to this construct. These
```
dneto0: I don't understand what "related" means here.
The examples aren't helping.

source/val/Construct.h
```diff
@@ -28,29 +28,88 @@
...
-  const BasicBlock* get_merge() const;
-  const BasicBlock* get_continue() const;
+  /// Returns the dominator block of the construct. This is usually the header
+  /// block or the first block of the construct.
+  BasicBlock* get_dominator();
 
-  BasicBlock* get_header();
-  BasicBlock* get_merge();
-  BasicBlock* get_continue();
+  /// Returns the exit block of the construct. This is usually the merge block
```
dneto0: I think we can be more precise here:  For a continue construct this is the block with the backedge.  Otherwise it's the merge block for the corresponding header block.

source/val/Construct.h
```diff
@@ -28,29 +28,88 @@
...
+  /// constructs can be the loop construct, the corresponding loop construct,
+  /// the case construct that are part of the same OpSwitch instruction
+  std::vector<Construct*> corresponding_constructs_;
+
+  /// @brief Dominator block for the construct
+  ///
+  /// The dominator block for the construct. Depending on the construct this may
+  /// be a selection header, a continue target of a loop, a loop header or a
+  /// Target or Default block of a switch
+  BasicBlock* base_dominator_;
```
dneto0: The name is a little awkward.  How about the entry_block_
That has symmetry with the exit_block_, and is still accurate and doesn't need fuzziness with "base".

source/val/Construct.h
```diff
@@ -28,29 +28,88 @@
...
+  /// Returns the dominator block of the construct. This is usually the header
+  /// block or the first block of the construct.
+  const BasicBlock* get_dominator() const;
 
-  const BasicBlock* get_header() const;
-  const BasicBlock* get_merge() const;
-  const BasicBlock* get_continue() const;
+  /// Returns the dominator block of the construct. This is usually the header
+  /// block or the first block of the construct.
+  BasicBlock* get_dominator();
```
dneto0: Suggest calling this  get_entry(); (see below)

source/val/Construct.h
umar456: Sorry, there is a typo in the examples.

Here is a table that describe what is in a corresponding consturct


| this construct | corresponding construct          |
|-------------------|-----------------------------------------|
| loop           | continue                         |
| continue       | loop                             |
| case           | other cases in the same OpSwitch |

source/val/Construct.cpp
```diff
@@ -26,19 +26,51 @@
...
+  return corresponding_constructs_;
+}
+std::vector<Construct*>& Construct::get_corresponding_constructs() {
+  return corresponding_constructs_;
+}
+
+bool ValidateConstructSize(ConstructType type, size_t size) {
+  switch (type) {
+    case ConstructType::kSelection: return size == 0;
+    case ConstructType::kContinue:  return size == 1;
```
dneto0: I'm not sure this is right for kContinue, kLoop orkCase.

For kContinue:  A simple loop could have the loop header be the continue target, with the corresponding "loop construct" not existing at all.
For kCase, there might only be a default case in the OpSwitch, so the corresponding constructs might be empty.  Yes, it's degenerate, but it's valid.

umar456: I am creating constructs for kLoop and kContinue even if they represent the same block. So they will always have one corresponding construct.

But the kCase ... case is incorrect. It should be size >= 1;

source/val/BasicBlock.h
```diff
@@ -78,21 +78,26 @@ class BasicBlock {
...
 
-  /// Sets the reacability of the basic block in the CFG
-  void set_reachability(bool reachability) { reachable_ = reachability; }
+  /// Sets the reachability of the basic block in the CFG
+  void set_reachable(bool reachability) { reachable_ = reachability; }
 
   /// Sets the type of the BasicBlock
-  void set_type(BlockType type) { type_.set(static_cast<int>(type)); }
+  void set_type(BlockType type) {
+    type_.set(type);
```
dneto0: This won't work if type is the Undefined type.  Either make that part of the contract (in the blurb) and perhaps assert; or make setting to Undefined clear all the bits.

source/val/BasicBlock.h
```diff
@@ -89,6 +89,7 @@ class BasicBlock {
 
   /// Sets the type of the BasicBlock
   void set_type(BlockType type) {
+    if (type == kBlockTypeUndefined) type_.reset();
     type_.set(type);
```
dneto0: This can't be right.   When calling set_type with Undefined, then you still get that first bit set.  Maybe do the type_.set(type) in the else clause.

source/val/Function.h
```diff
@@ -179,6 +192,9 @@ class Function {
   /// The block that is currently being parsed
   BasicBlock* current_block_;
 
+  /// A psudo exit block that is the successor to all return blocks
+  BasicBlock exit_block_;
```
dneto0: Strongly suggest actually calling it pseudo_exit_block_

Also, it's worth saying that by default its ID will be 0.  By SPIR-V rules, that's never a valid ID, so at least the psuedo exit block won't ever compare as equal to any other block's id.

source/val/BasicBlock.cpp
```diff
@@ -35,25 +35,38 @@ namespace libspirv {
...
+const BasicBlock* BasicBlock::GetImmediatePostDominator() const {
+  return immediate_post_dominator_;
+}
+
 BasicBlock* BasicBlock::GetImmediateDominator() { return immediate_dominator_; }
+BasicBlock* BasicBlock::GetImmediatePostDominator() {
+  return immediate_post_dominator_;
+}
 
 void BasicBlock::RegisterSuccessors(vector<BasicBlock*> next_blocks) {
```
dneto0: Minor: Suggest making next_blocks const ref to avoid the copy.

source/val/Construct.cpp
dneto0: Ah, so that's definitely worth a comment about kContinue and kLoop.

source/val/Function.h
dneto0: Oh, never mind.  I see it is set to an invalid ID in the constructor.     Either way works.   I don't know if the validator has already checked the bounds on any Ids coming in, so there *might* be a risk that a module actually uses the kInvalidID.  ?  I'd have to look elsewhere in the validator.

source/val/Function.cpp
```diff
@@ -93,15 +99,35 @@ spv_result_t Function::RegisterLoopMerge(uint32_t merge_id,
...
+
+  current_block_->set_type(kBlockTypeLoop);
+  merge_block.set_type(kBlockTypeMerge);
+  continue_block.set_type(kBlockTypeContinue);
+  assert(current_block_ &&
```
dneto0: Move the assert before you dereference current_block_ at line 106.

source/val/Function.cpp
```diff
@@ -93,15 +99,35 @@ spv_result_t Function::RegisterLoopMerge(uint32_t merge_id,
...
+         "RegisterLoopMerge must be called when called within a block");
+  cfg_constructs_.emplace_back(ConstructType::kLoop, get_current_block(),
```
dneto0: Could use current_block_ instead of get_current_block().  ??

source/val/Function.cpp
```diff
@@ -93,15 +99,35 @@ spv_result_t Function::RegisterLoopMerge(uint32_t merge_id,
...
+                               &merge_block);
+  Construct& loop_construct = cfg_constructs_.back();
+  cfg_constructs_.emplace_back(ConstructType::kContinue, &continue_block,
+                               nullptr);
```
dneto0: Now that the exit argument defaults to nullptr, you can drop this third argument.

source/val/Function.cpp
```diff
@@ -93,15 +99,35 @@ spv_result_t Function::RegisterLoopMerge(uint32_t merge_id,
...
 
 spv_result_t Function::RegisterSelectionMerge(uint32_t merge_id) {
   RegisterBlock(merge_id, false);
-  cfg_constructs_.emplace_back(get_current_block(), &blocks_.at(merge_id));
+  BasicBlock& merge_block = blocks_.at(merge_id);
+  current_block_->set_type(kBlockTypeHeader);
+  merge_block.set_type(kBlockTypeMerge);
+
+  cfg_constructs_.emplace_back(ConstructType::kSelection, get_current_block(),
+                               &blocks_.at(merge_id));
```
dneto0: Third param can be replaced by the simpler &merge_block

source/val/Construct.h
dneto0: Right.  I see that now in RegisterLoopMerge code.
This table (in ascii art form) is perfect.

source/validate_cfg.cpp
```diff
@@ -143,10 +149,14 @@ const vector<BasicBlock*>* successor(const BasicBlock* b) {
...
 
+const vector<BasicBlock*>* predecessor(const BasicBlock* b) {
+  return b->get_predecessors();
+}
+
 }  // namespace
 
 vector<pair<BasicBlock*, BasicBlock*>> CalculateDominators(
-    vector<cbb_ptr>& postorder) {
+    vector<cbb_ptr>& postorder, get_blocks_func predecessor_func) {
```
dneto0: The prototype and blurb is now out of date in validate.h.
Please update them.

source/validate_cfg.cpp
```diff
@@ -213,9 +223,11 @@ vector<pair<BasicBlock*, BasicBlock*>> CalculateDominators(
   return out;
 }
 
-void UpdateImmediateDominators(vector<pair<bb_ptr, bb_ptr>>& dom_edges) {
+void UpdateImmediateDominators(
```
dneto0: The prototype and blurb is now out of date in validate.h.
Please update them.

source/validate_cfg.cpp
```diff
@@ -244,30 +256,152 @@ spv_result_t FirstBlockAssert(ValidationState_t& _, uint32_t target) {
...
-  if (_.get_current_function().IsMergeBlock(merge_block)) {
+  if (_.get_current_function().IsBlockType(merge_block, kBlockTypeMerge)) {
     return _.diag(SPV_ERROR_INVALID_CFG)
            << "Block " << _.getIdName(merge_block)
            << " is already a merge block for another header";
   }
   return SPV_SUCCESS;
 }
 
+void UpdateContinueConstructExitBlocks(
```
dneto0: Needs a blurb.

source/validate_cfg.cpp
umar456: Would it be better if I removed these functions from the header and put them in the anonymous namespace in the current file? It is not currently being used in other functions.

source/validate_cfg.cpp
```diff
@@ -244,30 +254,154 @@ spv_result_t FirstBlockAssert(ValidationState_t& _, uint32_t target) {
...
     }
 
+    /// Check all backedges target only loop headers and have exactly one
+    /// back-edge branching to it
+    set<uint32_t> loop_headers;
+    for (auto back_edge : back_edges) {
+      uint32_t back_edge_block;
+      uint32_t header_block;
+      tie(back_edge_block, header_block) = back_edge;
+      if (!function.IsBlockType(header_block, kBlockTypeLoop)) {
```
dneto0: Structured control flow is a requirement only if the Shader capability is declared.
Subtle point. :-)

source/validate_cfg.cpp
```diff
@@ -244,30 +254,154 @@ spv_result_t FirstBlockAssert(ValidationState_t& _, uint32_t target) {
...
+        return _.diag(SPV_ERROR_INVALID_CFG)
+               << "Back-edges (" << _.getIdName(back_edge_block) << " -> "
+               << _.getIdName(header_block)
+               << ") can only be formed between a block and a loop header.";
+      }
+      bool success;
+      tie(ignore, success) = loop_headers.insert(header_block);
+      if (!success) {
```
dneto0: Similarly, only if structured control flow is required.

source/validate_cfg.cpp
```diff
@@ -284,41 +418,39 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
...
-      }
-      return _.diag(SPV_ERROR_INVALID_CFG)
-             << "Block(s) " << ss.str() << "\b}"
-             << " are referenced but not defined in function "
-             << _.getIdName(function.get_id());
-    }
-
-    // Check all headers dominate their merge blocks
+    // Check construct rules
     for (Construct& construct : function.get_constructs()) {
```
dneto0: Only check these rules when structured control flow is required.

source/validate_cfg.cpp
```diff
@@ -284,41 +418,39 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
...
-      auto header = construct.get_header();
-      auto merge = construct.get_merge();
-      // auto cont = construct.get_continue();
+      auto header = construct.get_entry();
+      auto merge = construct.get_exit();
 
+      //  for a given loop, its back-edge block must post dominate the
+      //  OpLoopMerge's Continue Target , and that Continue Target must dominate
+      //  that back edge block
```
dneto0: The comment here makes it seem like you're only checking loops.  But this first bit's comment should really be: if the merge block is reachable, then it's dominated by the header.

source/validate_cfg.cpp
```diff
@@ -244,30 +254,154 @@ spv_result_t FirstBlockAssert(ValidationState_t& _, uint32_t target) {
...
+    case ConstructType::kCase:
+      construct_name = "case";
+      header_name = "case block";
+      exit_name = "exit block";  // TODO(umar): there has to be a better name
+      break;
+    default:
+      assert(1 == 0 && "Not defined type");
+  }
+  // TODO(umar): Add header block for continue constructs to error message
+  return "The " + construct_name + " construct with the " + header_name + " " +
```
dneto0: In the post-dominate case, this reads incorrectly, e.g. I get:

The continue construct with the continue target 7 does not post dominate the back-edge block 10

That middle text should be "is not post dominated by"

test/Validate.CFG.cpp
```diff
@@ -707,9 +707,178 @@ TEST_F(ValidateCFG, DISABLED_BackEdgeBlockDoesntPostDominateContinueTargetBad) {
...
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_CFG, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              MatchesRegex("The continue construct with the continue target "
+                           ".\\[loop\\] does not post dominate the "
+                           "back-edge block .\\[cont\\]"));
 }
 
 /// TODO(umar): Switch instructions
```
dneto0: Existing tests that check structured control flow rules should all use OpCapabilityShader.
Need at least one test with unstructured code, e.g. a loop with two backedges, and no OpLoopMerge or OpSelectionMerge instructions.

test/Validate.CFG.cpp
```diff
@@ -707,9 +707,178 @@ TEST_F(ValidateCFG, DISABLED_BackEdgeBlockDoesntPostDominateContinueTargetBad) {
 
   CompileSuccessfully(str);
   ASSERT_EQ(SPV_ERROR_INVALID_CFG, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              MatchesRegex("The continue construct with the continue target "
+                           ".\\[loop2\\] does not dominate the back-edge block "
```
dneto0: This isn't what the test name says it's going to test.
This test is failing on a dominance test, but the test name says it's going to check post-dominance.

test/Validate.CFG.cpp
```diff
@@ -707,9 +707,178 @@ TEST_F(ValidateCFG, DISABLED_BackEdgeBlockDoesntPostDominateContinueTargetBad) {
...
+                           ".\\[loop1_merge\\]"));
+}
+
+TEST_F(ValidateCFG, BranchingToNonLoopHeaderBlockBad) {
+  Block entry("entry");
```
dneto0: This is actually ok when not Shader.
This test should add OpCapability Shader

test/Validate.CFG.cpp
```diff
@@ -707,9 +707,178 @@ TEST_F(ValidateCFG, DISABLED_BackEdgeBlockDoesntPostDominateContinueTargetBad) {
...
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_CFG, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              MatchesRegex("Back-edges \\(.\\[f\\] -> .\\[split\\]\\) can only "
+                           "be formed between a block and a loop header."));
+}
+
+TEST_F(ValidateCFG, BranchingToSameNonLoopHeaderBlockBad) {
+  Block entry("entry");
```
dneto0: Similar.  Add Shader capability.

test/Validate.CFG.cpp
```diff
@@ -707,9 +707,178 @@ TEST_F(ValidateCFG, DISABLED_BackEdgeBlockDoesntPostDominateContinueTargetBad) {
...
+  str += exit;  //  Branches out of a continue construct
+  str += be_block >> loop;
+  str += merge;
+  str += "OpFunctionEnd";
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_CFG, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              MatchesRegex("The continue construct with the continue target "
+                           ".\\[cheader\\] does not post dominate the "
```
dneto0: This is where the "post-dominate" error message is shown to be backward.

test/Validate.CFG.cpp
```diff
@@ -707,9 +707,178 @@ TEST_F(ValidateCFG, DISABLED_BackEdgeBlockDoesntPostDominateContinueTargetBad) {
...
+  str += "OpFunctionEnd";
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_CFG, ValidateInstructions());
+  EXPECT_THAT(
+      getDiagnosticString(),
+      MatchesRegex("Loop header .\\[loop\\] targeted by multiple back-edges"));
+}
+
+TEST_F(ValidateCFG, BranchOutWithReturnBad) {
```
dneto0: Bad test name.  Should be ContinueTargetMustBePostDominatedByBackEdge or similar.

test/Validate.CFG.cpp
```diff
@@ -707,9 +707,178 @@ TEST_F(ValidateCFG, DISABLED_BackEdgeBlockDoesntPostDominateContinueTargetBad) {
...
+  str += cont >> vector<Block>({loop, merge});
+  str += merge;
+  str += "OpFunctionEnd";
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_CFG, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              MatchesRegex("The continue construct with the continue target "
+                           ".\\[loop\\] does not post dominate the "
+                           "back-edge block .\\[cont\\]"));
```
dneto0: Expect the message to change. Another post-dominate case.

dneto0:
Ok. Phew!
Two bugs:
 - construct error wording is wrong in the post-dominate case
- structured control flow rules are only checked when OpCapability Shader is used.  You *might* want to abstract that to a property of the validation state. In future we might have other situations that check for structured control flow rules.

Then coding stuff:  naming of tests is not quite right sometimes.
Other more minor stuff about comments.

umar456:
Addressed feedback

source/validate_cfg.cpp
```diff
@@ -328,8 +331,70 @@ string ConstructErrorString(Construct construct, string header_string,
...
   return "The " + construct_name + " construct with the " + header_name + " " +
-         header_string + " does not " + dominate_text + " the " + exit_name +
-         " " + exit_string;
+         header_string + " " + dominate_text + " the " + exit_name + " " +
+         exit_string;
+}
+
+spv_result_t StructuredControlFlowChecks(
+    ValidationState_t& _, Function function,
+    vector<pair<uint32_t, uint32_t>> back_edges) {
```
dneto0: Should be const ref to save the copy.

test/Validate.CFG.cpp
```diff
@@ -160,9 +160,17 @@ Block& operator>>(Block& lhs, Block& successor) {
   return lhs;
 }
 
-string header =
-    "OpCapability Shader\n"
-    "OpMemoryModel Logical GLSL450\n";
+string header(SpvCapability cap) {
+  static string *shader_header = new string(
```
dneto0: No reason to leak these.
Can just make these const char* then allow the implicit return with:
```
return (cap == SpvCapabilityShader) ? shader_header : kernel_header;
```

source/validate_cfg.cpp
```diff
@@ -337,7 +337,7 @@ string ConstructErrorString(Construct construct, string header_string,
 
 spv_result_t StructuredControlFlowChecks(
     ValidationState_t& _, Function function,
```
dneto0: function shoudld be const ref too.  :-(
It has a bunch of internal state.

dneto0:
I'm testing with 0d42c07 on vk-build-programs from the latest Vulkan CTS.  The validator dominator code seems to get stuck on one of them.  I'm investigating.  I don't even know which shader it is.

dneto0:
Thanks for the last fixes.  Pending testing results with Vulkan CTS.  I'm still investigating.

dneto0:
It seems like we get an infinite loop with dEQP test case
dEQP-VK.spirv_assembly.instruction.compute.opphi.block, attached
[opphi.spvasm.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/326768/opphi.spvasm.txt)

umar456:
I see. I will take a look.

dneto0:
(No rush.  I'm done for the day.)

Stopping it at random, the stack trace is
```
.... __k=@0x2d39fb8: 0x2d3baf0) at /usr/include/c++/4.8/bits/hashtable_policy.h:488
#4  0x0000000001524c15 in std::unordered_map<const libspirv::BasicBlock*, libspirv::CalculateDominators(const std::vector<const libspirv::BasicBlock*>&, libspirv::get_blocks_func)::block_detail, std::hash<const libspirv::BasicBlock*>, std::equal_to<const libspirv::BasicBlock*>, std::allocator<std::pair<const libspirv::BasicBlock* const, libspirv::CalculateDominators(const std::vector<const libspirv::BasicBlock*>&, libspirv::get_blocks_func)::block_detail> > >::operator[](const std::unordered_map<const libspirv::BasicBlock*, libspirv::CalculateDominators(const std::vector<const libspirv::BasicBlock*>&, libspirv::get_blocks_func)::block_detail, std::hash<const libspirv::BasicBlock*>, std::equal_to<const libspirv::BasicBlock*>, std::allocator<std::pair<const libspirv::BasicBlock* const, libspirv::CalculateDominators(const std::vector<const libspirv::BasicBlock*>&, libspirv::get_blocks_func)::block_detail> > >::key_type &) (this=0x7ffe045e1080, __k=@0x2d39fb8: 0x2d3baf0) at /usr/include/c++/4.8/bits/unordered_map.h:596
#5  0x000000000152222e in libspirv::CalculateDominators(std::vector<libspirv::BasicBlock const*, std::allocator<libspirv::BasicBlock const*> > const&, std::function<std::vector<libspirv::BasicBlock*, std::allocator<libspirv::BasicBlock*> > const* (libspirv::BasicBlock const*)>) (postorder=std::vector of length 6, capacity 8 = {...}, predecessor_func=...)
    at ../external/spirv-tools/src/source/validate_cfg.cpp:198
#6  0x0000000001523d9a in libspirv::PerformCfgChecks (_=...) at ../external/spirv-tools/src/source/validate_cfg.cpp:440
#7  0x000000000151b701 in spvValidate (context=0x2d38210, binary=0x7ffe045e1a20, pDiagnostic=0x7ffe045e1a10)
    at ../external/spirv-tools/src/source/validate.cpp:206
```

dneto0:
+2 thanks.   I'll squash and merge.

dneto0:
Squashed and pushed into master as f61db0bcc

Issue #233, opened 2016-06-14
Assembler, bin parser: Extended instructions allowed to have too many parameters


dneto0:
The parser is allowing an extended instruction to have too many parameters.
For example, the OpenCL pow instruction is only supposed to have two normal arguments, but the following is accepted by the assembler, and the result is happily disassembled by the binary-parser/disassembler.
```
          %1 = OpExtInstImport "OpenCL.std"
          %blam  = OpExtInst %a %1 pow %x %y %z
          OpNop
```
The %z argument should have triggered an error.

dneto0:
Also affects GLSL.std.450:
```
          %1 = OpExtInstImport "GLSL.std.450"
          %blam  = OpExtInst %a %1 Round %2 %3 %5 %7 %8
          OpNop
```
The Round extended instruction in GLSL is only supposed to take one normal argument.

dneto0:
I traced it to the SPIR-V core instruction grammar file:  For OpExtInst, there is an argument following the LiteralExtInstInteger operand whic his IdRef with quantifier "*".  In the assembler and binary parser, the value of the extended instruction integer determines the expected operands, but then the IdRef-quantifier-* is still on the expected-operands stack.

In other cases, e.g. OpImageFetch, the ImageOperands bit mask fully determines the operands to follow and the grammar does *not* include them in the ImageFetch instruction listing.   It's arguable whether the same rule should apply in the case of OpExtInst.   Since we've already published the grammar, and it's arguable how much checking other tools should have to do, I'm ok just patching this in the SPIRV-Tools setup phase.

PR #235, opened 2016-06-14
Fix ExtInst parsing: no IdRef* at end

dneto0:
The operands following the extended instruction literal
number are determined by the extended instruction itself.
So drop the zero-or-more IdRef pattern at the end of OpExtInst.

It's arguable whether this should actually be a grammar fix.  I've
chosen to patch this in SPIRV-Tools instead of in the grammar file.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/233

Also fix two test cases for OpenCL extended instructions.  These
errors of supplying too many operands are now detected.


dekimir:
+2

AWoloszyn:
+2

PR #236, opened 2016-06-17
Return ptr instead of ref for get_current_function

umar456:
This PR changes the return type for the `get_current_function` from a reference to a pointer.


dneto0:
This change isn't necessary, since you can always use ValidationState_t::is_in_function() instead of a null check against the proposed get_current_function().
This change weakens the implied contract of the code.  I'd rather it not go in.

umar456:
Sounds reasonable.

I wanted to do this to remove the necessity for that function. I had made a similar change in a previous commit to get_current_block and thought I would make this consistent

PR #237, opened 2016-06-17
Check ID definitions appear in dominator of use

umar456:
depends on #236 

This PR checks if definitions of ID appear in the dominators of the blocks of where they are used.

Takes into account of phi instructions that may appear in the header of a loop.

### DIFF
--- a/source/val/ValidationState.cpp
+++ b/source/val/ValidationState.cpp
@@ -280,9 +280,8 @@ DiagnosticStream ValidationState_t::diag(spv_result_t error_code) const {
 
 list<Function>& ValidationState_t::get_functions() { return module_functions_; }
 
-Function& ValidationState_t::get_current_function() {
-  assert(in_function_body());
-  return module_functions_.back();
+Function* ValidationState_t::get_current_function() {
+  return (in_function_) ? &module_functions_.back() : nullptr;
 }
 
 bool ValidationState_t::in_function_body() const { return in_function_; }
--- a/source/val/ValidationState.cpp
+++ b/source/val/ValidationState.cpp
@@ -28,11 +28,19 @@
 
 #include <cassert>
 
+#include <algorithm>
+#include <utility>
+#include <vector>
+
 #include "val/BasicBlock.h"
 #include "val/Construct.h"
 #include "val/Function.h"
 
+using std::copy;
 using std::list;
+using std::pair;
+using std::set_difference;
+using std::sort;
 using std::string;
 using std::vector;
 
@@ -199,6 +207,7 @@ ValidationState_t::ValidationState_t(spv_diagnostic* diagnostic,
       current_layout_section_(kLayoutCapabilities),
       module_functions_(),
       module_capabilities_(0u),
+      usedefs_(this),
       grammar_(context),
       addressing_model_(SpvAddressingModelLogical),
       memory_model_(SpvMemoryModelSimple),
@@ -354,6 +363,93 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
          "ouside of a block");
   in_function_ = false;
   return SPV_SUCCESS;
+}
+
+ValidationState_t::UseDefTracker::UseDefTracker(ValidationState_t* state)
+    : module_(state) {}
+
+void ValidationState_t::UseDefTracker::AddDef(const spv_id_info_t& def) {
+  all_defs_[def.id] = def;
+  if (module_->in_block()) {
+    auto block = module_->get_current_function()->get_current_block();
+    block_defs_[block->get_id()].push_back(def.id);
+  } else {
+    module_defs_.insert(def.id);
+  }
+}
+
+void ValidationState_t::UseDefTracker::AddUse(uint32_t id, bool is_phi) {
+  all_uses_.insert(id);
+  if (module_->in_block()) {
+    auto block = module_->get_current_function()->get_current_block();
+    block_uses_[block->get_id()].insert({id, is_phi});
+  }
+}
+
+std::pair<bool, spv_id_info_t> ValidationState_t::UseDefTracker::FindDef(
+    uint32_t id) const {
+  if (all_defs_.count(id) == 0) {
+    return std::make_pair(false, spv_id_info_t{});
+  } else {
+    /// We are in a const function, so we cannot use defs.operator[]().
+    /// Luckily we know the key exists, so defs_.at() won't throw an
+    /// exception.
+    return std::make_pair(true, all_defs_.at(id));
+  }
+}
+
+std::unordered_set<uint32_t>
+ValidationState_t::UseDefTracker::FindUsesWithoutDefs() const {
+  auto diff = all_uses_;
+  for (const auto d : all_defs_) diff.erase(d.first);
+  return diff;
+}
+
+void ValidationState_t::UseDefTracker::UpdateDefs(Function* func) {
+  auto& ordered_blocks = func->get_blocks();
+
+  // Because dominators must appear before the blocks they dominate in the
+  // binary , we only need to visit the immidiate dominators to propagate the
+  // ids throughout the CFG
+  for (auto block : ordered_blocks) {
+    auto& block_defs = block_defs_[block->get_id()];
+    auto idom = block->GetImmediateDominator();
+    if (block->is_reachable() && block != idom) {
+      assert(idom);
+      auto& idom_defs = block_defs_[idom->get_id()];
+      block_defs.reserve(block_defs.size() + idom_defs.size());
+      copy(begin(idom_defs), end(idom_defs), back_inserter(block_defs));
+    } else {
+      // This is the first block or it is unreachable in the CFG
+      block_defs.reserve(block_defs.size() + module_defs_.size());
+      copy(begin(module_defs_), end(module_defs_), back_inserter(block_defs));
+    }
+    sort(begin(block_defs), end(block_defs));
+  }
+}
+
+std::map<uint32_t, vector<uint32_t>>
+ValidationState_t::UseDefTracker::CheckDefs(Function* func) {
+  std::map<uint32_t, vector<uint32_t>> undef_ids;
+
+  for (auto block : func->get_blocks()) {
+    auto& block_defs = block_defs_[block->get_id()];
+    auto& block_uses = block_uses_[block->get_id()];
+    vector<uint32_t> uses_vec;
+    uses_vec.reserve(block_uses.size());
+    for (auto& use : block_uses) {
+      if (use.second == false) {
+        uses_vec.push_back(use.first);
+      }
+    }
+    set_difference(begin(uses_vec), end(uses_vec), begin(block_defs),
+                   end(block_defs), back_inserter(undef_ids[block->get_id()]));
+    if (undef_ids[block->get_id()].empty()) {
+      undef_ids.erase(block->get_id());
+    }
+  }
+
+  return undef_ids;
 }
 
 }  /// namespace libspirv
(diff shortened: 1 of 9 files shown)


source/val/ValidationState.h
```diff
@@ -42,17 +43,24 @@
...
 // Info about a result ID.
-typedef struct spv_id_info_t {
+struct spv_id_info_t {
   /// Id value.
   uint32_t id;
   /// Type id, or 0 if no type.
   uint32_t type_id;
   /// Opcode of the instruction defining the id.
   SpvOp opcode;
+  /// The block where the id was defined. kInvalidId if module scoped
```
dneto0: Why not use ID value of 0.  That's never a valid ID in SPIR-V.

source/val/ValidationState.cpp
```diff
@@ -357,4 +365,91 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
   return SPV_SUCCESS;
 }
 
+ValidationState_t::UseDefTracker::UseDefTracker(ValidationState_t* state)
+    : module_(state) {}
+
+void ValidationState_t::UseDefTracker::AddDef(const spv_id_info_t& def) {
+  all_defs_[def.id] = def;
+  if (module_->in_block()) {
```
dneto0: The spv_id_info_t already has the ID of the defining block, if any.  So you can just check that instead of poking at the module_ member.   That would be more self-contained.

source/val/ValidationState.cpp
```diff
@@ -357,4 +365,91 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
...
+    auto block = module_->get_current_function()->get_current_block();
+    block_defs_[block->get_id()].push_back(def.id);
+  } else {
+    module_defs_.insert(def.id);
+  }
+}
+
+void ValidationState_t::UseDefTracker::AddUse(uint32_t id, bool is_phi) {
+  all_uses_.insert(id);
+  if (module_->in_block()) {
```
dneto0: This is fine.
However, in the back of my mind I'm concerned with the coupling to the validation state.
Alternately, this method could take another argument to name the ID of the defining block, or 0 if outside of a block.

Then together with the suggested change to AddDef we would remove the need to point back to the validation state.  That would make the use-def-tracker more self-contained and hence easier to test.
For another CL.

source/val/ValidationState.cpp
```diff
@@ -357,4 +365,91 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
...
+
+std::unordered_set<uint32_t>
+ValidationState_t::UseDefTracker::FindUsesWithoutDefs() const {
+  auto diff = all_uses_;
+  for (const auto d : all_defs_) diff.erase(d.first);
+  return diff;
+}
+
+void ValidationState_t::UseDefTracker::UpdateDefs(Function* func) {
+  auto& ordered_blocks = func->get_blocks();
```
dneto0: Need to update the blurb for get_blocks to say it returns them in an order that respects dominance.  Or in the order they appear in the module. (or both?)

source/val/ValidationState.cpp
```diff
@@ -357,4 +365,91 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
...
+
+  // Because dominators must appear before the blocks they dominate in the
```
dneto0: Is there  a check that confirms that block order respects dominance?

Also "immidiate" -> "immediate"

source/val/ValidationState.cpp
```diff
@@ -357,4 +365,91 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
...
+  // binary , we only need to visit the immidiate dominators to propagate the
+  // ids throughout the CFG
+  for (auto block : ordered_blocks) {
+    auto& block_defs = block_defs_[block->get_id()];
+    auto idom = block->GetImmediateDominator();
+    if (block->is_reachable() && block != idom) {
```
dneto0: I'm a little fuzzy on whether this idom variable is always non-null.  I think it is, once I've traced back through the various algorithms.  Could do with an assert here.

Normally the idom of a node strictly dominates the node, which means it can't be the node.
For example, is the idom of the entry block the entry block itself?
The blurb on CalculateDominators is too vague to tell this apart.

dneto0: Never mind. I was misreading this code....

source/val/ValidationState.cpp
```diff
@@ -357,4 +365,91 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
...
+      assert(idom);
+      auto& idom_defs = block_defs_[idom->get_id()];
+      block_defs.reserve(block_defs.size() + idom_defs.size());
+      copy(begin(idom_defs), end(idom_defs), back_inserter(block_defs));
+    } else {
+      // This is the first block or it is unreachable in the CFG
+      block_defs.reserve(block_defs.size() + module_defs_.size());
+      copy(begin(module_defs_), end(module_defs_), back_inserter(block_defs));
```
dneto0: This is not quite right.
OpFunctionParameter values are defined outside any block, so they incorrectly end up in the module_defs_ bucket.
So this code will allow an OpFunctionParameter from one function definition to be used in the body of a later function definition.

Example:

```
lOpCapability Shader
OpMemoryModel Logical Simple

%void = OpTypeVoid
%i32 = OpTypeInt 32 1
%voidfn = OpTypeFunction %void %i32

%f0 = OpFunction %void None %voidfn
%a0 = OpFunctionParameter %i32
%entry0 = OpLabel
OpReturn
OpFunctionEnd

%f1 = OpFunction %void None %voidfn
%a1 = OpFunctionParameter %i32
%entry1 = OpLabel
%oops = OpCopyObject %i32 %a0  ; this should not be allowed
OpReturn
OpFunctionEnd
```

source/val/ValidationState.h
```diff
@@ -138,36 +144,58 @@ class ValidationState_t {
...
-      } else {
-        /// We are in a const function, so we cannot use defs.operator[]().
-        /// Luckily we know the key exists, so defs_.at() won't throw an
-        /// exception.
-        return std::make_pair(true, defs_.at(id));
-      }
-    }
+    std::pair<bool, spv_id_info_t> FindDef(uint32_t id) const;
+
+    /// Propagates the definitions of IDs from the dominators to the blocks they
```
dneto0: The blurb is a little confusing.  I'd say

"Updates bookkeeping to record that an ID definition in a given block is visible in blocks it dominates."

The note should also say that it should only be run once, otherwise you'll get redundant data.
?

source/validate.cpp
```diff
@@ -125,18 +125,24 @@ void DebugInstructionPass(ValidationState_t& _,
...
       break;
   }
-
 }
 
 // Collects use-def info about an instruction's IDs.
 void ProcessIds(ValidationState_t& _, const spv_parsed_instruction_t& inst) {
   if (inst.result_id) {
+    uint32_t curr_block = libspirv::kInvalidId;
+    if (_.get_current_function() &&
```
dneto0: The test against _.get_current_function() could be replaced with _.in_function_body()

source/val/ValidationState.h
```diff
@@ -138,36 +144,58 @@ class ValidationState_t {
...
+    /// dominate for a given function
+    ///
+    /// NOTE: This function requires that the CFG dominators are correctly set
+    /// for each block in the function
+    void UpdateDefs(Function* func);
+
+    /// Returns a map of block IDs that incorrectly used an operand that was not
+    /// first defined by its dominator blocks.
```
dneto0: Return value is unclear.
Is the key the ID that's badly used?  And what's the value for that key?

source/val/ValidationState.h
umar456: Huh. I didn't realize that. It will make things easier.

source/val/ValidationState.cpp
```diff
@@ -357,4 +365,91 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
...
+  std::map<uint32_t, vector<uint32_t>> undef_ids;
+
+  for (auto block : func->get_blocks()) {
+    auto& block_defs = block_defs_[block->get_id()];
+    auto& block_uses = block_uses_[block->get_id()];
+    vector<uint32_t> uses_vec;
+    uses_vec.reserve(block_uses.size());
+    for (auto& use : block_uses) {
+      if (use.second == false) {
+        uses_vec.push_back(use.first);
```
dneto0: Something weird here.
An OpBranch at the end of a block counts as a "use" the label ID it branches to.
But in a case such as the following:
```
               OpCapability Shader
               OpMemoryModel Logical Simple
          %1 = OpTypeVoid
          %2 = OpTypeInt 32 1
          %3 = OpTypeFunction %1 %2
          %4 = OpFunction %1 None %3
          %5 = OpLabel
               OpBranch %6
          %6 = OpLabel
               OpReturn
               OpFunctionEnd
```
This code will say that ID 6 (the label of the last block) is used by block 5 (the first block), but then doesn't say it's a bad use.
When I add prints, it says that block 5 "defines" ID 6.  That can't be right.

source/val/ValidationState.cpp
umar456: Yes, its done [here](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/validate_cfg.cpp#L443)

source/val/ValidationState.cpp
dneto0: Ah.  I think label IDs are not counted as definitions, since the validation state is only "in a block" *after* the label has been fully processed.  That's a weird accident because the ProcessIds handler is called before CfgPass.
And so the "Current block" is not registered at the moment you see the label.  So then the label ID is put into the module scope bucket.  That's *really* confusing.  For one thing, the block_defs_ vectors will all be bloated without good reason, and the module_defs_ is incorrectly named.

I think we need to be more careful about partitioning the data.

dneto0:
I'm concerned about the way block_defs_ ends up storing a lot of data.  The number of elements is going to be roughly number of blocks times the number of ids.   It's kind of redundant to have every "module scope" ID in every block's list.
An alternate design is to have a function that answers the dominance relation between two *block IDs*.  Then to see if an ID id1 is badly used, map id1 to the block BUse that uses it, and find the block BDef that defines id1, and then check if BDef dominates BUse.  The table backing the block-dominance function is going to be very small.  (You'd need to handle the corner cases:  ID defined outside of all functions, ID defined as function parameter to my own function, ID defined as func parameter to other functions.)

For me that alternate design would be more clearly correct, and probably easier to work with.

dneto0:
As long as it's documented, you could take care of one special case:  Function parameters could  be considered as being "defined" by the entry block to their own function.

source/val/ValidationState.h
dneto0: Yeah, but my recent PR for psuedo-entry node cheats and uses Id 0 for the psuedo-entry node.   So please keep this one.  :-)

umar456:
I agree that it could end up storing a lot of redundant data. I am going to close this PR and implementing this using another approach.

dneto0:
Thanks.  To expand a bit:   Yeah, formally there's a time-space tradeoff here in terms of number of operations.  But using very small data means you'll probably get much better locality and may end up being faster.

umar456:
I understand.

I was worried since the blocks are allocated in the unordered map, doing a traversal of the dominator tree would result in several cache misses for each use. I was assuming that since the set operations in this algorithm have a fairly predictive memory access pattern, it would outperform the other approach. But like you mentioned in your approach has a better chance of fitting in the cache so it might outperform this approach even with the irregular accesses. I could also memoise the traversal so I don't take a hit for each traversal.

PR #238, opened 2016-06-21
Add classes for representing SPIR-V constructs in memory.

antiagainst:
The first offspring from #219.

### DIFF
--- a/source/opt/basic_block.h
+++ b/source/opt/basic_block.h
@@ -1,0 +1,87 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+// This file defines the language constructs for representing a SPIR-V
+// module in memory.
+
+#ifndef LIBSPIRV_OPT_BASIC_BLOCK_H_
+#define LIBSPIRV_OPT_BASIC_BLOCK_H_
+
+#include <functional>
+#include <vector>
+#include <utility>
+
+#include "instruction.h"
+
+namespace spvtools {
+namespace opt {
+namespace ir {
+
+class Function;
+
+// A SPIR-V basic block.
+class BasicBlock {
+ public:
+  // Creates a basic block with the given enclosing |function| and starting
+  // |label|.
+  BasicBlock(Instruction&& label)
+      : function_(nullptr), label_(std::move(label)) {}
+
+  // Sets the enclosing function for this basic block.
+  void SetParent(Function* function) { function_ = function; }
+  // Appends an instruction to this basic block.
+  void AddInstruction(Instruction&& i) { insts_.push_back(std::move(i)); }
+
+  // Runs the given function |f| on each instruction in this basic block.
+  inline void ForEachInst(const std::function<void(Instruction*)>& f);
+
+  // Pushes the binary segments for this instruction into the back of *|binary|.
+  // If |skip_nop| is true and this is a OpNop, do nothing.
+  inline void ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const;
+
+ private:
+  Function* function_;              // The enclosing function.
+  Instruction label_;               // The label starting this basic block.
+  std::vector<Instruction> insts_;  // Instructions inside this basic block.
+};
+
+inline void BasicBlock::ForEachInst(
+    const std::function<void(Instruction*)>& f) {
+  label_.ForEachInst(f);
+  for (auto& inst : insts_) f(&inst);
+}
+
+inline void BasicBlock::ToBinary(std::vector<uint32_t>* binary,
+                                 bool skip_nop) const {
+  label_.ToBinary(binary, skip_nop);
+  for (const auto& inst : insts_) inst.ToBinary(binary, skip_nop);
+}
+
+}  // namespace ir
+}  // namespace opt
+}  // namespace spvtools
+
+#endif  // LIBSPIRV_OPT_BASIC_BLOCK_H_
--- a/source/opt/reflect.h
+++ b/source/opt/reflect.h
@@ -1,0 +1,65 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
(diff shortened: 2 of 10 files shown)


source/opt/constructs.h
```diff
@@ -0,0 +1,306 @@
...
+namespace spvtools {
+namespace opt {
+namespace ir {
+
+class Function;
+class Module;
+
+// About operand:
+//
+// In the SPIR-V specification, the term "operand" is used to mean any single
```
dneto0: Good explanation!

source/opt/constructs.h
```diff
@@ -0,0 +1,306 @@
...
+  // instruction, if any.
+  Instruction(const spv_parsed_instruction_t& inst,
+              std::vector<Instruction>&& dbg_line = {});
+
+  SpvOp opcode() const { return opcode_; }
+  uint32_t type_id() const { return type_id_; }
+  uint32_t result_id() const { return result_id_; }
+  // Returns the vector of line-related debug instructions attached to this
+  // instruction and the caller can directly modify them.
+  std::vector<Instruction>& dbg_line_insts() { return dbg_line_insts_; }
```
dneto0: Do we want a const form of this accessor in case we want this data from a const Instruction?

source/opt/constructs.h
```diff
@@ -0,0 +1,306 @@
...
+    return static_cast<uint32_t>(operands_.size());
+  }
+  // Gets the number of SPIR-V words occupied by all logical operands.
+  uint32_t NumOperandWords() const {
+    return NumInOperandWords() + TypeResultIdCount();
+  }
+  // Gets the |index|-th logical operand.
+  const Operand& GetOperand(uint32_t index) const;
+  // Gets the |index|-th logical operand as a single SPIR-V word. This will only
+  // consider the first word in the logical operand, even which has more than
```
dneto0: "even which" -> "even when it"

"one words" -> "one word"

source/opt/constructs.h
```diff
@@ -0,0 +1,306 @@
...
+    type_id_ = result_id_ = 0;
+    operands_.clear();
+  }
+
+  // Runs the given function |f| on this instruction and preceding debug
+  // instructions.
+  void ForEachInst(const std::function<void(Instruction*)>& f);
+
+  // Pushes the binary segments for this instruction into the back of *|binary|.
+  // If |keep_nop| is true and this is a OpNop, do nothing.
```
dneto0: The name keep_op is confusing.  It seems to mean the opposite of the description.
Maybe call it skip_nop.  ?

dneto0:
The constructs.h and constructs.cpp filename is really confusing.  The validator has a concept of "construct" that matches the SPIR-V spec structured control flow construct.  That's a strong concept that we should not confuse here.
Either split up these files one per main class, or rename the file to something like "ir.h" and "ir.cpp".
(I *tend* to like one per class.  But coder's choice at this point.)  Note that you might end up with a warning from "ar" on  Darwin when file basenames collide.  That will happen with source/val/function.cpp and source/opt/function.cpp...

source/opt/constructs.h
antiagainst: Done.

source/opt/constructs.h
antiagainst: Done.

source/opt/constructs.h
antiagainst: Good catch. Actually it took me a while to understand that this is indeed reversed. Changed.

antiagainst:
@dneto0 Renamed the files to `ir.*`. I won't actually bother splitting it into multiple files since each class is so small right now. We can do that latter.

source/opt/reflect.h
```diff
@@ -0,0 +1,61 @@
...
+#ifndef LIBSPIRV_OPT_REFLECT_H_
+#define LIBSPIRV_OPT_REFLECT_H_
+
+#include "spirv/1.1/spirv.h"
+
+namespace spvtools {
+namespace opt {
+namespace ir {
+
+inline bool IsDebugInst(SpvOp opcode) {
```
dneto0: It might be worth a note saying that over time as SPIR-V evolves, you might get opcodes that don't fit neatly into ranges.  For example, in the future you might get new kinds of type instructions and they will have opcodes far different (with other opcodes in between).

source/opt/ir.cpp
```diff
@@ -0,0 +1,177 @@
...
+    const auto& current_payload = inst.operands[i];
+    std::vector<uint32_t> words(
+        inst.words + current_payload.offset,
+        inst.words + current_payload.offset + current_payload.num_words);
+    operands_.emplace_back(current_payload.type, std::move(words));
+  }
+}
+
+const Operand& Instruction::GetOperand(uint32_t index) const {
+  if (index >= operands_.size()) {
```
dneto0: It's more usual to write something like this:

```
assert( index < operands_.size() && "operand index out of bounds");
```

source/opt/ir.cpp
```diff
@@ -0,0 +1,177 @@
...
+    assert(0 && "operand index out of bound");
+  }
+
+  return operands_[index];
+};
+
+uint32_t Instruction::GetSingleWordOperand(uint32_t index) const {
+  if (index >= operands_.size()) {
+    // TODO(antiagainst): do better than panicking.
```
dneto0: Why do something better than panic?  All clients of this code will be other code, and user input has already been sanitized.

source/opt/ir.cpp
```diff
@@ -0,0 +1,177 @@
...
+    assert(0 && "operand index out of bound");
+  }
+  if (operands_[index].words.size() != 1) {
```
dneto0: The blurb for this method implies that it's usable with logical operands having more than one word.    I guess you should fix the blurb?

source/opt/ir.cpp
```diff
@@ -0,0 +1,177 @@
...
+    assert(0 && "expected the operand only taking one word");
+  }
+
+  return operands_[index].words.front();
```
dneto0: Could have simplified the code by writing:
```
const auto& words = GetOperand(index); // this will do the index assert for you.
...
```

source/opt/ir.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+uint32_t Instruction::NumInOperandWords() const {
+  uint32_t size = 0;
+  for (uint32_t i = TypeResultIdCount(); i < operands_.size(); ++i)
+    size += static_cast<uint32_t>(operands_[i].words.size());
+  return size;
+}
+
+void Instruction::ForEachInst(const std::function<void(Instruction*)>& f) {
+  f(this);
```
dneto0: Since the debug instructions have to appear before the regular instruction, I expected f(this) to be called last.

source/opt/constructs.h
```diff
@@ -0,0 +1,305 @@
...
+  // empty.
+  std::vector<Instruction> dbg_line_insts_;
+};
+
+// A SPIR-V basic block.
+class BasicBlock {
+ public:
+  // Creates a basic block with the given enclosing |function| and starting
+  // |label|.
+  BasicBlock(Function* function, Instruction&& label)
```
dneto0: Again, since the label is an rvalue reference, should I assume it's implied that this constructor takes ownership of the label instruction?
Should we say so in the blurb?  I'm leaning toward yes.

source/opt/constructs.h
```diff
@@ -0,0 +1,305 @@
...
+// operands are operands that represent ids generated from operations (result
+// type id or result id). For example, for "OpIAdd %rtype %rid %inop1 %inop2",
+// "%inop1" and "%inop2" are in operands, while "%rtype" and "%rid" are out
+// operands.
+
+// A *logical* operand to a SPIR-V instruction. It can be the type id, result
+// id, or other additional operands carried in an instruction.
+struct Operand {
+  Operand(spv_operand_type_t t, std::vector<uint32_t>&& w)
+      : type(t), words(std::move(w)) {}
```
dneto0: Is the std::move here redundant?

dneto0: Oh wait, if the actual argument is an lvalue reference, then by reference collapsing rules, then w itself will be an lvalue reference. In that case you do want the std::move.

source/opt/constructs.h
```diff
@@ -0,0 +1,305 @@
...
+  std::vector<Instruction> dbg_line_insts_;
+};
+
+// A SPIR-V basic block.
+class BasicBlock {
+ public:
+  // Creates a basic block with the given enclosing |function| and starting
+  // |label|.
+  BasicBlock(Function* function, Instruction&& label)
+      : function_(function), label_(label) {}
```
dneto0: Missing std::move on label in the member initialization?

source/opt/reflect.h
antiagainst: Done.

source/opt/ir.cpp
antiagainst: Nice! I don't know why I wrote like that and didn't notice it. :(

source/opt/ir.cpp
antiagainst: Hmm, reasonable. Removed.

source/opt/ir.cpp
antiagainst: Fixed the blurb.

source/opt/ir.cpp
antiagainst: Done.

source/opt/ir.cpp
antiagainst: Good catch! Done.

source/opt/constructs.h
antiagainst: I'm not sure that's necessary. rvalue ref itself is already a signal of moving semantics. The caller should be clear that the passed in object will be moved.

source/opt/constructs.h
antiagainst: That's right.

source/opt/constructs.h
antiagainst: Good catch. Done.

antiagainst:
@dneto0: And I'd like to take back of my previous comment on splitting the `ir.{h|cpp}` file. It seems now is "later". :)

dneto0:
You'll need a corresponding change in google/shaderc in the Android.mk file.

source/opt/basic_block.h
```diff
@@ -0,0 +1,86 @@
...
+
+class Function;
+
+// A SPIR-V basic block.
+class BasicBlock {
+ public:
+  // Creates a basic block with the given enclosing |function| and starting
+  // |label|.
+  BasicBlock(Instruction&& label)
+      : function_(nullptr), label_(std::move(label)) {}
```
dneto0: std::move is from ```<utility>```. Need to include it.

dneto0:
Looks ok except for the missing #include.  (I didn't check exhaustively for other ICWYU violations)

dneto0:
+2
Please wait to submit until there's a matching update to google/shaderc's third_party/Android.mk file.

antiagainst:
Definitely. Working on that.

On Fri, Jun 24, 2016, 6:28 PM David Neto <notifications@github.com> wrote:

> +2
> Please wait to submit until there's a matching update to google/shaderc's
> third_party/Android.mk file.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/238#issuecomment-228478086>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAdx-N8Iecc6qXMWFT0d5MeN-N08JxlNks5qPFoSgaJpZM4I7M4d>
> .
>
-- 
Thanks,
Lei

antiagainst:
Actually this will just work fine with shaderc; we haven't actually references any thing from the opt part in shaderc yet (even in spirv-tools itself).

antiagainst:
So it won't cause a build failure for shaderc. We just don't have `libSPIRV-Tools-opt.so` in `libshaderc_combined.so` if keeping the current `third_party/Android.mk`.

antiagainst:
https://github.com/google/shaderc/pull/229 is put up for `Android.mk` since eventually we'll need it.

But I'd suggest we just put it there and keep updating it until the opt component is fairly complete and ready to be exported via shaderc.

PR #242, opened 2016-06-24
Use pseudo entry and pseudo exit blocks for dominance.

dneto0:
For dominance calculations we use an "augmented" CFG
where we always add a pseudo-entry node that is the predecessor
in the augmented CFG to any nodes that have no predecessors in the
regular CFG.  Similarly, we add a pseudo-exit node that is the
predecessor in the augmented CFG that is a successor to any
node that has no successors in the regular CFG.

Fixes a subtle problem where we were implicitly creating
the block_details for the pseudo-exit node since it didn't
appear in the idoms map, and yet we referenced it.  In such a case the
contents of the block details could be garbage, or zero-initialized.
That sometimes caused incorrect calculation of immediate dominators
and post-dominators.  For example, on a debug build where the details
could be zero-initialized, the dominator of an unreachable block would
be given as the pseudo-exit node.  Bizarre.

### DIFF
--- a/source/val/BasicBlock.h
+++ b/source/val/BasicBlock.h
@@ -123,6 +123,14 @@ class BasicBlock {
   /// Adds @p next BasicBlocks as successors of this BasicBlock
   void RegisterSuccessors(const std::vector<BasicBlock*>& next = {});
 
+  /// Set the successors to this block, without updating other internal state,
+  /// and without updating the other blocks.
+  void SetSuccessorsUnsafe(std::vector<BasicBlock*>&& others);
+
+  /// Set the predecessors to this block, without updating other internal state,
+  /// and without updating the other blocks.
+  void SetPredecessorsUnsafe(std::vector<BasicBlock*>&& others);
+
   /// Returns true if the id of the BasicBlock matches
   bool operator==(const BasicBlock& other) const { return other.id_ == id_; }
 
--- a/source/val/Function.cpp
+++ b/source/val/Function.cpp
@@ -69,9 +69,13 @@ Function::Function(uint32_t id, uint32_t result_type_id,
       result_type_id_(result_type_id),
       function_control_(function_control),
       declaration_type_(FunctionDecl::kFunctionDeclUnknown),
+      end_has_been_registered_(false),
       blocks_(),
       current_block_(nullptr),
+      pseudo_entry_block_(0),
       pseudo_exit_block_(kInvalidId),
+      pseudo_entry_blocks_({&pseudo_entry_block_}),
+      pseudo_exit_blocks_({&pseudo_exit_block_}),
       cfg_constructs_(),
       variable_ids_(),
       parameter_ids_() {}
@@ -206,15 +210,27 @@ void Function::RegisterBlockEnd(vector<uint32_t> next_list,
     next_blocks.push_back(&inserted_block->second);
   }
 
-  if (branch_instruction == SpvOpReturn ||
-      branch_instruction == SpvOpReturnValue) {
-    assert(next_blocks.empty());
-    next_blocks.push_back(&pseudo_exit_block_);
-  }
   current_block_->RegisterBranchInstruction(branch_instruction);
   current_block_->RegisterSuccessors(next_blocks);
   current_block_ = nullptr;
   return;
+}
+
+void Function::RegisterFunctionEnd() {
+  if (!end_has_been_registered_) {
+    end_has_been_registered_ = true;
+
+    // Compute the successors of the pseudo-entry block, and
+    // the predecessors of the pseudo exit block.
+    vector<BasicBlock*> sources;
+    vector<BasicBlock*> sinks;
+    for (const auto b : ordered_blocks_) {
+      if (b->get_predecessors()->empty()) sources.push_back(b);
+      if (b->get_successors()->empty()) sinks.push_back(b);
+    }
+    pseudo_entry_block_.SetSuccessorsUnsafe(std::move(sources));
+    pseudo_exit_block_.SetPredecessorsUnsafe(std::move(sinks));
+  }
 }
 
 size_t Function::get_block_count() const { return blocks_.size(); }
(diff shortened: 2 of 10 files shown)


dneto0:
Hi @umar456 this PR fixes some problems we saw with the switch-with-unreachable-default, but only on Linux debug builds, and Windows builds.  I think it was accidentally passing on Linux and Mac release builds.
The root cause was really subtle, basically because we were asking for the postdom order for the pseudo exit block, and creating garbage bookkeeping.  

The key idea in the fix is to take care of unreachable blocks in a more uniform way, by adding a pseudo-entry node.  It's symmetric with pseudo-exit node. 
Also, the pseudo nodes are never in any regular blocks' predecessor or successor lists.   Instead we temporarily fake an "augmented CFG" only during the dominance calculation.

I'd like reviews from @umar456 and maybe @antiagainst and @AWoloszyn

dneto0:
I also clarified the result of the dominance calculation.
The dominance calculation could use some unit-testing love.

source/val/BasicBlock.h
```diff
@@ -123,6 +123,18 @@ class BasicBlock {
   /// Adds @p next BasicBlocks as successors of this BasicBlock
   void RegisterSuccessors(const std::vector<BasicBlock*>& next = {});
 
+  /// Set the successors to this block, without updating other internal state,
+  /// and without updating the other blocks.
+  void SetSuccessorsUnsafe(std::vector<BasicBlock*>&& others) {
+    successors_ = others;
```
antiagainst: I assume you'll need `std::move(others)` here?

source/val/BasicBlock.h
```diff
@@ -123,6 +123,18 @@ class BasicBlock {
...
+  }
+
+  /// Set the predecessors to this block, without updating other internal state,
+  /// and without updating the other blocks.
+  void SetPredecessorsUnsafe(std::vector<BasicBlock*>&& others) {
+    predecessors_ = others;
```
antiagainst: Same as above.

source/val/Function.cpp
```diff
@@ -231,6 +245,11 @@ vector<BasicBlock*>& Function::get_blocks() { return ordered_blocks_; }
 const BasicBlock* Function::get_current_block() const { return current_block_; }
 BasicBlock* Function::get_current_block() { return current_block_; }
 
+BasicBlock* Function::get_pseudo_entry_block() { return &pseudo_entry_block_; }
```
antiagainst: According to the style guide, simple getter methods should be just named as the corresponding field with trailing `_` stripped. But I see that all the getter functions are prefixed with `get` here. So, maybe for a later CL to convert them all.

source/validate.h
```diff
@@ -68,7 +74,10 @@ using get_blocks_func =
 /// @param[in] predecessor_func Function used to get the predecessor nodes of a
 ///                             block
 ///
-/// @return a set of dominator edges represented as a pair of blocks
+/// @return the dominator tree of the graph, as a vector of pairs of nodes.
+/// The first node is the pair is a node in the graph. The second node in the pair
```
antiagainst: `... in the pair...`

source/validate_cfg.cpp
```diff
@@ -408,45 +392,83 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
...
+    // CFG.  The augmented CFG is guaranteed to have a single source node
+    // (i.e. not having predecessors) and a single exit node (i.e. not
+    // having successors).  However, there might be isolated strongly
+    // connected components that are not reachable by following successors
+    // from the pseudo entry node, and not reachable by following
+    // predecessors from the pseudo exit node.
+
+    // We need vectors to use as the predecessors (in the augmented CFG)
+    // for the source nodes of the original CFG.  It must have a stable
+    // address for the duraction of the calculation.
```
umar456: duraction -> duration

source/validate_cfg.cpp
```diff
@@ -408,45 +392,83 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
...
+    const vector<BasicBlock*> pseudo_entry_vec = {pseudo_entry};
+    // Similarly, we need a vector to be used as the successors (in the
+    // augmented CFG) for sinks in the original CFG.
+    const vector<BasicBlock*> pseudo_exit_vec = {pseudo_exit};
+    // Returns the predecessors of a block in the augmented CFG.
+    auto augmented_predecessor_fn =
+        [pseudo_entry, &pseudo_entry_vec](const BasicBlock* block) {
+          auto predecessors = block->get_predecessors();
+          return (block != pseudo_entry && predecessors->empty())
+                     ? &pseudo_entry_vec
```
antiagainst: Hmm, mac@Travis is complaining about this line: `address of stack memory associated with local variable 'pseudo_entry_vec' returned`. But it seems to work fine on my macOS with AppleClang 7.3. So maybe we need to update the xcode version? https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version

source/val/BasicBlock.h
dneto0: Yes, indeed, you're right. Since the "others" variable is an lvalue itself.  Good catch!  (rvalue ref stuff is really subtle!)

source/val/Function.cpp
dneto0: Agreed.  I've done it this way for consistency with the surrounding code for now.
I already raise this with Umar on a previous review, but it can be deferred.

source/validate_cfg.cpp
dneto0: That's 100% deliberate.  I can also work around it by having that particular vector be served up by the Function object.

source/validate.cpp
```diff
@@ -185,6 +185,11 @@ spv_result_t spvValidate(const spv_const_context context,
                                 binary->wordCount, setHeader,
                                 ProcessInstruction, pDiagnostic));
 
+  // Finish analyzing functions, in case a final OpFunctionEnd was not supplied.
+  for (auto& f : vstate.get_functions()) {
+    f.RegisterFunctionEnd();
+  }
```
umar456: It would be better to call this in ValidationState_t::RegisterFunctionEnd(). I don't see anything in the spec that allows you to end a function without OpFunctionEnd.

dneto0: Agreed the spec requires an OpFunctionEnd.  I noticed some tests didn't have it probably because we don't check for it yet?
I'll move it.

umar456: Hmm I see. I can add explicit checks for this in another PR.

antiagainst:
+1

source/validate.cpp
dneto0: I've just added them.   :-)

dneto0:
Addressed all feedback.

umar456:
+2

dneto0:
Squashed.

dneto0:
Thanks for the reviews!

PR #246, opened 2016-06-27
Classes & tests for in-memory IR builder

antiagainst:
A part of #219.

### DIFF
--- a/source/opt/ir_loader.h
+++ b/source/opt/ir_loader.h
@@ -1,0 +1,83 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#ifndef LIBSPIRV_OPT_IR_LOADER_H_
+#define LIBSPIRV_OPT_IR_LOADER_H_
+
+#include <memory>
+
+#include "basic_block.h"
+#include "instruction.h"
+#include "module.h"
+#include "spirv-tools/libspirv.h"
+
+namespace spvtools {
+namespace opt {
+namespace ir {
+
+// Loader class for constructing SPIR-V in-memory IR representation. Methods in
+// this class are designed to work with the interface for spvBinaryParse() in
+// libspirv.h so that we can leverage the syntax checks implemented behind it.
+//
+// The user is expected to call SetModuleHeader() to fill in the module's
+// header, and then AddInstruction() for each decoded instruction, and finally
+// EndModule() to finalize the module. The instructions processed in sequence
+// by AddInstruction() should comprise a valid SPIR-V module.
+class IrLoader {
+ public:
+  // Instantiates a builder to construct the given |module| gradually.
+  IrLoader(Module* module) : module_(module) {}
+
+  // Sets the fields in the module's header to the given parameters.
+  void SetModuleHeader(uint32_t magic, uint32_t version, uint32_t generator,
+                       uint32_t bound, uint32_t reserved) {
+    module_->SetHeader({magic, version, generator, bound, reserved});
+  }
+  // Adds an instruction to the module. This method will properly capture and
+  // store the data provided in |inst| so that |inst| is no longer needed after
+  // returning.
+  void AddInstruction(const spv_parsed_instruction_t* inst);
+  // Finalizes the module construction. This must be called after the module
+  // header has been set and all instructions have been added.
+  // Resolves internal bookkeeping.
+  void EndModule();
+
+ private:
+  // The module to be built.
+  Module* module_;
+  // The current Function under construction.
+  std::unique_ptr<Function> function_;
+  // The current BasicBlock under construction.
+  std::unique_ptr<BasicBlock> block_;
+  // Line related debug instructions accumulated thus far.
+  std::vector<Instruction> dbg_line_info_;
+};
+
+}  // namespace ir
+}  // namespace opt
+}  // namespace spvtools
+
+#endif  // LIBSPIRV_OPT_IR_LOADER_H_
--- a/test/opt/opt_test_common.h
+++ b/test/opt/opt_test_common.h
@@ -1,0 +1,54 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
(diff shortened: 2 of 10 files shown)


test/opt/opt_test_common.h
```diff
@@ -0,0 +1,128 @@
...
+
+  return result;
+}
+
+// Skips all leading comment lines and returns the pointer to the start of
+// instructions. This function relies on the fact that a comment line strictly
+// has the prefix of ';' and postfix of '\n'.
+const char* SkipComments(const char* assembly) {
+  while (*assembly == ';') {
+    while (*assembly != '\n') ++assembly;
```
dneto0: A new line can appear in a string. So this isn't right.
You have to also not be in a quoted string.  And to detect that you also need to skip over escaped-quotes.

Need unit tests for this.
But I see you use this only to scan a disassembled output.  How about adding an option to disassembly to prevent it from generating any comments at all.  Then you don't need this functionality.

source/opt/ir_builder.h
```diff
@@ -0,0 +1,85 @@
...
+namespace ir {
+
+// Builder class for constructing SPIR-V in-memory IR representation. Methods in
+// this class are designed to work with the interface for spvBinaryParse() in
+// libspirv.h so that we can leverage the syntax checks implemented behind it.
+//
+// The user is expected to call SetModuleHeader() to fill in the module's
+// header, and then AddInstruction() for each decoded instruction, and finally
+// Link() to linking everything together.
+class IrBuilder {
```
dneto0: If this is all this class is going to do, then it's better to call it IRLoader.
Normally, an IR Builder (like in LLVM or in Glslang) is much more helpful in constructing a module.  E.g. being able to set the build point at a random spot in the module, keeping type tables, etc.

test/opt/opt_test_common.h
```diff
@@ -0,0 +1,128 @@
...
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#ifndef LIBSPIRV_TEST_OPT_OPT_TEST_COMMON_H_
+#define LIBSPIRV_TEST_OPT_OPT_TEST_COMMON_H_
+
+#include <memory>
```
dneto0: Should have <vector> as well.

test/opt/opt_test_common.h
```diff
@@ -0,0 +1,128 @@
...
+  std::string result(SkipComments(text->str), text->str + text->length);
+
+  spvDiagnosticDestroy(diagnostic);
+  spvTextDestroy(text);
+  spvContextDestroy(context);
+
+  return result;
+}
+
+spv_result_t SetSpvHeader(void* builder, spv_endianness_t, uint32_t magic,
```
dneto0: I'd prefer to not expose these two functions.  Best to hide them and the body of BuildModule inside a .cpp file.
Also, they need blurbs.

source/opt/ir_builder.h
```diff
@@ -0,0 +1,85 @@
...
+namespace opt {
+namespace ir {
+
+// Builder class for constructing SPIR-V in-memory IR representation. Methods in
+// this class are designed to work with the interface for spvBinaryParse() in
+// libspirv.h so that we can leverage the syntax checks implemented behind it.
+//
+// The user is expected to call SetModuleHeader() to fill in the module's
+// header, and then AddInstruction() for each decoded instruction, and finally
+// Link() to linking everything together.
```
dneto0: The name Link is suggestive of linking multiple modules together.
Better to call it EndModule() or Finish() or something like that.

source/opt/ir_builder.h
```diff
@@ -0,0 +1,85 @@
...
+  // Sets the fields in the module's header to the given parameters.
+  void SetModuleHeader(uint32_t magic, uint32_t version, uint32_t generator,
+                       uint32_t bound, uint32_t reserved) {
+    module_->SetHeader({magic, version, generator, bound, reserved});
+  }
+  // Adds an instruction to the module. This method will properly capture and
+  // store the data provided in |inst| so that |inst| is no longer needed after
+  // returning.
+  void AddInstruction(const spv_parsed_instruction_t* inst);
+  // Resolves internal references among the module, functions, basic blocks,
```
dneto0: This blurb is confusing since the user of the API doesn't know the internal details.
Could just say:

```
// Ends module creation.  This must be called after the module has been set and all instructions have been added.
// Resolves internal bookkeeping.
void EndModule();
```

source/opt/ir_builder.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+void IrBuilder::AddInstruction(const spv_parsed_instruction_t* inst) {
+  const auto opcode = static_cast<SpvOp>(inst->opcode);
+  if (IsDebugLineInst(opcode)) {
+    dbg_line_info_.push_back(Instruction(*inst));
+    return;
+  }
+
+  Instruction spv_inst(*inst, std::move(dbg_line_info_));
+  dbg_line_info_.clear();
```
dneto0: Seems like the Instruction constructor should do this clear after it has moved the contents of db_line_info.

source/opt/ir_builder.h
```diff
@@ -0,0 +1,85 @@
...
+#include "spirv-tools/libspirv.h"
+
+namespace spvtools {
+namespace opt {
+namespace ir {
+
+// Builder class for constructing SPIR-V in-memory IR representation. Methods in
+// this class are designed to work with the interface for spvBinaryParse() in
+// libspirv.h so that we can leverage the syntax checks implemented behind it.
+//
```
dneto0: Should say that the class assumes that the calls to set-header and add-instruction should represent a valid SPIR-V module.  (Internal bookkeeping relies on it.)

source/opt/ir_builder.cpp
```diff
@@ -0,0 +1,117 @@
...
+    block_ = nullptr;
+  } else {
+    if (function_ == nullptr) {  // Outside function definition
+      assert(block_ == nullptr);
+      if (opcode == SpvOpCapability) {
+        module_->AddCapability(std::move(spv_inst));
+      } else if (opcode == SpvOpExtension) {
+        module_->AddExtension(std::move(spv_inst));
+      } else if (opcode == SpvOpExtInstImport) {
+        module_->AddExtInstSet(std::move(spv_inst));
```
dneto0: Hm. Minor thing.  The method name (and the ext_inst_sets_ member of Module) is inconsistent with the instruction name OpExtInstImport.   I think the inconsistency is suboptimal.  :-)

source/opt/ir_builder.cpp
```diff
@@ -0,0 +1,117 @@
...
+      } else if (opcode == SpvOpMemoryModel) {
+        module_->SetMemoryModel(std::move(spv_inst));
+      } else if (opcode == SpvOpEntryPoint) {
+        module_->AddEntryPoint(std::move(spv_inst));
+      } else if (opcode == SpvOpExecutionMode) {
+        module_->AddExecutionMode(std::move(spv_inst));
+      } else if (IsDebugInst(opcode)) {
+        module_->AddDebugInst(std::move(spv_inst));
+      } else if (IsAnnotationInst(opcode)) {
```
dneto0: Minor.  Similarly, all these instructions are decorations.  Better to call this  IsDecorationInst, etc.

source/opt/ir_builder.cpp
```diff
@@ -0,0 +1,117 @@
...
+        module_->AddAnnotationInst(std::move(spv_inst));
+      } else if (IsTypeInst(opcode)) {
+        module_->AddType(std::move(spv_inst));
+      } else if (IsConstantInst(opcode)) {
+        module_->AddConstant(std::move(spv_inst));
+      } else if (opcode == SpvOpVariable) {
+        module_->AddVariable(std::move(spv_inst));
```
dneto0: Caution:   In the SPIR-V module layout rules, module-scope variables are grouped with constants and types.  

Also, OpVariable with Function storage class are must go inside the function body, and only there.

Perhaps you need to update the description of Module::variables_ member to say what it really means.  Does the vector get cleared once the first function starts being defined?

source/opt/ir_builder.cpp
```diff
@@ -0,0 +1,117 @@
...
+        assert(opcode == SpvOpFunctionParameter);
+        function_->AddParameter(std::move(spv_inst));
+      } else {
+        block_->AddInstruction(std::move(spv_inst));
+      }
+    }
+  }
+}
+
+void IrBuilder::Link() {
```
dneto0: Seems to me this work can be done right when the objects are created.  In that case we don't need this method at all.

dneto0: Oh.  Wait. It's because of the std::moves....
Hm. Then yes, you do need this method.

dneto0:
-1

test/opt/opt_test_common.h
antiagainst: This function was only intended to be used in the test to strip the preamble header comments. Now we have #247 in, this function is removed.

source/opt/ir_builder.h
antiagainst: Good naming! ;)

antiagainst: Done.

test/opt/opt_test_common.h
antiagainst: Added more header files.

test/opt/opt_test_common.h
antiagainst: Put the impl. into a `.cpp` file. So now it should be clear whether a function is exported or not.

source/opt/ir_builder.h
antiagainst: Good naming! I Like it. :D

source/opt/ir_builder.h
antiagainst: Done.

source/opt/ir_builder.cpp
antiagainst: The constructor takes a rvalue, so it cannot do the clean up.

Actually I'd guess STL will automatically do the clear for us when the vector is moved away. But just put a `clear()` here to make sure.

source/opt/ir_builder.h
antiagainst: Done.

source/opt/ir_builder.cpp
antiagainst: Done.

source/opt/ir_builder.cpp
antiagainst: The spec calls this section "Annotation" and later may have more instructions that are not for decoration.

source/opt/ir_builder.cpp
antiagainst: Good point. Revised. OpVariables for local variables are put in the basic blocks.

source/opt/ir_builder.cpp
antiagainst: Yes.

source/opt/ir_loader.h
```diff
@@ -0,0 +1,83 @@
...
+  // Sets the fields in the module's header to the given parameters.
+  void SetModuleHeader(uint32_t magic, uint32_t version, uint32_t generator,
+                       uint32_t bound, uint32_t reserved) {
+    module_->SetHeader({magic, version, generator, bound, reserved});
+  }
+  // Adds an instruction to the module. This method will properly capture and
+  // store the data provided in |inst| so that |inst| is no longer needed after
+  // returning.
+  void AddInstruction(const spv_parsed_instruction_t* inst);
+  // Finalizes the module construction. This must be called after the module has
```
dneto0: "after the module *header* has been set"?

dneto0:
+2 provided you fix the blurb on EndModule.  :-)

source/opt/ir_loader.h
antiagainst: Sorry. Copy-pasta cut one word out. :(
Done.

PR #249, opened 2016-06-28
Promote ir namespace and create draft libspirv.{h|c}pp.

antiagainst:


source/opt/libspirv.hpp
```diff
@@ -24,31 +24,35 @@
...
-#include "source/opt/module.h"
+#include "module.h"
+#include "spirv-tools/libspirv.h"
 
 namespace spvtools {
-namespace opt {
 
-// TODO(antiagainst): expand and export these functions as the C++ interface in
-// libspirv.hpp.
+// TODO(antiagainst): add parameters for target environment and error message.
```
dneto0: Let's add the context object before making this API public.
I think it should be spvtools::ir::Context  and then the Assemble, Disassemble, and BuildModule functions are methods on the context object.

dneto0:
Please add the context object now.

source/opt/libspirv.hpp
antiagainst: Wrapped all functions in a class.

But I feel like `Context` is too vague as a name. I'd prefer `SpvTools`. It's much clear when reading together with, like `SpvTools.Assemble`.

dneto0:
Please add a test that shows failure of the assembler on some source program that uses a 1.1 feature.  For example, OpSizeOf is new in 1.1

antiagainst:
@dneto0: done.

dneto0:
+2
Please squash at least the first two commits together.

PR #250, opened 2016-06-28
Add Pass, PassManager, and StripDebugInfoPass.

antiagainst:
A part of #219.

### DIFF
--- a/source/opt/pass_manager.h
+++ b/source/opt/pass_manager.h
@@ -1,0 +1,84 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#ifndef LIBSPIRV_OPT_PASS_MANAGER_H_
+#define LIBSPIRV_OPT_PASS_MANAGER_H_
+
+#include <cassert>
+#include <memory>
+#include <vector>
+
+#include "module.h"
+#include "passes.h"
+
+namespace spvtools {
+namespace opt {
+
+// The pass manager, responsible for tracking and running passes.
+// Clients should first call AddPass() to add passes and then call Run()
+// to run on a module. Passes are executed in the exact order of added.
+//
+// TODO(antiagainst): The pass manager is fairly simple right now. Eventually it
+// should support pass dependency, common functionality (like def-use analysis)
+// sharing, etc.
+class PassManager {
+ public:
+  // Adds a pass.
+  void AddPass(std::unique_ptr<Pass> pass) {
+    passes_.push_back(std::move(pass));
+  }
+  template <typename PassT>
+  void AddPass() {
+    passes_.emplace_back(new PassT);
+  }
+
+  // Returns the number of passes added.
+  uint32_t NumPasses() const { return static_cast<uint32_t>(passes_.size()); }
+  // Returns a pointer to the |index|th pass added.
+  Pass* GetPass(uint32_t index) const {
+    assert(index < passes_.size() && "index out of bound");
+    return passes_[index].get();
+  }
+
+  // Runs all passes on the given |module|.
+  void Run(ir::Module* module) {
+    for (const auto& pass : passes_) {
+      // TODO(antiagainst): Currently we ignore the return value of the pass,
+      // which indicates whether the module has been modified, since there is
+      // nothing shared between passes right now.
+      pass->Process(module);
+    }
+  }
+
+ private:
+  // A vector of passes. Order matters.
+  std::vector<std::unique_ptr<Pass>> passes_;
+};
+
+}  // namespace opt
+}  // namespace spvtools
+
+#endif  // LIBSPIRV_OPT_PASS_MANAGER_H_
--- a/source/opt/passes.h
+++ b/source/opt/passes.h
@@ -1,0 +1,64 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#ifndef LIBSPIRV_OPT_PASSES_H_
+#define LIBSPIRV_OPT_PASSES_H_
+
+#include <memory>
+
+#include "module.h"
+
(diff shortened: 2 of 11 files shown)


source/opt/pass_manager.h
```diff
@@ -0,0 +1,73 @@
...
+
+#include "module.h"
+#include "passes.h"
+
+namespace spvtools {
+namespace opt {
+
+// The pass manager, responsible for tracking and running passes.
+// The user is expected to first call AddPass() to add passes and then call
+// Run() to run on a module. Passes are executed in the exact order of added.
```
dneto0: "The user are expected".   
Either "Users are expected" or "The user is expected".
Or alternately "Clients should first call ..."

source/opt/pass_manager.h
```diff
@@ -0,0 +1,73 @@
...
+//
+// TODO(antiagainst): This manager is fairly simple right now. Support pass
```
dneto0: Do you mean the pass manager should eventually support dependency ,etc...?

source/opt/pass_manager.h
```diff
@@ -0,0 +1,73 @@
...
+  }
+  template <typename PassT>
+  void AddPass() {
+    passes_.emplace_back(new PassT);
+  }
+
+  // Returns the number of passes added.
+  uint32_t NumPasses() const { return static_cast<uint32_t>(passes_.size()); }
+
+  // Runs all passes on the given |module| and returns true if all passes
```
dneto0: What does it mean for a pass to fail?  Is that useful information?
In LLVM a pass returns true if it changed the module.  That's useful for its pass manager so it can know whether analyses are invalidated.
This difference in interface might confuse people working with both projects.

source/opt/pass_manager.h
```diff
@@ -0,0 +1,73 @@
...
+  // succeed.
+  bool Run(ir::Module* module) {
+    for (const auto& pass : passes_) {
+      if (!pass->Process(module)) return false;
```
dneto0: This does more than what the blurb says.  It will stop running passes once one of them fails.

source/opt/pass_manager.h
antiagainst: Done.

source/opt/passes.h
```diff
@@ -0,0 +1,58 @@
...
+  // Processes the given |module| and returns true on success. The given
+  // |module| may be modified for optimization.
+  virtual bool Process(ir::Module* module) = 0;
+};
+
+// The optimization pass for removing debug instructions (as documented in
+// Section 3.32.2 of the SPIR-V spec).
+class StripDebugInfoPass : public Pass {
+ public:
+  const char* name() const override { return "RemoveDebugInfo"; }
```
dneto0: Strange that the name uses "Remove" and the class name uses "Strip".  I prefer "Strip" because it's more traditional, reminding me of the strip program.

test/opt/test_passes.cpp
```diff
@@ -0,0 +1,195 @@
...
+  const std::vector<std::string> dbg_opcodes{
+      // clang-format off
+    "OpSourceContinued", "OpSource", "OpSourceExtension",
+    "OpName", "OpMemberName", "OpString",
+    "OpLine", "OpNoLine", "OpModuleProcessed"
+      // clang-format on
+  };
+
+  std::ostringstream oss;
+  for (const auto& inst : insts) {
```
dneto0: In theory you could refactor SKipDbgInstJoin and Join in terms of a common method like "JoinWithout" which takes the list of input strings, the set of strings to avoid, and a delimiter.  But that's probably too low a level to make much of a maintenance difference. 
Don't bother.

test/opt/test_passes.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  // Adds a pass to be run.
+  template <typename PassT>
+  void AddPass() {
+    manager_.AddPass<PassT>();
+  }
+
+  // Runs the pass on the binary assembled from the |original| assembly, and
+  // checks whether the optimized binary can be disassembled to the |expected|
+  // assembly.
```
dneto0: I suggest that the blurb should suggest use of SCOPED_TRACE so that when a failure occurs inside this method, then the developer can find out which caller was responsible.

test/opt/test_passes.cpp
```diff
@@ -0,0 +1,195 @@
...
+  void RunAndCheck(const std::string& original, const std::string& expected) {
+    assert(manager_.NumPasses());
```
dneto0: Why not ASSERT_TRUE ?

test/opt/test_passes.cpp
```diff
@@ -0,0 +1,195 @@
...
+      "OpModuleProcessed \"43\"",
+      "OpModuleProcessed \"44\"",
+  };
+  text.insert(text.begin() + 4, more_text.cbegin(), more_text.cend());
+  RunAndCheck(Join(text), SkipDbgInstJoin(text));
+}
+
+using StripDebugInfoTest = PassTest<::testing::TestWithParam<std::string>>;
+
+TEST_P(StripDebugInfoTest, Kind) {
```
dneto0: This test doesn't access GetParam().  So the parameterized test is essentially a no-op.  :-)

test/opt/test_passes.cpp
```diff
@@ -0,0 +1,195 @@
...
+      "OpLine %3 3 3",
+      "%6 = OpLabel",
+      "OpLine %3 4 4",
+      "OpNoLine",
+      "OpReturn",
+      "OpLine %3 4 4",
+      "OpNoLine",
+      "OpFunctionEnd",
+  };
+  RunAndCheck(Join(text), SkipDbgInstJoin(text));
```
dneto0: But how can you trust SkipDbgInstJoin and Join?  
You need tests for those basic things too, I think.

test/opt/test_passes.cpp
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// Template class for testing passes. It contains some handy utility methods for
+// running passes and checking results.
+//
+// To write value-Parameterized tests:
+//   using ValueParamTest = PassTest<::testing::TestWithParam<std::string>>;
+// To use as normal fixture:
+//   using FixtureTest = PassTest<::testing::Test>;
+template <typename GT>
```
dneto0: I don't know why the template parameter is "GT".  I can't figure out the G part.

source/opt/pass_manager.h
```diff
@@ -0,0 +1,73 @@
...
+// The pass manager, responsible for tracking and running passes.
+// The user is expected to first call AddPass() to add passes and then call
+// Run() to run on a module. Passes are executed in the exact order of added.
+//
+// TODO(antiagainst): This manager is fairly simple right now. Support pass
+// dependency, common functionality (like def-use analysis) sharing, etc.
+class PassManager {
+ public:
+  // Adds a pass.
+  void AddPass(std::unique_ptr<Pass> pass) {
```
dneto0: You haven't tested this method.

source/opt/pass_manager.h
```diff
@@ -0,0 +1,73 @@
...
+    passes_.push_back(std::move(pass));
+  }
+  template <typename PassT>
+  void AddPass() {
+    passes_.emplace_back(new PassT);
+  }
+
+  // Returns the number of passes added.
+  uint32_t NumPasses() const { return static_cast<uint32_t>(passes_.size()); }
```
dneto0: You haven't tested this method.  Should have 0, 1, 2 cases.

source/opt/pass_manager.h
```diff
@@ -0,0 +1,73 @@
...
+
+  // Runs all passes on the given |module| and returns true if all passes
+  // succeed.
+  bool Run(ir::Module* module) {
```
dneto0: You haven't tested:
-  the false return case
- the true return case where there is more than one pass.

dneto0:
-1

source/opt/pass_manager.h
antiagainst: Right. Modified the comment to make it more clear.

source/opt/pass_manager.h
antiagainst: Good call. Actually using the return value as indication of successful or not is kinda not well thought. Changed to adopt the LLVM way.

source/opt/pass_manager.h
antiagainst: Changed the signature & modified the blurb.

source/opt/passes.h
antiagainst: Good catch. Actually I was renaming all of them, but still missed one. :(

test/opt/test_passes.cpp
antiagainst: Done.

test/opt/test_passes.cpp
antiagainst: Because failure here means the test writer fails to add a pass. It's not that the test itself has failed.

test/opt/test_passes.cpp
antiagainst: Actually it has, in the `text` vector. :)

test/opt/test_passes.cpp
antiagainst: Tested.

test/opt/test_passes.cpp
antiagainst: Was to mean **G**oogle**T**est. But, yeah, maybe only I can understand it. Renamed.

source/opt/pass_manager.h
antiagainst: Done.

source/opt/pass_manager.h
antiagainst: Done.

test/opt/pass_utils.cpp
```diff
@@ -0,0 +1,77 @@
...
+                     [inst](const char* op) {
+                       return std::string(inst).find(op) != std::string::npos;
+                     });
+}
+
+}  // anonymous namespace
+
+namespace spvtools {
+
+std::string SelectiveJoin(const std::vector<const char*>& strings,
```
dneto0: Should have a blurb.

source/opt/pass_manager.h
antiagainst: Signature changed. N/A anymore.

test/opt/pass_utils.cpp
```diff
@@ -0,0 +1,77 @@
...
+                          const std::function<bool(const char*)>& skip_dictator,
+                          char delimiter) {
+  std::ostringstream oss;
+  for (const auto* str : strings) {
+    if (!skip_dictator(str)) oss << str << delimiter;
+  }
+  return oss.str();
+}
+
+std::string JoinAllInsts(const std::vector<const char*>& insts) {
```
dneto0: Blurb

test/opt/pass_utils.cpp
```diff
@@ -0,0 +1,77 @@
...
+  return SelectiveJoin(insts, [](const char*) { return false; });
+}
+
+std::string JoinNonDebugInsts(const std::vector<const char*>& insts) {
```
dneto0: Blurb

dneto0: Never mind.

test/opt/pass_utils.cpp
dneto0: Never mind.

test/opt/pass_fixture.h
```diff
@@ -0,0 +1,85 @@
...
+  template <typename PassT>
+  void AddPass() {
+    manager_.AddPass<PassT>();
+  }
+
+  // Runs the pass on the binary assembled from the |original| assembly, and
+  // checks whether the optimized binary can be disassembled to the |expected|
+  // assembly. Callers are suggested to use SCOPED_TRACE() for better messages.
+  void RunAndCheck(const std::string& original, const std::string& expected) {
+    assert(manager_.NumPasses());
```
dneto0: Why not ASSERT_TRUE?

test/opt/test_passes.cpp
dneto0: Oops. you're right.  :-)

test/opt/pass_fixture.h
dneto0: The regular assert might not trigger on release builds.  But ASSERT_TRUE would.
I am guessing here that you want to guard against accidentally making a test without passes. That's ok.

dneto0:
Needs a test to make sure that the Run() module actually runs all the passes in order.  All your tests that run passes only run a single pass.
Suggesting making a custom pass class that is constructed  a vector of words, and then just appends that to the end of the module.  And then add two instances of that class as passes (with different parameters).  The result doesn't have to validate.  It just has to assemble and disassemble.

-1

test/opt/pass_fixture.h
antiagainst: Yes. That's my intent.

antiagainst:
@dneto0: done.

dneto0:
+2 Please squash

PR #253, opened 2016-06-29
Test v1.1 loop dependency support

dneto0:
For DependencyInfinite and DependencyLength, test
that they don't require a capability to be turned on.
Also, that they are assembled, binary parsed, and disassembled
correctly.

### DIFF
--- a/test/BinaryToText.cpp
+++ b/test/BinaryToText.cpp
@@ -35,11 +35,14 @@
 
 namespace {
 
+using ::testing::Combine;
 using ::testing::Eq;
 using ::testing::HasSubstr;
 using spvtest::AutoText;
 using spvtest::ScopedContext;
 using spvtest::TextToBinaryTest;
+using std::get;
+using std::tuple;
 
 class BinaryToText : public ::testing::Test {
  public:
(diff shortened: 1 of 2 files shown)


test/BinaryToText.cpp
```diff
@@ -245,7 +245,10 @@ using RoundTripInstructionsTest =
     spvtest::TextToBinaryTestBase<::testing::TestWithParam<std::string>>;
 
 TEST_P(RoundTripInstructionsTest, Sample) {
-  EXPECT_THAT(EncodeAndDecodeSuccessfully(GetParam()), Eq(GetParam()));
+  EXPECT_THAT(
+      EncodeAndDecodeSuccessfully(GetParam(), SPV_BINARY_TO_TEXT_OPTION_NONE,
+                                  SPV_ENV_UNIVERSAL_1_1),
```
dekimir: This turns off all 1.0 testing, tho. :(

test/BinaryToText.cpp
```diff
@@ -285,6 +288,8 @@ INSTANTIATE_TEST_CASE_P(LoopControlMasks, RoundTripInstructionsTest,
                             "OpLoopMerge %1 %2 Unroll\n",
                             "OpLoopMerge %1 %2 DontUnroll\n",
                             "OpLoopMerge %1 %2 Unroll|DontUnroll\n",
+                            "OpLoopMerge %1 %2 DependencyInfinite\n",
+                            "OpLoopMerge %1 %2 DependencyLength 8\n",
```
dekimir: Perhaps this should go into `TextToBinary.ControlFlow.cpp`, together with other `OpLoopMerge` tests?

See 13877a3b0e9d9e3217f726178f6d26f1bf1867cc.

test/BinaryToText.cpp
dneto0: Fixed.
New commit parameterizes the BinaryToText round trip tests over both v1.0 and 1.1.
The new dependency operands to OpLoopMerge are only tested in 1.1.

test/BinaryToText.cpp
dneto0: Most of the existing tests test the assembler parts in TextToBinary.*  and the diassembler/round trip is in BinaryToText.  I'd rather not change the pattern.  Yes, there are some tests in various TextToBinary.* that are round-trip, but that's the exception, not the rule.

dneto0:
Addressed all feedback.

dekimir:
+2

dneto0:
Rebased, then pushed into master.

PR #254, opened 2016-06-29
Add the spirv-opt command line tool.

antiagainst:
a part of #219.

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -25,7 +25,7 @@
 # MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
 if (NOT ${SPIRV_SKIP_EXECUTABLES})
-  set(SPIRV_INSTALL_TARGETS spirv-as spirv-dis spirv-val)
+  set(SPIRV_INSTALL_TARGETS spirv-as spirv-dis spirv-val spirv-opt)
 
   add_executable(spirv-as ${CMAKE_CURRENT_SOURCE_DIR}/as/as.cpp)
   spvtools_default_compile_options(spirv-as)
@@ -43,6 +43,12 @@ if (NOT ${SPIRV_SKIP_EXECUTABLES})
   target_link_libraries(spirv-val PRIVATE ${SPIRV_TOOLS})
   target_include_directories(spirv-val
     PRIVATE ${spirv-tools_BINARY_DIR} ${spirv-tools_SOURCE_DIR}/source)
+
+  add_executable(spirv-opt ${CMAKE_CURRENT_SOURCE_DIR}/opt/opt.cpp)
+  spvtools_default_compile_options(spirv-opt)
+  target_link_libraries(spirv-opt PRIVATE SPIRV-Tools-opt ${SPIRV_TOOLS})
+  target_include_directories(spirv-opt PRIVATE
+    ${spirv-tools_SOURCE_DIR}/source ${spirv-tools_BINARY_DIR})
 
   install(TARGETS ${SPIRV_INSTALL_TARGETS}
     RUNTIME DESTINATION bin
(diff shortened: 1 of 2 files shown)


tools/CMakeLists.txt
```diff
@@ -44,6 +44,12 @@ if (NOT ${SPIRV_SKIP_EXECUTABLES})
   target_include_directories(spirv-val
     PRIVATE ${spirv-tools_BINARY_DIR} ${spirv-tools_SOURCE_DIR}/source)
 
+  add_executable(spirv-opt ${CMAKE_CURRENT_SOURCE_DIR}/opt/opt.cpp)
+  spvtools_default_compile_options(spirv-opt)
+  target_link_libraries(spirv-opt PRIVATE SPIRV-Tools-opt ${SPIRV_TOOLS})
+  target_include_directories(spirv-opt PRIVATE
+    ${spirv-tools_SOURCE_DIR}/source ${spirv-tools_BINARY_DIR})
+
   install(TARGETS ${SPIRV_INSTALL_TARGETS}
```
AWoloszyn: We should probably add spirv-opt to the list of SPIRV_INSTALL_TARGETS

tools/opt/opt.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+The SPIR-V binary is read from <input>. If no file is specified,
+or if <input> is "-", then the binary is read from standard input.
+if <output> is "-", then the optimized output is written to
+standard output.
+
+NOTE: The optimizer is a work in progress.
+
+Options:
+  --strip-debug-info
```
AWoloszyn: If we wanted to have similar syntax to llvm-opt this would e --strip-debug.

tools/opt/opt.cpp
```diff
@@ -0,0 +1,160 @@
...
+      } else if (0 == strcmp(cur_arg, "-o")) {
+        if (!out_file && argi + 1 < argc) {
+          out_file = argv[++argi];
+        } else {
+          PrintUsage(argv[0]);
+          return 1;
+        }
+      } else if (0 == strcmp(cur_arg, "--strip-debug-info")) {
+        pass_manager.AddPass<opt::StripDebugInfoPass>();
+      } else if (0 == cur_arg[1]) {
```
AWoloszyn: Nit: if you are comparing against a char you should probably use the '\0' instead of 0.

tools/opt/opt.cpp
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  std::vector<uint32_t> source;
+  const bool use_file = in_file && strcmp("-", in_file);
+  if (FILE* fp = (use_file ? fopen(in_file, "rb") : stdin)) {
+    uint32_t buf[1024];
+    while (size_t len = fread(buf, sizeof(uint32_t),
+                              sizeof(buf) / sizeof(uint32_t), fp)) {
+      source.insert(source.end(), buf, buf + len);
+    }
```
AWoloszyn: Since we are reading 4-bytes at a time, we might want to make sure that we have read the entire file by the end (i.e. there are no stray bytes after the last word)

tools/CMakeLists.txt
antiagainst: Good catch! Done.

tools/opt/opt.cpp
antiagainst: Right. Renamed.

tools/opt/opt.cpp
antiagainst: Done.

tools/opt/opt.cpp
antiagainst: By trying on my Linux machine, I found that `fp` is advanced even if `fread()` cannot read in a whole word. So here i use `ftell()` to do the check.

AWoloszyn:
LGTM

PR #256, opened 2016-07-04
Fix the name mismatch in CMakeLists.txt

Qining:
Change from spirv-headers to SPIV-Headers


antiagainst:
-1

We are using lowercase names for `spirv-headers` when `git clone`. So this won't work.

Qining:
All right,  but the default name of repo:
https://github.com/KhronosGroup/SPIRV-Headers
is SPIRV-Headers. Probably we need to add one line in the CMake error message to tell the user to checkout the repo with a specific name.

PR #257, opened 2016-07-04
DRY io functions in various command line tools.

antiagainst:


tools/io.h
```diff
@@ -0,0 +1,89 @@
...
+
+// Writes the given |data| into the file named as |filename| using the given
+// |mode|, assuming |data| is an array of |count| elements of type |T|. If
+// |filename| is nullptr or "-", writes to standard output. If any error occurs,
+// returns false and outputs error message to standard error.
+template <typename T>
+bool WriteFile(const char* filename, const char* mode, const T* data,
+               size_t count) {
+  const bool use_stdout =
+      !filename || (filename[0] == '-' && filename[1] == '\0');
```
dneto0: Could have used strcmp here, like we did in ReadFile.
But not serious enough to merit a change.

dneto0:
+2  
This is just a refactor.  Did you consider writing tests for ReadFile and WriteFile?

antiagainst:
Was thinking for a short while and then gave up. It would be nice to have, but, testing file system related stuff on multiple platforms is a pain. I'm not sure the gain is big enough compared to the current state.

PR #258, opened 2016-07-04
Finalize v2016.0

dneto0:


README.md
```diff
@@ -198,6 +198,16 @@ Use option `-h` to print help.
 The output includes syntax colouring when printing to the standard output stream,
 on Linux, Windows, and OS X.
 
+### Optimizer tool
+
+The optimizer processes a SPIR-V binary module, applying transformations
+in specified order.
```
antiagainst: in the specified order.

CHANGES
```diff
@@ -16,6 +16,8 @@ v2016.0-dev 2016-04-22
      - Supports floating point widths of 16, 32, 64 bits
      - Supports integer widths up to 64 bits
    - Validator is incomplete
+     - Validates some control flow graph rules
```
antiagainst: Also ID checks?

README.md
```diff
@@ -198,6 +198,16 @@ Use option `-h` to print help.
...
+
+This is a work in progress, with initially only few available transformations.
```
antiagainst: Prefix with `Warning` like the validator?

dneto0: I left off a Warning because I think it's a different level of expectation.

dneto0:
Addressed feedback

antiagainst:
LGTM.

Let's submit this today so the time stamp is actually correct. ;)

Issue #261, opened 2016-07-07
disable capability checks for ClipDistance and CullDistance, for now


dneto0:
There is open debate about when a ClipDistance or CullDistance capability is required.
For example, see https://github.com/KhronosGroup/glslang/issues/334
Until the dust settles, it might be prudent to disable those capability checks.

PR #262, opened 2016-07-07
Turn off ClipDistance CullDistance cap checks for Vulkan

dneto0:
Turn them off until resolution of the debate over how they should be checked.
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/261


AWoloszyn:
+2

dneto0:
Rebased and pushed into master as 06581f5ce6cb

PR #264, opened 2016-07-13
Refactor Id definition and use tracking

umar456:
* Creates an ID class which manages definition and use of IDs
* Moved tracking code from validate.cpp to validate_id.cpp
* Rename and combine SsaPass and ProcessIds into IdPass
* Remove module dependency in Function


source/val/Id.h
```diff
@@ -0,0 +1,134 @@
...
+class BasicBlock;
+
+enum IdType {
+  kIdTypeUnknown,
+  kIdTypeLabel,
+  kIdTypeType,
+  kIdTypeObject,
+};
+
+class Id {
```
dneto0: Need a blurb, I think.
The constructors distinguish between a "complete" and "incomplete" Id.  The class blurb should describe that concept.

source/val/Id.h
```diff
@@ -0,0 +1,134 @@
...
+  /// This constructor creates an incomplete Id. This constructor can be used to
+  /// create Id that are used to find other Ids
+  explicit Id(const uint32_t result_id = 0);
+
+  /// This constructor creates a complete Id.
+  explicit Id(const spv_parsed_instruction_t* inst,
+              Function* function = nullptr, BasicBlock* block = nullptr);
+
+  /// Registers a use of the Id
+  void RegisterUse(const Function* function = nullptr);
```
dneto0: Does this mean that the ID is used in the given function?

source/val/Id.h
```diff
@@ -0,0 +1,134 @@
...
+  /// The integer that identifies the Id
+  uint32_t id_;
+
+  /// The type of the Id
+  uint32_t type_id_;
+
+  /// The opcode used to define the Id
+  SpvOp opcode_;
+
+  /// The function in which the Id was defined
```
dneto0: Or, presumably, nullptr if the Id is defined outside a function.

source/val/Id.h
```diff
@@ -0,0 +1,134 @@
...
+  /// The block in which the Id was defined
+  BasicBlock* defining_block_;
+
+  /// The blocks in which the Id was used
+  std::set<const BasicBlock*> uses_;
+
+  /// The words of the instuction that defined the Id
+  std::vector<uint32_t> words_;
+
+#define OPERATOR(OP)                                     \
```
dneto0: The use of a macro here adds little.  Meh.  Ok to keep.

source/val/Id.h
```diff
@@ -0,0 +1,134 @@
...
+  SpvOp opcode_;
+
+  /// The function in which the Id was defined
+  Function* defining_function_;
+
+  /// The block in which the Id was defined
+  BasicBlock* defining_block_;
+
+  /// The blocks in which the Id was used
+  std::set<const BasicBlock*> uses_;
```
dneto0: Without reading the implementation, I don't see how this member can be populated since the RegisterUse only takes a function as argument.

source/val/Id.h
```diff
@@ -0,0 +1,134 @@
...
+OPERATOR(<);
+OPERATOR(==);
+#undef OPERATOR
+
+}  // namespace libspirv
+
+// custom specialization of std::hash for Id
+namespace std {
+template <>
+struct hash<libspirv::Id> {
```
dneto0: Ordinarily we don't specialize std::hash
https://google.github.io/styleguide/cppguide.html#std_hash 

However, this is just a pass-through down to std::hash<uint32_t>, so ok for now.
The danger is if the Id class becomes more complex, but the risk of that is low.

source/val/Id.h
dneto0: Ah,  I see now.  The blurb needs to be updated to say this method registers a use of this Id in the "current block" of the given function.
Since we only care to track the basic blocks, I'd just make the argument a BasicBlock directly.  It's simpler: reduces coupling with the Function's API and in principle would make it easier to test.

source/val/ValidationState.cpp
```diff
@@ -358,4 +370,26 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
...
+  }
+}
+
+void ValidationState_t::RegisterUseId(uint32_t used_id) {
+  auto used = all_definitions_.find(used_id);
+  if (used != end(all_definitions_)) {
+    if (in_function_body())
+      used->second.RegisterUse(&current_function());
+    else
+      used->second.RegisterUse(nullptr);
```
dneto0: This else clause ultimately does nothing.  I'd just skip it.

source/val/ValidationState.cpp
```diff
@@ -358,4 +370,26 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
...
+  } else {
+    all_definitions_[inst.result_id] = Id{&inst};
+  }
+}
+
+void ValidationState_t::RegisterUseId(uint32_t used_id) {
+  auto used = all_definitions_.find(used_id);
+  if (used != end(all_definitions_)) {
+    if (in_function_body())
+      used->second.RegisterUse(&current_function());
```
dneto0: Recommend changing RegisterUse to take the using BasicBlock as its argument.  That changes this invocation, of course, to be guarded by in_function_body() && in_block()

dneto0:
Finished review.  Mostly minor stuff.

dneto0:
Ok.  +2 essential feedback addressed.  I'll squash and merge.

dneto0:
Squashed and pushed into master as 66c94928a32

PR #265, opened 2016-07-14
Check ID definition block is a dominator of use block

umar456:
The validator now checks that the ID definition appears in a block that dominates the block where they are used.  This is using a simpler approach than PR #237.

Relies on #264

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,8 @@ Revision history for SPIRV-Tools
 
 v2016.2-dev 2016-07-19
  - Start v2016.2
+ - Validator is incomplete
+   - Checks ID use block is dominated by definition block
 
 v2016.1 2016-07-19
  - Fix https://github.com/KhronosGroup/SPIRV-Tools/issues/261
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -2401,8 +2401,9 @@ spv_result_t CheckIdDefinitionDominateUse(const ValidationState_t& _) {
           }
         }
       } else {
-        // If the Ids appear within a function but not in a block, then make
-        // sure all references to that Id appear within the same function
+        // If the Ids defined within a function but not in a block(i.e. function
+        // parameters, block ids), then make sure all references to that Id
+        // appear within the same function
         bool found = false;
         for (auto use : definition.second.uses()) {
           tie(ignore, found) = func->GetBlock(use->id());
@@ -2439,8 +2440,9 @@ spv_result_t IdPass(ValidationState_t& _,
     switch (type) {
       case SPV_OPERAND_TYPE_RESULT_ID:
         // NOTE: Multiple Id definitions are being checked by the binary parser
+        // NOTE: result Id is added *after* all of the other Ids have been
+        // checked to avoid premature use in the same instruction
         _.RemoveIfForwardDeclared(*operand_ptr);
-        _.AddId(*inst);
         ret = SPV_SUCCESS;
         break;
       case SPV_OPERAND_TYPE_ID:
@@ -2465,6 +2467,9 @@ spv_result_t IdPass(ValidationState_t& _,
     if (SPV_SUCCESS != ret) {
       return ret;
     }
+  }
+  if (inst->result_id) {
+    _.AddId(*inst);
   }
   return SPV_SUCCESS;
 }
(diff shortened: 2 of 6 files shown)


CHANGES
```diff
@@ -4,6 +4,8 @@ v2016.1-dev 2016-07-04
  - Start v2016.1
  - Fix https://github.com/KhronosGroup/SPIRV-Tools/issues/261
    Turn off ClipDistance and CullDistance capability checks for Vulkan.
+ - Validator is incomplete
+   - Checks ID use block is dominated by definition block
```
dneto0: v2016.2-dev has begun.  Please move this up.  (And thanks for being pro-active about updating CHANGES!)

source/validate_id.cpp
```diff
@@ -2396,6 +2438,7 @@ spv_result_t IdPass(ValidationState_t& _,
     auto ret = SPV_ERROR_INTERNAL;
     switch (type) {
       case SPV_OPERAND_TYPE_RESULT_ID:
+        // NOTE: Multiple Id definitions are being checked by the binary parser
```
dneto0: Good comment.  !

test/Validate.CFG.cpp
```diff
@@ -91,8 +91,12 @@ class Block {
       : label_(label), body_(), type_(type), successors_() {}
 
   /// Sets the instructions which will appear in the body of the block
-  Block& setBody(std::string body) {
-    body_ = body;
+  Block& setBody(std::string body, bool append = false) {
```
dneto0: I would have made a new method.  appendBody(string).
But meh.  No big deal in test code.

source/validate_id.cpp
```diff
@@ -2376,13 +2376,55 @@ function<bool(unsigned)> getCanBeForwardDeclaredFunction(SpvOp opcode) {
...
+        // If the Id is defined within a block then make sure all references to
+        // that Id appear in a blocks that are dominated by the defining block
+        for (auto use : definition.second.uses()) {
+          if (use->dom_end() == find(use->dom_begin(), use->dom_end(), block)) {
+            return _.diag(SPV_ERROR_INVALID_ID)
+                   << "ID " << _.getIdName(definition.first)
+                   << " defined in block " << _.getIdName(block->id())
+                   << " does not dominate its use in block "
+                   << _.getIdName(use->id());
+          }
```
dneto0: TODO:   If the use is in the same block as the definition, then the use must appear after the definition.

source/validate_id.cpp
```diff
@@ -2376,13 +2376,55 @@ function<bool(unsigned)> getCanBeForwardDeclaredFunction(SpvOp opcode) {
...
+        }
+      } else {
+        // If the Ids appear within a function but not in a block, then make
```
dneto0: "appear" -> "are defined"

dneto0: Might be useful to remind the reader that this occurs for function parameters.
(And I can't remember the detail of whether an OpLabel is considered as defined in its own block or not.)

test/Validate.SSA.cpp
```diff
@@ -1014,5 +1014,148 @@ TEST_F(ValidateSSA, PhiMissingLabelBad) {
   EXPECT_THAT(getDiagnosticString(), HasSubstr("missing"));
 }
 
-// TODO(umar): OpGroupMemberDecorate
+TEST_F(ValidateSSA, IdDominatesItsUseGood) {
```
dneto0: This is the case where uses are in different blocks from their definitions.
I'd also add another case where a use is in the same block as its definition.  (Or add that case to this one.)

test/Validate.SSA.cpp
```diff
@@ -1014,5 +1014,148 @@ TEST_F(ValidateSSA, PhiMissingLabelBad) {
...
+               OpFunctionEnd
+)";
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(
+      getDiagnosticString(),
+      MatchesRegex("ID .\\[eleven\\] defined in block .\\[true_block\\] does "
+                   "not dominate its use in block .\\[false_block\\]"));
+}
+
```
dneto0: Need a test case where the def and use are in the same block.  See above for possible cases.

dneto0:
First pass review complete.
Missing checks for def and use in the same basic block.

source/validate_id.cpp
umar456: Good catch!

I was already checking simple def and use cases in the IdPass but I was not taking into account use in the same instruction as the definition.

dneto0:
+2.  I'll squash and merge

dneto0:
Pushed into master as efc782d74

PR #263, opened 2016-07-08
Nice disasm

dneto0:
This is a long-requested feature.  It should make disassembly more readable.
It might enhance testabilty for SPIR-V generators since it will make it easier to write matchers against disassembled output.

PR #269, opened 2016-07-19
Disassembling: map IDs to friendly names.

antiagainst:
Add a `FriendlyNameMapper` to deduce friendly names for IDs based on
`OpName`, type structure, etc.

Taking over PR #263 @dneto0 did, fix issues here since I cannot push into
his branch.


antiagainst:
+2

PR #271, opened 2016-07-22
WIP: Add -load <pass-module> for spirv-opt

Qining:
Add PassRegistry to support out-of-source passes.

A `hello` pass:
```
#include "module.h"
// "pass.h" includes "pass_registry.h", which is supposed to be a public header
#include "pass.h"

#include <iostream>

namespace {

class HelloPass : public spvtools::opt::Pass {
 public:
  const char* name() const override { return "hello pass"; }
  bool Process(spvtools::ir::Module*) override {
    std::cerr << name() << std::endl;
    return false;
  }
};

// Register the pass to the registry
spvtools::opt::RegisterPass<HelloPass> X("hello");

}  // anonymous namespace
```

An example module with `hello` pass should be compiled with:
`-std=c++11 -fPIC -shared`

TODO:
Support Windows.
Tests for '-load <module>'.

### DIFF
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -6,8 +6,11 @@ add_library(SPIRV-Tools-opt
   libspirv.hpp
   module.h
   reflect.h
+  pass.h
   passes.h
   pass_manager.h
+  pass_registry_impl.h
+  pass_registry.h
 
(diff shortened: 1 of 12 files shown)


Qining:
close as this is not necessary for now.

Issue #270, opened 2016-07-21
Vulkan conformance test gets access violation in spvValidate


antiagainst:
Specifics
Windows 7, Nvidia GTX 660 Ti, Driver 368.39 

Have Vulkan SDK 1.0.21 installed for validation layers and set environment variable VK_INSTANCE_LAYERS=VK_LAYER_LUNARG_standard_validation

To reproduce I run all of the dEQP-VK.glsl.functions.control_flow.* tests from the Vulkan Conformance Test Suite

Message:
  Exception thrown: read access violation.
  this was nullptr. (debugger confirms that variable "this" is NULL)

Location:
  BasicBlock.h:

```
76 /// Returns true if the block is reachable in the CFG
77 bool reachable() const { return reachable_; }
```

Call Stack:
```
	VkLayer_core_validation.dll!libspirv::BasicBlock::reachable() Line 77	C++

 	VkLayer_core_validation.dll!libspirv::StructuredControlFlowChecks(const libspirv::ValidationState_t & _, const libspirv::Function & function, const std::vector<std::pair<unsigned int,unsigned int>,std::allocator<std::pair<unsigned int,unsigned int> > > & back_edges) Line 349	C++

 	VkLayer_core_validation.dll!libspirv::PerformCfgChecks(libspirv::ValidationState_t & _) Line 478	C++

 	VkLayer_core_validation.dll!spvValidate(const spv_context_t * const context, spv_const_binary_t * const binary, spv_diagnostic_t * * pDiagnostic) Line 209	C++

 	VkLayer_core_validation.dll!core_validation::CreateShaderModule(VkDevice_T * device, const VkShaderModuleCreateInfo * pCreateInfo, const VkAllocationCallbacks * pAllocator, VkShaderModule_T * * pShaderModule) Line 9069	C++

 	VkLayer_object_tracker.dll!object_tracker::CreateShaderModule(VkDevice_T * device, const VkShaderModuleCreateInfo * pCreateInfo, const VkAllocationCallbacks * pAllocator, VkShaderModule_T * * pShaderModule) Line 3583	C++

 	VkLayer_parameter_validation.dll!parameter_validation::CreateShaderModule(VkDevice_T * device, const VkShaderModuleCreateInfo * pCreateInfo, const VkAllocationCallbacks * pAllocator, VkShaderModule_T * * pShaderModule) Line 2621	C++

[...]

PR #272, opened 2016-07-22
Fixes segfault for loops without back-edges

umar456:
This fixes issue #270 which appeared because the loop construct did not have a back-edge branching to it. The the edge branching from the continue target is not a back-edge because it is unreachable in the CFG from the first block.

### DIFF
--- a/source/validate_cfg.cpp
+++ b/source/validate_cfg.cpp
@@ -32,6 +32,7 @@
 #include <functional>
 #include <set>
 #include <string>
+#include <tuple>
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
@@ -47,12 +48,14 @@ using std::function;
 using std::get;
 using std::ignore;
 using std::make_pair;
+using std::make_tuple;
 using std::numeric_limits;
 using std::pair;
 using std::set;
 using std::string;
 using std::tie;
 using std::transform;
+using std::tuple;
 using std::unordered_map;
 using std::unordered_set;
 using std::vector;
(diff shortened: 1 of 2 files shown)


antiagainst:
+1. Would prefer @dneto0 to double check this.

source/validate_cfg.cpp
```diff
@@ -268,21 +271,52 @@ void UpdateContinueConstructExitBlocks(
...
+      exit_name = "merge block";
+      break;
+    case ConstructType::kContinue:
+      construct_name = "continue";
+      header_name = "continue target";
+      exit_name = "back-edge block";
+      break;
+    case ConstructType::kCase:
+      construct_name = "case";
+      header_name = "case block";
```
dneto0: Optional:  How about "case entry block" and "case exit block". ?    If I feel strongly about this I can change it in a subsequent CL.

source/validate_cfg.cpp
```diff
@@ -268,21 +271,52 @@ void UpdateContinueConstructExitBlocks(
...
   if (post_dominate) {
     dominate_text = "is not post dominated by";
   } else {
     dominate_text = "does not dominate";
   }
 
+  tie(construct_name, header_name, exit_name) =
+      ConstructNames(construct.type());
+
   switch (construct.type()) {
```
dneto0: This should be dead code since you've replaced it with the call to ConstructNames

dneto0:
+2   I'll squash and merge

dneto0:
Merged into master as 6c61bf2dfad

PR #274, opened 2016-07-26
Freeze spec consts to their default value

Qining:
Add a pass to freeze spec constants their default value. This pass does
not fold the frozen spec constants so does not handle SpecConstantOp
instructions and SpecConstantComposite instructions.

Work in progress as this CL add OpINotEqual to the whitelist of
supported opcodes of OpSpecConstantOp.

### DIFF
--- a/source/assembly_grammar.cpp
+++ b/source/assembly_grammar.cpp
@@ -150,7 +150,6 @@ const SpecConstantOpcodeEntry kOpSpecConstantOpcodes[] = {
     CASE(Select),
     // Comparison
     CASE(IEqual),
-    CASE(INotEqual),
     CASE(ULessThan),
     CASE(SLessThan),
     CASE(UGreaterThan),
@@ -167,7 +166,7 @@ const SpecConstantOpcodeEntry kOpSpecConstantOpcodes[] = {
 };
 
 // The 58 is determined by counting the opcodes listed in the spec.
-static_assert(59 == sizeof(kOpSpecConstantOpcodes)/sizeof(kOpSpecConstantOpcodes[0]),
+static_assert(58 == sizeof(kOpSpecConstantOpcodes)/sizeof(kOpSpecConstantOpcodes[0]),
               "OpSpecConstantOp opcode table is incomplete");
 #undef CASE
 // clang-format on
--- a/source/assembly_grammar.cpp
+++ b/source/assembly_grammar.cpp
@@ -150,6 +150,7 @@ const SpecConstantOpcodeEntry kOpSpecConstantOpcodes[] = {
     CASE(Select),
     // Comparison
     CASE(IEqual),
+    CASE(INotEqual),
     CASE(ULessThan),
     CASE(SLessThan),
     CASE(UGreaterThan),
@@ -166,7 +167,7 @@ const SpecConstantOpcodeEntry kOpSpecConstantOpcodes[] = {
 };
 
 // The 58 is determined by counting the opcodes listed in the spec.
-static_assert(58 == sizeof(kOpSpecConstantOpcodes)/sizeof(kOpSpecConstantOpcodes[0]),
+static_assert(59 == sizeof(kOpSpecConstantOpcodes)/sizeof(kOpSpecConstantOpcodes[0]),
               "OpSpecConstantOp opcode table is incomplete");
 #undef CASE
 // clang-format on
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -86,6 +86,7 @@ class Instruction {
               std::vector<Instruction>&& dbg_line = {});
 
   SpvOp opcode() const { return opcode_; }
+  void SetOpcode(SpvOp opcode) { opcode_ = opcode; }
   uint32_t type_id() const { return type_id_; }
   uint32_t result_id() const { return result_id_; }
   // Returns the vector of line-related debug instructions attached to this
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -90,7 +90,7 @@ class Instruction {
   // make the instruction to be invalid.
   // TODO(qining): Remove this function when instruction building and insertion
   // is well implemented.
-  void SetOpcode(SpvOp opcode) { opcode_ = opcode; }
+  void SetOpcode(SpvOp op) { opcode_ = op; }
   uint32_t type_id() const { return type_id_; }
   uint32_t result_id() const { return result_id_; }
   // Returns the vector of line-related debug instructions attached to this
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -86,6 +86,10 @@ class Instruction {
               std::vector<Instruction>&& dbg_line = {});
 
   SpvOp opcode() const { return opcode_; }
+  // Sets the opcode of this instruction to a specific opcode. Note this may
+  // make the instruction to be invalid.
+  // TODO(qining): Remove this function when instruction building and insertion
+  // is well implemented.
   void SetOpcode(SpvOp opcode) { opcode_ = opcode; }
   uint32_t type_id() const { return type_id_; }
   uint32_t result_id() const { return result_id_; }
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -87,7 +87,7 @@ class Instruction {
 
   SpvOp opcode() const { return opcode_; }
   // Sets the opcode of this instruction to a specific opcode. Note this may
-  // make the instruction to be invalid.
+  // invalidate the instruction.
   // TODO(qining): Remove this function when instruction building and insertion
   // is well implemented.
   void SetOpcode(SpvOp op) { opcode_ = op; }
--- a/source/opt/passes.cpp
+++ b/source/opt/passes.cpp
@@ -53,6 +53,11 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
     } else if (inst->opcode() == SpvOp::SpvOpSpecConstantFalse) {
       inst->SetOpcode(SpvOp::SpvOpConstantFalse);
       modified = true;
+    } else if (inst->opcode() == SpvOp::SpvOpDecorate &&
+        inst->GetInOperand(1).words[0] == SpvDecoration::SpvDecorationSpecId) {
+      // If this is a decoration instrution and the decoration is Spec ID, we
+      // should remove the instrution.
+      inst->ToNop();
     }
   });
   return modified;
--- a/source/opt/passes.cpp
+++ b/source/opt/passes.cpp
@@ -41,5 +41,22 @@ bool StripDebugInfoPass::Process(ir::Module* module) {
   return modified;
 }
 
+bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
+  bool modified = false;
+  module->ForEachInst([&modified](ir::Instruction* inst) {
+    if (inst->opcode() == SpvOp::SpvOpSpecConstant) {
+      inst->SetOpcode(SpvOp::SpvOpConstant);
+      modified = true;
+    } else if (inst->opcode() == SpvOp::SpvOpSpecConstantTrue) {
+      inst->SetOpcode(SpvOp::SpvOpConstantTrue);
+      modified = true;
+    } else if (inst->opcode() == SpvOp::SpvOpSpecConstantFalse) {
+      inst->SetOpcode(SpvOp::SpvOpConstantFalse);
+      modified = true;
+    }
+  });
+  return modified;
+}
+
 }  // namespace opt
 }  // namespace spvtools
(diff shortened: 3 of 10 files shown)


source/assembly_grammar.cpp
```diff
@@ -150,6 +150,7 @@ const SpecConstantOpcodeEntry kOpSpecConstantOpcodes[] = {
     CASE(Select),
     // Comparison
     CASE(IEqual),
+    CASE(INotEqual),
```
dneto0: Please make this a separate change.  It implies a spec change, which is a big thing and needs to be resolved separately.

source/assembly_grammar.cpp
```diff
@@ -166,7 +167,7 @@ const SpecConstantOpcodeEntry kOpSpecConstantOpcodes[] = {
 };
 
 // The 58 is determined by counting the opcodes listed in the spec.
```
dneto0: Remove the "58" from the comment. It's stale!

source/opt/instruction.h
```diff
@@ -86,6 +86,7 @@ class Instruction {
               std::vector<Instruction>&& dbg_line = {});
 
   SpvOp opcode() const { return opcode_; }
+  void SetOpcode(SpvOp opcode) { opcode_ = opcode; }
```
dneto0: This is potentially unsafe.  This method should get a blurb, and the blurb should mention that in many cases operands will have to change!

test/opt/test_freeze_spec_const.cpp
```diff
@@ -0,0 +1,72 @@
...
+                            "%2 = OpConstant %1 1"),
+            std::make_tuple("%1 = OpTypeInt 32 0", "%2 = OpSpecConstant %1 1",
+                            "%2 = OpConstant %1 1"),
+            std::make_tuple("%1 = OpTypeFloat 32",
+                            "%2 = OpSpecConstant %1 3.14",
+                            "%2 = OpConstant %1 3.14"),
+            std::make_tuple("%1 = OpTypeFloat 64",
+                            "%2 = OpSpecConstant %1 3.1415926",
+                            "%2 = OpConstant %1 3.1415926"),
+            std::make_tuple("%1 = OpTypeBool", "%2 = OpSpecConstantTrue %1",
```
dneto0: Should test OpSpecConstantFalse

dneto0:
You should also remove the SpecId decoration on the instructions that you modify. Otherwise the module is invalid.

dneto0:
Review complete.

source/assembly_grammar.cpp
Qining: Undo'ed this change in the 2nd commit

source/assembly_grammar.cpp
Qining: Should I remove this '58' here? Or in the CL that resolves the spec issue?

source/opt/instruction.h
Qining: Blurb added in 2nd CL.

test/opt/test_freeze_spec_const.cpp
Qining: Added in 2nd CL.

test/opt/test_freeze_spec_const.cpp
```diff
@@ -0,0 +1,127 @@
...
+          "%5 = OpLabel",
+               "OpReturn",
+               "OpFunctionEnd",
+      // clang-format on
+  };
+  std::string optimized;
+  bool modified = false;
+  std::tie(optimized, modified) =
+      SinglePassRunAndGetDisassembly<opt::FreezeSpecConstantValuePass>(
+          JoinAllInsts(text));
```
Qining: I found it is hard to use the SinglePassRunAndCheck() here. Optimization and assembling again may change the variable ID so a plain text comparison will fail.

source/opt/instruction.h
```diff
@@ -86,6 +86,11 @@ class Instruction {
               std::vector<Instruction>&& dbg_line = {});
 
   SpvOp opcode() const { return opcode_; }
+  // Sets the opcode of this instruction to a specific opcode. Note this may
+  // make the instruction to be invalid.
```
antiagainst: `invalidate the instruction`.

source/opt/passes.cpp
```diff
@@ -41,5 +41,27 @@ bool StripDebugInfoPass::Process(ir::Module* module) {
   return modified;
 }
 
+bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
+  bool modified = false;
+  module->ForEachInst([&modified](ir::Instruction* inst) {
+    if (inst->opcode() == SpvOp::SpvOpSpecConstant) {
```
antiagainst: I'd suggest to create a local variable for `inst->opcode()`.

source/opt/passes.cpp
```diff
@@ -41,5 +41,27 @@ bool StripDebugInfoPass::Process(ir::Module* module) {
...
+      inst->SetOpcode(SpvOp::SpvOpConstant);
+      modified = true;
+    } else if (inst->opcode() == SpvOp::SpvOpSpecConstantTrue) {
+      inst->SetOpcode(SpvOp::SpvOpConstantTrue);
+      modified = true;
+    } else if (inst->opcode() == SpvOp::SpvOpSpecConstantFalse) {
+      inst->SetOpcode(SpvOp::SpvOpConstantFalse);
+      modified = true;
+    } else if (inst->opcode() == SpvOp::SpvOpDecorate &&
+        inst->GetInOperand(1).words[0] == SpvDecoration::SpvDecorationSpecId) {
```
antiagainst: There is a `GetSingleWordInOperand()` for `GetInOperand().words[0]`.

source/opt/passes.h
```diff
@@ -58,6 +58,14 @@ class StripDebugInfoPass : public Pass {
   bool Process(ir::Module* module) override;
 };
 
+class FreezeSpecConstantValuePass : public Pass {
+ public:
+  const char* name() const override {
+    return "Freeze spec constants to their default value";
```
antiagainst: This looks like a description instead of a name for me. ;)

test/opt/pass_fixture.h
```diff
@@ -52,26 +53,50 @@ class PassTest : public TestT {
...
 
+  // Runs the given |pass| on the binary assembled from the |assembly|, and
+  // disassebles the optimized binary. Returns a tuple of disassembly string
+  // and the boolean value returned from pass Process() function.
+  std::tuple<std::string, bool> GetOptimizedDisassembly(
+      opt::Pass* pass, const std::string& original, bool skip_nop = false) {
+    std::unique_ptr<ir::Module> module = tools_.BuildModule(original);
+    EXPECT_NE(nullptr, module);
+    if (!module) {
+      return std::make_tuple(std::string(), false);
```
antiagainst: `return {{}, false}`?

test/opt/pass_fixture.h
```diff
@@ -52,26 +53,50 @@ class PassTest : public TestT {
...
+
+    std::vector<uint32_t> binary;
+    module->ToBinary(&binary, skip_nop);
+    std::string optimized;
+    EXPECT_EQ(SPV_SUCCESS, tools_.Disassemble(binary, &optimized));
+    return std::make_tuple(optimized, modified);
+  }
+
+  // Runs a single pass of class |PassT| on the binary assembled from the
+  // |assembly|, disassebles the optimized binary. Returns a tuple of
```
antiagainst: s/disassebles/disassembles/

test/opt/pass_fixture.h
```diff
@@ -52,26 +53,50 @@ class PassTest : public TestT {
...
+  // disassembly string and the boolean value from the pass Process() function.
+  template <typename PassT>
+  std::tuple<std::string, bool> SinglePassRunAndGetDisassembly(
```
antiagainst: `SinglePassRunAndDisassemble`?

test/opt/pass_fixture.h
antiagainst: Hmm, seems like it's C++14.

test/opt/test_freeze_spec_const.cpp
```diff
@@ -0,0 +1,127 @@
...
+         "%11 = OpSpecConstant %10 3.14159265358979",
+         "%12 = OpTypeBool",
+         "%13 = OpSpecConstantTrue %12",
+          "%4 = OpFunction %2 None %3",
+          "%5 = OpLabel",
+               "OpReturn",
+               "OpFunctionEnd",
+      // clang-format on
+  };
+  std::string optimized;
```
antiagainst: Looks like we can reduce the above test case and use friendly names for ids in the above so that we don't need to duplicate the check logic in the above? I mean, using `SinglePassCheck` and `SelectiveJoin`.

source/opt/passes.cpp
dneto0: I think this would read better as a switch statement.  And in that case you don't need the local variable for t inst->opcode()

test/opt/test_freeze_spec_const.cpp
Qining: Not using SinglePassRunAndCheck() here because we will have ID mismatch for some variables.
#280 might partially mitigate the issue here, but it seems that eventually we will need a 'id agnostic' matcher.

source/opt/passes.h
dneto0: We should have a rule that the name() is the same as the option used on the spirv-opt command line.

test/opt/test_freeze_spec_const.cpp
Qining: Changed back to using SinglePassRunAndCheck() with SelectiveJoin() and new added FindAndReplace() utility function in commit: 72af6c7bec38e13c9f3353c82638e6d06d38eded

source/opt/instruction.h
Qining: Fixed in 72af6c7bec38e13c9f3353c82638e6d06d38eded

source/opt/passes.cpp
Qining: Use switch() in 72af6c7bec38e13c9f3353c82638e6d06d38eded

source/opt/passes.cpp
Qining: Use GetSingleWordInOperand() in 72af6c7bec38e13c9f3353c82638e6d06d38eded

source/opt/passes.h
Qining: Changed StripDebugInfoPass::name() to return "strip-debug"
Changed FreezeSpecConstantValuePass::name() to return "freeze-spec-const"
Both match with their command line arguments

test/opt/pass_fixture.h
Qining: Fixed in 72af6c7bec38e13c9f3353c82638e6d06d38eded

test/opt/pass_fixture.h
Qining: Fixed in 72af6c7bec38e13c9f3353c82638e6d06d38eded

test/opt/test_freeze_spec_const.cpp
Qining: Fixed in 72af6c7bec38e13c9f3353c82638e6d06d38eded

test/opt/test_freeze_spec_const.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include <algorithm>
+#include <tuple>
+#include <vector>
+
+namespace {
+
+using namespace spvtools;
+
+using FreezeSpecConstantValueTypeTest = PassTest<::testing::TestWithParam<
+    std::tuple<const char*, const char*, const char*>>>;
```
dneto0: It would be nicer to make this a first class struct so we can use readable names for the members.  That will make the tests easier to read.

dneto0:
-1.  Requesting a struct definition to make the test more readable.

test/opt/test_freeze_spec_const.cpp
Qining: Fixed in 4e9132a

dneto0:
LGTM +1

test/opt/test_pass_manager.cpp
```diff
@@ -36,18 +36,18 @@ TEST(PassManager, Interface) {
...
   manager.AddPass<opt::StripDebugInfoPass>();
   EXPECT_EQ(1u, manager.NumPasses());
-  EXPECT_STREQ("StripDebugInfo", manager.GetPass(0)->name());
+  EXPECT_STREQ("strip-debug", manager.GetPass(0)->name());
 
   manager.AddPass(std::unique_ptr<opt::NullPass>(new opt::NullPass));
   EXPECT_EQ(2u, manager.NumPasses());
-  EXPECT_STREQ("StripDebugInfo", manager.GetPass(0)->name());
+  EXPECT_STREQ("strip-debug", manager.GetPass(0)->name());
   EXPECT_STREQ("Null", manager.GetPass(1)->name());
```
antiagainst: I'd suggest to change `Null` to `null` to be consistent.

antiagainst:
LGTM. Just need to change the name of the null pass.

antiagainst:
Oh, besides, don't forget to update the `CHANGES` file.

Qining:
Squashed and merged in a24506266bbdb849785678402d247c43acd234e7

test/opt/test_pass_manager.cpp
Qining: Fixed in dcedb886a88209d8985f7e34af75008ab54b7350

PR #280, opened 2016-07-27
Use friendly disassembly output

Qining:


dneto0:
It seems to me that the SpvTools::Disassemble method should take the options mask as an argument.  It's exactly the kind of thing that someone might want to control even in a higher level API.
It would be ok to have a default argument for the mask.
Otherwise, the change looks ok.

antiagainst:
LGTM.

Didn't go through the test input/ouput text changes carefully, but I won't think it will be a big problem.

Qining:
Squashed and merged d503d75

PR #282, opened 2016-07-27
Optimization: Add def use analysis

antiagainst:


Qining:
Hope to have `GetDef()` and `GetUses()` so that user won't access the map directly.

test/opt/test_def_use.cpp
```diff
@@ -0,0 +1,571 @@
...
+              {7,
+                {
+                  "OpSelectionMerge %7 None",
+                  "OpBranchConditional %8 %9 %7",
+                }
+              },
+              {8, {"OpBranchConditional %8 %9 %7"}},
+              {9, {"OpBranchConditional %8 %9 %7"}},
+            }
+          }
```
Qining: What about add some tests for forward-reference?

source/opt/def_use_manager.h
```diff
@@ -0,0 +1,81 @@
...
+  // structures in this class.
+  void AnalyzeDefUse(ir::Module* module);
+  // Replaces all uses of |before| id with |after| id. Returns true if any
+  // replacement happens. This method does not kill the definition of the
+  // |before| id.
+  bool ReplaceAllUsesWith(uint32_t before, uint32_t after);
+  // Turns the instruction defining the given |id| into a Nop. Returns true on
+  // success, false if the given |id| is not defined at all. This method does
+  // not erase the uses of |id|.
+  bool KillDef(uint32_t id);
```
Qining: I think we may have a check of the uses for the deleting id here? This is easy to check and looks useful to me.

antiagainst: You mean return some error code to show that the id to be deleted still have uses?

antiagainst:
Makes sense, I'll add `GetUses` and `GetDef`.

source/opt/def_use_manager.h
Qining: Oops, never mind, now I think that is not necessary.

test/opt/test_def_use.cpp
antiagainst: Done.

antiagainst:
Added `GetDef` and `GetUses`.

Qining:
LGTM +1

source/opt/def_use_manager.cpp
```diff
@@ -0,0 +1,93 @@
...
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#include <cassert>
+#include <functional>
+
+#include "def_use_manager.h"
```
dneto0: This should go first.  It's the header that defines the interface for this module.

source/opt/def_use_manager.h
```diff
@@ -0,0 +1,89 @@
...
+
+#include <list>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace spvtools {
+
+namespace ir {
+class Instruction;
```
dneto0: FYI.  Google style guide discourages forward declarations.
Ok to keep for now.

source/opt/def_use_manager.h
```diff
@@ -0,0 +1,89 @@
...
+// A class for analyzing and managing defs and uses in an ir::Module.
+class DefUseManager {
+ public:
+  using InstLocPair = std::pair<ir::Instruction*, uint32_t>;
+  using IdToDefMap = std::unordered_map<uint32_t, ir::Instruction*>;
+  using IdToUseMap = std::unordered_map<uint32_t, std::list<InstLocPair>>;
+
+  // Analyzes the defs and uses in the given |module| and populates data
+  // structures in this class.
+  void AnalyzeDefUse(ir::Module* module);
```
dneto0: Since this is just an analysis, I expect that Module will not be changed.  So I expected to see the argument as  const ir::Module& module

But you can't do that since there is no const overload of Module::ForEachInst(const std::function<void(const Instruction*)> & ?

source/opt/def_use_manager.h
```diff
@@ -0,0 +1,89 @@
...
+  std::list<InstLocPair>* GetUses(uint32_t id);
+
+  // Returns the map from ids to their def instructions.
+  const IdToDefMap& id_to_defs() const { return id_to_defs_; }
+  // Returns the map from ids to their uses in instructions.
+  const IdToUseMap& id_to_uses() const { return id_to_uses_; }
+
+  // Turns the instruction defining the given |id| into a Nop. Returns true on
+  // success, false if the given |id| is not defined at all. This method does
+  // not erase the uses of |id|.
```
dneto0: Does this method also remove all record of uses of this definition?

source/opt/def_use_manager.h
```diff
@@ -0,0 +1,89 @@
...
+class Module;
+}
+
+namespace opt {
+namespace analysis {
+
+// A class for analyzing and managing defs and uses in an ir::Module.
+class DefUseManager {
+ public:
+  using InstLocPair = std::pair<ir::Instruction*, uint32_t>;
```
dneto0: Should this be called "Use" instead?  It feels like we should have a more first-class concept of a use.

source/opt/def_use_manager.h
```diff
@@ -0,0 +1,89 @@
...
+  using IdToDefMap = std::unordered_map<uint32_t, ir::Instruction*>;
+  using IdToUseMap = std::unordered_map<uint32_t, std::list<InstLocPair>>;
```
dneto0: I strongly suggest making a name for the set-of-uses type.  Especially since std::list will be very slow when there are a lot of elements.

dneto0: IdToUseMap -> IdToUsesMap

dneto0: Speaking of which, why use a std::list?  There should be a very good reason to do so, e..g if you want iterators relatively stable.

test/opt/test_def_use.cpp
```diff
@@ -0,0 +1,599 @@
...
+// Disassembles the given |module| and returns the disassembly.
+std::string DisassembleModule(ir::Module* module) {
+  SpvTools tools(SPV_ENV_UNIVERSAL_1_1);
+
+  std::vector<uint32_t> binary;
+  module->ToBinary(&binary, /* skip_nop = */ false);
+
+  std::string text;
+  // We'll need to check the underlying id numbers.
+  // So turn off friendly names for ids.
```
dneto0: :-)

source/opt/def_use_manager.cpp
```diff
@@ -0,0 +1,93 @@
...
+       ++it) {
+    auto* inst = it->first;
+    auto loc = it->second;
+    // Make the modification in the instruction.
+    inst->SetOperand(loc, {after});
+    // Register the use of |after| id into id_to_uses_.
+    // TODO(antiagainst): de-duplication.
+    id_to_uses_[after].emplace_back(inst, loc);
+  }
+  id_to_uses_.erase(before);
```
dneto0: This is wrong if before == after.
If before==after then this should be a no-op.  (That might be worth mentioning in the blurb for the function.)

dneto0: And we'll need a test case for this.

source/opt/def_use_manager.h
dneto0: Actually, it's really unclear what the uint32_t is.  Is it the logical operand Id?  Could it be the type Id for the target instruction?

source/opt/def_use_manager.cpp
```diff
@@ -0,0 +1,93 @@
...
+  return true;
+}
+
+void DefUseManager::AnalyzeInstDefUse(ir::Instruction* inst) {
+  if (inst->result_id() != 0) id_to_defs_[inst->result_id()] = inst;
+
+  for (uint32_t i = 0; i < inst->NumOperands(); ++i) {
+    const auto type = inst->GetOperand(i).type;
+    if (type == SPV_OPERAND_TYPE_ID || type == SPV_OPERAND_TYPE_TYPE_ID ||
```
dneto0: Worth having a comment here: any Id operand that is not a result-id.
(Also, I prefer switches in this kind of code...)

test/opt/test_def_use.cpp
```diff
@@ -0,0 +1,599 @@
...
+      { // no use, no replace request
+        "", {}, "", {},
+      },
+      { // no use, some replace requests
+        "OpMemoryModel Logical GLSL450",
+        {{1, 2}, {3, 4}, {7, 8}, {7, 9}, {7, 10}, {2, 10}, {3, 10}},
+        "OpMemoryModel Logical GLSL450",
+        {},
+      },
+      { // replace one use
```
dneto0: Add test case:  Replace an Id with itself.

test/opt/test_def_use.cpp
```diff
@@ -0,0 +1,599 @@
...
+              {
+                "%2 = OpTypeVector %10 2",
+                "%3 = OpTypeVector %10 3",
+                "%4 = OpTypeVector %10 4",
+              }
+            },
+            {20, {"%5 = OpTypeMatrix %20 2"}},
+            {3, {"%6 = OpTypeMatrix %3 3"}},
+            {40, {"%7 = OpTypeMatrix %40 4"}},
+          },
```
dneto0: Another test case:
  Replace a case Id in OpSwitch over 64-bit integers.  (other than the default case)
This is interesting because it will check that you are replacing the right word when the literals are 2 words wide.

(We also don't have lower level tests for SetOperand and GetSingleWordOperand.  But at least a case here will check it at slightly above unit test level.)

test/opt/test_def_use.cpp
```diff
@@ -0,0 +1,599 @@
...
+        {
+          {
+            {2, "%2 = OpTypeVector %1 2"},
+            {4, "%4 = OpTypeVector %1 4"},
+          },
+          {
+            {1,
+              {
+                "%2 = OpTypeVector %1 2",
+                "OpNop", // %3 = OpTypeVector %1 3
```
dneto0: This is unexpected.  The OpNop is no longer a use of 1. 
KillDef should remove the uses , I think.

test/opt/test_def_use.cpp
```diff
@@ -0,0 +1,599 @@
...
+                "%4 = OpTypeVector %1 4",
+              }
+            },
+            {3,
+              {
+                "OpNop" // %5 = OpTypeMatrix %3 3
```
dneto0: Also will need a test case where killing a def will make someone's uses list empty, in which case that list should be removed?

dneto0:
Review complete.

source/opt/def_use_manager.cpp
```diff
@@ -0,0 +1,93 @@
...
+  module->ForEachInst(std::bind(&DefUseManager::AnalyzeInstDefUse, this,
+                                std::placeholders::_1));
+}
+
+ir::Instruction* DefUseManager::GetDef(uint32_t id) {
+  if (id_to_defs_.count(id) == 0) return nullptr;
+  return id_to_defs_.at(id);
+}
+
+std::list<DefUseManager::InstLocPair>* DefUseManager::GetUses(uint32_t id) {
```
Qining: Maybe not necessary in this CL. But it would be better if we define a type for std::list<DefUseManager::InstLocPair>

test/opt/test_def_use.cpp
```diff
@@ -0,0 +1,599 @@
...
+  manager.AnalyzeDefUse(module.get());
+
+  CheckDefUse(tc.du, manager.id_to_defs(), manager.id_to_uses());
+}
+
+// clang-format off
+INSTANTIATE_TEST_CASE_P(
+    TestCase, ParseDefUseTest,
+    ::testing::ValuesIn(std::vector<ParseDefUseCase>{
+        {"", {{}, {}}},                              // no instruction
```
dneto0: Please add a test case where an OpPhi is both the definintion and use of a value.

Note that at some point LLVM decided that mention of a value on the right-hand-side of a phi is *not* considered a use.  But our code does consider it a use.
(And we might want to make that clear in the comments for the DefUseManager)

test/opt/test_def_use.cpp
```diff
@@ -0,0 +1,599 @@
...
+}
+
+// clang-format off
+INSTANTIATE_TEST_CASE_P(
+    TestCase, ReplaceUseTest,
+    ::testing::ValuesIn(std::vector<ReplaceUseCase>{
+      { // no use, no replace request
+        "", {}, "", {},
+      },
+      { // no use, some replace requests
```
dneto0: Please add a test case where an OpPhi both produces and consumes a value.

test/opt/test_def_use.cpp
```diff
@@ -0,0 +1,599 @@
...
+}
+
+// clang-format off
+INSTANTIATE_TEST_CASE_P(
+    TestCase, KillDefTest,
+    ::testing::ValuesIn(std::vector<KillDefCase>{
+      { // no def, no use, no kill
+        "", {}, "", {}
+      },
+      { // kill nothing
```
dneto0: Add a test case where an OpPhi both produces and consumes a value.

dneto0:
(Very nice test cases, by the way!)

test/opt/test_def_use.cpp
Qining: I think it would be clearer with && ~SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES

source/opt/def_use_manager.cpp
antiagainst: Done.

source/opt/def_use_manager.h
antiagainst: Done.

source/opt/def_use_manager.h
antiagainst: Right... I'll add the const overload later.

source/opt/def_use_manager.h
antiagainst: Now it does.

source/opt/def_use_manager.h
antiagainst: Added a `Use` struct and comments for it.

source/opt/def_use_manager.h
antiagainst: Done.

`list` is good for removing in the middle. Needed in the new impl for `KillDef`.

source/opt/def_use_manager.cpp
antiagainst: Good catch! Done. Test case added.

source/opt/def_use_manager.cpp
antiagainst: Done. Changed to `switch`.

test/opt/test_def_use.cpp
antiagainst: Done.

test/opt/test_def_use.cpp
antiagainst: I've a spotcheck test for it now.

test/opt/test_def_use.cpp
antiagainst: No longer like this.

test/opt/test_def_use.cpp
antiagainst: Now the `OpPhi` test case demonstrates this.

source/opt/def_use_manager.cpp
antiagainst: Done.

test/opt/test_def_use.cpp
antiagainst: Done.

test/opt/test_def_use.cpp
antiagainst: Done.

test/opt/test_def_use.cpp
antiagainst: Done.

test/opt/test_def_use.cpp
antiagainst: Added.

antiagainst:
Addressed all comments in the fourth commit.

source/opt/def_use_manager.h
```diff
@@ -32,54 +32,70 @@
...
+  using IdToParamMap = std::unordered_map<uint32_t, std::vector<uint32_t>>;
+
   // Analyzes the defs and uses in the given |inst|.
   void AnalyzeInstDefUse(ir::Instruction* inst);
 
-  IdToDefMap id_to_defs_;  // Mapping from ids to their definitions
-  IdToUseMap id_to_uses_;  // Mapping from ids to their uses
+  IdToDefMap id_to_def_;       // Mapping from ids to their definitions
+  IdToUsesMap id_to_uses_;     // Mapping from ids to their uses
+  IdToParamMap id_to_params_;  // Mapping from ids to those ids used as operands
```
dneto0: "as operands" -> "as function parameters" ?

source/opt/def_use_manager.cpp
```diff
@@ -42,48 +42,76 @@ void DefUseManager::AnalyzeDefUse(ir::Module* module) {
...
+    switch (inst->GetOperand(i).type) {
+      // For any id type but result id type
+      case SPV_OPERAND_TYPE_ID:
+      case SPV_OPERAND_TYPE_TYPE_ID:
+      case SPV_OPERAND_TYPE_MEMORY_SEMANTICS_ID:
+      case SPV_OPERAND_TYPE_SCOPE_ID: {
+        uint32_t use_id = inst->GetSingleWordOperand(i);
+        // use_id is used by the instruction generating def_id.
+        id_to_uses_[use_id].push_back({inst, i});
+        id_to_params_[def_id].push_back(use_id);
```
dneto0: What if def_id is 0?  This will happen if the current instruction is something like OpDecorate (which does not return a result id).
Seem wrong.

source/opt/def_use_manager.h
dneto0: Hm.  Now that I look at the code to build this, it's a map from the Ids of instructions that use an id, to the list of ids they use, and it can contain duplicates.
Better to call this id_to_used_ids_ or at least have a better comment.

test/opt/test_def_use.cpp
```diff
@@ -47,7 +52,9 @@ std::string DisassembleInst(ir::Instruction* inst) {
   std::string text;
   // We'll need to check the underlying id numbers.
   // So turn off friendly names for ids.
-  tools.Disassemble(binary, &text, SPV_BINARY_TO_TEXT_OPTION_NO_HEADER);
+  tools.Disassemble(binary, &text,
+                    SPV_BINARY_TO_TEXT_OPTION_NO_HEADER &
+                        ~SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES);
```
dneto0: I'm surprised you have to explicitly exclude FRIENDLY_NAMES since it's not the default. ?

test/opt/test_def_use.cpp
```diff
@@ -595,5 +884,123 @@ INSTANTIATE_TEST_CASE_P(
     })
 );
 // clang-format on
+//
+TEST(DefUseTest, OpSwith) {
```
dneto0: "OpSwitch"

dneto0:
I'd like to see a case where the same instruction both defines and uses an Id.  This is valid for an OpPhi.

E.g.
  
```
  %entry = OpLabel
   %zero = OpConstantZero
   OpBranch %loop

 %loop = OpLabel
     %i = OpPhi %i32 %i %loop %zero %entry ; Both defines and uses %i  !!
   OpBranch %loop

test/opt/test_def_use.cpp
antiagainst: It's to make it more clear. But I don't care that much. Will remove.

test/opt/test_def_use.cpp
antiagainst: :)

source/opt/def_use_manager.h
antiagainst: Good call. I've revised the name and comments. It should be more clear now!

source/opt/def_use_manager.cpp
antiagainst: Good catch!! I must be too tired when coding this. Fixed.

test/opt/test_def_use.cpp
antiagainst: Done.

antiagainst:
Cool code snippet!

Didn't anticipated that you can do this kind of work using `OpPhi`. So `OpPhi` is indeed the puppy for compiler testing. ;)

Added.

Addressed all comments thus far.

dneto0:
+2
Thanks for the TODO.  :-)

Issue #279, opened 2016-07-26
validator: goes into infinite loop when a block branches to itself and nothing else


dneto0:
Triggered by this input:
```
           OpCapability Shader
           OpMemoryModel Logical GLSL450
%voidt   = OpTypeVoid
%funct   = OpTypeFunction %voidt

%blah    = OpFunction %voidt None %funct
%entry   = OpLabel
           OpBranch %loop
%loop    = OpLabel
           OpLoopMerge %exit %loop None
           OpBranch %loop
%exit    = OpLabel
           OpReturn
           OpFunctionEnd
```

It's stuck in libspirv::CalculateDominators

dneto0:
The OpCapability Shader is essential to forcing the infinite loop.

umar456:
It seems to happen when you branch with an OpBranch instead of OpBranchConditional. I had a [unit test](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/Validate.CFG.cpp#L639) for this scenario but it only tested the sane case :).

Looking into it now.

dneto0:
Adding some logging, it's only showing up in the post-dom calculation.  It looks like the undefined_dom value is 3 instead of 5 like it is for the (forward) dom calculation.  Might be my bug in how I did the augmentation.   :-(
I'm gonna quit for the night and come at this fresh tomorrow.  If you can see it quickly then great. :-)

umar456:
That's probably because the post-dom tree only contains one node (+2 pseudo nodes).

dneto0:
Ah, right.
Thinking about it offline, I guess the problem is that the exit pseudo node isn't connected to the loop node, but should be (directly or indirectly).  I think we only connect the exit node to nodes that end with OpReturn and others that don't end in branches?   If so this might just need a "reachable-in-reverse" concept from an exit node.

dneto0:
Yes, that's the problem.  When trying to do the intersection of the dominators of predecessors to a given node, we go through the "predecessors".  But one of the predecessors in the post-dom calculation is not reachable from the pseudo-exit node, even in the augmented CFG.  So we end up dereferencing idoms[<unreachable-node>] which default-constructs a block_detail struct and that gives us a dominator member of 0 and a postorder_index member of 0 and so finger1  and finger2 are garbage.

[...]

PR #283, opened 2016-07-28
Fix infinite loop in dominance calculation.

dneto0:
Ensure the dominance calculation visits all nodes in the CFG.
The successor list of the pseudo-entry node is augmented with
a single node in each cycle that otherwise would not be visited.
Similarly, the predecssors list of the pseduo-exit node is augmented
with the a single node in each cycle that otherwise would not
be visited.

Pulls DepthFirstSearch out so it's accessible outside of the dominator
calculation.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/279

### DIFF
--- a/source/val/Function.cpp
+++ b/source/val/Function.cpp
@@ -29,10 +29,13 @@
 #include <cassert>
 
 #include <algorithm>
+#include <unordered_set>
+#include <unordered_map>
 #include <utility>
 
 #include "val/BasicBlock.h"
 #include "val/Construct.h"
+#include "validate.h"
 
 using std::ignore;
 using std::list;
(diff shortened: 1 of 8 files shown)


dneto0:
I'd like @umar456 to take a look.

source/val/Function.cpp
```diff
@@ -41,6 +44,53 @@ using std::pair;
...
+  auto traverse_from_root = [&mark_visited, &succ_func, &ignore_block,
+                             &ignore_blocks](const BasicBlock* entry) {
+    DepthFirstTraversal(entry, succ_func, mark_visited, ignore_block,
+                        ignore_blocks);
+  };
+
+  std::vector<BasicBlock*> result;
+
+  // First collect nodes without predecessors.
+  for (auto block : blocks) {
```
umar456: It would probably be a good idea to perform the reachability propagation using this approach. I don't like how I am performing that now. I can do that in a future PR.

source/validate.h
```diff
@@ -57,6 +57,32 @@ class ValidationState_t;
...
+/// @param[in] preorder   A function that will be called for every block in a
+///                       CFG following preorder traversal semantics
+/// @param[in] postorder  A function that will be called for every block in a
+///                       CFG following postorder traversal semantics
+/// @param[in] backedge   A function that will be called when a backedge is
+///                       encountered during a traversal
+/// NOTE: The @p successor_func and predecessor_func each return a pointer to a
+/// collection such that iterators to that collection remain valid for the
+/// lifetime
+/// of the algorithm
```
umar456: Nit: unnecessary line break

source/val/Function.h
```diff
@@ -149,31 +150,40 @@ class Function {
...
+  /// Returns the block predecessors function for the augmented CFG.
+  GetBlocksFunction AugmentedCFGPredecessorsFunction() const;
 
-  /// Returns a vector with just the pseudo exit block.
-  /// This serves as the successors of each sink node in the CFG when computing
-  /// dominators.
-  const std::vector<BasicBlock*>* pseudo_exit_blocks() const {
-    return &pseudo_exit_blocks_;
-  }
+  /// Returns the block successor function for the augmented CFG.
```
umar456: dangling comment

umar456:
Minor issues. But everything looks good.

source/val/Function.h
dneto0: Fixed in c978b7247

source/validate.h
dneto0: Fixed in c978b7247

dneto0:
Rebased and addressed all comments.

dneto0:
The requested changes were trivial updates to comments.  I'll take it as an LGTM.

PR #284, opened 2016-07-28
Fix a test failure in pass manager test

Qining:
Fix the test failure in a24506266bbdb849785678402d247c43acd234e7
Change "Null" to "null" in pass manager test


antiagainst:
+2

antiagainst:
We discourage using the merge button. So better to avoid doing this next time.

dneto0:
Right.  We only want linear history in the SPIRV-Tools master branch.  So don't use GitHub's merge buttong.  Instead, rebase and then push into master.

Issue #286, opened 2016-07-29
Validator: def-use error incorrectly generated when the use is in an OpPhi


dneto0:
This affects a Vulkan CTS test.

The essence of the case is:
```
           OpCapability Shader
           OpMemoryModel Logical GLSL450

           OpName %copy "copy"
           OpName %false "false"
           OpName %if_true "if_true"
           OpName %exit "exit"

%voidt   = OpTypeVoid
%funct   = OpTypeFunction %voidt
%boolt   = OpTypeBool
%false   = OpConstantFalse %boolt

%main    = OpFunction %voidt None %funct
%entry   = OpLabel
           OpSelectionMerge %exit None
           OpBranchConditional %false %if_true %exit

%if_true = OpLabel
%copy    = OpCopyObject %boolt %false
           OpBranch %exit

%exit    = OpLabel
%v       = OpPhi %boolt %false %entry %copy %if_true
           OpReturn
           OpFunctionEnd
```

The error reported is:   error: 21: ID 1[copy] defined in block 3[if_true] does not dominate its use in block 4[exit]

PR #287, opened 2016-07-29
Phi dominance check

dneto0:
This is a workaround to https://github.com/KhronosGroup/SPIRV-Tools/issues/286 
It disables def-use dominance checks for OpPhi (variable,predecessor) arguments.

### DIFF
--- a/test/Validate.SSA.cpp
+++ b/test/Validate.SSA.cpp
@@ -545,6 +545,7 @@ const string kBasicTypes = R"(
 %zero      = OpConstant %intt 0
 %one       = OpConstant %intt 1
 %ten       = OpConstant %intt 10
+%false     = OpConstantFalse %boolt
 )";
 
 const string kKernelTypesAndConstants = R"(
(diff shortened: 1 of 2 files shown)


dneto0:
I filed isseu https://github.com/KhronosGroup/SPIRV-Tools/issues/288 to track the enhancement that would enable the (proposed) disabled test.

umar456:
Ahh, sorry about that. I had made a special case for OpPhi in my previous implementation but I didn't move to the new commit. This seems like a reasonable workaround until i implement a proper fix.
LGTM +2

dneto0:
Rebased

test/Validate.SSA.cpp
```diff
@@ -1172,6 +1173,67 @@ TEST_F(ValidateSSA, PhiUseDoesntDominateUseOfPhiOperandUsedBeforeDefinitionBad)
...
+%value       = OpPhi %boolt %false %entry %copy %if_true
+               OpReturn
+               OpFunctionEnd
+)";
+
+  CompileSuccessfully(str);
+  ASSERT_EQ(SPV_SUCCESS, ValidateInstructions()) << getDiagnosticString();
+}
+
+TEST_F(ValidateSSA, DISABLED_PhiVariableDefMustComeFromBlockDominatingThePredecessorBad) {
```
umar456: Hmm. I am sure this was the intention but I don't see a reference to this rule anywhere in the spec. Here is what the instruction definition says about the `Parent` `Variable` pair:

```
Operands are a sequence of pairs: (/Variable 1/, /Parent 1/ block), (/Variable
2/, /Parent 2/ block),... Each /Parent i/ block is the label of an immediate
predecessor in the CFG of the current block. A /Parent i/ block must not appear
more than once in the operand sequence. All /Variables/ must have a type
matching /Result Type/.
```

dneto0: It's a requirement for SSA that it work that way. 
The SPIR-V spec says the module must be in SSA form but doesn't spell out all the rules for SSA.
The best I can point to is the section on SSA form: https://www.khronos.org/registry/spir-v/specs/1.1/SPIRV.html#_a_id_ssa_a_ssa_form  especially at the end where it mentions the use of a Phi.

Also, in the universal validation rules under the SSA bullet it says "Uses in a phi-function in a loop may consume definitions in the loop that don’t dominate the use."  Which is true (and vague) but also leaves out the forward-only but divergent then reconvergent control flow case like in this example (diamond structure).

PR #292, opened 2016-08-02
Simple Assembly code builder for test uses

Qining:
AssemblyBuilder contains boilplates and adds OpName instructions for all
added defining instructions.

### DIFF
--- a/test/opt/assembly_builder.h
+++ b/test/opt/assembly_builder.h
@@ -1,0 +1,250 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#ifndef LIBSPIRV_TEST_OPT_ASSEMBLY_BUILDER
+#define LIBSPIRV_TEST_OPT_ASSEMBLY_BUILDER
+
+#include <algorithm>
+#include <sstream>
+#include <string>
+#include <unordered_set>
+#include <vector>
+
+namespace spvtools {
+
+// A simple SPIR-V assembly code builder for test uses. It builds an SPIR-V
+// assembly module from vectors of assembly strings. It allows users to add
+// instructions to the main function and the type-constants-globals section
+// directly. It relies on OpName instructions and friendly-name disassembling
+// to keep the ID names unchanged after assembling.
+//
+// An assembly module is divided into several sections, matching with the
+// SPIR-V Logical Layout:
+//  Global Preamble:
+//    OpCapability instructions;
+//    OpExtension instructions and OpExtInstImport instructions;
+//    OpMemoryModel instruction;
+//    OpEntryPoint and OpExecutionMode instruction;
+//    OpString, OpSourceExtension, OpSource and OpSourceContinued instructions.
+//  Names:
+//    OpName instructions.
+//  Annotations:
+//    OpDecorate, OpMemberDecorate, OpGroupDecorate, OpGroupMemberDecorate and
+//    OpDecorationGroup.
+//  Types, Constants and Global variables:
+//    Types, constants and global variables declaration instructions.
+//  Main Function:
+//    Main function instructions.
+//  Main Function Postamble:
+//    The return and function end instructions.
+//
+// The assembly code is built by concatenating all the strings in the above
+// sections.
+//
+// Users define the contents in section <Type, Constants and Global Variables>
+// and <Main Function>. The <Names> section is to hold the names for IDs to
+// keep them unchanged before and after assembling. All defined IDs to be added
+// to this code builder will be assigned with a global name through OpName
+// instruction. The name is extracted from the definition instruction.
+//  E.g. adding instruction: %var_a = OpConstant %int 2, will also add an
+//  instruction: OpName %var_a, "var_a".
+//
+// Note that the name must not be used on more than one defined IDs and
+// friendly-name disassembling must be enabled so that OpName instructions will
+// be respected.
+class AssemblyBuilder {
+  // The base ID value for spec constants.
+  static const uint32_t SPEC_ID_BASE = 200;
+
+ public:
+  // Initalize a minimal SPIR-V assembly code as the template. The minimal
+  // module contains an empty main function and some predefined names for the
+  // main function.
+  AssemblyBuilder()
+      : spec_id_counter_(SPEC_ID_BASE),
+        global_preamble_({
+            // clang-format off
+                  "OpCapability Shader",
+                  "OpCapability Float64",
+             "%1 = OpExtInstImport \"GLSL.std.450\"",
+                  "OpMemoryModel Logical GLSL450",
+                  "OpEntryPoint Vertex %main \"main\"",
+            // clang-format on
+        }),
+        names_(),
+        annotations_(),
+        types_consts_globals_(),
+        main_func_(),
+        main_func_postamble_({
+            "OpReturn", "OpFunctionEnd",
+        }) {
+    AppendTypesConstantsGlobals({
+        "%void = OpTypeVoid", "%main_func_type = OpTypeFunction %void",
+    });
+    AppendInMain({
+        "%main = OpFunction %void None %main_func_type",
+        "%main_func_entry_block = OpLabel",
+    });
+  }
+
+  // Appends instructions to the types-constants-globals section and returns
+  // the reference of this assembly builder. IDs defined in the given code will
+  // be added to the Names section and then be registered with OpName
+  // instruction. Corresponding decoration instruction will be added for spec
+  // constants defined with opcode: 'OpSpecConstant'.
+  AssemblyBuilder& AppendTypesConstantsGlobals(
+      const std::vector<std::string>& vec_asm_code) {
+    AddNamesForResultIDsIn(vec_asm_code);
+    // Check spec constants defined with OpSpecConstant.
+    for (auto& inst_str : vec_asm_code) {
+      if (inst_str.find("= OpSpecConstant ") != std::string::npos ||
+          inst_str.find("= OpSpecConstantTrue ") != std::string::npos ||
+          inst_str.find("= OpSpecConstantFalse ") != std::string::npos) {
+        AddSpecIDFor(GetResultIDName(inst_str));
+      }
+    }
+    types_consts_globals_.insert(types_consts_globals_.end(),
+                                 vec_asm_code.begin(), vec_asm_code.end());
+    return *this;
+  }
+
+  // Appends instructions to the main function block, which is already labelled
+  // with "main_func_entry_block". Returns the reference of this assembly
+  // builder. IDs defined in the given code will be added to the Names section
+  // and then be registered with OpName instruction.
+  AssemblyBuilder& AppendInMain(const std::vector<std::string>& vec_asm_code) {
+    AddNamesForResultIDsIn(vec_asm_code);
+    main_func_.insert(main_func_.end(), vec_asm_code.begin(),
+                      vec_asm_code.end());
+    return *this;
+  }
+
+  // Appends annotation instructions to the annotation section, and returns the
+  // reference of this assembly builder.
+  AssemblyBuilder& AppendAnnotations(
+      const std::vector<std::string>& vec_annotations) {
+    annotations_.insert(annotations_.end(), vec_annotations.begin(),
+                        vec_annotations.end());
+    return *this;
+  }
+
+  // Get the SPIR-V assembly code as string.
+  std::string GetCode() const {
+    std::ostringstream ss;
+    for (const auto& line : global_preamble_) {
+      ss << line << std::endl;
+    }
+    for (const auto& line : names_) {
+      ss << line << std::endl;
+    }
+    for (const auto& line : annotations_) {
+      ss << line << std::endl;
+    }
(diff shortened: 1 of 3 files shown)


test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+
+#include <algorithm>
+#include <sstream>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace spvtools {
+
+// A simlple SPIR-V assmebly code builder for test uses. It builds an SPIR-V
```
dneto0: "simple"  "assembly"

test/opt/pass_utils.h
```diff
@@ -27,8 +27,11 @@
 #ifndef LIBSPIRV_TEST_OPT_PASS_UTILS_H_
 #define LIBSPIRV_TEST_OPT_PASS_UTILS_H_
 
+#include <algorithm>
```
dneto0: Why add new header dependencies here?  Only add them if the interface described in this header needs them.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+// assembly module from vectors of assembly strings. It allows users to add
+// instructions to the main function and the type-constants-globals section
+// directly. It relies on OpName instructions and friendly-name disassembling
+// to keep the ID names unchanged after assembling.
+//
+// An assembly module is divided into several sections:
```
dneto0: Do these correspond to ranges of sections as in the Logical Layout?  If yes, then it would be nice to say so.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+//  Global Preamble:
+//    declare capabilities, extensions, entry points, etc.
+//  Names:
+//    declare the names for IDs with OpName instructions.
+//  Types, Constants and Global variables:
+//    declare types, constants and global variables.
+//  Main Function:
```
dneto0: Is this really restricted to just one function?

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+//    contains instructions in main function, just after OpLable for the entry
```
dneto0: "OpLabel"

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+//    block.
+//  Main Function Postamble:
+//    contains the return and fucntion end instruction.
+//
+// The assembly code is built by concatenating all the strings in the above
+// section.
```
dneto0: "sections"

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+// to this code builder will be assigned with a global name through OpName
+// instruction. The name is extracted from the definition instruction.
+//  E.g. adding instruction: %var_a = OpConstant %int 2, will also add an
+//  instruction: OpName %var_a, "var_a".
+//
+// Note that the name must not be used on more than defined IDs and friendly-name
+// disassembling must be enabled so that OpName instructions will be respected.
+class AssemblyBuilder {
+ public:
+  // Initalize an minimal SPIR-V assembly code as the template. The minimal
```
dneto0: "a minimal ..."
"The minimal module contains..."

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+                             [](char c) { return c == ' ' || c == '%'; }),
+                   name.end());
+        AddOpNameIfNotExist(name);
+      }
+    }
+  }
+
+  // The vector that contains common preambles shared across all test SPIR-V
+  // code.
+  std::vector<std::string> global_preamble_vec_;
```
dneto0: I'd avoid the "vec"  part of these names.  The reader already knows it's a vector from its type.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+  // variables (aka. the Types-Constants-Globals section).
+  std::vector<std::string> types_consts_globals_vec_;
+  // The vector that contains the preamble of main function.
+  std::vector<std::string> main_func_preamble_vec_;
+  // The vector that contains the code in main function's entry block.
+  std::vector<std::string> main_func_entry_block_vec_;
+  // The vector that contains the postamble of main function body.
+  std::vector<std::string> main_func_postamble_vec_;
+  // All of the defined variable names.
+  std::unordered_map<std::string, size_t> names_;
```
dneto0: What does the name string map to?

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+            // clang-format on
+        }),
+        main_func_entry_block_vec_(),
+        main_func_postamble_vec_({
+            // clang-format off
+                  "OpReturn",
+                  "OpFunctionEnd",
+            // clang-format on
+        }),
+        names_({{"main", 0}, {"main_func_entry_block", 1}, {"main_func_type", 2}}) {}
```
dneto0: This seems coupled to the initialization of names_vec_.   Better to leave the initialization empty, then construct the values automatically in the body of the constructor.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+
+  // Appends instructions to the types-constants-globals section.
```
dneto0: Say what it returns.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+      const std::vector<std::string>& vec_asm_code) {
+    AddNamesFor(vec_asm_code);
+    types_consts_globals_vec_.insert(types_consts_globals_vec_.end(),
+                                     vec_asm_code.begin(), vec_asm_code.end());
+    return *this;
+  }
+
+  // Appends instructions to the main function block, start within
+  // "main_func_entry_block".
+  AssemblyBuilder& AppendInMain(const std::vector<std::string>& vec_asm_code) {
```
dneto0: Say what it returns.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+    AddNamesFor(vec_asm_code);
+    main_func_entry_block_vec_.insert(main_func_entry_block_vec_.end(),
+                                      vec_asm_code.begin(), vec_asm_code.end());
+    return *this;
+  }
+
+  // Get the SPIR-V assembly code in string.
+  std::string GetCode() const {
+    std::stringstream ss;
+    for (auto& line : global_preamble_vec_) {
```
dneto0: This is a lot of repeated code.
I wonder if instead we should have had an array of vectors, and the first level is indexed by segment enum like kGlobalPreamble, kNames, kTypesConstantsGLobals, kMainFuncPreamble etc.

But not worth changing now.  We could think about this later.

test/opt/assembly_builder.h
Qining: This one only has one function. But we can have another 'section' to contain whatever instructions an user wants to have. Just for now the main function is enough for my tests.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,207 @@
...
+// disassembling must be enabled so that OpName instructions will be respected.
+class AssemblyBuilder {
+ public:
+  // Initalize an minimal SPIR-V assembly code as the template. The minimal
+  // contains an empty main function and some predefined names for the main
+  // function.
+  AssemblyBuilder()
+      : global_preamble_vec_({
+            // clang-format off
+                  "OpCapability Shader",
```
dneto0: This seems rather specific.  But it's ok because it's all you've needed.  If we need to, we can always refactor later.

test/opt/test_assembly_builder.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+#include "pass_fixture.h"
+#include "pass_utils.h"
+
+namespace {
+
+using namespace spvtools;
+using AssemblyBuilderTest = PassTest<::testing::Test>;
+
+TEST_F(AssemblyBuilderTest, MinialShader) {
```
dneto0: "MinimalShader"

dneto0:
-1 
Relatively minor feedback.

test/opt/assembly_builder.h
Qining: I was thinking about that. Matching with Logical Layout is definitely good, just a little bit overkill for my current tests, like the capabilities, extension and memory models usually do not change across passes for now. I think it depends on how far we want to go with this simple builder (so far it is just extracted from the tests for dead-constant-elimination).

test/opt/assembly_builder.h
Qining: Fixed in 1ab7d6689503e71a08a0809a4009722a93c434fe

test/opt/assembly_builder.h
Qining: Fixed in 1ab7d6689503e71a08a0809a4009722a93c434fe

test/opt/assembly_builder.h
Qining: Removed the vec_ suffix in 1ab7d6689503e71a08a0809a4009722a93c434fe

test/opt/assembly_builder.h
Qining: It maps to the position of the corresponding OpName instruction in the name_ (original names_vec_) vector. Currently just using it as a set, the position is not used, but I was thinking about removing instructions (now I think removing might not be necessary at all).

test/opt/assembly_builder.h
Qining: Fixed in 1ab7d6689503e71a08a0809a4009722a93c434fe

test/opt/assembly_builder.h
Qining: Good idea. I should have a TODO here I guess?

test/opt/assembly_builder.h
Qining: Fixed in 1ab7d6689503e71a08a0809a4009722a93c434fe

test/opt/assembly_builder.h
dneto0: My point was that it would match *ranges*. That's important because the entire assembly is a plain concatenation of the regions you've got now.   I think the answer to my original question is "yes, each of your regions here corresponds to contiguous subsets of the logical layout sections".  
No big deal though.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,247 @@
...
+    }
+    for (auto& line : main_func_postamble_) {
+      ss << line << std::endl;
+    }
+    return ss.str();
+  }
+
+ private:
+  // Adds a given name to the Name section with OpName. If the given name has
+  // beed added before, do nothing.
```
dneto0: "beed" -> "been"

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,247 @@
...
+  // The vector that contains the code to declare types, constants and global
+  // variables (aka. the Types-Constants-Globals section).
+  std::vector<std::string> types_consts_globals_;
+  // The vector that contains the code in main function's entry block.
+  std::vector<std::string> main_func_;
+  // The vector that contains the postamble of main function body.
+  std::vector<std::string> main_func_postamble_;
+  // All of the defined variable names and the positions of their OpName
+  // instructions in the 'Names' section.
+  std::unordered_map<std::string, size_t> names_bookkeepping;
```
dneto0: Ok.  I saw in your comment you're only using it as a set.
  I recommend just using an unordered set then (YAGNI), and call it used_names_.
The "names_bookkeeping" name is not very descriptive.  If you do want to store the index in the future, then call it "name_index_".  "Bookkeeping" is too generic.

test/opt/assembly_builder.h
dneto0: Either way.  No need to add the TODO.

dneto0:
-1 relatively minor feedback
 "beed" -> "been"
and change the name bookkeeping to just a set.

Qining:
commit: 3e90a70 fix:
1. misspelling "beed" -> "been"
2. name_bookkeeping -> used_names_

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,246 @@
...
+//
+// Users define the contents in section <Type, Constants and Global Variables>
+// and <Main Function>. The <Names> section is to hold the names for IDs to
+// keep them unchanged before and after assembling. All defined IDs to be added
+// to this code builder will be assigned with a global name through OpName
+// instruction. The name is extracted from the definition instruction.
+//  E.g. adding instruction: %var_a = OpConstant %int 2, will also add an
+//  instruction: OpName %var_a, "var_a".
+//
+// Note that the name must not be used on more than defined IDs and
```
antiagainst: more than **one** defined ... ?

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,246 @@
...
+  // Appends annotation instructions to the annotation section, and returns the
+  // reference of this assembly builder.
+  AssemblyBuilder& AppendAnnotations(
+      const std::vector<std::string>& vec_annotations) {
+    annotations_.insert(annotations_.end(), vec_annotations.begin(),
+                        vec_annotations.end());
+    return *this;
+  }
+
+  // Get the SPIR-V assembly code in string.
```
antiagainst: Get**s** the ... **as** string.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,246 @@
...
+  std::string GetCode() const {
+    std::stringstream ss;
```
antiagainst: `std::ostringstream` suffices.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,246 @@
...
+    for (auto& line : global_preamble_) {
```
antiagainst: Although `auto` will be deduced as `const std::string` here (?), I'd think use `const auto` will be more clear.

test/opt/assembly_builder.h
```diff
@@ -0,0 +1,246 @@
...
+    }
+    for (auto& line : main_func_postamble_) {
+      ss << line << std::endl;
+    }
+    return ss.str();
+  }
+
+ private:
+  // Adds a given name to the Name section with OpName. If the given name has
+  // been added before, do nothing.
```
antiagainst: do**es** nothing.

antiagainst:
Just some nits.

Generally LGTM, +1.

test/opt/assembly_builder.h
Qining: Fixed in d83a6b7

test/opt/assembly_builder.h
Qining: Yes, I should add the `cosnt`, fixed in d83a6b7

test/opt/assembly_builder.h
Qining: Changed to `ostringstream` in d83a6b7

test/opt/assembly_builder.h
Qining: fixed in d83a6b7

test/opt/assembly_builder.h
Qining: Fixed in d83a6b7

test/opt/assembly_builder.h
Qining: Yes, I should make it clearer. The blurb has been updated in d83a6b7.

test/opt/assembly_builder.h
```diff
@@ -41,20 +41,25 @@ namespace spvtools {
...
 //  Types, Constants and Global variables:
-//    declare types, constants and global variables.
+//    Types, constants and global variables declaration instructions.
 //  Main Function:
-//    contains instructions in main function, just after OpLabel for the entry
-//    block.
+//    Main function instructions.
 //  Main Function Postamble:
-//    contains the return and fucntion end instruction.
+//    The return and fucntion end instructions.
```
dneto0: "fucntion" -> "function"

test/opt/assembly_builder.h
```diff
@@ -239,7 +243,7 @@ class AssemblyBuilder {
   std::vector<std::string> main_func_postamble_;
   // All of the defined variable names and the positions of their OpName
```
dneto0: Comment is out of date.  :-(

dneto0:
-1 minor things.

test/opt/assembly_builder.h
Qining: Fixed in 9620e34

test/opt/assembly_builder.h
Qining: Fixed in 9620e34

dneto0:
+2  Please rebase first before merging

Qining:
rebased and merged.

PR #293, opened 2016-08-04
Dead constants elimination

Qining:
A pass to remove dead constants, including both front-end constants and spec
constants.

This pass does not handle dead variables and types.

### DIFF
--- a/source/opt/module.h
+++ b/source/opt/module.h
@@ -94,6 +94,8 @@ class Module {
   // Returns a vector of pointers to type-declaration instructions in this
   // module.
   std::vector<Instruction*> types();
+  // Returns the constant-defining instructions.
+  std::vector<Instruction*> GetConstants();
   const std::vector<Instruction>& debugs() const { return debugs_; }
   std::vector<Instruction>& debugs() { return debugs_; }
   const std::vector<Instruction>& annotations() const { return annotations_; }
--- a/source/opt/passes.cpp
+++ b/source/opt/passes.cpp
@@ -26,6 +26,14 @@
 
 #include "passes.h"
 
+#include <algorithm>
+#include <iostream>
+#include <unordered_map>
+#include <unordered_set>
+
+#include "def_use_manager.h"
+#include "reflect.h"
+
 namespace spvtools {
 namespace opt {
 
@@ -69,6 +77,94 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
     }
   });
   return modified;
+}
+
+bool EliminateDeadConstantPass::Process(ir::Module* module) {
+  analysis::DefUseManager def_use;
+  def_use.AnalyzeDefUse(module);
+  std::unordered_set<ir::Instruction*> working_list;
+  // Traverse all the instructions to get the initial set of dead constants as
+  // working list and count number of real uses for constants. Uses in
+  // annotation instructions do not count.
+  std::unordered_map<ir::Instruction*, size_t> use_counts;
+  std::vector<ir::Instruction*> constants = module->GetConstants();
+  for (auto* c : constants) {
+    uint32_t const_id = c->result_id();
+    size_t count = 0;
+    if (analysis::UseList* uses = def_use.GetUses(const_id)) {
+      count =
+          std::count_if(uses->begin(), uses->end(), [](const analysis::Use& u) {
+            return !(ir::IsAnnotationInst(u.inst->opcode()) ||
+                     ir::IsDebugInst(u.inst->opcode()));
+          });
+    }
+    use_counts[c] = count;
+    if (!count) {
+      working_list.insert(c);
+    }
+  }
+
+  // Start from the constants with 0 uses, back trace through the def-use chain
+  // to find all dead constants.
+  std::unordered_set<ir::Instruction*> dead_consts;
+  while (!working_list.empty()) {
+    ir::Instruction* inst = *working_list.begin();
+    // Back propagate if the instruction contains IDs in its operands.
+    switch (inst->opcode()) {
+      case SpvOp::SpvOpConstantComposite:
+      case SpvOp::SpvOpSpecConstantComposite:
+      case SpvOp::SpvOpSpecConstantOp:
+        for (uint32_t i = 0; i < inst->NumInOperands(); i++) {
+          // SpecConstantOp instruction contains 'opcode' as its operand. Need
+          // to exclude such operands when decreasing uses.
+          if (inst->GetInOperand(i).type != SPV_OPERAND_TYPE_ID) {
+            continue;
+          }
+          uint32_t operand_id = inst->GetSingleWordInOperand(i);
+          ir::Instruction* def_inst = def_use.GetDef(operand_id);
+          // If the use_count does not have any count for the def_inst,
+          // def_inst must not be a constant, and should be ignored here.
+          if (!use_counts.count(def_inst)) {
+            continue;
+          }
+          // The number of uses should never be less then 0, so it can not be
+          // less than 1 before it decreases.
+          assert(use_counts[def_inst] > 0);
+          --use_counts[def_inst];
+          if (!use_counts[def_inst]) {
+            working_list.insert(def_inst);
+          }
+        }
+        break;
+      default:
+        break;
+    }
+    dead_consts.insert(inst);
+    working_list.erase(inst);
+  }
+
+  // Find all annotation and debug instructions that are referencing dead
+  // constants.
+  std::unordered_set<ir::Instruction*> dead_others;
+  for (auto* dc : dead_consts) {
+    if (analysis::UseList* uses = def_use.GetUses(dc->result_id())) {
+      for (const auto& u : *uses) {
+        if (ir::IsAnnotationInst(u.inst->opcode()) ||
+            ir::IsDebugInst(u.inst->opcode())) {
+          dead_others.insert(u.inst);
+        }
+      }
+    }
+  }
+
+  // Turn all dead instructions and uses of them to nop
+  for (auto* dc : dead_consts) {
+    def_use.KillDef(dc->result_id());
+  }
+  for (auto* da : dead_others) {
+    da->ToNop();
+  }
+  return !dead_consts.empty();
 }
 
 }  // namespace opt
--- a/test/opt/test_eliminate_dead_const.cpp
+++ b/test/opt/test_eliminate_dead_const.cpp
@@ -1,0 +1,852 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#include "assembly_builder.h"
+#include "pass_fixture.h"
+#include "pass_utils.h"
+
+#include <algorithm>
+#include <cstdarg>
+#include <iostream>
+#include <sstream>
+#include <unordered_set>
+
+namespace {
+
+using namespace spvtools;
+
+using EliminateDeadConstantBasicTest = PassTest<::testing::Test>;
+
+TEST_F(EliminateDeadConstantBasicTest, BasicAllDeadConstants) {
+  const std::vector<const char*> text = {
+      // clang-format off
+               "OpCapability Shader",
+               "OpCapability Float64",
+          "%1 = OpExtInstImport \"GLSL.std.450\"",
+               "OpMemoryModel Logical GLSL450",
+               "OpEntryPoint Vertex %main \"main\"",
+               "OpName %main \"main\"",
+       "%void = OpTypeVoid",
+          "%4 = OpTypeFunction %void",
+       "%bool = OpTypeBool",
+          "%6 = OpConstantTrue %bool",
+          "%7 = OpConstantFalse %bool",
+        "%int = OpTypeInt 32 1",
+          "%9 = OpConstant %int 1",
+       "%uint = OpTypeInt 32 0",
+         "%11 = OpConstant %uint 2",
+      "%float = OpTypeFloat 32",
+         "%13 = OpConstant %float 3.14",
+     "%double = OpTypeFloat 64",
+         "%15 = OpConstant %double 3.14159265358979",
+       "%main = OpFunction %void None %4",
+         "%16 = OpLabel",
+               "OpReturn",
+               "OpFunctionEnd",
+      // clang-format on
+  };
+  // None of the above constants is ever used, so all of them should be
+  // eliminated.
+  const char* const_decl_opcodes[] = {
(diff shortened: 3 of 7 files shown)


test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,759 @@
...
+    ::testing::ValuesIn(std::vector<EliminateDeadConstantTestCase>({
+        // Tests eliminating dead constant type ivec2. One dead constant vector
+        // and one used constant vector, each built from its own group of
+        // scalar constants.
+        {
+            /* .used_consts = */
+            {
+                "%used_int_x = OpConstant %int 1",
+                "%used_int_y = OpConstant %int 2",
+                "%used_v2int = OpConstantComposite %v2int "
```
antiagainst: Since we use manual format for this kind of list, it would be nice to avoid splitting the same string into multiple lines. I feel it's quite hard to read and differentiate between multiple strings (via `,`) or the same string crossing multiple lines (without `,`).

source/opt/passes.cpp
```diff
@@ -71,5 +78,84 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+  // Traverse all the instructions to get the initial set of dead constants as
+  // working list and count number of real uses for constants. Uses in
+  // annotation instructions do not count.
+  std::unordered_map<ir::Instruction*, size_t> use_counts;
+  std::vector<ir::Instruction*> constants = module->constants();
+  for (auto& c : constants) {
+    uint32_t const_id = c->result_id();
+    size_t count = 0;
+    if (analysis::UseList* uses = def_use.GetUses(const_id)) {
+      for (auto& u : *uses) {
```
antiagainst: `auto*`?

source/opt/passes.cpp
```diff
@@ -71,5 +78,84 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+        if (!ir::IsAnnotationInst(u.inst->opcode()) &&
+            !ir::IsDebugInst(u.inst->opcode())) {
+          count += 1;
```
antiagainst: `++count`?

source/opt/passes.cpp
```diff
@@ -71,5 +78,84 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+bool EliminateDeadConstantPass::Process(ir::Module* module) {
+  analysis::DefUseManager def_use;
+  def_use.AnalyzeDefUse(module);
+  std::unordered_set<ir::Instruction*> working_list;
+  // Traverse all the instructions to get the initial set of dead constants as
+  // working list and count number of real uses for constants. Uses in
+  // annotation instructions do not count.
+  std::unordered_map<ir::Instruction*, size_t> use_counts;
+  std::vector<ir::Instruction*> constants = module->constants();
+  for (auto& c : constants) {
```
antiagainst: `auto*`?

source/opt/passes.cpp
```diff
@@ -71,5 +78,84 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+    uint32_t const_id = c->result_id();
+    size_t count = 0;
+    if (analysis::UseList* uses = def_use.GetUses(const_id)) {
```
antiagainst: I would suggest to use
```
const auto* use_list = def_use.GetUses(...);
std::count(use_list.begin(), use_list.end(),
                 [](const Use& u) { return ... });
```

source/opt/passes.cpp
```diff
@@ -71,5 +78,84 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+    if (!count) {
+      working_list.insert(c);
+    }
+  }
+
+  // Start from the constants with 0 uses, back trace through the def-use chain
+  // to find all dead constants.
+  std::unordered_set<ir::Instruction*> dead_consts;
+  while (!working_list.empty()) {
+    ir::Instruction* inst = *working_list.begin();
```
antiagainst: `working_list.front()`?

source/opt/passes.cpp
```diff
@@ -71,5 +78,84 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+    // Back propagate if the instruction contains IDs in its operands.
+    switch (inst->opcode()) {
+      case SpvOp::SpvOpConstantComposite:
+      case SpvOp::SpvOpSpecConstantComposite:
+      case SpvOp::SpvOpSpecConstantOp:
+        for (uint32_t i = 0; i < inst->NumInOperands(); i++) {
+          uint32_t operand_id = inst->GetSingleWordInOperand(i);
+          ir::Instruction* def_inst = def_use.GetDef(operand_id);
+          use_counts[def_inst] -= 1;
```
antiagainst: `--use_counts[def_inst]`?

antiagainst: Please add an `assert(use_counts[def_inst] > 0)` here to further sanitize and potentially blow up if something is wrong internally.

source/opt/passes.cpp
```diff
@@ -71,5 +78,84 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+        }
+      }
+    }
+  }
+
+  // Turn all dead instructions and uses of them to nop
+  bool modified = false;
+  for (auto& dc : dead_consts) {
+    dc->dbg_line_insts().clear();
+    dc->ToNop();
```
antiagainst: I think we should use `KillDef` instead of `ToNop` here.

source/opt/passes.cpp
```diff
@@ -71,5 +78,84 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+          dead_others.insert(u.inst);
+        }
+      }
+    }
+  }
+
+  // Turn all dead instructions and uses of them to nop
+  bool modified = false;
+  for (auto& dc : dead_consts) {
+    dc->dbg_line_insts().clear();
```
antiagainst: We cannot erase `dbg_line_insts`. Those debug instructions are just physically adjacent (ahead of) to this instruction, but they are not a logical part of this instruction. Turning this instruction to `OpNop` should not affect them. (The weirdness of this issue comes from the flexibility of line-related debug instructions.)

source/opt/passes.cpp
```diff
@@ -71,5 +78,84 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+    dc->ToNop();
+    modified = true;
+  }
+  for (auto& da : dead_others) {
+    da->ToNop();
+  }
+  return modified;
```
antiagainst: `return !dead_consts.empty()`?

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,759 @@
...
+         "%13 = OpConstant %float 3.14",
+     "%double = OpTypeFloat 64",
+         "%15 = OpConstant %double 3.14159265358979",
+       "%main = OpFunction %void None %4",
+         "%16 = OpLabel",
+               "OpReturn",
+               "OpFunctionEnd",
+      // clang-format on
+  };
+  // None constants is ever used, so all of them should be eliminated.
```
antiagainst: `None of the above`

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,759 @@
...
+namespace {
+
+using namespace spvtools;
+
+using EliminateDeadConstantBasicTest = PassTest<::testing::Test>;
+
+TEST_F(EliminateDeadConstantBasicTest, BasicAllDeadConstants) {
+  std::vector<const char*> text = {
+      // clang-format off
+               "OpCapability Shader",
```
antiagainst: Actually these preamble stuff can be deleted since they are not needed in this test (we don't really care about the validation here). But I don't care that much.

source/opt/passes.cpp
antiagainst: Sorry, no such API.

test/opt/test_eliminate_dead_const.cpp
Qining: Fixed in 8e16da0

source/opt/passes.cpp
Qining: This is better, fixed in 8e16da0

source/opt/passes.cpp
Qining: Fixed in 8e16da0

source/opt/passes.cpp
Qining: Changed to use KillDef() in 8e16da0

source/opt/passes.cpp
Qining: Fixed in 8e16da0

source/opt/passes.cpp
Qining: Smart, fixed in 8e16da0

source/opt/module.h
```diff
@@ -94,6 +94,7 @@ class Module {
   // Returns a vector of pointers to type-declaration instructions in this
   // module.
   std::vector<Instruction*> types();
+  std::vector<Instruction*> constants();
```
dneto0: Needs a blurb.  Also, since it's not just a member access (and shouldn't be cached), then this needs a first-letter-capitalized method name like Constants() or GetConstants()

source/opt/passes.cpp
```diff
@@ -71,5 +79,88 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
   return modified;
 }
 
+bool EliminateDeadConstantPass::Process(ir::Module* module) {
```
dneto0: At some point the passes.cpp file will gain too many responsibilities.  At some point the non-triival passses should be moved  / placed into their own files.

antiagainst: Agreed.

Qining: I have that as part of the "-load *.so" CL #271. That definitely one thing needs to do.

In #271, the abstract `pass` class is extracted to a `pass.h` file, and all passes should only needs to include that header.

test/opt/test_eliminate_dead_const.cpp
Qining: Hmm, the assembly builder includes such preamble stuff for now. Probably not necessary to remove the preamble here, but good to know that.

source/opt/module.h
Qining: Fixed in 2d78f1f

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,717 @@
...
+using EliminateDeadConstantTest =
+    PassTest<::testing::TestWithParam<EliminateDeadConstantTestCase>>;
+
+TEST_P(EliminateDeadConstantTest, Custom) {
+  auto& tc = GetParam();
+  AssemblyBuilder builder;
+  builder.AppendTypesConstantsGlobals(CommonTypes)
+      .AppendTypesConstantsGlobals(tc.used_consts)
+      .AppendInMain(tc.main_insts);
+  std::string expected = builder.GetCode();
```
antiagainst: const

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,717 @@
...
+  builder.AppendTypesConstantsGlobals(tc.dead_consts);
+  std::string assembly_with_dead_const = builder.GetCode();
```
antiagainst: const

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,717 @@
...
+            {
+                "%dead_const_double = OpConstant %double 3.14",
+            },
+        },
+        // clang-format on
+    })));
+
+INSTANTIATE_TEST_CASE_P(
+    VectorTypeConstants, EliminateDeadConstantTest,
+    ::testing::ValuesIn(std::vector<EliminateDeadConstantTestCase>({
```
antiagainst: It would be nice to diverse the vector size a little bit in these tests.

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,717 @@
...
+
+INSTANTIATE_TEST_CASE_P(
+    StructTypeConstants, EliminateDeadConstantTest,
+    ::testing::ValuesIn(std::vector<EliminateDeadConstantTestCase>({
+        // clang-format off
+        // A plain struct type dead constants. All of its components are dead
+        // constants too.
+        {
+            /* .used_consts = */
+            {
```
antiagainst: Can save some lines by condensing like the following.

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,717 @@
...
+            },
+            /* .main_insts = */ {},
+            /* .dead_consts = */
+            {
+                "%dead_bool = OpConstantTrue %bool",
+                "%dead_int = OpConstant %int 1",
+                "%dead_float = OpConstant %float 2.5",
+                "%dead_double = OpConstant %double 3.1415926535",
+                "%dead_inner_struct = OpConstantComposite %inner_struct "
```
antiagainst: Please use one line for this `dead_inner_struct`

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,717 @@
...
+            {
+                "%dead_bool = OpConstantTrue %bool",
+                "%dead_float = OpConstant %float 2.5",
+                "%dead_inner_struct = OpConstantComposite %inner_struct %dead_bool %used_int %dead_float %used_double",
+                "%dead_int = OpConstant %int 2",
+                "%dead_double = OpConstant %double 1.428571428514",
+                "%dead_outer_struct = OpConstantComposite %outer_struct %dead_inner_struct %dead_int %dead_double",
+            },
+        },
+        // clang-format on
```
antiagainst: It would be nice to see a test case in which both the inner and outer struct have dead and used elements.

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,717 @@
...
+    })));
+
+INSTANTIATE_TEST_CASE_P(
+    VectorTypeSpecConstants, EliminateDeadConstantTest,
+    ::testing::ValuesIn(std::vector<EliminateDeadConstantTestCase>({
+        // clang-format off
+        // Bool vector type spec constants. One vector has all component dead,
+        // another vector has one dead boolean and one used boolean.
+        {
```
antiagainst: I feel like it may be ok to just do spotcheck instead of enumerating all possible types here. But I don't care that much.

test/opt/test_eliminate_dead_const.cpp
```diff
@@ -0,0 +1,717 @@
...
+                "%b_x = OpSpecConstantOp %int CompositeExtract %vec_b 0",
+
+                // Extract vector
+                "%a_xy = OpSpecConstantOp %v2int VectorShuffle %vec_a %vec_a 1 0",
+                "%b_xy = OpSpecConstantOp %v2int VectorShuffle %vec_b %vec_b 1 0",
+            },
+        },
+        // clang-format on
+    })));
+
```
antiagainst: It would be nice to have another spotcheck test case that has very (well, comparatively) complex references of used and dead constants to increase the confidence level. :)

antiagainst:
Review completed.

Super nice tests! :)

source/opt/passes.cpp
antiagainst: Also need blurb here.

test/opt/test_eliminate_dead_const.cpp
Qining: Added a test with long def-use chain. Nice to find a bug with that long test.

Fixed in f6d3c82

test/opt/test_eliminate_dead_const.cpp
Qining: Fixed in f6d3c82

test/opt/test_eliminate_dead_const.cpp
Qining: Fixed in f6d3c82

test/opt/test_eliminate_dead_const.cpp
Qining: Fixed in f6d3c82

test/opt/test_eliminate_dead_const.cpp
Qining: added vector with size 3 in f6d3c82

test/opt/test_eliminate_dead_const.cpp
Qining: Fixed in f6d3c82

test/opt/test_eliminate_dead_const.cpp
Qining: Fixed in f6d3c82

source/opt/passes.cpp
```diff
@@ -71,5 +79,93 @@ bool FreezeSpecConstantValuePass::Process(ir::Module* module) {
...
+      }
+    }
+  }
+
+  // Turn all dead instructions and uses of them to nop
+  for (auto* dc : dead_consts) {
+    def_use.KillDef(dc->result_id());
+  }
+  for (auto* da : dead_others) {
+    da->ToNop();
```
antiagainst: Would prefer to use `KillDef` here, too.

Besides, I don't see tests exercising this code snippet? Would be nice to add some tests containing debug/annotation instructions.

Qining: Cannot use KillDef as those instructions do not have result ID. Tests are comming.

Qining: Actually this code snippet has been exercised. The assembly code builder adds `OpDecorate SpecId` instruction for all the primary spec constants, and each defined constant has an `OpName` instruction. Those instructions must be removed after the pass so that the result can match with the expected result, in which we don't add those constants.

antiagainst: OK, I see.

antiagainst:
LGTM

Qining:
@dneto0 This CL is ready for review. PTAL

Issue #295, opened 2016-08-04
validator: SSA def-use check should only check reachable uses


dneto0:
Example:
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
       %void = OpTypeVoid
          %2 = OpTypeFunction %void
       %bool = OpTypeBool
          %4 = OpConstantTrue %bool
          %5 = OpFunction %void None %2
          %6 = OpLabel
          %7 = OpCopyObject %bool %4
               OpReturn
          %8 = OpLabel
          %9 = OpCopyObject %bool %7
               OpReturn
               OpFunctionEnd
```
Today the validator says:
error: 17: ID 3[def] defined in block 1[entry] does not dominate its use in block 2[unreach]

PR #296, opened 2016-08-04
Only check def dominance of reachable uses

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/295


umar456:
LGTM +2

PR #298, opened 2016-08-04
update_build_version.py: create directory first if not existing and always refresh.

antiagainst:
Also with a face lifting commit.

### DIFF
--- a/utils/update_build_version.py
+++ b/utils/update_build_version.py
@@ -25,7 +25,8 @@
 #  - A longer string with the project name, the software version number, and
 #    git commit information for the directory.  The commit information
 #    is the output of "git describe" if that succeeds, or "git rev-parse HEAD"
-#    if that succeeds, or otherwise a message containing the phrase "unknown hash".
+#    if that succeeds, or otherwise a message containing the phrase
+#    "unknown hash".
 # The string contents are escaped as necessary.
 
 from __future__ import print_function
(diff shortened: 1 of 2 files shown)


utils/update_build_version.py
```diff
@@ -71,41 +72,44 @@ def deduce_software_version(dir):
...
+        os.makedirs(output_dir)
 
     software_version = deduce_software_version(sys.argv[1])
     new_content = '"{}", "SPIRV-Tools {} {}"\n'.format(
         software_version, software_version,
         describe(sys.argv[1]).replace('"', '\\"'))
-    if os.path.isfile(output_file) and new_content == open(output_file, 'r').read():
-        sys.exit(0)
-    open(output_file, 'w').write(new_content)
+
```
dneto0: This new behaviour will cause constant rebuilding.  Please keep the old logic here.

utils/update_build_version.py
```diff
@@ -71,41 +72,44 @@ def deduce_software_version(dir):
...
 
 
 def main():
     if len(sys.argv) != 3:
-        print('usage: {0} <spirv-tools_dir> <output-file>'.format(sys.argv[0]))
+        print('usage: {} <spirv-tools-dir> <output-file>'.format(sys.argv[0]))
         sys.exit(1)
 
     output_file = sys.argv[2]
+    output_dir = os.path.dirname(output_file)
```
dneto0: This make sense.

dneto0:
-1 Need to keep the old behaviour of "write only if contents are different"

utils/update_build_version.py
antiagainst: Done.

dneto0:
+2

Issue #297, opened 2016-08-04
Validator: incorrectly errors out when an unreachable block branches to a merge block


dneto0:
I'm sure this is my fault for adding pseudo-entry nodes.

Example input that is actually valid:
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
       %bool = OpTypeBool
          %7 = OpConstantTrue %bool
       %main = OpFunction %void None %3
          %5 = OpLabel
               OpSelectionMerge %9 None
               OpBranchConditional %7 %8 %9
          %8 = OpLabel
               OpSelectionMerge %11 None
               OpBranchConditional %7 %10 %13
         %10 = OpLabel
               OpReturn
         %13 = OpLabel
               OpReturn
         %11 = OpLabel
               OpBranch %9
          %9 = OpLabel
               OpReturn
               OpFunctionEnd
```
The validator generates error:
>error: 28: The selection construct with the selection header 5 does not dominate the merge block 9

The problem is that the pseudo-entry node has an arc into unreachable block 11, and 11 branches to 9, and 9 is a merge block for selection header block 5.  But since there's an alternate path from the *pseudo-entry* node (0) into 9, then 5 does not dominate 9.

dneto0:
![y](https://cloud.githubusercontent.com/assets/13054574/17417649/6b7a3018-5a62-11e6-9d85-08934bb00743.png)

dneto0:
The above shows the CFG (without psuedo entry or pseudo exit nodes).  The dashed lines are arcs from a header node to their merge blocks.  Solid arcs are branches.

dneto0:
Incidentally, the original shader that gave rise to the SPIR-V assembly is:
```
#version 310 es

void main() {
  if (true) {
    if (true)
      return;
    else
      return;
  }
}
```
That reflects the essence of some Vulkan CTS examples.

PR #300, opened 2016-08-05
Fix dominance calculation

dneto0:
Fixes dominance calculation when there is a forward arc from an
unreachable block A to a reachable block B.  Before this fix, we would
say that B is not dominated by the graph entry node, and instead say
that the immediate dominator of B is the psuedo-entry node of the
augmented CFG.

The fix:

- Dominance is defined in terms of a traversal from the entry block
  of the CFG.  So the forward DFS should start from the function
  entry block, not the pseudo-entry-block.

- When following edges backward during dominance calculations, only go to
  nodes that are actually reachable in the forward traversal.
  Important: the sense of reachability flips around when computing
  post-dominance.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/297

### DIFF
--- a/test/Validate.CFG.cpp
+++ b/test/Validate.CFG.cpp
@@ -1149,6 +1149,41 @@ TEST_F(ValidateCFG, LoopWithoutBackEdgesBad) {
                            "back-edges but the standard requires exactly one"));
 }
 
+TEST_P(ValidateCFG,
+       NestedConstructWithUnreachableMergeBlockBranchingToOuterMergeBlock) {
+  // Test for https://github.com/KhronosGroup/SPIRV-Tools/issues/297
+  // The nested construct has an unreachable merge block.  In the
+  // augmented CFG that merge block
+  // we still determine that the
+  bool is_shader = GetParam() == SpvCapabilityShader;
+  Block entry("entry", SpvOpBranchConditional);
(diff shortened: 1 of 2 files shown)


dneto0:
Rebased

dneto0:
I'm interested in feedback from @umar456

test/Validate.CFG.cpp
```diff
@@ -1149,6 +1149,41 @@ TEST_F(ValidateCFG, LoopWithoutBackEdgesBad) {
...
+  Block inner_merge("inner_merge");
+  Block exit("exit", SpvOpReturn);
+
+  entry.SetBody("%cond    = OpSLessThan %intt %one %two\n");
+  if (is_shader) {
+    entry.AppendBody("OpSelectionMerge %exit None\n");
+    inner_head.SetBody("OpSelectionMerge %inner_merge None\n");
+  }
+
+  string str = header(GetParam()) + nameOps("entry", "inner_merge", "exit") +
```
umar456: You don't need named Ids if you expect the module to validate successfully. I only add those when I need to validate the error message. Not a big deal.

dneto0: Ack.  It's also helpful to make sure Ids have numbers 1,2,3....
Good for debugging.  :-)

umar456:
+2

PR #301, opened 2016-08-05
Check configuration and gmock existence when adding unit tests.

antiagainst:


test/CMakeLists.txt
```diff
@@ -30,118 +30,121 @@
...
       "To enable tests place googletest in '<spirv-dir>/external/googletest'.")
   endif()
 endif()
+
+function(add_spvtools_unittest)
+  if (NOT "${SPIRV_SKIP_TESTS}" AND TARGET gmock_main)
+    set(one_value_args TARGET)
+    set(multi_value_args SRCS LIBS)
+    cmake_parse_arguments(
+      ARG "" "${one_value_args}" "${multi_value_args}" ${ARGN})
```
AWoloszyn: This might be a bit more readable as 
cmake_parse_arguments(
  ARG
 ""
 "TARGET"
 "SRCS;LIBS"
 ${ARG})

AWoloszyn:
One small nit, but +2 otherwise.

test/CMakeLists.txt
antiagainst: Hmm, I feel that `{one|multi}_value_args` is more readable for me. Otherwise, it seems we need to reference the doc for `cmake_parse_arguments` to know what does each parameter mean.

PR #302, opened 2016-08-05
Rename a variable so it's consistent with spec

dneto0:
In the spec "continue block" is a block with a branch
to the Continue Target.  It's not the Continue Target.


dneto0:
A quick one for @umar456 to review.  :-)

umar456:
ha +2

dneto0:
Rebased and pushed into master as bb4c0b1af6

PR #303, opened 2016-08-05
Add BasicBlock methods: dominates postdominates

dneto0:


dneto0:
In coding intermediate solutions for other tasks, I was using this as a much more readable alternative to the std::find recipe I've seen a few places.  It also handles reflexivity as well ( a block always dominates itself).

source/val/BasicBlock.h
```diff
@@ -31,6 +31,7 @@
 
 #include <cstdint>
 
+#include <algorithm>
```
umar456: I like to add non-trivial functions in the `cpp` files to avoid bringing in large headers. hmm, it seems that `algorithm` is being included in `validate.h` so this will not really hurt compile times. Your call.

umar456:
LGTM +2

dneto0:
Rebased and pushed into master

PR #305, opened 2016-08-05
Validator cfg fixes

dneto0:
- Find unreachable continue targets.  Look for back edges
  with a DFS traversal separate from the dominance traversals,
  where we count the OpLoopMerge from the header to the continue
  target as an edge in the graph.

- It's ok for a loop to have multiple back edges, provided
  they are all from the same block, and we call that the latch block.
  This may require a clarification/fix in the SPIR-V spec.

- Compute postdominance correctly for infinite loop:
  Bias *predecessor* traversal root finding so that you use
  a later block in the original list.  This ensures that
  for certain simple infinite loops in the CFG where neither
  block branches to a node without successors, that we'll
  compute the loop header as dominating the latch block, and the
  latch block as postdominating the loop header.

### DIFF
--- a/source/val/Function.cpp
+++ b/source/val/Function.cpp
@@ -50,7 +50,9 @@ using libspirv::BasicBlock;
 
 // Computes a minimal set of root nodes required to traverse, in the forward
 // direction, the CFG represented by the given vector of blocks, and successor
-// and predecessor functions.
+// and predecessor functions.  When considering adding two nodes, each having
+// predecessors, favour using the one that appears earlier on the input blocks
+// list.
 std::vector<BasicBlock*> TraversalRoots(const std::vector<BasicBlock*>& blocks,
                                         libspirv::get_blocks_func succ_func,
                                         libspirv::get_blocks_func pred_func) {
@@ -200,7 +202,6 @@ void Function::RegisterBlockEnd(vector<uint32_t> next_list,
   assert(
       current_block_ &&
       "RegisterBlockEnd can only be called when parsing a binary in a block");
-
   vector<BasicBlock*> next_blocks;
   next_blocks.reserve(next_list.size());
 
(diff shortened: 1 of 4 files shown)


source/val/Function.cpp
```diff
@@ -292,6 +308,16 @@ Function::GetBlocksFunction Function::AugmentedCFGSuccessorsFunction() const {
   };
 }
 
+Function::GetBlocksFunction
+Function::AugmentedCFGSuccessorsFunctionIncludingHeaderToContinueEdge() const {
+  return [this](const BasicBlock* block) {
+    auto where = loop_header_successors_plus_continue_target_map_.find(block);
+    return where == loop_header_successors_plus_continue_target_map_.end()
+               ? AugmentedCFGSuccessorsFunction()(block)
```
dneto0: This could be faster.  We're creating a new lambda in that inner method.

source/val/Function.cpp
```diff
@@ -215,6 +216,21 @@ void Function::RegisterBlockEnd(vector<uint32_t> next_list,
...
+  if (current_block_->is_type(kBlockTypeLoop)) {
+    // For each loop header, record the set of its successors, and include
+    // its continue target if the continue target is not the loop header
+    // itself.
+    std::vector<BasicBlock*>& next_blocks_plus_continue_target =
+        loop_header_successors_plus_continue_target_map_[current_block_];
+    next_blocks_plus_continue_target = next_blocks;
+    // If this block is marked as Loop-type,  then the continue construct is
+    // the most recently created CFG construct.
+    auto continue_target = cfg_constructs_.back().entry_block();
```
umar456: Calling the [corresponding_construction](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/val/Construct.h#L119) method is a better way to get the continue construct of a loop block.

dneto0: Right.  But at this point I haven't got the Construct for this header block.  And Function doesn't have a map where I can look that up quickly.  I agree what I've got is a bit dodgy since it imposes a coupling between the implementations of the two methods. 

I plan to add a map in Function to find the construct from its entry block.

source/val/Function.h
```diff
@@ -262,6 +265,12 @@ class Function {
   std::unordered_map<const BasicBlock*, std::vector<BasicBlock*>>
       augmented_predecessors_map_;
 
+  // Maps a structured loop header to its CFG successors and also its
+  // continue target if that continue target is not the loop header
+  // itself. This might have duplicates.
+  std::unordered_map<const BasicBlock*, std::vector<BasicBlock*>>
```
umar456: This variable seems to be more aligned with a construct object then a function. Hmm, that would make things tricky. This is fine for now.

source/val/Function.cpp
dneto0: Fixed in 605847f15bc

umar456:
LGTM

PR #308, opened 2016-08-06
Remove superfluous headers

umar456:


source/validate.h
```diff
@@ -27,30 +27,18 @@
...
-#include "spirv_definition.h"
 #include "table.h"
-#include "val/BasicBlock.h"
-
-// Structures
 
 namespace libspirv {
 
 class ValidationState_t;
+class BasicBlock;
```
dneto0: I know forward declarations of classes are common practice.
However, the Google style guide strongly discourages them since they *can* hinder adjusting the API that exports the declared class.
https://google.github.io/styleguide/cppguide.html#Forward_Declarations.

Since BasicBlock is entirely contained in this project, I'll allow this.  In future, let's just #include the defining header.

dneto0:
+2
Rebased and pushed into master as fd965c9e7e

PR #309, opened 2016-08-07
Refactor Id class into Instruction class

umar456:
Refactor the Id class into the instruction class. This new class can represent
instructions with and without `result_id`

This approach is useful when tracking the definition dominance over use of Ids.
The previous approach was only tracking the basic block in which the Id was
used whereas the new approach tracks the actual instruction as well as its index.


source/val/Instruction.h
```diff
@@ -0,0 +1,131 @@
...
+  /// Returns a vector of pairs of all references to this instruction's result
+  /// id. The first element is the instruction in which this result id was
+  /// referenced and the second is the index of the word in that instruction
+  /// where this result id appeared
+  const std::vector<std::pair<const Instruction*, uint32_t>>& uses() const {
+    return uses_;
+  }
+
+  /// The word used to define the Instruction
+  uint32_t words(size_t index) const { return words_[index]; }
```
dneto0: Quibble:  I'd have called this "word" instead of "words" since it only returns a single word.
Similar below with "operand(size_t)" below.
But not serious enough to block the pull request.

source/val/Instruction.h
```diff
@@ -0,0 +1,131 @@
...
+  /// The words used to define the Instruction
+  const std::vector<uint32_t>& words() const { return words_; }
+
+  /// The operand of the Instruction at \p index
+  const spv_parsed_operand_t& operands(size_t index) const {
+    return operands_[index];
+  }
+
+  /// The operands of the Instruction
+  const std::vector<spv_parsed_operand_t>& operands() const {
```
dneto0: I imagine a future CL would make an Operand class.  But that can wait, clearly.

dneto0: It might be worth pointing out that operands can be wider than a word.  So that words(i) and operands(i) don't necessarily line up.   Some people miss that.
(Not a blocking issue.)

source/val/ValidationState.cpp
```diff
@@ -251,20 +253,25 @@ vector<uint32_t> ValidationState_t::UnresolvedForwardIds() const {
...
     /// We are in a const function, so we cannot use defs.operator[]().
     /// Luckily we know the key exists, so defs_.at() won't throw an
     /// exception.
-    return &all_definitions_.at(id);
+    return all_definitions_.at(id);
   }
 }
 
+Instruction* ValidationState_t::FindDef(uint32_t id) {
+  return const_cast<Instruction*>(
```
dneto0: The const casts are smelly. 

The other FindDef is already a const function, so why do you need the `const_cast<const ValidationState_t*>` ?

dneto0: Frankly, I'd rather reimplement the body than add const casts.

source/val/ValidationState.h
```diff
@@ -160,17 +159,24 @@ class ValidationState_t {
...
+  /// nullptr
+  const Instruction* FindDef(uint32_t id) const;
 
   /// Finds id's def, if it exists.  If found, returns the definition otherwise
   /// nullptr
-  const Id* FindDef(uint32_t id) const;
+  Instruction* FindDef(uint32_t id);
 
-  const std::unordered_map<uint32_t, Id>& all_definitions() const {
+  /// Returns a vector of instructions in the order they appear in the binary
```
dneto0: It returns a list, not a vector.

source/val/ValidationState.h
```diff
@@ -197,7 +203,11 @@ class ValidationState_t {
   spv_capability_mask_t
       module_capabilities_;  /// Module's declared capabilities.
 
-  std::unordered_map<uint32_t, Id> all_definitions_;
+  /// List of all instructions in the order they appear in the binary
+  std::list<Instruction> ordered_instructions_;
```
dneto0: Please use std::deque as it is better than std::list for this use case:  More densely packed data (fewer pointers), more cache friendly, and still has stable internal references since you only use emplace_back to add to it.  Also, emplace_back and back are const time.

source/val/Instruction.h
dneto0: Hm.  Looks like you didn't actually use this.

source/validate_id.cpp
```diff
@@ -2377,6 +2382,29 @@ function<bool(unsigned)> getCanBeForwardDeclaredFunction(SpvOp opcode) {
 
 namespace libspirv {
 
+spv_result_t UpdateIdUse(ValidationState_t& _) {
+  for (const auto& inst : _.ordered_instructions()) {
+    for (auto& operand : inst.operands()) {
+      const spv_operand_type_t& type = operand.type;
+      const uint32_t operand_id = inst.words()[operand.offset];
```
dneto0: This was a candidate for Instruction::words(size_t)

source/validate_id.cpp
```diff
@@ -2377,6 +2382,29 @@ function<bool(unsigned)> getCanBeForwardDeclaredFunction(SpvOp opcode) {
...
+
+      switch (type) {
+        case SPV_OPERAND_TYPE_VARIABLE_ID:
```
dneto0: Please use spvIsIdType(type) from operand.h instead.
There are already a few places where we see the same 5 type ids in a row, and that's a bit painful, but they're all in switch statements with other cases that need to be addressed.

source/validate_id.cpp
```diff
@@ -2387,32 +2415,38 @@ namespace libspirv {
...
-            return _.diag(SPV_ERROR_INVALID_ID)
-                   << "ID " << _.getIdName(definition.first)
-                   << " defined in block " << _.getIdName(block->id())
-                   << " does not dominate its use in block "
-                   << _.getIdName(use->id());
+        for (auto& use_index_pair : definition.second->uses()) {
+          const Instruction* use = use_index_pair.first;
+          if (const BasicBlock* use_block = use->block()) {
+            if (!use_block->reachable()) continue;
+            if (use->opcode() != SpvOpPhi &&
```
dneto0: The old OpPhi special case still checked the def-use of the result type.   :-(
But I guess I didn't have a test case for that?

source/val/Instruction.h
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /// The word used to define the Instruction
+  uint32_t words(size_t index) const { return words_[index]; }
+
+  /// The words used to define the Instruction
+  const std::vector<uint32_t>& words() const { return words_; }
+
+  /// The operand of the Instruction at \p index
+  const spv_parsed_operand_t& operands(size_t index) const {
```
dneto0: This appears to be dead code. Please remove it for now.

dneto0:
Somewhat minor issues:
 - minor cleanups
 - dead methods
 - missing checking def-use of OpPhi type operand. (My bad not having a test for that.)

dneto0:
(Note to myself: Now the validator has a first class Instruction class, and so does the optimizer.  Eventually it would be nice to have a common one.  But that's on me.)

source/val/Instruction.h
umar456: Wouldn't this function take that into account? Or do you mean to add this as a comment?

source/validate_id.cpp
umar456: Good idea. I am skipping SPV_OPERAND_TYPE_RESULT_ID in this switch but I can add a continue condition for that one.

source/validate_id.cpp
umar456: It is currently disabled but I have a PR ready as soon as this goes in.

source/val/ValidationState.cpp
umar456: The const version of the  `FindDef` function is returning a const `Instruction` pointer. Because I need to track the usage of these instructions, I need to be able to modify the Instruction object. This function returns a non-const `Instruction` pointer which can be used for that purpose and since I didn't want to repeat code I took this approach(Item 3 of Effective C++).

Another approach would be to make the `uses_` object mutable or repeat code as you said. I thought this was the safest of the 3 options. I don't have a particularly strong opinion on the matter. I can remove the consts.

source/val/Instruction.h
umar456: Fixed in 1c17c61

source/validate_id.cpp
umar456: Fixed in 1c17c61

source/validate_id.cpp
umar456: Fixed in 1c17c61

source/val/ValidationState.h
umar456: Good idea. I always forget about deque.

Fixed in 1c17c61

source/val/ValidationState.h
umar456: Fixed in 1c17c61

source/val/ValidationState.cpp
umar456: Fixed in 1c17c61

source/val/Instruction.h
dneto0: I meant to add that as a comment.  (But not a blocking issue)

source/val/ValidationState.cpp
dneto0: Good honest programmers can disagree.  If needed, point a complainer straight back at me.  :-)
(But, I guess I need brushing up on Effective C++...)

dneto0:
+2

dneto0:
Rebased and pushed into master as 816f29805b

umar456:
It seems that the fixes that I submitted(1c17c61) were not merged in the push to master.

dneto0:
My mistake.  I cherry-picked it and pushed onto master as 8ea1268f849

PR #310, opened 2016-08-08
Use NEW behavior for policy CMP0054.

antiagainst:
The NEW behavior is to not dereference variables or interpret keywords
that have been quoted or bracketed.

For more information, see
https://cmake.org/cmake/help/v3.1/policy/CMP0054.html.

This is to suppress a warning when using CMake 3.1.3+.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,13 @@
 # MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
 cmake_minimum_required(VERSION 2.8.12)
+if (POLICY CMP0054)
+  # Avoid dereferencing variables or interpret keywords that have been
+  # quoted or bracketed.
+  # https://cmake.org/cmake/help/v3.1/policy/CMP0054.html
+  cmake_policy(SET CMP0054 NEW)
+endif()
+
 project(spirv-tools)
 enable_testing()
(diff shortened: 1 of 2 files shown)


antiagainst:
First time developing on a Windows machine. Yay!

CMakeLists.txt
```diff
@@ -25,6 +25,8 @@
 # MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
 cmake_minimum_required(VERSION 2.8.12)
+cmake_policy(SET CMP0054 NEW)
```
umar456: This should be wrapped around  with`if(POLICY CMP0054)` since the policy was introduced in 3.1 and the minimum required cmake version for this file is 2.8.12.

antiagainst: Good catch! Somehow the cmake on my Ubuntu 14.04 got linked to CMake 3.5. So no warnings surfaced when testing.

dneto0:
+1

umar456:
+1

dneto0:
I'll test on Windows and Linux.  CMake versions TBD.

dneto0:
+2
Tested on Linux with Cmake 2.8.12.2
Tested on Windows with Cmake 3.2.3

dneto0:
Rebased and pushed into master as  2fb502b1d

PR #312, opened 2016-08-08
Print the shader when module building failed

Qining:
Just print the shader text that failed in either assembling or disassembling during optimization tests.


antiagainst:
+2

PR #314, opened 2016-08-09
Fix minor mistakes in the tests of dead-const-elim

Qining:


test/opt/test_eliminate_dead_const.cpp
```diff
@@ -746,8 +746,8 @@ INSTANTIATE_TEST_CASE_P(
                 "%b_x = OpSpecConstantOp %int CompositeExtract %vec_b 0",
 
                 // Extract vector
-                "%a_xy = OpSpecConstantOp %v2int VectorShuffle %vec_a %vec_a 1 0",
```
antiagainst: So essentially this is just because we name them as `a_xy` and `b_xy`?

antiagainst:
+2

test/opt/test_eliminate_dead_const.cpp
Qining: Yes, just to make it consistent here.

antiagainst:
LGTM. please squash before submtting.

PR #315, opened 2016-08-09
Separate pass.h from passes.h

Qining:
Future pass implementations should only need to include pass.h

This is also necessary for the implementation of pass registry to support -load=*.so in future.


antiagainst:
+2

PR #316, opened 2016-08-09
Check dominance of OpPhi parent and variable pairs

### LINKED ISSUE #288: Validator: Should check def-use dominance rules for OpPhi (variable,parent) operands
See example in disabled (proposed) test DISABLED_PhiVariableDefMustComeFromBlockDominatingThePredecessorBad

```
%func        = OpFunction %voidt None %vfunct
%entry       = OpLabel
               OpBranchConditional %false %if_true %if_false

 %if_true     = OpLabel
%true_copy   = OpCopyObject %boolt %false
                OpBranch %exit
 
%if_false    = OpLabel
%false_copy  = OpCopyObject %boolt %false
               OpBranch %exit

; The (variable,Id) pairs are swapped.
%exit        = OpLabel
%value       = OpPhi %boolt %true_copy %if_false %false_copy %if_true
               OpReturn
               OpFunctionEnd
```

We should be able to check that for a particular (variable,parent) pair of an OpPhi, that %variable is defined in a block that dominates %parent.  That check should generate an error for the above case.

umar456:
Checks the dominance of the variable definitions in the OpPhi instruction with their corresponding parent blocks.

Fixes #288

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -2414,6 +2414,7 @@ spv_result_t UpdateIdUse(ValidationState_t& _) {
 /// NOTE: This function does NOT check module scoped functions which are
 /// checked during the initial binary parse in the IdPass below
 spv_result_t CheckIdDefinitionDominateUse(const ValidationState_t& _) {
+  unordered_set<const Instruction*> phi_instructions;
   for (const auto& definition : _.all_definitions()) {
     // Check only those definitions defined in a function
     if (const Function* func = definition.second->function()) {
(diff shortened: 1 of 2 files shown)


dneto0:
+2
This still misses checking the use of the result type ID for the OpPhi.  That was the regression in the previous CL. I'll add the test and the check in another CL.

dneto0:
I'll patch it for the words(i) -> word(i) name change that I forgot to push in earlier.

dneto0:
Rebased, slightly patched, and pushed into master as 8ea1268f849

dneto0:
I said:
>This still misses checking the use of the result type ID for the OpPhi. That was the regression in the previous CL. I'll add the test and the check in another CL.

But now I don't think that can be caught in this part of the code.  The result type is a type id, and those are defined outside of all basic blocks.  So this case is already covered by a few more general things:
- that IDs are defined before use, when that definition occurs outside of a basic block, and
- the logical layout constraints: type declarations must come before function bodies
- the binary parser has already checked that that Id must be a type ID (if my memory serves correctly)

So I think it's all a-ok.

PR #317, opened 2016-08-09
Manually create move the constructor/assignment for Instruction.

antiagainst:
This is because some old visual studio versions (e.g., 2013) do
not support automatically generating move constructors/assignments.


dneto0:
+2

PR #321, opened 2016-08-10
Create an iterator class for in-memory representation.

antiagainst:
This PR depends on #322.

The third commit is interesting here.

### DIFF
--- a/source/opt/iterator.h
+++ b/source/opt/iterator.h
@@ -1,0 +1,171 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#ifndef LIBSPIRV_OPT_ITERATOR_H_
+#define LIBSPIRV_OPT_ITERATOR_H_
+
+#include <memory>
+#include <type_traits>
+#include <vector>
+
+namespace spvtools {
+namespace ir {
+
+// An ad hoc iterator class for std::vector<std::unique_ptr<|ValueType|>>. The
+// purpose of this iterator class is to provide transparent access to those
+// std::unique_ptr managed elements in the vector, behaving like we are using
+// std::vector<|ValueType|>.
+template <typename ValueType, bool IsConst = false>
+class UptrVectorIterator {
+ public:
+  using pointer =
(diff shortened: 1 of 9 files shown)


source/opt/iterator.h
```diff
@@ -0,0 +1,153 @@
...
+  // invalidated. Otherwise, those previous iterators pointing to after the
+  // insertion point will be invalidated.
+  inline UptrVectorIterator InsertBefore(Uptr value);
+
+ private:
+  UptrVector* container_;        // The container we are manipulating.
+  UnderlyingIterator iterator_;  // The raw iterator from the container.
+};
+
+// Handly class for a (begin, end) iterator pair.
```
dneto0: "Handy"

source/opt/iterator.h
```diff
@@ -0,0 +1,153 @@
...
+template <typename IteratorType>
+class IteratorRange {
+ public:
+  IteratorRange(IteratorType b, IteratorType e) : begin_(b), end_(e) {}
+
+  IteratorType begin() const { return begin_; }
+  IteratorType end() const { return end_; }
+
+  bool empty() const { return begin_ == end_; }
+  size_t size() const { return end_ = begin_; }
```
dneto0: end_ - begin_

antiagainst: Err... I need to write test for it.

source/opt/passes.cpp
```diff
@@ -39,7 +39,7 @@ namespace opt {
 
 bool StripDebugInfoPass::Process(ir::Module* module) {
   bool modified = !module->debugs().empty();
-  module->debugs().clear();
+  for (auto& inst : module->debugs()) inst.ToNop();
```
dneto0: Shouldn't we also do module->debugs().clear() ?

source/opt/iterator.h
```diff
@@ -0,0 +1,153 @@
...
+
+ private:
+  IteratorType begin_;
+  IteratorType end_;
+};
+
+template <typename VT, bool IC>
```
Qining: VT => ValueType, IC => IsConst to keep it consistent?

source/opt/iterator.h
```diff
@@ -0,0 +1,153 @@
...
+inline UptrVectorIterator<VT, IC> UptrVectorIterator<VT, IC>::operator++() {
```
Qining: Should we return a reference here?

source/opt/module.h
```diff
@@ -176,6 +186,44 @@ inline void Module::AddFunction(std::unique_ptr<Function> f) {
...
+}
+
+inline IteratorRange<Module::inst_iterator> Module::debugs() {
+  return IteratorRange<inst_iterator>(inst_iterator(&debugs_, debugs_.begin()),
+                                      inst_iterator(&debugs_, debugs_.end()));
+}
+
+inline IteratorRange<Module::const_inst_iterator> Module::debugs() const {
+  return IteratorRange<const_inst_iterator>(
+      const_inst_iterator(&debugs_, debugs_.begin()),
```
Qining: cbegin() instead of begin()?

source/opt/iterator.h
antiagainst: Done.

source/opt/iterator.h
antiagainst: Done. also added tests.

source/opt/passes.cpp
antiagainst: No need anymore. They are all turned into Nop, which will be filtered at the end.

source/opt/iterator.h
antiagainst: I feel like this is more concise since we are not far away from the above full declaration.

source/opt/iterator.h
antiagainst: Good catch! Done.

source/opt/module.h
antiagainst: Right. Done.

source/opt/passes.cpp
dneto0: Still need to clear the vector.

test/opt/test_iterator.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include "gmock/gmock.h"
+
+#include "opt/iterator.h"
+
+namespace {
+
+using namespace spvtools;
+using ::testing::ContainerEq;
+
+TEST(Iterator, IncrementDecrementDeref) {
```
dneto0: There's a lot of functionality tested here.  Idealy they would be smaller scale...
But this is pretty good.

test/opt/test_iterator.cpp
```diff
@@ -0,0 +1,133 @@
...
+  EXPECT_THAT(actual, ContainerEq(expected));
+
+  actual.clear();
+  for (int i = 0; i < count; ++i) {
+    actual.push_back(*(--end));
+  }
+  EXPECT_EQ(ir::UptrVectorIterator<int>(&data, data.begin()), end);
+  std::reverse(actual.begin(), actual.end());
+  EXPECT_THAT(actual, ContainerEq(expected));
+}
```
dneto0: Add TODO: test pre-increment and post-decrement.

dneto0:
-1 
Still think the StripDebugPass should clear the debug insts collection.
Also, please add a TODO to test the other increment modes on the iterator.

test/opt/test_iterator.cpp
antiagainst: Splitted into multiple smaller ones.

test/opt/test_iterator.cpp
antiagainst: Done. ;)

antiagainst:
> Still think the StripDebugPass should clear the debug insts collection.

That works too. Changed to use `clear`.

dneto0:
+2

PR #322, opened 2016-08-10
Change the interfaces of in-memory representation to use pointers.

antiagainst:
Previously we use vectors of objects and move semantics to handle
ownership. That approach has the flaw that inserting an object into
the middle of a vector, which may trigger a vector reallocation,
can invalidate some addresses taken from instructions.

Now the in-memory representation internally uses vector of unique
pointers to handle ownership. Since objects are explicitly heap-
allocated now, pointers to them won't be invalidated by vector
resizing anymore.

A follow up PR #321 introduces an iterator class.

### DIFF
--- a/source/opt/function.h
+++ b/source/opt/function.h
@@ -28,8 +28,9 @@
 #define LIBSPIRV_OPT_CONSTRUCTS_H_
 
 #include <functional>
-#include <vector>
+#include <memory>
 #include <utility>
+#include <vector>
 
 #include "basic_block.h"
 #include "instruction.h"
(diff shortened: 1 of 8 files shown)


source/opt/function.h
```diff
@@ -65,13 +69,26 @@ class Function {
...
 
  private:
-  Module* module_;        // The enclosing module.
-  Instruction def_inst_;  // The instruction definining this function.
-  std::vector<Instruction> params_;  // All parameters to this function.
-  std::vector<BasicBlock> blocks_;   // All basic blocks inside this function.
-  Instruction end_inst_;             // The OpFunctionEnd instruction.
+  // The enclosing module.
+  Module* module_;
+  // The instruction definining this function.
```
dneto0: Minor: "The OpFunction instruction that begins the definition of this function."
(not blocking)

dneto0:
+2
My only comment was a possible improvement to a comment.  Not blocking.

source/opt/function.h
antiagainst: Fixed.

PR #325, opened 2016-08-10
Let Module delegate DefUseManager

Qining:
Embed the DefUseManager to the optimization Module. Now user can call
GetUses, GetDef, KillInst (new added), KillDef from the Module directly.

### DIFF
--- a/source/opt/def_use_manager.cpp
+++ b/source/opt/def_use_manager.cpp
@@ -37,11 +37,13 @@ namespace opt {
 namespace analysis {
 
 void DefUseManager::AnalyzeDefUse(ir::Module* module) {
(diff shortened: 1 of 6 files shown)


dneto0:
-1 doesn't this make the code less modular?

Qining:
But when people create/insert new instructions or remove instruction, it is intuitive to go to Module to find the API, instead of the def-use manager, as the def-use manager does not own the instructions. 

And Def-Use info related to the changes should also be updated automatically. This needs the module to access the def-use info.

Making def-use manager a member of module solves both the problem.

Qining:
Close as this is not beneficial.

PR #327, opened 2016-08-10
Fix build error about passing initializer-list to constructor.

antiagainst:
VS2013 is not happy with using initializer-list to initialize
vector parameters to explicit constructors taking one vector.

Sorry for this. The previous PR passed all checks, but that's
before I submit the PR to turn on VS2013 buildbots.


dneto0:
+2

PR #328, opened 2016-08-11
Use friendly names for Ids

dneto0:


dneto0:
Example output:
vert

tools/cfg/bin_to_dot.cpp
```diff
@@ -41,7 +42,8 @@ const char* kContinueStyle = "style=dotted";
 // a SPIR-V module.
 class DotConverter {
  public:
-  DotConverter(std::iostream* out) : out_(*out) {}
+  DotConverter(libspirv::NameMapper name_mapper, std::iostream* out)
```
antiagainst: Double copy here? Either use `const libspirv::NameMapper&` or use `name_mapper_(std::move(name_mapper))`?

dneto0: I didn't care too much about efficiency.  In this case the name mapper is a lambda that captures a single "this" pointer.  But fixed in  892239b7cc anyway.

antiagainst:
+@

PR #329, opened 2016-08-11
Add GetId(Type* type) API to the type manager

Qining:


source/opt/type_manager.cpp
```diff
@@ -39,6 +39,11 @@ Type* TypeManager::GetType(uint32_t id) const {
   return nullptr;
 }
 
+uint32_t TypeManager::GetId(Type* type) const {
+  if (type_to_id_.count(type) != 0) return type_to_id_.at(type);
```
dneto0: This is fine.  But we could have stored and reused the iterator from find(..) instead of doing the lookup twice.
(Also affects GetType method)

dneto0:
+2

PR #334, opened 2016-08-11
Turn off modifying methods for const iterators.

antiagainst:


source/opt/iterator.h
```diff
@@ -158,8 +160,9 @@ inline bool UptrVectorIterator<VT, IC>::operator<(
 }
 
 template <typename VT, bool IC>
-inline UptrVectorIterator<VT, IC> UptrVectorIterator<VT, IC>::InsertBefore(
-    Uptr value) {
+template <bool C>
```
dneto0: It's confusing to have different template names in the declaration and this definiton.
In the declaration we have:  IsConst as the outer template param, and IC as the inner template param.
In this definition we have: IC as the outer termplate param, and C as the inner template param.

I see the readability gain from using short names like "IC".  But the confusion comes because the same short name (IC) is used for two different things.  Please rename to avoid this kind of conflict.  E.g. `<bool MethodForConstObj = IsConst>` at the declaration, and `<bool MethodForConstObj>` at the definition.

dneto0:
-1 just for a naming confusion.

source/opt/iterator.h
antiagainst: That's fairly reasonable. Using `IsConstForMethod` now. `IC` now is consistently a short version for `IsConst`.

dneto0:
+2

PR #335, opened 2016-08-12
Avoid non-oneliner definition in class and add missing iterators.

antiagainst:


source/opt/basic_block.h
```diff
@@ -45,16 +46,23 @@ class Function;
...
 class BasicBlock {
  public:
+  using iterator = UptrVectorIterator<Instruction>;
+  using const_iterator = UptrVectorIterator<Instruction, true>;
+
   // Creates a basic block with the given enclosing |function| and starting
   // |label|.
-  BasicBlock(std::unique_ptr<Instruction> label)
-      : function_(nullptr), label_(std::move(label)) {}
+  inline BasicBlock(std::unique_ptr<Instruction> label);
```
Qining: Blurbs need update

antiagainst: Done.

Qining:
lgtm +1

source/opt/function.h
```diff
@@ -48,17 +48,15 @@ class Function {
   using const_iterator = UptrVectorIterator<BasicBlock, true>;
 
   // Creates a function instance declared by the given instruction |def_inst|.
```
dneto0: We know this is has to be an OpFunction instruction, right?  Might as well say so.

dneto0:
Suggested minor improvement to one blurb.
But +2

source/opt/function.h
antiagainst: Done.

PR #336, opened 2016-08-12
Move the def-use analysis for single inst to public

Qining:


antiagainst:
We'd better also to rearrange the implementation in the `.cpp` file accordingly.

antiagainst:
+2. Please squash before committing.

source/opt/def_use_manager.cpp
```diff
@@ -41,6 +41,28 @@ void DefUseManager::AnalyzeDefUse(ir::Module* module) {
...
+    switch (inst->GetOperand(i).type) {
+      // For any id type but result id type
+      case SPV_OPERAND_TYPE_ID:
+      case SPV_OPERAND_TYPE_TYPE_ID:
+      case SPV_OPERAND_TYPE_MEMORY_SEMANTICS_ID:
+      case SPV_OPERAND_TYPE_SCOPE_ID: {
+        uint32_t use_id = inst->GetSingleWordOperand(i);
+        // use_id is used by the instruction generating def_id.
+        id_to_uses_[use_id].push_back({inst, i});
+        if (def_id != 0) result_id_to_used_ids_[def_id].push_back(use_id);
```
dneto0: This is not correct if you've already analyzed an instruction for this result id before.
This method should reset result_id_to_used_ids_[def_id] at the very start.

source/opt/def_use_manager.h
```diff
@@ -63,6 +63,9 @@ class DefUseManager {
   // const overload for ForEachInst().
   void AnalyzeDefUse(ir::Module* module);
 
+  // Analyzes the defs and uses in the given |inst|.
+  void AnalyzeInstDefUse(ir::Instruction* inst);
```
dneto0: Need more in the blurb now.  (Or maybe you always needed more in the blurb...)
What if I analyze the same instruction multiple times?  What if I've analyzed some instructions but not others yet.

Basically, the consistency of the data in this manager is in question. Before you could guess that it had no data until AnalyzeDefUse(module) was called.  Then after that you knew it had all the information.  But now that we can do things incrementally, it seems a bit up in the air.

source/opt/def_use_manager.h
```diff
@@ -63,6 +63,9 @@ class DefUseManager {
   // const overload for ForEachInst().
   void AnalyzeDefUse(ir::Module* module);
```
dneto0: What happens if I call AnalyzeDefUse multiple times?  Does it clear the old data first (I would expect it to.)

dneto0:
-1

Qining:
Fix uploaded to CL #337

PR #337, opened 2016-08-12
Clear old def-use record when re-analyze an ID

Qining:
This is to fix the bug in #336

### DIFF
--- a/source/opt/def_use_manager.cpp
+++ b/source/opt/def_use_manager.cpp
@@ -43,6 +43,9 @@ void DefUseManager::AnalyzeDefUse(ir::Module* module) {
 
 void DefUseManager::AnalyzeInstDefUse(ir::Instruction* inst) {
   const uint32_t def_id = inst->result_id();
+  // Clear the records of def_id first if it has been recorded before.
+  ClearDef(def_id);
+
   if (def_id != 0) id_to_def_[def_id] = inst;
 
   for (uint32_t i = 0; i < inst->NumOperands(); ++i) {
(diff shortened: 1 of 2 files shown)


source/opt/def_use_manager.cpp
```diff
@@ -116,6 +105,29 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
   return true;
 }
 
+bool DefUseManager::ClearRecordFor(uint32_t def_id) {
+  if (id_to_def_.count(def_id) == 0) return false;
+
+  // Go through all ids usd by this instruction, remove this instruction's
```
dneto0: "usd" -> "used"

Qining: Fixed in 60b2ab0

source/opt/def_use_manager.h
```diff
@@ -91,6 +91,14 @@ class DefUseManager {
   bool ReplaceAllUsesWith(uint32_t before, uint32_t after);
 
  private:
+  // Clear the internal def-use record of a defined id. Returns true on
+  // success, false if the given |def_id| is not recorded by this manager. This
+  // method erases both the uses of |def_id| and the |def_id|-generating
+  // instruction's use information kept in this manager, but not the operands
+  // in the original instructions. This method does not turn the
```
dneto0: The last sentence is a little weird. I never would have expected it to mutate the instruction.

source/opt/def_use_manager.cpp
```diff
@@ -43,6 +43,13 @@ void DefUseManager::AnalyzeDefUse(ir::Module* module) {
 
 void DefUseManager::AnalyzeInstDefUse(ir::Instruction* inst) {
   const uint32_t def_id = inst->result_id();
+
+  // If the inst result id has already been recorded in this manager, we need
+  // to clear its old records first then re-analyze its use-def record.
+  if (id_to_def_.count(def_id) != 0) {
```
dneto0: Just call ClearRecordFor since it will do the count check anyway.

source/opt/def_use_manager.h
```diff
@@ -91,6 +91,14 @@ class DefUseManager {
...
+  // |def_id|-generating instruction to nop.
+  bool ClearRecordFor(uint32_t def_id);
```
dneto0: Better name would be ClearRecordForDefinition or ClearDefinition(uint32_t def_id)

dneto0:
-1

source/opt/def_use_manager.h
Qining: Changed to `ClearDef()` to be consistent with others: `GetDef()`, `KillDef()`.

source/opt/def_use_manager.h
Qining: Last sentence removed in 3087458.

source/opt/def_use_manager.cpp
Qining: Fixed in 3087458

dneto0:
+2 please squash first.
Why does ClearDef return a bool when you never use the result?

dneto0:
+2  please squash

dneto0:
oh, there's only one commit.  so it's already squashed. :-)

PR #338, opened 2016-08-12
Add iterator getter interface for types_values_

Qining:


source/opt/module.h
```diff
@@ -110,6 +110,12 @@ class Module {
   IteratorRange<inst_iterator> annotations();
   IteratorRange<const_inst_iterator> annotations() const;
 
+  // Iterators for types, constants and global variables instructions.
```
dneto0: This lines up almost perfectly with SPIR-V spec section 9 of the logical layout.
Does this also iterate over OpUndef that are outside of function bodies?
(See latest SPIR-V specs 1.0 rev 7 or 1.1 rev 3)  https://www.khronos.org/registry/spir-v/specs/1.1/SPIRV.html#_changes_from_version_1_00_revision_6

antiagainst: No. `OpUndef` is not included in this vector. Need to change this behavior to comply with r6.

antiagainst: I'll have a PR for it.

antiagainst: Here it is: #345.

antiagainst:
+2

Issue #340, opened 2016-08-12
Potential race on mkdir in update_build_version.py


dneto0:
There are reports of the following part of update_build_version.py errorin out with a File exists exception.
```
    if not os.path.isdir(output_dir):
        os.makedirs(output_dir)
```
Probably a good idea to use the solution in http://stackoverflow.com/questions/600268/mkdir-p-functionality-in-python to make it non-racy.  (And without needing an extra module to be installed.)

dneto0:
Hm.  The way google/shaderc does it is that the update_build_version.py script is not responsible for making the directory, rather the build system has to make the directory and instead update_build_version.py just assumes it's going to write its output file in the current directory.

PR #341, opened 2016-08-12
Avoid race on mkdir

dneto0:
Should fix https://github.com/KhronosGroup/SPIRV-Tools/issues/340


AWoloszyn:
+2 LGTM

dneto0:
Rebased and pushed into master as 033b7d00f06022c2

PR #345, opened 2016-08-12
Allow OpUndef appearing in the type and global variable section.

antiagainst:


dneto0:
Please also make a test case using OpUndef inside a basic block inside a function.

antiagainst:
Added a new test.

Qining:
+2

Qining:
Hmm, I'd prefer to rename the function "RoundtripCheck" to something like AssembleThenDisassemble.

source/opt/ir_loader.cpp
```diff
@@ -87,7 +87,7 @@ void IrLoader::AddInstruction(const spv_parsed_instruction_t* inst) {
         module_->AddType(std::move(spv_inst));
       } else if (IsConstantInst(opcode)) {
         module_->AddConstant(std::move(spv_inst));
-      } else if (opcode == SpvOpVariable) {
+      } else if (opcode == SpvOpVariable || opcode == SpvOpUndef) {
```
dneto0: But an Undef is not a GlobalVariable.
Maybe the method is misnamed?

antiagainst: What about only having `AddType` and `AddGlobalValue`? We can use the latter to add constants, variables, & `OpUndef`s.

dneto0: Sounds good to me.

antiagainst: Done.

dneto0:
+2

PR #346, opened 2016-08-12
Add another construtor for opt::ir::Instruction and Operand

Qining:


source/opt/instruction.cpp
```diff
@@ -50,6 +50,20 @@ Instruction::Instruction(const spv_parsed_instruction_t& inst,
   }
 }
 
+Instruction::Instruction(SpvOp op, uint32_t type_id, uint32_t result_id,
+                         const std::vector<Operand>& in_operands)
+    : opcode_(op), type_id_(type_id), result_id_(result_id), operands_() {
+  if (type_id_ != 0) {
+    operands_.emplace_back(
+        Operand(spv_operand_type_t::SPV_OPERAND_TYPE_TYPE_ID, {type_id_}));
```
antiagainst: Hmm... I think this is still constructing a temporary object and then constructed? Considering `emplace_back(SPV_OPERAND_TYPE_TYPE_ID, std::vector<uint32_t>{type_id})`?

source/opt/instruction.h
```diff
@@ -85,6 +88,11 @@ class Instruction {
   Instruction(const spv_parsed_instruction_t& inst,
               std::vector<Instruction>&& dbg_line = {});
 
+  // Creates an instruction with the given opcode |op|, |result_id|, |type_id|
```
antiagainst: given |op|, |type_id|, |result_id|, and |in_operands| to match the order.

source/opt/instruction.h
```diff
@@ -62,6 +62,9 @@ struct Operand {
   Operand(spv_operand_type_t t, std::vector<uint32_t>&& w)
       : type(t), words(std::move(w)) {}
 
+  Operand(spv_operand_type_t t, const std::vector<uint32_t>& w)
```
antiagainst: Hmm... If you'd think it's necessary to keep a copy of `w` after passing it in this function, what about changing the above function to take `std::vector<uint32_t>`? That way, we only have one method to maintain, and the performance is on par:
* If passed in a rvalue, it will be moved into `w` and then moved to `words`. An extra move compared to this implementation.
* If passed in a lvalue, it will be copied to `w` and then moved to `words`. An extra move compared to this implementation.

But this is nit picking. Coder's choice.

source/opt/instruction.h
```diff
@@ -85,6 +88,11 @@ class Instruction {
   Instruction(const spv_parsed_instruction_t& inst,
               std::vector<Instruction>&& dbg_line = {});
 
+  // Creates an instruction with the given opcode |op|, |result_id|, |type_id|
+  // and |in_operands|.
+  Instruction(SpvOp op, uint32_t type_id, uint32_t result_id,
+              const std::vector<Operand>& in_operands);
```
antiagainst: Similar as above. But, your choice.

source/opt/instruction.cpp
Qining: It is interesting, I need to use std::initializer_list<uint32_t> for the inner vector, otherwise `emplace_back()` can not deduce the type of the initializer list.

Fixed in 4911cf6

antiagainst:
+2

source/opt/instruction.cpp
```diff
@@ -50,6 +50,20 @@ Instruction::Instruction(const spv_parsed_instruction_t& inst,
   }
 }
 
+Instruction::Instruction(SpvOp op, uint32_t ty_id, uint32_t res_id,
+                         const std::vector<Operand>& in_operands)
+    : opcode_(op), type_id_(ty_id), result_id_(res_id), operands_() {
+  if (type_id_ != 0) {
+    operands_.emplace_back(spv_operand_type_t::SPV_OPERAND_TYPE_TYPE_ID,
+                           std::initializer_list<uint32_t>{type_id_});
```
antiagainst: IWYU. `#include <initializer_list>`.

PR #347, opened 2016-08-13
Add more tests and KillInst() to DefUseManager

Qining:
Add KillInst() so that def-use manager can handle instrutions that do
not have result ids.

Add ClearDef() and ClearInst() two private functions.

Add tests for AnalyzeInstDefUse().

Add tests for KillInst().

Add a test that call AnalyzeDefUse() on two different modules.

### DIFF
--- a/source/opt/def_use_manager.cpp
+++ b/source/opt/def_use_manager.cpp
@@ -37,33 +37,32 @@ namespace opt {
 namespace analysis {
 
 void DefUseManager::AnalyzeDefUse(ir::Module* module) {
+  Reset();
   module->ForEachInst(std::bind(&DefUseManager::AnalyzeInstDefUse, this,
                                 std::placeholders::_1));
 }
 
 void DefUseManager::AnalyzeInstDefUse(ir::Instruction* inst) {
-  const uint32_t def_id = inst->result_id();
-  // Clear the records of def_id first if it has been recorded before.
-  ClearDef(def_id);
+  // Clear the records of the given instruction if it has been analyzed before.
+  ClearInst(inst);
 
-  if (def_id != 0) id_to_def_[def_id] = inst;
+  const uint32_t def_id = inst->result_id();
+  if (def_id != 0) {
+    // If the new instruction defines an existing result id, clear the records
+    // of the existing id first.
+    ClearDef(def_id);
+    id_to_def_[def_id] = inst;
+  }
 
   for (uint32_t i = 0; i < inst->NumOperands(); ++i) {
-    switch (inst->GetOperand(i).type) {
-      // For any id type but result id type
-      case SPV_OPERAND_TYPE_ID:
-      case SPV_OPERAND_TYPE_TYPE_ID:
-      case SPV_OPERAND_TYPE_MEMORY_SEMANTICS_ID:
-      case SPV_OPERAND_TYPE_SCOPE_ID: {
-        uint32_t use_id = inst->GetSingleWordOperand(i);
-        // use_id is used by the instruction generating def_id.
-        id_to_uses_[use_id].push_back({inst, i});
-        if (def_id != 0) result_id_to_used_ids_[def_id].push_back(use_id);
-      } break;
-      default:
-        break;
+    if (ShouldRecord(inst->GetOperand(i))) {
+      uint32_t use_id = inst->GetSingleWordOperand(i);
+      // use_id is used by this instruction.
+      id_to_uses_[use_id].push_back({inst, i});
     }
   }
+
+  analyzed_insts_.insert(inst);
 }
 
 ir::Instruction* DefUseManager::GetDef(uint32_t id) {
@@ -78,11 +77,14 @@ UseList* DefUseManager::GetUses(uint32_t id) {
 
 bool DefUseManager::KillDef(uint32_t id) {
   if (id_to_def_.count(id) == 0) return false;
-
-  ir::Instruction* defining_inst = id_to_def_.at(id);
-  ClearDef(id);
-  defining_inst->ToNop();
+  KillInst(id_to_def_[id]);
   return true;
+}
+
+void DefUseManager::KillInst(ir::Instruction* inst) {
+  if (!inst) return;
+  ClearInst(inst);
+  inst->ToNop();
 }
 
 bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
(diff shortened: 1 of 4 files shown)


source/opt/def_use_manager.h
```diff
@@ -84,27 +87,50 @@ class DefUseManager {
...
+  // If the instruction is defining a result id, the uses of the result id will
+  // also be removed. Note that if |inst| has not been analyzed before, this
+  // function does nothing even though |inst| may define an existing result id.
+  void ClearInst(ir::Instruction* inst);
+
+  // Returns true if the operand's id should be recorded, otherwise returns
+  // false;
+  bool ShouldRecord(const ir::Operand& operand) const;
+
+  // Erases the record that: instruction |user| uses id |used_id|.
```
dneto0: remove the ":"

source/opt/def_use_manager.h
```diff
@@ -84,27 +87,50 @@ class DefUseManager {
...
+  // has been analyzed by this manager. The use information of its operand ids
+  // will be updated: "the record: |inst| uses |operand id| will be removed".
+  // If the instruction is defining a result id, the uses of the result id will
+  // also be removed. Note that if |inst| has not been analyzed before, this
+  // function does nothing even though |inst| may define an existing result id.
+  void ClearInst(ir::Instruction* inst);
+
+  // Returns true if the operand's id should be recorded, otherwise returns
+  // false;
+  bool ShouldRecord(const ir::Operand& operand) const;
```
dneto0: This is just like saying "Returns the value that it returns". ?   ;-P
Under what conditions should an operand Id be recorded?

dneto0: Suggest the function name should be IsIdUse(...)
Because it's true for any Id that is not a definition, hence it's an Id that's a use.

source/opt/def_use_manager.cpp
```diff
@@ -103,24 +105,59 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
...
+  // before.
+  if (!inst || analyzed_insts_.count(inst) == 0) return;
+
+  // Go through all ids, except the result id, used by this instruction, remove
+  // this instruction's uses of those ids.
+  for (uint32_t i = 0; i < inst->NumOperands(); i++) {
+    if (ShouldRecord(inst->GetOperand(i))) {
+      uint32_t operand_id = inst->GetSingleWordOperand(i);
+      // Skip if the operand id is not recorded.
+      if (id_to_uses_.count(operand_id) == 0) {
```
dneto0: Put  the guard inside EreaseInstUseIdRecord; make it resilient to being called with an Id that isn't already in the map.

source/opt/def_use_manager.cpp
```diff
@@ -103,24 +105,59 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
...
+    case SPV_OPERAND_TYPE_SCOPE_ID:
+      return true;
+    default:
+      return false;
+  }
+}
+
+void DefUseManager::EraseInstUseIdRecord(const ir::Instruction& user,
+                                         uint32_t used_id) {
+  auto& uses = id_to_uses_[used_id];
```
dneto0: Use id_to_uses_.find(...) to guard against accessing a used_id that isn't already in the map.  Doing that will make this map resilient to callers that don't do the check.
(And update the blurb to say it's a no-op on an Id that isn't already in the map.)

source/opt/def_use_manager.cpp
```diff
@@ -103,24 +105,59 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
 
 void DefUseManager::ClearDef(uint32_t def_id) {
   if (id_to_def_.count(def_id) == 0) return;
```
dneto0: This does the lookup twice, which could be wasteful.

source/opt/def_use_manager.h
```diff
@@ -84,27 +87,50 @@ class DefUseManager {
...
+
+  // Resets the internal records
+  void Reset() {
+    analyzed_insts_.clear();
+    id_to_def_.clear();
+    id_to_uses_.clear();
+  }
+
+  std::unordered_set<ir::Instruction*>
+      analyzed_insts_;      // Analyzed instructions
```
dneto0: I don't think we need this map at all.  It just guards clearing of the def+uses of an instruction.  But the clear method should be resilient to the case where the instruction hasn't been analyzed before.
?!

test/opt/test_def_use.cpp
```diff
@@ -1155,4 +1157,271 @@ TEST(DefUseTest, OpSwitch) {
...
+
+// Creates an OpBranch |target_id| instruction.
+ir::Instruction BranchInstruction(uint32_t target_id) {
+  return ir::Instruction(SpvOp::SpvOpBranch, 0, 0,
+                         {
+                             ir::Operand(SPV_OPERAND_TYPE_ID, {target_id}),
+                         });
+}
+
+struct AnalyzeInstDefUseTestCase {
```
dneto0: Needs a comment.

test/opt/test_def_use.cpp
```diff
@@ -1155,4 +1157,271 @@ TEST(DefUseTest, OpSwitch) {
...
+  std::vector<ir::Instruction> insts;
+  InstDefUse du;
```
dneto0: Is this the expected analysis result?  Say so.

test/opt/test_def_use.cpp
```diff
@@ -1155,4 +1157,271 @@ TEST(DefUseTest, OpSwitch) {
...
+          OpTypeIntInstruction(1),
+          ConstantBoolInstruction(true, 1, 2),
+        },
+        {
+          { // defs
+            {1, "%1 = OpTypeInt 32 1"},
+            {2, "%2 = OpConstantTrue %1"},
+          },
+          { // uses
+            {1,{"%2 = OpConstantTrue %1",}},
```
dneto0: there's a confusing extra comma after %1"....

test/opt/test_def_use.cpp
```diff
@@ -1155,4 +1157,271 @@ TEST(DefUseTest, OpSwitch) {
...
+          ConstantBoolInstruction(true, 1, 2),
+          // The def-use info of the following instruction should overwrite the
+          // records of the above one.
+          ConstantBoolInstruction(false, 3, 2),
+        },
+        {
+          // defs
+          {{2, "%2 = OpConstantFalse %3"},},
+          // uses
+          {{3, {"%2 = OpConstantFalse %3",}},}
```
dneto0: Too many commas.

test/opt/test_def_use.cpp
```diff
@@ -1155,4 +1157,271 @@ TEST(DefUseTest, OpSwitch) {
...
+        // not have result id.
+        {
+          BranchInstruction(2),
+          LabelInstruction(2),
+        },
+        {
+          // defs
+          {{2, "%2 = OpLabel"},},
+          // uses
+          {{2, {"OpBranch %2",}},},
```
dneto0: Too many commas.  This isn't python

test/opt/test_def_use.cpp
```diff
@@ -1155,4 +1157,271 @@ TEST(DefUseTest, OpSwitch) {
...
+
+using KillInstTest = ::testing::TestWithParam<KillInstTestCase>;
+
+TEST_P(KillInstTest, Case) {
+  auto tc = GetParam();
+
+  // Build module.
+  const std::vector<const char*> text = {tc.before};
+  std::unique_ptr<ir::Module> module =
+      SpvTools(SPV_ENV_UNIVERSAL_1_1).BuildModule(JoinAllInsts(text));
```
dneto0: Why did you make a vector of a single element when you're going to join all its elements to make the string you already had?

test/opt/test_def_use.cpp
```diff
@@ -1155,4 +1157,271 @@ TEST(DefUseTest, OpSwitch) {
...
+  EXPECT_EQ(tc.after, DisassembleModule(module.get()));
+  CheckDef(tc.du, manager.id_to_defs());
+  CheckUse(tc.du, manager.id_to_uses());
+}
+
+// clang-format off
+INSTANTIATE_TEST_CASE_P(
+    TestCase, KillInstTest,
+    ::testing::ValuesIn(std::vector<KillInstTestCase>{
+      // Kill id defining instrutions.
```
dneto0: "instructions"  was missing a c

dneto0:
-1
Tests look ok.
Several issues with the code need addressing.

source/opt/def_use_manager.h
Qining: There may be four cases (depends on whether we want to make ir::Instruction to be immutable later) we need to differentiate:
1) a new instruction without result_id but with already analyzed operands.
2) a new instruction with result_id and also has its operands analyzed.
3) an analyzed but mutated instruction, the instruction does have a result id.
4) an analyzed but mutated instruction, the instruction does not have a result id.

For case 1), the desired result is that, the def-use records of the analyzed operands should be +1 to have the new added instruction as their new uses.

For case 2), the desired result is that, the original result_id defining instruction should be cleared first, then the new added instruction can update the records for the involved operands. This is a -1 followed by +1 for the operands.

For case 3), we should -1 first then +1, same as a new instruction with existing result id.

For case 4), we should also -1 first then +1, same as a new instruction with existing result id, but this is different with case 1).

To differentiate the four cases we have two options:

a) See if the instruction has a result id.
If the instruction has a result id, clear internal records for the id, then analyze the new instruction. Otherwise analyze the new instruction directly. But this way we can not differentiate case 1 and case 4. We will -1 for the operands in case 1.

b) Use a set to hold all the analyzed instructions.
If the instruction has been analyzed, clear its record first.
If the result id has been analyzed before, clear its def_id record also.
Add record for the given instruction as if it is a new instruction.

Actually the fundamental solution to this hassle is to use instruction* instead of id(uint32_t) for the internal maps then we can unify the behavior. But that may changed the structure a lot. This is the reason why we need such a set here.

dneto0: My point was that asking to remove a use from an instruction *should* be a no-op if that use is not actually recorded.  Then your cases 1 and 4 don't need to be distinguished. 

But I see there is complexity in having instruction's operands already be analyzed from the definer's perspective, and then you try to analyze the user's perspective.  I have a headache and I'd like a second opinion on this.

source/opt/def_use_manager.h
Qining: Fixed in 4d7c019

source/opt/def_use_manager.h
Qining: Fixed in 4d7c019

source/opt/def_use_manager.cpp
Qining: Fixed in 4d7c019.
Actually the iteration of the operands are also bundled in a private function.

source/opt/def_use_manager.cpp
Qining: Fixed in 4d7c019.

And it doesn't call ClearInst() anymore, so for instructions with result_id, we won't look for the maps twice.

source/opt/def_use_manager.cpp
Qining: Fixed in 4d7c019.

And it doesn't call ClearInst() anymore, so for instructions with result_id, we won't look for the maps twice.

test/opt/test_def_use.cpp
Qining: Change the name `du` to `expected_def_use` in 4d7c019

test/opt/test_def_use.cpp
Qining: Removed in 4d7c019.

Actually I find the commas help me to know the hierarchy of the test cases, but I can remove it for sure.

test/opt/test_def_use.cpp
Qining: Fixed in 4d7c019

test/opt/test_def_use.cpp
Qining: Fixed in 4d7c019

source/opt/def_use_manager.h
Qining: Oh, yes, you are right. If the target instruction is really a new instruction, ClearInst() won't change the def-use records of the target instruction's operand id.

I missed the "is same instruction" check when erasing use records.

Qining:
Close as it is moved to #356

PR #348, opened 2016-08-13
Do not use SetOpCode() to freeze spec constants

Qining:
Remove the opt::ir::Instruction::SetOpCode() interface.

Spec constant freezing is done by adding new front-end constant
instructions and turning the original spec constant instruction to Nop.

### DIFF
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -100,11 +100,6 @@ class Instruction {
   Instruction& operator=(Instruction&&);
 
   SpvOp opcode() const { return opcode_; }
-  // Sets the opcode of this instruction to a specific opcode. Note this may
-  // invalidate the instruction.
-  // TODO(qining): Remove this function when instruction building and insertion
-  // is well implemented.
-  void SetOpcode(SpvOp op) { opcode_ = op; }
   uint32_t type_id() const { return type_id_; }
   uint32_t result_id() const { return result_id_; }
   // Returns the vector of line-related debug instructions attached to this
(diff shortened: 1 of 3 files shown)


antiagainst:
I'm afraid I don't quite get the rationale behind this PR.

For me, the old approach is clear (turning `OpSpec*` to `Op*` directly) and efficient (since we don't create extra copies and insert them, which is a costly operation). Would you mind to explain why the new way?

dneto0:
(also "Constants" is mis-spelled in the title)

Qining:
I think the reason for having this is that, setting opcode directly is generally an unsafe behavior. And that is also the reason why we have a TODO at that location. There may be rules for creating valid instructions, like different constructor for different subclass of instructions, and in the constructor we can impose rules to make sure the built instruction is somewhat 'more' likely to be valid. So having an interface to set arbitrary opcode looks quite dangerous.

But if we really want to keep the optimization framework lightweight, and let the user to take the full responsibility to generate valid code, we can keep the SetOpcode().

antiagainst:
The current design is to keep the in-memory representation just a vehicle of data, nothing more. It is lightweight and the user is expected to guarantee its validness. This gives us the benefits of not imposing heavy analyses on passes that does not need them (e.g., def use is not used for stripping debug info) and flexibility in writing tests (we can sometimes create invalid in-memory representation for testing's ease).

So I think it's preferable to keep the old `SetOpcode` way.

Qining:
closed as we decided to keep the SetOpcode()

PR #350, opened 2016-08-15
Derive iterator from std::iterator to reap standarad algorithms.

antiagainst:
By deriving from `std::iterator`, `iterator_traits` will be properly
set up for our custom iterator type, thus we can use algorithms
from STL with our custom iterators.


dneto0:
+1

Qining:
lgtm +1

antiagainst:
Landed as 9747d33d08cd3ef069c08b37ff94a971707a5741

Issue #349, opened 2016-08-14
io.h fails to compile with GCC 6.1


atom-symbol:
GCC 4.9.3: compilation succeeds
GCC 6.1.0: compilation fails

    In file included from /var/tmp/portage/media-libs/vulkan-base-1.0.21/work/vulkan-base-1.0.21/external/spirv-tools/tools/as/as.cpp:33:0:
    /var/tmp/portage/media-libs/vulkan-base-1.0.21/work/vulkan-base-1.0.21/external/spirv-tools/tools/io.h: In function 'bool ReadFile(const char*, const char*, std::vector<T>*) [with T = char]':
    /var/tmp/portage/media-libs/vulkan-base-1.0.21/work/vulkan-base-1.0.21/external/spirv-tools/tools/io.h:53:16: error: ignoring return value of 'long int ftell(FILE*)', declared with attribute warn_unused_result [-Werror=unused-result]
           if (ftell(fp) % sizeof(T)) {
           ~~~~~^~~~
    cc1plus: all warnings being treated as errors

Please try to compile the project, if possible, with GCC 4.9, 5.4, 6.1 and maybe with Clang-3.9 before every release. Thanks.

umar456:
The result of `ftell` is being used in the modulus operation. I think this is a problem with the compiler and not spirv-tools.

atom-symbol:
You are right to some extent. But `sizeof(T=char)` is 1, and `(x%1)` is always 0, so the compiler is in fact right that the returned value isn't being used.

If we want to allow compilers to get smarter over time, I recommend the following fix:

    if (sizeof(T) > 1 && ftell(fp) % sizeof(T) != 0) { ... }

Alternatively, if we want to prevent compilers from getting smarter we can ignore the error message and close this issue.

So ... make your choice.

dneto0:
Thank you @atom-symbol for the bug report.  It should be fixed in master now.

PR #351, opened 2016-08-15
Avoid integer modulo by 1, to avoid warning

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/349


antiagainst:
+2

PR #354, opened 2016-08-16
Replace the SetOperand() with SetInOperand() and SetResultType()

Qining:
The result id of an instruction should never be changed, as it can be considered as another representation of the instruction.


source/opt/def_use_manager.cpp
```diff
@@ -86,8 +86,17 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
 
   for (auto it = id_to_uses_[before].cbegin(); it != id_to_uses_[before].cend();
        ++it) {
-    // Make the modification in the instruction.
-    it->inst->SetOperand(it->operand_index, {after});
+    uint32_t type_result_id_count =
```
antiagainst: const

source/opt/instruction.h
```diff
@@ -179,10 +181,18 @@ inline const Operand& Instruction::GetOperand(uint32_t index) const {
...
+                                      std::vector<uint32_t>&& data) {
+  assert(index + TypeResultIdCount() < operands_.size() &&
+         "operand index out of bound");
+  operands_[index + TypeResultIdCount()].words = std::move(data);
+}
+
+inline void Instruction::SetResultType(uint32_t ty_id) {
+  if (type_id_ != 0) {
+    type_id_ = ty_id;
+    operands_[0] = {SPV_OPERAND_TYPE_TYPE_ID, {ty_id}};
```
antiagainst: I'd think it would be nice to also do an `assert` here:

```
assert(operands_.front().type == SPV_OPERAND_TYPE_TYPE_ID);
operands_.front().words = {type_id};
```

source/opt/instruction.h
```diff
@@ -128,8 +128,10 @@ class Instruction {
   // not expected to be used with logical operands consisting of multiple SPIR-V
   // words.
   uint32_t GetSingleWordOperand(uint32_t index) const;
-  // Sets the |index|-th operand's data to the given |data|.
-  inline void SetOperand(uint32_t index, std::vector<uint32_t>&& data);
+  // Sets the |index|-th in-operand's data to the given |data|.
+  inline void SetInOperand(uint32_t index, std::vector<uint32_t>&& data);
+  // Sets the result type id.
+  inline void SetResultType(uint32_t ty_id);
```
antiagainst: I'd suggest we put some comments/rationale after this to warn against defining another function to tweak the result id. Otherwise, somebody may add it without knowing this later.

source/opt/def_use_manager.cpp
```diff
@@ -86,8 +86,17 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
...
+        (it->inst->result_id() != 0) + (it->inst->type_id() != 0);
+    if (it->inst->type_id() != 0 && it->operand_index == 0) {
```
antiagainst: This `if`-statement seems a little strange for me. What about

```
if (it->operand_index) {
  assert(it->inst->type_id() != 0 && "result type id considered as using while the instruction doesn't have a result type id";
  it->inst->SetResultType(after);
```

source/opt/instruction.h
Qining: Blurbs added in a7c772a

source/opt/instruction.h
Qining: assertion added in a7c772a

source/opt/def_use_manager.cpp
Qining: fixed in a7c772a

source/opt/def_use_manager.cpp
Qining: Added in 80c7770

antiagainst:
+2

PR #355, opened 2016-08-16
Make clear that we are using ptrdiff_t as the iterator difference type.

antiagainst:
And `ptrdiff_t` is a implementation defined signed type. Comparing it
with unsigned number literal causes compiler warnings.


Qining:
+2

PR #356, opened 2016-08-16
Add tests for individual instruction def-use analysis

Qining:
Rebase and refine the CL #347 .

Removed the test for analyzing two modules, as current Module is bind to a module when it is created.

Based on CL #354

### DIFF
--- a/source/opt/def_use_manager.cpp
+++ b/source/opt/def_use_manager.cpp
@@ -38,10 +38,22 @@ namespace analysis {
 
 void DefUseManager::AnalyzeInstDefUse(ir::Instruction* inst) {
   const uint32_t def_id = inst->result_id();
-  // Clear the records of def_id first if it has been recorded before.
-  ClearDef(def_id);
+  if (def_id != 0) {
+    auto iter = id_to_def_.find(def_id);
+    if (iter != id_to_def_.end()) {
+      // Clear the original instruction that defining the same result id of the
+      // new instruction.
+      ClearInst(iter->second);
+    }
+    id_to_def_[def_id] = inst;
+  } else {
+    ClearInst(inst);
+  }
 
-  if (def_id != 0) id_to_def_[def_id] = inst;
+  // Create entry for the given instruction. Note that the instruction may
+  // not have any in-operands. In such cases, we still need a entry for those
+  // instructions so this manager knows it has seen the instruction later.
+  inst_to_used_ids_[inst] = {};
 
   for (uint32_t i = 0; i < inst->NumOperands(); ++i) {
     switch (inst->GetOperand(i).type) {
@@ -53,7 +65,7 @@ void DefUseManager::AnalyzeInstDefUse(ir::Instruction* inst) {
         uint32_t use_id = inst->GetSingleWordOperand(i);
         // use_id is used by the instruction generating def_id.
         id_to_uses_[use_id].push_back({inst, i});
-        if (def_id != 0) result_id_to_used_ids_[def_id].push_back(use_id);
+        inst_to_used_ids_[inst].push_back(use_id);
       } break;
       default:
         break;
@@ -62,22 +74,28 @@ void DefUseManager::AnalyzeInstDefUse(ir::Instruction* inst) {
 }
 
 ir::Instruction* DefUseManager::GetDef(uint32_t id) {
-  if (id_to_def_.count(id) == 0) return nullptr;
-  return id_to_def_.at(id);
+  auto iter = id_to_def_.find(id);
+  if (iter == id_to_def_.end()) return nullptr;
+  return iter->second;
 }
 
 UseList* DefUseManager::GetUses(uint32_t id) {
-  if (id_to_uses_.count(id) == 0) return nullptr;
-  return &id_to_uses_.at(id);
+  auto iter = id_to_uses_.find(id);
+  if (iter == id_to_uses_.end()) return nullptr;
+  return &iter->second;
 }
 
 bool DefUseManager::KillDef(uint32_t id) {
-  if (id_to_def_.count(id) == 0) return false;
-
-  ir::Instruction* defining_inst = id_to_def_.at(id);
-  ClearDef(id);
-  defining_inst->ToNop();
+  auto iter = id_to_def_.find(id);
+  if (iter == id_to_def_.end()) return false;
+  KillInst(iter->second);
   return true;
+}
+
+void DefUseManager::KillInst(ir::Instruction* inst) {
+  if (!inst) return;
+  ClearInst(inst);
+  inst->ToNop();
 }
 
 bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
@@ -88,14 +106,21 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
        ++it) {
     const uint32_t type_result_id_count =
         (it->inst->result_id() != 0) + (it->inst->type_id() != 0);
-    if (!it->operand_index) {
-      assert(it->inst->type_id() != 0 &&
-             "result type id considered as using while the instruction doesn't "
-             "have a result type id");
-      it->inst->SetResultType(after);
+
+    if (it->operand_index < type_result_id_count) {
+      // Update the type_id. Note that result id is immutable so it should
+      // never be updated.
+      if (it->inst->type_id() != 0 && it->operand_index == 0) {
+        it->inst->SetResultType(after);
+      } else if (it->inst->type_id() == 0) {
+        assert(false &&
+               "Result type id considered as using while the instruction "
+               "doesn't have a result type id.");
+      } else {
+        assert(false && "Trying Setting the result id which is immutable.");
+      }
     } else {
-      assert(it->operand_index >= type_result_id_count &&
-             "the operand to be set is not an in-operand.");
+      // Update an in-operand.
       uint32_t in_operand_pos = it->operand_index - type_result_id_count;
       // Make the modification in the instruction.
       it->inst->SetInOperand(in_operand_pos, {after});
@@ -109,30 +134,42 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
 }
 
 void DefUseManager::AnalyzeDefUse(ir::Module* module) {
+  if (!module) return;
   module->ForEachInst(std::bind(&DefUseManager::AnalyzeInstDefUse, this,
                                 std::placeholders::_1));
 }
 
-void DefUseManager::ClearDef(uint32_t def_id) {
-  if (id_to_def_.count(def_id) == 0) return;
+void DefUseManager::ClearInst(ir::Instruction* inst) {
+  auto iter = inst_to_used_ids_.find(inst);
+  if (iter != inst_to_used_ids_.end()) {
+    EraseUseRecordsOfOperandIds(inst);
+    if (inst->result_id() != 0) {
+      id_to_uses_.erase(inst->result_id());  // Remove all uses of this id.
+      id_to_def_.erase(inst->result_id());
+    }
+  }
+}
 
+void DefUseManager::EraseUseRecordsOfOperandIds(const ir::Instruction* inst) {
   // Go through all ids used by this instruction, remove this instruction's
   // uses of them.
-  for (const auto use_id : result_id_to_used_ids_[def_id]) {
-    if (id_to_uses_.count(use_id) == 0) continue;
-    auto& uses = id_to_uses_[use_id];
-    for (auto it = uses.begin(); it != uses.end();) {
-      if (it->inst->result_id() == def_id) {
-        it = uses.erase(it);
-      } else {
-        ++it;
+  auto iter = inst_to_used_ids_.find(inst);
+  if (iter != inst_to_used_ids_.end()) {
+    for (const auto use_id : iter->second) {
+      auto uses_iter = id_to_uses_.find(use_id);
+      if (uses_iter == id_to_uses_.end()) continue;
+      auto& uses = uses_iter->second;
+      for (auto it = uses.begin(); it != uses.end();) {
+        if (it->inst == inst) {
+          it = uses.erase(it);
+        } else {
+          ++it;
+        }
       }
+      if (uses.empty()) id_to_uses_.erase(use_id);
     }
-    if (uses.empty()) id_to_uses_.erase(use_id);
+    inst_to_used_ids_.erase(inst);
   }
-  result_id_to_used_ids_.erase(def_id);
-  id_to_uses_.erase(def_id);  // Remove all uses of this id.
-  id_to_def_.erase(def_id);
 }
 
 }  // namespace analysis
--- a/source/opt/def_use_manager.h
+++ b/source/opt/def_use_manager.h
@@ -80,10 +80,13 @@ class DefUseManager {
 
   // Turns the instruction defining the given |id| into a Nop. Returns true on
   // success, false if the given |id| is not defined at all. This method also
-  // erases both the uses of |id| and the |id|-generating instruction's use
-  // information kept in this manager, but not the operands in the original
-  // instructions.
+  // erases both the uses of |id| and the information of this |id|-generating
+  // instruction's uses of its operands.
   bool KillDef(uint32_t id);
+  // Turns the given instruction |inst| to a Nop. This method erases the
+  // information of the given instruction's uses of its operands. If |inst|
+  // defines an result id, the uses of the result id will also be erased.
+  void KillInst(ir::Instruction* inst);
   // Replaces all uses of |before| id with |after| id. Returns true if any
   // replacement happens. This method does not kill the definition of the
   // |before| id. If |after| is the same as |before|, does nothing and returns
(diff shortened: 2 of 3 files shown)


source/opt/def_use_manager.cpp
```diff
@@ -109,30 +116,48 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
 }
 
 void DefUseManager::AnalyzeDefUse(ir::Module* module) {
+  if (!module) return;
```
antiagainst: Nice! ;)

source/opt/def_use_manager.cpp
```diff
@@ -73,13 +77,16 @@ UseList* DefUseManager::GetUses(uint32_t id) {
 
 bool DefUseManager::KillDef(uint32_t id) {
   if (id_to_def_.count(id) == 0) return false;
-
-  ir::Instruction* defining_inst = id_to_def_.at(id);
-  ClearDef(id);
-  defining_inst->ToNop();
+  KillInst(id_to_def_[id]);
```
antiagainst: Could use an iterator to avoid searching twice.

source/opt/def_use_manager.h
```diff
@@ -92,23 +96,34 @@ class DefUseManager {
...
+  // method will update the use information of the operand ids of |inst|. The
+  // record: |inst| uses an |id|, will be removed from the use records of |id|.
+  // If |inst| defines an result id, the use record of this result id will also
+  // be removed. Does nothing if |inst| was not analyzed before.
+  void ClearInst(ir::Instruction* inst);
+
+  // Erases the records that a given instruction |inst| uses its operand ids.
+  void EraseUseRecordsOfOperandIds(const ir::Instruction* inst);
+
+  using InstToUsedIdsMap =
```
antiagainst: Type declaration should appear at the very beginning of each section (after `private:` for this case).

source/opt/def_use_manager.h
```diff
@@ -92,23 +96,34 @@ class DefUseManager {
...
+      std::unordered_map<const ir::Instruction*, std::vector<uint32_t>>;
 
   IdToDefMap id_to_def_;    // Mapping from ids to their definitions
   IdToUsesMap id_to_uses_;  // Mapping from ids to their uses
   // Mapping from result ids to the ids used in the instructions generating the
```
antiagainst: `Mapping from instructions to ...`

source/opt/def_use_manager.cpp
```diff
@@ -38,10 +38,14 @@ namespace analysis {
 
 void DefUseManager::AnalyzeInstDefUse(ir::Instruction* inst) {
   const uint32_t def_id = inst->result_id();
-  // Clear the records of def_id first if it has been recorded before.
-  ClearDef(def_id);
+  if (def_id != 0) {
+    ClearDef(def_id);
```
antiagainst: It seems this is the only use of `ClearDef`? Actually we can use `ClearInst(inst)` here directly so that we can remove `ClearDef`?

source/opt/def_use_manager.cpp
```diff
@@ -109,30 +116,48 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
...
-    auto& uses = id_to_uses_[use_id];
-    for (auto it = uses.begin(); it != uses.end();) {
-      if (it->inst->result_id() == def_id) {
-        it = uses.erase(it);
-      } else {
-        ++it;
+  auto iter = inst_to_used_ids_.find(inst);
+  if (iter != inst_to_used_ids_.end()) {
+    for (const auto use_id : iter->second) {
+      if (id_to_uses_.count(use_id) == 0) continue;
```
antiagainst: Can also use iterator to avoid searching twice here.

source/opt/def_use_manager.cpp
```diff
@@ -38,10 +38,14 @@ namespace analysis {
...
-  ClearDef(def_id);
+  if (def_id != 0) {
+    ClearDef(def_id);
+    id_to_def_[def_id] = inst;
+  } else {
+    ClearInst(inst);
+  }
 
-  if (def_id != 0) id_to_def_[def_id] = inst;
+  inst_to_used_ids_[inst].reserve(inst->NumOperands());
```
antiagainst: I don't think we need this. This will highly likely reserve more than what's needed.

Most of the instructions take <= 2 operand IDs and I think `std::vector` should already allocate enough space and optimize that case.

source/opt/def_use_manager.cpp
```diff
@@ -109,30 +116,48 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
...
 
 void DefUseManager::AnalyzeDefUse(ir::Module* module) {
+  if (!module) return;
   module->ForEachInst(std::bind(&DefUseManager::AnalyzeInstDefUse, this,
                                 std::placeholders::_1));
 }
 
 void DefUseManager::ClearDef(uint32_t def_id) {
-  if (id_to_def_.count(def_id) == 0) return;
+  auto iter = id_to_def_.find(def_id);
```
antiagainst: As noted above, this method can potentially be removed.

source/opt/def_use_manager.h
```diff
@@ -84,6 +84,10 @@ class DefUseManager {
   // information kept in this manager, but not the operands in the original
   // instructions.
   bool KillDef(uint32_t id);
+  // Turns the given instruction |inst| to a Nop, also erases both the use
+  // records of the result id of |inst| (if any) and the corresponding use
+  // records of: |inst| useing |inst|'s operand ids.
```
antiagainst: This line reads a little awkward. Consider something similar to `KillDef`?

source/opt/def_use_manager.h
```diff
@@ -92,23 +96,34 @@ class DefUseManager {
...
-  // manager, but not the operands in the original instructions.
+  // manager, but not the operands in the original instructions. Does nothing
+  // if |def_id| was not recorded before.
   void ClearDef(uint32_t def_id);
 
-  using ResultIdToUsedIdsMap =
-      std::unordered_map<uint32_t, std::vector<uint32_t>>;
+  // Clear the internal def-use record of the given instruction |inst|. This
+  // method will update the use information of the operand ids of |inst|. The
+  // record: |inst| uses an |id|, will be removed from the use records of |id|.
```
antiagainst: I'd suggest to limit the use of `|<param>|` to reference a given parameter in the signature.

antiagainst: Also, `The record---|inst| uses an id---will be ...`?

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
   }
 }
 
+// Creates an |result_id| = OpTypeInt 32 1 instruction.
+ir::Instruction OpTypeIntInstruction(uint32_t result_id) {
```
antiagainst: Int32Instruction?

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
...
+  return ir::Instruction(SpvOp::SpvOpTypeInt, 0, result_id,
+                         {ir::Operand(SPV_OPERAND_TYPE_LITERAL_INTEGER, {32}),
+                          ir::Operand(SPV_OPERAND_TYPE_LITERAL_INTEGER, {1})});
+}
+
+// Creates an |result_id| = OpConstantTrue/Flase |type_id| instruction.
+ir::Instruction ConstantBoolInstruction(bool value, uint32_t type_id,
+                                        uint32_t result_id) {
+  return ir::Instruction(
```
antiagainst: Just curious. Will `return {...}` work?

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
...
+// Creates an |result_id| = OpLabel instruction.
+ir::Instruction LabelInstruction(uint32_t result_id) {
+  return ir::Instruction(SpvOp::SpvOpLabel, 0, result_id, {});
+}
+
+// Creates an OpBranch |target_id| instruction.
+ir::Instruction BranchInstruction(uint32_t target_id) {
+  return ir::Instruction(SpvOp::SpvOpBranch, 0, 0,
+                         {
+                             ir::Operand(SPV_OPERAND_TYPE_ID, {target_id}),
```
antiagainst: It seems we can leverage clang-format a little bit here.

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
...
+                         });
+}
+
+struct AnalyzeInstDefUseTestCase {
+  std::vector<ir::Instruction> insts;  // instrutions to be analyzed in order.
+  InstDefUse expected_define_use;      // expected analysis result.
```
antiagainst: `expected_def_use` is fine.

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
...
+          OpTypeIntInstruction(1),
+          ConstantBoolInstruction(true, 1, 2),
+        },
+        {
+          { // defs
+            {1, "%1 = OpTypeInt 32 1"},
+            {2, "%2 = OpConstantTrue %1"},
+          },
+          { // uses
+            {1,{"%2 = OpConstantTrue %1"}},
```
antiagainst: space ;)

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
...
+          },
+        },
+      },
+      { // Analyze two instrutions that have same result id. The def use info
```
antiagainst: Nice!

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
...
+          // uses
+          {{2, {"OpBranch %2"}}},
+        }
+      }
+      }));
+// clang-format on
+
+struct KillInstTestCase {
+  const char* before;
+  std::unordered_set<uint32_t> indices_inst_to_kill;
```
antiagainst: I feel `indices_for_insts_to_kill` is more clear for me.

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
...
+    manager.AnalyzeInstDefUse(&inst);
+  }
+
+  CheckDef(tc.expected_define_use, manager.id_to_defs());
+  CheckUse(tc.expected_define_use, manager.id_to_uses());
+}
+
+// clang-format off
+INSTANTIATE_TEST_CASE_P(
+    TestCase, AnalyzeInstDefUseTest,
```
antiagainst: Since we use `AnalyzeInstDefUse` for `AnalyzeDefUse` internally and the later already gets fairly good coverage, I'm OK with this relatively small test suite for `AnalyzeInstDefUseTest`.

But I think we should add some cases for a `DefUseManager` initialized from a non-`nullptr` module with its `AnalyzeInstDefUse` subsequently?

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
...
+        "%4 = OpLabel "
+        "     OpBranch %5 "
+        "%5 = OpLabel "
+        "     OpBranch %6 "
+        "%6 = OpLabel "
+        "     OpBranch %4 "
+        "%7 = OpLabel "
+        "     OpReturn "
+        "     OpFunctionEnd",
+        {0, 1, 3, 5, 7},
```
antiagainst: I'd suggest to keep some of the id-generating insts to increase the diversity of this case.

test/opt/test_def_use.cpp
```diff
@@ -1151,4 +1152,221 @@ TEST(DefUseTest, OpSwitch) {
...
+        "%4 = OpLabel "
+        "     OpBranch %5 "
+        "%5 = OpLabel "
+        "     OpBranch %6 "
+        "%6 = OpLabel "
+        "     OpBranch %4 "
+        "%7 = OpLabel "
+        "     OpReturn "
+        "     OpFunctionEnd",
+        {2, 4, 6},
```
antiagainst: Also keep one, maybe.

antiagainst:
Done review.

The main concern from me is the lacking of test cases for `DefUseManager` initialized from non-`nullptr` module and with its `AnalyzeInstDefUse` called subsequently. The rest are just minor issues.

antiagainst:
Very nice code and tests thus far! :)

source/opt/def_use_manager.cpp
Qining: Fixed in a15622a

source/opt/def_use_manager.h
Qining: Fixed in a15622a

source/opt/def_use_manager.h
Qining: Fixed in a15622a

source/opt/def_use_manager.cpp
Qining: Cleardef() is removed in a15622a

source/opt/def_use_manager.cpp
Qining: Fixed in a15622a

source/opt/def_use_manager.cpp
Qining: Changed to `inst_to_used_ids_[inst] = {}`.

This is needed because we need to create entry even when the instruction does not have any operands. `inst_to_uses_ids_` is also used a set of analyzed instructions

source/opt/def_use_manager.cpp
Qining: Removed in a15622a

source/opt/def_use_manager.h
Qining: Updated the blurbs for both KillInst() and KillDef()

test/opt/test_def_use.cpp
Qining: Fixed in a15622a

test/opt/test_def_use.cpp
Qining: unnecessary comment removed in a15622a

test/opt/test_def_use.cpp
Qining: This the formatting result from clang-format. Probably we should disable it here?

test/opt/test_def_use.cpp
Qining: Fixed in a15622a

test/opt/test_def_use.cpp
Qining: Fixed in a15622a

test/opt/test_def_use.cpp
Qining: New tests for "analyzing additional instructions with an existing module" have been added in a15622a

test/opt/test_def_use.cpp
Qining: It does work.

source/opt/def_use_manager.h
Qining: Updated the blurb in daa199b

test/opt/test_def_use.cpp
```diff
@@ -1322,9 +1368,9 @@ INSTANTIATE_TEST_CASE_P(
         "OpFunctionEnd",
         {
           // no defs
```
antiagainst: Comments need to be updated.

test/opt/test_def_use.cpp
```diff
@@ -1180,9 +1180,13 @@ ir::Instruction BranchInstruction(uint32_t target_id) {
                          });
 }
 
+// Test case for analyzing individual instructions.
 struct AnalyzeInstDefUseTestCase {
   std::vector<ir::Instruction> insts;  // instrutions to be analyzed in order.
-  InstDefUse expected_define_use;      // expected analysis result.
+  const char*
+      module_text;  // If not empty, the module will be analyzed before above
```
antiagainst: No need to mention `If not empty` since we still assemble it even empty.

test/opt/test_def_use.cpp
Qining: Oops, forgot to update the comment, fixed in c7f6a94

test/opt/test_def_use.cpp
Qining: Removed in c7f6a94. It is really nice that we can assemble then disassemble an empty module

source/opt/def_use_manager.cpp
```diff
@@ -95,12 +106,21 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
...
+      } else if (it->inst->type_id() == 0) {
+        assert(false &&
+               "Result type id considered as using while the instruction "
+               "doesn't have a result type id.");
+      } else {
+        assert(false && "Trying Setting the result id which is immutable.");
+      }
     } else {
+      // Update an in-operand.
       assert(it->operand_index >= type_result_id_count &&
```
antiagainst: This `assert` is not necessary anymore.

Qining: Fixed in 28abd03

source/opt/def_use_manager.h
```diff
@@ -95,34 +94,31 @@ class DefUseManager {
...
-
   // Clear the internal def-use record of the given instruction |inst|. This
   // method will update the use information of the operand ids of |inst|. The
   // record: |inst| uses an |id|, will be removed from the use records of |id|.
   // If |inst| defines an result id, the use record of this result id will also
   // be removed. Does nothing if |inst| was not analyzed before.
   void ClearInst(ir::Instruction* inst);
 
-  // Erases the records that a given instruction |inst| uses its operand ids.
+  // Erases the records that a given instruction uses its operand ids.
```
antiagainst: Generally we want to avoid mentioning implementation details in function blurbs. I feel that mentioning updates of `id_to_uses_` & `id_to_used_ids_` should be avoided. I think the name is fairly self-explaining and the original blurb is actually good.

test/opt/test_def_use.cpp
antiagainst: Well, weird and ugly... Anyway, I won't argue with the tool. ;)

source/opt/def_use_manager.h
Qining: Hmm, changed back to the original blurb in 919f627

antiagainst:
LGTM. Thanks!

PR #358, opened 2016-08-17
Get the width from integer and float type

Qining:


source/opt/types.h
```diff
@@ -123,14 +123,15 @@ class Type {
...
-  Integer(uint32_t width, bool is_signed) : width_(width), signed_(is_signed) {}
+  Integer(uint32_t w, bool is_signed) : width_(w), signed_(is_signed) {}
   Integer(const Integer&) = default;
 
   bool IsSame(Type* that) const override;
   std::string str() const override;
 
   Integer* AsInteger() override { return this; }
   const Integer* AsInteger() const override { return this; }
+  uint32_t width() const {return width_;}
```
antiagainst: clang-format

test/opt/test_types.cpp
```diff
@@ -218,4 +218,20 @@ TEST(Types, AllTypes) {
   }
 }
 
+#define TestWidthOfType(ty, ...)                                      \
+  TEST(Types, TypeWidthOf##ty) {                                      \
+    std::vector<uint32_t> widths = {1, 2, 4, 8, 16, 32, 48, 64, 128}; \
+    std::vector<std::unique_ptr<ty>> types;                           \
+    for (uint32_t w : widths) {                                       \
+      types.emplace_back(new ty(w, ##__VA_ARGS__));                   \
```
antiagainst: Ha, I'm afraid this `##__VA_ARGS__` trick suffers on some old VS compilers? (Although we only test >= VS2013 in buildbot, it's still better to be able to compile on older compilers.)

I'm not opposed to duplicate the code for Integer & Float anyway, it's just two cases.

antiagainst: http://stackoverflow.com/questions/5588855/standard-alternative-to-gccs-va-args-trick

Qining:
All comments fixed in 6c5bdc9

antiagainst:
LGTM

PR #360, opened 2016-08-18
Unify constants pass

Qining:
Unify the constant uses in SPIR-V module. If two constants are defined exactly
the same, only one of them will be kept and all the uses of the removed
constant will be redirected to the kept one.

### DIFF
--- a/source/opt/unify_const_pass.cpp
+++ b/source/opt/unify_const_pass.cpp
@@ -1,0 +1,183 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#include "unify_const_pass.h"
+
+#include <unordered_map>
+#include <utility>
+
+#include "def_use_manager.h"
+#include "make_unique.h"
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+
+// The trie that stores a bunch of result ids and, for a given instruction,
+// searches the result id that has been defined with the same opcode, type and
+// operands.
+class ResultIdTrie {
+ public:
+  ResultIdTrie() : root_(new Node) {}
+
+  // For a given instruction, extracts its opcode, type id and operand words
+  // as an array of keys, looks up the trie to find a result id which is stored
+  // with the same opcode, type id and operand words. If none of such result id
+  // is found, creates a trie node with those keys, stores the instruction's
+  // result id and returns that result id. If an existing result id is found,
+  // returns the existing result id.
+  uint32_t LookupEquivalentResultFor(const ir::Instruction& inst) {
+    auto keys = GetLookUpKeys(inst);
+    auto* node = root_.get();
+    for (uint32_t key : keys) {
+      node = node->GetOrCreateTrieNodeFor(key);
+    }
+    if (node->result_id() == 0) {
+      node->SetResultId(inst.result_id());
+    }
+    return node->result_id();
+  }
+
+ private:
+  // The trie node to store result ids.
+  class Node {
+   public:
+    using TrieNodeMap = std::unordered_map<uint32_t, std::unique_ptr<Node>>;
+
+    Node() : result_id_(0), next_() {}
+    uint32_t result_id() const { return result_id_; }
+
+    // Sets the result id stored in this node.
+    void SetResultId(uint32_t id) { result_id_ = id; }
+
+    // Searches for the child trie node with the given key. If the node is
+    // found, returns that node. Otherwise creates an empty child node with
+    // that key and returns that newly created node.
+    Node* GetOrCreateTrieNodeFor(uint32_t key) {
+      auto iter = next_.find(key);
+      if (iter == next_.end()) {
+        // insert a new node and return the node.
+        return next_.insert(std::make_pair(key, MakeUnique<Node>()))
+            .first->second.get();
+      }
+      return iter->second.get();
+    }
+
+   private:
+    // The result id stored in this node. 0 means this node is empty.
+    uint32_t result_id_;
+    // The mapping from the keys to the child nodes of this node.
+    TrieNodeMap next_;
+  };
+
+  // Returns a vector of the opcode followed by the words in the raw SPIR-V
+  // instruction encoding but without the result id.
+  std::vector<uint32_t> GetLookUpKeys(const ir::Instruction& inst) {
+    std::vector<uint32_t> keys;
+    // Need to use the opcode, otherwise there might be a conflict with the
+    // following case when <op>'s binary value equals xx's id:
+    //  OpSpecConstantOp tt <op> yy zz
+    //  OpSpecConstantComposite tt xx yy zz;
+    keys.push_back(static_cast<uint32_t>(inst.opcode()));
+    for (const auto& operand : inst) {
+      if (operand.type == SPV_OPERAND_TYPE_RESULT_ID) continue;
+      keys.insert(keys.end(), operand.words.cbegin(), operand.words.cend());
+    }
+    return keys;
+  }
+
+  std::unique_ptr<Node> root_;  // The root node of the trie.
+};
+}  // anonymous namespace
+
+bool UnifyConstantPass::Process(ir::Module* module) {
+  bool modified = false;
+  ResultIdTrie defined_constants;
+  analysis::DefUseManager def_use_mgr(module);
+
+  for (ir::Instruction& inst : module->types_values()) {
+    // Do not handle the instruction when there are decorations upon the result
+    // id.
+    if (def_use_mgr.GetAnnotations(inst.result_id()).size() != 0) {
+      continue;
+    }
+
+    // The overall algorithm is to store the result ids of all the eligible
+    // constants encountered so far in a trie. For a constant defining
+    // instruction under consideration, use its opcode, result type id and
+    // words in operands as an array of keys to lookup the trie. If a result id
+    // can be found for that array of keys, a constant with exactly the same
+    // value must has been defined before, the constant under processing
+    // should be replaced by the constant previously defined. If no such result
+    // id can be found for that array of keys, this must be the first time a
+    // constant with its value be defined, we then create a new trie node to
+    // store the result id with the keys. When replacing a duplicated constant
+    // with a previously defined constant, all the uses of the duplicated
+    // constant, which must be placed after the duplicated constant defining
+    // instruction, will be updated. This way, the descendants of the
+    // previously defined constant and the duplicated constant will both refer
+    // to the previously defined constant. So that the operand ids which are
+    // used in key arrays will be the ids of the unified constants, when
+    // processing is up to a descendant. This makes comparing the key array
+    // always valid for judging duplication.
+    switch (inst.opcode()) {
+      case SpvOp::SpvOpConstantTrue:
+      case SpvOp::SpvOpConstantFalse:
+      case SpvOp::SpvOpConstant:
+      case SpvOp::SpvOpConstantNull:
+      case SpvOp::SpvOpConstantSampler:
+      case SpvOp::SpvOpConstantComposite:
+      // Only spec constants defined with OpSpecConstantOp and
+      // OpSpecConstantComposite should be processed in this pass. Spec
+      // constants defined with OpSpecConstant{|True|False} are decorated with
+      // 'SpecId' decoration and all of them should be treated as unique.
+      // 'SpecId' is not applicable to SpecConstants defined with
+      // OpSpecConstant{Op|Composite}, their values are not necessary to be
+      // unique. When all the operands/compoents are the same between two
+      // OpSpecConstant{Op|Composite} results, their result values must be the
+      // same so are unifiable.
+      case SpvOp::SpvOpSpecConstantOp:
+      case SpvOp::SpvOpSpecConstantComposite: {
+        uint32_t id = defined_constants.LookupEquivalentResultFor(inst);
+        if (id != inst.result_id()) {
+          // The constant is a duplicated one, use the cached constant to
+          // replace the uses of this duplicated one, then turn it to nop.
+          def_use_mgr.ReplaceAllUsesWith(inst.result_id(), id);
+          def_use_mgr.KillInst(&inst);
+          modified = true;
+        }
+        break;
+      }
+      default:
+        break;
+    }
+  }
+  return modified;
+}
+
+}  // opt
+}  // namespace spvtools
--- a/source/opt/unify_const_pass.h
+++ b/source/opt/unify_const_pass.h
@@ -24,16 +24,35 @@
 // TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 // MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
-#ifndef LIBSPIRV_OPT_PASSES_H_
-#define LIBSPIRV_OPT_PASSES_H_
+#ifndef LIBSPIRV_OPT_UNIFY_CONSTANT_PASS_H_
+#define LIBSPIRV_OPT_UNIFY_CONSTANT_PASS_H_
 
-// A single header to include all passes.
+#include "module.h"
+#include "pass.h"
 
-#include "eliminate_dead_constant_pass.h"
-#include "fold_spec_constant_op_and_composite_pass.h"
-#include "freeze_spec_constant_value_pass.h"
-#include "null_pass.h"
-#include "strip_debug_info_pass.h"
-#include "unify_const_pass.h"
+namespace spvtools {
+namespace opt {
 
-#endif  // LIBSPIRV_OPT_PASSES_H_
+// The optimization pass to de-duplicate the constants. Constants with exactly
+// same values and identical form will be unified and only one constant will be
+// kept for each unique pair of type and value.
+// There are several cases not handled by this pass:
+//  1) Constants defined by OpConstantNull instructions (null constants) and
+//  constants defined by OpConstantFalse, OpConstant or OpConstantComposite
+//  with value(s) 0 (zero-valued normal constants) are not considered
+//  equivalent. So null constants won't be used to replace zero-valued normal
+//  constants, and other constants won't replace the null constants either.
+//  2) Whenever there are decorations to the constant's result id or its type
+//  id, the constants won't be handled, which means, it won't be used to
+//  replace any other constants, neither can other constants replace it.
+//  3) NaN in float point format with different bit patterns are not unified.
+class UnifyConstantPass : public Pass {
+ public:
+  const char* name() const override { return "unify-const"; }
+  bool Process(ir::Module*) override;
+};
+
+}  // namespace opt
+}  // namespace spvtools
+
+#endif // LIBSPIRV_OPT_UNIFY_CONSTANT_PASS_H_
--- a/test/opt/test_unify_const.cpp
+++ b/test/opt/test_unify_const.cpp
@@ -1,0 +1,992 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and/or associated documentation files (the
+// "Materials"), to deal in the Materials without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Materials, and to
+// permit persons to whom the Materials are furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be included
+// in all copies or substantial portions of the Materials.
+//
+// MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+// KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+// SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+//    https://www.khronos.org/registry/
+//
+// THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+// IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+// CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+// TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#include <unordered_set>
+
+#include "assembly_builder.h"
+#include "pass_fixture.h"
+#include "pass_utils.h"
+
+namespace {
+using namespace spvtools;
+
+// Returns the types defining instructions commonly used in many tests.
+std::vector<std::string> CommonTypes() {
+  return std::vector<std::string>{
+      // clang-format off
+    // scalar types
+    "%bool = OpTypeBool",
+    "%uint = OpTypeInt 32 0",
+    "%int = OpTypeInt 32 1",
+    "%uint64 = OpTypeInt 64 0",
+    "%int64 = OpTypeInt 64 1",
+    "%float = OpTypeFloat 32",
+    "%double = OpTypeFloat 64",
+    // vector types
+    "%v2bool = OpTypeVector %bool 2",
+    "%v2uint = OpTypeVector %uint 2",
+    "%v2int = OpTypeVector %int 2",
+    "%v3int = OpTypeVector %int 3",
+    "%v4int = OpTypeVector %int 4",
+    "%v2float = OpTypeVector %float 2",
+    "%v3float = OpTypeVector %float 3",
+    "%v2double = OpTypeVector %double 2",
+    // struct types
+    "%inner_struct = OpTypeStruct %bool %float",
+    "%outer_struct = OpTypeStruct %inner_struct %int %double",
+    "%flat_struct = OpTypeStruct %bool %int %float %double",
+    // variable pointer types
+    "%_pf_bool = OpTypePointer Function %bool",
+    "%_pf_uint = OpTypePointer Function %uint",
+    "%_pf_int = OpTypePointer Function %int",
+    "%_pf_uint64 = OpTypePointer Function %uint64",
+    "%_pf_int64 = OpTypePointer Function %int64",
+    "%_pf_float = OpTypePointer Function %float",
+    "%_pf_double = OpTypePointer Function %double",
+    "%_pf_v2int = OpTypePointer Function %v2int",
+    "%_pf_v3int = OpTypePointer Function %v3int",
+    "%_pf_v4int = OpTypePointer Function %v4int",
+    "%_pf_v2float = OpTypePointer Function %v2float",
+    "%_pf_v3float = OpTypePointer Function %v3float",
+    "%_pf_v2double = OpTypePointer Function %v2double",
+    "%_pf_inner_struct = OpTypePointer Function %inner_struct",
+    "%_pf_outer_struct = OpTypePointer Function %outer_struct",
+    "%_pf_flat_struct = OpTypePointer Function %flat_struct",
+      // clang-format on
+  };
+}
+
+// A helper function to strip OpName instructions from the given string of
+// disassembly code and put those debug instructions to a set. Returns the
+// string with all OpName instruction stripped and a set of OpName
+// instructions.
+std::tuple<std::string, std::unordered_set<std::string>>
+StripOpNameInstructionsToSet(const std::string& str) {
+  std::stringstream ss(str);
+  std::ostringstream oss;
+  std::string inst_str;
+  std::unordered_set<std::string> opname_instructions;
+  while (std::getline(ss, inst_str, '\n')) {
+    if (inst_str.find("OpName %") == std::string::npos) {
+      oss << inst_str << '\n';
+    } else {
+      opname_instructions.insert(inst_str);
+    }
+  }
+  return std::make_tuple(oss.str(), std::move(opname_instructions));
+}
+
+// The test fixture for all tests of UnifyConstantPass. This fixture defines
+// the rule of checking: all the optimized code should be exactly the same as
+// the expected code, except the OpName instructions, which can be different in
+// order.
+template <typename T>
+class UnifyConstantTest : public PassTest<T> {
+ protected:
+  // Runs UnifyConstantPass on the code built from the given |test_builder|,
+  // and checks whether the optimization result matches with the code built
+  // from |expected_builder|.
+  void Check(const AssemblyBuilder& expected_builder,
+             const AssemblyBuilder& test_builder) {
+    // unoptimized code
+    const std::string original_before_strip = test_builder.GetCode();
+    std::string original_without_opnames;
+    std::unordered_set<std::string> original_opnames;
+    std::tie(original_without_opnames, original_opnames) =
+        StripOpNameInstructionsToSet(original_before_strip);
+
+    // expected code
+    std::string expected_without_opnames;
+    std::unordered_set<std::string> expected_opnames;
+    std::tie(expected_without_opnames, expected_opnames) =
+        StripOpNameInstructionsToSet(expected_builder.GetCode());
+
+    // optimized code
+    std::string optimized_before_strip;
+    bool modified = false;
+    std::tie(optimized_before_strip, modified) =
+        this->template SinglePassRunAndDisassemble<opt::UnifyConstantPass>(
+            test_builder.GetCode(),
+            /* skip_nop = */ true);
+    std::string optimized_without_opnames;
+    std::unordered_set<std::string> optimized_opnames;
+    std::tie(optimized_without_opnames, optimized_opnames) =
+        StripOpNameInstructionsToSet(optimized_before_strip);
+
+    // Flag "modified" should be returned correctly.
+    EXPECT_EQ(expected_without_opnames != original_without_opnames, modified);
+    // Code except OpName instructions should be exactly the same.
+    EXPECT_EQ(expected_without_opnames, optimized_without_opnames);
+    // OpName instructions can be in different order, but the content must be
+    // the same.
+    EXPECT_EQ(expected_opnames, optimized_opnames);
+  }
+};
+
+using UnifyFrontEndConstantSingleTest =
+    UnifyConstantTest<PassTest<::testing::Test>>;
+
+TEST_F(UnifyFrontEndConstantSingleTest, Basic) {
+  AssemblyBuilder test_builder;
+  AssemblyBuilder expected_builder;
+
+  test_builder
+      .AppendTypesConstantsGlobals({
+          "%uint = OpTypeInt 32 0", "%_pf_uint = OpTypePointer Function %uint",
+          "%unsigned_1 = OpConstant %uint 1",
+          "%unsigned_1_duplicate = OpConstant %uint 1",  // duplicated constant
+      })
+      .AppendInMain({
+          "%uint_var = OpVariable %_pf_uint Function",
+          "OpStore %uint_var %unsigned_1_duplicate",
+      });
+
+  expected_builder
+      .AppendTypesConstantsGlobals({
+          "%uint = OpTypeInt 32 0", "%_pf_uint = OpTypePointer Function %uint",
+          "%unsigned_1 = OpConstant %uint 1",
+      })
+      .AppendInMain({
+          "%uint_var = OpVariable %_pf_uint Function",
+          "OpStore %uint_var %unsigned_1",
+      })
+      .AppendNames({
+          "OpName %unsigned_1 \"unsigned_1_duplicate\"",  // the OpName
+                                                          // instruction of the
+                                                          // removed duplicated
+                                                          // constant won't be
+                                                          // erased.
+      });
+  Check(expected_builder, test_builder);
+}
+
+TEST_F(UnifyFrontEndConstantSingleTest, SkipWhenResultIdHasDecorations) {
+  AssemblyBuilder test_builder;
+  AssemblyBuilder expected_builder;
+
+  test_builder
+      .AppendAnnotations({
+          // So far we don't have valid decorations for constants. This is
+          // preparing for the future updates of SPIR-V.
+          // TODO(qining): change to a valid decoration once they are available.
+          "OpDecorate %f_1 RelaxedPrecision",
+          "OpDecorate %f_2_dup RelaxedPrecision",
+      })
+      .AppendTypesConstantsGlobals({
+          // clang-format off
+          "%float = OpTypeFloat 32",
+          "%_pf_float = OpTypePointer Function %float",
+          "%f_1 = OpConstant %float 1",
+          // %f_1 has decoration, so %f_1 will not be used to replace %f_1_dup.
+          "%f_1_dup = OpConstant %float 1",
+          "%f_2 = OpConstant %float 2",
+          // %_2_dup has decoration, so %f_2 will not replace %f_2_dup.
+          "%f_2_dup = OpConstant %float 2",
+          // no decoration for %f_3 or %f_3_dup, %f_3_dup should be replaced.
+          "%f_3 = OpConstant %float 3",
+          "%f_3_dup = OpConstant %float 3",
+          // clang-format on
+      })
+      .AppendInMain({
+          // clang-format off
+          "%f_var = OpVariable %_pf_float Function",
+          "OpStore %f_var %f_1_dup",
+          "OpStore %f_var %f_2_dup",
+          "OpStore %f_var %f_3_dup",
+          // clang-format on
+      });
+
+  expected_builder
+    .AppendAnnotations({
+          "OpDecorate %f_1 RelaxedPrecision",
+          "OpDecorate %f_2_dup RelaxedPrecision",
+        })
+  .AppendTypesConstantsGlobals({
+          // clang-format off
+          "%float = OpTypeFloat 32",
+          "%_pf_float = OpTypePointer Function %float",
+          "%f_1 = OpConstant %float 1",
+          "%f_1_dup = OpConstant %float 1",
+          "%f_2 = OpConstant %float 2",
+          "%f_2_dup = OpConstant %float 2",
+          "%f_3 = OpConstant %float 3",
+          // clang-format on
+      })
+  .AppendInMain({
+          // clang-format off
+          "%f_var = OpVariable %_pf_float Function",
+          "OpStore %f_var %f_1_dup",
+          "OpStore %f_var %f_2_dup",
+          "OpStore %f_var %f_3",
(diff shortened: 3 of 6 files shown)


source/opt/passes.h
```diff
@@ -74,6 +75,15 @@ class EliminateDeadConstantPass : public Pass {
   bool Process(ir::Module*) override;
 };
 
+// The optimization pass to deduplicate the constants. Front-end constants with
+// same value will be unified and only one constants will be kept for each
```
antiagainst: A little bit not clear for me here is about the same floating point values but with different bit patterns. Are we also unifying them?

source/opt/unify_const.cpp
```diff
@@ -0,0 +1,169 @@
...
+    result |= operand.words[i] << (i * 32);
+  }
+  return result;
+}
+
+// A helper function to ease trie node looking up. Returns a trie node pointer
+// pointing to the node identified by the given key in the given trie node map.
+// If there is no such node for the given key, a new trie node will be built
+// and then a pointer to the new built node will be returned.
+ResultIdTrieNode* GetOrCreateIteratorFor(
```
antiagainst: It looks like this function should be a member method of `class ResultIdTrieNode`.

source/opt/unify_const.cpp
```diff
@@ -0,0 +1,169 @@
...
+    keys.push_back(GetOperandValueInUint64(inst.GetOperand(i)));
+  }
+  return keys;
+}
+
+// Tries to find the same constant defining instruction of the given |inst| in
+// the trie pointed by the given trie node |node|. If a same instruction is
+// found in the trie, returns its result id. Otherwise, creates an trie node
+// for the given instruction then returns its result id.
+uint32_t LookupOrInsert(ResultIdTrieNode* node, const ir::Instruction& inst) {
```
antiagainst: It looks like this function should be a member method of `class ResultIdTrieNode`.

source/opt/unify_const.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+// Using a trie node to store encountered constants and look up later with a
+// given constant defining instruction's opcode, result type id and operands.
+// Each ResultIdTrieNode stores a result id, if it does represent an
+// encountered constant. ResultIdTrieNode also guide the lookup process by
+// holds the map toward the next ResultIdTrieNode to check with.
+struct ResultIdTrieNode {
+  ResultIdTrieNode() : result_id(0), next() {}
+  uint32_t result_id;
+  std::unordered_map<uint64_t, std::unique_ptr<ResultIdTrieNode>> next;
```
antiagainst: Should probability `using` a short name for this type.

source/opt/unify_const.cpp
```diff
@@ -0,0 +1,169 @@
...
+};
+
+// A helper function to extract the operand's raw value in 64-bit unsigned
+// integers. Note that this does not support operands longer than 64-bit.
+uint64_t GetOperandValueInUint64(const ir::Operand& operand) {
```
antiagainst: It seems we should put a little more functionalities in `struct Operand` to let we get it directly as 32-bit or 64-bit values. So we don't need this function anymore.

source/opt/unify_const.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+bool UnifyConstantPass::Process(ir::Module* module) {
+  bool modified = false;
+  ResultIdTrieNode defined_constants;
+  std::unordered_set<ir::Instruction*> duplicated_cosnts;
+  analysis::DefUseManager def_use_mgr(module);
+
+  for (ir::Instruction& inst : module->types_values()) {
+    switch (inst.opcode()) {
+      case SpvOp::SpvOpConstantTrue:
```
antiagainst: It seems we should put some comments to explain why `OpSpecConstant{|True|False}` is missing here.

antiagainst: Add also mention that `SpecId` should not applied to `SpecConstant{Composite|Op}`, so it's ok to hash them here.

antiagainst:
Cases to cover:

* `OpConstantNull %int` can generate the same value as `OpConstant %int 0`. They should be unified if possible.
* Decorations can be different for constants with the same value. Should not ignore them.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,533 @@
...
+// MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+
+#include "assembly_builder.h"
+#include "pass_fixture.h"
+#include "pass_utils.h"
+
+namespace {
+using namespace spvtools;
+
+const std::vector<std::string> CommonTypes = {
```
antiagainst: Should probably have a common header for this. I see it appears in multiple files.

source/opt/passes.h
```diff
@@ -28,6 +28,7 @@
 #define LIBSPIRV_OPT_PASSES_H_
 
 #include <memory>
+#include <unordered_map>
```
antiagainst: This can be moved to the cpp file since it's not used in this header file.

source/opt/passes.h
antiagainst: `... deduplicate constants. Normal constants of the same value (with identical bit pattern) ...`

Also should mention the effects of `Decoration`s.

source/opt/unify_const.cpp
```diff
@@ -0,0 +1,169 @@
...
+  uint64_t result = 0;
+  for (size_t i = 0; i < operand.words.size(); i++) {
+    result |= operand.words[i] << (i * 32);
+  }
+  return result;
+}
+
+// A helper function to ease trie node looking up. Returns a trie node pointer
+// pointing to the node identified by the given key in the given trie node map.
+// If there is no such node for the given key, a new trie node will be built
```
antiagainst: `no existing node for |key|, creates one and returns its pointer.`

source/opt/unify_const.cpp
```diff
@@ -0,0 +1,169 @@
...
+// and then a pointer to the new built node will be returned.
+ResultIdTrieNode* GetOrCreateIteratorFor(
+    uint64_t key,
+    std::unordered_map<uint64_t, std::unique_ptr<ResultIdTrieNode>>* m) {
+  auto iter = m->find(key);
+  if (iter == m->end()) {
+    // insert new entry
+    (*m)[key] = std::unique_ptr<ResultIdTrieNode>(new ResultIdTrieNode);
```
antiagainst: Can use `return m->insert(...).first` I think

source/opt/unify_const.cpp
```diff
@@ -0,0 +1,169 @@
...
+    iter = m->find(key);
+  }
+  return iter->second.get();
+}
+
+// A helper function to construct the keys used to identify a constant. The
+// keys are consisent of 1) opcode, 2) result type id, 3) all other operands.
```
antiagainst: `keys are composed of`?

source/opt/unify_const.cpp
Qining: Fixed in 5d7fa1e

source/opt/passes.h
Qining: Float point values with different bit patterns are considered different.

Updated in 5d7fa1e
Now, constants with any decorations on their result ids or type ids won't be processed any more. They won't be used to replace other constants, and other constants won't replace them neither (even though the decorations may be the same).

source/opt/unify_const.cpp
Qining: Fixed in 5d7fa1e

source/opt/unify_const.cpp
Qining: Updated in 5d7fa1e.

Actually we should not need to care the length of an operand at all....

test/opt/test_unify_const.cpp
Qining: Probably will do that in another CL, such a change will affect too many tests.

source/opt/passes.h
Qining: fixed in 5d7fa1e

source/opt/unify_const.cpp
Qining: Function updated in 5d7fa1e. Blurbs updated

source/opt/unify_const.cpp
Qining: Blurbs updated

source/opt/unify_const.cpp
Qining: updated in bc933e0

source/opt/unify_const.cpp
Qining: Blurbs added in bc933e0

source/opt/unify_const.cpp
Qining: Adopted in bc933e0

source/opt/unify_const_pass.h
```diff
@@ -0,0 +1,57 @@
...
+#define LIBSPIRV_OPT_UNIFY_CONSTANT_PASS_H_
+
+#include "module.h"
+#include "pass.h"
+
+namespace spvtools {
+namespace opt {
+
+// The optimization pass to de-duplicate the constants. Constants with exactly
+// same values will be unified and only one constant will be kept for each
```
antiagainst: with exactly same values and identical form?

source/opt/unify_const_pass.h
```diff
@@ -0,0 +1,57 @@
...
+// unique pair of type and value. There are several cases not handled by this
+// pass:
+//  1) Constants defined by OpConstantNull instructions (null constants) and
+//  constants defined by OpConstantFalse, OpConstant or OpConstantComposite
+//  with value(s) 0 (zero-valued normal constants) are not considered
+//  equivalent. So null constants won't be used to replace zero-valued normal
+//  constants, neither vice versa.
```
antiagainst: I think neither is redundant here.

source/opt/unify_const_pass.h
```diff
@@ -0,0 +1,57 @@
...
+//  2) Whenever there are decorations to the constant's result id or its type
+//  id, the constants won't be handled, which means, it won't be used to
+//  replace any other constants, neither other constants can replace it.
```
antiagainst: neither can ... ?

source/opt/unify_const_pass.h
```diff
@@ -0,0 +1,57 @@
...
+class UnifyConstantPass : public Pass {
+ public:
+  const char* name() const override { return "unify-const"; }
```
antiagainst: It's better to keep various names consistent. The class is named as `UnifyConstantPass`, so I'd suggest to return `unify-constant` here. Also the file name.

source/opt/unify_const_pass.h
```diff
@@ -0,0 +1,57 @@
...
+// The optimization pass to de-duplicate the constants. Constants with exactly
+// same values will be unified and only one constant will be kept for each
+// unique pair of type and value. There are several cases not handled by this
+// pass:
+//  1) Constants defined by OpConstantNull instructions (null constants) and
+//  constants defined by OpConstantFalse, OpConstant or OpConstantComposite
+//  with value(s) 0 (zero-valued normal constants) are not considered
+//  equivalent. So null constants won't be used to replace zero-valued normal
+//  constants, neither vice versa.
+//  2) Whenever there are decorations to the constant's result id or its type
```
antiagainst: Constants with any decoration on its result id or result type id will be skipped by this pass and won't be unified with any other constants.

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+  using TrieNodeMap =
+      std::unordered_map<uint32_t, std::unique_ptr<ResultIdTrieNode>>;
+
+  ResultIdTrieNode() : result_id_(0), next_() {}
+  uint32_t result_id() const { return result_id_; }
+
+  // Sets the result id stored in this node.
+  void SetResultId(uint32_t id) { result_id_ = id; }
+
+  // Searches for the child trie node with the given key. If the node is found,
```
antiagainst: children trie nodes

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+  // returns that node. Otherwise creates an empty child node with that key and
+  // returns that newly created node.
+  ResultIdTrieNode* GetOrCreateTrieNodeFor(uint32_t key) {
+    auto iter = next_.find(key);
+    if (iter == next_.end()) {
+      // insert new entry
+      return next_.insert(std::make_pair(key, MakeUnique<ResultIdTrieNode>()))
```
antiagainst: `iter = next_.insert(...)`?

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include "def_use_manager.h"
+#include "make_unique.h"
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+
+// The trie node to store result ids.
+class ResultIdTrieNode {
```
antiagainst: Is it possible to put this class inside the `ResultIdTrie` and name it just as `Node` maybe?

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+// looking up with a given instruction.
+class ResultIdTrie {
+ public:
+  ResultIdTrie() : root_(new ResultIdTrieNode) {}
+
+  // For a given instruction, extracts its opcode, type id and operand words
+  // as an array of keys, looks up the trie to find an result id which is
+  // stored with the same opcode, type id and operand words. If none of such
+  // result id is found, creates a trie node with those keys, stores the
+  // instruction's result id and returns that result id. If a existing result
```
antiagainst: an existing

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+};
+
+// The trie that stores a bunch of result ids and performs equivalent result id
+// looking up with a given instruction.
+class ResultIdTrie {
+ public:
+  ResultIdTrie() : root_(new ResultIdTrieNode) {}
+
+  // For a given instruction, extracts its opcode, type id and operand words
+  // as an array of keys, looks up the trie to find an result id which is
```
antiagainst: a result id

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  std::unique_ptr<ResultIdTrieNode> root_;  // The root node of the trie.
+};
+}  // anonymous namespace
+
+bool UnifyConstantPass::Process(ir::Module* module) {
+  bool modified = false;
+  ResultIdTrie defined_constants;
+  std::unordered_set<ir::Instruction*> duplicated_cosnts;
```
antiagainst: duplicated_constants

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+    // Do not handle the instruction when there are decorations upon either
+    // the result id or the type id.
+    if (def_use_mgr.GetAnnotations(inst.result_id()).size() != 0 ||
+        def_use_mgr.GetAnnotations(inst.type_id()).size() != 0) {
+      continue;
+    }
+
+    // The overall algorithm is to store the result ids of all the eligible
+    // constants encountered so far in a trie. For a comming constant defining
+    // instruction, use its opcode, result type id and operand words as an
```
antiagainst: in operands

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+    switch (inst.opcode()) {
+      case SpvOp::SpvOpConstantTrue:
+      case SpvOp::SpvOpConstantFalse:
+      case SpvOp::SpvOpConstant:
+      case SpvOp::SpvOpConstantNull:
+      case SpvOp::SpvOpConstantSampler:
+      case SpvOp::SpvOpConstantComposite:
+      // Only spec constants defined with OpSpecConstantOp and
+      // OpSpecConstantComposite should be processed in this pass. Spec
+      // constants defined with OpSpecConstant{|True|False|} are decorated with
```
antiagainst: The last bar in `{|True|False|}` is redundant

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+  for (ir::Instruction& inst : module->types_values()) {
+    // Do not handle the instruction when there are decorations upon either
+    // the result id or the type id.
+    if (def_use_mgr.GetAnnotations(inst.result_id()).size() != 0 ||
+        def_use_mgr.GetAnnotations(inst.type_id()).size() != 0) {
+      continue;
+    }
+
+    // The overall algorithm is to store the result ids of all the eligible
+    // constants encountered so far in a trie. For a comming constant defining
```
antiagainst: for the constant-defining instruction under consideration?

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+    // instruction, use its opcode, result type id and operand words as an
+    // array of keys to lookup the trie. If an result id can be found for that
+    // array of keys, a constant with exactly the same value(s) must has been
```
antiagainst: exactly the same value

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+    // defined before, the constant under processing should be replace by the
```
antiagainst: replaced

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+    // constant previously defined. If none result id can be found for that
```
antiagainst: no such

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+    // array of keys, this must be the first time a constant with its value
```
antiagainst: the first time a constant is being defined

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+    // be defined, we then create a new trie node to store the result id with
+    // the keys. When replacing a duplicated constant with a previously defined
+    // constant, all the uses of the duplicated constant, which must be placed
+    // after the duplicated constant defining instruction, will be updated.
+    // This way, the descendants of the previously defined constant and the
+    // duplicated constant will both refer to the previously defined constant.
+    // So that the operand ids that used in key arrays will always be ids of
```
antiagainst: ... will be ids of the unified constants, when processing is up to a descendant.

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,185 @@
...
+  bool modified = false;
+  ResultIdTrie defined_constants;
+  std::unordered_set<ir::Instruction*> duplicated_cosnts;
+  analysis::DefUseManager def_use_mgr(module);
+
+  for (ir::Instruction& inst : module->types_values()) {
+    // Do not handle the instruction when there are decorations upon either
+    // the result id or the type id.
+    if (def_use_mgr.GetAnnotations(inst.result_id()).size() != 0 ||
+        def_use_mgr.GetAnnotations(inst.type_id()).size() != 0) {
```
antiagainst: If type id is the same, the decorations on types should be the same. So it's fine to have decorations on types since we use the type id as a part of the key.

source/opt/unify_const_pass.cpp
Qining: Hmm, I got 'no viable overload =' if I do it that way.

source/opt/unify_const_pass.cpp
Qining: Fixed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.cpp
Qining: removed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.cpp
Qining: fixed in a013120

source/opt/unify_const_pass.h
Qining: fixed in a013120

source/opt/unify_const_pass.h
Qining: fixed in a013120

source/opt/unify_const_pass.h
Qining: updated in a013120

source/opt/unify_const_pass.h
Qining: hmm, I think it is better to explicitly explain the 'instruction' won't be considered at all. Anyway I've updated this in a013120

source/opt/unify_const_pass.cpp
Qining: moved and rename in f70318784

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,186 @@
...
+        std::unordered_map<uint32_t, std::unique_ptr<Node>>;
+
+    Node() : result_id_(0), next_() {}
+    uint32_t result_id() const { return result_id_; }
+
+    // Sets the result id stored in this node.
+    void SetResultId(uint32_t id) { result_id_ = id; }
+
+    // Searches for the child trie node with the given key. If the node is
+    // found,
```
antiagainst: format

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,186 @@
...
+      node->SetResultId(inst.result_id());
+    }
+    return node->result_id();
+  }
+
+ private:
+  // The trie node to store result ids.
+  class Node {
+   public:
+    using TrieNodeMap =
```
antiagainst: `#include <unordered_map>`

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,754 @@
...
+    } else {
+      opname_instructions.insert(inst_str);
+    }
+  }
+  return std::make_tuple(oss.str(), std::move(opname_instructions));
+}
+
+// The test fixture for all tests of UnifyConstantPass. This fixture setups
+// two assembly code builders, one for the expected code and another for
+// pre-optimized code. This fixture also defines the rule of checking: all the
```
antiagainst: unoptimized? Pre-optimized sounds like something already optimized before.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,754 @@
...
+            /* skip_nop = */ true);
+    std::string optimized_without_opnames;
+    std::unordered_set<std::string> optimized_opnames;
+    std::tie(optimized_without_opnames, optimized_opnames) =
+        StripOpNameInstructionsToSet(optimized_before_strip);
+
+    // Flag "modified" should be returned correctly.
+    EXPECT_EQ(expected_without_opnames != original_without_opnames, modified);
+    // Code except OpName instructions should be exactly the same.
+    EXPECT_EQ(optimized_without_opnames, expected_without_opnames);
```
antiagainst: expected should be placed as the first parameter, as a convention.

Qining: fixed in 48d55dc

test/opt/test_unify_const.cpp
Qining: fixed in 48d55dc

source/opt/unify_const_pass.cpp
Qining: fixed in 48d55dc.

unused unordered_set header is removed

source/opt/unify_const_pass.cpp
Qining: fixed in 48d55dc

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,754 @@
...
+      .AppendNames({
+          "OpName %float_1 \"float_1_duplicate\"",
+          "OpName %double_1 \"double_1_duplicate\"",
+      });
+  Check();
+}
+
+struct UnifyConstantTestCase {
+  // new constants that should be kept.
+  std::vector<std::string> new_consts;
```
antiagainst: Instead of `new_consts` and `use_new_consts`, what about `preserved_consts` and `use_preserved_consts`? I feel the later is more clear.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,754 @@
...
+  // expected uses of the new defined constants.
+  std::vector<std::string> use_new_consts;
+  // duplicated constants of the new constants.
+  std::vector<std::string> duplicate_consts;
```
antiagainst: duplicated.

source/opt/unify_const_pass.h
```diff
@@ -0,0 +1,57 @@
...
+// kept for each unique pair of type and value. There are several cases not
+// handled by this pass:
+//  1) Constants defined by OpConstantNull instructions (null constants) and
+//  constants defined by OpConstantFalse, OpConstant or OpConstantComposite
+//  with value(s) 0 (zero-valued normal constants) are not considered
+//  equivalent. So null constants won't be used to replace zero-valued normal
+//  constants, and other constants won't replace the null constants either.
+//  2) Whenever there are decorations to the constant's result id or its type
+//  id, the constants won't be handled, which means, it won't be used to
+//  replace any other constants, neither can other constants replace it.
```
antiagainst: Also worth a mention that NaN with different bit patterns are not unified.

test/opt/test_unify_const.cpp
antiagainst: Also remember to update all comments in the following.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,754 @@
...
+  // The updated OpName instructions of the duplicated constants.
+  std::vector<std::string> remapped_names;
+};
+
+using UnifyFrontEndConstantParamTest = UnifyConstantTest<
+    PassTest<::testing::TestWithParam<UnifyConstantTestCase>>>;
+
+TEST_P(UnifyFrontEndConstantParamTest, TestCase) {
+  auto& tc = GetParam();
+  expected_builder_.AppendTypesConstantsGlobals(tc.new_consts)
```
antiagainst: Worth a comment here saying `// duplicate constants are killed in the expected output.`

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,754 @@
...
+      .AppendInMain(tc.use_new_consts)
+      .AppendNames(tc.remapped_names);
```
antiagainst: Worth a comment here saying `// With their attached debug instruction migrated to the corresponding preserved constants.`

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,754 @@
...
+          // use new constants in main
+          {
+            "%vec_var = OpVariable %_pf_v4int Function",
+            "OpStore %vec_var %vec",
+          },
+          // duplicated constants
+          {
+            "%signed_1_duplicate = OpConstant %int 1",
+            "%signed_2_duplicate = OpConstant %int 2",
+            "%signed_3_duplicate = OpConstant %int 3",
```
antiagainst: It would be better to leave some elements the preserved one.

antiagainst:
Nice tests!

Would be even nicer with the following tests:

* `OpConstantNull` together normal `OpConstant` generating zeros
* NaN with different bit patterns
* `OpSpecConstant{|True|False}` with different SpecIds

None of the above should be unified.

test/opt/test_unify_const.cpp
Qining: modified in 00970f9

test/opt/test_unify_const.cpp
Qining: added in 00970f9

test/opt/test_unify_const.cpp
Qining: added in 00970f9

source/opt/unify_const_pass.h
Qining: added in 00970f9

test/opt/test_unify_const.cpp
Qining: updated in 00970f9

Qining:
Added tests in 00970f9:
* OpConstantNull together with normal OpConstant and OpConstantComposite
* NaN with different bit pattern
* SpecConstants with SpecId decorations

Qining:
rebased

source/opt/unify_const_pass.h
```diff
@@ -35,8 +35,9 @@ namespace opt {
 
 // The optimization pass to de-duplicate the constants. Constants with exactly
 // same values and identical form will be unified and only one constant will be
-// kept for each unique pair of type and value. There are several cases not
-// handled by this pass:
+// kept for each unique pair of type and value. Note that NaN in float point
```
antiagainst: This can be listed as another case like the following two.

test/opt/test_unify_const.cpp
```diff
@@ -238,16 +238,17 @@ TEST_F(UnifyFrontEndConstantSingleTest, SkipWhenResultIdHasDecorations) {
...
+  std::vector<std::string> use_preserved_consts;
+  // duplicated constants of the preserved constants.
   std::vector<std::string> duplicate_consts;
-  // uses of the duplicated constants, expected to be updated to use the new
-  // constants.
+  // uses of the duplicated constants, expected to be updated to use the
+  // preserved constants.
   std::vector<std::string> use_duplicate_consts;
-  // The updated OpName instructions of the duplicated constants.
+  // The updated OpName instructions that originally refers to duplicated
```
antiagainst: s/refers/refer/

test/opt/test_unify_const.cpp
```diff
@@ -256,22 +257,26 @@ using UnifyFrontEndConstantParamTest = UnifyConstantTest<
...
-      .AppendInMain(tc.use_new_consts)
-      .AppendNames(tc.remapped_names);
-  test_builder_.AppendTypesConstantsGlobals(tc.new_consts)
+
+  test_builder_.AppendTypesConstantsGlobals(tc.preserved_consts)
       .AppendTypesConstantsGlobals(tc.duplicate_consts)
       .AppendInMain(tc.use_duplicate_consts);
+
+  // Duplicated constants are killed in the expected output, and the debug
+  // instructions attached to thoe duplicated instructions will be migrated to
```
antiagainst: s/thoe/the/

test/opt/test_unify_const.cpp
```diff
@@ -518,9 +553,101 @@ INSTANTIATE_TEST_CASE_P(
...
+            "OpStore %bool_var %spec_bool_1",
+            "OpStore %bool_var %spec_bool_2",
+            "%int_var = OpVariable %_pf_int Function",
+            "OpStore %int_var %spec_int_1",
+            "OpStore %int_var %spec_int_2",
+          },
+          // duplicated constants. No duplicated instruction to remove in this
+          // case.
+          {},
+          // use duplicated constants in main. Same to the 'preserved constants
```
antiagainst: Same as

source/opt/unify_const_pass.h
Qining: added in c94cbada

test/opt/test_unify_const.cpp
Qining: fixed in c94cbda

test/opt/test_unify_const.cpp
Qining: fixed in a previous commit

test/opt/test_unify_const.cpp
Qining: fixed in c94cbda

antiagainst:
Cool! +1.

Qining:
@dneto0 PTAL

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,184 @@
...
+   private:
+    // The result id stored in this node. 0 means this node is empty.
+    uint32_t result_id_;
+    // The mapping from the keys to the child nodes of this node.
+    TrieNodeMap next_;
+  };
+
+  // Extracts the opcode, result type id and operand words from an instruction.
+  // Returns a vector of those values in 32-bit unsigned integers.
+  std::vector<uint32_t> GetLookUpKeys(const ir::Instruction& inst) {
```
dneto0: It's hilarious but this is exactly just the set of words in the raw SPIR-V encoding after the opcode.

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,184 @@
...
+    std::vector<uint32_t> keys;
+    // Need to use the opcode, otherwise there might be a conflict with the
+    // following case when <op>'s binary value equals xx's id:
+    //  OpSpecConstantOp tt <op> yy zz
+    //  OpSpecConstantComposite tt xx yy zz;
+    keys.push_back(static_cast<uint32_t>(inst.opcode()));
+    for (uint32_t i = 0; i < inst.NumOperands(); i++) {
```
dneto0: Prefer range-for here:
```
for (const auto& operand : inst) {
   ...
}
```

source/opt/unify_const_pass.cpp
dneto0: Wait, you *don't* include the result id.  So the comment is wrong.

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,184 @@
...
+#include "def_use_manager.h"
+#include "make_unique.h"
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+
+// The trie that stores a bunch of result ids and performs equivalent result id
+// looking up with a given instruction.
```
dneto0: Be careful about the notion of "equivalence" here.  For operations with side effects or which otherwise touch the outside world, then one is not collapsable into another. It's only "equivalent" for things that would be pure functions (only operate on their operands and don't have side effects).
Sometimes they're called cominbators.

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,184 @@
...
+  analysis::DefUseManager def_use_mgr(module);
+
+  for (ir::Instruction& inst : module->types_values()) {
+    // Do not handle the instruction when there are decorations upon the result
+    // id.
+    if (def_use_mgr.GetAnnotations(inst.result_id()).size() != 0) {
+      continue;
+    }
+
+    // The overall algorithm is to store the result ids of all the eligible
```
dneto0: Great comment.  Describes all the subtle points, including some I had not anticipated.

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,184 @@
...
+      // OpSpecConstant{Op|Composite} results, their result values must be the
+      // same so are unifiable.
+      case SpvOp::SpvOpSpecConstantOp:
+      case SpvOp::SpvOpSpecConstantComposite: {
+        uint32_t id = defined_constants.LookupEquivalentResultFor(inst);
+        if (id != inst.result_id()) {
+          // The constant is a duplicated one, use the cached constant to
+          // replace the uses of this duplicated one, then turn it to nop.
+          def_use_mgr.ReplaceAllUsesWith(inst.result_id(), id);
+          def_use_mgr.KillInst(&inst);
```
dneto0: A weird thing with any use of KillInst.  If it has debug line instructions in it, then mabye those should be transferred to the next instruction.  ?  
Or do that in the remove-nops operations?
For another CL.
Would like @antiagainst opinion on this.

dneto0: I should say that handling debug info during code optimizations is always going to be somewhat broken no matter what you do.

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+
+// The trie that stores a bunch of result ids and performs equivalent result id
+// looking up with a given instruction.
+class ResultIdTrie {
+ public:
```
dneto0: This is a really good use of a trie.

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,184 @@
...
+    }
+    return node->result_id();
+  }
+
+ private:
+  // The trie node to store result ids.
+  class Node {
+   public:
+    using TrieNodeMap =
+        std::unordered_map<uint32_t, std::unique_ptr<Node>>;
```
dneto0: I suspect a more specialized data structure would perform better.  But we have no tuning set, so let's not worry about it now. :-)

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+  return std::make_tuple(oss.str(), std::move(opname_instructions));
+}
+
+// The test fixture for all tests of UnifyConstantPass. This fixture setups
+// two assembly code builders, one for the expected code and another for
+// unoptimized code. This fixture also defines the rule of checking: all the
+// optimized code should be exactly the same as the expected code, except the
+// OpName instructions, which can be different in order.
+template <typename T>
+class UnifyConstantTest : public PassTest<T> {
```
dneto0: This suffers a bit from complexity.
See https://g3doc.corp.google.com/devtools/library_club/g3doc/totw/122.md?cl=head  (h/t to stefanus)
I'll keep reading.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+ protected:
+  virtual void SetUp() override {
```
dneto0: Avoid Setup in fixtures.  Just do the initialization inline in the test.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+    test_builder_.AppendTypesConstantsGlobals(CommonTypes());
+    expected_builder_.AppendTypesConstantsGlobals(CommonTypes());
+  }
+
+  // Checks the validity of the optimization result.
+  void Check() {
```
dneto0: Prefer free functions with arguments.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+          // clang-format off
+          "%bool_true = OpConstantTrue %bool",
+          "%signed_1 = OpConstant %int 1",
+          "%float_1 = OpConstant %float 1",
+          "%double_1 = OpConstant %double 1",
+          "%s = OpConstantComposite %flat_struct %bool_true %signed_1 %float_1 %double_1",
+          // The duplicated constants without decorations should not be skipped.
+          "%float_1_duplicate = OpConstant %float 1",
+          "%double_1_duplicate = OpConstant %double 1",
+          // The type "flat_struct" has decoration, so the following
```
dneto0: "flat_struct" isn't decorated. The result value is.  ?!

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+          "OpName %unsigned_1 \"unsigned_1_duplicate\"",  // the OpName
+                                                          // instruction of the
+                                                          // removed duplicated
+                                                          // constant won't be
+                                                          // erased.
+      });
+  Check();
+}
+
+TEST_F(UnifyFrontEndConstantSingleTest, SkipWhenResultIdHasDecorations) {
```
dneto0: I think you can make your point here with a simple float constant with RelaxedPrecision decoration in one case and nothing in the other.

You also need a test case where you have types differ, e.g. two structs with same members but one struct type has member decoration on its element.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+          // remapped names
+          {
+            "OpName %bool_null \"bool_null_duplicate\"",
+            "OpName %signed_null \"signed_null_duplicate\"",
+            "OpName %signed_64_null \"signed_64_null_duplicate\"",
+            "OpName %float_null \"float_null_duplicate\"",
+            "OpName %double_null \"double_null_duplicate\"",
+          },
+        },
+        // TODO(qining): support constant sampler
```
dneto0: If you don't test samplers, don't enable it in the code.

source/opt/unify_const_pass.cpp
Qining: Err.. actually the comment is `result type id`.... anyway, it will be updated soon

test/opt/test_unify_const.cpp
Qining: Having arguments is fine. But this function will need the `SinglePassRunAndCheck()` method provided by the PassTest fixture. If we still want to pack the checkers in one function, we will need it in a fixture.

However, the setup one will be removed in the next commit

test/opt/test_unify_const.cpp
Qining: Oops, stale comments....I should have removed them before.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+          },
+          // remapped names
+          {
+            "OpName %null_inner \"null_inner_dup\"",
+            "OpName %null_outer \"null_outer_dup\"",
+            "OpName %inner_zero \"inner_zero_dup\"",
+            "OpName %outer_zero \"outer_zero_dup\"",
+          },
+        },
+        // Spec Constants with SpecId decortion should be skipped.
```
dneto0: "decoration"

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+        {
+          // preserved constants
+          {
+            // Assembly builder will add OpDecorate SpecId instruction for the
+            // following spec constant instructions automatically.
```
dneto0: That automatic part means this test is a bit fragile....  If the assembly builder changes then this test may silently break.
 Avoid creeping complexity like this.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+            "%spec_signed_add_duplicate = OpSpecConstantOp %int IAdd %spec_signed_1 %spec_signed_2",
+          },
+          // use duplicated contants in main
+          {
+            "%int_var = OpVariable %_pf_int Function",
+            "OpStore %int_var %spec_signed_add_duplicate",
+          },
+          // remapped names
+          {
+            "OpName %spec_signed_add \"spec_signed_add_duplicate\"",
```
dneto0: Nice.

test/opt/test_unify_const.cpp
```diff
@@ -0,0 +1,880 @@
...
+          // preserved constants
+          {
+            "%array_size = OpConstant %int 4",
+            "%type_arr_int_4 = OpTypeArray %int %array_size",
+            "%signed_0 = OpConstant %int 100",
+            "%signed_1 = OpConstant %int 1",
+            "%signed_2 = OpSpecConstantOp %int IAdd %signed_0 %signed_1",
+            "%signed_3 = OpSpecConstantOp %int ISub %signed_0 %signed_2",
+            "%signed_4 = OpSpecConstantOp %int IAdd %signed_0 %signed_3",
+            "%signed_5 = OpSpecConstantOp %int ISub %signed_0 %signed_4",
```
dneto0: You had me at length 3...

dneto0:
-1   Some minor things, but they should be fixed.

test/opt/test_unify_const.cpp
Qining: fixed in e9d61bc

test/opt/test_unify_const.cpp
Qining: a test of Constant Sampler is added. Lucky we don't have Null Constant of this type.

test/opt/test_unify_const.cpp
Qining: Updated this part with two tests in e9d61bc.
1) float w/o decoration
2) struct w/o member decoration

test/opt/test_unify_const.cpp
Qining: removed in e9d61bc

test/opt/test_unify_const.cpp
Qining: updated with arguments in e9d61bc

source/opt/unify_const_pass.cpp
Qining: updated in e9d61bc. I think describing the notion of 'equivalent when not having side effect' sort of hard to describe here.....

source/opt/unify_const_pass.cpp
Qining: updated in e9d61bc

source/opt/unify_const_pass.cpp
antiagainst: Yes, agreed.

For newly added instructions, the existing line debug info do not apply at all. If the line-debug instruction's host instruction is killed, a meaningful approach is that still let the line-debug instruction cover whatever is following the original host instruction. Actually that's exactly what we are doing now: #391. (Thanks for fixing the `ForEachInst` and `ToNop` stuff! It's great to see this works out of the box at the moment.)

So I would suggest we just keep what's there right now.

source/opt/unify_const_pass.cpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+   private:
+    // The result id stored in this node. 0 means this node is empty.
+    uint32_t result_id_;
+    // The mapping from the keys to the child nodes of this node.
+    TrieNodeMap next_;
+  };
+
+  // Returns a vector of of words in the raw SPIR-V encoding but excluding the
```
dneto0: Not really.  The word count part of the opcode word is not represented. But it's redundant.  Just fix the comment to make it accurate.   (I had misread it the last time!)
Suggest:
Returns a vector of the opcode followed by the words in the raw SPIR-V instruction encoding but without the result ID.

dneto0:
-1 One last thing.

source/opt/unify_const_pass.cpp
Qining: Fixed in 7d42b5e

dneto0:
+2 Please rebase and squash

Qining:
squashed and rebased. All bot passed before squashing

PR #324, opened 2016-08-10
Fold spec constants to front-end constants

Qining:
For the spec constants defined by OpSpecConstantOp and
OpSpecContantComposite, if all of their operands are constants with
determined values (front-end constants), get the correct values of
the spec constants and re-define them as front-end constants.

In short, this pass replaces all the spec constants with front-end
constants when possible.

Note that this pass does not unify front-end constants. This means it is
possible to have multiple constants defined with same value, due to new
added constant instructions.

PR #361, opened 2016-08-19
Add a home brewed make_unique to ease creating unique_ptrs.

antiagainst:
@Qining PTAL.

Should also benefit your #324.


dneto0:
+2

PR #362, opened 2016-08-20
2. Add macros for error and status logging.

antiagainst:
Depends on #401.

Add the following macros for logging purpose:
* `SPIRV_ASSERT`
* `SPIRV_DEBUG`
* `SPIRV_UNIMPLEMENTED`
* `SPIRV_UNREACHABLE`

The last two is always turned on, while the first two can only
be turned on in debug build.

### DIFF
--- a/source/opt/log.h
+++ b/source/opt/log.h
@@ -1,0 +1,202 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SPIRV_TOOLS_LOG_H_
+#define SPIRV_TOOLS_LOG_H_
+
+#include <cstdio>
+#include <cstdlib>
+#include <utility>
+#include <vector>
+
+#include "message.h"
+
+// Asserts the given condition is true. Otherwise, send a message to the
+// consumer. Accepts the following formats:
+//
+// SPIRV_ASSERT(<message-consumer>, <condition-expression>);
+// SPIRV_ASSERT(<message-consumer>, <condition-expression>, <message>);
+// SPIRV_ASSERT(<message-consumer>, <condition-expression>,
+//              <message-format>,   <variable-arguments>);
+//
+// In the third format, the number of <variable-arguments> cannot exceed (5 -
+// 2). If more arguments are wanted, grow PP_ARG_N and PP_NARGS in the below.
+#if !defined(NDEBUG)
+#define SPIRV_ASSERT(consumer, ...) SPIRV_ASSERT_IMPL(consumer, __VA_ARGS__)
+#else
+#define SPIRV_ASSERT(consumer, ...)
+#endif
+
+// Logs a debug message to the consumer. Accepts the following formats:
+//
+// SPIRV_DEBUG(<message-consumer>, <message>);
+// SPIRV_DEBUG(<message-consumer>, <message-format>, <variable-arguments>);
+//
+// In the second format, the number of <variable-arguments> cannot exceed (5 -
+// 1). If more arguments are wanted, grow PP_ARG_N and PP_NARGS in the below.
+#if !defined(NDEBUG) && defined(SPIRV_LOG_DEBUG)
+#define SPIRV_DEBUG(consumer, ...) SPIRV_DEBUG_IMPL(consumer, __VA_ARGS__)
+#else
+#define SPIRV_DEBUG(consumer, ...)
+#endif
+
+// Logs an error message to the consumer saying the given feature is
+// unimplemented.
+#define SPIRV_UNIMPLEMENTED(consumer, feature)                     \
+  do {                                                             \
+    spvtools::Log(consumer, MessageLevel::InternalError, __FILE__, \
+                  {__LINE__, 0, 0}, "unimplemented: " feature);    \
+  } while (0)
+
+// Logs an error message to the consumer saying the code location
+// should be unreachable.
+#define SPIRV_UNREACHABLE(consumer)                                \
+  do {                                                             \
+    spvtools::Log(consumer, MessageLevel::InternalError, __FILE__, \
+                  {__LINE__, 0, 0}, "unreachable");                \
+  } while (0)
+
+// Helper macros for concatenating arguments.
+#define SPIRV_CONCATENATE(a, b) SPIRV_CONCATENATE_(a, b)
+#define SPIRV_CONCATENATE_(a, b) a##b
+
+// Helper macro to force expanding __VA_ARGS__ to satisfy MSVC compiler.
+#define PP_EXPAND(x) x
+
+namespace spvtools {
+
+// Calls the given |consumer| by supplying  the |message|. The |message| is from
+// the given |file| and |location| and of the given severity |level|.
+inline void Log(const MessageConsumer& consumer, MessageLevel level,
+                const char* file, const spv_position_t& position,
+                const char* message) {
+  if (consumer != nullptr) consumer(level, file, position, message);
+}
+
+// Calls the given |consumer| by supplying the message composed according to the
+// given |format|. The |message| is from the given |file| and |location| and of
+// the given severity |level|.
+template <typename... Args>
+void Logf(const MessageConsumer& consumer, MessageLevel level, const char* file,
+          const spv_position_t& position, const char* format, Args&&... args) {
+#if defined(_MSC_VER) && _MSC_VER < 1900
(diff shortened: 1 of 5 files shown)


source/opt/log.h
```diff
@@ -0,0 +1,186 @@
...
+#define SPIRV_LOG_LEVEL_DEBUG 2
+
+// By default we use SPIRV_LOG_LEVEL_ASSERT.
+#ifndef SPIRV_LOG_LEVEL
+#define SPIRV_LOG_LEVEL SPIRV_LOG_LEVEL_ASSERT
+#endif
+
+// TODO(antiagainst): add SPIRV_FATAL.
+
+// Asserts the given condition is true. Otherwise, print a message to stderr and
```
dneto0: I really don't like libraries printing messages to stderr.   A library client should be able to redirect diagnostics as it likes.  I'd rather have a logging context object that encapsulates the diagnostics output channel.  And that logging context object should probably be a member of the spv_context_t. 
Then feel free to use a naming convention and macros to substitute the current context's log context object into a simple calling sequence.

dneto0: For example, I think it's ugly that the unit tests print messages to stderr.  That shows bad encapsulation.

antiagainst: Totally agreed. I intended to put a TODO in the implementation, but somehow forgot. :(

I also don't like the current diagnostic mechanism. It's printing to `stderr` and the user has no way to redirect that except reading the content of `spv_diagnostic` directly. Even from the perspective of C interface, we should also provide something like `spvDiagnosticGet` or `spvDiagnosticPrintToFile`-ish stuff. And I feel that `spvDiagnosticCreate` is not something the user should call, especially with its parameters; it's more something internal.

I'd like to consolidate these two. (I mean, hijack all `spv_diagnostic` implementations with something new and keep the original C interface. For the upcoming C++ interface, `spv_diagnostic` should just go out of the scene.)

source/opt/log.h
```diff
@@ -0,0 +1,202 @@
...
+// consumer and abort the program. Accepts the following formats:
+//
+// SPIRV_ASSERT(<message-consumer>, <condition-expression>);
+// SPIRV_ASSERT(<message-consumer>, <condition-expression>, <message>);
+// SPIRV_ASSERT(<message-consumer>, <condition-expression>,
+//              <message-format>,   <variable-arguments>);
+//
+// In the third format, the number of <variable-arguments> cannot exceed (5 -
+// 2). If more arguments are wanted, grow PP_ARG_N and PP_NARGS in the below.
+#if !defined(NDEBUG)
```
Qining: I think it is better to have:
```
if defined(NDEBUG)
<define to none>
else
<define to something>
```

just a little bit easier to read.

dneto0:
I'll wait for a rebase on top of https://github.com/KhronosGroup/SPIRV-Tools/pull/401 before reviewing.

source/opt/log.h
antiagainst: I think the convention is just to define the meaningful one first and the none one later.

antiagainst:
@dneto0: rebased.

Sorry this one doesn't have tests. These macros will abort the program, but not through `abort()`, which googletest death tests depend on. As for `SPIRV_DEBUG`, it's only for a single build type with extra options turned on.

But I do use these macros in latter PRs. So they should be fine.

source/opt/log.h
```diff
@@ -0,0 +1,202 @@
...
+#else
+#define SPIRV_DEBUG(consumer, ...)
+#endif
+
+// Logs an error message to the consumer saying the given feature is
+// unimplemented and abort the program.
+#define SPIRV_UNIMPLEMENTED(consumer, feature)                               \
+  do {                                                                       \
+    spvtools::Log(consumer, MessageLevel::Fatal, __FILE__, {__LINE__, 0, 0}, \
+                  "unimplemented: " feature);                                \
```
dneto0: I don't see the abort part. Is the blurb out of date?

source/opt/log.h
```diff
@@ -0,0 +1,202 @@
...
+  } while (0)
+
+// Logs an error message to the consumer saying the code location
+// should be unreachable and abort the program.
```
dneto0: what abort?

source/opt/log.h
```diff
@@ -0,0 +1,202 @@
...
+#define SPIRV_UNREACHABLE(consumer)                                \
+  do {                                                             \
+    spvtools::Log(consumer, MessageLevel::InternalError, __FILE__, \
+                  {__LINE__, 0, 0}, "unreachable");                \
+  } while (0)
+
+// Helper macros for stringify the given argument.
+#define SPIRV_STRINGIFY(s) SPIRV_STRINGIFY_(s)
```
dneto0: I don't understand why you have two versions of this macro.

source/opt/log.h
```diff
@@ -0,0 +1,202 @@
...
+// the given |file| and |location| and of the given severity |level|.
+inline void Log(const MessageConsumer& consumer, MessageLevel level,
+                const char* file, const spv_position_t& position,
+                const char* message) {
+  if (consumer != nullptr) consumer(level, file, position, message);
+
+  // TODO(antiagainst): This could tear down the whole problem if called in a
+  // library. But do we have better solution?
+  if (level == MessageLevel::Fatal || level == MessageLevel::InternalError)
+    exit(EXIT_FAILURE);
```
dneto0: Libraries should not exit.

source/opt/log.h
```diff
@@ -0,0 +1,202 @@
...
+
+// Helper macro to force expanding __VA_ARGS__ to satisfy MSVC compiler.
+#define PP_EXPAND(x) x
+
+namespace spvtools {
+
+enum { INIT_BUFFER_SIZE = 256 };
+
+// Calls the given |consumer| by supplying  the |message|. The |message| is from
+// the given |file| and |location| and of the given severity |level|.
```
dneto0: Blurb doesn't say it exits on fatal or internal error.

source/opt/log.h
```diff
@@ -0,0 +1,202 @@
...
+inline void Log(const MessageConsumer& consumer, MessageLevel level,
+                const char* file, const spv_position_t& position,
+                const char* message) {
+  if (consumer != nullptr) consumer(level, file, position, message);
+
+  // TODO(antiagainst): This could tear down the whole problem if called in a
```
dneto0: problem -> process ??

source/opt/log.h
dneto0: Note: You can't actually test this case (except through weird weird mechanisms.  So that's a smell.

At least two alternatives:
 - accept an "exit" function pointer that is called in this case.  That allows you to inject the dependency.   But it would be something like "throw an exception" which is ugly, or setjmp/longjmp which is worse in C++.
 - don't exit!   I like this alternative.   For example, maybe Log should return the error classification enum, and then the caller can choose to do what they want.  You can then wrap common behaviours as you like.

dneto0:
-1

source/opt/log.h
antiagainst: Removed now.

source/opt/log.h
antiagainst: Removed now.

source/opt/log.h
antiagainst: Good catch! They are all not used and removed. :)

source/opt/log.h
antiagainst: Not existing any more.

source/opt/log.h
antiagainst: N/A anymore.

source/opt/log.h
antiagainst: The No.2 approach is taken. Much easier to write tests now.

source/opt/log.h
```diff
@@ -0,0 +1,198 @@
...
+// the given |file| and |location| and of the given severity |level|.
+inline void Log(const MessageConsumer& consumer, MessageLevel level,
+                const char* file, const spv_position_t& position,
+                const char* message) {
+  if (consumer != nullptr) consumer(level, file, position, message);
+}
+
+#if defined(_MSC_VER) && _MSC_VER < 1900
+// Sadly, snprintf() is not supported until Visual Studio 2015!
+#define snprintf _snprintf
```
dneto0: You must remember to #undef it again at the end of this header.

dneto0:
-1 have to #undef snprintf again at the end of the log.h (conditionally of course)

source/opt/log.h
antiagainst: Ha... I will do that.

antiagainst: Squashed all commits into one. Also addressed this issue.

source/opt/log.h
```diff
@@ -0,0 +1,202 @@
...
+// Helper macros for concatenating arguments.
+#define SPIRV_CONCATENATE(a, b) SPIRV_CONCATENATE_(a, b)
+#define SPIRV_CONCATENATE_(a, b) a##b
+
+// Helper macro to force expanding __VA_ARGS__ to satisfy MSVC compiler.
+#define PP_EXPAND(x) x
+
+namespace spvtools {
+
+enum { INIT_BUFFER_SIZE = 256 };
```
dneto0: No need to put this outside the function.
Move it inside the body, and use new-style naming, e.g.  kInitBufferSize.
See the Google C++ style guide.

antiagainst: Done.

dneto0:
+2
Please squash

PR #364, opened 2016-08-22
Add spirv-lesspipe.sh

dneto0:
Suggested by @steve-lunarg in issue 359.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@ v2016.3-dev 2016-08-11
  - Add spirv-cfg, an experimental tool to dump the control flow graph
    as a GraphiViz "dot" graph
  - Add optimization pass: Eliminate dead constants.
+ - Add spirv-lesspipe.sh filter utility
  - Fixes issues:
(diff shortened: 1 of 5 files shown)


dneto0:
For Windows users, it might be more friendly to add a .sh suffix to lesspipe since executables might only be run via an extension association.

steve-lunarg:
Agreed about the .sh suffix.  That's a common convention in the *nix world anyway, so seems appropriate for both.

Also, though it's a very minor thing, I suppose technically the script could be #!/bin/sh rather than #!/bin/bash, which is a slightly less demanding request.

dneto0:
- Added the .sh suffix
- The script now passes through extra arguments to spirv-dis, so the user can specify things like --raw-id and --no-color options in the LESSOPEN configuration.
- documented the change in CHANGES

Ready for review again.

antiagainst:
+2

dneto0:
Rebased. Keeping +2

Issue #365, opened 2016-08-22
Validator should not check for capability when seeing decorations of ClipDistance, CullDistance, or PointSize


chrisforbes:
See KhronosGroup/Vulkan-LoaderAndValidationLayers#858, KhronosGroup/glslang#472.

johnkslang:
I fixed the title, as this should not be generalized beyond ClipDistance, CullDistance, and PointSize.

dneto0:
I made that change to the validator on July 7: 
https://github.com/KhronosGroup/SPIRV-Tools/commit/06581f5ce6cb636ed1c6611da371656c26a1d871

But it's conditioned on using a Vulkan 1.0 context, since it was a Vulkan decision to relax that rule.
@chrisforbes Please confirm that you have code at least as fresh as that.
  
The master branch of the Vulkan layers tree specifies revision 6c61bf2dfadf70ab from July 22 which should be fresh enough.  And its only call to spvContextCreate does use SPV_ENV_VULKAN_1_0.

dneto0:
Commented on the layers bug suggesting that maybe the command line tool was not being used, and that you should use `--target-env vulkan1.0` option.

johnkslang:
What situation should error on ClipDistance (et. al.) Decoration with no capability?  I don't think there is one; rather the validator should unconditionally tolerate ClipDistance/CullDistance/PointSize decoration, and not require specific options be specified to do so.

dneto0:
The basic rule applied is any *mention* of an enum that lists a required capability may only be in a module that declares that capability. That's even as a decoration.
The specific rule that overrides that default is in the Vulkan env spec, which is why the relaxation of the check is only done for Vulkan 1.0 targets.

In particular we now have OpenGL environments in SPIRV-Tools.  Almost certainly the same rule override would apply there. Since it's basically the same GLSL variation is going to be used any time ClipDistance (and friends) is to be used, there's a strong argument to say the rule should always be relaxed for ClipDistance, CullDistance, PointSize.  I would guess that's your recommendation @johnkslang  ?

johnkslang:
Yes, the exception is driven by whether it is ClipDistance/CullDistance/PointSize, and not driven by what the environment is.

The general rule (decoration requires capability), would also apply universally, unless it was  ClipDistance/CullDistance/PointSize.

So, slice along decoration name, not along environment.

dneto0:
Ok.  I'll do that.

[...]

PR #368, opened 2016-08-23
Relax ClipDistance, CullDistance capability check in all environments

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/365


test/Validate.Capability.cpp
```diff
@@ -1175,6 +1193,15 @@ TEST_P(ValidateCapabilityVulkan10, Capability) {
       << test_code;
 }
 
+TEST_P(ValidateCapabilityOpenGL40, Capability) {
+  const string test_code = MakeAssembly(GetParam());
+  std::cout << test_code << std::endl;
```
antiagainst: This line is used for debug? Also in the above `TEST_P`.

dneto0: Fixed in 358cb29

antiagainst:
+2

PR #370, opened 2016-08-24
ForEachInst optionally runs on attached debug line insts

dneto0:
Also:
- Add const forms of ForEachInst
- Rewrite Module::ToBinary in terms of ForEachInst
- Add Instruction::ToBinaryWithoutAttachedDebugInsts
- Delete the ToBinary method on Function, BasicBlock, and Instruction
  since it can now be implemented with ForEachInst in a less confusing
  way, e.g. without recursion.
- Preserve debug line instructions on OpFunctionEnd (and store that
  instruction as a unique-pointer, for regularity).

### DIFF
--- a/source/opt/basic_block.h
+++ b/source/opt/basic_block.h
@@ -62,19 +62,19 @@ class BasicBlock {
   const_iterator cbegin() { return const_iterator(&insts_, insts_.cbegin()); }
   const_iterator cend() { return const_iterator(&insts_, insts_.cend()); }
 
-  // Runs the given function |f| on each instruction in this basic block.
-  inline void ForEachInst(const std::function<void(Instruction*)>& f);
-
-  // Pushes the binary segments for this instruction into the back of *|binary|.
-  // If |skip_nop| is true and this is a OpNop, do nothing.
-  inline void ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const;
+  // Runs the given function |f| on each instruction in this basic block, and
+  // optionally on the debug line instructions that might precede them.
+  inline void ForEachInst(const std::function<void(Instruction*)>& f,
+                          bool run_on_debug_line_insts = false);
+  inline void ForEachInst(const std::function<void(const Instruction*)>& f,
+                          bool run_on_debug_line_insts = false) const;
 
  private:
   // The enclosing function.
   Function* function_;
   // The label starting this basic block.
   std::unique_ptr<Instruction> label_;
-  // Instructions inside this basic block.
+  // Instructions inside this basic block, but not the OpLabel.
   std::vector<std::unique_ptr<Instruction>> insts_;
 };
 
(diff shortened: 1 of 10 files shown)


source/opt/basic_block.h
```diff
@@ -62,19 +62,19 @@ class BasicBlock {
...
-  inline void ForEachInst(const std::function<void(Instruction*)>& f);
-
-  // Pushes the binary segments for this instruction into the back of *|binary|.
-  // If |skip_nop| is true and this is a OpNop, do nothing.
-  inline void ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const;
+  // Runs the given function |f| on each instruction in this basic block, and
+  // optionally on the debug line instructions that might precede them.
+  inline void ForEachInst(const std::function<void(Instruction*)>& f,
+                          bool run_on_debug_line_insts = false);
+  inline void ForEachInst(const std::function<void(const Instruction*)>& f,
```
antiagainst: Maybe we should change the parameter to `const std::function<void(const Instruction&)>` so that we don't need those `static_cast` to aid compilers to do resolution? This also satisfies the style guide.

source/opt/ir_loader.cpp
```diff
@@ -52,6 +52,7 @@ void IrLoader::AddInstruction(const spv_parsed_instruction_t* inst) {
   } else if (opcode == SpvOpFunctionEnd) {
     assert(function_ != nullptr);
     assert(block_ == nullptr);
+    function_->SetFunctionEnd(std::move(spv_inst));
```
antiagainst: Yay! Thanks for fixing this!

source/opt/basic_block.h
dneto0: Good idea.  Implemented in the second commit.

dneto0:
Rebased, and implemented suggestion to change the signature of the ForEachInst argument.  Looks better!

dneto0:
Whoops.  Visual Studio 2013 can't disambiguate the overloads of Instruction::ForEachInst.
```
c:\projects\spirv-tools\source\opt\basic_block.h(97): error C2666: 'spvtools::ir::Instruction::ForEachInst' : 2 overloads have similar conversions [C:\projects\spirv-tools\build\source\opt\SPIRV-Tools-opt.vcxproj]
          c:\projects\spirv-tools\source\opt\instruction.h(162): could be 'void spvtools::ir::Instruction::ForEachInst(const std::function<void (const spvtools::ir::Instruction &)> &,bool) const'
          c:\projects\spirv-tools\source\opt\instruction.h(160): or       'void spvtools::ir::Instruction::ForEachInst(const std::function<void (spvtools::ir::Instruction *)> &,bool)'
          while trying to match the argument list '(const std::function<void (const spvtools::ir::Instruction &)>, bool)'
          note: qualification adjustment (const/volatile) may be causing the ambiguity
```

dneto0:
So now it's back to the original code, but rebased once.

antiagainst:
LGTM once the buildbot is happy.

PR #372, opened 2016-08-25
Add Instruction cbegin and cend to access operands

dneto0:


dneto0:
This would be useful for some spots in https://github.com/KhronosGroup/SPIRV-Tools/pull/324 
where you just want to iterate over operands but don't need their index at the same time.  So you can just do a range-based-for.

Looking for @antiagainst review

source/opt/instruction.h
```diff
@@ -117,6 +119,10 @@ class Instruction {
     return dbg_line_insts_;
   }
 
+  // Const begin and end iterators for operands.
+  const_iterator cbegin() { return operands_.cbegin(); }
```
antiagainst: `const_iterator ...() const`?

test/opt/test_instruction.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+TEST(InstructionTest, CreateWithOpcodeAndOperands) {
+  auto i32 = MakeInstruction(SpvOpTypeInt, {44, 32, 1});
+  spv_parsed_operand_t parsed_operands[] = {
+      {1, 1, SPV_OPERAND_TYPE_RESULT_ID, SPV_NUMBER_NONE, 0},
+      {2, 1, SPV_OPERAND_TYPE_LITERAL_INTEGER, SPV_NUMBER_UNSIGNED_INT, 32},
+      {3, 1, SPV_OPERAND_TYPE_LITERAL_INTEGER, SPV_NUMBER_UNSIGNED_INT, 1},
+  };
+  spv_parsed_instruction_t parsed_inst{i32.data(),
+                                       uint16_t(i32.size()),
```
antiagainst: Wow... `clang-format` sometime really does a "nice" job.

But I won't complain anyway.

test/opt/test_instruction.cpp
```diff
@@ -0,0 +1,118 @@
...
+  EXPECT_EQ(3u, inst.NumOperandWords());
+  EXPECT_EQ(2u, inst.NumInOperandWords());
+  auto cbegin = inst.cbegin();
+  auto cend = inst.cend();
+  EXPECT_NE(cend, inst.cbegin());
+
+  // Spot check iteration across operands.
+
+  auto iter = inst.cbegin();
+  const auto& operand0 = *iter;
```
antiagainst: It seems we can do something like
```
for (int i = 0; i < 3; ++i) {
  const auto& operand = *iter;
  EXPECT_THAT(operand.type, Eq(parsed_operands[i]..));
  EXPECT_THAT(operand.words, Eq(std::vector<uint32_t>{i32[i + 1]}));
  EXPECT_NE(cend, iter++);
}
```

dneto0: Some advice tells you not to have too much flow control in your tests.  I rewrote it your way anyway.

source/opt/instruction.h
dneto0: Thanks!  Good catch.

dneto0:
Rebased

antiagainst:
LGTM

PR #373, opened 2016-08-25
IrLoader gracefully handles incomplete blocks and functions

dneto0:
This lets us write smaller test cases with the IrLoader, avoiding
boilerplate for function begin/end, and basic block begin/end.

Also ForEachInst is more forgiving of cases where a basic block
doesn't have a label, and when a function doesn't have a defining
or end instruction.

### DIFF
--- a/source/opt/basic_block.h
+++ b/source/opt/basic_block.h
@@ -87,15 +87,16 @@ inline void BasicBlock::AddInstruction(std::unique_ptr<Instruction> i) {
 
 inline void BasicBlock::ForEachInst(const std::function<void(Instruction*)>& f,
                                     bool run_on_debug_line_insts) {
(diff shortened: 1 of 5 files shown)


dneto0:
Rebased

Qining:
So this is adding guards for de-references? lgtm

@antiagainst PTAL

antiagainst:
+2

dneto0:
@qining Yes, I definitely need the guard on a function's end instruction, in the case where the function hasn't been closed off.  There's a test for that.  The guard on the defining instruction is there out of symmetry and defensiveness. I don't think you can make an ir::Function without a defining instruction.

antiagainst:
Thanks for this! :)

PR #376, opened 2016-08-26
Add Module::ComputeIdBound

dneto0:
When IDs change, something should use this utility method
and Module::SetIdBound before writing out the binary.

ForEachInst is more forgiving of cases where a basic block doesn't
have a label, and when a function doesn't have a defining instruction
or end instruction.

### DIFF
--- a/source/opt/module.cpp
+++ b/source/opt/module.cpp
@@ -25,6 +25,10 @@
 // MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
 #include "module.h"
+
+#include <algorithm>
+
+#include "operand.h"
 #include "reflect.h"
 
 namespace spvtools {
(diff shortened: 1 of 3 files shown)


test/opt/test_instruction.cpp
```diff
@@ -36,6 +36,7 @@ namespace {
 using spvtest::MakeInstruction;
 using spvtools::ir::Instruction;
 using spvtools::ir::Operand;
+using spvtools::ir::Operand;
```
antiagainst: :)

source/opt/module.cpp
```diff
@@ -117,5 +119,22 @@ void Module::ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const {
   ForEachInst(write_inst, true);
 }
 
+uint32_t Module::ComputeIdBound() const {
+  uint32_t highest = 0;
+
+  ForEachInst(
+      [&highest](const Instruction* inst) {
+        // Use a const-cast just to access begin() and end() for the range-for.
+        for (const auto& operand : *const_cast<Instruction*>(inst)) {
```
antiagainst: Wow. I really wish we have a way to avoid doing all these casts. Windows compilers, sigh.

antiagainst: Why going through all the ids instead of just the result ids?

Is it valid to just "use" some id without defining it eventually?

antiagainst: OK, I see tests below. But do we really need that?

antiagainst:
It seems the second paragraph in the commit message also need to be removed.

source/opt/module.cpp
antiagainst: Well, necessary.

dneto0: Good question.  It's not much more work scan everything.  It's easier to write tests this way. :-)

dneto0:
Removed the double "using" declaration.
Fixed the commit comment.
Rebased

source/opt/module.cpp
antiagainst: Sorry for this. It seems we can get away with this cast by adding the following into `Instruction`?

```cpp
const_iterator begin() const { return operands_.cbegin(); }
const_iterator end() const { return operands_.cend(); }

```

dneto0:
Fixed std::max problem on Windows.  Needed to include <algorithm>

antiagainst:
+2

dneto0:
Rebased

PR #377, opened 2016-08-26
Add a method in the types to check whether there is no decoration on a type.

Qining:


Qining:
@antiagainst

Qining:
@antiagainst  updated

antiagainst:
+2

dneto0:
I want to see tests for this kind of thing.

antiagainst:
@dneto0: Oh, my fault. I should requested that. So, I write it here: #378.

Issue #371, opened 2016-08-24
optimizer passes are not updating the ID bound in the header


dneto0:
Several steps are required to fix this.
* Add an API to update the module header word.  (done)
* Rework how you iterate over instructions #370 
* Add method to compute a tight ID Bound. (TODO)
* Ensure the ID bound is updated at appropriate times.

PR #381, opened 2016-08-26
Pass manager recomputes Id bound automatically.

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/371
in the sense that the id bound is correct after all the passes
have been run.  But it might be inconsistent between passes.

### DIFF
--- a/test/opt/test_pass_manager.cpp
+++ b/test/opt/test_pass_manager.cpp
@@ -24,12 +24,17 @@
 // TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 // MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
-#include "pass_fixture.h"
+#include "gmock/gmock.h"
+
+#include "module_utils.h"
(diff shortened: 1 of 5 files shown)


test/opt/test_pass_manager.cpp
```diff
@@ -91,4 +96,44 @@ TEST_F(PassManagerTest, Run) {
...
+  manager.Run(&module);
+  // The Id has been updated automatically, even though the pass
+  // did not update it.
+  EXPECT_THAT(GetIdBound(module), Eq(101u));
+
+  // Try one more time!
+  manager.AddPass(MakeUnique<AppendTypeVoidInstPass>(200));
+  manager.Run(&module);
+  EXPECT_THAT(GetIdBound(module), Eq(201u));
+}
```
antiagainst: Probably we should try adding another `AppendTypeVoidInstPass` with a much smaller id to make sure that the id bound won't decrease?

dneto0:
Rebased, and addressed all comments.

antiagainst:
+2

PR #384, opened 2016-08-27
Add libspirv::CapabilitySet

dneto0:
It's optimized for the common case, where capabilities have value
at most 63.


dneto0:
There are two commits here.  The first one adds libspirv::CapabilitySet and tests it.
The second one uses CapabilitySet to store sets of capabilities (duh) instead of using a 64-bit mask.

We'll need this for the case where extensions will have capabilities with values higher than 63.

dneto0:
Looks like I have to adapt to VisualStudio 2013 incompatibilities with std::initializer_list constructors...

dneto0:
With lots of help from @awoloszyn we found that on VS 2013, the ValidationState member initializer for module_capabilities_({}) was causing CapabilitySet(0) to be called, and hence insert enum valued 0 into the set.  :-(
We fixed it by just using module_capabilities_()

Also, added "explicit" qualifier to the single-argument constructor for CapabilitySet.

dneto0:
Rebased

source/capability_set.h
```diff
@@ -0,0 +1,116 @@
...
+// It is fast and compact for the common case, where capability values
+// are at most 63.  But it can represent capabilities with larger values,
+// as may appear in extensions.
+class CapabilitySet {
+ private:
+  using OverflowSetType = std::set<uint32_t>;
+
+ public:
+  // Construct an empty set.
+  CapabilitySet() {}
```
antiagainst: `= default`?

source/capability_set.h
```diff
@@ -0,0 +1,116 @@
...
+  // Returns true if the capability represented as a 32-bit word is in the set.
+  bool Contains(uint32_t word) const;
+
+  // Applies f to each capability in the set, in order from smallest enum
+  // value to largest.
+  void ForEach(std::function<void(SpvCapability)> f) const;
+
+ private:
+  // Returns true if the set has capabilities with value greater than 63.
+  bool HasOverflow() { return overflow_.get() != nullptr; }
```
antiagainst: I think `.get()` is redundant here.

source/capability_set.cpp
```diff
@@ -0,0 +1,75 @@
...
+void CapabilitySet::Add(uint32_t word) {
+  if (auto new_bits = AsMask(word)) {
+    mask_ |= new_bits;
+  } else {
+    Overflow().insert(word);
+  }
+}
+
+bool CapabilitySet::Contains(uint32_t word) const {
+  // We shouldn't call Overflow() since this is a const // method.
```
antiagainst: `const method`

source/capability_set.cpp
```diff
@@ -0,0 +1,75 @@
...
+  return false;
+}
+
+// Applies f to each capability in the set, in order from smallest enum
+// value to largest.
+void CapabilitySet::ForEach(std::function<void(SpvCapability)> f) const {
+  for (uint32_t i = 0; i < 64; ++i) {
+    if (mask_ & AsMask(i)) f(static_cast<SpvCapability>(i));
+  }
+  if (overflow_.get()) {
```
antiagainst: I think `.get()` is redundant here.

source/capability_set.h
```diff
@@ -0,0 +1,116 @@
...
+
+namespace libspirv {
+
+// A set of values of type SpvCapability.
+// It is fast and compact for the common case, where capability values
+// are at most 63.  But it can represent capabilities with larger values,
+// as may appear in extensions.
+class CapabilitySet {
+ private:
+  using OverflowSetType = std::set<uint32_t>;
```
antiagainst: `unordered_set`?

utils/generate_grammar_tables.py
```diff
@@ -59,10 +37,7 @@ def compose_capability_mask(caps):
     Returns:
       a string containing the hexadecimal value of the bit mask
     """
-    assert len(CAPABILITY_BIT_MAPPING) != 0
-    bits = [CAPABILITY_BIT_MAPPING[c] for c in caps]
-    caps_mask = functools.reduce(lambda m, b: m | (1 << b), bits, 0)
-    return '0x{:04x}'.format(caps_mask)
+    return "{" + ", ".join(['SpvCapability{}'.format(c) for c in caps]) + "}"
```
antiagainst: I'd suggest to use `"...".format()`. This is kinda hard to read.

utils/generate_grammar_tables.py
```diff
@@ -31,26 +28,7 @@ def make_path_to_file(f):
...
-            print(
-                "error: capability enumerant {} valued {} is over 63; "
-                "spv_capability_mask_t doesn't support this".format(
-                    enumerant['enumerant'], enumerant['value']))
-            exit(1)
-        CAPABILITY_BIT_MAPPING[enumerant['enumerant']] = enumerant['value']
-
-
-def compose_capability_mask(caps):
+def compose_capability_list(caps):
```
antiagainst: Nice to see complicated logic got killed! ;)

utils/generate_grammar_tables.py
```diff
@@ -59,10 +37,7 @@ def compose_capability_mask(caps):
     Returns:
       a string containing the hexadecimal value of the bit mask
```
antiagainst: This needs to be updated.

utils/generate_grammar_tables.py
antiagainst: Well, never mind. Just my eyes. ;)

source/capability_set.h
dneto0: The ForEach method promises to call the capabilities in order.  So I want an ordered set here. I'll add the comment saying so.

dneto0:
Rebased and addressed all comments.

antiagainst:
LGTM

dneto0:
Rebased

PR #387, opened 2016-08-29
Get annotation instructions for a given id

Qining:


source/opt/def_use_manager.cpp
```diff
@@ -85,6 +87,24 @@ UseList* DefUseManager::GetUses(uint32_t id) {
   return &iter->second;
 }
 
+const UseList* DefUseManager::GetUses(uint32_t id) const {
```
antiagainst: `const &`

source/opt/def_use_manager.h
```diff
@@ -72,6 +73,9 @@ class DefUseManager {
   // Returns the use instructions for the given |id|. If there is no uses of
   // |id|, returns nullptr.
   UseList* GetUses(uint32_t id);
+  const UseList* GetUses(uint32_t id) const;
+  // Returns a set of annotation instrunctions whose target is the given |id|.
+  std::unordered_set<ir::Instruction*> GetAnnotations(uint32_t id) const;
```
antiagainst: Why `unordered_set` instead of a plain `vector`?

source/opt/def_use_manager.cpp
Qining: that way what should be return if the id is not found?

antiagainst: Never mind.

test/opt/test_def_use.cpp
```diff
@@ -1414,4 +1414,41 @@ INSTANTIATE_TEST_CASE_P(
       }));
 // clang-format on
 
+TEST(DefUseManagerSingleTest, GetAnnotations) {
+  // Code
+  const char* code =
+      "OpDecorate %1 Block "
+      "OpDecorate %1 RelaxedPrecision "
```
antiagainst: We should also add some debug instructions here.

Qining: One debug instruction is added in 92d6fd7. The debug instruction won't be counted as an annotation instruction.

test/opt/test_def_use.cpp
```diff
@@ -1414,4 +1414,51 @@ INSTANTIATE_TEST_CASE_P(
...
+  EXPECT_EQ(2u, annotations_of_1.size());
+  EXPECT_EQ(0u, annotations_of_2.size());
+  EXPECT_EQ(0u, annotations_of_3.size());
+  EXPECT_EQ(0u, annotations_of_4.size());
+
+  using DecorationSet = std::unordered_set<SpvDecoration, std::hash<uint32_t>>;
+  DecorationSet actual_decorations;
+  DecorationSet expected_decorations;
+
+  for (auto* anno_inst : annotations_of_1) {
```
antiagainst: I think you can directly check the two elements in the vector...

test/opt/test_def_use.cpp
```diff
@@ -1414,4 +1414,51 @@ INSTANTIATE_TEST_CASE_P(
       }));
 // clang-format on
 
+TEST(DefUseManagerSingleTest, GetAnnotations) {
+  // Code
+  const char* code =
+      "OpName %1 \"struct_type\" "
+      "OpName %3 \"int_type\" "
+      "OpDecorate %1 Block "
+      "OpDecorate %1 RelaxedPrecision "
```
antiagainst: DecorationGroup

Qining: Just realized we have DisassembleInst() function available. Changed back to use parameterized tests in 5f94bfb

source/opt/def_use_manager.h
```diff
@@ -72,6 +72,14 @@ class DefUseManager {
   // Returns the use instructions for the given |id|. If there is no uses of
   // |id|, returns nullptr.
   UseList* GetUses(uint32_t id);
+  const UseList* GetUses(uint32_t id) const;
+  // Returns the annotation instrunctions which are a directly use of the given
```
antiagainst: direct

source/opt/def_use_manager.h
```diff
@@ -72,6 +72,14 @@ class DefUseManager {
...
+  // |id|. This means when the decorations are applied through decoration
+  // group(s), this function will just return the OpGroupDecorate
+  // instrcution(s) which refer to the given id as an operand. The OpDecorate
+  // instructions which decorates the decoration group will not be in the
```
antiagainst: which decorate

source/opt/def_use_manager.h
```diff
@@ -72,6 +72,14 @@ class DefUseManager {
...
+  // return vector.
```
antiagainst: will not be returned.

antiagainst:
+2. Just two grammar issues.

source/opt/def_use_manager.h
Qining: Fixed in e95acf1

source/opt/def_use_manager.h
Qining: fixed in e95acf1

source/opt/def_use_manager.h
Qining: fixed in e95acf1

Qining:
Both CI passed, squashed.

PR #388, opened 2016-08-31
Add forwarding so that passes' ctor can have args

Qining:


source/opt/null_pass.h
```diff
@@ -35,6 +35,13 @@ namespace opt {
 
 // A null pass that does nothing.
 class NullPass : public Pass {
+ public:
+  NullPass() : Pass() {}
+  // Interfaces for different arguments.
+  NullPass(uint32_t arg) : Pass() { (void)arg; }
```
antiagainst: I think you can get away with compiler warnings by omitting the parameter names.

test/opt/test_pass_manager.cpp
```diff
@@ -54,6 +56,17 @@ TEST(PassManager, Interface) {
   EXPECT_STREQ("strip-debug", manager.GetPass(0)->name());
   EXPECT_STREQ("null", manager.GetPass(1)->name());
   EXPECT_STREQ("strip-debug", manager.GetPass(2)->name());
+
+  manager.AddPass<opt::NullPass>(1u);
```
antiagainst: Instead of using `NullPass`, I would suggest that we create another pass here with various constructors for testing purpose only. `NullPass` is a user-facing pass, it should have meaningful interface.

test/opt/test_pass_manager.cpp
```diff
@@ -54,6 +56,17 @@ TEST(PassManager, Interface) {
...
+  manager.AddPass<opt::NullPass>("null pass args");
+  manager.AddPass<opt::NullPass>(std::initializer_list<int>{1, 2});
+  EXPECT_EQ(6u, manager.NumPasses());
+  EXPECT_STREQ("strip-debug", manager.GetPass(0)->name());
+  EXPECT_STREQ("null", manager.GetPass(1)->name());
+  EXPECT_STREQ("strip-debug", manager.GetPass(2)->name());
+  EXPECT_STREQ("null", manager.GetPass(3)->name());
```
antiagainst: Additionally you can do `static_cast` to the real type and access some of the data passed in via constructor.

test/opt/test_pass_manager.cpp
```diff
@@ -66,6 +79,24 @@ class AppendOpNopPass : public opt::Pass {
...
+// A pass that appends specified number of OpNop instructions to the debug
+// section.
+class AppendMultipleOpNopPass : public opt::Pass {
+ public:
+  AppendMultipleOpNopPass(uint32_t num_nop) : num_nop_(num_nop) {}
+  const char* name() const override { return "AppendOpNop"; }
+  bool Process(ir::Module* module) override {
+    for (uint32_t i = 0; i < num_nop_; i++) {
+      auto inst = MakeUnique<ir::Instruction>();
+      module->AddDebugInst(std::move(inst));
```
antiagainst: `AddDebugInst(MakeUnique<...>())`?

source/opt/pass_manager.h
```diff
@@ -50,9 +50,9 @@ class PassManager {
   void AddPass(std::unique_ptr<Pass> pass) {
     passes_.push_back(std::move(pass));
   }
-  template <typename PassT>
-  void AddPass() {
-    passes_.emplace_back(new PassT);
+  template <typename PassT, typename... Args>
```
antiagainst: blurb

test/opt/test_pass_manager.cpp
```diff
@@ -94,6 +125,10 @@ TEST_F(PassManagerTest, Run) {
   AddPass<DuplicateInstPass>();
   AddPass<AppendOpNopPass>();
   RunAndCheck(text.c_str(), (text + "OpSource ESSL 310\nOpNop\n").c_str());
+
+  RenewPassManger();
+  AddPass<AppendMultipleOpNopPass>(3);
+  RunAndCheck(text.c_str(), (text + "OpNop\nOpNop\nOpNop\n").c_str());
```
antiagainst: It would be nice to also see a use of `SinglePassRunAndCheck()` to make sure that `Args&&...` actually works. :)

source/opt/pass_manager.h
```diff
@@ -50,9 +50,9 @@ class PassManager {
...
+  void AddPass(Args&&... args) {
```
antiagainst: Cool stuff!

test/opt/pass_fixture.h
```diff
@@ -71,41 +71,42 @@ class PassTest : public TestT {
...
   // |original| assembly, and checks whether the optimized binary can be
   // disassembled to the |expected| assembly. This does *not* involve pass
   // manager. Callers are suggested to use SCOPED_TRACE() for better messages.
-  template <typename PassT>
+  template <typename PassT, typename... Args>
   void SinglePassRunAndCheck(const std::string& original,
-                             const std::string& expected,
-                             bool skip_nop = false) {
+                             const std::string& expected, bool skip_nop = false,
+                             Args&&... args) {
```
antiagainst: Cool stuff!

test/opt/test_pass_manager.cpp
Qining: I think that is covered in `Run` test.

test/opt/test_pass_manager.cpp
Qining: Hmm, I think that would make the line too long. I think the current form is easier to read

test/opt/test_pass_manager.cpp
Qining: Hmm, my question where to put the tests? So fart we don't have test for pass fixtures

source/opt/null_pass.h
Qining: updated in 06e098670

test/opt/test_pass_manager.cpp
Qining: fixed in 06e098670

source/opt/pass_manager.h
Qining: added in 06e098670

test/opt/test_pass_manager.cpp
Qining: This will be used in the next pass : set-spec-const-default-value. And will be tested at that time.

test/opt/test_pass_manager.cpp
```diff
@@ -38,6 +38,17 @@ using namespace spvtools;
 using spvtest::GetIdBound;
 using ::testing::Eq;
 
+// A null pass whose construtor accept arguments
```
antiagainst: accepts

antiagainst:
+2. Just need to fix the buildbot failure.

Qining:
dee6931 removed the default value of `skip_nop` arguments for function: `SinglePassRunAndCheck()` and `SinglePassRunAndDisassemble()`.

Qining:
Squashed and merged. All bots passed before squashing

PR #389, opened 2016-08-31
Refactor CapabilitySet into templated class EnumSet

dneto0:


Qining:
It looks like exactly the same thing as the capability set except make it a template for general uses. Looks good to me.

dneto0:
Yes, exactly.   I did this because I'm going to need to this to represent that certain things are dependent upon known extensions.  And I'll make the known-extensions-list an enum, and I'll need a compact representation of a set of enums, all of which are small.

dneto0:
Rebased and merged into master as 909d7f9bf5

PR #390, opened 2016-08-31
Set default values for spec constants with SpecId

Qining:
The pass instance is constructed with a map from spec id (uint32_t) to
default values in string format. The default value strings will be
parsed to numbers according to the target spec constant type.

If the Spec Id decoration is found to be applied on multiple different
target ids, that decoration instruction (OpDecorate or OpGroupDecorate)
will be skipped. But other decoration instrucitons may still be
processed.

### DIFF
--- a/source/opt/set_spec_constant_default_value_pass.cpp
+++ b/source/opt/set_spec_constant_default_value_pass.cpp
@@ -1,0 +1,252 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "set_spec_constant_default_value_pass.h"
+
+#include <cstring>
+#include <tuple>
+#include <unordered_map>
+#include <vector>
+
+#include "spirv-tools/libspirv.h"
+#include "util/parse_number.h"
+
+#include "def_use_manager.h"
+#include "type_manager.h"
+#include "types.h"
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+using spvutils::NumberType;
+using spvutils::EncodeNumberStatus;
+using spvutils::ParseAndEncodeNumber;
+
+// Given a numeric value in a null-terminated c string and the expected type of
+// the value, parses the string and encodes it in a vector of words. If the
+// value is a scalar integer or floating point value, encodes the value in
+// SPIR-V encoding format. If the value is 'false' or 'true', returns a vector
+// with single word with value 0 or 1 respectively. Returns the vector
+// containing the encoded value on success. Otherwise returns an empty vector.
+std::vector<uint32_t> ParseDefaultValueStr(const char* text,
+                                           const analysis::Type* type) {
+  std::vector<uint32_t> result;
+  if (!strcmp(text, "true")) {
+    result.push_back(1u);
+  } else if (!strcmp(text, "false")) {
+    result.push_back(0u);
+  } else {
+    NumberType number_type = {32, SPV_NUMBER_UNSIGNED_INT};
+    if (const auto* IT = type->AsInteger()) {
+      number_type.bitwidth = IT->width();
+      number_type.kind =
+          IT->IsSigned() ? SPV_NUMBER_SIGNED_INT : SPV_NUMBER_UNSIGNED_INT;
+    } else if (const auto* FT = type->AsFloat()) {
+      number_type.bitwidth = FT->width();
+      number_type.kind = SPV_NUMBER_FLOATING;
+    }
+    EncodeNumberStatus rc = ParseAndEncodeNumber(
+        text, number_type, [&result](uint32_t word) { result.push_back(word); },
+        nullptr);
+    // Clear the result vector on failure.
+    if (rc != EncodeNumberStatus::kSuccess) {
+      result.clear();
+    }
+  }
+  return result;
+}
+
+// Returns true if the given instruction's result id could have a SpecId
+// decoration.
+bool CanHaveSpecIdDecoration(const ir::Instruction& inst) {
+  switch (inst.opcode()) {
+    case SpvOp::SpvOpSpecConstant:
+    case SpvOp::SpvOpSpecConstantFalse:
+    case SpvOp::SpvOpSpecConstantTrue:
+      return true;
+    default:
+      return false;
+  }
+}
+
+// Given a decoration group defining instruction that is decorated with SpecId
+// decoration, finds the spec constant defining instruction which is the real
+// target of the SpecId decoration. Returns the spec constant defining
+// instruction if such an instruction is found, otherwise returns a nullptr.
+ir::Instruction* GetSpecIdTargetFromDecorationGroup(
+    const ir::Instruction& decoration_group_defining_inst,
+    analysis::DefUseManager* def_use_mgr) {
+  // Find the OpGroupDecorate instruction which consumes the given decoration
+  // group. Note that the given decoration group has SpecId decoration, which
+  // is unique for different spec constants. So the decoration group cannot be
+  // consumed by different OpGroupDecorate instructions. Therefore we only need
+  // the first OpGroupDecoration instruction that uses the given decoration
+  // group.
+  ir::Instruction* group_decorate_inst = nullptr;
+  for (const auto& u :
+       *def_use_mgr->GetUses(decoration_group_defining_inst.result_id())) {
+    if (u.inst->opcode() == SpvOp::SpvOpGroupDecorate) {
+      group_decorate_inst = u.inst;
+      break;
+    }
+  }
+  if (!group_decorate_inst) return nullptr;
+
+  // Scan through the target ids of the OpGroupDecorate instruction. There
+  // should be only one spec constant target consumes the SpecId decoration.
+  // If multiple target ids are presented in the OpGroupDecorate instruction,
+  // they must be the same one that defined by an eligible spec constant
+  // instruction. If the OpGroupDecorate instruction has different target ids
+  // or a target id is not defined by an eligible spec cosntant instruction,
+  // returns a nullptr.
+  ir::Instruction* target_inst = nullptr;
+  for (uint32_t i = 1; i < group_decorate_inst->NumInOperands(); i++) {
+    // All the operands of a OpGroupDecorate instruction should be of type
+    // SPV_OPERAND_TYPE_ID.
+    uint32_t candidate_id = group_decorate_inst->GetSingleWordInOperand(i);
+    ir::Instruction* candidate_inst = def_use_mgr->GetDef(candidate_id);
+
+    if (!candidate_inst) {
+      continue;
+    }
+
+    if (!target_inst) {
+      // If the spec constant target has not been found yet, check if the
+      // candidate instruction is the target.
+      if (CanHaveSpecIdDecoration(*candidate_inst)) {
+        target_inst = candidate_inst;
+      } else {
+        // Spec id decoration should not be applied on other instructions.
+        // TODO(qining): Emit an error message in the invalid case once the
+        // error handling is done.
+        return nullptr;
+      }
+    } else {
+      // If the spec constant target has been found, check if the candidate
+      // instruction is the same one as the target. The module is invalid if
+      // the candidate instruction is different with the found target.
+      // TODO(qining): Emit an error messaage in the invalid case once the
+      // error handling is done.
+      if (candidate_inst != target_inst) return nullptr;
+    }
+  }
+  return target_inst;
+}
+};
+
+bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
+  // The operand index of decoration target in an OpDecorate instruction.
+  const uint32_t kTargetIdOperandIndex = 0;
+  // The operand index of the decoration literal in an OpDecorate instruction.
+  const uint32_t kDecorationOperandIndex = 1;
+  // The operand index of Spec id literal value in an OpDecorate SpecId
+  // instruction.
+  const uint32_t kSpecIdLiteralOperandIndex = 2;
+  // The number of operands in an OpDecorate SpecId instruction.
+  const uint32_t kOpDecorateSpecIdNumOperands = 3;
+  // The in-operand index of the default value in a OpSpecConstant instruction.
+  const uint32_t kOpSpecConstantLiteralInOperandIndex = 0;
+
+  bool modified = false;
+  analysis::DefUseManager def_use_mgr(module);
+  analysis::TypeManager type_mgr(*module);
+  // Scan through all the annotation instructions to find 'OpDecorate SpecId'
+  // instructions. Then extract the decoration target of those instructions.
+  // The decoration targets should be spec constant defining instructions with
+  // opcode: OpSpecConstant{|True|False}. The spec id of those spec constants
+  // will be used to look up their new default values in the mapping from
+  // spec id to new default value strings. Once a new default value string
+  // is found for a spec id, the string will be parsed according to the target
+  // spec constant type. The parsed value will be used to replace the original
+  // default value of the target spec constant.
+  for (ir::Instruction& inst : module->annotations()) {
+    // Only process 'OpDecorate SpecId' instructions
+    if (inst.opcode() != SpvOp::SpvOpDecorate) continue;
+    if (inst.NumOperands() != kOpDecorateSpecIdNumOperands) continue;
+    if (inst.GetSingleWordInOperand(kDecorationOperandIndex) !=
+        uint32_t(SpvDecoration::SpvDecorationSpecId)) {
+      continue;
+    }
+
+    // 'inst' is an OpDecorate SpecId instruction.
+    uint32_t spec_id = inst.GetSingleWordOperand(kSpecIdLiteralOperandIndex);
+    uint32_t target_id = inst.GetSingleWordOperand(kTargetIdOperandIndex);
+
+    // Find the spec constant defining instruction. Note that the
+    // target_id might be a decoration group id.
+    ir::Instruction* spec_inst = nullptr;
+    if (ir::Instruction* target_inst = def_use_mgr.GetDef(target_id)) {
+      if (target_inst->opcode() == SpvOp::SpvOpDecorationGroup) {
+        spec_inst =
+            GetSpecIdTargetFromDecorationGroup(*target_inst, &def_use_mgr);
+      } else {
+        spec_inst = target_inst;
+      }
+    } else {
+      continue;
+    }
+    if (!spec_inst) continue;
+
+    // Search for the new default value for this spec id.
+    auto iter = spec_id_to_value_.find(spec_id);
+    if (iter == spec_id_to_value_.end()) continue;
+
+    // Gets the string of the default value and parses it to bit pattern
+    // with the type of the spec constant.
+    const std::string& default_value_str = iter->second;
+    std::vector<uint32_t> bit_pattern = ParseDefaultValueStr(
+        default_value_str.c_str(), type_mgr.GetType(spec_inst->type_id()));
+    if (bit_pattern.empty()) continue;
+
+    // Update the operand bit patterns of the spec constant defining
+    // instruction.
+    switch (spec_inst->opcode()) {
+      case SpvOp::SpvOpSpecConstant:
+        // If the new value is the same with the original value, no
+        // need to do anything. Otherwise update the operand words.
+        if (spec_inst->GetInOperand(kOpSpecConstantLiteralInOperandIndex)
+                .words != bit_pattern) {
+          spec_inst->SetInOperand(kOpSpecConstantLiteralInOperandIndex,
+                                  std::move(bit_pattern));
+          modified = true;
+        }
+        break;
+      case SpvOp::SpvOpSpecConstantTrue:
+        // If the new value is also 'true', no need to change anything.
+        // Otherwise, set the opcode to OpSpecConstantFalse;
+        if (!static_cast<bool>(bit_pattern.front())) {
+          spec_inst->SetOpcode(SpvOp::SpvOpSpecConstantFalse);
+          modified = true;
+        }
+        break;
+      case SpvOp::SpvOpSpecConstantFalse:
+        // If the new value is also 'false', no need to change anything.
+        // Otherwise, set the opcode to OpSpecConstantTrue;
+        if (static_cast<bool>(bit_pattern.front())) {
+          spec_inst->SetOpcode(SpvOp::SpvOpSpecConstantTrue);
+          modified = true;
+        }
+        break;
+      default:
+        break;
+    }
+    // No need to update the DefUse manager, as this pass does not change any
+    // ids.
+  }
+  return modified;
+}
+
+}  // namespace opt
+}  // namespace spvtools
--- a/source/opt/set_spec_constant_default_value_pass.h
+++ b/source/opt/set_spec_constant_default_value_pass.h
@@ -1,0 +1,54 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_OPT_SET_SPEC_CONSTANT_DEFAULT_VALUE_PASS_H_
+#define LIBSPIRV_OPT_SET_SPEC_CONSTANT_DEFAULT_VALUE_PASS_H_
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include "module.h"
+#include "pass.h"
+
+namespace spvtools {
+namespace opt {
+
+// The optimization pass that sets the default values for the spec constants
+// that have SpecId decorations (i.e. those defined by
+// OpSpecConstant{|True|False} instructions).
+class SetSpecConstantDefaultValuePass : public Pass {
+ public:
+  using SpecIdToValueStrMap = std::unordered_map<uint32_t, std::string>;
+  using SpecIdToInstMap = std::unordered_map<uint32_t, ir::Instruction*>;
+
+  // Constructs a pass instance with a map from spec ids to default values.
+  explicit SetSpecConstantDefaultValuePass(
+      const SpecIdToValueStrMap& default_values)
+      : Pass(), spec_id_to_value_(default_values) {}
+  explicit SetSpecConstantDefaultValuePass(SpecIdToValueStrMap&& default_values)
+      : Pass(), spec_id_to_value_(std::move(default_values)) {}
+
+  const char* name() const override { return "set-spec-const-default-value"; }
+  bool Process(ir::Module*) override;
+
+ private:
+  // The mapping from spec ids to their default values to be set.
+  const SpecIdToValueStrMap spec_id_to_value_;
+};
+
+}  // namespace opt
+}  // namespace spvtools
+
+#endif  // LIBSPIRV_OPT_SET_SPEC_CONSTANT_DEFAULT_VALUE_PASS_H_
--- a/test/opt/test_set_spec_const_default_value.cpp
+++ b/test/opt/test_set_spec_const_default_value.cpp
@@ -1,0 +1,433 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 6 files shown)


source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+#include <unordered_map>
+
+#include "def_use_manager.h"
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+const uint32_t target_id_operand_index = 0;
```
dneto0: Google style guide for these is  kTargetIdOperandIndex

dneto0: Also, why aren't these just in the function body?

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,108 @@
...
+bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
+  bool modified = false;
+  analysis::DefUseManager def_use_mgr(module);
+  // Scan through all the annotation instructions to find 'OpDecorate SpecId'
+  // instructions. Then extract the decoration target id of those instructions
+  // to get the target spec-constant-defining instructions. Replace the default
+  // values (the default bit patterns) with the corresponding values in the
+  // id_to_value_ map.
+  for (ir::Instruction& inst : module->annotations()) {
+    switch (inst.opcode()) {
```
dneto0: You're only doing one case, so might as well just do
```
if (inst.opcode() == SpvOp::SpvOpDecorate) {
...
}
```

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+#include "def_use_manager.h"
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+const uint32_t target_id_operand_index = 0;
+const uint32_t spec_id_operand_index = 2;
+const uint32_t opdecorate_num_operands = 3;
```
dneto0: This is true for SpecId decorations, but not all decorations.  Better to capture that in the name. Or have a comment saying we only care about that case.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,108 @@
...
+};
+
+bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
+  bool modified = false;
+  analysis::DefUseManager def_use_mgr(module);
+  // Scan through all the annotation instructions to find 'OpDecorate SpecId'
+  // instructions. Then extract the decoration target id of those instructions
+  // to get the target spec-constant-defining instructions. Replace the default
+  // values (the default bit patterns) with the corresponding values in the
+  // id_to_value_ map.
```
dneto0: This algorithm doesn't work when the decoration occurs through a decoration group.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,108 @@
...
+  for (ir::Instruction& inst : module->annotations()) {
+    switch (inst.opcode()) {
+      case SpvOp::SpvOpDecorate:
+        if (inst.NumOperands() == opdecorate_num_operands) {
+          uint32_t spec_id = inst.GetSingleWordOperand(spec_id_operand_index);
```
dneto0: You haven't checked that the decoration is actually a SpecId decoration.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,108 @@
...
+          uint32_t target_id =
+              inst.GetSingleWordOperand(target_id_operand_index);
+          ir::Instruction* spec_inst = def_use_mgr.GetDef(target_id);
+          auto iter = spec_id_to_value_.find(spec_id);
+          if (iter != spec_id_to_value_.end()) {
```
dneto0: Break up this method.  The "continue" you use below is a branch to quite a non-local spot.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Nice catch! Hmm, now I have two cases for the switch on the opcode....

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -0,0 +1,165 @@
...
+          "%1 = OpSpecConstant %int 2147483647\n",
+        },
+        // 4. 64-bit uint type
+        {
+          // code
+          "OpDecorate %1 SpecId 100\n"
+          "%ulong = OpTypeInt 64 0\n"
+          "%1 = OpSpecConstant %ulong 100\n",
+          // default values
+          SpecIdToValueMap{{100, {0xffffffff, 0xffffffff}}},
```
dneto0: Make the two words different so you know you've put them in the right order.  Suggest subtracing 1 from the first word.
That should subtract one from the decimal result.

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -0,0 +1,165 @@
...
+          "OpDecorate %2 SpecId 202\n"
+          "OpDecorate %3 SpecId 203\n"
+          "%bool = OpTypeBool\n"
+          "%int = OpTypeInt 32 1\n"
+          "%double = OpTypeFloat 64\n"
+          "%1 = OpSpecConstant %double 0x1.fffffffffffffp+1024\n"
+          "%2 = OpSpecConstant %int 2048\n"
+          "%3 = OpSpecConstantFalse %bool\n",
+        },
+    }));
```
dneto0: Add tests to show that true is returned when the pass modifies, and false when it does not modify.

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -0,0 +1,165 @@
...
+
```
dneto0: Add a test to show that you actually check the SpecId enum and not something liek ArrayStride that also has a uint parameter.
E.g.  OpDecorate %1 ArrayStride 4
      
and then try to modify something with a spec Id of 4.  Make sure that %1 target is not changed.

dneto0:
-1 Some bugs, missing tests, and some style.

test/opt/test_set_spec_const_default_value.cpp
Qining: `SinglePassRunAndCheck()` already checks the `modified` flag.

Qining:
This CL depends on #400 .

test/opt/test_set_spec_const_default_value.cpp
Qining: More tests added in 551c877

test/opt/test_set_spec_const_default_value.cpp
Qining: Updated in 551c877

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Updated in 551c877.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 551c877

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 551c877

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Comments added for constants in d21f2a9

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Updated in d21f2a9

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed and moved to function body in 551c877

Qining:
Rebased as #400 has been merged.
This CL also depends on #408.

Qining:
rebased to master as #408 is merged.

Qining:
PTAL @dneto0 @antiagainst

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+#include "set_spec_constant_default_value_pass.h"
+
+#include <cstring>
+#include <tuple>
+#include <unordered_map>
+#include <vector>
+
+#include "def_use_manager.h"
+#include "spirv-tools/libspirv.h"
```
dneto0: Nitpicky:  Order of includes:  Other libraries headers come before your own libraries.
The optimization library is separate from libSPIRV-Tools.a.

(I haven't been checking this bit very closely at all lately.)

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,229 @@
...
+using spvutils::ParseAndEncodeNumber;
+
+// Given a numeric value in string form and the expected type of the value,
+// parse the string and encode it in a vector of words. If the value is a
+// scalar integer or floating point value, encodes the value in SPIR-V encoding
+// format. If the value is 'false' or 'true', returns a vector with single
+// word with value 0 or 1 respectively. Returns the vector containing the
+// encoded value on success. Otherwise returns an empty vector.
+std::vector<uint32_t> ParseDefaultValueStr(const char* text,
+                                           const analysis::Type* type) {
```
dneto0: Shouldn't type be const-ref?  You're not changing it, and it can't be nullptr.?

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,229 @@
...
+  std::vector<uint32_t> result;
+  if (!strncmp(text, "true", 4)) {
```
dneto0: This will accept "trueish" as true.
Similarly for the false test.

source/opt/set_spec_constant_default_value_pass.cpp
dneto0: Oh, wait, are you going to use nullptr for boolean types?
In that case the first sentence of the comment is wrong: it's not always a numeric value being parsed.  (SPIR-V is explicit that numeric types exclude booleans.)

So the comment is wrong, needs updating.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,229 @@
...
+        nullptr);
+    // Clear the result vector on failure.
+    if (rc != EncodeNumberStatus::kSuccess) {
+      result.clear();
+    }
+  }
+  return result;
+}
+
+// Returns true if the given instruction's result id should have a SpecId
```
dneto0: Without context, this should be "could" not "should".

source/opt/set_spec_constant_default_value_pass.cpp
Qining: We down-cast the type pointer in this function, so it cannot be a const-reference.

It won't be nullptr for boolean type, it should be a `Bool`, but we only have the parent type pointer.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: fixed in 0d108d8

source/opt/set_spec_constant_default_value_pass.cpp
Qining: changed to `strcmp()`. Also updated the blurb to mention: "null-terminated c string" in 0d108d8

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Changed in 0d108d8

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+using spvutils::NumberType;
+using spvutils::EncodeNumberStatus;
+using spvutils::ParseAndEncodeNumber;
+
+// Given a numeric value in null-terminated c string form and the expected type
```
antiagainst: in a null-...

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+// of the value, parse the string and encode it in a vector of words. If the
```
antiagainst: parses

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -0,0 +1,54 @@
...
+#include "module.h"
+#include "pass.h"
+
+namespace spvtools {
+namespace opt {
+
+// The optimization pass that sets the default values for the spec constants
+// that have SpecId decorations (i.e. those defined by
+// OpSpecConstant{|True|False} instructions).
+class SetSpecConstantDefaultValuePass : public Pass {
```
antiagainst: Do we need to have the `Default` part in the name? It's just **fixing** the values of spec constants. Putting `Default` here let me feel like they are free to take some other value in some way.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+  std::vector<uint32_t> result;
+  if (!strcmp(text, "true")) {
+    result.push_back(1u);
+  } else if (!strcmp(text, "false")) {
+    result.push_back(0u);
+  } else {
+    NumberType number_type = {32, SPV_NUMBER_SIGNED_INT};
+    if (const auto* IT = type->AsInteger()) {
+      number_type.bitwidth = IT->width();
+      number_type.kind =
```
antiagainst: `if (!IT->IsSigned) number_type.kind = SPV_NUMBER_UNSIGNED_INT`?

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+    if (rc != EncodeNumberStatus::kSuccess) {
+      result.clear();
+    }
+  }
+  return result;
+}
+
+// Returns true if the given instruction's result id could have a SpecId
+// decoration.
+bool ShouldHaveSpecIdDecoration(const ir::Instruction& inst) {
```
antiagainst: Should? Could?

source/opt/set_spec_constant_default_value_pass.h
Qining: I think 'default value' here is more accurate. This is what the Spec says and I think it is true that other players in the pipe may change the value again later.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Changed to 'Can'.

But now I sort of feel that we should use 'should'. Because it is invalid that a Spec constant defined with `OpSpecConstant`, `OpSpecContantTrue` and `OpSpecConstantFalse` does not have `SpecId` decoration. This is the same for the blurb @dneto0 . To me it looks like the instruction must have a Spec Id decoration.

Anyway, not a serious problem in my opinion.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in de4c4a8

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in de4c4a8

source/opt/set_spec_constant_default_value_pass.h
antiagainst: Yeah, my fault. Suddenly realized that the "fixing" procedure is done in another pass.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -19,11 +19,12 @@
 #include <unordered_map>
 #include <vector>
 
+#include "util/parse_number.h"
+
 #include "def_use_manager.h"
 #include "spirv-tools/libspirv.h"
```
dneto0: spirv-tools/libspirv.h is also from libSPIRV-Tools.a, a different library.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+  for (ir::Instruction& inst : module->annotations()) {
+    // Only process 'OpDecorate SpecId' instructions
+    if (inst.opcode() != SpvOp::SpvOpDecorate) continue;
+    if (inst.NumOperands() != kOpDecorateSpecIdNumOperands) continue;
+    if (inst.GetSingleWordInOperand(kDecorationOperandIndex) !=
+        uint32_t(SpvDecoration::SpvDecorationSpecId)) {
+      continue;
+    }
+
+    // 'inst' is a OpDecorate SpecId instruction.
```
antiagainst: an

source/opt/set_spec_constant_default_value_pass.cpp
dneto0: I'm fine with the code as-is. Less stateful, more functional.

dneto0: Oh. I see the state was already set in the initializer.  Hm.  Either way is ok for me.

dneto0: Actually, I *DO* like the code the way it is.   That's because it's less sensitive to a change of the initializer.  (And I *do* want the initializer changed.)

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+// containing the encoded value on success. Otherwise returns an empty vector.
+std::vector<uint32_t> ParseDefaultValueStr(const char* text,
+                                           const analysis::Type* type) {
+  std::vector<uint32_t> result;
+  if (!strcmp(text, "true")) {
+    result.push_back(1u);
+  } else if (!strcmp(text, "false")) {
+    result.push_back(0u);
+  } else {
+    NumberType number_type = {32, SPV_NUMBER_SIGNED_INT};
```
dneto0: It turns out that all the untyped literals in SPIR-V are unsigned integers.  So I'd prefer this to be {32, SPV_NUMBER_UNSIGNED_INT}.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: So should I change the initializer to `NumberType number;` without any initial value?

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+    case SpvOp::SpvOpSpecConstantTrue:
+      return true;
+    default:
+      return false;
+  }
+}
+
+// Given a decoration group defining instruction that is decorated with SpecId
+// decoration, finds the spec constant defining instruction which is the real
+// consumer of the SpecId decoration. Returns the spec constant defining
```
dneto0: "consumer" -> "target" reads better I think.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Or I can change have the initial width = 0 and the initial type = SPV_NUMBER_NONE. I think this is cleaner

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+// instruction if such an instruction is found, otherwise returns a nullptr.
+ir::Instruction* GetSpecIdTargetFromDecorationGroup(
+    const ir::Instruction& decoration_group_defining_inst,
+    analysis::DefUseManager* def_use_mgr) {
+  // Find the OpGroupDecorate instruction which consumes the given decoration
+  // group. Note that the given decoration group has SpecId decoration, which
+  // is unique for different spec constants. So the decoration group cannot be
+  // consumed by different OpGroupDecorate instructions. Thereforth we only
```
dneto0: "Therefore"

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+  // need the first OpGroupDecortion instruction that uses the given decoration
```
dneto0: "OpGroupDecoration"

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -0,0 +1,230 @@
...
+        break;
+      case SpvOp::SpvOpSpecConstantTrue:
+        // If the new value is also 'true', no need to change anything.
+        // Otherwise, set the opcode to OpSpecConstantFalse;
+        if (static_cast<bool>(bit_pattern.front())) {
+          break;
+        } else {
+          spec_inst->SetOpcode(SpvOp::SpvOpSpecConstantFalse);
+          modified = true;
+        }
```
antiagainst:
```
if (!static_cast<bool>(...)) {
  spec_inst-> ...
  modified = true;
} break;
```

?

antiagainst: And the same for the rest two.

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -0,0 +1,334 @@
...
+        // 0. Empty.
+        {"", SpecIdToValueStrMap{}, ""},
+        // 1. Empty with non-empty values to set.
+        {"", SpecIdToValueStrMap{{1, "100"}, {2, "200"}}, ""},
+        // 2. Bool type.
+        {
+            // code
+            "OpDecorate %1 SpecId 100\n"
+            "%bool = OpTypeBool\n"
+            "%1 = OpSpecConstantTrue %bool\n",
```
antiagainst: Would be nice to also have a case to try to set a `True` to a `True`.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 8718600

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -0,0 +1,334 @@
...
+            "%1 = OpSpecConstantFalse %bool\n",
+        },
+        // 3. 32-bit int type.
+        {
+            // code
+            "OpDecorate %1 SpecId 100\n"
+            "OpDecorate %2 SpecId 101\n"
+            "%int = OpTypeInt 32 1\n"
+            "%1 = OpSpecConstant %int 10\n"
+            "%2 = OpSpecConstant %int 11\n",
```
antiagainst: Like bool, also a case change the value to the original value.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 8718600

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 8718600

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 8718600

test/opt/test_set_spec_const_default_value.cpp
antiagainst: Also cases for changing `False` to `True` & `False`>

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Updated in 8718600

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 8718600

source/opt/set_spec_constant_default_value_pass.cpp
Qining: fixed in 8718600

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -0,0 +1,334 @@
...
+            "%4 = OpSpecConstant %int 300\n",
+            // default values
+            SpecIdToValueStrMap{{100, "0xffffffff"}},
+            // expected
+            "OpDecorate %1 SpecId 100\n"
+            "%1 = OpDecorationGroup\n"
+            "OpGroupDecorate %1 %2 %3 %4\n"
+            "%int = OpTypeInt 32 1\n"
+            "%2 = OpConstant %int 100\n"
+            // Only the first spec constant target is handled.
```
antiagainst: Well, I think for this case, we should avoid anything and just return an error. (Error handling to be added to the framework though.)

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -0,0 +1,334 @@
...
+            // default values
+            SpecIdToValueStrMap{{100, "0x7fffffff"}, {4, "0x00000001"}},
+            // expected
+            "OpDecorate %1 SpecId 100\n"
+            "OpDecorate %1 ArrayStride 4\n"
+            "%1 = OpDecorationGroup\n"
+            "OpGroupDecorate %1 %2\n"
+            "%int = OpTypeInt 32 1\n"
+            "%2 = OpSpecConstant %int 2147483647\n",
+        },
```
antiagainst: Nice tests!

test/opt/test_set_spec_const_default_value.cpp
Qining: Fixed in 5f0626b

More tests added

test/opt/test_set_spec_const_default_value.cpp
Qining: Added in 5f0626b

test/opt/test_set_spec_const_default_value.cpp
Qining: added in 5f0626b

Qining:
All comments above have been resolved in 5f0626b

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -324,11 +402,29 @@ INSTANTIATE_TEST_CASE_P(
             "%1 = OpDecorationGroup\n"
             "OpGroupDecorate %1 %2 %3 %4\n"
             "%int = OpTypeInt 32 1\n"
-            "%2 = OpConstant %int 100\n"
+            "%2 = OpSpecConstant %int 100\n"
             // Only the first spec constant target is handled.
```
antiagainst: This line should be deleted now.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -104,22 +104,46 @@ ir::Instruction* GetSpecIdTargetFromDecorationGroup(
...
+  // returns a nullptr.
+  ir::Instruction* target_inst = nullptr;
+  for (uint32_t i = 1; i < group_decorate_inst->NumInOperands(); i++) {
     // All the operands of a OpGroupDecorate instruction should be of type
     // SPV_OPERAND_TYPE_ID.
-    uint32_t id = operand.words.front();
-    ir::Instruction* target_inst = def_use_mgr->GetDef(id);
-    if (target_inst && CanHaveSpecIdDecoration(*target_inst)) {
-      return target_inst;
+    uint32_t candidate_id = group_decorate_inst->GetInOperand(i).words.front();
```
antiagainst: GetSingleWordInOperand(i)

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -104,22 +104,46 @@ ir::Instruction* GetSpecIdTargetFromDecorationGroup(
...
+    ir::Instruction* candidate_inst = def_use_mgr->GetDef(candidate_id);
+
+    if (!candidate_inst) {
+      // The candidate instruction is not found, continue
```
antiagainst: This comment is redundant. It's already clear from the code.

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -245,6 +249,80 @@ INSTANTIATE_TEST_CASE_P(
...
+                                {202, "0.142857"}},
+            // expected
+            "OpDecorate %1 SpecId 201\n"
+            "OpDecorate %2 SpecId 202\n"
+            "%float = OpTypeFloat 32\n"
+            "%double = OpTypeFloat 64\n"
+            "%1 = OpSpecConstant %float 3.14159\n"
+            "%2 = OpSpecConstant %double 0.142857\n",
+        },
+        // 17. OpGroupDecorate may have multiple same target ids defined by a
```
antiagainst: may have multiple target ids of the same value. ?

Qining: Fixed in 3e73167

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 3e73167

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 3e73167

test/opt/test_set_spec_const_default_value.cpp
Qining: Deleted in 3e73167

antiagainst:
+1

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -0,0 +1,429 @@
...
+        // 3. 32-bit int type.
+        {
+            // code
+            "OpDecorate %1 SpecId 100\n"
+            "OpDecorate %2 SpecId 101\n"
+            "%int = OpTypeInt 32 1\n"
+            "%1 = OpSpecConstant %int 10\n"
+            "%2 = OpSpecConstant %int 11\n",
+            // default values
+            SpecIdToValueStrMap{{100, "2147483647"}, {101, "0xffffffff"}},
```
dneto0: I'd like to see a case for 32-bit int where the string value has a negative sign.

E.g.   "-42" should go through.  That makes sure you've asked to parse the right type.

dneto0:
-1  Need a test case to show that you've parsed "-42" correctly as a signed int.  Your signed int tests only have positive values.

dneto0:
+2

Issue #359, opened 2016-08-18
auto-(dis)assemble from various tools using spirv-dis/as


steve-lunarg:
As this isn't exactly an "issue", please feel free to delete if it's a breach of protocol.  I wasn't sure where else to put it, and I thought these might be handy for some people using SPIRV-Tools.
#### To allow the "less" pager (under *nix, BSD, and possibly cygwin) to transparently invoke spirv-dis on SPIRV binary files:

Place a script like this somewhere handy, and make it executable (chmod +x):

```
#!/bin/bash

case "$1" in
    *.spv) spirv-dis --raw-id "$1" 2>/dev/null;;
    *) exit 1;;
esac

exit $?
```

Then point the LESSOPEN variable to it, like so:

```
export LESSOPEN='|~/bin/lesspipe.sh "%s"'
```

Assuming spirv-dis is in your search path, you can now directly "less -R my_module.spv" on a binary and see it in a handy human readable form.  The -R is to preserve ANSI colors; you can add --no-color to the script if you don't like that.
#### To use spirv-dis and spirv-as to automatically (dis)assemble a SPIR-V binary file when loading or saving from emacs:

Place these lines into your .emacs file.  Then loading a binary will show the human readable asm in asm-mode in the editor buffer.  You can edit it within the constraints of validity, and when you re-save the file it will be assembled back to a SPIR-V binary.  Again you'll need the tools in your search path.  Handy for making hand tweaks to binary modules.

```
;; Handle auto-(dis)assemble of SPV binaries on save/load
(add-to-list 'jka-compr-compression-info-list
             '["\\.spv"
               "Assembling SPIRV" "spirv-as" ("-o" "-")
               "Disassembling SPIRV" "spirv-dis" ("--no-color" "--raw-id")
               t nil "\003\002\043\007"])

(add-to-list 'auto-mode-alist '("\\.spv\\'" . asm-mode))
```

dneto0:
These are good ideas!

I made https://github.com/KhronosGroup/SPIRV-Tools/pull/364 to capture the less feature.  Please take a look.  It adds a spirv-lesspipe program that gets installed appropriately. 
The only funny bit is the people might have different preferences for default flags.  E.g. I didn't keep the --raw-id flag in the script, and others might not regularly use -R.   But it's a reasonable default.
Please take a look and provide a +1 on that MR if you like it.

[...]

PR #399, opened 2016-09-02
add spirv-dis/spirv-as support for emacs when loading .spv binaries

steve-lunarg:
I've marked this PR as WIP for the moment, since I've not submitted anything to SPIRV-Tools before, have probably mangled something, and it could use a good looking over.  This PR adds automatic spirv-dis and spirv-as support when saving and loading binaries in emacs, per discussion in #359.

Some particulars to note:

- Unlike the other tools, this installs into an absolute path (ignoring CMAKE_INSTALL_PREFIX), since that is the only way it will be found upon emacs startup.  This might be considered intrusive, and the functionality is a bit niche at any rate.  Unclear to me if it's worthwhile in the balance, given it's only useful to emacs users.  That's a choice for SPIRV-Tools maintainers.  It might be a stronger case with a vi version as well, but I don't know how to make that.

- It will only install the file into /etc/emacs/site-start.d if that directory already exists, which is typically the case on systems with emacs installed, and not otherwise.  This avoids extraneous pollution.  The condition is evaluated only at cmake invocation time, not at "make install" time.

### DIFF
--- a/tools/emacs/CMakeLists.txt
+++ b/tools/emacs/CMakeLists.txt
@@ -1,0 +1,43 @@
+# Copyright (c) 2016 LunarG Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Install a script for use with the auto-compression feature of emacs(1).
+# Upon loading a file with the .spv extension, the file will be disassembled
+# using spirv-dis, and the result colorized with asm-mode in emacs.  The file
+# may be edited within the constraints of validity, and when re-saved will be
+# re-assembled using spirv-as.
+
+# It is required that those tools be in your PATH.  If that is not the case
+# when starting emacs, the path can be modified as in this example:
+# (setenv "PATH" (concat (getenv "PATH") ":/path/to/spirv/tools"))
+#
+# See https://github.com/KhronosGroup/SPIRV-Tools/issues/359
+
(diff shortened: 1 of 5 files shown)


tools/emacs/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+# See https://github.com/KhronosGroup/SPIRV-Tools/issues/359
+
+# This is an absolute directory, and ignores CMAKE_INSTALL_PREFIX, or
+# it will not be found by emacs upon startup.  It is only installed if
+# the directory /etc/emacs/site-start.d already exists *at the time of
+# cmake invocation, not at the time of make install*.  This is
+# typically true if emacs is installed on the system, and not
+# otherwise.
+
+if(EXISTS /etc/emacs/site-start.d)
```
dneto0: As you noted, this ignores the CMAKE_INSTALL_PREFIX.   I think that's intrusive-by-default, so I'd like this to be guarded by an explicit opt-in cmake setting, e.g SPIRV_TOOLS_INSTALL_EMACS_HELPERS.

dneto0:
Thanks for the contribution.  Two requests:
- guard the install by an opt-in cmake setting
- mention this functionality in the README.

Caveat for this feature is that the assembler assigns numbers to IDs in sequential fashion, so edit+save with this feature will change ID numbering by default.  That merits mention in the plugin doc and in the README.  (And perhaps I should extend the assembly syntax to allow ID preservation, but that's another story.)

In other news, I took a few minutes to see about equivalent functionality for vim.  It looks doable, but I didn't pursue it all the way.  Definitely not necessary for you to pursue; I was just curious.

steve-lunarg:
OK, will do.  (Most likely in ~1 week).

steve-lunarg:
>   guard the install by an opt-in cmake setting
>   mention this functionality in the README.

I updated the PR for those.  It also mentions that a load/edit/save cycle will change the IDs.  If there's ever an option to preserve IDs in spirv-as, it'll be easy to add that in.

dneto0:
I added another commit to add the new variable as a CMake option.
Otherwise I get a CMake error if I define it on the command line.  I'm not clear why it worked for you when set.  Anyway, it's now merged with the tweak/fix.

PR #401, opened 2016-09-04
1. Callback mechanism for error handling

antiagainst:
Every time an event happens in the library that the user should be
aware of, the callback will be invoked.

The existing diagnostic mechanism is hijacked internally by a
callback that creates an diagnostic object each time an event
happens.

### DIFF
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -61,6 +61,7 @@ class Parser {
          spv_parsed_header_fn_t parsed_header_fn,
          spv_parsed_instruction_fn_t parsed_instruction_fn)
       : grammar_(context),
+        consumer_(context->consumer),
         user_data_(user_data),
         parsed_header_fn_(parsed_header_fn),
         parsed_instruction_fn_(parsed_instruction_fn) {}
@@ -120,8 +121,7 @@ class Parser {
   // returned object will be propagated to the current parse's diagnostic
   // object.
   libspirv::DiagnosticStream diagnostic(spv_result_t error) {
-    return libspirv::DiagnosticStream({0, 0, _.word_index}, _.diagnostic,
-                                      error);
+    return libspirv::DiagnosticStream({0, 0, _.word_index}, consumer_, error);
   }
 
   // Returns a diagnostic stream object with the default parse error code.
@@ -156,6 +156,7 @@ class Parser {
   // Data members
 
   const libspirv::AssemblyGrammar grammar_;        // SPIR-V syntax utility.
+  const spvtools::MessageConsumer& consumer_;      // Message consumer callback.
   void* const user_data_;                          // Context for the callbacks
   const spv_parsed_header_fn_t parsed_header_fn_;  // Parsed header callback
   const spv_parsed_instruction_fn_t
@@ -752,7 +753,12 @@ spv_result_t spvBinaryParse(const spv_const_context context, void* user_data,
                             spv_parsed_header_fn_t parsed_header,
                             spv_parsed_instruction_fn_t parsed_instruction,
                             spv_diagnostic* diagnostic) {
-  Parser parser(context, user_data, parsed_header, parsed_instruction);
+  spv_context_t hijack_context = *context;
+  if (diagnostic) {
+    *diagnostic = nullptr;
+    libspirv::UseDiagnosticAsMessageConsumer(&hijack_context, diagnostic);
+  }
+  Parser parser(&hijack_context, user_data, parsed_header, parsed_instruction);
   return parser.parse(code, num_words, diagnostic);
 }
 
--- a/source/diagnostic.cpp
+++ b/source/diagnostic.cpp
@@ -20,6 +20,7 @@
 #include <iostream>
 
 #include "spirv-tools/libspirv.h"
+#include "table.h"
 
 // Diagnostic API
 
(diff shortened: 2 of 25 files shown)


dneto0:
I don't think you need to do this.  When you want to substitute your own message consumer, make a *new* context object by cloning the old one, and set the message consumer in that new one.  Then destroy it at the end of the call.
For one thing that preserves the const-ness and therefore implicitly continues to allow concurrent use of the outside context object.

antiagainst:
@dneto0: Very nice suggestion. Done.

source/diagnostic.h
```diff
@@ -59,10 +56,18 @@ class DiagnosticStream {
...
+  const spvtools::MessageConsumer& consumer_;  // Message consumer callback.
   spv_result_t error_;
 };
 
+// Changes the MessageConsumer in |context| to one that updates |diagnostic|
+// with the last message received.
+//
+// This function expects that |diagnostic| is not nullptr and its content is a
+// nullptr.
+void UseDiagnosticAsMessageConsumer(spv_context context,
```
Qining: Is this a static function?

source/binary.cpp
```diff
@@ -752,7 +753,13 @@ spv_result_t spvBinaryParse(const spv_const_context context, void* user_data,
                             spv_parsed_header_fn_t parsed_header,
                             spv_parsed_instruction_fn_t parsed_instruction,
                             spv_diagnostic* diagnostic) {
-  Parser parser(context, user_data, parsed_header, parsed_instruction);
+  if (!context->consumer && !diagnostic) return SPV_ERROR_INVALID_DIAGNOSTIC;
```
dneto0: I don't like this.
I should be able to use the binary parser without collecting detailed feedback. I could trust the binaries I'm passing in.

dneto0: I should have had a test for that case.  Please add one.

source/table.cpp
```diff
@@ -41,7 +41,15 @@ spv_context spvContextCreate(spv_target_env env) {
   spvOperandTableGet(&operand_table, env);
   spvExtInstTableGet(&ext_inst_table, env);
 
-  return new spv_context_t{env, opcode_table, operand_table, ext_inst_table};
+  return new spv_context_t{
+      env, opcode_table, operand_table, ext_inst_table,
+      nullptr /* an empty MessageConsumer */
```
dneto0: Is this a good default?  Should we instead default to a message consumer that just does nothing with the message?  Then it's always safe to call.

dneto0: Then again I see how later you can avoid constructing objects in the failure case.  So I can see the case on efficiency grounds....

antiagainst: That's reasonable.

It's just conceptually more clear to use a null std::function here to mean no consumer at all. But actually doesn't bring any benefits other than that.

I'll change this to a default consumer just ignoring messages.

source/binary.cpp
dneto0: I've merged a test case for this scenario.

antiagainst: I see. Revised in 5c7107d.

source/table.cpp
antiagainst: Revised in 5c7107d.

source/table.h
```diff
@@ -87,8 +88,12 @@ struct spv_context_t {
   const spv_opcode_table opcode_table;
   const spv_operand_table operand_table;
   const spv_ext_inst_table ext_inst_table;
+  spvtools::MessageConsumer consumer;
 };
 
+void SetContextMessageConsumer(spv_context context,
```
dneto0: Needs blurb.

source/table.cpp
```diff
@@ -41,7 +41,16 @@ spv_context spvContextCreate(spv_target_env env) {
   spvOperandTableGet(&operand_table, env);
   spvExtInstTableGet(&ext_inst_table, env);
 
-  return new spv_context_t{env, opcode_table, operand_table, ext_inst_table};
+  return new spv_context_t{
+      env, opcode_table, operand_table, ext_inst_table,
+      /* A deault message consumer that ignores all messages */
+      [](spvtools::MessageLevel, const char*, const spv_position_t&,
```
dneto0: Now that I know a std::function is comparable to nullptr and assignable from nullptr, then we may as well make the default nullptr since the user could always have set a nullptr function.  When we will invoke the consumer, we always have to check for the nullptr anyways.
Sorry about the flip flop.

source/diagnostic.cpp
```diff
@@ -68,12 +69,47 @@ spv_result_t spvDiagnosticPrint(const spv_diagnostic diagnostic) {
...
+      case SPV_ERROR_INVALID_TABLE:
+        level = MessageLevel::InternalError;
+        break;
+      case SPV_ERROR_OUT_OF_MEMORY:
+        level = MessageLevel::Fatal;
+        break;
+      default:
+        break;
+    }
+    consumer_(level, "", position_, stream_.str().c_str());
```
dneto0: I'm surprised and impressed that this is the only place where we ever used to call spvDiagnosticCreate.  Whew!

test/BinaryParse.cpp
```diff
@@ -258,6 +260,94 @@ TEST_F(BinaryParseTest, NullDiagnosticsIsOkForBadParse) {
...
+  EXPECT_CALL(client_, Instruction(_)).Times(0);  // No instruction callback.
+  EXPECT_EQ(SPV_SUCCESS,
+            spvBinaryParse(ctx, &client_, words.data(), words.size(),
+                           invoke_header, invoke_instruction, nullptr));
+  EXPECT_EQ(0, invocation);
+
+  spvContextDestroy(ctx);
+}
+
+TEST_F(BinaryParseTest, SpecifyConsumerNullDiagnosticsForBadParse) {
```
dneto0: You also need a test where diagnostic is nullptr, and you've set a nullptr for the message consumer.
That makes sure you've guarded the call to the consumer with a nullptr check.

source/table.cpp
dneto0: This also regains the efficiency of not creating objects in the nullptr case.

source/table.h
antiagainst: Done.

source/table.cpp
antiagainst: Done in the third commit.

source/diagnostic.cpp
antiagainst: Right! I'm surprised too when changing this.

test/BinaryParse.cpp
antiagainst: Done in the third commit.

PR #405, opened 2016-09-07
Prefix test binaries with `test_` and create separate binaries for validation tests.

antiagainst:
Two commits included.


dneto0:
+2
cc' @umar456 The validation tests are being moved/refactored into different executables.

PR #406, opened 2016-09-08
3. Add MessageConsumer to PassManager, Pass, and analysis interfaces.

antiagainst:
Depends on #362.

Use the callback error-handling mechanism introduced in #401
in Pass, PassManager, and analyses.

Also convert some uses of assert() in optimization code to use
SPIRV_ASSERT(), SPIRV_UNIMPLEMENTED(), or SPIRV_UNREACHABLE()
accordingly introduced in #362 .

### DIFF
--- a/source/message.h
+++ b/source/message.h
@@ -16,6 +16,7 @@
 #define SPIRV_TOOLS_MESSAGE_H_
 
 #include <functional>
+#include <string>
 
 #include "spirv-tools/libspirv.h"
 
@@ -41,6 +42,17 @@ using MessageConsumer = std::function<void(
     MessageLevel /* level */, const char* /* source */,
     const spv_position_t& /* position */, const char* /* message */
     )>;
+
+// A message consumer that ignores all messages.
+inline void IgnoreMessage(MessageLevel, const char*, const spv_position_t&,
+                          const char*) {}
+
+// A helper function to compose and return a string from the message in the
+// following format:
+//   "<level>: <source>:<line>:<column>:<index>: <message>"
+std::string StringifyMessage(MessageLevel level, const char* source,
+                             const spv_position_t& position,
+                             const char* message);
 
 }  // namespace spvtools
 
--- a/source/opt/def_use_manager.cpp
+++ b/source/opt/def_use_manager.cpp
@@ -14,10 +14,10 @@
 
 #include "def_use_manager.h"
 
-#include <cassert>
 #include <functional>
 
 #include "instruction.h"
+#include "log.h"
 #include "module.h"
 #include "reflect.h"
 
@@ -80,11 +80,10 @@ const UseList* DefUseManager::GetUses(uint32_t id) const {
   return &iter->second;
 }
 
-std::vector<ir::Instruction*> DefUseManager::GetAnnotations(
-    uint32_t id) const {
+std::vector<ir::Instruction*> DefUseManager::GetAnnotations(uint32_t id) const {
   std::vector<ir::Instruction*> annos;
   const auto* uses = GetUses(id);
-  if (!uses)  return annos;
+  if (!uses) return annos;
   for (const auto& c : *uses) {
     if (ir::IsAnnotationInst(c.inst->opcode())) {
       annos.push_back(c.inst);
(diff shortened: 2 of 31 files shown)


source/opt/def_use_manager.cpp
```diff
@@ -121,11 +120,12 @@ bool DefUseManager::ReplaceAllUsesWith(uint32_t before, uint32_t after) {
       if (it->inst->type_id() != 0 && it->operand_index == 0) {
         it->inst->SetResultType(after);
       } else if (it->inst->type_id() == 0) {
-        assert(false &&
-               "Result type id considered as using while the instruction "
-               "doesn't have a result type id.");
+        SPIRV_ASSERT(consumer_, false,
+                     "Result type id considered as use while the instruction "
+                     "doesn't have a result type id.");
```
dneto0: This is a regression from previous behaviour.  On a debug build this no longer halts the program.
Suggest that SPIRV_ASSERT* and friends themselves call assert(...) in the failure case.

source/message.h
```diff
@@ -42,6 +44,25 @@ using MessageConsumer = std::function<void(
...
+std::string StringifyMessage(MessageLevel level, const char* source,
+                             const spv_position_t& position,
+                             const char* message);
+
+// Prints the given message to stderr following the format in
+// StringifyMessage().
+inline void PrintMessageToStderr(MessageLevel level, const char* source,
+                                 const spv_position_t& position,
+                                 const char* message) {
+  std::cerr << StringifyMessage(level, source, position, message);
```
dneto0: Should hide this body.  Then you won't need to #include iostream and make client compilation slower.
Also, I suggest having the following more general thing:
```
inline MessageConsumer MessagePrinterToStream(std::ostream* out) { 
 return [out](MessageLevel level, const char* source, const spv_position_t& position, const char* message) { 
    *out << StringifyMessage(level, source, position, message);
};
}
```
I think you shouldn't provide PrintMessageToStderr once you have the more general thing.  In a real program you'll only see the std::cerr case once, in the main.cpp file.

source/opt/def_use_manager.h
```diff
@@ -106,13 +112,21 @@ class DefUseManager {
...
-  IdToUsesMap id_to_uses_;  // Mapping from ids to their uses
+  const MessageConsumer& consumer_;  // Message consumer.
+  IdToDefMap id_to_def_;             // Mapping from ids to their definitions
+  IdToUsesMap id_to_uses_;           // Mapping from ids to their uses
   // Mapping from instructions to the ids used in the instructions generating
   // the result ids.
   InstToUsedIdsMap inst_to_used_ids_;
 };
 
+inline DefUseManager::DefUseManager(const MessageConsumer& consumer,
```
dneto0: I would just put this body right at the declaration.

source/opt/def_use_manager.h
```diff
@@ -106,13 +112,21 @@ class DefUseManager {
...
+                                    ir::Module* module)
+    : consumer_(consumer) {
+  (void)consumer_;
```
dneto0: The (void)consumer_ is not needed is it?

test/opt/pass_fixture.h
```diff
@@ -39,7 +40,9 @@ template <typename TestT>
 class PassTest : public TestT {
  public:
   PassTest()
-      : tools_(SPV_ENV_UNIVERSAL_1_1), manager_(new opt::PassManager()) {}
+      : consumer_(PrintMessageToStderr),
```
dneto0: For tests, I want the default to be *no* output.

test/opt/test_def_use.cpp
```diff
@@ -132,7 +132,7 @@ TEST_P(ParseDefUseTest, Case) {
   ASSERT_NE(nullptr, module);
 
   // Analyze def and use.
-  opt::analysis::DefUseManager manager(module.get());
+  opt::analysis::DefUseManager manager(PrintMessageToStderr, module.get());
```
dneto0: I prefer *no* output in tests.

test/opt/test_pass_manager.cpp
```diff
@@ -29,23 +29,25 @@ using ::testing::Eq;
...
+      : NullPass(c) {}
+  NullPassWithArgs(const MessageConsumer& c, const std::vector<int>&, uint32_t)
+      : NullPass(c) {}
 
   const char* name() const override { return "null-with-args"; }
 };
 
 TEST(PassManager, Interface) {
-  opt::PassManager manager;
+  opt::PassManager manager(PrintMessageToStderr);
```
dneto0: Ignore output in tests.  (Unless you're testing the output.)

source/opt/def_use_manager.cpp
antiagainst: Good catch! Done in the second commit. But I used `std::exit`, which does the "all C streams are flushed and closed" cleanup.

source/message.h
antiagainst: Removed.

source/opt/def_use_manager.h
antiagainst: Moved.

source/opt/def_use_manager.h
antiagainst: It's needed since we only have two SPIRV_ASSERT using it. Will error out on release build b/c unused field. I moved this statement to where SPIRV_ASSERT uses consumer to make it more readable.

test/opt/pass_fixture.h
antiagainst: Using IgnoreMessage instead.

test/opt/test_def_use.cpp
antiagainst: Done.

test/opt/test_pass_manager.cpp
antiagainst: Done.

PR #407, opened 2016-09-08
Merge diagnostic tests and single them out in a separate binary.

antiagainst:


dneto0:
+2
I don't think there was a huge benefit to merging into one file.  But it's ok.

PR #409, opened 2016-09-09
Build "spec id->default val str" mapping from string

Qining:
Add function `ParseDefaultValuesString()` to build the spec id->default
value string mapping required by `SetSpecConstantDefaultValuePass`.

This CL depends on #390

### DIFF
--- a/source/opt/set_spec_constant_default_value_pass.cpp
+++ b/source/opt/set_spec_constant_default_value_pass.cpp
@@ -15,6 +15,8 @@
 #include "set_spec_constant_default_value_pass.h"
 
 #include <cstring>
+#include <cctype>
+#include <string>
 #include <tuple>
 #include <unordered_map>
 #include <vector>
@@ -23,6 +25,7 @@
 #include "util/parse_number.h"
 
 #include "def_use_manager.h"
+#include "make_unique.h"
 #include "type_manager.h"
 #include "types.h"
 
@@ -32,6 +35,7 @@ namespace opt {
 namespace {
 using spvutils::NumberType;
 using spvutils::EncodeNumberStatus;
+using spvutils::ParseNumber;
 using spvutils::ParseAndEncodeNumber;
 
 // Given a numeric value in a null-terminated c string and the expected type of
@@ -246,6 +250,59 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
     // ids.
   }
   return modified;
+}
+
+// Returns true if the given char is ':', '\0' or considered as blank space
+// (i.e.: '\n', '\r', '\v', '\t', '\f' and ' ').
+bool IsSeparator(char ch) {
+  return std::strchr(":\0", ch) || std::isspace(ch) != 0;
+}
+
+std::unique_ptr<SetSpecConstantDefaultValuePass::SpecIdToValueStrMap>
+SetSpecConstantDefaultValuePass::ParseDefaultValuesString(const char* str) {
+  if (!str) return nullptr;
+
+  auto spec_id_to_value = MakeUnique<SpecIdToValueStrMap>();
+
+  // The parsing loop, break when points to the end.
+  while (*str) {
+    // Find the spec id.
+    while (std::isspace(*str)) str++;  // skip leading spaces.
+    const char* entry_begin = str;
+    while (!IsSeparator(*str)) str++;
+    const char* entry_end = str;
+    std::string spec_id_str(entry_begin, entry_end - entry_begin);
+    uint32_t spec_id = 0;
+    if (!ParseNumber(spec_id_str.c_str(), &spec_id)) {
+      // The spec id is not a valid uint32 number.
+      return nullptr;
+    }
+    auto iter = spec_id_to_value->find(spec_id);
+    if (iter != spec_id_to_value->end()) {
+      // Same spec id has been defined before
+      return nullptr;
+    }
+    // Find the ':', spaces between the spec id and the ':' are not allowed.
+    if (*str++ != ':') {
+      // ':' not found
+      return nullptr;
+    }
+    // Find the value string
+    const char* val_begin = str;
+    while (!IsSeparator(*str)) str++;
+    const char* val_end = str;
+    if (val_end == val_begin) {
+      // Value string is empty.
+      return nullptr;
+    }
+    // Update the mapping with spec id and value string.
+    (*spec_id_to_value)[spec_id] = std::string(val_begin, val_end - val_begin);
+
+    // Skip trailing spaces.
+    while (std::isspace(*str)) str++;
+  }
+
+  return spec_id_to_value;
 }
 
 }  // namespace opt
--- a/source/opt/set_spec_constant_default_value_pass.h
+++ b/source/opt/set_spec_constant_default_value_pass.h
@@ -43,6 +43,42 @@ class SetSpecConstantDefaultValuePass : public Pass {
   const char* name() const override { return "set-spec-const-default-value"; }
   bool Process(ir::Module*) override;
 
+  // Parses the given null-terminated C string to get a mapping from Spec Id to
+  // default value strings. Returns a unique pointer of the mapping from spec
+  // ids to spec constant default value strings built from the given |str| on
+  // success. Returns a nullptr if the given string is not valid for building
+  // the mapping.
+  // A valid string for building the mapping should follow the rule below:
+  //
+  //  "<spec id A>:<default value for A> <spec id B>:<default value for B> ..."
+  //  Example:
+  //    "200:0x11   201:3.14   202:1.4728"
+  //
+  //  Entries are separated with blank spaces (i.e.:' ', '\n', '\r', '\t',
+  //  '\f', '\v'). Each entry corresponds to a Spec Id and default value pair.
+  //  Multiple spaces between, before or after entries are allowed. However,
+  //  spaces are not allowed within spec id or the default value string because
+  //  spaces are always considered as delimiter to separate entries.
+  //
+  //  In each entry, the spec id and value string is separated by ':'. Missing
+  //  ':' in any entry is invalid. And it is invalid to have blank spaces in
+  //  between the spec id and ':' or the default value and ':'.
+  //
+  //  <spec id>: specifies the spec id value.
+  //    The text must represent a valid uint32_t number.
+  //    Hex format with '0x' prefix is allowed.
+  //    Empty <spec id> is not allowed.
+  //    One spec id value can only be defined once, multiple default values
+  //      defined for the same spec id is not allowed. Spec ids with same value
+  //      but different formats (e.g. 0x100 and 256) are considered the same.
+  //
+  //  <default value>: the default value string.
+  //    Spaces before and after default value text is allowed.
+  //    Spaces within the text is not allowed.
+  //    Empty <default value> is not allowed.
+  static std::unique_ptr<SpecIdToValueStrMap> ParseDefaultValuesString(
+      const char* str);
+
  private:
   // The mapping from spec ids to their default values to be set.
   const SpecIdToValueStrMap spec_id_to_value_;
--- a/test/opt/test_set_spec_const_default_value.cpp
+++ b/test/opt/test_set_spec_const_default_value.cpp
@@ -14,11 +14,120 @@
 
 #include "pass_fixture.h"
 
+#include <gmock/gmock.h>
+
 namespace {
 using namespace spvtools;
 
+using testing::Eq;
+
 using SpecIdToValueStrMap =
     opt::SetSpecConstantDefaultValuePass::SpecIdToValueStrMap;
+
+struct DefaultValuesStringParsingTestCase {
+  const char* default_values_str;
+  bool expect_success;
+  SpecIdToValueStrMap expected_map;
+};
+
+using DefaultValuesStringParsingTest =
+    ::testing::TestWithParam<DefaultValuesStringParsingTestCase>;
+
+TEST_P(DefaultValuesStringParsingTest, TestCase) {
+  const auto& tc = GetParam();
+  auto actual_map =
+      opt::SetSpecConstantDefaultValuePass::ParseDefaultValuesString(
+          tc.default_values_str);
+  if (tc.expect_success) {
+    EXPECT_NE(nullptr, actual_map);
+    if (actual_map) EXPECT_THAT(*actual_map, Eq(tc.expected_map));
+  } else {
+    EXPECT_EQ(nullptr, actual_map);
+  }
+}
+
+INSTANTIATE_TEST_CASE_P(
+    ValidString, DefaultValuesStringParsingTest,
+    ::testing::ValuesIn(std::vector<DefaultValuesStringParsingTestCase>{
+        // 0. empty map
+        {"", true, SpecIdToValueStrMap{}},
+        // 1. one pair
+        {"100:1024", true, SpecIdToValueStrMap{{100, "1024"}}},
+        // 2. two pairs
+        {"100:1024 200:2048", true,
+         SpecIdToValueStrMap{{100, "1024"}, {200, "2048"}}},
+        // 3. spaces between entries
+        {"100:1024 \n \r \t \v \f 200:2048", true,
+         SpecIdToValueStrMap{{100, "1024"}, {200, "2048"}}},
+        // 4. \t, \n, \r and spaces before spec id
+        {"   \n \r\t \t \v \f 100:1024", true,
+         SpecIdToValueStrMap{{100, "1024"}}},
+        // 5. \t, \n, \r and spaces after value string
+        {"100:1024   \n \r\t \t \v \f ", true,
+         SpecIdToValueStrMap{{100, "1024"}}},
+        // 6. maximum spec id
+        {"4294967295:0", true, SpecIdToValueStrMap{{4294967295, "0"}}},
+        // 7. minimum spec id
+        {"0:100", true, SpecIdToValueStrMap{{0, "100"}}},
+        // 8. random content without spaces are allowed
+        {"200:random_stuff", true, SpecIdToValueStrMap{{200, "random_stuff"}}},
+        // 9. support hex format spec id (just because we use the
+        // ParseNumber() utility)
+        {"0x100:1024", true, SpecIdToValueStrMap{{256, "1024"}}},
+        // 10. multiple entries
+        {"101:1 102:2 103:3 104:4 200:201 9999:1000 0x100:333", true,
+         SpecIdToValueStrMap{{101, "1"},
+                             {102, "2"},
+                             {103, "3"},
+                             {104, "4"},
+                             {200, "201"},
+                             {9999, "1000"},
+                             {256, "333"}}},
+        // 11. default value in hex float format
+        {"100:0x0.3p10", true, SpecIdToValueStrMap{{100, "0x0.3p10"}}},
+        // 12. default value in decimal float format
+        {"100:1.5e-13", true, SpecIdToValueStrMap{{100, "1.5e-13"}}},
+    }));
+
+INSTANTIATE_TEST_CASE_P(
+    InvalidString, DefaultValuesStringParsingTest,
+    ::testing::ValuesIn(std::vector<DefaultValuesStringParsingTestCase>{
+        // 0. missing default value
+        {"100:", false, SpecIdToValueStrMap{}},
+        // 1. spec id is not an integer
+        {"100.0:200", false, SpecIdToValueStrMap{}},
+        // 2. spec id is not a number
+        {"something_not_a_number:1", false, SpecIdToValueStrMap{}},
+        // 3. only spec id number
+        {"100", false, SpecIdToValueStrMap{}},
+        // 4. same spec id defined multiple times
+        {"100:20 100:21", false, SpecIdToValueStrMap{}},
+        // 5. Multiple definition of an identical spec id in different forms
+        // is not allowed
+        {"0x100:100 256:200", false, SpecIdToValueStrMap{}},
+        // 6. empty spec id
+        {":3", false, SpecIdToValueStrMap{}},
+        // 7. only colon
+        {":", false, SpecIdToValueStrMap{}},
+        // 8. spec id overflow
+        {"4294967296:200", false, SpecIdToValueStrMap{}},
+        // 9. spec id less than 0
+        {"-1:200", false, SpecIdToValueStrMap{}},
+        // 10. nullptr
+        {nullptr, false, SpecIdToValueStrMap{}},
+        // 11. only a number is invalid
+        {"1234", false, SpecIdToValueStrMap{}},
+        // 12. invalid entry separator
+        {"12:34;23:14", false, SpecIdToValueStrMap{}},
+        // 13. invalid spec id and default value separator
+        {"12@34", false, SpecIdToValueStrMap{}},
+        // 14. spaces before colon
+        {"100   :1024", false, SpecIdToValueStrMap{}},
+        // 15. spaces after colon
+        {"100:   1024", false, SpecIdToValueStrMap{}},
+        // 16. spec id represented in hex float format is invalid
+        {"0x3p10:200", false, SpecIdToValueStrMap{}},
+    }));
 
 struct SetSpecConstantDefaultValueTestCase {
   const char* code;


source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,93 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+bool IsEnd(const char* const ptr) {
+  if (*ptr == '\0' || *ptr == '}') {
+    return true;
+  }
+  return false;
+}
+
+// Returns true if the char pointed by the given pointer is ' ', '\n', '\t',
+// '\r'.
+bool IsSpace(const char* const ptr) {
```
antiagainst: `std::isspace`? But I kinda remember there are some pitfalls of using that? @dneto0

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,6 +43,39 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
+  // strings. Returns a unique pointer of the mapping from spec ids to spec
+  // constant default value strings built from the given |str| on success. If
+  // the given string is not valid for building the mapping, returns a nullptr.
+  // A valid string for building the mapping should follow the below example:
+  //
+  //  { <spec id A> : <default value for A>, <spec id B> : <default value for
+  //  B>, ...}
+  //
+  //  The string must starts with '{', and should ends with '}'. However the
+  //  missing the ending '}' may still be valid for parsing.
```
antiagainst: Let's be rigid. Missing `}` should be considered invalid. Allowing special cases may bite us in unexpected way in the future.

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,6 +43,39 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
+  //
+  //  Entries are separated with ','. Each entry is corresponding to a Spec Id
+  //  and default value pair. Empty entry is not allowed.
```
antiagainst: Empty entries.

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,6 +43,39 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
+  //
+  //  In each entry, the spec id and value string is separated by ':'. Missing
+  //  ':' in any entry is not allowed.
+  //
+  //  <spec id>: specifies the spec id value.
+  //    the text must represent a valid uint32_t number.
+  //    spaces before and afer the number text is allowed.
+  //    hex format with '0x' prefix is allowed.
+  //    spaces withing the number text is not allowed.
```
antiagainst: within

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,6 +43,39 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
+  //    empty <spec id> is not allowed.
+  //
+  //  <default value>: the defalut value string.
```
antiagainst: default

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,93 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+bool IsEntryContent(const char* const ptr) {
+  if (!IsSpace(ptr) && !IsEnd(ptr) && *ptr != ',' && *ptr != ':') {
+    return true;
+  }
+  return false;
+}
+
+std::unique_ptr<SetSpecConstantDefaultValuePass::SpecIdToValueStrMap>
+SetSpecConstantDefaultValuePass::ParseDefaultValuesString(const char* str) {
+  if (*str++ != '{') return nullptr;
```
antiagainst: Should we guard against `str == nullptr`?

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,93 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
   return modified;
 }
 
+// Returns true if the char pointed by the given pointer is '\0' or '}'.
+bool IsEnd(const char* const ptr) {
```
antiagainst: guard against `ptr == nullptr`? Or should be clear in the comment `ptr` should be non-`nullptr`.

Besides, this can be labelled as `inline` function.

`return *ptr == '\0' || *ptr == '}'`

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,6 +43,39 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
+  // the given string is not valid for building the mapping, returns a nullptr.
+  // A valid string for building the mapping should follow the below example:
+  //
+  //  { <spec id A> : <default value for A>, <spec id B> : <default value for
+  //  B>, ...}
+  //
+  //  The string must starts with '{', and should ends with '}'. However the
+  //  missing the ending '}' may still be valid for parsing.
+  //
+  //  Entries are separated with ','. Each entry is corresponding to a Spec Id
```
antiagainst: corresponds to

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,93 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+    const char* entry_end = str;
+    std::string spec_id_str(entry_begin, entry_end - entry_begin);
+    uint32_t spec_id = 0;
+    if (!ParseNumber(spec_id_str.c_str(), &spec_id)) {
+      // The spec id is not a valid uint32 number.
+      return nullptr;
+    }
+    auto iter = spec_id_to_value->find(spec_id);
+    if (iter != spec_id_to_value->end()) {
+      // defining spec id multiple times
```
antiagainst: This should also be documented in the function blurb.

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -20,6 +20,120 @@ using namespace spvtools;
...
+  // const auto& tc = GetParam();
+  // SinglePassRunAndCheck<opt::SetSpecConstantDefaultValuePass>(
+  // tc.code, tc.expected, [> skip_nop = <] false, tc.default_values);
+  const auto& tc = GetParam();
+  auto actual_map =
+      opt::SetSpecConstantDefaultValuePass::ParseDefaultValuesString(
+          tc.default_values_str);
+  if (tc.expect_success) {
+    EXPECT_NE(nullptr, actual_map);
+    EXPECT_EQ(tc.expected_map, *actual_map.get());
```
antiagainst: I think `*actual_map` will do here?

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -20,6 +20,120 @@ using namespace spvtools;
...
+        {"{100:1024   \n \r\t \t  }", true, SpecIdToValueStrMap{{100, "1024"}}},
+        // 8. maximum spec id
+        {"{4294967295:0}", true, SpecIdToValueStrMap{{4294967295, "0"}}},
+        // 9. minimum spec id
+        {"{0:100}", true, SpecIdToValueStrMap{{0, "100"}}},
+        // 10. random content without spaces are allowed
+        {"{200: random_stuff}", true,
+         SpecIdToValueStrMap{{200, "random_stuff"}}},
+        // 11. missing ending '}' is fine for parsing
+        {"{100:1024", true, SpecIdToValueStrMap{{100, "1024"}}},
```
antiagainst: I would suggest let's fail this case.

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -20,6 +20,120 @@ using namespace spvtools;
...
+    EXPECT_EQ(tc.expected_map, *actual_map.get());
+  } else {
+    EXPECT_EQ(nullptr, actual_map);
+  }
+}
+
+INSTANTIATE_TEST_CASE_P(
+    ValidString, DefaultValuesStringParsingTest,
+    ::testing::ValuesIn(std::vector<DefaultValuesStringParsingTestCase>{
+        // 0. empty map
```
antiagainst: `{nullptr, ...}` case

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -20,6 +20,120 @@ using namespace spvtools;
...
+        {"{4294967295:0}", true, SpecIdToValueStrMap{{4294967295, "0"}}},
+        // 9. minimum spec id
+        {"{0:100}", true, SpecIdToValueStrMap{{0, "100"}}},
+        // 10. random content without spaces are allowed
+        {"{200: random_stuff}", true,
+         SpecIdToValueStrMap{{200, "random_stuff"}}},
+        // 11. missing ending '}' is fine for parsing
+        {"{100:1024", true, SpecIdToValueStrMap{{100, "1024"}}},
+        // 12. multiple ending '}' is fine for parsing
+        {"{100:1024}}", true, SpecIdToValueStrMap{{100, "1024"}}},
```
antiagainst: Hmm, l'd also suggest to fail this case.

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -20,6 +20,120 @@ using namespace spvtools;
...
+        {"{4294967296:0}", false, SpecIdToValueStrMap{}},
+        // 10. spec id less than 0
+        {"{-1:0}", false, SpecIdToValueStrMap{}},
+        // 11. empty string
+        {"", false, SpecIdToValueStrMap{}},
+        // 12. empty comma is not allow
+        {
+            "{100:1024,,200:2048}", false, SpecIdToValueStrMap{},
+        },
+        // 13. empty comma is not allow
```
antiagainst: not allowed

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Nice catch! We definitely should guard against it.

source/opt/set_spec_constant_default_value_pass.h
antiagainst: The reason is that, without a `}`, we are not clear whether it's user's intent and the string is complete or the string somehow gets truncated.

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -20,6 +20,120 @@ using namespace spvtools;
...
+        {"{,}", false, SpecIdToValueStrMap{}},
+        // 14. missing comma
+        {"{100:1024 200:2048}", false, SpecIdToValueStrMap{}},
```
antiagainst: Also cases for

1. only a number: `1234`
2. without `{}`: `12: 34`
3. hybrid: `{1abc3d: www}`
4. some other symbols: `{12: 34; 56: 78}`
5. some other symbols: `{12;34}`

etc.

Qining: Hmm, I thought some of them are already covered by current test cases. Anyway they will be added.

Qining: Tests added in 5aa7b55

test/opt/test_set_spec_const_default_value.cpp
Qining: Fixed in 5aa7b55

test/opt/test_set_spec_const_default_value.cpp
Qining: Hmm, currently all the content after the first '}' is ignored. So it doesn't matter whether they are anything after '}' at all. But anyway, this can be added later.

test/opt/test_set_spec_const_default_value.cpp
Qining: Added in 5aa7b55

test/opt/test_set_spec_const_default_value.cpp
Qining: Fixed in 5aa7b55

test/opt/test_set_spec_const_default_value.cpp
Qining: changed in 5aa7b55

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Blurb updated in 5aa7b55

source/opt/set_spec_constant_default_value_pass.h
Qining: fixed in 5aa7b55

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Updated in 5aa7b55

source/opt/set_spec_constant_default_value_pass.h
Qining: Fixed in 5aa7b55

source/opt/set_spec_constant_default_value_pass.h
Qining: Fixed in 5aa7b55

source/opt/set_spec_constant_default_value_pass.h
Qining: Fixed in 5aa7b55

source/opt/set_spec_constant_default_value_pass.h
Qining: Fixed in 5aa7b55

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Should I have '\v' '\f' in the tests?

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,35 +43,42 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
-  //    hex format with '0x' prefix is allowed.
-  //    spaces withing the number text is not allowed.
-  //    empty <spec id> is not allowed.
+  //    The text must represent a valid uint32_t number.
+  //    Spaces before and afer the number text is allowed.
+  //    Hex format with '0x' prefix is allowed.
+  //    Spaces within the number text is not allowed.
+  //    Empty <spec id> is not allowed.
+  //    One spec id value can only be defined once, multiple default values
+  //      defined for a same spec id is not allowed. Spec ids with same value
```
antiagainst: We don't say a same. Same is always together with the.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -251,16 +251,15 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+  if (!ptr) return false;
+  return *ptr == '\0' || *ptr == '}';
 }
 
 // Returns true if the char pointed by the given pointer is ' ', '\n', '\t',
 // '\r'.
-bool IsSpace(const char* const ptr) {
+inline bool IsSpace(const char* const ptr) {
+  if (!ptr) return false;
   if (*ptr == ' ' || *ptr == '\n' || *ptr == '\r' || *ptr == '\t') {
```
antiagainst: This can also be written as `return *ptr == ' ' || ...`, which is less verbose, and it's also quite readable.

Same for the following function.

antiagainst: Maybe we should just call `std::isspace` here.

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,35 +43,42 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
+  // success. Returns a nullptr if the given string is not valid for building
+  // the mapping.
+  // A valid string for building the mapping should follow the rule below:
   //
   //  { <spec id A> : <default value for A>, <spec id B> : <default value for
   //  B>, ...}
   //
-  //  The string must starts with '{', and should ends with '}'. However the
-  //  missing the ending '}' may still be valid for parsing.
+  //  The string must starts with '{', and end with '}'. Missing '{' or '}'
```
antiagainst: must start

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,35 +43,42 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
+  //  results into a parse failure. The parsing will end at the first '}' found
```
antiagainst: results in

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,35 +43,42 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
-  //    spaces withing the number text is not allowed.
-  //    empty <spec id> is not allowed.
+  //    The text must represent a valid uint32_t number.
+  //    Spaces before and afer the number text is allowed.
+  //    Hex format with '0x' prefix is allowed.
+  //    Spaces within the number text is not allowed.
+  //    Empty <spec id> is not allowed.
+  //    One spec id value can only be defined once, multiple default values
+  //      defined for a same spec id is not allowed. Spec ids with same value
+  //      but different formats (e.g. 0x100 and 256) are considered the same.
```
antiagainst: Nice. Would like to see a test case for this.

Qining: Added in 1403493

source/opt/set_spec_constant_default_value_pass.h
Qining: Fixed in 1403493

source/opt/set_spec_constant_default_value_pass.h
Qining: Fixed in 1403493

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 1403493

source/opt/set_spec_constant_default_value_pass.h
Qining: Fixed in 1403493

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,6 +43,46 @@ class SetSpecConstantDefaultValuePass : public Pass {
...
   bool Process(ir::Module*) override;
 
+  // Parses the given null-terminated c string to get a mapping from Spec Id to
+  // default value strings. Returns a unique pointer of the mapping from spec
+  // ids to spec constant default value strings built from the given |str| on
+  // success. Returns a nullptr if the given string is not valid for building
+  // the mapping.
+  // A valid string for building the mapping should follow the rule below:
+  //
+  //  { <spec id A> : <default value for A>, <spec id B> : <default value for
```
dneto0: This is a more complex syntax than you needed.  
Since spaces are not allowed inside a number, you could have used space as a delimiter between key/value pairs.
Then you just split on whitespace, and then for each term split on colon.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,85 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
   return modified;
 }
 
+// A helper function for parsing spec constant default value string. Returns
+// true if the char pointed by the given pointer is considered as the end of
```
dneto0: or if the given pointer is null.

dneto0: The blurb should not say what context the function is used in, since that tends to go stale. Just say what it computes.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,85 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+// the string, which means it is a '\0' or '}'.
+inline bool IsEnd(const char* const ptr) {
+  if (!ptr) return false;
+  return *ptr == '\0' || *ptr == '}';
+}
+
+// A helper function for parsing spec constant default value string. Returns
```
dneto0: Just say what it computes, not what context it's executed in (since that's irrelevant and likely to go stale.)

source/opt/set_spec_constant_default_value_pass.cpp
dneto0: You don't even use the nullptr case.  Just remove it?  And the blurb could say the pointer must non-null.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,85 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+// true if the char pointed by the given pointer is considered as the content
+// of an entry, which means it is not a space, end, ':' or ','.
+inline bool IsEntryContent(const char* const ptr) {
+  if (!ptr) return false;
+  return std::isspace(*ptr) == 0 && !IsEnd(ptr) && *ptr != ',' && *ptr != ':';
```
dneto0: Everything beyond the isspace is a good candidate for
  if ( std::strchr("},:", *ptr) == nullptr )  {  I am entry content }
(I'm even half tempted to replace the isspace with it too, except for locale complications...  Meh.)

dneto0: But then again, if you had a simpler syntax all this would be less complex in the first place.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,85 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+      // value string is empty.
+      return nullptr;
+    }
+    // Update the mapping with spec id and value string.
+    (*spec_id_to_value)[spec_id] = std::string(val_begin, val_end - val_begin);
+
+    // skip spaces after value string but before ',' or '}'
+    while (std::isspace(*str) != 0) str++;
+
+    // Entries must be separated by ',', if the next non-space char is not
```
dneto0: This complexity is caused by the complexity of the syntax you're accepting.  Not needed.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,85 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+    // the end sign: '}', a comma must present.
+    if (*str == ',') {
+      str++;
+    } else if (IsEnd(str)) {
+      continue;
+    } else {
+      return nullptr;
+    }
+    // skip tailing spaces
```
dneto0: trailing.

dneto0:
Simplify the syntax you accept.  It doesn't help the user, and it makes your code and tests more complex.
It makes it more likely that you're missing test cases.

source/opt/set_spec_constant_default_value_pass.h
Qining: Refactored in 19328b6

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Removed in 19328b6

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Refactored in 19328b6

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Refactored in 19328b6

Qining:
The string syntax has been simplified in 19328b6. Tests also updated.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -250,36 +250,24 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
-// A helper function for parsing spec constant default value string. Returns
-// true if the char pointed by the given pointer is considered as the content
-// of an entry, which means it is not a space, end, ':' or ','.
-inline bool IsEntryContent(const char* const ptr) {
-  if (!ptr) return false;
-  return std::isspace(*ptr) == 0 && !IsEnd(ptr) && *ptr != ',' && *ptr != ':';
+// Returns true if the given char is ':', '\0' or considered as blank space
+// (i.e.: '\n', '\r', '\v', '\t', '\f' and ' ').
+bool IsSeparator(char ch) {
+  return std::strchr(":\0", ch) || std::isspace(ch) != 0;
```
dneto0: std::strchr needs #include <cstring>
std::isspace needs #include <cctype>

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -289,43 +277,28 @@ SetSpecConstantDefaultValuePass::ParseDefaultValuesString(const char* str) {
...
-    // the end sign: '}', a comma must present.
-    if (*str == ',') {
-      str++;
-    } else if (IsEnd(str)) {
-      continue;
-    } else {
-      return nullptr;
-    }
-    // skip tailing spaces
+    // Skip tailing spaces.
```
dneto0: "trailing"

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,58 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
   return modified;
 }
 
+// Returns true if the given char is ':', '\0' or considered as blank space
+// (i.e.: '\n', '\r', '\v', '\t', '\f' and ' ').
+bool IsSeparator(char ch) {
+  return std::strchr(":\0", ch) || std::isspace(ch) != 0;
```
dneto0: need #include <cctype> and #include <cstring>

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,58 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+}
+
+std::unique_ptr<SetSpecConstantDefaultValuePass::SpecIdToValueStrMap>
+SetSpecConstantDefaultValuePass::ParseDefaultValuesString(const char* str) {
+  if (!str) return nullptr;
+
+  auto spec_id_to_value = MakeUnique<SpecIdToValueStrMap>();
+
+  // The parsing loop, break when points to the end.
+  while (!std::strchr("\0", *str)) {
```
dneto0: just need `while (*str) {`

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,58 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+    while (!IsSeparator(*str)) str++;
+    const char* val_end = str;
+    if (val_end == val_begin) {
+      // Value string is empty.
+      return nullptr;
+    }
+    // Update the mapping with spec id and value string.
+    (*spec_id_to_value)[spec_id] = std::string(val_begin, val_end - val_begin);
+
+    // Skip tailing spaces.
```
dneto0: "trailing"

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +250,58 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+std::unique_ptr<SetSpecConstantDefaultValuePass::SpecIdToValueStrMap>
+SetSpecConstantDefaultValuePass::ParseDefaultValuesString(const char* str) {
+  if (!str) return nullptr;
+
+  auto spec_id_to_value = MakeUnique<SpecIdToValueStrMap>();
+
+  // The parsing loop, break when points to the end.
+  while (!std::strchr("\0", *str)) {
+    // Find the spec id.
+    while (std::isspace(*str) != 0) str++;  // skip heading spaces.
```
dneto0: "heading" -> "leading"

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -43,6 +43,42 @@ class SetSpecConstantDefaultValuePass : public Pass {
   const char* name() const override { return "set-spec-const-default-value"; }
   bool Process(ir::Module*) override;
 
+  // Parses the given null-terminated c string to get a mapping from Spec Id to
```
dneto0: "c" -> "C"

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -20,6 +20,105 @@ using namespace spvtools;
...
+    ::testing::TestWithParam<DefaultValuesStringParsingTestCase>;
+
+TEST_P(DefaultValuesStringParsingTest, TestCase) {
+  const auto& tc = GetParam();
+  auto actual_map =
+      opt::SetSpecConstantDefaultValuePass::ParseDefaultValuesString(
+          tc.default_values_str);
+  if (tc.expect_success) {
+    EXPECT_NE(nullptr, actual_map);
+    if (actual_map) EXPECT_EQ(tc.expected_map, *actual_map);
```
dneto0: Suggest using EXPECT_THAT and  an appropriate container matcher, so you get better error message on failure.

test/opt/test_set_spec_const_default_value.cpp
```diff
@@ -20,6 +20,105 @@ using namespace spvtools;
...
+         SpecIdToValueStrMap{{100, "1024"}}},
+        // 6. maximum spec id
+        {"4294967295:0", true, SpecIdToValueStrMap{{4294967295, "0"}}},
+        // 7. minimum spec id
+        {"0:100", true, SpecIdToValueStrMap{{0, "100"}}},
+        // 8. random content without spaces are allowed
+        {"200:random_stuff", true, SpecIdToValueStrMap{{200, "random_stuff"}}},
+        // 9. support hex format spec id (just because we use the
+        // ParseNumber() utility)
+        {"0x100:1024", true, SpecIdToValueStrMap{{256, "1024"}}},
```
dneto0: Add positive cases where the value is a hex float.  Or an ordinarily expressed double.  What about 1e10 1.5e-13

dneto0:
-1 review complete

test/opt/test_set_spec_const_default_value.cpp
Qining: Fixed in 65847f7

test/opt/test_set_spec_const_default_value.cpp
Qining: This is so handy! Changed in 65847f7

source/opt/set_spec_constant_default_value_pass.h
Qining: Changed in 65847f7

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 65847f7

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 65847f7

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Updated in 65847f7

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Added back in 65847f7

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Fixed in 65847f7

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Added back in 65847f7

Qining:
Comments above resolved.
Rebased in 152c1fc

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -248,5 +252,58 @@ bool SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
...
+std::unique_ptr<SetSpecConstantDefaultValuePass::SpecIdToValueStrMap>
+SetSpecConstantDefaultValuePass::ParseDefaultValuesString(const char* str) {
+  if (!str) return nullptr;
+
+  auto spec_id_to_value = MakeUnique<SpecIdToValueStrMap>();
+
+  // The parsing loop, break when points to the end.
+  while (*str) {
+    // Find the spec id.
+    while (std::isspace(*str) != 0) str++;  // skip leading spaces.
```
dneto0: The `!= 0` is redundant.

dneto0:
-1 just one more thing.  :-)

dneto0:
+2

PR #410, opened 2016-09-12
4. C++ interface for assembling, disassembling, & validating

antiagainst:
Depends on #406.

### DIFF
--- a/source/opt/libspirv.cpp
+++ b/source/opt/libspirv.cpp
@@ -16,46 +16,58 @@
 
 #include "ir_loader.h"
 #include "make_unique.h"
+#include "message.h"
 #include "table.h"
 
 namespace spvtools {
 
+// Structs for holding the data members for SpvTools.
+struct SpvTools::Impl {
+  explicit Impl(spv_target_env env) : context(spvContextCreate(env)) {
+    // The default consumer in spv_context_t is a null consumer, which provides
+    // equivalent functionality (from the user's perspective) as a real consumer
+    // does nothing.
+  }
+  ~Impl() { spvContextDestroy(context); }
+
+  spv_context context;  // C interface context object.
+};
+
+SpvTools::SpvTools(spv_target_env env) : impl_(new Impl(env)) {}
+
+SpvTools::~SpvTools() {}
+
 void SpvTools::SetMessageConsumer(MessageConsumer consumer) {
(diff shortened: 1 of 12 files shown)


source/opt/libspirv.hpp
```diff
@@ -20,49 +20,64 @@
...
+  // Disassembles the given SPIR-V |binary| with the given |options| and writes
+  // the assembly to |text|. Returns ture on successful disassembling. |text|
+  // will be kept untouched if diassembling is unsuccessful.
+  bool Disassemble(const std::vector<uint32_t>& binary, std::string* text,
+                   uint32_t options = kDefaultDisassembleOption) const;
 
-  // Builds and returns a Module from the given SPIR-V assembly |text|.
-  std::unique_ptr<ir::Module> BuildModule(const std::string& text);
+  // Validates the given SPIR-V |binary|. Returns true if no issues are found.
+  // Otherwise, returns false and commuicates issues via the message consumer
```
dneto0: "communicates"

source/opt/libspirv.cpp
```diff
@@ -16,91 +16,58 @@
...
-  if (status == SPV_SUCCESS) return module;
-  return nullptr;
+  return status == SPV_SUCCESS;
 }
 
-std::unique_ptr<ir::Module> SpvTools::BuildModule(const std::string& text) {
-  std::vector<uint32_t> binary;
-  if (Assemble(text, &binary) != SPV_SUCCESS) return nullptr;
-  return BuildModule(binary);
+bool SpvTools::Validate(const std::vector<uint32_t>& binary) const {
```
dneto0: My first impression is to be a little bit wary of losing information by only returning a bool.  For example, we've talked about succeeding but with warnings in case the validator sees extensions it doesn't understand.  But the message consumer already can express warnings that still allow you to succeed.  So Ok I think!
It just feels weird given that several Khronos APIs use 0 for success and other ints for failure.

But still ok.

antiagainst: Yes. However, we are using `bool` here which is more clear and should avoid the "which number represents success" problem. ;)

source/opt/libspirv.hpp
antiagainst: Fixed and submitted.

PR #411, opened 2016-09-12
5. C++ interface for optimization

antiagainst:
Depends on #410.

A step towards addressing #2.

### DIFF
--- a/source/opt/optimizer.cpp
+++ b/source/opt/optimizer.cpp
@@ -1,0 +1,121 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "optimizer.hpp"
+
+#include "build_module.h"
+#include "make_unique.h"
+#include "pass_manager.h"
+#include "passes.h"
+
+namespace spvtools {
+
+struct Optimizer::PassToken::Impl {
+  Impl(std::unique_ptr<opt::Pass> p) : pass(std::move(p)) {}
+
+  std::unique_ptr<opt::Pass> pass;  // Internal implementation pass.
+};
+
+Optimizer::PassToken::PassToken(
+    std::unique_ptr<Optimizer::PassToken::Impl> impl)
+    : impl_(std::move(impl)) {}
+Optimizer::PassToken::PassToken(PassToken&& that)
+    : impl_(std::move(that.impl_)) {}
+
+Optimizer::PassToken& Optimizer::PassToken::operator=(PassToken&& that) {
+  impl_ = std::move(that.impl_);
+  return *this;
+}
+
+Optimizer::PassToken::~PassToken() {}
+
+struct Optimizer::Impl {
+  explicit Impl(spv_target_env env) : target_env(env), pass_manager() {}
+
+  const spv_target_env target_env;  // Target environment.
+  opt::PassManager pass_manager;    // Internal implementation pass manager.
+};
+
+Optimizer::Optimizer(spv_target_env env) : impl_(new Impl(env)) {}
+
+Optimizer::~Optimizer() {}
+
+void Optimizer::SetMessageConsumer(MessageConsumer c) {
+  // All passes' message consumer needs to be updated.
+  for (uint32_t i = 0; i < impl_->pass_manager.NumPasses(); ++i) {
+    impl_->pass_manager.GetPass(i)->SetMessageConsumer(c);
+  }
+  impl_->pass_manager.SetMessageConsumer(std::move(c));
+}
+
+Optimizer& Optimizer::RegisterPass(PassToken&& p) {
+  // Change to use the pass manager's consumer.
+  p.impl_->pass->SetMessageConsumer(impl_->pass_manager.consumer());
+  impl_->pass_manager.AddPass(std::move(p.impl_->pass));
+  return *this;
+}
+
+bool Optimizer::Run(const uint32_t* original_binary,
+                    const size_t original_binary_size,
+                    std::vector<uint32_t>* optimized_binary) const {
+  std::unique_ptr<ir::Module> module =
+      BuildModule(impl_->target_env, impl_->pass_manager.consumer(),
+                  original_binary, original_binary_size);
+  if (module == nullptr) return false;
+
+  auto status = impl_->pass_manager.Run(module.get());
+  if (status == opt::Pass::Status::SuccessWithChange) {
+    optimized_binary->clear();
+    module->ToBinary(optimized_binary, /* skip_nop = */ true);
+  }
+
+  return status != opt::Pass::Status::Failure;
+}
+
+Optimizer::PassToken CreateNullPass() {
+  return MakeUnique<Optimizer::PassToken::Impl>(MakeUnique<opt::NullPass>());
+}
+
+Optimizer::PassToken CreateStripDebugInfoPass() {
+  return MakeUnique<Optimizer::PassToken::Impl>(
+      MakeUnique<opt::StripDebugInfoPass>());
+}
+
+Optimizer::PassToken CreateSetSpecConstantDefaultValuePass(
+    const std::unordered_map<uint32_t, std::string>& id_value_map) {
+  return MakeUnique<Optimizer::PassToken::Impl>(
+      MakeUnique<opt::SetSpecConstantDefaultValuePass>(id_value_map));
+}
+
+Optimizer::PassToken CreateFreezeSpecConstantValuePass() {
+  return MakeUnique<Optimizer::PassToken::Impl>(
+      MakeUnique<opt::FreezeSpecConstantValuePass>());
+}
+
+Optimizer::PassToken CreateFoldSpecConstantOpAndCompositePass() {
+  return MakeUnique<Optimizer::PassToken::Impl>(
+      MakeUnique<opt::FoldSpecConstantOpAndCompositePass>());
+}
+
+Optimizer::PassToken CreateUnifyConstantPass() {
+  return MakeUnique<Optimizer::PassToken::Impl>(
+      MakeUnique<opt::UnifyConstantPass>());
+}
+
+Optimizer::PassToken CreateEliminateDeadConstantPass() {
+  return MakeUnique<Optimizer::PassToken::Impl>(
+      MakeUnique<opt::EliminateDeadConstantPass>());
+}
+
+}  // namespace spvtools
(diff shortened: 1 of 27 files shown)


source/opt/optimizer.hpp
```diff
@@ -0,0 +1,151 @@
...
+  ~Optimizer();
+
+  // Sets the message consumer to the given |consumer|. The |consumer| will be
+  // invoked once for each message communicated from the library.
+  void SetMessageConsumer(MessageConsumer consumer);
+
+  // Optimizes the given SPIR-V |binary| module.
+  // Returns true on successful optimization, no matter the module is modified
+  // or not. Returns false if errors occur when processing |binary| using any
+  // of the registered passes. All passes registered after the error-reporting
```
dneto0: Make it clearer that these last clauses only apply on failure of a pass:  "In that case, no further passes are executed and the contents in |binary| may be invalid."

source/opt/optimizer.hpp
```diff
@@ -0,0 +1,151 @@
...
+  // Registers a eliminate-dead-constant pass to this optimizer.
+  // A eliminate-dead-constant pass removes dead constants, including normal
+  // contants defined by OpConstant, OpConstantComposite, OpConstantTrue, or
+  // OpConstantFalse and spec constants defined by OpSpecConstant,
+  // OpSpecConstantComposite, OpSpecConstantTrue, OpSpecConstantFalse or
+  // OpSpecConstantOp.
+  Optimizer& RegisterEliminateDeadConstantPass();
+
+  // Clears all passes previously registered.
+  Optimizer& ClearRegisteredPasses();
```
dneto0: You don't need this.  Pretty much in any case where you want this you should just make a new optimizer and set its message consumer.

source/opt/pass_manager.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace spvtools {
+namespace opt {
+
+Pass::Status PassManager::Run(ir::Module* module) {
+  auto status = static_cast<uint32_t>(Pass::Status::SuccessWithoutChange);
+  for (const auto& pass : passes_) {
+    const auto one_status = pass->Process(module);
+    if (one_status == Pass::Status::Failure) return one_status;
+    status = static_cast<uint32_t>(one_status) & status;
```
dneto0: You could have made status a Pass::Status and changed this assignment to just:
```
  if (one_status == Pass::StatusWithChange) status = one_status;
```
Then you wouldn't need any casting anywhere else.
You also don't need any special encoding for the values.  It might just read a little bit easier too.

source/opt/optimizer.hpp
```diff
@@ -0,0 +1,151 @@
...
+  // of the registered passes. All passes registered after the error-reporting
+  // pass will be skipped. And the contents in |binary| may be invalid.
+  //
+  // Use the Register*Pass() methods in the following to register passes first.
+  // Passes will be run in the exact order of registration.
+  bool Optimize(std::vector<uint32_t>* binary) const;
+
+  // Registers a null pass to this optimizer.
+  // A null pass does nothing to the SPIR-V module to be optimized.
+  Optimizer& RegisterNullPass();
```
dneto0: I don't like that every time we add a new optimization we change the ABI to this class.
I'd rather split this into two parts. This Optimizer class accumulates registered transforms/passes via a single method, e.g. 
```
Optimizer& RegisterTransform(std::unique_ptr<Optimizer::Transform>)
```
An Optimizer::Transform is just a forward-declared struct.

Then have free functions that return such objects: 
```
std::unique_ptr<Optimizer::Transform> spvtools::NullTransform();
std::unique_ptr<Optimizer::Transform> spvtools::SetSpecConstantDefaultValueTransform( blah blah blah);
```

(I'm only using "Transform" in case you want to avoid the name "Pass".  But "Pass" would work as well.)

source/opt/optimizer.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+  // Optimizes the given SPIR-V |binary| module.
+  // Returns true on successful optimization, no matter the module is modified
+  // or not. Returns false if errors occur when processing |binary| using any
+  // of the registered passes. All passes registered after the error-reporting
+  // pass will be skipped. And the contents in |binary| may be invalid.
+  //
+  // Use the Register*Pass() methods in the following to register passes first.
+  // Passes will be run in the exact order of registration.
+  bool Optimize(std::vector<uint32_t>* binary) const;
```
dneto0: Why not call this "Run"?

(I find it a bit weird to overwrite the input vector.  I always think in terms of streaming pipelines. 
 Why not separate the input from output?  Have the input specified as a pointer+size, and a pointer-to-vector output.
Bah, that's only slightly more general, and probably just harder to use correctly.
Ok to keep.  )

source/opt/optimizer.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+  // Destructs this instance.
+  ~Optimizer();
+
+  // Sets the message consumer to the given |consumer|. The |consumer| will be
+  // invoked once for each message communicated from the library.
+  void SetMessageConsumer(MessageConsumer consumer);
+
+  // Optimizes the given SPIR-V |binary| module.
+  // Returns true on successful optimization, no matter the module is modified
```
antiagainst: Done.

source/opt/optimizer.hpp
antiagainst: Done.

source/opt/optimizer.hpp
antiagainst: Done.

source/opt/optimizer.hpp
antiagainst: Done. To create them as free functions, and also hide the internals of the token, we need to have another usage of PIMPL idiom for the PassToken...

source/opt/optimizer.hpp
antiagainst: Right. Remove.

source/opt/pass_manager.cpp
antiagainst: Good catch!! Done.

antiagainst:
The first three commits are untouched. Comments addressing stuff are in the later three commits.

source/opt/optimizer.hpp
```diff
@@ -53,99 +73,98 @@ class Optimizer {
...
+  // exact order of registration.
+  Optimizer& RegisterPass(PassToken&& pass);
+
+  // Optimizes the given SPIR-V module |original_binary| and writes the
+  // optimized binary into |optimized_binary|.
+  // Returns true on successful optimization, whether or not the module is
+  // modified. Returns false if errors occur when processing |original_binary|
+  // using any of the registered passes. In that case, no further passes are
+  // excuted and the contents in |optimized_binary| may be invalid.
+  bool Run(const uint32_t* original_binary, size_t original_binary_size,
```
dneto0: You should say that it's ok for original_binary to alias the start of the optimized_binary vector.

source/opt/optimizer.hpp
```diff
@@ -32,6 +32,26 @@ namespace spvtools {
...
+
+    // Tokens can only be moved. Copying is disabled.
+    PassToken(const PassToken&) = delete;
+    PassToken(PassToken&&);
+    PassToken& operator=(const PassToken&) = delete;
+    PassToken& operator=(PassToken&&);
+
+    ~PassToken();
+
+    std::unique_ptr<Impl> impl_;  // Unique pointer to internal data.
```
dneto0: Do you want impl_ to be private?

source/opt/optimizer.hpp
```diff
@@ -53,99 +73,98 @@ class Optimizer {
...
-  // contants defined by OpConstant, OpConstantComposite, OpConstantTrue, or
-  // OpConstantFalse and spec constants defined by OpSpecConstant,
-  // OpSpecConstantComposite, OpSpecConstantTrue, OpSpecConstantFalse or
-  // OpSpecConstantOp.
-  Optimizer& RegisterEliminateDeadConstantPass();
-
-  // Clears all passes previously registered.
-  Optimizer& ClearRegisteredPasses();
+  // Registers the given |pass| to this optimizer. Passes will be run in the
+  // exact order of registration.
```
dneto0: Should say that this method will take ownership of the passtoken.

source/opt/optimizer.cpp
```diff
@@ -15,83 +15,107 @@
...
 };
 
 Optimizer::Optimizer(spv_target_env env) : impl_(new Impl(env)) {}
 
 Optimizer::~Optimizer() {}
 
 void Optimizer::SetMessageConsumer(MessageConsumer c) {
-  impl_->message_consumer = c;
+  // All passes' message consumer needs to be updated.
+  for (uint32_t i = 0; i < impl_->pass_manager.NumPasses(); ++i) {
```
dneto0: For another CL, but it seems we could simplify by adding begin() and end() to PassManager

test/cpp_interface.cpp
```diff
@@ -211,27 +211,57 @@ TEST(CppInterface, OptimizeMulitplePasses) {
...
-  o.RegisterStripDebugInfoPass()
-      .RegisterFreezeSpecConstantValuePass()
-      .ClearRegisteredPasses();
-  EXPECT_TRUE(o.Optimize(&binary));
+  EXPECT_TRUE(t.Assemble("OpSource GLSL 450", &binary));
 
-  std::string text;
-  EXPECT_TRUE(t.Disassemble(binary, &text));
-  EXPECT_EQ(original_text, text);
+  {  // Do nothing with tokens.
```
dneto0: Nice.

test/cpp_interface.cpp
```diff
@@ -211,27 +211,57 @@ TEST(CppInterface, OptimizeMulitplePasses) {
...
+    CreateFreezeSpecConstantValuePass();
+    auto token = CreateUnifyConstantPass();
+  }
+
+  {  // Reassignment
```
dneto0: Really, these should be separate tests.

test/cpp_interface.cpp
```diff
@@ -211,27 +211,57 @@ TEST(CppInterface, OptimizeMulitplePasses) {
...
+    auto token = CreateNullPass();
+    token = CreateStripDebugInfoPass();
+
+    EXPECT_TRUE(Optimizer(SPV_ENV_UNIVERSAL_1_1)
+                    .RegisterPass(std::move(token))
+                    .Run(binary.data(), binary.size(), &binary));
```
dneto0: Note you're also testing that the input vector is aliased with the output vector.  That should be called out, maybe even in its own test.

source/opt/optimizer.hpp
antiagainst: Not quite necessary. That will require another indirection via a getter method `impl()`, and that getter method will be exposed in the header again. Distraction for the users.

So I'd prefer to keep it like this.

source/opt/optimizer.hpp
antiagainst: Added. Actually the `&&` in the signature forces that.

source/opt/optimizer.hpp
antiagainst: Done.

source/opt/optimizer.cpp
antiagainst: Sure.

test/cpp_interface.cpp
antiagainst: Splitted.

test/cpp_interface.cpp
antiagainst: Created a dedicated test for it.

source/opt/optimizer.hpp
dneto0: Approved.  Style guide would indicate this should be named without a trailing underscore since it's a struct member.  Your choice.

PR #412, opened 2016-09-13
Guard against nullptr in ParseNumber()

Qining:
Returns false if the given string is as nullptr.


source/util/parse_number.h
```diff
@@ -171,6 +171,7 @@ bool ParseNumber(const char* text, T* value_pointer) {
   static_assert(sizeof(T) > 1,
                 "Don't use a single-byte type this parse method");
```
antiagainst: The error message is strange.

Maybe
Single-byte types not supported in this function?

antiagainst:
+2. Just need to fix a weird error message.

Qining:
Guards against nullptr are also added to ParseAndEncodeXXX() functions.
Tests are also added.

antiagainst:
+2

PR #413, opened 2016-09-13
Make friendly names for built-in variables.

dneto0:
Handles only OpDecorate

Does not handle:
- decorations on struct members
- decoration via OpGroupDecorate


dneto0:
Rebased

antiagainst:
LGTM

Issue #415, opened 2016-09-14
Validation exception for Phi-function in a loop


AlexeySotkin:
spirv-val tool reports an error:

> error: 48: The following forward referenced IDs have not be defined:
13[result.0]

 on the following line:
`%result_0 = OpPhi %uint %12 %entry %result_0 %if_then %inc %if_end`

But I think the following exception from the [specification (2.16.1 SSA)](https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#_universal_validation_rules) is applicable in this case

> Uses in a phi-function in a loop may consume definitions in the loop that don’t dominate the use

...and therefore validator should not report an error. Please fix it.
The problem can be reproduced with attached [files](https://github.com/KhronosGroup/SPIRV-Tools/files/471981/tmp.zip)

dneto0:
Confirmed.  In the disassembly below, %result_0 is defined in block %while_cond, which dominates block %if_end.

![tmp](https://cloud.githubusercontent.com/assets/13054574/18514616/47548cba-7a60-11e6-86e7-876a07a32528.png)


```
      %entry = OpLabel
               OpBranch %while_cond

 %while_cond = OpLabel
   %result_0 = OpPhi %uint %12 %entry %result_0 %if_then %inc %if_end
 %num_addr_0 = OpPhi %uint %num %entry %dec %if_then %dec %if_end
     %tobool = OpINotEqual %bool %num_addr_0 %18
               OpBranchConditional %tobool %while_body %while_end

 %while_body = OpLabel
        %dec = OpIAdd %uint %num_addr_0 %21
    %tobool1 = OpINotEqual %bool %dec %18
               OpBranchConditional %tobool1 %if_then %if_end

    %if_then = OpLabel
               OpBranch %while_cond

     %if_end = OpLabel
        %inc = OpIAdd %uint %result_0 %12
               OpBranch %while_cond

  %while_end = OpLabel
               OpReturnValue %result_0
               OpFunctionEnd

```

dneto0:
I reduced it further to this case.  The reported error is: "error: 13: The following forward referenced IDs have not be defined:  7"

```
               OpCapability Addresses
               OpCapability Kernel
               OpMemoryModel Physical32 OpenCL
       %uint = OpTypeInt 32 0
          %2 = OpConstantNull %uint
          %3 = OpTypeFunction %uint
          %4 = OpFunction %uint None %3
          %5 = OpLabel
[...]

PR #416, opened 2016-09-14
Fix validator SSA check: Phi can use its own value sometimes

dneto0:
Defer removal of a Phi's result id from the undefined-forward-reference
set until after you've scanned the arguments.  The reordering is only
significant for Phi.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/415

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -2352,6 +2352,7 @@ function<bool(unsigned)> getCanBeForwardDeclaredFunction(SpvOp opcode) {
       break;
 
     case SpvOpFunctionCall:
+      // The Function parameter.
       out = [](unsigned index) { return index == 2; };
       break;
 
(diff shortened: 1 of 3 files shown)


dneto0:
@umar456  PTAL

source/validate_id.cpp
```diff
@@ -2479,31 +2483,45 @@ spv_result_t IdPass(ValidationState_t& _,
...
       case SPV_OPERAND_TYPE_RESULT_ID:
-        // NOTE: Multiple Id definitions are being checked by the binary parser
-        // NOTE: result Id is added *after* all of the other Ids have been
-        // checked to avoid premature use in the same instruction
-        _.RemoveIfForwardDeclared(*operand_ptr);
+        // NOTE: Multiple Id definitions are being checked by the binary parser.
+        //
+        // Defer undefined-forward-reference removal until after we've analyzed
+        // the remaining operands to this instruction.  Deferral only matters
+        // for
```
umar456: Nit: Superfluous line break

umar456:
+2

source/validate_id.cpp
dneto0: I put that extra linebreak in to emphasize that it's a new thing I'm talking about, not extending the Note.

dneto0:
Rebased

PR #418, opened 2016-09-14
Clarify: Ok to pass null diagnostic to spvBinaryParse

dneto0:
It's a valid use case: I might mostly trust my binaries and I don't
care to to collect detailed diagnostic feedback.


antiagainst:
+2

dneto0:
Rebased

PR #419, opened 2016-09-15
Update validate interface to accept binary pointers

umar456:
The current version of `spvValidate` accepts a `spv_const_binary` struct as one of its arguments. This works well when the module is created from the `spvTextToBinary` function but it is cumbersome when working vector of words.

This PR will create a new function `spvValidateBinary` which accepts a pointer to words instead of the `spv_const_binary` object.

### DIFF
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -177,7 +177,15 @@ UNUSED(void PrintDotGraph(ValidationState_t& _, libspirv::Function func)) {
 spv_result_t spvValidate(const spv_const_context context,
                          const spv_const_binary binary,
                          spv_diagnostic* pDiagnostic) {
+  return spvValidateBinary(context, binary->code, binary->wordCount,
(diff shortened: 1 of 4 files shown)


source/validate.cpp
```diff
@@ -177,7 +177,15 @@ UNUSED(void PrintDotGraph(ValidationState_t& _, libspirv::Function func)) {
 spv_result_t spvValidate(const spv_const_context context,
                          const spv_const_binary binary,
                          spv_diagnostic* pDiagnostic) {
+  return spvValidateBinary(context, binary->code, binary->wordCount,
+                           pDiagnostic);
+}
```
dneto0: May as well follow convention and insert a 1 line gap here.

umar456: fixed in 7dea7b8

dneto0:
Rebased and pushed into master as bf467da9c9

PR #420, opened 2016-09-16
Add basic support for SPV_KHR_shader_ballot

dneto0:
Requires use of SPIRV-Headers that has support
for SPV_KHR_shader_ballot.

Adds assembler, disassembler, binary parser support.

Adds general support for allowing an operand to be
only enabled by a set of extensions.

TODO: Validator support for extension checking.


dneto0:
Rebased and pushed into master as 9382035a22a

PR #421, opened 2016-09-16
Set the default source name as "input".

antiagainst:


dneto0:
Seems ok.  But it looks like a filename.  Did you consider something that looks less like a filename, such as "(input)" ?
+2

dneto0:
Rebased and pushed into master as 3590279a7

PR #422, opened 2016-09-16
Do normal errors instead of asserts in IrLoader

antiagainst:
* Add methods for getting positions of instructions and operands.
* Add SPIRV_ERROR() macro for returning user errors.

### DIFF
--- a/source/opt/log.h
+++ b/source/opt/log.h
@@ -78,20 +78,20 @@
 namespace spvtools {
 
 // Calls the given |consumer| by supplying  the |message|. The |message| is from
-// the given |file| and |location| and of the given severity |level|.
+// the given |source| and |location| and of the given severity |level|.
 inline void Log(const MessageConsumer& consumer, spv_message_level_t level,
-                const char* file, const spv_position_t& position,
+                const char* source, const spv_position_t& position,
                 const char* message) {
-  if (consumer != nullptr) consumer(level, file, position, message);
+  if (consumer != nullptr) consumer(level, source, position, message);
 }
 
 // Calls the given |consumer| by supplying the message composed according to the
-// given |format|. The |message| is from the given |file| and |location| and of
-// the given severity |level|.
+// given |format|. The |message| is from the given |source| and |location| and
+// of the given severity |level|.
 template <typename... Args>
 void Logf(const MessageConsumer& consumer, spv_message_level_t level,
-          const char* file, const spv_position_t& position, const char* format,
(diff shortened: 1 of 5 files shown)


source/opt/instruction.h
```diff
@@ -93,6 +93,17 @@ class Instruction {
   Instruction(Instruction&&);
   Instruction& operator=(Instruction&&);
 
+  // Returns the position for this instruction or one of its operand.
+  // The position is encoded as a spv_postion_t struct, with its
+  // * line field to mean the position of this instruction in the module
+  //   (starting from 1, 0 means invalid position),
+  // * column field to mean the the position of operand in this instruction
```
dneto0: Another weird thing.  Why not use the OpLine instructions to indicate position.  Whose position is it really?  Some high level language source file position, or inside-the-binary-module position.  It's not really clear except by comments.  This data could be easily misused.  (Also, position inside the binary changes with code transforms, discussed elsewhere in this review.)

source/opt/log.h
```diff
@@ -39,6 +39,21 @@
 #define SPIRV_ASSERT(consumer, ...)
 #endif
 
+// Logs an error message to consumer. Accepts the following formats:
```
dneto0: This is a lot of machinery to just delegate to spvtools::Log or spvtools::Logf but with one parameter (the message level substituted).  Why not have spvtools::LogError and spvtools::LogErrorF that just do that?  A using declaration to avoid typing the namespace would keep it short.

dneto0: (I know I haven't complained about the SPIRV_DEBUG macros, but the same criticism could apply there too.)

antiagainst:
@dneto0: fair point. I've stopped adding fields into Instruction itself for IrLoader's usage. This may end up invalidating the majority of the above comments. Sorry about it.

Still, the `source` is made up. I'm not sure we should have it passed in by the user; the functionality can be used via a library, and adding an entry point for source name seems overkill.

dneto0:
> Still, the source is made up. I'm not sure we should have it passed in by the user; the functionality can be used via a library, and adding an entry point for source name seems overkill.

Two options:
- make the source name an operand of the IrLoader constructor.  You could even have a second constructor that  defaults the name to "source" or "input" or whatever
- add a source-setter method. ( And default to "Source" or "input").

I somewhat prefer the constructor parameter.

antiagainst:
@dneto0: I've added a method for setting the input source in the last commit.

PR #425, opened 2016-09-21
Add some more overloads into the C++ interface

antiagainst:
The user may not be able to control the input; it's performance defect to force making a `vector` from C array in cases.

Depends on #424.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ validator, and optimizer for SPIR-V. Except for the optimizer, all are based
 on a common static library.  The library contains all of the implementation
 details, and is used in the standalone tools whilst also enabling integration
 into other code bases directly. The optimizer implementation resides in its
-own library.
+own library, which depends on the core library.
 
 The interfaces are still under development, and are expected to change.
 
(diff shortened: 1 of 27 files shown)


README.md
```diff
@@ -186,6 +195,12 @@ There are three main entry points into the library.
   for the header and each parsed instruction.  The disassembler is implemented
   as a client of `spvBinaryParse`.
 * `spvValidate` implements the validator functionality. *Incomplete*
+* `spvValidateBinary` implements the validator functionality. *Incomplete*
+
+The C++ interface is comprised of two classes, `SpirvTools` and `Optimizer`,
+both in the `spvtools` namespace.
+* `SpirvTools` provides `Assemble`, `Disassemble`, and `Validate` methods.
+* `Optimizer` provides the `Optimize` method.
```
dneto0: This is now "Run"

include/spirv-tools/libspirv.hpp
```diff
@@ -82,17 +82,26 @@ class SpirvTools {
   // Returns true on successful assembling. |binary| will be kept untouched if
   // assembling is unsuccessful.
   bool Assemble(const std::string& text, std::vector<uint32_t>* binary) const;
+  // |text_size| specifies the number of bytes in |text|.
+  bool Assemble(const char* text, size_t text_size,
```
dneto0: I think we don't need the terminating Null. That's unusual and should be stated here.

antiagainst: Added comments.

Actually this is to be consistent with `spvTextToBinary`, which takes a size parameter. Should we do normal null-terminating instead? It's kinda safe to require a size, but less conventional.

dneto0:
+2

PR #426, opened 2016-09-21
Change to use enum instead of enum class for MessageLevel.

antiagainst:
This allows us to create C interface with `MessageLevel` later.


dneto0:
Looks ok, even though it seems weird to qualify the enumerant name with the enum type name.  ?!  A corner I'm unsure about.
You get an LGTM if both bots pass.

PR #428, opened 2016-09-21
Clean up header dependency.

antiagainst:


source/opt/fold_spec_constant_op_and_composite_pass.cpp
```diff
@@ -249,6 +247,20 @@ FoldSpecConstantOpAndCompositePass::FoldSpecConstantOpAndCompositePass()
...
+Pass::Status FoldSpecConstantOpAndCompositePass::Process(ir::Module* module) {
+  Initialize(module);
+  return ProcessImpl(module);
+}
+
+void FoldSpecConstantOpAndCompositePass::Initialize(ir::Module* module) {
+  type_mgr_.reset(new analysis::TypeManager(consumer(), *module));
+  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+  for (const auto& id_def : def_use_mgr_->id_to_defs()) {
+    max_id_ = std::max(max_id_, id_def.first);
```
dneto0: std::Max needs algorithm.  Add it here, remove it from the .h file.

source/opt/fold_spec_constant_op_and_composite_pass.h
```diff
@@ -35,22 +35,13 @@ class FoldSpecConstantOpAndCompositePass : public Pass {
   FoldSpecConstantOpAndCompositePass();
 
   const char* name() const override { return "fold-spec-const-op-composite"; }
-  Status Process(ir::Module* module) override {
-    Initialize(module);
-    return ProcessImpl(module);
-  };
+
+  Status Process(ir::Module* module) override;
```
dneto0: Suspect you don't need algorithm or memory anymore.  Add it to the corresponding .cpp file?

source/opt/fold_spec_constant_op_and_composite_pass.cpp
antiagainst: Good catch!

source/opt/fold_spec_constant_op_and_composite_pass.h
antiagainst: Right. I've moved some impl. code into `.cpp` so `<algorithm>` definitely need to be moved. For `<memory>`, we have `std::unique_ptr` in the header.

Issue #429, opened 2016-09-22
spirv-val doesn't work properly with OpTypeForwardPointer


echuraev:
spirv-val get me the following error: error: 14: ID 6 has not been defined. But I think that spirv file is correct. I think that the problem is in spirv-val tool and how it works with OpTypeForwardPointer.
Files with spirv text representation, spirv binary and llvm ir are attached. I added extension "*.txt" for all of them because githug does not allow attach another types of files.
Step to reproduce: 
1. Run: spirv-val layout.spv.txt
2. Get error: error: 14: ID 6 has not been defined

llvm-ir: [layout.ll.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/486960/layout.ll.txt)
spirv text representation: [layout.spt.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/486961/layout.spt.txt)
spirv binary: [layout.spv.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/486962/layout.spv.txt)

umar456:
Confirmed. Fix should be out shortly.

umar456:
Reduced the failing kernel to:

```LLVM
               OpCapability Kernel
               OpCapability Addresses
               OpMemoryModel Logical OpenCL
               OpName %intptrt "intptrt"
               OpTypeForwardPointer %intptrt UniformConstant
       %uint = OpTypeInt 32 0
    %intptrt = OpTypePointer UniformConstant %uint
```

Also exposes a bug for `OpTypeStruct`

```LLVM
               OpCapability Kernel
               OpCapability Addresses
               OpMemoryModel Logical OpenCL
               OpName %structptr "structptr"
               OpTypeForwardPointer %structptr UniformConstant
       %uint = OpTypeInt 32 0
    %structt = OpTypeStruct %uint %structptr
 %structptr  = OpTypePointer UniformConstant %structt
```

PR #432, opened 2016-09-22
Fix forward reference issues with forward pointer

umar456:
Fix for #429 

* Allows OpTypeForwardPointer to reference IDs not yet declared in
  the module
* Allows OpTypeStruct to reference IDs not yet declared in the
  module

Possible Issue: OpTypeStruct should only allow forward references
if the ID is a pointer that is referenced by a forward pointer. Need
Type support in Validator which is currently a work in progress.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,9 @@ Revision history for SPIRV-Tools
 v2016.6-dev 2016-09-16
  - Published the C++ interface for assembling, disassembling, validation, and
    optimization.
+ - Fixes issues:
+   #429: Validator: Allow OpTypeForwardPointer and OpTypeStruct to reference
+     undefined IDs
 
 v2016.5 2016-09-16
  - Support SPV_KHR_shader_ballot in assembler, disassembler, parser.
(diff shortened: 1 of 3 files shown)


dneto0:
This is a correct relaxation of the rule.


>Possible Issue: OpTypeStruct should only allow forward references
if the ID is a pointer that is referenced by a forward pointer. 

Agreed.

We can tighten up the check here just a bit: OpTypeStruct word 0 can't be a forward reference since that's the result ID.  I'll do that in a followup CL.

+2

dneto0:
Rebased and pushed into master as a6ef1ea29

umar456:
The result_id is not passed into that function but I guess it would be beneficial to handle all cases if this function was used elsewhere.

dneto0:
Never mind on this:
>We can tighten up the check here just a bit: OpTypeStruct word 0 can't be a forward reference since that's the result ID. I'll do that in a followup CL.

Doesn't apply since the calling code only checks for Id parameters that are *not* result ids.  So there's no way to make a test case that triggers the difference.

dneto0:
... and now I see your comment from 4 minutes ago.  :-)

PR #433, opened 2016-09-22
Renaming test files to comply with Google style guide.

antiagainst:
Also prefixing test files with `test_' to make it more clear.

Autumn house cleaning series no.1.


antiagainst:
@dneto0: And I find that the style guide is saying `.cc` should be the suffix for impl. Never fail to be surprised by the style guide. Maybe I'm just following the guide N months ago.

antiagainst:
@dneto0: I've revised the PR to change to use the `*_test.cpp` convention. And all the test files for validation is kept unchanged. So it won't cause problems for 3rd party anymore.

dneto0:
I'll take ownership of rebasing 3rd party merge requests such as https://github.com/KhronosGroup/SPIRV-Tools/pull/448

PR #434, opened 2016-09-22
Fix operand class test coverage table missing column.

antiagainst:
This will make the table beautiful and searchable on Github.

Autumn house cleaning no.2.


antiagainst:
@dneto0: Maybe this table is obsolete and should be removed?

dneto0:
The lines you changed now have carriage returns in them.  Otherwise it looks ok.

PR #435, opened 2016-09-22
Command line option for set default value for spec constant

Qining:
Format:
```
--set-spec-constant-default-value <spec id A>:<default value A> <spec id
B>:<default value B>...
```

Example:
  shader: `test.vert`
```

layout(constant_id = 100) const int myint = 10;
layout(constant_id = 101) const int myuint = 100;
layout(constant_id = 200) const float myfloat = 1.25;
layout(constant_id = 201) const double mydouble = 2.34;

void main() {}
```
  command line:
```
spirv-opt --set-spec-const-default-value 100:12   101:200 200:1.2323
201:1.2345 test.vert -o output.spv
```
  output:
```
      ...
               OpDecorate %7 SpecId 100
               OpDecorate %8 SpecId 101
               OpDecorate %10 SpecId 200
               OpDecorate %12 SpecId 201
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
        %int = OpTypeInt 32 1
          %7 = OpSpecConstant %int 12
          %8 = OpSpecConstant %int 200
      %float = OpTypeFloat 32
         %10 = OpSpecConstant %float 1.23232
     %double = OpTypeFloat 64
         %12 = OpSpecConstant %double 2.34232
      ...
```

### DIFF
--- a/tools/opt/opt.cpp
+++ b/tools/opt/opt.cpp
@@ -14,6 +14,8 @@
 
 #include <cstring>
 #include <iostream>
+#include <memory>
+#include <sstream>
 #include <vector>
 
 #include "message.h"
@@ -50,6 +52,13 @@ Options:
                Fold the spec constants defined by OpSpecConstantOp or
                OpSpecConstantComposite instructions to front-end constants
                when possible.
+  --set-spec-const-default-value "<spec id>:<default value> ..."
+               Set the default values of the specialization constants with
+               <spec id>-<default value> pairs specified in a double-quoted
+               string. <spec id>-<default value> pairs must be separated by
+               blank spaces, and in each pair, spec id and default value must
+               be separated with colon ':' without any blank spaces in between.
+               e.g.: --set-spec-const-default-value "1:100 2:400"
   --unify-const
                Remove the duplicated constants.
   -h, --help   Print this help.
@@ -89,6 +98,18 @@ int main(int argc, char** argv) {
         }
       } else if (0 == strcmp(cur_arg, "--strip-debug")) {
         pass_manager.AddPass<opt::StripDebugInfoPass>();
+      } else if (0 == strcmp(cur_arg, "--set-spec-const-default-value")) {
+        if (++argi < argc) {
+          auto spec_ids_vals =
+              opt::SetSpecConstantDefaultValuePass::ParseDefaultValuesString(
+                  argv[argi]);
+          pass_manager.AddPass<opt::SetSpecConstantDefaultValuePass>(
+              std::move(*spec_ids_vals));
+        } else {
+          fprintf(stderr,
+                  "error: Expect a string of <spec id>-<default value> pairs.");
+          return 1;
+        }
       } else if (0 == strcmp(cur_arg, "--freeze-spec-const")) {
         pass_manager.AddPass<opt::FreezeSpecConstantValuePass>();
       } else if (0 == strcmp(cur_arg, "--eliminate-dead-const")) {


tools/opt/opt.cpp
```diff
@@ -50,6 +52,13 @@ NOTE: The optimizer is a work in progress.
                Fold the spec constants defined by OpSpecConstantOp or
                OpSpecConstantComposite instructions to front-end constants
                when possible.
+  --set-spec-const-default-value <spec id>:<default value> ...
+               Set the default values of the specialization constants with
```
antiagainst: Should change plural format to singular one.

tools/opt/opt.cpp
```diff
@@ -50,6 +52,13 @@ NOTE: The optimizer is a work in progress.
...
+               <spec id> to <default value>. Spec id:default value pairs should
```
antiagainst: `<spec-id>-<default-value> pairs ...`

tools/opt/opt.cpp
```diff
@@ -50,6 +52,13 @@ NOTE: The optimizer is a work in progress.
...
+               be separated by blank spaces, and in each pair, spec id and
+               default value should be separated with colon ':' without any
+               blank space in between.
+               e.g.: --set-spec-const-default-value 1:100 2:400 <other args>
```
antiagainst: No need to have `<other args>` since this is already an example.

tools/opt/opt.cpp
```diff
@@ -89,6 +98,21 @@ int main(int argc, char** argv) {
         }
       } else if (0 == strcmp(cur_arg, "--strip-debug")) {
         pass_manager.AddPass<opt::StripDebugInfoPass>();
+      } else if (0 == strcmp(cur_arg, "--set-spec-const-default-value")) {
+        std::ostringstream spec_ids_vals_stream;
+        // All the arguments with ':' inside will be considered as
```
antiagainst: I personally don't think this is a good way to distinguish arguments. We may have additional options in the future that takes `:` somehow, like `--some-future-option=key:value`, which will be considered as `<spec-id>:<default-value>` pair here.

What about the format like `--set-spec-const-default-value="k1:v1 k2:v2"`? The quotation should act as delimiters for us. That's also the typical way of passing values containing spaces. It's also much easier to extract the whole string here.

Qining: I think using " " is a good idea, will add it soon.

tools/opt/opt.cpp
Qining: updated in c72ebb3

tools/opt/opt.cpp
Qining: updated in 3806bc2

tools/opt/opt.cpp
Qining: removed in 3806bc2

tools/opt/opt.cpp
Qining: Fixed in 3806bc2

tools/opt/opt.cpp
antiagainst: It seems this is still not fixed?

tools/opt/opt.cpp
```diff
@@ -50,6 +52,13 @@ NOTE: The optimizer is a work in progress.
                Fold the spec constants defined by OpSpecConstantOp or
                OpSpecConstantComposite instructions to front-end constants
                when possible.
+  --set-spec-const-default-value "<spec id>:<default value> ..."
+               Set the default values of the specialization constants with
+               <spec id>-<default value> pairs specified in a double-quoted
+               string. Spec id:default value pairs must be separated by blank
```
antiagainst:
```
 <spec id>-<default value> pairs must be separated by blank
spaces, and in each pair, <spec id> and <default value> must be
separated with colon ':' without any blank space in between.
```

tools/opt/opt.cpp
```diff
@@ -89,6 +98,12 @@ int main(int argc, char** argv) {
         }
       } else if (0 == strcmp(cur_arg, "--strip-debug")) {
         pass_manager.AddPass<opt::StripDebugInfoPass>();
+      } else if (0 == strcmp(cur_arg, "--set-spec-const-default-value")) {
+        auto spec_ids_vals =
+            opt::SetSpecConstantDefaultValuePass::ParseDefaultValuesString(
+                argv[++argi]);
```
antiagainst: Should guard against `++argi >= argc`.

Qining: Fixed in f01e3fb

tools/opt/opt.cpp
Qining: Fixed in f01e3fb

tools/opt/opt.cpp
Qining: Oops, err, fixed now in f01e3fb

Qining:
At least a string of <spec id>-<default value> pairs in the command line arguments is required now in 01df41d

PR #436, opened 2016-09-22
Handle parse failure for setting spec constants

dneto0:


tools/opt/opt.cpp
```diff
@@ -103,11 +103,18 @@ int main(int argc, char** argv) {
...
+                    "--set-spec-const-default-value: %s\n",
+                    argv[argi]);
+            return 1;
+          }
           pass_manager.AddPass<opt::SetSpecConstantDefaultValuePass>(
               std::move(*spec_ids_vals));
         } else {
           fprintf(stderr,
-                  "error: Expect a string of <spec id>-<default value> pairs.");
+                  "error: Expected a string of <spec id>:<default value> pairs.");
```
antiagainst: It seems we'd better also to change the representation in the help message to `<spec id>:<default value>`.

tools/opt/opt.cpp
```diff
@@ -103,11 +103,18 @@ int main(int argc, char** argv) {
           auto spec_ids_vals =
               opt::SetSpecConstantDefaultValuePass::ParseDefaultValuesString(
                   argv[argi]);
+          if (!spec_ids_vals) {
```
antiagainst: Interesting! Forgot the case that `ParseDefaultValuesString` can also return `nullptr`. Thanks for fixing it!

tools/opt/opt.cpp
dneto0: Fixed in next commit

dneto0:
Keeping the approval.

PR #439, opened 2016-09-28
Add EliminateDeadCodePass.

antiagainst:
It does
- Remove dead extended instruction set imports
- Remove dead types
- Remove dead constants
- Remove dead global variables with internal visibility
- Remove dead instructions in basic blocks

It does not
- Remove debug instructions
- Change basic block or function structures

### DIFF
--- a/source/opt/eliminate_dead_code_pass.cpp
+++ b/source/opt/eliminate_dead_code_pass.cpp
@@ -1,0 +1,129 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


source/opt/eliminate_dead_code_pass.cpp
```diff
@@ -0,0 +1,129 @@
...
+namespace {
+
+// Returns true if the given |inst| is a dead instruction:
+// * generating result id, and
+// * no uses or uses are all debug instructions, and
+// * if is variable-defining instruction, having internal visibility.
+bool IsDeadInst(const analysis::DefUseManager& def_use_mgr,
+                const ir::Instruction& inst) {
+  // Instructions not generating result ids are used for their side effects.
+  if (inst.result_id() == 0) return false;
```
chrisforbes: I don't think this is quite correct. Consider the various atomics which generate a result id, but have side effects which must be performed even if the result is ignored.

antiagainst: Yep, you are right.

This PR is obsolete and I believe @greg-lunarg will have a better implementation. :) So I'll close it.

PR #442, opened 2016-10-04
Fix compilation errors when building inside Chromium

Kangz:
Example of an error:
    spirv-tools/source/validate_cfg.cpp:516:45: error: chosen constructor is
    explicit in copy-initialization:
      _.current_function().RegisterBlockEnd({}, opcode);


antiagainst:
+2

PR #445, opened 2016-10-14
Do not gitignore source/opt/build_module.*

ChrisGautier:
Modifications to e.g. ``source/opt/build_module.cpp`` are ignored due to the content of .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-build*
+/build*
 .ycm_extra_conf.py*
 compile_commands.json
 /external/googletest/


dneto0:
@ChrisGautier: I'm not experiencing that problem.  That is, if I change source/opt/build_module.cpp then git sees the change.  I'm using git 2.8.0.  Did the gitignore rules change?

dneto0:
In particular, the gitignore manual page (for 2.8.0 and also the latest) say:

> If the pattern does not contain a slash /, Git treats it as a shell glob pattern and checks for a match against the pathname relative to the location of the .gitignore file (relative to the toplevel of the work tree if not from a .gitignore file).

That's the clause that applies in this case.  So source/opt/build_module.cpp does not match the rule since the "build.*" pattern is relative to the location of the .gitignore file itself.

ChrisGautier:
I'm using git 2.7.4 (but it does not actually look relevant). Its man page has the same sentence.

Interestingly, if you change ``build_module.h``, you will indeed see the differences with the unmodified ``.gitignore``. However, if you create a new file whose name starts with ``build``, ``git status`` will not list it. If you try ``git add``ing it, git complains.
```
% touch source/opt/build_module.h2
% # git status does not show the file
% git add source/opt/build_module.h2
The following paths are ignored by one of your .gitignore files:
source/opt/build_module.h2
Use -f if you really want to add them.
```

I got the problem when I was trying to import SPIR-V tools into my own private repository.

So it looks like a git bug where the semantics of .gitignore changes depending on whether you are adding a file, or changing it. Removing a file works as expected.

dneto0:
Thanks for clarifying.  I get the same behaviour with git 2.8.0.   I'll merge.

ChrisGautier:
Thanks!

PR #450, opened 2016-10-24
Search for "SPIRV-Headers", as to work on case-sensitive file systems.

mattst88:
The name of the repository with the SPIRV headers is "SPIRV-Headers" and so the name of the directory will be as such. On a case-sensitive filesystem, the existing CMake code would fail to find the repository.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -26,7 +26,7 @@ matrix:
 
 # scripts that run after cloning repository
 install:
-  - git clone https://github.com/KhronosGroup/SPIRV-Headers.git external/spirv-headers
+  - git clone https://github.com/KhronosGroup/SPIRV-Headers.git external/SPIRV-Headers
   - git clone https://github.com/google/googletest.git external/googletest
 
 build:
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 .ycm_extra_conf.py*
 compile_commands.json
 /external/googletest/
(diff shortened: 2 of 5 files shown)


mattst88:
I should be covered by Intel. Should I resubmit the patch with my @intel.com address (which is matt dot turner at)? Would that be sufficient?

dneto0:
@mattst88 I'm not sure if that will cover it.  It's worth a try.  I guess that would work on the theory that the CLA-checker will check organization membership as registered with GitHub itself?

Regarding the content of the PR, I work on a case-sensitive filesystem.  But in my setup I check out the SPIRV-Headers repo with an explicit named target directory.  That's why the current configuration works.   But I'd have to check if I have any configurations that rely on that branch of the if.
Hey @antiagainst can you take a look to see if it's ok to make this change?

antiagainst:
I'm afraid this change will break all existing checkouts, since SPIRV-Headers has been the dependency for SPIRV-Tools for a while. After clients update their SPIRV-Tools with this change, no SPIRV-Headers will be found unless they notice the repo directory checkout change.

I agree that the case for the repo checkout is kinda inconsistent with the project name; but from SPIRV-Tool's perspective, SPIRV-Headers is just an external project and should be freely checked out into any directory, say, even `the-header-repo` or something like that.

mattst88:
Okay, thanks. Closing.

PR #452, opened 2016-10-27
enum_set: Fix bool performance warning.

null77:
Implicit casts from int to bool cause a warning in visual studio.

PTAL


antiagainst:
LGTM

Issue #451, opened 2016-10-27
source/instruction.cpp and source/val/Instruction.cpp conflict in external build config


null77:
On Windows, instruction and Instruction evaluate to the same filename, so the visual studio linker in my custom build outputs the warning:

1>D:\angle\build\Debug_x64\obj\spirv_tools\Instruction.obj : warning LNK4042: object specified more than once; extras ignored

This forces the latest compiled "instruction" file to be linked, which can produce link errors.

My solution is to rename one or both of the files, but I wasn't sure on the preferred naming scheme. Can someone fix this, or tell me what a good preferred name is? Would be much appreciated, I know it doesn't affect all configs, but would make my life easier.

dneto0:
Ouch.  There's a similar warning on my OSX build, but I've never seen a functional problem.  We should fix this.

I just took a look.  source/instruction.cpp is fairly trivial.  We can get rid of it by moving its single-line function body into the header with explicit inlining.

Thanks for raising this issue.

null77:
Thanks a lot for fixing this David!

PR #453, opened 2016-10-28
Remove source/instruction.cpp

dneto0:
Avoids linker problems on Windows and OSX with duplicate object file
names.  Avoids collision with source/val/instruction.cpp

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/451


dneto0:
FYI https://github.com/google/shaderc would require an update to react to this change.

antiagainst:
It seems `source/instruction.cpp` is not actually removed from git?

dneto0:
Right.  Fixed. :-)

dneto0:
Rebased

dneto0:
squashed

PR #458, opened 2016-11-10
Adding validation code for more data rules.

ehsannas:
These rules are under "Data Rules" in 2.16.1 (Universal Validation
Rules) part of the SPIR-V 1.1 Specification document:

* Scalar floating-point types can be parameterized only as 32 bit, plus
any additional sizes enabled by capabilities.

* Scalar integer types can be parameterized only as 32 bit, plus any
additional sizes enabled by capabilities.

* Vector types can only be parameterized with numerical types or the
OpTypeBool type.

* Matrix types can only be parameterized with floating-point types.

* Matrix types can only be parameterized as having only 2, 3, or 4
columns.

* Specialization constants (see Specialization) are limited to integers,
Booleans, floating-point numbers, and vectors of these.

Note that a few existing tests had to be updated to meet the data validation rules.

### DIFF
--- a/source/validate_datarules.cpp
+++ b/source/validate_datarules.cpp
@@ -59,6 +59,7 @@ spv_result_t ValidateVecNumComponents(ValidationState_t& _,
 // Validates that the number of bits specifed for a float type is valid.
 // Scalar floating-point types can be parameterized only with 32-bits.
 // Float16 capability allows using a 16-bit OpTypeFloat.
+// Float16Buffer capability allows creation of a 16-bit OpTypeFloat.
 // Float64 capability allows using a 64-bit OpTypeFloat.
 spv_result_t ValidateFloatSize(ValidationState_t& _,
                                const spv_parsed_instruction_t* inst) {
@@ -68,12 +69,13 @@ spv_result_t ValidateFloatSize(ValidationState_t& _,
     return SPV_SUCCESS;
   }
   if (num_bits == 16) {
-    if (_.HasCapability(SpvCapabilityFloat16)) {
+    if (_.HasCapability(SpvCapabilityFloat16) ||
+        _.HasCapability(SpvCapabilityFloat16Buffer)) {
       return SPV_SUCCESS;
     }
     return _.diag(SPV_ERROR_INVALID_DATA)
            << "Using a 16-bit floating point "
-           << "type requires the Float16 capability.";
+           << "type requires the Float16 or Float16Buffer capability.";
   }
   if (num_bits == 64) {
     if (_.HasCapability(SpvCapabilityFloat64)) {
@@ -124,8 +126,8 @@ spv_result_t ValidateIntSize(ValidationState_t& _,
 }
 
 // Validates that the matrix is parameterized with floating-point types.
-spv_result_t ValidateMatrixDataType(ValidationState_t& _,
-                                    const spv_parsed_instruction_t* inst) {
+spv_result_t ValidateMatrixColumnType(ValidationState_t& _,
+                                      const spv_parsed_instruction_t* inst) {
   // Find the component type of matrix columns (must be vector).
   // Operand 1 is the <id> of the type specified for matrix columns.
   auto type_id = inst->words[inst->operands[1].offset];
@@ -179,9 +181,8 @@ spv_result_t ValidateSpecConstNumerical(ValidationState_t& _,
 // Validates that OpSpecConstantTrue and OpSpecConstantFalse specialize to bool.
 spv_result_t ValidateSpecConstBoolean(ValidationState_t& _,
                                       const spv_parsed_instruction_t* inst) {
-  // Operand 0 is the <id> of the type that we're specializing to.
-  auto type_id = inst->words[inst->operands[0].offset];
-  auto type_instruction = _.FindDef(type_id);
+  // Find out the type that we're specializing to.
+  auto type_instruction = _.FindDef(inst->type_id);
   if (type_instruction->opcode() != SpvOpTypeBool) {
     return _.diag(SPV_ERROR_INVALID_ID) << "Specialization constant must be "
                                            "a boolean type.";
@@ -211,10 +212,12 @@ spv_result_t DataRulesPass(ValidationState_t& _,
       break;
     }
     case SpvOpTypeMatrix: {
-      if (auto error = ValidateMatrixDataType(_, inst)) return error;
+      if (auto error = ValidateMatrixColumnType(_, inst)) return error;
       if (auto error = ValidateMatrixNumCols(_, inst)) return error;
       break;
     }
+    // TODO(ehsan): Add OpSpecConstantComposite validation code.
+    // TODO(ehsan): Add OpSpecConstantOp validation code (if any).
     case SpvOpSpecConstant: {
       if (auto error = ValidateSpecConstNumerical(_, inst)) return error;
       break;
--- a/source/validate_datarules.cpp
+++ b/source/validate_datarules.cpp
@@ -24,6 +24,7 @@
 #include "opcode.h"
 #include "operand.h"
 #include "val/validation_state.h"
+#include "val/instruction.h"
 
 using libspirv::CapabilitySet;
 using libspirv::DiagnosticStream;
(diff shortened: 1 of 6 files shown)


test/val/val_data_test.cpp
```diff
@@ -117,3 +147,233 @@ TEST_F(ValidateData, vec15) {
...
+}
+
+TEST_F(ValidateData, int64_bad) {
+  string str = header + "%2 = OpTypeInt 64 1";
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_ERROR_INVALID_DATA, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(), HasSubstr(missing_int64_cap_error));
+}
+
+// number of bits in an integer may only one of: {8,16,32,64}
```
Qining: may be only one of?

ehsannas:
I agree.
Some tests were constructing programs that did not meet the data validation rules.

test/val/val_data_test.cpp
```diff
@@ -117,3 +147,233 @@ TEST_F(ValidateData, vec15) {
...
+}
+
+TEST_F(ValidateData, float64_bad) {
+  string str = header + "%2 = OpTypeFloat 64";
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_ERROR_INVALID_DATA, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(), HasSubstr(missing_float64_cap_error));
+}
+
+TEST_F(ValidateData, float64_invalid_bits) {
```
Qining: Add a same blurb as the integer test?

test/val/val_data_test.cpp
```diff
@@ -117,3 +147,233 @@ TEST_F(ValidateData, vec15) {
...
+%mat33  =  OpTypeMatrix %vec3 5
+)";
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_ERROR_INVALID_DATA, ValidateInstructions());
+  EXPECT_THAT(
+      getDiagnosticString(),
+      HasSubstr("can only be parameterized as having only 2, 3, or 4 columns"));
+}
+
+TEST_F(ValidateData, specialize_const_to_int) {
```
Qining: specialize_int or specialize_const_to_int. The following tests should have their name in the same pattern.

source/validate_datarules.cpp
```diff
@@ -24,39 +24,163 @@
...
+  }
+  return _.diag(SPV_ERROR_INVALID_DATA) << "Invalid number of bits ("
+                                        << num_bits << ") used for OpTypeInt.";
+}
+
+// Validates that the matrix is parameterized with floating-point types.
+spv_result_t ValidateMatrixDataType(ValidationState_t& _,
+                                    const spv_parsed_instruction_t* inst) {
+  // Find the component type of matrix columns (must be vector).
+  auto type_id = inst->words[inst->operands[1].offset];
```
Qining: Nitpicky, but hope to have a comment about the meaning of '1' here. We use those constants to index specific operands, I feel we'd better leave a clue somehow in case we need to come back in future.

source/validate_datarules.cpp
```diff
@@ -68,9 +192,36 @@ namespace libspirv {
...
+    case SpvOpTypeInt: {
+      if (auto error = ValidateIntSize(_, inst)) return error;
+      break;
+    }
+    case SpvOpTypeMatrix: {
+      if (auto error = ValidateMatrixDataType(_, inst)) return error;
+      if (auto error = ValidateMatrixNumCols(_, inst)) return error;
+      break;
+    }
+    case SpvOpSpecConstant: {
```
Qining: I remember there are more opcodes that can produce spec constant, like OpSpecConstantComposite and OpSpecConstantOp. But I think we can leave them for the next CL. May be have a TODO here?

source/validate_datarules.cpp
```diff
@@ -68,9 +192,36 @@ namespace libspirv {
...
+      if (auto error = ValidateSpecConstNumerical(_, inst)) return error;
+      break;
+    }
+    case SpvOpSpecConstantFalse:
+    case SpvOpSpecConstantTrue: {
+      if (auto error = ValidateSpecConstBoolean(_, inst)) return error;
+      break;
+    }
+    // TODO(ehsan): add more data rules validation here.
+    default: { break; }
```
Qining: Just curious. Any reason to have a `{ break; }` here?

test/val/val_data_test.cpp
ehsannas: 👍

test/val/val_data_test.cpp
ehsannas: 👍

source/validate_datarules.cpp
ehsannas: sure, I'll add some comments. (yeah, it's the index of the operand)

source/validate_datarules.cpp
ehsannas: yeah, those are to be done in a later changelist. I'll add a TODO comment.

source/validate_datarules.cpp
ehsannas: It's not necessary to have it. 
My thought process was: if we add some code for the default branch later, we might forget to add 'break', so why not do it now? :)
The style guide doesn't have a strong preference it seems.

test/val/val_data_test.cpp
ehsannas: 👍

Qining:
@dneto0 PTAL

source/validate_datarules.cpp
```diff
@@ -31,32 +32,187 @@ using libspirv::ValidationState_t;
...
+    }
+    return _.diag(SPV_ERROR_INVALID_DATA)
+           << "Using a 64-bit integer type requires the Int64 capability.";
+  }
+  return _.diag(SPV_ERROR_INVALID_DATA) << "Invalid number of bits ("
+                                        << num_bits << ") used for OpTypeInt.";
+}
+
+// Validates that the matrix is parameterized with floating-point types.
+spv_result_t ValidateMatrixDataType(ValidationState_t& _,
```
dneto0: Suggest Data --> Column since that matches the spec better.
So ValidateMatrixColumnType

source/validate_datarules.cpp
```diff
@@ -31,32 +32,187 @@ using libspirv::ValidationState_t;
...
+                                             "number.";
+  }
+  return SPV_SUCCESS;
+}
+
+// Validates that OpSpecConstantTrue and OpSpecConstantFalse specialize to bool.
+spv_result_t ValidateSpecConstBoolean(ValidationState_t& _,
+                                      const spv_parsed_instruction_t* inst) {
+  // Operand 0 is the <id> of the type that we're specializing to.
+  auto type_id = inst->words[inst->operands[0].offset];
```
dneto0: You can just use inst->type_id directly.  Since it's so common type_id and result_id are already extracted for you.

source/validate_datarules.cpp
dneto0: Generally, just add the TODO.   You never know how long it will take to get a review.  :-)

source/validate_id.cpp
```diff
@@ -355,13 +355,35 @@ bool idUsage::isValid<SpvOpTypeStruct>(const spv_instruction_t* inst,
...
-    auto memberType = module_.FindDef(inst->words[memberTypeIndex]);
+    auto memberTypeId = inst->words[memberTypeIndex];
+    auto memberType = module_.FindDef(memberTypeId);
     if (!memberType || !spvOpcodeGeneratesType(memberType->opcode())) {
       DIAG(memberTypeIndex) << "OpTypeStruct Member Type <id> '"
                             << inst->words[memberTypeIndex]
                             << "' is not a type.";
       return false;
     }
+    if (memberType && module_.IsForwardPointer(memberTypeId)) {
```
dneto0: Unrelated to the commit comment.

test/val/val_data_test.cpp
```diff
@@ -117,3 +153,300 @@ TEST_F(ValidateData, vec15) {
...
+
+TEST_F(ValidateData, int64_bad) {
+  string str = header + "%2 = OpTypeInt 64 1";
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_ERROR_INVALID_DATA, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(), HasSubstr(missing_int64_cap_error));
+}
+
+// Number of bits in an integer may be only one of: {8,16,32,64}
+TEST_F(ValidateData, int64_invalid_bits) {
```
dneto0: The "int64_" prefix on the test name is misleading.

test/val/val_data_test.cpp
```diff
@@ -117,3 +153,300 @@ TEST_F(ValidateData, vec15) {
...
+
+TEST_F(ValidateData, float64_bad) {
+  string str = header + "%2 = OpTypeFloat 64";
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_ERROR_INVALID_DATA, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(), HasSubstr(missing_float64_cap_error));
+}
+
+// Number of bits in a float may be only one of: {16,32,64}
+TEST_F(ValidateData, float64_invalid_bits) {
```
dneto0: The "float64_" prefix on the test name is misleading.

dneto0:
Squashed and pushed into master as d0620061

PR #459, opened 2016-11-15
Adding validation code for OpTypeStruct

ehsannas:
According to the Data Rules section of 2.16.1. Universal Validation
Rules of the SPIR-V Spec:

Forward reference operands in an OpTypeStruct
* must be later declared with OpTypePointer
* the type pointed to must be an OpTypeStruct
* had an earlier OpTypeForwardPointer forward reference to the same <id>

### DIFF
--- a/source/validate_datarules.cpp
+++ b/source/validate_datarules.cpp
@@ -190,6 +190,33 @@ spv_result_t ValidateSpecConstBoolean(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Records the <id> of the forward pointer to be used for validation.
+spv_result_t ValidateForwardPointer(ValidationState_t& _,
+                                    const spv_parsed_instruction_t* inst) {
+  // Record the <id> (which is operand 0) to ensure it's used properly.
+  // OpTypeStruct can only include undefined pointers that are
+  // previously declared as a ForwardPointer
+  return (_.RegisterForwardPointer(inst->words[inst->operands[0].offset]));
+}
+
+// Validates that any undefined component of the struct is a forward pointer.
+// It is valid to declare a forward pointer, and use its <id> as one of the
+// components of a struct.
+spv_result_t ValidateStruct(ValidationState_t& _,
+                            const spv_parsed_instruction_t* inst) {
+  // Struct components are operands 1, 2, etc.
+  for (unsigned i = 1; i < inst->num_operands; i++) {
+    auto type_id = inst->words[inst->operands[i].offset];
+    auto type_instruction = _.FindDef(type_id);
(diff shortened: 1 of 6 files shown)


source/validate_datarules.cpp
```diff
@@ -190,6 +190,33 @@ spv_result_t ValidateSpecConstBoolean(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Records the <id> of the forward pointer to be used for validation.
+spv_result_t ValidateForwardPointer(ValidationState_t& _,
```
dneto0: It's arguably excessive to make this a full blown "Validation" routine since no checks are occurring in this method.  You could have just directly called RegisterForwardPointer instead of calling this function.  But I can see how it's nice to maintain parallel structure at the call site.

Ok to keep.

source/validate_id.cpp
```diff
@@ -355,13 +355,35 @@ bool idUsage::isValid<SpvOpTypeStruct>(const spv_instruction_t* inst,
...
+      // Find out the type that the pointer is pointing to (must be struct)
+      // word 3 is the <id> of the type being pointed to.
+      auto typePointingTo = module_.FindDef(memberType->words()[3]);
+      if (typePointingTo && typePointingTo->opcode() != SpvOpTypeStruct) {
+        // Forward declared operands of a struct may only point to a struct.
+        DIAG(memberTypeIndex)
+            << "A forward reference operand in an "
+               "OpTypeStruct must be an OpTypePointer that "
+               "points to an OpTypeStruct. Found OpTypePointer that points to "
+            << spvOpcodeString(static_cast<SpvOp>(typePointingTo->opcode()))
```
dneto0: Suggest prefixing the opcode string with "Op" for consistency with other parts of this message.

test/val/val_data_test.cpp
```diff
@@ -390,3 +396,69 @@ TEST_F(ValidateData, specialize_boolean_to_int) {
...
+%uintt = OpTypeInt 32 0
+%struct_B = OpTypeStruct %uintt %_ptr_Generic_struct_A
+%_ptr_Generic_struct_A = OpTypePointer Generic %uintt
+)";
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("A forward reference operand in an OpTypeStruct must "
+                        "be an OpTypePointer that points to an OpTypeStruct. "
+                        "Found OpTypePointer that points to TypeInt."));
```
dneto0: The bare "TypeInt" looks wrong here. Prefix it with "Op".

dneto0:
Squashed and pushed into master as 8c414eb579876

PR #461, opened 2016-11-16
Adding validation for OpSpecConstantComposite.

ehsannas:
SpecConstantComposite may specialize to a vector, matrix, array, or
struct. In each case, the number of components and type of components
that are being specialized to must match the expected result type.

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -710,11 +710,196 @@ bool idUsage::isValid<SpvOpSpecConstantFalse>(const spv_instruction_t* inst,
   return true;
 }
 
-#if 0
 template <>
-bool idUsage::isValid<SpvOpSpecConstantComposite>(
-    const spv_instruction_t *inst, const spv_opcode_desc opcodeEntry) {}
-#endif
+bool idUsage::isValid<SpvOpSpecConstantComposite>(const spv_instruction_t* inst,
+                                                  const spv_opcode_desc) {
+  // The result type must be a composite type.
+  auto resultTypeIndex = 1;
+  auto resultType = module_.FindDef(inst->words[resultTypeIndex]);
+  if (!resultType || !spvOpcodeIsComposite(resultType->opcode())) {
+    DIAG(resultTypeIndex) << "OpSpecConstantComposite Result Type <id> '"
+                          << inst->words[resultTypeIndex]
+                          << "' is not a composite type.";
+    return false;
+  }
+  // Validation checks differ based on the type of composite type.
+  auto constituentCount = inst->words.size() - 3;
+  switch (resultType->opcode()) {
+    // For Vectors, the following must be met:
+    // * Number of constituents in the result type and the vector must match.
+    // * All the components of the vector must have the same type (or specialize
+    // to the same type). OpConstant and OpSpecConstant are allowed.
+    case SpvOpTypeVector: {
+      auto componentCount = resultType->words()[3];
+      if (componentCount != constituentCount) {
+        DIAG(inst->words.size() - 1)
+            << "OpSpecConstantComposite Constituent <id> count does not match "
+               "Result Type <id> '"
+            << resultType->id() << "'s vector component count.";
+        return false;
+      }
+      auto componentType = module_.FindDef(resultType->words()[2]);
+      assert(componentType);
+      for (size_t constituentIndex = 3; constituentIndex < inst->words.size();
+           constituentIndex++) {
+        auto constituent = module_.FindDef(inst->words[constituentIndex]);
+        if (!constituent ||
+            !spvOpcodeIsConstantOrUndef(constituent->opcode())) {
+          DIAG(constituentIndex) << "OpSpecConstantComposite Constituent <id> '"
+                                 << inst->words[constituentIndex]
+                                 << "' is not a constant or undef.";
+          return false;
+        }
+        auto constituentResultType = module_.FindDef(constituent->type_id());
+        if (!constituentResultType ||
+            componentType->opcode() != constituentResultType->opcode()) {
+          DIAG(constituentIndex) << "OpSpecConstantComposite Constituent <id> '"
+                                 << inst->words[constituentIndex]
+                                 << "'s type does not match Result Type <id> '"
+                                 << resultType->id()
+                                 << "'s vector element type.";
+          return false;
+        }
+      }
+      break;
+    }
+    case SpvOpTypeMatrix: {
+      auto columnCount = resultType->words()[3];
+      if (columnCount != constituentCount) {
+        DIAG(inst->words.size() - 1)
+            << "OpSpecConstantComposite Constituent <id> count does not match "
+               "Result Type <id> '"
+            << resultType->id() << "'s matrix column count.";
+        return false;
+      }
+
+      auto columnType = module_.FindDef(resultType->words()[2]);
+      assert(columnType);
+      auto componentCount = columnType->words()[3];
+      auto componentType = module_.FindDef(columnType->words()[2]);
+      assert(componentType);
+
+      for (size_t constituentIndex = 3; constituentIndex < inst->words.size();
+           constituentIndex++) {
+        auto constituent = module_.FindDef(inst->words[constituentIndex]);
+        auto constituentOpCode = constituent->opcode();
+        if (!constituent ||
+            !(SpvOpSpecConstantComposite == constituentOpCode ||
+              SpvOpConstantComposite == constituentOpCode ||
+              SpvOpUndef == constituentOpCode)) {
+          // The message says "... or undef" because the spec does not say
+          // undef is a constant.
+          DIAG(constituentIndex) << "OpSpecConstantComposite Constituent <id> '"
+                                 << inst->words[constituentIndex]
+                                 << "' is not a constant composite or undef.";
+          return false;
+        }
+        auto vector = module_.FindDef(constituent->type_id());
+        assert(vector);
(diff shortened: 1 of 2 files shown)


test/val/val_id_test.cpp
```diff
@@ -941,7 +941,299 @@ TEST_F(ValidateID, OpSpecConstantBad) {
...
+}
+
+// Invalid: result type is float.
+TEST_F(ValidateID, OpSpecConstantCompositeVectorResultTypeBad) {
+  const char* spirv = R"(
+%1 = OpTypeFloat 32
+%2 = OpTypeVector %1 4
+%3 = OpSpecConstant %1 3.14
+%4 = OpSpecConstantComposite %1 %3 %3 %3 %3)";
+  CHECK(spirv, SPV_ERROR_INVALID_ID);
```
dneto0: Let's check message substrings as well.  That way it's easier to tell if we've covered all the error cases in the code.

source/validate_id.cpp
```diff
@@ -710,11 +710,196 @@ bool idUsage::isValid<SpvOpSpecConstantFalse>(const spv_instruction_t* inst,
...
-bool idUsage::isValid<SpvOpSpecConstantComposite>(
-    const spv_instruction_t *inst, const spv_opcode_desc opcodeEntry) {}
-#endif
+bool idUsage::isValid<SpvOpSpecConstantComposite>(const spv_instruction_t* inst,
+                                                  const spv_opcode_desc) {
+  // The result type must be a composite type.
+  auto resultTypeIndex = 1;
+  auto resultType = module_.FindDef(inst->words[resultTypeIndex]);
+  if (!resultType || !spvOpcodeIsComposite(resultType->opcode())) {
+    DIAG(resultTypeIndex) << "OpSpecConstantComposite Result Type <id> '"
```
dneto0: Missing a test case for this diagnostic.

source/validate_id.cpp
```diff
@@ -710,11 +710,196 @@ bool idUsage::isValid<SpvOpSpecConstantFalse>(const spv_instruction_t* inst,
...
+                          << inst->words[resultTypeIndex]
+                          << "' is not a composite type.";
+    return false;
+  }
+  // Validation checks differ based on the type of composite type.
+  auto constituentCount = inst->words.size() - 3;
+  switch (resultType->opcode()) {
+    // For Vectors, the following must be met:
+    // * Number of constituents in the result type and the vector must match.
+    // * All the components of the vector must have the same type (or specialize
```
dneto0: This condition "All components of the vector must have the same type.." should be checkable when we first see the vector type declaration.  We should do it there, since we need it there anyway.
That would simplify this code.

source/validate_id.cpp
```diff
@@ -710,11 +710,196 @@ bool idUsage::isValid<SpvOpSpecConstantFalse>(const spv_instruction_t* inst,
...
+
+      for (size_t constituentIndex = 3; constituentIndex < inst->words.size();
+           constituentIndex++) {
+        auto constituent = module_.FindDef(inst->words[constituentIndex]);
+        auto constituentOpCode = constituent->opcode();
+        if (!constituent ||
+            !(SpvOpSpecConstantComposite == constituentOpCode ||
+              SpvOpConstantComposite == constituentOpCode ||
+              SpvOpUndef == constituentOpCode)) {
+          // The message says "... or undef" because the spec does not say
```
dneto0: Good comment.

source/validate_id.cpp
```diff
@@ -710,11 +710,196 @@ bool idUsage::isValid<SpvOpSpecConstantFalse>(const spv_instruction_t* inst,
...
+        assert(vector);
+        if (columnType->opcode() != vector->opcode()) {
+          DIAG(constituentIndex) << "OpSpecConstantComposite Constituent <id> '"
+                                 << inst->words[constituentIndex]
+                                 << "' type does not match Result Type <id> '"
+                                 << resultType->id()
+                                 << "'s matrix column type.";
+          return false;
+        }
+        auto vectorComponentType = module_.FindDef(vector->words()[2]);
```
dneto0: I think this and the next check are redundant?
I think that's so because the matrix components are vector types, and if those match then we already know they match in their own vector element and element-counts. ?

source/validate_id.cpp
ehsannas: This comment is perhaps not well-written; I apologize. I should rewrite the comment.
What this code does is that it ensures that each of the operands of OpSpecConstantComposite has the proper type. For example, for the code below, it ensures that %f1 and %f2 and %f3 and %f4 have the type that %vec4 components should have (i.e. ensures they're all float). There is no instruction before OpSpecConstantComposite that could validate this.

```
%floatt = OpTypeFloat 32
%vec4  = OpTypeVector %floatt 4
%f1       = OpSpecConstant %floatt 1.0
%f2       = OpSpecConstant %floatt 2.0
%f3       = OpSpecConstant %floatt 3.0
%f4       = OpSpecConstant %floatt 4.0
%spec_constt = OpSpecConstantComposite %vec4 %f1 %f2 %f3 %f4
```

Does that make sense? Or am I missing your point?

source/validate_id.cpp
ehsannas: Similarly, here OpSpecConstantComposite is the command which ties the different columns into 1 matrix. So this is where we can validate that different columns have the same component type and have the same size. We don't look at all the entries in all of the columns; we just check that the column vector component types and sizes match for different columns.

(OpSpecConstantComposite / OpConstantComposite that builds each column does validate that all entires in the column vector are of the same type and that the size of the vector is correct.)

Does that make sense?

dneto0: Ah, there's a uniqueness condition I'm relying on.  It's invalid to declare two different vector types with the same element type and element count.  That's not yet checked, but ought to be.

From the spec. 2.8 paragraph 6:

"Non-aggregate types are different: It is invalid to declare multiple type <id>s for the same scalar, vector, or matrix type. That is, non-aggregate type declarations must all have different opcodes or operands. (Note that non-aggregate types cannot be decorated in ways that affect their type.)"

Ok. I recommend:

1. adding a TODO saying these checks should be come dead code eventually.
2. test for the message text.  Once the vector-uniqueness checks are added, then we'll see these messages not be generated, and we can delete these checks.

ehsannas: Ah, that makes sense. Will do.

ehsannas:
:+1: Validator passes on all tests in Vulkan CTS

test/val/val_id_test.cpp
```diff
@@ -112,6 +114,26 @@ const char kOpenCLMemoryModel64[] = R"(
...
+    spvDiagnosticDestroy(diagnostic);                                          \
+    ASSERT_EQ(SPV_SUCCESS, error) << shader;                                   \
+  }                                                                            \
+  spv_result_t result = spvValidate(context, get_const_binary(), &diagnostic); \
+  if (SPV_SUCCESS != result) {                                                 \
+    spvDiagnosticPrint(diagnostic);                                            \
+    EXPECT_THAT(diagnostic->error, HasSubstr(err_substr));                     \
+    spvDiagnosticDestroy(diagnostic);                                          \
+  }                                                                            \
+  ASSERT_EQ(expected, result);                                                 \
```
dneto0: Normally use EXPECT_EQ .
Use ASSERT_EQ when you check a condition that gates any further execution.  For example, use ASSERT_NE(nullptr, p)  when in following steps you dereference pointer p.  The assert prevents a crash in a test failure case.

test/val/val_id_test.cpp
```diff
@@ -112,6 +114,26 @@ const char kOpenCLMemoryModel64[] = R"(
   ASSERT_EQ(expected, result);                                                 \
   spvContextDestroy(context);
 
+#define CHECK_WITH_DIAGNOSTIC(str, expected, err_substr)                       \
```
dneto0: This is quite a lot of new code, in a macro.

Much better to follow the example in val_ssa_test.cpp 
There each test is just a string constant, then two or three more lines.  The heavy lifting of object management is taken care of in the test fixture base class.

dneto0: It's a good exercise to streamline this code now.  Please use the fixture, e.g.
```
using ValidateWithMessage = spvtest::ValidateBase<bool>;
...
TEST_F(ValidateWithMessage, ...)  {
   char str[] = " .... ";
   CompileSuccessfully(str);
   EXPECT_THAT(getDiagnosticString(), HasSubstr("foobar"));
}
```

ehsannas:
👍 Validator passes on VKCTS.

test/val/val_id_test.cpp
```diff
@@ -33,14 +35,9 @@ using std::ostringstream;
...
-class ValidateID : public ::testing::Test {
- public:
-  virtual void TearDown() { spvBinaryDestroy(binary); }
-  spv_const_binary get_const_binary() { return spv_const_binary(binary); }
-  spv_binary binary;
-};
+using ValidateIdWithMessage = spvtest::ValidateBase<bool>;
 
-const char kGLSL450MemoryModel[] = R"(
+string kGLSL450MemoryModel = R"(
```
dneto0: FYI the Google C++ style guide forbids static-duration objects with C++ constructors.  std::string is an example.  But the reasons for it don't really apply for a test executable that starts and shuts down so trivially quickly.

dneto0:
Squashed and pushed into master as https://github.com/KhronosGroup/SPIRV-Tools/commit/6fa6a3f64703745425bab6bf51da984f765656e2

PR #448, opened 2016-10-21
[WIP] Type checking

umar456:
This PR attempts to create a data structure which can completely
represent all type requirements of the SPIR-V spec. This object
will be used to validate the types of a binary SPIR-V binary. The
current state of the struct is incomplete as it does not provide
the necessary fields to represent Capability decoration, and other
requirements of the spec.

I have created a list of type requirements for all instructions in the
spec here:
https://gist.github.com/umar456/c6cb1ca3cad11d5e7841e40572405308

The PR also contains code to validate type aliasing rules of the spec.

PR #462, opened 2016-11-19
WIP:  Rebased: Validator type checking

dneto0:
Do not review.

This is a rebase of @umar456's work in https://github.com/KhronosGroup/SPIRV-Tools/pull/448
Key differences are:

* React to file name changing
* Squash compile fixes into previous commit (e.g. signed-vs.-unsigned comparison, include cassert)
* Collapse copyright notice fixes into original commits of those .cpp files

Warning: The tests do not pass.   In some cases I checked it's a matter of a different failure being triggered.


source/val/type.h
```diff
@@ -0,0 +1,71 @@
...
+    if (the_category == category_ &&
+        operands().size() == size_t(inst.num_words - 2) &&
+        equal(begin(operands()), end(operands()),
+              inst.words + inst.operands->offset + 1)) {
+      return true;
+    }
+    return false;
+  }
+
+  bool IsType(spv_type_category_t cat) { return cat & category_; }
```
ehsannas: if `category_` is `SPV_TYPE_CATEGORY_SCALAR`, then `IsType(SPV_TYPE_CATEGORY_BOOL)` would return TRUE. I find that confusing (and it could lead to wrong usage if caller is not careful).
A Bool is a Scalar but a Scalar is not necessarily a Bool.

source/val/type.h
```diff
@@ -0,0 +1,71 @@
...
+  bool IsTypeAny(const std::vector<spv_type_category_t>& possibilities) {
+    return any_of(begin(possibilities), end(possibilities),
```
ehsannas: +1 for using `std::any_of`

source/val/type.h
```diff
@@ -0,0 +1,71 @@
...
+                  [this](spv_type_category_t cat) { return cat & category_; });
```
ehsannas: This lambda is basically the `IsType` member function. I wonder if there's a way to refer to that function instead of repeating yourself.

source/val/validation_state.cpp
```diff
@@ -395,4 +395,20 @@ void ValidationState_t::RegisterInstruction(
...
+    if (type.IsAlias(inst)) {
+      return &type;
+    }
+  }
+  return nullptr;
+}
+
+bool ValidationState_t::AddType(const spv_parsed_instruction_t& inst) {
+  types_.emplace(make_pair(inst.result_id, Type{inst}));
+  return true;
```
ehsannas: Why not make the function `void` if always returning `true`?

umar456:
I think the approach I was taking with this PR were kind of optimistic. I don't think it will be possible to expand this to the rest of the types.

I would suggest you look at this table: https://gist.github.com/umar456/c6cb1ca3cad11d5e7841e40572405308

I made this to get a better idea of the type requirements from the spec. I think you should be able to extract a set of general rules that can be used to create a data driven approach to type validation. You may need some special cases for OpTypeFunction and OpPhi but it should handle most instructions.

source/val/type.h
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2015-2016 The Khronos Group Inc.
```
dneto0: KhronosGroup.  This was written under contract to Khronos.  I believe it was as "work for hire".

dneto0:
Thanks @umar456 for laying the groundwork and thinking through the problem.  It's invaluable.

source/validate_id.cpp
```diff
@@ -2352,8 +2325,8 @@ bool idUsage::isValid(const spv_instruction_t* inst) {
     CASE(OpTypePointer)
     CASE(OpTypeFunction)
     CASE(OpTypePipe)
-    CASE(OpConstantTrue)
-    CASE(OpConstantFalse)
+    TODO(OpConstantTrue)
+    TODO(OpConstantFalse)
     CASE(OpConstantComposite)
     CASE(OpConstantSampler)
```
ehsannas: must be changed to `TODO(OpConstantSampler)` as the function has been removed.

source/validate_type.cpp
```diff
@@ -0,0 +1,202 @@
...
+// TODO(umar): Support decoration dependant checks
+// TODO(umar): Support type width and sign dependant checks
+struct spv_type_desc_t {
+  const SpvOp opcode;
+  const vector<spv_type_category_t> result_type;
+};
+
+// NOTE: This will be generated at build time for a specific version
+// of spirv. Ideally this should be generated by the spec and added
+// to the build script as an external dependency.
```
ehsannas: 👍  Overall I think this is a good approach. But we should definitely make it data-driven.

source/validate_type.cpp
```diff
@@ -0,0 +1,202 @@
...
+      begin(result_types), end(result_types),
+      [opcode](const spv_type_desc_t& type) { return type.opcode == opcode; });
+
+  if (found != end(result_types)) {
+    auto* type = _.types(operand);
+    if (!type->IsTypeAny(found->result_type)) {
+      return _.diag(SPV_ERROR_INVALID_TYPE)
+             << "Opcode " << spvOpcodeString(opcode)
+             << " requires the result type to be " +
+                    TypeFlagsToString(found->result_type) + " found " +
```
ehsannas: Improve the error message (punctuation).

source/validate_type.cpp
```diff
@@ -0,0 +1,202 @@
...
+    case SpvOpTypeFunction:       ret = SPV_TYPE_CATEGORY_TYPE_FUNCTION; break;
+    case SpvOpTypeOpaque:         ret = SPV_TYPE_CATEGORY_TYPE_OPAQUE; break;
+    case SpvOpTypeEvent:          ret = SPV_TYPE_CATEGORY_TYPE_EVENT; break;
+    case SpvOpTypeDeviceEvent:    ret = SPV_TYPE_CATEGORY_TYPE_DEVICEEVENT; break;
+    case SpvOpTypeReserveId:      ret = SPV_TYPE_CATEGORY_TYPE_RESERVEID; break;
+    case SpvOpTypeQueue:          ret = SPV_TYPE_CATEGORY_TYPE_QUEUE; break;
+    case SpvOpTypePipe:           ret = SPV_TYPE_CATEGORY_TYPE_PIPE; break;
+    case SpvOpTypeForwardPointer: ret = SPV_TYPE_CATEGORY_TYPE_FORWARDPOINTER; break;
+    default:
+      assert(1 == 0 && "Not Defined");
```
ehsannas: simplify to:
```
assert(false && "Reached undefined SpvOpType.");
```

source/validate_type.cpp
```diff
@@ -0,0 +1,202 @@
...
+}
+// clang-format on
+
+spv_result_t TypePass(ValidationState_t& _,
+                      const spv_parsed_instruction_t* inst) {
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+  if (spvOpcodeGeneratesType(opcode)) {
+    // Aggregate types can contain aliases
+    if (opcode != SpvOpTypeArray && opcode != SpvOpTypeStruct &&
+        opcode != SpvOpTypeRuntimeArray) {
```
ehsannas: I couldn't find the definition of SpvOpTypeRuntimeArray in the SPIRV 1.1 spec.

test/val/val_fixtures.h
```diff
@@ -18,32 +18,65 @@
...
+      spv_target_env env = SPV_ENV_UNIVERSAL_1_0) {
+    return spvValidateBinary(ScopedContext(env).context,
+                       words_.data(), words_.size(),
+                       &diagnostic_);
+  }
 
-  std::string getDiagnosticString();
-  spv_position_t getErrorPosition();
+  std::string getDiagnosticString() {
+    //std::cout << diagnostic_->error << std::endl;
```
ehsannas: remove debugging code.

test/val/val_fixtures.h
```diff
@@ -18,32 +18,65 @@
...
+    return std::string(diagnostic_->error); }
+  spv_position_t getErrorPosition() { return diagnostic_->position; }
 
-  spv_binary binary_;
   spv_diagnostic diagnostic_;
+  std::vector<uint32_t> words_;
```
ehsannas: I see that `spv_binary binary_` has been replaced with `std::vector<uint32_t> words_`.
And also val_fixtures.cpp has been fully removed (and some functionality has been moved to val_fixtures.h). Is there a particular reason for these changes?

I think this part of the CL may be either removed (not necessary) or separated out from the type checking code.

antiagainst:
This PR has been here for quite some time. Maybe we should devote some time to rebase and merge it in instead of keeping it here and letting it lag behind.

dneto0:
@atgoo has been writing type checks one piece at a time.  Closing.

PR #463, opened 2016-11-22
Validation for OpSampledImage instruction.

ehsannas:
This change implements the validation for usages of OpSampledImage
instruction as described in the Data Rules section of the Universal
Validation Rules of the SPIR-V Spec.

All OpSampledImage instructions must be in the same block in which their Result <id> are consumed. Result <id> from OpSampledImage instructions must not appear as operands to OpPhi instructions or OpSelect instructions, or any instructions other than the image lookup and query instructions specified to take an operand whose type is OpTypeSampledImage.

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -398,23 +398,24 @@ void ValidationState_t::RegisterInstruction(
   // If the instruction is using an OpTypeSampledImage as an operand, it should
   // be recorded. The validator will ensure that all usages of an
   // OpTypeSampledImage and its definition are in the same basic block.
-  for (auto i = 0; i < inst.num_operands; ++i) {
-    auto operand = inst.words[inst.operands[i].offset];
-    auto operand_inst = FindDef(operand);
-    if (operand_inst && SpvOpSampledImage == operand_inst->opcode()) {
-      RegisterSampledImageConsumer(operand, inst.result_id);
+  for (uint16_t i = 0; i < inst.num_operands; ++i) {
+    const spv_parsed_operand_t& operand = inst.operands[i];
+    const uint32_t operand_word = inst.words[operand.offset];
+    if (SPV_OPERAND_TYPE_ID == operand.type) {
+      Instruction* operand_inst = FindDef(operand_word);
+      if (operand_inst && SpvOpSampledImage == operand_inst->opcode()) {
+        RegisterSampledImageConsumer(operand_word, inst.result_id);
+      }
     }
   }
 }
 
-std::vector<uint32_t> ValidationState_t::getSampledImageConsumers(
+const std::vector<uint32_t> ValidationState_t::getSampledImageConsumers(
     uint32_t sampled_image_id) const {
   std::vector<uint32_t> result;
   auto iter = sampled_image_consumers_.find(sampled_image_id);
   if (iter != sampled_image_consumers_.end()) {
-    for (auto id : iter->second) {
-      result.push_back(id);
-    }
+    result = iter->second;
   }
   return result;
 }
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -394,5 +394,34 @@ void ValidationState_t::RegisterInstruction(
   if (id) {
     all_definitions_.insert(make_pair(id, &ordered_instructions_.back()));
   }
+
+  // If the instruction is using an OpTypeSampledImage as an operand, it should
+  // be recorded. The validator will ensure that all usages of an
+  // OpTypeSampledImage and its definition are in the same basic block.
+  for (auto i = 0; i < inst.num_operands; ++i) {
+    auto operand = inst.words[inst.operands[i].offset];
+    auto operand_inst = FindDef(operand);
+    if (operand_inst && SpvOpSampledImage == operand_inst->opcode()) {
+      RegisterSampledImageConsumer(operand, inst.result_id);
+    }
+  }
 }
+
+std::vector<uint32_t> ValidationState_t::getSampledImageConsumers(
+    uint32_t sampled_image_id) const {
+  std::vector<uint32_t> result;
+  auto iter = sampled_image_consumers_.find(sampled_image_id);
+  if (iter != sampled_image_consumers_.end()) {
+    for (auto id : iter->second) {
+      result.push_back(id);
+    }
+  }
+  return result;
+}
+
+void ValidationState_t::RegisterSampledImageConsumer(uint32_t sampled_image_id,
+                                                     uint32_t consumer_id) {
+  sampled_image_consumers_[sampled_image_id].push_back(consumer_id);
+}
+
 }  /// namespace libspirv
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -400,8 +400,8 @@ void ValidationState_t::RegisterInstruction(
   // OpTypeSampledImage and its definition are in the same basic block.
   for (uint16_t i = 0; i < inst.num_operands; ++i) {
     const spv_parsed_operand_t& operand = inst.operands[i];
-    const uint32_t operand_word = inst.words[operand.offset];
     if (SPV_OPERAND_TYPE_ID == operand.type) {
+      const uint32_t operand_word = inst.words[operand.offset];
       Instruction* operand_inst = FindDef(operand_word);
       if (operand_inst && SpvOpSampledImage == operand_inst->opcode()) {
         RegisterSampledImageConsumer(operand_word, inst.result_id);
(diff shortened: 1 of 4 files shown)


ehsannas:
Validator passed VKCTS

source/val/validation_state.cpp
```diff
@@ -394,5 +394,34 @@ void ValidationState_t::RegisterInstruction(
   if (id) {
     all_definitions_.insert(make_pair(id, &ordered_instructions_.back()));
   }
+
+  // If the instruction is using an OpTypeSampledImage as an operand, it should
+  // be recorded. The validator will ensure that all usages of an
+  // OpTypeSampledImage and its definition are in the same basic block.
+  for (auto i = 0; i < inst.num_operands; ++i) {
```
dneto0: In this declaration of i "auto" is a longer fancy name for "int".  Since the literal value 0 is of type int.  :-)

source/val/validation_state.cpp
```diff
@@ -394,5 +394,34 @@ void ValidationState_t::RegisterInstruction(
...
+    auto operand = inst.words[inst.operands[i].offset];
+    auto operand_inst = FindDef(operand);
```
dneto0: This must be guarded by a check that the operand type is SPV_OPERAND_TYPE_ID.

The proposed code can lead to false positives when the operand is, for example, a literal integer whose value happens to match the result id of an an OpSampledImage instruction.

source/val/validation_state.cpp
```diff
@@ -394,5 +394,34 @@ void ValidationState_t::RegisterInstruction(
...
+    if (operand_inst && SpvOpSampledImage == operand_inst->opcode()) {
+      RegisterSampledImageConsumer(operand, inst.result_id);
```
dneto0: Side note, not to be resolved by this issue.
The optimization framework has full definition-use tracking, of which this ID-tracking is a special case.  It's too bad the validator and optimizer don't share more code. :-(

source/validate_id.cpp
```diff
@@ -734,6 +734,48 @@ bool idUsage::isValid<SpvOpSpecConstantFalse>(const spv_instruction_t* inst,
...
+  auto resultID = inst->words[resultTypeIndex];
+  auto sampledImageInstr = module_.FindDef(resultID);
+  // We need to validate 2 things:
+  // * All OpSampledImage instructions must be in the same block in which their
+  // Result <id> are consumed.
+  // * Result <id> from OpSampledImage instructions must not appear as operands
+  // to OpPhi instructions or OpSelect instructions, or any instructions other
+  // than the image lookup and query instructions specified to take an operand
+  // whose type is OpTypeSampledImage.
+  std::vector<uint32_t> consumers = module_.getSampledImageConsumers(resultID);
```
dneto0: make consumers const ref

source/val/validation_state.cpp
```diff
@@ -394,5 +394,34 @@ void ValidationState_t::RegisterInstruction(
...
+    }
+  }
+}
+
+std::vector<uint32_t> ValidationState_t::getSampledImageConsumers(
+    uint32_t sampled_image_id) const {
+  std::vector<uint32_t> result;
+  auto iter = sampled_image_consumers_.find(sampled_image_id);
+  if (iter != sampled_image_consumers_.end()) {
+    for (auto id : iter->second) {
```
dneto0: No need to iterate to fill the vector here.  Make the method return a const ref to the vector. Return *iter.  Then I think no copies are made.

source/val/validation_state.h
```diff
@@ -179,6 +179,14 @@ class ValidationState_t {
     return all_definitions_;
   }
 
+  /// Returns a vector containing the Ids of instructions that consume the given
+  /// SampledImage id.
+  std::vector<uint32_t> getSampledImageConsumers(uint32_t id) const;
```
dneto0: Make this return const ref.

source/val/validation_state.cpp
```diff
@@ -394,5 +394,34 @@ void ValidationState_t::RegisterInstruction(
...
+      result.push_back(id);
+    }
+  }
+  return result;
```
dneto0: And here just return a newly constructed emtpy vector:  `return  std::vector<uint32_t>{};`

source/validate_id.cpp
```diff
@@ -734,6 +734,48 @@ bool idUsage::isValid<SpvOpSpecConstantFalse>(const spv_instruction_t* inst,
...
+        DIAG(resultTypeIndex)
+            << "All OpSampledImage instructions must be in the same block in "
+               "which their Result <id> are consumed. OpSampledImage Result "
+               "Type <id> '"
+            << resultID << "' has a consumer in a different basic "
+                           "block. The consumer instruction <id> is '"
+            << consumer_id << "'.";
+        return false;
+      }
+      if (consumer_opcode == SpvOpPhi || consumer_opcode == SpvOpSelect) {
```
dneto0: Add a comment saying this check is incomplete.  E.g. you don't want the sampled image value to be the Object argument of a Store instruction.
This merits a "TODO" to say maybe we should do better, e.g. by scanning for "Sampled Image" in the description field in the grammar file.

test/val/val_id_test.cpp
```diff
@@ -1925,6 +1925,169 @@ TEST_F(ValidateIdWithMessage, OpFunctionCallArgumentTypeBad) {
...
+         %24             = OpConstantComposite %v2float %float_2 %float_2
+         %40             = OpTypeBool
+         %41             = OpSpecConstantTrue %40
+       %main             = OpFunction %void None %3
+          %5             = OpLabel
+          %x             = OpVariable %_ptr_Function_v4float Function
+         %13             = OpLoad %10 %tex
+         %17             = OpLoad %14 %s
+         %25             = OpSampledImage %18 %13 %17
+         OpBranchConditional %41 %50 %70
```
dneto0: This will fail validation in Shader due to structured control flow rules.  You're missing an OpSelectionMerge.

Much better still is to use two basic blocks, and an unconditional branch from the first to the second.

test/val/val_id_test.cpp
```diff
@@ -1925,6 +1925,169 @@ TEST_F(ValidateIdWithMessage, OpFunctionCallArgumentTypeBad) {
   CompileSuccessfully(spirv.c_str());
   EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
 }
+
+// Valid: OpSampledImage result <id> is used in the same block by
+// OpImageSampleImplictLod
+TEST_F(ValidateIdWithMessage, OpSampledImageGood) {
+  string spirv = kGLSL450MemoryModel + R"(
+       %void             = OpTypeVoid
```
dneto0: This is a lot of boilerplate to set up the types and function preamble.  Better to make everything before the first OpLabel its own module-scope constant character array, so that each test can focus on just its own special characteristics.

This will also mean you'll want to make nicer names for the %<number> IDs.

source/val/validation_state.cpp
ehsannas: Yes! Good catch

source/val/validation_state.cpp
ehsannas: hmm... I can't return a reference to a local temporary object, right?
should I instead return a const pointer rather than a const reference?

source/val/validation_state.cpp
```diff
@@ -398,23 +398,24 @@ void ValidationState_t::RegisterInstruction(
...
   // be recorded. The validator will ensure that all usages of an
   // OpTypeSampledImage and its definition are in the same basic block.
-  for (auto i = 0; i < inst.num_operands; ++i) {
-    auto operand = inst.words[inst.operands[i].offset];
-    auto operand_inst = FindDef(operand);
-    if (operand_inst && SpvOpSampledImage == operand_inst->opcode()) {
-      RegisterSampledImageConsumer(operand, inst.result_id);
+  for (uint16_t i = 0; i < inst.num_operands; ++i) {
+    const spv_parsed_operand_t& operand = inst.operands[i];
+    const uint32_t operand_word = inst.words[operand.offset];
```
dneto0: operand_word is only used inside the "if", so move it down there.

source/val/validation_state.cpp
```diff
@@ -398,23 +398,24 @@ void ValidationState_t::RegisterInstruction(
...
+      Instruction* operand_inst = FindDef(operand_word);
+      if (operand_inst && SpvOpSampledImage == operand_inst->opcode()) {
+        RegisterSampledImageConsumer(operand_word, inst.result_id);
+      }
     }
   }
 }
 
-std::vector<uint32_t> ValidationState_t::getSampledImageConsumers(
+const std::vector<uint32_t> ValidationState_t::getSampledImageConsumers(
```
dneto0: Sorry I made you try this.  But a const here *without* the ref is kind of useless.  At this point it's just clutter. :-(

dneto0:
I'll wait for the bots to come back green before squashing and pushing into master.

dneto0:
Squashed and pushed into master as f72189c249b

PR #465, opened 2016-11-24
Validate limits (ID bound)

ehsannas:
Adding validation checks that all IDs are within the ID bound specified in the
SPIR-V header. Unit tests have also been added for this and pass successfully.

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -425,4 +425,7 @@ void ValidationState_t::RegisterSampledImageConsumer(uint32_t sampled_image_id,
   sampled_image_consumers_[sampled_image_id].push_back(consumer_id);
 }
 
+uint32_t ValidationState_t::getIdBound() const { return id_bound_; }
+
+void ValidationState_t::setIdBound(const uint32_t bound) { id_bound_ = bound; }
 }  /// namespace libspirv
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -131,47 +131,13 @@ spv_result_t CapCheck(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
-// Checks that the number of characters in a string doesn't exceed the limit.
-spv_result_t LimitCheckString(ValidationState_t& _,
-                              const spv_parsed_instruction_t* inst) {
-  // By definition, num_words of an operand is a 16-bit unsigned which will not
-  // exceed the limit of characters for a string literal.
-  // This function checks that a NULL termination must be found before reaching
-  // the end of the string words.
-  for (int i = 0; i < inst->num_operands; ++i) {
-    auto operand = inst->operands[i];
-    if (SPV_OPERAND_TYPE_LITERAL_STRING == operand.type) {
-      bool null_terminated = false;
-      for (auto word_i = 0; !null_terminated && word_i < operand.num_words;
-           ++word_i) {
-        const uint32_t word = inst->words[operand.offset + word_i];
-        // check if any BYTE in the word is NULL.
-        if ((word & 0x000000ff) == 0 || (word & 0x0000ff00) == 0 ||
-            (word & 0x00ff0000) == 0 || (word & 0xff000000) == 0) {
-          null_terminated = true;
-        }
-      }
-      if (!null_terminated) {
-        return _.diag(SPV_ERROR_INVALID_BINARY)
-               << "Found a string that is not null-terminated.";
-      }
-    }
-  }
-  return SPV_SUCCESS;
-}
-
 // Checks that the Resuld <id> is within the valid bound.
 spv_result_t LimitCheckIdBound(ValidationState_t& _,
                                const spv_parsed_instruction_t* inst) {
   for (int i = 0; i < inst->num_operands; ++i) {
-    auto operand = inst->operands[i];
+    const auto& operand = inst->operands[i];
     if (SPV_OPERAND_TYPE_RESULT_ID == operand.type) {
       const uint32_t result_id = inst->words[operand.offset];
-      if (result_id <= 0) {
-        return _.diag(SPV_ERROR_INVALID_BINARY)
-               << "Result <id> '" << result_id
-               << "' is not a valid ID as it must be larger than 0.";
-      }
       if (result_id >= _.getIdBound()) {
         return _.diag(SPV_ERROR_INVALID_BINARY)
                << "Result <id> '" << result_id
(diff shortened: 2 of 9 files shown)


ehsannas:
This change will not be pushed as is. The failure in the "test_val_capability" unit test must be resolved first.

dneto0:
Right.  The failing test should be modified to override the Id bound to make it bigger than the 64 in the faked Id.

source/val/validation_state.cpp
```diff
@@ -425,4 +425,7 @@ void ValidationState_t::RegisterSampledImageConsumer(uint32_t sampled_image_id,
   sampled_image_consumers_[sampled_image_id].push_back(consumer_id);
 }
 
+uint32_t ValidationState_t::getIdBound() const { return id_bound_; }
```
dneto0: It's good to match surrounding code, but you should know that Google C++ style guide permits (encourages?) accessors like this to just use the root of the member name.
E.g . in this case it would be    id_bound().  
Also, it's perfectly fine to put the body in the class declaration, and hence be available for inlining.  Similar for the setter.
See last paragraph in https://google.github.io/styleguide/cppguide.html#Function_Names

source/validate_instruction.cpp
```diff
@@ -131,6 +131,68 @@ spv_result_t CapCheck(ValidationState_t& _,
...
+        const uint32_t word = inst->words[operand.offset + word_i];
+        // check if any BYTE in the word is NULL.
+        if ((word & 0x000000ff) == 0 || (word & 0x0000ff00) == 0 ||
+            (word & 0x00ff0000) == 0 || (word & 0xff000000) == 0) {
+          null_terminated = true;
+        }
+      }
+      if (!null_terminated) {
+        return _.diag(SPV_ERROR_INVALID_BINARY)
+               << "Found a string that is not null-terminated.";
```
dneto0: Need a test case to trigger this.
I recommend creating test strings programmatically.  :-)

source/validate_instruction.cpp
```diff
@@ -131,6 +131,68 @@ spv_result_t CapCheck(ValidationState_t& _,
...
+spv_result_t LimitCheckString(ValidationState_t& _,
+                              const spv_parsed_instruction_t* inst) {
+  // By definition, num_words of an operand is a 16-bit unsigned which will not
+  // exceed the limit of characters for a string literal.
+  // This function checks that a NULL termination must be found before reaching
+  // the end of the string words.
+  for (int i = 0; i < inst->num_operands; ++i) {
+    auto operand = inst->operands[i];
+    if (SPV_OPERAND_TYPE_LITERAL_STRING == operand.type) {
+      bool null_terminated = false;
```
dneto0: Save yourself a loop by leveraging strlen.
There's a null byte if strlen is less than 4 * num words.
```
const size_t width_in_bytes = operand.num_words * sizeof(uint32_t);
const bool null_terminated = std::strnlen_s(reinterpret_cast<const char*>(inst->words + operand.offset)), width_in_bytes) == width_in_bytes;
```

source/val/validation_state.cpp
ehsannas: Good to know. 👍

source/validate_instruction.cpp
```diff
@@ -131,6 +131,68 @@ spv_result_t CapCheck(ValidationState_t& _,
...
+
+// Checks that the Resuld <id> is within the valid bound.
+spv_result_t LimitCheckIdBound(ValidationState_t& _,
+                               const spv_parsed_instruction_t* inst) {
+  for (int i = 0; i < inst->num_operands; ++i) {
+    auto operand = inst->operands[i];
+    if (SPV_OPERAND_TYPE_RESULT_ID == operand.type) {
+      const uint32_t result_id = inst->words[operand.offset];
+      if (result_id <= 0) {
+        return _.diag(SPV_ERROR_INVALID_BINARY)
```
ehsannas: I see. I can remove it then!

source/validate_instruction.cpp
ehsannas: this is clever! :-)

shouldn't it be:
```
const bool null_terminated = std::strnlen_s(reinterpret_cast<const char*>(inst->words + operand.offset)), width_in_bytes) <= width_in_bytes;
```

ehsannas: It seems that the compiler does not like `strlen_s`

`no member named 'strnlen_s' in namespace 'std'`

adding `#include <string.h>` and `#include <string>` did not help.
I also couldn't find the function in cplusplus.com (although it is in cppreference.com). maybe it's only supported my MS?

dneto0: Dang.  I found it in http://en.cppreference.com/w/cpp/string/byte/strlen  
But that says "Since C11". And that's not C++11.  Ugh.

Alternatives are std::find or std::memchr.  I think std::memchr is the better choice.
```
const size_t width_in_bytes = operand.num_words * sizeof(uint32_t);
const bool null_terminated = nullptr != std::memchr(reinterpret_cast<const char*>(inst->words + operand.offset), 0, width_in_bytes);
```

dneto0:
Suggesting std::memchr.

source/validate_instruction.cpp
ehsannas: I intend to remove this function completely. It seems that if a string is not null-terminated, it will be caught before the validator gets to run. I found this out after manipulating the bytes in my unit test:

```
error: 17: Invalid word count: OpName starting at word 13 says it has 3 words, but found 4 words instead.
```
or this:
```
error: 15: End of input reached while decoding OpName starting at word 13: truncated literal string operand at word offset 2.
```

source/validate_instruction.cpp
```diff
@@ -131,6 +131,26 @@ spv_result_t CapCheck(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Checks that the Resuld <id> is within the valid bound.
+spv_result_t LimitCheckIdBound(ValidationState_t& _,
+                               const spv_parsed_instruction_t* inst) {
+  for (int i = 0; i < inst->num_operands; ++i) {
```
dneto0: Sorry I didn't think of this before.  But instruction already has a result_id field parsed out for you.
Either it's 0 (meaning no result Id), or it's the result_id that the binary parser parsed out.
So no need for a loop here at all.  :-)

dneto0:
Yes, a good idea to just check the one thing in this PR. 
Still requesting changes though.  This can be simplified further.

dneto0:
Be sure to update the commit comment to reflect the reduced scope.

source/validate_instruction.cpp
ehsannas: Oh that's right. Thanks 👍

dneto0:
Squashed and pushed into master as bd5b0bfca1bf

PR #466, opened 2016-11-24
Validate limits (Number of Struct members)

ehsannas:
Added validation checks for the number of members in a struct. Also added unit tests for it.

You can only look at the diff for the last 2 commits as the first 2 commits are already included in another pull request.


ehsannas:
This CL will also fail one unit test that actually does fail the ID bound check. The test is called "SemanticsIdIsAnIdNotALiteral"

dneto0:
This is failing the bots.

dneto0:
If I read your comment about the failing test correctly, you'll have to modify that test to avoid the failure unrelated to the intent of that test.  That probably means manually overriding the id bound in the header of the generated binary, after you've assembled the source text.

PR #470, opened 2016-11-28
OpFunctionCall limits

ehsannas:
Adding validation code based on Universal Limits (Section 2.17 SPIR-V
Spec). Number of actual arguments passed to OpFunctionCall may not be
larger than 255.


ehsannas:
Please don't review this one.
There is another Limit rules: "Number of function parameters, per function declaration may not be larger than 255". If this rule is implemented, then the OpFunctionCall will always be valid by definition because the number of actual arguments passed to OpFunctionCall must match the function declaration (via OpTypeFunction).

PR #471, opened 2016-11-28
Validate the number of arguments in OpTypeFunction

ehsannas:
According to the SPIR-V spec (section 2.17: Universal Limits), the
OpTypeFunction instruction may not take more than 255 arguments for the
function. Also added unit tests for it.

### DIFF
--- a/test/val/val_limits_test.cpp
+++ b/test/val/val_limits_test.cpp
@@ -162,3 +162,35 @@ OpFunctionEnd
                         "exceeds the limit (16383)."));
 }
 
+// Valid: OpTypeFunction with 255 arguments.
+TEST_F(ValidateLimits, OpTypeFunctionGood) {
+  int num_args = 255;
+  std::ostringstream spirv;
+  spirv << header << R"(
+%1 = OpTypeInt 32 0
+%2 = OpTypeFunction %1)";
+  // add parameters
+  for (int i = 0; i < num_args; ++i) {
+    spirv << " %1";
+  }
+  CompileSuccessfully(spirv.str());
+  EXPECT_EQ(SPV_SUCCESS, ValidateInstructions());
+}
+
+// Invalid: OpTypeFunction with 256 arguments. (limit is 255 according to the
+// spec Universal Limits (2.17).
+TEST_F(ValidateLimits, OpTypeFunctionBad) {
+  int num_args = 256;
+  std::ostringstream spirv;
+  spirv << header << R"(
+%1 = OpTypeInt 32 0
+%2 = OpTypeFunction %1)";
+  for (int i = 0; i < num_args; ++i) {
+    spirv << " %1";
+  }
+  CompileSuccessfully(spirv.str());
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("OpTypeFunction may not take more than 255 arguments. "
+                        "OpTypeFunction <id> '2' has 256 arguments."));
+}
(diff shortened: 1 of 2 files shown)


umar456:
It was my understanding that the Universal Limits section imposed limits on the *minimum* requirements that implementations and validators should follow. The spec says:

> Validators must either
> • inform when these limits are crossed, or
> • be explicitly parameterized with larger limits.

It seems that the validator is treating these limits as maximum limits rather than minimums.

ehsannas:
My understanding is that these numbers are the __smallest upper limits__. i.e. Unless specified otherwise, these are the upper bounds that must be met.
The code is doing the first bullet: "inform when these limits are crossed"
Does that make sense?

umar456:
Hmm. Maybe @dneto0 can clarify.

From the Universal Limits section:
>These quantities are minimum limits for all implementations and validators. Implementations are allowed to support larger quantities.

If an implementation supports larger limit (let's say > 255 parameter functions) then it cannot use the validator because if minimal limits are crossed the validator will throw an error.

dneto0:
This is a good thing to discuss.

I think we agree that if any module satisfies the limits 2.17 then it can't be rejected by an implementation on the grounds of limits.
The second clause says that if a client can accept modules using beyond-universal-limits, then it should parameterize its calls to the validator with those larger limits.

I'd be ok being in the situation where that extra parameterization is future work.  So far I don't know of any demand for those larger limits. Sure, you could say there is no demand so far because we didn't check it yet.  But it's a reasonable stance to say that the validator should be protecting other parts of the tooling pipeline against invalid modules, and that those other parts could reasonably assume that modules have kept within those universal bounds.

In any case, this merits a TODO at the the validator declaration saying that it only universal bounds are permitted; that the further parameterization of limits is not yet implemented.

test/val/val_limits_test.cpp
```diff
@@ -76,3 +76,33 @@ TEST_F(ValidateLimits, idEqualToBoundBad) {
...
 
+// Valid: OpTypeFunction with 255 arguments.
+TEST_F(ValidateLimits, OpTypeFunctionGood) {
+  int num_args = 255;
+  string spirv = header + R"(
+%1 = OpTypeInt 32 0
+%2 = OpTypeFunction %1)";
+  // add parameters
+  for (int i = 0; i < num_args; ++i) {
+    spirv += " %1";
```
dneto0: Use std::ostringstream.

test/val/val_limits_test.cpp
```diff
@@ -76,3 +76,33 @@ TEST_F(ValidateLimits, idEqualToBoundBad) {
...
+
+// Invalid: OpTypeFunction with 256 arguments. (limit is 255 according to the
+// spec Universal Limits (2.17).
+TEST_F(ValidateLimits, OpTypeFunctionBad) {
+  int num_args = 256;
+  string spirv = header + R"(
+%1 = OpTypeInt 32 0
+%2 = OpTypeFunction %1)";
+  for (int i = 0; i < num_args; ++i) {
+    spirv += " %1";
```
dneto0: Use std::ostringstream

dneto0:
Please rebase against master. I'm getting conflicts. :-)

test/val/val_limits_test.cpp
```diff
@@ -162,3 +162,35 @@ OpFunctionEnd
                         "exceeds the limit (16383)."));
 }
 
+// Valid: OpTypeFunction with 255 arguments.
+TEST_F(ValidateLimits, OpTypeFunctionGood) {
+  int num_args = 255;
```
dneto0: Could be const.  Nit.  Fine to keep.

dneto0:
Squashed and pushed into master as 490617133e

dneto0:
I forgot to resolve @umar456's concern.
Yesterday I committed 5f2d40915e4 which updates the README and CHANGES to describe how limits are treated, and that it's future work to be able to parameterize the validator's limits.

PR #472, opened 2016-11-29
Validation for number of Struct members.

ehsannas:
The number of members in a struct may not exceed 16,383.
Also Adding unit tests for Struct size limit check.

### DIFF
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -147,11 +147,11 @@ spv_result_t LimitCheckStruct(ValidationState_t& _,
                               const spv_parsed_instruction_t* inst) {
   // Number of members is the number of operands of the instruction minus 1.
   // One operand is the result ID.
-  uint16_t limit = 0x3fff;
+  const uint16_t limit = 0x3fff;
   if (SpvOpTypeStruct == inst->opcode && inst->num_operands - 1 > limit) {
     return _.diag(SPV_ERROR_INVALID_BINARY)
            << "Number of OpTypeStruct members (" << inst->num_operands - 1
-           << ") has exceeded the limit (16,383).";
+           << ") has exceeded the limit (" << limit << ").";
   }
   return SPV_SUCCESS;
 }
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -142,6 +142,20 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Checks that the number of OpTypeStruct members is within the limit.
+spv_result_t LimitCheckStruct(ValidationState_t& _,
+                              const spv_parsed_instruction_t* inst) {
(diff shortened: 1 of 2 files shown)


test/val/val_limits_test.cpp
```diff
@@ -76,3 +76,28 @@ TEST_F(ValidateLimits, idEqualToBoundBad) {
       HasSubstr("Result <id> '64' must be less than the ID bound '64'."));
 }
 
+TEST_F(ValidateLimits, structNumMembersGood) {
+  string str = header + R"(
+%1 = OpTypeInt 32 0
+%2 = OpTypeStruct)";
+  for (int i = 0; i < 16383; ++i) {
+    str += " %1";
```
dneto0: This will cause a quadratic number of memory operations.
Instead, use a std::ostringstream to accumulate the values, and then construct the string once with .str()

source/validate_instruction.cpp
```diff
@@ -142,6 +142,20 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Checks that the number of OpTypeStruct members is within the limit.
+spv_result_t LimitCheckStruct(ValidationState_t& _,
+                              const spv_parsed_instruction_t* inst) {
+  // Number of members is the number of operands of the instruction minus 1.
+  // One operand is the result ID.
+  uint16_t limit = 0x3fff;
```
dneto0: const

source/validate_instruction.cpp
```diff
@@ -142,6 +142,20 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
...
+  if (SpvOpTypeStruct == inst->opcode && inst->num_operands - 1 > limit) {
+    return _.diag(SPV_ERROR_INVALID_BINARY)
+           << "Number of OpTypeStruct members (" << inst->num_operands - 1
+           << ") has exceeded the limit (16,383).";
```
dneto0: May as well just use  "<< limit" here, just to be DRY (don't repeat yourself).
Yeah, there's no comma, but at least no locale problem with interpretation of the comma either.

test/val/val_limits_test.cpp
```diff
@@ -76,3 +76,28 @@ TEST_F(ValidateLimits, idEqualToBoundBad) {
...
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_SUCCESS, ValidateInstructions());
+}
+
+TEST_F(ValidateLimits, structNumMembersExceededBad) {
+  string str = header + R"(
+%1 = OpTypeInt 32 0
+%2 = OpTypeStruct)";
+  for (int i = 0; i < 16384; ++i) {
+    str += " %1";
```
dneto0: Here again.

dneto0:
Squashed and pushed into master as bef80716d70a0b

PR #473, opened 2016-11-29
Adding validation code for OpSwitch limits

ehsannas:
The number of (literal, label) pairs passed to OpSwitch may not exceed
16,383. Added code to validate this and added unit tests for it.

Also fixed a typo in another validator error message.

### DIFF
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -152,10 +152,11 @@ spv_result_t LimitCheckSwitch(ValidationState_t& _,
     // literal,label pairs come after the first 2 operands.
     // It is guaranteed at this point that num_operands is an even numner.
     unsigned int num_pairs = (inst->num_operands - 2) / 2;
-    if (num_pairs > 16383) {
+    const unsigned int num_pairs_limit = 16383;
+    if (num_pairs > num_pairs_limit) {
       return _.diag(SPV_ERROR_INVALID_BINARY)
              << "Number of (literal, label) pairs in OpSwitch (" << num_pairs
-             << ") exceeds the limit (16,383).";
+             << ") exceeds the limit (" << num_pairs_limit << ").";
     }
   }
   return SPV_SUCCESS;
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -142,6 +142,25 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Checks that the number of (literal, label) pairs in OpSwitch is within the
+// limit.
+spv_result_t LimitCheckSwitch(ValidationState_t& _,
+                              const spv_parsed_instruction_t* inst) {
+  if (SpvOpSwitch == inst->opcode) {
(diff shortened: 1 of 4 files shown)


source/validate_instruction.cpp
```diff
@@ -142,6 +142,25 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
...
+// limit.
+spv_result_t LimitCheckSwitch(ValidationState_t& _,
+                              const spv_parsed_instruction_t* inst) {
+  if (SpvOpSwitch == inst->opcode) {
+    // The instruction syntax is as follows:
+    // OpSwitch <selector ID> <Default ID> literal label literal label ...
+    // literal,label pairs come after the first 2 operands.
+    // It is guaranteed at this point that num_operands is an even numner.
+    unsigned int num_pairs = (inst->num_operands - 2) / 2;
+    if (num_pairs > 16383) {
```
dneto0: Arguably to don't-repeat-yourself (DRY), make the 16383 a const value, and use it in the error message too.

test/val/val_limits_test.cpp
```diff
@@ -76,3 +76,60 @@ TEST_F(ValidateLimits, idEqualToBoundBad) {
...
+// Valid: Switch statement has 16,383 branches.
+TEST_F(ValidateLimits, switchNumBranchesGood) {
+  string str = header + R"(
+%1 = OpTypeVoid
+%2 = OpTypeFunction %1
+%3 = OpTypeInt 32 0
+%4 = OpConstant %3 1234
+%5 = OpFunction %1 None %2
+%7 = OpLabel
+%8 = OpIAdd %4 %4 %4
```
dneto0: That first %4 should be %3.    This will fail validation once type checking is performed.

test/val/val_limits_test.cpp
```diff
@@ -76,3 +76,60 @@ TEST_F(ValidateLimits, idEqualToBoundBad) {
...
+%9 = OpSwitch %4 %10)";
+
+  // Now add the (literal, label) pairs
+  for (int i = 0; i < 16383; ++i) {
+    str += " 1 %10";
```
dneto0: Use a std::ostringstream to accumulate the value.  To avoid quadratic behaviour.

test/val/val_limits_test.cpp
```diff
@@ -76,3 +76,60 @@ TEST_F(ValidateLimits, idEqualToBoundBad) {
...
+// Invalid: Switch statement has 16,384 branches.
+TEST_F(ValidateLimits, switchNumBranchesBad) {
+  string str = header + R"(
+%1 = OpTypeVoid
+%2 = OpTypeFunction %1
+%3 = OpTypeInt 32 0
+%4 = OpConstant %3 1234
+%5 = OpFunction %1 None %2
+%7 = OpLabel
+%8 = OpIAdd %4 %4 %4
```
dneto0: That first %4 should be %3

dneto0:
Rebased, squashed, and pushed into master as 3c8bc80e3a

PR #474, opened 2016-11-29
Validate the number of global and local variables.

ehsannas:
According to the Universal Limits section of the SPIR-V Spec (2.17), the
number of global variables may not exceed 65,535 and the number of local
variables may not exceed 524,287.

Also added unit tests for each one.

### DIFF
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -176,6 +176,31 @@ spv_result_t LimitCheckSwitch(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Ensure the number of variables of the given class does not exceed the limit.
+spv_result_t LimitCheckNumVars(ValidationState_t& _,
+                               const SpvStorageClass storage_class) {
+  if (SpvStorageClassFunction == storage_class) {
+    _.incrementNumLocalVars();
+    const uint32_t num_local_vars_limit = 0x7FFFF;
+    if (_.num_local_vars() > num_local_vars_limit) {
+      return _.diag(SPV_ERROR_INVALID_BINARY)
+             << "Number of local variables ('Function' Storage Class) "
+                "exceeded the valid limit ("
+             << num_local_vars_limit << ").";
+    }
+  } else {
+    _.incrementNumGlobalVars();
+    const uint32_t num_global_vars_limit = 0xFFFF;
+    if (_.num_global_vars() > num_global_vars_limit) {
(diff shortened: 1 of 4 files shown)


source/validate_instruction.cpp
```diff
@@ -176,6 +176,31 @@ spv_result_t LimitCheckSwitch(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Ensure the number of variables of the given class does not exceed the limit.
+spv_result_t LimitCheckNumVars(ValidationState_t& _,
+                               const SpvStorageClass storage_class) {
+  if (SpvStorageClassFunction == storage_class) {
```
dneto0: This is fine but it somewhat duplicates the logic in the callee function that separates out Function storage class variables from non-Function storage class.

test/val/val_limits_test.cpp
```diff
@@ -162,3 +162,95 @@ OpFunctionEnd
                         "exceeds the limit (16383)."));
 }
 
+// Valid: module has 65,535 global variables.
+TEST_F(ValidateLimits, numGlobalVarsGood) {
+  int num_globals = 65535;
```
dneto0: Could have been const.

source/validate_instruction.cpp
```diff
@@ -176,6 +176,31 @@ spv_result_t LimitCheckSwitch(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Ensure the number of variables of the given class does not exceed the limit.
```
dneto0: Google style is to be "descriptive" rather than "imperative":  So "Ensures" rather than "Ensure".  But this file is not consistently doing this.

https://google.github.io/styleguide/cppguide.html#Function_Comments
"These comments should be descriptive ("Opens the file") rather than imperative ("Open the file"); "

test/val/val_limits_test.cpp
```diff
@@ -162,3 +162,95 @@ OpFunctionEnd
                         "exceeds the limit (16383)."));
 }
 
+// Valid: module has 65,535 global variables.
+TEST_F(ValidateLimits, numGlobalVarsGood) {
```
dneto0: Convention is to name tests starting with upper-case first letter.

dneto0:
Rebased, slightly patched, and pushed into master as  4b26d51

dneto0:
Failing bot will be fixed by a Travis config update in another MR.

PR #476, opened 2016-12-01
Adding validation code for OpAccessChain.

ehsannas:
* Result Type must be an OpTypePointer. Its Type operand must be the
type reached by walking the Base’s type hierarchy down to the last
provided index in Indexes, and its Storage Class operand must be the
same as the Storage Class of Base.

* Base must be a pointer, pointing to the base of a composite object.

* Indexes walk the type hierarchy to the desired depth, potentially down
to scalar granularity. The first index in Indexes will select the
top-level member/element/component/element of the base composite. All
composite constituents use zero-based numbering, as described by their
OpType... instruction. The second index will apply similarly to that
result, and so on. Once any non-composite type is reached, there must
be no remaining (unused) indexes. Each of the Indexes must:
- be a scalar integer type,
- be an OpConstant when indexing into a structure.

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -746,8 +746,7 @@ bool idUsage::isValid<SpvOpSampledImage>(const spv_instruction_t* inst,
   // to OpPhi instructions or OpSelect instructions, or any instructions other
   // than the image lookup and query instructions specified to take an operand
   // whose type is OpTypeSampledImage.
-  std::vector<uint32_t> consumers =
-      module_.getSampledImageConsumers(resultID);
+  std::vector<uint32_t> consumers = module_.getSampledImageConsumers(resultID);
   if (!consumers.empty()) {
     for (auto consumer_id : consumers) {
       auto consumer_instr = module_.FindDef(consumer_id);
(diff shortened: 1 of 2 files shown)


source/validate_id.cpp
```diff
@@ -1177,11 +1176,143 @@ bool idUsage::isValid<SpvOpCopyMemorySized>(const spv_instruction_t* inst,
...
+  auto baseTypeStorageClass = baseTypeInstr->word(2);
+  if (resultTypeStorageClass != baseTypeStorageClass) {
+    DIAG(resultTypeIndex) << "The result pointer storage class and base "
+                             "pointer storage class in OpAccessChain do not "
+                             "match.";
+    return false;
+  }
+
+  // The type pointed to by OpTypePointer (word 3) must be a composite type.
+  auto typePointedTo = module_.FindDef(baseTypeInstr->word(3));
```
ehsannas: 👍 You're right; the check is taken care of in the loop.

source/validate_id.cpp
```diff
@@ -1177,11 +1176,143 @@ bool idUsage::isValid<SpvOpCopyMemorySized>(const spv_instruction_t* inst,
...
-#if 0
 template <>
-bool idUsage::isValid<SpvOpAccessChain>(const spv_instruction_t *inst,
-                                        const spv_opcode_desc opcodeEntry) {}
-#endif
+bool idUsage::isValid<SpvOpAccessChain>(const spv_instruction_t* inst,
+                                        const spv_opcode_desc) {
+  // The result type must be OpTypePointer. Result Type is at word 1.
+  auto resultTypeIndex = 1;
+  auto resultTypeInstr = module_.FindDef(inst->words[resultTypeIndex]);
```
ehsannas: It turns out that here we have access to a `spv_instruction_t*` which does not have a `type_id`.
`spv_parsed_instruction_t*` and `Instruction*` do have the `type_id`.

We could do:
```
Instruction* cur_instr = module_.FindDef(inst->resultTypeId);
```
and then replace `inst->words[resultTypeIndex]` with `cur_instr->type_id()`
But that would be more expensive.

source/validate_id.cpp
```diff
@@ -1237,7 +1232,10 @@ bool idUsage::isValid<SpvOpAccessChain>(const spv_instruction_t* inst,
         << num_indexes_limit << ". Found " << num_indexes << " indexes.";
     return false;
   }
-
+  if (num_indexes <= 0) {
```
dneto0: This works only because we rely on the binary parser to validate that at least we have 4 words in the instruction.
Otherwise this code would be wrong because num_indexes is a size_t and is therefor unsigned.
The safer pattern in general is to check inst->words.size() >= 4.

But this code works, subtlly.

source/validate_id.cpp
dneto0: Right. My mistake.  I was thinking of the parsed instruction struct.

dneto0:
Rebased, squashed, and pushed into master as bab1aad85.

Note that Travis CI Mac OSX buildbots were stalled for ~3 days.  So bypassing that check.

PR #477, opened 2016-12-01
No need to install cmake for new osx images.

ehsannas:
It seems that the osx image has been updated and already includes ctest.
As a result, all jobs fail with this error:
Error: cmake-3.6.2 already installed


dneto0:
Rebased and pushed into master as e6e7bd7d301d

PR #478, opened 2016-12-05
Add a test that exercises wrong usage of result id

ehsannas:
Tests that Result \<id\> resulting from an instruction within a function may not be used outside that function.

### DIFF
--- a/test/val/val_id_test.cpp
+++ b/test/val/val_id_test.cpp
@@ -2787,17 +2787,15 @@ TEST_F(ValidateIdWithMessage, ResultIdUsedOutsideOfFunctionBad) {
 %1 = OpTypeVoid
 %2 = OpTypeFunction %1
 %3 = OpTypeInt 32 0
-%4 = OpConstant %3 1234
-%5 = OpTypePointer Function %3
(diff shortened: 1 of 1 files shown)


test/val/val_id_test.cpp
```diff
@@ -2780,6 +2780,35 @@ TEST_F(ValidateIdWithMessage, OpStoreBitcastNonPointerBad) {
...
 }
 
+// Result <id> resulting from an instruction within a function may not be used
+// outside that function.
+TEST_F(ValidateIdWithMessage, ResultIdUsedOutsideOfFunctionBad) {
+  string spirv = kGLSL450MemoryModel + R"(
+%1 = OpTypeVoid
+%2 = OpTypeFunction %1
+%3 = OpTypeInt 32 0
+%4 = OpConstant %3 1234
```
dneto0: You can shorten this example so that you don't need the %4 (OpConstant) or %9 (OpIAdd).

Change the OpStore at line 2800 into an OpLoad.

dneto0:
Squashed and pushed into master as 1f6123fa5

PR #479, opened 2016-12-05
Fixing a bug in OpAccessChain validation code.

ehsannas:
The validation code for OpAccessChain was missing OpTypeRuntimeArray as
a possible type that can be indexed into.

This was caught by running the validator on VKCTS.

Also adding unit tests for it.


ehsannas:
:+1: Validator now passes on all VKCTS.

dneto0:
Rebased and pushed into master as aadf696fce

PR #480, opened 2016-12-06
Validation for struct nesting depth.

ehsannas:
According to sectin 2.17 in SPIR-V Spec, the structure nesting depth may
not be larger than 255. This is interpreted as structures nested in
structures. The code does not look into arrays or follow pointers to see
if it reaches a structure downstream.

### DIFF
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -16,6 +16,7 @@
 
 #include "validate.h"
 
+#include <algorithm>
 #include <cassert>
 
 #include <sstream>
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -145,13 +145,50 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
 // Checks that the number of OpTypeStruct members is within the limit.
 spv_result_t LimitCheckStruct(ValidationState_t& _,
                               const spv_parsed_instruction_t* inst) {
+  if (SpvOpTypeStruct != inst->opcode) {
+    return SPV_SUCCESS;
+  }
+
   // Number of members is the number of operands of the instruction minus 1.
   // One operand is the result ID.
   const uint16_t limit = 0x3fff;
-  if (SpvOpTypeStruct == inst->opcode && inst->num_operands - 1 > limit) {
+  if (inst->num_operands - 1 > limit) {
     return _.diag(SPV_ERROR_INVALID_BINARY)
            << "Number of OpTypeStruct members (" << inst->num_operands - 1
            << ") has exceeded the limit (" << limit << ").";
+  }
+
+  // Section 2.17 of SPIRV Spec specifies that the "Structure Nesting Depth"
+  // must be less than or equal to 255.
+  // This is interpreted as structures including other structures as members.
+  // The code does not follow pointers or look into arrays to see if we reach a
+  // structure downstream.
+  //
+  // This lambda takes a valid struct ID and finds its nesting depth.
+  // The nesting depth of a struct is 1+(largest depth of any member).
+  // Scalars are at depth 0.
+  std::function<int(uint32_t)> find_struct_depth_recursive = [&](uint32_t id) {
+    int max_member_depth = 0;
+    auto cur_instr = _.FindDef(id);
+    // Struct members start at word 2 of OpTypeStruct instruction.
+    for (size_t word_i = 2; word_i < cur_instr->words().size(); ++word_i) {
+      auto member = cur_instr->word(word_i);
+      auto memberTypeInstr = _.FindDef(member);
+      if (memberTypeInstr && SpvOpTypeStruct == memberTypeInstr->opcode()) {
+        max_member_depth =
+            std::max(max_member_depth,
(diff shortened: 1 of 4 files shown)


source/validate_instruction.cpp
```diff
@@ -145,14 +145,51 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
...
+  std::function<int(uint32_t)> find_struct_depth_recursive = [&](uint32_t id) {
+    int max_member_depth = 0;
+    auto cur_instr = _.FindDef(id);
+    // Struct members start at word 2 of OpTypeStruct instruction.
+    for (size_t word_i = 2; word_i < cur_instr->words().size(); ++word_i) {
+      auto member = cur_instr->word(word_i);
+      auto memberTypeInstr = _.FindDef(member);
+      if (memberTypeInstr && SpvOpTypeStruct == memberTypeInstr->opcode()) {
+        max_member_depth =
+            std::max(max_member_depth,
```
dneto0: Windows bot caught this:  You need to include the &lt;algorithm&gt; header to get std::max

source/validate_instruction.cpp
```diff
@@ -145,14 +146,51 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
...
+  // structure downstream.
+  //
+  // This lambda takes a valid struct ID and finds its nesting depth.
+  // The nesting depth of a struct is 1+(largest depth of any member).
+  // Scalars are at depth 0.
+  std::function<int(uint32_t)> find_struct_depth_recursive = [&](uint32_t id) {
+    int max_member_depth = 0;
+    auto cur_instr = _.FindDef(id);
+    // Struct members start at word 2 of OpTypeStruct instruction.
+    for (size_t word_i = 2; word_i < cur_instr->words().size(); ++word_i) {
```
dneto0: This could (should) be rewritten with the 4 argument form of std::accumulate.

source/validate_instruction.cpp
```diff
@@ -145,14 +146,51 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
...
+  // Section 2.17 of SPIRV Spec specifies that the "Structure Nesting Depth"
+  // must be less than or equal to 255.
+  // This is interpreted as structures including other structures as members.
+  // The code does not follow pointers or look into arrays to see if we reach a
+  // structure downstream.
+  //
+  // This lambda takes a valid struct ID and finds its nesting depth.
+  // The nesting depth of a struct is 1+(largest depth of any member).
+  // Scalars are at depth 0.
+  std::function<int(uint32_t)> find_struct_depth_recursive = [&](uint32_t id) {
```
dneto0: This algorithm suffers from an exponential runtime vulnerability.

Example, 
```struct S0 { int; int };
struct S1 { S0; S0; };
struct S2 { S2; S2; };
...
struct S40 { S39; S39 };
```
To compute the nesting depth of S40, this algorithm will take  about 2**40 steps.

Better to use memoization.  The simple way to do it is to compute and store depths as you scan the module.  This works because the members of a struct are always defined before the struct itself.
(And an OpTypeForwardPointer is just a pointer, which is a leaf anyway.)
This implies an extra member for the validation state.  Probably a std::unordered_map from result ID of struct definitions to depth.  When an ID is not in the map then use the value 0 in the reduction.

source/validate_instruction.cpp
```diff
@@ -145,14 +146,51 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
...
+      if (memberTypeInstr && SpvOpTypeStruct == memberTypeInstr->opcode()) {
+        max_member_depth =
+            std::max(max_member_depth,
+                     find_struct_depth_recursive(memberTypeInstr->id()));
+      }
+    }
+    return 1 + max_member_depth;
+  };
+
+  int depth_limit = 255;
```
dneto0: Use const where possible.

source/validate_instruction.cpp
```diff
@@ -145,14 +146,51 @@ spv_result_t LimitCheckIdBound(ValidationState_t& _,
...
+  int cur_depth = find_struct_depth_recursive(inst->result_id);
```
dneto0: const

source/validate_instruction.cpp
ehsannas: I actually thought of this. I thought to myself that structs are probably not that deep and I was trying to avoid adding more members to the validation state. After reading your example, it's clear that adding an unordered_map to the validation state is the way to go. :+1:

source/validate_instruction.cpp
ehsannas: I don't see how/why to use `std::accumulate`?... I only need to do `1+max{d1,d2,d3,...}` where `di` is the depth of member i.

dneto0: Right.  The idea was to use std::accumulate to do the max for you so you don't write the loop bounds etc.  Some people really like avoiding open-coded loops.   But the std::accumulate method would use a lambda.

I'm not too sticky on this.   Existing code looks ok in this instance.

dneto0:
Rebased, squashed and pushed into master as 4fb79b54f28

PR #481, opened 2016-12-06
Adding validation for OpInBoundsAccessChain.

ehsannas:
The validation for this instruction is similar to OpAccessChain.

### DIFF
--- a/test/val/val_id_test.cpp
+++ b/test/val/val_id_test.cpp
@@ -2245,7 +2245,7 @@ OpFunctionEnd
               HasSubstr("No Indexes were passes to OpInBoundsAccessChain."));
 }
 
-// Valid: 255 indexes passed to OpAccessChain. Limit is 255.
+// Valid: 255 indexes passed to OpInBoundsAccessChain. Limit is 255.
 TEST_F(ValidateIdWithMessage, OpInBoundsAccessChainTooManyIndexesGood) {
   int depth = 255;
   std::string header = kGLSL450MemoryModel + opAccessChainSpirvSetup;
--- a/test/val/val_id_test.cpp
+++ b/test/val/val_id_test.cpp
@@ -2103,7 +2103,271 @@ OpFunctionEnd
                         "base <id> (OpTypeFloat)."));
 }
 
-// TODO: OpInBoundsAccessChain
+// Valid: Access a float in a matrix using OpInBoundsAccessChain
+TEST_F(ValidateIdWithMessage, OpInBoundsAccessChainGood) {
+  string spirv = kGLSL450MemoryModel + opAccessChainSpirvSetup + R"(
+%float_entry = OpInBoundsAccessChain %_ptr_Private_float %my_matrix %int_0 %int_1
+OpReturn
+OpFunctionEnd
+  )";
+  CompileSuccessfully(spirv);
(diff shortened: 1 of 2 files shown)


test/val/val_id_test.cpp
```diff
@@ -2103,7 +2103,271 @@ OpFunctionEnd
                         "base <id> (OpTypeFloat)."));
 }
 
-// TODO: OpInBoundsAccessChain
+// Valid: Access a float in a matrix using OpInBoundsAccessChain
```
dneto0: There are two schools of thought.
1. Tests should be dead simple straight line code.  This is the purist party line in Google.  It can lead to lots of duplication.  But within reason that duplication is worth price in test stimplicity.
2. Parameterize if you can.    This has the cost of extra complexity in the generic parts of the tests.

We've adopted the parameterize strategy in many of the tests in SPIRV-Tools.   There are reasonable arguments both ways.   This wholesale duplication is acceptable, but we're going to get another 2-fold increase in the next while for the Ptr* variants of access chains.  It's a bit ugly.

Probably not worth the refactor now.  But for next set of tests (not the AccessChain ones), let's try to use test parameterization.

test/val/val_id_test.cpp
```diff
@@ -2103,7 +2103,271 @@ OpFunctionEnd
...
+OpFunctionEnd
+  )";
+  CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("No Indexes were passes to OpInBoundsAccessChain."));
+}
+
+// Invalid: 256 indexes passed to OpInBoundsAccessChain. Limit is 255.
+TEST_F(ValidateIdWithMessage, OpInBoundsAccessChainTooManyIndecesBad) {
```
dneto0: Let's use "Indexes" or "Indices" consistently.  "Indeces" is not a word. :-)

Likely affects the existing tests too.

test/val/val_id_test.cpp
```diff
@@ -2103,7 +2103,271 @@ OpFunctionEnd
...
+OpReturn
+OpFunctionEnd
+  )";
+  CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("No Indexes were passes to OpInBoundsAccessChain."));
+}
+
+// Invalid: 256 indexes passed to OpInBoundsAccessChain. Limit is 255.
```
dneto0: We should also exercise the use of 255 indices, to prove that that works.

Likely affects the existing tests for OpAccessChain.

ehsannas:
Also clang-formatted.

test/val/val_id_test.cpp
```diff
@@ -2198,8 +2245,48 @@ OpFunctionEnd
               HasSubstr("No Indexes were passes to OpInBoundsAccessChain."));
 }
 
+// Valid: 255 indexes passed to OpAccessChain. Limit is 255.
```
dneto0: OpInBoundsAccessChain

dneto0:
Squashed and pushed into master as d004cd46d

Issue #482, opened 2016-12-07
spirv-val still complains about using OpVariable initializers


aarongreig:
The following was added in revision five of the spec:
> Initializer must be an `<id>` from a constant instruction or a global (module scope) OpVariable instruction.

But trying something like
```
%10 = OpTypeInt 32 1
%11 = OpTypePointer Private %10
 %3 = OpVariable %11 Private
 %4 = OpVariable %11 Private %3
```
gives me
> error: 90: OpVariable Initializer <id> '3' is not a constant.

from spirv-val

dneto0:
Whoops. Agreed that's a problem.

PR #483, opened 2016-12-07
An OpVariable initializer can be a module-scope variable

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/482

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -988,7 +988,7 @@ bool idUsage::isValid<SpvOpSpecConstantOp>(const spv_instruction_t *inst) {}
 
 template <>
 bool idUsage::isValid<SpvOpVariable>(const spv_instruction_t* inst,
-                                     const spv_opcode_desc opcodeEntry) {
+                                     const spv_opcode_desc) {
   auto resultTypeIndex = 1;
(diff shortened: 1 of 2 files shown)


source/validate_id.cpp
```diff
@@ -997,13 +997,19 @@ bool idUsage::isValid<SpvOpVariable>(const spv_instruction_t* inst,
                           << "' is not a pointer type.";
     return false;
   }
-  if (opcodeEntry->numTypes < inst->words.size()) {
-    auto initialiserIndex = 4;
-    auto initialiser = module_.FindDef(inst->words[initialiserIndex]);
-    if (!initialiser || !spvOpcodeIsConstant(initialiser->opcode())) {
+  const auto initialiserIndex = 4;
```
ehsannas: Since you are removing the only usage of `opcodeEntry` parameter that's passed to the function, you need to update the function definition header to:
```
bool idUsage::isValid<SpvOpVariable>(const spv_instruction_t* inst, const spv_opcode_desc) {
```

dneto0: Ah. Submitted a fix.  I'll keep the LGTM if the bots go green.

dneto0:
Squashed.  Keeping the LGTM

PR #505, opened 2016-12-09
Validation code for Control Flow Nesting depth.

ehsannas:
According to Universal Limits section (2.17) of SPIRV Spec, the
control-flow nesting depth may not exceed 1023.

We can check this by utilizing the existing DepthFirstTraversal code
that operates on the control-flow graph.

Also added 2 unit tests.

### DIFF
--- a/source/validate_cfg.cpp
+++ b/source/validate_cfg.cpp
@@ -456,9 +456,8 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
               block_depth[merge_block] = depth;
               block_depth[continue_block] = depth + 1;
               ++depth;
-            }
-            // Header block of a selection.
-            else if (b->is_type(kBlockTypeHeader)) {
+            } else if (b->is_type(kBlockTypeHeader)) {
+              // Header block of a selection.
               const auto& construct = function.FindConstructForEntryBlock(b);
               const auto& merge_block = construct.exit_block();
               block_depth[b] = depth;
--- a/source/validate_cfg.cpp
+++ b/source/validate_cfg.cpp
@@ -403,7 +403,7 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
     auto ignore_block = [](cbb_ptr) {};
     auto ignore_edge = [](cbb_ptr, cbb_ptr) {};
     if (!function.ordered_blocks().empty()) {
-      /// calculate dominators
+      /// Calculate dominators
       DepthFirstTraversal(
           function.first_block(), function.AugmentedCFGSuccessorsFunction(),
           ignore_block, [&](cbb_ptr b) { postorder.push_back(b); },
@@ -414,7 +414,7 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
         edge.first->SetImmediateDominator(edge.second);
       }
 
-      /// calculate post dominators
+      /// Calculate post dominators
       DepthFirstTraversal(
           function.pseudo_exit_block(),
           function.AugmentedCFGPredecessorsFunction(), ignore_block,
@@ -424,7 +424,7 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
       for (auto edge : postdom_edges) {
         edge.first->SetImmediatePostDominator(edge.second);
       }
-      /// calculate back edges.
+      /// Calculate back edges.
       DepthFirstTraversal(
           function.pseudo_entry_block(),
           function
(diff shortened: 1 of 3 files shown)


source/validate_cfg.cpp
```diff
@@ -424,14 +424,63 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
...
+      /// Calculate back edges.
       DepthFirstTraversal(
           function.pseudo_entry_block(),
           function
               .AugmentedCFGSuccessorsFunctionIncludingHeaderToContinueEdge(),
           ignore_block, ignore_block, [&](cbb_ptr from, cbb_ptr to) {
             back_edges.emplace_back(from->id(), to->id());
           });
+
+      // Perform a DepthFirstTraversal that ensures CFG nesting depth is valid.
```
dneto0: This is a big enough chunk of functionality that I'd split this out into its own function.  It can assume the function has a body.

test/val/val_limits_test.cpp
```diff
@@ -22,9 +22,10 @@
 #include "unit_spirv.h"
 #include "val_fixtures.h"
 
+namespace {
```
dneto0: Good. Thanks.

source/validate_cfg.cpp
```diff
@@ -424,14 +424,63 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
...
+      const int cfg_depth_limit = 1023;
+      bool cfg_depth_limit_violated = false;
+      int depth = 0;
+      std::unordered_map<cbb_ptr, int> block_depth;
+      DepthFirstTraversal(
+          function.first_block(), function.AugmentedCFGSuccessorsFunction(),
+          [&](cbb_ptr b) {
```
dneto0: The style guide tells us to avoid overly-broad lambda captures.  In this case I think you only need four new variables you added here.

(Yes, I see there are other cases in this code. I was cutting some slack earlier.)
The situation would be better if you pulled this section of functionality into its own function. Then there's less surrounding state to capture in the first place.

source/validate_cfg.cpp
dneto0: You should explain in words how you determine depth of a block:
- It's 0 for the function entry block
- the non-header blocks in a loop or selection construct (that aren't in yet a deeper nested construct) have 1 more than the header
- the blocks in a continue construct (that aren't in a deeper nested construct) have 1 more than the header

test/val/val_limits_test.cpp
```diff
@@ -320,3 +321,71 @@ TEST_F(ValidateLimits, StructNestingDepthBad) {
       HasSubstr(
           "Structure Nesting Depth may not be larger than 255. Found 256."));
 }
+
+// clang-format off
+// Generates an SPIRV program with the given control flow nesting depth
+void GenerateSpirvProgramWithCfgNestingDepth(std::string& str, int depth) {
```
dneto0: It's worth noting that the nesting is along a single "cone" or spine.  There aren no other nested constructs elsewhere in the function.

dneto0:
Hmmmm.  I'm reconsidering.

Here's an example to think about:

```
void foo() {
   for (a=0; 
          a<100;  // depth 0
          a++       // depth 1 because this is the continue construct
          ) {
      //depth 1
      if(b) {
         // depth 2
         if (c) {
            // depth 3
            d = 1;
            continue;  // this branches out directly to the block containing a++ up at the top.
         }
      }
   }
}
```

But now I look at it again, that's the magic of marking structured control flow:

* The continue target has its depth marked *when we visit its corresponding loop header*.
* And the loop header *dominates* all the blocks inside the loop, and its header, by structured control flow rules.
* Similarly for a selection construct.

At the very least the algorithm needs more justification in its comments.  I'm going to think more about correctness.

source/validate_cfg.cpp
```diff
@@ -424,14 +424,63 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
...
+            } else if (b->is_type(kBlockTypeHeader)) {
+              // Header block of a selection.
+              const auto& construct = function.FindConstructForEntryBlock(b);
+              const auto& merge_block = construct.exit_block();
+              block_depth[b] = depth;
+              block_depth[merge_block] = depth;
+              ++depth;
+            } else if (block_depth.find(b) == block_depth.end()) {
+              block_depth[b] = depth;
+            }
```
dneto0: Let's say the current block b is a ContinueTarget that's supposed to be depth 1, and we got here from a very deeply nested block (e.g. depth 10) that executed a 'continue' branch out here.  Then the "depth" variable will still be 10, and we would end up setting the depth of the remaining blocks in the continue construct (aside from the continue target) to 10 or more.  That's a bug, I think.
It shows that we need something more sophisticated than depth first search with a single depth variable.

Does a stack of depth values work?  I think probably not.  In the same scenario we don't want to pop off all those deeper values from the stack when we hit the continue target  from a deep escaping branch.

ehsannas: No apology needed. I'm glad 

> we would end up setting the depth of the remaining blocks in the continue construct (aside from the continue target) to 10 or more. That's a bug, I think.

You're right. There's a subtle bug here: I hadn't thought about the fact that continue blocks could be multiple levels deep. The depth of deeper levels of the continue block would be calculated incorrectly here.

ehsannas: I'm thinking that fixing this bug might give us the correct solution. But I'll try to think of counter examples.

ehsannas:
As a result of discussions that came up after this pull request, I have adopted another (simpler) methodology for finding the control flow nesting depth which is implemented in a different PR.

PR #507, opened 2016-12-16
Generate spvasm.vim

dneto0:
Generate a vim syntax file for SPIR-V assembly.
Copy the resulting spvasm.vim into your $HOME/.vim/syntax directory
to get syntax highlighting in Vim.


dneto0:
This PR adds the generate-the-syntax-file target to the ALL target.  I'm having second thoughts and would like to make it a non-default target.

utils/generate_vim_syntax.py
```diff
@@ -0,0 +1,184 @@
...
+
+" The assembler treats the leading minus sign as part of the number token.
+syn match   spvasmNumber /-\?\<\d\+\>/
+syn match   spvasmFloat  /-\?\<\d\+\.\d*\(e[+-]\d\+\)\?\>/
+
+" Hex floats.
+" In general, C++ requires at least one hex digit, and the decimal
+" point is optional.
+" First case: hex digits before the optional decimal, no trailing hex digits.
+syn match   spvasmFloat  /-\?0[xX]\\x\+.*p[-+]\d\+\>/
```
antiagainst: The POWER of regex. I didn't check this carefully.

utils/generate_vim_syntax.py
```diff
@@ -0,0 +1,184 @@
...
+
+POSTAMBLE="""
+
+syntax keyword spvasmTodo TODO FIXME contained
+
+syn match   spvasmIdNumber /%\d\+\>/
+
+" The assembler treats the leading minus sign as part of the number token.
+syn match   spvasmNumber /-\?\<\d\+\>/
+syn match   spvasmFloat  /-\?\<\d\+\.\d*\(e[+-]\d\+\)\?\>/
```
antiagainst: Also covered by the following comment?

utils/generate_vim_syntax.py
```diff
@@ -0,0 +1,184 @@
...
+  HiLink spvasmTodo Todo
+
+  delcommand HiLink
+endif
+
+let b:current_syntax = "spvasm"
+"""
+
+# This list is taken from the description of OpSpecConstantOp in SPIR-V 1.1.
+SPEC_CONSTANT_OP_OPCODES = """
```
antiagainst: Maybe we should extend the grammar file with a field for each opcode to indicate whether it's possible to be used for specConstantOp.

I generally think the more information we put into the grammar file, the more useful it will be.

dneto0: I added a TODO for this suggestion.

utils/generate_vim_syntax.py
dneto0: Next commit fixed it.

dneto0:
Removed the \< and \> from the hex float rules. They were causing `-0X1a.p+12` to not match.

Squashed.

Keeping the LGTM.

Issue #508, opened 2016-12-19
CYGWIN support


Ethatron:
It's trivial to support. Add this to CMakeLists.txt:

elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "CYGWIN")
  add_definitions(-DSPIRV_WINDOWS)

And then in "source/binary.cpp", kill the exotic strnlen:

      const size_t string_num_content_bytes =
          std::min(strlen(string), remaining_input_bytes);

dneto0:
What is your configuration?
- what Cmake generator?  Unix makefiles?
- what compiler?  GCC 5.4.0 installed from within Cygwin?

dneto0:
That replacement for strnlen is not safe in the way I want it to be.  Given malicious input, it could cause a buffer overrun.  I'll investigate an alternative.

dneto0:
Hi @Ethatron: Are you using googletest under Cygwin?  I have not been able to make googletest work for me under Cygwin with GCC.  It won't pass its own self-tests, and it causes at least one crash in the tests for SPIRV-Tools.
See #510 for details.

Let me know if you have more success than I do.

Ethatron:
@dneto0 The configuration was all "lastest" from cygwin: cmake from cygwin and latest gcc
I did only go for the "dis" command, and it worked AFAIR.

PR #509, opened 2016-12-19
Support building under CYGWIN

dneto0:
- Parse CHANGES file with Universal Python line endings in case
  the source tree was checked out with Windows line endings.
- Use our own clone of strnlen_s which might not be available
  everywhere.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/508


antiagainst:
LGTM, just need to let the bot happy.

If we are gonna support CYGWIN, would be nice to add appeyvor config for it.

dneto0:
Added a commit to fix the signed/unsigned compare issue.  I'll keep the LGTM otherwise.

I added https://github.com/KhronosGroup/SPIRV-Tools/projects/2#card-1161778 for bot support.

dneto0:
Rebased

dneto0:
Squashed

dneto0:
Rebased

PR #511, opened 2016-12-22
Validation code for control flow nesting depth.

ehsannas:
According to Section 2.17 (Universal Limits) of the SPIR-V Spec, the
control flow nesting depth may not be larger than 1023.

### DIFF
--- a/source/val/function.cpp
+++ b/source/val/function.cpp
@@ -374,50 +374,52 @@ Construct& Function::FindConstructForEntryBlock(const BasicBlock* entry_block) {
 }
 
 int Function::GetBlockDepth(BasicBlock* bb) {
+  // Guard against nullptr.
   if (!bb) {
     return 0;
   }
   // Only calculate the depth if it's not already calculated.
   // This function uses memoization to avoid duplicate CFG depth calculations.
-  if (block_depth_.find(bb) == block_depth_.end()) {
-    if (bb->is_type(kBlockTypeMerge)) {
-      // If this is a merge block, its depth is equal to the block before
-      // branching.
-      BasicBlock* header = merge_block_header_[bb];
-      assert(header);
-      block_depth_[bb] = GetBlockDepth(header);
-    } else if (bb->is_type(kBlockTypeContinue)) {
-      // The depth of the continue block entry point is 1 + loop header depth.
-      Construct* continue_construct = entry_block_to_construct_[bb];
-      assert(continue_construct);
-      // Continue construct has only 1 corresponding construct (loop header).
-      Construct* loop_construct =
-          continue_construct->corresponding_constructs()[0];
-      assert(loop_construct);
-      BasicBlock* loop_header = loop_construct->entry_block();
-      // The continue target may be the loop itself (while 1).
-      // In such cases, the depth of the continue block is: 1 + depth of the
-      // loop's dominator block.
-      if (loop_header == bb) {
-        block_depth_[bb] = 1 + GetBlockDepth(bb->immediate_dominator());
-      } else {
-        block_depth_[bb] = 1 + GetBlockDepth(loop_header);
-      }
-    } else {
-      BasicBlock* bb_dom = bb->immediate_dominator();
-      if (!bb_dom || bb_dom == bb) {
-        // This block has no dominator, so it's at depth 0.
-        block_depth_[bb] = 0;
-      } else if (bb_dom->is_type(kBlockTypeHeader) ||
-                 bb_dom->is_type(kBlockTypeLoop)) {
-        // The dominator of the given block is a header block. So, the nesting
-        // depth of this block is: 1 + nesting depth of the header.
-        block_depth_[bb] = 1 + GetBlockDepth(bb_dom);
-      } else {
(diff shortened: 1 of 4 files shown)


ehsannas:
There seems to be a segmentation fault in the bots which I'm not getting on my machine. I'll investigate.

source/val/function.h
```diff
@@ -174,6 +174,9 @@ class Function {
   /// Returns the block predecessors function for the augmented CFG.
   GetBlocksFunction AugmentedCFGPredecessorsFunction() const;
 
+  /// Returns the control flow nesting depth of the given basic block.
+  int GetBlockDepth(BasicBlock* bb);
```
dneto0: You must say this should only be called after dominators have been computed.

It's discretionary whether you should define exactly what you mean by nesting depth.  Either document it here, or in the function.

dneto0: I see there's a reasonable set of definitions there. That will do for those who are curious.

test/val/val_limits_test.cpp
```diff
@@ -320,3 +321,71 @@ TEST_F(ValidateLimits, StructNestingDepthBad) {
       HasSubstr(
           "Structure Nesting Depth may not be larger than 255. Found 256."));
 }
+
+// clang-format off
```
dneto0: Turning off formatting for a whole function seems ... excessive.  What are you really trying to protect?

source/val/function.cpp
```diff
@@ -371,4 +373,45 @@ Construct& Function::FindConstructForEntryBlock(const BasicBlock* entry_block) {
...
+      assert(continue_construct);
+      // Continue construct has only 1 corresponding construct (loop header).
+      Construct* loop_construct =
+          continue_construct->corresponding_constructs()[0];
+      assert(loop_construct);
+      BasicBlock* loop_header = loop_construct->entry_block();
+      block_depth_[bb] = 1 + GetBlockDepth(loop_header);
+    } else {
+      BasicBlock* bb_dom = bb->immediate_dominator();
+      if (!bb_dom || bb_dom == bb) {
```
dneto0: I believe bb_dom can never be null.  So some might get confused about the !bb_dom condition.  But you could argue it's defensive.
Also, perhaps the doc for BasicBlock::immediate_dominator should say its result is never null. 

ok to keep.

source/val/function.h
dneto0: Should also say that control flow constructs should have been identified.

dneto0: Hmm.  Maybe the algorithm only works if you know you have structured control flow.  So I suggest you guard the check by whether you have Shader capability.

test/val/val_limits_test.cpp
```diff
@@ -320,3 +321,71 @@ TEST_F(ValidateLimits, StructNestingDepthBad) {
...
+// Generates an SPIRV program with the given control flow nesting depth
+void GenerateSpirvProgramWithCfgNestingDepth(std::string& str, int depth) {
```
dneto0: This test is good as far as it goes.
It doesn't capture the issue of having a complex the old algorithm incorrectly calculating the depths of the nodes in a complex continue target that can be reached from an inner construct.  However, making that test case won't be very clear if you only check for the overall validator error message.

The validator has this issue in general: it really begs for exposing its internals so they can be better unit tested.  This is true for the dominance calculations, and now for the block depth calculations.

I made https://github.com/KhronosGroup/SPIRV-Tools/projects/1#card-1185526 and https://github.com/KhronosGroup/SPIRV-Tools/projects/1#card-1185539 for that refactoring-and-testing work.

source/val/function.cpp
```diff
@@ -395,7 +395,14 @@ int Function::GetBlockDepth(BasicBlock* bb) {
           continue_construct->corresponding_constructs()[0];
       assert(loop_construct);
       BasicBlock* loop_header = loop_construct->entry_block();
-      block_depth_[bb] = 1 + GetBlockDepth(loop_header);
+      // The continue target may be the loop itself (while 1).
+      // In such cases, the depth of the continue block is: 1 + depth of the
+      // loop's dominator block.
+      if (loop_header == bb) {
+        block_depth_[bb] = 1 + GetBlockDepth(bb->immediate_dominator());
```
dneto0: What if bb has no predecssor?  Then you'll end up with an infinite loop.

This says to me that the !bb_dom || bb_dom == bb case should be taken care of *first*, at line 382 or so.

test/val/val_limits_test.cpp
ehsannas: The clang formatting pushed every "<<" of the ostringstream to a new line which made it difficult to see the spir-v text I was trying to build. 

Keeping clang-format off allows us to see what string goes on each line of the spir-v code. I thought it might be more readable this way. The entire function is just building the spir-v code string.

source/val/function.cpp
dneto0: In fact you should add a test case for that.  It's the same as the other self-loop  case, but where there is no branch from the entry point to the self-loop.

dneto0:
Squashed and pushed into master as 6993fc413d

PR #514, opened 2017-01-03
Refactoring unit tests

ehsannas:
Using parameterized unit tests to avoid duplicate code that runs the
tests of OpAccessChain and OpInBoundsAccessChain.

This is also a steppingstone to adding tests for OpPtrAccessChain and
OpInBoundsPtrAccessChain.


dneto0:
+2

PR #515, opened 2017-01-03
Adding validation code for PtrAccessChain.

ehsannas:
Validation for OpPtrAccessChain is similar to OpAccessChain with the
following difference: OpPtrAccessChain takes an extra argument (word 4)
which is the Element <id> argument.

Validation for OpInBoundsPtrAccessChain is also similar to OpPtrAccessChain.

Also added tests for all access chain instructions:
Modified the existing parameterized tests to accommodate OpPtrAccessChain and
OpInBoundsPtrAccessChain.

Also fixed a typo in previous commits.

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -1251,7 +1251,7 @@ bool idUsage::isValid<SpvOpAccessChain>(const spv_instruction_t* inst,
     return false;
   }
   if (num_indexes <= 0) {
-    DIAG(resultTypeIndex) << "No Indexes were passes to " << instr_name << ".";
+    DIAG(resultTypeIndex) << "No Indexes were passed to " << instr_name << ".";
     return false;
   }
   // Indexes walk the type hierarchy to the desired depth, potentially down to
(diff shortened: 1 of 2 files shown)


source/validate_id.cpp
```diff
@@ -1344,6 +1344,28 @@ bool idUsage::isValid<SpvOpInBoundsAccessChain>(
...
+                                           const spv_opcode_desc opcodeEntry) {
+  // OpPtrAccessChain's validation rules are similar to OpAccessChain, with one
+  // difference: word 4 must be id of an integer (Element <id>).
+  // The grammar guarantees that there are at least 5 words in the instruction
+  // (i.e. if there are fewer than 5 words, the SPIR-V code will not compile.)
+  int elem_index = 4;
+  // We can remove the Element <id> from the instruction words, and simply call
+  // the validation code of OpAccessChain.
+  spv_instruction_t new_inst = *inst;
+  new_inst.words.erase(new_inst.words.begin() + elem_index);
```
dneto0: At first I thought it was dodgy to reuse validation for a different instruction.  But it's ok because we're only doing validation of Ids in this class.  The instruction word does not matter.

PR #518, opened 2017-01-04
Validation for Composite Extract and Insert.

ehsannas:
The validity of each command is checked based on the description in
SPIR-V Spec Section 3.32.12 (Composite Instructions).

Also checked that the number of indexes passed to these commands does
not exceed the limit described in 2.17 (Universal Limits).

Also added unit tests for each one.

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -1630,7 +1630,7 @@ bool walkCompositeTypeHierarchy(
     std::vector<uint32_t>::const_iterator word_iter_end,
     const libspirv::Instruction* base,
     const libspirv::Instruction** result_type_instr,
-    std::function<std::string(void)> instr_name, std::ostringstream* error) {
+    std::function<std::string(void)> instr_name, std::ostream* error) {
   auto cur_type = base;
   for (; word_iter != word_iter_end; ++word_iter) {
     switch (cur_type->opcode()) {
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -1612,17 +1612,203 @@ bool idUsage::isValid<OpCompositeConstruct>(
     const spv_instruction_t *inst, const spv_opcode_desc opcodeEntry) {}
 #endif
 
-#if 0
 template <>
-bool idUsage::isValid<OpCompositeExtract>(
-    const spv_instruction_t *inst, const spv_opcode_desc opcodeEntry) {}
-#endif
+bool idUsage::isValid<SpvOpCompositeExtract>(
+    const spv_instruction_t *inst, const spv_opcode_desc) {
+  std::string instr_name =
+      "Op" + std::string(spvOpcodeString(static_cast<SpvOp>(inst->opcode)));
 
-#if 0
+  // Remember the result type. Result Type is at word 1.
+  // This will be used to make sure the indexing results in the same type.
+  auto resultTypeIndex = 1;
+  auto resultTypeInstr = module_.FindDef(inst->words[resultTypeIndex]);
+
+  // The Composite <id> is at word 3. ID definition checks ensure this id is
+  // already defined.
+  auto baseIdIndex = 3;
+  auto baseInstr = module_.FindDef(inst->words[baseIdIndex]);
+  auto curTypeInstr = module_.FindDef(baseInstr->type_id());
+
+  // Check Universal Limit (SPIR-V Spec. Section 2.17).
+  // The number of indexes passed to OpCompositeExtract may not exceed 255.
+  // The instruction includes 4 words + N words (for N indexes)
+  const size_t num_indexes = inst->words.size() - 4;
+  const size_t num_indexes_limit = 255;
+  if (num_indexes > num_indexes_limit) {
+    DIAG(resultTypeIndex) << "The number of indexes in " << instr_name
+                          << " may not exceed " << num_indexes_limit
+                          << ". Found " << num_indexes << " indexes.";
+    return false;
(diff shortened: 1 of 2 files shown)


source/validate_id.cpp
```diff
@@ -1612,17 +1612,202 @@ bool idUsage::isValid<OpCompositeConstruct>(
...
 #endif
 
-#if 0
 template <>
-bool idUsage::isValid<OpCompositeExtract>(
-    const spv_instruction_t *inst, const spv_opcode_desc opcodeEntry) {}
-#endif
+bool idUsage::isValid<SpvOpCompositeExtract>(const spv_instruction_t* inst,
+                                             const spv_opcode_desc) {
+  std::string instr_name =
```
dneto0: This does a bunch of allocations and a (slow for now) lookup.  But the common case is there is no error, so it's wasteful.
I'd rather make a lambda here that you then call when needed.

source/validate_id.cpp
```diff
@@ -1612,17 +1612,202 @@ bool idUsage::isValid<OpCompositeConstruct>(
...
+    DIAG(resultTypeIndex) << "The number of indexes in " << instr_name
+                          << " may not exceed " << num_indexes_limit
+                          << ". Found " << num_indexes << " indexes.";
+    return false;
+  }
+  if (num_indexes <= 0) {
+    DIAG(resultTypeIndex) << "No Indexes were passed to " << instr_name << ".";
+    return false;
+  }
+  // Indexes walk the type hierarchy to the desired depth, potentially down to
```
dneto0: Don't duplicate this code. Refactor it.  (Be more "DRY").

At first I thought there would be enough to share with OpAccessChain type walking code as well, but probably not since the composite insert/extract instructions need literal values.  So let's not share code that far.

test/val/val_id_test.cpp
```diff
@@ -2531,6 +2531,327 @@ OpFunctionEnd)";
                         "as an operand of <id> '24'."));
 }
 
+// Valid: Get a float in a matrix using CompositeExtract.
+// Valid: Insert float into a matrix using CompositeInsert.
+TEST_F(ValidateIdWithMessage, CompositeExtractInsertGood) {
+  ostringstream spirv;
+  spirv << kGLSL450MemoryModel << opAccessChainSpirvSetup << std::endl;
```
dneto0: opAccessChainSpirvSetup   should only be for AccessChain  tests.  ?

How about a rename to kDeeplyNestedStructureSetup.
Also, as per style guide, you should not have static duration non-POD values.  Make it a char[] instead.

test/val/val_id_test.cpp
```diff
@@ -2531,6 +2531,327 @@ OpFunctionEnd)";
...
+  std::ostringstream spirv;
+  spirv << header << std::endl;
+
+  // Build nested structures. Struct 'i' contains struct 'i-1'
+  spirv << "%s_depth_1 = OpTypeStruct %float\n";
+  for (int i = 2; i <= depth; ++i) {
+    spirv << "%s_depth_" << i << " = OpTypeStruct %s_depth_" << i - 1 << "\n";
+  }
+
+  // Define Pointer and Variable to use for the AccessChain instruction.
```
dneto0: AccessChain.... comment out of date.

test/val/val_id_test.cpp
ehsannas: Good idea 👍

source/validate_id.cpp
```diff
@@ -1612,11 +1612,77 @@ bool idUsage::isValid<OpCompositeConstruct>(
...
+// (unused) indexes.
+// Returns true on success and false otherwise.
+// If successful, the final type reached by indexing is returned by reference.
+// If an error occurs, the error string is returned by reference.
+bool walkCompositeTypeHierarchy(
+    const ValidationState_t& module,
+    std::vector<uint32_t>::const_iterator word_iter,
+    std::vector<uint32_t>::const_iterator word_iter_end,
+    const libspirv::Instruction* base,
+    const libspirv::Instruction*& result_type_instr,
```
dneto0: General thing: Style guide says don't use non-const ref arguments.  At the call site it's hard (impossible) to see what's modifiable inside the called function.  Use a pointer argument instead.
https://google.github.io/styleguide/cppguide.html#Reference_Arguments

source/validate_id.cpp
```diff
@@ -1612,11 +1612,77 @@ bool idUsage::isValid<OpCompositeConstruct>(
...
+    std::function<std::string(void)> instr_name, std::stringstream& error) {
```
dneto0: std::ostream is more general, and preferable here.

source/validate_id.cpp
```diff
@@ -1629,18 +1629,18 @@ bool walkCompositeTypeHierarchy(
     std::vector<uint32_t>::const_iterator word_iter,
     std::vector<uint32_t>::const_iterator word_iter_end,
     const libspirv::Instruction* base,
-    const libspirv::Instruction*& result_type_instr,
-    std::function<std::string(void)> instr_name, std::stringstream& error) {
-  result_type_instr = base;
+    const libspirv::Instruction** result_type_instr,
+    std::function<std::string(void)> instr_name, std::ostringstream* error) {
```
dneto0: std::ostream is the more basic type to use in the callee.  In the callee method your'e only shifting text onto it.

dneto0:
Rebased, squashed, and pushed into master as d1868b9

Issue #517, opened 2017-01-04
Vulkan CTS test gets assert in Function::GetBlockDepth


Tony-LunarG:
Vulkan CTS test dEQP-VK.spirv_assembly.instruction.graphics.selection_block_order.out_of_order_vert gets a "vector subscript out of range" assert when run with validation layers on Windows.  The stack trace points to PerformCfgChecks calling Function::GetBlockDepth then at line 402 continue_construct->corresponding_constructs is empty and the vector code asserts.  This started happening with commit 6993fc4 "Validation code for control flow nesting depth".

ehsannas:
I'll take a look tonight or tomorrow.

ehsannas:
![1](https://cloud.githubusercontent.com/assets/3828215/21686677/3ec1de36-d334-11e6-853d-a1162b1868e1.png)

This is a representation of the control flow graph in of the failing test. 

The issue is caused because `74` is both the entry block of the continue construct and also the entry block of the selection construct.

Here's the SPIRV code for this test:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos SPIR-V Tools Assembler; 0
; Bound: 84
; Schema: 0
               OpCapability Shader
               OpCapability ClipDistance
               OpCapability CullDistance
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main" %_ %position %vtxColor %color %gl_VertexIndex %gl_InstanceIndex
               OpName %main "main"
               OpName %gl_PerVertex "gl_PerVertex"
               OpMemberName %gl_PerVertex 0 "gl_Position"
               OpMemberName %gl_PerVertex 1 "gl_PointSize"
               OpMemberName %gl_PerVertex 2 "gl_ClipDistance"
               OpMemberName %gl_PerVertex 3 "gl_CullDistance"
               OpName %testfun_vf4_ "testfun(vf4;"
               OpName %_ ""
               OpName %position "position"
               OpName %vtxColor "vtxColor"
               OpName %color "color"
               OpName %gl_VertexIndex "gl_VertexIndex"
               OpName %gl_InstanceIndex "gl_InstanceIndex"
               OpMemberDecorate %gl_PerVertex 0 BuiltIn Position
               OpMemberDecorate %gl_PerVertex 1 BuiltIn PointSize
               OpMemberDecorate %gl_PerVertex 2 BuiltIn ClipDistance
[...]

PR #519, opened 2017-01-05
Fixes issue #517

ehsannas:
entry_block_to_construct_ maps an entry block to its construct. The key
in this map (the entry block) is not unique, and therefore the entry for
the continue construct gets overwritten when the selection construct is
discovered.

Since a given block may be the entry block of different types of
constructs, the (basic_block, construct_type) pair should be able to
uniquely identify the construct.

### DIFF
--- a/source/val/construct.h
+++ b/source/val/construct.h
@@ -21,7 +21,7 @@
 namespace libspirv {
 
 enum class ConstructType {
-  kNone,
+  kNone = 0,
   /// The set of blocks dominated by a selection header, minus the set of blocks
   /// dominated by the header's merge block
   kSelection,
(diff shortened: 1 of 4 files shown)


ehsannas:
Verified that it fixes the issue that the VK CTS was running into.

Side note: To avoid this from happening again:
* I will be running CTS on every PR on spirv-tools.
* I am also writing a unit test that exercises the code that caused this crash.

source/val/construct.h
```diff
@@ -123,4 +123,15 @@ class Construct {
 
 }  /// namespace libspirv
 
+// Override std::hash for libspirv::ConstructType so that the enum can be used
+// as a key in an unordered_map.
+namespace std {
```
dneto0: Style guide says don't override std::hash.
https://google.github.io/styleguide/cppguide.html#std_hash

Make your own class name for the is implementation.

dneto0:
The code change looks good.  I'd like to see a followup CL with a test exercising this case.  I'd rather we catch this bug in our own test suite rather than in downstream Vulkan CTS.

dneto0:
Squashed and pushed into master as 109ce225c9

Issue #490, opened 2016-12-08
Check universal rule for Entry point and execution model


dneto0:
2.16.1

"There is at least one OpEntryPoint instruction, unless the Linkage capability is being used.

No function can be targeted by both an OpEntryPoint instruction and an OpFunctionCall instruction."

PR #522, opened 2017-01-10
Validation of OpEntryPoint usage.

ehsannas:
According to the SPIRV Spec (2.16.1):
* There is at least one OpEntryPoint instruction, unless the Linkage
capability is being used.

* No function can be targeted by both an OpEntryPoint instruction and an
OpFunctionCall instruction.

Update to the unit tests:
* Most unit tests were written without considering this validation rule. Therefore I had to update several unit tests. 
* In most cases, adding the "Linkage" capability was the harmless fix. 
* Since unit tests in val_capability_test.cpp are actually testing capability dependencies, I have actually added an "OpEntryPoint" to those tests.
* I also added 2 new unit tests to test my new code.

Fixes Issue #490 
Passes VK CTS.

### DIFF
--- a/source/val/validation_state.h
+++ b/source/val/validation_state.h
@@ -129,9 +129,14 @@ class ValidationState_t {
   std::vector<uint32_t>& entry_points() { return entry_points_; }
   const std::vector<uint32_t>& entry_points() const { return entry_points_; }
 
-  /// Returns a list of function IDs that are target of OpFunctionCall.
-  std::vector<uint32_t>& function_call_targets() {
-    return function_call_targets_;
+  /// Inserts an <id> to the set of functions that are target of OpFunctionCall.
+  void AddFunctionCallTarget(const uint32_t id) {
+    function_call_targets_.insert(id);
+  }
+
+  /// Returns whether or not a function<id> is the target of OpFunctionCall.
+  bool IsFunctionCallTarget(const uint32_t id) {
+    return (function_call_targets_.find(id) != function_call_targets_.end());
   }
 
   /// Registers the capability and its dependent capabilities
@@ -261,7 +266,7 @@ class ValidationState_t {
   std::vector<uint32_t> entry_points_;
 
   /// Functions IDs that are target of OpFunctionCall.
-  std::vector<uint32_t> function_call_targets_;
+  std::unordered_set<uint32_t> function_call_targets_;
 
   /// ID Bound from the Header
   uint32_t id_bound_;
--- a/source/val/validation_state.h
+++ b/source/val/validation_state.h
@@ -129,6 +129,11 @@ class ValidationState_t {
   std::vector<uint32_t>& entry_points() { return entry_points_; }
   const std::vector<uint32_t>& entry_points() const { return entry_points_; }
 
+  /// Returns a list of function IDs that are target of OpFunctionCall.
+  std::vector<uint32_t>& function_call_targets() {
+    return function_call_targets_;
+  }
+
   /// Registers the capability and its dependent capabilities
   void RegisterCapability(SpvCapability cap);
 
@@ -254,6 +259,9 @@ class ValidationState_t {
 
   /// IDs that are entry points, ie, arguments to OpEntryPoint.
   std::vector<uint32_t> entry_points_;
+
+  /// Functions IDs that are target of OpFunctionCall.
+  std::vector<uint32_t> function_call_targets_;
 
   /// ID Bound from the Header
   uint32_t id_bound_;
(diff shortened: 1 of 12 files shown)


source/val/validation_state.h
```diff
@@ -255,6 +260,9 @@ class ValidationState_t {
   /// IDs that are entry points, ie, arguments to OpEntryPoint.
   std::vector<uint32_t> entry_points_;
 
+  /// Functions IDs that are target of OpFunctionCall.
+  std::vector<uint32_t> function_call_targets_;
```
dneto0: It would be better to make this an unordered_map.  We don't want repeated entries if there are several calls to the same function.  We want a fast lookup instead of a linear lookup.   We want O(n+m) time when there are n calls and m function definitions but instead using a std::vector gives us O(n**2 + m).

test/val/val_cfg_test.cpp
```diff
@@ -1069,6 +1072,7 @@ TEST_F(ValidateCFG, OpSwitchToUnreachableBlock) {
 
   string str = R"(
 OpCapability Shader
+OpCapability Linkage
```
dneto0: Don't need this one.  Already have an entry point.

test/val/val_cfg_test.cpp
```diff
@@ -1108,6 +1112,7 @@ OpDecorate %id BuiltIn GlobalInvocationId
 TEST_F(ValidateCFG, LoopWithZeroBackEdgesBad) {
   string str = R"(
            OpCapability Shader
+           OpCapability Linkage
```
dneto0: Don't need this one.

test/val/val_cfg_test.cpp
```diff
@@ -1133,6 +1138,7 @@ TEST_F(ValidateCFG, LoopWithZeroBackEdgesBad) {
 TEST_F(ValidateCFG, LoopWithBackEdgeFromUnreachableContinueConstructGood) {
   string str = R"(
            OpCapability Shader
+           OpCapability Linkage
```
dneto0: Don't need this one.

test/val/val_cfg_test.cpp
```diff
@@ -1338,6 +1345,42 @@ TEST_F(ValidateCFG, BasicBlockIsEntryBlockOfTwoConstructsGood) {
...
+// Invalid. A function may not be a target of both OpEntryPoint and
+// OpFunctionCall.
+TEST_F(ValidateCFG, FunctionIsTargetOfEntryPointAndFunctionCallBad) {
+  std::string spirv = R"(
+           OpCapability Shader
+           OpMemoryModel Logical GLSL450
+           OpEntryPoint Fragment %foo "foo"
+%voidt   = OpTypeVoid
+%funct   = OpTypeFunction %voidt
+%floatt  = OpTypeFloat 32
```
dneto0: %floatt is unused. Delete this line.

test/val/val_cfg_test.cpp
```diff
@@ -1338,6 +1345,42 @@ TEST_F(ValidateCFG, BasicBlockIsEntryBlockOfTwoConstructsGood) {
   EXPECT_EQ(SPV_SUCCESS, ValidateInstructions());
 }
 
+// Tests that not having OpEntryPoint causes an error.
+TEST_F(ValidateCFG, NoEntryPointBad) {
```
dneto0: This isn't really a CFG test.  It's misclassified, by name certainly. Perhaps it needs to move to a different file too.

test/val/val_cfg_test.cpp
```diff
@@ -1338,6 +1345,42 @@ TEST_F(ValidateCFG, BasicBlockIsEntryBlockOfTwoConstructsGood) {
...
+           OpFunctionEnd
+      )";
+  CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_ERROR_INVALID_BINARY, ValidateInstructions());
+  EXPECT_THAT(
+      getDiagnosticString(),
+      HasSubstr("A function (1) may not be targeted by both an OpEntryPoint "
+                "instruction and an OpFunctionCall instruction."));
+}
+
```
dneto0: Should also make a case specifically to test that a module with a function but no entry point is valid when you have Linkage.
All the other cases you've introduced in this MR are incidental since those other tests are testing other things.

source/val/validation_state.h
dneto0: It might also be true that entry_points could be an unordered map.  But that would be for another MR.

test/c_interface_test.cpp
```diff
@@ -26,7 +26,9 @@ using namespace spvtools;
 TEST(CInterface, DefaultConsumerNullDiagnosticForValidInput) {
   auto context = spvContextCreate(SPV_ENV_UNIVERSAL_1_1);
   const char input_text[] =
-      "OpCapability Shader\nOpMemoryModel Logical GLSL450";
+      "OpCapability Shader\n"
+      "OpCapability Linkage\n"
```
dneto0: It turns out that Vulkan does not permit Linkage capability. So this combination doesn't correspond to a real environment.  But that's ok.  A distinct Vulkan env check is yet to be coded.

dneto0:
Rebased, trivial merge fix, then pushed into master as 1c11c86

PR #523, opened 2017-01-11
Introducing a new flow for running the Validator.

ehsannas:
We are adding a new API which can be called to run the SPIR-V validator,
and retrieve the ValidationState_t object. This is very useful for
unit testing.

I have also added basic unit tests that demonstrate usage of this flow
and ease of use to verify correctness.

This change makes it easier for future developments to be unit testes.
Passes VK CTS.

### DIFF
--- a/source/validate.h
+++ b/source/validate.h
@@ -190,11 +190,15 @@ spv_result_t spvValidateIDs(const spv_instruction_t* pInstructions,
                             spv_position position,
                             const spvtools::MessageConsumer& consumer);
 
-// Performs validation for the SPIRV-V module binary and returns the
-// ValidationState to the caller.
-spv_result_t ValidateAndReturnValidationState(
+namespace spvtools {
+// Performs validation for the SPIRV-V module binary.
+// The main difference between this API and spvValidateBinary is that the
+// "Validation State" is not destroyed upon function return; it lives on and is
+// pointed to by the vstate unique_ptr.
+spv_result_t ValidateBinaryAndKeepValidationState(
     const spv_const_context context, const uint32_t* words,
     const size_t num_words, spv_diagnostic* pDiagnostic,
     std::unique_ptr<libspirv::ValidationState_t>* vstate);
+}  // namespace spvtools
 
 #endif  // LIBSPIRV_VALIDATE_H_
--- a/source/validate.h
+++ b/source/validate.h
@@ -190,4 +190,11 @@ spv_result_t spvValidateIDs(const spv_instruction_t* pInstructions,
                             spv_position position,
                             const spvtools::MessageConsumer& consumer);
 
+// Performs validation for the SPIRV-V module binary and returns the
+// ValidationState to the caller.
+spv_result_t ValidateAndReturnValidationState(
+    const spv_const_context context, const uint32_t* words,
+    const size_t num_words, spv_diagnostic* pDiagnostic,
+    std::unique_ptr<libspirv::ValidationState_t>* vstate);
+
 #endif  // LIBSPIRV_VALIDATE_H_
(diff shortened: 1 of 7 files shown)


source/validate.h
```diff
@@ -190,4 +190,11 @@ spv_result_t spvValidateIDs(const spv_instruction_t* pInstructions,
                             spv_position position,
                             const spvtools::MessageConsumer& consumer);
 
+// Performs validation for the SPIRV-V module binary and returns the
+// ValidationState to the caller.
+spv_result_t ValidateAndReturnValidationState(
```
dneto0: This should be inside the spvtools namespace.

All the public names in the C API start with "spv". We want to keep that pattern here as well.  So with a name that doesn't start with "spv" you should put it inside the spvtools namespace.

source/validate.h
```diff
@@ -190,4 +190,11 @@ spv_result_t spvValidateIDs(const spv_instruction_t* pInstructions,
                             spv_position position,
                             const spvtools::MessageConsumer& consumer);
 
+// Performs validation for the SPIRV-V module binary and returns the
+// ValidationState to the caller.
```
dneto0: Well, it just returns a status.  If you want to say "return" the maybe return an actual pair of status and unique-ptr.

test/CMakeLists.txt
```diff
@@ -50,6 +50,8 @@ function(add_spvtools_unittest)
     target_include_directories(${target} PRIVATE
       ${SPIRV_HEADER_INCLUDE_DIR}
       ${spirv-tools_SOURCE_DIR}
+      ${spirv-tools_SOURCE_DIR}/source
```
dneto0: Hm. We already have spirv-tools_SOURCE_DIR, so instead of adding these new dirs, your #include should have the relative path from there: e.g.  "source/decoration.h" and "source/val/validation_state.h"

test/val/val_fixtures.h
```diff
@@ -18,6 +18,7 @@
 #define LIBSPIRV_TEST_VALIDATE_FIXTURES_H_
 
 #include "unit_spirv.h"
+#include "validation_state.h"
```
dneto0: Better to be #include "source/val/validation_state.h" or whatever.

test/val/val_fixtures.h
```diff
@@ -45,11 +46,17 @@ class ValidateBase : public ::testing::Test,
   // spvValidate function
   spv_result_t ValidateInstructions(spv_target_env env = SPV_ENV_UNIVERSAL_1_0);
 
+  // Performs validation on the SPIR-V code and retrieves the
+  // ValidationState object.
```
dneto0: Should say what the return value is, and where the validation object goes.  It is stored into vstate_.


"Performs validation. Returns the status and stores validation state into the vstate_ member."

test/val/val_validation_state_test.cpp
```diff
@@ -58,7 +58,7 @@ TEST_F(ValidationStateTest, CheckNumGlobalVars) {
   )";
   CompileSuccessfully(spirv);
   EXPECT_EQ(SPV_SUCCESS, ValidateAndRetrieveValidationState());
-  EXPECT_EQ(2, vstate_->num_global_vars());
+  EXPECT_EQ(unsigned(2), vstate_->num_global_vars());
```
dneto0: This is ok.  I usually just use a "u" suffix.  It's short and sweet.
The compiler will do the upconversion to size_t, but at least 2u is of the right signedness type.

E.g. 2u

ehsannas: Oh. I just learned something new :) 👍

dneto0:
Rebased, squashed and pushed into master as 68e36ec7

PR #525, opened 2017-01-12
Update Travis CI configuration.

antiagainst:
* Stop requiring sudo.
* Stop updating brew.
* Shadow clone of dependencies.
* Fix `nproc` on macOS.


antiagainst:
Again, without `sudo` the  compiler panics. So keep it for now.

PR #526, opened 2017-01-13
Decoration

ehsannas:
Only need to review a7c7cf6. Other 2 commits are already up for review (pr #523 )

* Added the decoration class as well as the code that registers the
decorations for each \<id\> and also decorations for struct members.

* Added unit tests for decorations in ValidationState as well as
decoration id tests.

### DIFF
--- a/source/val/decoration.h
+++ b/source/val/decoration.h
@@ -1,0 +1,46 @@
+// Copyright (c) 2016 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_VAL_DECORATION_H_
+#define LIBSPIRV_VAL_DECORATION_H_
+
+#include <unordered_map>
+#include <vector>
+
+namespace libspirv {
+
+class Decoration {
+ public:
+  Decoration(SpvDecoration t, std::vector<uint32_t>& parameters,
+             uint32_t member_index = -1)
+      : dec_type_(t), params_(parameters), struct_member_index_(member_index) {}
+
+  void set_struct_member_index(uint32_t index) { struct_member_index_ = index; }
+  int struct_member_index() { return struct_member_index_; }
+  SpvDecoration dec_type() { return dec_type_; }
+  std::vector<uint32_t>& params() { return params_; }
+
+ private:
+  SpvDecoration dec_type_;
+  std::vector<uint32_t> params_;
+
+  // If the decoration applies to a member of a structure type, then the index
+  // of the member is stored here.
+  int struct_member_index_;
+};
+
+}  // namespace libspirv
+
+#endif  /// LIBSPIRV_VAL_DECORATION_H_
+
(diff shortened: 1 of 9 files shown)


ehsannas:
Passes VK CTS.

dneto0:
Sorry for the inline comments. I hoped it would go into a review of the PR but since I was just reviwing a single commit it spammed you.

In any case:
>Yes, a decoration object doesn't store the target <id>. That's intentional.  ...

Yes, I can infer that as part of this code review and by reading the delta.  But this comment you put in the PR belongs *as comments in the code* basically.  That is, don't explain it in the code review stream, explain it in the code that you check in.  General rule.

I've stopped commenting now.  

Example uses in other tests of the uniform-initialiation syntax is here:
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/text_to_binary_test.cpp#L226
There I didn't use std::vector for the list of paramters that make up the vector to MakeInstruction's std::vector argument.

source/val/decoration.h
```diff
@@ -0,0 +1,110 @@
...
+    return (dec_type_ == rhs.dec_type_ && params_ == rhs.params_ &&
+            struct_member_index_ == rhs.struct_member_index_);
+  }
+
+ private:
+  SpvDecoration dec_type_;
+  std::vector<uint32_t> params_;
+
+  // If the decoration applies to a member of a structure type, then the index
+  // of the member is stored here.
```
dneto0: An index is always stored here.  You need the "otherwise" case.  "Otherwise, this is kInvalidIndex".

test/val/val_decoration_test.cpp
```diff
@@ -0,0 +1,160 @@
...
+    %1 = OpTypeRuntimeArray %2
+    ; Since %1 is used first in Decoration, it gets id 1.
+)";
+  const uint32_t id = 1;
+  CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_SUCCESS, ValidateAndRetrieveValidationState());
+  // Must have 2 decorations.
+  EXPECT_THAT(vstate_->id_decorations(id),
+              Eq(vector<Decoration>{
+                  Decoration(SpvDecorationArrayStride, vector<uint32_t>{4}),
```
dneto0: Did you try this without vector<uint32_t>  annotation?  I think it should work.
(Yup, I just tried GCC.)

The tricky thing is only that you need the vector<Decoration> to be explicit because the Eq template needs a little help to match the vstate_->id_decorations(id) value under test.

test/val/val_decoration_test.cpp
```diff
@@ -0,0 +1,160 @@
...
+         %3 = OpVariable %_ptr_Uniform__struct_10 Uniform
+ %_struct_11 = OpTypeStruct %_runtimearr_float
+%_ptr_Uniform__struct_11 = OpTypePointer Uniform %_struct_11
+         %4 = OpVariable %_ptr_Uniform__struct_11 Uniform
+  )";
+  CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_SUCCESS, ValidateAndRetrieveValidationState());
+
+  // Decoration group has 3 decorations.
+  auto check_decorations_for_id = [this](const uint32_t obj_id) {
```
dneto0: Putting the EXPECT_THAT inside a function will obscure the location of the failure if/when failure occurs.

Alternately ,to save typing you could assign the vector<Decoration> value to a constant value, and then do three EXPECT_THATs.

source/val/decoration.h
```diff
@@ -0,0 +1,110 @@
...
+// OpMemberDecorate %struct 2 Offset 2
+//            dec_type_ = SpvDecorationOffset
+//              params_ = vector { 2 }
+// struct_member_index_ = 2
+//
+// Note that the Decoration object does not store the target <id>. It is
+// possible for the same decoration to be applied to several <id>s (and they
+// might be assigned using separate spir-v instructions, possibly using an
+// assignment through GroupDecorate). It is important, however, that each <id>
+// knows the decorations that applies to it. Therefore, the ValidationState_t
```
ehsannas: :+1: Great point. +1 for the quote.

dneto0:
Squashed and pushed into master as fcf7df069

Issue #499, opened 2016-12-08
Check universal rule for Global (Module Scope) variables


dneto0:
2.16.1: "It is illegal to initialize an imported variable. This means that a module-scope OpVariable with initialization value cannot be marked with the Import Linkage Type."

dneto0:
Fixed with d75bf5e

PR #527, opened 2017-01-16
Validation for decoration rules. Fixes issue #499.

ehsannas:
In this change the SPIRV Spec Section 2.16.1 is implemented:
"It is illegal to initialize an imported variable. This means
that a module-scope OpVariable with initialization value cannot be
marked with the Import Linkage Type."

Also added unit tests.


dneto0:
Please rebase.

source/val/decoration.h
```diff
@@ -62,7 +62,9 @@ class Decoration {
   void set_struct_member_index(uint32_t index) { struct_member_index_ = index; }
   int struct_member_index() { return struct_member_index_; }
   SpvDecoration dec_type() { return dec_type_; }
+  SpvDecoration dec_type() const { return dec_type_; }
```
dneto0: In retrospect dec_type maybe should have just been called type.
Meh.

source/validate_decorations.cpp
```diff
@@ -0,0 +1,57 @@
...
+// Validates that decorations have been applied properly.
+spv_result_t ValidateDecorations(ValidationState_t& vstate) {
+  // According the SPIR-V Spec 2.16.1, it is illegal to initialize an imported
+  // variable. This means that a module-scope OpVariable with initialization
+  // value cannot be marked with the Import Linkage Type (import type id = 1).
+  for (auto global_var_id : vstate.global_vars()) {
+    // Initializer <id> is an optional argument for OpVariable. If initializer
+    // <id> is present, the instruction will have 5 words.
+    auto variable_instr = vstate.FindDef(global_var_id);
+    if (variable_instr->words().size() == size_t(5)) {
```
dneto0: 5u would have done here.
Meh.

test/val/val_decoration_test.cpp
```diff
@@ -146,5 +146,37 @@ TEST_F(ValidateDecorations, ValidateGroupMemberDecorateRegistration) {
...
+TEST_F(ValidateDecorations, LinkageImportUsedForInitializedVariableBad) {
+  string spirv = R"(
+               OpCapability Shader
+               OpCapability Linkage
+               OpMemoryModel Logical GLSL450
+               OpDecorate %target LinkageAttributes "link_ptr" Import
+      %float = OpTypeFloat 32
+ %_ptr_float = OpTypePointer Uniform %float
+      %var_1 = OpVariable %_ptr_float Input
+     %target = OpVariable %_ptr_float Uniform %var_1
```
dneto0: The intializer for a float variable should be a float value.  But %var_1 is itself a pointer to float.
So this test would fail for the wrong reason if we had type checking for initializers.

Suggest that instead of %var_1 you write:

%zero = OpConstantNull %float

Then make %zero your initializer value.

test/val/val_decoration_test.cpp
```diff
@@ -146,5 +146,37 @@ TEST_F(ValidateDecorations, ValidateGroupMemberDecorateRegistration) {
...
+TEST_F(ValidateDecorations, LinkageExportUsedForInitializedVariableGood) {
+  string spirv = R"(
+               OpCapability Shader
+               OpCapability Linkage
+               OpMemoryModel Logical GLSL450
+               OpDecorate %target LinkageAttributes "link_ptr" Export
+      %float = OpTypeFloat 32
+ %_ptr_float = OpTypePointer Uniform %float
+      %var_1 = OpVariable %_ptr_float Output
+     %target = OpVariable %_ptr_float Uniform %var_1
```
dneto0: Same here. Make and use a %zero

source/val/decoration.h
ehsannas: yeah, I think so too. 👍
I'll make this change after the current PRs go through because making the change now would cause rebase conflicts.

dneto0:
Squashed and pushed into master as d75bf5ef03

PR #528, opened 2017-01-17
make build reproducible

bmwiedemann:
see https://reproducible-builds.org/ for why this is good

### DIFF
--- a/utils/update_build_version.py
+++ b/utils/update_build_version.py
@@ -33,10 +33,12 @@ from __future__ import print_function
 
 import datetime
 import errno
+import os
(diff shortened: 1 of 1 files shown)


dneto0:
I agree, reproducible builds are a good idea.

Thank you for pointing out this instance of non-reproducibility.   The could be better by adding a comment describing why that code sequence is doing what it's doing.  Also, the formatting is off.

dneto0:
I added a comment explaining why the code does what it does, and fixed python formatting by splitting the expression into more statements with intermediate variables.

PR #529, opened 2017-01-17
Built in decorations

ehsannas:
When applied to a structure-type member, all members of that structure
type must also be decorated with BuiltIn. (No allowed mixing of built-in
variables and non-built-in variables within a single structure.)

When applied to a structure-type member, that structure type cannot be
contained as a member of another structure type.

There is at most one object per Storage Class that can contain a
structure type containing members decorated with BuiltIn, consumed per
entry-point.

Passes VK CTS.

### DIFF
--- a/source/validate_decorations.cpp
+++ b/source/validate_decorations.cpp
@@ -20,8 +20,27 @@
 #include "opcode.h"
 #include "val/validation_state.h"
 
+using libspirv::Decoration;
 using libspirv::DiagnosticStream;
 using libspirv::ValidationState_t;
+
+namespace {
+
+// Returns whether the given structure has BuiltIn members.
+bool isBuiltInStruct(uint32_t struct_id, ValidationState_t& vstate) {
+  // According to the Universal Validation Rules of the SPIR-V Spec, if any one
+  // member is BuiltIn, then all members must be BuiltIn. Therefore, it is
+  // enough to check just one member.
+  for (const auto& decoration : vstate.id_decorations(struct_id)) {
+    if (SpvDecorationBuiltIn == decoration.dec_type() &&
+        Decoration::kInvalidMember != decoration.struct_member_index()) {
+      return true;
+    }
+  }
+  return false;
+}
+
+}  // anonymous namespace
 
 namespace libspirv {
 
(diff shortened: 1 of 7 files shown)


dneto0:
Please rebase against master.  That will help me focus on the actual change being applied.

source/validate_decorations.cpp
```diff
@@ -48,7 +65,49 @@ spv_result_t ValidateDecorations(ValidationState_t& vstate) {
...
+      Instruction* ptr_instr = vstate.FindDef(ptr_id);
+      // It is guaranteed (by validator ID checks) that ptr_instr is
+      // OpTypePointer. Word 3 of this instruction is the type being pointed to.
+      const uint32_t type_id = ptr_instr->word(3);
+      Instruction* type_instr = vstate.FindDef(type_id);
+      if (type_instr && SpvOpTypeStruct == type_instr->opcode() &&
+          isBuiltInStruct(type_id, vstate)) {
+        const auto storage_class =
+            static_cast<SpvStorageClass>(var_instr->word(3));
+        if (storage_class != SpvStorageClassInput &&
```
dneto0: This error is independent of whether the current thing has builtins.  So pull this check outside of the "if" at line 87.

source/validate_decorations.cpp
```diff
@@ -48,7 +65,49 @@ spv_result_t ValidateDecorations(ValidationState_t& vstate) {
...
+            storage_class != SpvStorageClassOutput) {
+          return vstate.diag(SPV_ERROR_INVALID_ID)
+                 << "OpEntryPoint interfaces must be OpVariables with "
```
dneto0: Improve the error:  Name the entry point by ID.

source/validate_decorations.cpp
```diff
@@ -48,7 +65,49 @@ spv_result_t ValidateDecorations(ValidationState_t& vstate) {
...
+                 << storage_class << ".";
+        }
+        if (storage_class == SpvStorageClassInput) ++num_builtin_inputs;
+        if (storage_class == SpvStorageClassOutput) ++num_builtin_outputs;
+        if (num_builtin_inputs > 1 || num_builtin_outputs > 1) break;
+      }
+    }
+    if (num_builtin_inputs > 1 || num_builtin_outputs > 1) {
+      return vstate.diag(SPV_ERROR_INVALID_BINARY)
+             << "There must be at most one object per Storage Class that can "
```
dneto0: Improve the error:  Name the entry point by ID.

source/validate_id.cpp
```diff
@@ -418,6 +431,25 @@ bool idUsage::isValid<SpvOpTypeStruct>(const spv_instruction_t* inst,
...
+  for (auto decoration : vstate.id_decorations(inst->words[1])) {
+    if (decoration.dec_type() == SpvDecorationBuiltIn &&
+        decoration.struct_member_index() != Decoration::kInvalidMember) {
+      built_in_members.insert(decoration.struct_member_index());
+    }
+  }
+  int num_struct_members = static_cast<int>(inst->words.size() - 2);
+  int num_builtin_members = static_cast<int>(built_in_members.size());
+  if (num_builtin_members > 0 && num_builtin_members != num_struct_members) {
+    DIAG(0) << "When BuiltIn decoration is applied to a structure-type member, "
```
dneto0: Improve the error:  Name the struct type by ID.

test/val/val_decoration_test.cpp
```diff
@@ -178,5 +178,139 @@ TEST_F(ValidateDecorations, LinkageExportUsedForInitializedVariableGood) {
...
+  CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_ERROR_INVALID_BINARY, ValidateAndRetrieveValidationState());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("There must be at most one object per Storage Class "
+                        "that can contain a structure type containing members "
+                        "decorated with BuiltIn, consumed per entry-point."));
+}
+
+TEST_F(ValidateDecorations,
+       OneBuiltInObjectPerStorageClassConsumedByOpEntryPointGood) {
```
dneto0: Add an OpEntryPoint case which uses interface variables unrelated to built-ins.  It should pass validation.

source/validate_decorations.cpp
ehsannas: Good point.

dneto0:
Squashed and pushed into master as  9c71c572e5

Issue #497, opened 2016-12-08
Check universal rule for Linkage Attributes decorations


dneto0:
2.16.1 says:

"The Linkage Attributes Decoration cannot be applied to functions targeted by an OpEntryPoint instruction."

dneto0:
Fixed by https://github.com/KhronosGroup/SPIRV-Tools/pull/530

PR #530, opened 2017-01-17
Fixes Issue #497

ehsannas:
The Linkage Attributes Decoration cannot be applied to functions
targeted by an OpEntryPoint instruction.

### DIFF
--- a/source/validate_decorations.cpp
+++ b/source/validate_decorations.cpp
@@ -107,7 +107,22 @@ spv_result_t ValidateDecorations(ValidationState_t& vstate) {
                 "BuiltIn, consumed per entry-point. Entry Point id "
              << entry_point << " does not meet this requirement.";
     }
+    // The LinkageAttributes Decoration cannot be applied to functions targeted
+    // by an OpEntryPoint instruction
+    for (auto& decoration : vstate.id_decorations(entry_point)) {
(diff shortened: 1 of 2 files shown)


dneto0:
Please rebase.
Overall looks ok.  Please make the message more friendly by saying *which* object has the problem.  Target ID will work.  But the linkage attribute string name is even nicer.

source/validate_decorations.cpp
```diff
@@ -107,8 +107,23 @@ spv_result_t ValidateDecorations(ValidationState_t& vstate) {
                 "BuiltIn, consumed per entry-point. Entry Point id "
              << entry_point << " does not meet this requirement.";
     }
+    // The LinkageAttributes Decoration cannot be applied to functions targeted
+    // by an OpEntryPoint instruction
+    for (auto& decoration : vstate.id_decorations(entry_point)) {
+      if (SpvDecorationLinkageAttributes == decoration.dec_type()) {
+        const char* linkage_name =
```
dneto0: If we end up doing this a lot, then we should make a helper method on Decoration that returns a utils::string_view or at worst const char* for the string starting at a word pointer.  Basically just hide the ugliness.

But as a distributed systems prof said, "You do something once, it's a trick.  You do it twice, it's a technique."

Issue #493, opened 2016-12-08
Check function call operands and return type agree with the called function type


dneto0:

dneto0:
Is the SPIR-V spec missing this validation rule?

dneto0:
Fixed with eb7d2534d

PR #531, opened 2017-01-17
Fixes issue #493.

ehsannas:
Check function call operands and return type agree with the called
function type.

The validation code exists. I have fixed the unit tests.


dneto0:
Rebased and pushed into master as eb7d2534d

Issue #532, opened 2017-01-19
Incorrect validation of OpPtrAccessChain  and OpInBoundsPtrAccessChain 


AlexeySotkin:
For `OpPtrAccessChain` and `OpInBoundsPtrAccessChain` SPIR-V specification says that word count for these instructions is "5 + variable". It means that minimum size of the instruction is 5 words, i.e. these instructions without indexes are valid.
But in this case the validator reports "No Indexes were passed to OpInBoundsPtrAccessChain” error.
I think indexes operands for these 2 instructions should be optional.

[SPIRV to reproduce the problem](https://github.com/KhronosGroup/SPIRV-Tools/files/716849/tmp.zip)

ehsannas:
Thanks Alexey for pointing this out. I will fix this shortly.

dneto0:
I made https://github.com/KhronosGroup/SPIRV-Tools/pull/533 out of @ehsannas' change.
I reviewed it and pushed it into master.

PR #533, opened 2017-01-19
Fixes issue #532.

dneto0:
It is acceptable for OpAccessChain, OpInBoundsAccessChain,
OpPtrAccessChain, OpInBoundsPtrAccessChain, OpCompositeInsert, and
OpCompositeExtract to not take any indexes as arguments. In such cases,
no indexing will be done on the Base pointer/composite.


dneto0:
This is really @ehsannas' change.

dneto0:
I've already reviewed it.  
+2

PR #534, opened 2017-01-20
Add exhaustive function call inlining to spirv-opt

greg-lunarg:
Inlining is done for all functions designated as entry points.

This pass is intended primarily to be used with other (yet to be added) passes designed to reduce SPIR-V file size, also targeting entry point functions. In this usage, inlining serves in lieu of interprocedural analysis.

The next pass will target elimination of function-scope variables and their loads, stores and access chains.


source/opt/inline_pass.h
```diff
@@ -0,0 +1,89 @@
...
+                std::unique_ptr<ir::BasicBlock>& bp);
+
+  void AddLoad(uint32_t typeId, uint32_t resultId, uint32_t srcId,
+               std::unique_ptr<ir::BasicBlock>& bp);
+
+  // Return in newBlocks the result of inlining the call at call_ii within
+  // its block call_bi. Also return in newVars additional OpVariable
+  // instructions required by and to be inserted into the caller function
+  // after the block call_bi is replaced with newBlocks.
+  void GenInlineCode(std::vector<std::unique_ptr<ir::BasicBlock>>& newBlocks,
```
dneto0: Maybe the answers to these algorithmic questions should be documented at a module-scope comment at the top of the .cpp file.  They might not merit answering at this method declaration.  But they should be written down somewhere.

source/opt/iterator.h
```diff
@@ -183,6 +201,30 @@ inline
...
+template <bool IsConstForMethod>
+inline
+    typename std::enable_if<!IsConstForMethod, UptrVectorIterator<VT, IC>>::type
+    UptrVectorIterator<VT, IC>::MoveBefore(UptrVector& values) {
+  const auto pos = iterator_ - container_->begin();
+  const auto origsz = container_->size();
+  container_->resize(origsz + values.size());
+  std::move_backward(container_->begin() + pos, container_->begin() + origsz,
+                     container_->end());
+  std::move(values.begin(), values.end(), container_->begin() + pos);
```
dneto0: TIL.  I can't find a reference for this instantiation of std::move, though I can guess what this does. :-)

source/opt/iterator.h
```diff
@@ -81,6 +81,24 @@ class UptrVectorIterator
...
   InsertBefore(Uptr value);
 
+  // Inserts the given |valueVector| to the position pointed to by this iterator
+  // and returns an iterator to the newly inserted |valueVector|.
+  // If the underlying vector changes capacity, all previous iterators will be
+  // invalidated. Otherwise, those previous iterators pointing to after the
+  // insertion point will be invalidated.
+  template <bool IsConstForMethod = IsConst>
+  inline typename std::enable_if<!IsConstForMethod, UptrVectorIterator>::type
+  MoveBefore(UptrVector& valueVector);
```
dneto0: Something is amiss.  If this is really a move then the parameter should be an rvalue reference (&&).
And the description should say it *moves* the values in the vector.

But if it doesn't move the contents of the called parameter, then this method should be renamed to InsertBefore (just another overload of the previous method.)  It may be that you're getting a temporary copy of the vector at the call boundary that you didn't intend.

source/opt/module.h
```diff
@@ -51,7 +51,7 @@ class Module {
   // Sets the header to the given |header|.
   void SetHeader(const ModuleHeader& header) { header_ = header; }
   // Sets the Id bound.
-  void SetIdBound(uint32_t bound) { header_.bound = bound; }
+  inline void SetIdBound(uint32_t bound) { header_.bound = bound; }
```
dneto0: When the body is present at the declaration of a method in a class, then it's already implicitly inline, IIRC.

source/validate_id.cpp
```diff
@@ -3044,7 +3044,7 @@ function<bool(unsigned)> getCanBeForwardDeclaredFunction(SpvOp opcode) {
       break;
 
     case SpvOpPhi:
-      out = [](unsigned index) { return index > 1; };
+      out = [](unsigned index) { return index > 1 && index % 2 == 1; };
```
dneto0: No so.
The point of an OpPhi is to bring value from a later block into an earlier block, when you can't yet "see" those later definitions.


  %n = OpPhi %int %next_n %label %zero %entry
   ...
   ; loop variable update.
   %next_n = OpIAdd %int %n %one  ; this is a forward declaration 
```

File: source/opt/inline_pass.cpp [hunk 6]
```diff
@@ -0,0 +1,432 @@
+// Copyright (c) 2016 The Khronos Group Inc.
+// Copyright (c) 2016 Valve Corporation
+// Copyright (c) 2016 LunarG Inc.
```
greg-lunarg: OK.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,432 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "inline_pass.h"
+#include "iterator.h"
+
+#include <unordered_map>
+
+static const int SPV_ENTRY_POINT_FUNCTION_ID = 1;
```
greg-lunarg: I added a comment and have adopted the Google style here. I did keep the Spv prefix; it gives a little context at the use points. I also left off the Index suffix since I added the comment and the role of an index is fairly clear at the use points.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,432 @@
...
+static const int SPV_FUNCTION_CALL_FUNCTION_ID = 2;
+static const int SPV_FUNCTION_CALL_ARGUMENT_ID = 3;
+static const int SPV_FUNCTION_PARAMETER_RESULT_ID = 1;
+static const int SPV_STORE_OBJECT_ID = 1;
```
greg-lunarg: Deleted.

source/opt/basic_block.h
```diff
@@ -90,6 +101,33 @@ inline void BasicBlock::ForEachInst(
         ->ForEachInst(f, run_on_debug_line_insts);
 }
 
+inline void BasicBlock::ForEachPhiInst(
+    const std::function<void(Instruction*)>& f, bool run_on_debug_line_insts) {
+  for (auto& inst : insts_) {
+    if (inst->opcode() != SpvOpPhi) break;
```
greg-lunarg: No change (yet). My reading of the code is that OpLine (and OpNoLines) preceding an instruction are stored in the dbg_line_insts_ member of the Instruction class, so we shouldn't see them in the insts_ member of a BasicBlock.

source/opt/basic_block.h
```diff
@@ -90,6 +101,33 @@ inline void BasicBlock::ForEachInst(
...
+    inst->ForEachInst(f, run_on_debug_line_insts);
+  }
+}
+
+inline void BasicBlock::ForEachSucc(const std::function<void(uint32_t)>& f) {
+  auto br = insts_.end() - 1;
```
greg-lunarg: Done.

source/opt/instruction.h
```diff
@@ -228,6 +245,28 @@ inline void Instruction::ForEachInst(
   f(this);
 }
 
+inline void Instruction::ForEachInId(const std::function<void(uint32_t*)>& f) {
+  for (auto& opnd : operands_)
```
greg-lunarg: Done.

Actually, there is now two versions of ForEachInId: a const version and one that allows for changing of the In Operands through the pointer. The const cannot be applied for the non-const version, but it can for the const version.

source/opt/inline_pass.h
```diff
@@ -0,0 +1,89 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_OPT_INLINE_PASS_H_
+#define LIBSPIRV_OPT_INLINE_PASS_H_
+
+#include <algorithm>
```
greg-lunarg: Done.

source/opt/inline_pass.h
```diff
@@ -0,0 +1,89 @@
...
+  // Map from function's result id to function
+  std::unordered_map<uint32_t, ir::Function*> id2function_;
+
+  // Map from block's label id to block
+  std::unordered_map<uint32_t, ir::BasicBlock*> id2block_;
+
+  // Next unused ID
+  uint32_t nextId_;
+
+  inline void finalizeNextId(ir::Module* module) {
```
greg-lunarg: Done.

source/opt/inline_pass.h
```diff
@@ -0,0 +1,89 @@
...
+    module->SetIdBound(nextId_);
+  }
+  inline uint32_t getNextId() { return nextId_++; }
```
greg-lunarg: Done.

source/opt/inline_pass.h
```diff
@@ -0,0 +1,89 @@
...
+  // all code that is inlined into func.
+  bool Inline(ir::Function* func);
+
+  // Find pointer to type in module, return its resultId. 0 if not found.
+  uint32_t FindPointerToType(uint32_t typeId);
+
+  // Add pointer to type to module and return resultId.
+  uint32_t AddPointerToType(uint32_t typeId);
+
+  void AddBranch(uint32_t labelId, std::unique_ptr<ir::BasicBlock>& bp);
```
greg-lunarg: Done.

source/opt/inline_pass.h
```diff
@@ -0,0 +1,89 @@
...
+
+  void AddStore(uint32_t destId, uint32_t valId,
```
greg-lunarg: destId (and srcId) changed to ptrId to be consistent with SPIR-V spec.

source/opt/inline_pass.h
greg-lunarg: Refs have been changed to pointers.

The content of the calling block call_bi is replicated in the blocks of newBlocks so that the old calling block can just be replaced by newBlocks. The code from the calling block is integrated into newBlocks such that additional branches are avoided. If the callee is a single block, the resulting newBlocks will be a single block.

In the case of debug info in the callee, it is cloned along with its instruction.

Early returns are replaced by storing to a local return variable and branching to a (created) exit block where the local variable is returned.

Formal parameters are trivially mapped to their actual parameters since they are read-only by definition of SPIR-V.

 I have updated this method comment to be clearer about all this.

source/opt/module.h
greg-lunarg: K.

source/opt/iterator.h
greg-lunarg: :)

source/validate_id.cpp
greg-lunarg: Reversed.

source/opt/iterator.h
greg-lunarg: It is documented here:

http://en.cppreference.com/w/cpp/algorithm/move

This is why I included <algorithm>.

source/opt/iterator.h
greg-lunarg: Well, if the previous routine could be called InsertBefore, then I guess this one can be as well. They are generally doing the same thing.

I made valueVector a ref for efficiency, but will change it to a pointer.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+  return modified;
+}
+
+void InlinePass::Initialize(ir::Module* module) {
+  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+
+  // Initialize next unused Id
+  nextId_ = 0;
+  for (const auto& id_def : def_use_mgr_->id_to_defs()) {
+    nextId_ = std::max(nextId_, id_def.first);
```
dneto0: The bots fail on this line.
std::max requires you '#include &lt;algorithm&gt;'

greg-lunarg: OK. Will add back.

dneto0:
Thanks for making the changes.  The changes are good.
Still to review: core of the inline algorithm, and the tests.  Sorry for the delay.

source/opt/iterator.h
dneto0: Nit:  I don't think reference arg and pointer arg have different efficiency.  In this case references mainly (only?) get your a front-end-language guarantee that the referenced thing is not null.

But no matter, you've already made the change I requested. :-)

greg-lunarg:
np :)


Greg Fischer

Project Manager and Senior Engineer

LunarG, Inc. - 3D Driver Innovations

greg@lunarg.com <alan@lunarg.com>

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+            // if previous instruction was early return, insert branch
+            // instruction
+            // to return block
+            if (prevInstWasReturn) {
+              if (returnLabelId == 0) returnLabelId = this->TakeNextId();
+              AddBranch(returnLabelId, &bp);
+              prevInstWasReturn = false;
+            }
+            // finish current block (if it exists) and get label for next block
+            uint32_t labelId;
```
antiagainst: Uninitialized variable.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// Generate callee code into newBlocks to be inlined for the function call at
+// call_ii. Also add new function variables into caller func
+
+void InlinePass::GenInlineCode(
+    std::vector<std::unique_ptr<ir::BasicBlock>>* newBlocks,
+    std::vector<std::unique_ptr<ir::Instruction>>* newVars,
+    ir::UptrVectorIterator<ir::Instruction> call_ii,
+    ir::UptrVectorIterator<ir::BasicBlock> call_bi) {
```
antiagainst: IMHO `fcall_inst` and `fcall_basic_block` is more clear as parameter names.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+                    new ir::Instruction(*cii));
+                // remember OpSampledImages for possible regeneration
+                if (spv_inst->opcode() == SpvOpSampledImage) {
+                  auto* samp_inst_ptr = spv_inst.get();
+                  preCallSI[spv_inst->result_id()] = samp_inst_ptr;
+                }
+                bp->AddInstruction(std::move(spv_inst));
+              }
+            } else
+              multiBlocks = true;
```
antiagainst: `else { ... }`. `if` and `else` should match w.r.t. `{}`. If one of them has `{}`, the other should do.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+    std::unique_ptr<ir::Instruction> var_inst(new ir::Instruction(
+        SpvOpVariable, returnVarTypeId, returnVarId, in_operands));
+    newVars->push_back(std::move(var_inst));
+  }
+
+  // Clone and map callee code
+  bool prevInstWasReturn = false;
+  uint32_t returnLabelId = 0;
+  bool multiBlocks = false;
+  std::unique_ptr<ir::BasicBlock> bp;
```
antiagainst: I'd suggest renaming this as `new_basic_block`.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+            }
+            // load return value into result id of call, if it exists
+            if (returnVarId != 0) {
+              const uint32_t resId = call_ii->result_id();
+              assert(resId != 0);
+              AddLoad(calleeTypeId, resId, returnVarId, &bp);
+            }
+            // copy remaining instructions from caller block.
+            auto cii = call_ii;
+            cii++;
```
antiagainst: This can be moved into the init part of the `for` loop.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+                  std::initializer_list<uint32_t>{ptrId}));
+  std::unique_ptr<ir::Instruction> newLoad(
+      new ir::Instruction(SpvOpLoad, typeId, resultId, load_in_operands));
+  (*bp)->AddInstruction(std::move(newLoad));
+}
+
+// Generate callee code into newBlocks to be inlined for the function call at
+// call_ii. Also add new function variables into caller func
+
+void InlinePass::GenInlineCode(
```
antiagainst: This method is quite large which negatively impacts the readability. Maybe consider break it into several methods?

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+              });
+            });
+          });
+        }
+        // replace old calling block with new block(s)
+        bi = bi.Erase();
+        bi = bi.InsertBefore(&newBlocks);
+        // insert new function variables
+        if (newVars.size() > 0) {
+          auto vbi = func->begin();
```
antiagainst: `func->begin()->begin()->InsertBefore(...)`?

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+          id2block_[blk->GetLabelId()] = &*blk;
+        }
+        // update phi functions in succesor blocks if call block
+        // is replaced with more than one block
+        if (newBlocks.size() > 1) {
+          const auto firstBlk = newBlocks.begin();
+          const auto lastBlk = newBlocks.end() - 1;
+          const uint32_t firstId = (*firstBlk)->GetLabelId();
+          const uint32_t lastId = (*lastBlk)->GetLabelId();
+          (*lastBlk)->ForEachSucc([&firstId, &lastId, this](uint32_t succ) {
```
antiagainst: Nice! :)

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+// call_ii. Also add new function variables into caller func
+
+void InlinePass::GenInlineCode(
+    std::vector<std::unique_ptr<ir::BasicBlock>>* newBlocks,
+    std::vector<std::unique_ptr<ir::Instruction>>* newVars,
+    ir::UptrVectorIterator<ir::Instruction> call_ii,
+    ir::UptrVectorIterator<ir::BasicBlock> call_bi) {
+  // Map from callee id to caller id
+  std::unordered_map<uint32_t, uint32_t> callee2caller;
+  // Pre-call OpSampledImage Insts
```
dneto0: The handling you do for OpSampledImage must also be done for OpImage.
This is due to validation rule:

"Instructions for extracting a scalar image or scalar sampler out of a composite must only use dynamically-uniform indexes. They must be in the same block in which their Result id are consumed."

(I don't see how to extract a scalar sampler from an OpSampledImage.)

dneto0:
Overall I found the core GenInline* function really long.  I'd suggest shortening it with more helper functions to make specific instructions.  Finding other ways to shorten it would also be good.

source/opt/basic_block.h
```diff
@@ -45,6 +45,9 @@ class BasicBlock {
   // Appends an instruction to this basic block.
   inline void AddInstruction(std::unique_ptr<Instruction> i);
 
+  // Returns the id of the label at the top of this block
+  inline uint32_t GetLabelId() { return label_->result_id(); }
```
greg-lunarg: Agree.

source/opt/basic_block.h
```diff
@@ -57,6 +60,14 @@ class BasicBlock {
   inline void ForEachInst(const std::function<void(const Instruction*)>& f,
                           bool run_on_debug_line_insts = false) const;
 
+  // Runs the given function |f| on each label id of each successor block
+  inline void ForEachSucc(const std::function<void(const uint32_t)>& f);
```
greg-lunarg: OK.

source/opt/basic_block.h
```diff
@@ -90,6 +101,34 @@ inline void BasicBlock::ForEachInst(
...
 
+inline void BasicBlock::ForEachPhiInst(
+    const std::function<void(Instruction*)>& f, bool run_on_debug_line_insts) {
+  for (auto& inst : insts_) {
+    if (inst->opcode() != SpvOpPhi) break;
+    inst->ForEachInst(f, run_on_debug_line_insts);
+  }
+}
+
+inline void BasicBlock::ForEachSucc(
```
greg-lunarg: There isn't a .cpp file in opt/. Create one?

source/opt/basic_block.h
```diff
@@ -90,6 +101,34 @@ inline void BasicBlock::ForEachInst(
...
+    const std::function<void(const uint32_t)>& f) {
+  const auto br = &*insts_.back();
+  switch (br->opcode()) {
+    case SpvOpBranch: {
+      f(br->GetOperand(0).words[0]);
+    } break;
+    case SpvOpBranchConditional:
+    case SpvOpSwitch: {
+      int cnt = 0;
```
greg-lunarg: Changed to is_first.

source/opt/basic_block.h
antiagainst: Yep, feel free to do that.

source/opt/function.h
```diff
@@ -49,6 +49,15 @@ class Function {
   // Saves the given function end instruction.
   inline void SetFunctionEnd(std::unique_ptr<Instruction> end_inst);
 
+  // Returns function's id
+  inline uint32_t GetResultId() { return def_inst_->result_id(); }
```
greg-lunarg: OK.

source/opt/inline_pass.h
```diff
@@ -0,0 +1,104 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_OPT_INLINE_PASS_H_
+#define LIBSPIRV_OPT_INLINE_PASS_H_
+
+#include <memory>
```
greg-lunarg: OK

source/opt/inline_pass.h
```diff
@@ -0,0 +1,104 @@
...
+
+ private:
+  // Map from function's result id to function
+  std::unordered_map<uint32_t, ir::Function*> id2function_;
+
+  // Map from block's label id to block
+  std::unordered_map<uint32_t, ir::BasicBlock*> id2block_;
+
+  // Next unused ID
+  uint32_t nextId_;
```
greg-lunarg: OK

source/opt/inline_pass.h
```diff
@@ -0,0 +1,104 @@
...
+
+  // Write the next available Id back to the module
+  inline void FinalizeNextId(ir::Module* module) {
+    module->SetIdBound(nextId_);
+  }
+
+  // Return the next available Id and increment it
+  inline uint32_t TakeNextId() { return nextId_++; }
```
greg-lunarg: Even with the auto-increment? g++ doesn't like it.

source/opt/inline_pass.h
```diff
@@ -0,0 +1,104 @@
...
+
+  // Exhaustively inline all function calls in func as well as in
+  // all code that is inlined into func.
+  bool Inline(ir::Function* func);
+
+  // Find pointer to type in module, return its resultId. 0 if not found.
+  uint32_t FindPointerToType(uint32_t typeId);
```
greg-lunarg: OK.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "inline_pass.h"
+
+// Indices of operands in SPIR-V instructions
+
+static const int kSpvEntryPointFunctionId = 1;
```
greg-lunarg: This is an <id> in SPIR-V terminology, so my preference would be to stay with that spelling.

source/opt/instruction.h
```diff
@@ -228,6 +246,34 @@ inline void Instruction::ForEachInst(
...
+}
+
+inline void Instruction::ForEachInId(
+    const std::function<void(const uint32_t*)>& f) const {
+  for (const auto& opnd : operands_)
+    if (opnd.type == SPV_OPERAND_TYPE_ID) f(&opnd.words[0]);
+}
+
+inline bool Instruction::IsControlFlow() const {
+  bool hasLabels = false;
```
greg-lunarg: OK

source/opt/instruction.h
```diff
@@ -228,6 +246,34 @@ inline void Instruction::ForEachInst(
...
+  switch (opcode_) {
+    case SpvOpSelectionMerge:
+    case SpvOpBranch:
+    case SpvOpLoopMerge:
+    case SpvOpBranchConditional:
+    case SpvOpSwitch:
+    case SpvOpPhi:
+      hasLabels = true;
```
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+namespace spvtools {
+namespace opt {
+
+uint32_t InlinePass::FindPointerToType(uint32_t typeId) {
+  ir::Module::inst_iterator type_itr = module_->types_values_begin();
+  for (; type_itr != module_->types_values_end(); ++type_itr) {
+    const ir::Instruction* type_inst = &*type_itr;
+    if (type_inst->opcode() == SpvOpTypePointer &&
+        type_inst->GetOperand(kSpvTypePointerTypeId).words[0] == typeId &&
```
greg-lunarg: How about changing the name of this to GetOneWordOperand()?

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+static const int kSpvFuncitonCallArgumentId = 3;
+static const int kSpvFunctionParameterResultId = 1;
+static const int kSpvReturnValueId = 0;
+static const int kSpvTypePointerStorageClass = 1;
+static const int kSpvTypePointerTypeId = 2;
+
+namespace spvtools {
+namespace opt {
+
+uint32_t InlinePass::FindPointerToType(uint32_t typeId) {
```
greg-lunarg: What would you want to call the method?

Another option is delete this method and just make the call

... = GetId(Pointer(GetType(type_id), SpvStorageClassFunction));

Which would you prefer?

greg-lunarg: I am leaning toward the second.

source/opt/inline_pass.h
antiagainst: Sorry, please ignore. :)

source/opt/inline_pass.cpp
antiagainst: I'm fine with that, but please do the renaming in another pull request since it is already used by many other places. :)

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+        type_inst->GetOperand(kSpvTypePointerStorageClass).words[0] ==
+            SpvStorageClassFunction)
+      break;
+  }
+  return (type_itr != module_->types_values_end()) ? type_itr->result_id() : 0;
+}
+
+uint32_t InlinePass::AddPointerToType(uint32_t typeId) {
+  uint32_t resultId = TakeNextId();
+  std::vector<ir::Operand> in_operands;
```
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+                           std::initializer_list<uint32_t>{uint32_t(typeId)});
+  std::unique_ptr<ir::Instruction> type_inst(
+      new ir::Instruction(SpvOpTypePointer, 0, resultId, in_operands));
+  module_->AddType(std::move(type_inst));
+  return resultId;
+}
+
+void InlinePass::AddBranch(uint32_t labelId,
+                           std::unique_ptr<ir::BasicBlock>* bp) {
+  std::vector<ir::Operand> branch_in_operands;
```
greg-lunarg: OK

source/opt/inline_pass.cpp
greg-lunarg: Changed to call_inst_itr and call_block_itr.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+  std::unordered_map<uint32_t, ir::Instruction*> preCallSI;
+  // Post-call OpSampledImage Ids
+  std::unordered_map<uint32_t, uint32_t> postCallSI;
+
+  const uint32_t calleeId =
+      call_ii->GetOperand(kSpvFunctionCallFunctionId).words[0];
+  ir::Function* calleeFn = id2function_[calleeId];
+
+  // Map parameters to actual arguments
+  int i = 0;
```
greg-lunarg: OK. Use param_idx.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+  calleeFn->ForEachParam(
+      [&call_ii, &i, &callee2caller](const ir::Instruction* cpi) {
+        const uint32_t pid =
+            cpi->GetOperand(kSpvFunctionParameterResultId).words[0];
```
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+        callee2caller[pid] =
+            call_ii->GetOperand(kSpvFuncitonCallArgumentId + i).words[0];
```
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+        i++;
+      });
+
+  // Define caller local variables for all callee variables and create map to
+  // them
+  auto cbi = calleeFn->begin();
```
greg-lunarg: Renamed to callee_block_itr and callee_var_itr.

source/opt/inline_pass.cpp
greg-lunarg: OK. Renamed to new_blk_ptr.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+       &preCallSI, this](const ir::Instruction* cpi) {
+        switch (cpi->opcode()) {
+          case SpvOpFunction:
+          case SpvOpFunctionParameter:
+          case SpvOpVariable:
+            // already processed
+            break;
+          case SpvOpLabel: {
+            // if previous instruction was early return, insert branch
+            // instruction
```
greg-lunarg: OK

source/opt/inline_pass.cpp
greg-lunarg: I do not see in the style guide that variables must be initialized. This variable is set on both paths of the following if. Still initialize? To zero?

source/opt/inline_pass.cpp
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+            assert(returnVarId != 0);
+            uint32_t valId = cpi->GetInOperand(kSpvReturnValueId).words[0];
+            const auto mapItr = callee2caller.find(valId);
+            if (mapItr != callee2caller.end()) {
+              valId = mapItr->second;
+            }
+            AddStore(returnVarId, valId, &bp);
+
+            // Remember we saw a return; if followed by a label, will need to
+            // insert
```
greg-lunarg: OK. Surprised clang-format doesn't do this

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+            // branch
+            prevInstWasReturn = true;
+          } break;
+          case SpvOpReturn: {
+            // Remember we saw a return; if followed by a label, will need to
+            // insert
```
greg-lunarg: OK. See above.

source/opt/inline_pass.cpp
greg-lunarg: Thanks! :)

source/opt/inline_pass.cpp
greg-lunarg: OK

source/opt/inline_pass.cpp
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+        id2function_[e.GetOperand(kSpvEntryPointFunctionId).words[0]];
+    modified = modified || Inline(fn);
+  }
+
+  FinalizeNextId(module_);
+
+  return modified ? Status::SuccessWithChange : Status::SuccessWithoutChange;
+}
+
+InlinePass::InlinePass()
```
greg-lunarg: OK

source/opt/basic_block.h
greg-lunarg: OK

source/opt/inline_pass.h
```diff
@@ -0,0 +1,104 @@
...
+  // Write the next available Id back to the module
+  inline void FinalizeNextId(ir::Module* module) {
+    module->SetIdBound(nextId_);
+  }
+
+  // Return the next available Id and increment it
+  inline uint32_t TakeNextId() { return nextId_++; }
+
+  // Exhaustively inline all function calls in func as well as in
+  // all code that is inlined into func.
```
greg-lunarg: OK

source/opt/inline_pass.h
```diff
@@ -0,0 +1,104 @@
...
+  bool Inline(ir::Function* func);
+
+  // Find pointer to type in module, return its resultId. 0 if not found.
+  uint32_t FindPointerToType(uint32_t typeId);
+
+  // Add pointer to type to module and return resultId.
+  uint32_t AddPointerToType(uint32_t typeId);
```
greg-lunarg: Added storage_id param

source/opt/inline_pass.h
```diff
@@ -0,0 +1,104 @@
...
+  // in newBlocks. Any additional branches are avoided. Debug instructions
+  // are cloned along with their callee instructions. Early returns are
+  // replaced by storing to a local return variable and branching to a 
+  // (created) exit block where the local variable is returned. Formal
+  // parameters are trivially mapped to their actual parameters.
+  //
+  // Also return in newVars additional OpVariable instructions required by 
+  // and to be inserted into the caller function after the block call_bi is 
+  // replaced with newBlocks.
+  void GenInlineCode(std::vector<std::unique_ptr<ir::BasicBlock>>* newBlocks,
```
greg-lunarg: Re the first bullet: I am not sure this is true. Likely I have not explained this well enough. The method creates a set of blocks such that they replicate the code of the original block with the call at call_inst_itr inlined. The original block is not changed by this method and technically is not included in new_blocks.

Re the second bullet, I think I do say that the original block can be replaced with the new blocks. I think the second bullet is implied by this statement and the rules of SPIR-V.

Does this help or would you still wish a change in the comment?

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+uint32_t InlinePass::FindPointerToType(uint32_t typeId) {
+  ir::Module::inst_iterator type_itr = module_->types_values_begin();
+  for (; type_itr != module_->types_values_end(); ++type_itr) {
+    const ir::Instruction* type_inst = &*type_itr;
+    if (type_inst->opcode() == SpvOpTypePointer &&
+        type_inst->GetOperand(kSpvTypePointerTypeId).words[0] == typeId &&
+        type_inst->GetOperand(kSpvTypePointerStorageClass).words[0] ==
+            SpvStorageClassFunction)
+      break;
```
greg-lunarg: Good point.

This function may disappear in lieu of using the type manager. Waiting to hear back from Lei on this subject.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+      ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID,
+                  std::initializer_list<uint32_t>{labelId}));
+  std::unique_ptr<ir::Instruction> newBranch(
+      new ir::Instruction(SpvOpBranch, 0, 0, branch_in_operands));
+  (*bp)->AddInstruction(std::move(newBranch));
+}
+
+void InlinePass::AddStore(uint32_t ptrId, uint32_t valId,
+                          std::unique_ptr<ir::BasicBlock>* bp) {
+  std::vector<ir::Operand> store_in_operands;
```
greg-lunarg: Done

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+      ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID,
+                  std::initializer_list<uint32_t>{valId}));
+  std::unique_ptr<ir::Instruction> newStore(
+      new ir::Instruction(SpvOpStore, 0, 0, store_in_operands));
+  (*bp)->AddInstruction(std::move(newStore));
+}
+
+void InlinePass::AddLoad(uint32_t typeId, uint32_t resultId, uint32_t ptrId,
+                         std::unique_ptr<ir::BasicBlock>* bp) {
+  std::vector<ir::Operand> load_in_operands;
```
greg-lunarg: Done

source/opt/inline_pass.cpp
greg-lunarg: Added TODO for now.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+  load_in_operands.push_back(
+      ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID,
+                  std::initializer_list<uint32_t>{ptrId}));
+  std::unique_ptr<ir::Instruction> newLoad(
+      new ir::Instruction(SpvOpLoad, typeId, resultId, load_in_operands));
+  (*bp)->AddInstruction(std::move(newLoad));
+}
+
+// Generate callee code into newBlocks to be inlined for the function call at
```
greg-lunarg: Deleted.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+// call_ii. Also add new function variables into caller func
+
+void InlinePass::GenInlineCode(
+    std::vector<std::unique_ptr<ir::BasicBlock>>* newBlocks,
+    std::vector<std::unique_ptr<ir::Instruction>>* newVars,
+    ir::UptrVectorIterator<ir::Instruction> call_ii,
+    ir::UptrVectorIterator<ir::BasicBlock> call_bi) {
+  // Map from callee id to caller id
+  std::unordered_map<uint32_t, uint32_t> callee2caller;
```
greg-lunarg: Good point. This is vague. This is a map used to convert all callee ids to their equivalent id in the caller as callee instructions are moved into the caller. I will try to make that more clear.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+                }
+                bp->AddInstruction(std::move(spv_inst));
+              }
+            } else
+              multiBlocks = true;
+          } break;
+          case SpvOpReturnValue: {
+            // store return value to return variable
+            assert(returnVarId != 0);
+            uint32_t valId = cpi->GetInOperand(kSpvReturnValueId).words[0];
```
greg-lunarg: Done

source/opt/inline_pass.cpp
greg-lunarg: I don't argue with your comment, although I must admit that I am having trouble understanding the "same block" rule for both OpImage and OpSampledImage. Maybe this isn't the place to have this discussion ultimately, but maybe I can bounce this off you first before taking it further.

It seems that any shader compiler can do the same analysis that I have to do to track down the instructions of interest when finally generating code. It seems to me it is sufficient that an OpSampledImage or OpImage must dominate its use. That plus prohibition against OpPhi and OpSelects (and OpStores?) having these as operands should be sufficient to guard against the kind of behavior that shader compiler writers are concerned about.

I object here partly because the addition of OpImage support to this algorithm will actually cause an escalation in the algorithm. Previously I could simply clone the OpSampledImage in the final block and be done, but now I have to scan the operands of the cloned instruction looking for OpImages. And I will have to scan OpImage operands looking for OpSampledImage. Basically, I will have to add an iterative/recursive component to the algorithm that will handle an alternating cascade of these instructions. This additional complication caused me to reconsider challenging the "same block" rule.

What do you think?

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+    std::vector<std::unique_ptr<ir::BasicBlock>>* newBlocks,
+    std::vector<std::unique_ptr<ir::Instruction>>* newVars,
+    ir::UptrVectorIterator<ir::Instruction> call_ii,
+    ir::UptrVectorIterator<ir::BasicBlock> call_bi) {
+  // Map from callee id to caller id
+  std::unordered_map<uint32_t, uint32_t> callee2caller;
+  // Pre-call OpSampledImage Insts
+  std::unordered_map<uint32_t, ir::Instruction*> preCallSI;
+  // Post-call OpSampledImage Ids
+  std::unordered_map<uint32_t, uint32_t> postCallSI;
```
greg-lunarg: Postponed pending "same block" discussion outcome.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+void InlinePass::GenInlineCode(
+    std::vector<std::unique_ptr<ir::BasicBlock>>* newBlocks,
+    std::vector<std::unique_ptr<ir::Instruction>>* newVars,
+    ir::UptrVectorIterator<ir::Instruction> call_ii,
+    ir::UptrVectorIterator<ir::BasicBlock> call_bi) {
+  // Map from callee id to caller id
+  std::unordered_map<uint32_t, uint32_t> callee2caller;
+  // Pre-call OpSampledImage Insts
+  std::unordered_map<uint32_t, ir::Instruction*> preCallSI;
```
greg-lunarg: Postponed pending "same block" discussion outcome.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+  // Post-call OpSampledImage Ids
+  std::unordered_map<uint32_t, uint32_t> postCallSI;
+
+  const uint32_t calleeId =
+      call_ii->GetOperand(kSpvFunctionCallFunctionId).words[0];
```
greg-lunarg: Done

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+    // find or create ptr to callee return type
+    uint32_t returnVarTypeId = FindPointerToType(calleeTypeId);
+    if (returnVarTypeId == 0) returnVarTypeId = AddPointerToType(calleeTypeId);
+    // Add return var to new function scope variables
+    returnVarId = TakeNextId();
+    std::vector<ir::Operand> in_operands;
+    in_operands.emplace_back(
+        spv_operand_type_t::SPV_OPERAND_TYPE_STORAGE_CLASS,
+        std::initializer_list<uint32_t>{uint32_t(SpvStorageClassFunction)});
+    std::unique_ptr<ir::Instruction> var_inst(new ir::Instruction(
```
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+        SpvOpVariable, returnVarTypeId, returnVarId, in_operands));
+    newVars->push_back(std::move(var_inst));
+  }
+
+  // Clone and map callee code
+  bool prevInstWasReturn = false;
```
greg-lunarg: Hm, I cannot think of the case where this is not up-to-date. It is initialized to false. It is only set to true when an OpReturn is seen. There are only two instructions that can follow: an OpLabel and an OpFunctionEnd. After OpLabel, it is set back to false (still up-to-date). I don't set it after OpFunctionEnd because then the iteration is complete and the variable is irrelevant. Right? Should I set it here anyway? Is there a case I am missing?

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+            prevInstWasReturn = true;
+          } break;
+          case SpvOpFunctionEnd: {
+            // if there was an early return, create return label/block
+            // if previous instruction was return, insert branch instruction
+            // to return block
+            if (returnLabelId != 0) {
+              if (prevInstWasReturn) AddBranch(returnLabelId, &bp);
+              newBlocks->push_back(std::move(bp));
+              const std::vector<ir::Operand> label_in_operands;
```
greg-lunarg: OK.

source/opt/inline_pass.cpp
greg-lunarg: PS. And I would love to get rid of the logic I already have for this. It's a major chunk of code for a case that rarely arises. Would be a lot cleaner and simpler without it.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+                          std::unique_ptr<ir::Instruction> samp_inst(
+                              new ir::Instruction(*inInst));
+                          const uint32_t rid = samp_inst->result_id();
+                          const uint32_t nid = this->TakeNextId();
+                          samp_inst->SetResultId(nid);
+                          postCallSI[rid] = nid;
+                          *iid = nid;
+                          bp->AddInstruction(std::move(samp_inst));
+                        }
+                      } else
```
greg-lunarg: Done.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+            spv_inst->ForEachInId([&callee2caller, &cpi, this](uint32_t* iid) {
+              const auto mapItr = callee2caller.find(*iid);
+              if (mapItr != callee2caller.end()) {
+                *iid = mapItr->second;
+              } else if (cpi->IsControlFlow()) {
+                const ir::Instruction* inst =
+                    def_use_mgr_->id_to_defs().find(*iid)->second;
+                if (inst->opcode() == SpvOpLabel) {
+                  // forward label reference. allocate a new label id, map it,
+                  // use
```
greg-lunarg: Fixed

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+            }
+            bp->AddInstruction(std::move(spv_inst));
+          } break;
+        }
+      });
+}
+
+bool InlinePass::Inline(ir::Function* func) {
+  bool modified = false;
+  for (auto bi = func->begin(); bi != func->end(); bi++) {
```
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+  for (auto& fn : *module_) {
+    id2function_[fn.GetResultId()] = &fn;
+    for (auto& blk : fn) {
+      id2block_[blk.GetLabelId()] = &blk;
+    }
+  }
+};
+
+Pass::Status InlinePass::ProcessImpl() {
+  // do exhaustive inlining on each entry point function in module
```
greg-lunarg: Yes. I think this will be a separate pass. Currently spirv-remap does a lot of dead code elimination (including dead function), although eventually I think that capability should migrate into spirv-opt.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+bool InlinePass::Inline(ir::Function* func) {
+  bool modified = false;
+  for (auto bi = func->begin(); bi != func->end(); bi++) {
+    for (auto ii = bi->begin(); ii != bi->end();) {
+      if (ii->opcode() == SpvOp::SpvOpFunctionCall) {
+        // Inline call
+        std::vector<std::unique_ptr<ir::BasicBlock>> newBlocks;
+        std::vector<std::unique_ptr<ir::Instruction>> newVars;
+        GenInlineCode(&newBlocks, &newVars, ii, bi);
+        // update block map given replacement blocks
```
greg-lunarg: OK.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+        for (auto& blk : newBlocks) {
```
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+          id2block_[blk->GetLabelId()] = &*blk;
+        }
+        // update phi functions in succesor blocks if call block
```
greg-lunarg: OK

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+        // is replaced with more than one block
```
greg-lunarg: On the first point, a slight correction: it is the label_id that survives the inlining (it is re-used in the first new block), but the original block is erased.

Are you suggesting a change here?

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+  }
+  return modified;
+}
+
+void InlinePass::Initialize(ir::Module* module) {
+  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+
+  // Initialize next unused Id
+  nextId_ = 0;
+  for (const auto& id_def : def_use_mgr_->id_to_defs()) {
```
greg-lunarg: I will make the change now.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+    }
+  }
+};
+
+Pass::Status InlinePass::ProcessImpl() {
+  // do exhaustive inlining on each entry point function in module
+  bool modified = false;
+  for (auto& e : module_->entry_points()) {
+    ir::Function* fn =
+        id2function_[e.GetOperand(kSpvEntryPointFunctionId).words[0]];
```
greg-lunarg: Done

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+    modified = modified || Inline(fn);
+  }
+
+  FinalizeNextId(module_);
```
greg-lunarg: OK. Will remove.

source/opt/iterator.h
```diff
@@ -81,6 +81,24 @@ class UptrVectorIterator
   inline typename std::enable_if<!IsConstForMethod, UptrVectorIterator>::type
   InsertBefore(Uptr value);
 
+  // Inserts the given |valueVector| to the position pointed to by this iterator
+  // and returns an iterator to the newly inserted |valueVector|.
```
greg-lunarg: OK

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,1071 @@
...
+"%foo2_f1_f1_ = OpFunction %float None %8",
+          "%f = OpFunctionParameter %_ptr_Function_float",
+         "%f2 = OpFunctionParameter %_ptr_Function_float",
+         "%12 = OpLabel",
+         "%19 = OpLoad %float %f",
+         "%20 = OpLoad %float %f2",
+         "%21 = OpFMul %float %19 %20",
+               "OpReturnValue %21",
+               "OpFunctionEnd",
+   "%foo_vf4_ = OpFunction %float None %15",
```
antiagainst: It helps to put the original GLSL source code for this assembly functions in comments.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,1071 @@
...
+               "OpReturnValue %45",
+               "OpFunctionEnd",
+      // clang-format on
+  };
+  SinglePassRunAndCheck<opt::InlinePass>(JoinAllInsts(before),
+                                                 JoinAllInsts(after),
+                                                 /* skip_nop = */ false);
+}
+
+TEST_F(InlineTest, InOut) {
```
antiagainst: I'd suggest to make this more clear. AFAICT, its InOutParameter, not InOutVariable.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,1071 @@
...
+         "%65 = OpLoad %float %r",
+               "OpReturnValue %65",
+               "OpFunctionEnd",
+      // clang-format on
+  };
+  SinglePassRunAndCheck<opt::InlinePass>(JoinAllInsts(before),
+                                                 JoinAllInsts(after),
+                                                 /* skip_nop = */ false);
+}
+
```
antiagainst: Consider test cases for

* Empty modules
* Modules without function definitions
* Modules in which all functions do not call other functions
* Recursive functions (calling self & calling each other) (Will this explode actually?)
* Caller and callee both accessing the same global variable
* Functions with `OpLine` & `OpNoLine`
* etc.

?

include/spirv-tools/optimizer.hpp
```diff
@@ -167,6 +167,10 @@ Optimizer::PassToken CreateUnifyConstantPass();
 // OpSpecConstantOp.
 Optimizer::PassToken CreateEliminateDeadConstantPass();
 
+// Creates a inline pass.
```
greg-lunarg: Done.

include/spirv-tools/optimizer.hpp
```diff
@@ -167,6 +167,10 @@ Optimizer::PassToken CreateUnifyConstantPass();
...
+// An inline pass exhaustively inlines all function calls.
+Optimizer::PassToken CreateInlinePass();
```
greg-lunarg: Done. Indeed, this pass only performs inlining on functions which are marked as entry points. More judicious application of inlining is left for future work.

source/opt/module.h
```diff
@@ -84,13 +84,19 @@ class Module {
   std::vector<Instruction*> GetConstants();
   std::vector<const Instruction*> GetConstants() const;
 
+  inline uint32_t id_bound() { return header_.bound; }
```
greg-lunarg: Done.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,1071 @@
...
+TEST_F(InlineTest, Simple) {
+  const std::vector<const char*> before = {
+      // clang-format off
+               "OpCapability Shader",
+          "%1 = OpExtInstImport \"GLSL.std.450\"",
+               "OpMemoryModel Logical GLSL450",
+               "OpEntryPoint Fragment %main \"main\" %BaseColor %gl_FragColor",
+               "OpExecutionMode %main OriginUpperLeft",
+               "OpSource GLSL 140",
+               "OpName %main \"main\"",
```
greg-lunarg: I think the names improve readability and allow for better visual comparison between the before and after. The names are created by default by glslang and I don't see a way to suppress them. It would require extra steps to strip them out. I sense that you agree that the names improve readability. Please let me know if you really want me to remove them.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,1071 @@
...
+               "OpName %param \"param\"",
+               "OpName %gl_FragColor \"gl_FragColor\"",
+       "%void = OpTypeVoid",
+          "%3 = OpTypeFunction %void",
+      "%float = OpTypeFloat 32",
+    "%v4float = OpTypeVector %float 4",
+"%_ptr_Function_v4float = OpTypePointer Function %v4float",
+          "%9 = OpTypeFunction %float %_ptr_Function_v4float",
+       "%uint = OpTypeInt 32 0",
+         "%14 = OpConstant %uint 0",
```
greg-lunarg: As you are probably aware, these disassemblies were created by spirv-dis.

test/opt/inline_test.cpp
antiagainst: Yes, they are quite useful to keep instructions readable. No need to remove. Just me ranted. :)

test/opt/inline_test.cpp
antiagainst: Yep. No need to take action here.

greg-lunarg: In order to coalesce shared sections of before and after, I had to reassemble and disassemble the befores. When I did, this issue went away.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,1071 @@
...
+         "%20 = OpLoad %float %19",
+         "%21 = OpFAdd %float %17 %20",
+               "OpReturnValue %21",
+               "OpFunctionEnd",
+      // clang-format on
+  };
+
+  const std::vector<const char*> after = {
+      // clang-format off
+               "OpCapability Shader",
```
greg-lunarg: Done. Besides isolating the changed sections, this shrank the file by 20%.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,1071 @@
...
+
+TEST_F(InlineTest, Nested) {
+  const std::vector<const char*> before = {
+      // clang-format off
+               "OpCapability Shader",
+          "%1 = OpExtInstImport \"GLSL.std.450\"",
+               "OpMemoryModel Logical GLSL450",
+               "OpEntryPoint Fragment %main \"main\" %BaseColor %gl_FragColor",
+               "OpExecutionMode %main OriginUpperLeft",
+               "OpSource GLSL 140",
```
greg-lunarg: Done.

source/opt/function.h
```diff
@@ -49,6 +49,12 @@ class Function {
   // Saves the given function end instruction.
   inline void SetFunctionEnd(std::unique_ptr<Instruction> end_inst);
 
+  // Returns function's id
+  inline uint32_t result_id() const { return def_inst_->result_id(); }
+
+  // Returns function's id
```
dneto0: Returns function's type id.

source/opt/inline_pass.h
```diff
@@ -0,0 +1,132 @@
...
+
+  const char* name() const override { return "inline"; }
+
+ private:
+  // Return the next available Id and increment it.
+  inline uint32_t TakeNextId() { return next_id_++; }
+
+  // Find pointer to type and storage in module, return its resultId,
+  // 0 if not found. TODO(greg-lunarg): Move this into type manager.
+  uint32_t FindPointerToType(uint32_t type_id, uint32_t storage_id);
```
dneto0: The "storage_id" parameter is actually being used as the literal value for StorageClass.  It's not an Id.  So the name is misleading.   

Additionally, you should tighten the type to SpvStorageClass.  Then it would be hard to misunderstand or use incorrectly.  That comes at the cost of some casting, but I think that's probably worth it.

source/opt/inline_pass.h
```diff
@@ -0,0 +1,132 @@
...
+
+  // Add pointer to type to module and return resultId.
+  uint32_t AddPointerToType(uint32_t type_id, uint32_t storage_id);
```
dneto0: Here as well:  uint32_t storage_id  -->  SpvStorageClass storage_class

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+    if (type_inst->opcode() == SpvOpTypePointer &&
+        type_inst->GetSingleWordOperand(kSpvTypePointerTypeId) == type_id &&
+        type_inst->GetSingleWordOperand(kSpvTypePointerStorageClass) ==
+            storage_id)
+      return type_inst->result_id();
+  }
+  return 0;
+}
+
+uint32_t InlinePass::AddPointerToType(uint32_t type_id, uint32_t storage_id) {
```
dneto0: uint32_t storage_id --> SpvStorageClass storage_class

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "inline_pass.h"
+
+// Indices of operands in SPIR-V instructions
+
+static const int kSpvEntryPointFunctionId = 1;
+static const int kSpvFunctionCallFunctionId = 2;
+static const int kSpvFuncitonCallArgumentId = 3;
```
dneto0: While we're here, let's fix spelling "Funciton" --> "Function".  :-)

source/opt/inline_pass.h
dneto0: I just reread the code in detail.  You're right.  Thanks!

dneto0: I think there's a tiny kernel that's still true:  The clone of the caller block retains its original ID.  The phi fixups in InlinePass::Inline rely on this, since the phi fixups only occur if newBlocks has at least two blocks.  But this is a pretty minor corner case.

source/opt/inline_pass.cpp
dneto0: I sympathize that this causes complexity in this algorithm.

The rules are deliberate.  Checking the SPIR-V spec changelog, I refer you to Khronos private Bugzilla bugs 14241 and 14683.  (I believe you have access).

Whenever there is a tradeoff in complexity between a SPIR-V generator or consumer, the tendency is to make the consumer simpler.  The consumer is typically an API implementation (driver), which helps avoid bugs and improve consistency across the ecosystem.

Regarding how to handle this, I think we're very much helped by the fact that the OpSampledImage and OpImage instructions are always in the same block.  So sequential order also implies dominance order.  So I think you can (a) keep *both* OpSampledImage and OpImage instructions in the same preCall and postCall collections, and keep the old-id to new-idm mapping up to date while you are rewriting them in the multi-block case.
(We'd need tests for this too.)

source/opt/inline_pass.cpp
dneto0: Ah, right.  My bad.

source/opt/inline_pass.cpp
dneto0: Right, the label_id of the caller blocks survives.  

Given there are other changes needed, I'd like the method comment on GenInlineCode to say that the last last block in new_blocks is the exit block of the inlined code.

tools/opt/opt.cpp
```diff
@@ -61,6 +61,8 @@ NOTE: The optimizer is a work in progress.
                e.g.: --set-spec-const-default-value "1:100 2:400"
   --unify-const
                Remove the duplicated constants.
+  --inline
+               Exhaustively inline all function calls
```
dneto0: Description is wrong:  It only inlines entry points.

Also, let's avoid confusion with LLVM's opt command.  There "-inline" means something else.
I prefer to rename this option  "--inline-entry-points" or even "--inline-entry-points-all"

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,430 @@
...
+        }
+        // Create first/next block.
+        new_blk_ptr.reset(new ir::BasicBlock(NewLabel(labelId)));
+        if (firstBlock) {
+          // Copy contents of original caller block up to call instruction.
+          for (auto cii = call_block_itr->begin(); cii != call_inst_itr;
+               cii++) {
+            std::unique_ptr<ir::Instruction> cp_inst(new ir::Instruction(*cii));
+            // Remember same-block ops for possible regeneration.
+            if (cp_inst->opcode() == SpvOpSampledImage) {
```
dneto0: Reiterating the need to handle OpImage similarly.  More details:  https://github.com/KhronosGroup/SPIRV-Tools/pull/534#discussion_r108220528

source/opt/inline_pass.cpp
greg-lunarg: Had to re-add this. At Lei's request I just added validation to the inline tests. As it turns out, the tests don't use the passmanager (at least the inline test doesn't) so validation was complaining about the stale id bound.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,1071 @@
...
+         "%28 = OpLoad %float %27",
+         "%29 = OpAccessChain %_ptr_Function_float %bar %uint_1",
+         "%30 = OpLoad %float %29",
+         "%31 = OpFAdd %float %28 %30",
+               "OpReturnValue %31",
+               "OpFunctionEnd",
+      // clang-format on
+  };
+  SinglePassRunAndCheck<opt::InlinePass>(JoinAllInsts(before),
+                                                 JoinAllInsts(after),
```
greg-lunarg: OK. I did this but now am wondering if you were suggesting this as a separate mod. Should I include it in this PR?

test/opt/inline_test.cpp
greg-lunarg: OK.

test/opt/inline_test.cpp
dneto0: I'd be ok adding TODOs for this in the test file, and adding the actual tests in subsequent CLs.

SPIR-V core allows recursion I believe all Khronos client APIs ban recursion.  I'd say that handling the recursive case can be deferred: file an enhancement issue for it.

test/opt/inline_test.cpp
dneto0: This is a really significant optimization, and its tests are complete valid modules.  I'd feel better if this was in this PR.  The inline tests should run validation as well.  Don't change the other existing tests.

test/opt/inline_test.cpp
greg-lunarg: Done.

test/opt/inline_test.cpp
greg-lunarg: OK. Thanks.

test/opt/inline_test.cpp
greg-lunarg: OK. Thanks.

source/opt/function.h
greg-lunarg: Done.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,837 @@
...
+namespace {
+
+using namespace spvtools;
+
+using InlineTest = PassTest<::testing::Test>;
+
+TEST_F(InlineTest, Simple) {
+  const std::vector<const char*> predefs = {
+      // clang-format off
+               "OpCapability Shader",
```
greg-lunarg: Done.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,837 @@
...
+"%_ptr_Input_v4float = OpTypePointer Input %v4float",
+  "%BaseColor = OpVariable %_ptr_Input_v4float Input",
+"%_ptr_Output_v4float = OpTypePointer Output %v4float",
+"%gl_FragColor = OpVariable %_ptr_Output_v4float Output",
+      // clang-format on
+  };
+
+  const std::vector<const char*> nonEntryFuncs = {
+      // clang-format off
+   "%foo_vf4_ = OpFunction %float None %15",
```
greg-lunarg: Done.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,837 @@
...
+        "%bar = OpFunctionParameter %_ptr_Function_v4float",
+         "%28 = OpLabel",
+          "%r = OpVariable %_ptr_Function_float Function",
+         "%29 = OpAccessChain %_ptr_Function_float %bar %uint_0",
+         "%30 = OpLoad %float %29",
+               "OpStore %r %30",
+         "%31 = OpLoad %float %r",
+         "%32 = OpFOrdLessThan %bool %31 %float_0",
+               "OpSelectionMerge %33 None",
```
greg-lunarg: Agreed.

source/opt/inline_pass.h
greg-lunarg: Done.

source/opt/inline_pass.h
greg-lunarg: Done.

source/opt/inline_pass.cpp
greg-lunarg: Done.

source/opt/inline_pass.cpp
greg-lunarg: At least I was consistent :)

source/opt/inline_pass.h
```diff
@@ -0,0 +1,132 @@
...
+      std::unique_ptr<ir::BasicBlock>* block_ptr);
+
+  // Return in new_blocks the result of inlining the call at call_inst_itr
+  // within its block at call_block_itr. The block at call_block_itr can
+  // just be replaced with the blocks in new_blocks. Any additional branches
+  // are avoided. Debug instructions are cloned along with their callee
+  // instructions. Early returns are replaced by a store to a local return
+  // variable and a branch to a (created) exit block where the local variable
+  // is returned. Formal parameters are trivially mapped to their actual
+  // parameters.
```
greg-lunarg: Done.

tools/opt/opt.cpp
greg-lunarg: OK. I chose --inline-entry-points-all.

test/opt/inline_test.cpp
```diff
@@ -0,0 +1,837 @@
+// Copyright (c) 2016 Google Inc.
```
greg-lunarg: Thanks :)

source/opt/inline_pass.cpp
greg-lunarg: Are you open to a recursive algorithm here? Essentially, I think I can implement this by only adding a recursive call to CloneSameBlockOps just after cloning the same block instruction, which would generate any same-block-ops needed for the cloned instruction. Otherwise I am going to split this algorithm into two loops, one to build up the stack of instructions to insert, and one to walk back down it.

greg-lunarg: I will check the recursive solution in.

source/opt/inline_pass.cpp
greg-lunarg: Done.

source/opt/inline_pass.cpp
greg-lunarg: Done.

source/opt/inline_pass.cpp
```diff
@@ -0,0 +1,439 @@
...
+      ->ForEachInId([&postCallSB, &preCallSB, &block_ptr, this](uint32_t* iid) {
+        const auto mapItr = (*postCallSB).find(*iid);
+        if (mapItr == (*postCallSB).end()) {
+          const auto mapItr2 = (*preCallSB).find(*iid);
+          if (mapItr2 != (*preCallSB).end()) {
+            // Clone pre-call same-block ops, map result id.
+            const ir::Instruction* inInst = mapItr2->second;
+            std::unique_ptr<ir::Instruction> sb_inst(
+                new ir::Instruction(*inInst));
+            CloneSameBlockOps(&sb_inst, postCallSB, preCallSB, block_ptr);
```
dneto0: This upgrade to handle OpImage is spot on. You made it look easy.  :-)

dneto0:
Ok, looks like the int to SpvStorageClass requires an explicit cast to satisfy MSVC.  I like constructor casts. :-)  
In the interest of reusing some of my rebase work, I'll post an MR with the squashed commit and the fix, then wait for bots again.

dneto0:
[TIL.  GitHub has a feature where users with write access can add new commits to a pull request branch. That would have saved me the trouble of making a new branch. :-)
]

dneto0:
Squashed, rebased, and some trivial conflicts resolved.
Content is now on master with commit 04fcc6674

Thank you!

greg-lunarg:
Thanks David and Lei!

Greg


Greg Fischer

Project Manager and Senior Engineer

LunarG, Inc. - 3D Driver Innovations

greg@lunarg.com <alan@lunarg.com>

source/opt/inline_pass.cpp
greg-lunarg: Done.

Issue #495, opened 2016-12-08
Verify that validator checks Logical Addressing rules.


dneto0:
Ensure all tests check the right error message is issued.
Older validator tests might only check the return value, which is weaker than we want.

s-perron:
We have made many improvements. Without specifics, it is hard to say if this is done. If something is still missing, please reopen.

PR #535, opened 2017-01-24
Making the unit tests more robust.

ehsannas:
It is best to check the error messages of unit tests that fail
validation. This will ensure that a validation failure is due to what we
expect and not due to some secondary reason.

In this CL, I have updated the tests in val_id_test.cpp.

### DIFF
--- a/test/val/val_id_test.cpp
+++ b/test/val/val_id_test.cpp
@@ -147,13 +147,13 @@ TEST_F(ValidateIdWithMessage, OpLineGood) {
 }
 TEST_F(ValidateIdWithMessage, OpLineFileBad) {
   string spirv = kGLSL450MemoryModel + R"(
-     OpLine %2 0 0
-%2 = OpTypeInt 32 0
-%3 = OpTypePointer Input %2
-%4 = OpVariable %3 Input)";
+  %1 = OpTypeInt 32 0
+     OpLine %1 0 0
+  )";
   CompileSuccessfully(spirv.c_str());
   EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
-  EXPECT_THAT(getDiagnosticString(), HasSubstr("ID 1 has not been defined"));
+  EXPECT_THAT(getDiagnosticString(),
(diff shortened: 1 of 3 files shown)


ehsannas:
Also addresses concerns in issue #495

test/val/val_id_test.cpp
```diff
@@ -148,6 +153,7 @@ TEST_F(ValidateIdWithMessage, OpLineFileBad) {
 %4 = OpVariable %3 Input)";
   CompileSuccessfully(spirv.c_str());
   EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(), HasSubstr("ID 1 has not been defined"));
```
dneto0: This is not the right message.  The message is an SSA problem.
But the actual bug is at line 150, since %2 is supposed to reference an OpString that names a file.

I don't even know why this message would be defined because I don't think anything references %1.

test/val/val_id_test.cpp
```diff
@@ -527,12 +555,13 @@ class OpTypeArrayLengthTest
   ~OpTypeArrayLengthTest() { spvDiagnosticDestroy(diagnostic_); }
 
   // Runs spvValidate() on v, printing any errors via spvDiagnosticPrint().
-  spv_result_t Val(const SpirvVector& v) {
+  spv_result_t Val(const SpirvVector& v, const char* expected_err = "") {
```
dneto0: In this case it's better to make this const std::string&  parameter.
You never allow it to be nullptr.  And you construct the std::string later anyway in the error case.  Hm. Are you trying to save the construction time?  I'm not worried about that kind of optimization in test code. In test code the balance shifts toward clarity over a small optimization like this.

test/val/val_id_test.cpp
ehsannas: 👍  good catch. I had actually read every single error message to make sure we're getting the correct message. Looks like I had missed this one.

(ID 1 is in fact %2... that's why the error is complaining that ID 1 is not defined).

dneto0:
Rebased, squashed, and pushed into master as affbc3184

Issue #489, opened 2016-12-08
Check universal rule for Functions


dneto0:
2.16.1:

"A function declaration (an OpFunction with no basic blocks), must have a Linkage Attributes Decoration with the Import Linkage Type.

A function definition (an OpFunction with basic blocks) cannot be decorated with the Import Linkage Type.

A function cannot have both a declaration and a definition (no forward declarations)."

dneto0:
Fixed by https://github.com/KhronosGroup/SPIRV-Tools/pull/544

PR #544, opened 2017-01-29
Fixes issue #489.

ehsannas:
From the SPIR-V Spec 2.16.1:
* A function declaration (an OpFunction with no basic blocks), must have
a Linkage Attributes Decoration with the Import Linkage Type.
* A function definition (an OpFunction with basic blocks) cannot be
decorated with the Import Linkage Type.

Also refactored the `ValidateDecorations` function and reused newly written functions.

### DIFF
--- a/source/validate_decorations.cpp
+++ b/source/validate_decorations.cpp
@@ -45,7 +45,8 @@ bool hasImportLinkageAttribute(uint32_t id, ValidationState_t& vstate) {
   return std::any_of(decorations.begin(), decorations.end(),
                      [](const Decoration& d) {
                        return SpvDecorationLinkageAttributes == d.dec_type() &&
-                              d.params().size() >= 2u && d.params().back() == 1;
+                              d.params().size() >= 2u &&
+                              d.params().back() == SpvLinkageTypeImport;
                      });
 }
 
--- a/source/validate_decorations.cpp
+++ b/source/validate_decorations.cpp
@@ -23,6 +23,7 @@
 
 using libspirv::Decoration;
 using libspirv::DiagnosticStream;
+using libspirv::Instruction;
 using libspirv::ValidationState_t;
 
 namespace {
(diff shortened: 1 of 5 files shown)


source/validate_decorations.cpp
```diff
@@ -38,33 +39,59 @@ bool isBuiltInStruct(uint32_t struct_id, ValidationState_t& vstate) {
...
 }
 
-}  // anonymous namespace
+// Returns true if the given ID has the Import LinkageAttributes decoration.
+bool hasImportLinkageAttribute(uint32_t id, ValidationState_t& vstate) {
+  const auto& decorations = vstate.id_decorations(id);
+  return std::any_of(decorations.begin(), decorations.end(),
+                     [](const Decoration& d) {
+                       return SpvDecorationLinkageAttributes == d.dec_type() &&
+                              d.params().size() >= 2u && d.params().back() == 1;
```
dneto0: That magic 1 should probably be the enum SpvLinkageTypeImport

test/val/val_decoration_test.cpp
```diff
@@ -360,5 +360,74 @@ TEST_F(ValidateDecorations, EntryPointFunctionHasLinkageAttributeBad) {
                 "an OpEntryPoint instruction."));
 }
 
+TEST_F(ValidateDecorations, FunctionDeclarationWithoutImportLinkageBad) {
```
dneto0: I'd like to see a test where we exercise the fact you got the "1" value right for the linkage attribute.
It's the same as this case, except you add an *Export* linkage decoration to the function.

ehsannas: Good idea 👍

dneto0:
Waiting for bots to go green.

dneto0:
Rebased, squashed, and pushed into master as 2046e244

Issue #551, opened 2017-02-09
Structured control flow: If a merge block is reachable, then it must be *strictly* dominated by its header block.


dneto0:

dneto0:
I'm taking this.

PR #552, opened 2017-02-09
Check strict domination of merge block

dneto0:
If a merge block is reachable, then it must be *strictly* dominated
by its header.  Until now we've allowed the header and the merge
block to be the same.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/551

Also: Use dominates and postdominates methods on BasicBlock to
improve readability.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,8 @@ v2016.7-dev 2017-01-06
    #508: Support compilation under CYGWIN
    #517: Fix validation when continue (or case) contstruct is also the head of a
      nested control construct.
+   #551: If a merge block is reachable, it must be *strictly* dominated by its
+     header.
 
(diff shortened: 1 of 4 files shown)


dneto0:
Whoops.  Test failures.  Will fix.

dneto0:
Ok. This passes internal tests.  Also, validation succeeds on all current Vulkan CTS tests.

sheredom:
LGTM

dneto0:
@ehsannas : Yeah. In this case a plain text test could have been easier to read.  But I wanted to follow the surrounding tests that use the more elaborate methodology.  Many of the other cases are rather more involved, and arguably more readable the way they are.

dneto0:
Rebased

Issue #492, opened 2016-12-08
Check Kernel validation rules


dneto0:
"The Signedness in OpTypeInt must always be 0."

dneto0:
PR #553 has been squashed and pushed into master

PR #553, opened 2017-02-09
The Signedness in OpTypeInt must always be 0.

ehsannas:
When Kernel capability is used, The Signedness in OpTypeInt must always
be 0. Fixes issue #492.

### DIFF
--- a/test/val/val_capability_test.cpp
+++ b/test/val/val_capability_test.cpp
@@ -1492,6 +1492,7 @@ OpMemoryModel Logical OpenCL
 %i32    = OpTypeInt 32 0
 )";
   CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_SUCCESS, ValidateInstructions());
 }
 
 TEST_F(ValidateCapability, IntSignednessKernelBad) {
@@ -1517,6 +1518,7 @@ OpMemoryModel Logical GLSL450
 %i32    = OpTypeInt 32 1
 )";
   CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_SUCCESS, ValidateInstructions());
 }
 
 }  // namespace anonymous
--- a/test/val/val_capability_test.cpp
+++ b/test/val/val_capability_test.cpp
@@ -37,6 +37,7 @@ using std::string;
 using std::tuple;
 using std::vector;
 using testing::Combine;
+using testing::HasSubstr;
 using testing::Values;
 using testing::ValuesIn;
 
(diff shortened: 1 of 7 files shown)


test/val/val_capability_test.cpp
```diff
@@ -812,207 +813,207 @@ INSTANTIATE_TEST_CASE_P(Decoration, ValidateCapability,
 make_pair(string(kOpenCLMemoryModel) +
           "OpEntryPoint Kernel %func \"compute\" \n"
           "OpDecorate %intt RelaxedPrecision\n"
-          "%intt = OpTypeInt 32 1\n" + string(kVoidFVoid),
+          "%intt = OpTypeInt 32 0\n" + string(kVoidFVoid),
```
dneto0: Whoops sorry about that.  :-)

test/val/val_capability_test.cpp
```diff
@@ -1483,4 +1484,39 @@ OpFunctionEnd
...
 }
 
+TEST_F(ValidateCapability, IntSignednessKernelGood) {
+  const std::string spirv = R"(
+OpCapability Kernel
+OpCapability Linkage
+OpMemoryModel Logical OpenCL
+%i32    = OpTypeInt 32 0
+)";
+  CompileSuccessfully(spirv);
```
dneto0: Gotta validate it too.  :-)

test/val/val_capability_test.cpp
```diff
@@ -1483,4 +1484,39 @@ OpFunctionEnd
...
+
+TEST_F(ValidateCapability, IntSignednessShaderGood) {
+  const std::string spirv = R"(
+OpCapability Shader
+OpCapability Linkage
+OpMemoryModel Logical GLSL450
+%u32    = OpTypeInt 32 0
+%i32    = OpTypeInt 32 1
+)";
+  CompileSuccessfully(spirv);
```
dneto0: Must validate it too.

test/val/val_id_test.cpp
```diff
@@ -537,9 +537,16 @@ enum Signed { kSigned, kUnsigned };
 // Creates an assembly snippet declaring OpTypeArray with the given length.
 string MakeArrayLength(const string& len, Signed isSigned, int width) {
   ostringstream ss;
-  ss << kGLSL450MemoryModel;
-  ss << " %t = OpTypeInt " << width << (isSigned == kSigned ? " 1" : " 0")
-     << " %l = OpConstant %t " << len << " %a = OpTypeArray %t %l";
+  ss << R"(
+    OpCapability Shader
+    OpCapability Linkage
```
dneto0: We might run into problems with this later because Vulkan doesn't allow Linkage.  But let's address that once it becomes a problem.  I know you did this because you don't want to have to name an entry point.

test/val/val_capability_test.cpp
ehsannas: Oops :-)

dneto0:
Squashed and pushed into master as f2867d7485

Issue #549, opened 2017-02-07
Add permissive mode which accepts nearly anything if an unrecognized extension is used. Uncertain whether this mode should be the default.


dneto0:

ehsannas:
@dneto0 how do we know if an extension is "unrecognized" ?

dneto0:
"Recognized by the validator" = there is code in the validator to accommodate that extension's changes to the validation rules.
Currently no extensions are recognized.  But there are extensions published on the SPIR-V registry:
https://www.khronos.org/registry/spir-v/

For example `SPV_KHR_shader_ballot` says its new features (a few more instructions, and a few enums) are enabled by a new capability, and that capabilty should only be recognized by current environments only when we additionally also see OpExtension "SPV_KHR_shader_ballot".   

Thankfully, the grammar file already is set up to automatically set up the dependency checks from enums to capabilities.
So, for example consider this example:
```
        OpCapability Shader
        OpCapability SubgroupBallotKHR
        OpMemoryModel Logical GLSL450
        OpEntryPoint Vertex %foo "foo"

%void = OpTypeVoid
%voidfn = OpTypeFunction %void
%bool = OpTypeBool
%true = OpConstantTrue %bool

%foo =  OpFunction %void None %voidfn

%entry = OpLabel
   %v = OpSubgroupBallotKHR %void  %true ; Right, shouldn't be void, but that's beside the point right now
        OpReturn

        OpFunctionEnd
```

The validator *allows* OpSubgroupBallotKHR to be used because it sees the corresponding capability SubgroupBallotKHR.  But if you comment out that capability declaration, then the validator complains, saying:

```
error: 10: Opcode SubgroupBallotKHR requires one of these capabilities: SubgroupBallotKHR
```
Excellent!

So the remaining piece is:  Unless an environment specifically has brought a particular capability into its core functionality, then use of that capability must be accompanied by a matching OpExtension instruction.    None of the current target environments include SPV_KHR_shader_ballot in core.  But a future environment, e.g. Vulkan20.0 could.

dneto0:
Note that the JSON grammar file encodes the dependency from the capability to the extension name.
E.g. for SPV_KHR_shader_ballot: https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/1.0/spirv.core.grammar.json#L5101 

So we'd just need some general purpose code to use that in a check.

dneto0:
Debate about how to expose this at the API level:
[...]

PR #554, opened 2017-02-13
Permissive mode

ehsannas:
Fixes issue #549

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -370,7 +370,7 @@ typedef enum {
 const char* spvTargetEnvDescription(spv_target_env env);
 
 // Creates a context object.  Returns null if env is invalid.
-spv_context spvContextCreate(spv_target_env env);
+spv_context spvContextCreate(spv_target_env env, bool permissive = false);
 
 // Destroys the given context object.
 void spvContextDestroy(spv_context context);
--- a/source/table.h
+++ b/source/table.h
@@ -31,6 +31,9 @@ enum class Extension {
 
 using ExtensionSet = EnumSet<Extension>;
 
+// Returns true if the input is a known extension.
+bool isValidExtensionName(std::string ext_name);
+
 }  // namespace libspirv
 
 typedef struct spv_opcode_desc_t {
@@ -106,6 +109,7 @@ struct spv_context_t {
   const spv_opcode_table opcode_table;
   const spv_operand_table operand_table;
   const spv_ext_inst_table ext_inst_table;
+  bool permissive;
   spvtools::MessageConsumer consumer;
 };
 
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -249,8 +249,6 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
   // If unrecognized extensions are used and 'permissive' mode is enabled,
   // consider the module valid.
   if (context.permissive) {
-    // Go over the given extensions and figure out if any unrecognized
-    // extensions are used.
     for (uint64_t instIndex = 0;
          instIndex < instructions.size() &&
          SpvOpMemoryModel != instructions[instIndex].opcode;
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -22,6 +22,7 @@
 #include <iterator>
 #include <sstream>
 #include <string>
+#include <unordered_set>
 #include <vector>
 
 #include "binary.h"
@@ -32,6 +33,7 @@
 #include "spirv-tools/libspirv.h"
 #include "spirv_constant.h"
 #include "spirv_endian.h"
+#include "table.h"
 #include "val/construct.h"
 #include "val/function.h"
 #include "val/validation_state.h"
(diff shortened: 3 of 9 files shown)


include/spirv-tools/libspirv.h
```diff
@@ -370,7 +370,7 @@ typedef enum {
 const char* spvTargetEnvDescription(spv_target_env env);
 
 // Creates a context object.  Returns null if env is invalid.
-spv_context spvContextCreate(spv_target_env env);
+spv_context spvContextCreate(spv_target_env env, bool permissive = false);
```
dneto0: This breaks backward compatibility.  Also, this is supposed to be a C API, so default options should not be used.

The "permissive" mode should only apply to validation, so it really should be only on the validation API.
I think the right answer is to add a new validation entry point.  Question is whether it's hardcoded to be permissive, or whether we're adding  a general way to add more kinds of options other than just allow more when unrecognized extensions are given.
Early on (before first public release) the validator had an options word with bits for various kinds of checks, e.g. ID checks vs. others.  We got rid of that because it was hard to use: only an insider really knew what checks were turned on for what bits.  But that could work again if we're careful about what options are exposed.
We'd have two:   default = check everything we can, and relaxed-when-we-see-an-unrecognized-extension.  If we can think of another coarse-grain option then this might be the right way to go.

There are two entry points right now spvValidate and spvValidateBinary.   I'd make a new entry point based on the second one.  e.g. spvValidateBinaryWithOptions

source/table.h
```diff
@@ -31,6 +31,9 @@ enum class Extension {
 
 using ExtensionSet = EnumSet<Extension>;
 
+// Returns true if the input is a known extension.
+bool isValidExtensionName(std::string ext_name);
```
dneto0: Suggest Valid -> Known

source/table.h
```diff
@@ -106,6 +109,7 @@ struct spv_context_t {
   const spv_opcode_table opcode_table;
   const spv_operand_table operand_table;
   const spv_ext_inst_table ext_inst_table;
+  bool permissive;
```
dneto0: Need a better comment on this.
Permissive here means: binary parsing must still succeed (and hence we must have a valid instruction table for it), but then no validation rules are checked.

source/validate.cpp
```diff
@@ -66,6 +68,14 @@ spv_result_t spvValidateIDs(const spv_instruction_t* pInsts,
   return SPV_SUCCESS;
 }
 
+// Returns true if the input is a known extension.
+// TODO: The list of recognized extensions should be populated atuomatically.
```
dneto0: But from what.   What does a maintainer do when a new extension comes along?  Do they add the string here first, and then implement the extensions checks?  Should say so.

source/validate.cpp
```diff
@@ -215,13 +226,52 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
...
+  // If unrecognized extensions are used and 'permissive' mode is enabled,
+  // consider the module valid.
+  if (context.permissive) {
+    for (uint64_t instIndex = 0;
+         instIndex < instructions.size() &&
+         SpvOpMemoryModel != instructions[instIndex].opcode;
+         ++instIndex) {
+      auto instr = instructions[instIndex];
+      if (SpvOpExtension == instr.opcode) {
+        std::string extension_name =
```
dneto0: const

test/val/val_state_test.cpp
```diff
@@ -84,4 +88,30 @@ TEST_F(ValidationState_HasAnyOfTest, MultiCapMask) {
...
+
+  // This should fail because the above SPIR-V module is missing an
+  // OpEntryPoint.
+  bool permissive = false;
+  EXPECT_EQ(SPV_ERROR_INVALID_BINARY,
+            ValidateInstructions(SPV_ENV_UNIVERSAL_1_0, permissive));
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("No OpEntryPoint instruction was found."));
+
+  // This should pass because we are using the permissive mode, and we are using
```
dneto0: Really should be a different test case.

include/spirv-tools/libspirv.h
ehsannas: if there's only 1 option, then there's no need to create an options word and use spvValidateBinaryWithOptions. We can just add spvValidateBinaryPermissive. 
(i.e. 
spvValidateBinary = the existing function would be the default
spvValidateBinaryPermissive = the permissive version of the validator)

I guess my question is: do we think there will be more command line options for the Validator executable in the future?

source/validate.cpp
```diff
@@ -215,13 +226,52 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
...
+    spvInstructionCopy(&binary->code[index], static_cast<SpvOp>(opcode),
+                       wordCount, endian, &inst);
+    instructions.push_back(inst);
+    index += wordCount;
+  }
 
-  if (vstate->in_function_body())
+  // If unrecognized extensions are used and 'permissive' mode is enabled,
+  // consider the module valid.
+  if (context.permissive) {
```
dneto0: You need more than this anyway.  The full spvBinaryParse is done above at line 229, and that will do what it calls "inline checks".  If you really want to exit early, then you'll have to put the check for unknown extension in the middle of that parsing and set a bit in the validation state.  Also, the ProcessInstruction function should exit early with success if it sees that bit set.

include/spirv-tools/libspirv.h
antiagainst: If we need to add a new entry point anyway, I'd prefer to have that new entry point designed as an extensible one. It's hard to predict the future; if we indeed need another entry point in the future, introducing yet another entry point is not optimal, especially if the new variant is orthogonal to the existing options.

Issue #548, opened 2017-02-07
Use of OpImageSparseSampleProj* is invalid


dneto0:

dneto0:
Fixed by https://github.com/KhronosGroup/SPIRV-Tools/pull/556

PR #556, opened 2017-02-15
Fixes issue #548

adamvy:
Add validation checks for the reserved OpImageSparseSampleProj* opcodes.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -54,7 +54,6 @@ typedef enum spv_result_t {
   SPV_ERROR_INVALID_LAYOUT = -12,
   SPV_ERROR_INVALID_CAPABILITY = -13,
   SPV_ERROR_INVALID_DATA = -14,  // Indicates data rules validation failure.
-  SPV_ERROR_RESERVED_OPCODE = -15,
   SPV_FORCE_32_BIT_ENUM(spv_result_t)
 } spv_result_t;
 
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -54,6 +54,7 @@ typedef enum spv_result_t {
   SPV_ERROR_INVALID_LAYOUT = -12,
   SPV_ERROR_INVALID_CAPABILITY = -13,
   SPV_ERROR_INVALID_DATA = -14,  // Indicates data rules validation failure.
+  SPV_ERROR_RESERVED_OPCODE = -15,
   SPV_FORCE_32_BIT_ENUM(spv_result_t)
 } spv_result_t;
 
--- a/test/val/val_instructions_test.cpp
+++ b/test/val/val_instructions_test.cpp
@@ -1,0 +1,158 @@
+// Copyright (c) 2017 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Validation tests for illegal instructions
+
+#include <sstream>
+#include <string>
+#include <tuple>
+
+#include "gmock/gmock.h"
+#include "val_fixtures.h"
+
+using ::testing::HasSubstr;
+
+using ValidateIns = spvtest::ValidateBase<std::string>;
+
+namespace {
+
+TEST_F(ValidateIns, OpImageSparseSampleProjImplicitLod) {
+  char str[] = R"(
+             OpCapability Shader
+             OpCapability Linkage
+             OpCapability Sampled1D
+             OpCapability SparseResidency
+             OpMemoryModel Logical GLSL450
+%void      = OpTypeVoid
+%int       = OpTypeInt 32 0
+%float     = OpTypeFloat 32
+%floatp    = OpTypePointer Input %float
+%fnt       = OpTypeFunction %void
+%imgt      = OpTypeImage %float 1D 0 0 0 0 Unknown
+%sampledt  = OpTypeSampledImage %imgt
+%sampledp  = OpTypePointer Uniform %sampledt
+%img       = OpVariable %sampledp Uniform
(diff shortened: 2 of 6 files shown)


adamvy:
I'm not sure I put all the pieces in the ideal spots, I'm open to suggestion for where these bits should live.

Thanks!

include/spirv-tools/libspirv.h
```diff
@@ -54,6 +54,7 @@ typedef enum spv_result_t {
   SPV_ERROR_INVALID_LAYOUT = -12,
   SPV_ERROR_INVALID_CAPABILITY = -13,
   SPV_ERROR_INVALID_DATA = -14,  // Indicates data rules validation failure.
+  SPV_ERROR_RESERVED_OPCODE = -15,
```
dneto0: I would have gone for a reuse of INVALID_VALUE, interpreting the opcode as a value.
But this works too.
Most new validation rules don't need a new enum value here.  It's not clear what is gained by very fine grained error codes.

test/val/val_instructions_test.cpp
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) 2017 The Khronos Group Inc.
```
dneto0: Prefer Google Inc. here since you wrote this whole thing as work product for Google.
But this is acceptable too.

test/val/val_instructions_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+%void      = OpTypeVoid
+%int       = OpTypeInt 32 0
+%float     = OpTypeFloat 32
+%floatp    = OpTypePointer Input %float
+%fnt       = OpTypeFunction %void
+%imgt      = OpTypeImage %float 1D 0 0 0 0 Unknown
+%sampledt  = OpTypeSampledImage %imgt
+%sampledp  = OpTypePointer Uniform %sampledt
+%img       = OpVariable %sampledp Uniform
+%coord     = OpVariable %floatp Input
```
dneto0: For a small reduction in example side, you could have made this 
```
%coord = OpConstantNull %float
```
Then you wouldn't need the %floatp type.  :-)

test/val/val_instructions_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+using ::testing::HasSubstr;
+
+using ValidateIns = spvtest::ValidateBase<std::string>;
+
+namespace {
+
+TEST_F(ValidateIns, OpImageSparseSampleProjImplicitLod) {
+  char str[] = R"(
+             OpCapability Shader
+             OpCapability Linkage
```
dneto0: This is ok. To make a slightly more realistic test, follow Vulkan rules:  In Vulkan, the Linkage capability is not allowed. Instead, you'd make an OpEntryPoint for the function.  That would also list the Input variable operand. (But you could get rid of that with another trick I give below.)

test/val/val_instructions_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+#include "gmock/gmock.h"
+#include "val_fixtures.h"
+
+using ::testing::HasSubstr;
+
+using ValidateIns = spvtest::ValidateBase<std::string>;
+
+namespace {
+
```
dneto0: High level thing to fix: There's a lot of repetition here between the three cases.

It's preferable to make a function to return the std::string containing the common preamble assembly text: everything up to the actual instruction being tested.  That will shorten the test, and keep each focused on what's actually different about it.

(Google style guide forbids non-POD variables with static duration.  So just use a function to return the std::string).

Alternately, you can use  gtest parameterized test infrastructure to present a bunch of cases as a collection of structs.  E.g. see the TEST_P and INSTANTIATE_TEST_CASE_P at https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/text_to_binary.constant_test.cpp#L33
In that case you'd have to make sure to capture the differences in the operands for the various instructions.  So a test case is both the instruction name, and its different operand lists.

test/val/val_instructions_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+             OpCapability Linkage
+             OpCapability Sampled1D
+             OpCapability SparseResidency
+             OpMemoryModel Logical GLSL450
+%void      = OpTypeVoid
+%int       = OpTypeInt 32 0
+%float     = OpTypeFloat 32
+%floatp    = OpTypePointer Input %float
+%fnt       = OpTypeFunction %void
+%imgt      = OpTypeImage %float 1D 0 0 0 0 Unknown
```
dneto0: Vulkan images tend not to be Unknown format.
But this is ok.

include/spirv-tools/libspirv.h
adamvy: Sounds good.  INVALID_VALUE sounded to me like it would be used in cases where a constant was initialized to the wrong type or something.

Updated to use INVALID_VALUE instead.

test/val/val_instructions_test.cpp
adamvy: Changed to Google Inc.  Wasn't sure what the policy was for this repo.

test/val/val_instructions_test.cpp
adamvy: Makes sense, I refactored it to use TEST_P and INSTANTIATE_TEST_CASE_P now that I understand what they do.

Right now I'm using a std::pair as the test parameter and the operands are represented just as a second string.  Not sure if I should have made it a vector that I join, but a second string seems to work well enough.  There's also a bit of duplication left as the first 3 operands are the same for each case, but I feel like it would be less readable to move those out.

test/val/val_instructions_test.cpp
adamvy: Changed to use OpEntryPoint and the image is left as an Input variable, but the other variables were changed to constants.

test/val/val_instructions_test.cpp
adamvy: Updated to Rgba32f :)

Also changed it to 2D so I can drop the OpCapability Sampled1D

test/val/val_instructions_test.cpp
adamvy: Made most of the inputs OpConstantNull now, and removed the %floatp type

adamvy:
Should be ready for another look now.

test/val/val_instructions_test.cpp
dneto0: Your new commit strikes a good balance.  Simplicity and readability wins in test code, even at the cost of a little duplication.  (Yes, we make judgement calls all over the place.)

dneto0:
Squashed and pushed into master as  c0949703b

Issue #557, opened 2017-02-16
Add a validator entry point that takes an opaque options object.


dneto0:
This is a stepping stone for supporting:
- parameterization of limit checks.  See issue #500
- adding a mode to provide as many error messages as possible

Should be similar to the shaderc_compiler_options object and its uses.  https://github.com/google/shaderc/blob/master/libshaderc/include/shaderc/shaderc.h#L217

ehsannas:
I have some code that I can modify to serve this case

dneto0:
Fixed by https://github.com/KhronosGroup/SPIRV-Tools/pull/558

Issue #500, opened 2016-12-08
Parameterize the validator's limit checks.


dneto0:
2.17 says:
"Validators must either inform when these limits are crossed, or be explicitly parameterized with larger limits."

We need to design the API for parameterizing the limits, and implement it.

dneto0:
Implementation should depend on https://github.com/KhronosGroup/SPIRV-Tools/issues/557

dneto0:
Fixed by https://github.com/KhronosGroup/SPIRV-Tools/pull/558

PR #558, opened 2017-02-17
Validator Options

ehsannas:
* Add a validator entry point that takes an opaque options object. (issue #557 )
* To demonstrate usage, one options has been added which aligns with issue #500 (Parameterize the validator's limit checks)

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -378,6 +378,7 @@ typedef enum {
   validator_limit_max_global_variables,
   validator_limit_max_switch_branches,
   validator_limit_max_function_args,
+  validator_limit_max_control_flow_nesting_depth,
 } spv_validator_limit;
 
 // Returns a string describing the given SPIR-V target environment.
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -376,6 +376,7 @@ typedef enum {
   validator_limit_max_struct_depth,
   validator_limit_max_local_variables,
   validator_limit_max_global_variables,
+  validator_limit_max_switch_branches,
 } spv_validator_limit;
 
 // Returns a string describing the given SPIR-V target environment.
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -377,6 +377,7 @@ typedef enum {
   validator_limit_max_local_variables,
   validator_limit_max_global_variables,
   validator_limit_max_switch_branches,
+  validator_limit_max_function_args,
 } spv_validator_limit;
 
 // Returns a string describing the given SPIR-V target environment.
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -373,6 +373,7 @@ typedef enum {
 // SPIR-V Validator can be parameterized with the following Universal Limits.
 typedef enum {
   validator_limit_max_struct_members,
+  validator_limit_max_struct_depth,
   validator_limit_max_local_variables,
   validator_limit_max_global_variables,
 } spv_validator_limit;
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -379,6 +379,7 @@ typedef enum {
   validator_limit_max_switch_branches,
   validator_limit_max_function_args,
   validator_limit_max_control_flow_nesting_depth,
+  validator_limit_max_access_chain_indexes,
 } spv_validator_limit;
 
 // Returns a string describing the given SPIR-V target environment.
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -427,8 +427,9 @@ spv_result_t spvValidate(const spv_const_context context,
                          const spv_const_binary binary,
                          spv_diagnostic* diagnostic);
 
-// Validates a SPIR-V binary for correctness. Any errors will be written into
-// *diagnostic if diagnostic is non-null.
+// Validates a SPIR-V binary for correctness. Uses the provided Validator
+// options. Any errors will be written into *diagnostic if diagnostic is
+// non-null.
 spv_result_t spvValidateWithOptions(const spv_const_context context,
                                     const spv_const_validator_options options,
                                     const spv_const_binary binary,
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -327,6 +327,8 @@ typedef struct spv_diagnostic_t {
 // Its object is used by various translation API functions.
 typedef struct spv_context_t spv_context_t;
 
+typedef struct spv_validator_options_t spv_validator_options_t;
+
 // Type Definitions
 
 typedef spv_const_binary_t* spv_const_binary;
@@ -336,6 +338,8 @@ typedef spv_position_t* spv_position;
 typedef spv_diagnostic_t* spv_diagnostic;
 typedef const spv_context_t* spv_const_context;
 typedef spv_context_t* spv_context;
+typedef spv_validator_options_t* spv_validator_options;
+typedef const spv_validator_options_t* spv_const_validator_options;
 
 // Platform API
 
@@ -375,6 +379,21 @@ spv_context spvContextCreate(spv_target_env env);
 // Destroys the given context object.
 void spvContextDestroy(spv_context context);
 
+// Creates a Validator options object with default options.
+spv_validator_options spvValidatorOptionsCreate();
+
+// Destroys the given Validator options object.
+void spvValidatorOptionsDestroy(spv_validator_options options);
+
+// Records the maximum number of structure members that is considered valid in
+// the given Validator options object. <options> argument must be non-null.
+void spvValidatorOptionsSetMaxStructMembers(spv_validator_options options,
+                                            const char* limit);
+
+// Returns the maximum number of structure members that is considered valid in
+// the given Validator options object. <options> argument must be non-null.
+int spvValidatorOptionsGetMaxStructMembers(spv_const_validator_options options);
+
 // Encodes the given SPIR-V assembly text to its binary representation. The
 // length parameter specifies the number of bytes for text. Encoded binary will
 // be stored into *binary. Any error will be written into *diagnostic if
@@ -407,6 +426,13 @@ void spvBinaryDestroy(spv_binary binary);
 spv_result_t spvValidate(const spv_const_context context,
                          const spv_const_binary binary,
                          spv_diagnostic* diagnostic);
+
+// Validates a SPIR-V binary for correctness. Any errors will be written into
+// *diagnostic if diagnostic is non-null.
+spv_result_t spvValidateWithOptions(const spv_const_context context,
+                                    const spv_const_validator_options options,
+                                    const spv_const_binary binary,
+                                    spv_diagnostic* diagnostic);
 
 // Validates a raw SPIR-V binary for correctness. Any errors will be written
 // into *diagnostic if diagnostic is non-null.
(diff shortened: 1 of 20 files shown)


ehsannas:
Looks like Android doesn't know std::stoi 😞 
Build failure is when BUILD_NDK=ON

include/spirv-tools/libspirv.h
```diff
@@ -375,6 +379,21 @@ spv_context spvContextCreate(spv_target_env env);
 // Destroys the given context object.
 void spvContextDestroy(spv_context context);
 
+// Creates a Validator options object with default options.
+spv_validator_options spvValidatorOptionsCreate();
+
+// Destroys the given Validator options object.
+void spvValidatorOptionsDestroy(spv_validator_options options);
+
+// Records the maximum number of structure members that is considered valid in
```
dneto0: Each new limit would imply another entry point.  We can do better by having a single "set limit" method, which takes an enum.

Follow the example in shaderc: https://github.com/google/shaderc/blob/master/libshaderc/include/shaderc/shaderc.h#L345 

The only question is what type the number should have. Either int or uint32_t.  Shaderc went with int.
The universal limits table in the spec shows that a 32 bit signed int is sufficient.  I could go either way.
So far we have no use for negative limits, so I'd probably lean toward uint32_t.
If we ever need a 64bit limit then we can add another entry point.

include/spirv-tools/libspirv.h
```diff
@@ -375,6 +379,21 @@ spv_context spvContextCreate(spv_target_env env);
...
+// the given Validator options object. <options> argument must be non-null.
+void spvValidatorOptionsSetMaxStructMembers(spv_validator_options options,
+                                            const char* limit);
+
+// Returns the maximum number of structure members that is considered valid in
```
dneto0: This does not need to be in the public API.
In general getters on options are not required, and arguably only pollute the interface.

include/spirv-tools/libspirv.h
```diff
@@ -375,6 +379,21 @@ spv_context spvContextCreate(spv_target_env env);
...
+// the given Validator options object. <options> argument must be non-null.
```
dneto0: "must be non-null"... and not a dangling pointer. :-)
How about "must be a valid options object".  And then in the creation method you say "returns a valid options object. The object remains vaild until it is passed into spvValidatorOptionsDestroy".

source/table.cpp
```diff
@@ -47,6 +48,34 @@ spv_context spvContextCreate(spv_target_env env) {
...
+
+void spvValidatorOptionsSetMaxStructMembers(spv_validator_options options,
+                                            const char* limit) {
+  assert(options && "Validator options object may not be Null");
+  if (limit) {
+    int limit_int;
+    int success = sscanf(limit, "%d", &limit_int);
+    // The Minimum limits are specified in the SPIR-V Spec, so we only apply an
+    // increase in the limit.
+    if (success && limit_int > options->max_struct_members) {
```
dneto0: I think set really means set.  Don't check with the existing limit.
The user is expressing explicit intent.

source/table.cpp
```diff
@@ -47,6 +48,34 @@ spv_context spvContextCreate(spv_target_env env) {
...
+spv_validator_options spvValidatorOptionsCreate() {
+  return new spv_validator_options_t;
+}
+
+void spvValidatorOptionsDestroy(spv_validator_options options) {
+  delete options;
+}
+
+void spvValidatorOptionsSetMaxStructMembers(spv_validator_options options,
+                                            const char* limit) {
```
dneto0: Should take an integer argument directly.  Don't parse a string in this API.
(And this method should be enum-driven. See above.)

source/validate.cpp
```diff
@@ -299,23 +299,42 @@ spv_result_t spvValidateBinary(const spv_const_context context,
   }
 
   // Create the ValidationState using the context.
-  ValidationState_t vstate(&hijack_context);
+  // This interface is used for default command line options.
+  ValidationState_t vstate(&hijack_context, nullptr);
```
dneto0: Isn't this dangerous?    What parts of the code assume a non-null options object?
Wouldn't it be better to create an options object that we destroy at the end of this scope?

source/val/validation_state.h
```diff
@@ -290,11 +294,14 @@ class ValidationState_t {
   bool IsStructTypeWithBuiltInMember(uint32_t id) const {
     return (builtin_structs_.find(id) != builtin_structs_.end());
   }
+
  private:
   ValidationState_t(const ValidationState_t&);
 
   const spv_const_context context_;
 
+  const spv_const_validator_options options_;
```
dneto0: Is this always a valid options object?  Must say so one way or the other.
Probably safer to assume it always is.
The constructor should check.

tools/val/val.cpp
```diff
@@ -33,22 +33,31 @@ or if the filename is "-", then the binary is read from standard input.
...
 int main(int argc, char** argv) {
   const char* inFile = nullptr;
+  const char* max_struct_members = nullptr;
   spv_target_env target_env = SPV_ENV_UNIVERSAL_1_1;
 
   for (int argi = 1; argi < argc; ++argi) {
     const char* cur_arg = argv[argi];
     if ('-' == cur_arg[0]) {
-      if (0 == strcmp(cur_arg, "--version")) {
+      if (0 == strcmp(cur_arg, "--max_struct_members")) {
```
dneto0: Consider --limit option to generalize.  See glslc's -flimit=' settings '  ?

dneto0: Or maybe that's too complex.  Shader compilation has about 50 limit settings. SPIR-V only about 15.

include/spirv-tools/libspirv.h
ehsannas: My intention was to have the spv_validator_options structure that holds _all_ command line arguments (not just limits). Therefore, there would be no need to create any new entry point.

[Any new command line option can be added as a member of the struct.](https://github.com/KhronosGroup/SPIRV-Tools/pull/558/files/6b36d00538394091b5f8f11041948576bf2358f3#diff-99eab7706bbdca4c52fcc674bd6710a3R113)

Does that make sense? or did I miss your point?

include/spirv-tools/libspirv.h
ehsannas: :+1:

source/table.cpp
ehsannas: Agreed. :+1:

source/validate.cpp
ehsannas: Yes, you're right. :+1:

tools/val/val.cpp
dneto0: We talked about this offline.  I'm less concerned about independent options on the command line.
Using a regular structure like `--max-struct-members` lets the code switch on the prefix "--max-" and then handle different limits in a table-driven manner because you'd call a common underlying API.  You'd just need a table mapping the suffix string to the limit enum.

Also, I dislike mixing hyphens and underscores in command line options.  So make them hyphens. :-)

ehsannas:
This actually passes the builds/checks. The reason for failure is a connection timeout.

antiagainst:
@ehsannas: Yep. I just wish we have complete control over projects in Khronos Group so that we can restart the test job as needed.

include/spirv-tools/libspirv.h
```diff
@@ -366,6 +370,18 @@ typedef enum {
   SPV_ENV_OPENGL_4_5,  // OpenGL 4.5 plus GL_ARB_gl_spirv, latest revisions.
 } spv_target_env;
 
+// SPIR-V Validator can be parameterized with the following Universal Limits.
+typedef enum {
+  validator_limit_max_struct_members,
```
dneto0: In C, enumerant names are not scoped.  So please prefix this with `spv_`.

But thanks for making this enum in the first place!

source/spirv_validator_options.h
```diff
@@ -0,0 +1,56 @@
...
+    /* max_access_chain_indexes */ 255,
+};
+
+// Manages command line options passed to the SPIR-V Validator. New struct
+// members may be added for any new option.
+struct spv_validator_options_t {
+  spv_validator_options_t()
+      : universalLimits(kDefaultValidatorUniversalLimits) {}
+
+  validator_universal_limits_t universalLimits;
```
dneto0: Google style convention for data members would be `universal_limits_`

source/spirv_validator_options.h
```diff
@@ -0,0 +1,56 @@
...
+  uint32_t max_struct_depth;
+  uint32_t max_local_variables;
+  uint32_t max_global_variables;
+  uint32_t max_switch_branches;
+  uint32_t max_function_args;
+  uint32_t max_control_flow_nesting_depth;
+  uint32_t max_access_chain_indexes;
+};
+
+const validator_universal_limits_t kDefaultValidatorUniversalLimits = {
```
dneto0: I recommend you use default member initialization instead:
http://en.cppreference.com/w/cpp/language/data_members

That would possibly be more robust to future changes, is more readable, and is shorter.  You'd just need a comment toward the top saying the default initialization is to the default universal limit.

source/spirv_validator_options.cpp
```diff
@@ -0,0 +1,77 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cassert>
+#include <cstring>
+
+#include "spirv_validator_options.h"
+
+bool spvParseUniversalLimitsOptions(const char* s, spv_validator_limit* type) {
+  auto match = [s](const char* b) {
```
dneto0: Good use of lambda.  :-)

test/val/val_state_test.cpp
```diff
@@ -35,10 +36,13 @@ using std::vector;
 class ValidationStateTest : public testing::Test {
  public:
   ValidationStateTest()
-      : context_(spvContextCreate(SPV_ENV_UNIVERSAL_1_0)), state_(context_) {}
+      : context_(spvContextCreate(SPV_ENV_UNIVERSAL_1_0)),
+        options_(spvValidatorOptionsCreate()),
+        state_(context_, options_) {}
```
dneto0: You're leaking the options object.  Add a destructor to destroy it.

tools/val/val.cpp
```diff
@@ -33,73 +34,122 @@ or if the filename is "-", then the binary is read from standard input.
...
         fprintf(stderr, "error: More than one input file specified\n");
-        return 1;
+        continue_processing = false;
+        return_code = 1;
       }
     }
   }
 
+  // Exit if command line parsing was not successful.
+  if (!continue_processing) {
```
dneto0: This is fine.   However, another two ways to handle this:
1. Don't care about leaking the object because this is the main function and it's going to exit soon.  :-)  Slightly ugly, but if you do that then add a comment saying "Deliberately leaking this object."
2. Do options object management unconditionally in this top level main function, but then wrap the "actually do stuff" remainder of the code in a function.  This is more complex than what you have.

source/spirv_validator_options.h
ehsannas: very cool! :+1:

source/spirv_validator_options.h
```diff
@@ -21,35 +21,26 @@
...
-const validator_universal_limits_t kDefaultValidatorUniversalLimits = {
-    /* max_struct_members */ 16383,
-    /* max_struct_depth */ 255,
-    /* max_local_variables */ 524287,
-    /* max_global_variables */ 65535,
-    /* max_switch_branches */ 16383,
-    /* max_function_args */ 255,
-    /* max_control_flow_netsting_depth */ 1023,
-    /* max_access_chain_indexes */ 255,
+  uint32_t max_struct_members{16383};
```
dneto0: This is fine.  For some reason, I had only seen equals initializaitions for this kind of thing, e.g.
`uint32_t max_struct_members = 16383;`

But this form is correct.

dneto0:
Squashed and pushed into master as  da4ae05

PR #560, opened 2017-02-23
Issue 559: check type declaration uniqueness

atgoo:
Adds PassTypeUnique to the validator.
Disallows repeated declarations of all types except for aggregates.

### DIFF
--- a/source/val/validation_state.h
+++ b/source/val/validation_state.h
@@ -16,6 +16,7 @@
 #define LIBSPIRV_VAL_VALIDATIONSTATE_H_
 
 #include <deque>
+#include <set>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
@@ -304,6 +305,10 @@ class ValidationState_t {
   // Returns the state of optional features.
   const Feature& features() const { return features_; }
 
+  /// Adds the instruction data to unique_type_declarations_.
+  /// Returns false if an identical type declaration already exists.
+  bool RegisterUniqueTypeDeclaration(const spv_parsed_instruction_t& inst);
+
  private:
   ValidationState_t(const ValidationState_t&);
 
@@ -370,6 +375,12 @@ class ValidationState_t {
 
   /// Stores the list of decorations for a given <id>
   std::unordered_map<uint32_t, std::vector<Decoration>> id_decorations_;
+
+  /// Stores type declarations which need to be unique (i.e. non-aggregates),
+  /// in the form [opcode, operand words], result_id is not stored.
+  /// Using ordered set to avoid the need for a vector hash function.
+  /// The size of this container is expected not to exceed double-digits.
+  std::set<std::vector<uint32_t>> unique_type_declarations_;
 
   AssemblyGrammar grammar_;
 
--- a/source/validate_type_unique.cpp
+++ b/source/validate_type_unique.cpp
@@ -1,0 +1,48 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Ensures type declarations are unique unless allowed by the specification.
+
+#include "validate.h"
+
+#include "diagnostic.h"
+#include "opcode.h"
+#include "val/instruction.h"
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+// Validates that type declarations are unique, unless multiple declarations
+// of the same data type are allowed by the specification.
+// (see section 2.8 Types and Variables)
+spv_result_t TypeUniquePass(ValidationState_t& _,
+                            const spv_parsed_instruction_t* inst) {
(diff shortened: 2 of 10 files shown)


source/val/validation_state.h
```diff
@@ -354,6 +360,10 @@ class ValidationState_t {
   /// Stores the list of decorations for a given <id>
   std::unordered_map<uint32_t, std::vector<Decoration>> id_decorations_;
 
+  /// Stores type declarations which need to be unique (i.e. non-aggregates),
+  /// [opcode, operand words]->result_id.
+  std::map<std::vector<uint32_t>, uint32_t> unique_type_declarations_;
```
dneto0: Why is this a map instead of a set?  You never use the result id value.

Also, why is this an ordered container rather than an unordered container?

source/validate_type_unique.cpp
```diff
@@ -0,0 +1,49 @@
...
+// (see section 2.8 Types and Variables)
+spv_result_t TypeUniquePass(ValidationState_t& _,
+                            const spv_parsed_instruction_t* inst) {
+  const auto opcode = inst->opcode;
+  if (opcode == SpvOpTypeArray || opcode == SpvOpTypeStruct) {
+    // Duplicate declarations of aggregates are allowed.
+    return SPV_SUCCESS;
+  }
+
+  if ((opcode >= SpvOpTypeVoid && opcode <= SpvOpTypeForwardPointer) ||
```
dneto0: This expression captures the condition that the opcode is a type-creating instruction.   It seems a bit fragile. 
It's better to centralize it.

I just found spvOpcodeGeneratesType in opcode.h.  That would be better to use.
However, we need to fix it to add PipeStorage and NamedBarrier.
Also, it knowingly excludes ForwardPointer.  OpTypeForwardPointer does not create a new type, so it should be excluded from the uniqueness rule, I think.

test/val/val_type_unique_test.cpp
```diff
@@ -0,0 +1,156 @@
+// Copyright (c) 2016 Google Inc.
```
dneto0: 2017

test/val/val_type_unique_test.cpp
```diff
@@ -0,0 +1,156 @@
...
+namespace {
+
+using ::testing::HasSubstr;
+
+using std::string;
+using std::pair;
+
+using ValidateTypeUnique = spvtest::ValidateBase<pair<string, bool>>;
+
+const string header = R"(
```
dneto0: Google style guide forbids variables of class type with static storage duration.

In this case we prefer using a free function to return the value you want.   (I know the rest of the test code is inconsistent on this point.)

test/val/val_type_unique_test.cpp
```diff
@@ -0,0 +1,156 @@
...
+%falsel = OpLabel
+%sub1 = OpISub %uintt %a %b
+%sub2 = OpISub %uintt %a %b
+OpBranch %endl
+%endl = OpLabel
+OpReturn
+OpFunctionEnd
+)";
+
+string GetErrorString(SpvOp_ opcode) {
```
dneto0: This declaration needs a comment, like any other method or function declaration.

test/val/val_type_unique_test.cpp
```diff
@@ -0,0 +1,156 @@
...
+TEST_F(ValidateTypeUnique, duplicate_vfunc) {
+  string str = header + R"(
+%vfunct2 = OpTypeFunction %voidt
+)" + body;
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_ERROR_INVALID_DATA, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr(GetErrorString(SpvOpTypeFunction)));
+}
+
```
dneto0: I'd also like to see:
- OpTypeForwardPointer can be duplicated.
- OpTypePipeStorage, OpTypeNamedBarrier  (So we cover the new cases for spvOpcodeGeneratesType)

test/val/val_type_unique_test.cpp
```diff
@@ -0,0 +1,156 @@
...
+%boolt = OpTypeBool
+%array2 = OpTypeArray %vec3t %num3
+%voidt = OpTypeVoid
+%vfunct = OpTypeFunction %voidt
+%struct2 = OpTypeStruct %floatt %floatt %vec3t
+%false = OpConstantFalse %boolt
+%true = OpConstantTrue %boolt
+)";
+
+const string body = R"(
```
dneto0: This part of the code is not necessary to the tests.  You should remove this.

atgoo: This part contains multiple duplicate instructions to test that the validator only triggers on duplicate type declarations and not on other opcodes.

test/val/val_type_unique_test.cpp
atgoo: Done.

test/val/val_type_unique_test.cpp
```diff
@@ -0,0 +1,219 @@
...
+namespace {
+
+using ::testing::HasSubstr;
+
+using std::string;
+using std::pair;
+
+using ValidateTypeUnique = spvtest::ValidateBase<pair<string, bool>>;
+
+const string& GetHeader() {
```
dneto0: I don't think the const reference is buying you much.  The temp object has to exist for the lifetime of the calling expression anyway.
But this is fine.

dneto0: I misread the code here.  You're creating the static string variable internally, and then returning the reference to that.
This is correct, is definitely saving work on subsequent calls.  So this is good!

dneto0:
Rebased and pushed into master as 0e9c24fdd112

Issue #275, opened 2016-07-26
Validator: SpecId decoration should only be applied to OpSpecConstant instructions on scalars


dneto0:
(No particular priority.  I just wanted to record this rule while I had it in mind.)

dneto0:
Someone has started working on this.

PR #561, opened 2017-02-25
Validate that SpecId decoration target is an OpSpecConstant instruction on a scalar

aliyap:
Fixes: https://github.com/KhronosGroup/SPIRV-Tools/issues/275


source/opcode.h
```diff
@@ -68,6 +68,10 @@ int32_t spvOpcodeIsConstant(const SpvOp opcode);
 // Returns true if the given opcode is a constant or undef.
 bool spvOpcodeIsConstantOrUndef(const SpvOp opcode);
 
+// Determines if the given opcode is a scalar specialization constant. Returns
+// zero if false, non-zero otherwise.
+int32_t spvOpcodeIsScalarSpecConstant(const SpvOp opcode);
```
ehsannas: Is there a reason you haven't used `bool` as the return type? A boolean return type sounds more intuitive based on what the function does.

source/validate_id.cpp
```diff
@@ -134,6 +134,24 @@ bool idUsage::isValid<SpvOpLine>(const spv_instruction_t* inst,
...
+  if (decoration == SpvDecorationSpecId) {
+    auto targetIndex = 1;
+    auto target = module_.FindDef(inst->words[targetIndex]);
+    if (!target || !spvOpcodeIsScalarSpecConstant(target->opcode())) {
+      DIAG(targetIndex) << "OpDecorate SpectId decoration target <id> '"
+                        << inst->words[decorationIndex]
+                        << "' is not a scalar specialization constant.";
+      return false;
+    }
+  }
```
ehsannas: Since this function currently doesn't cover validation for all decorations, I think you should add a `TODO` comment stating that more validation code for OpDecorate should be added in the future.

test/val/val_id_test.cpp
```diff
@@ -3724,6 +3724,45 @@ OpFunctionEnd
...
+TEST_F(ValidateIdWithMessage, SpecIdTargetGood) {
+  string spirv = kGLSL450MemoryModel + R"(
+OpDecorate %3 SpecId 200
+OpDecorate %4 SpecId 201
+%1 = OpTypeVoid
+%2 = OpTypeFunction %1
+%int = OpTypeInt 32 0
+%bool = OpTypeBool
+%3 = OpSpecConstant %int 3
+%4 = OpSpecConstantTrue %bool
```
dneto0: Suggest adding OpSpecConstantFalse case here as well since it's so easy to do.

test/val/val_id_test.cpp
```diff
@@ -3724,6 +3724,45 @@ OpFunctionEnd
...
+OpReturn
+OpFunctionEnd
+  )";
+  CompileSuccessfully(spirv.c_str());
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(
+      getDiagnosticString(),
+      HasSubstr("OpDecorate SpectId decoration target <id> '1' is not a "
+                "scalar specialization constant."));
+}
```
dneto0: Suggest adding two cases, where the target is:
- an OpSpecConstantOp instruction
- an OpSpecConstnatComposite instruction

That's to prove that if an instruction name starts with "OpSpecConstant*" then it doesn't imply that it can be the target of a SpecId decoration.

dneto0: In the work for https://github.com/KhronosGroup/SPIRV-Tools/pull/560 we found that the validator is already broken on those two types.  So I recommend you hold on until that is fixed.

dneto0: I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/563 for the existing validator problem.

test/val/val_id_test.cpp
```diff
@@ -3798,6 +3798,89 @@ OpFunctionEnd
...
+%main = OpFunction %1 None %2
+%4 = OpLabel
+OpReturn
+OpFunctionEnd
+  )";
+  CompileSuccessfully(spirv.c_str());
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(
+      getDiagnosticString(),
+      HasSubstr("OpDecorate SpectId decoration target <id> '1' is not a "
```
dneto0: The reported ID looks funny.  It's 1 because the assembler assigns IDs sequentially from 1, based on which id name was mentioned first. Same thing happens in the later tests as well.

I'll patch this before merging in.

dneto0: I noticed some other tests in the same file have the same issue. No worries.

dneto0:
Squashed and pushed into master as  edb52647bdab14
ID rewrites in tests in master as 892b0099590

Thank you for your contribution!

PR #562, opened 2017-02-27
Dneto validation spv khr 16bit storage

dneto0:


source/val/validation_state.h
```diff
@@ -359,8 +369,17 @@ class ValidationState_t {
   SpvAddressingModel addressing_model_;
   SpvMemoryModel memory_model_;
 
+  // Is OpTypeInt allowed with 16 bit width?
```
ehsannas: `permit_int16_type_` and `permit_float16_type_` variables are not used. I think the `Feature` members are meant to do the work instead?

source/val/validation_state.h
```diff
@@ -290,6 +296,10 @@ class ValidationState_t {
   bool IsStructTypeWithBuiltInMember(uint32_t id) const {
     return (builtin_structs_.find(id) != builtin_structs_.end());
   }
+
+  // Returns the state of optional features.
+  const Feature& features() { return features_; }
```
antiagainst: This method can be declared as `const`.

source/val/validation_state.h
dneto0: Removed in next commit.  Thanks!

source/val/validation_state.cpp
```diff
@@ -332,6 +332,24 @@ void ValidationState_t::RegisterCapability(SpvCapability cap) {
...
+      break;
+    case SpvCapabilityFloat16:
+    case SpvCapabilityFloat16Buffer:
+      features_.declare_float16_type = true;
+      break;
+    case SpvCapabilityStorageUniformBufferBlock16:
+    case SpvCapabilityStorageUniform16:
+    case SpvCapabilityStoragePushConstant16:
+    case SpvCapabilityStorageInputOutput16:
+      features_.declare_int16_type = true;
```
antiagainst: Is this being too permissive? From my understanding, `SpvCapabilityInt16` enables more operations on int16 than `SpvCapabilityStorage*16`, but we don't actually distinguish that here. But we can handle this later maybe.

dneto0: Correct, Int16 also allows arithmetic operations on those types.  But the validator doesn't check that yet.  Also, that's why the feature is called just *declare* 16bit types.   Later we will add something like int16_arithmetic, and that will distinguish the cases.

source/val/validation_state.h
dneto0: Done in next commit.

dneto0:
All feedback addressed.

dneto0:
Squashed and pushed into master as af7125dfb0

PR #564, opened 2017-02-28
Refactored validation_state, added new types

atgoo:
- validation_state.cpp uses functions from opcode.h instead of in-place
switches which need to be updated.
- added new spirv 1.1 type declaration opcodes to a 'is op type
declaration' switch in opcode.cpp.


dneto0:
Rebased and pushed into master as 4ef3b3e0b

PR #566, opened 2017-02-28
Add a default virtual dtor to Pass and check for missing virtual dtors

jfroy:


dneto0:
Rebased and pushed into master as 6a2b514826

Issue #559, opened 2017-02-17
Check non-aggregate type uniqueness.


dneto0:
In section 2.8: "That is, non-aggregate type declarations must all have different opcodes or operands. "

dneto0:
The spec constant folder has a trie data class we can use to check for uniqueness.  We should refactor it out.

dneto0:
Actually, we probably don't have to go that far.  All the non-result-id words have to match. That's a simpler condition to check.

dneto0:
@atgoo is working on this

dneto0:
Fixed by https://github.com/KhronosGroup/SPIRV-Tools/pull/560

dneto0:
455 Vulkan CTS tests appear to fail with this check.

dneto0:
@atgoo determined that the validation failures are bugs in the CTS, and has filed an issue there.
Closing

dneto0:
Reopening. We're uncovering more validation failures in CTS.

PR #567, opened 2017-03-01
WIP: Validator: duplicate type declaration check fails silently

atgoo:
Validator check for uniqueness of type declarations
(commit 0e9c24fdd112e7742b9b5dcff35aee32e71a66fd)
was causing failures in vulkancts tests.

See https://github.com/KhronosGroup/SPIRV-Tools/issues/559

This check will now fail silently (return SPV_SUCCESS even if it fails).

### DIFF
--- a/source/validate_type_unique.cpp
+++ b/source/validate_type_unique.cpp
@@ -14,8 +14,6 @@
 
 // Ensures type declarations are unique unless allowed by the specification.
 
-#include <cstdio>
-
 #include "validate.h"
 
 #include "diagnostic.h"
--- a/source/validate_type_unique.cpp
+++ b/source/validate_type_unique.cpp
@@ -40,8 +40,7 @@ spv_result_t TypeUniquePass(ValidationState_t& _,
     if (!_.RegisterUniqueTypeDeclaration(*inst)) {
       // TODO(atgoo@github) Error logging temporarily disabled because it's
       // failing vulkancts tests. Message in the diagnostics is for unit tests.
-      fprintf(stderr, "WARNING: Duplicate non-aggregate type declarations are"
-              " not allowed. Opcode %d\n", inst->opcode);
+      // See https://github.com/KhronosGroup/SPIRV-Tools/issues/559
       // return _.diag(SPV_ERROR_INVALID_DATA)
       return _.diag(SPV_SUCCESS)
           << "Duplicate non-aggregate type declarations are not allowed."
--- a/source/validate_type_unique.cpp
+++ b/source/validate_type_unique.cpp
@@ -14,6 +14,8 @@
 
 // Ensures type declarations are unique unless allowed by the specification.
 
+#include <cstdio>
+
 #include "validate.h"
 
 #include "diagnostic.h"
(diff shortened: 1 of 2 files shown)


source/validate_type_unique.cpp
```diff
@@ -36,7 +38,12 @@ spv_result_t TypeUniquePass(ValidationState_t& _,
     }
 
     if (!_.RegisterUniqueTypeDeclaration(*inst)) {
-      return _.diag(SPV_ERROR_INVALID_DATA)
+      // TODO(atgoo@github) Error logging temporarily disabled because it's
+      // failing vulkancts tests. Message in the diagnostics is for unit tests.
+      fprintf(stderr, "WARNING: Duplicate non-aggregate type declarations are"
```
dneto0: I don't think you want this.
We shouldn't print things to stderr or stdout.  Always pass messages back through a diagnostic so it can be captured and redirected via the function pointer in the context object.

source/validate_type_unique.cpp
```diff
@@ -36,7 +38,12 @@ spv_result_t TypeUniquePass(ValidationState_t& _,
     }
 
     if (!_.RegisterUniqueTypeDeclaration(*inst)) {
-      return _.diag(SPV_ERROR_INVALID_DATA)
+      // TODO(atgoo@github) Error logging temporarily disabled because it's
```
dneto0: Suggest referencing the original issue in SPIRV-Tools.

test/val/val_type_unique_test.cpp
```diff
@@ -25,10 +25,14 @@ namespace {
...
 
 using std::string;
-using std::pair;
 
 using ValidateTypeUnique = spvtest::ValidateBase<bool>;
 
+// TODO(atgoo@github) Error logging temporarily disabled because it's failing
+// vulkancts tests.
+// const spv_result_t g_duplicate_type_error = SPV_ERROR_INVALID_DATA;
+const spv_result_t g_duplicate_type_error = SPV_SUCCESS;
```
dneto0: Google style guide would have this as kDuplicateTypeError.

source/validate_type_unique.cpp
atgoo: If I just return SPV_SUCCESS, then it won't be printed. Diagnostic is printed in tools/val/val.cpp, only if error code is not zero.

If I return SPV_WARNING, it won't be correctly handled (results are not checked for >= 0, but for == 0).

dneto0: ok.  Let's fix tools/val/val.cpp  in another MR. It should print the warning.

dneto0: I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/568 to fix the command line tool.

source/validate_type_unique.cpp
atgoo: Done.

dneto0:
We found that at least a few of the Vulkan CTS failures was in essentially hand-coded SPIR-V assembly.
There were duplicate instances of OpTypeInt 32 0 in the same module.  That's just a bug in the CTS test.

So I'm going to hold off on this PR until we confirm that all the CTS failures were of that nature.

dneto0:
@atgoo has found more distinct kinds of validation failures in the CTS when adding the check.
I'll disable this for now until we have a 100% clear picture or have fixed all CTS or complier or spec issues.

dneto0:
Squashed and merged into master as  94d94e1f

PR #570, opened 2017-03-01
Allow using FPRoundingMode when see VK_KHR_16bit_storage

antiagainst:
According to the extension, FPRoundingMode should be allowed to
use without requiring Kernel capability when VK_KHR_16bit_storage
is enabled.

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -298,6 +298,7 @@ void ValidationState_t::RegisterCapability(SpvCapability cap) {
     case SpvCapabilityStorageInputOutput16:
       features_.declare_int16_type = true;
       features_.declare_float16_type = true;
+      features_.free_fp_rounding_mode = true;
     default:
       break;
   }
--- a/source/val/validation_state.h
+++ b/source/val/validation_state.h
@@ -57,8 +57,11 @@ class ValidationState_t {
  public:
   // Features that can optionally be turned on by a capability.
   struct Feature {
-    bool declare_int16_type = false;    // Allow OpTypeInt with 16 bit width?
-    bool declare_float16_type = false;  // Allow OpTypeFloat with 16 bit width?
+    bool declare_int16_type = false;     // Allow OpTypeInt with 16 bit width?
+    bool declare_float16_type = false;   // Allow OpTypeFloat with 16 bit width?
+    bool free_fp_rounding_mode = false;  // Allow the FPRoundingMode decoration
+                                         // and its vaules to be used without
+                                         // requiring any capability
   };
 
   ValidationState_t(const spv_const_context context,
@@ -202,7 +205,7 @@ class ValidationState_t {
   /// Returns the memory model of this module, or Simple if uninitialized.
   SpvMemoryModel memory_model() const;
 
-  AssemblyGrammar& grammar() { return grammar_; }
+  const AssemblyGrammar& grammar() const { return grammar_; }
 
   /// Registers the instruction
   void RegisterInstruction(const spv_parsed_instruction_t& inst);
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -63,7 +63,7 @@ spv_result_t CapabilityError(ValidationState_t& _, int which_operand,
 }
 
 // Returns an operand's required capabilities.
-CapabilitySet RequiredCapabilities(const AssemblyGrammar& grammar,
+CapabilitySet RequiredCapabilities(const ValidationState_t& state,
                                    spv_operand_type_t type, uint32_t operand) {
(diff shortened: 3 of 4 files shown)


source/val/validation_state.cpp
```diff
@@ -303,6 +304,12 @@ void ValidationState_t::RegisterCapability(SpvCapability cap) {
   }
 }
 
+void ValidationState_t::RegisterExtension(const char* extension) {
+  if (!strcmp(extension, "SPV_KHR_16bit_storage")) {
+    features_.free_fp_rounding_mode = true;
```
dneto0: The flag should be flipped when we see the corresponding capabilities.
The reason is that a future version of an API can bring the extension into core, and then you'll have modules with the capabilities but not the OpExtension.

source/val/validation_state.h
```diff
@@ -172,6 +175,12 @@ class ValidationState_t {
   /// Registers the capability and its dependent capabilities
   void RegisterCapability(SpvCapability cap);
 
+  // Registers the given extension.
```
dneto0: I think you won't need this.

source/val/validation_state.cpp
```diff
@@ -15,6 +15,7 @@
 #include "val/validation_state.h"
 
 #include <cassert>
+#include <cstring>
```
dneto0: Won't need this.

source/validate_instruction.cpp
```diff
@@ -79,11 +80,24 @@ CapabilitySet RequiredCapabilities(const AssemblyGrammar& grammar,
...
   spv_operand_desc operand_desc;
-  if (SPV_SUCCESS == grammar.lookupOperand(type, operand, &operand_desc)) {
-    return operand_desc->capabilities;
+  const auto ret = state.grammar().lookupOperand(type, operand, &operand_desc);
+  if (ret == SPV_SUCCESS) {
+    CapabilitySet result = operand_desc->capabilities;
+
+    // Allow FPRoundingMode decoration if requested
+    if (state.features().free_fp_rounding_mode &&
+        !strcmp(operand_desc->name, "FPRoundingMode")) {
```
dneto0: Compare against the value not the string.

source/validate_instruction.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <cstring>
```
dneto0: Won't need this.

source/validate_instruction.cpp
```diff
@@ -331,6 +343,10 @@ spv_result_t RegisterDecorations(ValidationState_t& _,
 spv_result_t InstructionPass(ValidationState_t& _,
                              const spv_parsed_instruction_t* inst) {
   const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+  if (opcode == SpvOpExtension) {
```
dneto0: Won't need this now.

test/val/val_data_test.cpp
```diff
@@ -540,4 +541,46 @@ OpTypeForwardPointer %_ptr_Generic_struct_A Generic
   ASSERT_EQ(SPV_SUCCESS, ValidateInstructions());
 }
 
+TEST_F(ValidateData, ext_16bit_storage_allow_free_fp_rounding_mode) {
+  for (const char* ext : {"RTE", "RTZ", "RTP", "RTN"}) {
+    string str = string(R"(
+        OpCapability Shader
+        OpCapability Linkage
+        OpExtension "SPV_KHR_16bit_storage"
```
dneto0: Test against the capabilities, not the extension string.

source/val/validation_state.cpp
antiagainst: Done

source/val/validation_state.cpp
antiagainst: Yep. Changed.

source/val/validation_state.h
antiagainst: Removed.

source/validate_instruction.cpp
antiagainst: Removed

source/validate_instruction.cpp
antiagainst: That also requires to consider the `type`. Changed.

source/validate_instruction.cpp
antiagainst: Done

test/val/val_data_test.cpp
antiagainst: Done

dneto0:
Rebased and pushed into master as 40be4bdd34a0fce1dfcc97f571704e7980d09a36

PR #574, opened 2017-03-07
Test assembler, disassembler for recently published SPV_KHR_* extensions

dneto0:


dneto0:
Rebased and pushed into master as 4203720

PR #577, opened 2017-03-08
Fix nullptr dereferencing in val_fixtures.cpp

atgoo:


dneto0:
This update is to test code. It's not clear if what was there was a bug because the API methods were not described at all. So we don't know under what conditions the access was going to occur.  It's likely the intent was that those methods would only be used when a diagnostic object is generated. But there's no comment to that effect.

But I'm not so fussy with test code.  (Famous last words.)


Rebased and pushed into master as  9696393c

PR #579, opened 2017-03-08
Validator parses and registers OpExtension

atgoo:
Known extensions are saved in validation state. Unknown extension
produce a dignostic message, but do not fail the validation.


source/validate_instruction.cpp
```diff
@@ -340,9 +341,44 @@ spv_result_t RegisterDecorations(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+// Parses OpExtension instruction and registers extension.
```
dneto0: Should say what the return code means. ?  Especially since it's always SPV_SUCCESS

source/validate_instruction.cpp
```diff
@@ -340,9 +341,44 @@ spv_result_t RegisterDecorations(ValidationState_t& _,
...
+                               const spv_parsed_instruction_t* inst) {
+  assert(inst->opcode == SpvOpExtension);
+  assert(inst->num_operands == 1);
+
+  const auto& operand = inst->operands[0];
+  assert(operand.type == SPV_OPERAND_TYPE_LITERAL_STRING);
+  assert(inst->num_words > operand.offset);
+
+  const size_t max_length =
+      sizeof(*inst->words) * (inst->num_words - operand.offset);
```
dneto0: Might as well make the first term sizeof(uint32_t) because that's predefined for the SPIR-V spec.

source/validate_instruction.cpp
```diff
@@ -340,9 +341,44 @@ spv_result_t RegisterDecorations(ValidationState_t& _,
...
+spv_result_t RegisterExtension(ValidationState_t& _,
+                               const spv_parsed_instruction_t* inst) {
+  assert(inst->opcode == SpvOpExtension);
+  assert(inst->num_operands == 1);
+
+  const auto& operand = inst->operands[0];
+  assert(operand.type == SPV_OPERAND_TYPE_LITERAL_STRING);
+  assert(inst->num_words > operand.offset);
+
+  const size_t max_length =
```
dneto0: The binary parser already does the check for the nul terminator.   So this is a redundant sequence of code to compute max bounds and check again.

atgoo:
Done.

source/extensions.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "extensions.h"
+
+#include <string>
+
+namespace libspirv {
+
+bool ParseSpvExtensionFromString(const std::string& str, Extension* extension) {
+  if (str == "SPV_KHR_shader_ballot") {
+    *extension = Extension::kSPV_KHR_shader_ballot;
+    return true;
```
antiagainst: Why not using `else if`s instead of early returns?

source/extensions.h
```diff
@@ -0,0 +1,50 @@
...
+#include <string>
+
+#include "enum_set.h"
+#include "message.h"
+#include "spirv-tools/libspirv.hpp"
+
+namespace libspirv {
+
+// The known SPIR-V extensions.
+// TODO(dneto): Consider auto-generating this list?
```
antiagainst: +1

test/val/val_extensions_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+using ::testing::HasSubstr;
+using ::testing::Not;
+using ::testing::Values;
+
+using std::string;
+
+using ValidateKnownExtensions = spvtest::ValidateBase<string>;
+using ValidateUnknownExtension = spvtest::ValidateBase<bool>;
+
+// Returns expected error string if |extensions| is not recognized.
```
antiagainst: s/extensions/extension/

test/val/val_extensions_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_SUCCESS, ValidateInstructions());
+  // TODO: Reenable once https://github.com/KhronosGroup/SPIRV-Tools/pull/577
+  // is merged.
+  // EXPECT_THAT(getDiagnosticString(), Not(HasSubstr(GetErrorString(extension))));
+}
+
+TEST_F(ValidateUnknownExtension, FailSilently) {
+  const std::string extension = "ERROR_unknown_extension";
+  const string str =
```
antiagainst: I would add an extension with a truncated name out of a recognized one, like `SPV_KHR_shader_ballo`. It can guard problematic future changes to string comparison.

antiagainst: Also something like `SPV_KHR_shader_balloty". :)

antiagainst: Also something like `SPV_KHR_shader_balloty". :)

source/extensions.h
atgoo: https://github.com/KhronosGroup/SPIRV-Tools/issues/580

dneto0:
Squashed and pushed into master as 1c43cb855

PR #581, opened 2017-03-09
Refactored enum_set

atgoo:
- removed forgotten file enum_set.cpp
- added IsEmpty and HasAnyOf
- hidden unsafe functions Add(uint32_t), Contains(uint32_t)
- added new tests


dneto0:
Fails compilation in some of the bots.

test/enum_set_test.cpp
```diff
@@ -20,75 +20,123 @@
...
+  EnumSet<uint32_t> set(4);
+  EXPECT_FALSE(set.IsEmpty());
+}
+
+TEST(EnumSet, IsEmpty4) {
+  EnumSet<uint32_t> set(300);
+  EXPECT_FALSE(set.IsEmpty());
+}
+
+TEST(EnumSet, HasAnyOf) {
```
dneto0: You should split this into multiple tests, so each behaviour is tested separately.
Ideally, each test should have the structure:  arrange, act, assert.
Also, the guidance on naming is that test names should include the condition being tested and the expected outcome.

Why?  
- separation of concerns.  if something fails some day, it should be clear what went wrong and why.  It should be very isolated
- the name of the test should clearly indicate the intent.  That helps in diagnosing and responding to failures, or when we have to rewrite tests due to other changes in the code being tested.

dneto0:
Squashed and pushed into master as 1fb8c37b5718118

PR #582, opened 2017-03-10
Validator checks if operands require extension

atgoo:
If required extension is not found, returning new error
SPV_ERROR_MISSING_EXTENSION.

### DIFF
--- a/source/extensions.h
+++ b/source/extensions.h
@@ -18,14 +18,14 @@
 #include <string>
 
 #include "enum_set.h"
-#include "message.h"
-#include "spirv-tools/libspirv.hpp"
+#include "spirv-tools/libspirv.h"
 
 namespace libspirv {
 
 // The known SPIR-V extensions.
 // TODO(dneto): Consider auto-generating this list?
-// When updating this list, consider also updating ParseSpvExtensionFromString.
+// When updating this list, consider also updating ParseSpvExtensionFromString
+// and SpvExtensionToString.
 enum class Extension {
   kSPV_KHR_shader_ballot,
   kSPV_KHR_shader_draw_parameters,
@@ -42,8 +42,17 @@ enum class Extension {
 
 using ExtensionSet = EnumSet<Extension>;
 
+// Returns literal string operand of OpExtension instruction.
+std::string GetExtensionString(const spv_parsed_instruction_t* inst);
+
 // Finds Extension enum corresponding to |str|. Returns false if not found.
 bool ParseSpvExtensionFromString(const std::string& str, Extension* extension);
+
+// Returns text string corresponding to |extension|.
+std::string ExtensionToString(Extension extension);
+
+// Returns text string listing |extensions| separated by whitespace.
+std::string ExtensionSetToString(const ExtensionSet& extensions);
 
 }  // namespace libspirv
 
--- a/source/table.h
+++ b/source/table.h
@@ -40,11 +40,6 @@ typedef struct spv_operand_desc_t {
   const char* name;
   const uint32_t value;
   const libspirv::CapabilitySet capabilities;
-  // A set of extensions that enable this feature. If empty then this operand
-  // value is always enabled, i.e. it's in core.  The assembler, binary parser,
-  // and disassembler ignore this rule, so you can freely process invalid
-  // modules.
-  // TODO(dneto): Add validator support to check extensions.
   const libspirv::ExtensionSet extensions;
   const spv_operand_type_t operandTypes[16];  // TODO: Smaller/larger?
 } spv_operand_desc_t;
--- a/source/table.h
+++ b/source/table.h
@@ -40,6 +40,10 @@ typedef struct spv_operand_desc_t {
   const char* name;
   const uint32_t value;
   const libspirv::CapabilitySet capabilities;
+  // A set of extensions that enable this feature. If empty then this operand
+  // value is always enabled, i.e. it's in core. The assembler, binary parser,
+  // and disassembler ignore this rule, so you can freely process invalid
+  // modules.
   const libspirv::ExtensionSet extensions;
   const spv_operand_type_t operandTypes[16];  // TODO: Smaller/larger?
 } spv_operand_desc_t;
--- a/source/util/string_utils.h
+++ b/source/util/string_utils.h
@@ -1,0 +1,41 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_UTIL_STRING_UTILS_H_
+#define LIBSPIRV_UTIL_STRING_UTILS_H_
+
+#include <string>
+
+namespace spvutils {
+
+// Converts cardinal number to ordinal number string.
+std::string CardinalToOrdinal(size_t cardinal) {
+  const size_t mod10 = cardinal % 10;
+  const size_t mod100 = cardinal % 100;
+  std::string suffix;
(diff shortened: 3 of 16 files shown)


source/extensions.h
```diff
@@ -18,14 +18,14 @@
...
-#include "message.h"
-#include "spirv-tools/libspirv.hpp"
+#include "spirv-tools/libspirv.h"
 
 namespace libspirv {
 
 // The known SPIR-V extensions.
 // TODO(dneto): Consider auto-generating this list?
-// When updating this list, consider also updating ParseSpvExtensionFromString.
+// When updating this list, consider also updating ParseSpvExtensionFromString
```
dneto0: It's becoming even more important to automatically parse extensions from the JSON grammar file ...

source/table.h
```diff
@@ -40,11 +40,6 @@ typedef struct spv_operand_desc_t {
   const char* name;
   const uint32_t value;
   const libspirv::CapabilitySet capabilities;
-  // A set of extensions that enable this feature. If empty then this operand
```
dneto0: Everything but the TODO still applies.   In particular, the behaviour on empty set is not obvious without the comment.

source/validate.cpp
```diff
@@ -117,6 +119,34 @@ void DebugInstructionPass(ValidationState_t& _,
   }
 }
 
+// Parses OpExtension instruction and registers extension.
+void RegisterExtension(ValidationState_t& _,
+                       const spv_parsed_instruction_t* inst) {
+  const std::string extension_str = libspirv::GetExtensionString(inst);
+  Extension extension;
+  if (!ParseSpvExtensionFromString(extension_str, &extension)) {
+    _.diag(SPV_SUCCESS) << "Failed to parse OpExtension " << extension_str;
```
dneto0: This is ok but it might be clearer for the user to say "Found unrecognized extension ...."

source/validate.cpp
```diff
@@ -117,6 +119,34 @@ void DebugInstructionPass(ValidationState_t& _,
...
+    return;
+  }
+
+  _.RegisterExtension(extension);
+}
+
+spv_result_t ProcessExtensions(
```
dneto0: Needs a method comment.

source/validate.cpp
```diff
@@ -117,6 +119,34 @@ void DebugInstructionPass(ValidationState_t& _,
...
+    void* user_data, const spv_parsed_instruction_t* inst) {
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+  if (opcode == SpvOpCapability)
```
dneto0: It's very much worth saying how this fits with the parser:

For a valid module, the intial segment of instructions are just optional OpCapability followed by OpExtension. Terminate parsing early when you get past those.

source/validate.cpp
```diff
@@ -215,6 +245,16 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
            << "Invalid SPIR-V header.";
   }
 
+
+  // Look for OpExtension instructions and register extensions.
+  // Diagnostics if any will be produced in the next pass (ProcessInstruction).
+  if (auto error = spvBinaryParse(&context, vstate, words, num_words,
+                                  setHeader, ProcessExtensions,
```
dneto0: It would be clearer to use nullptr instead of setHeader here.  You only care about processing instructions, not the header.

source/validate.cpp
```diff
@@ -215,6 +245,16 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
...
+                                  /* diagnostic = */ nullptr)) {
+    if (error != SPV_REQUESTED_TERMINATION)
```
dneto0: But then you never get a diagnostic in this early return case.  You should just let it fall through.

source/validate_instruction.cpp
```diff
@@ -139,14 +154,33 @@ spv_result_t CapCheck(ValidationState_t& _,
...
+                            const spv_parsed_instruction_t* inst) {
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+  for (int i = 0; i < inst->num_operands; ++i) {
+    const auto& operand = inst->operands[i];
+    const uint32_t word = inst->words[operand.offset];
+    const ExtensionSet required_extensions =
+        RequiredExtensions(_, operand.type, word);
+    if (!_.HasAnyOfExtensions(required_extensions)) {
+      return _.diag(SPV_ERROR_MISSING_EXTENSION)
+          << "Operand " << word << " of " << spvOpcodeString(opcode)
```
dneto0: This could be ambiguous, in theory.  The word value  might be valid in another operand position.  At the very least you should print the operand index as well.

source/validate_instruction.cpp
```diff
@@ -139,14 +154,33 @@ spv_result_t CapCheck(ValidationState_t& _,
...
       }
     }
   }
   return SPV_SUCCESS;
 }
 
+// Checks that all required extensions were declared in the module.
+spv_result_t ExtensionCheck(ValidationState_t& _,
+                            const spv_parsed_instruction_t* inst) {
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
```
dneto0: In practice so far, extensions are written so that only OpCapability has an extensions list.  Then other features are enabled by that capability.
But this is ok.

source/validate.cpp
dneto0: Seems this one should do any logging at all. The diagnostics will go nowhere.

source/util/string_utils.h
```diff
@@ -0,0 +1,41 @@
...
+  if (mod10 == 1 && mod100 != 11)
+    suffix = "st";
+  else if (mod10 == 2 && mod100 != 12)
+    suffix = "nd";
+  else if (mod10 == 3 && mod100 != 13)
+    suffix = "rd";
+  else
+    suffix = "th";
+
+  return std::to_string(cardinal) + suffix;
```
dneto0: As I recall, this might not exist on Android, for reasons that won't be fixed soon.

I suggest you pull out the to_string from name_mapper.cpp into this file and make it a member of spvutils namespace.

Either that, or use a std::strinstream directly in this method instead of concatenating strings.

source/validate.cpp
```diff
@@ -117,6 +119,40 @@ void DebugInstructionPass(ValidationState_t& _,
...
+    return;
+  }
+
+  _.RegisterExtension(extension);
+}
+
+// Parses the beginning of the module searching for OpExtension instructions.
+// Registers extensions if recognized. Returns SPV_REQUESTED_TERMINATION
+// once an instruction which is not SpvOpCapability and SpvOpExtension is
+// encountered (according to the SPIR-V spec extensions are declared after
```
dneto0: The parenthetical remark is a full sentence.  So end the previous sentence with a period, and captializze "According..."

source/validate_instruction.cpp
```diff
@@ -139,14 +155,35 @@ spv_result_t CapCheck(ValidationState_t& _,
...
+  for (size_t operand_index = 0; operand_index < inst->num_operands;
+       ++operand_index) {
+    const auto& operand = inst->operands[operand_index];
+    const uint32_t word = inst->words[operand.offset];
+    const ExtensionSet required_extensions =
+        RequiredExtensions(_, operand.type, word);
+    if (!_.HasAnyOfExtensions(required_extensions)) {
+      return _.diag(SPV_ERROR_MISSING_EXTENSION)
+          << spvutils::CardinalToOrdinal(operand_index + 1) << " operand of "
```
dneto0: This could also have been "Operand " << operand_index << " of "...

test/string_utils_test.cpp
```diff
@@ -21,6 +21,35 @@
...
+  EXPECT_EQ("0", ToString(0U));
+  EXPECT_EQ("1000", ToString(1000U));
+  EXPECT_EQ("0", ToString(0ULL));
+  EXPECT_EQ("1000", ToString(1000ULL));
+}
+
+TEST(ToString, Float) {
+  EXPECT_EQ("0", ToString(0.f));
+  EXPECT_EQ("1000", ToString(1000.f));
+  EXPECT_EQ("-1.44", ToString(-1.44f));
```
dneto0: This value has a non-terminating fractional value in binary.  In general it's slightly risky to use such a value in an equality test.   E.g. I would have used something with a binary representation that terminated after fewer than 23 significand bits.

But this is fine, provided the bots pass. :-)

atgoo: Changed to -1.5

dneto0:
Squashed and pushed into master as b011633171

PR #585, opened 2017-03-15
Autogenererating enum-string mappings from grammar

atgoo:
Autogenerating the following code:
- extension enum
- extension-to-string
- string-to-extension
- capability-to-string

### DIFF
--- a/test/enum_string_mapping_test.cpp
+++ b/test/enum_string_mapping_test.cpp
@@ -28,7 +28,38 @@ using ::testing::Values;
 using ::testing::ValuesIn;
 
 using ExtensionTest = ::testing::TestWithParam<std::pair<Extension, std::string>>;
+using UnknownExtensionTest = ::testing::TestWithParam<std::string>;
 using CapabilityTest = ::testing::TestWithParam<std::pair<SpvCapability, std::string>>;
+
+TEST_P(ExtensionTest, TestExtensionFromString) {
+  const std::pair<Extension, std::string>& param = GetParam();
+  const Extension extension = param.first;
+  const std::string extension_str = param.second;
+  Extension result_extension;
+  ASSERT_TRUE(libspirv::GetExtensionFromString(extension_str, &result_extension));
+  EXPECT_EQ(extension, result_extension);
+}
+
+TEST_P(ExtensionTest, TestExtensionToString) {
+  const std::pair<Extension, std::string>& param = GetParam();
+  const Extension extension = param.first;
+  const std::string extension_str = param.second;
+  const std::string result_str = libspirv::ExtensionToString(extension);
+  EXPECT_EQ(extension_str, result_str);
+}
+
+TEST_P(UnknownExtensionTest, TestExtensionFromStringFails) {
+  Extension result_extension;
+  ASSERT_FALSE(libspirv::GetExtensionFromString(GetParam(), &result_extension));
+}
+
+TEST_P(CapabilityTest, TestCapabilityToString) {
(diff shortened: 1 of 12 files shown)


atgoo:
I want to redo the cmake part. Will reupload.

test/enum_string_mapping_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+TEST_P(ExtensionTest, TestExtensionToString) {
+  const std::pair<Extension, std::string> param = GetParam();
+  const Extension extension = param.first;
+  const std::string extension_str = param.second;
+  const std::string result_str = libspirv::ExtensionToString(extension);
+  EXPECT_EQ(extension_str, result_str);
+}
+
+TEST_P(CapabilityTest, TestCapabilityToString) {
+  const std::pair<SpvCapability, std::string> param = GetParam();
```
dneto0: This is ok, but could have been const ref.

test/enum_string_mapping_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+      {SpvCapabilityPerViewAttributesNV,
+       "PerViewAttributesNV"}
+    })));
+
+TEST_P(ExtensionTest, TestExtensionFromString) {
+  const std::pair<Extension, std::string> param = GetParam();
+  const Extension extension = param.first;
+  const std::string extension_str = param.second;
+  Extension result_extension;
+  ASSERT_TRUE(libspirv::GetExtensionFromString(extension_str, &result_extension));
```
dneto0: Also need failure cases for GetExtensionFromString

test/enum_string_mapping_test.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+using ::libspirv::Extension;
+
+using ::testing::Values;
+using ::testing::ValuesIn;
+
+using ExtensionTest = ::testing::TestWithParam<std::pair<Extension, std::string>>;
+using CapabilityTest = ::testing::TestWithParam<std::pair<SpvCapability, std::string>>;
+
+INSTANTIATE_TEST_CASE_P(AllExtensions, ExtensionTest,
```
dneto0: Typically we put the test class before the list of parameterized instances.

utils/generate_grammar_tables.py
```diff
@@ -348,6 +348,91 @@ def generate_operand_kind_table(enums):
...
+    function += '  };\n\n  return "";\n}'
+    return function
+
+
+def generate_string_to_extension_mapping(operands):
+    """Returns mapping function from strings to corresponding extensions."""
+    extensions = get_extension_list(operands)
+    function = 'bool GetExtensionFromString(' \
+        'const std::string& str, Extension* extension) {\n '
+    template = ' if (str == "{extension}")\n' \
```
dneto0: This bakes in the linear search.  If the number of extensions grows large then we might consider doing a regular STL lookup into an unordered map.  Suggest adding a TODO here.

At least this is no worse than the existing algorithm.

atgoo: Could also be an array of strings, accessible by index.

atgoo: The last comment was for extension->string.

dneto0: Yes.  An optimizing compiler can turn a switch on a dense enum into a table lookup or binary search.  At least it used to be that way.  :-)

dneto0:
Rebased, squashed, and pushed into master as c804c125c

PR #589, opened 2017-03-16
spirv-val prints info and warning messages to stdout

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/568

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -68,7 +68,7 @@ typedef enum spv_message_level_t {
                            // Will exit the program immediately. E.g.,
                            // unimplemented feature.
   SPV_MSG_ERROR,           // Normal error due to user input.
-  SPV_MSG_WARNINING,       // Warning information.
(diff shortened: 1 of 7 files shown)


dneto0:
Unfortunately, the spvDiagnostic mechanism doesn't allow you to send many messages.  So to get both an info an an error you have to resort to using the C++ interface, and use your own message consumer object.

include/spirv-tools/libspirv.h
```diff
@@ -68,7 +68,7 @@ typedef enum spv_message_level_t {
                            // Will exit the program immediately. E.g.,
                            // unimplemented feature.
   SPV_MSG_ERROR,           // Normal error due to user input.
-  SPV_MSG_WARNINING,       // Warning information.
```
antiagainst: Ouch, my bad. :(

dneto0:
Rebased.  This will also give Appveyor one more shot at testing this.  (Infra failure last time.)

Issue #588, opened 2017-03-16
spirv-val: wrong validation error reported when a pointer bitcast is attempted


sheredom:
A compiler I am working on output the following (incorrect for Vulkan) SPIR-V:

```
; SPIR-V
; Version: 1.0
; Generator: Codeplay; 0
; Bound: 27
; Schema: 0
        OpCapability Shader
        OpMemoryModel Logical GLSL450
        OpEntryPoint GLCompute %20 "foo"
        OpExecutionMode %20 LocalSize 1 1 1
        OpDecorate %4 ArrayStride 16
        OpMemberDecorate %5 0 Offset 0
        OpDecorate %5 BufferBlock
        OpDecorate %10 ArrayStride 16
        OpMemberDecorate %11 0 Offset 0
        OpDecorate %11 BufferBlock
        OpDecorate %18 DescriptorSet 0
        OpDecorate %18 Binding 0
        OpDecorate %19 DescriptorSet 0
        OpDecorate %19 Binding 1
%1 =    OpTypeInt 32 0
%2 =    OpTypeVector %1 4
%3 =    OpTypePointer Uniform %2
%4 =    OpTypeRuntimeArray %2
%5 =    OpTypeStruct %4
%6 =    OpTypePointer Uniform %5
%7 =    OpTypeFloat 32
%8 =    OpTypeVector %7 4
%9 =    OpTypePointer Uniform %8
%10 =   OpTypeRuntimeArray %8
%11 =   OpTypeStruct %10
%12 =   OpTypePointer Uniform %11
%13 =   OpTypeVoid
%14 =   OpTypeFunction %13
%15 =   OpConstant %1 0
%16 =   OpConstant %1 31
%17 =   OpConstantComposite %2 %16 %16 %16 %16
%18 =   OpVariable %6 Uniform
%19 =   OpVariable %12 Uniform
%20 =   OpFunction %13 None %14
%21 =   OpLabel
%22 =   OpAccessChain %3 %18 %15 %15
%23 =   OpAccessChain %9 %19 %15 %15
%24 =   OpBitcast %3 %23
%25 =   OpLoad %2 %24
%26 =   OpShiftRightArithmetic %2 %25 %17
        OpStore %22 %26
        OpReturn
        OpFunctionEnd
```

Note the `%24 =   OpBitcast %3 %23` - its trying to type pun an SSBO address!

The validator reports the following error:

> error: 157: OpLoad Pointer <id> '24' is not a pointer.

[...]

PR #590, opened 2017-03-16
Fix validator message for bad logical pointer

dneto0:
Affects OpLoad and OpStore validation.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/588


dneto0:
Rebased and pushed into master as 2206bb10ed59

PR #592, opened 2017-03-18
Only try to create the directory tree if the output file is not on the current directory

RenatoUtsch:
I'm trying to run the update_build_version.py script, and it is failing if I try to generate the output file to the current directory.

With this, it will only try to create the directory tree if it is not empty (the result of os.path.dirname("file_in_current_dir")).


dneto0:
Sorry, we can't accept changes without a CLA..
Fixed instead via https://github.com/KhronosGroup/SPIRV-Tools/pull/593

PR #593, opened 2017-03-20
Do nothing if trying to make the current dir

dneto0:


dneto0:
This is an alternative fix to https://github.com/KhronosGroup/SPIRV-Tools/pull/592
It solves the problem where it can occur rather than requiring all call sites to avoid the issue.  Sure, there's only one call site for now. But I like this better.

Issue #594, opened 2017-03-21
Missing support for disassembling vendor specific extensions (TimeAMD).


Jasper-Bekkers:
Related issue: https://github.com/google/shaderc/issues/325

I looked around a bit and it seems related to `spvOpcodeString` not having it's tables set up for vendor extensions.

The offending shader compiled with `shaderc`

```
#version 450

#extension GL_ARB_separate_shader_objects : enable
#extension GL_ARB_shading_language_420pack : enable
#extension GL_ARB_gpu_shader_int64 : enable
#extension GL_AMD_gcn_shader : enable

void main() 
{
	uint64_t v = timeAMD();
}
```

Jasper-Bekkers:
Hi David thanks for looking at this & the quick turnaround. After the shaderc issue I also run into this, when loading the SPIRV module with validation enabled:

`SPIR-V module not valid: Invalid extended instruction import 'SPV_AMD_gcn_shader'`

Does the above commit fix that as well?

dneto0:
> SPIR-V module not valid: Invalid extended instruction import 'SPV_AMD_gcn_shader'

> Does the above commit fix that as well?

Yes.  The validator recognizes the extension, but does not check the specific rules.  For the example you gave, the SPIR-V module passes validation silently.

PR #595, opened 2017-03-21
Support SPV_AMD_gcn_shader

dneto0:
Supported in assembler, disassembler, and binary parser.

The validator does not check SPV_AMD_gcn_shader validation rules
beyond parsing the extension.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/594


source/extinst.amd-gcn-shader.grammar.json
```diff
@@ -0,0 +1,23 @@
+{
+  "revision" : 2,
+  "instructions" : [
+    {
+      "opname" : "CubeFaceIndexAMD",
+      "opcode" : 1,
+      "operands" : [
+        { "kind" : "IdRef", "name" : "'P'" }
+      ]
```
antiagainst: Also add `"extensions" : ["SPV_AMD_gcn_shader"]`?

source/extinst.amd-gcn-shader.grammar.json
```diff
@@ -0,0 +1,23 @@
...
+      "opname" : "CubeFaceCoordAMD",
+      "opcode" : 2,
+      "operands" : [
+        { "kind" : "IdRef", "name" : "'P'" }
+      ]
+    },
+    {
+      "opname" : "TimeAMD",
+      "opcode" : 3
+    }
```
antiagainst: Also need `"capabilities" : ["Int64"]` here.

dneto0: That would mean that this feature is enabled by Int64.  The dependency goes the other way.  This will *require* the Int64 capability. The grammar file supports that kind of dependency only for Capability themselves.

source/extinst.amd-gcn-shader.grammar.json
dneto0: Hmm.  The grammar file documentation only describes the "extensions" field for ValueType enumerants.  Seems it should be added for Instruction objects and BitEnum as well.

I'll add the field for this PR anyway.

dneto0:
Feedback addressed. 
I have filed https://github.com/KhronosGroup/SPIRV-Tools/issues/596 to enhance the representation of extended instructions so they can depend on extensions, and not just capabilities.

antiagainst:
+2

Discussed about the `capabilities` field offline. It's more tricker than I originally thought. I'm happy with the current implementation and we can always fix up the grammar later since the capability is not used right now.

dneto0:
Rebased

PR #601, opened 2017-03-23
Update capability tests for SPV_KHR_16bit_storage rev 7

dneto0:
The SPIR-V core grammar file in a recent SPIRV-Headers
applied the fix from Rev 7 of SPV_KHR_16bit_storage:
FPRoundingMode enums are now enabled by the capabilities
introduced by that extension.

Update the SPIRV-Tools tests accordingly.


dneto0:
This is necessary to fix build breaks introduced by today's update to SPIRV-Headers repo.  Self-approving.

PR #603, opened 2017-03-26
Add various accessors needed to read and edit SPIRV code.

dsrbecky:


source/opt/basic_block.h
```diff
@@ -38,12 +38,14 @@ class BasicBlock {
   using const_iterator = UptrVectorIterator<Instruction, true>;
 
   // Creates a basic block with the given starting |label|.
-  inline explicit BasicBlock(std::unique_ptr<Instruction> label);
+  inline explicit BasicBlock(std::unique_ptr<Instruction> _label);
```
dneto0: I don't like this.  Surely nested scoping rules avoids the need to do this?!

source/opt/type_manager.h
```diff
@@ -62,20 +62,20 @@ class TypeManager {
...
-  void AnalyzeTypes(const spvtools::ir::Module& module);
-
   // Creates and returns a type from the given SPIR-V |inst|. Returns nullptr if
   // the given instruction is not for defining a type.
   Type* RecordIfTypeDefinition(const spvtools::ir::Instruction& inst);
   // Attaches the decoration encoded in |inst| to a type. Does nothing if the
   // given instruction is not a decoration instruction or not decorating a type.
   void AttachIfTypeDecoration(const spvtools::ir::Instruction& inst);
 
+ private:
```
dneto0: Unrelated to the functionality you are adding.

source/opt/types.cpp
```diff
@@ -179,8 +179,8 @@ std::string SampledImage::str() const {
   return oss.str();
 }
 
-Array::Array(Type* type, uint32_t length_id)
-    : element_type_(type), length_id_(length_id) {
+Array::Array(Type* type, uint32_t _lenght_id)
```
dneto0: Again.  Don't need this.  Also spelling. :-)

source/opt/types.h
```diff
@@ -265,6 +266,7 @@ class Struct : public Type {
 
   bool IsSame(Type* that) const override;
   std::string str() const override;
+  const std::vector<Type*> element_types() const { return element_types_; }
```
dneto0: Should be const ref.  If you want a copy, make a copy on the caller side.

source/opt/basic_block.h
dsrbecky: The scoping rules handle it correctly, but produces warning:
/ssd/go/src/github.com/google/gapid/third_party/SPIRV-Tools/source/opt/basic_block.h:71:65: error: declaration of ‘label’ shadows a member of 'this' [-Werror=shadow]

I do not like the underscore either though.  How about upper-casing the accessor method instead?

source/opt/types.h
```diff
@@ -231,6 +231,7 @@ class Array : public Type {
   bool IsSame(Type* that) const override;
   std::string str() const override;
   const Type* element_type() const { return element_type_; }
+  uint32_t LengthId() const { return length_id_; }
```
dneto0: This could have been length_id since it's a simple accessor.  But this is fine.

dsrbecky: It could not be lower case since otherwise "Array(Type* element_type, uint32_t length_id);" above would also result in compiler warning about shadowed variable.  I do not feel happy about either, but this feels like lesser evil than adding the underscores.

source/opt/type_manager.h
dsrbecky: PS: I have reverted this for now, and I will try to work around it in the client code.  However the motivation of making RecordIfTypeDefinition public was to enable adding of new types to the module.  As it is now, the type manager will get initialized at load time, but it will not see any types which get added to the module later...

PR #604, opened 2017-03-29
Dneto inline from mr534 by greg lunarg

dneto0:
Tiny fixup for https://github.com/KhronosGroup/SPIRV-Tools/pull/534

I'm not looking for a code review. :-)


dneto0:
Will wait for bots

dneto0:
I verified that this fixes the MSVC build failure.  It also passes its tests on Linux and Windows.

Will still wait for bots.

PR #606, opened 2017-03-31
Dont inline external functions

dneto0:


dneto0:
Cc: @greg-lunarg

source/opt/inline_pass.h
```diff
@@ -117,6 +117,9 @@ class InlinePass : public Pass {
                      ir::UptrVectorIterator<ir::Instruction> call_inst_itr,
                      ir::UptrVectorIterator<ir::BasicBlock> call_block_itr);
 
+  // Returns true if |inst| is a function call that can be inlined.
+  bool IsInlinableFunctionCall(const ir::Instruction* inst);
```
antiagainst: const function?

dneto0: The unordered map access causes it to be non-const.  I didn't bother to rewrite it with the iterator getter.  There's another similar id lookup in the GenInlineCode function that is the same.

antiagainst: Right... The const version of `operator[]` is `at`... But never mind.

Issue #602, opened 2017-03-23
Opt pass to flatten out group decoration


chrisforbes:
In LoaderAndValidationLayers we're starting to see things targeting SPIRV which use decoration groups. We'd like to be able to run a SPIRV->SPIRV pass before our validation which flattens all the variation out to simple OpDecorate/OpMemberDecorate.

dneto0:
Would you also be satisfied with an analysis that let you look up decorations affecting a target Id?

E.g.
``` 
     for ( auto& decoration : SpvAnalyzer.DecorationsTargeting(uint32_t target_id) ) {
         switch( decoration.type() ) {
          case OpDecorate:  ... inspect decoration and its potential operands;
          case OpMemberDecorate :  ... inspect member index, decoration and its potential operands;
         }
     }
```

chrisforbes:
@dneto0, that would be fine too.

dneto0:
I went with the option of transforming the SPIR-V module.  We don't have a public API for doing reflection or analysis. :-(

PR #610, opened 2017-04-02
Flatten decoration groups

dneto0:
Replaces OpDecorationGroup and its uses in OpGroupDecorate and OpMemberDecorate with repeated instances of OpDecorate and OpMemberDecorate.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/602

### DIFF
--- a/test/opt/flatten_decoration_test.cpp
+++ b/test/opt/flatten_decoration_test.cpp
@@ -1,0 +1,234 @@
+// Copyright (c) 2017 Valve Corporation
+// Copyright (c) 2017 LunarG Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gmock/gmock.h>
+
+#include "pass_fixture.h"
+#include "pass_utils.h"
+
+namespace {
+
+using namespace spvtools;
+
+// Returns the initial part of the assembly text for a valid
+// SPIR-V module, including instructions prior to decorations.
+std::string PreambleAssembly() {
+  return
+      R"(OpCapability Shader
+OpMemoryModel Logical GLSL450
+OpEntryPoint Fragment %main "main" %hue %saturation %value
+OpName %main "main"
+OpName %void_fn "void_fn"
+OpName %hue "hue"
+OpName %saturation "saturation"
+OpName %value "value"
+OpName %entry "entry"
+OpName %Point "Point"
+OpName %Camera "Camera"
+)";
+}
+
+// Retuns types
+std::string TypesAndFunctionsAssembly() {
+  return
+      R"(%void = OpTypeVoid
+%void_fn = OpTypeFunction %void
+%float = OpTypeFloat 32
+%Point = OpTypeStruct %float %float %float
+%Camera = OpTypeStruct %float %float
+%_ptr_Input_float = OpTypePointer Input %float
+%hue = OpVariable %_ptr_Input_float Input
+%saturation = OpVariable %_ptr_Input_float Input
+%value = OpVariable %_ptr_Input_float Input
+%main = OpFunction %void None %void_fn
+%entry = OpLabel
+OpReturn
+OpFunctionEnd
+)";
+}
+
+struct FlattenDecorationCase {
+  // Names and decorations before the pass.
+  std::string input;
+  // Names and decorations after the pass.
+  std::string expected;
+};
+
+using FlattenDecorationTest =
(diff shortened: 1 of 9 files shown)


test/opt/flatten_decoration_test.cpp
```diff
@@ -0,0 +1,183 @@
...
+                             "%group = OpDecorationGroup\n"
+                             "OpGroupMemberDecorate %group %Point 1\n",
+                             "OpMemberDecorate %Point 1 Flat\n"
+                             "OpMemberDecorate %Point 1 Offset 16\n"},
+                            // Multiple OpGroupDecorate
+                            {"OpName %group \"group\"\n"
+                             "OpDecorate %group Flat\n"
+                             "OpDecorate %group NoPerspective\n"
+                             "%group = OpDecorationGroup\n"
+                             "OpGroupDecorate %group %hue %value\n"
```
antiagainst: This is not `OpGroupMemberDecorate`?

test/opt/flatten_decoration_test.cpp
```diff
@@ -0,0 +1,183 @@
...
+                            // OpDecorationGroup without any uses, and
+                            // no OpName.
+                            {"%group = OpDecorationGroup\n", ""},
+
+                            // OpDecorationGroup without any uses, and
+                            // with OpName targeting it. Proves you must
+                            // remove the names as well.
+                            {"OpName %group \"group\"\n"
+                             "%group = OpDecorationGroup\n",
+                             ""},
```
antiagainst: `OpDecorationGroup` with `OpDecorate`s for it but no actual usage via `OpGroup{Member|}Decorate`?

dneto0: Yes, that's the purpose of this test case.  Maybe the comment needs improving?  The only use is by the OpName.

test/opt/flatten_decoration_test.cpp
dneto0: %hue and %value are scalars, so you use OpGroupDecorate on them.
Use OpGroupMemberDecorate when you want to apply a set of decorations to a member of a struct type.

test/opt/flatten_decoration_test.cpp
antiagainst: Sorry, was suggesting a new test case like:

```
OpDecorate %group <stuff>
OpDecorate %group <stuff>
%group = OpDecorationGroup
```

test/opt/flatten_decoration_test.cpp
antiagainst: Ah, sorry, this comment should be in the corresponding code snippet for `OpGroupMemberDecorate`.

test/opt/flatten_decoration_test.cpp
```diff
@@ -0,0 +1,183 @@
...
+                        ::testing::ValuesIn(std::vector<FlattenDecorationCase>{
+                            // One OpGroupDecorate
+                            {"OpName %group \"group\"\n"
+                             "OpDecorate %group Flat\n"
+                             "OpDecorate %group Offset 16\n"
+                             "%group = OpDecorationGroup\n"
+                             "OpGroupMemberDecorate %group %Point 1\n",
+                             "OpMemberDecorate %Point 1 Flat\n"
+                             "OpMemberDecorate %Point 1 Offset 16\n"},
+                            // Multiple OpGroupDecorate
```
antiagainst: I mean here actually.

test/opt/flatten_decoration_test.cpp
```diff
@@ -0,0 +1,183 @@
...
+                             "OpDecorate %group Location 42\n"
+                             "%group = OpDecorationGroup\n"
+                             "OpGroupDecorate %group %hue %saturation\n",
+                             "OpDecorate %hue Location 42\n"
+                             "OpDecorate %saturation Location 42\n"},
+                        }), );
+
+INSTANTIATE_TEST_CASE_P(OpGroupMemberDecorate, FlattenDecorationTest,
+                        ::testing::ValuesIn(std::vector<FlattenDecorationCase>{
+                            // One OpGroupDecorate
```
antiagainst: Needs to update the comment to `OpGroupMemberDecorate`.

dneto0:
Thanks again.  I'll take a look at the feedback on tests tomorrow.

test/opt/flatten_decoration_test.cpp
dneto0: Yes. Good catch.

test/opt/flatten_decoration_test.cpp
dneto0: This was a good test case.  Helped expose an infinite loop bug.

dneto0:
Addressed feedback. 
Added test cases of non-member and member decorations that are unrelated to group decorations. This found an infinite loop problem.

test/opt/flatten_decoration_test.cpp
```diff
@@ -180,4 +192,43 @@ INSTANTIATE_TEST_CASE_P(OpGroupMemberDecorate, FlattenDecorationTest,
                              "OpMemberDecorate %Point 2 Offset 16\n"},
                         }), );
 
+INSTANTIATE_TEST_CASE_P(UnrelatedDecorations, FlattenDecorationTest,
```
antiagainst: Nice!

dneto0:
Rebased and squashed into one commit.  Keeping approval.

Issue #611, opened 2017-04-02
spvtools::Optimizer's Run method fails to save the binary when the passes succeed without changes


dneto0:

dneto0:
E.g. when we use the Null pass, and the input binary storage area is not aliased to the output binary storage area, then the output binary storage area is not written to.

PR #612, opened 2017-04-02
Optimizer should save the binary when a pass succeeds without changes

dneto0:
Added a test for this too.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/611

### DIFF
--- a/source/opt/optimizer.cpp
+++ b/source/opt/optimizer.cpp
@@ -75,7 +75,10 @@ bool Optimizer::Run(const uint32_t* original_binary,
   if (module == nullptr) return false;
 
   auto status = impl_->pass_manager.Run(module.get());
-  if (status == opt::Pass::Status::SuccessWithChange) {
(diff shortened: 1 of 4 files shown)


dneto0:
Rebased

source/opt/optimizer.cpp
```diff
@@ -75,12 +75,13 @@ bool Optimizer::Run(const uint32_t* original_binary,
   if (module == nullptr) return false;
 
   auto status = impl_->pass_manager.Run(module.get());
-  if (status == opt::Pass::Status::SuccessWithChange) {
+  const bool succeeded = status != opt::Pass::Status::Failure;
+  if (succeeded) {
```
dneto0: Actually we can do a bit better.  We don't have to generate the binary if it's pass without change *and* the target binary storage aliases the input binary.

dneto0:
Fixed order of using decls, as noted in review for #613.
Keeping approval.

PR #613, opened 2017-04-02
Spirv opt uses optimizer class

dneto0:
This builds on https://github.com/KhronosGroup/SPIRV-Tools/pull/612

### DIFF
--- a/test/opt/optimizer_test.cpp
+++ b/test/opt/optimizer_test.cpp
@@ -1,0 +1,109 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gmock/gmock.h>
+
+#include "spirv-tools/libspirv.hpp"
+#include "spirv-tools/optimizer.hpp"
+
+#include "pass_fixture.h"
+
+namespace {
+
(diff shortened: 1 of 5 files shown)


tools/opt/opt.cpp
```diff
@@ -13,16 +13,16 @@
 // limitations under the License.
 
 #include <cstring>
+#include <algorithm>
```
antiagainst: Order

test/opt/optimizer_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <gmock/gmock.h>
+
+#include "spirv-tools/libspirv.hpp"
+#include "spirv-tools/optimizer.hpp"
+
+#include "pass_fixture.h"
+
+namespace {
+
+using spvtools::CreateStripDebugInfoPass;
```
antiagainst: Order

test/opt/optimizer_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+  Optimizer opt(SPV_ENV_UNIVERSAL_1_0);
+  opt.RegisterPass(CreateNullPass());
+  auto orig_size = binary.size();
+  // Now change the size.  Add a word that will be ignored
+  // by the optimizer.
+  binary.push_back(42);
+  EXPECT_THAT(orig_size + 1, Eq(binary.size()));
+  opt.Run(binary.data(), orig_size, &binary);  // This is the key.
+  // The binary vector should have been rewritten.
```
antiagainst: Nice!

tools/opt/opt.cpp
dneto0: C system files go before C++ files

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

:-)

dneto0:
Thanks for approval.  I'll fix the order of the using decl, keep approval. Then I'll submit after the bots come back green.

dneto0:
Rebased. Keeping approval. No need for review.  :-)

tools/opt/opt.cpp
antiagainst: TIrL! (Today I reLearned!)

PR #614, opened 2017-04-03
Add /EHs for targets for MSVC

antiagainst:


CMakeLists.txt
```diff
@@ -114,6 +114,12 @@ function(spvtools_default_compile_options TARGET)
     endif()
   endif()
 
+  if (MSVC)
+    # Specify /EHsc for exception handling. This makes using SPIRV-Tools as
+    # dependencies in other projects easier.
+    target_compile_options(${TARGET} PRIVATE /EHsc)
```
dneto0: So what happens if an allocation failure occurs deep inside SPIR-V Tools. Then an exception is thrown.  But most of the SPIRV-Tools API is extern "C".  What happens?

antiagainst: You are right. I haven't thought carefully about the indications of adding `/EHsc` flag. After reading the [spec](https://msdn.microsoft.com/en-us/library/1deeycx5.aspx) a little bit more, I think `/EHs` is more suitable for SPIRV-Tools?

dneto0: I think /EHs is better than /EHsc for SPIRV-Tools.  But does /EHs solve your problem?  If so then update the PR.  :-)

antiagainst: Yes, `/EHs` also works for me. Changed.

dneto0:
Rebased and pushed into master as ad3b08280d69

PR #618, opened 2017-04-06
Added statistical analysis tool (tool/stats)

atgoo:
Currently analyzes percentages of the following:
- version (how many 1.0 or 1.1)
- generator (how many were generated by this or that tool)
- extension (how many files have specific extensions)
- capability (how many files have specific capabilities)
- opcode prevalence

### DIFF
--- a/include/spirv-tools/libspirv.hpp
+++ b/include/spirv-tools/libspirv.hpp
@@ -22,6 +22,12 @@
 
 #include "spirv-tools/libspirv.h"
 
+namespace libspirv {
+
+struct SpirvStats;
+
+}  // namespace libspirv
+
 namespace spvtools {
 
 // Message consumer. The C strings for source and message are only alive for the
@@ -110,6 +116,9 @@ class SpirvTools {
   bool Validate(const uint32_t* binary, size_t binary_size) const;
   // Like the previous overload, but takes an options object.
   bool Validate(const uint32_t* binary, size_t binary_size, const ValidatorOptions& options) const;
+
+  bool AggregateStats(const uint32_t* binary, size_t binary_size,
+                      libspirv::SpirvStats* stats) const;
 
  private:
   struct Impl;  // Opaque struct for holding the data fields used by this class.
--- a/include/spirv-tools/libspirv.hpp
+++ b/include/spirv-tools/libspirv.hpp
@@ -22,12 +22,6 @@
 
 #include "spirv-tools/libspirv.h"
 
-namespace libspirv {
-
-struct SpirvStats;
-
-}  // namespace libspirv
-
 namespace spvtools {
 
 // Message consumer. The C strings for source and message are only alive for the
@@ -116,9 +110,6 @@ class SpirvTools {
   bool Validate(const uint32_t* binary, size_t binary_size) const;
   // Like the previous overload, but takes an options object.
   bool Validate(const uint32_t* binary, size_t binary_size, const ValidatorOptions& options) const;
-
-  bool AggregateStats(const uint32_t* binary, size_t binary_size,
-                      libspirv::SpirvStats* stats) const;
 
  private:
   struct Impl;  // Opaque struct for holding the data fields used by this class.
--- a/source/spirv_stats.cpp
+++ b/source/spirv_stats.cpp
@@ -1,0 +1,112 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "spirv_stats.h"
+
+#include <cassert>
+
+#include <algorithm>
+#include <string>
+#include <vector>
+
+#include "binary.h"
+#include "diagnostic.h"
+#include "enum_string_mapping.h"
+#include "extensions.h"
+#include "instruction.h"
+#include "opcode.h"
+#include "operand.h"
+#include "spirv-tools/libspirv.h"
+#include "spirv_endian.h"
+
+using libspirv::SpirvStats;
+
+namespace {
+
+spv_result_t ProcessHeader(
+    void* user_data, spv_endianness_t /* endian */, uint32_t /* magic */,
+    uint32_t version, uint32_t generator, uint32_t /* id_bound */,
+    uint32_t /* schema */) {
+  SpirvStats* stats =
+      reinterpret_cast<libspirv::SpirvStats*>(user_data);
+  ++stats->version_hist[version];
+  ++stats->generator_hist[generator];
+  return SPV_SUCCESS;
+}
+
+void ProcessCapability(SpirvStats* stats,
+                       const spv_parsed_instruction_t* inst) {
+  if (static_cast<SpvOp>(inst->opcode) != SpvOpCapability) return;
+  assert(inst->num_operands == 1);
+  const spv_parsed_operand_t& operand = inst->operands[0];
+  assert(operand.num_words == 1);
+  assert(operand.offset < inst->num_words);
+  const uint32_t capability = inst->words[operand.offset];
+  ++stats->capability_hist[capability];
+}
+
+void ProcessExtension(SpirvStats* stats,
+                      const spv_parsed_instruction_t* inst) {
+  if (static_cast<SpvOp>(inst->opcode) != SpvOpExtension) return;
+  const std::string extension = libspirv::GetExtensionString(inst);
+  ++stats->extension_hist[extension];
+}
+
+spv_result_t ProcessInstruction(
+    void* user_data, const spv_parsed_instruction_t* inst) {
+  SpirvStats* stats =
+      reinterpret_cast<libspirv::SpirvStats*>(user_data);
+
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+  ++stats->opcode_hist[opcode];
+
+  ProcessCapability(stats, inst);
+  ProcessExtension(stats, inst);
+
+  return SPV_SUCCESS;
+}
+
+}  // namespace
+
+namespace libspirv {
+
+spv_result_t AggregateStats(
+    const spv_context_t& context, const uint32_t* words, const size_t num_words,
+    spv_diagnostic* pDiagnostic, SpirvStats* stats) {
+  spv_const_binary_t binary = {words, num_words};
+
+  spv_endianness_t endian;
+  spv_position_t position = {};
+  if (spvBinaryEndianness(&binary, &endian)) {
+    return libspirv::DiagnosticStream(position, context.consumer,
+                                      SPV_ERROR_INVALID_BINARY)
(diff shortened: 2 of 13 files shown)


source/spirv_stats.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "opcode.h"
+#include "operand.h"
+#include "spirv-tools/libspirv.h"
+#include "spirv_endian.h"
+
+using libspirv::SpirvStats;
+
+namespace {
+
+spv_result_t ProcessHeader(
```
dneto0: Method comment

source/spirv_stats.cpp
```diff
@@ -0,0 +1,112 @@
...
+    uint32_t version, uint32_t generator, uint32_t /* id_bound */,
+    uint32_t /* schema */) {
+  SpirvStats* stats =
+      reinterpret_cast<libspirv::SpirvStats*>(user_data);
+  ++stats->version_hist[version];
+  ++stats->generator_hist[generator];
+  return SPV_SUCCESS;
+}
+
+void ProcessCapability(SpirvStats* stats,
```
dneto0: method comment

source/spirv_stats.cpp
```diff
@@ -0,0 +1,112 @@
...
+  if (static_cast<SpvOp>(inst->opcode) != SpvOpCapability) return;
+  assert(inst->num_operands == 1);
+  const spv_parsed_operand_t& operand = inst->operands[0];
+  assert(operand.num_words == 1);
+  assert(operand.offset < inst->num_words);
+  const uint32_t capability = inst->words[operand.offset];
+  ++stats->capability_hist[capability];
+}
+
+void ProcessExtension(SpirvStats* stats,
```
dneto0: method comment

source/spirv_stats.cpp
```diff
@@ -0,0 +1,112 @@
...
+                      const spv_parsed_instruction_t* inst) {
+  if (static_cast<SpvOp>(inst->opcode) != SpvOpExtension) return;
+  const std::string extension = libspirv::GetExtensionString(inst);
+  ++stats->extension_hist[extension];
+}
+
+spv_result_t ProcessInstruction(
```
dneto0: method coment

source/spirv_stats.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "spirv-tools/libspirv.hpp"
+
+namespace libspirv {
+
+struct SpirvStats {
```
dneto0: Comments. What do these fields store?

test/stats/stats_aggregate_test.cpp
```diff
@@ -0,0 +1,237 @@
...
+  spv_binary binary;
+  ASSERT_EQ(SPV_SUCCESS, spvTextToBinary(
+      ctx.context, code.c_str(), code.size(), &binary, nullptr));
+
+  ASSERT_EQ(SPV_SUCCESS, AggregateStats(*ctx.context, binary->code,
+                                        binary->wordCount, nullptr, stats));
+  spvBinaryDestroy(binary);
+}
+
+TEST(AggregateStats, Capability) {
```
dneto0: Test name could be better: AccumulateCapabilityCounts

test/stats/stats_aggregate_test.cpp
```diff
@@ -0,0 +1,237 @@
...
+  AggregateStats(code2, &stats);
+  EXPECT_EQ(5, stats.capability_hist.size());
+  EXPECT_EQ(2, stats.capability_hist.at(SpvCapabilityShader));
+  EXPECT_EQ(2, stats.capability_hist.at(SpvCapabilityAddresses));
+  EXPECT_EQ(2, stats.capability_hist.at(SpvCapabilityKernel));
+  EXPECT_EQ(2, stats.capability_hist.at(SpvCapabilityGenericPointer));
+  EXPECT_EQ(4, stats.capability_hist.at(SpvCapabilityLinkage));
+}
+
+TEST(AggregateStats, Extension) {
```
dneto0: Also, name could be AccumulateExtensionCounts

test/stats/stats_aggregate_test.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+TEST(AggregateStats, Generator) {
+  const std::string code1 = R"(
+OpCapability Shader
+OpCapability Linkage
+OpMemoryModel Logical GLSL450
+)";
+
+  constexpr uint32_t kGeneratorKhronosAssembler =
```
dneto0: This could just be const.  :-)

test/stats/stats_aggregate_test.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include "source/spirv_stats.h"
+#include "test_fixture.h"
+#include "unit_spirv.h"
+
+namespace {
+
+using libspirv::SpirvStats;
+using spvtest::ScopedContext;
+
+void AggregateStats(const std::string& code, SpirvStats* stats,
```
dneto0: It's very confusing that this helper function has the same name as the thing you are testing.

dneto0: Function needs a comment.

source/spirv_stats.cpp
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  spv_header_t header;
+  if (spvBinaryHeaderGet(&binary, endian, &header)) {
+    return libspirv::DiagnosticStream(position, context.consumer,
+                                      SPV_ERROR_INVALID_BINARY)
+        << "Invalid SPIR-V header.";
+  }
+
+  if (auto error = spvBinaryParse(&context, stats, words, num_words,
```
dneto0: Could skip the if entirely and just do "return spvBinaryParse...."

test/stats/stats_analyzer_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+#include "spirv/1.1/spirv.h"
+#include "test_fixture.h"
+#include "tools/stats/stats_analyzer.h"
+
+namespace {
+
+using libspirv::SpirvStats;
+
+void FillDefaultStats(SpirvStats* stats) {
```
dneto0: Function comment.

test/stats/stats_analyzer_test.cpp
```diff
@@ -0,0 +1,140 @@
...
+  stats->version_hist[0x00010000] = 40;
```
dneto0: This is a little fragile.  What if more members are added to a SpirvStats.
Should this method clear stats first?  E.g
```*stats = SpirvStats();```

File: tools/stats/stats.cpp [hunk 13]
```diff
@@ -0,0 +1,195 @@
...
+#include "stats_analyzer.h"
+#include "tools/io.h"
+
+using libspirv::SpirvStats;
+
+namespace {
+
+std::vector<std::string> g_input_files;
+
+void AddInputFile(const char* path) {
```
dneto0: Method comment.

tools/stats/stats.cpp
```diff
@@ -0,0 +1,195 @@
...
+void PrintUsage(char* argv0) {
+  printf(
+      R"(%s - Collect statistics from one or more SPIR-V binary file(s).
+
+USAGE: %s [options] [<filepaths>] [<dirpaths>]
+
+The SPIR-V binaries read from the given <filepaths> combined with the files
+found in the given <dirpaths> and their subdirectories.
+
+<dirpaths> feature is currently only implmeneted in Linux.
```
dneto0: "implemented"

tools/stats/stats.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include "source/spirv_stats.h"
+#include "spirv-tools/libspirv.hpp"
+#include "stats_analyzer.h"
+#include "tools/io.h"
+
+using libspirv::SpirvStats;
+
+namespace {
+
+std::vector<std::string> g_input_files;
```
dneto0: Don't use a global variable.  Pass the vector around by pointer.

tools/stats/stats.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+void AddInputFile(const char* path) {
+  g_input_files.push_back(path);
+}
+
+#ifdef __linux__
+int AddInputFile(const char* path, const struct stat*, int tflag, struct FTW*) {
```
dneto0: Method comment.

tools/stats/stats.cpp
```diff
@@ -0,0 +1,195 @@
...
+#ifdef __linux__
+  nftw(path, AddInputFile, 20, FTW_PHYS);
+#endif  // __linux__
+}
+
+void PrintUsage(char* argv0) {
+  printf(
+      R"(%s - Collect statistics from one or more SPIR-V binary file(s).
+
+USAGE: %s [options] [<filepaths>] [<dirpaths>]
```
dneto0: So this will include any regular file.  So if I dropped in a "README" somewhere it would try to be interpreted.

I would have opted to let the shell find the files for me.
For example:

```
find . -name "*.spv" -print0 | xargs -0 spirv-stats
```
Or if you're worried about command line length limits (which can be a problem on older Windows shells), then allow the user to list the files on stdin.
E.g.

```
find . -name "*.spv" | spirv-stats -
```

tools/stats/stats.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  spvtools::SpirvTools tools(SPV_ENV_UNIVERSAL_1_1);
+  tools.SetMessageConsumer(DiagnosticsMessageHandler);
+
+  libspirv::SpirvStats stats;
+
+  for (size_t index = 0; index < g_input_files.size(); ++index) {
+    constexpr size_t kMilestonePeriod = 1000;
+    if (index % kMilestonePeriod == kMilestonePeriod - 1)
+      std::cerr << "Processed " << index + 1 << " files..." << std::endl;
```
dneto0: This is not an error. So why is it going to cerr?

Recommend prefixing this by the program name, in case this is used with a lot of other tools. Also, guard it with a "-v" for "verbose" option.

include/spirv-tools/libspirv.hpp
```diff
@@ -111,6 +117,9 @@ class SpirvTools {
   // Like the previous overload, but takes an options object.
   bool Validate(const uint32_t* binary, size_t binary_size, const ValidatorOptions& options) const;
 
+  bool AggregateStats(const uint32_t* binary, size_t binary_size,
```
dneto0: But more importantly, why is this in the public header when you can't get any info out of it?  In fact you can't even call it if you just have the public header, because you can't form a valid non-null pointer to a SpirvStats object.

You should move this whole entry point inside of a header under source.  It's not something external code can use.

dneto0:
FYI build failure was caused by push to SPIRV-Headers yesterday triggering https://github.com/KhronosGroup/SPIRV-Tools/issues/620
Should be fixed by https://github.com/KhronosGroup/SPIRV-Tools/pull/621

test/stats/stats_aggregate_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+  const std::string code2 = R"(
+OpCapability Shader
+OpCapability Linkage
+OpMemoryModel Logical GLSL450
+)";
+
+  SpirvStats stats;
+
+  CompileAndAggregateStats(code1, &stats);
+  EXPECT_EQ(4, stats.capability_hist.size());
```
dneto0: Signed vs. unsigned comparison error in Clang.  Recommend you add "u" suffix to all these literal numbers that compare to counts.

dneto0:
Squashed and pushed into master as 4f216402b

Issue #597, opened 2017-03-21
Enable disassembly in presence of unknown extensions and unknown instructions


dneto0:
Requires extending the operand type enum to include "unknown word" concept.  The words will not register as IDs, and so aren't tracked as types in the binary parser.
The validator can't know what the values are, and so things like SSA checks could fail.  That's life.

dneto0:
Requested by @johnkslang

johnkslang:
FYI, background information: Quoting the SPIR-V specification:

> SPIR-V can be extended by multiple vendors or parties simultaneously:

> - Using the OpExtension instruction to require new semantics that must be supported. Such new semantics would come from an extension document.

> - Reserving (registering) ranges of the token values, as described further below.

> - Aided by instruction skipping, also further described below.

> ...

> **Instruction Skipping.** Tools are encouraged to skip opcodes for features they are not required to process. This is trivially enabled by the word count in an instruction, which makes it easier to add new instructions without breaking existing tools.

johnkslang:
About SSA:

- duplicate definition errors should still be done the same way; skipping instructions does not introduce duplication, so any duplication is still an error
- missing definition, when there are unknown instructions, can be a warning, in case an unknown instruction defined missing definition

This is further aided by the SSA values being dense and near 0, and having the maximum value declared in the header: you can always have a table with an entry for all potential SSA ids, defined or not, to track information in.

dneto0:
I think the following syntax is easy to use and fits in with the current code.
For an unknown instruction opcode, say 999, use syntax like:
```Op999 12 13 14 15``` where 12 13 14 15 are literal number operands.  You can have as many operands as you like, but they're encoded as literal numbers.

I think this is an easy enough update to the code without having to change how we make everything else table-driven. 

I think the "strict" mode of parsing can be applied at the validator layer.  The validator API already can take an opaque options argument so I can enable both relaxed and strict behaviours.    I think I prefer that the default remain "strict", and add an option to enable relaxed mode.

dneto0:
@AWoloszyn pointed out that the Op999 syntax was not very useful alone because the assembler maps Ids to sequential numbers, and if your new unknown instruction uses Ids then those Ids could have been moved.  So in the end a Op999 would likely only be truly useful if the entire module was of that form.

But we can sort of fix things by adding another feature where the assembler requires Ids to only be numbers, and then map those Id numbers faithfully into the binary.
[...]

PR #623, opened 2017-04-07
Assembler supports Op999 syntax for unknown instructions

dneto0:
Working toward https://github.com/KhronosGroup/SPIRV-Tools/issues/597

### DIFF
--- a/source/text.cpp
+++ b/source/text.cpp
@@ -469,6 +469,18 @@ spv_result_t encodeInstructionStartingWithImmediate(
   return SPV_SUCCESS;
 }
 
+// If |str| denotes a number between 0 and 65535, then return that number.
+// Otherwise return 0.
+static int As16BitNumber(const char* str) {
+  int result = 0;
+  char c;
+  // Succeed only if we parse one parameter
+  if ((1 == std::sscanf(str, "%u%c", &result, &c)) && result <= 0xffff)
+    return result;
+  // Failure case.
+  return 0;
+}
+
 /// @brief Translate single Opcode and operands to binary form
 ///
 /// @param[in] grammar the grammar to use for compilation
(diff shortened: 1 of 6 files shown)


dneto0:
@antiagainst has been doing almost all my SPIRV-Tools reviews lately. We should spread the load.
Also, @atgoo and @ehsannas could learn a bit how the assembler works and is tested.

dneto0:
Rebased

antiagainst:
Ah, sorry, missed this one. I'll review it this afternoon.

source/text.cpp
```diff
@@ -530,11 +542,35 @@ spv_result_t spvTextEncodeOpcode(const libspirv::AssemblyGrammar& grammar,
...
   const char* pInstName = opcodeName.data() + 2;
 
-  spv_opcode_desc opcodeEntry;
-  error = grammar.lookupOpcode(pInstName, &opcodeEntry);
-  if (error) {
-    return context->diagnostic(error) << "Invalid Opcode name '" << opcodeName
-                                      << "'";
+  spv_opcode_desc opcodeEntry = nullptr;
+
+  // If it's an unknown opcode, then store its description the object
```
antiagainst: This sentence reads strangely for me. Something is missing or redundant.

source/text.cpp
```diff
@@ -530,11 +542,35 @@ spv_result_t spvTextEncodeOpcode(const libspirv::AssemblyGrammar& grammar,
...
+  std::unique_ptr<spv_opcode_desc_t> unknown_opcode_desc = nullptr;
+  // First see if it's parseable as a non-zero 16 bit number. OpNop has
+  // opcode zero, and if you want that then just say OpNop instead of Op0.
+  if (auto unknown_opcode = As16BitNumber(pInstName)) {
+    // Support this as an unknown instruction.  Don't make an opcode name
+    // for it.  Assume it:
+    //  - does not have a type
+    //  - does not generate a result Id
+    //  - has zero or more literal number operands.
+    unknown_opcode_desc.reset(
```
antiagainst: I think if we gonna do this, it's better to check whether this number is the opcode value for some known instruction. 

This feature itself is already very scary. Instead of leaving the users totally unprotected, we should try our best to guard them against some silly mistakes.

test/text_to_binary_test.cpp
```diff
@@ -266,4 +266,45 @@ TEST(CreateContext, VulkanEnvironment) {
...
+}
+
+TEST_F(TextToBinaryTest, FailOnOpcodeWithTooBigNumber) {
+  EXPECT_EQ("Invalid Opcode name 'Op65536'", CompileFailure("Op65536"));
+}
+
+TEST_F(TextToBinaryTest, FailOnOpcodeWithExplicitZero) {
+  EXPECT_EQ("Invalid Opcode name 'Op0'", CompileFailure("Op0"));
+}
+
```
antiagainst: Test cases for:

* `Op9.9`, `Op-999`?
* Normal instructions taking variable number of parameters together with unknown instructions like `OpGroupDecorate %group %1 %2 %3; Op<number>`?
* Unknown instructions having the same opcode number as a known instruction?

source/text.cpp
```diff
@@ -469,6 +469,18 @@ spv_result_t encodeInstructionStartingWithImmediate(
   return SPV_SUCCESS;
 }
 
+// If |str| denotes a number between 0 and 65535, then return that number.
+// Otherwise return 0.
+static int As16BitNumber(const char* str) {
+  int result = 0;
+  char c;
+  // Succeed only if we parse one parameter
+  if ((1 == std::sscanf(str, "%u%c", &result, &c)) && result <= 0xffff)
```
atgoo: %u requires an unsigned int pointer and would also likely succeed on a negative number.

Consider
if ((1 == std::sscanf(str, "**%d**%c", &result, &c)) && **result >= 0** && result <= 0xffff)

atgoo:
Can we close this?

PR #627, opened 2017-04-12
Added option --preserve-numeric-ids to tools/spirv-as

atgoo:
Options --preserve-numeric-ids or SPV_TEXT_TO_BINARY_OPTION_PRESERVE_NUMERIC_IDS preserve original numeric ids during compilation.


test/raw_ids_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+  std::string after;
+  EXPECT_EQ(SPV_SUCCESS, ToBinaryAndBack(before, &after,
+                                         SPV_TEXT_TO_BINARY_OPTION_RAW_IDS,
+                                         SPV_BINARY_TO_TEXT_OPTION_NO_HEADER));
+
+  EXPECT_EQ(expected, after);
+}
+
+TEST(TextHandler, NumericIds) {
```
dneto0: How is this different from the previous test?

source/text.cpp
```diff
@@ -659,10 +659,12 @@ spv_result_t SetHeader(spv_target_env env, const uint32_t bound,
   return SPV_SUCCESS;
 }
 
-spv_result_t spvTextGetNumericIds(const libspirv::AssemblyGrammar& grammar,
-                                  const spvtools::MessageConsumer& consumer,
-                                  const spv_text text,
-                                  std::set<uint32_t>* numeric_ids) {
+// Collects all numeric ids in the module sorce into |numeric_ids|.
```
dneto0: "source"

include/spirv-tools/libspirv.h
```diff
@@ -246,7 +246,10 @@ typedef enum spv_number_kind_t {
 
 typedef enum spv_text_to_binary_options_t {
   SPV_TEXT_TO_BINARY_OPTION_NONE = SPV_BIT(0),
```
dneto0: Pre-existing problem, similar for disassembler options.  NONE should have expanded bitmask value of 0, not 1.  :-)
Don't fix it, it would be confusing.

dneto0:
Squashed, typo fixed, and pushed into master as b173d1c3c

PR #630, opened 2017-04-13
Added --compact-ids to /tools/opt

atgoo:
The pass remaps ids to a compact set which starts with %1 and
has no gaps.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -186,8 +186,7 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 Optimizer::PassToken CreateInlinePass();
 
 // Creates a compact ids pass.
-// The pass remaps result ids to a compact set which starts from %1 and doesn't
-// have any gaps.
+// The pass remaps result ids to a compact and gapless range starting from %1.
 Optimizer::PassToken CreateCompactIdsPass();
 
 }  // namespace spvtools
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -185,6 +185,11 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 // points are not changed.
 Optimizer::PassToken CreateInlinePass();
 
+// Creates a compact ids pass.
+// The pass remaps result ids to a compact set which starts from %1 and doesn't
+// have any gaps.
+Optimizer::PassToken CreateCompactIdsPass();
+
 }  // namespace spvtools
 
 #endif  // SPIRV_TOOLS_OPTIMIZER_HPP_
--- a/source/opt/compact_ids_pass.cpp
+++ b/source/opt/compact_ids_pass.cpp
@@ -1,0 +1,66 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "compact_ids_pass.h"
+
+#include <cassert>
+#include <unordered_map>
+
+namespace spvtools {
+namespace opt {
+
+using ir::Instruction;
+using ir::Operand;
+
+Pass::Status CompactIdsPass::Process(ir::Module* module) {
+  bool modified = false;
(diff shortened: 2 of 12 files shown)


include/spirv-tools/optimizer.hpp
```diff
@@ -185,6 +185,11 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 // points are not changed.
 Optimizer::PassToken CreateInlinePass();
 
+// Creates a compact ids pass.
+// The pass remaps result ids to a compact set which starts from %1 and doesn't
```
dneto0: Prefer "compact range" since I always read "compact set" as "closed and bounded".  Too much topology background I guess. :-)

source/opt/compact_ids_pass.cpp
```diff
@@ -0,0 +1,66 @@
...
+using ir::Operand;
+
+Pass::Status CompactIdsPass::Process(ir::Module* module) {
+  bool modified = false;
+  std::unordered_map<uint32_t, uint32_t> result_id_mapping;
+
+  module->ForEachInst([&result_id_mapping, &modified] (Instruction* inst) {
+    auto operand = inst->begin();
+    while (operand != inst->end()) {
+      switch (operand->type) {
```
dneto0: Since this is the only use for the switch I prefer you use the centralized function spvIsIdType from operand.h

source/opt/compact_ids_pass.cpp
```diff
@@ -0,0 +1,66 @@
...
+using ir::Instruction;
+using ir::Operand;
+
+Pass::Status CompactIdsPass::Process(ir::Module* module) {
+  bool modified = false;
+  std::unordered_map<uint32_t, uint32_t> result_id_mapping;
+
+  module->ForEachInst([&result_id_mapping, &modified] (Instruction* inst) {
+    auto operand = inst->begin();
+    while (operand != inst->end()) {
```
dneto0: A for loop would have been slightly more idiomatic.

```
for (auto operand = inst->begin() ; operand != inst->end() ; ++operand ) { 
...
}
```

test/opt/compact_ids_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+  SetAssembleOptions(SPV_TEXT_TO_BINARY_OPTION_PRESERVE_NUMERIC_IDS);
+  SetDisassembleOptions(SPV_BINARY_TO_TEXT_OPTION_NO_HEADER);
+  SinglePassRunAndCheck<opt::NullPass>(before, after, false, false);
+}
+
+TEST_F(CompactIdsTest, PassOn) {
+  const std::string before =
+R"(OpCapability Addresses
+OpCapability Kernel
+OpCapability GenericPointer
```
dneto0: Should include a test case where a use occurs before the definition.
E.g. something with an OpEntryPoint, since the OpEntryPoint instruction appears before the function it describes.

test/scripts/test_compact_ids.py
```diff
@@ -0,0 +1,104 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests correctness of opt pass tools/opt --compact-ids."""
+
+from __future__ import print_function
+
+import os.path
+import sys
+
+TEST_HOME_PATH = '/tmp/test_compact_ids'
```
dneto0: This is not safe when multiple builds are on the same machine.
You should have a program argument tell you where you can write files.  The test target in CMakeLists should specify a directory in the build output tree. e.g. a special directoroy under ${CMAKE_CURRENT_BIN_DIR}

dneto0: Oh, this isn't used anywhere?  Hm.. I'd recommend that TEST_HOME_PATH default to a directory under the current directory.  

Or use tempfile.mkdtemp

https://docs.python.org/2/library/tempfile.html 

But then you'll want an option to keep the directory if something fails.

tools/opt/opt.cpp
```diff
@@ -66,6 +66,9 @@ NOTE: The optimizer is a work in progress.
   --flatten-decorations
                Replace decoration groups with repeated OpDecorate and
                OpMemberDecorate instructions.
+  --compact-ids
+               Remap result ids to a compact set starting from %%1 and without
```
dneto0: compact range  :-)

source/opt/compact_ids_pass.cpp
atgoo: I don't see any iterators which start at the beginning of the module and iterate through every instruction.

test/scripts/test_compact_ids.py
atgoo: If something fails, then the case can be easily reproduced with a few bash commands.

source/opt/compact_ids_pass.cpp
dneto0: This is about iterating over operands in a single instruction, not all instructions in the module.

I meant that a single for loop would replace lines 31, 32, and 49.

Relatively minor.

dneto0:
Squashed and pushed into master as 1e309af80a44

Issue #628, opened 2017-04-12
Optimization: set-spec-constant-default-value pass should have an interface to accept bit patterns.


Qining:

dneto0:
Fixed by #632

PR #632, opened 2017-04-19
Add bit pattern interface for setting spec constants default values

Qining:
Add another constructor which accepts map<spec_id, bit_pattern> as input argument. Bit patterns are fixed to vector<uint32_t> type.

### DIFF
--- a/source/opt/set_spec_constant_default_value_pass.cpp
+++ b/source/opt/set_spec_constant_default_value_pass.cpp
@@ -241,34 +241,32 @@ Pass::Status SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
     if (!spec_inst) continue;
 
     // Get the default value bit pattern for this spec id.
-    std::vector<uint32_t> bit_pattern = [this, &spec_inst, &spec_id,
-                                         &type_mgr]() {
-      std::vector<uint32_t> empty_bit_pattern;
-      if (spec_id_to_value_str_.size() != 0) {
-        // Search for the new string-form default value for this spec id.
-        auto iter = spec_id_to_value_str_.find(spec_id);
-        if (iter == spec_id_to_value_str_.end()) {
-          return empty_bit_pattern;
-        }
+    std::vector<uint32_t> bit_pattern;
 
-        // Gets the string of the default value and parses it to bit pattern
-        // with the type of the spec constant.
-        const std::string& default_value_str = iter->second;
-        return ParseDefaultValueStr(default_value_str.c_str(),
-                                    type_mgr.GetType(spec_inst->type_id()));
-
-      } else {
-        // Search for the new bit-pattern-form default value for this spec id.
-        auto iter = spec_id_to_value_bit_pattern_.find(spec_id);
-        if (iter == spec_id_to_value_bit_pattern_.end()) {
-          return empty_bit_pattern;
-        }
-
-        // Gets the bit-pattern of the default value from the map directly.
-        return ParseDefaultValueBitPattern(
-            iter->second, type_mgr.GetType(spec_inst->type_id()));
+    if (spec_id_to_value_str_.size() != 0) {
+      // Search for the new string-form default value for this spec id.
+      auto iter = spec_id_to_value_str_.find(spec_id);
+      if (iter == spec_id_to_value_str_.end()) {
+        continue;
       }
-    }();
+
+      // Gets the string of the default value and parses it to bit pattern
+      // with the type of the spec constant.
+      const std::string& default_value_str = iter->second;
+      bit_pattern = ParseDefaultValueStr(default_value_str.c_str(),
+                                  type_mgr.GetType(spec_inst->type_id()));
+
+    } else {
+      // Search for the new bit-pattern-form default value for this spec id.
+      auto iter = spec_id_to_value_bit_pattern_.find(spec_id);
+      if (iter == spec_id_to_value_bit_pattern_.end()) {
+        continue;
+      }
+
+      // Gets the bit-pattern of the default value from the map directly.
+      bit_pattern = ParseDefaultValueBitPattern(
+          iter->second, type_mgr.GetType(spec_inst->type_id()));
+    }
 
     if (bit_pattern.empty()) continue;
 
--- a/source/opt/set_spec_constant_default_value_pass.cpp
+++ b/source/opt/set_spec_constant_default_value_pass.cpp
@@ -74,6 +74,7 @@ std::vector<uint32_t> ParseDefaultValueStr(const char* text,
   }
   return result;
 }
+
 // Given a bit pattern and a type, checks if the bit pattern is compatible
 // with the type. If so, returns the bit pattern, otherwise returns an empty
 // bit pattern. If the given bit pattern is empty, returns an empty bit
--- a/source/opt/set_spec_constant_default_value_pass.cpp
+++ b/source/opt/set_spec_constant_default_value_pass.cpp
@@ -14,6 +14,7 @@
 
 #include "set_spec_constant_default_value_pass.h"
 
+#include <algorithm>
 #include <cctype>
 #include <cstring>
 #include <tuple>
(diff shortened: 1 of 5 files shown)


Qining:
This CL is to resolve #628

source/opt/set_spec_constant_default_value_pass.h
```diff
@@ -78,8 +89,10 @@ class SetSpecConstantDefaultValuePass : public Pass {
       const char* str);
 
  private:
-  // The mapping from spec ids to their default values to be set.
-  const SpecIdToValueStrMap spec_id_to_value_;
+  // The mapping from spec ids to their string-form default values to be set.
```
dneto0: You should say that exactly one of these two collections will be populated.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -200,15 +240,36 @@ Pass::Status SetSpecConstantDefaultValuePass::Process(ir::Module* module) {
     }
     if (!spec_inst) continue;
 
-    // Search for the new default value for this spec id.
-    auto iter = spec_id_to_value_.find(spec_id);
-    if (iter == spec_id_to_value_.end()) continue;
+    // Get the default value bit pattern for this spec id.
+    std::vector<uint32_t> bit_pattern = [this, &spec_inst, &spec_id,
```
dneto0: This is ok.

It feels a bit like overkill to use a lambda like this. You could have (a) declared the bit_pattern variable, then (b) put the rest of this code inside braces, to introduce a new block scope.  And instead of using "return", just assign to the bit_pattern variable.

But again, this is ok.

If you keep this use of the lambda, you could make bit_pattern const.  :-)  That's an argument *for* using the lambda.

test/opt/set_spec_const_default_value_test.cpp
```diff
@@ -537,6 +541,519 @@ INSTANTIATE_TEST_CASE_P(
             "%int = OpTypeInt 32 1\n"
             "%int_100 = OpConstant %int 100\n",
         },
+        // 6. Boolean type spec constant cannot be set with numeric values in
```
dneto0: Good comment.

test/opt/set_spec_const_default_value_test.cpp
```diff
@@ -537,6 +541,519 @@ INSTANTIATE_TEST_CASE_P(
...
+            "%1 = OpSpecConstant %float 200\n"
+            "%2 = OpSpecConstant %float 201\n",
+            // default values
+            SpecIdToValueBitPatternMap{{101, {0xffffffff}},
+                                       {102, {0x40200000}}},
+            // expected
+            "OpDecorate %1 SpecId 101\n"
+            "OpDecorate %2 SpecId 102\n"
+            "%float = OpTypeFloat 32\n"
+            "%1 = OpSpecConstant %float -0x1.fffffep+128\n"
```
dneto0: Nice.  I'm so glad we were very careful with printing float numbers with the HexFloat class.  :-)

test/opt/set_spec_const_default_value_test.cpp
```diff
@@ -537,6 +541,519 @@ INSTANTIATE_TEST_CASE_P(
...
+            // default values
+            SpecIdToValueBitPatternMap{{100, {0xffffffff}}},
+            // expected
+            "OpDecorate %1 SpecId 100\n"
+            "%1 = OpDecorationGroup\n"
+            "OpGroupDecorate %1 %2 %2 %2\n"
+            "%int = OpTypeInt 32 1\n"
+            "%2 = OpSpecConstant %int -1\n",
+        },
+        // 18. For Boolean type spec constants,if any word in the bit pattern
```
dneto0: Cool!

test/opt/set_spec_const_default_value_test.cpp
```diff
@@ -537,6 +541,519 @@ INSTANTIATE_TEST_CASE_P(
...
+            "OpDecorate %1 SpecId 100\n"
+            "%int = OpTypeInt 32 1\n",
+            // default values
+            SpecIdToValueBitPatternMap{{100, {0x7fffffff}}},
+            // expected
+            "OpDecorate %1 SpecId 100\n"
+            "%int = OpTypeInt 32 1\n",
+        },
+        // 2. Do nothing when SpecId decoration is not attached to a
+        // non-spec-contant instruction.
```
dneto0: "constant"

test/opt/set_spec_const_default_value_test.cpp
```diff
@@ -537,6 +541,519 @@ INSTANTIATE_TEST_CASE_P(
...
+            "%1 = OpSpecConstantOp %int IAdd %3 %3\n",
+            // default values
+            SpecIdToValueBitPatternMap{{100, {0x7fffffff}}},
+            // expected
+            "OpDecorate %1 SpecId 100\n"
+            "%int = OpTypeInt 32 1\n"
+            "%3 = OpSpecConstant %int 101\n"
+            "%1 = OpSpecConstantOp %int IAdd %3 %3\n",
+        },
+        // 4. Do not crash and do nothing when SpecId decoration is applied to
```
dneto0: I'm glad this is a no-op.
I suspect that's an invalid scenario, but it probably falls into a gray area.

test/opt/set_spec_const_default_value_test.cpp
```diff
@@ -537,6 +541,519 @@ INSTANTIATE_TEST_CASE_P(
...
+            // expected
+            "OpDecorate %1 SpecId 100\n"
+            "%1 = OpDecorationGroup\n"
+            "OpGroupDecorate %1 %2 %3 %4\n"
+            "%int = OpTypeInt 32 1\n"
+            "%2 = OpSpecConstant %int 100\n"
+            "%3 = OpSpecConstant %int 200\n"
+            "%4 = OpSpecConstant %int 300\n",
+        },
+        // 5. Do not crash and do nothing when SpecId decoration is attached to
```
dneto0: Good

test/opt/set_spec_const_default_value_test.cpp
```diff
@@ -537,6 +541,519 @@ INSTANTIATE_TEST_CASE_P(
...
+            // default values
+            SpecIdToValueBitPatternMap{{100, {0xffffffff}}},
+            // expected
+            "OpDecorate %1 SpecId 100\n"
+            "%1 = OpDecorationGroup\n"
+            "OpGroupDecorate %1 %2\n"
+            "%int = OpTypeInt 32 1\n"
+            "%int_100 = OpConstant %int 100\n",
+        },
+        // 6. Incompatible input bit pattern with the type. Nothing should be
```
dneto0: Nice.

source/opt/set_spec_constant_default_value_pass.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include "set_spec_constant_default_value_pass.h"
 
+#include <algorithm>
```
Qining: Clang-format puts the <algorithm> in the first place. So I will keep it there then.

source/opt/set_spec_constant_default_value_pass.cpp
Qining: Changed back to normal way. Actually all the rest of the related code is in an if-else branch, so the scope is well limited. Originally I want to avoid copy, I just realized rvalue assignment will do the job for me so I can just assign to the bit_pattern vector.

bit_pattern actually cannot be constant, we do std::move() it later.

source/opt/set_spec_constant_default_value_pass.h
Qining: Added in f185207

source/opt/set_spec_constant_default_value_pass.cpp
dneto0: Right.  I think I've been misreading that guideline.   stdio.h would be the C header and cstdio is the corresponding C++ header.  Ooops!
Best not to fight with clang-format!

dneto0:
Rebased, squashed and pushed into master as  144f59e45130

PR #633, opened 2017-04-20
Added Markov chain analysis to stats

atgoo:
Added data structure to SpirvStats which is used to collect statistics
on opcodes following other opcodes.

Added a simple analysis print-out to spirv-stats.

### DIFF
--- a/source/spirv_stats.cpp
+++ b/source/spirv_stats.cpp
@@ -37,8 +37,8 @@ namespace {
 struct StatsContext {
   SpirvStats* stats;
 
-  // Opcodes of already processed instructions. Used and updated in
-  // ProcessOpcode.
+  // Opcodes of already processed instructions in the order as they appear in
+  // the module.
   std::vector<uint32_t> opcodes;
 };
 
@@ -87,8 +87,6 @@ void ProcessOpcode(StatsContext* ctx,
     auto& hist = (*step_it)[*opcode_it];
     ++hist[opcode];
   }
-
-  ctx->opcodes.push_back(opcode);
 }
 
 // Collects opcode usage statistics and calls other collectors.
@@ -99,6 +97,9 @@ spv_result_t ProcessInstruction(
   ProcessOpcode(ctx, inst);
   ProcessCapability(ctx, inst);
   ProcessExtension(ctx, inst);
+
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+  ctx->opcodes.push_back(opcode);
 
   return SPV_SUCCESS;
 }
--- a/source/spirv_stats.cpp
+++ b/source/spirv_stats.cpp
@@ -34,20 +34,27 @@ using libspirv::SpirvStats;
 
 namespace {
 
+struct StatsContext {
+  SpirvStats* stats;
+
+  // Opcodes of already processed instructions. Used and updated in
+  // ProcessOpcode.
+  std::vector<uint32_t> opcodes;
+};
+
 // Collects statistics from SPIR-V header (version, generator).
 spv_result_t ProcessHeader(
     void* user_data, spv_endianness_t /* endian */, uint32_t /* magic */,
     uint32_t version, uint32_t generator, uint32_t /* id_bound */,
     uint32_t /* schema */) {
-  SpirvStats* stats =
-      reinterpret_cast<libspirv::SpirvStats*>(user_data);
-  ++stats->version_hist[version];
-  ++stats->generator_hist[generator];
+  StatsContext* ctx = reinterpret_cast<StatsContext*>(user_data);
+  ++ctx->stats->version_hist[version];
+  ++ctx->stats->generator_hist[generator];
   return SPV_SUCCESS;
 }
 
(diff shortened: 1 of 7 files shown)


dneto0:
Windows bot failure.  Looks like an out-of-order output bug. Are you iterating over an unordered map?

source/spirv_stats.cpp
```diff
@@ -34,49 +34,71 @@ using libspirv::SpirvStats;
 
 namespace {
 
+struct StatsContext {
+  SpirvStats* stats;
+
+  // Opcodes of already processed instructions. Used and updated in
+  // ProcessOpcode.
+  std::vector<uint32_t> opcodes;
```
dneto0: The comment must say how this vector is indexed.  When I first read this I assumed it was indexed by the opcode value.
But reading the rest of the code I think this vector contains the opcodes for instructions, in the order listed in the module. For example, opcodes[k] is the opcode for the instruction at position k (0-based).

Also, to avoid staleness in comments, we normally describe the data itself rather than where it is used.

tools/stats/stats_analyzer.cpp
```diff
@@ -90,6 +90,21 @@ void WriteFreq(std::ostream& out, const std::unordered_map<Key, double>& freq,
   }
 }
 
+template <class Key>
```
dneto0: Need a function comment.

tools/stats/stats_analyzer.cpp
```diff
@@ -125,3 +140,43 @@ void StatsAnalyzer::WriteOpcode(std::ostream& out) {
   out << "Total unique opcodes used: " << opcode_freq_.size() << std::endl;
   WriteFreq(out, opcode_freq_, GetOpcodeString);
 }
+
+void StatsAnalyzer::WriteOpcodeMarkov(std::ostream& out) {
```
dneto0: Need a function comment.

test/stats/stats_aggregate_test.cpp
```diff
@@ -235,4 +235,85 @@ OpMemoryModel Logical GLSL450
...
+%i32 = OpTypeInt 32 1
+%u32 = OpTypeInt 32 0
+%f32 = OpTypeFloat 32
+)";
+
+  SpirvStats stats;
+  stats.opcode_markov_hist.resize(2);
+
+  CompileAndAggregateStats(code1, &stats);
+  ASSERT_EQ(2u, stats.opcode_markov_hist.size());
```
dneto0: No change requested.

But I suspect some of this might have been easier to read and write if you had used brace-initialization and ::testing::Eq and its ability to match containers?   In other tests that tends clean up checks against sizes of containers.

test/stats/stats_analyzer_test.cpp
```diff
@@ -140,4 +140,33 @@ TEST(StatsAnalyzer, Opcode) {
   EXPECT_EQ(expected_output, output);
 }
 
+TEST(StatsAnalyzer, OpcodeMarkov) {
```
dneto0: It's good to see testing of output routines. That's often overlooked!

tools/stats/stats_analyzer.cpp
```diff
@@ -125,3 +140,43 @@ void StatsAnalyzer::WriteOpcode(std::ostream& out) {
...
+  if (stats_.opcode_markov_hist.empty())
+    return;
+
+  const std::unordered_map<uint32_t, std::unordered_map<uint32_t, uint32_t>>&
+      cue_to_hist = stats_.opcode_markov_hist[0];
+
+  for (const auto& kv : cue_to_hist) {
```
dneto0: I think this is causing the test failure on Windows.  Accidental ordering of outputs.

tools/stats/stats_analyzer.cpp
```diff
@@ -125,3 +140,43 @@ void StatsAnalyzer::WriteOpcode(std::ostream& out) {
...
+    const uint32_t cue = kv.first;
+    const double kFrequentEnoughToAnalyze = 0.001;
```
dneto0: Good.  Don't need to stare at noise!

tools/stats/stats_analyzer.cpp
```diff
@@ -125,3 +140,43 @@ void StatsAnalyzer::WriteOpcode(std::ostream& out) {
...
+    for (const auto& pair : sorted_hist) {
+      const uint32_t kStatisticallySignificant = 10;
+      if (pair.second < kStatisticallySignificant) continue;
+
+      const double prior = opcode_freq_[pair.first];
+      const double posterior =
+          static_cast<double>(pair.second) / static_cast<double>(total);
+      out << GetOpcodeString(cue) << " -> " << GetOpcodeString(pair.first)
+          << " " << posterior * 100 << "% (base rate " << prior * 100
+          << "%, pair occurances " << pair.second << ")" << std::endl;
```
dneto0: "occurrences"

dneto0:
Looks good. Waiting for bots to pass.

dneto0:
Rebased and pushed into master as 87a3f651e2

PR #634, opened 2017-04-25
Inline: Do not inline functions with multiple returns (for now)

greg-lunarg:
Currently the exhaustive inliner is generating invalid SPIR-V when inlining functions with multiple returns. 

This fix will quickly stop generation of invalid SPIR-V by inhibiting inlining of any such functions.

I will follow this up shortly with inlining such functions where no return is in a loop. This is likely the majority of the cases.

### DIFF
--- a/source/opt/basic_block.h
+++ b/source/opt/basic_block.h
@@ -52,8 +52,12 @@ class BasicBlock {
 
   iterator begin() { return iterator(&insts_, insts_.begin()); }
   iterator end() { return iterator(&insts_, insts_.end()); }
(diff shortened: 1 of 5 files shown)


dneto0:
Thanks for taking quick action!

Looks good to me.  This will catch all bad cases in practice.
(The test covers the OpReturnValue case but not OpReturn.  Ok for me since the code looks right and we plan to expand the heuristic soon.)

There is one more weird possible weird case, isomorphic to:
```
void main() {
  while(true) { return; }
}
```
It turns out that Glslang generates a second return for this code.  So only a compiler that tries hard to eliminate dead code will generate the bad case.

Will wait for bots to complete before merging.

PR #635, opened 2017-04-26
Inline: Change "--inline-entry-points-all" to "-exhaustive"

greg-lunarg:
Make the name of the exhaustive inliner more clear so that people can avoid this behavior if they wish.


dneto0:
Will merge this in the morning.

dneto0:
Rebased and pushed into master as bad90d9f

Issue #637, opened 2017-05-01
Bit stream utils


atgoo:
Compression project requires back-end which would allow to write/read arbitrary bit messages to/from a buffer.

See https://github.com/KhronosGroup/SPIRV-Tools/issues/382 for more information on SPIR-V sompression.

PR #638, opened 2017-05-01
Added bit stream utils

atgoo:
Utils for working with bit stream buffers and text strings.

This CL is a foundation block for future development of SPIR-V compression. It fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/637.


source/util/bit_stream.h
```diff
@@ -0,0 +1,371 @@
...
+inline std::string BitsetToStream(const std::bitset<N>& bits, size_t num_bits = N) {
+  std::string str = bits.to_string().substr(N - num_bits);
+  std::reverse(str.begin(), str.end());
+  return str;
+}
+
+// Converts a left-to-right stream of bits to uint64.
+inline uint64_t StreamToBits(std::string str) {
+  std::reverse(str.begin(), str.end());
+  return std::bitset<64>(str).to_ullong();
```
dneto0: I'm not sure if this to_ullong method is availble on Android.  Will need to test.

source/util/bit_stream.h
```diff
@@ -0,0 +1,371 @@
...
+
+  // Returns written stream as std::string, padded with zeroes so that the
+  // length is a multiple of 64.
+  std::string GetStreamPadded64() const {
+    return BufferToStream(buffer_);
+  }
+
+ private:
+  std::vector<uint64_t> buffer_;
+  size_t end_;
```
dneto0: end_ member needs a comment.   It's no obvious what it is.  From GetNumBits it seems it's the number of bits written so far.

source/util/bit_stream.cpp
```diff
@@ -0,0 +1,379 @@
...
+    payload_written += chunk_length;
+    val = val >> chunk_length;
+
+    // Write a single bit to signal if there is more to come.
+    writer->WriteBits(val ? 1 : 0, 1);
+  }
+}
+
+// Reads data written with WriteVariableWidthInternal. |chunk_length| and
+// |max_payload| should be identical to those used to write the data.
```
dneto0: Should describe the return value.

source/util/bit_stream.cpp
```diff
@@ -0,0 +1,379 @@
...
+  return pos_ >= buffer_.size() * 64;
+}
+
+bool BitReaderWord64::OnlyZeroesLeft() const {
+  if (ReachedEnd())
+    return true;
+
+  const size_t index = pos_ / 64;
+  if (index < buffer_.size() - 1)
+    return false;
```
dneto0: Could this return false even though there might only be zeroes left?     For example if there are a million zero bits left and nothing else, then we'll return false from here.  But the specification of the function says we should have returned true?  
If so, then the function definition should be clarified about this one-sided error.  Maybe it should be NoOnesLeft instead?

source/util/bit_stream.cpp
```diff
@@ -0,0 +1,379 @@
...
+#include <type_traits>
+
+#include "util/bit_stream.h"
+
+namespace spvutils {
+
+namespace {
+
+// Returns if the system is little-endian. Unfortunately only works during
+// runtime.
```
atgoo: I don't agree with what I see there. It is not a problem to cast
uint64_t* to either uint32_t* or uint16_t* or uint8_t*.

Casting uint8_t* to uint16_t* can be a problem, due to alignment. Casting structs or classes can be a problem due to padding and virtual function tables. But I don't see any problem with casting a pointer to a bigger uint to a pointer to a smaller uint.

source/util/bit_stream.cpp
atgoo: Yes. The description says "if we are nearing the end and only zeroes are left", but I agree it's not very clear.

I'm not entirely sure I would need this function, as simply storing the data length in bits in the header could be simpler than assuring that no instruction can encode to an array of zero bits.

source/util/bit_stream.h
```diff
@@ -0,0 +1,371 @@
...
+  if (block_num & 1) {
+    // Negative.
+    return -1LL - ((block_num >> 1) << block_exponent) - pos;
+  } else {
+    // Positive.
+    return ((block_num >> 1) << block_exponent) + pos;
+  }
+}
+
+// Converts |buffer| to a stream of zeroes and ones.
```
atgoo: '0' and '1'. What I call stream is a string like this: "0001". I use the word stream instead of string, because it describes bits in the order they come. (i.e. "0001" is 16 and not 1).

source/util/bit_stream.h
```diff
@@ -0,0 +1,371 @@
...
+template <typename T>
+std::string BufferToStream(const std::vector<T>& buffer) {
+  std::stringstream ss;
+  for (auto it = buffer.begin(); it != buffer.end(); ++it) {
+    std::string str = std::bitset<sizeof(T) * 8>(*it).to_string();
```
atgoo: gnustl doesn't have std::to_string, but std::bitset::to_string seems to be there. I didn't test.

source/util/bit_stream.h
atgoo: I can see it gnustl. Didn't test.

source/util/bit_stream.cpp
dneto0: Have I misunderstood the C++ spec?  For example, I think casting uint64_t* to uint32_t* and then accessing the underlying object produces undefined behaviour.   (It might work on every compiler in existence, but that's a separate question.)

C++11 3.10 paragraph 10:

```
If a program attempts to access the stored value of an object through a glvalue of other than one of the
following types the behavior is undefined:52
— the dynamic type of the object,
— a cv-qualified version of the dynamic type of the object,
— a type similar (as defined in 4.4) to the dynamic type of the object,
— a type that is the signed or unsigned type corresponding to the dynamic type of the object,
— a type that is the signed or unsigned type corresponding to a cv-qualified version of the dynamic type
of the object,
— an aggregate or union type that includes one of the aforementioned types among its elements or nonstatic
data members (including, recursively, an element or non-static data member of a subaggregate
or contained union),
— a type that is a (possibly cv-qualified) base class type of the dynamic type of the object,
— a char or unsigned char type.
```

test/bit_stream.cpp
```diff
@@ -0,0 +1,1026 @@
...
+TEST(ZigZagCoding, Decode0) {
+  EXPECT_EQ(0, DecodeZigZag(0, 0));
+  EXPECT_EQ(-1, DecodeZigZag(1, 0));
+  EXPECT_EQ(1, DecodeZigZag(2, 0));
+  EXPECT_EQ(-2, DecodeZigZag(3, 0));
+}
+
+TEST(ZigZagCoding, Decode0SameAsNormalZigZag) {
+  for (int32_t i = -10000; i < 10000; i += 123) {
+    ASSERT_EQ(DecodeZigZag(i), DecodeZigZag(i, 0));
```
atgoo: It's intentional because of the loop. Also fixed in a few other places, where I used EXPECT inside a big loop.

dneto0:
Let me know when this is ready for re-review.

dneto0:
Build failure with Clang:
../external/googletest/googletest/include/gtest/gtest.h:1392:11: error: comparison of integers of different signs: 'const int' and 'const unsigned long' [-Werror,-Wsign-compare]
  if (lhs == rhs) {
      ~~~ ^  ~~~

atgoo:
Should be ready.

dneto0:
I've verified with an Android build via google/shaderc.  (I added the source/util/bit_stream.cpp to its third_party/Android.mk)

dneto0:
Rebased, squashed and pushed into master as 3eb716cae

PR #640, opened 2017-05-03
Stats analyzer uses validator

atgoo:
Stats analyzer calls validator to check the instruction and update
validator state.

Fixed unit tests (validator was failing).


dneto0:
Rebased and pushed into master as b4cf371936

PR #647, opened 2017-05-08
Inline: Inline early return function if no returns in loop.

greg-lunarg:
This creates a one-trip loop around the callee code before it is inlined. Returns are converted to branches to the loop's merge block. If a callee has any returns in a loop it is not inlined.

### DIFF
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -507,7 +507,7 @@ void InlinePass::ComputeStructuredSuccessors(ir::Function* func) {
   }
 }
 
-InlinePass::GetBlocksFunction InlinePass::SuccessorsFunction() {
+InlinePass::GetBlocksFunction InlinePass::StructuredSuccessorsFunction() {
   return [this](const ir::BasicBlock* block) {
     return &(block2structuredSuccs_[block]);
   };
@@ -526,7 +526,7 @@ bool InlinePass::hasNoReturnInLoop(ir::Function* func) {
   auto ignore_edge = [](cbb_ptr, cbb_ptr) {};
   std::list<const ir::BasicBlock*> structuredOrder;
   spvtools::CFA<ir::BasicBlock>::DepthFirstTraversal(
-    &*func->begin(), SuccessorsFunction(), ignore_block,
+    &*func->begin(), StructuredSuccessorsFunction(), ignore_block,
     [&](cbb_ptr b) { structuredOrder.push_front(b); }, ignore_edge);
   // Search for returns in loops. Only need to track outermost loop
   bool return_in_loop = false;
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -27,7 +27,6 @@ static const int kSpvTypePointerStorageClass = 1;
 static const int kSpvTypePointerTypeId = 2;
 static const int kSpvLoopMergeMergeBlockId = 0;
 static const int kSpvSelectionMergeMergeBlockId = 0;
-static const int kSpvBranchTargetLabelId = 0;
 
 namespace spvtools {
 namespace opt {
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -530,7 +530,7 @@ bool InlinePass::HasNoReturnInLoop(ir::Function* func) {
     return false;
   // Compute structured block order. This order has the property
   // that dominators are before all blocks they dominate and merge blocks
-  // are after all blocks that are in the control construct their header.
+  // are after all blocks that are in the control constructs of their header.
   ComputeStructuredSuccessors(func);
   auto ignore_block = [](cbb_ptr) {};
   auto ignore_edge = [](cbb_ptr, cbb_ptr) {};
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -388,7 +388,7 @@ void InlinePass::GenInlineCode(
           const auto mapItr = callee2caller.find(*iid);
           if (mapItr != callee2caller.end()) {
             *iid = mapItr->second;
-          } else if (cpi->has_labels()) {
+          } else if (cpi->HasLabels()) {
             const ir::Instruction* inst =
                 def_use_mgr_->id_to_defs().find(*iid)->second;
             if (inst->opcode() == SpvOpLabel) {
@@ -468,7 +468,7 @@ bool InlinePass::Inline(ir::Function* func) {
   return modified;
 }
 
-bool InlinePass::hasMultipleReturns(ir::Function* func) {
+bool InlinePass::HasMultipleReturns(ir::Function* func) {
   bool seenReturn = false;
   bool multipleReturns = false;
   for (auto& blk : *func) {
@@ -513,10 +513,10 @@ InlinePass::GetBlocksFunction InlinePass::StructuredSuccessorsFunction() {
   };
 }
 
-bool InlinePass::hasNoReturnInLoop(ir::Function* func) {
+bool InlinePass::HasNoReturnInLoop(ir::Function* func) {
   // If control not structured, do not do loop/return analysis
   // TODO: Analyze returns in non-structured control flow
-  if (!module_->hasCapability(SpvCapabilityShader))
+  if (!module_->HasCapability(SpvCapabilityShader))
     return false;
   // Compute structured block order. This order has the property
   // that dominators are before all blocks they dominate and merge blocks
@@ -557,13 +557,13 @@ bool InlinePass::hasNoReturnInLoop(ir::Function* func) {
 
 void InlinePass::ReturnAnalysis(ir::Function* func) {
   // Look for multiple returns
-  if (!hasMultipleReturns(func)) {
+  if (!HasMultipleReturns(func)) {
     no_return_in_loop_.insert(func->result_id());
     return;
   }
   early_return_.insert(func->result_id());
   // If multiple returns, see if any are in a loop
-  if (hasNoReturnInLoop(func))
+  if (HasNoReturnInLoop(func))
     no_return_in_loop_.insert(func->result_id());
 }
 
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -476,9 +476,10 @@ bool InlinePass::HasMultipleReturns(ir::Function* func) {
   bool seenReturn = false;
   bool multipleReturns = false;
   for (auto& blk : *func) {
-    auto lii = blk.cend();
-    --lii;
-    if (lii->opcode() == SpvOpReturn || lii->opcode() == SpvOpReturnValue) {
+    auto terminal_ii = blk.cend();
+    --terminal_ii;
+    if (terminal_ii->opcode() == SpvOpReturn || 
+        terminal_ii->opcode() == SpvOpReturnValue) {
       if (seenReturn) {
         multipleReturns = true;
         break;
@@ -490,15 +491,15 @@ bool InlinePass::HasMultipleReturns(ir::Function* func) {
 }
 
 uint32_t InlinePass::MergeBlockIdIfAny(const ir::BasicBlock& blk) {
-  auto mii = blk.cend();
-  --mii;
+  auto merge_ii = blk.cend();
+  --merge_ii;
   uint32_t mbid = 0;
-  if (mii != blk.cbegin()) {
-    --mii;
-    if (mii->opcode() == SpvOpLoopMerge)
-      mbid = mii->GetSingleWordOperand(kSpvLoopMergeMergeBlockId);
-    else if (mii->opcode() == SpvOpSelectionMerge)
-      mbid = mii->GetSingleWordOperand(kSpvSelectionMergeMergeBlockId);
+  if (merge_ii != blk.cbegin()) {
+    --merge_ii;
+    if (merge_ii->opcode() == SpvOpLoopMerge)
+      mbid = merge_ii->GetSingleWordOperand(kSpvLoopMergeMergeBlockId);
+    else if (merge_ii->opcode() == SpvOpSelectionMerge)
+      mbid = merge_ii->GetSingleWordOperand(kSpvSelectionMergeMergeBlockId);
   }
   return mbid;
 }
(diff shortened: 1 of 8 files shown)


dneto0:
Clang errors out due to an unused variable:

../source/opt/inline_pass.cpp:30:18: error: unused variable 'kSpvBranchTargetLabelId' [-Werror,-Wunused-const-variable]
static const int kSpvBranchTargetLabelId = 0;
                 ^

source/opt/module.h
```diff
@@ -86,6 +86,12 @@ class Module {
   std::vector<Instruction*> GetConstants();
   std::vector<const Instruction*> GetConstants() const;
 
+  // Return result id of global value with |opcode|, 0 if not present.
+  uint32_t GetGlobalValue(SpvOp opcode) const;
```
dneto0: At first it was puzzling that this method is not general enough to find *any* kind of global value.
But I see how you only do this to get an OpConstantFalse value.  I guess this is fine.

Similar with AddGlobalValue not being fully general, but we already have the general overload that takes a unique Instruction. So that's fine too.

source/opt/inline_pass.h
```diff
@@ -117,11 +134,31 @@ class InlinePass : public Pass {
...
-  // Returns true if |func| is a function that can be inlined.
-  bool IsInlinableFunction(const ir::Function* func);
+  // Compute structured successors for function |func|.
+  // A block's structured successors are preceded by its
+  // merge block if its a header. This assures correct depth
+  // first search in the presence of early returns and kills.
+  void ComputeStructuredSuccessors(ir::Function* func);
+  
+  // Return function to return successors for a given block
+  GetBlocksFunction SuccessorsFunction();
```
dneto0: Should say that this assumes ComputedStructuredSuccessors has been called.
For otherwise the data is invalid.

source/opt/inline_pass.cpp
```diff
@@ -402,23 +468,117 @@ bool InlinePass::Inline(ir::Function* func) {
...
+      seenReturn = true;
+    }
+  }
+  return multipleReturns;
+}
+
+void InlinePass::ComputeStructuredSuccessors(ir::Function* func) {
+  // If header, make merge block first successor.
+  for (auto& blk : *func) {
+    auto mii = blk.end();
```
dneto0: I find "mii" cryptic.

source/opt/inline_pass.cpp
```diff
@@ -402,23 +468,117 @@ bool InlinePass::Inline(ir::Function* func) {
...
+    mii--;
```
dneto0: A little thing:   For iterators, always prefer pre-decrement (or pre-increment) to post-increment (or post-increment).    That way the update is done in-place and the compiler won't have to make a copy of the iterator.

source/opt/inline_pass.cpp
```diff
@@ -402,23 +468,117 @@ bool InlinePass::Inline(ir::Function* func) {
...
+    uint32_t mbid = 0;
```
dneto0: Is mbid "merge block id"?

source/opt/inline_pass.cpp
```diff
@@ -402,23 +468,117 @@ bool InlinePass::Inline(ir::Function* func) {
...
+    if (mii != blk.begin()) {
```
dneto0: A vaild basic block always has at least two instructions: the label and the terminator instruction.
You're just being defensive here.  That's fine, but you should say so.

The practical impact is that for a valid module we'll never take the false branch.    That might show up one day in a code coverage report, so it will be nice to either (a)  make the invalid test case that proves you need this, or (b) add the comment.

source/opt/inline_pass.cpp
```diff
@@ -402,23 +468,117 @@ bool InlinePass::Inline(ir::Function* func) {
...
+      mii--;
```
dneto0: Prefer --mii.   (Still don't know what mii means.  Merge instruction iterator?)

source/opt/inline_pass.cpp
```diff
@@ -402,23 +468,117 @@ bool InlinePass::Inline(ir::Function* func) {
...
+      if (mii->opcode() == SpvOpLoopMerge)
+        mbid = mii->GetSingleWordOperand(kSpvLoopMergeMergeBlockId);
+      else if (mii->opcode() == SpvOpSelectionMerge)
+        mbid = mii->GetSingleWordOperand(kSpvSelectionMergeMergeBlockId);
+    }
```
dneto0: Optional: This code sequence could have been clearer if it were factored out into a function:  
```
// Returns the id of the merge block declared by a merge instruction in this block, if any.  If none, returns zero.
uint32_t MergeBlockIdIfAny(const ir::BasicBlock&);
```

source/opt/inline_pass.h
```diff
@@ -117,11 +134,31 @@ class InlinePass : public Pass {
...
                      ir::UptrVectorIterator<ir::BasicBlock> call_block_itr);
 
-  // Returns true if |inst| is a function call that can be inlined.
+  // Return true if |inst| is a function call that can be inlined.
   bool IsInlinableFunctionCall(const ir::Instruction* inst);
 
-  // Returns true if |func| is a function that can be inlined.
-  bool IsInlinableFunction(const ir::Function* func);
+  // Compute structured successors for function |func|.
+  // A block's structured successors are preceded by its
```
dneto0: Update: after reading the body of the function I think I understand now.  Except now the comment is weird:  For a given block B, is B's declared merge block a structured successor to B or not?   It think yes, but your sentence here implies otherwise.


Suggest something like this to replace this middle sentence:
```
The structured successors of a block B are the set of the blocks it branches to, together with its declared merge block if it has one.  When order matters, the merge block always appears first.
```

source/opt/inline_pass.h
```diff
@@ -117,11 +134,31 @@ class InlinePass : public Pass {
...
+  // Return true if |func| has multiple returns
+  bool hasMultipleReturns(ir::Function* func);
+
+  // Return true if |func| has no return in a loop. The current analysis
+  // requires structured control flow, so return false if control flow not
+  // structured ie. module is not a shader.
+  bool hasNoReturnInLoop(ir::Function* func);
+
+  // Find all functions with multiple returns and no returns in loops
+  void ReturnAnalysis(ir::Function* func);
```
dneto0: For a void function, prefer a verb rather than a noun:  AnalyzeReturns.

source/opt/inline_pass.h
```diff
@@ -117,11 +134,31 @@ class InlinePass : public Pass {
...
-  // Returns true if |inst| is a function call that can be inlined.
+  // Return true if |inst| is a function call that can be inlined.
   bool IsInlinableFunctionCall(const ir::Instruction* inst);
 
-  // Returns true if |func| is a function that can be inlined.
-  bool IsInlinableFunction(const ir::Function* func);
+  // Compute structured successors for function |func|.
+  // A block's structured successors are preceded by its
+  // merge block if its a header. This assures correct depth
+  // first search in the presence of early returns and kills.
```
dneto0: I'm not sure why OpKill is relevant here.  ?    OpUnreachable would be in the same category, no?

source/opt/inline_pass.h
```diff
@@ -117,11 +134,31 @@ class InlinePass : public Pass {
...
+  void ComputeStructuredSuccessors(ir::Function* func);
+  
+  // Return function to return successors for a given block
```
dneto0: Suggest "successors" -> "ordered structured successors"

source/opt/inline_pass.cpp
```diff
@@ -402,23 +468,117 @@ bool InlinePass::Inline(ir::Function* func) {
...
-  for (auto bi = func->cbegin(); bi != func->cend(); bi++) {
-    auto li = bi->cend();
-    li--;
-    if (li->opcode() == SpvOpReturn || li->opcode() == SpvOpReturnValue) {
-      if (returnCnt > 0)
-        return false;
-      returnCnt++;
+  // Compute structured block order. This order has the property
+  // that dominators are before all blocks they dominate and merge blocks
+  // are after all blocks that are control dependent on their header.
```
dneto0: "merge blocks are after all blocks that are control dependent on their header".
I think this is not quite right.  Isn't the merge block control dependent on the header that names the merge block?

suggest "for a given header block H, its merge block appears after blocks in the structured constructs headed by H, other than H itself".

source/opt/inline_pass.cpp
```diff
@@ -402,23 +468,117 @@ bool InlinePass::Inline(ir::Function* func) {
...
+  // Search for returns in loops. Only need to track outermost loop
+  bool return_in_loop = false;
+  uint32_t outerLoopMergeId = 0;
+  for (auto& blk : structuredOrder) {
+    // Exiting current outer loop
+    if (blk->id() == outerLoopMergeId)
+      outerLoopMergeId = 0;
+    // Return block
+    auto lii = blk->cend();
+    lii--;
```
dneto0: Prefer --lii

Side note: Seems it would be nice to have a .back() method on ir::BasicBlock.  :-)

source/opt/inline_pass.h
```diff
@@ -67,6 +82,8 @@ class InlinePass : public Pass {
   // Return new label.
   std::unique_ptr<ir::Instruction> NewLabel(uint32_t label_id);
 
+  uint32_t GetFalseId();
```
greg-lunarg: Yeah. In general, sorry for missing style issues you have caught previously.

source/opt/inline_pass.h
greg-lunarg: Yes, that's better.

source/opt/inline_pass.h
greg-lunarg: Assuming the mention of early returns here is apt, isn't OpKill an just early return from the entire shader? OpUnreachable is similar in that it has no predecessor, but since it isn't reachable, it was not clear that it causes a problem for DepthFirstSearch. If you think it should be mentioned, I am open to it.

source/opt/inline_pass.h
greg-lunarg: OK.

source/opt/inline_pass.h
greg-lunarg: OK.

source/opt/inline_pass.h
```diff
@@ -117,11 +134,31 @@ class InlinePass : public Pass {
...
+  // A block's structured successors are preceded by its
+  // merge block if its a header. This assures correct depth
+  // first search in the presence of early returns and kills.
+  void ComputeStructuredSuccessors(ir::Function* func);
+  
+  // Return function to return successors for a given block
+  GetBlocksFunction SuccessorsFunction();
+
+  // Return true if |func| has multiple returns
+  bool hasMultipleReturns(ir::Function* func);
```
greg-lunarg: Yep.

source/opt/inline_pass.cpp
```diff
@@ -202,14 +245,17 @@ void InlinePass::GenInlineCode(
   // Clone and map callee code. Copy caller block code to beginning of
   // first block and end of last block.
   bool prevInstWasReturn = false;
+  uint32_t headerId = 0;
```
dneto0: headerId is too generic a name.  Recommend singleTripLoopHeaderId for this, and singleTripLoopContinueId for the next var.

source/opt/module.h
greg-lunarg: Yeah. I thought someone could beef this up later if they wanted.

source/opt/inline_pass.h
```diff
@@ -139,9 +176,21 @@ class InlinePass : public Pass {
...
+  std::set<uint32_t> no_return_in_loop_;
+
+  // Set of ids of inlinable functions
   std::set<uint32_t> inlinable_;
 
+  // Map from block to its structured successor blocks
+  std::unordered_map<const ir::BasicBlock*, std::vector<ir::BasicBlock*>> block2structuredSuccs_;
+
+  // result id for OpConstantFalse
+  uint32_t falseId;
```
greg-lunarg: Yep.

source/opt/inline_pass.h
```diff
@@ -139,9 +176,21 @@ class InlinePass : public Pass {
...
+  // Set of ids of functions with early returns
+  std::set<uint32_t> early_return_;
+
+  // Set of ids of functions with no returns in loop
+  std::set<uint32_t> no_return_in_loop_;
+
+  // Set of ids of inlinable functions
   std::set<uint32_t> inlinable_;
 
+  // Map from block to its structured successor blocks
```
greg-lunarg: Can I just refer back to the comment for ComputeStructuredSuccessors()?

source/opt/module.h
```diff
@@ -99,6 +105,10 @@ class Module {
   inline IteratorRange<inst_iterator> entry_points();
   inline IteratorRange<const_inst_iterator> entry_points() const;
 
+  // Iterators for capability instructions contained in this module
```
greg-lunarg: Doh!

source/opt/module.h
```diff
@@ -132,6 +142,9 @@ class Module {
   // Returns 1 more than the maximum Id value mentioned in the module.
   uint32_t ComputeIdBound() const;
 
+  // Returns true if module has capability |cap|
+  bool hasCapability(uint32_t cap);
```
greg-lunarg: Done.

source/opt/inline_pass.cpp
greg-lunarg: Yes.

source/opt/inline_pass.cpp
greg-lunarg: > A vaild basic block always has at least two instructions: the label and the terminator instruction

That's what I originally thought too. But then reality in the form of a runtime abort stepped in.

I had forgotten that on the opt side, the OpLabel is kept separate from all the other instructions in a block. So when iterating over a block using begin/end, the OpLabel doesn't appear. Hence the extra logic.

source/opt/inline_pass.cpp
greg-lunarg: Yours is better, but as H will be ordered before its merge block as well, would it be clearer to just say?:  

... and merge blocks are after all blocks that are in the control construct of their header.

source/opt/inline_pass.cpp
greg-lunarg: Done.

source/opt/inline_pass.cpp
greg-lunarg: Done.

source/opt/inline_pass.cpp
greg-lunarg: merge_ii?

source/opt/inline_pass.cpp
greg-lunarg: Done. Yep.

source/opt/inline_pass.cpp
greg-lunarg: Done.

source/opt/inline_pass.h
greg-lunarg: Done.

source/opt/inline_pass.h
greg-lunarg: Yes. This algorithm is a bit clunky. I probably could mention that extra merge blocks in the successor list are harmless.

source/opt/inline_pass.h
dneto0: Yes, that's fine.  Debatable where the definition should go.    They're close enough to each other that it doesn't matter to me.

source/opt/inline_pass.cpp
dneto0: merge_ii would be better, and sufficient.  :-)

source/opt/inline_pass.cpp
dneto0: Oh.  I forgot that too.  Sharp corner. :-(

source/opt/inline_pass.cpp
dneto0: I guess that works too, and is less wordy.  Only tweak is that a loop can have two constructs: the body and the continue construct.  So maybe "constructs" or "construct(s)".

source/opt/inline_pass.h
dneto0: Thanks for accepting them in good spirits.  :-)

dneto0:
Rebased, squashed, and pushed into master as  bba812f4

PR #648, opened 2017-05-09
Stats analyzer aggregates OpConstant usage

atgoo:


source/val/instruction.h
```diff
@@ -76,6 +76,15 @@ class Instruction {
     return inst_;
   }
 
+  // Casts the words belonging to the operand under |index| to |T| and returns.
+  template <typename T>
+  T GetOperandAs(size_t index) const {
+    const spv_parsed_operand_t& operand = operands_.at(index);
+    assert(operand.num_words * 4 >= sizeof(T));
+    assert(operand.offset + operand.num_words <= inst_.num_words);
+    return *reinterpret_cast<const T*>(&words_[operand.offset]);
```
dneto0: I won't complain about undefined behaviour.  :-)

test/stats/stats_aggregate_test.cpp
```diff
@@ -339,4 +339,98 @@ OpMemoryModel Physical32 OpenCL
...
+%8 = OpConstant %s32 2
+%9 = OpConstant %s32 -32
+%10 = OpConstant %u64 64
+%11 = OpConstant %s64 -64
+)";
+
+  SpirvStats stats;
+
+  CompileAndAggregateStats(code1, &stats);
+  EXPECT_EQ(2u, stats.f32_constant_hist.size());
```
dneto0: This is fine, as it's not too cumbersome to test individual elements.

An alternate is to use ::testing::UnorderedElementsAre or ::testing::UnorderedElementsAreArrray.
Those might be easier to use / read when you have larger sets.
See https://github.com/google/googletest/blob/master/googlemock/docs/CheatSheet.md

source/spirv_stats.cpp
```diff
@@ -108,6 +109,49 @@ class StatsAggregator {
     }
   }
 
+  // Collects OpConstant statistics.
```
dneto0: Are you also interested in OpSpecConstant ?
Those might not be very prevalent in the ecosystem, but we're hoping....

source/val/instruction.h
atgoo: I could improve it to:
T x;
memcpy(&x, &words_[operand.offset], sizeof(T));
return x;

PR #649, opened 2017-05-11
Added extension SPV_VAL_ignore_type_decl_unique

atgoo:
If extension is present in the module, validator will ignore the "type
declaration must be unique" rule.


dneto0:
What's the motivation for this change?

The only risk with this is the namespace SPV_VAL_.    The schema for extensions is:
*  SPV_KHR_ for Khronos-ratified
* SPV_EXT_ for multi company but not Khronos
* SPV_<company>_ for a specific company

So "VAL" theoretically stomps on the namespace for a mythical company that wants "VAL".   Perhaps use "SPV_SPVTOOLS_"  to be more clear about it.

atgoo:
Type unique validator rule is disabled because it's failing vulkan cts tests. With this extension we can reenable it.

Could be SPV_VALIDATOR_.

source/validate_type_unique.cpp
```diff
@@ -26,9 +26,15 @@ namespace libspirv {
 // Validates that type declarations are unique, unless multiple declarations
 // of the same data type are allowed by the specification.
 // (see section 2.8 Types and Variables)
+// Doesn't do anything if SPV_VAL_ignore_type_decl_unique was declared in the
+// module.
```
dneto0: Ok.  But please add a comment here saying we're adding this only for some limited legacy uses, and this mechanism should not otherwise be used.

dneto0: Please change the name to SPV_SPVTOOLS_ignore.... or SPV_VALIDATOR_igore...

dneto0:
Rebased, squashed, and pushed into master as 20fe94666

PR #651, opened 2017-05-15
Add Local Access Chain Convert pass

greg-lunarg:
This pass is the next in a series of passes designed to reduce the size of SPIR-V files.

This pass targets function scope variables of math type (scalar, vector, matrix) or structs or arrays of math type. It further focuses on variables which are accessed only with loads, stores, and access chains with constant indices. It then converts all loads and stores of such  variables into equivalent sequences of loads, stores, extracts and inserts. This unifies access to these variables to a single mode, simplifying subsequent analysis and elimination of these variables along with loads and stores and allowing values to propagate further for better subsequent analysis and optimization.

I will add tests shortly. Just wanted to get this next pass moving into the review pipeline.

I have not "advertised" this option yet in the spirv-opt usage message. I was thinking it might be better to hold off advertising until we have checked in enough passes so that users can actually see a reduction in SPIR-V file size using these passes.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -201,8 +201,9 @@ Optimizer::PassToken CreateInlinePass();
 // variables into equivalent sequences of loads, stores, extracts and inserts.
 //
 // This pass only processes entry point functions. It currently only converts
-// non-nested, non-ptr access chains. It does process modules with non-32-bit
-// integer types present.
+// non-nested, non-ptr access chains. It does not process modules with
+// non-32-bit integer types present. Optional memory access options on loads
+// and stores are ignored as we are only processing function scope variables.
 //
 // This pass unifies access to these variables to a single mode and simplifies
 // subsequent analysis and elimination of these variables along with their
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -201,7 +201,8 @@ Optimizer::PassToken CreateInlinePass();
 // variables into equivalent sequences of loads, stores, extracts and inserts.
 //
 // This pass only processes entry point functions. It currently only converts
-// non-ptr access chains.
+// non-nested, non-ptr access chains. It does process modules with non-32-bit
+// integer types present.
 //
 // This pass unifies access to these variables to a single mode and simplifies
 // subsequent analysis and elimination of these variables along with their
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -199,9 +199,13 @@ Optimizer::PassToken CreateInlinePass();
 // variables which are accessed only with loads, stores and access chains
 // with constant indices. It then converts all loads and stores of such
 // variables into equivalent sequences of loads, stores, extracts and inserts.
-// This unifies access to these variables to a single mode and simplifies
-// subsequent analysis and elimination of these variables along with loads
-// and stores allowing values to propagate further.
+//
+// This pass only processes entry point functions.
+//
+// This pass unifies access to these variables to a single mode and simplifies
+// subsequent analysis and elimination of these variables along with their
+// loads and stores allowing values to propagate to their points of use where
+// possible.
 Optimizer::PassToken CreateLocalAccessChainConvertPass();
 
 // Creates a compact ids pass.
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -200,7 +200,8 @@ Optimizer::PassToken CreateInlinePass();
 // with constant indices. It then converts all loads and stores of such
 // variables into equivalent sequences of loads, stores, extracts and inserts.
 //
-// This pass only processes entry point functions.
+// This pass only processes entry point functions. It currently only converts
+// non-ptr access chains.
 //
 // This pass unifies access to these variables to a single mode and simplifies
 // subsequent analysis and elimination of these variables along with their
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -194,6 +194,16 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 // points are not changed.
 Optimizer::PassToken CreateInlinePass();
 
+// Creates a local access chain conversion pass.
+// A local access chain conversion pass identifies all function scope
+// variables which are accessed only with loads, stores and access chains
+// with constant indices. It then converts all loads and stores of such
+// variables into equivalent sequences of loads, stores, extracts and inserts.
+// This unifies access to these variables to a single mode and simplifies
+// subsequent analysis and elimination of these variables along with loads
+// and stores allowing values to propagate further.
+Optimizer::PassToken CreateLocalAccessChainConvertPass();
+
 // Creates a compact ids pass.
 // The pass remaps result ids to a compact and gapless range starting from %1.
 Optimizer::PassToken CreateCompactIdsPass();
--- a/source/opt/local_access_chain_convert_pass.cpp
+++ b/source/opt/local_access_chain_convert_pass.cpp
@@ -29,10 +29,6 @@
 */
 
 static const int kSpvEntryPointFunctionId = 1;
-static const int kSpvFunctionCallFunctionId = 0;
-static const int kSpvDecorationTargetId = 0;
-static const int kSpvDecorationDecoration = 1;
-static const int kSpvDecorationLinkageType = 3;
 static const int kSpvStorePtrId = 0;
 static const int kSpvStoreValId = 1;
 static const int kSpvLoadPtrId = 0;
@@ -40,18 +36,6 @@ static const int kSpvAccessChainPtrId = 0;
 static const int kSpvTypePointerStorageClass = 0;
 static const int kSpvTypePointerTypeId = 1;
 static const int kSpvConstantValue = 0;
-static const int kSpvExtractCompositeId = 0;
-static const int kSpvExtractIdx0 = 1;
-static const int kSpvInsertObjectId = 0;
-static const int kSpvInsertCompositeId = 1;
-static const int kSpvBranchCondConditionalId = 0;
-static const int kSpvBranchCondTrueLabId = 1;
-static const int kSpvBranchCondFalseLabId = 2;
-static const int kSpvSelectionMergeMergeBlockId = 0;
-static const int kSpvPhiVal0Id = 0;
-static const int kSpvPhiLab0Id = 1;
-static const int kSpvPhiVal1Id = 2;
-static const int kSpvLoopMergeMergeBlockId = 0;
 
 namespace spvtools {
 namespace opt {
--- a/source/opt/local_access_chain_convert_pass.cpp
+++ b/source/opt/local_access_chain_convert_pass.cpp
@@ -136,7 +136,8 @@ void LocalAccessChainConvertPass::ReplaceAndDeleteLoad(
   }
 }
 
-uint32_t LocalAccessChainConvertPass::GetPteTypeId(const ir::Instruction* ptrInst) {
+uint32_t LocalAccessChainConvertPass::GetPteTypeId(
+    const ir::Instruction* ptrInst) {
   const uint32_t ptrTypeId = ptrInst->type_id();
   const ir::Instruction* ptrTypeInst =
       def_use_mgr_->id_to_defs().find(ptrTypeId)->second;
@@ -148,18 +149,18 @@ void LocalAccessChainConvertPass::BuildAndAppendInst(
     uint32_t typeId,
     uint32_t resultId,
     const std::vector<ir::Operand>& in_opnds,
-    std::vector<std::unique_ptr<ir::Instruction>>& newInsts) {
+    std::vector<std::unique_ptr<ir::Instruction>>* newInsts) {
   std::unique_ptr<ir::Instruction> newInst(new ir::Instruction(
       opcode, typeId, resultId, in_opnds));
   def_use_mgr_->AnalyzeInstDefUse(&*newInst);
-  newInsts.emplace_back(std::move(newInst));
+  newInsts->emplace_back(std::move(newInst));
 }
 
 uint32_t LocalAccessChainConvertPass::BuildAndAppendVarLoad(
     const ir::Instruction* ptrInst,
     uint32_t* varId,
     uint32_t* varPteTypeId,
-    std::vector<std::unique_ptr<ir::Instruction>>& newInsts) {
+    std::vector<std::unique_ptr<ir::Instruction>>* newInsts) {
   const uint32_t ldResultId = TakeNextId();
   *varId = ptrInst->GetSingleWordInOperand(kSpvAccessChainPtrId);
   const ir::Instruction* varInst = def_use_mgr_->GetDef(*varId);
@@ -185,10 +186,9 @@ void LocalAccessChainConvertPass::AppendConstantOperands(
   });
 }
 
-void LocalAccessChainConvertPass::GenAccessChainLoadReplacement(
+uint32_t LocalAccessChainConvertPass::GenAccessChainLoadReplacement(
     const ir::Instruction* ptrInst,
-    std::vector<std::unique_ptr<ir::Instruction>>& newInsts,
-    uint32_t& resultId) {
+    std::vector<std::unique_ptr<ir::Instruction>>* newInsts) {
 
   // Build and append load of variable in ptrInst
   uint32_t varId;
@@ -202,15 +202,15 @@ void LocalAccessChainConvertPass::GenAccessChainLoadReplacement(
   std::vector<ir::Operand> ext_in_opnds = 
       {{spv_operand_type_t::SPV_OPERAND_TYPE_ID, {ldResultId}}};
   AppendConstantOperands(ptrInst, &ext_in_opnds);
-  BuildAndAppendInst(
-      SpvOpCompositeExtract, ptrPteTypeId, extResultId, ext_in_opnds, newInsts);
-  resultId = extResultId;
+  BuildAndAppendInst(SpvOpCompositeExtract, ptrPteTypeId, extResultId, 
+                     ext_in_opnds, newInsts);
+  return extResultId;
 }
 
 void LocalAccessChainConvertPass::GenAccessChainStoreReplacement(
     const ir::Instruction* ptrInst,
     uint32_t valId,
-    std::vector<std::unique_ptr<ir::Instruction>>& newInsts) {
+    std::vector<std::unique_ptr<ir::Instruction>>* newInsts) {
 
   // Build and append load of variable in ptrInst
   uint32_t varId;
@@ -286,8 +286,8 @@ bool LocalAccessChainConvertPass::LocalAccessChainConvert(ir::Function* func) {
         if (!IsTargetVar(varId))
           break;
         std::vector<std::unique_ptr<ir::Instruction>> newInsts;
-        uint32_t replId;
-        GenAccessChainLoadReplacement(ptrInst, newInsts, replId);
+        uint32_t replId =
+            GenAccessChainLoadReplacement(ptrInst, &newInsts);
         ReplaceAndDeleteLoad(&*ii, replId, ptrInst);
         ++ii;
         ii = ii.InsertBefore(&newInsts);
@@ -303,7 +303,7 @@ bool LocalAccessChainConvertPass::LocalAccessChainConvert(ir::Function* func) {
           break;
         std::vector<std::unique_ptr<ir::Instruction>> newInsts;
         uint32_t valId = ii->GetSingleWordInOperand(kSpvStoreValId);
-        GenAccessChainStoreReplacement(ptrInst, valId, newInsts);
+        GenAccessChainStoreReplacement(ptrInst, valId, &newInsts);
         def_use_mgr_->KillInst(&*ii);
         DeleteIfUseless(ptrInst);
         ++ii;
--- a/source/opt/local_access_chain_convert_pass.cpp
+++ b/source/opt/local_access_chain_convert_pass.cpp
@@ -143,6 +143,18 @@ uint32_t LocalAccessChainConvertPass::GetPteTypeId(const ir::Instruction* ptrIns
   return ptrTypeInst->GetSingleWordInOperand(kSpvTypePointerTypeId);
 }
 
+void LocalAccessChainConvertPass::BuildAndAppendInst(
+    SpvOp opcode,
+    uint32_t typeId,
+    uint32_t resultId,
+    const std::vector<ir::Operand>& in_opnds,
+    std::vector<std::unique_ptr<ir::Instruction>>& newInsts) {
+  std::unique_ptr<ir::Instruction> newInst(new ir::Instruction(
+      opcode, typeId, resultId, in_opnds));
+  def_use_mgr_->AnalyzeInstDefUse(&*newInst);
+  newInsts.emplace_back(std::move(newInst));
+}
+
 uint32_t LocalAccessChainConvertPass::BuildAndAppendVarLoad(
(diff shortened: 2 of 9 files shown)


dneto0:
> I have not "advertised" this option yet in the spirv-opt usage message.

That's fine with me. :-)

greg-lunarg:
This code was written before the review of inlining, so it still has some older usages. For example, I noticed I am not using GetSingleWordInOperand in this code. I will fix that.

greg-lunarg:
I can also probably do some consolidation in GenAccessChainLoad/StoreReplacement. Certainly the "Build and append Load" sections can get merged, and probably the Insert/Extract code as well.

greg-lunarg:
Can also use {}s to get rid of some of the Operand vector code.

greg-lunarg:
Looks like header file comments could use a bit of a scrubbing.

greg-lunarg:
I think its ready for another set of eyeballs.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,366 @@
...
+bool LocalAccessChainConvertPass::LocalAccessChainConvert(ir::Function* func) {
+  // Rule out variables accessed with non-constant indices
+  for (auto bi = func->begin(); bi != func->end(); ++bi) {
+    for (auto ii = bi->begin(); ii != bi->end(); ++ii) {
+      switch (ii->opcode()) {
+      case SpvOpStore:
+      case SpvOpLoad: {
+        uint32_t varId;
+        ir::Instruction* ptrInst = GetPtr(&*ii, &varId);
+        if (ptrInst->opcode() != SpvOpAccessChain)
```
dneto0: Treat InBoundsAccessChain the same way?

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,366 @@
...
+}
+
+bool LocalAccessChainConvertPass::IsTargetVar(uint32_t varId) {
+  if (seen_non_target_vars_.find(varId) != seen_non_target_vars_.end())
+    return false;
+  if (seen_target_vars_.find(varId) != seen_target_vars_.end())
+    return true;
+  const ir::Instruction* varInst =
+    def_use_mgr_->id_to_defs().find(varId)->second;
+  assert(varInst->opcode() == SpvOpVariable);
```
dneto0: If I read the caller correctly, I think this code will assert out if you have:

%f = OpFunction...
 %var = OpVariable  ...
  %ptr_a = OpAccessChain %var %zero
  %ptr_b = OpAccessChain %ptr_a %zero
  %value = OpLoad %ptr_b
 
This will assert out because varInst is %ptr_a in this case.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,366 @@
...
+  // Replace access chains of all targeted variables with equivalent
+  // extract and insert sequences
+  bool modified = false;
+  for (auto bi = func->begin(); bi != func->end(); ++bi) {
+    for (auto ii = bi->begin(); ii != bi->end(); ++ii) {
+      switch (ii->opcode()) {
+      case SpvOpLoad: {
+        uint32_t varId;
+        ir::Instruction* ptrInst = GetPtr(&*ii, &varId);
+        if (ptrInst->opcode() != SpvOpAccessChain)
```
dneto0: OpInBoundsAccessChain as well.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,147 @@
...
+  inline uint32_t TakeNextId() {
+    return next_id_++;
+  }
+
+  // Returns true if |typeInst| is a scalar type
+  // or a vector or matrix
+  bool IsMathType(const ir::Instruction* typeInst);
+
+  // Returns true if |typeInst| is a math type or a struct or array
+  // of a math type.
```
dneto0: Ah, I think I see now.    Limiting to single level of nesting reduces complexity everywhere else.  (Sorry but you're seeing my comments out-of-order.)

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,366 @@
...
+          break;
+        std::vector<std::unique_ptr<ir::Instruction>> newInsts;
+        uint32_t valId = ii->GetSingleWordInOperand(kSpvStoreValId);
+        GenAccessChainStoreReplacement(ptrInst, valId, &newInsts);
+        def_use_mgr_->KillInst(&*ii);
+        DeleteIfUseless(ptrInst);
+        ++ii;
+        ii = ii.InsertBefore(&newInsts);
+        ++ii;
+        ++ii;
```
dneto0: Somehow I've lost track of why this gets two ++ii whereas the load only gets one ++ii.  But I'm sure testing proves this is correct. :-)

test/opt/local_access_chain_convert_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::LocalAccessChainConvertPass>(
+      assembly, assembly, false, true);
+}
+
+// TODO(greg-lunarg): Add tests to verify handling of these cases:
+//
```
dneto0: Add TODO test case:
 Where an access chain result is used in two different operations, e.g.
   %ptr = OpAccessChain %ptr_type %var %zero
   %value = OpLoad %value_type %ptr  ; first use
   OpStore %ptr %second_value   ; second use

The replacing code will know to avoid deleting the %ptr instruction when processing the OpLoad, but it will know to delete it the second time.

greg-lunarg:
Thanks David. I will apply these comments to my other PR as well so you don't have to repeat yourself. I will comment on that PR when I am finished.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,397 @@
+// Copyright (c) 2016 The Khronos Group Inc.
```
greg-lunarg: Yep.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,397 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "local_access_chain_convert_pass.h"
+#include "iterator.h"
+
+/*
```
greg-lunarg: Doh!

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,366 @@
...
+}
+
+ir::Instruction* LocalAccessChainConvertPass::GetPtr(
+    ir::Instruction* ip,
+    uint32_t* varId) {
+  const uint32_t ptrId = ip->opcode() == SpvOpStore ?
+      ip->GetSingleWordInOperand(kSpvStorePtrId) :
+      ip->GetSingleWordInOperand(kSpvLoadPtrId);
+  ir::Instruction* ptrInst =
+    def_use_mgr_->id_to_defs().find(ptrId)->second;
```
greg-lunarg: Yes, better.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,366 @@
...
+  *varId = ptrInst->opcode() == SpvOpAccessChain ?
```
greg-lunarg: Changed to also convert OpInBoundsAccessChain. Noted non-support of pointer access chains.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,366 @@
...
+  return ptrInst;
+}
+
+bool LocalAccessChainConvertPass::IsTargetVar(uint32_t varId) {
+  if (seen_non_target_vars_.find(varId) != seen_non_target_vars_.end())
+    return false;
+  if (seen_target_vars_.find(varId) != seen_target_vars_.end())
+    return true;
+  const ir::Instruction* varInst =
+    def_use_mgr_->id_to_defs().find(varId)->second;
```
greg-lunarg: Yes.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,397 @@
...
+    return true;
+  if (typeInst->opcode() != SpvOpTypeStruct &&
+      typeInst->opcode() != SpvOpTypeArray)
+    return false;
+  int nonMathComp = 0;
+  typeInst->ForEachInId([&nonMathComp,this](const uint32_t* tid) {
+    ir::Instruction* compTypeInst =
+        def_use_mgr_->id_to_defs().find(*tid)->second;
+    // Ignore length operand in Array type
+    if (compTypeInst->opcode() == SpvOpConstant) return;
```
greg-lunarg: Done.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,397 @@
...
+    if (!IsMathType(compTypeInst)) ++nonMathComp;
+  });
+  return nonMathComp == 0;
+}
+
+ir::Instruction* LocalAccessChainConvertPass::GetPtr(
+    ir::Instruction* ip,
+    uint32_t* varId) {
+  const uint32_t ptrId = ip->opcode() == SpvOpStore ?
```
greg-lunarg: Done.

source/opt/local_access_chain_convert_pass.cpp
greg-lunarg: Good catch.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,366 @@
...
+void LocalAccessChainConvertPass::ReplaceAndDeleteLoad(
+    ir::Instruction* loadInst,
+    uint32_t replId,
+    ir::Instruction* ptrInst) {
+  const uint32_t loadId = loadInst->result_id();
+  (void) def_use_mgr_->ReplaceAllUsesWith(loadId, replId);
+  // remove load instruction
+  def_use_mgr_->KillInst(loadInst);
+  // if access chain, see if it can be removed as well
+  if (ptrInst->opcode() == SpvOpAccessChain) {
```
greg-lunarg: Done.

source/opt/local_access_chain_convert_pass.cpp
greg-lunarg: Done.

source/opt/local_access_chain_convert_pass.cpp
greg-lunarg: Yep.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) 2016 The Khronos Group Inc.
```
greg-lunarg: Done.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,366 @@
...
+}
+
+void LocalAccessChainConvertPass::AppendConstantOperands(
+    const ir::Instruction* ptrInst,
+    std::vector<ir::Operand>* in_opnds) {
+  uint32_t iidIdx = 0;
+  ptrInst->ForEachInId([&iidIdx, &in_opnds, this](const uint32_t *iid) {
+    if (iidIdx > 0) {
+      const ir::Instruction* cInst = def_use_mgr_->GetDef(*iid);
+      uint32_t val = cInst->GetSingleWordInOperand(kSpvConstantValue);
```
greg-lunarg: Done.

source/opt/local_access_chain_convert_pass.cpp
greg-lunarg: Yes :) Loads are replaced with two instructions: a Load and an Extract. Stores are replaced with three instructions: a Load, an Insert and a Store. Hence the extra ++ii.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,128 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_OPT_SSAMEM_PASS_H_
+#define LIBSPIRV_OPT_SSAMEM_PASS_H_
+
+
+#include <unordered_map>
```
greg-lunarg: Done.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,128 @@
...
+#include <unordered_set>
+#include <map>
+#include <algorithm>
+#include <utility>
+#include <queue>
+
+#include "def_use_manager.h"
```
greg-lunarg: Done.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,128 @@
...
+#include "pass.h"
+
+namespace spvtools {
+namespace opt {
+
+// See optimizer.hpp for documentation.
+class LocalAccessChainConvertPass : public Pass {
+ public:
+  LocalAccessChainConvertPass();
+  const char* name() const override { return "sroa"; }
```
greg-lunarg: Fixed.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,147 @@
...
+  Status Process(ir::Module*) override;
+
+ private:
+   // Module this pass is processing
```
greg-lunarg: Fixed.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,128 @@
...
+   ir::Module* module_;
+
+   // Def-Uses for the module we are processing
+   std::unique_ptr<analysis::DefUseManager> def_use_mgr_;
+
+  // Map from function's result id to function
+  std::unordered_map<uint32_t, ir::Function*> id2function_;
+
+  // Set of verified target types
```
greg-lunarg: Good questions.

As you likely have come to realize, the target type is documented at IsTargetType.

I have now moved the data member definitions below the function members so any readers will come across the "definition" of the term target type before any "use" of it. Hopefully that will reduce such questions.

"Verified" here means that they are variables that IsTargetVar() and other member functions have made a determination on and have been cached for future queries. I have changed the comments to make this a little clearer.

source/opt/local_access_chain_convert_pass.h
greg-lunarg: Yes. I have added a TODO to handle more complex types to clarify this.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,147 @@
...
+  bool IsTargetType(const ir::Instruction* typeInst);
+
+  // Given a load or store |ip|, return the pointer instruction.
+  // Also return the variable's id in |varId|.
+  ir::Instruction* GetPtr(ir::Instruction* ip, uint32_t* varId);
+
+  // Return true if |varId| is function scope variable of targeted type.
+  bool IsTargetVar(uint32_t varId);
+
+  // Delete |inst| if it has no uses. Assumes |inst| has a non-zero resultId.
```
greg-lunarg: Good point though about OpNames. I am currently using spirv-remap to get rid of Debug Ops and do a final DCE. At some point my goal is to add such a pass to spirv-opt.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,147 @@
...
+
+  // Returns true if |typeInst| is a math type or a struct or array
+  // of a math type.
+  bool IsTargetType(const ir::Instruction* typeInst);
+
+  // Given a load or store |ip|, return the pointer instruction.
+  // Also return the variable's id in |varId|.
+  ir::Instruction* GetPtr(ir::Instruction* ip, uint32_t* varId);
+
+  // Return true if |varId| is function scope variable of targeted type.
```
greg-lunarg: Done.

greg-lunarg: I have just added code that encapsulates the first phase of this pass into the function FindTargetVars(). It uses IsTargetVar() but also updates the caches when it finds a target variable which is accessed using a non-constant access chain, which makes it a non-target
variable. I have updated the comments for both member functions to clarify this.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,147 @@
...
+
+  // Returns true if |typeInst| is a scalar type
+  // or a vector or matrix
+  bool IsMathType(const ir::Instruction* typeInst);
+
+  // Returns true if |typeInst| is a math type or a struct or array
+  // of a math type.
+  bool IsTargetType(const ir::Instruction* typeInst);
+
+  // Given a load or store |ip|, return the pointer instruction.
```
greg-lunarg: I have made an attempt to clarify this routine. Yes, it's really only useful for a load or store on a variable or an access chain with a variable base. And yes, if the scope of this pass is expanded, this function's capability would likely expand.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,147 @@
...
+    const std::vector<ir::Operand>& in_opnds,
+    std::vector<std::unique_ptr<ir::Instruction>>* newInsts);
+
+  // Build load of variable in |ptrInst| and append to |newInsts|.
+  // Return var in |varId| and its pointee type in |varPteTypeId|.
+  uint32_t BuildAndAppendVarLoad(const ir::Instruction* ptrInst,
+    uint32_t* varId, uint32_t* varPteTypeId,
+    std::vector<std::unique_ptr<ir::Instruction>>* newInsts);
+
+  // Append constant operands from access chain |ptrInst| to
```
greg-lunarg: OK.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,147 @@
...
+  void GenAccessChainStoreReplacement(const ir::Instruction* ptrInst,
+      uint32_t valId,
+      std::vector<std::unique_ptr<ir::Instruction>>* newInsts);
+
+  // For the (constant index) access chain |ptrInst|, create an
+  // equivalent load and extract. Append to |newInsts|.
+  uint32_t GenAccessChainLoadReplacement(const ir::Instruction* ptrInst,
+      std::vector<std::unique_ptr<ir::Instruction>>* newInsts);
+
+  // Return true if all indices of access chain |acp| are constant
```
greg-lunarg: OK. And changed to const.

include/spirv-tools/optimizer.hpp
```diff
@@ -194,6 +194,20 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 // points are not changed.
 Optimizer::PassToken CreateInlinePass();
 
+// Creates a local access chain conversion pass.
+// A local access chain conversion pass identifies all function scope
+// variables which are accessed only with loads, stores and access chains
+// with constant indices. It then converts all loads and stores of such
```
greg-lunarg: Done.

test/opt/local_access_chain_convert_test.cpp
greg-lunarg: Just added the test.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,369 @@
...
+      ir::Instruction* opInst = def_use_mgr_->GetDef(*tid);
+      if (opInst->opcode() != SpvOpConstant) ++nonConstCnt;
+    }
+    ++inIdx;
+  });
+  return nonConstCnt == 0;
+}
+
+void LocalAccessChainConvertPass::FindTargetVars(ir::Function* func) {
+  for (auto bi = func->begin(); bi != func->end(); ++bi) {
```
dneto0: Optional,Minor:  This is an opportunity for range for.

```
for (auto& block : *func) {
  for (auto& inst : *block) {
    Use inst directly.
  }
}
```

source/opt/local_access_chain_convert_pass.h
dneto0: Yeah.  There's already a strip-debug pass in SPIRV-Tools., which is simpler.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,167 @@
...
+  void DeleteIfUseless(ir::Instruction* inst);
+
+  // Replace all instances of |loadInst|'s id with |replId| and delete
+  // |loadInst| and its pointer |ptrInst| if it is a useless access chain.
+  void ReplaceAndDeleteLoad(ir::Instruction* loadInst,
+    uint32_t replId,
+    ir::Instruction* ptrInst);
+
+  // Return type id for |ptrInst|'s pointee
+  uint32_t GetPteTypeId(const ir::Instruction* ptrInst) const;
```
dneto0: Please rename to GetPointeeTypeId.  It's too easy to misread Pte as Ptr

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+  // Build and append Store
+  BuildAndAppendInst(SpvOpStore, 0, 0, 
+      {{spv_operand_type_t::SPV_OPERAND_TYPE_ID, {varId}}, 
+       {spv_operand_type_t::SPV_OPERAND_TYPE_ID, {insResultId}}},
+      newInsts);
+}
+
+bool LocalAccessChainConvertPass::IsConstantIndexAccessChain(
+    ir::Instruction* acp) const {
```
dneto0: Please make acp const.

source/opt/local_access_chain_convert_pass.h
```diff
@@ -0,0 +1,167 @@
...
+  // Identify all function scope variables of target type which are 
+  // accessed only with loads, stores and access chains with constant
+  // indices. Convert all loads and stores of such variables into equivalent
+  // loads, stores, extracts and inserts. This unifies access to these
+  // variables to a single mode and simplifies analysis and optimization.
+  // See IsTargetType() for targeted types.
+  //
+  // Nested access chains and pointer access chains are not currently
+  // converted.
+  bool LocalAccessChainConvert(ir::Function* func);
```
dneto0: Please rename to ConvertLocalAccessChains

test/opt/local_access_chain_convert_test.cpp
```diff
@@ -0,0 +1,342 @@
...
+//    Assorted vector and matrix types
+//    Assorted struct array types
+//    Assorted scalar types
+//    Assorted non-target types
+//    OpInBoundsAccessChain
```
dneto0: Please add one InBoundsAccessChain test.  (Don't duplicate the entire cross product of test scenarios! )

source/opt/local_access_chain_convert_pass.h
greg-lunarg: Sorry. Don't know how I missed these.

source/opt/local_access_chain_convert_pass.h
greg-lunarg: Done

source/opt/local_access_chain_convert_pass.cpp
greg-lunarg: Done.

test/opt/local_access_chain_convert_test.cpp
greg-lunarg: Done.

dneto0:
Rebased, squashed, and pushed into master as aa7e687ef

Thanks!

PR #655, opened 2017-05-18
Added compression tool tools/spirv-markv

atgoo:
Command line application is located at tools/spirv-markv
API at include/spirv-tools/markv.h

At the moment only very basic compression is implemented, mostly varint. Scope of supported SPIR-V opcodes is also limited.


source/util/bit_stream.h
```diff
@@ -362,8 +404,14 @@ class BitReaderWord64 : public BitReaderInterface {
   // Consuming the original buffer and casting it to uint64 is difficult,
   // as it would potentially cause data misalignment and poor performance.
```
dneto0: More than just poor performance:  It would be invalid on some architectures. (SPARC would trigger a bus error)

source/comp/markv_codec.cpp
```diff
@@ -0,0 +1,1194 @@
...
+  char* cstr = new char[str.length() + 1];
+  assert(cstr);
+  std::strncpy(cstr, str.c_str(), str.length());
+  cstr[str.length()] = '\0';
+  out->str = cstr;
+  out->length = str.length();
+  return out;
+}
+
+// Base class for MARK-V encoder and decoder.
```
dneto0: Since this is your contact point with the valdiation state, should say how you'll be using it.  It seems you're using it to maintain certain bookkeeping (and give examples: id bound, unresolved forward ids, vector of parsed instructions...), and to help process diagnostics.

tools/comp/markv.cpp
```diff
@@ -0,0 +1,229 @@
...
+            fprintf(stderr, "error: More than one input file specified\n");
+            return 1;
+          }
+        } break;
+        default:
+          print_usage(argv[0]);
+          return 1;
+      }
+    } else if (0 == strcmp("e", argv[argi])) {
+      task = kEncode;
```
dneto0: What if the filename is "e"  ?

Then you'll never pick it up.

E.g. spirv-markv e e

tools/comp/markv.cpp
```diff
@@ -0,0 +1,229 @@
...
+    want_comments = false;
+  }
+
+  const bool write_to_stdout = output_filename == nullptr ||
+      0 == strcmp(output_filename, "-");
+
+  spv_text comments = nullptr;
+  spv_text* comments_ptr = want_comments ? &comments : nullptr;
+
+  ScopedContext ctx(SPV_ENV_UNIVERSAL_1_1);
```
dneto0: Make this UNIVERSAL_1_2 since that's the latest now.

source/comp/markv_codec.cpp
```diff
@@ -0,0 +1,1194 @@
...
+      const size_t length = spv_strnlen_s(src, operand.num_words * 4);
+      if (length == operand.num_words * 4)
+        return vstate_.diag(SPV_ERROR_INVALID_BINARY)
+            << "Failed to find terminal character of literal string";
+      for (size_t i = 0; i < length + 1; ++i)
+        writer_.WriteUnencoded(src[i]);
+    } else {
+      for (int i = 0; i < operand.num_words; ++i) {
+        const uint32_t word = instruction.word(operand.offset + i);
+        EncodeOperandWord(operand.type, word);
```
atgoo: It's not so much a bug, but a question if I want to run
StripDebugInfoPass
FlattenDecorationPass
CompactIdsPass
on all binaries coming to the encoder.

source/comp/markv_codec.cpp
```diff
@@ -0,0 +1,1194 @@
...
+
+        if (opcode == SpvOpExtInst && parsed_operand.offset == 3) {
+          assert(0);
+        }
+      }
+      break;
+    }
+
+    case SPV_OPERAND_TYPE_EXTENSION_INSTRUCTION_NUMBER: {
+      assert(0);
```
atgoo: I prefer assert(0 && "not implemented"); instead of a TODO.

source/comp/markv_codec.cpp
```diff
@@ -0,0 +1,1194 @@
...
+        if (auto error = DecodeLiteralNumber(parsed_operand))
+          return error;
+      }
+      break;
+
+    case SPV_OPERAND_TYPE_LITERAL_STRING:
+    case SPV_OPERAND_TYPE_OPTIONAL_LITERAL_STRING: {
+      parsed_operand.type = SPV_OPERAND_TYPE_LITERAL_STRING;
+      std::vector<char> str;
+      while (true) {
```
atgoo: I'm not sure how is it vulnerable. We give a buffer to a function and ask it to read until the end or until '\0'. This is not a kerneI function, so the user of the API can't really do anything more harmful by calling the decoder with a bogus buffer, than they can do without it.

source/comp/markv_codec.cpp
```diff
@@ -0,0 +1,1194 @@
...
+};
+
+const MarkvModel* GetDefaultModel() {
+  static MarkvModel model;
+  return &model;
+}
+
+// Returns chunk length used for variable length encoding of spirv operand
+// words. Returns zero if operand type corresponds to potentially multiple
+// words or a word which is not expected to profit from variable width encoding.
```
dneto0: > Zero means do not encode, so it's safe.

That should be in the comment.   Saying "... not expected to profit from variable width encoding" does not say quite enough.

source/comp/markv_codec.cpp
dneto0: I looked further. The reader_ is your own creation rather than a reader passed in.  So you are relying on the boundedness of the memory buffer, correctness of the reader code.  If the input buffer's string is not null-terminated, then I presume the ReadUnencoded method will return false and you'll get the invalid binary.

Probably worth a comment, since infinite loops are so easily dangerous.

test/comp/markv_codec_test.cpp
```diff
@@ -0,0 +1,346 @@
...
+          SPV_TEXT_TO_BINARY_OPTION_PRESERVE_NUMERIC_IDS);
+  ASSERT_FALSE(binary_to_encode.empty());
+
+  spv_markv_binary markv_binary = nullptr;
+  std::string encoder_comments;
+  Encode(binary_to_encode, &markv_binary, &encoder_comments);
+  ASSERT_NE(nullptr, markv_binary);
+
+  // std::cerr << encoder_comments << std::endl;
+  std::cerr << "SPIR-V size: " << expected_binary.size() * 4 << std::endl;
```
dneto0: Test code should not normally have outputs.

test/comp/markv_codec_test.cpp
```diff
@@ -0,0 +1,346 @@
...
+%5 = OpTypeArray %2 %4
+%6 = OpTypeVoid
+%7 = OpTypeFunction %6
+%1 = OpFunction %6 None %7
+%8 = OpLabel
+OpReturn
+OpFunctionEnd
+)");
+}
+
```
dneto0: I'd like to see tests for:

- OpSwitch on a 64 bit value.  (proves that type tracking works somewhat)
- A test with instructions with a variable number of arguments (other than a function type), e.g. OpAccessChain
- A bit mask operand, e.g. OpLoopMerge with a combination of DontUnroll and DependencyLength 10.

source/comp/markv_codec.cpp
```diff
@@ -0,0 +1,1428 @@
...
+  }
+
+  spv_validator_options validator_options_;
+  ValidationState_t vstate_;
+  const libspirv::AssemblyGrammar grammar_;
+  MarkvHeader header_;
+  const MarkvModel* model_;
+
+  // Move-to-front list of all ids.
+  // TODO(atgoo@github.com) Consider a better move-to-front implementation.
```
atgoo: I've been designing a solution which would use timestamps and a rank/select tree. This way index-to-id becomes select element with k-th timestamp. Id-to-index would also need a hash map id-to-node, then rank(node) would be index.

source/comp/markv_codec.cpp
```diff
@@ -0,0 +1,1428 @@
...
+    if (chunk_length) {
+      return reader_.ReadVariableWidthU32(word, chunk_length);
+    } else {
+      return reader_.ReadUnencoded(word);
+    }
+  }
+
+  // Fetches the id from the move-to-front list and moves it to front.
+  uint32_t GetIdAndMoveToFront(uint16_t index) {
+    if (index >= move_to_front_ids_.size()) {
```
atgoo: list::size is O(1).

source/CMakeLists.txt
```diff
@@ -185,10 +185,12 @@ add_custom_target(spirv-tools-build-version
    DEPENDS ${SPIRV_TOOLS_BUILD_VERSION_INC})
 set_property(TARGET spirv-tools-build-version PROPERTY FOLDER "SPIRV-Tools build")
 
+add_subdirectory(comp)
 add_subdirectory(opt)
 
 set(SPIRV_SOURCES
   ${spirv-tools_SOURCE_DIR}/include/spirv-tools/libspirv.h
+  ${spirv-tools_SOURCE_DIR}/include/spirv-tools/markv.h
```
dneto0: This is not addressed.  The project builds just fine without this.

dneto0:
Do you think all feedback addressed?  Ready for review?

atgoo:
> Do you think all feedback addressed? Ready for review?

Yes

source/comp/markv_codec.cpp
```diff
@@ -0,0 +1,1478 @@
...
+    return ValidateInstructionAndUpdateValidationState(&vstate_, &inst);
+  }
+
+  // Returns the current instruction (the one last processed by the validator).
+  const Instruction& GetCurrentInstruction() const {
+    return vstate_.ordered_instructions().back();
+  }
+
+  // Returns true if the opcode has a variable number of operands.
+  bool OpcodeHasVariableNumberOfOperands(SpvOp opcode) const {
```
dneto0: I think this does not handle things like OpDecorate which can have differing numbers of operands based on one of its enum values.
E.g.   OpDecorate %target BuiltIn Position
         OpDecorate %target Volatile

Discussed offline, and perhaps it's safer to use a whitelist of instructions that certainly have a fixed number of logical operands.  And for quality-of-compression purposes, you only need to cover the top 20 or whatever until you get the long tail.

atgoo:
> Just one more thing required, which is to make the determination of variable-number-of-operands safer.

Done.

include/spirv-tools/markv.h
```diff
@@ -0,0 +1,87 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// MARK-V is a compression format for SPIR-V binaries. It strips away
+// non-essential information (such as result ids which can be regenerated) and
+// uses various bit reduction techiniques to reduce the size of the binary.
+
```
dneto0: Should say there is no forward compatibility promise, at least for now.

For example, the whitelist of instructions that are known to have a fixed number of operands will likely change, and that changes the encoding/decoding format of a given module.

tools/comp/markv.cpp
```diff
@@ -0,0 +1,243 @@
...
+  ~ScopedContext() { spvContextDestroy(context); }
+  spv_context context;
+};
+
+void print_usage(char* argv0) {
+  printf(
+      R"(%s - Encodes or decodes a SPIR-V binary to or from a MARK-V binary.
+
+USAGE: %s [e|d] [options] [<filename>]
+
```
dneto0: Need a statement about lack of promise of forward compatibility.

dneto0:
Also need a rebase, to eliminate the conflict.

dneto0:
Squashed and pushed into master as 73e8dac5b92

Issue #608, opened 2017-03-31
Add a way to retain a DefUseManager between passes, if a pass promises to keep it up to date during operation.


dneto0:
I'm discussing some ideas with @antiagainst and @greg-lunarg on how to do this well.

dneto0:
LLVM does a lot of work to make def-use info very very compact and fast to access.  To first order it's only right to retain it between passes when a pass specifically maintains it properly.  So the question is how to do that.

Considerations and design ideas:
- Who holds onto the data?  ir::Module seems a natural choice but it feels like that would overload its responsibilities.  Better to get the functionality by composing from other objects.
- So I'd rather have a new object, an "optimization context" if you will.  I'd add Pass::ProcessInContext method that it takes a pointer to the optimization context rather than the module.  
    - the context object contains the module, the def use manager, and anything else that is carried from one pass to the next.
- We can then make passes templated over its macro behaviours such as "I maintain the def use manager", or "i properly maintain the Module ID bound"
- The default (existing passes) would have the default instantiation which is "I don't maintain the def use manager", and "i don't maintain the module ID bound".
- Follow LLVM's pattern:  Instead of creating a DefUseManager on its own, a  pass requests it from the context object.  The context has enough info to know whether it has to create one from scratch or whether it can reuse the one it already has.  
- I don't think we need it yet, but if necessary we can add an explicit "InvalidateDefUseManager" method on the context object in case a pass completely reconstructs the def use manager during its operation.
- Fortunately the public API exposes PassToken instead of the raw Pass itself. So we are free to update the Pass API.  :-)

dneto0:
More thoughts:

* Clearly the optimization context is good for any kind of stateful analysis, such as (for example) decoration analysis
* For safety sake, invert the invalidation API so that by default the optimizer context assumes that pass invalidates all analyses.  But a pass that knows it maintains a particular analysis can use a call back to indicate that it has kept it up to date.

dnovillo:
I'm starting to look at adding some new options to spirvopt.  Analysis invalidation will likely affect my work.  I can take a stab at fixing this first.

dneto0:
cc @s-perron

dneto0:
More thoughts:

- I think we should change be more bold and remove the existing Pass::Process(ir::Module) method to Pass::(ir::IRContext) (or whatever the new class name is)
- We now have a real DecorationManager class.
- When we add a types-and-constants pool (or pools), it should be owned by the IRContext
[...]

PR #656, opened 2017-05-19
Add pass to eliminate local loads and stores in basic block.

greg-lunarg:
This pass is another in a series of passes designed to reduce SPIR-V file sizes using classic code optimization techniques.

For every entry point function, remove unneeded loads and stores of
function scope variables. For each targeted variable load, if previous store
to that variable in the block, replace the load's result id with the value id of the
store. If previous load within the block, replace the current load's result id
with the previous load's result id. In either case, delete the current
load. Finally, check if any remaining stores are useless, and delete store
and variable if possible.

The presense of access chain references and function calls can inhibit
the above optimization.

This pass is most effective if preceeded by Inlining and 
LocalAccessChainConvert. This pass will reduce the work needed to be done
by LocalSingleStoreElim and LocalSSARewrite.

You will notice that many members of this latest pass are redundant with those in LocalAccessChainConvertPass. Ultimately these will be collected in an Optimization Context along with state that will persist across passes. See issue #608.


greg-lunarg:
OK. I think I have incorporated all comments from the access chain convert pass into this PR.

Again, you will note that this pass contains a number of member functions that are duplicates from the access chain conversion pass. These should all be folded into the proposed optimization context class.

This pass will handle nested access chains the same as non-nested. The behavior of GetPtr (and its comment) have been updated to deal with this.

This pass will not process modules that have non-logical addressing.

dneto0:
Please rebase this against master.

include/spirv-tools/optimizer.hpp
```diff
@@ -194,6 +194,26 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
...
+// For every entry point function, do single block memory optimization of 
+// function variables referenced only with non-access-chain loads and stores.
+// For each targeted variable load, if previous store to that variable in the
+// block, replace the load's result id with the value id of the store.
+// If previous load within the block, replace the current load's result id
+// with the previous load's result id. In either case, delete the current
+// load. Finally, check if any remaining stores are useless, and delete store
+// and variable if possible.
+//
+// The presense of access chain references and function calls can inhibit
```
dneto0: "presence"

include/spirv-tools/optimizer.hpp
```diff
@@ -194,6 +194,26 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
...
+// the above optimization.
+//
+// Only modules with logical addressing are currently processed.
```
dneto0: Good call, and good comment!

source/opt/local_single_block_elim_pass.cpp
```diff
@@ -0,0 +1,329 @@
...
+static const int kSpvTypePointerTypeId = 1;
+
+namespace spvtools {
+namespace opt {
+
+bool LocalSingleBlockElimPass::IsNonPtrAccessChain(const SpvOp opcode) const {
+  return opcode == SpvOpAccessChain || opcode == SpvOpInBoundsAccessChain;
+}
+
+bool LocalSingleBlockElimPass::IsMathType(
```
dneto0: About the name of the class.  The pass eliminates loads and stores, but the class name makes it sound like it's eliminating Blocks.    It might be unwieldy, but 
LocalSingleBlockLoadStoreElimPass would be more accurate.
(Sorry. )

source/opt/local_single_block_elim_pass.h
```diff
@@ -0,0 +1,140 @@
...
+
+  // Return true if |storeInst| is not to function variable or if its
+  // base variable has a load
+  bool IsLiveStore(ir::Instruction* storeInst);
+
+  // Add stores using |ptr_id| to |insts|
+  void AddStores(uint32_t ptr_id, std::queue<ir::Instruction*>* insts);
+
+  // Delete |inst| and iterate DCE on all its operands 
+  void DCEInst(ir::Instruction* inst);
```
dneto0: I understand that this will make for nicer looking test and debug.
But we should consider just making this a separate general purpose pass.   Suggest adding a TODO to pull this out.

source/opt/local_single_block_elim_pass.h
```diff
@@ -0,0 +1,140 @@
...
+  // If variable is not cached, return true if variable is a function scope 
+  // variable of target type, false otherwise. Updates caches of target 
+  // and non-target variables.
+  bool IsTargetVar(uint32_t varId);
+
+  // Replace all instances of |loadInst|'s id with |replId| and delete
+  // |loadInst|.
+  void ReplaceAndDeleteLoad(ir::Instruction* loadInst, uint32_t replId);
+
+  // Return true if any instruction loads from |varId|
```
dneto0: Should qualify "any instruction in this function" loads from |varId|

source/opt/local_single_block_elim_pass.cpp
```diff
@@ -0,0 +1,329 @@
...
+  (void) def_use_mgr_->ReplaceAllUsesWith(loadId, replId);
+  DCEInst(loadInst);
+}
+
+bool LocalSingleBlockElimPass::HasLoads(uint32_t varId) const {
+  analysis::UseList* uses = def_use_mgr_->GetUses(varId);
+  if (uses == nullptr)
+    return false;
+  for (auto u : *uses) {
+    if (IsNonPtrAccessChain(u.inst->opcode())) {
```
dneto0: Should also cover the case of OpCopyObject.
That's another way to generate a pointer in Logical addressing model

E.g. 
  %var = OpVariable ...
   %varcopy = OpCopyObject %ptrtype %var
   %value = OpLoad %type %varcopy

source/opt/local_single_block_elim_pass.cpp
```diff
@@ -0,0 +1,329 @@
...
+  return nonMathComp == 0;
+}
+
+ir::Instruction* LocalSingleBlockElimPass::GetPtr(
+      ir::Instruction* ip, uint32_t* varId) {
+  *varId = ip->GetSingleWordInOperand(
+      ip->opcode() == SpvOpStore ?  kSpvStorePtrId : kSpvLoadPtrId);
+  ir::Instruction* ptrInst = def_use_mgr_->GetDef(*varId);
+  ir::Instruction* varInst = ptrInst;
+  while (IsNonPtrAccessChain(varInst->opcode())) {
```
dneto0: I realize now this should also cover the case of copying the pointer through OpCopyObject.

source/opt/local_single_block_elim_pass.cpp
```diff
@@ -0,0 +1,329 @@
...
+  (void) GetPtr(storeInst, &varId);
+  return IsLiveVar(varId);
+}
+
+void LocalSingleBlockElimPass::AddStores(
+    uint32_t ptr_id, std::queue<ir::Instruction*>* insts) {
+  analysis::UseList* uses = def_use_mgr_->GetUses(ptr_id);
+  if (uses != nullptr) {
+    for (auto u : *uses) {
+      if (IsNonPtrAccessChain(u.inst->opcode()))
```
dneto0: Trace through OpCopyObject

include/spirv-tools/optimizer.hpp
greg-lunarg: Yep

include/spirv-tools/optimizer.hpp
greg-lunarg: Thanks!

source/opt/local_single_block_elim_pass.cpp
greg-lunarg: s'ok :)

source/opt/local_single_block_elim_pass.cpp
greg-lunarg: Hmmm. Somehow this slipped under my radar. Good catch. Although I must admit that I feel like I must not completely understand this operation (or maybe SPIR-V) because I am having trouble seeing the utility of this operation. Why would one want to create a copy of an object in a language where all ids are SSA?

source/opt/local_single_block_elim_pass.h
greg-lunarg: I will eventually be submitting a PR for a pass that essentially does this across all (entry point) functions.

Are you also suggesting that we *not* call DCEInst in this pass?

source/opt/local_single_block_elim_pass.h
greg-lunarg: Actually, this will give an answer for the whole module because def_use_mgr_ does its analysis across the module. In fact, I think we do want an answer for the whole module. In fact, your question helped me find a bug.

Of course, for function scope variables, only their function will *directly* be loading from them. But it is possible that if the address is passed to a function call, other functions can indirectly load from them, right? So I added logic that assumes that if the varId (actually a ptrId) is passed to a FunctionCall we should assume it is loaded.

source/opt/local_single_block_elim_pass.cpp
dneto0: Indeed I've said exactly the same thing.  (See Khronos internal bug 14178)

It turns out OpCopyObject is useful to convert to "Conventional SSA" form which is easier to handle for some purposes. See original motivation in http://bboissin.appspot.com/static/upload/bboissin-outssa-cgo09-slides.pdf

source/opt/local_single_block_elim_pass.h
dneto0: Yes, I _think_ that's where we should head, to keep code complexity lower.  What do you think?
At the very least let's have a TODO to reconsider this at some point.

source/opt/local_single_block_elim_pass.cpp
```diff
@@ -0,0 +1,337 @@
...
+      ids.push(*iid);
+    });
+    uint32_t varId = 0;
+    // Remember variable if dead load
+    if (di->opcode() == SpvOpLoad)
+      (void) GetPtr(di, &varId);
+    def_use_mgr_->KillInst(di);
+    // For all operands with no remaining uses, add their instruction
+    // to the dead instruction queue.
+    while (!ids.empty()) {
```
dneto0: It seems that ids could have been a vector, and this loop could have been range for.  I don't see why this is a while loop.
Recommend the more idiomatic for loop.

source/opt/local_single_block_elim_pass.h
```diff
@@ -0,0 +1,140 @@
...
+  bool IsLiveVar(uint32_t varId) const;
+
+  // Return true if |storeInst| is not to function variable or if its
+  // base variable has a load
+  bool IsLiveStore(ir::Instruction* storeInst);
+
+  // Add stores using |ptr_id| to |insts|
+  void AddStores(uint32_t ptr_id, std::queue<ir::Instruction*>* insts);
+
+  // Delete |inst| and iterate DCE on all its operands
```
dneto0: Should say this won't delete an OpLabel

source/opt/local_single_block_elim_pass.h
```diff
@@ -0,0 +1,140 @@
...
+  // Map from function's result id to function
+  std::unordered_map<uint32_t, ir::Function*> id2function_;
+
+  // Cache of previously seen target types
+  std::unordered_set<uint32_t> seen_target_vars_;
+
+  // Cache of previously seen non-target types
+  std::unordered_set<uint32_t> seen_non_target_vars_;
+
+  // Map from variable to its live store in block
```
dneto0: These next 3 vars should be described more clearly.
They are reset each block, and incrementally updated while scanning the block from start to finish. 

When you say "live store" or "live  load" it means "the store in this block that most recently set the variable's contents", and "live load" means "the first load in this block that has read the value contained in the variable at the time this instruction is executed".

Should also say that members of these two maps are candidates for elimination.  That will explain why when you reach a function call it's conservative to just clear all these maps.

source/opt/local_single_block_elim_pass.h
```diff
@@ -0,0 +1,140 @@
...
+  std::unordered_map<uint32_t, ir::Instruction*> var2store_;
+
+  // Map from variable to its live load in block
+  std::unordered_map<uint32_t, ir::Instruction*> var2load_;
+
+  // Set of undeletable variables
```
dneto0: Should say why a variable is undeletable. 
Something like "There is a live load of the variable, whether or not it's a target variable."

test/opt/local_single_block_elim.cpp
```diff
@@ -0,0 +1,401 @@
...
+)";
+
+  SinglePassRunAndCheck<opt::LocalSingleBlockLoadStoreElimPass>(
+        assembly, assembly, false, true);
+}
+
+// TODO(greg-lunarg): Add tests to verify handling of these cases:
+//
+//    Other target variable types
+//    InBounds Access Chains
```
dneto0: I'd like to see a simple test showing tracing through OpObjectCopy

source/opt/local_single_block_elim_pass.cpp
greg-lunarg: Thanks for the background, David.

The problem I see is that the existence of this operator seems to require that every time you grab an operand, you have to check to see if its a CopyObject. And have logic which does the trace through in the case that it is.

Thus, while CSSA may have its benefits, preserving it also has its cost in both compile time and compiler code bloat.

In light of this, I wonder how necessary preserving CSSA is for this pass. This pass is intended as a pre-SSA pass, used to make SSA rewrite more efficient by eliminating variables, loads and stores that can be eliminated using less expensive methods.

So I am doubting that it is very useful for this pass (and my other pre-SSA passes) to preserve CSSA, since once a shader is in CSSA form, it will not need to run this pass.

In light of this, I am pondering the following approach: we document that these passes do not process shaders with CopyObject. The pass will check for the presence of CopyObject and return immediately if found. This check will be preserved in the proposed Optimization Context so that subsequent pres-SSA and SSA-rewrite passes can avoid the check.

How do you feel about this approach?

source/opt/local_single_block_elim_pass.h
greg-lunarg: The idea here was to derive some efficiency from the fact that we had some idea of what needed to be deleted, what likely needed to be deleted and what likely didn't need to be deleted.

For instance, for this pass, we knew that any loads whose ids were replaced could be deleted without having to look at use/def info. We also knew which loads' we didn't have to look at. Maybe not a huge efficiency gain, but something.

Yes, we do go back and (admittedly inefficiently) re-scan each block looking for stores  to delete. But we could have saved off the stores that we did link with a replaced load and only considered them and avoided a full scan. We still could do that and maybe we should.

My thinking on DCEInst was that it would be a generic recursive instruction deleter used by all the passes (again, possibly part of the proposed Optimizer Context), so the complexity would be somewhat limited in that way.

If we did write a separate DCE pass, we would probably want to design it to optionally only look at loads and stores, instead of all instructions. And we would need to do the load/store analysis in two full scans: first to remove any and all loads, and another scan to look at the stores, which can't be fully considered until all possible loads are removed.

While I think I still favor a more targeted approach to deal with load and store deletion, I am open to continuing to consider splitting this into a separate pass, so I will add it as a TODO.

source/opt/local_single_block_elim_pass.cpp
greg-lunarg: Good idea.

source/opt/local_single_block_elim_pass.h
greg-lunarg: Yep.

source/opt/local_single_block_elim_pass.h
greg-lunarg: OK

source/opt/local_single_block_elim_pass.h
greg-lunarg: OK

source/opt/local_single_block_elim_pass.cpp
greg-lunarg: See reply to next comment.

source/opt/local_single_block_elim_pass.cpp
greg-lunarg: See reply to previous comment.

source/opt/local_single_block_elim_pass.cpp
greg-lunarg: I have proactively added code for my solution anticipating that you will agree. If you do not agree with this approach, we can just delete it and craft a different approach.

test/opt/local_single_block_elim.cpp
greg-lunarg: Anticipating that you will agree to my proposal for dealing with CopyObject, I have added a test that demonstrates that we will not process functions containing CopyObject. If you do not agree with this approach, we can delete the test and craft a new one.

source/opt/local_single_block_elim_pass.h
dneto0: Sounds good to me.  Thanks for adding the TODO.  It's good to anticipate the next reader's potential question. It signals there are tradeoffs here, and we've chosen one.

source/opt/local_single_block_elim_pass.h
```diff
@@ -121,13 +121,25 @@ class LocalSingleBlockLoadStoreElimPass : public Pass {
...
+  // at the start of each block and incrementally updated as the block
+  // is scanned. The stores are candidates for elimination. The map is
+  // conservatively cleared when a function call is encountered.
   std::unordered_map<uint32_t, ir::Instruction*> var2store_;
 
-  // Map from variable to its live load in block
+  // Map from function scope variable to a load of that variable in the
+  // current block whose value is currently valid. This map is cleared
+  // at the start of each block and incrementally updated as the block
+  // is scanned. The stores are candidates for elimination. The map is
```
dneto0: I think it's "loads are candidates for elimination"... but I think that will be clear enough to the reader that they'll auto-correct.

source/opt/local_single_block_elim_pass.cpp
```diff
@@ -202,6 +202,14 @@ void LocalSingleBlockLoadStoreElimPass::DCEInst(ir::Instruction* inst) {
 
 bool LocalSingleBlockLoadStoreElimPass::LocalSingleBlockLoadStoreElim(
     ir::Function* func) {
+  // Verify no CopyObject ops in function. This is a pre-SSA pass and
+  // is generally is not useful for code already in CSSA form.
```
dneto0: Good comment.
Personally I'd spell out "Conventional SSA" since it's not universally known.    (As in, I didn't know it and found the wrong thing first time I looked.) But this is ok.

dneto0:
Rebased, squashed, and pushed into master as  7c8da66bc

PR #657, opened 2017-05-19
Pull CalculateDominators into CFA Class

greg-lunarg:


dneto0:
Rebased and pushed into master as d6f2979

PR #658, opened 2017-05-19
Rename files to avoid libtool same member name warning on macOS

antiagainst:
Avoid having the same name for cpp files in different directories.


dneto0:
There's a lot of work going on right now in these files.  I don't want to disturb that just yet.
The warning is annoying but I actively ignore it.

We also can get conflicts with Glslang, for example.

antiagainst:
That's fair. It's not a big deal esp. sine it's for macOS.

PR #659, opened 2017-05-23
Fixed disassembly of signed 16 bit int

trenouf:
A literal operand of 32 or fewer bits occupies a 32 bit word in binary
SPIR-V. The spec does not say that the remaining bits in the word have
to be any particular value, but the disassembler was assuming that they
were zero or sign extended as applicable to the signedness of the type.

I saw a SPIR-V binary that was affected by this; the spir-v generator
had set the extra bits to 0 for a negative signed 16 bit literal, and
the disassembler gave the wrong result:

           %6 = OpTypeInt 16 1
          %31 = OpConstant %6 65436

Assembling that disassembly causes an error.

With this commit, the disassembler assumes nothing about the extra bits,
and the disassembly is correct:

          %31 = OpConstant %6 -100

### DIFF
--- a/source/parsed_operand.cpp
+++ b/source/parsed_operand.cpp
@@ -32,10 +32,10 @@ void EmitNumericLiteral(std::ostream* out, const spv_parsed_instruction_t& inst,
   if (operand.num_words == 1) {
     switch (operand.number_kind) {
       case SPV_NUMBER_SIGNED_INT:
-        *out << int32_t(word);
+        *out << (int32_t(word << (32 - operand.number_bit_width)) >> (32 - operand.number_bit_width));
         break;
       case SPV_NUMBER_UNSIGNED_INT:
-        *out << word;
+        *out << (uint32_t(word << (32 - operand.number_bit_width)) >> (32 - operand.number_bit_width));
         break;
       case SPV_NUMBER_FLOATING:
         if (operand.number_bit_width == 16) {


trenouf:
@dneto0 Please review; thanks. (Is this how you are supposed to ask a specific person for a review?)

dneto0:
Thanks for the issue.

I think spec does 

In 2.2.1 it says this about literal numbers:

Literal Number: A numeric value consuming one or more words. An instruction will determine what type a literal will be interpreted as. When the type’s bit width is larger than one word, the literal’s low-order words appear first. When the type’s bit width is less than 32-bits, the literal’s value appears in the low-order bits of the word, and the high-order bits must be 0 for a floating-point type, or 0 for an integer type with Signedness of 0, or sign extended when Signedness is 1. (Similarly for the remaining bits of widths larger than 32 bits but not a multiple of 32 bits.)

The relevant part is toward the end.

So it seems to me that the literal word has 16 high order bits being 0, and the bit 15 being 1.  This is invalid by the spec for a signed 16 bit int.  
It's a judgement call between two choices:

1. disassemble it successfully but catch the problem in the validator.
2. fail disassembly.  That seems very unhelpful.

Clearly, we disassemble it but the validator does not catch this problem.

I haven't taken a look at your patch.  Do you agree that the generated constant is invalid?

dneto0:
Looking at your patch, I don't want to apply it.  It hides the fact that the upper 16 bits are zero. So it's an inaccurate view of what's in the binary, even if what's there is invalid.  I think the disassembler is doing the right thing here, even though the input is invalid.

trenouf:
OK. So it sounds like the bug is in the generator for generating it incorrectly, and the validator for not catching it.

I believe the generator used in this case is some version of the Khronos GLSL compiler, but I need to track down exactly what.

dneto0:
Ok.  I'll close this PR.  I've added an issue/task for the validator enhancement.  (The validator is a long way off from being complete.)  I'd appreciate you filing an issue upstream with whatever generator produced that code in the first place.

Thanks!

PR #661, opened 2017-05-25
Added Huffman codec to utils

atgoo:
The codec is intended to be used by compression tools and is designed to be compatible with bit_stream utils, but can be built and used independently.

### DIFF
--- a/source/util/bit_stream.h
+++ b/source/util/bit_stream.h
@@ -17,6 +17,7 @@
 #ifndef LIBSPIRV_UTIL_BIT_STREAM_H_
 #define LIBSPIRV_UTIL_BIT_STREAM_H_
 
+#include <algorithm>
 #include <bitset>
 #include <cstdint>
 #include <string>
--- a/source/util/huffman_codec.h
+++ b/source/util/huffman_codec.h
@@ -19,6 +19,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <functional>
 #include <queue>
 #include <iomanip>
 #include <map>
--- a/source/util/huffman_codec.h
+++ b/source/util/huffman_codec.h
@@ -25,6 +25,7 @@
 #include <map>
 #include <memory>
 #include <ostream>
+#include <sstream>
 #include <stack>
 #include <tuple>
 #include <unordered_map>
@@ -53,7 +54,7 @@ class HuffmanCodec {
     queue_vector.reserve(hist.size());
     std::priority_queue<Node*, std::vector<Node*>,
         std::function<bool(const Node*, const Node*)>>
-	    queue(LeftIsBigger, std::move(queue_vector));
+            queue(LeftIsBigger, std::move(queue_vector));
 
     // Put all leaves in the queue.
     for (const auto& pair : hist) {
@@ -73,7 +74,7 @@ class HuffmanCodec {
       Node* right = queue.top();
       queue.pop();
 
-      // If the queue is empty at this point, then the last node contains is
+      // If the queue is empty at this point, then the last node is
       // the root of the complete Huffman tree.
       if (queue.empty()) {
         root_ = right;
@@ -83,7 +84,7 @@ class HuffmanCodec {
       Node* left = queue.top();
       queue.pop();
 
-      // Combile left and right into a new tree and push it into the queue.
+      // Combine left and right into a new tree and push it into the queue.
       Node* parent = CreateNode();
       parent->weight = right->weight + left->weight;
       parent->left = left;
@@ -99,10 +100,14 @@ class HuffmanCodec {
   // w------w------'x'
   //        w------'y'
   // Where w stands for the weight of the node.
+  // Right tree branches appear above left branches. Taking the right path
+  // adds 1 to the code, taking the left adds 0.
   void PrintTree(std::ostream& out) {
     PrintTreeInternal(out, root_, 0);
   }
 
+  // Traverses the tree and prints the Huffman table: value, code
+  // and optionally node weight for every leaf.
   void PrintTable(std::ostream& out, bool print_weights = true) {
     std::queue<std::pair<Node*, std::string>> queue;
     queue.emplace(root_, "");
@@ -126,12 +131,15 @@ class HuffmanCodec {
     }
   }
 
+  // Returns the Huffman table. The table was built at at construction time,
+  // this function just returns a const reference.
   const std::unordered_map<Val, std::pair<uint64_t, size_t>>&
       GetEncodingTable() const {
     return encoding_table_;
   }
 
-  // Encodes |val| and stores the value in the lower |num_bits| of |bits|.
+  // Encodes |val| and stores its Huffman code in the lower |num_bits| of
+  // |bits|. Returns false of |val| is not in the Huffman table.
   bool Encode(const Val& val, uint64_t* bits, size_t* num_bits) {
     auto it = encoding_table_.find(val);
     if (it == encoding_table_.end())
@@ -142,7 +150,10 @@ class HuffmanCodec {
   }
 
   // Reads bits one-by-one using callback |read_bit| until a match is found.
-  // Matching value is stored in |val|. Returns false if the stream terminates
+  // Matching value is stored in |val|. Returns false if |read_bit| terminates
+  // before a code was mathced.
+  // |read_bit| has type bool func(bool* bit). When called, the next bit is
+  // stored in |bit|. |read_bit| returns false if the stream terminates
   // prematurely.
   bool DecodeFromStream(const std::function<bool(bool*)>& read_bit, Val* val) {
     Node* node = root_;
@@ -173,11 +184,16 @@ class HuffmanCodec {
   struct Node {
     Val val = Val();
     uint32_t weight = 0;
+    // Ids are issued sequentially starting from 1. Ids are used as an ordering
+    // tie-breaker, to make sure that the ordering (and resulting coding scheme)
+    // are consistent accross multiple platforms.
     uint32_t id = 0;
     Node* left = nullptr;
     Node* right = nullptr;
   };
 
+  // Returns true if |left| has bigger weight than |right|. Node ids are
+  // used as tie-breaker.
   static bool LeftIsBigger(const Node* left, const Node* right) {
     if (left->weight == right->weight) {
       assert (left->id != right->id);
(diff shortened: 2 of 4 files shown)


source/util/bit_stream.h
```diff
@@ -17,6 +17,7 @@
 #ifndef LIBSPIRV_UTIL_BIT_STREAM_H_
 #define LIBSPIRV_UTIL_BIT_STREAM_H_
 
+#include <algorithm>
```
dneto0: For std::reverse.

It turns out we also need cassert to make this self contained.  Can be in another CL.

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+
+  // Prints the Huffman tree in the following format:
+  // w------w------'x'
+  //        w------'y'
+  // Where w stands for the weight of the node.
+  void PrintTree(std::ostream& out) {
+    PrintTreeInternal(out, root_, 0);
+  }
+
+  void PrintTable(std::ostream& out, bool print_weights = true) {
```
dneto0: Method comment.

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+      parent->right = right;
+      queue.push(parent);
+    }
+
+    // Traverse the tree and form encoding table.
+    CreateEncodingTable();
+  }
+
+  // Prints the Huffman tree in the following format:
+  // w------w------'x'
```
dneto0: Good comment.  Examples are golden!

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+        if (node->left)
+          queue.emplace(node->left, code + "0");
+
+        if (node->right)
+          queue.emplace(node->right, code + "1");
+      }
+    }
+  }
+
+  const std::unordered_map<Val, std::pair<uint64_t, size_t>>&
```
dneto0: Method comment.

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+    auto it = encoding_table_.find(val);
+    if (it == encoding_table_.end())
+      return false;
+    *bits = it->second.first;
+    *num_bits = it->second.second;
+    return true;
+  }
+
+  // Reads bits one-by-one using callback |read_bit| until a match is found.
+  // Matching value is stored in |val|. Returns false if the stream terminates
```
dneto0: Presumably a decoded false maps to 0 and true maps to 1.  You should say so.

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+      }
+    }
+  }
+
+  const std::unordered_map<Val, std::pair<uint64_t, size_t>>&
+      GetEncodingTable() const {
+    return encoding_table_;
+  }
+
+  // Encodes |val| and stores the value in the lower |num_bits| of |bits|.
```
dneto0: "stores the value" ->  "stores the Huffman code for |val|"  ?

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+    assert (0);
+    return false;
+  }
+
+ private:
+  // Huffman tree node.
+  struct Node {
+    Val val = Val();
+    uint32_t weight = 0;
+    uint32_t id = 0;
```
dneto0: What is the meaning of "id" ?
Should say something like "Nodes are assigned sequential ids as they are created, starting with 1.  Ids are used to break ordering ties between values with equal weight."

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+    Node* left = nullptr;
+    Node* right = nullptr;
+  };
+
+  static bool LeftIsBigger(const Node* left, const Node* right) {
```
dneto0: Method comment.

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+    if (left->weight == right->weight) {
+      assert (left->id != right->id);
+      return left->id > right->id;
+    }
+    return left->weight > right->weight;
+  }
+
+  // Prints subtree (helper function used by PrintTree).
+  static void PrintTreeInternal(std::ostream& out, Node* node, size_t depth) {
```
dneto0: References normally should be const.  Recommend std::ostream* here.
But this is such a common pattern that we've let it slip by before.   Ok to keep if you really want.

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+    if (!node)
+      return;
+
+    if (!node->right && !node->left) {
+      out << node->val << std::endl;
+    } else {
+      if (node->right) {
+        out << std::setfill('-') << std::left << std::setw(7) << node->right->weight;
```
dneto0: You should restore the fill char and width properties of the stream.  It's good stream hygiene.
E.g. see https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/util/hex_float.h#L719

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+        PrintTreeInternal(out, node->right, depth + 1);
+      }
+
+      if (node->left) {
+        out << std::string(depth * 7, ' ');
```
dneto0: Recommend making a local constant to replace the magic "7" values in this function.

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+    // and pushing the root of the new tree in the queue.
+    while (true) {
+      // We push a node at the end of each iteration, so the queue is never
+      // supposed to be empty at this point, unless there are no leaves, but
+      // that case was already handled.
+      assert(!queue.empty());
+      Node* right = queue.top();
+      queue.pop();
+
+      // If the queue is empty at this point, then the last node contains is
```
dneto0: Grammar: delete the "contains" on this line.

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+      // the root of the complete Huffman tree.
+      if (queue.empty()) {
+        root_ = right;
+        break;
+      }
+
+      Node* left = queue.top();
+      queue.pop();
+
+      // Combile left and right into a new tree and push it into the queue.
```
dneto0: "Combine"

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+        out << std::setfill('-') << std::left << std::setw(7) << node->left->weight;
+        PrintTreeInternal(out, node->left, depth + 1);
+      }
+    }
+  }
+
+  // Traverses the Huffman tree and saves paths to the leaves as bit
+  // sequences to encoding_table_.
+  void CreateEncodingTable() {
+    std::queue<std::tuple<Node*, uint64_t, size_t>> queue;
```
dneto0: What are the uint64_t and size_t here? 
Probably better to make a struct so you'd have better names?

dneto0: Looks like the uint64_t is the prefix.  Why are 64 bits going to be enough?

Intuition says a worst case code length would be for a degenerately imbalanced tree.  Why does that not happen?

source/util/huffman_codec.h
```diff
@@ -0,0 +1,258 @@
...
+        auto insertion_result = encoding_table_.emplace(
+            node->val, std::pair<uint64_t, size_t>(bits, depth));
+        assert(insertion_result.second);
+        (void)insertion_result;
+      } else {
+        if (node->left)
+          queue.emplace(node->left, bits, depth + 1);
+
+        if (node->right)
+          queue.emplace(node->right, bits | (1ULL << depth), depth + 1);
```
dneto0: I'd find it slightly easier to read the 1ULL as uint64_t(1)

source/util/huffman_codec.h
dneto0: Explain the return value.

source/util/huffman_codec.h
dneto0: Have to explain who the callback works.  There are two bools returned: one by return and the other by pointer param.  I shouldn't have to read the body of this method to know how to make a read_bit callback.

source/util/huffman_codec.h
dneto0: Should say that the "1" subtree appears above its  "0" sibling.

test/huffman_codec.cpp
```diff
@@ -0,0 +1,220 @@
...
+  std::string bits_;
+  size_t pos_ = 0;
+};
+
+TEST(Huffman, PrintTree) {
+  HuffmanCodec<std::string> huffman(GetTestSet());
+  std::stringstream ss;
+  huffman.PrintTree(ss);
+
+  const std::string expected = std::string(R"(
```
dneto0: Elegant test cases.  I'm glad you put this tree output here first!   I'm sure you're glad too.  :-)

source/util/huffman_codec.h
atgoo: I thinks in order to fill 64 bits with imbalanced values we would need something like uint128_t for the histogram counts :)

atgoo: Actually no, it just needs to grow faster than fibonacci. But our counts are uint32_t, so we are safe.

dneto0: I figured that would be the essence of the argument.  You need to say that in a comment in the code.

Furthermore, what if I give you a big set of items, each with weight 0.  That's allowed by the signature of the API.  How do you know you won't get a badly balanced tree?

atgoo: I don't think there is such a thing a poorly balanced Huffman tree. If the frequencies are growing at ~2^n then we would have a ladder, but this can't happen. When all nodes have equal weights, then we have a flat table ordered by ids. All codes would have similar length. That's not an edge case, but rather anticlimactic one.

source/util/huffman_codec.h
```diff
@@ -0,0 +1,294 @@
...
+      return;
+
+    const size_t kTextFieldWidth = 7;
+
+    if (!node->right && !node->left) {
+      out << node->val << std::endl;
+    } else {
+      if (node->right) {
+        std::stringstream label;
+        label << std::setfill('-') << std::left << std::setw(kTextFieldWidth)
```
dneto0: Unaddressed comment from before: https://github.com/KhronosGroup/SPIRV-Tools/pull/661#discussion_r121531093

source/util/huffman_codec.h
dneto0: If all the weights are 0 then everything is ordered by 0 rather than by weights.  Then I think the queue always picks off the nodes with largest IDs to combine them.  So you get a ladder.  ??

This problem goes away if you have as a precondition to the entire table that the weights are assumed to be strictly positive.

Also, you must put the analysis in the *comments* rather than in the review.

source/util/huffman_codec.h
atgoo: These options are not global and the stream object is destroyed anyway.

dneto0: It's a mutation of a the out parameter.   So you if you want to keep it this way then your method comment should mention this side effect.

dneto0: Whoops. Misread that.

dneto0:
Remaining changes required:

- HuffmanCodec constructor comment should say that it assumes the histogram does not have entries with zero counts. If you want to be more relaxed then that, then give bounds somehow. But I'd just recommend being strict.
- CreateEncodingTable implementation comment (around the Context struct, probably) should say that because weights are assumed non-zero, then the huffman tree is nearly balanced, and so 64 bits is sufficient as a path label down to any leaf.

atgoo:
> Remaining changes required:

Done.

dneto0:
Squashed and pushed into master as 40a282961

PR #662, opened 2017-05-25
CFA: Move TraversalRoots and ComputeAugmentedCFG into CFA

greg-lunarg:
This the final step in leveraging the dominator calculation code from the validator in other parts of spirv-tools, specifically spirv-opt.

CFA continues to be stateless, but pushing data into it in subsequent updates remains a possibility, and is perhaps desirable.


dneto0:
Indeed, right now the validator stores immediate dominator in its representation of a block itself. But those dominators should probably move to state in the CFA object (or maybe as returned value from the dominator calculation).  ??

PR #665, opened 2017-05-28
Support building on FreeBSD

rdb:
This patch lets me build SPIRV-Tools on FreeBSD 11.  I ran the test suite and there were no failures.


dneto0:
Thank you for the contribution!

PR #668, opened 2017-05-31
Add LocalSingleStoreElimPass

greg-lunarg:
This is another pass in a series of passes designed to reduce the size of
SPIR-V files.

For each entry point function, this pass eliminates loads and stores for 
function scope variable that are stored to only once, where possible. Only
 whole variable loads and stores are eliminated; access-chain references are
not optimized. Replace all loads of such variables with the value that is
stored and eliminate any resulting dead code.

Currently, the presence of access chains and function calls can inhibit this
pass, however the Inlining and LocalAccessChainConvert passes can make it
more effective.

This pass will reduce the work needed to be done by LocalSingleBlockElim
and LocalSSARewrite and can improve the effectiveness of other passes such
as DeadBranchElimination which depend on values for their analysis.

### DIFF
--- a/source/opt/local_single_store_elim_pass.cpp
+++ b/source/opt/local_single_store_elim_pass.cpp
@@ -132,14 +132,14 @@ void LocalSingleStoreElimPass::SingleStoreAnalyze(ir::Function* func) {
           non_ssa_vars_.insert(varId);
           continue;
         }
-        // If already stored, disqualify it
+        // Ignore variables with multiple stores
         if (ssa_var2store_.find(varId) != ssa_var2store_.end()) {
           non_ssa_vars_.insert(varId);
           ssa_var2store_.erase(varId);
           continue;
         }
-        // Remember iterator of variable's store and it's
-        // ordinal position in function
+        // Remember pointer to variable's store and it's
+        // ordinal position in block
         ssa_var2store_[varId] = &*ii;
         store2idx_[&*ii] = instIdx;
         store2blk_[&*ii] = &*bi;
--- a/source/opt/local_single_store_elim_pass.cpp
+++ b/source/opt/local_single_store_elim_pass.cpp
@@ -27,7 +27,6 @@ static const int kSpvLoadPtrId = 0;
 static const int kSpvAccessChainPtrId = 0;
 static const int kSpvTypePointerStorageClass = 0;
 static const int kSpvTypePointerTypeId = 1;
-static const int kSpvExtInstInstruction = 0;
 
 // Universal Limit of ResultID + 1
 static const int kInvalidId = 0x400000;
--- a/source/opt/local_single_store_elim_pass.cpp
+++ b/source/opt/local_single_store_elim_pass.cpp
@@ -234,7 +234,7 @@ bool LocalSingleStoreElimPass::Dominates(
     ir::BasicBlock* blk0, uint32_t idx0,
     ir::BasicBlock* blk1, uint32_t idx1) {
   if (blk0 == blk1)
-    return idx0 < idx1;
+    return idx0 <= idx1;
   ir::BasicBlock* b = blk1;
   while (idom_[b] != b) {
     b = idom_[b];
--- a/source/opt/local_single_store_elim_pass.cpp
+++ b/source/opt/local_single_store_elim_pass.cpp
@@ -1,0 +1,444 @@
+// Copyright (c) 2017 The Khronos Group Inc.
+// Copyright (c) 2017 Valve Corporation
+// Copyright (c) 2017 LunarG Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cfa.h"
+#include "iterator.h"
+#include "local_single_store_elim_pass.h"
+
+static const int kSpvEntryPointFunctionId = 1;
+static const int kSpvStorePtrId = 0;
+static const int kSpvStoreValId = 1;
+static const int kSpvLoadPtrId = 0;
+static const int kSpvAccessChainPtrId = 0;
+static const int kSpvTypePointerStorageClass = 0;
+static const int kSpvTypePointerTypeId = 1;
+
+// Universal Limit of ResultID + 1
+static const int kInvalidId = 0x400000;
+
+namespace spvtools {
+namespace opt {
+
+bool LocalSingleStoreElimPass::IsNonPtrAccessChain(const SpvOp opcode) const {
+  return opcode == SpvOpAccessChain || opcode == SpvOpInBoundsAccessChain;
+}
+
+bool LocalSingleStoreElimPass::IsMathType(
+    const ir::Instruction* typeInst) const {
+  switch (typeInst->opcode()) {
+  case SpvOpTypeInt:
+  case SpvOpTypeFloat:
+  case SpvOpTypeBool:
+  case SpvOpTypeVector:
+  case SpvOpTypeMatrix:
+    return true;
+  default:
+    break;
+  }
+  return false;
+}
+
+bool LocalSingleStoreElimPass::IsTargetType(
+    const ir::Instruction* typeInst) const {
+  if (IsMathType(typeInst))
+    return true;
+  if (typeInst->opcode() == SpvOpTypeArray)
+    return IsMathType(def_use_mgr_->GetDef(typeInst->GetSingleWordOperand(1)));
+  if (typeInst->opcode() != SpvOpTypeStruct)
+    return false;
+  // All struct members must be math type
+  int nonMathComp = 0;
+  typeInst->ForEachInId([&nonMathComp,this](const uint32_t* tid) {
+    ir::Instruction* compTypeInst = def_use_mgr_->GetDef(*tid);
+    if (!IsMathType(compTypeInst)) ++nonMathComp;
+  });
+  return nonMathComp == 0;
+}
+
+ir::Instruction* LocalSingleStoreElimPass::GetPtr(
+      ir::Instruction* ip, uint32_t* varId) {
+  *varId = ip->GetSingleWordInOperand(
+      ip->opcode() == SpvOpStore ?  kSpvStorePtrId : kSpvLoadPtrId);
+  ir::Instruction* ptrInst = def_use_mgr_->GetDef(*varId);
+  ir::Instruction* varInst = ptrInst;
+  while (IsNonPtrAccessChain(varInst->opcode())) {
+    *varId = varInst->GetSingleWordInOperand(kSpvAccessChainPtrId);
+    varInst = def_use_mgr_->GetDef(*varId);
+  }
+  return ptrInst;
+}
+
+bool LocalSingleStoreElimPass::IsTargetVar(uint32_t varId) {
+  if (seen_non_target_vars_.find(varId) != seen_non_target_vars_.end())
+    return false;
+  if (seen_target_vars_.find(varId) != seen_target_vars_.end())
+    return true;
+  const ir::Instruction* varInst = def_use_mgr_->GetDef(varId);
+  assert(varInst->opcode() == SpvOpVariable);
+  const uint32_t varTypeId = varInst->type_id();
+  const ir::Instruction* varTypeInst = def_use_mgr_->GetDef(varTypeId);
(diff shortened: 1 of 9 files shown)


dneto0:
Please rebase against master

greg-lunarg:
Rebased.

test/opt/local_single_store_elim_test.cpp
```diff
@@ -0,0 +1,214 @@
...
+)";
+
+  SinglePassRunAndCheck<opt::LocalSingleStoreElimPass>(assembly, assembly,
+      true, true);
+}
+
+// TODO(greg-lunarg): Add tests to verify handling of these cases:
+//
+//    Other types
+//    Multiple loads optimized
```
dneto0: I'd like to see these cases before merging:
- multiple loads (like you say here)
- where there is a single store to a vector, that store dominates a load of one of its elements. Basically:
        OpStore %vec2var %vec2
        %ptr = OpAccessChain %float %vec2var %one
         %floatval = OpLoad %float %ptr
   Should not result in replacement
- variable has an initializer (e.g. OpConstantNull), and a single store is via access chain to the second element, and that store dominates a load of the entire vector.  Should not result in replacement of the load. (Extra work would need to be done to analyze the value enough to eliminate it.)
- Should not optimize if the pointer to a load came from OpCopyObject.
- Should not optimize if there are stores-by-another-instruction:  GLSL extended Modf, Frexp, OpAtomic*

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -0,0 +1,445 @@
...
+static const int kInvalidId = 0x400000;
+
+namespace spvtools {
+namespace opt {
+
+bool LocalSingleStoreElimPass::IsNonPtrAccessChain(const SpvOp opcode) const {
+  return opcode == SpvOpAccessChain || opcode == SpvOpInBoundsAccessChain;
+}
+
+bool LocalSingleStoreElimPass::IsMathType(
```
dneto0: We've now got 3 passes with common functionality.  I've filed a "card" for future work to refactor them into a single implementation.
https://github.com/KhronosGroup/SPIRV-Tools/projects/3#card-3359497

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -0,0 +1,445 @@
...
+          non_ssa_vars_.insert(varId);
+          continue;
+        }
+        // If already stored, disqualify it
+        if (ssa_var2store_.find(varId) != ssa_var2store_.end()) {
+          non_ssa_vars_.insert(varId);
+          ssa_var2store_.erase(varId);
+          continue;
+        }
+        // Remember iterator of variable's store and it's
```
dneto0: You're not recording the *iterator* but rather the ir::Instruction pointer.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -0,0 +1,445 @@
...
+          non_ssa_vars_.insert(varId);
+          ssa_var2store_.erase(varId);
+          continue;
+        }
+        // Verify target type and function storage class
+        if (!IsTargetVar(varId)) {
+          non_ssa_vars_.insert(varId);
+          continue;
+        }
+        // If already stored, disqualify it
```
dneto0: I think a clearer comment would be:  "Ignore variables with multiple stores."

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -0,0 +1,445 @@
...
+  return [this](const ir::BasicBlock* block) {
+    auto apmi = augmented_predecessors_map_.find(block);
+    if (apmi != augmented_predecessors_map_.end())
+      return &(*apmi).second;
+    auto pmi = predecessors_map_.find(block);
+    return &(*pmi).second;
+  };
+}
+
+void LocalSingleStoreElimPass::CalculateImmediateDominators(
```
dneto0: I presume this is lifted from the validator.  :-)

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -0,0 +1,445 @@
...
+  idom_.clear();
+  for (auto edge : edges)
+    idom_[edge.first] = edge.second;
+}
+
+bool LocalSingleStoreElimPass::Dominates(
+    ir::BasicBlock* blk0, uint32_t idx0,
+    ir::BasicBlock* blk1, uint32_t idx1) {
+  if (blk0 == blk1)
+    return idx0 < idx1;
```
dneto0: An instruction always dominates itself.  You probably don't care about that case, but I'd like to be consistent with the standard definition.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -0,0 +1,445 @@
...
+    uint32_t instIdx = 0;
+    for (auto ii = bi->begin(); ii != bi->end(); ++ii, ++instIdx) {
+      if (ii->opcode() != SpvOpLoad)
+        continue;
+      uint32_t varId;
+      ir::Instruction* ptrInst = GetPtr(&*ii, &varId);
+      // Skip access chain loads
+      if (IsNonPtrAccessChain(ptrInst->opcode()))
+        continue;
+      assert(ptrInst->opcode() == SpvOpVariable);
```
dneto0: Use "Continue" in the false case instead of asserting out.

That will let you gracefully handle the case where the pointer is from OpCopyObject.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -0,0 +1,445 @@
...
+        store2idx_[&*ii] = instIdx;
+        store2blk_[&*ii] = &*bi;
+      } break;
+      case SpvOpFunctionCall: {
+        // For now, empty SSA variable set and terminate analysis
+        // TODO(): Add logic to more optimally handle function calls.
+        ssa_var2store_.clear();
+        return;
+      } break;
+      default:
```
dneto0: I'm thinking of other weird cases, and there are lots of them.
E.g.    Other instructions that can modify the variable:  GLSL extended Modf, Frexp;  Also OpAtomic*
Other instructions can reference the store:  OpLifetimeStart, OpLifetimeStop.   OpSelect, OpCopyObject.

Maybe best thing to do here is to scan instructions other than Load, and insert any id arguments into non_ssa_vars_.   
It's not clear to me if this is best done during an initial pass.  Maybe it's clearer that way?

Is that the generic way to handle the OpFunctionCall  exception?

Actually, this also applies probably to local_single_block_elim_pass  ??  I filed https://github.com/KhronosGroup/SPIRV-Tools/projects/3#card-3359802  to track.

source/opt/local_single_store_elim_pass.h
```diff
@@ -0,0 +1,212 @@
...
+  // variable of target type, false otherwise. Updates caches of target 
+  // and non-target variables.
+  bool IsTargetVar(uint32_t varId);
+
+  // Find all function scope variables in |func| that are stored to
+  // only once (SSA) and map to their stored value id. Only analyze
+  // variables of scalar, vector, matrix types and struct and array
+  // types comprising only these types. Currently this analysis is
+  // is not done in the presence of function calls. TODO(): Allow
+  // analysis in the prensence of function calls.
```
greg-lunarg: Doh!

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: At one point you had thought that you and your team were going to at least put the infrastructure in place to support not only common functionality between passes, but also a persistent state along with a mechanism to semi-automatically invalidate the state if needed. Is this still in your plans?

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: OK.

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: Yep.

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: Presuming you are recommending idx0 <= idx1. Good catch.

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: Good idea.

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: Yes, from a number of places. Perhaps there is more leveraging that can be done here.

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: I appreciate you catching these.

For the moment, my preference is to completely bail on seeing OpAtomics, Modf, Frexp or OpCopyMemory and leave better handling of these as a TODO.

My reading of OpLifetime* is that these are optimization hints and can be safely ignored. Let me know if you disagree.

Re OpSelect, according to the spec validation rules, OpSelect is not a designated consumer or producer of pointers under logical addressing. This pass bails on non-logical addressing modules.

I will address OpCopyObject in the following comment.

greg-lunarg: Thinking about this a little more, I think ignoring OpLifetime* is lossy, so I am going to add them to the "bail" list.

Re OpCopyObject, like LocalSingleBlockLoadStoreElim, this pass is intended as a Pre-SSA rewrite pass, so I assert it doesn't make sense to spend logic handling CSSA. So I will add OpCopyObject to the bail list as well.

greg-lunarg: And yes, I will put the "bail" code in a separate, initial pass so it is clearer.

source/opt/local_single_store_elim_pass.cpp
dneto0: Yes.  That's why I've filed a card to track it.  Admittedly, it's not on the "hot" list.

dneto0: On the other hand, some of these are more specific than context-wide functionality.  A few of these, like "is this a target variable" is more specific to the load-store elimination strategies, and likely belong in a helper class specific to these few passes.    Exact strategy to be determined.

source/opt/local_single_store_elim_pass.cpp
dneto0: All sounds good to me.

source/opt/local_single_store_elim_pass.cpp
dneto0: Nothing to worry about for now.  I just didn't read the code very carefully, as that's a bunch of stuff I've rewritten about a year ago.

test/opt/local_single_store_elim_test.cpp
greg-lunarg: Done. Combined the last two since they are covered by the same logic.

greg-lunarg:
I think everything has now been addressed.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -0,0 +1,494 @@
...
+      case SpvOpFunctionCall:
+        return true;
+      // This is a pre-SSA pass, so Conventional SSA form including
+      // OpCopyObject is not supported.
+      case SpvOpCopyObject:
+        return true;
+      // Frexp and Modf extended ops reference memory and are unsupported
+      case SpvOpExtInst: {
+        uint32_t op = inst.GetSingleWordInOperand(kSpvExtInstInstruction);
+        if (op == GLSLstd450Frexp || op == GLSLstd450Modf)
```
dneto0: Hmmm.  How do you know this extended instruction is from GLSL.std.450?

There are similar bad cases in OpenCL extended instructions, e.g. frexp, fract, lgamma_r, and a few dozen more(!)

Overall, the strategy of looking for a few bad instructions is both too weak and too strong. Too weak because it doesn't cover new cases you can't anticipate, like new extended instruction sets.  Too strong because an AtomicXor on a variable in StorageBuffer storage class shouldn't affect the ability to eliminate loads and stores in Function storage class.

I think it's best to go with what I suggested earlier:  In SingleStorageAnalyze, scan instructions other than Load (and Store), and insert any id arguments into non_ssa_vars_.  That should fail safely:  newly invented instructions that use a pointer-to-variable will cause the algorithm to gracefully avoid the optimization.

I realize this reverses my previous approval to a pre-pass to bail on bad cases. Now that I see the code I see the problem with it.

greg-lunarg: Yes, I agree with the approach. Thanks!

greg-lunarg: I pretty much did what you suggested, but took a slightly different path.

In SingleStoreAnalyze, rather than scan every instruction, when I have a candidate variable that I have discovered through a Store, I call HasOnlySupportedRefs(), which looks at all the variable's uses and makes sure they are ultimately only loads and stores, either direct or through access chains. If it fails, I add the variable to non_ssa_vars_. If it passes, the routine caches it in supported_ref_ptrs_ to avoid doing the analysis again. I think this might be slightly more efficient than looking at every operand of every instruction.

dneto0:
Rebased, squashed, and pushed into master as 0c5722fc019

PR #669, opened 2017-06-01
Add InsertExtractElimPass

greg-lunarg:
This pass is another in a series designed to reduce SPIR-V file size.

This pass processes each entry point function in the module, searching for
extracts on a sequence of inserts. It further searches the sequence for an
insert with indices identical to the extract. If such an insert is
found before hitting a conflicting insert, the extract's result id is
replaced with the id of the values from the insert.

Besides removing extracts this pass enables subsequent dead code elimination
passes to delete the inserts. This pass performs best after access chains are
converted to inserts and extracts and local loads and stores are eliminated.


source/opt/insert_extract_elim.cpp
```diff
@@ -0,0 +1,125 @@
...
+};
+
+Pass::Status InsertExtractElimPass::ProcessImpl() {
+  bool modified = false;
+
+  // Process all entry point functions.
+  for (auto& e : module_->entry_points()) {
+    ir::Function* fn =
+        id2function_[e.GetSingleWordOperand(kSpvEntryPointFunctionId)];
+    modified = modified || EliminateInsertExtract(fn);
```
dneto0: Short circuiting seems the wrong thing here.    It means you'll stop after modifying only one entry point.   That will be the common case with glslang-generated modules, but wrong in general.

(This afflicts the other passes too, it seems.  I checked all three local_*cpp and inline_pass.cpp)

source/opt/insert_extract_elim.cpp
```diff
@@ -0,0 +1,125 @@
...
+  return true;
+}
+
+bool InsertExtractElimPass::ExtInsConflict(const ir::Instruction* extInst,
+    const ir::Instruction* insInst) const {
+  if (extInst->NumInOperands() == insInst->NumInOperands() - 1)
+    return false;
+  uint32_t extNumIdx = extInst->NumInOperands() - 1;
+  uint32_t insNumIdx = insInst->NumInOperands() - 2;
+  uint32_t numIdx = extNumIdx < insNumIdx ? extNumIdx : insNumIdx;
```
dneto0: Prefer std::min here.   It's easier to read.  (But you'll need to include `<algorithm>`)

source/opt/insert_extract_elim.cpp
```diff
@@ -0,0 +1,125 @@
...
+          if (ExtInsConflict(&*ii, cinst))
+            break;
+          if (ExtInsMatch(&*ii, cinst)) {
+            replId = cinst->GetSingleWordInOperand(kSpvInsertObjectId);
+            break;
+          }
+          cid = cinst->GetSingleWordInOperand(kSpvInsertCompositeId);
+          cinst = def_use_mgr_->GetDef(cid);
+        }
+        if (replId == 0)
```
dneto0: This test against 0 and break works but I read it as slightly fragile.
I would have reversed the test:
```
if (replId !=0 ) {
   ... do the replacement...
} 
break;
```

test/opt/insert_extract_elim_test.cpp
```diff
@@ -0,0 +1,263 @@
...
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::InsertExtractElimPass>(assembly, 
+      assembly, true, true);
+}
+
+// TODO(greg-lunarg): Add tests to verify handling of these cases:
+// Non-target types not optimized
```
dneto0: This line doesn't apply?  We don't have particular target types in mind.

test/opt/insert_extract_elim_test.cpp
```diff
@@ -0,0 +1,263 @@
...
+%_ptr_Function_v4float = OpTypePointer Function %v4float
+%_ptr_Output_v4float = OpTypePointer Output %v4float
+%gl_FragColor = OpVariable %_ptr_Output_v4float Output
+%main = OpFunction %void None %8
+%18 = OpLabel
+%s0 = OpVariable %_ptr_Function_S_t Function
+%19 = OpLoad %v4float %BaseColor
+%20 = OpLoad %S_t %s0
+%21 = OpCompositeInsert %S_t %19 %20 1
+%22 = OpCompositeInsert %S_t %float_0 %21 1 2
```
dneto0: This is a conflicting insert because its index list is longer than the extract index list. 

Please add a case where the extract index list is longer than the conflicting insert list. 
E.g. insert a vec4, but the later extract is pulling out one of its components.

source/opt/insert_extract_elim.cpp
dneto0: Actually, suggest this form:

```
 modified = EliminateInsertExtract(fn) || modified;
```

That's just as concise and avoids the bug.

source/opt/insert_extract_elim.cpp
greg-lunarg: OK.

source/opt/insert_extract_elim.cpp
greg-lunarg: OK.

source/opt/insert_extract_elim.cpp
greg-lunarg: OK.

source/opt/insert_extract_elim.cpp
```diff
@@ -0,0 +1,125 @@
...
+namespace spvtools {
+namespace opt {
+
+bool InsertExtractElimPass::ExtInsMatch(const ir::Instruction* extInst,
+    const ir::Instruction* insInst) const {
+  if (extInst->NumInOperands() != insInst->NumInOperands() - 1)
+    return false;
+  uint32_t numIdx = extInst->NumInOperands() - 1;
+  for (uint32_t i = 0; i < numIdx; ++i)
+    if (extInst->GetSingleWordInOperand(i + 1) !=
```
atgoo: Are indexes guaranteed to be in the same order?

test/opt/insert_extract_elim_test.cpp
greg-lunarg: Done. And one better. I added an insert that matched the extract's component, but added a full insert in between that wipes out the first insert. The extract is not eliminated.

test/opt/insert_extract_elim_test.cpp
greg-lunarg: Gone.

source/opt/insert_extract_elim.cpp
greg-lunarg: Yes. For an insert and an extract both addressing the same component in the same type, the indices will be in the same order.

dneto0:
Rebased, squashed, and pushed into master as 6136bf9e0b8f8

source/opt/insert_extract_elim.cpp
atgoo: Sorry, this is the wrong function. In ExtInsConflict the indices are compared in order as they come. So for example if I do the following operations
insert some_vec3 0, 1, 2
insert some_vec2 1, 3
extract 0, 1, 2
then value at arr[1] was changed but it won't be detected.
(maybe I don't understand how these operations work)

greg-lunarg: Your example is not quite consistent with my understanding of insert. For example, if the second insert inserted a vec2, then the first insert would insert a float.

Could you come up with a valid SPIR-V sequence that expresses your concern?

atgoo: I don't know how this is used in SPIR-V. So nevermind.

PR #672, opened 2017-06-07
Don't install googletest and googlemock

rdb:
When building with the Google Test framework in the external directory, calling `make install` would also install Google Test and Mock libraries and header files into my system.  This change prevents that.


antiagainst:
Thanks, @rdb!

PR #674, opened 2017-06-07
Add BlockMergePass

greg-lunarg:
This pass is another in a series designed to reduce the size of SPIR-V files.

This pass searches for blocks with a single Branch to a block with no
other predecessors. It merges the blocks into a single block.

The pass is most useful after Dead Branch Elimination, which can leave
such sequences of blocks. Merging them makes subsequent optimization passes
more effective, such as single block local store-load elimination.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -196,6 +196,9 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 //
 // While this pass reduces the number of occurrences of this sequence, at
 // this time it does not guarantee all such sequences are eliminated.
+//
+// Presence of phi instructions can inhibit this optimization. Handling
+// these is left for future improvements. 
 Optimizer::PassToken CreateBlockMergePass();
 
 // Creates an inline pass.
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -187,7 +187,8 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 
 // Creates a block merge pass.
 // This pass searches for blocks with a single Branch to a block with no
-// other predecessors. It merges the blocks into a single block.
+// other predecessors or reference (for example, by OpSelectionMerge or
+// OpLoopMerge). It merges the blocks into a single block. 
 //
 // The pass is most useful after Dead Branch Elimination, which can leave
 // such sequences of blocks. Merging them makes subsequent passes more
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -185,6 +185,15 @@ Optimizer::PassToken CreateUnifyConstantPass();
 // OpSpecConstantOp.
 Optimizer::PassToken CreateEliminateDeadConstantPass();
 
+// Creates a block merge pass.
+// This pass searches for blocks with a single Branch to a block with no
+// other predecessors. It merges the blocks into a single block.
+//
+// The pass is most useful after Dead Branch Elimination, which can leave
+// such sequences of blocks. Merging them makes subsequent passes more
+// effective, such as single block local store-load elimination.
+Optimizer::PassToken CreateBlockMergePass();
+
 // Creates an inline pass.
 // An inline pass exhaustively inlines all function calls in all functions
 // designated as an entry point. The intent is to enable, albeit through
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -187,8 +187,8 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 
 // Creates a block merge pass.
 // This pass searches for blocks with a single Branch to a block with no
-// other predecessors or reference (for example, by OpSelectionMerge or
-// OpLoopMerge). It merges the blocks into a single block. 
+// other predecessors and merges the blocks into a single block. Continue
+// blocks and Merge blocks are not candidates for the second block.
 //
 // The pass is most useful after Dead Branch Elimination, which can leave
 // such sequences of blocks. Merging them makes subsequent passes more
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -192,6 +192,9 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 // The pass is most useful after Dead Branch Elimination, which can leave
 // such sequences of blocks. Merging them makes subsequent passes more
 // effective, such as single block local store-load elimination.
+//
+// While this pass reduces the number of occurrences of this sequence, at
+// this time it does not guarantee all such sequences are eliminated.
 Optimizer::PassToken CreateBlockMergePass();
 
 // Creates an inline pass.
--- a/source/opt/basic_block.h
+++ b/source/opt/basic_block.h
@@ -42,10 +42,15 @@ class BasicBlock {
 
   // Sets the enclosing function for this basic block.
   void SetParent(Function* function) { function_ = function; }
+
   // Appends an instruction to this basic block.
   inline void AddInstruction(std::unique_ptr<Instruction> i);
+
+  // Appends all of block's instructions (except label) to this block
+  inline void AddInstructions(BasicBlock* bp);
+
   // The label starting this basic block.
-  Instruction& Label() { return *label_; }
+  Instruction* GetLabelInst() { return &*label_; }
 
   // Returns the id of the label at the top of this block
   inline uint32_t id() const { return label_->result_id(); }
@@ -89,6 +94,11 @@ inline BasicBlock::BasicBlock(std::unique_ptr<Instruction> label)
 
 inline void BasicBlock::AddInstruction(std::unique_ptr<Instruction> i) {
   insts_.emplace_back(std::move(i));
+}
+
+inline void BasicBlock::AddInstructions(BasicBlock* bp) {
+  auto bEnd = end();
+  (void) bEnd.InsertBefore(&bp->insts_);
 }
 
 inline void BasicBlock::ForEachInst(const std::function<void(Instruction*)>& f,
--- a/source/opt/basic_block.h
+++ b/source/opt/basic_block.h
@@ -64,6 +64,11 @@ class BasicBlock {
     return const_iterator(&insts_, insts_.cend());
   }
 
+  iterator tail() {
+    assert(!insts_.empty());
+    return iterator(&insts_, std::prev(insts_.end()));
+  }
+
   // Runs the given function |f| on each instruction in this basic block, and
   // optionally on the debug line instructions that might precede them.
   inline void ForEachInst(const std::function<void(Instruction*)>& f,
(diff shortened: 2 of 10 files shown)


test/opt/block_merge_test.cpp
```diff
@@ -0,0 +1,279 @@
...
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::BlockMergePass>(
+      predefs + before, predefs + after, true, true);
+}
+
+// TODO(greg-lunarg): Add tests to verify handling of these cases:
+//
```
dneto0: I'd like to see cases:
 - a block B which is the target of a single branch, but is also the merge block.  Result is no-op. Example is something like:
```
     %entry = OpLabel
     OpBranch %loop
     %loop = OpLabel
     OpLoopMerge %merge %continueTarget
     OpBranch %body
     %body = OpLabel
     OpBranch %merge   ;  this is a "break" block
     %continueTarget = OpLabel   ; finish off the loop.  Yes, this is unreachable, but valid.
     OpBranch %loop

     %merge = OpLabel   ; the algorithm would be tempted to merge this away, but should not since it has 2 uses
     OpReturn
``` 

- a loop, but without structured control flow  (i.e. no OpLoopMerge).  Result is no-op.  You'd use Kernel and Addresses capabilities.  The point is that loop header block can't be merged since it has at least two uses: the block that branches to it to start the loop, and the backedge.

source/opt/block_merge_pass.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return false;
+  --iItr;
+  return iItr->opcode() == SpvOpLoopMerge;
+}
+
+bool BlockMergePass::MergeBlocks(ir::Function* func) {
+  bool modified = false;
+  for (auto bi = func->begin(); bi != func->end(); ) {
+    // Do not merge loop header blocks, at least for now.
+    if (IsLoopHeader(&*bi)) {
```
greg-lunarg: The comments could probably be clearer here. We are looking for a sequence of two blocks where the first has only one successor and the second has only one predecessor. Here, bi is the first block. The first block can have multiple preds and still be a candidate. And I think almost by definition a loop header can only have one successor. So theoretically, this case could occur, but it wasn't a priority for me at the moment and I didn't want to deal with extra logic of moving the merge op or checking that the next block wasn't a header. This logic was easier.

Also I don't think that this logic causes a requirement for structured control flow for this pass, it just doesn't optimize a particular structured construct at this time.

greg-lunarg: If you decide this is ok, I will also add a high-level comment that it is not guaranteed that all such cases are eliminated, in case someone needed this quality in their code.

source/opt/block_merge_pass.cpp
```diff
@@ -0,0 +1,109 @@
...
+    // no other predecessors
+    auto ii = bi->end();
+    --ii;
+    ir::Instruction* br = &*ii;
+    if (br->opcode() != SpvOpBranch) {
+      ++bi;
+      continue;
+    }
+    uint32_t labId = br->GetSingleWordInOperand(0);
+    analysis::UseList* uses = def_use_mgr_->GetUses(labId);
```
greg-lunarg: Good points both.

source/opt/block_merge_pass.cpp
```diff
@@ -0,0 +1,109 @@
...
+    if (uses->size() > 1) {
+      ++bi;
+      continue;
+    }
+    // Merge blocks
+    def_use_mgr_->KillInst(br);
+    auto sbi = bi;
+    for (; sbi != func->end(); ++sbi)
```
greg-lunarg: Good idea.

source/opt/block_merge_pass.cpp
```diff
@@ -0,0 +1,109 @@
...
+  for (auto& fn : *module_) 
+    id2function_[fn.result_id()] = &fn;
+
+  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module_));
+};
+
+Pass::Status BlockMergePass::ProcessImpl() {
+  bool modified = false;
+
+  // Call Mem2Reg on all remaining functions.
```
greg-lunarg: Doh!

include/spirv-tools/optimizer.hpp
```diff
@@ -185,6 +185,15 @@ Optimizer::PassToken CreateUnifyConstantPass();
 // OpSpecConstantOp.
 Optimizer::PassToken CreateEliminateDeadConstantPass();
 
+// Creates a block merge pass.
+// This pass searches for blocks with a single Branch to a block with no
+// other predecessors. It merges the blocks into a single block.
```
greg-lunarg: Re OpPhi and OpName, these feel almost more like bugs rather than design. Neither of these seem to have a legitimate reason to stop this optimization. I will probably just add code to remove these restrictions rather than document them.

source/opt/block_merge_pass.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+    uint32_t labId = br->GetSingleWordInOperand(0);
+    analysis::UseList* uses = def_use_mgr_->GetUses(labId);
+    if (uses->size() > 1) {
+      ++bi;
+      continue;
+    }
+    // Merge blocks
+    def_use_mgr_->KillInst(br);
+    auto sbi = bi;
```
atgoo: It it guaranteed that sbi comes after bi?

source/opt/block_merge_pass.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "iterator.h"
+
+static const int kSpvEntryPointFunctionId = 1;
+
+namespace spvtools {
+namespace opt {
+
+bool BlockMergePass::IsLoopHeader(ir::BasicBlock* block_ptr) {
+  auto iItr = block_ptr->end();
+  --iItr;
```
greg-lunarg: True, however I believe that valid SPIR-V will always guarantee that every block has at least one instruction, the terminating instruction. I am assuming that only valid SPIR-V comes into a pass, and only valid SPIR-V can leave a pass.

source/opt/block_merge_pass.cpp
greg-lunarg: From the spec: The order of blocks in a function must satisfy the rule that blocks appear before all blocks they dominate. bi is sbi's only predecessor, therefore bi dominates sbi.

greg-lunarg: I am adding a comment to document this logic since it is a little surprising.

source/opt/basic_block.h
```diff
@@ -42,10 +42,15 @@ class BasicBlock {
...
+
   // Appends an instruction to this basic block.
   inline void AddInstruction(std::unique_ptr<Instruction> i);
+
+  // Appends all of block's instructions (except label) to this block
+  inline void AddInstructions(BasicBlock* bp);
+
   // The label starting this basic block.
-  Instruction& Label() { return *label_; }
+  Instruction* GetLabelInst() { return &*label_; }
```
greg-lunarg: OK

test/opt/block_merge_test.cpp
greg-lunarg: First test added. Also added a single block before the continue block to show that continue block is not merged as a second block either.

As for the second test request, as I clarified above, the multiple edges entering a loop header block are not sufficient to prevent consideration of a loop header block as the first block of a mergeable block pair. If the OpLoopMerge statement is removed, the algorithm would correctly consider merging the loop header and the following block if the second block only had one predecessor.

greg-lunarg:
Didn't notice the seg faults when I ran the tests before checkin :( Fixed that.

Thinking a little more, I also decided to bail on OpPhis afterall, at least for the time being, leaving it as a TODO. Happily, the only time OpPhi's become an issue, it is when they reference the second block, but if they reference the second block, it then has a ref count > 1 and the merge is inhibited. I think I documented all this sufficiently.

source/opt/block_merge_pass.cpp
atgoo: Maybe add
iterator tail() {
 assert(!insts_.empty());
 return iterator(&insts_, std::prev(insts_.end()));
}
to BasicBlock

greg-lunarg: OK

greg-lunarg:
Also fixed multiple entry point / modified issue.

source/opt/block_merge_pass.cpp
greg-lunarg: Comment added.

source/opt/basic_block.h
```diff
@@ -59,6 +64,11 @@ class BasicBlock {
     return const_iterator(&insts_, insts_.cend());
   }
 
+  iterator tail() {
```
atgoo: Consider also
const_iterator ctail() {
  assert(!insts_.empty());
  return const_iterator(&insts_, std::prev(insts_.cend()));
}

source/opt/basic_block.h
```diff
@@ -64,6 +64,11 @@ class BasicBlock {
...
```
dneto0: I prefer "back()" as more idiomatic. That's what standard containers have (vector, queue, deque, array, string, list...)

atgoo: I didn't want to call it back(), because back() returns reference, not iterator.

greg-lunarg:
I think I have at least addressed all issues. Please let me know if any are unsatisfactory.

source/opt/block_merge_pass.cpp
```diff
@@ -0,0 +1,139 @@
...
+bool BlockMergePass::IsLoopHeader(ir::BasicBlock* block_ptr) {
+  auto iItr = block_ptr->tail();
+  if (iItr == block_ptr->begin())
+    return false;
+  --iItr;
+  return iItr->opcode() == SpvOpLoopMerge;
+}
+
+bool BlockMergePass::HasMultipleRefs(uint32_t labId) {
+  analysis::UseList* uses = def_use_mgr_->GetUses(labId);
```
greg-lunarg: OK

source/opt/block_merge_pass.cpp
```diff
@@ -0,0 +1,139 @@
...
+  int rcnt = 0;
+  for (auto u : *uses) {
+    // Don't count OpName
+    if (u.inst->opcode() == SpvOpName)
+      continue;
+    ++rcnt;
+  }
+  return rcnt > 1;
```
greg-lunarg: OK

source/opt/block_merge_pass.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+void BlockMergePass::KillInstAndName(ir::Instruction* inst) {
+  uint32_t id = inst->result_id();
+  if (id != 0) {
+    analysis::UseList* uses = def_use_mgr_->GetUses(id);
+    if (uses != nullptr) {
+      ir::Instruction* ni = nullptr;
+      for (auto u : *uses)
+        if(u.inst->opcode() == SpvOpName)
```
greg-lunarg: OK

dneto0:
Rebased, squashed, and pushed into master as ad1d0351a0ba3

PR #677, opened 2017-06-14
Add AggressiveDCEPass

greg-lunarg:
This is another pass in a series designed to reduce SPIR-V file size.

This pass eliminates unused code from functions. In addition,
it detects and eliminates code which may have spurious uses but which do
not contribute to the output of the function. The most common cause of
such code sequences is summations in loops whose result is no longer used
due to dead code elimination. This optimization has additional compile
time cost over standard dead code elimination.

This pass only processes entry point functions. It also only processes
shaders with logical addressing. It currently will not process functions
with function calls.

This pass will be made more effective by first running passes that remove
dead control flow and inlines function calls.

This pass can be especially useful after running Local Access Chain
Conversion, which tends to cause cycles of dead code to be left after
Store/Load elimination passes are completed. These cycles cannot be
eliminated with standard dead code elimination.

Note: Similar to the LLVM ADCE pass, this pass currently does not remove useless control structures. The logic for these (emptied) structures can be a source of not-insignificant code sequences. This work is TODO.


source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -0,0 +1,210 @@
...
+    else if (op != SpvOpLoad)
+      AddStores(u.inst->result_id());
+  }
+}
+
+bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
+  bool modified = false;
+  // Add non-local stores, block terminating and merge instructions
+  // to worklist. If function call encountered, return false, unmodified.
+  // TODO(): Improve in presence of function calls
```
atgoo: Consider adding your username or email address to TODO.

source/opt/basic_block.h
```diff
@@ -45,7 +45,7 @@ class BasicBlock {
   // Appends an instruction to this basic block.
   inline void AddInstruction(std::unique_ptr<Instruction> i);
   // The label starting this basic block.
-  Instruction& Label() { return *label_; }
+  Instruction* GetLabelInst() { return &*label_; }
```
dneto0: I can't find any uses of this method.   Can we eliminate this altogether?
(Also, why would returning the pointer be preferable? Since it's always non-null isn't it better to return a reference?)

source/opt/aggressive_dead_code_elim_pass.h
```diff
@@ -0,0 +1,103 @@
...
+  // Add all store instruction which use |ptrId|, directly or indirectly,
+  // to the live instruction worklist.
+  void AddStores(uint32_t ptrId);
+
+  // Return true if variable with |varId| is function scope
+  bool IsLocalVar(uint32_t varId);
+
+  // For function |func|, mark all Stores to non-function-scope variables
+  // and block terminating instructions as live. Recursively mark the values
+  // they use. When complete, delete any non-live instructions. Return true
```
dneto0: This is too aggressive, e.g. potentially sweeping away:
- synchronization instructions like OpMemoryBarrier and OpControlBarrier
- extended instructions.   Some, like Frexp, can modify storage. 
- unknown extended instructions.
- collective functions like SubgroupBallotKHR
- atomic  instructions

Perhaps, ugly though it may be, you should use a whitelist of known "safe" instructions.  An instruction is safe if it is a combinator: its effect is purely determined by its arguments, and has no side effects.   It seems the kind of list that should be centralized. (In the long run, maybe it's a fundamental property that ought to be captured in the SPIR-V grammar files.)

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -0,0 +1,210 @@
...
+  if (!module_->HasCapability(SpvCapabilityShader))
+    return Status::SuccessWithoutChange;
+
+  // Current functionality assumes logical addressing only
+  // TODO(): Handle non-logical addressing
+  if (module_->HasCapability(SpvCapabilityAddresses))
+    return Status::SuccessWithoutChange;
+
+  bool modified = false;
+  // Call Mem2Reg on all remaining functions.
```
dneto0: Out of date comment?

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -0,0 +1,210 @@
...
+  for (auto& e : module_->entry_points()) {
+    ir::Function* fn =
+        id2function_[e.GetSingleWordOperand(kSpvEntryPointFunctionId)];
+    modified = modified || AggressiveDCE(fn);
```
dneto0: AggressiveDCE(fn) || modified

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -0,0 +1,210 @@
...
+void AggressiveDCEPass::AddStores(uint32_t ptrId) {
+  analysis::UseList* uses = def_use_mgr_->GetUses(ptrId);
+  if (uses == nullptr)
+    return;
+  for (auto u : *uses) {
+    SpvOp op = u.inst->opcode();
+    if (op == SpvOpStore)
+      worklist_.push(u.inst);
+    else if (op != SpvOpLoad)
+      AddStores(u.inst->result_id());
```
dneto0: I wonder if we could get into an infinite loop.
For example, with SPV_KHR_variable_pointers, a pointer can be the result of an OpPhi.
So if you stick that in a loop, then you can get a cycle where a pointer could trace back to itself.

E.g.


```
%entry = OpLabel
OpBranch %loop

%loop =OpLabel
%ptr = OpPhi %ptr_type %ptr %loop  %loop %base_ptr %entry
...
OpBranchConditional %loop
```

Suggest fixing this with a container of visited ptrIds.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -0,0 +1,210 @@
...
+namespace opt {
+
+bool AggressiveDCEPass::IsNonPtrAccessChain(const SpvOp opcode) const {
+  return opcode == SpvOpAccessChain || opcode == SpvOpInBoundsAccessChain;
+}
+
+ir::Instruction* AggressiveDCEPass::GetPtr(
+      ir::Instruction* ip, uint32_t* varId) {
+  *varId = ip->GetSingleWordInOperand(
+      ip->opcode() == SpvOpStore ?  kSpvStorePtrId : kSpvLoadPtrId);
```
greg-lunarg: OK. But keeping k* at global scope to avoid (potential) duplication.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -0,0 +1,210 @@
...
+    for (auto& inst : blk) {
+      if (live_insts_.find(&inst) != live_insts_.end())
+        continue;
+      dead_insts_.insert(&inst);
+    }
+  }
+  // Remove debug statements referencing dead instructions. This must
+  // be done before killing the instructions, otherwise there are dead
+  // objects in the def/use database.
+  for (auto& di : module_->debugs()) {
```
dneto0: Also need to be concerned about  NoContraction and RelaxedPrecision decorations on arithmetic operations. Those are real cases but you should be able to handle any kind of decoration since the list of possible decorations may grow over time.

Sadly, decorations can get pretty complicated: either be applied directly via OpDecorate or indirectly via OpGroupDecorate.

Need a test for these other scenarios.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -0,0 +1,210 @@
...
+  return varTypeInst->GetSingleWordInOperand(kSpvTypePointerStorageClass) ==
+      SpvStorageClassFunction;
+}
+
+void AggressiveDCEPass::AddStores(uint32_t ptrId) {
+  analysis::UseList* uses = def_use_mgr_->GetUses(ptrId);
+  if (uses == nullptr)
+    return;
+  for (auto u : *uses) {
+    SpvOp op = u.inst->opcode();
```
greg-lunarg: OK

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: OK

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Yep

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Yep

source/opt/basic_block.h
greg-lunarg: This was part of an aborted attempt to eliminate unused control flow using instruction pointers. I will eliminate for now.

source/opt/aggressive_dead_code_elim_pass.h
greg-lunarg: Good catch! And good suggestions!

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Again, good catch!

source/opt/aggressive_dead_code_elim_pass.h
greg-lunarg: Are there header files for the various vendor extensions eg  SPV_AMD_shader_explicit_vertex_parameter?

dneto0: Support for 4 more AMD extensions was just added today.   The raw source of truth is the json files in source/exinst*json.  But there is no corresponding .h file with the tokens as enums.

But you can make assembly-based tests. See examples toward end of test/text_to_binary.extension_test.cpp around line 290.

greg-lunarg: I have added logic to mark as live anything that is not a combinator. For now, my list of combinators come from the list of valid operators for shaders with logical addressing.

I still need to add tests for this new logic.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: This reveals a larger issue, that we need to explicitly white list approved extensions for this (and probably every) optimization and bail if an extension is present which is not on  the approved list.

For example, say that an extension is added that makes all function scope variables volatile. That would break this and several other of my optimizations.

I have added logic which, for the moment, prohibits all extensions with a TODO to add support at a later date.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: I added logic to delete OpDecorate and OpDecorateId if their target id is dead. But I just added OpGroupDecorate to the live worklist (so that none of its ids will go dead) and added a TODO to deal with it nicely later.

Still need to add tests.

greg-lunarg:
Added tests for non-combinator instruction (eg Frexp) not eliminated, extended combinator instructions eliminated, decoration for dead instruction eliminated.

Also added logic to bail on presence of OpCopyMemory. Added TODO to handle it.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -0,0 +1,504 @@
...
+    SpvOpImageSampleDrefExplicitLod,
+    SpvOpImageSampleProjImplicitLod,
+    SpvOpImageSampleProjExplicitLod,
+    SpvOpImageSampleProjDrefImplicitLod,
+    SpvOpImageSampleProjDrefExplicitLod,
+    SpvOpImageFetch,
+    SpvOpImageGather,
+    SpvOpImageDrefGather,
+    SpvOpImageRead,
+    SpvOpImage,
```
dneto0: Could add instructions enabled by the ImageQuerycapability.
Optional: Add TODO mentioning these.

dneto0:
Squashed and pushed into master as 9de4e698
Minor code review feedback applied in 06d4fd52c2

PR #678, opened 2017-06-16
Added log(n) move-to-front implementation

atgoo:
The implementation is based on AVL and order statistic tree.

Intended to be used by SPIR-V compression algorithms.


dneto0:
Fails MSVC build.

source/util/move_to_front.h
```diff
@@ -278,9 +278,10 @@ class MoveToFront {
   // 10H3S4----5H1S1-----D2
   //           15H2S2----12H1S1----D3
   // Right links are horisontal, left links step down one line.
```
klemun: typo: horisontal -> horizontal

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+  size_t RankFromId(uint32_t id);
+
+  // Returns id corresponding to a 0-indexed rank in the move-to-front sequence
+  // and moves id to the front. Example:
+  // Before the call: 4 8 2 1 7
+  // IdFromRank(1) returns 8
+  // After the call: 8 4 2 1 7
+  uint32_t IdFromRank(size_t rank);
+
+  // Permanently removes the id from the move-to-front sequence.
```
dneto0: The method also generates an Id if rank is given as the current size.  (Is that a case we care about?  It seems weird to have this.)

source/util/move_to_front.cpp
```diff
@@ -0,0 +1,336 @@
...
+  assert(old_size == GetSize());
+
+  return rank;
+}
+
+uint32_t MoveToFront::IdFromRank(size_t rank) {
+  const size_t old_size = GetSize();
+  if (rank >= old_size) {
+    assert(rank == old_size);
+    const uint32_t new_id = static_cast<uint32_t>(rank + 1);
```
dneto0: I think new_id could conflict with an Id already in the sequence.  
But I thought the Node.id field is supposed to be unique.

Do we even care about this case of adding a new node to the sequence via IdFromRank?

source/util/move_to_front.cpp
```diff
@@ -0,0 +1,336 @@
...
+  PrintTreeInternal(out, RightOf(node), depth + 1, print_timestamp);
+
+  if (LeftOf(node)) {
+    out << std::string(depth * text_field_width, ' ');
+    PrintTreeInternal(out, LeftOf(node), depth + 1, print_timestamp);
+  }
+}
+
+void MoveToFront::InsertNode(uint16_t node) {
+  if (!root_) {
```
dneto0: assert(TimestampOf(node))

That checks whether time has rolled over.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+
+  // Removes the value from the internal tree data structure. For testing only.
+  void TestRemove(uint32_t val) {
+    const auto it = id_to_node_.find(val);
+    assert (it != id_to_node_.end());
+    RemoveNode(it->second);
+  }
+
+  // Prints the internal tree data structure to |out|. For testing only.
+  void PrintTree(std::ostream& out, bool print_timestamp = false) const {
```
dneto0: Aw.  style guide says references must be const.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+      handle = RightOf(handle);
+    }
+    return parent;
+  }
+
+  // Prints the internal tree data structure for debug purposes in the following
+  // format:
+  // 10H3S4----5H1S1-----D2
+  //           15H2S2----12H1S1----D3
+  // Right links are horisontal, left links step down one line.
```
dneto0: "horizontal"

test/move_to_front_test.cpp
```diff
@@ -0,0 +1,551 @@
...
+
+void CheckTree(const MoveToFront& mft, const std::string& expected,
+               bool print_timestamp = false) {
+  std::stringstream ss;
+  mft.PrintTree(ss, print_timestamp);
+  EXPECT_EQ(expected, ss.str());
+}
+
+TEST(MoveToFront, InsertLeftRotation) {
+  MoveToFront mft;
```
dneto0: "mtf"

test/move_to_front_test.cpp
```diff
@@ -0,0 +1,551 @@
...
+using spvutils::MoveToFront;
+
+void CheckTree(const MoveToFront& mft, const std::string& expected,
+               bool print_timestamp = false) {
+  std::stringstream ss;
+  mft.PrintTree(ss, print_timestamp);
+  EXPECT_EQ(expected, ss.str());
+}
+
+TEST(MoveToFront, InsertLeftRotation) {
```
dneto0: Add test for empty tree: Should have empty string result.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+  struct Node {
+    // Fields are in order from biggest bit width to smallest.
+    // SPIR-V id.
+    uint32_t id = 0;
+    // Timestamp from a logical clock which updates every time the element is
+    // accessed.
+    uint32_t timestamp = 0;
+    // The size of the node's subtee, including the node.
+    // SizeOf(LeftOf(node)) + SizeOf(RightOf(node)) + 1.
+    uint16_t size = 0;
```
atgoo: No, 512 kb of memory at most.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+    // Handles to connected nodes.
+    uint16_t left = 0;
+    uint16_t right = 0;
+    uint16_t parent = 0;
+    // Distance to the farthest leaf.
```
atgoo: There is only one NIL. Leaves are NIL (one and the same), root parent also.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+    // Leaves have height 0, real nodes at least 1.
+    uint8_t height = 0;
```
atgoo: We should run into hardware problems long before we reach tree height 255. But you have a point, I'm not saving anything at all, because after that field would be a gap, and after the gap the next struct.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+  };
+
+  // Creates node and sets correct values. Nodes should be created only through
+  // this function.
```
atgoo: Ok. That node is not really a node.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+  // root parent reference a singleton under handle 0. Although dereferencing
+  // a null pointer is not possible, inappropriate access to handle 0 would
+  // cause an assertion. Handles are not garbage collected if id is deprecated.
+  // But handles are recycled when a node is repositioned.
+
+  // Node accessor methods. Naming is designed to be similar to natural
+  // language as these functions tend to be used in sequences, for example:
+  // ParentOf(LeftestDescendentOf(RightOf(node)))
+
+  // Returns immutable id of the node referenced by |handle|.
```
atgoo: Yes, I can drop immutable in the comment.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+  inline uint32_t IdOf(uint16_t handle) const {
```
atgoo: I have thought it's up to the compiler to decide how to interpret the hint. Does inline then even have any use except for "it's ok for the out-of-class definition of this function to appear in multiple modules"?

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+  // 10H3S4----5H1S1-----D2
+  //           15H2S2----12H1S1----D3
+  // Right links are horisontal, left links step down one line.
+  // 5H1S1 is read as id 5, height 1, size 1. Optionally node label can also
+  // contain timestamp (5H1S1T15). D3 stands for depth 3.
+  void PrintTreeInternal(std::ostream& out, uint16_t node, size_t depth,
+                         bool print_timestamp) const;
+
+  // Inserts node handle in the tree.
+  void InsertNode(uint16_t node);
```
atgoo: I probably should replace "handle" with "node" in the accessor functions for consistency.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+  // Rotates |node| left, reasigns all connections and returns the node
+  // which takes place of the |node|.
+  uint16_t RotateLeft(const uint16_t node);
+
+  // Rotates |node| right, reasigns all connections and returns the node
+  // which takes place of the |node|.
+  uint16_t RotateRight(const uint16_t node);
+
+  uint16_t root_ = 0;
+  uint32_t next_timestamp_ = 1;
```
atgoo: 0 is timestamp of NIL.

source/util/move_to_front.h
dneto0: C++11 7.1.2:
para 2:
A function declaration (8.3.5, 9.3, 11.3) with an inline specifier declares an inline function. The inline
specifier indicates to the implementation that inline substitution of the function body at the point of call
is to be preferred to the usual function call mechanism. An implementation is not required to perform this
inline substitution at the point of call; however, even if this inline substitution is omitted, the other rules
for inline functions defined by 7.1.2 shall still be respected.

para 3: A function defined within a class definition is an inline function. 

----
I see I read a little too much into this.   A function defined within a class definition is an inline function, but only has the inline *specifier* if you add it.  And para2 says it's the inline *specifier* that indicates prefer-to-inline.

That being said, I generally shy away from directing the compiler to do optimization stuff that I haven't really proven with tuning data.   Usually, programmers are worse at blind optimization than they think.  For example, there could be a lot of code hidden behind the std::vector::at method (what with bounds checking, exception throwing...).  And when you nest these accessors all together it's not 100% clear what the right thing to do is.  So I personally would leave the inline off and let the compiler make an educated guess based on code size. 

You can keep this if you like, but I personally would remove it.  It does have a read-ability cost for having the extra word.

source/util/move_to_front.cpp
```diff
@@ -0,0 +1,336 @@
...
+    root_ = node;
+    return;
+  }
+
+  uint16_t iter = root_;
+  uint16_t parent = 0;
+
+  bool right_child;
```
klemun: @atgoo typo in the new comment: of -> or

source/util/move_to_front.h
```diff
@@ -103,172 +110,172 @@ class MoveToFront {
...
+  uint32_t& MutableHeightOf(uint32_t node) {
+    assert(node);
+    return nodes_.at(node).height;
   }
 
-  // Returns true iff |handle| is left child of its parent.
-  inline bool IsLeftChild(uint16_t handle) const {
-    assert(handle);
-    return LeftOf(ParentOf(handle)) == handle;
+  // Returns true iff |node| is left child of its parent.
```
klemun: typo: iff -> if  (unless this is intentional - the next 2 comments have this as well)

atgoo: This is intentional, iff in this context means if and only if.
https://en.wikipedia.org/wiki/If_and_only_if

source/util/move_to_front.cpp
atgoo: Done.

source/util/move_to_front.h
klemun: Oh, I see. Thanks for explaining.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+#include <cstdint>
+#include <ostream>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+namespace spvutils {
+
+// Log(n) move-to-front implementation. Implements two main functions:
+// IdFromRank - access id by its 0-indexed rank in the move-to-front sequence.
```
dneto0: Say this in the comment, not the review.

source/util/move_to_front.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+namespace spvutils {
+
+size_t MoveToFront::RankFromId(uint32_t id) {
+  const size_t old_size = GetSize();
+  (void)old_size;
+  const auto it = id_to_node_.find(id);
+
+  if (it == id_to_node_.end()) {
+    assert(id == next_id_);
```
dneto0: I think this assertion is weird.  It's a strange coupling between this MTF sequence and the client.

I think you've written that the expected operation is for the client to call IdForRank(0) to generate new Ids.   But since you have a simple id allocation scheme, this forces RankFromId to be restrictive in this way.

source/util/move_to_front.h
```diff
@@ -0,0 +1,325 @@
...
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+namespace spvutils {
+
+// Log(n) move-to-front implementation. Implements two main functions:
+// IdFromRank - access id by its 1-indexed rank in the move-to-front sequence.
+// RankFromId - get the rank of the given id in the move-to-front sequence.
+// Accessing an id with any of the two functions moves the id to the front
```
dneto0: .... But in this implementation ids are not unrestricted.  You *require* a new Id to be next_id_.

That still seems a weird choice.  How do I know what the next id is?

The current setup seems strangely coupled to its client, and not very reusable.  I'd rather have arbitrary Ids allocated by the client, and only one way to inject a new Id into the MTF sequence.  I'd be happy with an explicit insertion.

source/util/move_to_front.h
```diff
@@ -0,0 +1,325 @@
...
+// of the sequence (rank of 1).
+//
+// The implementation is based on an AVL-based order statistic tree.
+//
```
dneto0: Feedback unaddressed from before:  Should describe the ordering by timestamp; larger timestamps to the left.

source/util/move_to_front.h
```diff
@@ -0,0 +1,325 @@
...
+  // with DeprecateId(). But handles are recycled when a node is repositioned.
+
+  // Internal tree data structure node.
+  struct Node {
+    // Fields are in order from biggest bit width to smallest.
+    // SPIR-V id.
+    uint32_t id = 0;
+    // Timestamp from a logical clock which updates every time the element is
+    // accessed through IdFromRank or RankFromId.
+    uint32_t timestamp = 0;
```
dneto0: Feedback unaddressed from before.  What happens when timestamp rolls over?
I'd be happier with this as a 64-bit type.

source/util/move_to_front.h
```diff
@@ -0,0 +1,318 @@
...
+
+ private:
+  // Internal tree data structure node.
+  struct Node {
+    // Fields are in order from biggest bit width to smallest.
+    // SPIR-V id.
+    uint32_t id = 0;
+    // Timestamp from a logical clock which updates every time the element is
+    // accessed.
+    uint32_t timestamp = 0;
```
dneto0: Which makes this a not very reusable component.

But most importantly, put this comment in the code, not the review.  The review will be lost in time.

source/util/move_to_front.h
dneto0: So it's confusing to say "Leaves" here without this extra comment because it seems like there are many leaf nodes.  But there's only one leaf node.

Again, put these comments in the code, not the review.

source/util/move_to_front.cpp
```diff
@@ -0,0 +1,336 @@
...
+  if (it == id_to_node_.end()) {
+    InsertNode(CreateNode(next_timestamp_++, id));
+    assert(old_size + 1 == GetSize());
+    return old_size;
+  }
+
+  uint16_t target = it->second;
+
+  uint16_t node = target;
+  size_t rank = SizeOf(LeftOf(node));
```
dneto0: I don't understand.  uint16_t is a valid type, implemented by the compiler.

And the code remains unchanged, i.e. no static_cast<uint16>(...).    In any case, this is a minor point.

source/util/move_to_front.cpp
```diff
@@ -0,0 +1,336 @@
...
+    parent = ParentOf(parent);
+  }
+}
+
+uint16_t MoveToFront::RemoveNode(uint16_t node) {
+  // Instead of removing the |node| find another node which is easier to
+  // remove and swap them.
+  if (const uint16_t scapegoat = RightestDescendantOf(LeftOf(node))) {
+    std::swap(MutableIdOf(node), MutableIdOf(scapegoat));
+    std::swap(MutableTimestampOf(node), MutableTimestampOf(scapegoat));
```
dneto0: When I originally read the code I missed the LeftOf part.  So makes sense that the scapegoat can take the place of the removed node.

But I think symmetry argues that the leftest-of-the-right-child would be equally acceptable.   Minor point.

source/util/move_to_front.cpp
```diff
@@ -0,0 +1,336 @@
...
+    id_to_node_[IdOf(node)] = node;
+    id_to_node_[IdOf(scapegoat)] = scapegoat;
+    node = scapegoat;
+  }
+
+  // node may have only one child at this point.
+  assert(!RightOf(node) || !LeftOf(node));
+
```
dneto0: ok

atgoo:
To make it reusable I would have to move the whole thing into the header and template the id.

dneto0:
The reusability I'm looking for is for the *client* to choose the id values, not the MTF container.
I don't need templating over the Id type (or value type more generically).

atgoo:
> The reusability I'm looking for is for the client to choose the id values, not the MTF container.
I don't need templating over the Id type (or value type more generically).

What's the point of it being a bit more generic, but not completely generic? The way it is now, the header can be taken and reused elsewhere.

dneto0:
The current API is difficult to use correctly, and unnecessarily constrains the client into using the IDs that the MTF sequence chooses.

To use the API correctly, while adding a new Id with RankForId, the Id has to be exactly what next_id_ is, but there's no way to get that.  Which means the client has to track the same number that the MTF object has internally.  That's error prone.

Also, it constrains the solution space for the codec client.

With slight tweaks the API will be easier to use and much less error prone.  I'm reminded of Scott Meyer's advice to API designers: 
http://www.aristeia.com/Papers/IEEE_Software_JulAug_2004_revised.htm 
It comes out more empahtically in https://www.youtube.com/watch?v=5tg1ONG18H8

dneto0:
https://www.youtube.com/watch?v=5tg1ONG18H8#t=4m08s and the next two minutes
"If they do not use the interface correctly, then what that means is it's your fault.  It's not their fault.  You cannot blame the user. You cannot blame the caller. They're trying to do the right thing.  If they're failing to do the right thing, there is something wrong with the interface that's permitting them to make an error that should not be permissible."

"Software should be easy to use correctly, and hard to use incorrectly."

atgoo:
> To use the API correctly, while adding a new Id with RankForId, the Id has to be exactly what next_id_ is, but there's no way to get that.

That has been fixed.

source/util/move_to_front.h
```diff
@@ -0,0 +1,649 @@
...
+  std::vector<Node> nodes_;
+
+  // Maps ids to node handles.
+  std::unordered_map<Val, uint32_t> value_to_node_;
+};
+
+template <typename Val>
+bool MoveToFront<Val>::Insert(const Val& value) {
+  auto it = value_to_node_.find(value);
+  if (it != value_to_node_.end() && IsInTree(it->second))
```
dneto0: *Might* be worth a comment to say that reinsertion of a node that had been removed will reuse its old node.  That would explain the second clause here.

source/util/move_to_front.h
```diff
@@ -0,0 +1,649 @@
...
+  // Inserts value in the move-to-front sequence. Does nothing if the value is
+  // already in the sequence. Returns true if insertion was successful.
+  // The inserted value is placed at the front of the sequence (rank 1).
+  bool Insert(const Val& value);
+
+  // Removes value from move-to-front sequence. Returns false iff the value
+  // was not found.
+  bool Remove(const Val& value);
+
+  // Computes 1-indexed rank of value in the move-to-front sequence and moves
```
dneto0: The RankFromValue and ValueFromRank APIs work.

But I bet it would be nicer to use if they return size_t and const Value& respectively by imposing the preconditions that the rank be valid and the value be already in the set.

No big deal. This is acceptable.

dneto0:
Rebased, squashed, and pushed into master as 8d3882a40

PR #680, opened 2017-06-20
Transform multiple entry points

dneto0:
Don't stop just after one because of short-circuiting logical-or.


dneto0:
cc: @greg-lunarg 

And yes, I didn't write the tests for this.  :-)

PR #681, opened 2017-06-21
Add support for SPV AMD extensions

amdrexu:


dneto0:
The code looked ok to me, but the bots are failing.

dneto0:
I'm looking at this.  Seems like a missing dependency on the way to ${spirv-tools_BINARY_DIR}/enum_string_mapping.inc.

dneto0:
It turns out there are missing dependencies, but that's not the problem here.

The problem here is that the enum/string mappings for vendor and other non-standard extensions are generated from a table inside utils/generate_grammar_tables.py.
The following diff fixes the issue:
```
--- a/utils/generate_grammar_tables.py
+++ b/utils/generate_grammar_tables.py
@@ -27,7 +27,16 @@ PYGEN_VARIABLE_PREFIX = 'pygen_variable'
 
 # Extensions to recognize, but which don't come from the SPIRV-V core grammar.
 NONSTANDARD_EXTENSIONS = [
+    # TODO(dneto): Vendor extension names should really be derived from the
+    # content of .json files.
     'SPV_AMD_gcn_shader',
+    'SPV_AMD_shader_ballot',
+    'SPV_AMD_shader_explicit_vertex_parameter',
+    'SPV_AMD_shader_trinary_minmax',
+    # The following don't have an extended instruction set grammar file.
+    'SPV_AMD_gpu_shader_half_float',
+    'SPV_AMD_texture_gather_bias_lod',
+    'SPV_AMD_gpu_shader_int16',
     # Validator would ignore type declaration unique check. Should only be used
     # for legacy autogenerated test files containing multiple instances of the
     # same type declaration, if fixing the test by other methods is too

```

dneto0:
I'll make a new PR to combine my fix.  Then I'll let the bots verify.  If all is good, I'll approve that PR and close this.

amdrexu:
@dneto0 Thanks very much for your kind help. I do forget to change this file, which leads to the test failures. I think you patch is fine. So please merge it with mine and integrate them into master branch.

dneto0:
Merged via https://github.com/KhronosGroup/SPIRV-Tools/pull/682

PR #686, opened 2017-06-27
Travis CI: stop requiring sudo and use make instead of ninja

antiagainst:
This gives us shorter testing time since we can use the container
based testing environment, which starts faster. Also using make
means we don't need to install ninja separately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ os:
 
 # Use Ubuntu 14.04 LTS (Trusty) as the Linux testing environment.
 dist: trusty
-sudo: required
+sudo: false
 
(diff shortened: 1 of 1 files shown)


antiagainst:
Testing out for now; no need to review.

antiagainst:
Ready for review now.

.travis.yml
```diff
@@ -35,18 +35,14 @@ matrix:
 cache:
   apt: true
 
+git:
+  depth: 5
```
dneto0: Why depth 5?  Why not depth 1?

antiagainst: Was a little bit random choice. :)

PR #688, opened 2017-06-27
Various easy perf wins

chrisforbes:


dneto0:
Rebased and pushed into master, ending in d431b69c28

source/validate.cpp
```diff
@@ -326,7 +328,7 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
     spv_instruction_t inst;
     spvInstructionCopy(&binary->code[index], static_cast<SpvOp>(opcode),
                        wordCount, endian, &inst);
-    instructions.push_back(inst);
+    instructions.emplace_back(std::move(inst));
```
atgoo: std::move decreases performance.

These two should work the same:
instructions.push_back(inst);
instructions.emplace_back(inst);

but
instructions.emplace_back(std::move(inst));
has an additional memory write to inst.

source/binary.cpp
```diff
@@ -269,17 +273,18 @@ spv_result_t Parser::parseInstruction() {
...
+    _.endian_converted_words.reserve(25);
   }
 
   // After a successful parse of the instruction, the inst.operands member
   // will point to this vector's storage.
-  std::vector<spv_parsed_operand_t> operands;
   // Most instructions have fewer than 25 logical operands.
-  operands.reserve(25);
+  _.operands.clear();
+  _.operands.reserve(25);
```
atgoo: This reserve can be done in the constructor.

source/binary.cpp
```diff
@@ -269,17 +273,18 @@ spv_result_t Parser::parseInstruction() {
   // If the module's endianness is different from the host native endianness,
   // then converted_words contains the the endian-translated words in the
   // instruction.
-  std::vector<uint32_t> endian_converted_words = {first_word};
+  _.endian_converted_words.clear();
+  _.endian_converted_words.push_back(first_word);
   if (_.requires_endian_conversion) {
     // Most instructions have fewer than 25 words.
-    endian_converted_words.reserve(25);
+    _.endian_converted_words.reserve(25);
```
atgoo: This reserve can be done in the constructor.

source/validate.cpp
ratchetfreak: but the copy (without std::move) needs an extra allocation which is then followed by a deallocation in the destructor of the original.

chrisforbes: @atgoo, the issue here is the `std::vector` embedded in `inst`. I will take another look at it and undo this change if it's really losing.

source/binary.cpp
chrisforbes: Indeed it can. I will move those there, in both cases.

source/validate.cpp
atgoo: > the issue here is the std::vector embedded in inst. I will take another look at it and undo this change if it's really losing

Yes, if instruction is not POD, then what I said doesn't apply.

atgoo: I see, I confused it with spv_parsed_instruction_t which is POD.

PR #689, opened 2017-06-28
Second round of perf changes

chrisforbes:
Chops off another 25s / 15% or so (so far) off my internal workload.

This is very WIP -- there are a number of tests that assume a pattern is a std::deque (and some of them just verify that a deque works...; also various TODO dropped in the code where I can do better.)

### DIFF
--- a/source/assembly_grammar.h
+++ b/source/assembly_grammar.h
@@ -95,17 +95,19 @@ class AssemblyGrammar {
   spv_result_t lookupExtInst(spv_ext_inst_type_t type, uint32_t firstWord,
                              spv_ext_inst_desc* extInst) const;
 
-  // Inserts the operands expected after the given typed mask onto the front
+  // Inserts the operands expected after the given typed mask onto the end
   // of the given pattern.
   //
-  // Each set bit in the mask represents zero or more operand types that should
-  // be prepended onto the pattern. Operands for a less significant bit always
-  // appear before operands for a more significant bit.
+  // Each set bit in the mask represents zero or more operand types that
+  // should be appended onto the pattern. Operands for a less significant
+  // bit must always match before operands for a more significant bit, so
+  // the operands for a less significant bit must appear closer to the end
+  // of the pattern stack.
   //
   // If a set bit is unknown, then we assume it has no operands.
-  void prependOperandTypesForMask(const spv_operand_type_t type,
-                                  const uint32_t mask,
-                                  spv_operand_pattern_t* pattern) const;
+  void pushOperandTypesForMask(const spv_operand_type_t type,
+                               const uint32_t mask,
+                               spv_operand_pattern_t* pattern) const;
 
  private:
   const spv_target_env target_env_;
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -202,6 +202,7 @@ class Parser {
     // Used by parseOperand
     std::vector<spv_parsed_operand_t> operands;
     std::vector<uint32_t> endian_converted_words;
+    spv_operand_pattern_t expected_operands;
   } _;
 };
 
@@ -310,13 +311,14 @@ spv_result_t Parser::parseInstruction() {
   // has its own logical operands (such as the LocalSize operand for
   // ExecutionMode), or for extended instructions that may have their
   // own operands depending on the selected extended instruction.
-  spv_operand_pattern_t expected_operands(
-      opcode_desc->operandTypes,
-      opcode_desc->operandTypes + opcode_desc->numTypes);
+  _.expected_operands.clear();
+  _.expected_operands.reserve(opcode_desc->numTypes);
+  for (auto i = 0; i < opcode_desc->numTypes; i++)
+      _.expected_operands.push_back(opcode_desc->operandTypes[opcode_desc->numTypes - i - 1]);
 
   while (_.word_index < inst_offset + inst_word_count) {
     const uint16_t inst_word_index = uint16_t(_.word_index - inst_offset);
-    if (expected_operands.empty()) {
+    if (_.expected_operands.empty()) {
       return diagnostic() << "Invalid instruction Op" << opcode_desc->name
                           << " starting at word " << inst_offset
                           << ": expected no more operands after "
@@ -325,17 +327,17 @@ spv_result_t Parser::parseInstruction() {
                           << inst_word_count << ".";
     }
 
-    spv_operand_type_t type = spvTakeFirstMatchableOperand(&expected_operands);
+    spv_operand_type_t type = spvTakeFirstMatchableOperand(&_.expected_operands);
 
     if (auto error =
             parseOperand(inst_offset, &inst, type, &_.endian_converted_words,
-                         &_.operands, &expected_operands)) {
+                         &_.operands, &_.expected_operands)) {
       return error;
     }
   }
 
-  if (!expected_operands.empty() &&
-      !spvOperandIsOptional(expected_operands.front())) {
+  if (!_.expected_operands.empty() &&
+      !spvOperandIsOptional(_.expected_operands.back())) {
     return diagnostic() << "End of input reached while decoding Op"
                         << opcode_desc->name << " starting at word "
                         << inst_offset << ": expected more operands after "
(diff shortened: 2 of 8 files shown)


source/operand.cpp
```diff
@@ -262,24 +267,23 @@ bool spvExpandOperandSequenceOnce(spv_operand_type_t type,
                                   spv_operand_pattern_t* pattern) {
   switch (type) {
     case SPV_OPERAND_TYPE_VARIABLE_ID:
-      pattern->insert(pattern->begin(), {SPV_OPERAND_TYPE_OPTIONAL_ID, type});
+      pattern->insert(pattern->end(), {type, SPV_OPERAND_TYPE_OPTIONAL_ID});
```
atgoo: Is this not just emplace_back?

source/operand.cpp
```diff
@@ -292,25 +296,29 @@ spv_operand_type_t spvTakeFirstMatchableOperand(
...
-    alternatePattern.push_back(SPV_OPERAND_TYPE_OPTIONAL_CIV);
+  // TODO: make this not terrible again: scan for RESULT_ID first,
+  // then decide what to do. (chrisforbes) not really bothered by
+  // this being a bit slow-- is a weird case in assembler path.
+
+  for (auto it = pattern.crbegin(); it != pattern.crend(); it++) {
+      if (*it == SPV_OPERAND_TYPE_RESULT_ID) {
+          alternatePattern.insert(alternatePattern.begin(), *it);
+          alternatePattern.insert(alternatePattern.begin(), SPV_OPERAND_TYPE_OPTIONAL_CIV);
+          return alternatePattern;
```
atgoo: Consider
```
spv_operand_pattern_t alternatePattern(index + 2, SPV_OPERAND_TYPE_OPTIONAL_CIV);
alternatePattern[index + 1] = SPV_OPERAND_TYPE_RESULT_ID;
return alternatePattern;
```

atgoo: I don't know what this function is doing but it looks like we want to return
{o, x, o, ..., o}
if result_id is found, or
{o}
if not.

source/operand.cpp
chrisforbes: It can be, yes.

source/operand.cpp
chrisforbes: Will do something along those lines. I'd not really cared very much about making this function good -- doesn't appear in my profile at all -- but not difficult to do this right.

dneto0:
I haven't looked at your patch.  But I can give you an insight as to why the code is the way it is.
The parser is doing classic parse-matching: constantly rewriting the left end of the pattern, and consuming inputs and matching against symbols in those patterns.   I *think* of it as the left end of the pattern, but it can be flipped around so that input symbols match against the end of a vector instead of the start of a dequeue.  Perhaps that's exactly what you've done.  :-)

source/binary.cpp
```diff
@@ -266,25 +274,15 @@ spv_result_t Parser::parseInstruction() {
 
   const uint32_t first_word = peek();
 
-  // TODO(dneto): If it's too expensive to construct the following "words"
```
dneto0: W00t

source/binary.cpp
```diff
@@ -310,13 +308,13 @@ spv_result_t Parser::parseInstruction() {
   // has its own logical operands (such as the LocalSize operand for
   // ExecutionMode), or for extended instructions that may have their
   // own operands depending on the selected extended instruction.
-  spv_operand_pattern_t expected_operands(
-      opcode_desc->operandTypes,
-      opcode_desc->operandTypes + opcode_desc->numTypes);
+  _.expected_operands.clear();
+  for (auto i = 0; i < opcode_desc->numTypes; i++)
```
dneto0: I expect i is int and numTypes field is uint16_t and you'll get a compile error.  Recommend doing:

```
auto types_array = &(opcode_desc->operandTypes[0]);
for (unsigned i =0, numTypes = opcode_desc->numTypes ; i < numTypes ; i++)  {
 ... =  types_array[ numTypes - i - 1]
```

source/operand.cpp
dneto0: Indeed it's a true corner case for making extra-weird test cases.  Don't bother making this fast.

chrisforbes:
@dneto0 is indeed what this does :).

source/binary.cpp
atgoo: The compiler likely compares i not to numTypes, but to int(numTypes), because it doesn't know how to compare int and uint16_t. So it would compile and work correctly.

chrisforbes:
Updated:
- Fix tests that work directly with patterns, to think about them as the other way around.
- Rename 'prepend' to 'push' everywhere. I think this is reasonably clear?
- Fix text parser (was building its patterns backward, which went poorly)
- Use spv_operand_pattern_t name consistently -- we had a vector in one place before.

test/operand_pattern_test.cpp
```diff
@@ -199,20 +199,29 @@ TEST(AlternatePatternFollowingImmediate, MultipleNonResultIds) {
 TEST(AlternatePatternFollowingImmediate, ResultIdFront) {
   EXPECT_THAT(spvAlternatePatternFollowingImmediate(
                   {SPV_OPERAND_TYPE_RESULT_ID, SPV_OPERAND_TYPE_ID}),
-              Eq(spv_operand_pattern_t{SPV_OPERAND_TYPE_RESULT_ID,
+              Eq(spv_operand_pattern_t{SPV_OPERAND_TYPE_OPTIONAL_CIV,
```
atgoo: Are the input parameters not supposed to be in the reversed order?
```
EXPECT_THAT(spvAlternatePatternFollowingImmediate(
    {SPV_OPERAND_TYPE_ID, SPV_OPERAND_TYPE_RESULT_ID}),
```
I don't know if it's important, but it looks like you are changing the original tests if the input parameters are given in the same order as before, but the function expects them to be reversed.

chrisforbes: These tests are patterns in, patterns out. I chose to fix up just the results rather than reverse both input and output patterns. I believe we still have good coverage of all the places a result id may be, across this test and its two siblings, but exactly what behavior is covered by each test moves around a little. Is that OK?

atgoo: It's ok with me. It just looked confusing.

source/operand.h
```diff
@@ -27,7 +27,7 @@
 //
 // As we parse an instruction in text or binary form from left to right,
 // we pull and push from the front of the pattern.
```
dneto0: "we pop and push at the end of the pattern vector.   Symbols later in the pattern vector are matched against the input before symbols earlier in the pattern vector are matched."

Probably also worth saying why:
"Using a vector in this way reduces memory traffic, which is good for performance."

source/operand.h
```diff
@@ -65,8 +65,8 @@ bool spvOperandIsVariable(spv_operand_type_t type);
 // Inserts a list of operand types into the front of the given pattern.
```
dneto0: "Append a list of operand types to the end of the pattern vector.  "

source/operand.cpp
```diff
@@ -218,26 +219,28 @@ const char* spvOperandTypeStr(spv_operand_type_t type) {
...
 
-void spvPrependOperandTypesForMask(const spv_operand_table operandTable,
-                                   const spv_operand_type_t type,
-                                   const uint32_t mask,
-                                   spv_operand_pattern_t* pattern) {
+void spvPushOperandTypesForMask(const spv_operand_table operandTable,
+                                const spv_operand_type_t type,
+                                const uint32_t mask,
+                                spv_operand_pattern_t* pattern) {
   // Scan from highest bits to lowest bits because we will prepend in LIFO
```
dneto0: prepend -> append

test/operand_pattern_test.cpp
dneto0: Confusing at first, but this works for me.

chrisforbes:
Comments all updated as requested.

chrisforbes:
Adjusted again to fix new mark-v code.

source/assembly_grammar.h
```diff
@@ -95,17 +95,19 @@ class AssemblyGrammar {
...
 
-  // Inserts the operands expected after the given typed mask onto the front
+  // Inserts the operands expected after the given typed mask onto the end
   // of the given pattern.
   //
-  // Each set bit in the mask represents zero or more operand types that should
-  // be prepended onto the pattern. Operands for a less significant bit always
-  // appear before operands for a more significant bit.
+  // Each set bit in the mask represents zero or more operand types that
+  // should be prepended onto the pattern. Operands for a less significant
```
dneto0: prepended -> appended

chrisforbes:
Fixed :)

dneto0:
Rebased, squashed, and pushed into master as 78338d5ba

PR #690, opened 2017-06-29
Added fixed width encoding to bit_stream

atgoo:
Fixed width encoding is intended to be used for small unsigned integers
when the upper bound is known both to the encoder and the decoder
(for example move-to-front rank).


dneto0:
Pushed onto master as  e842c17e
(Whoops. I rebased when I didn't have to.)

PR #691, opened 2017-06-30
Add LocalSSAElimPass

greg-lunarg:
This pass is another in a series designed to reduce the size of SPIR-V files.

For every entry point function, eliminate all loads and stores of function
scope variables only referenced with non-access-chain loads and stores.
Eliminate the variables as well. 

Only shader modules with logical addressing are currently processed.
In addition, modules with any extensions enabled are not processed. This
is left for future work.

The presence of access chain references and function calls can inhibit
this optimization. Hence, this pass is most effective if preceeded by exhaustive inlining
and  LocalAccessChainConvert.

LocalSingleStoreElim, LocalSingleBlockElim and other code reduction passes (eg DeadBranchElimination) will reduce the work that this pass has to do.


source/opt/local_ssa_elim_pass.h
```diff
@@ -0,0 +1,254 @@
...
+
+  // Initialize SSA map for loop header block by merging SSA maps 
+  // from all predecessors. If any value ids differ for any variable
+  // across predecessors, create a phi function and use that value id
+  // for the variable in the new SSA map. Assumes all predecessors have
+  // been visited by SSARewrite except the back edge. Use a dummy value
+  // in the phi for the back edge until the back edge block is visited
+  // and patch the phi value then.
+  void SSABlockInitLoopHeader(std::list<ir::BasicBlock*>::iterator block_itr);
+
```
greg-lunarg: Yeah, the poorly named SSABlockInitSelectionMerge routine actually handles all multi-predecessor non-loop-header blocks including continue blocks. I will fix the name and comment.

source/opt/local_ssa_elim_pass.h
```diff
@@ -0,0 +1,254 @@
...
+  // Patch phis in loop header block now that the map is complete for the
+  // backedge predecessor. Specifically, for each phi, find the value
+  // corresponding to the backedge predecessor. That contains the variable id
+  // that this phi corresponds to. Change this phi operand to the the value
+  // which corresponds to that variable in the predecessor map.
+  void PatchPhis(uint32_t header_id, uint32_t back_id);
+
+  // Return true if all extensions in this module are allowed by this pass.
+  // Currently, no extensions are allowed. glsl_std_450 extended instructions
```
greg-lunarg: Yeah. I just didn't want people to think we weren't handling glsl_std_450 instructions, because we are. I will work on that line.

source/opt/local_ssa_elim_pass.h
```diff
@@ -0,0 +1,254 @@
...
+  // are allowed.
+  bool AllExtensionsAllowed() const;
```
greg-lunarg: OK

greg-lunarg:
I was iterating though an unordered set which caused a slight diff on one of the platforms. Will fix.

source/opt/local_ssa_elim_pass.h
```diff
@@ -0,0 +1,254 @@
...
+  void ComputeStructuredOrder(ir::Function* func,
+      std::list<ir::BasicBlock*>* order);
+
+  // Return true if loop header block
+  bool IsLoopHeader(ir::BasicBlock* block_ptr) const;
+
+  // Initialize SSA map for block with single predecessor
+  void SSABlockInitSinglePred(ir::BasicBlock* block_ptr);
+
+  // Return true if variable is loaded after the label
```
greg-lunarg: "After" mean in any succeeding block. Will fix comment.

source/opt/local_ssa_elim_pass.h
```diff
@@ -0,0 +1,254 @@
...
+  // order has the property that dominators come before all blocks they
+  // dominate and merge blocks come after all blocks that are in the control
+  // constructs of their header.
+  void ComputeStructuredOrder(ir::Function* func,
+      std::list<ir::BasicBlock*>* order);
+
+  // Return true if loop header block
+  bool IsLoopHeader(ir::BasicBlock* block_ptr) const;
+
+  // Initialize SSA map for block with single predecessor
```
greg-lunarg: I changed this comment (and all others) to refer to label2ssa_map_, and improved the comment there.

include/spirv-tools/optimizer.hpp
```diff
@@ -230,6 +230,23 @@ Optimizer::PassToken CreateInlinePass();
...
+// the above optimization.
+//
+// Only shader modules with logical addressing are currently processed.
+// Currently modules with any extensions enabled are not processed. This
+// is left for future work.
+//
+// This pass is most effective if preceeded by Inlining and 
+// LocalAccessChainConvert. LocalSingleStoreElim and LocalSingleBlockElim
+// will reduce the work that this pass has to do.
+Optimizer::PassToken CreateLocalSSAElimPass();
```
greg-lunarg: My thinking here was to utilize a similar naming mechanism for the other previous similar phases: --eliminate-local-<method>, where method could be single-store or single-block. How about --eliminate-local-multiple-store?

dneto0: I see how --eliminate-local-multiple-store echoes --eliminate-local-single-store.  Seems ok to me, as long as the user help text is accurate.

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+    }
+    // If val is the same for all predecessors, enter it in map
+    if (!needsPhi) {
+      label2ssa_map_[label].insert(var_val);
+      continue;
+    }
+    // Val differs across predecessors. Add phi op to block and 
+    // add its result id to the map. For live back edge predecessor,
+    // use the variable id. We will patch this after visiting back
+    // edge predecessor. For predessors that do not define a value,
```
dneto0: "predecessors"

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+
+uint32_t LocalSSAElimPass::GetPointeeTypeId(
+    const ir::Instruction* ptrInst) const {
+  const uint32_t ptrTypeId = ptrInst->type_id();
+  const ir::Instruction* ptrTypeInst = def_use_mgr_->GetDef(ptrTypeId);
+  return ptrTypeInst->GetSingleWordInOperand(kTypePointerTypeIdInIdx);
+}
+
+void LocalSSAElimPass::SSABlockInitLoopHeader(
+    std::list<ir::BasicBlock*>::iterator block_itr) {
```
dneto0: This is a tricky algorithm.  Your comments inside this function are spot on!

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+  auto insertItr = block_ptr->begin();
+  for (auto var_val : liveVars) {
+    const uint32_t varId = var_val.first;
+    if (!IsLiveAfter(varId, label))
+      continue;
+    const uint32_t val0Id = var_val.second;
+    bool differs = false;
+    for (uint32_t predLabel : label2preds_[label]) {
+      const auto var_val_itr = label2ssa_map_[predLabel].find(varId);
+      // Missing values cause a difference
```
dneto0: Worth saying  "... because we'll need to use an undef for that predecessor".

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+      continue;
+    const uint32_t val0Id = var_val.second;
+    bool needsPhi = false;
+    if (val0Id != 0) {
+      for (uint32_t predLabel : label2preds_[label]) {
+        // Skip back edge predecessor.
+        if (predLabel == backLabel)
+          continue;
+        const auto var_val_itr = label2ssa_map_[predLabel].find(varId);
+        // Missing values do not cause difference
```
dneto0: Why do missing values not cause a difference when processing loops, but they do cause a difference in the multi-pred case?   Should explain why we consider no difference here.  Is this a bug?

Example:
Let's say this header H has two predecessors P and Q; P does not dominate Q; that variable V has value p0 when coming from P, that p0 is defined in P, and V has no value when coming from Q.  This code says no difference needs to be resolved.   Then let's say the loop headed by H loads from V.  We're trying to replace that load with an SSA value, and I think this code will replace it with p0.  But that use of p0 in the loop is not dominated by the definition in P since we could have reached H via Q instead of via P.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -0,0 +1,1239 @@
...
+%29 = OpLabel
+%32 = OpAccessChain %_ptr_Input_float %BC %39
+%33 = OpLoad %float %32
+%34 = OpFAdd %float %38 %33
+OpBranch %25
+%25 = OpLabel
+%36 = OpIAdd %int %39 %int_1
+OpBranch %23
+%24 = OpLabel
+OpStore %fo %38
```
dneto0: %38 is the OpPhi.  So you're storing the value that *f* had at the *top* of the loop, not the bottom.  That means you're missing one of the additions ?

source/opt/local_ssa_elim_pass.h
```diff
@@ -0,0 +1,257 @@
...
+
+  // Return true if loop header block
+  bool IsLoopHeader(ir::BasicBlock* block_ptr) const;
+
+  // Initialize label2ssa_map_ entry for block |block_ptr| with single
+  // predecessor.
+  void SSABlockInitSinglePred(ir::BasicBlock* block_ptr);
+
+  // Return true if variable is loaded in block with |label| or in
+  // any succeeding block.
```
greg-lunarg: Yes.

source/opt/local_ssa_elim_pass.h
```diff
@@ -0,0 +1,257 @@
...
+  bool IsLiveAfter(uint32_t var_id, uint32_t label) const;
+
+  // Initialize label2ssa_map_ entry for loop header block pointed to
+  // |block_itr| by merging entries from all predecessors. If any value
+  // ids differ for any variable across predecessors, create a phi function
+  // in the block and use that value id for the variable in the new map.
+  // Assumes all predecessors have been visited by SSARewrite except the
```
greg-lunarg: Yes, now EliminateLocalMultiStore. Fixed.

source/opt/local_ssa_elim_pass.h
```diff
@@ -0,0 +1,257 @@
...
+  std::unordered_set<uint32_t> seen_target_vars_;
+
+  // Cache of previously seen non-target types
+  std::unordered_set<uint32_t> seen_non_target_vars_;
+
+  // Set of label ids of visited blocks
+  std::unordered_set<uint32_t> visitedBlocks_;
+
+  // Map from type to undef for current function
+  std::unordered_map<uint32_t, uint32_t> type2undefs_;
```
greg-lunarg: OK.

include/spirv-tools/optimizer.hpp
```diff
@@ -230,6 +230,23 @@ Optimizer::PassToken CreateInlinePass();
 // by LocalSingleStoreElim and LocalSSARewrite.
```
greg-lunarg: Yes, now LocalMultiStoreElim.

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+      AddStores(varId, &deadInsts);
+    deadInsts.pop();
+  }
+}
+
+bool LocalSSAElimPass::HasOnlySupportedRefs(uint32_t varId) {
+  if (supported_ref_vars_.find(varId) != supported_ref_vars_.end())
+    return true;
+  analysis::UseList* uses = def_use_mgr_->GetUses(varId);
+  assert(uses != nullptr);
```
greg-lunarg: OK.

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+void LocalSSAElimPass::InitSSARewrite(ir::Function& func) {
+  // Init predecessors
+  label2preds_.clear();
+  for (auto& blk : func) {
+    uint32_t blkId = blk.id();
+    blk.ForEachSuccessorLabel([&blkId, this](uint32_t sbid) {
+      label2preds_[sbid].push_back(blkId);
+    });
+  }
+  // Remove variables with non-load/store refs from target variable set
```
greg-lunarg: Added additive language.

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+    else if (merge_ii->opcode() == SpvOpSelectionMerge)
+      mbid = merge_ii->GetSingleWordInOperand(kSelectionMergeMergeBlockIdInIdx);
+  }
+  return mbid;
+}
+
+void LocalSSAElimPass::ComputeStructuredSuccessors(ir::Function* func) {
+  for (auto& blk : *func) {
+    // If no predecessors in function, make successor to pseudo entry
+    if (label2preds_[blk.id()].size() == 0)
```
greg-lunarg: Well, I felt like I couldn't just ignore these blocks. If the user doesn't want them processed, the user should delete them.

greg-lunarg: PS. Looking at this routine, I realized I needed to make the fix that I made to structured ordering for DeadBranchElimination, that is, to include the continue target as well as the merge block to the beginning of the successor list.

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+  const uint32_t label = block_ptr->id();
+  const uint32_t predLabel = label2preds_[label].front();
+  assert(visitedBlocks_.find(predLabel) != visitedBlocks_.end());
+  label2ssa_map_[label] = label2ssa_map_[predLabel];
+}
+
+bool LocalSSAElimPass::IsLiveAfter(uint32_t var_id, uint32_t label) const {
+  // For now, return very conservative result: true. This will result in
+  // correct, but possibly usused, phi code to be generated. A subsequent
+  // DCE pass should eliminate this code.
```
greg-lunarg: Interesting

source/opt/local_ssa_elim_pass.cpp
greg-lunarg: Thanks!

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+  uint32_t backLabel = 0;
+  for (uint32_t predLabel : label2preds_[label])
+    if (visitedBlocks_.find(predLabel) == visitedBlocks_.end()) {
+      assert(backLabel == 0);
+      backLabel = predLabel;
+      break;
+    }
+  assert(backLabel != 0);
+  // Determine merge block.
+  auto mergeInst = (*block_itr)->begin();
```
greg-lunarg: Good catch.

This is only called on a loop header so it is safe to just reference the second-to-last instruction.

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+  // Add all stored variables in loop. Set their default value id to zero.
+  for (auto bi = block_itr; (*bi)->id() != mergeLabel; ++bi) {
+    ir::BasicBlock* bp = *bi;
+    for (auto ii = bp->begin(); ii != bp->end(); ++ii) {
+      if (ii->opcode() != SpvOpStore)
+        continue;
+      uint32_t varId;
+      (void) GetPtr(&*ii, &varId);
+      if (!IsTargetVar(varId))
+        break;
```
greg-lunarg: Yeah, I am just trying to keep indentation down.

Anyway, good catch! This was a copy and paste error. I had copied out of a case statement where break *is* the thing you want to do.

source/opt/local_ssa_elim_pass.cpp
greg-lunarg: Yes, you are right. Good catch. This was some early (bad) thinking before I realized that defs had to dominate uses. I fixed this in the other merge routine but didn't come back and fix it here.

source/opt/local_ssa_elim_pass.cpp
greg-lunarg: Doh

source/opt/local_ssa_elim_pass.cpp
greg-lunarg: Sure.

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+      phiItr->SetInOperand(idx, { valItr->second });
+      // Analyze uses now that they are complete
+      def_use_mgr_->AnalyzeInstUse(&*phiItr);
+    }
+  }
+}
+
+bool LocalSSAElimPass::LocalSSAElim(ir::Function* func) {
+  InitSSARewrite(*func);
+  // Process all blocks in structured order.
```
greg-lunarg: It is just one way (the simplest?) to make sure that all predecessor blocks are processed before a block itself.

greg-lunarg: Improved the comment.

source/opt/local_ssa_elim_pass.cpp
```diff
@@ -0,0 +1,759 @@
...
+      id2block_[blk.id()] = &blk;
+  }
+
+  // Initialize Target Type Caches
+  seen_target_vars_.clear();
+  seen_non_target_vars_.clear();
+
+  // Initialize set of variables only referenced by supported operations
+  supported_ref_vars_.clear();
+
```
greg-lunarg: I was presuming that a pass was only Process'ed once.

Would all the pass-level collection names be preceeded by _., as in _.visitedBlocks_?

It doesn't seem like there is a lot of difference between enumerating the collections in the Context or in Initialize.

For the moment I will clear them all in Initialize() and we can refine from there.

test/opt/local_ssa_elim_test.cpp
greg-lunarg: I think this is right. Remember that this is a for loop with no other break:

  //     float f = 0.0;
  //     for (int i=0; i<4; i++) {
  //       f = f + BC[i];
  //     }
  //     fo = f;

Notice its sole conditional/break is at the top of the loop before any computation. So the result to be stored into fo will indeed have been computed in the previous loop and will indeed be in the phi corresponding to f at the top of the loop at the time of the break.

greg-lunarg:
I think I have addressed all issues/concerns up to this point. Let me know if there is anything I missed or anything else.

test/opt/local_ssa_elim_test.cpp
dneto0: Of course you're right.  I'm not used to seeing the OpPhi before the exiting branch at the top of the loop. (!)

source/opt/local_ssa_elim_pass.cpp
dneto0: ::warm-and-fuzzy-emoji::

source/opt/local_ssa_elim_pass.cpp
dneto0: Agreed.   I was thinking out loud.  Sprinkling _. all over the place would be cumbersome.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -0,0 +1,1239 @@
...
+OpBranch %25
+%26 = OpLabel
+%44 = OpLoad %float %f
+OpStore %fo %44
+OpReturn
+OpFunctionEnd
+)";
+
+  const std::string after =
+      R"(%47 = OpUndef %float
```
dneto0: Nice to see this Undef

test/opt/local_ssa_elim_test.cpp
```diff
@@ -0,0 +1,1239 @@
...
+OpBranchConditional %38 %40 %39
+%40 = OpLabel
+OpBranch %26
+%39 = OpLabel
+OpBranch %27
+%27 = OpLabel
+%43 = OpIAdd %int %46 %int_1
+OpBranch %25
+%26 = OpLabel
+%49 = OpPhi %float %48 %28 %36 %40
```
dneto0: Nice to see the interior variable "t" is handled correctly in a general/regular way.  Even if that means generating phi values that end up unused.  We can sweep those away later.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -0,0 +1,1239 @@
...
+}
+
+TEST_F(LocalSSAElimTest, SwapProblem) {
+  // #version 140
+  // 
+  // in float fe;
+  // out float fo;
+  // 
+  // void main()
+  // {
```
dneto0: Good test case.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -0,0 +1,1239 @@
...
+
+  const std::string after =
+      R"(%43 = OpUndef %float
+%main = OpFunction %void None %11
+%23 = OpLabel
+%24 = OpLoad %float %fe
+%25 = OpConvertFToS %int %24
+OpBranch %26
+%26 = OpLabel
+%40 = OpPhi %float %float_0 %23 %41 %28
```
dneto0: It's really nice to see this just fall out.  (ok, with lots of your hard work!)  This assembly couldn't have been coded any nicer by hand.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -0,0 +1,1239 @@
...
+%41 = OpLabel
+OpBranch %26
+%40 = OpLabel
+OpBranch %27
+%27 = OpLabel
+%43 = OpIAdd %int %46 %int_1
+OpBranch %25
+%26 = OpLabel
+%49 = OpPhi %float %45 %28 %37 %41
+%50 = OpPhi %float %48 %28 %45 %41
```
dneto0: Nice resolution of t here.

dneto0:
Rebased, squashed, and pushed into master as cc8bad3a5

PR #695, opened 2017-07-04
Add option to skip install; also implement GNUInstallDirs

d3x0r:
Add option to skip install; also implement GNUInstallDirs; for some systems lib64 is the appropriate output path and not 'lib'.

define an option SKIP_SPIRV_TOOLS_INSTALL which internally becomes ENABLE_SPIRV_TOOLS_INSTALL to control whether install rules are used.
If this is used as a static library just included by other projects, installation of this is not required.


pierremoreau:
If you want to skip the install, then you just need to not run `make install`, or `ninja install`, etc. Or am I missing something else?
Also, you are missing a conditional around the `install()` in `source/comp/CMakeLists.txt`.

d3x0r:
I do want to run make install but on an application that's including shaderc which is in turn including this.  It ends up cluttering the output because I'm just using it as a static library and do not want to include headers and static libraries with the final product.

I only need to build the code to link to it.

I'll add that; but spiv-tools/source/comp was not included in the latest vulkan SDK so I didn't see it; updated now.

dneto0:
Does this negatively impact Windows?  (I'm testing now to see the effect.)

dneto0:
I also confirmed that GNUInstallDir is available on cmake 2.8.12 which is the baseline for this project.

dneto0:
Rebased and pushed into master as  fd70a1d7a

Issue #697, opened 2017-07-04
spvtools::ir::Instruction::ForEachInId skips memory-semantics-id and scope-id operands


dneto0:
Should use spvIsIdType to check against the several possibilities of operand types that could be an Id.

PR #698, opened 2017-07-04
Include memory and semantics IDs when iterating over inbound IDs

dneto0:
Fixes Instruction::ForEachInId so it covers
SPV_OPERAND_TYPE_MEMORY_SEMANTICS_ID and SPV_OPERAND_TYPE_SCOPE_ID.
Future proof a bit by using the common spvIsIdType routine.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/697


dneto0:
Rebased.  Added update to CHANGES file.  Keeping approval.

PR #701, opened 2017-07-10
Opt: Add CommonUniformElim pass

greg-lunarg:
This is another in a series of passes designed to work together to reduce SPIR-V file size.

This pass consolidates all identical uniform reads into a single uniform read. It begins by converting all uniform access chain loads with constant indices into an equivalent load and extract. It then combines all common loads and all common extracts with a single constant index, at times moving the common code to a block which dominates all uses.

This pass requires structured control flow, logical addressing and currently does not support any extensions. It also does not optimize decorated loads or access chains.

### DIFF
--- a/source/opt/common_uniform_elim_pass.cpp
+++ b/source/opt/common_uniform_elim_pass.cpp
@@ -367,7 +367,7 @@ bool CommonUniformElimPass::CommonUniformLoadElimination(ir::Function* func) {
   for (auto bi = structuredOrder.begin(); bi != structuredOrder.end(); ++bi) {
     ir::BasicBlock* bp = *bi;
     // Check if we are exiting outermost control construct. If so, remember
-    // new load insertion point.
+    // new load insertion point. Trying to keep register pressure down.
     if (mergeBlockId == bp->id()) {
       mergeBlockId = 0;
       insertItr = bp->begin();
--- a/source/opt/common_uniform_elim_pass.cpp
+++ b/source/opt/common_uniform_elim_pass.cpp
@@ -459,6 +459,7 @@ bool CommonUniformElimPass::CommonExtractElimination(ir::Function* func) {
     for (auto ii = bi->begin(); ii != bi->end(); ++ii) {
       if (ii->opcode() != SpvOpCompositeExtract)
         continue;
+      // TODO(greg-lunarg): Support multiple indices
       if (ii->NumInOperands() > 2)
         continue;
       if (HasUnsupportedDecorates(ii->result_id()))
--- a/source/opt/common_uniform_elim_pass.cpp
+++ b/source/opt/common_uniform_elim_pass.cpp
@@ -277,9 +277,6 @@ bool CommonUniformElimPass::UniformAccessChainConvert(ir::Function* func) {
 
 void CommonUniformElimPass::ComputeStructuredSuccessors(ir::Function* func) {
   for (auto& blk : *func) {
-    // If no predecessors in function, make successor to pseudo entry
-    if (label2preds_[blk.id()].size() == 0)
-      block2structured_succs_[&pseudo_entry_block_].push_back(&blk);
     // If header, make merge block first successor.
     uint32_t cbid;
     const uint32_t mbid = MergeBlockIdIfAny(blk, &cbid);
@@ -310,7 +307,7 @@ void CommonUniformElimPass::ComputeStructuredOrder(
   
   order->clear();
   spvtools::CFA<ir::BasicBlock>::DepthFirstTraversal(
-      &pseudo_entry_block_, get_structured_successors, ignore_block,
+      &*func->begin(), get_structured_successors, ignore_block,
       post_order, ignore_edge);
 }
 
@@ -329,9 +326,6 @@ bool CommonUniformElimPass::CommonUniformLoadElimination(ir::Function* func) {
     ++insertItr;
   uint32_t mergeBlockId = 0;
   for (auto bi = structuredOrder.begin(); bi != structuredOrder.end(); ++bi) {
-    // Skip pseudo entry block
-    if (*bi == &pseudo_entry_block_)
-      continue;
     ir::BasicBlock* bp = *bi;
     // Check if we are exiting outermost control construct. If so, remember
     // new load insertion point.
@@ -502,10 +496,7 @@ Pass::Status CommonUniformElimPass::ProcessImpl() {
 }
 
 CommonUniformElimPass::CommonUniformElimPass()
-    : module_(nullptr), def_use_mgr_(nullptr),
-      pseudo_entry_block_(std::unique_ptr<ir::Instruction>(
-          new ir::Instruction(SpvOpLabel, 0, 0, {}))),
-      next_id_(0) {}
+    : module_(nullptr), def_use_mgr_(nullptr), next_id_(0) {}
 
 Pass::Status CommonUniformElimPass::Process(ir::Module* module) {
   Initialize(module);
--- a/source/opt/common_uniform_elim_pass.cpp
+++ b/source/opt/common_uniform_elim_pass.cpp
@@ -111,6 +111,10 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
   *varId = ip->GetSingleWordInOperand(
       op == SpvOpStore ? kStorePtrIdInIdx : kLoadPtrIdInIdx);
   ir::Instruction* ptrInst = def_use_mgr_->GetDef(*varId);
+  while (ptrInst->opcode() == SpvOpCopyObject) {
+    *varId = ptrInst->GetSingleWordInOperand(kCopyObjectOperandInIdx);
+    ptrInst = def_use_mgr_->GetDef(*varId);
+  }
   ir::Instruction* varInst = ptrInst;
   while (varInst->opcode() != SpvOpVariable) {
     if (IsNonPtrAccessChain(varInst->opcode())) {
--- a/source/opt/common_uniform_elim_pass.cpp
+++ b/source/opt/common_uniform_elim_pass.cpp
@@ -1,0 +1,517 @@
+// Copyright (c) 2017 The Khronos Group Inc.
+// Copyright (c) 2017 Valve Corporation
+// Copyright (c) 2017 LunarG Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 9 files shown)


greg-lunarg:
This does have a slight issue: I will need to prohibit moving loads of OpTypeImage and OpTypeSampler out of the blocks they are used in. Will have a patch shortly.

source/opt/common_uniform_elim_pass.h
```diff
@@ -0,0 +1,214 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_OPT_COMMON_UNIFORM_ELIM_PASS_H_
+#define LIBSPIRV_OPT_COMMON_UNIFORM_ELIM_PASS_H_
+
+#include <unordered_map>
```
dneto0: These are out of order.   (Non-blocking, as this is easy enough to overlook.)

source/opt/common_uniform_elim_pass.h
```diff
@@ -0,0 +1,214 @@
...
+  // been loaded in the same block for types whose loads cannot cross blocks.
+  bool CommonUniformLoadElimBlock(ir::Function* func);
+
+  // Eliminate duplicated extracts of same id. Extract may be moved to same
+  // block as the id definition. This is primarily intended for extracts
+  // from uniform loads. Most effective if preceded by
+  // CommonUniformLoadElimination().
+  bool CommonExtractElimination(ir::Function* func);
+
+  // Initialize the label2SSA map entry for a block. Insert phi instructions
```
dneto0: These two lines don't belong?   Second line ends in a sentence fragment.

source/opt/common_uniform_elim_pass.h
```diff
@@ -0,0 +1,214 @@
...
+      block2structured_succs_;
+
+  // Map from block's label id to its predecessor blocks ids
+  std::unordered_map<uint32_t, std::vector<uint32_t>> label2preds_;
+
+  // Map from uniform variable id to its common load id
+  std::unordered_map<uint32_t, uint32_t> uniform2load_id_;
+
+  // Map of extract composite ids to map of indices to insts
+  std::unordered_map<uint32_t, std::unordered_map<uint32_t,
```
dneto0: Without reading the implementation, this type is a bit mysterious.  In   comp2idx2inst_[id][id2] it's not clear to me what id2 is, or why it maps to a list of instructions.

I'll suspend my question/recommendation until I read more.

greg-lunarg: Given:

%100 = OpCompositeExtract %1, 0
...
%200 = OpCompositeExtract %1, 0

both instructions would be inserted into comp2idx2inst[1][0].

Later, this data structure is used coalesce all extracts of the same composite value and index into one extract.

greg-lunarg: Because coalescing involves moving the first extract (ie some work and potential downstream register pressure), I didn't want to initiate the move until I knew I had two or more extracts to coalesce. So I collected them first in this data structure and only do the coalesce when >=2 instructions are present.

source/opt/common_uniform_elim_pass.h
greg-lunarg: Yep. Thanks. Deleted.

test/opt/common_uniform_elim_test.cpp
```diff
@@ -0,0 +1,672 @@
...
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::CommonUniformElimPass>(
+      predefs + before, predefs + after, true, true);
+}
+
+TEST_F(CommonUniformElimTest, Basic2) {
+  // Note: This test exemplifies the following:
```
dneto0: Comment does not match the test. 

This comment is the same as for Basic1. 

But Basic2 has an OpFNegate but Basic1 does not.

source/opt/common_uniform_elim_pass.h
```diff
@@ -0,0 +1,214 @@
...
+  // is not enabled. It also currently does not support any extensions.
+  bool EliminateCommonUniform(ir::Function* func);
+
+  // Initialize extensions whitelist
+  void InitExtensions();
+
+  // Return true if all extensions in this module are allowed by this pass.
+  bool AllExtensionsSupported() const;
+
+  // Return true if all uses of varId are only through supported reference
```
greg-lunarg: Re OpMemberDecorate, in this context I was only concerned with Decorates that could be applied to non-type instructions, so I have renamed this function to IsNonTypeDecorate, and fixed the comment.

As for OpGroupDecorate, I would rather not redundantly document this in optimizer.hpp for each pass which does not support it, which are numerous. I do have TODOs in each pass file that prohibits OpGroupDecorate. I have also opened an issue so that users can see this more readily. Will that suffice?

source/opt/common_uniform_elim_pass.h
```diff
@@ -0,0 +1,214 @@
...
+
+  // Map from block's label id to its predecessor blocks ids
+  std::unordered_map<uint32_t, std::vector<uint32_t>> label2preds_;
+
+  // Map from uniform variable id to its common load id
+  std::unordered_map<uint32_t, uint32_t> uniform2load_id_;
+
+  // Map of extract composite ids to map of indices to insts
+  std::unordered_map<uint32_t, std::unordered_map<uint32_t,
+      std::list<ir::Instruction*>>> comp2idx2inst_;
```
greg-lunarg: Changing this will require me to rewrite some of my code. I would rather put the time into higher priorities. I will add a TODO for this point.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -0,0 +1,613 @@
...
+    return false;
+  for (auto u : *uses) {
+    const SpvOp op = u.inst->opcode();
+    if (IsDecorate(op))
+      return true;
+  }
+  return false;
+}
+
+bool CommonUniformElimPass::HasOnlyNamesAndDecorates(uint32_t id) const {
```
greg-lunarg: I have added code to the main routine to explicitly prohibit GroupDecorate (as I have for many of the other passes as well). I have added a TODO there.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -0,0 +1,613 @@
...
+  const uint32_t ptrPteTypeId = GetPointeeTypeId(ptrInst);
+  std::vector<ir::Operand> ext_in_opnds;
+  ext_in_opnds.push_back(
+    ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID,
+      std::initializer_list<uint32_t>{ldResultId}));
+  uint32_t iidIdx = 0;
+  ptrInst->ForEachInId([&iidIdx, &ext_in_opnds, this](const uint32_t *iid) {
+    if (iidIdx > 0) {
+      const ir::Instruction* cInst = def_use_mgr_->GetDef(*iid);
+      uint32_t val = cInst->GetSingleWordInOperand(kConstantValueInIdx);
```
greg-lunarg: Like LocalAccessChainConvert, I have added code bailing on shaders with non-32-bit ints. I have also added a TODO.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -0,0 +1,613 @@
...
+      ext_in_opnds.push_back(
+        ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_LITERAL_INTEGER,
```
greg-lunarg: Yes, I saw the new issue. Thanks.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -0,0 +1,613 @@
...
+        continue;
+      if (!IsConstantIndexAccessChain(ptrInst))
+        continue;
+      if (HasUnsupportedDecorates(ii->result_id()))
+        continue;
+      if (HasUnsupportedDecorates(ptrInst->result_id()))
+        continue;
+      std::vector<std::unique_ptr<ir::Instruction>> newInsts;
+      uint32_t replId;
+      GenACLoadRepl(ptrInst, newInsts, replId);
```
greg-lunarg: Yep. Fixed.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -0,0 +1,613 @@
...
+  // Find insertion point in first block to copy non-dominating loads.
+  auto insertItr = func->begin()->begin();
+  while (insertItr->opcode() == SpvOpVariable ||
+      insertItr->opcode() == SpvOpNop)
+    ++insertItr;
+  uint32_t mergeBlockId = 0;
+  for (auto bi = structuredOrder.begin(); bi != structuredOrder.end(); ++bi) {
+    ir::BasicBlock* bp = *bi;
+    // Check if we are exiting outermost control construct. If so, remember
+    // new load insertion point.
```
greg-lunarg: OK. Added comment.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -0,0 +1,613 @@
...
+  return modified;
+}
+
+bool CommonUniformElimPass::CommonExtractElimination(ir::Function* func) {
+  // Find all composite ids with duplicate extracts.
+  for (auto bi = func->begin(); bi != func->end(); ++bi) {
+    for (auto ii = bi->begin(); ii != bi->end(); ++ii) {
+      if (ii->opcode() != SpvOpCompositeExtract)
+        continue;
+      if (ii->NumInOperands() > 2)
```
greg-lunarg: Yes, this is a simplification and yes, I will add a comment. There is some opportunity for multiple indices, but it is in the noise. It would also complicate the (already complicated) data structure I am using.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -0,0 +1,613 @@
...
+
+    modified |= CommonUniformLoadElimBlock(func);
+    return modified;
+}
+
+void CommonUniformElimPass::Initialize(ir::Module* module) {
+
+  module_ = module;
+
+  // Initialize function and block maps
```
greg-lunarg: Doh!

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -0,0 +1,613 @@
...
+    "SPV_KHR_multiview",
+    "SPV_NVX_multiview_per_view_attributes",
+    "SPV_NV_viewport_array2",
+    "SPV_NV_stereo_view_rendering",
+    "SPV_NV_sample_mask_override_coverage",
+    "SPV_NV_geometry_shader_passthrough",
+    "SPV_AMD_texture_gather_bias_lod",
+    "SPV_KHR_storage_buffer_storage_class",
+    // SPV_KHR_variable_pointers
+    //   Currently do not support extended pointer expressions
```
greg-lunarg: Thanks!

greg-lunarg:
I think have addressed all concerns. Let me know if I missed something or you see anything else.

dneto0:
Rebased, squashed, and pushed into master as f4b29f3bf79

PR #704, opened 2017-07-11
DeadBranchElim: Only remove blocks with no predecessors.

greg-lunarg:
Must be careful not to remove blocks pointed at by unreachable blocks. This change makes the algorithm cleaner and removes a fairly redundant data structure.

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -218,27 +218,11 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
     def_use_mgr_->KillInst(mergeInst);
 
     // Iterate to merge block adding dead blocks to elimination set
-    std::unordered_set<uint32_t> deadSuccIds;
-    deadSuccIds.insert(deadLabId);
     auto dbi = bi;
     ++dbi;
     uint32_t dLabId = (*dbi)->id();
     while (dLabId != mergeLabId) {
-      if (deadSuccIds.find(dLabId) != deadSuccIds.end()) {
-        // Add successor blocks to dead successor set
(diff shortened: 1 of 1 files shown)


source/opt/dead_branch_elim_pass.cpp
```diff
@@ -248,14 +232,19 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
...
-    // If eliminated branch is to merge label, add current block to dead blocks.
+    // elimBlocks are now blocks which cannot precede merge block. Also,
+    // if eliminated branch is to merge label, remember the conditional block
+    // also cannot precede merge block.
+    uint32_t deadCondLabId = 0;
     if (deadLabId == mergeLabId)
-      deadSuccIds.insert((*bi)->id());
-    (*dbi)->ForEachPhiInst([&deadSuccIds, this](ir::Instruction* phiInst) {
+      deadCondLabId = (*bi)->id();
+    (*dbi)->ForEachPhiInst([&elimBlocks, &deadCondLabId, this](
```
dneto0: Slightly better to capture deadCondLabId by value.  You're not modifying it ever.

dneto0:
> Must be careful not to remove blocks pointed at by unreachable blocks.

Does that mean this fixes a bug?   This PR comment seems to suggest that's true.  If so, would be nice to document it with an issue. (And a test case.)
But for now I'm just accepting.

dneto0:
Rebased and pushed into master as  e2544ddc90

greg-lunarg:
This change was based more on a thought experiment. As I reviewed the algorithm in the context of recent discussions of unreachable blocks, I realized it could leave unreachable blocks referring to a deleted label, which would be invalid. I also realized the fix would result in cleaner code. Could we just chalk this change up to cleaner code and call the thought experiement bug fix a side benefit? :)

dneto0:
Yes!  :-)

PR #705, opened 2017-07-12
LocalMultiStore: Always put varId for backedge on loop phi function.

greg-lunarg:
And always patch the backedge operand when patching phi functions. This
approach is more correct and cleaner. The previous code was generating
incorrect phis when the backedge block had no predecessors.


dneto0:
Rebased and pushed into master as  fe24e03

PR #706, opened 2017-07-13
Fix Visual Studio size_t cast compiler warning

lenny-lunarg:
Visual Studio was complaining about possible loss of data on 64-bit builds, due to an implicit cast from `size_t` to `int`. This changes the data to use an int all around.

Since the project is set to treat warnings as errors by default, this means that 64-bit builds of the repo were failing unless the user specifically disabled the `/Werror` flag.

### DIFF
--- a/source/comp/markv_codec.cpp
+++ b/source/comp/markv_codec.cpp
@@ -1367,7 +1367,7 @@ spv_result_t MarkvDecoder::DecodeInstruction(spv_parsed_instruction_t* inst) {
 
   assert(inst->num_words == std::accumulate(
       parsed_operands_.begin(), parsed_operands_.end(), 1,
-      [](size_t num_words, const spv_parsed_operand_t& operand) {
+      [](int num_words, const spv_parsed_operand_t& operand) {
         return num_words += operand.num_words;
   }) && "num_words in instruction doesn't correspond to the sum of num_words"
         "in the operands");


lenny-lunarg:
I'm not sure why appveyor failed. It marked the build as failed as soon as I created this pull request, and I'm pretty sure that it never ran anything. If you know a way to re-run appveyor, that would be helpful.

pierremoreau:
@lenny-lunarg If you hover over the “AppVeryor was unable to …” message, you will see the whole line. In this case, it did not run any tests as it somehow considers this pull request as non-mergeable.

dneto0:
This PR is one commit on top of current master, so it should be trivially mergeable.   I tried to see if I could restart the build but since it never started I can't do that.

And on the substance of the issue ...  I think MSVC has it wrong.  The num_words fields are uint16_t, the initializer 1 is an int, and so the resolved type should be int rather than size_t.  But that's in theory and we should live with the compilers we have.

I'll accept the patch if you can get it past the bots.  

@lenny-lunarg: Let's assume this is just an Appveyor glitch. I suggest you make a trivial spacing change, force-commit the result on the same branch, and let's see if Appveyor can build it.

dneto0:
@lenny-lunarg What version of Visual Studio were you using?   I'm reviewing our Appveyor config and it has VS 2013, VS 2015, Debug and Release, but only uses "Any CPU".  So I guess 64-bit coverage is not included.  
I'd like to cover a 64-bit case, but I don't want to extend testing time too much for the additional coverage.

lenny-lunarg:
I was using Visual Studio 2017, but I also had the same problem with Visual Studio 2015. The reason the appveyor tests don't cover 64-bit builds is that they let cmake use the default generator. Cmake, however, has different generators for 32 and 64-bit builds on Windows and always defaults to 32-bit. So to run a 64-bit build the line:

`cmake ..`

would need to be changed to something like:

`cmake -G"Visual Studio 14 2015 Win64" ..`

to enable 64-bit builds. I'm not aware of a way to force cmake to use a 64-bit generator, but otherwise choose the visual studio version to use, but it might exist (it seems like there should be a way). I haven't used appveyor much, so I'm not really sure how this would have to be set for appveyor to do that.

Also, I'll go make some trivial change to the commit and see if that kicks off a build.

dneto0:
@lenny-lunarg : Thanks for kicking it off again.  It's in progress.

Also, that's great background info on cmake vs. 64bit.    Our appveyor config is in https://github.com/KhronosGroup/SPIRV-Tools/blob/master/.appveyor.yml 
You can see we explicitly invoke the initial cmake.   I think with come imagination (and env variable substitution) we can convince AppVeyor to make 64-bit builds.  I filed https://github.com/KhronosGroup/SPIRV-Tools/projects/2#card-3770821 to track the task of adding such a config.

PR #726, opened 2017-07-19
Opt: Add extension whitelists to size-reduction passes.

greg-lunarg:
Because extensions change validation rules it is prudent to keep white-lists
of approved extensions for each pass.

Currently only SPV_KHR_variable_pointers is disallowed in passes which
do pointer analysis. Positive and negative tests of the general extensions
white-list mechanism were added to aggressive_dce but cover all passes.


antiagainst:
Squashed, rebased, and merged as https://github.com/KhronosGroup/SPIRV-Tools/commit/118241558173ae5816500ab82821279996d8ee71

PR #732, opened 2017-07-24
Change HasOnlySupportedRefs to consider OpCopyObject

antiagainst:


antiagainst:
@greg-lunarg: Hey Greg, I'd appreciate if you can take a look at the second commit to make sure everything is fine. :)

greg-lunarg:
Thanks. Looking...

greg-lunarg:
Good catch and good attempt at fix. I think the following is even "more" what is needed:

     if (IsNonPtrAccessChain(op) || op == SpvOpCopyObject) {
       if (!HasOnlySupportedRefs(u.inst->result_id()))
         return false;
     }
     else if (op != SpvOpStore && op != SpvOpLoad && op != SpvOpName)
       return false;

The goal is to make sure that we are restricting to variables which are ultimately only consumed by Loads, Stores or Names in order to simplify data dependence checking. We want to avoid variables consumed by OpFrexp, OpFunctionCall, etc.

We would want this change in both:

local_single_block_elim_pass.cpp
local_single_store_elim_pass.cpp

antiagainst:
@greg-lunarg: Gotcha. Thanks for the suggestion! Done. :)

antiagainst:
Hey @greg-lunarg, I'll merge this PR if you think it's fine now. :)

greg-lunarg:
Yes. I think it is fine. Thanks for your fixes.

PR #733, opened 2017-07-24
Add multi-sequence move-to-front implementation

atgoo:
Add MultiMoveToFront class which supports multiple move-to-front
sequences and allows to promote value in all sequences at once.

Added caching for last accessed sequence handle and last accessed value
in each sequence.

### DIFF
--- a/source/util/move_to_front.h
+++ b/source/util/move_to_front.h
@@ -20,7 +20,9 @@
 #include <cstdint>
 #include <iomanip>
 #include <iostream>
+#include <map>
 #include <ostream>
+#include <set>
 #include <sstream>
 #include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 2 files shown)


test/move_to_front_test.cpp
```diff
@@ -780,6 +798,88 @@ TEST(MoveToFront, String) {
...
+  EXPECT_EQ(1u, rank);
+}
+
+TEST(MultiMoveToFront, Empty) {
+  MultiMoveToFront<std::string> multi_mtf;
+
+  uint32_t rank = 0;
+  std::string value;
+
+  EXPECT_EQ(0u, multi_mtf.GetSize(1001));
```
dneto0: Good.

dneto0:
Rebased and pushed into master as  30bee67

source/util/move_to_front.h
```diff
@@ -305,6 +314,113 @@ class MoveToFront {
...
+    if (!cached_mtf_ || cached_handle_ != handle) {
+      cached_handle_ = handle;
+      cached_mtf_ = &mtfs_[handle];
+    }
+
+    return *cached_mtf_;
+  }
+
+  // Container holding MoveToFront objects. Map key is sequence handle.
+  std::map<uint32_t, MoveToFront<Val>> mtfs_;
```
atgoo: std::map is used to have predictable lifetime for *cached_mtf_, I don't think unordered_map is guaranteed to preserve it.

PR #737, opened 2017-07-31
Added C++ code generation to spirv-stats

atgoo:
The tool can now generate C++ code returning some of the histograms and
Huffman codecs generated from those histograms.

### DIFF
--- a/tools/stats/stats.cpp
+++ b/tools/stats/stats.cpp
@@ -14,6 +14,7 @@
 
 #include <cassert>
 #include <cstring>
+#include <fstream>
 #include <iostream>
 #include <unordered_map>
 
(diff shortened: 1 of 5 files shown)


tools/stats/stats.cpp
```diff
@@ -43,8 +44,35 @@ TIP: In order to collect statistics from all .spv files under current dir use
...
 Options:
-  -h, --help                       Print this help.
-  -v, --verbose                    Print additional info to stderr.
+  -h, --help
+                   Print this help.
+
+  -v, --verbose
+                   Print additional info to stderr.
+
+  --codegen_opcode_hist
```
dneto0: I think we normally use hyphens in command line options instead of underscores.  But this is your tool so you get to choose.  :-)

tools/stats/stats_analyzer.cpp
```diff
@@ -239,3 +560,207 @@ void StatsAnalyzer::WriteOpcodeMarkov(std::ostream& out) {
...
+    const double kMaxValue = 1000.0;
+    uint32_t value = uint32_t(kMaxValue * double(count) / double(total));
+    if (value == 0)
+      value = 1;
+    out << "    { SpvOp" << GetOpcodeString(opcode)
+        << ", " << value << " },\n";
+  }
+
+  // Add kMarkvNoneOfTheAbove as a signal for unknown opcode.
+  out << "    { kMarkvNoneOfTheAbove, " << 10 << " },\n";
```
dneto0: Why not just use literal string " 10 " as well?

tools/stats/stats_analyzer.cpp
```diff
@@ -15,19 +15,340 @@
...
-#include "spirv/1.1/spirv.h"
 
 using libspirv::SpirvStats;
 
 namespace {
 
+// Returns all SPIR-V v1.2 opcodes.
+std::vector<uint32_t> GetAllOpcodes() {
+  return std::vector<uint32_t>({
+    SpvOpNop,
```
dneto0: Would be slightly nicer to generate this automatically.  But this is good enough for your stats.

dneto0:
Rebased and pushed into master as 55b73a036

PR #738, opened 2017-08-01
Opt: Add MemPass, move all shared functions to it.

greg-lunarg:
Get rid of duplicated functions from the size-reduction passes. All classes that use these shared functions now derive from MemPass.


greg-lunarg:
Getting rid of duplicated code is the first step to making changes to enable the optimizations that glslang needs to generate valid SPIR-V.

source/opt/mem_pass.h
```diff
@@ -0,0 +1,127 @@
...
+
+#include "basic_block.h"
+#include "def_use_manager.h"
+#include "module.h"
+#include "pass.h"
+
+namespace spvtools {
+namespace opt {
+
+// See optimizer.hpp for documentation.
```
dneto0: Need a real comment here.  MemPass isn't mentioned in optimizer.hpp

test/opt/local_access_chain_convert_test.cpp
```diff
@@ -69,6 +69,34 @@ OpName %gl_FragColor "gl_FragColor"
 %gl_FragColor = OpVariable %_ptr_Output_v4float Output
 )";
 
+  const std::string predefs_after =
```
dneto0: I see that %int_1 = OpConstant %int 1 is now removed by the pass.  Presumably that's because this pass is now using a more generalized set of utilities.  Ok by me.

source/opt/mem_pass.h
dneto0: I fixed this myself.  (After I hit the approval button)

dneto0:
Rebased and pushed into master as  c1b46eed.  That also fixes the class comment for MemPass.

PR #746, opened 2017-08-03
Update README.md with new code-reduction passes

greg-lunarg:
I have also made an attempt to group/classify them to add clarity.


dneto0:
Trailing space removed, and pushed into master as   0bd8300fc2  
I also documented the  common uniform load removal in CHANGES, in the subsquent commit

PR #747, opened 2017-08-03
Inline: Split out InlineExhaustivePass from InlinePass

greg-lunarg:
InlinePass is the generic inliner that other inliners will derive from. Ultimately should move the main inlining routine into InlinePass, parameterized by a function for inline policy and a function for what do if the function is not inlinable.


dneto0:
Rebased and pushed into master as e28bd39997801

PR #748, opened 2017-08-03
Add id descriptor feature to SPIR-V

atgoo:
Id descriptors are computed as a recursive hash of all instructions used
to define an id. Descriptors are invarint of actual id values and
the similar code in different files would produce the same descriptors.

Multiple ids can have the same descriptor. For example
%1 = OpConstant %u32 1
%2 = OpConstant %u32 1
would produce two ids with the same descriptor. But
%3 = OpConstant %s32 1
%4 = OpConstant %u32 2
would have descriptors different from %1 and %2.

Descriptors will be used as handles of move-to-front sequences in SPIR-V
compression.


source/id_descriptor.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <iostream>
+
+#include "opcode.h"
+#include "operand.h"
+
+namespace libspirv {
+
+namespace {
+
+uint32_t HashU32Array(const std::vector<uint32_t>& words) {
```
dneto0: method comment.

source/id_descriptor.cpp
```diff
@@ -0,0 +1,74 @@
...
+  const uint32_t kMagic = 2654435761;
```
dneto0: how do you choose these magic constants?
How do you know it's relatively good?

source/id_descriptor.h
```diff
@@ -0,0 +1,56 @@
...
+#include <vector>
+
+#include "spirv-tools/libspirv.hpp"
+
+namespace libspirv {
+
+// Computes and stores id descriptors.
+//
+// Descriptors are computed as hash of all words in the instruction where ids
+// were substituted with previosly computed descriptors.
```
dneto0: "previously"

source/id_descriptor.h
```diff
@@ -0,0 +1,56 @@
...
+class IdDescriptorCollection {
+ public:
+  IdDescriptorCollection() {
+    words_.reserve(16);
+  }
+
+  // Issues new descriptor for the result id of the given instruction.
+  // This function needs to be sequentially called for every instruction in the
+  // module.
+  uint32_t IssueNewDescriptor(const spv_parsed_instruction_t& inst);
```
dneto0: But it might be a reused descriptor.  So the "New" part of the name seems wrong.
Suggest "MakeDescriptorForInst" ?

Also say that the return value is the descriptor is associated with the instruction.

source/id_descriptor.h
```diff
@@ -0,0 +1,56 @@
...
+  uint32_t GetDescriptor(uint32_t id) const {
+    const auto it = id_to_descriptor_.find(id);
+    if (it == id_to_descriptor_.end())
+      return 0;
+    return it->second;
+  }
+
+ private:
+  std::unordered_map<uint32_t, uint32_t> id_to_descriptor_;
+  std::vector<uint32_t> words_;
```
dneto0: Should say this words_ vector is scratch space used during descriptor generation.  the fact it's a class data member is just an optimization to avoid allocations.

source/id_descriptor.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+}  // namespace
+
+uint32_t IdDescriptorCollection::IssueNewDescriptor(
+    const spv_parsed_instruction_t& inst) {
+  if (!inst.result_id)
+    return 0;
+
+  words_.clear();
```
dneto0: Minor:  Why not clear at the end rather than the beginning?  Leaving it at the end will mean you have data you'll never use again, lying around.

source/id_descriptor.cpp
atgoo: The magic constant comes from the so called Knuth's multiplicative hash. Other things were just made up. It doesn't need to be particularly good.

source/id_descriptor.cpp
atgoo: Ok. But strictly speaking the data would still be there even after clear().

atgoo:
Feedback addressed.

source/id_descriptor.cpp
dneto0: Ok, the allocation is still there.  True.  So no there's no performance difference.  Good point.

source/id_descriptor.cpp
dneto0: You should put that in the code comment. (!)

dneto0:
Rebased, squashed, and pushed into master as 78cf86150

PR #749, opened 2017-08-03
Mem2Reg: Allow Image and Sampler types as base target types.

greg-lunarg:
One step to support opaque type propagation in glslang.


dneto0:
Rebased and pushed into master as d9a450121e

PR #751, opened 2017-08-04
AccessChainConvert: Add HasOnlySupportedRefs()

greg-lunarg:
This avoids conversion on variables which will not ultimately be optimized.
Also removed an obsolete restriction from FindTargetVars(). Also added
decorates to supported refs (eg. RelaxedPrecision). Also fixed name to
IsNonTypeDecorate().


dneto0:
Rebased and pushed into master as  f0fe601dc

PR #753, opened 2017-08-07
ADCE: Add support for function calls

greg-lunarg:
ADCE will now generate correct code in the presence of function calls.
This is needed for opaque type optimization needed by glslang. Currently
all function calls are marked as live. TODO: mark calls live only if they
write a non-local.

Also remove duplicate declaration from optimizer.hpp

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -376,18 +376,6 @@ Optimizer::PassToken CreateCommonUniformElimPass();
 // eliminated with standard dead code elimination.
 Optimizer::PassToken CreateAggressiveDCEPass();
 
-// Creates a pass to consolidate uniform references.
-// For each entry point function in the module, first change all constant index
-// access chain loads into equivalent composite extracts. Then consolidate 
-// identical uniform loads into one uniform load. Finally, consolidate
-// identical uniform extracts into one uniform extract. This may require
-// moving a load or extract to a point which dominates all uses.
-//
-// This pass requires a module to have structured control flow ie shader
-// capability. It also requires logical addressing ie Addresses capability
-// is not enabled. It also currently does not support any extensions.
-Optimizer::PassToken CreateCommonUniformElimPass();
-
 // Creates a compact ids pass.
 // The pass remaps result ids to a compact and gapless range starting from %1.
 Optimizer::PassToken CreateCompactIdsPass();
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -54,7 +54,7 @@ void AggressiveDCEPass::AddStores(uint32_t ptrId) {
       } break;
       case SpvOpLoad:
         break;
-      // Assume it stores eg frexp, modf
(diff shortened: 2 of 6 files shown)


include/spirv-tools/optimizer.hpp
```diff
@@ -376,18 +376,6 @@ Optimizer::PassToken CreateCommonUniformElimPass();
 // eliminated with standard dead code elimination.
 Optimizer::PassToken CreateAggressiveDCEPass();
 
-// Creates a pass to consolidate uniform references.
```
dneto0: This seems out of place.   Need to rebase?

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -90,20 +90,34 @@ bool AggressiveDCEPass::AllExtensionsSupported() const {
...
-  if (dead_insts_.find(tInst) != dead_insts_.end())
+  if (dead_insts_.find(tInst) != dead_insts_.end()) {
     def_use_mgr_->KillInst(inst);
+    return true;
+  }
+  return false;
+}
+
+void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
+  // Only process locals
```
dneto0: This is fine.  You only need to process locals because the loads of non-locals are already included in the live set earlier in the flow.

test/opt/aggressive_dead_code_elim_test.cpp
```diff
@@ -680,6 +680,133 @@ OpFunctionEnd
       assembly, assembly, true, true);
 }
 
+TEST_F(AggressiveDCETest, ElimWithCall) {
+  // This demonstrates that "dead" function calls are not eliminated.
+  // Also demonstrates that DCE will happen in presense of function call.
```
dneto0: "presence", but no matter.

include/spirv-tools/optimizer.hpp
greg-lunarg: Not a mistake. This was an extra declaration that somehow snuck in. I piggybacked it into this PR rather than create a new PR for something so minor.

test/opt/aggressive_dead_code_elim_test.cpp
greg-lunarg: Doh! Will he never learn?

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Yes. This was just to abstract processing of "real" loads and the virtual load represented in passing a variable to a function call.

include/spirv-tools/optimizer.hpp
dneto0: oh.  Ok. Maybe it was a bad merge on my part.  certainly possible in all my rebasing.

test/opt/aggressive_dead_code_elim_test.cpp
dneto0: Not enough to block!

dneto0:
Rebased, typo fixed, and pushed into master as b0310a4156345

Issue #755, opened 2017-08-09
inliner generates invalid OpLabel if callee has single return in block that does not appear last


dneto0:
Example code:

```
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main"
               OpSource GLSL 450
               OpName %main "main"
               OpName %foo_ "foo("
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
       %bool = OpTypeBool
       %true = OpConstantTrue %bool
       %main = OpFunction %void None %3
          %5 = OpLabel
         %12 = OpFunctionCall %void %foo_
               OpReturn
               OpFunctionEnd
       %foo_ = OpFunction %void None %3
          %7 = OpLabel
               OpSelectionMerge %11 None
               OpBranchConditional %true %10 %11
         %11 = OpLabel
               OpReturn  ;  This causes returnLabelId to be non-zero, but earlyReturn is false.
         %10 = OpLabel
               OpBranch %11
               OpFunctionEnd
```

It's generated from this vertex shader, but where I reorder, in the SPIR-V assembly, the last two basic blocks of the "foo" function:

```
#version 450

void foo() {
 if (true) {
 }
}
void main() {
  foo();
}
```

When running spirv-opt with inline-entry-points-exhaustive:

- returnLabelId is non-zero since we encountered an OpReturn well before the OpFunctionEnd
- but earlyReturn is false
- but at OpFunctionEnd we generate "continue target" with OpLabel for result id singleTripLoopContinueId.  But that is still zero.
- this causes a new instruction with OpLabel opcode, but zero result Id, and generates a 1-word OpLabel instruction, which is invalid.

Likely remedy is to guard the codegen of continue target block by earlyReturn.  Still have to create the block with label Id returnLabelId.

dneto0:
I expect to have  patch soon.
Cc @greg-lunarg

dneto0:
I have a code fix at https://github.com/dneto0/SPIRV-Tools/tree/inline-avoid-nil-label 
Need to make a test for it.

PR #757, opened 2017-08-10
Inliner: callee can have early return that isn't multi-return

dneto0:
Avoid generating an invalid OpLabel.
Create the continue target for the single-trip loop only if
you actually created the header for the single-trip loop.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/755

### DIFF
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -228,9 +228,9 @@ void InlinePass::GenInlineCode(
   ir::Function* calleeFn = id2function_[call_inst_itr->GetSingleWordOperand(
       kSpvFunctionCallFunctionId)];
 
-  // Check for early returns
-  auto fi = early_return_.find(calleeFn->result_id());
-  bool earlyReturn = fi != early_return_.end();
+  // Check for multiple returns in the callee.
+  auto fi = multi_return_funcs_.find(calleeFn->result_id());
+  const bool multiReturn = fi != multi_return_funcs_.end();
 
   // Map parameters to actual arguments.
   MapParams(calleeFn, call_inst_itr, &callee2caller);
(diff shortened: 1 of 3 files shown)


dneto0:
I'd appreciate a review from @greg-lunarg

source/opt/inline_pass.cpp
```diff
@@ -228,9 +228,9 @@ void InlinePass::GenInlineCode(
   ir::Function* calleeFn = id2function_[call_inst_itr->GetSingleWordOperand(
       kSpvFunctionCallFunctionId)];
 
-  // Check for early returns
-  auto fi = early_return_.find(calleeFn->result_id());
-  bool earlyReturn = fi != early_return_.end();
+  // Check for multiple returns in the callee.
+  auto fi = multi_return_funcs_.find(calleeFn->result_id());
+  bool multiReturn = fi != multi_return_funcs_.end();
```
atgoo: Consider
const bool multiReturn = multi_return_funcs_.count(calleeFn->result_id());

source/opt/inline_pass.cpp
```diff
@@ -250,11 +250,14 @@ void InlinePass::GenInlineCode(
   uint32_t returnLabelId = 0;
   bool multiBlocks = false;
   const uint32_t calleeTypeId = calleeFn->type_id();
+  // new_blk_ptr is a new basic block in the caller.  New instructions are
+  // written to it.  It is created when we encounter the OpLabel
+  // of the first calle block.
```
atgoo: typo: calle block

source/opt/inline_pass.cpp
dneto0: Fixed in next round.  I also updated the lambda to capture multiReturn by value rather than reference.

dneto0:
All feedback addressed.

greg-lunarg:
Good catch. The fix works and is solid and the name changes are an improvement. Thanks!

dneto0:
Thanks.  I'll take that as formal approval.

PR #761, opened 2017-08-14
Huffman codec can serialize to text

atgoo:
Refactored the Huffman codec implementation and added ability to
serialize to C++-like text format. This would reduce the time-complexity
if loading hard-coded codecs.

### DIFF
--- a/source/util/huffman_codec.h
+++ b/source/util/huffman_codec.h
@@ -38,31 +38,53 @@ namespace spvutils {
 // literal).
 template <class Val>
 class HuffmanCodec {
-  struct Node;
-
  public:
+  // Huffman tree node.
+  struct Node {
+    Node() {}
+
+    // Creates Node from serialization leaving weight and id undefined.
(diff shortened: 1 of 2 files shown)


source/util/huffman_codec.h
```diff
@@ -83,52 +105,98 @@ class HuffmanCodec {
...
+    const std::string indent1 = std::string(indent_num_whitespaces, ' ');
+    const std::string indent2 = std::string(indent_num_whitespaces + 2, ' ');
+
+    std::stringstream code;
+    code << "(" << root_ << ", {\n";
+
+    for (const Node& node : nodes_) {
+      code << indent2 << "{";
+      if (value_is_text)
+        code << "\"";
```
dneto0: This text encoding scheme is fine for simple cases.  If you want to deserialize it for subtle cases then you have to use more complicated escaping.  For example, if the node names themselves contain quote characters.
But this is fine for your current uses.

dneto0:
Rebased and pushed into master as 17d941af4f

PR #762, opened 2017-08-15
Opt: Have "size" passes process full entry point call tree.

greg-lunarg:
Includes code to deal correctly with OpFunctionParameter. This change
is needed by opaque propagation which may not exhaustively inline
entry point functions.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -33,9 +33,13 @@ const uint32_t kExtInstInstructionInIndx = 1;
 
 bool AggressiveDCEPass::IsLocalVar(uint32_t varId) {
   const ir::Instruction* varInst = def_use_mgr_->GetDef(varId);
-  assert(varInst->opcode() == SpvOpVariable);
+  const SpvOp op = varInst->opcode();
+  if (op != SpvOpVariable && op != SpvOpFunctionParameter) 
+    return false;
   const uint32_t varTypeId = varInst->type_id();
   const ir::Instruction* varTypeInst = def_use_mgr_->GetDef(varTypeId);
+  if (varTypeInst->opcode() != SpvOpTypePointer)
+    return false;
   return varTypeInst->GetSingleWordInOperand(kTypePointerStorageClassInIdx) ==
       SpvStorageClassFunction;
 }
@@ -54,7 +58,7 @@ void AggressiveDCEPass::AddStores(uint32_t ptrId) {
       } break;
       case SpvOpLoad:
         break;
-      // Assume it stores eg frexp, modf, function call
+      // If default, assume it stores eg frexp, modf, function call
       case SpvOpStore:
       default: {
         if (live_insts_.find(u.inst) == live_insts_.end())
@@ -136,7 +140,8 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
             worklist_.push(&inst);
         } break;
         default: {
-          // eg. control flow, function call, atomics
+          // eg. control flow, function call, atomics, function param,
+          // function return
           // TODO(greg-lunarg): function calls live only if write to non-local
           if (!IsCombinator(op))
             worklist_.push(&inst);
@@ -176,6 +181,10 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
         (void) GetPtr(*iid, &varId);
         ProcessLoad(varId);
       });
+    }
+    // If function parameter, treat as if it's result id is loaded from
+    else if (liveInst->opcode() == SpvOpFunctionParameter) {
+      ProcessLoad(liveInst->result_id());
     }
     worklist_.pop();
   }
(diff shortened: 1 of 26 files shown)


dneto0:
Please rebase to fix the conflicts.

source/opt/pass.cpp
```diff
@@ -0,0 +1,70 @@
...
+bool Pass::ProcessEntryPointCallTree(
+    ProcessFunction& pfn, ir::Module* module) {
+  // Map from function's result id to function
+  std::unordered_map<uint32_t, ir::Function*> id2function;
+  id2function.clear();
+  for (auto& fn : *module)
+    id2function[fn.result_id()] = &fn;
+  // Process call tree
+  bool modified = false;
+  std::unordered_set<uint32_t> todo;
```
dneto0: I'm concerned that processing order is unspecified.  This could lead to unstable results across versions or platforms, which would be a pain for testing and debugging.

I think changing "todo" to a std::queue and "next" to a vector would get us most of the way toward deterministic behaviour.  (The entry points iteration is a deterministic scan of a vector.)

Yes there might be duplicated entries in each but they're eliminated by the check against "done".  I'm not concerned about the temporary slight waste of space.  In practice you can reduce some of the space use by filtering against the "done" collection before insertion into "todo";  optional as I think the extra lookup would probably slow it down on average.   But that's just a guess.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -117,7 +116,8 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
     ptrInst = def_use_mgr_->GetDef(*varId);
   }
   ir::Instruction* varInst = ptrInst;
-  while (varInst->opcode() != SpvOpVariable) {
+  while (varInst->opcode() != SpvOpVariable &&
```
dneto0: Need to update the description of GetPtr in common_uniform_elim_pass.h, including the fact that the varId parameter is now badly named.  Maybe it should be objId?

source/opt/common_uniform_elim_pass.h
```diff
@@ -178,9 +178,6 @@ class CommonUniformElimPass : public Pass {
   // Def-Uses for the module we are processing
   std::unique_ptr<analysis::DefUseManager> def_use_mgr_;
 
-  // Map from function's result id to function
```
dneto0: Thanks for deleting these now-unused variables.  That kind of cleanup is easy to miss. :-)

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -79,6 +74,10 @@ void LocalSingleStoreElimPass::SingleStoreAnalyze(ir::Function* func) {
           non_ssa_vars_.insert(varId);
           continue;
         }
+        if (!HasOnlySupportedRefs(varId)) {
```
dneto0: This move is subtle, but I trust you.

source/opt/mem_pass.cpp
```diff
@@ -81,7 +81,13 @@ bool MemPass::IsPtr(uint32_t ptrId) {
     ptrInst = def_use_mgr_->GetDef(varId);
   }
   const SpvOp op = ptrInst->opcode();
-  return op == SpvOpVariable || IsNonPtrAccessChain(op);
+  if (op == SpvOpVariable || IsNonPtrAccessChain(op))
+    return true;
+  if (op != SpvOpFunctionParameter)
```
dneto0: Need to update the method comment for MemPass::IsPtr

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -33,9 +32,13 @@ const uint32_t kExtInstInstructionInIndx = 1;
 
 bool AggressiveDCEPass::IsLocalVar(uint32_t varId) {
```
greg-lunarg: Fixed.

source/opt/pass.cpp
```diff
@@ -0,0 +1,70 @@
...
+    for (auto ii = bi->begin(); ii != bi->end(); ++ii)
+      if (ii->opcode() == SpvOpFunctionCall)
+        next->insert(ii->GetSingleWordInOperand(0));
+}
+
+bool Pass::ProcessEntryPointCallTree(
+    ProcessFunction& pfn, ir::Module* module) {
+  // Map from function's result id to function
+  std::unordered_map<uint32_t, ir::Function*> id2function;
+  id2function.clear();
```
greg-lunarg: Good point. A little cut-and-paste error.

source/opt/pass.cpp
greg-lunarg: Good catch.

source/opt/common_uniform_elim_pass.cpp
greg-lunarg: That would be better.

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: But I appreciate your trust ;)

source/opt/mem_pass.cpp
greg-lunarg: Done.

dneto0:
Squashed, rebased, and pushed into master as c8c86a0d36

PR #764, opened 2017-08-16
Support supplying extra definitions via CMAKE variable

antiagainst:
For addressing https://github.com/Microsoft/DirectXShaderCompiler/pull/529

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,11 @@ The following CMake options are supported:
   warnings encountered by enabling the compiler-specific compiler front-end
   option.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -206,6 +206,10 @@ The following CMake options are supported:
   warnings encountered by enabling the compiler-specific compiler front-end
   option.
 
+Additionally, you can pass additional defintions to SPIR-V tools via setting
```
dneto0: "pass additional C preprocessor definitions"....

And give an example. 
I had to look up exactly what add_definitions does in CMake.

PR #767, opened 2017-08-18
Opt: Create InlineOpaquePass

greg-lunarg:
Only inline calls to functions with opaque params or return ie only inline calls needed to complete opaque propagation. This is in support of opaque propagation needed by glslang.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -213,7 +213,8 @@ Optimizer::PassToken CreateInlineExhaustivePass();
 // Creates an opaque inline pass.
 // An opaque inline pass inlines all function calls in all functions in all
 // entry point call trees where the called function contains an opaque type
-// in either its parameters or return value. The intent is to enable, albeit
+// in either its parameter types or return type. An opaque type is currently
+// defined as Image, Sampler or SampledImage. The intent is to enable, albeit
 // through brute force, analysis and optimization across these function calls
 // by subsequent passes in order to remove the storing of opaque types which is // not legal in Vulkan. Functions that are not in the call tree of an entry
 // point are not changed.
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -201,14 +201,23 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 // these is left for future improvements. 
 Optimizer::PassToken CreateBlockMergePass();
 
-// Creates an inline pass.
-// An inline pass exhaustively inlines all function calls in all functions
-// designated as an entry point. The intent is to enable, albeit through
-// brute force, analysis and optimization across function calls by subsequent
-// passes. As the inlining is exhaustive, there is no attempt to optimize for
-// size or runtime performance. Functions that are not designated as entry
-// points are not changed.
+// Creates an exhaustive inline pass.
+// An inline pass attempts to exhaustively inline all function calls in
+// all functions in an entry point call tree. The intent is to enable,
+// albeit through brute force, analysis and optimization across function
+// calls by subsequent optimization passes. As the inlining is exhaustive,
+// there is no attempt to optimize for size or runtime performance. Functions
+// that are not in the call tree of an entry point are not changed.
 Optimizer::PassToken CreateInlineExhaustivePass();
+  
+// Creates an opaque inline pass.
+// An inline pass inlines all function calls in all functions in an entry
+// point call tree. The intent is to enable, albeit through brute force,
+// analysis and optimization across these function calls by subsequent
+// passes in order to remove the storing of opaque types which is not
+// legal in Vulkan. Functions that are not in the call tree of an entry
+// point are not changed.
+Optimizer::PassToken CreateInlineOpaquePass();
   
 // Creates a single-block local variable load/store elimination pass.
 // For every entry point function, do single block memory optimization of 
@@ -228,6 +237,8 @@ Optimizer::PassToken CreateInlineExhaustivePass();
 // This pass is most effective if preceeded by Inlining and 
 // LocalAccessChainConvert. This pass will reduce the work needed to be done
 // by LocalSingleStoreElim and LocalMultiStoreElim.
+//
+// Only functions in the call tree of an entry point are processed.
 Optimizer::PassToken CreateLocalSingleBlockLoadStoreElimPass();
 
 // Create dead branch elimination pass.
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -202,8 +202,8 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
 Optimizer::PassToken CreateBlockMergePass();
 
 // Creates an exhaustive inline pass.
-// An inline pass attempts to exhaustively inline all function calls in
-// all functions in an entry point call tree. The intent is to enable,
+// An exhaustive inline pass attempts to exhaustively inline all function
+// calls in all functions in an entry point call tree. The intent is to enable,
 // albeit through brute force, analysis and optimization across function
 // calls by subsequent optimization passes. As the inlining is exhaustive,
 // there is no attempt to optimize for size or runtime performance. Functions
@@ -211,11 +211,11 @@ Optimizer::PassToken CreateBlockMergePass();
 Optimizer::PassToken CreateInlineExhaustivePass();
   
 // Creates an opaque inline pass.
-// An inline pass inlines all function calls in all functions in an entry
-// point call tree. The intent is to enable, albeit through brute force,
-// analysis and optimization across these function calls by subsequent
-// passes in order to remove the storing of opaque types which is not
-// legal in Vulkan. Functions that are not in the call tree of an entry
+// An opaque inline pass inlines all function calls in all functions in all
+// entry point call trees where the called function contains an opaque type
+// in either its parameters or return value. The intent is to enable, albeit
+// through brute force, analysis and optimization across these function calls
+// by subsequent passes in order to remove the storing of opaque types which is // not legal in Vulkan. Functions that are not in the call tree of an entry
 // point are not changed.
 Optimizer::PassToken CreateInlineOpaquePass();
   
--- a/source/opt/inline_opaque_pass.cpp
+++ b/source/opt/inline_opaque_pass.cpp
@@ -38,6 +38,7 @@ bool InlineOpaquePass::IsOpaqueType(uint32_t typeId) {
     default:
       break;
   }
+  // TODO(greg-lunarg): Handle arrays containing opaque type
   if (typeInst->opcode() != SpvOpTypeStruct)
     return false;
   // Return true if any member is opaque
--- a/source/opt/inline_opaque_pass.cpp
+++ b/source/opt/inline_opaque_pass.cpp
@@ -1,0 +1,121 @@
+// Copyright (c) 2017 The Khronos Group Inc.
+// Copyright (c) 2017 Valve Corporation
+// Copyright (c) 2017 LunarG Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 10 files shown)


johnkslang:
What about functions that set opaques, but without them being paramaters or return values? Can this policy be extended to include any functions that write to opaque types?  In fact, that could probably be the sole thing to look for, independent of parameter and return types: simply does it contain an opaque l-value.

greg-lunarg:
@johnkslang 

One thing you may not be aware of is that we just added code so that all the propagation passes operate on all functions in the entry point call tree, not just the entry point. So they will do propagation in functions that are not inlined into the entry point.

For example:

... = OpFunction main
...
... = OpFunctionCall foo ...
...
OpFunctioEnd

... = OpFunction foo ...
...
<Self contained code that loads, stores, loads again and uses an opaque type>
...
OpFunctionEnd

(I think) we don't necessarily want to inline foo into main since the remaining passes will now fully process and eliminate any opaque memory ops from foo without having to inline it into main.

Does this address your concern? If not, can you sketch out an example that gives a little more detail on your concern?

source/opt/inline_opaque_pass.cpp
```diff
@@ -0,0 +1,121 @@
...
+      } else {
+        ++ii;
+      }
+    }
+  }
+  return modified;
+}
+
+void InlineOpaquePass::Initialize(ir::Module* module) {
+  InitializeInline(module);
```
dneto0: I just checked that this method has no comment in inline_pass.h
Please fix that.  :-)

source/opt/inline_opaque_pass.cpp
```diff
@@ -0,0 +1,121 @@
...
+    case SpvOpTypeImage:
+    case SpvOpTypeSampledImage:
+      return true;
+    case SpvOpTypePointer:
+      return IsOpaqueType(typeInst->GetSingleWordInOperand(
+          kTypePointerTypeIdInIdx));
+    default:
+      break;
+  }
+  if (typeInst->opcode() != SpvOpTypeStruct)
```
dneto0: What about Array and RuntimeArray?  

Ok, RuntimeArray would be impossible in the shading langauges I know.  But *perhaps* Array.
Recommend a TODO here, rather than any code change.    Because if you did add Array, then I'd ask for a test. :-)

include/spirv-tools/optimizer.hpp
```diff
@@ -201,15 +201,24 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
...
+// that are not in the call tree of an entry point are not changed.
 Optimizer::PassToken CreateInlineExhaustivePass();
   
+// Creates an opaque inline pass.
+// An inline pass inlines all function calls in all functions in an entry
+// point call tree. The intent is to enable, albeit through brute force,
+// analysis and optimization across these function calls by subsequent
+// passes in order to remove the storing of opaque types which is not
+// legal in Vulkan. Functions that are not in the call tree of an entry
+// point are not changed.
```
dneto0: Should say what you mean by "opaque".  For now it's restricted to Image, Sampler, SampledImage, Structs containing such, or pointers containing such.

Need that clarification because OpenCL has lots of types that are opaque as well.

tools/opt/opt.cpp
```diff
@@ -68,46 +68,55 @@ NOTE: The optimizer is a work in progress.
...
                any gaps.
   --inline-entry-points-exhaustive
-               Exhaustively inline all function calls in entry point functions.
-               Currently does not inline calls to functions with early return
+               Exhaustively inline all function calls in entry point call tree
+               functions. Currently does not inline calls to functions with
+               early return in a loop.
+  --inline-entry-points-opaque
+               Inline all function calls in entry point call tree functions
+               where the callee has opaque type parameters or return value.
```
dneto0: I'd clarify about what opaque means. Suggest "A type is considered opaque if it is an image, sampler, sampled image, is a pointer to one, or is a struct containing one."

test/opt/inline_opaque_test.cpp
```diff
@@ -0,0 +1,412 @@
...
+#include "pass_fixture.h"
+#include "pass_utils.h"
+
+namespace {
+
+using namespace spvtools;
+
+using InlineOpaqueTest = PassTest<::testing::Test>;
+
+TEST_F(InlineOpaqueTest, InlineOpaqueArg) {
```
dneto0: Worth being more specific:   InlineCallWithStructArgContainingSampledImage

Key for me is that your IsOpaqueType had to look inside an OpTypeStruct to see the opaqueness.

dneto0:
@johnkslang : 
 
@greg-lunarg  made a case that I think addresses the root of your concern.  See https://github.com/KhronosGroup/SPIRV-Tools/pull/767/files#diff-ca22934be6a97d842bfe974dd25b3ae2R223 

It's of the form he described:  main calls main2, main2 calls foo_struct_S_t_vf2_vf21_ and this last function takes a struct param containing an opaque.  The result is as Greg describes: the leaf function is inlined into main2, and that's all.    The point is that the opaque object reference in main2 must have come from a module-scope variable, and so "main" has no need to inline main2.

dneto0:
Also, high level comment: With @greg-lunarg 's recent refactorings this code is getting easier and easier to read, and hopefully write.  :-)

include/spirv-tools/optimizer.hpp
```diff
@@ -201,15 +201,24 @@ Optimizer::PassToken CreateEliminateDeadConstantPass();
...
+// An inline pass attempts to exhaustively inline all function calls in
+// all functions in an entry point call tree. The intent is to enable,
+// albeit through brute force, analysis and optimization across function
+// calls by subsequent optimization passes. As the inlining is exhaustive,
+// there is no attempt to optimize for size or runtime performance. Functions
+// that are not in the call tree of an entry point are not changed.
 Optimizer::PassToken CreateInlineExhaustivePass();
   
+// Creates an opaque inline pass.
+// An inline pass inlines all function calls in all functions in an entry
```
greg-lunarg: I should actually read stuff before I check it in :(

include/spirv-tools/optimizer.hpp
greg-lunarg: Good idea.

source/opt/inline_opaque_pass.h
```diff
@@ -0,0 +1,61 @@
...
+namespace opt {
+
+// See optimizer.hpp for documentation.
+class InlineOpaquePass : public InlinePass {
+
+ public:
+  InlineOpaquePass();
+  Status Process(ir::Module*) override;
+
+  const char* name() const override { return "inline-exhaustive"; }
```
greg-lunarg: Gah!

source/opt/inline_opaque_pass.cpp
greg-lunarg: Added TODO.

source/opt/inline_opaque_pass.cpp
greg-lunarg: You are thorough!

tools/opt/opt.cpp
greg-lunarg: Fixed.

test/opt/inline_opaque_test.cpp
greg-lunarg: OK.

include/spirv-tools/optimizer.hpp
dneto0: Well, it's not that inefficient if I catch it.  :-)
We're probably in the right spot tradeoff-wise.

dneto0:
Rebased, squashed, and pushed into master as 1803679f9

johnkslang:
> One thing you may not be aware of is that we just added code so that all the propagation passes operate on all functions in the entry point call tree, not just the entry point. So they will do propagation in functions that are not inlined into the entry point.

Okay, sounds good, thanks.

greg-lunarg:
@johnkslang @dneto0

Thinking about this more, I think John is right. We may need to inline any subroutine with opaque types. And more. Here is why:

The main issue is dead branch elimination. Essentially, we need to resolve multiple branches that store different opaque values into the same variable. But this potentially means inlining more than just functions with opaque in their interface. For example, here is a case where a function (foo) must be inlined even though it has no opaque types at all:

%main = OpFunction ...
...
%1 = OpFunctionCall %foo
OpSelectionMerge %4
OpBranchConditional %1 %2 %3
%2 = OpLabel
%5 = OpLoad ... $image0
OpStore %image %5
OpBranch %4
%3 = OpLabel
%5 = OpLoad ... $image1
OpStore %image %5
OpBranch %4
%4 = OpBranch
%6 = OpLoad %image
%7 = OpImageSample ... %6 ...
...
OpFunctionEnd
%foo = OpFunction ...
%10 = OpLabel
%11 = OpConstantFalse
OpReturnValue %11
OpFunctionEnd

While this is simplistic, one can imagine more likely and more complex cases which pose the same problem.

And related to John's question, functions containing the opaque code above but depending on (non-opaque) parameters will need to be inlined so that needed argument propagation can occur.

Given the current regimen of using inlining to do interprocedural analysis, computing the minimal set of inlining required for a call tree with opaque type could be fairly complex.

At this point, a first-order approximation could be to inline into main all functions with opaque type and all functions in all blocks and structured components which dominate the call as well as any loops which include the call.

Slightly better, one might imagine an iterative solution which does inlining and propagation until opaque stores have been eliminated. But I don't think our tool chest is sufficient at the moment for such a solution and some additional passes will be needed.

As opaque inlining sits for the moment, it is not a complete solution and as such, probably not worth advertising. I will probably continue to think about the iterative solution, but will probably start working to add the first-order approximation above so we have some kind of solution short of exhaustive inlining.

dneto0:
Interesting.

But that's a lot of complexity that might not be needed.  I presume Microsoft's fxc only does so much work, and bails out at some point.  There is no spec saying how far it goes; it's just using heuristics, and those heuristics are changing over time.

I'd only want to put in the work if there are cases we know about that fxc does resolve, and our current tooling does not.

greg-lunarg:
Exhaustive should always resolve any case. It should be the bailout.

A good question is whether the additional complexity here is worth the inlining that is avoided.

Your comparison to fxc is apt. Is there any evidence that fxc does anything but exhaustive inlining?

Might be interesting to take a handful of HLSL shaders with opaque stores and see what percentage of inlining can be avoided. That might give us a handle on how much complexity is justified.

johnkslang:
I believe the use cases are driven by large projects sharing code, where interface setup (to get to a common interface) is encapsulated in these functions that translate from one opaque interface to another, needing some basic control flow around assignments to opaques.

johnkslang:
FYI, I put together a small shader that blows up if exhaustively inlined.

The SPIR-V text is ~10 times bigger than the shader, no inlining.

The FXC output text is ~100 times bigger than the shader, exhaustive inlining.

I think some instruction caches would rather preserve the function calls.  This is a chance we have to do better with both performance and size, at the same time.

greg-lunarg:
I agree that minimizing inlining is an opportunity for us.

Here is one idea for an incremental approach to minimizing inlining when attempting to eliminate invalid opaque usages.

1) Inline all calls of functions with opaque args or return values and run "propagation" passes if changed.
2)  If shader *still* contains invalid opaque usages, inline all calls of functions with opaque locals and run "propagation" passes if changed.
3)  If shader *still* contains invalid opaque usages, inline all calls in blocks which contain or dominate blocks with opaque code and run "propagation" passes if changed.
4)  If shader *still* contains invalid opaque usages, do exhaustive inlining and run "propagation" passes.

There might be other intermediate stages, but hopefully the basic approach is clear.

Let me know if you see issues or other opportunities for improvement.

greg-lunarg:
Note in the preceding post that I am proposing a pass that detects the presence of invalid (opaque) SPIR-V. If we decide this is a good approach, this would obviate the need for infrastructure in glslang to detect and report invalid SPIR-V, at least for the moment.

This pass would not change code, but only report back "found" or "not found". We could add two more return statuses for a pass, or we could piggy-back on SuccessWithChange and SuccessWithoutChange.

PR #769, opened 2017-08-20
Fix: background color was erroneously reset on Win32 platform.

LukasBanana:
Fix: background color was erroneously reset on Win32 platform.
This was especially annoying on the Windows PowerShell, where the background color is not black per default.


dneto0:
Thanks for the contribution.  I hope to test and review it today.

dneto0:
Rebased and pushed onto master as  4fe8e389a

Thanks for the fix!

PR #775, opened 2017-08-24
README: Add reference to SPIR-V size reduction white paper

greg-lunarg:


dneto0:
Rebased and pushed onto master as cdeb669541

PR #777, opened 2017-08-28
vim Disassembler: Print colour codes only when writing to a terminal

dneto0:


tools/dis/dis.cpp
```diff
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
+#include <stdio.h> // Need fileno
```
AWoloszyn: If you want, you could switch this to a CMake configuration step, 
CHECK_INCLUDE_FILE(unistd.h HAS_UNISTD)
CHECK_FUNCTION_EXISTS(isatty, HAS_ISATTY)

But I don't mind this either.

dneto0: Good to know.

dneto0:
Rebased

PR #779, opened 2017-08-28
Fix mingw build (source/print.cpp)

atgoo:
source/print.cpp doesn't compile due to integer conversion.

Fix not tested (no windows machine).


dneto0:
The MinGW build runs on a Linux machine, so you can replicate this.  :-)

dneto0:
I still get the error with your patch:

...spirv-tools/source/print.cpp:45:9: error: conversion to ‘WORD {aka short unsigned int}’ from ‘int’ may alter its value [-Werror=conversion]
   color |= (bufInfo.wAttributes & WORD(0xfff0));
         ^

dneto0:
I'm fixing the patch and testing.

PR #780, opened 2017-08-29
InlineOpaque: Remove from usage until complete opaque policy is desig…

greg-lunarg:
…ned.


dneto0:
Rebased and pushed into master as  b9b9a533344ed7

Issue #781, opened 2017-08-29
Validator: Should allow OpTypePointer with same pointee type and storage class


dneto0:
This is exercised by SPV_KHR_variable_pointers.

See the "Types" subsection of
https://www.khronos.org/registry/spir-v/extensions/KHR/SPV_KHR_variable_pointers.html#_modifications_to_the_spir_v_specification_version_1_1

dneto0:
Could either be precise, and enable this behaviour only when either of the variable pointers capabilities is enabled, or could be blunt and always enable this.

dneto0:
The variable pointers extension allows you to have two OpTypePointer with the same storage class and pointee type, but the pointer types have ArrayStride decorations.  The decorations might have the same value or different.

dneto0:
Fixed by #783

PR #783, opened 2017-08-29
Extension allows multiple same OpTypePointer types

atgoo:
SPV_KHR_variable_pointers allows OpTypePointer to declare multiple
identical types

https://github.com/KhronosGroup/SPIRV-Tools/issues/781

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -426,17 +426,20 @@ bool ValidationState_t::RegisterUniqueTypeDeclaration(
     key.insert(key.end(), inst.words + words_begin, inst.words + words_end);
   }
 
+  // If module has extension SPV_KHR_variable_pointers, then OpTypePointer
+  // is allowed to create identical types so long as their decorations differ.
   if (inst.opcode == SpvOpTypePointer &&
       HasExtension(Extension::kSPV_KHR_variable_pointers)) {
     const std::vector<Decoration>& decorations = id_decorations(inst.result_id);
     std::map<SpvDecoration, const std::vector<uint32_t>*> decoration_params;
     for (const auto& decoration : decorations) {
-      const auto result = decoration_params.emplace(decoration.dec_type(),
-                                                    &decoration.params());
-      (void)result;
-      assert(result.second);
+      decoration_params[decoration.dec_type()] = &decoration.params();
     }
     for (const auto& kv : decoration_params) {
+      // Append decoration data to the key, sorted by decoration type.
+      // Use magic number to separate unrelated chunks of data.
+      const uint32_t kMagicSeparator = 0xffff9296;
+      key.push_back(kMagicSeparator);
       key.push_back(static_cast<uint32_t>(kv.first));
       key.insert(key.end(), kv.second->begin(), kv.second->end());
     }
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -15,7 +15,6 @@
 #include "val/validation_state.h"
 
 #include <cassert>
-#include <map>
 
 #include "opcode.h"
 #include "val/basic_block.h"
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -426,25 +426,6 @@ bool ValidationState_t::RegisterUniqueTypeDeclaration(
     key.insert(key.end(), inst.words + words_begin, inst.words + words_end);
   }
 
-  // If module has extension SPV_KHR_variable_pointers, then OpTypePointer
-  // is allowed to create identical types so long as their decorations differ.
-  if (inst.opcode == SpvOpTypePointer &&
-      HasExtension(Extension::kSPV_KHR_variable_pointers)) {
-    const std::vector<Decoration>& decorations = id_decorations(inst.result_id);
-    std::map<SpvDecoration, const std::vector<uint32_t>*> decoration_params;
-    for (const auto& decoration : decorations) {
-      decoration_params[decoration.dec_type()] = &decoration.params();
-    }
-    for (const auto& kv : decoration_params) {
-      // Append decoration data to the key, sorted by decoration type.
-      // Use magic number to separate unrelated chunks of data.
-      const uint32_t kMagicSeparator = 0xffff9296;
-      key.push_back(kMagicSeparator);
-      key.push_back(static_cast<uint32_t>(kv.first));
-      key.insert(key.end(), kv.second->begin(), kv.second->end());
-    }
-  }
-
   return unique_type_declarations_.insert(std::move(key)).second;
 }
 }  /// namespace libspirv
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -15,6 +15,7 @@
 #include "val/validation_state.h"
 
 #include <cassert>
+#include <map>
 
 #include "opcode.h"
 #include "val/basic_block.h"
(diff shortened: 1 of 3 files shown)


test/val/val_type_unique_test.cpp
```diff
@@ -238,4 +238,57 @@ OpMemoryModel Physical32 OpenCL
               Not(HasSubstr(GetErrorString(SpvOpTypeVoid))));
 }
 
+TEST_F(ValidateTypeUnique, PointerTypesSameArrayStrideNoExtension) {
```
dneto0: The test name is misleading to me:  The array strides here are different, but the test name is ....SameArrayStride...

source/val/validation_state.cpp
```diff
@@ -425,6 +426,22 @@ bool ValidationState_t::RegisterUniqueTypeDeclaration(
...
 
+  if (inst.opcode == SpvOpTypePointer &&
+      HasExtension(Extension::kSPV_KHR_variable_pointers)) {
+    const std::vector<Decoration>& decorations = id_decorations(inst.result_id);
+    std::map<SpvDecoration, const std::vector<uint32_t>*> decoration_params;
+    for (const auto& decoration : decorations) {
+      const auto result = decoration_params.emplace(decoration.dec_type(),
+                                                    &decoration.params());
+      (void)result;
+      assert(result.second);
```
dneto0: I don't think you should assert here? There's nothing wrong with redundant decorations, except that it wastes space.
I don't think the storage behind id_decorations has unique'd it.

source/val/validation_state.cpp
```diff
@@ -425,6 +426,22 @@ bool ValidationState_t::RegisterUniqueTypeDeclaration(
     key.insert(key.end(), inst.words + words_begin, inst.words + words_end);
   }
 
+  if (inst.opcode == SpvOpTypePointer &&
```
dneto0: Better to have a comment describing the rule you're enforcing.

source/val/validation_state.cpp
dneto0: It's worth saying you're assuming that a single decoration enum can only be set once per target ID.
It think that's true currently, and is likely to be true always.  But I don't recall a SPIR-V validation rule about that. (Perhaps an oversight.)

test/val/val_type_unique_test.cpp
```diff
@@ -238,4 +238,57 @@ OpMemoryModel Physical32 OpenCL
...
+OpDecorate %ptr1 ArrayStride 4
+OpDecorate %ptr2 ArrayStride 8
+%u32 = OpTypeInt 32 0
+%ptr1 = OpTypePointer Input %u32
+%ptr2 = OpTypePointer Input %u32
+)";
+  CompileSuccessfully(str.c_str());
+  ASSERT_EQ(SPV_SUCCESS, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr(GetErrorString(SpvOpTypePointer)));
```
dneto0: I'm confused.  The validation succeeds, but generates an error string?

Does this test even pass?

source/validate_type_unique.cpp
```diff
@@ -49,7 +49,7 @@ spv_result_t TypeUniquePass(ValidationState_t& _,
       // return _.diag(SPV_ERROR_INVALID_DATA)
       return _.diag(SPV_SUCCESS)
           << "Duplicate non-aggregate type declarations are not allowed."
-          << " Opcode: " << inst->opcode;
+          << " Opcode: " << spvOpcodeString(SpvOp(inst->opcode));
```
dneto0: This is an improvement. But in actual practice I've found having the ID generated by this instruction is easier to work with.  It pinpoints the exact instruction at fault.

test/val/val_type_unique_test.cpp
```diff
@@ -238,4 +238,57 @@ OpMemoryModel Physical32 OpenCL
               Not(HasSubstr(GetErrorString(SpvOpTypeVoid))));
 }
 
+TEST_F(ValidateTypeUnique, PointerTypesSameArrayStrideNoExtension) {
+  string str = R"(
+OpCapability Shader
+OpCapability Linkage
+OpMemoryModel Logical GLSL450
+OpDecorate %ptr1 ArrayStride 4
```
dneto0: Or maybe you just have a typo here?

source/val/validation_state.cpp
atgoo: I will assume that decorations are not unique, but only use the last one.

source/val/validation_state.cpp
atgoo: Done.

source/validate_type_unique.cpp
atgoo: Done

test/val/val_type_unique_test.cpp
atgoo: CopyPaste

test/val/val_type_unique_test.cpp
atgoo: It logs a warning but returns SPV_SUCCESS.

source/val/validation_state.cpp
```diff
@@ -425,6 +426,25 @@ bool ValidationState_t::RegisterUniqueTypeDeclaration(
...
+      HasExtension(Extension::kSPV_KHR_variable_pointers)) {
+    const std::vector<Decoration>& decorations = id_decorations(inst.result_id);
+    std::map<SpvDecoration, const std::vector<uint32_t>*> decoration_params;
+    for (const auto& decoration : decorations) {
+      decoration_params[decoration.dec_type()] = &decoration.params();
+    }
+    for (const auto& kv : decoration_params) {
+      // Append decoration data to the key, sorted by decoration type.
+      // Use magic number to separate unrelated chunks of data.
+      const uint32_t kMagicSeparator = 0xffff9296;
```
dneto0: Low level comment:  This magic number mechanism is fragile.  Better to push a kv.second.size() then its contents.

(But really we should not care about duplication of decorations.)

source/val/validation_state.cpp
```diff
@@ -425,6 +426,25 @@ bool ValidationState_t::RegisterUniqueTypeDeclaration(
     key.insert(key.end(), inst.words + words_begin, inst.words + words_end);
   }
 
+  // If module has extension SPV_KHR_variable_pointers, then OpTypePointer
+  // is allowed to create identical types so long as their decorations differ.
```
dneto0: Looking again, this is not the rule.
The SPV_KHR_variable_pointers extension says 
> Pointer types are also allowed to have multiple id's for the same opcode and operands, to allow for differing ArrayStride Array Stride decoration values.

> Also, in that paragraph, "non-aggregate types" will then generally be "non-aggregate non-pointer types."

From 2.8 in the SPIR-V spec:
> It is valid to declare multiple aggregate type <id>s having the same opcode and operands. This is to allow multiple instances of aggregate types with the same structure to be decorated differently. (Different decorations are not required; two different aggregate type ids are allowed to have identical declarations and decorations, and will still be two different types.)

That means the validator should allow distinct OpTypePointer with the same operands, and with the same decorations.

dneto0: So I think all this new code boils down to:
```
if (inst.opcode == SpvOpTypePointer && HasExtension....) return true;
```
Even better:  put the check up in the callee, along with the aggregate types.  That is, modify TypeUniquePass to do the pointer-and-has-extension check.

atgoo:
Ready for review.

dneto0:
Hmm.  Except that the commit message was inaccurate.  I'll fix it on the rebase.

dneto0:
Rebased, squashed, and pushed into master as 725284c2ef8

PR #786, opened 2017-08-30
Add validation pass for arithmetic operations

atgoo:
The pass checks if arithmetic operations (such as OpFMul) receive
correct operands.


dneto0:
I've also checked this against the Vulkan CTS.  Fortunately, the Vulkan CTS shaders continue to pass validation. :-)

dneto0:
Rebased and pushed onto master as 82df4bbd688a

Issue #790, opened 2017-08-31
Inliner: Fails to remap callee block entry ID when splitting for multi-return callee case


dneto0:
This can cause a dominance error in a Phi inside the cloned callee.

I found this by inspection while working on #787 

Example input:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %1 "main"
               OpSource OpenCL_C 120
        %int = OpTypeInt 32 1
      %int_0 = OpConstant %int 0
      %int_1 = OpConstant %int 1
      %int_2 = OpConstant %int 2
      %int_3 = OpConstant %int 3
      %int_4 = OpConstant %int 4
       %void = OpTypeVoid

          %9 = OpTypeFunction %void
         %10 = OpFunction %void None %9
         %11 = OpLabel
         %12 = OpCopyObject %int %int_0
               OpBranch %13
         %13 = OpLabel
         %14 = OpPhi %int %12 %11
         %15 = OpCopyObject %int %int_1
               OpReturn
         %16 = OpLabel
         %17 = OpCopyObject %int %int_2
               OpReturn

               OpFunctionEnd
          %1 = OpFunction %void None %9
         %18 = OpLabel
         %19 = OpCopyObject %int %int_3
         %20 = OpFunctionCall %void %10
         %21 = OpCopyObject %int %int_4
               OpReturn
               OpFunctionEnd
```
The function %9 has multiple returns, and is called from entry point %1.  So when inlining we make a single-trip loop to enclose the cloned callee. 
The result of inlining is:
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %1 "main"
               OpSource OpenCL_C 120
        %int = OpTypeInt 32 1
      %int_0 = OpConstant %int 0
      %int_1 = OpConstant %int 1
      %int_2 = OpConstant %int 2
      %int_3 = OpConstant %int 3
      %int_4 = OpConstant %int 4
       %void = OpTypeVoid
          %9 = OpTypeFunction %void
       %bool = OpTypeBool
      %false = OpConstantFalse %bool
         %10 = OpFunction %void None %9
         %11 = OpLabel
         %12 = OpCopyObject %int %int_0
               OpBranch %13
         %13 = OpLabel
         %14 = OpPhi %int %12 %11
         %15 = OpCopyObject %int %int_1
               OpReturn
         %16 = OpLabel
[...]

PR #791, opened 2017-08-31
Inliner: Remap callee entry block id to single-trip loop header

dneto0:
Otherwise cloned phis can be invalid.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/790


dneto0:
Hi @greg-lunarg : I'd like your feedback.

dneto0:
Latest commit fixes the comment to emphasize that the mapping for the callee's entry block is being *re*-set here.

dneto0:
Rebased to pick up the test fixes for the newly aliased names.

dneto0:
Rebased

PR #793, opened 2017-09-01
Make enums for all currently published extensions

dneto0:
Use the list from the SPIR-V registry page.  Also, capture it as
a string so it's much easier to update via copy-paste.

The validator will accept modules that declare these known
extensions.  However, we might not know about new tokens or
instructions declared in them.  For that we need grammar updates
applied to SPIRV-Headers.


dneto0:
Rebased and pushed into master as 0d3b8329a43

PR #794, opened 2017-09-01
Fix mingw build (source/print.cpp)

dneto0:
source/print.cpp doesn't compile due to integer conversion.

Tested by @dneto0 on a Windows machine.


dneto0:
This is my adaptation of #779 
I reviewed it there and have modified it to work, and tested it manually on Windows.

LGTM

Issue #796, opened 2017-09-01
validator:  OpModuleProcessed should be usable anywhere in the module (1.1 and later)


dneto0:
Reported by @sheredom 

Indeed, the SPIR-V spec places no restrictions on where it should be allowed.
The validator is complaining about it anywhere other than within a basic block.

dneto0:
Closing. SPIR WG decided OpModuleProcessed belongs in its own layout section 7c.
That's covered by https://github.com/KhronosGroup/SPIRV-Tools/pull/812

PR #797, opened 2017-09-01
WIP:  validator: OpModuleProcessed allowed anywhere, for 1.1 and after

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/796

### DIFF
--- a/test/val/val_layout_test.cpp
+++ b/test/val/val_layout_test.cpp
@@ -34,8 +34,9 @@ using std::tie;
 using std::tuple;
 using std::vector;
 
-using ::testing::StrEq;
+using ::testing::Eq;
 using ::testing::HasSubstr;
+using ::testing::StrEq;
 using libspirv::spvResultToString;
 
 using pred_type = function<spv_result_t(int)>;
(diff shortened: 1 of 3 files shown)


test/val/val_layout_test.cpp
```diff
@@ -491,5 +492,54 @@ TEST_F(ValidateEntryPoint, NoEntryPointWithLinkageCapGood) {
...
+  EXPECT_THAT(getDiagnosticString(), Eq(""));
+}
+
+TEST_F(ValidateModuleProcessed, OkAnywhereInSPIRV12) {
+  CompileSuccessfully(GetModuleWithModuleProcessed(), SPV_ENV_UNIVERSAL_1_1);
+  EXPECT_EQ(SPV_SUCCESS, ValidateInstructions(SPV_ENV_UNIVERSAL_1_2));
+  EXPECT_THAT(getDiagnosticString(), Eq(""));
+}
+
+TEST_F(ValidateModuleProcessed, FailsIn10) {
```
sheredom: Does this fail on the first `OpModuleProcessed`, or any of them?

Since `OpModuleProcessed` isn't a valid 1.0 opcode I just want to be sure that its invalid everywhere for 1.0, and not just inbetween the `OpFunction` & `OpLabel`.

dneto0:
I'm confirming with the SPIR WG that this is the right behaviour.

dneto0:
SPIR WG agreed to restrict OpModuleProcessed to a "7c" section of the layout.  So this is the wrong fix.

dneto0:
Withdrawing this PR.

Issue #800, opened 2017-09-02
inliner: fails to move OpLoopMerge to top of loop when inlining multi-block function into loop header


dneto0:
This is like #787 but it applied to loops that aren't just single-block loops. It happens in general.

Example:
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpSource OpenCL_C 120

       %bool = OpTypeBool
       %true = OpConstantTrue %bool
        %int = OpTypeInt 32 1
       %int_1 = OpConstant %int 1  ; these constants are used as markers
       %int_2 = OpConstant %int 2
       %int_3 = OpConstant %int 3
       %int_4 = OpConstant %int 4
       %int_5 = OpConstant %int 5
        %void = OpTypeVoid
      %voidfn = OpTypeFunction %void

        %foo = OpFunction %void None %voidfn
   %fooentry = OpLabel
         %c1 = OpCopyObject %int %int_1
               OpBranch %fooexit
    %fooexit = OpLabel
         %c2 = OpCopyObject %int %int_2
               OpReturn
               OpFunctionEnd

    %main = OpFunction %void None %voidfn
      %entry = OpLabel
               OpBranch %loop

       %loop = OpLabel
         %c3 = OpCopyObject %int %int_3
        %nil = OpFunctionCall %void %foo
         %c4 = OpCopyObject %int %int_4
               OpLoopMerge %merge %body None
               OpBranchConditional %true %body %merge

       %body = OpLabel
         %c5 = OpCopyObject %int %int_5
               OpBranchConditional %true %loop %merge

      %merge = OpLabel
               OpReturn
               OpFunctionEnd
```

The salient bits are:
- Callee function foo has two blocks. So it will generate multiple blocks when inlined
- The call to foo occurs in the loop header block:  between its Label and LoopMerge instructions
- This example is different from #787 because in this case the loop header branches forward into a new body block rather than just looping back. 

Inlining produces:
```
              OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %1 "main"
               OpSource OpenCL_C 120
       %bool = OpTypeBool
       %true = OpConstantTrue %bool
        %int = OpTypeInt 32 1
      %int_1 = OpConstant %int 1
      %int_2 = OpConstant %int 2
      %int_3 = OpConstant %int 3
      %int_4 = OpConstant %int 4
[...]

PR #801, opened 2017-09-02
Inliner: Fix LoopMerge when inline into loop header of multi block loop

dneto0:
This adapts the fix for the single-block loop.  Split the loop like
before.  But when we move the OpLoopMerge back to the loop header,
redirect the continue target only when the original loop was a single
block loop.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/800


dneto0:
Another review request for @greg-lunarg.
This generalizes the single-block-loop fix from before.

dneto0:
Rebased and added CHANGES entry.

PR #809, opened 2017-09-06
Updated markv_autogen

atgoo:
- now includes a table of all descriptors with coding scheme
(improves performance by 5% by allowing to avoid creation of
move-to-front sequences which will never be used)
- increased the size of markv_autogen.inc, clang doesn't seem
to have the long compilation time problem now
(probably was inadvertently fixed by using Huffman codec
serialization)


dneto0:
Rebased, squashed, and pushed into master as 9b14dd0cb453

PR #810, opened 2017-09-07
CFG Cleanup pass - Remove unreachable blocks.

dnovillo:
This change adds a new CFG cleanup pass to spirv-opt.  The intent of the pass is to act as a repository of a variety of common CFG cleanup actions (straightening, removal of unreachable code, etc).

It's my first change to SPIRV-Tools, so I'm sure it's full of inconsistencies and missing bits.  One of the things I've been thinking is that it may make sense to coalesce other existing CFG cleanup passes into this one in the hopes of reducing the number of CFG/IR traversals needed during cleanup.

Or, perhaps, it's preferable to keep these passes separate.  I understand that compile time is not an issue now, but I'm not sure if this might be challenging in the future (in my experience, this tends to be true at some point).

The code works and passes the trivial test I've added.  I need to add more, but I want a first review pass so I can learn preferred idioms and ways of structuring the code.

Thanks.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -393,6 +393,14 @@ Optimizer::PassToken CreateCompactIdsPass();
 // Creates a remove duplicate capabilities pass.
 Optimizer::PassToken CreateRemoveDuplicatesPass();
 
+// Creates a CFG cleanup pass.
+// This pass removes cruft from the control flow graph of functions that are
+// reachable from entry points and exported functions. It currently includes the
+// following functionality:
+//
+// - Removal of unreachable basic blocks.
+Optimizer::PassToken CreateCFGCleanupPass();
+
 }  // namespace spvtools
 
 #endif  // SPIRV_TOOLS_OPTIMIZER_HPP_
--- a/source/opt/cfg_cleanup_pass.cpp
+++ b/source/opt/cfg_cleanup_pass.cpp
@@ -1,0 +1,287 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file implements a pass to cleanup the CFG to remove superfluous
+// constructs (e.g., unreachable basic blocks, empty control flow structures,
+// etc)
+
+#include <queue>
+#include <unordered_set>
+
+#include "cfg_cleanup_pass.h"
+
+#include "function.h"
+#include "module.h"
+
+namespace spvtools {
+namespace opt {
+
+uint32_t CFGCleanupPass::TypeToUndef(uint32_t type_id) {
+  const auto uitr = type2undefs_.find(type_id);
+  if (uitr != type2undefs_.end()) {
+    return uitr->second;
+  }
+
+  const uint32_t undefId = TakeNextId();
+  std::unique_ptr<ir::Instruction> undef_inst(
+      new ir::Instruction(SpvOpUndef, type_id, undefId, {}));
+  def_use_mgr_->AnalyzeInstDefUse(&*undef_inst);
+  module_->AddGlobalValue(std::move(undef_inst));
+  type2undefs_[type_id] = undefId;
+
+  return undefId;
+}
+
+// Remove all |phi| operands coming from unreachable blocks (i.e., blocks not in
+// |reachable_blocks|).  There are two types of removal that this function can
+// perform:
+//
+// 1- Any operand that comes directly from an unreachable block is completely
+//    removed.  Since the block is unreachable, the edge between the unreachable
+//    block and the block holding |phi| has been removed.
+//
+// 2- Any operand that comes via a live block and was defined at an unreachable
+//    block gets its value replaced with an OpUndef value. Since the argument
+//    was generated in an unreachable block, it no longer exists, so it cannot
+//    be referenced.  However, since the value does not reach |phi| directly
+//    from the unreachable block, the operand cannot be removed from |phi|.
+//    Therefore, we replace the argument value with OpUndef.
+//
+// For example, in the switch() below, assume that we want to remove the
+// argument with value %11 coming from block %41.
+//
+//          [ ... ]
+//          %41 = OpLabel                    <--- Unreachable block
+//          %11 = OpLoad %int %y
+//          [ ... ]
+//                OpSelectionMerge %16 None
+//                OpSwitch %12 %16 10 %13 13 %14 18 %15
+//          %13 = OpLabel
+//                OpBranch %16
+//          %14 = OpLabel
+//                OpStore %outparm %int_14
+//                OpBranch %16
+//          %15 = OpLabel
+//                OpStore %outparm %int_15
+//                OpBranch %16
+//          %16 = OpLabel
+//          %30 = OpPhi %int %11 %41 %int_42 %13 %11 %14 %11 %15
+//
+// Since %41 is now an unreachable block, the first operand of |phi| needs to
+// be removed completely.  But the operands (%11 %14) and (%11 %15) cannot be
+// removed because %14 and %15 are reachable blocks.  Since %11 no longer exist,
+// in those arguments, we replace all references to %11 with an OpUndef value.
+// This results in |phi| looking like:
+//
+//           %50 = OpUndef %int
+//           [ ... ]
+//           %30 = OpPhi %int %int_42 %13 %50 %14 %50 %15
+void CFGCleanupPass::RemovePhiOperands(
+    ir::Instruction* phi,
+    std::unordered_set<ir::BasicBlock*> reachable_blocks) {
+  std::vector<ir::Operand> keep_operands;
+  uint32_t type_id = 0;
+  uint32_t undef_id = 0;
+
+  // Traverse all the operands in |phi|. Build the new operand vector by adding
+  // all the original operands from |phi| except the unwanted ones.
+  bool undef_created = false;
+  for (uint32_t i = 0; i < phi->NumOperands();) {
+    if (i < 2) {
+      // The first two arguments are always preserved.
+      keep_operands.push_back(phi->GetOperand(i));
+      ++i;
+      continue;
+    }
+
+    // The remaining Phi arguments come in pairs. Index 'i' contains the
+    // variable id, index 'i + 1' is the originating block id.
+    assert(i % 2 == 0 && i < phi->NumOperands() - 1 &&
+           "malformed Phi arguments");
+
+    ir::BasicBlock *in_block = label2block_[phi->GetSingleWordOperand(i + 1)];
+    if (reachable_blocks.find(in_block) == reachable_blocks.end()) {
+      // If the incoming block is unreachable, remove both operands as this
+      // means that the |phi| has lost an incoming edge.
+      i += 2;
+      continue;
+    }
+
+    // In all other cases, the operand must be kept but may need to be changed.
+    uint32_t arg_id = phi->GetSingleWordOperand(i);
+    ir::BasicBlock *def_block = def_block_[arg_id];
+    if (def_block &&
+        reachable_blocks.find(def_block_[arg_id]) == reachable_blocks.end()) {
+      // If the current |phi| argument was defined in an unreachable block, it
+      // means that this |phi| argument is no longer defined. Replace it with
+      // |undef_id|.
+      if (!undef_created) {
+        type_id = def_use_mgr_->GetDef(arg_id)->type_id();
+        undef_id = TypeToUndef(type_id);
+        undef_created = true;
+      }
+      keep_operands.push_back(
+          ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {undef_id}));
+    } else {
+      // Otherwise, the argument comes from a reachable block or from no block
+      // at all (meaning that it was defined in the global section of the
+      // program).  In both cases, keep the argument intact.
+      keep_operands.push_back(phi->GetOperand(i));
+    }
+
+    keep_operands.push_back(phi->GetOperand(i + 1));
+
+    i += 2;
+  }
+
+  phi->ReplaceOperands(keep_operands);
+}
+
+void CFGCleanupPass::RemoveBlock(ir::Function::iterator* bi) {
+  auto& rm_block = **bi;
+
+  // Remove instructions from the block.
+  rm_block.ForEachInst([&rm_block, this](ir::Instruction* inst) {
+    // Note that we do not kill the block label instruction here. The label
+    // instruction is needed to identify the block, which is needed by the
+    // removal of phi operands.
+    if (inst != rm_block.GetLabelInst()) {
+      KillNamesAndDecorates(inst);
+      def_use_mgr_->KillInst(inst);
+    }
+  });
+
+  // Remove the label instruction last.
+  auto label = rm_block.GetLabelInst();
+  KillNamesAndDecorates(label);
+  def_use_mgr_->KillInst(label);
+
+  *bi = bi->Erase();
+}
+
+bool CFGCleanupPass::RemoveUnreachableBlocks(ir::Function* func) {
+  bool modified = false;
+
+  // Mark reachable all blocks reachable from the function's entry block.
+  std::unordered_set<ir::BasicBlock*> reachable_blocks;
+  std::unordered_set<ir::BasicBlock*> visited_blocks;
+  std::queue<ir::BasicBlock*> worklist;
+  reachable_blocks.insert(func->entry().get());
+
+  // Initially mark the function entry point as reachable.
+  worklist.push(func->entry().get());
+
+  auto mark_reachable = [&reachable_blocks, &visited_blocks, &worklist,
+                         this](uint32_t label_id) {
+    auto successor = label2block_[label_id];
+    if (visited_blocks.count(successor) == 0) {
+      reachable_blocks.insert(successor);
+      worklist.push(successor);
+      visited_blocks.insert(successor);
+    }
+  };
+
+  // Transitively mark all blocks reachable from the entry as reachable.
+  while (!worklist.empty()) {
+    ir::BasicBlock* block = worklist.front();
+    worklist.pop();
+
+    // All the successors of a live block are also live.
+    block->ForEachSuccessorLabel(mark_reachable);
+
+    // All the Merge and ContinueTarget blocks of a live block are also live.
+    block->ForMergeAndContinueLabel(mark_reachable);
+  }
+
+  // Update operands of Phi nodes that reference unreachable blocks.
+  for (auto& block : *func) {
+    // If the block is about to be removed, don't bother updating its
+    // Phi instructions.
+    if (reachable_blocks.count(&block) == 0) {
+      continue;
+    }
+
+    // If the block is reachable and has Phi instructions, remove all
+    // operands from its Phi instructions that reference unreachable blocks.
+    if (block.HasPhiInstructions()) {
+      block.ForEachPhiInst(
+          [&block, &reachable_blocks, this](ir::Instruction* phi) {
+            RemovePhiOperands(phi, reachable_blocks);
+          });
+    }
+  }
+
+  // Erase unreachable blocks.
+  for (auto ebi = func->begin(); ebi != func->end();) {
+    if (reachable_blocks.count(&*ebi) == 0) {
+      RemoveBlock(&ebi);
+      modified = true;
+    } else {
+      ++ebi;
+    }
+  }
+
+  return modified;
+}
+
+bool CFGCleanupPass::CFGCleanup(ir::Function* func) {
+  bool modified = false;
+  modified |= RemoveUnreachableBlocks(func);
+  return modified;
+}
+
+void CFGCleanupPass::Initialize(ir::Module* module) {
+  // Initialize the DefUse manager. TODO(dnovillo): Re-factor all this into the
+  // module or some other context class for the optimizer.
+  module_ = module;
+  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+  FindNamedOrDecoratedIds();
+
+  // Initialize next unused Id. TODO(dnovillo): Re-factor into the module or
+  // some other context class for the optimizer.
+  next_id_ = module_->id_bound();
+
+  // Initialize block lookup map.
+  label2block_.clear();
+  for (auto& fn : *module) {
+    for (auto& block : fn) {
+      label2block_[block.id()] = &block;
+
+      // Build a map between SSA names to the block they are defined in.
+      // TODO(dnovillo): This is expensive and unnecessary if ir::Instruction
+      // instances could figure out what basic block they belong to. Remove this
+      // once this is possible.
+      block.ForEachInst([this, &block](ir::Instruction* inst) {
+        uint32_t result_id = inst->result_id();
+        if (result_id > 0) {
+          def_block_[result_id] = &block;
+        }
+      });
+    }
+  }
+}
+
+Pass::Status CFGCleanupPass::Process(ir::Module* module) {
+  Initialize(module);
+
+  // Process all entry point functions.
+  ProcessFunction pfn = [this](ir::Function* fp) { return CFGCleanup(fp); };
+  bool modified = ProcessReachableCallTree(pfn, module);
+  FinalizeNextId(module_);
+  return modified ? Pass::Status::SuccessWithChange
+                  : Pass::Status::SuccessWithoutChange;
+}
+
+}  // namespace opt
+}  // namespace spvtools
--- a/source/opt/cfg_cleanup_pass.h
+++ b/source/opt/cfg_cleanup_pass.h
@@ -1,0 +1,88 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_OPT_CFG_CLEANUP_PASS_H_
+#define LIBSPIRV_OPT_CFG_CLEANUP_PASS_H_
+
+#include "function.h"
+#include "mem_pass.h"
+#include "module.h"
+
+namespace spvtools {
+namespace opt {
+
+class CFGCleanupPass : public MemPass {
+ public:
+  CFGCleanupPass() = default;
+  const char* name() const override { return "cfg-cleanup"; }
+  Status Process(ir::Module*) override;
+
+ private:
+  // Call all the cleanup helper functions on |func|.
+  bool CFGCleanup(ir::Function* func);
+
+  // Remove all the unreachable basic blocks in |func|.
+  bool RemoveUnreachableBlocks(ir::Function* func);
+
+  // Remove the block pointed by the iterator |*bi|. This also removes
+  // all the instructions in the pointed-to block.
+  void RemoveBlock(ir::Function::iterator* bi);
+
+  // Initialize the pass.
+  void Initialize(ir::Module* module);
+
+  // Remove Phi operands in |phi| that are coming from blocks not in
+  // |reachable_blocks|.
+  void RemovePhiOperands(ir::Instruction* phi,
+                         std::unordered_set<ir::BasicBlock*> reachable_blocks);
+
+  // Return the next available Id and increment it. TODO(dnovillo): Refactor
+  // into a new type pool manager to be used for all passes.
+  inline uint32_t TakeNextId() { return next_id_++; }
+
+  // Save next available id into |module|. TODO(dnovillo): Refactor
+  // into a new type pool manager to be used for all passes.
+  inline void FinalizeNextId(ir::Module* module) {
+    module->SetIdBound(next_id_);
+  }
+
+  // Return undef in function for type. Create and insert an undef after the
+  // first non-variable in the function if it doesn't already exist. Add
+  // undef to function undef map. TODO(dnovillo): Refactor into a new
+  // type pool manager to be used for all passes.
+  uint32_t TypeToUndef(uint32_t type_id);
+
+  // Map from block's label id to block. TODO(dnovillo): Basic blocks ought to
+  // have basic blocks in their pred/succ list.
+  std::unordered_map<uint32_t, ir::BasicBlock*> label2block_;
+
+  // Map from an instruction result ID to the block that holds it.
+  // TODO(dnovillo): This would be unnecessary if ir::Instruction instances
+  // knew what basic block they belong to.
+  std::unordered_map<uint32_t, ir::BasicBlock*> def_block_;
+
+  // Map from type to undef values. TODO(dnovillo): This is replicated from
+  // class LocalMultiStoreElimPass. It should be refactored into a type
+  // pool manager.
+  std::unordered_map<uint32_t, uint32_t> type2undefs_;
+
+  // Next unused ID. TODO(dnovillo): Refactor this to some common utility class.
+  // Seems to be implemented in very many passes.
+  uint32_t next_id_;
+};
+
+}  // namespace opt
+}  // namespace spvtools
+
+#endif
--- a/test/opt/cfg_cleanup_test.cpp
+++ b/test/opt/cfg_cleanup_test.cpp
@@ -1,0 +1,448 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "pass_fixture.h"
+#include "pass_utils.h"
+
+namespace {
+
+using namespace spvtools;
+
+using CFGCleanupTest = PassTest<::testing::Test>;
+
+TEST_F(CFGCleanupTest, RemoveUnreachableBlocks) {
+  const std::string declarations = R"(OpCapability Shader
+%1 = OpExtInstImport "GLSL.std.450"
+OpMemoryModel Logical GLSL450
+OpEntryPoint Fragment %main "main" %inf %outf4
+OpExecutionMode %main OriginUpperLeft
+OpSource GLSL 450
+OpName %main "main"
+OpName %inf "inf"
+OpName %outf4 "outf4"
+OpDecorate %inf Location 0
+OpDecorate %outf4 Location 0
+%void = OpTypeVoid
+%6 = OpTypeFunction %void
+%float = OpTypeFloat 32
+%_ptr_Input_float = OpTypePointer Input %float
+%inf = OpVariable %_ptr_Input_float Input
+%float_2 = OpConstant %float 2
+%bool = OpTypeBool
+%v4float = OpTypeVector %float 4
+%_ptr_Output_v4float = OpTypePointer Output %v4float
+%outf4 = OpVariable %_ptr_Output_v4float Output
+%float_n0_5 = OpConstant %float -0.5
+)";
+
+  const std::string body_before = R"(%main = OpFunction %void None %6
+%14 = OpLabel
+OpSelectionMerge %17 None
+OpBranch %18
+%19 = OpLabel
+%20 = OpLoad %float %inf
+%21 = OpCompositeConstruct %v4float %20 %20 %20 %20
+OpStore %outf4 %21
+OpBranch %17
+%18 = OpLabel
+%22 = OpLoad %float %inf
+%23 = OpFAdd %float %22 %float_n0_5
+%24 = OpCompositeConstruct %v4float %23 %23 %23 %23
+OpStore %outf4 %24
+OpBranch %17
+%17 = OpLabel
+OpReturn
+OpFunctionEnd
+)";
+
+  const std::string body_after = R"(%main = OpFunction %void None %6
+%14 = OpLabel
+OpSelectionMerge %15 None
+OpBranch %16
+%16 = OpLabel
+%20 = OpLoad %float %inf
+%21 = OpFAdd %float %20 %float_n0_5
+%22 = OpCompositeConstruct %v4float %21 %21 %21 %21
+OpStore %outf4 %22
+OpBranch %15
+%15 = OpLabel
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::CFGCleanupPass>(
+      declarations + body_before, declarations + body_after, true, true);
+}
+
+TEST_F(CFGCleanupTest, RemoveDecorations) {
+  const std::string before = R"(
+                       OpCapability Shader
+                  %1 = OpExtInstImport "GLSL.std.450"
+                       OpMemoryModel Logical GLSL450
+                       OpEntryPoint Fragment %main "main"
+                       OpName %main "main"
+                       OpName %x "x"
+                       OpName %dead "dead"
+                       OpDecorate %x RelaxedPrecision
+                       OpDecorate %dead RelaxedPrecision
+               %void = OpTypeVoid
+                  %6 = OpTypeFunction %void
+              %float = OpTypeFloat 32
+%_ptr_Function_float = OpTypePointer Function %float
+            %float_2 = OpConstant %float 2
+            %float_4 = OpConstant %float 4
+
+               %main = OpFunction %void None %6
+                 %14 = OpLabel
+                  %x = OpVariable %_ptr_Function_float Function
+                       OpSelectionMerge %17 None
+                       OpBranch %18
+                 %19 = OpLabel
+               %dead = OpVariable %_ptr_Function_float Function
+                       OpStore %dead %float_2
+                       OpBranch %17
+                 %18 = OpLabel
+                       OpStore %x %float_4
+                       OpBranch %17
+                 %17 = OpLabel
+                       OpReturn
+                       OpFunctionEnd
+)";
+
+  const std::string after = R"(OpCapability Shader
+%1 = OpExtInstImport "GLSL.std.450"
+OpMemoryModel Logical GLSL450
+OpEntryPoint Fragment %main "main"
+OpName %main "main"
+OpName %x "x"
+OpDecorate %x RelaxedPrecision
+%void = OpTypeVoid
+%6 = OpTypeFunction %void
+%float = OpTypeFloat 32
+%_ptr_Function_float = OpTypePointer Function %float
+%float_2 = OpConstant %float 2
+%float_4 = OpConstant %float 4
+%main = OpFunction %void None %6
+%11 = OpLabel
+%x = OpVariable %_ptr_Function_float Function
+OpSelectionMerge %12 None
+OpBranch %13
+%13 = OpLabel
+OpStore %x %float_4
+OpBranch %12
+%12 = OpLabel
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::CFGCleanupPass>(before, after, true, true);
+}
+
+
+TEST_F(CFGCleanupTest, UpdatePhis) {
+  const std::string before = R"(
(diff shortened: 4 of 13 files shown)


dneto0:
I haven't looked at the code yet, but am curious about "straightening".
Something that has been requested is:
 - SPIR-V has a rule requiring that if block A dominates block B, then A appears in the module before B.
 - It would be nice if we had a pass that enforced:  If B post-dominates A, then B appears after A.
 - Furthermore, it would be nice if, when structured control flow is in effect, the continue construct for a loop appears after the loop construct for the loop. (See section 2.11 for those definitions)

greg-lunarg:
A few notes on the most recent comment on straightening by @dneto0:

- Another order "nicety" (I think) would be to have the merge block follow all blocks in the control-flow construct  and continue construct for which it is the merge block.

- There is code in spirv-opt that figures out an ordering that satisfies this requirement plus those from @dneto0. Look for ComputeStructuredOrder(). Note that there are several copies laying around. I think these can (and should) be consolidated into one at some point.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,88 @@
...
+          auto successor = label2block_[label_id];
+          reachable_blocks.insert(successor);
+          worklist.push(successor);
+        });
+  }
+
+  // Erase unreachable blocks.
+  for (auto ebi = func->begin(); ebi != func->end();)
+    if (reachable_blocks.find(&*ebi) == reachable_blocks.end()) {
+      ebi = ebi.Erase();
```
dneto0: You're too used to LLVM.  Remember that in the current ir::* class hierarchy, instructions only have IDs for the things they use, and the DefUseManager is a separate layer over it.

So deleting an Instruction will not delete the references to that instruction.
Thinks you'll have to clean up:
 - OpName,  possibly OpMemberName (for struct types, so does not apply in this pass)
 - OpDecoration, OpDecorationGroup
 - OpPhi, e.g. you delete A, and so all values defined in A, but that block and its values might be used in an OpPhi in a live block.

Use DefUseManager to trace uses.  It has a convenience KillInst that converts an existing instruction to a Nop and kill the uses.  That'a a common idiom you'll see:  KillInst, then the Module's write-to-output function can be used to skip over all nops.  It works.  However, cleaning up blocks is another matter.   Check Greg's dead-branch-elim code for examples of how to clean up blocks.

And then we have to agree on whether this pass promises to keep up the DefUseManager up to date. By default it does not have to make that promise.  But our direction is to "save time" by agreeing to keep it up to date, and then have the pass manager track whether the DefUseManager is stale or not.

include/spirv-tools/optimizer.hpp
```diff
@@ -393,6 +393,13 @@ Optimizer::PassToken CreateAggressiveDCEPass();
 // The pass remaps result ids to a compact and gapless range starting from %1.
 Optimizer::PassToken CreateCompactIdsPass();
 
+// Creates a CFG cleanup pass.
+// This pass removes cruft from the control flow graph. It currently includes
+// the following functionality:
```
dneto0: The pass only operates on functions reachable from some entry point.  Should say so.

test/opt/cfg_cleanup_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+OpStore %outf4 %22
+OpBranch %15
+%15 = OpLabel
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::CFGCleanupPass>(
+      declarations + body_before, declarations + body_after, true, true);
+}
```
dneto0: Need cases for:
 - showing cleanup of OpName and OpDecoration, OpGroupDecorate
 - showing cleanup of OpPhis in live blocks that remain
 - keeping unreachable merge block
 - keeping unreachable continue target, and the loop backedge it dominates even if it's not in the same block
 - in these latter two cases: non-trivial defs and uses inside those unreachable blocks

tools/opt/opt.cpp
```diff
@@ -112,9 +112,12 @@ NOTE: The optimizer is a work in progress.
                Join two blocks into a single block if the second has the
                first as its only predecessor. Performed only on entry point
                call tree functions.
-  -h, --help   
+  --cfg-cleanup
+               Cleanup the control flow graph. This will remove any unnecessary
+               code from the CFG like unreachable code.
```
dneto0: .... "Peformed only on entry point call tree functions"

dneto0:
I think this PR is still live but needs feedback addressed.

dneto0:
Also, please rebase to avoid conflicts.

dnovillo:
On Wed, Sep 20, 2017 at 4:23 PM, David Neto <notifications@github.com>
wrote:

> I think this PR is still live but needs feedback addressed.
>

It is, but I've been traveling and distracted with shiny things.  Will
respond soon.

include/spirv-tools/optimizer.hpp
dnovillo: Done.

dnovillo:
On Fri, Sep 8, 2017 at 1:23 PM, David Neto <notifications@github.com> wrote:

> You also have to preserve any Merge block or ContinueTarget mentioned by
> an OpLoopMerge or OpSelectionMerge instruction, even if they are
> unreachable by control flow.
>
> BasicBlock::ForMergeAndContinueLabel is your friend here.
>

Thanks.  Done.


> Now, if the merge block is unreachable, then
>
>    - you can delete that merge blocks' instructions. (Reasonable to defer
>    to a
>
> ... future patch?


>
>    - the merge block's own CFG successors might not be reachable either.
>    Similar for the ContinueTarget, except that you must also preserve the
>    loop backedge, and that loop backedge might not be in the same block as the
>    ContinueTarget.
>
> But if they are not reachable, then they would be caught by the other
logic.  Perhaps I'm just not following you here.


>
>
>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r137830811>
> :
>
> > +bool CFGCleanupPass::RemoveUnreachableBlocks(ir::Function* func) {
> +  bool modified = false;
> +
> +  // Mark reachable all blocks reachable from the function's entry block.
> +  std::unordered_set<ir::BasicBlock*> reachable_blocks;
> +  std::queue<ir::BasicBlock*> worklist;
> +  reachable_blocks.insert(func->entry().get());
> +  worklist.push(func->entry().get());
> +  while (!worklist.empty()) {
> +    ir::BasicBlock* block = worklist.front();
> +    worklist.pop();
> +    block->ForEachSuccessorLabel(
> +        [&reachable_blocks, &worklist, this](uint32_t label_id) {
> +          auto successor = label2block_[label_id];
> +          reachable_blocks.insert(successor);
> +          worklist.push(successor);
>
> I think you should only push it onto the worklist if it wasn't already in
> the reachable blocks set.
>
> Does this code terminate when it sees a loop?
>
It doesn't.  Fixed.

------------------------------
>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r137831660>
> :
>
> > +  reachable_blocks.insert(func->entry().get());
> +  worklist.push(func->entry().get());
> +  while (!worklist.empty()) {
> +    ir::BasicBlock* block = worklist.front();
> +    worklist.pop();
> +    block->ForEachSuccessorLabel(
> +        [&reachable_blocks, &worklist, this](uint32_t label_id) {
> +          auto successor = label2block_[label_id];
> +          reachable_blocks.insert(successor);
> +          worklist.push(successor);
> +        });
> +  }
> +
> +  // Erase unreachable blocks.
> +  for (auto ebi = func->begin(); ebi != func->end();)
> +    if (reachable_blocks.find(&*ebi) == reachable_blocks.end()) {
>
> Some prefer using the count method instead of check against end iterator.
>
> if (reachable.count(&*ebi))) { ++ebi ; }  else {ebi = ebi.Erase(); modified=true;}
>
>
Done.


>  ------------------------------
>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r137831813>
> :
>
> > +    worklist.pop();
> +    block->ForEachSuccessorLabel(
> +        [&reachable_blocks, &worklist, this](uint32_t label_id) {
> +          auto successor = label2block_[label_id];
> +          reachable_blocks.insert(successor);
> +          worklist.push(successor);
> +        });
> +  }
> +
> +  // Erase unreachable blocks.
> +  for (auto ebi = func->begin(); ebi != func->end();)
> +    if (reachable_blocks.find(&*ebi) == reachable_blocks.end()) {
> +      ebi = ebi.Erase();
> +      modified = true;
> +    } else {
> +      ebi++;
>
> Always prefer pre-order increment/decrement on non-primitive types.
> Iterators in particular. This saves a copy.
>
It doesn't, given the implementation of the iterator classes, but changed
anyway (and nowadays, I do expect a good optimizing compiler to figure this
one out on its own, but it's not always the case ;).


>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r137843639>
> :
>
> > +  worklist.push(func->entry().get());
> +  while (!worklist.empty()) {
> +    ir::BasicBlock* block = worklist.front();
> +    worklist.pop();
> +    block->ForEachSuccessorLabel(
> +        [&reachable_blocks, &worklist, this](uint32_t label_id) {
> +          auto successor = label2block_[label_id];
> +          reachable_blocks.insert(successor);
> +          worklist.push(successor);
> +        });
> +  }
> +
> +  // Erase unreachable blocks.
> +  for (auto ebi = func->begin(); ebi != func->end();)
> +    if (reachable_blocks.find(&*ebi) == reachable_blocks.end()) {
> +      ebi = ebi.Erase();
>
> You're too used to LLVM. Remember that in the current ir::* class
> hierarchy, instructions only have IDs for the things they use, and the
> DefUseManager is a separate layer over it.
>
Gah!  Yes.  So, this is something that would be  awesome if we can change.
Some ideas coming up in the next few days.

So deleting an Instruction will not delete the references to that
> instruction.
> Thinks you'll have to clean up:
>
>    - OpName, possibly OpMemberName (for struct types, so does not apply
>    in this pass)
>    - OpDecoration, OpDecorationGroup
>    - OpPhi, e.g. you delete A, and so all values defined in A, but that
>    block and its values might be used in an OpPhi in a live block.
>
> Use DefUseManager to trace uses. It has a convenience KillInst that
> converts an existing instruction to a Nop and kill the uses. That'a a
> common idiom you'll see: KillInst, then the Module's write-to-output
> function can be used to skip over all nops. It works. However, cleaning up
> blocks is another matter. Check Greg's dead-branch-elim code for examples
> of how to clean up blocks.
>
Thanks.  Done.


> And then we have to agree on whether this pass promises to keep up the
> DefUseManager up to date. By default it does not have to make that promise.
> But our direction is to "save time" by agreeing to keep it up to date, and
> then have the pass manager track whether the DefUseManager is stale or not.
>
OK.   This is part of another conversation, but I think we can make some
improvements here that will mean not having to update all these satellite
data structures when doing basic operations like removing instructions,
edges or basic blocks.

------------------------------
>
> In include/spirv-tools/optimizer.hpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r137844149>
> :
>
> > @@ -393,6 +393,13 @@ Optimizer::PassToken CreateAggressiveDCEPass();
>  // The pass remaps result ids to a compact and gapless range starting from %1.
>  Optimizer::PassToken CreateCompactIdsPass();
>
> +// Creates a CFG cleanup pass.
> +// This pass removes cruft from the control flow graph. It currently includes
> +// the following functionality:
>
> The pass only operates on functions reachable from some entry point.
> Should say so.
>

Done.


>
> In test/opt/cfg_cleanup_test.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r137844637>
> :
>
> > +OpSelectionMerge %15 None
> +OpBranch %16
> +%16 = OpLabel
> +%20 = OpLoad %float %inf
> +%21 = OpFAdd %float %20 %float_n0_5
> +%22 = OpCompositeConstruct %v4float %21 %21 %21 %21
> +OpStore %outf4 %22
> +OpBranch %15
> +%15 = OpLabel
> +OpReturn
> +OpFunctionEnd
> +)";
> +
> +  SinglePassRunAndCheck<opt::CFGCleanupPass>(
> +      declarations + body_before, declarations + body_after, true, true);
> +}
>
> Need cases for:
>
>    - showing cleanup of OpName and OpDecoration, OpGroupDecorate
>
>
Added a test for this one, but I'm causing some failures during assembly
that I'm still trying to figure out.


>
>    - showing cleanup of OpPhis in live blocks that remain
>
> Well, I'm not sure how I can get this with this change.  This cleanup does
not flatten out branches nor multi-branches. Perhaps from non-structured
control flow we could get this.


>
>    - keeping unreachable merge block
>    - keeping unreachable continue target, and the loop backedge it
>    dominates even if it's not in the same block
>
>  Will do.


>    - in these latter two cases: non-trivial defs and uses inside those
>    unreachable blocks
>
> Hmm, ok, but when a block is gone, we remove all its instructions.  This
is covered by  the existing tests?  Or are you thinking something else?


>
>
> In tools/opt/opt.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r137844700>
> :
>
> > @@ -112,9 +112,12 @@ NOTE: The optimizer is a work in progress.
>                 Join two blocks into a single block if the second has the
>                 first as its only predecessor. Performed only on entry point
>                 call tree functions.
> -  -h, --help
> +  --cfg-cleanup
> +               Cleanup the control flow graph. This will remove any unnecessary
> +               code from the CFG like unreachable code.
>
> .... "Peformed only on entry point call tree functions"
>

Done.

dnovillo:
On Thu, Sep 21, 2017 at 4:50 PM, Diego Novillo <dnovillo@google.com> wrote:

>
> Added a test for this one, but I'm causing some failures during assembly
> that I'm still trying to figure out.
>

Done.  I had written the wrong assembly.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+    // All the successors of a live block are also live.
+    block->ForEachSuccessorLabel(mark_reachable);
+
+    // All the Merge and ContinueTarget blocks of a live block are also live.
+    block->ForMergeAndContinueLabel(mark_reachable);
+  }
+
+  // Erase unreachable blocks.
+  for (auto ebi = func->begin(); ebi != func->end();)
```
dneto0: Strongly prefer braces around multi-line bodies.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,117 @@
...
+namespace spvtools {
+namespace opt {
+
+void CFGCleanupPass::RemoveBlock(ir::Function::iterator* bi) {
+  auto &block = **bi;
+  block.ForEachInst([this](ir::Instruction* inst) {
+    KillNamesAndDecorates(inst);
+    def_use_mgr_->KillInst(inst);
+  });
+  *bi = bi->Erase();
```
dneto0: Need to clean up phis  in live blocks that have this block as a predecessor.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+}
+
+Pass::Status CFGCleanupPass::Process(ir::Module* module) {
+  Initialize(module);
+
+  // Process all entry point functions.
+  ProcessFunction pfn = [this](ir::Function* fp) { return CFGCleanup(fp); };
+  bool modified = ProcessEntryPointCallTree(pfn, module);
```
dneto0: I suggest generalizing this pass to use ProcessReachableCallTree that @s-perron recently added.
That is, also cover functions reachable from exported functions.

Would also have to update the pass description.

dnovillo:
On Tue, Sep 26, 2017 at 5:28 PM, David Neto <notifications@github.com>
wrote:
>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r141124881>
> :
>
> > +  };
> +
> +  // Transitively mark all blocks reachable from the entry as reachable.
> +  while (!worklist.empty()) {
> +    ir::BasicBlock* block = worklist.front();
> +    worklist.pop();
> +
> +    // All the successors of a live block are also live.
> +    block->ForEachSuccessorLabel(mark_reachable);
> +
> +    // All the Merge and ContinueTarget blocks of a live block are also live.
> +    block->ForMergeAndContinueLabel(mark_reachable);
> +  }
> +
> +  // Erase unreachable blocks.
> +  for (auto ebi = func->begin(); ebi != func->end();)
>
> Strongly prefer braces around multi-line bodies.
>

Done.


>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r141126105>
> :
>
> > +
> +#include "cfg_cleanup_pass.h"
> +
> +#include "function.h"
> +#include "module.h"
> +
> +namespace spvtools {
> +namespace opt {
> +
> +void CFGCleanupPass::RemoveBlock(ir::Function::iterator* bi) {
> +  auto &block = **bi;
> +  block.ForEachInst([this](ir::Instruction* inst) {
> +    KillNamesAndDecorates(inst);
> +    def_use_mgr_->KillInst(inst);
> +  });
> +  *bi = bi->Erase();
>
> Need to clean up phis in live blocks that have this block as a predecessor.
>

Done.  PTAL.

This reminds me that we'll want to add more CFG cleanup including removal
of degenerate Phis.  These are particularly common after propagation and
dce.

source/opt/cfg_cleanup_pass.cpp
dnovillo: Oops. Forgot this.  I'm adding this now.

dnovillo: OK, now done.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,170 @@
...
+    // removal of PHI operands.
+    if (inst != block.GetLabelInst()) {
+      RemoveFromReachedPhiOperands(block, inst);
+      KillNamesAndDecorates(inst);
+      def_use_mgr_->KillInst(inst);
+    }
+  });
+
+  // Remove the label instruction last.
+  def_use_mgr_->KillInst(block.GetLabelInst());
```
dneto0: What if the label has a name?

This code will produce invalid SPIR-V when run on the following input:
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpSource GLSL 430
               OpName %main "main"
               OpName %dead "dead"
       %void = OpTypeVoid
          %5 = OpTypeFunction %void
       %main = OpFunction %void None %5
          %6 = OpLabel
               OpReturn
       %dead = OpLabel
               OpReturn
               OpFunctionEnd
```

%dead is dead, and will be swept away.  But then the OpName targeting %dead will be left around.

Should use KillNamesAndDecorates on the label instruction.

Please add the above as a test case too.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  for (auto u : *uses) {
+    if (u.inst->opcode() != SpvOpPhi) {
+      continue;
+    }
+
+    ir::Instruction* phi_inst = u.inst;
+    std::vector<ir::Operand> keep_operands;
+    for (uint32_t i = 0; i < phi_inst->NumOperands(); i++) {
```
dneto0: Start i from 2 here, then you won't need the test against i >= 2 later on.
(But also move the "Phi arguments start at index 2" comment to just before the loop)

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,170 @@
...
+      const ir::Operand& var_op = phi_inst->GetOperand(i);
+      if (i >= 2 && i < phi_inst->NumOperands() - 1) {
+        // PHI arguments start at index 2. Each argument consists of two
+        // operands: the variable id and the originating block id.
+        const ir::Operand& block_op = phi_inst->GetOperand(i + 1);
+        assert(var_op.words.size() == 1 && block_op.words.size() == 1 &&
+               "Phi operands should have exactly one word in them.");
+        uint32_t var_id = var_op.words.front();
```
dneto0: It's a little more idiomatic t write:

```
 uint32_t var_id = GetSingleWordOperand(i);
 uint32_t block_id = GetSingleWordOperand(i+1);
```
rather than accessing the guts of an Operand instance.

Separately, both var_id and block_id should be const.

dnovillo:
I think this is ready for review again.  Any other changes?  Thanks.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,170 @@
...
+        uint32_t block_id = block_op.words.front();
+        if (var_id == inst_id && block_id == block.id()) {
+          i++;
+          continue;
+        }
+      }
+
+      keep_operands.push_back(var_op);
```
dneto0: Where's block_op ?  This smells like a bug.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,170 @@
...
+      if (i >= 2 && i < phi_inst->NumOperands() - 1) {
+        // PHI arguments start at index 2. Each argument consists of two
+        // operands: the variable id and the originating block id.
+        const ir::Operand& block_op = phi_inst->GetOperand(i + 1);
+        assert(var_op.words.size() == 1 && block_op.words.size() == 1 &&
+               "Phi operands should have exactly one word in them.");
+        uint32_t var_id = var_op.words.front();
+        uint32_t block_id = block_op.words.front();
+        if (var_id == inst_id && block_id == block.id()) {
+          i++;
```
dneto0: This increment looks wrong.

I think you'll get out of phase. 

I recommend:
- the "for" statement should be:    `for (i = 2 ; i < phi_inst->NumOperands() ; i+= 2)`
- then the body of the loop either adds the var_op and block_op as a pair, or not.

dnovillo:
On Thu, Oct 5, 2017 at 2:55 PM, David Neto <notifications@github.com> wrote:

> What if the label has a name?
>
> This code will produce invalid SPIR-V when run on the following input:
>
>                OpCapability Shader
>           %1 = OpExtInstImport "GLSL.std.450"
>                OpMemoryModel Logical GLSL450
>                OpEntryPoint Vertex %main "main"
>                OpSource GLSL 430
>                OpName %main "main"
>                OpName %dead "dead"
>        %void = OpTypeVoid
>           %5 = OpTypeFunction %void
>        %main = OpFunction %void None %5
>           %6 = OpLabel
>                OpReturn
>        %dead = OpLabel
>                OpReturn
>                OpFunctionEnd
>
> %dead is dead, and will be swept away. But then the OpName targeting %dead
> will be left around.
>
> Should use KillNamesAndDecorates on the label instruction.
>
> Please add the above as a test case too.
>

Done.


>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r142958693>
> :
>
> > +    return;
> +  }
> +
> +  analysis::UseList* uses = def_use_mgr_->GetUses(inst_id);
> +  if (uses == nullptr) {
> +    return;
> +  }
> +
> +  for (auto u : *uses) {
> +    if (u.inst->opcode() != SpvOpPhi) {
> +      continue;
> +    }
> +
> +    ir::Instruction* phi_inst = u.inst;
> +    std::vector<ir::Operand> keep_operands;
> +    for (uint32_t i = 0; i < phi_inst->NumOperands(); i++) {
>
> Start i from 2 here, then you won't need the test against i >= 2 later on.
> (But also move the "Phi arguments start at index 2" comment to just before
> the loop)
>

Can't.  We need to reproduce the full operand vector for the Phi.  Only the
unwanted operands need to be removed. The test for i>=2 is needed so we
only match the actual Phi operands.  This makes me think we may want to
have sub-classes of ir::Instruction to avoid this kind of thing.  It's
slightly odd that each Phi logical operand is actually 2 operands.


>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r142959318>
> :
>
> > +  for (auto u : *uses) {
> +    if (u.inst->opcode() != SpvOpPhi) {
> +      continue;
> +    }
> +
> +    ir::Instruction* phi_inst = u.inst;
> +    std::vector<ir::Operand> keep_operands;
> +    for (uint32_t i = 0; i < phi_inst->NumOperands(); i++) {
> +      const ir::Operand& var_op = phi_inst->GetOperand(i);
> +      if (i >= 2 && i < phi_inst->NumOperands() - 1) {
> +        // PHI arguments start at index 2. Each argument consists of two
> +        // operands: the variable id and the originating block id.
> +        const ir::Operand& block_op = phi_inst->GetOperand(i + 1);
> +        assert(var_op.words.size() == 1 && block_op.words.size() == 1 &&
> +               "Phi operands should have exactly one word in them.");
> +        uint32_t var_id = var_op.words.front();
>
> It's a little more idiomatic t write:
>
>  uint32_t var_id = GetSingleWordOperand(i);
>  uint32_t block_id = GetSingleWordOperand(i+1);
>
> rather than accessing the guts of an Operand instance.
>
> Separately, both var_id and block_id should be const.
>
 Done.

------------------------------
>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r142961769>
> :
>
> > +      const ir::Operand& var_op = phi_inst->GetOperand(i);
> +      if (i >= 2 && i < phi_inst->NumOperands() - 1) {
> +        // PHI arguments start at index 2. Each argument consists of two
> +        // operands: the variable id and the originating block id.
> +        const ir::Operand& block_op = phi_inst->GetOperand(i + 1);
> +        assert(var_op.words.size() == 1 && block_op.words.size() == 1 &&
> +               "Phi operands should have exactly one word in them.");
> +        uint32_t var_id = var_op.words.front();
> +        uint32_t block_id = block_op.words.front();
> +        if (var_id == inst_id && block_id == block.id()) {
> +          i++;
> +          continue;
> +        }
> +      }
> +
> +      keep_operands.push_back(var_op);
>
> Where's block_op ? This smells like a bug.
>

Hmm?  It's the operand right after the variable.  I think we need a better
API to Phi operands.  I'll try to re-work this loop to be less confusing.


>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r142961903>
> :
>
> > +    }
> +
> +    ir::Instruction* phi_inst = u.inst;
> +    std::vector<ir::Operand> keep_operands;
> +    for (uint32_t i = 0; i < phi_inst->NumOperands(); i++) {
> +      const ir::Operand& var_op = phi_inst->GetOperand(i);
> +      if (i >= 2 && i < phi_inst->NumOperands() - 1) {
> +        // PHI arguments start at index 2. Each argument consists of two
> +        // operands: the variable id and the originating block id.
> +        const ir::Operand& block_op = phi_inst->GetOperand(i + 1);
> +        assert(var_op.words.size() == 1 && block_op.words.size() == 1 &&
> +               "Phi operands should have exactly one word in them.");
> +        uint32_t var_id = var_op.words.front();
> +        uint32_t block_id = block_op.words.front();
> +        if (var_id == inst_id && block_id == block.id()) {
> +          i++;
>
> This increment looks wrong.
>
> I think you'll get out of phase.
>
It doesn't, but I agree that the code is confusing.  I'll re-work it to not
be so opaque.

> I recommend:
>
>    - the "for" statement should be: for (i = 2 ; i <
>    phi_inst->NumOperands() ; i+= 2)
>
> Can't do this.  All operands need to be pushed into the new operand
vector.

dneto0:
> Can't.  We need to reproduce the full operand vector for the Phi. 

I see.   I get it now.   I think it would be clearer to move the i++ out of the for statement and only do it in the cases inside the loop.  So you'd have something like:

```
 if ( i < 2 ) {
  keep_operands.push_back( ...);
  ++i;
 } else {
    // handle (var, pred) case, as a pair
    either keep both or neither
    i += 2;
}
```

> >  Where's block_op ? This smells like a bug.

> Hmm?  It's the operand right after the variable.  I think we need a better
API to Phi operands.  I'll try to re-work this loop to be less confusing.

Right.  My problem here is that the code at the end of the function replaces all the (var, predecessor) operands with the sequence of operands contained in keep_operands.   So you have to put both the var_op and the block_op.  I think you were missing the block_op?

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,186 @@
...
+// Remove Phi operand {|rm_arg_id|, |rm_block_id|} from |phi|. This works by
+// building a new operand vector for |phi|.  The new operand vector will have
+// all the original operands except the operand {|rm_arg_id|, |rm_block_id|}.
+void RemovePhiOperand(ir::Instruction* phi, uint32_t rm_arg_id,
+                      uint32_t rm_block_id) {
+  std::vector<ir::Operand> keep_operands;
+
+  // Traverse all the operands in |phi|. Build the new operand vector by adding
+  // all the original operands from |phi| except the unwanted one.
+  for (uint32_t i = 0; i < phi->NumOperands();) {
```
dneto0: Thanks.  This is much easier to understand.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  analysis::UseList* uses = def_use_mgr_->GetUses(inst_id);
+  if (uses == nullptr) {
+    return;
+  }
+
+  for (auto u : *uses) {
+    if (u.inst->opcode() == SpvOpPhi) {
+      RemovePhiOperand(u.inst, inst_id, block.id());
```
dneto0: In SPIR-V the block in a Phi is the immediate predecessor block, not necessarily the block in which the associated value is defined.

Example:

%block0  defines value %e
Then %block0 branches into %block1, then %block1 branches to %block2. 
%block2 has a phi with (%e, %block1).
Then when killing %block0 you're going to delete the definition of %e but miss removing (%e, %block1) from the phi.

So you need an alternative case where you substitute an undef value for inst used in a subsequent phi.
E.g. something like in https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/local_ssa_elim_pass.cpp#L271 where an undef value is generated for the given type.

You'll need a test case for this scenario.

dneto0:
Also, please rebase against master.  :-)

dnovillo:
On Tue, Oct 10, 2017 at 12:37 PM, David Neto <notifications@github.com>
wrote:

> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r143782193>
> :
>
> > +
> +void CFGCleanupPass::RemoveFromReachedPhiOperands(const ir::BasicBlock& block,
> +                                                  ir::Instruction* inst) {
> +  uint32_t inst_id = inst->result_id();
> +  if (inst_id == 0) {
> +    return;
> +  }
> +
> +  analysis::UseList* uses = def_use_mgr_->GetUses(inst_id);
> +  if (uses == nullptr) {
> +    return;
> +  }
> +
> +  for (auto u : *uses) {
> +    if (u.inst->opcode() == SpvOpPhi) {
> +      RemovePhiOperand(u.inst, inst_id, block.id());
>
> In SPIR-V the block in a Phi is the immediate predecessor block, not
> necessarily the block in which the associated value is defined.
>
> Example:
>
> %block0 defines value %e
> Then %block0 branches into %block1, then %block1 branches to %block2.
> %block2 has a phi with (%e, %block1).
> Then when killing %block0 you're going to delete the definition of %e but
> miss removing (%e, %block1) from the phi.
>
> So you need an alternative case where you substitute an undef value for
> inst used in a subsequent phi.
> E.g. something like in https://github.com/KhronosGrou
> p/SPIRV-Tools/blob/master/source/opt/local_ssa_elim_pass.cpp#L271 where
> an undef value is generated for the given type.
>
> You'll need a test case for this scenario.
>

I'm thinking... What if I just match the value?  It's an SSA name, so I
just need to match it.  If the same value is referenced in multiple
arguments, then they all need to be removed.  Pseudocode (apologies)
%b6: jump %b5
%b0: %20 = ...
%b5: switch (%3) {
  %b1: case 0:
     [ ... ] (no new name for %20)
     break;
  %b2: case 10:
     [...] (no new name for %20)
   case 23:
     %b3: %21 = ...
}
%22 = phi(%20 %b0, %20 %b1, %20 %b2, %21 %b3)

This code should remove all the phi arguments for %20, right?  Just
matching %20 should be enough?

dneto0:
> %22 = phi(%20 %b0, %20 %b1, %20 %b2, %21 %b3)

> This code should remove all the phi arguments for %20, right?  Just
matching %20 should be enough?

Yes, but there are two cases: 
-  The phi parent block is the block you are removing.  Remove the pair (%20 %b0)
- The phi parent block is not the block you are removing.  Replace the %20 with an Undef of the right type.  (%u %b1) etc. where %u is defined appropriately.

dneto0:
Stop the presses.  I thought about this further, and it's kind of weird.

Let's look at that second case again:  The phi parent block is not the block you are removing. 
From your example,  %20 %b1 is one such pair.   Now, if value %20 reaches %22 via %b1, then a use of %20 could exist in %b1. That means the definition of %20 dominates %b1.  But you're removing %20 because it's in a dead block %b0.     That implies that %b0 must be dead too.  So you shouldn't have to bother replacing %20 by an undef in the pair %20 %b1 because you're going to eventually remove the pair anyway.

test/opt/cfg_cleanup_test.cpp
```diff
@@ -0,0 +1,272 @@
...
+         %11 = OpLoad %int %y
+               OpBranch %21
+         %16 = OpLabel
+         %20 = OpIAdd %int %11 %int_42
+               OpBranch %17
+         %21 = OpLabel
+         %24 = OpISub %int %11 %int_23
+               OpBranch %17
+         %17 = OpLabel
+         %31 = OpPhi %int %20 %16 %24 %21
```
dneto0: This case proves that the pair (%20 %16) is removed.  But it's removed because the value %20 is defined in a removed block.

You should also have a case where you remove a pair because the block is dead.  E.g. if you had

constant:  %int_0  = OpConstant %int 0

and then you had 
 %31 = OpPhi %int %int_0 %16 %24 %21

This should also result in removing the pair (%int_0 %16)

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,186 @@
...
+      RemoveFromReachedPhiOperands(block, inst);
+      KillNamesAndDecorates(inst);
+      def_use_mgr_->KillInst(inst);
+    }
+  });
+
+  // Remove the label instruction last.
+  auto label = block.GetLabelInst();
+  KillNamesAndDecorates(label);
+  def_use_mgr_->KillInst(label);
```
dneto0: You'll also need to remove this label from any phis that might use it.
E.g. if removing block A, then remove any pairs (X A) from any Phis.   For example X could be a constant value (as in the example I gave earlier.)

dnovillo:
So, I think I've convinced myself that we cannot remove a PHI argument
unless both the block (i.e., the label) is being removed.  In the example I
gave:

%b6: jump %b5
%b0: %20 = ...
%b5: switch (%3) {
  %b1: case 0:
     [ ... ] (no new name for %20)
     break;
  %b2: case 10:
     [...] (no new name for %20)
  %b3: case 23:
     %21 = ...
}
%22 = phi(%20 %b0, %20 %b1, %20 %b2, %21 %b3)

In here, we are removing %b0, so %20 is being removed.  Removing %b0 means
that the PHI node %22 gets its in-degree reduced by 1.  So instead of 4
pairs of (value, block) arguments, it should have 3.

%22 = phi(%20 %b0, %20 %b1,    %20 %b2,    %21 %b3) must be converted to
%22 = phi(         %Undef %b1, %Undef %b2, %21 %b3).

We cannot remove the arguments for %b1 nor %b2, because those blocks are
still there.  We'd get an invalid PHI, if we did.  So, I think the right
way to update PHI nodes is what you had proposed originally:


   1. If the argument's block matches an unreachable node, the argument
   must be removed.
   2. If the argument's value matches a removed value, but the block is
   still live, then the argument's value must be replaced with Undef.


Does this sound reasonable?

dnovillo:
I implemented this approach.  PTAL.

Thanks.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,269 @@
...
+// 2- Any operand whose value is |rm_arg_id|, and doesn't come from
+//    |rm_block_id|, gets its value replaced with an OpUndef value. Since
+//    |rm_arg_id| was generated in the unreachable block |rm_block_id|, it no
+//    longer exists, so it cannot be referenced.  However, since the value
+//    does not reach the Phi instruction directly from |rm_block_id|, the
+//    operand cannot be removed from the Phi instruction.  Therefore, we replace
+//    the argument value with OpUndef.  Note that there may be several
+//    operands in |phi| that meet this requirement.
+//
+// For example, in the switch() below, assume that we want to remove the
```
dneto0: Good example!

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,269 @@
...
+  std::unique_ptr<ir::Instruction> undef_inst(
+      new ir::Instruction(SpvOpUndef, type_id, undefId, {}));
+  def_use_mgr_->AnalyzeInstDefUse(&*undef_inst);
+  module_->AddGlobalValue(std::move(undef_inst));
+  type2undefs_[type_id] = undefId;
+
+  return undefId;
+}
+
+// Remove Phi operand (|rm_arg_id|, |rm_block_id|) from |phi|. There are two
```
dneto0: Excellent comment.

source/opt/cfg_cleanup_pass.h
```diff
@@ -0,0 +1,88 @@
...
+
+  // Initialize the pass.
+  void Initialize(ir::Module* module);
+
+  // Remove the result from |inst| from every Phi instruction reached by
+  // |inst|. The instruction is coming from basic block |block|.
+  void RemoveFromReachedPhiOperands(const ir::BasicBlock& block,
+                                    ir::Instruction* inst);
+
+  // Remove Phi operand (|rm_arg_id|, |rm_block_id|) from |phi|.
```
dneto0: This needs slightly more explanation?
```
Remove Phi operand involving either |rm_arg_id| or |rm_block_id|:
- Remove any pair (var, |rm_block_id|)
- Otherwise, replace (|rm_arg_id|, block) with ( U, block ) where U is the undefined value with same type as |rm_arg_id|
```

dneto0: It would also be handy if you allowed |rm_arg_id| to be zero.  Then you can use it for the cleanup of OpLabel mentioned in phis, e.g. where the value is some OpConstant.

source/opt/cfg_cleanup_pass.cpp
dneto0: Still need to take care of this.

If RemovePhiOperand allowed a 0 arg id then you could use that.

dneto0:
The Travis build failure looks to be unrelated.  I'm investigating that.

dnovillo:
On Fri, Oct 13, 2017 at 4:52 PM, David Neto <notifications@github.com>
wrote:

> In source/opt/cfg_cleanup_pass.h
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r144603538>
> :
>
> > +  // Remove all the unreachable basic blocks in |func|.
> +  bool RemoveUnreachableBlocks(ir::Function* func);
> +
> +  // Remove the block pointed by the iterator |*bi|. This also removes
> +  // all the instructions in the pointed-to block.
> +  void RemoveBlock(ir::Function::iterator *bi);
> +
> +  // Initialize the pass.
> +  void Initialize(ir::Module* module);
> +
> +  // Remove the result from |inst| from every Phi instruction reached by
> +  // |inst|. The instruction is coming from basic block |block|.
> +  void RemoveFromReachedPhiOperands(const ir::BasicBlock& block,
> +                                    ir::Instruction* inst);
> +
> +  // Remove Phi operand (|rm_arg_id|, |rm_block_id|) from |phi|.
>
> This needs slightly more explanation?
>
> Remove Phi operand involving either |rm_arg_id| or |rm_block_id|:
> - Remove any pair (var, |rm_block_id|)
> - Otherwise, replace (|rm_arg_id|, block) with ( U, block ) where U is the undefined value with same type as |rm_arg_id|
>
>
I've re-implemented this code.  It turns out that we need to analyze *ALL*
the Phi instructions in the program.  Before, the code was trying to be
selective and only analyze Phis reached by unreachable blocks, but that's
not always the case. A Phi operand may be reached by an unreachable block
that has no def-use edges into it.  I added a test for it:

          %5 = OpLabel
               OpBranch %14
         %40 = OpLabel
          %9 = OpLoad %int %y
         %12 = OpSGreaterThan %bool %9 %int_10
               OpSelectionMerge %14 None
               OpBranchConditional %12 %13 %14
         %13 = OpLabel
               OpBranch %14
         %14 = OpLabel
         %25 = OpPhi %int %24 %5 %int_23 %13

Block %13 here is unreachable, but the argument %int_23 is not defined
there.  So, by just following def-use edges, we would never analyze that
Phi instruction.

 ------------------------------
>
> In source/opt/cfg_cleanup_pass.h
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r144604386>
> :
>
> > +  // Remove all the unreachable basic blocks in |func|.
> +  bool RemoveUnreachableBlocks(ir::Function* func);
> +
> +  // Remove the block pointed by the iterator |*bi|. This also removes
> +  // all the instructions in the pointed-to block.
> +  void RemoveBlock(ir::Function::iterator *bi);
> +
> +  // Initialize the pass.
> +  void Initialize(ir::Module* module);
> +
> +  // Remove the result from |inst| from every Phi instruction reached by
> +  // |inst|. The instruction is coming from basic block |block|.
> +  void RemoveFromReachedPhiOperands(const ir::BasicBlock& block,
> +                                    ir::Instruction* inst);
> +
> +  // Remove Phi operand (|rm_arg_id|, |rm_block_id|) from |phi|.
>
> It would also be handy if you allowed |rm_arg_id| to be zero. Then you can
> use it for the cleanup of OpLabel mentioned in phis, e.g. where the value
> is some OpConstant.
>

The new code should handle this case.  The new test is an example of this.



>
> In source/opt/cfg_cleanup_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/810#discussion_r144604756>
> :
>
> > +  auto& block = **bi;
> +  block.ForEachInst([&block, this](ir::Instruction* inst) {
> +    // Note that we do not kill the block label instruction here. The label
> +    // instruction is needed to identify the block, which is needed by the
> +    // removal of PHI operands.
> +    if (inst != block.GetLabelInst()) {
> +      RemoveFromReachedPhiOperands(block, inst);
> +      KillNamesAndDecorates(inst);
> +      def_use_mgr_->KillInst(inst);
> +    }
> +  });
> +
> +  // Remove the label instruction last.
> +  auto label = block.GetLabelInst();
> +  KillNamesAndDecorates(label);
> +  def_use_mgr_->KillInst(label);
>
> Still need to take care of this.
>
> If RemovePhiOperand allowed a 0 arg id then you could use that.
>

Sorry.  Take care of what?  I'm not following you here.

PTAL.  Thanks.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,287 @@
...
+void CFGCleanupPass::RemovePhiOperands(
+    ir::Instruction* phi,
+    std::unordered_set<ir::BasicBlock*> reachable_blocks) {
+  std::vector<ir::Operand> keep_operands;
+  uint32_t type_id = 0;
+  uint32_t undef_id = 0;
+
+  // Traverse all the operands in |phi|. Build the new operand vector by adding
+  // all the original operands from |phi| except the unwanted ones.
+  bool undef_created = false;
```
dneto0: You could use the convention that undef_id > 0 if and only if you've created the undef value.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+  *bi = bi->Erase();
+}
+
+bool CFGCleanupPass::RemoveUnreachableBlocks(ir::Function* func) {
+  bool modified = false;
+
+  // Mark reachable all blocks reachable from the function's entry block.
+  std::unordered_set<ir::BasicBlock*> reachable_blocks;
+  std::unordered_set<ir::BasicBlock*> visited_blocks;
```
dneto0: It seems to me that you could remove visited_blocks in favour of using reachable_blocks.  But it's not that important, and this way is readable.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -0,0 +1,287 @@
...
+  for (auto& block : *func) {
+    // If the block is about to be removed, don't bother updating its
+    // Phi instructions.
+    if (reachable_blocks.count(&block) == 0) {
+      continue;
+    }
+
+    // If the block is reachable and has Phi instructions, remove all
+    // operands from its Phi instructions that reference unreachable blocks.
+    if (block.HasPhiInstructions()) {
```
dneto0: It's redundant to check HasPhiInstructions  here.  Just run ForEachPhiInst directly, and be happy if it never calls the lambda.

dneto0:
I forgot.  This should also add source/opt/cfg_cleanup_pass.cpp to the Android.mk.  I'll do that before merging.

dneto0:
hm. I think I merged the wrong thing. :-(

dnovillo:
Anything for me to fix?

dneto0:
Ok. I pushed in the right thing now.  :-)

Rebased and merged into master as  c75704ec and fixup 8ec62deb2

dneto0:
I think what happened is that I had a stale 702 branch lying around from a previous iteration of this review.  I've now put on the actual head of your 702, suitably rebased.  So all good now.  First time this has happened to me.  :-(

dnovillo:
Cool.  Thanks!

PR #811, opened 2017-09-07
Add validate logicals pass to the validator

atgoo:
New pass checks operands of all instructions listed under
3.32.15. Relational and Logical Instructions

### DIFF
--- a/source/validate_logicals.cpp
+++ b/source/validate_logicals.cpp
@@ -229,90 +229,11 @@ spv_result_t LogicalsPass(ValidationState_t& _,
     }
 
     case SpvOpIEqual:
-    case SpvOpINotEqual: {
-      if (!_.IsBoolScalarType(result_type) &&
-          !_.IsBoolVectorType(result_type))
-        return _.diag(SPV_ERROR_INVALID_DATA)
-            << "Expected bool scalar or vector type as Result Type: "
-            << spvOpcodeString(opcode);
-
-      const uint32_t left_type = GetOperandTypeId(_, inst, 2);
-      const uint32_t right_type = GetOperandTypeId(_, inst, 3);
-
-      if (!left_type || (!_.IsIntScalarType(left_type) &&
-                         !_.IsIntVectorType(left_type)))
-        return _.diag(SPV_ERROR_INVALID_DATA)
-            << "Expected operands to be scalar or vector int: "
-            << spvOpcodeString(opcode);
-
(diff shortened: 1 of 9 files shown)


source/validate_logicals.cpp
```diff
@@ -0,0 +1,357 @@
...
+#include "val/instruction.h"
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+namespace {
+
+// Returns operand word for given instruction and operand index.
+// The operand is expected to only have one word.
+inline uint32_t GetOperandWord(const spv_parsed_instruction_t* inst,
```
ehsannas: All calls to `GetOperandWord` are prepended with `_.GetTypeId`. For example:
```cpp
const uint32_t operand_type = _.GetTypeId(GetOperandWord(inst, index));
```

I suggest you change the function to:
```cpp
inline uint32_t GetOperandType(
  ValidationState_t& _ ,
  const spv_parsed_instruction_t * inst,
  size_t operand_index)
```
and the change the function to return the type:
```cpp
return _.GetTypeId(inst->words[operand.offset]);
```
This will simplify the call sites

dneto0: This is a good simplification of the idiom.  But I suggest appending "Id" to the name, to get GetOperandTypeId  because we might want to save GetOperandType so it returns an ir::Instruction pointer.

source/validate_logicals.cpp
```diff
@@ -0,0 +1,357 @@
...
+    case SpvOpSelect: {
+      uint32_t dimension = 1;
+      {
+        const Instruction* type_inst = _.FindDef(inst->type_id);
+        assert(type_inst);
+
+        const SpvOp type_opcode = type_inst->opcode();
+        switch (type_opcode) {
+          case SpvOpTypePointer: {
+            if (!_.HasCapability(SpvCapabilityVariablePointers))
```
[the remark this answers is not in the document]
dneto0: FYI.  Please use validation state's features_ member to check for whether variable pointers is turned on.  
I want to grow that thing into a bag of bools, each of which is turned on by some up-front decision based on capabilities and extensions.  For example, future SPIR-V specs might pull some things into core, and you want that disjunction logic to exist in only one spot.

dneto0:
Waiting for bots to go green before merging.

dneto0:
Rebased, squashed, and pushed into master as cf85ad1429d

PR #815, opened 2017-09-08
Create a new optimization pass, strength reduction

s-perron:
Create a new optimization pass, strength reduction, which will replace integer multiplication by a constant power of 2 with an equivalent bit shift.  More optimizations could be added later.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -185,6 +185,12 @@ Optimizer::PassToken CreateUnifyConstantPass();
 // OpSpecConstantOp.
 Optimizer::PassToken CreateEliminateDeadConstantPass();
 
+// Creates a strenth-reduction pass.
+// A strenth-reduction pass will look for opportunities to replace an
+// instruction with an equivalent and less expensive one.  For example,
+// multiplying by a power of 2 can be replaced by a bit shift.
+Optimizer::PassToken CreateStrengthReductionPass();
+
 // Creates a block merge pass.
 // This pass searches for blocks with a single Branch to a block with no
 // other predecessors and merges the blocks into a single block. Continue
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -185,8 +185,8 @@ Optimizer::PassToken CreateUnifyConstantPass();
 // OpSpecConstantOp.
 Optimizer::PassToken CreateEliminateDeadConstantPass();
 
-// Creates a strenth-reduction pass.
-// A strenth-reduction pass will look for opportunities to replace an
+// Creates a strength-reduction pass.
+// A strength-reduction pass will look for opportunities to replace an
 // instruction with an equivalent and less expensive one.  For example,
 // multiplying by a power of 2 can be replaced by a bit shift.
 Optimizer::PassToken CreateStrengthReductionPass();
--- a/source/opt/strength_reduction_pass.cpp
+++ b/source/opt/strength_reduction_pass.cpp
@@ -27,7 +27,7 @@
 namespace {
 // Count the number of trailing zeros in the binary representation of
 // |constVal|.
-uint32_t CountLeadingZeros(uint32_t constVal) {
+uint32_t CountTrailingZeros(uint32_t constVal) {
   // Faster if we use the hardware count trailing zeros instruction.
   // If not available, we could create a table.
   uint32_t shiftAmount = 0;
@@ -38,7 +38,7 @@ uint32_t CountLeadingZeros(uint32_t constVal) {
   return shiftAmount;
 }
 
-// Check if |val| is a power of 2 or not.
+// Return true if |val| is a power of 2.
 bool IsPowerOf2(uint32_t val) {
   // The idea is that the & will clear out the least
   // significant 1 bit.  If it is a power of 2, then
@@ -58,7 +58,7 @@ Pass::Status StrengthReductionPass::Process(ir::Module* module) {
   def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
   int32_type_id_ = 0;
   uint32_type_id_ = 0;
-  memset(constant_ids_, 0, sizeof(constant_ids_));
+  std::memset(constant_ids_, 0, sizeof(constant_ids_));
   next_id_ = module->IdBound();
   module_ = module;
 
@@ -91,7 +91,7 @@ bool StrengthReductionPass::ReplaceMultiplyByPowerOf2(
 
       if (IsPowerOf2(constVal)) {
         modified = true;
-        uint32_t shiftAmount = CountLeadingZeros(constVal);
+        uint32_t shiftAmount = CountTrailingZeros(constVal);
         uint32_t shiftConstResultId = GetConstantId(shiftAmount);
 
         // Create the new instruction.
@@ -125,23 +125,22 @@ bool StrengthReductionPass::ReplaceMultiplyByPowerOf2(
 }
 
 void StrengthReductionPass::FindIntTypesAndConstants() {
-  for (auto typeIter = module_->types_values_begin();
-       typeIter != module_->types_values_end(); ++typeIter) {
-    switch (typeIter->opcode()) {
+  for (auto iter = module_->types_values_begin();
+       iter != module_->types_values_end(); ++iter) {
+    switch (iter->opcode()) {
       case SpvOp::SpvOpTypeInt:
-        if (typeIter->GetSingleWordOperand(1) == 32) {
-          if (typeIter->GetSingleWordOperand(2) == 1) {
-            int32_type_id_ = typeIter->result_id();
+        if (iter->GetSingleWordOperand(1) == 32) {
+          if (iter->GetSingleWordOperand(2) == 1) {
+            int32_type_id_ = iter->result_id();
           } else {
-            uint32_type_id_ = typeIter->result_id();
+            uint32_type_id_ = iter->result_id();
           }
         }
         break;
       case SpvOp::SpvOpConstant:
-        if (typeIter->GetSingleWordOperand(0) == uint32_type_id_) {
-          uint32_t value = typeIter->GetSingleWordOperand(2);
-          if (value <= 32)
-            constant_ids_[value] = typeIter->GetSingleWordOperand(1);
+        if (iter->type_id() == uint32_type_id_) {
+          uint32_t value = iter->GetSingleWordOperand(2);
+          if (value <= 32) constant_ids_[value] = iter->result_id();
         }
         break;
       default:
--- a/source/opt/strength_reduction_pass.cpp
+++ b/source/opt/strength_reduction_pass.cpp
@@ -1,0 +1,191 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "strength_reduction_pass.h"
+
+#include <algorithm>
+#include <cstdio>
+#include <unordered_map>
+#include <unordered_set>
+
+#include "def_use_manager.h"
+#include "log.h"
+#include "reflect.h"
+
+namespace {
+// Utility function to count the number of trailing zeros in constVal.
+uint32_t GetShiftAmount(uint32_t constVal) {
+  // Faster if we use the hardware count trailing zeros instruction.
+  // If not available, we could create a table.
+  uint32_t shiftAmount = 0;
+  while (constVal != 1) {
+    ++shiftAmount;
+    constVal = (constVal >> 1);
+  }
+  return shiftAmount;
+}
+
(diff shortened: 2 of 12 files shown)


include/spirv-tools/optimizer.hpp
```diff
@@ -185,6 +185,12 @@ Optimizer::PassToken CreateUnifyConstantPass();
 // OpSpecConstantOp.
 Optimizer::PassToken CreateEliminateDeadConstantPass();
 
+// Creates a strenth-reduction pass.
+// A strenth-reduction pass will look for opportunities to replace an
```
dneto0: "strength"

source/opt/strength_reduction_pass.h
```diff
@@ -0,0 +1,64 @@
...
+// See optimizer.hpp for documentation.
+class StrengthReductionPass : public Pass {
+ public:
+  const char* name() const override { return "strength-reduction"; }
+  Status Process(ir::Module*) override;
+
+ private:
+  // Replaces multiple by power of 2 with an equivalent bit shift.
+  // Returns true if something changed.
+  bool ReplaceMultiplyByPowerOf2(ir::BasicBlock::iterator&);
```
dneto0: Style guide: use a pointer here.  Reference parameters should be const.
(The short of it:  It's much easier to see at the call site that the parameter could be modified.)

source/opt/strength_reduction_pass.h
```diff
@@ -0,0 +1,64 @@
...
+  // it, or create it.  We do not want duplicates.
+  void FindIntTypes();
+
+  // Simple driver to go through every instruction.  Returns true if something
+  // changed.
+  bool ScanFunctions();
+
+  // Will create the type for an unsigned 32-bit integer with id |uint32_type_id_|.
+  // This functions assumes one does not already exist.
+  uint32_t CreateUint32Type();
```
dneto0: What is the point of the return value if it's going to be stored in uint32_type_id_ ?

source/opt/strength_reduction_pass.h
```diff
@@ -0,0 +1,64 @@
...
+
+  // Def-Uses for the module we are processing
+  std::unique_ptr<analysis::DefUseManager> def_use_mgr_;
+
+  // Type ids for the types of interest.
+  uint32_t int32_type_id_;
+  uint32_t uint32_type_id_;
+  bool must_create_type_;
```
dneto0: You could probably avoid needing this bool.   A value of uint32_type_id_ = 0 could indicate that you didn't find one yet.

source/opt/strength_reduction_pass.h
```diff
@@ -0,0 +1,64 @@
...
+
+  // Will create the type for an unsigned 32-bit integer with id |uint32_type_id_|.
+  // This functions assumes one does not already exist.
+  uint32_t CreateUint32Type();
+
+  // Def-Uses for the module we are processing
+  std::unique_ptr<analysis::DefUseManager> def_use_mgr_;
+
+  // Type ids for the types of interest.
+  uint32_t int32_type_id_;
```
dneto0: What if the types don't exist?  If so these should be 0. (No valid Id may be 0.)

source/opt/strength_reduction_pass.h
```diff
@@ -0,0 +1,64 @@
...
+  // Replaces multiple by power of 2 with an equivalent bit shift.
+  // Returns true if something changed.
+  bool ReplaceMultiplyByPowerOf2(ir::BasicBlock::iterator&);
+
+  // Scan the types in the module looking for the the integer types that we are
+  // interested in.  The shift operation needs an unsigned int.  We need to find
+  // it, or create it.  We do not want duplicates.
+  void FindIntTypes();
+
+  // Simple driver to go through every instruction.  Returns true if something
```
dneto0: "Simple" is unnecessary here.  Actually the whole first sentence is kind of beside the point.  How about "Replaces some instructions in function bodies with presumably cheaper ones."

source/opt/strength_reduction_pass.h
```diff
@@ -0,0 +1,64 @@
...
+
+  // Def-Uses for the module we are processing
+  std::unique_ptr<analysis::DefUseManager> def_use_mgr_;
+
+  // Type ids for the types of interest.
+  uint32_t int32_type_id_;
+  uint32_t uint32_type_id_;
+  bool must_create_type_;
+
+  uint32_t next_id_;
```
dneto0: Say what this is.

source/opt/strength_reduction_pass.h
dneto0: Ah.  This function doesn't actually modify uint32_type_id_.  So I wouldn't mention that in the method comment.

source/opt/strength_reduction_pass.cpp
```diff
@@ -0,0 +1,191 @@
...
+  // If not available, we could create a table.
+  uint32_t shiftAmount = 0;
+  while (constVal != 1) {
+    ++shiftAmount;
+    constVal = (constVal >> 1);
+  }
+  return shiftAmount;
+}
+
+// Quick check if |val| is a power of 2 or not.
```
dneto0: A method comment (i.e. appears outside) should focus on the functional behaviour: "Returns true if |val| is a power of two."

The *algorithm* description goes inside, where the executable statements reside.

source/opt/strength_reduction_pass.cpp
```diff
@@ -0,0 +1,191 @@
...
+#include <cstdio>
+#include <unordered_map>
+#include <unordered_set>
+
+#include "def_use_manager.h"
+#include "log.h"
+#include "reflect.h"
+
+namespace {
+// Utility function to count the number of trailing zeros in constVal.
```
dneto0: "Utility function" phrase isn't really helping here. Just say what it does.

Why not just name the function CountTrailingZeroes?  That describes what it does rather than what it could be used for.

source/opt/strength_reduction_pass.cpp
```diff
@@ -0,0 +1,191 @@
...
+uint32_t GetShiftAmount(uint32_t constVal) {
+  // Faster if we use the hardware count trailing zeros instruction.
+  // If not available, we could create a table.
+  uint32_t shiftAmount = 0;
+  while (constVal != 1) {
```
dneto0: The method description didn't give a precondition on |constVal|.  So this isn't quite right.
You seem to assume that constVal is a power of 2.

source/opt/strength_reduction_pass.cpp
dneto0: ... after reading the body ... :   You need to say that constVal is assumed to be a power of 2.

s-perron:
Yes, I have written test, but @dneto0 I want to talk to you about the testing strategy.  Right now the test are somewhat incomplete.  I will add more, but I also don't know the best way to try to verify the test.

s-perron:
I added code that will avoid duplicating constants.

test/opt/strength_reduction_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+using namespace spvtools;
+
+using ::testing::HasSubstr;
+using ::testing::MatchesRegex;
+
+using StrengthReductionBasicTest = PassTest<::testing::Test>;
+
+// Test to make sure we replace 5*8.
+TEST_F(StrengthReductionBasicTest, BasicReplaceMulBy8a) {
```
s-perron: I'll update the name of the test to "BasicReplaceMulBy8"

test/opt/strength_reduction_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+      JoinAllInsts(text), /* skip_nop = */ true);
+
+  EXPECT_EQ(opt::Pass::Status::SuccessWithChange, std::get<1>(result));
+  const std::string& output = std::get<0>(result);
+  EXPECT_THAT(output, Not(HasSubstr("OpIMul")));
+  EXPECT_THAT(output, HasSubstr("OpShiftLeftLogical %uint %uint_5 %uint_3"));
+}
+
+// Test to make sure we replace 8*5.
+TEST_F(StrengthReductionBasicTest, BasicReplaceMulBy8b) {
```
s-perron: I'll update the name and comment above it to indicate this is a multiply by 16 now.

source/opt/strength_reduction_pass.h
```diff
@@ -0,0 +1,75 @@
...
+  // Scan the types and constants in the module looking for the the integer types that we are
+  // interested in.  The shift operation needs a small unsigned integer.  We need to find
+  // them or create them.  We do not want duplicates.
+  void FindIntTypesAndConstants();
+
+  // Get the id for the given constant.  If it does not exist, it will be
+  // created.  The parameter must be between 0 and 32 inclusive.
+  uint32_t GetConstantId(uint32_t);
+
+  // Replaces certain instructions in function bodies with presumable cheaper
```
s-perron: "presumable" should be "presumably"

source/opt/strength_reduction_pass.cpp
```diff
@@ -0,0 +1,211 @@
...
+          } else {
+            uint32_type_id_ = typeIter->result_id();
+          }
+        }
+        break;
+      case SpvOp::SpvOpConstant:
+        if (typeIter->GetSingleWordOperand(0) == uint32_type_id_) {
+          uint32_t value = typeIter->GetSingleWordOperand(2);
+          if (value <= 32)
+            constant_ids_[value] = typeIter->GetSingleWordOperand(1);
```
dneto0: Prefer using result_id() over GetSingleWordOperand(1) as it's a little more readable, and echoes setting of int32_type_id_ and uint32_type_id_

test/opt/strength_reduction_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+using ::testing::MatchesRegex;
+
+using StrengthReductionBasicTest = PassTest<::testing::Test>;
+
+// Test to make sure we replace 5*8.
+TEST_F(StrengthReductionBasicTest, BasicReplaceMulBy8a) {
+  const std::vector<const char*> text = {
+      // clang-format off
+               "OpCapability Shader",
+               "OpCapability Float64",
```
dneto0: Don't need this Float64 capability.

test/opt/strength_reduction_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+using namespace spvtools;
+
+using ::testing::HasSubstr;
+using ::testing::MatchesRegex;
+
+using StrengthReductionBasicTest = PassTest<::testing::Test>;
+
+// Test to make sure we replace 5*8.
+TEST_F(StrengthReductionBasicTest, BasicReplaceMulBy8a) {
+  const std::vector<const char*> text = {
```
dneto0: You may find it easier to use a long raw literal string.  It's less fussy and less work to pull in from an external source.
e.g. see many of the tests for the SSA passes, or inlining.

test/opt/strength_reduction_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+  auto result = SinglePassRunAndDisassemble<opt::StrengthReductionPass>(
+      JoinAllInsts(text), /* skip_nop = */ true);
+
+  EXPECT_EQ(opt::Pass::Status::SuccessWithChange, std::get<1>(result));
+  const std::string& output = std::get<0>(result);
+  EXPECT_THAT(output, Not(MatchesRegex(".*OpConstant %uint 3.*OpConstant %uint 3.*")));
+  EXPECT_THAT(output, Not(MatchesRegex(".*OpTypeInt 32 0.*OpTypeInt 32 0.*")));
+}
+
```
dneto0: You also need tests to demonstrate:
- the new instruction is put in the right place
- the result id of the new instruction is put anywhere the old one was used
- two IMul with the same power-of-2 factor will only generate a single new constant
Nice to have:
- an IMul with two power-of-2 factors.  Should only absorb one.

test/opt/strength_reduction_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+      // clang-format on
+  };
+
+  auto result = SinglePassRunAndDisassemble<opt::StrengthReductionPass>(
+      JoinAllInsts(text), /* skip_nop = */ true);
+
+  EXPECT_EQ(opt::Pass::Status::SuccessWithChange, std::get<1>(result));
+  const std::string& output = std::get<0>(result);
+  EXPECT_THAT(output, Not(MatchesRegex(".*OpConstant %uint 3.*OpConstant %uint 3.*")));
+  EXPECT_THAT(output, Not(MatchesRegex(".*OpTypeInt 32 0.*OpTypeInt 32 0.*")));
```
dneto0: (Interesting, I thought we tried gtest regex matches in the past and they didn't work on somep platforms... But maybe googletest has improved since then)

Another way to cover these cases is to do full string compare on the input and output assembly code.  See the inliner tests for examples.  I recommend doing that for new tests.

test/opt/pass_utils.h
```diff
@@ -49,6 +49,16 @@ std::string JoinAllInsts(const std::vector<const char*>& insts);
 // will be ignored.
 std::string JoinNonDebugInsts(const std::vector<const char*>& insts);
 
+// Returns a vector that contains the contents of |a| followed by the contents
+// of |b|.
+template <typename T>
+std::vector<T> Concat(const std::vector<T>& a, const std::vector<T>& b) {
```
dneto0: This is fine.
But many uses of this function is actually really concatenating 3 vectors.
Elsewhere we've made a function that takes a vector of vectors.  That involves more copies, so it isn't great.  Maybe just add a 2nd overload that takes 3 arguments. ?

Definitely optional, though.

dneto0:
One more thing I just noticed:  Your commit should start with a single self-contained line, then a blank line.  I've modified the commit message when I've been squashing the commits.

dneto0:
Rebased, squashed, and pushed into master as e4c7d8e748a2

PR #817, opened 2017-09-13
Add -O, -Os and -Oconfig flags.

dnovillo:
NOTE: This is still WIP.  I wanted to provide an early patch to give an idea of where I intend to go with this.  I would appreciate feedback on the direction this is going, so I can adjust it accordingly.

Major missing bits:

1- Proper expansion for -O.  I added some transformations mostly to have something to debug,
    but I am sure that list is incomplete and/or makes little sense.
2- Similarly for -Os.
3- Unit tests
4- There is some debugging code that still needs to be taken out.

======================================================================

These flags are expanded to a series of spirv-opt flags with the
following semantics:

-O: expands to passes that attempt to improve the performance of the
    generated code.

-Os: expands to passes that attempt to reduce the size of the generated
     code.

-Oconfig=<file> expands to the sequence of passes determined by the
                flags specified in the user-provided file.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -77,16 +77,31 @@ class Optimizer {
   // method.
   Optimizer& RegisterPass(PassToken&& pass);
 
+  // Registers passes that attempt to improve performance of generated code.
+  // This sequence of passes is subject to constant review and will change
+  // from time to time.
+  Optimizer& RegisterPerformancePasses();
+
+  // Registers passes that attempt to improve the size of generated code.
+  // This sequence of passes is subject to constant review and will change
+  // from time to time.
+  Optimizer& RegisterSizePasses();
+
   // Optimizes the given SPIR-V module |original_binary| and writes the
   // optimized binary into |optimized_binary|.
   // Returns true on successful optimization, whether or not the module is
   // modified. Returns false if errors occur when processing |original_binary|
   // using any of the registered passes. In that case, no further passes are
-  // excuted and the contents in |optimized_binary| may be invalid.
+  // executed and the contents in |optimized_binary| may be invalid.
   //
   // It's allowed to alias |original_binary| to the start of |optimized_binary|.
   bool Run(const uint32_t* original_binary, size_t original_binary_size,
            std::vector<uint32_t>* optimized_binary) const;
+
+  // Returns a vector of strings with all the pass names added to this
+  // optimizer's pass manager. These strings are valid until the associated
+  // pass manager is destroyed.
+  std::vector<const char *> GetPassNames() const;
 
  private:
   struct Impl;                  // Opaque struct for holding internal data.
--- a/source/opt/pass.h
+++ b/source/opt/pass.h
@@ -56,6 +56,11 @@ class Pass {
   virtual ~Pass() = default;
 
   // Returns a descriptive name for this pass.
+  //
+  // NOTE: When deriving a new pass class, make sure you make the name
+  // compatible with the corresponding spirv-opt command-line flag. For example,
+  // if you add the flag --my-pass to spirv-opt, make this function return
+  // "my-pass" (no leading hyphens).
   virtual const char* name() const = 0;
 
   // Sets the message consumer to the given |consumer|. |consumer| which will be
--- a/tools/opt/opt.cpp
+++ b/tools/opt/opt.cpp
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <cstring>
 #include <algorithm>
+#include <cassert>
+#include <cstring>
+#include <fstream>
 #include <iostream>
 #include <memory>
 #include <sstream>
@@ -26,6 +28,36 @@
 #include "tools/io.h"
 
 using namespace spvtools;
+
+namespace {
+
+// Status and actions to perform after parsing command-line arguments.
+enum OptActions { OPT_CONTINUE, OPT_STOP };
+
+struct OptStatus {
+  OptActions action;
+  int code;
+};
+
+std::string GetListOfPassesAsString(const spvtools::Optimizer& optimizer) {
+  std::stringstream ss;
+  for (const auto& name : optimizer.GetPassNames()) {
+    ss << "\n\t\t" << name;
+  }
+  return ss.str();
+}
+
+std::string GetOptimizationPasses() {
+  spvtools::Optimizer optimizer(SPV_ENV_UNIVERSAL_1_2);
+  optimizer.RegisterPerformancePasses();
+  return GetListOfPassesAsString(optimizer);
+}
+
+std::string GetSizePasses() {
+  spvtools::Optimizer optimizer(SPV_ENV_UNIVERSAL_1_2);
+  optimizer.RegisterSizePasses();
+  return GetListOfPassesAsString(optimizer);
+}
 
 void PrintUsage(const char* program) {
   printf(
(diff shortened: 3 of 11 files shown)


tools/opt/opt.cpp
```diff
@@ -112,28 +126,119 @@ NOTE: The optimizer is a work in progress.
                Join two blocks into a single block if the second has the
                first as its only predecessor. Performed only on entry point
                call tree functions.
-  -h, --help   
+  -O
```
greg-lunarg: I would not advertise -O and -Os until we are fairly confident they are close to what we want. We want to make sure initial experience with these options is positive. If you follow my recommendations for what to include for them, then I would be fine with advertising here.

tools/opt/opt-passes.h
```diff
@@ -0,0 +1,21 @@
...
+#include <string>
+
+namespace spvtools {
+
+// Performance optimization flags expanded by -O.
+//
+// Using the -O flag in spirv-opt's command line is equivalent to specifying
+// the following flags in this exact sequence.
+std::vector<std::string> OptimizationPasses = {
+    "--inline-entry-points-opaque",
```
greg-lunarg: See my recommendations for -Os. I would include those here as well. They are designed to eliminate all loads and stores to function scope variable which is a performance win besides a size win.

I would not use opaque inlining here. It was not designed with performance in mind (and it doesn't even quite solve the problem it was originally designed for). I am currently writing a "default" inliner which I think might be more acceptable here, but until that is ready I think the exhaustive inliner is going to be our best bet. Without some form of inlining, the other size-reduction passes will be much less effective for shaders with function calls.

tools/opt/size-passes.h
```diff
@@ -0,0 +1,24 @@
+#ifndef SPIRV_TOOLS_SIZE_PASSES_H_
+#define SPIRV_TOOLS_SIZE_PASSES_H_
+
+#include <vector>
+#include <string>
+
+namespace spvtools {
+
+// Size optimization flags expanded by -Os.
```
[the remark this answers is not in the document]
dneto0: I haven't given this data point before about exhaustive inlining:  In my work on clspv, I'm working on some codebases originating from the OpenCL world.  Some of the code I've seen has a large handful of entry points in the same compilation units, and they call down into common functions.   Exhaustive inlining for all entry points would increase size.

tools/opt/opt.cpp
```diff
@@ -23,10 +25,22 @@
 #include "spirv-tools/optimizer.hpp"
 
 #include "message.h"
+#include "opt-passes.h"
+#include "size-passes.h"
 #include "tools/io.h"
 
 using namespace spvtools;
 
+std::string Flags2Str(const std::vector<std::string>& flags) {
```
dneto0: Need a method comment.

Also, I prefer To to 2.

tools/opt/opt.cpp
```diff
@@ -112,28 +126,119 @@ NOTE: The optimizer is a work in progress.
...
+  flags.push_back(prog_name);
 
+  if (0 == strcmp(opt_flag, "-O")) {
+    flags.insert(flags.end(), OptimizationPasses.begin(),
+                 OptimizationPasses.end());
+  } else if (0 == strcmp(opt_flag, "-Os")) {
+    flags.insert(flags.end(), SizePasses.begin(), SizePasses.end());
+  } else if (0 == strncmp(opt_flag, "-Oconfig=", sizeof("-Oconfig=") - 1)) {
+    std::vector<std::string> file_flags;
+    if (!ReadFlagsFromFile(opt_flag, &file_flags)) {
```
dneto0: So I guess -Oconfig=foo inside a flags file does not work?   That's fine, I'm sure.  :-)

tools/opt/size-passes.h
```diff
@@ -0,0 +1,24 @@
...
+//
+// Using the -Os flag in spirv-opt's command line is equivalent to specifying
+// the following flags in this exact sequence.
+std::vector<std::string> SizePasses = {
```
dneto0: Make a function to return this.  

"Variables of class type with static storage duration are forbidden: they cause hard-to-find bugs due to indeterminate order of construction and destruction"

tools/opt/size-passes.h
```diff
@@ -0,0 +1,24 @@
...
+    "--strip-debug",
+    "--inline-entry-points-exhaustive",
+    "--eliminate-dead-code-aggressive",
+    "--eliminate-dead-branches",
+    "--merge-blocks",
+    "--flatten-decorations",
```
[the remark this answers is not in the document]
dneto0: Agree.  Also, I don't know of any tools that produce grouped decorations.  :-)

tools/opt/size-passes.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <vector>
+#include <string>
+
+namespace spvtools {
+
+// Size optimization flags expanded by -Os.
+//
+// Using the -Os flag in spirv-opt's command line is equivalent to specifying
+// the following flags in this exact sequence.
```
dneto0: Recommend a note this list is subject to continual tuning.

tools/opt/size-passes.h
```diff
@@ -0,0 +1,24 @@
...
+std::vector<std::string> SizePasses = {
+    "--strip-debug",
+    "--inline-entry-points-exhaustive",
+    "--eliminate-dead-code-aggressive",
+    "--eliminate-dead-branches",
+    "--merge-blocks",
+    "--flatten-decorations",
+    "--compact-ids"
```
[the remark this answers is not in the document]
dneto0: Definitely agree on this one.  spirv-remap works hard to set Ids so they look similar-ish across modules.

tools/opt/opt.cpp
```diff
@@ -112,28 +126,119 @@ NOTE: The optimizer is a work in progress.
...
+
+               The following two invocations to spirv-opt are equivalent:
+
+               $ spirv-opt -Oconfig=opts.cfg program.spv
+
+               $ spirv-opt --inline-entry-points-exhaustive \
+                    --eliminate-dead-code-aggressive program.spv
+
+               Note that the flags -O, -Os and -Oconfig=<file> are also
+               affected by other flags present on the command line. For
```
dneto0: Can we improve the "affected" phrasing?  How about:

Is this better?

> The -O, -Os, and -Oconfig flags act as macros.  Using one of them is equivalent to explicitly inserting the underlying flags at that position in the command line.

tools/opt/opt.cpp
dnovillo: I think marking it [experimental] is the way to go.  No point hiding it if we want users to try it out.

tools/opt/size-passes.h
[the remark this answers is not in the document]
dnovillo: Thanks, folks.  But I really didn't put any thought in these strings.  They're very literally options I just picked at random.  I would like the dev community to give an initial list of flags you think might be good for -O and -Os.  I very literally do not care what we put here for now. I don't have a good set of benchmarks to try it for now.

tools/opt/size-passes.h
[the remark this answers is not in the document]
dnovillo: Thanks. I've added these passes for now.  We'll be testing them internally and we'll see how they work.  I'm particularly perplexed at the exhaustive inliner, but I don't really have a good idea of why it's having this effect.

tools/opt/opt-passes.h
dnovillo: Thanks.  As a starting point, I've made both lists identical for now.  I've added comments stating that they're in flux.

include/spirv-tools/optimizer.hpp
```diff
@@ -77,6 +77,10 @@ class Optimizer {
   // method.
   Optimizer& RegisterPass(PassToken&& pass);
 
+  // Registers passes that attempt to improve performance of generated code.
+  // The sequence of passes is defined in the PerformancePasses vector.
+  void RegisterPerformancePasses();
```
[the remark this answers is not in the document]
dnovillo: Done. I actually do not need this anymore.

tools/opt/opt.cpp
dnovillo: Done.

tools/opt/opt.cpp
```diff
@@ -23,10 +25,22 @@
...
 
 #include "message.h"
+#include "opt-passes.h"
+#include "size-passes.h"
 #include "tools/io.h"
 
 using namespace spvtools;
 
+std::string Flags2Str(const std::vector<std::string>& flags) {
+  std::string s;
```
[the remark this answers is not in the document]
dnovillo: Done.

tools/opt/opt.cpp
```diff
@@ -112,28 +126,119 @@ NOTE: The optimizer is a work in progress.
...
-  -h, --help   
+  -O
+               Optimize for performance. Apply a sequence of transformations
+               in an attempt to improve the performance of the generated
+               code. This flag is equivalent to specifying the following
+               flags: %s
+  -Os
+               Optimize for size. Apply a sequence of transformations in an
+               attempt to minimize the size of the generated code. This flag
+               is equivalent to specifying the following flags: %s
```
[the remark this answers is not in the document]
dnovillo: Done.

tools/opt/opt.cpp
```diff
@@ -112,28 +126,119 @@ NOTE: The optimizer is a work in progress.
...
 )",
-      program, program);
+      program, program, Flags2Str(OptimizationPasses).c_str(),
+      Flags2Str(SizePasses).c_str());
 }
 
-int main(int argc, char** argv) {
-  const char* in_file = nullptr;
-  const char* out_file = nullptr;
+bool ReadFlagsFromFile(const char* oconfig_flag,
```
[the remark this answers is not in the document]
dnovillo: Done

tools/opt/opt.cpp
```diff
@@ -112,28 +126,119 @@ NOTE: The optimizer is a work in progress.
...
-  optimizer.SetMessageConsumer([](spv_message_level_t level, const char* source,
-                                  const spv_position_t& position,
-                                  const char* message) {
-    std::cerr << StringifyMessage(level, source, position, message)
-              << std::endl;
-  });
+  while (!input_file.eof()) {
+    std::string flag;
+    input_file >> flag;
+    if (flag.length() > 0 && flag[0] != '#') {
```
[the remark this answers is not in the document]
dnovillo: Done.

tools/opt/opt.cpp
```diff
@@ -112,28 +126,119 @@ NOTE: The optimizer is a work in progress.
...
+      file_flags->push_back(flag);
+    }
+  }
+
+  return true;
+}
+
+int ParseFlags(int argc, const char** argv, Optimizer* optimizer,
```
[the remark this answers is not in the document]
dnovillo: Done.

tools/opt/opt.cpp
```diff
@@ -112,28 +126,119 @@ NOTE: The optimizer is a work in progress.
...
+               const char** in_file, const char** out_file);
+
+int ParseOptFlags(const char* prog_name, const char* opt_flag,
```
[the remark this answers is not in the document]
dnovillo: Done.

tools/opt/opt.cpp
dnovillo: Yeah, I was debating whether to allow it or not. I'm disallowing it for now, but we can review this later.  It's possible to have -O and -Os inside the configuration file, of course.  I figured a -Oconfig= inside a configuration file is just asking for trouble.

tools/opt/size-passes.h
dnovillo: Done.

tools/opt/size-passes.h
dnovillo: Done.  I've reworked the whole -O and -Os handling.  They are now part of the Optimizer API. There are two new methods: RegisterPerformancePasses and RegisterSizePasses.  These do the direct instantiation of the related passes.  Having them inside the Optimizer API allows us to invoke them from places other than spirv-opt.

tools/opt/opt.cpp
dnovillo: Done.

dnovillo:
I've addressed all the requested changes.  PTAL. Thanks.

dnovillo:
On Thu, Oct 5, 2017 at 5:07 PM, David Neto <notifications@github.com> wrote:

> *@dneto0* requested changes on this pull request.
>
> Low level item: Method comment on GetPassNames should talk about lifetime
> of the char pointers.
>
> High level item: The name advertised by a particular pass via its name()
> method is often different from what we use on the command line. So that
> will be confusing.
> Two options:
>
>    1. Explain the difference. (Yucky)
>    2. Change the internal name to the public name (without leading
>    double-hyphen). And forever catch a particular discrepancy in code review.
>
> I prefer option 2. I don't think anybody really depends on the specific
> result of the name() method yet.
>
> What do you think?
>

Yeah, that was bugging me too but I forgot to talk about it in my response
earlier.  I prefer #2 too. We could also set the name at pass construction
time.  The driver would pass these strings directly from the flags.  But I
think it's easier to give them a fixed string name and catch it in code
review.

I've changed all the passes to advertise their spirv-opt flag name.



>
> In include/spirv-tools/optimizer.hpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/817#discussion_r142996373>
> :
>
> >    //
>    // It's allowed to alias |original_binary| to the start of |optimized_binary|.
>    bool Run(const uint32_t* original_binary, size_t original_binary_size,
>             std::vector<uint32_t>* optimized_binary) const;
>
> +  // Returns a vector of strings with all the pass names added to this
> +  // optimizer's pass manager.
>
> Since they are const char* you should say how long the pointers are valid.
> They are valid until the pass manager is destroyed.
>
> (Or, change it to return std::string)
>
Done.  No point converting to std::string since they come from the const
char * runtime pool.

PTAL.  Thanks.

source/opt/pass.h
```diff
@@ -56,6 +56,11 @@ class Pass {
   virtual ~Pass() = default;
 
   // Returns a descriptive name for this pass.
+  //
+  // NOTE: When deriving a new pass class, make sure you make the name
+  // compatible with the corresponding spirv-opt command-line flag. For example,
+  // if you add the flag --my-pass to spirv-opt, make this function return
+  // "my-pass" (no leading hyphens).
```

dneto0:
Rebased and pushed into master as c90d7305e7d8

PR #818, opened 2017-09-14
Recognize SPV_AMD_shader_fragment_mask

dneto0:


dneto0:
Self-approving this one.

PR #819, opened 2017-09-14
Dead function elimination

s-perron:
This is the first pass at the code that will remove function that are not needed.  Please let me know if there is anything I am missing.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -102,6 +102,12 @@ Optimizer::PassToken CreateNullPass();
 // Section 3.32.2 of the SPIR-V spec) of the SPIR-V module to be optimized.
 Optimizer::PassToken CreateStripDebugInfoPass();
 
+// Creates an eliminate-dead-functions pass.
+// An eliminate-dead-functions pass will remove all functions that are not in the
+// call trees rooted at entry points and exported functions.  These functions
+// are not needed because they will never be called.
+Optimizer::PassToken CreateEliminateDeadFunctionsPass();
+
 // Creates a set-spec-constant-default-value pass from a mapping from spec-ids
 // to the default values in the form of string.
 // A set-spec-constant-default-value pass sets the default values for the
@@ -204,7 +210,7 @@ Optimizer::PassToken CreateStrengthReductionPass();
 // this time it does not guarantee all such sequences are eliminated.
 //
 // Presence of phi instructions can inhibit this optimization. Handling
-// these is left for future improvements. 
+// these is left for future improvements.
 Optimizer::PassToken CreateBlockMergePass();
 
 // Creates an exhaustive inline pass.
@@ -215,7 +221,7 @@ Optimizer::PassToken CreateBlockMergePass();
 // there is no attempt to optimize for size or runtime performance. Functions
 // that are not in the call tree of an entry point are not changed.
 Optimizer::PassToken CreateInlineExhaustivePass();
-  
+
 // Creates an opaque inline pass.
 // An opaque inline pass inlines all function calls in all functions in all
 // entry point call trees where the called function contains an opaque type
@@ -226,9 +232,9 @@ Optimizer::PassToken CreateInlineExhaustivePass();
 // not legal in Vulkan. Functions that are not in the call tree of an entry
 // point are not changed.
 Optimizer::PassToken CreateInlineOpaquePass();
-  
+
 // Creates a single-block local variable load/store elimination pass.
-// For every entry point function, do single block memory optimization of 
+// For every entry point function, do single block memory optimization of
 // function variables referenced only with non-access-chain loads and stores.
 // For each targeted variable load, if previous store to that variable in the
 // block, replace the load's result id with the value id of the store.
@@ -240,9 +246,9 @@ Optimizer::PassToken CreateInlineOpaquePass();
 // The presence of access chain references and function calls can inhibit
 // the above optimization.
 //
-// Only modules with logical addressing are currently processed. 
+// Only modules with logical addressing are currently processed.
 //
-// This pass is most effective if preceeded by Inlining and 
+// This pass is most effective if preceeded by Inlining and
 // LocalAccessChainConvert. This pass will reduce the work needed to be done
 // by LocalSingleStoreElim and LocalMultiStoreElim.
 //
@@ -266,7 +272,7 @@ Optimizer::PassToken CreateDeadBranchElimPass();
 // Creates an SSA local variable load/store elimination pass.
 // For every entry point function, eliminate all loads and stores of function
 // scope variables only referenced with non-access-chain loads and stores.
-// Eliminate the variables as well. 
+// Eliminate the variables as well.
 //
 // The presence of access chain references and function calls can inhibit
 // the above optimization.
@@ -275,7 +281,7 @@ Optimizer::PassToken CreateDeadBranchElimPass();
 // Currently modules with any extensions enabled are not processed. This
 // is left for future work.
 //
-// This pass is most effective if preceeded by Inlining and 
+// This pass is most effective if preceeded by Inlining and
 // LocalAccessChainConvert. LocalSingleStoreElim and LocalSingleBlockElim
 // will reduce the work that this pass has to do.
 Optimizer::PassToken CreateLocalMultiStoreElimPass();
@@ -320,7 +326,7 @@ Optimizer::PassToken CreateLocalAccessChainConvertPass();
 Optimizer::PassToken CreateAggressiveDCEPass();
 
 // Creates a local single store elimination pass.
-// For each entry point function, this pass eliminates loads and stores for 
+// For each entry point function, this pass eliminates loads and stores for
 // function scope variable that are stored to only once, where possible. Only
 // whole variable loads and stores are eliminated; access-chain references are
 // not optimized. Replace all loads of such variables with the value that is
@@ -362,7 +368,7 @@ Optimizer::PassToken CreateDeadBranchElimPass();
 
 // Creates a pass to consolidate uniform references.
 // For each entry point function in the module, first change all constant index
-// access chain loads into equivalent composite extracts. Then consolidate 
+// access chain loads into equivalent composite extracts. Then consolidate
 // identical uniform loads into one uniform load. Finally, consolidate
 // identical uniform extracts into one uniform extract. This may require
 // moving a load or extract to a point which dominates all uses.
--- a/source/opt/eliminate_dead_functions_pass.cpp
+++ b/source/opt/eliminate_dead_functions_pass.cpp
@@ -21,6 +21,7 @@ namespace opt {
 
 Pass::Status EliminateDeadFunctionsPass::Process(ir::Module* module) {
   bool modified = false;
+  module_ = module;
 
   // Identify live functions first.  Those that are not live
   // are dead.
@@ -32,6 +33,7 @@ Pass::Status EliminateDeadFunctionsPass::Process(ir::Module* module) {
   ProcessReachableCallTree(mark_live, module);
 
   def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+  FindNamedOrDecoratedIds();
   for (auto funcIter = module->begin(); funcIter != module->end();) {
     if (live_function_set.count(&*funcIter) == 0) {
       modified = true;
--- a/source/opt/eliminate_dead_functions_pass.cpp
+++ b/source/opt/eliminate_dead_functions_pass.cpp
@@ -1,0 +1,59 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "eliminate_dead_functions_pass.h"
+
+#include <unordered_set>
+
+namespace spvtools {
+namespace opt {
+
+Pass::Status EliminateDeadFunctionsPass::Process(ir::Module* module) {
+  bool modified = false;
+
+  // Identify live functions first.  Those that are not live
+  // are dead.
+  std::unordered_set<const ir::Function*> live_function_set;
+  ProcessFunction mark_live = [&live_function_set](ir::Function* fp) {
+    live_function_set.insert(fp);
+    return false;
+  };
+  ProcessReachableCallTree(mark_live, module);
+
+  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+  for (auto funcIter = module->begin(); funcIter != module->end();) {
+    if (live_function_set.count(&*funcIter) == 0) {
+      modified = true;
+      EliminateFunction(&*funcIter);
+      funcIter = funcIter.Erase();
+    } else {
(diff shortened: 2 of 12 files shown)


include/spirv-tools/optimizer.hpp
```diff
@@ -102,6 +102,12 @@ Optimizer::PassToken CreateNullPass();
 // Section 3.32.2 of the SPIR-V spec) of the SPIR-V module to be optimized.
 Optimizer::PassToken CreateStripDebugInfoPass();
 
+// Creates a eliminate-dead-functions pass.
+// A eliminate-dead-functions pass will remove all functions that are not entry
+// points, and cannot be reached from and entry point.  These functions are not
```
dneto0: "and" -> "an".
Also have to preserve functions reachable from a function with external linkage.

source/opt/eliminate_dead_functions_pass.h
```diff
@@ -0,0 +1,39 @@
...
+#include "def_use_manager.h"
+#include "function.h"
+#include "mem_pass.h"
+#include "module.h"
+
+namespace spvtools {
+namespace opt {
+
+// See optimizer.hpp for documentation.
+class EliminateDeadFunctionsPass : public MemPass {
```
dneto0: MemPass is increasingly misnamed.  :-)
But that's not for this CL to fix.

source/opt/eliminate_dead_functions_pass.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace spvtools {
+namespace opt {
+
+Pass::Status EliminateDeadFunctionsPass::Process(ir::Module* module) {
+  bool modified = false;
+
+  // Identify live functions first.  Those that are not live
+  // are dead.
+  std::unordered_set<const ir::Function*> liveFunctionSet;
```
dneto0: I prefer underscores to camelcase for variable names.
The style guide agrees with me: https://google.github.io/styleguide/cppguide.html#Variable_Names

source/opt/eliminate_dead_functions_pass.cpp
```diff
@@ -0,0 +1,59 @@
...
+  ProcessFunction pfn = [&liveFunctionSet](ir::Function* fp) {
```
dneto0: "pfn" isn't a good name.   "mark_live" is better.

source/opt/eliminate_dead_functions_pass.cpp
```diff
@@ -0,0 +1,59 @@
...
+    liveFunctionSet.insert(fp);
+    return false;
+  };
+  ProcessReachableCallTree(pfn, module);
```
dneto0: Also need to preserve functions with external linkage, and functions reachable from them.

source/opt/eliminate_dead_functions_pass.cpp
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+
+  return modified ? Pass::Status::SuccessWithChange
+                  : Pass::Status::SuccessWithoutChange;
+}
+
+void EliminateDeadFunctionsPass::EliminateFunction(ir::Function* func) {
+  // Remove all of the instruction in the function body
+  func->ForEachInst(
```
dneto0: It's kind of awesome it boils down to just this.

source/opt/pass.h
```diff
@@ -67,9 +67,14 @@ class Pass {
   // Add to |todo| all ids of functions called in |func|.
   void AddCalls(ir::Function* func, std::queue<uint32_t>* todo);
 
-  // 
+  // Applies |pfn| to every function in the call trees that are rooted at the
```
dneto0: I guess I forgot to ask for the description the first time around.
Must say what the return value means:
"Returns true if any call to the process function returns true.  By convention the process function should return true if it modified the module."

source/opt/pass.h
```diff
@@ -67,9 +67,14 @@ class Pass {
...
+  // entry points.
   bool ProcessEntryPointCallTree(ProcessFunction& pfn, ir::Module* module);
 
+  // Applies |pfn| to every function in the call trees rooted at the entry
+  // points and exported functions.
```
dneto0: Describe the return value.

source/opt/pass.cpp
```diff
@@ -60,6 +60,46 @@ bool Pass::ProcessEntryPointCallTree(
...
+  std::unordered_set<uint32_t> done;
+
+  // Add all entry points since they can be reached from outside the module.
+  for (auto& e : module->entry_points())
+    todo.push(e.GetSingleWordInOperand(kEntryPointFunctionIdInIdx));
+
+  // Add all exported functions since they can be reached from outside the
+  // module.
+  for (auto& a : module->annotations()) {
+    if (a.opcode() == SpvOp::SpvOpDecorate) {
```
dneto0: Sadly, need a TODO here to handle group decorations. (See also #821)

source/opt/pass.cpp
```diff
@@ -60,6 +60,46 @@ bool Pass::ProcessEntryPointCallTree(
...
+      if (a.GetSingleWordOperand(1) == SpvDecoration::SpvDecorationLinkageAttributes) {
+        uint32_t lastOperand = a.NumOperands()-1;
+        if (a.GetSingleWordInOperand(lastOperand) == SpvLinkageType::SpvLinkageTypeExport) {
```
dneto0: In this instruction GetSingleWordOperand is the same as GetSingleWordInOperand.  So I suggest using one or the other but not both.

source/opt/pass.cpp
```diff
@@ -60,6 +60,46 @@ bool Pass::ProcessEntryPointCallTree(
...
+          todo.push(a.GetSingleWordOperand(0));
+        }
+      }
+    }
+  }
+
+  while (!todo.empty()) {
```
dneto0: This bit is common with ProcessEntryPointCallTree.

Suggest refactoring the two functions to extract this common bit and make it usable externally, e.g. ProcessCallTreeFromRoots

source/opt/eliminate_dead_functions_pass.cpp
s-perron: I do not understand that this comment is referring to.  The function `ProcessReachableCallTree` will look at functions with the export linkage decoration.  Is there something else?

test/opt/eliminate_dead_functions_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include <vector>
+
+#include <gmock/gmock.h>
+
+#include "assembly_builder.h"
+#include "pass_fixture.h"
+#include "pass_utils.h"
+
+
+template <typename T> std::vector<T> concat(const std::vector<T> &a, const std::vector<T> &b) {
```
s-perron: This will change when I merge it with my other ongoing pull request: https://github.com/KhronosGroup/SPIRV-Tools/pull/815/.  I will not have this definition here; it will be moved to pass_utils.h.

source/opt/eliminate_dead_functions_pass.cpp
dneto0: I made this comment before reading the method comment for ProcessReachableCallTree.   I'm withdrawing my complaint. :-)

source/opt/pass.cpp
```diff
@@ -25,41 +25,73 @@ namespace {
...
+  for (auto& a : module->annotations()) {
+    // TODO: Handle group decorations as well.  Currently not generate by any
+    // front-end, but could be coming.
+    if (a.opcode() == SpvOp::SpvOpDecorate) {
+      if (a.GetSingleWordOperand(1) ==
+          SpvDecoration::SpvDecorationLinkageAttributes) {
+        uint32_t lastOperand = a.NumOperands() - 1;
+        if (a.GetSingleWordOperand(lastOperand) ==
+            SpvLinkageType::SpvLinkageTypeExport) {
+          roots.push(a.GetSingleWordOperand(0));
```
dneto0: I just remembered that you also need to check that the ID is an ID of a function.   E.g. you can also have variables with external linkage, but you should not include them.
(You'll need a negative test case for this too.)

test/opt/eliminate_dead_functions_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+  // The function Dead should be removed because it is never called.
+  const std::vector<const char*> common_code = {
+      // clang-format off
+               "OpCapability Shader",
+          "%1 = OpExtInstImport \"GLSL.std.450\"",
+               "OpMemoryModel Logical GLSL450",
+               "OpEntryPoint Fragment %main \"main\" %gl_FragColor",
+               "OpExecutionMode %main OriginUpperLeft",
+               "OpSource GLSL 150",
+               "OpSourceExtension \"GL_GOOGLE_cpp_style_line_directive\"",
```
dneto0: Delete the OpSourceExtension.  They're not needed.  :-)

Also don't need OpExtInstImport.

test/opt/eliminate_dead_functions_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+               "OpFunctionEnd",
+  "%Constant_ = OpFunction %float None %9",
+         "%20 = OpLabel",
+               "OpReturnValue %float_0_8",
+               "OpFunctionEnd"
+      // clang-format on
+  };
+
+  auto result = SinglePassRunAndDisassemble<opt::EliminateDeadFunctionsPass>(
+      JoinAllInsts(text), /* skip_nop = */ true);
```
dneto0: Should check that the result is the same as the concat of the input.

test/opt/eliminate_dead_functions_test.cpp
```diff
@@ -0,0 +1,200 @@
...
```
dneto0: Check that the result is the concat of the input.

test/opt/eliminate_dead_functions_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+         "%19 = OpLabel",
+               "OpReturnValue %float_0_5",
+               "OpFunctionEnd",
+      // clang-format on
+  };
+
+  SetAssembleOptions(SPV_TEXT_TO_BINARY_OPTION_PRESERVE_NUMERIC_IDS);
+  SinglePassRunAndCheck<opt::EliminateDeadFunctionsPass>(
+      JoinAllInsts(concat(common_code, dead_function)),
+      JoinAllInsts(common_code), /* skip_nop = */ true);
```
dneto0: Worth noting that this also proves that appropriate OpNames have been removed.

Suggest extending the test case to include a decoration on an instruction inside one of the basic blocks.  For example, change the body to include a floating point add:

```      
     "%Dead = OpFunction %float None %9",
     "%19 = OpLabel",
     "%sum = OpFAdd %float %float_0_5 %float_0_5",
     "OpReturnValue %sum",
     "OpFunctionEnd",
```

And add a decoration in the decorations section:
 ```
OpDecorate %sum RelaxedPrecision
```

Then prove that even the decoration disappears.

File: test/opt/pass_test.cpp [hunk 18]
```diff
@@ -0,0 +1,281 @@
...
+               OpFunctionEnd
+         %12 = OpFunction %float None %9
+         %20 = OpLabel
+         %21 = OpFunctionCall %float %12
+               OpReturnValue %21
+               OpFunctionEnd
+)";
+  // clang-format on
+
+  std::unique_ptr<ir::Module> module =
```
dneto0: All the tests in this file have the same basic pattern.  You could encapsulate the following steps in a function: (a) assemble with preserve IDs, (b) run a dummy pass to collect visited ids, (c) return that list. 

That would be a nice cleanup, but optional.
Also, the tests do a bunch of work with various values.  It's usually best to eliminate all but absolutely necessary instructions.  The functionality in this test file is simple enough that it isn't a big problem, but you should get into the habit of pruning as much as possible.

test/opt/eliminate_dead_functions_test.cpp
```diff
@@ -0,0 +1,189 @@
...
+    "%float_1 = OpConstant %float 1",
+       "%main = OpFunction %void None %7",
+         "%15 = OpLabel",
+         "%16 = OpFunctionCall %float %Constant",
+         "%17 = OpFunctionCall %float %Dead",
+         "%18 = OpCompositeConstruct %v4float %16 %17 %float_0_8 %float_1",
+               "OpStore %gl_FragColor %18",
+               "OpReturn",
+               "OpFunctionEnd",
+      "%Dead = OpFunction %float None %9",
```
dneto0: It's misleading to call the %Dead function "Dead".  :-)

test/opt/eliminate_dead_functions_test.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+using namespace spvtools;
+using ::testing::HasSubstr;
+
+using EliminateDeadFunctionsBasicTest = PassTest<::testing::Test>;
+
+TEST_F(EliminateDeadFunctionsBasicTest, BasicDeleteDeadFunction) {
+  // The function Dead should be removed because it is never called.
+  const std::vector<const char*> common_code = {
+      // clang-format off
```
dneto0: Please eliminate all the instructions that don't illustrate the issue.
You can get by with functions that take no params and return void.  Then you don't need any value types, and almost all blocks consist of only OpLabel,  OpReturn, and OpFunctionCall

test/opt/eliminate_dead_functions_test.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include "pass_fixture.h"
+#include "pass_utils.h"
+
+namespace {
+
+using namespace spvtools;
+using ::testing::HasSubstr;
+
+using EliminateDeadFunctionsBasicTest = PassTest<::testing::Test>;
+
```
dneto0: Please add a test prove that when you remove a function, you also remove decoration instructions that target IDs inside that function.

E.g. try this GLSL shader:
#version 450

precision lowp float;
precision highp int;
void main() {
  lowp float x = 1.0;
  lowp float y = 1.0;
  lowp float z =  x + y;
}

This will generate OpVariable and OpFAdd inside the function, and there are OpDecorate ... RelaxedPrecision targeting those IDs.
(You'll have to artificially create a dead function to wrap around them since glslang only emits reachable functions.)

greg-lunarg:
Did a cursory review of the main code. Looks good.

Good catch on exported functions. I have added an issue to apply size optimizations to exported functions as well.

source/opt/eliminate_dead_functions_pass.h
greg-lunarg: Suggestions?

source/opt/pass.cpp
```diff
@@ -25,41 +25,79 @@ namespace {
...
   // Process call tree
   bool modified = false;
-  std::queue<uint32_t> todo;
   std::unordered_set<uint32_t> done;
-  for (auto& e : module->entry_points())
-    todo.push(e.GetSingleWordInOperand(kEntryPointFunctionIdInIdx));
-  while (!todo.empty()) {
-    const uint32_t fi = todo.front();
+
+  while (!roots->empty()) {
```
atgoo: Petty performance nitpicks:
1. Consider calling pop() before adding new elements. Deque might recycle the front chunk and use it to add new elements at the back instead of allocating a new one.

2. You are doing the same map lookup twice. Consider
```
if (done.insert(fi).second) {
  ... call AddCalls
}
```

source/opt/eliminate_dead_functions_pass.cpp
greg-lunarg: Yes, and the Erase on the function iterator is pretty cool too!

test/opt/eliminate_dead_functions_test.cpp
dneto0: Either this was is now out of date, or I was wrong.  :-)

s-perron: Out of date, I changed it after your comments were made.

dneto0:
Rebased and pushed into master as e43c91046b47b

PR #820, opened 2017-09-14
ExtractInsert: Handle rudimentary Composite ops

greg-lunarg:
This optimizes a single index extract whose composite value terminates with a
CompositeConstruct (or ConstantComposite) by evaluating to the correct
component. This was needed for opaque legalization.

This highlights the need/opportunity to improve this optimization to deal
with more complex composite expressions including currently handled ops
plus Null ops and special vector composition. A TODO has been added.

### DIFF
--- a/source/opt/insert_extract_elim.cpp
+++ b/source/opt/insert_extract_elim.cpp
@@ -55,6 +55,11 @@ bool InsertExtractElimPass::ExtInsConflict(const ir::Instruction* extInst,
   return true;
 }
 
+bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
+  ir::Instruction* typeInst = def_use_mgr_->GetDef(typeId);
+  return typeInst->opcode() == SpvOpTypeVector;
+}
+
 bool InsertExtractElimPass::EliminateInsertExtract(ir::Function* func) {
   bool modified = false;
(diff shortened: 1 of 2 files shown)


source/opt/insert_extract_elim.cpp
```diff
@@ -74,6 +79,33 @@ bool InsertExtractElimPass::EliminateInsertExtract(ir::Function* func) {
...
+          // TODO(greg-lunarg): Handle multiple-indices, ConstantNull, special
+          // vector composition, and additional CompositeInsert.
+          if ((cinst->opcode() == SpvOpCompositeConstruct ||
+               cinst->opcode() == SpvOpConstantComposite) &&
+              (*ii).NumInOperands() == 2) {
+            uint32_t compIdx = (*ii).GetSingleWordInOperand(1);
+            if (IsVectorType(cinst->type_id())) {
+              if (compIdx < cinst->NumInOperands()) {
+                uint32_t i = 0;
+                for (; i <= compIdx; i++) {
```
dneto0: I *think* this verification is unnecessary for valid SPIR-V. 
But ok to keep it.

I'd like to see the scenario that forces this code to exist.

dneto0:
Rebased, squashed, and pushed into master as  7be791aaaa

I'd like to see a test as well, but I believe it will be forthcoming.

PR #822, opened 2017-09-18
Add remaining opcodes to arithmetics validation

atgoo:
Add validation rules for:
- OpIAddCarry
- OpISubBorrow
- OpUMulExtended
- OpSMulExtended

Includes some refactoring of old code.

### DIFF
--- a/source/val/validation_state.h
+++ b/source/val/validation_state.h
@@ -358,6 +358,12 @@ class ValidationState_t {
       uint32_t id, uint32_t* num_rows, uint32_t* num_cols,
       uint32_t* column_type, uint32_t* component_type) const;
 
+  // Collects struct member types into |member_types|.
+  // Returns false iff not struct type or has no members.
+  // Deletes prior contents of |member_types|.
+  bool GetStructMemberTypes(
+      uint32_t struct_type_id, std::vector<uint32_t>* member_types) const;
+
   // Returns true iff |id| is a type corresponding to the name of the function.
   // Only works for types not for objects.
   bool IsFloatScalarType(uint32_t id) const;
--- a/source/validate_arithmetics.cpp
+++ b/source/validate_arithmetics.cpp
@@ -412,7 +412,50 @@ spv_result_t ArithmeticsPass(ValidationState_t& _,
       break;
     }
 
-    // TODO(atgoo@github.com): Support other operations.
+    case SpvOpIAddCarry:
+    case SpvOpISubBorrow:
+    case SpvOpUMulExtended:
+    case SpvOpSMulExtended: {
+      std::vector<uint32_t> result_types;
+      if (!_.GetStructMemberTypes(result_type, &result_types))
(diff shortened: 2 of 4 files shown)


source/val/validation_state.h
```diff
@@ -358,6 +358,11 @@ class ValidationState_t {
       uint32_t id, uint32_t* num_rows, uint32_t* num_cols,
       uint32_t* column_type, uint32_t* component_type) const;
 
+  // Collects struct member types into |member_types|.
+  // Returns false if parsing fails (wrong syntaxis or not struct type).
```
dneto0: "syntaxis"?

dneto0: Also returns false if the struct has no members.

Also, it's probably ok that the vector is not updated in the failure case.  Probably better to say that in the method comment as well.

source/validate_arithmetics.cpp
```diff
@@ -405,7 +412,51 @@ spv_result_t ArithmeticsPass(ValidationState_t& _,
...
+            << spvOpcodeString(opcode);
+
+      const uint32_t left_type_id = GetOperandTypeId(_, inst, 2);
+      const uint32_t right_type_id = GetOperandTypeId(_, inst, 3);
+
+      if (left_type_id != result_types[0] || right_type_id != result_types[0])
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Expected both operands to be of Result Type member type: "
+            << spvOpcodeString(opcode);
+
```
dneto0: Please add TODO for IAddCarry ISubBorrow and UMulExtended: The result types members must have signedness of 0.

dneto0:
Also please indicate if you have run the Vulkan CTS shaders through this extra functionality.

source/validate_arithmetics.cpp
atgoo: That is handled above
`  if (!_.IsUnsignedIntScalarType(result_types[0]) &&
       !_.IsUnsignedIntVectorType(result_types[0]))
        return _.diag(SPV_ERROR_INVALID_DATA)
            << "Expected Result Type struct member types to be unsigned "
            << "integer scalar or vector: " << spvOpcodeString(opcode);`

dneto0: Ah.  I missed that reading the code too quickly.  I didn't quite see the if test against the opcode.

dneto0:
I'll wait for the bots to go green.

dneto0:
Whoops.  Still need the typo fix and updated method comment.

dneto0:
Please fix the typo and method comment.

atgoo:
Ready for review.

dneto0:
Code looks ok.

But:

> Also please indicate if you have run the Vulkan CTS shaders through this extra functionality.

dneto0:
I've verified this works with latest Vulkan CTS tests.

dneto0:
Rebased and pushed into master as dcf42433a63c

Issue #824, opened 2017-09-19
spirv-opt: assert/crash with --eliminate-local-multi-store pass


danginsburg:
I am using SPIRV-Tools master (to commit c6dfc1188) and SPIRV-Headers master (to commit 2bb92e6fe) and hit an assert/crash with the attached shader in --eliminate-local-multi-store pass.

To repro the assert, you can run the command-line:

spirv-opt.exe --eliminate-local-multi-store lightandcopy.spv -o lightandcopy_opt.spv
Assertion failed: index < operands_.size() && "operand index out of bound", file d:\dev\source2\main\src\thirdparty\spirv-tools\source\opt\instruction.h, line 197

The stack of the assert looks like this:
```
 	vfx_gl.dll!common_assert_to_message_box<wchar_t>(const wchar_t * const expression, const wchar_t * const file_name, const unsigned int line_number, void * const return_address) Line 351	C++
 	vfx_gl.dll!common_assert<wchar_t>(const wchar_t * const expression, const wchar_t * const file_name, const unsigned int line_number, void * const return_address) Line 387	C++
 	vfx_gl.dll!_wassert(const wchar_t * expression, const wchar_t * file_name, unsigned int line_number) Line 405	C++
 	vfx_gl.dll!spvtools::ir::Instruction::GetOperand(unsigned int index) Line 197	C++
 	vfx_gl.dll!spvtools::ir::Instruction::GetSingleWordOperand(unsigned int index) Line 72	C++
 	vfx_gl.dll!spvtools::ir::Instruction::GetSingleWordInOperand(unsigned int index) Line 153	C++
 	vfx_gl.dll!spvtools::opt::LocalMultiStoreElimPass::GetPointeeTypeId(const spvtools::ir::Instruction * ptrInst) Line 176	C++
 	vfx_gl.dll!spvtools::opt::LocalMultiStoreElimPass::PatchPhis(unsigned int header_id, unsigned int back_id) Line 390	C++
>	vfx_gl.dll!spvtools::opt::LocalMultiStoreElimPass::EliminateMultiStoreLocal(spvtools::ir::Function * func) Line 467	C++
 	vfx_gl.dll!spvtools::opt::LocalMultiStoreElimPass::ProcessImpl::__l2::<lambda>(spvtools::ir::Function * fp) Line 550	C++
 	[External Code]	
[...]

PR #825, opened 2017-09-19
ElimLocalMultiStore: Reset structured successors for each function

greg-lunarg:
Fixes crash reported in issue #824


greg-lunarg:
@dneto0 I looked at the other definitions of ComputeStructuredSuccessors (I really need to consolidate these).

The reason the "multi-store" version failed is because it uses a pseudo-entry-block. The p-e-b was collecting successors across functions. Technically I could have just cleared the p-e-b's successors, but it was just easier to clear them all.

None of the other versions of CSS use a p-e-b. The reason is because "multi-store" needed all blocks in a function to be processed, even if they weren't reached. The other passes don't care if unreached blocks don't get processed.

Issue #826, opened 2017-09-19
crash in --eliminate-local-multi-store


dneto0:
Example from Vulkan CTS, reduces to this:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %1 "main"
          %2 = OpTypeBool
          %3 = OpConstantTrue %2
          %4 = OpTypeVoid
          %5 = OpTypeFunction %4
          %6 = OpTypeFloat 32
          %7 = OpTypePointer Function %6
          %8 = OpConstant %6 0
          %9 = OpConstant %6 1
          %1 = OpFunction %4 None %5
         %10 = OpLabel
         %11 = OpVariable %7 Function %8
         %12 = OpVariable %7 Function
         %13 = OpLoad %6 %11
               OpBranch %14
         %14 = OpLabel
         %15 = OpPhi %6 %13 %10 %16 %14
         %16 = OpFAdd %6 %15 %9
               OpLoopMerge %17 %14 None
               OpBranchConditional %3 %14 %17
         %17 = OpLabel
               OpStore %12 %15
               OpReturn
               OpFunctionEnd
```

The problem appears in processing %16, in the PatchPhis method.

dneto0:
Cc @greg-lunarg

dneto0:
The live var set is empty when processing the self-loop block with id 14.  And the pass crashes when processing the (16, 14) pair at the end of the OpPhi with id 15.

dneto0:
I have a fix.  Making a test for it.

greg-lunarg:
Yeah, the code assumes all phi's need to be patched, in actuality, only phi's generated by the elim-multi-store pass need to patched. Is that your fix?

dneto0:
> Yeah, the code assumes all phi's need to be patched, in actuality, only phi's generated by the elim-multi-store pass need to patched. Is that your fix?

Yes.  Please take a look at #828

PR #828, opened 2017-09-20
elim-multi-store: only patch loop header phis that we created

dneto0:
There can already be OpPhi instructions in a loop header that
are unrelated to the optimization.  We should not be patching those.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/826


dneto0:
@greg-lunarg Please take a look.

greg-lunarg:
Looks good.

dneto0:
Thanks.  

I'll wait for bots to go green.  There was a hiccup on the OSX bot not being able to check out the code, so I restarted the job.

PR #829, opened 2017-09-21
Validator: fix logicals pass for OpSelect pointers

atgoo:
OpSelect works with pointers also when capability
VariablePointersStorageBuffer is declared (before worked only with
capability VariablePointers).


antiagainst:
@ehsannas: We typically do rebase merge (if there is only one commit) or squash merge (if there are multiple commits). Creating merge commits should be avoided to have a clean history. :)

ehsannas:
@antiagainst oops, my bad. Is there anything I need to do for this one now?

antiagainst:
@ehsannas: We cannot rewrite `master` history. No worries for this one. Just keep in mind for the future.

Issue #502, opened 2016-12-09
Use O(log n) table lookup algorithms for opcodes


dneto0:
Currently the opcode table is just an array, and we do opcode and opcode-string lookups with straight linear search.

See spvOpcodeTableNameLookup and spvOpcodeTableValueLookup in source/opcode.cpp

The underlying tables are generated via the "spvtools_core_tables" invocations in source/CMakeLists.txt.  There's a python script that parses the JSON and then generates .inc files in the build output directory.  Those are then #included in source/opcode.cpp into static tables.

We can do much better by generating the underlying tables in sorted order.  Then use std::lower_bound to find the first matching entry. 

But then we need two tables: one sorted by opcode, and the other sorted by opcode string name.  I'd like the sorted-by-opcode table to be the primary table, and the sorted-by-string-name table to just have an index into the other one.  The reason is that the fast path should be processing a binary module as in a GPU driver, as opposed to doing name lookups for disassembly or assembly as occurs during development.

A wrinkle:  We have to allow for the same opcode to be used with two different string names.  This is almost certainly going to happen over time as Khronos extensions to SPIR-V get incorporated into the core of a future version of SPIR-V.  If that extension describes a new instruction, e.g. OpFooKHR then it would drop the KHR suffix to become OpFoo when it goes into core.

So the work is to:
* Update  utils/generate_grammar_tables.py to generate two opcode tables, sorted by opcode and sorted by opcode string, where it currently generates one.  The sort should be stable, i.e. preserve the order of equal elements in the input.
* Use std::lower_bound for the lookups. (I don't want to see an open-coded binary search.  Those are notoriously tricky to write correctly, and more so in the presence of equal elements.)

ratchetfreak:
Extensions should use the OpExtInst to introduce new opcodes. Once they get into core then they will get promoted to proper opcodes.

dneto0:
@ratchetfreak That's a design choice.

There are already KHR extensions that introduce instructions as new (direct) opcodes.
Eg. see OpSubgroupBallotKHR in https://www.khronos.org/registry/spir-v/extensions/KHR/SPV_KHR_shader_ballot.html

The reasoning is that the working group feels there is a strong chance that the functionality will be pulled into core at a future date with unchanged semantics.  And if that's the case, then tooling that just deals with the numbers of opcodes won't have to change.

dneto0:
cc: @antiagainst

jcaraban:
[...]

PR #830, opened 2017-09-21
Avoid using global static variables

antiagainst:
Previously we have several grammar tables defined as global static
variables and these grammar table entries contains non-POD struct
fields (CapabilitySet/ExtensionSet). The initialization of these
non-POD struct fields may require calling operator new. If used
as a library and the caller defines its own operator new, things
can screw up.

This pull request changes all global static variables into
function static variables, which is lazy evaluated in a thread
safe way as guaranteed by C++11.

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -27,25 +27,25 @@
 #include "spirv_endian.h"
 
 namespace {
+struct OpcodeDescPtrLen {
+  const spv_opcode_desc_t* ptr;
+  uint32_t len;
+};
 
-// Descriptions of each opcode.  Each entry describes the format of the
-// instruction that follows a particular opcode.
-const spv_opcode_desc_t opcodeTableEntries_1_0[] = {
-#include "core.insts-1.0.inc"
-};
-const spv_opcode_desc_t opcodeTableEntries_1_1[] = {
-#include "core.insts-1.1.inc"
-};
-const spv_opcode_desc_t opcodeTableEntries_1_2[] = {
+OpcodeDescPtrLen getOpcodeTableEntries_1_2() {
+  static const spv_opcode_desc_t opcodeTableEntries_1_2[] = {
 #include "core.insts-1.2.inc"
-};
+  };
+
+  return {opcodeTableEntries_1_2, ARRAY_SIZE(opcodeTableEntries_1_2)};
+}
 
 // Represents a vendor tool entry in the SPIR-V XML Regsitry.
 struct VendorTool {
   uint32_t value;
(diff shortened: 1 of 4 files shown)


antiagainst:
I found this issue when trying to call optimizer in spiregg.

Apparently DXC provides its own `operator new` in the dxcompiler DLL.  Optimizer is called in that DLL so it means we need to initialize the global variables in SPIRV-Tools before actually doing other stuff for the DLL. Then the custom `operator new` is called for the static variable init. However, data needed by `operator new` is not initialized yet. Boom.

I cannot thank @AWoloszyn enough for helping me pinpoint this issue amazingly fast. He has a much joyful way to fix this; but I feel the above lazy evaluation is less controversial.

antiagainst:
Non-POD global variables are problematic. I appreciate C++ style guide more now.

source/opcode.cpp
```diff
@@ -173,10 +183,12 @@ void spvInstructionCopy(const uint32_t* words, const SpvOp opcode,
 const char* spvOpcodeString(const SpvOp opcode) {
   // Use the latest SPIR-V version, which should be backward-compatible with all
   // previous ones.
-  for (uint32_t i = 0; i < ARRAY_SIZE(opcodeTableEntries_1_2); ++i) {
-    if (opcodeTableEntries_1_2[i].opcode == opcode)
-      return opcodeTableEntries_1_2[i].name;
+  const auto entries = getOpcodeTableEntries_1_2();
+
+  for (uint32_t i = 0; i < entries.len; ++i) {
```
atgoo: Perhaps create a static map on the first call.

Or if the array is in order then std::binary_search (with a lambda returning opcode) can be used to locate the right index in log(n) time.

antiagainst: Reasonable. This function is used fairly frequently; using map should save some time. Will do in another pull request. Thanks for the suggestion!

dneto0: That's https://github.com/KhronosGroup/SPIRV-Tools/issues/502

dneto0: Also, I get complaints about memory leaks.  So the create-on-first call means we keep it for subsequent calls.  And that's viewed as a memory leak.  I'm ok with at least not making things worse for now.

The plan in #502 suggests building the sorted tables at SPIRV-Tools build time, so that we still only need POD data.

antiagainst: So @dneto0  has more plans for this change. I'll defer doing that.

dneto0:
Rebased and pushed into master as 16981f87f

PR #831, opened 2017-09-22
Add bitwise operations validator pass

atgoo:
The pass checks correctness of operand types of all bitwise instructions
(opcode range from SpvOpShiftRightLogical to SpvOpBitCount).

### DIFF
--- a/source/validate_bitwise.cpp
+++ b/source/validate_bitwise.cpp
@@ -1,0 +1,248 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Validates correctness of bitwise instructions.
+
+#include "validate.h"
+
+#include "diagnostic.h"
+#include "opcode.h"
+#include "val/instruction.h"
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+namespace {
+
+// Returns operand word for given instruction and operand index.
+// The operand is expected to only have one word.
+inline uint32_t GetOperandWord(const spv_parsed_instruction_t* inst,
+                               size_t operand_index) {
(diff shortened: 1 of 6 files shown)


atgoo:
vk-gl-cts master test successful.

source/validate_bitwise.cpp
```diff
@@ -0,0 +1,248 @@
...
+#include "val/instruction.h"
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+namespace {
+
+// Returns operand word for given instruction and operand index.
+// The operand is expected to only have one word.
+inline uint32_t GetOperandWord(const spv_parsed_instruction_t* inst,
```
dneto0: I've seen this a few times.

A future CL could refactor this into a common spot in source/instruction.cpp.

However, the better change is a refactor of the internal representation of SPIR-V instructions, commonizing functionality between the optimizer and the validator.

So I'm happy to wait.  This is fine.

source/validate_bitwise.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    case SpvOpBitCount: {
+      if (!_.IsIntScalarType(result_type) &&
+          !_.IsIntVectorType(result_type))
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Expected int scalar or vector type as Result Type: "
+            << spvOpcodeString(opcode);
+
+      const uint32_t base_type = GetOperandTypeId(_, inst, 2);
+        if (!base_type ||
```
dneto0: Indentation is weird.

source/validate_bitwise.cpp
```diff
@@ -0,0 +1,248 @@
...
+              << "Expected Base Type to be int scalar or vector: "
+              << spvOpcodeString(opcode);
+
+      const uint32_t base_dimension = _.GetDimension(base_type);
+      const uint32_t result_dimension = _.GetDimension(result_type);
+
+      if (base_dimension != result_dimension)
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Expected Base dimension to be equal to Result Type dimension: "
+            << spvOpcodeString(opcode);
```
dneto0: Also for Result Type:  "The components must be wide enough to hold the unsigned Width of Base as an unsigned value. That is, no sign bit is needed or counted when checking for a wide enough result width."

So you should be doing a width check here too.

dneto0:
Rebased and pushd into master as 64d5e5214f3

source/validate_bitwise.cpp
atgoo: But this is not Instruction, this is spv_parsed_instruction_t. There seems to be no good place for it.

PR #836, opened 2017-09-25
Detach MARK-V from the validator

atgoo:
MARK-V codec was previously dependent on the validation state.
Now it doesn't need the validator to function, but can still optionally
create it and validate every instruction once it's decoded.


dneto0:
Rebased and pushed into master as 976e4218d59

PR #841, opened 2017-09-26
Remove duplicated declaration of CreateAggressiveDCEPass().

antiagainst:
We have the same declaration already at the end of the file


dneto0:
Rebased and pushed into master as  99cd25c4

Issue #827, opened 2017-09-20
spirv-opt segfault when combining options --compact-ids --inline-entry-points-exhaustive


hevrard:
spirv-opt segfaults when invoked with both options --compact-ids --inline-entry-points-exhaustive , e.g.:

`spirv-opt --compact-ids --inline-entry-points-exhaustive shader.spv -o shader_opt.spv`

This seems to be independent of the shader contents, the shader must still be a well-formed spir-v file.

dneto0:
I've tried 2 different shaders, and both orderings of optimizations, but haven't reproduced it.

Please provide a reproduction case.

hevrard:
Here comes a shader for reproduction, uploaded as a text file to work around github limitations.

[shader.spv.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/1318292/shader.spv.txt)

dneto0:
I can reproduce the problem and have seen some bad data internally.  Thanks!

dneto0:
The bug is in the compact-ids pass.
On this line it directly modifies the underlying word storing the result id.
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/compact_ids_pass.cpp#L35

But the Instruction instance caches the  result id and type id in members result_id_ and type_id_, respectively.

This just proves again that caching is one of the hardest things in computing.  Ugh.

dneto0:
Assigned to @atgoo 

The interesting thing is the design of the solution is not obvious.

Do we maximize safety by hiding the `words` member of an ir::Operand and force the modification to go through an API on the instruction?  I suspect there are many good uses of direct access.   

Or do we fix the problem locally in the compact-ids pass and take the risk that the same mistake might be made elsewhere.  I'd like @atgoo to propose a design direction before posting a PR.  The right answer might only be clear once you try to hide the operand data members and see if you lose much.

dneto0:
I made ir::Operand::words private and got 31 unique compilation failures.
* Many uses in instruction.h and instruction.cpp are writes, and it would be ok to add a special method for that, because those methods know that they have to update the cached result_id_ and type_id_.
* Most other uses appear to be reads of the data. That would be satisifed by a const & accessor to the words vector.

I suspect the right solution is to make ir::Operand members private and add appropriate getter and setter.

dneto0:
I'm taking this.

dneto0:
[...]

PR #842, opened 2017-09-27
Compact-ids pass should update instruction's result_id member

dneto0:
Also update the result type field.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/827

### DIFF
--- a/test/opt/compact_ids_test.cpp
+++ b/test/opt/compact_ids_test.cpp
@@ -14,6 +14,9 @@
 
 #include <gmock/gmock.h>
 
+#include "spirv-tools/libspirv.hpp"
+#include "spirv-tools/optimizer.hpp"
+
 #include "pass_fixture.h"
 #include "pass_utils.h"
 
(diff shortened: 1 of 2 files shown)


test/opt/compact_ids_test.cpp
```diff
@@ -87,4 +90,57 @@ OpFunctionEnd
...
+  auto assembled = tools.Assemble(
+      input, &binary, SPV_TEXT_TO_BINARY_OPTION_PRESERVE_NUMERIC_IDS);
+  EXPECT_TRUE(assembled);
+
+  spvtools::Optimizer optimizer(env);
+  optimizer.RegisterPass(CreateCompactIdsPass());
+  // The exhaustive inliner will use the result_id
+  optimizer.RegisterPass(CreateInlineExhaustivePass());
+
+  // This should not crash!
```
atgoo: A badge of code quality...

dneto0: Yeah.  I debated on the "proper" way to test this.  First thought was to expose the Instruction objects directly, but that would break the current information-hiding.  That's generally ok, but we've got plans to overhaul the instruction classes and didn't want to cause churn at this moment.

dneto0:
Waiting for the bots to go green.

Issue #846, opened 2017-09-28
DefUseManager does not scale well on large modules


dneto0:
Analyzing a module with thousands of instructions is very slow.
E.g. see AppVeyor failures in #693 where it's taking more than 15 minutes to analyze a module with 65K instructions.

Manually sampling shows the slowness is in DefUseManager::AnalyzeDefUse, in AnalyzeInstDef -> CLearInst -> EraseUseRecordsOfOperandIds and thereabouts.

dneto0:
FYI @s-perron for your consideration.   :-)

dneto0:
Tiny progress with #848, getting 3x speed up on a module with ~65K instructions, on Windows debug build.

s-perron:
The long running test is:

17 - spirv-tools-test_link_global_values_amount (Timeout)

dneto0:
@s-perron is planning a refactor of the ir::Instruction class that will move responsibility of use-to-def mapping out of DefUseManager.  What remains is the def-to-use mapping, and that's the slow bit: it's based on std::list and scales very badly when there are many uses.  (E.g. think of 10K instructions resulting in uint result; then the uint type has 10K uses, and each lookup is linear time.)

I surveyed the existing usage pattern of DefUseManager::GetUses.   All the cases boil down to:
- detect whether an Id has any uses
- if so, do something for each use.
- that "something"  really does access the underlying instruction, e.g. inspecting the opcode.
- clients of DefUseManager ignore the operand_index field.  This is surprising.
   - the operand_index field is *only* used within DefUseManager::ReplaceAllUsesWith.  That itself could be eliminated by scanning over all operands that are of id type, an replacing those that match the 'before' value.

Therefore:
 - Internally for each Use we can just store the using instruction for each use rather than instruction and operand_index. (!)
 - We could replace the GetUses interface with 
      - a ForEachUser that iterates over all instruction that uses the given id (this is the common case)
      - a ForEachUse that iterates over all (instruction, operand_index) pairs.  (Which could be derived via scanning)

- The underlying storage could be an (ordered) std::set<Instruction*, Instruction*> mapping def instruction to use instruction, and where the less comparison is less on the instruction ids.  
   - The iteration over uses for a def would start use the lower_bound method to find first candidate iter, and then continue until you run off the end of the set or reach the next def instruction.
   - Maintenance operations have logarithmic operation complexity. (Though of course the cache behaviour could be not so great.)
[...]

PR #693, opened 2017-07-02
WIP: Implement a SPIR-V linker

pierremoreau:
Still a work-in-progress and I can most likely reuse more code, but at least I adapted the code to use some of the SPIRV-Tools interface/code.

Missing:

- [x] Update SPIR-V version support from 1.0 to 1.2
- [X] Remove duplicate types
- [x] Remove duplicate decorations
- [ ] Remove duplicate names
- [X] Remove duplicate categories
- [x] Remove duplicate `SpvOpExtInstImport`
- [x] Tests
- [x] Ensure that an import/export pair of variables or functions have the same interface before accepting to link

Questions:

- Should the linker require the given binaries to be valid SPIR-V binaries?
- Should the linker require the given binaries to have the same endianness as the host?

PR #848, opened 2017-09-28
Cache end iterators for speed

dneto0:
Helps scaling of DefUseManager on modules with many thousands
of instructions.


dneto0:
This might help #693 get over the hump

But really we need DefUseManager to scale better https://github.com/KhronosGroup/SPIRV-Tools/issues/846 
Or we need a complete overhaul of instructions and their uses.

AWoloszyn:
Looks good to me assuming the bots are happy.

dneto0:
Rebased and pushed into master as 17a843c6b0

Issue #843, opened 2017-09-27
Validate OpBitCount ResultType component bit width.  Ensure they are wide enough to hold width of base as an unsigned value.


dneto0:
Fix the TODO in https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/validate_bitwise.cpp#L239

jcaraban:
How can we possibly create a failing test for this requirement, if even the smallest unsigned integer (u8 = [0,256]) can always hold the width of the largest integer (64b < 256) ?

dneto0:
Aw dang.   SPIR-V can be used with arbitrary bit widths.
But the validation rules say you *only* get certain bit widths unless you enable more by capabilities. And we don't have a capability for any larger than 64 bits.

So you're right, there's no use implementing the check right now.  :-(
Can you remove the TODO?  :-)

dneto0:
Resolved by https://github.com/KhronosGroup/SPIRV-Tools/pull/851

PR #851, opened 2017-09-29
No use to check OpBitCount result width

jcaraban:
removed TODO


dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/843

(@jcaraban: If you write that "Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/843" in your commit message, then when the commit is merged it will close the issue.  That's usually what you want, i.e. when there is a one-to-one correspondence between issue and merge request.)

PR #852, opened 2017-09-29
AggressiveDCE: Fix to not treat parameter memory refs as local

greg-lunarg:
This fixes a bug that incorrectly deletes stores to parameters, which
can be used to return values from functions.


dneto0:
Rebased and pushed into master as da04f5640e1

PR #853, opened 2017-10-01
DiagnosticStream move ctor moves output duties to new object

dneto0:
… object

- Take over contents of the expiring message stream
- Prevent the expiring object from emitting anything during destruction

### DIFF
--- a/source/diagnostic.h
+++ b/source/diagnostic.h
@@ -33,12 +33,14 @@ class DiagnosticStream {
                    spv_result_t error)
       : position_(position), consumer_(consumer), error_(error) {}
 
-  DiagnosticStream(DiagnosticStream&& other)
-      : stream_(other.stream_.str()),
-        position_(other.position_),
-        consumer_(other.consumer_),
-        error_(other.error_) {}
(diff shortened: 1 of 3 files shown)


antiagainst:
Oh, well, I think the purpose of this change is fine but apparently the buildbot is not happy with some details. ;-P

BTW, the first line of the Git message is a little bit too long.

source/diagnostic.h
```diff
@@ -52,9 +54,9 @@ class DiagnosticStream {
   operator spv_result_t() { return error_; }
 
  private:
-  std::stringstream stream_;
+  std::ostringstream stream_;
   spv_position_t position_;
-  const spvtools::MessageConsumer& consumer_;  // Message consumer callback.
+  spvtools::MessageConsumer consumer_;  // Message consumer callback.
```
dneto0: The test segfaults if I leave this as const ref.   So make the constructor copy the message consumer rather than keeping a reference to it.

dneto0:
Please take another look. 
- Now tested on Linux (but not on OSX again).
- I shortened the commit message (basically, responsibilities --> duties)
- DiagnosticStream construct copies the consumer rather than keeping a reference to it.

PR #856, opened 2017-10-03
Refactored MARK-V API

atgoo:
- switched from C to C++
- moved MARK-V model creation from backend to frontend
- The same MARK-V model object can be used to encode/decode multiple
files
- Added MARK-V model factory (currently only one option)
- Added --validate option to spirv-markv (run validation while
encoding/decoding)


source/comp/markv.h
```diff
@@ -0,0 +1,63 @@
...
+
+// Encodes the given SPIR-V binary to MARK-V binary.
+// If |comments| is not nullptr, it would contain a textual description of
+// how encoding was done (with snippets of disassembly and bit sequences).
+spv_result_t SpirvToMarkv(spv_const_context context,
+                             const std::vector<uint32_t>& spirv,
+                             const markv_encoder_options_t& options,
+                             const MarkvModel* markv_model,
+                             std::vector<uint8_t>* markv,
+                             std::string* comments, spv_diagnostic* diagnostic);
```
dneto0: Our C++ APIs use  MessageConsumer to emit messages. (For one thing the spv_diagnostic is not good at sending out more than one message.)

source/comp/markv.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include <string>
+#include <vector>
+
+#include "markv_model.h"
+#include "spirv-tools/libspirv.h"
+
+namespace spvtools {
+
+struct markv_encoder_options_t {
```
dneto0: Prefer CamelCase with leading capital for C++ types.

I usually prefer an opaque options object.   I guess this is ok provided all future additions to the struct get member initializers.  Otherwise it will be too easy for clients to only explicitly initialize some;  and without a member initializer you'll get default zero-initialized (essentially).  But it's a sharp edge that doesn't have to remain.

(Also with an opaque object the setter methods can return a reference to the options object, and then clients can chain setters.  It's a nice pattern.)

source/comp/markv.h
```diff
@@ -0,0 +1,63 @@
...
+// make it more similar to other compressed SPIR-V files to further improve
+// compression of the dataset.
+
+#ifndef SPIRV_TOOLS_MARKV_HPP_
+#define SPIRV_TOOLS_MARKV_HPP_
+
+#include <string>
+#include <vector>
+
+#include "markv_model.h"
```
dneto0: order of includes.  headers from outside this module should precede headers from inside this module.

tools/comp/markv.cpp
```diff
@@ -20,9 +20,10 @@
 #include <memory>
 #include <vector>
 
+#include "markv_model_factory.h"
```
dneto0: order of includes.  Headers from outside this module should precede headers from inside this module.

I'm counting the 'comp' module as separate from the rest of SPIRV-Tools

source/comp/markv.h
```diff
@@ -0,0 +1,63 @@
...
+  bool validate_spirv_binary = false;
+};
+
+// Encodes the given SPIR-V binary to MARK-V binary.
+// If |comments| is not nullptr, it would contain a textual description of
+// how encoding was done (with snippets of disassembly and bit sequences).
+spv_result_t SpirvToMarkv(spv_const_context context,
+                             const std::vector<uint32_t>& spirv,
+                             const markv_encoder_options_t& options,
+                             const MarkvModel* markv_model,
```
dneto0: I believe your code assumes markv_model is non-null.   Since it's const, you should make it const-ref to enforce the contract.

source/comp/markv_codec.cpp
```diff
@@ -586,9 +356,11 @@ class MarkvCodecBase {
   };
 
   explicit MarkvCodecBase(spv_const_context context,
-                          spv_validator_options validator_options)
+                          spv_validator_options validator_options,
+                          const MarkvModel* model)
```
dneto0: Need a comment saying model must be non-null.  Or change it to const ref (even better)

source/comp/markv_codec.cpp
```diff
@@ -778,8 +566,9 @@ class MarkvCodecBase {
 class MarkvEncoder : public MarkvCodecBase {
  public:
   MarkvEncoder(spv_const_context context,
-               spv_const_markv_encoder_options options)
-      : MarkvCodecBase(context, GetValidatorOptions(options)),
+               const markv_encoder_options_t& options,
+               const MarkvModel* model)
```
dneto0: Recommend making model const-ref

source/comp/markv_codec.cpp
atgoo: Using const ref in tha API. I prefer to use a pointer with an object loaned to a class, as it implicitly suggests lifetime. I.e. it can't be done this way
MarkvEncoder encoder(MarkvModel());
encoder.DoSomething();

atgoo:
Fixed some issues. The "headers from outside this module should precede" rule seems to be against the style guide.

source/comp/markv.h
dneto0: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes says

-----------
In dir/foo.cc or dir/foo_test.cc, whose main purpose is to implement or test the stuff in dir2/foo2.h, order your includes as follows:

dir2/foo2.h.
C system files.
C++ system files.
Other libraries' .h files.
Your project's .h files
----------

So it depends on your interpretation of "module".  The rule uses "library" and "project" in a possibly ambiguous way.  I take "module" = "library" and the markv_model.h is in the SPIRV-Tools-comp library, while libspriv.hpp declares content from other SPIRV-Tools libaries.

Do you agree?

source/comp/markv_codec.cpp
dneto0: Ok, so you chose to make it a const pointer.  It would be an improvement to assert that the pointer is not null.

dneto0: I'm a bit confused now.  The new commit deletes the copy and assignment constructors.  So is it still possible to have a dangling reference as you describe?

For example, this looks ok to me, (and passes on an ASAN execution):
```
#include <iostream>

class Foo {
 public:
  explicit Foo(int n) : num_(n) {}
  int num() const { return num_; }

 private:
  Foo(const Foo&) = delete;
  const Foo& operator=(const Foo&) = delete;
  int num_;
};

class Bing {
 public:
  explicit Bing(const Foo& f) : f_(f) {}
  int eval() { return f_.num(); }

 private:
  const Foo& f_;
};

int bar() {
  Bing twelve(Foo(12));
  return twelve.eval() + 3;
}

int main() {
  std::cout << bar() << std::endl;
  return 0;
}
```
In this example Bing holds onto a const reference to a Foo that *looks* like it has expired.  But I think the lifetime of the Foo(12) is extended by Bing's holding onto the reference.

dneto0:
Please take another look at my feedback.

source/comp/markv_codec.cpp
atgoo: I'm not convinced.

C++ 2003 12.2/5
The temporary to which the reference is bound or the temporary that is the complete object to a subobject of which the temporary is bound persists for the lifetime of the reference except as specified below. A temporary bound to a reference member in a constructor’s ctor-initializer (12.6.2) persists until the constructor exits. A temporary bound to a reference parameter in a function call (5.2.2) persists until the completion of the full expression containing the call.

C++ 2011 12.2/5
The temporary to which the reference is bound or the temporary that is the complete object of a subobject to which the reference is bound persists
for the lifetime of the reference except:
(5.1) — A temporary object bound to a reference parameter in a function call (5.2.2) persists until the completion of the full-expression containing the call.
(5.2) — The lifetime of a temporary bound to the returned value in a function return statement (6.6.3) is not extended; the temporary is destroyed at the end of the full-expression in the return statement.
(5.3) — A temporary bound to a reference in a new-initializer (5.3.4) persists until the completion of the full-expression containing the new-initializer.

dneto0:
We talked offline.   @atgoo pointed out that the external API is ok to use const ref because they are function calls.  In that scenario, lifetime extension, which extends until the end of the function call, is good enough.  The inside case is a constructor, and the lifetime extension is good only until the end of the constructor call.  Then the object goes out of scope but with the saved reference now pointing at an invalid object.

So I consider that point resolved.

source/comp/markv.h
dneto0: Perhaps this is too small to argue over.  The rest of SPIRV-Tools is not uniformly good in the header include order department.

dneto0:
Rebased, squashed, and pushed into master as 2401fc0a726ce

PR #860, opened 2017-10-05
Remove duplicate dead branch elim pass declaration

atgoo:
The function had two declarations in the same header with somewhat
different comments.


dneto0:
Rebased and pushed into master as 4b1577a0cbcfdbd

PR #866, opened 2017-10-10
Generic intrusive linked list class.

s-perron:
This commit is the initial implementation of the intrusive linked list
class.  It includes the implementation in the header files, and unit
test.

### DIFF
--- a/source/opt/ilist.h
+++ b/source/opt/ilist.h
@@ -1,0 +1,194 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_OPT_ILIST_H_
+#define LIBSPIRV_OPT_ILIST_H_
+
+#include <cassert>
+#include <type_traits>
+
+#include "ilist_node.h"
+
+namespace spvtools {
+namespace utils {
+
+// An IntrusiveList is a generic implementation of a doubly-linked list.  The
+// intended convention for using this container is:
+//
+//      class Node : public IntrusiveNodeBase<Node> {
+//        // Node that "Node", the class being defined is the template.
+//        // Must have a default constructor accessible to List.
+//        // Add whatever data is needed in the node
+//      };
+//
+//      typedef IntrusiveList<Node> List;
+//
+// You can also inherit from IntrusiveList instead of a typedef if you want to
+// add more functionality.
+//
+// The condition on the template for IntrusiveNodeBase is there to add some type
+// checking to the container.  The compiler will still allow inserting elements
+// of type IntrusiveNodeBase<Node>, but that would be an error. This assumption
+// allows NextNode and PreviousNode to return pointers to Node, and casting will
+// not be required by the user.
+
+template <class NodeType>
+class IntrusiveList {
+ public:
+  static_assert(
+      std::is_base_of<IntrusiveNodeBase<NodeType>, NodeType>::value,
+      "The type from the node must be derived from IntrusiveNodeBase, with "
+      "itself in the template.");
+
+  // Creates an empty list.
+  inline IntrusiveList();
+
+  // Moves the contents of the given list to the list being constructed.
+  IntrusiveList(IntrusiveList&&);
+
+  // Destorys the list.  Note that the elements of the list will not be deleted,
+  // but thy will be removed from the list.
+  ~IntrusiveList();
+
+  // Moves all of the elements in the list on the RHS to the list on the LHS.
+  IntrusiveList& operator=(IntrusiveList&&);
+
+  // Basetype for iterators so an IntrusiveList can be traversed like STL
+  // containers.
+  template <class T>
+  class iterator_template {
+   public:
+    inline iterator_template(const iterator_template& i) : node_(i.node_) {}
+    inline iterator_template& operator++() {
+      node_ = node_->NextNode();
+      return *this;
+    }
+    inline iterator_template& operator--() {
+      node_ = node_->PreviousNode();
+      return *this;
+    }
+    inline iterator_template& operator=(const iterator_template& i) {
+      node_ = i.node_;
+      return *this;
+    }
+    inline T& operator*() const { return *node_; }
+    inline T* operator->() const { return node_; }
+
+    friend inline bool operator==(const iterator_template& lhs,
+                                  const iterator_template& rhs) {
+      return lhs.node_ == rhs.node_;
+    }
+    friend inline bool operator!=(const iterator_template& lhs,
+                                  const iterator_template& rhs) {
+      return !(lhs == rhs);
+    }
+
+   private:
+    inline iterator_template(T* node) { node_ = node; }
+    T* node_;
+
+    friend IntrusiveList;
+  };
+
+  typedef iterator_template<NodeType> iterator;
+  typedef iterator_template<const NodeType> const_iterator;
+
+  iterator begin();
+  iterator end();
+  const_iterator begin() const;
+  const_iterator end() const;
+
+  // Appends |node| to the end of the list.
+  void push_back(NodeType* node);
+
+ private:
+  // Doing a deep copy of the list does not make sense if the list does not own
+  // the data.  It is not clear who will own the newly created data.  Making
+  // copies illegal for that reason.
+  IntrusiveList(const IntrusiveList&);
+  IntrusiveList& operator=(const IntrusiveList&);
+
+  // A special node used to represent both the start and end of the list,
+  // without being part of the list.
+  NodeType sentinel_;
+};
+
+// Implementation of IntrusiveList
+
+template <class NodeType>
+inline IntrusiveList<NodeType>::IntrusiveList() : sentinel_() {
+  sentinel_.next_node_ = &sentinel_;
+  sentinel_.previous_node_ = &sentinel_;
+  sentinel_.is_sentinel_ = true;
+}
+
+template <class NodeType>
+IntrusiveList<NodeType>::IntrusiveList(IntrusiveList&& list) {
+  this->sentinel_ = list.sentinel_;
+  this->sentinel_.next_node_->previous_node_ = &sentinel_;
+  this->sentinel_.previous_node_->next_node_ = &sentinel_;
+
+  list.sentinel_.next_node_ = &list.sentinel_;
+  list.sentinel_.previous_node_ = &list.sentinel_;
+}
+
+template <class NodeType>
+IntrusiveList<NodeType>::~IntrusiveList() {
+  for (auto i : *this) i.RemoveFromList();
+}
+
+template <class NodeType>
+IntrusiveList<NodeType>& IntrusiveList<NodeType>::operator=(
+    IntrusiveList<NodeType>&& list) {
+  this->sentinel_ = list.sentinel_;
+  this->sentinel_.next_node_->previous_node_ = &sentinel_;
+  this->sentinel_.previous_node_->next_node_ = &sentinel_;
+
+  list.sentinel_.next_node_ = &list.sentinel_;
+  list.sentinel_.previous_node_ = &list.sentinel_;
+  return *this;
+}
+
+template <class NodeType>
+inline typename IntrusiveList<NodeType>::iterator
+IntrusiveList<NodeType>::begin() {
+  return iterator(sentinel_.NextNode());
+}
+
+template <class NodeType>
+inline typename IntrusiveList<NodeType>::iterator
+IntrusiveList<NodeType>::end() {
+  return iterator(nullptr);
+}
+
+template <class NodeType>
+inline typename IntrusiveList<NodeType>::const_iterator
+IntrusiveList<NodeType>::begin() const {
+  return const_iterator(sentinel_.NextNode());
+}
+
+template <class NodeType>
+inline typename IntrusiveList<NodeType>::const_iterator
+IntrusiveList<NodeType>::end() const {
+  return const_iterator(nullptr);
+}
+template <class NodeType>
+void IntrusiveList<NodeType>::push_back(NodeType* node) {
+  node->InsertBefore(&sentinel_);
+}
+
+}  // namespace utils
+}  // namespace spvtools
+
+#endif  // LIBSPIRV_OPT_ILIST_H_
--- a/source/opt/ilist_node.h
+++ b/source/opt/ilist_node.h
@@ -1,0 +1,127 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_OPT_ILIST_NODE_H_
+#define LIBSPIRV_OPT_ILIST_NODE_H_
+
+#include <cassert>
+
+namespace spvtools {
+namespace utils {
+
+template <class NodeType>
+class IntrusiveList;
+
+// IntrusiveNodeBase is the class when creating classes than can be inserted
+// into an IntrusiveList.  See the comments in ilist.h on how to use the class.
+
+template <class NodeType>
+class IntrusiveNodeBase {
+ public:
+  // Creates a new node that is not in a list.
+  inline IntrusiveNodeBase();
+
+  // Returns the node that comes after the given node in the list, if one
+  // exists.  If the given node is not in a list or is at the end of the list,
+  // the return value is nullptr.
+  inline NodeType* NextNode() const;
+
+  // Returns the node that comes before the given node in the list, if one
+  // exists.  If the given node is not in a list or is at the start of the
+  // list, the return value is nullptr.
+  inline NodeType* PreviousNode() const;
+
+  // Will insert the given node immediately before pos in the list.
+  // This the given node is already in a list, it will first be removed
+  // from that list.
+  //
+  // It is assumed that the given node is of type NodeType.  It is an error is
+  // pos is not already in a list.
+  inline void InsertBefore(NodeType* pos);
+
+  // Will insert the given node immediately after pos in the list.
+  // This the given node is already in a list, it will first be removed
+  // from that list.
+  //
+  // It is assumed that the given node is of type NodeType.  It is an error is
+  // pos is not already in a list.
+  inline void InsertAfter(NodeType* pos);
+
+  // Removes the given node from the list.  Note that this does not free any
+  // storage related to the node.
+  inline void RemoveFromList();
+
+ private:
+  NodeType* next_node_;
+  NodeType* previous_node_;
+  bool is_sentinel_;
+
+  friend IntrusiveList<NodeType>;
+};
+
+// Implementation of IntrusiveNodeBase
+
+template <class NodeType>
+inline IntrusiveNodeBase<NodeType>::IntrusiveNodeBase()
+    : next_node_(nullptr), previous_node_(nullptr), is_sentinel_(false) {}
+
+template <class NodeType>
+inline NodeType* IntrusiveNodeBase<NodeType>::NextNode() const {
+  if (!next_node_->is_sentinel_) return next_node_;
+  return nullptr;
+}
+
+template <class NodeType>
+inline NodeType* IntrusiveNodeBase<NodeType>::PreviousNode() const {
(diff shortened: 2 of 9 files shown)


s-perron:
This implementation does not contain everything the class will need.  It is just the basics to get started.  We can continue to new functionality as needed.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+
+  // Moves all of the elements in the list on the RHS to the list on the LHS.
+  IntrusiveList& operator=(IntrusiveList&&);
+
+  // Basetype for iterators so an IntrusiveList can be traversed like STL
+  // containers.
+  template <class T>
+  class iterator_template {
+   public:
+    inline iterator_template(const iterator_template& i) : node_(i.node_) {}
```
dneto0: The "inline" on all these is redundant.  I'd remove them.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+// An IntrusiveList is a generic implementation of a doubly-linked list.  The
+// intended convention for using this container is:
+//
+//      class Node : public IntrusiveNodeBase<Node> {
+//        // Node that "Node", the class being defined is the template.
+//        // Must have a default constructor accessible to List.
+//        // Add whatever data is needed in the node
+//      };
+//
+//      typedef IntrusiveList<Node> List;
```
dneto0: Should we prefer a alias declaration via 'using'?

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+   private:
+    inline iterator_template(T* node) { node_ = node; }
+    T* node_;
+
+    friend IntrusiveList;
+  };
+
+  typedef iterator_template<NodeType> iterator;
```
dneto0: I think prefer `using iterator = iterator_template<NodeType>;`

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+  typedef iterator_template<const NodeType> const_iterator;
+
+  iterator begin();
+  iterator end();
+  const_iterator begin() const;
```
dneto0: Would be nice to also have cbegin and cend.

Should these be noexcept as well, to follow the lead of various std::*::begin methods? See http://en.cppreference.com/w/cpp/container/vector/begin

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+  const_iterator end() const;
+
+  // Appends |node| to the end of the list.
+  void push_back(NodeType* node);
```
dneto0: Assumes node is not already on the list?  Should say so.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+
+ private:
+  // Doing a deep copy of the list does not make sense if the list does not own
+  // the data.  It is not clear who will own the newly created data.  Making
+  // copies illegal for that reason.
```
dneto0: Good decision and good comment.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+  // Moves all of the elements in the list on the RHS to the list on the LHS.
+  IntrusiveList& operator=(IntrusiveList&&);
+
+  // Basetype for iterators so an IntrusiveList can be traversed like STL
+  // containers.
+  template <class T>
+  class iterator_template {
+   public:
+    inline iterator_template(const iterator_template& i) : node_(i.node_) {}
+    inline iterator_template& operator++() {
```
dneto0: This method is invalid to call on an end iterator.  If that's intentional then you should have a method comment saying so.
Otherwise guard the assignment, and have a method comment saying what it does on an end iterator.

I see it's useful to have an efficiency gain when you avoid the guard.  I could go either way.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+      node_ = node_->NextNode();
+      return *this;
+    }
+    inline iterator_template& operator--() {
+      node_ = node_->PreviousNode();
```
dneto0: Unsafe to call on an end iterator.  But that's a useful pattern, e.g. scanning a list in reverse order.

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+  // It is assumed that the given node is of type NodeType.  It is an error is
+  // pos is not already in a list.
+  inline void InsertAfter(NodeType* pos);
+
+  // Removes the given node from the list.  Note that this does not free any
+  // storage related to the node.
+  inline void RemoveFromList();
+
+ private:
+  NodeType* next_node_;
```
dneto0: Should state the invariants:

```
The next and previous nodes on this list.  
If this node is on a list, then both next and previous point to distinct nodes on the list.  If this node is not on a list then both are nullptr.
```

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+  NodeType* previous_node_;
+  bool is_sentinel_;
```
dneto0: // True only for the sentinel node stored in the list itself.

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+  // Will insert the given node immediately after pos in the list.
+  // This the given node is already in a list, it will first be removed
+  // from that list.
+  //
+  // It is assumed that the given node is of type NodeType.  It is an error is
+  // pos is not already in a list.
+  inline void InsertAfter(NodeType* pos);
+
+  // Removes the given node from the list.  Note that this does not free any
+  // storage related to the node.
```
dneto0: Assumes this node is on a list.

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+  this->previous_node_ = pos;
+  this->next_node_ = pos->next_node_;
+  pos->next_node_ = static_cast<NodeType*>(this);
+  this->next_node_->previous_node_ = static_cast<NodeType*>(this);
+}
+
+template <class NodeType>
+inline void IntrusiveNodeBase<NodeType>::RemoveFromList() {
+  assert(!this->is_sentinel_ && "Sentinel nodes cannot be moved around.");
+  this->next_node_->previous_node_ = this->previous_node_;
```
dneto0: assert( previous_node_)
assert( next_node_ )

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+    friend IntrusiveList;
+  };
+
+  typedef iterator_template<NodeType> iterator;
+  typedef iterator_template<const NodeType> const_iterator;
+
+  iterator begin();
+  iterator end();
+  const_iterator begin() const;
+  const_iterator end() const;
```
atgoo: You might also want to iterate back-to-front.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+inline typename IntrusiveList<NodeType>::const_iterator
+IntrusiveList<NodeType>::begin() const {
+  return const_iterator(sentinel_.NextNode());
+}
+
+template <class NodeType>
+inline typename IntrusiveList<NodeType>::const_iterator
+IntrusiveList<NodeType>::end() const {
+  return const_iterator(nullptr);
+}
```
atgoo: Nit: missing blank line.

test/opt/ilist_test.cpp
```diff
@@ -0,0 +1,222 @@
...
+// nodes to build the list.
+TestList BuildList(TestNode nodes[], int n) {
+  TestList list;
+  for (int i = 0; i < n; i++) {
+    nodes[i].data = i;
+    list.push_back(&nodes[i]);
+  }
+  return list;
+}
+
```
dneto0: The iterators are tested in various ways by the range-for constructs.

I'd write specific tests for begin(), end(), increment and decrement.  In particular, there's a possible ambiguity about what ++ and -- do on end and begin iterators (respectively.)

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+  // Basetype for iterators so an IntrusiveList can be traversed like STL
+  // containers.
+  template <class T>
+  class iterator_template {
+   public:
+    inline iterator_template(const iterator_template& i) : node_(i.node_) {}
+    inline iterator_template& operator++() {
+      node_ = node_->NextNode();
+      return *this;
+    }
```
atgoo: Nit: missing blank line.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+    inline iterator_template& operator=(const iterator_template& i) {
+      node_ = i.node_;
+      return *this;
+    }
+    inline T& operator*() const { return *node_; }
+    inline T* operator->() const { return node_; }
+
+    friend inline bool operator==(const iterator_template& lhs,
```
atgoo: "friend" should be used to decorate an externally defined operator. Am I missing something?

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+                                  const iterator_template& rhs) {
+      return lhs.node_ == rhs.node_;
+    }
+    friend inline bool operator!=(const iterator_template& lhs,
+                                  const iterator_template& rhs) {
+      return !(lhs == rhs);
+    }
+
+   private:
+    inline iterator_template(T* node) { node_ = node; }
```
atgoo: Consider defining or explicitly disabling the default constructor.

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+
+template <class NodeType>
+inline void IntrusiveNodeBase<NodeType>::InsertBefore(NodeType* pos) {
+  assert(!this->is_sentinel_ && "Sentinel nodes cannot be moved around.");
+  assert(pos->previous_node_ != nullptr && "Pos should already be in a list.");
+  if (this->previous_node_ != nullptr) this->RemoveFromList();
+
+  this->next_node_ = pos;
+  this->previous_node_ = pos->previous_node_;
+  pos->previous_node_ = static_cast<NodeType*>(this);
```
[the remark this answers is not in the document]
s-perron: They do not have the same type.  "this" has type IntrusiveNodeBase<NodeType>.

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+  // Returns the node that comes before the given node in the list, if one
+  // exists.  If the given node is not in a list or is at the start of the
+  // list, the return value is nullptr.
+  inline NodeType* PreviousNode() const;
+
+  // Will insert the given node immediately before pos in the list.
+  // This the given node is already in a list, it will first be removed
+  // from that list.
+  //
+  // It is assumed that the given node is of type NodeType.  It is an error is
```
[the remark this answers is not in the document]
s-perron: I will still compile if "this" points to an IntrusiveNodeBase<NodeType>, so the compiler will allow them to be added to the list.  We could could get the compiler to flag this as an error if we make the insert functions private in the base class.  Then they will have to be made explicitly public in NodeType.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
+// Copyright (c) 2017 Google Inc.
```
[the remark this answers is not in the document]
s-perron: Good call.  Done.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+#include "ilist_node.h"
+
+namespace spvtools {
+namespace utils {
+
+// An IntrusiveList is a generic implementation of a doubly-linked list.  The
+// intended convention for using this container is:
+//
+//      class Node : public IntrusiveNodeBase<Node> {
+//        // Node that "Node", the class being defined is the template.
```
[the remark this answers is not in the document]
s-perron: Done

source/opt/ilist.h
s-perron: done

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+      "itself in the template.");
+
+  // Creates an empty list.
+  inline IntrusiveList();
+
+  // Moves the contents of the given list to the list being constructed.
+  IntrusiveList(IntrusiveList&&);
+
+  // Destorys the list.  Note that the elements of the list will not be deleted,
+  // but thy will be removed from the list.
```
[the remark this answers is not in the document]
s-perron: Done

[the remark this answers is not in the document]
s-perron: Thanks, and fixed.

source/opt/ilist.h
s-perron: When I was thinking of operator++ and operator--, it does not make sense to increment end().  At the same time, I was think that we could add reverse iterators (with rbegin and rend) if we want to traverse a list in reverse.  However, looking on stackoverflow, it seems like people will decrement end().  I'll make that legal.

I'll make end() point to the sentinel node, but this means I can't use PreviousNode() and NexNode() to move the iterator.

source/opt/ilist.h
s-perron: done

source/opt/ilist.h
s-perron: done

source/opt/ilist.h
s-perron: The "friend" is needed because these are not member functions.  They are in the namespace, but not the class, and they need access to the members of the class.  "friend" is needed even if they are defined inline.

Note that I defined them inline instead of just a prototype in the class because the template instantiations were getting really hard to get right with the templated class containing a templated class.  That is why I defined everything in the Iterator class.

source/opt/ilist.h
s-perron: Done

source/opt/ilist.h
s-perron: Done

source/opt/ilist.h
s-perron: Done

source/opt/ilist.h
s-perron: Yes, but I was think that would be one of the things we can add as we need it.  I can see adding new stuff to the class as needed.

source/opt/ilist.h
s-perron: Add a comment.  |node| will be removed from any list it is already it first.

source/opt/ilist.h
```diff
@@ -0,0 +1,194 @@
...
+  const_iterator end() const;
+
+  // Appends |node| to the end of the list.
+  void push_back(NodeType* node);
+
+ private:
+  // Doing a deep copy of the list does not make sense if the list does not own
+  // the data.  It is not clear who will own the newly created data.  Making
+  // copies illegal for that reason.
+  IntrusiveList(const IntrusiveList&);
```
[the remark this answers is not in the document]
s-perron: Good call.  Done.

source/opt/ilist.h
s-perron: done

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+
+#include <cassert>
+
+namespace spvtools {
+namespace utils {
+
+template <class NodeType>
+class IntrusiveList;
+
+// IntrusiveNodeBase is the class when creating classes than can be inserted
```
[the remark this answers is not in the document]
s-perron: Fixed the comment.

I can change the two spaces at the end of sentences.  Most style guides say 1 space, but that is because generally people don't use monospace fonts.  Not us programmers.

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+  // exists.  If the given node is not in a list or is at the end of the list,
+  // the return value is nullptr.
+  inline NodeType* NextNode() const;
+
+  // Returns the node that comes before the given node in the list, if one
+  // exists.  If the given node is not in a list or is at the start of the
+  // list, the return value is nullptr.
+  inline NodeType* PreviousNode() const;
+
+  // Will insert the given node immediately before pos in the list.
```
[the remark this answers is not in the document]
s-perron: done

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+  // This the given node is already in a list, it will first be removed
+  // from that list.
```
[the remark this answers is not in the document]
s-perron: done

source/opt/ilist_node.h
```diff
@@ -0,0 +1,127 @@
...
+  //
+  // It is assumed that the given node is of type NodeType.  It is an error is
+  // pos is not already in a list.
+  inline void InsertBefore(NodeType* pos);
+
+  // Will insert the given node immediately after pos in the list.
```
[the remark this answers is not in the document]
s-perron: done

source/opt/ilist_node.h
s-perron: done

source/opt/ilist_node.h
s-perron: done

source/opt/ilist_node.h
s-perron: done

source/opt/ilist_node.h
s-perron: done

test/opt/CMakeLists.txt
```diff
@@ -183,3 +183,8 @@ add_spvtools_unittest(TARGET pass_strength_reduction
   SRCS strength_reduction_test.cpp pass_utils.cpp
   LIBS SPIRV-Tools-opt
 )
+
+add_spvtools_unittest(TARGET util_intrusive_list
+  SRCS ilist_test.cpp pass_utils.cpp
```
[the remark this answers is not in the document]
s-perron: removed.

test/opt/ilist_test.cpp
s-perron: done

dneto0:
VisualStudio 2013 gives error C3646: 'noexcept': unknown override specifier.   :-(

atgoo:
...In the ISO C++11 Standard, noexcept operator was introduced and is supported in Visual Studio 2015 and later...
https://msdn.microsoft.com/en-us/library/wfa0edys.aspx

You would probable have to do something with MSC_VER.

source/util/ilist.h
```diff
@@ -179,37 +183,37 @@ IntrusiveList<NodeType>& IntrusiveList<NodeType>::operator=(
...
 template <class NodeType>
 inline typename IntrusiveList<NodeType>::const_iterator
-IntrusiveList<NodeType>::cbegin() const noexcept {
+IntrusiveList<NodeType>::cbegin() const {
   return const_iterator(sentinel_.next_node_);
 }
 
 template <class NodeType>
 inline typename IntrusiveList<NodeType>::const_iterator
-IntrusiveList<NodeType>::cend() const noexcept {
```
atgoo: You could still have it as NOEXCEPT which would be nothing for MSVC under 2015. But I don't insist, it was David's idea to use noexcept in the first place.

source/util/ilist.h
```diff
@@ -95,29 +99,42 @@ class IntrusiveList {
...
-  iterator begin();
-  iterator end();
-  const_iterator begin() const;
-  const_iterator end() const;
-
-  // Appends |node| to the end of the list.
+  using iterator = iterator_template<NodeType>;
+  using const_iterator = iterator_template<const NodeType>;
+
+  // Various types of iterators for the start (begin) and one past the end (end)
```
dneto0: suggest changing "one past the end" to "one past the last element".
Fortunately your next sentences make it clear.

test/util/ilist_test.cpp
```diff
@@ -65,6 +65,46 @@ TestList BuildList(TestNode nodes[], int n) {
   return list;
 }
 
+// Test decrementing begin()
+TEST(IListTest, DecrementingBegin) {
```
dneto0: Thanks for these extra tests.

The wraparound behaviour of the iterators is good to document, but hopefully nobody relies on it.

source/util/ilist.h
```diff
@@ -95,29 +99,42 @@ class IntrusiveList {
...
+  // of the list.
+  //
+  // Decrementing |end()| iterator will give and iterator pointing to the last
+  // element in the list, if one exists.
+  //
+  // Incrementing |end()| iterator will give |begin()|.
```
dneto0: This wraparound behaviour is good to document as an internal behaviour, as it makes operation very clear.
Hopefully nobody relies on it.  It's an implementation detail.  I'm tempted to say that here.  But this feedback is below the bar.

source/opt/ilist_node.h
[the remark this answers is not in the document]
dneto0: I'm pretty old school.  I learned touch-typing on a mechanical typewriter and am in the habit of two spaces after a period.  I'm ok with others doing it either single-space or double-space.

source/util/ilist.h
dneto0: Ok to leave out.  :-)

dneto0: We compile without exceptions anyway.

dneto0:
Rebased, squashed, and pushed into master as 720beb161a7c3802cbc

source/util/ilist_node.h
```diff
@@ -0,0 +1,135 @@
...
+// IntrusiveNodeBase is the base class for nodes in an IntrusiveList.
+// See the comments in ilist.h on how to use the class.
+
+template <class NodeType>
+class IntrusiveNodeBase {
+ public:
+  // Creates a new node that is not in a list.
+  inline IntrusiveNodeBase();
+
+  // Returns the node that comes after the given node in the list, if one
```
dnovillo: Comment out of date here?  The next node is the next node for the calling instance, right?  The function doesn't take a parameter.

source/util/ilist_node.h
```diff
@@ -0,0 +1,135 @@
...
+  // exists.  If the given node is not in a list or is at the end of the list,
+  // the return value is nullptr.
+  inline NodeType* NextNode() const;
+
+  // Returns the node that comes before the given node in the list, if one
```
dnovillo: Likewise here.

Issue #770, opened 2017-08-22
Validator should check storage class of builtin variables


ianromanick:
For example, this SPIR-V shader says that FragCoord is an output, and the validator is perfectly happy with that.

    000000 07230203 00010200 00000000 00000010
    000010 00000000 00020011 00000001 0003000e
    000020 00000000 00000000 0005000f 00000000
    000030 00000001 6e69616d 00000000 0005000f
    000040 00000004 00000002 6e69616d 00000000
    000050 00040047 00000003 0000000b 00000000
    000060 00040047 0000000f 0000000b 0000000f
    000070 00020013 00000004 00030021 00000005
    000080 00000004 00030016 00000006 00000020
    000090 00040017 00000007 00000006 00000004
    0000a0 00040020 00000008 00000001 00000007
    0000b0 00040020 00000009 00000003 00000007
    0000c0 0004003b 00000009 0000000f 00000003
    0000d0 0004003b 00000008 00000003 00000001
    0000e0 0004003b 00000009 0000000c 00000003
    0000f0 0004002b 00000006 0000000b 00000000
    000100 0007002c 00000007 0000000a 0000000b
    000110 0000000b 0000000b 0000000b 00050036
    000120 00000004 00000001 00000000 00000005
    000130 000200f8 0000000d 0003003e 00000003
    000140 0000000a 000100fd 00010038 00050036
    000150 00000004 00000002 00000000 00000005
    000160 000200f8 0000000e 0003003e 0000000c
    000170 0000000a 000100fd 00010038

dneto0:
The validator is incomplete.  Unfinished work is broken down and tracked in this "project": https://github.com/KhronosGroup/SPIRV-Tools/projects/1

dneto0:
Suggest for @jcaraban.  Intel could have been helped by this check.

jcaraban:
Do we agree that these two SPIR-V codes should trigger the error? First, the output Position builtin is decorated as input, second, the input FragCoord builtin is decorated as output.
```llvm
          OpCapability Shader
          OpCapability Linkage
          OpMemoryModel Logical GLSL450
          OpDecorate %bltin BuiltIn Position
%float  = OpTypeFloat 32
%v4f    = OpTypeVector %float 4
%v4fin  = OpTypePointer Input %v4f
%bltin  = OpVariable %v4fin Input
```
```llvm
          OpCapability Shader
          OpCapability Linkage
          OpMemoryModel Logical GLSL450
[...]

PR #867, opened 2017-10-10
WIP built-ins tests and validation

jcaraban:
This WIP PR incorporates tests and validation for built-in variables.
It started as a quick fix to #770 (check storage class of built-ins),
but I realized that built-ins require a whole section on their own.
Therefore I made some decisions, and I would like to hear if your opinion.

- Split the Built-ins validation and testing from the Decorations
- Used parameterized tests, because most validation logic is repetitive
- Used individual tests, for built-ins presenting exceptional conditions
- Structured the code to be easily extended with future built-ins

Also, the logic in validate_builtins.cpp is so repetitive,
that it could pay off to auto-generate it from some description.

Finally, I noticed that some related validations are happening in other files,
e.g., below this capabilities restrictions on PrimitiveId is checked in valdiate_instructions.cpp

_“In a fragment shader, any variable decorated with PrimitiveId must be declared using the Input storage class, and either the Geometry or Tessellation capability must also be declared.”_

**Note:** work in process, submitting to get early feedback

### DIFF
--- a/source/validate_builtins.cpp
+++ b/source/validate_builtins.cpp
@@ -32,7 +32,7 @@ namespace { // utils
 enum AnyValue { kAnyBits=-3, kAnySize=-2, kAnySign=-1 };
 enum Signedness { kUnsigned=0, kSigned=1 };
 
-// Struct modeling the OpTypeInt/Float/Bool "elemental" types 
+// Struct representing the OpTypeInt/Float/Bool "elemental" types
 struct ElemType {
   SpvOp type; // Elemental type (e.g. int, float, bool)
   int bits; // Bits number (e.g. 32, 16, 64), 0 means any number
@@ -42,7 +42,7 @@ struct ElemType {
     : type(type_), bits(bits_), sign(sign_) { }
 };
 
-// Struct modeling the OpTypeVector/Array "composed" types
+// Struct representing the OpTypeVector/Array "composed" types
 struct CompType {
   SpvOp type; // Composed type (e.g. vector, array)
   int size; // Number of components, 0 means any number
@@ -52,6 +52,7 @@ struct CompType {
 };
 
 // Returns the integer signedness of the target environment
+// Depending on SPV_ENV_VULKAN / CL, the int is signed or unsgined
 int Sign(ValidationState_t& vstate) {
   switch (vstate.context()->target_env) {
     case SPV_ENV_VULKAN_1_0:
@@ -60,10 +61,10 @@ int Sign(ValidationState_t& vstate) {
     case SPV_ENV_OPENGL_4_2:
     case SPV_ENV_OPENGL_4_3:
     case SPV_ENV_OPENGL_4_5:
-      return kSigned;
+      return kSigned; // Shader models prefer integers ints
     case SPV_ENV_OPENCL_2_1:
     case SPV_ENV_OPENCL_2_2:
-      return kUnsigned;
+      return kUnsigned; // Kernel model required unsigned ints
     default:
       return kAnySign;
   }
@@ -162,7 +163,7 @@ std::vector<uint32_t> getExecutionModes(const Instruction* entry_point) {
 // Ensures |exec_model| is included in the list of compatible |models|
 spv_result_t CheckExecutionModel(ValidationState_t& vstate, const string& name,
       uint32_t exec_model, const std::vector<uint32_t>& models) {
-  if (not IsIncluded(exec_model, models)) {
+  if (!IsIncluded(exec_model, models)) {
     return vstate.diag(SPV_ERROR_INVALID_ID) << name
         << " built-in is restricted to certain EXECUTION MODELS"
            " (see SPIR-V, Vulkan, OpenGL, OpenCL specifications)";
@@ -177,7 +178,7 @@ spv_result_t CheckStorageClass(ValidationState_t& vstate,
   // If 'exec_model' is included in the list of 'models'
   if (IsIncluded(exec_model, models)) {
     // But 'storage_class' is not one of the compatible 'storages'
-    if (not IsIncluded(storage_class, storages)) {
+    if (!IsIncluded(storage_class, storages)) {
       return vstate.diag(SPV_ERROR_INVALID_ID) << name
           << " built-in is restricted to certain STORAGE CLASSES,"
              " depending on the execution model (see SPIR-V spec)";
@@ -543,8 +544,8 @@ spv_result_t CheckPrimitiveId(ValidationState_t& vstate, const Instruction* inst
   //                 should it be tested in validate_builtins.cpp instead?
 
   #if 0
-  if (not vstate.HasCapability(SpvCapabilityGeometry) &&
-      not vstate.HasCapability(SpvCapabilityTessellation)) {
+  if (!vstate.HasCapability(SpvCapabilityGeometry) &&
+      !vstate.HasCapability(SpvCapabilityTessellation)) {
     return vstate.diag(SPV_ERROR_INVALID_ID)
         << "In a fragment shader, any variable decorated with "
            "PrimitiveId must be declared using the Input storage "
@@ -993,7 +994,7 @@ spv_result_t CheckFragDepth(ValidationState_t& vstate, const Instruction* inst,
   auto entry_vec = getEntryPoints(vstate,inst);
   for (auto entry_point : entry_vec) {
     auto mode_vec = getExecutionModes(entry_point);
-    if (not IsIncluded(SpvExecutionModeDepthReplacing,mode_vec)) {
+    if (!IsIncluded(SpvExecutionModeDepthReplacing,mode_vec)) {
       return vstate.diag(SPV_ERROR_INVALID_ID)
           << "To write to FragDepth, a shader must declare the "
             "DepthReplacing execution mode (see Vulkan spec)";
@@ -1070,14 +1071,12 @@ spv_result_t CheckNumWorkgroups(ValidationState_t& vstate, const Instruction* in
   CompType comp = { SpvOpTypeVector, 3 };
   if (auto error = CheckBuiltInType(vstate, name, inst, elem, comp))
     return error;
-  
-  // TODO(jcaraban): depending on SPV_ENV_VULKAN / CL, the int is signed or unsgined
 
   return SPV_SUCCESS;
 }
 
 spv_result_t CheckWorkgroupSize(ValidationState_t& vstate, const Instruction* inst,
-    uint32_t storage_class, uint32_t exec_model)
+    uint32_t , uint32_t exec_model)
 {
   const string name = "WorkgroupSize";
   std::vector<uint32_t> exec_vec, stor_vec;
@@ -1085,8 +1084,7 @@ spv_result_t CheckWorkgroupSize(ValidationState_t& vstate, const Instruction* in
   // If an object is decorated with the WorkgroupSize decoration,
   // this must take precedence over any execution mode set for LocalSize
 
-  // TODO(jcaraban): can this be verified now?
-  //                 i.e. is this a compile or runtime restriction?
+  // TODO(jcaraban): can this be verified? or is this a rule for the drivers?
 
   // The WorkgroupSize decoration must be used only within compute shaders
   exec_vec = { SpvExecutionModelGLCompute,
@@ -1104,8 +1102,7 @@ spv_result_t CheckWorkgroupSize(ValidationState_t& vstate, const Instruction* in
   }
 
   // WrokgroupSize should not have storage class, and if it does,
-  // other validation should complain before reaching this line
-  assert(storage_class == SpvStorageClassMax);
+  // other validation pass should complain about it
   
   // The object decorated with WorkgroupSize must be declared as
   // a three-component vector of 32-bit integers.
@@ -2268,7 +2265,7 @@ spv_result_t ValidateBuiltIns(ValidationState_t& _)
   // For every instruction decorated with a built-in...
   for (const auto& inst_ref : _.ordered_instructions()) {
     const Instruction *inst = &inst_ref;
-    if (not IsBuiltIn(inst, _))
+    if (!IsBuiltIn(inst, _))
       continue;
     auto storage_class = GetStorageClass(inst);
     auto built_in = GetBuiltInEnum(inst, _);
--- a/source/validate_builtins.cpp
+++ b/source/validate_builtins.cpp
@@ -21,6 +21,7 @@
 #include "opcode.h"
 #include "val/validation_state.h"
 
+using std::string;
 using libspirv::Decoration;
 using libspirv::DiagnosticStream;
 using libspirv::Instruction;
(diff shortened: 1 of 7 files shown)


source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+  }
+  return exec_vec;
+}
+
+// Ensures the excution model is included in the list of compatible models
+spv_result_t checkExecutionModel(ValidationState_t& vstate,
+      uint32_t exec_model, std::vector<uint32_t> models) {
+  if (not isIncludedIn(exec_model,models)) {
+    return vstate.diag(SPV_ERROR_INVALID_ID)
+        << "Built-in variables are restricted to only certain"
```
dneto0: It's much more friendly for the error message to state the builtin and the execution model that it was found in.

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+spv_result_t checkStorageClass(ValidationState_t& vstate,
+      uint32_t exec_model, uint32_t storage_class,
+      std::vector<uint32_t> models, std::vector<uint32_t> storages)
+{
+  // If 'exec_model' is included the list of 'models'
+  if (isIncludedIn(exec_model,models)) {
+    // But 'storage_class' is not one of the compatible 'storages'
+    if (not isIncludedIn(storage_class,storages)) {
+      return vstate.diag(SPV_ERROR_INVALID_ID)
+          << "Built-in variables must match the STORAGE CLASS "
```
dneto0: A more friendly message would state the storage class and execution model.

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+    }
+  }
+  return SPV_SUCCESS;
+}
+
+// Ensures the variable type matches the built-in type
+spv_result_t checkBuiltInType(ValidationState_t& vstate, Instruction inst,
+      SpvOp elem_type, SpvOp comp_type=SpvOpNop, uint32_t comp_size=0)
+{
+  auto error = [&](ValidationState_t& state) -> spv_result_t {
```
dneto0: You should only apply these rules in certain target environments.    If you're only sure about Vulkan then be specific to Vulkan.  There might be large overlap with OpenGL's use of SPIR-V. (e.g. OpenGL 4.6)

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+  };
+  auto *curr_inst = &inst;
+  // Gets instruction type
+  curr_inst = vstate.FindDef(curr_inst->type_id());
+  assert(curr_inst);
+  // The built-in variable must be of OpType Pointer
+  if (curr_inst->opcode() != SpvOpTypePointer) {
+    return error(vstate);
+  }
+  // The Pointer type is stored in its 3rd word
```
dneto0: Recommend rephrasing "The type pointed to is given in word 3."

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+      return vstate.diag(SPV_ERROR_INVALID_ID)
+          << "Built-in variables must match the STORAGE CLASS "
+             "of the built-in (see e.g. Vulkan specification)";
+    }
+  }
+  return SPV_SUCCESS;
+}
+
+// Ensures the variable type matches the built-in type
+spv_result_t checkBuiltInType(ValidationState_t& vstate, Instruction inst,
```
dneto0: Need a better function comment.
What are comp_type and comp_size?

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+        BUILTIN_CASE( SampleMask                  )
+        BUILTIN_CASE( FragDepth                   )
+        BUILTIN_CASE( HelperInvocation            )
+        BUILTIN_CASE( NumWorkgroups               )
+        BUILTIN_CASE( WorkgroupSize               )
+        BUILTIN_CASE( WorkgroupId                 )
+        BUILTIN_CASE( LocalInvocationId           )
+        BUILTIN_CASE( GlobalInvocationId          )
+        BUILTIN_CASE( LocalInvocationIndex        )
+        // TODO(jcaraban): the built-ins below are not clearly documented
```
dneto0: Not all builtins are used by Vulkan.    If the Vulkan environment spec doesn't describe it then it's not allowed in Vulkan.

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+  return SPV_SUCCESS;
+}
+
+spv_result_t CheckInstanceIndex(ValidationState_t& vstate, Instruction inst,
+    uint32_t storage_class, uint32_t exec_model)
+{
+  // TODO(jcaraban): many built-ins show same logic, e.g. Vertex/InstanceIndex
+  return CheckVertexIndex(vstate,inst,storage_class,exec_model);
+}
+/*
```
dneto0: Use #if 0 and a TODO comment.
That's less fragile than a multi-line comment.

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+
+  // The variable decorated with GlobalInvocationId must be declared as
+  // a scalar 32-bit integer.
+  SpvOp elem_type = SpvOpTypeInt;
+  if (auto error = checkBuiltInType(vstate,inst,elem_type))
+    return error;
+
+  return SPV_SUCCESS;
+}
+/*
```
dneto0: Use #if 0 instead of multi-line comment.
And add a TODO

test/val/val_builtins_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+using std::string;
+using std::vector;
+using std::tie;
+using std::tuple;
+using std::make_tuple;
+using ::testing::HasSubstr;
+using ::testing::Eq;
+using libspirv::Decoration;
+
+using tuple4string = tuple<string,string,string,string>;
```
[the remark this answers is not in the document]
jcaraban: Great idea, will do

test/val/val_builtins_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+    %label = OpLabel
+             OpReturn
+             OpFunctionEnd
+  )";
+
+  return capabilities + extensions + memory + entries
+         + modes + decorations + types + definitions;
+}
+
+// clang-format off
```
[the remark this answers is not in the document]
jcaraban: Understood

test/val/val_builtins_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+#define BUILTIN(builtin,type,storage,exec) \
+  make_tuple(string(builtin),string(type),string(storage),string(exec))
+
+TEST_P(ValidateBuiltIn, BuiltInValid) {
+  CompileSuccessfully(GenerateShaderCode(GetParam()));
```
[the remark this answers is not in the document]
jcaraban: Not sure how to do this @dneto0, as simple as passing SPV_ENV_VULKAN_1_0 to CompileSuccessfully? Shall I then split the tests in Vulkan and OpenCL parts?

test/val/val_builtins_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+       BUILTIN("FragCoord","%vecf4","Input","Fragment"),
+       BUILTIN("PointCoord","%vecf2","Input","Fragment"),
+       BUILTIN("FrontFacing","%bool","Input","Fragment"),
+       BUILTIN("SampleId","%uint32","Input","Fragment"),
+       BUILTIN("SamplePosition","%vecf2","Input","Fragment"),
+       BUILTIN("SampleMask","%arru2","Input","Fragment"),
+       BUILTIN("FragDepth","%float","Output","Fragment"),
+       BUILTIN("HelperInvocation","%bool","Input","Fragment"),
+       BUILTIN("NumWorkgroups","%vecu3","Input","GLCompute"),
+       //BUILTIN("WorkgroupSize","%vecu3","Uniform","GLCompute"),
```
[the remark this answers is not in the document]
jcaraban: I didn't know how to treat WorkgroupSize because of the following rule: _"The object decorated with WorkgroupSize must be a specialization constant or a constant."_ I think it requires individual tests

test/val/val_builtins_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+       BUILTIN("WorkgroupId","%vecu3","Input","GLCompute"),
+       BUILTIN("LocalInvocationId","%vecu3","Input","GLCompute"),
+       BUILTIN("GlobalInvocationId","%vecu3","Input","GLCompute"),
+       BUILTIN("LocalInvocationIndex","%uint32","Input","GLCompute"),
+       // WorkDim, GLobalSize, EnqueueWorkgroupSize, GLobalOffset
```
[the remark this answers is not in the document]
jcaraban: The Vulkan Spec is very clear about its built-ins, but the OpenCL Spec not so much. For example, WorkDim is only mentioned as an scalar argument to clEnqueueNDRange, but I could not find a formal definition anywhere. I will try to figure these built-ins out.

test/val/val_builtins_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+       // GLobalLinearId, SubgroupSize, SubgroupMaxSize, NumSubgroups
+       // NumEnqueuedSubgroups, SubgroupId, SubroupLocalInvocationId,
+       BUILTIN("VertexIndex","%uint32","Input","Vertex"),
+       BUILTIN("InstanceIndex","%uint32","Input","Vertex")
+       // BaseVertex, BaseInstance, DrawIndex, DeviceIndex, ViewIndex
+  ), );
+
+
+TEST_P(ValidateBuiltInExecutionModel, BuiltInExecutionModel) {
+  CompileSuccessfully(GenerateShaderCode(GetParam()));
```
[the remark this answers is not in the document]
jcaraban: What about the OpenCL built-ins such as WorkDim, GlobalSize, etc? Shall I move those to another test case that sets the OpenCL environment?

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+#include "val/validation_state.h"
+
+using libspirv::Decoration;
+using libspirv::DiagnosticStream;
+using libspirv::Instruction;
+using libspirv::ValidationState_t;
+
+namespace { // utils
+
+bool isIncludedIn(uint32_t elem, std::vector<uint32_t> vec) {
```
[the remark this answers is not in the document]
jcaraban: Roger

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+  switch (inst.opcode()) {
+    case SpvOpTypePointer: return inst.words()[2];
+    case SpvOpTypeForwardPointer: return inst.words()[2];
+    case SpvOpVariable: return inst.words()[3];
+    case SpvOpGenericCastToPtrExplicit: return inst.words()[4];
+    default: return SpvStorageClassMax;
+  }
+}
+
+// TODO(jcaraban): how to find the entry_points in a more direct way?
```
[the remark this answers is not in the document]
jcaraban: Makes sense, the current representation is a little awkward in some case

source/validate_builtins.cpp
jcaraban: Agree, it was in my plans. I will update the tests accordingly.

source/validate_builtins.cpp
jcaraban: sure

source/validate_builtins.cpp
jcaraban: I'm confused by the target environments, need to have a look at the specs

jcaraban:
Thanks, I will resubmit once I figure out the issue with the target environments, which is confusing me at the moment. Would it be a good idea to splits the tests in Vulkan and OpenCL parts?

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+           "EXECUTION MODELS (see e.g. Vulkan specification)";
+  }
+  return SPV_SUCCESS;
+}
+
+// Ensures the pair 'exec_model','storage_class' are compatible
+spv_result_t checkStorageClass(ValidationState_t& vstate,
+      uint32_t exec_model, uint32_t storage_class,
+      std::vector<uint32_t> models, std::vector<uint32_t> storages)
+{
```
atgoo: Please move '{' to the previous line.
void Func() {
}

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+  // If 'exec_model' is included the list of 'models'
```
atgoo: in the list

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+// Ensures the variable type matches the built-in type
+spv_result_t checkBuiltInType(ValidationState_t& vstate, Instruction inst,
+      SpvOp elem_type, SpvOp comp_type=SpvOpNop, uint32_t comp_size=0)
+{
+  auto error = [&](ValidationState_t& state) -> spv_result_t {
+    return state.diag(SPV_ERROR_INVALID_ID)
+        << "Built-in variables must match the DATA TYPE "
+           "of the built-in (see e.g. Vulkan specification)";
+  };
+  auto *curr_inst = &inst;
```
atgoo: auto* curr_inst
or better
Instruction* curr_inst

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+  // Gets instruction type
+  curr_inst = vstate.FindDef(curr_inst->type_id());
+  assert(curr_inst);
+  // The built-in variable must be of OpType Pointer
+  if (curr_inst->opcode() != SpvOpTypePointer) {
+    return error(vstate);
+  }
+  // The Pointer type is stored in its 3rd word
+  curr_inst = vstate.FindDef(curr_inst->words()[3]);
```
atgoo: Could also be
curr_inst->word(3)

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+    // It must match the composed type...
+    if (curr_inst->opcode() != comp_type) {
+      return error(vstate);
+    }
+    // ... and it must match the number of components / size
+    uint32_t size;
+    if (comp_type == SpvOpTypeVector) { // size as literal
+      size = curr_inst->words()[3];
+    } else if (comp_type == SpvOpTypeArray) { // size as constant
+      auto *size_inst = vstate.FindDef(curr_inst->words()[3]);
```
atgoo: auto* size_inst

source/validate_builtins.cpp
```diff
@@ -0,0 +1,1474 @@
...
+      assert(size_inst);
+      size = size_inst->words()[3];
```
atgoo: Can the size be defined with OpSpecConstant?

jcaraban: Yes, and it should work since both OpConstant and OpSpecConstant hold the literal in the word 3. BTW, shall I check whether the instruction is one of these two, or such validation logic belongs somewhere else?

test/val/val_builtins_test.cpp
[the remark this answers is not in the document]
jcaraban: Another issue with the struct solution is that it makes the testing less readable. So, while the tuple version shows failed test like: 
`[  FAILED  ] BuiltInValid/ValidateBuiltIn.BuiltInValid/0, where GetParam() = tuple<"Position","%vecf4","Output","Vertex">`,
 the struct version dumps this info:
 `[  FAILED  ] BuiltInValid/ValidateBuiltIn.BuiltInValid/0, where GetParam() = 128-byte object <A0-71 40-02 00-00 00-00 08-00 00-00 00-00 00-00 50-6F 73-69 74-69 6F-6E 00-00 00-00 00-00 00-00 C0-71 40-02 00-00 00-00 06-00 00-00 00-00 00-00 25-76 65-63 66-34 00-00 00-00 00-00 00-00 00-00 E0-71 40-02 00-00 00-00 06-00 00-00 00-00 00-00 4F-75 74-70 75-74 00-00 00-00 00-00 00-00 00-00 00-72 40-02 00-00 00-00 06-00 00-00 00-00 00-00 56-65 72-74 65-78 00-00 00-00 00-00 00-00 00-00> (0 ms)`

jcaraban:
ops, I closed the the MR by error

jcaraban:
I don't think I can continue this MR without further guidance.

I split the tests in two sets, for Vulkan_1_0 and OpenCL_2_0. Some built-ins are shared by both environments (e.g. WorkgroupId), therefore both Vulkan and OpenCL need tests for them.

Still, I don't think I handled the target environments correctly. For example, I didn't see the need to write any specialized logic for Vulkan or OpenCL in validate_builtins.cpp.

I wrote tests that pass, tests that fail for the Execution Model, tests that fail for the Storage Class, test that fail for the Data Type and few individual tests for special conditions. Apart from some confusing TODOs, I don't know if more tests are needed, or perhaps I wrote too many already.

There is also the issue of code repetition in the validation code, which now reaches 2000 lines.

jcaraban:
Question @dneto0, can a single variable be decorated with multiple built-ins?

jcaraban:
About built-ins used by both Kernel and GLCompute models, e.g. _GlobalInvocationId_.

In Kernel model they must use unsigned integers, due to the rules in section 2.16.3.
In GLCompute model, I understand from the Vulkan spec that they should use signed ints.

The Vulkan spec says for e.g. _SampleId_ that ``The variable decorated with SampleId must be declared as a scalar 32-bit integer.`` and in fact **glslc** translates such built-in to ``OpTypeInt 32 1``.

For _GlobalInvocationId_, the Vulkan spec says ``The variable decorated with GlobalInvocationId must be declared as a three-component vector of 32-bit integers.``, so I infer it should use signed ints too.

However, **glslc** now outputs ``OpTypeInt 32 0``, therefore using unsigned integers.
Who's wrong: glslc, the spec or me?

dneto0:
FYI. @atgoo has since added a bunch of validation for Builtin variables for Vulkan shaders.

dj2:
Is this PR still valid or should we close it?

dj2:
Going to close this for now. We can re-open if we find it's still useful.

PR #868, opened 2017-10-11
Reenable validate type unique pass

atgoo:
Vulkan CTS patch fixing the instances of non-unique type declaration in
autogenerated code has recently been submitted.


atgoo:
Passes on Vulkan CTS master.

atgoo:
It seems the Vulkan fix was not transferred to other branches, and only exists in master.

dneto0:
I asked Vulkan CTS lead if it's ok to turn this on now.
This MR is WIP until we get a "yes" answer.

dneto0:
Got a positive response from Vulkan CTS lead.

dneto0:
Rebased and pushed into master as 8de8dd8c8c69

PR #869, opened 2017-10-11
Validator checks OpReturn called from void func

atgoo:
Added check into validate_cfg which checks that OpReturn is not called
from functions which are supposed to return a value.

### DIFF
--- a/source/val/function.h
+++ b/source/val/function.h
@@ -137,7 +137,7 @@ class Function {
   uint32_t id() const { return id_; }
 
   /// Returns return type id of the function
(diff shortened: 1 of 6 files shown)


atgoo:
Vulkan CTS master passes with this change.

dneto0:
I also filed https://github.com/KhronosGroup/SPIRV-Tools/issues/876 for the opposite condition: OpReturnValue result type must match the function's return type.

dneto0:
Waiting for bots to go green

dneto0:
Rebased, squashed, and pushed into master as  b54997e6ebd9

PR #873, opened 2017-10-12
DeadBranchElim: Slightly more defensive coding

dneto0:


dneto0:
This addresses my minor concerns from #863 

Please take a look @greg-lunarg

greg-lunarg:
Looks good.

My take away is that we need to document assumptions based on the spec using asserts.

Among other benefits, it future proofs the code from changes to the spec.

dneto0:
> My take away is that we need to document assumptions based on the spec using asserts.

I don't intend that to be a hard-and-fast rule.  It's a judgement call, based on readability concerns.  Some things are easier to remember than others.  The asserts or comments are an assist to the reader, and carry some redundancy. 

Code review is for correctness *and* readability.  Unfortunately readability is fuzzy and subjective.   So let's keep talking.  :-)

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -118,6 +118,7 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
   ir::Instruction* sInst = def_use_mgr_->GetDef(selId);
   uint32_t typeId = sInst->type_id();
   ir::Instruction* typeInst = def_use_mgr_->GetDef(typeId);
```
atgoo: assert(typeInst);

dneto0: Point taken.  I changed the following test to be

if (!typeInst || (typeInst->opcode() != SpvOpTypeInt)) return false;

dneto0:
Rebased. @atgoo please take another look

dneto0:
Hi @atgoo can you take another look?

dneto0:
Rebased

PR #874, opened 2017-10-12
ADCE: Treat privates like locals in entry point with no calls

greg-lunarg:
This is needed for ongoing legalization of HLSL. It allows removal
of accesses to textures/buffers stored into private variables that are not used.


dneto0:
Will wait for the bots to go green.

dneto0:
Rebased and pushed into master as 1a9061a2be

PR #875, opened 2017-10-12
Opt: Move *NextId functionality into MemPass

greg-lunarg:
Part of continued refactoring.

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -397,7 +397,7 @@ void DeadBranchElimPass::Initialize(ir::Module* module) {
   def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module_));
 
   // Initialize next unused Id.
-  next_id_ = module->id_bound();
+  InitNextId();
 
   // Initialize extension whitelist
   InitExtensions();
@@ -435,7 +435,7 @@ Pass::Status DeadBranchElimPass::ProcessImpl() {
     return EliminateDeadBranches(fp);
   };
   bool modified = ProcessEntryPointCallTree(pfn, module_);
-  FinalizeNextId(module_);
+  FinalizeNextId();
   return modified ? Status::SuccessWithChange : Status::SuccessWithoutChange;
 }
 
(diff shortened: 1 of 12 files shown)


source/opt/dead_branch_elim_pass.cpp
```diff
@@ -397,7 +397,7 @@ void DeadBranchElimPass::Initialize(ir::Module* module) {
   def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module_));
 
   // Initialize next unused Id.
-  next_id_ = module->id_bound();
+  InitNextId();
```
dnovillo: Shouldn't this be moved to the base Pass class?  It doesn't seem useful to me that every pass needs to do this manual step all the time.  I had to do something similar in the CFG cleanup pass (https://github.com/KhronosGroup/SPIRV-Tools/pull/810/)

dneto0: @greg-lunarg and I had talked about moving this to my long-talked-about IRContext utility class.  This is a step in the right direction.   ::shrug:: ?

dnovillo: Oh, in that case, never mind.  I'm going to be adding an IRContext utility class soonish to fix the plethora of TODOs I left in my CFG cleanup pass.  I'll deal with it when I do.  I'm OK with this intermediate step as it is in the right direction.

greg-lunarg: I created MemPass (derived from Pass) to hold shared functionality which I found I needed in the size reduction passes but hadn't yet been contained in the dozen passes that preceded the ones I wrote. I wasn't sure it made sense to push this functionality all the way to Pass. But I am open to this if people feel the benefits outweigh the costs or see other benefits for changes to this architecture.

greg-lunarg: Once we decide where these should end up, I would propose creating functions Initialize(Mem)Pass and Finalize(Mem)Pass which initialize and finalize those members which they contain.

dnovillo: Sure.  It could also be moved into the pass manager, so that individual passes don't have to deal with all this.

dneto0: Ok.  I'll merge this and @dnovillo can do the refactor later.

dneto0:
Waiting for bots to go green.

dneto0:
Rebased and pushed into master as 1e7994c08

Issue #876, opened 2017-10-12
OpReturnValue result type is the same as function return type


dneto0:

dneto0:
I'm claiming this.

atgoo:
Is this not checked by IdPass? It seems to be doing a bit of everything.

dneto0:
Yes, it's already checked.  I added a test in #877

PR #877, opened 2017-10-12
Validator: Test OpReturnValue type check

dneto0:
The check already existed.  I added a test for it.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/876


dneto0:
Rebased. Keeping LGTM

PR #879, opened 2017-10-13
Change BasicBlock to use InstructionList to hold instructions.

s-perron:
This is the first step in replacing the std::vector of Instruction
pointers to using and intrusive linked list.

To this end, we created the InstructionList class.  It inherites from
the IntrusiveList class, but add the extra concept of ownership.  An
InstructionList owns the instruction that are in it.  This is to be
consistent with the current ownership rules where the vector owns the
instruction that are in it.

The other larger change is that the inst_ member of the BasicBlock class
was changed to using the InstructionList class.

### DIFF
--- a/source/opt/basic_block.cpp
+++ b/source/opt/basic_block.cpp
@@ -23,9 +23,8 @@ BasicBlock::BasicBlock(const BasicBlock& bb)
     : function_(nullptr),
       label_(MakeUnique<Instruction>(bb.GetLabelInst())),
       insts_() {
-  insts_.reserve(bb.insts_.size());
   for (auto& inst : bb.insts_)
-    AddInstruction(MakeUnique<Instruction>(*inst.get()));
+    AddInstruction(std::unique_ptr<Instruction>(inst.Clone()));
 }
 
 const Instruction* BasicBlock::GetMergeInst() const {
@@ -78,7 +77,7 @@ Instruction* BasicBlock::GetLoopMergeInst() {
 
 void BasicBlock::ForEachSuccessorLabel(
     const std::function<void(const uint32_t)>& f) {
-  const auto br = &*insts_.back();
+  const auto br = &insts_.back();
   switch (br->opcode()) {
     case SpvOpBranch: {
       f(br->GetOperand(0).words[0]);
@@ -102,9 +101,9 @@ void BasicBlock::ForMergeAndContinueLabel(
   --ii;
   if (ii == insts_.begin()) return;
   --ii;
-  if ((*ii)->opcode() == SpvOpSelectionMerge || 
-      (*ii)->opcode() == SpvOpLoopMerge)
-    (*ii)->ForEachInId([&f](const uint32_t* idp) {
+  if (ii->opcode() == SpvOpSelectionMerge || 
+      ii->opcode() == SpvOpLoopMerge)
+    ii->ForEachInId([&f](const uint32_t* idp) {
       f(*idp);
     });
 }
--- a/source/opt/basic_block.h
+++ b/source/opt/basic_block.h
@@ -126,7 +126,7 @@ inline void BasicBlock::AddInstruction(std::unique_ptr<Instruction> i) {
 
 inline void BasicBlock::AddInstructions(BasicBlock* bp) {
   auto bEnd = end();
-  (void)bEnd.InsertBefore(&bp->insts_);
+  (void) bEnd.MoveBefore(&bp->insts_);
 }
 
 inline void BasicBlock::ForEachInst(const std::function<void(Instruction*)>& f,
--- a/source/opt/basic_block.h
+++ b/source/opt/basic_block.h
@@ -24,6 +24,7 @@
 #include <vector>
 
 #include "instruction.h"
+#include "instruction_list.h"
 #include "iterator.h"
 
 namespace spvtools {
@@ -34,8 +35,8 @@ class Function;
 // A SPIR-V basic block.
 class BasicBlock {
  public:
-  using iterator = UptrVectorIterator<Instruction>;
-  using const_iterator = UptrVectorIterator<Instruction, true>;
+  using iterator = InstructionList::iterator;
+  using const_iterator = InstructionList::const_iterator;
 
   // Creates a basic block with the given starting |label|.
   inline explicit BasicBlock(std::unique_ptr<Instruction> label);
@@ -71,26 +72,22 @@ class BasicBlock {
   // Returns the id of the label at the top of this block
   inline uint32_t id() const { return label_->result_id(); }
 
-  iterator begin() { return iterator(&insts_, insts_.begin()); }
-  iterator end() { return iterator(&insts_, insts_.end()); }
-  const_iterator cbegin() const {
-    return const_iterator(&insts_, insts_.cbegin());
-  }
-  const_iterator cend() const {
-    return const_iterator(&insts_, insts_.cend());
-  }
+  iterator begin() { return insts_.begin(); }
+  iterator end() { return insts_.end(); }
+  const_iterator cbegin() const { return insts_.cbegin(); }
+  const_iterator cend() const { return insts_.cend(); }
 
   // Returns an iterator pointing to the last instruction.  This may only
   // be used if this block has an instruction other than the OpLabel
   // that defines it.
   iterator tail() {
     assert(!insts_.empty());
-    return iterator(&insts_, std::prev(insts_.end()));
+    return --end();
   }
   // Returns a const iterator, but othewrise similar to tail().
   const_iterator ctail() const {
     assert(!insts_.empty());
-    return const_iterator(&insts_, std::prev(insts_.cend()));
+    return --insts_.cend();
   }
 
   // Runs the given function |f| on each instruction in this basic block, and
@@ -106,12 +103,10 @@ class BasicBlock {
                              bool run_on_debug_line_insts = false);
 
   // Runs the given function |f| on each label id of each successor block
-  void ForEachSuccessorLabel(
-      const std::function<void(const uint32_t)>& f);
+  void ForEachSuccessorLabel(const std::function<void(const uint32_t)>& f);
 
   // Runs the given function |f| on the merge and continue label, if any
-  void ForMergeAndContinueLabel(
-      const std::function<void(const uint32_t)>& f);
+  void ForMergeAndContinueLabel(const std::function<void(const uint32_t)>& f);
 
  private:
   // The enclosing function.
@@ -119,25 +114,25 @@ class BasicBlock {
   // The label starting this basic block.
   std::unique_ptr<Instruction> label_;
   // Instructions inside this basic block, but not the OpLabel.
-  std::vector<std::unique_ptr<Instruction>> insts_;
+  InstructionList insts_;
 };
 
 inline BasicBlock::BasicBlock(std::unique_ptr<Instruction> label)
     : function_(nullptr), label_(std::move(label)) {}
 
 inline void BasicBlock::AddInstruction(std::unique_ptr<Instruction> i) {
-  insts_.emplace_back(std::move(i));
+  insts_.push_back(i.release());
 }
 
 inline void BasicBlock::AddInstructions(BasicBlock* bp) {
   auto bEnd = end();
-  (void) bEnd.InsertBefore(&bp->insts_);
+  (void)bEnd.InsertBefore(&bp->insts_);
 }
 
 inline void BasicBlock::ForEachInst(const std::function<void(Instruction*)>& f,
                                     bool run_on_debug_line_insts) {
   if (label_) label_->ForEachInst(f, run_on_debug_line_insts);
-  for (auto& inst : insts_) inst->ForEachInst(f, run_on_debug_line_insts);
+  for (auto& inst : insts_) inst.ForEachInst(f, run_on_debug_line_insts);
 }
 
 inline void BasicBlock::ForEachInst(
@@ -147,15 +142,15 @@ inline void BasicBlock::ForEachInst(
     static_cast<const Instruction*>(label_.get())
         ->ForEachInst(f, run_on_debug_line_insts);
   for (const auto& inst : insts_)
-    static_cast<const Instruction*>(inst.get())
+    static_cast<const Instruction*>(&inst)
         ->ForEachInst(f, run_on_debug_line_insts);
 }
 
 inline void BasicBlock::ForEachPhiInst(
     const std::function<void(Instruction*)>& f, bool run_on_debug_line_insts) {
   for (auto& inst : insts_) {
-    if (inst->opcode() != SpvOpPhi) break;
-    inst->ForEachInst(f, run_on_debug_line_insts);
+    if (inst.opcode() != SpvOpPhi) break;
+    inst.ForEachInst(f, run_on_debug_line_insts);
   }
 }
 
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -126,7 +126,7 @@ uint32_t InlinePass::GetFalseId() {
 
 void InlinePass::MapParams(
     ir::Function* calleeFn,
-    ir::UptrVectorIterator<ir::Instruction> call_inst_itr,
+    ir::BasicBlock::iterator call_inst_itr,
     std::unordered_map<uint32_t, uint32_t>* callee2caller) {
   int param_idx = 0;
   calleeFn->ForEachParam(
@@ -145,8 +145,7 @@ void InlinePass::CloneAndMapLocals(
   auto callee_block_itr = calleeFn->begin();
   auto callee_var_itr = callee_block_itr->begin();
   while (callee_var_itr->opcode() == SpvOp::SpvOpVariable) {
-    std::unique_ptr<ir::Instruction> var_inst(
-        new ir::Instruction(*callee_var_itr));
+    std::unique_ptr<ir::Instruction> var_inst(callee_var_itr->Clone());
     uint32_t newId = TakeNextId();
     var_inst->SetResultId(newId);
     (*callee2caller)[callee_var_itr->result_id()] = newId;
@@ -196,8 +195,7 @@ void InlinePass::CloneSameBlockOps(
           if (mapItr2 != (*preCallSB).end()) {
             // Clone pre-call same-block ops, map result id.
             const ir::Instruction* inInst = mapItr2->second;
-            std::unique_ptr<ir::Instruction> sb_inst(
-                new ir::Instruction(*inInst));
+            std::unique_ptr<ir::Instruction> sb_inst(inInst->Clone());
             CloneSameBlockOps(&sb_inst, postCallSB, preCallSB, block_ptr);
             const uint32_t rid = sb_inst->result_id();
             const uint32_t nid = this->TakeNextId();
@@ -216,7 +214,7 @@ void InlinePass::CloneSameBlockOps(
 void InlinePass::GenInlineCode(
     std::vector<std::unique_ptr<ir::BasicBlock>>* new_blocks,
     std::vector<std::unique_ptr<ir::Instruction>>* new_vars,
-    ir::UptrVectorIterator<ir::Instruction> call_inst_itr,
+    ir::BasicBlock::iterator call_inst_itr,
     ir::UptrVectorIterator<ir::BasicBlock> call_block_itr) {
   // Map from all ids in the callee to their equivalent id in the caller
   // as callee instructions are copied into caller.
@@ -328,7 +326,7 @@ void InlinePass::GenInlineCode(
           // Copy contents of original caller block up to call instruction.
           for (auto cii = call_block_itr->begin(); cii != call_inst_itr;
                ++cii) {
-            std::unique_ptr<ir::Instruction> cp_inst(new ir::Instruction(*cii));
+            std::unique_ptr<ir::Instruction> cp_inst(cii->Clone());
             // Remember same-block ops for possible regeneration.
             if (IsSameBlockOp(&*cp_inst)) {
               auto* sb_inst_ptr = cp_inst.get();
@@ -438,7 +436,7 @@ void InlinePass::GenInlineCode(
         // Copy remaining instructions from caller block.
         auto cii = call_inst_itr;
         for (++cii; cii != call_block_itr->end(); ++cii) {
-          std::unique_ptr<ir::Instruction> cp_inst(new ir::Instruction(*cii));
+          std::unique_ptr<ir::Instruction> cp_inst(cii->Clone());
           // If multiple blocks generated, regenerate any same-block
           // instruction that has not been seen in this last block.
           if (multiBlocks) {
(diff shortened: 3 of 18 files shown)


dneto0:
Bots are unhappy about unused parameters, at least
https://travis-ci.org/KhronosGroup/SPIRV-Tools/jobs/287679913#L589

s-perron:
I think this is ready for the first pass of reviews.  I finally got the windows build to work.  I still have a few things to do:

- [ ] Add test for InsertBefore functions.
- [ ] Add test that the destructors for the ownership properties of the InstructionList are correct.
- [ ] Remove some of the more expensive asserts (i.e. calls to "Check").

source/opt/inline_pass.cpp
```diff
@@ -438,7 +436,7 @@ void InlinePass::GenInlineCode(
         // Copy remaining instructions from caller block.
         auto cii = call_inst_itr;
         for (++cii; cii != call_block_itr->end(); ++cii) {
-          std::unique_ptr<ir::Instruction> cp_inst(new ir::Instruction(*cii));
+          std::unique_ptr<ir::Instruction> cp_inst(cii->Clone());
```
dnovillo: Marking the Instruction copy constructor with =delete() wreaks havoc atm, right?

source/opt/instruction.h
```diff
@@ -97,12 +107,21 @@ class Instruction {
...
   Instruction(const Instruction&) = default;
   Instruction& operator=(const Instruction&) = default;
 
   Instruction(Instruction&&);
   Instruction& operator=(Instruction&&);
 
+  // Return a newly allocated instruction that has the same operands, result,
+  // and type as |this|.  The new instruction is not linked into any list.
+  // It is the responsibility of the caller to make sure that the storage is
+  // remove. It is the caller's responsibility to make sure that there is only
```
dnovillo: "the storage is remove"?

source/opt/instruction.h
```diff
@@ -217,6 +241,8 @@ class Instruction {
   // Instructions representing OpLine or OpNonLine itself, this field should be
   // empty.
   std::vector<Instruction> dbg_line_insts_;
```
dnovillo: Shouldn't these also be an InstructionList?  Though I guess that can be changed later.

source/opt/instruction_list.cpp
```diff
@@ -0,0 +1,45 @@
...
+InstructionList::~InstructionList() {
+  while (!empty()) {
+    Instruction* inst = &front();
+    inst->RemoveFromList();
+    delete inst;
+  }
+}
+
+InstructionList::iterator InstructionList::iterator::InsertBefore(
+    std::vector<std::unique_ptr<Instruction>>* list) {
```
dnovillo: Would it make sense to have an ::InsertBefore() that accepted an InstructionList?

source/opt/instruction_list.h
```diff
@@ -0,0 +1,74 @@
...
+    iterator(const utils::IntrusiveList<Instruction>::iterator& i)
+        : utils::IntrusiveList<Instruction>::iterator(&*i) {}
+    iterator(Instruction* i) : utils::IntrusiveList<Instruction>::iterator(i) {}
+
+    // The nodes in |list| will be moved to the list that |this| points to.  The
+    // positions of the nodes will be immediately before the element pointed to
+    // by the iterator.  The return value will be an iterator pointing to the
+    // first of the newly inserted elements.  Ownership if the elements in
+    // |list| is now pass on to the new list that contains them.
+    iterator InsertBefore(std::vector<std::unique_ptr<Instruction>>* list);
```
dnovillo: s/Ownership if/Ownership of/
s/is now pass/is now passed/

source/opt/instruction_list.h
```diff
@@ -0,0 +1,74 @@
...
+
+    // The node |i| will be inserted immediatly before |this|. The return value
```
dnovillo: s/immediatly/immediately/

source/util/ilist_node.h
```diff
@@ -81,6 +99,51 @@ template <class NodeType>
...
+  next_node_ = nullptr;
+  previous_node_ = nullptr;
+  is_sentinel_ = false;
+}
+
+template<class NodeType>
+inline IntrusiveNodeBase<NodeType>& IntrusiveNodeBase<NodeType>::operator=(
+    const IntrusiveNodeBase&) {
+  assert(!is_sentinel_);
+  if (IsInAList()) RemoveFromList();
```
dnovillo: The preference is to always use braces for control flow bodies, even if they're only one line.  I don't think the codebase is very consistent about that, but @dneto0 has requested changes on that before.

source/opt/basic_block.h
```diff
@@ -106,38 +103,36 @@ class BasicBlock {
...
+  void ForMergeAndContinueLabel(const std::function<void(const uint32_t)>& f);
 
  private:
   // The enclosing function.
   Function* function_;
   // The label starting this basic block.
   std::unique_ptr<Instruction> label_;
   // Instructions inside this basic block, but not the OpLabel.
-  std::vector<std::unique_ptr<Instruction>> insts_;
+  InstructionList insts_;
```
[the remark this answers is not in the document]
s-perron: Yes, I have thought of this.  However, I want to delay that decision until we have a better understanding of the trade-offs.  The task we want to accomplish is "Given an Instruction, find the basic block it is contained in."  If an instruction contains a pointer to the list it is contained in, then merging basic block in say code straightening, will be linear time: we have to traverse every instruction to update the pointer to the containing list.  However, it could make finding the basic block really fast.

To find the basic block, we can also make it such that we have to walk the basic block until we get to the label.  So finding the containing basic block is linear time, but merging two block is constant, just change a few pointers.

We'll have to decide what is more important.

source/opt/basic_block.h
```diff
@@ -106,38 +103,36 @@ class BasicBlock {
...
   auto bEnd = end();
-  (void) bEnd.InsertBefore(&bp->insts_);
+  (void) bEnd.MoveBefore(&bp->insts_);
 }
 
 inline void BasicBlock::ForEachInst(const std::function<void(Instruction*)>& f,
                                     bool run_on_debug_line_insts) {
   if (label_) label_->ForEachInst(f, run_on_debug_line_insts);
-  for (auto& inst : insts_) inst->ForEachInst(f, run_on_debug_line_insts);
+  for (auto& inst : insts_) inst.ForEachInst(f, run_on_debug_line_insts);
```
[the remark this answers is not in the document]
s-perron: There is no "ForEachInst" implemented in the InstructionList class.  In this loop, we are calling ForEachInst on an instruction, not the instruction list.  It does this because the Instruction class contains a vector of instruction for the debug information.  We want to call it on the Instruction itself, and associated debug instructions.

source/opt/inline_pass.cpp
s-perron: Yes, it does.  I will eventually do that.  I think I need to change the container for the debug instruction inside the Instruction class, and we should be good.

source/opt/instruction.h
s-perron: done

source/opt/instruction.h
s-perron: Yes, they will be.  I wanted to keep the changes smaller.

source/opt/instruction_list.cpp
s-perron: There is one that is part of the base class.  I wrote this one because it existed for the old iterator type.  As we change the code that uses it we can phase it out.  I will add a comment to say it is deprecated, and new code should not use it.

source/opt/instruction_list.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include "spirv-tools/libspirv.h"
+#include "spirv/1.2/spirv.h"
+
+namespace spvtools {
+namespace ir {
+
+// This class is intended to be the container for Instructions.  This container
+// owns the instructions that are in it.  When removing an Instruction from the
+// list, the caller is assumming responsibility for deleting the storage.
```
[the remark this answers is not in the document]
s-perron: done

source/opt/instruction_list.h
s-perron: done

source/opt/instruction_list.h
s-perron: done

source/util/ilist_node.h
s-perron: done

source/opt/basic_block.cpp
```diff
@@ -23,9 +23,8 @@ BasicBlock::BasicBlock(const BasicBlock& bb)
     : function_(nullptr),
       label_(MakeUnique<Instruction>(bb.GetLabelInst())),
       insts_() {
-  insts_.reserve(bb.insts_.size());
   for (auto& inst : bb.insts_)
-    AddInstruction(MakeUnique<Instruction>(*inst.get()));
+    AddInstruction(std::unique_ptr<Instruction>(inst.Clone()));
```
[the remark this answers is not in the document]
s-perron: I made the change to avoid using the copy constructor that MakeUnique uses.  The new code works with all of the compiler that are use in our continuous builds.

source/opt/basic_block.h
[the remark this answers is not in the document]
dnovillo: I agree.  Let's leave the decision for when we really need it.  Perhaps we could add a helper that does the instruction->block mapping on-demand.  We can factor out the code in #810 for that.

source/opt/basic_block.h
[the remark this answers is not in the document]
dnovillo: Ah, right.

source/opt/instruction.h
dnovillo: SGTM

test/opt/instruction_list_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+// Need to make sure the elements are inserted in the correct order, and the
+// return value points to the correct location.
+TEST(InstructionListTest, InsertBefore1) {
+  InstructionList list;
+  std::vector<Instruction*> created_instructions;
+  for (int i = 0; i < 4; i++) {
+    std::unique_ptr<Instruction> inst(new Instruction());
+    created_instructions.push_back(inst.get());
+    auto new_element = list.end().InsertBefore(std::move(inst));
+    EXPECT_THAT(&*new_element, created_instructions.back());
```
atgoo: EXPECT_EQ?

source/opt/instruction_list.h
```diff
@@ -0,0 +1,92 @@
...
+
+namespace spvtools {
+namespace ir {
+
+// This class is intended to be the container for Instructions.  This container
+// owns the instructions that are in it.  When removing an Instruction from the
+// list, the caller is assuming responsibility for deleting the storage.
+//
+// TODO: Because there are a number of other data structures that will want
+// pointers to instruction, owndership should probably be moved to the module.
```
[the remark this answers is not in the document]
s-perron: done

source/opt/instruction_list.h
```diff
@@ -0,0 +1,92 @@
...
+    // positions of the nodes will be immediately before the element pointed to
+    // by the iterator.  The return value will be an iterator pointing to the
+    // first of the newly inserted elements.  Ownership of the elements in
+    // |list| is now passed on to |*this|.
+    iterator InsertBefore(std::vector<std::unique_ptr<Instruction>>* list);
+
+    // The node |i| will be inserted immediately before |this|. The return value
+    // will be an iterator pointing to the newly inserted node.  The owner of
+    // |*i| becomes |*this|
+    iterator InsertBefore(std::unique_ptr<Instruction> i);
```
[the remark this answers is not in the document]
s-perron: Right.  I fixed it.

test/opt/instruction_list_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  std::vector<Instruction*> output;
+  for (auto& i : list) {
+    output.push_back(&i);
+  }
+  EXPECT_THAT(output, ContainerEq(created_instructions));
```
atgoo: This seems awkwardly limiting as you can have output only in the order they were created.
Perhaps comparing data_ and not pointers would be better, as you can then specify in which order the elements should come.

test/opt/instruction_list_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+    created_instructions_.push_back(this);
+    static int counter = 0;
+    data = ++counter;
+  }
+
+  ~TestInstruction() { deleted_instructions_.push_back(this); }
+
+  int data;
+  static std::vector<TestInstruction*> created_instructions_;
+  static std::vector<TestInstruction*> deleted_instructions_;
```
atgoo: Consider identifying instructions by data_ instead of pointers.

source/util/ilist_node.h
```diff
@@ -120,15 +185,46 @@ inline void IntrusiveNodeBase<NodeType>::InsertAfter(NodeType* pos) {
...
 
   this->next_node_->previous_node_ = this->previous_node_;
   this->previous_node_->next_node_ = this->next_node_;
   this->next_node_ = nullptr;
   this->previous_node_ = nullptr;
 }
 
+template<class NodeType>
+void IntrusiveNodeBase<NodeType>::ReplaceWith(NodeType* target) {
+  if (is_sentinel_) {
```
atgoo: I don't understand the logic in this if/else. Please check if you are not confusing 'this' with 'target'.

source/util/ilist_node.h
```diff
@@ -120,15 +185,46 @@ inline void IntrusiveNodeBase<NodeType>::InsertAfter(NodeType* pos) {
...
+    target->previous_node_->next_node_ = static_cast<NodeType*>(target);
+
+    if (!this->is_sentinel_) {
+      this->next_node_ = nullptr;
+      this->previous_node_ = nullptr;
+    } else {
+      this->next_node_ = static_cast<NodeType*>(this);
+      this->previous_node_ = static_cast<NodeType*>(this);
+    }
+  } else {
```
atgoo: Please add comments.

source/opt/instruction_list.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+
+InstructionList::iterator InstructionList::iterator::InsertBefore(
+    std::unique_ptr<Instruction>&& i) {
+  i.get()->InsertBefore(node_);
+  return iterator(i.release());
+}
+
+}  // namespace ir
+}  // namespace spvtools
```
atgoo: trailing whitespace

source/util/ilist_node.h
```diff
@@ -31,6 +31,18 @@ class IntrusiveNodeBase {
  public:
   // Creates a new node that is not in a list.
   inline IntrusiveNodeBase();
+  inline IntrusiveNodeBase(const IntrusiveNodeBase&);
+  inline IntrusiveNodeBase& operator=(const IntrusiveNodeBase&);
+  inline IntrusiveNodeBase(IntrusiveNodeBase&& that);
+
+  // Will destroy a node.  It is an error to destroy a node that is part of a
+  // list, unless it is an sentinel.
```
atgoo: a sentinel

source/util/ilist_node.h
```diff
@@ -59,10 +71,16 @@ class IntrusiveNodeBase {
...
 
   // Removes the given node from the list.  It is assumed that the node is
-  // in a list.  Note that this does not free any storage related to the node.
+  // in a list.  Note that this does not free any storage related to the node,
+  // it becomes the caller's responsibility to free the storage.
   inline void RemoveFromList();
 
- private:
+ protected:
+  // This function will replace |this| with |target|.  No nodes that are not
```
atgoo: Replaces |this|...

source/util/ilist.h
```diff
@@ -133,13 +157,27 @@ class IntrusiveList {
   // will be removed from that list first.
   void push_back(NodeType* node);
 
- private:
+  // Return true if the list is empty.
```
atgoo: Returns

source/util/ilist.h
```diff
@@ -98,7 +100,29 @@ class IntrusiveList {
       return !(lhs == rhs);
     }
 
-   private:
+    // The nodes in |list| will be moved to the list that |this| points to.  The
```
atgoo: Moves the nodes

source/opt/instruction_list.h
```diff
@@ -0,0 +1,92 @@
...
+
+  class iterator : public utils::IntrusiveList<Instruction>::iterator {
+   public:
+    iterator(const utils::IntrusiveList<Instruction>::iterator& i)
+        : utils::IntrusiveList<Instruction>::iterator(&*i) {}
+    iterator(Instruction* i) : utils::IntrusiveList<Instruction>::iterator(i) {}
+
+    // DEPRECATED: Please use MoveBefore with an InstructionList instead.
+    //
+    // The nodes in |list| will be moved to the list that |this| points to.  The
```
atgoo: Moves the nodes

source/opt/instruction.h
```diff
@@ -97,12 +107,23 @@ class Instruction {
...
+  // std::vector<Instruction>.
   Instruction(const Instruction&) = default;
   Instruction& operator=(const Instruction&) = default;
 
   Instruction(Instruction&&);
   Instruction& operator=(Instruction&&);
 
+  virtual ~Instruction() = default;
+
+  // Return a newly allocated instruction that has the same operands, result,
```
atgoo: Returns

source/util/ilist_node.h
```diff
@@ -31,6 +31,18 @@ class IntrusiveNodeBase {
  public:
   // Creates a new node that is not in a list.
   inline IntrusiveNodeBase();
+  inline IntrusiveNodeBase(const IntrusiveNodeBase&);
+  inline IntrusiveNodeBase& operator=(const IntrusiveNodeBase&);
+  inline IntrusiveNodeBase(IntrusiveNodeBase&& that);
+
+  // Will destroy a node.  It is an error to destroy a node that is part of a
```
atgoo: Destroys

source/util/ilist_node.h
s-perron: Yes, there were some problems there.  I fixed it up, please take another look when I do a push.

source/util/ilist_node.h
s-perron: Fixed up a bit.  Let me know what you think.

test/opt/instruction_list_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+ public:
+  TestInstruction() : Instruction() {
+    created_instructions_.push_back(this);
+    static int counter = 0;
+    data = ++counter;
+  }
+
+  ~TestInstruction() { deleted_instructions_.push_back(this); }
+
+  int data;
```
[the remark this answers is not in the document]
s-perron: I just removed it because it is not used.  This class was just used to trace the destructor.

test/opt/instruction_list_test.cpp
s-perron: I prefer the pointers because I need to know the same memory locations are destroyed, not something that happens to have the same data.

test/opt/instruction_list_test.cpp
s-perron: I renamed created_instructions to make it clearer.  The important part is that the instruction appear in the list in the same order as they are inserted.  Using "data_" is ackward because that field does not exist in an instruction.  If I use TestInstructions, there is lots of extra casting that is needed to access "data_".

The other consideration is that I want to make sure the address remain stable while moving things around.  We will have other data structures like the def-use manager that will have pointers to instructions.  It is important that we have stable address to help avoid dangling pointers.  I'll add a comment about that.

test/opt/instruction_list_test.cpp
s-perron: Fixed.

source/opt/instruction.cpp
```diff
@@ -68,6 +70,16 @@ Instruction& Instruction::operator=(Instruction&& that) {
   return *this;
 }
 
+Instruction* Instruction::Clone() const {
+  Instruction * result = new Instruction();
```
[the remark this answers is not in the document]
s-perron: done

source/opt/instruction.h
s-perron: done

source/opt/instruction_list.cpp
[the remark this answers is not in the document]
s-perron: done

source/opt/instruction_list.cpp
s-perron: done

source/util/ilist.h
s-perron: done

source/util/ilist.h
s-perron: Done

source/util/ilist_node.h
s-perron: Done

source/util/ilist_node.h
s-perron: Done

source/util/ilist_node.h
s-perron: Done

source/util/ilist_node.h
```diff
@@ -76,11 +76,14 @@ class IntrusiveNodeBase {
   inline void RemoveFromList();
 
  protected:
-  // This function will replace |this| with |target|.  No nodes that are not
+  // Replaces |this| with |target|.  No nodes that are not
```
atgoo: Replaces |this| with |target|. |this| and |target| need to be either
- both sentinels, in which case the entire list is moved
*or*
- both not sentinels, in which case only single node is moved
No storage is deleted.

source/util/ilist_node.h
```diff
@@ -196,33 +199,50 @@ inline void IntrusiveNodeBase<NodeType>::RemoveFromList() {
...
+      // a sentinel node to not be in a list.
       this->next_node_ = static_cast<NodeType*>(this);
       this->previous_node_ = static_cast<NodeType*>(this);
     }
   } else {
-    target->next_node_ = static_cast<NodeType*>(target);
-    target->previous_node_ = static_cast<NodeType*>(target);
+    // If |this| points to itself, it must be a sentinel node with an empty
+    // list.  Reset |this| so that it is the head of an empty list.  We want
+    // |target| to be the same.  The asserts above guarantee that.
```
atgoo: > Reset |this| so that it is the head of an empty list.

You are not actually resetting anything.

source/util/ilist_node.h
```diff
@@ -196,33 +199,50 @@ inline void IntrusiveNodeBase<NodeType>::RemoveFromList() {
 
 template<class NodeType>
 void IntrusiveNodeBase<NodeType>::ReplaceWith(NodeType* target) {
-  if (is_sentinel_) {
-    assert(target->next_node_ == target);
-    assert(this->is_sentinel_);
+  if (this->is_sentinel_) {
+    target->IsEmptyList();
```
atgoo: ` target->IsEmptyList();`

assert(target->is_sentinel_)?

source/util/ilist_node.h
s-perron: done

source/util/ilist_node.h
s-perron: Should be assert(target->IsEmptyList()), and that checks target->is_sentinel_.  I fixed that up.

dneto0:
Ok.  Needs a rebase since there are now conflicts.  (It's fairly trivial, and you'll keep your LGTM)

dneto0:
oh wait, I can do the resolve.  I'm brave.

dneto0:
No.  I'm not that brave.   

@s-perron Please rebase this.

s-perron:
Rebased, squashed, and merged as bb7802b.

PR #880, opened 2017-10-13
Validate SpvOpVectorShuffle

jozefkucia:


source/validate_id.cpp
```diff
@@ -1701,12 +1701,41 @@ bool idUsage::isValid<OpVectorInsertDynamic>(
...
+  // Component operands.
+  auto componentCount = inst->words.size() - 5;
+  auto resultVectorDimension = resultType->words()[3];
+  if (componentCount != resultVectorDimension) {
+    DIAG(inst->words.size() - 1)
+        << instr_name() << " component literals count does not match "
+                           "Result Type <id> '"
+        << resultType->id() << "'s vector component count.";
+    return false;
+  }
```
dneto0: This is good as far as it goes.
But there are more rules to check:
- each literal is either 0xFFFFFFFF,
   or as an unsigned integer at most  the sum of the lengths of the input vectors, minus 1

jozefkucia: Thanks for the review. I've added checks for the remaining rules.

dneto0:
Waiting for the bots to pass

dneto0:
One bot failed spuriously.  I kicked it off again.

dneto0:
Rebased, squashed, and pushed into master as 90862fe4b1c6

Issue #750, opened 2017-08-04
Opt: --eliminate-common-uniform does not honor volatile


greg-lunarg:
Need to fix --eliminate-common-uniform so that it does not eliminate loads of volatile uniform buffer blocks and volatile images

dneto0:
Fixed by #881

PR #881, opened 2017-10-14
Fixed --eliminate-common-uniform so that it does not eliminate loads of volatile variables

daniel-schuermann:
Issue #750

### DIFF
--- a/source/opt/common_uniform_elim_pass.cpp
+++ b/source/opt/common_uniform_elim_pass.cpp
@@ -15,9 +15,7 @@
 // limitations under the License.
 
 #include "common_uniform_elim_pass.h"
-
 #include "cfa.h"
-#include "iterator.h"
 
 namespace spvtools {
 namespace opt {
@@ -128,6 +126,68 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
     objInst = def_use_mgr_->GetDef(*objId);
   }
   return ptrInst;
+}
+
+bool CommonUniformElimPass::IsVolatileStruct(uint32_t type_id) {
+  assert(def_use_mgr_->GetDef(type_id)->opcode() == SpvOpTypeStruct);
+  bool has_volatile_deco = false;
+  dec_mgr_->ForEachDecoration(type_id, SpvDecorationVolatile,
+                              [&has_volatile_deco](const ir::Instruction&){ has_volatile_deco = true;});
+  return has_volatile_deco;
+}
+
+bool CommonUniformElimPass::IsAccessChainToVolatileStructType(const ir::Instruction &AccessChainInst) {
+  assert(AccessChainInst.opcode() == SpvOpAccessChain);
+
+  uint32_t ptr_id = AccessChainInst.GetSingleWordInOperand(0);
+  const ir::Instruction* ptr_inst = def_use_mgr_->GetDef(ptr_id);
+  uint32_t pointee_type_id = GetPointeeTypeId(ptr_inst);
+  const uint32_t num_operands = AccessChainInst.NumOperands();
+
+  // walk the type tree:
+  for (uint32_t idx = 3; idx < num_operands; ++idx) {
+    ir::Instruction* pointee_type = def_use_mgr_->GetDef(pointee_type_id);
+
+    switch (pointee_type->opcode()) {
+      case SpvOpTypeMatrix:
+      case SpvOpTypeVector:
+      case SpvOpTypeArray:
+      case SpvOpTypeRuntimeArray:
+        pointee_type_id = pointee_type->GetSingleWordOperand(1);
+        break;
+      case SpvOpTypeStruct:
+        // check for volatile decorations:
+        if (IsVolatileStruct(pointee_type_id)) return true;
+
+        if (idx < num_operands - 1) {
+          const uint32_t index_id = AccessChainInst.GetSingleWordOperand(idx);
+          const ir::Instruction* index_inst = def_use_mgr_->GetDef(index_id);
+          uint32_t index_value = index_inst->GetSingleWordOperand(2); // TODO: replace with GetUintValueFromConstant()
+          pointee_type_id = pointee_type->GetSingleWordInOperand(index_value);
+        }
+        break;
+      default:
+        assert(false && "Unhandled pointee type.");
+    }
+  }
+  return false;
+}
+
+bool CommonUniformElimPass::IsVolatileLoad(const ir::Instruction& loadInst) {
+  assert(loadInst.opcode() == SpvOpLoad);
+  // Check if this Load instruction has Volatile Memory Access flag
+  if (loadInst.NumOperands() == 4) {
+    uint32_t memory_access_mask = loadInst.GetSingleWordOperand(3);
+    if (memory_access_mask & SpvMemoryAccessVolatileMask)
+      return true;
+  }
+  // If we load a struct directly (result type is struct),
+  // check if the struct is decorated volatile
+  uint32_t type_id = loadInst.type_id();
+  if (def_use_mgr_->GetDef(type_id)->opcode() == SpvOpTypeStruct)
+    return IsVolatileStruct(type_id);
+  else
+    return false;
 }
 
 bool CommonUniformElimPass::IsUniformVar(uint32_t varId) {
@@ -301,6 +361,10 @@ bool CommonUniformElimPass::UniformAccessChainConvert(ir::Function* func) {
         continue;
       if (HasUnsupportedDecorates(ptrInst->result_id()))
         continue;
+      if (IsVolatileLoad(*ii))
+        continue;
+      if (IsAccessChainToVolatileStructType(*ptrInst))
+        continue;
       std::vector<std::unique_ptr<ir::Instruction>> newInsts;
       uint32_t replId;
       GenACLoadRepl(ptrInst, &newInsts, &replId);
@@ -386,6 +450,8 @@ bool CommonUniformElimPass::CommonUniformLoadElimination(ir::Function* func) {
         continue;
       if (HasUnsupportedDecorates(ii->result_id()))
         continue;
+      if (IsVolatileLoad(*ii))
+        continue;
       uint32_t replId;
       const auto uItr = uniform2load_id_.find(varId);
       if (uItr != uniform2load_id_.end()) {
@@ -437,6 +503,8 @@ bool CommonUniformElimPass::CommonUniformLoadElimBlock(ir::Function* func) {
       if (!IsSamplerOrImageVar(varId))
         continue;
       if (HasUnsupportedDecorates(ii->result_id()))
+        continue;
+      if (IsVolatileLoad(*ii))
         continue;
       uint32_t replId;
       const auto uItr = uniform2load_id_.find(varId);
@@ -527,6 +595,7 @@ void CommonUniformElimPass::Initialize(ir::Module* module) {
 
   // TODO(greg-lunarg): Use def/use from previous pass
   def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module_));
+  dec_mgr_.reset(new analysis::DecorationManager(module_));
 
   // Initialize next unused Id.
   next_id_ = module->id_bound();
--- a/source/opt/common_uniform_elim_pass.h
+++ b/source/opt/common_uniform_elim_pass.h
@@ -25,6 +25,7 @@
 #include <utility>
 
 #include "def_use_manager.h"
+#include "decoration_manager.h"
 #include "module.h"
 #include "basic_block.h"
 #include "pass.h"
@@ -65,6 +66,20 @@ class CommonUniformElimPass : public Pass {
 
   // Return true if variable is uniform
   bool IsUniformVar(uint32_t varId);
+
+  // Given the type id for a struct type, checks if the struct type
+  // or any struct member is volatile decorated
+  bool IsVolatileStruct(uint32_t type_id);
+
+  // Given an OpAccessChain instruction, return true
+  // if the accessed variable belongs to a volatile
+  // decorated object or member of a struct type
+  bool IsAccessChainToVolatileStructType(const ir::Instruction &AccessChainInst);
+
+  // Given an OpLoad instruction, return true if
+  // OpLoad has a Volatile Memory Access flag or if
+  // the resulting type is a volatile decorated struct
+  bool IsVolatileLoad(const ir::Instruction& loadInst);
 
   // Return true if any uses of |id| are decorate ops.
   bool HasUnsupportedDecorates(uint32_t id) const;
@@ -178,6 +193,9 @@ class CommonUniformElimPass : public Pass {
 
   // Def-Uses for the module we are processing
   std::unique_ptr<analysis::DefUseManager> def_use_mgr_;
+
+  // Decorations for the module we are processing
+  std::unique_ptr<analysis::DecorationManager> dec_mgr_;
 
   // Map from block's label id to block.
   std::unordered_map<uint32_t, ir::BasicBlock*> id2block_;
--- a/test/opt/common_uniform_elim_test.cpp
+++ b/test/opt/common_uniform_elim_test.cpp
@@ -14,7 +14,6 @@
 // limitations under the License.
 
 #include "pass_fixture.h"
-#include "pass_utils.h"
 
 namespace {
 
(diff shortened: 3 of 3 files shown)


dneto0:
FYI @greg-lunarg

dneto0:
Hi @daniel-schuermann.  You need to (digitally) sign the Contributor's License Agreement (CLA) before we can accept your contribution.

daniel-schuermann:
Of course! Sorry, I overlooked the first comment.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -130,6 +130,32 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
...
+        break;
+      default: continue;
+    }
+    if (decoration == SpvDecorationVolatile)
+      return true;
+  }
+  return false;
+}
+
+bool CommonUniformElimPass::IsVolatileLoad(ir::Instruction *loadInst) {
```
s-perron: If loadInst is not changed the type should be a "const ir::Instruction&".

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -301,6 +327,10 @@ bool CommonUniformElimPass::UniformAccessChainConvert(ir::Function* func) {
         continue;
       if (HasUnsupportedDecorates(ptrInst->result_id()))
         continue;
+      if (IsVolatileVar(varId))
+        continue;
+      if (IsVolatileLoad(&*ii))
```
s-perron: Since "IsVolatileLoad" is a less expensive check, it might be useful to move it up.  No point walking all of the decorations if this check will fail anyway.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -386,6 +416,10 @@ bool CommonUniformElimPass::CommonUniformLoadElimination(ir::Function* func) {
         continue;
       if (HasUnsupportedDecorates(ii->result_id()))
         continue;
+      if (IsVolatileVar(varId))
+        continue;
+      if (IsVolatileLoad(&*ii))
```
s-perron: Same as above.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -438,6 +472,10 @@ bool CommonUniformElimPass::CommonUniformLoadElimBlock(ir::Function* func) {
...
+        continue;
```
s-perron: Same as above.

source/opt/common_uniform_elim_pass.h
```diff
@@ -66,6 +67,12 @@ class CommonUniformElimPass : public Pass {
   // Return true if variable is uniform
   bool IsUniformVar(uint32_t varId);
 
+  // Return true if variable has Volatile decoration
+  bool IsVolatileVar(uint32_t varId);
+
+  // Return true if Load has Volatile Memory Access flag
```
s-perron: Should say that loadInst is assumed to be an OpLoad instruction.

test/opt/common_uniform_elim_test.cpp
```diff
@@ -664,6 +664,307 @@ OpFunctionEnd
       predefs + before, predefs + after, true, true);
 }
 
+  TEST_F(CommonUniformElimTest, Volatile1) {
```
s-perron: Is this test any different that "Basic1"?  With no volatile values, what are you checking?

test/opt/common_uniform_elim_test.cpp
```diff
@@ -664,6 +664,307 @@ OpFunctionEnd
...
+  SinglePassRunAndCheck<opt::CommonUniformElimPass>(
+  predefs + before, predefs + after, true, true);
+}
+
+TEST_F(CommonUniformElimTest, Volatile2) {
+  // Note: This test exemplifies the following:
+  // - Common uniform (%_) load floated to nearest non-controlled block
+  // - Common extract (g_F) floated to non-controlled block
+  // - Non-common extract (g_F2) not floated, but common uniform load shared
+  //
```
s-perron: Are the above comments still accurate now that  U_t is changed to volatile?

test/opt/common_uniform_elim_test.cpp
```diff
@@ -664,6 +664,307 @@ OpFunctionEnd
...
+  // } ;
+  //
+  // void main()
+  // {
+  //     vec4 v = BaseColor;
+  //     if (fi > 0) {
+  //       v = v * g_F;
+  //     }
+  //     else {
+  //       float f2 = g_F2 - g_F;
```
s-perron: Since we know that the load of g_F2 won't be moved because it is not loaded on any other path, is it still needed for this test?

test/opt/common_uniform_elim_test.cpp
```diff
@@ -664,6 +664,307 @@ OpFunctionEnd
...
+OpStore %v %44
+OpBranch %30
+%30 = OpLabel
+%45 = OpLoad %v4float %v
+OpStore %gl_FragColor %45
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::CommonUniformElimPass>(
```
s-perron: It looks like you are testing that nothing happens.  You could use SinglePassRunAndDisassemble and check that the result was a SuccessWithoutChange.

At the very least, get rid of "after" and just use "before".  It will be easier to see that nothing is suppose to change.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -130,6 +130,32 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
...
 
+bool CommonUniformElimPass::IsVolatileVar(uint32_t varId) {
+  for (auto * dec : dec_mgr_->GetDecorationsFor(varId, false)) {
+    uint32_t decoration;
+    switch(dec->opcode()) {
+      case SpvOpDecorate: decoration = dec->GetSingleWordOperand(1);
+        break;
+      case SpvOpMemberDecorate: decoration = dec->GetSingleWordOperand(2);
+        break;
+      default: continue;
```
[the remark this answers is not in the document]
daniel-schuermann: GetDecorationsFor() should only return OpDecoration or OpMemberDecoration as far as I understand. Nonetheless, I added an assert.

test/opt/common_uniform_elim_test.cpp
daniel-schuermann: Forgot to change the description: I changed one Load to Volatile.

test/opt/common_uniform_elim_test.cpp
daniel-schuermann: I would leave it for consistency. I the volatile1 test, this load gets replaced by extract, here it doesn't.

daniel-schuermann: I would leave it for consistency. In volatile1, this load gets changed to extract, here it doesn't.

daniel-schuermann:
Thank you @s-perron for your feedback!
I included your requested changes. Please let me know, if there is something left.

test/opt/common_uniform_elim_test.cpp
```diff
@@ -664,6 +664,268 @@ OpFunctionEnd
...
+OpBranch %30
+%30 = OpLabel
+%45 = OpLoad %v4float %v
+OpStore %gl_FragColor %45
+OpReturn
+OpFunctionEnd
+)";
+
+  opt::Pass::Status res = std::get<1>(SinglePassRunAndDisassemble<opt::CommonUniformElimPass>(text, true));
+  EXPECT_TRUE(res == opt::Pass::Status::SuccessWithoutChange);
```
s-perron: You can use EXPECT_EQ here.  It will give a better message explaining the failure if it ever fails.

s-perron: Sorry EXPECT_THAT is the preferred way of doing things.

daniel-schuermann: No problem, give me two hours and I will be at home to change that.

dneto0: No worries about this one.   We don't have hard-and-fast rules, and we haven't written them down.

For booleans, EXPECT_TRUE and EXPECT_FALSE are probably the best alternatives.

For other  containers (like vectors, strings, and sets), EXPECT_THAT  with Eq or other predicates is often the best because the failure messages are really good.  Also you get a lot of flexibility.

For stuff in between, like scalars, I'm fine with either EXPECT_EQ  or EXPECT_THAT with Eq.   Out of habit and consistency with the container checks, I tend to use EXPECT_THAT.

dneto0: Now I read it again after more sips of coffee.    :-)

It's better to avoid building an expression to check.  I agree with @s-perron on this one.  The preferred form is:

```
EXPECT_THAT(res, Eq(opt::Pass::Status::SuccessWithoutChange) 
```

daniel-schuermann: I left EXPECT_EQ, if that is fine for you. The error messages are identical and EXPECT_THAT requires additional #includes.

daniel-schuermann:
Just found a breaking issue :(
OpMemberDecorates are still ignored. So, if you have
struct S {
volatile float a;
float b; }
Then, a is optimized, regardless. Please mark as broken (or something). I will shortly add a non-working test just in case.

daniel-schuermann:
Seems, my intention was right and volatile is decorated at the type and not at the variable. I rewrote the test (now with glslang), which is now not working. Shall I open a new PR, when I'm done or leave this open until the test works?

dneto0:
It's true that for a struct variable decorations are applied on the struct type's members.

I'm a bit confused, so I'll just state my high level request:

I'd like this to go in as one pull request.  It's less confusing later on that way.

So what is left to do?   What is broken?

daniel-schuermann:
Sorry for the confusion...
Broken is the check for volatile structs / struct members. Working is the check for volatile Loads.
Left to do is getting the type id for a variable, should be doable over the week-end (if not tonight).

dneto0:
Ok.  Thanks for the reply.  There is no rush from my end. Take your time!  :-)

daniel-schuermann:
This should do it now. Sorry again.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -130,6 +130,51 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
   return ptrInst;
 }
 
+bool CommonUniformElimPass::IsVolatileStructType(const ir::Instruction& AccessChainInst) {
+  assert(AccessChainInst.opcode() == SpvOpAccessChain);
+  uint32_t memberTypeId = AccessChainInst.GetSingleWordOperand(3);
```
dneto0: This is a confusing name.  It's the Id of the first index value.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -130,6 +130,51 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
...
+  // OpConstant
```
dneto0: You can only know if this is a constant if you've proven that the base pointer has a struct type.  You should check that first;  return false before this point if it's not a struct type.

Now, it's true that the code for Vulkan will always have a struct as the type of a UBO or SSBO.    But I think OpenGL's use allows "loose" uniforms, so you should guard against that here.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -130,6 +130,51 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
...
+  const ir::Instruction* memberTypeDef = def_use_mgr_->GetDef(memberTypeId);
+  uint32_t member = memberTypeDef->GetSingleWordOperand(2);
+  uint32_t VarId = AccessChainInst.GetSingleWordOperand(2);
+  // OpVariable
+  const ir::Instruction* VarInst = def_use_mgr_->GetDef(VarId);
+  uint32_t ptrTypeId = VarInst->GetSingleWordOperand(0);
+  // OpTypePointer
+  const ir::Instruction* TypePtrInst = def_use_mgr_->GetDef(ptrTypeId);
+  uint32_t typeId = TypePtrInst->GetSingleWordInOperand(kTypePointerTypeIdInIdx);
+  for (auto * dec : dec_mgr_->GetDecorationsFor(typeId, false)) {
```
dneto0: (Idea for refactoring later:  We should have a ForEachDecoration idiom.)

No request for a change here.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -130,6 +130,51 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
...
+    uint32_t decoration = 0;
+    switch (dec->opcode()) {
+      case SpvOpDecorate:
+        decoration = dec->GetSingleWordOperand(1);
+        break;
+      case SpvOpMemberDecorate:
+        // check if decoration applies to member in OpAccessChain
+        if (member != dec->GetSingleWordOperand(1))
```
dneto0: This code does the right thing if there are two member decorations.

This is an interesting test case.

struct S {
    lowp volatile float x;  
};

I would expect both  
  OpMemberDecorate %S 0 RelaxedPrecision  // possibly, I haven't checked
  OpMemberDecorate %S 0 Volatile

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -130,6 +130,51 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
...
+  // OpConstant
+  const ir::Instruction* memberTypeDef = def_use_mgr_->GetDef(memberTypeId);
+  uint32_t member = memberTypeDef->GetSingleWordOperand(2);
+  uint32_t VarId = AccessChainInst.GetSingleWordOperand(2);
+  // OpVariable
+  const ir::Instruction* VarInst = def_use_mgr_->GetDef(VarId);
+  uint32_t ptrTypeId = VarInst->GetSingleWordOperand(0);
+  // OpTypePointer
+  const ir::Instruction* TypePtrInst = def_use_mgr_->GetDef(ptrTypeId);
+  uint32_t typeId = TypePtrInst->GetSingleWordInOperand(kTypePointerTypeIdInIdx);
```
dneto0: This sees the volatile when:
 - the base pointer is of struct type S
 - the volatile is on a member of S

It misses cases where the volatile is on a nested struct.  E.g.
 struct Inner {
      volatile float x;
  }; 
  struct Outer {
    Inner a;
  };

layout(set = 0, binding = 0)  buffer B {
  Outer o;
} Bdat;
...
   float f = Bdat.o.a.x;


Either:
 -  add a TODO for the nested struct case, or
 -  handle the nested struct case.   

For the nested struct case, you'll have to walk the type tree, determining the pointer type at each step.   We really should have a common utility for this, but we don't yet.  For now you can look at this sample work-in-progress code that I wrote that does something similar:

https://github.com/dneto0/SPIRV-Tools/blob/robust-access/source/opt/graphics_robust_access_pass.cpp#L188

source/opt/common_uniform_elim_pass.h
```diff
@@ -66,6 +67,14 @@ class CommonUniformElimPass : public Pass {
   // Return true if variable is uniform
   bool IsUniformVar(uint32_t varId);
 
+  // Given an OpAccessChain instruction, return true
```
dneto0: You should say more about the assumptions:
- the base pointer is an OpVariable
- the variable is of type pointer-to-struct

Also say it only sees volatile on a member of that struct type.  (See discussion above)

source/opt/common_uniform_elim_pass.cpp
dneto0: I suggest you add a test case like this.  It will ensure the code isn't broken by a rewrite.

source/opt/common_uniform_elim_pass.cpp
daniel-schuermann: Good point, although I don't know how to write this in glsl. I made a test with a single inner volatile struct and took you approach to walk the type tree.

source/opt/common_uniform_elim_pass.cpp
daniel-schuermann: I tested it and it's fine. It adds just another decoration as expected. I iterate over all decorations, so there shouldn't be a problem.
However, I realized that the we already don't want to optimize if _any_ member is decorated volatile, so i skipped that part.

daniel-schuermann:
Thanks for all the suggestions. If everything is fine, I will squash the commits. Or should I do this always directly?

dneto0:
Thanks for the update.

I usually prefer for code changes be presented as new commits rather than squashed commits.  It makes it easier to see what changed in GitHub's interface.    The approver (usually me) will squash, rebase, and retest before pushing onto master.   That will maintain linear history, and ensure that everything should work at each commit in the history.

(I plan to write up our process and explain why.)

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -130,6 +128,68 @@ ir::Instruction* CommonUniformElimPass::GetPtr(
...
+        pointee_type_id = pointee_type->GetSingleWordOperand(1);
+        break;
+      case SpvOpTypeStruct:
+        // check for volatile decorations:
+        if (IsVolatileStruct(pointee_type_id)) return true;
+
+        if (idx < num_operands - 1) {
+          const uint32_t index_id = AccessChainInst.GetSingleWordOperand(idx);
+          const ir::Instruction* index_inst = def_use_mgr_->GetDef(index_id);
+          uint32_t index_value = index_inst->GetSingleWordOperand(2); // TODO: replace with GetUintValueFromConstant()
```
dneto0: There's a universal limit of 16383 struct members.  So this actually should work forever.  :-)
See the table in section 2.17.

It's certainly ok to keep the TODO.

dneto0:
Rebased and pushed into master as  97990dc

PR #883, opened 2017-10-16
DeadBranchElim: Simplify and fix dead block elimination

greg-lunarg:
The previous algorithm would leave invalid code in the case of unreachable
  blocks pointing into a dead branch. It would leave the unreachable blocks
  branching to labels that no longer exist. The previous algorithm also left
  unreachable blocks in some cases (eg a loop dominated by an orphaned
  merge block). This fix also addresses that.
    
  This code will soon be replaced with the coming CFG cleanup, but in the meantime...

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -24,6 +24,7 @@ namespace opt {
 
 namespace {
 
+const uint32_t kBranchTargetLabIdInIdx = 0;
 const uint32_t kBranchCondTrueLabIdInIdx = 1;
 const uint32_t kBranchCondFalseLabIdInIdx = 2;
 const uint32_t kSelectionMergeMergeBlockIdInIdx = 0;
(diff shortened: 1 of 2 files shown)


source/opt/dead_branch_elim_pass.cpp
```diff
@@ -185,20 +186,57 @@ bool DeadBranchElimPass::GetSelectionBranch(ir::BasicBlock* bp,
...
+        }
+        const uint32_t fLabId = ii->GetSingleWordInOperand(
+            kBranchCondFalseLabIdInIdx);
+        if (visited.find(fLabId) != visited.end())
+          backedges_.insert(&*ii);
+      } break;
+      default:
+        break;
+    }
+    visited.insert((*bi)->id());
```
s-perron: I think you need to insert into visited at the start of the loop.  Otherwise you will miss back edges of loops that contain just a single basic block.

dneto0: Agreed.

greg-lunarg:
I have addressed the one issue @s-perron noted.

I agree with ultimately replacing the dead block elimination here, but I would request that we allow this fix through so that @danginsburg can proceed while we work that out.

dneto0:
Rebased, squashed, and pushed into master as e3a720933046

PR #884, opened 2017-10-17
Refactored compression debugger

atgoo:
Markv codec now receives two optional callbacks:
- LogConsumer for internal codec logging
- DebugConsumer for testing if encoding->decoding produces the original
results.

tools/comp/spirv-markv now has new task 't' which tests the codec on a given file

### DIFF
--- a/source/comp/markv.h
+++ b/source/comp/markv.h
@@ -29,35 +29,48 @@
 
 namespace spvtools {
 
-struct MarkvEncoderOptions {
+struct MarkvCodecOptions {
   bool validate_spirv_binary = false;
 };
 
-struct MarkvDecoderOptions {
-  bool validate_spirv_binary = false;
-};
+// Debug callback. Called once per instruction.
+// |words| is instriuction SPIR-V words.
+// |bits| is a string of MARK-V bits used to encode the instruction.
+// |comment| contains all logs generated while processing the instruction.
+using MarkvDebugConsumer = std::function<bool(
+    const std::vector<uint32_t>& words, const std::string& bits,
+    const std::string& comment)>;
+
+// Logging callback. Called often (if decoder reads a single bit, the log
+// consumer will receive 1 character string with that bit).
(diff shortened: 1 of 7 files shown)


source/comp/markv.h
```diff
@@ -29,35 +29,48 @@
...
-struct MarkvEncoderOptions {
+struct MarkvCodecOptions {
   bool validate_spirv_binary = false;
 };
 
-struct MarkvDecoderOptions {
-  bool validate_spirv_binary = false;
-};
+// Debug callback. Called once per instruction.
+// |words| is instriuction SPIR-V words.
```
dneto0: "instruction"

source/comp/markv.h
```diff
@@ -29,35 +29,48 @@
...
+// |bits| is a string of MARK-V bits used to encode the instruction.
```
dneto0: What's the representation?  Does a 0 bit become a "0" character?

source/comp/markv.h
```diff
@@ -29,35 +29,48 @@
...
+// |comment| contains all logs generated while processing the instruction.
+using MarkvDebugConsumer = std::function<bool(
+    const std::vector<uint32_t>& words, const std::string& bits,
+    const std::string& comment)>;
+
+// Logging callback. Called often (if decoder reads a single bit, the log
```
dneto0: What is in the "snippet"?  Is it just informational?

source/comp/markv_codec.cpp
```diff
@@ -290,6 +292,8 @@ class CommentLogger {
   }
 
   void AppendBitSequence(const std::string& str) {
+    if (debug_consumer_)
```
dneto0: Today I learned that a std::function is convertible to bool, and its meaning.    :-)

atgoo:
Ready for review.

dneto0:
Rebased, squashed, and pushed into master as cfd95f3d

PR #885, opened 2017-10-17
Re-factor generic constant folding code out of fold spec constants pass

dnovillo:
There are no functional changes in this patch.  The generic folding
routines in FoldSpecConstantOpAndCompositePass are now inside opt/fold.{cpp,h}.
This code will be used by the upcoming constant propagation pass.  In
time, we'll add more expression folding and simplification into these
two files.


Qining:
lgtm, and thanks for adding back the missing terminating comments for anonymous namespaces.

PR #886, opened 2017-10-18
val & test of literal's upper bits

### LINKED ISSUE #660: Verify upper bits of literal numbers that aren't a multiple of 32 bits.
See 
https://github.com/KhronosGroup/SPIRV-Tools/pull/659

jcaraban:
Fixes #660 

Submitting for feedback.

### DIFF
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -111,7 +111,7 @@ class Parser {
   spv_result_t setNumericTypeInfoForType(spv_parsed_operand_t* parsed_operand,
                                          uint32_t type_id);
 
-  // For literal numbers that aren not multiple of 32 bits (i.e. 16b, 8b),
+  // For literal numbers that are not multiple of 32 bits (i.e. 16b, 8b),
   // verifies that the upper bits of the word are all 0 or sign-extended
   spv_result_t verifyLiteralUpperBits(
     const spv_parsed_operand_t& parsed_operand, uint32_t word);
@@ -741,13 +741,15 @@ spv_result_t Parser::setNumericTypeInfoForType(
 
 spv_result_t Parser::verifyLiteralUpperBits(
     const spv_parsed_operand_t& parsed_operand, uint32_t word) {
-  // Verify the upper bits of literal numbers not multiple of 32 bits
+  // Verify that the upper bits are either set to 0, or sign-extended
   auto verify = [&](uint32_t mask, std::vector<uint32_t> valid_patterns)
       -> spv_result_t {
     uint32_t upper = word & mask;
-    for (uint32_t valid : valid_patterns)
-        if (upper == valid)
-          return SPV_SUCCESS;
+    for (uint32_t valid : valid_patterns) {
+      if (upper == valid) {
+        return SPV_SUCCESS;
+      }
+    }
     return diagnostic() << "Upper bits of literal number are invalid";
   };
 
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -111,6 +111,11 @@ class Parser {
   spv_result_t setNumericTypeInfoForType(spv_parsed_operand_t* parsed_operand,
                                          uint32_t type_id);
 
+  // For literal numbers that aren not multiple of 32 bits (i.e. 16b, 8b),
+  // verifies that the upper bits of the word are all 0 or sign-extended
+  spv_result_t verifyLiteralUpperBits(
+    const spv_parsed_operand_t& parsed_operand, uint32_t word);
+
   // Records the number type for an instruction at the given offset, if that
   // instruction generates a type.  For types that aren't scalar numbers,
   // record something with number kind SPV_NUMBER_NONE.
@@ -505,7 +510,7 @@ spv_result_t Parser::parseOperand(size_t inst_offset,
       break;
 
     case SPV_OPERAND_TYPE_TYPED_LITERAL_NUMBER:
-    case SPV_OPERAND_TYPE_OPTIONAL_TYPED_LITERAL_INTEGER:
+    case SPV_OPERAND_TYPE_OPTIONAL_TYPED_LITERAL_INTEGER: {
       parsed_operand.type = SPV_OPERAND_TYPE_TYPED_LITERAL_NUMBER;
       if (opcode == SpvOpSwitch) {
         // The literal operands have the same type as the value
@@ -541,7 +546,9 @@ spv_result_t Parser::parseOperand(size_t inst_offset,
                 setNumericTypeInfoForType(&parsed_operand, inst->type_id))
           return error;
       }
-      break;
+      if (auto error = verifyLiteralUpperBits(parsed_operand,word))
+        return error;
+    } break;
 
     case SPV_OPERAND_TYPE_LITERAL_STRING:
     case SPV_OPERAND_TYPE_OPTIONAL_LITERAL_STRING: {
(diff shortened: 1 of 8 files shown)


source/binary.cpp
```diff
@@ -111,6 +111,11 @@ class Parser {
   spv_result_t setNumericTypeInfoForType(spv_parsed_operand_t* parsed_operand,
                                          uint32_t type_id);
 
+  // For literal numbers that aren not multiple of 32 bits (i.e. 16b, 8b),
```
dnovillo: s/aren not/are not/

source/binary.cpp
```diff
@@ -732,6 +739,41 @@ spv_result_t Parser::setNumericTypeInfoForType(
   return SPV_SUCCESS;
 }
 
+spv_result_t Parser::verifyLiteralUpperBits(
+    const spv_parsed_operand_t& parsed_operand, uint32_t word) {
+  // Verify the upper bits of literal numbers not multiple of 32 bits
+  auto verify = [&](uint32_t mask, std::vector<uint32_t> valid_patterns)
```
dnovillo: s/not multiple/are not multiple/ ?

Also, this seems weirdly indented.

source/binary.cpp
```diff
@@ -732,6 +739,41 @@ spv_result_t Parser::setNumericTypeInfoForType(
...
+      -> spv_result_t {
+    uint32_t upper = word & mask;
+    for (uint32_t valid : valid_patterns)
```
dnovillo: Prefer braces around single-line control-flow bodies.

source/binary.cpp
```diff
@@ -732,6 +739,41 @@ spv_result_t Parser::setNumericTypeInfoForType(
...
+        if (upper == valid)
+          return SPV_SUCCESS;
+    return diagnostic() << "Upper bits of literal number are invalid";
+  };
+
+  switch (parsed_operand.number_kind) {
+    case SPV_NUMBER_SIGNED_INT:
+      switch (parsed_operand.number_bit_width) {
+        case 8:  return verify(0xFFFFFF00,{0x00000000,0xFFFFFF00});
+        case 16: return verify(0xFFFF0000,{0x00000000,0xFFFF0000});
```
dnovillo: Question out of sheer ignorance.  There are no 32 bit integers?

source/binary.cpp
dnovillo: s/aren not/are not/

source/binary.cpp
```diff
@@ -732,6 +739,41 @@ spv_result_t Parser::setNumericTypeInfoForType(
   return SPV_SUCCESS;
 }
 
+spv_result_t Parser::verifyLiteralUpperBits(
+    const spv_parsed_operand_t& parsed_operand, uint32_t word) {
+  // Verify the upper bits of literal numbers not multiple of 32 bits
```
dnovillo: I'm not sure I understand this comment.  This is verifying that the upper bits are either not set or sign-extended, right?  Not sure how that is considered "multiple of 32 bits".

source/binary.cpp
dnovillo: Prefer braces around one-line control-flow bodies.

source/binary.cpp
jcaraban: This check only targets those literal numbers smaller than the word size, i.e. smaller than 32 bits. Therefore, any 32 bit and 64 bit literals pass this test automatically, through the default branch. Thanks for the comments!

source/binary.cpp
```diff
@@ -732,6 +739,43 @@ spv_result_t Parser::setNumericTypeInfoForType(
   return SPV_SUCCESS;
 }
 
+spv_result_t Parser::verifyLiteralUpperBits(
+    const spv_parsed_operand_t& parsed_operand, uint32_t word) {
+  // Verify that the upper bits are either set to 0, or sign-extended
+  auto verify = [&](uint32_t mask, std::vector<uint32_t> valid_patterns)
```
dneto0: FYI:  The style guide we're using has some non-obvious advice about lambda captures.  See https://google.github.io/styleguide/cppguide.html#Lambda_expressions  

"Use default capture by reference ([&]) only when the lifetime of the lambda is obviously shorter than any potential captures."

Your use here is ok by that guide.  :-)

source/binary.cpp
```diff
@@ -732,6 +739,43 @@ spv_result_t Parser::setNumericTypeInfoForType(
...
+        return SPV_SUCCESS;
+      }
+    }
+    return diagnostic() << "Upper bits of literal number are invalid";
+  };
+
+  switch (parsed_operand.number_kind) {
+    case SPV_NUMBER_SIGNED_INT:
+      switch (parsed_operand.number_bit_width) {
+        case 8:  return verify(0xFFFFFF00,{0x00000000,0xFFFFFF00});
```
dneto0: Generic SPIR-V can have any-width numeric types.  So this code is too specific.
It should work with 1-bit , 12-bit, 51-bit integers, for example.

So far we've limited SPIRV-Tools to handle any integer bit width up to and including 64 bits.

For floats, we only handle 16, 32, and 64 bit floats.

source/binary.cpp
```diff
@@ -541,7 +546,9 @@ spv_result_t Parser::parseOperand(size_t inst_offset,
                 setNumericTypeInfoForType(&parsed_operand, inst->type_id))
           return error;
       }
-      break;
+      if (auto error = verifyLiteralUpperBits(parsed_operand,word))
```
dneto0: This should be done in the validator instead of in the binary parser.

Why?   The binary parser is the lowest layer: both the disassembler and the validator are built on top of it.  The disassembler is used to help debug problems.  In a bad case you're catching, it's usually preferable to see a disassembly rather than a truncated failed parse.

So, the binary parser generally does as little as possible.  It only does enough to ensure well-formedness, with validity being a whole other bunch of work.

Since you can defer these checks to the validator, you should do so.

And I realize none of this is written down where you'd be able to find it.   :-(

test/binary_to_text_test.cpp
```diff
@@ -189,6 +189,16 @@ INSTANTIATE_TEST_CASE_P(
          "%2 = OpTypeVector %1 4",
          spvtest::MakeInstruction(SpvOpConstant, {2, 3, 999}),
          "Type Id 2 is not a scalar numeric type"},
+        // Three tests to check the upper bits of literal numbers
+        {"%1 = OpTypeInt 8 1",
+        spvtest::MakeInstruction(SpvOpConstant, {1, 2, 0xABCDEF00}),
+        "Upper bits of literal number are invalid"},
+        {"%1 = OpTypeInt 16 1",
```
dneto0: You should also exercise cases between 33 and 63 bits.

source/binary.cpp
jcaraban: I targeted the binary parser because there things happen at word level.
As a result, I could handled all literals with a same function, a same logic.

The validator works at instruction level, with different logics checking each of them.
As a result, now I need to inject new code to every instruction accepting literals.

right @dneto0? or do we have a common place in the validator where to check literals?

test/binary_to_text_test.cpp
jcaraban: I wonder, is there is a better option than parameterized tests (TEST_P + INSTANTIATE) for cases like this, where we need the very same logic for all bit sizes?

jcaraban:
@dneto0 I’m stopping this MR until I receive further guidance,
given that I don’t know where to place the _literal’s upper bits_ verification.

Verifying at instruction level seems a bad approach to me,
because there are just too many opcodes accepting literals.
A quick search in the spec reveals about 40 cases using literals...

Verifying at word/operand level in the binary parser seems more appropriate,
but the disassembler, which also uses the parser, would then be affected.

Would it be possible to have a word/operand-pass in the validator? 
I don’t see other way to easily validate the 40 cases with literals.

dneto0:
I suggest adding a step to ProcessInstruction in validate.cpp.

The spv_parsed_instruction_t has operand-kind information for each operand.  For literal number operands it has kind (floating, or signed int or unsigned int) and bit width.

jcaraban:
By "step" I understood a new Pass, and that's what I did.
However, a whole pass with its tests file and executable seems overkill.
Perhaps _literals validation_ logic could be moved into another, more general pass.

source/validate_literals.cpp
```diff
@@ -0,0 +1,89 @@
...
+// Verifies that the upper bits of the given upper |word| with given
+// lower |width| are zero- or sign-extended when |sign| is true
+bool verifyUpperBits(uint32_t word, uint32_t width, bool sign) {
+  assert(width < 32);
+  uint32_t upper_mask = (0xFFFFFFFF >> width) << width;
+  uint32_t upper_bits = word & upper_mask;
+  if (upper_bits == 0x00000000)
+      return true; // zero-extended
+  if (sign && upper_bits == upper_mask)
+      return true; // sign-extended
```
dneto0: This is not right.  It allows the following case:
```
               OpCapability Shader
               OpCapability Linkage
               OpCapability Int16
               OpMemoryModel Logical GLSL450
%short = OpTypeInt 16 1
%bad = OpConstant %short 0xff00
```

So this is a signed 16-bit integer but the sign bit is 0 and the mask is 1.

test/val/val_literals_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+              + inst_id + " must be 0 for a floating-point type, "
+              "or 0 for an integer type with Signedness of 0, "
+              "or sign extended when Signedness is 1"));
+}
+
+INSTANTIATE_TEST_CASE_P(LiteralsShaderCases, ValidateLiteralsShader,
+    ::testing::Values(
+      "%11 = OpConstant %int16  !0xABCD0000",
+      "%11 = OpConstant %uint16 !0xABCD0000",
+      "%11 = OpConstant %half   !0xABCD0000"
```
dneto0: Also need a case like   OpConstant %int16 0xff00

source/validate_literals.cpp
```diff
@@ -0,0 +1,89 @@
...
+spv_result_t LiteralsPass(ValidationState_t& _,
+                            const spv_parsed_instruction_t* inst) {
+  // For every operand that is a literal number
+  for (uint16_t i=0; i<inst->num_operands; i++) {
+    const spv_parsed_operand_t* operand = inst->operands + i;
+    if (!IsLiteralNumber(operand)) continue;
+
+    // The upper bits are always in the last word (little-endian)
+    int last_index = operand->offset + operand->num_words - 1;
+    uint32_t upper_word = inst->words[last_index];
```
dneto0: Generally I recommend using const.

source/validate_literals.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+// Verifies that the upper bits of the given upper |word| with given
+// lower |width| are zero- or sign-extended when |sign| is true
+bool verifyUpperBits(uint32_t word, uint32_t width, bool sign) {
+  assert(width < 32);
+  uint32_t upper_mask = (0xFFFFFFFF >> width) << width;
+  uint32_t upper_bits = word & upper_mask;
+  if (upper_bits == 0x00000000)
+      return true; // zero-extended
+  if (sign && upper_bits == upper_mask)
```
dneto0: The "sign" variable only says whether it's a signed number format.  It doesn't say whether the sign bit is 0 or 1.

dneto0:
Content pushed into master via #1092

Issue #855, opened 2017-10-03
spirv-tools does not build with gcc 4.8 (default for steam-runtime)


danginsburg:
I get several instances of the following internal compiler error attempting to build spirv-tools with gcc 4.8:

```
In file included from /data/data/source2_main/src/thirdparty/glslang/External/spirv-tools/source/util/parse_number.h:23:0,
                 from /data/data/source2_main/src/thirdparty/glslang/External/spirv-tools/source/util/parse_number.cpp:15:
/data/data/source2_main/src/thirdparty/glslang/External/spirv-tools/source/util/hex_float.h: In instantiation of 'typename other_T::uint_type spvutils::HexFloat<T, Traits>::getRoundedNormalizedSignificand(spvutils::round_direction, bool*) [with other_T = spvutils::HexFloat<spvutils::FloatProxy<spvutils::Float16>, spvutils::HexFloatTraits<spvutils::FloatProxy<spvutils::Float16> > >; T = spvutils::FloatProxy<float>; Traits = spvutils::HexFloatTraits<spvutils::FloatProxy<float> >; typename other_T::uint_type = short unsigned int]':
/data/data/source2_main/src/thirdparty/glslang/External/spirv-tools/source/util/hex_float.h:580:40:   required from 'void spvutils::HexFloat<T, Traits>::castTo(other_T&, spvutils::round_direction) [with other_T = spvutils::HexFloat<spvutils::FloatProxy<spvutils::Float16>, spvutils::HexFloatTraits<spvutils::FloatProxy<spvutils::Float16> > >; T = spvutils::FloatProxy<float>; Traits = spvutils::HexFloatTraits<spvutils::FloatProxy<float> >]'
/data/data/source2_main/src/thirdparty/glslang/External/spirv-tools/source/util/hex_float.h:815:51:   required from here
/data/data/source2_main/src/thirdparty/glslang/External/spirv-tools/source/util/hex_float.h:490:63: internal compiler error: in dependent_type_p, at cp/pt.c:19496
             : negatable_left_shift<num_throwaway_bits>::val(1u);
                                                               ^
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-4.8/README.Bugs> for instructions.
```

gcc-4.8 is currently the default compiler for steam-runtime (https://github.com/ValveSoftware/steam-runtime)

dneto0:
[...]

PR #887, opened 2017-10-18
Android.mk: add source/opt/fold.cpp

dneto0:


dneto0:
This compensates for inclusion of #855 and is needed to ensure we can merge https://github.com/google/shaderc/pull/391

dneto0:
LGTM myself.

PR #888, opened 2017-10-19
Add OpBranchConditional validation

nsubtil:
Validates that the type of the condition variable for
OpBranchConditional is valid.

### DIFF
--- a/source/validate_branch.cpp
+++ b/source/validate_branch.cpp
@@ -1,0 +1,68 @@
+// Copyright (c) 2017 NVIDIA Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Validates branch instructions
+
+#include "validate.h"
+
+#include "diagnostic.h"
+#include "opcode.h"
+#include "val/instruction.h"
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+namespace {
+
+// Returns operand word for given instruction and operand index.
+// The operand is expected to only have one word.
+inline uint32_t GetOperandWord(const spv_parsed_instruction_t* inst,
+                               size_t operand_index) {
+  assert(operand_index < inst->num_operands);
+  const spv_parsed_operand_t& operand = inst->operands[operand_index];
+  assert(operand.num_words == 1);
+  return inst->words[operand.offset];
+}
+
+// Returns the type id of instruction operand at |operand_index|.
+// The operand is expected to be an id.
+inline uint32_t GetOperandTypeId(ValidationState_t& _,
+                               const spv_parsed_instruction_t* inst,
+                               size_t operand_index) {
+  return _.GetTypeId(GetOperandWord(inst, operand_index));
+}
+
+} // namespace
+
+spv_result_t BranchPass(ValidationState_t& _,
+                        const spv_parsed_instruction_t* inst) {
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+
+  switch(opcode) {
+    case SpvOpBranchConditional:
+      if (inst->num_operands < 3) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Expected a minimum of 3 operands in " << spvOpcodeString(opcode);
+      }
+
+      if (!_.IsBoolScalarType(GetOperandTypeId(_, inst, 0))) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Condition operand for " << spvOpcodeString(opcode) << " must be of bool scalar type.";
+        }
+    }
+
+    return SPV_SUCCESS;
+}
+
+} // namespace libspirv
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -255,6 +255,7 @@ set(SPIRV_SOURCES
   ${CMAKE_CURRENT_SOURCE_DIR}/validate.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/validate_arithmetics.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/validate_bitwise.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/validate_branch.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/validate_cfg.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/validate_capability.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/validate_datarules.cpp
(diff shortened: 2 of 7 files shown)


dneto0:
Thanks for the contribution!

You'll need to fix the problems detected by the bots.  Click through on the "Details"  for the Travis build.
E.g. you'll find this error: https://travis-ci.org/KhronosGroup/SPIRV-Tools/jobs/289768690#L547

source/validate_branch.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+} // namespace
+
+spv_result_t BranchPass(ValidationState_t& _,
+                        const spv_parsed_instruction_t* inst) {
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+
+  switch(opcode) {
+    case SpvOpBranchConditional:
```
atgoo: Consider adding braces:
case SpvOpBranchConditional: {
}

source/validate_branch.cpp
```diff
@@ -0,0 +1,68 @@
...
+      if (inst->num_operands < 3) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Expected a minimum of 3 operands in " << spvOpcodeString(opcode);
+      }
+
+      if (!_.IsBoolScalarType(GetOperandTypeId(_, inst, 0))) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Condition operand for " << spvOpcodeString(opcode) << " must be of bool scalar type.";
+        }
+    }
```
atgoo: You need an empty default: case, or it won't build.

source/validate_branch.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+} // namespace
+
+spv_result_t BranchPass(ValidationState_t& _,
+                        const spv_parsed_instruction_t* inst) {
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+
+  switch(opcode) {
+    case SpvOpBranchConditional:
+      if (inst->num_operands < 3) {
```
atgoo: inst->num_operands should be either 3 or 5. If 5 then the last two need to be literal integers.

source/validate_branch.cpp
```diff
@@ -0,0 +1,68 @@
...
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Expected a minimum of 3 operands in " << spvOpcodeString(opcode);
+      }
+
+      if (!_.IsBoolScalarType(GetOperandTypeId(_, inst, 0))) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Condition operand for " << spvOpcodeString(opcode) << " must be of bool scalar type.";
+        }
```
atgoo: Operands 1 and 2 have already been declared, they should be OpLabel.

nsubtil:
Yeah, I wasn't sure if this was the right pattern to follow for branch validation. Is there an example of how to do this after the entire module is parsed? Would that be more amenable here?

For context, this stems from a case we've seen in the wild, where the condition operand is not a boolean scalar. Since I haven't contributed to SPIRV-Tools before, I was hoping to use this one tiny case as a pipe-cleaner to iron out any process or style issues --- based on the feedback, that seems to be working :) So I'm happy to rework if that's better, and also extend it to validate all the conditions mentioned in the spec.

nsubtil:
Would the idUsage checks be a suitable place to check the conditional type as well? Or is the intent that we'd implement the target label ID validation there and the conditional value type checking somewhere else?

I *think* the conditional variable ID declaration has to precede the branch, though on second look it's not clear that it does. Can you clarify how these details are defined in the spec?

atgoo:
Operand 0: Bool scalar needs to be defined before OpBranchConditional
Operand 1, 2: OpLabel can come after OpBranchConditional
Operand 3, 4: just literal numbers.

So everything except labels can be checked here. But labels can only be checked if they already exist.

Where are two options:
1. Check labels in IdPass and everything else while parsing, but I'm not sure if we need a new file, since there is already validate_cfg.cpp which processes this opcode.
2. Check everything in IdPass.

What do you think, David?

dneto0:
Thanks @atgoo for catching this subtle issue.

But I think this statement is wrong:

>  IdPass is called after the entire module was parsed and could be more suitable for this kind of check.

I think you meant spvValidateIDs?

(Looking at  ValidateBinaryUsingContextAndValidationState  it splits into two phases:
The spvBinaryParse uses ProcessInstruction as the callback, and so ProcessInstruction is being called incrementally as each instruction is encountered.   That applies to the things it calls, including IdPass.)

But spvValidateIDs, and ultimately the code in validate_id.cpp is run after the entire module has been seen once, and ID definitions are tracked.

I think it's better to just move the checks to idUsage::isValid<SpvOpBranchConditional> in validate_id.cpp.

.... and then I see that the idUsge object doesn't have a reference to the validation state.  Sadly that validate_id.cpp is the oldest (legacy) part of the validator.  ::facepalm::

So, I recommend:
 - give the idUsage a reference to the validation state so it can do ID lookups like the newer parts of the validator.  That means plumbing the validator-state object down via spvValidateIDs and spvValidateInstructionIDs
 - add your BranchConditional check to isUsage.

dneto0:
I agree with @atgoo that it's better not add a new file for this.  It should fit in an existing module.
If you're concerned about recording the copyright (which is very understandable), then add the NVIDIA copyright below the current one.  And if that causes the copyright script to fail, then generalize it.  :-)

atgoo:
I meant this function:
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/validate_id.cpp#L2656

It has validation state, but you can't use all of its features. For example you can't ask for a reference to the current function because there is no current function.

nsubtil:
Picking this back up...

@dneto0, are you suggesting that all of the validation per instruction should be done in isUsage? In other words, the initial pass would only do enough work to resolve all of the IDs, and then all of the per-instruction validation is done in one pass after that?

This would make a lot of sense to me, as it would ensure that all of the validation for a given opcode is effectively done in one location, making it easier to verify. (It would also open up the possibility of auto-generating some of this code from the spec grammar in the future, which might be desirable.)

dneto0:
@nsubtil :  Essentially yes:  There are two phases:
The first phase is streaming-only and so it can't see future IDs
The second phase has access to tables built over the entire module.  This is where your OpBranchConditional check must go, since OpBranchConditional IDs may be forward references.

About auto-generating parts of the checks.  The binary parser uses tables generated from the JSON grammar files in the SPIRV-Headers repo.  So you get well-formedness (a.k.a. syntax) checks for free. 

I'm reticent about going much further since the spec text itself is not particuarly amenable to bullet-proof translation into rules.  @atgoo has checks for a few classes of instructions can be written cleanly and very readably.  (e.g. type checks for arithmetic, etc.)

nsubtil:
Yeah, parsing English text isn't easy :)

In my mind, we could start formalizing some of these rules in the spec grammar --- it would probably be easier than writing the corresponding C++ by hand. In the ideal scenario a large portion of the validation code would be auto-generated, with hand-written code for less common cases.

Admittedly, however, I don't yet have a very good sense for how uniform these checks are in practice. This came to mind while working on two follow-on PRs that expand the OpBranch* validation and add a bunch of checks to OpImage* instructions, where I found myself doing similar checks for all of them. I haven't really had exposure to the majority of opcodes, however.

Meanwhile, I'll work on fixing up this PR. Thanks for the insights and discussion!

atgoo:
I'm currently working on validator checks for all OpImageXXX instructions.
https://github.com/KhronosGroup/SPIRV-Tools/issues/900

nsubtil:
Superseded by #925

PR #905, opened 2017-10-20
Add pass to remove dead variables at the module level.

s-perron:
There does not seem to be any pass that remove global variables.  I
think we could use one.  This pass will look specifically for global
variables that are not referenced and are not exported.  Any decoration
associated with the variable will also be removed.  However, this could
cause types or constants to become unreferenced.  They will not be
removed.  Another pass will have to be called to remove those.

### DIFF
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -21,6 +21,7 @@ add_library(SPIRV-Tools-opt
   compact_ids_pass.h
   constants.h
   dead_branch_elim_pass.h
+        dead_variable_elimination.h
   decoration_manager.h
   def_use_manager.h
   eliminate_dead_constant_pass.h
@@ -66,6 +67,7 @@ add_library(SPIRV-Tools-opt
   decoration_manager.cpp
   def_use_manager.cpp
   dead_branch_elim_pass.cpp
+        dead_variable_elimination.cpp
   eliminate_dead_constant_pass.cpp
   flatten_decoration_pass.cpp
   fold.cpp
--- a/source/opt/dead_variable_elimination.cpp
+++ b/source/opt/dead_variable_elimination.cpp
@@ -1,0 +1,105 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "dead_variable_elimination.h"
+
+#include "reflect.h"
+
+namespace spvtools {
+namespace opt {
+
+Pass::Status DeadVariableElimination::Process(spvtools::ir::Module* module) {
+  bool modified = false;
+  module_ = module;
+  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+  FindNamedOrDecoratedIds();
+  decoration_manager_ = std::unique_ptr<analysis::DecorationManager>(
+      new analysis::DecorationManager(module));
+
+  std::vector<uint32_t> ids_to_remove;
+
+  // Get the reference count for all of the global OpVariable instructions.
+  for (auto& inst : module->types_values()) {
+    if (inst.opcode() == SpvOp::SpvOpVariable) {
+      size_t count = 0;
+      uint32_t result_id = inst.result_id();
+
+      // Check the linkage.  If it is exported, it could be reference somewhere
+      // else, so we must keep the variable around.
+      const ir::Instruction* linkage_instruction =
+          decoration_manager_->GetDecoration(result_id,
+                                             SpvDecorationLinkageAttributes);
+      if (linkage_instruction != NULL) {
+        uint32_t last_operand = linkage_instruction->NumOperands() - 1;
+        if (linkage_instruction->GetSingleWordOperand(last_operand) ==
+            SpvLinkageTypeExport) {
+          count = MUST_KEEP;
+        }
+      }
+
+      if (count != MUST_KEEP) {
+        // If we don't have to keep the instruction for other reasons, then look
+        // at the uses and count the number of real references.
+        if (analysis::UseList* uses = def_use_mgr_->GetUses(result_id)) {
+          count = std::count_if(
+              uses->begin(), uses->end(), [](const analysis::Use& u) {
+                return (!ir::IsAnnotationInst(u.inst->opcode()) &&
+                    u.inst->opcode() != SpvOpName);
+              });
+        }
+      }
+      reference_count_[result_id] = count;
+      if (count == 0) {
+        ids_to_remove.push_back(result_id);
+      }
+    }
+  }
+
+  // Remove all of the variables that have a reference count of 0.
+  if (!ids_to_remove.empty()) {
+    modified = true;
+    for (auto result_id : ids_to_remove) {
+      DeleteVariable(result_id);
+    }
+  }
+  return (modified ? Status::SuccessWithChange : Status::SuccessWithoutChange);
+}
+
+void DeadVariableElimination::DeleteVariable(uint32_t result_id) {
+  ir::Instruction* inst = def_use_mgr_->GetDef(result_id);
+  assert(inst->opcode() == SpvOpVariable &&
+      "Should not be trying to delete anything other than an OpVariable.");
+
+  // Look for an initializer that references another variable.  We need to know
+  // if that variable can be deleted after the reference is removed.
+  if (inst->NumOperands() == 4) {
+    ir::Instruction* initializer =
+        def_use_mgr_->GetDef(inst->GetSingleWordOperand(3));
+    if (initializer->opcode() == SpvOpVariable) {
+      uint32_t initializer_id = initializer->result_id();
+      size_t& count = reference_count_[initializer_id];
+      if (count != MUST_KEEP) {
+        --count;
+      }
+
+      if (count == 0) {
+        DeleteVariable(initializer_id);
+      }
+    }
+  }
+  this->KillNamesAndDecorates(result_id);
+  def_use_mgr_->KillDef(result_id);
+}
+}  // namespace opt
+}  // namespace spvtools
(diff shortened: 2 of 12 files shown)


source/opt/decoration_manager.h
```diff
@@ -50,6 +50,8 @@ class DecorationManager {
   bool AreDecorationsTheSame(const ir::Instruction* inst1,
                              const ir::Instruction* inst2) const;
 
+  const ir::Instruction* GetDecoration(uint32_t id, uint32_t decoration) const;
```
dneto0: Need a method comment.

source/opt/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ add_library(SPIRV-Tools-opt
   compact_ids_pass.h
   constants.h
   dead_branch_elim_pass.h
+        dead_variable_elimination.h
```
dneto0: spacing is wonky

source/opt/CMakeLists.txt
```diff
@@ -66,6 +67,7 @@ add_library(SPIRV-Tools-opt
   decoration_manager.cpp
   def_use_manager.cpp
   dead_branch_elim_pass.cpp
+        dead_variable_elimination.cpp
```
dneto0: spacing is wonky.
more importantly, needs a corresponding update to Android.mk

source/opt/decoration_manager.cpp
```diff
@@ -229,6 +229,35 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(uint32_t id,
   return decorations;
 }
 
+const ir::Instruction* DecorationManager::GetDecoration(
+    uint32_t id, uint32_t decoration) const {
+  auto decoration_list = id_to_decoration_insts_.find(id);
+  if (decoration_list != id_to_decoration_insts_.end()) {
```
dneto0: I'm slightly tempted to write this in terms of GetDecorationsFor.    But not that tempted.

No change requested.

source/opt/decoration_manager.h
dneto0: A decoration enum *could* applied more than one time to a target.  In fact the LinkageAttributes is such a case.
You can both import *and* export a variable, yes?

So Suggest this should either return a vector, or better do a ForEachDecorationForIdOfKind(uint32_t id, uint32_t, std::function<void(const Instruction*)>)

source/opt/dead_variable_elimination.h
```diff
@@ -0,0 +1,54 @@
...
+  // Note that the special value MUST_KEEP is used to indicate that the
+  // instruction cannot be deleted for reasons other that is being explicitly
+  // referenced.
+  std::unordered_map<uint32_t, size_t> reference_count_;
+
+  //  Decoration manager to help organize decorations.
+  std::unique_ptr<analysis::DecorationManager> decoration_manager_;
+
+  // Special value used to indicate that an id cannot be safely deleted.
+  const uint32_t MUST_KEEP = INT_MAX;
```
dneto0: Style guide says to spell this as  kMustKeep

Also, I tend to use this form instead, rather than allocating storage via a const member.
enum { kMustKeep =  INT_MAX };

source/opt/dead_variable_elimination.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+namespace spvtools {
+namespace opt {
+
+Pass::Status DeadVariableElimination::Process(spvtools::ir::Module* module) {
+  bool modified = false;
+  module_ = module;
+  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+  FindNamedOrDecoratedIds();
+  decoration_manager_ = std::unique_ptr<analysis::DecorationManager>(
```
dneto0: This could be made into a local variable.

source/opt/dead_variable_elimination.cpp
```diff
@@ -0,0 +1,105 @@
...
+  // Get the reference count for all of the global OpVariable instructions.
+  for (auto& inst : module->types_values()) {
+    if (inst.opcode() == SpvOp::SpvOpVariable) {
+      size_t count = 0;
+      uint32_t result_id = inst.result_id();
+
+      // Check the linkage.  If it is exported, it could be reference somewhere
+      // else, so we must keep the variable around.
+      const ir::Instruction* linkage_instruction =
+          decoration_manager_->GetDecoration(result_id,
```
dneto0: Ooooh.  A variable could both be imported *and* exported.  And depending on the order of those decorations, you'll miss the wrong one and incorrectly delete it.

So I think you really need a GetDecorationsForIdOfKind(result_id,   SpvDecoration)

source/opt/dead_variable_elimination.cpp
```diff
@@ -0,0 +1,105 @@
...
+  ir::Instruction* inst = def_use_mgr_->GetDef(result_id);
+  assert(inst->opcode() == SpvOpVariable &&
+      "Should not be trying to delete anything other than an OpVariable.");
+
+  // Look for an initializer that references another variable.  We need to know
+  // if that variable can be deleted after the reference is removed.
+  if (inst->NumOperands() == 4) {
+    ir::Instruction* initializer =
+        def_use_mgr_->GetDef(inst->GetSingleWordOperand(3));
+    if (initializer->opcode() == SpvOpVariable) {
```
dneto0: This is correct but too specific to be fully general.

E.g. OpenCL code can look like this:

```
__constant int c[3] = {0,1,2};
int __constant * constant cp = c +1;

kernel void foo(global int *A, int i ) {
  *A = 12;
}
```
The point is that cp's initializer is allowed to be an expression which is an OpSpecConstantOp that uses an PtrAccessChain as the opcode.  So you won't see far enough down into the expression tree to see that you could also delete the 'c' after you delete the 'cp'.

I recommend you add a TODO to do slightly better.
(And I think the better algorithm is to use a work-list strategy in the outer function.)

OR, maybe the better algorithm is mixed in with eliminating unused constants.  ??
You'd need the transitive closure all together working with both variables and constants.  Ugh.

source/opt/dead_variable_elimination.cpp
```diff
@@ -0,0 +1,105 @@
...
+                                             SpvDecorationLinkageAttributes);
+      if (linkage_instruction != NULL) {
+        uint32_t last_operand = linkage_instruction->NumOperands() - 1;
+        if (linkage_instruction->GetSingleWordOperand(last_operand) ==
+            SpvLinkageTypeExport) {
+          count = MUST_KEEP;
+        }
+      }
+
+      if (count != MUST_KEEP) {
```
dnovillo: Likewise here.

source/opt/dead_variable_elimination.cpp
```diff
@@ -0,0 +1,105 @@
...
+// limitations under the License.
+
+#include "dead_variable_elimination.h"
+
+#include "reflect.h"
+
+namespace spvtools {
+namespace opt {
+
+Pass::Status DeadVariableElimination::Process(spvtools::ir::Module* module) {
```
dnovillo: Some documentation explaining how this works?

source/opt/dead_variable_elimination.cpp
```diff
@@ -0,0 +1,105 @@
...
+  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+  FindNamedOrDecoratedIds();
+  decoration_manager_ = std::unique_ptr<analysis::DecorationManager>(
+      new analysis::DecorationManager(module));
+
+  std::vector<uint32_t> ids_to_remove;
+
+  // Get the reference count for all of the global OpVariable instructions.
+  for (auto& inst : module->types_values()) {
+    if (inst.opcode() == SpvOp::SpvOpVariable) {
```
dnovillo: I slightly prefer the form that leads to less indentation:

```
if (inst.opcode() != SpvOp::SpvOpVariable) {
   continue;
}
```

source/opt/decoration_manager.cpp
```diff
@@ -180,13 +181,12 @@ void DecorationManager::AnalyzeDecorations(ir::Module* module) {
 }
 
 template <typename T>
-std::vector<T> DecorationManager::InternalGetDecorationsFor(uint32_t id,
-                                                    bool include_linkage) {
```
dnovillo: I think this got mis-formatted.  Can you run clang-format on it?

source/opt/decoration_manager.h
dnovillo: Documentation?

source/opt/dead_variable_elimination.cpp
s-perron: done

source/opt/dead_variable_elimination.cpp
s-perron: done

source/opt/CMakeLists.txt
s-perron: done

source/opt/CMakeLists.txt
s-perron: fixed

source/opt/dead_variable_elimination.cpp
s-perron: This one merges back before the end of the loop.  I have to keep it like this.

source/opt/dead_variable_elimination.cpp
s-perron: I've added a to do.  To handle this, we will have to remove the OpSpecConstantOp, which my codes does not do.  I think we should make that the goal when fixing https://github.com/KhronosGroup/SPIRV-Tools/issues/906.  For now, I added a TODO.

source/opt/dead_variable_elimination.h
s-perron: done

test/opt/dead_variable_elim_test.cpp
```diff
@@ -0,0 +1,300 @@
+// Copyright (c) 2017 Valve Corporation
```
atgoo: Google?

source/opt/decoration_manager.h
s-perron: done.  I went with the ForEach method, but used a shorter name, and added a comment.

test/opt/dead_variable_elim_test.cpp
s-perron: right.

source/opt/dead_variable_elimination.cpp
s-perron: done

source/opt/CMakeLists.txt
dnovillo: Still seems over-indented.

source/opt/CMakeLists.txt
dnovillo: Seems over-indented.  Could it be just a rendering issue on the web client?

dneto0:
Waiting for bots to go green.

s-perron:
Will handle @dnovillo request in another pull request.

dneto0:
Rebased, squashed, and pushed into master as 5834719fc17

PR #907, opened 2017-10-20
Added dummy android test application

atgoo:
The application is needed to test Android.mk build.


dneto0:
Rebased and pushed into master as cfb711a68bb

PR #910, opened 2017-10-21
restrict opcodes targeting OpDecorationGroup

### LINKED ISSUE #616: Result id of OpDecorationGroup can only be used as the target of OpName, OpGroupDecorate, and OpGroupMemberDecorate
I believe that's the right rule.  If it's not true then the "flatten decorations" optimization pass is incorrect.
https://github.com/KhronosGroup/SPIRV-Tools/pull/610

jcaraban:
Fixes #616 

Submitting for feedback.

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -181,6 +181,27 @@ bool idUsage::isValid<SpvOpMemberDecorate>(const spv_instruction_t* inst,
 }
 
 template <>
+bool idUsage::isValid<SpvOpDecorationGroup>(const spv_instruction_t* inst,
+                                     const spv_opcode_desc) {
+  auto decorationGroupIndex = 1;
+  auto decorationGroup = module_.FindDef(inst->words[decorationGroupIndex]);
+
+  for (auto pair : decorationGroup->uses()) {
+    auto use = pair.first;
+    if (use->opcode() != SpvOpDecorate &&
+        use->opcode() != SpvOpGroupDecorate &&
+        use->opcode() != SpvOpGroupMemberDecorate &&
+        use->opcode() != SpvOpName ) {
+      DIAG(decorationGroupIndex) << "Result id of OpDecorationGroup can only "
+                                 << "be targeted by OpName, OpGroupDecorate, "
+                                 << "OpDecorate, and OpGroupMemberDecorate";
+      return false;
+    }
+  }                                        
+  return true;
+}
+
+template <>
 bool idUsage::isValid<SpvOpGroupDecorate>(const spv_instruction_t* inst,
                                           const spv_opcode_desc) {
   auto decorationGroupIndex = 1;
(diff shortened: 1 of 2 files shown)


source/validate_id.cpp
```diff
@@ -181,6 +181,27 @@ bool idUsage::isValid<SpvOpMemberDecorate>(const spv_instruction_t* inst,
 }
 
 template <>
+bool idUsage::isValid<SpvOpDecorationGroup>(const spv_instruction_t* inst,
+                                     const spv_opcode_desc) {
+  auto decorationGroupIndex = 1;
+  auto decorationGroup = module_.FindDef(inst->words[decorationGroupIndex]);
+
+  for (auto pair : decorationGroup->uses()) {
```
dneto0: Slightly prefer auto&

source/validate_id.cpp
```diff
@@ -181,6 +181,27 @@ bool idUsage::isValid<SpvOpMemberDecorate>(const spv_instruction_t* inst,
...
+    auto use = pair.first;
+    if (use->opcode() != SpvOpDecorate &&
+        use->opcode() != SpvOpGroupDecorate &&
+        use->opcode() != SpvOpGroupMemberDecorate &&
+        use->opcode() != SpvOpName ) {
+      DIAG(decorationGroupIndex) << "Result id of OpDecorationGroup can only "
```
dneto0: It would be more friendly to give more information so we can identify the problem in a big SPIR-V module:
- give the result ID of the OpDecorationGroup (so  you can easily find it)
- give the Opcode of the offending instruction.  Optional: If it has a result Id then give that too.

Generally, the validator's error messages are pretty poor.  I'd like to incrementally make them better.

dneto0:
FYI.  If you write "Fixes " and then the URL of the issue you're fixing, then GitHub will automatically close the issue when the commit appears on master.  That saves a step.
Just be sure to fix the entire issue.

dneto0:
Rebased and pushed into master as 13e659894

source/validate_id.cpp
```diff
@@ -181,6 +181,27 @@ bool idUsage::isValid<SpvOpMemberDecorate>(const spv_instruction_t* inst,
...
 
 template <>
+bool idUsage::isValid<SpvOpDecorationGroup>(const spv_instruction_t* inst,
+                                     const spv_opcode_desc) {
+  auto decorationGroupIndex = 1;
+  auto decorationGroup = module_.FindDef(inst->words[decorationGroupIndex]);
+
+  for (auto pair : decorationGroup->uses()) {
+    auto use = pair.first;
+    if (use->opcode() != SpvOpDecorate &&
```
dneto0: An alternative way to write this is:


 switch (use->opcode()) {
  case SpvOpDecorate:
  case SpvOpGroupDecorate:
  case SpvOpName:
      break;
  default:
      ... issue the error

 }

dneto0: I tend to like that better because it evaluates the opcode only once, and it's easier to scan the list of opcodes and the logic.

Issue #911, opened 2017-10-23
optimizer: OpLine and OpNoLine can't fall in debug1, debug2, or debug3 sections


dneto0:
Found in review of #908 by @antiagainst 

Should fix the definition of IsDebug1Inst and IsDebug2Inst functions

PR #912, opened 2017-10-23
Optimizer: Line and NoLine are not debug1 or debug2

dneto0:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/911


dneto0:
It's not clear to me there is a fruitful test.

I could write special tests that verify the IsDebug1Inst IsDebug2Inst functions directly.  That seems not very useful as it just replicates the reflect.h code.

I could write an ir_loader test to show that if I put an OpLine or OpNoLine instruction in the wrong place that it will end up in the right place.  That is, it can change an invalid module into a valid one.  But we don't want to force that responsibility forever. 

I'm open to other suggestions.

PR #913, opened 2017-10-23
Run dead variable elimination when using -O and -Os

s-perron:
We want to run the optimization when using -O and -Os, but it was not
added at part of https://github.com/KhronosGroup/SPIRV-Tools/pull/905.
This change will add that a well as some minor formatting changes
requested in that same pull request.


dneto0:
Rebased and pushed into master as 8d6e4dbc727

PR #915, opened 2017-10-24
Change the sections in the module to use the InstructionList class.

s-perron:
This change will replace a number of the
std::vector<std::unique_ptr<Instruction>> member of the module to
InstructionList.  This is for consistency and to make it easier to
delete instructions that are no longer needed.


dneto0:
Waiting for bots to pass

dneto0:
Rebased and pushed into master as 94dc66b74d5

Issue #893, opened 2017-10-19
Extend Phi placement to support unstructured control flow


dnovillo:
This is a pre-requisite for #889.  There exists an implementation for Phi placement in https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/local_ssa_elim_pass.cpp#L410.  However, that implementation is limited to structured control flow input.

To compute the minimal or pruned SSA form needed to insert Phi instructions, implement the algorithms referenced in https://en.wikipedia.org/wiki/Static_single_assignment_form.

This will require

- [ ] Implementation of a dominator tree.
- [ ] Implementation of dominance frontiers.

ratchetfreak:
There are other ways of placing phi nodes. 

Start from the function entry point and give each variable a id (which will be all of them). and add the basic block to the set O.

In every basic block except the first for each variable you get a new intermidiate ID and add a zero-source opPhi for linked to each variable.

Every store is replaced with a OpCopyObject with the same source and new ID, the new id is then linked to the variable for the remainder of the basic block.

Every load is replaced with a OpCopyObject using the Id of the opPhi at the beginning of the basic block or the previous opStore that was replaced with OpCopyObject 

For every basic block go to the next block and ad the final ID for each variable to the corresponding opPhi.

Then run a a DIE to remove and opPhis and OpCopyObjects whose results are not used and a copy propagation pass that also propagates each single source opPhis.

dnovillo:
> There are other ways of placing phi nodes.

Sure.  Any advantages over the traditional pruned or minimal SSA form?

ratchetfreak:
Biggest advantage is that it can be implemented as a series of simple passes each being able to be verified for correctness. (if you make a few changes like instead of opStore being replaced have the OpCopyObject being inserted next to it). Copy and single-source opPhi propagation being useful passes in and of themselves though.

Biggest disadvantage is that it is multiple passes that each depend on the input being of a specific form.

PR #916, opened 2017-10-24
ADCE: Dead if elimination

greg-lunarg:
This pass essentially eliminates "if" control flow when the "then" and "else"
parts are empty.

Mark structured conditional branches live only if one or more instructions
in their associated construct is marked live. After closure, replace dead
structured conditional branches with a branch to its merge and remove
dead blocks.

Besides eliminating a not-insignificant amount of dead code (up to %30)
from real shaders, this optimization is most necessary for legalization as there are
real life codes that are expecting the frontend/optimizer/spirv-generator to eliminate
references to objects that are referenced statically but not dynamically under
compile-time-constant conditions.

This code does its own dead block elimination, but will ultimately be replaced with
a call the newly integrated cfg cleanup code.


dneto0:
cfg-cleanup has already been merged.
I would prefer to not write more code to eliminate blocks.  This MR has net of more than 200 lines of code in the .cpp file.  Can you rewrite this to just update the branch at the top of the selection construct?

I haven't looked at the code or the cases, but I'm concerned about blindingly branching to the merge block.  If all arms of the selection construct do a "return" then the merge block was unreachable and should remain that way.

dneto0:
I guess my question is: Have you tried cutting this down to only mutating the branch at the top, and then running cfg-cleanup?  That would be preferred.

greg-lunarg:
The code to do the actual elimination of blocks is approximately 2% of the new code. I can refactor using cfg cleanup and retest. I was just trying to make this code available as soon as possible as there are Vulkan developers being held up waiting for it.

Are you ok with my proposal of pushing the heart of the cfg-cleanup code into MemPass and calling it as a utility from ADCE?

FYI, a lot of the code in the MR is just a replication of the "StructuredOrder" code from DeadBranchElim. Ultimately some form of this code will pushed into MemPass (or Pass) and shared, but right now I have two slightly different versions of this code both deriving off of MemPass so I need to resolve that. FWIW, the incompatibility is due to having to deal with unreachable blocks in the SSA code.

greg-lunarg:
FYI, I still need to add tests for this functionality. But I wanted to start the review of the non-test code as soon as possible.

dnovillo:
I think this should be part of DCE or at the very least part of the CFG cleanup code.  This is essentially straightening an if-then-else diamond with no body, right?  It's similar to straightening an if(true) or if(false) diamond (which we will leave when constant propagating more commonly).

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -123,40 +129,193 @@ void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
   live_local_vars_.insert(varId);
 }
 
+uint32_t AggressiveDCEPass::MergeBlockIdIfAny(
```
dneto0: There are copies of this already in InlinePass, LocalMultiStoreElimPass, DeadBranchElimPass, CommonUniformElimPass.  Let's not add a fifth copy.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -123,40 +129,193 @@ void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
...
+    }
+    else if (merge_ii->opcode() == SpvOpSelectionMerge) {
+      mbid = merge_ii->GetSingleWordInOperand(
+          kSelectionMergeMergeBlockIdInIdx);
+    }
+  }
+  return mbid;
+}
+
+void AggressiveDCEPass::ComputeStructuredSuccessors(ir::Function* func) {
```
dneto0: Similar, there are several copies of this elsewhere.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -123,40 +129,193 @@ void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
...
+        block2structured_succs_[&blk].push_back(id2block_[cbid]);
+    }
+    // add true successors
+    blk.ForEachSuccessorLabel([&blk, this](uint32_t sbid) {
+      block2structured_succs_[&blk].push_back(id2block_[sbid]);
+    });
+  }
+}
+
+void AggressiveDCEPass::ComputeStructuredOrder(
```
dneto0: Similar. Copied code.

greg-lunarg:
@dnovillo Please note that this is indeed part of DCE.

Please also note that this is not the straightening you mentioned. What you describe is already done in DeadBranchElimination.

This is when you have an "if" where the "then" and "else" part are both empty. This let's you replace the conditional branch with an unconditional branch (to the merge block), most importantly getting rid of the condition, which can be considerable in size and contain (useless) references to external objects.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -123,40 +129,193 @@ void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
...
   call_in_func_ = false;
   func_is_entry_point_ = false;
   private_stores_.clear();
-  for (auto& blk : *func) {
-    for (auto& inst : blk) {
-      uint32_t op = inst.opcode();
+  std::stack<bool> skip_control_flow;
+  std::stack<uint32_t> currentMergeBlockId;
+  skip_control_flow.push(false);
+  currentMergeBlockId.push(false);
```
dneto0: Did you mean to push 0, signifiying no merge block at all?  The false is misleading.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -123,40 +129,193 @@ void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
...
+            AddToWorklist(&*ii);
+        } break;
+        case SpvOpLoopMerge: {
+          skip_control_flow.push(false);
+          currentMergeBlockId.push(
+              ii->GetSingleWordInOperand(kSelectionMergeMergeBlockIdInIdx));
+          AddToWorklist(&*ii);
+        } break;
+        case SpvOpSelectionMerge: {
+          auto bri = ii;
```
dneto0: bri is branch_instr_iter?

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -123,40 +129,193 @@ void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
...
           // eg. GLSL frexp, modf
-          if (!IsCombinatorExt(&inst))
-            worklist_.push(&inst);
+          if (!IsCombinatorExt(&*ii))
+            AddToWorklist(&*ii);
+        } break;
+        case SpvOpLoopMerge: {
+          skip_control_flow.push(false);
+          currentMergeBlockId.push(
+              ii->GetSingleWordInOperand(kSelectionMergeMergeBlockIdInIdx));
```
dneto0: This is a LoopMerge, so the fact that kSelectionMergeMergeBlockIdInIdx works is accidental.

dnovillo:
On Tue, Oct 24, 2017 at 5:50 PM, greg-lunarg <notifications@github.com>
wrote:

> The code to do the actual elimination of blocks is approximately 2% of the
> new code. I can refactor using cfg cleanup and retest. I was just trying to
> make this code available as soon as possible as there are Vulkan developers
> being held up waiting for it.
>
> Are you ok with my proposal of pushing the heart of the cfg-cleanup code
> into MemPass and calling it as a utility from ADCE?
>

I'm OK either way.  cfg-cleanup will eventually need to be called from the
pass manager for passes that request it.  Typically, DCE, copy-prop,
const-prop, PRE... all these passes leave unreachable code, degenerate
Phis, squashable conditional jumps, etc.  I would not call it directly from
ADCE, but it's fine for now since we still don't have the ability to tell
ADCE to register its post-pass cleanup.

dnovillo:
On Tue, Oct 24, 2017 at 7:08 PM, greg-lunarg <notifications@github.com>
wrote:

> @dnovillo <https://github.com/dnovillo> Please note that this is indeed
> part of DCE.
>

But why is it a separate pass?  That's what confuses me a bit.  These look
like things that should be done by a single DCE pass.  We have several
instances of code that does DCE-things.  They should all be part of a
single pass that deals with all types of dead code.  Seems too fine-grained
atm.

greg-lunarg:
@dnovillo This is an "aggressive" DCE pass. It is a completely different algorithm than a "standard" DCE.

A "standard" DCE deletes a dead instruction and then looks to see what has been made dead because of the deletion. It is (arguably) a cheaper form of DCE but it cannot remove dead code cycles, like an induction value in a loop that is never used except by itself.

An "aggressive" DCE starts by assuming everything is dead. It then marks all statements that are alive (like a store to an output variable) and then marks all its operands live recursively. This is more expensive but it gets rid of dead code cycles. I already have a bug from a Vulkan development group that needs dead loop elimination to get rid of unused external references in a shader which are tripping the (Vulkan) validator. Fxc gets rid of these, so we are going to have to as well. ADCE is a way to get rid of these. I plan to start writing that code next.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Yes. Surprised I did that.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Yeah, I think that was a cut-and-paste error.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Yep.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Yes. Sorry. I am just trying to get this code out as quickly as possible. I know of two Vulkan development groups that are blocked waiting for this code.

The LocalMultiStoreElimPass version is different than the others (it includes all blocks, not just unreachable ones). That is what is holding up merging these all together. That and I still need to move CommonUniformElimPass under MemPass. I will do this under a separate PR.

dnovillo:
On Tue, Oct 24, 2017 at 7:45 PM, greg-lunarg <notifications@github.com>
wrote:

> @dnovillo <https://github.com/dnovillo> This is an "aggressive" DCE pass.
> It is a completely different algorithm than a "standard" DCE.
>
> A "standard" DCE deletes a dead instruction and then looks to see what has
> been made dead because of the deletion. It is (arguably) a cheaper form of
> DCE but it cannot remove dead code cycles, like an induction value in a
> loop that is never used except by itself.
>
> An "aggressive" DCE starts by assuming everything is dead. It then marks
> all statements that are alive (like a store to an output variable) and then
> marks all its operands live recursively. This is more expensive but it gets
> rid of dead code cycles. I already have a bug from a Vulkan development
> group that needs dead loop elimination to get rid of unused external
> references in a shader which are tripping the (Vulkan) validator. Fxc gets
> rid of these, so we are going to have to as well. ADCE is a way to get rid
> of these. I plan to start writing that code next.
>

Yes, that's the DCE I'm referring to.  I'm not sure what you are trying to
say here.  Sorry.

In particular, I'm interested in https://dl.acm.org/citation.cfm?id=178256

s-perron:
I'll reiterate what I said before, with some extra.  This is starting to merge two different optimizations: dead code elimination, and code straightening.  I think they should be two separate things for a few reasons:

1) Simpler code.
2) Other things create branches over empty blocks, so the code straightening needs to be done anyway.
3) Mixing it this way will not get everything.  For example, and if-then-else where else block is empty, should be turned into an if-then.  We will need code straightening anyway.

We should favor less and simpler code.

The argument against is that if we can fold the branch in DCE, then we can get rid of more in one shot.  There are two reasons I don't think this outweighs my reasons above.

1) We can get the same results by iterating.  DCE, straightening, DCE, etc.  In practice, you don't have to iterate very much to get rid of everything.
2) In a real optimization sequence, DCE and straightening are clean up passes that should be run a lot.  You end up getting lots of iterations anyway.

dnovillo:
On Wed, Oct 25, 2017 at 1:36 AM, Steven Perron <notifications@github.com>
wrote:

> I'll reiterate what I said before, with some extra. This is starting to
> merge two different optimizations: dead code elimination, and code
> straightening. I think they should be two separate things for a few reasons:
>
>    1. Simpler code.
>    2. Other things create branches over empty blocks, so the code
>    straightening needs to be done anyway.
>    3. Mixing it this way will not get everything. For example, and
>    if-then-else where else block is empty, should be turned into an if-then.
>    We will need code straightening anyway.
>
>
Agreed.  DCE and others should just leave unreachable code, empty control
flow and collapsable predicated branches alone.  After they run, we
schedule a cleanup pass to take care of everything.  This separation of
concerns will simplify code and reduce duplication.


>
>
> We should favor less and simpler code.
>
> The argument against is that if we can fold the branch in DCE, then we can
> get rid of more in one shot. There are two reasons I don't think this
> outweighs my reasons above.
>
>    1. We can get the same results by iterating. DCE, straightening, DCE,
>    etc. In practice, you don't have to iterate very much to get rid of
>    everything.
>    2. In a real optimization sequence, DCE and straightening are clean up
>    passes that should be run a lot. You end up getting lots of iterations
>    anyway.
>
>
Yup.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -123,40 +129,193 @@ void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
...
+  *branchInst = &*ii;
+  --ii;
+  if (ii->opcode() != SpvOpSelectionMerge)
+    return false;
+  *mergeInst = &*ii;
+  *mergeBlockId = ii->GetSingleWordInOperand(kSelectionMergeMergeBlockIdInIdx);
+  return true;
+}
+
+void AggressiveDCEPass::ComputeBlock2BranchMaps(
```
s-perron: This reminds me of the control dependence graph (CDG).  If we keep the code this way, it might be a good idea to move this analysis out into it own things.  Build the CDG, and then it can easily be reused by other transformations that need it.  It is not needed by many thing, but at least is will be available.  Plus it will make it easier to test on it own.

greg-lunarg:
Thanks for your continued concerns and ideas.

First, I don't think we need to talk about unreachable blocks anymore. I have agreed to use cfg-cleanup.

As for repeating DCE and a form of "straightening" that detects empty then-else blocks, there is a piece of the puzzle that I have not talked about: dead loop elimination (DLE).

First let me say that DLE and dead if elimination(DIE) are not just niceties. There are two Vulkan development groups that I know of that are blocked waiting for this. The lack of these is leaving references to objects that were deleted by fxc and are causing binding errors in Vulkan validation.

DLE is not as simple as DIE. Its not incredibly complex either, but it requires some analysis to determine if any values generated in a loop are used outside of the loop. So any separate pass doing this will need to construct data structures that determine control structures and their blocks, etc. Basically doing the same thing I am doing with DIE. So there is no savings by removing this from ADCE. We will need to do it somewhere.

I am going to be able to  easily extend DIE to hand DLE (I started it late yesterday and will have it done today). So I am going to be able to use the same framework to solve DCE, DIE and DLE, and solve them completely in just one pass without iteration.

You are right that this doesn't solve things like simplifying empty else blocks. That is a much simpler problem that should be part of the cfg-cleanup or merge-blocks or somewhere else.

You are also right in recognizing CDG. That is the underlying theory to my algorithm.

I am hoping that you will see my solution as an elegant and cost effective and timely way to extend the ADCE framework to also doing two critical optimizations, DIE and DLE.

source/opt/aggressive_dead_code_elim_pass.cpp
s-perron: A lot of these things should be moved into a class representing the CFG.

dnovillo:
On Thu, Oct 26, 2017 at 1:49 PM, Steven Perron <notifications@github.com>
wrote:

>
> In source/opt/aggressive_dead_code_elim_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/916#discussion_r147146637>
> :
>
> > +  *cbid = 0;
> +  if (merge_ii != blk.cbegin()) {
> +    --merge_ii;
> +    if (merge_ii->opcode() == SpvOpLoopMerge) {
> +      mbid = merge_ii->GetSingleWordInOperand(kLoopMergeMergeBlockIdInIdx);
> +      *cbid = merge_ii->GetSingleWordInOperand(kLoopMergeContinueBlockIdInIdx);
> +    }
> +    else if (merge_ii->opcode() == SpvOpSelectionMerge) {
> +      mbid = merge_ii->GetSingleWordInOperand(
> +          kSelectionMergeMergeBlockIdInIdx);
> +    }
> +  }
> +  return mbid;
> +}
> +
> +void AggressiveDCEPass::ComputeStructuredSuccessors(ir::Function* func) {
>
> A lot of these things should be moved into a class representing the CFG.
>

Yes.  I am currently doing step #1 of this (within
https://github.com/KhronosGroup/SPIRV-Tools/issues/893 - I'll be pushing
the new update later today).  I'm moving them into MemPass to first remove
the redundant copies.  I will then create a CFG class holding these.
Perhaps the code in source/cfa.h could be a good spot for this
functionality.

source/opt/aggressive_dead_code_elim_pass.cpp
s-perron: I think we should try to be careful about what we assume.  If we are doing something that only works because we have structured control flow, we should add asserts and comments indicating that.  They should be close to the part of the code that uses that assumption.  We don't want the OpenCL world accidentally use this code.  It also makes it obvious which parts need to be updated if we want this optimization to work for things other than Vulkan shaders.

Also because the concept of control dependence is generally known, it might be a good idea to talk about it in those terms.  When I see a function name ComputeBlock2BranchMaps, I have no idea how they are related.  Are they the branches at the end of said block?  Even if there are comments in the header, I do not want to go to the header every time I see a function call.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -218,12 +384,20 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
...
+    for (auto ii = (*bi)->begin(); ii != (*bi)->end(); ++ii) {
+      if (IsLive(&*ii))
         continue;
-      dead_insts_.insert(&inst);
+      if (IsBranch(ii->opcode())) {
+        if (ii == (*bi)->begin())
+          continue;
+        auto mi = ii;
+        --mi;
+        if (mi->opcode() != SpvOpSelectionMerge)
```
s-perron: Is there a reason this will not work (or we don't want it to work) with branches that are not marked as OpSelectionMerge?  That reason should be documented as a comment.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -241,15 +415,41 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
...
+        KillAllInsts(*bi);
+        dead_blocks.insert(*bi);
+      }
+    }
+    else {
+      ++bi;
+    }
   }
+  // Erase dead blocks from function
+  for (auto ebi = func->begin(); ebi != func->end(); )
```
s-perron: Formatting seems off for this loop.

source/opt/aggressive_dead_code_elim_pass.cpp
dneto0: > I think we should try to be careful about what we assume. If we are doing something that only works because we have structured control flow, we should add asserts and comments indicating that.

We've been handling this with the first guard (and early return) at the top of ProcessImpl.  It checks for Shader, which then requires structured control flow.  The assumption that structured control flow is required is somewhat pervasive, and I wouldn't want to pepper all the existing code with repeat of the same assert. It's something that all of us working in this code are now accustomed to.  

Perhaps it merits a comment, but I think the fact that we're using a structuredOrder parameter is a very very strong signal.  Also, the code is peppered with ....MergBlockId...  which again is a very strong signal.

Given all that I'm inclined to say we don't need yet another signal boost that this requires/assumes structured control flow.

dneto0: > Even if there are comments in the header, I do not want to go to the header every time I see a function call.

I've given pretty strong guidance to follow the Google C++ style guide on this topic.  See 
https://google.github.io/styleguide/cppguide.html#Function_Comments

In particular:
* The method comment is given in the header.  Don't repeat it at the function body because it's liable to get out of date.
* The method comment describes what the function does from the client's perspective
* Comments about the implementation details (if it's non-obvious) are placed inside the function body.

How does this apply in this instance?   The PR is good, in my view, regarding placement of the comments.  I think it could use more explanatory text about the relation of the two block2merge_ and block2branch_ to the more general CDG concept.  I would put that in one place, probably at the declaration of those members.  Another reasonable spot is an explanatory paragraph before the declaration of this method.

dneto0: Saying it another way, I'm strongly inclined to document the *data*, and then the methods are generally self-explanatory.  (I think Fred Brooks said something about this... though I think it was framed in terms of "tables")

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -218,12 +384,20 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
...
-  for (auto& blk : *func) {
-    for (auto& inst : blk) {
-      if (live_insts_.find(&inst) != live_insts_.end())
+  // Mark all non-live instructions dead, except non-if branches
+  for (auto bi = structuredOrder.begin(); bi != structuredOrder.end(); ++bi) {
+    for (auto ii = (*bi)->begin(); ii != (*bi)->end(); ++ii) {
+      if (IsLive(&*ii))
         continue;
-      dead_insts_.insert(&inst);
+      if (IsBranch(ii->opcode())) {
```
dneto0: I was tempted to suggest use of IsStructuredIfHeader here.  Is this any better?

```
if (IsBranch(ii->opcode()) && IsStructuredIfHeader(*bi, nullptr, nullptr, nullptr)) continue;
```
But that would also need a tweak to IsStructuredHeader to allow null for those last three args.  You just want the bool result.

I don't know if that's worth it though.  Your choice.

source/opt/aggressive_dead_code_elim_pass.cpp
s-perron: Sorry, my suggestion was not about the comments, but the name of the function and data.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -178,24 +337,31 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
...
     liveInst->ForEachInId([this](const uint32_t* iid) {
       ir::Instruction* inInst = def_use_mgr_->GetDef(*iid);
-      if (live_insts_.find(inInst) == live_insts_.end())
-        worklist_.push(inInst);
+      if (!IsLive(inInst))
+        AddToWorklist(inInst);
     });
+    // If in a structured if construct, add the controlling conditional branch
+    // and its merge
+    ir::BasicBlock* blk = inst2block_[liveInst];
```
dneto0: I'm surprised this didn't walk up the stack of nested structured-ifs and mark all those ancestors live too.

What happens if there is a doubly-nested structured if,  and there is a live instruction inside the inner-if.  It seem the inner if branch (and merge) are marked live, but the outer one will still remain marked dead, and we'll accidentally delete the entire thing.
E.g.
```
  if (true) { // outer
     if (true) { // inner
        foo = 1; // something live.   the inst2block_ -> block2branch_ mapping will point to inner, but not outer
     }
  }
```
But then later the outer will be seen as dead and the whole thing swept away?

dneto0: Thinking about this more, this is handled because then the branch instruction at inner goes onto the work list and then we'll naturally end up walking up the stack.

Wow, this deserves a comment and a test case.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -241,15 +415,41 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
...
+        mergeBlockId = 
+            ii->GetSingleWordInOperand(kSelectionMergeMergeBlockIdInIdx);
+      def_use_mgr_->KillInst(&*ii);
       modified = true;
     }
+    // If a structured if was deleted, add a branch to its merge block,
+    // and traverse to the merge block killing all instructions on the way
+    // and remembering dead blocks for later deletion from function.
+    if (mergeBlockId != 0) {
+      AddBranch(mergeBlockId, *bi);
```
dneto0: How are we guaranteed that the block still exists at all?   I think this is actually ok because the block's instruction iteration skips over the OpLabel at the start. But it's worth a comment.

Is there a test case where the entire function is dead?  That would be a nice illustration that this aspect works.

test/opt/aggressive_dead_code_elim_test.cpp
```diff
@@ -1388,6 +1388,271 @@ OpFunctionEnd
...
+%26 = OpCompositeConstruct %v4float %25 %25 %25 %25
+OpStore %OutColor %26
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::AggressiveDCEPass>(
+      assembly, assembly, true, true);
+}
+
```
dneto0: Please add a doubly-nested if case where the outer one is live only because the inner one is live:

Analogous to:
```
  if (true) { // outer
     if (true) { // inner
        foo = 1; // something live.   the inst2block_ -> block2branch_ mapping will point to inner, but not outer
     }
  }
```

Please add a test case where the whole function goes dead because the nested-if is dead.
```
void foo(bool c) {
   if (c) {
      uint x = 2; // dead
   }
}
```
We should be left with a single block having only OpLabel and OpReturn

source/opt/aggressive_dead_code_elim_pass.cpp
dneto0: Specifically, OpSwitch is the other case.   Totally fine to not add it now since that's way more testing burden.  Agree it would be nice to have a comment saying you've deliberately scoped it down to just if-selection

test/opt/aggressive_dead_code_elim_test.cpp
```diff
@@ -1388,6 +1388,271 @@ OpFunctionEnd
       assembly, assembly, true, true);
 }
 
+TEST_F(AggressiveDCETest, EliminateDeadIf) {
```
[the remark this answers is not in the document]
greg-lunarg: I have added a test for IfThen and also for nested constructs, both eliminated and not.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: I agree the names are too terse. I will change the name of this method to ComputeBlock2HeaderMaps and the name of the structures to block2headerMerge and block2headerBranch.

greg-lunarg:
@dneto0 Yes, sorry, did not mean to imply with that update that I was done. I usually push my updates at the end of the day to reduce risk of lost work.

I will certainly address all the comments made and will push that work shortly. Thanks!

dneto0:
> @dneto0 Yes, sorry, did not mean to imply with that update that I was done. I usually push my updates at the end of the day to reduce risk of lost work. 

I didn't mean to imply it was a problem. I did a quick review and I wanted to checkpoint my own review.  I planned to use my list as my review TODO.

In other news #921 was just pushed into master.  It's a big refactor that will affect ADCE.  It's functionally orthogonal to your change, but you'll have to rebase against it first.

greg-lunarg:
I think I have addressed all comments except the duplicate StructuredOrder functions. I will see to the merging and that next.

greg-lunarg:
I have removed the duplicate StructuredOrder code and am now using CFGCleanup to remove any dead blocks. I think I have now addressed all comments to this point.

dnovillo:
Thanks for putting up with the rug-pulling, Greg!  I've got more, but I'm waiting for this and #934 to go in first.

dneto0:
I'm rebasing.

dneto0:
Rebased, squashed, and pushed into master as 94bec26afe1

@dnovillo  There might be some stray duplications remaining in the added code.  I made it compile and work, but didn't check for a 100% code de-dup.

And thanks again @greg-lunarg !

dnovillo:
> @dnovillo There might be some stray duplications remaining in the added code. I made it 
> compile and work, but didn't check for a 100% code de-dup.

No worries.  I still have other re-factorings to do.  I'll take care of any remaining duplicates.

Issue #728, opened 2017-07-20
Opt: Inlining needs to preserve decorations


greg-lunarg:
Currently the inliner is not generating decorations for inlined code which corresponds to function code which has decorations. An example of decorations that are relevant: RelaxedPrecision, NoContraction.

dnovillo:
Fixed.

PR #919, opened 2017-10-25
Fix: Preserve decorations of inlined functions.

daniel-schuermann:
Fixes issue #728.
(More test cases e.g. with group decorations are desirable.)
+small bugfix in ForEachDecoration()

### DIFF
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -261,49 +261,51 @@ void DecorationManager::CloneDecorations(uint32_t from, uint32_t to) {
   if (decoration_list == id_to_decoration_insts_.end()) return;
   for (ir::Instruction* inst : decoration_list->second) {
     switch (inst->opcode()) {
-      case SpvOpGroupDecorate:
-        // add |to| to list of decorated id's
-        inst->AddOperand(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
-        id_to_decoration_insts_[to].push_back(inst);
+      case SpvOpNop: break;
+      case SpvOpGroupDecorate: {
+        // replace decoration instruction with one containing the new id as well.
+        std::vector<ir::Operand> new_operands;
+        new_operands.insert(inst->begin(), inst->begin(), inst->end());
+        new_operands.push_back(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
+        std::unique_ptr<ir::Instruction> new_inst(new ir::Instruction(inst->opcode(), inst->type_id(), inst->result_id(), new_operands));
+        id_to_decoration_insts_[from].push_back(new_inst.get());
+        module_->AddAnnotationInst(std::move(new_inst));
+        inst->ToNop();
         break;
+      }
       case SpvOpGroupMemberDecorate: {
-        bool is_id = false; // we start with decoration group id, followed by target-id, literal, target-id...
-        bool to_add = false;
-        // for each (id == from), get literal and add (to, literal) as operands
-        inst->ForEachInOperand([&](uint32_t* operand) {
-          if (is_id) {
-            if (*operand == from) {
-              to_add = true;
-              inst->AddOperand(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
-            }
-            is_id = false;
-          } else { // is literal
-            if (to_add) {
-              to_add = false;
-              inst->AddOperand(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_LITERAL_INTEGER, {*operand}));
-            }
-            is_id = true;
+        // replace decoration with one which contains additional the same pairs of (id, literal) for the new id:
+        std::vector<ir::Operand> new_operands;
+        new_operands.insert(inst->begin(), inst->begin(), inst->end());
+        for (auto op = ++inst->begin(); op != inst->end(); ++op) { // emit decoration group id
+          if (op->words[0] == from) { // add new pair of operands: (to, literal)
+            new_operands.push_back(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
+            ++op;
+            new_operands.push_back(*op);
+          } else {
+            ++op;
           }
-        });
-        assert(is_id && "mal-formed SpvOpGroupMemberDecorate encountered.");
-        id_to_decoration_insts_[to].push_back(inst);
+        }
+        std::unique_ptr<ir::Instruction> new_inst(new ir::Instruction(inst->opcode(), inst->type_id(), inst->result_id(), new_operands));
+        id_to_decoration_insts_[from].push_back(new_inst.get());
+        module_->AddAnnotationInst(std::move(new_inst));
+        inst->ToNop();
         break;
       }
       case SpvOpDecorate:
       case SpvOpMemberDecorate:
       case SpvOpDecorateId: {
         // simply clone decoration and change |target-id| to |to|
-        std::unique_ptr<ir::Instruction> newInst(inst->Clone());
-        newInst->SetInOperand(0, {to});
-        id_to_decoration_insts_[to].push_back(newInst.get());
-        module_->AddAnnotationInst(std::move(newInst));
+        std::unique_ptr<ir::Instruction> new_inst(inst->Clone());
+        new_inst->SetInOperand(0, {to});
+        id_to_decoration_insts_[to].push_back(new_inst.get());
+        module_->AddAnnotationInst(std::move(new_inst));
         break;
       }
       default:
         assert(false && "Unexpected decoration instruction");
     }
   }
-
 }
 
 }  // namespace analysis
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -235,16 +235,12 @@ void DecorationManager::ForEachDecoration(uint32_t id,
                                           std::function<void(const ir::Instruction&)> f) {
   for (const ir::Instruction* inst : GetDecorationsFor(id, true)) {
     switch (inst->opcode()) {
-      case SpvOpDecorate:
-        if (inst->GetSingleWordInOperand(1) == decoration) {
-          f(*inst);
-        }
-        break;
       case SpvOpMemberDecorate:
         if (inst->GetSingleWordInOperand(2) == decoration) {
           f(*inst);
         }
         break;
+      case SpvOpDecorate:
       case SpvOpDecorateId:
         if (inst->GetSingleWordInOperand(1) == decoration) {
           f(*inst);
@@ -260,38 +256,26 @@ void DecorationManager::CloneDecorations(uint32_t from, uint32_t to) {
   auto const decoration_list = id_to_decoration_insts_.find(from);
   if (decoration_list == id_to_decoration_insts_.end()) return;
   for (ir::Instruction* inst : decoration_list->second) {
+    // TODO: update def-use manager (additional use of |to| )
     switch (inst->opcode()) {
-      case SpvOpNop: break;
-      case SpvOpGroupDecorate: {
-        // replace decoration instruction with one containing the new id as well.
-        std::vector<ir::Operand> new_operands;
-        new_operands.insert(inst->begin(), inst->begin(), inst->end());
-        new_operands.push_back(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
-        std::unique_ptr<ir::Instruction> new_inst(new ir::Instruction(inst->opcode(), inst->type_id(), inst->result_id(), new_operands));
-        id_to_decoration_insts_[from].push_back(new_inst.get());
-        module_->AddAnnotationInst(std::move(new_inst));
-        inst->ToNop();
+      case SpvOpGroupDecorate:
+        // add |to| to list of decorated id's
+        inst->AddOperand(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
+        id_to_decoration_insts_[to].push_back(inst);
         break;
-      }
-      case SpvOpGroupMemberDecorate: {
-        // replace decoration with one which contains additional the same pairs of (id, literal) for the new id:
-        std::vector<ir::Operand> new_operands;
-        new_operands.insert(inst->begin(), inst->begin(), inst->end());
+      case SpvOpGroupMemberDecorate:
+        // for each (id == from), add (to, literal) as operands
         for (auto op = ++inst->begin(); op != inst->end(); ++op) { // emit decoration group id
           if (op->words[0] == from) { // add new pair of operands: (to, literal)
-            new_operands.push_back(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
+            inst->AddOperand(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
             ++op;
-            new_operands.push_back(*op);
+            inst->AddOperand(*op);
           } else {
             ++op;
           }
         }
-        std::unique_ptr<ir::Instruction> new_inst(new ir::Instruction(inst->opcode(), inst->type_id(), inst->result_id(), new_operands));
-        id_to_decoration_insts_[from].push_back(new_inst.get());
-        module_->AddAnnotationInst(std::move(new_inst));
-        inst->ToNop();
+        id_to_decoration_insts_[to].push_back(inst);
         break;
-      }
       case SpvOpDecorate:
       case SpvOpMemberDecorate:
       case SpvOpDecorateId: {
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -230,33 +230,60 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
   return decorations;
 }
 
-void DecorationManager::ForEachDecoration(
-    uint32_t id, uint32_t decoration,
-    std::function<void(const ir::Instruction&)> f) const {
-  auto decoration_list = id_to_decoration_insts_.find(id);
-  if (decoration_list != id_to_decoration_insts_.end()) {
-    for (const ir::Instruction* inst : decoration_list->second) {
-      switch (inst->opcode()) {
-        case SpvOpDecorate:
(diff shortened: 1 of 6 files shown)


source/opt/decoration_manager.cpp
```diff
@@ -231,31 +231,78 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+      case SpvOpGroupDecorate:
+        // add |to| to list of decorated id's
+        inst->AddOperand(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
+        id_to_decoration_insts_[to].push_back(inst);
+        break;
+      case SpvOpGroupMemberDecorate: {
+        bool is_id = false; // we start with decoration group id, followed by target-id, literal, target-id...
+        bool to_add = false;
+        // for each (id == from), get literal and add (to, literal) as operands
+        inst->ForEachInOperand([&](uint32_t* operand) {
```
s-perron: I don't know what other people think, but my preference is to have a loop where 1 iteration of the loop treat the pair instead of using a boolean that keeps flipping and the use of to_add.  I would suggest:
```
for( int op = <first operand>; op < inst.NumOperands(); op += 2 ) {
  if ( <id is "from"> ) {
    // Add pair of operands <to, next literal>.
  }
}
```
To me this is easier to read.  @dneto0, do you have a preference for this type of code?

source/opt/decoration_manager.cpp
```diff
@@ -231,31 +231,78 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+          if (is_id) {
+            if (*operand == from) {
+              to_add = true;
+              inst->AddOperand(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
```
s-perron: I'll need to think if this is a good way of changing an instruction.  We need to think about keeping analysis up to date.  The DefUseManager in particular.  If there is one, how will it know to keep that up to date?  In this case (inlining), it might not be a big deal, but, in general, it will be.

source/opt/decoration_manager.cpp
```diff
@@ -231,31 +231,78 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+        break;
       }
+      case SpvOpDecorate:
+      case SpvOpMemberDecorate:
+      case SpvOpDecorateId: {
+        // simply clone decoration and change |target-id| to |to|
+        std::unique_ptr<ir::Instruction> newInst(inst->Clone());
+        newInst->SetInOperand(0, {to});
+        id_to_decoration_insts_[to].push_back(newInst.get());
+        module_->AddAnnotationInst(std::move(newInst));
```
s-perron: I prefer this idiom for adding instruction.  When we add an IRContext to the module, the module can take care of updating the analysis (including the decoration manager).

source/opt/decoration_manager.h
```diff
@@ -54,7 +57,13 @@ class DecorationManager {
   // |f| is run on each decoration instruction for |id| with decoration
   // |decoration|.
   void ForEachDecoration(uint32_t id, uint32_t decoration,
-                         std::function<void(const ir::Instruction& f)>) const;
+                         std::function<void(const ir::Instruction& f)>);
+
+  // Clone all decorations from one id |from|.
+  // The cloned decorations are assigned to the given id |to| and are
+  // added to the module. The purpose is to decorate cloned instructions.
+  // This function does not check, if the id |to| is already decorated.
```
s-perron: No comma.

source/opt/decoration_manager.cpp
```diff
@@ -231,31 +231,78 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+      default:
+        assert(false && "Unexpected decoration instruction");
+    }
+  }
+}
+
+void DecorationManager::CloneDecorations(uint32_t from, uint32_t to) {
+  auto const decoration_list = id_to_decoration_insts_.find(from);
+  if (decoration_list == id_to_decoration_insts_.end()) return;
+  for (ir::Instruction* inst : decoration_list->second) {
```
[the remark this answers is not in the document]
daniel-schuermann: I just saw that it contains a potential bug. id_to_decoration_insts_ also contains group decorations which are not handled but raise an assertion.

source/opt/decoration_manager.cpp
daniel-schuermann: So, in the other cases we better add a new instruction and tonop() the old one?

source/opt/decoration_manager.cpp
daniel-schuermann: I am fine with both but would have preferred paired operands each of two words ;)

source/opt/decoration_manager.cpp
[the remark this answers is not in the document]
s-perron: In this case, you want to rewrite the OpGroupDecorate and OpGroupMemberDecorate, which GetDecorations for will remove.  Makes sense.

source/opt/decoration_manager.cpp
s-perron: Yes, I think that is a good way of doing it for now.

source/opt/decoration_manager.cpp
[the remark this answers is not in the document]
daniel-schuermann: CloneDecorations() and ForEachDecoration() are completely independent. I changed ForEachDecoration, because it had unhandled cases. I think, the intention was to have a function which goes through all (direct and indirect) specified Op(Member)Decorations and not group decorations. So, I fixed it.

source/opt/decoration_manager.cpp
```diff
@@ -231,29 +231,78 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
-        case SpvOpMemberDecorate:
-          if (inst->GetSingleWordInOperand(2) == decoration) {
-            f(*inst);
-          }
-          break;
-        case SpvOpDecorateId:
-          if (inst->GetSingleWordInOperand(1) == decoration) {
-            f(*inst);
+                                          std::function<void(const ir::Instruction&)> f) {
+  for (const ir::Instruction* inst : GetDecorationsFor(id, true)) {
```
dneto0: I see why you did this.  You wanted to expose the effective decorations.  The group decorations were getting in the way.

Please clarify the method comment for this method. (i.e. in the header) to indicate this intent.

It's not your problem, but it's unfortunate that GetDecorationsFor allocates a new vector that we then throw away.  I suspect a good future enhancement is to have a ForEachDecoration overload that doesn't constrain the decoration kind.  This method could be written in terms of that more general one.

source/opt/decoration_manager.cpp
```diff
@@ -231,29 +231,78 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+  for (ir::Instruction* inst : decoration_list->second) {
+    switch (inst->opcode()) {
+      case SpvOpNop: break;
+      case SpvOpGroupDecorate: {
+        // replace decoration instruction with one containing the new id as well.
+        std::vector<ir::Operand> new_operands;
+        new_operands.insert(inst->begin(), inst->begin(), inst->end());
+        new_operands.push_back(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
+        std::unique_ptr<ir::Instruction> new_inst(new ir::Instruction(inst->opcode(), inst->type_id(), inst->result_id(), new_operands));
+        id_to_decoration_insts_[from].push_back(new_inst.get());
```
dneto0: I expected `to` as the map index.  ??

source/opt/decoration_manager.cpp
```diff
@@ -231,29 +231,78 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+            new_operands.push_back(*op);
+          } else {
+            ++op;
           }
-          break;
-        default:
-          assert(false && "Unexpected decoration instruction");
+        }
+        std::unique_ptr<ir::Instruction> new_inst(new ir::Instruction(inst->opcode(), inst->type_id(), inst->result_id(), new_operands));
+        id_to_decoration_insts_[from].push_back(new_inst.get());
```
dneto0: I expected `to` as the map index.

source/opt/decoration_manager.cpp
```diff
@@ -231,29 +231,78 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+    switch (inst->opcode()) {
+      case SpvOpNop: break;
+      case SpvOpGroupDecorate: {
+        // replace decoration instruction with one containing the new id as well.
+        std::vector<ir::Operand> new_operands;
+        new_operands.insert(inst->begin(), inst->begin(), inst->end());
+        new_operands.push_back(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
+        std::unique_ptr<ir::Instruction> new_inst(new ir::Instruction(inst->opcode(), inst->type_id(), inst->result_id(), new_operands));
+        id_to_decoration_insts_[from].push_back(new_inst.get());
+        module_->AddAnnotationInst(std::move(new_inst));
```
dneto0: Unfortunately the DefUseManagers that might exist are guaranteed to be out of date.
There is no great solution for this (we don't have an IRContext that knows all).  The only thing I can think of is to use a callback.

source/opt/decoration_manager.cpp
dneto0: Also, for all IDs x in the in-operand-list of inst, id_to_decoration_insts_[x] incorrectly contains &inst.  That stale reference must be removed.

source/opt/inline_pass.cpp
```diff
@@ -147,6 +147,7 @@ void InlinePass::CloneAndMapLocals(
   while (callee_var_itr->opcode() == SpvOp::SpvOpVariable) {
     std::unique_ptr<ir::Instruction> var_inst(callee_var_itr->Clone());
     uint32_t newId = TakeNextId();
+    dec_mgr_->CloneDecorations(callee_var_itr->result_id(), newId);
```
dneto0: the def-use-manager is out of date.

source/opt/decoration_manager.cpp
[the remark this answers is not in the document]
daniel-schuermann: I see. Shall I put it in a second PR?

source/opt/decoration_manager.cpp
daniel-schuermann: That's the question. Shouldn't this (and updating dec_mgr) be done in module?

source/opt/decoration_manager.h
```diff
@@ -54,7 +57,13 @@ class DecorationManager {
   // |f| is run on each decoration instruction for |id| with decoration
   // |decoration|.
   void ForEachDecoration(uint32_t id, uint32_t decoration,
-                         std::function<void(const ir::Instruction& f)>) const;
+                         std::function<void(const ir::Instruction& f)>);
```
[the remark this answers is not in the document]
daniel-schuermann: calling non-const function...

source/opt/decoration_manager.cpp
daniel-schuermann: You are right, it's both, because the old instruction is invalidated (or not, see below).

source/opt/instruction.h
```diff
@@ -163,10 +163,6 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   }
   // Gets the |index|-th logical operand.
   inline const Operand& GetOperand(uint32_t index) const;
-  // Adds |operand| to the list of operands of this instruction.
-  // It is the responsibility of the caller to make sure
-  // that the instruction remains valid.
-  inline void AddOperand(Operand operand);
```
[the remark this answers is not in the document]
daniel-schuermann: Okay, I also prefer this way, this would also keep the def-use-mgr correct for groupDecorations.

source/opt/decoration_manager.cpp
daniel-schuermann: Yes...
Had it right, before, with AddOperand(). If this function is okay for you, I'll do it by adding the operands and def_use_mgr as well as dec_mgr remain up to date.

daniel-schuermann:
I tried to incorporate all requested changes. Remains only the def-use manager outdated. This shouldn't be a problem for decorations at all and less for the inline pass. However, if you think, this needs a temporary solution, I can add a call-back.

s-perron:
I talked with David Neto this morning.  Yes you should add a callback.  The pattern he was considering was passing a function  f(Instruction* inst, boolean add) to CloneDecorations.

Then before changing an instruction "inst", you call "f(inst, false)" to notify whatever need to know that this instruction is going to be changed.  After changing it call "f(inst,true)" to say you are done updating it.  If you are just creating a new instruction, you don't need the first call.  If you are deleting an instruction you don't need the second call.

On the callee side, you can create a lambda or function that will inform the def-use manager that is should remove or add information about the instruction.

@dneto0 Please correct me if I have not correctly described your idea.  Daniel, if you have other idea, feel free to share.

FYI, https://github.com/KhronosGroup/SPIRV-Tools/issues/608 will be used to try to find a better way of keeping the def-use manager up to date.

dneto0:
@atgoo is also reviewing.

source/opt/decoration_manager.cpp
```diff
@@ -231,29 +231,62 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+      case SpvOpGroupDecorate:
+        // add |to| to list of decorated id's
+        inst->AddOperand(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
+        id_to_decoration_insts_[to].push_back(inst);
+        break;
+      case SpvOpGroupMemberDecorate:
+        // for each (id == from), add (to, literal) as operands
+        for (auto op = ++inst->begin(); op != inst->end(); ++op) { // emit decoration group id
+          if (op->words[0] == from) { // add new pair of operands: (to, literal)
+            inst->AddOperand(ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to}));
```
atgoo: This looks like a bug. std::vector::push_back may or may not invalidate |op| depending on if the vector has enough capacity.

daniel-schuermann: You mean if the allocation fails?

atgoo: The problem is that you are modifying a vector which you are iterating over. The C++ standard doesn't guarantee that you can do that.
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf
§ 23.3.7.5 vector modifiers
...
void push_back(const T& x);
void push_back(T&& x);
Remarks: Causes reallocation if the new size is greater than the old capacity. If no reallocation happens, all the iterators and references before the insertion point remain valid.
...

The standard guarantees that iterators remain valid only if vector has enough capacity. So there is a chance that after AddOperand |op| is not a valid iterator anymore.

daniel-schuermann: Thanks for clarifying!
So, I better create a new vector and append afterwards? (Already changed operand to &&.)

atgoo: I think you can still the same process if you use the vector indirectly with functions
GetNumOperands() and GetOperand(). But appending afterwards also works.

daniel-schuermann:
Can I rebase?

source/opt/decoration_manager.cpp
```diff
@@ -231,29 +231,68 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+        // add |to| to list of decorated id's
+        inst->AddOperand(std::move(
+          ir::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_ID, {to})));
+        id_to_decoration_insts_[to].push_back(inst);
+        f(*inst, true);
+        break;
+      case SpvOpGroupMemberDecorate:
+        f(*inst, false);
+        // for each (id == from), add (to, literal) as operands
+        for (uint32_t i = 1; i < inst->NumOperands(); i+=2) {
```
atgoo: Consider
const uint32_t num_operand =  inst->NumOperands();
for (uint32_t i = 1; i < num_operands; i+=2) {
since you don't need to iterate over operands you are adding.

source/opt/inline_pass.h
```diff
@@ -181,6 +182,12 @@ class InlinePass : public Pass {
   // Def/Use database
   std::unique_ptr<analysis::DefUseManager> def_use_mgr_;
 
+  // Update the DefUseManager when cloning decorations.
+  std::function<void(ir::Instruction&, bool)> UpdateDefUseMgr;
```
atgoo: It is a class variable, so I think correct naming should be update_def_use_mgr_.
Maybe callback_def_use_mgr_ for more clarity.

source/opt/decoration_manager.cpp
```diff
@@ -231,29 +231,68 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
-        case SpvOpMemberDecorate:
-          if (inst->GetSingleWordInOperand(2) == decoration) {
-            f(*inst);
-          }
-          break;
-        case SpvOpDecorateId:
-          if (inst->GetSingleWordInOperand(1) == decoration) {
-            f(*inst);
+                                          std::function<void(const ir::Instruction&)> f) {
+  for (const ir::Instruction* inst : GetDecorationsFor(id, true)) {
```
atgoo: Consider
assert(f);

source/opt/inline_pass.h
atgoo: Ah, it's not a callback.

source/opt/decoration_manager.h
```diff
@@ -30,7 +30,10 @@ namespace analysis {
 class DecorationManager {
  public:
   // Constructs a decoration manager from the given |module|
-  DecorationManager(ir::Module* module) { AnalyzeDecorations(module); }
+  DecorationManager(ir::Module* module) {
+    AnalyzeDecorations(module);
```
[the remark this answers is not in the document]
daniel-schuermann: Not sure if this PR isn't going out of scope. I was fixing the inline pass ...

s-perron:
@daniel-schuermann Are you still working on this?  You might need to resolve the conflicts and make sure the continuous integration test pass.

daniel-schuermann:
@s-perron Sorry, I thought I shouldn't rebase. Then, I squash as well and you have a final look?

s-perron:
You can squash if you like.  The rebase is the important part because of the conflicts.

s-perron:
Squashed, rebased, and merged as a76d097.

Sorry for the delay with this.  I should have told you that you can rebase earlier.

PR #921, opened 2017-10-25
Re-factor Phi insertion code out of LocalMultiStoreElimPass.

### LINKED ISSUE #893: Extend Phi placement to support unstructured control flow

dnovillo:
This moves the Phi insertion code up into MemPass.  It partially fixes #893, but it is missing some more re-factoring.  In moving this code around I found several other instances of something similar in other passes.  In subsequent PRs, I'll remove those redundancies as well.

This is an NFC.


source/opt/cfg_cleanup_pass.cpp
```diff
@@ -240,16 +224,9 @@ bool CFGCleanupPass::CFGCleanup(ir::Function* func) {
 }
 
 void CFGCleanupPass::Initialize(ir::Module* module) {
-  // Initialize the DefUse manager. TODO(dnovillo): Re-factor all this into the
-  // module or some other context class for the optimizer.
-  module_ = module;
-  def_use_mgr_.reset(new analysis::DefUseManager(consumer(), module));
+  InitializeProcessing(module);
   FindNamedOrDecoratedIds();
```
s-perron: Should  "FindNamedOrDecoratedIds" be included in "InitializeProcessing"?  All it does is set up a data structure so that "KillNamesAndDecorates" actually works.  Since this is all stuff related to MemPass it would be nice that the passes that extend MemPass do not have to call it.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -276,7 +253,7 @@ Pass::Status CFGCleanupPass::Process(ir::Module* module) {
   // Process all entry point functions.
   ProcessFunction pfn = [this](ir::Function* fp) { return CFGCleanup(fp); };
   bool modified = ProcessReachableCallTree(pfn, module);
-  FinalizeNextId(module_);
+  FinalizeNextId();
```
s-perron: Is there a reason we don't get rid of FinalizeNextId() and just incorporate into asking for a new id?  I don't see a benefit of having it one its own, and it could be a source of error if someone forgets to call it.

source/opt/cfg_cleanup_pass.cpp
```diff
@@ -127,8 +111,8 @@ void CFGCleanupPass::RemovePhiOperands(
       // means that this |phi| argument is no longer defined. Replace it with
       // |undef_id|.
       if (!undef_id) {
-        type_id = def_use_mgr_->GetDef(arg_id)->type_id();
-        undef_id = TypeToUndef(type_id);
+        type_id = get_def_use_mgr()->GetDef(arg_id)->type_id();
+        undef_id = Type2Undef(type_id);
```
dneto0: I prefer "To" to "2" in these situations.
"Nothing compares 2 u" is great for a song, but not for a method name. :-)

This is not a request to change.    :-)

source/opt/cfg_cleanup_pass.cpp
dneto0: This code was written before the DecorationManager existed.  I expect it will be refactored in the future to avoid the customized step.

For now I suggest a TODO.   I would not block this MR for lack of the TODO.
I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/928 for that refactor.

source/opt/cfg_cleanup_pass.cpp
dneto0: I'm ok deferring this.

I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/929 for this refactor.

source/opt/common_uniform_elim_pass.h
```diff
@@ -117,23 +111,25 @@ class CommonUniformElimPass : public Pass {
...
   // A block's structured successors are the blocks it branches to
   // together with its declared merge block if it has one.
   // When order matters, the merge block always appears first.
-  // This assures correct depth first search in the presence of early 
+  // This assures correct depth first search in the presence of early
   // returns and kills. If the successor vector contain duplicates
   // if the merge block, they are safely ignored by DFS.
+  //
+  // TODO(dnovillo): This pass computes structured successors slightly different
+  // than the implementation in class Pass. Can this be re-factored?
```
dneto0: Good catch!  Thanks for the TODO!

source/opt/local_access_chain_convert_pass.h
```diff
@@ -44,7 +44,9 @@ class LocalAccessChainConvertPass : public MemPass {
 
  private:
   // Return true if all refs through |ptrId| are only loads or stores and
-  // cache ptrId in supported_ref_ptrs_.
+  // cache ptrId in supported_ref_ptrs_. TODO(dnovillo): This function is
```
dneto0: Good catch!

source/opt/mem_pass.cpp
```diff
@@ -262,37 +217,448 @@ void MemPass::DCEInst(ir::Instruction* inst) {
...
+    phiItr->SetInOperand(idx, {valId});
+    // Analyze uses now that they are complete
+    get_def_use_mgr()->AnalyzeInstUse(&*phiItr);
+  }
+}
+
+Pass::Status MemPass::InsertPhiInstructions(ir::Function* func) {
+  // TODO(dnovillo) the current Phi placement mechanism assumes structured
+  // control-flow. This should be generalized
+  // (https://github.com/KhronosGroup/SPIRV-Tools/issues/893).
```
dneto0: +1
Good comment

source/opt/pass.h
```diff
@@ -96,8 +108,92 @@ class Pass {
   // succesful to indicate whether changes are made to the module.
   virtual Status Process(ir::Module* module) = 0;
 
+ protected:
+  // Initialize basic data structures for the pass. This sets up the def-use
+  // manager, module and other attributes. TODO(dnovillo): Some of this should
+  // be done during pass instantiation. Other things should be outside the pass
+  // altogether (e.g., def-use manager).
```

source/opt/cfg_cleanup_pass.cpp
dnovillo: Yeah, good idea.  Done.

dneto0:
Rebased, squashed, and pushed into master as 1040a95b

source/opt/cfg_cleanup_pass.cpp
dnovillo: D'oh!  Of course.  Done.

PR #923, opened 2017-10-25
Add option to relax validation of store types.

s-perron:
There are a number of users of spriv-opt that are hitting errors
because of stores with different types.  In general, this is wrong, but,
in these cases, the types are the exact same except for decorations.

The proper solution to the problem is being discussed, but in the
meantime I would add an option to accept these cases.

The options is "--relax-store-struct", and it can be used with the
validator or the optimizer.

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -207,8 +207,8 @@ bool idUsage::isValid<SpvOpGroupDecorate>(const spv_instruction_t* inst,
   auto decorationGroup = module_.FindDef(inst->words[decorationGroupIndex]);
   if (!decorationGroup || SpvOpDecorationGroup != decorationGroup->opcode()) {
     DIAG(decorationGroupIndex)
-        << "OpGroupDecorate Decoration group <id> '"
-        << inst->words[decorationGroupIndex] << "' is not a decoration group.";
+    << "OpGroupDecorate Decoration group <id> '"
+    << inst->words[decorationGroupIndex] << "' is not a decoration group.";
     return false;
   }
   return true;
@@ -275,8 +275,8 @@ bool idUsage::isValid<SpvOpEntryPoint>(const spv_instruction_t* inst,
     auto entryPointType = module_.FindDef(entryPoint->words()[4]);
     if (!entryPointType || 3 != entryPointType->words().size()) {
       DIAG(entryPointIndex)
-          << "OpEntryPoint Entry Point <id> '" << inst->words[entryPointIndex]
-          << "'s function parameter count is not zero.";
+      << "OpEntryPoint Entry Point <id> '" << inst->words[entryPointIndex]
+      << "'s function parameter count is not zero.";
       return false;
     }
   }
@@ -436,8 +436,8 @@ bool idUsage::isValid<SpvOpTypeStruct>(const spv_instruction_t* inst,
     auto memberType = module_.FindDef(memberTypeId);
     if (!memberType || !spvOpcodeGeneratesType(memberType->opcode())) {
       DIAG(memberTypeIndex)
-          << "OpTypeStruct Member Type <id> '" << inst->words[memberTypeIndex]
-          << "' is not a type.";
+      << "OpTypeStruct Member Type <id> '" << inst->words[memberTypeIndex]
+      << "' is not a type.";
       return false;
     }
     if (SpvOpTypeStruct == memberType->opcode() &&
@@ -617,10 +617,10 @@ bool idUsage::isValid<SpvOpConstantComposite>(const spv_instruction_t* inst,
         if (!constituentResultType ||
             componentType->opcode() != constituentResultType->opcode()) {
           DIAG(constituentIndex)
-              << "OpConstantComposite Constituent <id> '"
-              << inst->words[constituentIndex]
-              << "'s type does not match Result Type <id> '" << resultType->id()
-              << "'s vector element type.";
+          << "OpConstantComposite Constituent <id> '"
+          << inst->words[constituentIndex]
+          << "'s type does not match Result Type <id> '" << resultType->id()
+          << "'s vector element type.";
           return false;
         }
       }
@@ -646,7 +646,7 @@ bool idUsage::isValid<SpvOpConstantComposite>(const spv_instruction_t* inst,
            constituentIndex++) {
         auto constituent = module_.FindDef(inst->words[constituentIndex]);
         if (!constituent || !(SpvOpConstantComposite == constituent->opcode() ||
-                              SpvOpUndef == constituent->opcode())) {
+            SpvOpUndef == constituent->opcode())) {
           // The message says "... or undef" because the spec does not say
           // undef is a constant.
           DIAG(constituentIndex) << "OpConstantComposite Constituent <id> '"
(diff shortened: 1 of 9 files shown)


s-perron:
FYI, I ran clang-format on the files I changed.  That created a lot of purely formatting change.  I can undo some of that if people want.

tools/opt/opt.cpp
```diff
@@ -155,6 +156,10 @@ NOTE: The optimizer is a work in progress.
                Replaces instructions with equivalent and less expensive ones.
   --eliminate-dead-variables
                Deletes module scope variables that are not referenced.
+  --relax-store-struct
+               Allow store from one struct type to a different, but 
+               interchangeable, type.  This option is forwarded to the
```
dneto0: I think  ."... to a different struct type with compatible layout and members".

test/val/val_id_test.cpp
```diff
@@ -2165,6 +2162,177 @@ TEST_F(ValidateIdWithMessage, OpStoreTypeBad) {
...
+%16 = OpCompositeConstruct %4 %11 %11
+      OpStore %12 %16
+      OpReturn
+      OpFunctionEnd)";
+  spvValidatorOptionsSetRelaxStoreStruct(options_, true);
+  CompileSuccessfully(spirv.c_str());
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("OpStore Pointer <id> '12's type does not match Object "
+                            "<id> '4's type."));
```
dneto0: Add a TODO for exercising the other layout-affecting decorations.

Please add a test for showing the relaxation will allow matching of distinct types where the layouts are the same but some *other* kind of decoration is different between them.  E.g. RelaxedPrecision on a float member in one struct but not the other.

source/validate_id.cpp
```diff
@@ -1199,10 +1215,14 @@ bool idUsage::isValid<SpvOpStore>(const spv_instruction_t* inst,
...
-    DIAG(pointerIndex) << "OpStore Pointer <id> '" << inst->words[pointerIndex]
-                       << "'s type does not match Object <id> '"
-                       << objectType->id() << "'s type.";
-    return false;
+    if (!module_.options()->relax_struct_store ||
+        !AreInterchangeableSructs(type, objectType)) {
+      DIAG(pointerIndex) << "OpStore Pointer <id> '"
+                         << inst->words[pointerIndex]
+                         << "'s type does not match Object <id> '"
+                         << objectType->id() << "'s type.";
```
atgoo: Consider having two different error messages for id mismatch and data layout mismatch.

source/validate_id.cpp
```diff
@@ -2495,6 +2518,75 @@ bool idUsage::isValid(const spv_instruction_t* inst) {
...
+      case SpvDecorationArrayStride:
+      case SpvDecorationOffset:
+      case SpvDecorationMatrixStride:
+      case SpvDecorationRowMajor:
+      case SpvDecorationColMajor:
+        // Since these effect the layout of the struct, they must be present in
+        // both structs.
+        if (std::find(type2_decorations.begin(), type2_decorations.end(),
+                      decoration) == type2_decorations.end()) {
+          return false;
```
atgoo: Can type2 have decorations which type1 doesn't have? (question out if ignorance)

s-perron: I've added a comment explaining this.  I don't think they can in a legal program.  The layout has to be completely determined.  For example, every member in the struct has to have an offset.  If the offset for the first member in type1 matches the offset for the first member in type 2, then type 2 cannot have a different giving the first member a different offset.  This same reasoning applies to all of the possible decorations.

s-perron:
Updated.  Ready for another review.

source/validate_id.cpp
dneto0: The Vulkan environment spec requires that layout decorations be complete.  (e.g. all structs have offsets for all members, all arrays have strides, etc.)  So if the member type is the of the same optype* then they'll also have layout decorations.
I expect OpenGL does the same.

test/val/val_id_test.cpp
```diff
@@ -2162,7 +2159,180 @@ TEST_F(ValidateIdWithMessage, OpStoreTypeBad) {
...
+      OpStore %12 %16
+      OpReturn
+      OpFunctionEnd)";
+  spvValidatorOptionsSetRelaxStoreStruct(options_, true);
+  CompileSuccessfully(spirv.c_str());
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(
+      getDiagnosticString(),
+      HasSubstr("OpStore Pointer <id> '12's layout does not match Object "
+                    "<id> '16's layout."));
```
dneto0: You missed the test requests from my first review.:

Add a TODO for exercising the other layout-affecting decorations.

Please add a test for showing the relaxation will allow matching of distinct types where the layouts are the same but some other kind of decoration is different between them. E.g. RelaxedPrecision on a float member in one struct but not the other.

tools/val/val.cpp
```diff
@@ -44,6 +44,7 @@ NOTE: The validator is a work in progress.
   --max-function-args              <maximum number arguments allowed per function>
   --max-control-flow-nesting-depth <maximum Control Flow nesting depth allowed>
   --max-access-chain-indexes       <maximum number of indexes allowed to use for Access Chain instructions>
+  --relax-struct-store             Allow store from one struct type to a different type with compatible layout and members.
```
dneto0: Avoid going past 80 columns.

test/val/val_id_test.cpp
s-perron: Done

tools/val/val.cpp
s-perron: I fixed the description for the option I added.  I can fix the other options too, but you'll have to tell me what you expect.

test/val/val_id_test.cpp
s-perron: I added two TODOs.  One to add more test for layout compatible arrays and matrices.  The other to update checks for layout compatible structs to consider matrices and arrays.

source/validate_id.cpp
dneto0: I apologize, I answered incorrectly because I forgot a key qualification.  @atgoo is right to question this.

_When objects are in UniformConstant, Uniform, PushConstant, and StorageBuffer storage classes_, then their layout must be explicit.    But I think this code could come across cases where type1 is a struct value that was never associated with an object on of those storage classes.  And so you can get the case where type2 has layout constraints but type1 does not.

source/validate_id.cpp
```diff
@@ -2495,6 +2528,88 @@ bool idUsage::isValid(const spv_instruction_t* inst) {
...
+      "type1 must be and OpTypeStruct instruction.");
+  assert(type2->opcode() == SpvOpTypeStruct &&
+      "type2 must be and OpTypeStruct instruction.");
+  const std::vector<Decoration>& type1_decorations =
+      module_.id_decorations(type1->id());
+  const std::vector<Decoration>& type2_decorations =
+      module_.id_decorations(type2->id());
+
+  // For shader modules, it is required that composite objects have
+  // decorations that completely define their layout.  See section 2.16.2 of
```
dneto0: This applies to composite objects in certain storage classes.  (Sometimes colloquially called "transparent" memory, but it's the storage classes which map to storage that the application CPU code writes to.)  See 2.16.2 about composite objects (and also add StorageBuffer storage class... )

tools/val/val.cpp
dneto0: Thanks.  Just keep the fixes to the one you added.

source/validate_id.cpp
```diff
@@ -972,8 +975,8 @@ bool idUsage::isValid<SpvOpSpecConstantComposite>(const spv_instruction_t* inst,
         auto constituent = module_.FindDef(inst->words[constituentIndex]);
         auto constituentOpCode = constituent->opcode();
         if (!constituent || !(SpvOpSpecConstantComposite == constituentOpCode ||
-            SpvOpConstantComposite == constituentOpCode ||
-            SpvOpUndef == constituentOpCode)) {
+                              SpvOpConstantComposite == constituentOpCode ||
```
dneto0: There's a lot of reformatting going on.  Is one tool inconsistent with another?

s-perron: Formatting is fixed, but I can't get clang-format to do the right thing.

dneto0:
Rebased, squashed, and pushed into master as 716138ee14

PR #925, opened 2017-10-26
Add validation for OpBranchConditional

nsubtil:


nsubtil:
Continues work started in #888

source/validate_id.cpp
```diff
@@ -2002,11 +2002,54 @@ bool idUsage::isValid<OpBranch>(const spv_instruction_t *inst,
...
+    const spv_instruction_t *inst, const spv_opcode_desc opcodeEntry)
+{
+  const size_t numOperands = inst->words.size() - 1;
+  constexpr size_t condOperandIndex = 1;
+  constexpr size_t targetTrueIndex = 2;
+  constexpr size_t targetFalseIndex = 3;
+  constexpr size_t weightTrueIndex = 4;
+  constexpr size_t weightFalseIndex = 5;
+
+  // num_operands is either 3 or 5 --- if 5, the last two need to be literal integers
```
nsubtil: I'm not actually sure how to validate that the last two operands are literal integers. It seems like they're just two dwords in the instruction that can always be interpreted as literal integers.

Is there something that we need to check here?

source/validate_id.cpp
```diff
@@ -2002,11 +2002,52 @@ bool idUsage::isValid<OpBranch>(const spv_instruction_t *inst,
...
-#if 0
 template <>
-bool idUsage::isValid<OpBranchConditional>(
-    const spv_instruction_t *inst, const spv_opcode_desc opcodeEntry) {}
-#endif
+bool idUsage::isValid<SpvOpBranchConditional>(
+    const spv_instruction_t *inst, const spv_opcode_desc)
+{
+  const size_t numOperands = inst->words.size() - 1;
+  constexpr size_t condOperandIndex = 1;
```
atgoo: Consider const. Microsoft doesn't fully support the C++ standard.
https://msdn.microsoft.com/en-us/library/hh567368.aspx

source/validate_id.cpp
```diff
@@ -2002,11 +2002,52 @@ bool idUsage::isValid<OpBranch>(const spv_instruction_t *inst,
...
 #endif
 
-#if 0
 template <>
-bool idUsage::isValid<OpBranchConditional>(
-    const spv_instruction_t *inst, const spv_opcode_desc opcodeEntry) {}
-#endif
+bool idUsage::isValid<SpvOpBranchConditional>(
+    const spv_instruction_t *inst, const spv_opcode_desc)
+{
```
atgoo: Consider moving the brace to the previous line for consistency.
f () {
}

source/validate_id.cpp
```diff
@@ -2002,11 +2002,52 @@ bool idUsage::isValid<OpBranch>(const spv_instruction_t *inst,
...
+    }
+
+    auto targetOpFalse = module_.FindDef(inst->words[targetFalseIndex]);
+    if (!targetOpFalse || SpvOpLabel != targetOpFalse->opcode()) {
+      DIAG(0) << "The 'False Label' operand for OpBranchConditional must be the ID of an OpLabel instruction";
+      ret = false;
+    }
+
+    if (!ret) {
+      return false;
```
atgoo: Consider
return ret;

source/validate_id.cpp
```diff
@@ -2002,11 +2002,52 @@ bool idUsage::isValid<OpBranch>(const spv_instruction_t *inst,
...
+    }
+  }
```
atgoo: You don't really need this {} block.

test/val/val_id_test.cpp
```diff
@@ -3946,7 +3946,147 @@ TEST_F(ValidateIdWithMessage, OpVectorShuffleLiterals) {
 // TODO: OpLoopMerge
 // TODO: OpSelectionMerge
 // TODO: OpBranch
-// TODO: OpBranchConditional
+
+string BranchConditionalSetup = R"(
```
atgoo: Please move the string definitions to the top of the file.

source/validate_id.cpp
dneto0: Yup.  We support VS 2013.  Sadly, constexpr can't be used in SPIRV-Tools.
Use const here.  Or even an enum.  (Use const)

source/validate_id.cpp
atgoo: I don't think so.

If you write code directly in SPIR-V (i.e. not in assembly, but in raw SPIR-V) and you put here an id referencing the integer you want to use instead of a literal integer, then there is pretty much nothing we can do: the id would be interpreted as the weight and the code would look perfectly valid.

source/validate_id.cpp
```diff
@@ -2002,11 +2002,45 @@ bool idUsage::isValid<OpBranch>(const spv_instruction_t *inst,
...
+  if (numOperands != 3 &&
+      numOperands != 5) {
+    DIAG(0) << "OpBranchConditional requires either 3 or 5 parameters";
+    return false;
+  }
+
+  bool ret = true;
+
+  // grab the condition operand and check that it is a bool
+  auto condOp = module_.FindDef(inst->words[condOperandIndex]);
```
dneto0: Recommend const whenever we can.  That's true for the three operand variables.

source/validate_id.cpp
```diff
@@ -2002,11 +2002,45 @@ bool idUsage::isValid<OpBranch>(const spv_instruction_t *inst,
...
+  if (!condOp || !module_.IsBoolScalarType(condOp->type_id())) {
+    DIAG(0) << "Condition operand for OpBranchConditional must be of boolean type";
+    ret = false;
+  }
+
+  // target operands must be OpLabel
+  auto targetOpTrue = module_.FindDef(inst->words[targetTrueIndex]);
```
dneto0: Should also add a comment that validate_cfg.cpp checks that trueLabel and falseLabel are in the same function.  (It's the first check in PerformCfgChecks in that source file.)    So we don't need to check that condition here.

test/val/val_id_test.cpp
```diff
@@ -94,6 +94,52 @@ string sampledImageSetup = R"(
             %sampler_inst = OpLoad %sampler_type %s
 )";
 
+string BranchConditionalSetup = R"(
```
dneto0: FYI,  the style guide says "Variables of class type with static storage duration are forbidden"
For an explanation, see https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables 

I know this file breaks the rule multiple times.
Since this is test code, it's not critical to fix.  But we do enforce the rule in the library code.

The alternative is either  const char array, pointer-to-const-char, or a function returning a static string value.

atgoo: That rule exists probably because a very large program would initialize all such variables on startup even if it will never use them. Doesn't really apply in this case.

dneto0: It's for a bunch of reasons, but the most critical ones are the confusion and bugs that arise from indeterminate initialization order and indeterminate destruction order. 

It happens in real life with SPIRV-Tools:   When SPIRV-Tools is used in the MicrosoftDirectXShaderCompiler, the main process replaces the memory allocator at some point (I believe after 'main' starts), and if you have variables of the banned type, then there is trouble at process shutdown because the deallocator mismatches the allocator.

Yes, that does not apply for test code (which this is an instance of).  That's why it's a FYI in this case.

nsubtil:
I believe I have addressed all the review feedback. @dneto0 or @atgoo, can you take one final look?

dneto0:
Rebased and pushed into master as d861ceffd

PR #927, opened 2017-10-27
Compression: removed 'presumed index' feature

atgoo:
The feature used to improve compression of const integers which were
presumed to be indices. Now obsolete as descriptor-based compression
does this in a more generalized way.

### DIFF
--- a/source/comp/markv_codec.cpp
+++ b/source/comp/markv_codec.cpp
@@ -103,8 +103,6 @@ enum : uint64_t {
   kMtfTypeIntScalarOrVector,
   // All types declared as return types in OpTypeFunction.
   kMtfTypeReturnedByFunction,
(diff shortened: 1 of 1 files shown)


source/comp/markv_codec.cpp
```diff
@@ -271,7 +265,7 @@ size_t GetNumBitsToNextByte(size_t bit_pos) {
 // Defines and returns current MARK-V version.
 uint32_t GetMarkvVersion() {
   const uint32_t kVersionMajor = 1;
-  const uint32_t kVersionMinor = 2;
+  const uint32_t kVersionMinor = 3;
```
dneto0: thank you!

dneto0:
Rebased and pushed into master as 6724c2725

PR #930, opened 2017-10-29
Tooling - CMake - Export symbol when creating shared library on Windows

theblackunknown:
Hello people,

Not sure if that be useful for anyone, but I am experimenting with this project and noticed that no symbols gets exported when creating a shared library on Windows.

So I modified build script and headers to select what should be exported or not.
The approach should be quite portable as I rely on CMake ``GenerateExportHeader`` module.

I haven*t touched travis or appveyor file to add windows shared library build to CI as I don't know what this is the recommended way to do so, but if merged it should go there to verify shared library correctly generate export symbol

Let me know what you think of it and if it worth a merge to upstream.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -22,6 +22,8 @@
 
 #include "libspirv.hpp"
 
+#include "spirv-tools-opt_export.h"
+
 namespace spvtools {
 
 // C++ interface for SPIR-V optimization functionalities. It wraps the context
@@ -38,15 +40,15 @@ class Optimizer {
   struct PassToken {
     struct Impl;  // Opaque struct for holding inernal data.
 
-    PassToken(std::unique_ptr<Impl>);
+    SPIRV_TOOLS_OPT_EXPORT PassToken(std::unique_ptr<Impl>);
 
     // Tokens can only be moved. Copying is disabled.
-    PassToken(const PassToken&) = delete;
-    PassToken(PassToken&&);
-    PassToken& operator=(const PassToken&) = delete;
-    PassToken& operator=(PassToken&&);
+    SPIRV_TOOLS_OPT_EXPORT PassToken(const PassToken&) = delete;
+    SPIRV_TOOLS_OPT_EXPORT PassToken(PassToken&&);
+    SPIRV_TOOLS_OPT_EXPORT PassToken& operator=(const PassToken&) = delete;
+    SPIRV_TOOLS_OPT_EXPORT PassToken& operator=(PassToken&&);
 
-    ~PassToken();
+    SPIRV_TOOLS_OPT_EXPORT ~PassToken();
 
     std::unique_ptr<Impl> impl_;  // Unique pointer to internal data.
   };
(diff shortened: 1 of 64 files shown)


dneto0:
Hi @Kangz can you comment?  (You added the shared lib support).
As I recall, isn't there a less intrusive way to do this, via setting build defaults?

source/CMakeLists.txt
```diff
@@ -188,89 +188,61 @@ add_custom_target(spirv-tools-build-version
...
-add_subdirectory(link)
+add_library(${SPIRV_TOOLS} "")
 
-set(SPIRV_SOURCES
-  ${spirv-tools_SOURCE_DIR}/include/spirv-tools/libspirv.h
+include(GenerateExportHeader)
+generate_export_header(${SPIRV_TOOLS}
+  BASE_NAME spirv-tools)
 
-  ${CMAKE_CURRENT_SOURCE_DIR}/util/bitutils.h
```
Kangz: .h files are important to keep in the source list so that they are shown in IDEs like visual studio.

source/CMakeLists.txt
```diff
@@ -188,89 +188,61 @@ add_custom_target(spirv-tools-build-version
...
 set_property(TARGET spirv-tools-build-version PROPERTY FOLDER "SPIRV-Tools build")
 
-add_subdirectory(comp)
-add_subdirectory(opt)
-add_subdirectory(link)
+add_library(${SPIRV_TOOLS} "")
 
-set(SPIRV_SOURCES
-  ${spirv-tools_SOURCE_DIR}/include/spirv-tools/libspirv.h
+include(GenerateExportHeader)
```
Kangz: Why not make your own export header, it is simple to do and makes the code more clear imho.

include/spirv-tools/optimizer.hpp
```diff
@@ -57,35 +59,35 @@ class Optimizer {
   // The constructed instance will have an empty message consumer, which just
   // ignores all messages from the library. Use SetMessageConsumer() to supply
   // one if messages are of concern.
-  explicit Optimizer(spv_target_env env);
+  SPIRV_TOOLS_OPT_EXPORT explicit Optimizer(spv_target_env env);
```
Kangz: You can make the whole class SPIRV_TOOLS_OPT_EXPORT instead of every function individually.

Kangz:
@dneto0 he changes are a bit more intrusive then they could if whole classes were exported and the export headers were written manually (vs. generated inside CMake).

Exporting symbols / classes individually is the preferred way to do this in the Chromium style. Since this is not Chromium, the alternative is to have the following CMake code:

```
set_target_properties(${Target} PROPERTIES
                      WINDOWS_EXPORT_ALL_SYMBOLS true
                      C_VISIBILITY_PRESET visible)
```

include/spirv-tools/optimizer.hpp
dneto0: Yes, that was my recollection.  (I haven't had to do this in some time...)
That is much preferred: mark an entire class rather than individual methods.

dneto0:
@Kangz : Thanks for the advice/confirmation.  

My summary:
* I strongly prefer not to export all symbols. 
* The files in include/spirv-tools/* are all meant to be the export interface of the various libraries.  It bears a review to double-check before applying export annotations.

theblackunknown:
I totally agree on your conclusion that those changes are quite intrusive, at the cost of having finer-grained control over what should be exposed or not.
As @Kangz  mentionned using CMake ``WINDOWS_EXPORT_ALL_SYMBOLS `` and ``[LANG]_VISIBILITY_PRESET`` should be equivalent and less intrusive if we want to sacrifice the finer-grained control.
If we go that way also ``CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS `` and ``CMAKE_[LANG]_VISIBILITY_PRESET`` at the root CMakeLists.txt to set the default value for every subsequent library resulting in less changes.

I can try experimenting with those, in the meantime let me know what are your thoughts on this

theblackunknown:
So ``CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`` works pretty well and it has no if you are not on Windows.
One downside is that linker is called everytime you trigger a ``cmake --build`` with generator "Visual Studio Code 15 2017".
It is not that long but definitively annoying when I try to have a quick iteration cycle.

**NOTE** Using Ninja or NMake Makefiles this problem does not show up

dj2:
Is this PR still valid, or can we close it?

Kangz:
It should be closed, I fixed it in a PR that was based on this one.

theblackunknown:
I am fine to have it closed, I didn't touched it in a while. 
Thanks @Kangz to have fixed it, I'll give it a try !

PR #932, opened 2017-10-30
Lowered initial capacity of move-to-front sequences

atgoo:
Also fixed outdated comments.


dneto0:
Rebased and pushed into master as 7299fb5b

PR #933, opened 2017-10-30
Add a new class opt::CFG to represent the CFG for the module.

dnovillo:
This class moves some of the CFG-related functionality into a new
class opt::CFG.  There is some other code related to the CFG in the
inliner and in opt::LocalSingleStoreElimPass that should also be moved,
but that require more changes than this pure restructuring.

I will move those bits in a follow-up PR.

Currently, the CFG is computed every time a pass is instantiated, but
this should be later moved to the new IRContext class that @s-perron is
working on.

### DIFF
--- a/source/opt/basic_block.cpp
+++ b/source/opt/basic_block.cpp
@@ -19,6 +19,14 @@
 namespace spvtools {
 namespace ir {
 
+namespace {
+
+const uint32_t kLoopMergeContinueBlockIdInIdx = 1;
+const uint32_t kLoopMergeMergeBlockIdInIdx = 0;
+const uint32_t kSelectionMergeMergeBlockIdInIdx = 0;
+
+}  // namespace
+
 BasicBlock::BasicBlock(const BasicBlock& bb)
     : function_(nullptr),
       label_(MakeUnique<Instruction>(bb.GetLabelInst())),
@@ -35,7 +43,7 @@ const Instruction* BasicBlock::GetMergeInst() const {
   if (iter != cbegin()) {
     --iter;
     const auto opcode = iter->opcode();
-    if (opcode == SpvOpLoopMerge || opcode  == SpvOpSelectionMerge) {
+    if (opcode == SpvOpLoopMerge || opcode == SpvOpSelectionMerge) {
       result = &*iter;
     }
   }
@@ -50,7 +58,7 @@ Instruction* BasicBlock::GetMergeInst() {
   if (iter != begin()) {
     --iter;
     const auto opcode = iter->opcode();
-    if (opcode == SpvOpLoopMerge || opcode  == SpvOpSelectionMerge) {
+    if (opcode == SpvOpLoopMerge || opcode == SpvOpSelectionMerge) {
       result = &*iter;
     }
   }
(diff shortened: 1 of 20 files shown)


source/opt/basic_block.cpp
```diff
@@ -101,13 +109,31 @@ void BasicBlock::ForMergeAndContinueLabel(
...
-  if (ii->opcode() == SpvOpSelectionMerge || 
-      ii->opcode() == SpvOpLoopMerge)
-    ii->ForEachInId([&f](const uint32_t* idp) {
-      f(*idp);
-    });
+  if (ii->opcode() == SpvOpSelectionMerge || ii->opcode() == SpvOpLoopMerge)
+    ii->ForEachInId([&f](const uint32_t* idp) { f(*idp); });
+}
+
+uint32_t BasicBlock::MergeBlockIdIfAny(uint32_t* cbid) {
```
s-perron: Not that this needs to be done in this PR, but it is worth splitting this into two function?  One to get the merge block, and another to get the continue block?  It does not look like there is any looping going on.  I won't fight for it if others disagree.

source/opt/basic_block.cpp
```diff
@@ -101,13 +109,31 @@ void BasicBlock::ForMergeAndContinueLabel(
...
+  auto merge_ii = cend();
+  --merge_ii;
+  if (cbid != nullptr) {
+    *cbid = 0;
+  }
+  uint32_t mbid = 0;
```
s-perron: This code looks a lot like trying to fine the merge instruction.  Could we call "GetMergeInst" instead?

source/opt/basic_block.h
```diff
@@ -118,6 +120,22 @@ class BasicBlock {
     return count > 0;
   }
 
+  // Return true if this block is a loop header block.
+  bool IsLoopHeader() const {
+    auto iItr = cend();
```
s-perron: could we just use "GetLoopMergeInst() != nullptr"?  The other issue is that all of this only works for structured control flow.  Will we want it to be more general than that?  How should we let OpenCL people know they cannot use this function?

source/opt/cfg.h
```diff
@@ -0,0 +1,104 @@
...
+  bool IsPseudoExitBlock(ir::BasicBlock* block_ptr) const {
+    return block_ptr == &pseudo_exit_block_;
+  }
+
+  // Compute structured successors for function |func|. A block's structured
+  // successors are the blocks it branches to together with its declared merge
+  // block if it has one. When order matters, the merge block always appears
+  // first. This assures correct depth first search in the presence of early
+  // returns and kills. If the successor vector contain duplicates if the merge
+  // block, they are safely ignored by DFS. TODO(dnovillo): This belongs in a
```
s-perron: Can you remove the TODO now?

source/opt/cfg.h
```diff
@@ -0,0 +1,104 @@
...
+  // CFG class.
+  void ComputeStructuredSuccessors(ir::Function* func);
+
+  // Compute structured block order for |func| into |structuredOrder|. This
+  // order has the property that dominators come before all blocks they
+  // dominate and merge blocks come after all blocks that are in the control
+  // constructs of their header. TODO(dnovillo): This belongs in a CFG class.
```
s-perron: Remove the TODO if it is done.

s-perron:
Also, the continuous builds are failing, so you need to check that out as well.

dneto0:
A bunch of changes were just pushed into master.  Please rebase.

dnovillo:
> A bunch of changes were just pushed into master. Please rebase.

Done.

source/opt/basic_block.cpp
dnovillo: Yeah, this function bugged me a bit because it does two distinct things.  I've added BasicBlock::ContinueBlockIdIfAny().

source/opt/basic_block.h
dnovillo: Done.

I've also asserted that we're processing a shader module in the SSA re-writer and the structured control compute.

dnovillo:
Re-based, re-tested, and pushed to master.

PR #934, opened 2017-10-31
Add the IRContext (part 1)

s-perron:
This is the first part of adding the IRContext.  This class is meant to
hold the extra data that is build on top of the module that it
owns.

The first part will simply create the IRContext class and get it passed
to the passes in place of the module.  For now it does not have any
functionality of its own, but it acts more as a wrapper for the module.

The functions that I added to the IRContext are those that either
traverse the headers or add to them.  I did this because we may decide
to have other ways of dealing with these sections (for example adding a
type pool, or use the decoration manager).

I also added the function that add to the header because the IRContext
needs to know when an instruction is added to update other data
structures appropriately.

Note that there is still lots of work that needs to be done.  There are
still many places that change the module, and do not inform the context.
That will be the next step.

### DIFF
--- a/source/opt/block_merge_pass.cpp
+++ b/source/opt/block_merge_pass.cpp
@@ -17,6 +17,7 @@
 #include "block_merge_pass.h"
 
 #include "iterator.h"
+#include "ir_context.h"
 
 namespace spvtools {
 namespace opt {
@@ -94,8 +95,8 @@ bool BlockMergePass::MergeBlocks(ir::Function* func) {
   return modified;
 }
 
-void BlockMergePass::Initialize(ir::Module* module) {
-  InitializeProcessing(module);
+void BlockMergePass::Initialize(ir::IRContext* c) {
+  InitializeProcessing(c);
 
   // Initialize extension whitelist
   InitExtensions();
@@ -126,8 +127,8 @@ Pass::Status BlockMergePass::ProcessImpl() {
 
 BlockMergePass::BlockMergePass() {}
 
-Pass::Status BlockMergePass::Process(ir::Module* module) {
-  Initialize(module);
+Pass::Status BlockMergePass::Process(ir::IRContext* c) {
+  Initialize(c);
   return ProcessImpl();
 }
 
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -1,0 +1,292 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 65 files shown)


source/opt/ir_context.h
```diff
@@ -0,0 +1,293 @@
...
+  inline Module::inst_iterator debug3_begin();
+  inline Module::inst_iterator debug3_end();
+  inline IteratorRange<Module::inst_iterator> debugs3();
+  inline IteratorRange<Module::const_inst_iterator> debugs3() const;
+
+  // Clears all debug instructions (excluding OpLine & OpNoLine).
+  inline void debug_clear();
+
+  // Appends a capability instruction to this module.
+  inline void AddCapability(std::unique_ptr<Instruction> c);
```
atgoo: std::unique_ptr<Instruction>&&?

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -249,8 +249,8 @@ bool LocalAccessChainConvertPass::ConvertLocalAccessChains(ir::Function* func) {
   return modified;
 }
 
-void LocalAccessChainConvertPass::Initialize(ir::Module* module) {
-  InitializeProcessing(module);
+void LocalAccessChainConvertPass::Initialize(ir::IRContext* c) {
+  InitializeProcessing(c);
```
dnovillo: Not a change request now, but for future PRs.  Could you move this call to Pass::InitializeProcessing() somewhere that gets called automatically?  I would like to avoid having the passes call this.  Ideally, Pass::InitializeProcessing would be part of the Pass constructor itself.

source/opt/ir_context.h
ratchetfreak: this is actually fine you can move the unique_ptr into the param and then the function will have to move again into the final location.

It's an extra move but for unique_ptr it's just a pointer copy.

atgoo: I don't agree. In one case you "can move", in another you "must move".

ratchetfreak: unique_ptr must be moved anyway

atgoo: That's good of them. So the only functionality we are getting from this is that we do the move twice instead of doing it once?

s-perron: I'll change it just to be more precise.

source/opt/local_access_chain_convert_pass.cpp
s-perron: Yes, I want to do something like that.  When I add the code that builds and invalidates things in the IRContext, I will do that.

source/opt/block_merge_pass.cpp
```diff
@@ -94,8 +95,8 @@ bool BlockMergePass::MergeBlocks(ir::Function* func) {
   return modified;
 }
 
-void BlockMergePass::Initialize(ir::Module* module) {
-  InitializeProcessing(module);
+void BlockMergePass::Initialize(ir::IRContext* c) {
```
dneto0: Tiny thing:   My preference order is:   'context' > 'ctx' > 'c'.
I see that context() is a method, so I would have used ctx here.

But that's definitely non-blocking feedback, given how little the identifier is used.

dneto0:
Waiting for the bots to pass.

PR #935, opened 2017-10-31
Fix copyright check when the build directory is in the source directory.

s-perron:
The IDE I am using places the cmake build directory inside the source
directory.  When that happens the copyright check fails because the
build process creates a file CMakeCXXCompilerId.cpp.  That file does
not have the copyright information, so the test fails.

I want this file to be ignored, so I added its directory, CompilerIdCXX,
to the list of directories to be ignored.


dneto0:
Rebased and pushed into master as b1b494835d

PR #938, opened 2017-11-01
Validate storage class of target pointer for OpStore

nsubtil:


test/val/val_id_test.cpp
```diff
@@ -2092,7 +2092,7 @@ TEST_F(ValidateIdWithMessage, OpStoreGood) {
   string spirv = kGLSL450MemoryModel + R"(
 %1 = OpTypeVoid
 %2 = OpTypeInt 32 0
-%3 = OpTypePointer UniformConstant %2
+%3 = OpTypePointer Uniform %2
```
nsubtil: These seem like invalid OpStore targets to me since UniformConstant is read-only, so I updated them to Uniform.

Note that I did not change the corresponding OpVariable storage class. The spec does not seem to require that these match, which I find a bit odd --- it's not clear to me that this is actually valid either.

dneto0: > These seem like invalid OpStore targets to me since UniformConstant is read-only, so I updated them to Uniform.

Thank you!

> Note that I did not change the corresponding OpVariable storage class. The spec does not seem to require that these match, which I find a bit odd --- it's not clear to me that this is actually valid either.

That's missing functionality in the validator's OpVariable check.  I added https://github.com/KhronosGroup/SPIRV-Tools/issues/941 for it.

The spec says in OpVariable: "
Result Type must be an OpTypePointer. Its Type operand is the type of object in memory.

Storage Class is the Storage Class of the memory holding the object. It cannot be Generic."

I read that as implicitly constraining the storage class of the pointer type to be the same as the storage class of the OpVariable.    But it would be fair to request an explicit clarification.

test/val/val_id_test.cpp
```diff
@@ -2481,6 +2481,45 @@ TEST_F(ValidateIdWithMessage, DISABLED_OpStoreFunction) {
...
+
+               OpReturn
+               OpFunctionEnd
+  )";
+
+  CompileSuccessfully(spirv.c_str());
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("storage class is read-only"));
+}
```
dneto0: Would have been nice to have tests for the other storage classes that are read-only. 

Also, this is a bit more complex than needed, since this example involves a built-in.  Conceivably someone could break the code by writing a check based on builtins.

dneto0:
Rebased and pushed into master as 2dddb8193b49

PR #940, opened 2017-11-02
Move class CFG from namespace opt to namespace ir.

dnovillo:
It makes more sense to have the CFG inside the ir name space, as it is
descriptive of the representation.


s-perron:
LGTM

PR #948, opened 2017-11-06
Opt: Fix HasLoads to not interpret decoration as load.

greg-lunarg:
Multi-store elimination was aborting because HasLoads was incorrectly interpreting a variable decoration 
(RelaxedPrecision) as a load.

This was blocking use of spirv-opt in the Vulkan CTS.


dneto0:
Should push this onto master after bots pass.   Request that @dnovillo or @s-perron do this.

dnovillo:
Re-based, tested and pushed as ac04b2faeabc15811e8c542a5d7a36dbdba243b8

Issue #946, opened 2017-11-06
Dubious optimization of UBO blocks


Themaister:
Using spirv-opt -O -o opt.spv test.spv

Input SPIR-V
```
; SPIR-V
; Version: 1.0
; Generator: Google Shaderc over Glslang; 2
; Bound: 80
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main" %gl_InstanceIndex %Position %__0
               OpSource GLSL 450
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %main "main"
               OpName %World "World"
               OpName %StaticMeshInfo "StaticMeshInfo"
               OpMemberName %StaticMeshInfo 0 "Model"
               OpMemberName %StaticMeshInfo 1 "Normal"
               OpName %PerVertexData "PerVertexData"
               OpMemberName %PerVertexData 0 "infos"
               OpName %_ ""
               OpName %gl_InstanceIndex "gl_InstanceIndex"
               OpName %Position "Position"
               OpName %gl_PerVertex "gl_PerVertex"
               OpMemberName %gl_PerVertex 0 "gl_Position"
               OpMemberName %gl_PerVertex 1 "gl_PointSize"
               OpMemberName %gl_PerVertex 2 "gl_ClipDistance"
               OpMemberName %gl_PerVertex 3 "gl_CullDistance"
               OpName %__0 ""
               OpName %RenderParameters "RenderParameters"
               OpMemberName %RenderParameters 0 "projection"
               OpMemberName %RenderParameters 1 "view"
               OpMemberName %RenderParameters 2 "view_projection"
               OpMemberName %RenderParameters 3 "inv_projection"
               OpMemberName %RenderParameters 4 "inv_view"
               OpMemberName %RenderParameters 5 "inv_view_projection"
               OpMemberName %RenderParameters 6 "inv_local_view_projection"
               OpMemberName %RenderParameters 7 "camera_position"
               OpMemberName %RenderParameters 8 "camera_front"
               OpMemberName %RenderParameters 9 "camera_right"
               OpMemberName %RenderParameters 10 "camera_up"
               OpMemberName %RenderParameters 11 "z_near"
               OpMemberName %RenderParameters 12 "z_far"
               OpName %global "global"
[...]

PR #949, opened 2017-11-07
Opt: Remove CommonUniformElimination from -O and -Os (for now)

greg-lunarg:
It is causing crashes for some drivers. Will try to re-enable it once
existing drivers are better able to deal with it.

### DIFF
--- a/source/opt/optimizer.cpp
+++ b/source/opt/optimizer.cpp
@@ -77,7 +77,8 @@ Optimizer& Optimizer::RegisterPerformancePasses() {
       .RegisterPass(CreateBlockMergePass())
       .RegisterPass(CreateLocalMultiStoreElimPass())
       .RegisterPass(CreateInsertExtractElimPass())
-      .RegisterPass(CreateCommonUniformElimPass())
+      // Currently causing driver crashes
+      // .RegisterPass(CreateCommonUniformElimPass())
       .RegisterPass(CreateDeadVariableEliminationPass());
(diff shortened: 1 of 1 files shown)


source/opt/optimizer.cpp
```diff
@@ -77,7 +77,8 @@ Optimizer& Optimizer::RegisterPerformancePasses() {
       .RegisterPass(CreateBlockMergePass())
       .RegisterPass(CreateLocalMultiStoreElimPass())
       .RegisterPass(CreateInsertExtractElimPass())
-      .RegisterPass(CreateCommonUniformElimPass())
+      // Currently causing driver crashes
```
dnovillo: Could you reference the issue? It's #946, IIRC.

source/opt/optimizer.cpp
```diff
@@ -92,7 +93,8 @@ Optimizer& Optimizer::RegisterSizePasses() {
...
```
dnovillo: Likewise here.  Thanks.

source/opt/optimizer.cpp
greg-lunarg: Done.

source/opt/optimizer.cpp
greg-lunarg: Done.

dnovillo:
Rebased, tested and pushed as commit d86c7ce808

PR #950, opened 2017-11-07
Add the IRContext (part 2): Add def-use manager

s-perron:
This change will move the instances of the def-use manager to the
IRContext.  This allows it to persists across optimization, and does
not have to be rebuilt multiple times.

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -233,8 +233,8 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
     const uint32_t mergeLabId =
         mergeInst->GetSingleWordInOperand(kSelectionMergeMergeBlockIdInIdx);
     AddBranch(liveLabId, *bi);
-    get_def_use_mgr()->KillInst(br);
-    get_def_use_mgr()->KillInst(mergeInst);
+    context()->KillInst(br);
+    context()->KillInst(mergeInst);
 
     modified = true;
 
@@ -331,8 +331,8 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
       }
       const uint32_t phiId = pii->result_id();
       KillNamesAndDecorates(phiId);
-      (void)get_def_use_mgr()->ReplaceAllUsesWith(phiId, replId);
-      get_def_use_mgr()->KillInst(&*pii);
+      (void)context()->ReplaceAllUsesWith(phiId, replId);
+      context()->KillInst(&*pii);
     }
   }
 
--- a/source/opt/def_use_manager.h
+++ b/source/opt/def_use_manager.h
@@ -49,8 +49,7 @@ class DefUseManager {
   // will be communicated to the outside via the given message |consumer|. This
   // instance only keeps a reference to the |consumer|, so the |consumer| should
   // outlive this instance.
-  DefUseManager(const MessageConsumer& consumer, ir::Module* module)
-      : consumer_(consumer) {
+  DefUseManager(ir::Module* module) {
     AnalyzeDefUse(module);
   }
 
@@ -88,29 +87,6 @@ class DefUseManager {
   // Returns the map from ids to their uses in instructions.
   const IdToUsesMap& id_to_uses() const { return id_to_uses_; }
 
-  // Turns the instruction defining the given |id| into a Nop. Returns true on
-  // success, false if the given |id| is not defined at all. This method also
-  // erases both the uses of |id| and the information of this |id|-generating
-  // instruction's uses of its operands.
-  bool KillDef(uint32_t id);
-  // Turns the given instruction |inst| to a Nop. This method erases the
-  // information of the given instruction's uses of its operands. If |inst|
-  // defines an result id, the uses of the result id will also be erased.
-  void KillInst(ir::Instruction* inst);
-  // Replaces all uses of |before| id with |after| id. Returns true if any
-  // replacement happens. This method does not kill the definition of the
-  // |before| id. If |after| is the same as |before|, does nothing and returns
-  // false.
-  bool ReplaceAllUsesWith(uint32_t before, uint32_t after);
-
- private:
-  using InstToUsedIdsMap =
-      std::unordered_map<const ir::Instruction*, std::vector<uint32_t>>;
-
-  // Analyzes the defs and uses in the given |module| and populates data
-  // structures in this class. Does nothing if |module| is nullptr.
-  void AnalyzeDefUse(ir::Module* module);
-
   // Clear the internal def-use record of the given instruction |inst|. This
   // method will update the use information of the operand ids of |inst|. The
   // record: |inst| uses an |id|, will be removed from the use records of |id|.
@@ -121,11 +97,17 @@ class DefUseManager {
   // Erases the records that a given instruction uses its operand ids.
   void EraseUseRecordsOfOperandIds(const ir::Instruction* inst);
 
-  const MessageConsumer& consumer_;  // Message consumer.
+ private:
+  using InstToUsedIdsMap =
+      std::unordered_map<const ir::Instruction*, std::vector<uint32_t>>;
+
+  // Analyzes the defs and uses in the given |module| and populates data
+  // structures in this class. Does nothing if |module| is nullptr.
+  void AnalyzeDefUse(ir::Module* module);
+
   IdToDefMap id_to_def_;             // Mapping from ids to their definitions
   IdToUsesMap id_to_uses_;           // Mapping from ids to their uses
-  // Mapping from instructions to the ids used in the instructions generating
-  // the result ids.
+  // Mapping from instructions to the ids used in the instruction.
   InstToUsedIdsMap inst_to_used_ids_;
 };
 
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -15,6 +15,7 @@
 #ifndef SPIRV_TOOLS_IR_CONTEXT_H
 #define SPIRV_TOOLS_IR_CONTEXT_H
 
+#include "def_use_manager.h"
 #include "module.h"
 
 #include <iostream>
@@ -24,7 +25,23 @@ namespace ir {
 
 class IRContext {
  public:
-  IRContext(std::unique_ptr<Module>&& m) : module_(std::move(m)) {}
+  enum Analysis {
+    kAnalysisNone = 0x0,
+    kAnalysisBegin = 0x1,
+    kAnalysisDefUse = kAnalysisBegin,
+    kAnalysisEnd = 0x2
+  };
+
+  friend inline Analysis operator|(Analysis lhs, Analysis rhs);
+  friend inline Analysis& operator|=(Analysis& lhs, Analysis rhs);
+  friend inline Analysis operator<<(Analysis a, int shift);
+  friend inline Analysis& operator<<=(Analysis& a, int shift);
+
+  IRContext(std::unique_ptr<Module>&& m, spvtools::MessageConsumer c)
+      : module_(std::move(m)),
+        consumer_(std::move(c)),
+        def_use_mgr_(nullptr),
+        valid_analyses_(kAnalysisNone) {}
   Module* module() const { return module_.get(); }
 
   inline void SetIdBound(uint32_t i);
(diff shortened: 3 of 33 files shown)


s-perron:
This is the first pass at moving the def-use manager to the IRContext.  I still have a few things to do:

- [ ] Add test for the IRContext.
- [ ] Add code for "Consistent" so we can identify when somebody failed to invalidate something.

I may leave the second part to another PR.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -331,8 +331,8 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
       }
       const uint32_t phiId = pii->result_id();
       KillNamesAndDecorates(phiId);
-      (void)get_def_use_mgr()->ReplaceAllUsesWith(phiId, replId);
-      get_def_use_mgr()->KillInst(&*pii);
+      (void)context()->ReplaceAllUsesWith(phiId, replId);
+      context()->KillInst(&*pii);
```
dnovillo: Out of curiosity.  What was the rationale you used to move some functions into IRContext vs the DefUse manager?  Is it instruction manipulation that goes in IRContext now?  And pure analysis stays in DefUse Manager?

source/opt/ir_context.h
```diff
@@ -121,8 +132,64 @@ class IRContext {
...
+
+  // Replaces all uses of |before| id with |after| id. Returns true if any
+  // replacement happens. This method does not kill the definition of the
+  // |before| id. If |after| is the same as |before|, does nothing and returns
+  // false.
+  bool ReplaceAllUsesWith(uint32_t before, uint32_t after);
+
  private:
   std::unique_ptr<Module> module_;
+  spvtools::MessageConsumer consumer_;
```
dnovillo: This is causing -Werror build failures because consumer_ is never used.  Perhaps add it later when you have uses of it?

source/opt/dead_branch_elim_pass.cpp
s-perron: Yes, functions that change the IR are part of the IRContext.  This way the def-use manager will not have to be aware of other analyses that are added in the future.

source/opt/ir_context.h
s-perron: It is being used.  It is returned by the "consumer()" member function, which is used to pass the consumer on to the def-use manager when it is being built (ir_context.h line 144).

s-perron: Looking at it closer, it was "consumer_" in the def use manger that was causing the problem.  The uses of it were moved to the IRContext.

I've removed it.

source/opt/dead_branch_elim_pass.cpp
dnovillo: Sounds good.

source/opt/ir_context.h
```diff
@@ -121,8 +132,64 @@ class IRContext {
   // Appends a function to this module.
   inline void AddFunction(std::unique_ptr<Function>&& f);
 
+  // Returns a pointer to a def-use manager.  If the def-use manager is
+  // invalid, it is rebuilt first.
+  opt::analysis::DefUseManager* get_def_use_mgr() {
+    if (!AreAnalysesValid(kAnalysisDefUse)) BuildDefUseManager();
```
dnovillo: Suggest braces around single-line bodies.

source/opt/ir_context.h
```diff
@@ -121,8 +132,64 @@ class IRContext {
...
+  // invoked every time there is a message to be communicated to the outside.
+  void SetMessageConsumer(spvtools::MessageConsumer c) {
+    consumer_ = std::move(c);
+  }
+
+  // Returns the reference to the message consumer for this pass.
+  const spvtools::MessageConsumer& consumer() const { return consumer_; }
+
+  // Rebuilds the analyses in |set| that are invalid.
+  void BuildAnalyses(Analysis set);
```
dnovillo: Slight semantic confusion.  BuildDefUseManager() is a forced build, but BuildAnalyses() is not.  Perhaps rename BuildAnalyses() to BuildInvalidAnalyses()?

source/opt/ir_context.h
```diff
@@ -121,8 +132,64 @@ class IRContext {
...
+
+  // Invalidates all of the analyses except for those in |preserved_analyses|.
+  void InvalidateAnalyses(Analysis preserved_analyses);
+
+  // Returns true if all of the valid analyses are consistent with module.
+  bool Consistent() { return true; }
```
dnovillo: Shouldn't this check the state of kAnalysisDefUse?  Also, it's a bit unclear what 'consistent' means here.  Does it mean that all the analyses are in a valid state?

source/opt/ir_context.h
dnovillo: Ah, yes.  It was the one in DefUseManager.  Thanks.

source/opt/pass.cpp
```diff
@@ -106,6 +102,17 @@ bool Pass::ProcessCallTreeFromRoots(
   return modified;
 }
 
+Pass::Status Pass::Run(ir::IRContext* ctx) {
+  Pass::Status status = Process(ctx);
```
dnovillo: Note for subsequent PRs.  This is where we'd move the call to Pass::InitializeProcessing(), right?

source/opt/ir_context.h
s-perron: Still have to implement this function.  It is just a stub right now.  It will be used only for debugging.  I possibly add calls to it in assert to make sure someone changed the code without invalidating the analysis.

source/opt/pass.cpp
s-perron: Yes it is.

source/opt/ir_context.h
dneto0: Or GetAnalyses, with method comment that it will reuse valid analyses, and rebuild invalid analysis.

source/opt/def_use_manager.h
```diff
@@ -121,7 +97,14 @@ class DefUseManager {
...
+  using InstToUsedIdsMap =
+      std::unordered_map<const ir::Instruction*, std::vector<uint32_t>>;
+
+  // Analyzes the defs and uses in the given |module| and populates data
+  // structures in this class. Does nothing if |module| is nullptr.
+  void AnalyzeDefUse(ir::Module* module);
+
   IdToDefMap id_to_def_;             // Mapping from ids to their definitions
   IdToUsesMap id_to_uses_;           // Mapping from ids to their uses
   // Mapping from instructions to the ids used in the instructions generating
```
dneto0: Not part of this CL, but I'm confused by this comment.

source/opt/ir_context.h
```diff
@@ -121,8 +133,61 @@ class IRContext {
...
+  }
+
+  // Returns the reference to the message consumer for this pass.
+  const spvtools::MessageConsumer& consumer() const { return consumer_; }
+
+  // Rebuilds the analyses in |set| that are invalid.
+  void BuildAnalyses(Analysis set);
+
+  // Invalidates all of the analyses except for those in |preserved_analyses|.
+  void InvalidateAnalyses(Analysis preserved_analyses);
```
dneto0: The reversal of sense will look really confusing when you don't have the method comment fresh in your mind.

Suggest InvalidateAnalysesExcepting  or InvalidateAnalysesExceptFor

source/opt/ir_context.h
```diff
@@ -121,8 +133,61 @@ class IRContext {
...
+  // invalid, it is rebuilt first.
+  opt::analysis::DefUseManager* get_def_use_mgr() {
+    if (!AreAnalysesValid(kAnalysisDefUse)) BuildDefUseManager();
+    return def_use_mgr_.get();
+  }
+
+  // Builds the def-use manager from scratch, even if it was already valid.
+  void BuildDefUseManager() {
+    def_use_mgr_.reset(new opt::analysis::DefUseManager(module()));
+    valid_analyses_ = (Analysis)(valid_analyses_ | kAnalysisDefUse);
```
dneto0: Prefer static_cast over c-style cast.

The masking and casting is ok, but we could also make an operator|  overload for this purpose.

source/opt/ir_context.h
```diff
@@ -121,8 +133,61 @@ class IRContext {
...
+  // Returns the reference to the message consumer for this pass.
+  const spvtools::MessageConsumer& consumer() const { return consumer_; }
+
+  // Rebuilds the analyses in |set| that are invalid.
+  void BuildAnalyses(Analysis set);
+
+  // Invalidates all of the analyses except for those in |preserved_analyses|.
+  void InvalidateAnalyses(Analysis preserved_analyses);
+
+  // Turns the instruction defining the given |id| into a Nop. Returns true on
```
dneto0: For the long term:  Since we now have easy deletion of instructions, should we change this to mean deletion of the instruction?  Changing it to no-op was an artefact of not wanting to manipulate the vector of instructions.

source/opt/pass.h
```diff
@@ -111,10 +111,18 @@ class Pass {
...
 
-  // Processes the given |module|. Returns Status::Failure if errors occur when
+
+  // Run the pass on the given |module|. Returns Status::Failure if errors occur when
   // processing. Returns the corresponding Status::Success if processing is
-  // succesful to indicate whether changes are made to the module.
-  virtual Status Process(ir::IRContext* context) = 0;
+  // succesful to indicate whether changes are made to the module.  If there
+  // were any changes it will also invalidate the analyses in the IRContext
+  // that are not preserved.
```

source/opt/pass.h
```diff
@@ -111,10 +111,18 @@ class Pass {
...
+  virtual Status Run(ir::IRContext* ctx) final;
+
+  // Returns the set of analyses that the pass is guanreteed to preserve.
```
dneto0: guaranteed

source/opt/ir_context.h
```diff
@@ -24,7 +25,18 @@ namespace ir {
 
 class IRContext {
  public:
-  IRContext(std::unique_ptr<Module>&& m) : module_(std::move(m)) {}
+  enum Analysis {
+    kAnalysisNone = 0x0,
+    kAnalysisFirst = 0x1,
```
dneto0: Quibbles:
 
Seems kAnalysisFirst should be kAnalysisBegin = 0x1, and then kAnalysisDefUse = kAnalysisBegin

Also  kAnalysisLast should be renamed kAnalysisEnd  ?

dneto0:
Thanks for doing this change!  This is a really nice cleanup, and it has been long on our list of things to do.

dneto0:
I assume a future MR will decorate the various existing passes to indicate which analyses they preserve.  That's too much for this CL, which is primarily a refactor.     The most interesting passes are the mem2reg style passes.

source/opt/ir_context.h
s-perron: Yes, I looked into doing that.  The problem is that current code is written assuming the current instruction iterator is not invalidated after calling killing the instruction.  To fix that, we need to rewrite that code.  I did not want to muddy the waters in this CL by doing that as well.

source/opt/def_use_manager.h
s-perron: Fixed up.  From other comments I think it was originally a map from and id to the ids used in the instruction that defines the id.

source/opt/ir_context.h
s-perron: done

source/opt/ir_context.h
s-perron: done

source/opt/ir_context.h
s-perron: Defining the operator is a good idea.  I did that.

source/opt/ir_context.h
s-perron: I went with BuildInvalidAnalyses.  The "GetAnalyes" might make people think something is being returned, when it is not.

source/opt/ir_context.h
s-perron: Done.  Went with InvalidateAnalysesExceptFor.

source/opt/ir_context.h
dnovillo: +1

dnovillo:
@dneto0 this LGTM to push and we need it in before #952 can land.  @s-perron seems to have addressed all your feedback.  If you have more, I propose addressing them in another PR.

s-perron:
rebase, squashed and merged at f32d11f.

PR #952, opened 2017-11-08
Re-format files in source, source/opt, source/util, source/val and tools

dnovillo:
NFC. This just makes sure every file is formatted following the
formatting definition in .clang-format.


dnovillo:
Re-based and pushed as d2938e48427cb6e8d5996712c23496d62e3c08d1

PR #956, opened 2017-11-09
Tests: Add optional dependency on Effcee stateful matcher

dneto0:
Add Effcee as an optional dependency for use in tests.  In future it will
be a required dependency.

Effcee is a stateful pattern matcher that has much of the functionality
of LLVM's FileCheck, except in library form.  Effcee makes it much easier
to write tests for optimization passes.

Demonstrate its use in a test for the strength-reduction pass.

Update README.md with example commands of how to get sources.

Update Appveyor and Travis-CI build rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,8 +173,8 @@ We intend to maintain a linear history on the GitHub `master` branch.
   [googletest][googletest] sources, not provided
 * `external/effcee`: Location of [Effcee][effcee] sources, if the `effcee` library
   is not already configured by an enclosing project.
-* `external/effcee/third_party/re2`: Location of [RE2][re2] sources, if the
-  `effcee` library is not already configured by an enclosing project.
+* `external/re2`: Location of [RE2][re2] sources, if the `effcee` library is not already
+  configured by an enclosing project.
   (The Effcee project already requires RE2.)
 * `include/`: API clients should add this directory to the include search path
 * `external/spirv-headers`: Intended location for
@@ -190,7 +190,7 @@ Example of getting sources, assuming SPIRV-Tools is configured as a standalone p
     git clone https://github.com/KhronosGroup/SPIRV-Headers.git spirv-tools/external/spirv-headers
     git clone https://github.com/google/googletest.git          spirv-tools/external/googletest
     git clone https://github.com/google/effcee.git              spirv-tools/external/effcee
-    git clone https://github.com/google/re2.git                 spirv-tools/external/effcee/third_party/re2
+    git clone https://github.com/google/re2.git                 spirv-tools/external/re2
 
 ### Tests
 
@@ -218,7 +218,7 @@ Effcee itself depends on [RE2][re2].
 * If SPIRV-Tools is configured as part of a larger project that already uses
   Effcee, then that project should include Effcee before SPIRV-Tools.
 * Otherwise, SPIRV-Tools expects Effcee sources to appear in `external/effcee`
-  and RE2 sources to appear in `external/effcee/third_party/re2`.
+  and RE2 sources to appear in `external/re2`.
 
 Currently Effcee is an optional dependency, but soon it will be required.
 
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ Some tests depend on the [Effcee][effcee] library for stateful matching.
 Effcee itself depends on [RE2][re2].
 
 * If SPIRV-Tools is configured as part of a larger project that already uses
-  Effcee, then that project should include Effcee before SPIRV-Toolls.
+  Effcee, then that project should include Effcee before SPIRV-Tools.
 * Otherwise, SPIRV-Tools expects Effcee sources to appear in `external/effcee`
   and RE2 sources to appear in `external/effcee/third_party/re2`.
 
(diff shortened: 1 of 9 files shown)


dneto0:
This enables a significant change in how we test.  We can use more expressive pattern-matching, just like FileCheck in LLVM.  I included an example.

It  *also* makes building tests more heavyweight since we add two libraries to each test executable.

README.md
```diff
@@ -197,6 +210,18 @@ tests:
 The fix is included on the googletest master branch any time after 2015-11-10.
 In particular, googletest must be newer than version 1.7.0.
 
+### Optional dependency on Effcee
+
+Some tests depend on the [Effcee][effcee] library for stateful matching.
+Effcee itself depends on [RE2][re2].
+
+* If SPIRV-Tools is configured as part of a larger project that already uses
+  Effcee, then that project should include Effcee before SPIRV-Toolls.
```
dnovillo: s/SPIRV-Toolls/SPIRV-Tools/

test/opt/strength_reduction_test.cpp
```diff
@@ -63,36 +63,45 @@ TEST_F(StrengthReductionBasicTest, BasicReplaceMulBy8) {
   EXPECT_THAT(output, HasSubstr("OpShiftLeftLogical %uint %uint_5 %uint_3"));
 }
 
-// Test to make sure we replace 16*5.
+// TODO(dneto): Add Effcee as requried dependency, and make this unconditional.
```
dnovillo: s/requried/required/

test/opt/strength_reduction_test.cpp
```diff
@@ -63,36 +63,45 @@ TEST_F(StrengthReductionBasicTest, BasicReplaceMulBy8) {
...
+               OpCapability Shader
+          %1 = OpExtInstImport "GLSL.std.450"
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint Vertex %main "main"
+               OpName %main "main"
+       %void = OpTypeVoid
+          %4 = OpTypeFunction %void
+; We know disassembly will produce %uint here, but
+;  CHECK: %uint = OpTypeInt 32 0
+;  CHECK-DAG: [[five:%[a-zA-Z_\d]+]] = OpConstant %uint 5
```
dnovillo: YAY!

test/opt/strength_reduction_test.cpp
dnovillo: Also, I'm wondering if we could somehow make the harness execute tests marked as EFFCEE so that the test itself can run unconditionally when EFFCEE is enabled.  If that's too annoying, that's fine.  Adding #ifdefs is certainly not a big deal.

dnovillo: Oh, and of course, our bots now have to run with and without EFFCEE enabled...

README.md
```diff
@@ -179,6 +184,14 @@ We intend to maintain a linear history on the GitHub `master` branch.
 * `test/`: Tests, using the [googletest][googletest] framework
 * `tools/`: Command line executables
 
+Example of getting sources, assuming SPIRV-Tools is configured as a standalone project:
+
+    git clone https://github.com/KhronosGroup/SPIRV-Tools.git   spirv-tools
+    git clone https://github.com/KhronosGroup/SPIRV-Headers.git spirv-tools/external/spirv-headers
+    git clone https://github.com/google/googletest.git          spirv-tools/external/googletest
+    git clone https://github.com/google/effcee.git              spirv-tools/external/effcee
+    git clone https://github.com/google/re2.git                 spirv-tools/external/effcee/third_party/re2
```
antiagainst: I think we can checkout re2 into `external/` directly?

external/CMakeLists.txt
```diff
@@ -54,4 +54,26 @@ if (NOT ${SPIRV_SKIP_EXECUTABLES})
       set_property(TARGET ${target} PROPERTY FOLDER GoogleTest)
     endforeach()
   endif()
+
```
antiagainst: We can check and `add_subdirectory` re2 here so that we can checkout re2 into `external/`.

test/opt/strength_reduction_test.cpp
dneto0: My plan is to do a 3-step soft launch:
- add as conditional dependency now
- notify downstream projects so they can react, e.g. Android NDK, and LunarG
- make it a required dependency

If it's just a few weeks between steps 1 and 3 then I think we can skip running bots having Effcee off.

README.md
dneto0: Yes, we should be able to do that.  I'll make that change.

dneto0:
The bot logs showed effcee as not being configured.  That's wrong.  I'll fix that too.

dneto0:
I want to see the bots exercise the effcee tests and *not* the RE2 tests.

dneto0:
Travis bot failed for the Android build.  That's because even in Android mode, we say we don't want tests, but we were still trying to add the re2 and effcee libraries as dependencies, and they don't have Android.mk support.

I added a commit to stop depending on effcee and re2 when we've requested option SPIRV_SKIP_TESTS. That condition is implied when we ask for SPIRV_SKIP_EXECUTABLES.

dneto0:
Bots pass and show it is running effcee-test and *not* running the RE2 tests.  That's intended.

I'll squash, rebase if necessary, then push into master.

dneto0:
Squashed and pushed into master as 76555bd4ba

PR #957, opened 2017-11-09
Have all MemPasses preserve the def-use manager.

s-perron:
Originally the passes that extended from MemPass were those that are
of the def-use manager.  I am assuming they would be able to preserve
it because of that.

The IRContext relies on the pass to tell it if something is invalidated.
It is possible that the pass lied.  To help identify those situations,
we will check if the valid analyses are correct after each pass.

This will be enabled by default for the debug build, and disabled in the
production build.  It can be disabled in the debug build by adding
"-DSPIRV_CHECK_CONTEXT=OFF" to the cmake command.

This is part of #608.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,6 +173,13 @@ if ("${SPIRV_SKIP_EXECUTABLES}")
   set(SPIRV_SKIP_TESTS ON)
 endif()
 
+# Defaults to ON.  The checks can be time consuming.
+# Turn off if they take too long.
+option(SPIRV_CHECK_CONTEXT "In a debug build, check if the IR context is in a valid state." ON)
+if (${SPIRV_CHECK_CONTEXT})
+  add_definitions(-DSPIRV_CHECK_CONTEXT)
+endif()
+
 add_subdirectory(external)
 
 add_subdirectory(source)
--- a/source/opt/cfg_cleanup_pass.h
+++ b/source/opt/cfg_cleanup_pass.h
@@ -28,6 +28,10 @@ class CFGCleanupPass : public MemPass {
   const char* name() const override { return "cfg-cleanup"; }
   Status Process(ir::IRContext* c) override;
 
+  ir::IRContext::Analysis GetPreservedAnalyses() override {
+    return ir::IRContext::kAnalysisDefUse;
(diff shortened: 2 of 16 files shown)


source/opt/cfg_cleanup_pass.h
```diff
@@ -28,6 +28,10 @@ class CFGCleanupPass : public MemPass {
   const char* name() const override { return "cfg-cleanup"; }
   Status Process(ir::IRContext* c) override;
 
+  ir::IRContext::Analysis GetPreservedAnalyses() override {
+    return ir::IRContext::kAnalysisDefUse;
```
dnovillo: Is this right?  CFGCleanupPass should not preserve DefUse analysis.

CMakeLists.txt
```diff
@@ -173,6 +173,11 @@ if ("${SPIRV_SKIP_EXECUTABLES}")
   set(SPIRV_SKIP_TESTS ON)
 endif()
 
+option(SPIRV_CHECK_CONTEXT "In a debug build, check if the IR context is in a valid state." ON)
```
dnovillo: Why not all the time?

source/opt/mem_pass.cpp
```diff
@@ -755,7 +755,9 @@ void MemPass::RemovePhiOperands(
     i += 2;
   }
 
+  context()->ForgetUses(phi);
   phi->ReplaceOperands(keep_operands);
+  context()->AnalyzeUses(phi);
```
dnovillo: Neat.  I think this is what's causing my segfaults in https://github.com/dnovillo/SPIRV-Tools/tree/instr-to-block

CMakeLists.txt
s-perron: I can be an expensive check.  So I would like to allow developers to turn it off if they want.  It is on by default for the debug build.

s-perron:
The failure with the Android build does not look like it is related to my change.  It seems like a directory was temporarily unavailable, which caused the error.

CMakeLists.txt
dnovillo: Ah, OK.  Could you document why it's an option.  LGTM with that change.

Issue #857, opened 2017-10-05
Add merge-return pass


dneto0:
Transform each  reachable function so that:
- it has at most one basic block with an OpReturn or OpReturnValue
- that basic block post-dominates all reachable blocks (modulo those ending with OpKill)

Expose a general purpose helper function that does this transform on a single SPIR-V function.

Motivated by https://github.com/KhronosGroup/SPIRV-Tools/issues/850

cwabbott0:
Btw, we already have a pass to do this in Mesa: https://cgit.freedesktop.org/mesa/mesa/tree/src/compiler/nir/nir_lower_returns.c -- feel free to steal the algorithm!

Note that what it's doing is more complicated than just wrapping the entire thing in a loop, and should generate better code in most cases, at least on Intel.

s-perron:
@alan-baker is working on this.

dneto0:
@s-perron is working on this

PR #958, opened 2017-11-09
Merge return pass Issue:857

alan-baker:
Adds a merge return pass #857 .

Unifies all return in a function into a single return location using a new basic block and (if necessary) new phi. Does not perform any other optimizations.

### DIFF
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -43,6 +43,7 @@ add_library(SPIRV-Tools-opt
   local_single_store_elim_pass.h
   local_ssa_elim_pass.h
   log.h
+  merge_return_pass.h
   module.h
   null_pass.h
   reflect.h
@@ -88,6 +89,7 @@ add_library(SPIRV-Tools-opt
   local_single_block_elim_pass.cpp
   local_single_store_elim_pass.cpp
   local_ssa_elim_pass.cpp
+  merge_return_pass.cpp
   module.cpp
   eliminate_dead_functions_pass.cpp
   remove_duplicates_pass.cpp
--- a/source/opt/merge_return_pass.cpp
+++ b/source/opt/merge_return_pass.cpp
@@ -27,18 +27,20 @@ Pass::Status MergeReturnPass::Process(ir::IRContext *irContext) {
 
   bool modified = false;
   for (auto &function : *get_module()) {
-    std::vector<ir::BasicBlock*> returnBlocks = collectReturnBlocks(&function);
+    std::vector<ir::BasicBlock *> returnBlocks = collectReturnBlocks(&function);
     modified |= mergeReturnBlocks(&function, returnBlocks);
   }
 
   return modified ? Status::SuccessWithChange : Status::SuccessWithoutChange;
 }
 
-std::vector<ir::BasicBlock*> MergeReturnPass::collectReturnBlocks(ir::Function *function) {
-  std::vector<ir::BasicBlock*> returnBlocks;
+std::vector<ir::BasicBlock *> MergeReturnPass::collectReturnBlocks(
+    ir::Function *function) {
+  std::vector<ir::BasicBlock *> returnBlocks;
   for (auto &block : *function) {
     ir::Instruction &terminator = *block.tail();
-    if (terminator.opcode() == SpvOpReturn || terminator.opcode() == SpvOpReturnValue) {
+    if (terminator.opcode() == SpvOpReturn ||
+        terminator.opcode() == SpvOpReturnValue) {
       returnBlocks.push_back(&block);
     }
   }
@@ -46,18 +48,21 @@ std::vector<ir::BasicBlock*> MergeReturnPass::collectReturnBlocks(ir::Function *
   return std::move(returnBlocks);
 }
 
-bool MergeReturnPass::mergeReturnBlocks(ir::Function *function, const std::vector<ir::BasicBlock*> &returnBlocks) {
+bool MergeReturnPass::mergeReturnBlocks(
+    ir::Function *function, const std::vector<ir::BasicBlock *> &returnBlocks) {
   if (returnBlocks.size() <= 1) {
     // No work to do.
     return false;
   }
 
   // Create a label for the new return block
-  std::unique_ptr<ir::Instruction> returnLabel(new ir::Instruction(SpvOpLabel, 0u, TakeNextId(), {}));
+  std::unique_ptr<ir::Instruction> returnLabel(
+      new ir::Instruction(SpvOpLabel, 0u, TakeNextId(), {}));
   uint32_t returnId = returnLabel->result_id();
 
   // Create the new basic block
-  std::unique_ptr<ir::BasicBlock> returnBlockUPtr(new ir::BasicBlock(std::move(returnLabel)));
+  std::unique_ptr<ir::BasicBlock> returnBlockUPtr(
+      new ir::BasicBlock(std::move(returnLabel)));
   function->AddBasicBlock(std::move(returnBlockUPtr));
   ir::BasicBlock &returnBlock = *(--function->end());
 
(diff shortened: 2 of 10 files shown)


source/opt/CMakeLists.txt
```diff
@@ -88,6 +89,7 @@ add_library(SPIRV-Tools-opt
   local_single_block_elim_pass.cpp
   local_single_store_elim_pass.cpp
   local_ssa_elim_pass.cpp
+  merge_return_pass.cpp
```
dneto0: Need a corresponding change to Android.mk, to support Android builds via ndk-build.

source/opt/merge_return_pass.h
```diff
@@ -0,0 +1,50 @@
...
+// Documented in optimizer.hpp
+class MergeReturnPass : public Pass {
+ public:
+  const char *name() const override { return "merge-return-pass"; }
+  Status Process(ir::IRContext *) override;
+
+ private:
+  // Returns all BasicBlocks terminated by OpReturn or OpReturnValue in
+  // |function|.
+  std::vector<ir::BasicBlock *> collectReturnBlocks(ir::Function *function);
```
dneto0: style guide:  Name of nontrivial methods should start with capital letter.  (e.g. anything that isn't a trivial accessor)

source/opt/merge_return_pass.cpp
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  return modified ? Status::SuccessWithChange : Status::SuccessWithoutChange;
+}
+
+std::vector<ir::BasicBlock *> MergeReturnPass::collectReturnBlocks(
+    ir::Function *function) {
+  std::vector<ir::BasicBlock *> returnBlocks;
+  for (auto &block : *function) {
+    ir::Instruction &terminator = *block.tail();
```
dneto0: I think our format puts the & against the type, not the variable.
Did this get clang-formatted?

Super minor!

source/opt/merge_return_pass.cpp
```diff
@@ -0,0 +1,117 @@
...
+    return false;
+  }
+
+  // Create a label for the new return block
+  std::unique_ptr<ir::Instruction> returnLabel(
+      new ir::Instruction(SpvOpLabel, 0u, TakeNextId(), {}));
+  uint32_t returnId = returnLabel->result_id();
+
+  // Create the new basic block
+  std::unique_ptr<ir::BasicBlock> returnBlockUPtr(
```
dneto0: Style:  I'd just call this returnBlock.   Generally I'm not a fan of embedding type information in the name.    Especially in this case, the scope of the variable is small enough that it's easy to find the original definition any time you need to (or just remember it).

source/opt/merge_return_pass.cpp
```diff
@@ -0,0 +1,117 @@
...
+      new ir::BasicBlock(std::move(returnLabel)));
+  function->AddBasicBlock(std::move(returnBlockUPtr));
+  ir::BasicBlock &returnBlock = *(--function->end());
```
dneto0: Seems a bit baroque to pack so much into a single line...  But ok.

source/opt/merge_return_pass.cpp
```diff
@@ -0,0 +1,117 @@
...
+      phiOps.push_back(
+          {SPV_OPERAND_TYPE_ID, {block->tail()->GetSingleWordInOperand(0u)}});
+      phiOps.push_back({SPV_OPERAND_TYPE_ID, {block->id()}});
+    }
+  }
+
+  if (!phiOps.empty()) {
+    // Need a PHI node to select the correct return value.
+    uint32_t phiResultId = TakeNextId();
+    uint32_t phiTypeId = function->type_id();
```
dneto0: This is a sad irregularity of SPIR-V.   It's the *ONE* place where the type of the thing is not the same as the typeid of the thing.  :-(

test/opt/pass_merge_return_test.cpp
```diff
@@ -0,0 +1,260 @@
...
+OpMemoryModel Logical GLSL450
+%1 = OpTypeInt 32 0
+%2 = OpTypeBool
+%3 = OpConstantFalse %2
+%4 = OpConstant %1 0
+%5 = OpConstant %1 1
+%6 = OpTypeFunction %1
+%7 = OpFunction %1 None %6
+%8 = OpLabel
+OpBranchConditional %3 %9 %10
```
dneto0: Shader capability requires structured control flow.  
Seems it's missing OpSelectionMerge.  Does this pass validation? 

Either OpSelectionMerge as required to properly validate, or change Shader to Kernel.

source/opt/merge_return_pass.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (terminator.opcode() == SpvOpReturn ||
+        terminator.opcode() == SpvOpReturnValue) {
+      returnBlocks.push_back(&block);
+    }
+  }
+
+  return std::move(returnBlocks);
+}
+
+bool MergeReturnPass::mergeReturnBlocks(
```
dneto0: This is the right algorithm when structured control flow is not required.  If structured control flow is required, then you have to do a lot more work:  From within nested control flow, you have to branch out to the enclosing merge block; and you'll need to track whether you're on the "escape" path, via a bool flag.  It's complicated.... :-)

dneto0:
Suggest you take a look at Mesa's algorithm.  It does the structured breakout. 
See @cwabbot0's comment: 
https://github.com/KhronosGroup/SPIRV-Tools/issues/857#issuecomment-334952559

source/opt/merge_return_pass.cpp
alan-baker: I'm attempting to get the return type of the function here. Function::type_id() returns the type id of the defining instruction (OpFunction). Is this not correct?

source/opt/merge_return_pass.cpp
alan-baker: I did run clang format. It does the opposite of your expectation.

source/opt/merge_return_pass.cpp
alan-baker: I'll change this and similar places to use iterator instead. Hopefully, the overall readability increases.

test/opt/pass_merge_return_test.cpp
alan-baker: I'll change this and add a test to do nothing in the presence of Shader capability for now.

source/opt/merge_return_pass.cpp
alan-baker: Changed

source/opt/CMakeLists.txt
alan-baker: Thanks, fixed.

source/opt/merge_return_pass.h
alan-baker: Fixed

source/opt/merge_return_pass.cpp
alan-baker: Apparently I was running this incorrectly. Will fix.

source/opt/merge_return_pass.cpp
dneto0: Your code is correct.   It's just an irregularity of the spec.    The result id of the OpFunction is an OpFunction.  But the type id is the type of the value returned by the function, not type of the function.
Never mind. It just bugs me.

tools/opt/opt.cpp
```diff
@@ -155,6 +155,9 @@ NOTE: The optimizer is a work in progress.
   --merge-return
                Replace all return instructions with unconditional branches to
                a new basic block containing an unified return.
+
+               This pass does not currently support structured control flow. It
+               bails out if the shader capability is detected.
```
dneto0: "Bail out" is colloquial and ambiguous: does it result with success or failure?

Suggest replacing "bails out" with "does nothing" 
(Also in the method comments talking about bailing out).

dneto0:
Rebased, squashed, and pushed into master as a92d69b43

source/opt/merge_return_pass.cpp
```diff
@@ -0,0 +1,120 @@
...
+  return modified ? Status::SuccessWithChange : Status::SuccessWithoutChange;
+}
+
+std::vector<ir::BasicBlock*> MergeReturnPass::CollectReturnBlocks(
+    ir::Function* function) {
+  std::vector<ir::BasicBlock*> returnBlocks;
+  for (auto& block : *function) {
+    ir::Instruction& terminator = *block.tail();
+    if (terminator.opcode() == SpvOpReturn ||
+        terminator.opcode() == SpvOpReturnValue) {
```
dnovillo: Change to if (block.IsReturn())

PR #961, opened 2017-11-10
Add analysis to compute mappings between instructions and basic blocks.

dnovillo:
Note that this is currently running into some testsuite failures.  I believe these should be fixed by #957.

### DIFF
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -32,17 +32,17 @@ class IRContext {
   // 1. Enum values should be powers of 2. These are cast into uint32_t
   //    bitmasks, so we can have at most 31 analyses represented.
   //
-  // 2. Make sure it gets invalidated by IRContext methods that add or remove IR
-  //    elements (e.g., KillDef, KillInst, ReplaceAllUsesWith).
+  // 2. Make sure it gets invalidated or preserved by IRContext methods that add
+  //    or remove IR elements (e.g., KillDef, KillInst, ReplaceAllUsesWith).
   //
-  // 4. Add handling code in BuildInvalidAnalyses and
+  // 3. Add handling code in BuildInvalidAnalyses and
   //    InvalidateAnalysesExceptFor.
   enum Analysis {
     kAnalysisNone = 0 << 0,
     kAnalysisBegin = 1 << 0,
     kAnalysisDefUse = kAnalysisBegin,
     kAnalysisInstrToBlockMapping = 1 << 1,
-    kAnalysisEnd = kAnalysisInstrToBlockMapping
+    kAnalysisEnd = 1 << 2
   };
 
   friend inline Analysis operator|(Analysis lhs, Analysis rhs);
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -25,11 +25,24 @@ namespace ir {
 
 class IRContext {
  public:
+  // Available analyses.
+  //
+  // When adding a new analysis:
+  //
+  // 1. Enum values should be powers of 2. These are cast into uint32_t
+  //    bitmasks, so we can have at most 31 analyses represented.
+  //
+  // 2. Make sure it gets invalidated by IRContext methods that add or remove IR
+  //    elements (e.g., KillDef, KillInst, ReplaceAllUsesWith).
+  //
+  // 4. Add handling code in BuildInvalidAnalyses and
+  //    InvalidateAnalysesExceptFor.
   enum Analysis {
-    kAnalysisNone = 0x0,
-    kAnalysisBegin = 0x1,
+    kAnalysisNone = 0 << 0,
+    kAnalysisBegin = 1 << 0,
     kAnalysisDefUse = kAnalysisBegin,
-    kAnalysisEnd = 0x2
+    kAnalysisInstrToBlockMapping = 1 << 1,
+    kAnalysisEnd = kAnalysisInstrToBlockMapping
   };
 
   friend inline Analysis operator|(Analysis lhs, Analysis rhs);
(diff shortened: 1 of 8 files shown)


source/opt/ir_context.h
```diff
@@ -25,10 +25,20 @@ namespace ir {
...
+  // 1. Make sure it gets invalidated by IRContext methods that add or remove IR
+  //    elements (e.g., KillDef, KillInst, ReplaceAllUsesWith).
+  //
+  // 2. Add handling code in BuildInvalidAnalyses and
+  //    InvalidateAnalysesExceptFor.
   enum Analysis {
     kAnalysisNone = 0x0,
     kAnalysisBegin = 0x1,
     kAnalysisDefUse = kAnalysisBegin,
+    kAnalysisInstrToBlockMapping,
```
s-perron: Since we are going up by powers of 2,should explicitly set to value to 0x2?

source/opt/ir_context.h
```diff
@@ -25,10 +25,20 @@ namespace ir {
...
     kAnalysisEnd = 0x2
```
s-perron: This is suppose to be 1 past the end, so it should become 0x4.

test/opt/def_use_test.cpp
```diff
@@ -1318,7 +1318,9 @@ TEST_P(KillInstTest, Case) {
       BuildModule(SPV_ENV_UNIVERSAL_1_1, nullptr, tc.before);
   ASSERT_NE(nullptr, module);
   ir::IRContext context(std::move(module), spvtools::MessageConsumer());
-  context.BuildDefUseManager();
+
+  // Force a re-build of def-use manager the next time it's requested.
+  context.InvalidateAnalyses(ir::IRContext::Analysis::kAnalysisDefUse);
```
s-perron: this is not testing that KillInst is updating the def-use manger anymore.  We are suppose to build it and have it valid for the call to KillInst.  Then we check the chains at the end.  Now it is building it at the end.

source/opt/ir_context.h
```diff
@@ -25,10 +25,20 @@ namespace ir {
...
  public:
+  // Available analyses.
+  //
+  // When adding a new analysis:
+  //
+  // 1. Make sure it gets invalidated by IRContext methods that add or remove IR
+  //    elements (e.g., KillDef, KillInst, ReplaceAllUsesWith).
+  //
+  // 2. Add handling code in BuildInvalidAnalyses and
+  //    InvalidateAnalysesExceptFor.
```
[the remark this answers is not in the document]
dnovillo: Oh!  Of course they have to be powers of 2... silly me.

source/opt/ir_context.h
dnovillo: Yup.

source/opt/ir_context.h
dnovillo: Done.

test/opt/def_use_test.cpp
dnovillo: Ah, yeah.  If I add a call to get_def_use_mgr(), I would restore the original semantics of the test, right?

dnovillo:
I've made the requested changes.  It's still not ready to submit (waiting for #957 to land).

PTAL.

dnovillo:
I've fixed the failures.  I forgot that some instructions will never belong to any basic block.  The method get_instr_block() was blindly de-referencing the map element without checking that.

All tests should pass now.

@s-perron, PTAL.

source/opt/ir_context.h
```diff
@@ -25,11 +25,24 @@ namespace ir {
...
+  //
+  // When adding a new analysis:
+  //
+  // 1. Enum values should be powers of 2. These are cast into uint32_t
+  //    bitmasks, so we can have at most 31 analyses represented.
+  //
+  // 2. Make sure it gets invalidated by IRContext methods that add or remove IR
+  //    elements (e.g., KillDef, KillInst, ReplaceAllUsesWith).
+  //
+  // 4. Add handling code in BuildInvalidAnalyses and
```
s-perron: You missed point 3. :)

source/opt/ir_context.h
```diff
@@ -25,11 +25,24 @@ namespace ir {
...
+  //    InvalidateAnalysesExceptFor.
   enum Analysis {
-    kAnalysisNone = 0x0,
-    kAnalysisBegin = 0x1,
+    kAnalysisNone = 0 << 0,
+    kAnalysisBegin = 1 << 0,
     kAnalysisDefUse = kAnalysisBegin,
-    kAnalysisEnd = 0x2
+    kAnalysisInstrToBlockMapping = 1 << 1,
+    kAnalysisEnd = kAnalysisInstrToBlockMapping
```
s-perron: I had used kAnalysisEnd as "one-past-the-end".  See the IRContext test.  They do 

 for (Analysis i = IRContext::kAnalysisBegin; i < IRContext::kAnalysisEnd; i <<= 1)

I don't care which way it is, but it something needs to be changed.

dnovillo: Ah, yes.  One past the end seems more intuitive.  Fixed.

Also fixed the test AllPreserveFirstOnlyAfterPassWithChange (thanks for the hints!)

source/opt/ir_context.h
dnovillo: D'oh! :)

dnovillo:
Both issues fixed.  PTAL.

dnovillo:
Also fixed logic for preserving instr-to-block analysis in KillInst (thanks).  PTAL.

dnovillo:
Thanks.  Re-based, re-tested and pushed.

Issue #928, opened 2017-10-27
IRContext: replace uses of FindNamedOrDecoratedIds with use of the decoration manager that will live inside IRContext


dneto0:

s-perron:
I think this is done.

PR #965, opened 2017-11-14
Add the decoration manager to the IRContext.

s-perron:
To make the decoration manger available everywhere, and to reduce the
number of times it needs to be build, I add one the IRContext.

As the same time, I move code that modifies decoration instruction into
the IRContext from mempass and the decoration manager.  This will make
it easier to keep everything up to date.

This should take care of issue #928.

### DIFF
--- a/source/link/linker.cpp
+++ b/source/link/linker.cpp
@@ -227,16 +227,15 @@ spv_result_t Linker::Link(const uint32_t* const* binaries,
 
   // Phase 4: Find the import/export pairs
   LinkageTable linkings_to_do;
-  DecorationManager decoration_manager(linked_context.module());
-  res = GetImportExportPairs(consumer, linked_context,
-                             *linked_context.get_def_use_mgr(),
-                             decoration_manager, &linkings_to_do);
+  res = GetImportExportPairs(
+      consumer, linked_context, *linked_context.get_def_use_mgr(),
+      *linked_context.get_decoration_mgr(), &linkings_to_do);
   if (res != SPV_SUCCESS) return res;
 
   // Phase 5: Ensure the import and export have the same types and decorations.
   res = CheckImportExportCompatibility(consumer, linkings_to_do,
                                        *linked_context.get_def_use_mgr(),
-                                       decoration_manager);
+                                       *linked_context.get_decoration_mgr());
   if (res != SPV_SUCCESS) return res;
 
   // Phase 6: Remove duplicates
@@ -248,9 +247,9 @@ spv_result_t Linker::Link(const uint32_t* const* binaries,
 
   // Phase 7: Remove linkage specific instructions, such as import/export
   // attributes, linkage capability, etc. if applicable
-  res = RemoveLinkageSpecificInstructions(consumer, !options.GetCreateLibrary(),
-                                          linkings_to_do, &decoration_manager,
-                                          &linked_context);
+  res = RemoveLinkageSpecificInstructions(
+      consumer, !options.GetCreateLibrary(), linkings_to_do,
+      linked_context.get_decoration_mgr(), &linked_context);
   if (res != SPV_SUCCESS) return res;
 
   // Phase 8: Rematch import variables/functions to export variables/functions
--- a/source/opt/common_uniform_elim_pass.cpp
+++ b/source/opt/common_uniform_elim_pass.cpp
@@ -99,7 +99,7 @@ ir::Instruction* CommonUniformElimPass::GetPtr(ir::Instruction* ip,
 bool CommonUniformElimPass::IsVolatileStruct(uint32_t type_id) {
   assert(get_def_use_mgr()->GetDef(type_id)->opcode() == SpvOpTypeStruct);
   bool has_volatile_deco = false;
-  dec_mgr_->ForEachDecoration(type_id, SpvDecorationVolatile,
+  get_decoration_mgr()->ForEachDecoration(type_id, SpvDecorationVolatile,
                               [&has_volatile_deco](const ir::Instruction&) {
                                 has_volatile_deco = true;
                               });
@@ -195,33 +195,10 @@ bool CommonUniformElimPass::HasOnlyNamesAndDecorates(uint32_t id) const {
   return true;
 }
 
-void CommonUniformElimPass::KillNamesAndDecorates(uint32_t id) {
-  // TODO(greg-lunarg): Remove id from any OpGroupDecorate and
-  // kill if no other operands.
-  analysis::UseList* uses = get_def_use_mgr()->GetUses(id);
-  if (uses == nullptr) return;
-  std::list<ir::Instruction*> killList;
-  for (auto u : *uses) {
-    const SpvOp op = u.inst->opcode();
-    if (op != SpvOpName && !IsNonTypeDecorate(op)) continue;
-    killList.push_back(u.inst);
-  }
-  for (auto kip : killList) context()->KillInst(kip);
-}
-
-void CommonUniformElimPass::KillNamesAndDecorates(ir::Instruction* inst) {
-  // TODO(greg-lunarg): Remove inst from any OpGroupDecorate and
-  // kill if not other operands.
-  const uint32_t rId = inst->result_id();
-  if (rId == 0) return;
-  KillNamesAndDecorates(rId);
-}
-
 void CommonUniformElimPass::DeleteIfUseless(ir::Instruction* inst) {
   const uint32_t resId = inst->result_id();
   assert(resId != 0);
   if (HasOnlyNamesAndDecorates(resId)) {
-    KillNamesAndDecorates(resId);
     context()->KillInst(inst);
   }
 }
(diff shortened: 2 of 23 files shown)


source/opt/decoration_manager.cpp
```diff
@@ -144,41 +113,45 @@ void DecorationManager::AnalyzeDecorations() {
...
-        auto const target_id = inst.GetSingleWordInOperand(0u);
+    AddDecoration(&inst);
+  }
+}
+void DecorationManager::AddDecoration(ir::Instruction* inst) {
+  switch (inst->opcode()) {
+    case SpvOpDecorate:
+    case SpvOpDecorateId:
+    case SpvOpMemberDecorate:
+    case SpvOpName: {
```
dneto0: SpvOpName looks wrong here.  That would be a functional change.

source/opt/decoration_manager.h
```diff
@@ -70,6 +73,7 @@ class DecorationManager {
   // with |true| afterwards.
   void CloneDecorations(uint32_t from, uint32_t to, std::function<void(ir::Instruction&, bool)> f);
 
+  void AddDecoration(ir::Instruction* inst);
```
dneto0: Need a method comment.

source/opt/decoration_manager.h
```diff
@@ -90,6 +94,8 @@ class DecorationManager {
   IdToDecorationInstsMap group_to_decoration_insts_;
   // The enclosing module.
   ir::Module* module_;
+  void RemoveInstructionFromTarget(ir::Instruction* inst,
```
dneto0: Need a method comment. 

Also, I prefer to see all methods before all data members.  Please move this up.

source/opt/ir_context.cpp
```diff
@@ -125,12 +142,44 @@ void spvtools::ir::IRContext::ForgetUses(Instruction* inst) {
...
+
+void IRContext::KillNamesAndDecorates(uint32_t id) {
+  std::vector<ir::Instruction*> decorations =
+      get_decoration_mgr()->GetDecorationsFor(id, true);
+
+  for (Instruction* inst : decorations) {
+    KillInst(inst);
+  }
+
+  // Should the decoration manger take care of the SpvOpName Instruction as
```
dneto0: An OpName is not a decoration. Decorations are semantically significant but OpName is a debug instruction without semantic signficance.

I think this is the right place to delete the name.   It's right in the method name too.

"manger" -> "manager".

Once we agree, either delete this comment or put in the rationale for doing the name deletion here.

source/opt/ir_context.h
```diff
@@ -170,6 +173,15 @@ class IRContext {
     return (entry != instr_to_block_.end()) ? entry->second : nullptr;
   }
 
+  // Returns a pointer the decoration manager.  If the decoration manger is
+  // invalid, it is rebuilt first.
+  opt::analysis::DecorationManager* get_decoration_mgr() {
```
dneto0: Not to be resolved in this MR.  But why are we returning pointers when we could be returning references?   The result is never going to be null.

source/opt/ir_context.h
```diff
@@ -221,6 +233,12 @@ class IRContext {
   // will be updated accordingly.
   void AnalyzeUses(Instruction* inst);
 
+  // Kill all name and decorate ops using |id|
```
dneto0: I'd replace "using" with "targeting".  That would be clearer to me.

source/opt/decoration_manager.cpp
dnovillo: Perhaps I'm confused, but isn't SpvOpName another decoration?  There's also SpvOpMemberName.  Yes, this is new wrt existing code, so if these are indeed decorations, perhaps they could be added later.

source/opt/decoration_manager.cpp
```diff
@@ -300,6 +275,44 @@ void DecorationManager::CloneDecorations(uint32_t from, uint32_t to,
   }
 }
 
+void DecorationManager::RemoveDecoration(ir::Instruction* inst) {
+  switch (inst->opcode()) {
+    case SpvOpDecorate:
+    case SpvOpDecorateId:
+    case SpvOpMemberDecorate: {
```
dnovillo: If we handle SpvOpName in AddDecoration, it should also be handled here.

source/opt/ir_context.cpp
dnovillo: OK, but shouldn't this also handle SpvOpMemberName then?

source/link/linker.cpp
```diff
@@ -227,16 +227,15 @@ spv_result_t Linker::Link(const uint32_t* const* binaries,
...
+  res = GetImportExportPairs(
+      consumer, linked_context, *linked_context.get_def_use_mgr(),
+      *linked_context.get_decoration_mgr(), &linkings_to_do);
   if (res != SPV_SUCCESS) return res;
 
   // Phase 5: Ensure the import and export have the same types and decorations.
   res = CheckImportExportCompatibility(consumer, linkings_to_do,
                                        *linked_context.get_def_use_mgr(),
-                                       decoration_manager);
+                                       *linked_context.get_decoration_mgr());
```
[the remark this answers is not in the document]
s-perron: No.  `linked_context` is not a pointer.  The `*` is dereferencing the pointer returned by "get_decoration_mgr" so that it passed the decoration manager by reference.

source/opt/common_uniform_elim_pass.cpp
```diff
@@ -195,33 +195,10 @@ bool CommonUniformElimPass::HasOnlyNamesAndDecorates(uint32_t id) const {
...
-  const uint32_t rId = inst->result_id();
-  if (rId == 0) return;
-  KillNamesAndDecorates(rId);
-}
-
 void CommonUniformElimPass::DeleteIfUseless(ir::Instruction* inst) {
   const uint32_t resId = inst->result_id();
   assert(resId != 0);
   if (HasOnlyNamesAndDecorates(resId)) {
-    KillNamesAndDecorates(resId);
```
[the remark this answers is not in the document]
s-perron: It gets called automatically in KillInst and KillDef.  The ones that were left are those where you want to remove the decorations before call ReplaceAllUses.

dnovillo:
On Tue, Nov 14, 2017 at 3:16 PM, Steven Perron <notifications@github.com>
wrote:

>
>    // Phase 5: Ensure the import and export have the same types and decorations.
>    res = CheckImportExportCompatibility(consumer, linkings_to_do,
>                                         *linked_context.get_def_use_mgr(),
> -                                       decoration_manager);
> +                                       *linked_context.get_decoration_mgr());
>
> No. linked_context is not a pointer. The * is dereferencing the pointer
> returned by "get_decoration_mgr" so that it passed the decoration manager
> by reference.
>

Ugh.  Right.

source/opt/decoration_manager.cpp
s-perron: The OpName are not in the annotations section.  They are part of the debug information.  SpvOpMemberName add a name for a type.  It does not take an general result id.  I was debating whether or not I should get the DecorationManager to handle names.  I added this, then started to rethink this.  I forgot to remove this.  I'll do that.

source/opt/decoration_manager.cpp
s-perron: I'll remove the OpName from AddDecoration.

source/opt/ir_context.cpp
s-perron: Probably.  OpMemberName is a little different because it gets applied to members of a type, the specific applications where this function was used before did not care about removing types, so it was not an issue.

I can add it in.

s-perron: I think I agree with @dneto0  that the name does not have semantic significance.  This will create special cases in the decoration manager when we ask if two ids have the same decorations.  Does that include the name?  I'll keep the removal of the names this way.

source/opt/ir_context.h
s-perron: I'll keep it consistent with the other analyses for now.  Then we can change them after.

source/opt/ir_context.h
s-perron: sounds good.

source/opt/decoration_manager.h
s-perron: done

source/opt/decoration_manager.h
s-perron: done.

s-perron:
I made the suggested changes.  Let me know if more are needed.  The only thing that I think needs to be done is to add test for KillNameAndDecorates to make sure it deletes the OpMemberNames.  Existing test are covering the removal of the decorations and OpName.

source/opt/common_uniform_elim_pass.cpp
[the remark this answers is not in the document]
dnovillo: +1

s-perron:
Test are added, should be good for another look.

PR #967, opened 2017-11-14
Appveyor: skip building tags

antiagainst:


dneto0:
Hmm. Maybe the commit message should have been "Travis: Stop pushing tags on success"  :-)

Actually, is this a problem on other bots that we've set up, e.g. Shaderc and Glslang?

antiagainst:
It it related to the new configuration for auto deployment to GitHub Releases. So we have similar problem for Glslang: https://github.com/KhronosGroup/glslang/pull/1155. Fix merged there.

Travis need a tag annotating the current ToT and the tag's name should be the release name (i.e., `master-tot`);  Otherwise it cannot push the build artifacts to the proper release. So we cannot stop pushing tags on success.

PR #969, opened 2017-11-15
Unique instruction ids

alan-baker:
In preparation for #846, this request adds an unique id to all instructions. This id is issued by the IRContext. This means instructions are effectively owned by a context. For convenience, Instruction now caches a pointer to the IRContext.

* Instruction constructors now take a context where possible
  * could not remove default or copy constructor yet, but they should **NOT** be used explicitly
* Removed instances of instruction/block/function copies or moves and replaced them with Clones.
* Removed copy constructors for blocks and functions. 
* Added an internal ability for the linker to verify unique-ness of ids in the merged module (realistically the only place dealing with multiple contexts).

### DIFF
--- a/include/spirv-tools/linker.hpp
+++ b/include/spirv-tools/linker.hpp
@@ -38,14 +38,18 @@ class LinkerOptions {
   // The returned value will be true if creating a library, and false if
   // creating an executable.
   bool GetCreateLibrary() const { return createLibrary_; }
+
   // Sets whether a library or an executable should be produced.
   void SetCreateLibrary(bool create_library) {
     createLibrary_ = create_library;
   }
 
-  // Returns whether to verify the ids in the merged module
+  // Returns whether to verify the uniqueness of the unique ids in the merged
+  // context.
   bool GetVerifyIds() const { return verifyIds_; }
-  // Sets whether to verify the ids in the merged module
+
+  // Sets whether to verify the uniqueness of the unique ids in the merged
+  // context.
   void SetVerifyIds(bool verifyIds) {
     verifyIds_ = verifyIds;
   }
--- a/include/spirv-tools/linker.hpp
+++ b/include/spirv-tools/linker.hpp
@@ -26,7 +26,9 @@ namespace spvtools {
 
 class LinkerOptions {
  public:
-  LinkerOptions() : createLibrary_(false) {}
+  LinkerOptions()
+      : createLibrary_(false),
+        verifyIds_(false) {}
 
   // Returns whether a library or an executable should be produced by the
   // linking phase.
@@ -41,8 +43,16 @@ class LinkerOptions {
     createLibrary_ = create_library;
   }
 
+  // Returns whether to verify the ids in the merged module
+  bool GetVerifyIds() const { return verifyIds_; }
+  // Sets whether to verify the ids in the merged module
+  void SetVerifyIds(bool verifyIds) {
+    verifyIds_ = verifyIds;
+  }
+
  private:
   bool createLibrary_;
+  bool verifyIds_;
 };
 
 class Linker {
--- a/source/link/linker.cpp
+++ b/source/link/linker.cpp
@@ -70,15 +70,16 @@ using LinkageTable = std::vector<LinkageEntry>;
 // is returned in |max_id_bound|.
 //
 // Both |modules| and |max_id_bound| should not be null, and |modules| should
-// not be empty either.
+// not be empty either. Furthermore |modules| should not contain any null
+// pointers.
 static spv_result_t ShiftIdsInModules(
     const MessageConsumer& consumer,
     std::vector<ir::Module*>* modules, uint32_t* max_id_bound);
 
 // Generates the header for the linked module and returns it in |header|.
 //
-// |header| should not be null, |modules| should not be empty and
-// |max_id_bound| should be strictly greater than 0.
+// |header| should not be null, |modules| should not be empty and pointers
+// should be non-null. |max_id_bound| should be strictly greater than 0.
 //
 // TODO(pierremoreau): What to do when binaries use different versions of
 //                     SPIR-V? For now, use the max of all versions found in
@@ -88,9 +89,10 @@ static spv_result_t GenerateHeader(
     const std::vector<ir::Module*>& modules,
     uint32_t max_id_bound, ir::ModuleHeader* header);
 
-// Merge all the modules from |inModules| into |linked_module|.
+// Merge all the modules from |inModules| into a single module owned by
+// |linked_context|.
 //
-// |linked_module| should not be null.
+// |linked_context| should not be null.
 static spv_result_t MergeModules(
     const MessageConsumer& consumer,
     const std::vector<Module*>& inModules,
@@ -125,7 +127,7 @@ static spv_result_t CheckImportExportCompatibility(
 // functions, declarations of imported variables, import (and export if
 // necessary) linkage attribtes.
 //
-// |linked_module| and |decoration_manager| should not be null, and the
+// |linked_context| and |decoration_manager| should not be null, and the
 // 'RemoveDuplicatePass' should be run first.
 //
 // TODO(pierremoreau): Linkage attributes applied by a group decoration are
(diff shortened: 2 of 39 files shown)


source/opt/decoration_manager.cpp
```diff
@@ -261,7 +261,7 @@ void DecorationManager::CloneDecorations(
       case SpvOpMemberDecorate:
       case SpvOpDecorateId: {
         // simply clone decoration and change |target-id| to |to|
-        std::unique_ptr<ir::Instruction> new_inst(inst->Clone());
+        std::unique_ptr<ir::Instruction> new_inst(inst->Clone(module_->context()));
         new_inst->SetInOperand(0, {to});
```
s-perron: Not for this PR, but I forgot to move this function to the IRContext.

source/opt/instruction.h
```diff
@@ -83,29 +84,30 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   using iterator = std::vector<Operand>::iterator;
   using const_iterator = std::vector<Operand>::const_iterator;
 
-  // Creates a default OpNop instruction.
+  // This exists solely for containers that can't do without. Should be removed.
```
s-perron: Should still keep the comment about what it does, even if we want to eventually remove it.

source/opt/instruction.h
```diff
@@ -241,15 +246,22 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   // Returns true if the instruction annotates an id with a decoration.
   inline bool IsDecoration();
 
+  inline bool operator==(const Instruction&) const;
+  inline bool operator!=(const Instruction&) const;
+  inline bool operator<(const Instruction&) const;
+
  private:
+
```
s-perron: No blank line here.

include/spirv-tools/linker.hpp
```diff
@@ -41,8 +43,16 @@ class LinkerOptions {
     createLibrary_ = create_library;
   }
 
+  // Returns whether to verify the ids in the merged module
+  bool GetVerifyIds() const { return verifyIds_; }
+  // Sets whether to verify the ids in the merged module
```
dnovillo: Blank line above.

source/link/linker.cpp
```diff
@@ -186,7 +192,8 @@ spv_result_t Linker::Link(const uint32_t* const* binaries,
                                       SPV_ERROR_INVALID_BINARY)
            << "No modules were given.";
 
-  std::vector<std::unique_ptr<Module>> modules;
+  std::vector<std::unique_ptr<IRContext>> contexts;
+  std::vector<Module*> modules;
```
dnovillo: Curious why did the vector of Modules change from a vector of unique_ptr to a vector of raw ptr?

source/opt/build_module.h
```diff
@@ -18,23 +18,25 @@
...
 namespace spvtools {
 
-// Builds and returns an ir::Module from the given SPIR-V |binary|. |size|
-// specifies number of words in |binary|. The |binary| will be decoded
-// according to the given target |env|. Returns nullptr if erors occur and
-// sends the errors to |consumer|.
-std::unique_ptr<ir::Module> BuildModule(spv_target_env env,
+// Builds an ir::Module returns the owning ir::IRContext from the given SPIR-V
+// |binary|. |size| specifies number of words in |binary|. The |binary| will be
+// decoded according to the given target |env|. Returns nullptr if erors occur
```
dnovillo: s/erors/errors/

source/opt/build_module.h
```diff
@@ -18,23 +18,25 @@
...
                                         MessageConsumer consumer,
                                         const uint32_t* binary, size_t size);
 
-// Builds and returns an ir::Module from the given SPIR-V assembly |text|.
-// The |text| will be encoded according to the given target |env|. Returns
-// nullptr if erors occur and sends the errors to |consumer|.
-std::unique_ptr<ir::Module> BuildModule(
+// Builds an ir::Module and returns the owning ir::IRContext from the given
+// SPIR-V assembly |text|.  The |text| will be encoded according to the given
+// target |env|. Returns nullptr if erors occur and sends the errors to
```
dnovillo: s/erors/errors/  (the misspelling was there before, but since you're here... ;)

source/opt/ir_context.h
```diff
@@ -239,6 +263,14 @@ class IRContext {
   // Kill all name and decorate ops targeting the result id of |inst|.
   void KillNamesAndDecorates(ir::Instruction* inst);
 
+  // Return the next unique id for use by an instruction
+  inline uint32_t TakeNextUniqueId() {
+    assert(unique_id_ != std::numeric_limits<uint32_t>::max());
+
+    // Skip zero
```
dnovillo: Terminate all comments with '.' (yeah, we're not consistent :)

source/opt/module.h
```diff
@@ -223,11 +225,17 @@ class Module {
   // Returns 0 if not found.
   uint32_t GetExtInstImportId(const char* extstr);
 
+  // Sets the associated context for this module
+  void SetContext(IRContext* c) { context_ = c; }
+  // Gets the associated context for this module
```
dnovillo: Blank line above, please.

source/link/linker.cpp
alan-baker: Convenience. The modules are owned by the vector<unique_ptr<IRContext>>, but most of the code referred to the module vector. So to avoid further rewrites I also tracked raw pointers to the modules to avoid larger changes in the linker.

source/opt/ir_context.h
```diff
@@ -53,11 +54,34 @@ class IRContext {
   friend inline Analysis operator<<(Analysis a, int shift);
   friend inline Analysis& operator<<=(Analysis& a, int shift);
 
+  // Trivial constructor for testing purposes
```
dneto0: What goes wrong if someone uses this by accident?  Seems like we should more strenuously steer people away from this.

If you truly need something for testing purposes, then I recommend making a static method to make it, eg..

static IRContext* MakeIRContextForTest() ... and then hide the details


...
Then I read the tests and made a suggestion there to use a helper function to make the dummy IRContexts.

source/opt/ir_context.h
```diff
@@ -239,6 +263,14 @@ class IRContext {
   // Kill all name and decorate ops targeting the result id of |inst|.
   void KillNamesAndDecorates(ir::Instruction* inst);
 
+  // Return the next unique id for use by an instruction
```
dneto0: While we're nitpicky, prefer "Returns" over "Return".
See style guide:  https://google.github.io/styleguide/cppguide.html#Function_Comments 
In particular:  "These comments should be descriptive ("Opens the file") rather than imperative ("Open the file"); the comment describes the function, it does not tell the function what to do"

Get used to writing comments in that style.  Yes, the codebase is inconsistent on this, but I'm holding Googlers to the Google style guide for this rule.

source/opt/ir_context.h
```diff
@@ -264,6 +296,7 @@ class IRContext {
     valid_analyses_ = valid_analyses_ | kAnalysisDecorations;
   }
 
+  uint32_t unique_id_;
```
dneto0: Needs a comment. to describe this, especially how it starts at 0 but is not the next unique ID to be used.  It's one *less* than the next unique id to be used.

source/opt/ir_loader.cpp
```diff
@@ -20,22 +20,22 @@
 namespace spvtools {
 namespace ir {
 
-IrLoader::IrLoader(const MessageConsumer& consumer, Module* module)
+IrLoader::IrLoader(const MessageConsumer& consumer, Module* m)
```
dneto0: Are you doing this to avoid shadowing?  Probably, and that's ok. :-)

test/opt/instruction_test.cpp
```diff
@@ -168,7 +175,8 @@ TEST(InstructionTest, OperandConstIterators) {
 }
 
 TEST(InstructionTest, OperandIterators) {
-  Instruction inst(kSampleParsedInstruction);
+  IRContext context;
```
dneto0: Instead of having a default constructor that nobody else in the world uses, I recommend making a helper function in this module to make a dummy, but with few words.

e.g. 
static IRContext DummyContext() { return IRContext(Module(), ...); }

Then in the functions:
IRContext& context = DummyContext();

And then let lifetime extension carry it through.

source/opt/ir_loader.cpp
alan-baker: Yes this is because of shadowing.

alan-baker:
I removed the default IRContext and while I was at it I reworked some large portions of the DefUseManager tests in anticipation of the next round of changes. Previously I was creating dummy contexts, but this would have needed to be rewritten when the storage of uses changes anyways. Removing the temporary contexts required changing the way several of the tests functioned. I removed some that no longer made sense and replaced others that were awkward to instantiate via the parameterizations. Unfortunately, this has increased the size of the diff considerably.

dneto0:
Rebased, squashed, and pushed into master as a771713e4

PR #970, opened 2017-11-15
Git should ignore external repos

dneto0:
Ignore googletest, spirv-headers, effcee, and re2


alan-baker:
Rebased and push as dbf9b3762cd485d05387b495495e1a359da8ee37

Issue #960, opened 2017-11-10
Implement a more general redundancy elimination pass.


s-perron:
As far as I know, the only form of redundancy elimination in spirv-opt is common-uniform-elimination.  However that is very specific to loads of uniforms.  However, there are many other reasons for redundancy elimination.  First there is generic removal of duplicate code that comes just as a natural part of translating high level languages.  For example, consider

`
#version 430

layout(set = 0, binding = 0) uniform sampler2DShadow g_Texture;
layout(location = 0) out float SV_TARGET0;
vec4 u_xlat0;
vec4 u_xlat1;
vec4 u_xlat2;
vec2 u_xlat3;
vec4 u_xlat4;
vec4 u_xlat5;
float u_xlat6;
vec2 u_xlat12;
float u_xlat18;
void main()
{
    float refZ = 0.25;
    u_xlat0 = textureGatherOffset(g_Texture, vec2(0.0, 0.0), refZ, ivec2(-2, -2));
    u_xlat1 = textureGatherOffset(g_Texture, vec2(0.0, 0.0), refZ, ivec2(0, -2));
    u_xlat2 = textureGatherOffset(g_Texture, vec2(0.0, 0.0), refZ, ivec2(2, -2));
    return;
}
`

Compile with the glslc (from shaderc), `glslc -S t.frag`.  Looking at the output file, you will see that the gather instructions look very different even if the only real difference is in the last operand:
`
         %26 = OpImageDrefGather %v4float %17 %20 %21 ConstOffset %25                                                                                                                                                                                                                                                       
         %32 = OpImageDrefGather %v4float %28 %20 %29 ConstOffset %31
         %38 = OpImageDrefGather %v4float %34 %20 %35 ConstOffset %37
`

For example %17, %28, and %34 are all defined the exact same way.  Same for %21, %29, and %35, but that is not obvious.

To help with these things we need a general redundancy elimination that will get rid of the extra loads, and make the instructions that use them look the same in places where they are actually the same.

s-perron:
My suggestion for this is to do a hashing based value numbering pass to deal with redundancy for the symbols with ssa name.  We have to deal with memory locations, which are not in ssa form, differently.  Function scope variables can given ssa names by "local-multi-store-elimination" pass.  Something like "local-uniform-elimination" could be used to fix up any type of read-only memory, but we would want a solution that will not generate a load of an entire ubo (see #946).

[...]

PR #971, opened 2017-11-16
Create a local redundancy elimination pass

s-perron:
Creates a pass that removes redundant instructions within the same basic
block.  This will be implemented using a hash based value numbering
algorithm.  This is the first part of the fix for #960.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -76,10 +76,12 @@ SPVTOOLS_OPT_SRC_FILES := \
 		source/opt/ir_context.cpp \
 		source/opt/ir_loader.cpp \
 		source/opt/local_access_chain_convert_pass.cpp \
+		source/opt/local_redundancy_elimination.cpp \
 		source/opt/local_single_block_elim_pass.cpp \
 		source/opt/local_single_store_elim_pass.cpp \
 		source/opt/local_ssa_elim_pass.cpp \
 		source/opt/mem_pass.cpp \
+		source/opt/merge_return_pass.cpp \
 		source/opt/module.cpp \
 		source/opt/optimizer.cpp \
 		source/opt/pass.cpp \
@@ -91,7 +93,7 @@ SPVTOOLS_OPT_SRC_FILES := \
 		source/opt/type_manager.cpp \
 		source/opt/types.cpp \
 		source/opt/unify_const_pass.cpp \
-		source/opt/merge_return_pass.cpp
+		source/opt/value_number_table.cpp
 
 # Locations of grammar files.
 SPV_CORE10_GRAMMAR=$(SPVHEADERS_LOCAL_PATH)/include/spirv/1.0/spirv.core.grammar.json
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -318,7 +318,7 @@ int32_t spvOpcodeGeneratesType(SpvOp op) {
 }
 
 bool spvOpcodeIsDecoration(const SpvOp opcode) {
-  switch(opcode) {
+  switch (opcode) {
     case SpvOpDecorate:
     case SpvOpDecorateId:
     case SpvOpMemberDecorate:
@@ -329,4 +329,59 @@ bool spvOpcodeIsDecoration(const SpvOp opcode) {
       break;
   }
   return false;
+}
+
+bool spvOpcodeIsLoad(const SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpLoad:
+    case SpvOpImageSampleExplicitLod:
+    case SpvOpImageSampleImplicitLod:
+    case SpvOpImageSampleDrefImplicitLod:
+    case SpvOpImageSampleDrefExplicitLod:
+    case SpvOpImageSampleProjImplicitLod:
+    case SpvOpImageSampleProjExplicitLod:
+    case SpvOpImageSampleProjDrefImplicitLod:
+    case SpvOpImageSampleProjDrefExplicitLod:
+    case SpvOpImageFetch:
+    case SpvOpImageGather:
+    case SpvOpImageDrefGather:
+    case SpvOpImageRead:
+    case SpvOpImageSparseSampleImplicitLod:
+    case SpvOpImageSparseSampleExplicitLod:
+    case SpvOpImageSparseSampleDrefExplicitLod:
+    case SpvOpImageSparseSampleDrefImplicitLod:
+    case SpvOpImageSparseFetch:
+    case SpvOpImageSparseGather:
+    case SpvOpImageSparseDrefGather:
+    case SpvOpImageSparseRead:
+      return true;
+    default:
+      return false;
+  }
+}
+
+bool spvOpcodeIsAtomicOp(const SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpAtomicLoad:
+    case SpvOpAtomicStore:
+    case SpvOpAtomicExchange:
+    case SpvOpAtomicCompareExchange:
+    case SpvOpAtomicCompareExchangeWeak:
+    case SpvOpAtomicIIncrement:
+    case SpvOpAtomicIDecrement:
+    case SpvOpAtomicIAdd:
+    case SpvOpAtomicISub:
+    case SpvOpAtomicSMin:
+    case SpvOpAtomicUMin:
+    case SpvOpAtomicSMax:
+    case SpvOpAtomicUMax:
+    case SpvOpAtomicAnd:
+    case SpvOpAtomicOr:
+    case SpvOpAtomicXor:
+    case SpvOpAtomicFlagTestAndSet:
+    case SpvOpAtomicFlagClear:
+      return true;
+    default:
+      return false;
+  }
 }
--- a/source/opcode.h
+++ b/source/opcode.h
@@ -87,7 +87,14 @@ bool spvOpcodeReturnsLogicalVariablePointer(const SpvOp opcode);
 // non-zero otherwise.
 int32_t spvOpcodeGeneratesType(SpvOp opcode);
 
-// Returns true if the opcode add a decoration to an id.
+// Returns true if the opcode adds a decoration to an id.
 bool spvOpcodeIsDecoration(const SpvOp opcode);
+
+// Returns true if the opcode is a load from memory into a result id.  This
+// function only considers core instructions.
+bool spvOpcodeIsLoad(const SpvOp opcode);
+
+// Returns true if the opcode is an atomic operation.
+bool spvOpcodeIsAtomicOp(const SpvOp opcode);
 
 #endif  // LIBSPIRV_OPCODE_H_
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -39,6 +39,7 @@ add_library(SPIRV-Tools-opt
   ir_loader.h
   ir_context.h
   local_access_chain_convert_pass.h
+  local_redundancy_elimination.h
   local_single_block_elim_pass.h
   local_single_store_elim_pass.h
   local_ssa_elim_pass.h
@@ -59,6 +60,7 @@ add_library(SPIRV-Tools-opt
   types.h
   type_manager.h
   unify_const_pass.h
+  value_number_table.h
 
   aggressive_dead_code_elim_pass.cpp
   basic_block.cpp
@@ -83,9 +85,11 @@ add_library(SPIRV-Tools-opt
   inline_opaque_pass.cpp
   insert_extract_elim.cpp
   instruction.cpp
+  instruction_list.cpp
   ir_loader.cpp
   ir_context.cpp
   local_access_chain_convert_pass.cpp
+  local_redundancy_elimination.cpp
   local_single_block_elim_pass.cpp
   local_single_store_elim_pass.cpp
   local_ssa_elim_pass.cpp
@@ -103,7 +107,7 @@ add_library(SPIRV-Tools-opt
   types.cpp
   type_manager.cpp
   unify_const_pass.cpp
-  instruction_list.cpp
+  value_number_table.cpp
 )
 
 spvtools_default_compile_options(SPIRV-Tools-opt)
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -17,10 +17,19 @@
 
 #include <initializer_list>
 
+#include "ir_context.h"
 #include "reflect.h"
 
 namespace spvtools {
 namespace ir {
+
+namespace {
+// Indices used to get particular operands out of instructions using InOperand.
+const uint32_t kTypeImageDimIndex = 1;
+const uint32_t kLoadBaseIndex = 0;
+const uint32_t kVariableStorageClassIndex = 0;
+const uint32_t kTypeImageSampledIndex = 5;
+}  // namespace
 
 Instruction::Instruction(IRContext* c)
     : utils::IntrusiveNodeBase<Instruction>(),
@@ -57,7 +66,8 @@ Instruction::Instruction(IRContext* c, const spv_parsed_instruction_t& inst,
   }
 }
 
-Instruction::Instruction(IRContext* c, SpvOp op, uint32_t ty_id, uint32_t res_id,
+Instruction::Instruction(IRContext* c, SpvOp op, uint32_t ty_id,
+                         uint32_t res_id,
                          const std::vector<Operand>& in_operands)
     : utils::IntrusiveNodeBase<Instruction>(),
       context_(c),
@@ -96,7 +106,7 @@ Instruction& Instruction::operator=(Instruction&& that) {
   return *this;
 }
 
-Instruction* Instruction::Clone(IRContext *c) const {
+Instruction* Instruction::Clone(IRContext* c) const {
   Instruction* clone = new Instruction(c);
   clone->opcode_ = opcode_;
   clone->type_id_ = type_id_;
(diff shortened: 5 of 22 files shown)


Android.mk
```diff
@@ -91,6 +91,8 @@ SPVTOOLS_OPT_SRC_FILES := \
 		source/opt/types.cpp \
 		source/opt/unify_const_pass.cpp \
 		source/opt/merge_return_pass.cpp
+		sourc/opt/value_number_table.cpp \
+		sourc/opt/value_numbering_pass.cpp
```
s-perron: I renamed this file to local_redundancy_elimination.cpp, but the refactoring tool I used missed this.  This is causing the android build failures.  I'll fix that up.

source/opt/value_number_table.h
```diff
@@ -0,0 +1,73 @@
...
+// instruction compute the same value.
+class ComputeSameValue {
+ public:
+  bool operator()(const ir::Instruction& lhs, const ir::Instruction& rhs) const;
+};
+
+// The hash function used in the value number table.
+class ValueTableHash {
+ public:
+  std::size_t operator()(const spvtools::ir::Instruction& inst) const noexcept;
```
s-perron: The noexcept is causing the VS2013 build to fail.  However, I believe I need it for other builds.  I'll have to look into that.  I might need some macro magic to fix this.

source/opt/CMakeLists.txt
```diff
@@ -59,6 +59,8 @@ add_library(SPIRV-Tools-opt
   types.h
   type_manager.h
   unify_const_pass.h
+  value_number_table.h
+        local_redundancy_elimination.h
```
alan-baker: odd formatting

source/opt/CMakeLists.txt
```diff
@@ -103,7 +106,8 @@ add_library(SPIRV-Tools-opt
   types.cpp
   type_manager.cpp
   unify_const_pass.cpp
-  instruction_list.cpp
+  value_number_table.cpp
+        local_redundancy_elimination.cpp
```
alan-baker: here too

source/opt/instruction.h
```diff
@@ -241,6 +242,18 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   // Returns true if the instruction annotates an id with a decoration.
   inline bool IsDecoration();
 
+  // Returns true if the instruction is known to be a load from read-only
+  // memory.
+  bool IsReadOnlyLoad(IRContext* ctx) const;
+
+  // Returns the instruction that gives the base address of an address
+  // calculation.
```
alan-baker: You should document the IsLoad() pre-condition

source/opt/value_number_table.cpp
```diff
@@ -0,0 +1,97 @@
...
+    const spvtools::ir::Instruction& inst) const {
+  std::size_t hash = 0;
+  std::hash<uint32_t> hash_function;
+
+  // We hash the in operands only because we are want instructions that are
+  // the same except for the result to hash to the same value.
+  for (uint32_t i = 0; i < inst.NumInOperands(); ++i) {
+    const auto& opnd = inst.GetInOperand(i);
+    for (uint32_t word : opnd.words) {
+      hash ^= hash_function(word);
```
alan-baker: Should we use a better hash combination than just xor?

s-perron: I'm open to suggestions.  I know essentially nothing about hash functions except that it is hard to get right.

atgoo: Indeed, this would mean that "a-b" is the same as "b-a"

`std::hash<std::u32string>`?

I use a home-made function, but only because I need it to be consistent over multiple platforms.
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/id_descriptor.cpp#L28

source/opt/value_number_table.cpp
```diff
@@ -0,0 +1,97 @@
...
+}
+
+std::size_t ValueTableHash::operator()(
+    const spvtools::ir::Instruction& inst) const {
+  std::size_t hash = 0;
+  std::hash<uint32_t> hash_function;
+
+  // We hash the in operands only because we are want instructions that are
+  // the same except for the result to hash to the same value.
+  for (uint32_t i = 0; i < inst.NumInOperands(); ++i) {
```
atgoo: Why not hash the opcode too?

source/opt/local_redundancy_elimination.h
```diff
@@ -0,0 +1,45 @@
...
+
+// See optimizer.hpp for documentation.
+class LocalRedundancyEliminationPass : public Pass {
+ public:
+  const char* name() const override { return "local-redundancy-elimination"; }
+  Status Process(ir::IRContext*) override;
+
+ private:
+  // Deletes instructions in |block| whose value is in |vnTable| or is computed
+  // earlier in |block|.  The values computed in |block| are added to |vnTable|.
```
dnovillo: But if it's already in |vnTable|, it's no longer a local CSE.  Or are you thinking of using this same mechanism for global CSE?

source/opt/local_redundancy_elimination.h
```diff
@@ -0,0 +1,45 @@
...
+  // |value_to_ids| is a map from value number to the result ids known to
+  // contain those values.  One value needs to map to multiple ids because the
+  // ids may contain the same value, but have different decorations.
```
dnovillo: Perhaps do an actual map?  IDs can be sparse (perhaps it doesn't really matter, I'm not sure).

source/opt/value_number_table.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <algorithm>
+
+namespace spvtools {
+namespace opt {
+
+uint32_t ValueNumberTable::GetValueNumber(spvtools::ir::Instruction* inst) {
+  // TODO: Need to implement the substitution of operands by their value number
+  // before hashing.
+  // TODO: Implement a normal form for opcodes that commute like integer
+  // addition.  This will let us know that a+b is the same value as b+a.
```
atgoo: IEEE 754 float addition and multiplication is supposed to be commutative. Still it feels like playing with fire.

source/opt/value_number_table.cpp
```diff
@@ -0,0 +1,97 @@
...
+  }
+  return true;
+}
+
+std::size_t ValueTableHash::operator()(
+    const spvtools::ir::Instruction& inst) const {
+  std::size_t hash = 0;
+  std::hash<uint32_t> hash_function;
+
+  // We hash the in operands only because we are want instructions that are
```
dnovillo: s/we are want/we want/.

I'm not following, if you don't hash the opcode, won't you get the same hash value for '%5 * %6' and '%5 + %6'?

Android.mk
```diff
@@ -90,7 +91,8 @@ SPVTOOLS_OPT_SRC_FILES := \
 		source/opt/type_manager.cpp \
 		source/opt/types.cpp \
 		source/opt/unify_const_pass.cpp \
-		source/opt/merge_return_pass.cpp
+		source/opt/merge_return_pass.cpp \
```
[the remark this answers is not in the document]
s-perron: done

source/opt/CMakeLists.txt
s-perron: Fixed, plus I moved it into the correct place.  That is what happens you trust the tools too much.

source/opt/CMakeLists.txt
```diff
@@ -59,6 +59,8 @@ add_library(SPIRV-Tools-opt
   types.h
   type_manager.h
   unify_const_pass.h
+  value_number_table.h
+        local_redundancy_elimination.h
```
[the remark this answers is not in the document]
s-perron: Done

source/opt/CMakeLists.txt
s-perron: Done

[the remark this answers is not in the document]
s-perron: Done

source/opt/instruction.cpp
```diff
@@ -109,5 +110,82 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
+        break;
+      default:
+        done = true;
+        break;
+    }
+  }
+  return base_inst;
+}
+
+bool Instruction::IsLoad() const {
```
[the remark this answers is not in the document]
s-perron: Done.

source/opt/instruction.h
s-perron: done.

source/opt/instruction.h
```diff
@@ -241,6 +242,18 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
...
+  Instruction* GetBaseAddress(IRContext* ctx) const;
+
+  // Returns true if the instruction is a load from memory.  This does not
+  // include memory-to-memory instructions.
+  bool IsLoad() const;
```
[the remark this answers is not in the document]
s-perron: I prefer to keep the name to keep it close the the OpLoad instructions.  This is for OpLoad instructions or others that are similar.  I don't think if an instruction like OpCopyMemory as loads because nothing is ever visible outside of memory.  However, I have been a part of this discussion before on whether or not a memcpy like instruction is a load, so I just wanted to be explicit.

Looking at the spec, the copy operations are described as loads, for whatever that is worth.

If we want to take the position that OpCopyMemory is a load, I am fine with changing the name.

source/opt/local_redundancy_elimination.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+Pass::Status LocalRedundancyEliminationPass::Process(ir::IRContext* c) {
+  InitializeProcessing(c);
+  bool modified = false;
+
+  for (auto& func : *get_module()) {
+    for (auto& bb : func) {
+      // Resetting the value number table for every basic block because we just
+      // want the opportunities within a basic block.  This will help keep
+      // register pressure down.
```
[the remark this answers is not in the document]
s-perron: Yes, I am planning on have a global CSE pass as well.  People can choose which one they want.  As you observe below, I have written the code in such a way that most of the code can be reused for the global CSE.  The only difference will be the traversal of the basic blocks.

source/opt/local_redundancy_elimination.cpp
```diff
@@ -0,0 +1,80 @@
...
+  auto func = [this, vnTable, &modified, value_to_ids](ir::Instruction* inst) {
+    if (inst->result_id() == 0) {
+      return;
+    }
+
+    uint32_t value = vnTable->GetValueNumber(inst);
+    if (value >= value_to_ids->size()) {
+      value_to_ids->resize(value + 1);
+    }
+    std::vector<uint32_t>& representative_set = (*value_to_ids)[value];
```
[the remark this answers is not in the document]
s-perron: IDs may be sparse, but the value numbers are dense.

source/opt/local_redundancy_elimination.cpp
```diff
@@ -0,0 +1,80 @@
...
+    bool found_representative = false;
+    for (uint32_t representative_id : representative_set) {
```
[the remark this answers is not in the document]
s-perron: I've added some comments, and renamed variables.  Hopefully it is clearer now.  Let me know.

source/opt/local_redundancy_elimination.h
```diff
@@ -0,0 +1,45 @@
...
+#define LIBSPIRV_OPT_LOCAL_REDUNDANCY_ELIMINATION_H_
+
+#include "ir_context.h"
+#include "pass.h"
+#include "value_number_table.h"
+
+namespace spvtools {
+namespace opt {
+
+// See optimizer.hpp for documentation.
```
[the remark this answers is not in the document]
s-perron: Done.

source/opt/local_redundancy_elimination.h
s-perron: I wanted this to be general enough to be used for a global version of the same algorithm.  When I start on the global version, we can have a discussion on the exact algorithm, but this function should be reused.

source/opt/local_redundancy_elimination.h
s-perron: The values are the value numbers in the value table.  Those numbers start at 1 and are incremented every time we need a new value number, they will generally be dense.  Right now the numbers are being reset for every basic block.  For the global version, it expect the numbers to be reset for every function, so they will still be fairly dense.

source/opt/value_number_table.cpp
s-perron: Yes, for floating point arithmetic we will have to be careful.  I would expect that we can only do this if we have "relaxed precision" or the like.  We'll have to cross that bridge when we come to it.

source/opt/value_number_table.cpp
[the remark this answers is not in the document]
s-perron: fixed.

s-perron: Yes, I should have included the opcode.

source/opt/value_number_table.cpp
s-perron: fixed.

source/opt/value_number_table.cpp
s-perron: Yes, "a-b" and "b-a" would hash to the same value.  Too many collisions.  Thanks @atgoo for the suggestion.  I'll use `std::hash<std::u32string>`.  The next thing to watch out for is if this effects the compile time negatively.

source/opt/value_number_table.h
```diff
@@ -0,0 +1,73 @@
...
+#include <cstdint>
+#include <unordered_map>
+#include "instruction.h"
+#include "ir_context.h"
+
+namespace spvtools {
+namespace opt {
+
+// Uses as the comparison in the value number table to determine if two
+// instruction compute the same value.
```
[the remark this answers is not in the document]
s-perron: Fix up a bit.

source/opt/value_number_table.h
```diff
@@ -0,0 +1,73 @@
...
+  bool operator()(const ir::Instruction& lhs, const ir::Instruction& rhs) const;
+};
+
+// The hash function used in the value number table.
+class ValueTableHash {
+ public:
+  std::size_t operator()(const spvtools::ir::Instruction& inst) const;
+};
+
+// This class implement the value number analysis.  It is using a hash-based
```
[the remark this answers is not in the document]
s-perron: fixed.

s-perron:
Need to fix the `IsReadOnlyLoad` function.  @antiagainst pointed out some errors.

source/opt/instruction.cpp
```diff
@@ -109,5 +110,166 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
+    return false;
+  }
+
+  if (GetSingleWordInOperand(0) == SpvDimBuffer) {
+    return false;
+  }
+
+  // Check if the image is sampled.  If we do not know for sure that it is,
+  // then assume it is a storage image.
+  return GetSingleWordInOperand(6) != 1;
```
dnovillo: Use a const int instead of hardcoding 6?

source/opt/instruction.cpp
```diff
@@ -109,5 +110,166 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
+    return false;
+  }
+
+  if (GetSingleWordInOperand(0) != SpvDimBuffer) {
+    return false;
+  }
+
+  // Check if the image is sampled.  If we do not know for sure that it is,
+  // then assume it is a storage texel buffer.
+  return GetSingleWordInOperand(6) != 1;
```
dnovillo: Likewise.

source/opt/instruction.cpp
```diff
@@ -109,5 +110,166 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
+}
+
+bool Instruction::IsVulkanStorageBuffer(IRContext* ctx) const {
+  // Is there a difference between a "Storage buffer" and a "dynamic storage
+  // buffer" in SPIR-V and do we age about the difference?
```
dnovillo: "do we age"?

source/opt/instruction.h
```diff
@@ -15,11 +15,11 @@
 #ifndef LIBSPIRV_OPT_INSTRUCTION_H_
 #define LIBSPIRV_OPT_INSTRUCTION_H_
 
+#include <opcode.h>
```
dnovillo: Better use "opcode.h"

source/opt/instruction.h
```diff
@@ -106,6 +107,7 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   // Creates an instruction with the given opcode |op|, type id: |ty_id|,
   // result id: |res_id| and input operands: |in_operands|.
   Instruction(SpvOp op, uint32_t ty_id, uint32_t res_id,
+
```
dnovillo: No blank line.

source/opt/instruction.h
```diff
@@ -239,7 +241,45 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   void ReplaceOperands(const std::vector<Operand>& new_operands);
 
   // Returns true if the instruction annotates an id with a decoration.
-  inline bool IsDecoration();
+  inline bool IsDecoration() const;
+
+  // Returns true if the instruction is known to be a load from read-only
+  // memory.
```
dnovillo: Document why |ctx| is required (in the others too).

s-perron:
I need to add test for the new functions I added to the instruction class, IsReadOnly in particular.  Also, I was told with combination of flags are read only for kernel, so I'll update the code to handle that as well.

s-perron:
I also need to handle atomic operations.  I'll add a comment about volatile.  They do not require a special case because they are not read only load.

source/opt/instruction.cpp
s-perron: Done, for this and other constant indices for operands.

source/opt/instruction.cpp
s-perron: Done.

source/opt/instruction.cpp
s-perron: Fixed, but I'm still looking for an answer to the question.

source/opt/instruction.h
s-perron: done

source/opt/instruction.h
s-perron: fixed.

source/opt/instruction.h
s-perron: I'll skip that.  If #969 goes through first, I'll be removing the parameter, and just using the context in the instruction if it exists.

s-perron:
Should be ready for another review. @dnovillo

dnovillo:
On Mon, Nov 20, 2017 at 3:43 PM, Steven Perron <notifications@github.com>
wrote:

> Should be ready for another review. @dnovillo
> <https://github.com/dnovillo>
>

Thanks.  Will take a look.

source/opt/instruction.cpp
```diff
@@ -16,11 +16,20 @@
...
 #include <initializer_list>
 
+#include "ir_context.h"
 #include "reflect.h"
 
 namespace spvtools {
 namespace ir {
 
+namespace {
+// Indicies used to get particular operands out of instructions using InOperand.
```
dnovillo: s/Indicies/Indices/

source/opt/instruction.h
dnovillo: Sounds good.

s-perron:
Rebased and squashed.

source/opt/value_number_table.cpp
dneto0: Forgive the drive-by comment:

For graphics APIs, the default is "fast relaxed" math: assume you can reassociate, commute, and all sorts of numerically unstable things. That can be turned off if there is a "NoContraction" decoration.

For compute APIs (OpenCL), the default is the same as C: can't reassociate or contract, etc.  The rules can be relaxed by applying "FP Fast Math" flags on basic operations, including the Fast and AllowRecip flags.

source/opt/value_number_table.cpp
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  // Atomic operation often represent shared memory, so something else could
+  // have changed the inputs.  It atomic instruction need a unique value number.
+  if (inst->IsAtomicOp()) {
+    uint32_t value_number = TakeNextValueNumber();
+    id_to_value_[inst->result_id()] = value_number;
+    return value_number;
+  }
+
```
dneto0: Should grant new value numbers for all group instructions since they represent communication with the outside world.

Similar for pipe reads : OpPipeRead, OpReservedReadPipe.

... which makes me think that you should be using a whitelist of what instructions are actually combinators.  
See IsCombinator*  aggressive_dead_code_elim_pass.

The code in the rest of this function only correctly says yes when the instructions are combinators.

source/opcode.cpp
```diff
@@ -330,3 +330,57 @@ bool spvOpcodeIsDecoration(const SpvOp opcode) {
...
+      return false;
+  }
+}
+
+bool spvOpcodeIsAtomicOp(const SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpAtomicLoad:
+    case SpvOpAtomicStore:
+    case SpvOpAtomicExchange:
+    case SpvOpAtomicCompareExchange:
```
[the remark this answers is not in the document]
s-perron: Done

source/opcode.h
```diff
@@ -87,7 +87,13 @@ bool spvOpcodeReturnsLogicalVariablePointer(const SpvOp opcode);
 // non-zero otherwise.
 int32_t spvOpcodeGeneratesType(SpvOp opcode);
 
-// Returns true if the opcode add a decoration to an id.
+// Returns true if the opcode adds a decoration to an id.
 bool spvOpcodeIsDecoration(const SpvOp opcode);
 
+// Returns true if the opcode is a load from memory into a result id.
+bool spvOpcodeIsLoad(const SpvOp opcode);
```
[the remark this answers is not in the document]
s-perron: Added a comment that we are only considering core instructions.

source/opt/instruction.cpp
```diff
@@ -133,5 +142,186 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
 }
+
+bool Instruction::IsReadOnlyLoad() const {
+if (IsLoad()) {
+  ir::Instruction* address_def = GetBaseAddress();
+  if (address_def->opcode() == SpvOpLoad ||
+      address_def->opcode() == SpvOpFunctionParameter) {
+    return false;
+  }
+  assert(address_def->opcode() == SpvOpVariable &&
```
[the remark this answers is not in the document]
s-perron: done.

source/opt/instruction.cpp
```diff
@@ -133,5 +142,186 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
+ir::Instruction* base_inst = context()->get_def_use_mgr()->GetDef(base);
+bool done = false;
+while (!done) {
+  switch (base_inst->opcode()) {
+    case SpvOpAccessChain:
+    case SpvOpInBoundsAccessChain:
+    case SpvOpPtrAccessChain:
+    case SpvOpInBoundsPtrAccessChain:
+    case SpvOpImageTexelPointer:
+      base = base_inst->GetSingleWordInOperand(kLoadBaseIndex);
```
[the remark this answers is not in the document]
s-perron: Fixed the first two comments.  I'll leave the third.  In that case we will return the bitcast as the base pointer, and value numbering will assume it is not read only.  If there is a specific case others want to get it can be extended to get look through those instructions as well.

source/opt/instruction.cpp
```diff
@@ -133,5 +142,186 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
+      context()->get_def_use_mgr()->GetDef(GetSingleWordInOperand(1));
+  return base_type->IsVulkanStorageImage();
+}
+
+if (opcode() != SpvOpTypeImage) {
+  return false;
+}
+
+if (GetSingleWordInOperand(kTypeImageDimIndex) == SpvDimBuffer) {
+    return false;
```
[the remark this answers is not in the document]
s-perron: Discussed with @dneto0.  Even if it is not spelled out explicitly, this seems to be the only way to tell the difference between a StorageTexelBuffer and a StorageImage.  We opened an issue to look into this. I've added  a TODO here to follow up with that.

source/opt/instruction.cpp
```diff
@@ -133,5 +142,186 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
+        context()->get_def_use_mgr()->GetDef(GetSingleWordInOperand(1));
+    return base_type->IsVulkanSampledImage();
+  }
+
+  if (opcode() != SpvOpTypeImage) {
+    return false;
+  }
+
+  if (GetSingleWordInOperand(kTypeImageDimIndex) == SpvDimBuffer) {
+    return false;
```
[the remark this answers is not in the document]
s-perron: See above.

source/opt/instruction.cpp
```diff
@@ -133,5 +142,186 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
+}
+
+bool Instruction::IsVulkanStorageBuffer() const {
+  // Is there a difference between a "Storage buffer" and a "dynamic storage
+  // buffer" in SPIR-V and do we care about the difference?
+  if (opcode() == SpvOpTypePointer) {
+    ir::Instruction* base_type =
+        context()->get_def_use_mgr()->GetDef(GetSingleWordInOperand(1));
+    return base_type->IsVulkanStorageBuffer();
+  }
```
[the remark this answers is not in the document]
s-perron: Okay I've changed all of the the functions that identify descriptor types.  Now the return true only if they are a pointer type that points to <descriptor type>.  That way we can see the storage type on the OpTypePointer.

I did not see any documentation for the "new way" in the vulkan spec.  I implemented it based on your comment.

source/opt/instruction.cpp
```diff
@@ -133,5 +142,186 @@ void Instruction::ReplaceOperands(const std::vector<Operand>& new_operands) {
...
+        context()->get_def_use_mgr()->GetDef(GetSingleWordInOperand(1));
+    return base_type->IsVulkanUniformBuffer();
+  }
+
+  if (opcode() != SpvOpTypeStruct) {
+    return false;
+  }
+
+  bool is_block = false;
+  context()->get_decoration_mgr()->ForEachDecoration(
```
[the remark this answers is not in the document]
s-perron: See above.

source/opt/instruction.h
```diff
@@ -245,7 +248,48 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
...
+  bool IsReadOnlyLoad() const;
+
+  // Returns the instruction that gives the base address of an address
+  // calculation.  The instruction must be a load instruction.
+  Instruction* GetBaseAddress() const;
+
+  // Returns true if the instruction is a load from memory into a result id.
+  // This does not include memory-to-memory instructions.
+  inline bool IsLoad() const;
```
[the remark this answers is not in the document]
s-perron: commented that it considers core instructions only.  The longer name would be too cumbersome.  I prefer leaving the name as is.

source/opt/local_redundancy_elimination.h
```diff
@@ -0,0 +1,52 @@
...
+  const char* name() const override { return "local-redundancy-elimination"; }
+  Status Process(ir::IRContext*) override;
+
+ private:
+  // Deletes instructions in |block| whose value is in |vnTable| or is computed
+  // earlier in |block|. The values computed in |block| are added to |vnTable|.
+  // |value_to_ids| is a map from a value number to the result ids known to
+  // contain those values. The definition of the ids in value_to_ids must
+  // dominate |block|. One value needs to map to multiple ids because the ids
+  // may contain the same value, but have different decorations.
```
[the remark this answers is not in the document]
s-perron: done.

source/opt/value_number_table.cpp
dneto0: Derivative instructions are another case where there's communication with other invocations behind your back.  Treat them like a group instruction, or a volatile load for example.

s-perron: Done, I factored out the combinator stuff from ADCE into the IRContext.  Now it is generally available.

source/opt/value_number_table.cpp
```diff
@@ -0,0 +1,118 @@
...
+  if (lhs.NumInOperands() != rhs.NumInOperands()) {
+    return false;
+  }
+
+  for (uint32_t i = 0; i < lhs.NumInOperands(); ++i) {
+    if (lhs.GetInOperand(i) != rhs.GetInOperand(i)) {
+      return false;
+    }
```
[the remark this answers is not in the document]
s-perron: done.

source/opt/value_number_table.cpp
```diff
@@ -0,0 +1,118 @@
...
+  }
+  return true;
+}
+
+std::size_t ValueTableHash::operator()(
+    const spvtools::ir::Instruction& inst) const {
+  // We hash the opcode and in-operands, not the result, because we want
+  // instructions that are the same except for the result to hash to the same
+  // value.
```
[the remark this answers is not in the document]
s-perron: done

source/opt/value_number_table.cpp
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  // If not, assign it a new value number.
+  uint32_t value_number = TakeNextValueNumber();
+  id_to_value_[inst->result_id()] = value_number;
+  instruction_to_value_[*inst] = value_number;
+  return value_number;
+}
+
+bool ComputeSameValue::operator()(const ir::Instruction& lhs,
```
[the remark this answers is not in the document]
s-perron: I added a test for this.

source/opt/ir_context.cpp
```diff
@@ -183,5 +188,248 @@ void IRContext::KillNamesAndDecorates(Instruction* inst) {
...
+        SpvOpImageSparseDrefGather,
+        SpvOpImageSparseTexelsResident,
+        SpvOpImageSparseRead,
+        SpvOpSizeOf
+        // TODO(dneto): Add instructions enabled by ImageQuery
+    });
+  }
+}
+
+void IRContext::AddCombinatorsForExtesnion(ir::Instruction* extension) {
```
dneto0: Typo "Extension" is misspelled.
I can fix that before merge.

source/opt/ir_context.cpp
```diff
@@ -183,5 +188,248 @@ void IRContext::KillNamesAndDecorates(Instruction* inst) {
...
+        SpvOpNot,
+        SpvOpBitFieldInsert,
+        SpvOpBitFieldSExtract,
+        SpvOpBitFieldUExtract,
+        SpvOpBitReverse,
+        SpvOpBitCount,
+        SpvOpDPdx,
+        SpvOpDPdy,
+        SpvOpFwidth,
+        SpvOpDPdxFine,
```
dneto0: Something I learned today:  The derivative instructions exchange information with other invocations (in the same quad), so they should not be considered a combinator.  However, this commit is mainly a refactor, so I'll file a separate issue to fix that later.

source/opt/ir_context.h
```diff
@@ -263,6 +264,24 @@ class IRContext {
     return ++unique_id_;
   }
 
+  // Returns true if |inst| is a combinator in the current context.
+  // |combinator_ops_| is built if it has not been arleady.
```
dneto0: "already"

dneto0:
I have a squashed version that has typo fixes.  I'll push that into master once the bots come back clean.

I have filed https://github.com/KhronosGroup/SPIRV-Tools/issues/994 to remove derivatives from the list of combinators.

dneto0:
Build failure on Windows.

C:\projects\spirv-tools\source\opt\ir_context.cpp(419): error C2593: 'operator =' is ambiguous [C:\projects\spirv-tools\build\source\opt\SPIRV-Tools-opt.vcxproj] 

See AppVeyor logs

Please fix that error and the typos.

source/opt/ir_context.cpp
s-perron: Fixed.

source/opt/ir_context.h
s-perron: fixed.

Issue #968, opened 2017-11-15
Need legalization pass recipe in optimizer header


antiagainst:
Right now we have `RegisterPerformancePasses()` and `RegisterSizePasses()` in `optimizer.hpp`.

It would also be nice to have a `RegisterLegalizationPasses()` so that both glslang and spiregg can share the same source of truth. Right now they are separately encoded in both projects:

https://github.com/KhronosGroup/glslang/blob/4036ef154fb0e11a2b235afce3e9b3b819cf5fea/SPIRV/GlslangToSpv.cpp#L6060-L6071

https://github.com/Microsoft/DirectXShaderCompiler/blob/6788ebea4a3a4652d2b182ffe99bdd1bf3a1451c/tools/clang/lib/SPIRV/SPIRVEmitter.cpp#L185-L203

dneto0:
#980 was merged

PR #980, opened 2017-11-17
Add RegisterLegalizationPasses() into the interface

antiagainst:
Addresses #968

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -77,11 +77,6 @@ class Optimizer {
   // method.
   Optimizer& RegisterPass(PassToken&& pass);
 
-  // Registers passes that attempt to legalize the generated code.
-  // This sequence of passes is subject to constant review and will change
-  // from time to time.
-  Optimizer& RegisterLegalizationPasses();
-
   // Registers passes that attempt to improve performance of generated code.
   // This sequence of passes is subject to constant review and will change
   // from time to time.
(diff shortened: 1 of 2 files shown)


antiagainst:
This is the same recipe as used in glslang right now: https://github.com/KhronosGroup/glslang/blob/4036ef154fb0e11a2b235afce3e9b3b819cf5fea/SPIRV/GlslangToSpv.cpp#L6060-L6071

antiagainst:
cc @greg-lunarg

s-perron:
Do we have any ideas which ones are really for legalization and which ones for optimization?  This looks like the list that recipe that LunarG suggests to their customer and what we used for the performance and size optimization recipes.

include/spirv-tools/optimizer.hpp
```diff
@@ -77,6 +77,11 @@ class Optimizer {
   // method.
   Optimizer& RegisterPass(PassToken&& pass);
 
+  // Registers passes that attempt to legalize the generated code.
+  // This sequence of passes is subject to constant review and will change
```
dneto0: It's worth saying these are passes used to legalize code originally generated by legacy HLSL.  Also say where to find a more current list, i.e. in Glslang

antiagainst:
@s-perron: Legalization passes are largely general optimization passes since the transformations needed to legalize are similar: function inlining, load/store eliminating, const propagating and folding, DCE, etc. But legalization emphasizes transformations that flatten structs/arrays, which I'd assume performance/size optimization does not. Also legalization does not really care about loop unrolling and other performance oriented passes. So there are subtle differences.

Right now these recipes are almost identical. But they will drift away gradually I think.

dnovillo:
I wish they were not referred to "optimizations", though.  I understand
that they are transformations required to produce canonical/legal vulkan.
Perhaps we can call them "legalizing transformations"?  I'm being picky,
but the term "optimization" is very specific.  It does not make much sense
to need an optimization for correctness.  They should be orthogonal.

OK, stepping off the soap box.

antiagainst:
Haha, yes, I totally agree. You'll notice that in my previous reply I've used optimization/legalization/transformation carefully. ;-P

include/spirv-tools/optimizer.hpp
antiagainst: Done.

dneto0:
Rebased, squashed, and pushed into master as aec60b8158

Issue #979, opened 2017-11-17
Opt: Optimize loads/stores through nested structs


greg-lunarg:
When glslang stops flattening structs with opaques types, we will need to optimize nested structs as we now optimize non-nested structs.

See hlsl.flattenSubset.frag test in glslang.

Original report from @johnkslang

I claim this issue.

PR #981, opened 2017-11-18
Optimize loads/stores on nested structs

greg-lunarg:
Enhance ExtractInsertElimination to optimize sequences created for nested structs.

Addresses legalization issue #979 reported by @johnkslang 

I will add some test cases shortly, but thought I would get the review started.


s-perron:
I have been thinking about this.  I believe something like this is needed.  I'll take a closer look when the test cases are added.  The only things that jumps out at me right now is the heavy use of abbreviated names made is hard for me to follow at first.

greg-lunarg:
I have now adapted/added tests for nested structs.

dneto0:
Rebased, squashed, and pushed into master as e28edd458

PR #985, opened 2017-11-20
Generic value propagation engine.

dnovillo:
_**NOTE: I am adding tests and debugging at the moment.  I'm sending this initial version for review and discussion.  This will be used as the basis for SSA-CCP, copy propagation and range propagation (if we decide to add it).**_

This class implements a generic value propagation algorithm based on the
conditional constant propagation algorithm proposed in

     Constant propagation with conditional branches,
     Wegman and Zadeck, ACM TOPLAS 13(2):181-210.

The implementation is based on

     A Propagation Engine for GCC
     Diego Novillo, GCC Summit 2005
     http://ols.fedoraproject.org/GCC/Reprints-2005/novillo-Reprint.pdf

The purpose of this implementation is to act as a common framework for any
transformation that needs to propagate values from statements producing new
values to statements using those values.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -300,7 +300,7 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
   for (auto bi = structuredOrder.begin(); bi != structuredOrder.end(); ++bi) {
     for (auto ii = (*bi)->begin(); ii != (*bi)->end(); ++ii) {
       if (IsLive(&*ii)) continue;
-      if (IsBranch(ii->opcode()) &&
+      if (ii->IsBranch() &&
           !IsStructuredIfHeader(*bi, nullptr, nullptr, nullptr))
         continue;
       dead_insts_.insert(&*ii);
--- a/source/opt/basic_block.h
+++ b/source/opt/basic_block.h
@@ -138,6 +138,10 @@ class BasicBlock {
   // this block, if any.  If none, returns zero.
   uint32_t ContinueBlockIdIfAny() const;
 
+  // Returns true if this basic block exits this function and returns to its
+  // caller.
+  bool IsReturn() const { return ctail()->IsReturn(); }
+
  private:
   // The enclosing function.
   Function* function_;
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -295,6 +295,19 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   // Returns true if the instruction is an atom operation.
   inline bool IsAtomicOp() const;
 
+  // Returns true if this instruction is a branch or switch instruction (either
+  // conditional or not).
+  bool IsBranch() const { return spvOpcodeIsBranch(opcode()); }
+
+  // Returns true if this instruction causes the function to finish execution
+  // and return to its caller
+  bool IsReturn() const { return spvOpcodeIsReturn(opcode()); }
+
+  // Returns true if this instruction is a basic block terminator.
+  bool IsBlockTerminator() const {
+    return spvOpcodeIsBlockTerminator(opcode());
+  }
+
   inline bool operator==(const Instruction&) const;
   inline bool operator!=(const Instruction&) const;
   inline bool operator<(const Instruction&) const;
@@ -344,8 +357,7 @@ inline const Operand& Instruction::GetOperand(uint32_t index) const {
 };
 
 inline void Instruction::AddOperand(Operand&& operand) {
-  operands_.push_back(operand);
-  return;
+  operands_.push_back(std::move(operand));
 }
 
 inline void Instruction::SetInOperand(uint32_t index,
--- a/source/opt/propagator.cpp
+++ b/source/opt/propagator.cpp
@@ -1,0 +1,232 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "propagator.h"
+
+namespace spvtools {
+namespace opt {
+
+void SSAPropagator::AddControlEdge(const Edge& edge) {
+  ir::BasicBlock* dest_bb = edge.dest;
+
+  // Refuse to add the exit block to the work list.
+  if (dest_bb == cfg_->pseudo_exit_block()) {
+    return;
+  }
+
+  // Try to mark the edge executable.  If it was already in the set of
+  // executable edges, do nothing.
+  if (!MarkEdgeExecutable(edge)) {
+    return;
+  }
+
+  // If the edge had not already been marked executable, add the destination
+  // basic block to the work list.
+  blocks_.push(dest_bb);
+}
+
+void SSAPropagator::AddSSAEdges(uint32_t id) {
+  get_def_use_mgr()->ForEachUser(id, [this](ir::Instruction *instr) {
+    // If the basic block for |instr| has not been simulated yet, do nothing.
+    if (!BlockHasBeenSimulated(ctx_->get_instr_block(instr))) {
+      return;
+    }
+
+    if (ShouldSimulateAgain(instr)) {
+      ssa_edge_uses_.push(instr);
+    }
+  });
+}
+
+bool SSAPropagator::Simulate(ir::Instruction* instr) {
+  bool changed = false;
+
+  // Don't bother visiting instructions that should not be simulated again.
+  if (!ShouldSimulateAgain(instr)) {
+    return changed;
+  }
+
+  ir::BasicBlock* dest_bb = nullptr;
+  PropStatus status = visit_fn_(instr, &dest_bb);
+
+  if (status == kVarying) {
+    // The statement produces a varying result, add it to the list of statements
+    // not to simulate anymore and add its SSA def-use edges for simulation.
+    DontSimulateAgain(instr);
+    if (instr->result_id() > 0) {
+      AddSSAEdges(instr->result_id());
+    }
+
+    // If |instr| is a block terminator, add all the control edges out of its
+    // block.
+    if (instr->IsBlockTerminator()) {
+      ir::BasicBlock* block = ctx_->get_instr_block(instr);
+      for (const auto& e : bb_succs_.at(block)) {
+        AddControlEdge(e);
+      }
+    }
+    return false;
+  } else if (status == kInteresting) {
+    // If the instruction produced a new interesting value, add the SSA edge
+    // for its result ID.
+    if (instr->result_id() > 0) {
+      AddSSAEdges(instr->result_id());
+    }
+
+    // If there are multiple outgoing control flow edges and we know which one
+    // will be taken, add the destination block to the CFG work list.
+    if (dest_bb) {
+      blocks_.push(dest_bb);
+    }
+    changed = true;
+  }
+
+  // At this point, we are dealing with instructions that are in status
+  // kInteresting or kNotInteresting.  To decide whether this instruction should
+  // be simulated again, we examine its operands.  If at least one operand O is
+  // defined at an instruction D that should be simulated again, then the output
+  // of D might affect |instr|, so we should simulate |instr| again.
+  bool has_operands_to_simulate = false;
+  ir::BasicBlock* instr_bb = ctx_->get_instr_block(instr);
+  if (instr->opcode() == SpvOpPhi) {
+    // For Phi instructions, an operand causes the Phi to be simulated again if
+    // the operand comes from an edge that has not yet been traversed or if its
+    // definition should be simulated again.
+    for (uint32_t i = 2; i < instr->NumOperands(); i += 2) {
+      // Phi arguments come in pairs. Index 'i' contains the
+      // variable id, index 'i + 1' is the originating block id.
+      assert(i % 2 == 0 && i < instr->NumOperands() - 1 &&
+             "malformed Phi arguments");
+
+      uint32_t arg_id = instr->GetSingleWordOperand(i);
+      ir::Instruction* arg_def_instr = get_def_use_mgr()->GetDef(arg_id);
+      uint32_t in_label_id = instr->GetSingleWordOperand(i + 1);
+      ir::Instruction* in_label_instr = get_def_use_mgr()->GetDef(in_label_id);
+      ir::BasicBlock* in_bb = ctx_->get_instr_block(in_label_instr);
+      Edge edge(in_bb, instr_bb);
+
+      if (!IsEdgeExecutable(edge) || ShouldSimulateAgain(arg_def_instr)) {
+        has_operands_to_simulate = true;
+        break;
+      }
+    }
+  } else {
+    // For regular instructions, check if the defining instruction of each
+    // operand needs to be simulated again.  If so, then this instruction should
+    // also be simulated again.
+    instr->ForEachInId([&has_operands_to_simulate, this](const uint32_t* use) {
+      ir::Instruction* def_instr = get_def_use_mgr()->GetDef(*use);
+      if (ShouldSimulateAgain(def_instr)) {
+        has_operands_to_simulate = true;
+        return;
+      }
+    });
+  }
+
+  if (!has_operands_to_simulate) {
+    DontSimulateAgain(instr);
+  }
+
+  return changed;
+}
+
+bool SSAPropagator::Simulate(ir::BasicBlock* block) {
+  if (block == cfg_->pseudo_exit_block()) {
+    return false;
+  }
+
+  // Always simulate Phi instructions, even if we have simulated this block
+  // before. We do this because Phi instructions receive their inputs from
+  // incoming edges. When those edges are marked executable, the corresponding
+  // operand can be simulated.
+  bool changed = false;
+  block->ForEachPhiInst(
+      [&changed, this](ir::Instruction* instr) { changed |= Simulate(instr); });
+
+  // If this is the first time this block is being simulated, simulate every
+  // statement in it.
+  if (!BlockHasBeenSimulated(block)) {
+    block->ForEachInst([this, &changed](ir::Instruction* instr) {
+      if (instr->opcode() != SpvOpPhi) {
+        changed |= Simulate(instr);
+      }
+    });
+
+    MarkBlockSimulated(block);
+
+    // If this block has exactly one successor, mark the edge to its successor
+    // as executable.
+    if (bb_succs_.at(block).size() == 1) {
+      AddControlEdge(bb_succs_.at(block).at(0));
+    }
+  }
+
+  return changed;
+}
+
+void SSAPropagator::Initialize(ir::Function* fn) {
+  // Compute predecessor and successor blocks for every block in |fn|'s CFG.
+  // TODO(dnovillo): Move this to ir::CFG and always build them. Alternately,
+  // move it to IRContext and build CFG preds/succs on-demand.
+  bb_succs_[cfg_->pseudo_entry_block()].push_back(
+      Edge(cfg_->pseudo_entry_block(), fn->entry().get()));
+
+  for (auto& block : *fn) {
+    block.ForEachSuccessorLabel([this, &block](uint32_t label_id) {
+      ir::BasicBlock* succ_bb =
+          ctx_->get_instr_block(get_def_use_mgr()->GetDef(label_id));
+      bb_succs_[&block].push_back(Edge(&block, succ_bb));
+      bb_preds_[succ_bb].push_back(Edge(succ_bb, &block));
+    });
+    if (block.IsReturn()) {
+      bb_succs_[&block].push_back(Edge(&block, cfg_->pseudo_exit_block()));
+      bb_preds_[cfg_->pseudo_exit_block()].push_back(
+          Edge(cfg_->pseudo_exit_block(), &block));
+    }
+  }
+
+  // Add the edges out of the entry block to seed the propagator.
+  const auto& entry_succs = bb_succs_[cfg_->pseudo_entry_block()];
+  for (const auto& e : entry_succs) {
+    AddControlEdge(e);
+  }
+}
+
+bool SSAPropagator::Run(ir::Function* fn) {
+  Initialize(fn);
+
+  bool changed = false;
+  while (!blocks_.empty() || !ssa_edge_uses_.empty()) {
+    // Simulate all blocks first. Simulating blocks will add SSA edges to
+    // follow after all the blocks have been simulated.
+    if (!blocks_.empty()) {
+      auto block = blocks_.front();
+      changed |= Simulate(block);
+      blocks_.pop();
+      continue;
+    }
+
+    // Simulate edges from the SSA queue.
+    if (!ssa_edge_uses_.empty()) {
+      ir::Instruction* instr = ssa_edge_uses_.front();
+      changed |= Simulate(instr);
+      ssa_edge_uses_.pop();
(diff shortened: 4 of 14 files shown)


source/opt/instruction.h
```diff
@@ -265,7 +280,7 @@ inline const Operand& Instruction::GetOperand(uint32_t index) const {
   return operands_[index];
 };
 
-inline void Instruction::AddOperand(Operand &&operand) {
+inline void Instruction::AddOperand(Operand&& operand) {
   operands_.push_back(operand);
```
atgoo: This probably was supposed to be
operands_.push_back(std::move(operand));

source/opt/propagator.cpp
```diff
@@ -0,0 +1,215 @@
...
+// limitations under the License.
+
+#include "propagator.h"
+
+namespace spvtools {
+namespace opt {
+
+template <typename ValueType>
+void SSAPropagator<ValueType>::AddControlEdge(const Edge& edge) {
+  const ir::BasicBlock *dest_bb = edge.dst;
```
atgoo: const ir::BasicBlock*

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+
+#include "cfg.h"
+#include "ir_context.h"
+#include "module.h"
+
+namespace spvtools {
+namespace opt {
+
+// Represents a CFG control edge. TODO(dnovillo): Move it to ir::CFG and always
+// build it. Alternately, move it to IRContext and build it on-demand.
```
s-perron: The TODO is not clear what is being built.  This is just a CFG edge.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+  bool operator==(const Edge& o) const { return src == o.src && dst == o.dst; }
+};
+
+}  // namespace opt
+}  // namespace spvtools
+
+// Hashing operator for Edge. Needed to instantiate unordered sets of Edges.
+namespace std {
+template <>
+struct std::hash<spvtools::opt::Edge> {
```
atgoo: You are already in the namespace.
std::hash -> hash

source/opt/propagator.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  if (status == Varying) {
+    // The statement produces a varying result, add it to the list of statements
+    // not to simulate anymore and add its SSA def-use edges for simulation.
+    DontSimulateAgain(instr);
+    if (instr->result_id() > 0) {
+      AddSSAEdges(instr->result_id());
+    }
+
+    /* If STMT transfers control out of its basic block, add
```
s-perron: STMT?

source/opt/propagator.cpp
```diff
@@ -0,0 +1,215 @@
...
+template <typename ValueType>
+void SSAPropagator<ValueType>::Simulate(ir::Instruction* instr) {
+  // Don't bother visiting instructions that should not be simulated again.
+  if (!ShouldSimulateAgain(instr)) {
+    return;
+  }
+
+  ir::BasicBlock* dest_bb = nullptr;
+  uint32_t result_id = 0;
+  PropStatus status = visit_fn_(instr, &dest_bb, &result_id);
```
s-perron: You pass in the address |result_id|, but |result_id| is not used afterward.  Why do you need it?

source/opt/propagator.cpp
```diff
@@ -0,0 +1,215 @@
...
+void SSAPropagator<ValueType>::Simulate(ir::BasicBlock* block) {
+  if (block == cfg_->pseudo_exit_block()) {
+    return;
+  }
+
+  // Always simulate Phi instructions, even if we have simulated this block
+  // before. We do this because Phi instructions receive their inputs from
+  // incoming edges. When those edges are marked executable, the corresponding
+  // operand can be evaluated.
+  block->ForEachPhiInst([this](ir::Instruction* instr) { Simulate(instr); });
```
s-perron: I don't understand why this is necessary.  Won't the phi nodes get added to "sss_edges_" when its operands are simulated.  So the phi nodes will get processed anyway if they need to.  Is that right?

source/opt/instruction.h
```diff
@@ -241,6 +242,20 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   // Returns true if the instruction annotates an id with a decoration.
   inline bool IsDecoration();
 
+  // Returns true if this instruction is a branch instruction (either
```
alan-baker: Mention switch support.

source/opt/propagator.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  // If this is the frist time this block is being simulated, evaluate every
```
alan-baker: frist -> first

source/opt/propagator.cpp
```diff
@@ -0,0 +1,215 @@
...
+void SSAPropagator<ValueType>::Initialize(ir::Function* fn) {
+  cfg_->ComputeStructuredOrder(fn, cfg_->pseudo_entry_block(),
+                                ordered_blocks_);
+
+  // Compute predecessor and successor blocks for every block in |fn|'s CFG.
+  for (const auto& block : *fn) {
+    block.ForEachSuccessorLabel([this, &block](uint32_t label_id) {
+      ir::BasicBlock* succ_bb =
+          ctx_->get_instr_block(get_def_use_mgr()->GetDef(label_id));
+      bb_succs_[block].push(Edge(&block, succ_bb));
```
alan-baker: Shouldn't the CFG manage these edges? Is that future work?

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+template <typename ValueType>
+class SSAPropagator {
+ public:
+  // Lattice values used for propagation. See class documentation for
+  // a description.
+  enum PropStatus { NotInteresting, Interesting, Varying };
+
+  using VisitFunction = std::function<PropStatus(ir::Instruction*)>;
+
+  SSAPropagator(ir::IRContext* context, ir::CFG* cfg, VisitFunction& visit_fn)
```
alan-baker: const VisitFunction&

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+  // CFG blocks in structured order. Every block in this list comes after its
+  // dominators.
+  std::list<ir::BasicBlock*> ordered_blocks_;
+
+  // Set of instructions that should not be simulated again because they have
+  // been found to be in the Varying state.
+  std::unordered_set<ir::Instruction*> do_not_simulate_;
+
+  // Map between a basic block and its predecessor edges.
+  std::unordered_map<ir::BasicBlock*, std::list<Edge>> bb_preds_;
```
alan-baker: Please repeat the todo about moving these into CFG (which is to be moved into IRContext).

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+}  // namespace opt
+}  // namespace spvtools
+
+// Hashing operator for Edge. Needed to instantiate unordered sets of Edges.
+namespace std {
+template <>
+struct std::hash<spvtools::opt::Edge> {
+  size_t operator()(const spvtools::opt::Edge& e) const {
+    return std::hash<const spvtools::ir::BasicBlock*>()(e.src) ^
+           std::hash<const spvtools::ir::BasicBlock*>()(e.dst);
```
atgoo: Unless the hash is supposed to be symmetric I would prefer
hash1 * kConst + hash2.

source/opt/instruction.h
```diff
@@ -15,12 +15,13 @@
 #ifndef LIBSPIRV_OPT_INSTRUCTION_H_
 #define LIBSPIRV_OPT_INSTRUCTION_H_
 
+#include <opcode.h>
```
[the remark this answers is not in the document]
dnovillo: Done.

source/opt/instruction.h
dnovillo: Done.

source/opt/propagator.cpp
dnovillo: Fixed.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+
+  // Gets the value for |id|. If |id| has not been assigned a value, it
+  // throws a runtime assertion.
+  ValueType GetValue(uint32_t id) const { return values_.at(id); }
+
+ private:
+  // Initialize processing.
+  void Initialize(ir::Function* fn);
+
+  // Simulate the execution of |block| by evaluating every instruction in it.
```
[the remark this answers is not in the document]
dnovillo: They are.  Changed to use 'simulate' everywhere.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+    return do_not_simulate_.find(instr) != do_not_simulate_.end();
+  }
+
+  // Add |instr| to the set of instructions not to simulate again.
+  void DontSimulateAgain(ir::Instruction* instr) {
+    do_not_simulate_.insert(instr);
+  }
+
+  // Returns true if |block| has been simulated already.
+  bool IsBlockSimulated(ir::BasicBlock* block) {
```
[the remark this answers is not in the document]
dnovillo: I'm not sure how else to say this.  The predicate does precisely that.  If every instruction has been simulated, it returns true.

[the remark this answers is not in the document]
dnovillo: May be BlockHasBeenSimulated?

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  // Marks block |block| as simulated.
+  void MarkBlockSimulated(ir::BasicBlock* block) {
+    simulated_blocks_.insert(block);
+  }
+
+  // Marks |edge| as executable.  Returns false if the edge was already marked
+  // as executable. Otherwise, the edge is marked executable for the first time
+  // and it returns true.
```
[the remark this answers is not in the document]
dnovillo: Fixed.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+  bool MarkEdgeExecutable(const Edge& edge) {
+    return executable_edges_.insert(edge).second;
+  }
+
+  // Returns true if |edge| has been marked as executable.
+  bool IsEdgeExecutable(const Edge& edge) {
+    return executable_edges_.find(edge) != executable_edges_.end();
+  }
+
+  // Return a pointer to the def-use manager for this context.
```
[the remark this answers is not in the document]
dnovillo: Fixed.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+  analysis::DefUseManager* get_def_use_mgr() const {
+    return ctx_->get_def_use_mgr();
+  }
+
+  // Add the destination block of the CFG edge |e| to the work list.  If |e|
+  // had already been executed, do nothing.
```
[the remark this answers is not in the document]
dnovillo: Done.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+  void AddControlEdge(const Edge& e);
+
+  // Add all the use statements at the end of SSA edges coming out the
+  // definition for |id|.
```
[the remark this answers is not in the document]
dnovillo: Paper terminology.  Fixed.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+  void AddSSAEdges(uint32_t id);
+
+  // Function that visit instructions during simulation. The output of this
+  // function is used to determine if the simulated instruction produced a value
+  // interesting for propagation. The actual value produced by the instruction
+  // must be stored in the |values_| map.
```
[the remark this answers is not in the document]
dnovillo: Not needed anymore.  This was the only thing that was forcing the whole class to be a template.  It's unnecessary, so I moved this as a requirement for the user-provided callback.

source/opt/instruction.h
dnovillo: Done.

source/opt/propagator.cpp
dnovillo: Done

source/opt/propagator.h
dnovillo: Done.

source/opt/propagator.cpp
dnovillo: Fixed.

source/opt/propagator.cpp
dnovillo: Phi nodes get operands enabled not only by SSA edges, a Phi operand becomes available also when the incoming CFG edge is marked executable.  That's why this code checks which operands have an executable predecessor.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+namespace spvtools {
+namespace opt {
+
+// Represents a CFG control edge. TODO(dnovillo): Move it to ir::CFG and always
+// build it. Alternately, move it to IRContext and build it on-demand.
+struct Edge {
+  explicit Edge(const ir::BasicBlock* b1, const ir::BasicBlock* b2)
+      : src(b1), dst(b2) {}
+  const ir::BasicBlock* src;
+  const ir::BasicBlock* dst;
```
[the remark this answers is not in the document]
dnovillo: Fixed.

source/opt/propagator.h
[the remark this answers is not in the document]
dnovillo: I've taken the easy way out and I'm using std::set for edges.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+
+  // Run the propagator on function |fn|. Returns true if changes were made to
+  // the function. Otherwise, it returns false.
+  bool Run(ir::Function* fn);
+
+  // Sets the value for |id| to |value|.
+  void SetValue(uint32_t id, ValueType value) { values_[id] = value; }
+
+  // Gets the value for |id|. If |id| has not been assigned a value, it
+  // throws a runtime assertion.
```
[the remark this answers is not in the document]
dnovillo: Fixed. Moved this out of the propagator. The callback function is responsible for keeping track of the value table.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+  void Simulate(ir::BasicBlock* block);
+
+  // Simulate the execution of |instr| by replacing all the known values in
+  // every operand and determining whether the result is interesting for
+  // propagation. This invokes the callback function |visit_fn_| to determine
+  // the value computed by |instr|.
+  void Simulate(ir::Instruction* instr);
+
+  // Returns true if |instr| should be simulated again. This tests whether the
+  // instruction is in the set of instructions not to simulate again.
```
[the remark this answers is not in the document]
dnovillo: Fixed.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+
+  // Add the destination block of the CFG edge |e| to the work list.  If |e|
+  // had already been executed, do nothing.
+  void AddControlEdge(const Edge& e);
+
+  // Add all the use statements at the end of SSA edges coming out the
+  // definition for |id|.
+  void AddSSAEdges(uint32_t id);
+
+  // Function that visit instructions during simulation. The output of this
```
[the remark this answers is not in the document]
dnovillo: Fixed.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+  // IR context to use.
+  ir::IRContext* ctx_;
+
+  // CFG to propagate values on. TODO(dnovillo): The CFG should be part of
+  // IRContext.
+  ir::CFG* cfg_;
+
+  // SSA def-use edges to traverse. Each entry is a destination statement for an
+  // SSA def-use edge as returned by |def_use_manager_|.
+  std::queue<ir::Instruction*> ssa_edges_;
```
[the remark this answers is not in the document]
dnovillo: Renamed to ssa_edge_uses_

source/opt/propagator.cpp
dnovillo: Fixed.

source/opt/propagator.h
dnovillo: Fixed.

source/opt/propagator.h
dnovillo: Done.

source/opt/propagator.cpp
```diff
@@ -0,0 +1,215 @@
...
+      bb_preds_[succ_bb].push(Edge(succ_bb, &block));
+    });
+  }
+
+  // Add the edges out of the entry block to seed the propagator.
+  std::list<Edge> entry_succs = bb_succs_[fn->entry()];
```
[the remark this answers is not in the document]
dnovillo: Fixed.

source/opt/propagator.cpp
```diff
@@ -0,0 +1,215 @@
...
+  for (const auto& e : bb_succs_[fn->entry()]) {
```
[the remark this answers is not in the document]
dnovillo: Fixed.

source/opt/propagator.h
```diff
@@ -0,0 +1,242 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_OPT_PROPAGATOR_H_
+#define LIBSPIRV_OPT_PROPAGATOR_H_
+
+#include <functional>
+#include <unordered_map>
+#include <unordered_set>
+#include <queue>
```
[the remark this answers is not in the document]
dnovillo: Fixed.

source/opt/propagator.h
dnovillo: Removed.  Using std::set.

source/opt/propagator.h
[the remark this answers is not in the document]
dnovillo: Fixed.

source/opt/propagator.cpp
```diff
@@ -0,0 +1,228 @@
...
+  // move it to IRContext and build CFG preds/succs on-demand.
+  bb_succs_[cfg_->pseudo_entry_block()].push_back(
+      Edge(cfg_->pseudo_entry_block(), fn->entry().get()));
+
+  for (auto& block : *fn) {
+    block.ForEachSuccessorLabel([this, &block](uint32_t label_id) {
+      ir::BasicBlock* succ_bb =
+          ctx_->get_instr_block(get_def_use_mgr()->GetDef(label_id));
+      bb_succs_[&block].push_back(Edge(&block, succ_bb));
+      bb_preds_[succ_bb].push_back(Edge(succ_bb, &block));
```
atgoo: I don't see bb_preds_ being used anywhere. Should bb_preds_ also feature pseudo-blocks?

dnovillo: Good point.  Added.  Predecessors are not really used in the propagator, but I'm thinking of moving this into ir::CFG and I *think* I may need it in SSA-CCP itself.

dnovillo:
@s-perron, @alan-baker with the new tests, this PR is ready for more reviews.  PTAL.  Thanks.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -321,7 +321,7 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
   for (auto bi = structuredOrder.begin(); bi != structuredOrder.end(); ++bi) {
     for (auto ii = (*bi)->begin(); ii != (*bi)->end(); ++ii) {
       if (IsLive(&*ii)) continue;
-      if (IsBranch(ii->opcode()) &&
+      if (ii->IsBranch() &&
```
alan-baker: You've extended the functionality of this pass. You should test it with an OpSwitch.

source/opt/basic_block.h
```diff
@@ -138,6 +138,10 @@ class BasicBlock {
   // this block, if any.  If none, returns zero.
   uint32_t ContinueBlockIdIfAny() const;
 
+  // Returns true if this basic block exits this function and returns to its
+  // caller.
+  bool IsReturn() const { return insts_.back().IsReturn(); }
```
alan-baker: return tail()->IsReturn(). Fewer restrictions on the underlying implementation of insts_.

source/opt/instruction.h
```diff
@@ -247,6 +247,20 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
...
+  // conditional or not).
+  bool IsBranch() const { return spvOpcodeIsBranch(opcode()); }
+
+  // Returns true if this instruction causes the function to finish execution
+  // and return to its caller
+  bool IsReturn() const { return spvOpcodeIsReturn(opcode()); }
+
+  // Returns true if this instruction is a basic block terminator.
+  bool IsBlockTerminator() const {
+    return IsBranch() || IsReturn() || opcode() == SpvOpKill ||
```
alan-baker: Consider making this congruent with the other tests.

source/opt/propagator.h
```diff
@@ -0,0 +1,248 @@
...
+//              contains the statements at the end of a def-use edge that need
+//              to be re-visited when an instruction produces a kVarying or
+//              kInteresting result.
+//
+// 4- Simulation terminates when all work queues are drained.
+//
+//
+// EXAMPLE: Basic constant store propagator.
+//
+// Suppose we want to propagate all constant assignments of the form "OpStore
```
s-perron: I don't know if this is a good example.  No propagating seems to be going on, and it is unclear what should happen if there are two store to the same variable with different constants.  If I was trying to figure out how my |value_fn| function is suppose to work so I can use the propagator, I would be confused.

dnovillo:
On Thu, Nov 23, 2017 at 12:00 PM, Steven Perron <notifications@github.com>
wrote:

>
> In source/opt/propagator.h
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/985#discussion_r152846942>
> :
>
> > +//      CFG blocks: contains the queue of blocks to be simulated.
> +//             Blocks are added to this queue if their incoming edges are
> +//             executable.
> +//
> +//      SSA Edges: An SSA edge is a def-use edge between a value-producing
> +//              instruction and its use instruction.  The SSA edges list
> +//              contains the statements at the end of a def-use edge that need
> +//              to be re-visited when an instruction produces a kVarying or
> +//              kInteresting result.
> +//
> +// 4- Simulation terminates when all work queues are drained.
> +//
> +//
> +// EXAMPLE: Basic constant store propagator.
> +//
> +// Suppose we want to propagate all constant assignments of the form "OpStore
>
> I don't know if this is a good example. No propagating seems to be going
> on, and it is unclear what should happen if there are two store to the same
> variable with different constants. If I was trying to figure out how my
> |value_fn| function is suppose to work so I can use the propagator, I would
> be confused.
>

I have to balance between a very detailed example vs an example that gives
a brief idea of how it works.  I'll try to add a more complete example.
But, in the end, the paper references are the ones to go read when
implementing your own propagation.

dnovillo:
On Thu, Nov 23, 2017 at 12:13 PM, Steven Perron <notifications@github.com>
wrote:

> *@s-perron* requested changes on this pull request.
>
> I'm still unclear about the requirements that the visit function has the
> meet. For example, what happens if it returns that all branch instructions
> are not interesting? Looking at the implementation of the propagator, it
> seems like a whole lot of blocks will not be visited. Is that intended? If
> so, more explanation is needed. If not it needs to be fixed up.
>

Yes, that's exactly what happens.  I'll rephrase and re-arrange to make it
more clear.  All the requirements are there now.  I'll re-arrange.

dnovillo:
@s-perron added more details on the user-provided visit function.  I decided against creating a more detailed example in the docs.  I prefer to have the tests be that example.  The tests have the advantage that they will be updated with any API changes.

Will address the comments from @alan-baker

s-perron:
Need to make the builds bot happy, but otherwise LGTM.

source/opt/aggressive_dead_code_elim_pass.cpp
dnovillo: Added a new test, but it fails because ADCE does not seem to handle OpSwitch at all (#1021)

source/opt/basic_block.h
dnovillo: Done.

source/opt/instruction.h
dnovillo: Done.

dnovillo:
Re-based, re-tested and pushed to master as 74327845aa045ea854186af286a7f34f0125d6a4

dneto0:
This MR is proximate cause of Windows debug build failure.  PTAL @dnovillo

PR #986, opened 2017-11-20
Fix move semantics in iterator make_range

atgoo:


dneto0:
Rebased and pushed into master as b14291581

PR #987, opened 2017-11-21
Speed up DefUseManager for large modules

alan-baker:
Reworks storage of use lists for #846.

Replaces unordered_map<uint32_t, list<Use>> with set<pair<Instruction*,Instruction*>>. Storage now only stores def -> user. ForEachUser and ForEachUse replace GetUses. ForEachUse extends ForEachUser by scanning operands.

A side effect of the change is that definitions must be registered with the DefUseManager before analyzing or iterating over the uses.

Note: I still need to analyze the windows run time impact of this change (the original impetus for this issue). I will not finalize this review unless the speed up is significant.


s-perron:
FYI, on the linux build before your change (https://travis-ci.org/KhronosGroup/SPIRV-Tools/jobs/305003853):

82/82 Test #19: spirv-tools-test_link_global_values_amount ...........   Passed   **15.11 sec**

With your change (https://travis-ci.org/KhronosGroup/SPIRV-Tools/jobs/305055686):

33/82 Test #19: spirv-tools-test_link_global_values_amount ...........   Passed    **0.94 sec**

source/opt/def_use_manager.cpp
```diff
@@ -76,33 +76,71 @@ const ir::Instruction* DefUseManager::GetDef(uint32_t id) const {
...
 std::vector<ir::Instruction*> DefUseManager::GetAnnotations(uint32_t id) const {
   std::vector<ir::Instruction*> annos;
-  const auto* uses = GetUses(id);
-  if (!uses) return annos;
-  for (const auto& c : *uses) {
-    if (ir::IsAnnotationInst(c.inst->opcode())) {
-      annos.push_back(c.inst);
+  if (!GetDef(id)) return annos;
+
+  ForEachUser(id, [&annos](ir::Instruction* user) {
```
atgoo: Consider
```
const auto* def = GetDef(id);
if (!def) return annos;
ForEachUser(def, [&annos](ir::Instruction* user) {
```

source/opt/def_use_manager.cpp
```diff
@@ -111,7 +149,14 @@ void DefUseManager::ClearInst(ir::Instruction* inst) {
...
     EraseUseRecordsOfOperandIds(inst);
     if (inst->result_id() != 0) {
-      id_to_uses_.erase(inst->result_id());  // Remove all uses of this id.
+      // Remove all uses of this inst.
+      auto user_end = id_to_users_.end();
+      auto user_iter = id_to_users_.lower_bound(UserEntry(inst, nullptr));
+      while (user_iter != user_end && user_iter->first == inst) {
+        // Increment to next element before invalidating iterator.
+        auto use = user_iter++;
+        id_to_users_.erase(use);
```
atgoo: Consider
`user_iter = id_to_users_.erase(user_iter);`
or 
```
auto begin = id_to_users_.lower_bound(UserEntry(inst, nullptr));
auto end = begin;
while (end != user_end && end->first == inst) { end++; }
id_to_users_.erase(begin, end);
```
(likely has better time complexity)

source/opt/def_use_manager.h
```diff
@@ -99,7 +144,9 @@ class DefUseManager {
   // Returns the map from ids to their def instructions.
   const IdToDefMap& id_to_defs() const { return id_to_def_; }
   // Returns the map from ids to their uses in instructions.
-  const IdToUsesMap& id_to_uses() const { return id_to_uses_; }
+  //const IdToUsesMap& id_to_uses() const { return id_to_uses_; }
```
atgoo: The line is not needed anymore.

source/opt/eliminate_dead_constant_pass.cpp
```diff
@@ -36,16 +36,15 @@ Pass::Status EliminateDeadConstantPass::Process(ir::IRContext* irContext) {
...
-          std::count_if(uses->begin(), uses->end(), [](const analysis::Use& u) {
-            return !(ir::IsAnnotationInst(u.inst->opcode()) ||
-                     ir::IsDebug1Inst(u.inst->opcode()) ||
-                     ir::IsDebug2Inst(u.inst->opcode()) ||
-                     ir::IsDebug3Inst(u.inst->opcode()));
-          });
-    }
+    irContext->get_def_use_mgr()->ForEachUse(
+        const_id, [&count](ir::Instruction* user, uint32_t index) {
+          (void)index;
```
atgoo: Consider
`[&count](ir::Instruction* user, uint32_t /*index*/)`

dneto0:
@alan-baker and I looked into the Windows debug build hang.  The debug version of the set insertion checks that the less functor has weak ordering. It did not, and the machine raised a Dr. Watson window and hung.

source/opt/def_use_manager.cpp
alan-baker: Good suggestion

source/opt/def_use_manager.cpp
alan-baker: Good suggestion

source/opt/def_use_manager.h
alan-baker: Will do

source/opt/eliminate_dead_constant_pass.cpp
alan-baker: Good suggestion

source/opt/def_use_manager.cpp
```diff
@@ -76,33 +76,71 @@ const ir::Instruction* DefUseManager::GetDef(uint32_t id) const {
...
+void DefUseManager::ForEachUser(uint32_t id,
+                                const std::function<void(ir::Instruction*)>& f) const {
+  ForEachUser(GetDef(id), f);
+}
+
+void DefUseManager::ForEachUse(const ir::Instruction* def,
+                               const std::function<void(ir::Instruction*, uint32_t)>& f) const {
+  // Ensure that |def| has been registered.
+  assert(def && def == GetDef(def->result_id()) && "Definition is not registered.");
+  auto iter = id_to_users_.lower_bound(UserEntry(const_cast<ir::Instruction*>(def), nullptr));
```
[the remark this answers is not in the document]
alan-baker: Good suggestion

alan-baker:
Windows debug build passed and the whole offending test suite took 10sec. I think this review is ready for more eyes. PTAL.

source/opt/def_use_manager.cpp
```diff
@@ -76,33 +78,87 @@ const ir::Instruction* DefUseManager::GetDef(uint32_t id) const {
...
+                                const ir::Instruction* inst) const {
+  return UsersNotEnd(iter, id_to_users_.end(), inst);
+}
+
+void DefUseManager::ForEachUser(const ir::Instruction* def,
+                                const std::function<void(ir::Instruction*)>& f) const {
+  // Ensure that |def| has been registered.
+  assert(def && def == GetDef(def->result_id()) && "Definition is not registered.");
+  auto end = id_to_users_.end();
+  for (auto iter = UsersBegin(def); UsersNotEnd(iter, end, def); ++iter) {
```
dnovillo: line break after ';'

source/opt/def_use_manager.h
```diff
@@ -53,13 +54,41 @@ inline bool operator<(const Use& lhs, const Use& rhs) {
...
+    // If lhs.first and rhs.first are both null, fall through to checking the
+    // second entries.
+    if (lhs.first && rhs.first) {
+      if (lhs.first->unique_id() < rhs.first->unique_id())
+        return true;
+      if (rhs.first->unique_id() < lhs.first->unique_id())
+        return false;
+    }
+
+    // Return false on equality
```
dnovillo: '.' at end of comment.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -31,18 +31,22 @@ const uint32_t kStoreValIdInIdx = 1;
...
-  assert(uses != nullptr);
-  for (auto u : *uses) {
-    SpvOp op = u.inst->opcode();
+  bool hasOnlySupportedRefs = true;
+  get_def_use_mgr()->ForEachUser(ptrId, [this,&hasOnlySupportedRefs](ir::Instruction* user) {
+    SpvOp op = user->opcode();
     if (IsNonPtrAccessChain(op) || op == SpvOpCopyObject) {
-      if (!HasOnlySupportedRefs(u.inst->result_id())) return false;
+      if (!HasOnlySupportedRefs(user->result_id())) {
+        hasOnlySupportedRefs = false;
```
dnovillo: The logic has shifted somewhat here.  This now sets the flag to false, but does not cut the search short.  Intended?  Or am I just confused?

source/opt/mem_pass.cpp
```diff
@@ -118,13 +118,14 @@ ir::Instruction* MemPass::GetPtr(ir::Instruction* ip, uint32_t* varId) {
...
-  for (auto u : *uses) {
-    const SpvOp op = u.inst->opcode();
-    if (op != SpvOpName && !IsNonTypeDecorate(op)) return false;
-  }
-  return true;
+  bool hasOnlyNamesAndDecorates = true;
+  get_def_use_mgr()->ForEachUser(id, [this, &hasOnlyNamesAndDecorates](ir::Instruction* user) {
+    SpvOp op = user->opcode();
+    if (op != SpvOpName && !IsNonTypeDecorate(op)) {
+      hasOnlyNamesAndDecorates = false;
```
dnovillo: Likewise here.

source/opt/def_use_manager.h
```diff
@@ -53,13 +54,41 @@ inline bool operator<(const Use& lhs, const Use& rhs) {
   return lhs.operand_index < rhs.operand_index;
 }
 
-using UseList = std::list<Use>;
+using UserEntry = std::pair<ir::Instruction*, ir::Instruction*>;
```
dneto0: Which is the use and which is the user?  Must say so.
I'm going to assume the .first entry is the thing being used.  The .second entry is the thing using it.
Are they allowed to be nullptr?  And what does that mean?

source/opt/def_use_manager.h
```diff
@@ -53,13 +54,41 @@ inline bool operator<(const Use& lhs, const Use& rhs) {
...
+
+struct less_than_id_ptr {
```
dneto0: Suggest renaming this to UserEntryLess

dneto0: Should say what this is trying to do.  It's a weak less-than ordering (or whatever), as required for the comparator for a std::set.

source/opt/def_use_manager.h
```diff
@@ -53,13 +54,41 @@ inline bool operator<(const Use& lhs, const Use& rhs) {
...
+  bool operator()(const UserEntry& lhs, const UserEntry& rhs) const {
+    if (!lhs.first && rhs.first)
```
dneto0: // A null use always compares less than a real use.

source/opt/def_use_manager.h
```diff
@@ -53,13 +54,41 @@ inline bool operator<(const Use& lhs, const Use& rhs) {
...
+      return true;
+    if (lhs.first && !rhs.first)
+      return false;
+
+    // If lhs.first and rhs.first are both null, fall through to checking the
+    // second entries.
+    if (lhs.first && rhs.first) {
+      if (lhs.first->unique_id() < rhs.first->unique_id())
```
dneto0: // If both are real uses, then compare IDs.

source/opt/def_use_manager.h
```diff
@@ -53,13 +54,41 @@ inline bool operator<(const Use& lhs, const Use& rhs) {
...
+using UserEntry = std::pair<ir::Instruction*, ir::Instruction*>;
+
+struct less_than_id_ptr {
+  bool operator()(const UserEntry& lhs, const UserEntry& rhs) const {
+    if (!lhs.first && rhs.first)
+      return true;
+    if (lhs.first && !rhs.first)
+      return false;
+
+    // If lhs.first and rhs.first are both null, fall through to checking the
```
dneto0: Feels like this comment belongs after the "if" construct that starts on line 68, i.e. move to  around line 74.

source/opt/def_use_manager.h
```diff
@@ -120,12 +172,23 @@ class DefUseManager {
   using InstToUsedIdsMap =
       std::unordered_map<const ir::Instruction*, std::vector<uint32_t>>;
 
+  // Returns the first location that {|def|, nullptr} could be inserted into the
+  // users map with violating ordering.
```
dneto0: "without" ?

source/opt/def_use_manager.h
```diff
@@ -120,12 +172,23 @@ class DefUseManager {
...
+  IdToUsersMap::const_iterator UsersBegin(const ir::Instruction* def) const;
+
+  // Returns true if |iter| has not reached the end of |def|'s users.
+  bool UsersNotEnd(const IdToUsersMap::const_iterator& iter,
+                   const ir::Instruction* def) const;
+  bool UsersNotEnd(const IdToUsersMap::const_iterator& iter,
+                   const IdToUsersMap::const_iterator& cached_end,
```
dneto0: And what is cached_end ?

source/opt/def_use_manager.cpp
```diff
@@ -76,33 +78,87 @@ const ir::Instruction* DefUseManager::GetDef(uint32_t id) const {
...
+void DefUseManager::ForEachUse(const ir::Instruction* def,
+                               const std::function<void(ir::Instruction*, uint32_t)>& f) const {
+  // Ensure that |def| has been registered.
+  assert(def && def == GetDef(def->result_id()) && "Definition is not registered.");
+  auto end = id_to_users_.end();
+  for (auto iter = UsersBegin(def); UsersNotEnd(iter, end, def); ++iter) {
+    ir::Instruction* user = iter->second;
+    for (uint32_t idx = 0; idx != user->NumOperands(); ++idx) {
+      const ir::Operand& op = user->GetOperand(idx);
+      switch (op.type) {
```
dneto0: Prefer    if (op.type != SPV_OPERAND_TYPE_RESULT_ID && spvOpIsIdType(op.type)) ...

Just in case I have to add another id-type enum some day.

source/opt/local_single_store_elim_pass.cpp
alan-baker: Intended. The early return doesn't work with ForEachUser().

source/opt/mem_pass.cpp
alan-baker: Same as above.

source/opt/def_use_manager.h
alan-baker: Will fix.

source/opt/def_use_manager.cpp
alan-baker: I'm confused. Where do you think the line break is missing?

source/opt/def_use_manager.cpp
```diff
@@ -156,17 +195,8 @@ bool operator==(const DefUseManager& lhs, const DefUseManager& rhs) {
...
-    if (rhs_iter == rhs.id_to_uses_.end()) {
-      return false;
-    }
-    use.second.sort();
-    UseList rhs_uselist = rhs_iter->second;
-    rhs_uselist.sort();
-    if (use.second != rhs_uselist) {
-      return false;
-    }
+  if (lhs.id_to_users_ != rhs.id_to_users_) {
```
dneto0: Not for this CL, but .... Why do we need an equality comparison on a DefUseManager ???

source/opt/strength_reduction_pass.cpp
```diff
@@ -164,6 +164,10 @@ uint32_t StrengthReductionPass::GetConstantId(uint32_t val) {
         context(), SpvOp::SpvOpConstant, uint32_type_id_, resultId, {constant}));
     get_module()->AddGlobalValue(std::move(newConstant));
 
+    // Notify the DefUseManager about this constant.
```
dneto0: Good catch

test/link/global_values_amount_test.cpp
```diff
@@ -105,9 +105,6 @@ class EntryPoints : public spvtest::LinkerTest {
   spvtest::Binaries binaries;
 };
 
-// TODO(dneto): Fix performance issue for debug builds on Windows
```
dneto0: Victory!

test/opt/def_use_test.cpp
```diff
@@ -33,6 +33,22 @@ using ::testing::UnorderedElementsAreArray;
 using namespace spvtools;
 using spvtools::opt::analysis::DefUseManager;
 
+uint32_t NumUses(const std::unique_ptr<ir::IRContext> &context, uint32_t id) {
```
dneto0: Add a method comment.

dnovillo:
On Thu, Nov 23, 2017 at 10:22 AM, alan-baker <notifications@github.com>
wrote:
>
> +  // Ensure that |def| has been registered.
> +  assert(def && def == GetDef(def->result_id()) && "Definition is not registered.");
> +  auto end = id_to_users_.end();
> +  for (auto iter = UsersBegin(def); UsersNotEnd(iter, end, def); ++iter) {
>
> I'm confused. Where do you think the line break is missing?
>

Never mind.  I completely misread that line.

test/opt/def_use_test.cpp
dneto0: LLVM has method hasNUsesOrMore which can sometimes be faster than counting all the uses, but is a quick way to tell if a thing has any uses or a unique use.    In future we might want something similar as an Instruction method.

test/opt/def_use_test.cpp
```diff
@@ -71,12 +87,36 @@ void CheckDef(const InstDefUse& expected_defs_uses,
...
+    if (def->opcode() != SpvOpConstant) {
+      // Constants don't disassemble properly without a full context.
+      EXPECT_EQ(expected_def, DisassembleInst(actual_defs.at(id)));
+    }
   }
 }
 
+using UserMap = std::unordered_map<uint32_t, std::vector<ir::Instruction*>>;
+
+UserMap BuildAllUsers(const DefUseManager* mgr, uint32_t idBound) {
```
dneto0: Add a method comment.

source/opt/local_single_store_elim_pass.cpp
dneto0: Not confused.   I've debated in my head whether the ForEachUse(r) methods should return a bool to indicate "keep going".  Then an early exit is to return false.

I don't know if that makes the common use too cumbersome.  ?

Or should we have another method with the short-circuit.  But I don't have a good name for it: ForEachUseContinue
ForEachUseAnd

???  What's a good name?

I'm ok with doing that modification in a future CL.  Or if @dnovillo wants it now I'm cool with that too.

source/opt/block_merge_pass.cpp
```diff
@@ -23,27 +23,26 @@ namespace spvtools {
...
-  const analysis::UseList* uses = get_def_use_mgr()->GetUses(labId);
   int rcnt = 0;
-  for (const auto u : *uses) {
-    // Don't count OpName
-    if (u.inst->opcode() == SpvOpName) continue;
-    if (rcnt == 1) return true;
-    ++rcnt;
-  }
-  return false;
+  get_def_use_mgr()->ForEachUse(labId,
```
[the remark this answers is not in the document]
alan-baker: Will change.

source/opt/local_single_store_elim_pass.cpp
alan-baker: I thought about supporting the early returns, but it felt awkward for most of the code I wanted to write.

I think having new methods is probably better than forcing every lambda to have "return true" at the end. In that case, we can keep the basic functionality with the same unambiguous name. I don't have any good suggestions for the new method name though.

source/opt/def_use_manager.cpp
alan-baker: This is used in the consistency checks (in debug builds) to ensure analyses passes promise to preserve are actually correct.

source/opt/def_use_manager.cpp
alan-baker: Thanks, didn't know about that helper.

source/opt/def_use_manager.h
alan-baker: I'll add comments.

source/opt/def_use_manager.h
alan-baker: Will comment.

source/opt/def_use_manager.h
alan-baker: Thanks

source/opt/def_use_manager.h
alan-baker: Will comment.

test/opt/def_use_test.cpp
alan-baker: Will do

test/opt/def_use_test.cpp
alan-baker: Will do.

source/opt/def_use_manager.cpp
```diff
@@ -50,7 +50,9 @@ void DefUseManager::AnalyzeInstUse(ir::Instruction* inst) {
       case SPV_OPERAND_TYPE_MEMORY_SEMANTICS_ID:
       case SPV_OPERAND_TYPE_SCOPE_ID: {
         uint32_t use_id = inst->GetSingleWordOperand(i);
-        id_to_uses_[use_id].push_back({inst, i});
+        ir::Instruction* def = GetDef(use_id);
```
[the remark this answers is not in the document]
alan-baker: Done

source/opt/eliminate_dead_constant_pass.cpp
```diff
@@ -96,25 +95,23 @@ Pass::Status EliminateDeadConstantPass::Process(ir::IRContext* irContext) {
...
-            ir::IsDebug3Inst(u.inst->opcode())) {
-          dead_others.insert(u.inst);
-        }
+    irContext->get_def_use_mgr()->ForEachUser(dc, [&dead_others](ir::Instruction* user) {
+      SpvOp op = user->opcode();
+      if (ir::IsAnnotationInst(op) ||
+          ir::IsDebug1Inst(op) ||
+          ir::IsDebug2Inst(op) ||
+          ir::IsDebug3Inst(op)) {
+        dead_others.insert(user);
```
s-perron: KillDef below not deletes the names and decorations, so we can probably delete everything related to ```dead_others```

source/opt/strength_reduction_pass.cpp
s-perron: We should probably call the IRContext's version of AddGlobalValue, and have that update the def-use manager.

source/opt/def_use_manager.h
```diff
@@ -54,17 +54,35 @@ inline bool operator<(const Use& lhs, const Use& rhs) {
...
+// the minor key so that all the users of a particular definition are
+// consecutive in a container.
+//
+// A null user always compares less than a real user. This is done provide easy
+// values to search for the beginning of the users of a particular definition
+// (i.e. using {def, nullptr}).
+struct UserEntryLess {
   bool operator()(const UserEntry& lhs, const UserEntry& rhs) const {
+    // If lhs.first and rhs.first are both null, fall through to checking the
+    // second entries.
```
dneto0: If the def must never be null, then some of this code is purely defensive.... which is fine.

source/opt/def_use_manager.h
```diff
@@ -80,6 +98,7 @@ struct less_than_id_ptr {
     if (!rhs.second)
       return false;
 
+    // If neither user is null then compare unqiue ids.
```
dneto0: "unique"

source/opt/def_use_manager.h
```diff
@@ -54,17 +54,35 @@ inline bool operator<(const Use& lhs, const Use& rhs) {
...
 using UserEntry = std::pair<ir::Instruction*, ir::Instruction*>;
 
-struct less_than_id_ptr {
+// Orders UserEntry for use in associative containers (i.e. less than ordering).
+//
+// The definition of an UserEntry is treated as the major key and the users as
+// the minor key so that all the users of a particular definition are
+// consecutive in a container.
+//
+// A null user always compares less than a real user. This is done provide easy
```
dneto0: "done to provide"

source/opt/strength_reduction_pass.cpp
dneto0: I filed #1000 for that.

source/opt/eliminate_dead_constant_pass.cpp
dneto0: So true.  Good catch.  (Removing "not" from your sentence first..)

Or even invoke KillInst on each member of dead_consts rather than KillDef.

Agree that is good for a future cleanup.

dneto0:
Hi @alan-baker please rebase.  It's not a particularly pleasant to merge manually.

dneto0:
And I just pushed #999 into master, so that's another reason to rebase.

alan-baker:
Rebased and pushed with those small comment fixes.

dneto0:
Rebased, squashed, and pushed into master as 746bfd210a

PR #992, opened 2017-11-22
WIP: Add derivatives validation pass

atgoo:
Checks operands of instructions in opcode range from OpDPdx to
OpFwidthCoarse.

### DIFF
--- a/source/validate_derivatives.cpp
+++ b/source/validate_derivatives.cpp
@@ -39,8 +39,7 @@ spv_result_t DerivativesPass(ValidationState_t& _,
     case SpvOpDPdxCoarse:
     case SpvOpDPdyCoarse:
     case SpvOpFwidthCoarse: {
-      if (!_.IsFloatScalarType(result_type) &&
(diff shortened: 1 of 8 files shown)


source/validate_derivatives.cpp
```diff
@@ -0,0 +1,66 @@
...
+               << spvOpcodeString(opcode);
+      }
+
+      const uint32_t p_type = _.GetOperandTypeId(inst, 2);
+      if (p_type != result_type) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+               << "Expected P type and Result Type to be the same: "
+               << spvOpcodeString(opcode);
+      }
+
```
dneto0: Deserves a comment here saying that the only other condition is that these instructions are only usable by the Fragment execution model. But that's must be checked elsewhere (after the entire module is read) since this instruction might be in a function that isn't itself an entry point.

dneto0:
Rebased, squashed, and pushed into master as ab892f7bd6

PR #993, opened 2017-11-22
Move SetContextMessageConsumer into libspirv namespace

dneto0:
Avoid polluting the global namespace.


dneto0:
Rebased, squashed, and pushed into master as c2999273

PR #995, opened 2017-11-22
Test for pollution of the global namespace

dneto0:
Works on Linux only for now.  That's a good start.

This is WIP because we need to fix these two problems it finds:

check_symbol_exports: error: Unescaped exported symbol: SetContextMessageConsumer(spv_context_t*, std::function<void (spv_message_level_t, char const*, spv_position_t const&, char const*)>)
check_symbol_exports: error: Unescaped exported symbol: ValidateBinaryUsingContextAndValidationState(spv_context_t const&, unsigned int const*, unsigned long, spv_diagnostic_t**, libspirv::ValidationState_t*)

#993 fixes the first one.  I have yet to fix the second one.

### DIFF
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -229,14 +229,6 @@ UNUSED(void PrintDotGraph(ValidationState_t& _, libspirv::Function func)) {
     printf("}\n");
   }
 }
-}  // anonymous namespace
-
-spv_result_t spvValidate(const spv_const_context context,
-                         const spv_const_binary binary,
-                         spv_diagnostic* pDiagnostic) {
-  return spvValidateBinary(context, binary->code, binary->wordCount,
-                           pDiagnostic);
-}
 
 spv_result_t ValidateBinaryUsingContextAndValidationState(
     const spv_context_t& context, const uint32_t* words, const size_t num_words,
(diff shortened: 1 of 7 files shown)


dneto0:
I've now fixed all the bad exports this test caught.

dneto0:
Please take a look.

source/validate.cpp
```diff
@@ -229,14 +229,6 @@ UNUSED(void PrintDotGraph(ValidationState_t& _, libspirv::Function func)) {
...
-}  // anonymous namespace
-
-spv_result_t spvValidate(const spv_const_context context,
-                         const spv_const_binary binary,
-                         spv_diagnostic* pDiagnostic) {
-  return spvValidateBinary(context, binary->code, binary->wordCount,
-                           pDiagnostic);
-}
 
 spv_result_t ValidateBinaryUsingContextAndValidationState(
```
dneto0: This function use to be in the global namespace.  The delta moves it within scope of the anonymous namespace.

source/validate.cpp
```diff
@@ -341,6 +333,14 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
...
                         context.ext_inst_table, *vstate, &position);
 }
+}  // anonymous namespace
+
+spv_result_t spvValidate(const spv_const_context context,
+                         const spv_const_binary binary,
+                         spv_diagnostic* pDiagnostic) {
+  return spvValidateBinary(context, binary->code, binary->wordCount,
+                           pDiagnostic);
+}
```
dnovillo: Sorry if I'm missing something obvious.  But it looks like the function is still out of the anonymous namespace.

dneto0: I wasn't specific.  :-)

ValidateBinaryUsingContextAndValidationState was outside the anonymous namespace.  I moved spvValidate down and extended the anonymous namespace region down to enclose ValidateBinaryUsingContextAndValidationState.

dneto0: spvValidate is part of our public extern "C" API, and should remain there.

dnovillo: Ah, ok.  I figured I was missing something :)

Thanks.

dneto0:
Rebased and pushed into master as d9129f00a

PR #996, opened 2017-11-23
Add validation checks for Execution Model

atgoo:
Currently checks that these instructions are called from entry points
with Fragment execution model.
OpImageImplicit*
OpImageQueryLod
OpKill

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -18,6 +18,7 @@
 
 #include <algorithm>
 #include <iostream>
+#include <stack>
 #include <unordered_set>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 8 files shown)


atgoo:
Passes Vulkan CTS master.

source/validate_id.cpp
```diff
@@ -300,6 +302,31 @@ bool idUsage::isValid<SpvOpEntryPoint>(const spv_instruction_t* inst,
       return false;
     }
   }
+
+  std::stack<uint32_t> call_stack;
+  call_stack.push(entryPoint->id());
+  while (!call_stack.empty()) {
```
dneto0: Recommend using a visited set to avoid cycling if there are cycles in the call graph.

That's possible to express in SPIR-V, but banned by all the known client APIs.  But the validator should not loop forever on such a bad input.

dneto0:
Build failure due to shadowing violation
https://travis-ci.org/KhronosGroup/SPIRV-Tools/jobs/306430365#L590

dneto0:
Rebased, squashed, and pushed into master as 3e08a3f7

Issue #994, opened 2017-11-22
Derivative instructions are not combinators


dneto0:
Remove the derivative instructions from the optimizer's list of combinators.

dneto0:
Blocked until #971 is merged

dneto0:
In a derivative instruction, an invocation exchanges information with its peers in the quad being executed.  So it has side effects beyond calculating a value.

dneto0:
Cc: @greg-lunarg  :  This is a functional change to the aggressive dead code elim, but I think a necessary fix.

PR #997, opened 2017-11-23
Remove derivative instructions from the list of combinators.

s-perron:
These instructions compute their value based the value of the immediate
neighbours of the current fragment.  This means the result is not
defined purely by the operands of the instruction.

Fixes issue #994.


dneto0:
Rebased and pushed into master as 0b1cb27f83b

PR #999, opened 2017-11-23
Allow derived access chain without uses in access chain conversion

antiagainst:


antiagainst:
Would appreciate @greg-lunarg 's review too

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -139,7 +139,17 @@ bool LocalAccessChainConvertPass::IsConstantIndexAccessChain(
...
+  if (!uses) {
+    // This is a variable (or access chain to a variable) that has no uses.
+    // We won't encounter loads or stores for this <result-id> per se, but
+    // this <result-id> may be derived from some other variable (or access
+    // chain). Return true here can unblock the access chain conversion of
+    // the root variable. This particular <result-id> won't be touched and
+    // can be handled in dead code elimination.
+    return true;
+  }
+
```
dnovillo: This should not be needed after @alan-baker PR #987 is in.  @alan-baker could you confirm?

antiagainst:
Oh, forgot that this function is duplicated at two other places... Will add them too.

source/opt/local_access_chain_convert_pass.cpp
dneto0: Agree.  #987 changes the API to provide a ForEachUse or ForEachUser and you can't get nullptr at all.

antiagainst: Oh, great! That would be better than this ad-hoc fix.

dneto0:
Rebased, squashed, and pushed into master as b02c9a58

antiagainst:
Thanks, @dneto0! :)

Issue #1004, opened 2017-11-24
use-after-free in Instruction::unique_id()


dneto0:
Running unit tests with address sanitizer.

Config line is 
cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS=-fsanitize=address -DCMAKE_C_FLAGS=-fsanitize=address -DCMAKE_C_COMPILER=clang-3.6 -DCMAKE_CXX_COMPILER=clang++-3.6 .. -DRE2_BUILD_TESTING=OFF

Test output is:

83/85 Test #18: spirv-tools-test_link_global_values_amount ...........***Failed    7.72 sec
Running main() from gmock_main.cc
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from EntryPoints
[ RUN      ] EntryPoints.UnderLimit
=================================================================
==22040==ERROR: AddressSanitizer: heap-use-after-free on address 0x60b000b03e54 at pc 0x0000006cfc74 bp 0x7ffc049f6910 sp 0x7ffc049f6908
READ of size 4 at 0x60b000b03e54 thread T0
    #0 0x6cfc73 in spvtools::ir::Instruction::unique_id() const /build/dneto/SPIRV-Tools/build/../source/opt/instruction.h:142:5
    #1 0xa64886 in spvtools::opt::analysis::UserEntryLess::operator()(std::pair<spvtools::ir::Instruction*, spvtools::ir::Instruction*> const&, std::pair<spvtools::ir::Instruction*, spvtools::ir::Instruction*> const&) const /build/dneto/SPIRV-Tools/source/opt/def_use_manager.h:102:38
[...]

PR #1006, opened 2017-11-24
Notify the context of instructions that are being erased

alan-baker:
Fixes use-after-free error in RemoveDuplicatesPass (#1004)

Removed instructions were being left in the DefUseManager despite having been erased.


dneto0:
Rebased and pushed into master as 0cae89e79e

PR #1010, opened 2017-11-24
Relaxed OpImageRead validation rules

atgoo:
Removed the check that result type of OpImageRead should be a vector4.
Will reenable/adapt once the spec is clarified on what the right
dimension should be.

### DIFF
--- a/test/val/val_image_test.cpp
+++ b/test/val/val_image_test.cpp
@@ -2262,6 +2262,8 @@ TEST_F(ValidateImage, ReadNeedCapabilityImageCubeArray) {
       "ImageRead"));
 }
 
+#if 0
+// TODO(atgoo@github.com) Disabled until the spec is clarified.
 TEST_F(ValidateImage, ReadWrongResultType) {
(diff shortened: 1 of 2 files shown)


test/val/val_image_test.cpp
```diff
@@ -2262,6 +2262,8 @@ TEST_F(ValidateImage, ReadNeedCapabilityImageCubeArray) {
       "ImageRead"));
 }
 
+#if 0
+// TODO(atgoo@github.com) Disabled until the spec is clarified.
 TEST_F(ValidateImage, ReadWrongResultType) {
```
dneto0: If you have a broken test that you cannot fix right away, you can add the DISABLED_ prefix to its name. 

https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#temporarily-disabling-tests

But this is fine for now.

dneto0:
Rebased.  I changed the #if 0 to DISABLED_ prefix on the two tests to disable.
Then pushed into master as f84f266977

Issue #1001, opened 2017-11-23
Fragment execution model required for derivative instructions


dneto0:
#996 covers cases: implicit Lod sampling, query Lod, OpKill

dneto0:
Done.

PR #1013, opened 2017-11-24
Derivative opcodes require Fragment exec model

atgoo:
Added validator check that all derivative opcodes require Fragment
execution model.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1001


atgoo:
Passes Vulkan CTS master.

dneto0:
Rebased and pushed into master as d8b2013ec

PR #1014, opened 2017-11-24
Relaxed OpImageWrite texel type check

atgoo:
Relax texel type check in OpImageWrite until the spec is clarified.


dneto0:
Rebased, updated CHANGES, and pushed into master as c170afd93

Issue #991, opened 2017-11-22
spirv-val invalid on glslangvalidator generated spv file for opGroupXNonUniformAMD


aurolou:
When generating the spv file with glslangvalidator from the glsl file and then validating the generated spv file with spirv-val, "Operand 4 of GroupXNonUniformAMD requires one of these capabilities: Kernel", is returned for the add/min/maxInvocationsNonUniformAMD functions.
[shader.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/1496606/shader.zip)

dneto0:
The SPV_AMD_shader_ballot extension clearly intends to use group operations because the new instructions are useless without group operations.  The extension spec should say to add the extension as an additional enabling condition for the group operation.

GL_AMD_shader_ballot says that Reduce, InclusiveScan, ExclusiveScan are usable. 
I'll fix the validator to match the intent.

dneto0:
Fixed with #1015

PR #1015, opened 2017-11-24
Fix validation of group ops in SPV_AMD_shader_ballot

dneto0:
This needs custom code since the rules from the extension
are not encoded in the grammar.

Changes are:
- The new group instructions don't require Group capability
  when the extension is declared.
- The Reduce, InclusiveScan, ExclusiveScan normally require the Kernel
  capability, but don't when the extension is declared.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/991

### DIFF
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -67,6 +67,39 @@ spv_result_t CapabilityError(ValidationState_t& _, int which_operand,
          << " requires one of these capabilities: " << required_capabilities;
 }
 
+// Returns capabilities that enable an opcode.  An empty result is interpreted
+// as no prohibition of use of the opcode.  If the result is non-empty, then
+// the opcode may only be used if at least one of the capabilities is specified
+// by the module.
+CapabilitySet EnablingCapabilitiesForOp(const ValidationState_t& state,
+                                        SpvOp opcode) {
+  // Exceptions for SPV_AMD_shader_ballot
+  switch (opcode) {
+    // Normally these would require Group capability
+    case SpvOpGroupIAddNonUniformAMD:
+    case SpvOpGroupFAddNonUniformAMD:
+    case SpvOpGroupFMinNonUniformAMD:
+    case SpvOpGroupUMinNonUniformAMD:
+    case SpvOpGroupSMinNonUniformAMD:
+    case SpvOpGroupFMaxNonUniformAMD:
+    case SpvOpGroupUMaxNonUniformAMD:
+    case SpvOpGroupSMaxNonUniformAMD:
+      if (state.HasExtension(libspirv::kSPV_AMD_shader_ballot))
+        return CapabilitySet();
+      break;
+    default:
+      break;
+  }
+  // Look it up in the grammar
+  spv_opcode_desc opcode_desc = {};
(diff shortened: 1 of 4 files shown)


source/validate_instruction.cpp
```diff
@@ -67,6 +67,36 @@ spv_result_t CapabilityError(ValidationState_t& _, int which_operand,
          << " requires one of these capabilities: " << required_capabilities;
 }
 
+// Returns an opcode's required capabilities.
```
atgoo: It's not required, but rather enabling. Required means that all of them are required. But in reality any of the capabilities in the list can enable the opcode.

test/val/val_extensions_test.cpp
```diff
@@ -118,4 +118,70 @@ TEST_F(ValidateExtensionCapabilities, DeclCapabilityFailure) {
...
+
+  const string assembly = preamble + GetParam() + "\nOpReturn OpFunctionEnd";
+
+  CompileSuccessfully(assembly.c_str());
+  ASSERT_EQ(SPV_SUCCESS, ValidateInstructions()) << getDiagnosticString();
+}
+
+INSTANTIATE_TEST_CASE_P(ExpectSuccess, ValidateAMDShaderBallotCapabilities,
+                        Values(
+                            // clang-format off
```
atgoo: The indentation is strange. Consider putting this before the entire statement.

test/val/val_extensions_test.cpp
```diff
@@ -118,4 +118,70 @@ TEST_F(ValidateExtensionCapabilities, DeclCapabilityFailure) {
...
+  "%umax_reduce = OpGroupUMaxNonUniformAMD %uint %scope Reduce %uint_const",
+  "%umax_iscan = OpGroupUMaxNonUniformAMD %uint %scope InclusiveScan %uint_const",
+  "%umax_escan = OpGroupUMaxNonUniformAMD %uint %scope ExclusiveScan %uint_const",
+
+  "%smax_reduce = OpGroupUMaxNonUniformAMD %int %scope Reduce %int_const",
+  "%smax_iscan = OpGroupUMaxNonUniformAMD %int %scope InclusiveScan %int_const",
+  "%smax_escan = OpGroupUMaxNonUniformAMD %int %scope ExclusiveScan %int_const"
+                            // clang-format on
+                            ));
+
```
atgoo: Consider testing that it fails without the extension.

source/validate_instruction.cpp
dneto0: Fixed

test/val/val_extensions_test.cpp
dneto0: Fixed as a side effect of adding validation failure cases.

test/val/val_extensions_test.cpp
dneto0: Fixed.

dneto0:
Please take another look.

dneto0:
Rebased and squashed.   The previous Windows bot failure was caused by someone else's commit.
There is a current windows bot failure due to #1034

dneto0:
Rebased and pushed into master as 3c2e4c7d9

PR #1016, opened 2017-11-24
Validate OpTypeImage and OpTypeSampledImage

atgoo:
Added new validation rules to the validate image pass.

### DIFF
--- a/source/validate_image.cpp
+++ b/source/validate_image.cpp
@@ -600,6 +600,80 @@ spv_result_t ImagePass(ValidationState_t& _,
   }
 
   switch (opcode) {
+    case SpvOpTypeImage: {
+      assert(result_type == 0);
+
+      ImageTypeInfo info;
+      if (!GetImageTypeInfo(_, inst->words[1], &info)) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "OpTypeImage: corrupt definition";
+      }
+
+      const SpvOp sampled_type_opcode = _.GetIdOpcode(info.sampled_type);
+      if (sampled_type_opcode != SpvOpTypeVoid &&
+          sampled_type_opcode != SpvOpTypeInt &&
+          sampled_type_opcode != SpvOpTypeFloat) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Expected Sampled Type to be either void or numerical scalar "
+            << "type: " << spvOpcodeString(opcode);
+      }
+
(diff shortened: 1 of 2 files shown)


atgoo:
Passes Vulkan CTS master.

source/validate_image.cpp
```diff
@@ -594,6 +594,80 @@ spv_result_t ImagePass(ValidationState_t& _,
...
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Expected Sampled Type to be either void or numerical scalar "
+            << "type: " << spvOpcodeString(opcode);
+      }
+
+      // Dim is checked elsewhere.
+
+      if (info.depth > 2) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+            << "Invalid Depth " << info.depth << " (must be 0, 1 or 2): "
```
antiagainst: Nit: Hmm, I would typically expect the error message to be something like `invalid Depth parameter for OpTypeImage: <the-wrong-number>`, or `OpTypeImage: invalid Depth parameter: <the-wrong-number>`. The current error message reads like we are providing `TypeImage` as the `Depth` parameter.

test/val/val_image_test.cpp
```diff
@@ -310,6 +305,120 @@ OpFunctionEnd)";
   return ss.str();
 }
 
+std::string GetShaderHeader(
+    const std::string& capabilities_and_extensions = "") {
+  std::stringstream ss;
```
antiagainst: `ostringstream`?

source/validate_image.cpp
```diff
@@ -980,6 +1055,13 @@ spv_result_t ImagePass(ValidationState_t& _,
             << "Corrupt image type definition";
       }
 
+      if (info.dim == SpvDimSubpassData) {
```
dneto0: I reviewed other places which should check for SubpassData.   
An image with Dim=SubpassData is required to have Sampled = 2 (i.e. unsampled), so that already excludes the majority of image operations.   OpImageWrite already bans an image with Dim = SubpassData.  

The only interesting case I found is OpImageSparseRead;  I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/1022 to ban the use of images with Dim=SubpassData.  That's not explicitly spec'd anywhere but I'm pretty sure the (Vulkan) use case for SubpassData is different from sparse memory.

dneto0:
Rebased, squashed, and pushed into master as  e1ceff9f

PR #1018, opened 2017-11-27
Re-format source tree - NFC.

dnovillo:
Re-formatted the source tree with the command:

$ /usr/bin/clang-format -style=file -i \
    $(find include source tools test utils -name '*.cpp' -or -name '*.h')

This required a fix to source/val/decoration.h.  It was not including
spirv.h, which broke builds when the #include headers were re-ordered by
clang-format.

### DIFF
--- a/source/val/decoration.h
+++ b/source/val/decoration.h
@@ -15,6 +15,7 @@
 #ifndef LIBSPIRV_VAL_DECORATION_H_
 #define LIBSPIRV_VAL_DECORATION_H_
 
+#include <cstdint>
 #include <unordered_map>
 #include <vector>
 


test/huffman_codec.cpp
```diff
@@ -88,7 +76,8 @@ TEST(Huffman, PrintTree) {
               6------3------i
                      3------1------x
                             2------h
-)").substr(1);
+)")
+                                   .substr(1);
```
atgoo: This doesn't look right.

test/link/binary_version_test.cpp
```diff
@@ -22,27 +22,20 @@ using BinaryVersion = spvtest::LinkerTest;
 TEST_F(BinaryVersion, LinkerChoosesMaxSpirvVersion) {
   spvtest::Binaries binaries = {
       {
-          SpvMagicNumber,
```
atgoo: This seems to be intentional format. Consider turning auto-format off.

test/move_to_front_test.cpp
```diff
@@ -128,13 +123,15 @@ TEST(MoveToFront, InsertLeftRotation) {
 
   CheckTree(mtf, std::string(R"(
 30H2S2----20H1S1----D2
-)").substr(1));
+)")
+                     .substr(1));
```
atgoo: This doesn't look right.

test/link/binary_version_test.cpp
dnovillo: Done, though I would rather let clang-format do its thing and not have many exceptions.

test/move_to_front_test.cpp
dnovillo: This one is borderline for me.  There's too many lines that would need to be disabled and it's easier to just let clang-format decide here.

s-perron: I agree it does not look good, but it is in an part of the code that people are going to read very often.  Since it happens in a few different file in a lot of places, I don't think it is worth fixing until some needs to actually change those files.  I suggest just letting clang-format do its thing.

dnovillo:
On Mon, Nov 27, 2017 at 2:08 PM, Steven Perron <notifications@github.com>
wrote:

> *@s-perron* approved this pull request.
>
> I'm assuming clang-format is being run correctly. Everything LGTM.
>

Yeah.  It's executed from the top directory with -style=file.  This picks
up the .clang-format file at the top of the tree.

dnovillo:
Re-based, pushed to master as commit 491b112fd235f28fac9734bdbdaea3cd0cb69b13

PR #1019, opened 2017-11-27
Convert the CFG to an on-demand analysis - NFC.

dnovillo:
This fixes some TODOs by moving the CFG into the IRContext as an
analysis.


dnovillo:
Re-based, re-tested and pushed as 9f20799fb4129f4b1174fb168fb4508f46f64c5b

PR #1020, opened 2017-11-27
Dominator Tree Analysis (#3)

davors72:
Support for dominator and post dominator analysis on ir::Functions. This patch contains a DominatorTree class which builds the tree and DominatorAnalysis and DominatorAnalysisPass classes for interfacing with and caching the built trees.


dnovillo:
Thanks!  I'm starting the review today.

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,395 @@
...
+    }
+  } else {
+    // Technically, this is not needed, but it unifies
+    // the handling of dominator and postdom tree later on.
+    Successors[dummyStartNode].push_back(F.entry().get());
+    Predecessors[F.entry().get()].push_back(dummyStartNode);
+    for (BasicBlock& BB : F) {
+      BasicBlockListTy& SuccList = Successors[&BB];
+
+      BB.ForEachSuccessorLabel([&](const uint32_t successorID) {
```
s-perron: This does not take into account "OpSelectionMerge" instructions. It is possible that the merge point is not technically reachable, but may still need to be in the dominator tree. I think this depends on how we are going to use dominator tree. We should have a discussion on how to handle this. I'll try to come up with some odd cases and see what we should be doing about them.

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,395 @@
...
+  // postorder list.
+  std::vector<const ir::BasicBlock*> postorder;
+  auto postorder_function = [&](const ir::BasicBlock* b) {
+    postorder.push_back(b);
+  };
+
+  // CFA::CalculateDominators requires std::vector<ir::BasicBlock*>
+  // BB are derived from F, so we need to const cast it at some point
+  // no modification is made on F.
+  BasicBlockSuccessorHelper<ir::BasicBlock> helper{
```
dnovillo: Please also run clang-format over the code.  This seems oddly formatted.

source/opt/dominator_tree.h
```diff
@@ -0,0 +1,146 @@
...
+  // Returns the immediate dominator of basicblock A.
+  ir::BasicBlock* ImmediateDominator(const ir::BasicBlock* A) const;
+
+  // Returns the immediate dominator of basicblock A.
+  ir::BasicBlock* ImmediateDominator(uint32_t A) const;
+
+  // Returns true if BasicBlock A is reachable by this tree. A node would be
+  // unreachable if it cannot be reached by traversal from the start node or for
+  // a postdominator tree, cannot be reached from the exit nodes.
+  bool Reachable(const ir::BasicBlock* A) const;
```
dnovillo: Cool.  I think we can use this in CFGCleanup instead of the ad-hoc reachability analysis it does now (#1026)

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,395 @@
...
+}
+
+// Small type trait to get the function class type.
+template <typename BBType>
+struct GetFunctionClass {
+  using FunctionType = ir::Function;
+};
+
+// This helper class is basically a massive workaround for the current way that
+// depth first is implemented.
```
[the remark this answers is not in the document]
davors72: The current DFT function takes in an std::function to return the predecessors and one to return the successors of any given BB. They each return a 'const std::vector<BB*>*' which means (as it's a pointer) we have to precompute the two lists and maintain them. For our case it could just be done on the fly since they aren't used again.

Workaround is probably not the right word, I'll update the comment to make it clearer why we need that class.

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,395 @@
...
+template <typename BBType>
+class BasicBlockSuccessorHelper {
+  // This should eventually become const ir::BasicBlock.
```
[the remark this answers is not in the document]
Naghasan: Because of the interface of CFA, forcing a const will required to do large update of the library.

source/opt/dominator_tree.cpp
Naghasan: I did, I just ran it again to check, no changes

source/opt/dominator_analysis_pass.h
```diff
@@ -0,0 +1,157 @@
...
+
+  // Returns true if A is reachable from the entry.
+  inline bool IsReachable(uint32_t node_id) const {
+    return tree_.ReachableFromRoots(node_id);
+  }
+
+  // Dump the tree structure into the given stream in the dot format.
+  inline void DumpAsDot(std::ostream& out) const { tree_.DumpTreeAsDot(out); }
+
+  // Returns true if this is a postdomiator tree.
```
dnovillo: s/postdomiator/postdominator/

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,383 @@
...
+          typename PostLambda>
+static void DepthFirstSearch(const BBType* bb, SuccessorLambda successors,
+                             PreLambda pre, PostLambda post) {
+  // Ignore backedge operation.
+  auto nop_backedge = [](const BBType*, const BBType*) {};
+  CFA<BBType>::DepthFirstTraversal(bb, successors, pre, post, nop_backedge);
+}
+
+// Wrapper around CFA::DepthFirstTraversal to provide an interface to perform
+// depth first search on generic BasicBlock types. This overload is for only
```
dnovillo: "This overload is for only" -> "This overload is only for"?  Or perhaps rephrase if I totally missed the intended meaning.

source/opt/dominator_tree.cpp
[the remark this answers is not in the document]
dnovillo: Ah, thanks.  Now I understand the issue better.  I'm fine with this.

source/opt/dominator_tree.cpp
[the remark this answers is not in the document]
dnovillo: SGTM

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,383 @@
...
+  BasicBlockSuccessorHelper(Function& func, BasicBlock* dummy_start_node,
+                            bool post);
+
+  // CFA::CalculateDominators requires std::vector<BasicBlock*>.
+  using GetBlocksFunction =
+      std::function<const std::vector<BasicBlock*>*(const BasicBlock*)>;
+
+  // Returns the list of predecessor functions.
+  GetBlocksFunction GetPredFunctor() {
+    return [&](const BasicBlock* bb) {
```
dnovillo: No implicit capture, please.

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,383 @@
...
+  BasicBlockMapTy predecessors_;
+
+  // Build a bi-directional graph from the CFG of F.
+  // If invert_graph_ is true, all edge are reverted (successors becomes
+  // predecessors and vise versa).
+  // For convenience, the start of the graph is dummyStartNode. The dominator
+  // tree construction requires a unique entry node, which cannot be guarantied
+  // for the postdominator graph. The dummyStartNode BB is here to gather all
+  // entry nodes.
+  void CreateSuccessorMap(Function& f, BasicBlock* dummy_start_node);
```
dnovillo: dummy_start_node should not be needed.  You can use CFG::pseudo_entry_node()

source/opt/dominator_tree.cpp
dnovillo: Thanks.  It's the initializer list that tricked me.

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,383 @@
...
+  // Skip over empty functions.
+  if (f->cbegin() == f->cend()) {
+    return;
+  }
+
+  std::unique_ptr<ir::Instruction> dummy_label{new ir::Instruction(
+      f->GetParent()->context(), SpvOp::SpvOpLabel, 0, -1, {})};
+  // Create a dummy start node which will point to all of the roots of the tree
+  // to allow us to work with a singular root.
+  ir::BasicBlock dummy_start_node(std::move(dummy_label));
```
dnovillo: Should not be needed.  Please use CFG::pseudo_entry_block()

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+  auto getSucc = [&](const DominatorTreeNode* node) {
+    return &node->childrens_;
+  };
+
+  for (auto root : roots_) DepthFirstSearch(root, getSucc, preFunc, postFunc);
+}
+
+void DominatorTree::DumpTreeAsDot(std::ostream& out_stream) const {
+  out_stream << "digraph {\n";
```
dnovillo: Nice!

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,383 @@
...
+        out_stream << "Dummy -> " << node->bb_->id() << " [style=dotted];\n";
+      }
+    });
+  }
+  out_stream << "}\n";
+}
+
+void DominatorTree::Visit(
+    const DominatorTreeNode* node,
+    std::function<void(const DominatorTreeNode*)> func) const {
```
dnovillo: Could we make this functor return 'bool' so that it can stop traversal?

source/opt/dominator_tree.h
```diff
@@ -0,0 +1,180 @@
...
+
+  // List of DominatorTreeNode to define the list of roots
+  using DominatorTreeNodeList = std::vector<DominatorTreeNode*>;
+  using roots_iterator = DominatorTreeNodeList::iterator;
+  using roots_const_iterator = DominatorTreeNodeList::const_iterator;
+
+  DominatorTree() : postdominator_(false) {}
+  explicit DominatorTree(bool post) : postdominator_(post) {}
+
+  iterator begin() { return nodes_.begin(); }
```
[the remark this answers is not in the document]
dnovillo: @s-perron, I believe the intent is to use DominatorTree::Visit() for dom-tree traversals in dom-order.

source/opt/dominator_tree.h
```diff
@@ -0,0 +1,180 @@
...
+    roots_.clear();
+  }
+
+ private:
+  // Adds the BasicBlock to the tree structure if it doesn't already exist.
+  DominatorTreeNode* GetOrInsertNode(ir::BasicBlock* bb);
+
+  // Applies the std::function 'func' to 'node' then applies it to node's
+  // children.
+  void Visit(const DominatorTreeNode* node,
```
dnovillo: Make it public, please.  This is going to be widely used by optimizers.

source/opt/basic_block.h
```diff
@@ -98,6 +100,14 @@ class BasicBlock {
     return --insts_.cend();
   }
 
+  // return if the basic block has at least one successor
+  inline bool hasSuccessor() const {
```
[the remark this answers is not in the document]
alan-baker: This isn't true. OpUnreachable and OpKill can also terminate a block without successors. 'return ctail()->IsBranch();' is the helper you want.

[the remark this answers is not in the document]
Naghasan: Yes, that's how we fixed it

source/opt/dominator_analysis_pass.h
```diff
@@ -0,0 +1,157 @@
...
+    return StrictlyDominates(a->id(), b->id());
+  }
+
+  // Returns true if BasicBlock A strictly dominates BasicBlock B. Same as above
+  // only using the BasicBlock IDs.
+  inline bool StrictlyDominates(uint32_t a, uint32_t b) const {
+    return tree_.StrictlyDominates(a, b);
+  }
+
+  // Return the immediate dominator of node A or return NULL if it is an entry
```
alan-baker: Minor, but here and elsewhere in this file your comments referring to parameters do not match parameters names. Preferably refer to parameters as |param|. "Returns the immediate dominator of |node| or NULL if it is an entry node."

Also, here uses "Returns".

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,383 @@
...
+    };
+  }
+
+ private:
+  bool invert_graph_;
+  BasicBlockMapTy successors_;
+  BasicBlockMapTy predecessors_;
+
+  // Build a bi-directional graph from the CFG of F.
+  // If invert_graph_ is true, all edge are reverted (successors becomes
```
alan-baker: Is it really bi-directional? If edges get reversed in one case it is just a DAG.

reverted -> reversed.

source/opt/dominator_tree.cpp
```diff
@@ -0,0 +1,383 @@
...
+  // predecessors and vise versa).
+  // For convenience, the start of the graph is dummyStartNode. The dominator
+  // tree construction requires a unique entry node, which cannot be guarantied
```
alan-baker: guarantied -> guaranteed

source/opt/dominator_tree.h
```diff
@@ -0,0 +1,180 @@
...
+  // Wrapper function which gets the list of BasicBlock->DominatingBasicBlock
+  // from the CFA and stores it in the edges parameter.
+  //
+  // The |edges| vector will contain the dominator tree as pairs of nodes.
+  // The first node in the pair is a node in the graph. The second node in the
+  // pair is its immediate dominator.
+  // The root of the tree has him self as immediate dominator.
+  void GetDominatorEdges(
+      const ir::Function* f, ir::BasicBlock* dummy_start_node,
+      std::vector<std::pair<ir::BasicBlock*, ir::BasicBlock*>>& edges);
```
alan-baker: Style guide: Only const reference arguments. Please change edges to a pointer to vector.

source/opt/dominator_tree.h
```diff
@@ -0,0 +1,180 @@
...
+        parent_(nullptr),
+        childrens_({}),
+        dfs_num_pre_(-1),
+        dfs_num_post_(-1) {}
+
+  inline uint32_t id() const { return bb_->id(); }
+
+  ir::BasicBlock* bb_;
+  DominatorTreeNode* parent_;
+  std::vector<DominatorTreeNode*> childrens_;
```
alan-baker: childrens_ -> children_

source/opt/dominator_tree.h
```diff
@@ -0,0 +1,180 @@
...
+  const_iterator cend() const { return nodes_.end(); }
+
+  roots_iterator roots_begin() { return roots_.begin(); }
+  roots_iterator roots_end() { return roots_.end(); }
+  roots_const_iterator roots_begin() const { return roots_cbegin(); }
+  roots_const_iterator roots_end() const { return roots_cend(); }
+  roots_const_iterator roots_cbegin() const { return roots_.begin(); }
+  roots_const_iterator roots_cend() const { return roots_.end(); }
+
+  // Get the unique root of the tree.
```
alan-baker: It seems like some of this would be simpler using both a pseudo entry (which is used) and a pseudo exit block. Then there would be a single root for both cases and that root would never be a true basic block. Opinions?

davors72: That's actually how we implemented it at first but we regard the pseudo entry node an implementation detail of the algorithm and as such avoided exposing it as much as possible. It also means that every tree in a module would share the same root because the pseudo entry/exit blocks are retrieved from the cfg on a module scope which could lead to some confusion.

source/opt/dominator_tree.cpp
davors72: The bi-directional graph refers to the successor and predecessor maps in that helper class. They'll always maintain a link forward and back

source/opt/dominator_analysis_pass.h
```diff
@@ -0,0 +1,157 @@
...
+  explicit DominatorAnalysisBase(bool isPostDom) : tree_(isPostDom) {}
+
+  // Calculate the dominator (or postdominator) tree for given function F.
+  inline void InitializeTree(const ir::Function* f) { tree_.InitializeTree(f); }
+
+  // Returns true if BasicBlock A dominates BasicBlock B.
+  inline bool Dominates(const ir::BasicBlock* a,
+                        const ir::BasicBlock* b) const {
+    if (!a || !b) return false;
+    return Dominates(a->id(), b->id());
```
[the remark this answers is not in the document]
davors72: We treat the dummy node as an implementation detail so I'm not sure that it should. Conceptually yes it is at the top of the tree but the immediate dominator of a root node (say b) will be nullptr but then you could do Dominates(a, b) with a as pseudo_entry_block and get true which doesn't make much sense. Also, every function in a module will share the same pseudo_entry_block as it's retrieved from the CFG at module level.

source/opt/dominator_tree.h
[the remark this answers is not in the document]
davors72: Yeah I think it's best to use DominatorTree::Visit for the ordered traversals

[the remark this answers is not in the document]
s-perron: Okay

source/opt/basic_block.h
```diff
@@ -98,6 +100,12 @@ class BasicBlock {
     return --insts_.cend();
   }
 
+  // Return true if the basic block has at least one successor
+  inline bool hasSuccessor() const {
+    const auto br = &insts_.back();
+    return br->IsBranch();
```
dnovillo: More compact and less implementation-dependent: "return ctail()->IsBranch();"

dnovillo:
Additionally, there are formatting errors with the patch.  Please run clang-format.  https://travis-ci.org/KhronosGroup/SPIRV-Tools/jobs/311943875

source/opt/basic_block.h
```diff
@@ -98,6 +100,12 @@ class BasicBlock {
     return --insts_.cend();
   }
 
+  // Return true if the basic block has at least one successor
```
alan-baker: Returns and missing period.[Function comments](https://google.github.io/styleguide/cppguide.html#Function_Comments)

source/opt/dominator_analysis.h
```diff
@@ -0,0 +1,132 @@
...
+
+namespace spvtools {
+namespace opt {
+
+// Interface to perform dominator or postdominator analysis on a given function.
+class DominatorAnalysisBase {
+ public:
+  explicit DominatorAnalysisBase(bool is_post_dom) : tree_(is_post_dom) {}
+
+  // Calculate the dominator (or postdominator) tree for given function |f|.
```
alan-baker: Calculates.[Function comments](https://google.github.io/styleguide/cppguide.html#Function_Comments)

source/opt/dominator_analysis.h
```diff
@@ -0,0 +1,132 @@
...
+  inline void InitializeTree(const ir::Function* f, const ir::CFG& cfg) {
+    tree_.InitializeTree(f, cfg);
+  }
+
+  // Return true if BasicBlock |a| dominates BasicBlock |b|.
```
alan-baker: Returns.[Function comments](https://google.github.io/styleguide/cppguide.html#Function_Comments)

source/opt/dominator_analysis.h
```diff
@@ -0,0 +1,132 @@
...
+  inline bool Dominates(const ir::BasicBlock* a,
+                        const ir::BasicBlock* b) const {
+    if (!a || !b) return false;
+    return Dominates(a->id(), b->id());
+  }
+
+  // Return true if BasicBlock |a| dominates BasicBlock |b|. Same as above only
```
alan-baker: Returns.[Function comments](https://google.github.io/styleguide/cppguide.html#Function_Comments)

source/opt/dominator_analysis.h
```diff
@@ -0,0 +1,132 @@
...
+  // using the BasicBlock IDs.
+  inline bool Dominates(uint32_t a, uint32_t b) const {
+    return tree_.Dominates(a, b);
+  }
+
+  // Return true if BasicBlock |a| strictly dominates BasicBlock |b|.
```
alan-baker: Returns.[Function comments](https://google.github.io/styleguide/cppguide.html#Function_Comments)

source/opt/dominator_analysis.h
```diff
@@ -0,0 +1,132 @@
...
+  inline bool StrictlyDominates(const ir::BasicBlock* a,
+                                const ir::BasicBlock* b) const {
+    if (!a || !b) return false;
+    return StrictlyDominates(a->id(), b->id());
+  }
+
+  // Return true if BasicBlock |a| strictly dominates BasicBlock |b|. Same as
```
alan-baker: Returns.[Function comments](https://google.github.io/styleguide/cppguide.html#Function_Comments)

source/opt/dominator_analysis.h
```diff
@@ -0,0 +1,132 @@
...
+  // above only using the BasicBlock IDs.
+  inline bool StrictlyDominates(uint32_t a, uint32_t b) const {
+    return tree_.StrictlyDominates(a, b);
+  }
+
+  // Return the immediate dominator of |node| or return nullptr if it is has no
```
alan-baker: Returns.[Function comments](https://google.github.io/styleguide/cppguide.html#Function_Comments)

source/opt/dominator_analysis.h
```diff
@@ -0,0 +1,132 @@
...
+  // dominator.
+  inline ir::BasicBlock* ImmediateDominator(const ir::BasicBlock* node) const {
+    if (!node) return nullptr;
+    return tree_.ImmediateDominator(node);
+  }
+
+  // Return the immediate dominator of |node_id| or return nullptr if it is has
```
alan-baker: Returns.[Function comments](https://google.github.io/styleguide/cppguide.html#Function_Comments)

source/opt/dominator_analysis.h
```diff
@@ -0,0 +1,132 @@
...
+  // no dominator. Same as above but operates on IDs.
+  inline ir::BasicBlock* ImmediateDominator(uint32_t node_id) const {
+    return tree_.ImmediateDominator(node_id);
+  }
+
+  // Return true if |node| is reachable from the entry.
```
alan-baker: Returns.[Function comments](https://google.github.io/styleguide/cppguide.html#Function_Comments)

dnovillo:
Waiting for the CLA to be signed and for bots to finish.

s-perron:
Rebase, squashed, and merged as 8ba68fa9.

dnovillo:
Thanks.  The bots were still going when I checked last night.

PR #1023, opened 2017-11-27
Update README about the automatic master-tot relase

antiagainst:


antiagainst:
That is the initial release date of `master-tot`. We keep rewriting the artifacts of the same release. So it looks like it is outdated, but it is not. You can see there is a "0 commits to ... since this release". And if you click on the SHA1 of this release (on the left, after the `master-tot` tag), it is the current top of the tree.

dneto0:
> And if you click on the SHA1 of this release (on the left, after the master-tot tag), it is the current top of the tree.

I see that. Thanks.

dneto0:
The windows bot failure is already fixed, I think.

PR #1024, opened 2017-11-28
Fix windows build. unsigned vs signed comparison in EXPECT_EQ.

dnovillo:
This will fix the windows bots (https://travis-ci.org/KhronosGroup/SPIRV-Tools/jobs/308381700)


dnovillo:
Re-based and pushed as 8ffed974ac5dbf049612c670eb283b5ce03f344b

PR #1028, opened 2017-11-28
Compute value number when the value table is constructed.

s-perron:
Computing the value numbers on demand, as we do now, can lead to
different results depending on the order in which the users asks for
the value numbers.  To make things more stable, we compute them ahead
of time.

The algorithm used is similar to the dominator tree algorithm in [Preston Briggs, Keith D. Cooper, and L. Taylor Simpson. 1997. Value numbering. Softw. Pract. Exper. 27, 6 (June 1997), 701-724.](https://www.cs.rice.edu/~keith/Promo/CRPC-TR94517.pdf.gz).

### DIFF
--- a/source/opt/local_redundancy_elimination.cpp
+++ b/source/opt/local_redundancy_elimination.cpp
@@ -23,14 +23,10 @@ Pass::Status LocalRedundancyEliminationPass::Process(ir::IRContext* c) {
   InitializeProcessing(c);
 
   bool modified = false;
+  ValueNumberTable vnTable(context());
 
   for (auto& func : *get_module()) {
     for (auto& bb : func) {
-      // Resetting the value number table for every basic block because we just
-      // want the opportunities within a basic block. This will help keep
-      // register pressure down.
-      ValueNumberTable vnTable(context());
-
       // Keeps track of all ids that contain a given value number. We keep
       // track of multiple values because they could have the same value, but
       // different decorations.
@@ -39,7 +35,6 @@ Pass::Status LocalRedundancyEliminationPass::Process(ir::IRContext* c) {
         modified = true;
     }
   }
-
   return (modified ? Status::SuccessWithChange : Status::SuccessWithoutChange);
 }
 
@@ -54,6 +49,11 @@ bool LocalRedundancyEliminationPass::EliminateRedundanciesInBB(
     }
 
     uint32_t value = vnTable->GetValueNumber(inst);
+
+    if (value == 0) {
+      return;
+    }
+
     if (value >= value_to_ids->size()) {
       value_to_ids->resize(value + 1);
     }
--- a/source/opt/value_number_table.cpp
+++ b/source/opt/value_number_table.cpp
@@ -16,29 +16,49 @@
 
 #include <algorithm>
 
+#include "cfg.h"
+
 namespace spvtools {
 namespace opt {
 
 uint32_t ValueNumberTable::GetValueNumber(spvtools::ir::Instruction* inst) {
-  // TODO: Need to implement the substitution of operands by their value number
-  // before hashing.
-  // TODO: Implement a normal form for opcodes that commute like integer
-  // addition.  This will let us know that a+b is the same value as b+a.
   assert(inst->result_id() != 0 &&
          "inst must have a result id to get a value number.");
 
   // Check if this instruction already has a value.
-  auto id_to_val = id_to_value_.find(inst->result_id());
-  if (id_to_val != id_to_value_.end()) {
-    return id_to_val->second;
+  auto result_id_to_val = id_to_value_.find(inst->result_id());
+  if (result_id_to_val != id_to_value_.end()) {
+    return result_id_to_val->second;
+  }
+  return 0;
+}
+
+uint32_t ValueNumberTable::AssignValueNumber(ir::Instruction* inst) {
+  // If it already has a value return that.
+  uint32_t value = GetValueNumber(inst);
+  if (value != 0) {
+    return value;
   }
 
   // If the instruction has other side effects, then it must
   // have its own value number.
+  // OpSampledImage and OpImage must remain in the same basic block in which
+  // they are used, because of this we will assign each one it own value number.
   if (!context()->IsCombinatorInstruction(inst)) {
-    uint32_t value_number = TakeNextValueNumber();
-    id_to_value_[inst->result_id()] = value_number;
-    return value_number;
+    value = TakeNextValueNumber();
+    id_to_value_[inst->result_id()] = value;
+    return value;
+  }
+
+  switch (inst->opcode()) {
+    case SpvOpSampledImage:
+    case SpvOpImage:
+    case SpvOpVariable:
+      value = TakeNextValueNumber();
+      id_to_value_[inst->result_id()] = value;
+      return value;
+    default:
+      break;
   }
 
   // If it is a load from memory that can be modified, we have to assume the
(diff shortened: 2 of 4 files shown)


source/opt/value_number_table.cpp
```diff
@@ -48,24 +58,115 @@ uint32_t ValueNumberTable::GetValueNumber(spvtools::ir::Instruction* inst) {
...
+  }
+
+  // Replace all of the operands by their value number.  The sign bit will be
+  // set to distinguish between an id and a value number.
+  ir::Instruction value_ins(context(), inst->opcode(), inst->type_id(),
+                            inst->result_id(), {});
+  for (uint32_t o = 0; o < inst->NumInOperands(); ++o) {
+    const ir::Operand& op = inst->GetInOperand(o);
+    switch (op.type) {
+      case SPV_OPERAND_TYPE_ID: {
```
dneto0: This misses scope ID and memory semantics ID.  Prefer  if(spvOpIsIdTytpe(op.type)), which works here because you're already skipping past result-id and type-id (which is SPV_OPERAND_TYPE_TYPE_ID)

source/opt/value_number_table.cpp
```diff
@@ -48,24 +58,115 @@ uint32_t ValueNumberTable::GetValueNumber(spvtools::ir::Instruction* inst) {
...
+
+  for (auto& inst : context()->module()->ext_inst_imports()) {
+    if (inst.result_id() != 0) {
+      AssignValueNumber(&inst);
+    }
+  }
+
+  for (ir::Function& func : *context()->module()) {
+    std::list<ir::BasicBlock*> post_order;
+    context()->cfg()->ComputeStructuredOrder(&func, &*func.begin(),
```
dneto0: This is relying on structured control flow.  This constraint must be mentioned in the header file.
Later when we have general dominance-respecting-traversal, we can update it for the general case.

dneto0: Hmm.  Why not rely on SPIR-V's rule that the physical order of blocks in a module respects dominance order?  That's a universal validation rule.

source/opt/value_number_table.cpp
```diff
@@ -48,24 +58,115 @@ uint32_t ValueNumberTable::GetValueNumber(spvtools::ir::Instruction* inst) {
...
+    if (value != 0) {
+      id_to_value_[inst->result_id()] = value;
+      return value;
+    }
   }
 
+  // Phi nodes are a type of copy.  If all of the inputs have the same value
+  // number, then we can assign the result of the phi the same value number.
+  if (inst->opcode() == SpvOpPhi) {
+    value = GetValueNumber(inst->GetSingleWordInOperand(0));
```
dneto0: This is somewhat tricky.  It merits a test.

s-perron: Yes, I wrote test yesterday, but have not committed them yet.  I wrote test for phi nodes, CopyObj, and various types of control flow.

source/opt/value_number_table.cpp
s-perron: Done.

source/opt/value_number_table.cpp
s-perron: Yes, that should be exactly what I want.

source/opt/value_number_table.cpp
```diff
@@ -16,29 +16,49 @@
...
 
   // If the instruction has other side effects, then it must
   // have its own value number.
+  // OpSampledImage and OpImage must remain in the same basic block in which
+  // they are used, because of this we will assign each one it own value number.
   if (!context()->IsCombinatorInstruction(inst)) {
-    uint32_t value_number = TakeNextValueNumber();
-    id_to_value_[inst->result_id()] = value_number;
-    return value_number;
+    value = TakeNextValueNumber();
```
alan-baker: Why not just return an invalid value number? These special instructions (combinators, images, etc) seem to be treated differently than other things which don't have useful values. Below you refuse to assign a value to a phi if all the inputs are not the same, but you could assign a new value number. The two cases seem inconsistent.

source/opt/value_number_table.cpp
```diff
@@ -48,24 +68,111 @@ uint32_t ValueNumberTable::GetValueNumber(spvtools::ir::Instruction* inst) {
...
+    value = GetValueNumber(inst->GetSingleWordInOperand(0));
+    if (value != 0) {
+      id_to_value_[inst->result_id()] = value;
+      return value;
+    }
+  }
+
+  // Phi nodes are a type of copy.  If all of the inputs have the same value
+  // number, then we can assign the result of the phi the same value number.
+  if (inst->opcode() == SpvOpPhi) {
```
alan-baker: Although it would likely not be frequent, should this not handle duplicate phi nodes?
%5 = OpPhi %1 %2 %3 %4 %5
%6 = OpPhi %1 %2 %3 %4 %5

source/opt/local_redundancy_elimination.cpp
```diff
@@ -23,14 +23,10 @@ Pass::Status LocalRedundancyEliminationPass::Process(ir::IRContext* c) {
   InitializeProcessing(c);
 
   bool modified = false;
+  ValueNumberTable vnTable(context());
```
dnovillo: If you are not resetting the VNs at every basic block, isn't this effectively doing global VN?

source/opt/value_number_table.cpp
alan-baker: Misread below code.

source/opt/value_number_table.cpp
```diff
@@ -48,24 +68,111 @@ uint32_t ValueNumberTable::GetValueNumber(spvtools::ir::Instruction* inst) {
...
+    value = GetValueNumber(inst->GetSingleWordInOperand(0));
+    if (value != 0) {
+      for (uint32_t op = 2; op < inst->NumInOperands(); op += 2) {
```
dnovillo: Ugh, we really need a PhiOperand iterator.  Nothing to do with your change.  This just bugs me.

source/opt/value_number_table.cpp
```diff
@@ -48,24 +68,111 @@ uint32_t ValueNumberTable::GetValueNumber(spvtools::ir::Instruction* inst) {
...
+    if (inst.result_id() != 0) {
+      AssignValueNumber(&inst);
+    }
+  }
+
+  for (ir::Function& func : *context()->module()) {
+    // For best results we want to traverse the code in reverse post order.
+    // This happens naturally because of the forward referencing rules.
+    for (ir::BasicBlock& block : func) {
+      for (ir::Instruction& inst : block) {
```
dnovillo: But, this won't guarantee dom order visits, right?  You'd need to compute structured ordering before doing this.

source/opt/value_number_table.cpp
s-perron: I always assign a value number to phis.  I want to assign a value number to every instruction.  The first reason is that it leads to fewer special cases: I don't have to look for the invalid value number.  Second, I don't know how people will use them, so a more general approach seems better.

source/opt/local_redundancy_elimination.cpp
s-perron: No, the "value_to_ids" map gets reset for every basic block.  So the pass does not know which ids are available and contain the value number unless then come from the same basic block.

dnovillo: Ah, ok.  So, this contains the global table, but it uses the local value_to_ids for the actual transformation.  Thanks.

source/opt/value_number_table.cpp
s-perron: Yes it does. In section 2.16.1 under the heading for the CFG is says:

The order of blocks in a function must satisfy the rule that blocks appear before all blocks they dominate.

If is have a diamond there is not guarantee that it will visit the side nodes before the merge node.  However, I am not too concerned about that case.  If we do, good, if we don't, not a problem.

s-perron:
I've address @dneto0  comments.  If there are others I can address in another PR.  I'll merge.

dnovillo:
On Mon, Dec 4, 2017 at 11:01 AM, Steven Perron <notifications@github.com>
wrote:

> I've address @dneto0 <https://github.com/dneto0> comments. If there are
> others I can address in another PR. I'll merge.
>

SGTM

Issue #989, opened 2017-11-22
Odd loop is wrongly optimized.


Themaister:
I have this (very reduced) test shader in SPIRV-Cross:
```
#version 310 es
layout(local_size_x = 1) in;

layout(std430, binding = 0) readonly buffer SSBO
{
    mat4 mvp;
    vec4 in_data[];
};

layout(std430, binding = 1) writeonly buffer SSBO2
{
    vec4 out_data[];
};

void main()
{
    int k = 0;
	uint i = 0u;
    uint ident = gl_GlobalInvocationID.x;
	for (;;)
	{
		i += 2u;
		if (i > 20u)
			break;
	}

    vec4 idat = in_data[ident];
    out_data[ident] = idat;
}
```

When I compile that with glslang I get:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 1
; Bound: 56
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main" %gl_GlobalInvocationID
               OpExecutionMode %main LocalSize 1 1 1
               OpSource ESSL 310
               OpName %main "main"
               OpName %k "k"
               OpName %i "i"
               OpName %ident "ident"
               OpName %gl_GlobalInvocationID "gl_GlobalInvocationID"
               OpName %idat "idat"
               OpName %SSBO "SSBO"
               OpMemberName %SSBO 0 "mvp"
               OpMemberName %SSBO 1 "in_data"
               OpName %_ ""
               OpName %SSBO2 "SSBO2"
               OpMemberName %SSBO2 0 "out_data"
               OpName %__0 ""
               OpDecorate %gl_GlobalInvocationID BuiltIn GlobalInvocationId
               OpDecorate %_runtimearr_v4float ArrayStride 16
               OpMemberDecorate %SSBO 0 ColMajor
               OpMemberDecorate %SSBO 0 NonWritable
               OpMemberDecorate %SSBO 0 Offset 0
               OpMemberDecorate %SSBO 0 MatrixStride 16
               OpMemberDecorate %SSBO 1 NonWritable
               OpMemberDecorate %SSBO 1 Offset 64
               OpDecorate %SSBO BufferBlock
               OpDecorate %_ DescriptorSet 0
               OpDecorate %_ Binding 0
[...]

PR #1030, opened 2017-11-28
AggressiveDCE: Add merge and continue branches for live loop.

greg-lunarg:
This fixes issue #989


greg-lunarg:
@dneto0 Doh! Forgot the test. It is pretty much the code from #989, but I will add it right now.

greg-lunarg:
Yes, you have indeed surmised the mechanism of the solution. This solution will continue to work when we add the code to do empty loop removal.

dneto0:
Question:  In the example from #989, what happens if the break statement is replaced by an OpReturn or OpKill (if a fragment shader).  Do we need special exceptions for those as well, i.e. add them to the worklist?

greg-lunarg:
Re OpReturn, OpKill: These are not combinators ie are considered to have external effect, and thus are added to the initial worklist. No additional special exceptions are needed.

dneto0:
The bot failure is actually occuring on master, and is unrelated to this change.
I will push this into master.

dneto0:
Rebased and pushed into master as 8dd3d93cf66

PR #1032, opened 2017-11-29
Linker: Fix incorrect exit status.

daniel-schuermann:
The Linker returned exit code 1 on success. This patch fixes this behavior.

### DIFF
--- a/tools/link/linker.cpp
+++ b/tools/link/linker.cpp
@@ -140,11 +140,11 @@ int main(int argc, char** argv) {
   });
 
   std::vector<uint32_t> linkingResult;
-  bool succeed = linker.Link(contents, linkingResult, options);
+  bool success = linker.Link(contents, linkingResult, options);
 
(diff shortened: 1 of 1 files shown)


pierremoreau:
Oops, indeed. I might have had Linker::link return a bool at some point in the development, before moving to spv_result_t but forgot to change the linker tool.
I agree with @dneto0’s suggestion.

daniel-schuermann:
I understand that something like this can slip through :)
Indeed, |success| is also misleading. Although bool was already projecting down to a fixed number, this way it is more robust. Thanks for reviewing.

pierremoreau:
LGTM, and thanks for the fix!

dneto0:
Side note:  SPIRV-Tools does not have a test framework for checking the command line tools.  It's on a TODO list.  :-)

dneto0:
Rebased and pushed into master as 4c58086c56

PR #1033, opened 2017-11-29
Preserve SpvOpDecorationGroup when removing duplicates

daniel-schuermann:
Bugfix: The RemoveDuplicatesPass eliminated all Decoration Groups but one. This resulted in invalid SPIR-V modules, when there was previously more than one Decoration Group. Added a simple check if the opcode is SpvOpDecorationGroup to leave them inside.

### DIFF
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -87,7 +87,8 @@ bool DecorationManager::AreDecorationsTheSame(
   ir::Instruction tmpA = *inst1, tmpB = *inst2;
   if (tmpA.opcode() != tmpB.opcode() ||
       tmpA.NumInOperands() != tmpB.NumInOperands() ||
-      tmpA.opcode() == SpvOpNop || tmpB.opcode() == SpvOpNop)
+      tmpA.opcode() == SpvOpNop ||
+      tmpA.opcode() == SpvOpDecorationGroup)
     return false;
 
   for (uint32_t i = (tmpA.opcode() == SpvOpDecorate) ? 1u : 2u;


pierremoreau:
The code was absolutely not properly handling SpvOpDecorationGroup, sorry for that. Adding something like the following code (untested), before the for-loop, should fix it:
```
if (tmpA.opcode() == SpvOpDecorationGroup)
    return tmpA.result_id() == tmpB.result_id();
```

I doubt I added any tests for this, I’ll work on that.

daniel-schuermann:
I can add it, but can this actually happen? I cannot imagine a case where two SpvOpDecorationGroup would have the same result_id.

pierremoreau:
You are right, cause even if linking two modules, defining the same group of decorations, together, each decoration group should have a different ID, and hopefully would of them will be removed as duplicate. And defining the same ID twice shouldn’t be legal.
But in the hypothetical case where you have twice the same SpvOpDecorationGroup, AreDecorationsTheSame should still return true IMHO.

daniel-schuermann:
Okay, sounds plausible. Although I don't expect that this can harm anything, I would like to check this (tomorrow) before committing. You can add a change request, if you want this to be added.

pierremoreau:
I will add the change in another pull request containing some more fixes to the DecorationManager, as well as more unit tests.

daniel-schuermann:
@pierremoreau Do you also add a test case for multiple SpvOpDecorationGroup? Because my kernels are generated from OpenMP and are very cumbersome...
Then, this PR can be closed as soon as Pierre adds his.
Nonetheless, the RemoveDuplicatesPass needs a lot more testing than only one.

pierremoreau:
Yes, I’ll try to cover as many different decoration cases as possible, multiple SpvOpDecorationGroup included.

pierremoreau:
Just a small update: I continued working on it over the weekend but found more bugs within the decoration manager as I was adding tests. Current WIP can be found on [this branch](https://github.com/pierremoreau/SPIRV-Tools/tree/decoration_manager_fixes).

pierremoreau:
This can be closed as #1113 which fixes the issue has been merged and tests have been added.

dneto0:
> This can be closed as #1113 which fixes the issue has been merged and tests have been added. 

Closing

PR #1036, opened 2017-11-29
Fix #1034 - Give Edge::operator<() weak ordering semantics.

### LINKED ISSUE #1034: std::less on spvtools::opt::Edge is not a weak ordering
This is causing Windows Debug build timeout failures.

dnovillo:
This should fix #1034.  It changes the predicate on operator< to use
label IDs from each block and add them together to define a weak
ordering for std::set.

NOTE: This bug is happening on Windows, I do not have a Windows machine yet, so I'll wait for the bots to finish to make sure that the bug is properly fixed.

### DIFF
--- a/source/opt/propagator.h
+++ b/source/opt/propagator.h
@@ -18,6 +18,7 @@
 #include <functional>
 #include <queue>
 #include <set>
+#include <tuple>
 #include <unordered_map>
 #include <unordered_set>
 #include <vector>
@@ -36,7 +37,8 @@ struct Edge {
   ir::BasicBlock* source;
   ir::BasicBlock* dest;
   bool operator<(const Edge& o) const {
-    return source->id() + dest->id() < o.source->id() + o.dest->id();
+    return std::make_pair(source->id(), dest->id()) <
+           std::make_pair(o.source->id(), o.dest->id());
   }
 };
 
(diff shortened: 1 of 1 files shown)


source/opt/propagator.h
```diff
@@ -36,7 +36,7 @@ struct Edge {
   ir::BasicBlock* source;
   ir::BasicBlock* dest;
   bool operator<(const Edge& o) const {
-    return source < o.source || dest < o.dest;
+    return source->id() + dest->id() < o.source->id() + o.dest->id();
```
dneto0: At first I worried about unsigned overlflow. But that has defined behaviour.  Also, SPIR-V universal limits bound IDs to at most 0x3ffff.

dnovillo: Yeah, Given the limits on instructions, I figured the id space cannot be bigger than 16 bits.

atgoo: I don't think edges (2, 4) and (1, 5) should be considered equal. Please use std::pair, std::tie or manual comparisons:
```
if (source < o.source)
  return true;
if (o.source < source)
  return false;
if (dest < o.dest)
  return true;
if (o.dest < dest)
  return false;
```

dnovillo:
Re-based and pushed as 8cfa0c40e0afd5b6fcbedf622fd45afd2de2a055

PR #1037, opened 2017-11-29
Explicitly discard result of std::remove

dneto0:
Fixes Android arm-64-v8a build with NDK r14.

### DIFF
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -305,12 +305,13 @@ void DecorationManager::RemoveInstructionFromTarget(ir::Instruction* inst,
                                                     const uint32_t target_id) {
   auto const group_iter = group_to_decoration_insts_.find(target_id);
   if (group_iter != group_to_decoration_insts_.end()) {
-    remove(group_iter->second.begin(), group_iter->second.end(), inst);
+    auto& insts = group_iter->second;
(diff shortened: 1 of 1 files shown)


source/opt/decoration_manager.cpp
```diff
@@ -305,12 +305,13 @@ void DecorationManager::RemoveInstructionFromTarget(ir::Instruction* inst,
                                                     const uint32_t target_id) {
   auto const group_iter = group_to_decoration_insts_.find(target_id);
   if (group_iter != group_to_decoration_insts_.end()) {
-    remove(group_iter->second.begin(), group_iter->second.end(), inst);
+    (void)std::remove(group_iter->second.begin(), group_iter->second.end(),
+                      inst);
```
atgoo: Consider
group_iter->second.erase(std::remove(group_iter->second.begin(), group_iter->second.end(), inst), group_iter->second.end());

dneto0: That physically resizes the array, which seems unnecessary.  I don't have strong intuition about whether that's significantly wasteful.  I'll make the change (logically, anyway).

dneto0: Vector, not array.

dneto0:
Rebased against master, and rewrote the patch to erase the removed elements.

dneto0:
Please take another look.

dneto0:
Rebased and pushed into master as 188cd3780d76

PR #1039, opened 2017-11-30
Validate capabilities in OpenCL SPIR-V binaries

pierremoreau:
Extends the validator to not only check for Vulkan SPIR-V capabilities, but also for OpenCL SPIR-V ones. OpenCL 1.2 and 2.0 can ingest SPIR-V binaries thanks to the cl_khr_program_il extension, therefore add target environment enums for those.


pierremoreau:
The set of guaranteed capabilities is different between the OpenCL full profile and embedded profile, however there is currently no way to distinguish between them in SPIRV-Tools. Should we keep the current OpenCL target environment enums for the full profile, and introduce some SPV_ENV_OPENCL_EMBEDDED_x.y ones for embedded, or have SPV_ENV_OPENCL_FULL_x.y/SPV_ENV_OPENCL_EMBEDDED_x.y and get rid of the existing SPV_ENV_OPENCL_x.y ones?

pierremoreau:
I forgot to extend val_capability_test.cpp, my bad; I will do that later.

include/spirv-tools/libspirv.h
```diff
@@ -374,6 +374,8 @@ typedef enum {
   SPV_ENV_UNIVERSAL_1_0,  // SPIR-V 1.0 latest revision, no other restrictions.
   SPV_ENV_VULKAN_1_0,     // Vulkan 1.0 latest revision.
   SPV_ENV_UNIVERSAL_1_1,  // SPIR-V 1.1 latest revision, no other restrictions.
+  SPV_ENV_OPENCL_1_2,     // OpenCL 1.2 plus cl_khr_il_program, latest revision.
```
dneto0: Lets avoid reordering values of the other enums in the list, so please put the new ones at the end.

source/validate_capability.cpp
```diff
@@ -77,6 +77,54 @@ bool IsSupportOptionalVulkan_1_0(uint32_t capability) {
   return false;
 }
 
+bool IsSupportGuaranteedOpenCL_1_2(uint32_t capability) {
+  switch (capability) {
+    case SpvCapabilityAddresses:
+    case SpvCapabilityFloat16Buffer:
+    case SpvCapabilityGroups:
+    case SpvCapabilityInt64:
```
dneto0: The embedded profile requires cles_khr_int64 in order to support Int64

dneto0: I'm ok with this if we interpret an OpenCL environment without qualification as the full profile.   That should be documented in the enum definition.

include/spirv-tools/libspirv.h
dneto0: Please put a comment at the top of the enum declaration saying that by default we mean the OpenCL full profile.

dneto0:
> The set of guaranteed capabilities is different between the OpenCL full profile and embedded profile, however there is currently no way to distinguish between them in SPIRV-Tools. Should we keep the current OpenCL target environment enums for the full profile, and introduce some SPV_ENV_OPENCL_EMBEDDED_x.y ones for embedded, or have SPV_ENV_OPENCL_FULL_x.y/SPV_ENV_OPENCL_EMBEDDED_x.y and get rid of the existing SPV_ENV_OPENCL_x.y ones?

I suggest the first option.  Assume full profile, and make a special name for embedded profile.

source/validate_capability.cpp
pierremoreau: I completely ignored the embedded profile for now, until we decided out to differentiate it from the full profile (if we wanted to make that distinction). I’ll update the series to handle the embedded profile as well.

pierremoreau:
> I can't find the cl_khr_program_il extension text (!)

Maybe if I gave you the proper name, you would have found it: it’s cl_khr_il_program, not cl_khr_program_il 😓.
It is not part of the list of extensions found at https://www.khronos.org/registry/OpenCL/, but you can find its specification in the OpenCL [1.2](https://www.khronos.org/registry/OpenCL/specs/opencl-1.2-extensions.pdf) and [2.0](https://www.khronos.org/registry/OpenCL/specs/opencl-1.2-extensions.pdf) extension specification document. (It’s also not part of any OpenCL headers right now, I plan to open an issue about it on the OpenCL-Headers repo.)

I will update the pull request to include SPV_ENV_OPENCL_EMBEDDED_x.y enums and handle them.

include/spirv-tools/libspirv.h
pierremoreau: I edited the comment next to each OPENCL flag to indicate it means the full profile: is that enough do you prefer having a comment at the top as well?

pierremoreau:
I added the embedded profile but still didn’t touch val_capability_test.cpp. I’ll probably need to go in and update the target environments supported by some of the tools.

pierremoreau:
Please, take a look.

I believe it is now ready for review. There are a few things I am not 100% confident about the tests:
1. Is this enough?
2. Should I keep tests like “OpenCL22FullEnabledByCapability” when capabilities enabled by other capabilities are currently the same for every OpenCL version, so it doesn’t bring anything more than what “OpenCL12FullEnabledByCapability” is doing?

On a side note, I started adding OpenCL extensions to SPIRV-Tools but decided to leave it out of this pull request. I could take another stab at it if you prefer to have it part of this pull request, but it looked like it wouldn’t as easy as first thought, as there is no mapping between capabilities and extensions (for example, in “spirv.core.grammar.json”, capability “StorageBuffer16BitAccess” is associated to extenion “SPV_KHR_16bit_storage”.

dneto0:
> Should I keep tests like “OpenCL22FullEnabledByCapability” when capabilities enabled by other capabilities are currently the same for every OpenCL version, so it doesn’t bring anything more than what “OpenCL12FullEnabledByCapability” is doing?

This is at your discretion.  It's fine to keep them.  It's informative and very explicit this way. It saves an inference or lookup into the relevant spec.

dneto0:
Rebased, squashed, and pushed into master as 12447d8465

Issue #1012, opened 2017-11-24
Dref should be the depth-comparison reference value


antiagainst:
https://github.com/KhronosGroup/SPIRV-Tools/blob/e1a6f8d1c3ec4f814d6160ab46b8f397f89f8184/source/validate_image.cpp#L771-L776

Dref should be a float scalar based on my understanding of HLSL. But the spec is unclear about its type:

> Dref is the depth-comparison reference value.

PR #1042, opened 2017-11-30
Fix Dref type check in validator

antiagainst:
Dref should be of 32-bit scalar floating type.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1012

### DIFF
--- a/source/validate_image.cpp
+++ b/source/validate_image.cpp
@@ -852,7 +852,7 @@ spv_result_t ImagePass(ValidationState_t& _,
       }
 
       const uint32_t dref_type = _.GetOperandTypeId(inst, 4);
-      if (!_.IsFloatScalarType(dref_type)) {
+      if (!_.IsFloatScalarType(dref_type) || _.GetBitWidth(dref_type) != 32) {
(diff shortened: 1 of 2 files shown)


source/validate_image.cpp
```diff
@@ -852,10 +852,10 @@ spv_result_t ImagePass(ValidationState_t& _,
...
 
       const uint32_t dref_type = _.GetOperandTypeId(inst, 4);
-      if (dref_type != info.sampled_type) {
+      if (!_.IsFloatScalarType(dref_type)) {
         return _.diag(SPV_ERROR_INVALID_DATA)
-               << "Expected Dref to be of Image 'Sampled Type': "
-               << spvOpcodeString(opcode);
+               << spvOpcodeString(opcode)
+               << ": Expected Dref to be of 32-bit float type";
       }
```
atgoo: There is another function for bit width:
_.GetBitWidth(dref_type) != 32

antiagainst: Yep! Modified. Thanks!

dneto0:
Rebased, squashed, and pushed into master as 0dd4ee27b16

PR #1043, opened 2017-12-01
Apply EXCLUDE_FROM_ALL to effcee

antiagainst:
Saves us a few targets to build by default


dneto0:
There are two very fast-running effcee tests that normally run.  But now they fail because the underlying code was not built ?

I would still like the SPIRV-Tools public bots to run those tests, to make sure our test infra is working at least in basic ways.   Can you reframe this?

antiagainst:
Hmm. Works fine for me on MacOS. Didn't check windows previously. I'll abandon this PR then since it saves just a small amount of time.

PR #1044, opened 2017-12-01
Appveyor: use ninja instead of MSBuild

antiagainst:


dneto0:
Ninja's faster, right?

dneto0:
Rebased and pushed onto master as b8aeab82ea97

antiagainst:
Yes, it's faster. :)

Issue #1029, opened 2017-11-28
Run clang-format on pull requests


pierremoreau:
It’s easy to forget running clang-format before submitting a pull request. Reviewers currently look for wrong formatting, but it could be checked for automatically by a build bot, or as part of an additional check being run on the pull request.

Do some of you have experience with some syntax formatting check being run along the other checks on pull requests?
Maybe something like [this](https://github.com/dokan-dev/dokany/commit/c8ffc494482e299ed025855214db839a561ceab7)? Though it seems like the test still appears as successful on GitHub, even if the formatting needed to be fixed.

dneto0:
I try not to be too picky with formatting with pull requests.  Normally I find something else more important that I'd like fixed.  :-)

Hey @antiagainst what do you think?

I assume that's PowerShell. I've never learned PowerShell. :-)

dnovillo:
I am *ALL* for something like this.  I am always using clang-format to keep my code formatted and I sometimes have tons of noise in my patches when editing existing files.

I'm also unfamiliar with this scripting language.  But if it does what it says, it's got my vote.

antiagainst:
Oh, it suddenly comes to me that clang-format has a mode to only format the changed code: https://clang.llvm.org/docs/ClangFormat.html#script-for-patch-reformatting. So it may be fine.

dnovillo:
Yeah, I was going to suggest that option for clang-format.  I've also found that if your style file is stable, the output is stable.  Particularly when using a well-established format like the Google style

antiagainst:
Yep.

We don't want to add more checks into Appveyor. Appveyor is running all jobs sequentially. It already takes more than 1 hour to run a push. Adding checks to Travis is better, IMHO.

dnovillo:
@antiagainst, thanks.  Would you be able to make this change then?  I'm not sure where to start, but if you give me a couple of pointers I can get it done.

antiagainst:
Not a problem. I can do it.

Before that, can we do a wholesale reformatting over the current codebase?

PR #1045, opened 2017-12-01
Travis CI: add clang-format check

antiagainst:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1029

### DIFF
--- a/utils/check_code_format.sh
+++ b/utils/check_code_format.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+# Copyright (c) 2017 Google Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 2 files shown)


utils/check_code_format.sh
```diff
@@ -0,0 +1,24 @@
...
+# This script assumes to be invoked at the project root directory.
+
+FILES_TO_CHECK=$(git diff --name-only master | grep -E ".*\.(cpp|cc|c\+\+|cxx|c|h|hpp)$")
+
+if [ -z "${FILES_TO_CHECK}" ]; then
+  echo "No source code to check for formatting."
+  exit 0
+fi
+
+FORMAT_DIFF=$(git diff -U0 master -- ${FILES_TO_CHECK} | ./utils/clang-format-diff.py -p1 -style=Google)
```
pierremoreau: Shouldn’t it rather be using the style defined in the .clang-format files (which is the Google style, but with one additional tweak)?

utils/clang-format-diff.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python
+#
+#===- clang-format-diff.py - ClangFormat Diff Reformatter ----*- python -*--===#
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
+# License. See LICENSE.TXT for details.
```
dnovillo: LICENSE.TXT should be included here.

utils/check_code_format.sh
antiagainst: Done

utils/clang-format-diff.py
antiagainst: I've changed to download this file for each build.

PR #1047, opened 2017-12-01
Add validation rules for atomic instructions

atgoo:
Validates all OpAtomicXXX instructions.

### DIFF
--- a/source/validate_atomics.cpp
+++ b/source/validate_atomics.cpp
@@ -25,12 +25,13 @@ namespace libspirv {
 
 // Validates correctness of atomic instructions.
 spv_result_t AtomicsPass(ValidationState_t& _,
-                             const spv_parsed_instruction_t* inst) {
+                         const spv_parsed_instruction_t* inst) {
   const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
   const uint32_t result_type = inst->type_id;
 
   switch (opcode) {
     case SpvOpAtomicLoad:
+    case SpvOpAtomicStore:
     case SpvOpAtomicExchange:
     case SpvOpAtomicCompareExchange:
     case SpvOpAtomicCompareExchangeWeak:
@@ -44,24 +45,36 @@ spv_result_t AtomicsPass(ValidationState_t& _,
     case SpvOpAtomicUMax:
     case SpvOpAtomicAnd:
     case SpvOpAtomicOr:
-    case SpvOpAtomicXor: {
+    case SpvOpAtomicXor:
+    case SpvOpAtomicFlagTestAndSet:
+    case SpvOpAtomicFlagClear: {
       if (opcode == SpvOpAtomicLoad || opcode == SpvOpAtomicExchange ||
           opcode == SpvOpAtomicCompareExchange) {
         if (!_.IsFloatScalarType(result_type) &&
             !_.IsIntScalarType(result_type)) {
           return _.diag(SPV_ERROR_INVALID_DATA)
-              << spvOpcodeString(opcode)
-              << ": expected Result Type to be int or float scalar type";
+                 << spvOpcodeString(opcode)
+                 << ": expected Result Type to be int or float scalar type";
         }
+      } else if (opcode == SpvOpAtomicFlagTestAndSet) {
+        if (!_.IsBoolScalarType(result_type)) {
+          return _.diag(SPV_ERROR_INVALID_DATA)
+                 << spvOpcodeString(opcode)
+                 << ": expected Result Type to be bool scalar type";
+        }
+      } else if (opcode == SpvOpAtomicFlagClear || opcode == SpvOpAtomicStore) {
+        assert(result_type == 0);
       } else {
         if (!_.IsIntScalarType(result_type)) {
           return _.diag(SPV_ERROR_INVALID_DATA)
-              << spvOpcodeString(opcode)
-              << ": expected Result Type to be int scalar type";
+                 << spvOpcodeString(opcode)
+                 << ": expected Result Type to be int scalar type";
         }
       }
 
-      const uint32_t pointer_type = _.GetOperandTypeId(inst, 2);
+      uint32_t operand_index =
+          opcode == SpvOpAtomicFlagClear || opcode == SpvOpAtomicStore ? 0 : 2;
+      const uint32_t pointer_type = _.GetOperandTypeId(inst, operand_index++);
 
       uint32_t data_type = 0;
       uint32_t storage_class = 0;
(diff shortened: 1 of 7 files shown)


atgoo:
Passes Vulkan CTS master.

source/validate_atomics.cpp
```diff
@@ -0,0 +1,260 @@
...
+        }
+      }
+
+      break;
+    }
+
+    case SpvOpAtomicStore: {
+      const uint32_t pointer_type = _.GetOperandTypeId(inst, 0);
+      uint32_t data_type = 0;
+      uint32_t storage_class = 0;
```
antiagainst: We are not checking storage class here?

2.16.1. Universal Validation Rules

> The pointers taken by atomic operation instructions must be a pointer into one of the following Storage Classes:
>  *  Uniform when used with the BufferBlock Decoration
>  *  Workgroup
>  *  CrossWorkgroup
>  *  Generic
>  *  AtomicCounter
>  *  Image

source/validate_atomics.cpp
```diff
@@ -0,0 +1,260 @@
...
+    case SpvOpAtomicIAdd:
+    case SpvOpAtomicISub:
+    case SpvOpAtomicSMin:
+    case SpvOpAtomicUMin:
+    case SpvOpAtomicSMax:
+    case SpvOpAtomicUMax:
+    case SpvOpAtomicAnd:
+    case SpvOpAtomicOr:
+    case SpvOpAtomicXor: {
+      if (opcode == SpvOpAtomicLoad || opcode == SpvOpAtomicExchange ||
```
antiagainst: Hmm, I also find the following requirements in 2.16.1. Universal Validation Rules:

> All pointers used in atomic operation instructions must be pointers to one of the following:
> *    32-bit scalar integer
> *    64-bit scalar integer

@dneto0: The above effectively forbids having floats as return types?

source/validate_atomics.cpp
```diff
@@ -0,0 +1,260 @@
...
+      uint32_t storage_class = 0;
+      if (!_.GetPointerTypeInfo(pointer_type, &data_type, &storage_class)) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+               << spvOpcodeString(opcode)
+               << ": expected Pointer to be of type OpTypePointer";
+      }
+
+      if (data_type != result_type) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+               << spvOpcodeString(opcode)
```
antiagainst: Nit: Maybe it's just my eyes. But I generally find it a little bit hard to associate the error message with the the problematic parameter since we don't have clear marking of which parameter is wrong. I'd suggest putting the parameter names in `<>` to make it clear, e.g., `expected <Pointer> to point to a value of type <Result Type>`. What's better, add colors to them! :)

source/validate_atomics.cpp
```diff
@@ -0,0 +1,260 @@
...
+               << ": expected Pointer to point to a value of type Result Type";
+      }
+
+      const uint32_t scope_type = _.GetOperandTypeId(inst, 3);
+      if (!_.IsIntScalarType(scope_type) || _.GetBitWidth(scope_type) != 32) {
```
antiagainst: Scope should be a SPIR-V constant whose value is <= 4. I'm not sure whether we have constant query methods to check this.

source/validate_atomics.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+      const uint32_t memory_semantics_type = _.GetOperandTypeId(inst, 2);
+      if (!_.IsIntScalarType(memory_semantics_type) ||
+          _.GetBitWidth(memory_semantics_type) != 32) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+               << spvOpcodeString(opcode)
+               << ": expected Memory Semantics to be 32-bit int";
+      }
+
+      if (opcode == SpvOpAtomicStore) {
```
antiagainst: Redundant check?

source/validate_atomics.cpp
```diff
@@ -0,0 +1,260 @@
...
+              << spvOpcodeString(opcode)
+              << ": expected Value type and the type pointed to by Pointer to"
+              << " be the same";
+        }
+      }
+
+      break;
+    }
+
+    case SpvOpAtomicFlagTestAndSet: {
```
antiagainst: I feel this case can be merged with the main case for `SpvOpAtomicLoad` and others to share some checking logic?

source/validate_atomics.cpp
```diff
@@ -0,0 +1,260 @@
...
+          _.GetBitWidth(memory_semantics_type1) != 32) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+               << spvOpcodeString(opcode)
+               << ": expected Memory Semantics to be 32-bit int";
+      }
+
+      break;
+    }
+
+    case SpvOpAtomicFlagClear: {
```
antiagainst: I feel this case can be merged with the case for `SpvOpAtomicStore` to share some logic?

source/validate_atomics.cpp
atgoo: I already have hundreds of error messages which use capitalization to mark keywords. I don't want to have to change it everywhere.

source/validate_atomics.cpp
atgoo: Done.

source/validate_atomics.cpp
antiagainst: That's fine.

source/validate_atomics.cpp
dneto0: Correct.   I know of no atomic-float operations.

atgoo: So, is this a spec bug?
OpAtomicLoad

Atomically load through Pointer using the given Semantics. All subparts of the value that is loaded will be read atomically with respect to all other atomic accesses to it within Scope.

Result Type must be a scalar of integer type or floating-point type.

source/validate_atomics.cpp
```diff
@@ -0,0 +1,196 @@
...
+      uint32_t data_type = 0;
+      uint32_t storage_class = 0;
+      if (!_.GetPointerTypeInfo(pointer_type, &data_type, &storage_class)) {
+        return _.diag(SPV_ERROR_INVALID_DATA)
+               << spvOpcodeString(opcode)
+               << ": expected Pointer to be of type OpTypePointer";
+      }
+
+      switch (storage_class) {
+        case SpvStorageClassUniform:
```
dneto0: Should also allow StorageBuffer storage class.
That was introduced in SPV_KHR_storage_buffer_storage_class, but that's not incorporated in 1.0, 1.1, 1.2.

dneto0:
Rebased, squashed, and merged as 532b327d4

PR #1049, opened 2017-12-01
Add simplified instructions for pushing a reviewed PR.

dnovillo:
github keeps PRs in the refs/pull namespace, so it is a bit simpler to just get a PR directly from there instead of the more convoluted process of getting the user id and branch to review.

For reviewing PR 9999, the flow then becomes:

$ git fetch upstream
$ git checkout pr/9999

The rest remains the same.  The new documentation explains how to do the first time setup and subsequent workflow.

I've tried it with other PRs and it is no different than before.  Just shorter.  I will also try it with this one, once approved.


pierremoreau:
Nice, I didn’t know one could do that!

Issue #1051, opened 2017-12-04
DCEInst is trying to delete the same instruction twice.


s-perron:
When calling DCEInst, if an instruction is deleted that contains the same ids twice, it is possible that the instruction defining that id will end up in the work queue twice.  This leads to the that instruction being processed after it has been killed.

PR #1052, opened 2017-12-04
DCEInst kill the same instruction twice.

s-perron:
In DCEInst, it is possible that the same instruction ends up in the
queue multiple times, if the same id is used multiple times in the
same instruction.

The solution is to keep the ids in a set, to ensure no duplication in
the list.

Fixes issue #1051

### DIFF
--- a/source/opt/mem_pass.cpp
+++ b/source/opt/mem_pass.cpp
@@ -184,7 +184,7 @@ void MemPass::AddStores(uint32_t ptr_id, std::queue<ir::Instruction*>* insts) {
 }
 
 void MemPass::DCEInst(ir::Instruction* inst,
-                      const function<void(ir::Instruction * )>& call_back) {
+                      const function<void(ir::Instruction*)>& call_back) {
   std::queue<ir::Instruction*> deadInsts;
   deadInsts.push(inst);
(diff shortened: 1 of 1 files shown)


source/opt/mem_pass.cpp
```diff
@@ -195,8 +195,8 @@ void MemPass::DCEInst(ir::Instruction* inst,
       continue;
     }
     // Remember operands
-    std::vector<uint32_t> ids;
-    di->ForEachInId([&ids](uint32_t* iid) { ids.push_back(*iid); });
+    std::set<uint32_t> ids;
+    di->ForEachInId([&ids](uint32_t* iid) { ids.insert(*iid); });
```
atgoo: Consider
const uint32_t* iid

atgoo:
This fixes a crash in spirv-opt --eliminate-local-single-block that I observed earlier today.

source/opt/mem_pass.cpp
s-perron: Yes that would be better, but I don't think it is worth waiting for the build bot again.

s-perron:
I'll merge if the build bots are successful.

PR #1054, opened 2017-12-04
Remove cfg_ field from SSAPropagator class - NFC.

dnovillo:
When I moved the CFG into IRContext
(https://github.com/KhronosGroup/SPIRV-Tools/pull/1019), I forgot to
update SSAPropagator to stop requiring one.

Fixed with this patch.


s-perron:
LGTM

dnovillo:
Re-based and pushed as e9ecc0cbfda13f9c2731dc02209bd7153fe1d203.

PR #1056, opened 2017-12-04
Change Release builds to Release with debug info on Linux

dnovillo:
For Linux builds, we want to add debug information on the generated binary (for timing and profiling purposes).  This patch changes Release builds to RelWithDebInfo.


dnovillo:
Travis build passed.  Submitted as 726573a0e546e0d7ce5431db00d309542488406d.

PR #1061, opened 2017-12-05
Add a scalar replacement pass

alan-baker:
The pass considers all function scopevariables of composite type. If there are accesses to individual elements (and it is legal) the pass replaces the variable with a variable for each composite element and updates all the uses.

Added the pass to -O (not -Os)
Added NumUses and NumUsers to DefUseManager
Added some helper methods for the inst to block mapping in context
Added some helper methods for specific constant types

### DIFF
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -20,6 +20,7 @@
 #include "decoration_manager.h"
 #include "def_use_manager.h"
 #include "dominator_analysis.h"
+#include "feature_manager.h"
 #include "module.h"
 #include "type_manager.h"
 
@@ -370,6 +371,14 @@ class IRContext {
   // Return the next available SSA id and increment it.
   inline uint32_t TakeNextId() { return module()->TakeNextIdBound(); }
 
+  opt::FeatureManager* get_feature_mgr() {
+    if (!feature_mgr_.get()) {
+      AnalyzeFeatures();
+    }
+    return feature_mgr_.get();
+  }
+
+
  private:
   // Builds the def-use manager from scratch, even if it was already valid.
   void BuildDefUseManager() {
@@ -400,6 +409,11 @@ class IRContext {
     valid_analyses_ = valid_analyses_ | kAnalysisCFG;
   }
 
+  void AnalyzeFeatures() {
+    feature_mgr_.reset(new opt::FeatureManager());
+    feature_mgr_->Analyze(module());
+  }
+
   // Scans a module looking for it capabilities, and initializes combinator_ops_
   // accordingly.
   void InitializeCombinators();
@@ -428,6 +442,7 @@ class IRContext {
 
   // The instruction decoration manager for |module_|.
   std::unique_ptr<opt::analysis::DecorationManager> decoration_mgr_;
+  std::unique_ptr<opt::FeatureManager> feature_mgr_;
 
   // A map from instructions the the basic block they belong to. This mapping is
   // built on-demand when get_instr_block() is called.
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -197,6 +197,19 @@ class IRContext {
     auto entry = instr_to_block_.find(instr);
     return (entry != instr_to_block_.end()) ? entry->second : nullptr;
   }
+  ir::BasicBlock* get_instr_block(uint32_t id) {
+    ir::Instruction* def = get_def_use_mgr()->GetDef(id);
+    return get_instr_block(def);
+  }
+
+  // Sets the basic block for |inst|. Re-builds the mapping if it has become invalid.
+  void set_instr_block(ir::Instruction* inst, ir::BasicBlock* block) {
+    if (AreAnalysesValid(kAnalysisInstrToBlockMapping)) {
+      instr_to_block_[inst] = block;
+    } else {
+      BuildInstrToBlockMapping();
+    }
+  }
 
   // Returns a pointer the decoration manager.  If the decoration manger is
   // invalid, it is rebuilt first.
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -197,6 +197,11 @@ class IRContext {
     auto entry = instr_to_block_.find(instr);
     return (entry != instr_to_block_.end()) ? entry->second : nullptr;
   }
+
+  // Returns the basic block for |id|. Re-builds the instruction block map, if
+  // needed.
+  //
+  // |id| must be a registered definition.
   ir::BasicBlock* get_instr_block(uint32_t id) {
     ir::Instruction* def = get_def_use_mgr()->GetDef(id);
     return get_instr_block(def);
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -409,6 +409,7 @@ class IRContext {
     valid_analyses_ = valid_analyses_ | kAnalysisCFG;
   }
 
+  // Analyzes the features in the owned module. Builds the manager if required.
   void AnalyzeFeatures() {
     feature_mgr_.reset(new opt::FeatureManager());
     feature_mgr_->Analyze(module());
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -378,7 +378,6 @@ class IRContext {
     return feature_mgr_.get();
   }
 
-
  private:
   // Builds the def-use manager from scratch, even if it was already valid.
   void BuildDefUseManager() {
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -202,7 +202,8 @@ class IRContext {
     return get_instr_block(def);
   }
 
-  // Sets the basic block for |inst|. Re-builds the mapping if it has become invalid.
+  // Sets the basic block for |inst|. Re-builds the mapping if it has become
+  // invalid.
   void set_instr_block(ir::Instruction* inst, ir::BasicBlock* block) {
     if (AreAnalysesValid(kAnalysisInstrToBlockMapping)) {
       instr_to_block_[inst] = block;
--- a/source/opt/scalar_replacement_pass.cpp
+++ b/source/opt/scalar_replacement_pass.cpp
@@ -627,7 +627,8 @@ bool ScalarReplacementPass::CheckUses(const ir::Instruction* inst,
   return ok;
 }
 
-bool ScalarReplacementPass::CheckUsesRelaxed(const ir::Instruction* inst) const {
+bool ScalarReplacementPass::CheckUsesRelaxed(
+    const ir::Instruction* inst) const {
   bool ok = true;
   get_def_use_mgr()->ForEachUse(
       inst, [this, &ok](const ir::Instruction* user, uint32_t index) {
--- a/source/opt/scalar_replacement_pass.cpp
+++ b/source/opt/scalar_replacement_pass.cpp
@@ -489,13 +489,7 @@ bool ScalarReplacementPass::CanReplaceVariable(
     return false;
 
   const ir::Instruction* typeInst = GetStorageType(varInst);
-  if (!CheckType(typeInst)) return false;
-
-  if (!CheckAnnotations(varInst)) return false;
-
-  if (!CheckUses(varInst)) return false;
-
-  return true;
+  return CheckType(typeInst) && CheckAnnotations(varInst) && CheckUses(varInst);
 }
 
 bool ScalarReplacementPass::CheckType(const ir::Instruction* typeInst) const {
@@ -579,9 +573,8 @@ bool ScalarReplacementPass::CheckAnnotations(
 }
 
 bool ScalarReplacementPass::CheckUses(const ir::Instruction* inst) const {
-  bool ok = true;
   VariableStats stats = {0, 0};
-  CheckUses(inst, &stats, &ok);
+  bool ok = CheckUses(inst, &stats);
 
   // TODO(alanbaker): Extend this to some meaningful heuristics about when
   // SRoA is valuable.
@@ -590,10 +583,11 @@ bool ScalarReplacementPass::CheckUses(const ir::Instruction* inst) const {
   return ok;
 }
 
-void ScalarReplacementPass::CheckUses(const ir::Instruction* inst,
-                                      VariableStats* stats, bool* ok) const {
+bool ScalarReplacementPass::CheckUses(const ir::Instruction* inst,
+                                      VariableStats* stats) const {
+  bool ok = true;
   get_def_use_mgr()->ForEachUse(
-      inst, [this, stats, ok](const ir::Instruction* user, uint32_t index) {
+      inst, [this, stats, &ok](const ir::Instruction* user, uint32_t index) {
         // Annotations are check as a group separately.
         if (!ir::IsAnnotationInst(user->opcode())) {
           switch (user->opcode()) {
@@ -603,58 +597,62 @@ void ScalarReplacementPass::CheckUses(const ir::Instruction* inst,
                 uint32_t id = user->GetSingleWordOperand(3u);
                 const ir::Instruction* opInst = get_def_use_mgr()->GetDef(id);
                 if (!ir::IsCompileTimeConstantInst(opInst->opcode())) {
-                  *ok = false;
+                  ok = false;
                 } else {
-                  CheckUsesRelaxed(user, ok);
+                  if (!CheckUsesRelaxed(user)) ok = false;
                 }
                 stats->num_partial_accesses++;
               } else {
-                *ok = false;
+                ok = false;
               }
               break;
             case SpvOpLoad:
-              if (!CheckLoad(user, index)) *ok = false;
+              if (!CheckLoad(user, index)) ok = false;
               stats->num_full_accesses++;
               break;
             case SpvOpStore:
-              if (!CheckStore(user, index)) *ok = false;
+              if (!CheckStore(user, index)) ok = false;
               stats->num_full_accesses++;
               break;
             case SpvOpName:
             case SpvOpMemberName:
               break;
             default:
-              *ok = false;
+              ok = false;
               break;
           }
         }
       });
+
+  return ok;
 }
 
-void ScalarReplacementPass::CheckUsesRelaxed(const ir::Instruction* inst,
-                                             bool* ok) const {
+bool ScalarReplacementPass::CheckUsesRelaxed(const ir::Instruction* inst) const {
+  bool ok = true;
   get_def_use_mgr()->ForEachUse(
-      inst, [this, ok](const ir::Instruction* user, uint32_t index) {
+      inst, [this, &ok](const ir::Instruction* user, uint32_t index) {
         switch (user->opcode()) {
           case SpvOpAccessChain:
           case SpvOpInBoundsAccessChain:
             if (index != 2u) {
-              *ok = false;
+              ok = false;
             } else {
-              CheckUsesRelaxed(user, ok);
+              if (!CheckUsesRelaxed(user)) ok = false;
             }
             break;
           case SpvOpLoad:
-            if (!CheckLoad(user, index)) *ok = false;
+            if (!CheckLoad(user, index)) ok = false;
             break;
           case SpvOpStore:
-            if (!CheckStore(user, index)) *ok = false;
+            if (!CheckStore(user, index)) ok = false;
             break;
           default:
-            *ok = false;
+            ok = false;
             break;
         }
       });
+
+  return ok;
 }
 
 bool ScalarReplacementPass::CheckLoad(const ir::Instruction* inst,
--- a/source/opt/scalar_replacement_pass.cpp
+++ b/source/opt/scalar_replacement_pass.cpp
@@ -209,7 +209,7 @@ bool ScalarReplacementPass::ReplaceAccessChain(
   // indexes) or a direct use of the replacement variable.
   uint32_t indexId = chain->GetSingleWordInOperand(1u);
   const ir::Instruction* index = get_def_use_mgr()->GetDef(indexId);
-  uint64_t indexValue = GetConstantInteger(index);
+  size_t indexValue = GetConstantInteger(index);
   if (indexValue > replacements.size()) {
     // Out of bounds access, this is illegal IR.
     return false;
@@ -410,7 +410,7 @@ void ScalarReplacementPass::GetOrCreateInitialValue(ir::Instruction* source,
   }
 }
 
-uint64_t ScalarReplacementPass::GetIntegerLiteral(const ir::Operand& op) const {
+size_t ScalarReplacementPass::GetIntegerLiteral(const ir::Operand& op) const {
   assert(op.words.size() <= 2);
   uint64_t len = 0;
   for (uint32_t i = 0; i != op.words.size(); ++i) {
@@ -419,7 +419,7 @@ uint64_t ScalarReplacementPass::GetIntegerLiteral(const ir::Operand& op) const {
   return len;
 }
 
-uint64_t ScalarReplacementPass::GetConstantInteger(const ir::Instruction* constant) const {
+size_t ScalarReplacementPass::GetConstantInteger(const ir::Instruction* constant) const {
   assert(constant->opcode() == SpvOpConstant || constant->opcode() == SpvOpConstantNull);
   if (constant->opcode() == SpvOpConstantNull) {
     return 0;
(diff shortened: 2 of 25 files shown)


source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+  CheckUses(inst, &stats, &ok);
+
+  // TODO(alanbaker): Extend this to some meaningful heuristics about when SRoA
+  // is valuable.
+  if (stats.num_partial_accesses == 0) ok = false;
+
+  return ok;
+}
+
+void ScalarReplacementPass::CheckUses(const ir::Instruction* inst,
```
s-perron: Why not have it return a bool instead of having |ok| as a parameter?

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+        switch (user->opcode()) {
+          case SpvOpAccessChain:
+          case SpvOpInBoundsAccessChain:
+            if (index == 2u) {
+              uint32_t id = user->GetSingleWordOperand(3u);
+              const ir::Instruction* opInst = get_def_use_mgr()->GetDef(id);
+              if (!ir::IsCompileTimeConstantInst(opInst->opcode())) {
+                *ok = false;
+              } else {
+                CheckUsesRelaxed(user, ok);
```
s-perron: Not obvious  what you are looking for and why.  Can you add to the comments in the header file?  My guess is that you want to make sure the address does not "escape".  However, why would if matter if the address of the member escapes as say a parameter on a function call?  You can pass it the address of the new variable.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+}
+
+bool ScalarReplacementPass::CheckUses(const ir::Instruction* inst) const {
+  bool ok = true;
+  VariableStats stats = {0, 0};
+  CheckUses(inst, &stats, &ok);
+
+  // TODO(alanbaker): Extend this to some meaningful heuristics about when SRoA
+  // is valuable.
+  if (stats.num_partial_accesses == 0) ok = false;
```
s-perron: +1

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+bool ScalarReplacementPass::ReplaceVariable(
+    ir::Instruction* inst, std::queue<ir::Instruction*>* worklist) {
+  std::vector<ir::Instruction*> replacements;
+  CreateReplacementVariables(inst, &replacements);
+
+  bool ok = true;
+  std::vector<ir::Instruction*> dead;
+  dead.push_back(inst);
+  get_def_use_mgr()->ForEachUser(
+      inst, [this, &ok, replacements, &dead](ir::Instruction* user) {
```
s-perron: Looks like a value capture of |replacements|, could this cause many call to the copy constructor for the vector?

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+      constantId = iter->second;
+    }
+
+    // Create a new constant extract.
+    newInitId = TakeNextId();
+    context()->AddGlobalValue(MakeUnique<ir::Instruction>(
+        context(), SpvOpSpecConstantOp, storageId, newInitId,
+        std::initializer_list<ir::Operand>{
+            {SPV_OPERAND_TYPE_SPEC_CONSTANT_OP_NUMBER, {SpvOpCompositeExtract}},
+            {SPV_OPERAND_TYPE_ID, {constantId}}}));
```
s-perron: Are you sure this is the correct instruction?  An OpCompositeExtract takes more than 1 operand.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+  } else if (init->opcode() == SpvOpConstantComposite) {
+    // Get the appropriate index constant.
+    newInitId = init->GetSingleWordInOperand(index);
+    ir::Instruction* element = get_def_use_mgr()->GetDef(newInitId);
+    if (element->opcode() == SpvOpUndef) {
+      // Undef is not a valid initializer for a variable.
+      newInitId = 0;
+    }
+  } else {
+    assert(false);
```
s-perron: The spec specifically states that the initializer could be the id of a global variable.  Add a comment that you are specifically disallowing it.  Also I don't see the code in |CanReplaceVariable| that disallows this case.  Should be added there.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+  for (uint32_t i = 0; i != op.words.size(); ++i) {
+    len |= (op.words[i] << (32 * i));
+  }
+  return len;
+}
+
+size_t ScalarReplacementPass::GetConstantInteger(
+    const ir::Instruction* constant) const {
+  assert(constant->opcode() == SpvOpConstant ||
+         constant->opcode() == SpvOpConstantNull);
```
s-perron: Should assert in the type as well?  OpConstant is uses for floating point constants as well.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+
+  uint32_t ptrTypeId = inst->type_id();
+  uint32_t typeId =
+      get_def_use_mgr()->GetDef(ptrTypeId)->GetSingleWordInOperand(1u);
+  return get_def_use_mgr()->GetDef(typeId);
+}
+
+bool ScalarReplacementPass::CanReplaceVariable(
+    const ir::Instruction* varInst) const {
+  assert(varInst->opcode() == SpvOpVariable);
```
s-perron: Might want to check for decorations on variables as well.  A variable could be decorated as volatile for example.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+
+    default:
+      assert(false && "Unexpected type.");
+      break;
+  }
+}
+
+void ScalarReplacementPass::CreateVariable(
+    uint32_t typeId, ir::Instruction* varInst, uint32_t index,
+    std::vector<ir::Instruction*>* replacements) {
```
s-perron: Somewhere in here you might need to copy decorations from the original struct to the new variable.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,598 @@
...
+      if (GetArrayLength(typeInst) > MAX_NUM_ELEMENTS) return false;
+      return true;
+    // Specifically including matrix and vector in an attempt to reduce the
+    // number of vector registers required.
+    case SpvOpTypeMatrix:
+    case SpvOpTypeVector:
+      if (GetNumElements(typeInst) > MAX_NUM_ELEMENTS) return false;
+      return true;
+
+    case SpvOpTypeRuntimeArray:
```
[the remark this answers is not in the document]
alan-baker: I just wanted to explicitly call out that it should not be handled despite being an aggregate type.

source/opt/scalar_replacement_pass.cpp
alan-baker: Convenience due to the recursive nature and interaction with ForEachUser lambda.

source/opt/scalar_replacement_pass.cpp
alan-baker: This checks that the pointer ends up in a load or store and not something else. If this were a kernel, I could do ptrtoint and then weird stuff. I don't want that to be allowed.

source/opt/scalar_replacement_pass.cpp
alan-baker: I'm not sure that's appropriate, but I will peruse the decorations section in the spec.

source/opt/scalar_replacement_pass.cpp
alan-baker: Function scope variables cannot be decorated as volatile, but I'll see if there are other decorations that are inappropriate.

source/opt/scalar_replacement_pass.cpp
alan-baker: Sure I can add that.

source/opt/scalar_replacement_pass.cpp
alan-baker: Since OpVariable returns a pointer, I won't see any valid cases for scalarization (only composite types are handled).

source/opt/scalar_replacement_pass.cpp
alan-baker: I'll double check. It passes the validator though.

source/opt/scalar_replacement_pass.cpp
alan-baker: Good catch.

source/opt/ir_context.h
```diff
@@ -192,6 +192,20 @@ class IRContext {
     auto entry = instr_to_block_.find(instr);
     return (entry != instr_to_block_.end()) ? entry->second : nullptr;
   }
+  ir::BasicBlock* get_instr_block(uint32_t id) {
```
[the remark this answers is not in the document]
alan-baker: Ok

source/opt/scalar_replacement_pass.cpp
alan-baker: You're right, I'll fix this.

source/opt/scalar_replacement_pass.cpp
alan-baker: I've added code to handle decorations.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+              assert(false && "Unexpected opcode");
+              break;
+          }
+        }
+      });
+
+  // There was an illegal access.
+  if (!ok) return false;
+
+  // Clean up some dead code.
```
dnovillo: Why not just leave this to DCE?  Makes the pass slightly cleaner.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+    context()->set_instr_block(&*iter, block);
+
+    // Create the store.
+    std::unique_ptr<ir::Instruction> newStore(
+        new ir::Instruction(context(), SpvOpStore, 0, 0,
+                            std::initializer_list<ir::Operand>{
+                                {SPV_OPERAND_TYPE_ID, {var->result_id()}},
+                                {SPV_OPERAND_TYPE_ID, {extractId}}}));
+    // Copy memory access attributes.
+    for (uint32_t i = 2; i < store->NumInOperands(); ++i) {
```
dnovillo: Why skipping the first two?  This needs a comment.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+bool ScalarReplacementPass::ReplaceAccessChain(
+    ir::Instruction* chain, const std::vector<ir::Instruction*>& replacements) {
+  // Replaces the access chain with either another access chain (with one fewer
+  // indexes) or a direct use of the replacement variable.
+  uint32_t indexId = chain->GetSingleWordInOperand(1u);
+  const ir::Instruction* index = get_def_use_mgr()->GetDef(indexId);
+  size_t indexValue = GetConstantInteger(index);
+  if (indexValue > replacements.size()) {
+    // Out of bounds access, this is illegal IR.
+    return false;
```
dnovillo: Easier to just assert?  Nothing good will happen from this point on.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+    if (chain->NumInOperands() > 2) {
+      // Replace input access chain with another access chain.
+      ir::BasicBlock::iterator chainIter(chain);
+      uint32_t replacementId = TakeNextId();
+      std::unique_ptr<ir::Instruction> replacementChain(new ir::Instruction(
+          context(), chain->opcode(), chain->type_id(), replacementId,
+          std::initializer_list<ir::Operand>{
+              {SPV_OPERAND_TYPE_ID, {var->result_id()}}}));
+      // Add the remaining indexes.
+      for (uint32_t i = 2; i < chain->NumInOperands(); ++i) {
```
dnovillo: Likewise.  You're using *InOperands() so it's not immediately obvious why the loop starts at 2.

source/opt/scalar_replacement_pass.h
```diff
@@ -0,0 +1,215 @@
...
+  // If there is an initial value for |source| for element |index|, it is
+  // appended as an operand on |newVar|. If the initial value is OpUndef, no
+  // initial value is added to |newVar|.
+  void GetOrCreateInitialValue(ir::Instruction* source, uint32_t index,
+                               ir::Instruction* newVar);
+
+  // Returns the result id of a 32 bit integer.
+  //
+  // Searches all types for a decoration-less integer before creating one.
+  uint32_t GetIntId();
```
dnovillo: Why not use TypeManager?

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+
+  // If varInst was initialized, make sure to initialize its replacement.
+  GetOrCreateInitialValue(varInst, index, inst);
+  get_def_use_mgr()->AnalyzeInstDefUse(inst);
+  context()->set_instr_block(inst, block);
+
+  replacements->push_back(inst);
+}
+
+uint32_t ScalarReplacementPass::GetOrCreatePointerType(uint32_t id) {
```
dnovillo: This is likely better implemented in analysis::TypeManager.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+          {SPV_OPERAND_TYPE_STORAGE_CLASS, {SpvStorageClassFunction}},
+          {SPV_OPERAND_TYPE_ID, {id}}}));
+  ir::Instruction* ptr = &*--context()->types_values_end();
+  get_def_use_mgr()->AnalyzeInstDefUse(ptr);
+  pointee_to_pointer_[id] = ptrId;
+
+  return ptrId;
+}
+
+uint32_t ScalarReplacementPass::GetIntId() {
```
dnovillo: Likewise with this one.  The TypeManager would give you an existing Int, if it finds one (or should ;)

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+                                                    uint32_t index,
+                                                    ir::Instruction* newVar) {
+  assert(source->opcode() == SpvOpVariable);
+  if (source->NumInOperands() < 2) return;
+
+  uint32_t initId = source->GetSingleWordInOperand(1u);
+  uint32_t storageId = GetStorageType(newVar)->result_id();
+  ir::Instruction* init = get_def_use_mgr()->GetDef(initId);
+  uint32_t newInitId = 0;
+  if (init->opcode() == SpvOpConstantNull) {
```
dnovillo: There is code in fold_spec_constant_op_and_composite_pass.cpp that does something similar, but I'm currently in the midst of re-factoring it for use with CCP.  Add a TODO for me to deal with this after the dust settles?

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+  } else {
+    assert(false);
+  }
+
+  if (newInitId != 0) {
+    newVar->AddOperand({SPV_OPERAND_TYPE_ID, {newInitId}});
+  }
+}
+
+size_t ScalarReplacementPass::GetIntegerLiteral(const ir::Operand& op) const {
```
dnovillo: This looks like a useful method for Operand or Instruction.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+  assert(op.words.size() <= 2);
+  size_t len = 0;
+  for (uint32_t i = 0; i != op.words.size(); ++i) {
+    len |= (op.words[i] << (32 * i));
+  }
+  return len;
+}
+
+size_t ScalarReplacementPass::GetConstantInteger(
```
dnovillo: Likewise.  There's actually a few more.  Not sure if they all make sense, but any methods that are now taking an ir::Instruction* as parameter may be good for just adding inside ir::Instruction.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+  if (!CheckTypeAnnotations(get_def_use_mgr()->GetDef(varInst->type_id())))
+    return false;
+
+  const ir::Instruction* typeInst = GetStorageType(varInst);
+  if (!CheckType(typeInst)) return false;
+
+  if (!CheckAnnotations(varInst)) return false;
+
+  if (!CheckUses(varInst)) return false;
+
```
dnovillo: Consider just 'return CheckType && CheckAnnotations && CheckUses'?

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+      if (GetNumElements(typeInst) > MAX_NUM_ELEMENTS) return false;
+      return true;
+
+    case SpvOpTypeRuntimeArray:
+    default:
+      return false;
+  }
+}
+
+bool ScalarReplacementPass::CheckTypeAnnotations(
```
dnovillo: The name of this predicate is confusing.  What is it checking?

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,702 @@
...
+        break;
+      default:
+        return false;
+    }
+  }
+
+  return true;
+}
+
+bool ScalarReplacementPass::CheckAnnotations(
```
dnovillo: Likewise here.

source/opt/scalar_replacement_pass.cpp
dnovillo: But you can capture &ok in the lambda and return it.

test/opt/pass_fixture.h
```diff
@@ -89,11 +89,23 @@ class PassTest : public TestT {
   // disassembly string and the boolean value from the pass Process() function.
   template <typename PassT, typename... Args>
   std::tuple<std::string, opt::Pass::Status> SinglePassRunAndDisassemble(
-      const std::string& assembly, bool skip_nop, Args&&... args) {
+      const std::string& assembly, bool skip_nop, bool do_validation,
```
dnovillo: default do_validation to 'true'?

tools/opt/opt.cpp
```diff
@@ -197,17 +195,23 @@ Options (in lexicographical order):
                'spirv-opt --merge-blocks -O ...' applies the transformation
                --merge-blocks followed by all the transformations implied by
                -O.
-  --relax-store-struct
+  --relax-struct-store
```
dnovillo: I'd rather not change flag names here.  There may be external users relying on the previous name.  Better leave it for another PR.

source/opt/ir_context.h
```diff
@@ -357,6 +384,12 @@ class IRContext {
     valid_analyses_ = valid_analyses_ | kAnalysisCFG;
   }
 
+  // Analyzes the features in the owned module. Builds the manager if required.
+  void AnalyzeFeatures() {
+    feature_mgr_.reset(new opt::FeatureManager());
+    feature_mgr_->Analyze(module());
+  }
```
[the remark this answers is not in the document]
alan-baker: I don't think we're planning on adding any optimizations that enable features, so it shouldn't change.

source/opt/scalar_replacement_pass.cpp
alan-baker: It makes it easier to test and I know exactly what is dead as opposed to scanning the whole module.

source/opt/scalar_replacement_pass.cpp
alan-baker: Ok

source/opt/scalar_replacement_pass.cpp
alan-baker: This causes the pass to return Failure, which I think is better than only catching it in debug.

source/opt/scalar_replacement_pass.cpp
alan-baker: ok

source/opt/scalar_replacement_pass.cpp
alan-baker: The type manager is pretty bare bones right now. I agree this method would be better implemented elsewhere, but I'd rather have a separate change for that.

source/opt/scalar_replacement_pass.cpp
alan-baker: Same as above, I'd prefer to leave this refactor to another PR.

source/opt/scalar_replacement_pass.cpp
alan-baker: ok

source/opt/scalar_replacement_pass.cpp
alan-baker: As mentioned in the header it checks annotations on a type. This is different than CheckAnnotations because different annotations are acceptable for variables and types.

test/opt/pass_fixture.h
alan-baker: I didn't want to make that choice for other tests. Its possible they are using invalid IR for their testcases.

tools/opt/opt.cpp
alan-baker: This option exists in the validator as --relax-struct-store, but in opt it was --relax-store-struct. This simply unifies those options since they do the same thing.

source/opt/scalar_replacement_pass.cpp
alan-baker: I take it back, I'll make what changes I can to include the type manager.

source/opt/scalar_replacement_pass.cpp
alan-baker: This dead, I've removed it.

source/opt/scalar_replacement_pass.cpp
alan-baker: I'll leave a todo to improve the TypeManager, its not very useful as is.

source/opt/scalar_replacement_pass.cpp
alan-baker: I'd like to leave that for a larger discussion of sub-classes. Same with the comments about instruction functions.

source/opt/scalar_replacement_pass.h
alan-baker: This is dead, removed.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,678 @@
...
+namespace opt {
+
+// Heuristic aggregate element limit.
+const uint32_t MAX_NUM_ELEMENTS = 100u;
+
+Pass::Status ScalarReplacementPass::Process(ir::IRContext* c) {
+  InitializeProcessing(c);
+
+  Status status = Status::SuccessWithoutChange;
+  for (auto& f : *get_module()) {
```
dnovillo: Use ProcessReachableCallTree() here?

source/opt/scalar_replacement_pass.cpp
dnovillo: Hmm, OK.  I generally prefer to keep transformations simple, but I agree that this may not be a big deal.

source/opt/scalar_replacement_pass.cpp
s-perron: Now that we have dead function elimination running, I don't think it makes a difference.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -0,0 +1,678 @@
...
+      return true;
+    case SpvOpTypeArray:
+      if (GetArrayLength(typeInst) > MAX_NUM_ELEMENTS) return false;
+      return true;
+    // Specifically including matrix and vector in an attempt to reduce the
+    // number of vector registers required.
+    case SpvOpTypeMatrix:
+    case SpvOpTypeVector:
+      if (GetNumElements(typeInst) > MAX_NUM_ELEMENTS) return false;
+      return true;
```
s-perron: I looked at how this pass affects code.  I think for now it would be a good idea to not split matrices and vectors.  In another PR, we could see how to improve the heuristics for them.

s-perron:
Rebased, squashed and merged as 867451f4.

Issue #700, opened 2017-07-07
Global value numbering.


dneto0:
Should eliminate duplicate computed values, e.g. in constants section.
Cover Undef cases as well.  Extra undefs could be introduced by Type2Undef method in https://github.com/KhronosGroup/SPIRV-Tools/pull/691

dneto0:
See also opportunity in IfThenElse test in local_ssa_elim_test.

in vec4 BaseColor;
vec4 v;
if (c) { v = BaseColor * 0.5; } else { v = BaseColor + vec4(1.0); }

Results in redundant load of BaseColor.    Could be eliminated with GVN on Input storage class because it's constant (?)

s-perron:
@dneto0 I implemented a GVN based commoning algorithm, but it will not get this case.  The reason is that BaseColor is not loaded multiple times on the same path.

Usually commoning algorithms want to be "lazy".  That is they try not to compute (or load) a value too far before its use because that needlessly increases register pressure.  If you really want to get this case, let me know, and we can discuss how we could do it.

dneto0:
Never mind. Your commoning algorithm is good enough!

PR #1062, opened 2017-12-06
Add global redundancy elimination

s-perron:
Adds a pass that looks for redundant instruction in a function, and
removes them.  The algorithm is a hash table based value numbering
algorithm that traverses the dominator tree.

This pass removes completely redundant instructions, not partially
redundant ones.

This finish issue #960, and #700.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -432,6 +432,11 @@ Optimizer::PassToken CreateMergeReturnPass();
 // This pass will look for instructions in the same basic block that compute the
 // same value, and remove the redundant ones.
 Optimizer::PassToken CreateLocalRedundancyEliminationPass();
+
+// Create global value numbering pass.
+// This pass will look for instructions where the same value is computed on all
+// paths leading to the instruction.  Those instructions are deleted.
+Optimizer::PassToken CreateRedundancyEliminationPass();
 }  // namespace spvtools
 
 #endif  // SPIRV_TOOLS_OPTIMIZER_HPP_
--- a/source/opt/local_redundancy_elimination.h
+++ b/source/opt/local_redundancy_elimination.h
@@ -33,18 +33,29 @@ class LocalRedundancyEliminationPass : public Pass {
  public:
   const char* name() const override { return "local-redundancy-elimination"; }
   Status Process(ir::IRContext*) override;
+  virtual ir::IRContext::Analysis GetPreservedAnalyses() override {
+    return ir::IRContext::kAnalysisDefUse |
+           ir::IRContext::kAnalysisInstrToBlockMapping |
+           ir::IRContext::kAnalysisDecorations |
+           ir::IRContext::kAnalysisCombinators | ir::IRContext::kAnalysisCFG |
+           ir::IRContext::kAnalysisDominatorAnalysis;
+  }
 
- private:
-  // Deletes instructions in |block| whose value is in |vnTable| or is computed
-  // earlier in |block|. The values computed in |block| are added to |vnTable|.
-  // |value_to_ids| is a map from a value number to the result ids known to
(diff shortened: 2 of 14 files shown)


test/opt/redundancy_elimination_test.cpp
```diff
@@ -0,0 +1,276 @@
...
+#include "pass_utils.h"
+
+namespace {
+
+using namespace spvtools;
+
+using ::testing::HasSubstr;
+using ::testing::MatchesRegex;
+
+using LocalRedundancyEliminationTest = PassTest<::testing::Test>;
```
alan-baker: LocalRedundancyEliminationTest -> RedundancyEliminationTest

test/opt/redundancy_elimination_test.cpp
```diff
@@ -0,0 +1,276 @@
...
+               OpFunctionEnd
+
+  )";
+  auto result = SinglePassRunAndDisassemble<opt::RedundancyEliminationPass>(
+      text, /* skip_nop = */ true);
+  EXPECT_EQ(opt::Pass::Status::SuccessWithoutChange, std::get<1>(result));
+}
+
+// Keep the add.  Even if it is redundant on all paths, there is no single id
+// whose definition dominates the add and contains thae same value.
```
alan-baker: thae -> the

dnovillo:
Looking at this now.   Apologies for the delay.

source/opt/local_redundancy_elimination.h
```diff
@@ -33,18 +33,29 @@ class LocalRedundancyEliminationPass : public Pass {
  public:
   const char* name() const override { return "local-redundancy-elimination"; }
   Status Process(ir::IRContext*) override;
+  virtual ir::IRContext::Analysis GetPreservedAnalyses() override {
+    return ir::IRContext::kAnalysisDefUse |
+           ir::IRContext::kAnalysisInstrToBlockMapping |
+           ir::IRContext::kAnalysisDecorations |
+           ir::IRContext::kAnalysisCombinators | ir::IRContext::kAnalysisCFG |
+           ir::IRContext::kAnalysisDominatorAnalysis;
```
dnovillo: So, should we start listing the analyses that we do *not* preserve?

include/spirv-tools/optimizer.hpp
```diff
@@ -432,6 +432,11 @@ Optimizer::PassToken CreateMergeReturnPass();
 // This pass will look for instructions in the same basic block that compute the
 // same value, and remove the redundant ones.
 Optimizer::PassToken CreateLocalRedundancyEliminationPass();
+
+// Create global value numbering pass.
+// This pass will look for instructions in the same function that compute the
+// same value, and remove the redundant ones.
```
dnovillo: Clarify that this is done in cases where there is **total** redundancy: "compute the same value through all execution paths", maybe?

test/opt/redundancy_elimination_test.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+using ::testing::HasSubstr;
+using ::testing::MatchesRegex;
+
+using LocalRedundancyEliminationTest = PassTest<::testing::Test>;
+
+#ifdef SPIRV_EFFCEE
+// Test that it can get a simple case of local redundancy elimination.
+// The rest of the test check for extra functionality.
+TEST_F(LocalRedundancyEliminationTest, RemoveRedundantLocalAdd) {
```
dnovillo: Local?  Seems like a remnant from the local RE PR.

s-perron:
Made the suggested changes.  I will merge once all of the bots are done.  You have until then to add other comments.

source/opt/local_redundancy_elimination.h
s-perron: That is an option.

PR #1063, opened 2017-12-06
Add depth first iterator for tree structures

Naghasan:
Add the possibility to walk a tree structure in a depth first order using an iterator.

Included in this patch:
 - Add generic depth first iterator for tree structures;
 - Update the dominator tree to use this iterator instead of "randomly" iterate over the nodes.

### DIFF
--- a/source/opt/dominator_tree.h
+++ b/source/opt/dominator_tree.h
@@ -22,6 +22,7 @@
 
 #include "cfg.h"
 #include "module.h"
+#include "tree_iterator.h"
 
 namespace spvtools {
 namespace opt {
@@ -35,6 +36,16 @@ struct DominatorTreeNode {
         children_({}),
         dfs_num_pre_(-1),
         dfs_num_post_(-1) {}
+
+  using iterator = std::vector<DominatorTreeNode*>::iterator;
+  using const_iterator = std::vector<DominatorTreeNode*>::const_iterator;
+
+  iterator begin() { return children_.begin(); }
+  iterator end() { return children_.end(); }
+  const_iterator begin() const { return cbegin(); }
+  const_iterator end() const { return cend(); }
+  const_iterator cbegin() const { return children_.begin(); }
+  const_iterator cend() const { return children_.end(); }
 
   inline uint32_t id() const { return bb_->id(); }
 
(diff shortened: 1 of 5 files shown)


source/opt/tree_iterator.h
```diff
@@ -0,0 +1,121 @@
...
+      return;
+    }
+    std::pair<NodePtr, NodeIterator>& next_it = parent_iterators_.top();
+    // set the new node.
+    current_ = *next_it.second;
+    // update the iterator for the next child.
+    ++next_it.second;
+    // if we finished with node, pop it.
+    if (next_it.first->end() == next_it.second) parent_iterators_.pop();
+    // If our current node is not a leaf, store the iteration state for latter.
```
dnovillo: s/latter/later/

source/opt/tree_iterator.h
```diff
@@ -0,0 +1,121 @@
...
+  TreeDFIterator& operator++() {
+    MoveToNextNode();
+    return *this;
+  }
+
+  TreeDFIterator operator++(int) {
+    TreeDFIterator tmp = *this;
+    ++*this;
+    return tmp;
+  }
```
dnovillo: Do we really need a postfix ++ here?  Not sure how it would be useful.

Naghasan: That was only for conformance with InputIterator concept

dnovillo: ok, thanks.

dnovillo:
Waiting for bots to finish.

source/opt/dominator_tree.h
```diff
@@ -67,12 +78,12 @@ class DominatorTree {
   DominatorTree() : postdominator_(false) {}
   explicit DominatorTree(bool post) : postdominator_(post) {}
 
-  iterator begin() { return nodes_.begin(); }
-  iterator end() { return nodes_.end(); }
+  iterator begin() { return iterator(GetRoot()); }
```
s-perron: Should add comments telling uses about the order.

source/opt/dominator_tree.h
```diff
@@ -144,30 +155,20 @@ class DominatorTree {
 
   // Applies the std::function |func| to all nodes in the dominator tree.
```
s-perron: Should mention that what the user can expect from the order of the traversal.

Naghasan:
@s-perron I documented the traversal ordering.

dnovillo:
Rebased and pushed upstream as 731d1899b19bc5151ca5460a627d2f5cdb2dd5d3

Issue #622, opened 2017-04-07
inline-entry-points-all: forgets to delete names of Ids that go away


dneto0:
Inliner forgets to delete OpName targeting the inlined return value of a called function.

Example input is below. In particular, `%foo_result` is the name given to the result of the function call.

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main"
               OpName %main "main"
               OpName %main_entry "main_entry"
               OpName %foo_result "foo_result"
               OpName %void_fn "void_fn"
               OpName %foo "foo"
               OpName %foo_entry "foo_entry"
       %void = OpTypeVoid
    %void_fn = OpTypeFunction %void
        %foo = OpFunction %void None %void_fn
  %foo_entry = OpLabel
               OpReturn
               OpFunctionEnd
       %main = OpFunction %void None %void_fn
 %main_entry = OpLabel
 %foo_result = OpFunctionCall %void %foo
               OpReturn
               OpFunctionEnd
```

After running the inliner I get:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main"
               OpName %main "main"
               OpName %main_entry "main_entry"
               OpName %foo_result "foo_result"
               OpName %void_fn "void_fn"
               OpName %foo "foo"
               OpName %foo_entry "foo_entry"
       %void = OpTypeVoid
    %void_fn = OpTypeFunction %void
        %foo = OpFunction %void None %void_fn
  %foo_entry = OpLabel
               OpReturn
               OpFunctionEnd
       %main = OpFunction %void None %void_fn
 %main_entry = OpLabel
               OpReturn
               OpFunctionEnd
```
We still have the `OpName %foo_result` but that Id is no longer defined, and this module fails validation:

```
error: 19: The following forward referenced IDs have not been defined:
3[foo_result]
```

It's a bug, yes, but I don't think this is particularly high priority since I only encountered it with a hand-coded assembly module.

Cc @greg-lunarg

greg-lunarg:
Is the "before" shader valid? foo is defined as a void function but it is called with a return value. The spec says that void functions do not return a value, but there does not appear to be a way to call a function without a return value. So that means that %foo_result's type is void. Is that valid? There doesn't seem to be validation rule that rules out values or variables from having type void. Should there be? My code had (apparently naively) assumed there wasn't a return value for void functions.

[...]

PR #1066, opened 2017-12-07
Kill names and decoration in inlining.

s-perron:
Currently when inlining a call, the name and decorations for the result of the call is not deleted.  This should be changed.  Just before deleting the basic block that contains the call, we delete the name and decorations for the call.

This fixes issue #622.


s-perron:
I think I should add a test as well.

PR #1067, opened 2017-12-07
ADCE: Empty Loop Elimination

greg-lunarg:
This entirely eliminates loops which do not contain live code.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -101,21 +101,33 @@ void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
   live_local_vars_.insert(varId);
 }
 
-bool AggressiveDCEPass::IsStructuredIfHeader(ir::BasicBlock* bp,
-                                             ir::Instruction** mergeInst,
-                                             ir::Instruction** branchInst,
-                                             uint32_t* mergeBlockId) {
-  auto ii = bp->end();
-  --ii;
-  if (ii->opcode() != SpvOpBranchConditional) return false;
-  if (ii == bp->begin()) return false;
-  if (branchInst != nullptr) *branchInst = &*ii;
-  --ii;
-  if (ii->opcode() != SpvOpSelectionMerge) return false;
-  if (mergeInst != nullptr) *mergeInst = &*ii;
-  if (mergeBlockId != nullptr)
-    *mergeBlockId =
-        ii->GetSingleWordInOperand(kSelectionMergeMergeBlockIdInIdx);
+bool AggressiveDCEPass::IsStructuredHeader(ir::BasicBlock* bp, SpvOp mergeOp,
+                                           ir::Instruction** mergeInst,
+                                           ir::Instruction** branchInst,
+                                           uint32_t* mergeBlockId) {
+  ir::Instruction* tempMergeInst;
+  if (!IsStructuredIfOrLoopHeader(bp, &tempMergeInst, branchInst, mergeBlockId))
+    return false;
+  if (mergeInst != nullptr) *mergeInst = tempMergeInst;
+  return tempMergeInst->opcode() == mergeOp;
+}
+
+bool AggressiveDCEPass::IsStructuredIfOrLoopHeader(ir::BasicBlock* bp,
+                                                   ir::Instruction** mergeInst,
+                                                   ir::Instruction** branchInst,
+                                                   uint32_t* mergeBlockId) {
+  auto brii = bp->end();
+  --brii;
+  if (brii == bp->begin()) return false;
+  auto mii = brii;
+  --mii;
+  if (mii->opcode() == SpvOpSelectionMerge) {
+    if (brii->opcode() != SpvOpBranchConditional) return false;
+  } else if (mii->opcode() != SpvOpLoopMerge)
+    return false;
+  if (branchInst != nullptr) *branchInst = &*brii;
+  if (mergeInst != nullptr) *mergeInst = &*mii;
+  if (mergeBlockId != nullptr) *mergeBlockId = mii->GetSingleWordInOperand(0);
   return true;
 }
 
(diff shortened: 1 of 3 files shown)


source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -101,21 +101,33 @@ void AggressiveDCEPass::ProcessLoad(uint32_t varId) {
...
+    return false;
+  if (mergeInst != nullptr) *mergeInst = tempMergeInst;
+  return tempMergeInst->opcode() == mergeOp;
+}
+
+bool AggressiveDCEPass::IsStructuredIfOrLoopHeader(ir::BasicBlock* bp,
+                                                   ir::Instruction** mergeInst,
+                                                   ir::Instruction** branchInst,
+                                                   uint32_t* mergeBlockId) {
+  auto brii = bp->end();
```
alan-baker: BasicBlock has some useful functionality in this regard. Pleas rewrite in terms of BasicBlock::GetMergeInst, BasicBlock::tail and BasicBlock::MergeIdIfAny.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -268,18 +291,33 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
   while (!worklist_.empty()) {
     ir::Instruction* liveInst = worklist_.front();
     // Add all operand instructions if not already live
-    liveInst->ForEachInId([this](const uint32_t* iid) {
+    SpvOp op = liveInst->opcode();
+    liveInst->ForEachInId([&op, &liveInst, this](const uint32_t* iid) {
```
alan-baker: op is unused.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -130,17 +142,27 @@ void AggressiveDCEPass::ComputeBlock2HeaderMaps(
...
       currentBranchInst.pop();
     }
-    block2headerMerge_[*bi] = currentMergeInst.top();
-    block2headerBranch_[*bi] = currentBranchInst.top();
     ir::Instruction* mergeInst;
     ir::Instruction* branchInst;
     uint32_t mergeBlockId;
-    if (IsStructuredIfHeader(*bi, &mergeInst, &branchInst, &mergeBlockId)) {
+    // If there is live code in loop header, the loop is live
+    if (IsStructuredHeader(*bi, SpvOpLoopMerge, &mergeInst, &branchInst,
```
alan-baker: I would consider reverting IsStructuredHeader to return true for either case and then conditionally block2HeaderMerge and block2HeaderBranch instead of invoking IsStructuredHeader twice.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -268,18 +291,33 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
...
-      AddToWorklist(block2headerMerge_[blk]);
+      ir::Instruction* mergeInst = block2headerMerge_[blk];
+      AddToWorklist(mergeInst);
+      // If in a loop, mark all branches targeting merge block
+      // and continue block as live.
+      if (mergeInst->opcode() == SpvOpLoopMerge) {
+        AddBranchesToWorklist(
+            mergeInst->GetSingleWordInOperand(kLoopMergeMergeBlockIdInIdx));
+        AddBranchesToWorklist(
+            mergeInst->GetSingleWordInOperand(kLoopMergeContinueBlockIdInIdx));
```
s-perron: I spent a lot of time looking at this code.  I believe that adding all of the branches that target the merge block is the right thing to do.  However, I don't know about the continue block.  If the continue target is the same block as the merge block of an empty structure, I think the empty structure will be kept alive.  I created an example for this to check if I am write, and ADCE generates invalid code.  The error is not new, so I opened #1075.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Yep. Gone.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: Got ride of IsStructuredHeader completely and now call IsStructuredIfOrLoopHeader once.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: OK.

source/opt/aggressive_dead_code_elim_pass.cpp
greg-lunarg: OK. Thanks. I will look at that next.

greg-lunarg:
Thanks for the reviews. I think have addressed all suggestions up to this point.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -180,19 +189,23 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
...
   // TODO(greg-lunarg): Handle Frexp, Modf more optimally
   call_in_func_ = false;
   func_is_entry_point_ = false;
   private_stores_.clear();
-  // Stacks to keep track of when we are inside an if-construct. When not
-  // immediately inside an in-construct,  we must assume all branches are live.
+  // Stacks to keep track of when we are inside an if- or loop-construct.
+  // When not immediately inside an if- or loop-construct,  we must assume
+  // all branches are live as we may be inside of a control construct (ie
+  // switch) which is not part of the ADCE analysis.
```
dnovillo: This should not be necessary, really.  I don't see why control structures should be special wrt ADCE.  You simply need to know how to remove the whole structure when you find it dead.  What am I missing?  All this tracking makes the logic more difficult to follow.

greg-lunarg: I can see how the reason for this can be unclear.

I am actually looking at this logic right now to fix an issue that @s-perron found with empty-if, so I will see if I can either improve the logic or the documentation in this area as part of that issue.

greg-lunarg:
I think the remaining appveyor issue is the infamous "effcee" problem.

Issue #1068, opened 2017-12-07
Add option in spirv-opt to skip validation.


s-perron:
There might be scenarios where it is useful to skip validation in spirv-opt.

The main one is when you have SPIR-V that is technically illegal, but in a way that is intended.  If you are not able to change the SPIR-V, then this is your only option.  For example, if you want to run the legalization passes to make illegal code from an HSLS front-end legal.

PR #1069, opened 2017-12-07
Add option to spirv-opt to skip the validator.

s-perron:
The option --skip-validation is added.  When used it will skip the
validation step.

Also rearranged the help text to fix two options that are out of order.

Implements issue #1068.

### DIFF
--- a/tools/opt/opt.cpp
+++ b/tools/opt/opt.cpp
@@ -140,6 +140,9 @@ Options (in lexicographical order):
                Exhaustively inline all function calls in entry point call tree
                functions. Currently does not inline calls to functions with
                early return in a loop.
+  --local-redundancy-elimination
+               Looks for instructions in the same basic block that compute the
+               same value, and deletes the redundant ones.
   --merge-blocks
                Join two blocks into a single block if the second has the
                first as its only predecessor. Performed only on entry point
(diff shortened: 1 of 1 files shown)


tools/opt/opt.cpp
```diff
@@ -463,22 +470,24 @@ int main(int argc, const char** argv) {
...
-  if (error) {
-    spvDiagnosticPrint(diagnostic);
+  if (!skip_validator) {
+    // Let's do validation first.
+    spv_context context = spvContextCreate(target_env);
+    spv_diagnostic diagnostic = nullptr;
+    spv_const_binary_t binary_struct = {binary.data(), binary.size()};
+    spv_result_t error =
+        spvValidateWithOptions(context, options, &binary_struct, &diagnostic);
+    if (error) {
```
atgoo: Consider
```
  if (error) {
      spvDiagnosticPrint(diagnostic);
  }
  spvDiagnosticDestroy(diagnostic);
  spvValidatorOptionsDestroy(options);
  spvContextDestroy(context) 
  return error;
```

s-perron: The way the diff looks is confusing, but we want to `return error` only if there was an error.  We would have to do

```
if (error) {
  spvDiagnosticPrint(diagnostic);
}
  spvDiagnosticDestroy(diagnostic);
  spvValidatorOptionsDestroy(options);
  spvContextDestroy(context) 
if (error) {
  return error;
}
```
I think the extra `if` is not very nice, and it only saves 1 line of code.

tools/opt/opt.cpp
```diff
@@ -208,6 +206,11 @@ Options (in lexicographical order):
                blank spaces, and in each pair, spec id and default value must
                be separated with colon ':' without any blank spaces in between.
                e.g.: --set-spec-const-default-value "1:100 2:400"
+  --skip-validation
+               Will not validate the SPIR-V before optimizing.  If the SPIR-V
+               is invalid, the optimizer may fail or generate incorrect output.
```
dnovillo: Add a warning that states that this should be used with care and only in special circumstances.

tools/opt/opt.cpp
atgoo: Sorry, I thought it was the end of the function.

Issue #1071, opened 2017-12-07
Improve TypeManager


alan-baker:
The TypeManager has good intentions, but the implementation is not very usable. Ideally, it could be used for finding the declaration of a given type, but since it does looks up based on addresses it can only find previously indexed types. This means users cannot construct a type and see if its been declared.

Another issue is that decorations are added after the types have been created. This should be changed so that decorations are indexed at the same time as types. Furthermore, there should be a clear system of describing which decorations change the meaning of a type.

The SPIRV spec says non-aggregate types cannot be decorated in ways that change the meaning of their type (the variable pointers extension modifies that to non-pointer non-aggregate types). Perhaps not all decorations should matter and be recorded in a type.

alan-baker:
Pushed as 616908503d747a052022d5f5af86647ec373855f

PR #1072, opened 2017-12-07
Add a new constant manager class.

dnovillo:
This patch adds a new constant manager class to interface with
analysis::Constant.  The new constant manager lives in ir::IRContext
together with the type manager (analysis::TypeManager).

The new analysis::ConstantManager is used by the spec constant folder
and the constant propagator (in progress).

**NOTE**:  The patch currently runs into testsuite failures due to the instruction unique IDs.  I'm investigating, but in the meantime, I would appreciate a review for the additional type and constant managers in the IRContext.

### DIFF
--- a/source/opt/constants.cpp
+++ b/source/opt/constants.cpp
@@ -22,11 +22,12 @@ namespace spvtools {
 namespace opt {
 namespace analysis {
 
-analysis::Type* ConstantManager::GetType(const ir::Instruction* inst) {
+analysis::Type* ConstantManager::GetType(const ir::Instruction* inst) const {
   return context()->get_type_mgr()->GetType(inst->type_id());
 }
 
-uint32_t ConstantManager::FindRecordedConstant(const analysis::Constant* c) {
+uint32_t ConstantManager::FindRecordedConstant(
+    const analysis::Constant* c) const {
   auto iter = const_val_to_id_.find(c);
   if (iter == const_val_to_id_.end()) {
     return 0;
@@ -36,7 +37,7 @@ uint32_t ConstantManager::FindRecordedConstant(const analysis::Constant* c) {
 }
 
 std::vector<const analysis::Constant*> ConstantManager::GetConstantsFromIds(
-    const std::vector<uint32_t>& ids) {
+    const std::vector<uint32_t>& ids) const {
   std::vector<const analysis::Constant*> constants;
   for (uint32_t id : ids) {
     if (analysis::Constant* c = FindRecordedConstant(id)) {
@@ -63,7 +64,7 @@ ir::Instruction* ConstantManager::BuildInstructionAndAddToModule(
   return new_inst_ptr;
 }
 
-analysis::Constant* ConstantManager::FindRecordedConstant(uint32_t id) {
+analysis::Constant* ConstantManager::FindRecordedConstant(uint32_t id) const {
   auto iter = id_to_const_val_.find(id);
   if (iter == id_to_const_val_.end()) {
     return nullptr;
@@ -74,7 +75,7 @@ analysis::Constant* ConstantManager::FindRecordedConstant(uint32_t id) {
 
 std::unique_ptr<analysis::Constant> ConstantManager::CreateConstant(
     const analysis::Type* type,
-    const std::vector<uint32_t>& literal_words_or_ids) {
+    const std::vector<uint32_t>& literal_words_or_ids) const {
   std::unique_ptr<analysis::Constant> new_const;
   if (literal_words_or_ids.size() == 0) {
     // Constant declared with OpConstantNull
@@ -124,7 +125,7 @@ std::unique_ptr<analysis::Constant> ConstantManager::CreateConstant(
 }
 
 std::unique_ptr<analysis::Constant> ConstantManager::CreateConstantFromInst(
-    ir::Instruction* inst) {
+    ir::Instruction* inst) const {
   std::vector<uint32_t> literal_words_or_ids;
   std::unique_ptr<analysis::Constant> new_const;
 
@@ -156,7 +157,7 @@ std::unique_ptr<analysis::Constant> ConstantManager::CreateConstantFromInst(
 }
 
 std::unique_ptr<ir::Instruction> ConstantManager::CreateInstruction(
-    uint32_t id, analysis::Constant* c) {
+    uint32_t id, analysis::Constant* c) const {
   if (c->AsNullConstant()) {
     return MakeUnique<ir::Instruction>(
         context(), SpvOp::SpvOpConstantNull,
@@ -190,7 +191,7 @@ std::unique_ptr<ir::Instruction> ConstantManager::CreateInstruction(
 }
 
 std::unique_ptr<ir::Instruction> ConstantManager::CreateCompositeInstruction(
-    uint32_t result_id, analysis::CompositeConstant* cc) {
+    uint32_t result_id, analysis::CompositeConstant* cc) const {
   std::vector<ir::Operand> operands;
   for (const analysis::Constant* component_const : cc->GetComponents()) {
     uint32_t id = FindRecordedConstant(component_const);
--- a/source/opt/constants.cpp
+++ b/source/opt/constants.cpp
@@ -1,0 +1,214 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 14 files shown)


dnovillo:
@alan-baker, @s-perron I've fixed the test failures.  The patch is now ready for a full review.  There will be some additional changes to constant.{cpp,h} for features needed by the propagator, but I will include those in the final CCP patch.

Qining:
lgtm.

source/opt/ir_context.h
```diff
@@ -204,6 +205,24 @@ class IRContext {
...
+  // created yet, it creates one.  NOTE: Once created, the constant manager
+  // remains active and it is never re-built.
+  opt::analysis::ConstantManager* get_constant_mgr() {
+    if (!constant_mgr_)
+      constant_mgr_.reset(new opt::analysis::ConstantManager(this));
+    return constant_mgr_.get();
+  }
+
+  // Returns a pointer to the type manager.  If no type manager has been created
+  // yet, it creates one. NOTE: Once created, the type manager remains active it
```
alan-baker: I don't think this is appropriate currently. The TypeManager currently cannot add types after its initial analysis. So if a new type is ever added, it will not show up in the TypeManager. I have filed #1071 to improve the TypeManager.

source/opt/ir_context.h
```diff
@@ -59,13 +61,14 @@ class IRContext {
...
-  // Create an |IRContext| that contains an owned |Module|
+  // Creates an |IRContext| that contains an owned |Module|
   IRContext(spvtools::MessageConsumer c)
       : unique_id_(0),
         module_(new Module()),
         consumer_(std::move(c)),
         def_use_mgr_(nullptr),
-        valid_analyses_(kAnalysisNone) {
+        valid_analyses_(kAnalysisNone),
+        constant_mgr_(nullptr) {
```
alan-baker: Why didn't you initialize type_mgr_ also?

source/opt/ir_context.h
```diff
@@ -74,15 +77,13 @@ class IRContext {
         module_(std::move(m)),
         consumer_(std::move(c)),
         def_use_mgr_(nullptr),
-        valid_analyses_(kAnalysisNone) {
+        valid_analyses_(kAnalysisNone),
+        constant_mgr_(nullptr) {
```
alan-baker: Same here

source/opt/constants.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+namespace spvtools {
+namespace opt {
+namespace analysis {
+
+analysis::Type* ConstantManager::GetType(const ir::Instruction* inst) {
+  return context()->get_type_mgr()->GetType(inst->type_id());
+}
+
+uint32_t ConstantManager::FindRecordedConstant(const analysis::Constant* c) {
```
alan-baker: This should be a const method.

source/opt/constants.cpp
```diff
@@ -0,0 +1,214 @@
...
+  auto new_inst = CreateInstruction(new_id, new_const);
+  if (!new_inst) return nullptr;
+  auto* new_inst_ptr = new_inst.get();
+  *pos = pos->InsertBefore(std::move(new_inst));
+  ++(*pos);
+  context()->get_def_use_mgr()->AnalyzeInstDefUse(new_inst_ptr);
+  return new_inst_ptr;
+}
+
+analysis::Constant* ConstantManager::FindRecordedConstant(uint32_t id) {
```
alan-baker: Also this one.

source/opt/constants.cpp
```diff
@@ -0,0 +1,214 @@
...
+    return MakeUnique<analysis::VectorConstant>(vt, components);
+  } else if (auto* st = type->AsStruct()) {
+    auto components = GetConstantsFromIds(literal_words_or_ids);
+    if (components.empty()) return nullptr;
+    return MakeUnique<analysis::StructConstant>(st, components);
+  } else if (auto* at = type->AsArray()) {
+    auto components = GetConstantsFromIds(literal_words_or_ids);
+    if (components.empty()) return nullptr;
+    return MakeUnique<analysis::ArrayConstant>(at, components);
+  } else {
```
alan-baker: Missing matrix handling?

source/opt/ir_context.h
dnovillo: I've made TypeManager::AnalyzeTypes a public method for now.  I want to keep this re-factor as NFC as possible.  I'd rather extend the TypeManager as part of the fix for #1071.

source/opt/constants.cpp
dnovillo: Done.

source/opt/constants.cpp
dnovillo: Done.

source/opt/constants.cpp
dnovillo: Yes, but to be done in a separate PR extending this functionality.  I want to keep this NFC.

source/opt/ir_context.h
dnovillo: Missed it.  Done.

source/opt/ir_context.h
dnovillo: Done.

source/opt/constants.cpp
```diff
@@ -0,0 +1,215 @@
...
+}
+
+std::vector<const analysis::Constant*> ConstantManager::GetConstantsFromIds(
+    const std::vector<uint32_t>& ids) const {
+  std::vector<const analysis::Constant*> constants;
+  for (uint32_t id : ids) {
+    if (analysis::Constant* c = FindRecordedConstant(id)) {
+      constants.push_back(c);
+    } else {
+      return {};
```
s-perron: Why return an empty vector if only one constant failed?

dnovillo:
Re-tested, re-based and pushed as 241dcacc04d661166c57d5ad3ec71f47a71d03e3

source/opt/constants.cpp
```diff
@@ -0,0 +1,215 @@
...
+  analysis::Constant* new_const = c.get();
+  uint32_t new_id = context()->TakeNextId();
+  const_val_to_id_[new_const] = new_id;
+  id_to_const_val_[new_id] = std::move(c);
+  auto new_inst = CreateInstruction(new_id, new_const);
+  if (!new_inst) return nullptr;
+  auto* new_inst_ptr = new_inst.get();
+  *pos = pos->InsertBefore(std::move(new_inst));
+  ++(*pos);
+  context()->get_def_use_mgr()->AnalyzeInstDefUse(new_inst_ptr);
```
s-perron: We should add something to the context to do this.  Then it can update the def use manager if it is built.  This will rebuild it even if it is not needed.  See `IRContext::AnalyzeUses` for how this could be done.

source/opt/constants.cpp
```diff
@@ -0,0 +1,215 @@
...
+    assert(literal_words_or_ids.size() == 1 &&
+           "Bool constant should be declared with one operand");
+    return MakeUnique<analysis::BoolConstant>(bt, literal_words_or_ids.front());
+  } else if (auto* it = type->AsInteger()) {
+    return MakeUnique<analysis::IntConstant>(it, literal_words_or_ids);
+  } else if (auto* ft = type->AsFloat()) {
+    return MakeUnique<analysis::FloatConstant>(ft, literal_words_or_ids);
+  } else if (auto* vt = type->AsVector()) {
+    auto components = GetConstantsFromIds(literal_words_or_ids);
+    if (components.empty()) return nullptr;
```
s-perron: I'm trying to understand the uses cases.  Why return nullptr instead of asserting?  Do we envision a situation where the caller wants a constant vector, and can do something meaningful with the nullptr?  The only things I can think of is that the caller made a mistake.  They have an id for a non-constant, or they did not register the constant.

source/opt/constants.cpp
```diff
@@ -0,0 +1,215 @@
...
+    if (!std::all_of(components.begin(), components.end(),
+                     [](const analysis::Constant* c) {
+                       if (c->type()->AsBool() || c->type()->AsInteger() ||
+                           c->type()->AsFloat()) {
+                         return true;
+                       } else {
+                         return false;
+                       }
+                     }))
+      return nullptr;
```
s-perron: Again here, the caller would be trying to create something illegal.  I feel an assert would be better.

source/opt/constants.cpp
dnovillo: This is code from the spec constant folder.  Any functionality changes I need in this code will come from CCP.

dnovillo:
Commented on one of your questions.  Any extensions to this code are better done in a new PR.  I'll add some extensions with CCP.  If you see any other extensions or modifications you may want, please file an issue and send it my way.  I'll be playing around this code in the next few days.

PR #1079, opened 2017-12-09
Export a config file for pkg-config

### LINKED ISSUE #1031: Export SPIRV-Tools as a “package”
Providing a set of pkg-config (and CMake or others) “package files” would make it easier for other projects using SPIRV-Tools to locate it and use it, rather than having to implement some custom detection method.

I have some patches for generating pkg-config’s .pc file, which I could submit if you are interested.

(I have come across this issue I am planning to use SPIRV-Tools in [Mesa](https://www.mesa3d.org/) (in its implementation of the OpenCL API), which uses autotools and meson as build systems, both of which can rely on pkg-config for checking for dependencies. I was previously looking for the header files, as well as symbols defined in the library to check whether or not it had the linker in it, for example.)

pierremoreau:
Fixes #1031

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -249,6 +249,7 @@ if(CURRENT_VERSION MATCHES "-dev")
 else()
   string(APPEND CURRENT_VERSION ".1")
 endif()
+configure_file(CHANGES CHANGES.junk)
 configure_file("./cmake/SPIRV-Tools.pc.in" "SPIRV-Tools.pc" @ONLY)
 
 # Install pkg-config file
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,11 @@ if (POLICY CMP0054)
 endif()
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
-project(spirv-tools)
+project(spirv-tools
+  VERSION
+    2017.2
+  DESCRIPTION
+    "Tools for SPIR-V")
 enable_testing()
 set(SPIRV_TOOLS "SPIRV-Tools")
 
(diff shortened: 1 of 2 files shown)


pierremoreau:
It’s my first time writing a pkg-config configuration file, so I might have made some basic errors. I can confirm that it works though, as I was able to integrate SPIRV-Tools into Mesa using the autotools primitives for looking up a pkg-config package.

If using the CMake API for setting up the version, it expects the version to be of the format MAJOR[.MINOR[.PATCH]], so it won’t accept the “-dev” suffix. Should we consider PATCH=0 to be the dev version, or not use the CMake API for setting up the version?

dneto0:
In principle this seems ok to me.

It fails the Android build.  That should be fixed.

There is extra overhead in that right now the version is set by the string in the CHANGES file.  It's one more step to update the version number in CMake.   I think I can live with that.  :-)

And we can adopt the convention that the .0 means dev, and .1 or higher is more current.   However, the code will evolve for some time before having that minor version updated.  In other words, it's a moving target while being .0

pierremoreau:
I went with using the project’s description and version variables, but nothing prevents us from creating custom ones in CMake, and that way keep the “-dev” in the version, or have them directly in the .pc file. I went with what seemed to be the more “CMake-way”, but we don’t have to take that path.

pierremoreau:
Found the issue for Android: the “DESCRIPTION” property for the `project()` command was only introduced in CMake 3.9; I did not bump the minimum required version which is currently at 2.8.12, and the Android build was using CMake 3.2.2, causing it to fail.

@dneto0 I don’t think it’s worth it to bump the CMake requirements to 3.9, but should I store the description as a variable in CMake or directly inline it into the pkg-config? This is also valid for the version of the project (see the comment above).

dneto0:
I see.
I'd like to keep the version information only taken from the CHANGEs file.  So how about a two-stage process: 
- first stage does a CMake-configure but skips setting the version.  This writes SPIRV-Tools.pc.2 (in the build output directory)
- enhance the call to update_build_version.py so has a new responsibility to copy SPIRV-Tools.pc.2 to SPIRV-Tools.pc but substitute the version number

dneto0:
I agree I don't want to increase the minimum required version of CMake.  For example, the Android NDK build would have to be updated first to make a newer CMake available (I can't remember which it uses right now.)

pierremoreau:
How about reading CHANGES from CMake, and extracting the first version we find (for example, in Bash by running `grep "^v[0-9]\+.* [0-9]\+-[0-9]\+-[0-9]\+$" CHANGES | head -1 | sed -e 's/^v\([0-9]\+.*\) [0-9]\+-[0-9]\+-[0-9]\+$/\1/'`)? That way we get everything in one pass.

pierremoreau:
Please take another look.

CMake will retrieve the first new version declaration found in CHANGES, will extract the version string, and replace “-dev” by “.0” as neither pkg-config nor CMake appreciate having a string in the version; pkg-config just strips it, so for “2017.2-dev” it actually returns “2017.2”.
Also, the description is now directly embedded in the pkg-config file, rather than first set in CMake, then inserted in the pkg-config file.

pierremoreau:
The three patches should solve the two problem you mentioned. The `configure_file()` on CHANGES is kind of a hack, but it works.

dneto0:
I prefer #1103 for various reasons. Take a look.

pierremoreau:
Superseded by #1103 .

PR #1080, opened 2017-12-11
Force gtest to expose ::testing::Combine

dneto0:
Work around faulty logic in googletest, where ::testing::Combine
is accidentally disabled for VS 2017.
See https://github.com/google/googletest/issues/1352


dneto0:
I'd like to see the bots pass first...

dneto0:
Please take another look.  I tested locally with VS 2017 Community Edition.

Issue #1083, opened 2017-12-12
Inlining does not set the parent of new basic blocks.


s-perron:
This is an internal representation issue, so I cannot give an specific test case.  However, after inlining, the basic blocks created by inlining return nullptr when GetParent is called.

PR #1084, opened 2017-12-12
Set the parent for basic blocks during inlining.

s-perron:
Inlining is not setting the parent (function) for each basic block.
This can cause problems for later optimizations.  The solution is to set
the parent for each new block just before it is linked into the
function.

Fixes issue #1083.


s-perron:
@dnovillo Added a test.  Not very complicated, but it fails without my change, passes with it.

antiagainst:
Windows failed because of re2 discards support for VS 2013. Should be fixed by #1085 .

s-perron:
Thanks Lei.  I'll merge this change then.

PR #1087, opened 2017-12-12
Fix uninitialized warning at -Os.

dnovillo:
-Os is not smart enough to realize that this variable is actually never used uninitialized.  Added an initializer to help it.


dnovillo:
Re-based and pushed to master as 853a3d6c

Issue #1075, opened 2017-12-08
ADCE deletes label, but leaves a branch to the label.


s-perron:
If an empty structured-if has a merge block that is also the continue target of a loop construct, the block in the if is removed, but the branch is left as is.  This leaves a use of a label id that has no definition.

The following SPIR-V demonstrates the problem.  Assemble it, and run it through ADCE.

```
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %2 "main" %3
               OpExecutionMode %2 OriginUpperLeft
               OpSource GLSL 430
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %2 "main"
               OpDecorate %3 Location 0
          %4 = OpTypeVoid
          %5 = OpTypeFunction %4
          %6 = OpTypeFloat 32
          %8 = OpConstant %6 0
          %9 = OpTypeInt 32 1
         %10 = OpTypePointer Function %9
         %11 = OpConstant %9 0
         %12 = OpConstant %9 10
         %13 = OpTypeBool
         %14 = OpConstant %6 1
         %15 = OpConstant %9 1
         %16 = OpTypePointer Output %6
          %3 = OpVariable %16 Output
          %2 = OpFunction %4 None %5
         %20 = OpLabel
               OpBranch %21
         %21 = OpLabel
         %22 = OpPhi %6 %8 %20 %23 %24
         %25 = OpPhi %9 %11 %20 %26 %24
               OpLoopMerge %27 %24 None
               OpBranch %28
         %28 = OpLabel
         %29 = OpSLessThan %13 %25 %12
               OpBranchConditional %29 %30 %27
         %30 = OpLabel
         %23 = OpFAdd %6 %22 %14
         %31 = OpConvertFToS %9 %23
         %32 = OpSGreaterThan %13 %25 %31
               OpSelectionMerge %24 None
               OpBranchConditional %32 %33 %24
         %33 = OpLabel
               OpBranch %24
         %24 = OpLabel
         %26 = OpIAdd %9 %25 %15
               OpBranch %21
         %27 = OpLabel
               OpStore %3 %22
               OpReturn
               OpFunctionEnd
```

greg-lunarg:
Good catch! 

[...]

PR #1088, opened 2017-12-13
ADCE: Only mark true breaks and continues of live loops

greg-lunarg:
Previously we marked *all* branches to a loop's continue and merge blocks when it was discovered to be live, but some of these branches are not true breaks or continues.

This fixes issue #1075

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -174,38 +174,37 @@ void AggressiveDCEPass::AddBranch(uint32_t labelId, ir::BasicBlock* bp) {
 }
 
 void AggressiveDCEPass::AddBreaksAndContinuesToWorklist(
-      ir::Instruction* loopMerge) {
+    ir::Instruction* loopMerge) {
   const uint32_t mergeId =
       loopMerge->GetSingleWordInOperand(kLoopMergeMergeBlockIdInIdx);
-  get_def_use_mgr()->ForEachUser(mergeId,
-                                 [&loopMerge, this](ir::Instruction* user) {
-    // A branch to the merge block can only be a break if it is nested in the
-    // current loop
-    SpvOp op = user->opcode();
-    if (op != SpvOpBranchConditional && op != SpvOpBranch) return;
-    ir::Instruction* branchInst = user;
-    while (true) {
-      ir::BasicBlock* blk = inst2block_[branchInst];
-      ir::Instruction* hdrBranch = block2headerBranch_[blk];
-      if (hdrBranch == nullptr) return;
-      ir::Instruction* hdrMerge = branch2merge_[hdrBranch];
-      if (hdrMerge == loopMerge) break;
-      branchInst = hdrBranch;
-    }
-    if (!IsLive(user)) AddToWorklist(user);
-  });
+  get_def_use_mgr()->ForEachUser(
+      mergeId, [&loopMerge, this](ir::Instruction* user) {
+        // A branch to the merge block can only be a break if it is nested in
+        // the current loop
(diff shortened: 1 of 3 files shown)


source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -146,8 +146,9 @@ void AggressiveDCEPass::ComputeBlock2HeaderMaps(
       currentMergeInst.push(mergeInst);
       currentBranchInst.push(branchInst);
     }
-    block2headerMerge_[*bi] = currentMergeInst.top();
-    block2headerBranch_[*bi] = currentBranchInst.top();
+    ir::Instruction* cbi = currentBranchInst.top();
+    block2headerBranch_[*bi] = cbi;
+    branch2merge_[cbi] = currentMergeInst.top();
```
s-perron: I'm lost at this point.  Some comments will probably be needed.  Why is this line executed for every block.  Won't it be setting the same entry to the same value multiple times?

Also, now that I look at it closer, why do we need three different stack on the go.  The current branch instruction will always be the instruction after the merge instruction.  The merge block is an operand to the merge instruction.  Can't we just keep one of them?

Then at the same time, why have branch2merge_ at all?  Can't we just step back one instruction from the branch instruction?

greg-lunarg: > Can't we just step back one instruction from the branch instruction?

Does that work? If I have Instruction* ip pointing at an instruction, can I do --ip and have it point at the preceding instruction?

greg-lunarg: Re: multiple updates: Yeah, when I changed to branch2merge_, I spaced that it didn't need to get updated for every block.

I got rid of two of the stacks. I would like to leave the branch2merge_ map because it saves me some code in a couple other places in the class.

greg-lunarg:
Hold off on merging this. I think I have found a case where incorrect code would be generated.

greg-lunarg:
I realized that a continue can take the form of a conditional branch with a merge instruction as long as the merge block is not the continue block.

greg-lunarg:
I think this can be merged now.

source/opt/aggressive_dead_code_elim_pass.cpp
s-perron: You cannot do `--inst` with an instruction pointer, but you can do `inst->NextNode()` and `inst->PreviousNode()` to the get next and previous instruction in the basic block.  They do not cross basic block boundaries.

greg-lunarg:
That’s cool. I think I need to make one more improvement anyway: similar to the last test I added but for break.

greg-lunarg:
OK. I think its ready for the second review.

dneto0:
I'm looking now.

source/opt/aggressive_dead_code_elim_pass.h
```diff
@@ -140,9 +141,8 @@ class AggressiveDCEPass : public MemPass {
   // immediate controlling structured if.
```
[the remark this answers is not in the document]
dneto0: Additionally, a loop header block will point to its own branch. But an "if" header block points to an enclosing construct's branch (if one exists).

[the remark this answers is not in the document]
dneto0: I'll make these comment fixes before pushing into master.

dneto0:
Rebased, squashed, comment updated, and pushed into master as 0f804063154

greg-lunarg:
Thanks for the review and the comment fixes!

dneto0:
You're welcome.
I'm working on an answer to "when do you need a merge instruction" in #1075....

PR #1089, opened 2017-12-13
Type manager improvements

alan-baker:
Changes for #1071.

Changes TypeManager to hash types instead of type pointers. This allows lookup of arbitrarily created types. Users still must be careful when dealing with ambiguous types (e.g. multiple declarations of the same struct). To that end -O and -Os now run duplicate removal.

In general users should use TypeManager::GetTypeInstruction to get any type they need (with the caveat about ambiguous types). If the requested type already exists, it will be returned otherwise all the necessary instructions to define the input type will be added to the module. If users call AddType, they should also register that instruction with the TypeManager to keep it up-to-date. I have not made the type manager an analysis.

Because of ambiguous types, I've modified the constant manager to be able to provide an optional type override when necessary.

### DIFF
--- a/source/opt/constants.cpp
+++ b/source/opt/constants.cpp
@@ -51,12 +51,12 @@ std::vector<const analysis::Constant*> ConstantManager::GetConstantsFromIds(
 
 ir::Instruction* ConstantManager::BuildInstructionAndAddToModule(
     std::unique_ptr<analysis::Constant> c, ir::Module::inst_iterator* pos,
-    uint32_t typeId) {
+    uint32_t type_id) {
   analysis::Constant* new_const = c.get();
   uint32_t new_id = context()->TakeNextId();
   const_val_to_id_[new_const] = new_id;
   id_to_const_val_[new_id] = std::move(c);
-  auto new_inst = CreateInstruction(new_id, new_const, typeId);
+  auto new_inst = CreateInstruction(new_id, new_const, type_id);
   if (!new_inst) return nullptr;
   auto* new_inst_ptr = new_inst.get();
   *pos = pos->InsertBefore(std::move(new_inst));
@@ -158,9 +158,9 @@ std::unique_ptr<analysis::Constant> ConstantManager::CreateConstantFromInst(
 }
 
 std::unique_ptr<ir::Instruction> ConstantManager::CreateInstruction(
-    uint32_t id, analysis::Constant* c, uint32_t typeId) const {
+    uint32_t id, analysis::Constant* c, uint32_t type_id) const {
   uint32_t type =
-      (typeId == 0) ? context()->get_type_mgr()->GetId(c->type()) : typeId;
+      (type_id == 0) ? context()->get_type_mgr()->GetId(c->type()) : type_id;
   if (c->AsNullConstant()) {
     return MakeUnique<ir::Instruction>(context(), SpvOp::SpvOpConstantNull,
                                        type, id,
@@ -183,7 +183,7 @@ std::unique_ptr<ir::Instruction> ConstantManager::CreateInstruction(
             spv_operand_type_t::SPV_OPERAND_TYPE_TYPED_LITERAL_NUMBER,
             fc->words())});
   } else if (analysis::CompositeConstant* cc = c->AsCompositeConstant()) {
-    return CreateCompositeInstruction(id, cc, typeId);
+    return CreateCompositeInstruction(id, cc, type_id);
   } else {
     return nullptr;
   }
@@ -191,7 +191,7 @@ std::unique_ptr<ir::Instruction> ConstantManager::CreateInstruction(
 
 std::unique_ptr<ir::Instruction> ConstantManager::CreateCompositeInstruction(
     uint32_t result_id, analysis::CompositeConstant* cc,
-    uint32_t typeId) const {
+    uint32_t type_id) const {
   std::vector<ir::Operand> operands;
   for (const analysis::Constant* component_const : cc->GetComponents()) {
     uint32_t id = FindRecordedConstant(component_const);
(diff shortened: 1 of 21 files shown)


source/opt/types.h
```diff
@@ -171,32 +247,45 @@ class Matrix : public Type {
...
   std::string str() const override;
 
   Image* AsImage() override { return this; }
   const Image* AsImage() const override { return this; }
 
+  const Type* sampled_type() const { return sampled_type_; }
+  SpvDim dim() const { return dim_; }
+  uint32_t depth() const { return depth_; }
+  uint32_t arrayed() const { return arrayed_; }
+  uint32_t ms() const { return ms_; }
```
s-perron: This acronym is does not have an obvious meaning.  Is it possible to replace with this IsMultiSampled.  Also, shouldn't it return a bool since the value is either 0 or 1?  Same for arrayed.

source/opt/types.h
```diff
@@ -171,32 +247,45 @@ class Matrix : public Type {
...
+  uint32_t sampled() const { return sampled_; }
```
s-perron: I don't know what is best here, but this can have values 0, 1, or 2.  Do we want this to have an enum?  Code that uses it will be much more readable.

source/opt/constants.cpp
```diff
@@ -50,12 +50,13 @@ std::vector<const analysis::Constant*> ConstantManager::GetConstantsFromIds(
 }
 
 ir::Instruction* ConstantManager::BuildInstructionAndAddToModule(
-    std::unique_ptr<analysis::Constant> c, ir::Module::inst_iterator* pos) {
+    std::unique_ptr<analysis::Constant> c, ir::Module::inst_iterator* pos,
+    uint32_t typeId) {
```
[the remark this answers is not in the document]
alan-baker: Will change

source/opt/types.h
alan-baker: I'll make them bool and name them is_arrayed and is_multisampled.

source/opt/types.h
alan-baker: I'm going to leave this as is for now at least.

dnovillo:
Looking at this now.

source/opt/ir_context.cpp
```diff
@@ -95,6 +95,10 @@ Instruction* IRContext::KillInst(ir::Instruction* inst) {
     }
   }
 
+  if (type_mgr_) {
+    type_mgr_->RemoveId(inst->result_id());
```
s-perron: We will be killing a to of instructions that are not type.  Since the type manager is always valid this will mean a lot of hash table look ups.  Would if be better to have a quick check of the opcode?  If it is not OpType* then don't try to remove it.

source/opt/type_manager.h
```diff
@@ -62,16 +86,45 @@ class TypeManager {
...
+
+  // Registers |id| to |type|.
+  //
+  // If GetId(|type|) already returns a non-zero id, the return value will be
+  // unchanged.
+  void RegisterType(uint32_t id, const Type& type);
+
+  // Removes knowledge of |id| from the manager.
+  //
+  // If |id| is an ambiguous type, GetId for |id|'s type may return a new id.
```
s-perron: Comment looks out of place.

source/opt/fold_spec_constant_op_and_composite_pass.cpp
```diff
@@ -160,26 +160,54 @@ bool FoldSpecConstantOpAndCompositePass::ProcessOpSpecConstantOp(
   return true;
 }
 
+uint32_t FoldSpecConstantOpAndCompositePass::GetTypeComponent(
+    uint32_t typeId, uint32_t element) const {
```
dnovillo: Seems like a useful function for the type manager.

source/opt/inline_pass.cpp
```diff
@@ -33,6 +33,14 @@ namespace opt {
 
 uint32_t InlinePass::FindPointerToType(uint32_t type_id,
                                        SpvStorageClass storage_class) {
+  opt::analysis::Type* pointeeTy = context()->get_type_mgr()->GetType(type_id);
```
dnovillo: just analysis::Type.  It happens in other spots too.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -335,8 +335,18 @@ uint32_t ScalarReplacementPass::GetOrCreatePointerType(uint32_t id) {
   auto iter = pointee_to_pointer_.find(id);
   if (iter != pointee_to_pointer_.end()) return iter->second;
 
-  // TODO(alanbaker): Make the type manager useful and then replace this code.
+  opt::analysis::Type* pointeeTy = context()->get_type_mgr()->GetType(id);
+  opt::analysis::Pointer pointerTy(pointeeTy, SpvStorageClassFunction);
```
dnovillo: Could this pattern be encoded in the type manager as a single function call?  It happens in a couple of places.  It could return a std::tie<pointeeTy, pointerTy> or an std::pair.  Not a biggie, though.

source/opt/constants.h
```diff
@@ -325,22 +325,39 @@ class ConstantManager {
   // points to the same instruction before and after the insertion. This is the
   // only method that actually manages id creation/assignment and instruction
   // creation/insertion for a new Constant instance.
+  //
+  // |type_id| is an optional argument for disambiguating equivalent types. If
+  // |type_id| is specified, it is used as the type of the constant. Otherwise
+  // the type of the constant is derived by getting an id from the type manager
+  // for |c|.
```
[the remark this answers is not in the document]
alan-baker: I needed to do this because the TypeManager doesn't have a reliable id mapping for ambiguous types. Users of the constant manager do. This is a drawback to be able to search for id-less types.

source/opt/inline_pass.cpp
alan-baker: Will change

source/opt/fold_spec_constant_op_and_composite_pass.cpp
alan-baker: I'll add an utility function to Instruction.

source/opt/scalar_replacement_pass.cpp
alan-baker: Ok

alan-baker:
Rebased, squashed and pushed as 616908503d747a052022d5f5af86647ec373855f

PR #1090, opened 2017-12-13
Refactor spir-v headers includes

atgoo:
Moved spirv, glsl-std-450, opencl-std header includes into one place to avoid headers of different version being used across the codebase, unless intended.


source/latest_version_spirv_header.h
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_LATEST_VERSION_SPIRV_HEADER_H_
+#define LIBSPIRV_LATEST_VERSION_SPIRV_HEADER_H_
+
+#include "spirv/1.2/spirv.h"
+
+inline const char* GetSpirvHeaderVersionString() {
+  return "1.2";
```
dneto0: This function is not used.

source/latest_version_opencl_std_header.h
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_LATEST_VERSION_OPENCL_STD_HEADER_H_
+#define LIBSPIRV_LATEST_VERSION_OPENCL_STD_HEADER_H_
+
+#include "spirv/1.2/OpenCL.std.h"
+
+inline const char* GetOpenClStdHeaderVersionString() {
```
dneto0: This function is not used.

source/latest_version_glsl_std_450_header.h
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_LATEST_VERSION_GLSL_STD_450_HEADER_H_
+#define LIBSPIRV_LATEST_VERSION_GLSL_STD_450_HEADER_H_
+
+#include "spirv/1.2/GLSL.std.450.h"
+
+inline const char* GetGlslStd450HeaderVersionString() {
```
dneto0: This function is not used.
It also pollutes the global namespace.

dneto0: GLSL.std.450.h declares a GLSLstd450Version = 100;   GLSLstd450Revision=3 that can serve this same purpose.

source/latest_version_opencl_std_header.h
dneto0: The OpenCL header does *not* have a self-reported version number.

source/latest_version_spirv_header.h
dneto0: spirv.h has a self-reported version number via SPV_VERSION and SPV_REVISION macros, and static variables SpvVersion and SpvRevision.

dneto0:
Rebased, squashed, and pushed into master as af7d5799a5c

PR #1094, opened 2017-12-14
Allow pointers to pointers in logical addressing mode.

s-perron:
A few optimizations are updates to handle code that is suppose to be
using the logical addressing mode, but still has variables that contain
pointers as long as the pointer are to opaque objects.  This is called
"relaxed logical addressing".  It applies only to spirv-opt.  spirv-val will
not use the relaxed rules.

|Instruction::GetBaseAddress| will check that pointers that are use meet
the relaxed logical addressing rules.  Optimization that now handle
relaxed logical addressing instead of logical addressing are:

 - aggressive dead-code elimination
 - local access chain convert
 - local store elimination passes.

Some of the work needed for #1096.

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -417,3 +417,23 @@ bool spvOpcodeIsBlockTerminator(SpvOp opcode) {
   return spvOpcodeIsBranch(opcode) || spvOpcodeIsReturn(opcode) ||
          opcode == SpvOpKill || opcode == SpvOpUnreachable;
 }
+
+bool spvOpcodeIsBaseOpaqueType(SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpTypeImage:
+    case SpvOpTypeSampler:
+    case SpvOpTypeSampledImage:
+    case SpvOpTypeOpaque:
+    case SpvOpTypeEvent:
+    case SpvOpTypeDeviceEvent:
+    case SpvOpTypeReserveId:
+    case SpvOpTypeQueue:
+    case SpvOpTypePipe:
+    case SpvOpTypeForwardPointer:
+    case SpvOpTypePipeStorage:
+    case SpvOpTypeNamedBarrier:
+      return true;
+    default:
+      return false;
+  }
+}
--- a/source/opcode.h
+++ b/source/opcode.h
@@ -106,4 +106,6 @@ bool spvOpcodeIsReturn(SpvOp opcode);
 // Returns true if the given opcode is a basic block terminator.
 bool spvOpcodeIsBlockTerminator(SpvOp opcode);
 
+// Returns true if the given opcode always defines an opaque type.
+bool spvOpcodeIsBaseOpaqueType(SpvOp opcode);
 #endif  // LIBSPIRV_OPCODE_H_
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -36,6 +36,7 @@ const uint32_t kLoopMergeContinueBlockIdInIdx = 1;
 }  // namespace
 
 bool AggressiveDCEPass::IsVarOfStorage(uint32_t varId, uint32_t storageClass) {
+  if (varId == 0) return false;
   const ir::Instruction* varInst = get_def_use_mgr()->GetDef(varId);
   const SpvOp op = varInst->opcode();
   if (op != SpvOpVariable) return false;
@@ -355,7 +356,9 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
     if (liveInst->opcode() == SpvOpLoad) {
       uint32_t varId;
       (void)GetPtr(liveInst, &varId);
-      ProcessLoad(varId);
+      if (varId != 0) {
+        ProcessLoad(varId);
+      }
     }
     // If function call, treat as if it loads from all pointer arguments
     else if (liveInst->opcode() == SpvOpFunctionCall) {
@@ -466,7 +469,8 @@ Pass::Status AggressiveDCEPass::ProcessImpl() {
   // TODO(greg-lunarg): Handle additional capabilities
   if (!get_module()->HasCapability(SpvCapabilityShader))
     return Status::SuccessWithoutChange;
-  // Current functionality assumes logical addressing only
+  // Current functionality assumes relaxed logical addressing (see
+  // instruction.h)
   // TODO(greg-lunarg): Handle non-logical addressing
   if (get_module()->HasCapability(SpvCapabilityAddresses))
     return Status::SuccessWithoutChange;
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -156,8 +156,11 @@ bool Instruction::IsReadOnlyLoad() const {
 }
 
 Instruction* Instruction::GetBaseAddress() const {
-  assert(IsLoad() &&
-         "GetBaseAddress should only be called on load instructions.");
+  assert((IsLoad() || opcode() == SpvOpStore || opcode() == SpvOpAccessChain ||
+          opcode() == SpvOpInBoundsAccessChain ||
+          opcode() == SpvOpCopyObject) &&
+         "GetBaseAddress should only be called on instructions that take a "
+         "pointer or image.");
   uint32_t base = GetSingleWordInOperand(kLoadBaseIndex);
   ir::Instruction* base_inst = context()->get_def_use_mgr()->GetDef(base);
   bool done = false;
@@ -179,6 +182,9 @@ Instruction* Instruction::GetBaseAddress() const {
         break;
     }
   }
+
+  assert(base_inst->IsValidBasePointer() &&
+         "We cannot have a base pointer come from this load");
   return base_inst;
 }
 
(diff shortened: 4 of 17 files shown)


s-perron:
@greg-lunarg Greg can you take a look at this as well to make sure it is reasonable?

source/opcode.cpp
```diff
@@ -417,3 +417,22 @@ bool spvOpcodeIsBlockTerminator(SpvOp opcode) {
...
+    case SpvOpTypeImage:
+    case SpvOpTypeSampler:
+    case SpvOpTypeSampledImage:
+    case SpvOpTypeOpaque:
+    case SpvOpTypeEvent:
+    case SpvOpTypeDeviceEvent:
+    case SpvOpTypeReserveId:
+    case SpvOpTypeQueue:
+    case SpvOpTypePipe:
+    case SpvOpTypeForwardPointer:
```
dneto0: I expected that OpTypeForwardPointer should be the same as OpTypePointer, and should not be on this list.  But I see the SPIR-V spec says it's an opaque type.
But OpTypeForwardPointer doesn't even have a result id....   :-)

I guess this is fine.  You can't do better.

source/opcode.h
```diff
@@ -106,4 +106,6 @@ bool spvOpcodeIsReturn(SpvOp opcode);
 // Returns true if the given opcode is a basic block terminator.
 bool spvOpcodeIsBlockTerminator(SpvOp opcode);
 
+// Returns true if the given opcode is a basic block terminator.
```
dneto0: copypasta

source/opt/instruction.h
```diff
@@ -35,6 +35,19 @@ class IRContext;
...
 
+// Relaxed logcial addressing:
+//
+// In the logical addressing model, pointers cannot be stored or loaded.  This
+// is a useful assumption because it simplifies the aliasing significantly.
+// However, for the purpose of legalizing HLSL type code, we will have to allow
+// storing and loading of pointers to opaque objects and runtime arrays.  This
+// relaxation of the rule still implies that function and private scope
+// variables do not have any aliasing, so we can treat them as before.  This
+// will be cased the relaxed logical addressing model.
```
dneto0: "will be cased" ---> "will be the case in"

greg-lunarg:
OK

source/opt/instruction.h
```diff
@@ -35,6 +35,19 @@ class IRContext;
 class Module;
 class InstructionList;
 
+// Relaxed logcial addressing:
+//
+// In the logical addressing model, pointers cannot be stored or loaded.  This
+// is a useful assumption because it simplifies the aliasing significantly.
+// However, for the purpose of legalizing HLSL type code, we will have to allow
```
dneto0: Suggest "legalizing HLSL type code" --> "legalizing code generated from HLSL"

source/opt/mem_pass.cpp
```diff
@@ -170,6 +179,11 @@ bool MemPass::IsLiveStore(ir::Instruction* storeInst) {
   // get store's variable
   uint32_t varId;
   (void)GetPtr(storeInst, &varId);
+  if (varId == 0) {
+    // If we do know now which variable we are accessing, assume the store is
```
dneto0: If we do *not* know which variable .... ?

source/opcode.cpp
```diff
@@ -417,3 +417,22 @@ bool spvOpcodeIsBlockTerminator(SpvOp opcode) {
   return spvOpcodeIsBranch(opcode) || spvOpcodeIsReturn(opcode) ||
          opcode == SpvOpKill || opcode == SpvOpUnreachable;
 }
+bool spvOpcodeIsOpaqueType(SpvOp opcode) {
```
alan-baker: According to the spec an opaque-ness is transitive. Any type containing or pointing to or containing a pointer to an opaque type is also opaque. Should this function instead take an instruction and recurse through its types?

source/opt/instruction.cpp
```diff
@@ -369,5 +374,32 @@ Instruction* Instruction::InsertBefore(std::unique_ptr<Instruction>&& i) {
   i.get()->InsertBefore(this);
   return i.release();
 }
+
+bool Instruction::IsValidBasePointer(Instruction* inst) const {
+  if (context()->module()->HasCapability(SpvCapabilityAddresses)) {
```
dneto0: This is really permissive.   Suggest updating the method comment saying we assume inst is an instruction that yields a pointer.

Or move move the check for pointer type up above this line.

source/opt/mem_pass.cpp
alan-baker: do -> don't

source/opt/instruction.h
```diff
@@ -328,6 +342,10 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   bool IsReadOnlyVariableShaders() const;
   bool IsReadOnlyVariableKernel() const;
 
+  // Returns true if it is valid to use the result if |inst| as the base
+  // pointer for a load or store.
```
dneto0: Should say that this will return true if using physical addressing or if relaxed-logical addressing rules are satisfied.

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -36,6 +36,7 @@ const uint32_t kLoopMergeContinueBlockIdInIdx = 1;
 }  // namespace
 
 bool AggressiveDCEPass::IsVarOfStorage(uint32_t varId, uint32_t storageClass) {
+  if (varId == 0) return false;
```
dneto0: Seems to merit an update to the method comment.

source/opt/mem_pass.cpp
```diff
@@ -92,21 +97,25 @@ bool MemPass::IsPtr(uint32_t ptrId) {
...
-      *varId = varInst->GetSingleWordInOperand(kCopyObjectOperandInIdx);
-    }
-    varInst = get_def_use_mgr()->GetDef(*varId);
+  if (varInst->opcode() == SpvOpVariable) {
+    *varId = varInst->result_id();
+  } else {
+    *varId = 0;
+  }
+
+  while (ptrInst->opcode() == SpvOpCopyObject) {
```
dneto0: At first I was concerned that functionality might have been changed by moving this down. But the GetBasePointer logic traces through OpCopyObject.   And I think the rest checks out.

So this is fine by me.

source/opt/instruction.cpp
```diff
@@ -156,7 +156,9 @@ bool Instruction::IsReadOnlyLoad() const {
 }
 
 Instruction* Instruction::GetBaseAddress() const {
-  assert(IsLoad() &&
+  assert((IsLoad() || opcode() == SpvOpStore || opcode() == SpvOpAccessChain ||
+          opcode() == SpvOpInBoundsAccessChain ||
+          opcode() == SpvOpCopyObject) &&
          "GetBaseAddress should only be called on load instructions.");
```
dneto0: The assert text is now out of date.

dneto0:
Please update the commit message to say this is only affecting the optimizer.  Don't leave the impression that the validator is affected.

source/opcode.cpp
s-perron: I'll rename this function to `spvOpcodeIsBaseOpaqueType`.  It cannot take an instruction because the file is shared with the validator, and instructions are represented differently.

I think I will need something more generic for the check that currently calls this.

s-perron:
Looking at a larger set of shaders exposed some extra cases I need to handle.

source/opcode.cpp
s-perron: I've add `IsOpaqueType` to the instruction class.  That will look re

source/opt/mem_pass.cpp
s-perron: fixed.

source/opcode.cpp
s-perron: Okay, I'll leave it as is.

source/opcode.h
s-perron: Fixed.

source/opt/aggressive_dead_code_elim_pass.cpp
s-perron: Done.

source/opt/instruction.cpp
s-perron: Fixed.

source/opt/instruction.cpp
s-perron: I've moved the check for pointer type.  I've also the check for the opcodes of non variables.  For those I don't case so much about the opcodes.  I care about the type.

source/opt/instruction.h
s-perron: I meant "will be called".  Changed to that.

source/opt/instruction.h
s-perron: done.

source/opt/mem_pass.cpp
s-perron: done.

s-perron:
Made most of the changes.  Still need to add test.

greg-lunarg:
Did I miss a discussion of this somewhere? Its a little hard to evaluate something when you don't understand how it is going to be used.

greg-lunarg:
The title of the PR should say "...pointers to opaques...", right?

greg-lunarg:
So is the plan to use pointers to opaque types in structs, etc?

If so, are the IHVs going to support that?

Or is the optimizer going to propagate these pointers? If so, how is this different than just propagating opaque typed objects themselves?

s-perron:
@greg-lunarg I'll create an issue where I explain the issue.  It is not really about all opaque object, but with structured buffers.

s-perron:
@greg-lunarg I created issue #1096 that explains the motivation.  I am being more general than needed.  I can pull back a bit if we really need to.

greg-lunarg:
Great! Thanks! I will look at it right now.

greg-lunarg:
This all looks fine.

Yes, I see that the current code could be open to propagating pointers to more than uniforms & buffers, but I agree its probably not going to be a problem since those exceptions are not really valid in shaders anyway.

s-perron:
I believe all of the comments have been addressed.  Please take another look.

source/opt/instruction.cpp
```diff
@@ -389,5 +395,49 @@ Instruction* Instruction::InsertBefore(std::unique_ptr<Instruction>&& i) {
...
+      context()->get_def_use_mgr()->GetDef(pointee_type_id);
+
+  if (pointee_type_inst->IsOpaqueType()) {
+    return true;
+  }
+  return false;
+}
+
+bool Instruction::IsOpaqueType() const {
+  if (opcode() == SpvOpTypeStruct) {
```
alan-baker: I think you need to check the element type of arrays too. You could have an array of opaque types I believe. I see no restriction in the spec.

source/opt/instruction.h
```diff
@@ -35,6 +35,19 @@ class IRContext;
 class Module;
 class InstructionList;
 
+// Relaxed logcial addressing:
+//
+// In the logical addressing model, pointers cannot be stored or loaded.  This
```
dnovillo: copied?  'store' and 'load' have the immediate implication that the pointer is being de-referenced.

source/opt/instruction.h
```diff
@@ -35,6 +35,19 @@ class IRContext;
...
+// is a useful assumption because it simplifies the aliasing significantly.
+// However, for the purpose of legalizing code generated from HLSL, we will have
+// to allow storing and loading of pointers to opaque objects and runtime
```
dnovillo: Same here.

source/opt/instruction.h
```diff
@@ -312,6 +326,11 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
     return spvOpcodeIsBlockTerminator(opcode());
   }
 
+  // Return true if |this| is an instruction that define an opaque type.  Since
+  // runtime array are similar characteristics they are included as opaque
```
dnovillo: s/are/have/

source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -466,7 +469,8 @@ Pass::Status AggressiveDCEPass::ProcessImpl() {
   // TODO(greg-lunarg): Handle additional capabilities
   if (!get_module()->HasCapability(SpvCapabilityShader))
     return Status::SuccessWithoutChange;
-  // Current functionality assumes logical addressing only
```
dnovillo: So, there's no actual way of testing for relaxed logical addressing?

s-perron: There is not quick check that can be done at the start of the pass.  However, it is check and enforced through the assert that are done in the `GetPtr` calls.

source/opt/instruction.h
s-perron: I do mean "store" and "load".  The value of the the pointer cannot be stored into a variable because we are not allowed to have variables of type pointer to pointer.

source/opt/instruction.h
s-perron: Same as above.

PR #1095, opened 2017-12-14
Add primitive instruction validation pass

jeremy-lunarg:


dneto0:
I reformatted the one line, then committed to master as cdfbf26c13

source/validate_primitives.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "validate.h"
+
+#include "diagnostic.h"
+#include "opcode.h"
+#include "val/instruction.h"
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+// Validates correctness of composite instructions.
```
atgoo: Not composite.

source/validate_primitives.cpp
```diff
@@ -0,0 +1,58 @@
...
+spv_result_t PrimitivesPass(ValidationState_t& _,
+                            const spv_parsed_instruction_t* inst) {
+  const SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+
+  switch (opcode) {
+    case SpvOpEmitStreamVertex:
+    case SpvOpEndStreamPrimitive: {
+      const uint32_t stream_type = _.GetOperandTypeId(inst, 0);
```
atgoo: There is also function _.GetTypeId(id). (since you already have variable stream_id).

PR #1097, opened 2017-12-15
Test validation success for OpEmitVertex OpEndPrimitive, execution model for primitive instructions

dneto0:


dneto0:
Followup to #1095 
cc: @jeremy-lunarg

dneto0:
I filed #1098 and #1099 for more validation that can be done for primitive instructions.

dneto0:
Fixed formatting.

test/val/val_primitives_test.cpp
```diff
@@ -28,13 +28,14 @@ using ValidatePrimitives = spvtest::ValidateBase<bool>;
 
 std::string GenerateShaderCode(
     const std::string& body,
+    const bool declare_geometry_streams_capability = true,
     const std::string& capabilities_and_extensions = "",
     const std::string& execution_model = "Geometry") {
```
atgoo: It also should validate and test that it doesn't work when execution model is not Geometry.

Validation can be done calling this function:
    _.current_function().RegisterExecutionModelLimitation(
        SpvExecutionModelGeometry,
        "Primitive instructions require Geometry execution model");
when a primitive instruction is encountered.

test/val/val_primitives_test.cpp
```diff
@@ -28,13 +28,14 @@ using ValidatePrimitives = spvtest::ValidateBase<bool>;
 
 std::string GenerateShaderCode(
     const std::string& body,
+    const bool declare_geometry_streams_capability = true,
```
atgoo: This can be passed through capabilities_and_extensions.

test/val/val_primitives_test.cpp
atgoo: (this can be done in the next pull request)

jeremy-lunarg: `_.current_function().RegisterExecutionModelLimitation(
SpvExecutionModelGeometry,
"Primitive instructions require Geometry execution model");`

@atgoo 
I tried to use this naively and failed on `assert(in_function_body())` within `current_function()`.

test/val/val_primitives_test.cpp
dneto0: I've reworked it to use that parameter instead.  Good suggestion.

test/val/val_primitives_test.cpp
dneto0: I've done this in the current pull request, in the last (new) commit.

dneto0:
All feedback addressed.  Please take another look.

test/val/val_primitives_test.cpp
```diff
@@ -63,6 +63,27 @@ OpFunctionEnd)";
   return ss.str();
 }
 
+// Returns SPIR-V assembly fragment representing a function call,
+// the end of the callee body, and the preamble and body of the called
+// function with the given body, but missing the final return and
+// function-end.  The result is of the form where it can be used in the
+// |body| argument to GenerateShaderCode.
+std::string CallAndCallee(std::string body) {
```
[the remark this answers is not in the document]
dneto0: As I recall the compiler complained when this formal parameter was a const reference.   So I made this pretty dumb by just making non-ref string.  It's dumb and it works.  Clarity is the most important thing in test code, especially at this level of detail.

Thanks for the review.

dneto0:
Rebased, squashed, and merged as  8135dd637

Fixed #1098

Issue #889, opened 2017-10-19
Implement Sparse Conditional Constant Propagation (SSA-CCP)


dnovillo:
This will implement the algorithm proposed in https://dl.acm.org/citation.cfm?id=103136.  This implementation has several pre-requisites:

- [x] Phi placement.  Although there is an implementation of Phi placement in local_ssa_elim_pass.cpp, it assumes structured control flow.  A more generalized form of Phi placement (i.e., pruned SSA form) may be required.
- [x] CFG edges.  Needed for the simulation done during constant propagation.
- [x] Propagation engine.  The propagation logic for SSA-CCP can be shared by other transformations (value range propagation, attribute and copy propagation).
- [x] Constant folding.  Some of the logic needed to fold integer constants exists in opt/fold.{cpp,h}.  More logic will need to be added to fold floating point.
- [ ] Control Flow cleanups.  In particular, straightening of conditional branches.  Constant propagation can convert predicates for conditional jumps into compile-time constant values 'true' or 'false'.  This allows CFG cleanups to completely remove the conditional branch.

I will be creating new issues for these pre-requisites and link them to this issue.

dnovillo:
Fixed with 4ba9dcc8.  The remaining issue is being tracked in #1124.

dneto0:
Reopening since the last step of CFG cleanups is still pending.

dneto0:
Never mind. I see #1124 is open for the last bit

PR #1102, opened 2017-12-15
Implement SSA CCP (SSA Conditional Constant Propagation)

dnovillo:
This implements the conditional constant propagation pass proposed in

    Constant propagation with conditional branches,
    Wegman and Zadeck, ACM TOPLAS 13(2):181-210.

The implementation uses the SSA propagation engine added in #985 and is the final fix needed for #889.

The implementation required several changes in the constant manager.  I believe I should separate those changes into a separate PR.  I'll see about doing this early next week.

The test cases cover all the possible CFG situations the propagator can find itself in.  CCP has the potential of leaving unreachable regions in the CFG (when it simplifies conditionals) and dead code (when it propagates an assignment).

I've scheduled CCP in -O and -Os so that DCE and CFGCleanup run after it.  We'll need to adjust this as we test with real code.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -452,6 +452,18 @@ Optimizer::PassToken CreateScalarReplacementPass();
 // used in only one function.  Those variables are moved to the function storage
 // class in the function that they are used.
 Optimizer::PassToken CreatePrivateToLocalPass();
+
+// Creates a conditional constant propagation (CCP) pass.
+// This pass implements the SSA-CCP algorithm in
+//
+//      Constant propagation with conditional branches,
+//      Wegman and Zadeck, ACM TOPLAS 13(2):181-210.
+//
+// Constant values in expressions and conditional jumps are folded and
+// simplified. This may reduce code size by removing never executed jump targets
+// and computations with constant operands.
+Optimizer::PassToken CreateCCPPass();
+
 }  // namespace spvtools
 
 #endif  // SPIRV_TOOLS_OPTIMIZER_HPP_
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -16,6 +16,7 @@ add_library(SPIRV-Tools-opt
   basic_block.h
   block_merge_pass.h
   build_module.h
+  ccp_pass.h
   cfg_cleanup_pass.h
   cfg.h
   common_uniform_elim_pass.h
@@ -74,6 +75,7 @@ add_library(SPIRV-Tools-opt
   basic_block.cpp
   block_merge_pass.cpp
   build_module.cpp
+  ccp_pass.cpp
   cfg_cleanup_pass.cpp
   cfg.cpp
   common_uniform_elim_pass.cpp
--- a/source/opt/ccp_pass.cpp
+++ b/source/opt/ccp_pass.cpp
@@ -1,0 +1,264 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file implements conditional constant propagation as described in
+//
+//      Constant propagation with conditional branches,
+//      Wegman and Zadeck, ACM TOPLAS 13(2):181-210.
+#include "ccp_pass.h"
+
+#include "fold.h"
+#include "function.h"
+#include "module.h"
+#include "propagator.h"
+
+namespace spvtools {
+namespace opt {
+
+SSAPropagator::PropStatus CCPPass::VisitPhi(ir::Instruction* phi) {
+  uint32_t meet_val_id = 0;
+
+  // Implement the lattice meet operation. The result of this Phi instruction is
+  // interesting only if the meet operation over arguments coming through
+  // executable edges yields the same constant value.
+  for (uint32_t i = 2; i < phi->NumOperands(); i += 2) {
+    if (!propagator_->IsPhiArgExecutable(phi, i)) {
+      // Ignore arguments coming through non-executable edges.
+      continue;
+    }
+    uint32_t phi_arg_id = phi->GetSingleWordOperand(i);
+    auto it = values_.find(phi_arg_id);
+    if (it != values_.end()) {
+      // We found an argument with a constant value.  Apply the meet operation
+      // with the previous arguments.
+      if (meet_val_id == 0) {
+        // This is the first argument we find.  Initialize the result to its
+        // constant value id.
+        meet_val_id = it->second;
+      } else if (it->second == meet_val_id) {
+        // The argument is the same constant value already computed. Continue
+        // looking.
+        continue;
+      } else {
+        // We found another constant value, but it is different from the
+        // previous computed meet value.  This Phi will never be constant.
+        return SSAPropagator::kVarying;
+      }
+    } else {
+      // If any argument is not a constant, the Phi produces nothing
+      // interesting for now. The propagator will callback again, if needed.
+      return SSAPropagator::kNotInteresting;
+    }
+  }
+
+  // If there are no incoming executable edges, the meet ID will still be 0. In
+  // that case, return not interesting to evaluate the Phi node again.
+  if (meet_val_id == 0) {
+    return SSAPropagator::kNotInteresting;
+  }
+
+  // All the operands have the same constant value represented by |meet_val_id|.
+  // Set the Phi's result to that value and declare it interesting.
+  values_[phi->result_id()] = meet_val_id;
+  return SSAPropagator::kInteresting;
+}
+
+SSAPropagator::PropStatus CCPPass::VisitAssignment(ir::Instruction* instr) {
+  assert(instr->result_id() != 0 &&
+         "Expecting an instruction that produces a result");
+
+  // If this is a copy operation, and the RHS is a known constant, assign its
+  // value to the LHS.
+  if (instr->opcode() == SpvOpCopyObject) {
+    uint32_t rhs_id = instr->GetSingleWordInOperand(0);
+    auto it = values_.find(rhs_id);
+    if (it != values_.end()) {
+      values_[instr->result_id()] = it->second;
+      return SSAPropagator::kInteresting;
+    }
+    return SSAPropagator::kNotInteresting;
+  }
+
+  // Instructions with a RHS that cannot produce a constant are always varying.
+  if (!instr->IsFoldable()) {
+    return SSAPropagator::kVarying;
+  }
+
+  // Otherwise, see if the RHS of the assignment folds into a constant value.
+  std::vector<uint32_t> cst_val_ids;
+  for (uint32_t i = 0; i < instr->NumInOperands(); i++) {
+    uint32_t op_id = instr->GetSingleWordInOperand(i);
+    auto it = values_.find(op_id);
+    if (it != values_.end()) {
+      cst_val_ids.push_back(it->second);
+    } else {
+      break;
+    }
+  }
+
+  // If we did not find a constant value for every operand in the instruction,
+  // do not bother folding it.  Indicate that this instruction does not produce
+  // an interesting value for now.
+  auto constants = const_mgr_->GetConstantsFromIds(cst_val_ids);
+  if (constants.size() == 0) {
+    return SSAPropagator::kNotInteresting;
+  }
+
+  // Otherwise, fold the instruction with all the operands to produce a new
+  // constant.
+  uint32_t result_val = FoldScalars(instr->opcode(), constants);
+  const analysis::Constant* result_const =
+      const_mgr_->GetConstant(const_mgr_->GetType(instr), {result_val});
+  ir::Instruction* const_decl =
+      const_mgr_->GetDefiningInstruction(result_const);
+  values_[instr->result_id()] = const_decl->result_id();
+  return SSAPropagator::kInteresting;
+}
+
+SSAPropagator::PropStatus CCPPass::VisitBranch(ir::Instruction* instr,
+                                               ir::BasicBlock** dest_bb) const {
+  assert(instr->IsBranch() && "Expected a branch instruction.");
+  uint32_t dest_label = 0;
+
+  if (instr->opcode() == SpvOpBranch) {
+    // An unconditional jump always goes to its unique destination.
+    dest_label = instr->GetSingleWordInOperand(0);
+  } else if (instr->opcode() == SpvOpBranchConditional) {
+    // For a conditional branch, determine whether the predicate selector has a
+    // known value in |values_|.  If it does, set the destination block
+    // according to the selector's boolean value.
+    uint32_t pred_id = instr->GetSingleWordOperand(0);
+    auto it = values_.find(pred_id);
+    if (it == values_.end()) {
+      // The predicate has an unknown value, either branch could be taken.
+      *dest_bb = nullptr;
+      return SSAPropagator::kVarying;
+    }
+
+    // Get the constant value for the predicate selector from the value table.
+    // Use it to decide which branch will be taken.
+    uint32_t pred_val_id = it->second;
+    const analysis::Constant* c = const_mgr_->FindDeclaredConstant(pred_val_id);
+    assert(c && "Expected to find a constant declaration for a known value.");
+    const analysis::BoolConstant* val = c->AsBoolConstant();
+    dest_label = val->value() ? instr->GetSingleWordOperand(1)
+                              : instr->GetSingleWordOperand(2);
+  } else {
+    // For an OpSwitch, extract the value taken by the switch selector and check
+    // which of the target literals it matches.  The branch associated with that
+    // literal is the taken branch.
+    assert(instr->opcode() == SpvOpSwitch);
+    uint32_t select_id = instr->GetSingleWordOperand(0);
+    auto it = values_.find(select_id);
+    if (it == values_.end()) {
+      // The selector has an unknown value, any of the branches could be taken.
+      *dest_bb = nullptr;
+      return SSAPropagator::kVarying;
+    }
+
+    // Get the constant value for the selector from the value table. Use it to
+    // decide which branch will be taken.
+    uint32_t select_val_id = it->second;
+    const analysis::Constant* c =
+        const_mgr_->FindDeclaredConstant(select_val_id);
+    assert(c && "Expected to find a constant declaration for a known value.");
+    const analysis::IntConstant* val = c->AsIntConstant();
+
+    // Start assuming that the selector will take the default value;
+    dest_label = instr->GetSingleWordOperand(1);
+    for (uint32_t i = 2; i < instr->NumOperands(); i += 2) {
+      if (val->words()[0] == instr->GetSingleWordOperand(i)) {
+        dest_label = instr->GetSingleWordOperand(i + 1);
+        break;
+      }
+    }
+  }
+
+  assert(dest_label && "Destination label should be set at this point.");
+  *dest_bb = context()->cfg()->block(dest_label);
+  return SSAPropagator::kInteresting;
+}
+
+SSAPropagator::PropStatus CCPPass::VisitInstruction(ir::Instruction* instr,
+                                                    ir::BasicBlock** dest_bb) {
+  *dest_bb = nullptr;
+  if (instr->opcode() == SpvOpPhi) {
+    return VisitPhi(instr);
+  } else if (instr->IsBranch()) {
+    return VisitBranch(instr, dest_bb);
+  } else if (instr->result_id()) {
+    return VisitAssignment(instr);
+  }
+  return SSAPropagator::kVarying;
+}
+
+bool CCPPass::ReplaceValues() {
+  bool retval = false;
+  for (const auto& it : values_) {
+    uint32_t id = it.first;
+    uint32_t cst_id = it.second;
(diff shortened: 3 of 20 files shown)


source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,267 @@
...
+  return SSAPropagator::kVarying;
+}
+
+bool CCPPass::ReplaceValues() {
+  bool retval = false;
+  for (const auto& it : values_) {
+    uint32_t id = it.first;
+    uint32_t cst_id = it.second;
+    if (id != cst_id) {
+      context()->get_def_use_mgr()->ForEachUser(
```
alan-baker: Why isn't this just IRContext::ReplaceAllUsesWith? Separately DefUseManager::ForEachUse will go through all the operands so you don't have to write the nested loop

source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,267 @@
...
+  InitializeProcessing(c);
+
+  const_mgr_ = context()->get_constant_mgr();
+
+  // Populate the constant table with values from constant declarations in the
+  // module.  The values of each OpConstant declaration is the identity
+  // assignment (i.e., each constant is its own value).
+  for (const auto& inst : c->module()->GetConstants()) {
+    values_[inst->result_id()] = inst->result_id();
+    const_mgr_->MapConstantToInst(const_mgr_->GetConstantFromInst(inst), inst);
```
alan-baker: Add ConstantManager::MapInst and merge these functions together?

source/opt/constants.cpp
```diff
@@ -50,50 +40,52 @@ std::vector<const analysis::Constant*> ConstantManager::GetConstantsFromIds(
...
   }
 }
 
-std::unique_ptr<analysis::Constant> ConstantManager::CreateConstant(
-    const analysis::Type* type,
-    const std::vector<uint32_t>& literal_words_or_ids) const {
-  std::unique_ptr<analysis::Constant> new_const;
+const Constant* ConstantManager::CreateConstant(
+    const Type* type, const std::vector<uint32_t>& literal_words_or_ids) const {
+  std::unique_ptr<Constant> new_const;
```
alan-baker: Unused variable

test/opt/ccp_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+          %7 = OpCopyObject %int %9
+
+               OpBranch %25
+         %23 = OpLabel
+          %8 = OpCopyObject %int %int_4
+               OpBranch %25
+         %25 = OpLabel
+
+; %int_4 should have propagated to both OpPhi operands.
+; CHECK: OpPhi %int %int_4 [[id:%\d+]] %int_4 [[id:%\d+]]
```
alan-baker: If you want to capture the math, you can use {{<regex>}} for plain regexes.

test/opt/ccp_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+; CHECK: OpStore %outparm %int_7
+               OpStore %outparm %35
+               OpReturn
+               OpFunctionEnd
+               )";
+
+  SinglePassRunAndMatch<opt::CCPPass>(spv_asm, true);
+}
+
+TEST_F(CCPTest, SimplifySwitches) {
```
alan-baker: You should have another test for taking the default value of the switch.

source/opt/ccp_pass.cpp
dnovillo: Thanks.  I don't know what I was thinking.  Fixed.

source/opt/ccp_pass.cpp
dnovillo: MapInst?

source/opt/constants.cpp
dnovillo: Fixed. Odd that the compiler didn't get it.

test/opt/ccp_test.cpp
dnovillo: Done.

test/opt/ccp_test.cpp
dnovillo: Done.

source/opt/ccp_pass.cpp
alan-baker: I suggest merging "const_mgr_->MapConstantToInst(const_mgr_->GetConstantFromInst(inst), inst)" together into a single new function maybe called MapInst. Since the input since to rely on other constant manager information it might be convenient to provide the single function too.

include/spirv-tools/optimizer.hpp
```diff
@@ -442,6 +442,18 @@ Optimizer::PassToken CreateRedundancyEliminationPass();
...
+
+// Creates a conditional constant propagation (CCP) pass.
+// This pass implements the SSA-CCP algorithm in
+//
+//      Constant propagation with conditional branches,
+//      Wegman and Zadeck, ACM TOPLAS 13(2):181-210.
+//
+// Constant values in expressions and conditional jumps are folded and
+// simplified. This may reduce code size by removing never executed jump targets
+// and computations with constant operands.
```
s-perron: Does this pass remove the jumps or is that left for cfg cleanup?

source/opt/CMakeLists.txt
```diff
@@ -73,6 +74,7 @@ add_library(SPIRV-Tools-opt
   basic_block.cpp
   block_merge_pass.cpp
   build_module.cpp
+  ccp_pass.cpp
```
s-perron: Need to add this to the android.mk file as well.  I don't see that.

source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+    } else {
+      break;
+    }
+  }
+
+  // If we did not find a constant value for every operand in the instruction,
+  // do not bother folding it.  Indicate that this instruction does not produce
+  // an interesting value for now.  TODO(dnovillo): Some instructions will never
+  // produce an interesting constant value (e.g., most function calls).  Return
+  // kVarying for those.
```
s-perron: Instructions that are not "Combinators" should be kVarying.  See `IsCombinatorInstruction` in the IRContext.

source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+      return SSAPropagator::kInteresting;
+    }
+    return SSAPropagator::kNotInteresting;
+  }
+
+  // Otherwise, see if the RHS of the assignment folds into a constant value.
+  std::vector<uint32_t> cst_val_ids;
+  for (uint32_t i = 0; i < instr->NumInOperands(); i++) {
+    uint32_t op_id = instr->GetSingleWordInOperand(i);
+    auto it = values_.find(op_id);
```
s-perron: Not all operands are ids.  Some are literals that represent other things.

source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+    if (it != values_.end()) {
+      cst_val_ids.push_back(it->second);
+    } else {
+      break;
```
s-perron: How do you tell if you exited early?  Does it matter?  My guess is that this is an id that is not constant, so you do care.

source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+  // produce an interesting constant value (e.g., most function calls).  Return
+  // kVarying for those.
+  auto constants = const_mgr_->GetConstantsFromIds(cst_val_ids);
+  if (constants.size() == 0) {
+    return SSAPropagator::kNotInteresting;
+  }
+
+  // Otherwise, fold the instruction with all the operands to produce a new
+  // constant.
+  uint32_t result_val = FoldScalars(instr->opcode(), constants);
```
s-perron: Are you sure you can successfully fold all instructions?  What happens if you cannot?  Shouldn't it become kVarying?

source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+  values_[instr->result_id()] = const_decl->result_id();
+  return SSAPropagator::kInteresting;
+}
+
+SSAPropagator::PropStatus CCPPass::VisitBranch(ir::Instruction* instr,
+                                               ir::BasicBlock** dest_bb) const {
+  assert(instr->IsBranch() && "Expected a branch instruction.");
+  uint32_t dest_label = 0;
+
+  if (instr->opcode() == SpvOpBranch) {
```
s-perron: It is only three cases, but would this be cleaner as a switch?  Also, we could code this more defensively with a default case that assert for unknown branch instructions.  If a new branch is added with `IsBranch` changing, it will let people know they have to add a case.

source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+  return retval;
+}
+
+bool CCPPass::PropagateConstants(ir::Function* fp) {
+  const auto visit_fn = [this](ir::Instruction* instr,
+                               ir::BasicBlock** dest_bb) {
+    return VisitInstruction(instr, dest_bb);
+  };
+
+  InsertPhiInstructions(fp);
```
s-perron: Do we really want this to be part of CCP, and not its own pass?  If another pass already did this, do we really want to do it again?

source/opt/ccp_pass.h
```diff
@@ -0,0 +1,86 @@
...
+ private:
+  // Initializes the pass.
+  void Initialize(ir::IRContext* c);
+
+  // Runs constant propagation on the given function |fp|. Returns true if any
+  // constants were propagated and the IR modified.
+  bool PropagateConstants(ir::Function* fp);
+
+  // Visits a single instruction |instr|.  If the instruction is a conditional
+  // branch with that always jumps to the same basic block, it sets the
```
s-perron: Should be "branch that always jumps"

source/opt/ccp_pass.h
```diff
@@ -0,0 +1,86 @@
...
+
+  // Visits an SSA assignment instruction |instr|.  If the RHS of |instr| folds
+  // into a constant value C, then the LHS of |instr| is assigned the value C in
+  // |values_|.
+  SSAPropagator::PropStatus VisitAssignment(ir::Instruction* instr);
+
+  // Visits a branch instruction |instr|. If the branch is conditional
+  // (OpBranchConditional or OpSwitch), and the value of its selector is known,
+  // |dest_bb| will be set to the corresponding destination block.
+  SSAPropagator::PropStatus VisitBranch(ir::Instruction* instr,
```
s-perron: |dest_bb| is also set if the branch in an unconditional branch.

source/opt/ccp_pass.cpp
dnovillo: Well, there's not much to switch on.   The branches are protected against unknown branches in VisitBranch and the third case is not about specific opcodes.

source/opt/ccp_pass.h
dnovillo: Fixed.

source/opt/ccp_pass.h
dnovillo: Fixed.

source/opt/propagator.h
```diff
@@ -187,10 +187,14 @@ class SSAPropagator {
   SSAPropagator(ir::IRContext* context, const VisitFunction& visit_fn)
       : ctx_(context), visit_fn_(visit_fn) {}
 
-  // Run the propagator on function |fn|. Returns true if changes were made to
+  // Runs the propagator on function |fn|. Returns true if changes were made to
   // the function. Otherwise, it returns false.
   bool Run(ir::Function* fn);
 
+  // Returns true if the |i|th argument for |phi| comes through a CFG edge that
+  // has been marked executable.
```
dneto0: MIght be worth saying the numbering for |i| is 0-based and includes result id and type id ??
Or just say you're going to use it with GetSingleWordOperand

source/opt/ccp_pass.cpp
dneto0: Still seems borken to me.  I recommend using Instruction::ForEachInId

source/opt/fold.h
```diff
@@ -25,11 +25,13 @@ namespace spvtools {
 namespace opt {
 
 uint32_t FoldScalars(SpvOp opcode,
-                     const std::vector<analysis::Constant*>& operands);
+                     const std::vector<const analysis::Constant*>& operands);
```
dneto0: This needs a method comment out here in the header.    I don't want to have to read the code to know what's going on.

source/opt/fold.h
```diff
@@ -25,11 +25,13 @@ namespace spvtools {
...
 
 std::vector<uint32_t> FoldVectors(
     SpvOp opcode, uint32_t num_dims,
-    const std::vector<analysis::Constant*>& operands);
+    const std::vector<const analysis::Constant*>& operands);
+
+bool IsFoldableOpcode(SpvOp opcode);
```
dneto0: Needs a method comment.

source/opt/ccp_pass.cpp
dneto0: I agree with @s-perron here.  If you find an id that is non-constant, then you should bail early here rather than relying on FoldScalars later on to try to fail.  

I **could** see an instruction being foldable even with a non-constant Id.   (e.g. %mul = OpIMul %uint %x %uint_0) however, in that case how does FoldScalars line up the constant-values with the operands.  The Id that is non-constant is missing from the vector and the folding call is nonsensical.

source/opt/ccp_pass.cpp
dneto0: Also worth saying you only handle scalars with bit patterns of 32-bits.

source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,264 @@
...
+      const_mgr_->GetDefiningInstruction(result_const);
+  values_[instr->result_id()] = const_decl->result_id();
+  return SSAPropagator::kInteresting;
+}
+
+SSAPropagator::PropStatus CCPPass::VisitBranch(ir::Instruction* instr,
+                                               ir::BasicBlock** dest_bb) const {
+  assert(instr->IsBranch() && "Expected a branch instruction.");
+  uint32_t dest_label = 0;
+
```
dneto0: Optional:    *dest_bb =nullptr right at the start.  It's defensive.

source/opt/ccp_pass.cpp
```diff
@@ -0,0 +1,264 @@
...
+    uint32_t select_val_id = it->second;
+    const analysis::Constant* c =
+        const_mgr_->FindDeclaredConstant(select_val_id);
+    assert(c && "Expected to find a constant declaration for a known value.");
+    const analysis::IntConstant* val = c->AsIntConstant();
+
+    // Start assuming that the selector will take the default value;
+    dest_label = instr->GetSingleWordOperand(1);
+    for (uint32_t i = 2; i < instr->NumOperands(); i += 2) {
+      if (val->words()[0] == instr->GetSingleWordOperand(i)) {
```
dneto0: This doesn't work with OpSwitch over 64 bit values.  You must handle or bail on that case.

dnovillo:
Gah. Sorry @dneto0.  I pushed too early.  4ba9dcc8.  I'll address David's comments and fix in a new PR.

dneto0:
I filed #1130 for things I'd like fixed.

dnovillo:
On Thu, Dec 21, 2017 at 2:41 PM, David Neto <notifications@github.com>
wrote:

> *@dneto0* requested changes on this pull request.
>
> I'm still reviewing....
>
> In source/opt/propagator.h
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1102#discussion_r158356568>
> :
>
> >    // the function. Otherwise, it returns false.
>    bool Run(ir::Function* fn);
>
> +  // Returns true if the |i|th argument for |phi| comes through a CFG edge that
> +  // has been marked executable.
>
> MIght be worth saying the numbering for |i| is 0-based and includes result
> id and type id ??
> Or just say you're going to use it with GetSingleWordOperand
>

Done.


>
> In source/opt/ccp_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1102#discussion_r158358188>
> :
>
> > +  // value to the LHS.
> +  if (instr->opcode() == SpvOpCopyObject) {
> +    uint32_t rhs_id = instr->GetSingleWordInOperand(0);
> +    auto it = values_.find(rhs_id);
> +    if (it != values_.end()) {
> +      values_[instr->result_id()] = it->second;
> +      return SSAPropagator::kInteresting;
> +    }
> +    return SSAPropagator::kNotInteresting;
> +  }
> +
> +  // Otherwise, see if the RHS of the assignment folds into a constant value.
> +  std::vector<uint32_t> cst_val_ids;
> +  for (uint32_t i = 0; i < instr->NumInOperands(); i++) {
> +    uint32_t op_id = instr->GetSingleWordInOperand(i);
> +    auto it = values_.find(op_id);
>
> Still seems borken to me. I recommend using Instruction::ForEachInId
>

Done.



>
> In source/opt/fold.h
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1102#discussion_r158359022>
> :
>
> > @@ -25,11 +25,13 @@ namespace spvtools {
>  namespace opt {
>
>  uint32_t FoldScalars(SpvOp opcode,
> -                     const std::vector<analysis::Constant*>& operands);
> +                     const std::vector<const analysis::Constant*>& operands);
>
> This needs a method comment out here in the header. I don't want to have
> to read the code to know what's going on.
>

Done.


>
> In source/opt/fold.h
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1102#discussion_r158359039>
> :
>
> >
>  std::vector<uint32_t> FoldVectors(
>      SpvOp opcode, uint32_t num_dims,
> -    const std::vector<analysis::Constant*>& operands);
> +    const std::vector<const analysis::Constant*>& operands);
> +
> +bool IsFoldableOpcode(SpvOp opcode);
>
> Needs a method comment.
>

Done.



>
> In source/opt/ccp_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1102#discussion_r158360474>
> :
>
> > +    if (it != values_.end()) {
> +      values_[instr->result_id()] = it->second;
> +      return SSAPropagator::kInteresting;
> +    }
> +    return SSAPropagator::kNotInteresting;
> +  }
> +
> +  // Otherwise, see if the RHS of the assignment folds into a constant value.
> +  std::vector<uint32_t> cst_val_ids;
> +  for (uint32_t i = 0; i < instr->NumInOperands(); i++) {
> +    uint32_t op_id = instr->GetSingleWordInOperand(i);
> +    auto it = values_.find(op_id);
> +    if (it != values_.end()) {
> +      cst_val_ids.push_back(it->second);
> +    } else {
> +      break;
>
> I agree with @s-perron <https://github.com/s-perron> here. If you find an
> id that is non-constant, then you should bail early here rather than
> relying on FoldScalars later on to try to fail.
>
> I *could* see an instruction being foldable even with a non-constant Id.
> (e.g. %mul = OpIMul %uint %x %uint_0) however, in that case how does
> FoldScalars line up the constant-values with the operands. The Id that is
> non-constant is missing from the vector and the folding call is nonsensical.
>

We do bail early.  If we cannot obtain a Constant instance out of ALL the
operands in the instruction, the |constants| vector down below will be
empty.  I've now added an IsFoldableConstant() method in the folder.  This
way, we can return varying if any constants are unsupported.


>
> In source/opt/ccp_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1102#discussion_r158360885>
> :
>
> > +    }
> +  }
> +
> +  // If we did not find a constant value for every operand in the instruction,
> +  // do not bother folding it.  Indicate that this instruction does not produce
> +  // an interesting value for now.  TODO(dnovillo): Some instructions will never
> +  // produce an interesting constant value (e.g., most function calls).  Return
> +  // kVarying for those.
> +  auto constants = const_mgr_->GetConstantsFromIds(cst_val_ids);
> +  if (constants.size() == 0) {
> +    return SSAPropagator::kNotInteresting;
> +  }
> +
> +  // Otherwise, fold the instruction with all the operands to produce a new
> +  // constant.
> +  uint32_t result_val = FoldScalars(instr->opcode(), constants);
>
> Also worth saying you only handle scalars with bit patterns of 32-bits.
>
Done.  New IsFoldableConstant() helper in fold.cpp


>
> In source/opt/ccp_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1102#discussion_r158361174>
> :
>
> > +  // Otherwise, fold the instruction with all the operands to produce a new
> +  // constant.
> +  uint32_t result_val = FoldScalars(instr->opcode(), constants);
> +  const analysis::Constant* result_const =
> +      const_mgr_->GetConstant(const_mgr_->GetType(instr), {result_val});
> +  ir::Instruction* const_decl =
> +      const_mgr_->GetDefiningInstruction(result_const);
> +  values_[instr->result_id()] = const_decl->result_id();
> +  return SSAPropagator::kInteresting;
> +}
> +
> +SSAPropagator::PropStatus CCPPass::VisitBranch(ir::Instruction* instr,
> +                                               ir::BasicBlock** dest_bb) const {
> +  assert(instr->IsBranch() && "Expected a branch instruction.");
> +  uint32_t dest_label = 0;
> +
>
> Optional: *dest_bb =nullptr right at the start. It's defensive.
>

It's set in every path, and the caller, but I floated it above for clarity.


>
> In source/opt/ccp_pass.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1102#discussion_r158361677>
> :
>
> > +      *dest_bb = nullptr;
> +      return SSAPropagator::kVarying;
> +    }
> +
> +    // Get the constant value for the selector from the value table. Use it to
> +    // decide which branch will be taken.
> +    uint32_t select_val_id = it->second;
> +    const analysis::Constant* c =
> +        const_mgr_->FindDeclaredConstant(select_val_id);
> +    assert(c && "Expected to find a constant declaration for a known value.");
> +    const analysis::IntConstant* val = c->AsIntConstant();
> +
> +    // Start assuming that the selector will take the default value;
> +    dest_label = instr->GetSingleWordOperand(1);
> +    for (uint32_t i = 2; i < instr->NumOperands(); i += 2) {
> +      if (val->words()[0] == instr->GetSingleWordOperand(i)) {
>
> This doesn't work with OpSwitch over 64 bit values. You must handle or
> bail on that case.
>

Done.

PR #1103, opened 2017-12-15
Export as package

dneto0:
Based on #1079, but uses a new spirv-tools-pkg-config target to ensure precise dependencies.


dneto0:
Hey @pierremoreau I prefer this way of doing it.  It pulls some of the details out, makes a target that's easier to invoke, but most importantly has precise dependency tracking.

pierremoreau:
Oh, I didn’t know you could invoke a particular CMake file like that, really neat!
And I also didn’t think of always adding “.1” to the version, and removing it along the “-dev” in case of a development version. 👍 

LGTM

dneto0:
Thanks for the review.
I'll merge this one in once the bots pass.

dneto0:
Rebased, squashed, and pushed into master as  8c0501259

pierremoreau:
Thank you for improving the series and merging it. :-)

PR #1104, opened 2017-12-15
Remove concept of FIRST_CONCRETE_* operand types

dneto0:
Cleans up the header, and encapsulates the concept of a concrete operand type in a function to be called rather than with enum range tricks.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -96,11 +96,8 @@ typedef enum spv_operand_type_t {
   // A sentinel value.
   SPV_OPERAND_TYPE_NONE = 0,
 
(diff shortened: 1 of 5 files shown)


antiagainst:
Do we intend to do something similar for `FIRST_VARIABLE`/`FIRST_OPTIONAL`?

dneto0:
> Do we intend to do something similar for FIRST_VARIABLE/FIRST_OPTIONAL? 

I didn't have a need for that one yet. I did the concrete enum change as a first step to another set of changes I have waiting to go out.

dneto0:
Rebased and  pushed into master as 0dbe184d320

Issue #1086, opened 2017-12-12
Convert private variables to function scope


s-perron:
Many optimizations work with function scope variables, but not private scope.  The reason is the locality: if a variable is used only in one function it can be replaced by the result ids.

The current optimization scheme usually involves aggressive inlining that leaves a single function.  In this case, private variables are essentially function scope.  It would be nice to have an optimization that is able to convert private variables that are referenced in a single function into a function scope variable in that function.

PR #1105, opened 2017-12-18
Convert private variables to function scope.

s-perron:
When a private variable is used in a single function, it can be
converted to a function scope variable in that function.  This adds a
pass that does that.  The pass can be enabled using the option
`--private-to-local`.

This transformation allows other transformations to act on these
variables.

Also moved `FindPointerToType` from the inline class to the type manager.

Fixes issue #1086.

### DIFF
--- a/source/opt/private_to_local_pass.cpp
+++ b/source/opt/private_to_local_pass.cpp
@@ -1,0 +1,181 @@
+// Copyright (c) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "private_to_local_pass.h"
+
+#include "ir_context.h"
+
+namespace {
+const uint32_t kVariableStorageClassInIdx = 0;
+const uint32_t kSpvTypePointerTypeIdInIdx = 1;
+}  // namespace
+
+namespace spvtools {
+namespace opt {
+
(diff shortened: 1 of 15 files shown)


source/opt/private_to_local_pass.cpp
```diff
@@ -0,0 +1,175 @@
...
+const int kSpvTypePointerTypeIdInIdx = 1;
+}  // namespace
+
+namespace spvtools {
+namespace opt {
+
+Pass::Status PrivateToLocalPass::Process(ir::IRContext* c) {
+  InitializeProcessing(c);
+  bool modified = false;
+
```
alan-baker: The semantics on the lifetime of a private scope variable outside of shaders models are unclear. You should predicate this pass on the shader capabality.

s-perron: From the spec, I think that private variable only exist in shader modules.  I'll add the check, so we don't do pointless work.

source/opt/private_to_local_pass.cpp
```diff
@@ -0,0 +1,180 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "private_to_local_pass.h"
+
+#include "ir_context.h"
+
+namespace {
+const uint32_t kVariableStorageClassInIdx = 0;
+const int kSpvTypePointerTypeIdInIdx = 1;
```
dnovillo: uint32_t?

source/opt/private_to_local_pass.cpp
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  modified = !variables_to_move.empty();
+  for (auto p : variables_to_move) {
+    MoveVariable(p.first, p.second);
+  }
+
+  return (modified ? Status::SuccessWithChange : Status::SuccessWithoutChange);
+}
+ir::Function* PrivateToLocalPass::FindLocalFunction(
```
dnovillo: Blank line above for readability.

source/opt/private_to_local_pass.cpp
s-perron: done

source/opt/private_to_local_pass.cpp
s-perron: done

PR #1107, opened 2017-12-18
Fix/improve primitive validation

jeremy-lunarg:


dneto0:
Rebased, squashed, and pushed into master as 0d8ea486527

PR #1109, opened 2017-12-18
Opt: Remove commented out duplicated type_id() function

pierremoreau:
This code was wrongly added by #693.


dneto0:
Rebased and pushed onto master as f359635

Issue #1108, opened 2017-12-18
Remove duplicates incorrectly removes non-duplicate decorations


alan-baker:
RemoveDuplicatesPass attempts to remove duplicate decorations by asking AreDecorationsTheSame(). DecorationManager::AreDecorationsTheSame() does not check the target operands (target id for OpDecorate and type and element for OpMemberDecorate).

pierremoreau:
I’ll send a pull request shortly which reworks `DecorationManager::AreDecorationsTheSame()` and fixes this bug, as well as another one, and adds tests for remove duplicates and parts of the decoration manager.

alan-baker:
@pierremoreau I don't think DecorationManager is incorrect. I plan to fix how RemoveDuplicatesPass determines which decorations are duplicates. I'm currently working on this. I forgot to move it to in progress.

pierremoreau:
This should do it, I think: https://github.com/pierremoreau/SPIRV-Tools/commit/92364300ed07ac224976999ea99c7bfb5752b15b . I have more patches on https://github.com/pierremoreau/SPIRV-Tools/tree/decoration_manager_fixes and am currently adding test for RemoveDuplicatesPass.

pierremoreau:
I submitted a pull request with my work. It contains another fix to AreDecorationsTheSame() as well as unit tests for that function and HaveTheSameDecorations(). I also added unit tests for the various part of RemoveDuplicatesPass. If you think anything is missing from it, feel free to add/comment  on it.

alan-baker:
Fixed in 424f744db16879c09ce9c839de4954eca7580565

PR #1110, opened 2017-12-18
Decoration manager fixes and adding unit tests

pierremoreau:
* Restrict `DecorationManager::AreDecorationsTheSame()` to `SpvOpDecorate`, `SpvOpMemberDecorate` and `SpvOpDecorateId`, optionally allow it to compare the targets of the decorations, and improve the documentation of the function.
* Remove some unused code in `RemoveDuplicatesPass::RemoveDuplicateDecorations()`.
* Add tests for the `DecorationManager` and `RemoveDuplicatesPass`.

Fixes #1033 and #1108

### DIFF
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -50,7 +50,7 @@ bool DecorationManager::HaveTheSameDecorations(uint32_t id1,
   for (const ir::Instruction* inst1 : decorationsFor1) {
     bool didFindAMatch = false;
     for (const ir::Instruction* inst2 : decorationsFor2) {
-      if (AreDecorationsTheSame(inst1, inst2)) {
+      if (AreDecorationsTheSame(inst1, inst2, true)) {
         didFindAMatch = true;
         break;
       }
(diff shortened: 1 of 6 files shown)


pierremoreau:
I’ll have another look at it tomorrow, but it should be ready for reviews.

I couldn’t find an easy way to generate a single instruction how of a string, so I went with creating the instruction by hand in “test/opt/decoration_manager_test.cpp”, which isn’t really easy to read (nor write).

dneto0:
The core logic fix is good. 
@alan-baker agrees and plans to make a new PR based on this one to patch the string handling in the test code to make GCC happy.

pierremoreau:
OK. Regarding the generation of a single (or handful) `ir::Instruction` from a string, do you think the parser could provide a function to do that (and reuse the existing logic there) or ti should be hacked up whenever needed (as only rarely needed)?

dneto0:
> OK. Regarding the generation of a single (or handful) ir::Instruction from a string, do you think the parser could provide a function to do that (and reuse the existing logic there) or ti should be hacked up whenever needed (as only rarely needed)?

I think what you have now is acceptable, but I agree it will be cumbersome to continue writing test code that way.  I suggest having th following function available to test code:

// Assembles the given string into a module in the given context, and returns the resulting list of instructions.  
// Numeric Ids are preserved during assembly.
std::vector<spvtools::ir::Instruction*> Assemble(spvootls::ir::IRContext* context, std::string assembly);

What do you think?

pierremoreau:
That sounds good! I’m not sure how useful providing the context would be in those cases, but it seems better to have the caller manage it, rather than creating it inside the function and hope for the caller to release it correctly.

alan-baker:
Rebased, squashed and pushed as 424f744db16879c09ce9c839de4954eca7580565

PR #1113, opened 2017-12-19
Decoration manager fixes and adding unit tests

alan-baker:
PR from @pierremoreau (#1110) with compiler fixes.

* Restrict `DecorationManager::AreDecorationsTheSame()` to `SpvOpDecorate`, `SpvOpMemberDecorate` and `SpvOpDecorateId`, optionally allow it to compare the targets of the decorations, and improve the documentation of the function.
* Remove some unused code in `RemoveDuplicatesPass::RemoveDuplicateDecorations()`.
* Add tests for the `DecorationManager` and `RemoveDuplicatesPass`.

Fixes #1033 and #1108

### DIFF
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -50,7 +50,7 @@ bool DecorationManager::HaveTheSameDecorations(uint32_t id1,
   for (const ir::Instruction* inst1 : decorationsFor1) {
     bool didFindAMatch = false;
     for (const ir::Instruction* inst2 : decorationsFor2) {
-      if (AreDecorationsTheSame(inst1, inst2)) {
+      if (AreDecorationsTheSame(inst1, inst2, true)) {
         didFindAMatch = true;
         break;
       }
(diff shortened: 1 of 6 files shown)


source/opt/decoration_manager.cpp
```diff
@@ -60,39 +60,27 @@ bool DecorationManager::HaveTheSameDecorations(uint32_t id1,
...
-  //  decorateIdToDecorate(deco2) : deco2;
-  //
-  if (inst1->opcode() == SpvOpDecorateId || inst2->opcode() == SpvOpDecorateId)
-    return false;
+// TODO(pierremoreau): If OpDecorateId is referencing an OpConstant, one could
+//                     check that the constants are the same rather than just
+//                     looking at the constant ID.
+bool DecorationManager::AreDecorationsTheSame(const ir::Instruction* inst1,
+                                              const ir::Instruction* inst2,
+                                              bool ignore_target) const {
```
dnovillo: Would it be useful to default ignore_target to 'true'?  Not a big fan of default values, but it seems to be mostly used with this value.

dneto0: Even better in my mind is to avoid a bool param by using two different method names that more explicitly call out the difference.  I find that an accumulation of bool params tends to become hard-to-use since you always have to go back and read the comment to remember what sense the bool has.

But I haven't thought enough to invent good method names for the two cases.  Normally I think of a "decoration" as the the attributes being applied, not including the target.  But here we're using "decoration" to mean "decoration instruction".  So maybe 
SameDecorationValueAndTarget vs. SameDecorationValue  are the two cases?

alan-baker: I will leave it for another PR

alan-baker:
Rebased, squashed and pushed as 424f744db16879c09ce9c839de4954eca7580565

PR #1114, opened 2017-12-19
Make a string parameter const ref

dneto0:
Addresses feedback from #1097


dneto0:
Rebased and pushed onto master as  c9a881ecc6

Issue #1111, opened 2017-12-19
TypeManager can reference deleted memory


alan-baker:
The id to type map stores unique pointers of types. These pointers may be used in other types. If the TypeManager removes a mapping for a particular id, the memory for the type is free'd. Instead the type manager should have a pool of allocated types and store regular pointer in the id to type map.

alan-baker:
Rebased, squashed and pushed as 1ab8ad654a550d28593b1b38ac303adc66b1a241

PR #1117, opened 2017-12-19
Fix type manager memory management

alan-baker:
Addresses #1111.

Instead of mapping ids to unique_ptrs, the type manager now stores a pool of unique type pointers. The id to type map now maps to regular type pointers.

I made some modifications to RemoveDuplicatesPass to remove some (effectively) duplicated code comparing types. I also eliminated an instance of instruction being copied into a vector instead of instruction pointers.

I also made remove duplicates run-able from the command line.

### DIFF
--- a/source/opt/type_manager.cpp
+++ b/source/opt/type_manager.cpp
@@ -40,7 +40,7 @@ TypeManager::TypeManager(const MessageConsumer& consumer,
 
 Type* TypeManager::GetType(uint32_t id) const {
   auto iter = id_to_type_.find(id);
-  if (iter != id_to_type_.end()) return (*iter).second.get();
+  if (iter != id_to_type_.end()) return (*iter).second;
   return nullptr;
 }
 
@@ -67,7 +67,6 @@ ForwardPointer* TypeManager::GetForwardPointer(uint32_t index) const {
 
 void TypeManager::AnalyzeTypes(const spvtools::ir::Module& module) {
   for (const auto* inst : module.GetTypes()) RecordIfTypeDefinition(*inst);
-  for (const auto& inst : module.annotations()) AttachIfTypeDecoration(inst);
 }
 
 void TypeManager::RemoveId(uint32_t id) {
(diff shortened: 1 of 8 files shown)


alan-baker:
@pierremoreau please take a look at the changes to remove duplicates and the linker.

source/opt/type_manager.cpp
```diff
@@ -81,17 +80,17 @@ void TypeManager::RemoveId(uint32_t id) {
...
-        type_to_id_.erase(type.get());
-        type_to_id_[pair.second.get()] = pair.first;
+        type_to_id_.erase(type);
+        type_to_id_[pair.second] = pair.first;
         found = true;
         break;
       }
       // No equivalent ambiguous type, remove mapping.
-      if (!found) type_to_id_.erase(type.get());
+      if (!found) type_to_id_.erase(type);
```
dneto0: Looks like this should be outside the for loop at line 80.  Otherwise it's going to indiscriminantly delete types. ?!

source/opt/type_manager.cpp
```diff
@@ -347,10 +346,10 @@ void TypeManager::CreateDecoration(uint32_t target,
 }
 
 void TypeManager::RegisterType(uint32_t id, const Type& type) {
-  auto& t = id_to_type_[id];
-  t.reset(type.Clone().release());
-  if (GetId(t.get()) == 0) {
-    type_to_id_[t.get()] = id;
+  auto pair = type_pool_.insert(type.Clone());
```
dneto0: I think this needs a comment.
The comparison and hash on the  type pool will avoid inserting new clones of types that already match something already there.  In case of an attempted repeated insertion the clone will be destroyed by the end of this function.  That is, repeated insertions of the same Type will at most keep one corresponding object in the type pool..

source/opt/type_manager.h
```diff
@@ -120,6 +133,9 @@ class TypeManager {
  private:
```
dneto0: RegisterType comment needs clarification:

// If GetId(|type|) already returns a non-zero id, the return value will be  unchanged.


// If GetId(|type|) already returns a non-zero id, that mapping will not be changed.

source/opt/type_manager.cpp
dneto0: Missing test?

test/opt/type_manager_test.cpp
```diff
@@ -552,6 +552,69 @@ OpMemoryModel Logical GLSL450
...
+  EXPECT_NE(context, nullptr);
+
+  const Type* type1 = context->get_type_mgr()->GetType(2u);
+  const Type* type2 = context->get_type_mgr()->GetType(3u);
+  EXPECT_NE(type1, nullptr);
+  EXPECT_NE(type2, nullptr);
+  EXPECT_FALSE(type1->IsSame(type2));
+}
+
+TEST(TypeManager, MemoryManagement) {
```
dneto0: Rename test to RemovingIdAvoidsUseAfterFree

source/opt/type_manager.cpp
alan-baker: will fix

source/opt/type_manager.cpp
alan-baker: will fix

source/opt/type_manager.h
alan-baker: ok

test/opt/type_manager_test.cpp
alan-baker: ok

alan-baker:
Rebased, squashed and pushed as 1ab8ad654a550d28593b1b38ac303adc66b1a241

Issue #1118, opened 2017-12-19
Update the legalization passes


s-perron:
For Spiregg, we want to try a different set of optimizations for legalization.  The idea is limit the number of optimization that are expected to be preformed on the illegal code.  The idea will be to do a thorough copy propagation.  All optimizations must be there to handle some specific problem related to copy propagation.

The new list of optimization will be:

1. Inlining - Place the definition and use in the same function to enable copy propagation.
1. Dead function elimination - remove references to private variables in function that are not used.  This will in general leave accesses to private variables in just 1 function.
1. Private to local - make private variables function scope to make it obvious that all uses are in a single function.
1. Scalar replacement - Split up aggregates so that copy propagation can act on each member individually.
1. Multi Store Elimination - Remove the loads and stores and keep values in the intermediate values.  They are easier to deal with because they are in ssa form.  This optimization also does a lot of it own copy propagation.
1. Insert-extract elimination - Scalar replacement can leave sequences of OpCompositConstruct that are then split up using OpCompositExtract.  This pass does the copy propagation through those.
1. CFG cleanup - remove branches that are know to always branch in one direction, and propagate the simplified Phi nodes.
1. Dead code elimination - Cleanup any unused code that could leave remnants of the illegal code.

antiagainst:
To keep track of known issues:

* HS/DS/GS is not supported right now because of insufficient capability probing #1101
* Some pointer-to-pointer types are not cleaned because of lacking of dead type elimination #833

antiagainst:
cc @ehsannas @chaoticbob

s-perron:
@antiagainst  pointed out that I might want "Dead branch elimination":

```
if (true) {
a = ...
} else {
b = ...
}
c = phi(a, b)
```

I thought cfg cleanup was suppose to clean this up, but apparently not.  @greg-lunarg @dnovillo Can the two of you comment on exactly what each optimization is suppose to do or not do?  How will the constant propagator handle this?

greg-lunarg:
Re --convert-local-access-chains, I am pretty sure that I can write a shader that will depend on this in order to legalize. Let me know if you would like me to write one.

greg-lunarg:
I am not completely up on the capabilities of CFGCleanup or the constant propagator. I can tell you that DBE converts all OpBranchConditionals with a constant condition to the appropriate OpBranch (and removes any newly dead blocks).

dnovillo:
[...]

PR #1119, opened 2017-12-19
Update the legalization passes.

s-perron:
Changes the set of optimizations done for legalization.  While doing
this, I added documentation to explain why we want each optimization.

A new option "--legalize-hlsl" is added so the legalization passes can
be easily run from the command line.

Implements issue #1118.


source/opt/optimizer.cpp
```diff
@@ -66,19 +66,50 @@ Optimizer& Optimizer::RegisterPass(PassToken&& p) {
...
+          // scalar replacement.
+          .RegisterPass(CreateInsertExtractElimPass())
+          // TODO: Add constant propagation here
+          // May need loop unrolling here see
+          // https://github.com/Microsoft/DirectXShaderCompiler/pull/930
+          .RegisterPass(CreateCFGCleanupPass())
+          // Get rid of unused code that leave traces of the illegal code.
+          .RegisterPass(CreateAggressiveDCEPass())
+          // TODO: Remove this once ADCE can do it.
+          .RegisterPass(CreateDeadVariableEliminationPass());
```
antiagainst: Can we have the auxiliary `CreateCompactIdsPass()` here? Running this recipe bloats the result-id ranges used.

source/opt/optimizer.cpp
```diff
@@ -66,19 +66,50 @@ Optimizer& Optimizer::RegisterPass(PassToken&& p) {
   return *this;
 }
 
+// The legalization passes take a spir-v shader generated by an HLSL front-end
+// and turn it into a valid vulkan spir-v shader.  There are two ways in which
+// the code will be invalid at the start:
+//
+// 1) There will of opaque objects, like images, which will be passed around
```
antiagainst: s/of/be/

source/opt/optimizer.cpp
dneto0: But ID compaction is not strictly a legalization.  If you think it's beneficial for other reasons then you should apply it separately. ??

antiagainst: That's true. I can register it separately then.

tools/opt/opt.cpp
```diff
@@ -421,6 +434,8 @@ OptStatus ParseFlags(int argc, const char** argv, Optimizer* optimizer,
         optimizer->RegisterPerformancePasses();
       } else if (0 == strcmp(cur_arg, "-Os")) {
         optimizer->RegisterSizePasses();
+      } else if (0 == strcmp(cur_arg, "-legalize-hlsl")) {
```
antiagainst: Should use double dash here.

dneto0:
Rebased, squashed, and pushed into master as 7505d24225

Issue #1101, opened 2017-12-15
Insufficient handling of capabilities


s-perron:
In the module class, there is a function `HasCapability`.  It is suppose to return true of the module has a certain capability.  It is implemented by looking for the capability in question in the list of OpCapability instruction.  The problem is that this is not enough.

When one capability depends on another, the other is implicitly enabled, even if it is not explicitly present.  This needs to be improved.

s-perron:
@alan-baker suggested we add this to the feature manager.  As we read each capability, the feature manger keeps track of all explicitly and implicitly enabled capabilities.

The name of the function on the module could change to `HasExplicitCapability` or something like that.

dneto0:
The validator handles this transitive closure in ValidationState_t::RegisterCapability.

Agree the optimizer needs the transitive closure too.

dneto0:
Fixed by #1122

PR #1122, opened 2017-12-20
Store all enabled capabilities in the feature manger.

s-perron:
In order to keep track of all of the implicit capabilities as well as
the explicit ones, we will add them all to the feature manager.  That is
the object that needs to be queried when checking if a capability is
enabled.

The name of the "HasCapability" function in the module was changed to
make it more obvious that it does not check for implied capabilities.

Fixes issue #1101.


source/opt/feature_manager.cpp
```diff
@@ -30,5 +37,202 @@ void FeatureManager::Analyze(ir::Module* module) {
   }
 }
 
+void FeatureManager::AddCapabilities(ir::Module* module) {
+  std::stack<SpvCapability> capabilities_to_add;
```
dneto0: Make this code data-driven.  The grammar tables encode the capability dependency graph.  See ValidationState_t::RegisterCapability
This will eliminate the entire switch statement below. (!!!)

It seems the IRContext doesn't have an spv_context, which surprises me.   It should in fact have a libspirv::AssemblyGrammar which does a boatload of stuff in a data-driven way.

test/opt/feature_manager_test.cpp
```diff
@@ -84,3 +84,51 @@ OpExtension "SPV_KHR_storage_buffer_storage_class"
...
+  const std::string text = R"(
+OpCapability Shader
+OpMemoryModel Logical GLSL450
+  )";
+
+  std::unique_ptr<ir::IRContext> context =
+      BuildModule(SPV_ENV_UNIVERSAL_1_2, nullptr, text);
+  ASSERT_NE(context, nullptr);
+
+  ASSERT_TRUE(context->get_feature_mgr()->HasCapability(SpvCapabilityShader));
```
dneto0: Nit: These should be EXPECT_TRUE, EXPECT_FALSE.

test/opt/feature_manager_test.cpp
```diff
@@ -84,3 +84,51 @@ OpExtension "SPV_KHR_storage_buffer_storage_class"
...
+  const std::string text = R"(
+OpCapability Kernel
+OpMemoryModel Logical GLSL450
+  )";
+
+  std::unique_ptr<ir::IRContext> context =
+      BuildModule(SPV_ENV_UNIVERSAL_1_2, nullptr, text);
+  ASSERT_NE(context, nullptr);
+
+  ASSERT_FALSE(context->get_feature_mgr()->HasCapability(SpvCapabilityShader));
```
dneto0: Similar

test/opt/feature_manager_test.cpp
```diff
@@ -84,3 +84,51 @@ OpExtension "SPV_KHR_storage_buffer_storage_class"
...
+OpMemoryModel Logical GLSL450
+  )";
+
+  std::unique_ptr<ir::IRContext> context =
+      BuildModule(SPV_ENV_UNIVERSAL_1_2, nullptr, text);
+  ASSERT_NE(context, nullptr);
+
+  // Check multiple levels of indirection.  Tessellation implies Shader, which
+  // implies Matrix.
+  ASSERT_TRUE(
```
dneto0: Similar.

source/opt/feature_manager.cpp
antiagainst: Done.

test/opt/feature_manager_test.cpp
antiagainst: Done

test/opt/feature_manager_test.cpp
antiagainst: Done

test/opt/feature_manager_test.cpp
antiagainst: Done

dneto0:
Rebased, squashed, and pushed into master as 756b277fb

PR #1123, opened 2017-12-20
Add asm, dis support for DebugInfo extended instruction set

dneto0:
Add grammar file for DebugInfo extended instruction set
- Each new operand enum kind in extinst.debuginfo.grammar.json maps
  to a new value in spv_operand_type_t.
- Add new concrete enum operand types for DebugInfo

Generate a C header for the DebugInfo extended instruction set

Add table lookup of DebugInfo extended instrutions

Handle the debug info operand types in binary parser,
disassembler, and assembler.

Add DebugInfo round trip tests for assembler, disassembler

Android.mk: Support DebugInfo extended instruction set

The extinst.debuginfo.grammar.json file is currently part of
SPIRV-Tools source.

It contributes operand type enums, so it has to be processed
along with the core grammar files.

We also generate a C header DebugInfo.h.

Add necessary grammar file processing to Android.mk.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -15,6 +15,11 @@
 set(GRAMMAR_PROCESSING_SCRIPT "${spirv-tools_SOURCE_DIR}/utils/generate_grammar_tables.py")
 set(VIMSYNTAX_PROCESSING_SCRIPT "${spirv-tools_SOURCE_DIR}/utils/generate_vim_syntax.py")
 set(XML_REGISTRY_PROCESSING_SCRIPT "${spirv-tools_SOURCE_DIR}/utils/generate_registry_tables.py")
+set(LANG_HEADER_PROCESSING_SCRIPT "${spirv-tools_SOURCE_DIR}/utils/generate_language_headers.py")
+
+# For now, assume the DebugInfo grammar file is in the current directory.
+# It might migrate to SPIRV-Headers.
+set(DEBUGINFO_GRAMMAR_JSON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/extinst.debuginfo.grammar.json")
 
 # macro() definitions are used in the following because we need to append .inc
 # file paths into some global lists (*_CPP_DEPENDS). And those global lists are
@@ -27,9 +32,10 @@ macro(spvtools_core_tables VERSION)
   add_custom_command(OUTPUT ${GRAMMAR_INSTS_INC_FILE} ${GRAMMAR_KINDS_INC_FILE}
     COMMAND ${PYTHON_EXECUTABLE} ${GRAMMAR_PROCESSING_SCRIPT}
       --spirv-core-grammar=${GRAMMAR_JSON_FILE}
+      --extinst-debuginfo-grammar=${DEBUGINFO_GRAMMAR_JSON_FILE}
       --core-insts-output=${GRAMMAR_INSTS_INC_FILE}
       --operand-kinds-output=${GRAMMAR_KINDS_INC_FILE}
-    DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${GRAMMAR_JSON_FILE}
+    DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${GRAMMAR_JSON_FILE} ${DEBUGINFO_GRAMMAR_JSON_FILE}
     COMMENT "Generate info tables for SPIR-V v${VERSION} core instructions and operands.")
   list(APPEND OPCODE_CPP_DEPENDS ${GRAMMAR_INSTS_INC_FILE})
   list(APPEND OPERAND_CPP_DEPENDS ${GRAMMAR_KINDS_INC_FILE})
@@ -43,9 +49,10 @@ macro(spvtools_enum_string_mapping VERSION)
      ${GRAMMAR_ENUM_STRING_MAPPING_INC_FILE}
     COMMAND ${PYTHON_EXECUTABLE} ${GRAMMAR_PROCESSING_SCRIPT}
       --spirv-core-grammar=${GRAMMAR_JSON_FILE}
+      --extinst-debuginfo-grammar=${DEBUGINFO_GRAMMAR_JSON_FILE}
       --extension-enum-output=${GRAMMAR_EXTENSION_ENUM_INC_FILE}
       --enum-string-mapping-output=${GRAMMAR_ENUM_STRING_MAPPING_INC_FILE}
-    DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${GRAMMAR_JSON_FILE}
+    DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${GRAMMAR_JSON_FILE} ${DEBUGINFO_GRAMMAR_JSON_FILE}
     COMMENT "Generate enum-string mapping for SPIR-V v${VERSION}.")
   list(APPEND EXTENSION_H_DEPENDS ${GRAMMAR_EXTENSION_ENUM_INC_FILE})
   list(APPEND ENUM_STRING_MAPPING_CPP_DEPENDS ${GRAMMAR_ENUM_STRING_MAPPING_INC_FILE})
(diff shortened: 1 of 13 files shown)


dneto0:
FYI: @AlexeySotkin

dneto0:
Rebased and fixed the formatting in the test code.  :-)

utils/generate_language_headers.py
```diff
@@ -0,0 +1,172 @@
...
+    def uses_guards(self):
+        return True
+
+    def comment_prefix(self):
+        return "// "
+
+    def const_definition(self, prefix, var, value):
+        # Use an anonymous enum.  Don't use a static const int variable because
+        # that can bloat binary size.
+        return 'enum {0} {1}{2} = {3}, {1}{2}_Max = 0x7fffffff {4};'.format(
```
antiagainst: What is this `*_Max` used for? I think we should avoid generating code into the header unless there is a clear need because once it's there, it would be hard to take back.

utils/generate_language_headers.py
```diff
@@ -0,0 +1,172 @@
...
+                '{', prefix, var, value, '}')
+
+    def enum_prefix(self, prefix, name):
+        return 'enum {}{} {}'.format(prefix, name, '{')
+
+    def enum_end(self, prefix, enum):
+        return '    {}{}Max = 0x7ffffff\n{};\n'.format(prefix, enum, '}')
```
antiagainst: IMHO, this `*Max` marker is less useful than a `*Last` or `*Next` since `*Max` is almost always `0x7fffffff`. But the core SPIR-V header adopts this scheme anyway.

utils/generate_language_headers.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def generate(self, grammar):
+        """Returns a string that is the language-specific header for the given grammar"""
+
+        parts = []
+        if grammar.copyright:
+            parts.extend(["{}{}".format(self.comment_prefix(), f) for f in grammar.copyright])
+        parts.append('')
+
+        guard = '{}_H_'.format(grammar.name)
```
antiagainst: I think this may be too short and likely to collide with some other header? What about `SPIRV_EXT_{}_H_` or something like that?

utils/generate_language_headers.py
```diff
@@ -0,0 +1,172 @@
...
+    def comment_prefix(self):
+        return "// "
+
+    def const_definition(self, prefix, var, value):
+        # Use an anonymous enum.  Don't use a static const int variable because
+        # that can bloat binary size.
+        return 'enum {0} {1}{2} = {3}, {1}{2}_Max = 0x7fffffff {4};'.format(
+                '{', prefix, var, value, '}')
+
+    def enum_prefix(self, prefix, name):
```
antiagainst: Do we want `typedef enum <name> { ... } <name>` here? That's how the core SPIR-V header defines enums. It will enable using the `enum` `<name>` without the `enum` keyword.

utils/generate_language_headers.py
```diff
@@ -0,0 +1,172 @@
...
+        if grammar.copyright:
+            parts.extend(["{}{}".format(self.comment_prefix(), f) for f in grammar.copyright])
+        parts.append('')
+
+        guard = '{}_H_'.format(grammar.name)
+        if self.uses_guards:
+            parts.append('#ifndef {}'.format(guard))
+            parts.append('#define {}'.format(guard))
+        parts.append('')
+
```
antiagainst: Do we want
```c++
#ifdef __cplusplus
  extern "C" {
#endif
```
here maybe?

source/CMakeLists.txt
```diff
@@ -109,6 +115,22 @@ macro(spvtools_vendor_tables VENDOR_TABLE)
...
+  set(OUT_H ${OUTBASE}.h)
+  add_custom_command(OUTPUT ${OUT_H}
+    COMMAND ${PYTHON_EXECUTABLE} ${LANG_HEADER_PROCESSING_SCRIPT}
+      --extinst-name=${NAME}
+      --extinst-grammar=${GRAMMAR_FILE}
+      --extinst-output-base=${OUTBASE}
+    DEPENDS ${LANG_HEADER_PROCESSING_SCRIPT} ${GRAMMAR_FILE}
+    COMMENT "Generate language specific header for ${NAME}.")
+  list(APPEND EXTINST_CPP_DEPENDS ${OUT_H})
+  add_custom_target(spirv-tools-${NAME} DEPENDS ${OUT_H})
```
antiagainst: Maybe use `spirv-tools-header-${NAME}` here?

I  see the names passed into `spvtools_vendor_tables()` and `spvtools_extinst_lang_headers()` only differ w.r.t. cases. If we provide the same name to the above two macros, targets will collide. It is subtle.

source/disassemble.cpp
```diff
@@ -277,6 +281,7 @@ void Disassembler::EmitOperand(const spv_parsed_instruction_t& inst,
     case SPV_OPERAND_TYPE_IMAGE:
     case SPV_OPERAND_TYPE_MEMORY_ACCESS:
     case SPV_OPERAND_TYPE_SELECTION_CONTROL:
+    case SPV_OPERAND_TYPE_DEBUG_INFO_FLAGS:
```
antiagainst: I think the whole list can be checked using `spvOperandIsConcreteMask()`?

source/disassemble.cpp
```diff
@@ -265,7 +265,11 @@ void Disassembler::EmitOperand(const spv_parsed_instruction_t& inst,
     case SPV_OPERAND_TYPE_BUILT_IN:
     case SPV_OPERAND_TYPE_GROUP_OPERATION:
     case SPV_OPERAND_TYPE_KERNEL_ENQ_FLAGS:
-    case SPV_OPERAND_TYPE_KERNEL_PROFILING_INFO: {
+    case SPV_OPERAND_TYPE_KERNEL_PROFILING_INFO:
+    case SPV_OPERAND_TYPE_DEBUG_BASE_TYPE_ATTRIBUTE_ENCODING:
+    case SPV_OPERAND_TYPE_DEBUG_COMPOSITE_TYPE:
+    case SPV_OPERAND_TYPE_DEBUG_TYPE_QUALIFIER:
+    case SPV_OPERAND_TYPE_DEBUG_OPERATION: {
```
antiagainst: Maybe the whole list can be checked using `spvOperandIsConcrete()`?

dneto0: Not here.  This stack of cases is for things that are enums that are not bit-flags.

There is a potential cleanup here and other places to sequences of if/else where each branch is a check via a function like spvOperandIsValueEnum and spvOperandIsBitEnum and so on.    (Or to make yet another more abstract set of enums.)   But I don't want to make that change as part of this CL.

source/disassemble.cpp
dneto0: Yes.
It changes the switch to a sequence of if/else.  I didn't want to make that big of a change in this CL to add incremental functionality.  It should be a separate cleanup.

utils/generate_language_headers.py
dneto0: It's there to ensure that the underlying type takes up 32-bits at least. 

The SPIR-V headers do this kind of thing:
https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/1.0/spirv.h#L69

utils/generate_language_headers.py
dneto0: There's collisions if you make all the enums typedefs.  :-(

In enum DebugInfoInstructions there is:
  DebugInfoDebugTypeQualifier = 4,
and yet there is also:
 enum DebugInfoDebugTypeQualifier {

If someone wants convenience in C they can make their own typedef.  If they want convenience in C++ they can use a using declaration.

dneto0:
Feedback addressed, except I deliberately choose not to:
 - change the switch structure. (Keep the extra case listings)
 - create typedefs.  Since I want to avoid identifier clashes

utils/generate_language_headers.py
antiagainst: I see. Thanks!

utils/generate_language_headers.py
antiagainst: Ah. Names...

dneto0:
Rebased, squashed, and pushed into master as 59de6100b

PR #1125, opened 2017-12-21
Add loop descriptors

Naghasan:
Add loop descriptors and some required dominator tree extensions.

Patch addition:

- Add post-order tree iterator.

- Add DominatorTreeNode extensions:
  - Add begin/end methods to do pre-order and post-order tree traversal from a given DominatorTreeNode

- Add DominatorTree extensions:
  - Add begin/end methods to do pre-order and post-order tree traversal
  - Tree traversal ignore by default the pseudo entry block
  - Retrieve a DominatorTreeNode from a basic block

- Add loop descriptor:
  - Add a LoopDescriptor class to register all loops in a given function.
  - Add a Loop class to describe a loop:
    - Loop parent
    - Nested loops
    - Loop depth
    - Loop header, merge, continue and preheader
    - Basic blocks that belong to the loop


dnovillo:
Thanks.  Looking at this today.

source/opt/dominator_tree.cpp
```diff
@@ -227,23 +227,28 @@ bool DominatorTree::StrictlyDominates(const ir::BasicBlock* a,
...
+  if (a == b) return false;
+  return Dominates(a, b);
+}
+
 bool DominatorTree::Dominates(uint32_t a, uint32_t b) const {
   // Check that both of the inputs are actual nodes.
-  auto a_itr = nodes_.find(a);
-  auto b_itr = nodes_.find(b);
-  if (a_itr == nodes_.end() || b_itr == nodes_.end()) return false;
+  const DominatorTreeNode* a_node = (*this)[a];
```
dneto0: This will create entries (with nullptr) for IDs that are not present.  That's undesirable.  The old behaviour was better.

source/opt/loop_descriptor.h
```diff
@@ -0,0 +1,246 @@
...
+  inline BasicBlock* GetHeaderBlock() { return loop_header_; }
+  inline const BasicBlock* GetHeaderBlock() const { return loop_header_; }
+
+  // Returns the latch basic block (basic block that holds the back-edge).
+  inline BasicBlock* GetLatchBlock() { return loop_continue_; }
+  inline const BasicBlock* GetLatchBlock() const { return loop_continue_; }
+
+  // Returns the BasicBlock which marks the end of the loop.
+  inline BasicBlock* GetMergeBlock() { return loop_merge_; }
+  inline const BasicBlock* GetMergeBlock() const { return loop_merge_; }
```
dnovillo: So, this will not work with compute kernels?  If it does, please document what happens when this function is used in unstructured IR.

source/opt/loop_descriptor.h
```diff
@@ -0,0 +1,246 @@
...
+  inline const Loop* GetParent() const { return parent_; }
+
+  inline bool HasParent() const { return parent_; }
+
+  // Returns true if this loop is itself nested within another loop.
+  inline bool IsNested() const { return parent_ != nullptr; }
+
+  // Returns the set of all basic blocks contained within the loop. Will be all
+  // BasicBlocks dominated by the header which are not also dominated by the
+  // loop merge block.
```
dnovillo: Will it include the header and merge blocks?  Just the body?

source/opt/loop_descriptor.h
```diff
@@ -0,0 +1,246 @@
...
+    return IsInsideLoop(bb->id());
+  }
+
+  // Returns true if the basic block id |bb_id| is inside this loop.
+  inline bool IsInsideLoop(uint32_t bb_id) const {
+    return loop_basic_blocks_.count(bb_id);
+  }
+
+  // Adds the Basic Block |bb| this loop and its parents.
+  void AddBasicBlockToLoop(const BasicBlock* bb) {
```
dnovillo: Seems like this should check that |bb| is actually dominated by the header and postdominated by the merge blocks (in the case of structured loops)

source/opt/loop_descriptor.h
```diff
@@ -0,0 +1,246 @@
...
+  // Set the loop preheader if it exists.
+  void SetLoopPreheader(IRContext* context,
+                        opt::DominatorAnalysis* dom_analysis);
+
+  // This is only to allow LoopDescriptor::dummy_top_loop_ to add top level
+  // loops as child.
+  friend class LoopDescriptor;
+};
+
+class LoopDescriptor {
```
dnovillo: Please add documentation for this class.

source/opt/tree_iterator.h
```diff
@@ -115,6 +115,131 @@ class TreeDFIterator {
...
+    return tmp;
+  }
+
+ private:
+  explicit inline PostOrderTreeDFIterator(NodePtr top_node)
+      : current_(top_node) {
+    if (current_) WalkToLeaf();
+  }
+
+  // end() iterator.
```
dnovillo: I don't understand this part of the comment.  This is a constructor.

source/opt/loop_descriptor.cpp
```diff
@@ -0,0 +1,164 @@
...
+      // The id of the continue basic block of this loop.
+      uint32_t continue_bb_id = merge_inst->GetSingleWordOperand(1);
+
+      // The continue target of this loop.
+      BasicBlock* merge_bb = context->cfg()->block(merge_bb_id);
+
+      // The continue target of this loop.
+      BasicBlock* continue_bb = context->cfg()->block(continue_bb_id);
+
+      // The basicblock containing the merge instruction.
```
dnovillo: s/basicblock/basic block/

source/opt/loop_descriptor.cpp
```diff
@@ -0,0 +1,164 @@
...
+      BasicBlock* header_bb = context->get_instr_block(merge_inst);
+
+      // Add the loop the list of all the loops in the function.
```
dnovillo: s/the/to the/

source/opt/dominator_tree.cpp
```diff
@@ -227,23 +227,28 @@ bool DominatorTree::StrictlyDominates(const ir::BasicBlock* a,
...
+  const DominatorTreeNode* b_node = (*this)[b];
```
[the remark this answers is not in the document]
Naghasan: Not necessary, that's true for maps but not for array/vector. Although, there might be the assumption that the returned element will always be valid (which is not true here).

source/opt/dominator_tree.cpp
Naghasan: I overloaded the operator (that's calling DominatorTree::operator[]). If a is not in the tree, the returned pointer is null.

source/opt/dominator_tree.cpp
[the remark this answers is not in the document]
Naghasan: @dnovillo @dneto0  Maybe `DominatorTree::operator[]` should be changed to a `GetTreeNode` (as this seems to be confusing)

source/opt/dominator_tree.h
```diff
@@ -80,13 +109,27 @@ class DominatorTree {
 
   // Depth first iterators.
   // Traverse the dominator tree in a depth first pre-order.
-  iterator begin() { return iterator(GetRoot()); }
+  // The pseudo-block is ignored.
+  iterator begin() { return ++iterator(GetRoot()); }
```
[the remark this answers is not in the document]
Naghasan: What do you mean ? The only thing that changed is the added `++` to skip the pseudo-entry block (as the basic block behind is not a valid basick block). The change was made on the const and non-const begin.

source/opt/loop_descriptor.cpp
```diff
@@ -0,0 +1,164 @@
...
+  //
+  // FIXME: if instead of having "ForEach*" functions, we had iterators, the
+  // standard library would be usable...
+  bool is_preheader = true;
+  uint32_t loop_header_id = loop_header_->id();
+  loop_pred->ForEachSuccessorLabel(
+      [&is_preheader, loop_header_id](const uint32_t id) {
+        if (id != loop_header_id) is_preheader = false;
+      });
+  if (is_preheader) loop_preheader_ = loop_pred;
```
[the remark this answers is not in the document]
Naghasan: Not sure to understand, this check is to make sure the candidate is a preheader.

At this stage, `loop_pred` is the unique predecessor of the loop header that is outside the loop. The lambda above checks that all successors of `loop_pred` is the loop header. So if `is_preheader` is true, `loop_pred` qualifies as a loop preheader.

source/opt/loop_descriptor.h
Naghasan: I added this method as public so a client could update the loop descriptor (to preserve the analysis). In such situation, the dominator tree might not be updated but that can be a requirement.

The check should be an assert right ?

source/opt/tree_iterator.h
Naghasan: That's the Ctor for the end() iterator, I fixed that

source/opt/loop_descriptor.h
```diff
@@ -0,0 +1,262 @@
...
+      : loop_header_(nullptr),
+        loop_continue_(nullptr),
+        loop_merge_(nullptr),
+        loop_preheader_(nullptr),
+        parent_(nullptr) {}
+
+  Loop(IRContext* context, opt::DominatorAnalysis* analysis, BasicBlock* header,
+       BasicBlock* continue_target, BasicBlock* merge_target);
+
+  // Iterators which allows access to the nested loops.
```
s-perron: Not clear which loops will be visited.  I get the impression it will only be the immediately nested loops (or some other appropriate language).  In a triply nested loop, the begin and end will not visit the inner most loop.  Is that correct?

source/opt/loop_descriptor.h
```diff
@@ -0,0 +1,262 @@
...
+    assert(postdom_analysis->Dominates(GetMergeBlock(), bb));
+#endif  // NDEBUG
+
+    for (Loop* loop = this; loop != nullptr; loop = loop->parent_) {
+      loop_basic_blocks_.insert(bb->id());
+    }
+  }
+
+  // Returns true if the parent basic block of |inst| belong to this loop.
+  inline bool IsLoopInvariant(Instruction* inst) const {
```
s-perron: This is confusing to me.  My understanding of a loop invariant instruction is one that computes the same value for every iteration of the loop.  This seems to be true if the instruction is inside the loop, but it says nothing about the value being computed.

source/opt/loop_descriptor.cpp
```diff
@@ -0,0 +1,162 @@
...
+  // Safe guard against invalid code, SPIR-V spec forbids loop with the entry
+  // node as header.
+  assert(loop_pred && "The header node is the entry block ?");
+
+  // So we have a unique basic block that can enter this loop.
+  // If this loop is the unique successor of this block, then it is a loop
+  // preheader.
+  bool is_preheader = true;
+  uint32_t loop_header_id = loop_header_->id();
+  loop_pred->ForEachSuccessorLabel(
```
s-perron: There should be a cleaner way of doing this.  All you want to say is that loop_pred has a single successor.  Because it is a predecessor of the header (that is how you found loop_pred), you can assume it has at least the header as a successor.

source/opt/loop_descriptor.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+void LoopDescriptor::PopulateList(const Function* f) {
+  IRContext* context = f->GetParent()->context();
+
+  opt::DominatorAnalysis* dom_analysis =
+      context->GetDominatorAnalysis(f, *context->cfg());
+
+  loops_.clear();
+
+  // Traverse the tree and apply the above functor to find all the OpLoopMerge
```
s-perron: I'm not sure which functor you are referring to.

source/opt/loop_descriptor.cpp
```diff
@@ -0,0 +1,162 @@
...
+       ir::make_range(dom_tree.post_begin(), dom_tree.post_end())) {
+    Instruction* merge_inst = node.bb_->GetLoopMergeInst();
+    if (merge_inst) {
+      // The id of the continue basic block of this loop.
+      uint32_t merge_bb_id = merge_inst->GetSingleWordOperand(0);
+
+      // The id of the continue basic block of this loop.
+      uint32_t continue_bb_id = merge_inst->GetSingleWordOperand(1);
+
+      // The continue target of this loop.
```
s-perron: Is this comment correct? I think it should be the merge target.

source/opt/loop_descriptor.cpp
```diff
@@ -0,0 +1,162 @@
...
+  // loop_merge_inst you still need to check dominance between each block
+  // manually.
+  opt::DominatorTree& dom_tree = dom_analysis->GetDomTree();
+  // Post-order traversal of the dominator tree: inner loop will be inserted
+  // first.
+  for (opt::DominatorTreeNode& node :
+       ir::make_range(dom_tree.post_begin(), dom_tree.post_end())) {
+    Instruction* merge_inst = node.bb_->GetLoopMergeInst();
+    if (merge_inst) {
+      // The id of the continue basic block of this loop.
```
s-perron: I think you mean the merge basic block, not the continue basic block.

source/opt/loop_descriptor.cpp
```diff
@@ -0,0 +1,162 @@
...
+                                           continue_bb, merge_bb));
+      Loop* current_loop = loops_.back().get();
+
+      // We have a bottom-up construction, so if this loop has nested-loops,
+      // they are by construction at the tail of the loop list.
+      for (auto itr = loops_.rbegin() + 1; itr != loops_.rend(); ++itr) {
+        Loop* previous_loop = itr->get();
+
+        // If the loop already has a parent, then it has been processed.
+        if (previous_loop->HasParent()) break;
```
s-perron: I'm not convinced this is correct:

```
void foo() {
  for( ... ) { // loop 4
    for( ... ) {} // loop 1
    for( ... ) {  // loop 3
      for ( ... ) {} // loop 2
    }
  }
}
```

Think you could process in the order in which they are numbered.  Then in this case, loop 4 will not be set as the parent of loop 1.  If this is wrong, please explain what I have missed.

test/opt/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 add_subdirectory(dominator_tree)
+add_subdirectory(loop_optimisations)
```
s-perron: We have been using the other spelling "optimization" in this project.

source/opt/loop_descriptor.cpp
Naghasan: Oh good catch.

We do a post-order df traversal for the dominator tree (from the main loop : `ir::make_range(dom_tree.post_begin(), dom_tree.post_end())`), so the inner most loops are visited first. So here loop 3 is dominated by loop 1, so loop 3 will be processed first.

But with this example:
```
void foo() {
  for( ... ) { // loop 4
    for( ... ) {  // loop 3
      for ( ... ) {} // loop 2
    }
    for( ... ) {} // loop 1
  }
}
```
The visit order will be 1, 2, 3, 4. When processing loop 3, loop 2 will have 3 as parent, and when processing loop 4, we will stop at loop 2 leaving loop 1 unset...

source/opt/loop_descriptor.cpp
Naghasan: That's true the canonical case is the preheader has 1 successor which is the loop header. I wanted to also handle a case where multiple branches point to the loop header (because of a pass didn't clean up the CFG).

source/opt/loop_descriptor.cpp
Naghasan: Left over of the some code refactoring. I removed it.

source/opt/loop_descriptor.cpp
Naghasan: Indeed, thanks.

source/opt/loop_descriptor.cpp
Naghasan: That's right.

source/opt/loop_descriptor.h
Naghasan: Yes, this allow only the visit of the children in the tree. That's required for to allow DF tree iterators to work properly. The iterator of LoopDescriptor do a post-order visit of the tree.

But we could add post_begin/post_end and pre_begin/pre_end visitors, it is kind of trivial now.

source/opt/loop_descriptor.h
Naghasan: Well I totally agree with you, I suppose a better name for the function would be more "IsDynamicallyLoopInvariant" (are we recomputing this value at each step). But after a loop invariant code motion, this also say something about the value being computed.

That's implemented this way for performance reasons. To take into account the value being computed, we need to walk through the use-def chain for each query and so to make this efficient we will need to cache the result. The resulting problem is that the cache could quickly become invalid and selective invalidation might be intrusive for users. I have the feeling that this would be more efficient (in terms of memory and speed) if it stay this way.

test/opt/CMakeLists.txt
Naghasan: Good point.

source/opt/loop_descriptor.h
s-perron: Could you update the comment to make it clearer?  I was not looking for a change in behaviour, but a more precise comment.

source/opt/loop_descriptor.h
s-perron: To avoid confusion, can we rename the function to overload "IsInsideLoop"?  It is a much easier to understand and does not carry with the understanding of "invariance", which is really a statement of the value computed.

source/opt/loop_descriptor.h
Naghasan: I updated it

source/opt/loop_descriptor.cpp
s-perron: Okay, I see what you mean, but I'm not sure of the value.  Leave it as is, and other can comment if they want.

s-perron:
@Naghasan I think all of @dnovillo comments have been dealt with.  Can you fix the failures on the continuous build, and then I can merge the PR.

Naghasan:
@s-perron The failure is now resolved

s-perron:
Rebased, squashed, and merged as e8ad02f3.

source/opt/dominator_tree.cpp
```diff
@@ -227,23 +227,28 @@ bool DominatorTree::StrictlyDominates(const ir::BasicBlock* a,
   return DominatorTree::StrictlyDominates(a->id(), b->id());
 }
 
+bool DominatorTree::StrictlyDominates(const DominatorTreeNode* a,
+                                      const DominatorTreeNode* b) const {
+  if (a == b) return false;
```
dnovillo: Should this assert (a && b)?

source/opt/loop_descriptor.h
dnovillo: This did not seem to get addressed.

source/opt/tree_iterator.h
```diff
@@ -115,6 +115,131 @@ class TreeDFIterator {
...
 
+// Helper class to iterate over a tree in a depth first post-order.
+// The class assumes the data structure is a tree, tree node type implements a
+// forward iterator.
+// At each step, the iterator holds the pointer to the current node and state of
+// the walk.
+// The state is recorded by stacking the iteration position of the node
+// children. To move to the next node, the iterator:
+//  - Looks at the top of the stack;
+//  - If the children iterator has reach the end, then the node become the
```
dnovillo: s/reach/reached/
s/become/becomes/

PR #1126, opened 2017-12-21
CONTRIBUTING: If you fixed a bug, say so

dneto0:


CONTRIBUTING.md
```diff
@@ -57,6 +57,10 @@ your changes:
 *   Create a pull request (PR) with your patch.
 *   Make sure the PR description clearly identified the problem, explains the
     solution, and references the issue if applicable.
+*   If your patch completely fixes bug 1234, the commit message should say
+    `Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1234`
```
pierremoreau: It should work as well by just writing `Fixes #1234`, unless you want to have the full URL in the commit.

CONTRIBUTING.md
```diff
@@ -57,6 +57,10 @@ your changes:
...
+    When you do this, the issue will be closed automatically when the commit
+    goes into master.  Also, this helps us update the [CHANGES](CHANGES) file.
```
pierremoreau: Should new features/fixes be added to the CHANGES file as part of the pull request?

dneto0: No.  That leads to useless conflicts between PRs that are reviewed concurrently.

CONTRIBUTING.md
dneto0: I wondered whether that was true but never tried the experiment.   
It's harmless to have the full URL.  And it's more informative if/when the code migrates to other repositories.  E.g. SPIRV-Tools ends up in the Android open source tree, and then how would a new reader know where to go?

CONTRIBUTING.md
pierremoreau: That’s a good point.

CONTRIBUTING.md
pierremoreau: Here is the syntax for automatically closing issues when a fix is pushed: https://help.github.com/articles/closing-issues-using-keywords/
I am fine having the URL in, just wanted to mention the alternative just in case.

pierremoreau:
Looks good to me

CONTRIBUTING.md
dneto0: OMG.  You can close an issue in a different repository.

> https://help.github.com/articles/closing-issues-using-keywords/#closing-an-issue-in-a-different-repository 

Isn't that a bad idea?  I hope there's some kind of access control.... (!)

pierremoreau: There is the last line: “will close the referenced issue in that repository, **provided you have push access to that repository**.” (emphasis mine). ;-) Thankfully, otherwise it would be crazy!

dneto0:
Rebased and pushed onto master as a27d673ce

PR #1128, opened 2017-12-21
Add ExtInst validation pass (GLSL only for now)

atgoo:
Validates all GLSL.std.450 extended instructions.

### DIFF
--- a/source/validate_ext_inst.cpp
+++ b/source/validate_ext_inst.cpp
@@ -149,8 +149,7 @@ spv_result_t ExtInstPass(ValidationState_t& _,
                    << "Result Type";
           }
 
-          if (ext_inst_key == GLSLstd450FindILsb ||
-              ext_inst_key == GLSLstd450FindUMsb ||
+          if (ext_inst_key == GLSLstd450FindUMsb ||
               ext_inst_key == GLSLstd450FindSMsb) {
             if (result_type_bit_width != 32) {
               return _.diag(SPV_ERROR_INVALID_DATA)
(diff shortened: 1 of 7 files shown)


source/validate_ext_inst.cpp
```diff
@@ -0,0 +1,755 @@
...
+          }
+
+          if (result_type_bit_width != _.GetBitWidth(operand_type)) {
+            return _.diag(SPV_ERROR_INVALID_DATA)
+                   << ext_inst_name() << ": "
+                   << "expected all operands to have the same bit width as "
+                   << "Result Type";
+          }
+
+          if (ext_inst_key == GLSLstd450FindILsb ||
```
dneto0: FindILsb does not have this restriction.
(Was confirmed by Khronos internal issue #173 if you want to look it up.)

atgoo: It seems that the issue was left undecided.

source/validate_ext_inst.cpp
```diff
@@ -0,0 +1,755 @@
...
+        uint32_t interpolant_storage_class = 0;
+        uint32_t interpolant_data_type = 0;
+        if (!_.GetPointerTypeInfo(interpolant_type, &interpolant_data_type,
+                                  &interpolant_storage_class)) {
+          return _.diag(SPV_ERROR_INVALID_DATA)
+                 << ext_inst_name() << ": "
+                 << "expected Interpolant to be a pointer";
+        }
+
+        if (result_type != interpolant_data_type) {
```
dneto0: Yes, thanks for thinking this one through.  I've filed Khronos internal issue 260 to fix the description in the spec.

source/validate_ext_inst.cpp
dneto0: We decided to leave ILsb as it is currently specified, without the constraint.
The undecided part was whether to relax the UMsb SMsb instructions.   We're waiting for a request change it.

dneto0:
Rebased, squashed, and pushed into master as 25d396b4a2

Issue #1130, opened 2017-12-21
constant propagator issues


dneto0:
Please address request for changes in:
https://github.com/KhronosGroup/SPIRV-Tools/pull/1102#pullrequestreview-85156526

Issue #1132, opened 2017-12-22
Add support for integer arithmetic different than 32-bits in folder


dnovillo:
FoldScalars() and FoldVectors() currently operate on 32-bit integers only.  They should be extended to handle 64-bit arithmetic as well.

dneto0:
Make sure they work with 16bit ints and floats too, or refuse to fold.

dnovillo:
Good point.  I've updated the mission statement for this issue.

s-perron:
We have written a new folding infrastructure that can do this.

Issue #943, opened 2017-11-02
Implement folding for floating point operations.


dnovillo:
The existing folder only handles integer arithmetic.  Support for floating point folding should follow the Vulkan spec (https://www.khronos.org/registry/vulkan/specs/1.0/pdf/vkspec.pdf)

dnovillo:
Needed for #889

antiagainst:
Folding of floating points can be tricky. I'd assume we should do this gradually by enabling easy cases first.

dneto0:
The case in #731 should be safe.

As for general folding, Graphics environments assume "fast relaxed math" by default, but Kernel (OpenCL) is like C by default, and is much more careful.   Also, each environment can flip the behaviour via decorations.  
E.g. Graphics code can use NoContraction to prevent certain transforms. 
Kernel code can use FPFastMathMode to enable lots of stuff, including full fast-math mode.

dnovillo:
#1182 has another example of floating point folding that should be handled (equality and inequality)

zeux:
I was looking into this a bit. Are there any SPIRV attributes that actually prevent constant folding for floating point instructions? There are obviously complications for non-constant simplifications such as `a * 0.0` which can't be simplified to 0.0 unless fast-math is enabled; but folding `3.0 * 4.0` to `7.0` or `1.0 > 0.9` to `true` should always be safe and correct?

zeux:
This is now partially addressed (105441360043248e91ec1c8b981f9541dab11083, 6669d8163dca26db521dfea477fd8589c47fb4c8) but still missing some operations (float->int conversion, mix, possibly others)

s-perron:
@zeux Just saw the question from Jan 30.  For shaders, we only need to look at the non-contract decoration.  As for the rules on precision, we need to follow https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#spirvenv-precision-operation.

In particular, "NaNs may not be generated.  Instructions that operate on a NaN may not result in a NaN.".  So we can fold `a * 0.0` to `0.0`.

zeux:
@s-perron Nice, thanks! This is basically `CanFoldFloatingPoint` in const_folding_rules.cpp; `a * 0.0` would be useful to fold to `0.0`, I've also noticed a few cases of `mix(a, b, 0.0)`/`mix(a, b, 1.0)` in our shaders which could be folded when contractions are enabled as well.

[...]

PR #1133, opened 2017-12-22
Fix https://github.com/KhronosGroup/SPIRV-Tools/issues/1130

dnovillo:
This addresses review feedback for the CCP implementation (which fixes
https://github.com/KhronosGroup/SPIRV-Tools/issues/889).

This adds more protection around the folding of instructions that would
not be supported by the folder.

@dneto0, there are more extensions that should be added to CCP.  Notably, 64-bit integer operations (#1132) and floating point operations (#943).

I will handle these in the coming days.


dneto0:
Rebased and pushed into master as 1acce992

Issue #1112, opened 2017-12-19
DecorationManager::HaveTheSameDecorations should be symmetric


dneto0:
There is a TODO for this.  I'm filing an issue for this because the method name implies the check is symmetric, but it is not symmetric

Cc: @pierremoreau

pierremoreau:
If no one started working on this one, I’ll give it a shot.

dneto0:
Thanks.  I don't think anybody else has started it.

I will be on vacation the next week, so take your time.  :-)

PR #1136, opened 2017-12-24
Fix DecorationManager::HaveTheSameDecorations

pierremoreau:
* SpvOpDecorateId was not recognised by `IsAnnotationInst()`, causing some of the new tests to fail.
* `HaveTheSameDecorations()` is now symmetric and the testing is no longer quadratic thanks to sorting the list and removing duplicates prior to the test.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1112

### DIFF
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -112,12 +112,9 @@ bool DecorationManager::HaveTheSameDecorations(uint32_t id1,
                             &decorateIdInstructionsFor2,
                             &decorateMemberInstructionsFor2);
 
-  if (decorateInstructionsFor1 != decorateInstructionsFor2) return false;
-  if (decorateIdInstructionsFor1 != decorateIdInstructionsFor2) return false;
-  if (decorateMemberInstructionsFor1 != decorateMemberInstructionsFor2)
-    return false;
-
-  return true;
+  return decorateInstructionsFor1 == decorateInstructionsFor2 &&
+         decorateIdInstructionsFor1 == decorateIdInstructionsFor2 &&
+         decorateMemberInstructionsFor1 == decorateMemberInstructionsFor2;
 }
 
 // TODO(pierremoreau): If OpDecorateId is referencing an OpConstant, one could
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -16,8 +16,8 @@
 
 #include <algorithm>
 #include <iostream>
+#include <set>
 #include <stack>
-#include <unordered_set>
 
 namespace spvtools {
 namespace opt {
@@ -43,7 +43,7 @@ std::vector<const ir::Instruction*> DecorationManager::GetDecorationsFor(
 bool DecorationManager::HaveTheSameDecorations(uint32_t id1,
                                                uint32_t id2) const {
   using InstructionList = std::vector<const ir::Instruction*>;
-  using UniqueSortedList = std::vector<std::u32string>;
+  using UniqueSortedList = std::set<std::u32string>;
 
   const InstructionList decorationsFor1 = GetDecorationsFor(id1, false);
   const InstructionList decorationsFor2 = GetDecorationsFor(id2, false);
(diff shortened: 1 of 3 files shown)


source/opt/decoration_manager.cpp
```diff
@@ -18,6 +18,93 @@
...
+bool compareBeforeDecorate(const Instruction* inst1, const Instruction* inst2) {
+  if (inst1->GetSingleWordInOperand(1u) > inst2->GetSingleWordInOperand(1u))
+    return false;
+  if (inst1->NumInOperands() > inst2->NumInOperands()) return false;
+  for (uint32_t i = 2u; i < inst1->NumInOperands(); ++i) {
+    const Operand& op1 = inst1->GetInOperand(i);
+    const Operand& op2 = inst2->GetInOperand(i);
+    if (op1.words.size() > op2.words.size()) return false;
+    for (uint32_t j = 0u; j < op1.words.size(); ++j)
+      if (op1.words[j] > op2.words[j]) return false;
```
atgoo: Missing case
if (op1.words[j] **<** op2.words[j]) return **true**;

source/opt/decoration_manager.cpp
```diff
@@ -18,6 +18,93 @@
...
+// * the decoration of |inst1| is strictly less than the one from |inst2|;
+// * and |inst1| has strictly fewer operands than |inst2|;
+// * and when comparing operands
+//   * |inst1|'s operand has strictly fewer words than |inst2|'s one;
+//   * and each word of |inst1|'s operand is strictly less than the ones from
+//     |inst2| (using their representation as uint32_t).
+bool compareBeforeDecorate(const Instruction* inst1, const Instruction* inst2) {
+  if (inst1->GetSingleWordInOperand(1u) > inst2->GetSingleWordInOperand(1u))
+    return false;
+  if (inst1->NumInOperands() > inst2->NumInOperands()) return false;
```
atgoo: Missing case 
if (inst1->NumInOperands() **<** inst2->NumInOperands()) return **true**;

source/opt/decoration_manager.cpp
```diff
@@ -18,6 +18,93 @@
...
+// * the member index of |inst1| is strictly less than the one from |inst2|;
+// * and |inst1| has strictly fewer operands than |inst2|;
+// * and when comparing operands
+//   * |inst1|'s operand has strictly fewer words than |inst2|'s one;
+//   * and each word of |inst1|'s operand is strictly less than the ones from
+//     |inst2| (using their representation as uint32_t).
+bool compareBeforeMemberDecorate(const Instruction* inst1,
+                                 const Instruction* inst2) {
+  if (inst1->GetSingleWordInOperand(2u) > inst2->GetSingleWordInOperand(2u))
+    return false;
```
atgoo: Same problem as in the above function: missing return true cases.

source/opt/decoration_manager.cpp
```diff
@@ -18,6 +18,93 @@
...
+  if (inst1->NumInOperands() > inst2->NumInOperands()) return false;
+  for (uint32_t i = 3u; i < inst1->NumInOperands(); ++i) {
+    const Operand& op1 = inst1->GetInOperand(i);
+    const Operand& op2 = inst2->GetInOperand(i);
+    if (op1.words.size() > op2.words.size()) return false;
+    for (uint32_t j = 0u; j < op1.words.size(); ++j)
+      if (op1.words[j] > op2.words[j]) return false;
+  }
+  return true;
```
atgoo: return false if operands are equal

source/opt/decoration_manager.cpp
```diff
@@ -18,6 +18,93 @@
...
+}
+
+// Returns true if |inst1| is equal to |inst2|, for SpvOpDecorate and
+// SpvOpDecorateId.
+//
+// When checking for equality, the target is not considered. The actual check
+// is the same as done by compareBeforeDecorate, but replacing all instances of
+// "strictly less than" by "equal to".
+bool compareEqualDecorate(const Instruction* inst1, const Instruction* inst2) {
+  if (inst1->NumInOperands() != inst2->NumInOperands()) return false;
```
atgoo: Consider comparing all words after target instead of operands.

source/opt/decoration_manager.cpp
```diff
@@ -39,24 +126,84 @@ std::vector<const ir::Instruction*> DecorationManager::GetDecorationsFor(
...
+  generateUniqueLists(decorationsFor2, &decorateInstructionsFor2,
+                      &decorateIdInstructionsFor2,
+                      &decorateMemberInstructionsFor2);
+
+  // This function checks that both lists contain the exact same decorations.
+  const auto compareLists = [this](const InstructionList& list1,
+                                   const InstructionList& list2) {
+    if (list1.size() != list2.size()) return false;
+    for (uint32_t i = 0u; i < list1.size(); ++i)
+      if (!AreDecorationsTheSame(list1[i], list2[i], true)) return false;
```
atgoo: Readability nit:
AreDecorationsTheSame(list1[i], list2[i], /* ignore_target = */ true)

atgoo:
Fixing the comparison function is likely to fix the timeout in windows tests.

source/opt/decoration_manager.cpp
```diff
@@ -39,24 +126,84 @@ std::vector<const ir::Instruction*> DecorationManager::GetDecorationsFor(
...
-        break;
+  using InstructionList = std::vector<const ir::Instruction*>;
+
+  const InstructionList decorationsFor1 = GetDecorationsFor(id1, false);
+  const InstructionList decorationsFor2 = GetDecorationsFor(id2, false);
+
+  // This function will sort and remove all duplicates for each of the three
+  // given lists.
+  const auto generateUniqueLists = [](const InstructionList& decorationList,
+                                      InstructionList* decorateList,
```
atgoo: If you just need to remove duplicates and compare the content, then you could use
std::unordered_set<std::u32string>
This way you won't need most of the custom code.

source/opt/decoration_manager.cpp
pierremoreau: The `bool operator!=(const Operand& op1, const Operand& op2)` relies on the equal operator of Operand, which does a word to word comparison already (as it uses the == operator of std::vector).

source/opt/decoration_manager.cpp
pierremoreau: Hum, and sorting to avoid a quadratic behaviour wouldn’t be necessary as we would rely on the hash map of the unordered set? I can do the change. Why an unordered set of `std::u32string` though (I didn’t even know that type existed), rather than a std::string?

pierremoreau:
Please have another look.

I made the change to using `u32string` and `std::unordered_set` to get a unique set of instructions. I still needed to sort them (and for that, copy from the unordered set to a `std::vector`).
It did solve the timeout issue in the Windows tests, which was a bit surprising considering it had no performance impact on my debug clang build on Linux on a 5 year old laptop (~4ms before, and still ~4ms after).

source/opt/decoration_manager.cpp
```diff
@@ -39,25 +40,81 @@ std::vector<const ir::Instruction*> DecorationManager::GetDecorationsFor(
...
+            case SpvOpDecorateId:
+              decorateIdListTmp.emplace(
+                  reinterpret_cast<char32_t*>(binaryInst.data()),
+                  binaryInst.size());
+              break;
+            default:
+              break;
+          }
+        }
+        std::copy(decorateListTmp.begin(), decorateListTmp.end(),
```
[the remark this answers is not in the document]
pierremoreau: I do need to sort them, otherwise it says that `{ Decorate %1 Constant, Decorate %1 Restrict }` and `{ Decorate %1 Restrict, Decorate %1 Constant }` are different, which is not the expected behaviour.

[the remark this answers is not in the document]
atgoo: If sets have the same content, then operator== would return true regardless of insertion order.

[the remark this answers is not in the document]
pierremoreau: Thanks for reminding me about std::set… I should have gone for that rather than emulating them with that unordered_map + vector.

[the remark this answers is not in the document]
pierremoreau: > If sets have the same content, then operator== would return true regardless of insertion order.

I meant for unordered sets, not ordered ones.

pierremoreau:
I reworked HaveTheSameDecorations to use sets rather than unordered sets + vectors, and explicitly built the decoration payload rather than using “ToBinaryWithoutAttachedDebugInsts”, as suggested.

source/opt/decoration_manager.cpp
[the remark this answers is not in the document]
atgoo: operator== would have worked the same for unordered_set as for set, I'm not sure why you were having problems.

If you are using set, you can also use std::vector<uint32_t> instead of u32string.

pierremoreau:
I also updated some of the names and comments, which were no longer meaningful.

dneto0:
Rebased, squashed, and pushed into master as 702852bd2

pierremoreau:
Thank you for the reviews!

Issue #1139, opened 2017-12-24
Turn the two `Linker::link()` functions into free functions


pierremoreau:
From https://github.com/KhronosGroup/SPIRV-Tools/pull/693#discussion_r141394561

>  Given that not much state needs to be retained, another acceptable solution is that the external API could have been two free functions, where the target environment and message consumer would be two more arguments.

dneto0:
WIP #1149

dneto0:
Fixed by #1149

Issue #768, opened 2017-08-18
Add a module linker (combine contents of multiple modules)


dneto0:
@pierremoreau has claimed this and is very far along.  See https://github.com/KhronosGroup/SPIRV-Tools/pull/693
Creating this issue to track it on the projects page.

LoopDawg:
@johnkslang pointed me to this issue.  I was (very tentatively) thinking about a module merger tool, but to a quick glance, it looks like there would be a lot of overlap with this linker, so the motivation for it may be covered or nearly covered by this.

Does the linker handle:

 (1) Eliminating logically identical functions when the IDs within the functional duplicates are different between module A and module B?  I believe this comes up sometimes when ubershaders are baked down.

(2) Introduction of multiple entry points into the linked result?  E.g, if standalone modules A and B are linked, then the result would contain AEntryPoints() and BEntryPoints()?

Related random thought: the current spirv-remap tool was constrained (by request, to ease integration into existing toolchains) to not look across modules, and to produce valid SPIRV modules, both of which limit its effectiveness.  Some of the underpinnings one would want for (1) above, and the similar in spirit redundant type elimination I believe this linker already does, are the same ones you might imagine for multi-module-scoped ID remapping.  However, I don't know if both things would be useful to have.  If you can merge modules, why not do that and eliminate redundancies for real, rather than fall back on a compressor dictionary with its potential horizon limits etc?

Perhaps some in-house toolchain couldn't handle multiple entry point modules.  Or perhaps there's still enough win from similar but not redundant code.  Still, it seems like the case for looking across modules gets weaker, once modules can be merged.

Also related: un-baking, and specialization constants.

dneto0:
Answering:

No to (1).   Eliminating logically redundant functions could be delivered as a code transform pass, available in spirv-opt (and underlying API).   I filed #805 to track that as a separate task.

Yes to (2).   
I just realized that entry point names can collide.  I don't know if the linker errors out, but it should.  Also, GLSL shaders are always named "main" by default.  I believe Glslang lets you set the name of the entry point when compiling for HLSL but not for GLSL.  (That seems an oversight.)

I agree that merging modules increases the opportunity for eliminating redundant code.  !

Vulkan conformance tests cover the case of multiple entry points in the same module.  So at least drivers should handle it.

pierremoreau:
[...]

PR #1149, opened 2018-01-03
Address remaining comments from the linker pull request

pierremoreau:
As very little information was kept in the Linker class, we can get rid of the whole class and have the `Link()` as free functions instead; the environment target as well as the consumer are passed along through an `spv_context` object.
The resulting linked_binary is passed as a pointer rather than a reference to follow the Google C++ Style guidelines; variables were also renamed to follow Google’s C++ Style guidelines.
The remaining changes are mostly about adding comments and TODOs, and some other formatting.

Addresses remaining comments from https://github.com/KhronosGroup/SPIRV-Tools/issues/693
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/768
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1139

### DIFF
--- a/include/spirv-tools/libspirv.hpp
+++ b/include/spirv-tools/libspirv.hpp
@@ -31,6 +31,10 @@ using MessageConsumer = std::function<void(
     const spv_position_t& /* position */, const char* /* message */
     )>;
 
+// Create a context with the targeted environemnt |env| and set its message
+// consumer to |consumer|.
+spv_context CreateContext(spv_target_env env, MessageConsumer consumer);
+
 // A RAII wrapper around a validator options object.
 class ValidatorOptions {
  public:
--- a/include/spirv-tools/libspirv.hpp
+++ b/include/spirv-tools/libspirv.hpp
@@ -31,9 +31,38 @@ using MessageConsumer = std::function<void(
     const spv_position_t& /* position */, const char* /* message */
     )>;
 
-// Create a context with the targeted environemnt |env| and set its message
-// consumer to |consumer|.
-spv_context CreateContext(spv_target_env env, MessageConsumer consumer);
+// C++ RAII wrapper around the C context object spv_context.
+class Context {
+ public:
+  // Constructs a context targeting the given environment |env|.
+  //
+  // The constructed instance will have an empty message consumer, which just
+  // ignores all messages from the library. Use SetMessageConsumer() to supply
+  // one if messages are of concern.
+  explicit Context(spv_target_env env);
+
+  // Enables move constructor/assignment operations.
+  Context(Context&& other);
+  Context& operator=(Context&& other);
+
+  // Disables copy constructor/assignment operations.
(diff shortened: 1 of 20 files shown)


include/spirv-tools/libspirv.hpp
```diff
@@ -31,6 +31,10 @@ using MessageConsumer = std::function<void(
     const spv_position_t& /* position */, const char* /* message */
     )>;
 
+// Create a context with the targeted environemnt |env| and set its message
```
dneto0: This seems naked.  It provides the creation idiom but not the destruction idiom.  The complexity is exposed in that you have to manually call spvContextDestroy in all the exit paths.

It would be better to have a RAII class that is convertible to an spv_context reference.  E.g.:

```
class Context {
  public:
   Context(spv_target_env  e, MessageConsumer mc);
   Context(Context&& other);
   ~Context() { spvContextDestroy(c); }
   spv_context&  operator() { return context_; }
  private:
   spv_context context_;
}
```


Inside a .cpp file:
```
   Context::Context(spv_target_env e, MessageConsumer mc) { 
     context_ = spvContextCreate(e); SetContextMessageConsumer(context_, mc); 
   }
   Context::Context(Context&& other)  { context_ = other.context_; other.context_ = nullptr; }
```

Then in linker.cpp you can create a Context object and pass it in as the argument where spv_context& is expected.

dneto0: Also the Context's copy constructor and assignment operators should be deleted.

pierremoreau: That would be better, indeed. Another option would be to pass an `spv_diagnostic*` to `Link()` to make it more similar to other entry points like `spvValidate()`, `spvBinaryParse()`, etc. It would be more C-friendly, and I could even get rid of `std::vector<>` and have it in “libspirv.h”.

pierremoreau: Which option would you prefer (between going with the C-API, with spv_diagnostic, or the Context class)? @dneto0

dneto0: I prefer the Context class for now.  The spv_diagnostic API design does not really work when you might emit more than one message.  The message consumer is better.
(But I also regret that that is such a complex lambda!  It should just have taken a string or string-view)

If someone wants a C API then we can make a C API version that wraps around the C++ version.  :-)

Thanks for asking.

pierremoreau: Sounds good, I’ll add the Context class. :-)

pierremoreau:
I added the Context class. I guess one could replace the “struct Impl” from SpirvTools by the Context class now.

Please have another look.

dneto0:
Rebased, squashed, and pushed into master as 7183ad526e6c

pierremoreau:
Thank you for the review @dneto0 !

PR #1150, opened 2018-01-03
Do not insert Phi nodes in CCP propagator.

dnovillo:
In CCP we should not need to insert Phi nodes because CCP never looks at
loads/stores.  This required adjusting two tests that relied on Phi
instructions being inserted.  I changed the tests to have the Phi
instructions pre-inserted.

I also added a new test to make sure that CCP does not try to look
through stores and loads.

Although this is just an efficiency fix for CCP, it is
also working around a bug in Phi insertion.  When Phi instructions are
inserted, they are never associated a basic block.  This causes a
segfault when the propagator tries to lookup CFG edges when analyzing
Phi instructions.

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -413,9 +413,13 @@ bool spvOpcodeIsReturn(SpvOp opcode) {
   }
 }
 
+bool spvOpcodeIsReturnOrAbort(SpvOp opcode) {
+  return spvOpcodeIsReturn(opcode) || opcode == SpvOpKill ||
+         opcode == SpvOpUnreachable;
+}
+
 bool spvOpcodeIsBlockTerminator(SpvOp opcode) {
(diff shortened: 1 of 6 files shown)


s-perron:
You did not update the Performance Passes and the Size passes to so that CCP is run after Local Multi store elimination, which is the most general load-store removal pass optimizations.  Is that intended?

dnovillo:
> You did not update the Performance Passes and the Size passes to so that CCP is run after Local Multi 
> store elimination, which is the most general load-store removal pass optimizations. Is that intended?

Thanks.  Fixed.

dneto0:
The phi-insertion bug is #1154

dneto0:
This passes tests on Windows with VS2017

dneto0:
This passes tests on OSX Apple LLVM version 9.0.0 (clang-900.0.38)

dnovillo:
Rebased and pushed as 135150a1a85e5b516a7ec299ba3b8f8513f2bc0b

Issue #1003, opened 2017-11-24
Stop using ToNop to remove instructions.


s-perron:
When an instruction is removed from the ir, we currently call "ToNop", which does not delete the instruction.  This causes later transformation that traverse the ir to look at a lot of nop instructions, which is just simply time consuming.  The OpNop instructions will also interfere with pattern matching transformations the we will be adding because the window used for peephole optimization will have to deal with an bunch of OpNops.  The change will take place in two steps:

- [x] Change KillInst so that it deletes instructions, and does not call ToNop.
- [ ] Change all other calls to calls to KillInst.

PR #1152, opened 2018-01-03
Replace calls to `ToNop` by `KillInst`.

s-perron:
Calling `ToNop` leaves around instructions that are pointless.  In
general it is better to remove the instruction completely.  That way
other optimizations will not need to look at them.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1003.


s-perron:
pushed as e8f2890c.

PR #1158, opened 2018-01-03
Add --print-before-all optimizer option

dneto0:
Disassembly is written to the standard error stream

### DIFF
--- a/source/opt/pass_manager.cpp
+++ b/source/opt/pass_manager.cpp
@@ -13,18 +13,41 @@
 // limitations under the License.
 
 #include "pass_manager.h"
+
+#include <iostream>
+#include <vector>
+
 #include "ir_context.h"
+#include "spirv-tools/libspirv.hpp"
 
 namespace spvtools {
+
 namespace opt {
 
 Pass::Status PassManager::Run(ir::IRContext* context) {
   auto status = Pass::Status::SuccessWithoutChange;
+
+  // If print_all_stream_ is not null, prints the disassembly of the module
+  // to that stream, with the given preamble and optionally the pass name.
+  auto print_disassembly = [&context, this](const char* preamble, Pass* pass) {
+    if (print_all_stream_) {
+      std::vector<uint32_t> binary;
+      context->module()->ToBinary(&binary, false);
+      SpirvTools t(SPV_ENV_UNIVERSAL_1_2);
+      std::string disassembly;
+      t.Disassemble(binary, &disassembly, 0);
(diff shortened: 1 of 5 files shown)


source/opt/pass_manager.cpp
```diff
@@ -21,6 +26,15 @@ namespace opt {
...
+    if (print_before_all_) {
+      std::vector<uint32_t> binary;
+      context->module()->ToBinary(&binary, false);
+      SpirvTools t(SPV_ENV_UNIVERSAL_1_2);
+      std::string out;
+      t.Disassemble(binary, &out, 0);
+      std::cerr << "; IR before pass " << pass->name() << "\n"
+                << out << std::endl;
+    }
     const auto one_status = pass->Run(context);
```
dnovillo: So, if this is the last pass we are running, it would be useful to print the IR after it.  The most common use I'm picturing with this is when debugging a single pass:

$ spirv-opt --print-before-all --my-pass in.spv -o out.spv

I'll want the IR after my pass is done, too.

A further enhancement to this (but not needed now) is to emit the dumps to a text file so they can be analyzed with diff.

dneto0: I modeled this after LLVM's opt -print-before-all, but I agree that would be more convenient.  How about changing the option to -print-all then it would print before each pass and once after the last pass.
I'll do that.

dneto0: LLVM's print-before-all writes to stderr, which is why I did stderr.  However, I realize the API should use an output stream rather than a fixed file.  I'll make that change too.

dneto0:
Please take another look.
Now the API takes an ostream parameter rather than a boolean.  spirv-opt still only writes to stderr.

source/opt/pass_manager.cpp
```diff
@@ -13,18 +13,41 @@
...
 
 namespace spvtools {
+
 namespace opt {
 
 Pass::Status PassManager::Run(ir::IRContext* context) {
   auto status = Pass::Status::SuccessWithoutChange;
+
+  // If print_all_stream_ is not null, prints the disassembly of the module
+  // to that stream, with the given preamble and optionally the pass // name.
```
dnovillo: No '//' between 'pass' and 'name'.

dneto0: Fixed. Thanks.

dneto0:
Keeping approval as I've now fixed the typo.

dneto0:
Rebased and pushed into master as c32e79eeefe15

PR #1162, opened 2018-01-04
Allow getting the base pointer of an image load/store.

### LINKED ISSUE #1160: assert(base_inst->IsValidBasePointer()) in Instruction::GetBaseAddress()
When running the optimization passes on the attached shader, this assertion fails:

```
  assert(base_inst->IsValidBasePointer() &&
         "We cannot have a base pointer come from this load");
```

[test.spv.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/1602043/test.spv.zip)

s-perron:
In value numbering, we treat loads and stores of images, ie OpImageLoad,
as a memory operation where it is interested in the "base address" of
the instruction.  In those cases, it is an image instruction.

The problem is that `Instruction::GetBaseAddress()` does not account for
the image instructions, so the assert at the end to make sure it found
a valid base address for its addressing mode fails.

The solution is to look at the load/store instruction to determine how
the assertion should be done.

Fixes #1160.

### DIFF
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -182,8 +182,21 @@ Instruction* Instruction::GetBaseAddress() const {
     }
   }
 
-  assert(base_inst->IsValidBasePointer() &&
-         "We cannot have a base pointer come from this load");
+  switch (opcode()) {
+    case SpvOpLoad:
+    case SpvOpStore:
+    case SpvOpAccessChain:
+    case SpvOpInBoundsAccessChain:
+    case SpvOpCopyObject:
+      // A load or store through a pointer.
+      assert(base_inst->IsValidBasePointer() &&
+             "We cannot have a base pointer come from this load");
+      break;
+    default:
+      // A load or store of an image.
+      assert(base_inst->IsValidBaseImage() && "We are expecting an image.");
+      break;
+  }
   return base_inst;
 }
 
@@ -423,6 +436,17 @@ bool Instruction::IsValidBasePointer() const {
     return true;
   }
   return false;
+}
+
+bool Instruction::IsValidBaseImage() const {
+  uint32_t tid = type_id();
+  if (tid == 0) {
+    return false;
+  }
+
+  ir::Instruction* type = context()->get_def_use_mgr()->GetDef(tid);
+  return (type->opcode() == SpvOpTypeImage ||
+          type->opcode() == SpvOpTypeSampledImage);
 }
 
 bool Instruction::IsOpaqueType() const {
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -268,16 +268,17 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   bool IsReadOnlyLoad() const;
 
   // Returns the instruction that gives the base address of an address
-  // calculation.  The instruction must be a load instruction.  In logical
-  // addressing mode, will return an OpVariable or OpFunctionParameter
-  // instruction. For relaxed logical addressing, it would also return a load of
-  // a pointer to an opaque object.  For physical addressing mode, could return
-  // other types of instructions.
+  // calculation.  The instruction must be a load, as defined by |IsLoad|,
+  // store, copy, or access chain instruction.  In logical addressing mode, will
+  // return an OpVariable or OpFunctionParameter instruction. For relaxed
+  // logical addressing, it would also return a load of a pointer to an opaque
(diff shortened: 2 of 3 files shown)


source/opt/instruction.h
```diff
@@ -363,6 +363,7 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   // logical addressing rules when using logical addressing.  Normal validation
   // rules for physical addressing.
   bool IsValidBasePointer() const;
+  bool IsValidBaseImage() const;
```
dneto0: Would need a comment saying this can be used as the base pointer an OpImageRead (or OpImageWrite too??)

source/opt/instruction.cpp
```diff
@@ -182,8 +182,21 @@ Instruction* Instruction::GetBaseAddress() const {
     }
   }
 
-  assert(base_inst->IsValidBasePointer() &&
-         "We cannot have a base pointer come from this load");
+  switch (opcode()) {
```
dneto0: Need to update the method comment for GetBaseAddress since it currently says "The current instruction must be a load operation".  That has to be generalized.

test/opt/instruction_test.cpp
```diff
@@ -577,4 +578,81 @@ TEST_F(OpaqueTypeTest, OpaqueStructTypes) {
     EXPECT_TRUE(type->IsOpaqueType());
   }
 }
+
+TEST_F(GetBaseTest, SampledImageLoad) {
```
dneto0: This is not a "load" but rather a sampling operation.  Please change the test name.

test/opt/instruction_test.cpp
```diff
@@ -577,4 +578,81 @@ TEST_F(OpaqueTypeTest, OpaqueStructTypes) {
...
+)";
+
+  std::unique_ptr<ir::IRContext> context =
+      BuildModule(SPV_ENV_UNIVERSAL_1_2, nullptr, text);
+  Instruction* load = context->get_def_use_mgr()->GetDef(21);
+  Instruction* base = context->get_def_use_mgr()->GetDef(20);
+  EXPECT_TRUE(load->GetBaseAddress() == base);
+}
+
+TEST_F(GetBaseTest, ImageLoad) {
```
dneto0: ImageLoad --> ImageRead

source/opt/instruction.cpp
s-perron: done.

test/opt/instruction_test.cpp
s-perron: Done

source/opt/instruction.h
s-perron: Added a comment let me know if it is okay.

test/opt/instruction_test.cpp
s-perron: In the `OpImageSampleImplicitLod` what does the "Lod" mean?  I understood that as a short form for "load".  I've changed the name, but I would still like this clarified.

s-perron:
The continuous build failures was a network issue.  It should be fine.

test/opt/instruction_test.cpp
dneto0: LOD = Level-of-detail;  a texture being sampled could have many slices at different resolutions, to be used when your view is "far" vs. "near".  See https://en.wikipedia.org/wiki/Mipmap

source/opt/instruction.h
```diff
@@ -268,16 +268,17 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
...
+  // instructions.
   Instruction* GetBaseAddress() const;
 
-  // Returns true if the instruction is a load from memory into a result id. It
-  // considers only core instructions. Memory-to-memory instructions are not
-  // considered loads.
+  // Returns true if the instruction loads from memory or samples an image, and
+  // stores the result into an id. It considers only core instructions.
+  // Memory-to-memory instructions are not considered loads.
   inline bool IsLoad() const;
```
dneto0: I wish we had a good word for "reads from memory" but which doesn't overload one of the instruction names to the exclusion of others.  At least the method comment describes it well enough for now.

Maybe this should be "IsLoadOrSamplingOp" ?  or "IsMemoryRead".

And looking at spvOpcodeIsLoad, I see it excludes atomic operations.....  

Let me mull this over for later.

dneto0:
Rebased, squashed, and pushed into master as  ccb921dd2bc6dd0

PR #1163, opened 2018-01-04
Address review comments from #985

dnovillo:
This addresses the review comments from #985.


dnovillo:
Pushed upstream as 5b52626eaa4bc599ba26b6b3fbee733cde23ba0f

Issue #1168, opened 2018-01-05
dead branch elim: backedge bookkeeping was getting stale


dneto0:
Originally https://github.com/KhronosGroup/glslang/issues/1205 

Symptom is that sometimes dead branch elimination would mistakenly delete a live block, resulting in invalid dangling references to other live blocks.

Diagnosis is in https://github.com/KhronosGroup/glslang/issues/1205#issuecomment-355429806

PR #1169, opened 2018-01-05
dead branch elim: Track killed backedges

dneto0:
When deleting branches and blocks, also remove them from
the backedges set, in case they were there.

This prevents us from keeping stale pointers to deleted Instruction
objects.  That memory could be used later by another instruction,
incorrectly signaling that something has a backedge reference, and
the dead branch eliminator could end up deleting live blocks.

Adds accessor method ir::BasicBlock::terminator

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1168

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -220,6 +220,7 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
     const uint32_t mergeLabId =
         mergeInst->GetSingleWordInOperand(kSelectionMergeMergeBlockIdInIdx);
     AddBranch(liveLabId, *bi);
+    backedges_.erase(br);
     context()->KillInst(br);
     context()->KillInst(mergeInst);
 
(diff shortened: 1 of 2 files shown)


dneto0:
@greg-lunarg please take a look.  I coded up the "delete terminators" solution.  It's not bad.
I prefer this solution because I don't want to keep invalid pointers lying around. That could get really confusing when debugging later on.

dneto0:
I can't think of a good way to unit test this new functionality.  The original problem only showed up under stress testing.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -220,6 +220,7 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
     const uint32_t mergeLabId =
         mergeInst->GetSingleWordInOperand(kSelectionMergeMergeBlockIdInIdx);
     AddBranch(liveLabId, *bi);
+    backedges_.erase(br);
```
greg-lunarg: I think this may never be a backedge, so I am worried it could be confusing. I think it is safe, so whatever you think best here.

dneto0: I think this case occurs for a single-block loop.  Granted, Glslang doesn't generate loops that way, so for now those are typically hand-constructed cases.

But this merits adding a test case, and perhaps a comment as well.  I'll do that in another PR.

Issue #1159, opened 2018-01-03
SSAPropagator::Run hangs on a complex shader


zeux:
Running the standard set of optimization passes:

			tools.RegisterSizePasses();
			tools.RegisterPerformancePasses();

On the attached shader hangs the optimizer; it looks like the loop in SSAPropagator::Run never terminates. This is most likely a semi-recent regression - the shader built & optimized fine with glslang & SPIRV-Tools around 1 month ago.

[test.spv.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/1602027/test.spv.zip)

dnovillo:
Thanks for the report.  Triaging.

dneto0:
I just saw a hang with --ccp on the test case from  https://github.com/KhronosGroup/glslang/issues/1205 Let me know if you want me to attach it here.

dnovillo:
Yes, please.  If it's not the same root cause, I'll create a new issue.

dneto0:
Here's the test case from Glslang issue 1205. 
[glslang-1205.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/1604318/glslang-1205.zip)

It hangs in spirv-opt just running --ccp

dnovillo:
The propagator is getting into an infinite loop when analyzing phi instructions, the SSA edge list is getting populated with the same ssa edge over and over again:

```
(gdb) p ssa_edge_uses_
$6 = std::queue wrapping: std::deque with 78441 elements = {0xa787e0, 0xa787e0, 0xa787e0,
  0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0,
  0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0, 0xa787e0, [...]
```

Reducing for a smaller reproducer.

Issue #896, opened 2017-10-19
Extend constant folding to do arithmetic simplifications


dnovillo:
This is a pre-requisite for #889.

Currently, a constant folding library exists for integer types (https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/fold.cpp).  This will need to be extended to fold floating point expressions and other simplifications (e.g., %21 - %21 folds to 0 for any integer result ID).

With the current implementation, we can get some of the required behaviour for SSA-CCP, but this folding helper will need to be extended.

dneto0:
- [ ] If all value operands of a Phi %phi are value %x, and %phi is not %x, then replace all uses of %phi with %x

dneto0:
- [ ] Undef absorption.  See #710

s-perron:
@dneto0 I think we did everything except for the undef absorption.  Since there is #710 is opened for that, 

@dnovillo Do you have specific simplifications that you want done?  Can they be listed here so we can know when we are done?  Otherwise, I would like to close this.  We have the ability to add the constant folding rules that we want.  A few are implemented.

s-perron:
Marking as done.  We have the infrastructure to get this done, and we can do most of the obvious stuff.  If we want to do anything specific open a new issue with an example.

PR #1171, opened 2018-01-05
Fix infinite simulation cycles in SSA propagator.

dnovillo:
This fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1159.  I
had missed a nuance in the original algorithm.  When simulating Phi
instructions, the SSA edges out of a Phi instruction should never be
added to the list of edges to simulate.

Phi instructions can be in SSA def-use cycles with other Phi
instructions.  This was causing the propagator to fall into an infinite
loop when the same def-use edge kept being added to the queue.

The original algorithm in the paper specifically separates the visit of
a Phi instruction vs the visit of a regular instruction.  This fix makes
the implementation match the original algorithm.

### DIFF
--- a/test/opt/ccp_test.cpp
+++ b/test/opt/ccp_test.cpp
@@ -415,6 +415,55 @@ TEST_F(CCPTest, HandleAbortInstructions) {
   SinglePassRunAndMatch<opt::CCPPass>(spv_asm, true);
 }
 
+TEST_F(CCPTest, SSAWebCycles) {
+  // Test reduced from https://github.com/KhronosGroup/SPIRV-Tools/issues/1159
+  // When there is a cycle in the SSA def-use web, the propagator was getting
+  // into an infinite loop.  SSA edges for Phi instructions should not be
+  // added to the edges to simulate.
+  const std::string spv_asm = R"(
+               OpCapability Shader
+          %1 = OpExtInstImport "GLSL.std.450"
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint Fragment %main "main"
+               OpExecutionMode %main OriginUpperLeft
+               OpSource GLSL 450
+               OpName %main "main"
+       %void = OpTypeVoid
+          %3 = OpTypeFunction %void
+        %int = OpTypeInt 32 1
+%_ptr_Function_int = OpTypePointer Function %int
+      %int_0 = OpConstant %int 0
+      %int_4 = OpConstant %int 4
+       %bool = OpTypeBool
+      %int_1 = OpConstant %int 1
+%_ptr_Output_int = OpTypePointer Output %int
+       %main = OpFunction %void None %3
+          %5 = OpLabel
+               OpBranch %11
+         %11 = OpLabel
+         %29 = OpPhi %int %int_0 %5 %22 %14
+         %30 = OpPhi %int %int_0 %5 %25 %14
+               OpLoopMerge %13 %14 None
+               OpBranch %15
+         %15 = OpLabel
+         %19 = OpSLessThan %bool %30 %int_4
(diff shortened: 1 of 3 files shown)


test/opt/ccp_test.cpp
```diff
@@ -415,6 +415,55 @@ TEST_F(CCPTest, HandleAbortInstructions) {
...
+         %15 = OpLabel
+         %19 = OpSLessThan %bool %30 %int_4
+; CHECK: OpBranchConditional %true {{%\d+}} {{%\d+}}
+               OpBranchConditional %19 %12 %13
+         %12 = OpLabel
+; CHECK: OpIAdd %int %int_0 %int_0
+         %22 = OpIAdd %int %29 %30
+               OpBranch %14
+         %14 = OpLabel
+; CHECK: OpPhi %int %int_0 {{%\d+}} %int_0 {{%\d+}}
```
dneto0: This suggests an optimization:  If all value operands of a Phi are the same value %x then replace uses of the phi with %x.   I'll add that to #896

dnovillo: That's already done, actually.  But in this case, we never get to it because we are purposely not adding SSA edges out of a Phi.  If the value of that Phi is used in a non-phi expression dominated by the Phi, it will be replaced with 0 (because the meet operator done in VisitPhi will return 0 as the meet value).

dnovillo: We test for this in other tests in test_ccp.cpp

dnovillo:
Thanks for the quick review.  Re-based and pushed as 716718a5e969f6b4e73cbc864db59a754a83aab3

test/opt/ccp_test.cpp
dneto0: I thought I had seen that somehwere but couldn't find it.  I didn't look far enough.

So to summarize (for myself):
- constant prop will get rid of the the non-phi dominated uses
- we could end up with these things as dead, and then dead code elim will get rid of most of them
- what remains is phis in cycles like this. 

That seems harmless/rare enough that it would be low priority to fix.

dnovillo:
On Fri, Jan 5, 2018 at 10:50 AM, David Neto <notifications@github.com>
wrote:

> *@dneto0* commented on this pull request.
>
> In test/opt/ccp_test.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1171#discussion_r159908100>
> :
>
> > +               OpBranch %11
> +         %11 = OpLabel
> +         %29 = OpPhi %int %int_0 %5 %22 %14
> +         %30 = OpPhi %int %int_0 %5 %25 %14
> +               OpLoopMerge %13 %14 None
> +               OpBranch %15
> +         %15 = OpLabel
> +         %19 = OpSLessThan %bool %30 %int_4
> +; CHECK: OpBranchConditional %true {{%\d+}} {{%\d+}}
> +               OpBranchConditional %19 %12 %13
> +         %12 = OpLabel
> +; CHECK: OpIAdd %int %int_0 %int_0
> +         %22 = OpIAdd %int %29 %30
> +               OpBranch %14
> +         %14 = OpLabel
> +; CHECK: OpPhi %int %int_0 {{%\d+}} %int_0 {{%\d+}}
>
> I thought I had seen that somehwere but couldn't find it. I didn't look
> far enough.
>
> So to summarize (for myself):
>
>    - constant prop will get rid of the the non-phi dominated uses
>    - we could end up with these things as dead, and then dead code elim
>    will get rid of most of them
>    - what remains is phis in cycles like this.
>
> That seems harmless/rare enough that it would be low priority to fix.
>

Correct.  The constant propagation problem is actually undecidable in
general, so all the algorithms are an actual approximation.  SSA-CCP
happens to be one of the fastest and powerful.

Issue #1143, opened 2017-12-29
Opt: --ccp (and -O and -Os) generates incorrect code for integer for-loops


greg-lunarg:
Test case:

; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 3
; Bound: 43
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main" %o
               OpExecutionMode %main OriginUpperLeft
               OpSource GLSL 430
               OpName %main "main"
               OpName %U_t "U_t"
               OpMemberName %U_t 0 "g_F"
               OpName %_ ""
               OpName %o "o"
               OpDecorate %_arr_float_uint_10 ArrayStride 4
               OpMemberDecorate %U_t 0 Offset 0
               OpDecorate %U_t BufferBlock
               OpDecorate %_ DescriptorSet 0
               OpDecorate %o Location 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
%_ptr_Function_float = OpTypePointer Function %float
    %float_0 = OpConstant %float 0
        %int = OpTypeInt 32 1
%_ptr_Function_int = OpTypePointer Function %int
      %int_0 = OpConstant %int 0
     %int_10 = OpConstant %int 10
       %bool = OpTypeBool
       %uint = OpTypeInt 32 0
    %uint_10 = OpConstant %uint 10
%_arr_float_uint_10 = OpTypeArray %float %uint_10
        %U_t = OpTypeStruct %_arr_float_uint_10
%_ptr_Uniform_U_t = OpTypePointer Uniform %U_t
          %_ = OpVariable %_ptr_Uniform_U_t Uniform
%_ptr_Uniform_float = OpTypePointer Uniform %float
      %int_1 = OpConstant %int 1
%_ptr_Output_float = OpTypePointer Output %float
          %o = OpVariable %_ptr_Output_float Output
       %main = OpFunction %void None %3
          %5 = OpLabel
               OpBranch %14
         %14 = OpLabel
         %41 = OpPhi %float %float_0 %5 %34 %17
         %42 = OpPhi %int %int_0 %5 %37 %17
               OpLoopMerge %16 %17 None
               OpBranch %18
         %18 = OpLabel
         %22 = OpSLessThan %bool %42 %int_10
[...]

PR #1172, opened 2018-01-05
Remove CCP from size and performance recipes, pending bugfixes

dneto0:
Currently CCP is incorrectly optimizing loops.
See https://github.com/KhronosGroup/SPIRV-Tools/issues/1143


dneto0:
cc: @greg-lunarg

dneto0:
Rebased and pushed into master as 3fbbd3c772

Issue #1157, opened 2018-01-03
Dead branch elim can generate invalid def use chains during processing


alan-baker:
If dead branch elim erases a section of code that is later referenced in a phi node, the pass may assert about missing definitions because because definition is removed before updating its uses.

dneto0:
It looks like https://github.com/KhronosGroup/glslang/issues/1205 might be the same issue as this.  There is a test case there.

greg-lunarg:
@alan-baker Have you looked at KhronosGroup/glslang#1205 yet? I will start with that unless you think you already have a solution.

greg-lunarg:
If you have something for this issue, pass me a branch and I will see if it also fixes the other issue.

alan-baker:
Here is a minimal testcase (run spirv-opt --eliminate-dead-branches in debug to trigger assertion).

>OpCapability Shader
>OpMemoryModel Logical GLSL450
>OpEntryPoint Fragment %func "main"
>OpExecutionMode %func OriginUpperLeft
>%void = OpTypeVoid
>%bool = OpTypeBool
>%1 = OpTypeFunction %void
>%false = OpConstantFalse %bool
>%true = OpConstantTrue %bool
>%bool_ptr = OpTypePointer Input %bool
>%var = OpVariable %bool_ptr Input
>%func = OpFunction %void None %1
>%entry = OpLabel
>OpSelectionMerge %outer_merge None
>OpBranchConditional %false %inner_header %else
>%inner_header = OpLabel
>%load = OpLoad %bool %var
>OpSelectionMerge %inner_merge None
>OpBranchConditional %load %inner_then %inner_merge
>%inner_then = OpLabel
>OpBranch %inner_merge
>%inner_merge = OpLabel
>%inner_phi = OpPhi %bool %false %inner_header %true %inner_then
>OpBranch %outer_merge
>%else = OpLabel
>OpBranch %outer_merge
>%outer_merge = OpLabel
>%outer_phi = OpPhi %bool %inner_phi %inner_merge %false %else
>OpReturn
>OpFunctionEnd

alan-baker:
The above example attempts to delete only %inner_merge and not %inner_then or %inner_merge which creates invalid IR.

alan-baker:
![minimal-cfg](https://user-images.githubusercontent.com/33432579/34578449-fb5a5536-f152-11e7-9927-3b10670b2d08.png)

Erases 11, but not 14 or 15.

[...]

PR #1173, opened 2018-01-05
Rewriting dead branch elimination.

alan-baker:
Addresses #1157 

Pass now paints live blocks and fixes constant branches and switches as
it goes. No longer requires structured control flow. It also removes
unreachable blocks as a side effect. It fixes the IR (phis) before doing
any code removal (other than terminator changes).

Added several unit tests for updated/new functionality.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -295,9 +295,8 @@ Optimizer::PassToken CreateLocalSingleBlockLoadStoreElimPass();
 // For all phi functions in merge block, replace all uses with the id
 // corresponding to the living predecessor.
 //
-// This pass only works on shaders (guaranteed to have structured control
-// flow). Note that some such branches and blocks may be left to avoid
-// creating invalid control flow. Improving this is left to future work.
+// Note that some branches and blocks may be left to avoid creating invalid
+// control flow. Improving this is left to future work.
 //
 // This pass is most effective when preceeded by passes which eliminate
 // local loads and stores, effectively propagating constant values where
--- a/source/opt/basic_block.cpp
+++ b/source/opt/basic_block.cpp
@@ -87,7 +87,7 @@ Instruction* BasicBlock::GetLoopMergeInst() {
 }
 
 void BasicBlock::ForEachSuccessorLabel(
-    const std::function<void(const uint32_t)>& f) {
+    const std::function<void(const uint32_t)>& f) const {
   const auto br = &insts_.back();
   switch (br->opcode()) {
     case SpvOpBranch: {
@@ -104,6 +104,15 @@ void BasicBlock::ForEachSuccessorLabel(
     default:
       break;
   }
+}
+
+bool BasicBlock::IsSuccessor(const ir::BasicBlock* block) const {
+  uint32_t succId = block->id();
+  bool isSuccessor = false;
+  ForEachSuccessorLabel([&isSuccessor, succId](const uint32_t label) {
+    if (label == succId) isSuccessor = true;
+  });
+  return isSuccessor;
 }
 
 void BasicBlock::ForMergeAndContinueLabel(
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -77,21 +77,27 @@ void DeadBranchElimPass::AddBranch(uint32_t labelId, ir::BasicBlock* bp) {
       context(), SpvOpBranch, 0, 0,
       {{spv_operand_type_t::SPV_OPERAND_TYPE_ID, {labelId}}}));
   get_def_use_mgr()->AnalyzeInstDefUse(&*newBranch);
-  context()->set_instr_block(&*newBranch, bp);
   bp->AddInstruction(std::move(newBranch));
 }
 
 bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
+  // This pass only requires correct instruction block mappings for the input.
+  // This pass does not preserve the block mapping, so it is not kept
+  // up-to-date during processing.
   auto get_parent_block = [this](uint32_t id) {
     return context()->get_instr_block(get_def_use_mgr()->GetDef(id));
   };
 
   // Mark live blocks reachable from the entry. Simplify constant branches and
-  // switches as you proceed to limit the number of live blocks. Be careful not
+  // switches as you proceed, to limit the number of live blocks. Be careful not
   // to eliminate backedges even if they are dead, but the header is live.
+  // Likewise, unreachable merge blocks named in live merge instruction must be
+  // retained (though they may be clobbered).
+  //
+  // |continues| maps the continue target to its corresponding header.
   std::unordered_map<ir::BasicBlock*, ir::BasicBlock*> continues;
   bool modified = false;
-  std::unordered_set<ir::BasicBlock*> liveBlocks;
+  std::unordered_set<ir::BasicBlock*> live_blocks;
   std::vector<ir::BasicBlock*> stack;
   stack.push_back(&*func->begin());
   while (!stack.empty()) {
@@ -99,37 +105,37 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
     stack.pop_back();
 
     // Live blocks doubles as visited set.
-    if (!liveBlocks.insert(block).second) continue;
+    if (!live_blocks.insert(block).second) continue;
 
-    uint32_t contId = block->ContinueBlockIdIfAny();
-    if (contId != 0) continues[get_parent_block(contId)] = block;
+    uint32_t cont_id = block->ContinueBlockIdIfAny();
+    if (cont_id != 0) continues[get_parent_block(cont_id)] = block;
 
-    ir::Instruction* terminator = &*block->tail();
-    uint32_t liveLabId = 0;
+    ir::Instruction* terminator = block->terminator();
+    uint32_t live_lab_id = 0;
     // Check if the terminator has a single valid successor.
     if (terminator->opcode() == SpvOpBranchConditional) {
       bool condVal;
       if (GetConstCondition(terminator->GetSingleWordInOperand(0u), &condVal)) {
-        liveLabId = terminator->GetSingleWordInOperand(
+        live_lab_id = terminator->GetSingleWordInOperand(
             condVal ? kBranchCondTrueLabIdInIdx : kBranchCondFalseLabIdInIdx);
       }
     } else if (terminator->opcode() == SpvOpSwitch) {
-      uint32_t selVal;
-      if (GetConstInteger(terminator->GetSingleWordInOperand(0u), &selVal)) {
-        // Search switch operands for selector value, set liveLabId to
+      uint32_t sel_val;
+      if (GetConstInteger(terminator->GetSingleWordInOperand(0u), &sel_val)) {
+        // Search switch operands for selector value, set live_lab_id to
         // corresponding label, use default if not found.
         uint32_t icnt = 0;
-        uint32_t caseVal;
+        uint32_t case_val;
         terminator->ForEachInOperand(
-            [&icnt, &caseVal, &selVal, &liveLabId](const uint32_t* idp) {
+            [&icnt, &case_val, &sel_val, &live_lab_id](const uint32_t* idp) {
               if (icnt == 1) {
                 // Start with default label.
-                liveLabId = *idp;
+                live_lab_id = *idp;
               } else if (icnt > 1) {
                 if (icnt % 2 == 0) {
-                  caseVal = *idp;
+                  case_val = *idp;
                 } else {
-                  if (caseVal == selVal) liveLabId = *idp;
+                  if (case_val == sel_val) live_lab_id = *idp;
                 }
               }
               ++icnt;
@@ -138,24 +144,24 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
     }
 
     // Don't simplify branches of continue blocks. A path from the continue to
-    // the merge is required.
+    // the header is required.
     // TODO(alan-baker): They can be simplified iff there remains a path to the
     // backedge. Structured control flow should guarantee one path hits the
     // backedge, but I've removed the requirement for structured control flow
     // from this pass.
-    bool simplify = liveLabId != 0 && !continues.count(block);
+    bool simplify = live_lab_id != 0 && !continues.count(block);
 
     if (simplify) {
       modified = true;
       // Replace with unconditional branch.
       // Remove the merge instruction if it is a selection merge.
-      AddBranch(liveLabId, block);
+      AddBranch(live_lab_id, block);
       context()->KillInst(terminator);
       ir::Instruction* mergeInst = block->GetMergeInst();
       if (mergeInst->opcode() == SpvOpSelectionMerge) {
         context()->KillInst(mergeInst);
       }
-      stack.push_back(get_parent_block(liveLabId));
+      stack.push_back(get_parent_block(live_lab_id));
     } else {
       // All successors are live.
       block->ForEachSuccessorLabel(
@@ -168,28 +174,31 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
   // Check for unreachable merge and continue blocks with live headers, those
   // blocks must remain. Continues are tracked separately so that when updating
   // live phi nodes with an edge from a continue I can replace it with an
-  // undef (because I still clobber the continue block).
-  std::unordered_set<ir::BasicBlock*> unreachableMerges;
-  std::unordered_map<ir::BasicBlock*, ir::BasicBlock*> unreachableContinues;
-  for (auto block : liveBlocks) {
-    uint32_t mergeId = block->MergeBlockIdIfAny();
-    if (mergeId != 0) {
-      ir::BasicBlock* mergeBlock = get_parent_block(mergeId);
-      if (!liveBlocks.count(mergeBlock)) {
-        unreachableMerges.insert(mergeBlock);
+  // undef (because we clobber the instructions inside continue block).
+  //
+  // |unreachable_continues| maps continue targets that cannot be reached to
+  // merge instruction that declares them.
+  std::unordered_set<ir::BasicBlock*> unreachable_merges;
+  std::unordered_map<ir::BasicBlock*, ir::BasicBlock*> unreachable_continues;
+  for (auto block : live_blocks) {
+    uint32_t merge_id = block->MergeBlockIdIfAny();
+    if (merge_id != 0) {
+      ir::BasicBlock* merge_block = get_parent_block(merge_id);
+      if (!live_blocks.count(merge_block)) {
+        unreachable_merges.insert(merge_block);
       }
-      uint32_t contId = block->ContinueBlockIdIfAny();
-      if (contId != 0) {
-        ir::BasicBlock* contBlock = get_parent_block(contId);
-        if (!liveBlocks.count(contBlock)) {
-          unreachableContinues[contBlock] = block;
+      uint32_t cont_id = block->ContinueBlockIdIfAny();
+      if (cont_id != 0) {
+        ir::BasicBlock* cont_block = get_parent_block(cont_id);
+        if (!live_blocks.count(cont_block)) {
+          unreachable_continues[cont_block] = block;
         }
       }
     }
   }
 
   for (auto& block : *func) {
-    if (liveBlocks.count(&block)) {
+    if (live_blocks.count(&block)) {
       // Fix phis so that only live (or unremovable) incoming edges are
       // present. If the block now only has a single live incoming edge, remove
       // the phi and replace its uses with its data input.
@@ -207,7 +216,7 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
         for (uint32_t i = 1; i < inst->NumInOperands(); i += 2) {
           ir::BasicBlock* inc =
               get_parent_block(inst->GetSingleWordInOperand(i));
-          if (unreachableContinues.count(inc)) {
+          if (unreachable_continues.count(inc)) {
             // Replace incoming value with undef.
             operands.emplace_back(
                 SPV_OPERAND_TYPE_ID,
@@ -215,7 +224,7 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
             operands.push_back(inst->GetInOperand(i));
             changed = true;
             backedge_added = true;
-          } else if (liveBlocks.count(inc) && inc->IsSuccessor(&block)) {
+          } else if (live_blocks.count(inc) && inc->IsSuccessor(&block)) {
             // Keep live incoming edge.
             operands.push_back(inst->GetInOperand(i - 1));
             operands.push_back(inst->GetInOperand(i));
@@ -265,7 +274,7 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
 
   // Erase dead blocks.
   for (auto ebi = func->begin(); ebi != func->end();) {
-    if (unreachableMerges.count(&*ebi)) {
+    if (unreachable_merges.count(&*ebi)) {
       // Make unreachable, but leave the label.
       KillAllInsts(&*ebi, false);
       // Add unreachable terminator.
@@ -274,18 +283,19 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
                                       std::initializer_list<ir::Operand>{}));
       ++ebi;
       modified = true;
-    } else if (unreachableContinues.count(&*ebi)) {
+    } else if (unreachable_continues.count(&*ebi)) {
       // Make unreachable, but leave the label.
       KillAllInsts(&*ebi, false);
       // Add unconditional branch to header.
-      uint32_t contId = unreachableContinues[&*ebi]->id();
-      ebi->AddInstruction(MakeUnique<ir::Instruction>(
-          context(), SpvOpBranch, 0, 0,
-          std::initializer_list<ir::Operand>{{SPV_OPERAND_TYPE_ID, {contId}}}));
+      uint32_t cont_id = unreachable_continues[&*ebi]->id();
+      ebi->AddInstruction(
+          MakeUnique<ir::Instruction>(context(), SpvOpBranch, 0, 0,
+                                      std::initializer_list<ir::Operand>{
+                                          {SPV_OPERAND_TYPE_ID, {cont_id}}}));
       get_def_use_mgr()->AnalyzeInstUse(&*ebi->tail());
       ++ebi;
       modified = true;
-    } else if (!liveBlocks.count(&*ebi)) {
+    } else if (!live_blocks.count(&*ebi)) {
       // Kill this block.
       KillAllInsts(&*ebi);
       ebi = ebi.Erase();
@@ -327,7 +337,6 @@ Pass::Status DeadBranchElimPass::ProcessImpl() {
   ProcessFunction pfn = [this](ir::Function* fp) {
     return EliminateDeadBranches(fp);
   };
-  // bool modified = ProcessEntryPointCallTree(pfn, get_module());
   bool modified = ProcessReachableCallTree(pfn, context());
   return modified ? Status::SuccessWithChange : Status::SuccessWithoutChange;
 }
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -213,7 +213,7 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
                 std::initializer_list<uint32_t>{Type2Undef(inst->type_id())});
             operands.push_back(inst->GetInOperand(i));
             changed = true;
-          } else if (liveBlocks.count(inc)) {
+          } else if (liveBlocks.count(inc) && inc->IsSuccessor(&block)) {
             // Keep live incoming edge.
             operands.push_back(inst->GetInOperand(i - 1));
             operands.push_back(inst->GetInOperand(i));
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -120,24 +120,20 @@ bool DeadBranchElimPass::EliminateDeadBranches(ir::Function* func) {
         // corresponding label, use default if not found.
         uint32_t icnt = 0;
         uint32_t caseVal;
-        terminator->ForEachInOperand([
-          &icnt,
-          &caseVal,
-          &selVal,
-          &liveLabId,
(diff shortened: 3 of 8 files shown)


dneto0:
Clang errors out due to unused captures.

test/opt/dead_branch_elim_test.cpp
```diff
@@ -1334,7 +1334,7 @@ OpBranch %33
 OpStore %oc %19
 OpBranch %28
 %29 = OpLabel
-OpBranchConditional %false %27 %28
+OpBranch %27
```
dneto0: There is no longer a backedge for this loop.  That's invalid.

alan-baker: This branch is the backedge (%27 contains OpLoopMerge). The exit to the merge is no longer present.

dneto0: whoops.  You're right.

greg-lunarg:
VS does not like the trailing comma after liveLabId. :(

dneto0:
> VS does not like the trailing comma after liveLabId. :(

And it shows up as an error or warning in YouCompleteMe with my GCC.

dneto0:
While we're here, @greg-lunarg : This rewrite handles some extra cases that @alan-baker is seeing.
  
An example is where there's a selection construct but control flow reconverges early, one block before the merge block, and that early block has a Phi.  The old algorithm doesn't know to fixup that phi because it never sees it.  The result with the old algorithm is a phi with dangling references to the dead block that has been removed.  See the EarlyReconvergence test
https://github.com/KhronosGroup/SPIRV-Tools/pull/1173/files#diff-34499625bbeb0676d52a53a9454d9e96R1461

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
-  for (auto bi = structuredOrder.begin(); bi != structuredOrder.end(); ++bi) {
-    // Skip blocks that are already in the elimination set
-    if (elimBlocks.find(*bi) != elimBlocks.end()) continue;
-    // Skip blocks that don't have conditional branch preceded
-    // by OpSelectionMerge
-    ir::Instruction* br;
-    ir::Instruction* mergeInst;
-    uint32_t condId;
-    if (!GetSelectionBranch(*bi, &br, &mergeInst, &condId)) continue;
+  auto get_parent_block = [this](uint32_t id) {
```
dneto0: Should say that we only need the instr-to-block mapping in for instructions that exist when we start processing or that we explicitly analyze later. You don't need to update it for instructions that are removed.
That explains why later creation of branch instructions don't need to invalidate or update the analysis.

greg-lunarg:
> While we're here, @greg-lunarg : This rewrite handles some extra cases that @alan-baker is seeing.

Totally cool. However, I am seeing regressions. I am tracking them down now.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
-      dLabId = (*dbi)->id();
+  // Check for unreachable merge and continue blocks with live headers, those
+  // blocks must remain. Continues are tracked separately so that when updating
+  // live phi nodes with an edge from a continue I can replace it with an
+  // undef (because I still clobber the continue block).
+  std::unordered_set<ir::BasicBlock*> unreachableMerges;
+  std::unordered_map<ir::BasicBlock*, ir::BasicBlock*> unreachableContinues;
+  for (auto block : liveBlocks) {
+    uint32_t mergeId = block->MergeBlockIdIfAny();
+    if (mergeId != 0) {
```
dneto0: Can shorten this by one line by putting the declaration inside the "if".
```
if (auto merge_id = block->MergeBlockIdIfAny()) {
...
```

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
-      elimBlocks.insert(*dbi);
-      ++dbi;
-      if (dbi == structuredOrder.end()) break;
-      dLabId = (*dbi)->id();
+  // Check for unreachable merge and continue blocks with live headers, those
+  // blocks must remain. Continues are tracked separately so that when updating
+  // live phi nodes with an edge from a continue I can replace it with an
+  // undef (because I still clobber the continue block).
+  std::unordered_set<ir::BasicBlock*> unreachableMerges;
+  std::unordered_map<ir::BasicBlock*, ir::BasicBlock*> unreachableContinues;
```
dneto0: Describe what is mapped to what.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
-      backedges_.erase((*dbi)->terminator());
-      KillAllInsts(*dbi);
-      elimBlocks.insert(*dbi);
-      ++dbi;
-      if (dbi == structuredOrder.end()) break;
-      dLabId = (*dbi)->id();
+  // Check for unreachable merge and continue blocks with live headers, those
+  // blocks must remain. Continues are tracked separately so that when updating
+  // live phi nodes with an edge from a continue I can replace it with an
+  // undef (because I still clobber the continue block).
```
dneto0: "(because we clobber the instructions inside of the continue block)."  

Most important is the clarification that the definition instructions inside the continue block are removed, but the continue block itself survives.

The "still" wasn't doing anything for you here.  And I use the royal "we" because the reader should feel they own the code too.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
+        if (!liveBlocks.count(contBlock)) {
+          unreachableContinues[contBlock] = block;
+        }
+      }
     }
+  }
 
-    // If last block reached, look for next dead branch
-    if (dbi == structuredOrder.end()) continue;
+  for (auto& block : *func) {
```
dneto0: Given the length of this function, you need a comment saying what this phase does.    I guess I suggest moving the comment starting with "Fix phis" out to the above the iteration over the blocks.

Then rehprase it as   "Fix phis in reachable blocks so that ...."

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
-          if (deadPreds.find(cfg()->block(*idp)) == deadPreds.end()) {
-            ++lcnt;
-            lidx = icnt - 1;
+        if (changed) {
+          // Replace the phi with either a single value or a rebuilt phi.
+          uint32_t replId;
+          if (operands.size() == 2) {
+            replId = operands[0].words[0];
+          } else {
+            replId = TakeNextId();
```
dneto0: If we're going to delete the old phi we should reuse its result id, at least in principle.
  
That makes it tricky to keep inst-def-use up to date but it should be doable.  You'd have to kill the old instruction then add the new one.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
     }
   }
 
-  // Erase dead blocks
-  for (auto ebi = func->begin(); ebi != func->end();)
-    if (elimBlocks.find(&*ebi) != elimBlocks.end())
+  // Erase dead blocks.
+  for (auto ebi = func->begin(); ebi != func->end();) {
+    if (unreachableMerges.count(&*ebi)) {
+      // Make unreachable, but leave the label.
```
dneto0: Might be worth saying the unreachable merge and continue blocks need to remain they can be targeted by a merge instruction.
Any other use of the label is, perforce, a branch from yet another unreachable block.  And that other unreachable block will itself either (a) disappear or (b) be emptied out and its terminator replaced with an OpUnreachable or a trivial backedge.

greg-lunarg:
Here is a pre-packageed test with a fairly simple piece of code for which this PR currently generates incorrect code. Essentially, it is an if-then with a phi in the merge block. This pattern is quite commonly generated by glslang for short-circuit boolean operations, although I believe the code in this test reasonably exemplifies the basic problem.

There might be other regressions as well. I will start with this and will re-run my regression testing once this problem is addressed.

TEST_F(DeadBranchElimTest, IfThenPhiTrue) {
  // Test handling of phi in merge block after dead branch elimination.
  // Note: The SPIR-V has had store/load elimination and phi insertion
  //
  // #version 140
  //
  // void main()
  // {
  //     vec4 v;
  //     v = vec4(0.0,0.0,0.0,0.0);
  //     if (true)
  //       v = vec4(1.0,1.0,1.0,1.0);
  //     gl_FragColor = v;
  // }

  const std::string predefs =
      R"(OpCapability Shader
OpMemoryModel Logical GLSL450
OpEntryPoint Fragment %main "main" %gl_FragColor
OpSource GLSL 140
OpName %main "main"
OpName %gl_FragColor "gl_FragColor"
%void = OpTypeVoid
%5 = OpTypeFunction %void
%bool = OpTypeBool
%true = OpConstantTrue %bool
%float = OpTypeFloat 32
%_ptr_Function_v4float = OpTypePointer Function %v4float
%float_0 = OpConstant %float 0
%12 = OpConstantComposite %v4float %float_0 %float_0 %float_0 %float_0
%float_1 = OpConstant %float 1
%14 = OpConstantComposite %v4float %float_1 %float_1 %float_1 %float_1
%_ptr_Output_v4float = OpTypePointer Output %v4float
%gl_FragColor = OpVariable %_ptr_Output_v4float Output
%_ptr_Input_v4float = OpTypePointer Input %v4float
)";

  const std::string before =
      R"(%main = OpFunction %void None %5
%17 = OpLabel
OpSelectionMerge %18 None
OpBranchConditional %true %19 %18
%19 = OpLabel
OpBranch %18
%18 = OpLabel
%20 = OpPhi %v4float %12 %19 %14 %17
OpStore %gl_FragColor %20
OpReturn
OpFunctionEnd
)";

  const std::string after =
      R"(%main = OpFunction %void None %5
%17 = OpLabel
OpBranch %19
%19 = OpLabel
OpBranch %18
%18 = OpLabel
OpStore %gl_FragColor %12
OpReturn
OpFunctionEnd
)";

  SinglePassRunAndCheck<opt::DeadBranchElimPass>(predefs + before,
                                                 predefs + after, true, true);
}

greg-lunarg:
Sorry, the GLSL in the test is backwards. Here is what it should be:

// #version 140
//
// void main()
// {
// vec4 v;
// v = vec4(1.0,1.0,1.0,1.0);
// if (true)
// v = vec4(0.0,0.0,0.0,0.0);
// gl_FragColor = v;
// }

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
-                caseVal = *idp;
-              } else {
-                if (caseVal == selVal) liveLabId = *idp;
-              }
+      if (GetConstInteger(terminator->GetSingleWordInOperand(0u), &selVal)) {
+        // Search switch operands for selector value, set liveLabId to
+        // corresponding label, use default if not found.
+        uint32_t icnt = 0;
+        uint32_t caseVal;
+        terminator->ForEachInOperand([
```
s-perron: I know this was code that was already there, but, since we are doing a large change, could we change this to a for-loop.  Plus it can avoid all of the logic with `icnt`, which I always find hard to follow.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
+          ir::BasicBlock* inc =
+              get_parent_block(inst->GetSingleWordInOperand(i));
+          if (unreachableContinues.count(inc)) {
+            // Replace incoming value with undef.
+            operands.emplace_back(
+                SPV_OPERAND_TYPE_ID,
+                std::initializer_list<uint32_t>{Type2Undef(inst->type_id())});
+            operands.push_back(inst->GetInOperand(i));
+            changed = true;
+          } else if (liveBlocks.count(inc)) {
```
s-perron: To fix the problem that Greg identified you need to check that the edge still exists, not just that the block is live.

source/opt/dead_branch_elim_pass.cpp
s-perron: I would suggest splitting it up into separate functions.  I found the function long as well.  So there would be a `FindLiveBlocks`, `MarkStructuredTargetsAsLive`, `UpdatePhis`, and `DeleteDeadBlocks`.  Or something like that.  it would make it much easier to read, at least for me.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
+      KillAllInsts(&*ebi, false);
+      // Add unreachable terminator.
+      ebi->AddInstruction(
+          MakeUnique<ir::Instruction>(context(), SpvOpUnreachable, 0, 0,
+                                      std::initializer_list<ir::Operand>{}));
+      ++ebi;
+      modified = true;
+    } else if (unreachableContinues.count(&*ebi)) {
+      // Make unreachable, but leave the label.
+      KillAllInsts(&*ebi, false);
```
s-perron: This seems inconsistent with the handling of the Phi nodes above.  You delete the original branch, and all edges related to it, and add one back in for the header.  However, when updating the Phi node, you did not delete any edges.  You just changed it to Undef.  This is fine for the header, if the continue block branches to the header right away.  In other cases, the header will need a new entry for the Phi nodes.  For other block that are still live this will leave invalid Phi nodes.

I've assumed two things:

1) The back edge is not always in the continue block, only that it post-dominates the continue block.
2) The back edge can be a conditional branch, in the continue block, and the merge block can be reached some other way.

If these are both wrong, then there should not be a problem.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
   }
 
-  // Erase dead blocks
-  for (auto ebi = func->begin(); ebi != func->end();)
-    if (elimBlocks.find(&*ebi) != elimBlocks.end())
+  // Erase dead blocks.
+  for (auto ebi = func->begin(); ebi != func->end();) {
+    if (unreachableMerges.count(&*ebi)) {
+      // Make unreachable, but leave the label.
+      KillAllInsts(&*ebi, false);
```
s-perron: It is hard to know what the false means here, even with the comment above.  Consider the suggestion inhttps://google.github.io/styleguide/cppguide.html#Implementation_Comments, "Function Argument Comments" section.

alan-baker:
Thanks for the testcase Greg.

source/opt/dead_branch_elim_pass.cpp
alan-baker: If the continue is unreachable then blocks succeeding it on the backedge path are also unreachable. I guess what's missing from the phi update for the header is to add an undef to the phi from the continue instead which ever block it got a value from previously.

This is incorrect if OpLoopMerge does not imply structured control flow though (since I removed the requirement of Shader capability).

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,211 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
 }
 
 void DeadBranchElimPass::AddBranch(uint32_t labelId, ir::BasicBlock* bp) {
+  assert(get_def_use_mgr()->GetDef(labelId) != nullptr);
   std::unique_ptr<ir::Instruction> newBranch(new ir::Instruction(
       context(), SpvOpBranch, 0, 0,
       {{spv_operand_type_t::SPV_OPERAND_TYPE_ID, {labelId}}}));
   get_def_use_mgr()->AnalyzeInstDefUse(&*newBranch);
+  context()->set_instr_block(&*newBranch, bp);
```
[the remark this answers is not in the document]
alan-baker: I'll confirm its unnecessary and remove it.

source/opt/dead_branch_elim_pass.cpp
alan-baker: I'll add a comment.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
+  auto get_parent_block = [this](uint32_t id) {
+    return context()->get_instr_block(get_def_use_mgr()->GetDef(id));
+  };
 
-    // If constant condition/selector, replace conditional branch/switch
-    // with unconditional branch and delete merge
-    uint32_t liveLabId;
-    if (br->opcode() == SpvOpBranchConditional) {
+  // Mark live blocks reachable from the entry. Simplify constant branches and
+  // switches as you proceed to limit the number of live blocks. Be careful not
```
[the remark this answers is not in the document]
alan-baker: Done

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
+  // to eliminate backedges even if they are dead, but the header is live.
+  std::unordered_map<ir::BasicBlock*, ir::BasicBlock*> continues;
```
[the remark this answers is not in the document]
alan-baker: Will comment

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
+  bool modified = false;
+  std::unordered_set<ir::BasicBlock*> liveBlocks;
```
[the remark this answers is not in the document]
alan-baker: I'll fix the local variables.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
+    ir::BasicBlock* block = stack.back();
+    stack.pop_back();
+
+    // Live blocks doubles as visited set.
+    if (!liveBlocks.insert(block).second) continue;
+
+    uint32_t contId = block->ContinueBlockIdIfAny();
+    if (contId != 0) continues[get_parent_block(contId)] = block;
+
+    ir::Instruction* terminator = &*block->tail();
```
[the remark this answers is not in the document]
alan-baker: Thanks

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +72,215 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
-    auto dbi = bi;
-    ++dbi;
-    uint32_t dLabId = (*dbi)->id();
-    while (dLabId != mergeLabId) {
-      if (!HasNonPhiNonBackedgeRef(dLabId)) {
-        // Kill use/def for all instructions and mark block for elimination
-        backedges_.erase((*dbi)->terminator());
-        KillAllInsts(*dbi);
-        elimBlocks.insert(*dbi);
+    // Don't simplify branches of continue blocks. A path from the continue to
```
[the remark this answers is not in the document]
alan-baker: Good catch, meant to say header.

source/opt/dead_branch_elim_pass.cpp
alan-baker: Updated

source/opt/dead_branch_elim_pass.cpp
alan-baker: Ok

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -364,7 +314,8 @@ Pass::Status DeadBranchElimPass::ProcessImpl() {
   ProcessFunction pfn = [this](ir::Function* fp) {
     return EliminateDeadBranches(fp);
   };
-  bool modified = ProcessEntryPointCallTree(pfn, get_module());
+  // bool modified = ProcessEntryPointCallTree(pfn, get_module());
```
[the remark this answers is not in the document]
alan-baker: Done

source/opt/dead_branch_elim_pass.cpp
alan-baker: I'll do some refactoring.

source/opt/dead_branch_elim_pass.cpp
alan-baker: Ok I'll do this.

source/opt/dead_branch_elim_pass.cpp
alan-baker: I'll add some more explanation here.

s-perron:
Here is an example of a continue target that is dead.  The continue target ends with a conditional branch.  The branch goes to the header or the merge block (which is not dead).  The Phi node in the merge bloc would have to be updated.

[continue_problem.spvasm.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/1612034/continue_problem.spvasm.zip)

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +73,268 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
-      {{spv_operand_type_t::SPV_OPERAND_TYPE_ID, {condId}},
-       {spv_operand_type_t::SPV_OPERAND_TYPE_ID, {trueLabId}},
-       {spv_operand_type_t::SPV_OPERAND_TYPE_ID, {falseLabId}}}));
-  get_def_use_mgr()->AnalyzeInstDefUse(&*newBranchCond);
-  bp->AddInstruction(std::move(newBranchCond));
-}
+bool DeadBranchElimPass::MarkLiveBlocks(
+    ir::Function* func, std::unordered_set<ir::BasicBlock*>* live_blocks) {
+  // |continues| maps the continue target to its corresponding header.
+  std::unordered_map<ir::BasicBlock*, ir::BasicBlock*> continues;
```
s-perron: Why is this a map?  Looks like a set would do.

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +73,268 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
-      uint32_t lcnt = 0;
-      uint32_t lidx = 0;
-      uint32_t icnt = 0;
-      inst->ForEachInId([&deadPreds, &icnt, &lcnt, &lidx, this](uint32_t* idp) {
-        if (icnt % 2 == 1) {
-          if (deadPreds.find(cfg()->block(*idp)) == deadPreds.end()) {
-            ++lcnt;
-            lidx = icnt - 1;
+          // Replace the phi with either a single value or a rebuilt phi.
+          // uint32_t replId;
```
dnovillo: Remnants of old code?

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +73,268 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
+          //
+          // We always have type and result id operands. So this phi has a
+          // single source if are two more operands beyond those.
```
dnovillo: s/are/there are/

source/opt/dead_branch_elim_pass.cpp
```diff
@@ -73,260 +73,268 @@ bool DeadBranchElimPass::GetConstInteger(uint32_t selId, uint32_t* selVal) {
...
+          if (operands.size() == 4) {
+            // First input data operands is at index 2.
+            uint32_t replId = operands[2u].words[0];
+            context()->ReplaceAllUsesWith(inst->result_id(), replId);
+            iter = context()->KillInst(&*inst);
+          } else {
+            // We've rewritten the operands, so first instruct the def/use
+            // manager to forget uses that in the phi before we replace them.
```
dnovillo: s/that//

include/spirv-tools/optimizer.hpp
```diff
@@ -295,9 +295,8 @@ Optimizer::PassToken CreateLocalSingleBlockLoadStoreElimPass();
 // For all phi functions in merge block, replace all uses with the id
 // corresponding to the living predecessor.
 //
-// This pass only works on shaders (guaranteed to have structured control
-// flow). Note that some such branches and blocks may be left to avoid
-// creating invalid control flow. Improving this is left to future work.
+// Note that some branches and blocks may be left to avoid creating invalid
+// control flow. Improving this is left to future work.
```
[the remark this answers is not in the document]
alan-baker: Compute kernels can still optionally declare structured constructs. In fact, it is currently the only way to specify unroll preferences and dependency information :/ I've left a TODO to improve the handling on constant conditional branches in the continue block.

source/opt/basic_block.cpp
```diff
@@ -106,6 +106,15 @@ void BasicBlock::ForEachSuccessorLabel(
   }
 }
 
+bool BasicBlock::IsSuccessor(const ir::BasicBlock* block) const {
+  uint32_t succId = block->id();
+  bool isSuccessor = false;
+  ForEachSuccessorLabel([&isSuccessor, succId](const uint32_t label) {
+    if (label == succId) isSuccessor = true;
```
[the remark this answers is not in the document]
alan-baker: ForEachSuccessor takes a function with a void return. This is a similar issue to ForEachUser in the def/use manager. We probably should have variants, but we got stuck on naming previously.

source/opt/dead_branch_elim_pass.cpp
alan-baker: Looks like.

source/opt/dead_branch_elim_pass.h
```diff
@@ -60,34 +57,71 @@ class DeadBranchElimPass : public MemPass {
...
   // and convert to a Branch to the indicated label. Delete resulting dead
-  // blocks. Assumes only structured control flow in shader. Note some such
-  // branches and blocks may be left to avoid creating invalid control flow.
-  // TODO(greg-lunarg): Remove remaining constant conditional branches and
-  // dead blocks.
+  // blocks. Note some such branches and blocks may be left to avoid creating
+  // invalid control flow.
+  // TODO(greg-lunarg): Remove remaining constant conditional branches and dead
+  // blocks.
   bool EliminateDeadBranches(ir::Function* func);
```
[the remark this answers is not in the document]
alan-baker: Yes, they should be merged, but I'd rather put that in a separate PR.

source/opt/dead_branch_elim_pass.cpp
alan-baker: Thanks

alan-baker: Ok

source/opt/dead_branch_elim_pass.cpp
alan-baker: You're right. I think I originally made it a map assuming I'd need the mapped value, but never ended up using it.

source/opt/dead_branch_elim_pass.h
```diff
@@ -60,34 +57,71 @@ class DeadBranchElimPass : public MemPass {
...
+      ir::Function* func,
+      const std::unordered_set<ir::BasicBlock*>& live_blocks,
+      const std::unordered_map<ir::BasicBlock*, ir::BasicBlock*>&
+          unreachable_continues);
+
+  // Erases dead blocks. Any block captured in |unreachable_merges| or
+  // |unreachable_continues| is a dead block that is required to remain due to
+  // a live merge instruction in the corresponding header. These blocks will
+  // have their instructions clobbered and will become a label and terminator.
+  // Unreachable merge blocks are terminated by OpReachable, while unreachable
```
dneto0: OpUnreachable

source/opt/dead_branch_elim_pass.h
```diff
@@ -60,34 +57,71 @@ class DeadBranchElimPass : public MemPass {
...
+  // continue blocks are terminated by an unconditional branch to the header.
+  // Otherwise, blocks are dead if not explicitly captured in |live_blocks| and
+  // are totally removed.
+  //
+  // |unreachable_continues| maps continue targets that cannot be reached to
+  // merge instruction that declares them.
```
dneto0: "to the corresponding header block" rather than "to merge instruction"

test/opt/dead_branch_elim_test.cpp
```diff
@@ -1346,6 +1346,569 @@ OpFunctionEnd
...
+OpSelectionMerge %14 None
+OpBranchConditional %false %5 %6
+%5 = OpLabel
+%10 = OpLogicalAnd %bool %true %false
+OpBranch %7
+%6 = OpLabel
+%11 = OpLogicalOr %bool %true %false
+OpBranch %7
+%7 = OpLabel
+; This phi is in a block preceeding the merge %14!
```
dneto0: Good

source/opt/dead_branch_elim_pass.h
```diff
@@ -60,34 +57,71 @@ class DeadBranchElimPass : public MemPass {
...
+  void MarkUnreachableStructuredTargets(
+      const std::unordered_set<ir::BasicBlock*>& live_blocks,
+      std::unordered_set<ir::BasicBlock*>* unreachable_merges,
+      std::unordered_map<ir::BasicBlock*, ir::BasicBlock*>*
+          unreachable_continues);
+
+  // Fix phis in reachable blocks so that only live (or unremovable) incoming
+  // edges are present. If the block now only has a single live incoming edge,
+  // remove the phi and replace its uses with its data input.
```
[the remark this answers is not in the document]
dneto0: Maybe the method comment should be updated to address this situation.  How about adding:

"If the single remaining incoming edge is from the phi itself, then the phi is in an unreachable single-block-loop.  Either the block is dead and will be removed, or it's reachable from an unreachable continue target.  In the latter case that continue target block will be collapsed into a block that only branches back to its header and we'll eliminate the block with the phi."

greg-lunarg:
It has passed my regression testing.

dneto0:
Rebased, squashed, and pushed into master as 1b6cfd34092ad

PR #1179, opened 2018-01-07
Allow relaxing validation of pointers in logical addressing mode

antiagainst:
In HLSL structured buffer legalization, pointer to pointer types
are emitted to indicate a structured buffer variable should be
treated as an alias of some other variable. We need an option to
relax the check of pointer types in logical addressing mode to
catch other validation errors.


dneto0:
Rebased and pushed into master as 44f27f928

PR #1180, opened 2018-01-08
Fix method comment for BasicBlock::MegeBlockIdIfAny

### LINKED ISSUE #1177: BasicBlock::ContinueBlockIdIfAny should exit early if it reaches an OpSelectionMerge
All uses of ContinueBlockIdIfAny immediately follow a use of MergeBlockIdIfAny. So a selection block will always be fully scanned when it can exit early.
It can exit early because OpSelectionMerge and OpLoopMerge can never be used in the same block.

dneto0:
Fixes #1177


dneto0:
The Travis failure is failure is an infrastructure failure: can't download the code.

PR #1181, opened 2018-01-08
Fix constant propagation of induction variables.

dnovillo:
This fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1143.
When an instruction transitions from constant to bottom (varying) in the
lattice, we were telling the propagator that the instruction was
varying, but never updating the actual value in the values table.

This led to incorrect value substitutions at the end of propagation.

### DIFF
--- a/source/opt/ccp_pass.cpp
+++ b/source/opt/ccp_pass.cpp
@@ -27,6 +27,25 @@
 namespace spvtools {
 namespace opt {
 
+namespace {
+
+// This SSA id is never defined nor referenced in the IR.  It is a special ID
+// which represents varying values.  When an ID is found to have a varying
+// value, its entry in the |values_| table maps to kVaryingSSAId.
+const uint32_t kVaryingSSAId = std::numeric_limits<uint32_t>::max();
+
+}  // namespace
+
+bool CCPPass::IsVaryingValue(uint32_t id) const { return id == kVaryingSSAId; }
+
+SSAPropagator::PropStatus CCPPass::MarkInstructionVarying(
+    ir::Instruction* instr) {
+  assert(instr->result_id() != 0 &&
+         "Instructions with no result cannot be marked varying.");
+  values_[instr->result_id()] = kVaryingSSAId;
+  return SSAPropagator::kVarying;
+}
+
 SSAPropagator::PropStatus CCPPass::VisitPhi(ir::Instruction* phi) {
   uint32_t meet_val_id = 0;
 
@@ -52,9 +71,10 @@ SSAPropagator::PropStatus CCPPass::VisitPhi(ir::Instruction* phi) {
         // looking.
         continue;
       } else {
-        // We found another constant value, but it is different from the
-        // previous computed meet value.  This Phi will never be constant.
-        return SSAPropagator::kVarying;
+        // We either found a varying value, or another constant value different
+        // from the previous computed meet value.  This Phi will never be
+        // constant.
+        return MarkInstructionVarying(phi);
       }
     } else {
       // If any argument is not a constant, the Phi produces nothing
@@ -85,24 +105,33 @@ SSAPropagator::PropStatus CCPPass::VisitAssignment(ir::Instruction* instr) {
     uint32_t rhs_id = instr->GetSingleWordInOperand(0);
     auto it = values_.find(rhs_id);
     if (it != values_.end()) {
-      values_[instr->result_id()] = it->second;
-      return SSAPropagator::kInteresting;
+      if (IsVaryingValue(it->second)) {
+        return MarkInstructionVarying(instr);
+      } else {
+        values_[instr->result_id()] = it->second;
+        return SSAPropagator::kInteresting;
+      }
     }
     return SSAPropagator::kNotInteresting;
   }
 
   // Instructions with a RHS that cannot produce a constant are always varying.
   if (!instr->IsFoldable()) {
-    return SSAPropagator::kVarying;
+    return MarkInstructionVarying(instr);
   }
 
   // Otherwise, see if the RHS of the assignment folds into a constant value.
   std::vector<uint32_t> cst_val_ids;
   bool missing_constants = false;
-  instr->ForEachInId([this, &cst_val_ids, &missing_constants](uint32_t* op_id) {
+  bool varying_values = false;
+  instr->ForEachInId([this, &cst_val_ids, &missing_constants,
+                      &varying_values](uint32_t* op_id) {
     auto it = values_.find(*op_id);
     if (it == values_.end()) {
       missing_constants = true;
+      return;
+    } else if (IsVaryingValue(it->second)) {
+      varying_values = true;
       return;
     }
     cst_val_ids.push_back(it->second);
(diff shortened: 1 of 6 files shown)


test/opt/ccp_test.cpp
```diff
@@ -464,6 +464,60 @@ TEST_F(CCPTest, SSAWebCycles) {
...
+
+; This Phi should not have all constant arguments:
+; CHECK: OpPhi %int %int_0 {{%\d+}} {{%\d+}} {{%\d+}}
+         %22 = OpPhi %int %int_0 %12 %21 %15
+               OpLoopMerge %14 %15 None
+               OpBranch %16
+         %16 = OpLabel
+
+; The Phi should never be considered to have the value %int_0.
+; CHECK: OpSLessThan %bool {{%\d+}} %int_10
```
alan-baker: It might be clearer to capture the result id of the phi above and ensure it is used in the less than.

test/opt/ccp_test.cpp
```diff
@@ -464,6 +464,60 @@ TEST_F(CCPTest, SSAWebCycles) {
...
+         %18 = OpSLessThan %bool %22 %int_10
+
+; This conditional was wrongly converted into an always-true jump due to the
+; bad meet evaluation of %27.
+; CHECK: OpBranchConditional {{%\d+}} {{%\d+}} {{%\d+}}
```
alan-baker: Same here, capture the less than result and ensure its the condition of the branch.

source/opt/ccp_pass.h
```diff
@@ -67,6 +67,13 @@ class CCPPass : public MemPass {
   // otherwise.
   bool ReplaceValues();
 
+  // Marks |instr| as varying by registering a varying value for its result
+  // into the |values_| table. Returns SSAPropagator::kVarying.
+  SSAPropagator::PropStatus MarkInstructionVarying(ir::Instruction* instr);
+
+  // Returns true if |id| is a varying value.
```
alan-baker: The parameter name seem slightly confusing. |id| is not a SSA id, but I think that would the intuition. Suggest changing it to |val|.

source/opt/ccp_pass.cpp
```diff
@@ -27,6 +27,18 @@
 namespace spvtools {
 namespace opt {
 
+bool CCPPass::IsVaryingValue(uint32_t id) const {
+  return id == std::numeric_limits<uint32_t>::max();
```
dneto0: You should use a named constant value for this, to document and enforce consistency with MarkInstructionVarying.

E.g.   enum { kVaryingValueSentinelId = std::numeric_limits<uint32_t>::max(); };

source/opt/ccp_pass.h
dneto0: Recommend renaming the method to IsVaryingValueId.
And the method comment should be 

// Returns true if |id| is the sentinel id indicating a varying value.  It is never a valid SSA id.


OR: delete the method and test against kVaryingSentinelValueId everywhere.  That is probably less readable, though.

test/opt/ccp_test.cpp
```diff
@@ -464,6 +464,60 @@ TEST_F(CCPTest, SSAWebCycles) {
...
+               OpLoopMerge %14 %15 None
+               OpBranch %16
+         %16 = OpLabel
+
+; The Phi should never be considered to have the value %int_0.
+; CHECK: OpSLessThan %bool {{%\d+}} %int_10
+         %18 = OpSLessThan %bool %22 %int_10
+
+; This conditional was wrongly converted into an always-true jump due to the
+; bad meet evaluation of %27.
```
dneto0: I think you mean %22 here.  The example has changed since you first looked at it.

test/opt/ccp_test.cpp
```diff
@@ -464,6 +464,60 @@ TEST_F(CCPTest, SSAWebCycles) {
...
+       %bool = OpTypeBool
+      %int_1 = OpConstant %int 1
+       %main = OpFunction %void None %5
+         %12 = OpLabel
+               OpBranch %13
+         %13 = OpLabel
+
+; This Phi should not have all constant arguments:
+; CHECK: OpPhi %int %int_0 {{%\d+}} {{%\d+}} {{%\d+}}
+         %22 = OpPhi %int %int_0 %12 %21 %15
```
dneto0: Should test like this:

```
; CHECK  [[phi:%\w+]] = OpPhi %int %int_0 {{%\w+}} [[induction_var:%\w+]] {{\d+}}
```
Then later as @alan-baker suggests, use [[phi]] in the OpSLessThan.
Also later add
```
; CHECK: [[induction_var]] = OpIAdd %int [[phi]] %int_1
```

source/opt/ccp_pass.h
```diff
@@ -67,6 +67,13 @@ class CCPPass : public MemPass {
...
+  bool IsVaryingValue(uint32_t id) const;
+
   // Constant manager for the parent IR context.  Used to record new constants
   // generated during propagation.
   analysis::ConstantManager* const_mgr_;
```
dneto0: You must also update the comment for the values_ member. Should say that an SSA id can map to the kVaryingValueSentinelId to indicate that the id has been proven to be varying.

source/opt/propagator.cpp
```diff
@@ -37,22 +37,28 @@ void SSAPropagator::AddControlEdge(const Edge& edge) {
...
       instr->result_id(), [this](ir::Instruction* use_instr) {
         // If the basic block for |use_instr| has not been simulated yet, do
-        // nothing.
+        // nothing.  The instruction |use_instr| will be simulated next time the
+        // block is scheduled.
         if (!BlockHasBeenSimulated(ctx_->get_instr_block(use_instr))) {
           return;
         }
 
+        // If |use_instr| is a Phi, ignore this edge.  Phi instructions can form
```
dneto0: I'd probably put this first in the sequence of tests.  It might save some time.

test/opt/ccp_test.cpp
dnovillo: Done.

test/opt/ccp_test.cpp
dnovillo: Done.

source/opt/ccp_pass.h
dnovillo: Done.

source/opt/propagator.cpp
dnovillo: Done.

source/opt/ccp_pass.cpp
```diff
@@ -87,23 +107,30 @@ SSAPropagator::PropStatus CCPPass::VisitAssignment(ir::Instruction* instr) {
     if (it != values_.end()) {
       values_[instr->result_id()] = it->second;
       return SSAPropagator::kInteresting;
+    } else if (IsVaryingValue(it->second)) {
```
s-perron: This looks odd.  If `it == values_.end()` then load `it->second`?

dnovillo: Fixed, thanks.  The case can actually never happen because the propagator is guaranteed to always have generated a value of the RHS of an OpCopyObject (it works in def-use and dominator order, after all).  However, it would've failed to mark the LHS varying if the RHS was varying and would mark the instruction not-interesting, causing an additional go around the propagator.

dnovillo:
Re-based and pushed to head as e5560d64.

Issue #1183, opened 2018-01-09
PackDouble2x32 is mis-validated


HansKristian-ARM:
Getting an error:
```
error: 249: GLSL.std.450 PackDouble2x32: expected operand V to be a 32-bit float vector of size 2
```
but spec says:
```
The v operand must be a vector of 2 components whose type is a 32-bit integer.
```

dneto0:
Fixed by #1186

PR #1186, opened 2018-01-09
Fix validation rules for GLSL.450.std Pack/UnpackDouble2x32

atgoo:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1183


dneto0:
The Travis failure was infrastructure unable to download sources.

dneto0:
Rebased and pushed into master as d54a286c75

Issue #1164, opened 2018-01-04
Folding non-constant expressions


s-perron:
Many times there are expressions that can be simplified even if not all of the operands are constants.  A few of these are:

```
n+0
n*0
n*1
n / 1
b && true
b && false
b || true
b || false
```
It can be useful to simplify these, especially if they fold to a constant like `b && false`.

This issue will have a few different parts to it:

- [x] Write a utility that folds any instruction if possible.
- [x] Make use of this new utility in CCP
- [x] Make use of this new utility in GVN

dnovillo:
This is already reported in #896.  I'll assign it to you, if you're ready to start working on this.

s-perron:
We have done all of the main parts of this.  We still have lots of folding rules we can add, but separate issues should be opened for those that are needed.

PR #1188, opened 2018-01-09
Add generic folding function and use in CCP

s-perron:
The current folding routines have a very cumbersome interface, make them
harder to use, and not a obvious how to extend.

This change is to create a new interface for the folding routines, and
show how it can be used by calling it from CCP.

This does not make a significant change to the behaviour of CCP.  In
general it should produce the same code as before; however it is
possible that an instruction that takes 32-bit integers as inputs and
the result is not a 32-bit integer or bool will not be folded as before.

Contributes to #1164.

### DIFF
--- a/source/opt/ccp_pass.cpp
+++ b/source/opt/ccp_pass.cpp
@@ -120,58 +120,41 @@ SSAPropagator::PropStatus CCPPass::VisitAssignment(ir::Instruction* instr) {
     return MarkInstructionVarying(instr);
   }
 
-  // Otherwise, see if the RHS of the assignment folds into a constant value.
-  std::vector<uint32_t> cst_val_ids;
-  bool missing_constants = false;
-  bool varying_values = false;
-  instr->ForEachInId([this, &cst_val_ids, &missing_constants,
-                      &varying_values](uint32_t* op_id) {
+  // See if the RHS of the assignment folds into a constant value.
+  auto map_func = [this](uint32_t id) {
+    auto it = values_.find(id);
+    if (it == values_.end() || IsVaryingValue(it->second)) {
+      return id;
+    }
+    return it->second;
+  };
+  ir::Instruction* folded_inst =
+      opt::FoldInstructionToConstant(instr, map_func);
+  if (folded_inst != nullptr) {
+    // We do not want to change the body of the function by adding new
+    // instructions.  When folding we can only generate new constants.
+    assert(folded_inst->IsConstant() && "CCP is only interested in constant.");
+    values_[instr->result_id()] = folded_inst->result_id();
+    return SSAPropagator::kInteresting;
+  }
+
+  // If not, see if there is a least one unknown operand to the instruction.  If
+  // so, we might be able to fold it later.
+  bool could_be_improved = false;
+  instr->ForEachInId([this, &could_be_improved](uint32_t* op_id) {
     auto it = values_.find(*op_id);
     if (it == values_.end()) {
-      missing_constants = true;
-      return;
-    } else if (IsVaryingValue(it->second)) {
-      varying_values = true;
+      could_be_improved = true;
       return;
     }
-    cst_val_ids.push_back(it->second);
   });
-
-  // If we did not find a constant value for every operand in the instruction,
-  // do not bother folding it.  Indicate that this instruction does not produce
-  // an interesting value for now.
-  if (missing_constants) {
+  if (could_be_improved) {
     return SSAPropagator::kNotInteresting;
   }
 
-  // If we found at least one varying value, the instruction will never fold
-  // into anything interesting.  Mark it varying.
-  if (varying_values) {
-    return MarkInstructionVarying(instr);
-  }
-
-  auto constants = const_mgr_->GetConstantsFromIds(cst_val_ids);
-  assert(constants.size() != 0 && "Found undeclared constants");
-
-  // If any of the constants are not supported by the folder, we will not be
-  // able to produce a constant out of this instruction.  Consider it varying
-  // in that case.
-  if (!std::all_of(constants.begin(), constants.end(),
-                   [](const analysis::Constant* cst) {
-                     return IsFoldableConstant(cst);
-                   })) {
-    return MarkInstructionVarying(instr);
-  }
-
-  // Otherwise, fold the instruction with all the operands to produce a new
-  // constant.
-  uint32_t result_val = FoldScalars(instr->opcode(), constants);
-  const analysis::Constant* result_const =
-      const_mgr_->GetConstant(const_mgr_->GetType(instr), {result_val});
-  ir::Instruction* const_decl =
-      const_mgr_->GetDefiningInstruction(result_const);
-  values_[instr->result_id()] = const_decl->result_id();
-  return SSAPropagator::kInteresting;
+  // Otherwise, we will never be able to fold this instruction, so mark it
+  // varying.
+  return MarkInstructionVarying(instr);
 }
 
 SSAPropagator::PropStatus CCPPass::VisitBranch(ir::Instruction* instr,
--- a/source/opt/fold.cpp
+++ b/source/opt/fold.cpp
@@ -14,6 +14,7 @@
 
 #include "fold.h"
 #include "def_use_manager.h"
+#include "ir_context.h"
 
 #include <cassert>
 #include <vector>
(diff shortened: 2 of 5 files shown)


source/opt/fold.h
```diff
@@ -59,6 +59,29 @@ bool IsFoldableOpcode(SpvOp opcode);
 // Returns true if |cst| is supported by FoldScalars and FoldVectors.
 bool IsFoldableConstant(const analysis::Constant* cst);
 
+// Returns true if |FoldInstruction| will fold an instruction whose result type
```
dneto0: suggest changing "will" to "could".  It's more accurate, and it's in line with the method comment on IsFoldable.

source/opt/instruction.h
```diff
@@ -347,6 +347,8 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   Instruction* InsertBefore(std::unique_ptr<Instruction>&& i);
   using utils::IntrusiveNodeBase<Instruction>::InsertBefore;
 
+  inline bool IsConstant() const;
```
dneto0: Should have a method comment. 
You want to include OpContstant* but exclude OpSpecConstant*

(The SPIR-V spec includes spec constant instructions in "constant instructions", so it's best to be crystal clear here.)

source/opt/fold.h
```diff
@@ -59,6 +59,29 @@ bool IsFoldableOpcode(SpvOp opcode);
...
+// is |type_inst|.
+bool IsFoldableType(ir::Instruction* type_inst);
+
+// Tries to use a simpler instruction to compute the result of |inst|, when the
```
dneto0: Suggest "create" instead of "user" here.

source/opt/ccp_pass.cpp
```diff
@@ -120,58 +120,37 @@ SSAPropagator::PropStatus CCPPass::VisitAssignment(ir::Instruction* instr) {
...
-  // Otherwise, see if the RHS of the assignment folds into a constant value.
-  std::vector<uint32_t> cst_val_ids;
-  bool missing_constants = false;
-  bool varying_values = false;
-  instr->ForEachInId([this, &cst_val_ids, &missing_constants,
-                      &varying_values](uint32_t* op_id) {
+  // See if the RHS of the assignment folds into a constant value.
+  auto map_func = [this](uint32_t id) {
+    auto it = values_.find(id);
+    if (it == values_.end() || it->second == kVaryingSSAId) {
```
alan-baker: Use IsVaryingValue(it->second) in the latter half of the expression.

source/opt/fold.cpp
```diff
@@ -289,5 +290,54 @@ bool IsFoldableConstant(const analysis::Constant* cst) {
     return cst->AsNullConstant() != nullptr;
 }
 
+ir::Instruction* FoldInstruction(
+    ir::Instruction* inst,
+    std::function<uint32_t(uint32_t)> id_map ) {
+  if (!IsFoldableOpcode(inst->opcode())) {
```
dneto0: Why is this not `if (!inst->IsFoldable()) { return false; }`

Suggest a comment here.

source/opt/fold.cpp
```diff
@@ -289,5 +290,54 @@ bool IsFoldableConstant(const analysis::Constant* cst) {
...
+    return nullptr;
+  }
+
+  ir::IRContext* context = inst->context();
+  analysis::ConstantManager* const_mgr = context->get_constant_mgr();
+
+  std::vector<const analysis::Constant*> constants;
```
dneto0: I'd add a comment for this phase:  
`// Try to find or create an equivalent constant value.`

source/opt/fold.cpp
```diff
@@ -289,5 +290,54 @@ bool IsFoldableConstant(const analysis::Constant* cst) {
...
+      });
+
+  if (!missing_constants) {
+    uint32_t result_val = FoldScalars(inst->opcode(), constants);
+    const analysis::Constant* result_const =
+        const_mgr->GetConstant(const_mgr->GetType(inst), {result_val});
+    return const_mgr->GetDefiningInstruction(result_const);
+  }
+
+  // Check for constant parameters.
```
dneto0: This comment is a bit confusing.

I'd add a TODO, e.g. 
`// TODO: Other folding cases, such as arithmetic and logic absorption.`

source/opt/fold.cpp
```diff
@@ -289,5 +290,54 @@ bool IsFoldableConstant(const analysis::Constant* cst) {
...
+  return nullptr;
+}
+
+bool IsFoldableType(ir::Instruction* type_inst) {
+  // Support 32-bit integers.
+  if (type_inst->opcode() == SpvOpTypeInt) {
+    return type_inst->GetSingleWordInOperand(0) == 32;
```
alan-baker: Strictly speaking literal operands maybe more than one word. Perhaps assert that there is only one word or check that all other words are 0?

source/opt/ccp_pass.cpp
```diff
@@ -120,58 +120,37 @@ SSAPropagator::PropStatus CCPPass::VisitAssignment(ir::Instruction* instr) {
...
+      return id;
+    }
+    return it->second;
+  };
+  ir::Instruction* folded_inst = opt::FoldInstruction(instr, map_func);
+  if (folded_inst != nullptr && folded_inst->IsConstant()) {
```
dneto0: What if we improve the folder to generate non-constant instructions. Then we would leak the folded_inst in those cases.
For example, what if the folder implements a form of strength reduction.

source/opt/ccp_pass.cpp
```diff
@@ -120,58 +120,37 @@ SSAPropagator::PropStatus CCPPass::VisitAssignment(ir::Instruction* instr) {
...
-  uint32_t result_val = FoldScalars(instr->opcode(), constants);
-  const analysis::Constant* result_const =
-      const_mgr_->GetConstant(const_mgr_->GetType(instr), {result_val});
-  ir::Instruction* const_decl =
-      const_mgr_->GetDefiningInstruction(result_const);
-  values_[instr->result_id()] = const_decl->result_id();
-  return SSAPropagator::kInteresting;
+  // Otherwise, we will never be able to fold this instruction, so mark it
+  // varying.
+  return MarkInstructionVarying(instr);
```
dneto0: This seems really aggressive.

source/opt/fold.cpp
dnovillo: Shouldn't this be !inst->IsFoldable()?  If not, it's inconsistent with the predicate that protects it in CCP.

source/opt/fold.cpp
s-perron: For OpTypeInt, the width is limited to a single word. https://www.khronos.org/registry/spir-v/specs/1.2/SPIRV.html#OpTypeInt

source/opt/ccp_pass.cpp
s-perron: I had thought of this a little.  The problem is assessing who is the owner of the instruction pointer to by `folded_inst`.  It cannot be `folded_inst` because most of the time we will be returning a value that already exists in the code, like an OpConstant.  Even though I did not document it, if `FoldInstruction` creates a new instruction, it will have to link it into the code somewhere so that it is owned by a basic block or other list.  I did not document that because I am still unsure how to decide which block it belongs to, should there be a requirement that the input instruction be in a list, and the new instructions be placed in that list?

I could say that for now, `FoldInstruction` will not create a new instruction, but that will probably change soon enough.

Thanks for bring this up, because I forgot to.  It should be discussed.

source/opt/ccp_pass.cpp
s-perron: This is less aggressive than it was before.  The previous code would say the instruction was varying if there was a least 1 input operand that was varying.  

Now, we will mark it as varying if every operand has moved from the top of the lattice (not in values_), and we were unable to fold.  At this point, things can only get worse for folding.  A constant will becoming bottom (varying).  If we can't fold to a constant now, then we will not be able to fold later.  So we might as well go straight to the bottom (varying).

source/opt/ccp_pass.cpp
s-perron: I misspoke.  Current if a new instruction is created it is a constant, and it is added to the section that holds all of the constants.  When we create other instructions, it will become more complicated.

source/opt/ccp_pass.cpp
s-perron: done

source/opt/fold.cpp
s-perron: done.

source/opt/fold.cpp
s-perron: Done.

source/opt/fold.cpp
s-perron: Done.

source/opt/fold.h
s-perron: Done.

source/opt/fold.cpp
s-perron: done.

source/opt/instruction.h
s-perron: Done.

s-perron:
I think I have responded to all of the comment, except for the discussion on the owner of the created instructions.

source/opt/ccp_pass.cpp
dneto0: This is correct but fragile.   I'd recommend being a little more defensive:

```
if (folded_inst) {
  if (folded_inst->IsConstant()) { ... }
  else { 
    assert and a TODO.
  }
}
```
But I can put that in in a subsequent CL.

s-perron:
Should be ready for another round of reviews on the interface.  I should add some test for what already exists.

source/opt/ccp_pass.cpp
s-perron: I've updated the folding routines, and created two version.  One that will only return constants, which is called from here.  I've added an assert to make sure the folded instruction is a constant.

s-perron:
I'll merge this change, start working on adding new simplifications, and create a good way of testing the folding instructions.

PR #1189, opened 2018-01-10
InsertExtractElim: Optimize through VectorShuffle, Mix

greg-lunarg:
This improves Extract replacement to continue through VectorShuffle.
It will also continue through Mix with 0.0 or 1.0 in the a-value of the desired
component.

To facilitate optimization of VectorShuffle, the algorithm was refactored
to pass around the indices of the extract in a vector rather than pass the
extract instruction itself. This allows the indices to be modified as the
algorithm progresses.

These two optimizations were necessary to remove significant dead code
(including dead samples from an unbound texture) from an HLSL shader
in a real-world Vulkan app ported from DX. Failure to remove the dead
samples caused validation layer warnings. These optimizations are required
for fxc equivalence.

### DIFF
--- a/source/opt/insert_extract_elim.cpp
+++ b/source/opt/insert_extract_elim.cpp
@@ -36,6 +36,7 @@ const uint32_t kVectorShuffleVec2IdInIdx = 1;
 const uint32_t kVectorShuffleCompsInIdx = 2;
 const uint32_t kTypeVectorCompTypeIdInIdx = 0;
 const uint32_t kTypeVectorLengthInIdx = 1;
+const uint32_t kTypeFloatWidthInIdx = 0;
 const uint32_t kExtInstSetIdInIdx = 0;
 const uint32_t kExtInstInstructionInIdx = 1;
 const uint32_t kFMixXIdInIdx = 2;
@@ -145,6 +146,11 @@ uint32_t InsertExtractElimPass::DoExtract(ir::Instruction* compInst,
         break;
       ir::Instruction* a_comp_inst = get_def_use_mgr()->GetDef(a_comp_id);
       if (a_comp_inst->opcode() != SpvOpConstant)
+        break;
+      // If a value is not 32-bit, give up
+      uint32_t a_comp_type_id = a_comp_inst->type_id();
+      ir::Instruction* a_comp_type = get_def_use_mgr()->GetDef(a_comp_type_id);
+      if (a_comp_type->GetSingleWordInOperand(kTypeFloatWidthInIdx) != 32)
         break;
       uint32_t u = a_comp_inst->GetSingleWordInOperand(kConstantValueInIdx);
       float* fp = reinterpret_cast<float*>(&u);
--- a/source/opt/insert_extract_elim.cpp
+++ b/source/opt/insert_extract_elim.cpp
@@ -45,15 +45,13 @@ const uint32_t kFMixAIdInIdx = 4;
 
 }  // anonymous namespace
 
-bool InsertExtractElimPass::ExtInsMatch(
-    const std::vector<uint32_t>& extIndices, const ir::Instruction* insInst,
-    const uint32_t extOffset) const {
+bool InsertExtractElimPass::ExtInsMatch(const std::vector<uint32_t>& extIndices,
+                                        const ir::Instruction* insInst,
+                                        const uint32_t extOffset) const {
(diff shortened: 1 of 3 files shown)


source/opt/insert_extract_elim.cpp
```diff
@@ -63,6 +77,124 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
...
+      uint32_t a_id = cinst->GetSingleWordInOperand(kFMixAIdInIdx);
+      ir::Instruction* a_inst = get_def_use_mgr()->GetDef(a_id);
+      uint32_t a_comp_id = DoExtract(a_inst, &aIndices, 0);
+      if (a_comp_id == 0)
+        break;
+      ir::Instruction* a_comp_inst = get_def_use_mgr()->GetDef(a_comp_id);
+      if (a_comp_inst->opcode() != SpvOpConstant)
+        break;
+      uint32_t u = a_comp_inst->GetSingleWordInOperand(kConstantValueInIdx);
+      float* fp = reinterpret_cast<float*>(&u);
```
s-perron: The specification for FMix says that the operands are floating point type, but does not specify their width.  You might need to add a check that it is 32-bit.

source/opt/insert_extract_elim.cpp
```diff
@@ -63,6 +77,124 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
...
+        cinst->opcode() == SpvOpConstantComposite) {
+      uint32_t vec_comp_type_id =
+        ctype_inst->GetSingleWordInOperand(kTypeVectorCompTypeIdInIdx);
+      if (compIdx < cinst->NumInOperands()) {
+        uint32_t i = 0;
+        for (; i <= compIdx; i++) {
+          uint32_t compId = cinst->GetSingleWordInOperand(i);
+          ir::Instruction* componentInst = get_def_use_mgr()->GetDef(compId);
+          if (componentInst->type_id() != vec_comp_type_id)
+            break;
```
s-perron: Is this case that is being looked for legal?  If something is of type vector everything must be the same type.  I guess I am  unclear of what "vector composition" means.

source/opt/insert_extract_elim.h
```diff
@@ -40,24 +40,29 @@ class InsertExtractElimPass : public Pass {
   Status Process(ir::IRContext*) override;
 
  private:
-  // Return true if indices of extract |extInst| starting at |extOffset|
+  // Return true if indices in |extIndices| starting at |extOffset|
   // match indices of insert |insInst|.
```
s-perron: It is also not clear how `extIndices` is to be interpreted.  I was able for figure it out because I knew it was replacing the indices from in an extract instruction.  Might be worth making the comments a little clearer.

source/opt/insert_extract_elim.cpp
greg-lunarg: K.

test/opt/insert_extract_elim_test.cpp
```diff
@@ -539,6 +539,270 @@ OpFunctionEnd
...
+OpStore %OutColor %23
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndCheck<opt::InsertExtractElimPass>(
+      predefs + before, predefs + after, true, true);
+}
+
+TEST_F(InsertExtractElimTest, VectorShuffle2) {
```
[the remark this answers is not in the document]
greg-lunarg: K

source/opt/insert_extract_elim.cpp
greg-lunarg: This chunk of code was designed to handle both OpCompositeConstruct and OpConstantComposite. They are identical except that OpCompositeConstruct allows one additional construction: if constructing a vector, the constituents can be vectors with the same component type as the result of the OpCompositeConstruct. For the moment we are just bailing when we see that, but technically this can be optimized through.

source/opt/insert_extract_elim.h
greg-lunarg: K

greg-lunarg:
I think I have addressed all outstanding issues. Thanks for the reviews so far!

s-perron:
I'll merge this tonight if nobody else has any comments.

s-perron:
Rebased, squashed, and merged as 5eafc00a.

Issue #988, opened 2017-11-21
Add pretty-printing (disassembling) for instructions, basic blocks and operands.


dnovillo:

dneto0:
@alan-baker discovered some code in the test tree that disassembles a single instruction. But it fails on a single OpConstant instruction, likely because there is no definition of the type id.

The right way to do partial-disassembly is to specialize the full-module disassembler so you still run through the entire module but only pay attention to some of the strings coming out for the instructions you're interested in.

dneto0:
Some more on this:  Yes, you need the information from earlier in the module, especially for type information.   So you need to send in the entire module, and then watch for the particular instruction that you want to disassemble.  You could specify that as a word offset from the start of the module, but you probably don't have that offset information handy.   

Instead, you could look for the sequence of binary words that make up the instruction, and look for a match that is instruction-aligned. I think it's true that two instructions with exactly the same sequence of words should always map to the same disassembly string.  So it doesn't matter if there are two copies of something like OpBranchConditional %c %truebranch %falsebranch because they should disassemble to the string. (Note that instructions that return a result id are always unique'd since IDs live in a global number space.)

The other thing that helps get to a solution is that spvtools::ir::Instruction now holds onto its IRContext, and so can find its containing module.  So it has full information to reproduce its own disassembly.

alan-baker:
Rebased, squashed and committed as 672494da13d54cf72d41edc67ecc6b53c561afc6.

PR #1190, opened 2018-01-10
Adding ostream operators for IR structures

alan-baker:
Addresses #988.

* Added for Instruction, BasicBlock, Function and Module
* Uses new disassembly functionality that can disassemble individual
instructions
 * For debug use only (no caching is done)
 * Each output converts module to binary, parses and outputs an
 individual instruction
* Added a test for whole module output


CMakeLists.txt
```diff
@@ -85,7 +85,7 @@ if(${COMPILER_IS_LIKE_GNU})
     set(SPIRV_WARNINGS ${SPIRV_WARNINGS} -Werror)
   endif()
 elseif(MSVC)
-  set(SPIRV_WARNINGS -D_CRT_SECURE_NO_WARNINGS /wd4800)
+  set(SPIRV_WARNINGS -D_CRT_SECURE_NO_WARNINGS -D_SCL_SECURE_NO_WARNINGS /wd4800)
```
s-perron: Can you explain why you need to disable the warning?  Not in the code, but in a comment in the PR.  I don't know what the warning is, so I have not idea if this is a good idea or not.

source/disassemble.cpp
```diff
@@ -352,6 +352,56 @@ spv_result_t DisassembleInstruction(
   return disassembler->HandleInstruction(*parsed_instruction);
 }
 
+// Simple wrapper class to provide extra data necessary for targetted
```
s-perron: Targeted does not use the double t.

CMakeLists.txt
dneto0: Would be nice to document why this was needed.

alan-baker: Windows issues warning C4996 (and we have warnings as errors) for the std::equal I added to compare uint32_t's. It suggests using a version of std::equal that requires c++14 or a Windows specific extension. The warning is due to unchecked iterators (from the pointer increments). Linux and Mac are ok with this code and we only use c++11 so I can't use the newer version of std::equal.

alan-baker: I should also mention the other define there "_CRT_SECURE_NO_WARNINGS" generates the same warning number in Visual Studio.

source/opt/instruction.cpp
```diff
@@ -470,5 +470,36 @@ bool Instruction::IsOpaqueType() const {
...
+  std::vector<uint32_t> module_binary;
+  inst.context()->module()->ToBinary(&module_binary, /* skip_nop = */ false);
+
+  // Convert the instruction to binary. This is used to identify the correct
+  // stream of words to output from the module.
+  std::vector<uint32_t> inst_binary;
+  inst.ToBinaryWithoutAttachedDebugInsts(&inst_binary);
+
+  // Do not generate a header.
+  uint32_t options = SPV_BINARY_TO_TEXT_OPTION_NO_HEADER;
```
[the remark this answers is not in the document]
alan-baker: Interesting. From a debugging perspective I feel the raw ids are more valuable. How about a refactor as follows?

// In Instruction. Always or's in the NO_HEADER option.
std::string Instruction::PrettyPrint(uint32_t options) const;

Then change to the ostream operator to use pretty print as I've set it up (raw ids). This would give people option to change other options.

test/opt/module_test.cpp
```diff
@@ -74,4 +76,34 @@ TEST(ModuleTest, ComputeIdBound) {
                 ->ComputeIdBound());
 }
 
+TEST(ModuleTest, OstreamOperator) {
```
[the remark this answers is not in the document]
alan-baker: Ok

source/disassemble.cpp
```diff
@@ -352,6 +352,56 @@ spv_result_t DisassembleInstruction(
...
+  const size_t word_count_;
+};
+
+spv_result_t DisassembleTargetHeader(void* /*user_data*/,
+                                     spv_endianness_t /*endian*/,
+                                     uint32_t /* magic */, uint32_t /*version*/,
+                                     uint32_t /*generator*/,
+                                     uint32_t /*id_bound*/,
+                                     uint32_t /*schema*/) {
+  return SPV_SUCCESS;
```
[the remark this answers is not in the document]
alan-baker: Ok

source/disassemble.cpp
```diff
@@ -352,6 +352,56 @@ spv_result_t DisassembleInstruction(
...
+  // Check if this is the instruction we want to disassemble.
+  if (!wrapped->stop() &&
+      wrapped->word_count() == parsed_instruction->num_words &&
+      std::equal(wrapped->inst_binary(),
+                 wrapped->inst_binary() + wrapped->word_count(),
+                 parsed_instruction->words)) {
+    // Found the target instruction. Disassemble it and signal that we should
+    // stop searching so we don't output the same instruction again.
+    wrapped->set_stop(true);
+    return wrapped->disassembler()->HandleInstruction(*parsed_instruction);
```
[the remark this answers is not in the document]
alan-baker: Ok

alan-baker:
I changed to check_copyright.py to get it to work with my new header. A full fix for 2018 should be deferred to another PR.

Issue #1154, opened 2018-01-03
MemPass:InsertPhiInstructions does not update instr -> basic block mapping


alan-baker:
InsertPhiInstructions does not update the instr -> block mapping, nor does it mark the analysis as invalid.

I plan to fix this by keeping the analysis up-to-date.

alan-baker:
Rebased, squashed and committed as eb0c73dad6102fc0d4f03c62fe910348bae43a11.

PR #1191, opened 2018-01-11
Maintain instruction to block mapping in phi insertion

alan-baker:
Addresses #1154.

* Changed MemPass::InsertPhiInstructions to set basic blocks for new
phis
* Local SSA elim now maintains instr to block mapping
 * Added a test and confirmed it fails without the updated phis
* IRContext::set_instr_block no longer builds the map if the analysis is
invalid
* Added instruction to block mapping verification to
IRContext::IsConsistent()

### DIFF
--- a/source/opt/ir_context.cpp
+++ b/source/opt/ir_context.cpp
@@ -181,6 +181,19 @@ bool IRContext::IsConsistent() {
       return false;
     }
   }
+  if (AreAnalysesValid(kAnalysisInstrToBlockMapping)) {
+    for (auto& func : *module()) {
+      for (auto& block : func) {
+        bool ok = true;
+        block.ForEachInst([this, &block, &ok](ir::Instruction* inst) {
+          if (get_instr_block(inst) != &block) {
+            ok = false;
+          }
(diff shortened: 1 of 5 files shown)


dneto0:
clang really wants you to remove a std::move

source/opt/ir_context.cpp
```diff
@@ -181,6 +181,19 @@ bool IRContext::IsConsistent() {
       return false;
     }
   }
+  if (AreAnalysesValid(kAnalysisInstrToBlockMapping)) {
+    for (auto& func : *module()) {
+      for (auto& block : func) {
+        bool ok = true;
+        block.ForEachInst([this, &block, &ok](ir::Instruction* inst) {
+          if (get_instr_block(inst) != &block) {
+            ok = false;
```
dnovillo: return false here?  no need to continue checking the rest of the block.

alan-baker:
Rebased, squashed and pushed as eb0c73dad6102fc0d4f03c62fe910348bae43a11.

Closes #1154

Issue #1192, opened 2018-01-12
Add early return versions of ForEach* methods


alan-baker:
There are currently many cases where code could be more efficient if the ForEach* method took a functor with a boolean return value (as opposed to a void return). To avoid confusion, add early return versions as WhileEach* (e.g. ForEachInId -> WhileEachInId). The ForEach* method should be implemented in terms of the WhileEach* method. Example:

void WhileEachInId(std::function<bool(uint32_t id)> f);

void ForEachInId(std::function<void(uint32_t id)> f) {
  WhileEachInId([f](uint32_t id) {
    f(id);
    return true;
  }
}

dneto0:
Fixed by #1196

PR #1196, opened 2018-01-12
Early exit iteration methods

alan-baker:
Addresses #1192 

* Looked through code for instances where code would benefit from early
exit
 * ForEach* reimplemented in terms of WhileEach*
 * Added a corresponding WhileEach* method and updated the code


dneto0:
Squashed and pushed into master as 6587d3f8a31b636

PR #1197, opened 2018-01-12
Remove redundant passes from legalization passes

s-perron:
With work that Alan has done, some passes have become redundant.  ADCE
now removed unused variables.  Dead branch elimination removes
unreachable blocks.  This means we can remove CFG Cleanup and dead
variable elimination.

Contributes to #1118.


dneto0:
Rebased and pushed into master as 8cb0aec7242

PR #1198, opened 2018-01-12
Add MatrixConstant

greg-lunarg:
Shaders with matrix constant are seg faulting on Windows when --ccp enabled.


dneto0:
The formatting is off, but I can take care of that before pushing to master.

dneto0:
Fixed formatting, retested, and pushed into master as  c2aadb02d9

PR #1200, opened 2018-01-13
Skip SpecConstants in CCP.

### LINKED ISSUE #1199: Error in CCP when processing OpSpecConstant* instructions.
At the moment specialization constants look like constants to ccp.  This
causes a problem because they are handled differently by the constant
manager.

I choose to simply skip over them, and not try to add them to the value
table.  We can do specialization before ccp if we want to be able to
propagate these values.

s-perron:
At the moment specialization constants look like constants to ccp.  This
causes a problem because they are handled differently by the constant
manager.

I choose to simply skip over them, and not try to add them to the value
table.  We can do specialization before ccp if we want to be able to
propagate these values.

Fixes #1199.


s-perron:
rebase and merged as 6cc772c3.

PR #1202, opened 2018-01-15
Add generic folding function and use in CCP

s-perron:
The current folding routines have a very cumbersome interface, make them
harder to use, and not a obvious how to extend.

This change is to create a new interface for the folding routines, and
show how it can be used by calling it from CCP.

This does not make a significant change to the behaviour of CCP.  In
general it should produce the same code as before; however it is
possible that an instruction that takes 32-bit integers as inputs and
the result is not a 32-bit integer or bool will not be folded as before.

Contributes to #1164

### DIFF
--- a/source/opt/constants.h
+++ b/source/opt/constants.h
@@ -120,6 +120,31 @@ class IntConstant : public ScalarConstant {
   IntConstant* AsIntConstant() override { return this; }
   const IntConstant* AsIntConstant() const override { return this; }
 
+  int32_t GetS32BitValue() const {
+    // Relies on signed values smaller than 32-bit being sign extended.  See
+    // section 2.2.1 of the SPIR-V spec.
+    assert(words().size() == 1);
+    return words()[0];
+  }
+
+  uint32_t GetU32BitValue() const {
+    // Relies on unsigned values smaller than 32-bit being zero extended.  See
+    // section 2.2.1 of the SPIR-V spec.
+    assert(words().size() == 1);
+    return words()[0];
+  }
+
+  bool IsZero() const {
+    bool is_zero = true;
+    for (uint32_t v : words()) {
+      if (v != 0) {
+        is_zero = false;
+        break;
+      }
+    }
+    return is_zero;
+  }
+
   // Make a copy of this IntConstant instance.
   std::unique_ptr<IntConstant> CopyIntConstant() const {
     return MakeUnique<IntConstant>(type_->AsInteger(), words_);
--- a/source/opt/fold.cpp
+++ b/source/opt/fold.cpp
@@ -13,16 +13,30 @@
 // limitations under the License.
 
 #include "fold.h"
+
 #include "def_use_manager.h"
 #include "ir_context.h"
 
 #include <cassert>
+#include <cstdint>
 #include <vector>
 
 namespace spvtools {
 namespace opt {
 
 namespace {
+
+#ifndef INT32_MIN
+#define INT32_MIN (-2147483648)
+#endif
+
+#ifndef INT32_MAX
+#define INT32_MAX 2147483647
+#endif
+
+#ifndef UINT32_MAX
+#define UINT32_MAX 0xffffffff /* 4294967295U */
+#endif
 
 // Returns the single-word result from performing the given unary operation on
 // the operand value which is passed in as a 32-bit word.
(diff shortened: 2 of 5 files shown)


source/opt/fold.cpp
```diff
@@ -195,6 +195,252 @@ uint32_t FoldScalars(SpvOp opcode,
...
+            return true;
+          }
+        }
+      }
+      break;
+
+      // Comparison
+    case SpvOp::SpvOpULessThan:
+      if (constants[0] != nullptr &&
+          constants[0]->GetU32BitValue() == UINT32_MAX) {
```
dneto0: UINT32_MAX requires cstdint

(Also, style guide requires a reorder of the headers:   Self header first, then system headers, then others)

source/opt/fold.cpp
```diff
@@ -195,6 +195,252 @@ uint32_t FoldScalars(SpvOp opcode,
...
+      // This changes undefined behaviour (ie divide by 0) into a 0.
+      for (uint32_t i = 0; i < 2; i++) {
+        if (constants[i] != nullptr && constants[i]->IsZero()) {
+          *result = 0;
+          return true;
+        }
+      }
+      break;
+
+      // Shifting
```
dneto0: Formatting looks off.  But I'm willing to believe the formatter thinks the comment is part of the previous case.

source/opt/fold.cpp
```diff
@@ -195,6 +195,252 @@ uint32_t FoldScalars(SpvOp opcode,
...
+          return true;
+        }
+      }
+      break;
+
+      // Bitwise operations
+    case SpvOp::SpvOpBitwiseOr:
+      for (uint32_t i = 0; i < 2; i++) {
+        if (constants[i] != nullptr) {
+          uint32_t shift_amount = constants[i]->GetU32BitValue();
```
dneto0: shift_amount is the wrong name.  How about "mask"

source/opt/fold.cpp
```diff
@@ -195,6 +197,252 @@ uint32_t FoldScalars(SpvOp opcode,
...
+          if (shift_amount == 0xFFFFFFFF) {
```
dneto0: I suppose we could generalize slightly for other widths, e.g. 16 bit all-1 mask is 0xFFFF.    But we can leave that for later.  Maybe add a TODO.

source/opt/constants.h
```diff
@@ -120,6 +120,27 @@ class IntConstant : public ScalarConstant {
   IntConstant* AsIntConstant() override { return this; }
   const IntConstant* AsIntConstant() const override { return this; }
 
+  int32_t GetS32BitValue() const {
+    assert(words().size() == 1);
```
dneto0: This relies critically on the fact that SPIR-V requires signed integer literal values be sign-extended.  It's worth saying that. (Section 2.2.1, definition of Literal Number.)

test/opt/fold_test.cpp
```diff
@@ -0,0 +1,515 @@
...
+    const opt::analysis::IntConstant* result =
+        const_mrg->GetConstantFromInst(inst)->AsIntConstant();
+    EXPECT_NE(result, nullptr);
+    if (result != nullptr) {
+      EXPECT_EQ(result->GetU32BitValue(), tc.expected_result);
+    }
+  }
+}
+
+const std::string header = R"(OpCapability Shader
```
dneto0: Avoid objects of class type with static duration.  

Use a function that returns a string instead.

test/opt/fold_test.cpp
```diff
@@ -0,0 +1,515 @@
...
+          "%main_lab = OpLabel\n" +
+          "%n = OpVariable %_ptr_int Function\n" +
+          "%load = OpLoad %int %n\n" +
+          "%2 = OpSGreaterThanEqual %bool %int_max %load\n" +
+          "OpReturn\n" +
+          "OpFunctionEnd",
+      2, true)
+                        ));
+// clang-format on
+
```
dneto0: Please add:

 - non-folding cases for int and bool caused by:
     - 64bit ints aren't folded (yet)
     - non-constant params
 - a couple of folding cases for 16-bit integers

source/opt/constants.h
s-perron: Done.

source/opt/fold.cpp
s-perron: Fixed.  The formatter simply trusts where you place it.

source/opt/fold.cpp
s-perron: Done.

source/opt/fold.cpp
s-perron: done

test/opt/fold_test.cpp
s-perron: done.

test/opt/fold_test.cpp
s-perron: Even though some of the code is anticipating folding values that are less than 32-bits in length (like FoldBinaryIntegerOpToConstant), the public interface still requires a result that is an 32-bit int or a boolean value.  Since the `IsFoldable` is the gate keeper for all of this and it is used in other places, I would prefer not to actually handle 16-bit integers yet.  Just leave it as not too much work.  I'll add some cases where the folder returns `nullptr`.

s-perron:
I realized I missed something.  I forgot the pass the id map through to the new folding routines.  I'll create a follow up work item to get that done.  I'd like to get this much through first.

test/opt/CMakeLists.txt
```diff
@@ -262,6 +262,7 @@ add_spvtools_unittest(TARGET ccp
   LIBS SPIRV-Tools-opt
 )
 
+<<<<<<< HEAD
```
alan-baker: Bad merge

source/opt/constants.h
```diff
@@ -120,6 +120,31 @@ class IntConstant : public ScalarConstant {
   IntConstant* AsIntConstant() override { return this; }
   const IntConstant* AsIntConstant() const override { return this; }
 
+  int32_t GetS32BitValue() const {
+    // Relies on signed values smaller than 32-bit begin sign extended.  See
```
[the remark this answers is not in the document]
s-perron: Done.

source/opt/constants.h
```diff
@@ -120,6 +120,31 @@ class IntConstant : public ScalarConstant {
...
+    // section 2.2.1 of the SPIR-V spec.
+    assert(words().size() == 1);
+    return words()[0];
+  }
+
+  uint32_t GetU32BitValue() const {
+    // Relies on unsigned values smaller than 32-bit begin zero extended.  See
```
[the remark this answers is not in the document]
s-perron: Done.

source/opt/fold.cpp
```diff
@@ -195,6 +197,255 @@ uint32_t FoldScalars(SpvOp opcode,
...
+        }
+      }
+      break;
+
+    // Shifting
+    case SpvOp::SpvOpShiftRightLogical:
+    case SpvOp::SpvOpShiftLeftLogical:
+      if (constants[1] != nullptr) {
+        uint32_t shift_amount = constants[1]->GetU32BitValue();
+        if (shift_amount >= 32) {
```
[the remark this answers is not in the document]
s-perron: Done

s-perron:
@alan-baker Yes there is still lots of work to do.  The next step is what I already mention: pass the id map along.  Then add cases that do not fold to constants, and use that in the redundancy elimination pass.  Then, then we can start adding the folding that we need, 16-bit ints, floating point, etc..

Issue #1203, opened 2018-01-16
Array out of bounds in GLSL caused by incorrect SPIR-V


redorav:
I reported a bug over at the SPIRV-Cross project about an array out of bounds issue where a variable named 645 is being used as a literal to access an array (so the code in glsl ends up as Array[645] instead of Array[_645]. The bug is this https://github.com/KhronosGroup/SPIRV-Cross/issues/397

This seems to be a very recent change as I'm on 5eafc00 commit eb0c73dad6102fc0d4f03c62fe910348bae43a11

which is only 5 days ago. I don't have a simple repro but the SPIRV-Cross bug contains the faulty SPIR-V.

alan-baker:
@redorav can you share the input to spirv-opt?

redorav:
I've had to redo the repro and I lost the original one, but here's another shader that shows the same issue.

[ArrayOutOfBounds_spv.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/1635849/ArrayOutOfBounds_spv.txt)
[ArrayOutOfBounds_optimized_spv.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/1635850/ArrayOutOfBounds_optimized_spv.txt)

The resulting glsl shows the following code:

```glsl
for (int _484 = 0; _484 < 5; _484++)
{
    _486[_484] = clamp(In_uvSet01.xy + _23.g_RingEffectPS.ringOffsets[484].xy, _23.g_RingEffectPS.uvClamp.xy, _23.g_RingEffectPS.uvClamp.zw);
}
```

As you can see, ringOffsets is using 484 directly, instead of _484 as it's meant to be using

redorav:
I'm using a custom recipe, this is the code. I try to base these on the performance passes, but I've had to sometimes disable some that cause bugs which is why I keep a custom one. Perhaps I'm doing something wrong?

```hlsl
optimizer.RegisterPass(spvtools::CreateRemoveDuplicatesPass());
optimizer.RegisterPass(spvtools::CreateMergeReturnPass());
optimizer.RegisterPass(spvtools::CreateInlineExhaustivePass());
optimizer.RegisterPass(spvtools::CreateEliminateDeadFunctionsPass());
optimizer.RegisterPass(spvtools::CreateScalarReplacementPass());
optimizer.RegisterPass(spvtools::CreateLocalAccessChainConvertPass());
[...]

PR #1204, opened 2018-01-16
Fixes missing increment in common uniform elim

alan-baker:
* Addresses #1203
* Increments inIdx in IsConstantIndexAccessChain
 * added test to catch the bug


dneto0:
Windows debug build: built and passed tests but failed to upload artifacts

PR #1206, opened 2018-01-16
Capturing value table by reference in local redundancy elim

alan-baker:
Value table was being copied instead of getting passed by reference. Profiling showed a reduction in runtime from ~90sec to ~23sec.


antiagainst:
Well, one character actually.

dneto0:
Travis might be wedged.   Let's push this into master tomorrow if it hasn't *failed* a Travis build.

dneto0:
Travis Linux bots finished about 3 hrs ago.

dneto0:
Rebased and pushed into master as 3a0eb44da

Issue #1210, opened 2018-01-17
Recent change broke ADCE


alan-baker:
80b743a570a6a616484d9aa9233774fed6cc5463 (#1021) broke ADCE by inserting a potential infinite loop.

dneto0:
Fixed by #1211

PR #1211, opened 2018-01-17
Fixing missing early exit from break identification

alan-baker:
Addresses #1210 

Previous commit 80b743a570a6a616484d9aa9233774fed6cc5463. It removed a check that the user was a branch or conditional branch. Should have been updated to say it was still a "branch" (branch, conditional branch or switch).


dneto0:
In discussions with @alan-baker:
- in trying to produce a test case, we have found several other bugs in ADCE
- by inspection, this PR clearly fixes a clear regression introduced by https://github.com/KhronosGroup/SPIRV-Tools/commit/80b743a570a6a616484d9aa9233774fed6cc5463 

Accepting this PR makes quality incrementally better. @alan-baker will continue to make more test cases and file more issues that are to be addressed by possibly more intrusive PRs.

PR #1215, opened 2018-01-18
CFG: force the creation of a predecessor entry for all basic block.

Naghasan:
This ensure that all basic blocks in a function have a valid entry the CFG object.

The entry block has no predecessors but remains a valid basic block
for which we might want to query the number of predecessors. Because it has no predecessor, no entry were created in the `label2preds_` map.

Some unreachable basic blocks may not have predecessors as well.


dneto0:
Rebased and pushed onto master as 0b1372a8cac

PR #1216, opened 2018-01-18
Adding testcase for #1210

alan-baker:
Minimal testcase that catches the infinite loop from #1210. Confirmed it fails without the fix and passes with it.


dneto0:
Rebased and pushed onto master as 18618061cb8

PR #1217, opened 2018-01-18
Introduce an instruction builder helper class.

Naghasan:
The idea is to have a common place to abstract the instruction building and insertion process and allow automatic update of some analysis. Analysis are updated only if the user request it.

So far, the class can only construct the instructions I needed for the work I'm doing (so very little):
  - Phi instructions;
  - Conditional and unconditional branches;
  - Merge selection instructions.

Only the def-use manager updates are implemented.

### DIFF
--- a/source/opt/ir_builder.h
+++ b/source/opt/ir_builder.h
@@ -22,34 +22,41 @@
 namespace spvtools {
 namespace opt {
 
-// In SPIR-V, ids are encoded as uint16_t, this id is guarantied to be always
+// In SPIR-V, ids are encoded as uint16_t, this id is guaranteed to be always
 // invalid.
-constexpr uint32_t kInvalidId = (uint32_t)-1;
+constexpr uint32_t kInvalidId = std::numeric_limits<uint32_t>::max();
 
 // Helper class to abstract instruction construction and insertion.
 // |AnalysisToPreserve| asks the InstructionBuilder to preserve requested
 // analysis.
 // Supported analysis:
 //   - Def-use analysis
+//   - Instruction to block analysis
 template <ir::IRContext::Analysis AnalysisToPreserve =
               ir::IRContext::kAnalysisNone>
 class InstructionBuilder {
-  static_assert(!(AnalysisToPreserve & ~ir::IRContext::kAnalysisDefUse),
+  static_assert(!(AnalysisToPreserve &
+                  ~(ir::IRContext::kAnalysisDefUse |
+                    ir::IRContext::kAnalysisInstrToBlockMapping)),
                 "Only Def-use analysis update is supported");
 
  public:
   using InsertionPointTy = spvtools::ir::BasicBlock::iterator;
 
-  InstructionBuilder(ir::IRContext* context, InsertionPointTy insert_before)
-      : context_(context), insert_before_(insert_before) {}
+  InstructionBuilder(ir::IRContext* context, ir::BasicBlock* parent,
+                     InsertionPointTy insert_before)
+      : context_(context), parent_(parent), insert_before_(insert_before) {}
 
   // Creates a new selection merge instruction.
   // The id |merge_id| is the merge basic block id.
-  ir::Instruction* AddSelectionMerge(uint32_t merge_id) {
+  ir::Instruction* AddSelectionMerge(
+      uint32_t merge_id,
+      uint32_t selection_control = SpvSelectionControlMaskNone) {
     std::unique_ptr<ir::Instruction> new_branch_merge(new ir::Instruction(
         GetContext(), SpvOpSelectionMerge, 0, 0,
         {{spv_operand_type_t::SPV_OPERAND_TYPE_ID, {merge_id}},
-         {spv_operand_type_t::SPV_OPERAND_TYPE_LITERAL_INTEGER, {0}}}));
+         {spv_operand_type_t::SPV_OPERAND_TYPE_SELECTION_CONTROL,
+          {selection_control}}}));
     return AddInstruction(std::move(new_branch_merge));
   }
 
@@ -65,22 +72,25 @@ class InstructionBuilder {
 
   // Creates a new conditional instruction and the associated selection merge
   // instruction if requested.
-  // The id |cond_id| is the id to the the condition instruction, must be of
+  // The id |cond_id| is the id of the condition instruction, must be of
   // type bool.
-  // The id |true_id| is the basic block id to branch to if the condition is
-  // true.
-  // The id |false_id| is the basic block id to branch to if the condition is
-  // false.
-  // The id |merge_id| is the merge basic block id for the selection merge
-  // instruction. If |merge_id| equals kInvalidId then no selection merge
+  // The id |true_id| is the id of the basic block to branch to if the condition
+  // is true.
+  // The id |false_id| is the id of the basic block to branch to if the
+  // condition is false.
+  // The id |merge_id| is the id of the merge basic block for the selection
+  // merge instruction. If |merge_id| equals kInvalidId then no selection merge
   // instruction will be created.
+  // The value |selection_control| is the selection control flag for the
+  // selection merge instruction.
   // Note that the user must makes sure the final basic block is
   // well formed.
-  ir::Instruction* AddBranchCond(uint32_t cond_id, uint32_t true_id,
-                                 uint32_t false_id,
-                                 uint32_t merge_id = kInvalidId) {
+  ir::Instruction* AddConditionalBranch(
+      uint32_t cond_id, uint32_t true_id, uint32_t false_id,
+      uint32_t merge_id = kInvalidId,
+      uint32_t selection_control = SpvSelectionControlMaskNone) {
     if (merge_id != kInvalidId) {
-      AddSelectionMerge(merge_id);
+      AddSelectionMerge(merge_id, selection_control);
     }
     std::unique_ptr<ir::Instruction> new_branch(new ir::Instruction(
         GetContext(), SpvOpBranchConditional, 0, 0,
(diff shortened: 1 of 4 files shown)


source/opt/ir_builder.h
```diff
@@ -0,0 +1,144 @@
...
+  InstructionBuilder(ir::IRContext* context, InsertionPointTy insert_before)
+      : context_(context), insert_before_(insert_before) {}
+
+  // Creates a new selection merge instruction.
+  // The id |merge_id| is the merge basic block id.
+  ir::Instruction* AddSelectionMerge(uint32_t merge_id) {
+    std::unique_ptr<ir::Instruction> new_branch_merge(new ir::Instruction(
+        GetContext(), SpvOpSelectionMerge, 0, 0,
+        {{spv_operand_type_t::SPV_OPERAND_TYPE_ID, {merge_id}},
+         {spv_operand_type_t::SPV_OPERAND_TYPE_LITERAL_INTEGER, {0}}}));
```
alan-baker: This should be:
{SPV_OPERAND_TYPE_SELECTION_CONTROL, {SpvSelectionControlMaskNone}}

I would also suggest changing the control to an argument and defaulting it to none.

source/opt/ir_builder.h
```diff
@@ -0,0 +1,144 @@
...
+namespace opt {
+
+// In SPIR-V, ids are encoded as uint16_t, this id is guarantied to be always
+// invalid.
+constexpr uint32_t kInvalidId = (uint32_t)-1;
+
+// Helper class to abstract instruction construction and insertion.
+// |AnalysisToPreserve| asks the InstructionBuilder to preserve requested
+// analysis.
+// Supported analysis:
```
alan-baker: At least you should also support updating the instruction to block mapping.

source/opt/ir_builder.h
```diff
@@ -0,0 +1,144 @@
...
+#include "opt/basic_block.h"
+#include "opt/instruction.h"
+#include "opt/ir_context.h"
+
+namespace spvtools {
+namespace opt {
+
+// In SPIR-V, ids are encoded as uint16_t, this id is guarantied to be always
+// invalid.
+constexpr uint32_t kInvalidId = (uint32_t)-1;
```
alan-baker: std::numeric_limits<uint32_t>::max()

source/opt/ir_builder.h
```diff
@@ -0,0 +1,144 @@
...
+
+  // Creates a phi instruction.
+  // The id |type| must be the id of the phi instruction's type.
+  // The vector |incomings| must be a sequence of pairs of <definition id,
+  // parent id>.
+  ir::Instruction* AddPhi(uint32_t type,
+                          const std::vector<uint32_t>& incomings) {
+    assert(incomings.size() % 2 == 0 && "A sequence of pairs is expected");
+    std::vector<ir::Operand> phi_ops;
+    for (size_t i = 0; i < incomings.size(); i += 2) {
```
alan-baker: I get your thinking here, but I feel like it would be better to go with the simpler code of adding each input operand separately (e.g. increment by 1 and add 1 element in the body).

source/opt/ir_builder.h
```diff
@@ -0,0 +1,144 @@
...
+  ir::Instruction* AddBranch(uint32_t label_id) {
+    std::unique_ptr<ir::Instruction> new_branch(new ir::Instruction(
+        GetContext(), SpvOpBranch, 0, 0,
+        {{spv_operand_type_t::SPV_OPERAND_TYPE_ID, {label_id}}}));
+    return AddInstruction(std::move(new_branch));
+  }
+
+  // Creates a new conditional instruction and the associated selection merge
+  // instruction if requested.
+  // The id |cond_id| is the id to the the condition instruction, must be of
```
alan-baker: is the id of the condition instruction.

source/opt/ir_builder.h
```diff
@@ -0,0 +1,144 @@
...
+  // type bool.
+  // The id |true_id| is the basic block id to branch to if the condition is
```
alan-baker: similar here

source/opt/ir_builder.h
```diff
@@ -0,0 +1,144 @@
...
+  // true.
+  // The id |false_id| is the basic block id to branch to if the condition is
```
alan-baker: similar here

source/opt/ir_builder.h
```diff
@@ -0,0 +1,144 @@
...
+#define LIBSPIRV_OPT_IR_BUILDER_H_
+
+#include "opt/basic_block.h"
+#include "opt/instruction.h"
+#include "opt/ir_context.h"
+
+namespace spvtools {
+namespace opt {
+
+// In SPIR-V, ids are encoded as uint16_t, this id is guarantied to be always
```
alan-baker: guarantied -> guaranteed

source/opt/ir_builder.h
```diff
@@ -0,0 +1,144 @@
...
+  // The id |true_id| is the basic block id to branch to if the condition is
+  // true.
+  // The id |false_id| is the basic block id to branch to if the condition is
+  // false.
+  // The id |merge_id| is the merge basic block id for the selection merge
+  // instruction. If |merge_id| equals kInvalidId then no selection merge
+  // instruction will be created.
+  // Note that the user must makes sure the final basic block is
+  // well formed.
+  ir::Instruction* AddBranchCond(uint32_t cond_id, uint32_t true_id,
```
alan-baker: I'd rather not save the characters and just name this AddBranchConditional (or AddConditionalBranch).

source/opt/ir_builder.h
Naghasan: I agree, this should be an argument

source/opt/ir_builder.h
Naghasan: I understand, on the other hand, I always found disturbing when you get an extra increment inside the body of a for loop.

In the code base, phi node handling are commonly done using this idiom. Is that legacy you are trying to get rid of ?

alan-baker: Not per se, but in this case I think it is unnecessary. You are simple iterating through each input. While they are paired logically, I think the code benefits only using the idiom when necessary.

Naghasan:
@alan-baker I addressed your comment. Note that I made operator overload for | and << constexpr (so they can be used in the static_assert)

source/opt/ir_builder.h
```diff
@@ -0,0 +1,169 @@
...
+  // The id |type| must be the id of the phi instruction's type.
+  // The vector |incomings| must be a sequence of pairs of <definition id,
+  // parent id>.
+  ir::Instruction* AddPhi(uint32_t type,
+                          const std::vector<uint32_t>& incomings) {
+    assert(incomings.size() % 2 == 0 && "A sequence of pairs is expected");
+    std::vector<ir::Operand> phi_ops;
+    for (size_t i = 0; i < incomings.size(); i++) {
+      phi_ops.push_back({SPV_OPERAND_TYPE_ID, {incomings[i]}});
+      i++;
```
alan-baker: This isn't what I meant. Sorry if I was unclear. I wanted:

for (size_t i = 0; i != incomings.size(); ++i) {
  phi_ops.push_back({SPV_OPERAND_TYPE_ID, {incomings[i]}});
}

source/opt/ir_builder.h
```diff
@@ -0,0 +1,169 @@
...
+    return insn_ptr;
+  }
+
+  // Returns the insertion point iterator.
+  InsertionPointTy GetInsertPoint() { return insert_before_; }
+
+  // Returns the context which instructions are constructed for.
+  ir::IRContext* GetContext() const { return context_; }
+
+  // Returns the set of preserved analysis.
```
alan-baker: Should make this Analyses now

source/opt/ir_builder.h
```diff
@@ -0,0 +1,169 @@
...
+namespace opt {
+
+// In SPIR-V, ids are encoded as uint16_t, this id is guaranteed to be always
+// invalid.
+constexpr uint32_t kInvalidId = std::numeric_limits<uint32_t>::max();
+
+// Helper class to abstract instruction construction and insertion.
+// |AnalysisToPreserve| asks the InstructionBuilder to preserve requested
+// analysis.
+// Supported analysis:
```
alan-baker: anlaysis -> analyses

source/opt/ir_builder.h
```diff
@@ -0,0 +1,169 @@
...
+//   - Def-use analysis
+//   - Instruction to block analysis
+template <ir::IRContext::Analysis AnalysisToPreserve =
```
alan-baker: analysis -> analyses

source/opt/ir_builder.h
```diff
@@ -0,0 +1,169 @@
...
+class InstructionBuilder {
+  static_assert(!(AnalysisToPreserve &
+                  ~(ir::IRContext::kAnalysisDefUse |
+                    ir::IRContext::kAnalysisInstrToBlockMapping)),
+                "Only Def-use analysis update is supported");
+
+ public:
+  using InsertionPointTy = spvtools::ir::BasicBlock::iterator;
+
+  InstructionBuilder(ir::IRContext* context, ir::BasicBlock* parent,
```
s-perron: We should be able to get parent block from the parent block of the insertion point.  Should only get it if the instToBlockMapping is already built, and needs to be preserved.  I don't want to force user to look up the basic block when it is not readily available.

source/opt/ir_builder.h
```diff
@@ -0,0 +1,169 @@
...
+                     InsertionPointTy insert_before)
```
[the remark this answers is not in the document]
Naghasan: I used the iterator to allow insertion at the end of a basic block. An `Instruction*` can be turned into an InsertionPointTy, so we can still have both.

We need a Ctor that takes an `Instruction*` (insertion before this instruction) and a `BasicBlock*` (insertion at the end). That should cover the use cases and the remove the need for that extra parameter.

I will make the current Ctor internal, it can be moved as public if there is a need for this overload.

source/opt/ir_builder.h
Naghasan: I think you were clear, I had my mind traped at "phi operands are pairs of" ... Sorry about that.

s-perron:
I'll merge this right now.

s-perron:
Rebased, squashed, and merged as cf3b2a58.

PR #1219, opened 2018-01-18
DeadInsertElim: Detect and DCE dead Inserts

greg-lunarg:
This adds Dead Insert Elimination to the end of the --eliminate-insert-extract pass. See the new tests for examples of code that will benefit.

Essentially, this removes OpCompositeInsert instructions which are not used, either because there is no instruction which uses the value at the index it is inserted, or because a subsequent insert intercepts any such use.

This code has been seen to remove significant amounts of dead code from real-life HLSL shaders being ported to Vulkan. In fact, it is needed to remove dead texture samples which cause Vulkan validation layer errors (unbound textures and samplers) if not removed . Such DCE is thus required for fxc equivalence and legalization.

This analysis operates across "chains" of Inserts which can also contain Phi instructions.

### DIFF
--- a/source/opt/insert_extract_elim.cpp
+++ b/source/opt/insert_extract_elim.cpp
@@ -30,6 +30,10 @@ namespace {
 const uint32_t kExtractCompositeIdInIdx = 0;
 const uint32_t kInsertObjectIdInIdx = 0;
 const uint32_t kInsertCompositeIdInIdx = 1;
+const uint32_t kTypeVectorCountInIdx = 1;
+const uint32_t kTypeMatrixCountInIdx = 1;
+const uint32_t kTypeArrayLengthIdInIdx = 1;
+const uint32_t kTypeIntWidthInIdx = 0;
 const uint32_t kConstantValueInIdx = 0;
 const uint32_t kVectorShuffleVec1IdInIdx = 0;
 const uint32_t kVectorShuffleVec2IdInIdx = 1;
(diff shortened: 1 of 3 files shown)


s-perron:
Does this extra code do anything that cannot be accomplished by adding a pass of ADCE at the end of list of optimizations?  We wanted to amalgamate all of the dead code removal stuff, so I'm a little tentative about adding more.  Lots of transformations create dead code, and we probably don't want some thing like this in everyone of them.  

Also, I see the potential for creep.  Currently it handles a subset of instructions, but there may come a case where we decide it needs to handle a slightly larger class.  That could keep happening.

Can we have further justification beyond the existence of dead code that needs to be removed?

greg-lunarg:
@s-perron Good questions.

Yes, I believe that this optimization cannot be accomplished with the current passes rearranged in some way. ADCE essentially removes instructions whose ultimate result_id is not used, and STOREs whose variable is not loaded. The Inserts we are removing do not fall into these categories.

The problem of removing these dead inserts is inherent with vector code. Like any optimization, there is probably more than one way to achieve removal of this code. But after quite a bit of thought, it seemed to me that this was the most cost effective in terms of lines of code. For example, we could also remove this dead code by performing scalar replacement on all vector code. But how many lines of code would it take to put that scalar code back into vector code? I am guessing more than I am introducing here.

Ultimately it might make the most sense as a standalone pass called near the end of optimization. For the moment I put it at the end of EliminateInsertExtract for several reasons. One is that it shares several utilities with EliminateInsertExtract. Another is that it eliminates dead inserts created by EliminateInsertExtract. Another is that when opportunities are created for EliminateInsertExtract they can also be created for DeadInsertElim as well, so they would likely be naturally paired anyway. Another is that EliminateInsertExtract usually reduces the work that DeadInsertElim does, so it is natural for them to execute in this order. I didn't want to go through the extra work of creating a new pass until the final disposition of this code was determined. But I am not insistent in the way this is packaged or called. If someone has a better idea, I would be open to it.

As you observed, there is room to improve this code by adding support/analysis for additional instructions that can use Insert instructions (eg VectorShuffle). I only added the support that was needed at the moment by a Vulkan development team I have been working with. If our goal is fxc equivalence and comprehensive legalization capability, then eventually I think we may need to add these improvements.

The bottom line is that there is a team of Vulkan developers that needs to have their dead code removed today. If someone can figure out a way to remove the code with fewer lines of code than I have written, that would be great. But until then I propose we make this capability available to the community.

s-perron:
Thanks.  For the explanation.  I'll take a look, and see how this works.  The first things that jumps out at me is that there is not unit test for an OpPhi in a loop header.  Also, I think this will have to be improved at some point, so we can think about how to do this more completely.  For example, the test "ConflictingInsertPreventsOptimization2", still has dead code because the `%23 = OpLoad %S_t %s0` is not needed.  I have a feeling you will end up rewriting this code to get cases like that at some point.

When I think of what you are trying to do in the Insert-extract-elimination as a whole, I think instruction simplification/fold/InstCombine.  It can get a lot of the things you are doing in this pass.  However, the difficulty with the folding method is that it usually looks at a window of instructions, but you might need to look at arbitrary long chains.  This make it a little more difficult.  Your method might be worth it.  Unfortunately, I am only working a half day today, and will be traveling next week.  I'll think about this more, and look at the code more carefully.  I'll try to get back to you early next week.

greg-lunarg:
Thanks. Not extremely urgent, but I did want everyone to understand that this is addressing real problems for real Vulkan developers today.

In the "after" code for ConflictingInsertPreventsOptimization2, %23 is not completely "dead". It is the input value, albeit unused, of the Insert (one of the "fun" complications of Insert) so we can't eliminate it completely. It will eventually become an Undef value (after SSA rewrite) so ultimately it works out. So I am not seeing a need for DeadInsertElim to do more here.

I will think on a test with a loop in it.

source/opt/insert_extract_elim.cpp
```diff
@@ -75,6 +79,207 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
...
+      return typeInst->NumInOperands();
+    } break;
+    default: { return 0; } break;
+  }
+}
+
+void InsertExtractElimPass::MarkInsertChain(ir::Instruction* insert,
+                                            std::vector<uint32_t>* pExtIndices,
+                                            uint32_t extOffset) {
+  // If extract indices are empty, mark all subcomponents if type
```
s-perron: It is a little confusing because I cannot tell if `insert` is suppose to be an OpCompositeInsert instruction or not.  Looking at the code the naming seems to suggest it is an insert instruction, but it will not always be.  The code works, but could lead to errors when other people try to add code.  I suggest having an early exit at the start.

source/opt/insert_extract_elim.cpp
```diff
@@ -75,6 +79,207 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
...
+  // is constant length.
+  if (pExtIndices == nullptr) {
+    uint32_t cnum = ComponentNum(insert->type_id());
+    if (cnum > 0) {
+      std::vector<uint32_t> extIndices;
+      for (uint32_t i = 0; i < cnum; i++) {
+        extIndices.clear();
+        extIndices.push_back(i);
+        MarkInsertChain(insert, &extIndices, 0);
```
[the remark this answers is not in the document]
s-perron: Do you want this called for no inset instructions?  It is not going to do anything.

source/opt/insert_extract_elim.cpp
```diff
@@ -223,6 +428,7 @@ bool InsertExtractElimPass::EliminateInsertExtract(ir::Function* func) {
       }
     }
   }
+  modified |= EliminateDeadInserts(func);
```
s-perron: Since this is just being tacked on to the end, it would be better as it own transformation.  This will create more flexibility, and make the files and classes smaller and easier to read.

source/opt/insert_extract_elim.cpp
```diff
@@ -75,6 +79,207 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
...
+  liveInserts_.erase(insInst->result_id());
+}
+
+bool InsertExtractElimPass::EliminateDeadInserts(ir::Function* func) {
+  bool modified = false;
+  bool lastmodified = true;
+  // Each pass can delete dead instructions, thus potentially revealing
+  // new dead insertions ie insertions with no uses.
+  while (lastmodified) {
+    lastmodified = EliminateDeadInsertsOnePass(func);
```
s-perron: Could you add a test case that relies on this loop being run twice?

source/opt/insert_extract_elim.cpp
```diff
@@ -75,6 +79,207 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
...
+    // Get next insert in chain
+    const uint32_t compId =
+        insInst->GetSingleWordInOperand(kInsertCompositeIdInIdx);
+    insInst = get_def_use_mgr()->GetDef(compId);
+  }
+  // If insert chain ended with phi, do recursive call on each operand
+  if (insInst->opcode() != SpvOpPhi) return;
+  // If phi is already live, we have processed already. Return to
+  // avoid infinite loop
+  if (liveInserts_.find(insInst->result_id()) != liveInserts_.end()) return;
```
alan-baker: If (!liveInserts_.insert(instInst->result_id()).second) return;

Then you don't need the subsequent insert.

greg-lunarg:
@s-perron 
> Actually, I think simplifying that test is easier than that.

This is an interesting optimization and could very well be worthwhile in real life. One thought...

"Folding" could get this simple test, but remember that insert-extract-elimination may need to traverse a fairly lengthy insert chain to get from the extract to its corresponding insert. So the place for this optimization may be in insert-extract-elimination.

source/opt/insert_extract_elim.cpp
```diff
@@ -75,6 +79,207 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
...
+    case SpvOpTypeMatrix:
+    case SpvOpTypeArray:
+    case SpvOpTypeStruct:
+      return true;
+    default:
+      return false;
+  }
+}
+
+uint32_t InsertExtractElimPass::ComponentNum(uint32_t typeId) {
```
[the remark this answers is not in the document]
greg-lunarg: Given prior art (eg NumOperands, NumInOperands, etc) I went with NumComponents.

source/opt/insert_extract_elim.cpp
greg-lunarg: Changed insert to insertChain and added early exit (with comment).

source/opt/insert_extract_elim.cpp
```diff
@@ -75,6 +79,207 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
...
+  uint32_t icnt = 0;
+  insInst->ForEachInId([&icnt, &pExtIndices, &extOffset, this](uint32_t* idp) {
+    if (icnt % 2 == 0) {
+      ir::Instruction* pi = get_def_use_mgr()->GetDef(*idp);
+      MarkInsertChain(pi, pExtIndices, extOffset);
+    }
+    ++icnt;
+  });
+  // Remove phi from live set when finished
+  liveInserts_.erase(insInst->result_id());
```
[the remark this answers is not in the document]
greg-lunarg: Phis don't really need to be marked live. This optimization does not delete Phis. I was just using the liveInserts_ to mark visited Phis to avoid infinite loops in the insert chain traversal. When we are done visiting the Phi, they can be unmarked. I have added a new set called visitedPhis_ for this purpose so this isn't so confusing.

source/opt/insert_extract_elim.cpp
greg-lunarg: Sure. Can we do this in a subsequent commit?

source/opt/insert_extract_elim.cpp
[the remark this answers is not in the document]
greg-lunarg: What is the largest array you would consider reasonable for this algorithm?

source/opt/insert_extract_elim.cpp
```diff
@@ -75,6 +79,214 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
   return typeInst->opcode() == SpvOpTypeVector;
 }
 
+bool InsertExtractElimPass::IsComposite(uint32_t typeId) {
+  ir::Instruction* typeInst = get_def_use_mgr()->GetDef(typeId);
+  switch (typeInst->opcode()) {
```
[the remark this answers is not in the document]
greg-lunarg: Done.

source/opt/insert_extract_elim.cpp
greg-lunarg: Good trick. Thanks!

source/opt/insert_extract_elim.cpp
greg-lunarg: Done.

source/opt/insert_extract_elim.cpp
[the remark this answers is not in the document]
greg-lunarg: Since my customer did not need array inserts optimized and the value seems questionable, I am disabling for arrays for now.

greg-lunarg:
I think I have responded to all comments. Let me know if there are other concerns.

source/opt/insert_extract_elim.cpp
s-perron: Sure it can be in another commit.

source/opt/insert_extract_elim.cpp
```diff
@@ -75,6 +79,216 @@ bool InsertExtractElimPass::IsVectorType(uint32_t typeId) {
...
+  // Not currently optimizing array inserts.
+  ir::Instruction* typeInst = get_def_use_mgr()->GetDef(insertChain->type_id());
+  if (typeInst->opcode() == SpvOpTypeArray) return;
+  // Insert chains are only composed of inserts and phis
+  if (insertChain->opcode() != SpvOpCompositeInsert &&
+      insertChain->opcode() != SpvOpPhi)
+    return;
+  // If extract indices are empty, mark all subcomponents if type
+  // is constant length.
+  if (pExtIndices == nullptr) {
```
s-perron: We can leave this for now, but I'm think there must be a better way of doing this.  Inside this if statement you are probably looking for an insert that covers another insert.  We could look for that more directly.  We can leave that for a follow up issue.

s-perron:
Rebased, squashed, and merged as 684997eb.

source/opt/insert_extract_elim.cpp
greg-lunarg: The code in this "if" is meant to do more precise marking when all components have been requested to be marked (pExtIndices == nullptr). By making multiple single component mark calls we are indeed able to do "Insert After Insert" (and other) eliminations from this situation.

Issue #1212, opened 2018-01-17
ADCE produces invalid code: two branches in a row, missing block


dneto0:
@alan-baker and I found this while trying to make a test case for #1211 

The issue was introduced in  22faa2b083aa738ef3 when general block cleanup was added to ADCE.

Two test slightly different test cases:

a.spvasm:
```
                  OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main" %out_var
       %void = OpTypeVoid
     %voidfn = OpTypeFunction %void
      %float = OpTypeFloat 32
      %bool = OpTypeBool
      %true = OpConstantTrue %bool

%ptr_out_float = OpTypePointer Output %float
%ptr_in_float = OpTypePointer Input %float
    %out_var = OpVariable %ptr_out_float Output
     %float0 = OpConstant %float 0
     %float1 = OpConstant %float 1
     %float3 = OpConstant %float 3

    %main = OpFunction %void None %voidfn
          %entry = OpLabel
           OpBranch %outer_head


          %outer_head = OpLabel
          OpLoopMerge %outer_merge %outer_cont None
          OpBranchConditional %true %inner_head %outer_merge


          %inner_head = OpLabel
          OpStore %out_var %float3
          OpLoopMerge %inner_merge %inner_head None
          OpBranchConditional %true %inner_merge %inner_head

          %inner_merge = OpLabel
           OpBranch %outer_merge

          %outer_cont = OpLabel
           OpBranch %outer_head

          %outer_merge = OpLabel
           %out_phi = OpPhi %float %float0 %outer_head %float1 %inner_merge
           OpBranch %exit

       %exit = OpLabel
               OpReturn
               OpFunctionEnd

```
Running  --eliminate-dead-code-aggressive on this, I get:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %1 "main" %2
       %void = OpTypeVoid
          %4 = OpTypeFunction %void
      %float = OpTypeFloat 32
       %bool = OpTypeBool
       %true = OpConstantTrue %bool
%_ptr_Output_float = OpTypePointer Output %float
          %2 = OpVariable %_ptr_Output_float Output
    %float_3 = OpConstant %float 3
          %1 = OpFunction %void None %4
         %13 = OpLabel
               OpBranch %14
         %14 = OpLabel
               OpBranchConditional %true %17 %15   ;;;; There is no block 17
[...]

PR #1220, opened 2018-01-18
Simplifying code for adding instructions to worklist

alan-baker:
Addresses #1212

* AddToWorklist can now be called unconditionally
 * It will only add instructions that have not already been marked as
 live
 * Fixes a case where a merge was not added to the worklist because the
 branch was already marked as live
* Added two similar tests that fail without the fix


s-perron:
Rebased and merged as 73940aba.

Issue #670, opened 2017-06-02
spirv-val has inconsistent location reporting for errors


baldurk:
Running spirv-val on a particular module I was looking at I got this:

```
info: 507: Duplicate non-aggregate type declarations are not allowed. Opcode: 32
error: 3095: The result pointer storage class and base pointer storage class in OpAccessChain do not match.
```

507 being the opcode index, and 3095 being the word count offset of the parameter in the OpAccessChain.

The latter I don't think is useful directly, but being able to identify the particular word is nice. I'd suggest that instead the opcode index should always be used since it's easy to look up in the spirv-dis output (apart from the comments at the start with the header, it's just the line number), and instead the error should be displayed something like `error: filename.spv:574:1: ...` similar to line:column numbers in compiler outputs.

As an aside, I didn't realise it was the opcode index until I looked at the source, although that's partly because one of them wasn't so I dismissed that possibility. It might be a good idea to explicitly call that out, and add an option to spirv-dis to display the opcode indices on the left-hand side similar to how it can display the word offset in hex as a comment on the right.

----

Extra bonus points if you used the disassembly to display context on what exactly is wrong, like this mockup:

```                                                                                 
output_dump.spv:507: info: Duplicate non-aggregate type declarations are not allowed.
%_ptr_Uniform_float_0 = OpTypePointer Uniform %float
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

output_dump.spv:431: note: First type declaration here
%_ptr_Uniform_float = OpTypePointer Uniform %float
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

output_dump.spv:574:1: error: The result pointer storage class and base pointer storage class in OpAccessChain do not match.
      %25242 = OpAccessChain %_ptr_Uniform_v4float %_entryPointOutput %uint_1
                             ~~~~~~~~~~~~~~~~~~~~~ ^^^^^^^^^^^^^^^^^^^^^^^^^^

output_dump.spv:422: note: %_ptr_Uniform_v4float declared here
%_ptr_Uniform_v4float = OpTypePointer Uniform %v4float
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

output_dump.spv:422: note: %_entryPointOutput declared here
%_entryPointOutput = OpVariable %_ptr_Output_VS_OUTPUT_0 Output
[...]

PR #1221, opened 2018-01-18
WIP:  Validator disassembles offending instruction during "inline" phase of validation

dneto0:
This has a few moving parts, separated into different commits in this PR.  In reverse order:

   Disassemble the bad instruction during "inline" validation phase

   Add ValidationState_t method to disassemble an instruction
    
      Fixes byte offset accounting when disassembling just a single
      instruction.

    Validator state object holds a reference to the SPIR-V binary

    Validator uses SuffixableDiagnosticGenerator for diagnostics
    
       Currently no suffix is ever offered.

    validator: Register instructions early in the flow
    
       This is a step toward disassembling the at-fault instruction
       for any validation error.

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -16,6 +16,7 @@
 
 #include <cassert>
 
+#include "disassemble.h"
 #include "opcode.h"
 #include "val/basic_block.h"
 #include "val/construct.h"
@@ -774,6 +775,17 @@ bool ValidationState_t::GetConstantValUint64(uint32_t id, uint64_t* val) const {
     *val |= uint64_t(inst->word(4)) << 32;
   }
   return true;
+}
+
+std::string ValidationState_t::Disassemble(const Instruction& inst) {
+  const spv_parsed_instruction_t& c_inst(inst.c_inst());
+  uint32_t disassembly_options = SPV_BINARY_TO_TEXT_OPTION_NO_HEADER |
+                                 SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES |
+                                 SPV_BINARY_TO_TEXT_OPTION_SHOW_BYTE_OFFSET;
+
+  return spvtools::spvInstructionBinaryToText(
+      context()->target_env, c_inst.words, c_inst.num_words, words_, num_words_,
+      disassembly_options);
 }
 
 }  // namespace libspirv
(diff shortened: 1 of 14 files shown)


dneto0:
This is partial progress toward #670

source/val/validation_state.h
```diff
@@ -421,9 +463,15 @@ class ValidationState_t {
   /// Stores the Validator command line options. Must be a valid options object.
   const spv_const_validator_options options_;
 
+  /// The SPIR-V binary module we're validating.
```
atgoo: (not owned)

source/val/validation_state.cpp
```diff
@@ -418,6 +423,12 @@ void ValidationState_t::RegisterInstruction(
       }
     }
   }
+
+  return &(ordered_instructions_.back());
+}
+
+void ValidationState_t::RegisterDefinition(uint32_t id, Instruction* inst) {
+  all_definitions_.insert(make_pair(id, inst));
```
atgoo: all_definitions_.emplace(id, inst);

source/validate_id.cpp
```diff
@@ -2678,9 +2677,11 @@ spv_result_t IdPass(ValidationState_t& _,
     }
   }
   if (result_id) {
```
atgoo: I don't think there is a need for this block to be here. Consider moving it inside RegisterInstruction(). Then there is also no need to pipeline internal_instruction.

dneto0:
Clearly I forgot to consider the MarkvCodec's use of the ValidationState_t.  The encoder is probably not a problem, but the decoder's use might be a problem because there *is* no fully elaborated binary at the start.

Setting this to WIP, but I'll take feedback in the meantime.

atgoo:
MarkvCodec doesn't need to use ValidationState to encode/decode. It optionally uses it to validate.

The problem can be solved by giving validation state a callback which would return the current binary.

atgoo:
Does spvtools::spvInstructionBinaryToText work for an incomplete module?

dneto0:
> The problem can be solved by giving validation state a callback which would return the current binary.

That's a good idea.  I can make that work.  I had prototyped a solution where the Markv endcoder would not get the benefit of the instruction printing.  But your idea will make it better.

> Does spvtools::spvInstructionBinaryToText work for an incomplete module?

I believe so. 
The interesting bit is whether the "friendly name mapper" works on a partial module.  I believe it will behave sensibly, but I should update its comments to say so, and add a test. It consciously ignores the status result from spvBinaryParse.

dneto0:
Superceded by #1600, which has gone into master.
Yay

Issue #1214, opened 2018-01-17
ADCE Infinite loop


alan-baker:
Separate from #1210, the following SPIR-V produces an infinite loop in ADCE.

```
OpCapability Shader
OpMemoryModel Logical GLSL450
OpEntryPoint Fragment %func "func" %var
OpExecutionMode %func OriginUpperLeft
%void = OpTypeVoid
%bool = OpTypeBool
%true = OpConstantTrue %bool
%int = OpTypeInt 32 0
%int_ptr_Output = OpTypePointer Output %int
%var = OpVariable %int_ptr_Output Output
%int0 = OpConstant %int 0
%functy = OpTypeFunction %void
%func = OpFunction %void None %functy
%entry = OpLabel
OpBranch %outer_header
%outer_header = OpLabel
OpLoopMerge %outer_merge %outer_continue None
OpBranchConditional %true %inner_header %outer_continue
%inner_header = OpLabel
OpStore %var %int0
OpLoopMerge %inner_merge %inner_continue None
OpBranchConditional %true %inner_merge %inner_continue
%inner_continue = OpLabel
OpBranch %inner_header
%inner_merge = OpLabel
OpBranchConditional %true %outer_merge %outer_continue
%outer_continue = OpLabel
OpBranch %outer_header
%outer_merge = OpLabel
%phi = OpPhi %int %int0 %outer_header %int0 %inner_merge %int0 %continue
OpStore %var %phi
OpLoopMerge %merge %continue None
OpBranchConditional %true %merge %continue
%continue = OpLabel
OpBranch %outer_merge
%merge = OpLabel
OpReturn
OpFunctionEnd
```

![cfg](https://user-images.githubusercontent.com/33432579/35067761-004bafe0-fba3-11e7-8e46-bd44bc0a3c75.png)

When attempting to identify breaks from the outer loop (11, %outer_header), adce considers the users of the merge id. One of those is the branch in the continue for the following loop (18, %continue). That gets the traversal back to the header of the following loop (12, %outer_merge) and it can't exit, because it is a header, but not the expected one.

PR #1222, opened 2018-01-18
Fixes infinite loop in ADCE

alan-baker:
Addresses #1214 

* Addresses how breaks are indentified to prevent infinite loops when
back to back loop share a merge and header
* Added test to catch the bug

~~Note this will conflict with #1220. I'll rebase when #1220 is merged.~~

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -190,10 +190,14 @@ void AggressiveDCEPass::ComputeBlock2HeaderMaps(
     std::list<ir::BasicBlock*>& structuredOrder) {
   block2headerBranch_.clear();
   branch2merge_.clear();
+  structured_order_index_.clear();
   std::stack<ir::Instruction*> currentHeaderBranch;
   currentHeaderBranch.push(nullptr);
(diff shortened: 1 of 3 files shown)


source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -235,27 +239,26 @@ void AggressiveDCEPass::AddBranch(uint32_t labelId, ir::BasicBlock* bp) {
...
-          branchInst = hdrBranch;
-        }
-        if (!IsLive(user)) {
-          AddToWorklist(user);
-          // Add branch's merge if there is one
-          ir::Instruction* userMerge = branch2merge_[user];
-          if (userMerge != nullptr) AddToWorklist(userMerge);
+        ir::BasicBlock* block = context()->get_instr_block(user);
+        uint32_t index = structured_order_index_[block];
+        if (headerIndex < index && index < mergeIndex) {
```
dneto0: This is mercifully more readable.
Thanks!

alan-baker:
Rebase and committed as 3b780db7f8fe81fc0f7446f32f4f604f3cf2485d

Fixes #1214

Issue #711, opened 2017-07-14
Convert if-then-else to OpSelect to collapse control flow.


dneto0:
E.g. example reduced from IfThenElse test in local_ssa_elim_test

if (c) { v = x; } else { v = y; } z = v;
Collapse to    z = OpSelect c x y

alan-baker:
Found some examples in production shaders that would benefit from this opt and have started some initial work on it.

alan-baker:
Committed in 2e93e806e454e2cfefabf217320038188c6c1fb6

PR #1226, opened 2018-01-22
Initial implementation of if conversion

alan-baker:
Addresses #711 

Where possible converts phis into OpSelects. This can enable other optimizations like adce and block merging to simplify the CFG. I have instances of industrial shaders that require this optimization.

This initial implementation is limited in scope. It does not handle phis with more than two predecessors, nor does it perform any code motion. Both of these are potentially useful additions to the pass, but I want a base to work from before extending the pass.


source/opt/if_conversion.cpp
```diff
@@ -0,0 +1,209 @@
...
+        ++iter;
+      }
+
+      ir::BasicBlock* common = nullptr;
+      block.WhileEachPhiInst([this, &modified, &common, &to_kill, dominators,
+                              &block, &iter](ir::Instruction* phi) {
+        // A false return from this function stops iteration through subsequent
+        // phis. If any of the cfg aspects of the phi are incompatible (e.g.
+        // bad domination of incoming block), there is no point checking other
+        // phi nodes. If the function returns true, then this particular phi is
```
s-perron: It is also possible that the current phi is compatible.  What is the return value in that case?

source/opt/if_conversion.cpp
```diff
@@ -0,0 +1,209 @@
...
+        // incompatible, but subsequent phis may be compatible.
+
+        // TODO(alan-baker): Extend to more than two predecessors
+        if (phi->NumInOperands() != 4u) return false;
+
+        // This phi is not compatible, but subsequent phis might be.
+        if (!CheckType(phi->type_id())) return true;
+
+        ir::BasicBlock* inc0 = GetIncomingBlock(phi, 0u);
```
s-perron: Can't you do some of these checks before you start looking at the phi by asking for the predecessors of the current block?  It could make this a little simpler.  You have one function that checks for all of the conditions on the basic block.  If that says it is okay, then move on to looking at the individual phi instructions.

source/opt/if_conversion.cpp
```diff
@@ -0,0 +1,209 @@
...
+        if (dominators->Dominates(&block, inc0)) return false;
+
+        ir::BasicBlock* inc1 = GetIncomingBlock(phi, 1u);
+        if (dominators->Dominates(&block, inc1)) return false;
+
+        // All phis will have the same common dominator, so cache the result
+        // for this block. If there is no common dominator, we can stop
+        // traversing.
+        if (!common) common = CommonDominator(inc0, inc1, *dominators);
```
s-perron: This could be computed before the call to `WhileEachPhiInst` and passed in.  The CFG can give the predecessors by calling `cfg.preds(block_id)`.

source/opt/if_conversion.cpp
```diff
@@ -0,0 +1,209 @@
...
+        if ((then_block == &block && inc0 == common) ||
+            dominators->Dominates(then_block, inc0)) {
+          true_value = GetIncomingValue(phi, 0u);
+          false_value = GetIncomingValue(phi, 1u);
+        } else {
+          true_value = GetIncomingValue(phi, 1u);
+          false_value = GetIncomingValue(phi, 0u);
+        }
+
+        // If either incoming value is defined is a block that does not dominate
```
s-perron: s/is defined is/is defined in/

source/opt/if_conversion.cpp
```diff
@@ -0,0 +1,209 @@
...
+        if (false_def_block && !dominators->Dominates(false_def_block, &block))
+          return true;
+
+        analysis::Type* data_ty =
+            context()->get_type_mgr()->GetType(true_value->type_id());
+        if (analysis::Vector* vec_data_ty = data_ty->AsVector()) {
+          condition = SplatCondition(vec_data_ty, &block, iter, condition);
+        }
+
+        std::unique_ptr<ir::Instruction> select(new ir::Instruction(
```
s-perron: Try adding something to the IRBuilder.  Should make this a little cleaner.

source/opt/if_conversion.cpp
```diff
@@ -0,0 +1,209 @@
...
+  analysis::Vector bool_vec_ty(&bool_ty, vec_data_ty->element_count());
+  uint32_t bool_vec_id =
+      context()->get_type_mgr()->GetTypeInstruction(&bool_vec_ty);
+  std::vector<ir::Operand> ops;
+  for (size_t i = 0; i != vec_data_ty->element_count(); ++i) {
+    ops.emplace_back(SPV_OPERAND_TYPE_ID,
+                     std::initializer_list<uint32_t>{cond});
+  }
+  uint32_t splat_id = TakeNextId();
+  std::unique_ptr<ir::Instruction> splat(new ir::Instruction(
```
s-perron: IRBuilder again.

source/opt/if_conversion.cpp
```diff
@@ -0,0 +1,209 @@
...
+  return GetBlock(phi->GetSingleWordInOperand(in_index));
+}
+
+ir::Instruction* IfConversion::GetIncomingValue(ir::Instruction* phi,
+                                                uint32_t predecessor) {
+  uint32_t in_index = 2 * predecessor;
+  return get_def_use_mgr()->GetDef(phi->GetSingleWordInOperand(in_index));
+}
+
+ir::BasicBlock* IfConversion::CommonDominator(
```
s-perron: Would this be better in the Dominator Tree class?  I could see many other things using this.

source/opt/dominator_analysis.cpp
```diff
@@ -0,0 +1,41 @@
...
+  if (!b1 || !b2) return nullptr;
+
+  std::unordered_set<ir::BasicBlock*> seen;
+  ir::BasicBlock* block = b1;
+  while (block && seen.insert(block).second) {
+    block = ImmediateDominator(block);
+  }
+
+  block = b2;
+  while (block && seen.insert(block).second) {
```
s-perron: Why do you need to insert here?  This should be a tree, so no issues with loops.

s-perron:
Can be merged if all of the builds pass.

alan-baker:
Rebased, squashed and committed as 2e93e806e454e2cfefabf217320038188c6c1fb6

Fixes #711

Issue #1116, opened 2017-12-19
Vulkan: Images: OpSampledImage must only consume an “Image” operand whose type has its “Sampled” operand set to 1.


dneto0:
From Vulkan env spec (Appendix A), subsection "Validation rules within a module"

Issue #1115, opened 2017-12-19
Vulkan: Images: OpTypeImage must declare a scalar 32-bit float or 32-bit integer type for the “Sampled Type”. (RelaxedPrecision can be applied to a sampling instruction and to the variable holding the result of a sampling instruction.)


dneto0:
From Vulkan appendix A, section "Validation rules within a module"

OpTypeImage must declare a scalar 32-bit float or 32-bit integer type for the “Sampled Type”. (RelaxedPrecision can be applied to a sampling instruction and to the variable holding the result of a sampling instruction.)

PR #1227, opened 2018-01-23
Added Vulkan-specifc checks to image validation

atgoo:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1115
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1116

Implemented Vulkan-specific rules:
- OpTypeImage must declare a scalar 32-bit float or 32-bit integer type
for the “Sampled Type”.
- OpSampledImage must only consume an “Image” operand whose type has its
“Sampled” operand set to 1.

### DIFF
--- a/source/spirv_target_env.cpp
+++ b/source/spirv_target_env.cpp
@@ -146,3 +146,28 @@ bool spvParseTargetEnv(const char* s, spv_target_env* env) {
     return false;
   }
 }
+
+bool spvIsVulkanEnv(spv_target_env env) {
+  switch (env) {
+    case SPV_ENV_UNIVERSAL_1_0:
+    case SPV_ENV_OPENCL_1_2:
+    case SPV_ENV_OPENCL_EMBEDDED_1_2:
+    case SPV_ENV_OPENCL_2_0:
+    case SPV_ENV_OPENCL_EMBEDDED_2_0:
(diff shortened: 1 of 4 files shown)


atgoo:
Passes Vulkan CTS master.

source/spirv_target_env.cpp
```diff
@@ -146,3 +146,28 @@ bool spvParseTargetEnv(const char* s, spv_target_env* env) {
     return false;
   }
 }
+
+bool spvIsVulkanEnv(spv_target_env env) {
```
antiagainst: `return env == SPV_ENV_VULKAN_1_0`?

atgoo: What if someone adds SPV_ENV_VULKAN_1_1?

antiagainst: Then this code will need to be changed anyway, right? :)

antiagainst: I've no strong opinion towards this, so +2'd previously. Your choice. :)

atgoo: The evil plan is that the switch won't compile if you add a new env without putting it either into "Vulkan" or "not Vulkan".

antiagainst: That's a good point!

Issue #1091, opened 2017-12-13
Ban floating point case for OpAtomicLoad, OpAtomicExchange, OpAtomicCompareExchange


dneto0:
In graphics (Shader) environments, these instructions only operate on scalar integers.  Ban the floating point case.

OpenCL supports atomic_float

dneto0:
NV_shader_atomic_float has atomic  operations on floats, but that would be an extension that would have to be declared in the SPIR-V module.
https://www.khronos.org/registry/OpenGL/extensions/NV/NV_shader_atomic_float.txt

PR #1229, opened 2018-01-24
Validator: restricted some atomic ops for shaders

atgoo:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1091

Ban floating point case for OpAtomicLoad, OpAtomicExchange,
OpAtomicCompareExchange. In graphics (Shader) environments, these
instructions only operate on scalar integers. Ban the floating point
case. OpenCL supports atomic_float.


atgoo:
Passes Vulkan CTS master.

PR #1230, opened 2018-01-24
Initial support for loop unrolling

davors72:
This patch contains the initial support for loop unrolling in the form of a series of utility classes which perform the actual unrolling. Before performing the optimization we check that the incoming IR meets a series of assumptions made, some of these will be softened in subsequent patches.

The unroller will only unroll code which:

- Has had its induction variable as a phi node (i.e has run through opt::LocalMultiStoreElimPass prior)

- Induction variable is in a canonicalised form with the initial value, final value, and step all being constant integers, and that the step is an OpIAdd and the condition is OpSLessThan.

- That thee are no Phi instructions (other than the induction variable) which are used both inside the loop and outside.

-  That the loop is in the Structured Control Flow form as defined by the SPIR-V spec.

### DIFF
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -58,6 +58,7 @@ add_library(SPIRV-Tools-opt
   local_ssa_elim_pass.h
   log.h
   loop_descriptor.h
+  loop_unroller.h
   loop_utils.h
   make_unique.h
   mem_pass.h
@@ -130,6 +131,7 @@ add_library(SPIRV-Tools-opt
   local_ssa_elim_pass.cpp
   loop_descriptor.cpp
   loop_utils.cpp
+  loop_unroller.cpp
   mem_pass.cpp
   merge_return_pass.cpp
   module.cpp
--- a/source/opt/dominator_tree.h
+++ b/source/opt/dominator_tree.h
@@ -224,6 +224,19 @@ class DominatorTree {
     return true;
   }
 
+  // Applies the std::function |func| to all nodes in the dominator tree from
+  // |node| downwards. The boolean return from |func| is used to determine
+  // whether or not the children should also be traversed. Tree nodes are
+  // visited in a depth first pre-order.
+  void VisitChildrenIf(std::function<bool(DominatorTreeNode*)> func,
+                       iterator node) {
+    if (func(&*node)) {
+      for (auto n : *node) {
+        VisitChildrenIf(func, n->df_begin());
+      }
+    }
+  }
+
   // Returns the DominatorTreeNode associated with the basic block |bb|.
   // If the |bb| is unknown to the dominator tree, it returns null.
   inline DominatorTreeNode* GetTreeNode(ir::BasicBlock* bb) {
--- a/source/opt/ir_builder.h
+++ b/source/opt/ir_builder.h
@@ -16,9 +16,9 @@
 #define LIBSPIRV_OPT_IR_BUILDER_H_
 
 #include "opt/basic_block.h"
+#include "opt/constants.h"
 #include "opt/instruction.h"
 #include "opt/ir_context.h"
-
 namespace spvtools {
 namespace opt {
 
@@ -136,6 +136,12 @@ class InstructionBuilder {
     return AddInstruction(std::move(select));
   }
 
+  // Adds a signed int32 constant to the binary.
+  // The |value| parameter is the constant value to be added.
+  ir::Instruction* Add32BitSignedIntegerConstant(int32_t value) {
+    return Add32BitConstantInteger<int32_t>(value, true);
+  }
+
   // Create a composite construct.
   // |type| should be a composite type and the number of elements it has should
   // match the size od |ids|.
@@ -150,6 +156,38 @@ class InstructionBuilder {
         new ir::Instruction(GetContext(), SpvOpCompositeConstruct, type,
                             GetContext()->TakeNextId(), ops));
     return AddInstruction(std::move(construct));
+  }
+  // Adds an unsigned int32 constant to the binary.
+  // The |value| parameter is the constant value to be added.
+  ir::Instruction* Add32BitUnsignedIntegerConstant(uint32_t value) {
+    return Add32BitConstantInteger<uint32_t>(value, false);
+  }
+
+  // Adds either a signed or unsigned 32 bit integer constant to the binary
+  // depedning on the |sign|. If |sign| is true then the value is added as a
+  // signed constant otherwise as an unsigned constant. If |sign| is false the
+  // value must not be a negative number.
+  template <typename T>
+  ir::Instruction* Add32BitConstantInteger(T value, bool sign) {
+    // Assert that we are not trying to store a negative number in an unsigned
+    // type.
+    if (!sign)
+      assert(value > 0 &&
+             "Trying to add a signed integer with an unsigned type!");
+
+    // Get or create the integer type.
+    analysis::Integer int_type(32, sign);
+
+    // Even if the value is negative we need to pass the bit pattern as a
+    // uint32_t to GetConstant.
+    uint32_t word = value;
+
+    // Create the constant value.
+    const opt::analysis::Constant* constant =
+        GetContext()->get_constant_mgr()->GetConstant(&int_type, {word});
+
+    // Create the OpConstant instruction using the type and the value.
+    return GetContext()->get_constant_mgr()->GetDefiningInstruction(constant);
   }
 
   ir::Instruction* AddCompositeExtract(
--- a/source/opt/loop_descriptor.cpp
+++ b/source/opt/loop_descriptor.cpp
@@ -18,6 +18,7 @@
 #include <utility>
 #include <vector>
 
+#include "constants.h"
 #include "opt/cfg.h"
 #include "opt/dominator_tree.h"
 #include "opt/ir_builder.h"
@@ -29,6 +30,107 @@
 namespace spvtools {
 namespace ir {
 
+// Takes in a phi instruction |induction| and the loop |header| and returns the
+// step operation of the loop.
+ir::Instruction* Loop::GetInductionStepOperation(
+    const ir::Loop* loop, const ir::Instruction* induction) const {
+  // Induction must be a phi instruction.
+  assert(induction->opcode() == SpvOpPhi);
+
+  ir::Instruction* step = nullptr;
+
+  opt::analysis::DefUseManager* def_use_manager = context_->get_def_use_mgr();
+
+  // Traverse the incoming operands of the phi instruction.
+  for (uint32_t operand_id = 1; operand_id < induction->NumInOperands();
+       operand_id += 2) {
+    // Incoming edge.
+    ir::BasicBlock* incoming_block =
+        context_->cfg()->block(induction->GetSingleWordInOperand(operand_id));
+
+    // Check if the block is dominated by header, and thus coming from within
+    // the loop.
+    if (loop->IsInsideLoop(incoming_block)) {
+      step = def_use_manager->GetDef(
+          induction->GetSingleWordInOperand(operand_id - 1));
+      break;
+    }
+  }
+
+  if (!step || !IsSupportedStepOp(step->opcode())) {
+    return nullptr;
+  }
+
+  return step;
+}
+
+// Returns true if the |step| operation is an induction variable step operation
+// which is currently handled.
+bool Loop::IsSupportedStepOp(SpvOp step) const {
+  switch (step) {
+    case SpvOp::SpvOpISub:
+    case SpvOp::SpvOpIAdd:
+      return true;
+    default:
+      return false;
+  }
+}
+
+bool Loop::IsSupportedCondition(SpvOp condition) const {
+  switch (condition) {
+    // <
+    case SpvOp::SpvOpULessThan:
+    case SpvOp::SpvOpSLessThan:
+    // >
+    case SpvOp::SpvOpUGreaterThan:
+    case SpvOp::SpvOpSGreaterThan:
+      return true;
+    default:
+      return false;
+  }
+}
+
+// Extract the initial value from the |induction| OpPhi instruction and store it
+// in |value|. If the function couldn't find the initial value of |induction|
+// return false.
+bool Loop::GetInductionInitValue(const ir::Loop* loop,
+                                 const ir::Instruction* induction,
+                                 int64_t* value) const {
+  ir::Instruction* constant_instruction = nullptr;
+  opt::analysis::DefUseManager* def_use_manager = context_->get_def_use_mgr();
+
+  for (uint32_t operand_id = 0; operand_id < induction->NumInOperands();
+       operand_id += 2) {
+    ir::BasicBlock* bb = context_->cfg()->block(
+        induction->GetSingleWordInOperand(operand_id + 1));
+
+    if (!loop->IsInsideLoop(bb)) {
+      constant_instruction = def_use_manager->GetDef(
+          induction->GetSingleWordInOperand(operand_id));
+    }
+  }
+
+  if (!constant_instruction) return false;
+
+  const opt::analysis::Constant* constant =
+      context_->get_constant_mgr()->FindDeclaredConstant(
+          constant_instruction->result_id());
+  if (!constant) return false;
+
+  if (value) {
+    const opt::analysis::Integer* type =
+        constant->AsIntConstant()->type()->AsInteger();
+
+    if (type->IsSigned()) {
+      *value = constant->AsIntConstant()->GetS32BitValue();
+    } else {
+      *value = constant->AsIntConstant()->GetU32BitValue();
+    }
+  }
+
+  return true;
+}
+
 Loop::Loop(IRContext* context, opt::DominatorAnalysis* dom_analysis,
            BasicBlock* header, BasicBlock* continue_target,
            BasicBlock* merge_target)
@@ -37,12 +139,11 @@ Loop::Loop(IRContext* context, opt::DominatorAnalysis* dom_analysis,
       loop_continue_(continue_target),
       loop_merge_(merge_target),
       loop_preheader_(nullptr),
-      parent_(nullptr) {
+      parent_(nullptr),
+      loop_is_marked_for_removal_(false) {
   assert(context);
   assert(dom_analysis);
   loop_preheader_ = FindLoopPreheader(dom_analysis);
-  AddBasicBlockToLoop(header);
-  AddBasicBlockToLoop(continue_target);
 }
 
 BasicBlock* Loop::FindLoopPreheader(opt::DominatorAnalysis* dom_analysis) {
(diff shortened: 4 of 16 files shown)


source/opt/loop_descriptor.cpp
```diff
@@ -153,5 +250,122 @@ void LoopDescriptor::PopulateList(const Function* f) {
...
+
+  assert(condition->opcode() == SpvOpSLessThan);
+
+  // The right hand side operand of the operation.
+  const ir::Instruction* rhs_inst =
+      def_use_manager->GetDef(condition->GetSingleWordOperand(3));
+
+  // Exit out if we couldn't resolve the rhs to be a constant integer.
+  uint32_t condition_value = 0;
+  if (!GetConstant(rhs_inst, ir_context_, &condition_value)) return false;
```
s-perron: I think you could use the constant manager here.

```
const Constant* upper_bound = const_mgr->FindDeclaredConstant(condition->GetSingleWordOperand(3));
if (!upper_bound) return false;
// must be integer because of the opcode on the condition.
uint32_t condition_value = upper_bound->AsIntegerConstant()->GetUInt32BitValue(); 
```

If we can get rid of GetConstant, we can reduce the amount of code.

source/opt/loop_descriptor.h
```diff
@@ -154,6 +159,20 @@ class Loop {
...
+                              size_t* iterations) const;
+
+  inline bool HasUnrollLoopControl() const {
+    assert(loop_header_);
+    if (!loop_header_->GetLoopMergeInst()) return false;
+
+    return loop_header_->GetLoopMergeInst()->GetSingleWordOperand(2) == 1;
+  }
+
+  ir::BasicBlock* FindConditionBlock(const ir::Function& function);
```
s-perron: Should document the conditions on the functions above.

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
+// Copyright (c) 2018 Google Inc.
```
s-perron: Use "Google LLC".  We were just told to use that for new files.

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+// Implements loop util unrolling functionality for fully and partially
+// unrolling loops. Given a factor it will duplicate the loop that many times,
+// appending each one to the end of the old loop and removing backedges, to
+// create a new unrolled loop.
+//
+// 1 - User calls LoopUtils::FullyUnroll or LoopUtils::PartiallyUnroll with a
+// loop they wish to unroll. LoopUtils::CanPerformUnroll is used to
+// validate that a given loop can be unrolled. That method (alongwith the the
+// constructor of loop) checks that the IR is in the expected canonicalised
+// format.
```
s-perron: Is the expected format described anywhere?

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+//
+// 5 - If it is being partially unrolled: if the unrolling factor leaves the
+// loop with an even number of bodies with respect to the number of loop
+// iterations then step 3 is all that is needed. If it is uneven then we need to
+// duplicate the loop completely and unroll the duplicated loop to cover the
+// "uneven"  part and adjust the first loop to cover only the "even" part. For
+// instance if you request an unroll factor of 2 on a loop with 10 iterations
+// then copying the body twice would leave you with three bodies in the loop
+// while  the loop still iterates over each 4 times. So we make two loops one
+// iterating once then a second loop of three iterating 3 times.
```
s-perron: This is a little confusing to me. I think of the unroll factor as including the original copy of the code.  Unrolling a loop with 10 iteration by a factor of 2 means you end up with a loop that has 5 iterations, no residual.  Looking it up, it seems like the term gets used both ways.  I'm not sure which definition is more common, but GCC seems to use it the way I understand it (https://gcc.gnu.org/onlinedocs/gcc/Loop-Specific-Pragmas.html) because an unroll factor of 1 blocks unrolling.

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+
+  LoopUnrollerUtilsImpl(ir::IRContext* c, ir::Function* function)
+      : ir_context_(c),
+        function_(*function),
+        loop_condition_block_(nullptr),
+        loop_induction_variable_(nullptr),
+        number_of_loop_iterations_(0) {}
+
+  // Unroll the |loop| by given |factor| by copying the whole body |factor|
+  // times. The resulting basicblock structure will remain a loop.
```
s-perron: See my previous comment about the unrolling factor.

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+  void PartiallyUnroll(ir::Loop*, size_t factor);
+
+  // If partially unrolling the |loop| would leave the loop with too many bodies
+  // for its number of iterations then this method should be used. This method
+  // will duplicate the |loop| compltetely, makingthe duplicated loop the
```
s-perron: missing a spec "makingthe".

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+  // successor of the originals merge block. The original loop will have its
```
s-perron: original's

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+  // condition changed to loop over the uneven part and the duplicate will be
```
s-perron: A better name than "uneven part" might be the "residual".  I think this is commonly used in the compiler literature.  Just something to consider.  If you are following the language defined in a particular paper, add a comment at the top that you are following the implementation in that paper.

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+
+  // Remove the OpConditionalBranch instruction inside |conditional_block| used
+  // to branch to either exit or continue the loop and replace it with an
+  // unconditional OpBranch to block |new_target|.
+  void FoldConditionBlock(ir::BasicBlock* condtion_block, uint32_t new_target);
+
+  // Add all blocks_to_add_ to function_ at the |insert_point|.
+  void AddBlocksToFunction(const ir::BasicBlock* insert_point);
+
+  void DuplicateLoop(ir::Loop* old_loop, ir::Loop* new_loop);
```
s-perron: Might need some more detail on what this function does.  Where does the code for the new_loop go?

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+  // Using the map built by AssignNewResultIds, for each instruction in
+  // |basic_block| use
+  // that map to substitute the IDs used by instructions (in the operands) with
+  // the new ids.
+  void RemapOperands(ir::BasicBlock* basic_block);
+
+  // Copy the whole body of the loop, all blocks dominated by the |loop| header
+  // and not dominated by the |loop| merge. The copied body will be linked to by
+  // the old |loop| continue block and the new body will link to the |loop|
+  // header via the new continue block. |eliminate_conditions| is used to decid
```
s-perron: s/decid/decide

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+  // Copy a given |block_to_copy| in the |loop| and record the mapping of the
+  // old/new ids. |preserve_instructions| determines whether or not the method
+  // will modify (other than result_id) instructions which are copied.
+  void CopyBasicBlock(ir::Loop* loop, const ir::BasicBlock* block_to_copy,
+                      bool preserve_instructions);
+
+  // The actual implementation of the unroll step. Unrolls |loop| by given
+  // |factor| by copying the body by |factor| times. Also propagates the
+  // induction variable value throughout the copies.
+  void Unroll(ir::Loop* loop, size_t);
```
s-perron: might want to add the name |factor| after size_t since it is referenced in the comments.

test/opt/loop_optimizations/unroll_simple.cpp
```diff
@@ -0,0 +1,608 @@
...
+%19 = OpConstant %7 1
+%20 = OpTypeVector %12 4
+%21 = OpTypePointer Output %20
+%3 = OpVariable %21 Output
+%2 = OpFunction %5 None %6
+%22 = OpLabel
+%4 = OpVariable %16 Function
+OpBranch %23
+%23 = OpLabel
+%24 = OpPhi %7 %9 %22 %103 %102
```
alan-baker: This shouldn't be a phi anymore. %23 is not reachable from %102. Most of the phis look wrong. The subsequent ones reference %22, but there is no edge from %22.

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+    loop_condition_block_ = state_.new_condition_block;
+  }
+  assert(loop_condition_block_);
+
+  loop_induction_variable_ = loop->FindInductionVariable(loop_condition_block_);
+  assert(loop_induction_variable_);
+
+#ifndef NDEBUG
+  bool found =
+#endif
```
s-perron: I'm assuming you did this because of the unused variable error for a release build.  I think you can add a line:
```
bool found = ...;
found;  // So the release build does not complain.
assert(found);
```
I think this works and it is a little cleaner.

test/opt/loop_optimizations/unroll_simple.cpp
```diff
@@ -0,0 +1,608 @@
...
+OpBranch %37
+%37 = OpLabel
+%38 = OpAccessChain %18 %4 %34
+OpStore %38 %17
+OpBranch %39
+%39 = OpLabel
+%40 = OpIAdd %7 %34 %19
+OpBranch %42
+%42 = OpLabel
+%43 = OpPhi %7 %9 %22 %49 %48
```
alan-baker: This block has %39 as a predecessor, but it is not in the list incoming blocks. %48 doesn't reach this block. The value for the phi should %40 I pretty sure.

test/opt/loop_optimizations/unroll_simple.cpp
davors72: Yeah looks like that one should be removed when the loop is closed. For the others they're all unused so I was hoping we could schedule a cleanup pass after the loop unroller runs to kill them

alan-baker: The IR at the end of the pass should be valid. It is unfortunate that the validator doesn't flag these phis currently, but I will try to get that prioritized. This would be a different situation if the phis were valid IR, but dead code.

greg-lunarg: @alan-baker  We have someone that is currently working on validating exactly these phi issues, ie all and only block predecessors appear in block's phi exactly once.

source/opt/loop_unroller.cpp
zeux: fwiw `(void)found;` is a canonical way to suppress unused variable warning; `found;` causes warnings on some compilers because it's an unused expression

source/opt/dominator_tree.h
```diff
@@ -224,6 +224,17 @@ class DominatorTree {
     return true;
   }
 
+  // Applies the std::function |func| to all nodes in the dominator tree.
```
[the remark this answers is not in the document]
dnovillo: I'm confused at the "If" in the name.  The function will always visit the node even if the callback returns false.

source/opt/loop_descriptor.cpp
```diff
@@ -153,5 +250,122 @@ void LoopDescriptor::PopulateList(const Function* f) {
...
+
+  // If iterations is non null then store the value in that.
+  if (iterations) {
+    *iterations = (condition_value / step_value) - init_value;
+  }
+
+  return true;
+}
+
+ir::Instruction* Loop::FindInductionVariable(
```
dnovillo: There could be multiple induction variables in a loop.  Is this based on some existing algorithm?  There are several references in the literature that deal with induction variable detection and elimination.  It would be very useful if this was based on one of them.  Easier to document and maintain.

source/opt/loop_unroller.cpp
dnovillo: This is what is commonly understood, yes.

source/opt/loop_descriptor.cpp
dnovillo: Additionally, induction variables could have a positive or negative step.  This code seems to assume and require always monotonically increasing loops.  This should be documented or perhaps it's not too difficult to support?

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,793 @@
...
+  bool branching_assumtion =
+      branch.opcode() == SpvOpBranch &&
+      branch.GetSingleWordInOperand(0) == loop->GetHeaderBlock()->id();
+  if (!branching_assumtion) {
+    return false;
+  }
+
+  // Make sure the induction is the only phi instruction we have in the loop
+  // header.
+  for (const ir::Instruction& inst : *loop->GetHeaderBlock()) {
```
dnovillo: This seems too restrictive.  Why not just identify the phi instruction for the detected induction variable(s)?

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,907 @@
...
+// 4 - If the loop is to be fully unrolled then it is simply closed after step
+// 3, with the OpLoopMerge being deleted, the backedge removed, and the
+// condition blocks folded.
+//
+// 5 - If it is being partially unrolled: if the unrolling factor leaves the
+// loop with an even number of bodies with respect to the number of loop
+// iterations then step 3 is all that is needed. If it is uneven then we need to
+// duplicate the loop completely and unroll the duplicated loop to cover the
+// residual part and adjust the first loop to cover only the "even" part. For
+// instance if you request an unroll factor of 2 on a loop with 10 iterations
```
s-perron: Can we fix this up?  The more common understanding is that an unroll factor of 2 on a loop with 10 iterations would give a single loop with the original body twice, and 5 iterations.
```
for( int i = 0; i < 10; ++i ) {
   S(i);
}
```
unrolled with a factor of 2 becomes
```
for( int i = 0; i < 10; i += 2 ) {
   S(i);
   S(i+1);
}
```

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,910 @@
...
+  original_phi->SetInOperand(phi_index, {phi_label});
+}
+
+// Fully unroll the loop by partially unrolling it by the number of loop
+// iterations minus one for the body already accounted for.
+void LoopUnrollerUtilsImpl::FullyUnroll(ir::Loop* loop) {
+  // We unroll the loop by number of iterations in the loop.
+  Unroll(loop, number_of_loop_iterations_);
+
+  // The first condition block is perserved until now so it can be copied.
```
s-perron: s/perserved/preserved

source/opt/loop_unroller.cpp
```diff
@@ -0,0 +1,910 @@
...
+// Duplicate the |loop| body |factor| number of times while keeping the loop
+// backedge intact.
+void LoopUnrollerUtilsImpl::Unroll(ir::Loop* loop, size_t factor) {
+  state_ = LoopUnrollState{loop_induction_variable_, loop->GetLatchBlock(),
+                           loop_condition_block_};
+  for (size_t i = 0; i < factor - 1; ++i) {
+    CopyBody(loop, true);
+  }
+
+  uint32_t phi_index = GetPhiIndexFromLabel(state_.previous_continue_block_,
```
s-perron: If I got this right, this section is going back to the top to rewrite the OpPhi's value that comes from the backedge.  The issue that @dnovillo brought up earlier (multiple induction variables) is that there may be multiple OpPhi instructions that need to be rewritten.  Where are the other OpPhi instructions rewritten?

test/opt/loop_optimizations/unroll_assumptions.cpp
```diff
@@ -0,0 +1,318 @@
+// Copyright (c) 2018 Google Inc.
```
s-perron: LLC not Inc.

source/opt/loop_unroller.cpp
davors72: They won't be rewritten in this patch as it is, we check in the CanPerformUnroll function that the only phi used in the loop is the induction variable. We have a solution to rewrite the phis after we transform the loop into LCSAA and give it a single preheader. After doing that we can fairly easily extend the induction variable rewriting part to rewrite all the phis found. 

We could add that to this patch if required, otherwise it will be a followup PR. I added a note to the CanPerformUnroll method in LoopUtils.h to list the constraints we currently enforce.

source/opt/loop_unroller.cpp
davors72: Sorry that comment is out of date, the implementation now reflects the above

test/opt/loop_optimizations/unroll_simple.cpp
```diff
@@ -0,0 +1,2228 @@
...
+%12 = OpLabel
+OpBranch %13
+%13 = OpLabel
+OpBranch %18
+%18 = OpLabel
+%19 = OpSLessThan %9 %7 %8
+OpBranch %20
+%20 = OpLabel
+%21 = OpIEqual %9 %7 %10
+OpSelectionMerge %22 None
```
alan-baker: This isn't correct. I know it passes the validator. I've filed #1281 to fix that. I would suggest not handling breaks for now at least. Add a rule to CanPerformUnroll to check that the loop has only one exit block. Continues are also potentially troublesome. Structured control flow rules do not require an OpSelectionMerge in the case of breaks and continues. I would suggest adding a check in CanPerformUnroll to verify that any conditional branches to the continue block are preceeded by an OpSelectionMerge. If they are not, I would refuse to unroll them for now.

source/opt/ir_builder.h
```diff
@@ -151,6 +157,39 @@ class InstructionBuilder {
...
+  ir::Instruction* Add32BitUnsignedIntegerConstant(uint32_t value) {
+    return Add32BitConstantInteger<uint32_t>(value, false);
+  }
+
+  // Adds either a signed or unsigned 32 bit integer constant to the binary
+  // depedning on the |sign|. If |sign| is true then the value is added as a
+  // signed constant otherwise as an unsigned constant.
+  template <typename T>
+  ir::Instruction* Add32BitConstantInteger(T value, bool sign) {
+    if (sign) assert(value > 0);
```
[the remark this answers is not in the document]
davors72: Yeah that should be !sign, it's because the T doesn't have to be unsigned so if you try and store a negative value with an unsigned spirv type it will just end up storing some huge number. Updated now

s-perron:
Looking at it.  I think we can get this PR through, but we will need some follow up issues:

- [ ] Support multiple OpPhis in the loop header
- [ ] Support OpSLessThanEqual and the like in the loop condition.

s-perron:
I think all of Alan's comments and Diego's comments were at least initially addressed.  This is a large PR hitting 15 files.  It does not enable loop unrolling by default, so that lowers the risk.  I will merge it after the tests pass, and we can address any other concerns in a follow up PR.

@alan-baker @dnovillo FYI.

dnovillo:
Thanks @s-perron.  Yes, I'm fine with merging the PR as-is.  Further changes can be addressed in future PRs.

source/opt/CMakeLists.txt
```diff
@@ -128,6 +129,7 @@ add_library(SPIRV-Tools-opt
   local_ssa_elim_pass.cpp
   loop_descriptor.cpp
   loop_utils.cpp
+  loop_unroller.cpp
```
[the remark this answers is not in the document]
davors72: Yeah just caught that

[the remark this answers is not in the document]
davors72: It's there now

PR #1231, opened 2018-01-24
Add general folding infrastructure.

s-perron:
Create the folding engine that will

1) attempt to fold an instruction.
2) iterates on the folding so small folding rules can be easily combined.
3) insert new instructions when needed.

I've added the minimum number of rules needed to test the features above.

Contributes to #1164.

### DIFF
--- a/source/opt/fold.cpp
+++ b/source/opt/fold.cpp
@@ -15,6 +15,8 @@
 #include "fold.h"
 
 #include "def_use_manager.h"
+#include "folding_rules.h"
+#include "ir_builder.h"
 #include "ir_context.h"
 
 #include <cassert>
@@ -178,6 +180,42 @@ uint32_t OperateWords(SpvOp opcode,
       assert(false && "Invalid number of operands");
       return 0;
   }
+}
+
+bool FoldInstructionInternal(ir::Instruction* inst,
+                             std::function<uint32_t(uint32_t)> id_map) {
+  ir::IRContext* context = inst->context();
+  ir::Instruction* folded_inst = FoldInstructionToConstant(inst, id_map);
+  if (folded_inst != nullptr) {
+    inst->SetOpcode(SpvOpCopyObject);
+    inst->SetInOperands({{SPV_OPERAND_TYPE_ID, {folded_inst->result_id()}}});
+    return true;
+  }
+
+  SpvOp opcode = inst->opcode();
+  analysis::ConstantManager* const_manger = context->get_constant_mgr();
+
+  std::vector<const analysis::Constant*> constants;
+  for (uint32_t i = 0; i < inst->NumInOperands(); i++) {
+    const ir::Operand* operand = &inst->GetInOperand(i);
+    if (operand->type != SPV_OPERAND_TYPE_ID) {
+      constants.push_back(nullptr);
+    } else {
+      uint32_t id = id_map(operand->words[0]);
+      inst->SetInOperand(i, {id});
+      const analysis::Constant* constant =
+          const_manger->FindDeclaredConstant(id);
+      constants.push_back(constant);
+    }
+  }
+
+  static FoldingRules* rules = new FoldingRules();
+  for (FoldingRule rule : rules->GetRulesForOpcode(opcode)) {
+    if (rule(inst, constants)) {
+      return true;
+    }
+  }
+  return false;
 }
 
 }  // namespace
@@ -624,13 +662,26 @@ bool IsFoldableType(ir::Instruction* type_inst) {
 
 ir::Instruction* FoldInstruction(ir::Instruction* inst,
                                  std::function<uint32_t(uint32_t)> id_map) {
-  ir::Instruction* folded_inst = FoldInstructionToConstant(inst, id_map);
-  if (folded_inst != nullptr) {
-    return folded_inst;
+  ir::IRContext* context = inst->context();
+  bool modified = false;
+  std::unique_ptr<ir::Instruction> folded_inst(inst->Clone(context));
+  while (FoldInstructionInternal(&*folded_inst, id_map)) {
+    modified = true;
   }
 
-  // TODO: Add other folding opportunities that do not necessarily fold to a
-  // constant.
+  if (modified) {
+    if (folded_inst->opcode() == SpvOpCopyObject) {
+      analysis::DefUseManager* def_use_mgr = context->get_def_use_mgr();
+      return def_use_mgr->GetDef(folded_inst->GetSingleWordInOperand(0));
+    } else {
+      InstructionBuilder ir_builder(
+          context, inst,
+          ir::IRContext::kAnalysisDefUse |
+              ir::IRContext::kAnalysisInstrToBlockMapping);
+      folded_inst->SetResultId(context->TakeNextId());
+      return ir_builder.AddInstruction(std::move(folded_inst));
+    }
+  }
   return nullptr;
 }
 
--- a/source/opt/folding_rules.cpp
+++ b/source/opt/folding_rules.cpp
@@ -1,0 +1,129 @@
+// Copyright (c) 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "folding_rules.h"
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+const uint32_t kExtractCompositeIdInIdx = 0;
+const uint32_t kInsertObjectIdInIdx = 0;
+const uint32_t kInsertCompositeIdInIdx = 1;
+
+FoldingRule IntMultiplyBy1() {
+  return [](ir::Instruction* inst,
+            const std::vector<const analysis::Constant*>& constants) {
+    assert(inst->opcode() == SpvOpIMul && "Wrong opcode.  Should be OpIMul.");
+    for (uint32_t i = 0; i < 2; i++) {
+      if (constants[i] == nullptr) {
+        continue;
+      }
+      const analysis::IntConstant* int_constant = constants[i]->AsIntConstant();
+      if (int_constant->GetU32BitValue() == 1) {
+        inst->SetOpcode(SpvOpCopyObject);
+        inst->SetInOperands(
+            {{SPV_OPERAND_TYPE_ID, {inst->GetSingleWordInOperand(1 - i)}}});
+        return true;
+      }
+    }
+    return false;
+  };
+}
+
+FoldingRule InsertFeedingExtract() {
+  return [](ir::Instruction* inst,
+            const std::vector<const analysis::Constant*>&) {
+    assert(inst->opcode() == SpvOpCompositeExtract &&
+           "Wrong opcode.  Should be OpCompositeExtract.");
+    analysis::DefUseManager* def_use_mgr = inst->context()->get_def_use_mgr();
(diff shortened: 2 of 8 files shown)


s-perron:
@greg-lunarg FYI.

s-perron:
This is my idea for the design of the folding engine.  I think it makes it easy to add new rules, and they will automatically be combined.  The only part I do not like it the create and deletion of the instructions.  This could create a lot of churn, and lead to a lot of unused result ids.  Also, the allocation and deallocation of the instructions could be slow as well.

If anybody has any ideas for a better design let me know.

source/opt/fold.cpp
```diff
@@ -180,6 +182,53 @@ uint32_t OperateWords(SpvOp opcode,
...
+  analysis::ConstantManager* const_manger = context->get_constant_mgr();
+
+  std::vector<uint32_t> ids;
+  std::vector<const analysis::Constant*> constants;
+  for (uint32_t i = 0; i < inst->NumInOperands(); i++) {
+    const ir::Operand* operand = &inst->GetInOperand(i);
+    if (operand->type != SPV_OPERAND_TYPE_ID) {
+      ids.push_back(0);
+      constants.push_back(nullptr);
+      continue;
```
alan-baker: Just change this to an if/else structure

source/opt/fold.cpp
```diff
@@ -180,6 +182,53 @@ uint32_t OperateWords(SpvOp opcode,
...
+    }
+    ids.push_back(id_map(operand->words[0]));
+    const analysis::Constant* constant =
+        const_manger->FindDeclaredConstant(ids[i]);
+    constants.push_back(constant);
+  }
+
+  static FoldingRules rules;
```
alan-baker: No static function variable of class type.

alan-baker: What about just have another function that switches on opcode and calls an opcode specific folder? The opcode specific folder can then try all the various rules for that opcode.

source/opt/fold.cpp
```diff
@@ -624,14 +673,18 @@ bool IsFoldableType(ir::Instruction* type_inst) {
...
 ir::Instruction* FoldInstruction(ir::Instruction* inst,
                                  std::function<uint32_t(uint32_t)> id_map) {
-  ir::Instruction* folded_inst = FoldInstructionToConstant(inst, id_map);
-  if (folded_inst != nullptr) {
-    return folded_inst;
+  InstructionBuilder<ir::IRContext::kAnalysisDefUse> ir_builder(inst->context(),
+                                                                inst);
+  bool kill_inst = false;
+  ir::Instruction* folded_inst = nullptr;
+  ir::Instruction* next_attempt =
```
alan-baker: You can fold this into the condition test of the while loop.

source/opt/fold.cpp
```diff
@@ -180,6 +182,53 @@ uint32_t OperateWords(SpvOp opcode,
...
+    constants.push_back(constant);
+  }
+
+  static FoldingRules rules;
+  for (FoldingRule rule : rules.GetRulesForOpcode(opcode)) {
+    bool new_inst = false;
+    folded_inst = rule(inst, ids, constants, ir_builder, &new_inst);
+    if (folded_inst != nullptr) {
+      if (*kill_inst) {
+        context->KillInst(inst);
```
alan-baker: Maybe it would be better to leave the dead instructions in place and have a pass of adce clean them up.

source/opt/folding_rules.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+const uint32_t kExtractCompositeIdInIdx = 0;
+const uint32_t kInsertObjectIdInIdx = 0;
+const uint32_t kInsertCompositeIdInIdx = 1;
+
+FoldingRule MultipleBy1 =
```
alan-baker: Change to IntMultiplyBy1

test/opt/fold_test.cpp
```diff
@@ -1,5 +1,4 @@
-// Copyright (c) 2016 Google Inc.
-//
+// Copyright (c) 2016 Google Inc. //
```
alan-baker: Trailing //

source/opt/fold.cpp
s-perron: Think I'll change this to a static pointer, and initialize it with a call to new.  The other options is to move it so it lives somewhere else like the IRContext, but I don't like that because I don't want it used by anybody else.

I have a preference to avoid the switch.  Less code for people to look at if the want to try to understand what is going on.

source/opt/fold.cpp
s-perron: If we can avoid adding the instruction at all, then we won't have to kill.  The issue I have is that I do not want to create a lot of extra instructions.  Also, I prefer deleting instructions that we know are dead without any analysis.  Saves doing the analysis later.

source/opt/fold.cpp
s-perron: Yes, that does look better.

test/opt/fold_test.cpp
s-perron: Fixed.

s-perron:
To fix the issue of inserting and deleting the intermediate instructions, we can create a local instruction.  The folding rules overwrite that instruction when folding, and return a bool to indicate when it was updated.  If you want to fold to use an already existing instruction overwrite the dummy instruction with an OpCopyObject instruction that copies the result of that other instruction.

When folding is done, take appropriate action to insert the new instruction if necessary.

s-perron:
@dneto0 @dnovillo Just want to give this a bump.

source/opt/folding_rules.h
```diff
@@ -0,0 +1,53 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_UTIL_FOLDING_RULES_H_
+#define LIBSPIRV_UTIL_FOLDING_RULES_H_
+
+#include "constants.h"
```
dneto0: Include order:  In a header, system includes come first.
(In a .cpp, your own interface .h comes first, then system headers,...)

source/opt/folding_rules.h
```diff
@@ -0,0 +1,53 @@
...
+using FoldingRule =
+    std::function<bool(ir::Instruction*, const std::vector<uint32_t>&,
+                       const std::vector<const analysis::Constant*>&)>;
+
+class FoldingRules {
+ public:
+  FoldingRules();
+
+  const std::vector<FoldingRule>& GetRulesForOpcode(SpvOp opcode) {
+    static std::vector<FoldingRule> empty_vector;
```
dneto0: I find it a bit dubious to use a static variable here.  What's the benefit over making this a regular data member that is initialized during construction?

Actually, this goes against the style guide.  static vars should be POD.

source/opt/folding_rules.h
```diff
@@ -0,0 +1,53 @@
...
+
+using FoldingRule =
+    std::function<bool(ir::Instruction*, const std::vector<uint32_t>&,
+                       const std::vector<const analysis::Constant*>&)>;
+
+class FoldingRules {
+ public:
+  FoldingRules();
+
+  const std::vector<FoldingRule>& GetRulesForOpcode(SpvOp opcode) {
```
dneto0: Need a method comment.

source/opt/folding_rules.h
```diff
@@ -0,0 +1,53 @@
...
+#include <vector>
+
+namespace spvtools {
+namespace opt {
+
+using FoldingRule =
+    std::function<bool(ir::Instruction*, const std::vector<uint32_t>&,
+                       const std::vector<const analysis::Constant*>&)>;
+
+class FoldingRules {
```
dneto0: Would be nice to have a class comment.

source/opt/folding_rules.h
```diff
@@ -0,0 +1,53 @@
...
+#include "ir_builder.h"
+#include "ir_context.h"
+
+#include <cstdint>
+#include <vector>
+
+namespace spvtools {
+namespace opt {
+
+using FoldingRule =
```
dneto0: This needs an explanation, most of all.  What is the bool for?  What are the arguments"?

source/opt/folding_rules.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+const uint32_t kExtractCompositeIdInIdx = 0;
+const uint32_t kInsertObjectIdInIdx = 0;
+const uint32_t kInsertCompositeIdInIdx = 1;
+
+FoldingRule IntMultipleBy1 =
```
dneto0: I don't think this is POD.   Please make this the return value of a function.

test/opt/fold_test.cpp
```diff
@@ -1101,4 +640,494 @@ INSTANTIATE_TEST_CASE_P(TestCase, BooleanInstructionFoldingTestWithMap,
...
+            "OpReturn\n" +
+            "OpFunctionEnd",
+        2, 0),
+    // Test case 40: fold 1*n
+    InstructionFoldingCase<uint32_t>(
+        Header() + "%main = OpFunction %void None %void_func\n" +
+            "%main_lab = OpLabel\n" +
+            "%n = OpVariable %_ptr_int Function\n" +
+            "%3 = OpLoad %int %n\n" +
+            "%2 = OpIMul %int %int_1 %3\n" +
```
dneto0: Please test the case  where the 1 is the second arg.  Your code handles it but it's not tested here.
 %2 OpIMul %int %3 %int_1

source/opt/folding_rules.cpp
s-perron: Done.

source/opt/folding_rules.h
s-perron: Done

source/opt/folding_rules.h
s-perron: Yes, I wanted to add a detailed description on how to add a new folding rule.

source/opt/folding_rules.h
s-perron: Made it a class member.

test/opt/fold_test.cpp
s-perron: done

s-perron:
@dneto0 Should be good from another look.  Thanks.

source/opt/fold.cpp
```diff
@@ -624,13 +662,24 @@ bool IsFoldableType(ir::Instruction* type_inst) {
...
   }
 
-  // TODO: Add other folding opportunities that do not necessarily fold to a
-  // constant.
+  if (modified) {
+    if (folded_inst->opcode() == SpvOpCopyObject) {
+      analysis::DefUseManager* def_use_mgr = context->get_def_use_mgr();
+      return def_use_mgr->GetDef(folded_inst->GetSingleWordInOperand(0));
+    } else {
+      InstructionBuilder<ir::IRContext::kAnalysisDefUse> ir_builder(context,
```
dneto0: Sorry, InstructionBuilder is no longer a template.  Please rebase and rebuild.

source/opt/folding_rules.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+const uint32_t kExtractCompositeIdInIdx = 0;
+const uint32_t kInsertObjectIdInIdx = 0;
+const uint32_t kInsertCompositeIdInIdx = 1;
+
+FoldingRule IntMultipleBy1() {
```
dnovillo: s/IntMultipleBy1/IntMultiplyBy1/

s-perron: I don't think it is committed already, we were waiting for the Windows build.  I've fixed it.

dneto0:
I'm planning to test on Windows.  I'm having to download various VisualStudio versions.

dneto0:
I manually tested with VS2015 and VS2017.  I'll merge this.

dneto0:
Rebased, squashed, and pushed into master as bc1ec9418

PR #1232, opened 2018-01-25
Add LoopDescriptor as an IRContext analysis.

Naghasan:
This patch the loop descriptors an analysis.

Some function definitions were moved from header to source to avoid circular definition.


Naghasan:
Updated, my clang-format was 3.9 ...

s-perron:
Rebase, squashed, and merged as 6018de81.

Issue #252, opened 2016-06-29
Validator should check capability of the value referenced by semantics-id operand to OpAtomicIAdd


dneto0:
This is the "proper" fix for https://github.com/KhronosGroup/SPIRV-Tools/issues/248
The temporary workaround committed was to stop checking capabilities on parameters that are IDs.

dneto0:
Fixed by #1234

PR #1234, opened 2018-01-25
Add memory semantics checks to validate atomics

atgoo:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/252

Memory semantics checks are performed if the bit field was declared as int constant.

### DIFF
--- a/source/validate_atomics.cpp
+++ b/source/validate_atomics.cpp
@@ -23,6 +23,89 @@
 
 namespace libspirv {
 
+// Returns number of '1' bits in a word.
+uint32_t CountSetBits(uint32_t word) {
+  uint32_t count = 0;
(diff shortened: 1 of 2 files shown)


source/validate_atomics.cpp
```diff
@@ -23,6 +23,89 @@
 
 namespace libspirv {
 
+// Returns number of '1' bits in a word.
+uint32_t CountSetBits(uint32_t word) {
```
dneto0: FYI this is also known as "population count" or popcount which is sometimes a builtin in some compilers. 

But this is fine.  :-)

dneto0:
Rebased and pushed onto master as 44d88c8d9

Issue #1224, opened 2018-01-22
Handle phi updates in block merge pass


alan-baker:
The block merge pass bails on merging blocks if there are phis that reference the block that will be deleted. This should not inhibit the optimization, instead the phi nodes should be updated to indicate the updated incoming block.

A possible alternative for the block merging would be to delete the block with the unconditional branch, but that might not significantly limit the problem.

dneto0:
@alan-baker :  If this is fully fixed, then please close.

PR #1235, opened 2018-01-26
Merge blocks improvements

alan-baker:
Addresses #1224 

Handles more cases in merge blocks. Disallows only cases that are illegal:
* Non-single edge between blocks
* two headers
* merge and its header
* two merges

If another block is merged into a header, the merge instruction is relocated to just before the new terminator.

### DIFF
--- a/source/opt/block_merge_pass.cpp
+++ b/source/opt/block_merge_pass.cpp
@@ -106,8 +106,8 @@ bool BlockMergePass::IsHeader(uint32_t id) {
 }
 
 bool BlockMergePass::IsMerge(uint32_t id) {
-  return !get_def_use_mgr()->WhileEachUse(id, [this](ir::Instruction* user,
-                                                     uint32_t index) {
+  return !get_def_use_mgr()->WhileEachUse(id, [](ir::Instruction* user,
+                                                 uint32_t index) {
     SpvOp op = user->opcode();
     if ((op == SpvOpLoopMerge || op == SpvOpSelectionMerge) && index == 0u) {
       return false;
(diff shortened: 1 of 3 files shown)


source/opt/block_merge_pass.cpp
```diff
@@ -50,47 +38,88 @@ void BlockMergePass::KillInstAndName(ir::Instruction* inst) {
...
+
+    bool pred_is_header = IsHeader(&*bi);
+    bool succ_is_header = IsHeader(lab_id);
+    if (pred_is_header && succ_is_header) {
+      // Cannot merge two headers together.
+      ++bi;
+      continue;
+    }
+    if (pred_is_header && lab_id == bi->MergeBlockIdIfAny()) {
+      // Cannot merge the header into its own merge.
```
s-perron: Why can't we merge these?  If the only successor of the header is its own merge block, then we can merge then and remove the construct altogether.  What am I missing?

test/opt/block_merge_test.cpp
```diff
@@ -328,6 +271,255 @@ OpFunctionEnd
...
+
+  SinglePassRunAndMatch<opt::BlockMergePass>(text, true);
+}
+
+TEST_F(BlockMergeTest, MergeContinue) {
+  const std::string text = R"(
+; CHECK: OpBranch [[header:%\w+]]
+; CHECK: [[header]] = OpLabel
+; CHECK-NEXT: OpLogicalAnd
+; CHECK-NEXT: OpLoopMerge {{%\w+}} [[header]] None
```
dneto0: Sweet.  I had to think about this case while reading the code.  I'm glad you wrote a test for it.

dneto0:
Rebased, squashed, and merged as  e661da7941e

Issue #1093, opened 2017-12-14
Improve "requires capability" message to give the string name of the bad enum.


dneto0:
E.g. See https://github.com/KhronosGroup/glslang/issues/790#issuecomment-288865356

dneto0:
Fixed by #1237

PR #1237, opened 2018-01-26
Improved error message in val capabilities

atgoo:
The message was printing value of capability instead of a corresponding string.


dneto0:
Rebased and pushed into master as c86cb76a22

dneto0:
This is a fix for #1093

PR #1238, opened 2018-01-26
Add LoopUtils class to gather some loop transformation support

Naghasan:
This patch adds LoopUtils class to handle some loop related transformations. For now it has 2 transformations that will simplifies other transformations such as loop unroll or unswitch:
 - Dedicate exit blocks: this ensure that all exit basic block (out-of-loop basic blocks that have a predecessor in the loop) have all their predecessors in the loop;
 - Loop Closed SSA (LCSSA): this ensure that all definitions in a loop are used inside the loop or in a phi instruction in an exit basic block.

It also adds the following capabilities:
 - Loop::IsLCSSA to test if the loop is in a LCSSA form
 - Loop::GetOrCreatePreHeaderBlock that can build a loop preheader if required;
 - New methods to allow on the fly updates of the loop descriptors.
 - New methods to allow on the fly updates of the CFG analysis.
 - Instruction::SetOperand to allow expression of the index relative to Instruction::NumOperands (to be compatible with the index returned by DefUseManager::ForEachUse)


source/opt/loop_descriptor.cpp
```diff
@@ -103,6 +104,179 @@ bool Loop::IsBasicBlockInLoopSlow(const BasicBlock* bb) {
...
+      // All uses must be either:
+      //  - In the loop;
+      //  - In an exit block and in a phi instruction.
+      if (!def_use_mgr->WhileEachUser(
+              &insn, [&exit_blocks, context, this](ir::Instruction* use) {
+                BasicBlock* parent = context->get_instr_block(use);
+                assert(parent && "Invalid analysis");
+                if (IsInsideLoop(parent)) return true;
+                if (use->opcode() != SpvOpPhi) return false;
+                return !!exit_blocks.count(parent->id());
```
alan-baker: Remove the !!.

source/opt/loop_descriptor.h
```diff
@@ -68,14 +68,58 @@ class Loop {
...
+  }
 
   // Returns the latch basic block (basic block that holds the back-edge).
+  // These functions return nullptr if the loop is not structured (i.e. if it
+  // has more than one backedge).
   inline BasicBlock* GetLatchBlock() { return loop_continue_; }
   inline const BasicBlock* GetLatchBlock() const { return loop_continue_; }
-
-  // Returns the BasicBlock which marks the end of the loop.
+  // Sets |latch| as the loop unique continue block. A continue block must have
```
alan-baker: The latch and the continue blocks can be different blocks in structured control flow.

source/opt/loop_utils.cpp
```diff
@@ -0,0 +1,379 @@
...
+  ir::LoopDescriptor& loop_desc = *context_->GetLoopDescriptor(function);
+  ir::CFG& cfg = *context_->cfg();
+  opt::analysis::DefUseManager* def_use_mgr = context_->get_def_use_mgr();
+
+  constexpr ir::IRContext::Analysis PreservedAnalyses =
+      ir::IRContext::kAnalysisDefUse |
+      ir::IRContext::kAnalysisInstrToBlockMapping;
+
+  // Gathers the set of basic block that are not in this loop and have at least
+  // one predecessor in the loop and one not in the loop.
```
alan-baker: "and one not in the loop" does sound correct for the definition of an exit block.

source/opt/loop_utils.cpp
```diff
@@ -0,0 +1,379 @@
...
+  bool made_change = false;
+  // For each block, we create a new one that gathers all branches from
+  // the loop and fall into the block.
+  for (uint32_t non_dedicate_id : exit_bb_set) {
+    ir::BasicBlock* non_dedicate = cfg.block(non_dedicate_id);
+    const std::vector<uint32_t>& bb_pred = cfg.preds(non_dedicate_id);
+    // Ignore the block if:
+    //   - all the predecessors are in the loop;
+    //   - and has an unconditional branch;
+    //   - and any other instructions are phi.
```
alan-baker: You check that all other instructions are phis.

test/opt/loop_optimizations/lcssa.cpp
```diff
@@ -0,0 +1,351 @@
...
+  for (; i < 10; i++) {
+  }
+  if (i != 0) {
+    i = 1;
+  }
+}
+*/
+TEST_F(LCSSATest, SimpleLCSSA) {
+  const std::string text = R"(
+; CHECK: OpLoopMerge %32 %19 None
```
alan-baker: The test may be more robust if you rewrite the checks using regex variables:

```
; CHECK: OpLoopMerge [[merge:%\w+]]
; CHECK: [[merge]] = OpLabel
; CHECK-NEXT: [[phi:%\w+]] = OpPhi {{%\w+}} %30 %20
; CHECK-NEXT: OpBranch [[label:%\w+]]
; CHECK-NEXT: [[label]] = OpLabel
; CHECK-NEXT: OpINotEqual {{%\w+}} [[phi]] %9
```

source/opt/loop_utils.cpp
s-perron: Why does a dedicated exit have to contain only phi instructions?  This also contradicts the comment in the ".h" for this function that says structured control flow already has a dedicated exit.

At the same time, are you taking into consideration that merge block for the loop is not necessarily the exit block for the loop?  For example, 

```
for( int i = 0; i < 10; i++ ) {
// bb1
s += i;
}
// bb2
if (s > 10) //bb 3 s = 10;
// bb 4 selection merge
// bb 5 loop merge
...
```

In this case it is possible that the merge block for the OpLoopMerge in bb1 is bb5, not bb2.  I believe this is valid.  Does this effect the way things are being done?

source/opt/loop_utils.cpp
```diff
@@ -0,0 +1,379 @@
...
+  CreateLoopDedicateExits();
+
+  ir::Function* function = loop_->GetHeaderBlock()->GetParent();
+  ir::CFG& cfg = *context_->cfg();
+  opt::DominatorTree& dom_tree =
+      context_->GetDominatorAnalysis(function, cfg)->GetDomTree();
+
+  opt::analysis::DefUseManager* def_use_manager = context_->get_def_use_mgr();
+  std::unordered_set<ir::BasicBlock*> exit_bb;
+  for (uint32_t bb_id : loop_->GetBlocks()) {
```
s-perron: Can't you call `loop_->GetExitBlocks()` in place of the loop?

source/opt/loop_utils.cpp
```diff
@@ -0,0 +1,379 @@
...
+              }
+            }
+
+            for (ir::BasicBlock* e_bb : exit_bb) {
+              if (processed_exit.count(e_bb)) continue;
+              processed_exit.insert(e_bb);
+
+              // If the current exit basic block does not dominate |use| then
+              // |inst| does not escape through |e_bb|.
+              if (!dom_tree.Dominates(e_bb, use_parent)) continue;
```
s-perron: I'm not convinced this is true when the CFG is not structured.  Can you give more of an explanation?  I'm thinking a loop header defines a value, say `%2`.  It dominates its two exits nodes.  The two exit nodes then branch unconditionally to a node `bb4` that uses `%2`.  Then `bb4` is not dominated by either of the exit nodes, but `%2` still escapes through both of them.

What is suppose to happen in a case like this?

source/opt/basic_block.cpp
```diff
@@ -109,6 +109,28 @@ void BasicBlock::ForEachSuccessorLabel(
   }
 }
 
+void BasicBlock::ForEachSuccessorLabel(
+    const std::function<void(uint32_t*)>& f) {
+  auto br = &insts_.back();
+  switch (br->opcode()) {
+    case SpvOpBranch: {
+      uint32_t tmp_id = br->GetOperand(0).words[0];
```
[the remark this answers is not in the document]
Naghasan: @s-perron @alan-baker That would yield a `const uint32_t*` type (because GetOperand return a `const Operand&`  ) so that's incompatible with the type of `f`. Alternatively, a call to `ForEachInId` call can be made instead.

I added this member function to be able to redirect branches in an explicit manner. This is equivalent to `bb->tail()->ForEachInId` for most of the cases, the difference is some ids won't be labels.

source/opt/basic_block.h
```diff
@@ -133,6 +133,8 @@ class BasicBlock {
   void ForEachSuccessorLabel(
       const std::function<void(const uint32_t)>& f) const;
 
+  void ForEachSuccessorLabel(const std::function<void(uint32_t*)>& f);
```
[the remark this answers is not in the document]
Naghasan: Yes that was the intension, explicit in-place modification of the branch targets.

source/opt/loop_descriptor.cpp
```diff
@@ -103,6 +104,179 @@ bool Loop::IsBasicBlockInLoopSlow(const BasicBlock* bb) {
...
+      // Create a phi instruction if and only if the preheader_phi_ops has more
+      // than one pair.
+      if (preheader_phi_ops.size() > 2)
+        preheader_insn_def = builder.AddPhi(phi->type_id(), preheader_phi_ops);
+      else
+        preheader_insn_def =
+            context->get_def_use_mgr()->GetDef(preheader_phi_ops[0]);
+      // Build the new incoming branch.
+      header_phi_ops.push_back(preheader_insn_def->result_id());
+      header_phi_ops.push_back(loop_header_->id());
```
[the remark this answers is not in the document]
Naghasan: yes indeed... fixed and updated the tests

source/opt/loop_descriptor.cpp
Naghasan: Yes the lambda must return a bool, my old C reflex is to do this ... (that's a common idiom to do 0/1 normalization which also do a conversion to bool type in C++).

Another approach is to specify the type in the lambda declaration: `(ir::Instruction* use) -> bool`.

source/opt/loop_descriptor.h
Naghasan: Yes, I realized that now... I'll remove the update of the continue block.

source/opt/loop_utils.cpp
Naghasan: Hum very good point.

Specifically for your example: I don't think bb5 is a valid merge block (at least I find the choice strange, as bb2 post-dominate all exits), but I see your main point. I believe this code would generate a close structure (not quite the same though):
```
int i = 0;
while (i < 11) {
  s += i;
  if (i != 10) {
     if (s > 10) s = 10;
     break;
  }
  i++;
}
```
You have a similar pattern, but  would bb5 post-dominate all exits.

Generally, this function and the LCSSA transformation rely only on the capability to have access to the header block and the loop blocks. The update of the merge block will only happen if there is only 1 exit block (which is not correct if your example is valid).

source/opt/loop_utils.cpp
Naghasan: It is but the exit is not dominated by the loop so that can only happen in non-structured context:
```
if (cond) {
 for () {}
}
// merge-bb
```
It would remain valid to have the loop exit to be `merge-bb`, so it would have 2 incoming edges: from the if and from the loop.

source/opt/loop_utils.cpp
Naghasan: For the phi-instruction, I wasn't sure if we needed such a strong requirement. I just forgot to remove that part.

I added that code at some point because, I thought it would be easier to maintain the LCSSA if we have this property, but it is easy to do without.

source/opt/loop_utils.cpp
Naghasan: That should be using it indeed

source/opt/loop_descriptor.h
```diff
@@ -68,14 +68,58 @@ class Loop {
...
+  inline bool IsStructuredCompliant() const {
+    return GetLatchBlock() && GetMergeBlock();
+  }
+
+  // Returns true if the loop is structured according to SPIR-V rules:
+  //  - Has a merge block;
+  //  - Has a continue block;
+  //  - Has an OpLoopMerge instruction.
+  inline bool IsStructured() const {
+    return IsStructuredCompliant() && GetHeaderBlock()->GetLoopMergeInst();
```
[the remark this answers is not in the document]
Naghasan: No... I will kill that for now.

source/opt/loop_utils.cpp
Naghasan: That's a (stupid) mistake, this is true if use_parent does not post-dominates e_bb ...

I wanted to add a simple check that try to reduce the number of unnecessary phis, but by addressing the performance issue you raised for the rewriter, that can be done without that domination check.

Naghasan:
@alan-baker @s-perron I addressed all your comments but the one for ` BasicBlock::ForEachSuccessorLabel`. Notably, I added the optimization suggested when rewriting the use.

I also took more into account that a merge block can be different from the exit blocks. I think it would be beneficial to have the guaranty that escaping definitions should go through a phi for both exit blocks and merge blocks (at least at the beginning). I did that for those 2 reasons:
- Transformations that operate only on the repeated part of the loop (like some loop unroll), I don't think they should not worry about the merge construct (to avoid code duplication), so having the traditional LCSSA form would help here;
- Transformations that needs to duplicate the loop (like loop unswitch), they will duplicate every block of the loop, including the merge construct. So considering the merge block as the loop exit for LCSSA will help as well.

BTW, I saw the PR https://github.com/KhronosGroup/SPIRV-Tools/pull/1254, I'll update this PR once it is merge in.

source/opt/loop_utils.cpp
```diff
@@ -48,125 +48,249 @@ class LCSSARewriter {
...
-  }
+    // Return the new def to use for the basic block |bb_id|.
+    // If |bb_id| does not have a suitable def to use then we:
+    //   - return the common def used by all predecessors;
+    //   - if there is no common def, then we build a new phi instr at the
+    //     beginning of |bb_id| and return this new instruction.
+    ir::Instruction* GetOrBuildIncoming(uint32_t bb_id) {
+      assert(base_->cfg_->block(bb_id) != nullptr && "Unknown basic block");
+
+      ir::Instruction*& incoming_phi = bb_to_phi_[bb_id];
```
s-perron: I think there might be a problem with this uses of bb_to_phi_.

It seems like you will add any phi that uses the id when you "RegisterExitPhi".  It is possible that multiple phis uses the id.  How do you know the correct one, if any of the already existing ones are correct.

An example might be:

```
// bb1
m1=0;
for( int i = 0; i < 10; i++ ) {
   // bb2
   n2 = i;
   if (cond) {
      // bb3
      break;
   }
   // bb4
}
m2 = phi(n2, bb3, 0, bb4);
return n2+m2;
```
In this case, the phi defining `m2` might be added to bb_to_phi_, right?  However, when replacing the uses of `n2` in the return statement, you do not want to use the result of that phi.  You need a new one.  I think there might be an ordering issue with the calls to `RegisterExitPhi`, so it might sometimes get it right, but wrong other times.

source/opt/loop_utils.cpp
```diff
@@ -48,125 +48,249 @@ class LCSSARewriter {
...
+                  context](ir::Instruction* use, uint32_t operand_index) {
+            ir::BasicBlock* use_parent = context->get_instr_block(use);
+            assert(use_parent);
+            if (blocks.count(use_parent->id())) return;
+
+            if (use->opcode() == SpvOpPhi) {
+              // If the use is a Phi instruction and the incoming block is
+              // coming from the loop, then that's consistent with LCSSA form.
+              if (exit_bb.count(use_parent)) {
+                rewriter.RegisterExitPhi(use_parent, use);
```
s-perron: I don't see why you need or want this.  You probably don't want to replace a use with this phi, and it can overwrite the Phi created below.

Naghasan:
@s-perron I added the missing check to make sure a phi instruction this is actually eligible. I change the algorithm to make `UseRewriter` responsible to check/create the phi instruction on a need basis and so removed the `RegisterPhi`.

@dnovillo @alan-baker @s-perron I found an issue with the def/use manager when updating use map: for an instruction `a`, if you change one operand by an other, AnalyzeInstUse will not register that the instruction `a` no longer use this old definition. So `a` will keep showing up a user the old operand. I made a fix with the last commit, let me know if you are fine with the change

source/opt/def_use_manager.cpp
```diff
@@ -36,7 +36,8 @@ void DefUseManager::AnalyzeInstDef(ir::Instruction* inst) {
   }
 }
 
-void DefUseManager::AnalyzeInstUse(ir::Instruction* inst) {
+void DefUseManager::AnalyzeInstUse(ir::Instruction* inst, bool is_new_inst) {
+  if (!is_new_inst) EraseUseRecordsOfOperandIds(inst);
```
s-perron: Instead of adding a new parameter, you can check if `used_ids` (defined below) is empty or not.  Make this more transparent to the user.

Naghasan: Fixed

s-perron:
Rebase, squashed, and merged as 50e85c86.

source/opt/loop_descriptor.cpp
```diff
@@ -103,6 +103,197 @@ bool Loop::IsBasicBlockInLoopSlow(const BasicBlock* bb) {
...
+#ifndef NDEBUG
+  assert(merge->GetParent() && "The basic block does not belong to a function");
+  CFG& cfg = *merge->GetParent()->GetParent()->context()->cfg();
+
+  for (uint32_t pred : cfg.preds(merge->id())) {
+    assert(IsInsideLoop(pred) &&
+           "A predecessor of the merge block does not belong to the loop");
+  }
+#endif  // NDEBUG
+  assert(!IsInsideLoop(merge) && "The merge block is in the loop");
```
dnovillo: Might as well push this assert inside the #ifndef.

source/opt/loop_descriptor.cpp
```diff
@@ -103,6 +103,197 @@ bool Loop::IsBasicBlockInLoopSlow(const BasicBlock* bb) {
...
+  SetMergeBlockImpl(merge);
+  if (GetHeaderBlock()->GetLoopMergeInst()) {
+    UpdateLoopMergeInst();
+  }
+}
+
+void Loop::GetExitBlocks(IRContext* context,
+                         std::unordered_set<uint32_t>* exit_blocks) const {
+  ir::CFG* cfg = context->cfg();
+
```
dnovillo: Empty exit_blocks before the loop?  And add to the function's API doc.

source/opt/loop_descriptor.cpp
```diff
@@ -103,6 +103,197 @@ bool Loop::IsBasicBlockInLoopSlow(const BasicBlock* bb) {
...
+      if (!IsInsideLoop(succ)) {
+        exit_blocks->insert(succ);
+      }
+    });
+  }
+}
+
+void Loop::GetMergingBlocks(
+    IRContext* context, std::unordered_set<uint32_t>* merging_blocks) const {
+  assert(GetMergeBlock() && "This loop is not structured");
```
dnovillo: Likewise here for merging_blocks.

source/opt/loop_descriptor.cpp
```diff
@@ -103,6 +103,197 @@ bool Loop::IsBasicBlockInLoopSlow(const BasicBlock* bb) {
...
+    for (uint32_t pred_id : cfg->preds(bb->id())) {
+      if (!IsInsideLoop(pred_id) && !merging_blocks->count(pred_id)) {
+        to_visit.push(cfg->block(pred_id));
+      }
+    }
+  }
+}
+
+bool Loop::IsLCSSA() const {
+  IRContext* context = GetHeaderBlock()->GetParent()->GetParent()->context();
```
dnovillo: Isn't it easier to have a |context_| class member here?

source/opt/loop_utils.h
```diff
@@ -0,0 +1,75 @@
...
+
+namespace opt {
+
+// Set of basic loop transformation.
+class LoopUtils {
+ public:
+  LoopUtils(ir::IRContext* context, ir::Loop* loop)
+      : context_(context), loop_(loop) {}
+
+  // The make the current loop in the loop closed SSA form.
```
dnovillo: "This converts the current loop to loop closed SSA form"?

source/opt/loop_utils.h
```diff
@@ -0,0 +1,75 @@
...
+  // In the loop closed SSA, all loop exiting values goes through a dedicate SSA
```
dnovillo: s/goes/go/

dnovillo: And it's a dedicated Phi instruction, not SSA instruction, right?

PR #1239, opened 2018-01-26
Add adjacency validation pass.

jeremy-lunarg:
Validate OpPhi predecessors.
Validate OpLoopMerge successors.
Validate OpSelectionMerge successors.

### DIFF
--- a/source/validate_adjacency.cpp
+++ b/source/validate_adjacency.cpp
@@ -1,0 +1,84 @@
+// Copyright (c) 2018 LunarG Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Validates correctness of the intra-block preconditions of SPIR-V
+// instructions.
+
+#include "validate.h"
+
+#include <string>
+
+#include "diagnostic.h"
+#include "opcode.h"
+#include "val/instruction.h"
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+spv_result_t ValidateAdjacency(ValidationState_t& _) {
+  const auto& instructions = _.ordered_instructions();
+  for (auto i = instructions.cbegin(); i != instructions.cend(); ++i) {
+    switch (i->opcode()) {
+      case SpvOpPhi:
+        if (i != instructions.cbegin()) {
+          switch (prev(i)->opcode()) {
+            case SpvOpLabel:
+            case SpvOpPhi:
+            case SpvOpLine:
+              break;
+            default:
+              return _.diag(SPV_ERROR_INVALID_DATA)
+                     << "OpPhi must appear before all non-OpPhi instructions "
+                     << "(except for OpLine, which can be mixed with OpPhi).";
+          }
+        }
+        break;
+      case SpvOpLoopMerge:
+        if (next(i) != instructions.cend()) {
+          switch (next(i)->opcode()) {
+            case SpvOpBranch:
+            case SpvOpBranchConditional:
+              break;
+            default:
+              return _.diag(SPV_ERROR_INVALID_DATA)
+                     << "OpLoopMerge must immediately precede either an "
+                     << "OpBranch or OpBranchConditional instruction. "
+                     << "OpLoopMerge must be the second-to-last instruction in "
+                     << "its block.";
+          }
+        }
+        break;
+      case SpvOpSelectionMerge:
+        if (next(i) != instructions.cend()) {
+          switch (next(i)->opcode()) {
+            case SpvOpBranchConditional:
+            case SpvOpSwitch:
+              break;
+            default:
+              return _.diag(SPV_ERROR_INVALID_DATA)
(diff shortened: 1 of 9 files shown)


source/validate_adjacency.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include "diagnostic.h"
+#include "opcode.h"
+#include "val/instruction.h"
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+spv_result_t ValidateAdjacency(ValidationState_t& _) {
+  auto const instructions = _.ordered_instructions();
```
atgoo: ordered_instructions returns const std::deque<Instruction>&. Please use const auto& to avoid creating a new copy of the deque.

source/validate_adjacency.cpp
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2017 LunarG Inc.
```
atgoo: 2018

test/val/val_adjacency_test.cpp
```diff
@@ -0,0 +1,219 @@
+// Copyright (c) 2017 LunarG Inc.
```
atgoo: 2018

source/validate_adjacency.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+spv_result_t ValidateAdjacency(ValidationState_t& _) {
+  auto const instructions = _.ordered_instructions();
+  for (auto i = instructions.cbegin(); i != instructions.cend(); ++i) {
+    switch (i->opcode()) {
+      case SpvOpPhi:
+        switch (prev(i)->opcode()) {
+          case SpvOpLabel:
+          case SpvOpPhi:
+          case SpvOpLine:
```
dneto0: I would have expected OpNoLine in this list as well, but I see the spec's description of OpPhi only lists OpLine as an exception.

no request for a change.

source/validate_adjacency.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+spv_result_t ValidateAdjacency(ValidationState_t& _) {
+  auto const instructions = _.ordered_instructions();
+  for (auto i = instructions.cbegin(); i != instructions.cend(); ++i) {
+    switch (i->opcode()) {
+      case SpvOpPhi:
+        switch (prev(i)->opcode()) {
```
dneto0: It's worth saying that the instruction layout checks have ensured that prev(i) exists and is an instruction. (At least I hope that's true.)
I'd like to see a test case prove the validator is resilient to the case.

source/validate_adjacency.cpp
```diff
@@ -0,0 +1,78 @@
...
+          case SpvOpLine:
+            break;
+          default:
+            return _.diag(SPV_ERROR_INVALID_DATA)
+                   << "OpPhi must appear before all non-OpPhi instructions "
+                   << "(except for OpLine, which can be mixed with OpPhi).";
+        }
+        break;
+      case SpvOpLoopMerge:
+        switch (next(i)->opcode()) {
```
dneto0: Similar. I'd like to see test case to prove the validator works when LoopMerge is the last instruction in the module.

source/validate_adjacency.cpp
```diff
@@ -0,0 +1,78 @@
...
+            break;
+          default:
+            return _.diag(SPV_ERROR_INVALID_DATA)
+                   << "OpLoopMerge must immediately precede either an OpBranch "
+                   << "or OpBranchConditional instruction. OpLoopMerge must be "
+                   << "the second-to-last instruction in its block.";
+        }
+        break;
+      case SpvOpSelectionMerge:
+        switch (next(i)->opcode()) {
```
dneto0: I'd like a case where OpSelectionMerge is the last instruction in the module.

source/validate_adjacency.cpp
atgoo: At the very least we should be between OpFunction and OpFunctionEnd, so both prev() and next() should exist. It does make the code look flimsy though.

source/validate_adjacency.cpp
jeremy-lunarg: Done.

source/validate_adjacency.cpp
```diff
@@ -0,0 +1,78 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Validates correctness of the intra-block preconditions of SPIR-V instructions.
+
+// TODO: Rename this file to validate_adjacency.cpp.
```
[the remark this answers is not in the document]
jeremy-lunarg: Done.

source/validate_adjacency.cpp
jeremy-lunarg: Good catch. Fixed.

test/val/val_adjacency_test.cpp
jeremy-lunarg: Done.

source/validate_adjacency.cpp
jeremy-lunarg: Agreed. Added guards.

source/validate_adjacency.cpp
jeremy-lunarg: No change.

source/validate_adjacency.cpp
jeremy-lunarg: Done.

source/validate_adjacency.cpp
jeremy-lunarg: Done.

dneto0:
Rebased and pushed into master as cd68f2b17642

PR #1241, opened 2018-01-27
InsertExtractElim: Split out DeadInsertElim as separate pass

greg-lunarg:


s-perron:
rebase, squashed, and merged as f28b1061

PR #1242, opened 2018-01-29
Loop invariant code motion initial implementation

Alexander-Johnston:
This patch provides an initial implementation of a loop invariant code motion pass.
The pass expects eliminate-local-multi-store to have been run on the module first.

The pass works by performing the following algorithm on each loop in the module.

1. Add each invariant instruction in the loop to a queue of invariants
2. For each invariant instruction in the queue
2.1. Hoist it to the loops preheader
2.2. For each of the instructions users, check if they are now invariant, and if so add them to the invariants queue

And instruction is invariant iff

- It is an opcode with no side effects
- All instructions contained in its operands are outside the loop

### DIFF
--- a/source/opt/cfg.cpp
+++ b/source/opt/cfg.cpp
@@ -56,7 +56,9 @@ void CFG::RemoveNonExistingEdges(uint32_t blk_id) {
     const ir::BasicBlock* pred_blk = block(id);
     bool has_branch = false;
     pred_blk->ForEachSuccessorLabel([&has_branch, blk_id](uint32_t succ) {
-      if (succ == blk_id) has_branch = true;
+      if (succ == blk_id) {
+        has_branch = true;
+      }
     });
     if (has_branch) updated_pred_list.push_back(id);
   }
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -499,5 +499,80 @@ std::ostream& operator<<(std::ostream& str, const ir::Instruction& inst) {
   return str;
 }
 
+bool Instruction::IsOpcodeCodeMotionSafe() const {
+  switch (opcode_) {
+    case SpvOpVectorExtractDynamic:
+    case SpvOpVectorInsertDynamic:
+    case SpvOpVectorShuffle:
+    case SpvOpConvertFToU:
+    case SpvOpConvertFToS:
+    case SpvOpConvertSToF:
+    case SpvOpConvertUToF:
+    case SpvOpUConvert:
+    case SpvOpSConvert:
+    case SpvOpFConvert:
+    case SpvOpQuantizeToF16:
+    case SpvOpBitcast:
+    case SpvOpSNegate:
+    case SpvOpFNegate:
+    case SpvOpIAdd:
+    case SpvOpFAdd:
+    case SpvOpISub:
+    case SpvOpFSub:
+    case SpvOpIMul:
+    case SpvOpFMul:
+    case SpvOpUDiv:
+    case SpvOpSDiv:
+    case SpvOpFDiv:
+    case SpvOpUMod:
+    case SpvOpSRem:
+    case SpvOpSMod:
+    case SpvOpFRem:
+    case SpvOpFMod:
+    case SpvOpVectorTimesScalar:
+    case SpvOpMatrixTimesScalar:
+    case SpvOpVectorTimesMatrix:
+    case SpvOpMatrixTimesVector:
+    case SpvOpMatrixTimesMatrix:
+    case SpvOpLogicalEqual:
+    case SpvOpLogicalNotEqual:
+    case SpvOpLogicalOr:
+    case SpvOpLogicalAnd:
+    case SpvOpLogicalNot:
+    case SpvOpIEqual:
+    case SpvOpINotEqual:
+    case SpvOpUGreaterThan:
+    case SpvOpSGreaterThan:
+    case SpvOpUGreaterThanEqual:
+    case SpvOpSGreaterThanEqual:
+    case SpvOpULessThan:
+    case SpvOpSLessThan:
+    case SpvOpULessThanEqual:
+    case SpvOpSLessThanEqual:
+    case SpvOpFOrdEqual:
+    case SpvOpFUnordEqual:
+    case SpvOpFOrdNotEqual:
+    case SpvOpFUnordNotEqual:
+    case SpvOpFOrdLessThan:
+    case SpvOpFUnordLessThan:
+    case SpvOpFOrdGreaterThan:
+    case SpvOpFUnordGreaterThan:
+    case SpvOpFOrdLessThanEqual:
+    case SpvOpFUnordLessThanEqual:
+    case SpvOpFOrdGreaterThanEqual:
+    case SpvOpFUnordGreaterThanEqual:
+    case SpvOpShiftRightLogical:
+    case SpvOpShiftRightArithmetic:
+    case SpvOpShiftLeftLogical:
+    case SpvOpBitwiseOr:
+    case SpvOpBitwiseXor:
+    case SpvOpBitwiseAnd:
+    case SpvOpNot:
+      return true;
+    default:
+      return false;
+  }
+}
+
 }  // namespace ir
 }  // namespace spvtools
--- a/source/opt/licm_pass.cpp
+++ b/source/opt/licm_pass.cpp
@@ -1,0 +1,125 @@
+// Copyright (c) 2018 Google LLC.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "opt/licm_pass.h"
+#include "opt/module.h"
+#include "opt/pass.h"
+
+#include <queue>
+#include <utility>
+
+namespace spvtools {
+namespace opt {
+
+Pass::Status LICMPass::Process(ir::IRContext* c) {
+  InitializeProcessing(c);
+  bool modified = false;
+
+  if (c != nullptr) {
+    modified = ProcessIRContext();
+  }
+
+  return modified ? Status::SuccessWithChange : Status::SuccessWithoutChange;
+}
+
+bool LICMPass::ProcessIRContext() {
+  bool modified = false;
+  ir::Module* module = get_module();
+
+  // Process each function in the module
+  for (ir::Function& f : *module) {
+    modified |= ProcessFunction(&f);
+  }
+  return modified;
+}
+
+bool LICMPass::ProcessFunction(ir::Function* f) {
+  bool modified = false;
+  ir::LoopDescriptor* loop_descriptor = context()->GetLoopDescriptor(f);
+
+  // Process each loop in the function
+  for (ir::Loop& loop : *loop_descriptor) {
+    // Ignore nested loops, as we will process them in order in ProcessLoop
+    if (loop.IsNested()) {
+      continue;
+    }
+    modified |= ProcessLoop(&loop, f);
+  }
+  return modified;
+}
+
+bool LICMPass::ProcessLoop(ir::Loop* loop, ir::Function* f) {
+  bool modified = false;
+
+  // Process all nested loops first
+  for (ir::Loop* nested_loop : *loop) {
+    modified |= ProcessLoop(nested_loop, f);
+  }
+
+  std::vector<ir::BasicBlock*> loop_bbs{};
+  modified |= AnalyseAndHoistFromBB(loop, f, loop->GetHeaderBlock(), &loop_bbs);
+
+  for (size_t i = 0; i < loop_bbs.size(); ++i) {
+    ir::BasicBlock* bb = loop_bbs[i];
+    // do not delete the element
+    modified |= AnalyseAndHoistFromBB(loop, f, bb, &loop_bbs);
+  }
+
+  return modified;
+}
+
+bool LICMPass::AnalyseAndHoistFromBB(ir::Loop* loop, ir::Function* f,
+                                     ir::BasicBlock* bb,
+                                     std::vector<ir::BasicBlock*>* loop_bbs) {
+  bool modified = false;
+  std::function<void(ir::Instruction*)> hoist_inst =
+      [this, &loop, &modified](ir::Instruction* inst) {
+        if (loop->ShouldHoistInstruction(this->context(), inst)) {
+          HoistInstruction(loop, inst);
+          modified = true;
+        }
+      };
+
+  if (IsImmediatelyContainedInLoop(loop, f, bb)) {
+    bb->ForEachInst(hoist_inst, false);
+  }
+
+  opt::DominatorAnalysis* dom_analysis =
+      context()->GetDominatorAnalysis(f, *cfg());
+  opt::DominatorTree& dom_tree = dom_analysis->GetDomTree();
+
+  for (opt::DominatorTreeNode* child_dom_tree_node :
+       *dom_tree.GetTreeNode(bb)) {
+    if (loop->IsInsideLoop(child_dom_tree_node->bb_)) {
+      loop_bbs->push_back(child_dom_tree_node->bb_);
+    }
+  }
+
+  return modified;
+}
+
+bool LICMPass::IsImmediatelyContainedInLoop(ir::Loop* loop, ir::Function* f,
+                                            ir::BasicBlock* bb) {
+  ir::LoopDescriptor* loop_descriptor = context()->GetLoopDescriptor(f);
+  return loop == (*loop_descriptor)[bb->id()];
+}
+
+void LICMPass::HoistInstruction(ir::Loop* loop, ir::Instruction* inst) {
+  ir::BasicBlock* pre_header_bb = loop->GetOrCreatePreHeaderBlock();
+  inst->InsertBefore(std::move(&(*pre_header_bb->tail())));
+  context()->set_instr_block(inst, pre_header_bb);
+}
+
+}  // namespace opt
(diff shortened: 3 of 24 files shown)


source/opt/instruction.cpp
```diff
@@ -499,5 +499,51 @@ std::ostream& operator<<(std::ostream& str, const ir::Instruction& inst) {
   return str;
 }
 
+bool Instruction::HasSideEffects() const {
```
alan-baker: There is a IsCombinatorInstruction() function which is effectively the reverse of this list. Our preference is to whitelist operations instead of blacklisting them so that future extensions are not buggy until fixed.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+  // Process each function in the module
+  for (ir::Function& f : *module) {
+    modified |= ProcessFunction(&f);
+  }
+  return modified;
+}
+
+bool LICMPass::ProcessFunction(ir::Function* f) {
+  bool modified = false;
+  ir::LoopDescriptor loop_descriptor{f};
```
alan-baker: This is now available as an analysis.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+bool LICMPass::ProcessInstructionList(
+    ir::Loop* loop, std::queue<ir::Instruction*>* loop_iv_instr) {
+  if (loop_iv_instr->empty()) {
+    return false;
+  }
+
+  ir::BasicBlock* pre_header_bb = loop->GetPreHeaderBlock();
+
+  // If the input instruction is invariant, push it onto the invariants queue
+  const std::function<void(ir::Instruction*)> check_users_now_invariant =
```
alan-baker: Are you assuming the loop is in SSA? This appears to be missing a test that the user is in the loop.

source/opt/licm_pass.h
```diff
@@ -0,0 +1,78 @@
...
+
+class LICMPass : public Pass {
+ public:
+  LICMPass();
+
+  const char* name() const override { return "loop-invariant-code-motion"; }
+  Status Process(ir::IRContext*) override;
+
+ private:
+  // Searches the IRContext for functions and processes each, moving invairants
```
alan-baker: invairants -> invariants

source/opt/licm_pass.cpp
Alexander-Johnston: Yes, I am assuming the loop is in SSA.
You're right about the missing test for the user, thank you.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <queue>
+#include <utility>
+
+#include "opt/licm_pass.h"
```
s-perron: This include goes first: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

source/opt/licm_pass.h
```diff
@@ -0,0 +1,78 @@
...
+  // of its operands are outside the loop. If so, moves the instruction to a
+  // queue. Once the list has been iterated over, hoists each inst in the queue
+  // outside of the loop.
+  // This cycle continues until a full iteration over the instructions list
+  // results in no instructions being queued.
+  bool ProcessInstructionList(ir::Loop* loop,
+                              std::queue<ir::Instruction*>* loop_iv_instr);
+
+ private:
+  ir::IRContext* ir_context;
```
s-perron: A pass already has a copy of the context.  You just need to set things up right.  We should change this.  But call `Initialize Processing(context);` at the start of the transformation.  Then calling `context()` will return the context, there are shortcuts to get access to other things as well.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+  // Process each loop in the function
+  for (ir::Loop& loop : loop_descriptor) {
+    modified |= ProcessLoop(&loop, f);
+  }
+  return modified;
+}
+
+bool LICMPass::ProcessLoop(ir::Loop* loop, ir::Function* f) {
+  // Process all nested loops first
```
s-perron: Doesn't the loop that calls this function already do a post-order depth-first traversal?  For a deep loop nest, you will be processing the same loop many times.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+  GatherAllImmediatelyInvariantInstructions(loop, &loop_iv_instr);
+
+  return ProcessInstructionList(loop, &loop_iv_instr);
+}
+
+void LICMPass::GatherAllImmediatelyInvariantInstructions(
+    ir::Loop* loop, std::queue<ir::Instruction*>* loop_iv_instr) {
+  // For each instruction in each basic block in the loop, check if it is
+  // invariant. If so, push it to our invariants queue.
+  for (uint32_t bb_id : loop->GetBlocks()) {
```
s-perron: If you already processed the inner loops, there is not need to go over the blocks in the nested loop again.  You can cut down on the amount of work you need to do here.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+          loop_iv_instr->push(inst);
+        }
+      };
+
+  // While there are invariant instructions in the queue, hoist them outside of
+  // the loop, then add their users to the queue if they have become invariant
+  while (!loop_iv_instr->empty()) {
+    ir::Instruction* inst = loop_iv_instr->front();
+    loop_iv_instr->pop();
+    HoistInstruction(pre_header_bb, inst);
```
s-perron: As I think about this, we might be able to simplify this code, and remove the dependence on the def-use chains.  In `GatherAllImmediatelyInvariantInstructions`, if you traverse the blocks in dominance order starting with the header, and you hoist the instruction right away instead of waiting, then that first pass should be all that you need.  No need for the work list.

The idea is that if an instruction is loop invariant, then its input are outside the loop, or they are inside the loop, must dominate the use, and are invariant themselves.  So the feeders will have been moved first.

Personally, I don't like using the def-use chain (use-def chains are fine) because they are expensive to keep up to date.  Eventually we need to split the def-use and use-def chains into two separate analyses, and use the def-use chain rarely.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+bool LICMPass::AllOperandsOutsideLoop(ir::Loop* loop, ir::Instruction* inst) {
+  analysis::DefUseManager* def_use_mgr = ir_context->get_def_use_mgr();
+  bool all_outside_loop = true;
+
+  const std::function<void(uint32_t*)> operand_test_lambda =
+      [&def_use_mgr, &loop, &all_outside_loop](uint32_t* id) {
+        if (loop->IsInsideLoop(def_use_mgr->GetDef(*id))) {
+          all_outside_loop = false;
```
dnovillo: Return early here.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  std::queue<ir::Instruction*> loop_iv_instr{};
+  GatherAllImmediatelyInvariantInstructions(loop, &loop_iv_instr);
+
+  return ProcessInstructionList(loop, &loop_iv_instr);
+}
+
+void LICMPass::GatherAllImmediatelyInvariantInstructions(
+    ir::Loop* loop, std::queue<ir::Instruction*>* loop_iv_instr) {
```
dnovillo: Not totally convinced about this suggestion, but I'd like your take on it.  How about making this and AllOperandsOutsideLoop members of the Loop class?  They are a loop-related analysis and there's some likelihood that we may use them in other loop transformations.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  ir::BasicBlock* pre_header_bb = loop->GetPreHeaderBlock();
+
+  // If the input instruction is invariant, push it onto the invariants queue
+  const std::function<void(ir::Instruction*)> check_users_now_invariant =
+      [this, &loop, &loop_iv_instr](ir::Instruction* inst) {
+        if (inst->opcode() != SpvOpPhi && !inst->HasSideEffects() &&
+            AllOperandsOutsideLoop(loop, inst)) {
+          loop_iv_instr->push(inst);
```
dnovillo: What about instructions that cannot be moved?

source/opt/licm_pass.h
```diff
@@ -0,0 +1,78 @@
...
+  // Checks the function for loops, calling ProcessLoop on each one found.
+  // Returns true if a change was made to the function, false otherwise.
+  bool ProcessFunction(ir::Function* f);
+
+  // Checks for invariants in the loop and attempts to move them to the loops
+  // preheader. Works from inner loop to outer when nested loops are found.
+  // Returns true if a change was made to the loop, false otherwise.
+  bool ProcessLoop(ir::Loop* loop, ir::Function* f);
+
+  // Gathers all instructions in the loop whos opcodes do not have side effects
```
dnovillo: s/whos/whose/

source/opt/licm_pass.cpp
dnovillo: In fact, a barrier instruction for instance would prevent moving instructions downstream from it, even if they're LI.

source/opt/licm_pass.cpp
Alexander-Johnston: I quite like this suggestion. Do you think it's worth caching the invariants once calculated, or just returning the invariants as it does now? I feel like anything that needs to gather the invariants is likely to invalidate a cached list of them quite quickly.
Additionally, is it safe to use IsCombinatorInstruction to help determine the invariance, or are there combinators that may be invariant?

Alexander-Johnston: Instead though, as Steven suggested above we don't need `GatherAllImmediatelyInvariantInstructions` to produce an initial working list. We could just move `AreAllOperandsOutsideLoop` to the Loop class and simplify this pass in the process.

source/opt/licm_pass.cpp
Alexander-Johnston: Should the pass ignore any loops containing barriers then?

s-perron: Why would a barrier stop us from moving things?  A barrier is more about memory.  If the instruction is not a load or store of some type, then we should be able to move it.  On that note, I did not really pay attention to what happens to loads.  They are considered combinators.  Even if all of its operands are defined outside of the loop, the same variable is stored to then we have to leave it.

source/opt/instruction.cpp
s-perron: You might want to look at the value numbering code to see the conditions used there.  There will need to be more than just looking for combinators.

source/opt/licm_pass.cpp
dnovillo: Yes, I meant instructions accessing shared resources.  So loads/stores and anything else that may have hidden shared inputs/outputs.

source/opt/loop_descriptor.cpp
```diff
@@ -367,5 +367,26 @@ void LoopDescriptor::PopulateList(const Function* f) {
   }
 }
 
+bool Loop::IsInstructionInvariant(IRContext* context, Instruction* inst) {
```
alan-baker: Consider renaming this ShouldHoistInstruction().

alan-baker:
As for scheduling, we may need to do more testing, but generally if you fully unroll a loop you can expose many more opportunities for SROA so we may want to re-run the mem2reg set of passes (SROA and 3 mem2regs) afterwards. I agree this needs to come after the initial round mem2reg so the induction variables are promoted.

Probably we should file another issue or two. Determine an ordering and (separately, determine whether all three mem2regs are really necessary).

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,149 @@
...
+  bool modified = false;
+  if (!IsInsideNestedLoop(loop, f, bb)) {
+    bool hoisted = true;
+    while (hoisted) {
+      hoisted = false;
+      // When we perform a hoist it will invalidate this iterator, but we
+      // require the hoist so instructions later in the bb which depended on the
+      // hoisted instruction can now be hoisted. As such, we break iteration
+      // over the bb and start again with fresh iterators until no instructions
+      // are hoisted
```
s-perron: You can get around this by not using iterators.  The `ForEachInst` and `WhileEachInst` are written to handle deleting or moving the current instruction.  We should add that to the documentation for them because we rely on it in a few places.

Can be done in a follow up PR.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  ir::CFG cfg(get_module());
+  opt::DominatorAnalysis* dom_analysis =
+      context()->GetDominatorAnalysis(f, cfg);
+  opt::DominatorTree& dom_tree = dom_analysis->GetDomTree();
+
+  for (opt::DominatorTreeNode* child_dom_tree_node :
+       *dom_tree.GetTreeNode(bb)) {
+    if (dom_analysis->Dominates(header_bb, child_dom_tree_node->bb_) &&
```
s-perron: It is possible to replace this will a call to IsInLoop or something like that?  It would be easier to read.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,149 @@
...
+        !dom_analysis->Dominates(merge_bb, child_dom_tree_node->bb_)) {
+      loop_bbs->push(child_dom_tree_node->bb_);
+    }
+  }
+
+  return modified;
+}
+
+bool LICMPass::IsInsideNestedLoop(ir::Loop* loop, ir::Function* f,
```
s-perron: The name implies something different that what you are checking.  What if `bb` is not in `loop` at all.  Then it is not in a loop that is nested in `loop`, but this function will return true.

A better name might be `IsImmediatelyContainInLoop` (seems a little long), or anything you can think of.

source/opt/licm_pass.h
```diff
@@ -0,0 +1,69 @@
...
+
+  // Analyses each instruction in |bb|, hoisting invariants to |pre_header_bb|.
+  // Each child of |bb| wrt to |dom_tree| is pushed to |loop_bbs|
+  bool AnalyseAndHoistFromBB(ir::Loop* loop, ir::Function* f,
+                             ir::BasicBlock* header_bb,
+                             ir::BasicBlock* merge_bb,
+                             ir::BasicBlock* pre_header_bb, ir::BasicBlock* bb,
+                             std::queue<ir::BasicBlock*>* loop_bbs);
+
+  // Returns true if |bb| belongs to a block nested in |loop|
```
s-perron: I think you meant "if |bb| belongs to a loop nested in |loop|".  As I mentioned in another comment, this not what you are actually checking.  A name change and comment change might be in order.

source/opt/licm_pass.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+  // Get the pre_header. If it doesn't exist, force it's creation
+  ir::BasicBlock* pre_header_bb = loop->GetPreHeaderBlock();
+  if (!pre_header_bb) {
+    modified = true;
+    pre_header_bb = loop->GetOrCreatePreHeaderBlock();
+  }
+  ir::BasicBlock* header_bb = loop->GetHeaderBlock();
+  ir::BasicBlock* merge_bb = loop->GetMergeBlock();
+  std::queue<ir::BasicBlock*> loop_bbs{};
```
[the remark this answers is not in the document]
s-perron: Yes, we want to avoid std::deque and std::list, and std::queue is built on std::deque by default.  Please try to rework this.  Since you don't have to add too many basic block, you could do something like:

```
std::vector<BasicBlock*> loop_bbs;
...
for( int i = 0; i < loop_bbs.size(); ++i ) {
  ir::Basicblock* bb = loop_bbs[i];
  // do not delete the element.
  modified |= AnalyseAndHoistFromBB(loop, f, header_bb,
                                      merge_bb, pre_header_bb, bb, &loop_bbs);
}
```

source/opt/cfg.cpp
```diff
@@ -58,7 +58,6 @@ void CFG::RemoveNonExistingEdges(uint32_t blk_id) {
     pred_blk->ForEachSuccessorLabel([&has_branch, blk_id](uint32_t succ) {
       if (succ == blk_id) {
         has_branch = true;
-        return;
```
dnovillo: I remember requesting this.  @s-perron, I believe you said it wasn't needed.  Why, though?  Once we determine that there is a branch, we could just return from the lambda.  No point iterating over all the other successors.  Granted it's not a big saving, but I'm not sure why it's unsafe to just return once found.

s-perron: The return leaves that instance of the lambda.  It does not stop it from checking the other labels.

s-perron: If we had a "WhileEach*" version, we could return false to indicate that we should stop.

dnovillo:
On Wed, Feb 7, 2018, 13:30 Steven Perron, <notifications@github.com> wrote:

>
> In source/opt/cfg.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1242#discussion_r166711344>
> :
>
> > @@ -58,7 +58,6 @@ void CFG::RemoveNonExistingEdges(uint32_t blk_id) {
>      pred_blk->ForEachSuccessorLabel([&has_branch, blk_id](uint32_t succ) {
>        if (succ == blk_id) {
>          has_branch = true;
> -        return;
>
> If we had a "WhileEach*" version, we could return false to indicate that
> we should stop.
>

D'oh, of course. Thanks. I kept forgetting that the iteration is not
affected by the lambda. Apologies for the noise.

s-perron:
I think this is good to merge once all of the bots have passed.  Please rebase, and then I can merge it.

dnovillo:
LGTM

dnovillo:
Please fix the test failures.

s-perron:
I'll try to merge these right now.

s-perron:
Could you do a **rebase** and squash?  Not a merge.  We like keeping a linear history to make problem identification easy.  I tried rebasing, but it is not obvious how to do it.

s-perron:
Rebased and merged as 84ccd0b9.

Issue #1250, opened 2018-01-30
Dead branch elimination makes unnecessary changes


alan-baker:
While performing some phase ordering experiments, I came across some instances of dead branch elimination reporting SuccessWithChange after repeated runs of the pass. At least some of it stems from unreachable continue blocks. The pass should be improved to only report change on real changes (i.e. it should not make effective nop transforms).

dneto0:
Fixed with #1251

PR #1251, opened 2018-01-30
Prevent dead branch from making unnecessary changes

alan-baker:
Addresses #1250

* When handling unreachable merges and continues, do not optimize to the
same IR
 * pass did not check whether the unreachable blocks were in the
 optimized form before transforming them
* added a test to catch this issue


dneto0:
Rebased and pushed onto master as 16949236fe5

PR #1253, opened 2018-01-30
Add pass to replace invalid opcodes

### LINKED ISSUE #1247: HLSL legalization: invalid opcode for shader stage.
Another problem that has come up in the HLSL legalization area is the use of specific opcodes that are not legal for the given shader stage.  For example, an `OpImageSampleImplicitLod` instruction can only be used in  fragment shader.  The generic HLSL code that is written will contain instructions of these types in non-fragment shaders, but, on close examination, there result will not be used.

If the optimizer is able to do many optimizations, then these instructions can be removed.  The problem is that the users many not always want to optimize that much, or the optimizer may not be good enough to prove that the invalid instruction's results are not used.

The help with this problem, I would like to add a pass that will assume these invalid instructions are not used, and replace them with a constant value.

1. The instructions that will be replaced are those valid for shader, but not for the current shader stage.
1. They will be replaced by a specific constant.
1. A warning will be added to the message consumer for each instruction that is replaced, with file and line number when available.
1. If a module has multiple entry points for different shader stages, then the pass will do nothing.
    1. This limitation could be removed at a later time if we need to.

The advantages of this approach are:

1. Correct code will not see any behavior change, and will not have to wait for improvements to the optimizer in order to be able to run.
1. This allows the possibility of a non-optimized compile that makes debugging easier.

The disadvantages of this approach are:

1. For incorrect code, the compiler would have issued an error before.  Not the user will have to identify a possibly subtle error in the output of the shader.
    1. The ideal solution would be to have the shader "trap"; however, that type of instruction does not seem to be available.
    1.  These compiler errors will probably be false positives for the HLSL, so the error is not very useful anyway.
    1. This replacement will not be turned on by default, so users will be aware when this might be happening.
    1. The warnings can give an idea where the replacement is taking place allowing the user to at least look at those locations to see if they are a problem.

s-perron:
Creates a pass that will remove instructions that are invalid for the
current shader stage.  For the instruction to be considered for replacement

1) The opcode must be valid for a shader modules.
2) The opcode must be invalid for the current shader stage.
3) All entry points to the module must be for the same shader stage.
4) The function containing the instruction must be reachable from an entry point.

Fixes #1247.

### DIFF
--- a/source/opt/replace_invalid_opc.cpp
+++ b/source/opt/replace_invalid_opc.cpp
@@ -1,0 +1,207 @@
+// Copyright (c) 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "replace_invalid_opc.h"
+
+#include <bitset>
+
+namespace spvtools {
+namespace opt {
+
+Pass::Status ReplaceInvalidOpcodePass::Process(ir::IRContext* c) {
+  InitializeProcessing(c);
+  bool modified = false;
+
+  if (context()->get_feature_mgr()->HasCapability(SpvCapabilityLinkage)) {
+    return Status::SuccessWithoutChange;
+  }
+
+  SpvExecutionModel execution_model = GetExecutionModel();
+  if (execution_model == SpvExecutionModelKernel) {
+    // We do not handle kernels.
+    return Status::SuccessWithoutChange;
+  }
+  if (execution_model == SpvExecutionModelMax) {
+    // Mixed execution models for the entry points.  This case is not currently
+    // handled.
+    return Status::SuccessWithoutChange;
+  }
+
+  for (ir::Function& func : *get_module()) {
+    modified |= RewriteFunction(&func, execution_model);
+  }
+  return (modified ? Status::SuccessWithChange : Status::SuccessWithoutChange);
+}
+
+SpvExecutionModel ReplaceInvalidOpcodePass::GetExecutionModel() {
+  SpvExecutionModel result = SpvExecutionModelMax;
+  bool first = true;
+  for (ir::Instruction& entry_point : get_module()->entry_points()) {
+    if (first) {
+      result =
+          static_cast<SpvExecutionModel>(entry_point.GetSingleWordInOperand(0));
+      first = false;
+    } else {
+      SpvExecutionModel current_model =
+          static_cast<SpvExecutionModel>(entry_point.GetSingleWordInOperand(0));
+      if (current_model != result) {
+        result = SpvExecutionModelMax;
+        break;
+      }
+    }
+  }
+  return result;
+}
+
+bool ReplaceInvalidOpcodePass::RewriteFunction(ir::Function* function,
+                                               SpvExecutionModel model) {
+  bool modified = false;
+  ir::Instruction* last_line_dbg_inst = nullptr;
+  function->ForEachInst(
+      [model, &modified, &last_line_dbg_inst, this](ir::Instruction* inst) {
+        // Track the debug information so we can have a meaningful message.
+        if (inst->opcode() == SpvOpLabel || inst->opcode() == SpvOpNoLine) {
+          last_line_dbg_inst = nullptr;
+          return;
+        } else if (inst->opcode() == SpvOpLine) {
+          last_line_dbg_inst = inst;
+          return;
+        }
+
+        bool replace = false;
+        if (model != SpvExecutionModelFragment &&
+            IsFragmentShaderOnlyInstruction(inst)) {
+          replace = true;
+        }
+
+        if (model != SpvExecutionModelTessellationControl &&
+            model != SpvExecutionModelGLCompute) {
+          if (inst->opcode() == SpvOpControlBarrier) {
+            assert(model != SpvExecutionModelKernel &&
+                   "Expecting to be working on a shader module.");
+            replace = true;
+          }
+        }
+
+        if (replace) {
+          modified = true;
+          if (last_line_dbg_inst == nullptr) {
+            ReplaceInstruction(inst, nullptr, 0, 0);
+          } else {
+            // Get the name of the source file.
+            ir::Instruction* file_name = context()->get_def_use_mgr()->GetDef(
+                last_line_dbg_inst->GetSingleWordInOperand(0));
+            const char* source = reinterpret_cast<const char*>(
+                &file_name->GetInOperand(0).words[0]);
+
+            // Get the line number and column number.
+            uint32_t line_number =
+                last_line_dbg_inst->GetSingleWordInOperand(1);
+            uint32_t col_number = last_line_dbg_inst->GetSingleWordInOperand(2);
+
+            // Replace the instruction.
+            ReplaceInstruction(inst, source, line_number, col_number);
+          }
+        }
+      },
+      /* run_on_debug_line_insts = */ true);
+  return modified;
+}
+
+bool ReplaceInvalidOpcodePass::IsFragmentShaderOnlyInstruction(
+    ir::Instruction* inst) {
+  switch (inst->opcode()) {
+    case SpvOpDPdx:
+    case SpvOpDPdy:
+    case SpvOpFwidth:
+    case SpvOpDPdxFine:
+    case SpvOpDPdyFine:
+    case SpvOpFwidthFine:
+    case SpvOpDPdxCoarse:
+    case SpvOpDPdyCoarse:
+    case SpvOpFwidthCoarse:
+    case SpvOpImageSampleImplicitLod:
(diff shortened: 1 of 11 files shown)


s-perron:
This pull request does not remove OpKill or OpTypeImage with DIM=SubpassData.  They could be handled if needed, but we can do that as another PR.

source/opt/replace_invalid_opc.cpp
```diff
@@ -0,0 +1,149 @@
...
+  function->ForEachInst([model, &modified, this](ir::Instruction* inst) {
+    if (model != SpvExecutionModelFragment &&
+        IsFragmentShaderOnlyInstruction(inst)) {
+      modified = true;
+      ReplaceInstruction(inst);
+    }
+
+    if (model != SpvExecutionModelTessellationControl &&
+        model != SpvExecutionModelGLCompute) {
+      if (inst->opcode() == SpvOpControlBarrier) {
```
alan-baker: I didn't see any tests covering this case. Please add one.

s-perron:
I'll add that test.  I also need to update the test that remove the instructions to check that a warning is emitted.  I also need to emit the warning.

source/opt/replace_invalid_opc.cpp
```diff
@@ -0,0 +1,149 @@
...
+  InitializeProcessing(c);
+  bool modified = false;
+
+  if (context()->get_feature_mgr()->HasCapability(SpvCapabilityLinkage)) {
+    return Status::SuccessWithoutChange;
+  }
+
+  SpvExecutionModel execution_model = GetExecutionModel();
+  if (execution_model == SpvExecutionModelKernel ||
+      execution_model == SpvExecutionModelMax) {
```
atgoo: Is there a case when it should be equal SpvExecutionModelMax?

source/opt/replace_invalid_opc.cpp
```diff
@@ -0,0 +1,149 @@
...
+        result = SpvExecutionModelMax;
+        break;
+      }
+    }
+  }
+  return result;
+}
+
+bool ReplaceInvalidOpcodePass::RewriteFunction(ir::Function* function,
+                                               SpvExecutionModel model) {
```
atgoo: Consider
```
RewriteFunction(ir::Function* function, std::set<SpvExecutionModel> models)
```
You can traverse the call trees of each entry point and collect execution models for every function.

source/opt/replace_invalid_opc.cpp
```diff
@@ -0,0 +1,149 @@
...
+    case SpvOpDPdyCoarse:
+    case SpvOpFwidthCoarse:
+    case SpvOpImageSampleImplicitLod:
+    case SpvOpImageSampleDrefImplicitLod:
+    case SpvOpImageSampleProjImplicitLod:
+    case SpvOpImageSampleProjDrefImplicitLod:
+    case SpvOpImageSparseSampleImplicitLod:
+    case SpvOpImageSparseSampleDrefImplicitLod:
+    case SpvOpImageQueryLod:
+      // case SpvOpKill:
```
atgoo: Please add a comment explaining SpvOpKill.

source/opt/replace_invalid_opc.cpp
```diff
@@ -0,0 +1,149 @@
...
+    if (model != SpvExecutionModelFragment &&
+        IsFragmentShaderOnlyInstruction(inst)) {
+      modified = true;
+      ReplaceInstruction(inst);
+    }
+
+    if (model != SpvExecutionModelTessellationControl &&
+        model != SpvExecutionModelGLCompute) {
+      if (inst->opcode() == SpvOpControlBarrier) {
+        modified = true;
```
atgoo: Consider assert(model != SpvExecutionModelKernel);

source/opt/replace_invalid_opc.cpp
s-perron: This response applies to a few of the comments below.  The problem that can come up is if there are two entry points with different models, and there call trees are not disjoint.  A function in both call trees could be in a conflicting situation.  You need to keep the instruction because of one the first entry point, but it also needs to be removed for the second one.

This can lead to a mildly complicated analysis.  It can be done, but it complicates things.  Considering that our current set of shaders generates a single entry point per shader module, we did not consider it worth implementing.  So will just do a simple check (all entry points must have the same model).

For this check, we use SpvExecutionModelMax as a special value to indicate that the entry points have inconsistent model.  No shader can have the model SpvExecutionModelMax.  However, I will admit this is not future proof.

source/opt/replace_invalid_opc.cpp
s-perron: As I mentioned above, the actual use case are much simpler, so I wanted to keep the code simple.  If the need for more general solution, we can do it at that time.

source/opt/replace_invalid_opc.cpp
s-perron: done

source/opt/replace_invalid_opc.cpp
s-perron: done

source/opt/replace_invalid_opc.cpp
s-perron: Done

source/opt/replace_invalid_opc.cpp
atgoo: I asking for a comment not because I don't understand what it is doing, but to improve readability. Consider how the code looks to somebody who sees it for the first time.

This looks much more readable.
SpvExecutionModel execution_model = GetExecutionModel();
if (execution_model == SpvExecutionModelKernel) {
  // Nothing to do.
  return;
}

if (execution_model == SpvExecutionModelMax) {
  // Multiple execution models in the module. Case not handled.
  return;
}

source/opt/replace_invalid_opc.cpp
atgoo: Same here. I can see what you have written about SpvOpKill in the CL's description, but other people would only see the code, and it is not clear why SpvOpKill was commented.

source/opt/replace_invalid_opc.cpp
s-perron: Sorry, the original comment before out edited it seemed like you were asking under what circumstances `GetExecutionModel` would return ModelMax.  Now that I understand what you really meant, I think you have a good suggestion.  I'll fix that up.

source/opt/replace_invalid_opc.cpp
```diff
@@ -0,0 +1,207 @@
...
+  assert(special_const != nullptr);
+  return const_mgr->GetDefiningInstruction(special_const)->result_id();
+}
+
+std::string ReplaceInvalidOpcodePass::BuildWarningMessage(SpvOp opcode) {
+  spv_opcode_desc opcode_info;
+  context()->grammar().lookupOpcode(opcode, &opcode_info);
+  std::string message = "Removing ";
+  message += opcode_info->name;
+  message += " instruction because of incompatible execution model.";
```
antiagainst: This will concatenate multiple messages without whitespace characters in between if multiple instructions are replaced.

source/opt/replace_invalid_opc.cpp
```diff
@@ -0,0 +1,207 @@
...
+    std::vector<uint32_t> ids;
+    for (uint32_t i = 0; i < type->GetSingleWordInOperand(1); ++i) {
+      ids.push_back(component_const);
+    }
+    special_const = const_mgr->GetConstant(type_mgr->GetType(type_id), ids);
+  } else {
+    assert(type->opcode() == SpvOpTypeInt || type->opcode() == SpvOpTypeFloat);
+    std::vector<uint32_t> literal_words;
+    for (uint32_t i = 0; i < type->GetSingleWordInOperand(0); i += 32) {
+      literal_words.push_back(0xDEADBEEF);
```
dneto0: This is not valid for 16-bit integers or 16bit floats.
Signed integers are sign-extended out to a multiple of 32bits.
Unsigned integers are zero-padded.
16bit floats have 0 as upper bytes.

But I don't think these cases come up for the instructions that we're banning.  I suggest no change for now.

source/opt/replace_invalid_opc.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+    special_const =
+        const_mgr->GetConstant(type_mgr->GetType(type_id), literal_words);
+  }
```
dneto0: You could have an Else clause that generates an OpConstantNull.  That should work for just about everything else, including booleans.

But again, I don't think these cases come up. No change requested.

source/opt/replace_invalid_opc.cpp
dneto0: Wouldn't it be the responsibility of the message consumer to separate messages with newlines if that's appropriate?

test/opt/replace_invalid_opc_test.cpp
```diff
@@ -0,0 +1,590 @@
...
+namespace {
+
+using namespace spvtools;
+
+using ReplaceInvalidOpcodeTest = PassTest<::testing::Test>;
+
+#ifdef SPIRV_EFFCEE
+TEST_F(ReplaceInvalidOpcodeTest, ReplaceInstruction) {
+  const std::string text = R"(
+; CHECK: [[special_const:%\w+]] = OpConstant %float -6.25985e+18
```
dneto0: Ooh.  Beware that different platforms print floating point numbers differently.
I've thought of adding  a "force hex float output" option on the disassembler, which would guard against a failure here.

source/opt/replace_invalid_opc.cpp
antiagainst: Good point!

PR #1254, opened 2018-01-30
Remove constexpr from Analysis operators

alan-baker:
* Had to remove templating from InstructionBuilder as a result
 * now preserved analyses are specified as a constructor argument
* updated tests and uses
* changed static_assert to a runtime assert
 * this should probably get further changes in the future

### DIFF
--- a/source/opt/ir_builder.h
+++ b/source/opt/ir_builder.h
@@ -24,7 +24,7 @@ namespace opt {
 
 // In SPIR-V, ids are encoded as uint16_t, this id is guaranteed to be always
 // invalid.
-constexpr uint32_t kInvalidId = std::numeric_limits<uint32_t>::max();
+const uint32_t kInvalidId = std::numeric_limits<uint32_t>::max();
 
 // Helper class to abstract instruction construction and insertion.
 // The instruction builder can preserve the following analyses (specified via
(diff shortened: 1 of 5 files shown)


source/opt/ir_builder.h
```diff
@@ -27,32 +27,30 @@ namespace opt {
 constexpr uint32_t kInvalidId = std::numeric_limits<uint32_t>::max();
```
dneto0: I think this constexpr will blow up on VS2013 as well.
You can use  either
   const uint32_t kInvalidId = ....
or
  enum : uint32_t {  kInvalidId = ... } ;

source/opt/ir_builder.h
```diff
@@ -203,6 +208,7 @@ class InstructionBuilder {
   ir::IRContext* context_;
   ir::BasicBlock* parent_;
   InsertionPointTy insert_before_;
+  ir::IRContext::Analysis preserved_analyses_;
```
dneto0: This should probably be 'const'

dneto0:
Thanks for the quick turnaround!

dneto0:
Rebased, squashed, and pushed into master as 2735e0851e

PR #1257, opened 2018-01-31
Reordering performance passes ordering to produce better opts

alan-baker:
* Moved initial insert/extract passes later to cover more opportunities
* Added an extra set of passes to clean up opportunities exposed later
in the pipeline

I tested this on some industrial shader corpora. After this change all shaders optimize to fixed point in a single pass of -O. Previously many of the shaders required multiple passes to fully optimize.

### DIFF
--- a/source/opt/optimizer.cpp
+++ b/source/opt/optimizer.cpp
@@ -122,22 +122,22 @@ Optimizer& Optimizer::RegisterPerformancePasses() {
       .RegisterPass(CreateLocalAccessChainConvertPass())
       .RegisterPass(CreateLocalSingleBlockLoadStoreElimPass())
       .RegisterPass(CreateLocalSingleStoreElimPass())
-      .RegisterPass(CreateInsertExtractElimPass())
-      .RegisterPass(CreateDeadInsertElimPass())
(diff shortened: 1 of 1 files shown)


alan-baker:
I ran a couple of shader copora through repeated runs of spirv-opt -O until they reached a fixed point. I had modified the compiler to log which passes which performing optimizations. Through examination of the logs I was able to determine that the posted changes could get all of my test shaders to fixed point in a single pass of -O.

I did not measure compile time. I didn't notice any major compile time increases, but I only have anecdotal evidence. I think this can be decoupled from compile time as we haven't really put any effort yet into reducing compile times.

I also did not measure code size impacts. While reducing code size is very often beneficial for performance, that is not the major objective of -O. Looking at the passes that are newly repeated, I would expect that code size is reduced overall since those optimizations generally produce smaller code size.

greg-lunarg:
> @greg-lunarg <https://github.com/greg-lunarg>, could you test this with
your codebase?

Running -O on my benchmark shader gives size results essentially equivalent
to my current (and current best) size.

Greg

dneto0:
Rebased and pushed into master as abe113219e24

PR #1259, opened 2018-02-01
Add barrier instructions validation pass

atgoo:


atgoo:
Fails Vulkan CTS master.

source/validate_barriers.cpp
```diff
@@ -0,0 +1,285 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Validates correctness of barrier SPIR-V instructions.
+
+#include "latest_version_spirv_header.h"
+#include "validate.h"
```
dneto0: validate.h should be able to stand on its own.  So it seems it should include latest_version_spirv_header.h directly?

source/validate_barriers.cpp
```diff
@@ -0,0 +1,285 @@
...
+#include "spirv_target_env.h"
+#include "val/instruction.h"
+#include "val/validation_state.h"
+
+namespace libspirv {
+
+namespace {
+
+// Returns number of '1' bits in a word.
+uint32_t CountSetBits(uint32_t word) {
```
dneto0: This function already exists twice: in validate_atomics.cpp and validate_image.cpp.    It's ripe for refactoring rather than reimplementing.

source/validate_barriers.cpp
```diff
@@ -0,0 +1,285 @@
...
+  uint32_t count = 0;
+  while (word) {
+    word &= word - 1;
+    ++count;
+  }
+  return count;
+}
+
+// Evaluates int32 value from |id| if possible,
```
dneto0: Should say this works for signed and unsigned cases.

The "if possible" in the comment is less specific than the method name.

Recommend "Tries to evaluate a 32-bit signed or unsigned scalar integer constant.  Returns a tuple ..."

source/validate_barriers.cpp
```diff
@@ -0,0 +1,285 @@
...
+    return SPV_SUCCESS;
+  }
+
+  if (spvIsVulkanEnv(_.context()->target_env)) {
+    if (value != SpvScopeWorkgroup && value != SpvScopeSubgroup) {
+      return _.diag(SPV_ERROR_INVALID_DATA)
+             << spvOpcodeString(opcode)
+             << ": in Vulkan environment Execution Scope is limited to "
+                "Workgroup and Subgroup";
+    }
```
dneto0: Please add a TODO  comment to cover the other supported environments: OpenCL and OpenGL.

source/validate_barriers.cpp
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  if (spvIsVulkanEnv(_.context()->target_env)) {
+    if (value != SpvScopeDevice && value != SpvScopeWorkgroup &&
+        value != SpvScopeInvocation) {
+      return _.diag(SPV_ERROR_INVALID_DATA)
+             << spvOpcodeString(opcode)
+             << ": in Vulkan environment Memory Scope is limited to Device, "
+                "Workgroup and Invocation";
+    }
```
dneto0: Please add a TODO to cover the OpenCL and OpenGL environments.

source/validate_barriers.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+  if (num_memory_order_set_bits > 1) {
+    return _.diag(SPV_ERROR_INVALID_DATA)
+           << spvOpcodeString(opcode)
+           << ": Memory Semantics can have at most one of the following bits "
+              "set: Acquire, Release, AcquireRelease or SequentiallyConsistent";
+  }
+
+  if (spvIsVulkanEnv(_.context()->target_env)) {
+    // TODO(atgoo@github.com): the Vulkan spec is unclear if
```
dneto0: From the Vulkan env spec for SPIR-V (1.0.68):

"Sequentially consistent atomics and barriers are not supported and SequentiallyConsistent is treated as AcquireRelease. SequentiallyConsistent should not be used."
So it's supported but acts as acq_rel.

So I think your logic is correct.

source/validate_barriers.cpp
```diff
@@ -0,0 +1,285 @@
...
+             << ": expected Memory Semantics to include a Vulkan-supported "
+                "storage class";
+    }
+
+    if (opcode == SpvOpControlBarrier && value && !includes_storage_class) {
+      return _.diag(SPV_ERROR_INVALID_DATA)
+             << spvOpcodeString(opcode)
+             << ": expected Memory Semantics to include a Vulkan-supported "
+                "storage class if Memory Semantics is not None";
+    }
```
dneto0: Add TODO for OpenCL and OpenGL environments.

source/validate_barriers.cpp
```diff
@@ -0,0 +1,285 @@
...
+  if (spvIsVulkanEnv(_.context()->target_env)) {
+    // TODO(atgoo@github.com): the Vulkan spec is unclear if
+    // SequentiallyConsistent is supported.
+
+    const bool includes_storage_class =
+        value & (SpvMemorySemanticsUniformMemoryMask |
+                 SpvMemorySemanticsWorkgroupMemoryMask |
+                 SpvMemorySemanticsImageMemoryMask);
+
+    if (opcode == SpvOpMemoryBarrier && !includes_storage_class) {
```
dneto0: From the Vulkan env spec:  "OpMemoryBarrier must not use None".
This is badly worded. It should say "must include exactly one memory order value other than None" or similar.
I think this should also require num_memoyr_order_set_bits == 1.

dneto0:
Manually tested on on VS2017 Release.  Passes.

dneto0:
Rebased, squashed and pushed into master as 12e6860d073b

PR #1260, opened 2018-02-02
Prevent temporary copy of std::vector in decorations

tafuri:
The reason I found this was because I got a `bad_alloc` due to some other bug I am currently investigating.


dneto0:
Looking at types.cpp a little more, I found another case of unnecessary copying decorations, in Struct::GetExtraHashWords.

dneto0:
CLA is not signed.  We can't include this code.
Closing this in favour of #1261

PR #1261, opened 2018-02-02
Avoid vector copies in range-for loops in opt/types.cpp

dneto0:
Also be more explicit about iterated types in other range-for loops.

This subsumes #1260

### DIFF
--- a/source/opt/types.cpp
+++ b/source/opt/types.cpp
@@ -173,7 +173,7 @@ bool Type::operator==(const Type& other) const {
 
 void Type::GetHashWords(std::vector<uint32_t>* words) const {
   words->push_back(kind_);
-  for (auto d : decorations_) {
+  for (const auto& d : decorations_) {
     for (auto w : d) {
       words->push_back(w);
     }
@@ -407,7 +407,7 @@ void RuntimeArray::GetExtraHashWords(std::vector<uint32_t>* words) const {
 
 Struct::Struct(const std::vector<Type*>& types)
     : Type(kStruct), element_types_(types) {
-  for (auto* t : types) {
+  for (const auto* t : types) {
     (void)t;
     assert(!t->AsVoid());
   }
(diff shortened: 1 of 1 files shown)


source/opt/types.cpp
```diff
@@ -173,7 +173,7 @@ bool Type::operator==(const Type& other) const {
 
 void Type::GetHashWords(std::vector<uint32_t>* words) const {
   words->push_back(kind_);
-  for (auto d : decorations_) {
+  for (const auto& d : decorations_) {
```
dneto0: Use reference, not copy of vector.

source/opt/types.cpp
```diff
@@ -455,12 +455,12 @@ std::string Struct::str() const {
...
 void Struct::GetExtraHashWords(std::vector<uint32_t>* words) const {
-  for (auto t : element_types_) {
+  for (auto* t : element_types_) {
     t->GetHashWords(words);
   }
-  for (auto pair : element_decorations_) {
+  for (const auto& pair : element_decorations_) {
     words->push_back(pair.first);
-    for (auto d : pair.second) {
+    for (const auto& d : pair.second) {
```
dneto0: Use a reference, not an expensive copy.

source/opt/types.cpp
```diff
@@ -455,12 +455,12 @@ std::string Struct::str() const {
 }
 
 void Struct::GetExtraHashWords(std::vector<uint32_t>* words) const {
-  for (auto t : element_types_) {
+  for (auto* t : element_types_) {
```
atgoo: auto and auto* don't have much difference, except for more type safety.

source/opt/types.cpp
```diff
@@ -407,7 +407,7 @@ void RuntimeArray::GetExtraHashWords(std::vector<uint32_t>* words) const {
 
 Struct::Struct(const std::vector<Type*>& types)
     : Type(kStruct), element_types_(types) {
-  for (auto* t : types) {
+  for (const auto* t : types) {
```
atgoo: This looks like an empty loop for NDEBUG.

dneto0: Yes.

source/opt/types.cpp
dneto0: Yes.  After seeing #1260 I went through and annotated things more explicitly.  Having the * there makes it clearer this is an inexpensive iteration.   So it helps verifiability.

dneto0:
Rebased.  Keeping approval.

alan-baker:
Rebased and submitted as 3ef4bb600f2ac09fda73671617e7793b5891afea.

Tested manually on windows.

PR #1263, opened 2018-02-02
Build SPIRV-Tools as a shared library

### LINKED ISSUE #1240: Linking SPIRV-Tools with C code (shared/dynamic library request)
Using SPIRV-Tools in C code is a bit problematic.

Let's take as an example the following code:
```
#include <stdio.h>

int main(int argc, char **argv)
{
    spv_context context = spvContextCreate(SPV_ENV_VULKAN_1_0);
    spvContextDestroy(context);
    return 0;
}
```

The usual way of building it fails:
```
$ gcc main.c $(pkg-config --cflags --libs SPIRV-Tools)

/home/joseph/src/vulkan/install/lib64/libSPIRV-Tools.a(table.cpp.o): In function `spvContextCreate':
/home/joseph/src/vulkan/SPIRV-Tools/source/table.cpp:52: undefined reference to `operator new(unsigned long)'
/home/joseph/src/vulkan/install/lib64/libSPIRV-Tools.a(table.cpp.o): In function `spvContextDestroy':
/home/joseph/src/vulkan/SPIRV-Tools/source/table.cpp:55: undefined reference to `operator delete(void*)'

...

collect2: error: ld returned 1 exit status
```

I can fix this by linking with `libstdc++` or by using `g++`, but I would like to avoid this if possible.

Could SPIRV-Tools be also built as a shared library?

jozefkucia:
Fixes #1240.

Tested on Linux.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -31,7 +31,7 @@ endif()
 function(add_spvtools_unittest)
   if (NOT "${SPIRV_SKIP_TESTS}" AND TARGET gmock_main)
     set(one_value_args TARGET)
-    set(multi_value_args SRCS LIBS)
+    set(multi_value_args SRCS LIBS ENVIRONMENT)
     cmake_parse_arguments(
       ARG "" "${one_value_args}" "${multi_value_args}" ${ARGN})
     set(target test_${ARG_TARGET})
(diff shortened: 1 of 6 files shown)


test/CMakeLists.txt
```diff
@@ -156,7 +159,8 @@ add_spvtools_unittest(
 add_spvtools_unittest(
   TARGET c_interface_shared
   SRCS c_interface_test.cpp
-  LIBS ${SPIRV_TOOLS}-shared)
+  LIBS ${SPIRV_TOOLS}-shared
+  ENVIRONMENT PATH=$<TARGET_FILE_DIR:${SPIRV_TOOLS}-shared>)
```
dneto0: This is needed Windows so it can find the DLL.
I wonder why Linux (and OSX) don't need a modification for LD_LIBRARY_PATH?

dneto0:
Rebased, squashed, and pushed into master as 3013897556

test/CMakeLists.txt
jozefkucia: I think the answer is `RUNPATH`:

```
$ objdump -x test_c_interface_shared | grep PATH
  RUNPATH              /home/joseph/src/vulkan/SPIRV-Tools/build/source
```

jozefkucia:
Thanks!

PR #1266, opened 2018-02-05
Make use of the instruction folder.

s-perron:
Implementation of the simplification pass.

- Create pass that calls the instruction folder on each instruction and
  propagate instructions that fold to a copy.  This will do copy
  propagation as well.

- Did not use the propagator engine because I want to modify the instruction
  as we go along.  It allows the folding code to be much simpler because we 
 do not need the map for the ids since we can change the ids as we go along.

- Change folding to not allocate new instructions, but make changes in
  place.  This change had a big impact on compile time.

- Add simplification pass to the legalization passes in place of
  insert-extract elimination.


s-perron:
Still need to add test.  Please comment on the overall design for what is already there.

source/opt/folding_rules.cpp
```diff
@@ -113,6 +207,49 @@ FoldingRule InsertFeedingExtract() {
...
+
+        for (uint32_t i = 0; i < inst->NumInOperands(); i += 2) {
+          uint32_t op_id = inst->GetSingleWordInOperand(i);
+          if (op_id == inst->result_id()) {
+            continue;
+          }
+
+          ir::Instruction* op_inst = def_use_mgr->GetDef(op_id);
+          if (op_inst->opcode() == SpvOpUndef) {
+            continue;
```
s-perron: This case is probably wrong.  We might need to figure out something more precise because I do see Undef's in some real cases.

source/opt/simplification_pass.cpp
```diff
@@ -0,0 +1,114 @@
...
+        process_phis.insert(inst);
+      }
+
+      if (FoldInstruction(inst)) {
+        modified = true;
+        context()->AnalyzeUses(inst);
+        get_def_use_mgr()->ForEachUser(
+            inst, [&work_list, &process_phis](ir::Instruction* use) {
+              if (process_phis.count(use)) {
+                work_list.push_back(use);
```
alan-baker: Is it worth preventing adding the same phi to the worklist multiple times?

source/opt/simplification_pass.cpp
```diff
@@ -0,0 +1,114 @@
...
+  // Using a set to make sure entries are unique, and the traversal order is
+  // deterministic.
+  while (!work_list.empty()) {
+    ir::Instruction* inst = work_list.back();
+    work_list.pop_back();
+    if (FoldInstruction(inst)) {
+      modified = true;
+      context()->AnalyzeUses(inst);
+      get_def_use_mgr()->ForEachUser(inst, [&work_list](ir::Instruction* use) {
+        work_list.push_back(use);
```
alan-baker: Prevent multiple copies of the same user from being enqueued on the worklist? Might have to change to a breadth first traversal.

source/opt/folding_rules.cpp
```diff
@@ -42,6 +42,100 @@ FoldingRule IntMultipleBy1() {
...
+    std::vector<ir::Operand> operands;
+    operands.push_back({SPV_OPERAND_TYPE_ID, {element_id}});
+
+    // Add the remaining indicies for extraction.
+    for (uint32_t i = 2; i < inst->NumInOperands(); ++i) {
+      operands.push_back(
+          {SPV_OPERAND_TYPE_ID, {inst->GetSingleWordInOperand(i)}});
+    }
+
+    // If there were no extra indicies, then we have the final object.  No need
```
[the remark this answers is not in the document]
s-perron: done

source/opt/folding_rules.cpp
```diff
@@ -42,6 +42,100 @@ FoldingRule IntMultipleBy1() {
...
+    }
+
+    // The last check it to see that the object being extracted from is the
+    // correct type.
+    ir::Instruction* original_inst = def_use_mgr->GetDef(original_id);
+    if (original_inst->type_id() != inst->type_id()) {
+      return false;
+    }
+
+    // Simplfy by using the original object.
```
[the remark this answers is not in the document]
s-perron: done

source/opt/folding_rules.cpp
```diff
@@ -113,6 +207,51 @@ FoldingRule InsertFeedingExtract() {
...
+          }
+        }
+
+        if (incoming_value == 0) {
+          // Code looks invalid.  Don't do anything.
+          return false;
+        }
+
+        // We have a single incoming value.  Simplify using that value.
+        inst->SetOpcode(SpvOpCopyObject);
```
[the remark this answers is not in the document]
s-perron: Added a comment to `FoldInstruction`.

source/opt/folding_rules.cpp
```diff
@@ -124,6 +263,11 @@ spvtools::opt::FoldingRules::FoldingRules() {
   rules[SpvOpIMul].push_back(IntMultipleBy1());
```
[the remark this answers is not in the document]
s-perron: Done

source/opt/optimizer.cpp
```diff
@@ -103,7 +104,8 @@ Optimizer& Optimizer::RegisterLegalizationPasses() {
           .RegisterPass(CreateLocalMultiStoreElimPass())
           // Copy propagate members.  Cleans up code sequences generated by
           // scalar replacement.
-          .RegisterPass(CreateInsertExtractElimPass())
+          .RegisterPass(CreateSimplificationPass())
+          //.RegisterPass(CreateInsertExtractElimPass())
```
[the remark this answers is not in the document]
s-perron: I left is because I was playing around with it. I'll remove the old code.

For -O, I want to add more rules and then replace InsertExtractElimPass.  There are still somethings the folder cannot do that the other pass does.

For -Os, I will make @greg-lunarg aware of what is being done, but let him decide what he wants to do with that set of optimizations.

test/opt/fold_test.cpp
```diff
@@ -656,18 +660,27 @@ TEST_P(GeneralInstructionFoldingTest, Case) {
...
+    for (uint32_t i = 0; i < inst->NumInOperands(); ++i) {
+      EXPECT_EQ(inst->GetOperand(i), original_inst->GetOperand(i));
+    }
   }
 }
 
 // clang-format off
-INSTANTIATE_TEST_CASE_P(TestCase, GeneralInstructionFoldingTest,
-::testing::Values(
+INSTANTIATE_TEST_CASE_P(IntegerArimaticTestCases, GeneralInstructionFoldingTest,
```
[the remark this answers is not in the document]
s-perron: Done

source/opt/simplification_pass.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+  for (ir::Function& function : *get_module()) {
+    modified |= SimplifyFunction(&function);
+  }
+  return (modified ? Status::SuccessWithChange : Status::SuccessWithoutChange);
+}
+
+bool SimplificationPass::SimplifyFunction(ir::Function* function) {
+  bool modified = false;
+  // Phase 1: Traverse all instruction in dominance order.
```
dnovillo: s/instruction/instructions/

s-perron:
Yes, I will admit this will not get all of the cases that using the propagator will miss.  The things it will miss are those that get exposed because of dead branches.  We could run dead branch elimination before hand, but we would still miss those that we realize are dead during the simplification pass.  

I did not use the propagator because of the complication it would add to managing the memory of the instructions.  What I found very early was that creating and deleting instructions during this pass is very very expensive.  The compile time is much better if we do the changes in place.

However, the propagator relies on doing a complete analysis, and then changing the code.  The change from interesting to varying would require reverting back to the original.  This means creating copies of instructions, and keeping track of them.

The other issue is how to handle the lattice.  It seems like the propagator has only 3 levels {not interesting, interesting, varying}.  However, simplifying can give a lot of different "interesting" values, that might have a more complicated hierarchy.  They do not from a simple 3 level lattice.  The mapping for this is not very clear.  If there is a way of doing this, you can show me, and then we can talk about it.

dnovillo:
On Tue, Feb 6, 2018 at 2:52 PM, Steven Perron <notifications@github.com>
wrote:

> Yes, I will admit this will not get all of the cases that using the
> propagator will miss. The things it will miss are those that get exposed
> because of dead branches. We could run dead branch elimination before hand,
> but we would still miss those that we realize are dead during the
> simplification pass.
>
> I did not use the propagator because of the complication it would add to
> managing the memory of the instructions. What I found very early was that
> creating and deleting instructions during this pass is very very expensive.
> The compile time is much better if we do the changes in place.
>
> However, the propagator relies on doing a complete analysis, and then
> changing the code. The change from interesting to varying would require
> reverting back to the original. This means creating copies of instructions,
> and keeping track of them.
>
> The other issue is how to handle the lattice. It seems like the propagator
> has only 3 levels {not interesting, interesting, varying}. However,
> simplifying can give a lot of different "interesting" values, that might
> have a more complicated hierarchy. They do not from a simple 3 level
> lattice. The mapping for this is not very clear. If there is a way of doing
> this, you can show me, and then we can talk about it.
>

Yup, thanks for confirming.  It's indeed simpler to handle it this way.
For copy-prop, we could do a lattice, but it's true that for actual
simplification (x + y -> x + 0 -> x), we could end up with several levels.

Most likely easier to just schedule this pass a couple of times, or, even,
have the pass itself run N times until convergence (N < 3?).  But, let's
have the pass in first.  We can tweak later.

source/opt/cfg.cpp
```diff
@@ -111,5 +124,18 @@ void CFG::ComputeStructuredSuccessors(ir::Function* func) {
   }
 }
 
+void CFG::ComputePostOrderTraversal(BasicBlock* bb,
+                                    std::vector<BasicBlock*>& order,
```
alan-baker: Syle guide says all references must be const [here](https://google.github.io/styleguide/cppguide.html#Reference_Arguments).

source/opt/folding_rules.cpp
```diff
@@ -42,6 +42,100 @@ FoldingRule IntMultipleBy1() {
...
+    uint32_t cid = inst->GetSingleWordInOperand(kExtractCompositeIdInIdx);
+    ir::Instruction* cinst = def_use_mgr->GetDef(cid);
+
+    if (cinst->opcode() != SpvOpCompositeConstruct) {
+      return false;
+    }
+
+    // Get the element being extracted from the OpCompositeConstruct
+    uint32_t element_index = inst->GetSingleWordInOperand(1);
+    uint32_t element_id = cinst->GetSingleWordInOperand(element_index);
```
alan-baker: OpCompositeConstruct can create a vector out of component vectors (e.g. float4 out of two float2s or a float2 and 2 floats). If the code is:
```
%construct = OpCompositeConstruct %float4 %float2a %float2b
%extract = OpCompositeExtract %float %construct 3
```
Then this rule will have an out-of-bounds access on the operands. I think you need a special case for vectors. This should also be an unit test.

source/opt/folding_rules.cpp
```diff
@@ -42,6 +42,100 @@ FoldingRule IntMultipleBy1() {
...
+
+    std::vector<ir::Operand> operands;
+    operands.push_back({SPV_OPERAND_TYPE_ID, {element_id}});
+
+    // Add the remaining indicies for extraction.
```
alan-baker: indicies -> indices

source/opt/simplification_pass.cpp
```diff
@@ -0,0 +1,115 @@
...
+            process_phis.insert(inst);
+          }
+
+          if (inst->opcode() == SpvOpCopyObject || FoldInstruction(inst)) {
+            modified = true;
+            context()->AnalyzeUses(inst);
+            get_def_use_mgr()->ForEachUser(
+                inst, [&work_list, &process_phis,
+                       &in_work_list](ir::Instruction* use) {
+                  if (process_phis.count(use) && !in_work_list.count(use)) {
```
alan-baker: ... && !in_work_list.insert(use).second))  {

Allows you to remove the subsequent insert.

source/opt/simplification_pass.cpp
```diff
@@ -0,0 +1,115 @@
...
+  //          has already finished.
+  for (size_t i = 0; i < work_list.size(); ++i) {
+    ir::Instruction* inst = work_list[i];
+    in_work_list.erase(inst);
+    if (FoldInstruction(inst)) {
+      modified = true;
+      context()->AnalyzeUses(inst);
+      get_def_use_mgr()->ForEachUser(
+          inst, [&work_list, &in_work_list](ir::Instruction* use) {
+            if (!use->IsDecoration() && !in_work_list.count(use)) {
```
alan-baker: Same as above re:insert. Also you may want to skip OpName too if you're skipping decorations.

source/opt/cfg.cpp
s-perron: Yes, that is correct.  Fixed.

source/opt/folding_rules.cpp
s-perron: Good catch.  I've updated the folding rule to handle this.  I've also added a few tests.

source/opt/simplification_pass.cpp
s-perron: fixed.

source/opt/simplification_pass.cpp
s-perron: fixed.

PR #1267, opened 2018-02-05
VS2013: LoopDescriptor LoopContainerType can't contain unique_ptr

### LINKED ISSUE #1262: VisualStudio 2013 can't compile SPIRV-Tools
There is still interest in building SPIRV-Tools with VisualStudio 2013.  VS 2013 has known bugs in its C++11 support, so we would have to work around those to make it build.
We would have to exclude optional components used in testing -- effcee and RE2 -- since effcee depends on RE2 and RE2 rejects VS 2013.

I'm attaching a build log with VS2013.
[vs2013.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/1690925/vs2013.txt)

dneto0:
The loop descriptor must explicitly manage the storage for contained
Loop objects.

Fixes #1262

### DIFF
--- a/source/opt/loop_descriptor.cpp
+++ b/source/opt/loop_descriptor.cpp
@@ -296,7 +296,11 @@ bool Loop::IsLCSSA() const {
   return true;
 }
 
-LoopDescriptor::LoopDescriptor(const Function* f) { PopulateList(f); }
+LoopDescriptor::LoopDescriptor(const Function* f) : loops_() {
+  PopulateList(f);
+}
+
+LoopDescriptor::~LoopDescriptor() { ClearLoops(); }
 
 void LoopDescriptor::PopulateList(const Function* f) {
   IRContext* context = f->GetParent()->context();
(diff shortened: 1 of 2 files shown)


dneto0:
Tested on VS2013 Release, VS2017 Release, VS2017 Debug, 
Linux GCC 6.3 Release, GCC 6.3 Debug

atgoo:
Does vector<unique_ptr> not work in general with VS 2013, or is it maybe missing some constructors or operators which are autogenerated by other compilers?

dneto0:
> Does vector<unique_ptr> not work in general with VS 2013, or is it maybe missing some constructors or operators which are autogenerated by other compilers? 

I think VS2013 is missing a specialization of std::vector<std::unique_ptr>.
See my initial speculcation in https://github.com/KhronosGroup/glslang/pull/1234#issuecomment-362299717

However, I see there are other cases of std::vector<std::unique_ptr<...>> in SPIRV-Tools.  So there must be another ingredient that I haven't tracked down.   

An interesting thing about the use inside LoopDescriptor is that a LoopDescriptor is copy-constructed when populating a map in IRContext::GetLoopDescriptor.  That's why I needed to delete its copy-constructor and make a move-constructor instead.

dneto0:
Rebased and pushed into master.

PR #1268, opened 2018-02-05
ADCE: Fix combinator initialization

greg-lunarg:
The combinator initialization was only looking at the capabilities
in the original shader code and not the inferred capabilities.
Geometry and tessellation shaders compiled by glslang did not
have the Shader capability explicitly set in the original code as it
is inferred. So the combinator set was not initialized correctly
causing problems for ADCE.

### DIFF
--- a/source/opt/feature_manager.h
+++ b/source/opt/feature_manager.h
@@ -41,6 +41,11 @@ class FeatureManager {
   // Analyzes |module| and records enabled extensions and capabilities.
   void Analyze(ir::Module* module);
 
+  libspirv::CapabilitySet* GetCapabilities() { return &capabilities_; }
+  const libspirv::CapabilitySet* GetCapabilities() const {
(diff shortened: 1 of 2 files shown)


source/opt/feature_manager.h
```diff
@@ -41,6 +41,11 @@ class FeatureManager {
   // Analyzes |module| and records enabled extensions and capabilities.
   void Analyze(ir::Module* module);
 
+  libspirv::CapabilitySet* GetCapabilities() { return &capabilities_; }
```
dneto0: I would have gone for returning a reference, but this is fine.

dneto0:
I can handle the formatting fix upon integration.

dneto0:
Manually tested on VS 2017 and VS 2013.  Passes.
Rebased, reformatted, and pushed onto master as 860b2ee5fc850

PR #1269, opened 2018-02-05
Fix type manager memory management for multi-level types

### LINKED ISSUE #1265: Use after free in if-conversion
The underlying problem is memory management of types. It can be fixed in IfConversion, but the design intent of the TypeManager was to allow that sort of type construction so I consider this a problem in the TypeManager. Minimal testcase for exposing this issue:
```
TEST(TypeManager, RegisterAndRemoveId) {
  const std::string text = R"(
OpCapability Shader
OpMemoryModel Logical GLSL450
%1 = OpTypeInt 32 0
  )";

  std::unique_ptr<ir::IRContext> context =
      BuildModule(SPV_ENV_UNIVERSAL_1_2, nullptr, text,
                  SPV_TEXT_TO_BINARY_OPTION_PRESERVE_NUMERIC_IDS);
  EXPECT_NE(context, nullptr);

  uint32_t id = 2u;
  {
    // Ensure that u32 goes out of scope.
    Integer u32(32, false);
    Struct st({&u32});
    context->get_type_mgr()->RegisterType(id, st);
  }

  context->get_type_mgr()->RemoveId(id);
  EXPECT_EQ(nullptr, context->get_type_mgr()->GetType(id));
}
```

I think that TypeManager::RegisterType should rebuild the incoming type in terms of the type pool to prevent this sort of problem.

alan-baker:
Fixes #1265 

The type manager now rebuilds component types in terms of memory it owns in addition to the input type. The allows the usage that follows to be safe if that type is later removed during subsequent optimizations.
```
Bool bool_ty;
Vector vec_ty(&bool_ty, 3);
uint32_t id = context()->get_type_mgr()->GetTypeInstruction(&vec_ty);
```

### DIFF
--- a/source/opt/type_manager.cpp
+++ b/source/opt/type_manager.cpp
@@ -349,16 +349,166 @@ void TypeManager::CreateDecoration(uint32_t target,
   context()->get_def_use_mgr()->AnalyzeInstUse(inst);
 }
 
+Type* TypeManager::RebuildType(const Type& type) {
+  // The comparison and hash on the type pool will avoid inserting the rebuilt
+  // type if an equivalent type already exists. The rebuilt type will be deleted
+  // when it goes out of scope at the end of the function in that case. Repeated
+  // insertions of the same Type will atmost keep one corresponding object in
+  // the type pool.
+  Type* rebuilt_ty = nullptr;
+  switch (type.kind()) {
+#define DefineNoSubtypeCase(kind)                              \
+  case Type::k##kind:                                          \
+    rebuilt_ty = type_pool_.insert(type.Clone()).first->get(); \
+    break;
+    DefineNoSubtypeCase(Void);
+    DefineNoSubtypeCase(Bool);
+    DefineNoSubtypeCase(Integer);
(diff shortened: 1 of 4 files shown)


source/opt/type_manager.cpp
```diff
@@ -349,16 +349,145 @@ void TypeManager::CreateDecoration(uint32_t target,
   context()->get_def_use_mgr()->AnalyzeInstUse(inst);
 }
 
+Type* TypeManager::RebuildType(const Type& type) {
+  // The comparison and hash on the type pool will avoid inserting the rebuilt
+  // type if an equivalent type already exists. The rebuilt type will be deleted
+  // when it goes out of scope at the end of the function in that case. Repeated
+  // insertions of the same Type will atmost keep one corresponding object in
```
dnovillo: s/atmost/, at most,/

alan-baker:
Rebased, squashed and committed in 871022772e865f314696d04b7180c3aca3e46590.

Fixes #1265

PR #1270, opened 2018-02-06
Enable Visual Studio 2013 again

dneto0:
Disable use of Effcee and RE2 with MSVC compilers older than VisualStudio 2015
since RE2 doesn't support them.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,6 +9,7 @@ version: "{build}"
 os:
   - Visual Studio 2017
   - Visual Studio 2015
+  - Visual Studio 2013
 
 platform:
   - x64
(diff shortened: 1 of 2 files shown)


dneto0:
Updated the commit with my correct email

dnovillo:
Or, alternately, we could *always* enable RE if VS 2013 is not detected.  Would that work?

dneto0:
> This means that as long as we need to support VS 2013, we cannot demand RE in tests, right? 

Yes, unfortunately.

> Or, alternately, we could always enable RE if VS 2013 is not detected. Would that work?

Currently downstream projects are not required to download Effcee or RE2 source.  But we could change that by forcing the issue.

PR #1273, opened 2018-02-06
Add Vulkan-specific validation rules for atomics

atgoo:
Added atomic instructions validation rules from
https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#spirvenv-module-validation

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -766,4 +766,22 @@ bool ValidationState_t::GetConstantValUint64(uint32_t id, uint64_t* val) const {
   return true;
 }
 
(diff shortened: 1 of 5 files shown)


atgoo:
Fails Vulkan CTS master. Looking into this now.

atgoo:
Disabled the check which fails Vulkan CTS.

dneto0:
I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/1278 to re-enable the check that was disabled, and to track the CTS change in flight to fix it.

dneto0:
Rebased and pushed into master as a61e4c13562
I didn't squash this time because I want to keep separate the commit that disables the check, so it can more easily be reversed.

PR #1274, opened 2018-02-07
SROA: Do scalar replacement on structs with no partial references.

greg-lunarg:
This change allows for scalar replacement of structs with no references to their individual components. This was previously disabled, likely due to concern about increasing shader size.

While theoretically one might be able to write a shader for which this change may cause the shader to become larger when optimized, in practice all the shaders observed did not get larger, and a few shaders experienced moderate to large size decreases. This is because SROA allows better analysis of struct references enabling improved dead code detection, among other optimizations.

The new test in this commit demonstrates the code pattern which benefits from this change. It is a pattern similar to that generated when a struct valued function with early returns is inlined. This test also demonstrates how such a pattern can mask dead code and how this change allows for its deletion by subsequent passes.

### DIFF
--- a/test/opt/scalar_replacement_test.cpp
+++ b/test/opt/scalar_replacement_test.cpp
@@ -1139,7 +1139,7 @@ OpFunctionEnd
 #endif  // SPIRV_EFFCEE
 
 TEST_F(ScalarReplacementTest, ReplaceNoPartialRefs) {
-// Replace structs whose individual components are not referenced.
+  // Replace structs whose individual components are not referenced.
   const std::string before_predefs =
       R"(OpCapability Shader
 %1 = OpExtInstImport "GLSL.std.450"
--- a/test/opt/scalar_replacement_test.cpp
+++ b/test/opt/scalar_replacement_test.cpp
@@ -1138,4 +1138,201 @@ OpFunctionEnd
 }
 #endif  // SPIRV_EFFCEE
 
+TEST_F(ScalarReplacementTest, ReplaceNoPartialRefs) {
+// Replace structs whose individual components are not referenced.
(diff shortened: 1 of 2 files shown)


greg-lunarg:
@alan-baker Please review this change. Are you aware of specific real-life shaders for which this change would increase their optimized size?

s-perron:
The effect this has is that when we do the load-store removal, the OpPhi instructions that get generated will be split up.  Without this change, there will be 1 OpPhi merging the entire structure.  I can definitely see this helping some of the shaders I've been looking at.

test/opt/scalar_replacement_test.cpp
```diff
@@ -1138,4 +1138,201 @@ OpFunctionEnd
...
+OpBranch %46
+%46 = OpLabel
+%78 = OpPhi %float %35 %47 %41 %48
+OpStore %fo %78
+OpReturn
+OpFunctionEnd
+)";
+
+  AddPass<opt::ScalarReplacementPass>();
+  AddPass<opt::LocalMultiStoreElimPass>();
```
alan-baker: Please don't run extra passes for the unit test. I appreciate the intent, but these tests should focus on just scalar-replacement.

greg-lunarg:
I believe I have addressed all concerns with my last round of commits.

s-perron:
@greg-lunarg Does this have an issue associated with it?

greg-lunarg:
> Does this have an issue associated with it?

No. I wrote the PR to include what I would put into an issue.

alan-baker:
Rebased, squashed and committed as ca4457b4b6e20f8f37731a3cf086e963fa52ff31.

PR #1279, opened 2018-02-08
Add header files missing from CMakeLists.txt

s-perron:


s-perron:
The ide I use was giving me messages that they are not part of a project, and would not index them.

PR #1280, opened 2018-02-08
Fix generation of Vim syntax file

dneto0:


dneto0:
Rebased.  Keeping LGTM

PR #1283, opened 2018-02-09
Add folding of Extract and Construct constant instructions

s-perron:
- Create files for constant folding rules.
- Add the rules for OpCompositeConstruct and OpCompositeExtract.

### DIFF
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -22,6 +22,7 @@ add_library(SPIRV-Tools-opt
   common_uniform_elim_pass.h
   compact_ids_pass.h
   composite.h
+  const_folding_rules.h
   constants.h
   dead_branch_elim_pass.h
   dead_insert_elim_pass.h
@@ -94,6 +95,7 @@ add_library(SPIRV-Tools-opt
   common_uniform_elim_pass.cpp
   compact_ids_pass.cpp
   composite.cpp
+  const_folding_rules.cpp
   constants.cpp
   dead_branch_elim_pass.cpp
   dead_insert_elim_pass.cpp
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -1,0 +1,92 @@
+// Copyright (c) 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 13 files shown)


source/opt/const_folding_rules.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+#include "const_folding_rules.h"
+
+namespace spvtools {
+namespace opt {
+
+namespace {
+const uint32_t kExtractCompositeIdInIdx = 0;
+
+ConstantFoldingRule FoldExtractWithConstants() {
```
alan-baker: Need a method comment.

source/opt/const_folding_rules.cpp
```diff
@@ -0,0 +1,95 @@
...
+      auto cc = c->AsCompositeConstant();
+      assert(cc != nullptr);
+      auto components = cc->GetComponents();
+      c = components[element_index];
+    }
+    return c;
+  };
+}
+
+ConstantFoldingRule FoldCompositeWithConstants() {
```
alan-baker: Here too.

source/opt/const_folding_rules.cpp
```diff
@@ -0,0 +1,95 @@
...
+              const_mgr->GetDefiningInstruction(&null_const);
+          real_const = const_mgr->GetConstantFromInst(const_inst);
+        }
+        return real_const;
+      }
+
+      auto cc = c->AsCompositeConstant();
+      assert(cc != nullptr);
+      auto components = cc->GetComponents();
+      c = components[element_index];
```
alan-baker: OpConstantComposite can have OpUndef as an element. What happens in this case?

source/opt/CMakeLists.txt
```diff
@@ -94,6 +95,7 @@ add_library(SPIRV-Tools-opt
   common_uniform_elim_pass.cpp
   compact_ids_pass.cpp
   composite.cpp
+  const_folding_rules.cpp
```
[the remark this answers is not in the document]
s-perron: Fixed

source/opt/const_folding_rules.cpp
s-perron: done

source/opt/const_folding_rules.cpp
s-perron: done

source/opt/const_folding_rules.cpp
s-perron: I forgot about that.  I think "c" ends up becoming nullptr.  I'll have to check that.

s-perron: The "constant" with the undef is not considered a constant by the constant manager.  So we hit the early exit at the top.

I remember an issue with this because, when specializing constants, we are not allowed to fold an OpSpecConstantComposite that contains an OpUndef.  I did not want to deal with that, so I left it as is.  At the time, I wanted to treat all OpUndef as if the were OpNull in the constant manager.

I've added a test case for this.

source/opt/const_folding_rules.cpp
dneto0: Method comments go before the declaration.
Comments inside the body are implementation notes.

dneto0:
Rebased, squashed, and pushed into master as 1d7b1423f939027

PR #1287, opened 2018-02-10
Avoid using the def-use manager during inlining.

### LINKED ISSUE #1285: Random failures during inlining.
When I run many large shader through the optimizer, I get a few failures.  When I try rerunning in a debugger, they do not reproduce.  They cannot be made available here, but the error is an assert failure because the unique id for an instruction is 0.

With some tracing, I found that is happening when adding a new use to the def-use manager.  When trying to sort the uses by unique id, on of the already existing instruction has a unique id of 0.

I believe what is happening is that the def-use manager is used while inlining, so it gets build, but it goes stall when a function is inlined.  When that happens, the def-use manager contains a dangling pointer.  That memory is sometimes reused and over written with a 0.

s-perron:
There seems to only be a single location where the def-use manager is
used.  It is to get information about a type.  We can do that with the
type manager instead.

Fixes #1285

### DIFF
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -150,9 +150,8 @@ uint32_t InlinePass::CreateReturnVar(
     std::vector<std::unique_ptr<ir::Instruction>>* new_vars) {
   uint32_t returnVarId = 0;
   const uint32_t calleeTypeId = calleeFn->type_id();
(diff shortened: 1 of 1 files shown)


s-perron:
Functional.  See #1285.

dnovillo:
Ah thanks. I missed the reference to the bug.

PR #1288, opened 2018-02-12
Consistently include latest spirv.h header file.

dneto0:
Use indirection through latest_version_spirv.h

Also, when generating enum tables, use the unified1 JSON grammar since
it now has FragmentFullyCoveredEXT but the other JSON grammars don't.
They are starting to fall behind.

This is a followup to #1286


source/enum_string_mapping.h
```diff
@@ -17,9 +17,8 @@
 
 #include <string>
 
-#include "spirv/1.1/spirv.h"
-
 #include "extensions.h"
+#include "latest_version_spirv_header.h"
```
atgoo: This header should be included before other local headers. Please also add blank line, so that clang-format doesn't reorder.

source/name_mapper.cpp
```diff
@@ -23,8 +23,8 @@
 #include <unordered_set>
 
 #include "spirv-tools/libspirv.h"
-#include "spirv/1.2/spirv.h"
 
+#include "latest_version_spirv_header.h"
```
atgoo: Please add blank line, so that clang-format doesn't reorder headers.

source/opt/const_folding_rules.h
```diff
@@ -22,7 +22,7 @@
 #include "folding_rules.h"
 #include "ir_builder.h"
 #include "ir_context.h"
-#include "spirv/1.2/spirv.h"
+#include "latest_version_spirv_header.h"
```
atgoo: This header should be included before other local headers. Please also add blank line, so that clang-format doesn't reorder.

source/binary.cpp
```diff
@@ -25,9 +25,9 @@
 #include "assembly_grammar.h"
 #include "diagnostic.h"
 #include "ext_inst.h"
+#include "latest_version_spirv_header.h"
```
atgoo: This header should be included before other local headers. Please also add blank line, so that clang-format doesn't reorder.

source/enum_string_mapping.h
dneto0: I disagree. It's a local header.
From the style guide:

> Use standard order for readability and to avoid hidden dependencies: Related header, C library, C++ library, other libraries' .h, your project's .h.

source/name_mapper.cpp
dneto0: See above

source/opt/const_folding_rules.h
dneto0: Same

source/binary.cpp
dneto0: Same

dneto0:
Rebased.  Keeping the LGTM

dneto0:
Since this has been a while, I'll wait for the bots to go green.

PR #1293, opened 2018-02-13
Add id to name map

### LINKED ISSUE #1290: Too much time spent in `KillNamesAndDecorates`
I did some profiling of the optimizer on a about 15 shaders, and it seems like the optimizer is spending over half of its time in `KillNamesAndDecorates`.  This is not reasonable.

Looking at the code, my guess the time is spent in the search for the name corresponding to the id.  A lot of time is also spent in `GetSingleWordInOperand`, which is in that loop.

We should add a map from an id to its name.

s-perron:
Adding a map from an id to it set of OpName and OpMemberName
instructions.  This will be used in KillNameAndDecorates to kill the
names for the ids that are being removed.

In my test, the compile time for 50 shaders went from 1m57s to 55s.
This was on linux using the release build.

Fixes #1290.

### DIFF
--- a/source/opt/ir_context.cpp
+++ b/source/opt/ir_context.cpp
@@ -42,6 +42,9 @@ void IRContext::BuildInvalidAnalyses(IRContext::Analysis set) {
   if (set & kAnalysisLoopAnalysis) {
     ResetLoopAnalysis();
   }
+  if (set & kAnalysisNameMap) {
+    BuildIdToNameMap();
+  }
 }
 
 void IRContext::InvalidateAnalysesExceptFor(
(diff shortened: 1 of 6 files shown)


s-perron:
I case others are wondering, I chose not to use the def-use manager because not all transformations keep it up to date.  One example, is inlining.  It needs to remove the names, but does not keep the def-use manager up-to-date.  If we were to use it, the def-use manager would have to be rebuilt for every function call that is inlined.

source/opt/ir_context.cpp
```diff
@@ -206,6 +223,17 @@ void spvtools::ir::IRContext::ForgetUses(Instruction* inst) {
       get_decoration_mgr()->RemoveDecoration(inst);
     }
   }
+  if (id_to_name_) {
```
dneto0: This code is almost identical to the code at line 109.   I'd like to see this pulled out as a helper method.

s-perron:
@dneto0 The andriod build timed out.  I don't think this is an issue with my change.  Let me know if you think it is.

PR #1296, opened 2018-02-13
tools/linker: Allow setting --verify-ids on the command line

pierremoreau:


dneto0:
Rebased and pushed into master as 64298bfd

pierremoreau:
Thank you for the review.

Issue #1144, opened 2017-12-30
Allow for partial linking


doe300:
Problem:
The linker currently expects all functions imported by one of the modules linked together to be exported by another module, otherwise it fails. This works well for linking together a final binary module, but does not allow for linking partially-linked modules, which have still unresolved import linkages.

Use-case:
I have an [OpenCL compiler](https://github.com/doe300/VC4C), that uses the SPIRV-Tools to read SPIR-V input, link SPIR-V modules if necessary, and convert it into an internal representation. In the device-side standard-library (which is linked into the kernel code) I also define some additional intrinsic functions to access special hardware components (like mutex, semaphores, etc.). These functions are then replaced by my compiler with the corresponding hardware-instructions and therefore not defined in any SPIR-V module. This works fine with a single SPIR-V module, since imported functions are accepted by SPIRV-Tools, but not when linking modules, since the linker does not allow for any functions to remain imported. See also [here](https://github.com/doe300/VC4C/issues/1) for a discussion on this.

Solution:
Allow the linker to create "partially linked" modules by not throwing errors when there is no exported symbol matching an imported one, instead it should leave the imported symbol as such. Since this behaviour may not always be intended, it should be added as optional feature, e.g. as a new flag to `spvtools::LinkerOptions`.

pierremoreau:
I can have a look at allowing partially linked modules, when given the corresponding linker option.

pierremoreau:
Sorry for the wait. #1297 should allow you to do what you want. Please have a try and tell me if some things are not working as expected.

doe300:
Yes, that is exactly what I wanted and works like a charm. Thank you very much

PR #1297, opened 2018-02-13
linker: Allow modules to be partially linked

pierremoreau:
Fixes: https://github.com/KhronosGroup/SPIRV-Tools/issues/1144

### DIFF
--- a/include/spirv-tools/linker.hpp
+++ b/include/spirv-tools/linker.hpp
@@ -26,7 +26,10 @@ namespace spvtools {
 
 class LinkerOptions {
  public:
-  LinkerOptions() : create_library_(false), verify_ids_(false) {}
+  LinkerOptions()
+      : create_library_(false),
+        verify_ids_(false),
+        allow_partial_linkage_(false) {}
 
   // Returns whether a library or an executable should be produced by the
   // linking phase.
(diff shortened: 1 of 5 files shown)


pierremoreau:
The RelWithDebInfo build using g++ timed out, but all the other builds were successful.

antiagainst:
Hmm... It seems the performance of the Travis infrastructure is fluctuating again.

I restarted that job. Let's see what will happen.

pierremoreau:
Thanks for restarting the job. I had some timeout on another pull request, but I need to fix some formatting, so that will restart the job there.

pierremoreau:
Still the same build failing, but this time it is due to one of the unit test (spirv-tools-test_spirv_unit_tests), however that test is also failing on master.

dneto0:
I suspect the conflict is a trivial conflict with the --verify-ids change.  I can resolve that locally.

pierremoreau:
I would guess so as well; I initially had this series on top of the verify-ids one, and rebasing it onto master led to some conflicts, but I thought I was still better to have them separate.
Thanks for the review.

dneto0:
Rebased, resolved conflict, and pushed into master as bdd6617fa

Issue #1218, opened 2018-01-18
Linker should handle OpModuleProcessed


dneto0:
OpModuleProcessed instructions (introduced in SPIR-V 1.1) are accessed as ir::Module::debugs3().
That was added while the linker was being reviewed.

pierremoreau:
I’ll take it.

dneto0:
thanks. :-)

PR #1303, opened 2018-02-14
linker: merge debug annotations from category c)

pierremoreau:
Fixes: https://github.com/KhronosGroup/SPIRV-Tools/issues/1218

### DIFF
--- a/source/link/linker.cpp
+++ b/source/link/linker.cpp
@@ -446,6 +446,24 @@ static spv_result_t MergeModules(const MessageConsumer& consumer,
           std::unique_ptr<Instruction>(inst.Clone(linked_context)));
 
   for (const auto& module : input_modules)
+    for (const auto& inst : module->debugs3())
+      linked_module->AddDebug3Inst(
+          std::unique_ptr<Instruction>(inst.Clone(linked_context)));
+
+  // If the generated module uses SPIR-V 1.1 or higher, add an
+  // OpModuleProcessed instruction about the linking step.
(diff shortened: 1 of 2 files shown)


antiagainst:
Hi @pierremoreau, we have fixed the failure via #1319. You may want to rebase to trigger new builds.

pierremoreau:
Thanks for the notice @antiagainst ; let’s see if it succeeds now.

source/link/linker.cpp
```diff
@@ -445,6 +445,24 @@ static spv_result_t MergeModules(const MessageConsumer& consumer,
       linked_module->AddDebug2Inst(
           std::unique_ptr<Instruction>(inst.Clone(linked_context)));
 
+  for (const auto& module : input_modules)
+    for (const auto& inst : module->debugs3())
+      linked_module->AddDebug3Inst(
+          std::unique_ptr<Instruction>(inst.Clone(linked_context)));
+
+  // If the generated module uses SPIR-V 1.1 or higher, add an
```
dneto0: This is a nice touch.  :-)

dneto0:
Rebased, test fixed, and pushed into master as 9394272c

source/link/linker.cpp
pierremoreau: :-) I didn’t have it at first, but then decided I might as well make use of it.

pierremoreau:
Thanks for the review and fixing the partial-linkage test.

PR #1305, opened 2018-02-14
fix typo

nerijus:


dneto0:
Rebased and pushed onto master as 5d442fad2f

Issue #882, opened 2017-10-16
Disassembler rounds float constants that are near-to-1 up to 1


dneto0:
Example, a float value of 0x1.fffffep-1 is very nearly 1 but is not 1.
It should be displayed in a bit-preserving way, but instead is rounded up to 1.

So we get 
  OpConstant %float 1

when we should have had
   OpConstant %float 0x1.fffffep-1

dneto0:
I think people commonly want the default floating point string that their platform C++ library provides.  We should keep that as the default.

So I think we should add a disassembler option to force hex floats, for all float constants in the module. That gives us bit-for-bit preservation for floats through disassembler -> assembler round trip.

b-sumner:
It seems odd to me to make exact preservation of floating point constants through disassemble -> assemble just an option.  I'd say it is a requirement.  I don't really see any harm in outputting 0.1f as 0x1.99999ap-4.

dneto0:
> It seems odd to me to make exact preservation of floating point constants through disassemble -> assemble just an option. I'd say it is a requirement. I don't really see any harm in outputting 0.1f as 0x1.99999ap-4.

Most people aren't wired to read hex floating point very well. It's not the default for any other environment.  I think I'd have a lot of complaints if I flipped the default at this point.

Also, given variability across implementations, it's hard to identify "the" subset of cases that would benefit from bit-to-bit accuracy of hex floats.

b-sumner:
Should the disassembler issue warnings when it outputs floating point values that will not assemble to the same bits?

dneto0:
This appears to be fixed by #1306.

This input (from original report)
```
%float = OpTypeFloat 32
%near1 = OpConstant %float   0x1.fffffep-1
```
Assembles to a binary that disassembles to this:

```
; SPIR-V
; Version: 1.3
; Generator: Khronos SPIR-V Tools Assembler; 0
; Bound: 3
; Schema: 0
      %float = OpTypeFloat 32
%float_0_99999994 = OpConstant %float 0.99999994
```

That then re-assembles to a binary with same binary contents as the original (byte for byte equality).

So I think the problem is resolved.

PR #1306, opened 2018-02-14
spriv-dis: Use max_digits10 instead of digits10 for float output precision

greg-lunarg:
max_digits10 is needed for binary->text->binary, essentially, when one wants to disassemble and (re)assemble. digits10 is not sufficient and was causing lossiness for this process. One only need try the float32 value 16777215 which should be exactly representable but was getting truncated by spirv-dis. A test has been added for this value. A number of other tests needed to be "fixed" to take this new output precision into account.

### DIFF
--- a/test/binary_to_text_test.cpp
+++ b/test/binary_to_text_test.cpp
@@ -255,13 +255,14 @@ INSTANTIATE_TEST_CASE_P(
                 "%1 = OpTypeFloat 16\n%2 = OpConstant %1 0x1.ff4p+16\n",
                 "%1 = OpTypeFloat 16\n%2 = OpConstant %1 -0x1.d2cp-10\n",
                 // 32-bit floats
-                "%1 = OpTypeFloat 32\n%2 = OpConstant %1 -3.275\n",
+                "%1 = OpTypeFloat 32\n%2 = OpConstant %1 -3.2750001\n",
+                "%1 = OpTypeFloat 32\n%2 = OpConstant %1 16777215\n",
                 "%1 = OpTypeFloat 32\n%2 = OpConstant %1 0x1.8p+128\n", // NaN
                 "%1 = OpTypeFloat 32\n%2 = OpConstant %1 -0x1.0002p+128\n", // NaN
                 "%1 = OpTypeFloat 32\n%2 = OpConstant %1 0x1p+128\n", // Inf
                 "%1 = OpTypeFloat 32\n%2 = OpConstant %1 -0x1p+128\n", // -Inf
                 // 64-bit floats
-                "%1 = OpTypeFloat 64\n%2 = OpConstant %1 -3.275\n",
+                "%1 = OpTypeFloat 64\n%2 = OpConstant %1 -3.2749999999999999\n",
                 "%1 = OpTypeFloat 64\n%2 = OpConstant %1 0x1.ffffffffffffap-1023\n", // small normal
                 "%1 = OpTypeFloat 64\n%2 = OpConstant %1 -0x1.ffffffffffffap-1023\n",
                 "%1 = OpTypeFloat 64\n%2 = OpConstant %1 0x1.8p+1024\n", // NaN
--- a/test/name_mapper_test.cpp
+++ b/test/name_mapper_test.cpp
@@ -311,7 +311,8 @@ INSTANTIATE_TEST_CASE_P(
         {"%1 = OpTypeFloat 16\n%2 = OpConstant %1 -0x1.d2cp-10", 2,
          "half_n0x1_d2cpn10"},
         // 32-bit floats
-        {"%1 = OpTypeFloat 32\n%2 = OpConstant %1 -3.275", 2, "float_n3_275"},
+        {"%1 = OpTypeFloat 32\n%2 = OpConstant %1 -3.275", 2,
+         "float_n3_2750001"},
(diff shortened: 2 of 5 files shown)


dneto0:
I'd much rather resort to forcing hex-floats for cases you care about.
See https://github.com/KhronosGroup/SPIRV-Tools/issues/882 

The parser supports hex-float for bit-to-bit accurate recovery of a float number.  The disassembler doesn't support that yet but it would be relatively easy.

test/binary_to_text_test.cpp
```diff
@@ -255,13 +255,14 @@ INSTANTIATE_TEST_CASE_P(
...
                 // 32-bit floats
-                "%1 = OpTypeFloat 32\n%2 = OpConstant %1 -3.275\n",
+                "%1 = OpTypeFloat 32\n%2 = OpConstant %1 -3.2750001\n",
+                "%1 = OpTypeFloat 32\n%2 = OpConstant %1 16777215\n",
                 "%1 = OpTypeFloat 32\n%2 = OpConstant %1 0x1.8p+128\n", // NaN
                 "%1 = OpTypeFloat 32\n%2 = OpConstant %1 -0x1.0002p+128\n", // NaN
                 "%1 = OpTypeFloat 32\n%2 = OpConstant %1 0x1p+128\n", // Inf
                 "%1 = OpTypeFloat 32\n%2 = OpConstant %1 -0x1p+128\n", // -Inf
                 // 64-bit floats
-                "%1 = OpTypeFloat 64\n%2 = OpConstant %1 -3.275\n",
```
dneto0: For tests that care about some float number but not a particular one, I usually go for something that is an integer when multiplied by a power-of-2.  This one should have bee .375  (i.e. 3/8)

test/name_mapper_test.cpp
```diff
@@ -311,7 +311,8 @@ INSTANTIATE_TEST_CASE_P(
         {"%1 = OpTypeFloat 16\n%2 = OpConstant %1 -0x1.d2cp-10", 2,
          "half_n0x1_d2cpn10"},
         // 32-bit floats
-        {"%1 = OpTypeFloat 32\n%2 = OpConstant %1 -3.275", 2, "float_n3_275"},
```
dneto0: Same here. This should have been -3.375 not -3.275.

test/name_mapper_test.cpp
```diff
@@ -321,7 +322,8 @@ INSTANTIATE_TEST_CASE_P(
         {"%1 = OpTypeFloat 32\n%2 = OpConstant %1 -0x1p+128", 2,
          "float_n0x1p_128"},  // -Inf
                               // 64-bit floats
-        {"%1 = OpTypeFloat 64\n%2 = OpConstant %1 -3.275", 2, "double_n3_275"},
```
dneto0: Same

dneto0:
Are you trying to fix a portability of the tests or a "real" use case?
For absolute fidelity I'd rather implement hex-float option for disassembly #882.

dneto0:
For what it's worth, reliable base-10 printing and parsing of floating point numbers is a subject of ongoing research.
E.e. http://cseweb.ucsd.edu/~lerner/papers/fp-printing-popl16.pdf 
I don't expect implementations to fall in line any time soon.  Hex floats are the way to go here.

greg-lunarg:
While I was looking at floating point rounding issues in constant folding, I noticed that spirv-dis was dropping 2-3 digits of the significand of floats. I thought it strange that we were worrying about rounding mode accuracy while at the same time dropping 2-3 digits in spirv-dis. While graphics is probably not as picky about precision as HPC, it seemed that this would be something that HPC users would eventually bump into and care about so I was trying to be proactive. I have always thought that the ability to disassemble, edit and reassemble SPIR-V was a real strength of our ecosystem, and I thought it would be a good idea to not be causing inadvertant significant changes if possible.

I looked into this and it seemed fairly obvious to me that max_digits10 was a better precision here than digits10.

I don't doubt that hex-float is the way to go for high fidelity. But I thought if we could fix 99.999% of the problems with a one word change, that might be a good thing until something "better" could be settled on and implemented.

greg-lunarg:
Well, I should have anticipated that this change would hit other tests as well :(

@dneto Is the change to hex-float a done deal? Is it imminent? If yes to both, than it probably doesn't make sense spending time fixing these tests. Otherwise, maybe.

greg-lunarg:
@dneto OK. So it seems your thinking is to only do hex-float as an option.

Are you interested in a less-lossy default ie this change? If so, I will continue fixing this PR. If not, we can just drop it. Let me know.

dneto0:
Finally getting back to this.

> I don't doubt that hex-float is the way to go for high fidelity. But I thought if we could fix 99.999% of the problems with a one word change, that might be a good thing until something "better" could be settled on and implemented.

I'll by the "why not both" argument.  If you want to continue down this path then go for it. The hex-float option is not imminent.

When there are spurious-looking changes, e.g. when we have a random float value that isn't testing a specific limit, I suggest fixing the constants to be more binary-friendly. (e.g. .375 instead of .275 ).

dneto0:
I'm not sure whether you saw the extra failures found by the bots.  Is there a platform sensitivity? That would be concerning.

dneto0:
Superceded by #1444

antiagainst:
#1444 is merged. I'll close this one.

PR #1308, opened 2018-02-14
Invoke cmake via CMAKE_COMMAND variable

dneto0:
Need to do this in case cmake is not on the path.
This should fix the Android NDK build, as in when building the NDK
itself.


dneto0:
> Ha, so this is the culprit? Can never assume anything for cross compilation related stuff...

Yes.  It's a machine-configuration difference. Cmake is not on the path on the Android NDK builders.  Another risk is that the builders use the oldest-supported Apple toolchains.  But my local OSX testing machines have the latest tools.  So far that hasn't been a problem, but it's a risk.

dneto0:
Rebased and fixed the tabs/spaces problem, and retested. Thanks!

PR #1309, opened 2018-02-15
Add folding of redundant OpSelect insns

zeux:
We can fold OpSelect into one of the operands in two cases:

- condition is constant
- both results are the same

Even if the original shader doesn't have either of these, if-conversion
pass sometimes ends up generating instructions like

   %7127 = OpSelect %int %3220 %7058 %7058

And this optimization cleans them up.

### DIFF
--- a/source/opt/folding_rules.cpp
+++ b/source/opt/folding_rules.cpp
@@ -294,6 +294,38 @@ FoldingRule RedundantPhi() {
         return true;
       };
 }
+
+FoldingRule RedundantSelect() {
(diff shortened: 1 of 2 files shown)


zeux:
On a separate note - should instruction simplification pass be part of size/performance passes? It's currently only present in the list of legalization passes.

s-perron:
Yes, we will be adding it to the performance passes soon.  I was want to put things in a little bit at a time.  It seems to be working well in the legalization passes, so the performance passes will come soon.

PR #1310, opened 2018-02-15
Add constant folding rules for floating-point comparison

zeux:
This change handles all 6 regular comparison types in two variations,
ordered (true if values are ordered *and* comparison is true) and
unordered (true if values are unordered *or* comparison is true).

Ordered comparison matches the default floating-point behavior on host
but we use std::isnan to check ordering explicitly anyway.

This change also slightly reworks the floating-point folding support
code to make it possible to define a folding operation that returns
boolean instead of floating point.

### DIFF
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -281,14 +281,14 @@ bool CompareFloatingPoint(bool op_result, bool op_unordered,
       float fb = GetFloatFromConst(b);                                    \
       bool result = CompareFloatingPoint(                                 \
           fa op fb, std::isnan(fa) || std::isnan(fb), ord);               \
-      std::vector<uint32_t> words = {result};                             \
+      std::vector<uint32_t> words = {uint32_t(result)};                   \
       return const_mgr->GetConstant(result_type, words);                  \
     } else if (float_type->width() == 64) {                               \
       double fa = GetDoubleFromConst(a);                                  \
       double fb = GetDoubleFromConst(b);                                  \
       bool result = CompareFloatingPoint(                                 \
           fa op fb, std::isnan(fa) || std::isnan(fb), ord);               \
-      std::vector<uint32_t> words = {result};                             \
+      std::vector<uint32_t> words = {uint32_t(result)};                   \
       return const_mgr->GetConstant(result_type, words);                  \
     }                                                                     \
     return nullptr;                                                       \
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -156,18 +156,18 @@ ConstantFoldingRule FoldFloatingPointOp(FloatScalarFoldingRule scalar_rule) {
     }
 
     if (vector_type != nullptr) {
-      std::vector<const analysis::Constant*> a_componenets;
-      std::vector<const analysis::Constant*> b_componenets;
+      std::vector<const analysis::Constant*> a_components;
+      std::vector<const analysis::Constant*> b_components;
       std::vector<const analysis::Constant*> results_components;
 
-      a_componenets = GetVectorComponents(constants[0], const_mgr);
-      b_componenets = GetVectorComponents(constants[1], const_mgr);
+      a_components = GetVectorComponents(constants[0], const_mgr);
+      b_components = GetVectorComponents(constants[1], const_mgr);
 
       // Fold each component of the vector.
-      for (uint32_t i = 0; i < a_componenets.size(); ++i) {
+      for (uint32_t i = 0; i < a_components.size(); ++i) {
         results_components.push_back(scalar_rule(vector_type->element_type(),
-                                                 a_componenets[i],
-                                                 b_componenets[i], const_mgr));
+                                                 a_components[i],
+                                                 b_components[i], const_mgr));
         if (results_components[i] == nullptr) {
           return nullptr;
         }
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -105,8 +105,8 @@ ConstantFoldingRule FoldCompositeWithConstants() {
 // The interface for a function that returns the result of applying a scalar
 // floating-point binary operation on |a| and |b|.  The type of the return value
 // will be |type|.  The input constants must also be of type |type|.
-using FloatScalarFoldingRule = std::function<const analysis::FloatConstant*(
-    const analysis::Float* type, const analysis::Constant* a,
+using FloatScalarFoldingRule = std::function<const analysis::Constant*(
+    const analysis::Type* result_type, const analysis::Constant* a,
     const analysis::Constant* b, analysis::ConstantManager*)>;
 
 // Returns an std::vector containing the elements of |constant|.  The type of
@@ -146,7 +146,6 @@ ConstantFoldingRule FoldFloatingPointOp(FloatScalarFoldingRule scalar_rule) {
     analysis::TypeManager* type_mgr = context->get_type_mgr();
     const analysis::Type* result_type = type_mgr->GetType(inst->type_id());
     const analysis::Vector* vector_type = result_type->AsVector();
-    const analysis::Float* float_type = nullptr;
 
     if (!CanFoldFloatingPoint(context, inst->result_id())) {
       return nullptr;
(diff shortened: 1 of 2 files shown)


zeux:
Please let me know if I went overboard on tests or if you'd like to see some tests that are currently missing...

s-perron:
Thanks for adding these rules.  I know you were asking for them.  There are lots to add, and in the shaders I am working on other things come up more often.

zeux:
Added more tests. And no worries; these were easier to add now that previous folding changes established a framework :)

zeux:
Hmm... This change might be triggering an issue in one of our shaders. I previously tested this change in isolation, for example:

    spirv-opt -O shader1.spv -o shader2.spv
    spirv-opt --simplify-instructions shader2.spv -o shader3.spv

and

    spirv-opt --simplify-instructions shader1.spv -o shader4.spv

And these all work fine. However, trying to run a combined pass sequence like this:

    spirv-opt -O --simplify-instructions shader1.spv -o shader5.spv

Crashes inside the simplification pass with the following callstack when simplifying FOrdEqual:

```
#0  0x00000000007d3690 in spvtools::ir::Instruction::result_id (this=0x0) at /mnt/c/work/SPIRV-Tools/source/opt/instruction.h:154
#1  0x000000000086bfff in spvtools::opt::analysis::DefUseManager::UpdateDefUse (this=0xd46810, inst=0x0) at /mnt/c/work/SPIRV-Tools/source/opt/def_use_manager.cpp:75
#2  0x000000000087b7e5 in spvtools::ir::IRContext::UpdateDefUse (this=0xd53e10, inst=0x0) at /mnt/c/work/SPIRV-Tools/source/opt/ir_context.h:754
#3  0x000000000087ab6a in spvtools::opt::FoldInstructionToConstant(spvtools::ir::Instruction*, std::function<unsigned int (unsigned int)>) (inst=0xdafb60, id_map=...)
    at /mnt/c/work/SPIRV-Tools/source/opt/fold.cpp:643
#4  0x000000000087940e in spvtools::opt::(anonymous namespace)::FoldInstructionInternal (inst=0xdafb60) at /mnt/c/work/SPIRV-Tools/source/opt/fold.cpp:194
#5  0x000000000087ae37 in spvtools::opt::FoldInstruction (inst=0xdafb60) at /mnt/c/work/SPIRV-Tools/source/opt/fold.cpp:686
#6  0x00000000007f55d8 in spvtools::opt::SimplificationPass::<lambda(spvtools::ir::BasicBlock*)>::operator()(spvtools::ir::BasicBlock *) const (__closure=0xd27f60, bb=0xdd0ed0)
    at /mnt/c/work/SPIRV-Tools/source/opt/simplification_pass.cpp:59
```

What looks like is happening here is that the folding folds comparison to a bool constant:

```
(gdb) p *(spvtools::opt::analysis::BoolConstant*)folded_const
$5 = {<spvtools::opt::analysis::ScalarConstant> = {<spvtools::opt::analysis::Constant> = {_vptr.Constant = 0xcfc3f8 <vtable for spvtools::opt::analysis::BoolConstant+16>,
      type_ = 0xd60bc0}, words_ = std::vector of length 1, capacity 1 = {0}}, value_ = false}
```

However, the constant does *not* have a defining instruction, so const_inst is nullptr and UpdateDefUse crashes. I'm not sure how to handle this correctly - a simple null check & early return doesn't seem right?

s-perron:
Can you give a small example that shows the problem?  `const_inst` is the return value of `GetDefiningInstruction`, which is suppose to create the instruction if it does not already exist.

We should add a check for a nullprt, because that can happen.  However, I don't think your case should get a nullptr.

zeux:
The issue with a nullptr check is that if you have to return an Instruction, but if you just return nullptr the folding doesn't happen; we could synthesize a new instruction for the constant but that seems a bit like duct tape.

The issue is that constant manager thinks that the constant is registered and bound to an instruction (FindDeclaredConstant returns non-0 id), but def manager thinks that the id doesn't have a def anymore. I suspect that this is happening because the constant was registered with constant manager earlier, but the definition was removed by some other pass - so constant manager would still have the const->id mapping but the id is dead. Trying to make a small repro now.

zeux:
Yeah - my suspicion was correct; in this shader, I get this at some point during optimization:

...
%1158 = OpConstantFalse %538
...

which gets recorded in the constant-manager; at a later time, during simplify-instructions this:

```
%76 = OpConstant %6 0
...
%674 = OpConstant %6 0.75
...
%1251 = OpFOrdEqual %538 %674 %76
```

is folded to `false` which refers to 1158 in the constant manager, but doesn't appear in the IR that is fed to simplification pass.

zeux:
So I still need to make a simple repro for this, but as far as what we could do, I can imagine several options:

- Fix the problem in GetDefiningInstruction:

```
diff --git a/source/opt/constants.cpp b/source/opt/constants.cpp
index 336e061..d899561 100644
--- a/source/opt/constants.cpp
+++ b/source/opt/constants.cpp
@@ -67,12 +67,18 @@ ir::Instruction* ConstantManager::BuildInstructionAndAddToModule(
 ir::Instruction* ConstantManager::GetDefiningInstruction(
     const Constant* c, ir::Module::inst_iterator* pos) {
   uint32_t decl_id = FindDeclaredConstant(c);
-  if (decl_id == 0) {
+
+  // The mapping we recorded can become invalid if the constant has been removed
+  // by DCE If this happened, we recreate the constant instruction
+  ir::Instruction* def =
+      (decl_id == 0) ? nullptr : context()->get_def_use_mgr()->GetDef(decl_id);
+
+  if (def == nullptr) {
     auto iter = context()->types_values_end();
     if (pos == nullptr) pos = &iter;
     return BuildInstructionAndAddToModule(c, pos);
   } else {
-    return context()->get_def_use_mgr()->GetDef(decl_id);
+    return def;
   }
 }
```

This will break in odd ways if ids are ever reused during optimization, but otherwise might be fine.

- In PassManager, clean up dead constants from constant manager after each pass

- Remove constants from constant manager in DCE passes - I think there's more than one so all of them would need to change

s-perron:
I want to look at `IRContext::KillInst()`.  If it is deleting a constant it should update the constant manager.

zeux:
Ah yeah, that's better. This appears to work:

```
diff --git a/source/opt/constants.cpp b/source/opt/constants.cpp
index 336e061..d5b873f 100644
--- a/source/opt/constants.cpp
+++ b/source/opt/constants.cpp
@@ -72,7 +72,9 @@ ir::Instruction* ConstantManager::GetDefiningInstruction(
     if (pos == nullptr) pos = &iter;
     return BuildInstructionAndAddToModule(c, pos);
   } else {
-    return context()->get_def_use_mgr()->GetDef(decl_id);
+    auto def = context()->get_def_use_mgr()->GetDef(decl_id);
+    assert(def != nullptr);
+    return def;
   }
 }
 
diff --git a/source/opt/constants.h b/source/opt/constants.h
index e382e16..6ac9bef 100644
--- a/source/opt/constants.h
+++ b/source/opt/constants.h
@@ -518,6 +518,14 @@ class ConstantManager {
     return false;
   }
 
+  void RemoveId(uint32_t id) {
+    auto it = id_to_const_val_.find(id);
+    if (it != id_to_const_val_.end()) {
+      const_val_to_id_.erase(it->second);
+      id_to_const_val_.erase(it);
+    }
+  }
+
   // Records a new mapping between |inst| and |const_value|. This updates the
   // two mappings |id_to_const_val_| and |const_val_to_id_|.
   void MapConstantToInst(const Constant* const_value, ir::Instruction* inst) {
diff --git a/source/opt/ir_context.cpp b/source/opt/ir_context.cpp
index 895c8a9..a5efd80 100644
--- a/source/opt/ir_context.cpp
+++ b/source/opt/ir_context.cpp
@@ -106,6 +106,10 @@ Instruction* IRContext::KillInst(ir::Instruction* inst) {
     type_mgr_->RemoveId(inst->result_id());
   }
 
+  if (constant_mgr_ && inst->IsConstant()) {
+    constant_mgr_->RemoveId(inst->result_id());
+  }
+
   RemoveFromIdToName(inst);
 
   Instruction* next_instruction = nullptr;
```

s-perron:
Yes that is what we want.  Can you open a new PR with that fix?  We can merge that first, and then we will merge this one.

s-perron:
@zeux Same here.  Please rebase.

s-perron:
rebased, squashed, and merged as 10544136.

PR #1313, opened 2018-02-15
Make better use of simplification pass

### LINKED ISSUE #1118: Update the legalization passes

s-perron:
The simplification pass works better after all of the dead branches are
removed.  So swapping them around in the legalization passes.  Also
adding the simplification pass to performance passes right after dead
branch elimination.

Added CCP to the legalization passes so we can propagate the constants
into the branchs, and remove as many branches a possible.  CCP is
designed to still get opportunities even if the branches are dead, so it
is a good place for it.

Fixes #1118


s-perron:
Only test failure was a time out because of travis being slow.  Tested myself, and it looks okay.  Merging anyway.

PR #1315, opened 2018-02-15
Simplify OpPhi instructions referencing unreachable continues

### LINKED ISSUE #1314: Simplify OpPhi instructions referencing unreachable continues.
I've seen lots of OpPhi instructions at the start of a loop that has an unreachable continue.  If this OpPhi has only 1 other entry, then there is only 1 possible value for the OpPhi.  Here is one example:

```
OpBranch %2
%2 = OpLabel
%phi = OpPhi %bool %false %1 %true %continue
OpBranch %3
%3 = OpLabel
OpReturn
%continue = OpLabel
OpBranch %2
%merge = OpLabel
OpReturn
```

The %phi will always be %false because it is not possible to come along the other edge.  We should be able to replace the uses of the %phi with %false.

s-perron:
In dead branch elimination, we already recognize unreachable continue
blocks, and update OpPhi instruction accordingly.  This change adds an
extra check: if the head block has exactly 1 other incoming edge, then
replace the OpPhi with the value from that edge.

Fixes #1314.

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -211,12 +211,17 @@ bool DeadBranchElimPass::FixPhiNodesInLiveBlocks(
         operands.push_back(inst->GetOperand(0u));
         operands.push_back(inst->GetOperand(1u));
         // Iterate through the incoming labels and determine which to keep
-        // and/or modify.
+        // and/or modify.  If there in an unreachable continue block, there will
+        // be an edge from that block to the header.  We need to keep it to
+        // maintain the structured control flow.  If the header has more that 2
+        // incoming edges, then the OpPhi must have an entry for that edge.
+        // However, if there is only one other incoming edge, the OpPhi can be
+        // eliminated.
         for (uint32_t i = 1; i < inst->NumInOperands(); i += 2) {
           ir::BasicBlock* inc = GetParentBlock(inst->GetSingleWordInOperand(i));
           auto cont_iter = unreachable_continues.find(inc);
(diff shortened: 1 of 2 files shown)


source/opt/dead_branch_elim_pass.cpp
```diff
@@ -250,7 +255,8 @@ bool DeadBranchElimPass::FixPhiNodesInLiveBlocks(
           modified = true;
           uint32_t continue_id = block.ContinueBlockIdIfAny();
           if (!backedge_added && continue_id != 0 &&
-              unreachable_continues.count(GetParentBlock(continue_id))) {
+              unreachable_continues.count(GetParentBlock(continue_id)) &&
+              operands.size() > 4) {
```
alan-baker: I think this should be >=. In this case we haven't added a backedge (because the backedge block was the successor of an unreachable continue). So we have two valid incoming edges and would have a third from the unreachable continue.

Apparently no test covers this properly. Also please double check the subsequent code which special cases a two input phi node.

s-perron: In `operands`, the first two entries are the result id, and the result type.  If the first incoming egdse covers elements 3 and 4.  If there are move than 4, it is because there is a second incoming edge.

Do you want me to add a comment in the code?

alan-baker: Ok, clearly forgot how my own code worked.

s-perron:
The only failure in the bots was a time out for one of the builds.  I'll merge anyway.

PR #1316, opened 2018-02-15
Adding checks for bad CCP transitions and unsettled values

### LINKED ISSUE #1300: Identify and fix bad lattice transitions in CCP
#1295 found a bug in CCP where a phi went from interesting to not interesting. This is an invalid lattice transition. We should assert all transitions are valid during the transformation. Adding one such assert broke `CCPTest.SSAWebCycles`, so it appears there are more bugs.

alan-baker:
Fixes #1300

* Now track propagation status and assert on bad statuses
 * Added helper methods to access instruction propagation status
* Modified the phi meet operator to properly reflect the paper it is
based on
* Modified SSA edge addition so that all edge are added, but only on
state changes
* Fixed a bug in instruction simulation where interesting conditional
branches would not mark the interesting edge as executed
 * Added a test to catch this bug
* Added an ostream operator for SSAPropagator::PropStatus

### DIFF
--- a/source/opt/ccp_pass.cpp
+++ b/source/opt/ccp_pass.cpp
@@ -81,9 +81,10 @@ SSAPropagator::PropStatus CCPPass::VisitPhi(ir::Instruction* phi) {
         return MarkInstructionVarying(phi);
       }
     } else {
-      // If any argument is not a constant, the Phi produces nothing
-      // interesting for now. The propagator will callback again, if needed.
-      return SSAPropagator::kNotInteresting;
+      // The incoming value has no recorded value and is therefore not
+      // interesting. A not interesting value joined with any other value is the
+      // other value.
+      continue;
     }
   }
 
@@ -227,15 +228,17 @@ SSAPropagator::PropStatus CCPPass::VisitBranch(ir::Instruction* instr,
 
 SSAPropagator::PropStatus CCPPass::VisitInstruction(ir::Instruction* instr,
                                                     ir::BasicBlock** dest_bb) {
+  SSAPropagator::PropStatus status = SSAPropagator::kVarying;
   *dest_bb = nullptr;
   if (instr->opcode() == SpvOpPhi) {
-    return VisitPhi(instr);
+    status = VisitPhi(instr);
   } else if (instr->IsBranch()) {
-    return VisitBranch(instr, dest_bb);
+    status = VisitBranch(instr, dest_bb);
   } else if (instr->result_id()) {
-    return VisitAssignment(instr);
+    status = VisitAssignment(instr);
   }
-  return SSAPropagator::kVarying;
+  // return SSAPropagator::kVarying;
+  return status;
 }
 
 bool CCPPass::ReplaceValues() {
@@ -258,7 +261,19 @@ bool CCPPass::PropagateConstants(ir::Function* fp) {
 
   propagator_ =
       std::unique_ptr<SSAPropagator>(new SSAPropagator(context(), visit_fn));
-  if (propagator_->Run(fp)) {
+  bool modified = propagator_->Run(fp);
+
+#ifndef NDEBUG
+  // Verify all visited values have settled. No value that has been simulated
+  // should end on not interesting.
+  fp->ForEachInst([this](ir::Instruction* inst) {
+    assert(!propagator_->HasStatus(inst) ||
+           propagator_->Status(inst) != SSAPropagator::kNotInteresting &&
+               "Unsettled value");
+  });
+#endif
+
+  if (modified) {
     return ReplaceValues();
   }
 
--- a/source/opt/ccp_pass.cpp
+++ b/source/opt/ccp_pass.cpp
@@ -228,17 +228,15 @@ SSAPropagator::PropStatus CCPPass::VisitBranch(ir::Instruction* instr,
 
 SSAPropagator::PropStatus CCPPass::VisitInstruction(ir::Instruction* instr,
                                                     ir::BasicBlock** dest_bb) {
-  SSAPropagator::PropStatus status = SSAPropagator::kVarying;
   *dest_bb = nullptr;
   if (instr->opcode() == SpvOpPhi) {
-    status = VisitPhi(instr);
+    return VisitPhi(instr);
   } else if (instr->IsBranch()) {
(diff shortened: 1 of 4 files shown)


source/opt/ccp_pass.cpp
```diff
@@ -258,7 +259,19 @@ bool CCPPass::PropagateConstants(ir::Function* fp) {
...
+
+#ifndef NDEBUG
+  // Verify all visited values have settled. No value that has been simulated
+  // should end on not interesting.
+  fp->ForEachInst([this](ir::Instruction* inst) {
+    assert(!propagator_->HasStatus(inst) ||
+           propagator_->Status(inst) != SSAPropagator::kNotInteresting &&
+               "Unsettled value");
+  });
+#endif
```
s-perron: Not that is has to be done, but could we make this check part of the generic propagator.  That way we automatically get the checks when someone else tried to use it.

s-perron:
Look like you have a real compile error with gcc:

```
In file included from /usr/include/c++/4.8/cassert:43:0,
                 from /home/travis/build/KhronosGroup/SPIRV-Tools/source/util/hex_float.h:18,
                 from /home/travis/build/KhronosGroup/SPIRV-Tools/source/opt/constants.h:29,
                 from /home/travis/build/KhronosGroup/SPIRV-Tools/source/opt/ccp_pass.h:18,
                 from /home/travis/build/KhronosGroup/SPIRV-Tools/source/opt/ccp_pass.cpp:19:
/home/travis/build/KhronosGroup/SPIRV-Tools/source/opt/ccp_pass.cpp: In lambda function:
/home/travis/build/KhronosGroup/SPIRV-Tools/source/opt/ccp_pass.cpp:269:72: error: suggest parentheses around ‘&&’ within ‘||’ [-Werror=parentheses]
            propagator_->Status(inst) != SSAPropagator::kNotInteresting &&
```

source/opt/ccp_pass.cpp
alan-baker: Made this change when fixing the compile error.

dneto0:
Please rebase against master for a clean bot run.

test/opt/ccp_test.cpp
```diff
@@ -458,14 +458,15 @@ TEST_F(CCPTest, SSAWebCycles) {
          %22 = OpIAdd %int %29 %30
                OpBranch %14
          %14 = OpLabel
-; CHECK: OpPhi %int %int_0 {{%\d+}} %int_0 {{%\d+}}
-         %25 = OpPhi %int %int_0 %5 %30 %14
+; CHECK: OpPhi %int %int_0 {{%\d+}}
+         %25 = OpPhi %int %30 %12
```
dneto0: Yes, this must have been painful.

source/opt/propagator.cpp
```diff
@@ -36,22 +36,14 @@ void SSAPropagator::AddControlEdge(const Edge& edge) {
...
+void SSAPropagator::AddSSAEdges(ir::Instruction* instr) {
   // Ignore instructions that produce no result.
   if (instr->result_id() == 0) {
     return;
   }
 
   get_def_use_mgr()->ForEachUser(
-      instr->result_id(), [this, traverse_phis](ir::Instruction* use_instr) {
-        // If |use_instr| is a Phi, ignore this edge.  Phi instructions can form
-        // cycles in the def-use web, which would get the propagator into an
```
dnovillo: Oh, nice I always disliked this hack. However, how did your change guarantee that we don't get into an infinite loop anymore?

source/opt/propagator.h
```diff
@@ -198,6 +198,20 @@ class SSAPropagator {
...
+  bool HasStatus(ir::Instruction* inst) const { return statuses_.count(inst); }
+
+  // Returns the current propagation status of |inst|.
+  PropStatus Status(ir::Instruction* inst) const {
+    return statuses_.find(inst)->second;
+  }
+
+  // Records the propagation status |status| for |inst|.
+  void SetStatus(ir::Instruction* inst, PropStatus status) {
+    statuses_[inst] = status;
```
dnovillo: Move the checking for old status here, please. Better contained.

test/opt/ccp_test.cpp
dnovillo: Ugh. Sorry about this one.

source/opt/propagator.cpp
alan-baker: I tried both infinite loop cases linked through #1159 and neither hangs.

source/opt/propagator.h
alan-baker: Good idea, will do.

source/opt/propagator.cpp
s-perron: You will not get into an infinite loop because you only add the ssa edges if the status changed.  For each instruction, it status can change at most twice (not interesting -> interesting -> varying).  Then the simulations will stop.

dnovillo:
On Fri, Feb 16, 2018 at 12:58 PM, Steven Perron <notifications@github.com>
wrote:

>
> In source/opt/propagator.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1316#discussion_r168856248>
> :
>
> >    // Ignore instructions that produce no result.
>    if (instr->result_id() == 0) {
>      return;
>    }
>
>    get_def_use_mgr()->ForEachUser(
> -      instr->result_id(), [this, traverse_phis](ir::Instruction* use_instr) {
> -        // If |use_instr| is a Phi, ignore this edge.  Phi instructions can form
> -        // cycles in the def-use web, which would get the propagator into an
>
> You will not get into an infinite loop because you only add the ssa edges
> if the status changed. For each instruction, it status can change at most
> twice (not interesting -> interesting -> varying). Then the simulations
> will stop.
>

Right.  That's what's supposed to happen.  This was not happening before.
I think it's the fixed meet operation done for non-interesting results.
Instead of giving up and saying "not interesting", we now proceed with the
merge and reconsider on the second round.  Anyway, I think this is the
right fix and removing the previous hack is a good thing.

s-perron:
Still has a compiler error when compiling with GCC.

alan-baker:
Rebased, squashed and submitted as c3f34d8bf3532040592a8bbb7f8079bcb7ae4c6b

PR #1317, opened 2018-02-15
Remove constants from constant manager in KillInst

zeux:
Registering a constant in constant manager establishes a relation
between instruction that defined it and constant object. On complex
shaders this could result in the constant definition getting removed as
part of one of the DCE pass, and a subsequent simplification pass trying
to use the defining instruction for the constant.

To fix this, we now remove associated constant entries from constant
manager when killing constant instructions; the constant object is still
registered and can be remapped to a new instruction later.

GetDefiningInstruction shouldn't ever return nullptr after this change
so add an assertion to check for that.

### DIFF
--- a/source/opt/ir_context.cpp
+++ b/source/opt/ir_context.cpp
@@ -106,6 +106,10 @@ Instruction* IRContext::KillInst(ir::Instruction* inst) {
     type_mgr_->RemoveId(inst->result_id());
   }
 
+  if (constant_mgr_ && ir::IsConstantInst(inst->opcode())) {
+    constant_mgr_->RemoveId(inst->result_id());
+  }
+
   RemoveFromIdToName(inst);
 
   Instruction* next_instruction = nullptr;
--- a/source/opt/constants.cpp
+++ b/source/opt/constants.cpp
@@ -72,7 +72,9 @@ ir::Instruction* ConstantManager::GetDefiningInstruction(
     if (pos == nullptr) pos = &iter;
     return BuildInstructionAndAddToModule(c, pos);
   } else {
-    return context()->get_def_use_mgr()->GetDef(decl_id);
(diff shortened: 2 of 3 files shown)


dneto0:
Please rebase against master so we can see a clean bot run.  Bots were failing because of an external fix to capability dependencies, and our capability tests caught it.

source/opt/ir_context.cpp
```diff
@@ -106,6 +106,10 @@ Instruction* IRContext::KillInst(ir::Instruction* inst) {
     type_mgr_->RemoveId(inst->result_id());
   }
 
+  if (constant_mgr_ && inst->IsConstant()) {
```
alan-baker: The constant manager can generate a constant from a OpSpecConstantComposite (at least in some cases). This check will not identify those. Perhaps the cleanest way would be to add a method to constant manager to indicate supported constant opcodes? It would return true for IsConstant() and OpSpecConstantComposite.

zeux: Restricting just to OpSpecConstantComposite seems slightly odd - would it make sense to use ir::IsConstantInst(inst->opcode()) instead?

alan-baker: The oddness is why I propose the method test method belong to ConstantManager. Then it can completely own the implementation weirdness; however, RemoveId looks safe for all ids so I'm also ok with your suggestion.

zeux:
Rebased against master, changed IsConstant() call to ir::IsConstantInst that detects both constants and spec constants (Instruction::IsConstant needs to be renamed IMO... but probably in a separate change)

zeux:
Although I guess IsConstant *does* say that it doesn't return true for spec constants in a comment, I just missed it.

s-perron:
rebase and merged at 27d23a92.

PR #1321, opened 2018-02-16
Support for multiple inductions in loop unroller

davors72:
Adds support to unrolls loops with multiple induction variables. The condition remains that the each loop must only be bound by one induction variable.

`for(int i = 0, j =2, k=0; i < 10; ++i, ++j, ++k)`
Would be unrolled, however:
`for(int i = 0, j =2, k=0; i < 10 && k < 10; ++i, ++j, ++k)`
Will not be unrolled.

### DIFF
--- a/source/opt/ir_builder.h
+++ b/source/opt/ir_builder.h
@@ -172,11 +172,20 @@ class InstructionBuilder {
     // Assert that we are not trying to store a negative number in an unsigned
     // type.
     if (!sign)
-      assert(value > 0 &&
+      assert(value >= 0 &&
              "Trying to add a signed integer with an unsigned type!");
 
-    // Get or create the integer type.
-    analysis::Integer int_type(32, sign);
+    analysis::Integer int_type{32, sign};
+
+    // Get or create the integer type. This rebuilds the type and manages the
+    // memory for the rebuilt type.
+    uint32_t type_id =
+        GetContext()->get_type_mgr()->GetTypeInstruction(&int_type);
+
+    // Get the memory managed type so that it is safe to be stored by
+    // GetConstant.
+    analysis::Type* rebuilt_type =
+        GetContext()->get_type_mgr()->GetType(type_id);
 
     // Even if the value is negative we need to pass the bit pattern as a
     // uint32_t to GetConstant.
@@ -184,7 +193,7 @@ class InstructionBuilder {
 
     // Create the constant value.
     const opt::analysis::Constant* constant =
-        GetContext()->get_constant_mgr()->GetConstant(&int_type, {word});
+        GetContext()->get_constant_mgr()->GetConstant(rebuilt_type, {word});
 
     // Create the OpConstant instruction using the type and the value.
     return GetContext()->get_constant_mgr()->GetDefiningInstruction(constant);
--- a/source/opt/loop_descriptor.cpp
+++ b/source/opt/loop_descriptor.cpp
@@ -33,7 +33,7 @@ namespace ir {
 // Takes in a phi instruction |induction| and the loop |header| and returns the
 // step operation of the loop.
 ir::Instruction* Loop::GetInductionStepOperation(
-    const ir::Loop* loop, const ir::Instruction* induction) const {
+    const ir::Instruction* induction) const {
   // Induction must be a phi instruction.
   assert(induction->opcode() == SpvOpPhi);
 
@@ -50,7 +50,7 @@ ir::Instruction* Loop::GetInductionStepOperation(
 
     // Check if the block is dominated by header, and thus coming from within
     // the loop.
-    if (loop->IsInsideLoop(incoming_block)) {
+    if (IsInsideLoop(incoming_block)) {
       step = def_use_manager->GetDef(
           induction->GetSingleWordInOperand(operand_id - 1));
       break;
@@ -58,6 +58,21 @@ ir::Instruction* Loop::GetInductionStepOperation(
   }
 
   if (!step || !IsSupportedStepOp(step->opcode())) {
+    return nullptr;
+  }
+
+  // The induction variable which binds the loop must only be modified once.
+  uint32_t lhs = step->GetSingleWordInOperand(0);
+  uint32_t rhs = step->GetSingleWordInOperand(1);
+
+  // One of the left hand side or right hand side of the step instruction must
+  // be the induction phi and the other must be an OpConstant.
+  if (lhs != induction->result_id() && rhs != induction->result_id()) {
+    return nullptr;
+  }
+
+  if (def_use_manager->GetDef(lhs)->opcode() != SpvOp::SpvOpConstant &&
+      def_use_manager->GetDef(rhs)->opcode() != SpvOp::SpvOpConstant) {
     return nullptr;
   }
 
@@ -84,17 +99,52 @@ bool Loop::IsSupportedCondition(SpvOp condition) const {
     // >
     case SpvOp::SpvOpUGreaterThan:
     case SpvOp::SpvOpSGreaterThan:
+
+    // >=
+    case SpvOp::SpvOpSGreaterThanEqual:
+    case SpvOp::SpvOpUGreaterThanEqual:
+    // <=
+    case SpvOp::SpvOpSLessThanEqual:
+    case SpvOp::SpvOpULessThanEqual:
+
       return true;
     default:
       return false;
   }
 }
 
+int64_t Loop::GetResidualConditionValue(SpvOp condition, int64_t initial_value,
+                                        int64_t step_value,
+                                        size_t number_of_iterations,
+                                        size_t factor) {
+  int64_t remainder =
+      initial_value + (number_of_iterations % factor) * step_value;
+
+  // We subtract or add one as the above formula calculates the remainder if the
+  // loop where just less than or greater than. Adding or subtracting one should
+  // give a functionally equivalent value.
+  switch (condition) {
+    case SpvOp::SpvOpSGreaterThanEqual:
+    case SpvOp::SpvOpUGreaterThanEqual: {
+      remainder -= 1;
+      break;
+    }
+    case SpvOp::SpvOpSLessThanEqual:
+    case SpvOp::SpvOpULessThanEqual: {
+      remainder += 1;
+      break;
+    }
+
+    default:
+      break;
+  }
+  return remainder;
+}
+
 // Extract the initial value from the |induction| OpPhi instruction and store it
 // in |value|. If the function couldn't find the initial value of |induction|
 // return false.
-bool Loop::GetInductionInitValue(const ir::Loop* loop,
-                                 const ir::Instruction* induction,
+bool Loop::GetInductionInitValue(const ir::Instruction* induction,
                                  int64_t* value) const {
   ir::Instruction* constant_instruction = nullptr;
   opt::analysis::DefUseManager* def_use_manager = context_->get_def_use_mgr();
@@ -104,7 +154,7 @@ bool Loop::GetInductionInitValue(const ir::Loop* loop,
     ir::BasicBlock* bb = context_->cfg()->block(
         induction->GetSingleWordInOperand(operand_id + 1));
 
-    if (!loop->IsInsideLoop(bb)) {
+    if (!IsInsideLoop(bb)) {
       constant_instruction = def_use_manager->GetDef(
           induction->GetSingleWordInOperand(operand_id));
     }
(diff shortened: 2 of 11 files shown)


s-perron:
This making progress.  It can unroll one of the larger cases we want to get, but it then gives a validation error:

> error: 5193: Loop header 932 is targeted by 0 back-edge blocks but the standard requires exactly one

I'll try to reduce it to a smaller test case that I can share.  It looks like a problem handling structured control-flow.

s-perron:
It looks like the `OpLoopMerge` for the loops are not being removed even though the loop are fully unrolled.

davors72:
That sounds unusual, there are a few tests which check that the OpLoopMerge is removed. If you can send us a sample I'll take a look at it @s-perron

s-perron:
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main" %o
               OpSource GLSL 430
               OpDecorate %o Location 0
       %void = OpTypeVoid
          %6 = OpTypeFunction %void
        %int = OpTypeInt 32 1
     %int_n2 = OpConstant %int -2
      %int_2 = OpConstant %int 2
       %bool = OpTypeBool
      %float = OpTypeFloat 32
%_ptr_Output_float = OpTypePointer Output %float
          %o = OpVariable %_ptr_Output_float Output
    %float_1 = OpConstant %float 1
       %main = OpFunction %void None %6
         %15 = OpLabel
               OpBranch %16
         %16 = OpLabel
         %27 = OpPhi %int %int_n2 %15 %26 %18
         %21 = OpSLessThanEqual %bool %27 %int_2
               OpLoopMerge %17 %18 Unroll
               OpBranchConditional %21 %22 %17
         %22 = OpLabel
         %23 = OpLoad %float %o
         %24 = OpFAdd %float %23 %float_1
               OpStore %o %24
               OpBranch %18
         %18 = OpLabel
         %26 = OpIAdd %int %27 %int_2
               OpBranch %16
         %17 = OpLabel
               OpReturn
               OpFunctionEnd
```

Here is an example,  I don't think it has to do with this change.  The problem seem to be when the loop exit condition is in the same block as the OpLoopMerge.

davors72:
@s-perron Ok, pushed a fix. Seems issue is that there are temporarily two branch instruction in the condition block, and since the merge instruction is found using tail()-- it just finds the other branch. Because we needed to keep some instructions around in the pass we have a vector of instructions marked for future deletion and delete them at the end. This isn't necessary for the conditional blocks branch however so I just changed it to delete that on the spot.

I'll integrate that sample into the regression tests and will add some missing asserts too.

source/opt/loop_descriptor.cpp
```diff
@@ -639,7 +691,17 @@ int64_t Loop::GetIterations(SpvOp condition, int64_t condition_value,
   return result;
 }
 
-ir::Instruction* Loop::FindInductionVariable(
+// Returns the list of induction variables within the loop.
+void Loop::GetInductionVariables(
+    std::vector<ir::Instruction*>& induction_variables) const {
+  for (ir::Instruction& inst : *loop_header_) {
+    if (inst.opcode() == SpvOp::SpvOpPhi) {
+      induction_variables.push_back(&inst);
```
dnovillo: Hmm, is this really correct?  There may be Phi instructions for variables that are not really induction variables.  You also need the Phi to have a cycle from within the loop.  Could you point at where in the literature you are taking induction variable detection from?  Having a paper/article source makes documentation much easier.

This is a contrived example, but |z| below is not really an induction variable:

```
#version 450
layout(location  = 0) in flat int  y;
layout(location  = 0) out int  outparm;

void main()
{
  int i, z=3;
  #pragma unroll
  for (i = 0; i < 16; i++) {
    if (y > 10) z+=5;
   }
  outparm = z;
}
```

source/opt/ir_builder.h
```diff
@@ -172,19 +172,31 @@ class InstructionBuilder {
...
+    uint32_t type_id =
+        GetContext()->get_type_mgr()->GetTypeInstruction(&int_type);
+
+    // Get the memory managed type so that it is safe to be stored by
+    // GetConstant.
+    analysis::Type* rebuilt_type =
+        GetContext()->get_type_mgr()->GetType(type_id);
+
+    GetContext()->InvalidateAnalysesExceptFor(
+        ir::IRContext::Analysis::kAnalysisLoopAnalysis);
```
dnovillo: Could you send this separately?  I'm not sure why we have to invalidate all analyses here, for instance.  Is there a test case?

source/opt/loop_unroller.cpp
```diff
@@ -348,20 +386,44 @@ void LoopUnrollerUtilsImpl::PartiallyUnrollResidualFactor(ir::Loop* loop,
...
   loop_induction_variable_ = state_.new_phi;
-
   // Unroll the new loop by the factor with the usual -1 to account for the
   // existing block iteration.
   Unroll(new_loop, factor);
 
+  LinkLastPhisToStart(new_loop);
+  AddBlocksToLoop(new_loop);
+
+  // Add the merge block to the back of the binary.
```
dnovillo: back of the binary?  I don't understand the comment.

source/opt/loop_unroller.cpp
```diff
@@ -420,36 +491,58 @@ void LoopUnrollerUtilsImpl::PartiallyUnrollResidualFactor(ir::Loop* loop,
...
+// Mark this loop as DontUnroll as it will already be unrolled and it may not
+// be safe to unroll a previously partially unrolled loop.
+void LoopUnrollerUtilsImpl::MarkLoopControlAsDontUnroll(ir::Loop* loop) const {
+  ir::Instruction* loop_merge_inst = loop->GetHeaderBlock()->GetLoopMergeInst();
+  assert(loop_merge_inst &&
+         "Loop merge instruction could not be found after entering unroller "
+         "(should have exited before this)");
+
+  uint32_t loop_control_dont_unroll = 2;
+  uint32_t loop_control_index = 2;
```
dnovillo: Suggest creating static kIndex variables in an anonymous namespace (as done in other passes).

source/opt/loop_unroller.cpp
```diff
@@ -420,36 +491,58 @@ void LoopUnrollerUtilsImpl::PartiallyUnrollResidualFactor(ir::Loop* loop,
...
 }
 
-// Duplicate the |loop| body |factor| number of times while keeping the loop
-// backedge intact.
+// Duplicate the |loop| body |factor| - 1 number of times while keeping the loop
+// backedge intact. This will leave the loop with |factor| number of bodies
+// after accounting for the initial body.
 void LoopUnrollerUtilsImpl::Unroll(ir::Loop* loop, size_t factor) {
+  // If we unroll a loop partially it may not be safe to unroll it further.
+  MarkLoopControlAsDontUnroll(loop);
```
dnovillo: Under which conditions?  This seems to always mark the loop unsafe to unroll again.

source/opt/loop_unroller.cpp
```diff
@@ -420,36 +491,58 @@ void LoopUnrollerUtilsImpl::PartiallyUnrollResidualFactor(ir::Loop* loop,
...
 
-  uint32_t phi_index = GetPhiIndexFromLabel(state_.previous_continue_block_,
-                                            state_.previous_phi_);
-  uint32_t phi_variable =
-      state_.previous_phi_->GetSingleWordInOperand(phi_index - 1);
-  uint32_t phi_label = state_.previous_phi_->GetSingleWordInOperand(phi_index);
-
-  ir::Instruction* original_phi = loop_induction_variable_;
+void LoopUnrollerUtilsImpl::RemoveDeadInstructions() {
+  // Remove the dead phi instructions.
```
dnovillo: Assert that these are all Phis being killed?  Also, maybe this could be left to DCE?  Though I'm fine with leaving this here, since it seems to be straightforward.

davors72: Yeah that was what happened in the original patch but as Alan pointed out the phis aren't just dead but also invalid as the branches will change. Technically it doesn't have to just be phis but that is the main thing being removed, I'll update the comment

dnovillo: Ah, yes, please do.  Perhaps renaming the class member to |invalidated_phis_| may make it clearer.

source/opt/ir_builder.h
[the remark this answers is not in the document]
davors72: I don't think we do, sorry, it's a leftover from the previous implementation of the constant type creation mechanism which did need them to be invalidated. I'll remove that

source/opt/loop_unroller.cpp
davors72: It's a allusion to the fact that the blocks in the function are in eventual binary order so we have to be careful when we insert the new merge block to make sure it is at the end of the new loop list as to not break the dominator ordering rule. I'll clear up the comment

source/opt/loop_descriptor.cpp
davors72: The reason we need this is that we need to track all the phis which are coming into the loop so we can change the uses in the unrolled bodies and clean up uses outside the loop after fully unrolling. So we still need to track them whether or not they are strictly induction variables. It's a generic approach which we need for phis so we aren't following any paper on it, I think it should probably be renamed to header phis or something. 

With the work we are doing on dependence analysis we have a much better solution for finding induction variables and eliminating a lot of existing code by using the newer approach. (using the scalar analysis graph, this is closer to my understanding from literature). It will be useful to have a generic induction variable detection solution, and it lets us clean up a lot of code, but it would be good if it could be deferred to a later patch once we have some more of the analysis tools coming soon.

source/opt/loop_unroller.cpp
davors72: This comes back to the current induction variable mechanism, one of the constraints we enforce is that the loop is bound by one induction variable which is modified by a single constant in the step instruction. We use the values from that to work out the number of iterations in the loop. Since the loop now has an induction variable with multiple step instructions it won't work and will incorrectly calculate the number of iterations.

s-perron: I can't believe I missed this.  That case 

```
#version 430

layout(location = 0) out float o;

void main(void) {
  float s = 0.0;
  for (int j = 0; j < 3; j += 1) {
    s += 1.0;
                j += 1;
  }
  o = s;
}
```

Will not unrolled correctly.

davors72: No it shouldn't be getting unrolled at all. I'll make sure to check it doesn't and add a test if we don't already have one. It's one of the things we can support fairly easily with the more advanced induction variable detection code mentioned above

source/opt/loop_unroller.cpp
```diff
@@ -502,17 +509,17 @@ void LoopUnrollerUtilsImpl::MarkLoopControlAsDontUnroll(ir::Loop* loop) const {
...
 }
 
 // Duplicate the |loop| body |factor| - 1 number of times while keeping the loop
 // backedge intact. This will leave the loop with |factor| number of bodies
 // after accounting for the initial body.
 void LoopUnrollerUtilsImpl::Unroll(ir::Loop* loop, size_t factor) {
-  // If we unroll a loop partially it may not be safe to unroll it further.
+  // If we unroll a loop partially it will not be safe to unroll it further.
+  // This is due to the current method of calculating the number of loop
+  // iterations.
```
s-perron: I guess I have to ask what happens in the loop that makes it unsafe, and could the user have simply written the loop that way in the first place?

davors72: The step count will be incorrect because the current step count code can only handle a single step instruction and in the unrolled loop it will be a series of step operations. The unroller will check this so the user could write code with multiple step instructions but it won't be unrolled. This should be the same for the partially unrolled loop if it was not marked DontUnroll. Basically since we know the loop doesn't meet the conditions we just mark it as DontUnroll, I thought this was a more explicit way of doing it but it's probably just unnecessary redundancy as it would be checked anyway. I can remove it if that would clear up the code a bit more

s-perron:
Not that it has the order correct, I cannot find anything else I want changed.

source/opt/loop_descriptor.cpp
dnovillo: Sure.  This test will currently return many false positives, but they're harmless.

dnovillo:
Merging now.

dnovillo:
@davors72 the branch is not merging cleanly with master, could you rebase and squash against master? I'll push the result.

davors72:
Done @dnovillo

dnovillo:
Thanks!  Running final tests and merging.

dnovillo:
Re-tested and pushed as e354984b09624c7fa2200e40555c98ca1a8fa5c6.

Issue #1328, opened 2018-02-20
Improve the compile time of spirv-opt.


s-perron:
We have come across a number of places where spirv-opt takes much longer that it needs to.  This will be a catch all issue for all compile time changes.

dnovillo:
Currently implementing a new SSA rewriting algorithm to improve compile time during Phi insertion.

dnovillo:
New SSA rewriting algorithm merged #1369.  Observing speedups of up to 7x on windows builds.

s-perron:
The next issue I want to look into is the reduce the number of times we kill analyzes.  In particular, the def-use manger and the inst-to-block mapping.  They are both used very often, and are expensive to kill on Windows.  In one profile that I did, kill those two account for almost 1/3 of the time spend compiling the shader.  I'll open another issue to group all of those changes together.

ratchetfreak:
Is that time spent in actual killing or in rebuilding?

If in killing because of recursive destructors, then that can be optimized but using something like a dedicated allocation to kill everything in one go and nulling out the references in each instruction/block (or maybe a analysis generation number which if doesn't match means it's effectively null)?

s-perron:
On Windows the time is spent mostly in killing the data structures, but build takes time as well.

Those are interesting ideas, and I'll keep them in mind.  For now we already do the work to keep the analyses up-to-date, but just did not set the bit on a number of optimizations to say that they are still valid.  That is most of the work, and it is pretty simple.  So we will get that done.  The next step would be to try to change the implementation of the analyses.  That is where we may use some of your ideas.  Thanks @ratchetfreak.

s-perron:
This issue has been tracking the long compile time on Windows when using DXC.  A large improvement should be coming with https://github.com/Microsoft/DirectXShaderCompiler/pull/1729.  Once that is through, we will consider this done.

PR #1329, opened 2018-02-20
Reduce instruction create and deletion during inlining.

s-perron:
When inlining a function call the instructions in the same basic block
as the call get cloned.  The clone is added to the set of new blocks
containing the inlined code, and the original instructions are deleted.

This PR will change this so that we simply move the instructions to the
new blocks.  This saves on the creation and deletion of the
instructions.

Contributes to #1328.

### DIFF
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -317,8 +317,10 @@ void InlinePass::GenInlineCode(
         if (firstBlock) {
           // Copy contents of original caller block up to call instruction.
           for (auto cii = call_block_itr->begin(); cii != call_inst_itr;
-               ++cii) {
-            std::unique_ptr<ir::Instruction> cp_inst(cii->Clone(context()));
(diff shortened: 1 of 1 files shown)


source/opt/inline_pass.cpp
```diff
@@ -426,8 +428,9 @@ void InlinePass::GenInlineCode(
         }
         // Copy remaining instructions from caller block.
         auto cii = call_inst_itr;
-        for (++cii; cii != call_block_itr->end(); ++cii) {
-          std::unique_ptr<ir::Instruction> cp_inst(cii->Clone(context()));
+        for (ir::Instruction* inst = call_inst_itr->NextNode(); inst; inst = call_inst_itr->NextNode()) {
```
dnovillo: Formatting here?

dnovillo:
There's also a set-but-unused warning in the Travis builder.

Issue #1331, opened 2018-02-20
Improve CI build times


zeux:
For contributors, or at least for me :), CI build times are a significant source of pain atm. A few days ago when the builds were broken due to SPIRV-Headers change, after that was fixed there was a queue of over 8 hours worth of builds. Even without this during a normal day like today that's still an issue - if a change has been reviewed and needs rebase for example, it won't get merged until hours later.

Obviously we're limited by build infra to some extend, but I can't help but wonder if this can be improved. I haven't looked into this at all - I know SPIRV-Tools source takes a long time to compile (presumably because of include complexity?), but beyond that I'm not sure if we're using parallel builds on Travis - they have ~1.5 core capacity for open source builds IIRC - and there are a LOT of configurations built on Travis and I'm not sure they are all justified. Same for Appveyor - I don't have readily available data on which one is more problematic, but do we really need to build 4 VS versions worth of builds there?

antiagainst:
Yep, we definitely feel the same pain as you. I've tried a few times in the past to alleviate the problem, but for each time after a while we gain the weight back (because of increasing code complexity, re-adding build configurations, etc.)...

One of the reasons is definitely Travis & Appveyor infrastructure limitation. My impression is that the performance of Travis fluctuates a lot, and there are kinda periodic service incidents. Appveyor is more dependable, but only one thread per PR/push is supported. It essentially means no parallelism at all. With that said, they are free services, we are already vergy grateful. We can look into alternative services or get paid membership with them. 

Another reason is, as you said, we have two many build configurations.

For Windows, we definitely want VS 2013 (discussions at https://github.com/KhronosGroup/glslang/issues/42). We'll need VS 2017 Debug, since it is an advanced compiler and likely to expose problems. We also need VS 2017 Release for artifacts uploading. Maybe we can remove VS 2015 Release build.

For Travis, we need to support Linux, MacOS (https://github.com/KhronosGroup/SPIRV-Tools/issues/1064), and Android. We want both Clang and GCC: Clang is more likely to expose problems in PRs, and GCC is still the default compiler on Ubuntu. So it seems we cannot remove any of the existing configurations. Travis does have better parallelism support than Appveyor; we use to leverage that by using ninja. But it ends up having problems with memory and causing compiler internal errors. So we changed to use make. That prolongs the build time. I'll see what I can do to improve here.

dj2:
The build speed is a lot faster after the move to Kokoro. I'm going to mark this as fixed at this point.

PR #1336, opened 2018-02-21
Reduce CI build time

antiagainst:
* Removed VS 2015 build configuration
* Used ninja instead of make on Linux, MacOS

Trying to address #1331

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,7 +8,6 @@ version: "{build}"
 # Put it first so we get its feedback first.
 os:
   - Visual Studio 2017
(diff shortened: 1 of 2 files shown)


dneto0:
This already went in as https://github.com/KhronosGroup/SPIRV-Tools/commit/432dc40412d2f557e4c49867c9481adf7679d496

antiagainst:
Yep. Sorry for the confusion. I submitted the change for Appveyor in 432dc40, but not the Travis changes, which doesn't seem to help.

PR #1339, opened 2018-02-21
Handle OpConstantNull in branch/switch conditions

### LINKED ISSUE #1338: CCP asserts if conditional branch is done on a null value
VisitBranchCondition doesn't handle the case of a boolean null value.

alan-baker:
Fixes #1338.

* No longer assume the branch/switch condition must be bool or int
constants (respectively)
* Added a couple unit tests for each case

I searched elsewhere in ccp and didn't find any other instances where the type of constant was assumed.


alan-baker:
Rebased and submitted as 01760d2f0f6422b330ea99dd67da72230ef2bbd7

PR #1340, opened 2018-02-21
Preserve analysies in the simplification pass

s-perron:
Fixes a bug at the same time.  In `UpdateDefUse`, if the definition
already exists, we are not suppose to analyse it again.  When you do
the entries for the definition are deleted, and we don't want that.
The check for this was wrong.


s-perron:
I'll add a test.  It is a functional issue.

s-perron:
Rebased, squashed, and merged as 3f19c203.

Issue #1341, opened 2018-02-21
Opt: Atomics (and other instructions with side-effects) being eliminated


greg-lunarg:
This is happening in DCEInst() which is used to cleanup after other optimizations. It isn't checking that instructions are combinators (no side-effects) before deleting.

I am claiming this and will be submitting a PR shortly.

danginsburg:
I can confirm that PR #1342 fixes the issue.  The attached SPIR-V reproduces the bug, but is fixed by the PR.

[cs_no_opt_good.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/1748345/cs_no_opt_good.zip)

PR #1342, opened 2018-02-21
Opt: Check for side-effects in DCEInst()

greg-lunarg:
This function now checks for side-effects before adding operand
instructions to the dead instruction work list.

Because this fix puts more pressure on IsCombinatorInstruction() to
be correct, this commit adds all OpConstant* and OpType* instructions
to combinator_ops_ set.

This fixes issue #1341.


s-perron:
Squashed, rebased, and merged as 46a9ec9d.

dneto0:
Hey @s-perron is a corresponding change needed in ADCE?

s-perron:
I don't think so.  It already checks if the instruction is a combinator.

PR #1343, opened 2018-02-22
BuiltIn variables can't have Location/Component decorations

zeux:
As per Vulkan spec, BuiltIn variables can't have Location or Component
decorations. On some drivers, these can lead to driver crashing when
compiling the shader pipeline; for example, NVidia/AMD desktop drivers:
https://github.com/KhronosGroup/glslang/issues/1182.

This change adds validation and tests to catch this.

### DIFF
--- a/source/validate_decorations.cpp
+++ b/source/validate_decorations.cpp
@@ -28,6 +28,14 @@ using libspirv::ValidationState_t;
 
 namespace {
 
+// Returns whether the given variable has a BuiltIn decoration.
+bool isBuiltInVar(uint32_t var_id, ValidationState_t& vstate) {
+  const auto& decorations = vstate.id_decorations(var_id);
+  return std::any_of(
+      decorations.begin(), decorations.end(),
+      [](const Decoration& d) { return SpvDecorationBuiltIn == d.dec_type(); });
+}
+
 // Returns whether the given structure type has any members with BuiltIn
 // decoration.
 bool isBuiltInStruct(uint32_t struct_id, ValidationState_t& vstate) {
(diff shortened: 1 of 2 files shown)


s-perron:
@atgoo Can you take a look at this?  It is in the validator, so I am not a good person to review that code.

dneto0:
Ping @atgoo

zeux:
Updated to only check this for Vulkan env (the rule is part of Vulkan spec, not SPIRV spec).

source/validate_decorations.cpp
```diff
@@ -91,6 +100,22 @@ spv_result_t CheckImportedVariableInitialization(ValidationState_t& vstate) {
   return SPV_SUCCESS;
 }
 
+// Checks whether a builtin variable is valid.
+spv_result_t CheckBuiltInVariable(uint32_t var_id, ValidationState_t& vstate) {
+  const auto& decorations = vstate.id_decorations(var_id);
+  for (const auto& d : decorations) {
+    if (spvIsVulkanEnv(vstate.context()->target_env)) {
```
atgoo: This could have been an early return (if not Vulkan, then there is nothing to check). But since there might be other rules in the future, I think it's fine.

source/validate_decorations.cpp
```diff
@@ -91,6 +100,22 @@ spv_result_t CheckImportedVariableInitialization(ValidationState_t& vstate) {
...
+      if (d.dec_type() == SpvDecorationLocation ||
+          d.dec_type() == SpvDecorationComponent) {
+        return vstate.diag(SPV_ERROR_INVALID_ID)
+               << "A BuiltIn variable (id " << var_id
```
atgoo: I think "According to the Vulkan spec..." would make it clearer, but I don't insist.

dneto0:
Rebased, squashed, and pushed into master as 8b27ba834da734

Thanks!

PR #1346, opened 2018-02-23
Merge arithmetic instructions with constant operands

alan-baker:
Adds Folding rules for merging arithmetic instructions where some operands are constant. Current rules:
 * negate of (negate, multiply, divide, add or subtract)
 * multiply of (multiply, divide or negate)
 * divide of (multiply, divide or negate)
 * reciprocal of divide
 * add of (add, subtract or negate)
 * subtract of (add, subtract or negate)

All rules handle integer, floating point or vectors of those types.

Changed IntMultiplyBy1 to handle 64 bit integers.

All the rules I intended to add have been added now.


source/opt/folding_rules.cpp
```diff
@@ -27,6 +27,782 @@ const uint32_t kExtInstInstructionInIdx = 1;
...
+
+std::vector<uint32_t> ExtractInts(uint64_t val) {
+  std::vector<uint32_t> words;
+  words.push_back(static_cast<uint32_t>(val));
+  words.push_back(static_cast<uint32_t>(val >> 32));
+  return words;
+}
+
+// Negates the integer constant |c|. Returns the id of the defining instruction.
+uint32_t NegateIntegerConstant(analysis::ConstantManager* const_mgr,
```
s-perron: Why do you want the id instead of the constant?

source/opt/folding_rules.cpp
```diff
@@ -27,6 +27,782 @@ const uint32_t kExtInstInstructionInIdx = 1;
...
+    return false;
+  };
+}
+
+// Merges negate into a add or sub operation if that operation contains a
+// constant operand.
+// Cases:
+// -(x + 2) = -2 - x
+// -(2 + x) = -2 - x
+// -(x - 2) = x + 2
```
s-perron: -(x - 2) = -x + 2 = 2 - x

source/opt/folding_rules.cpp
```diff
@@ -27,6 +27,782 @@ const uint32_t kExtInstInstructionInIdx = 1;
...
+
+    if (op_inst->opcode() == SpvOpFAdd || op_inst->opcode() == SpvOpFSub ||
+        op_inst->opcode() == SpvOpIAdd || op_inst->opcode() == SpvOpISub) {
+      std::vector<const analysis::Constant*> op_constants =
+          const_mgr->GetOperandConstants(op_inst);
+      if (op_constants[0] || op_constants[1]) {
+        bool zero_is_variable = op_constants[0] == nullptr;
+        bool is_add = (op_inst->opcode() == SpvOpFAdd) ||
+                      (op_inst->opcode() == SpvOpIAdd);
+        bool swap_operands = !(!zero_is_variable && is_add);
```
s-perron: This might need to be fixed up.  I think you swap for any subtraction.  This would also simplify the code below.

source/opt/folding_rules.cpp
```diff
@@ -27,6 +27,782 @@ const uint32_t kExtInstInstructionInIdx = 1;
...
+    if (HasFloatingPoint(type) && !inst->IsFloatingPointFoldingAllowed())
+      return false;
+
+    uint32_t width = ElementWidth(type);
+    if (width != 32 && width != 64) return false;
+
+    // Determine the constant input and the variable input in |inst|.
+    const analysis::Constant* const_input1 = ConstInput(constants);
+    ir::Instruction* other_inst = NonConstInput(context, constants[0], inst);
+    if (!const_input1) return false;
```
s-perron: Move this up one line.  Avoid doing the extra work in `NonConstInput`.

source/opt/folding_rules.cpp
```diff
@@ -27,6 +27,782 @@ const uint32_t kExtInstInstructionInIdx = 1;
...
+    if (HasFloatingPoint(type) && !other_inst->IsFloatingPointFoldingAllowed())
+      return false;
+
+    if (other_inst->opcode() == inst->opcode()) {
+      std::vector<const analysis::Constant*> other_constants =
+          const_mgr->GetOperandConstants(other_inst);
+      if (other_constants[0] || other_constants[1]) {
```
s-perron: Why not call `ConstInput`  here?

source/opt/constants.h
```diff
@@ -83,6 +83,30 @@ class Constant {
   virtual const ArrayConstant* AsArrayConstant() const { return nullptr; }
   virtual const NullConstant* AsNullConstant() const { return nullptr; }
 
+  // Returns the float representation of the constant. Must be a 32 bit
+  // FloatConstant.
```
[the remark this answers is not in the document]
alan-baker: Yes, will change.

source/opt/ir_context.h
```diff
@@ -414,6 +414,10 @@ class IRContext {
   // its definitions and uses.
   inline void UpdateDefUse(Instruction* inst);
 
+  // Returns true if we are allowed to fold or otherwise manipulate the
+  // instruction that defines |id|.
+  bool CanFoldFloatingPoint(uint32_t id);
```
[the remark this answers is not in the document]
alan-baker: I had forgotten to remove the declaration apparently.

source/opt/folding_rules.cpp
alan-baker: To facilitate both building vector constants and setting the new operands.

source/opt/folding_rules.cpp
alan-baker: Yeah, oops.

alan-baker: The tests have the right case.

source/opt/folding_rules.cpp
alan-baker: Good point, added it after that point.

source/opt/folding_rules.cpp
alan-baker: Done

source/opt/folding_rules.cpp
alan-baker: It might be awkward to read, but its the same as !is_add || zero_is_variable. I can change it to that.

dneto0:
Rebased, squashed, and pushed into master as 802cf053c

PR #1347, opened 2018-02-23
Remove uses DCEInst and call ADCE

s-perron:
The algorithm used in DCEInst to remove dead code is very slow.  It is
fine if you only want to remove a small number of instructions, but, if
you need to remove a large number of instructions, then the algorithm in
ADCE is much faster.

This PR removes the calls to DCEInst in the load-store removal passes
and adds a pass of ADCE afterwards.

A number of different iterations of the order of optimization, and I
believe this is the best I could find.

The results I have on 3 sets of shaders are:

Legalization:

Set 1: 5.39 -> 5.01
Set 2: 13.98 -> 8.38
Set 3: 98.00 -> 96.26

Performance passes:

Set 1: 6.90 -> 5.23
Set 2: 10.11 -> 6.62
Set 3: 253.69 -> 253.74

Size reduction passes:

Set 1: 7.16 -> 7.25
Set 2: 17.17 -> 16.81
Set 3: 112.06 -> 107.71

Note that the third set's compile time is large because of the large
number of basic blocks, not so much because of the number of
instructions.  That is why we don't see much gain there.

Contributes to #1328.

### DIFF
--- a/source/opt/local_single_block_elim_pass.cpp
+++ b/source/opt/local_single_block_elim_pass.cpp
@@ -50,7 +50,6 @@ bool LocalSingleBlockLoadStoreElimPass::HasOnlySupportedRefs(uint32_t ptrId) {
 bool LocalSingleBlockLoadStoreElimPass::LocalSingleBlockLoadStoreElim(
     ir::Function* func) {
   // Perform local store/load and load/load elimination on each block
-  std::vector<ir::Instruction*> dead_instructions;
   bool modified = false;
   for (auto bi = func->begin(); bi != func->end(); ++bi) {
     var2store_.clear();
@@ -72,7 +71,6 @@ bool LocalSingleBlockLoadStoreElimPass::LocalSingleBlockLoadStoreElim(
             if (pinned_vars_.find(varId) == pinned_vars_.end()) {
               auto si = var2store_.find(varId);
               if (si != var2store_.end()) {
-                dead_instructions.push_back(si->second);
               }
             }
             var2store_[varId] = &*ii;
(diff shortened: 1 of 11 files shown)


s-perron:
@greg-lunarg FYI.

greg-lunarg:
This is a bit surprising. Theoretically the only things that will be deleted by these passes is loads and stores. This is because the value that is stored is used to replace the load's id, so little else besides the load and store should generally go away. So one would think that DCEInst should really not need to be doing a lot of work when called from these passes.

Using DCEInst to remove loads and stores from these passes is probably inefficient. Still, it feels like using ADCE to remove the dead loads and stores created by these passes is overkill.

I wonder if these passes just cheaply removed the relevant load and store instructions as they went (without the potentially expensive recursive DCEInst calls) we might be able to get rid of these new ADCE calls and reduce compile time further.

greg-lunarg:
On continued consideration, it is possible for the original shader to contain "dead stores" and significant dead code to start with. I did not see this a lot with the base I was working with but it certainly can happen. In this case I can see that ADCE could be more efficient.

So it feels like we are dealing with something that could be dependent on the code base.

I will look at the timing for my base.

s-perron:
@greg-lunarg Thanks for running some extra tests.  I agree change in compile time can very significantly depending on the coding style.  The ones I see the biggest improvement on are those that have very large structures (more than a hundred elements).  SROA create lots of loads and stores when it splits up those structures.

greg-lunarg:
If you think the main task is just removing loads and stores, you might want to try the experiment I suggested. It could reduce opt time further.

greg-lunarg:
My base is also compiling faster with this change by about 10%.

s-perron:
Thanks Greg.  I'll merge this change.  Then for future work, we can look into how to reduce the time even further by looking at how we delete the dead code.

PR #1348, opened 2018-02-26
Make IR builder use the type manager for constants

davors72:
This change makes the IR builder use the type manager to generate
OpTypeInts when creating OpConstants. This avoids dangling references
being stored by the created OpConstants.

### DIFF
--- a/test/opt/ir_builder.cpp
+++ b/test/opt/ir_builder.cpp
@@ -296,6 +296,120 @@ OpFunctionEnd
   Match(text, context.get());
 }
 
+TEST_F(IRBuilderTest, ConstantAdder) {
+  const std::string text = R"(
+; CHECK: [[uint:%\w+]] = OpTypeInt 32 0
+; CHECK: OpConstant [[uint]] 13
+; CHECK: [[sint:%\w+]] = OpTypeInt 32 1
+; CHECK: OpConstant [[sint]] -1
+; CHECK: OpConstant [[uint]] 1
+; CHECK: OpConstant [[sint]] 34
+; CHECK: OpConstant [[uint]] 0
+; CHECK: OpConstant [[sint]] 0
+OpCapability Shader
+OpCapability Linkage
+OpMemoryModel Logical GLSL450
+%1 = OpTypeVoid
+%2 = OpTypeFunction %1
+%3 = OpFunction %1 None %2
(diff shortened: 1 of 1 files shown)


test/opt/ir_builder.cpp
```diff
@@ -296,6 +296,96 @@ OpFunctionEnd
   Match(text, context.get());
 }
 
+TEST_F(IRBuilderTest, ConstantAdder) {
+  const std::string text = R"(
+; CHECK: [[uint:%\w+]] = OpTypeInt 32 0
+; CHECK: [[u0:%\w+]] = OpConstant [[uint]] 13
```
alan-baker: If you need to match a captured variable later, I would suggest not capturing it. Instead you could have:
```
; CHECK: [uint:%\w+]] = OpTypeInt 32 0
; CHECK: OpConstant [[uint]] 13
; CHECK: [[sint:%\w+]] = OpTypeInt 32 1
; CHECK: OpConstant [[sint]] -1
...
```

test/opt/ir_builder.cpp
```diff
@@ -296,6 +296,96 @@ OpFunctionEnd
...
+  std::unique_ptr<ir::IRContext> context =
+      BuildModule(SPV_ENV_UNIVERSAL_1_2, nullptr, text);
+  EXPECT_NE(nullptr, context);
+
+  opt::InstructionBuilder builder(
+      context.get(), &*context->module()->begin()->begin()->begin());
+  EXPECT_NE(nullptr, builder.Add32BitUnsignedIntegerConstant(13));
+  EXPECT_NE(nullptr, builder.Add32BitSignedIntegerConstant(-1));
+
+  // Try adding the same constants again to make sure they aren't added.
```
alan-baker: The check statements are insufficient for this. Effcee does not match full lines so `%uint_13 = OpConstant %uint 13` would match `OpConstant %uint 1`. You should either save the results of the first call and compare them to the second call or you could use some `CHECK-NOT` commands. Just be aware that the negative pattern is only ensured between two matches.

davors72:
Updated and rebased, @alan-baker. The main change went through with unroller patch so this PR is just the test now.

dneto0:
Rebased and pushed into master as 20b8cdb7

PR #1349, opened 2018-02-26
Opt: Add constant folding for FToI and IToF

greg-lunarg:
TODO: Double support

### DIFF
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -272,65 +272,63 @@ ConstantFoldingRule FoldFPBinaryOp(BinaryScalarFoldingRule scalar_rule) {
 // This macro defines a |UnaryScalarFoldingRule| that performs float to
 // integer conversion.
 // TODO(greg-lunarg): Support for 64-bit integer types.
-#define FOLD_FTOI_OP()                                                        \
-  [](const analysis::Type* result_type, const analysis::Constant* a,          \
-     analysis::ConstantManager* const_mgr) -> const analysis::Constant* {     \
-    assert(result_type != nullptr && a != nullptr);                           \
-    const analysis::Integer* integer_type = result_type->AsInteger();         \
-    const analysis::Float* float_type = a->type()->AsFloat();                 \
-    assert(float_type != nullptr);                                            \
-    assert(integer_type != nullptr);                                          \
-    if (integer_type->width() != 32)                                          \
-      return nullptr;                                                         \
-    if (float_type->width() == 32) {                                          \
-      float fa = a->GetFloat();                                               \
-      uint32_t result = integer_type->IsSigned()                              \
-                            ? static_cast<uint32_t>(static_cast<int32_t>(fa)) \
-                            : static_cast<uint32_t>(fa);                      \
-      std::vector<uint32_t> words = {result};                                 \
-      return const_mgr->GetConstant(result_type, words);                      \
-    } else if (float_type->width() == 64) {                                   \
-      double fa = a->GetDouble();                                             \
-      uint32_t result = integer_type->IsSigned()                              \
-                            ? static_cast<uint32_t>(static_cast<int32_t>(fa)) \
-                            : static_cast<uint32_t>(fa);                      \
-      std::vector<uint32_t> words = {result};                                 \
-      return const_mgr->GetConstant(result_type, words);                      \
-    }                                                                         \
-    return nullptr;                                                           \
-  }
+UnaryScalarFoldingRule FoldFToIOp =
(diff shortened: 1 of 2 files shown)


source/opt/const_folding_rules.cpp
```diff
@@ -196,7 +254,84 @@ double GetDoubleFromConst(const analysis::Constant* c) {
   }
 }
 
-// This macro defines a |FloatScalarFoldingRule| that applies |op|.  The
+// Returns the integer value of |c|.  The constant |c| must have type
+// |Integer|, and width |32|.
+uint32_t GetIntegerFromConst(const analysis::Constant* c) {
```
s-perron: Alan add some functions to the Constant class to do this.  When you rebase you should be able to use them.

source/opt/const_folding_rules.cpp
```diff
@@ -196,7 +254,84 @@ double GetDoubleFromConst(const analysis::Constant* c) {
...
+  } else {
+    assert(c->AsNullConstant() && "c must be an integer constant.");
+    return 0.0f;
+  }
+}
+
+// This macro defines a |UnaryScalarFoldingRule| that performs float to
+// integer conversion.
+// TODO(greg-lunarg): Support for 64-bit integer types.
+#define FOLD_FTOI_OP()                                                        \
```
s-perron: I used a macro for the floating point binary operations because I could not find a way to pass in the operator.  In this case there is not parameter to the macro.  If there is no good reason for it, I would prefer to not use a macro.

source/opt/const_folding_rules.cpp
```diff
@@ -196,7 +254,84 @@ double GetDoubleFromConst(const analysis::Constant* c) {
...
+                             ? static_cast<float>(static_cast<int32_t>(ua))   \
+                             : static_cast<float>(ua);                        \
+      spvutils::FloatProxy<float> result(result_val);                         \
+      std::vector<uint32_t> words = {result.data()};                          \
+      return const_mgr->GetConstant(result_type, words);                      \
+    } else if (float_type->width() == 64) {                                   \
+      double result_val = integer_type->IsSigned()                            \
+                              ? static_cast<double>(static_cast<int32_t>(ua)) \
+                              : static_cast<double>(ua);                      \
+      spvutils::FloatProxy<double> result(result_val);                        \
```
s-perron: You can double check, but I believe Alan added something to the constant class to do this.

source/opt/const_folding_rules.cpp
greg-lunarg: OK

source/opt/const_folding_rules.cpp
greg-lunarg: Gone.

source/opt/const_folding_rules.cpp
greg-lunarg: Couldn't find anything in Constant class.

source/opt/const_folding_rules.cpp
```diff
@@ -272,65 +272,63 @@ ConstantFoldingRule FoldFPBinaryOp(BinaryScalarFoldingRule scalar_rule) {
...
-      double fa = a->GetDouble();                                             \
-      uint32_t result = integer_type->IsSigned()                              \
-                            ? static_cast<uint32_t>(static_cast<int32_t>(fa)) \
-                            : static_cast<uint32_t>(fa);                      \
-      std::vector<uint32_t> words = {result};                                 \
-      return const_mgr->GetConstant(result_type, words);                      \
-    }                                                                         \
-    return nullptr;                                                           \
-  }
+UnaryScalarFoldingRule FoldFToIOp =
```
s-perron: Small problem.  This is technically a global that is not a trivially destructible (I think). https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

source/opt/const_folding_rules.cpp
```diff
@@ -272,65 +272,63 @@ ConstantFoldingRule FoldFPBinaryOp(BinaryScalarFoldingRule scalar_rule) {
...
-      double result_val = integer_type->IsSigned()                            \
-                              ? static_cast<double>(static_cast<int32_t>(ua)) \
-                              : static_cast<double>(ua);                      \
-      spvutils::FloatProxy<double> result(result_val);                        \
-      std::vector<uint32_t> words = result.GetWords();                        \
-      return const_mgr->GetConstant(result_type, words);                      \
-    }                                                                         \
-    return nullptr;                                                           \
-  }
+UnaryScalarFoldingRule FoldIToFOp =
```
s-perron: Same as above.

s-perron:
I made the requested change myself.

Rebased, squashed, and merged as bdaf8d56.

PR #1350, opened 2018-02-27
Add more folding rules for vector shuffle.

s-perron:
Adds rule to fold OpVectorShuffle with constant inputs.

Adds rules to fold OpCompositeExtrac being fed by an OpVectorShuffle.


dneto0:
Please resolve the conflict.  I tried a somewhat trivial resolution and got dozens of errors.

s-perron:
I found an error in it that I need to fix anyway.

s-perron:
Made enough changes.  Should have a new review.

PR #1353, opened 2018-02-27
Remove Function::GetBlocks pushed by accident

Naghasan:
@s-perron Asked me to remove this here https://github.com/KhronosGroup/SPIRV-Tools/pull/1292#discussion_r168240514, but it ended up being readded and pushed somehow...

So here the revert.


dneto0:
Rebased and pushed onto master as 90e1637ce43

PR #1355, opened 2018-02-27
Do not merge integer division.

### LINKED ISSUE #1354: Folding shouldn't merge integer divides
Released I made a mistake with #1346. Integer-based division should not be folded. An example:
```
%1 = OpSDiv %int %x %int_64
%2 = OpIMul %int %1 %int_64
```
This folds to the wrong result.

alan-baker:
Fixes #1354.

* Removes merging of div with a div or mul for integers
* Removed the ability of PerformIntegerOperation to do divide to be safe
* Updated tests

### DIFF
--- a/source/opt/folding_rules.cpp
+++ b/source/opt/folding_rules.cpp
@@ -577,8 +577,8 @@ FoldingRule MergeMulMulArithmetic() {
 // Merges divides into subsequent multiplies if each instruction contains one
 // constant operand. Does not support integer operations.
 // Cases:
-// 2 * (x / 2) = 4 / x
-// 2 * (2 / x) = x * 1
+// 2 * (x / 2) = x * 1
+// 2 * (2 / x) = 4 / x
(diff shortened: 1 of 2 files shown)


source/opt/folding_rules.cpp
```diff
@@ -598,7 +575,7 @@ FoldingRule MergeMulMulArithmetic() {
 }
 
 // Merges divides into subsequent multiplies if each instruction contains one
-// constant operand.
+// constant operand. Does not support integer operations.
 // Cases:
 // 2 * (x / 2) = 4 / x
```
s-perron: I know this is already in, but this should be `2 * (x / 2) = x`.  The comment below does not look right either.

dneto0:
One of the GCC builds timed out.  I restarted it.

dneto0:
Rebased, squashed, and pushed into master as 9457cabbce

PR #1356, opened 2018-02-27
Add Vulkan validation rules for BuiltIn variables

atgoo:
Added a framework for validation of BuiltIn variables. The framework
allows implementation of flexible abstract rules which are required for
built-ins as the information (decoration, definition, reference) is not
in one place, but is scattered all over the module.

Validation rules are implemented as a map
id -> list<functor(instrution)>

Ids which are dependent on built-in types or objects receive a task
list, such as "this id cannot be referenced from function which is
called from entry point with execution model X; propagate this rule
to your descendants in the global scope".

Also refactored test/val/val_fixtures.


atgoo:
WIP as most tests are still missing.

source/val/validation_state.h
```diff
@@ -268,9 +268,17 @@ class ValidationState_t {
     return id_decorations_[id];
   }
   const std::vector<Decoration>& id_decorations(uint32_t id) const {
+    // TODO: This would throw or generate SIGABRT if id has no
```
dneto0: We don't know the assumed contract for this function since there is no method comment.  :-)
Not your original commit so I'm just pleased you added this TODO. Thanks.

source/validate_builtins.cpp
```diff
@@ -0,0 +1,2471 @@
...
+          "Vulkan spec doesn't allow BuiltIn PrimitiveId to be used for "
+          "variables with Input storage class if execution model is "
+          "TessellationEvaluation.",
+          SpvExecutionModelTessellationEvaluation, decoration, built_in_inst,
+          referenced_from_inst, std::placeholders::_1));
+    }
+
+    if (storage_class == SpvStorageClassOutput) {
+      assert(function_id_ == 0);
+      id_to_at_reference_checks_[referenced_from_inst.id()].push_back(std::bind(
```
dneto0: Good comment.

source/validate_builtins.cpp
```diff
@@ -0,0 +1,2471 @@
...
+  // Run validation.
+  spv_result_t Run();
+
+ private:
+  // Goes through all decorations in the module, if decoration is BuiltIn
+  // validates the instruction defining the decorated id. Also seeds
+  // id_to_at_reference_checks_ with decorated ids.
+  spv_result_t ValidateBuiltInsAtDefinition();
+
+  // The following section contains functions which are called when id defined
```
dneto0: Good. This scheme avoids useless clutter.

source/validate_builtins.cpp
```diff
@@ -0,0 +1,2471 @@
...
+  std::map<uint32_t, std::list<std::function<spv_result_t(const Instruction&)>>>
+      id_to_at_reference_checks_;
+
+  // Id of the function we are currently inside. 0 if not inside a function.
+  uint32_t function_id_ = 0;
+
+  // Entry points which can (indirectly) call the current function.
+  // The pointer either points to a vector inside to function_to_entry_points_
+  // or to no_entry_points_. The pointer is guaranteed to never be null.
+  const std::vector<uint32_t> no_entry_points;
```
dneto0: (This is missing the trailing underscore but I'm not sweating that detail.)

source/validate_builtins.cpp
```diff
@@ -0,0 +1,2471 @@
...
+    entry_points_ = &no_entry_points;
+    execution_models_.clear();
+  }
+}
+
+void BuiltInsValidator::ComputeFunctionToEntryPointMapping() {
+  for (const Instruction& inst : _.ordered_instructions()) {
+    const SpvOp opcode = inst.opcode();
+    if (opcode == SpvOpFunction) {
+      // We are looking for opcodes which can only be found at the top of
```
dneto0: Good comment.

source/validate_builtins.cpp
```diff
@@ -0,0 +1,2471 @@
...
+
+  // Traverses call tree and computes function_to_entry_points_,
+  // entry_point_to_execution_model_ and entry_point_to_execution_mode_.
+  void ComputeFunctionToEntryPointMapping();
+
+  const ValidationState_t& _;
+
+  // Mapping id -> list of rules which validate instruction referencing the
+  // id. Rules can create new rules and add them to this container.
+  // Using std::map, and not std::unordered_map to avoid iterator invalidation
```
dneto0: Good comment!

dneto0:
Rebased, squashed, and pushed into master as fe9121f72144f

Thank you!  This has been a long-standing request.

PR #1358, opened 2018-02-28
Support null constants better in folding

### LINKED ISSUE #1357: Folding operations don't properly account for vector type NullConstants
PerformOperation doesn't handle null vector constants properly. Maybe elsewhere in that newly added code too.

alan-baker:
Fixes #1357 

* getFloatConstantKind() now handles OpConstantNull
* PerformOperation() now handles OpConstantNull for vectors
* Fixed some instances where we would attempt to merge a division by 0
* added tests


s-perron:
Rebased, squashed, and merged as ce5941a6.

Issue #709, opened 2017-07-14
Constant folding.


dneto0:
Claimed opportunity in #664

s-perron:
I have looked for more opportunities to do constant folding.  I modified the folding code to print the opcode of an instruction every time it identifies all of operands are constant and the folder does not fold the instruction.  Here are the results from the three sets.  This is using master as of right now (90e1637ce43a6f53a6961f640e5096c71a4dee76).

Note that some of these will fail to fold because of type of the operand.  For example, the IAdd instruction will only get folded for scalars, and not vectors.

@alan-baker FYI.  This is what we discussed today.

Set 1:
- [ ]       1 ConvertUToF
- [ ]       5 Dot
- [ ]      22 Select
- [ ]      28 VectorTimesScalar
- [ ]     124 Bitcast
- [ ]     145 ConvertSToF
- [ ]     555 FNegate

Set 2:

- [ ]       1 Not
- [ ]       2 CompositeInsert
- [ ]       2 INotEqual
- [ ]       3 BitCount
- [ ]      10 ConvertFToS
- [ ]      15 Dot
- [ ]      20 ConvertSToF
- [ ]     119 VectorTimesScalar
- [ ]     125 ConvertFToU
- [ ]     152 ConvertUToF
- [ ]     192 Bitcast
- [ ]     235 Select
- [ ]     888 FNegate

Set 3:

- [ ]       1 CompositeInsert
- [ ]       1 IMul
- [ ]       1 Select
- [ ]       3 IAdd
- [ ]      24 MatrixTimesVector
- [ ]      27 ConvertFToU
- [ ]      73 Bitcast
- [ ]     235 ConvertSToF
- [ ]     856 VectorTimesScalar
- [ ]    1938 Dot
- [ ]    6307 ConvertUToF
- [ ]    7422 FNegate

s-perron:
@jaebaek I'll do `OpDot`.

jaebaek:
Sure but please leave OpDiv for my learning!

jaebaek:
For OpSelect .. when I input

         %const_true = OpConstantTrue %bool
       %main = OpFunction %void None %3
          %5 = OpLabel
         %32 = OpSelect %float %const_true %float_1 %float_2
         %31 = OpCompositeConstruct %v4float %float_1 %float_1 %float_1 %32
         %29 = OpAccessChain %_ptr_Output_v4float %_ %int_0
               OpStore %29 %31
[...]

PR #1365, opened 2018-03-02
Handles more cases of redundant selects

alan-baker:
* Handles OpConstantNull and vector types
 * vector selects (except against a null) are converted to vector
 shuffles
* Added tests

Contributes to #709.


dnovillo:
The build error on windows seems to be a server timeout.

dneto0:
Please rebase so we can get another bot run.
(I just pushed #1362 into master)

alan-baker:
Last build was just a timeout.

Rebased and submitted as 52bceb3569fd36edc35275603ecb83a68ae81aa1.

PR #1367, opened 2018-03-02
WIP:  Add support for decorate_string hlsl_functionality1 extensions

antiagainst:


antiagainst:
This PR needs https://github.com/KhronosGroup/SPIRV-Headers/pull/61

source/operand.cpp
```diff
@@ -405,6 +405,9 @@ std::function<bool(unsigned)> spvOperandCanBeForwardDeclaredFunction(
     case SpvOpSelectionMerge:
     case SpvOpDecorate:
     case SpvOpMemberDecorate:
+    case SpvOpDecorateId:
```
dneto0: Missing OpDecorateId is actually a bug in the current code.  Good catch.

source/val/validation_state.cpp
```diff
@@ -73,6 +73,9 @@ bool IsInstructionInLayoutSection(ModuleLayoutSection layout, SpvOp op) {
         case SpvOpGroupDecorate:
         case SpvOpGroupMemberDecorate:
         case SpvOpDecorationGroup:
+        case SpvOpDecorateId:
```
dneto0: Hmm. Maybe this should delegate to IsAnnotationInst.  :-)
Not requested for this CL.

utils/generate_grammar_tables.py
```diff
@@ -53,6 +53,7 @@
 SPV_EXT_shader_viewport_index_layer
 SPV_AMD_shader_image_load_store_lod
 SPV_AMD_shader_fragment_mask
+SPV_GOOGLE_decorate_string
```
dneto0: I would include SPV_GOOGLE_hlsl_functionality1 here as well.
Both of these are equally represented in the core grammar file.

dneto0:
Superceded by #1371

utils/generate_grammar_tables.py
antiagainst: I think `SPV_GOOGLE_hlsl_functionality1` is automatically picked up because it is listed as the extension for the `HlslSemanticGOOGLE` decoration. This one not becuase of #1368. But it does not harm to also list here.

PR #1369, opened 2018-03-02
SSA rewrite pass.

dnovillo:
This pass replaces the load/store elimination passes.  It implements the
SSA re-writing algorithm proposed in

     Simple and Efficient Construction of Static Single Assignment Form.
     Braun M., Buchwald S., Hack S., Leißa R., Mallon C., Zwinkau A. (2013)
     In: Jhala R., De Bosschere K. (eds)
     Compiler Construction. CC 2013.
     Lecture Notes in Computer Science, vol 7791.
     Springer, Berlin, Heidelberg

     https://link.springer.com/chapter/10.1007/978-3-642-37051-9_6

In contrast to common eager algorithms based on dominance and dominance
frontier information, this algorithm works backwards from load operations.

When a target variable is loaded, it queries the variable's reaching
definition.  If the reaching definition is unknown at the current location,
it searches backwards in the CFG, inserting Phi instructions at join points
in the CFG along the way until it finds the desired store instruction.

The algorithm avoids repeated lookups using memoization.

For reducible CFGs, which are a superset of the structured CFGs in SPIRV,
this algorithm is proven to produce minimal SSA.  That is, it inserts the
minimal number of Phi instructions required to ensure the SSA property, but
some Phi instructions may be dead
(https://en.wikipedia.org/wiki/Static_single_assignment_form).


source/opt/mem_pass.h
```diff
@@ -68,21 +56,57 @@ class MemPass : public Pass {
...
+  // variable of target type, false otherwise. Updates caches of target and
+  // non-target variables.
+  bool IsSSATargetVar(uint32_t varId);
+
+ protected:
+  // Returns true if |typeInst| is a scalar type, or a vector or matrix.
+  bool IsSSABaseTargetType(const ir::Instruction* typeInst) const;
+
+  // Returns true if |typeInst| is one of the concrete types or a struct or
+  // array
```
dneto0: Bad formatting here.

source/opt/mem_pass.h
```diff
@@ -68,21 +56,57 @@ class MemPass : public Pass {
...
 
-  // Return true if all uses of |id| are only name or decorate ops.
+  // Returns true if |varId| is a variable that can be rewritten into SSA.
+  // Rewriting a variable into SSA means replacing all its load/store operations
+  // with operations on SSA IDs.  Target variables have the following properies:
+  //
+  // - The variable has local function scope. That is, it is declared by an
+  // OpVariable instruction with Function storage class.
+  //
+  // previously identified target variable.
```
dneto0: Sentence fragment here.

source/opt/mem_pass.h
```diff
@@ -68,21 +56,57 @@ class MemPass : public Pass {
...
+  // Returns false if |varId| is a previously identified non-target variable.
+  //
+  // Non-target variables are variable of function scope of a target type that
+  // are accessed with constant-index access chains. not accessed with
```
dneto0: Sentence fragment at the last 3 words of this line.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -112,25 +112,25 @@ OpStore %f %float_0
 OpStore %i %int_0
 OpBranch %23
 %23 = OpLabel
-%38 = OpPhi %float %float_0 %22 %34 %25
-%39 = OpPhi %int %int_0 %22 %36 %25
+%39 = OpPhi %float %float_0 %22 %34 %25
```
dneto0: This must have been fun

tools/opt/opt.cpp
```diff
@@ -184,9 +184,6 @@ Options (in lexicographical order):
                a new basic block containing an unified return.
                This pass does not currently support structured control flow. It
                makes no changes if the shader capability is detected.
-  --local-redundancy-elimination
```
dneto0: This redundancy now looks like an easter egg.  :-)

test/opt/local_ssa_elim_test.cpp
```diff
@@ -668,7 +661,7 @@ OpFunctionEnd
 )";
 
   const std::string after =
-      R"(%47 = OpUndef %float
+      R"(%50 = OpUndef %float
```
dneto0: This test looks ok.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -542,26 +536,25 @@ OpStore %ie %25
 OpStore %i %int_0
 OpBranch %26
 %26 = OpLabel
-%40 = OpPhi %float %float_0 %23 %41 %28
-%41 = OpPhi %float %float_1 %23 %40 %28
-%42 = OpPhi %int %int_0 %23 %38 %28
-%44 = OpPhi %float %43 %23 %40 %28
```
dneto0: I guess this %44 one was dead.  Ok.
Also, was the only use of the OpUndef.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -417,12 +413,11 @@ OpBranch %26
 OpStore %f %36
 OpBranch %27
 %27 = OpLabel
-%43 = OpIAdd %int %46 %int_1
+%43 = OpIAdd %int %45 %int_1
 OpStore %i %43
 OpBranch %25
 %26 = OpLabel
-%49 = OpPhi %float %48 %28 %36 %40
```
dneto0: This was dead.  Ok

test/opt/local_ssa_elim_test.cpp
```diff
@@ -251,16 +250,15 @@ OpStore %f %float_0
 OpStore %i %int_0
 OpBranch %24
 %24 = OpLabel
-%44 = OpPhi %float %float_0 %23 %48 %26
-%45 = OpPhi %int %int_0 %23 %42 %26
-%47 = OpPhi %float %46 %23 %33 %26
```
dneto0: This %47 was dead.

source/opt/function.cpp
```diff
@@ -78,12 +78,19 @@ void Function::ForEachParam(const std::function<void(const Instruction*)>& f,
...
-  func.ForEachInst([&str](const ir::Instruction* inst) {
-    str << *inst;
+  str << func.PrettyPrint();
+  return str;
+}
+
+std::string Function::PrettyPrint(uint32_t options) const {
+  std::string str;
+  ForEachInst([&str, options](const ir::Instruction* inst) {
+    str += inst->PrettyPrint(options);
```
dneto0: This will result in a lot of memory reallocations.
Streaming is much more preferable.

source/opt/ssa_rewrite_pass.h
```diff
@@ -0,0 +1,237 @@
...
+  // Finalizes processing of Phi candidates.  Once the whole function has been
+  // scanned for loads and stores, the CFG will still have some incomplete and
+  // trivial Phis.  This will add missing arguments and remove trivial Phi
+  // candidates.
+  void FinalizePhiCandidates();
+
+  // Map holding the value of every SSA-target variable at every basic block
+  // where the variable is stored. defs_at_block_[block][var_id] = val_id
+  // means that there is a store instruction for variable |var_id| at basic
+  // block |block| with value |val_id|.
```
s-perron: The comment is not correct.  There does not have to be a store to the variable.  At merge nodes, it could contain the val_id of the candidate OpPhi.

source/opt/ssa_rewrite_pass.h
```diff
@@ -0,0 +1,237 @@
...
+  BlockDefsMap defs_at_block_;
+
+  // Map, indexed by basic block, holding all the Phi candidates created at that
+  // basic block during SSA rewriting.  phi_candidates_[block] returns a map,
+  // indexed by variable ID, for the Phi candidate created for that variable at
+  // |block|.
+  BlockPhisMap phi_candidates_;
+
+  // SSA replacement table.  This maps variable IDs, resulting from a load
+  // operation, to the value IDs that will replace them after SSA rewriting.
```
s-perron: What do you mean by "variable IDs"?  I think it is suppose to be the result of an OpVariable instruction.  I think a better way of saying what I think this means is "This maps the result ids of load instruction to the value ids that will replace them."

source/opt/ssa_rewrite_pass.h
```diff
@@ -0,0 +1,237 @@
...
+  // After all the rewriting decisions are made, a final scan through the IR is
+  // done to replace all uses of the original load ID with the value ID.
+  std::unordered_map<uint32_t, uint32_t> load_replacement_;
+
+  // Set of blocks that have been sealed already.
+  std::unordered_set<ir::BasicBlock*> sealed_blocks_;
+
+  // Memory pass requesting the SSA rewriter.
+  MemPass* pass_;
```
s-perron: A MemPass containing another MemPass?  Is this so you can get some of the unilities in MemPass?  Which ones, and do they belong there?  I would expect it to store a context, and that is it.

source/opt/ssa_rewrite_pass.cpp
```diff
@@ -0,0 +1,481 @@
...
+      }
+
+      uint32_t type_id = pass_->GetPointeeTypeId(
+          pass_->get_def_use_mgr()->GetDef(phi_candidate.var_id));
+      std::vector<ir::Operand> phi_operands;
+      uint32_t ix = 0;
+      for (uint32_t pred_label : pass_->cfg()->preds(phi_candidate.bb->id())) {
+        uint32_t op_val_id = phi_candidate.phi_args[ix++];
+        phi_operands.push_back(
+            {spv_operand_type_t::SPV_OPERAND_TYPE_ID, {op_val_id}});
```
s-perron: We already identified that this is a potential problem because if `op_val_id` is the result of a trivial phi candidate, the instruction defining this id will not be generated.

The solution I propose is a larger change but it can work:

1) Change `phi_candidates_` to a map from the result id the of candidate to a `PhiCandidate`.  It does not seem like the map from the basic block is actually used anyway, so it should not have a big impact on existing code.

2) At this point add code that as long as `op_val_id` is the potential result of a trivial phi will set it to the single phi_arg.   Then generate the instruction using that value instead.

The same change will be needed when replacing loads below.

source/opt/function.h
```diff
@@ -103,6 +103,14 @@ class Function {
   void ForEachParam(const std::function<void(const Instruction*)>& f,
                     bool run_on_debug_line_insts = false) const;
 
+  // Pretty-prints all the basic blocks in this function into a std::string.
+  //
+  // Blocks are pretty-printed in reverse post-order.
```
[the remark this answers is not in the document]
dnovillo: Recommendation from the paper authors.  Helps minimize the empty Phis the algorithm inserts.

source/opt/ssa_rewrite_pass.h
dnovillo: No, the rewriter is a standalone class.  It can be used by any pass as a helper.  The SSARewriterPass is the pass that uses it.  I organized it this way so that I could use it from the local multi-store elimination pass.

source/opt/ssa_rewrite_pass.cpp
```diff
@@ -0,0 +1,481 @@
...
+            << phi_cand->PrettyPrint(pass_->cfg()) << " -> ";
+#endif
+
+  if (phi_cand->phi_args.size() == 0) {
+    AddPhiOperands(phi_cand);
+  } else {
+    uint32_t ix = 0;
+    for (const auto& pred : pass_->cfg()->preds(phi_cand->bb->id())) {
+      ir::BasicBlock* pred_bb = pass_->cfg()->block(pred);
+      assert(IsBlockSealed(pred_bb) &&
```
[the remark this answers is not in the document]
dnovillo: Done.

source/opt/ssa_rewrite_pass.cpp
dnovillo: This has changed quite a bit now. PTAL.

dnovillo:
Thanks for the reviews.  The code shifted quite a bit since you last saw it.  @dneto0, @s-perron, could you take a new look?  I'll go over the feedback on style.

source/opt/function.h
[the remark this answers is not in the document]
dnovillo: I'm sorry, I totally spaced out what part of the code you were referring to.  I thought it was the actual rewriter.  I did this in rpo because it's generally a good ordering for a CFG, but you are right about your comment on structured control flow, it may look wonky.  I'll change it.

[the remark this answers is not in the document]
dnovillo: ... And... looking at the code again, it doesn't even work in RPO!  It traverses in linear order using ForEachInst.  Sigh.  Comment updated.

dnovillo:
@dneto0, @s-perron, @alan-baker the pass is ready for review again.  It passes all the internal tests (runtime tests in progress).  The time differences in Linux are not as pronounced as in Windows, but it's generally much faster.  There are other tweaks that can be done (it can insert more Phis, but I would rather address that in subsequent PRs.)

@greg-lunarg could you test this branch with your code, please?  It's a fairly big algorithmic change and would like to have good coverage.  Thanks.

source/opt/ssa_rewrite_pass.cpp
```diff
@@ -0,0 +1,594 @@
...
+    const auto& var_it = current_defs.find(var_id);
+    if (var_it != current_defs.end()) {
+      return var_it->second;
+    }
+  }
+
+  // Otherwise, look up the value for |var_id| in |bb|'s predecessors.
+  uint32_t val_id = 0;
+  auto& predecessors = pass_->cfg()->preds(bb->id());
+  if (!IsBlockSealed(bb)) {
```
alan-baker: This would be easier to read if you just asserted that the block is sealed and started the if statement below.

source/opt/ssa_rewrite_pass.h
s-perron: Don't bother to change this.  I can be done later if we want to do something about it.

source/opt/mem_pass.cpp
```diff
@@ -873,5 +850,32 @@ bool MemPass::CFGCleanup(ir::Function* func) {
...
 
+void MemPass::CollectTargetVars(ir::Function* func) {
+  seen_target_vars_.clear();
+  seen_non_target_vars_.clear();
+  supported_ref_vars_.clear();
+  type2undefs_.clear();
+
+  // Collect target (and non-) variable sets. Remove variables with
+  // non-load/store refs from target variable set
+  for (auto& blk : *func) {
```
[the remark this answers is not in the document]
s-perron: Open a follow up issue for this.

source/opt/ssa_rewrite_pass.cpp
```diff
@@ -0,0 +1,577 @@
...
+      // |phi_to_remove.result_id()| with |repl_id|.
+      for (uint32_t& arg : user_phi->phi_args()) {
+        if (arg == phi_to_remove.result_id()) {
+          arg = repl_id;
+        }
+      }
+    } else {
+      // For regular loads, traverse the |load_replacement_| table looking for
+      // instances of |phi_to_remove|.
+      for (auto& it : load_replacement_) {
```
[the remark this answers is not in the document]
s-perron: Don't change anything for now.  We can monitor it and see if it causes problems.

source/opt/ssa_rewrite_pass.h
dnovillo: Fixed.

source/opt/mem_pass.h
```diff
@@ -68,21 +56,57 @@ class MemPass : public Pass {
...
+  // non-constant-index access chains. Also cache non-target variables.
+  //
+  // If variable is not cached, return true if variable is a function scope
```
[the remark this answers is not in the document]
dnovillo: No.  It returns the result from the cache (first couple of lines from the comment).

source/opt/ssa_rewrite_pass.h
dnovillo: We also access other pass functions (CollectTargetVars, GetUndefVal, etc).  It can't just use the context.

source/opt/mem_pass.cpp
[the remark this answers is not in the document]
dnovillo: Not only variables need to be local storage.  This collector also makes sure that they are accessed without any decorations.  This code predates me.  I needed to incorporate it from the old Phi inserter to avoid the random segfaults I was getting before.  Not every load/store to local functions can be converted into SSA ids.

This loop does not even show up in profiles that I've seen, so it's not a big concern.

source/opt/mem_pass.cpp
```diff
@@ -873,5 +850,32 @@ bool MemPass::CFGCleanup(ir::Function* func) {
   return modified;
 }
 
+void MemPass::CollectTargetVars(ir::Function* func) {
```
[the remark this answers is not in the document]
dnovillo: I was trying to minimize code shuffling changes.  There's a bunch of code that the SSA rewriter is making dead.  I'll kill it in a subsequent PR.

source/opt/ssa_rewrite_pass.h
```diff
@@ -0,0 +1,295 @@
...
+
+  // Processes the load operation |inst| in basic block |bb|. This extracts the
+  // variable ID being stored into, determines whether the variable is an
+  // SSA-target variable, and, if it is, it reads its reaching definition by
+  // calling |GetReachingDef|.
+  void ProcessLoad(ir::Instruction* inst, ir::BasicBlock* bb);
+
+  // Reads the current definition for variable |var_id| in basic block |bb|.
+  // If |var_id| is not defined in block |bb| it walks up the predecessors of
+  // |bb|, creating new Phi instructions along the way, if needed.
```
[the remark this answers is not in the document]
dnovillo: Candidates.  Fixed.  Thanks.

source/opt/ssa_rewrite_pass.cpp
```diff
@@ -0,0 +1,577 @@
...
+}
+
+SSARewriter::PhiCandidate& SSARewriter::CreatePhiCandidate(uint32_t var_id,
+                                                           ir::BasicBlock* bb) {
+  uint32_t phi_result_id = pass_->context()->TakeNextId();
+  auto result = phi_candidates_.emplace(
+      phi_result_id, PhiCandidate(var_id, phi_result_id, bb));
+
+  // We should never try to create more than one Phi candidate for the same
+  // |var_id| on |bb|.
```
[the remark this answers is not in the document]
dnovillo: The whole pass is needed because we access services other than the context.  In a subsequent PR, I'll see about moving those services into the rewriter, if they're not used anywhere else.  I did not want to add too much noise in this PR (it's already pretty big).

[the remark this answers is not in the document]
dnovillo: Good point about the check.  Yeah, that check is now useless.  Thanks.

source/opt/ssa_rewrite_pass.cpp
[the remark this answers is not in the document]
dnovillo: No other ideas, because this does not show up in the profile.  If it becomes a problem, it should be easy to fix.

source/opt/ssa_rewrite_pass.cpp
```diff
@@ -0,0 +1,594 @@
...
+
+void SSARewriter::FinalizePhiCandidates() {
+#if SSA_REWRITE_DEBUGGING_LEVEL > 1
+  std::cerr << "Finalizing Phi candidates:\n\n";
+  PrintPhiCandidates();
+  std::cerr << "\n";
+#endif
+
+  // Now, complete the collected candidates.
+  while (incomplete_phis_.size() > 0) {
```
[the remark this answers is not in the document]
dnovillo: There can be additions, actually.  The call to GetReachingDef() will generate new Phi candidates.

source/opt/ssa_rewrite_pass.cpp
```diff
@@ -0,0 +1,594 @@
...
+  uint32_t same_id = 0;
+  for (uint32_t arg_id : phi_candidate->phi_args()) {
+    if (arg_id == same_id || arg_id == phi_candidate->result_id()) {
+      // This is a self-reference operand or a reference to the same value ID.
+      continue;
+    }
+    if (same_id != 0) {
+      // This Phi candidate merges at least two values.  Therefore, it is not
+      // trivial.
+      phi_candidate->MarkCopyOf(0);
```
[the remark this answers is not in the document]
dnovillo: Sure.  Fixed.

source/opt/ssa_rewrite_pass.cpp
dnovillo: I'm trying to workaround the fact that this code will not abort if we are in a release build.  We don't really have a good internal-compiler-error service.  If assertions are disabled, this returns 0, which will cause failures even in release builds.  If there's a better way of doing this, I'll happily change this.

We should really add an ICE facility to spirv-opt...

dnovillo:
Rebased and pushed as 735d8a579ecb6c5b5d41c2a81d695b49f7bd3248

PR #1370, opened 2018-03-02
Define Disassemble only when Effcee is used in fold_test

pierremoreau:
Fixes compilation with clang 5.0, when SPIRV_WERROR is on.


dneto0:
Rebased and pushed onto master as 6cd6e5ebef0

pierremoreau:
Thanks!

PR #1371, opened 2018-03-02
Support SPV_GOOGLE_decorate_string and SPV_GOOGLE_hlsl_functionality1

dneto0:
This is the same as #1367 but fixes my one complaint about it.

Also, I'm doing this as a separate PR to kick off bots, now that https://github.com/KhronosGroup/SPIRV-Headers/pull/61 has been merged.


dneto0:
FYI: @antiagainst

Issue #1174, opened 2018-01-05
DecorationManager does not properly handle decoration groups


alan-baker:
When decorations of an instruction are killed, only the leaf instruction are handled. Groups and group decorates are not properly removed. Additionally, since leaves are removed, we could incorrectly remove decorations for a live value (because OpGroupDecorate can specify multiple targets).

pierremoreau:
The whole `RemoveDecoration()` needs to be redone, as it no longer works even for lead decorations (the mappings are removed within the manager, but the instructions remain). I’ll work on that

alan-baker:
If you take a look at my PR #1147. You can see how I approach removing all dead decorations. You may find that useful for this.

pierremoreau:
I have a WIP available at https://github.com/pierremoreau/SPIRV-Tools/tree/fix_removedecoration; I still need to do some cleanup and add more unit tests to it before submitting it.
I’ll have a look at your PR, thanks for the link.

pierremoreau:
I was wondering, would it be okay to have `DecorationManager::RemoveDecorationFrom()` not remove decoration groups with no targets, and leave that to a later pass? Logic for that should already be part of ACDE thanks to @alan-baker’s work.

pierremoreau:
(Note: this is assuming OpGroupDecorate can target other groups. If this is not the case, then everything’s fine.)

Issue #924, opened 2017-10-26
eliminate group-decorate-of-group-decorate case in DecorationManager


dneto0:
We can remove https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/decoration_manager.cpp#L161 and 
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/decoration_manager.cpp#L171 cases because in a valid SPIR-V module:

An OpDecorationGroup result-id can't be one of the targets for an OpGroupDecorate or OpGroupMemberDecorate.  An OpDecorationGroup can only be the target for OpDecorate and OpDecorateId.

We should remove the confusing code that tries to handle the invalid case.

pierremoreau:
Where in the spec does it say that an OpDecorationGroup can only be used as a target to OpDecorate and OpDecorateId? I would love to get rid of group-decorate-of-group-decorate-of-group-decorate etc. cases, but I couldn’t find such restrictions in the spec.

dneto0:
I'll see if I can recall the argument

From the OpDecorationGroup description:

> A collector for Decorations from OpDecorate and OpDecorateId instructions. All such decoration instructions targeting this OpDecorationGroup instruction must precede it. Subsequent OpGroupDecorate and OpGroupMemberDecorate instructions that consume this instruction’s Result <id> will apply these decorations to their targets.

From the first sentence, OpDecorate and OpDecorateId can target OpDecorationGroup.  Fine.
From the second, those OpDecorate and OpDecorateId must precede the targeted OpDecorationGroup.

I guess you're concerned about a situation where an OpGroupDecorate can use an OpDecorationGroup result id as a *Target* operand?  I interpret that as banned but I can see your concern.  Maybe I read the second sentence in the description without the "such", and so I'd get

> "All decoration instructions targeting this OpDecorationGroup instruction must precede it"

and I also further interpret that OpGroupDecorate (and member decorate) must follow the OpDecorationGroup it targets.

But I see your concern.  I'll take this to the SPIR working group.

pierremoreau:
> I guess you're concerned about a situation where an OpGroupDecorate can use an OpDecorationGroup result id as a Target operand?

Right, for example would the following be allowed:
```
OpDecorate %1 Constant
%1 = OpDecorationGroup
OpDecorate %2 Restrict
OpGroupDecorate %1 %2
%2 = OpDecorationGroup
OpGroupDecorate %2 %3
[...]

Issue #898, opened 2017-10-20
Linker:  Incorrect handling of removal of FuncParamAttr when that is a grouped decoration


dneto0:
RemoveLinkageSpecificInstructions has this code:
```
 // Remove FuncParamAttr decorations of imported functions' parameters.
  // From the SPIR-V specification, Sec. 2.13:
  //   When resolving imported functions, the Function Control and all Function
  //   Parameter Attributes are taken from the function definition, and not
  //   from the function declaration.
  for (const auto& linking_entry : linkings_to_do) {
    for (const auto parameter_id :
         linking_entry.imported_symbol.parameter_ids) {
      for (ir::Instruction* decoration :
           decoration_manager->GetDecorationsFor(parameter_id, false)) {
        switch (decoration->opcode()) {
          case SpvOpDecorate:
          case SpvOpMemberDecorate:
            if (decoration->GetSingleWordInOperand(1u) ==
                SpvDecorationFuncParamAttr)
              decoration->ToNop();
            break;
          default:
            break;
        }
      }
    }
  }
```

This does the wrong thing if the FuncParamAttr is on a grouped decoration, and that grouped decoration applies to other IDs.

E.g. 
   OpDecorate %group FuncParamAttr
   %group = OpDecorationGroup
   OpGroupDecorate %group  %a %b %c

If the above code is triggered on id %a, then it will change the OpDecorate to an OpNop and lose the effect of the decorations on %b and %c.

Thanks to @s-perron for asking the question that lead to this: The question is "Why does the decoration manager have a GetDecorationsFor overload that returns non-const Instruction pointers" ?

pierremoreau:
Oops, my bad… that is completely wrong indeed. I am not entirely sure how this should be handle in this case: should the `DecorationManager` be extended to allow removing a specific decoration from an ID, or have some custom code in `RemoveLinkageSpecificInstructions`?

dneto0:
No problem.  I missed it on code reviews. :-)

Good question on how to handle it.  Normally I'd say add a method to the decoration manager.  HOWEVER, updating instructions often involves updating bookkeeping in several different places.
For example, removing the group decoration from an ID should also update any life DefUseManagers.  But you don't want to couple everything to everything. 

You also don't want to make web of publish/subscribe relationships between managers.  That's really hard to trace and can kill performance.

[...]

PR #1373, opened 2018-03-05
Reimplement the DecorationManager

pierremoreau:
This reimplementation fixes several issues when removing decorations associated to an ID (partially addresses #1174 and gives tools for fixing #898), as well as making it easier to remove groups; a few additional tests have been added.

`DecorationManager::RemoveDecoration()` will still not delete dead decorations it created, but I do not think it is its job either; given the following input

```
OpCapability Shader
OpCapability Linkage
OpMemoryModel Logical GLSL450
OpDecorate %2 Restrict
%2      = OpDecorationGroup
OpGroupDecorate %2 %1 %3
OpDecorate %4 Invariant
%4      = OpDecorationGroup
OpGroupDecorate %4 %2
%uint   = OpTypeInt 32 0
%1      = OpVariable %uint Uniform
%3      = OpVariable %uint Uniform
```

which of the following two outputs would you expect `RemoveDecoration(2)` to produce:

```
OpCapability Shader
OpCapability Linkage
OpMemoryModel Logical GLSL450
%uint = OpTypeInt 32 0
%1 = OpVariable %uint Uniform
%3 = OpVariable %uint Uniform
```

or

```
OpCapability Shader
OpCapability Linkage
OpMemoryModel Logical GLSL450
OpDecorate %4 Invariant
%4      = OpDecorationGroup
%uint   = OpTypeInt 32 0
%1      = OpVariable %uint Uniform
%3      = OpVariable %uint Uniform
```

### DIFF
--- a/source/opt/decoration_manager.cpp
+++ b/source/opt/decoration_manager.cpp
@@ -23,7 +23,7 @@ namespace opt {
 namespace analysis {
 
 void DecorationManager::RemoveDecorationsFrom(uint32_t id) {
-  auto const ids_iter = id_to_decoration_insts_.find(id);
+  const auto ids_iter = id_to_decoration_insts_.find(id);
   if (ids_iter == id_to_decoration_insts_.end()) return;
   id_to_decoration_insts_.erase(ids_iter);
 }
@@ -44,32 +44,32 @@ bool DecorationManager::HaveTheSameDecorations(uint32_t id1,
   using InstructionList = std::vector<const ir::Instruction*>;
   using DecorationSet = std::set<std::u32string>;
 
-  const InstructionList decorationsFor1 = GetDecorationsFor(id1, false);
-  const InstructionList decorationsFor2 = GetDecorationsFor(id2, false);
+  const InstructionList decorations_for1 = GetDecorationsFor(id1, false);
+  const InstructionList decorations_for2 = GetDecorationsFor(id2, false);
 
   // This function splits the decoration instructions into different sets,
   // based on their opcode; only OpDecorate, OpDecorateId and OpMemberDecorate
   // are considered, the other opcodes are ignored.
   const auto fillDecorationSets =
-      [](const InstructionList& decorationList, DecorationSet* decorateSet,
-         DecorationSet* decorateIdSet, DecorationSet* memberDecorateSet) {
-        for (const ir::Instruction* inst : decorationList) {
-          std::u32string decorationPayload;
+      [](const InstructionList& decoration_list, DecorationSet* decorate_set,
+         DecorationSet* decorate_id_set, DecorationSet* member_decorate_set) {
+        for (const ir::Instruction* inst : decoration_list) {
+          std::u32string decoration_payload;
           // Ignore the opcode and the target as we do not want them to be
           // compared.
           for (uint32_t i = 1u; i < inst->NumInOperands(); ++i)
             for (uint32_t word : inst->GetInOperand(i).words)
-              decorationPayload.push_back(word);
+              decoration_payload.push_back(word);
 
           switch (inst->opcode()) {
             case SpvOpDecorate:
-              decorateSet->emplace(std::move(decorationPayload));
+              decorate_set->emplace(std::move(decoration_payload));
               break;
             case SpvOpMemberDecorate:
-              memberDecorateSet->emplace(std::move(decorationPayload));
+              member_decorate_set->emplace(std::move(decoration_payload));
               break;
             case SpvOpDecorateId:
-              decorateIdSet->emplace(std::move(decorationPayload));
+              decorate_id_set->emplace(std::move(decoration_payload));
               break;
             default:
               break;
@@ -77,21 +77,21 @@ bool DecorationManager::HaveTheSameDecorations(uint32_t id1,
         }
       };
 
-  DecorationSet decorateSetFor1;
-  DecorationSet decorateIdSetFor1;
-  DecorationSet memberDecorateSetFor1;
-  fillDecorationSets(decorationsFor1, &decorateSetFor1, &decorateIdSetFor1,
-                     &memberDecorateSetFor1);
+  DecorationSet decorate_set_for1;
+  DecorationSet decorate_id_set_for1;
+  DecorationSet member_decorate_set_for1;
+  fillDecorationSets(decorations_for1, &decorate_set_for1, &decorate_id_set_for1,
+                     &member_decorate_set_for1);
 
-  DecorationSet decorateSetFor2;
-  DecorationSet decorateIdSetFor2;
-  DecorationSet memberDecorateSetFor2;
-  fillDecorationSets(decorationsFor2, &decorateSetFor2, &decorateIdSetFor2,
-                     &memberDecorateSetFor2);
+  DecorationSet decorate_set_for2;
+  DecorationSet decorate_id_set_for2;
+  DecorationSet member_decorate_set_for2;
+  fillDecorationSets(decorations_for2, &decorate_set_for2, &decorate_id_set_for2,
+                     &member_decorate_set_for2);
 
-  return decorateSetFor1 == decorateSetFor2 &&
-         decorateIdSetFor1 == decorateIdSetFor2 &&
-         memberDecorateSetFor1 == memberDecorateSetFor2;
+  return decorate_set_for1 == decorate_set_for2 &&
+         decorate_id_set_for1 == decorate_id_set_for2 &&
+         member_decorate_set_for1 == member_decorate_set_for2;
 }
 
 // TODO(pierremoreau): If OpDecorateId is referencing an OpConstant, one could
@@ -143,8 +143,8 @@ void DecorationManager::AddDecoration(ir::Instruction* inst) {
     case SpvOpDecorate:
     case SpvOpDecorateId:
     case SpvOpMemberDecorate: {
-      auto const target_id = inst->GetSingleWordInOperand(0u);
-      auto const group_iter = group_to_decoration_insts_.find(target_id);
+      const auto target_id = inst->GetSingleWordInOperand(0u);
+      const auto group_iter = group_to_decoration_insts_.find(target_id);
       if (group_iter != group_to_decoration_insts_.end())
         group_iter->second.push_back(inst);
       else
@@ -153,8 +153,8 @@ void DecorationManager::AddDecoration(ir::Instruction* inst) {
     }
     case SpvOpGroupDecorate:
       for (uint32_t i = 1u; i < inst->NumInOperands(); ++i) {
-        auto const target_id = inst->GetSingleWordInOperand(i);
-        auto const group_iter = group_to_decoration_insts_.find(target_id);
+        const auto target_id = inst->GetSingleWordInOperand(i);
+        const auto group_iter = group_to_decoration_insts_.find(target_id);
         if (group_iter != group_to_decoration_insts_.end())
           group_iter->second.push_back(inst);
         else
@@ -163,8 +163,8 @@ void DecorationManager::AddDecoration(ir::Instruction* inst) {
       break;
     case SpvOpGroupMemberDecorate:
       for (uint32_t i = 1u; i < inst->NumInOperands(); i += 2u) {
-        auto const target_id = inst->GetSingleWordInOperand(i);
-        auto const group_iter = group_to_decoration_insts_.find(target_id);
+        const auto target_id = inst->GetSingleWordInOperand(i);
+        const auto group_iter = group_to_decoration_insts_.find(target_id);
         if (group_iter != group_to_decoration_insts_.end())
           group_iter->second.push_back(inst);
         else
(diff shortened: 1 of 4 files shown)


alan-baker:
Your example is invalid because OpGroupDecorate cannot target an OpDecorationGroup. Does that simplify the logic required in the DecorationManager?

pierremoreau:
That definitely would make some things simpler, however there is nothing in the spec preventing OpGroupDecorate from targeting other groups. @dneto0 brought up this point in #924 and was going to bring it up to the SPIR-V working group, but I haven’t heard back since.

source/opt/decoration_manager.cpp
```diff
@@ -18,14 +18,91 @@
...
+
+      const uint32_t last_operand_index = inst->NumInOperands() - stride;
+      if (i < last_operand_index)
+        inst->GetInOperand(i) = inst->GetInOperand(last_operand_index);
+      // Remove the associated literal, if it exists.
+      if (stride == 2u) {
+        if (i < last_operand_index)
+          inst->GetInOperand(i + 1u) =
+              inst->GetInOperand(last_operand_index + 1u);
+        inst->RemoveInOperand(last_operand_index + 1u);
```
s-perron: The use if `pred` seem inconsistent.  If the decoration is applied through a group you do not check it.  That does not seems right.

source/opt/decoration_manager.cpp
```diff
@@ -18,14 +18,91 @@
...
+    } else if (was_modified) {
+      indirect_decorations_to_remove.emplace(inst);
+      context->AnalyzeUses(inst);
+    }
+  }
+
+  // Schedule all instructions applying the group for removal if instructed as
+  // such by |pred|.
+  for (ir::Instruction* inst : decorations_info.decorate_insts)
+    insts_to_kill.push_back(inst);
```
s-perron: Can you do this if `pred` returned false for any of the direct decorations?

source/opt/decoration_manager.cpp
```diff
@@ -18,14 +18,91 @@
...
+  for (ir::Instruction* inst : insts_to_kill) context->KillInst(inst);
+
+  if (decorations_info.direct_decorations.empty() &&
+      decorations_info.indirect_decorations.empty() &&
+      decorations_info.decorate_insts.empty()) {
+    id_to_decoration_insts_.erase(ids_iter);
+  }
+
+  // Remove the OpDecorationGroup defining this group.
+  if (is_group) context->KillInst(context->get_def_use_mgr()->GetDef(id));
```
s-perron: You cannot delete this unless all uses of `id` are also killed, but what if `pred` returned false for something?

source/opt/decoration_manager.cpp
```diff
@@ -180,46 +235,42 @@ template <typename T>
...
-    if (group_iter != group_to_decoration_insts_.end()) {
-      // Process all the decorations applied by the group.
-      for (T inst : group_iter->second) process(inst);
-    } else {
-      // Something went wrong.
-      assert(false);
-      return std::vector<T>();
-    }
+  // Process the decorations of all groups applied to |id|.
+  std::stack<uint32_t> groups_to_process;
```
s-perron: Just use an std::vector.  We want to avoid more uses of std::stack, std::queue, std::list, and std::deueque.  They perform very poorly on windows.

source/opt/decoration_manager.cpp
```diff
@@ -180,46 +235,42 @@ template <typename T>
...
+  for (const ir::Instruction* inst : target_data.indirect_decorations)
+    groups_to_process.push(inst->GetSingleWordInOperand(0u));
+
+  std::unordered_set<uint32_t> group_ids;
```
s-perron: Do you think there will be duplicates?  If not std::vector would again be the better data structure.  You don't seem to do any searches.

source/opt/decoration_manager.cpp
```diff
@@ -260,16 +311,25 @@ void DecorationManager::ForEachDecoration(
...
   assert(f && "Missing function parameter f");
-  auto const decoration_list = id_to_decoration_insts_.find(from);
+  const auto decoration_list = id_to_decoration_insts_.find(from);
   if (decoration_list == id_to_decoration_insts_.end()) return;
-  for (ir::Instruction* inst : decoration_list->second) {
+  for (ir::Instruction* inst : decoration_list->second.direct_decorations) {
+    // simply clone decoration and change |target-id| to |to|
+    std::unique_ptr<ir::Instruction> new_inst(inst->Clone(module_->context()));
+    new_inst->SetInOperand(0, {to});
+    id_to_decoration_insts_[to].direct_decorations.push_back(new_inst.get());
```
s-perron: Doing the same hash lookup for every iteration.  Might want to move that out of the loop.

source/opt/decoration_manager.cpp
```diff
@@ -260,16 +311,25 @@ void DecorationManager::ForEachDecoration(
 void DecorationManager::CloneDecorations(
     uint32_t from, uint32_t to, std::function<void(ir::Instruction&, bool)> f) {
```
s-perron: Given that the decoration now knows about the context, we might be able to remove the call back `f`.  Just have the context analyze the new instructions.

source/opt/decoration_manager.h
```diff
@@ -35,10 +36,16 @@ class DecorationManager {
   }
   DecorationManager() = delete;
 
-  // Removes all decorations from |id|, which should not be a group ID.
-  void RemoveDecorationsFrom(uint32_t id);
+  // Removes all decorations from |id| for which |pred| returns true.
+  // |id| should not be a group ID.
+  void RemoveDecorationsFrom(uint32_t id,
```
s-perron: Can you explain why you are adding `pred`?  It will make the implementation much more difficult, so we should have a reason.

pierremoreau:
@alan-baker Ah, I missed the update to OpGroupDecorate in SPIR-V 1.2 rev 3 that explicitly says it can’t target groups. I’ll update the code accordingly.

source/opt/decoration_manager.cpp
pierremoreau: You are right, I completely missed that case.

source/opt/decoration_manager.cpp
pierremoreau: No, you are right, I should check that it has no direct decorations left.

source/opt/decoration_manager.cpp
pierremoreau: This should be in the if statement right above it, to guard against that eventuality.

source/opt/decoration_manager.cpp
pierremoreau: Okay, I’ll do the change to std::vector.

source/opt/decoration_manager.h
pierremoreau: The reason for adding pred was explained in #898, more specifically in @dneto0’s comment https://github.com/KhronosGroup/SPIRV-Tools/issues/898#issuecomment-338413513. Currently the linker has some code (which is wrong) that loops over all decorations to remove only LinkageAttributes Import ones, or FuncParamAttr ones. The idea is to have the DecorationManager do it, in a generic way, instead.

source/opt/decoration_manager.cpp
pierremoreau: I’ll do that.

source/opt/decoration_manager.cpp
pierremoreau: I think there could be duplicates, but I’ll try to come up with such a SPIR-V example to check whether it is possible or not.

source/opt/decoration_manager.cpp
pierremoreau: I’ll store a reference to |to|’s `TargetData` outside of the loop, and use it in this loop and in others further down in the function.

pierremoreau:
I forgot to add in one of the commits:

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/924

pierremoreau:
Since groups of groups aren’t valid, I could change the code to remove a dead group when removing its last target in RemoveDecoration(). Or should this be left to ADCE, which already handles dead decorations IIRC?

alan-baker:
You don't need to remove target-less group decorations. Feel free to leave them for ADCE.

pierremoreau:
> You don't need to remove target-less group decorations. Feel free to leave them for ADCE.

Okay, nice. In that case, I think this should also include:

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1174

source/opt/decoration_manager.cpp
```diff
@@ -46,6 +46,21 @@ void DecorationManager::RemoveDecorationsFrom(
   for (ir::Instruction* inst : decorations_info.indirect_decorations) {
     assert(inst->opcode() == SpvOpGroupDecorate ||
            inst->opcode() == SpvOpGroupMemberDecorate);
+
+    std::vector<ir::Instruction*> group_decorations_to_keep;
+    const uint32_t group_id = inst->GetSingleWordInOperand(0u);
+    const auto group_iter = id_to_decoration_insts_.find(group_id);
+    for (ir::Instruction* group_decoration :
+         group_iter->second.direct_decorations)
```
s-perron: Please add the braces.  It is easier to see what is in the scope of the for loop.

source/opt/decoration_manager.cpp
```diff
@@ -46,6 +46,21 @@ void DecorationManager::RemoveDecorationsFrom(
...
+      if (!pred(*group_decoration))
+        group_decorations_to_keep.push_back(group_decoration);
+
+    // If all decorations should be kept, move to the next group
+    if (group_decorations_to_keep.size() ==
+        group_iter->second.direct_decorations.size())
+      continue;
```
s-perron: If you are keeping all of the decorations, should you clear `group_decorations_to_keep` so you don't add them below?

source/opt/decoration_manager.cpp
```diff
@@ -242,35 +276,26 @@ std::vector<T> DecorationManager::InternalGetDecorationsFor(
...
+
   // Process |id|'s decorations.
-  decorations.reserve(target_data.direct_decorations.size());
-  for (ir::Instruction* inst : ids_iter->second.direct_decorations) {
-    const bool is_linkage =
-        inst->opcode() == SpvOpDecorate &&
-        inst->GetSingleWordInOperand(1u) == SpvDecorationLinkageAttributes;
-    if (include_linkage || !is_linkage) decorations.push_back(inst);
-  }
+  process_direct_decorations(ids_iter->second.direct_decorations);
```
s-perron: If there are only direct calls to this function, why use a lambda?

source/opt/decoration_manager.cpp
pierremoreau: You are correct, I need to do that. I’ll add a test that would fail with the current code, to detect possible regressions and check that the code is properly fixed.

source/opt/decoration_manager.cpp
pierremoreau: Will do

source/opt/decoration_manager.cpp
pierremoreau: The lambda is called twice: once on line 291, and once on line 298 inside the loop over groups’ direct decorations.

s-perron: I know it is called twice.  I'm not suggesting it should be inlined.  I'm just wondering why a regular function, where `include_linkage` and `decorations` are passed as parameters,  would not work.

pierremoreau: Ah sorry, I misunderstood your comment. A regular function could work as well, but I am not sure who else would call it, and since there was only one place needing it, a lambda seemed to be the best. Also, I was thinking I could replace `include_linkage` by a predicate function instead.

source/opt/decoration_manager.cpp
pierremoreau: Actually, looking at the code again, they won’t be added later on (I assume you are thinking of the code [right here](https://github.com/KhronosGroup/SPIRV-Tools/pull/1373/files/91f73ab072b9176e4e539a343d868f005847e8a4#diff-b567013ae7fa0970645039363ebc5cd7R99)): `group_decorations_to_keep` is local to the loop, so when continuing and going to the next loop iteration, `group_decorations_to_keep` will be a new, empty, vector.

I added two new tests to check that behaviour, which uncovered a bug where the medata was not updated with the clone instructions due to calling `AnalyzeDefUse()` (which does not update decorations) instead of `AnalyzeUses()`.
I alos added a test for the `CloneDecorations()` which was suffering from the same issue, as well as from an other bug (the list of direct decorations from groups was modified as we were iterating over it).

pierremoreau:
Please have another look.
(I added the fixes messages to the last commit.)

@s-perron I addressed your remaining comments, apart from the regular function one. I can make the lambda into a regular function if you want, but I would rather get rid of `include_linkage` and replace it by a predicate, similar to `RemoveDecorationsFrom()`.

s-perron:
The lambda is not a big deal.  I we can put it through anyway.  However, even if you change `include_linkage` to an `std::function`, it will not force `process_direct_decorations` to be a lambda.  You can simply pass the predicate function as a parameter to the function.

pierremoreau:
>  However, even if you change include_linkage to an std::function, it will not force process_direct_decorations to be a lambda. You can simply pass the predicate function as a parameter to the function.

No, in that case I would even drop the lambda and inline the code, though a regular function would work fine as well, as it would even be used by `RemoveDecorationsFrom()`.

s-perron:
Rebased, squashed, and merged as 5bd55f10.

pierremoreau:
Thank you all for the comments and reviews. :-)

PR #1377, opened 2018-03-05
Handle half folding gracefully

### LINKED ISSUE #1376: Optimizer internal error on float16 usage
I have a SPIR-V source that uses float16 (below):

[float16.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/1781904/float16.txt)

If you assemble and run the optimizer on it, you'll get an internal error.
```
C:\spirv-tools\build\tools\spirv-as.exe float16.txt -o float16.spv
C:\spirv-tools\build\tools\spirv-opt.exe -O -o float16_after_optimization.spv float16.spv
```

You get this assertion failure:
```
Assertion failed: type()->AsFloat()->width() == 32 && "Not a 32-bit floating point value.", file c:\spirv-tools\source\opt\constants.h, line 222
```
cc @antiagainst @alan-baker  @s-perron @dnovillo

alan-baker:
Fixes #1376

* Added early returns to folding rules to prevent half attempts
* Added some tests

### DIFF
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -342,6 +342,8 @@ UnaryScalarFoldingRule FoldIToFOp() {
     assert(result_type == a->type() && result_type == b->type());         \
     const analysis::Float* float_type = result_type->AsFloat();           \
     assert(float_type != nullptr);                                        \
+    if (float_type->width() != 32 && float_type->width() != 64)           \
+      return nullptr;                                                     \
     if (float_type->width() == 32) {                                      \
       float fa = a->GetFloat();                                           \
       float fb = b->GetFloat();                                           \
@@ -391,6 +393,8 @@ bool CompareFloatingPoint(bool op_result, bool op_unordered,
     assert(a->type() == b->type());                                       \
     const analysis::Float* float_type = a->type()->AsFloat();             \
     assert(float_type != nullptr);                                        \
+    if (float_type->width() != 32 && float_type->width() != 64)           \
+      return nullptr;                                                     \
     if (float_type->width() == 32) {                                      \
       float fa = a->GetFloat();                                           \
       float fb = b->GetFloat();                                           \
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -342,8 +342,6 @@ UnaryScalarFoldingRule FoldIToFOp() {
     assert(result_type == a->type() && result_type == b->type());         \
     const analysis::Float* float_type = result_type->AsFloat();           \
     assert(float_type != nullptr);                                        \
-    if (float_type->width() != 32 && float_type->width() != 64)           \
-      return nullptr;                                                     \
     if (float_type->width() == 32) {                                      \
       float fa = a->GetFloat();                                           \
       float fb = b->GetFloat();                                           \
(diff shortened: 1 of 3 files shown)


source/opt/const_folding_rules.cpp
```diff
@@ -342,6 +342,8 @@ UnaryScalarFoldingRule FoldIToFOp() {
     assert(result_type == a->type() && result_type == b->type());         \
     const analysis::Float* float_type = result_type->AsFloat();           \
     assert(float_type != nullptr);                                        \
+    if (float_type->width() != 32 && float_type->width() != 64)           \
+      return nullptr;                                                     \
```
s-perron: Isn't this already handled? There are two `if` below, and it neither of them are true, we return nullptr.

source/opt/const_folding_rules.cpp
```diff
@@ -391,6 +393,8 @@ bool CompareFloatingPoint(bool op_result, bool op_unordered,
     assert(a->type() == b->type());                                       \
     const analysis::Float* float_type = a->type()->AsFloat();             \
     assert(float_type != nullptr);                                        \
+    if (float_type->width() != 32 && float_type->width() != 64)           \
+      return nullptr;                                                     \
```
s-perron: Same here, I don't see why it is needed.

source/opt/const_folding_rules.cpp
alan-baker: Good point, guess I misread those.

source/opt/folding_rules.cpp
```diff
@@ -1638,8 +1638,12 @@ FloatConstantKind getFloatConstantKind(const analysis::Constant* constant) {
 
     return kind;
   } else if (const analysis::FloatConstant* fc = constant->AsFloatConstant()) {
-    double value = (fc->type()->AsFloat()->width() == 64) ? fc->GetDoubleValue()
-                                                          : fc->GetFloatValue();
+    if (fc->IsZero()) return FloatConstantKind::Zero;
```
zeux: Would it make sense to add something like GetDoubleValue() to FloatConstant class? As written this code now doesn't support 1.0 for half floats but does support 0.0 which seems somewhat odd.

s-perron: What are suggesting?  We have `GetDoubleValue()` in the FloatContant class.  It is called below.  Or are you suggesting that `GetDoubleValue()` should return the value as a double regardless of the width of the constant?

zeux: The latter; maybe this requires a new function name.

zeux: (it might be worth doing this separately from this PR, to get assertion on half float folding fixed first)

dneto0: LLVM APFloat has:
  bool isZero() const;   ... and other special predicates.
  double convertToDouble() conts;
  bool isExactlyValue(double V) const;

source/opt/folding_rules.cpp
```diff
@@ -1638,8 +1638,12 @@ FloatConstantKind getFloatConstantKind(const analysis::Constant* constant) {
...
+
+    uint32_t width = fc->type()->AsFloat()->width();
+    if (width != 32 && width != 64) return FloatConstantKind::Unknown;
+
+    double value = (width == 64) ? fc->GetDoubleValue() : fc->GetFloatValue();
```
dneto0: Seems like this would be an opportunity for a new ConvertToDouble() method.

alan-baker: Improvements to handling half precision float can come later.

Rebased and submitted as 5f50e6209c815e93733f4d58906b86a50ff2af32.

PR #1383, opened 2018-03-07
Support SPIR-V 1.3 and Vulkan 1.1

dneto0:
This requires the latest SPIRV-Headers that supports SPIR-V 1.3.


dneto0:
The bots did pass on this PR.
The "x" failure shown above is due to cancellation of the bots when this CL landed on master;  I had immediately added yet another commit on top of it, so the "master" jobs for this CL were cancelled in favour of jobs for the very latest master.

PR #1386, opened 2018-03-07
Grab correct input to calculate indices.

### LINKED ISSUE #1385: SPIR-V optimizer breaks SMAA shaders
Latest SPIR-V optimizer breaks SMAA shaders. The blend weights pass shader produces broken output resulting in the final output being not antialiased and having nasty line artifacts instead.
2018.1 and latest master are broken.
Using default speed optimizations.
I've bisected it to e661da7941e12cbc6a9100bd8cadbc0d61dd8b7a.
[Good shader](https://gist.github.com/turol/7970f90b1fc3f3c8c6709a5c552b1101)
[Broken shader](https://gist.github.com/turol/3797e98dddfaae70385f38e78dc74626)

My [SMAA demo](https://github.com/turol/smaaDemo) can be used to test this.
It currently ships with the last working SPIRV-tools built in.
Both Vulkan and OpenGL versions are affected.
Beware that there appears to be a an unrelated problem with MSVC and OpenGL and that version doesn't work. Either use only Vulkan under MSVC, compile with mingw or use Linux.
Shader SPIR-V is cached in `~/.local/share/SMAADemo` on Linux and
`C:\Users\<username>\AppData\Roaming\SMAADemo` on Windows.
The breaking shader is smaaBlendWeight.frag_SMAA_PRESET_ULTRA=1.spv.
Running the latest optimizer on this shader and running the demo again can be used to test this.

If you want to compile a version with modified SPIRV-Tools, on Windows just update the source under foreign and recompile using the provided .bat. On Linux you can add the following to local.mk to use an external build of SPIRV-Tools
```
INTERNAL_spirv-tools:=n
LDLIBS_spirv-tools:=-lSPIRV-Tools -lSPIRV-Tools-opt
LDFLAGS+=-L/path/to/SPIRV-Tools/build/source
LDFLAGS+=-L/path/to/SPIRV-Tools/build/source/opt
CFLAGS+=-isystem/path/to/SPIRV-Tools/include
CFLAGS+=-isystem/path/to/SPIRV-Tools/build
```

alan-baker:
Fixes #1385 

* Used input to extract instead of input to vector shuffle
* Added tests to catch the bug

### DIFF
--- a/source/opt/folding_rules.cpp
+++ b/source/opt/folding_rules.cpp
@@ -1465,7 +1465,7 @@ FoldingRule VectorShuffleFeedingExtract() {
 
     // Find the size of the first vector operand of the VectorShuffle
     ir::Instruction* first_input =
-        def_use_mgr->GetDef(inst->GetSingleWordInOperand(0));
+        def_use_mgr->GetDef(cinst->GetSingleWordInOperand(0));
     analysis::Type* first_input_type =
         type_mgr->GetType(first_input->type_id());
     assert(first_input_type->AsVector() &&
--- a/test/opt/fold_test.cpp
+++ b/test/opt/fold_test.cpp
@@ -153,6 +153,7 @@ OpName %main "main"
 %_ptr_half = OpTypePointer Function %half
 %_ptr_long = OpTypePointer Function %long
 %_ptr_v2int = OpTypePointer Function %v2int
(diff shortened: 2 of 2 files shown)


alan-baker:
Rebased and submitted as bc9cfee6faa67fe570312aa889596b96165a9e7f.

PR #1389, opened 2018-03-08
Add missing SPV extension strings

amdrexu:


dneto0:
I've reviewed the semantics of these extensions.   They are compatible with the semantics of those transforms.

BuiltIns that are Input or Output variables are already mentioned in the EntryPoint interface.  So they won't be touched by mem2reg or dead code elimination. That takes care of many cases here.

Passes that operate combinatorially likely won't need a whitelist of extensions because we already have a whitelist of instructions.  I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/1392 to remove the extension whitelists from those.

I have filed validator issues to relax some rules:
-  https://github.com/KhronosGroup/SPIRV-Tools/issues/1394 to relax Lod checking for SPV_AMD_shader_image_load_store
- https://github.com/KhronosGroup/SPIRV-Tools/issues/1393 to allow SPV_AMD_gpu_shader_half_float_fetch to declare the float 16 type

amdrexu:
Thanks for your clarification.

Issue #1378, opened 2018-03-05
Add -ftime-report to spirv-opt


dnovillo:
Given

$ spirv-opt shader.spv -o my_shader.svp -ftime-report
Pass   CPU time      WALL time     SYS time     RSS
pass1  ...
pass2 ....

Each pass needs to keep track of its own resource utilization.  On pass creation, the pass manager tells the pass whether it should emit resource utilization at the end of the pass.  If resource utilization is enabled, the pass should print

- [ ] CPU time
- [ ] WALL time
- [ ] SYS time
- [ ] Memory utilized

In Linux, we can do this with getrusage().  In Windows?

@jaebaek, could you look into this?

dnovillo:
Fixed as of 3b594e163009cb224e80155ff17971a4b2fc0cd4

PR #1397, opened 2018-03-08
Add --time-report to spirv-opt

jaebaek:
Add --time-report to spirv-opt  …
To check the resource utilization of each optimization pass, I add --ftime-report command line option.

Related issue: #1378


s-perron:
Need to fix the build failures:

```
In file included from /home/travis/build/KhronosGroup/SPIRV-Tools/source/util/timer.cpp:20:0:
/home/travis/build/KhronosGroup/SPIRV-Tools/source/util/timer.h:70:60: error: #elif with no expression
 #elif  // defined(SPIRV_ANDROID) || defined(SPIRV_LINUX) ||
```

source/util/timer.cpp
```diff
@@ -0,0 +1,105 @@
...
+    if (getrusage(RUSAGE_SELF, &usage_after) == -1) {
+      usage_fail = kGetrusageFail;
+    } else if (gettimeofday(&wall_after, NULL) == -1) {
+      usage_fail = kGettimeofdayFail;
+    }
+  }
+}
+
+void Timer::Report(const char* tag) {
+  switch (usage_fail) {
```
s-perron: You should probably check if `report_stream_` is `nullptr` before the switch.

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2018 Google Inc.
```
s-perron: Google LLC

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
...
+
+#ifndef LIBSPIRV_UTIL_TIMER_H_
+#define LIBSPIRV_UTIL_TIMER_H_
+
+#if defined(SPIRV_ANDROID) || defined(SPIRV_LINUX) || defined(SPIRV_MAC) || \
+    defined(SPIRV_FREEBSD)
+
+#define SPIRV_TIMER_ENABLED
+#define SPIRV_TIMER_DESCRIPTION(out) spvutils::TimerPrintDescription(out)
+#define SPIRV_TIMER_SCROPED(out, tag) \
```
s-perron: What is "scroped"?  I think you meant "scoped".  This is in a lot of places.

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
...
+  Timer(std::ostream* out) : report_stream_(out), usage_fail(kNotFailed) {}
+  void Start();
+  void Stop();
+  void Report(const char* tag);
+  void StopAndReport(const char* tag);
+
+ private:
+  std::ostream* report_stream_;
+
+  enum { kGetrusageFail, kGettimeofdayFail, kNotFailed } usage_fail;
```
s-perron: Might be better to make `kNotFailed` more positive.  It will ready easier.  Try `kSucceeded` or `kPassed`.

tools/opt/opt.cpp
```diff
@@ -239,6 +239,10 @@ Options (in lexicographical order):
   --print-all
                Print SPIR-V assembly to standard error output before each pass
                and after the last pass.
+  -ftime-report
+               Print the resource utilization of each pass (e.g., CPU time,
+               RSS). Currently it supports only Unix systems. This option is
+               same with -ftime-report in GCC.
```
s-perron: is the same as -ftime-report in GCC.

s-perron: Should we mention that it will be output to stderr and not stdout?

source/util/timer.h
jaebaek: It is shameful :( So sorry.

source/util/timer.cpp
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) 2018 Google Inc.
```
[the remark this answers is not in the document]
jaebaek: Done

source/util/timer.cpp
```diff
@@ -0,0 +1,105 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <sys/resource.h>
+#include <sys/time.h>
+#include <iomanip>
+#include <iostream>
+
+#include "util/timer.h"
```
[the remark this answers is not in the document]
jaebaek: Done

source/util/timer.cpp
jaebaek: Done

source/util/timer.h
jaebaek: Done

source/util/timer.h
jaebaek: Done

tools/opt/opt.cpp
jaebaek: Yes, that's right. I updated it.

source/opt/optimizer.cpp
```diff
@@ -211,6 +211,11 @@ Optimizer& Optimizer::SetPrintAll(std::ostream* out) {
   return *this;
 }
 
+Optimizer& Optimizer::SetFTimeReport(std::ostream* out) {
+  impl_->pass_manager.SetFTimeReport(out);
```
dnovillo: Optimizer::SetTimeReport

source/CMakeLists.txt
```diff
@@ -263,6 +264,7 @@ set(SPIRV_SOURCES
   ${CMAKE_CURRENT_SOURCE_DIR}/util/bit_stream.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/util/parse_number.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/util/string_utils.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/util/timer.cpp
```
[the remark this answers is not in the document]
jaebaek: Done

source/util/timer.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+namespace spvutils {
+
+#if defined(SPIRV_TIMER_ENABLED)
+
+namespace {
+inline double TimeDifference(const timeval& before, const timeval& after) {
```
dneto0: Needs a method comment.
Also, what are the units of the result?

source/util/timer.h
dneto0: It's just a typo.  Don't worry about it..  :-)

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
...
+
+#if defined(SPIRV_LINUX)
+#define SPIRV_MEMORY_MEASUREMENT_ENABLED
+#endif  // defined(SPIRV_LINUX)
+
+#include <sys/resource.h>
+
+namespace spvutils {
+
+void TimerPrintDescription(std::ostream* out);
```
dneto0: What is being described?
This would be better with a method comment to explain.

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
...
+
+class Timer {
+ public:
+  Timer(std::ostream* out) : report_stream_(out), usage_fail(kNotFailed) {}
+  void Start();
```
dneto0: These need method comments.

What is being measured?  Total accumulated time between each matched pair of start and stop?  Or does Start() erase all previous state?

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
...
+  void Stop();
+  void Report(const char* tag);
+  void StopAndReport(const char* tag);
```
dneto0: I would remove the StopAndReport because it doesn't give much advantage.

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
...
+
+ private:
+  std::ostream* report_stream_;
```
dneto0: Need data member comments.

source/util/timer.h
dneto0: Please follow Google style guide for naming the type.
https://google.github.io/styleguide/cppguide.html#Type_Names 

So it would be UsageFail.

Also, since success is a possible value, I would probably call this UsageStatus

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
...
+
+  enum { kGetrusageFail, kGettimeofdayFail, kNotFailed } usage_fail;
+
+  rusage usage_before;
+  rusage usage_after;
+  timeval wall_before;
+  timeval wall_after;
+};
+
+class ScropedTimer : Timer {
```
dneto0: Need a class comment.  The class comment can be say enough that constructor and destructor comments might not be needed.

source/util/timer.cpp
```diff
@@ -0,0 +1,105 @@
...
+                    << (usage_after.ru_maxrss - usage_before.ru_maxrss)
+                    << std::setw(12)
+                    << ((usage_after.ru_minflt - usage_before.ru_minflt) +
+                        (usage_after.ru_majflt - usage_before.ru_majflt))
+#endif  // defined(SPIRV_MEMORY_MEASUREMENT_ENABLED)
+                    << std::endl;
+  }
+}
+
+void Timer::StopAndReport(const char* tag) {
```
dneto0: Yeah, this method does not give us enough extra convenience.  I would remove it.

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
...
+  void Stop();
+  void Report(const char* tag);
+  void StopAndReport(const char* tag);
+
+ private:
+  std::ostream* report_stream_;
+
+  enum { kGetrusageFail, kGettimeofdayFail, kNotFailed } usage_fail;
+
+  rusage usage_before;
```
dneto0: Please follow naming conventions for data members:  These should end with "_"

source/util/timer.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "util/timer.h"
+
+namespace spvutils {
+
+#if defined(SPIRV_TIMER_ENABLED)
+
+namespace {
+inline double TimeDifference(const timeval& before, const timeval& after) {
+  return static_cast<double>(after.tv_sec - before.tv_sec) +
+         static_cast<double>((after.tv_usec - before.tv_usec) / 10000) * .01;
```
dneto0: Why split the arithmetic with tv_usec?  It's suseconds_t and that is an integral type.

I would cast to double and then divide by a million.

source/util/timer.cpp
```diff
@@ -0,0 +1,105 @@
...
+      usage_fail = kGettimeofdayFail;
+    }
+  }
+}
+
+void Timer::Stop() {
+  if (report_stream_ && usage_fail == kNotFailed) {
+    if (getrusage(RUSAGE_SELF, &usage_after) == -1) {
+      usage_fail = kGetrusageFail;
+    } else if (gettimeofday(&wall_after, NULL) == -1) {
```
dneto0: Why do we need this fallback?

source/util/timer.h
```diff
@@ -0,0 +1,79 @@
...
+#define SPIRV_MEMORY_MEASUREMENT_ENABLED
+#endif  // defined(SPIRV_LINUX)
+
+#include <sys/resource.h>
+
+namespace spvutils {
+
+void TimerPrintDescription(std::ostream* out);
+
+class Timer {
```
dneto0: Needs a method comment.
What is this trying to measure?

There are lots of kinds of time that are relevant.
- elapsed wall clock time
- total CPU time (e.g. adding up all time spent by all threads, in case we ever multithread the code)
- calendar time

See http://man7.org/linux/man-pages/man7/time.7.html 

Should this really be called an IntervalTimer?

source/util/timer.cpp
```diff
@@ -0,0 +1,105 @@
...
+#endif  // defined(SPIRV_MEMORY_MEASUREMENT_ENABLED)
+         << std::endl;
+  }
+}
+
+void Timer::Start() {
+  if (report_stream_) {
+    if (getrusage(RUSAGE_SELF, &usage_before) == -1) {
+      usage_fail = kGetrusageFail;
+    } else if (gettimeofday(&wall_before, NULL) == -1) {
```
dneto0: This should not be in an else clause.   When you always report the time you will print the time-of-day usage.
But since this is an else clause you only measure the time-of-day if the getrusage failed.  So the data will be likely be invalid for the time-of-day report.

tools/opt/opt.cpp
jaebaek: Done

source/opt/optimizer.cpp
jaebaek: Done

source/util/timer.cpp
jaebaek: Done

source/util/timer.h
jaebaek: Done

source/util/timer.h
jaebaek: Done

source/util/timer.h
jaebaek: Done

source/util/timer.cpp
jaebaek: Done

source/util/timer.h
jaebaek: Done

source/util/timer.cpp
```diff
@@ -0,0 +1,106 @@
...
+#if defined(SPIRV_MEMORY_MEASUREMENT_ENABLED)
+         << std::setw(12) << "RSS" << std::setw(12) << "Pagefault"
+#endif  // defined(SPIRV_MEMORY_MEASUREMENT_ENABLED)
+         << std::endl;
+  }
+}
+
+void Timer::Start() {
+  if (report_stream_) {
+    if (getrusage(RUSAGE_SELF, &usage_before) == -1) {
```
atgoo: This is a low-res clock. Please use clock_gettime(CLOCK_PROCESS_CPUTIME_ID).

source/util/timer.cpp
```diff
@@ -0,0 +1,106 @@
...
+      usage_status = kGetrusageFail;
+    } else if (gettimeofday(&wall_before, NULL) == -1) {
```
atgoo: This is a non-monotonic clock. Please use clock_gettime(CLOCK_MONOTONIC).

source/util/timer.cpp
jaebaek: Sorry about this. Because of lots of review comments, I missed this. Let me update it.

source/util/timer.h
jaebaek: Done

source/util/timer.cpp
jaebaek: I wanted to show two digits after decimal point. I forgot that I already did it by adopting std::ios_base::precision. I updated it.

source/util/timer.cpp
jaebaek: I have one question .. getrusage() and times() give the information of both user and system cpu times. How can I get user cpu time and system cpu time using clock_gettime()?

source/util/timer.cpp
jaebaek: getrusage(), gettimeofday(), times(), and clock_gettime() returns -1 when they fail. This code set the usage status as failure if one of those calls fail. Otherwise, it proceeds with the next call.

source/util/timer.cpp
jaebaek: Done.

source/util/timer.cpp
atgoo: I don't think we need to have split user/system time measurements. Can spirv-opt actually generate a lot of system time (unless it's constantly swapping to disk due to low RAM)?

By the way, times() has resolution of 100Hz.

jaebaek: As far as I checked, it is not that big (e.g., 0.12 second when CPU time is 100.49). However, I think just leaving them is not worse. I will leave some comments to explain those times.

test/timer_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+TEST(Timer, Sleep) {
+  const unsigned int usleep_time = 71234;
+  const double epsilon = 0.001;
+  std::ostringstream buf;
+
+  PrintTimerDescription(&buf);
+  Timer timer(&buf);
+  timer.Start();
+
+  usleep(usleep_time);
```
atgoo: This kind of tests are usually done through mocking. For example you make a MockTimer class where functions return values which you define in the test.

source/util/timer.cpp
```diff
@@ -0,0 +1,133 @@
...
+      return;
+    case kClockGettimeFailed:
+      *report_stream_ << std::setw(30) << tag
+                      << " ERROR: calling clock_gettime() fails";
+      return;
+    default:
+      break;
+  }
+
+  report_stream_->precision(2);
```
atgoo: We have much higher precision for cpu and wall times. Smaller optimization passes could take tens of microseconds, outputting 0.019ms is better than 0.00s.

I recommend output in milliseconds with three digits after decimal point for cpu and wall times, and integral milliseconds for user and system time.

test/timer_test.cpp
jaebaek: Done.

source/util/timer.cpp
jaebaek: Done.

source/util/timer.h
```diff
@@ -0,0 +1,353 @@
...
+
+ private:
+  // Returns the time gap between |from| and |to| in seconds.
+  double TimeDifference(const timeval& from, const timeval& to) {
+    return static_cast<double>(to.tv_sec - from.tv_sec) +
+           static_cast<double>(to.tv_usec - from.tv_usec) * .000001;
+  }
+
+  // Returns the time gap between |from| and |to| in seconds.
+  double TimeDifference(const timespec& from, const timespec& to) {
```
atgoo: static

source/util/timer.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include <map>
+#include <string>
+
+namespace spvutils {
+
+#if defined(SPIRV_TIMER_ENABLED)
+
+// A map data structure to search a CumulativeTimer object by its name. The key
+// of the map is the name and the value is the CumulativeTimer object.
+static std::map<const char*, CumulativeTimer*> CumulativeTimerMap;
```
atgoo: This is not allowed by the style guide.

But parts of this project don't follow the style guide...

source/opt/pass_manager.cpp
```diff
@@ -41,8 +42,10 @@ Pass::Status PassManager::Run(ir::IRContext* context) {
     }
   };
 
+  SPIRV_TIMER_DESCRIPTION(time_report_stream_, true);
```
atgoo: Readability nit.

SPIRV_TIMER_DESCRIPTION(time_report_stream_, /* measure_mem_usage = */ true);

source/util/timer.h
```diff
@@ -0,0 +1,353 @@
...
+           (usage_after_.ru_majflt - usage_before_.ru_majflt);
+  }
+
+  UsageStatus GetUsageStatus() const { return usage_status_; }
+
+  virtual ~Timer() {}
+
+ private:
+  // Returns the time gap between |from| and |to| in seconds.
+  double TimeDifference(const timeval& from, const timeval& to) {
```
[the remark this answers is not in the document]
jaebaek: Done.

source/util/timer.h
```diff
@@ -0,0 +1,353 @@
...
+    return static_cast<double>(to.tv_sec - from.tv_sec) +
+           static_cast<double>(to.tv_usec - from.tv_usec) * .000001;
+  }
```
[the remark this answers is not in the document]
jaebaek: I added an assert() to check if |from| is later than |to|.

source/util/timer.h
jaebaek: Done

source/opt/pass_manager.cpp
jaebaek: Done.

source/util/timer.h
```diff
@@ -0,0 +1,353 @@
...
+  // Returns the cumulative number of page faults for a range of code execution.
+  long PageFault() const override { return pgfaults_; }
+
+  // Delete this CumulativeTimer object from |CumulativeTimerMap|.
+  ~CumulativeTimer() { DeleteCumulativeTimer(name_); }
+
+ private:
+  // Add an element pair to |CumulativeTimerMap| whose key is |name| and value
+  // is |ctimer|.
+  void SetCumulativeTimer(const char* name, CumulativeTimer* ctimer);
```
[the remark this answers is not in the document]
jaebaek: Done.

source/util/timer.cpp
jaebaek: Hmm .. I cannot understand how to change this into the correct way. If I replace it with an array of pairs, it is not clear whether the length of array must be fixed or not. If I want to not fixed one, it looks like an implementation of the data structure from scratch.

jaebaek: Thanks! I followed the first way (i.e., drop the map structure and functions related to it).

jaebaek:
Any comments for this PR?

source/util/timer.cpp
atgoo: Now it's outputting 0.000s.

I'm not strongly insisting on millisecond output, but it should be better than this:
```
                     PASS name    CPU time   WALL time    USR time    SYS time         RSS   Pagefault
inline-entry-points-exhaustive       0.001       0.001       0.000       0.000           0          13
   convert-local-access-chains       0.003       0.003       0.004       0.000         720          84
  eliminate-local-single-block       0.001       0.001       0.000       0.000           0           0
  eliminate-local-single-store       0.000       0.000       0.000       0.000           0           2
      eliminate-insert-extract       0.000       0.000       0.000       0.000           0           0
eliminate-dead-code-aggressive       0.004       0.004       0.004       0.000           0           0
       eliminate-dead-branches       0.000       0.000       0.000       0.000           0           0
                  merge-blocks       0.000       0.000       0.000       0.000           0           0
  eliminate-local-single-block       0.001       0.001       0.000       0.000           0           0
  eliminate-local-single-store       0.000       0.000       0.000       0.000           0           0
   eliminate-local-multi-store       0.000       0.000       0.000       0.000           0           2
      eliminate-insert-extract       0.000       0.000       0.000       0.000           0           0
eliminate-dead-code-aggressive       0.001       0.001       0.000       0.000           0           0
      eliminate-common-uniform       0.001       0.001       0.004       0.000           0           0
```

Maybe tweak the precision based on the measured CPU time? If measured 95 micros, output 0.000095. If measured 95 millis, output 0.095.

Also, note that if you are measuring a duration of 0.5ms, with 12.5% probability you would output:
CPU time: 0.000
WALL time: 0.000
USR time: 0.004

jaebaek: How about using scientific floating-point notation instead (http://www.cplusplus.com/reference/ios/scientific/)?

dnovillo:
No scientific output please. They are harder to parse.

source/util/timer.cpp
```diff
@@ -0,0 +1,103 @@
...
+// time correct as much as possible. Calling functions to get CPU/Wall time must
+// closely surround the target code of measuring.
+void Timer::Start() {
+  if (report_stream_) {
+    if (getrusage(RUSAGE_SELF, &usage_before_) == -1) {
+      usage_status_ = kClockGettimeFailed;
+    } else if (clock_gettime(CLOCK_MONOTONIC, &wall_before_) == -1) {
+      usage_status_ = kClockGettimeFailed;
+    } else if (clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &cpu_before_) == -1) {
+      usage_status_ = kGetrusageFailed;
```
dnovillo: Why 'else if'?  If a timer fails, the other timers will not be read.  I think these if() should not be linked with 'else'.

source/util/timer.cpp
```diff
@@ -0,0 +1,103 @@
...
+      return;
+    case kClockGettimeFailed:
+      *report_stream_ << std::setw(30) << tag
+                      << " ERROR: calling clock_gettime() fails";
+      return;
+    default:
+      break;
+  }
+
+  report_stream_->precision(3);
```
dnovillo: Precision 2, please.  Anything past the 3rd decimal does not really matter.

source/util/timer.h
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  // Returns the measured SYS Time for a range of code execution.
+  virtual double SystemTime() {
+    return TimeDifference(usage_before_.ru_stime, usage_after_.ru_stime);
+  }
+
+  // Returns the measured RSS for a range of code execution.
+  virtual long RSS() const {
+    return usage_after_.ru_maxrss - usage_before_.ru_maxrss;
```
dnovillo: Please state that this returns the RSS delta since the pass started.  Label the column as 'RSS delta' too.  Otherwise, it can be confusing.

source/util/timer.h
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  // Returns the measured number of page faults for a range of code execution.
+  virtual long PageFault() const {
+    return (usage_after_.ru_minflt - usage_before_.ru_minflt) +
+           (usage_after_.ru_majflt - usage_before_.ru_majflt);
+  }
```
dnovillo: Similarly, here.

tools/opt/opt.cpp
```diff
@@ -239,6 +239,13 @@ Options (in lexicographical order):
   --print-all
                Print SPIR-V assembly to standard error output before each pass
                and after the last pass.
+  --time-report
```
dnovillo: Please move this down to preserve lexicographical order of the flag names.

source/util/timer.cpp
```diff
@@ -0,0 +1,103 @@
...
+namespace spvutils {
+
+#if defined(SPIRV_TIMER_ENABLED)
+
+// Print the description of resource types measured by Timer class. If |out| is
+// NULL, it does nothing. Otherwise, it prints resource types. The second is
+// optional and if it is true, the function also prints resource type fields
+// related to memory. Its default is false. In usual, this must be placed before
+// calling Timer::Report() to inform what those fields printed by
+// Timer::Report() indicate.
```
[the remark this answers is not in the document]
jaebaek: Done

source/util/timer.cpp
jaebaek: Done

source/util/timer.h
jaebaek: Done

source/util/timer.h
jaebaek: Done

tools/opt/opt.cpp
jaebaek: Done

source/util/timer.cpp
jaebaek: Hmm .. the logic here is
1. if some function fails (i.e., the return value is equal to -1), stop and set the error status.
2. else call some other function.

It should be

if (some_function(...) == -1) {
  // it fails ..
  /* report_error */
} else ...

I do not think it is incorrect .. but it would be confusing.

I will change this as
if (some_function(...) == -1) {
  usage_status_ == error;
  return;
}
/* call other functions */

dnovillo:
On Tue, Mar 20, 2018 at 11:02 AM Jaebaek Seo <notifications@github.com>
wrote:

> *@jaebaek* commented on this pull request.
>
> In source/util/timer.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1397#discussion_r175868476>
> :
>
> > +    }
> +    *out << std::endl;
> +  }
> +}
> +
> +// Do not change the order of invoking system calls. We want to make CPU/Wall
> +// time correct as much as possible. Calling functions to get CPU/Wall time must
> +// closely surround the target code of measuring.
> +void Timer::Start() {
> +  if (report_stream_) {
> +    if (getrusage(RUSAGE_SELF, &usage_before_) == -1) {
> +      usage_status_ = kClockGettimeFailed;
> +    } else if (clock_gettime(CLOCK_MONOTONIC, &wall_before_) == -1) {
> +      usage_status_ = kClockGettimeFailed;
> +    } else if (clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &cpu_before_) == -1) {
> +      usage_status_ = kGetrusageFailed;
>
> Hmm .. the logic here is
>
>    1. if some function fails (i.e., the return value is equal to -1),
>    stop and set the error status.
>    2. else call some other function.
>
> It should be
>
> if (some_function(...) == -1) {
> // it fails ..
> /* report_error */
> } else ...
>
> I do not think it is incorrect .. but it would be confusing.
>
> I will change this as
> if (some_function(...) == -1) {
> usage_status_ == error;
> return;
> }
> /* call other functions */
>

But, this is what's confusing.  Why return if any of these functions fail?
The others may not fail.

In the case of failure, the timer will print nothing at all?

>

jaebaek:
Ah .. yes. Currently, it prints only when no failure exists.
I will update it so that it prints each element if it does not fail.

2018-03-20 14:06 GMT-04:00 Diego Novillo <notifications@github.com>:

source/util/timer.cpp
jaebaek: Hmm .. it seems that your email is not shown here.

Currently, it prints only when no failure exists.
I will update it so that it prints each type of resource usage if it does not fail.

dnovillo:
On Tue, Mar 20, 2018 at 11:11 AM Jaebaek Seo <notifications@github.com>
wrote:

> Ah .. yes. Currently, it prints only when no failure exists.
> I will update it so that it prints each element if it does not fail.
>

SG.  Thanks.

>

dnovillo:
On Tue, Mar 20, 2018 at 11:12 AM Jaebaek Seo <notifications@github.com>
wrote:

> *@jaebaek* commented on this pull request.
>
> In source/util/timer.cpp
> <https://github.com/KhronosGroup/SPIRV-Tools/pull/1397#discussion_r175871783>
> :
>
> > +    }
> +    *out << std::endl;
> +  }
> +}
> +
> +// Do not change the order of invoking system calls. We want to make CPU/Wall
> +// time correct as much as possible. Calling functions to get CPU/Wall time must
> +// closely surround the target code of measuring.
> +void Timer::Start() {
> +  if (report_stream_) {
> +    if (getrusage(RUSAGE_SELF, &usage_before_) == -1) {
> +      usage_status_ = kClockGettimeFailed;
> +    } else if (clock_gettime(CLOCK_MONOTONIC, &wall_before_) == -1) {
> +      usage_status_ = kClockGettimeFailed;
> +    } else if (clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &cpu_before_) == -1) {
> +      usage_status_ = kGetrusageFailed;
>
> Hmm .. it seems that your email is not shown here.
>
> Currently, it prints only when no failure exists.
> I will update it so that it prints each type of resource usage if it does
> not fail.
>
Yeah, I'm responding from email.  github's mail gateway is not always
reliable.

>

dnovillo:
Rebased, retested and pushed as 3b594e163009cb224e80155ff17971a4b2fc0cd4

@atgoo, I think the latest version should address all your concerns. If that's not the case, please let @jaebaek know so he can address it in a follow-up PR.  Thanks.

PR #1411, opened 2018-03-13
Don't merge non-branch terminators into loop header.

### LINKED ISSUE #1402: OpLoopMerge is generated without a loop causing validation error
Before: 
[before.asm.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/1802727/before.asm.txt)
After `spirv-opt -O`:
[after.asm.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/1802729/after.asm.txt)

Error in spirv-val:
```
error: 73: OpLoopMerge must immediately precede either an OpBranch or OpBranchConditional instruction. OpLoopMerge must be the second-to-last instruction in its block.
```

alan-baker:
Fixes #1402

Added tests. Confirmed on testcase in #1402.

### DIFF
--- a/source/opt/block_merge_pass.cpp
+++ b/source/opt/block_merge_pass.cpp
@@ -69,8 +69,24 @@ bool BlockMergePass::MergeBlocks(ir::Function* func) {
       continue;
     }
 
-    // Merge blocks.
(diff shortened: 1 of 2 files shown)


dneto0:
Rebased and pushed onto master as 7e03e76a5f9

PR #1413, opened 2018-03-14
Migrate to use unified grammar tables

antiagainst:
Previously we keep a separate static grammar table for opcodes/
operands per SPIR-V version. This commit changes that to use a
single unified static grammar table for opcodes/operands.

This essentially changes how grammar facts are queried against
a certain target environment. There are only limited filtering
according to the desired target environment; a symbol is
considered as available as long as:

1. The target environment satisfies the minimal requirement of
   the symbol; or
2. There is at least one extension enabling this symbol.

Note that the second rule assumes the extension enabling the
symbol is indeed requested in the SPIR-V code; checking that
should be the validator's work.

Also fixed a few grammar related issues:
* Rounding mode capability requirements are moved to client APIs.
* Reserved symbols not available in any extension is no longer
  recognized by assembler.


antiagainst:
This pull request requires https://github.com/KhronosGroup/SPIRV-Headers/pull/62, so it is expected to fail bots ATM.

include/spirv-tools/libspirv.h
```diff
@@ -73,6 +73,7 @@ typedef enum spv_result_t {
   SPV_ERROR_INVALID_CAPABILITY = -13,
   SPV_ERROR_INVALID_DATA = -14,  // Indicates data rules validation failure.
   SPV_ERROR_MISSING_EXTENSION = -15,
+  SPV_ERROR_WRONG_VERSION = -16, // Indicates wrong SPIR-V version
```
antiagainst: I'm not sure about whether to introduce this new error type. It just seems there is no suitable existing ones.

utils/generate_grammar_tables.py
```diff
@@ -77,6 +77,16 @@ def make_path_to_file(f):
             raise
 
 
+def convert_min_required_version(version):
+    """Converts the minimal required SPIR-V version encoded in the
+    grammar to the symbol in SPIRV-Tools"""
+    if version is None:
+        return 'SPV_ENV_UNIVERSAL_1_0'
+    if version == 'None':
+        return '0xffffffffu'
```
antiagainst: I'm not sure about creating `SPV_ENV_UNIVERSAL_infinity` because it will be available to API consumers and we need do sanity check for it. But I don't have strong preferences. I can change if you have.

antiagainst:
With this, `libSPIRV-Tools.a`'s size drops from 1.9M to 1.6M.

source/opcode.cpp
```diff
@@ -89,85 +75,87 @@ void spvOpcodeSplit(const uint32_t word, uint16_t* pWordCount,
...
 
-spv_result_t spvOpcodeTableGet(spv_opcode_table* pInstTable,
-                               spv_target_env env) {
+spv_result_t spvOpcodeTableGet(spv_opcode_table* pInstTable, spv_target_env) {
   if (!pInstTable) return SPV_ERROR_INVALID_POINTER;
 
   // Descriptions of each opcode.  Each entry describes the format of the
   // instruction that follows a particular opcode.
 
-  switch (env) {
```
dneto0: I am really glad to never have to modify this again.  !

source/validate_instruction.cpp
```diff
@@ -127,22 +126,35 @@ CapabilitySet RequiredCapabilities(const ValidationState_t& state,
...
     // Allow FPRoundingMode decoration if requested.
-    if (state.features().free_fp_rounding_mode &&
-        type == SPV_OPERAND_TYPE_DECORATION &&
+    if (type == SPV_OPERAND_TYPE_DECORATION &&
         operand_desc->value == SpvDecorationFPRoundingMode) {
-      return CapabilitySet();
+      if (state.features().free_fp_rounding_mode) return CapabilitySet();
+
+      // Vulkan API requires more capabilities on rounding mode.
+      switch (state.context()->target_env) {
```
dneto0: Use spvIsVulkanEnv from source/spirv_target_env.h

utils/generate_grammar_tables.py
dneto0: This seems fine.

source/validate_instruction.cpp
antiagainst: Done

dneto0:
Rebased, squashed, and pushed into master as 1ef6b192602

PR #1415, opened 2018-03-21
Fix VS2013 build failures.

dnovillo:
VS 2013 does not accept defaulting the move constructor.  This fixes the build problems with it.

### DIFF
--- a/source/opt/ssa_rewrite_pass.h
+++ b/source/opt/ssa_rewrite_pass.h
@@ -57,10 +57,6 @@ class SSARewriter {
           is_complete_(false),
           users_() {}
 
-    PhiCandidate(const PhiCandidate&) = delete;
-    PhiCandidate(PhiCandidate&&) = default;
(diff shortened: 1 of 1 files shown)


source/opt/ssa_rewrite_pass.h
```diff
@@ -57,10 +57,6 @@ class SSARewriter {
           is_complete_(false),
           users_() {}
 
-    PhiCandidate(const PhiCandidate&) = delete;
-    PhiCandidate(PhiCandidate&&) = default;
```
s-perron: You can't have a default move constructor in vs2013, but you can write one if you think it is needed.  It could be useful because of the two std::vector members.  However, I don't think it will have a bug performance impact, so this is fine for now.

dnovillo:
Pushed upstream as 2e644e45785bb221294c32bf02a4ac867de49dc4

PR #1417, opened 2018-03-21
Fixes the compile errors generated by SPIRV_WARN_EVERYTHING and SPIRV_WERROR

hikiko:
This patch fixes the compile errors generated when the options
SPIRV_WARN_EVERYTHING and SPIRV_WERROR (that force -Wpedantic) are
set to CMake.

### DIFF
--- a/source/opt/types.h
+++ b/source/opt/types.h
@@ -517,15 +517,15 @@ class ForwardPointer : public Type {
                                                                      \
     void GetExtraHashWords(std::vector<uint32_t>*) const override {} \
   };  // namespace analysis
-DefineParameterlessType(Void, void);
-DefineParameterlessType(Bool, bool);
-DefineParameterlessType(Sampler, sampler);
-DefineParameterlessType(Event, event);
-DefineParameterlessType(DeviceEvent, device_event);
-DefineParameterlessType(ReserveId, reserve_id);
-DefineParameterlessType(Queue, queue);
-DefineParameterlessType(PipeStorage, pipe_storage);
-DefineParameterlessType(NamedBarrier, named_barrier);
+DefineParameterlessType(Void, void)
+DefineParameterlessType(Bool, bool)
+DefineParameterlessType(Sampler, sampler)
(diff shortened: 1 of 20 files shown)


source/opt/types.h
```diff
@@ -517,15 +517,15 @@ class ForwardPointer : public Type {
...
-DefineParameterlessType(Void, void);
-DefineParameterlessType(Bool, bool);
-DefineParameterlessType(Sampler, sampler);
-DefineParameterlessType(Event, event);
-DefineParameterlessType(DeviceEvent, device_event);
-DefineParameterlessType(ReserveId, reserve_id);
-DefineParameterlessType(Queue, queue);
-DefineParameterlessType(PipeStorage, pipe_storage);
-DefineParameterlessType(NamedBarrier, named_barrier);
+DefineParameterlessType(Void, void)
```
s-perron: I think the better fix here is to remove the semi-colon from the macro definition.  Also the comment at the end of the macro "namespace analysis" should be removed.  It seems wrong.

s-perron:
Clang-format is still not working.  I have heard that the version you use matters, also make sure you are using the .clang-format in the source tree.

In any case, this is the section the bot is complaining about:

```
--- source/opt/inline_exhaustive_pass.cpp	(before formatting)
+++ source/opt/inline_exhaustive_pass.cpp	(after formatting)
@@ -59,9 +59,7 @@
   return modified;
 }
 
-void InlineExhaustivePass::Initialize(ir::IRContext* c) {
-  InitializeInline(c);
-}
+void InlineExhaustivePass::Initialize(ir::IRContext* c) { InitializeInline(c); }
 
 Pass::Status InlineExhaustivePass::ProcessImpl() {
   // Attempt exhaustive inlining on each entry point function in module
```

s-perron:
The formatting is still off.  I'll fix it up myself, and merge.

s-perron:
Rebased, squashed, and merged as 045cc8f7

PR #1380, opened 2018-03-06
Change merge return pass to handle structured cfg.

s-perron:
We are seeing shaders that have multiple returns in a functions.  These
functions must get inlined for legalization purposes; however, the
inliner does not know how to inline functions that have multiple
returns.

The solution we will go with it to improve the merge return pass to
handle structured control flow.

Fixes #857.

PR #1419, opened 2018-03-21
Fixed remaining review comments from #1380

s-perron:


source/opt/def_use_manager.cpp
```diff
@@ -271,15 +271,11 @@ bool operator==(const DefUseManager& lhs, const DefUseManager& rhs) {
...
       if (rhs.id_to_users_.count(p) == 0) {
-        std::cerr << p.first->PrettyPrint() << std::endl;
-        std::cerr << p.second->PrettyPrint() << std::endl;
         return false;
       }
     }
     for (auto p : rhs.id_to_users_) {
       if (lhs.id_to_users_.count(p) == 0) {
-        std::cerr << p.first->PrettyPrint() << std::endl;
-        std::cerr << p.second->PrettyPrint() << std::endl;
```
dnovillo: Perhaps ifdef for when we're debugging this in the future? Up to you.

PR #1420, opened 2018-03-23
Adding OpPhi validation rules.

alan-baker:
Checks:
* phi type is a type instruction
* incoming value types match result type
* number of incoming blocks matches CFG predecessors
* matched pairs of values and blocks
* incoming blocks are CFG predecessors

Changes:
* Added tests
* Fixes SSA check for unreachable phi parents
* Fixes invalid cfg cleanup test

I tried to match the style of the file, including diagnostic messages.

### DIFF
--- a/source/validate_id.cpp
+++ b/source/validate_id.cpp
@@ -1786,11 +1786,72 @@ bool idUsage::isValid<SpvOpVectorShuffle>(const spv_instruction_t* inst,
   return true;
 }
 
-#if 0
 template <>
-bool idUsage::isValid<OpPhi>(const spv_instruction_t *inst,
-                             const spv_opcode_desc opcodeEntry) {}
-#endif
+bool idUsage::isValid<SpvOpPhi>(const spv_instruction_t* inst,
+                                const spv_opcode_desc opcodeEntry) {
+  (void)opcodeEntry;
+  auto thisInst = module_.FindDef(inst->words[2]);
+  SpvOp typeOp = module_.GetIdOpcode(thisInst->type_id());
+  if (!spvOpcodeGeneratesType(typeOp)) {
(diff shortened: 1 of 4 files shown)


source/validate_id.cpp
```diff
@@ -1786,11 +1786,72 @@ bool idUsage::isValid<SpvOpVectorShuffle>(const spv_instruction_t* inst,
...
+      auto resType = inst->words[1];
+      if (resType != incType->id()) {
+        DIAG(i) << "OpPhi's result type <id> " << module_.getIdName(resType)
+                << " does not match incoming value <id> "
+                << module_.getIdName(incId) << " type <id> "
+                << module_.getIdName(incType->id()) << ".";
+        return false;
+      }
+    } else {
+      // Incoming basic block must be an immediate predecessor of the phi's
```
atgoo: Nit: consider also a more targeted error message if
if (SpvOpLabel != module_.GetIdOpcode(incId))

source/validate_id.cpp
```diff
@@ -2606,7 +2671,8 @@ spv_result_t CheckIdDefinitionDominateUse(const ValidationState_t& _) {
       const Instruction* variable = _.FindDef(phi->word(i));
       const BasicBlock* parent =
           phi->function()->GetBlock(phi->word(i + 1)).first;
-      if (variable->block() && !variable->block()->dominates(*parent)) {
+      if (variable->block() && parent->reachable() &&
```
atgoo: Is it valid for a parent block to be not reachable (I see that this is the case in some of tests)?

Also this makes the comment above not accurate (// Check all OpPhi parent blocks are dominated by the variable...).

alan-baker:
Rebased, squashed and committed as 97c8fdccd2daf824ee030b6371ed521de0ac0496

PR #1421, opened 2018-03-23
Copy propagate arrays

s-perron:
The sprir-v generated from HLSL code contain many copyies of very large
arrays.  Not only are these time consumming, but they also cause
problems for drivers because they require too much space.

To work around this, we will implement an array copy propagation.  Note
that we will not implement a complete array data flow analysis in order
to implement this.  We will be looking for very simple cases:

1) The source must never be stored to.
2) The target must be stored to exactly once.
3) The store to the target must be a store to the entire array, and be a
copy of the entire source.
4) All loads of the target must be dominated by the store.

The hard part is keeping all of the types correct.  We do not want to
have to do too large a search to update everything, which may not be
possible, do we give up if we see any instruction that might be hard to
update.

Also in types.h, the element decorations are not stored in an std::map.
This change was done so the hashing algorithm for a Struct is
consistent.  With the std::unordered_map, the traversal order was
non-deterministic leading to the same type getting hashed to different
values.  See |Struct::GetExtraHashWords|.

### DIFF
--- a/source/opt/copy_prop_arrays.cpp
+++ b/source/opt/copy_prop_arrays.cpp
@@ -175,8 +175,8 @@ bool CopyPropagateArrays::HasValidReferencesOnly(ir::Instruction* ptr_inst,
       context()->GetDominatorAnalysis(store_block->GetParent(), *cfg());
 
   return get_def_use_mgr()->WhileEachUser(
-      ptr_inst, [this, store_inst, store_block, dominator_analysis,
-                 ptr_inst](ir::Instruction* use) {
+      ptr_inst,
+      [this, store_inst, dominator_analysis, ptr_inst](ir::Instruction* use) {
         if (use->opcode() == SpvOpLoad) {
           return dominator_analysis->Dominates(store_inst, use);
         } else if (use->opcode() == SpvOpAccessChain) {
@@ -338,8 +338,7 @@ bool CopyPropagateArrays::CanUpdateUses(ir::Instruction* original_ptr_inst,
   analysis::ConstantManager* const_mgr = context()->get_constant_mgr();
   analysis::DefUseManager* def_use_mgr = context()->get_def_use_mgr();
 
-  return def_use_mgr->WhileEachUse(original_ptr_inst, [this, original_ptr_inst,
-                                                       type_id, type_mgr,
+  return def_use_mgr->WhileEachUse(original_ptr_inst, [this, type_id, type_mgr,
                                                        const_mgr](
                                                           ir::Instruction* use,
                                                           uint32_t index) {
@@ -431,8 +430,7 @@ void CopyPropagateArrays::UpdateUses(ir::Instruction* original_ptr_inst,
   analysis::ConstantManager* const_mgr = context()->get_constant_mgr();
   analysis::DefUseManager* def_use_mgr = context()->get_def_use_mgr();
 
-  def_use_mgr->ForEachUse(original_ptr_inst, [this, original_ptr_inst,
-                                              new_ptr_inst, type_mgr,
+  def_use_mgr->ForEachUse(original_ptr_inst, [this, new_ptr_inst, type_mgr,
                                               const_mgr](ir::Instruction* use,
                                                          uint32_t index) {
     analysis::Pointer* pointer_type = nullptr;
--- a/source/opt/copy_prop_arrays.cpp
+++ b/source/opt/copy_prop_arrays.cpp
@@ -103,12 +103,12 @@ CopyPropagateArrays::FindSourceObjectIfPossible(ir::Instruction* var_inst) {
     return nullptr;
   }
   return source;
-}  // namespace opt
+}
 
 void CopyPropagateArrays::PropagateObject(ir::Instruction* var_inst,
                                           MemoryObject* source) {
   assert(var_inst->opcode() == SpvOpVariable &&
-         "This funciton propagtages variables.");
+         "This function propagates variables.");
 
   ir::Instruction* insertion_point = var_inst->NextNode();
   while (insertion_point->opcode() == SpvOpVariable) {
@@ -178,6 +178,9 @@ bool CopyPropagateArrays::HasValidReferencesOnly(ir::Instruction* ptr_inst,
       ptr_inst,
       [this, store_inst, dominator_analysis, ptr_inst](ir::Instruction* use) {
         if (use->opcode() == SpvOpLoad) {
+          // TODO: If there are many load in the same BB as |store_inst| the
+          // time to do the multiple traverses can add up.  Consider collecting
+          // those loads and doing a single traversal.
           return dominator_analysis->Dominates(store_inst, use);
         } else if (use->opcode() == SpvOpAccessChain) {
           return HasValidReferencesOnly(use, store_inst);
@@ -185,7 +188,9 @@ bool CopyPropagateArrays::HasValidReferencesOnly(ir::Instruction* ptr_inst,
           return true;
         } else if (use->opcode() == SpvOpStore) {
           // If we are storing to part of the object it is not an candidate.
-          return ptr_inst->opcode() == SpvOpVariable;
+          return ptr_inst->opcode() == SpvOpVariable &&
+                 store_inst->GetSingleWordInOperand(kStorePointerInOperand) ==
+                     ptr_inst->result_id();
         }
         // Some other instruction.  Be conservative.
         return false;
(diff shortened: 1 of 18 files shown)


s-perron:
I still have to create unit test.

source/opt/copy_prop_arrays.cpp
```diff
@@ -0,0 +1,562 @@
...
+        // Some other instruction.  Be conservative.
+        return false;
+      });
+}
+
+bool CopyPropagateArrays::HasValidReferencesOnly(ir::Instruction* ptr_inst,
+                                                 ir::Instruction* store_inst) {
+  ir::BasicBlock* store_block = context()->get_instr_block(store_inst);
+  opt::DominatorAnalysis* dominator_analysis =
+      context()->GetDominatorAnalysis(store_block->GetParent(), *cfg());
```
dnovillo: Optional.  Move this to a class instance variable and call it once per function?  Saves the constant lookup in GetDominatorAnalysis.  OTOH, we may not even need to compute dominance if there are no copy-prop opportunities, so... *shrug*.

source/opt/copy_prop_arrays.cpp
```diff
@@ -0,0 +1,562 @@
...
+  std::vector<uint32_t> components_in_reverse;
+  analysis::DefUseManager* def_use_mgr = context()->get_def_use_mgr();
+  analysis::ConstantManager* const_mgr = context()->get_constant_mgr();
+
+  ir::Instruction* current_inst = def_use_mgr->GetDef(
+      load_inst->GetSingleWordInOperand(kLoadPointerInOperand));
+
+  // Build the access chain for the memory object by collecting the indices used
+  // in the OpAccessChain instructions.  If we find a variable index, then
+  // return |nullptr| because we cannot know for use which memory location is
```
dnovillo: s/for use/for sure/ ?

source/opt/copy_prop_arrays.cpp
```diff
@@ -0,0 +1,562 @@
...
+        return nullptr;
+      }
+      assert(element_index_const->AsIntConstant());
+      components_in_reverse.push_back(
+          element_index_const->AsIntConstant()->GetU32());
+    }
+    current_inst = def_use_mgr->GetDef(current_inst->GetSingleWordInOperand(0));
+  }
+
+  // If the address in the load is not construction from an |OpVariable|
```
dnovillo: s/construction/constructed/

source/opt/copy_prop_arrays.cpp
```diff
@@ -0,0 +1,562 @@
...
+        const analysis::Type* pointee_type = pointer_type->pointee_type();
+
+        std::vector<uint32_t> access_chain;
+        for (uint32_t i = 1; i < use->NumInOperands(); ++i) {
+          const analysis::Constant* index_const =
+              const_mgr->FindDeclaredConstant(use->GetSingleWordInOperand(i));
+          if (index_const) {
+            access_chain.push_back(index_const->AsIntConstant()->GetU32());
+          } else {
+            // Variable index means the type is an type where every element
```
dnovillo: s/an type/a type/

s-perron:
Created some test.  Still want to create test where the elements in the arrays are struct containing other arrays.

source/opt/copy_prop_arrays.h
```diff
@@ -0,0 +1,191 @@
...
+   public:
+    // Construction a memory object that is owned by |var_inst|.  The iterator
+    // |begin| and |end| traverse a container of integers that identify which
+    // member of |var_inst| this memory object will represent.  These integers
+    // are interpreted the same way they would be in an |OpCompositeExtract|
+    // instruction.
+    template <class iterator>
+    MemoryObject(ir::Instruction* var_inst, iterator begin, iterator end);
+
+    // Change |this| to now point to the member identify by |access_chain|
```
dnovillo: s/identify/identified/

source/opt/copy_prop_arrays.h
```diff
@@ -0,0 +1,191 @@
...
+    // (starting from the current member).  The elements in |access_chain| are
+    // interpreted the same as the indicies in the |OpCompositeExtract|
+    // instruction.
+    void GetMember(const std::vector<uint32_t>& access_chain);
+
+    // Change |this| to now represent the first enclosing object to which it
+    // belongs.  (Remove the last element off the access_chain). It is invalid
+    // to call this function if |this| does not represent a member its owner.
```
dnovillo: "of its owner"?

source/opt/copy_prop_arrays.h
```diff
@@ -0,0 +1,191 @@
...
+    // Returns the owning variable that the memory object is contained in.
+    ir::Instruction* GetVariable() const { return variable_inst_; }
+
+    // Returns a vector of integers that can be used to access the specific
+    // member that |this| represents starting from the owning variable.  These
+    // values are to be interpreted the same way the indicies are in an
+    // |OpCompositeExtract| instruction.
+    const std::vector<uint32_t>& AccessChain() const { return access_chain_; }
+
+    // Returns the type id for a pointer to the memory object.
```
dnovillo: Can you describe what this function is doing?  At first, it seemed like it was similar to GetPointeeTypeId(), but it isn't and it's not immediately clear what it's trying to do.

source/opt/copy_prop_arrays.h
```diff
@@ -0,0 +1,191 @@
...
+      const analysis::Type* member_type =
+          type_mgr->GetMemberType(var_type, AccessChain());
+      uint32_t member_type_id = type_mgr->GetId(member_type);
+      assert(member_type != 0);
+      uint32_t member_pointer_type_id = type_mgr->FindPointerToType(
+          member_type_id, pointer_type->storage_class());
+      return member_pointer_type_id;
+    }
+
+    SpvStorageClass GetStorageClass() const {
```
dnovillo: Similar confusion here and with GetStorageType.

source/opt/dominator_analysis.cpp
```diff
@@ -37,5 +39,35 @@ ir::BasicBlock* DominatorAnalysisBase::CommonDominator(
...
+  for (ir::Instruction& inst : *bb_a) {
+    if (&inst == a) {
+      return true;
+    } else if (&inst == b) {
+      return false;
+    }
+  }
+  assert(false &&
+         "We did not find the load or store in the block they are "
+         "suppose to be in.");
```
dnovillo: s/suppose/supposed/

source/opt/dominator_analysis.cpp
```diff
@@ -37,5 +39,35 @@ ir::BasicBlock* DominatorAnalysisBase::CommonDominator(
...
+    return tree_.Dominates(bb_a, bb_b);
+  }
+
+  for (ir::Instruction& inst : *bb_a) {
+    if (&inst == a) {
+      return true;
+    } else if (&inst == b) {
+      return false;
+    }
+  }
```
dnovillo: If only one of the instructions has the wrong basic block information, this will silently return true/false.  Perhaps, traverse the whole block always?

source/opt/types.h
```diff
@@ -369,7 +370,7 @@ class Struct : public Type {
   bool decoration_empty() const override {
     return decorations_.empty() && element_decorations_.empty();
   }
-  const std::unordered_map<uint32_t, std::vector<std::vector<uint32_t>>>&
+  const std::map<uint32_t, std::vector<std::vector<uint32_t>>>&
```
dnovillo: Please document that we need an ordered map here for stable traversals.

source/opt/copy_prop_arrays.cpp
s-perron: I'll leave this as it for now.  Because it is just a lookup.  We could do a local cache or something, but not worth the complication for now.

source/opt/dominator_analysis.cpp
s-perron: In general, I do not want to do extra work.  I want to assert when I hit an know problem.  If you really want, I can add a loop that is guarded by an ifdef that can do this work.

source/opt/types.h
s-perron: Added the comment on the declaration of the member variable.

source/opt/copy_prop_arrays.cpp
```diff
@@ -0,0 +1,562 @@
...
+          return true;
+        } else if (use->opcode() == SpvOpStore) {
+          return false;
+        }
+        // Some other instruction.  Be conservative.
+        return false;
+      });
+}
+
+bool CopyPropagateArrays::HasValidReferencesOnly(ir::Instruction* ptr_inst,
```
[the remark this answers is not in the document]
s-perron: I'll add a TODO for this.  It could be bad if we have lots of loads  in the same basic block.  It may not happen.  The other TODOs are more important.  I'll get to it after those.

source/opt/dominator_analysis.cpp
dnovillo: Actually, I think it's fine as is.  This code is guarded with the predicate bb_a != bb_b, so if get_instr_block() is lying, we have more problems that will lead to other problems elsewhere.

s-perron:
Rebased, squashed and merged as c4dc046.

PR #1422, opened 2018-03-23
Don't produce validation false negatives for spec constant sized arrays

### LINKED ISSUE #1403: spirv-val gives false positive on spec constant array sizes
[array-size-spec-constant.asm.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/1802843/array-size-spec-constant.asm.txt)

Getting validation error:
```
error: 72: CompositeExtract: array access is out of bounds, array size is 0, but access index is 0
```

alan-baker:
Fixes #1403

Don't validate composite insert, extract and construct instructions
against spec constant sized arrays.
* Added predicate for spec constant opcodes
* Added tests

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -206,6 +206,19 @@ int32_t spvOpcodeIsScalarType(const SpvOp opcode) {
   }
 }
 
+int32_t spvOpcodeIsSpecConstant(const SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpSpecConstantTrue:
+    case SpvOpSpecConstantFalse:
+    case SpvOpSpecConstant:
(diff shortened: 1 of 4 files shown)


alan-baker:
I did not see that in the spec, perhaps @dneto0 can clarify.

dneto0:
The validator is running in a context where we don't know if the default values are going to be overridden at a later stage, e.g. pipeline creation time, or a further transform.

@atgoo : thanks for pointing out the other spots where this kind of assumption is made.

alan-baker:
Rebased and committed as 0a2ee65f57726ad8f7c24420bdb7166e5b1e4ba3. Filed #1434 to follow up additional assumptions.

PR #1424, opened 2018-03-23
Add missing function parameters in libspirv.h

cubanismo:
When building C code with gcc and the
-Wstrict-prototypes option, function declarations
and definitions that don't specify their argument
types generate warnings.  Functions that don't
take parameters need to specify (void) as their
parameter list, rather than leaving it empty.

Note this only applies to C, so only the functions
exported in C-compatible headers need fixing.  In
C++ functions can't be declared/defined without a
parameter list, so C++ can safely allow an empty
parameter list to imply (void).


cubanismo:
The contribution should be covered by some broader agreement between NVIDIA and Khronos from what I remember.  I don't think I'm supposed to sign the CLA myself.

cubanismo:
OK, after far too much back and forth with lawyers internally, I can click through the CLA here, so this should be good to go.

PR #1426, opened 2018-03-26
Handle more cases in copy propagate arrays.

s-perron:
When we change the type of an object that gets stored, we do not want to
change the type of the memory location being stored to.  In order to
still be able to do the rewrite, we will decompose and rebuild the
object so it is the type that can be stored.

### DIFF
--- a/source/opt/copy_prop_arrays.cpp
+++ b/source/opt/copy_prop_arrays.cpp
@@ -343,86 +343,88 @@ bool CopyPropagateArrays::CanUpdateUses(ir::Instruction* original_ptr_inst,
   analysis::ConstantManager* const_mgr = context()->get_constant_mgr();
   analysis::DefUseManager* def_use_mgr = context()->get_def_use_mgr();
 
-  return def_use_mgr->WhileEachUse(original_ptr_inst, [this, type_id, type_mgr,
-                                                       const_mgr](
-                                                          ir::Instruction* use,
-                                                          uint32_t index) {
-    analysis::Pointer* pointer_type = nullptr;
-    switch (use->opcode()) {
-      case SpvOpLoad: {
-        pointer_type = type_mgr->GetType(type_id)->AsPointer();
-        uint32_t new_type_id = type_mgr->GetId(pointer_type->pointee_type());
+  analysis::Type* type = type_mgr->GetType(type_id);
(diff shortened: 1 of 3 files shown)


source/opt/copy_prop_arrays.cpp
```diff
@@ -343,86 +343,88 @@ bool CopyPropagateArrays::CanUpdateUses(ir::Instruction* original_ptr_inst,
...
-        if (new_type_id != use->type_id()) {
-          return CanUpdateUses(use, new_type_id);
-        }
-        return true;
-      }
-      case SpvOpAccessChain: {
-        pointer_type = type_mgr->GetType(type_id)->AsPointer();
-        const analysis::Type* pointee_type = pointer_type->pointee_type();
+  if (!type->AsStruct() && !type->AsArray() && !type->AsPointer()) {
+    // If the type is not an aggragate, then the desired type must be the
```
dnovillo: s/aggragate/aggregate/

source/opt/copy_prop_arrays.cpp
```diff
@@ -515,14 +535,95 @@ void CopyPropagateArrays::UpdateUses(ir::Instruction* original_ptr_inst,
         // If the use is the object being stored, we do not know how to change
         // the type, so we assume that |CanUpdateUse| would have returned false,
         // and we should not have called this function.
```
dnovillo: Looks like this comment needs to be updated.

source/opt/copy_prop_arrays.cpp
```diff
@@ -515,14 +535,95 @@ void CopyPropagateArrays::UpdateUses(ir::Instruction* original_ptr_inst,
...
+    for (uint32_t i = 0; i < original_types.size(); i++) {
+      ir::Instruction* extract = ir_builder.AddCompositeExtract(
+          type_mgr->GetId(original_types[i]), object_inst->result_id(), {i});
+      element_ids.push_back(GenerateCopy(extract, type_mgr->GetId(new_types[i]),
+                                         insertion_position));
+    }
+    return ir_builder.AddCompositeConstruct(new_type_id, element_ids)
+        ->result_id();
+  } else {
+    // If we do not have an aggragate type, then we have a problem.  Either we
```
dnovillo: s/aggragate/aggregate/

PR #1428, opened 2018-03-27
Val: OpControlBarrier usable in any execution model

### LINKED ISSUE #1427: spirv-val gives false positive for OpControlBarrier
SPIR-V 1.3 says:

"Before version 1.3, it is only valid to use this instruction with TessellationControl, GLCompute, or Kernel execution models. There is no such restriction starting with version 1.3."

The validator doesn't respect this sentence. 

Ref: https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/validate_barriers.cpp#L175

dneto0:
Fixes #1427

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -447,6 +447,13 @@ uint32_t ValidationState_t::getIdBound() const { return id_bound_; }
 
 void ValidationState_t::setIdBound(const uint32_t bound) { id_bound_ = bound; }
 
+void ValidationState_t::RegisterSpirvVersion(uint32_t version) {
+  spirv_version_ = version;
+  if (spirv_version_ >= 0x10300) {
+    features_.control_barrier_usable_in_any_execution_model = true;
+  }
+}
+
 bool ValidationState_t::RegisterUniqueTypeDeclaration(
     const spv_parsed_instruction_t& inst) {
   std::vector<uint32_t> key;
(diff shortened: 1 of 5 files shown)


source/val/validation_state.cpp
```diff
@@ -447,6 +447,13 @@ uint32_t ValidationState_t::getIdBound() const { return id_bound_; }
 
 void ValidationState_t::setIdBound(const uint32_t bound) { id_bound_ = bound; }
 
+void ValidationState_t::RegisterSpirvVersion(uint32_t version) {
+  spirv_version_ = version;
+  if (spirv_version_ >= 0x10300) {
```
atgoo: That's the file version. We should make validation decisions based on the validator version.

I have an alternative fix:
https://github.com/KhronosGroup/SPIRV-Tools/pull/1429/files#diff-bdc10ce1ea59762351b41a27d6c1e2d0R176

dneto0: Ok.  Makes sense. 
It points out that we've got a hole:  When validating for SPIR-V 1.0, we should reject a module that has SPIR-V 1.3 as its version.  I'll file an issue for that.

dneto0:
Abandon in favour of #1429

dneto0:
I filed #1433 for rejecting a SPIR-V module from the "future"

PR #1431, opened 2018-03-27
Stop asking for extensions if feature avaiable in core SPIR-V

antiagainst:
Migrating to unified grammar means we sometimes have two fields
for a certain feature: version and extensions. It means the feature
in question can be used either in SPIR-V of advanced-enough
versions or in any SPIR-V with with the specified extensions.

Validator now respects the above rules.

### DIFF
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -163,10 +163,15 @@ ExtensionSet RequiredExtensions(const ValidationState_t& state,
   if (state.grammar().lookupOperand(type, operand, &operand_desc) ==
       SPV_SUCCESS) {
     assert(operand_desc);
+    // If this operand is incorporated into core SPIR-V before or in the current
+    // target environment, we don't require extensions anymore.
+    if (static_cast<uint32_t>(state.context()->target_env) >=
+        operand_desc->minVersion)
+      return {};
     return {operand_desc->numExtensions, operand_desc->extensions};
   }
 
-  return ExtensionSet();
+  return {};
 }
 
 }  // namespace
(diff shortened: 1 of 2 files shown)


source/validate_instruction.cpp
```diff
@@ -163,10 +163,15 @@ ExtensionSet RequiredExtensions(const ValidationState_t& state,
   if (state.grammar().lookupOperand(type, operand, &operand_desc) ==
       SPV_SUCCESS) {
     assert(operand_desc);
+    // If this operand is incorporated into core SPIR-V before or in the current
+    // target environment, we don't require extensions anymore.
+    if (static_cast<uint32_t>(state.context()->target_env) >=
```
dneto0: Should use spvVersionForTargetEnv here.
Don't rely on enum orderings.

test/val/val_extensions_test.cpp
```diff
@@ -228,4 +228,50 @@ TEST_P(ValidateAMDShaderBallotCapabilities, ExpectFailure) {
 INSTANTIATE_TEST_CASE_P(ExpectFailure, ValidateAMDShaderBallotCapabilities,
                         ValuesIn(AMDShaderBallotGroupInstructions()));
 
+struct ExtIntoCoreCase {
+  const char* cap;
```
dneto0: Should have some cases with Vulkan 1.0 and Vulkan 1.1.

source/validate_instruction.cpp
antiagainst: Very good catch! I've #1437 to solve it.

antiagainst: Done

test/val/val_extensions_test.cpp
antiagainst: Added tests for Vulkan target env.

alegal-arm:
any ETA here?

antiagainst:
@dneto0: I've addressed your comments. PTAL

dneto0:
Rebased, squashed, and pushed into master as 43ca2112b8f

Issue #1430, opened 2018-03-27
Misc improvements to array copy propagation.


s-perron:
- [x] Preserve analyses
- [x] Handle memory object from non-constant offsets.
- [x] Recognize more patterns for copying an object.

s-perron:
Finished by https://github.com/KhronosGroup/SPIRV-Tools/pull/1438.

PR #1435, opened 2018-03-28
Handle non-constant accesses in memory objects (copy prop arrays)

s-perron:
The first implementation of MemroyObject, which is used in copy
propagate arrays, forced the access chain to be like the access chains
in OpCompositeExtract.  This excluded the possibility of the memory
object from representing an array element that was extracted with a
variable index.   Looking at the code, that restriction is not
neccessary.  I also see some opportunities for doing this in some real
shaders.

Contributes to #1430.

### DIFF
--- a/source/opt/copy_prop_arrays.cpp
+++ b/source/opt/copy_prop_arrays.cpp
@@ -38,13 +38,20 @@ Pass::Status CopyPropagateArrays::Process(ir::IRContext* ctx) {
         continue;
       }
 
+      // Find the only store to the entire memory location, if it exists.
+      ir::Instruction* store_inst = FindStoreInstruction(&*var_inst);
+
+      if (!store_inst) {
+        continue;
+      }
+
       std::unique_ptr<MemoryObject> source_object =
-          FindSourceObjectIfPossible(&*var_inst);
+          FindSourceObjectIfPossible(&*var_inst, store_inst);
 
       if (source_object != nullptr) {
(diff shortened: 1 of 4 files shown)


source/opt/copy_prop_arrays.cpp
```diff
@@ -266,14 +246,25 @@ CopyPropagateArrays::BuildMemoryObjectFromExtract(
...
   std::unique_ptr<MemoryObject> result = GetSourceObjectIfAny(
       extract_inst->GetSingleWordInOperand(kCompositeExtractObjectInOperand));
 
   if (result) {
+    analysis::Integer int_type(32, false);
+    const analysis::Type* uint32_type =
+        context()->get_type_mgr()->GetRegisteredType(&int_type);
+
     std::vector<uint32_t> components;
+    // Convert the indecies in the extract instruction to a series of ids that
```
dnovillo: s/indecies/indices/

source/opt/type_manager.cpp
```diff
@@ -676,6 +676,8 @@ const Type* TypeManager::GetMemberType(
       parent_type = struct_type->element_types()[element_index];
     } else if (const analysis::Array* array_type = parent_type->AsArray()) {
       parent_type = array_type->element_type();
+    } else if (const analysis::RuntimeArray* runtime_array_type = parent_type->AsRuntimeArray()) {
```
dnovillo: Formatting?

PR #1436, opened 2018-03-28
Execution mode/model available in validation state

atgoo:
Refactored validate built-ins to make
GetExecutionModels(entry_point)
and
GetExecutionModes(entry_point)
available in validation state.

Entry points are allowed to have multiple execution modes and execution
models.

Finished the last missing feature in Vulkan built-ins validation:
FragDepth requires DepthReplacing.


dneto0:
Rebased, squashed, and pushed into master as da332cf332c7

PR #1437, opened 2018-03-28
Use standard SPIR-V version scheme for version requirement

antiagainst:
Previously we use symbols in spv_target_env as the minimum version
requirements for features. That makes version check implicitly
relies on the order of entries in the spv_target_env enum, which
also contains client APIs. Instead, we should use the standard
scheme for constructing SPIR-V version; and by doing that we can
also map client API entries to universial SPIR-V versions.

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -106,7 +106,7 @@ spv_result_t spvOpcodeTableNameLookup(spv_target_env env,
     // Note that the second rule assumes the extension enabling this instruction
     // is indeed requested in the SPIR-V code; checking that should be
     // validator's work.
-    if ((static_cast<uint32_t>(env) >= entry.minVersion ||
+    if ((spvVersionForTargetEnv(env) >= entry.minVersion ||
          entry.numExtensions > 0u) &&
         nameLength == strlen(entry.name) &&
         !strncmp(name, entry.name, nameLength)) {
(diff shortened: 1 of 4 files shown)


source/opcode.cpp
```diff
@@ -106,7 +106,7 @@ spv_result_t spvOpcodeTableNameLookup(spv_target_env env,
     // Note that the second rule assumes the extension enabling this instruction
     // is indeed requested in the SPIR-V code; checking that should be
     // validator's work.
-    if ((static_cast<uint32_t>(env) >= entry.minVersion ||
```
atgoo: How did this work?

antiagainst: It just happens that we have target envs for later versions of SPIR-V placed after earlier ones, and client API target envs placed after their required universal SPIR-V versions.

antiagainst:
The MacOS build failure is due to googletest framework: https://github.com/google/googletest/issues/1519

source/opcode.cpp
atgoo: But you are not comparing two enums.

On the left you have
static_cast<uint32_t>(SPV_ENV_UNIVERSAL_1_2) = 2

On the right you have it->minVersion which is probably something like 0x10002.

antiagainst: Previously the `generate_grammar_tables.py` were written to emit `SPV_ENV_UNIVERSAL_*` as the `minVersion`. So we were comparing enumerant values. This pull request changed that to use `SPV_SPIRV_VERSION_WORD()` everywhere. See https://github.com/KhronosGroup/SPIRV-Tools/pull/1437/files#diff-d07b0ddc717d188e2da6b3342256a34c

PR #1438, opened 2018-03-29
In copy-prop-arrays, indentify copies via OpCompositeInsert

s-perron:
When the original code copies an entire array or struct one element at a
time, this turns into a series of OpCompositeInsert instruction followed
by a store of the whole array.  We currently miss opportunities in copy
propagate arrays because we do not recognize this as a copy.

This commit adds code to copy propagate arrays to identify this code
pattern.

Also updates the performance passed to run array copy propagation.

### DIFF
--- a/source/opt/copy_prop_arrays.h
+++ b/source/opt/copy_prop_arrays.h
@@ -121,6 +121,10 @@ class CopyPropagateArrays : public MemPass {
       return pointer_type->storage_class();
     }
 
+    // Returns true if |other| represents memory that is contains inside of the
+    // memory represented by |this|.
+    bool Contains(MemoryObject* other);
+
    private:
     // The variable that owns this memory object.
     ir::Instruction* variable_inst_;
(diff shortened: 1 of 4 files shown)


source/opt/copy_prop_arrays.h
```diff
@@ -170,11 +174,20 @@ class CopyPropagateArrays : public MemPass {
...
   // |OpCompositeConstruct|.
   std::unique_ptr<MemoryObject> BuildMemoryObjectFromCompositeConstruct(
       ir::Instruction* conststruct_inst);
 
+  // Returns the memroy object that at some pointer was equivalent to the result
+  // of |insert_inst|.  If a memeory object cannot be identified, the return
+  // value is |nullptr\.  The opcode of |insert_inst| must be
+  // |OpCompositeInsert|.  This function looks for a series of
+  // |OpCompositeInsert| instructions That insert the elements one at a time in
+  // order from beginning to end.
```
dnovillo: Check the spelling of 'memory' above.

s-perron: Fix up the comment.  Not bothering with retesting for that.

PR #1441, opened 2018-03-29
Pinpoint a specific commit to bypass Googletest regression

antiagainst:
See https://github.com/google/googletest/issues/1519


antiagainst:
The culprit pull request is reverted.

PR #1442, opened 2018-03-30
Add missing <iterator> header for std::back_inserter

antiagainst:


antiagainst:
VS2015 is not happy because of missing `<iterator>` header for `std::back_inserter`: https://ci.appveyor.com/project/antiagainst/shaderc-rs/branch/master/job/76ym0x4visd0qucg

PR #1447, opened 2018-04-02
Teach array copy propagation about OpImageTexelPointer.

s-perron:
OpImageTexelPointer acts like a special kind of load.  It is not an
array load, but it also cannot be removed the same way a regular
load can.  The type of propagation that needs to be done is similar
to what we do for arrays, so I want to merge that code into that
optmization.

Contributers to #1445.


greg-lunarg:

PR #1448, opened 2018-04-02
Teach the private to local about OpImageTexelPointer.

s-perron:
OpImageTexelPointer acts like a special kind of load.  It is still
safe to change the storage class of a variable used in a
OpImageTexalPointer instruction.

Contributes to #1445.


greg-lunarg:

PR #1449, opened 2018-04-03
Fix major bug in validate_builtins

atgoo:
Fixed an early return in the loop, resulting in only one decoration
being checked.


dneto0:
Rebased and pushed into master as 691eed92cb

PR #1450, opened 2018-04-03
Fix bug in validate_builtins (definition checks were missing some cases)

atgoo:
At every definition of a builtin id, run at-reference-check rules on the
defining instruction as well.

Previosly the validation was missing the case when invalid storage class
was defined in the instruction which defines the built-in, and not in
the instruction which references the built-in.

### DIFF
--- a/source/validate_builtins.cpp
+++ b/source/validate_builtins.cpp
@@ -753,11 +753,7 @@ spv_result_t BuiltInsValidator::ValidateClipOrCullDistanceAtDefinition(
   }
 
   // Seed at reference checks with this built-in.
-  id_to_at_reference_checks_[inst.id()].push_back(
-      std::bind(&BuiltInsValidator::ValidateClipOrCullDistanceAtReference, this,
-                decoration, inst, inst, std::placeholders::_1));
-
-  return SPV_SUCCESS;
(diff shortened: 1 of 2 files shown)


source/validate_builtins.cpp
```diff
@@ -753,11 +753,7 @@ spv_result_t BuiltInsValidator::ValidateClipOrCullDistanceAtDefinition(
   }
 
   // Seed at reference checks with this built-in.
-  id_to_at_reference_checks_[inst.id()].push_back(
```
dneto0: Shouldn't the change be to *additionally* check the defining instruction?
Are we losing the checks at the references?  I'm confused

atgoo: ValidateClipOrCullDistanceAtReference will register itself at id_to_at_reference_checks_ on its own.

dneto0: I see, that was pre-existing functionality.  Thanks.

dneto0:
Rebased, patched, and pushed into master as d7fff408e308169

The patch was to add the extra execution modes with code at https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/val/val_builtins_test.cpp#L326 
It clones similar code in the other test fixtures just above it.  It's necessary to pass the DepthReplacing check that was introduced into master since this PR was first proposed.

PR #1451, opened 2018-04-04
OpName and decorations should not stop array copy prop.

s-perron:
Current names and decorations stop us from propagating arrays.  I don't think they should.  Definitely not the name.  Let me know if there are any specific decorations that would.


s-perron:
Reformatted before I merged.

PR #1452, opened 2018-04-04
linker: Properly remove FuncParamAttr from imported symbols

pierremoreau:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/898

### DIFF
--- a/source/link/linker.cpp
+++ b/source/link/linker.cpp
@@ -37,8 +37,8 @@
 
 namespace spvtools {
 
-using ir::IRContext;
 using ir::Instruction;
+using ir::IRContext;
 using ir::Module;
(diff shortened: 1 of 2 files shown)


pierremoreau:
I wanted to have this as part of a PR with more improvements to the linker regarding decorations and groups, but since I have no idea when I’ll have more time to spend on it, I thought I’d send the patch on its own for now, to fix the issue.

dneto0:
Rebased, squashed, and pushed into master as caf7da87e

pierremoreau:
Thanks! I’ll try to work on improving the linker further in the coming weeks, and hopefully bring it out of the WIP state. :-)

PR #1453, opened 2018-04-05
WIP: Fix SSA rewrite for loops with tight back edges

greg-lunarg:


greg-lunarg:
This passes all my size and correctness tests.

s-perron:
This change makes sense to me.  When you see the load in the loop, it does not find a store.  So it reaches the top of the loop still looking for a store.  So it decides to create a phi node.  When adding the phi operands it see the current block as a predecessor.  Since this block is sealed, it believes it can use the current definition on the block, which is the new phi node itself.  This turns the phi candidate  into a trivial phi because it missed the unprocessed store.

Since `IsBlockSealed` is only used when adding phi operands or finalizing phi candidates, this seems like the right change.  It will not effect finalizing of phi candidates.  We have to change AddPhiOperands.

The original comment before the block is sealed suggests it is needed to avoid a phi node when reading from a store in the same basic block as the load.  I don't think that will happen with this change.  When the store is processed it is added to `defs_at_block_`.   When processing the load, `GetReachingDef` will first check the current def in the current block (sealed or not) and it should find that store.  No phi candidate will be created.

greg-lunarg:
Thanks Steven.

I would recommend we try to resolve this soon as it appears that others are starting to see this problem #1456.

dnovillo:
Taking a look now.  @s-perron thanks for the analysis.  @greg-lunarg I'll review and test with our codebase.

dnovillo:
Yes, I agree.  This PR is the right fix.  Thanks for your patience @greg-lunarg.  Adding more detail for future reference (because I managed to trip myself with this one):

The slice of the CFG that is interesting for the bug is

```
25
|
v
30
|
v
31<-+
|   |
v   |
34--+
```
1. In block 25, we have a Phi candidate for `%f` with arguments `%47 = Phi[%float_0, %0]`. This merges `%float_0` and a yet unknown argument from the external loop backedge.
2. We are now processing block 34:
    1. The load `%35 = OpLoad %f` triggers a Phi candidate to be placed in block 31.
    2. The Phi candidate `%50 = Phi` needs two arguments.  The one coming from block 30 is `%47`. But the one coming from block 34 (which we are now processing and have marked sealed), finds `%50` itself as the reaching def for `%f`.
3. This wrongfully marks `%50` as a copy-of Phi, which ultimately makes both `%47` and `%50` copy-of Phis that get eliminated.

I'll rebase and commit.  Thanks for the fix!

dnovillo:
Re-based, tested internally and against testsuite.  Pushed as 6fbfe1c016bd8ffaf8f59db75e660d7d9a447843

PR #1454, opened 2018-04-05
Test asm/dis for SPV_NV_shader_subgroup_partitioned and SPV_EXT_descriptor_indexing

dneto0:


test/text_to_binary.extension_test.cpp
```diff
@@ -523,4 +523,137 @@ INSTANTIATE_TEST_CASE_P(
...
+            {"%2 = OpGroupIAdd %1 %3 PartitionedExclusiveScanNV %4\n",
+             MakeInstruction(SpvOpGroupIAdd, {1, 2, 3, 8, 4})},
+        })), );
+
+// SPV_EXT_descriptor_indexing
+
+INSTANTIATE_TEST_CASE_P(
+    SPV_EXT_descriptor_indexing, ExtensionRoundTripTest,
+    Combine(
+        Values(SPV_ENV_UNIVERSAL_1_0, SPV_ENV_UNIVERSAL_1_1,
```
antiagainst: Why not SPIR-V 1.3 and/or Vulkan 1.1?

dneto0: No reason.  It was copy-pasted from somewhere else old.  I'll add them.

PR #1455, opened 2018-04-05
More automatic extension support

dneto0:
Update grammar table generation:
- Get extensions from instructions, not just operand-kinds
- Don't explicitly list extensions that come from the SPIR-V core
  grammar or from a KHR extended instruction set grammar.

This makes it easier to support new extensions since the recommended
extension strategy is to add instructions to the core grammar file.

Also, test the validator has trivial support for passing through
the extensions SPV_NV_shader_subgroup_partitioned and
SPV_EXT_descriptor_indexing.


utils/generate_grammar_tables.py
```diff
@@ -524,15 +499,25 @@ def generate_operand_kind_table(enums):
...
-def get_extension_list(operands):
+def get_extension_list(instructions, operand_kinds):
     """Returns extensions as an alphabetically sorted list of strings."""
-    enumerants = sum([item.get('enumerants', []) for item in operands
-                      if item.get('category') in ['ValueEnum']], [])
 
-    extensions = sum([item.get('extensions', []) for item in enumerants
+    things_with_an_extensions_field = [item for item in instructions]
+
+    enumerants = sum([item.get('enumerants', []) for item in operand_kinds], [])
```
antiagainst: Genius

PR #1460, opened 2018-04-06
Fold OpDot

s-perron:
Adding three rules to fold OpDot (implemented as two).

- When an OpDot has two constants, then fold to the resulting const.

- When one of the inputs is the 0 vector, then fold to zero.

- When one of the inputs is a single 1 with 0s, then rewrite to an
OpCompositeExtract of the appropriate element.  This will help find
even more folding opportunities.

Contributes to #709.

### DIFF
--- a/source/opt/folding_rules.cpp
+++ b/source/opt/folding_rules.cpp
@@ -13,6 +13,9 @@
 // limitations under the License.
 
 #include "folding_rules.h"
+
+#include <limits>
+
 #include "latest_version_glsl_std_450_header.h"
 
 namespace spvtools {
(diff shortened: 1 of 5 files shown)


source/opt/folding_rules.cpp
```diff
@@ -1812,6 +1812,79 @@ FoldingRule RedundantFMix() {
   };
 }
 
+// This rule look for a dot with a constant vector containing a single 1 and
+// the rest 0s.  This is the same as doing an extract.
+FoldingRule DotProductDoingExtract() {
+  return [](ir::Instruction* inst,
+            const std::vector<const analysis::Constant*>& constants) {
+    assert(inst->opcode() == SpvOpDot && "Wrong opcode.  Should be OpOpDot.");
```
alan-baker: OpOpDot -> OpDot

dneto0:
Bot failures: /home/travis/build/KhronosGroup/SPIRV-Tools/source/opt/const_folding_rules.cpp:317:71: error: declaration of ‘const_mgr’ shadows a previous local [-Werror=shadow]
      analysis::ConstantManager* const_mgr) -> const analysis::Constant* {

PR #1461, opened 2018-04-06
Add constant folding for FNegate

jaebaek:
Let `OpFNegate` against a constant be the negative of the operand.

Contributes to #709.

### DIFF
--- a/test/opt/fold_test.cpp
+++ b/test/opt/fold_test.cpp
@@ -229,8 +229,6 @@ OpName %main "main"
 %v2double_4_4 = OpConstantComposite %v2double %double_4 %double_4
 %v2double_2_0p5 = OpConstantComposite %v2double %double_2 %double_0p5
 %v2double_null = OpConstantNull %v2double
-%float_nan = OpConstant %float -0x1.8p+128
-%double_nan = OpConstant %double -0x1.8p+1024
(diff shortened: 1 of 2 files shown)


test/opt/fold_test.cpp
```diff
@@ -799,7 +799,15 @@ ::testing::Values(
...
-        2, -std::numeric_limits<float>::infinity())
+        2, -std::numeric_limits<float>::infinity()),
+    // Test case 3: Fold -2.0
+    InstructionFoldingCase<float>(
+        Header() + "%main = OpFunction %void None %void_func\n" +
+            "%main_lab = OpLabel\n" +
+            "%2 = OpFNegate %float %float_2\n" +
+            "OpReturn\n" +
+            "OpFunctionEnd",
+        2, -2)
```
s-perron: I know we don't do this for many cases, but could you add a test for when the input is the result of and OpConstantNull.

Also add a test for a vector.

s-perron: Test double as well.

dneto0:
Hi @jaebaek : Please rebased your change and add tests as requested

test/opt/fold_test.cpp
jaebaek: Done

PR #1462, opened 2018-04-06
Validate binary version

### LINKED ISSUE #1433: If validating for an environment defined for SPIR-V N, then reject a module with SPIR-V version M where  M > N

alan-baker:
Fixes #1433

* Validates SPIR-V binary version against target environment
* Fixed a few header typos

Wasn't sure what to say about OpenCL 1.2

### DIFF
--- a/source/spirv_target_env.h
+++ b/source/spirv_target_env.h
@@ -24,4 +24,7 @@ bool spvParseTargetEnv(const char* s, spv_target_env* env);
 // Returns true if |env| is a VULKAN environment, false otherwise.
 bool spvIsVulkanEnv(spv_target_env env);
 
+// Returns the version number for the given SPIR-V target environment.
+uint32_t spvVersionForTargetEnv(spv_target_env env);
+
 #endif  // LIBSPIRV_SPIRV_TARGET_ENV_H_
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -234,81 +234,6 @@ UNUSED(void PrintDotGraph(ValidationState_t& _, libspirv::Function func)) {
   }
 }
 
-// Returns the major version number from |version|.
-uint32_t parseMajorVersion(uint32_t version) {
-  // 2.3 Table 1: Version bytes are 0 | major | minor | 0.
-  return version >> 16;
-}
-
-// Returns the minor version number from |version|.
-uint32_t parseMinorVersion(uint32_t version) {
-  // 2.3 Table 1: Version bytes are 0 | major | minor | 0.
-  return (version & 0xff00) >> 8;
-}
-
-// Returns true if SPIR-V version |version| supports the target environment
-// |env|.
-bool validateVersion(uint32_t version, const spv_target_env& env) {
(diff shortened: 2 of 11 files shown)


source/validate.cpp
```diff
@@ -234,6 +234,81 @@ UNUSED(void PrintDotGraph(ValidationState_t& _, libspirv::Function func)) {
...
+  return (version & 0xff00) >> 8;
+}
+
+// Returns true if SPIR-V version |version| supports the target environment
+// |env|.
+bool validateVersion(uint32_t version, const spv_target_env& env) {
+  uint32_t major = parseMajorVersion(version);
+  uint32_t minor = parseMinorVersion(version);
+  switch (env) {
+    case SPV_ENV_VULKAN_1_0:
```
dneto0: This function should use spvVersionForTargetEnv from spirv_constant.h
Then you won't need any specific cases, and that will reduce maintenance.

source/validate.cpp
```diff
@@ -234,6 +234,81 @@ UNUSED(void PrintDotGraph(ValidationState_t& _, libspirv::Function func)) {
   }
 }
 
+// Returns the major version number from |version|.
+uint32_t parseMajorVersion(uint32_t version) {
```
dneto0: See SPV_SPIRV_VERSION_MAJOR_PART and friends in source/spirv_constant.h

source/validate.cpp
dneto0: (And feel free to move that function from spirv_constant.h to spirv_target_env.h which seems like a more discoverable place for it.

source/spirv_target_env.h
```diff
@@ -24,4 +24,7 @@ bool spvParseTargetEnv(const char* s, spv_target_env* env);
 // Returns true if |env| is a VULKAN environment, false otherwise.
 bool spvIsVulkanEnv(spv_target_env env);
 
+// Returns the version number for the given SPIR-V target environment.
```
dneto0: Nice.  The implementation was *ALREADY* in source/spv_targe_env.cpp

source/validate.cpp
```diff
@@ -255,6 +255,16 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
            << "Invalid SPIR-V header.";
   }
 
+  if (header.version > spvVersionForTargetEnv(context.target_env)) {
```
dneto0: schweet

pierremoreau:
> Wasn't sure what to say about OpenCL 1.2

OpenCL 1.2 should work the same as OpenCL 2.0, as both needs the cl_khr_il_program extension to get SPIR-V support; there are some differences, like OpenCL 1.2 not having generic pointers but OpenCL 2.0 does, but version-wise, they both only require SPIR-V 1.0 support.

dneto0:
Rebased, squashed, and pushed into master as 42840d15e4

dneto0:
OpenCL 1.2 cases covered in #1464, pushed into master just now.

PR #1463, opened 2018-04-07
Fix PrimitiveId builtin check for Vulkan

antiagainst:
According to Vulkan spec 1.1.72:

> The PrimitiveId decoration must be used only within fragment,
> tessellation control, tessellation evaluation, and geometry shaders.

> In a tessellation control or tessellation evaluation shader, any
> variable decorated with PrimitiveId must be declared using the Input
> storage class.

We were enforcing that PrimitiveId can only be used with Output
storage class for TCS and TES before.

### DIFF
--- a/source/validate_builtins.cpp
+++ b/source/validate_builtins.cpp
@@ -1478,26 +1478,22 @@ spv_result_t BuiltInsValidator::ValidatePrimitiveIdAtReference(
              << " " << GetStorageClassDesc(referenced_from_inst);
     }
 
-    if (storage_class == SpvStorageClassInput) {
+    if (storage_class == SpvStorageClassOutput) {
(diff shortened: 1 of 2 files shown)


antiagainst:
This is blocking DXC to use the latest SPIRV-Tools.

dneto0:
The Vulkan 1.0.70 spec still says "Output" storage class.
But this was a bug, and fixed in Vulkan 1.1 with coimmit https://github.com/KhronosGroup/Vulkan-Docs/commit/ce60b9c88745ecded74296dfbe69dae7c1fb2e62 
It was tracked as Khronos internal Vulkan issue 1109.

dneto0:
Rebased and pushed onto master as 26a698c347

PR #1473, opened 2018-04-11
Add support for MIV and Delta test dependence analysis

tremmelg:
Provides functionality to perform MIV and Delta test dependence analysis tests.

GCD MIV test as described in Chapter 3 of "Optimizing Compilers for Modern Architectures: A Dependence-Based Approach" by Randy Allen, and Ken Kennedy.

Delta test as described in Figure 3 of "Practical Dependence Testing" by Gina Goff, Ken Kennedy, and Chau-Wen Tseng from PLDI '91.

### DIFF
--- a/source/opt/loop_dependence.cpp
+++ b/source/opt/loop_dependence.cpp
@@ -1355,7 +1355,7 @@ Constraint* LoopDependenceAnalysis::IntersectConstraints(
 // Propagate constraints function as described in section 5 of Practical
 // Dependence Testing, Goff, Kennedy, Tseng, 1991.
 SubscriptPair LoopDependenceAnalysis::PropagateConstraints(
-    SubscriptPair& subscript_pair,
+    const SubscriptPair& subscript_pair,
     const std::vector<Constraint*>& constraints) {
   SENode* new_first = subscript_pair.first;
   SENode* new_second = subscript_pair.second;
@@ -1424,19 +1424,15 @@ bool LoopDependenceAnalysis::DeltaTest(
                 [this]() { return make_constraint<DependenceNone>(); });
 
   // Separate SIV and MIV subscripts
-  auto first_MIV = std::partition(
-      std::begin(coupled_subscripts), std::end(coupled_subscripts),
-      [this](const SubscriptPair& p) { return IsSIV(p); });
-
   std::vector<SubscriptPair> siv_subscripts{};
   std::vector<SubscriptPair> miv_subscripts{};
 
-  for (auto it = std::begin(coupled_subscripts); it < first_MIV; ++it) {
-    siv_subscripts.push_back(*it);
-  }
-
-  for (auto it = first_MIV; it < std::end(coupled_subscripts); ++it) {
-    miv_subscripts.push_back(*it);
+  for (const auto& subscript_pair : coupled_subscripts) {
+    if (IsSIV(subscript_pair)) {
+      siv_subscripts.push_back(subscript_pair);
+    } else {
+      miv_subscripts.push_back(subscript_pair);
+    }
   }
 
   // Delta Test
--- a/source/opt/loop_dependence.cpp
+++ b/source/opt/loop_dependence.cpp
@@ -30,6 +30,157 @@ namespace opt {
 
 using SubscriptPair = std::pair<SENode*, SENode*>;
 
+namespace {
+
+// Calculate the greatest common divisor of a & b using Stein's algorithm.
+// https://en.wikipedia.org/wiki/Binary_GCD_algorithm
+int64_t GreatestCommonDivisor(int64_t a, int64_t b) {
+  // Simple cases
+  if (a == b) {
+    return a;
+  } else if (a == 0) {
+    return b;
+  } else if (b == 0) {
+    return a;
+  }
+
+  // Both even
+  if (a % 2 == 0 && b % 2 == 0) {
+    return 2 * GreatestCommonDivisor(a / 2, b / 2);
+  }
+
+  // Even a, odd b
+  if (a % 2 == 0 && b % 2 == 1) {
+    return GreatestCommonDivisor(a / 2, b);
+  }
+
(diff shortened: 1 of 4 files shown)


source/opt/loop_dependence.cpp
```diff
@@ -129,11 +134,51 @@ bool LoopDependenceAnalysis::GetDependence(const ir::Instruction* source,
...
+  for (auto it = first_coupled; it < std::end(sets_of_subscripts); ++it) {
+    auto coupled_instructions = *it;
+    std::vector<SubscriptPair> coupled_subscripts{};
+
+    for (const auto& elem : coupled_instructions) {
+      auto source_subscript = std::get<0>(elem);
+      auto destination_subscript = std::get<1>(elem);
+
+      SENode* source_node = scalar_evolution_.SimplifyExpression(
+          scalar_evolution_.AnalyzeInstruction(source_subscript));
```
dnovillo: Are we guaranteed that AnalyzeInstruction won't return a nullptr here?  SimplifyExpression won't like that.

source/opt/loop_dependence.cpp
```diff
@@ -817,6 +872,175 @@ bool LoopDependenceAnalysis::WeakCrossingSIVTest(
   return false;
 }
 
+// Calculate the greatest common divisor of a & b using Stein's algorithm.
+// https://en.wikipedia.org/wiki/Binary_GCD_algorithm
+int64_t GreatestCommonDivisor(int64_t a, int64_t b) {
```
dnovillo: Inside an anonymous namespace, please.  Likewise for all the other free functions in this file.  If the functions need to be shared across other files, perhaps make the class static?

source/opt/loop_dependence.cpp
```diff
@@ -817,6 +872,175 @@ bool LoopDependenceAnalysis::WeakCrossingSIVTest(
...
+
+// Check if node is affine, ie in the form: a0*i0 + a1*i1 + ... an*in + c
+// and contains only the following types of nodes: SERecurrentNode, SEAddNode
+// and SEConstantNode
+bool IsInCorrectFormForGCD(SENode* node) {
+  bool children_ok = true;
+
+  if (auto add_node = node->AsSEAddNode()) {
+    for (auto child : add_node->GetChildren()) {
+      children_ok = children_ok && IsInCorrectFormForGCD(child);
```
dnovillo: &= here?

source/opt/loop_dependence.cpp
```diff
@@ -883,5 +1111,524 @@ PartitionedSubscripts LoopDependenceAnalysis::PartitionSubscripts(
   return partitions;
 }
 
+bool IsConstant(const SENode* node) {
```
dnovillo: enclose in anonymous namespace, please.  Or, better yet, make this an class function of SENode?

source/opt/loop_dependence.cpp
```diff
@@ -883,5 +1111,524 @@ PartitionedSubscripts LoopDependenceAnalysis::PartitionSubscripts(
...
+  return node->AsSEConstantNode() != nullptr;
+}
+
+// Compare 2 fractions while first normalizing them, e.g. 2/4 and 4/8 will both
+// be simplified to 1/2 and then determined to be equal.
+bool NormalizeAndCompareFractions(int64_t numerator_0, int64_t denominator_0,
```
dnovillo: In anonymous namespace.

source/opt/loop_dependence.cpp
tremmelg: I've left them in the anonymous namespace, as they are only used here.

source/opt/loop_dependence.cpp
tremmelg: I've removed it altogether and replaced the uses with `SENode::AsSEConstantNode`

source/opt/loop_dependence.cpp
tremmelg: Done.

source/opt/loop_dependence.cpp
```diff
@@ -147,7 +192,11 @@ bool LoopDependenceAnalysis::GetDependence(const ir::Instruction* source,
   return false;
 }
 
-bool LoopDependenceAnalysis::ZIVTest(SENode* source, SENode* destination) {
+bool LoopDependenceAnalysis::ZIVTest(
```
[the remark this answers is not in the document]
tremmelg: Done. I also made the same change elsewhere for read-only arguments.

source/opt/loop_dependence.cpp
tremmelg: `AnalyzeInstruction` will always return a valid pointer. It will be `SEValueUnknown` or  `SECantCompute` if it encounters anything unexpected.

source/opt/loop_dependence.cpp
```diff
@@ -1111,28 +1134,6 @@ PartitionedSubscripts LoopDependenceAnalysis::PartitionSubscripts(
...
-  auto normalized_numerator_0 = numerator_0 / gcd_0;
-  auto normalized_denominator_0 = denominator_0 / gcd_0;
-  auto normalized_numerator_1 = numerator_1 / gcd_1;
-  auto normalized_denominator_1 = denominator_1 / gcd_1;
-
-  return normalized_numerator_0 == normalized_numerator_1 &&
-         normalized_denominator_0 == normalized_denominator_1;
-}
-
 std::shared_ptr<Constraint> LoopDependenceAnalysis::IntersectConstraints(
```
dnovillo: I forgot to comment on the use of shared_ptr<> before.  Is it absolutely necessary to use it?  The coding guideline strongly prefers unique_ptr with ownership transfer.  It's easier to understand than shared_ptr.

dnovillo:
Re-tested and squashed.  Pushed upstream as 0f335cf87e7a42e45b1a06958c4bc2d8a57e3978

PR #1475, opened 2018-04-11
Rewrite local-single-store-elim to not create large data structures.

s-perron:
The local-single-store-elim algorithm is not fundamentally bad.
However, when there are a large number of variables, some of the
maps that are used can become very large.  These large data structures
then take a very long time to be destroyed.  I've seen cases around 40%
if the time.

I've rewritten that algorithm to not use as much memory.  This give a
significant improvement when running a large number of shader through
DXC.

I've also made a small change to local-single-block-elim to delete the
loads that is has replaced.  That way local-single-store-elim will not
have to look at those.  local-single-store-elim now does the same thing.

The time for one set goes from 309s down to 126s.  For another set, the
time goes from 102s down to 88s.

### DIFF
--- a/source/opt/local_single_store_elim_pass.cpp
+++ b/source/opt/local_single_store_elim_pass.cpp
@@ -30,218 +30,24 @@ const uint32_t kVariableInitIdInIdx = 1;
 
 }  // anonymous namespace
 
-bool LocalSingleStoreElimPass::HasOnlySupportedRefs(uint32_t ptrId) {
-  if (supported_ref_ptrs_.find(ptrId) != supported_ref_ptrs_.end()) return true;
-  if (get_def_use_mgr()->WhileEachUser(ptrId, [this](ir::Instruction* user) {
-        SpvOp op = user->opcode();
-        if (IsNonPtrAccessChain(op) || op == SpvOpCopyObject) {
-          if (!HasOnlySupportedRefs(user->result_id())) {
-            return false;
-          }
-        } else if (op != SpvOpStore && op != SpvOpLoad && op != SpvOpName &&
-                   !IsNonTypeDecorate(op)) {
-          return false;
-        }
-        return true;
-      })) {
-    supported_ref_ptrs_.insert(ptrId);
-    return true;
-  }
-  return false;
-}
-
-void LocalSingleStoreElimPass::SingleStoreAnalyze(ir::Function* func) {
-  ssa_var2store_.clear();
-  non_ssa_vars_.clear();
-  store2idx_.clear();
-  store2blk_.clear();
-  for (auto bi = func->begin(); bi != func->end(); ++bi) {
-    uint32_t instIdx = 0;
-    for (auto ii = bi->begin(); ii != bi->end(); ++ii, ++instIdx) {
-      uint32_t varId = 0;
-      ir::Instruction* ptrInst = nullptr;
-      switch (ii->opcode()) {
-        case SpvOpStore: {
-          ptrInst = GetPtr(&*ii, &varId);
-        } break;
-        case SpvOpVariable: {
-          // If initializer, treat like store
-          if (ii->NumInOperands() > 1) {
-            varId = ii->result_id();
-            ptrInst = &*ii;
-          }
-        } break;
-        default:
-          break;
-      }  // switch
-      if (varId == 0) continue;
-      // Verify variable is target type
-      if (non_ssa_vars_.find(varId) != non_ssa_vars_.end()) continue;
-      if (ptrInst->opcode() != SpvOpVariable) {
-        non_ssa_vars_.insert(varId);
-        ssa_var2store_.erase(varId);
-        continue;
-      }
-      // Verify target type and function storage class
-      if (!IsTargetVar(varId)) {
-        non_ssa_vars_.insert(varId);
-        continue;
-      }
-      if (!HasOnlySupportedRefs(varId)) {
-        non_ssa_vars_.insert(varId);
-        continue;
-      }
-      // Ignore variables with multiple stores
-      if (ssa_var2store_.find(varId) != ssa_var2store_.end()) {
-        non_ssa_vars_.insert(varId);
-        ssa_var2store_.erase(varId);
-        continue;
-      }
-      // Remember pointer to variable's store and it's
-      // ordinal position in block
-      ssa_var2store_[varId] = &*ii;
-      store2idx_[&*ii] = instIdx;
-      store2blk_[&*ii] = &*bi;
-    }
-  }
-}
-
-LocalSingleStoreElimPass::GetBlocksFunction
-LocalSingleStoreElimPass::AugmentedCFGSuccessorsFunction() const {
-  return [this](const ir::BasicBlock* block) {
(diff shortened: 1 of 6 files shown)


dnovillo:
Seems to have formatting errors.  https://travis-ci.org/KhronosGroup/SPIRV-Tools/jobs/365286960

greg-lunarg:
Seeing a regression with this branch. Will isolate commit next...

greg-lunarg:
Yes, its this commit. Will start chopping and diffing...

s-perron:
@greg-lunarg Do you have a test case you can share that shows the problem you found?

greg-lunarg:
Shortly...

greg-lunarg:
It looks like the bug I am seeing is being uncovered by this change. I still have a little more work to track this down and confirm...

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -319,6 +120,127 @@ void LocalSingleStoreElimPass::InitExtensions() {
...
+  }
+
+  for (ir::Instruction* user : users) {
+    switch (user->opcode()) {
+      case SpvOpStore:
+        if (store_inst == nullptr) {
+          store_inst = user;
+        } else {
+          // More than 1 store.
+          store_inst = nullptr;
```
greg-lunarg: This is incorrect and causes bad code to be generated. This logic just forgets that a store was seen. One way to correct this is to replace this with "return nullptr;".

greg-lunarg:
On further inspection, it was this commit after all. I found the bug and submitted a suggested fix above.

My tests now pass.

I do have a few more minor suggestions and will submit them shortly.

source/opt/local_single_store_elim_pass.cpp
s-perron: That use to an `if-then-else` instead of the `switch`.  So the break broke out of the loop.  It also was originally in a larger part of the code instead of being in its own function do a return would not have worked.  Then I made changes and do not have a test for this.  Good catch.  I'll try to create a test case for this.  You suggestion should work.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -319,6 +120,127 @@ void LocalSingleStoreElimPass::InitExtensions() {
...
+
+  // If |var_inst| has an initializer, then that will count as a store.
+  if (var_inst->NumInOperands() > 1) {
+    store_inst = var_inst;
+  }
+
+  for (ir::Instruction* user : users) {
+    switch (user->opcode()) {
+      case SpvOpStore:
+        if (store_inst == nullptr) {
```
greg-lunarg: Don't we need a check here that the variable of interest is the *first* operand of the store? If it isn't, I think this algorithm will generate incorrect code. Yeah, having a variable show up in the second operand here is admittedly rare, but I am pretty sure it is legal, so we should guard against it.  Since it is rare, I haven't thought a lot about the optimal way to deal with it, but I think it would be *safe* to just return nullptr for the function. Since this event is rare, it wouldn't hurt performance all that much.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -319,6 +120,127 @@ void LocalSingleStoreElimPass::InitExtensions() {
...
+          // More than 1 store.
+          store_inst = nullptr;
+          break;
+        }
+        break;
+      case SpvOpAccessChain:
+      case SpvOpInBoundsAccessChain:
+        if (FeedsAStore(user)) {
+          // Has a partial store.  Cannot propagate that.
+          store_inst = nullptr;
```
greg-lunarg: I think we need a "return nullptr;" here too.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -319,6 +120,127 @@ void LocalSingleStoreElimPass::InitExtensions() {
...
+          break;
+        }
+        break;
+      case SpvOpLoad:
+      case SpvOpImageTexelPointer:
```
greg-lunarg: SpvOpImageTexelPointer can feed a "write" operation, right? Probably either need to verify or have this disqualify the variable to be safe, right?

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -319,6 +120,127 @@ void LocalSingleStoreElimPass::InitExtensions() {
...
+      case SpvOpName:
+      case SpvOpCopyObject:
+        break;
+      default:
+        if (!user->IsDecoration()) {
+          // Don't know if this instruction modifies the variable.
+          // Conservatively assume it is a store.
+          store_inst = nullptr;
```
greg-lunarg: Another "return nullptr;" here?

source/opt/local_single_store_elim_pass.cpp
s-perron: No I do not want it to disqualify the symbol.  It cannot feed and OpImageWrite, but it can feed atomic operations that write to the memory location.  However, I think this goes back to the issue we discussed before about the handle model.  When you do an OpImageWrite, what memory are you actually writing to?

```
   %gTexture = OpVariable %_ptr_UniformConstant_7 UniformConstant                                                                  
...
        %105 = OpLoad %7 %gTexture
               OpImageWrite %105 %int_0 %float_0
```

The write cannot be to %105 it is an ssa symbol.  It cannot be to the memory pointed to by %gTexture because it is in UniformConstant.  It write to the "image memory" that is access by looking at the data that %gTexture points to (the stuff in the UniformConstant memory.).

It is the same thing for OpImageTexelPointer.  The result of that instruction does not point to the function scope memory, but something else.  Whatever happens after to the result of the OpImageTexelPointer instruction, it cannot change the memory in the function scope variable.

source/opt/local_single_store_elim_pass.cpp
s-perron: done

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -319,6 +120,127 @@ void LocalSingleStoreElimPass::InitExtensions() {
...
+  return !def_use_mgr->WhileEachUser(inst, [this](ir::Instruction* user) {
+    switch (user->opcode()) {
+      case SpvOpStore:
+        return false;
+      case SpvOpAccessChain:
+      case SpvOpInBoundsAccessChain:
+      case SpvOpCopyObject:
+        return !FeedsAStore(user);
+      case SpvOpLoad:
+      case SpvOpImageTexelPointer:
```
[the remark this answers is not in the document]
s-perron: See my earlier comment.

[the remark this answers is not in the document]
greg-lunarg: OK.

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: Agreed.

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: OK

PR #1477, opened 2018-04-12
Adds missing validator help environments

### LINKED ISSUE #1471: vulkan1.1 is not listed as valid environment in spirv-val
It works, but it's not there in --help.

alan-baker:
Fixes #1471

* Added environments referenced in --version, but not mentioned in
--help

### DIFF
--- a/tools/val/val.cpp
+++ b/tools/val/val.cpp
@@ -51,7 +51,7 @@ Options:
                                    members.
   --version                        Display validator version information.
   --target-env                     {vulkan1.0|vulkan1.1|opencl2.2|spv1.0|spv1.1|spv1.2|spv1.3}
-                                   Use Vulkan 1.0, Vulkan 1.1, SPIR-V 1.0, OpenCL 2.2,
+                                   Use Vulkan 1.0, Vulkan 1.1, OpenCL 2.2, SPIR-V 1.0,
                                    SPIR-V 1.1, SPIR-V 1.2 or SPIR-V 1.3 validation rules.
 )",
       argv0, argv0);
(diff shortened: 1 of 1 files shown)


tools/val/val.cpp
```diff
@@ -50,8 +50,8 @@ NOTE: The validator is a work in progress.
                                    different type with compatible layout and
                                    members.
   --version                        Display validator version information.
-  --target-env                     {vulkan1.0|spv1.0|spv1.1|spv1.2}
-                                   Use Vulkan1.0/SPIR-V1.0/SPIR-V1.1/SPIR-V1.2 validation rules.
+  --target-env                     {vulkan1.0|vulkan1.1|opencl2.2|spv1.0|spv1.1|spv1.2}
+                                   Use Vulkan1.0/Vulkan1.1/SPIR-V1.0/OpenCL2.2/SPIR-V1.1/SPIR-V1.2 validation rules.
```
pierremoreau: You might want to swap SPIR-V1.0 and OpenCL2.2 on that line to match the previous one. Also, OpenCL2.1 should be supported, and maybe even OpenCL1.2 and OpenCL2.0 (and there embedded profile counterparts).

dneto0: I prefer the extra whitespace as shown in the assembler's help, rather than the slashes.
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/tools/as/as.cpp#L45 
I think the spacing makes it more readable as the lists get longer.  Also as the line wraps.

dneto0:
Rebased, squashed, and pushed into master as c3ee2105639

PR #1478, opened 2018-04-12
Allow subgroup memory scope for Vulkan 1.1

### LINKED ISSUE #1469: spirv-val fails to validate subgroup barriers
[test.spv.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/1897776/test.spv.txt)
`error: 196: ControlBarrier: in Vulkan environment Memory Scope is limited to Device, Workgroup and Invocation`

`--target-env vulkan1.1`

alan-baker:
Fixes #1469

* New error that prevents CrossDevice memory scope for all vulkan
* Old error specifically references Vulkan 1.0
* New tests

### DIFF
--- a/source/validate_barriers.cpp
+++ b/source/validate_barriers.cpp
@@ -81,11 +81,18 @@ spv_result_t ValidateMemoryScope(ValidationState_t& _,
   }
 
   if (spvIsVulkanEnv(_.context()->target_env)) {
-    if (value != SpvScopeDevice && value != SpvScopeWorkgroup &&
+    if (value == SpvScopeCrossDevice) {
+      return _.diag(SPV_ERROR_INVALID_DATA)
+             << spvOpcodeString(opcode)
+             << ": in Vulkan environment, Memory Scope cannot be CrossDevice";
+    }
+    if (_.context()->target_env == SPV_ENV_VULKAN_1_0 &&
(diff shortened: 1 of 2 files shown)


source/validate_barriers.cpp
```diff
@@ -81,11 +81,18 @@ spv_result_t ValidateMemoryScope(ValidationState_t& _,
   }
 
   if (spvIsVulkanEnv(_.context()->target_env)) {
-    if (value != SpvScopeDevice && value != SpvScopeWorkgroup &&
+    if (value == SpvScopeCrossDevice) {
+      return _.diag(SPV_ERROR_INVALID_DATA)
+             << spvOpcodeString(opcode)
+             << ": in Vulkan environment, Memory Scope cannot be CrossDevice";
+    }
+    if (_.context()->target_env == SPV_ENV_VULKAN_1_0 &&
```
dneto0: Normally I'd use a whitelist on the permitted scopes.  But the scope enum is already limited to a specific set, and CrossDevice is the only banned one after Vulkan 1.0.
So this works for me.  And it will naturally work well if/when Vulkan 1.2 comes out, for example.
The only concern is if another scope enum is added.  But we can deal with that remote possibility later.

dneto0:
I don't see why the AppVeyor job failed.   Please rebase to trigger a new build.

PR #1479, opened 2018-04-12
spirv-val: List vulkan1.1 and spirv1.3 targets

### LINKED ISSUE #1471: vulkan1.1 is not listed as valid environment in spirv-val

dneto0:
Fixes #1471


dneto0:
Close in favour of #1479

PR #1488, opened 2018-04-12
Change parameter to Mempass::RemovePhiOperands

s-perron:
Pass a hashtable by const ref instead of by value.  Big impact on
compile time.


dneto0:
> Big impact on compile time.

Can you quantify this?

s-perron:
I saw a 10% improvement on a set of shaders on Linux, and 14% improvement on Windows.

PR #1489, opened 2018-04-12
Add validator test for OpBranch

dneto0:
Add test for case where OpBranch branches to a value (a function value).
Previous tests only checked a label value (name of a block.).

Update validate_id.cpp to remove the TODO for OpBranch and say that it
is already checked in validate_cfg.cpp

Also updates the message generation to avoid the extra space at the end and the backspace(!)


dneto0:
Fixed formatting, and rebased. Keeping the LGTM.

dneto0:
Squashed and pushed into master as e8814be7328

PR #1490, opened 2018-04-12
Use a bit vector in ADCE

s-perron:
The unordered_set in ADCE that holds all of the live instructions takes
a very long time to be destroyed.  In some shaders, it takes over 40% of
the time.

If we look at the unique ids of the live instructions, I believe they
are dense enough make a simple bit vector a good choice for to hold that
data.  When I check the density of the bit vector for larger shaders, we
are usually using less than 4 bytes per element in the vector, and
almost always less than 16.

So, in this commit, I introduce a simple bit vector class, and
use it in ADCE.

This help improve the compile time for some shaders on windows by the
40% mentioned above.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.h
+++ b/source/opt/aggressive_dead_code_elim_pass.h
@@ -73,8 +73,7 @@ class AggressiveDCEPass : public MemPass {
 
   // Add |inst| to worklist_ and live_insts_.
   void AddToWorklist(ir::Instruction* inst) {
-    if (!IsLive(inst)) {
-      live_insts_.Set(inst->unique_id());
+    if (!live_insts_.Set(inst->unique_id())) {
       worklist_.push(inst);
     }
   }
--- a/source/opt/aggressive_dead_code_elim_pass.h
+++ b/source/opt/aggressive_dead_code_elim_pass.h
@@ -17,6 +17,7 @@
 #ifndef LIBSPIRV_OPT_AGGRESSIVE_DCE_PASS_H_
 #define LIBSPIRV_OPT_AGGRESSIVE_DCE_PASS_H_
 
+#include <util/bit_vector.h>
 #include <algorithm>
 #include <map>
 #include <queue>
@@ -60,7 +61,7 @@ class AggressiveDCEPass : public MemPass {
 
   // Return true if |inst| is marked live.
   bool IsLive(const ir::Instruction* inst) const {
-    return live_insts_.find(inst) != live_insts_.end();
+    return live_insts_.Get(inst->unique_id());
   }
 
   // Returns true if |inst| is dead.
@@ -72,7 +73,10 @@ class AggressiveDCEPass : public MemPass {
 
   // Add |inst| to worklist_ and live_insts_.
   void AddToWorklist(ir::Instruction* inst) {
-    if (live_insts_.insert(inst).second) worklist_.push(inst);
+    if (!IsLive(inst)) {
+      live_insts_.Set(inst->unique_id());
+      worklist_.push(inst);
+    }
   }
 
   // Add all store instruction which use |ptrId|, directly or indirectly,
@@ -168,7 +172,7 @@ class AggressiveDCEPass : public MemPass {
   std::vector<ir::Instruction*> private_stores_;
 
   // Live Instructions
-  std::unordered_set<const ir::Instruction*> live_insts_;
+  utils::BitVector live_insts_;
 
   // Live Local Variables
   std::unordered_set<uint32_t> live_local_vars_;
--- a/source/util/bit_vector.h
+++ b/source/util/bit_vector.h
@@ -35,8 +35,8 @@ class BitVector {
   // Creates a bit vector contianing 0s.
   BitVector() : bits_(kInitialNumBits / kBitContainerSize, 0) {}
 
-  // Sets the |i|th bit to 1.
-  void Set(uint32_t i) {
+  // Sets the |i|th bit to 1.  Returns the |i|th bit before it was set.
+  bool Set(uint32_t i) {
     uint32_t element_index = i / kBitContainerSize;
     uint32_t bit_in_element = i % kBitContainerSize;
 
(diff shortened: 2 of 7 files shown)


source/util/bit_vector.h
```diff
@@ -0,0 +1,96 @@
...
+// limitations under the License.
+
+#ifndef LIBSPIRV_UTILS_BIT_VECTOR_H_
+#define LIBSPIRV_UTILS_BIT_VECTOR_H_
+
+#include <cassert>
+#include <iostream>
+#include <vector>
+
+#include "ilist_node.h"
```
alan-baker: Unnecessary include?

source/util/bit_vector.h
```diff
@@ -0,0 +1,96 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_UTILS_BIT_VECTOR_H_
+#define LIBSPIRV_UTILS_BIT_VECTOR_H_
+
+#include <cassert>
+#include <iostream>
```
alan-baker: Change this to just ostream. Even better, make it iosfwd and add a cpp that implements ReportDensity and includes ostream.

source/util/bit_vector.h
```diff
@@ -0,0 +1,96 @@
...
+//
+// All bits default to zero, and the upper bound is 2^32-1.
+class BitVector {
+ public:
+  // Creates a bit vector contianing 0s.
+  BitVector() : bits(1024 / sizeof(BitContainer), 0) {}
+
+  // Sets the |i|th bit to 1.
+  void Set(uint32_t i) {
+    uint32_t element_index = i / sizeof(BitContainer);
```
alan-baker: sizeof returns the number of bytes. Should be dividing by 64 here to get the right vector element. The 63rd bit is in index 0, not index 7.

source/util/bit_vector.h
```diff
@@ -0,0 +1,96 @@
...
+    uint32_t bit_in_element = i % sizeof(BitContainer);
```
alan-baker: This is also wrong for 64 bits.

s-perron:
I believe I addressed all of @alan-baker suggestions.

@dnovillo The spec does not force std::vector<bool> to be a bit vector, it could be an array of bools.  See http://en.cppreference.com/w/cpp/container/vector_bool.

dnovillo:
On Thu, Apr 12, 2018 at 4:56 PM Steven Perron <notifications@github.com>
wrote:

> I believe I addressed all of @alan-baker <https://github.com/alan-baker>
> suggestions.
>
> @dnovillo <https://github.com/dnovillo> The spec does not force
> std::vector to be a bit vector, it could be an array of bools. See
> http://en.cppreference.com/w/cpp/container/vector_bool.
>

Hm, quite right.  I haven't run into an implementation that doesn't, but
better safe than sorry.

>

source/opt/aggressive_dead_code_elim_pass.h
```diff
@@ -72,7 +73,10 @@ class AggressiveDCEPass : public MemPass {
 
   // Add |inst| to worklist_ and live_insts_.
   void AddToWorklist(ir::Instruction* inst) {
-    if (live_insts_.insert(inst).second) worklist_.push(inst);
+    if (!IsLive(inst)) {
```
dneto0: This does two operations on the bit vector when before we only did one.

Seems we could use the abstraction of "ensure the bit is set, and tell me if it wasn't set before this operation".
Either that, or update Set to return a bool saying whether it actually changed the state of the value.  (I don't know if that would be slower to burden *every* Set operation that way.

source/util/bit_vector.h
```diff
@@ -0,0 +1,80 @@
...
+
+    return (bits[element_index] &
+            (static_cast<BitContainer>(1) << bit_in_element)) != 0;
+  }
+
+  void ReportDensity(std::ostream& out);
+
+ private:
+  using BitContainer = uint64_t;
+  static const uint32_t kBitContainerSize = 64;
```
dneto0: This can be an anonymous enum:

```
enum : int {  kBitContainerSize = 64 }
```

source/util/bit_vector.h
```diff
@@ -0,0 +1,80 @@
...
+namespace spvtools {
+namespace utils {
+
+// Implements a bit vector class.
+//
+// All bits default to zero, and the upper bound is 2^32-1.
+class BitVector {
+ public:
+  // Creates a bit vector contianing 0s.
+  BitVector() : bits(1024 / kBitContainerSize, 0) {}
```
dneto0: The 1024 is kind of magic.  Better to have

```
enum : int { kInitialNumBits  = 1024 };
```

source/util/bit_vector.h
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    return (bits[element_index] &
+            (static_cast<BitContainer>(1) << bit_in_element)) != 0;
+  }
+
+  void ReportDensity(std::ostream& out);
+
+ private:
+  using BitContainer = uint64_t;
```
dneto0: Minor style thing, but I'd prefer to see these using and constant definitions right at the top. (still in private visibility)

source/util/bit_vector.h
```diff
@@ -0,0 +1,80 @@
...
+
+    if (element_index >= bits.size()) {
+      return false;
+    }
+
+    return (bits[element_index] &
+            (static_cast<BitContainer>(1) << bit_in_element)) != 0;
+  }
+
+  void ReportDensity(std::ostream& out);
```
dneto0: Please add a method comment. :-)

source/util/bit_vector.h
```diff
@@ -0,0 +1,80 @@
...
+
+ private:
+  using BitContainer = uint64_t;
+  static const uint32_t kBitContainerSize = 64;
+  std::vector<BitContainer> bits;
```
dneto0: Member data names should end in underscore, by Google style guide.

test/util/bit_vector_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+TEST(BitVectorTest, Initialize) {
+  BitVector bvec;
+  for (int i = 1; i < 10000; i *= 2) {
+    EXPECT_FALSE(bvec.Get(i));
+  }
+}
+
+TEST(BitVectorTest, Set) {
+  BitVector bvec;
+  for (int i = 3; i < 10000; i *= 2) {
```
dneto0: It's worth adding a comment saying that 10000 > 1024, so this will excecise the resize code path.

s-perron:
@dneto0 @dnovillo  Just finished the last changes that Dave asked for.  Please have another look.

dneto0:
Rebased, squashed, and pushed into master as d42f65e7c

PR #1493, opened 2018-04-15
Travis CI: change to use the default email notification behavior

antiagainst:
See https://docs.travis-ci.com/user/notifications/#Default-notification-settings


dneto0:
Rebased and pushed into master as a3bb78274

Issue #1494, opened 2018-04-16
Remove the size limit in scalar replacement.


s-perron:
For legalization, we rely on scalar replacement to be able to propagate everything.  The legalization optimizations are not optional.  We don't do them, the shader will not run.

I am currently see a number of shaders where this is happening.  Resources are not getting propagated, and we are left with illegal code because they are members of structs with more than 100 members.

Because of this I think we will have to remove that limit altogether (at least for legalization).  However, this causes compile time problems.  First it take a lot of time for scalar replacement to create the new instructions.  Then the ssa rewritter takes a long time to rewrite everything.  Finally, ADCE takes a long time to clean up all of the code.

For this issue we have two parts:

- [ ] Remove the limit in scalar replacement.
- [x] Reduce the compile caused by change.

s-perron:
I counted the number of new variables created by scalar replacement for a group of ~4,400 shader.

The current code generates an average of 4,972 new variables per shader, with a high of 69,747 for one shader.

When the limit is removed the average goes up with 8,250 with a high of 1,341,137 for one shader.

If remove the limit and undo the change in ca4457b, the average is 2,570 with a high of 66,437 for one shader.

I will try to reduce the compile time by reducing the number of symbols created by scalar replacement.  I will guess that it should be possible to get to an average of around 2,000 per shader in this set.

greg-lunarg:
How about the following approach: do limited scalar replacement and if there is illegal code at the end, re-legalize with unlimited scalar replacement. Shaders with illegal code and very large structs will compile slower, but all other shaders will compile faster.

I have thought about a similar approach with inlining: only do inlining that doesn't increase size. If illegal code exists in the end, re-optimize with full inlining.

s-perron:
That might be a good idea.  For now, I want to try to reduce the number of symbols by looking at which ones are actually used.  For example, if there is a just one load of component 1, then don't create a variable for any other component.  I just got it implemented, and it does reduce the numbers a lot.  The shader that had over 1.3 million new symbols is down below 23,000 with that change.

s-perron:
With the change I suggested above the average number of new symbols is 1,555 and the maximum is 22,401.  This has no limit on the size of the structure.  I'll check the compile time next to see how things look.

s-perron:
With the changes for #1557  and #1558, we should be good to merge #1495.

PR #1495, opened 2018-04-16
Remove the limit on struct size in SROA.

s-perron:
Removes the limit on scalar replacement because it has to happen for
legalization.

Part of #1494.


s-perron:
A couple shader is our test suite take over about 4.5 hours to compile.  This cannot be merged until that is fixed.

s-perron:
@greg-lunarg I'm about to merge this.  Let me know if you have any concerns.  The `-Os` pass should stay the same.  The only difference is if the user uses `--legalize-hlsl` or explicitly uses `--scalar-replacement=<n>`.

PR #1497, opened 2018-04-16
ADCE: Remove OpDecorateStringGOOGLE

### LINKED ISSUE #1492: ADCE doesn't understand OpDecorateStringGOOGLE
In debug build, this fails the assert in ProcessGlobalValues.
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/aggressive_dead_code_elim_pass.cpp#L653

Blocking https://github.com/google/shaderc/pull/452

dneto0:
Also fix a few failures to set "modified" status when removing
global values.

Fixes #1492

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -603,7 +603,11 @@ bool AggressiveDCEPass::ProcessGlobalValues() {
       case SpvOpDecorate:
       case SpvOpMemberDecorate:
       case SpvOpDecorateId:
-        if (IsTargetDead(annotation)) context()->KillInst(annotation);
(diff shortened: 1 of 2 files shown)


dneto0:
> Can OpDecorateStringGoogle get applied to a group?

From 
https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/GOOGLE/SPV_GOOGLE_decorate_string.asciidoc

"Target is the <id> to decorate. It can potentially be any <id> that is a forward reference, except it must not be the <id> of an OpDecorationGroup."

dneto0:
> In any case, you should add this to the decoration sorting rules in DecorationLess to ensure a good processing order with groups.

Will do.

dneto0:
Rebased, squashed, and pushed into master as 152b9a681e

PR #1501, opened 2018-04-16
 Fix broken pointer of CommonUniformElimPass

jaebaek:
In `bool CommonUniformElimPass::CommonUniformLoadElimination(ir::Function* func)`,
`insertItr` can be a `inst` that will be removed by `ReplaceAndDeleteLoad()`.

This code checks if the candidate of `insertItr` can be removed and do not update `insertItr` in the case.

### DIFF
--- a/source/opt/common_uniform_elim_pass.cpp
+++ b/source/opt/common_uniform_elim_pass.cpp
@@ -356,21 +356,21 @@ bool CommonUniformElimPass::CommonUniformLoadElimination(ir::Function* func) {
     // new load insertion point. Trying to keep register pressure down.
     if (mergeBlockId == bp->id()) {
       mergeBlockId = 0;
-
-      // Update insertItr as |bp->begin()| only when it will not be removed.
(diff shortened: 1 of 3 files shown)


source/opt/common_uniform_elim_pass.cpp
```diff
@@ -356,7 +356,20 @@ bool CommonUniformElimPass::CommonUniformLoadElimination(ir::Function* func) {
     // new load insertion point. Trying to keep register pressure down.
     if (mergeBlockId == bp->id()) {
       mergeBlockId = 0;
-      insertItr = bp->begin();
+
+      // Update insertItr as |bp->begin()| only when it will not be removed.
+      // Without this check, it can set |insertItr| as a dangling pointer.
```
s-perron: Could the original value of insertItr from line 351 have the same problem?  If so, where would you put the insertion point?

jaebaek: Yes you are right. I updated it.

jaebaek:
Rebased and pushed into master as 430a2933

PR #1502, opened 2018-04-16
Vulkan 1.1 barrier execution scope checks for OpControlBarrier

### LINKED ISSUE #1483: Vulkan 1.1: execution scope must be Subgroup for OpControlBarrier in frag, vertex, tese, geom
https://github.com/KhronosGroup/Vulkan-Docs/blob/master/appendices/spirvenv.txt#L452

alan-baker:
Fixes #1483 

Refactored the function -> entry point callers out of builtin validation and into validation state. Removed on-the-fly construction of an entry point's call tree during validation in favour of the newly maintained map. Execution model limitations are now checked in the function instead of the entry point.

Added checks that for Vulkan 1.1, subgroup execution scope must be used for Fragment, Vertex, Geometry and TessellationEvaluation models for OpControlBarrier.

Added tests.

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -817,8 +817,8 @@ void ValidationState_t::ComputeFunctionToEntryPointMapping() {
 }
 
 const std::vector<uint32_t>& ValidationState_t::FunctionEntryPoints(
-    uint32_t function) const {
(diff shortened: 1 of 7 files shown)


dneto0:
Rebased, squashed, pushed into master as 38359ba8000

PR #1508, opened 2018-04-18
DeadInsertElim: Don't revisit select phi nodes during MarkInsertChain

### LINKED ISSUE #1487: Bad optimization: LocalMultiStoreElim causes infinite loop?
See details in https://github.com/google/shaderc/issues/450 and https://github.com/mpv-player/mpv/issues/5695 

Symptom is a bad transformation in LocalMultiStoreElim, very likely in SSA rewriter.

Behaviour seems like an infinite loop when that transform is applied, but works correctly without that transform.

Sending to @dnovillo

greg-lunarg:
Such revisiting is a waste of time and can cause prohibitively exponential analysis time. Fixes #1487.

### DIFF
--- a/source/opt/dead_insert_elim_pass.cpp
+++ b/source/opt/dead_insert_elim_pass.cpp
@@ -65,10 +65,9 @@ uint32_t DeadInsertElimPass::NumComponents(ir::Instruction* typeInst) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


s-perron:
Rebased, squashed, and merged as df700f6.

PR #1510, opened 2018-04-19
Constant folding for OpVectorTimesScalar

jaebaek:
Related issue: #709


jaebaek:
Rebased and pushed into master as 48802bad

Issue #1511, opened 2018-04-19
Generalize Dead-insert-elimination


s-perron:
Dead insert elimination looks for chains of instructions starting with an OpCompositeExtract followed by a series of OpCompositeInserts.  It then tried to remove insert whose value is never used because it is not extracted, or it replaced by another insert instruction.

This is useful but it is limited by the fact that there cannot be any instruction between the inserts.

We want to extend this to be able to 
- [x] handle instructions that treat vectors on a per component basis.   
- [x] handle intervening vector shuffle instructions,
- [x] be able to replace unused elements of OpCompositeConstruct
- [ ] handle inserts and extract for other composite types.

Another issue is that dead insert elimination repeats work.  With the longer chain of instructions that the above changes will cause, it could negatively effect compile time.

To be able to avoid special casing a lot of stuff in the current implementation, we believe it is best to create a new one where we keep track of which elements of each vector is live.  The current implementation keeps track of just the insert instructions that are needed.  This change is intrusive enough to warrant a new implementation.

s-perron:
Sorry forgot that we need to handle structs before we can remove dead-insert-elimination.  I've added that to the description.

s-perron:
Done what we are doing for now.

PR #1512, opened 2018-04-19
Vector DCE

s-perron:
Introduce a pass that does a DCE type analysis for vector elements
instead of the whole vector as a single element.

It will then rewrite instructions that are not used with something else.
For example, an instruction whose value are not used, even though it is
referenced, is replaced with an OpUndef.

Contributes to #1511.

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -496,3 +496,90 @@ bool spvOpcodeIsNonUniformGroupOperation(SpvOp opcode) {
       return false;
   }
 }
+
+bool spvOpcodeIsScalarizable(SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpPhi:
+    case SpvOpCopyObject:
+    case SpvOpConvertFToU:
+    case SpvOpConvertFToS:
+    case SpvOpConvertSToF:
+    case SpvOpConvertUToF:
+    case SpvOpUConvert:
+    case SpvOpSConvert:
+    case SpvOpFConvert:
+    case SpvOpQuantizeToF16:
+    case SpvOpVectorInsertDynamic:
+    case SpvOpSNegate:
+    case SpvOpFNegate:
+    case SpvOpIAdd:
+    case SpvOpFAdd:
+    case SpvOpISub:
+    case SpvOpFSub:
+    case SpvOpIMul:
+    case SpvOpFMul:
+    case SpvOpUDiv:
+    case SpvOpSDiv:
+    case SpvOpFDiv:
+    case SpvOpUMod:
+    case SpvOpSRem:
+    case SpvOpSMod:
+    case SpvOpFRem:
+    case SpvOpFMod:
+    case SpvOpVectorTimesScalar:
+    case SpvOpIAddCarry:
+    case SpvOpISubBorrow:
+    case SpvOpUMulExtended:
+    case SpvOpSMulExtended:
+    case SpvOpShiftRightLogical:
+    case SpvOpShiftRightArithmetic:
+    case SpvOpShiftLeftLogical:
+    case SpvOpBitwiseOr:
+    case SpvOpBitwiseAnd:
+    case SpvOpNot:
+    case SpvOpBitFieldInsert:
+    case SpvOpBitFieldSExtract:
+    case SpvOpBitFieldUExtract:
+    case SpvOpBitReverse:
+    case SpvOpBitCount:
+    case SpvOpIsNan:
+    case SpvOpIsInf:
+    case SpvOpIsFinite:
+    case SpvOpIsNormal:
+    case SpvOpSignBitSet:
+    case SpvOpLessOrGreater:
+    case SpvOpOrdered:
+    case SpvOpUnordered:
+    case SpvOpLogicalEqual:
+    case SpvOpLogicalNotEqual:
+    case SpvOpLogicalOr:
+    case SpvOpLogicalAnd:
+    case SpvOpLogicalNot:
+    case SpvOpSelect:
+    case SpvOpIEqual:
+    case SpvOpINotEqual:
+    case SpvOpUGreaterThan:
+    case SpvOpSGreaterThan:
+    case SpvOpUGreaterThanEqual:
+    case SpvOpSGreaterThanEqual:
+    case SpvOpULessThan:
+    case SpvOpSLessThan:
+    case SpvOpULessThanEqual:
+    case SpvOpSLessThanEqual:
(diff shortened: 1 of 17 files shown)


source/opt/vector_dce.cpp
```diff
@@ -0,0 +1,287 @@
...
+  analysis::DefUseManager* def_use_mgr = context()->get_def_use_mgr();
+
+  current_inst->ForEachInId([&work_list, &live_elements, this, live_components,
+                             def_use_mgr](uint32_t* operand_id) {
+    ir::Instruction* operand_inst = def_use_mgr->GetDef(*operand_id);
+
+    if (HasVectorResult(operand_inst)) {
+      WorkListItem new_item;
+      new_item.instruction = operand_inst;
+      new_item.components = live_elements;
```
alan-baker: This seems wrong for OpCompositeConstruct. If I construct a float4 from 2 float2s where the only elements 2 and 3 are live, this should translate to two items: {0,0} and {1,1}, but it appears you generate {0,0,1,1} twice.

source/opt/vector_dce.cpp
```diff
@@ -0,0 +1,287 @@
...
+        if (live_component == live_components.end()) {
+          // If this instruction is not in live_components then it does not
+          // produce a vector, or it is never referenced and ADCE will remove
+          // it.  No point in trying to differentiate.
+          return;
+        }
+
+        // If no element in the current instruction is used replace it with an
+        // OpUndef.
+        if (live_component->second.Empty()) {
```
alan-baker: Maybe just refactor Type2Undef from MemPass and generate a reusable undef? Then you can just replace all uses and not have special cases for OpPhi.

source/opt/def_use_manager.cpp
```diff
@@ -265,6 +265,17 @@ void DefUseManager::EraseUseRecordsOfOperandIds(const ir::Instruction* inst) {
 
 bool operator==(const DefUseManager& lhs, const DefUseManager& rhs) {
   if (lhs.id_to_def_ != rhs.id_to_def_) {
+    for (auto l : lhs.id_to_def_) {
+      if (rhs.id_to_def_.count(l.first) == 0) {
+        std::cerr << "Id " << l.first << " has not uses in the rhs.\n";
```
dnovillo: This will create spurious output that will confuse users.

source/opt/vector_dce.cpp
s-perron: This depends on what values are passed in for `live_elements`.  For the `OpCompositeConstruct` case (line 87), we pass in `all_components_live_`, which is 16 elements live for both inputs.  It is not as precise as it could be, which is why I left the TODO on line 85, but it is safe.

source/opt/vector_dce.cpp
```diff
@@ -0,0 +1,287 @@
...
+void VectorDCE::FindLiveComponents(ir::Function* function,
+                                   LiveComponentMap* live_components) {
+  std::vector<WorkListItem> work_list;
+
+  // Prime the work list.  We will assume that any instruction that does
+  // not result in a vector is live.
+  //
+  // TODO: If we want to remove all of the dead code with the current
+  // implementation, we will have to iterate VectorDCE and ADCE.  That can be
+  // avoided if we treat scalar results as if they are vectors with 1 element.
```
dnovillo: So, I wonder if at some point it will make better sense to give up and implement a proper memory SSA based on FUD chains so that we can do things like this and array copyprop more easily.  We've talked about this offline, and none of us are convinced.  But adding these ad-hoc passes is a long-term technical debt as well.

At some point, maybe better to use a single framework for reasoning about memory.  Passes that need it, build it, use it and tear it down.

No.  I'm not volunteering just yet ;)

source/opcode.cpp
```diff
@@ -454,3 +454,90 @@ bool spvOpcodeIsBaseOpaqueType(SpvOp opcode) {
       return false;
   }
 }
+
+bool spvOpcodeWithPerComponentResult(SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpPhi:
```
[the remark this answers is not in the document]
s-perron: done

source/opt/vector_dce.cpp
s-perron: I can can generate a global OpUndef and reuse the same one.  I did not do that because I wanted to reduce the amount if time spent updating the def-use chains.  With this change, there are very few changes to the def-use information.

I will change it.  I think it will generate cleaner code.  However, I won't factor out the stuff from mempass.  I don't want to muddy this PR.  We can do that on its own later.

source/opt/vector_dce.cpp
s-perron: Sure, but we would need to come up with a good design for it to make sure it can be used.

source/opcode.h
```diff
@@ -121,4 +121,8 @@ bool spvOpcodeIsBaseOpaqueType(SpvOp opcode);
 
 // Returns true if the given opcode is a non-uniform group operation.
 bool spvOpcodeIsNonUniformGroupOperation(SpvOp opcode);
+
+// Returns true if the result of an instruction with this opcode are computed
```
alan-baker: Comment is out of date.

s-perron:
@greg-lunarg It is still missing the iterative aspect.  We do just one pass, while dead-insert-elim would do the DCEInst, and then rerun until it reached a fixed point.  That would be the main difference.  I want to try to get these case by treating scalars as a single element vector.

greg-lunarg:
OK, thanks. As it turns out, when I turn off the closure mechanism for DIE, my tests still pass, so I think your basic algorithm still might be missing something. I will look at it.

Just as long as you don't rip out DIE until I can verify that VDCE is fully covering my shaders.

greg-lunarg:
The differences I am seeing are due to the fact that DIE handles structs and vectors where VDCE only handles vectors. So we will need to keep DIE around for the moment as I have a number of real-life shaders that require that structs are handled.

For the moment it might make sense for DIE to drop processing of vectors. That might reduce some redundant analysis.

But in the near future, I am thinking we are going to have to do similar optimization on OpCompositeConstruct (identify unused elements and replace them with Undefs), which could also be a struct. This would require capability similar to what is being done in VDCE. So it feels like in the long term we are going to have to make DIE more like VDCE or adapt VDCE to deal with structs. It feels like it will make the most sense for these to merge at some point.

s-perron:
@greg-lunarg I have have left DIE in for now.  We can discuss how to merge the two in a different form.

Issue #1500, opened 2018-04-16
Local Single Block Elim is not eliminating stores


greg-lunarg:
In local-single-block-elim, the propagation and elimination of the loads are happening, but the elimination of the stores and the variables are not.

These stores are preventing propagation of some values that could be propagated, resulting in a possible loss of optimization and legalization.

This is also a bit of a waste because this is leaving variables and stores to be re-analyzed by following passes, specifically local-single-store-elim and local-multi-store-elim.

For example, I have a real-live shader where a variable is stored-loaded-store-loaded-stored in the same block and then loaded in a subsequent block.  Since the stores are not eliminated, it prevents single-store-elim from  propagating a value that could have been propagated if the stores had been eliminated. This could make a difference for optimization and legalization.

In addition, the analysis by single-store-elim is wasted and has to be repeated by multi-store-elim, which ultimately propagates the value, but possibly too late.

ADCE does not remove these stores because it does not do Write-after-Write elimination.

It feels like a little additional work in this pass could cost-effectively eliminate these stores and enhance subsequent analysis.

dneto0:
I haven't looked at SSA-rewriter in detail, but doesn't its back-to-front conversion to SSA registers cause the first two stores to look dead?  I guess a concrete example would help think this one through.

greg-lunarg:
Yes, SSA rewrite will eventually propagate the value in this example, remove the final load and enable these stores to be removed. But SSA rewrite is much farther down the recipe. If this propagation can be done earlier by single-store, it potentially can enable other optimizations.

Here is a quick C example:

int a,b,c;
int x,y;

main() {
   int i;
   i += a;
   i += b;
   if (c > 0)
     x = i;
   else
     y = i;
}

greg-lunarg:
And a similar shader:

```
layout(location=0) in float a;

layout(location=0) out float x;
layout(location=1) out float y;

main() {
   float f = 0.0;
   f += 0.1;
   f += 0.1;
   if (a > 0.0)
      x = f;
   else
      y = f;
}
```

s-perron:
I believe this is done.

PR #1514, opened 2018-04-19
LocalSingleBlockElim: Add store-store elimination

greg-lunarg:
Eliminate unused store to variable if followed by store to same
variable in same block.

Most significantly, this cleans up stores made unused by this pass.
These useless stores can inhibit subsequent optimizations, specifically
LocalSingleStoreElim. Eliminating them can make subsequent optimization
more effective.

This fixes issue #1500

### DIFF
--- a/test/opt/local_single_block_elim.cpp
+++ b/test/opt/local_single_block_elim.cpp
@@ -183,87 +183,176 @@ OpFunctionEnd
       predefs + before, predefs + after, true, true);
 }
 
-TEST_F(LocalSingleBlockLoadStoreElimTest,
-       NoStoreElimIfInterveningAccessChainLoad) {
+TEST_F(LocalSingleBlockLoadStoreElimTest, StoreStoreElim) {
   //
-  // Note that even though the Load to %v is eliminated, the Store to %v
-  // is not eliminated due to the following access chain reference.
+  // Note first store to v is eliminated
   //
-  // #version 140
+  // #version 450
   //
-  // in vec4 BaseColor;
-  // flat in int Idx;
+  // layout(location = 0) in vec4 BaseColor;
+  // layout(location = 0) out vec4 OutColor;
   //
   // void main()
   // {
   //     vec4 v = BaseColor;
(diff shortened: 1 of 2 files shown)


test/opt/local_single_block_elim.cpp
```diff
@@ -183,87 +183,176 @@ OpFunctionEnd
...
+  //
+  // layout(location = 0) in vec4 BaseColor;
+  // layout(location = 0) out vec4 OutColor;
+  //
+  // void main()
+  // {
+  //     vec4 v = BaseColor;
+  //     v = v * 0.5;
+  //     OutColor = v;
+  // }
```
dnovillo: If we convert to SSA and run DCE, this should eliminate the fist store to ```v``` does it not?

dnovillo: It does, actually.  Running ```spirv-opt --ssa-rewrite --eliminate-dead-code-aggressive``` on the test case results in:

```
       %main = OpFunction %void None %7
         %14 = OpLabel
         %15 = OpLoad %v4float %BaseColor
         %17 = OpVectorTimesScalar %v4float %15 %float_0_5
               OpStore %OutColor %17
               OpReturn
               OpFunctionEnd
```

greg-lunarg: Yes, SSARewrite functionally can replace SingleBlockElim and SingleStoreElim, but that is not the way the current recipes are written.

Several months ago this change was made to the recipes and I had to go back and demonstrate that running SingleBlock and SingleStore before SSARewrite reduced overall compile time. This is also the way that LLVM does it, at least the last time I looked.

So I would request that if you decide to try this again you look at overall compile time. Until that time I am requesting this very simple and inexpensive improvement to SingleBlock.

dnovillo: Oh, interesting.  If this reduces overall compile-time, then we shouldn't even consider getting rid of SBE and SSE.  I'll merge your change.  Thanks!

PR #1516, opened 2018-04-20
Dominator analysis interface clean-up

Naghasan:
Remove the need to provide a CFG object when querying a dominator/post-dominator.
Updated all uses of the functions.


dnovillo:
Pushed to master as efc506192943262551728cbc8e6394e36ca2e348

PR #1517, opened 2018-04-20
Another fix for old XCode: std::set explicit ctor in test code

dneto0:


dneto0:
This problem is blocking Android NDK build.
LGTM.  I'm accountable for making this work.

PR #1518, opened 2018-04-23
Add Dead insert elmination back in.

s-perron:
This should have been part of the previous PR that introduced vector DCE.  It address @greg-lunarg comments.


s-perron:
@alan-baker I'm open to that. Do you have a particular reason?

alan-baker:
My thinking was is vector dce had a better run time, then run it first and let dead insert elim do the cleanup of the remainder.

s-perron:
That makes sense.  I made the chnage.

PR #1519, opened 2018-04-23
Support for loop fission.

davors72:
Pass to split loops when the loop exceeds a certain register pressure threshold. Threshold can be provided by the user as an integer value for total number of registers or a custom functor. Loops will check if they would violate any data dependency rules by splitting and will leave control flow intact for dead code elimination to remove afterwards.

### DIFF
--- a/source/opt/loop_fission.cpp
+++ b/source/opt/loop_fission.cpp
@@ -1,0 +1,508 @@
+// Copyright (c) 2018 Google LLC.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "opt/loop_fission.h"
+#include "opt/register_pressure.h"
+
+// Implement loop fission with an optional parameter to split only
+// if the register pressure in a given loop meets a certain criteria. This is
+// controlled via the constructors of LoopFissionPass.
+//
+// 1 - Build a list of loops to be split, these are top level loops (loops
+// without child loops themselves) which meet the register pressure criteria, as
+// determined by the ShouldSplitLoop method of LoopFissionPass.
+//
+// 2 - For each loop in the list, group each instruction into a set of related
+// instructions by traversing each instructions users and operands recursively.
+// We stop if we encounter an instruction we have seen before or an instruction
+// which we don't consider relevent (i.e OpLoopMerge). We then group these
+// groups into two different sets, one for the first loop and one for the
+// second.
+//
+// 3 - We then run CanPerformSplit to check that it would be legal to split a
+// loop using those two sets. We check that we haven't altered the relative
+// order load/stores appear in the binary and that we aren't breaking any
+// dependency between load/stores by splitting them into two loops. We also
+// check that none of the OpBranch instructions are dependent on a load as we
+// leave control flow structure intact and move only instructions in the body so
+// we want to avoid any loads with side affects or aliasing.
+//
+// 4 - We then split the loop by calling SplitLoop. This function clones the
+// loop and attaches it to the preheader and connects the new loops merge block
+// to the current loop header block. We then use the two sets built in step 2 to
+// remove instructions from each loop. If an instruction appears in the first
+// set it is removed from the second loop and vice versa.
+//
+// 5 - If the multiple split passes flag is set we check if each of the loops
+// still meet the register pressure criteria. If they do then we add them to the
+// list of loops to be split (created in step one) to allow for loops to be
+// split multiple times.
+//
+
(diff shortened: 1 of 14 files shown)


source/opt/loop_fission.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+    // Add the instruction to the set of instructions already seen, this breaks
+    // recursion and allows us to ignore certain instructions.
+    seen_instructions_.insert(user);
+
+    // We want to traverse the operands of branches but we don't want to include
+    // them in the set.
+    if (!user->IsBranch()) {
+      inst_set.insert(user);
+    }
```
dnovillo: But if the branch itself is not included, how is this going to help?  I'm missing something here.  ISTM that you'd be splitting the branches of the conditional, but not the predicate itself.

source/opt/loop_fission.cpp
```diff
@@ -0,0 +1,509 @@
...
+}
+
+bool LoopFissionImpl::CanPerformSplit() {
+  // Return false if any of the condition instructions in the loop depend on a
+  // load.
+  if (load_used_in_condition_) {
+    return false;
+  }
+
+  // Build a list of all parent loops of this loop. Loop dependency analysis
```
dnovillo: s/dependency/dependence/

source/opt/loop_fission.cpp
```diff
@@ -0,0 +1,509 @@
...
+  // Traverse the def use chain of |inst| and add the users and uses of |inst|
+  // which are in the same loop to the |returned_set|.
+  void TraverseUseDef(ir::Instruction* inst,
+                      std::set<ir::Instruction*>* returned_set,
+                      bool ignore_phi_users = false, bool report_loads = false);
+
+  // We group the instructions in the block into two different groups, the
+  // instructions to be kept in the original loop and the ones to be cloned into
+  // the new loop. As the cloned loop is attached to the preheader it will be
+  // the first loop and the second loop will be the original.
```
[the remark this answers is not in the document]
davors72: I agree, that's better.

source/opt/loop_fission.cpp
```diff
@@ -0,0 +1,509 @@
...
+  // related to the loop condition and any if conditions should any of those
+  // instructions be a load.
+  bool load_used_in_condition_;
+};
+
+bool LoopFissionImpl::MovableInstruction(const ir::Instruction& inst) const {
+  return inst.opcode() == SpvOp::SpvOpLoad ||
+         inst.opcode() == SpvOp::SpvOpStore ||
+         inst.opcode() == SpvOp::SpvOpSelectionMerge ||
+         inst.opcode() == SpvOp::SpvOpPhi || inst.IsOpcodeCodeMotionSafe();
```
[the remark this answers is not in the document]
davors72: This function is just to white list instructions which don't have side effects, we do the data dependence checks after. The IsOpcodeCodeMotionSafe provides a list of instructions which that pass considered to be safe to move so I reused it and added the above exceptions for the instructions not included as the list was mostly the same

source/opt/loop_fission.cpp
davors72: When we group the instructions we first loop over the conditional control flow instructions and call each instruction with this traverse function to ensure that it is added to the seen_instructions_ set. We just ignore the inst_sets produced by that pass but it will mean that any, say, OpBranchConditionals and associated instructions will be ignored by the actual grouping pass which creates the sets. 

This check was to then stop any unconditional branches which would not be part of the effective blacklist created by the seen_instructions_ set. Looking at it now, I think it's confusing to break up the logic like that so I've moved the check to the same part that we deal with conditional branches.

source/opt/loop_fission.cpp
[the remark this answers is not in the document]
dnovillo: Ah, ok, so IsOpcodeCodeMotionSafe() may have to be renamed to IsOpcodeSideEffectFree (as it is checking for a stronger invariant).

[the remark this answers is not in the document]
dnovillo: This does not need to be done in this PR.  We can always rename it later.

source/opt/loop_fission.cpp
```diff
@@ -0,0 +1,508 @@
...
+          // split, split it again.
+          if (ShouldSplitLoop(*loop, c)) new_loops_to_split.push_back(loop);
+        }
+      }
+
+      // If the split multiple times flag has been set add the new loops which
+      // meet the splitting criteria into the list of loops to be split on the
+      // next iteration.
+      if (split_multiple_times_) {
+        inner_most_loops = std::move(new_loops_to_split);
```
dnovillo: Hmm, doesn't this invalidate the loop iterator?  I'm not actually sure it does.

davors72: I don't think it should invalidate, since it's just moving a list of pointers to loops managed by the analysis so it's not actually touching anything internal to the loop or analysis. But we aren't iterating directly over the loop at this point anyway, we're iterating over the inner_most_loops list to sidestep the iterator invalidation issue

dnovillo:
Re-based, re-tested and submitted as 9a5dd6fe88882005ea0433857e688581608bb210

dnovillo:
Re-based, re-tested and pushed upstream as 1dc245806082c4f4e79fec7e1aef57720f0e1221

dnovillo:
Disregard my previous comment.  I meant to update #1515.

PR #1523, opened 2018-04-24
Fixes #1521: PadToWord() should use std::move() in && variant

### LINKED ISSUE #1521: Use std::move(str) suggested with -Wreturn-std-move
Error:

.../source/util/bit_stream.h:157:10: error: local variable 'str' will be copied despite being returned by name [-Werror,-Wreturn-std-move]
  return str;

cdwfs:


alan-baker:
Rebased and submitted as 72524db2dec0cf13a98c4b3e75ff10fcbdd86ac8

PR #1525, opened 2018-04-25
Fold OpVectorTimesScalar and OpPhi better.

s-perron:
If one of the operands to an OpVectorTimesScalar instruction is zero,
then the result will be the 0 vector.  Currently we do not fold the
insturction unless both operands are constants.  This change fixes that.

We also allow folding of OpPhi instructions where the incoming values
are either an OpUndef or the OpPhi instruction itself.  As with other
cases, this can be simplified to the OpUndef.

Contributes to #709.

### DIFF
--- a/test/opt/fold_test.cpp
+++ b/test/opt/fold_test.cpp
@@ -3855,14 +3855,13 @@ INSTANTIATE_TEST_CASE_P(MergeMulTest, MatchingInstructionFoldingTest,
       "OpFunctionEnd",
     2, true),
   // Test case 19: Fold OpVectorTimesScalar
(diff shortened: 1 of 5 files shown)


test/opt/fold_test.cpp
```diff
@@ -4767,4 +4784,4 @@ ::testing::Values(
         3, true)
 ));
 #endif
-}  // anonymous namespace
+}  // anonymous namespace
```
alan-baker: Seems odd.

s-perron: I don't know what is going on there.  I tried to undo it, but it did not seem to work.

PR #1529, opened 2018-04-26
Fixes handling of kill and unreachable ops in inlining.

### LINKED ISSUE #1527: Wrong  transformations for OpKill
For the following SPIR-V:

```llvm
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpSource HLSL 600
               OpName %bb_entry "bb.entry"
               OpName %src_main "src.main"
               OpName %main "main"
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
          %5 = OpLabel
          %6 = OpFunctionCall %void %src_main
               OpReturn
               OpFunctionEnd
   %src_main = OpFunction %void None %3
   %bb_entry = OpLabel
               OpKill
               OpFunctionEnd
```

spirv-opt --legalize-hlsl -O is turning it into

```llvm
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpSource HLSL 600
               OpName %main "main"
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
          %4 = OpLabel
               OpKill
               OpReturn
               OpFunctionEnd
```
which fails validation.

Originally reported by @Gaspard-- as https://github.com/Microsoft/DirectXShaderCompiler/issues/1250.

alan-baker:
Fixes #1527

* Adds handling for copying OpKill and OpUnreachable and forces the
generation of a new basic block
* Adds tests to check

### DIFF
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -289,6 +289,16 @@ void InlinePass::GenInlineCode(
       case SpvOpVariable:
         // Already processed
         break;
+      case SpvOpUnreachable:
+      case SpvOpKill: {
+        // Generate a return label so that we split the block with the function
+        // call. Copy the terminator into the new block.
+        if (returnLabelId == 0) returnLabelId = this->TakeNextId();
+        std::unique_ptr<ir::Instruction> terminator(
+            new ir::Instruction(context(), cpi->opcode(), 0, 0, {}));
+        new_blk_ptr->AddInstruction(std::move(terminator));
+        break;
+      }
       case SpvOpLabel: {
         // If previous instruction was early return, insert branch
         // instruction to return block.
--- a/test/opt/inline_test.cpp
+++ b/test/opt/inline_test.cpp
@@ -2587,6 +2587,126 @@ TEST_F(InlineTest, SetParent) {
   }
 }
 
+#ifdef SPIRV_EFFCEE
+TEST_F(InlineTest, OpKill) {
+  const std::string text = R"(
+; CHECK: OpFunction
+; CHECK-NEXT: OpLabel
+; CHECK-NEXT: OpKill
+; CHECK-NEXT: OpLabel
+; CHECK-NEXT: OpReturn
+; CHECK-NEXT: OpFunctionEnd
+OpCapability Shader
+OpMemoryModel Logical GLSL450
+OpEntryPoint Fragment %main "main"
+%void = OpTypeVoid
+%voidfuncty = OpTypeFunction %void
+%main = OpFunction %void None %voidfuncty
+%1 = OpLabel
+%2 = OpFunctionCall %void %func
+OpReturn
+OpFunctionEnd
+%func = OpFunction %void None %voidfuncty
+%3 = OpLabel
+OpKill
+OpFunctionEnd
+)";
+
+  SinglePassRunAndMatch<opt::InlineExhaustivePass>(text, true);
+}
+
+TEST_F(InlineTest, OpKillWithTrailingInstructions) {
+  const std::string text = R"(
+; CHECK: OpFunction
+; CHECK-NEXT: OpLabel
+; CHECK-NEXT: [[var:%\w+]] = OpVariable
+; CHECK-NEXT: OpKill
+; CHECK-NEXT: OpLabel
+; CHECK-NEXT: OpStore [[var]]
+; CHECK-NEXT: OpReturn
+; CHECK-NEXT: OpFunctionEnd
+OpCapability Shader
+OpMemoryModel Logical GLSL450
+OpEntryPoint Fragment %main "main"
+%void = OpTypeVoid
+%bool = OpTypeBool
+%true = OpConstantTrue %bool
+%bool_func_ptr = OpTypePointer Function %bool
+%voidfuncty = OpTypeFunction %void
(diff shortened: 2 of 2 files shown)


source/opt/inline_pass.cpp
```diff
@@ -289,6 +289,16 @@ void InlinePass::GenInlineCode(
       case SpvOpVariable:
         // Already processed
         break;
+      case SpvOpUnreachable:
+      case SpvOpKill: {
+        // Generate a return label so that we split the block with the function
+        // call. Copy the terminator into the new block.
+        if (returnLabelId == 0) returnLabelId = this->TakeNextId();
+        std::unique_ptr<ir::Instruction> terminator(
+            new ir::Instruction(context(), cpi->opcode(), 0, 0, {}));
```
s-perron: This would be easier to read if you used the `clone` function in the instruction class.

greg-lunarg:
While I will be the first to admit that this logic is not the easiest to work with, it feels like there might be a way to avoid creating a new (dead) block with a similar amount of additional logic. Let me stare at the code a bit more.

greg-lunarg:
Actually, I think this solution may actually generate invalid SPIR-V. Consider the case where the called function contains a KILL but it is not the last instruction in the function before the OpFunctionEnd. Also this does not fix the case where the last instruction before the OpFunctionEnd is a non-Return terminator like OpBranch, which would also be a problem.

I do agree with your general solution, though. And yes, there is a potential to generate a dead block after the call but we will just have to rely on following passes to clean it up. Trying to avoid this would just be overly complicated.

Here is a solution that I think will work:

Start over.

Create a new boolean prevInstNonReturnTerminator. Initialize to false.

Also, at the beginning of the OpLabel case, prevInstNonReturnTerminator= false.

At the end of the default case, if the opcode is a terminator, prevInstNonReturnTerminator= true. We know it is not a Return because those are caught by their own cases.

Finally, at the beginning of the OpFunctionEnd case:

if (prevInstNonReturnTerminator && returnLabelId == 0) returnLabelId = this->TakeNextId();

If it is easier, I can just create my own pull request. Let me know how you want to proceed.

alan-baker:
I don't see the problems you're describing. Do you have a case that doesn't work?  I just tried one where the branch is the last instruction before the OpFunctionEnd and it seems ok.
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
       %void = OpTypeVoid
       %bool = OpTypeBool
       %true = OpConstantTrue %bool
          %5 = OpTypeFunction %void
%_ptr_Function_bool = OpTypePointer Function %bool
          %1 = OpFunction %void None %5
          %7 = OpLabel
          %8 = OpVariable %_ptr_Function_bool Function
          %9 = OpLoad %bool %8
         %10 = OpFunctionCall %void %11
               OpStore %8 %9
               OpReturn
               OpFunctionEnd
         %11 = OpFunction %void None %5
         %12 = OpLabel
               OpBranch %13
         %13 = OpLabel
               OpLoopMerge %14 %15 None
               OpBranch %16
         %16 = OpLabel
               OpKill
         %14 = OpLabel
               OpReturn
         %15 = OpLabel
               OpBranch %13
               OpFunctionEnd
```

Turns into (minus the inlined function):
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
       %void = OpTypeVoid
       %bool = OpTypeBool
       %true = OpConstantTrue %bool
          %5 = OpTypeFunction %void
%_ptr_Function_bool = OpTypePointer Function %bool
          %1 = OpFunction %void None %5
          %7 = OpLabel
          %8 = OpVariable %_ptr_Function_bool Function
          %9 = OpLoad %bool %8
               OpBranch %17
         %17 = OpLabel
               OpLoopMerge %18 %19 None
               OpBranch %20
         %20 = OpLabel
               OpKill
         %18 = OpLabel
               OpBranch %21
         %19 = OpLabel
               OpBranch %17
         %21 = OpLabel
               OpStore %8 %9
               OpReturn
               OpFunctionEnd
```

I can add this as a unit test.

dneto0:
@greg-lunarg : I seems to me your solution is functionally the same, but you're trying to frame it more generically.  There are only so many terminator instructions, and the only ones that "leave" the function are the returns, and Kill and Unreachable.   @alan-baker 's proposed change explicitly enumerates those in the big switch statement.   

It would possibly improve the MR if a comment were added to emphasize the generic reason *why* OpUnreachable and OpKill require this behaviour, and that therefore any potential future instruction that had the same property of "leaving" the CFG should also be treated similarly.

How about:

> Handle the case of a block terminator that does not transfer control to another block in the CFG.  Aside from OpReturn and OpReturnValue (handled elsewhere), only OpUnreachable and OpKill are like this.

greg-lunarg:
My apologies. Actually running the cases I was worried about revealed forgotten logic in the original code which covered those cases.

Given that, it seems the following similar but slightly simpler logic passes all the inlining tests:

Start over. Add the following at the beginning of the OpFunctionEnd case:

// If previous instruction was not return, we will need a label
if (!prevInstWasReturn && returnLabelId == 0)
  returnLabelId = this->TakeNextId();

greg-lunarg:
Take or leave as you wish.

alan-baker:
Rebased, squashed and submitted as 4246abdc74a18c60b49316d5d9863ca2be61127a

PR #1535, opened 2018-04-30
Disallow entries into a construct except through the header

### LINKED ISSUE #1281: Validator fails to identify invalid structured control flow
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpSource GLSL 410
               OpName %main "main"
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
        %int = OpTypeInt 32 1
%_ptr_Function_int = OpTypePointer Function %int
      %int_0 = OpConstant %int 0
      %int_2 = OpConstant %int 2
       %bool = OpTypeBool
      %int_5 = OpConstant %int 5
      %int_1 = OpConstant %int 1
          %5 = OpLabel
               OpBranch %10
         %10 = OpLabel
               OpBranch %14
         %14 = OpLabel
         %18 = OpSLessThan %bool %int_0 %int_2
               OpBranch %11
         %11 = OpLabel
         %21 = OpIEqual %bool %int_0 %int_5
               OpSelectionMerge %23 None
               OpBranchConditional %21 %22 %23
         %22 = OpLabel
               OpBranch %12
         %23 = OpLabel
               OpBranch %13
         %13 = OpLabel
         %27 = OpIAdd %int %int_0 %int_1
               OpBranch %29
         %29 = OpLabel
               OpBranch %31
         %31 = OpLabel
         %32 = OpSLessThan %bool %27 %int_2
               OpBranch %33
         %33 = OpLabel
         %34 = OpIEqual %bool %27 %int_5
               OpSelectionMerge %36 None
               OpBranchConditional %34 %35 %36
         %35 = OpLabel
               OpBranch %12
         %36 = OpLabel
               OpBranch %37
         %37 = OpLabel
         %38 = OpIAdd %int %27 %int_1
               OpBranch %12
         %12 = OpLabel
               OpReturn
               OpFunctionEnd
```

cfg

The validator succeeds for the given SPIR-V.

alan-baker:
Fixes #1281 

* Constructs can now produce a set of their blocks.
 * Uses block depth to help detect breaks and continues
* Validation check does not consider unreachable predecessors
* Fixed an invalid test in common uniform elim tests
* Added a test for the driving testcase
 * Difficult to produce tests for loop and continue constructs due to unreachability


alan-baker:
I will check CTS against this, but have not yet.

source/val/construct.cpp
```diff
@@ -64,4 +67,52 @@ const BasicBlock* Construct::exit_block() const { return exit_block_; }
 BasicBlock* Construct::exit_block() { return exit_block_; }
 
 void Construct::set_exit(BasicBlock* block) { exit_block_ = block; }
+
+Construct::ConstructBlockSet Construct::blocks(Function* function) const {
+  auto header = entry_block();
+  auto merge = exit_block();
```
s-perron: Since a precondition for this function is that the exit block is set, can we assert that `merge != nullptr`?  Or are asserts not done in the validator?

source/val/construct.cpp
```diff
@@ -13,10 +13,13 @@
 // limitations under the License.
 
 #include "val/construct.h"
+#include "val/function.h"
 
 #include <cassert>
 #include <cstddef>
+#include <unordered_set>
 
+#include <iostream>
```
s-perron: I don't think this include is needed.  If it is, group it with the group above it.

source/val/construct.cpp
alan-baker: Added an assert for header and merge.

source/val/basic_block.h
```diff
@@ -205,7 +205,7 @@ class BasicBlock {
   std::vector<BasicBlock*> successors_;
 
   /// The type of the block
-  std::bitset<kBlockTypeCOUNT - 1> type_;
+  std::bitset<kBlockTypeCOUNT> type_;
```
dneto0: That must have been painful to find....

source/val/construct.cpp
```diff
@@ -64,4 +66,53 @@ const BasicBlock* Construct::exit_block() const { return exit_block_; }
 BasicBlock* Construct::exit_block() { return exit_block_; }
 
 void Construct::set_exit(BasicBlock* block) { exit_block_ = block; }
+
+Construct::ConstructBlockSet Construct::blocks(Function* function) const {
+  auto header = entry_block();
+  auto merge = exit_block();
+  assert(header && merge);
```
dneto0: It's a little thing, but I would have done two asserts, so it's easier to see what the failure was: missing header? or missing merge.

Not a request to change this.

source/val/construct.h
```diff
@@ -91,6 +101,12 @@ class Construct {
     return type_ == ConstructType::kLoop || type_ == ConstructType::kSelection;
   }
 
+  using ConstructBlockSet = std::set<BasicBlock*, less_than_id>;
+
+  // Returns the basic blocks in this construct. This function should not
+  // be called before the exit block is set.
```
dneto0: Also requires dominators to have been computed.  That's because it relies on Function::GetBlockDepth

source/val/construct.cpp
```diff
@@ -64,4 +66,53 @@ const BasicBlock* Construct::exit_block() const { return exit_block_; }
...
+      // Entered a corresponding construct.
+      continue;
+    }
+
+    int block_depth = function->GetBlockDepth(block);
+    if (block_depth < header_depth) {
+      // Broke to outer construct.
+      continue;
+    }
+    if (block_depth == header_depth && type() == ConstructType::kSelection &&
```
dneto0: I'm a bit fuzzy why you check for kSelection here.   Could use a comment.

test/val/val_cfg_test.cpp
```diff
@@ -1423,6 +1423,33 @@ TEST_F(ValidateCFG, OpReturnInNonVoidFunc) {
           "OpReturn can only be called from a function with void return type"));
 }
```
dneto0: Please add a test case which branches into a loop construct (bypassing the header).

dneto0:
Also, thanks for your patience in waiting for my review.
Please rebase against master.

test/val/val_cfg_test.cpp
alan-baker: I'm not sure this can be done with the implementation of the validator. When restricted to reachable blocks every case I try runs afoul of backedge rules. Branching around the header is no good because it prevents blocks from being part of the loop construct. Branching back into the loop is an extra backedge. Continuing into the middle of the loop construct is caught already.

dneto0: Ok.  Sounds good to me.

source/validate_cfg.cpp
```diff
@@ -242,6 +242,25 @@ spv_result_t StructuredControlFlowChecks(
                    _.getIdName(merge->id()), "is not post dominated by");
       }
     }
+
+    // Check that for all non-header blocks, all predecessors are within this
+    // construct.
+    Construct::ConstructBlockSet construct_blocks = construct.blocks(function);
```
dneto0: This constructs the block list for each construct, which will traverse all the blocks in the nested constructs as well.
There's a potential for somewhat pathological behaviour (quadratic?) but dependent heavily on construct nesting depth.
I think this is ok and alternatives would involve state (memoization).
Ok for now.

dneto0:
The AppVeyor failure looks to be spurious.   I think this is ok to merge.

PR #1537, opened 2018-04-30
Fold divide and multiply by same value.

### LINKED ISSUE #1531: Fold multiply and divide by same value.
We are not folding the following code:

```
         %75 = OpFDiv %v2float %73 %74
         %76 = OpFMul %v2float %75 %74
```

This is leaving lots of unnecessary code around.

s-perron:
We want to fold code like (x*y)/x and other permutations of this.

Fixes #1531.

### DIFF
--- a/source/opt/folding_rules.cpp
+++ b/source/opt/folding_rules.cpp
@@ -619,18 +619,36 @@ FoldingRule MergeMulMulArithmetic() {
 // 2 * (2 / x) = 4 / x
 // (x / 2) * 2 = x * 1
 // (2 / x) * 2 = 4 / x
+// (y / x) * x = y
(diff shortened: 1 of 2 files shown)


s-perron:
According to https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#spirvenv

The Nan is not a problem because of the line "NaNs may not be generated. Instructions that operate on a NaN may not result in a NaN."

The denorms might be an issue.  There is some wiggle room because it says "Any denormalized value input into a shader or potentially generated by any instruction in a shader may be flushed to 0."  This indicates (but does not formally allow) a lack of precision when dealing with denorms.  We can try to get clarification on this.

PR #1539, opened 2018-05-02
Added GetContinueBlock to loop class

davors72:
Previously, the loop class used the terms latch and continue block
interchangeably. This patch splits the two and corrects and tests some
uses of the old uses of GetLatchBlock.

### DIFF
--- a/source/opt/loop_descriptor.cpp
+++ b/source/opt/loop_descriptor.cpp
@@ -214,6 +214,7 @@ Loop::Loop(IRContext* context, opt::DominatorAnalysis* dom_analysis,
   assert(context);
   assert(dom_analysis);
   loop_preheader_ = FindLoopPreheader(dom_analysis);
+  loop_latch_ = FindLatchBlock();
 }
 
 BasicBlock* Loop::FindLoopPreheader(opt::DominatorAnalysis* dom_analysis) {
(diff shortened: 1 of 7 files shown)


source/opt/loop_descriptor.cpp
```diff
@@ -321,6 +327,26 @@ void Loop::SetPreHeaderBlock(BasicBlock* preheader) {
   loop_preheader_ = preheader;
 }
 
+ir::BasicBlock* Loop::FindLatchBlock() {
+  ir::CFG* cfg = context_->cfg();
+
+  opt::DominatorAnalysis* dominator_analysis =
+      context_->GetDominatorAnalysis(loop_header_->GetParent());
+
+  // Find the backedge to the header which should be dominated by the loop
```
dnovillo: This phrasing is a bit confusing.  The backedge is not dominated by the continue target.  It's the latch block.  And the latch block contains the backedge to the header, right?

source/opt/loop_descriptor.h
```diff
@@ -340,6 +351,8 @@ class Loop {
   // Returns the context associated this loop.
   IRContext* GetContext() const { return context_; }
 
+  ir::BasicBlock* FindLatchBlock();
```
dnovillo: Document this, please.

source/opt/loop_descriptor.cpp
davors72: Yeah it's a bit misleading, updated it now

dnovillo:
Thanks!  Re-basing  now.

PR #1540, opened 2018-05-02
Allow hoisting code in if-conversion. 

### LINKED ISSUE #1526: Fold Phi instructions where the value (but not the id) is the same on all paths.
```
               OpSelectionMerge %7200 None
               OpBranchConditional %7199 %7201 %7202
       %7201 = OpLabel
       %7205 = OpCompositeConstruct %v4float %float_0 %16282 %float_0 %float_1
               OpBranch %7200
       %7202 = OpLabel
       %7211 = OpCompositeConstruct %v4float %float_0 %16282 %float_0 %float_1
               OpBranch %7200
       %7200 = OpLabel
      %16275 = OpPhi %v4float %7205 %7201 %7211 %7202
```

Consider the snippet of code above.  The same value is computed on both side of the branch.  This means that regardless of which direction the branch takes %16275 will always get the same value.  We should be able to replace the Phi with the instruction to compute the value.  Then then the branch can be removed because the if-then-else block will be empty.

s-perron:
When doing if-conversion, we do not currently move code out of the side
nodes.  The reason for this is that it can increase the number of
instructions that get executed because both side nods will have to be
executed now.

In this commit, we add code to move an instruction, and all of the
instructions it depends on, out of a side node and into the header of
the selection construct.  However to keep the cost down, we only do it
when the two values in the OpPhi node compute the same value.  This way
we have to move only one of the instructions and the other becomes
unused most of the time.  So no real extra cost.

Fixes #1526.

### DIFF
--- a/source/opt/if_conversion.cpp
+++ b/source/opt/if_conversion.cpp
@@ -14,12 +14,15 @@
 
 #include "if_conversion.h"
 
+#include "value_number_table.h"
+
 namespace spvtools {
 namespace opt {
 
 Pass::Status IfConversion::Process(ir::IRContext* c) {
   InitializeProcessing(c);
 
+  const ValueNumberTable& vn_table = *context()->GetValueNumberTable();
   bool modified = false;
   std::vector<ir::Instruction*> to_kill;
   for (auto& func : *get_module()) {
@@ -41,7 +44,8 @@ Pass::Status IfConversion::Process(ir::IRContext* c) {
           ir::IRContext::kAnalysisDefUse |
               ir::IRContext::kAnalysisInstrToBlockMapping);
       block.ForEachPhiInst([this, &builder, &modified, &common, &to_kill,
-                            dominators, &block](ir::Instruction* phi) {
+                            dominators, &block,
+                            &vn_table](ir::Instruction* phi) {
         // This phi is not compatible, but subsequent phis might be.
         if (!CheckType(phi->type_id())) return;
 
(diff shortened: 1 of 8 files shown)


s-perron:
There is a thorny issue.  The value number table needs the context, and now the ir context contains a pointer to the value number table.  This creates problems for the headers.  Let me know if there is a better way to handle it, or if I should just remove it.

source/opt/if_conversion.cpp
```diff
@@ -71,16 +75,46 @@ Pass::Status IfConversion::Process(ir::IRContext* c) {
           false_value = GetIncomingValue(phi, 0u);
         }
 
+        ir::BasicBlock* true_def_block = context()->get_instr_block(true_value);
+        ir::BasicBlock* false_def_block =
+            context()->get_instr_block(false_value);
```
dnovillo: Minor nit, these seem only used inside the if().  Move them there?

source/opt/if_conversion.cpp
```diff
@@ -79,8 +79,13 @@ Pass::Status IfConversion::Process(ir::IRContext* c) {
         ir::BasicBlock* false_def_block =
             context()->get_instr_block(false_value);
 
+        std::cerr << "Looking at OpPhi: " << phi->PrettyPrint(SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES) << "\n";
```
dnovillo: You left some debugging code here.

s-perron:
Not at this time.  if conversion creates new statements that do not get a value number.  The other one that uses it is redundancy elimination.  I do did not change it to use the one in the ir context.  However, that one could.  We can create a follow up issue for that.

s-perron:
I've opened https://github.com/KhronosGroup/SPIRV-Tools/issues/1543 to follow up on Alan's comments.

PR #1544, opened 2018-05-04
Allow ADCE to remove more instructions.

s-perron:
At this time, DCE will only remove an instruction if it is a combinator.
However, there are certain non-combinator instructions that can be
safely removed if their results are not used.  The derivative
instructions are on example.

We are also missing some instructions from the list of combinators
those are added as the same time.


source/opt/ir_context.cpp
```diff
@@ -339,6 +339,12 @@ void IRContext::AddCombinatorsForCapability(uint32_t capability) {
         SpvOpImageDrefGather,
         SpvOpImageRead,
         SpvOpImage,
+        SpvOpImageQueryFormat,
+        SpvOpImageQueryOrder,
+        SpvOpImageQuerySizeLod,
+        SpvOpImageQuerySize,
+        SpvOpImageQueryLevels,
+        SpvOpImageQuerySamples,
```
dnovillo: There was a TODO(dneto) about ImageQuery at the end.  I think that's safe to delete now.

source/opt/instruction.cpp
```diff
@@ -701,5 +701,27 @@ bool Instruction::IsScalarizable() const {
...
+    case SpvOpDPdx:
+    case SpvOpDPdy:
+    case SpvOpFwidth:
+    case SpvOpDPdxFine:
+    case SpvOpDPdyFine:
+    case SpvOpFwidthFine:
+    case SpvOpDPdxCoarse:
+    case SpvOpDPdyCoarse:
+    case SpvOpFwidthCoarse:
+    case SpvOpImageQueryLod:
```
dnovillo: Shouldn't it be more consistent to move this ImageQuery op to AddCombinatorsForCapability?

source/opt/ir_context.cpp
s-perron: Done.

source/opt/instruction.cpp
s-perron: No because it is not a combinator.  In the description of the opcode:

> This instruction is only valid in the Fragment Execution Model. **In addition, it consumes an implicit derivative that can be affected by code motion.**

dnovillo: Ah.  Thanks.

PR #1548, opened 2018-05-07
Add folding rule for load feeding extracts.

### LINKED ISSUE #1547: Replace load->extract with access chain and load.
If you run the following hlsl through DXC, you end up with a situation where you load all 32 elements of `a` just to extract one of them:

```
struct S {
  uint f;
};


cbuffer gBuffer { uint a[32]; };

RWStructuredBuffer<S> gRWSBuffer;

uint foo(uint p[32]) {
  return p[1];
}

[numthreads(1,1,1)]
void main() {
   gRWSBuffer[0].f = foo(a);
}
```

The resulting spir-v is:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpExecutionMode %main LocalSize 1 1 1
               OpSource HLSL 600
               OpName %type_gBuffer "type.gBuffer"
               OpMemberName %type_gBuffer 0 "a"
               OpName %gBuffer "gBuffer"
               OpName %S "S"
               OpMemberName %S 0 "f"
               OpName %type_RWStructuredBuffer_S "type.RWStructuredBuffer.S"
               OpName %gRWSBuffer "gRWSBuffer"
               OpName %main "main"
               OpDecorate %_arr_uint_uint_32 ArrayStride 16
               OpMemberDecorate %type_gBuffer 0 Offset 0
               OpDecorate %type_gBuffer Block
               OpMemberDecorate %S 0 Offset 0
               OpDecorate %_runtimearr_S ArrayStride 4
               OpMemberDecorate %type_RWStructuredBuffer_S 0 Offset 0
               OpDecorate %type_RWStructuredBuffer_S BufferBlock
               OpDecorate %gBuffer DescriptorSet 0
               OpDecorate %gBuffer Binding 0
               OpDecorate %gRWSBuffer DescriptorSet 0
               OpDecorate %gRWSBuffer Binding 1
       %uint = OpTypeInt 32 0
    %uint_32 = OpConstant %uint 32
%_arr_uint_uint_32 = OpTypeArray %uint %uint_32
%type_gBuffer = OpTypeStruct %_arr_uint_uint_32
%_ptr_Uniform_type_gBuffer = OpTypePointer Uniform %type_gBuffer
          %S = OpTypeStruct %uint
%_runtimearr_S = OpTypeRuntimeArray %S
%type_RWStructuredBuffer_S = OpTypeStruct %_runtimearr_S
%_ptr_Uniform_type_RWStructuredBuffer_S = OpTypePointer Uniform %type_RWStructuredBuffer_S
        %int = OpTypeInt 32 1
       %void = OpTypeVoid
         %15 = OpTypeFunction %void
      %int_0 = OpConstant %int 0
%_ptr_Uniform__arr_uint_uint_32 = OpTypePointer Uniform %_arr_uint_uint_32
     %uint_0 = OpConstant %uint 0
%_ptr_Uniform_uint = OpTypePointer Uniform %uint
    %gBuffer = OpVariable %_ptr_Uniform_type_gBuffer Uniform
 %gRWSBuffer = OpVariable %_ptr_Uniform_type_RWStructuredBuffer_S Uniform
       %main = OpFunction %void None %15
         %20 = OpLabel
         %21 = OpAccessChain %_ptr_Uniform__arr_uint_uint_32 %gBuffer %int_0
         %22 = OpLoad %_arr_uint_uint_32 %21    ; Load of 32-element array.
         %23 = OpCompositeExtract %uint %22 1
         %24 = OpAccessChain %_ptr_Uniform_uint %gRWSBuffer %int_0 %uint_0 %int_0
               OpStore %24 %23
               OpReturn
               OpFunctionEnd
```

We want to avoid the unnecessary loads.  This has two advantages:

1) Less memory is used and less data transferred when the driver does not do optimize it.
2) Some drivers do not accept loads of larger arrays of say 1,000 elements, so this has to be done in order to run those shaders.

s-perron:
We have already disabled common uniform elimination because it created
sequences of loads an entire uniform object, then we extract just a
single element.  This caused problems in some drivers, and is just
generally slow because it loads more memory than needed.

However, there are other way to get into this situation, so I've added
a folding rule to write these into a load of the specific element.

This fixes #1547.


greg-lunarg:
My concern here is that if we did have a situation where all the elements of a global object were being used, this just causes a needless increase in code size and possibly a reduction in performance.

For example, what if the index in foo (in #1547) was "i" instead of "1"? Now this looks more like a de-optimization instead of an optimization, by doing a separate memory operation for each element.

It feels like #1547 should be fixed with array copy propagation, by running it very early, like right after inlining.

s-perron:
@greg-lunarg The problem with array copy propagation is that is relies of all of the other optimizations to make the copy of one array into another obvious.  If we do it earlier, it will just miss the opportunity.

If the element index was "i" then local-access-chain-convert would not have turn what was originally an OpAccessChain and OpLoad into an OpLoad and OpCompositeExtract.

Also, these large loads do not come from the source code.  They are an artifact of local-access-chain-convert.  We also do not know the capabilities of the hardware.  If the user writes these loads as separate loads (except for when they do a wholesale array copy), should we be the ones to merge the loads (as local-access-chain-convert does) or leave them to the driver to determine what is best for the specific hardware?  This is another example of the same problem in https://github.com/KhronosGroup/SPIRV-Tools/issues/946.

Do you have any data was can use to guide the heuristics?  In the limited code that I see, almost all cases are arrays with 1024+ elements and only a handful get used.

Note this does not happen for structured buffers they do not get an entire local copy, which is part of the cause.  This is also not a problem for any type that is declared with OpTypeImage or OpTypeRuntimeArray.

s-perron:
After thinking about @greg-lunarg comments a little more, and discussing with @dnovillo, another solution is to not do local-access-chain convert on arrays, and then trust array-copy-propagation as Greg had suggested.

dnovillo:
@s-perron are you going to try phrasing this in terms of not doing local-access-chain conversion?  I don't see an issue with the approach you have here.  But, if this is being hampered by l-a-c, I would rather taking that out first  (incidentally, I'm not sure I completely grasp the idea behind l-a-c conversion, but that's another discussion).

greg-lunarg:
I am fairly certain that LAC only acts on local/function scope variables, so I don't think that it is at fault here.  Unless someone changed it and I missed it. I am guessing that dxc is generating the Load/Extract pattern. Can someone confirm or contradict?

s-perron:
No it is LAC.  I only showed the final code after we have a number of optimizations applied.  I can attach the unoptimized spir-v that dxc generates if you want.

What happens is that there is a local copy (because of the parameter being passed by value) of the cbuffer.  The instruction sequence is: 

1. A load of the cbuffer.
1. A store to the local copy.
1. An OpAccessChain on the local copy.
1. An OpLoad of the single element from the local copy.

The first time we run LAC it sees the load as an OpAccessChain and OpLoad on a function scope symbol.  That gets converted to an OpCompositeExtract.  Then the ssa rewriter propagates the load of the cbuffer into the OpCompositeExtract.

greg-lunarg:
OK. Thanks.

I agree a solution that utilizes (a possibly improved) early call to array copy prop is still worth looking at.

I am fine with taking LAC out of the main recipes. It has outlived its usefulness there. It was originally meant to keep the number of local symbols down (rather than explode them ala scalar replacement) and to eliminate the need to do do analysis on local access chains by unifying everything under the insert/extract regimen. But since we have bit the scalar replacement bullet its usefulness is less obvious.

s-perron:
@greg-lunarg Do you want LAC to continue working as is?  I ask because I don't want to remove it completely.  It still want it to modifiy accesses to vectors an matricies.  I was thinking of limiting LAC; however, I understand that glslang does not use our optimization sets.  They probably need to the same behaviour.  Let me know what you think.

s-perron:
@greg-lunarg @dnovillo I've updated the PR.  I've made the changes a pass, and added some heuristics.  Let me know if I should change anything.

greg-lunarg:
Why does the pass not do a replacement if all elements are used?

s-perron:
That is the case you were worried about.  We could do a load of the whole struct in one shot, which may be faster it the proper hardware is present.  However, if we only use part of the load, then less memory bandwidth may be more beneficial.  I have a threshold that can be tuned in the future.

greg-lunarg:
I have changed my mind.

Since the frontends generally generate an access-chain/load sequence for these anyway, this will likely do little harm.

Another reason is that I think we can use it to facilitate array copy prop. Look at the code generated by glslangValidator for the very similar shader:

struct S {
  uint f;
};

cbuffer gBuffer { uint a[32]; uint i; };

RWStructuredBuffer<S> gRWSBuffer;

uint foo(uint p[32], uint i) {
  return p[i];
}

[numthreads(1,1,1)]
void main() {
   gRWSBuffer[0].f = foo(a, i);
}

glslangValidator generates the load/extract sequence to do these copys. We could use this new pass to put such code into a "canonical" form (access-chain/load). I haven't tested it yet, but I think that would then allow array copy prop to clean this code up.

greg-lunarg:
OK, thanks. I am happy that this has been split out of Simplification. I no longer have a problem with this going forward.

include/spirv-tools/optimizer.hpp
```diff
@@ -578,6 +578,12 @@ Optimizer::PassToken CreateCopyPropagateArraysPass();
 // a pass of ADCE will be able to remove.
 Optimizer::PassToken CreateVectorDCEPass();
 
+// Create a pass to reduce the size of loads.
+// This pass looks for loads of structures where only a few of it members are
```
dnovillo: s/it/its/

source/opt/copy_prop_arrays.cpp
```diff
@@ -269,7 +269,7 @@ CopyPropagateArrays::BuildMemoryObjectFromExtract(
     // Convert the indices in the extract instruction to a series of ids that
     // can be used by the |OpAccessChain| instruction.
     for (uint32_t i = 1; i < extract_inst->NumInOperands(); ++i) {
-      uint32_t index = extract_inst->GetSingleWordInOperand(1);
+      uint32_t index = extract_inst->GetSingleWordInOperand(i);
```
dnovillo: oh, subtle :)  Do you have a test case for this or did you catch it by visual inspection?

source/opt/reduce_load_size.cpp
```diff
@@ -0,0 +1,176 @@
...
+      op_inst, [&elements_used, &all_elements_used](ir::Instruction* use) {
+        if (use->opcode() != SpvOpCompositeExtract) {
+          all_elements_used = true;
+        }
+        elements_used.insert(use->GetSingleWordInOperand(1));
+      });
+
+  bool should_replace = false;
+  if (all_elements_used) {
+    should_replace = false;
```
dnovillo: Should we have a threshold different than 100% at which to do the replacement?  On big structures, if more than half the elements are used, maybe the replacement is not worth doing?  Not necessary to do this in this PR.  I'm not even sure if it's worth implementing at all.

source/opt/copy_prop_arrays.cpp
s-perron: We did not have a test case, but I added one.  I was looking at that code because I wanted to do something similar in the new pass, and noticed the error.

source/opt/reduce_load_size.cpp
s-perron: In the other case, where not every element is used, we check if the percent used is less than a threshold.  That value is currently 90%.  We can adjust that as we see the need.

Issue #1550, opened 2018-05-08
Do dead store elimination on variable in the Workgroup storage class.


s-perron:
If there is a variable in the Workgroup storage class that is stored to, but never loaded from, then we can remove the stores.  We do not do this in spirv-opt.

PR #1551, opened 2018-05-08
Remove dead Workgroup variables in ADCE.

s-perron:
If there is a shader with a variable in the workgroup storage class that
is stored to, but not loadeds, then we know nothing will read those
loads.  It should be safe to remove them.

This is implemented in ADCE by treating workgroup variables the same
way that private variables are treated.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1550.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -91,8 +91,15 @@ bool AggressiveDCEPass::IsVarOfStorage(uint32_t varId, uint32_t storageClass) {
 }
 
 bool AggressiveDCEPass::IsLocalVar(uint32_t varId) {
-  return IsVarOfStorage(varId, SpvStorageClassFunction) ||
-         (IsVarOfStorage(varId, SpvStorageClassPrivate) && private_like_local_);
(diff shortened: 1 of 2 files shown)


source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -327,7 +334,8 @@ bool AggressiveDCEPass::AggressiveDCE(ir::Function* func) {
           // Mark stores as live if their variable is not function scope
           // and is not private scope. Remember private stores for possible
           // later inclusion
-          if (IsVarOfStorage(varId, SpvStorageClassPrivate))
+          if (IsVarOfStorage(varId, SpvStorageClassPrivate) ||
+              IsVarOfStorage(varId, SpvStorageClassWorkgroup))
```
dnovillo: Why not call IsLocalVar() here?

dnovillo: Ah, ok.  Could you add a small comment here?

s-perron: Done.

PR #1555, opened 2018-05-10
Adjust tests according to grammar change

antiagainst:
* ConstOffsets now requires ImageGatherExtended
* Int8 does not require Kernel anymore

See https://github.com/KhronosGroup/SPIRV-Headers/commit/3a9bf0b12a5b1a0a4996b0066f1451345c293546


dneto0:
Approved.  

Failures were due to a SPIRV-Headers update that changed capability dependencies.
KhronosGroup/SPIRV-Headers@3a9bf0b

This caused failures in SPIRV-Tools validation tests, since they check some of those dependencies.

All future bot builds will fail until this patch goes in, so I'll push this to master now.

Issue #937, opened 2017-11-01
Add checks for storage layout decorations on Block and BufferBlock decorated structs


ianromanick:
Both Vulkan and GL_ARB_spirv has a requirement like:

    7.6.2.spv SPIR-V Uniform Offsets and Strides

    The SPIR-V decorations *GLSLShared* or *GLSLPacked* must not be used. A
    variable in the *Uniform* Storage Class decorated as a *Block* must be
    explicitly laid out using the *Offset*, *ArrayStride*, and *MatrixStride*
    decorations. If the variable is decorated as a *BufferBlock*, its offsets
    and strides must not contradict std430 alignment and minimum offset
    requirements.  Otherwise, its offsets and strides must not contradict
    std140 alignment and minimum offset requirements.

However, the validator doesn't seem to perform any checks for Offset, ArrayStride, etc. decorations when a structure is decorated with Block or BufferBlock.

dneto0:
#1568 was merged as part of #1639.  Remaining work has been split off into:

#1637 #1638 #1606

PR #1568, opened 2018-05-17
Add validation for structs decorated as Block or BufferBlock.

asuonpaa:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/937

The following checks are performed for structs that are decorated as Block or BufferBlock:

- No GLSLShared or GLSLPacked decoration should be used by any members.
- All members need to have Offset decoration. Arrays should have ArrayStride and matrices MatrixStride decorations.
- Blocks need to follow std140 rules.
- BufferBlocks need to follow std430 rules.


antiagainst:
Also, it would be nice if we can have an option to turn off checks for the offsets. Existing Vulkan layout rules are incompatible with DirectX, which causes quite some pain for porting games from DX to VK. So we end up having a [`-fvk-use-dx-layout`](https://github.com/Microsoft/DirectXShaderCompiler/pull/1198) to let DXC to emit offsets using DX layout rules, which devs find very helpful. We run spirv-val at the end of DXC's SPIR-V CodeGen. Such an option to turn off the checks for DX layouts will be very helpful.

An example of providing validator options:

https://github.com/KhronosGroup/SPIRV-Tools/blob/f1f7cc870e13e1ad1eb2f2e757c21aec4276c707/include/spirv-tools/libspirv.h#L466-L473

asuonpaa:
It's now checking for standard uniform/storage buffer layout. However I'm unsure what this rule means:

> The Offset decoration of a member must not place it between the end of a structure or an array and the next multiple of the base alignment of that structure or array.

So until I figure out what that means, this particular rule is unchecked.

I'll add the run time switch in a separate commit.

asuonpaa:
I think I figured out the remaining layout rule. I'll do a separate commit for that.

asuonpaa:
I suppose the checking should be only done for variables with Uniform or PushConstant storage class. Right now all structs are checked. I'm currently working on the changes for this.

source/validate_decorations.cpp
```diff
@@ -59,6 +60,282 @@ bool hasImportLinkageAttribute(uint32_t id, ValidationState_t& vstate) {
...
+  const auto inst = vstate.FindDef(member_id);
+  const auto words = inst->words();
+  uint32_t baseAlignment = 0;
+  switch (inst->opcode()) {
+    case SpvOpTypeInt:
+    case SpvOpTypeFloat:
+      baseAlignment = words[2] / 8;
+      break;
+    case SpvOpTypeVector:
+      baseAlignment = getBaseAlignment(words[2], roundUp, vstate);
```
alan-baker: This should be multiplied by the number of elements in the vector rounded up to a power of two (e.g. vec3 is aligned to vec4).

source/validate_decorations.cpp
```diff
@@ -59,6 +60,282 @@ bool hasImportLinkageAttribute(uint32_t id, ValidationState_t& vstate) {
...
+    case SpvOpTypeFloat:
+      return baseAlignment;
+    case SpvOpTypeVector: {
+      const auto componentId = words[2];
+      const auto numComponents = words[3];
+      const auto componentSize = getSize(componentId, roundUp, vstate);
+      return componentSize * numComponents;
+    }
+    case SpvOpTypeArray:
+      return (vstate.FindDef(words[3])->words()[3] - 1) * baseAlignment +
```
alan-baker: This seems wrong for an array sized by a OpSpecConstant. I don't think you can return a size in that case.

source/validate_decorations.cpp
```diff
@@ -59,6 +60,282 @@ bool hasImportLinkageAttribute(uint32_t id, ValidationState_t& vstate) {
...
+// - It is a vector with total size greater than 16 bytes and has its Offset
+// decorations placing its first byte at a non-integer multiple of 16.
+bool hasImproperStraddle(uint32_t id, uint32_t offset,
+                         ValidationState_t& vstate) {
+  const auto inst = vstate.FindDef(id);
+  const auto words = inst->words();
+  const auto componentId = words[2];
+  const auto numComponents = words[3];
+  const auto componentSize = vstate.FindDef(componentId)->words()[2] / 8;
+  const auto size = componentSize * numComponents;
```
alan-baker: Why not getSize(id)?

source/validate_decorations.cpp
asuonpaa: Fixed. I did have it like that in one of the previous versions. Somehow missed it here.

source/validate_decorations.cpp
asuonpaa: Now it's returning zero just like with run time arrays.

source/validate_decorations.cpp
asuonpaa: GetSize function was added later and I didn't realize it could be used there. Now it's being used like you suggested.

source/validate_decorations.cpp
```diff
@@ -160,9 +166,13 @@ uint32_t getSize(uint32_t member_id, bool roundUp, ValidationState_t& vstate) {
       const auto componentSize = getSize(componentId, roundUp, vstate);
       return componentSize * numComponents;
     }
-    case SpvOpTypeArray:
-      return (vstate.FindDef(words[3])->words()[3] - 1) * baseAlignment +
+    case SpvOpTypeArray: {
+      const auto sizeInst = vstate.FindDef(words[3]);
+      if (SpvOpSpecConstant == sizeInst->opcode()) return 0;
```
alan-baker: Sorry I wasn't clearer when I said OpSpecConstant. I meant the whole family. Use `spvOpcodeIsSpecConstant()`.

antiagainst:
@ehsannas: could you cross-check this pull request with the current DXC to make sure both are correctly implemented before merging? Thanks!

dneto0:
Sorry I haven't been on this.  Note that Uniform storage class with BufferBlock decoration is the same as StorageBuffer with Block decoration.  Hey @alan-baker if this PR doesn't cover the latter then please file an issue to extend it to that case.

alan-baker:
Filed #1606

ehsannas:
I'll look into whether this plays nice with DXC.

source/validate_decorations.cpp
```diff
@@ -180,6 +464,63 @@ spv_result_t CheckDecorationsOfEntryPoints(ValidationState_t& vstate) {
...
+                   << " must follow standard uniform buffer layout rules.";
+          } else if (isBufferBlock && !checkLayout(id, false, vstate)) {
+            return vstate.diag(SPV_ERROR_INVALID_ID)
+                   << "Structure id " << id << " decorated as BufferBlock"
+                   << " must follow standard storage buffer layout rules.";
+          }
+        }
+      }
+    }
+    //	  }
```
ehsannas: remove this line?

ehsannas:
This change introduces ~30% failure rate in DXC's SPIR-V CodeGen tests.

ehsannas:
as @antiagainst mentioned, would it be possible to add an option to the validator to turn off these checks for DX layouts? Otherwise we won't be able to use the SPIRV-Tools TOT due to large number of failures it'll cause.

alan-baker:
There was an option added "Relaxed block layout". Did you run the tests with that option?

asuonpaa:
Looks like it's missing code to expose the options to command line. I'll add the necessary changes.

ehsannas:
Thanks @asuonpaa for adding the option! This options does fix the SPIRV CodeGen tests in DXC.
However, the change as a whole still causes failures when tested on some internal workloads (tested on Windows).

The appveyor Windows buildbots for the past two commits to this PR have also failed:
https://ci.appveyor.com/project/Khronoswebmaster/spirv-tools/build/3384
https://ci.appveyor.com/project/Khronoswebmaster/spirv-tools/build/3385

I just gave it a try on my Windows machine. I took the TOT of SPIRV-Tools and cherry-picked this PR on top of it. Compilation was successful, but the following unit tests failed (run using `ctest`):

```
The following tests FAILED:
         35 - spirv-tools-test_pass_inline (Failed)
         76 - spirv-tools-test_pass_if_conversion (Failed)
        116 - spirv-tools-test_fusion_legal (Failed)
```

For the internal workload on Windows, I am seeing an assertion being hit
`assert(offset != 0xffffffff);`
which is line 192 of validate_decorations.cpp in the newly added code.

source/validate_decorations.cpp
```diff
@@ -59,6 +60,289 @@ bool hasImportLinkageAttribute(uint32_t id, ValidationState_t& vstate) {
...
+      const auto& lastMember = members.back();
+      uint32_t offset = 0xffffffff;
+      // Find the offset of the last element and add the size.
+      for (auto& decoration : vstate.id_decorations(member_id)) {
+        if (SpvDecorationOffset == decoration.dec_type() &&
+            decoration.struct_member_index() == (int)lastIdx) {
+          offset = decoration.params()[0];
+        }
+      }
+      assert(offset != 0xffffffff);
```
ehsannas: I see this assertion being hit on some workloads.

asuonpaa:
@ehsannas, I tried to reproduce the assertions by cloning the TOT and patching the pull request. I wasn't able to run ctest on Visual Studio build, but instead ran the individual tests you reported directly from VS. I saw no assertions. Then I built the same sources under linux subsystem and there ctest was also working. Everything was passing without errors there also.

Any ideas how to get this reproduced on my side? Maybe just share a zip of your code base?

ehsannas:
I am able to reproduce the ctest failure (using ninja and ctest).
More details:
```
[ RUN      ] InlineTest.OpVariableWithInit
c:\spirv-tools\test\opt\pass_fixture.h(179): error: Expected equality of these values:
  effcee::Result::Status::Ok
    Which is: 4-byte object <00-00 00-00>
  match_result.status()
    Which is: 4-byte object <01-00 00-00>
<stdin>:39:7: error: CHECK-NOT: string occurred!
%43 = OpLoad %float %41
      ^
<stdin>:6:14: note: CHECK-NOT: pattern specified here
; CHECK-NOT: OpLoad %float [[var]]
             ^
<stdin>:39:7: note: with variable "var" equal to "%41"
%43 = OpLoad %float %41
[  FAILED  ] InlineTest.OpVariableWithInit (111 ms)
```

asuonpaa:
@ehsannas : I'm still unable to reproduce. Here's what I did:

- cloned your fork
- checked out tmp2 branch
- launched visual studio command prompt
- built successfully using ninja
- ran ctest

All tests were passing for me. This happens on both NVidia and Intel GPUs on my Surface Book.

ehsannas:
@asuonpaa , It looks like the `ctest` failures occur even on the 'master' branch. So I'll open a separate issue for that. Sorry for the confusion.

But, I still do see crashes due to `assert(offset != 0xffffffff);`

I looked at the SPIRV code that causes the crash, and it contains some empty structs. Note that empty structures are valid in SPIR-V, and they are used more often than one might expect!

source/validate_decorations.cpp
ehsannas: Must take empty structs into account

ehsannas:
Thanks, @asuonpaa !

dneto0:
Hi @asuonpaa please rebase against master.  A change has gone into master that avoids the AppVeyor failure. I'd like to see a clean build bot run for this MR.

dneto0:
Sorry, now there are conflicts.  Please rebase one more time.  :-)

source/validate_decorations.cpp
```diff
@@ -225,6 +518,62 @@ spv_result_t CheckDescriptorSetArrayOfArrays(ValidationState_t& vstate) {
...
+                   << " must be explicitly laid out with Offset decorations.";
+          } else if (hasDecoration(id, SpvDecorationGLSLShared, vstate)) {
+            return vstate.diag(SPV_ERROR_INVALID_ID)
+                   << "Structure id " << id << " decorated as " << dec_str
+                   << " must not use GLSLShared decoration.";
+          } else if (hasDecoration(id, SpvDecorationGLSLPacked, vstate)) {
+            return vstate.diag(SPV_ERROR_INVALID_ID)
+                   << "Structure id " << id << " decorated as " << dec_str
+                   << " must not use GLSLPacked decoration.";
+          } else if (!checkForRequiredDecoration(id, SpvDecorationArrayStride,
```
dneto0: This is going to miss a case.

When the member is an array of matrices, or an array of an array of matrices, etc., then the MatrixStride is required on the struct member that is the outermost array type.

Example vertex shader in GLSL:

```
#version 450

layout (set=1, binding=2) uniform S { mat4x4 arr[2][3]; } B;

void main() {
  vec4 v = B.arr[1][2][1];
}
```
Here's the SPIR-V:
```
; SPIR-V
; Version: 1.0
; Generator: Google Shaderc over Glslang; 7
; Bound: 26
; Schema: 0
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpSource GLSL 450
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %main "main"
               OpName %v "v"
               OpName %S "S"
               OpMemberName %S 0 "arr"
               OpName %B "B"
               OpDecorate %_arr_mat4v4float_uint_3 ArrayStride 64
               OpDecorate %_arr__arr_mat4v4float_uint_3_uint_2 ArrayStride 192
               OpMemberDecorate %S 0 ColMajor
               OpMemberDecorate %S 0 Offset 0
               OpMemberDecorate %S 0 MatrixStride 16 ;;;;; This is the surprising part
               OpDecorate %S Block
               OpDecorate %B DescriptorSet 1
               OpDecorate %B Binding 2
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
%_ptr_Function_v4float = OpTypePointer Function %v4float
%mat4v4float = OpTypeMatrix %v4float 4
       %uint = OpTypeInt 32 0
     %uint_3 = OpConstant %uint 3
%_arr_mat4v4float_uint_3 = OpTypeArray %mat4v4float %uint_3
     %uint_2 = OpConstant %uint 2
%_arr__arr_mat4v4float_uint_3_uint_2 = OpTypeArray %_arr_mat4v4float_uint_3 %uint_2
          %S = OpTypeStruct %_arr__arr_mat4v4float_uint_3_uint_2
%_ptr_Uniform_S = OpTypePointer Uniform %S
          %B = OpVariable %_ptr_Uniform_S Uniform
        %int = OpTypeInt 32 1
      %int_0 = OpConstant %int 0
      %int_1 = OpConstant %int 1
      %int_2 = OpConstant %int 2
          %5 = OpLabel
          %v = OpVariable %_ptr_Function_v4float Function
         %24 = OpAccessChain %_ptr_Uniform_v4float %B %int_0 %int_1 %int_2 %int_1
         %25 = OpLoad %v4float %24
               OpStore %v %25
               OpReturn
               OpFunctionEnd
```
Note where the MatrixStride is applied: It's on member 0 of the struct, not on the matrix type.

I have filed #1637 to track this issue.

dneto0:
I'm getting compilation problems with floor not being declared.
I'll make a followup CL to fix that one problem in the way I suggested, and have someone else review.

source/validate_decorations.cpp
```diff
@@ -59,6 +60,298 @@ bool hasImportLinkageAttribute(uint32_t id, ValidationState_t& vstate) {
...
+  bool nestedStructsMissingOffset = false;
+  for (auto id : getStructMembers(struct_id, SpvOpTypeStruct, vstate)) {
+    if (isMissingOffsetInStruct(id, vstate)) {
+      nestedStructsMissingOffset = true;
+      break;
+    }
+  }
+  return nestedStructsMissingOffset ||
+         !std::all_of(hasOffset.begin(), hasOffset.end(),
+                      [](const bool b) { return b; });
```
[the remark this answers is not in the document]
dneto0: Meh.  I see how you process the indices in any order.  So the code isn't so easily rewritten with just some counters.  So never mind this request.

source/validate_decorations.cpp
```diff
@@ -59,6 +60,298 @@ bool hasImportLinkageAttribute(uint32_t id, ValidationState_t& vstate) {
...
+      return false;
+    if (offset >= padStart && offset + lastByte <= padEnd) return false;
+    // Check improper straddle of vectors.
+    if (SpvOpTypeVector == opcode && hasImproperStraddle(id, offset, vstate))
+      return false;
+    // Check struct members recursively.
+    if (SpvOpTypeStruct == opcode && !checkLayout(id, isBlock, vstate))
+      return false;
+    // Check matrix stride.
+    if (SpvOpTypeMatrix == opcode) {
```
dneto0, line 276: This is looking for a MatrixStride decoration on the *matrix type*. 
But the MatrixStride decoration decoration is on the struct member.  So this will not trigger.

I think this is the root cause of something I'm seeing now, i.e. that MatrixStride itself is not being validated.

Issue #1570, opened 2018-05-18
Replace insert-extract elimination by the simplification pass.


s-perron:
The insert-extract elimination pass is really a special purpose instruction folding pass.  This is the same thing that is implemented in the simplification pass, so we should be able to replace the former by the latter.

However, the simplification pass is missing some rules that are implemented in the insert-extract elimination pass.

For this to be considered done we will need to

- [x] Have folding rules that cover everything in the insert-extract elimination pass.
- [x] Replace the instances of insert-extract-elim in the standard optimizations sets by the simplification pass.
- [ ] Change the option `--eliminate-insert-extract` to call enable the simplification pass instead with documentation changes to deprecate the option.
- [ ] Delete all code that is not unused because of the first three changes.

greg-lunarg:
I can certainly imagine value in using the simplification infrastructure to do extract elimination, but I am slightly concerned with conflating all the other simplifications with it (x + 0.0 => x, etc). My concern is mostly around compile time. My instincts suggest we might want to do have these done separately as it feels like they would be useful at different points in an optimization recipe. Do we have data on simplification time together and separately?

s-perron:
About the conflation, what do you think should be kept separate?  When I think of the simplifications that are done, the only ones that I think we might want independent control of are the floating point folding rules like the one you mentioned above.  However, that can be turned off with the "NoContraction" decoration.  This is really a  FE issue.  If the user tell the FE that generates the spir-v that they don't want floating point operations folded, then they should add the decoration.

I'll try to collect some compile time numbers.  I expect the compile time will be higher, but the question is whether or not the compile time is worth it.

Do you have customer who actually do not want the extra folding, or is it just hypothetical?  For now the extra simplification is mostly just fold arithmetic (fix-point and floating-point).  However, I have see the OpPhi and OpSelect folding rules make a big difference.

greg-lunarg:
I would seperate the extract elimination and any other "vector" simplification (for example, OpVectorShuffle) from everything else.

The concern is strictly compile time. Someone may wish to create a recipe which does both kinds of simplfication, but at different frequencies and different points in the recipe.

s-perron:
I did some compile time test.  I compared three different optimization orders.  "current" is what we currently have for `--legalize-hlsl -O`.  Proposed is what I have implemented in #1571.  "other" is the same as current except I reordered the simplification pass and the insert-extract-elim pass in the legalization set.  I have attached the config file.  
[...]

PR #1571, opened 2018-05-18
Fold FMix feeding Extract, and use the simplification pass.

s-perron:
We add a new rule to the folding rules to fold an FMix feeding an
extract when the alpha value for the element being extracted is either
0 or 1.  In those case, we can simple extract from one of the operands
to the FMix.

With that change the simplification pass completely subsumes the
insert-extract elimination pass.  So we remove the insert-extract
elimination passes and replce them with calls to the simplification
pass.

In a follow up PR, we should delete the insert-extract elimination pass.

Contributes to https://github.com/KhronosGroup/SPIRV-Tools/issues/1570.


s-perron:
@greg-lunarg Could you let me know if I am missing anything in the simplification pass?

s-perron:
Yes I realized that after opening the PR I forgot the test.  Working on it now.  Thanks.

PR #1573, opened 2018-05-22
Validate presence of OpMemoryModel.

### LINKED ISSUE #1207: validator does not always insist on an OpMemoryModel instruction
Example input:
```OpCapability Linkage``` passes validation when it shouldn't.

### PR DESCRIPTION
**Pull request by dj2**
According to the SPIR-V Spec, section 2.4 Logical Layout of a Module there
should be a single required OpMemoryModel instruction provided. This CL adds
validation that OpMemoryModel is provided to the SPIR-V validator.

Fixes #1207

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -158,6 +158,7 @@ ValidationState_t::ValidationState_t(const spv_const_context ctx,
       local_vars_(),
       struct_nesting_depth_(),
       grammar_(ctx),
+      has_memory_model_specified_(false),
       addressing_model_(SpvAddressingModelLogical),
       memory_model_(SpvMemoryModelSimple),
       in_function_(false) {
@@ -363,6 +364,7 @@ bool ValidationState_t::HasAnyOfExtensions(
 }
 
 void ValidationState_t::set_addressing_model(SpvAddressingModel am) {
+  has_memory_model_specified_ = true;
   addressing_model_ = am;
 }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: source/val/validation_state.cpp [hunk 1]
```diff
@@ -158,6 +158,7 @@ ValidationState_t::ValidationState_t(const spv_const_context ctx,
       local_vars_(),
       struct_nesting_depth_(),
       grammar_(ctx),
+      has_memory_model_specified_(false),
```
alan-baker: I think it would be better to reuse `memory_model_`, but initialize it to SpvMemoryModelMax.

source/val/validation_state.h
```diff
@@ -262,6 +262,9 @@ class ValidationState_t {
   /// Sets the addressing model of this module (logical/physical).
   void set_addressing_model(SpvAddressingModel am);
 
+  /// Returns true if the OpMemoryModel was found.
+  bool has_memory_model_specified() const { return has_memory_model_specified_; }
```
alan-baker: With the above change, make this return `memory_model_ != SpvMemoryModelMax`.

source/val/validation_state.cpp
```diff
@@ -363,6 +364,7 @@ bool ValidationState_t::HasAnyOfExtensions(
 }
 
 void ValidationState_t::set_addressing_model(SpvAddressingModel am) {
+  has_memory_model_specified_ = true;
```
alan-baker: Remove

source/val/validation_state.cpp
```diff
@@ -371,6 +373,7 @@ SpvAddressingModel ValidationState_t::addressing_model() const {
 }
 
 void ValidationState_t::set_memory_model(SpvMemoryModel mm) {
+  has_memory_model_specified_ = true;
```
alan-baker: Remove

PR #1575, opened 2018-05-22
Allow registering external passes without a token

zeux:
Currently it's impossible for external code to register a pass because
the only source file that can create pass tokens is optimizer.cpp. This
makes it hard to add passes that can't be upstreamed since you can't run
them from the usual pass sequence without reimplementing Optimizer.

This change adds an overload for RegisterPass that takes unique_ptr to
opt::Pass; if external code implements opt::Pass it can register a
custom pass without having to add it to SPIRV-Tools source code.


s-perron:
Hello,

Can you be more specific about what you want to do?  Are you writing your own pass?

If not, then why can't you include `optimizer.hpp` and call the `Create*Pass` functions to create the PassToken for you?

zeux:
Yeah - I an writing my own pass. Unless I am missing something, your suggestion requires the pass declaration to be visible in optimizer.cpp which requires me to maintain a fork of SPIRV-Tools or to upstream my pass. The proposed solution would make it possible to maintain out-of-tree passes without changing SPIRV-Tools.

s-perron:
I'll merge if the windows build pass.  I can run clang format when I rebase and squash.

PR #1576, opened 2018-05-23
Validate that only a single OpMemoryModel is provided.

### LINKED ISSUE #1574: Validator allows multiple OpMemoryModel entries.
According to the spec there should be a single defined OpMemoryModel (Section 2.4 Logical Layout of a Module). Currently the validator will accept the following as valid:

    OpCapability Linkage
    OpCapability Shader
    OpMemoryModel Logical Simple
    OpMemoryModel Logical Simple

The second OpMemoryModel should be dis-allowed.

dj2:
This CL adds validation that only a single OpMemoryModel is provided in the
SPIR-V binary.
    
Fixes #1574


alan-baker:
Rebased, squashed and committed as 0a14a1f74840ebd172e4128a8789f44190d1e4e7

PR #1580, opened 2018-05-25
Allow duplicate pointer types

### LINKED ISSUE #1577: spirv-val reports "Duplicate non-aggregate type declarations..." error for pointer types
According to the 1.3 specification section 2.8:
`
[...] Pointer types are also allowed to have multiple <id>s for the same opcode and operands, to allow for differing decorations (e.g., Volatile) or different decoration values (e.g., different Array Stride values for the ArrayStride). [...] Non-aggregate non-pointer types are different: It is invalid to declare multiple type <id>s for the same scalar, vector, or matrix type. That is, non-aggregate non-pointer type declarations must all have different opcodes or operands. [...]
`

Which means spirv-val shouldn't check for duplicate pointer types.

alan-baker:
Fixes #1577

* Remove validation requiring unique pointer types unless variable
pointers extension enabled
* Modified scalar replacement to always look for an undecorated pointer

### DIFF
--- a/source/opt/scalar_replacement_pass.cpp
+++ b/source/opt/scalar_replacement_pass.cpp
@@ -377,11 +377,8 @@ uint32_t ScalarReplacementPass::GetOrCreatePointerType(uint32_t id) {
     if (global.opcode() == SpvOpTypePointer &&
         global.GetSingleWordInOperand(0u) == SpvStorageClassFunction &&
         global.GetSingleWordInOperand(1u) == id) {
-      if (!context()->get_feature_mgr()->HasExtension(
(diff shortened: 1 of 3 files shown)


s-perron:
The code looks okay, but it is not clear why you need the change in scalar replacement.  Could you give an explanation?

alan-baker:
Scalar replacement wants an undecorated pointer type. Previously it assumed you couldn't have duplicate pointer types unless the variable pointers extension was present. The spec language has been updated to allow duplicate pointers so I changed scalar replacement to always require an undecorated pointer.

alan-baker:
Rebased and submitted as badcf73d00c2a83e5272ab85cfa0c31e6564b1c5

PR #1583, opened 2018-05-28
WIP: Workaround undefined depth formats

AWoloszyn:
I have absolutely no intention on trying to check this in as-is. It was a first pass, I am just trying to get feedback on what are the "RIGHT" facilities from spirv-opt to use.

I will add specific questions

Problems that I already intend to fix, but wanted to get early feedback
1) Names. Nothing is consistent.
2) Nesting is a bit silly right now, obviously needs fixing.

### DIFF
--- a/source/opt/types.cpp
+++ b/source/opt/types.cpp
@@ -328,9 +328,23 @@ Image::Image(Type* type, SpvDim dimen, uint32_t d, bool array, bool multisample,
       ms_(multisample),
       sampled_(sampling),
       format_(f),
-      access_qualifier_(qualifier) {
+      access_qualifier_(qualifier),
+      has_access_qualifier_(true) {
   // TODO(antiagainst): check sampled_type
 }
+
+Image::Image(Type* type, SpvDim dimen, uint32_t d, bool array, bool multisample,
+             uint32_t sampling, SpvImageFormat f)
+    : Type(kImage),
+      sampled_type_(type),
+      dim_(dimen),
+      depth_(d),
+      arrayed_(array),
+      ms_(multisample),
+      sampled_(sampling),
+      format_(f),
+      access_qualifier_(SpvAccessQualifierReadOnly),
+      has_access_qualifier_(false) {}
 
 bool Image::IsSameImpl(const Type* that, IsSameCache* seen) const {
   const Image* it = that->AsImage();
@@ -339,14 +353,18 @@ bool Image::IsSameImpl(const Type* that, IsSameCache* seen) const {
          ms_ == it->ms_ && sampled_ == it->sampled_ && format_ == it->format_ &&
          access_qualifier_ == it->access_qualifier_ &&
          sampled_type_->IsSameImpl(it->sampled_type_, seen) &&
+         has_access_qualifier_ == it->has_access_qualifier_ &&
          HasSameDecorations(that);
 }
 
 std::string Image::str() const {
   std::ostringstream oss;
   oss << "image(" << sampled_type_->str() << ", " << dim_ << ", " << depth_
-      << ", " << arrayed_ << ", " << ms_ << ", " << sampled_ << ", " << format_
-      << ", " << access_qualifier_ << ")";
+      << ", " << arrayed_ << ", " << ms_ << ", " << sampled_ << ", " << format_;
+  if (has_access_qualifier_) {
+    oss << ", " << access_qualifier_;
+  }
+  oss << ")";
   return oss.str();
 }
 
@@ -359,7 +377,9 @@ void Image::GetExtraHashWords(std::vector<uint32_t>* words,
   words->push_back(ms_);
   words->push_back(sampled_);
   words->push_back(format_);
-  words->push_back(access_qualifier_);
+  if (has_access_qualifier_) {
+    words->push_back(access_qualifier_);
+  }
 }
 
 bool SampledImage::IsSameImpl(const Type* that, IsSameCache* seen) const {
--- a/source/opt/workaround_image_operands.cpp
+++ b/source/opt/workaround_image_operands.cpp
@@ -26,20 +26,101 @@ Pass::Status WorkaroundImageOperands::Process() {
   return (modified ? Status::SuccessWithChange : Status::SuccessWithoutChange);
 }
 
-bool WorkaroundImageOperands::FixupOpTypeImage() {
-  auto* def_use_mgr = context()->get_def_use_mgr();
-  auto* type_mgr = context()->get_type_mgr();
+void WorkaroundImageOperands::WalkSampledImageUse(
+    Instruction* op_type_sampled_image_usage, uint32_t idx,
+    Instruction* img_type, analysis::SampledImage* sampled_image_type,
+    analysis::DefUseManager* def_use_mgr) {
+  auto op_sampled_image = op_type_sampled_image_usage;
+  context()->get_def_use_mgr()->ForEachUse(
+      op_sampled_image,
+      [this, op_sampled_image, sampled_image_type, img_type](
+          opt::Instruction* sampled_image_usage, uint32_t idx) {
+        bool has_depth = false;
+        if (sampled_image_usage->opcode() == SpvOpImageDrefGather ||
+            sampled_image_usage->opcode() == SpvOpImageSampleDrefImplicitLod ||
+            sampled_image_usage->opcode() == SpvOpImageSampleDrefExplicitLod ||
+            sampled_image_usage->opcode() ==
+                SpvOpImageSampleProjDrefImplicitLod ||
+            sampled_image_usage->opcode() ==
+                SpvOpImageSampleProjDrefExplicitLod ||
+            sampled_image_usage->opcode() ==
+                SpvOpImageSparseSampleDrefImplicitLod ||
+            sampled_image_usage->opcode() ==
+                SpvOpImageSparseSampleDrefExplicitLod ||
+            sampled_image_usage->opcode() ==
+                SpvOpImageSparseSampleProjDrefImplicitLod ||
+            sampled_image_usage->opcode() ==
+                SpvOpImageSparseSampleProjDrefExplicitLod ||
+            sampled_image_usage->opcode() == SpvOpImageSparseDrefGather) {
+          has_depth = true;
+        }
+        if (!has_depth) {
+          return;
+        }
+        analysis::Image* image_type = context()
+                                          ->get_type_mgr()
+                                          ->GetType(img_type->result_id())
+                                          ->AsImage();
+        analysis::Type* t = context()->get_type_mgr()->GetType(
+            context()->get_type_mgr()->GetId(image_type->sampled_type()));
+        analysis::Image img(
+            t, image_type->dim(), has_depth ? 1 : 0, image_type->is_arrayed(),
+            image_type->is_multisampled(), image_type->sampled(),
+            image_type->format(), image_type->access_qualifier());
+        for (auto& dec : image_type->decorations()) {
+          img.AddDecoration(std::vector<uint32_t>(dec));
+        }
+        uint32_t new_image =
+            context()->get_type_mgr()->GetTypeInstruction(&img);
 
+        analysis::SampledImage sampled_img(
+            context()->get_type_mgr()->GetType(new_image));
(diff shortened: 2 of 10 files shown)


source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,173 @@
...
+bool DeanonymizeImages::FixupOpTypeImage() {
+  auto* def_use_mgr = context()->get_def_use_mgr();
+  auto* type_mgr = context()->get_type_mgr();
+
+  bool modified = false;
+  auto* module = get_module();
+
+  std::vector<analysis::Image*> unknown_depth_images;
+
+  for (auto type : *type_mgr) {
```
AWoloszyn: Should I even be using the type_mgr here? Or should I just deal directly with  ir::Instruction*s.
The typeMgr might be overkill?

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,173 @@
...
+        modified = true;
+      }
+    }
+  }
+
+  std::vector<ir::Instruction*> fixed_instructions;
+
+  for (auto img_type : unknown_depth_images) {
+    // Find all of the places we used this image
+    def_use_mgr->ForEachUse(
```
AWoloszyn: Similar to above, is using the def_use_mgr here the right solution, or should I just iterate all of the instructions instead.

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,173 @@
...
+                            sampled_img.AddDecoration(
+                                std::vector<uint32_t>(dec));
+                          }
+
+                          uint32_t sampled_image_type =
+                              type_mgr->GetTypeInstruction(&sampled_img);
+
+                          // Create a new OpSampledImage value that uses that
+                          // type.
+                          auto new_sampled_image =
```
AWoloszyn: If I use the def_use_mgr, how do I make sure that these get tracked correctly, I tried to register them, but it didn't seem to work.

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,173 @@
...
+                // We should also backtrack to find the actual textures used, and switch
+                // its type to the new depth image.
+                // If we have both? not sure what the right answer is, probably pick one
+                // or the other (depth = 1 is probably safe)
+          }
+        });
+  }
+
+  for (auto img_type : unknown_depth_images) {
+    analysis::Type* t =
```
AWoloszyn: Similar to above, I tried to use the def_use_mgr here to update all, refs to a new type, but it seemed to clobber all of my previous changes. (Probably because I wasn't updating it correctly?)

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,173 @@
...
+        type_mgr->GetType(type_mgr->GetId(img_type->sampled_type()));
+    ir::Instruction* inst =
```
AWoloszyn: It seems like I am mixing the ir::Instruction and analysis::Type* which seems wrong.

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,173 @@
...
+                              sampled_image_usage->opcode() ==
+                                  SpvOpImageSparseDrefGather) {
+                            has_depth = true;
+                          }
+                          // If any of these are a dref lookup
+                          if (!has_depth) {
+                            return;
+                          }
+                          analysis::Type* t = type_mgr->GetType(
+                              type_mgr->GetId(img_type->sampled_type()));
```
s-perron: Why not set `t` to `img_type->sampled_type()`?  Is there a reason `t` is not const?

source/opt/workaround_image_operands.cpp
s-perron: You need to call `context()->AnalyzeDefUse(new_sampled_image)` to make the def-use manager aware of the new instruction.  Do this after you have finished updating the instruction with the new result id.

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,173 @@
...
+                              op_sampled_image->Clone(context());
+                          new_sampled_image->SetResultType(sampled_image_type);
+
+                          new_sampled_image->InsertBefore(op_sampled_image);
+                          new_sampled_image->SetResultId(
+                              context()->TakeNextId());
+
+                          // Update the DRef instruction to point to the new sampled image.
+                          sampled_image_usage->SetOperand(
+                              idx, {new_sampled_image->result_id()});
```
s-perron: After setting the operand you should call `context->AnalyzeUses(sampled_image_usage)`.

source/opt/workaround_image_operands.cpp
s-perron: Yes, I pointed out what was missing above.

source/opt/workaround_image_operands.cpp
s-perron: This will make the TypeManager inconsistent, but we do not invalidate it.  Create a new type, and use `context->ReplaceAllUses(...)`.  This should be okay once the def-use manager is kept up-to-date.

source/opt/workaround_image_operands.cpp
AWoloszyn: The analysis::Image constructor takes a non-const analysis::Type*. 
This may have just been accidental (I doesn't LOOK like it needs a mutable type), but I was avoiding making external changes. I can make Image take a only hold onto a const Type, which would simplify this.

s-perron: There is some inconsistency in the use of const in the type manager.  This is fine for now.  We should have a follow up issue to fix it.  I have a PR #1584 that does some fix up.

dj2:
Is this change still in progress?

AWoloszyn:
This is technically still in progress. I have not had a lot of time recently to come back and finish it. I will try to get to it soon

AWoloszyn:
Thanks @s-perron . I think I have responded to all of your comments.
I still have to squash, and write tests. But I have a question:

Now that I use ReplaceAllUses, the new type ends up at the end of the type-manager. This is a problem since there are other types that depend on it (E.g. an OpTypePointer).

Is there any way to change the location that it will be inserted? Also am I possibly just doing this wrong.

source/opt/types.cpp
```diff
@@ -328,25 +328,43 @@ Image::Image(Type* type, SpvDim dimen, uint32_t d, bool array, bool multisample,
...
+      format_(f),
+      access_qualifier_(SpvAccessQualifierReadOnly),
+      has_access_qualifier_(false) {}
+
 bool Image::IsSameImpl(const Type* that, IsSameCache* seen) const {
   const Image* it = that->AsImage();
   if (!it) return false;
   return dim_ == it->dim_ && depth_ == it->depth_ && arrayed_ == it->arrayed_ &&
          ms_ == it->ms_ && sampled_ == it->sampled_ && format_ == it->format_ &&
          access_qualifier_ == it->access_qualifier_ &&
```
s-perron: In other parts, you ignore `access_qualifier_` if `has_access_qualifer_` is false.  Why not here as well?

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,160 @@
...
+  bool modified = false;
+  auto* module = get_module();
+  auto* def_use_mgr = context()->get_def_use_mgr();
+  std::vector<opt::Instruction*> unknown_depth_images;
+
+  for (Module::inst_iterator type_it = context()->types_values_begin();
+       type_it != context()->types_values_end(); ++type_it) {
+    // 2 means no indication as to whether this is a depth or non-depth image
+    if (type_it->opcode() == SpvOpTypeImage) {
+      if (type_it->GetSingleWordOperand(3) == 2) {
```
s-perron: Could use a named constant for `3` and `2`.  Might make it clearer.

source/opt/workaround_image_operands.h
```diff
@@ -0,0 +1,43 @@
...
+namespace opt {
+
+// See optimizer.hpp for documentation.
+class WorkaroundImageOperands : public Pass {
+ public:
+  const char* name() const override { return "workaround-image-operands"; }
+  Status Process() override;
+
+ private:
+  void WalkImageUse(Instruction* use, Instruction* img_type,
```
s-perron: Add comments to describe what the functions do.  It is not clear what happens during the walk.

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+        sampled_image_usage->SetOperand(idx, {new_sampled_image->result_id()});
+        context()->AnalyzeUses(sampled_image_usage);
+        context()->AnalyzeDefUse(new_sampled_image);
+      });
+}
+
+void WorkaroundImageOperands::WalkImageUse(
+    Instruction* inst, Instruction* img_type,
+    analysis::DefUseManager* def_use_mgr) {
```
s-perron: You probably don't need to pass the def_use_mgr around.  It is part of the context, which is available anywhere in the pass.

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,160 @@
...
+    Instruction* op_type_sampled_image_usage, uint32_t idx,
+    Instruction* img_type, analysis::SampledImage* sampled_image_type,
+    analysis::DefUseManager* def_use_mgr) {
+  auto op_sampled_image = op_type_sampled_image_usage;
+  context()->get_def_use_mgr()->ForEachUse(
+      op_sampled_image,
+      [this, op_sampled_image, sampled_image_type, img_type](
+          opt::Instruction* sampled_image_usage, uint32_t idx) {
+        bool has_depth = false;
+        if (sampled_image_usage->opcode() == SpvOpImageDrefGather ||
```
s-perron: A switch statement might be cleaner.

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,160 @@
...
+          return;
+        }
+        analysis::Image* image_type = context()
+                                          ->get_type_mgr()
+                                          ->GetType(img_type->result_id())
+                                          ->AsImage();
+        analysis::Type* t = context()->get_type_mgr()->GetType(
+            context()->get_type_mgr()->GetId(image_type->sampled_type()));
+        analysis::Image img(
+            t, image_type->dim(), has_depth ? 1 : 0, image_type->is_arrayed(),
```
s-perron: if `has_depth` is false, we would have already returned right?

source/opt/workaround_image_operands.cpp
```diff
@@ -0,0 +1,160 @@
...
+    analysis::Type* t = context()->get_type_mgr()->GetType(
+        context()->get_type_mgr()->GetId(image_type->sampled_type()));
+    analysis::Image img(t, image_type->dim(), 0, image_type->is_arrayed(),
+                        image_type->is_multisampled(), image_type->sampled(),
+                        image_type->format(), image_type->access_qualifier());
+    for (auto& dec : image_type->decorations()) {
+      img.AddDecoration(std::vector<uint32_t>(dec));
+    }
+    uint32_t new_image = context()->get_type_mgr()->GetTypeInstruction(&img);
+    context()->ReplaceAllUsesWith(img_type->result_id(), new_image);
```
s-perron: For this change, as you mentioned, it is possible that `new_image` does not dominate all of the uses of `img_type`, so using`ReplaceAllUsesWith` can create invalid code.

This can happen two ways:  `new_image` is a newly created type or it is an existing type that was declared after `img_type`.  I see two possible solution.  The first it to try to rewrite the `img_type`.  You can only do this if you are sure there it will not create a duplicate type.  

The other would be to move `new_image` to just before `img_type` (`new_image->InsertAfter(img_type)`).  You would have to do a scan to make sure `new_image` is being moved earlier and not later.

s-perron:
@AWoloszyn I this still something you want to add?  If you are not actively working on it, I would like to close the PR.  You will still have your branch if you want to resubmit later.

AWoloszyn:
I will close. I want to get around to this, but things keep getting in the way. I will re-submit if I get a chance. I have found a different workaround for the problem on the specific devices anyway.

Issue #1578, opened 2018-05-25
Segfault in spvtools::opt::analysis::Struct::Struct for structs pointing to itself


karolherbst:
SPIR-V:

```SPIRV
; SPIR-V
; Version: 1.0
; Generator: Khronos LLVM/SPIR-V Translator; 14
; Bound: 12
; Schema: 0
               OpCapability Addresses
               OpCapability Kernel
          %1 = OpExtInstImport "OpenCL.std"
               OpMemoryModel Physical64 OpenCL
               OpEntryPoint Kernel %7 "test"
               OpSource OpenCL_C 102000
               OpDecorate %11 FuncParamAttr NoCapture
         %11 = OpDecorationGroup
               OpGroupDecorate %11 %8 %9
               OpTypeForwardPointer %_ptr_CrossWorkgroup__struct_3 CrossWorkgroup
       %void = OpTypeVoid
  %_struct_3 = OpTypeStruct %_ptr_CrossWorkgroup__struct_3
%_ptr_CrossWorkgroup__struct_3 = OpTypePointer CrossWorkgroup %_struct_3
%_ptr_CrossWorkgroup__struct_3_0 = OpTypePointer CrossWorkgroup %_struct_3
          %6 = OpTypeFunction %void %_ptr_CrossWorkgroup__struct_3_0 %_ptr_CrossWorkgroup__struct_3_0
          %7 = OpFunction %void Pure %6
          %8 = OpFunctionParameter %_ptr_CrossWorkgroup__struct_3_0
          %9 = OpFunctionParameter %_ptr_CrossWorkgroup__struct_3_0
         %10 = OpLabel
               OpReturn
               OpFunctionEnd
```

ran as gdb --args spirv-opt --skip-validation -O (validation needs to be skipped due to #1577):
```
#0  0x00007ffff7a85db2 in spvtools::opt::analysis::Struct::Struct (this=0x634dc0, types=std::vector of length 1, capacity 1 = {...}) at ../source/opt/types.cpp:412
#1  0x00007ffff7a799bd in spvtools::opt::analysis::TypeManager::RecordIfTypeDefinition (this=0x634b40, inst=...) at ../source/opt/type_manager.cpp:556
#2  0x00007ffff7a7485c in spvtools::opt::analysis::TypeManager::AnalyzeTypes (this=0x634b40, module=...) at ../source/opt/type_manager.cpp:69
[...]

PR #1584, opened 2018-05-28
Handle types with self references.

s-perron:
By using forward pointers, we are able to define a struct that has a
pointer to itself.  This could be directly or indirectly.  The current
implementation of the type manager did not handle this case.  There are
three changes that are made in this commit inorder to handle this case:

1) Change the handling of OpTypeForwardPointer

The current handling of OpTypeForwardsPointer is broken if there is a
reference to the pointer before the real definition.  When build the
type that contain the forward delared pointer, the type manager will ask
for the type for that ID, and will get a nullptr because it does not
exists.  This nullptr is not handleded very well.

The change is to keep track of the incomplete types the first time
through all of the types.  An incomplete type is a ForwardPointer or any
type that references an incomplete type.

Then we implement a second pass through the incomplete types that will
complete them.

2) Hashing types.

When hashing a type, we want to uses all of the subtypes as part of the
hash.  However, with types that reference them selves, this creates an
infinite recursion.  To get around this, we keep track of which types
have been seen on the path from the root type.  If we have see the
current type already then we can stop the recursion.

3) Comparing types.

In order to check if two types are the same, we must check that all of
their subtypes are the same as well.  This also causes an infinit
recursion.  The solution is to stop comparing the subtypes if we are
trying to compare two pointer types that we are already in the middle of
comparing.  The ideas is that if the two pointer are different, then in
progress compare will return false itself.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1578.

### DIFF
--- a/source/opt/types.cpp
+++ b/source/opt/types.cpp
@@ -16,6 +16,7 @@
 #include <cassert>
 #include <cstdint>
 #include <sstream>
+#include <unordered_set>
 
 #include "types.h"
 
@@ -171,7 +172,12 @@ bool Type::operator==(const Type& other) const {
   }
 }
 
-void Type::GetHashWords(std::vector<uint32_t>* words) const {
+void Type::GetHashWords(std::vector<uint32_t>* words,
+                        std::unordered_set<const Type*>* seen) const {
+  if (!seen->insert(this).second) {
+    return;
+  }
+
   words->push_back(kind_);
   for (const auto& d : decorations_) {
     for (auto w : d) {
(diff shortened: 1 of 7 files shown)


s-perron:
A bit premature creating this PR.  I forgot I need to do more that 1 pass over the incomplete types.  That needs to be a loop until we reach a fixed point.  If there are still incomplete types after the fixed point, then there was a error in the input.

s-perron:
@alan-baker I think it is good for a review now.  The travis failure is just missing parenthesis in an assert.  I can fix that with any changes based on the code review.

source/opt/types.cpp
```diff
@@ -171,7 +172,13 @@ bool Type::operator==(const Type& other) const {
   }
 }
 
-void Type::GetHashWords(std::vector<uint32_t>* words) const {
+void Type::GetHashWords(std::vector<uint32_t>* words,
+                        std::unordered_set<const Type*>* seen) const {
+  if (seen->count(this)) {
```
alan-baker: if (!seen->insert(this).second)
  return

PR #1586, opened 2018-05-28
Disallow array-of-arrays with DescriptorSets when validating.

### LINKED ISSUE #1522: Vulkan: at most one level of arrays are permitted at the descriptor interface
Vulkan internal issue 1192.

SPV_EXT_descriptor_indexing introduces the ability to use arrays of descriptors.  GLSL permits arrays of arrays, but Vulkan (will) only accept SPIR-V having a single level of array at the descriptor interface.  Front ends (like Glslang) are free to flatten arrays of arrays into a single level array.

dj2:
This CL adds validation to disallow using an array-of-arrays when attached to a
DescriptorSet.

Fixes #1522


dj2:
@alan-baker does this seem like it's the right track? I didn't gate the check on anything at this point it's just looking for DescriptorSet with array-of-arrays.

test/val/val_decoration_test.cpp
```diff
@@ -548,4 +548,159 @@ OpFunctionEnd
                         "Component decorations"));
 }
 
+TEST_F(ValidateDecorations, RuntimeArrayOfDescriptorSetsIsAllowed) {
```
alan-baker: These tests are long enough that it would be useful to add the GLSL code in comments for them.

source/validate_decorations.cpp
```diff
@@ -180,6 +180,43 @@ spv_result_t CheckDecorationsOfEntryPoints(ValidationState_t& vstate) {
...
+        SpvOpTypeArray != typePtr->opcode()) {
+      continue;
+    }
+
+    // Check for a second level array
+    const auto secondaryTypePtr = vstate.FindDef(typePtr->words()[2]);
+    if (SpvOpTypeRuntimeArray == secondaryTypePtr->opcode() ||
+        SpvOpTypeArray == secondaryTypePtr->opcode()) {
+      return vstate.diag(SPV_ERROR_INVALID_ID)
+             << "Array of arrays is not allowed.";
```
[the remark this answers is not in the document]
dj2: done

test/val/val_decoration_test.cpp
dj2: done

test/val/val_decoration_test.cpp
```diff
@@ -548,4 +548,179 @@ OpFunctionEnd
...
+         %30 = OpImageSampleExplicitLod %v4float %25 %28 Lod %float_0
+               OpStore %v %30
+               OpReturn
+               OpFunctionEnd
+)";
+  CompileSuccessfully(spirv, env);
+
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateAndRetrieveValidationState(env));
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("Array of arrays is not allowed"));
```
alan-baker: I think you forgot to update this error message.

test/val/val_decoration_test.cpp
```diff
@@ -548,4 +548,179 @@ OpFunctionEnd
...
+       %main = OpFunction %void None %3
+          %5 = OpLabel
+               OpReturn
+               OpFunctionEnd
+)";
+  CompileSuccessfully(spirv, env);
+
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateAndRetrieveValidationState(env));
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("Array of arrays is not allowed"));
```
alan-baker: Same here

test/val/val_decoration_test.cpp
dj2: Sigh, yes I did. Thanks.

test/val/val_decoration_test.cpp
dj2: done.

PR #1589, opened 2018-05-31
Validate static uses of interfaces

### LINKED ISSUE #1120: OpEntryPoint: validate Interface objects
- Interface objects are Ids of variables in Input or Output storage class.
- If code in the static call tree of the entry point references an Input or Output variable, then the variable must be listed in the Interfaces list of the entry point

alan-baker:
Fixes #1120

Checks that all static uses of the Input and Output variables are listed
as interfaces in each corresponding entry point declaration.
 * Changed validation state to track interface lists
 * updated many tests

### DIFF
--- a/source/val/validation_state.h
+++ b/source/val/validation_state.h
@@ -155,12 +155,17 @@ class ValidationState_t {
   /// instruction
   bool in_block() const;
 
+  struct EntryPointDescription {
+    std::string name;
+    std::vector<uint32_t> interfaces;
+  };
+
   /// Registers |id| as an entry point with |execution_model| and |interfaces|.
   void RegisterEntryPoint(const uint32_t id, SpvExecutionModel execution_model,
-                          std::vector<uint32_t>&& interfaces) {
+                          EntryPointDescription&& desc) {
     entry_points_.push_back(id);
     entry_point_to_execution_models_[id].insert(execution_model);
-    entry_point_interfaces_[id].emplace_back(interfaces);
+    entry_point_descriptions_[id].emplace_back(desc);
   }
 
   /// Returns a list of entry point function ids
@@ -172,11 +177,10 @@ class ValidationState_t {
     entry_point_to_execution_modes_[entry_point].insert(execution_mode);
   }
 
-  /// Returns the interface lists of a given entry point. If the given id is not
-  /// a valid Entry Point id, std::out_of_range exception is thrown.
-  const std::vector<std::vector<uint32_t>>& entry_point_interfaces(
-      uint32_t entry_point) const {
-    return entry_point_interfaces_.at(entry_point);
+  /// Returns the interface descriptions of a given entry point.
+  const std::vector<EntryPointDescription>& entry_point_descriptions(
+      uint32_t entry_point) {
+    return entry_point_descriptions_.at(entry_point);
   }
 
   /// Returns Execution Models for the given Entry Point.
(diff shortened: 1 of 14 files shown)


source/val/validation_state.h
```diff
@@ -155,32 +155,26 @@ class ValidationState_t {
...
   void RegisterExecutionModeForEntryPoint(uint32_t entry_point,
                                           SpvExecutionMode execution_mode) {
     entry_point_to_execution_modes_[entry_point].insert(execution_mode);
   }
 
-  /// Returns the interfaces of a given entry point. If the given id is not a
-  /// valid Entry Point id, std::out_of_range exception is thrown.
-  const std::vector<uint32_t>& entry_point_interfaces(
+  /// Returns the interface lists of a given entry point. If the given id is not
+  /// a valid Entry Point id, std::out_of_range exception is thrown.
```
dneto0: Hm. We build without exceptions IIRC, so throwing is not a thing.
Anyway, ok to leave that in. It was a pre-existing issue.

source/validate_decorations.cpp
```diff
@@ -122,58 +122,66 @@ spv_result_t CheckDecorationsOfEntryPoints(ValidationState_t& vstate) {
...
-      }
-      if (type_instr && SpvOpTypeStruct == type_instr->opcode() &&
-          isBuiltInStruct(type_id, vstate)) {
-        if (storage_class == SpvStorageClassInput) ++num_builtin_inputs;
-        if (storage_class == SpvStorageClassOutput) ++num_builtin_outputs;
-        if (num_builtin_inputs > 1 || num_builtin_outputs > 1) break;
-        if (auto error = CheckBuiltInVariable(interface, vstate)) return error;
-      } else if (isBuiltInVar(interface, vstate)) {
-        if (auto error = CheckBuiltInVariable(interface, vstate)) return error;
+    for (const auto& interface_list : interfaces) {
```
dneto0: This only looks like a big change because indentation changed.  I wish GItHub differ would make that more clear...

test/val/CMakeLists.txt
```diff
@@ -192,3 +192,9 @@ add_spvtools_unittest(TARGET val_non_uniform
         ${VAL_TEST_COMMON_SRCS}
   LIBS ${SPIRV_TOOLS}
 )
+
+add_spvtools_unittest(TARGET val_interfaces
```
dneto0: Please rebase against master.  Insert this new test file in the appriopriate list:   val_ijklmnop

source/validate_interfaces.cpp
```diff
@@ -0,0 +1,112 @@
...
+        if (var->id() == interface) {
+          found = true;
+          break;
+        }
+      }
+      if (!found) {
+        return _.diag(SPV_ERROR_INVALID_ID)
+               << (var->word(3u) == SpvStorageClassInput ? "Input" : "Output")
+               << " variable id <" << var->id()
+               << "> is used by entry point id <" << id
```
dneto0: Can we print the entry point name as well?  It definitely has one.

source/val/validation_state.h
alan-baker: Ended up changing this since I changed the method again.

test/val/CMakeLists.txt
alan-baker: Done

source/validate_interfaces.cpp
alan-baker: Done. ValidationState_t now stores names and interface lists together in a simple struct.

dneto0:
Rebased, squashed, and pushed into master as 4f866abfd85

PR #1599, opened 2018-06-04
Improve vector-dce compile time.

s-perron:
The compile time for vector dce can be pretty long.  The first problem
is that we are passing a large map by value to a lambda.  That was
intended to be by reference.

The second change is to implement the map using a vector of BitVectors
instead of an std::unordered_map.  The std::unordered_map implementation
can be slow, and as we saw with ADCE, the vector is not so sparse that
the vector becomes slow.

### DIFF
--- a/source/opt/vector_dce.cpp
+++ b/source/opt/vector_dce.cpp
@@ -278,13 +278,18 @@ bool VectorDCE::RewriteInstructions(
     const VectorDCE::LiveComponentMap& live_components) {
   bool modified = false;
   function->ForEachInst(
-      [&modified, this, live_components](ir::Instruction* current_inst) {
+      [&modified, this, &live_components](ir::Instruction* current_inst) {
         if (!context()->IsCombinatorInstruction(current_inst)) {
           return;
         }
 
-        auto live_component = live_components.find(current_inst->result_id());
-        if (live_component == live_components.end()) {
+        if (current_inst->result_id() >= live_components.size()) {
+          return;
+        }
+
+        utils::BitVector* live_component =
+            live_components[current_inst->result_id()].get();
+        if (!live_component) {
           // If this instruction is not in live_components then it does not
           // produce a vector, or it is never referenced and ADCE will remove
           // it.  No point in trying to differentiate.
@@ -293,7 +298,7 @@ bool VectorDCE::RewriteInstructions(
 
         // If no element in the current instruction is used replace it with an
         // OpUndef.
-        if (live_component->second.Empty()) {
+        if (live_component->Empty()) {
           modified = true;
           uint32_t undef_id = this->Type2Undef(current_inst->type_id());
           context()->KillNamesAndDecorates(current_inst);
@@ -304,8 +309,7 @@ bool VectorDCE::RewriteInstructions(
 
         switch (current_inst->opcode()) {
           case SpvOpCompositeInsert:
-            modified |=
-                RewriteInsertInstruction(current_inst, live_component->second);
+            modified |= RewriteInsertInstruction(current_inst, *live_component);
             break;
           case SpvOpCompositeConstruct:
             // TODO: The members that are not live can be replaced by an undef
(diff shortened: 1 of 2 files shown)


ratchetfreak:
With all the map<uint32_t, T> performance issues lately wouldn't it be time to create a proper data structure for it?

A vector sized to maxid for when it's small and dense and a hybrid `vector<T>` + `map<uint32_t, T>` or sorted `vector<pair<uint32_t, T>>`  for when it's large&sparse.

dneto0:
> A vector sized to maxid for when it's small and dense and a hybrid vector<T> + map<uint32_t, T> or sorted vector<pair<uint32_t, T>> for when it's large&sparse.

You'd have to have a way to indicate an empty slot, which is at least one bit for each entry.

We have a hybrid data structure for enum sets optimized for the case where the Ids are less than 63; it was really made for capability sets.  See source/enum_set.h

source/opt/vector_dce.cpp
```diff
@@ -351,13 +355,19 @@ void VectorDCE::AddItemToWorkListIfNeeded(
     WorkListItem work_item, VectorDCE::LiveComponentMap* live_components,
     std::vector<WorkListItem>* work_list) {
   ir::Instruction* current_inst = work_item.instruction;
-  auto it = live_components->find(current_inst->result_id());
-  if (it == live_components->end()) {
-    live_components->emplace(
-        std::make_pair(current_inst->result_id(), work_item.components));
+
+  if (current_inst->result_id() >= live_components->size()) {
+    live_components->resize(current_inst->result_id() + 100);
```
dneto0: I think this is ok.  But it could be pathological.
Amortized constant time (in theory) would be gained by a multiplicative factor expansion, rather than additive term.

source/opt/vector_dce.cpp
```diff
@@ -351,13 +355,19 @@ void VectorDCE::AddItemToWorkListIfNeeded(
     WorkListItem work_item, VectorDCE::LiveComponentMap* live_components,
     std::vector<WorkListItem>* work_list) {
   ir::Instruction* current_inst = work_item.instruction;
-  auto it = live_components->find(current_inst->result_id());
-  if (it == live_components->end()) {
-    live_components->emplace(
-        std::make_pair(current_inst->result_id(), work_item.components));
+
+  if (current_inst->result_id() >= live_components->size()) {
```
dneto0: There's an opportunity to cache the result id. ... but that might be a bad micro-optimization.
This is fine.

source/opt/vector_dce.cpp
```diff
@@ -351,13 +355,19 @@ void VectorDCE::AddItemToWorkListIfNeeded(
...
+    live_components->resize(current_inst->result_id() + 100);
+  }
+
+  utils::BitVector* live_component =
```
dneto0: You could have saved the reference result from live_components->at(..) so you wouldn't do the lookup twice.... but again that's a micro-optimization that's probably not worth it.

ratchetfreak:
> You'd have to have a way to indicate an empty slot, which is at least one bit for each entry.

> We have a hybrid data structure for enum sets optimized for the case where the Ids are less than 63; it was really made for capability sets. See source/enum_set.h

Depends on whether T is nullable or not and if you want nulls as values.

I can think of several ways to implement this, however the optimal one depends on access patterns and such, so without examining those I shouldn't say too much. However it is starting to become clear that std::map is not the solution and std::vector directly indexed by the ID is vulnerable to pathological and malicious cases.

source/opt/vector_dce.cpp
s-perron: This is implemented by an std::vector.  The size of the vector is not the same as the amount of space allocated.  I am leaving it up to the implementation of std::vector to do a multiplicative expansion when the memory allocated needs expanded.

I wanted to keep a smaller size for the vector to reduce the number of elements that need to be constructed and destructed, even if it is only ans std::unique_ptr that remains null.

s-perron:
I was getting some odd results.  As I look at this more, the gain I thought I had is not there.  I'll close this without merging.  I don't think it will make a difference.

Issue #1581, opened 2018-05-27
Can't decipher cryptic error message


paniq:
For the following code, `spirv-val` is producing the error
```
error: 2856: OpStore Pointer <id> '137' is not a logical pointer.
```
which is completely indecipherable. The numbers 2856 or 137 do not appear in the source code, neither is 2856 a line number, and line 137 has nothing to do with the error. How am I supposed to make sense of this?

Even when passing `--preserve-numeric-ids` to `spirv-as`, the output produced by `spirv-val` does not become any clearer, as the new number is also not used in the source:

```
error: 2856: OpStore Pointer <id> '666' is not a logical pointer.
```

```
; SPIR-V
; Version: 1.0
; Generator: Unknown(21315); 20304
; Bound: 1121
; Schema: 0
               OpCapability Shader
               OpCapability SampledBuffer
               OpCapability ImageQuery
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %test_fragment "main"
               OpSource GLSL 450
               OpName %test_fragment "test-fragment"
               OpName %_ ""
               OpName %uv "uv"
               OpName %shader "shader"
               OpName %ShaderGlobals "ShaderGlobals"
               OpMemberName %ShaderGlobals 0 "aspect"
               OpMemberName %ShaderGlobals 1 "time"
               OpMemberName %ShaderGlobals 2 "frame"
               OpName %shglobals "shglobals"
               OpName %__0 ""
               OpName %__1 ""
               OpName %__2 ""
               OpName %__3 ""
               OpName %vec_type_new "vec-type-new"
               OpName %__4 ""
               OpName %__5 ""
               OpName %versor "versor"
               OpName %__6 ""
               OpName %__7 ""
               OpName %vec_type_new_0 "vec-type-new"
               OpName %__8 ""
               OpName %vec_type_new_1 "vec-type-new"
               OpName %versor_rotate "versor-rotate"
               OpName %__9 ""
               OpName %__10 ""
               OpName %__11 ""
               OpName %__12 ""
               OpName %__13 ""
               OpName %__14 ""
               OpName %__15 ""
               OpName %__16 ""
               OpName %__17 ""
               OpName %dot "dot"
               OpName %__18 ""
               OpName %__19 ""
               OpName %vector_reduce "vector-reduce"
               OpName %vec_type_new_2 "vec-type-new"
[...]

PR #1600, opened 2018-06-04
[val] Add extra context to error messages.

### LINKED ISSUE #670: spirv-val has inconsistent location reporting for errors

dj2:
This CL extends the error messages produced by the validator to output the
disassembly of the errored line.

The validation_id messages have also been updated to print the line number of
the error instead of the word number. Note, the error number is from the start
of the SPIR-V, it does not include any headers printed in the disassembled code.

Fixes #670, #1581


dj2:
Please let me know if this is on the right track.

For issue 1581 this would output:
```
error: line 735: OpStore Pointer <id> '137' is not a logical pointer.
  %obj = OpPhi %_ptr_Function_bool %493 %549 %493 %550 %493 %546 ; 0x00000014
```

and for the example provided in 670 it would output:
```
error: line 1: TypeInt cannot appear before the memory model instruction
  %uint = OpTypeInt 32 0 ; 0x00000014
```

I think for 670 we may also want to change the <id> '137' to output <id> '137[obj]' or whatever the getIdName() string is for that object.

dj2:
Found the issue, I need to run cmake with -DSPIRV_BUILD_COMPRESSION=ON to match the bots.

dj2:
Will need to do a bit of work to make this build/run with the markv stuff. Will ping when the CL is read for review again.

dneto0:
FYI.  The Markv decoder case is the sticky one because the uncompressed form of the SPIR-V module is not anywhere in memory.  See https://github.com/KhronosGroup/SPIRV-Tools/pull/1221#issuecomment-358789690 for @atgoo's suggestion on how to resolve this.

source/val/validation_state.cpp
```diff
@@ -829,4 +845,20 @@ const std::vector<uint32_t>& ValidationState_t::FunctionEntryPoints(
...
+  return Disassemble(c_inst.words, c_inst.num_words);
+}
+
+std::string ValidationState_t::Disassemble(const uint32_t* words,
+                                           uint16_t num_words) const {
+  uint32_t disassembly_options = SPV_BINARY_TO_TEXT_OPTION_NO_HEADER |
+                                 SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES |
+                                 SPV_BINARY_TO_TEXT_OPTION_SHOW_BYTE_OFFSET;
+
+  return spvtools::spvInstructionBinaryToText(context()->target_env, words,
```
dneto0: Let's fix this in a subsequent CL, but this can be wrong for things like 
`OpConstant %int 42` because you need type info to know that %int is actually an int, and therefore how to interpret the "42" word.  (E.g. compare with a floating point constant.)
A second bad case is OpSwitch where again you need type info to know how many words the literals are.

Interestingly, the spv_parsed_instruction_t has the necessary info.

An actually correct solution would generalize the disassembler so that you pass in the entire binary in and ask only for a specific instruction. It would have to run through the entire binary to collect the type info, and only accumulate string output for the requested instruction (by index).

And yeah, I messed this up on my original attempt. It was quick and dirty and only 95% right.

source/diagnostic.h
```diff
@@ -30,8 +30,12 @@ class DiagnosticStream {
  public:
   DiagnosticStream(spv_position_t position,
```
dneto0: Given that there are so many times you call this constructor with an empty string value for the disassembled_instruction member, I would have probably made a second constructor.

test/c_interface_test.cpp
```diff
@@ -194,8 +194,10 @@ TEST(CInterface, SpecifyConsumerNullDiagnosticForValidating) {
         // TODO(antiagainst): what validation reports is not a word offset here.
         // It is inconsistent with diassembler. Should be fixed.
         EXPECT_EQ(1u, position.index);
-        EXPECT_STREQ("Nop cannot appear before the memory model instruction",
-                     message);
+        EXPECT_STREQ(
+            "Nop cannot appear before the memory model instruction\n"
+            "  OpNop ; 0x00000014\n",
```
dneto0: Let's lose the byte offset since it's a lie in the general case.

tools/io.h
```diff
@@ -39,7 +39,10 @@ bool ReadFile(const char* filename, const char* mode, std::vector<T>* data) {
       }
     } else {
       if (sizeof(T) != 1 && (ftell(fp) % sizeof(T))) {
-        fprintf(stderr, "error: corrupted word found in file '%s'\n", filename);
+        fprintf(
+            stderr,
+            "error: file size should be a multiple of %ld; file '%s' corrput\n",
```
dneto0: Spelling of "corrupt"

source/val/validation_state.cpp
```diff
@@ -829,4 +845,20 @@ const std::vector<uint32_t>& ValidationState_t::FunctionEntryPoints(
...
+std::string ValidationState_t::Disassemble(const Instruction& inst) const {
+  const spv_parsed_instruction_t& c_inst(inst.c_inst());
+  return Disassemble(c_inst.words, c_inst.num_words);
+}
+
+std::string ValidationState_t::Disassemble(const uint32_t* words,
+                                           uint16_t num_words) const {
+  uint32_t disassembly_options = SPV_BINARY_TO_TEXT_OPTION_NO_HEADER |
+                                 SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES |
+                                 SPV_BINARY_TO_TEXT_OPTION_SHOW_BYTE_OFFSET;
```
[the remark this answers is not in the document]
dneto0: Actually, if the strategy changes to have the whole binary in hand, and you really do know which instruction position (index) it is, then you should keep this show-byte-offset because then it's the truth and extra information the user could use.

source/diagnostic.cpp
```diff
@@ -71,6 +71,7 @@ DiagnosticStream::DiagnosticStream(DiagnosticStream&& other)
     : stream_(),
       position_(other.position_),
       consumer_(other.consumer_),
+      disassembled_instruction_(other.disassembled_instruction_),
```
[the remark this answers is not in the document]
dj2: done

source/diagnostic.h
dj2: I'm on the fence for this one. It makes it a lot clearer where we don't have a disassembled_instruction and, hopefully, would make it harder to add a missing one in the future. It's also possible, some of the fields I have as blank could have values, like bad header and a couple others.

source/val/validation_state.cpp
```diff
@@ -142,9 +142,13 @@ bool IsInstructionInLayoutSection(ModuleLayoutSection layout, SpvOp op) {
 }  // anonymous namespace
 
 ValidationState_t::ValidationState_t(const spv_const_context ctx,
-                                     const spv_const_validator_options opt)
+                                     const spv_const_validator_options opt,
+                                     const uint32_t* words,
```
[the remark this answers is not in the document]
dj2: Yea, realized this when I hit the bots and figured out they use different flags. Going to try to work out Markv and will ping when the updates are done.

source/val/validation_state.cpp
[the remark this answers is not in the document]
dj2: Removed for now. It was always outputting 0x14 and I couldn't figure out where it came from. We can always put back later if desired.

source/val/validation_state.cpp
dj2: So, is the problem that sometimes it would decompile one of:
```
OpConstant %int 42
OpConstant %int 42.0
```
because it doesn't know if it's a float or an int?

test/c_interface_test.cpp
dj2: done

tools/io.h
dj2: done

dj2:
PTAL. I changed the Markv code to do validation at the beginning of encode and the end of decode so we should have the full binary available. This removes the need to create a ValidationState in the MarkvDecode constructor.

source/val/validation_state.cpp
dneto0: Yes, that's one problem.

Another is with 

`OpSwitch %value %default  12 %label0  43 %label1 52 %label3`

The %value and literals are supposed to be integer scalars.  But you need to see the OpTypeInt to know how many bits wide it is, and therefore how many words wide it is.  That's needed to know how to decode the words for the cases.  E.g. 6 words could be 3 pairs of (32-bit literal, target-id) or 2 pairs of (64bit literal, target-id)

dj2:
Rebased, will see if AppVeyor is happy now.

source/diagnostic.h
dneto0: > would make it harder to add a missing one in the future.  

That's a good argument.

source/val/validation_state.cpp
```diff
@@ -254,9 +258,19 @@ bool ValidationState_t::IsOpcodeInCurrentLayoutSection(SpvOp op) {
...
 DiagnosticStream ValidationState_t::diag(spv_result_t error_code) const {
+  return diag(error_code, instruction_counter_);
+}
+
+DiagnosticStream ValidationState_t::diag(spv_result_t error_code,
+                                         int instruction_counter) const {
+  std::string disassembly;
+  if (instruction_counter >= 0 && static_cast<size_t>(instruction_counter) <=
+                                      ordered_instructions_.size()) {
+    disassembly = Disassemble(ordered_instructions_[instruction_counter - 1]);
```
dneto0: This produces the right output in almost all situations.
Corner cases where it does not are where type info from prior instructions are required to correctly parse:  e.g. OpConstnat %int ....  and also OpSwitch 

That can wait for a future PR.  I'll file an issue for this.

dneto0: I filed #1627 but the results are better than I expected. :-)

PR #1601, opened 2018-06-04
[val] Output id names along with numbers in validate_id

dj2:
This CL updates the validate_id code to output the name of the object along with
the id number. There were a few instances which already output the name, this
just extends to all of them. Now, the output should say 123[obj] instead of just
123.

Issue #1581


dj2:
For the example in 1581, this would output:

```
error: 2856: OpStore Pointer <id> '137[obj]' is not a logical pointer.
```

dj2:
The travis failure is due to formatting differences, probably because I ran clang-format on mac and it did some changes the linux version doesn't like. The appveyor failure is complaining about taking too many resources?

dj2:
Reformatted the test file by hand, the clang-format on my Linux machine also doesn't match what Travis expects?

dneto0:
Seems like formatting is under control now?

As for AppVeyor, could be the cloud weather.  I recommend doing commit --amend and force pushing onto your branch again to trigger a rebuild.

Issue #1596, opened 2018-06-01
Can't assemble ViewportRelativeNV


dneto0:
There's something weird with operand lookup.  Trying to assemble the following:
```
OpDecorateId %1 HlslCounterBufferGOOGLE %2
OpDecorate %1 ViewportRelativeNV
```
Fails with:
```
error: 2: 15: Invalid decoration 'ViewportRelativeNV'.
```
This seems to be related to the fact that that token is defined in the grammar with `version : "None"` to indicate a "Reserved" token, and hence not defined by any SPIR-V version.

But that's weird because I can assemble the HlslCounterBufferGOOGLE line. 

Something weird is going on.

antiagainst:
I think it's because the grammar rules for `ViewportRelativeNV` is missing `extensions` field right now.

antiagainst:
I fixed missing `extensions` field for multiple SPIR-V extensions with multiple PRs, but seems missing this one.

PR #1603, opened 2018-06-05
Add optionally unconditional operand lookup

dneto0:
Assembler and disassembler always do unconditional lookups.
Validator operand capability check does unconditional lookup.

Now we can assemble and disassemble tokens like ViewportRelativeNV
which is enabled by a capability, which in turn is is enabled by
an extension.  But we also check capabilities appropriately.
Add tests for that.

Fixes issue #1596


dneto0:
I'm not convinced this is the right solution.

PR #1604, opened 2018-06-06
Operand lookup succeeds if it's enabled by a capability

### LINKED ISSUE #1596: Can't assemble ViewportRelativeNV

dneto0:
- Fix tests for basic group operations (e.g. Reduce) to allow for
new capabilities in SPIR-V 1.3 that enable them.
- Refactor operand capability check to avoid code duplication and
to put all checks that don't need table lookup before any table
lookup.
- Test round trip assembly/disassembly support for extension
SPV_NV_viewport_array2
- Test assembly and validation of decoration ViewportRelativeNV

Fixes #1596


test/binary_parse_test.cpp
```diff
@@ -887,8 +887,8 @@ INSTANTIATE_TEST_CASE_P(
          "component 32"},
         {"%2 = OpFunction %2 !31",
          "Invalid function control operand: 31 has invalid mask component 16"},
-        {"OpLoopMerge %1 %2 !7",
-         "Invalid loop control operand: 7 has invalid mask component 4"},
+        {"OpLoopMerge %1 %2 !1027",
```
dneto0: SPIR-V 1.1 introduced DependencyInfinite with bit value 4, so I bumped up the bad case to something larger than anything in the table in SPIR-V 1.3

test/text_to_binary.extension_test.cpp
```diff
@@ -145,19 +145,19 @@ INSTANTIATE_TEST_CASE_P(
                  MakeInstruction(SpvOpSubgroupBallotKHR, {1, 2, 3})},
                 {"%2 = OpSubgroupFirstInvocationKHR %1 %3\n",
                  MakeInstruction(SpvOpSubgroupFirstInvocationKHR, {1, 2, 3})},
-                {"OpDecorate %1 BuiltIn SubgroupEqMaskKHR\n",
+                {"OpDecorate %1 BuiltIn SubgroupEqMask\n",
```
dneto0: In SPIR-V 1.3 these toke numbers are pulled into core, so the preferred name is used here for the round trip assembly/disassembly test.  That drops the KHR suffix.

test/text_to_binary.extension_test.cpp
```diff
@@ -523,6 +523,34 @@ INSTANTIATE_TEST_CASE_P(
...
+                // The EXT and NV extensions have the same token number for this
+                // capability.
+                {"OpCapability ShaderViewportIndexLayerEXT\n",
+                 MakeInstruction(SpvOpCapability,
+                                 {SpvCapabilityShaderViewportIndexLayerNV})},
+                // Check the new capability's token number
+                {"OpCapability ShaderViewportIndexLayerEXT\n",
+                 MakeInstruction(SpvOpCapability, {5254})},
+                // Decorations
+                {"OpDecorate %1 ViewportRelativeNV\n",
```
dneto0: Proves that I've fixed #1596

dneto0:
Hmmm.  Looks like we are consistently running out of time for the VS2017 debug build.  (cc @dj2 )

dneto0:
The paid plans for AppVeyor have a limit of 60minutes per job.  That's much more room than the 20minutes that I'm seeing now.

dj2:
What's the definition of a Job on AppVeyor? Is that a build, or each of the things inside a build? If the sum of the 3 things inside the build are > 60 minutes does that fail or if only one of the things is > 60 minutes? 

(The pricing page says 'All plans have maximum build job execution time of 60 minutes.' which could be read either way....)

dj2:
Looking at the  AppVeyor boards, this can also mean you ran out of disk space as the error just says 'Build exceeded allowed resource quotas. Fix your build to consume less resources or contact AppVeyor support to request quotas increase.'

See: https://help.appveyor.com/discussions/problems/14560-build-exceeded-allowed-resource-quotas and 
https://help.appveyor.com/discussions/problems/14543-build-exceeded-allowed-resource-quotas-fix-your-build-to-consume-less-resources-or-contact-appveyor-support-to-request-quotas-increase

So, maybe we just need to log a support request to get the resources upgraded?

dneto0:
Rebased, squashed, and pushed into master as 363bfca2ed888

Issue #1609, opened 2018-06-07
Small size optimizations for vectors


s-perron:
In the Instruction class and the Operand, there are std::vector member that typically have  a small number of elements.  For example, the Operand class has a vector of uint32_t to hold the words for that operand.  However, in most cases the number of integers needed for the words is 1.

In this issue, we want to:

- [x] Implement a replacement for std::vector that is optimized for small sizes.
- [x] Looking into using it in the Operand class.
- [x] Look into using it in the Instruction class.
- [ ] Look into using it in the BitVector class.

dj2:
We have a class in PDFium which does the optimization for small sizes and we're trying to get rid of it. In the long run, while it seemed to make thingsy faster, there is a maintenance overhead associated with the class that gets tedious.

We've also found that most cases where that special class was used, we'd get the same, or similar performance gain from reserving and resizing a vector instead of using a fixed size or vector class thing.

s-perron:
@dj2 For the operand class,  the replacement is initialized and then not resized afterwards.  The size is known by the constructor.  I'm assuming it is not faster to do a default construction, reserve, resize, and then copy over the data.  Also these arrays are 1 element 98% of the time.  

We can talk about your suggestions because it could help in other places.

s-perron:
I've tried using the SmallVector class for BitVector to help compile time in vector DCE.  It does not help.  I think some of the profiling I was doing was pointing me in the wrong place.  I'll view this item as done.

PR #1610, opened 2018-06-07
Small vector optimization for operands.

s-perron:
We replace the std::vector in the Operand class by a new class that does
a small size optimization.  This helps improve compile time on Windows.

Tested on three sets of shaders.  Trying various values for the small
vector.  The optimal value for the operand class was 2.  However, for
the Instruction class, using an std::vector was optimal.  Size of "0"
means that an std::vector was used.

<table>
  <tr>
   <td>OperandData Size
   </td>
   <td colspan="3" >OperandList size
   </td>
  </tr>
  <tr>
   <td>
   </td>
   <td>0
   </td>
   <td>4
   </td>
   <td>8
   </td>
  </tr>
  <tr>
   <td>0
   </td>
   <td>489s (baseline)    
   </td>
   <td>544s
   </td>
   <td>684s
   </td>
  </tr>
  <tr>
   <td>1
   </td>
   <td>593s    
   </td>
   <td>
   </td>
   <td>
   </td>
  </tr>
  <tr>
   <td>2
   </td>
   <td>469s
   </td>
   <td>487s
   </td>
   <td>
   </td>
  </tr>
  <tr>
   <td>4
   </td>
   <td>473s
   </td>
   <td>570s
   </td>
   <td>
   </td>
  </tr>
  <tr>
   <td>8
   </td>
   <td>505s
   </td>
   <td>
   </td>
   <td>
   </td>
  </tr>
</table>

This is a single thread run of ~120 shaders.  

For the multithreaded run the results were the similar.  The time went from ~62sec to ~38sec.  The
optimal configuration was an 2 for the OperandData and an
std::vector for the OperandList.  Similar
expiriments were done with other sets of shaders.  The compile time still
improved, but not as much.

Contributes to https://github.com/KhronosGroup/SPIRV-Tools/issues/1609.

### DIFF
--- a/source/util/small_vector.h
+++ b/source/util/small_vector.h
@@ -1,0 +1,464 @@
+// Copyright (c) 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef LIBSPIRV_UTILS_SMALL_VECTOR_H_
+#define LIBSPIRV_UTILS_SMALL_VECTOR_H_
+
+#include <cassert>
+#include <iostream>
+#include <vector>
+
+#include "opt/make_unique.h"
+
+namespace spvtools {
+namespace utils {
+
+// The |SmallVector| class is intended to be a drop-in replacement for
+// |std::vector|.  The difference is in the implementation. A |SmallVector| is
+// optimized for when the number of elements in the vector are small.  Small is
+// defined by the template parameter |small_size|.
+//
+// Note that |SmallVector| is not always faster than an |std::vector|, so you
+// should experiment with different values for |small_size| and compare to
+// using and |std::vector|.
+//
+// TODO: I have implemented the public member functions from |std::vector| that
+// I needed.  If others are needed they should be implemented. Do not implement
+// public member functions that are not defined by std::vector.
+template <class T, size_t small_size>
+class SmallVector {
+ public:
+  using iterator = T*;
+  using const_iterator = const T*;
+
+  SmallVector()
+      : size_(0),
+        small_data_(reinterpret_cast<T*>(buffer)),
+        large_data_(nullptr) {}
+
+  SmallVector(const SmallVector& that) : SmallVector() { *this = that; }
+
+  SmallVector(SmallVector&& that) : SmallVector() { *this = std::move(that); }
+
+  SmallVector(const std::vector<T>& vec) : SmallVector() {
+    if (vec.size() > small_size) {
+      large_data_.reset(new std::vector<T>(vec));
+    } else {
+      size_ = vec.size();
+      for (uint32_t i = 0; i < size_; i++) {
+        new (small_data_ + i) T(vec[i]);
+      }
+    }
+  }
+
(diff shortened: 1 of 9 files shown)


dj2:
Out of curiosity, did you try calling .reserve() on the vector to set the number of operands it would need to store or are they just added to the back?

ratchetfreak:
The unique_ptr to a std::vector seems like a waste (double indirection), checking against empty is just as fast as checking a pointer so you may as well keep it inline if you don't mind the extra space needed. Also std::aligned_storage is a better solution than a char array.

You are likely to get better performance for indexing and push_back by having a `iterator begin;` as a field and use that for the operator[]. You can even use that field to check whether it is pointing to the small buffer as the test to see whether the large or small buffer is being used.

This lets you recycle the storage space:

```
T* begin;//always points to &sb or bb.data()
size_t length;
union {
    std::aligned_storage<sizeof(T)*small_size, alignof(T)> sb;
    std::vector<T> bb;
}
```

This does require more lifetime finagling but moveToBig is then a std::move of sb to a local std::vector, destruct of the sb elements and then a placement move construct of local to bb.

s-perron:
@ratchetfreak I implemented you suggestion (or a least part of it) in https://github.com/s-perron/SPIRV-Tools/tree/small_vector2.  It did not give any improvments.  In general I agree it would be better to reuse the same data; however, in this case I do not believe the amount of data being allocated is the bottleneck.  Even large shaders are not that large.  We just want to avoid a lot of calls to malloc and free.

For now, I will go with the simpler implementation.  I do not like unions all that much because the make it harder for the compiler to optimize, and easier for the developer to make a mistake.

I also tried Abseil, and it does give a small improvement over my implementation, but it is not very much.  So there is room for improvement.  We can also look into this a little more later, but I want to get the majority of the improvement first, then we can work to get the rest.

dneto0:
Overall the approach seems fine and the code looks good.

However, bots fail compilation:

/home/travis/build/KhronosGroup/SPIRV-Tools/source/util/small_vector.h:336:12: error: variable has incomplete type 'void'
      auto ret = large_data_->insert(new_pos, first, last);

dneto0:
At one point I wondered if the creation of operand lists is sufficiently constrained that in the vast majority of cases we have all the operand values in hand at one time, so we can allocate their storage in bulk.  But that's a more intrusive change to this one.

source/util/small_vector.h
```diff
@@ -0,0 +1,463 @@
...
+    }
+
+    // Move |last| and any elements after it their earlier position.
+    while (l != e) {
+      *f = std::move(*l);
+      ++f;
+      ++l;
+    }
+
+    // Destroy the elements that were suppose to be deleted.
```
dnovillo: s/suppose/supposed/

source/util/small_vector.h
```diff
@@ -0,0 +1,463 @@
...
+namespace spvtools {
+namespace utils {
+
+// The |SmallVector| class is intended to be a drop-in replacement for
+// |std::vector|.  The difference is in the implementation. A |SmallVector| is
+// optimized for when the number of elements in the vector are small.  Small is
+// defined by the template parameter |small_size|.
+//
+// Note that |SmallVector| is not always faster than an |std::vector|, so you
+// should expiriement with different  values for |small_size| and compare to
```
dnovillo: s/expiriement/experiment/

There's two spaces after 'different'.

source/util/small_vector.h
```diff
@@ -0,0 +1,464 @@
...
+namespace spvtools {
+namespace utils {
+
+// The |SmallVector| class is intended to be a drop-in replacement for
+// |std::vector|.  The difference is in the implementation. A |SmallVector| is
+// optimized for when the number of elements in the vector are small.  Small is
+// defined by the template parameter |small_size|.
+//
+// Note that |SmallVector| is not always faster than an |std::vector|, so you
+// should expirement with different values for |small_size| and compare to
```
dneto0: s/expirement/experiment/

:-)

PR #1612, opened 2018-06-11
 Setup gclient and presubmit file.

dj2:
This CL adds the necessary files to use gclient and the depot_tools with
the SPIRV-Tools directory. This allows doing things like `git cl format`
to format code pre-upload and `git cl presubmit -uf` to run presubmit
checks over the code.

The dependencies are all added to the DEPS file and will be
auto-downloaded. They are all pin'd to specific revisions so everyone
has the same checkout. Clang is included in the checkout so it will be
consistent over usages.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,4 +2,5 @@
 Language: Cpp
 BasedOnStyle: Google
 DerivePointerAlignment: false
+SortIncludes: true
 ...
--- a/DEPS
+++ b/DEPS
@@ -1,0 +1,85 @@
+use_relative_paths = True
+
+vars = {
+  'chromium_git': 'https://chromium.googlesource.com',
+  'github': 'https://github.com',
+
+  'buildtools_revision': 'ab7b6a7b350dd15804c87c20ce78982811fdd76f',
+  'clang_revision': 'abe5e4f9dc0f1df848c7a0efa05256253e77a7b7',
+  'effcee_revision': '04b624799f5a9dbaf3fa1dbed2ba9dce2fc8dcf2',
+  'googletest_revision': '98a0d007d7092b72eea0e501bb9ad17908a1a036',
+  're2_revision': '6cf8ccd82dbaab2668e9b13596c68183c9ecd13f',
+  'spirv_headers_revision': '3ce3e49d73b8abbf2ffe33f829f941fb2a40f552',
+}
+
+deps = {
+  'buildtools':
+      Var('chromium_git') + '/chromium/buildtools.git@' +
+          Var('buildtools_revision'),
+
+  'external/spirv-headers':
+      Var('github') +  '/KhronosGroup/SPIRV-Headers.git@' +
+          Var('spirv_headers_revision'),
+
+  'external/googletest':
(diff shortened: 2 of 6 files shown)


DEPS
```diff
@@ -0,0 +1,85 @@
+use_relative_paths = True
+
+vars = {
```
dneto0: Hm. This isn't the DEPS file format I remember.  Then again, it's been a long time and my memory could be considered....fuzzy.

DEPS
```diff
@@ -0,0 +1,85 @@
...
+  'chromium_git': 'https://chromium.googlesource.com',
+  'github': 'https://github.com',
+
+  'buildtools_revision': 'ab7b6a7b350dd15804c87c20ce78982811fdd76f',
+  'clang_revision': 'abe5e4f9dc0f1df848c7a0efa05256253e77a7b7',
+  'effcee_revision': '04b624799f5a9dbaf3fa1dbed2ba9dce2fc8dcf2',
+  'googletest_revision': '98a0d007d7092b72eea0e501bb9ad17908a1a036',
+  're2_revision': '6cf8ccd82dbaab2668e9b13596c68183c9ecd13f',
+  'spirv_headers_revision': '3ce3e49d73b8abbf2ffe33f829f941fb2a40f552',
```
dneto0: We'll need a defined process for updating these commit hashes.  I'm not saying it has to be now now now, but soon.

PRESUBMIT.py
```diff
@@ -0,0 +1,278 @@
+# Copyright 2018 The Chromium Authors. All rights reserved.
```
dneto0: Does this run afoul of the copyright check?

PRESUBMIT.py
```diff
@@ -0,0 +1,278 @@
...
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
```
dneto0: The license-file is Apache2 in this repo.  So... what do we do here to make it right?

dj2:
I think this is ready to go. I didn't update the README yet, I'd like to land this, then I can verify the instructions will work correctly on mac/win/linux. Landing these changes shouldn't affect anything with the current README instructions.

PRESUBMIT.py
```diff
@@ -1,6 +1,16 @@
-# Copyright 2018 The Chromium Authors. All rights reserved.
-# Use of this source code is governed by a BSD-style license that can be
-# found in the LICENSE file.
+# Copyright (c) 2018 The Khronos Group Inc.
```
dneto0: This is fine.  But we now usually put Google LLC for new files written by Google.  SPRV-Tools has mixed authorship.

.clang-format
```diff
@@ -2,4 +2,5 @@
 Language: Cpp
 BasedOnStyle: Google
 DerivePointerAlignment: false
+SortIncludes: true
```
dneto0: Ooh.  This is aggressive.  But it passes the bots, so I'm good with it.

dneto0:
Rebased, squashed, and pushed into master as 185406439f

PR #1614, opened 2018-06-11
Combine test suites into single unit test suite

dj2:
This CL take the majority of the SPIRV-Tools test suites and combines them into
a single binary. This takes the number of build targets on a clean build from
~678 to ~489. The number of suites run by ctest goes from ~150 to 5.

This has the intersting effect that it is a lot easier to see how long all of
the SPIRV-Tools unit tests take (30sec) as they get run in the same test suite
now.


dj2:
What do folks think about this, good idea, bad idea?

dj2:
@s-perron pointed out this broke with compression off, that should now be fixed and it should work with and without compression enabled.

dneto0:
I should have looked again before starting on my own path...  See #1616 
I was less aggressive.   I kept a run of "ctest -j 20" to around 10s still on my machine.  Fast is better than slow.   :-)

dj2:
Yea, I went with a bigger hammer. There is also the question of link speed, which this maybe a bit faster as it will like 1 tests instead of a bunch. How many targets does your version have?

I'm curious if the -j will effect tests inside a suite or just suites run in parallel?

dj2:
Interestingly, this failed with a Timeout on appveyor (although I have no idea why as it was well under the allowed time.....So, maybe a good reason to go with the small version.

dj2:
Closed in favour of #1616

Issue #1615, opened 2018-06-11
Spurious AppVeyor failures


dneto0:
Example: https://ci.appveyor.com/project/Khronoswebmaster/spirv-tools/branch/master/job/8ucofc97wrgi9359 

I think we're running into a resource exhaustion condition.  

See @dj2 's note at https://github.com/KhronosGroup/SPIRV-Tools/pull/1604#issuecomment-395113175 
 where it might be disk space issue.

MSVC debug builds are rather large.

Let's fix this because it's getting in the way of the progress of the project.

dj2:
I believe this is now fixed by @dneto0 with PR #1616 so I'm going to close this issue.

PR #1616, opened 2018-06-11
Make fewer test executables

dneto0:
Try to reduce the amount of disk space used by especially by debug builds,
which may be contributing to AppVeyor failures.

Collapses tests in categories:
- validator
- loop optimizations
- dominator analysis
- linker

Contributes to #1615

### DIFF
--- a/test/val/CMakeLists.txt
+++ b/test/val/CMakeLists.txt
@@ -18,177 +18,54 @@ set(VAL_TEST_COMMON_SRCS
   ${CMAKE_CURRENT_SOURCE_DIR}/val_fixtures.h
 )
 
-
-add_spvtools_unittest(TARGET val_capability
-  SRCS val_capability_test.cpp
-       ${VAL_TEST_COMMON_SRCS}
-  LIBS ${SPIRV_TOOLS}
-)
-
-add_spvtools_unittest(TARGET val_cfg
-  SRCS val_cfg_test.cpp
-       ${VAL_TEST_COMMON_SRCS}
-  LIBS ${SPIRV_TOOLS}
-)
-
-add_spvtools_unittest(TARGET val_id
-  SRCS val_id_test.cpp
-       ${VAL_TEST_COMMON_SRCS}
-  LIBS ${SPIRV_TOOLS}
-)
-
-add_spvtools_unittest(TARGET val_layout
-  SRCS val_layout_test.cpp
-       ${VAL_TEST_COMMON_SRCS}
-  LIBS ${SPIRV_TOOLS}
-)
-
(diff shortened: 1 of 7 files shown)


dneto0:
When I run "ctest -j 20" I still get consistent total test times of roughly 10s.
(That's part of why I left validator limits test on its own.  It's a tall pole.)

alan-baker:
Basically the same as #1616?

dj2:
This is basically a small version of https://github.com/KhronosGroup/SPIRV-Tools/pull/1614.

This version maybe faster for the -j20 variant as I have 5 suites and this will have a bunch. The longest suite in my version is 30sec. I don't know if ctest -j20 will run tests in parallel or just suites?

dneto0:
>  I don't know if ctest -j20 will run tests in parallel or just suites? 

It runs suites in parallel.  That is, runs test executables in parallel.
(And yah, I didn't see your PR before starting on mine.  :-)

dj2:
I'm fine with the big or small version. The big version means you only have to remember on suite name to run, the small version is faster when run parallel. I'm curious what causes the slow down for the big version I can see it pausing every now and then so is it just a few really slow tests or the number of tests that causes the 30sec runtime?

test/val/CMakeLists.txt
```diff
@@ -18,177 +18,54 @@ set(VAL_TEST_COMMON_SRCS
...
-
-add_spvtools_unittest(TARGET val_primitives
-        SRCS val_primitives_test.cpp
-       ${VAL_TEST_COMMON_SRCS}
-  LIBS ${SPIRV_TOOLS}
-)
-
-add_spvtools_unittest(TARGET val_ext_inst
-	SRCS val_ext_inst_test.cpp
+add_spvtools_unittest(TARGET val_abcde
```
dj2: heh, these are terrible test names .... Is there some logical way to break this apart, like numerics, something, something?

dneto0: Lol.

The abcde says the test covers  .cpp source files starting from a through e.  I had hoped it was an obvious mapping.

dneto0:
> I can see it pausing every now and then so is it just a few really slow tests or the number of tests that causes the 30sec runtime?

Most tests are super fast.  A small handful of tests are slow because they test limits on things like structured nesting depth and similar combinatorial explosions.

I think there is a happy medium somewhere.   Anyway, it will be good data to let both MRs run their course.

dneto0:
So AppVeyor passed: Yay
Clang-format check failed:  Boo.

Rebased, and fixed formatting.

dj2:
Test names make me sad but lgtm regardless.

PR #1617, opened 2018-06-12
Fix buffer read overrun in linker

dneto0:
Fixes an ASAN failure.

Was occuring when generating the OpModuleProcessed instruction declaring
that this module was processed by the linker.


dneto0:
Rebased and pushed onto master as b49cbf09c

PR #1622, opened 2018-06-18
Validate Ids before DataRules.

dj2:
The DataRule validators call FindDefs with the assumption that they
definitions being looked at can be found. This may not be true if we
have not validated identifiers first.

This CL flips the IdPass and DataRulesPass to fix this issue.


dj2:
Done.

PR #1623, opened 2018-06-18
 Validate within function body when doing a FunctionCall.

dj2:
When validating a FunctionCall we can trigger an assert if we are not
currently within a function body. This CL adds verification that we are
within a function before attempting to add a function call.


dj2:
Is _LAYOUT the right error in this case?

PR #1625, opened 2018-06-19
Enable Kokoro buildbots.

ehsannas:


ehsannas:
PTAL @antiagainst @dneto0 
The MSCV-2017-Debug still fails. But it looks like the bots are running for PRs now. So If this change is merged, we should at least see all the bots passing except for 1.
I have filed an issue for MSVC-2017 version to be updated by Kokoro team. And I'll look into Docker.

antiagainst:
Maybe we can disable the failing job for now and re-enable it later?

kokoro/android/build.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+# Copyright (c) 2017 Google Inc.
```
antiagainst: 2018 Google LLC?

kokoro/android/build.sh
```diff
@@ -0,0 +1,51 @@
...
+# Display commands being run.
+set -x
+
+BUILD_ROOT=$PWD
+SRC=$PWD/github/SPIRV-Tools
+TARGET_ARCH="armeabi-v7a with NEON"
+export ANDROID_NDK=/opt/android-ndk-r15c
+
+# Get NINJA.
+wget -q https://github.com/ninja-build/ninja/releases/download/v1.7.2/ninja-linux.zip
```
antiagainst: Kokoro does not have ninja installed by default?

The latest is v1.8.2, which is released on 2017-09-12: https://github.com/ninja-build/ninja/releases/tag/v1.8.2

Also for other places.

kokoro/ndk-build/build.sh
```diff
@@ -0,0 +1,59 @@
...
+# Fail on any error.
+set -e
+# Display commands being run.
+set -x
+
+BUILD_ROOT=$PWD
+SRC=$PWD/github/SPIRV-Tools
+
+# Set the glslang repo URL.
+GLSLANG_REPO_URL="https://github.com/KhronosGroup/glslang.git"
```
antiagainst: We do not need Glslang for SPIRV-Tools?

kokoro/android/build.sh
ehsannas: Done

kokoro/ndk-build/build.sh
ehsannas: Oops. Copy-pasta from shaderc... Done

antiagainst:
+1. I'd suggest we disable MSCV-2017-Debug for now and get this submitted. How do you think @dneto0?

dneto0:
> +1. I'd suggest we disable MSCV-2017-Debug for now and get this submitted. How do you think @dneto0?

I'd still want coverage of that case.  Maybe keep the AppVeyor bot *only* for MSVC-2017-debug and enable Kokoro for everything else.

ehsannas:
Yeah we don't want to lose coverage. I am not removing the existing bots until the new ones are all stable and green.

PR #1626, opened 2018-06-19
Instruction lookup succeeds if it's enabled by a capability

### LINKED ISSUE #1624: Can't assemble OpShuffleGroupINTEL : directly guarded only by capabilities
The assembler (actually internal symbol lookup) fails to find a symbol when that symbol:

- Is "Reserved", i.e. not in any version of the SPIR-V core spec
- Is not directly enabled by an extension
- Is directly guarded by at least one capability.

(Unfortunately this also often the recommended way to enable new instructions and enums: Introduce a new capability guarded by the extension, and then enable individual features with that capability.)

This is the instruction case similar to #1596 which was for operand enums.

dneto0:
Also add a corresponding check for capabilities in the validator.

Update previously existing test cases where an instruction used to fail
assembling because of a version check, but now they succeed because the
instruction is also guarded by a capability.  Now it should assemble.
Add tests to ensure that capabilities are checked appropriately.

The explicitly reserved instructions OpImageSparseSampleProj*
now assemble, but they fail validation.

Fixes #1624

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -108,7 +108,7 @@ spv_result_t spvOpcodeTableNameLookup(spv_target_env env,
     // is indeed requested in the SPIR-V code; checking that should be
     // validator's work.
     if ((spvVersionForTargetEnv(env) >= entry.minVersion ||
-         entry.numExtensions > 0u) &&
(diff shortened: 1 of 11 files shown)


dneto0:
Rebased and pushed into master as 8d65c8967

PR #1630, opened 2018-06-20
Add initial WebGPU env validation

Kangz:
@dneto0 PTAL!

This adds support for the "webgpu0" = "WIP WebGPU" environment and adds an initial check that disallows usage of OpUndef (+ a test).

Also some minor cleanups / warning fixes.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -412,6 +412,7 @@ typedef enum {
   SPV_ENV_OPENCL_EMBEDDED_2_2,  // OpenCL Embedded Profile 2.2 latest revision.
   SPV_ENV_UNIVERSAL_1_3,  // SPIR-V 1.3 latest revision, no other restrictions.
   SPV_ENV_VULKAN_1_1,     // Vulkan 1.1 latest revision.
+  SPV_ENV_WEBGPU_0,       // Work in progress WebGPU 1.0.
 } spv_target_env;
 
 // SPIR-V Validator can be parameterized with the following Universal Limits.
--- a/source/validate_instruction.cpp
+++ b/source/validate_instruction.cpp
@@ -82,8 +82,7 @@ CapabilitySet EnablingCapabilitiesForOp(const ValidationState_t& state,
     case SpvOpGroupFMaxNonUniformAMD:
     case SpvOpGroupUMaxNonUniformAMD:
     case SpvOpGroupSMaxNonUniformAMD:
-      if (state.HasExtension(kSPV_AMD_shader_ballot))
-        return CapabilitySet();
+      if (state.HasExtension(kSPV_AMD_shader_ballot)) return CapabilitySet();
       break;
     default:
       break;
@@ -179,7 +178,7 @@ ExtensionSet RequiredExtensions(const ValidationState_t& state,
   return {};
 }
 
-// Return SPV_ERROR_INVALID_BINARY and emit a diagnostic if the instruction
+// Returns SPV_ERROR_INVALID_BINARY and emits a diagnostic if the instruction
 // is explicitly reserved in the SPIR-V core spec.  Otherwise return
 // SPV_SUCCESS.
 spv_result_t ReservedCheck(ValidationState_t& _,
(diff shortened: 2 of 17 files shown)


source/validate_instruction.cpp
```diff
@@ -205,6 +205,12 @@ spv_result_t ReservedCheck(ValidationState_t& _,
       return _.diag(SPV_ERROR_INVALID_BINARY)
              << "Invalid Opcode name 'Op" << inst_desc->name << "'";
     }
+    case SpvOpUndef:
```
Kangz: This function is called CheckReserved, should the name be updated to reflect that we also do env-related checks?

include/spirv-tools/libspirv.h
```diff
@@ -412,6 +412,7 @@ typedef enum {
   SPV_ENV_OPENCL_EMBEDDED_2_2,  // OpenCL Embedded Profile 2.2 latest revision.
   SPV_ENV_UNIVERSAL_1_3,  // SPIR-V 1.3 latest revision, no other restrictions.
   SPV_ENV_VULKAN_1_1,     // Vulkan 1.1 latest revision.
+  SPV_ENV_WEBGPU_0, // Work in progress WebGPU 1.0.
```
dneto0: I expect clang-format will complain.

source/validate_instruction.cpp
dneto0: Also the comment on the method would need updating.

The alternative is to make a whole new function.

test/val/val_decoration_test.cpp
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Common validation fixtures for unit tests
+// Validation tests for decorations
```
dneto0: Yay. Thanks.

test/val/val_fixtures.h
```diff
@@ -43,8 +45,7 @@ class ValidateBase : public ::testing::Test,
   // This function overwrites the word at the given index with a new word.
   void OverwriteAssembledBinary(uint32_t index, uint32_t word);
 
-  // Performs validation on the SPIR-V code and compares the result of the
-  // spvValidate function
+  // Performs validation on the SPIR-V code.
```
dneto0: Thank you for fixing this.

test/val/val_webgpu_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+    %1 = OpUndef %float
+)";
+
+  CompileSuccessfully(spirv);
+
+  // Control case: this is allowed in SPIR-V 1.3
+  EXPECT_EQ(SPV_SUCCESS, ValidateInstructions(SPV_ENV_UNIVERSAL_1_3));
+
+  // Control case: OpUndef is disallowed in the WebGPU env
+  EXPECT_EQ(SPV_ERROR_INVALID_BINARY, ValidateInstructions(SPV_ENV_WEBGPU_0));
```
dneto0: You should add a check for the result of the diagnostic string.
E.g. see uses of HasSubstr in val_capability_test.cpp.  For example the checks I just added for SubgroupShuffleINTEL

include/spirv-tools/libspirv.h
Kangz: Done

test/val/val_decoration_test.cpp
Kangz: Ack

test/val/val_fixtures.h
Kangz: Ack

test/val/val_webgpu_test.cpp
Kangz: Done

source/validate_instruction.cpp
Kangz: Chose to add a new EnvironmentCheck function.

Kangz:
PTAL again, addressed comments, some more additional commenting of existing code snuck in, and an additional commit to reduce the visibility of symbols in validate_instruction.cpp

Kangz:
Thanks for the review, the PR passed CI.

test/val/val_version_test.cpp
```diff
@@ -112,6 +114,7 @@ INSTANTIATE_TEST_CASE_P(Universal, ValidateVersion,
     make_tuple(SPV_ENV_UNIVERSAL_1_1, SPV_ENV_OPENGL_4_2,    vulkan_spirv, false),
     make_tuple(SPV_ENV_UNIVERSAL_1_1, SPV_ENV_OPENGL_4_3,    vulkan_spirv, false),
     make_tuple(SPV_ENV_UNIVERSAL_1_1, SPV_ENV_OPENGL_4_5,    vulkan_spirv, false),
+    make_tuple(SPV_ENV_UNIVERSAL_1_0, SPV_ENV_WEBGPU_0,      vulkan_spirv, true),
```
alan-baker: This case is duplicated 4 times. I think you meant to update the universal environment comparison.

Kangz: Done

alan-baker:
This failure seems unrelated. Filed #1632 for it.

dneto0:
Rebased, squashed, and pushed into master as  ba602c90

Issue #1632, opened 2018-06-21
test_val_limits near timeout on Windows debug build


alan-baker:
https://ci.appveyor.com/project/Khronoswebmaster/spirv-tools/build/3440/job/gtvwpefl1r8rpgqg

The NumLocalVarsGood and NumLocalVarsBad tests run half a million variables. This greatly increases test runtime without improving the overall quality of the test.

We should artificially limit the variable number in the test via the existing option.

PR #1633, opened 2018-06-21
Validator test: artificially lower limit on local var limit

dneto0:
We need this to reduce the test time on Visual Studio debug builds.
AppVeyor times out at 300s for a process.

Artificially lower the local variable limit check for testing purposes.
We don't get much value from using the full 500K+ variables.
Use a limit of 5000 instead.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1632


alan-baker:
Rebased and committed as 4d99fcba9d9915eb3d8a923f2906aa66ef354dac

Issue #1638, opened 2018-06-25
validator: implement review feedback on Block / BufferBlock layout checks


dneto0:
See comments at https://github.com/KhronosGroup/SPIRV-Tools/pull/1568#pullrequestreview-131647492

Issue #1637, opened 2018-06-25
validator: Check matrix stride on struct member that is array of (array of ...) matrix


dneto0:
Example vertex shader in GLSL:

```
#version 450

layout (set=1, binding=2) uniform S { mat4x4 arr[2][3]; } B;

void main() {
  vec4 v = B.arr[1][2][1];
}
```
Here's the SPIR-V:
```
; SPIR-V
; Version: 1.0
; Generator: Google Shaderc over Glslang; 7
; Bound: 26
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main"
               OpSource GLSL 450
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %main "main"
               OpName %v "v"
               OpName %S "S"
               OpMemberName %S 0 "arr"
               OpName %B "B"
               OpDecorate %_arr_mat4v4float_uint_3 ArrayStride 64
               OpDecorate %_arr__arr_mat4v4float_uint_3_uint_2 ArrayStride 192
               OpMemberDecorate %S 0 ColMajor
               OpMemberDecorate %S 0 Offset 0
               OpMemberDecorate %S 0 MatrixStride 16 ;;;;; This is the surprising part
               OpDecorate %S Block
               OpDecorate %B DescriptorSet 1
               OpDecorate %B Binding 2
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
    %v4float = OpTypeVector %float 4
%_ptr_Function_v4float = OpTypePointer Function %v4float
%mat4v4float = OpTypeMatrix %v4float 4
       %uint = OpTypeInt 32 0
     %uint_3 = OpConstant %uint 3
%_arr_mat4v4float_uint_3 = OpTypeArray %mat4v4float %uint_3
     %uint_2 = OpConstant %uint 2
%_arr__arr_mat4v4float_uint_3_uint_2 = OpTypeArray %_arr_mat4v4float_uint_3 %uint_2
          %S = OpTypeStruct %_arr__arr_mat4v4float_uint_3_uint_2
%_ptr_Uniform_S = OpTypePointer Uniform %S
          %B = OpVariable %_ptr_Uniform_S Uniform
        %int = OpTypeInt 32 1
[...]

Issue #1606, opened 2018-06-06
Check layout of StorageBuffer with Block decoration


alan-baker:
Follow on to #937 

Layout rules for StorageBuffer with Block decoration should be the same as Uniform with BufferBlock decoration.

asuonpaa:
I'll grab this as soon as the original changes have landed.

PR #1639, opened 2018-06-25
Fix Pr1568: use integer arithmetic instead of floor

dneto0:
pr1568 didn't compile on my machine because the "floor" function was not declared ( missing cmath header?)

I prefer to use integer arithmetic anyway, and this PR is:
- squashed commit with all content from #1568 
- one more commit to change the use of floor to use integer arithmetic instead.

I still have other followups that I'd like to see, filed as #1637 and #1638


dneto0:
FYI: There is also #1606 to ensure StorageBuffer with Block decoration is treated the same as Uniform with BufferBlock decoration.

Issue #1372, opened 2018-03-03
The spirv-opt remove-duplicates pass makes a seemingly erroneous transformation of my shader


StrayLightning:
I have run into a problem with spirv-opt since the release of Vulkan SDK 1.0.68.0.  The remove-duplicates pass seems to be clobbering my shader.  Prior to that release, things were fine.  This may well be something that I am doing wrong, but I have been unable to determine what thus far.

spirv-opt from Vulkan SDK 1.0.68.0 self-identifies as:

    SPIRV-Tools v2018.1-dev v2018.0-6-g9e19fc0

The original SPIR-V is constructed by glslangValidator from the same SDK.

Some relevant bits from my shader:

	layout(set = 0, binding = 5, std140) uniform atmosphere_fragment_uniforms {
		float sun_radius;
	};

	layout(set = 1, binding = 12, std140) uniform planet_dynamic_uniforms {
		float lod_factor;
	};

Before any optimisation the disassembly shows the following section:

    OpName %planet_dynamic_uniforms "planet_dynamic_uniforms"
    OpMemberName %planet_dynamic_uniforms 0 "lod_factor"

After running just the remove-duplicates pass this has changed to:

    OpName %atmosphere_fragment_uniforms "planet_dynamic_uniforms"
    OpMemberName %atmosphere_fragment_uniforms 0 "lod_factor"

There are other changes, but off the bat this seems odd.

Running the optimiser with every -O pass except for remove-duplicates generates SPIR-V that appears to work fine.

I tried using spirv-opt from the CI build from commit 6cd6e5ebef0623a863ca2d8152edda75dd5001df, but with no change in behaviour.

The following is a cut-down that generates the odd difference when running just remove-duplicates:

	#version 450 core

	layout(set = 1, binding = 12, std140) uniform planet_dynamic_uniforms {
		float lod_factor;
	};

	layout(set = 0, binding = 5, std140) uniform atmosphere_fragment_uniforms {
		float sun_radius;
	};

	layout(location = 0) out vec4 out_colour;

	void main() {
		out_colour = vec4(1.0);
	}

To demonstrate the difference:

	glslangValidator -V -o cutdown.spv cutdown.frag
	spirv-opt -o cutdown.opt.spv --remove-duplicates cutdown.spv
	spirv-dis -o cutdown.s cutdown.spv
	spirv-dis -o cutdown.opt.s cutdown.opt.spv
	fc /l cutdown.s cutdown.opt.s

[...]

PR #1640, opened 2018-06-26
Don't merge types of resources

s-perron:
When doing reflection users care about the names of the variable, the
    name of the type, and the name of the members.  Remove duplicates breaks
    this because it removes the names one of the types when merging.
    
    To fix this we have to keep the different types around for each
    resource.  This commit adds code to remove duplicates to look for the
    types uses to describe resources, and make sure they do not get merged.
    
    This could have a negative effect on compile time, but it was not
    expected to be much.
    
    Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1372.


dneto0:
Rebased, squashed, and pushed into master as f393b0e480148

Issue #1631, opened 2018-06-21
Eliminate storing undef values into interface output variables


antiagainst:
For the following HLSL source code:

```hlsl
void MainVp(in float3 used_in_var : USED_IN_VAR,
	    in float3 unused_in_var : UNUSED_IN_VAR,
	    out float3 used_out_var : USED_OUT_VAR,
	    out float3 unused_out_var : UNUSED_OUT_VAR)
{
	used_out_var = used_in_var;
}
```

We are optimizing it into the following right now:

```llvm
         %19 = OpUndef %v3float
     %MainVp = OpFunction %void None %9
         %20 = OpLabel
         %21 = OpLoad %v3float %in_var_USED_IN_VAR
               OpStore %out_var_USED_OUT_VAR %21
               OpStore %out_var_UNUSED_OUT_VAR %19
```
Note that we are writing an undef value into the essentially unused output variable. This causes problems for certain reflection workflows. I think it's reasonable to avoid the store if the value to be stored is an undef value.

Initially reported by @Gaspard-- as https://github.com/Microsoft/DirectXShaderCompiler/issues/1362. See the original issue for background and more discussion.

Gaspard--:
While testing a workaround, I arrived to some interresting generated bytecode which may be related.
I wasn't able to build a POC or narrow down a cause, and sadly I can't share the source due to legal reasons.
Here's snippets of the bytecode which caught my attention:
```spirv
         %83 = OpUndef %v4float
         %84 = OpUndef %v3float
[...]
        %268 = OpCompositeConstruct %v3bool %196 %196 %196
        %269 = OpSelect %v3float %268 %84 %84
        %270 = OpCompositeConstruct %v4bool %196 %196 %196 %196
        %271 = OpSelect %v4float %270 %83 %83
[...]
               OpStore %out_var_LEGACY_TEXCOORD %271
               OpStore %out_var_WSPOSITION %269
```
The `OpSelect`s seem useless (but I may not know enough about spirv to know why it wouldn't be).
This seems like an extension of the case above, so I thought it was worth mentioning.

s-perron:
Okay, so my change has taken care of part of this.  However, the OpSelect not being folded away is strange.  We have code that is suppose to be removing thing.  Make sure you are running the simplification pass, which is the last thing run when using -O.

Let me know how things look.

Gaspard--:
[...]

PR #1641, opened 2018-06-26
Remove stores of undef.

s-perron:
When storing an undef, any value is valid, including the one already in
that memory location.  So we can avoid the store.

Part of https://github.com/KhronosGroup/SPIRV-Tools/issues/1631.

### DIFF
--- a/source/opt/folding_rules.cpp
+++ b/source/opt/folding_rules.cpp
@@ -31,6 +31,7 @@ const uint32_t kExtInstInstructionInIdx = 1;
 const uint32_t kFMixXIdInIdx = 2;
 const uint32_t kFMixYIdInIdx = 3;
 const uint32_t kFMixAIdInIdx = 4;
+const uint32_t kStoreObjectInIdx = 1;
 
 // Returns the element width of |type|.
 uint32_t ElementWidth(const analysis::Type* type) {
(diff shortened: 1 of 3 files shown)


source/opt/folding_rules.cpp
```diff
@@ -2055,6 +2074,8 @@ spvtools::opt::FoldingRules::FoldingRules() {
 
   rules_[SpvOpSelect].push_back(RedundantSelect());
 
+  rules_[SpvOpStore].push_back(StoringUndef());
```
alan-baker: What OpCopyMemory, OpCopyMemorySized, OpImageWrite and atomic operations that can write?

s-perron: We cannot remove `OpCopyMemory*` instructions because the source is a pointer, and the values being copied is what is pointed to by the source pointer.  The pointer should not be OpUndef, and we cannot tell if the memory the source pointer points to is undefined.

We should be able to remove OpImageWrite instructions, and atomic instruction.  I'll look into adding those.

s-perron: Can't remove atomic stores.  OpImageWrite's will not be removed because the check for volatile was not very clear.

Issue #1645, opened 2018-06-27
Tests fail with Python3


pow2clk:
When the default python for a system is python3, the ctests involving check_symbol_exports.py produce python errors due to a change in python's willingness to try treating byte data like a string when the programmer asks and hope for the best between python2 and python3:
```

 ctest --output-on-failure --tests-regex spirv-tools-symbol-exports
Test project /home/sadangel/opc/SPIRV-Tools/build
    Start 4: spirv-tools-symbol-exports-SPIRV-Tools
1/4 Test #4: spirv-tools-symbol-exports-SPIRV-Tools ..........***Failed    0.21 sec
Traceback (most recent call last):
  File "/home/sadangel/opc/SPIRV-Tools/utils/check_symbol_exports.py", line 92, in <module>
    main()
  File "/home/sadangel/opc/SPIRV-Tools/utils/check_symbol_exports.py", line 84, in main
    status = check_library(args.library)
  File "/home/sadangel/opc/SPIRV-Tools/utils/check_symbol_exports.py", line 60, in check_library
    for line in command_output(['objdump', '-t', library], '.').split('\n'):
TypeError: a bytes-like object is required, not 'str'

    Start 5: spirv-tools-symbol-exports-SPIRV-Tools-shared
2/4 Test #5: spirv-tools-symbol-exports-SPIRV-Tools-shared ...***Failed    0.09 sec
Traceback (most recent call last):
  File "/home/sadangel/opc/SPIRV-Tools/utils/check_symbol_exports.py", line 92, in <module>
    main()
  File "/home/sadangel/opc/SPIRV-Tools/utils/check_symbol_exports.py", line 84, in main
    status = check_library(args.library)
  File "/home/sadangel/opc/SPIRV-Tools/utils/check_symbol_exports.py", line 60, in check_library
    for line in command_output(['objdump', '-t', library], '.').split('\n'):
TypeError: a bytes-like object is required, not 'str'

    Start 6: spirv-tools-symbol-exports-SPIRV-Tools-opt
3/4 Test #6: spirv-tools-symbol-exports-SPIRV-Tools-opt ......***Failed    0.78 sec
Traceback (most recent call last):
  File "/home/sadangel/opc/SPIRV-Tools/utils/check_symbol_exports.py", line 92, in <module>
    main()
  File "/home/sadangel/opc/SPIRV-Tools/utils/check_symbol_exports.py", line 84, in main
    status = check_library(args.library)
[...]

PR #1646, opened 2018-06-27
check_symbol_exports on Python3, Mac

pow2clk:
subprocess.Popen returns byte data by default. Python2 was happy
to try to execute string operations on such data and hope for the
best, but python3 is more persnickety. Luckily, there's a simple
way to indicate to the Popen class that text data is wanted that
benefits the script. Just specifying universal_newlines will cause
the returned data to be text and also convert any system-specific
newlines to '\n' which the script relies on anyway.

Enabled on Mac as an incidental change after confirming that the
script works there just as well as it does on Linux.

It probably works on FreeBSD too, but I retired my BSD system years
ago. So I have no way to check.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1645

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,8 +168,7 @@ endif()
 
 find_host_package(PythonInterp)
 
-if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux" OR "${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin"
-                          OR "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows)
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux" OR "${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
   macro(spvtools_check_symbol_exports TARGET)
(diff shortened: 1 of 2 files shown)


pow2clk:
Forgot to reference https://github.com/KhronosGroup/SPIRV-Tools/issues/1645 . I fixed it on the web interface. Maybe that's good enough?

dneto0:
Are symbols on Windows really the same format as the Itanium mangling expected on Linux?

The check-symbol-exports was a spot check for namespace leakage.  I guess there was a possible hole on the Windows side, but I viewed it as low risk.

I'm happy to merge this if the bots pass.

pow2clk:
I was surprised about Windows too. I confess though that all I did was plug this change in and run. Everything came back happy and the tests did run. So I guess so?

pow2clk:
Turns out the Windows passing was a mirage. When the script finds it's not on a posix system, it pretends to pass and moves on. So patronizing!

dneto0:
Rebased, squashed, and pushed into master as 878b3b400b2f

PR #1650, opened 2018-06-27
GLSL.std.450 Refract Eta can be any float scalar

dneto0:
This is a decision from Khronos-internal SPIR-V spec issue 337.


dneto0:
Rebased and pushed into master as b67beca723

Issue #1648, opened 2018-06-27
Assert in Compact IDs optimization pass


pow2clk:
This was originally found in dxc and filed as https://github.com/google/DirectXShaderCompiler/issues/264 . Since I've confirmed it's entirely constrained to SPIRV-Tools, I'm opening this issue and closing that one. I'll copy an edited version of the explanation from there.

a1f9e1342e77e47251e2f001d95f8af8a523745a recently altered which Analyses were preserved for the compact IDs pass. Instead of preserving nothing, it preserved a list of things including `ir::IRContext::kAnalysisCFG` . This last is a problem because as potentially changes the result IDs, it can change the basic block ID predecessors. `IRContext::CheckCFG()` uses a stale basic block ID to index predecessors and fires an assert. This function is only called inside an assert itself, so this is entirely a debug issue as manifest here. Though I wouldn't say for certain it doesn't have other effects.

The way to observe this issue is to build spirv-opt with debug symbols and run it on the .spv file included in the attached zip since github won't let me upload mysterious file extensions, but zipped mysterious files are a-okay. Any unoptimized spir-v file will probably work.
[ppass.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/2142599/ppass.zip)

Simply running `spirv-opt  --compact-ids ppass.mO.spv -o ppass.spv` is insufficient because the CFG has to be valid at the start of the compact IDs pass or it will get validated as soon as `cfg()` is called in `CheckCFG()` anyway. So you have to do some other optimization pass first. I did: `spirv-opt  --redundancy-elimination --compact-ids ppass.mO.spv -o ppass.spv`

I have a fix for the problem. I hoped to have a test to prevent regressions, but I was looking over the existing compact_ids_test and I'm unsure that the issue can be reproduced unless I can be sure that another pass is run before it. Maybe I can work it into the pass_manager_test. I'm open to suggestions though.

s-perron:
Here is a snippet you can use when writing a test:

```
  BuildModule(SPV_ENV_UNIVERSAL_1_1, nullptr, test, SPV_TEXT_TO_BINARY_OPTION_PRESERVE_NUMERIC_IDS);
  ASSERT_NE(nullptr, context);
  auto pass = CreateCompactIdsPass();

  opt::CompactIdsPass compact_id_pass;
  context->BuildInvalidAnalyses(compact_id_pass.GetPreservedAnalyses());
  compact_id_pass.Run(context.get());
[...]

PR #1651, opened 2018-06-28
Add test for CFG alteration by compact IDs

pow2clk:
The Compact IDs pass can corrupt the CFG, but first the CFG has to
be setup. To do this, a test that builds the CFG, then performs the
compact IDs pass, then checks context consistency.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1648

### DIFF
--- a/test/opt/compact_ids_test.cpp
+++ b/test/opt/compact_ids_test.cpp
@@ -192,4 +192,84 @@ OpFunctionEnd
   EXPECT_THAT(disassembly, ::testing::Eq(expected));
 }
 
+// Test context consistency check after invalidating
+// CFG and others by compact IDs Pass.
+// Uses a GLSL shader with named labels for variety
+TEST(CompactIds, ConsistentCheck) {
+  const std::string input(R"(OpCapability Shader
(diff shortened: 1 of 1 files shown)


test/opt/compact_ids_test.cpp
```diff
@@ -192,4 +192,84 @@ OpFunctionEnd
...
+%_ptr_Output_v4float = OpTypePointer Output %v4float
+%in_var_A = OpVariable %_ptr_Input_v4float Input
+%out_var_SV_TARGET = OpVariable %_ptr_Output_v4float Output
+%main = OpFunction %void None %3
+%5 = OpLabel
+%12 = OpLoad %v4float %in_var_A
+%23 = OpVectorShuffle %v4float %12 %12 0 0 0 1
+OpStore %out_var_SV_TARGET %23
+OpReturn
+OpFunctionEnd
```
pow2clk: I realize that any spirv with uncompacted IDs would have worked, but I thought I'd use this because it was unlike that used for many of the existing tests.

pow2clk:
@s-perron will probably want to take a look at this.

PR #1654, opened 2018-06-29
Add a check for invalid exits from case construct.

### LINKED ISSUE #1618: Check in structured cfg that there are no branches into the middle of a case
#1281 added a general check, but the validator does not create case constructs. This should be fixed after #491 is fixed.

alan-baker:
Fixes #1618.

Adds a check that validates acceptable exits from case constructs. Case
constructs may only exit to another case construct, the corresponding
merge, an outer loop continue or outer loop merge.

I say this fixes #1618 because any reachable block that branches into the middle of a case construct would violate the dominance and split the calculated blocks of the case construct (see InvalidCaseExit test).

### DIFF
--- a/source/validate_cfg.cpp
+++ b/source/validate_cfg.cpp
@@ -173,21 +173,23 @@ string ConstructErrorString(const Construct& construct,
 // |case_fall_through|. Returns SPV_ERROR_INVALID_CFG if the case construct
 // headed by |target_block| branches to multiple case constructs.
 spv_result_t FindCaseFallThrough(
-    const ValidationState_t& _, const BasicBlock* target_block,
+    const ValidationState_t& _, BasicBlock* target_block,
     uint32_t* case_fall_through, const BasicBlock* merge,
-    const std::unordered_set<uint32_t>& case_targets) {
-  std::vector<const BasicBlock*> stack;
+    const std::unordered_set<uint32_t>& case_targets, Function* function) {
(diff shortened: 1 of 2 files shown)


source/validate_cfg.cpp
```diff
@@ -173,21 +173,23 @@ string ConstructErrorString(const Construct& construct,
...
-    const ValidationState_t& _, const BasicBlock* target_block,
+    const ValidationState_t& _, BasicBlock* target_block,
     uint32_t* case_fall_through, const BasicBlock* merge,
-    const std::unordered_set<uint32_t>& case_targets) {
-  std::vector<const BasicBlock*> stack;
+    const std::unordered_set<uint32_t>& case_targets, Function* function) {
+  std::vector<BasicBlock*> stack;
   stack.push_back(target_block);
   std::unordered_set<const BasicBlock*> visited;
+  bool target_reachable = target_block->reachable();
```
dneto0: Better for this to be const.  It took me a few seconds to see it never changed since target_block never changes.

source/validate_cfg.cpp
```diff
@@ -173,21 +173,23 @@ string ConstructErrorString(const Construct& construct,
...
+  int target_depth = function->GetBlockDepth(target_block);
```
dneto0: Should be const.

dneto0:
Rebased and pushed into master as c460f44fb

PR #1656, opened 2018-06-29
Revert change and stop running remove duplicates

### LINKED ISSUE #1655: Invalid types used after remove duplicates.
This is caused by my change for #1372.

An example is:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %1 "main"
               OpSource HLSL 600
               OpName %PositionAdjust "PositionAdjust"
               OpMemberName %PositionAdjust 0 "XAdjust"
               OpName %NormalAdjust "NormalAdjust"
               OpMemberName %NormalAdjust 0 "XDir"
               OpName %Constants "Constants"
               OpMemberDecorate %PositionAdjust 0 Offset 0
               OpMemberDecorate %NormalAdjust 0 Offset 0
               OpDecorate %Constants DescriptorSet 0
               OpDecorate %Constants Binding 0
               OpDecorate %5 DescriptorSet 1
               OpDecorate %5 Binding 0
      %float = OpTypeFloat 32
    %v3float = OpTypeVector %float 3
%PositionAdjust = OpTypeStruct %v3float
%NormalAdjust = OpTypeStruct %v3float
%_ptr_Uniform_NormalAdjust = OpTypePointer Uniform %NormalAdjust
%_ptr_Uniform_PositionAdjust = OpTypePointer Uniform %PositionAdjust
 %_struct_10 = OpTypeStruct %NormalAdjust
%_ptr_Uniform__struct_10 = OpTypePointer Uniform %_struct_10
          %5 = OpVariable %_ptr_Uniform_PositionAdjust Uniform
  %Constants = OpVariable %_ptr_Uniform__struct_10 Uniform
       %void = OpTypeVoid
         %13 = OpTypeFunction %void
       %uint = OpTypeInt 32 0
     %uint_0 = OpConstant %uint 0
          %1 = OpFunction %void None %13
         %16 = OpLabel
         %17 = OpAccessChain %_ptr_Uniform_NormalAdjust %Constants %uint_0
               OpReturn
               OpFunctionEnd
```

`%_ptr_Uniform_PositionAdjust` is used in a resource (`%5`), but `%_ptr_Uniform_PositionAdjust` is not.  No other replacements are allowed.  This causes a type mismatch on the OpAccessChain.

s-perron:
Fixes #1655.

The change for commit 455186d5ab037c0a659cb50301c9f57ef1313cc6, was not correct, and must expanded to handle more cases or reverted.  The value in running remove duplicates was compile time and code size.  Turning it off does not effect compile time.  So I chose to revert my change and turn off remove duplicates in the -O and -Os paths.  The extra complication of the code is not worth the gains we are not seeing.

### DIFF
--- a/test/opt/pass_remove_duplicates_test.cpp
+++ b/test/opt/pass_remove_duplicates_test.cpp
@@ -307,6 +307,9 @@ OpGroupDecorate %2 %4
   EXPECT_EQ(GetErrorMessage(), "");
 }
 
+// Test what happens when a type is a resource type.  For now we are merging
+// them, but, if we want to merge types and make reflection work (issue #1372),
+// we will not be able to merge %2 and %3 below.
 TEST_F(RemoveDuplicatesTest, DontMergeNestedResourceTypes) {
   const std::string spirv = R"(OpCapability Shader
 OpMemoryModel Logical GLSL450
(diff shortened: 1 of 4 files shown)


s-perron:
@greg-lunarg Can you check how this affects code size on the shaders that are important to you?  I did not check size because we are less concerned with it.

test/opt/pass_remove_duplicates_test.cpp
```diff
@@ -307,6 +307,9 @@ OpGroupDecorate %2 %4
   EXPECT_EQ(GetErrorMessage(), "");
 }
 
+// Test what happens when a type is a resource type.  For now we are merging
```
alan-baker: If you think its easier, consider just disabling the older version of the test.

s-perron: I had thought of that, but I believe the test could get lost that way.  This way if some tries to do something similar to what I did, this test will start to fail, and make it obvious which cases they need to handle.

s-perron:
Time sensitive, so I'll merge.  @greg-lunarg if there is an increase in size, let me know and we can figure out how to deal with it.

PR #1660, opened 2018-07-03
Produce better error diagnostics in the CFG validation.

### LINKED ISSUE #1657: Confusing assembly line shown for some validator errors
Some examples from the ValidateCFG test suite:
```
[ RUN      ] ValidateCFG.MultipleFallThroughBlocks
error: 22: Case construct that targets 10 has branches to multiple other case construct targets 12 and 11
  OpFunctionEnd

[       OK ] ValidateCFG.MultipleFallThroughBlocks (1 ms)
[ RUN      ] ValidateCFG.MultipleFallThroughToDefault
error: 22: Multiple case constructs have branches to the case construct that targets 10
  OpFunctionEnd

[       OK ] ValidateCFG.MultipleFallThroughToDefault (0 ms)
[ RUN      ] ValidateCFG.MultipleFallThroughToNonDefault
error: 22: Multiple case constructs have branches to the case construct that targets 12
  OpFunctionEnd

[       OK ] ValidateCFG.MultipleFallThroughToNonDefault (1 ms)
[ RUN      ] ValidateCFG.DuplicateTargetWithFallThrough
[       OK ] ValidateCFG.DuplicateTargetWithFallThrough (0 ms)
[ RUN      ] ValidateCFG.WrongOperandList
error: 22: Case construct that targets 12 has branches to the case construct that targets 11, but does not immediately precede it in the OpSwitch's target list
  OpFunctionEnd

[       OK ] ValidateCFG.WrongOperandList (0 ms)
[ RUN      ] ValidateCFG.WrongOperandListThroughDefault
error: 22: Case construct that targets 12 has branches to the case construct that targets 11, but does not immediately precede it in the OpSwitch's target list
  OpFunctionEnd

[       OK ] ValidateCFG.WrongOperandListThroughDefault (1 ms)
[ RUN      ] ValidateCFG.WrongOperandListNotLast
error: 24: Case construct that targets 12 has branches to the case construct that targets 11, but does not immediately precede it in the OpSwitch's target list
  OpFunctionEnd

[       OK ] ValidateCFG.WrongOperandListNotLast (0 ms)
[ RUN      ] ValidateCFG.InvalidCaseExit
error: 20: Case construct that targets 8 has invalid branch to block 10 (not another case construct, corresponding merge, outer loop merge or outer loop continue)
  OpFunctionEnd

[       OK ] ValidateCFG.InvalidCaseExit (1 ms)
```
The errors all show OpFunctionEnd as the associated disassembly, but that doesn't make sense for these messages. The CFG checks occur after a full parse of the binary so it may be that they are relying on a stale instruction count. I don't think some of these errors should even have an associated line for them as they require more context than a single line.

dj2:
This CL fixes up several issues with the diagnostic error line output
in the CFG validation code. For the cases where we can determine a
better line it has been output. For other cases, we removed the
diagnostic line and the error line number from the results.

Fixes #1657

### DIFF
--- a/source/val/basic_block.h
+++ b/source/val/basic_block.h
@@ -109,11 +109,11 @@ class BasicBlock {
   /// Ends the block without a successor
   void RegisterBranchInstruction(SpvOp branch_instruction);
 
-  /// Returns the initiator instruction for the block.
-  const Instruction* initiator() const { return initiator_; }
+  /// Returns the label instruction for the block, or nullptr if not set.
+  const Instruction* label() const { return label_; }
 
-  //// Registers the initiator instruction for the block.
-  void set_initiator(const Instruction* t) { initiator_ = t; }
+  //// Registers the label instruction for the block.
+  void set_label(const Instruction* t) { label_ = t; }
 
   /// Registers the terminator instruction for the block.
   void set_terminator(const Instruction* t) { terminator_ = t; }
@@ -224,8 +224,8 @@ class BasicBlock {
   /// True if the block is reachable in the CFG
   bool reachable_;
 
-  /// Initiator of this block.
-  const Instruction* initiator_;
+  /// label of this block, if any.
+  const Instruction* label_;
 
   /// Terminator of this block.
   const Instruction* terminator_;
--- a/source/val/basic_block.h
+++ b/source/val/basic_block.h
@@ -109,6 +109,12 @@ class BasicBlock {
   /// Ends the block without a successor
   void RegisterBranchInstruction(SpvOp branch_instruction);
 
+  /// Returns the initiator instruction for the block.
+  const Instruction* initiator() const { return initiator_; }
+
+  //// Registers the initiator instruction for the block.
+  void set_initiator(const Instruction* t) { initiator_ = t; }
+
   /// Registers the terminator instruction for the block.
   void set_terminator(const Instruction* t) { terminator_ = t; }
 
@@ -217,6 +223,9 @@ class BasicBlock {
 
   /// True if the block is reachable in the CFG
   bool reachable_;
+
+  /// Initiator of this block.
+  const Instruction* initiator_;
 
   /// Terminator of this block.
   const Instruction* terminator_;
--- a/source/validate.cpp
+++ b/source/validate.cpp
@@ -39,6 +39,7 @@
 #include "spirv_validator_options.h"
 #include "val/construct.h"
 #include "val/function.h"
+#include "val/instruction.h"
 #include "val/validation_state.h"
 
 using std::function;
@@ -53,6 +54,7 @@ using libspirv::CfgPass;
 using libspirv::DataRulesPass;
 using libspirv::Extension;
 using libspirv::IdPass;
+using libspirv::Instruction;
 using libspirv::InstructionPass;
 using libspirv::LiteralsPass;
 using libspirv::ModuleLayoutPass;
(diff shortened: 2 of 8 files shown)


source/validate.cpp
```diff
@@ -177,9 +179,12 @@ spv_result_t ProcessInstruction(void* user_data,
   // The IdPass check registers instructions and, therefore, must be called
   // before any instruction lookups are performed.
   if (auto error = IdPass(_, inst)) return error;
+
+  const Instruction* instruction = &(_.ordered_instructions().back());
+
   if (auto error = DataRulesPass(_, inst)) return error;
   if (auto error = ModuleLayoutPass(_, inst)) return error;
-  if (auto error = CfgPass(_, inst)) return error;
+  if (auto error = CfgPass(_, instruction)) return error;
```
dj2: I think this can be propagated through the other passes as well and then we always provide the higher level Instruction instead of the spv_parsed_instruction_t.

I didn't do it here to keep the change smaller.

source/validate_cfg.cpp
```diff
@@ -537,39 +543,40 @@ spv_result_t PerformCfgChecks(ValidationState_t& _) {
...
-  SpvOp opcode = static_cast<SpvOp>(inst->opcode);
+spv_result_t CfgPass(ValidationState_t& _, const Instruction* inst) {
+  SpvOp opcode = static_cast<SpvOp>(inst->opcode());
   switch (opcode) {
     case SpvOpLabel:
-      if (auto error = _.current_function().RegisterBlock(inst->result_id))
+      if (auto error = _.current_function().RegisterBlock(inst->id()))
         return error;
+
+      _.current_function().current_block()->set_initiator(inst);
```
alan-baker: This creates an odd disagreement in the validator. Due to the order of passes, OpLabel is *not* part of the basic block it starts. Can you add a TODO to fix that and file an issue?

source/val/basic_block.h
```diff
@@ -109,6 +109,12 @@ class BasicBlock {
   /// Ends the block without a successor
   void RegisterBranchInstruction(SpvOp branch_instruction);
 
+  /// Returns the initiator instruction for the block.
+  const Instruction* initiator() const { return initiator_; }
```
alan-baker: I think I'd prefer to call this label(). That is the only valid instruction to start a basic block.

source/validate_cfg.cpp
```diff
@@ -290,13 +295,12 @@ spv_result_t StructuredSwitchChecks(const ValidationState_t& _,
         // must immediately precede T2 in the list of OpSwitch Target operands.
         if ((switch_inst->operands().size() < j + 2) ||
             (case_fall_through != switch_inst->GetOperandAs<uint32_t>(j + 2))) {
-          return _.diag(SPV_ERROR_INVALID_CFG)
+          return _.diag(SPV_ERROR_INVALID_CFG, -1)
```
alan-baker: The OpSwitch if its accessible would probably be the right thing to disassemble here.

source/validate_cfg.cpp
dj2: Done #1661

source/val/basic_block.h
dj2: Done

source/validate_cfg.cpp
dj2: Nice, now they all give diagnostic again. Thanks.

PR #1663, opened 2018-07-03
Convert validation to use libspriv::Instruction where possible.

dj2:
For the instructions which execute after the IdPass check we can provide
the Instruction instead of the spv_parsed_instruction_t. This
Instruction class provides a bit more context (like the source line)
that is not available from spv_parsed_instruction_t.


dj2:
@alan-baker ping.

source/val/instruction.h
```diff
@@ -66,6 +66,10 @@ class Instruction {
   /// The words used to define the Instruction
   const std::vector<uint32_t>& words() const { return words_; }
 
+  const spv_parsed_operand_t& operand(size_t idx) const {
```
alan-baker: Comment please.

dneto0: Also, size_t is excessive since by SPIR-V format there are at most 64K words and therefore at most 64K operands (though I don't know of a case of more than 15 or so).

But I see you have changed some loops to use size_t indices.   So I'm not that concerned about this.

source/val/instruction.h
```diff
@@ -74,13 +78,18 @@ class Instruction {
...
+  }
+
   // Casts the words belonging to the operand under |index| to |T| and returns.
   template <typename T>
   T GetOperandAs(size_t index) const {
-    const spv_parsed_operand_t& operand = operands_.at(index);
-    assert(operand.num_words * 4 >= sizeof(T));
-    assert(operand.offset + operand.num_words <= inst_.num_words);
-    return *reinterpret_cast<const T*>(&words_[operand.offset]);
+    const spv_parsed_operand_t& op = operands_.at(index);
```
dneto0: I prefer the full word " operand" since "op" can often be confused for "opcode" especially since every SPIR-V instruction starts with "Op" as a prefix.  Having the full word here aids speed reading the code.

source/val/instruction.h
dj2: The size_t matches the existing GetOperandAs method as well. Using size_t here keeps things consistent.

source/val/instruction.h
dj2: Renamed to the_operand due to the operand() method taking the name.

dj2: Even better, o, given how short it is.

dj2:
PTAL.

PR #1669, opened 2018-07-04
Make sure the constant folder get the correct type.

s-perron:
There are a few locations where we need to handle duplicate types.  We
cannot merge them because they may be needed for reflection.  When this
happens we need do some extra lookups in the type manager.

The specific fixes are:

1) When generating a constant through `GetDefiningInstruction` accept
and use an id for the desired type of the constant.  This will make sure
you get the type that is needed.

2) In Private-to-local, make sure we to update the def-use chains when a
new pointer type is created.

3) In the type manager, make sure that `FindPointerToType` returns a
pointer that points to the given type and not a duplicate type.

4) In scalar replacment, make sure the null constants that are created
are the correct type.

### DIFF
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -795,7 +795,7 @@ void IRContext::AddAnnotationInst(std::unique_ptr<Instruction>&& a) {
 void IRContext::AddType(std::unique_ptr<Instruction>&& t) {
   module()->AddType(std::move(t));
   if (AreAnalysesValid(kAnalysisDefUse)) {
-    get_def_use_mgr()->AnalyzeInstDef(&*(--types_values_end()));
+    get_def_use_mgr()->AnalyzeInstDefUse(&*(--types_values_end()));
   }
 }
 
--- a/test/opt/constant_manager_test.cpp
+++ b/test/opt/constant_manager_test.cpp
@@ -1,0 +1,51 @@
+// Copyright (c) 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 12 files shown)


source/opt/ir_context.h
```diff
@@ -795,7 +795,7 @@ void IRContext::AddAnnotationInst(std::unique_ptr<Instruction>&& a) {
 void IRContext::AddType(std::unique_ptr<Instruction>&& t) {
   module()->AddType(std::move(t));
   if (AreAnalysesValid(kAnalysisDefUse)) {
-    get_def_use_mgr()->AnalyzeInstDef(&*(--types_values_end()));
+    get_def_use_mgr()->AnalyzeInstDefUse(&*(--types_values_end()));
```
dnovillo: Why are we analysing def-use now?  I don't have an issue with it, but I'm not sure I understand why we need this now.

test/opt/constant_manager_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+  StructConstant struct_const_1(struct_type_1->AsStruct());
+  ir::Instruction* const_inst_1 =
+      context->get_constant_mgr()->GetDefiningInstruction(&struct_const_1, 1);
+  EXPECT_EQ(const_inst_1->type_id(), 1);
+
+  Type* struct_type_2 = context->get_type_mgr()->GetType(2);
+  StructConstant struct_const_2(struct_type_2->AsStruct());
+  ir::Instruction* const_inst_2 =
+      context->get_constant_mgr()->GetDefiningInstruction(&struct_const_2, 2);
+  EXPECT_EQ(const_inst_2->type_id(), 2);
```
dnovillo: If we don't pass anything as the type_id here, should we expect a 1?

s-perron: It could be a 1 or a 2.  There is no clearly defined value when the type_id is not provided.

source/opt/ir_context.h
s-perron: When you analyze just the definitions, it clears all of the uses.  Then the uses in the type definition are not present.  This lead to the def-use chains being inconsistent.  I think that `AddType` should do nothing or do both.  Only doing the definition is odd.

Issue #1659, opened 2018-06-29
Optimizer memory leaks


jodavis42:
The optimizer has two fairly obvious memory leaks, both in fold.cpp.
One is line 202 in the function FoldInstructionInternal and the other is on line 214 in the function GetConstantFoldingRules.

Both of these could be turned into static non-pointers to get rid of the leaks from my quick tests.

dneto0:
I'd suggest these objects belong to the IRContext object.

s-perron:
When I wrote this code I did not want to make the object static to avoid issues with static variable construction and destruction order. See our style guide: https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables.

I had considered placing it in the IRContext, but decided against that.  This would expose the folding rules to anybody who has access to the IRContext.  I don't want that.  I want the internal implementation of the folding functions to remain hidden.

That is why I opted for this pattern.  Notice this is the last bullet point in the section on static and global variables in the style guide:

> If all else fails, you can create an object dynamically and never delete it by binding the pointer to a function-local static pointer variable: static const auto* const impl = new T(args...); (If the initialization is more complex, it can be moved into a function or lambda expression.)

Let me know if there are other issues I have not considered.

jodavis42:
So while I agree with using a static pointer instead of a static instance I feel that letting this leak isn't a good alternative.

In the application I'm working on, we try to make sure we don't have any memory leaks. However now that I'm using the spir-v optimizer I get false leak reports that make it significantly harder to track down leaks in our own code base as well as near impossible to set up automated leak tests. I imagine as spir-v gains more traction that this will be an issue for other users too.

Would it be possible to for you to register an atexit callback to clean this up? Or at least add a global Shutdown/Cleanup function so I can manually invoke cleanup?

s-perron:
Okay.  What I did was move the folding routines into a class, and that class owns the memory that is being leaked.  Then the context owns an instance of that class.  It should get destroyed when the context is destroyed.

PR #1672, opened 2018-07-05
Move folding routines into a class

s-perron:
The folding routines are currently global functions.  They also rely on
data in an std::map that holds the folding rules for each opcode.  This
causes that map to not have a clear owner, and therefore never gets
deleted.

There has been a request to delete this map.  To implement this, we will
create a InstructionFolder class that owns the maps.  The IRContext will
own the InstructionFolder instance.  Then the global functions will
become public memeber functions of the InstructionFolder.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/1659.

### DIFF
--- a/source/opt/fold.cpp
+++ b/source/opt/fold.cpp
@@ -41,9 +41,9 @@ namespace {
 #define UINT32_MAX 0xffffffff /* 4294967295U */
 #endif
 
-// Returns the single-word result from performing the given unary operation on
-// the operand value which is passed in as a 32-bit word.
-uint32_t UnaryOperate(SpvOp opcode, uint32_t operand) {
+}  // namespace
+
+uint32_t InstructionFolder::UnaryOperate(SpvOp opcode, uint32_t operand) const {
   switch (opcode) {
     // Arthimetics
     case SpvOp::SpvOpSNegate:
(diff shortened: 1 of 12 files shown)


source/opt/fold.cpp
```diff
@@ -199,28 +193,21 @@ bool FoldInstructionInternal(ir::Instruction* inst) {
   std::vector<const analysis::Constant*> constants =
       const_manager->GetOperandConstants(inst);
 
-  static FoldingRules* rules = new FoldingRules();
-  for (FoldingRule rule : rules->GetRulesForOpcode(opcode)) {
+  for (FoldingRule rule : GetFoldingRules().GetRulesForOpcode(opcode)) {
```
dneto0: Drive by: Would it be useful if this were a FoldingRule reference?

source/opt/fold.h
```diff
@@ -21,76 +21,145 @@
...
+class InstructionFolder {
+ public:
+  // Returns the result of folding a scalar instruction with the given |opcode|
+  // and |operands|. Each entry in |operands| is a pointer to an
+  // analysis::Constant instance, which should've been created with the constant
+  // manager (See IRContext::get_constant_mgr).
+  //
+  // It is an error to call this function with an opcode that does not pass the
+  // IsFoldableOpcode test. If any error occurs during folding, the folder will
+  // faill with a call to assert.
```
dneto0: spelling of "fail"

dneto0:
Rebased and fixed typos. Pushed onto master as  a45d4cac6

source/opt/fold.cpp
s-perron: A folding rule is a function object.  Not something large.  I don't think it will make a difference.

source/opt/fold.h
s-perron: Fixed.

PR #1673, opened 2018-07-06
Update SPIRV-Headers

dj2:
Roll SPIRV-Headers dep to pickup the new StorageBuffer8Bit extensions.


dneto0:
Rebased and pushed onto master as fa78d3bec9

PR #1674, opened 2018-07-06
Move comp code into comp namespace

dj2:
This CL moves the code in the comp/ directories into the comp namespace.

### DIFF
--- a/source/comp/markv.h
+++ b/source/comp/markv.h
@@ -24,10 +24,11 @@
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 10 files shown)


dneto0:
Need to ignore the Kokoro failures.  The test config can't be found because this MR was based off a version of master that didn't have them.
Waiting for Travis and AppVeyor

dneto0:
Rebased, squashed, and pushed into master as 9836b05a

PR #1675, opened 2018-07-06
Move utils/ to spvtools::utils namespace

dj2:
Currently the utils/ folder uses both spvutils:: and spvtools::utils.
This CL changes the namespace to consistently be spvtools::utils to match
the rest of the codebase.

### DIFF
--- a/test/bit_stream.cpp
+++ b/test/bit_stream.cpp
@@ -19,24 +19,9 @@
 #include "gmock/gmock.h"
 #include "util/bit_stream.h"
 
+namespace spvtools {
+namespace utils {
 namespace {
-
-using spvutils::BitReaderInterface;
-using spvutils::BitReaderWord64;
-using spvutils::BitsetToStream;
-using spvutils::BitsToStream;
-using spvutils::BitWriterInterface;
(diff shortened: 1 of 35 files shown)


test/bit_stream.cpp
```diff
@@ -19,25 +19,10 @@
 #include "gmock/gmock.h"
 #include "util/bit_stream.h"
 
+namespace spvtools {
```
dneto0: Huh.  I never thought of putting an unnamed namespace inside a real namespace.

dj2: It's nice because you can remove all the using statements below. Technically, I don't think the tests need to be wrapped in an anonymous namespace, but I don't think it hurts either.

dneto0: We like putting test code in anonymous namespace so there is no chance of having a dependency from the code being tested to the test code.

dneto0:
The conflict is fairly trivial.
The Travis failure is a timeout on the Android build.  Seems spurious.

dneto0:
Rebased, resolved merge conflict, and pushed into master as 76e0bde19

PR #1680, opened 2018-07-08
Move the ir namespace to opt.

dj2:
This CL moves the files in opt/ to consistenly be under the opt::
namespace. This frees up the ir:: namespace so it can be used to make a
shared ir represenation.


source/opt/block_merge_pass.h
```diff
@@ -38,33 +38,33 @@ class BlockMergePass : public Pass {
...
-           ir::IRContext::kAnalysisDecorations |
-           ir::IRContext::kAnalysisCombinators |
-           ir::IRContext::kAnalysisNameMap;
+  Status Process(opt::IRContext*) override;
+  virtual opt::IRContext::Analysis GetPreservedAnalyses() override {
+    return opt::IRContext::kAnalysisDefUse |
+           opt::IRContext::kAnalysisInstrToBlockMapping |
+           opt::IRContext::kAnalysisDecorations |
+           opt::IRContext::kAnalysisCombinators |
+           opt::IRContext::kAnalysisNameMap;
```
s-perron: All of the `opt::` are no longer needed.  Here and in a lot of other places.  How hard is it to have a tool clean these up?

dj2: It won't be hard, I just want to do it as a second pass. Tried it as one and the change gets huge and hard to debug.

PR #1681, opened 2018-07-09
Fix up code to make ClangTidy happy.

s-perron:
Just a few changes to pass `std::function` objects by const reference
instead of by value.


s-perron:
The message that clang tidy gave was:

> the parameter 'id_map' is copied for each invocation but only used as a const reference; consider making it a const reference

PR #1682, opened 2018-07-09
Move the validation code into the val:: namespace

dj2:
This CL moves the validation code to the val:: namespace. This makes it
clearer which instance of the Instruction and other classes are being
referred too.


dj2:
The Travis Android timed out but the kokoro Android builds all passed, so I'm going to land this without trying to kick Travis again.

PR #1693, opened 2018-07-10
Vulkan permits non-monotonic offsets for block members

### LINKED ISSUE #1685: Vulkan block layout does not require member offsets to be monotonic
Vulkan 14.5.4 Offset and Stride Assignment has a clause:

> The numeric order of Offset decorations need not follow member declaration order.

This contradicts the SPIR-V spec, but it overrides it. (I have filed an internal issue to resolve the conflict.)

dneto0:
Other environments do not.
Add tests for OpenGL 4.5 and SPIR-V universal 1.0 to ensure
they still check monotonic layout.

For universal 1.0, we're assuming it otherwise follows Vulkan
rules for block layout.

Fixes #1685

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -18,6 +18,7 @@
 #include <stack>
 
 #include "opcode.h"
+#include "spirv_target_env.h"
 #include "val/basic_block.h"
 #include "val/construct.h"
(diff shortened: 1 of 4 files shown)


dneto0:
Rebased and fixed formatting.

PR #1696, opened 2018-07-10
Reopen stdin for binary as needed

### LINKED ISSUE #1688: spirv-dis succeeds with a filename but fails when reading from stdin
`spirv-dis < foo.spv`
`...`
`error: 130: End of input reached while decoding OpDecorate starting at word 129: missing ID operand at word offset 1.`

As noted above, `spirv-dis <filename>` succeeds.

After inspecting `tools/io.h:ReadFile()`, this might be a case where you want to reopen stdin with `freopen(NULL,"rb",stdin)`?

I'm using the prebuilt Vulkan SDK 1.1.77.0 on Win10.

Example .spv attached.

[comp.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/2178208/comp.zip)

dneto0:
Fixes #1688

### DIFF
--- a/tools/io.h
+++ b/tools/io.h
@@ -21,13 +21,15 @@
 
 // Appends the content from the file named as |filename| to |data|, assuming
 // each element in the file is of type |T|. The file is opened with the given
-// |mode|. If |filename| is nullptr or "-", reads from the standard input. If
-// any error occurs, writes error messages to standard error and returns false.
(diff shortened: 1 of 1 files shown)


dneto0:
Looks like a spurious Travis failure on the Android build.  We've removed Travis support on master so I'll check this in.

dneto0:
Rebased

PR #1698, opened 2018-07-10
Rename ForEachId for clarity

dj2:
The ForEachId method has the side effect that it will also set the type
and result IDs. This CL renames the method to ForEachIdUpdateId to make
it to the effect clearer.

The non-const ForEachId is removed as it is unused.

### DIFF
--- a/source/link/linker.cpp
+++ b/source/link/linker.cpp
@@ -165,7 +165,7 @@ spv_result_t ShiftIdsInModules(const MessageConsumer& consumer,
        ++module_iter) {
     Module* module = *module_iter;
     module->ForEachInst([&id_bound](Instruction* insn) {
-      insn->ForEachId([&id_bound](uint32_t* id) { *id += id_bound; });
+      insn->ForEachIdUpdateId([&id_bound](uint32_t* id) { *id += id_bound; });
     });
     id_bound += module->IdBound() - 1u;
     if (id_bound > 0x3FFFFF)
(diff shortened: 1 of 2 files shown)


s-perron:
Personally, I'm not sure that this is any clearer, and I find it awkward to read.  The id is not always updated. Whether or not it is updated depends on what the function that acts on the id does.  In fact, I find the current name to be in line with the name of the std::for_each function.

My preference is to leave it as is; however, I don't think it is a big deal.  If most other people think the change is a good one, then I'm willing to go with it.

dj2:
(although, I do agree my new name is bad, I just can't come up with a better one)

s-perron:
I do not think that is true.  You can change the instruction in the `ForEachInst` function.  The `ForEachBlock*` you get a pointer to the basic block, and that basic block could be changed.  I think most of them are the same way, the function being passed in takes a non-const pointer to something.  That something can be changed, but it is not in most cases.

dj2:
In that case, the side effect is in the callers code (so, not a side effect but intended behaviour), so the caller should know it's going to happen. In this case, there is the extra side effect which isn't obvious from the call site.

dj2:
(Dropping this is also fine, if the current name is preferred)

dj2:
Chatted offline, the type_id_ and result_id_ fields are more like caches then the source of truth. So, this updates the cached version if we changed the value during the function callback. So, closing this as NamedAsIntended.

PR #1700, opened 2018-07-10
Remove context() method from opt::Function

dj2:
This CL removes the context() method from opt::Function. In the places
where the context() was used we can retrieve, or provide, the context in
another fashion.

#1703

### DIFF
--- a/source/opt/dominator_tree.h
+++ b/source/opt/dominator_tree.h
@@ -157,7 +157,7 @@ class DominatorTree {
   void DumpTreeAsDot(std::ostream& out_stream) const;
 
   // Build the (post-)dominator tree for the given control flow graph
-  // |cfg| and the function |f|. |f| must exist in the |cfg|. Any 
+  // |cfg| and the function |f|. |f| must exist in the |cfg|. Any
   // existing data in the dominator tree will be overwritten
   void InitializeTree(const CFG& cfg, const opt::Function* f);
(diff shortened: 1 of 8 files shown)


dj2:
Actually, hold off on this until I finish the Pass constructor CL ....

dneto0:
> Actually, hold off on this until I finish the Pass constructor CL ....

Convention is to add WIP to the title.  I did that for this CL.
Remove it when you want a review.

s-perron:
I'm not against this change, but could you create an umbrella issue that ties all of the refactoring changes together with an explanation of what you are trying to accomplish with the changes?  I know we have talked about it a little, but it would be good to make it public.

dj2:
Done. #1703

dj2:
The Pass::Run change has landed so this should be good to go. PTAL.

source/opt/dominator_tree.h
```diff
@@ -158,7 +158,7 @@ class DominatorTree {
 
   // Build the (post-)dominator tree for the function |f|
   // Any existing data will be overwritten
-  void InitializeTree(const opt::Function* f);
+  void InitializeTree(const CFG& cfg, const opt::Function* f);
```
s-perron: Should add comments about `cfg`.  Make sure to mention that `cfg` includes `f`.

PR #1701, opened 2018-07-10
Enforce block layout rules even when relaxed

### LINKED ISSUE #1697: relaxed block layout: Check all layout rules except those that are relaxed

dneto0:
- Vulkan 1.0 uses strict layout rules
- Vulkan 1.0 with relaxed-block-layout validator option
  enforces all rules except for the relaxation of vector
  offset.
- Vulkan 1.1 and later always supports relaxed block layout

Add spot check tests for the relaxed-block-layout scenarios.

Fixes #1697


dneto0:
Rebased

PR #1702, opened 2018-07-11
[WIP] Change Optimizer to build passes as late as possible.

dj2:
Currently the Optimizer pass objects are created during the Optimizer
setup, before Run is executed. This means we don't have access to the
IRContext when creating the passes.

In order to pass the IRContext to the Pass constructors we need to
create the pass later in the pipeline. This CL changes the Optimizer to
create PassTokens instead of Pass objects. The PassTokens are stored in
the PassManager and the token creates the real Pass object when we're
running all of the passes.

This has the side effect that we now tear down the passes after they're
used, and we don't have to clear the pass list for the optimizer as it
can be re-run on a different binary without re-creating the state.

#1703

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -24,10 +24,11 @@
 #include "libspirv.hpp"
 
 namespace spvtools {
-
 namespace opt {
-class Pass;
-}
+
+class PassToken;
+
+}  // namespace opt
 
 // C++ interface for SPIR-V optimization functionalities. It wraps the context
 // (including target environment and the corresponding SPIR-V grammar) and
@@ -49,7 +50,7 @@ class Optimizer {
     // below; for out-of-tree passes, use this constructor instead.
     // Note that this API isn't guaranteed to be stable and may change without
     // preserving source or binary compatibility in the future.
-    PassToken(std::unique_ptr<opt::Pass>&& pass);
+    PassToken(std::unique_ptr<opt::PassToken>&& pass);
 
     // Tokens can only be moved. Copying is disabled.
     PassToken(const PassToken&) = delete;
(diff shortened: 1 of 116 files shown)


include/spirv-tools/optimizer.hpp
```diff
@@ -49,7 +50,7 @@ class Optimizer {
     // below; for out-of-tree passes, use this constructor instead.
     // Note that this API isn't guaranteed to be stable and may change without
     // preserving source or binary compatibility in the future.
-    PassToken(std::unique_ptr<opt::Pass>&& pass);
+    PassToken(std::unique_ptr<opt::PassToken>&& pass);
```
dj2: This is public facing change. But, it changes from one opaque type to a second, so I think it's safe.

source/opt/pass_manager.cpp
```diff
@@ -62,7 +67,6 @@ Pass::Status PassManager::Run(opt::IRContext* context) {
   if (status == Pass::Status::SuccessWithChange) {
     context->module()->SetIdBound(context->module()->ComputeIdBound());
   }
-  passes_.clear();
```
dj2: This is a second, possibly, visible change (and was causing a linker test error). Previously the pass manager would dump all pass information when it was done so you could re-use the manager by setting up everything again.

This CL leaves the Tokens in place so the manager can be re-used with a given set of passes.

I can change this back to clearing if preferred, we just need to decide if we want the manager to b re-usable with new passes, or with the previously configured passes.

dj2:
Ok, tests pass locally so I think this is good for review now.

dj2:
A simplified version of this CL is in #1708 where the Pass::Run method stores the context instead of letting Pass::Process do it. The other change is a lot less invasive then this change.

dj2:
Going with #1708

PR #1707, opened 2018-07-11
Add folding rule to merge a vector shuffle feeding another one.

s-perron:
In some cases, there will be an OpVectorShuffle instruction feeding another where the end result only uses elements from two of the input vectors.  In those cases, the multiple OpVectorShuffle instructions can replaced by a single instructions.

### DIFF
--- a/test/opt/fold_test.cpp
+++ b/test/opt/fold_test.cpp
@@ -5650,6 +5650,174 @@ INSTANTIATE_TEST_CASE_P(StoreMatchingTest, MatchingInstructionWithNoResultFoldin
         0 /* OpStore */, true)
 ));
 
+INSTANTIATE_TEST_CASE_P(VectorShuffleMatchingTest, MatchingInstructionWithNoResultFoldingTest,
+::testing::Values(
+    // Test case 0: Basic test 1
+    InstructionFoldingCase<bool>(
+        Header() +
+            "; CHECK: OpVectorShuffle\n" +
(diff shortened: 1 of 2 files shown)


test/opt/fold_test.cpp
```diff
@@ -5652,5 +5652,173 @@ ::testing::Values(
...
+            "%4 = OpVariable %_ptr_v4double Function\n" +
+            "%5 = OpLoad %v4double %2\n" +
+            "%6 = OpLoad %v4double %3\n" +
+            "%7 = OpLoad %v4double %4\n" +
+            "%8 = OpVectorShuffle %v4double %5 %6 2 3 4 5\n" +
+            "%9 = OpVectorShuffle %v4double %7 %8 2 3 7 5\n" +
+            "OpReturn\n" +
+            "OpFunctionEnd",
+        9, false),
+    // Test case 6: Don't fold, need both operands of the feeder.
```
dnovillo: Check that the fold didn't actually happen here?

s-perron: That is what the `false` is at the end of the struct.  I'll add comments to clarify.  If no folding happens then the EFFCEE checks are not done.

dnovillo: Ah, thanks.  Yeah, I had forgotten about that subtlety.

s-perron: Should I add a comment to it?

dnovillo: Not really needed.  We have other cases, and that's the documented behaviour.  Thanks.

Issue #1703, opened 2018-07-11
Expose a representation of the IR through the public interface


dj2:
Currently we have no way for users of SPIRV-Tools to build a representation of the IR. Each user has to handle dominator analysis and all the other bits of turning the SPIR-V into a representation.

Using the IR code in source/opt is a good starting place for this representation but it needs to be split apart in order to extract it nicely. The namespacing has been changed to free up the IR namespace in #1680. We should also dis-entangle the IRContext from the various pieces if possible.

PR #1708, opened 2018-07-11
use Pass::Run to set the context on each pass.

dj2:
Currently the IRContext is passed into the Pass::Process method. It is
then up to the individual pass to store the context into the context_
variable. This CL changes the Run method to store the context before
calling Process which no-longer receives the context as a parameter.

Issue #1703


dj2:
This is the less invasive version of #1702 (and a followup that wasn't posted yet). With this version we set the context in the Run method and remove it from the Process methods. We don't change the constructor for the passes.

PR #1709, opened 2018-07-11
Pass the IRContext into the folding rules.

dj2:
This CL updates the folding rules to receive the IRContext as a paramter
instead of retrieving off of the Instruction.

Issue #1703

### DIFF
--- a/source/opt/fold.h
+++ b/source/opt/fold.h
@@ -28,6 +28,8 @@ namespace opt {
 
 class InstructionFolder {
  public:
+  InstructionFolder(IRContext* context) : context_(context) {}
+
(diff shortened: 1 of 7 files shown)


source/opt/fold.h
```diff
@@ -28,6 +28,8 @@ namespace opt {
 
 class InstructionFolder {
  public:
+  InstructionFolder(IRContext* context) : context_(context) {}
```
dneto0: Please make this an explicit constructor

dj2: Done. Editing on github isn't too bad, heh.

PR #1711, opened 2018-07-11
Add option to skip verifying block layout

antiagainst:
We need this to avoid emitting errors on DirectX layout rules.

### DIFF
--- a/tools/val/val.cpp
+++ b/tools/val/val.cpp
@@ -46,7 +46,10 @@ Options:
   --max-access-chain-indexes       <maximum number of indexes allowed to use for Access Chain instructions>
   --relax-logical-pointer          Allow allocating an object of a pointer type and returning
                                    a pointer value from a function in logical addressing mode
-  --relax-block-layout             Skips checking of standard uniform/storage buffer layout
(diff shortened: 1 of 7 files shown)


tools/val/val.cpp
```diff
@@ -46,7 +46,10 @@ NOTE: The validator is a work in progress.
   --max-access-chain-indexes       <maximum number of indexes allowed to use for Access Chain instructions>
   --relax-logical-pointer          Allow allocating an object of a pointer type and returning
                                    a pointer value from a function in logical addressing mode
-  --relax-block-layout             Skips checking of standard uniform/storage buffer layout
+  --relax-block-layout             Enable VK_HR_relaxed_block_layout when checking standard
+                                   uniform/storage buffer layout
+  --skip-block-layout              Skip checking standard uniform/storage buffer layout
+                                   uniform/storage buffer layout
```
dneto0: Actually this line does not belong. I will remove it before pushing into master

dneto0:
Rebased, fixed the help line, and pushed into master as 4db9c789

PR #1716, opened 2018-07-12
Cleanup extraneous namespace qualifies in source/opt.

dj2:
This CL follows up on the opt namespacing CLs by removing the
unnecessary opt:: and opt::analysis:: namespace prefixes.


dj2:
The followup as requested. This is basically just a sed command run over the source/opt files.

PR #1718, opened 2018-07-12
Replace opt::Instruction type and result cache with flags.

dj2:
Currentlty opt::Instruction class holds a cache of the result_id and
type_id for the instruction. That cache needs to be updated if the
underlying operand values are changes.

This CL changes the cache to being a flag if there is a type or result
id for the instruction. We then retrieve the value if needed from the
operands.


dj2:
Thoughts? This is an attempt at cleaning up the cached instruction parameters as we discussed after the abandoned ForEachId change.

source/opt/instruction.h
```diff
@@ -506,28 +512,27 @@ inline void Instruction::SetInOperands(OperandList&& new_operands) {
 }
 
 inline void Instruction::SetResultId(uint32_t res_id) {
-  result_id_ = res_id;
-  auto ridx = (type_id_ != 0) ? 1 : 0;
+  auto ridx = has_type_id_ ? 1 : 0;
```
dneto0: There's a weirdness here.  If res_id is 0, should we also update the has_result_id_ flag?  And vice versa?

I think we should. Needs
` has_result_id_ = res_id !=0;`

source/opt/instruction.h
```diff
@@ -506,28 +512,27 @@ inline void Instruction::SetInOperands(OperandList&& new_operands) {
...
   operands_[ridx].words = {res_id};
 }
 
 inline void Instruction::SetResultType(uint32_t ty_id) {
-  if (type_id_ != 0) {
-    type_id_ = ty_id;
-    assert(operands_.front().type == SPV_OPERAND_TYPE_TYPE_ID);
-    operands_.front().words = {ty_id};
-  }
+  if (!has_type_id_) return;
```
dneto0: This preserves behaviour:  an instruction has a type id if and only if it always had a type. 
It's a bit weird because you can't take a Nop instruction and turn it into something like OpConstant.  You have to construct an instruction with a type if it ever needs a type.
But that's all a pre-existing problem.

source/opt/instruction.h
```diff
@@ -576,9 +581,6 @@ inline void Instruction::ForEachInst(
 inline void Instruction::ForEachId(const std::function<void(uint32_t*)>& f) {
   for (auto& opnd : operands_)
     if (spvIsIdType(opnd.type)) f(&opnd.words[0]);
-  if (type_id_ != 0u) type_id_ = GetSingleWordOperand(0u);
```
dneto0: Wow. that was pretty weird.

source/opt/instruction.h
dj2: Or, maybe it should be assert(res_id != 0); If you  have a result_id and you set it to 0 here you'll still have an entry in the operand array for it which would be bad. I think we want to just disallow that from happening?

source/opt/instruction.h
dj2: I think the issue is adding and removing things from the Operand list if you add/remove type and result ids.

source/opt/instruction.h
dj2: Yea, those are the lines which triggered this in the first place, heh.

s-perron:
I'll check the compile time for these changes today.

source/opt/instruction.h
s-perron: I think we should assert.  Before this change if we did this change we would of had an error when trying to remove the result type (see below).  We can rely on `ToNop` to create a Nop.

If we need to be able to delete a result from an instruction because we changed it to something else, we can see the use case.

source/opt/instruction.h
s-perron: Yes there is a problem that the existence of a type and result id should be determined by the opcode, but we get problems with consistency, but for this PR just retain the same behaviour.

With that in mind, if we set the result type to "0" in the old implementation, then later checks would think there was no type id.  Then the operands would not be lined up correctly.  This was incorrect.  

I think we should document and assert that you cannot change the result type to 0.

source/opt/instruction.h
dj2: Done.

source/opt/instruction.h
dj2: Done.

dj2:
PTAL.

dj2:
ping. dneto@ does this look good now?

PR #1724, opened 2018-07-13
Fix finding constant with particular type.

s-perron:
With current implementation, the constant manager does not keep around
two constant with the same value but different types when the types
hash to the same value. So when you start looking for that constant you
will get a constant with the wrong type back.

I've made a few changes to the constant manager to fix this.  First off,
I have changed the map from constant to ids to be an std::multimap.
This way a single constant can be mapped to mutiple ids each
representing a different type.

Then when asking for an id of a constant, we can search all of the ids
associated with that constant in order to find the one with the correct
type.


source/opt/constants.cpp
```diff
@@ -131,6 +131,37 @@ std::vector<const Constant*> ConstantManager::GetOperandConstants(
...
 }
 
+uint32_t ConstantManager::FindDeclaredConstant(const Constant* c,
+                                               uint32_t type_id) const {
+  c = FindConstant(c);
+  if (c == nullptr) {
+    return 0;
+  }
+
+  auto iter = const_val_to_id_.lower_bound(c);
```
alan-baker: std::tie(start, end) = const_val_to_id_.equal_range(c);
while (start != end) {
  ...
}

source/opt/const_folding_rules.cpp
```diff
@@ -197,13 +197,25 @@ ConstantFoldingRule FoldCompositeWithConstants() {
...
     std::vector<uint32_t> ids;
-    for (const analysis::Constant* element_const : constants) {
+    for (uint32_t i = 0; i < constants.size(); ++i) {
+      const analysis::Constant* element_const = constants[i];
       if (element_const == nullptr) {
         return nullptr;
       }
-      uint32_t element_id = const_mgr->FindDeclaredConstant(element_const);
+
+      uint32_t component_type_id = 0;
```
alan-baker: We likely have a similar issue related to pointers in another function. Fix or file an issue please.

source/opt/constants.cpp
s-perron: Done.

source/opt/const_folding_rules.cpp
s-perron: I looked at all of the function in the constant manager.  Others member functions that return an instruction or id all take an optional type id as a parameter already.

I missed this one in an earlier change because I did not look closely enough at the implementation.  The const-to-id map would use the address of the constant for the hash, so I figured two constant that "equal" except for different types would both be in the map.

Can you point to any specific ones that I may have missed?

Issue #1719, opened 2018-07-12
Bad line number it spir-val check for vector shuffle


s-perron:
This shader does not validate, and it should not.  However, the error message is not very useful.  Especially the line number:

```
error: line 18446744073709551615: Component literal value 4 is greater than 3.
```

```
               OpCapability SampledBuffer
               OpCapability StorageImageExtendedFormats
               OpCapability Shader
        %978 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %CS "CS"
               OpExecutionMode %CS LocalSize 256 1 1
               OpSource HLSL 600
       %uint = OpTypeInt 32 0
      %float = OpTypeFloat 32
    %v4float = OpTypeVector %float 4
    %v2float = OpTypeVector %float 2
       %void = OpTypeVoid
        %548 = OpTypeFunction %void
%CS = OpFunction %void None %548
        %550 = OpLabel
       %6275 = OpUndef %v2float
       %6280 = OpUndef %v2float
       %6282 = OpVectorShuffle %v4float %6275 %6280 0 1 4 5
               OpReturn
               OpFunctionEnd     
```

dj2:
Fixed by #1725.

PR #1725, opened 2018-07-16
Fixup line number for OpVectorShuffle ID error.

dj2:
This CL updates the code to pull a valid instruction for the line number
when outputting a component error in OpVectorShuffle. The error line
isn't the best at this point as it points at the component, but it's
better then a -1 (turning to max<size_t>) that was being output.

The error messages has been updated to better reflect what the error is
attempting to say.

Issue #1719.

### DIFF
--- a/source/val/validate_id.cpp
+++ b/source/val/validate_id.cpp
@@ -1781,9 +1781,12 @@ bool idUsage::isValid<SpvOpVectorShuffle>(const spv_instruction_t* inst,
   for (size_t i = firstLiteralIndex; i < inst->words.size(); ++i) {
     auto literal = inst->words[i];
     if (literal != 0xFFFFFFFF && literal >= N) {
-      DIAG(module_.FindDef(inst->words[0]))
-          << "Component literal value " << literal << " is greater than "
-          << N - 1 << ".";
+      // TODO(dsinclair): This should provide the line number of the
+      // OpVectorShuffle but need the spv_instruction_t to be an Instruction
+      // in order to provide that value.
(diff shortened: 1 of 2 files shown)


dj2:
This is a partial fix to get rid of the gigantic line number. The line output ends up being for the Vector result, instead of the OpVectorShuffle instruction. Getting to the right line is going to be a much more invasive change, so I wanted to land this part first to at least make things a bit nicer.

source/val/validate_id.cpp
```diff
@@ -1781,9 +1781,12 @@ bool idUsage::isValid<SpvOpVectorShuffle>(const spv_instruction_t* inst,
   for (size_t i = firstLiteralIndex; i < inst->words.size(); ++i) {
     auto literal = inst->words[i];
     if (literal != 0xFFFFFFFF && literal >= N) {
-      DIAG(module_.FindDef(inst->words[0]))
-          << "Component literal value " << literal << " is greater than "
-          << N - 1 << ".";
+      // TODO(dsinclair): This should provide the line number of the
+      // OpVectorShuffle but need the spv_instruction_t to be an Instruction
+      // in order to provide that value.
+      DIAG(module_.FindDef(inst->words[1]))
```
dneto0: words[1] is the result *type* not the result id.  ?

dj2: Oh, using words[2] works because that's the Result <id>. I got confused by the type vs id distinction. Gives the correct line number now.

dneto0:
Rebased, squashed, and pushed into master as 8c7dab5caa5

PR #1726, opened 2018-07-18
Add GN Build

dj2:
This CL adds the necessary infrastructure to create a GN build. Only the libraries are generated at this point. Fuzzer targets have been added to show that building and linking of the static_libraries works correctly.


testing/test.gni
```diff
@@ -0,0 +1,413 @@
+# Copyright 2018 The Chromium Authors. All rights reserved.
```
dj2: This file is copied from Chromium, so I left the Chromium copyright and added the license text below.

dneto0: Thanks for calling that out.  Seems ok to me.

testing/fuzzers/spvtools_opt_fuzzer.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  spvtools::Optimizer optimizer(SPV_ENV_UNIVERSAL_1_3);
+  optimizer.SetMessageConsumer([](spv_message_level_t, const char*,
+                                  const spv_position_t&, const char*) {});
+
+  std::vector<uint32_t> input;
+  input.resize(size >> 2);
+
+  size_t count = 0;
```
dneto0: I would have gone with a std::memcpy.   Meh.

testing/fuzzers/spvtools_opt_fuzzer.cpp
```diff
@@ -0,0 +1,36 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cstdint>
+
+#include "spirv-tools/optimizer.hpp"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  spvtools::Optimizer optimizer(SPV_ENV_UNIVERSAL_1_3);
```
dneto0: The optimizer won't do anything unless you register some passes.  So let's expand this later.

dneto0:
Rebased, squashed, and pushed into master as 3b9f197be1

dneto0:
Yep, I rebased and squashed on my machine but failed to push ...  

I've rebased again and pushed as a114e1f30c34

Kangz:
Thank you!

PR #1728, opened 2018-07-19
Fix block ordering in dead branch elim

### LINKED ISSUE #1727: Dead branch elimination can produce invalid block orders
```
OpCapability Shader
OpMemoryModel Logical GLSL450
OpEntryPoint Fragment %func "func"
%void = OpTypeVoid
%bool = OpTypeBool
%true = OpConstantTrue %bool
%func_ty = OpTypeFunction %void
%func = OpFunction %void None %func_ty
%1 = OpLabel
OpSelectionMerge %3 None
OpBranchConditional %true %2 %3
%3 = OpLabel
OpReturn
%2 = OpLabel
%not = OpLogicalNot %bool %true
OpBranch %3
OpFunctionEnd
```

Dead branch elimination correctly reduces this case to straight line code, but it fails to fix the block ordering. `%2` must precede `%3` post-optimization since it dominates `%3`.

alan-baker:
Fixes #1727

* If the pass finds any dead branches it can optimize then at the end of
the pass it reorders basic blocks to ensure they satisfy block ordering
requirements
 * Added some new tests
* While investigating this issue, found and fixed a non-deterministic
ordering of dominators
 * Now the edges used to construct the dominator tree are sorted
 according to posorder traversal indices

Note: including multiple ordering functions limited the number of test updates required and kept disassembly reading more naturally.

### DIFF
--- a/source/cfa.h
+++ b/source/cfa.h
@@ -246,21 +246,20 @@ vector<pair<BB*, BB*>> CFA<BB>::CalculateDominators(
   }
 
   // Sort by postorder index to generate a deterministic ordering of edges.
-  std::sort(out.begin(), out.end(),
-            [&idoms](const pair<bb_ptr, bb_ptr>& lhs,
-                     const pair<bb_ptr, bb_ptr>& rhs) {
-              assert(lhs.first);
-              assert(lhs.second);
-              assert(rhs.first);
-              assert(rhs.second);
-              auto lhs_first_index = idoms[lhs.first].postorder_index;
-              auto lhs_second_index = idoms[lhs.second].postorder_index;
-              auto rhs_first_index = idoms[rhs.first].postorder_index;
-              auto rhs_second_index = idoms[rhs.second].postorder_index;
-              if (lhs_first_index < rhs_first_index) return true;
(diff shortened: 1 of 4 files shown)


source/cfa.h
```diff
@@ -244,6 +244,23 @@ vector<pair<BB*, BB*>> CFA<BB>::CalculateDominators(
...
+              assert(lhs.second);
+              assert(rhs.first);
+              assert(rhs.second);
+              auto lhs_first_index = idoms[lhs.first].postorder_index;
+              auto lhs_second_index = idoms[lhs.second].postorder_index;
+              auto rhs_first_index = idoms[rhs.first].postorder_index;
+              auto rhs_second_index = idoms[rhs.second].postorder_index;
+              if (lhs_first_index < rhs_first_index) return true;
+              if (rhs_first_index < lhs_first_index) return false;
+              return lhs_second_index < rhs_second_index;
```
dnovillo: Perhaps a bit more readable to build std::pair and use the operator< template?

alan-baker:
Rebased, squashed and committed as 28199b80b7c70a03c1002c81503ed480554d9ade

PR #1732, opened 2018-07-20
Handle undef literal value in vector shuffle

### LINKED ISSUE #1731: Folding rule VectorShuffleFeedingShuffle doesn't account for undefined value
OpVectorShuffle can use 0xffffffff (4294967295) to mean an undefined value. The folding rule does not account for this possibility and can access an invalid input operand as a result.

alan-baker:
Fixes #1731

* Updated folding rules related to vector shuffle to account for the
undef literal value:
 * FoldVectorShuffleFeedingShuffle
 * FoldVectorShuffleFeedingExtract
 * FoldVectorShuffleWithConstants
* These rules would commit memory violations due to treating the undef
literal value as an accessible composite component

### DIFF
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -95,9 +95,13 @@ ConstantFoldingRule FoldVectorShuffleWithConstants() {
     }
 
     std::vector<uint32_t> ids;
+    const uint32_t undef_literal_value = 0xffffffff;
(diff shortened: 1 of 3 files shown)


alan-baker:
Committed in https://github.com/KhronosGroup/SPIRV-Tools/commit/b49f76fd62f5840d848f8891c599ef91e6fa57bb

PR #1733, opened 2018-07-25
Combine access chains transform

alan-baker:
Adds a new transformation to combine consecutive access chains. Considers all types of access chains (OpAccessChain, OpInBoundsAccessChain, OpPtrAccessChain). Currently does not handle the ArrayStride decoration (requires a type size/alignment analysis). This transform has been tested against DXC and clspv generated code.

I did not add the transform as a folding rule because in some cases an extra instruction (addition) must be generated to combine indices. It did not seem worthwhile to have duplicated code to have some aspects handled in instruction folding.

Also adds a folding rule for integer addition with zero. This rule cleans up up 0 + x index combinations that are commonly generated by clspv.

### DIFF
--- a/source/opt/combine_access_chains.cpp
+++ b/source/opt/combine_access_chains.cpp
@@ -61,12 +61,6 @@ uint32_t CombineAccessChains::GetConstantValue(
     } else {
       return constant_inst->GetU32();
     }
-  } else if (constant_inst->type()->AsInteger()->width() <= 64) {
-    if (constant_inst->type()->AsInteger()->IsSigned()) {
-      return static_cast<uint32_t>(constant_inst->GetS64());
-    } else {
-      return static_cast<uint32_t>(constant_inst->GetU64());
-    }
   } else {
     assert(false);
     return 0u;
@@ -154,9 +148,6 @@ bool CombineAccessChains::CombineIndices(Instruction* ptr_input,
         constant_mgr->GetDefiningInstruction(new_value_constant);
     new_value_id = new_value_inst->result_id();
   } else if (!type->AsStruct() || combining_element_operands) {
-    // TODO(alan-baker): handle this unlikely case.
-    if (last_index_inst->type_id() != element_inst->type_id()) return false;
-
     // Generate an addition of the two indices.
     InstructionBuilder builder(
         context(), inst,
@@ -173,7 +164,7 @@ bool CombineAccessChains::CombineIndices(Instruction* ptr_input,
   return true;
 }
 
-bool CombineAccessChains::CreateNewOperands(
+bool CombineAccessChains::CreateNewInputOperands(
     Instruction* ptr_input, Instruction* inst,
     std::vector<Operand>* new_operands) {
   // Start by copying all the input operands of the feeder access chain.
@@ -202,8 +193,63 @@ bool CombineAccessChains::CreateNewOperands(
   return true;
 }
 
-bool CombineAccessChains::IsPtrAccessChain(SpvOp opcode) {
-  return opcode == SpvOpPtrAccessChain || opcode == SpvOpInBoundsPtrAccessChain;
+bool CombineAccessChains::CombineAccessChain(Instruction* inst) {
+  assert((inst->opcode() == SpvOpPtrAccessChain ||
+          inst->opcode() == SpvOpAccessChain ||
+          inst->opcode() == SpvOpInBoundsAccessChain ||
+          inst->opcode() == SpvOpInBoundsPtrAccessChain) &&
+         "Wrong opcode. Expected an access chain.");
+
+  Instruction* ptr_input =
+      context()->get_def_use_mgr()->GetDef(inst->GetSingleWordInOperand(0));
+  if (ptr_input->opcode() != SpvOpAccessChain &&
+      ptr_input->opcode() != SpvOpInBoundsAccessChain &&
+      ptr_input->opcode() != SpvOpPtrAccessChain &&
+      ptr_input->opcode() != SpvOpInBoundsPtrAccessChain) {
+    return false;
+  }
+
+  if (Has64BitIndices(inst) || Has64BitIndices(ptr_input)) return false;
+
+  // Handles the following cases:
+  // 1. |ptr_input| is an index-less access chain. Replace the pointer
+  //    in |inst| with |ptr_input|'s pointer.
+  // 2. |inst| is a index-less access chain. Change |inst| to an
+  //    OpCopyObject.
+  // 3. |inst| is not a pointer access chain.
+  //    |inst|'s indices are appended to |ptr_input|'s indices.
+  // 4. |ptr_input| is not pointer access chain.
+  //    |inst| is a pointer access chain.
+  //    |inst|'s element operand is combined with the last index in
+  //    |ptr_input| to form a new operand.
+  // 5. |ptr_input| is a pointer access chain.
+  //    Like the above scenario, |inst|'s element operand is combined
+  //    with |ptr_input|'s last index. This results is either a
+  //    combined element operand or combined regular index.
+
+  // TODO(alan-baker): Support this properly. Requires analyzing the
+  // size/alignment of the type and converting the stride into an element
+  // index.
+  uint32_t array_stride = GetArrayStride(ptr_input);
+  if (array_stride != 0) return false;
+
+  if (ptr_input->NumInOperands() == 1) {
+    // The input is effectively a no-op.
+    inst->SetInOperand(0, {ptr_input->GetSingleWordInOperand(0)});
+  } else if (inst->NumInOperands() == 1) {
+    // |inst| is a no-op, change it to a copy. Instruction simplification will
+    // clean it up.
+    inst->SetOpcode(SpvOpCopyObject);
+  } else {
+    std::vector<Operand> new_operands;
+    if (!CreateNewInputOperands(ptr_input, inst, &new_operands)) return false;
+
+    // Update the instruction.
+    inst->SetOpcode(UpdateOpcode(inst->opcode(), ptr_input->opcode()));
+    inst->SetInOperands(std::move(new_operands));
+    context()->AnalyzeUses(inst);
+  }
+  return true;
 }
 
 SpvOp CombineAccessChains::UpdateOpcode(SpvOp base_opcode, SpvOp input_opcode) {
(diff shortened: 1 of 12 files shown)


dneto0:
>  Also adds a folding rule for integer addition with zero. This rule cleans up up 0 + x index combinations that are commonly generated by clspv.

I think I just got rid of those.   :-)

But certainly that's no harm.

source/opt/combine_access_chains.h
```diff
@@ -0,0 +1,70 @@
...
+  bool CombineAccessChain(Instruction* inst);
+
+  // Returns the value of |constant_inst| as a uint32_t.
+  uint32_t GetConstantValue(const analysis::Constant* constant_inst);
+
+  // Returns the array stride of |inst|'s type.
+  uint32_t GetArrayStride(const Instruction* inst);
+
+  // Returns the type by resolving the index operands |inst|.
+  const analysis::Type* GetIndexedType(Instruction* inst);
```
dneto0: This is confusing without context.  Is inst a kind of an access chain inst?

source/opt/combine_access_chains.h
```diff
@@ -0,0 +1,70 @@
...
+    return IRContext::kAnalysisDefUse |
+           IRContext::kAnalysisInstrToBlockMapping |
+           IRContext::kAnalysisDecorations | IRContext::kAnalysisCombinators |
+           IRContext::kAnalysisCFG | IRContext::kAnalysisDominatorAnalysis |
+           IRContext::kAnalysisNameMap;
+  }
+
+ private:
+  // Combine access chains in |function|. Blocks are processed in reverse
+  // post-order.
```
dneto0: Returns true if the module was modified?

source/opt/combine_access_chains.h
```diff
@@ -0,0 +1,70 @@
...
+  // Returns the array stride of |inst|'s type.
+  uint32_t GetArrayStride(const Instruction* inst);
+
+  // Returns the type by resolving the index operands |inst|.
+  const analysis::Type* GetIndexedType(Instruction* inst);
+
+  // Populates |new_operands| with the operands for the combined access chain.
+  // Returns false if the access chains cannot be combined.
+  bool CreateNewOperands(Instruction* ptr_input, Instruction* inst,
+                         std::vector<Operand>* new_operands);
```
dneto0: Are these new_operands just the index operands? Does it include result type, or base pointer?

source/opt/folding_rules.cpp
```diff
@@ -1907,6 +1907,27 @@ FoldingRule RedundantFMix() {
...
+    assert(inst->opcode() == SpvOpIAdd && "Wrong opcode. Should be OpIAdd.");
+
+    uint32_t operand = std::numeric_limits<uint32_t>::max();
+    if (constants[0] && constants[0]->IsZero())
+      operand = inst->GetSingleWordInOperand(1);
+    else if (constants[1] && constants[1]->IsZero())
+      operand = inst->GetSingleWordInOperand(0);
+
+    if (operand != std::numeric_limits<uint32_t>::max()) {
+      inst->SetOpcode(SpvOpCopyObject);
```
dneto0: OpCopyObject requires result type to match operand type.  But OpIAdd can be used with mixed signedness operands.
If the types are different, use an OpBitcast instead.

And then if we want to be thorough we might want to have folding rules for bitcast-feeding-X where X has flexibility on the type of its operands. But that can come later.

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+    modified |= ProcessFunction(function);
+  }
+
+  return (modified ? Status::SuccessWithChange : Status::SuccessWithoutChange);
+}
+
+bool CombineAccessChains::ProcessFunction(Function& function) {
+  bool modified = false;
+
+  cfg()->ForEachBlockInReversePostOrder(
```
dneto0: Wow. I didn't know we had this.  :-)

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+      function.entry().get(), [&modified, this](BasicBlock* block) {
+        block->ForEachInst([&modified, this](Instruction* inst) {
+          switch (inst->opcode()) {
+            case SpvOpAccessChain:
+            case SpvOpInBoundsAccessChain:
+            case SpvOpPtrAccessChain:
```
dneto0: Also OpInBoundsPtrAccessChain  ?

source/opt/combine_access_chains.h
```diff
@@ -0,0 +1,70 @@
...
+  // Combine access chains in |function|. Blocks are processed in reverse
+  // post-order.
+  bool ProcessFunction(Function& function);
+
+  // Combines an access chain (normal, in bounds or pointer) |inst| if its base
+  // pointer is another access chain. Returns true if the access chain was
+  // modified.
+  bool CombineAccessChain(Instruction* inst);
+
+  // Returns the value of |constant_inst| as a uint32_t.
```
dneto0: If the constant is wider than 32 bits than we just take the 32LSB

source/opt/combine_access_chains.h
```diff
@@ -0,0 +1,70 @@
...
+           IRContext::kAnalysisCFG | IRContext::kAnalysisDominatorAnalysis |
+           IRContext::kAnalysisNameMap;
+  }
+
+ private:
+  // Combine access chains in |function|. Blocks are processed in reverse
+  // post-order.
+  bool ProcessFunction(Function& function);
+
+  // Combines an access chain (normal, in bounds or pointer) |inst| if its base
```
dneto0: FYI.  Style guide tells us to implement methods in declaration order.  This one is out of order.
Normally I don't notice these things, so I'm not really pushing to change what you have.

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+      ptr_input->opcode() != SpvOpInBoundsAccessChain &&
+      ptr_input->opcode() != SpvOpPtrAccessChain) {
+    return false;
+  }
+
+  // Handles the following cases:
+  // 1. |inst| is not a pointer access chain.
+  //    |inst|'s indices are appended to |ptr_input|'s indices.
+  // 2. |ptr_input| is not pointer access chain.
+  //    |inst|'s element operand is combined with the last index in
```
dneto0: In 2. you should be more clear to state that |inst| *is* a ptr access chain.  (It's not clear that 2 is an "else" clause of 1.)  
It's kind of implied by calling out inst's "element" operand (which exists in PtrAccessChain but not AccessChain), but that's easy to miss.

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+  //    |ptr_input| to form a new operand.
+  // 3. |ptr_input| is a pointer access chain.
+  //    Like the above scenario, |inst|'s element operand is combined
+  //    with |ptr_input|'s last index. This results is either a
+  //    combined element operand or combined regular index.
```
dneto0: FYI.  Access chains can have 0 indices.  Let's make sure that case is covered. (I haven't looked yet)

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+  // size/alignment of the type and converting the stride into an element
+  // index.
+  uint32_t array_stride = GetArrayStride(ptr_input);
+  if (array_stride != 0) return false;
+
+  std::vector<Operand> new_operands;
+  if (!CreateNewOperands(ptr_input, inst, &new_operands)) return false;
+
+  // Update the instruction.
+  if (ptr_input->opcode() == SpvOpInBoundsAccessChain &&
```
dneto0: This looks asymmetric enough to break down when OpInBoundsPtrAccessChain is supported.

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+      inst->opcode() != SpvOpInBoundsAccessChain) {
+    // Cannot guarantee the pointer is in bounds anymore.
+    inst->SetOpcode(SpvOpAccessChain);
+  } else {
+    inst->SetOpcode(ptr_input->opcode());
+  }
+  inst->SetInOperands(std::move(new_operands));
+  context()->AnalyzeUses(inst);
```
dneto0: What about analyzing the uses of ptr_inst  ?  It loses a use.

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+  // Deal with the last index of the feeder access chain.
+  if (inst->opcode() == SpvOpPtrAccessChain) {
+    // The last index of the feeder should be combined with the element operand
+    // of |inst|.
+    if (!CombineIndices(ptr_input, inst, new_operands)) return false;
+  } else {
+    // The indices aren't being combined so now add the last index operand of
+    // |ptr_input|.
+    new_operands->push_back(
```
dneto0: ptr_input might have 0  in-operands. (gotcha)

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+    InstructionBuilder builder(
+        context(), inst,
+        IRContext::kAnalysisDefUse | IRContext::kAnalysisInstrToBlockMapping);
+    Instruction* addition = builder.AddIAdd(last_index_inst->type_id(),
+                                            last_index_inst->result_id(),
+                                            element_inst->result_id());
+    new_value_id = addition->result_id();
+  } else {
+    // Indexing into structs must be constant, so bail out here.
+    return false;
```
dneto0: This is defensive, right?   A valid module would never get here, I think.?

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+  // operand of the PtrAccessChain (|inst|).
+  const bool combining_element_operands =
+      inst->opcode() == SpvOpPtrAccessChain &&
+      ptr_input->opcode() == SpvOpPtrAccessChain &&
+      ptr_input->NumInOperands() == 2;
+  uint32_t new_value_id = 0;
+  const analysis::Type* type = GetIndexedType(ptr_input);
+  if (last_index_constant && element_constant) {
+    // Combine the constants.
+    uint32_t new_value = GetConstantValue(last_index_constant) +
```
dneto0: If the index was 64-bits, I'd rather we not do this optimization at all.  We lose bits of the index.

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,253 @@
...
+                         GetConstantValue(element_constant);
+    const analysis::Constant* new_value_constant =
+        constant_mgr->GetConstant(last_index_constant->type(), {new_value});
+    Instruction* new_value_inst =
+        constant_mgr->GetDefiningInstruction(new_value_constant);
+    new_value_id = new_value_inst->result_id();
+  } else if (!type->AsStruct() || combining_element_operands) {
+    // TODO(alan-baker): handle this unlikely case.
+    if (last_index_inst->type_id() != element_inst->type_id()) return false;
```
dneto0: Actually, the signedness semantics depends on the consuming SPIR-V instruction, not on the type of the value.
(It's that subtle rule...)

So I think you don't have to bail here.

source/opt/combine_access_chains.h
alan-baker: Yes. I'll clarify.

source/opt/combine_access_chains.h
alan-baker: Yes, will change.

source/opt/combine_access_chains.h
alan-baker: These the input operands. Includes the base pointer, but not the result type or result id.

source/opt/folding_rules.cpp
alan-baker: Thanks, I didn't realize that.

source/opt/combine_access_chains.cpp
alan-baker: I've added support, but not pushed yet. Will come with the rebase.

source/opt/combine_access_chains.cpp
alan-baker: Thanks, will add some tests.

source/opt/combine_access_chains.cpp
alan-baker: AnalyzeUses is analyzing the operands of |inst|. This call should fix everything up, will double check.

source/opt/combine_access_chains.cpp
alan-baker: Not entirely. It covers the case where I have:
```
%gep = OpAccessChain %ptr <indices to a struct>
%ptr_ptr = OpPtrAccessChain %gep %non_const_element <indices>
```
The slide can't be converted properly.

source/opt/combine_access_chains.cpp
alan-baker: Sure.

source/opt/combine_access_chains.cpp
alan-baker: I re-factored this to use a helper.

source/opt/combine_access_chains.cpp
dneto0: Resolved.

source/opt/combine_access_chains.cpp
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  return input_opcode;
+}
+
+bool CombineAccessChains::IsPtrAccessChain(SpvOp opcode) {
+  return opcode == SpvOpPtrAccessChain || opcode == SpvOpInBoundsPtrAccessChain;
+}
+
+bool CombineAccessChains::Has64BitIndices(Instruction* inst) {
```
dneto0: Really this is HasAnyNon32BitIndices.  But no matter.

test/opt/fold_test.cpp
```diff
@@ -3785,6 +3873,36 @@ TEST_P(MatchingInstructionFoldingTest, Case) {
   }
 }
 
+INSTANTIATE_TEST_CASE_P(RedundantIntegerMatching, MatchingInstructionFoldingTest,
+::testing::Values(
+    // Test case 0: Fold 0 + n (change sign)
+    InstructionFoldingCase<bool>(
+        Header() +
+            "; CHECK: [[uint:%\\w+]] = OpTypeInt 32 0\n" +
+            "; CHECK: %2 = OpBitcast [[uint]] %3\n" +
```
dneto0: sweet

alan-baker:
Rebased, squashed and committed as 755e5c94207ede680cf5f1b84626f20e3a24524f

PR #1735, opened 2018-07-26
Add API to create passes out of a list of command-line flags.

dnovillo:
This re-implements the -Oconfig=<file> flag to use a new API that takes
a list of command-line flags representing optimization passes.

This moves the processing of flags that create new optimization passes
out of spirv-opt and into the library API.  Useful for other tools that
want to incorporate a facility similar to -Oconfig.

The main changes are:

1- Add a new public function Optimizer::RegisterPassesFromFlags. This
   takes a vector of strings.  Each string is assumed to have the form
   '--pass_name[=pass_args]'.  It creates and registers into the pass
   manager all the passes specified in the vector.  Each pass is
   validated internally.  Failure to create a pass instance causes the
   function to return false and a diagnostic is emitted to the
   registered message consumer.

2- Re-implements -Oconfig in spirv-opt to use the new API.

NOTE: There are refactorings that I started doing but would've made the PR too big.  I'll send subsequent PRs.

This also needs tests for the new pass.  I'll add them shortly.


source/opt/optimizer.cpp
```diff
@@ -201,12 +206,219 @@ Optimizer& Optimizer::RegisterSizePasses() {
...
+std::pair<std::string, std::string> SplitFlagArgs(const std::string& flag) {
+  assert(flag.find("--") == 0 && "ill-formed flag. It should start with '--'.");
+  size_t ix = flag.find('=');
+  return (ix != std::string::npos)
+             ? make_pair(flag.substr(2, ix - 1), flag.substr(ix + 1))
+             : make_pair(flag, std::string());
+}
+}  // namespace
+
+std::unique_ptr<Optimizer::PassToken> Optimizer::GetPassFromFlag(
```
alan-baker: Could this be changed to register the pass? Then we could accept -O, -Os and --legalize-hlsl in the config file.

dnovillo:
Added tests and re-worked logic to accept -O, -Os and --legalize-hlsl.  PTAL.

tools/opt/opt.cpp
```diff
@@ -505,182 +500,53 @@ OptStatus ParseFlags(int argc, const char** argv, Optimizer* optimizer,
...
-            "error: Unknown flag '%s'. Use --help for a list of valid flags\n",
-            cur_arg);
-        return {OPT_STOP, 1};
+        // Some passes used to accept the form '--pass arg', canonicalize them
+        // to '--pass=arg'.
+        pass_flags.push_back(CanonicalizeFlag(argv, argc, &argi));
+
+        // If we were requested to legalize SPIR-V generated from the HLSL
+        // front-end, skip validation.
+        if (0 == strcmp(cur_arg, "--legalize-hlsl")) *skip_validator = true;
```
alan-baker: This seems to be in a weird place. Also, this is not done if --legalize-hlsl is used in -Oconfig.

dnovillo: It needs to be there, because the flag affects the behaviour of spirv-opt.  The flag will be done out of -Oconfig because the processing of Oconfig flags simply reads the strings and passes them out to ParseFlags() recursively.

So, it doesn't matter where flags are read from.

dnovillo:
Re-tested and submitted as 99fe61e7240a6d232ea361fee7a8ce81557bd63a.

PR #1738, opened 2018-07-30
Wrap bin group in conditional for BUILD.gn

dj2:
This group needs to be wrapped in the if conditional otherwise the deps can't be found.


dj2:
Simple build change, landing TBR.

PR #1739, opened 2018-07-30
Remove unused idUsage members

dj2:
This CL removes unused parameters and members from the idUsage class.


dj2:
Abandonded. Going to try removing idUsage instead.

PR #1744, opened 2018-07-31
Make sure all instructions are in the ordered list.

dj2:
Currently, some instructions will be missing from the list of
ordered_instructions. This will cause issues due to the debug change
which passed the last instruction into subsequent passes.

This CL moves the addition to the ordered list out of the
RegisterInstruction method into AddOrderedInstruction. This method is
called first in ProcessInstruction and the CapabilitiesPass and IdPass
are updated to take an Instruction parameter.

### DIFF
--- a/source/val/validate.cpp
+++ b/source/val/validate.cpp
@@ -83,35 +83,6 @@ spv_result_t setHeader(void* user_data, spv_endianness_t endian, uint32_t magic,
   return SPV_SUCCESS;
 }
 
-// Improves diagnostic messages by collecting names of IDs
-// NOTE: This function returns void and is not involved in validation
-void DebugInstructionPass(ValidationState_t& _,
-                          const spv_parsed_instruction_t* inst) {
-  switch (inst->opcode) {
-    case SpvOpName: {
-      const uint32_t target = *(inst->words + inst->operands[0].offset);
-      const char* str =
-          reinterpret_cast<const char*>(inst->words + inst->operands[1].offset);
-      _.AssignNameToId(target, str);
-    } break;
-    case SpvOpMemberName: {
-      const uint32_t target = *(inst->words + inst->operands[0].offset);
-      const char* str =
-          reinterpret_cast<const char*>(inst->words + inst->operands[2].offset);
-      _.AssignNameToId(target, str);
-    } break;
-    case SpvOpSourceContinued:
-    case SpvOpSource:
-    case SpvOpSourceExtension:
-    case SpvOpString:
-    case SpvOpLine:
-    case SpvOpNoLine:
-
-    default:
-      break;
-  }
-}
-
 // Parses OpExtension instruction and registers extension.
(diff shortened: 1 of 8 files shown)


source/val/validate.h
```diff
@@ -112,7 +112,7 @@ spv_result_t ModuleLayoutPass(ValidationState_t& _, const Instruction* inst);
 spv_result_t CfgPass(ValidationState_t& _, const Instruction* inst);
 
 /// Performs Id and SSA validation of a module
-spv_result_t IdPass(ValidationState_t& _, const spv_parsed_instruction_t* inst);
+spv_result_t IdPass(ValidationState_t& _, Instruction* inst);
```
dj2: Note, non-const because when we call RegisterInstruction it needs a non-const instruction to add to all_definitions. Changing all_definitions to const .... ripples ....

source/val/validate.cpp
```diff
@@ -83,35 +83,6 @@ spv_result_t setHeader(void* user_data, spv_endianness_t endian, uint32_t magic,
   return SPV_SUCCESS;
 }
 
-// Improves diagnostic messages by collecting names of IDs
-// NOTE: This function returns void and is not involved in validation
-void DebugInstructionPass(ValidationState_t& _,
```
dj2: I moved this to ValidationState as it makes more sense over there.

source/val/validate.cpp
```diff
@@ -166,14 +137,11 @@ spv_result_t ProcessInstruction(void* user_data,
...
   }
 
-  DebugInstructionPass(_, inst);
-  if (auto error = CapabilityPass(_, inst)) return error;
-  // The IdPass check registers instructions and, therefore, must be called
-  // before any instruction lookups are performed.
-  if (auto error = IdPass(_, inst)) return error;
-
-  const Instruction* instruction = &(_.ordered_instructions().back());
+  auto* instruction = _.AddOrderedInstruction(inst);
```
dneto0: Or we could put this at line 122 beside the _.increment_instruction_count()....  but that also looks like the counter could be replaced with a size() call on the ordered instructions list.    (Likely the counter came first, and saving the instructions came much later.)

No change required for this review.

source/val/validate_id.cpp
```diff
@@ -2191,14 +2189,12 @@ spv_result_t IdPass(ValidationState_t& _,
...
-    if (SPV_SUCCESS != ret) {
-      return ret;
-    }
-  }
-  if (result_id) {
-    _.RemoveIfForwardDeclared(result_id);
+    if (SPV_SUCCESS != ret) return ret;
   }
-  _.RegisterInstruction(*inst);
+  if (result_id) _.RemoveIfForwardDeclared(result_id);
```
dneto0: I personally go back and forth between braces or no braces around a single short statement.  :-)

dj2: I prefer no braces, but I really wish we formatted it onto two lines.

PR #1745, opened 2018-07-31
Wrap entire timer.cpp in SPIRV_TIMER_ENABLED.

dj2:
This CL moves the SPIRV_TIMER_ENABLED preprocesser guard to encompass
the includes along with the source. Currently we will try to pull in
sys/resource.h on machines which may not have the file available and the
build will fail. If we don't need timers, then we don't need the
includes as well.

### DIFF
--- a/source/util/timer.cpp
+++ b/source/util/timer.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if defined(SPIRV_TIMER_ENABLED)
+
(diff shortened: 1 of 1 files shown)


dj2:
This is a speculative fix for https://logs.chromium.org/logs/chromium/buildbucket/cr-buildbucket.appspot.com/8939555512022232032/+/steps/compile__with_patch_/0/stdout

dneto0:
Rebased and pushed into master as  89901a8a4

PR #1749, opened 2018-07-31
Change ValidationState::diag to accept an Instruction.

dj2:
This CL changes the signature of diag() to accept an Instruction instead
of the instructions position.

A deprecated variant that accepts the position is available but will be
removed in the near future.

### DIFF
--- a/source/val/validation_state.h
+++ b/source/val/validation_state.h
@@ -508,7 +508,7 @@ class ValidationState_t {
  private:
   ValidationState_t(const ValidationState_t&);
 
-  /// Deprecated. Use the Instruction variant instead of this one.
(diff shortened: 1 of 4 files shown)


source/val/validation_state.h
```diff
@@ -154,7 +154,7 @@ class ValidationState_t {
   bool IsOpcodeInCurrentLayoutSection(SpvOp op);
 
   DiagnosticStream diag(spv_result_t error_code) const;
```
dj2: My goal is to remove this overload, and then we can remove the int overload I added in this CL and just have the Instruction variant. This removes the cases where we set the line based on where the parser is, vs setting it based on the instruction.

Issue #1176, opened 2018-01-05
Consider banning use of std::queue due to potential slowness with MSVC


dneto0:
By default std::queue uses a std::deque as the underlying container.

Some past experience with MSVC runtimes have shown that block sizes for std::deqeue could be quite small and you could end up jumping around memory a lot.

Need to investigate!

The optimizer and compressor use std::deque, in possibly performance-critical parts of the code.

dneto0:
Cc @AWoloszyn

dneto0:
cc: @dj2 

Same applies to std::deque as well.

PR #1755, opened 2018-08-01
Remove std::deque in favour of std::vector.

dj2:
This CL removes the two deque's from ValidationState and converts them
into std::vectors. In order to maintain the stability of instructions we
walk over the binary and counter the instructions and functions in the
ValidationState constructor and reserve the required number of items in
the module_functions_ and ordered_instructions_ vectors.

Issue #1176.

### DIFF
--- a/source/comp/markv_codec.cpp
+++ b/source/comp/markv_codec.cpp
@@ -49,6 +49,7 @@
 #include "markv_model.h"
 #include "opcode.h"
 #include "operand.h"
+#include "source/assembly_grammar.h"
 #include "spirv-tools/libspirv.h"
 #include "spirv_endian.h"
 #include "spirv_validator_options.h"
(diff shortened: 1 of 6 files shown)


source/val/validation_state.cpp
```diff
@@ -187,6 +191,21 @@ ValidationState_t::ValidationState_t(const spv_const_context ctx,
     default:
       break;
   }
+
+  // Only attempt to count if we have words, otherwise let the other validation
+  // fail and generate an error.
+  if (num_words > 0) {
```
dj2: I do this work in the constructor because ValidationState is used from both the validator and the stats aggregator. We either do it here or we add methods to do it that we have to remember to call from each place we want to use ValidationState in the future.

source/comp/markv_codec.cpp
```diff
@@ -49,6 +49,7 @@
 #include "markv_model.h"
 #include "opcode.h"
```
dneto0: For another time:  It seems these should be source/opcode.h and source/operand.h etc...

dj2: Yea, there are a bunch of includes which should be fixed. It's on my list, heh.

PR #1756, opened 2018-08-01
Remove using std::<foo> statements.

dj2:
Many of the files have using std::<foo> statements in them, but then the
use of <foo> will be inconsistently std::<foo> or <foo> scattered
through the file. This CL removes all of the using statements and
updates the code to have the required std:: prefix.


dj2:
Thoughts? This seems clearer to me, but what do y'all think?

dj2:
@alan-baker @s-perron thoughts on this cl?

test/val/val_layout_test.cpp
```diff
@@ -136,53 +126,54 @@ INSTANTIATE_TEST_CASE_P(InstructionsOrder,
     // are not "All" in the below.
     //
     //                                  | Instruction                | Line(s) valid          | Lines to compile
```
s-perron: The headers for this should should be realigned.

dj2: done

dj2:
Given @dneto0 is on the fence, @s-perron +1'd and @alan-baker is fine either way, I'm going to go ahead and merge this change.

PR #1758, opened 2018-08-01
Update diag() use in validate_builtin.

dj2:
This CL updates the calls to diag() in vlidate_builtings to provide the
instruction.


dj2:
This one was a bit tricker due to the callback functions, please give it a good look.

PR #1760, opened 2018-08-01
Update diag() calls in validate_cfg.

dj2:
This CL updates the diag() calls in validate_cfg to provide the
associated instruction. This fixes a couple places where we output the
last line of the file instead of the instruction as the disassembly.

### DIFF
--- a/source/val/validate_cfg.cpp
+++ b/source/val/validate_cfg.cpp
@@ -193,7 +193,7 @@ spv_result_t FindCaseFallThrough(
           continue;
         }
 
-        return _.diag(SPV_ERROR_INVALID_CFG)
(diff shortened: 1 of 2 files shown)


source/val/validate_cfg.cpp
```diff
@@ -330,7 +330,7 @@ spv_result_t StructuredControlFlowChecks(
     uint32_t header_block;
     tie(back_edge_block, header_block) = back_edge;
     if (!function->IsBlockType(header_block, kBlockTypeLoop)) {
-      return _.diag(SPV_ERROR_INVALID_CFG)
+      return _.diag(SPV_ERROR_INVALID_CFG, _.FindDef(back_edge_block))
```
dj2: I guessed at these three, but they seem to make sense (and give a better disassembly then end of file).

PR #1764, opened 2018-08-01
Update diag() calls in validate_decorations.

dj2:
Several of the diag() calls in validate_decorations do not provide the
line number, and will output the last line in the file. This CL updates
the diag() calls to provide the instruction of interest.


source/val/validate_decorations.cpp
```diff
@@ -336,7 +336,7 @@ spv_result_t checkLayout(uint32_t struct_id, const char* storage_class_str,
   auto fail = [&vstate, struct_id, storage_class_str, decoration_str,
                blockRules](uint32_t member_idx) -> DiagnosticStream {
     DiagnosticStream ds =
-        std::move(vstate.diag(SPV_ERROR_INVALID_ID)
+        std::move(vstate.diag(SPV_ERROR_INVALID_ID, vstate.FindDef(struct_id))
```
dj2: I think the instructions in this file make sense?

PR #1768, opened 2018-08-01
Update diag() calls in validate_instruction.

dj2:
This CL updates validate_instruction to pass the Instruction to diag().

### DIFF
--- a/source/val/validate_instruction.cpp
+++ b/source/val/validate_instruction.cpp
@@ -55,16 +55,6 @@ std::string ToString(const CapabilitySet& capabilities,
   return ss.str();
 }
 
-// Reports a missing-capability error to _'s diagnostic stream and returns
-// SPV_ERROR_INVALID_CAPABILITY.
-spv_result_t CapabilityError(const ValidationState_t& _, size_t which_operand,
-                             SpvOp opcode,
-                             const std::string& required_capabilities) {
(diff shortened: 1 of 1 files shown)


source/val/validate_instruction.cpp
```diff
@@ -55,16 +55,6 @@ std::string ToString(const CapabilitySet& capabilities,
   return ss.str();
 }
 
-// Reports a missing-capability error to _'s diagnostic stream and returns
-// SPV_ERROR_INVALID_CAPABILITY.
-spv_result_t CapabilityError(const ValidationState_t& _, size_t which_operand,
```
dj2: I moved this inline as it was only used in one place.

source/val/validate_instruction.cpp
```diff
@@ -100,7 +90,8 @@ CapabilitySet EnablingCapabilitiesForOp(const ValidationState_t& state,
 // in the module.  Otherwise issues an error message and returns
 // SPV_ERROR_INVALID_CAPABILITY.
 spv_result_t CheckRequiredCapabilities(const ValidationState_t& state,
-                                       SpvOp opcode, size_t which_operand,
+                                       const Instruction* inst,
```
dj2: The opcode changed to instruction here so we have access.

PR #1776, opened 2018-08-01
Remove diag() overloads.

dj2:
This CL removes the two diag() overloads and leaves only the version
which accepts an Instruction. This is safer as we never use the
implicit location from the validation state.

### DIFF
--- a/source/val/validate_id.cpp
+++ b/source/val/validate_id.cpp
@@ -2086,7 +2086,7 @@ spv_result_t CheckIdDefinitionDominateUse(const ValidationState_t& _) {
             if (use->opcode() == SpvOpPhi) {
               phi_instructions.insert(use);
             } else if (!block->dominates(*use->block())) {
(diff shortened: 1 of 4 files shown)


source/val/validate_id.cpp
```diff
@@ -2086,7 +2086,7 @@ spv_result_t CheckIdDefinitionDominateUse(const ValidationState_t& _) {
             if (use->opcode() == SpvOpPhi) {
               phi_instructions.insert(use);
             } else if (!block->dominates(*use->block())) {
-              return _.diag(SPV_ERROR_INVALID_ID)
+              return _.diag(SPV_ERROR_INVALID_ID, use_block->label())
```
dj2: This one may have conflicts with the work @alan-baker is doing to split up validate_id. I just wanted to get it up now so that the patch chain was complete.

dj2:
Heh, and of course it won't build until all the others have landed.

PR #1780, opened 2018-08-01
Split memory validation

alan-baker:
Moves memory instruction validation that occurred in validate_id.cpp into a separate pass.

* Tests were not modified.
* Some minor disassembly changes
* Some minor formatting changes


source/CMakeLists.txt
```diff
@@ -303,6 +303,7 @@ set(SPIRV_SOURCES
   ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_layout.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_literals.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_logicals.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_memory.cpp
```
dj2: Can you add this to BUILD.gn as well?

source/val/validate_memory.cpp
```diff
@@ -0,0 +1,599 @@
...
+
+}  // namespace
+
+spv_result_t ValidateMemoryInstructions(ValidationState_t& _) {
+  for (auto& inst : _.ordered_instructions()) {
+    switch (inst.opcode()) {
+      case SpvOpVariable:
+        if (auto error = ValidateVariable(_, inst)) return error;
+        break;
+      case SpvOpImageTexelPointer:
```
dj2: Move this down to 582 to keep all the empty ones together?

source/CMakeLists.txt
alan-baker: done

source/val/validate_memory.cpp
alan-baker: done

alan-baker:
Rebased, squashed and submitted as d49bedcaa68a907e2b201bc372d69256e50f2ab5

PR #1784, opened 2018-08-02
Remove ValidateInstructionAndUpdateValidationState

dj2:
This CL changes the stats aggregator to use
ValidateBinaryAndKeepValidationState to process the binary. This means
we can remove ValidateInstructionAndUpdateValidationState which expects
to be able to call ProcessInstruction in the validate anonymous
namespace. This decouples the stats aggregator from how validation
processes the binary.


source/spirv_stats.cpp
```diff
@@ -149,79 +130,74 @@ class StatsAggregator {
...
+    if (inst->opcode() != SpvOpExtension) return;
+    const std::string extension = GetExtensionString(&inst->c_inst());
     ++stats_->extension_hist[extension];
   }
 
   // Collects OpCode statistics.
-  void ProcessOpcode() {
-    auto inst_it = vstate_->ordered_instructions().rbegin();
-    const SpvOp opcode = inst_it->opcode();
+  void ProcessOpcode(const val::Instruction* inst, size_t idx) {
```
dj2: Please give this method a good look as it was doing things with iterators that I converted to indexes ...

PR #1786, opened 2018-08-02
Simplify validation ProcessInstruction

dj2:
This CL moves most of the logic out of validation ProcessInstruction and
groups it into validate. This places all of the validation logic in the
same place making it clearer what is running.

The Instruction class is changed to allow setting the function and block
after creation.

### DIFF
--- a/source/val/instruction.cpp
+++ b/source/val/instruction.cpp
@@ -19,33 +19,28 @@
 namespace spvtools {
 namespace val {
 
-#define OPERATOR(OP)                                                 \
-  bool operator OP(const Instruction& lhs, const Instruction& rhs) { \
-    return lhs.id() OP rhs.id();                                     \
-  }                                                                  \
-  bool operator OP(const Instruction& lhs, uint32_t rhs) {           \
-    return lhs.id() OP rhs;                                          \
-  }
-
-OPERATOR(<)
-OPERATOR(==)
-#undef OPERATOR
-
-Instruction::Instruction(const spv_parsed_instruction_t* inst,
-                         Function* defining_function,
-                         BasicBlock* defining_block)
+Instruction::Instruction(const spv_parsed_instruction_t* inst)
     : words_(inst->words, inst->words + inst->num_words),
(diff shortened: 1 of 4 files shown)


source/val/validate.cpp
```diff
@@ -237,7 +244,64 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
   }
 
   for (size_t i = 0; i < vstate->ordered_instructions().size(); ++i) {
-    const auto& instruction = vstate->ordered_instructions()[i];
+    auto& instruction = vstate->ordered_instructions()[i];
+
+    {
```
dj2: This is yucky, but changing ordered_instruction to provide a non-const instruction has ripples....

source/val/validation_state.cpp
```diff
@@ -455,17 +455,7 @@ spv_result_t ValidationState_t::RegisterFunctionEnd() {
 
 Instruction* ValidationState_t::AddOrderedInstruction(
     const spv_parsed_instruction_t* inst) {
-  if (in_function_body()) {
```
dj2: This is moved into the yucky bit above so that in_function_body() and in_block() are correctly defined by ModuleLayout.

source/val/instruction.cpp
```diff
@@ -19,34 +19,27 @@
...
-      block_(defining_block),
-      uses_() {}
+             inst->num_operands}) {}
 
 void Instruction::RegisterUse(const Instruction* inst, uint32_t index) {
   uses_.push_back(std::make_pair(inst, index));
 }
 
+bool operator<(const Instruction& lhs, const Instruction& rhs) {
+  return lhs.id() < rhs.id();
```
alan-baker: I see where this is a port from, but this doesn't seem correct. Do you know where this is used? Only using result id will cause all stores to appear equal for example.

dj2: I have no idea, I can do a followup CL to attempt to remove them and see what happens

alan-baker: SGTM

PR #1790, opened 2018-08-02
Validation within function body when doing a FunctionCall.

dj2:
When validating a FunctionCall we can trigger an assert if we are not
currently within a function body. This CL adds verification that we are
within a function before attempting to add a function call.

Issue 1789.


dj2:
This is the same PR as https://github.com/KhronosGroup/SPIRV-Tools/pull/1623 just rebased to master, didn't know how to re-open a PR.

I didn't move where the check was happening as it's still too late. We have to have the check before we do the AddFunctionCallTarget call as that will call current_function() and fail with a nullptr.

We could, possibly, add a validation run in before we do this work but that was a bigger change.

PR #1792, opened 2018-08-02
Script to fixup fuzzer results

dj2:


dj2:
It's still not great, but gets it from an error due to wrong size to:

; SPIR-V
; Version: 1.0
; Generator: Unknown(463); 0
; Bound: 26
; Schema: 33558784
     %196670 = OpFunctionCall %14679817 %4294965777 %66062
error: 11: Invalid instruction OpNop starting at word 10: expected no more operands after 1 words, but stated word count is 14.

PR #1795, opened 2018-08-03
Cleanup includes.

dj2:
This Cl cleans up the include paths to be realtive to the top level
directory. Various include-what-you-use fixes have been added.


dj2:
Looks like the android makefile is right and cmake is wrong, but I don't know how ot remove the include/ directory from the -I lines in cmake? I'll fixup the issues found by android when I get a chance.

dj2:
Actually, I think going the other way makes sense and leaving include/ in the -I's.  Will go that route instead.

dj2:
Done, let's see if that works better on Android.

PR #1796, opened 2018-08-03
Fixup Wno-self-assign warning

dj2:
The warning should only be applied for Clang builds, it was accidentally
also being included for GCC builds.


dnovillo:
Thanks!

PR #1797, opened 2018-08-03
Gn build overridables

Kangz:
PTAL, this adds overridable variables to the GN build to make it easier to integrate SPIRV-Tools in different projects by removing hardcoded paths to dependencies.

Potential users of this:
 - Chromium 
 - Dawn (our WebGPU implementation, why I'm doing this now)
 - ANGLE
 - Skia

Each commit should compile on its own, but the whole thing could be squashed too.


BUILD.gn
```diff
@@ -12,14 +12,12 @@
...
+
 import("//testing/test.gni")
 import("//build_overrides/build.gni")
 
-if (build_with_chromium) {
-  spirv_headers = "//third_party/spirv-headers/src"
-} else {
-  spirv_headers = "external/spirv-headers"
-}
+spirv_headers = spirv_tools_spirv_headers_dir
```
dneto0: Well that looks nicer!

BUILD.gn
```diff
@@ -12,14 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import("//build_overrides/spirv_tools.gni")
```
dneto0: I don't know much about GN. 
How does an external project like Dawn override the settings that we import here?

Kangz: It is purely a convention starteb by Chromium's `build_overrides/build.gni`: each project has a gni in "build_overrides" and does `import("//build_overrides/foo.gni")` first thing in its BUILD.gn. Since `//` is the root of the current GN execution, this will either import SPIRV-Tools's version if in SPIRV-Tools, or the one in `<dawn_source>/build_overrides` if we are in Dawn's source.

dneto0: Ah.  It's like duck-typing for builds.
As long as all the build-overrides file sets the right things for SPIRV-Tools then it all works out. Very nice.

dneto0:
Rebased, squashed, and pushed into master as 2d9a325264e3

Kangz:
Thank you!

PR #1798, opened 2018-08-03
Refactor where opcodes are validated

alan-baker:
* Replaced uses in opcode validation of current_function()
* Added non-const accessor to function lookup in ValidationState_t
* Updated a couple bad tests due to check reordering

Moved where opcode validation happens. It will be easier to refactor checks out of idUsage as a result of this change. Fixed passes to not rely on ValidationState_t::current_function().

Now the first loop through ordered instructions builds the infrastructure. The new later loop does individual opcode validation.

### DIFF
--- a/source/val/validate.cpp
+++ b/source/val/validate.cpp
@@ -242,9 +242,7 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
     return error;
   }
 
-  for (size_t i = 0; i < vstate->ordered_instructions().size(); ++i) {
-    auto& instruction = vstate->ordered_instructions()[i];
-
+  for (auto& instruction : vstate->ordered_instructions()) {
     {
       // In order to do this work outside of Process Instruction we need to be
       // able to, briefly, de-const the instruction.
--- a/source/val/validate.cpp
+++ b/source/val/validate.cpp
@@ -286,34 +286,12 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
 
       if (auto error = IdPass(*vstate, inst)) return error;
     }
-
     if (auto error = CapabilityPass(*vstate, &instruction)) return error;
     if (auto error = DataRulesPass(*vstate, &instruction)) return error;
     if (auto error = ModuleLayoutPass(*vstate, &instruction)) return error;
     if (auto error = CfgPass(*vstate, &instruction)) return error;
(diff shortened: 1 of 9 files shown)


test/val/val_image_test.cpp
```diff
@@ -339,19 +339,28 @@ OpFunctionEnd)";
   return ss.str();
 }
 
-std::string GetShaderHeader(
-    const std::string& capabilities_and_extensions = "") {
+std::string GetShaderHeader(const std::string& capabilities_and_extensions = "",
```
alan-baker: All the changes in this file were to fix SPIR-V having an unresolved forward reference (the entry point). Moving the image checks later was causing other failures.

source/val/validate.cpp
```diff
@@ -286,34 +286,12 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
 
       if (auto error = IdPass(*vstate, inst)) return error;
     }
-
```
dj2: I think line 245 and 246 can change to:

for (const auto& instruction : vstate->ordered_instructions()) {

since the pass that needed i is moved to the second batch.

source/val/validate.cpp
```diff
@@ -337,6 +315,32 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
   // for() above as it loops over all ordered_instructions internally.
   if (auto error = ValidateBuiltIns(*vstate)) return error;
 
+  // Validate individual opcodes.
+  for (size_t i = 0; i < vstate->ordered_instructions().size(); ++i) {
```
dj2: There are other things that are going to be missing like in_function_body() and in_block() I believe. Can we add a flag to validation state that those are available, set it false after the first for() exist and then assert if any of those special methods are hit in later passes?

source/val/validate.cpp
alan-baker: Will do

source/val/validate.cpp
alan-baker: Makes sense.

alan-baker: current_function() and in_function_body() look like they get set correctly. current_function() asserts that in_function_body() is true. Likewise, in_block() looks ok too.

alan-baker:
Rebased, squashed and committed as 2896b8f0e5f85a726d724d46c0641529280a7a31

PR #1799, opened 2018-08-03
Validate the input to Optimizer::Run

s-perron:
Run the validator in the optimization fuzzers.
    
    The optimizers assumes that the input to the optimizer is valid.  Since
    the fuzzers do not check that the input is valid before passing the
    spir-v to the optimizer, we are getting a few errors.
    
    The solution is to run the validator in the optimizer to validate the
    input.
    
    For the legalization passes, we need to add an extra option to the
    validator to accept certain types of variable pointers, even if the
    capability is not given.  At the same time, we changed the option
    "--legalize-hlsl" to relax the validator in the same way instead of
    turning it off.

Follow up issue #1810 was opened to improve the consistency of the message consumers.


s-perron:
I think I have most of the changes needed to run the validator in the optimizer.  All of the tests pass.  The only remaining issue is the output of the messages from the validator.  The PR in this state will change the output, and I don't think it is as understandable.

include/spirv-tools/libspirv.hpp
```diff
@@ -68,6 +68,8 @@ class Context {
 class ValidatorOptions {
  public:
   ValidatorOptions() : options_(spvValidatorOptionsCreate()) {}
+  ValidatorOptions(spv_validator_options options) : options_(options) {}
```
dj2: explicit ValidatorOptions(...)

source/libspirv.cpp
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "spirv-tools/libspirv.hpp"
+#include <iostream>
```
dj2: This should go above the libspirv.hpp header

source/libspirv.cpp
```diff
@@ -111,8 +112,15 @@ bool SpirvTools::Validate(const uint32_t* binary,
 bool SpirvTools::Validate(const uint32_t* binary, const size_t binary_size,
                           const ValidatorOptions& options) const {
   spv_const_binary_t the_binary{binary, binary_size};
-  return spvValidateWithOptions(impl_->context, options, &the_binary,
-                                nullptr) == SPV_SUCCESS;
+  spv_diagnostic diagnostic = nullptr;
```
dj2: Does spvValidateWIthOptions create the diagnostic? That's an ... odd interface

source/libspirv.cpp
```diff
@@ -111,8 +112,15 @@ bool SpirvTools::Validate(const uint32_t* binary,
...
+  bool valid = spvValidateWithOptions(impl_->context, options, &the_binary,
+                                      &diagnostic) == SPV_SUCCESS;
+  if (!valid && impl_->context->consumer) {
+    impl_->context->consumer.operator()(
```
dj2: So, this is just to feed messages back from the c api to the consumer? How did this work before?

test/cpp_interface_test.cpp
```diff
@@ -24,6 +24,15 @@ namespace {
 using ::testing::ContainerEq;
 using ::testing::HasSubstr;
 
+// Return a string that contains the minimum instructions needed to form
+// a valid module.  Other instructions can be appended to this string.
+std::string Header() {
```
dj2: Does this need to be a method? What about just having a

const char[] kHeader = R"( ... )";

test/opt/optimizer_test.cpp
```diff
@@ -25,10 +25,20 @@ namespace {
 
 using ::testing::Eq;
 
+// Return a string that contains the minimum instructions needed to form
+// a valid module.  Other instructions can be appended to this string.
+std::string Header() {
```
dj2: ditto

include/spirv-tools/libspirv.hpp
s-perron: Actually I do not need or want this one anymore, I'll undo this change, and add explicit to the default constructor.

source/libspirv.cpp
s-perron: Forgot to delete.

source/libspirv.cpp
s-perron: Yes it does, which is why you pass it a pointer to a pointer.  It is odd.

tools/opt/opt.cpp
```diff
@@ -590,28 +592,9 @@ int main(int argc, const char** argv) {
...
 
-  if (!skip_validator) {
-    // Let's do validation first.
-    spv_context context = spvContextCreate(target_env);
-    spv_diagnostic diagnostic = nullptr;
-    spv_const_binary_t binary_struct = {binary.data(), binary.size()};
-    spv_result_t error =
-        spvValidateWithOptions(context, options, &binary_struct, &diagnostic);
-    if (error) {
-      spvDiagnosticPrint(diagnostic);
```
s-perron: @dj2 This how the messages were printed before.  If there was an error, the diagnostic we passed to `spvDiagnosticPrint`, which always prints to stderr.  This is fine for code specific to spirv-opt, but I do not want to call if from public API.

source/libspirv.cpp
s-perron: See my other comment in tools/opt/opt.cpp.

include/spirv-tools/libspirv.hpp
dj2: Default constructor doesn't need it, just single argument constructors.

source/libspirv.cpp
dj2: Ugh.

test/cpp_interface_test.cpp
s-perron: Yes you are right, that is a POD, so I can be a global.  However, I kinda want an std::string to make concatenation simpler.  Since the string is not POD, I cannot have a global std::string.

tools/opt/opt.cpp
dj2: Right, makes sense. Thanks for pointing that out.

test/cpp_interface_test.cpp
dj2: Doesn't concat just work? I thought it would just turn the c-str into a std::string as needed?

s-perron:
The current version of this PR will not change the format of the output for validator errors, but will change the information in the messages from replace invalid opcode.

I think this is fine because those messages rely on OpLine debug information to be meaningful anyway, and in general the line information is not present.  Issue #1810 was opened to follow up with this and make messages more consistent.

include/spirv-tools/libspirv.hpp
```diff
@@ -180,6 +180,11 @@ class SpirvTools {
   std::unique_ptr<Impl> impl_;  // Unique pointer to implementation data.
 };
 
+// A message consumer that can be used by command line tools like spirv-opt and
+// spirv-val to display messages.
+void CLIMessageConsumer(spv_message_level_t level, const char*,
```
dj2: Can we move this into a file in tools/utils or something similar? I don't think we want this in the public API.

s-perron: Sure.  My thought was that is make is available to other so they don't have to write their own message consumer, but I don't have strong feelings about it either way.

source/CMakeLists.txt
```diff
@@ -219,6 +219,7 @@ set(SPIRV_SOURCES
   ${CMAKE_CURRENT_SOURCE_DIR}/util/bitutils.h
   ${CMAKE_CURRENT_SOURCE_DIR}/util/bit_stream.h
   ${CMAKE_CURRENT_SOURCE_DIR}/util/bit_vector.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/util/cli_consumer.h
```
dj2: Do we want this added to the general library? it's going to get shipped to folks that are just using libspirv when it's only used for the tools?

s-perron: Okay moved. Let me know if I should have changed the namespaces for the consumer as well.  I could not think of a good thing to change it to.

dj2: Awesome, thanks. I think spvtools::utils makes sense as the namespace.

PR #1801, opened 2018-08-03
Unify validation of OpCopyMemory*

### LINKED ISSUE #1800: OpCopyMemorySized validation of size operand is incorrect
The validator currently accepts OpConstant, OpSpecConstant and OpVariable. The spec says:
> Size is the number of bytes to copy. It must have a scalar integer type. If it is a constant instruction, the constant value cannot be 0. It is invalid for both the constant’s type to have Signedness of 1 and to have the sign bit set. Otherwise, as a run-time value, Size is treated as unsigned, and if its value is 0, no memory access will be made.

The check should be that the id has an integer type. This disallows variables. So it is both not permissive enough and too permissive at the same time.

alan-baker:
Fixes #1800

* Refactored duplication of code between OpCopyMemory and
OpCopyMemorySized validation
* Fixed some bugs in OpCopyMemorySized validation
* Replaced asserts with checks
* Added new tests

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -316,96 +316,86 @@ spv_result_t ValidateCopyMemory(ValidationState_t& _, const Instruction& inst) {
   }
 
   const auto target_pointer_type = _.FindDef(target->type_id());
-  assert(target_pointer_type);
-  const auto target_type = _.FindDef(target_pointer_type->words()[3]);
-  assert(target_type);
-  const auto source_pointer_type = _.FindDef(source->type_id());
-  assert(source_pointer_type);
-  const auto source_type = _.FindDef(source_pointer_type->words()[3]);
-  assert(source_type);
-  if (target_type->id() != source_type->id()) {
-    return _.diag(SPV_ERROR_INVALID_ID, &inst)
-           << "OpCopyMemory Target <id> '" << _.getIdName(source_id)
-           << "'s type does not match Source <id> '"
-           << _.getIdName(source_type->id()) << "'s type.";
-  }
-  return SPV_SUCCESS;
-}
-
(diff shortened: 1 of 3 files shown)


source/val/validate_memory.cpp
```diff
@@ -313,97 +313,87 @@ spv_result_t ValidateCopyMemory(ValidationState_t& _, const Instruction& inst) {
...
-      if (!size_type || SpvOpTypeInt != size_type->opcode()) {
-        return _.diag(SPV_ERROR_INVALID_ID, &inst)
-               << "OpCopyMemorySized Size <id> '" << _.getIdName(size_id)
-               << "'s variable type is not an integer type.";
-      }
-    } break;
-    default:
+
+  if (inst.opcode() == SpvOpCopyMemory) {
+    const auto target_type = _.FindDef(target_pointer_type->words()[3]);
```
dj2: Can these words()[3] be changed to GetOperandAs()'s?

source/val/validate_memory.cpp
```diff
@@ -313,97 +313,87 @@ spv_result_t ValidateCopyMemory(ValidationState_t& _, const Instruction& inst) {
...
+    }
+
+    if (target_type->id() != source_type->id()) {
+      return _.diag(SPV_ERROR_INVALID_ID, &inst)
+             << "Target <id> '" << _.getIdName(source_id)
+             << "'s type does not match Source <id> '"
+             << _.getIdName(source_type->id()) << "'s type.";
+    }
+  } else {
+    const auto size_index = 2;
```
dj2: drop size_index and inline if this is the only usage?

alan-baker:
Rebased, squashed and committed as 3a20879f4d0af1cc336ef3294d67bf199796be5d

PR #1802, opened 2018-08-03
Add testing framework for CLI tools.

### LINKED ISSUE #1778: -Oconfig should allow spaces in comment lines
#this.works

#this does not because each word after the first is passed on as a new flag

dnovillo:
This copies the testing harness from https://github.com/google/shaderc
to allow testing CLI tools.  Most of the setup is identical to the one
in shaderc, with small changes for naming and formatting.

New features needed for SPIRV-Tools include:

1. A new PlaceHolder subclass for spirv shaders.  This place holder
   calls spirv-as to convert assembly input into SPIRV bytecode. This is
   required for most tools in SPIRV-Tools.
2. A minimal testing file for testing basic functionality of spirv-opt.

I only added a very simple test for spirv-opt.  The patch was getting
too big for code review.

This PR also fixes #1778.  The fix for that bug includes tests that go in `test/tools/opt`.


dnovillo:
@antiagainst @dneto0 ping.

The PR now has all the tests I wanted to add to spirv-opt.  Tests caught a few inconsistencies in flag processing and bugs in pass naming.

I will now add a test for the fix to #1778, but I need this PR merged first.

source/opt/optimizer.cpp
```diff
@@ -427,6 +427,15 @@ bool Optimizer::RegisterPassFromFlag(const std::string& flag) {
     }
   } else if (pass_name == "loop-peeling") {
     RegisterPass(CreateLoopPeelingPass());
+  } else if (pass_name == "loop-peeling-threshold") {
+    int factor = (pass_args.size() > 0) ? atoi(pass_args.c_str()) : 0;
+    if (factor != 0) {
```
dj2: What happens if I pass -10?

dnovillo: Good catch, though different bug.  This code is simply restoring code I had inadvertently deleted.  I'll fix on a separate PR.  I don't want to change existing behaviour in this PR.

I've created #1817 to track.

dnovillo: On second thought,  the PR is fixing other bugs already.  Might as well Fix All The Bugs.

dj2:
Looks like the linux bots are validation failures now? @alan-baker any idea why this would suddenly break?

[ RUN      ] Dim/ValidateCapability.Capability/456
/tmpfs/src/github/SPIRV-Tools/test/val/val_capability_test.cpp:1647: Failure
Expected equality of these values:
  ExpectedResult(GetParam())
    Which is: 0
  ValidateInstructions(env)
    Which is: -13
target env: SPIR-V 1.3
test code:
OpCapability GroupNonUniformShuffleRelative
 OpCapability ImageBasic OpCapability Kernel OpMemoryModel Logical OpenCL  OpEntryPoint Kernel %func "compute" %voidt = OpTypeVoid %imgt = OpTypeImage %voidt 2D 0 0 0 0 Unknown %void_f = OpTypeFunction %voidt %func   = OpFunction %voidt None %void_f %label  = OpLabel           OpReturn           OpFunctionEnd 
error: 7: Operand 3 of TypeImage requires one of these capabilities: ImageMSArray 
  %3 = OpTypeImage %void 2D 0 0 0 0 Unknown

dnovillo:
@antiagainst @dneto0 This PR now fixes #1778 and #1817.  It's ready to review.

dnovillo:
Rebased, re-tested and pushed as 03000a3a384fabf97da22d709e45e860aba86be1.

PR #1803, opened 2018-08-04
Fixup warnings.

dj2:
This CL fixes the two storage class warnings and the various missing
explicit qualifiers.


dj2:
Closing causes more issues then I anticipated. Will do in chunks.

PR #1808, opened 2018-08-06
Treat clang-cl.exe differently than normal clang

shartte:
When clang is used as a MSVC frontend (clang-cl.exe) it does not behave exactly as GCC would. I.e. -Wall translates to /Wall, which is an alias for -Weverything. This causes massive C++98 compat warnings logspam, which this patch aims to fix.


shartte:
Some background info on this issue can also be found in the self-hosted clang build done by the LLVM team, since they were facing the same issue:

Their change for removing -Wall if clang-cl.exe is used:
https://reviews.llvm.org/D40603

Their script to detect clang-cl.exe:
https://github.com/llvm-mirror/llvm/blob/master/cmake/modules/CheckCompilerVersion.cmake

PR #1809, opened 2018-08-07
Split MarkV into multiple files

dj2:
This CL breaks the monolithic markv_codec file into files for the base
class, encoder, decoder and logger.


dj2:
@dneto0 @alan-baker ping

source/comp/markv.cpp
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2018 Google Inc.
```
alan-baker: The company should be Google LLC.

dj2:
```
Development/spirv-tools/SPIRV-Tools % gg "Google LLC"|wc -l
138
Development/spirv-tools/SPIRV-Tools % gg "Google Inc"|wc -l
298
```

dj2: Looks like a larger cleanup is needed here.

PR #1811, opened 2018-08-08
Split type opcode validation into separate pass

alan-baker:
Split type opcode validation into a new file. Unified it with the type unique pass.

Reordered opcode checking to match SPIR-V specification sections.


test/val/val_id_test.cpp
```diff
@@ -1037,23 +1037,6 @@ TEST_F(ValidateIdWithMessage,
                         "component count does not match Result Type <id> '4's "
                         "vector component count."));
 }
-TEST_F(ValidateIdWithMessage, OpConstantCompositeMatrixColumnTypeBad) {
```
alan-baker: This test gets caught by type checks now.

dj2: Should we convert it to check for the type error instead of removing it?

alan-baker: We do have tests separately for that so I figure it was redundant to add again. I can if you think there's value.

alan-baker:
Rebased, squashed and committed as f2a990022a4b99d11c5fd0567a7c16f67d4e3b59

PR #1813, opened 2018-08-08
Replace asserts with returns

alan-baker:
* Changes to satisfy fuzzer

### DIFF
--- a/source/parsed_operand.cpp
+++ b/source/parsed_operand.cpp
@@ -23,10 +23,12 @@ namespace spvtools {
 
 void EmitNumericLiteral(std::ostream* out, const spv_parsed_instruction_t& inst,
                         const spv_parsed_operand_t& operand) {
-  assert(operand.type == SPV_OPERAND_TYPE_LITERAL_INTEGER ||
-         operand.type == SPV_OPERAND_TYPE_TYPED_LITERAL_NUMBER);
-  assert(1 <= operand.num_words);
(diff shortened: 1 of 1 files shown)


source/parsed_operand.cpp
```diff
@@ -23,10 +23,12 @@ namespace spvtools {
 
 void EmitNumericLiteral(std::ostream* out, const spv_parsed_instruction_t& inst,
                         const spv_parsed_operand_t& operand) {
-  assert(operand.type == SPV_OPERAND_TYPE_LITERAL_INTEGER ||
-         operand.type == SPV_OPERAND_TYPE_TYPED_LITERAL_NUMBER);
-  assert(1 <= operand.num_words);
-  assert(operand.num_words <= 2);
+  if (operand.type != SPV_OPERAND_TYPE_LITERAL_INTEGER &&
+      operand.type != SPV_OPERAND_TYPE_TYPED_LITERAL_NUMBER)
+    return;
```
dj2: What does it mean for this to just return? Do we need to fail validation somewhere now? What will be output?

alan-baker: The validator actually only used this to produce diagnostics. This would affect the disassembler more. There will be no output, but then there wasn't previously in a release build.

The code lacks the plumbing necessary to issue internal errors currently.

alan-baker:
Committed as 983f8f02de963e60c1091e3f22f122ece6e29de6

PR #1820, opened 2018-08-10
Reenable --skip-validation.

s-perron:
In previous changes, the option `--skip-validation` was disabled.  This
change is to reenable it.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -160,7 +160,7 @@ class Optimizer {
   bool Run(const uint32_t* original_binary, size_t original_binary_size,
            std::vector<uint32_t>* optimized_binary) const;
 
-  // Same as above, except is passes |options| to the validator when trying to
+  // Same as above, except passes |options| to the validator when trying to
   // validate the binary.  If |skip_validation| is true, then the caller is
   // guaranteeing that |original_binary| is valid, and the validator will not
   // be run.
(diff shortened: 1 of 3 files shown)


include/spirv-tools/optimizer.hpp
```diff
@@ -151,16 +151,22 @@ class Optimizer {
...
+  // occur when processing |original_binary| using any of the registered passes.
+  // In that case, no further passes are executed and the contents in
+  // |optimized_binary| may be invalid.
   //
   // It's allowed to alias |original_binary| to the start of |optimized_binary|.
   bool Run(const uint32_t* original_binary, size_t original_binary_size,
            std::vector<uint32_t>* optimized_binary) const;
-  bool Run(const uint32_t* original_binary, size_t original_binary_size,
+
+  // Same as above, except is passes |options| to the validator when trying to
```
dj2: Same as above, except passes

johnkslang:
There is a problem here in that this, without an explicit request to do so, writes to stderr, which may not fit at all into a particular tool chain's use, especially programmatic use.

A tool should request that messages go to stderr, not have this happen by default.

s-perron:
@johnkslang Are you sure your comment was for this PR?  It does not add any output stderr, as far as I can tell.

johnkslang:
I am not sure, sorry.

What I saw was glslang's default use of SPIR-V tools getting double runs of the validator: one it requested and one it did not request, and the one it did not request was writing to stderr. My point is that should not be a default unrequested behavior; it should specifically requested, and it seemed directly tied to how this option is implemented in the API.

PR #1821, opened 2018-08-10
Split function opcode validation into new files.

alan-baker:
* Moved function opcode validation out of idUsage and into new files
 * minor style changes
 * General opcode checking is in validate_function.cpp
 * Execution limitation checking is in
 validate_execution_limitations.cpp
* Execution limitations was split into a new pass as it requires other
validation to register those limitations first.

Many of the numbers changed here, please take a closer look.


test/val/val_id_test.cpp
```diff
@@ -2651,7 +2651,7 @@ TEST_F(ValidateIdWithMessage, OpCopyMemoryVoidTarget) {
 %2 = OpTypeInt 32 0
 %3 = OpTypePointer Uniform %1
 %4 = OpTypePointer Uniform %2
-%5 = OpTypeFunction %1 %2 %3
+%5 = OpTypeFunction %1 %3 %4
```
alan-baker: This was incorrect, but the order of checks previously exited before hitting it.

test/val/val_id_test.cpp
```diff
@@ -2673,7 +2673,7 @@ TEST_F(ValidateIdWithMessage, OpCopyMemoryVoidSource) {
...
```
alan-baker: Same here.

test/val/val_id_test.cpp
```diff
@@ -3550,7 +3550,7 @@ TEST_F(ValidateIdWithMessage, OpFunctionResultTypeBad) {
   EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
   EXPECT_THAT(getDiagnosticString(),
               HasSubstr("OpFunction Result Type <id> '2' does not match the "
-                        "Function Type <id> '2's return type."));
+                        "Function Type's return type <id> '1'."));
```
alan-baker: The error seemed bad before as it referred to id 2 twice. I updated it to describe the ids of the actual mismatch.

source/val/validate_execution_limitations.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "source/val/validate.h"
+
+#include "source/val/function.h"
+#include "source/val/validation_state.h"
+
+namespace spvtools {
+namespace val {
+
+spv_result_t ValidateExecutionLimitations(ValidationState_t& _) {
+  for (const auto inst : _.ordered_instructions()) {
```
dj2: Let's move the for() out to the calling method to keep the signature the same (_, instructions) as the other passes. That way, if we have more of these they're easy to add.

source/val/validate_execution_limitations.cpp
```diff
@@ -0,0 +1,61 @@
...
+    if (inst.opcode() == SpvOpFunction) {
```
dj2: This can then be an if (inst.opcode() != SpvOpFunction) return SPV_SUCCESS;

source/val/validate_execution_limitations.cpp
```diff
@@ -0,0 +1,61 @@
...
+        return _.diag(SPV_ERROR_INTERNAL, &inst)
+               << "Internal error: missing function.";
+      }
+
+      for (uint32_t entry_id : _.FunctionEntryPoints(inst.id())) {
+        const auto* models = _.GetExecutionModels(entry_id);
+        if (models) {
+          if (models->empty()) {
+            return _.diag(SPV_ERROR_INTERNAL, &inst)
+                   << "Internal error: empty execution models.";
```
dj2: Can we add the function name here? "Internal error: empty execution models for function <foo>"?

source/val/validate_function.cpp
```diff
@@ -0,0 +1,183 @@
...
+  // NOTE: Find OpFunction & ensure OpFunctionParameter is not out of place.
+  size_t param_index = 0;
+  size_t inst_num = inst->LineNum() - 1;
+  if (inst_num == 0) {
+    return _.diag(SPV_ERROR_INVALID_LAYOUT, inst)
+           << "Function parameter cannot be the first instruction.";
+  }
+
+  auto func_inst = &_.ordered_instructions()[inst_num];
+  if (func_inst != inst) {
```
dj2: Is this check needed? Seems like we're just checking if we used the vector correctly?

alan-baker:
Rebased, squashed and committed as e7fdcdba75eda1f5a70d689cf50628389efa3ed0

PR #1824, opened 2018-08-10
Add ~ConstantManager() to avoid memory leak.

schwa423:
Our ASAN build detected a memory leak due to the ConstantManager never destroying the entries in constant_pool_

### DIFF
--- a/source/opt/constants.cpp
+++ b/source/opt/constants.cpp
@@ -112,6 +112,12 @@ ConstantManager::ConstantManager(IRContext* ctx) : ctx_(ctx) {
   }
 }
 
+ConstantManager::~ConstantManager() {
+  for (auto it = const_pool_.begin(); it != const_pool_.end(); ++it) {
+    delete *it;
+  }
+}
+
 Type* ConstantManager::GetType(const Instruction* inst) const {
   return context()->get_type_mgr()->GetType(inst->type_id());
 }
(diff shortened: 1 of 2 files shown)


s-perron:
Thanks for bringing this to our attention.  However, I am not sure this is the best fix.  The problem is that the issue of ownership is not properly handled, so we could easily get to a double-free situation or other memory leaks.

RegisterConstant is public.  After calling that function who is responsible for deleting the constant?  If it is added to the constant pool, then the constant manager will do it, but, if an equivalent constant is already in the pool, then it is still up to the caller.  That is how we get code like https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/fold_spec_constant_op_and_composite_pass.cpp#L285.  It is very easy for the caller of RegisterConstant to do the wrong thing.

I think the better change it to 

1) make the constant pool a set of unique_ptr.
2) Get `RegisterConsant` to take a `unique_ptr` instead of a plain pointer
3) Get `CreateConstant` to return a `unique_ptr`.

After making these changes it will probably reveal other changes that need to happen.

s-perron:
@schwa423 Are you still working on this?

schwa423:
No, I'm not working on this, too busy with other things.  Your suggestion to have RegisterConstant() take a unique_ptr sounds sane.

s-perron:
It was not a big change.  I did the change myself in a new PR.  I'll close this one.

schwa423:
Thanks s-perron!

PR #1827, opened 2018-08-13
Fixing heap overflow in validation.

alan-baker:
* Changed entry point validation to check storage class of variable
instead of pointer
 * added a test
* Moved several checks after opcode validation
 * These checks should be able to guarantee individual instructions are
 ok


alan-baker:
Rebased, squashed and committed as 397e02442e63781f66c70978fffab7551fdadf3f

PR #1830, opened 2018-08-13
Allow struct members of type void.

s-perron:
For some reason there is an assert the ensures that the member of a
struct is not of type void.  Looking at the spec it seems like this is
wrong.

When defining abstract type it says:

> Abstract Type: An OpTypeVoid or OpTypeBool, or OpTypePointer when
using the Logical addressing model, or any aggregate type containing
any of these.

This seems to imply it is possible.  I cannot find anything that would
forbid it.

I think all we need to do is remove the assert.  I've look at other code in the `Struct` class, and everything should still work when we have a void member.

### DIFF
--- a/source/opt/types.cpp
+++ b/source/opt/types.cpp
@@ -434,12 +434,7 @@ void RuntimeArray::ReplaceElementType(const Type* type) {
 }
 
 Struct::Struct(const std::vector<const Type*>& types)
-    : Type(kStruct), element_types_(types) {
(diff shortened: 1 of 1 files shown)


alan-baker:
> Structure: An ordered heterogeneous collection of any non-void types. When an object is part of a structure, it will also be called a member.

So void should be invalid.

s-perron:
Right.  Thanks.

PR #1834, opened 2018-08-13
Validate correct opcode uses of OpFunction

alan-baker:
* Filed Khronos SPIR-V issue 352
* Updated bad tests
* Added new test

### DIFF
--- a/test/val/val_cfg_test.cpp
+++ b/test/val/val_cfg_test.cpp
@@ -477,9 +477,10 @@ TEST_P(ValidateCFG, BranchTargetFirstBlockBadSinceEntryBlock) {
 
 TEST_P(ValidateCFG, BranchTargetFirstBlockBadSinceValue) {
   Block entry("entry");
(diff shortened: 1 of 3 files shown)


test/val/val_cfg_test.cpp
```diff
@@ -477,9 +477,10 @@ TEST_P(ValidateCFG, BranchTargetFirstBlockBadSinceEntryBlock) {
 
 TEST_P(ValidateCFG, BranchTargetFirstBlockBadSinceValue) {
   Block entry("entry");
+  entry.SetBody("%undef = OpUndef %voidt\n");
   Block bad("bad");
   Block end("end", SpvOpReturn);
-  Block badvalue("func");  // This referenes the function name.
+  Block badvalue("undef");  // This referenes the function name.
```
dj2: This comment seems wrong now?

alan-baker:
Rebased, squashed and committed as 8cb949ad3455ec767583bfe968de1e726f52fe85

PR #1837, opened 2018-08-14
Use MakeUnique.

dj2:
This CL replaces instances of reset(new ..) with MakeUnique.


dj2:
Mostly just to get rid of calls to new. MakeUnique makes it clearer that we're building a unique value as opposed to the reset(new) pattern.

PR #1838, opened 2018-08-14
Remove source/message.h

dj2:
The code in source/message was only used in a single set of tests to
format the output results. This CL changes the test to verify the
message instead of all the error values and removes the source/message
code.


test/opt/ir_loader_test.cpp
```diff
@@ -322,10 +321,9 @@ TEST(IrBuilder, KeepModuleProcessedInRightPlace) {
 // from the given |assembly|.
 void DoErrorMessageCheck(const std::string& assembly,
                          const std::string& error_message) {
-  auto consumer = [error_message](spv_message_level_t level, const char* source,
-                                  const spv_position_t& position,
-                                  const char* m) {
-    EXPECT_EQ(error_message, StringifyMessage(level, source, position, m));
+  auto consumer = [error_message](spv_message_level_t, const char*,
+                                  const spv_position_t&, const char* m) {
+    EXPECT_EQ(error_message, m);
```
s-perron: This change no longer checks that the error was found in the correct position.  It that really what we want?

s-perron: That could be why we ended up with such bad messages coming out of the validator.  I really think we should be checking the position as well.  I don't want these tests to check for less than they currently do.

dj2: Done.

dj2:
Changed to verify message and line number.

johnkslang:
Note that deletion of files breaks down-stream users.

johnkslang:
> Note that deletion of files breaks down-stream users.

Apparently that was improperly consumed, sorry for the distraction.

PR #1840, opened 2018-08-15
Change when instruction is registered in validator.

s-perron:
When doing the validator checks, an instruction is currently registered
at the end of IdPass.  This creates an inconsistency.  In IdPass, an
instruction that uses its own result will treat that use as a forward
reference.  Then in the following passes it will not because the
definition can be found.

It seems best to update the state after all of the check have been done
for the current instruction.  This makes it consistent for all of the
passes.

This makes a different when trying to verify OpTypeStruct.

Fixes https://crbug.com/874372.

### DIFF
--- a/source/val/validate.cpp
+++ b/source/val/validate.cpp
@@ -243,44 +243,48 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
   }
 
   for (auto& instruction : vstate->ordered_instructions()) {
-    // In order to do this work outside of Process Instruction we need to be
-    // able to, briefly, de-const the instruction.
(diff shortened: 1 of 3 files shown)


source/val/validate.cpp
```diff
@@ -243,53 +243,52 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
   }
 
   for (auto& instruction : vstate->ordered_instructions()) {
-    {
```
dj2: Instead of removing this block, can we wrap the usageon 291 in it's own block? I'd like to keep the non-const version of instruction as contained as possible.

PR #1844, opened 2018-08-17
Moving constant opcode validation into a new file

alan-baker:
* Split constant opcode validation out of idUsage and into
validate_constants.cpp
 * minor style fixes
 * reduced duplication
 * fixed an issue with array sizing


source/val/validate_constants.cpp
```diff
@@ -0,0 +1,346 @@
...
+      const auto length = _.FindDef(result_type->GetOperandAs<uint32_t>(2));
+      if (!length) {
+        return _.diag(SPV_ERROR_INVALID_ID, result_type)
+               << "Length is not defined.";
+      }
+      bool is_int32;
+      bool is_const;
+      uint32_t value;
+      std::tie(is_int32, is_const, value) = _.EvalInt32IfConst(length->id());
+      if (is_int32 && is_const && !spvOpcodeIsSpecConstant(length->opcode()) &&
```
alan-baker: I changed this to not check the length value for spec constants.

source/val/validate_constants.cpp
```diff
@@ -0,0 +1,346 @@
...
+           << "Op" << spvOpcodeString(inst->opcode()) << " Result Type <id> '"
+           << _.getIdName(inst->type_id()) << "' is not a boolean type.";
+  }
+
+  return SPV_SUCCESS;
+}
+
+spv_result_t ValidateConstantComposite(ValidationState_t& _,
+                                       const Instruction* inst) {
+  std::string opcode_name = std::string("Op") + spvOpcodeString(inst->opcode());
```
dneto0: Driveby: To avoid always-constructing and appending the std::string, I'd be tempted to make this a lambda to be evaluated only when needed.  I have no idea if that would be faster.

dneto0:
Rebased, squasehd, and pushed into master as c5b38062ec

PR #1846, opened 2018-08-20
Fix EvalInt32IfConst to fail on type instructions.

alan-baker:
Fixes https://crbug.com/875842

* EvalInt32IfConst dereferenced a null pointer if a type instruction was
sent as the id


dneto0:
VS2013 no longer works with googletest.

dneto0:
Rebased, squashed, and pushed into master as 197b4597

PR #1848, opened 2018-08-20
Handle breaks from structured-ifs in DCE.

s-perron:
dead code elimination assumes that are conditional branches except for
breaks and continues in loops will have an OpSelectionMerge before them.
That is not true when breaking out of a selection construct.

The fix is to look for breaks in selection constructs in the same place
we look for breaks and continues for loops.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -245,11 +245,11 @@ void AggressiveDCEPass::AddBranch(uint32_t labelId, BasicBlock* bp) {
 }
 
 void AggressiveDCEPass::AddBreaksAndContinuesToWorklist(
-    Instruction* loopMerge) {
-  BasicBlock* header = context()->get_instr_block(loopMerge);
+    Instruction* mergeInst) {
+  BasicBlock* header = context()->get_instr_block(mergeInst);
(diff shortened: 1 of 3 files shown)


source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -245,11 +245,11 @@ void AggressiveDCEPass::AddBranch(uint32_t labelId, BasicBlock* bp) {
...
 
 void AggressiveDCEPass::AddBreaksAndContinuesToWorklist(
-    Instruction* loopMerge) {
-  BasicBlock* header = context()->get_instr_block(loopMerge);
+    Instruction* mergeInst) {
+  BasicBlock* header = context()->get_instr_block(mergeInst);
   uint32_t headerIndex = structured_order_index_[header];
   const uint32_t mergeId =
-      loopMerge->GetSingleWordInOperand(kLoopMergeMergeBlockIdInIdx);
+      mergeInst->GetSingleWordInOperand(0);
```
dnovillo: Perhaps assert that this is, indeed, a merge instruction?  They're all guaranteed to have the merge ID in operand 0, right?

s-perron: Added the assert, and yes both merge instructions have the merge id at position 0.

PR #1849, opened 2018-08-20
Have PredicateBlocks jump the existing merge blocks.

### LINKED ISSUE #1845: Merge return causing infinite loop.
[before.spvasm.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/2298068/before.spvasm.txt)
[after.spvasm.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/2298069/after.spvasm.txt)

If you look at the attached files, you can see how merge return turns the input (`before.spvasm.txt`) into a shader with an infinite loop.  More details are in the comments of `after.spvasm.txt`.

s-perron:
In PredicateBlocks, we currently skip instructions with side effects,
but it still follows the same control flow (sort-of).  This causes a
problem, when we are trying to predicate code in a loop.  We skip all
of the code with side effects (IV increment), but still follow the
same control flow (jump back the start of the loop).  This creates an
infinite loop because the code will keep jumping back to the start of
the loop without changing the values that effect the exit condition.

This is a large change to merge-return.  When predicating a block that
is in a loop or merge construct, it will jump to the merge block of the
construct.  Once out of all constructs we will generate code as we did
before.

Fixes #1845.

### DIFF
--- a/source/opt/merge_return_pass.cpp
+++ b/source/opt/merge_return_pass.cpp
@@ -220,9 +220,6 @@ void MergeReturnPass::UpdatePhiNodes(BasicBlock* new_source, BasicBlock* target)
 void MergeReturnPass::CreatePhiNodesForInst(BasicBlock* merge_block,
                                             uint32_t predecessor,
                                             Instruction& inst) {
-
-  std::cerr << "Adding new phi nodes for inst " << inst.PrettyPrint(SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES) << std::endl;
-
   DominatorAnalysis* dom_tree =
       context()->GetDominatorAnalysis(merge_block->GetParent());
   BasicBlock* inst_bb = context()->get_instr_block(&inst);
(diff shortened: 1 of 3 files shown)


source/opt/merge_return_pass.cpp
```diff
@@ -251,41 +277,74 @@ void MergeReturnPass::CreatePhiNodesForInst(BasicBlock* merge_block,
...
+  const_block->ForEachSuccessorLabel([this, &block](const uint32_t idx) {
+    BasicBlock* succ_block = context()->get_instr_block(idx);
+    assert(block == nullptr);
+    block = succ_block;
+  });
+  assert(block &&
+         "Return blocks should have returns already replaced by a single "
+         "unconditional branch.");
+
+  (void)(order);
```
alan-baker: Dead parameter

s-perron: Removed.

PR #1851, opened 2018-08-21
[kokoro] Stop running tests for VS2013.

ehsannas:


dneto0:
check-format bot is failing.

ehsannas:
Apparently we want to stop compiling vs 2013 tests. This change was to stop running them on vs2013.

PR #1853, opened 2018-08-21
Fix VS2013 build break.

s-perron:


ehsannas:
2013 bot is back to the googletest failure. So, I think it should be good.

PR #1858, opened 2018-08-22
spirv-as: Fix error in help text.

karl-lunarg:
The assembler always reads assembly text, not binaries.


dneto0:
The bot failures look spurious to me, unrelated.

dnovillo:
I'll push this to master after all the bots finish.

PR #1859, opened 2018-08-22
Update the order when predicating blocks.

s-perron:
When doing predicate blocks in the merge return pass, we need to traverse every block in
structured order in order to keep track of which construct a block is
contained in.  The standard way of traversing code in structured order
is to create a list with all of the nodes in order.  However, when
predicating blocks, new blocks are created, and those blocks are missed.
This causes branches that go too far.

The solution is to update the order as new blocks are created.  Since
we are using an std::list, we do not have to worry about invalidation of
iterators when changing the list.

### DIFF
--- a/source/opt/merge_return_pass.cpp
+++ b/source/opt/merge_return_pass.cpp
@@ -403,10 +403,7 @@ void MergeReturnPass::PredicateBlock(
   predicated->insert(old_body);
 
   // Update |order| so old_block will be traversed.
-  auto pos = std::find(order->begin(), order->end(), block);
-  assert(pos != order->end());
-  ++pos;
-  order->insert(pos, old_body);
+  InsertAfterElement(block, old_body, order);
 
   if (tail_block == block) {
     tail_block = old_body;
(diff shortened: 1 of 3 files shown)


source/opt/merge_return_pass.cpp
```diff
@@ -403,6 +402,12 @@ void MergeReturnPass::PredicateBlock(
       function_->InsertBasicBlockAfter(std::move(new_block), block);
   predicated->insert(old_body);
 
+  // Update |order| so old_block will be traversed.
+  auto pos = std::find(order->begin(), order->end(), block);
+  assert(pos != order->end());
+  ++pos;
+  order->insert(pos, old_body);
+
```
dnovillo: Factor out into a helper function?  It's used a couple of times.

source/opt/merge_return_pass.h
```diff
@@ -215,19 +215,29 @@ class MergeReturnPass : public MemPass {
...
-  // |return_blocks|, extra code is added to jump around any code that should
-  // not be executed because the original code would have already returned. This
-  // involves adding new selections constructs to jump around these
-  // instructions.
+  // For every basic block that is reachable from |return_block|, extra code is
+  // added to jump around any code that should not be executed because the
+  // original code would have already returned. This involves adding new
+  // selections constructs to jump around these instructions.
+  //
+  // If new blocks are created the will be added to |order|.
```
dnovillo: s/the/they/.  Also, could you add a few words describing why this is necessary?

Issue #1861, opened 2018-08-23
Enable breaking from selection construct.


s-perron:
There seems to be at least 1 driver that does not like generating conditional branches that exit from a selection construct.  That was fine until #1850 was merged, because we did not generate those types of instructions.

This issue is to revert part of #1850 until the driver has been fixed.  We fix for the problem with infinite loops will still be fixed by break out of loops, which seems fine.

This issue will remain open until all of #1850 is reenabled.

s-perron:
The latest version of merge return is better.  No need to generate this code anymore.

PR #1862, opened 2018-08-23
Don't "break" from selection constructs.

s-perron:
If seems like at least 1 driver does not like a condition jump to the end
of a selection construct.  We are generating these in the merge return
pass.  This change stops merge return from generating this sequence.

Part of #1861.

### DIFF
--- a/test/opt/pass_merge_return_test.cpp
+++ b/test/opt/pass_merge_return_test.cpp
@@ -489,6 +489,9 @@ TEST_F(MergeReturnPassTest, SplitBlockUsedInPhi) {
   SinglePassRunAndMatch<MergeReturnPass>(before, false);
 }
 
+// TODO(#1861): Reenable these test when the breaks from selection constructs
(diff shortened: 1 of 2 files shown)


s-perron:
@dnovillo Done.

test/opt/pass_merge_return_test.cpp
```diff
@@ -489,6 +489,9 @@ TEST_F(MergeReturnPassTest, SplitBlockUsedInPhi) {
   SinglePassRunAndMatch<MergeReturnPass>(before, false);
 }
 
+// TODO(#1861): Reenable these test when the breaks from selection constructs
+// are reenabled.
+/*
```
dneto0: The proper way to disable a test is to prefix the test name with `DISABLED_`.  Then it is still enumerated and counted but not executed.

https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#temporarily-disabling-tests

PR #1869, opened 2018-08-28
Remove struct member offset monotonicity check

### LINKED ISSUE #1822: Validator: Do not check for increasing buffer offsets
Vulkan has accepted out-of-order for a long time, and an overlooked statement about this was just removed from the SPIR-V specification (rev. 3):

> * Khronos SPIR-V Issue #346: Remove ordered rule for structures in the memory layout: Vulkan allows out-of-order Offset layouts.

alan-baker:
Fixes #1822

* Remove check that struct member offsets must be monotonic
 * All environments match Vulkan behaviour now
 * updated offending tests


alan-baker:
Rebased, squashed and committed as cb0f1f565b29a38f0be313207165d873808c014a

Issue #1841, opened 2018-08-15
Gracefully handle id overflow in spirv-opt


s-perron:
If we have a spir-v module that uses close to the maximum number of ids, it is possible that optimizations will cause the id bound to go over the limit.

At this time, there are no checks that the optimizer does not go past the max id bound.  In fact the current behaviour of `IRContext::TakeNextId()` is to wrap around to 0 after INT_MAX is reached.  Then that value is blindly used by all callers of that function.  It is assumed that it is a valid id.  This needs to be fixed because it can lead to the optimizer failing in strange ways.  See http://crbug.com/873225 for a few examples.

We cannot exit the program in `IRContext::TakeNextId()` or anywhere else in the library code.  We do not want to force the calling program to exit because of an error in spirv-opt.  We need to return control to the caller and let it figure out what it wants to do.

The second issue is to figure out what the limit is suppose to be.  In the [spec](https://www.khronos.org/registry/spir-v/specs/1.2/SPIRV.html#_a_id_limits_a_universal_limits), there is a minimum limit of 0x3FFFFF.  However, since validators have to accept an option to increase that limit, it seems like spirv-opt should not always use that value for the limit.  Should there be an option to set the limit?

dneto0:
I think there are two things to resolve here:

-  Gracefully wrap around

- What to do if we really do run out of IDs and can't make more.

For the graceful wraparound, I would treat it like a stop-the-world garbage collection and be willing to pay a long latency to find the next bunch of free IDs.  For example, refresh the def-use-manager and find the 1000 lowest-numbered unused IDs by scanning (in sequential order) the id-to-def map.

What happens if we really do run out of IDs?   I think we should set a flag in the IR context, and then exit (gracefully?!) out of that pass, and then return an error from the pass manager.  This one is a bit more fuzzy to me since each pass will have its quirks around safe exiting checkpoints.

s-perron:
@dneto0 Any comments on what the limit should be and whether or not there should be an option to change the limit?

s-perron:
@dneto0 @dnovillo @alan-baker @dj2 

I just opened a PR for the first change I made for this issue.  This represents the main method I will use to implement this.  I will have a smallish change that propagates the error up.  Leaving TODOs around to identify where to continue in future PRs.  At each point, things should be getting better.

s-perron:
[...]

PR #1870, opened 2018-08-28
Add option for the max id bound.

s-perron:
* Adds an option to the CLI and API to specify the limit for the id
  bound.
* Modify `TakeNextIdBound` to return 0 when the limit is reached.

Part of #1841.

### DIFF
--- a/include/spirv-tools/libspirv.hpp
+++ b/include/spirv-tools/libspirv.hpp
@@ -24,10 +24,6 @@
 
 namespace spvtools {
 
-// The default id bound is to the minimum value for the id limit
-// in the spir-v specification under the section "Universal Limits".
-const uint32_t kDefaultMaxIdBound = 0x3FFFFF;
-
 // Message consumer. The C strings for source and message are only alive for the
 // specific invocation.
 using MessageConsumer = std::function<void(
@@ -110,6 +106,36 @@ class ValidatorOptions {
   spv_validator_options options_;
 };
 
+// A C++ wrapper around an optimization options object.
+class OptimizerOptions {
+ public:
+  OptimizerOptions() : options_(spvOptimizerOptionsCreate()) {}
+  ~OptimizerOptions() { spvOptimizerOptionsDestroy(options_); }
+
+  // Allow implicit conversion to the underlying object.
+  operator spv_optimizer_options() const { return options_; }
+
+  // Records whether or not the optimizer should run the validator before
+  // optimizing.  If |run| is true, the validator will be run.
+  void set_run_validator(bool run) {
+    spvOptimizerOptionsSetRunValidator(options_, run);
+  }
+
+  // Records the validator options that should be passed to the validator if it
+  // is run.
+  void set_validator_options(const ValidatorOptions& val_options) {
+    spvOptimizerOptionsSetValidatorOptions(options_, val_options);
+  }
+
+  // Records the maximum possible value for the id bound.
+  void set_max_id_bound(uint32_t new_bound) {
+    spvOptimizerOptionsSetMaxIdBound(options_, new_bound);
+  }
+
+ private:
+  spv_optimizer_options options_;
+};
+
 // C++ interface for SPIRV-Tools functionalities. It wraps the context
 // (including target environment and the corresponding SPIR-V grammar) and
 // provides methods for assembling, disassembling, and validating.
@@ -177,7 +203,7 @@ class SpirvTools {
   bool Validate(const uint32_t* binary, size_t binary_size) const;
   // Like the previous overload, but takes an options object.
   bool Validate(const uint32_t* binary, size_t binary_size,
-                const ValidatorOptions& options) const;
+                spv_validator_options options) const;
 
  private:
   struct Impl;  // Opaque struct for holding the data fields used by this class.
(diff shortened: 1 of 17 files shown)


include/spirv-tools/optimizer.hpp
```diff
@@ -163,10 +163,11 @@ class Optimizer {
   // Same as above, except passes |options| to the validator when trying to
   // validate the binary.  If |skip_validation| is true, then the caller is
   // guaranteeing that |original_binary| is valid, and the validator will not
-  // be run.
+  // be run.  The |max_id_bound| is the limit on the max id in the module.
   bool Run(const uint32_t* original_binary, const size_t original_binary_size,
            std::vector<uint32_t>* optimized_binary,
-           const ValidatorOptions& options, bool skip_validation = false) const;
+           const ValidatorOptions& options, bool skip_validation = false,
+           uint32_t max_id_bound = 0x3FFFFF) const;
```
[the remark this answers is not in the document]
s-perron: This is part of the public interface.  Do we want to break backwards compatibility?  Should be leave the original interface and just create a new one with the options struct?

My vote would be for a new one.  Implement the old one as a wrapper for the new one.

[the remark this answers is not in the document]
dj2: I'd say, make a new one, make this one a wrapper and add a DEPRECATED comment to this one.

source/opt/module.h
```diff
@@ -24,6 +24,10 @@
 #include "source/opt/instruction.h"
 #include "source/opt/iterator.h"
 
+namespace {
+const uint32_t kDefaultMaxIdBound = 4194303;
```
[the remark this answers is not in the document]
s-perron: Done, and moved to a single header file that is can access by all places that want to know this limit.

include/spirv-tools/optimizer.hpp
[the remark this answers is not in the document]
s-perron: Actually I should just add it to the validator options because the validator will need the same value eventually.

test/opt/module_test.cpp
```diff
@@ -139,6 +139,44 @@ OpFunctionEnd)";
...
+OpBranch %3
+%3 = OpLabel
+OpReturn
+OpFunctionEnd)";
+
+  std::unique_ptr<IRContext> context = BuildModule(text);
+  uint32_t current_bound = context->module()->id_bound();
+  context->set_max_id_bound(current_bound);
+  uint32_t next_id_bound = context->module()->TakeNextIdBound();
+  EXPECT_EQ(next_id_bound, 0);
```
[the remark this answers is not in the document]
s-perron: Done.

s-perron:
@dneto0 This is ready for another review.

include/spirv-tools/libspirv.hpp
```diff
@@ -173,7 +203,7 @@ class SpirvTools {
   bool Validate(const uint32_t* binary, size_t binary_size) const;
   // Like the previous overload, but takes an options object.
   bool Validate(const uint32_t* binary, size_t binary_size,
-                const ValidatorOptions& options) const;
+                spv_validator_options options) const;
```
dnovillo: why not const ref here?

include/spirv-tools/optimizer.hpp
```diff
@@ -160,13 +160,20 @@ class Optimizer {
...
   bool Run(const uint32_t* original_binary, const size_t original_binary_size,
            std::vector<uint32_t>* optimized_binary,
-           const ValidatorOptions& options, bool skip_validation = false) const;
+           const ValidatorOptions& options, bool skip_validation) const;
+
+  // Same as above, except it takes an options object.  See the documentation
+  // for |OptimizerOptions| to see which options can be set.
+  bool Run(const uint32_t* original_binary, const size_t original_binary_size,
+           std::vector<uint32_t>* optimized_binary,
+           spv_optimizer_options opt_options) const;
```
dnovillo: Same here.  const &?

tools/opt/opt.cpp
```diff
@@ -214,6 +213,9 @@ Options (in lexicographical order):
                growth threshold. The threshold prevents the loop peeling
                from happening if the code size increase created by
                the optimization is above the threshold.
+  --max-id-bound=<n>
+               Sets the maximum value for the id bound for the moudle.  The
+               default is the minimum value for this limit, 4,194,303.
```
dnovillo: Reference the spec here?  And make consistent with the hex notation used in the error message.

include/spirv-tools/libspirv.hpp
s-perron:
```
./include/spirv-tools/libspirv.h:typedef spv_validator_options_t* spv_validator_options;
```

That type is actually a pointer.  That is why.  This is done to expose the name in the API, but still hid the implementation.

include/spirv-tools/optimizer.hpp
s-perron: Same as above.  I followed the same pattern as the other options types.

include/spirv-tools/libspirv.hpp
dnovillo: Sure, but it could use the const pointer variant, right?  Not a big deal, though.

s-perron: Sure can could make it a const.  I'll make that change.

tools/opt/opt.cpp
s-perron: done.  I've had to rebase because of the new spir-v header files.

Issue #1405, opened 2018-03-12
spirv-val does not validate OpTypeBool inside Blocks


HansKristian-ARM:
Reference: https://github.com/KhronosGroup/SPIRV-Cross/issues/504
spirv-val does not catch when OpTypeBools are placed in buffer blocks, which is illegal.

dneto0:
Agreed we should check this.

Would be an application of this clause in OpTypeBool
"If they are stored (in conjunction with OpVariable), they can only be used with logical addressing operations, not physical, and only with non-externally visible shader Storage Classes: Workgroup, CrossWorkgroup, Private, and Function."

jaebaek:
2c2fee7 solves this issue.

PR #1872, opened 2018-08-30
Validator: check OpTypeBool inside Blocks (#1405)

jaebaek:
OpTypeBool can only be used with logical addressing operations, not
physical. Thus, it cannot be a member of buffer block i.e., struct
type. This commit lets Validator check it.

Note that it fails in validate decorations even without this commit when
Validator checks decorations of members in struct type. In that case,
it kills itself without any message i.e. `assert(0)`.

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -139,6 +139,47 @@ bool HasConflictingMemberOffsets(
   return false;
 }
 
+// If |skip_builtin| is true, returns true if |storage| contains bool within
+// it and no storage that contains the bool is builtin.
+// If |skip_builtin| is false, returns true if |storage| contains bool within
+// it.
+bool ContainsInvalidBool(ValidationState_t& _, const Instruction* storage,
+                         bool skip_builtin) {
+  if (skip_builtin) {
+    for (const Decoration& decoration : _.id_decorations(storage->id())) {
+      if (decoration.dec_type() == SpvDecorationBuiltIn) return false;
+    }
+  }
+
+  const size_t elem_type_index = 1;
+  uint32_t elem_type_id;
+  Instruction* elem_type;
+
+  switch (storage->opcode()) {
+    case SpvOpTypeBool:
+      return true;
+    case SpvOpTypeVector:
+    case SpvOpTypeMatrix:
+    case SpvOpTypeArray:
+    case SpvOpTypeRuntimeArray:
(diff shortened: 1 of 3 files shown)


source/val/validate_type.cpp
```diff
@@ -165,6 +165,11 @@ spv_result_t ValidateTypeStruct(ValidationState_t& _, const Instruction* inst) {
       return _.diag(SPV_ERROR_INVALID_ID, inst)
              << "Structures cannot contain a void type.";
     }
+    if (member_type->opcode() == SpvOpTypeBool) {
```
alan-baker: This is not the right place for this check. It is valid for a OpTypeBool to be a member of OpTypeStruct, but if it is then there are restrictions on which storage classes it may be used with.

jaebaek: Updated the code to check "OpTypeBool can be used only with non-externally visible shader Storage Classes: Workgroup, CrossWorkgroup, Private, and Function.".

test/val/val_id_test.cpp
```diff
@@ -1848,6 +1848,30 @@ OpFunctionEnd
   EXPECT_EQ(SPV_SUCCESS, ValidateInstructions());
 }
 
+TEST_F(ValidateIdWithMessage, OpVariableContainsOpTypeBoolBad) {
```
alan-baker: Please add some positive tests. Also make sure to cover the builtin exclusion you added.

source/val/validate_memory.cpp
```diff
@@ -139,6 +139,49 @@ bool HasConflictingMemberOffsets(
   return false;
 }
 
+// Returns true if |inst| contains |op| as its member or opcode of |inst| is
+// |op|.
+bool ContainsType(ValidationState_t& _, const Instruction* inst, SpvOp op) {
```
alan-baker: Why are only pointers and struct subtypes checked? It seems you should check OpTypeVector, OpTypeArray, OpTypeRuntimeArray and OpTypeForwardPointer (be careful checking forward pointers).

jaebaek: I updated this code to check all [*Composite* types](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#CompositeType).
I dropped OpTypePointer and OpTypeForwardPointer because the storage of value pointed by OpTypePointer or OpTypeForwardPointer may have a storage class different from the one of OpVariable.

jaebaek:
Rebased and pushed into master as 2c2fee7

PR #1873, opened 2018-08-30
Add helper methods to opt::Instruction to get storage class information

dj2:


source/opt/instruction.cpp
```diff
@@ -373,6 +373,16 @@ bool Instruction::IsReadOnlyVariableKernel() const {
   return storage_class == SpvStorageClassUniformConstant;
 }
 
+bool Instruction::IsStorageClassInput() const {
+  uint32_t storage_class = GetSingleWordInOperand(kVariableStorageClassIndex);
```
alan-baker: Why doesn't this work with OpTypePointer?

source/opt/instruction.h
```diff
@@ -349,6 +349,13 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   // uniform buffer.
   bool IsVulkanUniformBuffer() const;
 
+  // Returns true if the instruction defines a pointer type that points to an
+  // input class.
```
alan-baker: Should document the preconditions that this instruction is a variable.

dj2:
Fair enough, dropping this CL.

PR #1880, opened 2018-09-06
Add validation check for arrays of void type.

s-perron:
In the definition of an array
(https://www.khronos.org/registry/spir-v/specs/1.2/SPIRV.html#Array),
it specfically mentions that array elements have non-void type.  I've
added a check for that in this PR.

http://crbug.com/879016


dneto0:
You'll need to rebase to catch up to changes necessary to catch up with latest SPIRV-Headers.

PR #1886, opened 2018-09-11
Fixup BUILD.gn file

dj2:


dj2:
TBR= @s-perron will commit once the builders pass to fixup the Chromium breakage.

PR #1887, opened 2018-09-12
Respect --skip-validation in spirv-opt

antiagainst:
Previously a field in OptimizerOptions was set for --skip-validation,
but the OptimizerOptions was never used when calling optimizer.Run().

Also avoid copying the OptimizerOptions object.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -173,7 +173,7 @@ class Optimizer {
   // for |OptimizerOptions| to see which options can be set.
   bool Run(const uint32_t* original_binary, const size_t original_binary_size,
            std::vector<uint32_t>* optimized_binary,
-           const spv_optimizer_options& opt_options) const;
+           const spv_optimizer_options opt_options) const;
 
   // Returns a vector of strings with all the pass names added to this
   // optimizer's pass manager. These strings are valid until the associated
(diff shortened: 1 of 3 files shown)


include/spirv-tools/optimizer.hpp
```diff
@@ -173,7 +173,7 @@ class Optimizer {
   // for |OptimizerOptions| to see which options can be set.
   bool Run(const uint32_t* original_binary, const size_t original_binary_size,
            std::vector<uint32_t>* optimized_binary,
-           const spv_optimizer_options opt_options) const;
+           const spv_optimizer_options& opt_options) const;
```
s-perron: `spv_optimizer_options` is just a pointer if you look at the typedefs.  No need for the reference.

source/opt/optimizer.cpp
```diff
@@ -453,7 +453,7 @@ bool Optimizer::Run(const uint32_t* original_binary,
 bool Optimizer::Run(const uint32_t* original_binary,
                     const size_t original_binary_size,
                     std::vector<uint32_t>* optimized_binary,
-                    const spv_optimizer_options opt_options) const {
+                    const spv_optimizer_options& opt_options) const {
```
s-perron: Same as above.

include/spirv-tools/optimizer.hpp
dnovillo: No reference needed here (it's a pointer).  Yeah, it's a bit confusing.  Perhaps this should be cleaned up.

antiagainst: Ah, yes, sorry! Messed it up with `OptimizerOptions`. :(

antiagainst: Fixed.

source/opt/optimizer.cpp
antiagainst: Yep.

include/spirv-tools/optimizer.hpp
antiagainst: Yep.

PR #1893, opened 2018-09-17
Memory model capability check

alan-baker:
Basic check that if the Vulkan memory model capability is specified, the memory model must be VulkanKHR.


dneto0:
Rebased, squashed, and pushed into master as 149211

PR #1894, opened 2018-09-17
Memory model deprecated decorations

alan-baker:
Checks that Coherent and Volatile decorations cannot be used if the vulkan memory model is used.


dneto0:
Rebased, squashed, and pushed into master as 90a12b3d

PR #1896, opened 2018-09-17
Add dummy loop in merge-return.

s-perron:
The current implementation of merge return can create bad, but correct,
code.  When it is not in a loop construct, it will insert a lot of
extra branch around code.  The potentially large number of branches are
bad.  At the same time, it can separate code store to variables from
its uses hiding the fact that the store dominates the load.

This hurts the later analysis because the compiler thinks that multiple
values can reach a load, when there is really only 1.  This poorer
analysis leads to missed optimizations.

The solution is to create a dummy loop around the entire body of the
function, then we can break from that loop with a single branch.  Also
only new merge nodes would be those at the end of loops meaning that
most analysies will not be hurt.

Remove dead code for cases that are no longer possible.

It seems like some drivers expect there the be an OpSelectionMerge
before conditional branches, even if they are not strictly needed.
So we add them.


source/opt/basic_block.cpp
```diff
@@ -212,8 +212,11 @@ BasicBlock* BasicBlock::SplitBasicBlock(IRContext* context, uint32_t label_id,
                                         iterator iter) {
   assert(!insts_.empty());
 
-  BasicBlock* new_block = new BasicBlock(MakeUnique<Instruction>(
-      context, SpvOpLabel, 0, label_id, std::initializer_list<Operand>{}));
+  std::unique_ptr<BasicBlock> new_block_temp =
+      MakeUnique<BasicBlock>(MakeUnique<Instruction>(
+          context, SpvOpLabel, 0, label_id, std::initializer_list<Operand>{}));
+  BasicBlock* new_block = new_block_temp.get();
+  function_->InsertBasicBlockAfter(std::move(new_block_temp), this);
```
dnovillo: This code is now inserting the new block to the function.  Is the change in semantics needed?  I don't recall whether there was a good reason to do the previous semantics.  I think this is simpler, but this may need changes in every caller.

source/opt/merge_return_pass.cpp
```diff
@@ -820,5 +611,104 @@ void MergeReturnPass::InsertAfterElement(BasicBlock* element,
...
+
+  if (context()->AreAnalysesValid(IRContext::kAnalysisCFG)) {
+    cfg()->RegisterBlock(new_block);
+  }
+
+  return new_block;
+}
+
+void MergeReturnPass::CreateDummyLoop(
+
```
dnovillo: Remove blank line.

source/opt/basic_block.cpp
s-perron: Yes, I already changed all of the calls to this function.  I forgot to update the function comment, so I did that.

source/opt/merge_return_pass.cpp
s-perron: Done.

Issue #1482, opened 2018-04-12
Vulkan 1.1: FPRoundingMode decoration can only be used with float conversion instructions


dneto0:
https://github.com/KhronosGroup/Vulkan-Docs/blob/master/appendices/spirvenv.txt#L427

jaebaek:
This is unclear to me. SPIR-V spec says

"The **FPRoundingMode** decoration can be applied **only to a width-only conversion instruction** that is used as the Object operand of an OpStore storing through a pointer to a 16-bit floating-point object in the StorageBuffer, Uniform, PushConstant, Input, or Output Storage Classes"

In the unit test named `ValidateData.ext_16bit_storage_caps_allow_free_fp_rounding_mode`, it has the following SPIR-V assembly code:

```
OpDecorate %2 FPRoundingMode RTE
        %1 = OpTypeFloat 32
        %2 = OpConstant %1 1.25
```

Since `OpConstant` is not "a width-only conversion instruction", this SPIR-V assembly code must be invalid. Is it valid code?

jaebaek:
Rebased and pushed into master as 03cbf33a695ec84d41a68333920864876fca18d9

PR #1897, opened 2018-09-17
Validator: FPRoundingMode decoration (#1482)

jaebaek:
The following statement from SPIR-V spec is not checked by Validator:
The FPRoundingMode decoration can be applied only to a width-only conversion instruction that is used as the Object operand of an OpStore storing through a pointer to a 16-bit floating-point object in the StorageBuffer, Uniform, PushConstant, Input, or Output Storage Classes.

### DIFF
--- a/source/val/validate_decorations.cpp
+++ b/source/val/validate_decorations.cpp
@@ -902,15 +902,13 @@ spv_result_t CheckDecorationsOfConversions(ValidationState_t& vstate) {
                     "Object operand of an OpStore.";
         }
 
-        const auto object_id = store->word(use.second);
-        const auto ptr_id = store->GetOperandAs<uint32_t>(0);
-        if (object_id != inst->id() || ptr_id == inst->id()) {
+        if (use.second != 2) {
           return vstate.diag(SPV_ERROR_INVALID_ID, inst)
                  << "FPRoundingMode decoration can be applied only to the "
                     "Object operand of an OpStore.";
         }
 
-        const auto ptr_inst = vstate.FindDef(ptr_id);
+        const auto ptr_inst = vstate.FindDef(store->GetOperandAs<uint32_t>(0));
         const auto ptr_type =
             vstate.FindDef(ptr_inst->GetOperandAs<uint32_t>(0));
         const auto storage = ptr_type->GetOperandAs<uint32_t>(1);
--- a/source/val/validate_decorations.cpp
+++ b/source/val/validate_decorations.cpp
@@ -868,6 +868,9 @@ spv_result_t CheckVulkanMemoryModelDeprecatedDecorations(
 }
 
 spv_result_t CheckDecorationsOfConversions(ValidationState_t& vstate) {
+  // Validates FPRoundingMode decoration for Shader Capabilities
+  if (!vstate.HasCapability(SpvCapabilityShader)) return SPV_SUCCESS;
+
   for (const auto& kv : vstate.id_decorations()) {
     const uint32_t id = kv.first;
     const auto& decorations = kv.second;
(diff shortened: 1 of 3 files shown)


source/val/validate_decorations.cpp
```diff
@@ -840,6 +840,72 @@ spv_result_t CheckDecorationsOfBuffers(ValidationState_t& vstate) {
...
+      if (inst->opcode() != SpvOpFConvert) {
+        return vstate.diag(SPV_ERROR_INVALID_ID, inst)
+               << "FPRoundingMode decoration can be applied only to a "
+                  "width-only conversion instruction for floating-point "
+                  "object.";
+      }
+
+      // Validates Object operand of an OpStore
+      bool exist = false;
+      for (const auto& store : vstate.ordered_instructions()) {
```
alan-baker: Please rewrite this go through inst's uses instead of all ordered instructions.

source/val/validate_decorations.cpp
```diff
@@ -840,6 +840,72 @@ spv_result_t CheckDecorationsOfBuffers(ValidationState_t& vstate) {
...
+                storage != SpvStorageClassInput &&
+                storage != SpvStorageClassOutput) {
+              return vstate.diag(SPV_ERROR_INVALID_ID, inst)
+                     << "FPRoundingMode decoration can be applied only to the "
+                        "Object operand of an OpStore in the StorageBuffer, "
+                        "Uniform, PushConstant, Input, or Output Storage "
+                        "Classes.";
+            }
+            exist = true;
+            break;
```
alan-baker: What happens if this conversion is used by multiple stores? It seems only the first would be checked.

source/val/validate_decorations.cpp
```diff
@@ -864,7 +864,70 @@ spv_result_t CheckVulkanMemoryModelDeprecatedDecorations(
...
+      for (const auto& use : inst->uses()) {
+        const auto store = use.first;
+        if (store->opcode() != SpvOpStore) {
+            return vstate.diag(SPV_ERROR_INVALID_ID, inst)
+                   << "FPRoundingMode decoration can be applied only to the "
+                      "Object operand of an OpStore.";
+        }
+
+        const auto object_id = store->GetOperandAs<uint32_t>(1);
+        if (object_id != inst->id()) {
```
alan-baker: I would change this to use the second element of the use. This would avoid missing some bizarre state where we don't otherwise catch `OpStore %conv %conv`.

jaebaek: Updated.

source/val/validate_decorations.cpp
```diff
@@ -902,14 +902,15 @@ spv_result_t CheckDecorationsOfConversions(ValidationState_t& vstate) {
                     "Object operand of an OpStore.";
         }
 
-        const auto object_id = store->GetOperandAs<uint32_t>(1);
-        if (object_id != inst->id()) {
+        const auto object_id = store->word(use.second);
+        const auto ptr_id = store->GetOperandAs<uint32_t>(0);
+        if (object_id != inst->id() || ptr_id == inst->id()) {
```
alan-baker: If the first part of this check fails its because our internal data structures are inconsistent. The use is tracking which word `inst->id()` will appear in. The check should really just be:
```
if (use.second != 2) {
 return error
}

const auto ptr_id = ...
```

jaebaek: Oh ok. I updated it to check only the `use.second != 2`.

jaebaek:
PTAL

jaebaek:
Rebased and pushed into master as 03cbf33a695ec84d41a68333920864876fca18d9

PR #1900, opened 2018-09-18
Memory model image operands

alan-baker:
Validates new memory model image operands: NonPrivateTexelKHR, MakeTexelAvailableKHR and MakeTexelVisibleKHR.


dneto0:
Rebased, squashed, and pushed into master as  1c128aa9ef

PR #1901, opened 2018-09-18
Vulkan memory model memory semantics checks

alan-baker:
Ban sequentially consistent memory semantics with vulkan memory model. Also adds checks for memory semantics.

* Adds checks that OutputMemoryKHR, MakeAvailableKHR and MakeVisibleKHR
are only used if the VulkanMemoryModelKHR capabailty is present
* Added checks that MakeAvailableKHR requires release semantics
* Added checks that MakeVisibleKHR requires acquire semantics
* Added checks that MakeAvailableKHR and MakeVisibleKHR require a
storage class

### DIFF
--- a/source/val/validate_atomics.cpp
+++ b/source/val/validate_atomics.cpp
@@ -82,6 +82,18 @@ spv_result_t ValidateMemorySemantics(ValidationState_t& _,
     return SPV_SUCCESS;
   }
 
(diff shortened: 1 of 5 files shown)


dneto0:
After rebasing, I updated some tests to apply an OriginUpperLeft execution mode for fragment shaders.

Rebased and pushed into master as d652ed3029e

PR #1903, opened 2018-09-18
Add unrolling to the legalization passes

s-perron:
Adds unrolling to the legalization passes.

After enabling unrolling I found a bug when there is a self-referencing
phi node.  That has been fixed.

The test that checks for that the order of optimizations is correct also
needed to be updated.


source/opt/loop_unroller.cpp
```diff
@@ -842,19 +851,21 @@ void LoopUnrollerUtilsImpl::AssignNewResultIds(BasicBlock* basic_block) {
   }
 }
 
-// For all instructions in |basic_block| check if the operands used are from a
-// copied instruction and if so swap out the operand for the copy of it.
-void LoopUnrollerUtilsImpl::RemapOperands(BasicBlock* basic_block) {
-  for (Instruction& inst : *basic_block) {
-    auto remap_operands_to_new_ids = [this](uint32_t* id) {
-      auto itr = state_.new_inst.find(*id);
+void LoopUnrollerUtilsImpl::RemapOperands(Instruction* inst) {
```
dnovillo: Can you keep a variant of the original comment?  The new version doesn't have any documentation.

source/opt/loop_unroller.cpp
```diff
@@ -842,19 +851,21 @@ void LoopUnrollerUtilsImpl::AssignNewResultIds(BasicBlock* basic_block) {
...
+      *id = itr->second;
+    }
+  };
+
+  inst->ForEachInId(remap_operands_to_new_ids);
+}
 
-    inst.ForEachInId(remap_operands_to_new_ids);
+void LoopUnrollerUtilsImpl::RemapOperands(BasicBlock* basic_block) {
+  for (Instruction& inst : *basic_block) {
```
dnovillo: Or just move the original comment here.  It seems to still apply.

source/opt/loop_unroller.cpp
s-perron: The code standards that we use want the comments describing the function just before the declaration, not the definition.  The declaration at line 255 already has the comments, so the ones I removed are redundant.

source/opt/loop_unroller.cpp
dneto0: Yah, generally the method comment describes behaviour from the perspective of the caller.  Comments about the algorithm's internal operation (that are not relevant to the interface) belong inside the function body.

dnovillo:
Yeah. I keep forgetting that this is not optimized for my own workflow. I'm
a heavy tags user, tags take you to the definition not the declaration.

PR #1904, opened 2018-09-19
Update memory model

alan-baker:
Updates from Logical GLSL450 memory model to Logical VulkanKHR memory model.

* Adds capability and extension
* changes memory model
* removes coherent and volatile decorations and adds them to individual instructions as operands
* adds output memory semantics to control barriers in tessellation control shaders
* changes device memory scope to queue family scope


dneto0:
Please rebase to get bot coverage.   (It was a almost trivial.)

Issue #1486, opened 2018-04-12
Vulkan 1.1: limit group op for OpGroupNonUniformBallotBitCount


dneto0:
https://github.com/KhronosGroup/Vulkan-Docs/blob/master/appendices/spirvenv.txt#L472

jaebaek:
Rebased and pushed into master as 026309ab275a63f659d7a8d0a61b0dd6885923cc

PR #1906, opened 2018-09-19
Validator: OpGroupNonUniformBallotBitCount validation (#1486)

jaebaek:


source/val/validate_non_uniform.cpp
```diff
@@ -63,6 +63,27 @@ spv_result_t ValidateExecutionScope(ValidationState_t& _,
...
 }
 
+spv_result_t ValidateGroupNonUniformBallotBitCount(ValidationState_t& _,
+                                                   const Instruction* inst) {
+  // Scope is already checked by ValidateExecutionScope() above.
+
+  const uint32_t result_type = inst->type_id();
+  if (!_.IsUnsignedIntScalarType(result_type)) {
+    return _.diag(SPV_ERROR_INVALID_DATA, inst)
+           << "Expected Result Type to be unsigned int scalar";
```
alan-baker: Change to:
> Expected Result Type to be an unsigned integer type scalar.

source/val/validate_non_uniform.cpp
```diff
@@ -63,6 +63,27 @@ spv_result_t ValidateExecutionScope(ValidationState_t& _,
...
+
+  const auto value = inst->GetOperandAs<uint32_t>(4);
+  const auto value_type = _.FindDef(value)->type_id();
+  if (!_.IsUnsignedIntVectorType(value_type) ||
+      _.GetDimension(value_type) != 4) {
+    return _.diag(SPV_ERROR_INVALID_DATA, inst) << "Expected Value to be a "
+                                                   "vector of four components "
+                                                   "of integer type scalar";
+  }
+  return SPV_SUCCESS;
```
alan-baker: What about the requirement that _Operation_ is 0?

jaebaek: I thought "The identity _I_ for Operation is 0." means "identity _I_" used in the definition of `InclusiveScan` and `ExclusiveScan` as described [here](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Group_Operation) must be `0`.

If _Operation_ must be 0 for `OpGroupNonUniformBallotBitCount`, does it mean its operation must be always `Reduce`?

alan-baker: I'm no longer sure. So I will say, don't check anything.

source/val/validate_non_uniform.cpp
jaebaek: Done.

source/val/validate_non_uniform.cpp
jaebaek: I think we do not need to check `Operation` for `OpGroupNonUniformBallotBitCount`.

jaebaek:
Rebased and pushed into master as 026309ab275a63f659d7a8d0a61b0dd6885923cc

PR #1909, opened 2018-09-20
Memory access checks for vulkan mem model

alan-baker:
* MakePointerVisibleKHR cannot be used with OpStore
* MakePointerAvailableKHR cannot be used with OpLoad
* MakePointerAvailableKHR and MakePointerVisibleKHR both require
NonPrivatePointerKHR
* NonPrivatePointerKHR is limited to a subset of storage classes
* many tests

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -180,6 +180,96 @@ bool ContainsInvalidBool(ValidationState_t& _, const Instruction* storage,
   return false;
 }
 
+std::pair<SpvStorageClass, SpvStorageClass> GetStorageClass(
+    ValidationState_t& _, const Instruction* inst) {
+  SpvStorageClass dst_sc = SpvStorageClassMax;
+  SpvStorageClass src_sc = SpvStorageClassMax;
+  switch (inst->opcode()) {
+    case SpvOpLoad: {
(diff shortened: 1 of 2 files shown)


source/val/validate_memory.cpp
```diff
@@ -180,6 +180,96 @@ bool ContainsInvalidBool(ValidationState_t& _, const Instruction* storage,
...
+      return _.diag(SPV_ERROR_INVALID_ID, inst)
+             << "NonPrivatePointerKHR must be specified if "
+                "MakePointerVisibleKHR is specified.";
+    }
+  }
+
+  if (mask & SpvMemoryAccessNonPrivatePointerKHRMask) {
+    SpvStorageClass dst_sc, src_sc;
+    std::tie(dst_sc, src_sc) = GetStorageClass(_, inst);
+    if (dst_sc != SpvStorageClassUniform &&
```
dneto0: Weird asymmetry here:  this dst_sc does not check SpvStorageClassMax but the src_sc does.

alan-baker: All instructions use dst_sc (maybe a bad name), but only the copies use src_sc. So the condition is whether or not src_sc  should be a check.

PR #1910, opened 2018-09-20
Wrap tests needing effcee inside SPIRV_EFFCEE

antiagainst:


dneto0:
I only need to see one of the bot compiles to pass. Then I'd like to merge.

dneto0:
I'm happy with this.  I'll push onto master.

PR #1911, opened 2018-09-20
[kokoro] Add test job for disabling effcee

antiagainst:


antiagainst:
Hmm, it seems there are more work to do than just modifying the above scripts. The `CI-ubuntu-clang-no-effcee` job is not kicked off.

CC @ehsannas

antiagainst:
Okay, this time it's good. `CI-ubuntu-clang-no-effcee` is skipping effcee tests.

PR #1917, opened 2018-09-24
Allow 0 as argument to scalar replacement.

s-perron:
A limit of 0 for the scalar replacement options it used to indicate that
there is no limit.  The current implementation does not allow 0.  This
should be fixed.


dnovillo:
I forgot to ask.  Does this need a test change?  I remember adding tests for these things, but I don't remember if I tested this limit specifically.

dnovillo:
Yeah, I had added a test for this :)

https://source.cloud.google.com/results/invocations/01fbb386-3e01-4db5-b0a5-75c9d859b32f/targets/graphics_shader_compiler%2Fspirv-tools%2Fmacos-clang-debug%2Fpresubmit/log

Issue #487, opened 2016-12-08
Check that sample image and image operands are only used in designated instructions.


dneto0:
See where an operand must an object whose type is OpTypeSampledImage.
Similarly, operands which must be an object whose type is OpTypeImage.

This might fall under generic type checking task.

jcaraban:
I see two possible interpretations.

1. For every OpTypeImage, walk the uses and ensure they are one of compatible types.
This logic could go in ``isValid<OpTypeImage>`` in validate_id.cpp, or somewhere else?
2. For every opcode accepting an OpTypeImage operand, ensure the type matches.
This logic would definitely go in validate_id.cpp, as I see similar code there.

Although they overlap, 1) and 2) are not exactly the same.

atgoo:
Image operands are expected to appear in operands of type SPV_OPERAND_TYPE_IMAGE.

OpTypeSampledImage and OpTypeImage not appearing in non-image instructions should be covered by complete validation coverage of all opcodes.

I don't think there is any work to be done on this issue.

dneto0:
SPV_OPERAND_TYPE_IMAGE is used for the "image operands" mask word.  It's the special token enum for that case.
This is different from the result ID that is of type OpTypeImage or OpTypeSampledImage.
I think this does need explicit checks.  Certainly merits adding validator tests to ensure it's caught.

jaebaek:
Rebased and pushed into master as 4b4bd4c

PR #1918, opened 2018-09-24
Validator: Validate OpImageTexelPointer (#487)

jaebaek:
Checked all instructions whose object is OpTypeSampledImage or
OpTypeImage as suggested in #487. OpImageTexelPointer instruction
is missing and others look good. This commit adds only
OpImageTexelPointer.


jaebaek:
Rebased and pushed into master as 4b4bd4c53aaa020f7e349aede394d42476b7e3aa

PR #1929, opened 2018-09-26
Keep analyses live in unrolling

### LINKED ISSUE #1928: Loop unrolling increases compile time
We recently added loop unrolling to the set of legalization passes.  This is increasing the compile time a significant amount.  The problem seems to be that the def-use manager is constantly invalidated and rebuilt.  In the case that I am looking at, the def-use manger is rebuilt twice for every loop that gets unrolled.  In one shader, this means that the def-use manger is rebuilt 37 times during unrolling alone.

We know that this is a slow process.

s-perron:
Add code to keep the def-use manger and the inst-to-block mapping up-to-date.  This means we do not have to rebuild them later.

To make this work, we will have to have to find places to update the
def-use manager.  Updating the def-use manager is not straight forward
because we are unrolling loops, and we have circular references.

This forces one pass to register all of the definitions.  A second one
to analyze the uses.  Also because there will be references to the new
instructions in the old code, we want to register the definitions of the
new instructions early, so we can update the uses of the older code as
we go along.

The inst-to-block mapping is not too difficult.  It can be done as instructions are created.

Fixes #1928.

### DIFF
--- a/source/opt/loop_unroller.cpp
+++ b/source/opt/loop_unroller.cpp
@@ -746,9 +746,9 @@ void LoopUnrollerUtilsImpl::CloseUnrolledLoop(Loop* loop) {
 
   // Remove the final backedge to the header and make it point instead to the
   // merge block.
-  state_.previous_latch_block_->terminator()->SetInOperand(
-      0, {loop->GetMergeBlock()->id()});
-  context_->UpdateDefUse(state_.previous_latch_block_->terminator());
+  Instruction* latch_instruction = state_.previous_latch_block_->terminator();
+  latch_instruction->SetInOperand(0, {loop->GetMergeBlock()->id()});
+  context_->UpdateDefUse(latch_instruction);
 
   // Remove all induction variables as the phis will now be invalid. Replace all
   // uses with the constant initializer value (all uses of phis will be in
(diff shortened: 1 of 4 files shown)


source/opt/loop_unroller.cpp
```diff
@@ -736,8 +746,9 @@ void LoopUnrollerUtilsImpl::CloseUnrolledLoop(Loop* loop) {
 
   // Remove the final backedge to the header and make it point instead to the
   // merge block.
-  state_.previous_latch_block_->tail()->SetInOperand(
+  state_.previous_latch_block_->terminator()->SetInOperand(
```
dnovillo: Is this a NOP change?

s-perron: In the next line I needed the same instruction, but I needed an `Instruction*` instead of a `BasicBlock::Itertator`.  Just keep the two consistent.  I probably should have created a new variable that is used in both places.

s-perron:
In the debug version, we have checks that check if the context is consistent.  This is checked after every pass.  This means that, if the def-use manager is suppose to be valid, it is compared to a def-use manager that is build from scratch after every unit test case.

That is already happening with all of the unit test, and that was how I found the parts of the code that had to be fixed.

dnovillo:
On Wed, Sep 26, 2018 at 4:34 PM Steven Perron <notifications@github.com>
wrote:

> In the debug version, we have checks that check if the context is
> consistent. This is checked after every pass. This means that, if the
> def-use manager is suppose to be valid, it is compared to a def-use manager
> that is build from scratch after every unit test case.
>
> That is already happening with all of the unit test, and that was how I
> found the parts of the code that had to be fixed.
>

Oh, right!  I had totally forgotten about this.  That's exactly what I was
hoping we'd have.  Good.

>

antiagainst:
Ship it! Ship it!

PR #1930, opened 2018-09-26
Enable /MP4 (parallel build across 4 cores for MSVC) for SPIRV-Tools/source[/opt]

jeffbolznv:
cmake builds using MSVC normally get parallelism by building N projects in parallel, with each project using 1-2 CPU cores. But as spirv-tools has grown, it has become a serial bottleneck in CTS and glslang builds. Most other projects depend on it (unclear whether it is a real dependency or artificial) and it gets no parallelism.

Applying /MP (parallelize to N cores) risks using N^2 cores if N projects are doing it. This is intended to be a conservative change that uses a modest number of cores and only in the two libraries with the longest compile times. And N=4 solves 75% of the problem, the rest is diminishing returns ;-)

Some build times from my system:
SPIRV-Tools release build: 8m32s -> 4m57s
glslang release build: 6m59s -> 3m15s
Vulkan CTS release build (rough estimate): 12m -> 10m

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -371,3 +371,8 @@ if(ENABLE_SPIRV_TOOLS_INSTALL)
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 endif(ENABLE_SPIRV_TOOLS_INSTALL)
+
+if(MSVC)
+  # Enable parallel builds across four cores for this lib
+  add_definitions(/MP4)
+endif()
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -205,3 +205,7 @@ if(ENABLE_SPIRV_TOOLS_INSTALL)
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 endif(ENABLE_SPIRV_TOOLS_INSTALL)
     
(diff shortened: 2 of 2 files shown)


jeffbolznv:
I'm seeing occasional build failures in CTS with this change where the files generated by spvtools_vendor_tables are not always found. This is surprising, because add_custom_command is supposed to be the "right" way to handle generated files like this. It's pretty rare, which is making it tricky to debug.

jeffbolznv:
I've caught that error again and determined that it is caused by the generated files being built twice, once by SPIRV-Tools and once by SPIRV-Tools-shared. The second one is trying to write the files while the first is trying to compile ext_inst.cpp which includes them. So it is unrelated to this PR.

Sadly, it seems that spirv-tools goes out of its way not to put these generated files in a separate target and use add_custom_target (there is a comment about it).

s-perron:
Even if your change is not the problem, is it hidden without your change?  We don't want to introduce random failures on our build bots or for others.  We may have to fix that other issue before we merge this change.  Can you open an issue describing the problem about the files being built twice?

jeffbolznv:
I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/1941 for the generated file issue.

ringoz:
this /MP4 hack is incompatible with clang-cl, 
and not needed at all with Ninja CMake generator

s-perron:
@ringoz If you are seeing some type of error because of this, please open a new issue, and you can propose the changes that make it work in a pull request.

PR #1933, opened 2018-09-27
Set the C++ version to C++11

alan-baker:
Recent builds have been failing due to updated googletest warnings about non-c++11 code. Set the cmake c++ version to c++11.


s-perron:
Can you revert bfcdc913c4184430e4fca6923683c62723bf5956 at the same time, so we can see if all of the builds work with the latest googletest and your change?

alan-baker:
Testing with the revert of https://github.com/KhronosGroup/SPIRV-Tools/commit/bfcdc913c4184430e4fca6923683c62723bf5956

antiagainst:
Cool stuff!

Issue #1922, opened 2018-09-25
spirv-val does not validate that OpPhi instructions start the block correctly.


s-perron:
The check that an OpPhi is at the start of the basic block that it is contained in is not correct.

Here is an example that it gets it wrong:

```
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %2 "main"
               OpExecutionMode %2 OriginUpperLeft
               OpSource GLSL 400
          %3 = OpString "t.frag"
       %void = OpTypeVoid
          %5 = OpTypeFunction %void
        %int = OpTypeInt 32 1
      %int_0 = OpConstant %int 0
       %bool = OpTypeBool
       %true = OpConstantTrue %bool
      %int_1 = OpConstant %int 1
          %2 = OpFunction %void None %5
         %11 = OpLabel
               OpSelectionMerge %12 None
               OpBranchConditional %true %13 %12
         %13 = OpLabel
               OpBranch %12
         %12 = OpLabel
         %14 = OpUndef %int
               OpLine %3 1 1
         %15 = OpPhi %int %int_0 %11 %int_1 %13
               OpReturn
               OpFunctionEnd
```

alan-baker:
ValidateAdjacency is only check the immediately previous instruction to an OpPhi. This check should be reworked.

jaebaek:
Rebased and pushed into master as f0aa6f4e3a57d428725ef91b7005bb4371d78155

Issue #1923, opened 2018-09-25
Validate function scope variables are at the start of the entry block.


s-perron:
The validator check that function scope variables are defined in the entry block, but it does not check that they are at the start of that block as the specification says: https://www.khronos.org/registry/spir-v/specs/1.2/SPIRV.html#_a_id_logicallayout_a_logical_layout_of_a_module

Here is an example that the validator fails to catch:

```
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %2 "main"
               OpExecutionMode %2 OriginUpperLeft
               OpSource GLSL 400
       %void = OpTypeVoid
          %4 = OpTypeFunction %void
        %int = OpTypeInt 32 1
%_ptr_Function_int = OpTypePointer Function %int
          %2 = OpFunction %void None %4
          %7 = OpLabel
          %8 = OpUndef %int
          %9 = OpVariable %_ptr_Function_int Function
               OpStore %9 %8
               OpReturn
               OpFunctionEnd
```

PR #1935, opened 2018-09-27
Fixed Validator adjacency bug for OpPhi (#1922)

jaebaek:
OpPhi instruction must appear before all non-OpPhi instructions
except for OpLine. Without this commit, Validator does not check
the case that an OpPhi is preceeded by an OpLine and the OpLine is
preceeded by a non-OpPhi instruction that is not OpLine.

### DIFF
--- a/source/val/validate.cpp
+++ b/source/val/validate.cpp
@@ -333,10 +333,12 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
     if (auto error = NonUniformPass(*vstate, &instruction)) return error;
 
     if (auto error = LiteralsPass(*vstate, &instruction)) return error;
-    // Validate the preconditions involving adjacent instructions. e.g. SpvOpPhi
-    // must only be preceeded by SpvOpLabel, SpvOpPhi, or SpvOpLine.
+    // Validate the preconditions involving adjacent instructions.
     if (auto error = ValidateAdjacency(*vstate, i)) return error;
   }
+
+  // SpvOpPhi must only be preceeded by SpvOpLabel, SpvOpPhi, or SpvOpLine.
+  if (auto error = ValidateAdjacencyForOpPhi(*vstate)) return error;
 
   if (auto error = ValidateEntryPoints(*vstate)) return error;
   // CFG checks are performed after the binary has been parsed
--- a/source/val/validate.cpp
+++ b/source/val/validate.cpp
@@ -333,12 +333,11 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
     if (auto error = NonUniformPass(*vstate, &instruction)) return error;
 
     if (auto error = LiteralsPass(*vstate, &instruction)) return error;
-    // Validate the preconditions involving adjacent instructions.
-    if (auto error = ValidateAdjacency(*vstate, i)) return error;
   }
 
-  // SpvOpPhi must only be preceeded by SpvOpLabel, SpvOpPhi, or SpvOpLine.
-  if (auto error = ValidateAdjacencyForOpPhi(*vstate)) return error;
+  // Validate the preconditions involving adjacent instructions. e.g. SpvOpPhi
+  // must only be preceeded by SpvOpLabel, SpvOpPhi, or SpvOpLine.
+  if (auto error = ValidateAdjacency(*vstate)) return error;
 
   if (auto error = ValidateEntryPoints(*vstate)) return error;
   // CFG checks are performed after the binary has been parsed
--- a/source/val/validate_adjacency.cpp
+++ b/source/val/validate_adjacency.cpp
@@ -33,13 +33,14 @@ spv_result_t ValidateAdjacency(ValidationState_t& _, size_t idx) {
 
   switch (inst.opcode()) {
     case SpvOpPhi:
+      while (idx > 0 && instructions[idx - 1].opcode() == SpvOpLine) --idx;
       if (idx > 0) {
         switch (instructions[idx - 1].opcode()) {
           case SpvOpLabel:
           case SpvOpPhi:
-          case SpvOpLine:
             break;
           default:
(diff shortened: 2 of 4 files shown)


source/val/validate_adjacency.cpp
```diff
@@ -33,13 +33,14 @@ spv_result_t ValidateAdjacency(ValidationState_t& _, size_t idx) {
 
   switch (inst.opcode()) {
     case SpvOpPhi:
+      while (idx > 0 && instructions[idx - 1].opcode() == SpvOpLine) --idx;
```
alan-baker: I feel like this whole pass would be better if it got pulled out of the instruction loop in validate and moved below. Then you could just flag state instead of doing the same searches over and over again.

jaebaek: Do you mean adding a function like

```
spv_result_t ValidateAdjacencyForOpPhi(ValidationState_t& _) {
  const auto& instructions = _.ordered_instructions();
  bool valid_to_place = false;

  for (size_t i = 0; i < instructions.size(); ++i) {
    const auto& inst = instructions[i];
    switch (inst.opcode()) {
      case SpvOpLabel:
        valid_to_place = true;
        break;
      case SpvOpPhi:
      case SpvOpLine:
        if (!valid_to_place) {
          return _.diag(SPV_ERROR_INVALID_DATA, &inst)
              << "OpPhi must appear before all non-OpPhi instructions "
              << "(except for OpLine, which can be mixed with OpPhi).";
        }
        break;
      default:
        valid_to_place = false;
        break;
    }
  }

  return SPV_SUCCESS;
}
```

and calling it right after the instruction loop in validate?

alan-baker: Yes, something like that and calling right before ValidateEntryPoints(). Then we can use this structure for #1923 as well.

source/val/validate_adjacency.cpp
```diff
@@ -33,13 +33,14 @@ spv_result_t ValidateAdjacency(ValidationState_t& _, size_t idx) {
...
       if (idx > 0) {
         switch (instructions[idx - 1].opcode()) {
           case SpvOpLabel:
           case SpvOpPhi:
-          case SpvOpLine:
             break;
           default:
+            assert(instructions[idx - 1].opcode() != SpvOpLine);
```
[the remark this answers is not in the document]
jaebaek: Removed.

test/val/val_adjacency_test.cpp
```diff
@@ -196,6 +196,26 @@ OpNop
...
+OpSelectionMerge %end_label None
+OpBranchConditional %true %true_label %false_label
+%true_label = OpLabel
+OpBranch %end_label
+%false_label = OpLabel
+OpBranch %end_label
+%end_label = OpLabel
+OpNop
+OpLine %string 1 1
+%result = OpPhi %bool %true %true_label %false %false_label
```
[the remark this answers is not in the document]
jaebaek: Added. Thanks.

source/val/validate_adjacency.cpp
jaebaek: Updated.

source/val/validate_adjacency.cpp
```diff
@@ -27,25 +27,39 @@
...
+  bool valid_to_place = false;
+
+  for (size_t i = 0; i < instructions.size(); ++i) {
+    const auto& inst = instructions[i];
+    switch (inst.opcode()) {
+      case SpvOpLabel:
+        valid_to_place = true;
+        break;
+      case SpvOpPhi:
+        if (!valid_to_place) {
```
alan-baker: Can you add a test for the following:

```
...
%func = OpFunction ...
%entry = OpLabel
%1 = OpPhi %int
OpReturn
OpFunctionEnd
...
```

This should be invalid, however, I'm not sure if we check it

jaebaek: This is checked [here](https://github.com/KhronosGroup/SPIRV-Tools/blob/2c2fee797908ffb1aab96a3b7265baaada1a0e12/test/val/val_id_test.cpp#L4321)

jaebaek: Oh sorry. I was confused. Is the problem that `OpPhi` is located in the entry block of the function?

alan-baker: Yes, it will have zero predecessors in that case which would seem to match.

source/val/validate.cpp
```diff
@@ -333,11 +333,13 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
     if (auto error = NonUniformPass(*vstate, &instruction)) return error;
 
     if (auto error = LiteralsPass(*vstate, &instruction)) return error;
-    // Validate the preconditions involving adjacent instructions. e.g. SpvOpPhi
-    // must only be preceeded by SpvOpLabel, SpvOpPhi, or SpvOpLine.
+    // Validate the preconditions involving adjacent instructions.
     if (auto error = ValidateAdjacency(*vstate, i)) return error;
```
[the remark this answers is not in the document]
jaebaek: Done.

source/val/validate_adjacency.cpp
```diff
@@ -27,25 +27,39 @@
 namespace spvtools {
 namespace val {
 
+spv_result_t ValidateAdjacencyForOpPhi(ValidationState_t& _) {
+  const auto& instructions = _.ordered_instructions();
+  bool valid_to_place = false;
```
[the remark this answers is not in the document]
jaebaek: Because of the entry block check, I made it as a status check. Please take a look.

source/val/validate_adjacency.cpp
jaebaek: Done.

jaebaek:
Rebased and pushed into master as f0aa6f4e3a57d428725ef91b7005bb4371d78155

PR #1937, opened 2018-09-28
Analyze uses for all instructions.

s-perron:
The def-use manager needs to fill in the `inst_to_used_ids_` field for
every instruction.  This means we have to analyze the uses for every
instruction, even if they do not have any uses.

This mistake was not found earlier because there was a typo in the
equality check for def-use managers.  No new tests are needed.

While looking into this I found redundant work in block merge.  Cleaning
that up at the same time.

### DIFF
--- a/source/opt/def_use_manager.cpp
+++ b/source/opt/def_use_manager.cpp
@@ -278,7 +278,7 @@ bool operator==(const DefUseManager& lhs, const DefUseManager& rhs) {
     return false;
   }
 
-  if (lhs.inst_to_used_ids_ != lhs.inst_to_used_ids_) {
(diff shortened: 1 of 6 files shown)


source/opt/def_use_manager.cpp
```diff
@@ -278,7 +278,7 @@ bool operator==(const DefUseManager& lhs, const DefUseManager& rhs) {
     return false;
   }
 
-  if (lhs.inst_to_used_ids_ != lhs.inst_to_used_ids_) {
+  if (lhs.inst_to_used_ids_ != rhs.inst_to_used_ids_) {
```
alan-baker: LOL

PR #1938, opened 2018-09-28
Validator: function scope variable at start of entry block #1923

jaebaek:
All OpVariable instructions in a function must be the first
instructions in the first block.

### DIFF
--- a/source/val/validate_adjacency.cpp
+++ b/source/val/validate_adjacency.cpp
@@ -28,28 +28,34 @@ namespace spvtools {
 namespace val {
 
 enum {
+  // Status right after meeting OpFunction.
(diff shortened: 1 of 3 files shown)


jaebaek:
I am not sure we can allow `OpLine` between `OpLabel` instruction and `OpVariable` instructions.
Please double-check it.

alan-baker:
>I am not sure we can allow OpLine between OpLabel instruction and OpVariable instructions.
Please double-check it.

OpLine are allowed to be mixed in with OpVariable.

jaebaek:
Rebased and pushed into master as ebcc58b5f8e46b4820cc380dad3522dd58a7d1fe

PR #1940, opened 2018-10-01
Validate execution modes

### LINKED ISSUE #1908: Validator should reject OriginLowerLeft execution mode for vulkan environments
Vulkan requires shaders to use the OriginUpperLeft execution mode, but the validator also accepts OriginLowerLeft. I think the same issue applies to the PixelCenter* execution modes.

alan-baker:
Fixes #1908

Checks the rules in 2.16.2 related to execution modes, the requirements in the execution mode table and the extra requirements added by the Vulkan environment spec.

Many tests needed to be updated to have valid SPIR-V.


source/val/validate_mode_setting.cpp
```diff
@@ -53,6 +54,172 @@ spv_result_t ValidateEntryPoint(ValidationState_t& _, const Instruction* inst) {
...
+                                  case SpvExecutionModeDepthUnchanged:
+                                    return true;
+                                  default:
+                                    return false;
+                                }
+                              })) {
+          return _.diag(SPV_ERROR_INVALID_DATA, inst)
+                 << "Fragment execution model entry points can specify at most "
+                    "one of "
+                    "DepthGreater, DepthLess or DepthUnchanged execution "
```
dj2: Fix wrapping

source/val/validate_mode_setting.cpp
```diff
@@ -68,6 +235,122 @@ spv_result_t ValidateExecutionMode(ValidationState_t& _,
...
+    case SpvExecutionModeLocalSizeId:
+      if (!std::all_of(models->begin(), models->end(),
+                       [](const SpvExecutionModel& model) {
+                         return (model == SpvExecutionModelKernel) ||
+                                (model == SpvExecutionModelGLCompute);
+                       })) {
+        return _.diag(SPV_ERROR_INVALID_DATA, inst)
+               << "Execution mode can only be used with a Kernel or GLCompute "
+                  "execution "
+                  "model.";
```
dj2: Fix wrapping

source/val/validate_mode_setting.cpp
```diff
@@ -68,6 +235,122 @@ spv_result_t ValidateExecutionMode(ValidationState_t& _,
...
+      }
+      break;
+    case SpvExecutionModeTriangles:
+    case SpvExecutionModeOutputVertices:
+      if (!std::all_of(models->begin(), models->end(),
+                       [](const SpvExecutionModel& model) {
+                         switch (model) {
+                           case SpvExecutionModelGeometry:
+                           case SpvExecutionModelTessellationControl:
+                           case SpvExecutionModelTessellationEvaluation:
```
dj2: Do we care about MeshShadingNV for OutputVertices?

source/val/validate_mode_setting.cpp
```diff
@@ -68,6 +235,122 @@ spv_result_t ValidateExecutionMode(ValidationState_t& _,
...
+
+  const auto mode = inst->GetOperandAs<SpvExecutionMode>(1);
+  const auto* models = _.GetExecutionModels(entry_point_id);
+  switch (mode) {
+    case SpvExecutionModeInvocations:
+    case SpvExecutionModeInputPoints:
+    case SpvExecutionModeInputLines:
+    case SpvExecutionModeInputLinesAdjacency:
+    case SpvExecutionModeInputTrianglesAdjacency:
+    case SpvExecutionModeOutputPoints:
```
dj2: Do we care about MeshShadingNV?

source/val/validate_mode_setting.cpp
alan-baker: I added limited validation where it would otherwise fail valid code for MeshNV.

Issue #1939, opened 2018-09-28
Disallow searches in the type manager for non-unique types.


s-perron:
A common source of bugs in the optimizer is dealing with the type manager not returning the correct type.  For example, suppose I have type 1 `struct { int a; }` and type 2 `struct { int b; }`.  If I say I need the type of a struct that contain a single int, then the type manager can correctly return either type 1 or type 2.

This can be very easily overlooked especially because we use to be more aggressive at merging types, which hid a lot of these problems.

To make code more robust, I would like to disallow the constant manager from returning an id when it is possible that there are multiple possible answers.

This can be enforced by not including non-unique types in the type_to_id_ map.  A non-unique id will be defined by `Type::IsUnique`.

I don't know how much code will need to change, but fix will lead to more stable results.  As it stands now, the order in with types are defined can be the difference between hiding or exposing a bug.

PR #1942, opened 2018-10-01
Fix erroneous uses of the type manager in copy-prop-arrays.

s-perron:
There are a few spots where copy propagate arrays is trying
to go from a Type to an id, but the type is not unique.  When generating
code this pass needs specific ids, otherwise we get type mismatches.
However, the ambigous types means we can sometimes get the wrong type
and generate invalid code.

That code has been rewritten to not rely on the type manager, and just
look at the instructions instead.

I have opened https://github.com/KhronosGroup/SPIRV-Tools/issues/1939 to
try to get a way to make this more robust.

### DIFF
--- a/source/opt/copy_prop_arrays.cpp
+++ b/source/opt/copy_prop_arrays.cpp
@@ -26,6 +26,8 @@ const uint32_t kLoadPointerInOperand = 0;
 const uint32_t kStorePointerInOperand = 0;
 const uint32_t kStoreObjectInOperand = 1;
 const uint32_t kCompositeExtractObjectInOperand = 0;
+const uint32_t kTypePointerStorageClassInIdx = 0;
+const uint32_t kTypePointerPointeeInIdx = 1;
 
 }  // namespace
 
@@ -51,7 +53,7 @@ Pass::Status CopyPropagateArrays::Process() {
           FindSourceObjectIfPossible(&*var_inst, store_inst);
 
       if (source_object != nullptr) {
-        if (CanUpdateUses(&*var_inst, source_object->GetPointerTypeId())) {
+        if (CanUpdateUses(&*var_inst, source_object->GetPointerTypeId(this))) {
           modified = true;
           PropagateObject(&*var_inst, source_object.get(), store_inst);
         }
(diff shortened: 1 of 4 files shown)


source/opt/copy_prop_arrays.cpp
```diff
@@ -484,80 +486,76 @@ bool CopyPropagateArrays::CanUpdateUses(Instruction* original_ptr_inst,
...
+        // type of the value being stored.  This way we can always handled
+        // stores.
+        return true;
+      case SpvOpImageTexelPointer:
+      case SpvOpName:
+        return true;
+      default:
+        return use->IsDecoration();
+    }
+  });
```
dnovillo: Are there any changes in this code?  There seem to be just whitespace changes that are confusing the diff.

s-perron: Yes, there are changes in the code.  I think clang-format changed the start of it, which is just whitespace, but the bodies of the OpAccessChain and OpCompositeExtract case changed.

s-perron:
This pass is specifically for propagating array.  If it is not an array, then the other passes (sroa and ssa-rewrite) will handle it.  So this pass does not need to use the type manager.

There are still some uses of the type manager in the analysis portion of this pass.  Those could be removed eventually, but do not cause invalid code to be generated.  They many cause us to miss some opportunities.

source/opt/copy_prop_arrays.cpp
dnovillo: After an offline chat, this part was mostly changed because of clang-format.  The remaining uses of the type manager in here are not really important because this is a predicate that does not alter any code.  Any false negative that this gets are innocuous wrt correctness.

source/opt/type_manager.cpp
```diff
@@ -59,7 +59,9 @@ std::pair<Type*, std::unique_ptr<Pointer>> TypeManager::GetTypeAndPointerType(
 
 uint32_t TypeManager::GetId(const Type* type) const {
   auto iter = type_to_id_.find(type);
-  if (iter != type_to_id_.end()) return (*iter).second;
+  if (iter != type_to_id_.end()) {
+    return (*iter).second;
+  }
```
dnovillo: In our offline chat you mentioned that you'd want to get rid of this NOP change.  I don't mind it, actually.  I prefer every control construct to have braces even for one-liners.

PR #1943, opened 2018-10-01
Make EFFCEE required

### LINKED ISSUE #1912: Remove #ifdef EFFCEE guards in tests
Those guards were only added because we could still run most of our googletest tests but not the effcee tests.  That's because effcee (and its re2) could not build with VS2013, but googletest could build with VS2013.

But recently googletest stopped building under VS2013.  So there's no point in having the #ifdefs which just broke things.
(See https://github.com/KhronosGroup/SPIRV-Tools/pull/1910 and related break in https://github.com/google/shaderc/issues/477 )

This change should:
- [ ] remove the Effcee #ifdef guards
- [ ] update the README and bots to say we now required Effcee in the source tree
- [ ] be accompanied by an email to the developer mailing list about the change.

s-perron:
Makes EFFCEE a required component when the tests are build.  Many of our tests use it.

Fixes #1912.


test/opt/loop_optimizations/fusion_pass.cpp
```diff
@@ -17,18 +17,14 @@
 #include "gmock/gmock.h"
 #include "test/opt/pass_fixture.h"
 
-#ifdef SPIRV_EFFCEE
 #include "effcee/effcee.h"
```
dj2: Can you remove the blank line before the include and then hopefully the formatter should order them correctly (this and other files)

PR #1946, opened 2018-10-02
Use custom allocator in SPIR-V optimization

antiagainst:
This PR adds a custom allocator that allocates memory with
a 4096-byte footprint so that we can consolidate multiple small
allocations into one. This is expected to improve the runtime
performance especially on Windows.

Also converted std::map, std::unordered_map, std::multimap,
std::set, std::unordered_set, std::list, std::queue, std::stack
to use the custom allocator.


antiagainst:
Okay, lots of build configurations are unhappy. I'll submit the conversion for std::list, std::stack, and std::queue in another pull request.

dj2:
Have we tested this on windows to see if it does improve performance?

antiagainst:
@dj2: Yep. I'm seeing 2.5x on a shader corpus.

dj2:
Are we seeing the same speedup on the bots/other machines? Can you update the description with that performance boost?

source/opt/aggressive_dead_code_elim_pass.h
```diff
@@ -161,13 +160,13 @@ class AggressiveDCEPass : public MemPass {
   // immediate controlling structured if or loop.  A loop header block points
   // to its own branch instruction.  An if-selection block points to the branch
   // of an enclosing construct's header, if one exists.
-  std::unordered_map<BasicBlock*, Instruction*> block2headerBranch_;
+  CAUnorderedMap<BasicBlock*, Instruction*> block2headerBranch_;
```
dj2: Do we have to do every single one? Is it every map that is an issue or can we limit this? My fear is we're going to end up with a mix og CAUnorderedMap and std::unordered_map throughout. If we know of the critical sections and just allocate those it will be easier to document what's going on (and make the code change a lot smaller)

source/opt/allocator.h
```diff
@@ -0,0 +1,204 @@
+// Copyright (c) 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
```
dj2: Did you look at importing something like partition alloc instead of a custom one?

antiagainst: You mean https://chromium.googlesource.com/chromium/src/+/dcc13470a/third_party/WebKit/Source/wtf/PartitionAlloc.md?

IIRC, we previously discussed the possibility of leveraging some existing allocator. But it's an additional dependency; the external allocator has its own compiler requirements (yeah, VS2013). Copying in the code will have license problem; even if possible, we may end up pulling in too many source files. So we didn't do that.

I'm not quite happy of writing an allocator by myself, too. But it seems it's a more viable approach. Happy to explore alternatives for sure.

dj2: There is one in base/ as well: https://chromium.googlesource.com/chromium/src/+/master/base/allocator/

I'm just concerned because writing an allocator is .... hard ... and can have security implications.

source/opt/aggressive_dead_code_elim_pass.h
antiagainst: I think it's safer this way. We'll keep changing the code. Previously time-consuming sections can be less time-consuming, and vice versa. Unless we want to do periodical to adjust, which I think is too burdensome.

Converting `std::*` into `CA*` ones does not hurt, even if it is not very time-consuming.

Using mixed `std::*` ones and `CA*` ones won't be a problem I think. They are different types and the compiler can guarantee that new/delete/copy/etc. is properly handled.

The downside is that we now can use more memory for running a recipe. But it's just a trade-off.

source/opt/allocator.h
antiagainst: That's a lot of source files... I'm not quite sure, but is it easily extractable and importable here?

Yep. I totally agree. That's why I added a lot of reviwers. :) The one I have here is basically referencing what Clang `BumpPtrAllocator` is doing.

dj2: Yea, we pulled it into pdfium: https://pdfium.googlesource.com/pdfium/+/master/third_party/base/allocator/partition_allocator there are bits in there and in third_party/base that it needs. But, yea, it's large .....

source/opt/aggressive_dead_code_elim_pass.h
dj2: I guess my question is, can we get the performance benefit while making this less invasive? True, it might regress in the future, but it seems like the kind of thing we should be testing periodically (or getting some performance tests on the bots would be even better).

source/opt/allocator.h
```diff
@@ -0,0 +1,236 @@
...
+// A batch allocator.
+//
+// This allocator wraps around malloc/free and conducts memory allocation on
+// a |kPageSize|-byte footprint. This groups multiple small allocation into a
+// larger one so that we can improve performance.
+//
+// This allocator adopts an ever-increasing approach; new memory are kept
+// allocated even if previously requested memory are deallocated. It's only
+// at the time of the destruction of this allocator that all the memory gets
+// released.
```
s-perron: How does our memory consumption change?  We might need to be careful how much we use.  There are other consideration when you start to use too much memory.  Are we in an environment that does not allow us to use that much memory?  I have seen bugs from the fuzzer because of stuff like this.  It can effect how many parallel compiles can be done if the machines has limited space.  There are also performance issues because of a lack of memory locality.

This type of allocation is not always a win.

antiagainst:
I'll do more experiments and collect more data and post here.

source/opt/allocator.h
ratchetfreak: The way tls containers work is that the allocator's lifetime is tied to the container they are allocating for. So if the container gets actually destructed then the memory will get freed at that point.

However if the container is just cleared then the unused memory will accumulate.

source/opt/common_uniform_elim_pass.h
```diff
@@ -190,20 +189,19 @@ class CommonUniformElimPass : public Pass {
...
 
   // Map from uniform variable id to its common load id
-  std::unordered_map<uint32_t, uint32_t> uniform2load_id_;
+  CAUnorderedMap<uint32_t, uint32_t> uniform2load_id_;
 
   // Map of extract composite ids to map of indices to insts
   // TODO(greg-lunarg): Consider std::vector.
-  std::unordered_map<uint32_t,
-                     std::unordered_map<uint32_t, std::list<Instruction*>>>
+  CAUnorderedMap<uint32_t, CAUnorderedMap<uint32_t, std::list<Instruction*>>>
```
ratchetfreak: forgot the std::list. It's also a node-base container that will probably benefit from using an allocator.

source/opt/allocator.h
s-perron: The STL container will have an instance of `StlAllocator`.  When that gets destroyed it will not free memory.  It will only get freed when `AllocatorImpl` is destroyed, which is owned by the `Optimizer`.  None of this memory will be deleted until that object is destroyed.

source/opt/allocator.h
```diff
@@ -0,0 +1,234 @@
...
+  size_type max_size() const {
+    return std::numeric_limits<size_type>::max() / sizeof(value_type);
+  }
+};
+
+// STL uses the equality operator to determine if memory allocated by one
+// allocator can be deallocated with another.
+template <class T, class U>
+bool operator==(const StlAllocator<T>&, const StlAllocator<U>&) {
+  return false;  // For safety
```
s-perron: I don't think this is valid.  I don't know much about this, but I found some information at https://en.cppreference.com/w/cpp/named_req/Allocator.  If you copy construct the allocator, then an equality check between the new and original copies must return true.

antiagainst:
I analyzed max memory usage on a collection of 120+ real-world shaders and did the following chart

chart

x axis is shader case #, y axis is time/memory ratio. `default` means default allocator, while `custom` means the implementation in this PR.

Data is sorted according to memory ratio.

The native allocator in this PR can achieve 2~3x in speed with 2-3x more memory.

dneto0:
The chart is interesting but seems inconclusive. 

It wasn't obvious to me but the time chart is default / custom.  That's the opposite of what I'd normally use but I could see you wanting to do that to get a time vs. memory comparison.

Can you render the chart again with     custom / default for both memory and time.
Also, what is the geometric mean of the ratios, for both metrics.

s-perron:
The compile time issues will be fixed with https://github.com/Microsoft/DirectXShaderCompiler/pull/1729.

PR #1953, opened 2018-10-04
Don't scalarize spec constant sized arrays

### LINKED ISSUE #1952: linux only spirv-opt crashes with --scalar-replacement on trivial kernel
Trivial Kernel

```
#version 450

layout(constant_id = 0) const int workGroupSize = 32;
layout(constant_id = 1) const int n = 6;

layout(local_size_x = 1) in;

layout(binding = 0) buffer fbuf {
    float a_buffer[];
};

void main() {
    float cr[n];

    for (int i=0; i<n; ++i)
        a_buffer[i] = cr[i];
}
```

Compilation
```
/opt/vulkan/1.1.82.1/x86_64/bin/glslangValidator -V bug.comp -o bug.spv
```

Optimization
```
/opt/vulkan/1.1.82.1/x86_64/bin/spirv-opt --scalar-replacement bug.spv -o bug.spv
spirv-opt: /opt/vulkan/1.1.82.1/source/spirv-tools/source/opt/scalar_replacement_pass.cpp:475: size_t spvtools::opt::ScalarReplacementPass::GetConstantInteger(const spvtools::ir::Instruction*) const: Assertion `constant->opcode() == SpvOpConstant || constant->opcode() == SpvOpConstantNull' failed.
Aborted
```

Windows compiles/optimizes/runs with no issues.

alan-baker:
Fixes #1952

* Prevent scalarization of arrays that are sized by a specialization
constant

### DIFF
--- a/source/opt/scalar_replacement_pass.cpp
+++ b/source/opt/scalar_replacement_pass.cpp
@@ -500,6 +500,12 @@ size_t ScalarReplacementPass::GetNumElements(const Instruction* type) const {
   return len;
 }
 
+bool ScalarReplacementPass::IsSpecConstant(uint32_t id) const {
+  const Instruction* inst = get_def_use_mgr()->GetDef(id);
+  assert(id);
+  return spvOpcodeIsSpecConstant(inst->opcode());
+}
+
 Instruction* ScalarReplacementPass::GetStorageType(
     const Instruction* inst) const {
   assert(inst->opcode() == SpvOpVariable);
@@ -536,6 +542,7 @@ bool ScalarReplacementPass::CheckType(const Instruction* typeInst) const {
         return false;
       return true;
     case SpvOpTypeArray:
+      if (IsSpecConstant(typeInst->GetSingleWordInOperand(1u))) return false;
       if (IsLargerThanSizeLimit(GetArrayLength(typeInst))) return false;
       return true;
       // TODO(alanbaker): Develop some heuristics for when this should be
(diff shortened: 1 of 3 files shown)


source/opt/scalar_replacement_pass.cpp
```diff
@@ -500,6 +500,12 @@ size_t ScalarReplacementPass::GetNumElements(const Instruction* type) const {
   return len;
 }
 
+bool ScalarReplacementPass::IsSpecConstant(uint32_t id) const {
+  const Instruction* inst = get_def_use_mgr()->GetDef(id);
+  assert(id);
```
s-perron: Are you sure you want to assert on `id` and not `inst`?

source/opt/scalar_replacement_pass.cpp
```diff
@@ -536,6 +542,7 @@ bool ScalarReplacementPass::CheckType(const Instruction* typeInst) const {
         return false;
       return true;
     case SpvOpTypeArray:
+      if (IsSpecConstant(typeInst->GetSingleWordInOperand(1u))) return false;
```
s-perron: I know it is not done consistently, but I think we should move towards always adding braces, and splitting these into two lines.  I'm constantly rewriting these when I want to add break points inside the if-block.

alan-baker:
Rebased, squashed and committed as 3b5960174f6d652feb4fb912b7fd84a23bbcde7d

PR #1955, opened 2018-10-04
allow atomics on Function pointers for OpenCL

bashbaug:
For OpenCL environments, an atomic operation on a *Pointer* to a **Function** *Storage Class* is valid, but the SPIR-V validator currently flags this as an invalid operation.  This change relaxes the SPIR-V validator for OpenCL environments only to allow atomic operations on *Pointers* to the **Function** *Storage Class*.  There is no change for other environments.

### DIFF
--- a/test/val/val_atomics_test.cpp
+++ b/test/val/val_atomics_test.cpp
@@ -159,6 +159,8 @@ OpMemoryModel Physical32 OpenCL
 %f32vec4_var = OpVariable %f32vec4_ptr Workgroup
 
 %f32_ptr_function = OpTypePointer Function %f32
+%f32_ptr_uniformconstant = OpTypePointer UniformConstant %f32
+%f32_uc_var = OpVariable %f32_ptr_uniformconstant UniformConstant
 
 %main = OpFunction %void None %func
 %main_entry = OpLabel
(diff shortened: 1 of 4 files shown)


test/val/val_atomics_test.cpp
```diff
@@ -488,7 +515,7 @@ OpAtomicStore %f32vec4_var %device %relaxed %f32_1
 
 TEST_F(ValidateAtomics, AtomicStoreWrongPointerStorageType) {
   const std::string body = R"(
-%f32_var_function = OpVariable %f32_ptr_function Function
+%f32_var_function = OpVariable %f32_ptr_uniformconstant Function
```
alan-baker: Why did this change? The error message is unchanged and this OpVariable is incorrect in terms of storage classes now.

bashbaug: I made this change because an atomic store on a Function pointer is valid for at least some environments now, whereas an atomic store on a UniformConstant pointer is invalid for all environments.  I'm totally open to changing this back though.

Since this test compiles for the "universal" environment it will "fail correctly" even with a Function pointer.  Is that preferred?

If I change this back to a Function pointer I think I should also remove the new AtomicStoreVulkanFunctionPointerStorageType test, since it's effectively already covered here?

alan-baker: I'm ok with changing this to test the UniformConstant storage class, but I'd prefer the rest of the assembly is still valid. I don't like relying on the order of checks in the validator so I just want the mismatch in storage class between pointer and variable fixed.

bashbaug: Ah, I see now.  I'll fix this and the formatting issues.

bashbaug: I believe both this and the formatting issues are fixed.

Issue #1956, opened 2018-10-09
Opt: aggressive code elimination removes too much


kennyalive:
In the attached archive test.spv has  this condition:

```
     ....
         %40 = OpIEqual %bool %37 %uint_0
               OpSelectionMerge %if_merge None
               OpBranchConditional %40 %if_true %if_merge
    %if_true = OpLabel
         %43 = OpAccessChain %_ptr_Function_uint %group_id %int_0
         %44 = OpLoad %uint %43
         %46 = OpAccessChain %_ptr_Uniform_uint %buffer %int_0 %44
         %47 = OpLoad %uint %46
         %48 = OpBitcast %int %47
               OpStore %offset %48
               OpBranch %if_merge
      ....
```

After applying spirv-opt with -O that condition was removed (test-optimized.spv).

Also if I replace -O with -Oconfig with all options from -O except  --eliminate-dead-code-aggressive then conditional branch is not eliminated (test-optimized-non-aggressive.spv).

The above was tested with the latest master-tot binaries (SPIRV-Tools v2018.5-dev a3e3869540b7c2d2f01c846fc76453be55d1ac34). The same result if using VulkanSDK 1.1.82.1 or when I compile original shader with the latest DXC with -O3.

Old versions of spirv-opt did not expose this issue. The version that works well for us is the one that is used with DXC compiler with revision 524ffe2fda4ca5b2fcc271305681013fb2e61ba3 (it's relatively old, May 14, 2018). After that DXC compiler was updated to newer spirv-tools and starting from revision d849dc2b8672cf7e8527923b5b92f5875a3d60e2 (May 16, 2018) it exposes this optimization bug.

[report.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/2460619/report.zip)

greg-lunarg:
Thanks. Looking at it now...

greg-lunarg:
Suspecting spirv-tools 9ec3f81e5c5de63b864c1d568ee9b125dffc3367

greg-lunarg:
Confirmed. Suspecting AtomicIAdd (and other atomics) not being treated as loads.

s-perron:
Took a quick look.  I agree.

kennyalive:
This fixes the issue. Thanks.

PR #1958, opened 2018-10-09
Consider atomics that load when analyzing live stores in ADCE (#1956)

greg-lunarg:


greg-lunarg:
Sorry. Forgot to check this passed CI. I will hammer on this until it passes.

greg-lunarg:
Now passes CI.

PR #1963, opened 2018-10-10
Fold divisions by 0.

### LINKED ISSUE #1961: spirv-opt fails assertion in spvtools::opt::InstructionFolder::BinaryOperate
> SPIRV-Tools-master-linux-Debug/bin/spirv-opt --version
SPIRV-Tools v2018.5-dev a3e3869540b7c2d2f01c846fc76453be55d1ac34

On attached .spv file embedded in [shaderfile.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/2463444/shaderfile.zip):

SPIRV-Tools-master-linux-Debug/bin/spirv-opt shader.spv --eliminate-local-multi-store --simplify-instructions -o opt.spv

spirv-opt: /home/travis/build/KhronosGroup/SPIRV-Tools/source/opt/fold.cpp:76: uint32_t spvtools::opt::InstructionFolder::BinaryOperate(SpvOp, uint32_t, uint32_t) const: Assertion `b != 0u' failed.
Aborted

I'm using a Debug release to reproduce the assertion failure; with Release I get a crash.

Issue found using [GraphicsFuzz](https://github.com/google/graphicsfuzz).

s-perron:
The current implementation in the folder when seeing a division by zero
is to assert.  In the release build, the compiler will attempt to
compute the value, which causes its own problems.

The solution I will go with is to fold the division, and just give it
the value of 0.  The same goes for remainder and mod operations.

Fixes #1961.

### DIFF
--- a/source/opt/fold.cpp
+++ b/source/opt/fold.cpp
@@ -69,29 +69,49 @@ uint32_t InstructionFolder::BinaryOperate(SpvOp opcode, uint32_t a,
     case SpvOp::SpvOpIMul:
       return a * b;
     case SpvOp::SpvOpUDiv:
-      assert(b != 0);
-      return a / b;
+      if (b != 0) {
+        return a / b;
+      } else {
+        // Dividing by 0 is undefined, so we will just pick 0.
+        return 0;
+      }
     case SpvOp::SpvOpSDiv:
-      assert(b != 0u);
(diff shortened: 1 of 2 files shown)


ratchetfreak:
wouldn't folding into OpUndef be an option?

That way further folds can happen based on ops having one operand == undef

PR #1966, opened 2018-10-10
Check for unreachable blocks in merge-return.

### LINKED ISSUE #1962: spirv-opt segfaults or fails assertion in spvtools::opt::Instruction::GetBaseAddress
SPIRV-Tools-master-linux-Debug/bin/spirv-opt --version
SPIRV-Tools v2018.5-dev a3e3869

On attached .spv file embedded in [shader.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/2463841/shader.zip):

SPIRV-Tools-master-linux-Debug/bin/spirv-opt shader.spv --merge-return --eliminate-dead-code-aggressive -o out.spv
Segmentation fault

With debug build from top of tree:

SPIRV-Tools/cmake-build-debug/tools/spirv-opt shader.spv --merge-return --eliminate-dead-code-aggressive -o out.spv
spirv-opt: /home/afdx/SPIRV-Tools/source/opt/instruction.cpp:162: spvtools::opt::Instruction* spvtools::opt::Instruction::GetBaseAddress() const: Assertion `(IsLoad() || opcode() == SpvOpStore || opcode() == SpvOpAccessChain || opcode() == SpvOpInBoundsAccessChain || opcode() == SpvOpCopyObject || opcode() == SpvOpImageTexelPointer) && "GetBaseAddress should only be called on instructions that take a " "pointer or image."' failed.
Aborted

Issue found using [GraphicsFuzz](https://github.com/google/graphicsfuzz).

s-perron:
Merge return assumes that the only unreachable blocks are those needed
to keep the structured cfg valid.  Even those must be essentially empty
blocks.

If this is not the case, we get unpredictable behaviour.  This commit
add a check in merge return, and emits an error if it is not the case.

Fixes #1962.

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -93,7 +93,7 @@ BasicBlock* DeadBranchElimPass::GetParentBlock(uint32_t id) {
 
 bool DeadBranchElimPass::MarkLiveBlocks(
     Function* func, std::unordered_set<BasicBlock*>* live_blocks) {
-  StructuredCFGAnalysis cfgAnalysis(context());
+  StructuredCFGAnalysis* cfgAnalysis = context()->GetStructuedCFGAnalaysis();
 
   std::unordered_set<BasicBlock*> continues;
   std::vector<BasicBlock*> stack;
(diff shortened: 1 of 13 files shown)


s-perron:
We do for legalization.  I must have missed adding it for -O and -Os.  I'll do that.

PR #1969, opened 2018-10-11
Enable precompiled headers for MSVC for spirv-tools-opt

jeffbolznv:
This reduces build time for spirv-tools-opt by about 45 seconds on my system (after my previous /MP4 change to enable parallelism). This helps a lot with the spirv-tools bottleneck in the Vulkan CTS build.

### DIFF
--- a/source/opt/pch.h
+++ b/source/opt/pch.h
@@ -1,0 +1,32 @@
+// Copyright (c) 2018 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


source/opt/pch.h
```diff
@@ -0,0 +1,19 @@
+
+#include <algorithm>
+#include <functional>
+#include <map>
+#include <queue>
+#include <unordered_map>
+#include <unordered_set>
+#include <stack>
+#include <utility>
+#include <vector>
```
dj2: This list seems arbitrary and likely to fall out of date quickly. How long is the tools build time? Is that building everything including the tests or just the source code?

jeffbolznv: precompiled header includes don't have to be perfect, generally as long as you catch the top few headers (which often include most of the headers in the project) it's fine. We've used PCH in the NVIDIA driver build for many years and almost never update the PCH headers.

The total build time for the spirv-tools solution on my system went from about 5 minutes 10 seconds to about 4 minutes 30 seconds. Each spirv-tools-opt file went from around 3 seconds to 1 second.

There appears to be some problem with the PCH in the windows CI builds, which is surprising. I guess it must be Ninja-specific because I can't reproduce it with visual studio.

jeffbolznv:
The remaining failing builds appear to require a Google corporate account. Can somebody share the errors with me?

PR #1970, opened 2018-10-11
Validator: TaskNV can use LocalSize or LocalSizeId

### LINKED ISSUE #1968: Validator: Rev2 of NV mesh shaders extension allows LocalSize and LocalSizeId
Forked from https://github.com/KhronosGroup/glslang/issues/1522 

Revision 2 of [http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/NV/SPV_NV_mesh_shader.html](SPV_NV_mesh_shader) adds LocalSize  and LocalSizeId in mesh shaders.

Validator needs to be relaxed to match

dneto0:
Correponds to the update to Rev2 of SPV_NV_mesh_shader

Fixes #1968


dneto0:
FYI.  This should fix the Shaderc bots

Issue #1967, opened 2018-10-11
Loop removed by aggressive dead code elimination


afd:
With respect to the files in [this archive](https://github.com/KhronosGroup/SPIRV-Tools/files/2470523/tofile.zip),

spirv-opt --eliminate-dead-branches shader.spv -o shader.good.spv

and

spirv-opt --eliminate-dead-branches --eliminate-dead-code-aggressive shader.spv -o shader.bad.spv

leads to binaries that differ with respect to a nest of loops that is present in shader.good.spv but absent in shader.bad.spv (the associated disassembled files are included as .asm files).

The GLSL shader that shader.spv originates from is also included as shader.frag.

As can be seen most easily from the GLSL shader, the write to data[i] inside the inner-most loop cannot legally be removed.

Removing either of the --eliminate-dead-branches or --eliminate-dead-code-aggressive flags makes the problem go away.

I have compiled SPIRV-tools from 68ad72ca9535414a9fa837a6ad5b08dc5f5be769.

Maybe related to #1956?

Thanks!

Issue found using [GraphicsFuzz](http://github.com/google/GraphicsFuzz).

afd:
Included archive of files now.

s-perron:
@greg-lunarg Can you take a look at this?  The problem is that the only live instructions in the outer loop are in the inner loop.  Because the merge instruction for the inner loop is in the inner loop, there is nothing that propagates liveness to the outer loop.  That whole loop is removed.

I looked into fixing it myself, but there is no obvious answer.  I'll pass it off to you, since it is your code.

greg-lunarg:
Sure. Will do so shortly.

PR #1975, opened 2018-10-15
Fix ADCE liveness to include all enclosing control structures. (#1967)

greg-lunarg:
Was removing control structures which didn't have data dependency
with enclosed live loop and otherwise did not contain live code.
An example is a counting loop around a live loop.

Still need to add regression test.


greg-lunarg:
Removed WIP.

PR #1976, opened 2018-10-16
Checks for variable pointers

s-perron:
In logical addressing mode, we are not allowed to generate variables
pointers.  There is already a check for OpSelect.  However, OpPhi,
OpPtrAccessChain, and OpCopyObject are not
checked to make sure it does not generate an variable pointer.  I've
added those checks.

At the same time, there was already a check that there are no loads from
a variable pointer unless the right capabilities are used.  This check
was defieient.  First it only looks at the last instruction generating
the pointer to check if it is a variable pointer.  This is incomplete.
Also, even if there is no load, it is still wrong to generate the
variable pointer.  I've removed that check.

Fixes KhronosGroup#1957.

### DIFF
--- a/source/val/validate_cfg.cpp
+++ b/source/val/validate_cfg.cpp
@@ -59,13 +59,6 @@ spv_result_t ValidatePhi(ValidationState_t& _, const Instruction* inst) {
       return _.diag(SPV_ERROR_INVALID_DATA, inst)
              << "Using pointers with OpPhi requires capability "
              << "VariablePointers or VariablePointersStorageBuffer";
-    } else if (!_.features().variable_pointers) {
-      if (type_inst->GetOperandAs<uint32_t>(1) !=
-          SpvStorageClassStorageBuffer) {
-        return _.diag(SPV_ERROR_INVALID_DATA, inst)
-               << "Storage class of variable pointers must be StorageBuffer "
-                  "when using VariablePointersStorageBuffer";
-      }
     }
   }
 
--- a/source/val/validate_cfg.cpp
+++ b/source/val/validate_cfg.cpp
@@ -48,6 +48,27 @@ spv_result_t ValidatePhi(ValidationState_t& _, const Instruction* inst) {
               "basic blocks.";
   }
 
+  const Instruction* type_inst = _.FindDef(inst->type_id());
+  assert(type_inst);
+
+  const SpvOp type_opcode = type_inst->opcode();
+  if (type_opcode == SpvOpTypePointer &&
+      _.addressing_model() == SpvAddressingModelLogical) {
+    if (!_.features().variable_pointers &&
+        !_.features().variable_pointers_storage_buffer) {
+      return _.diag(SPV_ERROR_INVALID_DATA, inst)
+             << "Using pointers with OpPhi requires capability "
+             << "VariablePointers or VariablePointersStorageBuffer";
+    } else if (!_.features().variable_pointers) {
+      if (type_inst->GetOperandAs<uint32_t>(1) !=
+          SpvStorageClassStorageBuffer) {
+        return _.diag(SPV_ERROR_INVALID_DATA, inst)
+               << "Storage class of variable pointers must be StorageBuffer "
+                  "when using VariablePointersStorageBuffer";
+      }
+    }
+  }
+
   // Create a uniqued vector of predecessor ids for comparison against
   // incoming values. OpBranchConditional %cond %label %label produces two
   // predecessors in the CFG.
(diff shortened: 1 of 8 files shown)


source/val/validate_cfg.cpp
```diff
@@ -48,6 +48,27 @@ spv_result_t ValidatePhi(ValidationState_t& _, const Instruction* inst) {
...
+
+  const SpvOp type_opcode = type_inst->opcode();
+  if (type_opcode == SpvOpTypePointer &&
+      _.addressing_model() == SpvAddressingModelLogical) {
+    if (!_.features().variable_pointers &&
+        !_.features().variable_pointers_storage_buffer) {
+      return _.diag(SPV_ERROR_INVALID_DATA, inst)
+             << "Using pointers with OpPhi requires capability "
+             << "VariablePointers or VariablePointersStorageBuffer";
+    } else if (!_.features().variable_pointers) {
```
alan-baker: I can't find this restriction. The spec only limits storage class to StorageBuffer for OpLoad and OpStore. It seems to be an oversight that OpCopyMemory is not also restricted.

source/val/validate_composites.cpp
```diff
@@ -376,6 +376,27 @@ spv_result_t ValidateCopyObject(ValidationState_t& _, const Instruction* inst) {
     return _.diag(SPV_ERROR_INVALID_DATA, inst)
            << "Expected Result Type and Operand type to be the same";
   }
+
+  const Instruction* type_inst = _.FindDef(result_type);
+  assert(type_inst);
+
+  const SpvOp type_opcode = type_inst->opcode();
+  if (type_opcode == SpvOpTypePointer &&
```
alan-baker: OpCopyObject does not require variable pointers at all.

source/val/validate_memory.cpp
```diff
@@ -735,6 +727,28 @@ spv_result_t ValidateAccessChain(ValidationState_t& _,
...
+                                    const Instruction* inst) {
+  if (_.addressing_model() == SpvAddressingModelLogical) {
+    const uint32_t result_type = inst->type_id();
+    const Instruction* type_inst = _.FindDef(result_type);
+    if (!_.features().variable_pointers &&
+        !_.features().variable_pointers_storage_buffer) {
+      return _.diag(SPV_ERROR_INVALID_DATA, inst)
+             << "Generating variable pointers requires capability "
+             << "VariablePointers or VariablePointersStorageBuffer";
+    } else if (!_.features().variable_pointers) {
```
alan-baker: Same issue as with OpPhi.

source/val/validate_memory.cpp
```diff
@@ -334,18 +334,10 @@ spv_result_t ValidateLoad(ValidationState_t& _, const Instruction* inst) {
            << "' is not defined.";
   }
 
-  const bool uses_variable_pointers =
-      _.features().variable_pointers ||
-      _.features().variable_pointers_storage_buffer;
   const auto pointer_index = 2;
   const auto pointer_id = inst->GetOperandAs<uint32_t>(pointer_index);
   const auto pointer = _.FindDef(pointer_id);
-  if (!pointer ||
```
alan-baker: Why was this changed?

source/val/validate_cfg.cpp
s-perron: In the [capabilities table](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Capability) it says

>VariablePointersStorageBuffer
Allow variable pointers, each confined to a single Block-decorated struct in the StorageBuffer storage class.

It seem that adding the `VariablePointersStorageBuffer` capability only allows certain types of variable pointers.  Says nothing about loads or stores.

However, I do see in the [validation rules](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_validation_a_validation_rules) it seems to discuss what is allowed based on loads and stores.

I also just noticed that the validation rules explicitly allow coping pointers using `OpCopyObject`, but that is defined as a [variable pointer]([url](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#VariablePointer)).

I'll need to go back of some of this.

source/val/validate_composites.cpp
s-perron: What messed me up was the definition of variable pointer says that the pointer result of an OpCopyObject is a variable pointer.

source/val/validate_memory.cpp
s-perron: If you cannot generate a variable pointer, then you cannot use it, but I need to double check that now.

PR #1981, opened 2018-10-17
[WIP] Use temp files when generating files.

### LINKED ISSUE #1941: SPIRV-Tools and SPIRV-Tools-shared libraries generate the same files and break parallel builds
Here's a snippet from a parallel build of Vulkan CTS using MSVC:

```
9>------ Build started: Project: SPIRV-Tools, Configuration: Release x64 ------
...
9>  Generate extended instruction tables for spv-amd-shader-trinary-minmax.
...
23>------ Build started: Project: SPIRV-Tools-shared, Configuration: Release x64 ------
...
23>  Generate extended instruction tables for spv-amd-shader-explicit-vertex-parameter.
23>  Generate extended instruction tables for spv-amd-shader-trinary-minmax.
9>  extensions.cpp
9>C:\vulkancts\vk-gl-cts\external\spirv-tools\src\source\ext_inst.cpp(37): fatal error C1083: Cannot open include file: 'spv-amd-shader-trinary-minmax.insts.inc': No such file or directory
23>  Generate extended instruction tables for spv-amd-gcn-shader.
```

What's happening is that SPIRV-Tools (number 9) had already generated spv-amd-shader-trinary-minmax, but SPIRV-Tools-shared (number 23) is rebuilding it at the same time that SPIRV-Tools is compiling ext_inst.cpp that includes it. This happens maybe once in every 10-20 times I build.

This is exactly the issue described here: https://public.kitware.com/Bug/view.php?id=12311, and it is considered not a CMake bug but a bug in the makefiles. They resolved that issue by adding this documentation:

```
+      "Do not list the output in more than one independent target that may "
+      "build in parallel or the two instances of the rule may conflict "
+      "(instead use add_custom_target to drive the command and make the "
+      "other targets depend on that one).  "
```

But the SPIRV-Tools cmakefile intentionally (without realizing the consequences?) doesn't use add_custom_target, with this comment:

```
# The following .cpp files include the above generated .inc files.
# Add those .inc files as their dependencies.
#
# Why using such an awkward way?
# * If we use add_custom_target() to define a target to generate all .inc files
#   and let ${SPIRV_TOOLS} depend on it, then we need to run ninja twice every
#   time the grammar is updated: the first time is for generating those .inc
#   files, and the second time is for rebuilding .cpp files, when ninja finds
#   out that .inc files are updated.
# * If we use add_custom_command() with PRE_BUILD, then the grammar processing
#   script will always run no matter whether the grammar is updated.
# * add_dependencies() is used to add *target* dependencies to a target.
# * The following solution only generates .inc files when the script or the
#   grammar files is updated, and in a single ninja run.
```

I'm not an expert on cmake (and don't fully understand the ninja issue), but it seems to me that the "right" way to do this in cmake is to use add_custom_target and to find some other way to deal with the ninja issue.

s-perron:
When generating files from the json files, we currently open up the
destination file and write to it directly.  This can cause problems
if multiple processes try to write the same file.  One part of the build
can believe the file is valid while another is overwriting it.

To try to solve this problem, I have changed the python scripts to write
to a temp file, and then moves the temp file.  On Unix systems this move
should be atomic.  On Windows, if the file already exists, then an
exception will be raised.  When this happens we can just delete the temp
file. We will assume the process that created the file did it correctly.

This fixes #1941.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -110,9 +110,9 @@ macro(spvtools_vendor_tables VENDOR_TABLE)
       --vendor-insts-output=${INSTS_FILE}
     DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${GRAMMAR_FILE}
     COMMENT "Generate extended instruction tables for ${VENDOR_TABLE}.")
-  list(APPEND EXTINST_CPP_DEPENDS ${INSTS_FILE})
   add_custom_target(spirv-tools-${VENDOR_TABLE} DEPENDS ${INSTS_FILE})
   set_property(TARGET spirv-tools-${VENDOR_TABLE} PROPERTY FOLDER "SPIRV-Tools build")
+  list(APPEND EXTINST_CPP_DEPENDS spirv-tools-${VENDOR_TABLE})
 endmacro(spvtools_vendor_tables)
 
 macro(spvtools_extinst_lang_headers NAME GRAMMAR_FILE)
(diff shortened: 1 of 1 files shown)


utils/generate_grammar_tables.py
```diff
@@ -34,6 +36,17 @@
...
 
+def write_to_file(data, filename):
+    temp_file = tempfile.NamedTemporaryFile(mode='w', dir=os.path.dirname(filename), delete=False)
+    print(data,file=temp_file)
+    temp_file.flush()
+    temp_file.close()
+    try:
+      os.rename(temp_file.name, filename)
+    except OSError:
+      # Can happen on Windows when the file arleady exists.  We can just delete
```
antiagainst: I think this will have problems for incremental build when the grammar changes? The outdated file will always be in the binary directory preventing us from writing to it. It seems we should compare the contents and do force replacement if they differ?

s-perron: Right.  Fine for Unix systems, but not for Windows.

PR #1984, opened 2018-10-17
Fix ADCE to treat OpUnreachable correctly during liveness analysis

greg-lunarg:
ADCE liveness algorithm should treat OpUnreachable at least like other
branch instructions. It was being treated as always live which was
preventing useless structured constructs from being eliminated.
OpUnreachable is generated by dead branch elimination which is now
being required by merge return, so this fix should accompany that
change.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -464,7 +464,7 @@ bool AggressiveDCEPass::AggressiveDCE(Function* func) {
       if (varId != 0) {
         ProcessLoad(varId);
       }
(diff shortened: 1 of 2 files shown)


greg-lunarg:
All CI passed last night. Not sure how a change to aggressive_dead_code_elim_pass.cpp could start causing compile error in fold_test.cpp. Did clang get updated on the CI machine.

s-perron:
Other PR are failing as well.  Something changed on the CI machine.  We are looking into it now.

s-perron:
I manually ran the failing test with clang 4.0 instead of 3.8.  They passed.  Those failures are unrelated to this change.  I'll merge.

PR #1986, opened 2018-10-18
Add custom target to wrap around custom commands.

s-perron:
In CMake, we are not suppose to have multiple targets depend on the same
custom command.  To avoid this, we have to add a custom target around
the command.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -110,9 +110,9 @@ macro(spvtools_vendor_tables VENDOR_TABLE)
       --vendor-insts-output=${INSTS_FILE}
     DEPENDS ${GRAMMAR_PROCESSING_SCRIPT} ${GRAMMAR_FILE}
     COMMENT "Generate extended instruction tables for ${VENDOR_TABLE}.")
-  list(APPEND EXTINST_CPP_DEPENDS ${INSTS_FILE})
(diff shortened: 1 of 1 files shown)


source/CMakeLists.txt
```diff
@@ -159,36 +159,34 @@ list(APPEND OPCODE_CPP_DEPENDS ${GENERATOR_INC_FILE})
...
+  DEPENDS ${OPCODE_CPP_DEPENDS} ${OPERAND_CPP_DEPENDS})
+add_custom_target(enum_string_mapping
+  DEPENDS ${EXTENSION_H_DEPENDS} ${ENUM_STRING_MAPPING_CPP_DEPENDS})
+add_custom_target(extinst_tables
+  DEPENDS ${EXTINST_CPP_DEPENDS})
+
 set_source_files_properties(
   ${CMAKE_CURRENT_SOURCE_DIR}/opcode.cpp
-  PROPERTIES OBJECT_DEPENDS "${OPCODE_CPP_DEPENDS}")
+  PROPERTIES OBJECT_DEPENDS core_tables)
```
antiagainst: Is this actually working? The doc for OBJECT_DEPENDS (https://cmake.org/cmake/help/v3.0/prop_sf/OBJECT_DEPENDS.html) sys it should be files but we are providing targets here.

s-perron: It is working for the linux ninja build.

PR #1987, opened 2018-10-18
Add a nullptr check to array copy propagation.

s-perron:
We are missing a check for a nullptr that is causing things to fail.

Added an extra test case, and fixed up others.


dnovillo:
The changes to fold_test.cpp have somehow confused the clang build.  Not quite sure what template instantiation it's complaining about. 

https://source.cloud.google.com/results/invocations/f214422f-d2d8-4c55-96bc-8063de211f4b/targets/graphics_shader_compiler%2Fspirv-tools%2Flinux-clang-release%2Fpresubmit/log

s-perron:
@dnovillo The failures are unrelated to my change.  Can you please review as if they did not fail.  I run those test with clang 4.0 manually to make sure they pass.

PR #1989, opened 2018-10-18
Use per-configuration location for pch file

jeffbolznv:
Once the PCH change made it into Vulkan CTS, I ran into a problem where switching between release and debug builds complained about mismatching compile flags in the PCH, because they were inadvertently sharing a PCH file.

This change uses the MSVC builtin macro $(IntDir) (the path intermediate files go to) to select a per-config location for the PCH file.

### DIFF
--- a/source/opt/CMakeLists.txt
+++ b/source/opt/CMakeLists.txt
@@ -190,8 +190,13 @@ if(MSVC)
   add_definitions(/MP4)
 
   # Enable precompiled header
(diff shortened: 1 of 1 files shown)


jeffbolznv:
I'm struggling with ninja and non-ninja builds disagreeing on how to handle $ literals. There also appear to be some build failures that are unrelated to this change.

jeffbolznv:
I *think* this is working now (still a few more builds to go). Needed to distinguish between Visual Studio generators (a multi-generator) vs Ninja (a single generator, where $(IntDir) doesn't exist).

The failures in the clang build look unrelated to this change.

s-perron:
You can rebase with the master branch to pick up the fix for the clang build.  I'll wait until the windows build passes before I review.

jeffbolznv:
Thanks, all builds passed this time.

PR #2009, opened 2018-10-23
Fix dead branch elim infinite loop.

### LINKED ISSUE #2004: spirv-opt does not terminate with --eliminate-dead-branches
[This archive](https://github.com/KhronosGroup/SPIRV-Tools/files/2505293/tofile.zip) contains the files necessary to reproduce the issue.

shader.spv was generated from shader.frag using glslangValidator.

Running:

spirv-opt --eliminate-dead-branches --merge-return --inline-entry-points-exhaustive --eliminate-local-multi-store --ccp --eliminate-dead-branches shader.spv -o temp.spv

sends spirv-opt into a seemingly infinite loop.

spirv-opt --version
SPIRV-Tools v2018.6-dev v2018.5-81-g07d0f9df

Found using [GraphicsFuzz](https://github.com/google/graphicsfuzz).

s-perron:
When looking for a break from a selection construct, we do not realize
that a jump to the continue target of a loop containing the selection
is a break.  This causes and infinit loop, or possibly other failures.

Fixes #2004.

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -55,9 +55,7 @@ bool DeadBranchElimPass::GetConstCondition(uint32_t condId, bool* condVal) {
           GetConstCondition(cInst->GetSingleWordInOperand(0), &negVal);
       if (condIsConst) *condVal = !negVal;
     } break;
-    default: {
-      condIsConst = false;
-    } break;
+    default: { condIsConst = false; } break;
   }
   return condIsConst;
 }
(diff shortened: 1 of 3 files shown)


source/opt/dead_branch_elim_pass.cpp
```diff
@@ -440,28 +441,34 @@ Pass::Status DeadBranchElimPass::Process() {
...
+          // If a possible target is the |loop_merge_id| or |loop_continue_id|,
+          // which are not the current merge node, then we continue the search
+          // with the other target.
           for (uint32_t i = 1; i < 3; i++) {
             if (branch->GetSingleWordInOperand(i) == loop_merge_id &&
                 loop_merge_id != merge_block_id) {
               next_block_id = branch->GetSingleWordInOperand(3 - i);
               break;
             }
+            if (branch->GetSingleWordInOperand(i) == loop_continue_id) {
```
alan-baker: The comment doesn't match this code. The loop continue block could be the same block as the current merge block.

s-perron: Fixed.  However, I don't think it makes a difference either way.

PR #2011, opened 2018-10-23
Added support for Transform Feedback capabilities.

brett19:


brett19:
Note: There is a related change for Vulkan-ValidationLayers here: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/419

alan-baker:
Sorry for letting this languish for so long. I have opened #2088 with test fixes and your implementation.

PR #2014, opened 2018-10-26
Add base and core bindless validation instrumentation classes

greg-lunarg:


greg-lunarg:
Getting id shift in my tests on the linux-based platforms. Windows is passing. Any ideas?

greg-lunarg:
Figured out the id shift problem. I had two different routines that were doing TakeNextId() and was calling them as arguments to a function call where, I presume, the order of execution is not guaranteed across platforms.

include/spirv-tools/instrument.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+#ifndef INCLUDE_SPIRV_TOOLS_INSTRUMENT_HPP_
+#define INCLUDE_SPIRV_TOOLS_INSTRUMENT_HPP_
+
+namespace spvtools {
+
+// Error Codes
+// These are utilized in debug output record streams and identify which
+// validation error has occurred and written a specific record.
+static const uint32_t kInstErrorBindlessBounds = 0;
```
dj2: Do all of these need to be in the public api? If not, this should probably live in the opt/ directory and these should be enum classes or something.

greg-lunarg: Yes, these are all intended to be exported outside of spirv-opt. They will be used by the Vulkan bindless validation layer to read the data written by the instrumented bindless code and pass it to the Vulkan developer to understand what error occurred and where. Is there a better place for this file?

greg-lunarg: In fact, I had a request to add a few more, which I am going to do next.

dj2: Is there a bug or design doc I can look at to better understand how this works/ what it does?

greg-lunarg:
Building of the validation files is failing although I haven't made any (direct) changes in the area. Any thoughts on what is going on here or how to proceed?

dj2:
Looks like it failed on windows due to PDB issues. Usually re-building fixes the problem.

include/spirv-tools/instrument.hpp
greg-lunarg: I have now updated instrument.hpp to make it friendlier to external users/developers. Let me know if you have additional suggestions.

s-perron:
Related to what @dj2 was asking for, now that we have the official PR, is there a spec or something in gitlab that describes how all of this is suppose to work with all of the other parts?  What tasks will be validation layer be responsible for, and how will a user know which descriptor set and bindings the output will be located in?  Are these error codes defined in a specification for the feature?  What about the layouts for the messages?

I would like to know that the external interface is agreed upon and what it is  so I have a way of verifying that it is implemented correctly.

Also, should some of the things exist in spirv-headers, or something like that.  If they are in spirv-tools,  any project that uses this feature will have to depend on spirv-tools.

greg-lunarg:
> is there a spec or something in gitlab that describes how all of this is suppose to work with all of the other parts?

InstBindlessCheckPass is designed to partially address gitlab issue 1327 (Add Bounds Checking to Layers Using GPU-Assisted Validation) and InstrumentPass is designed to partially address 1293 (Develop tooling for shader instrumentation?). These are the only public documents for this work. The most recent design in 1327 should give you some context to understand the larger layer mechanism and how the instrumentation fits into it.

There is a lower-level design document that the layer developer is keeping here internal to LunarG, but as internal document, it is not clear how useful it would be for you to verify against it.

You should note that I have documented in inst_bindless_check_pass.h that the external design of InstBindlessCheckPass is wholly intended to address the needs of the bindless validation layer and that its external design may change as the needs of that layer change. This is meant to give a heads up to anyone else who builds anything against this design.

That said, the design of both InstBindlessCheckPass and the layer are pretty much set, at least for the core functionality. We have a version of the layer that was recently demo'd at the last F2F and is running solidly on at least one real game. So I am not expecting too much more change, which is why I am now attempting to check this in.

Can we just assume that the functionality that I document in inst_bindless_check_pass.h, instrument.hpp and optimizer.hpp is useful to the community, and verify that the code matches the documentation?

> how will a user know which descriptor set and bindings the output will be located in?

As is documented in optimizer.hpp, the descriptor set is passed in by the layer through CreateInstBindlessCheckPass(). As is documented in instrument.hpp, the binding for the stream output buffer is 0. Let me know if you think this should be documented some other way.

> Are these error codes defined in a specification for the feature? What about the layouts for the messages?

The documentation I have written is for the developer of the validation layer. The developer of the validation layer will decide what and how to document to the user of the layer since that is really a separate design.

> I would like to know that the external interface is agreed upon and what it is so I have a way of verifying that it is implemented correctly.

The external interface for the output buffer for InstBindlessCheckPass is essentially what is documented in instrument.hpp. The layer includes instrument.hpp and uses those static consts, so by definition there is agreement on the external interface.

> Also, should some of the things exist in spirv-headers, or something like that. If they are in spirv-tools, any project that uses this feature will have to depend on spirv-tools.

If the layer wishes to publish offsets somewhere to allow other tools to instrument to the layer, that would be their prerogative, but for now the layer is giving the control of the format of the output buffer to InstBindlessCheckPass.

greg-lunarg:
This is documented elsewhere, but I thought I would point out that this current commit only covers core bindless functionality. The core functionality assumes that descriptor arrays are of compile-time constant size and all descriptors are initialized.

Future commits will cover the vk_ext_descriptor_indexing extension, which allows for runtime-sized descriptor arrays and uninitialized descriptors (if unused). My current thinking is that I will extend InstBindlessCheckPass to handle runtime-sized descriptor arrays, but will create a separate pass to check for uninitialized descriptors. Both passes will use the same output buffer stream to record errors.

source/opt/inst_bindless_check_pass.cpp
```diff
@@ -0,0 +1,263 @@
...
+    case SpvOp::SpvOpImageSparseFetch:
+    case SpvOp::SpvOpImageSparseRead:
+    case SpvOp::SpvOpImageWrite:
+      image_id =
+          ref_inst_itr->GetSingleWordInOperand(kSpvImageSampleImageIdInIdx);
+      break;
+    default:
+      return;
+  }
+  Instruction* imageInst = get_def_use_mgr()->GetDef(image_id);
```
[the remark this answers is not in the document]
greg-lunarg: Done

source/opt/basic_block.h
```diff
@@ -71,6 +71,9 @@ class BasicBlock {
   // Appends all of block's instructions (except label) to this block
   inline void AddInstructions(BasicBlock* bp);
 
+  // The pointer to the label starting this basic block.
```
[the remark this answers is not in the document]
greg-lunarg: Good point.

So I changed the function slightly to pass back the lvalue of the label pointer which then allows me to do a std::move on the result of the function call.

Is that ok?

s-perron:
We talked internally about the header file, and it is fine.  I don't really like `GetLabel` returning a reference to a unique pointer, but I won't hold this up any longer.  We can change that later if we need a need.

dneto0:
> We talked internally about the header file, and it is fine. 

My view was that these constants are tightly coupled to the code that does the instrumentation.  The new instrumentation code is being added to SPIRV-Tools so it's natural to put this header here.  It's good that the new constants are in their own header file, so it's clear that it's a different part of the API surface.

In future, we might expose the optimizer's classes and APIs for representing, manipulating, and analyzing a SPIR-V module (a real _representation_ for update purposes).  In that case this instrumentation code could take a life of its own in its own repository (and independent maintenance, evolution, and leadership).  In that case the code doing the instrumentation and its interface header would move to that other repository.  That future is still open to us, and that's great.

PR #2023, opened 2018-10-31
Add /Zm flag as a workaround for VS2013 build

jeffbolznv:
The glslang build has an out of memory error related to precompiled headers in its VS2013 build, see https://github.com/KhronosGroup/glslang/pull/1543. This change adds a flag to increase the amount of memory available for PCH.

I found some comments from Microsoft that this isn't an issue anymore in VS2015 and newer, which is why it didn't show up in the spirv-tools build or in CTS.


alelenv:
I had a pull request just before you did the same :)
https://github.com/KhronosGroup/SPIRV-Tools/pull/2022

I am fine using that or abandoning it

jeffbolznv:
@alelenv I got a build failure in my first try, due to the flag mismatching between pch.cpp and the files that include pch.h. I can't see the CI results for your change, but I assume they'd have the same error. Hoping this second version is clean...

jeffbolznv:
All builds passed this time, so I think we should go with this one.

antiagainst:
Cool, thanks! I'll merge this one and close #2022.

PR #2024, opened 2018-10-31
Reduce number of test targets

dj2:
This CL combines a bunch of tests targets into single executables. This reduces the number of build targets and number of test suites to be run in ctest.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -175,6 +175,29 @@ add_spvtools_unittest(
   LIBS ${SPIRV_TOOLS})
 endif()
 
+
+add_spvtools_unittest(
+  TARGET bit_stream
+  SRCS bit_stream.cpp
+      ${CMAKE_CURRENT_SOURCE_DIR}/../source/comp/bit_stream.cpp
+      ${CMAKE_CURRENT_SOURCE_DIR}/../source/comp/bit_stream.h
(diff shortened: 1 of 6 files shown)


greg-lunarg:
Why did we do this?

I kind of liked the the ability to build and run a single test suite, especially if I was only working on one pass.

greg-lunarg:
OK. I figured out how to just run the suites I want. Building everything is still an annoyance, but it will probably have the same frequency as building all passes.

dj2:
This removes something like a hundred applications to link every time something changes. googletest has facilities (--gtest_filter="...") to run specific tests which handles having individual test suites.

Not sure what you mean by building everything is an annoyance, it was doing that already. Any change in opt, val or the main library would relink all of the test suites.

greg-lunarg:
> Not sure what you mean by building everything is an annoyance

Previously, through VS at least, I could build a single test target. That seems to no longer be an option.

As I said in the previous message, the frequency of rebuilding all tests will likely not be prohibitive.

s-perron:
You can still just build the single executable `test_opt`.  It will take as much time compile and link as a single small test.

I like this better because the IDE I use, clion, integrates well with googletest.  I can run the whole suite.  It shows me the individual tests that fail, and then I can rerun just a single test or group of tests.  With the multiple executables, I had extra work to do to find the right exe when something failed.

Issue #2018, opened 2018-10-30
spirv-opt: --eliminate-dead-code-aggressive can make OpUnreachable reachable


afd:
[This archive](https://github.com/KhronosGroup/SPIRV-Tools/files/2529819/spirvopt-bug.zip) contains the files necessary to reproduce the bug.

Files were generated as follows:

```
glslangValidator -V shader.frag -o shader.spv
spirv-opt shader.spv -o partlyoptimized.spv --eliminate-dead-branches --inline-entry-points-exhaustive
spirv-opt partlyoptimized.spv -o bad.spv --eliminate-dead-code-aggressive
```

Applying spirv-dis to bad.spv, we see that this block:

```
         %43 = OpLabel
               OpUnreachable
```

is reachable.  It was not reachable in partlyoptimized.spv (in that binary it exists but has no incoming edges).

```
spirv-opt --version
SPIRV-Tools v2018.6-dev v2018.5-83-g1c1e749f
```

Found using [GraphicsFuzz](https://github.com/google/graphicsfuzz).

greg-lunarg:
I can look at this since I touched this last.

greg-lunarg:
Pretty sure that partlyoptimized.spv is invalid: it contains a branch in a loop to a block outside the loop (%46) that is not the loop's merge block.

(Note that the most recent stable version of spirv-val does not catch this problem. This should probably be filed as an issue).

Next I will look at how partlyoptimized.spv is created and see where the invalid code is created.

greg-lunarg:
So it appears that inlining is generating the invalid code.

This inliner was not designed to handle this particular case: where the inlined routine had a return in a loop. That is why the merge-return pass was designed: to run before the inliner and get rid of this case.

Still, the inliner is supposed to check for this case and just not inline. I will check and see why that is not happening.

s-perron:
If we add merge return before inlining does it fix this problem?  I don't think it will.  Like inlining, it is looking for multiple OpReturn instructions.  We should open an separate issue for merge-return.

afd:
> Pretty sure that partlyoptimized.spv is invalid: it contains a branch in a loop to a block outside the loop (%46) that is not the loop's merge block.
> 
> (Note that the most recent stable version of spirv-val does not catch this problem. This should probably be filed as an issue).

Done: #2028.

s-perron:
Opened #2038 to handle the merge return improvement.  This issue can be closed now.

PR #2025, opened 2018-10-31
Don't assume one return means function can be inlined. (#2018)

greg-lunarg:


s-perron:
We might need to consider the case where a return is inside of a selection construct.  for example

```
OpFunction
...
OpSelectionMerge %m
OpBranchConditional %true %bb %m
%bb = OpLabel
OpReturn
%m = OpLabel
OpUnreachable
...
OpFunctionEnd
```

If that is the only return, then inlining will generate invalid code as well.

greg-lunarg:
Sorry. I missed this comment. Will respond shortly.

greg-lunarg:
Changed to WIP. I am going to add a couple test cases.

Note changes to a couple existing tests: these do not require a single-trip loop wrapper and under the new regime, the loop is no longer generated.

source/opt/inline_pass.cpp
```diff
@@ -601,6 +584,48 @@ InlinePass::GetBlocksFunction InlinePass::StructuredSuccessorsFunction() {
   };
 }
 
+bool InlinePass::HasNoReturnInSelection(Function* func) {
```
s-perron: You might be able to use `GetStructuedCFGAnalaysis` in the ir context.  I added it because this same code has already been written too often.  You can get the analysis, and do a single run through of all of the blocks in any order.

greg-lunarg: Done.

greg-lunarg:
Looks like the CI site went down. I can't look at any results, passing or failing. When I can look at passing results, I will resubmit.

PR #2026, opened 2018-11-01
Enable precompiled headers for spirv-tools(-shared) and some unit tests

jeffbolznv:
My original goal here was to enable this for SPIRV-Tools.lib and SPIRV-Tools-Shared.lib, since other projects like glslang and CTS build those. But it was easy enough to also enable it for the larger unit tests.

Total build time for Debug on my 6-core system goes from 4 minutes 40 seconds to 3 minutes 13 seconds.


test/CMakeLists.txt
```diff
@@ -30,12 +30,16 @@ endif()
...
     set(multi_value_args SRCS LIBS ENVIRONMENT)
     cmake_parse_arguments(
       ARG "" "${one_value_args}" "${multi_value_args}" ${ARGN})
     set(target test_${ARG_TARGET})
-    add_executable(${target} ${ARG_SRCS})
+    set(SRC_COPY ${ARG_SRCS})
+    if (DEFINED ARG_PCH_FILE)
+      PCH(SRC_COPY ${ARG_PCH_FILE})
+    endif()
+    add_executable(${target} ${SRC_COPY})
```
jeffbolznv: I couldn't figure out a way to do this without making a copy (SRC_COPY = ${ARG_SRCS}). But I'm still a novice at cmake, and there's probably a way to avoid this. Any suggestions?

jeffbolznv:
I eventually gave up on getting ninja to work. It only worked before because I had a single library using PCH. The way it puts all intermediate files in a single output folder causes no end of problems, and the workarounds I tried just led to other problems.

test/CMakeLists.txt
s-perron: I'm a novice with cmake as well, but why can't you just use `${ARG_SRCS}` in both places you use `${SRC_COPY}`?

dj2:
So, submitting this would mean that we can't use ninja on Windows?

source/CMakeLists.txt
```diff
@@ -340,6 +345,8 @@ set_source_files_properties(
   ${CMAKE_CURRENT_SOURCE_DIR}/software_version.cpp
   PROPERTIES OBJECT_DEPENDS "${SPIRV_TOOLS_BUILD_VERSION_INC}")
 
+PCH(SPIRV_SOURCES pch.cpp)
```
s-perron: There are too many files named pch.cpp and pch.h now.  It might be best to rename them so they identify their intended use.  For example, the test/opt/ one could be called `test_opt_pch*`.  That way it will be easier to make sure we do not accidentally pick up the wrong file.

CMakeLists.txt
```diff
@@ -212,6 +212,18 @@ if (${SPIRV_CHECK_CONTEXT})
   add_definitions(-DSPIRV_CHECK_CONTEXT)
 endif()
 
+# Precompiled header macro. Parameters are source file list and filename for pch cpp file.
+macro(PCH SRCS PCHCPP)
```
s-perron: I like the use of the macro.  Good idea.

CMakeLists.txt
```diff
@@ -212,6 +212,18 @@ if (${SPIRV_CHECK_CONTEXT})
...
+  if(MSVC AND CMAKE_GENERATOR MATCHES "^Visual Studio")
+    set(PCH_NAME "$(IntDir)\\pch.pch")
```
s-perron: Is it possible that you will be writing to the same `pch.pch` for each of the `pch.cpp` files you have?  Another reason to have different names.

test/CMakeLists.txt
jeffbolznv: I think it is because ARG_SRCS is a "multi value" thing (a collection of source file names), and I need to actually turn it into a single variable that is a list. I tried a few other things and none worked.

CMakeLists.txt
jeffbolznv: In Visual Studio, each project's pch.pch goes to a unique output directory. That doesn't happen with ninja, and was the primary source of problems with that build.

source/CMakeLists.txt
jeffbolznv: I can give them more unique names, though it doesn't actually help with the Ninja problems (I had tried that last night).

jeffbolznv: I've renamed all the pch files to reflect their directories.

test/CMakeLists.txt
jeffbolznv: I looked into this some more. I think the issue is that the PCH macro appends to the list that is its first argument, and you can't do that using a reference to the function parameter. So this makes a bit more sense now.

PR #2036, opened 2018-11-07
Validate that VertexId and InstanceId are not allowed in Vulkan.

sheredom:
The Vulkan specification does not permit use of the `VertexId` and `InstanceId` BuiltIn decorations, so add a check to ensure they are not being used when the target environment is Vulkan.


alan-baker:
Please fix the compile errors and I'll kick off new builds.

sheredom:
@alan-baker it should be fixed now!

Issue #2027, opened 2018-11-01
A proposal to preserve debug line info when passes delete instructions


greg-lunarg:
As part of the bindless validation and instrumentation, I have started looking more at debug/line info in spirv-opt as we are getting feedback that source-level information on the location of errors would be highly desirable.

I have noticed that SPIR-V line information (OpLine and OpNoLine) instructions are not propagated/cloned to all following affected instructions in the current spirv-opt framework, but essentially just attach to their nearest following non-line instruction. If that instruction is deleted during optimization passes, the line information is also deleted and lost for all subsequent, affected, non-deleted instructions.

For example, consider the following:

         %30 = OpLabel
               OpLine %1 31 0
         %47 = OpLoad %uint %u
         %49 = OpAccessChain %_ptr_UniformConstant_42 %g_tColor %47
         %50 = OpLoad %42 %49
         %54 = OpLoad %51 %g_sAniso
         %56 = OpSampledImage %55 %50 %54
         %58 = OpAccessChain %_ptr_Function_v2float %i %int_0
         %59 = OpLoad %v2float %58
         %60 = OpImageSampleImplicitLod %v4float %56 %59
         %62 = OpAccessChain %_ptr_Function_v4float %ps_output %int_0
               OpStore %62 %60

When the load of function-scope variable %u at the top of the block is ultimately optimized away, the OpLine will be deleted along with it, losing the source line information for the remaining instructions.

My current thinking to solving this is the following: two different passes.

One pass propagates/clones line instructions into all instructions they affect. This means every instruction has source info attached to it. It would generally be run before any optimization passes where the user wishes to maintain source line information. This makes line info easier to maintain as instructions are deleted, moved or replicated during optimization.

The other pass eliminates any line-instructions that are redundant. It would be run after optimization and before output of the SPIR-V to reduce the file size of the SPIR-V as much as possible while still retaining all line info for the module.

I think the analyses will be pretty straightforward as the scope of the line instructions ends at every block, thus restricting the analyses to single blocks.

There are alternatives solutions which might be more efficient in maximum space usage, but this solution seemed to be the simplest while maintaining the flexibility of the current infrastructure.

s-perron:
Yes, I agree with this plan.  In general, the way to solve this problem is that every instruction needs to somehow store its own source and file information.  In other compilers, it has been an index into a side table like the llvm metadata.  These types of passes will be needed regardless of how the data is stored.

[...]

PR #2039, opened 2018-11-10
Add passes to propagate and eliminate redundant line instructions (#2027).

greg-lunarg:
These are bookend passes designed to help preserve line information
across passes which delete, move and clone instructions. The propagation
pass attaches a debug line instruction to every instruction based on
SPIR-V line propagation rules. It should be performed before optimization.
The redundant line elimination pass eliminates all line instructions
which match the previous line instruction. This pass should be performed
at the end of optimization to reduce physical SPIR-V file size.


greg-lunarg:
This is submitted to address #2027

s-perron:
LGTM.

PR #2044, opened 2018-11-13
Validate variable types for UniformConstant storage in Vulkan (#2008)

### LINKED ISSUE #2008: Vulkan UniformConstant variables only support opaque types
> Variables identified with the UniformConstant storage class are used only as handles to refer to opaque resources. Such variables must be typed as OpTypeImage, OpTypeSampler, OpTypeSampledImage, or an array of one of these types.

zoddicus:
From the Vulkan 1.1 spec 14.5.2:
  Variables identified with the UniformConstant storage class are used
  only as handles to refer to opaque resources. Such variables must be
  typed as OpTypeImage, OpTypeSampler, OpTypeSampledImage, or an array
  of one of these types.

Fixes #2008

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -345,6 +345,7 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
         break;
       default:
         return _.diag(SPV_ERROR_INVALID_ID, inst)
+               << "From Vulkan spec, section 14.5.2:\n"
                << "Variables identified with the UniformConstant storage class "
                << "are used only as handles to refer to opaque resources. Such "
                << "variables must be typed as OpTypeImage, OpTypeSampler, "
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -328,19 +328,13 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
   // Vulkan spec 14.5.2
   if (spvIsVulkanEnv(_.context()->target_env) &&
       storage_class == SpvStorageClassUniformConstant) {
-    const auto pointer_type_index = 2;
-    const auto pointer_type_id =
-        result_type->GetOperandAs<uint32_t>(pointer_type_index);
-    auto variable_type = _.FindDef(pointer_type_id);
+    auto variable_type = _.FindDef(result_type->GetOperandAs<uint32_t>(2));
     auto variable_type_opcode = variable_type->opcode();
 
     // If the variable is actually an array extract the element type.
     if (variable_type_opcode == SpvOpTypeArray ||
         variable_type_opcode == SpvOpTypeRuntimeArray) {
-      const auto element_index = 1;
-      const auto element_id =
-          variable_type->GetOperandAs<uint32_t>(element_index);
-      variable_type = _.FindDef(element_id);
+      variable_type = _.FindDef(variable_type->GetOperandAs<uint32_t>(1));
       variable_type_opcode = variable_type->opcode();
     }
 
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -19,6 +19,7 @@
 #include <vector>
 
 #include "source/opcode.h"
+#include "source/spirv_target_env.h"
 #include "source/val/instruction.h"
 #include "source/val/validation_state.h"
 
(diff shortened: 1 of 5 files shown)


source/val/validate_memory.cpp
```diff
@@ -323,6 +324,40 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
              << "Classes: Workgroup, CrossWorkgroup, Private, and Function";
     }
   }
+
+  // Vulkan spec 14.5.2
+  if (spvIsVulkanEnv(_.context()->target_env) &&
+      storage_class == SpvStorageClassUniformConstant) {
+    const auto pointer_type_index = 2;
+    const auto pointer_type_id =
+        result_type->GetOperandAs<uint32_t>(pointer_type_index);
```
dj2: I'd just inline the 2 here (I know we do this in other places, but I don't think it adds much)

source/val/validate_memory.cpp
```diff
@@ -323,6 +324,40 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
...
+    auto variable_type = _.FindDef(pointer_type_id);
+    auto variable_type_opcode = variable_type->opcode();
+
+    // If the variable is actually an array extract the element type.
+    if (variable_type_opcode == SpvOpTypeArray ||
+        variable_type_opcode == SpvOpTypeRuntimeArray) {
+      const auto element_index = 1;
```
dj2: ditto

test/val/CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ add_spvtools_unittest(TARGET val_ijklmnop
        val_layout_test.cpp
        val_literals_test.cpp
        val_logicals_test.cpp
+       val_memory_test.cpp
```
dj2: Is this file missing?

source/val/validate_memory.cpp

test/val/CMakeLists.txt
zoddicus: yup, added

dj2:
This LG to me, but leaving the approval to @alan-baker.

source/val/validate_memory.cpp
```diff
@@ -323,6 +324,34 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
...
+
+  // Vulkan spec 14.5.2
+  if (spvIsVulkanEnv(_.context()->target_env) &&
+      storage_class == SpvStorageClassUniformConstant) {
+    auto variable_type = _.FindDef(result_type->GetOperandAs<uint32_t>(2));
+    auto variable_type_opcode = variable_type->opcode();
+
+    // If the variable is actually an array extract the element type.
+    if (variable_type_opcode == SpvOpTypeArray ||
+        variable_type_opcode == SpvOpTypeRuntimeArray) {
```
alan-baker: According the SPIR-V environment for Vulkan, only SSBOs (StorageBuffer with Block or Uniform with BufferBlock) can have a runtime array so drop this condition.

source/val/validate_memory.cpp
```diff
@@ -323,6 +324,34 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
...
+    }
+
+    switch (variable_type_opcode) {
+      case SpvOpTypeImage:
+      case SpvOpTypeSampler:
+      case SpvOpTypeSampledImage:
+        break;
+      default:
+        return _.diag(SPV_ERROR_INVALID_ID, inst)
+               << "Variables identified with the UniformConstant storage class "
```
alan-baker: I generally like to mention this is a Vulkan specific check so the user knows where to look.

zoddicus: Updated the comment at the beginning of this block to be more explicit.

source/val/validate_memory.cpp

source/val/validate_memory.cpp
alan-baker: I meant this error message. The user should have to read this code to know its a Vulkan-specific check.

zoddicus: my bad, I have added a reference to the spec in the error message

test/val/val_memory_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+%func = OpFunction %void None %functy
+%1 = OpLabel
+OpReturn
+OpFunctionEnd
+)";
+  CompileSuccessfully(spirv.c_str(), SPV_ENV_VULKAN_1_1);
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions(SPV_ENV_VULKAN_1_1));
+  EXPECT_THAT(
+      getDiagnosticString(),
+      HasSubstr("Variables identified with the UniformConstant storage class "
```
alan-baker: Please update this to reflect the full message.

test/val/val_memory_test.cpp
```diff
@@ -0,0 +1,133 @@
...
```
alan-baker: Please update this to reflect the full message.

zoddicus: Done

test/val/val_memory_test.cpp
zoddicus: Done

PR #2049, opened 2018-11-14
Allow InstanceId for NV ray tracing

### LINKED ISSUE #2046: Re-enable InstanceId for ray tracing shaders
#2036 disabled InstanceId for Vulkan.
This causes issues in glslang tests: https://github.com/KhronosGroup/glslang/issues/1576

It appears we made a mistake in the Vulkan and SPIR-V specs for NV_ray_tracing and documented the builtin for gl_InstanceId as InstanceIndex (as the rest of the vulkan shaders use) instead of adding InstanceId for the relevant shader stages. We are sorting out the spec issues, but at this point it's looking like we stick with InstanceId for closest-hit, any-hit, and intersection shaders

alan-baker:
Fixes #2046

* Allows InstanceId in the Vulkan environment if MeshShadingNV
capability is specified

### DIFF
--- a/source/val/validate_builtins.cpp
+++ b/source/val/validate_builtins.cpp
@@ -168,7 +168,7 @@ class BuiltInsValidator {
   spv_result_t ValidateVertexIndexAtDefinition(const Decoration& decoration,
                                                const Instruction& inst);
   spv_result_t ValidateVertexIdOrInstanceIdAtDefinition(
-      const Instruction& inst);
+      const Decoration& decoration, const Instruction& inst);
   spv_result_t ValidateWorkgroupSizeAtDefinition(const Decoration& decoration,
                                                  const Instruction& inst);
   // Used for GlobalInvocationId, LocalInvocationId, NumWorkgroups, WorkgroupId.
@@ -2088,8 +2088,11 @@ spv_result_t BuiltInsValidator::ValidateVertexIndexAtDefinition(
 }
 
 spv_result_t BuiltInsValidator::ValidateVertexIdOrInstanceIdAtDefinition(
-    const Instruction& inst) {
-  if (spvIsVulkanEnv(_.context()->target_env)) {
+    const Decoration& decoration, const Instruction& inst) {
+  const SpvBuiltIn label = SpvBuiltIn(decoration.params()[0]);
+  bool allow_instance_id = _.HasCapability(SpvCapabilityMeshShadingNV) &&
+                           label == SpvBuiltInInstanceId;
+  if (spvIsVulkanEnv(_.context()->target_env) && !allow_instance_id) {
     return _.diag(SPV_ERROR_INVALID_DATA, &inst)
            << "Vulkan spec doesn't allow BuiltIn VertexId/InstanceId "
               "to be used.";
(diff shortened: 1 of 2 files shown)


alan-baker:
@dgkoch is MeshShadingNV the right capability to gate this on?

dgkoch:
> @dgkoch is MeshShadingNV the right capability to gate this on?

No - that would enable it for SPV_NV_mesh_shading. 

You want to enable it for the the SPV_NV_ray_tracing extension, which means the RayTracingNV capability. 

Further more, we'd only want to allow them in following execution modes: 
- IntersectionNV
- AnyHitNV
- ClosestHitNV

source/val/validate_builtins.cpp
```diff
@@ -2088,8 +2088,11 @@ spv_result_t BuiltInsValidator::ValidateVertexIndexAtDefinition(
 }
 
 spv_result_t BuiltInsValidator::ValidateVertexIdOrInstanceIdAtDefinition(
-    const Instruction& inst) {
-  if (spvIsVulkanEnv(_.context()->target_env)) {
+    const Decoration& decoration, const Instruction& inst) {
+  const SpvBuiltIn label = SpvBuiltIn(decoration.params()[0]);
+  bool allow_instance_id = _.HasCapability(SpvCapabilityRayTracingNV) &&
+                           label == SpvBuiltInInstanceId;
```
dgkoch: I'm still not convinced that this is an adequate check.
From surveying other validation examples in this file, it seems like we'd need to add an "AtReference" version of the check and add in something like 

```
for (const SpvExecutionModel execution_model : execution_models_) {
    switch (execution_model) {
        case SpvExecutionModelIntersectionNV:
        case SpvExecutionModelAnyHitNV:
        case SpvExecutionModelClosestHitNV:
            break; // allowed
       default: 
           // error...
    }
} 
```

And that's where I got bogged down when I investigated adding the change :-/

Ie - having RayTracingNV capability should NOT enable this in existing stages, and should NOT enable it in all new stages that area added under RayTracingNV capability either.

test/val/val_builtins_test.cpp
```diff
@@ -2169,6 +2169,46 @@ OpFunctionEnd
                         "be declared when using BuiltIn FragDepth"));
 }
 
+TEST_F(ValidateBuiltIns, AllowInstanceIdWithMeshShader) {
```
dgkoch: Still says mesh shader here.

source/val/validate_builtins.cpp
alan-baker: My goal is not to validate the Nvidia extension. I'm just trying to avoid failing a shader that should be valid.

dneto0: We can file a followup issue for the more precise check.

dgkoch: > My goal is not to validate the Nvidia extension. I'm just trying to avoid failing a shader that should be valid.

Those seem to be contradictory. 

> We can file a followup issue for the more precise check.

Sure fine.

alan-baker: > Those seem to be contradictory.

They are not. False positives from the validator (invalid module being found valid) are more acceptable than false negatives. This change will eliminate the false negatives, but introduce false positives. I'm ok with that.

PR #2052, opened 2018-11-14
Ensure for OpVariable that result type and storage class operand agree

### LINKED ISSUE #941: OpVariable: Storage class must match storage class of the pointer type

zoddicus:
From SPIR-V spec, section 3.32.8 on OpVariable:
  Its Storage Class operand must be the same as the Storage Class
  operand of the result type.

Fixes #941

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -301,7 +301,7 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
     const auto initializer = _.FindDef(initializer_id);
     const auto is_module_scope_var =
         initializer && (initializer->opcode() == SpvOpVariable) &&
-        (initializer->GetOperandAs<uint32_t>(storage_class_index) !=
+        (initializer->GetOperandAs<SpvStorageClass>(storage_class_index) !=
          SpvStorageClassFunction);
     const auto is_constant =
         initializer && spvOpcodeIsConstant(initializer->opcode());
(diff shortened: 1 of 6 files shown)


test/val/val_memory_test.cpp
```diff
@@ -130,6 +130,52 @@ OpFunctionEnd
...
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(
+      getDiagnosticString(),
+      HasSubstr(
+          "From SPIR-V spec, section 3.32.8 on OpVariable:\n"
+          "Its Storage Class operand must be the same as the Storage Class "
+          "operand of the result type."));
+}
+
+TEST_F(ValidateMemory, MatchingStorageClassesBad) {
```
alan-baker: I think you meant MatchingStorageClassesGood?

source/val/validate_memory.cpp
```diff
@@ -325,6 +325,17 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
     }
   }
 
+  // Check that pointer type and variable type have the same storage class
+  const auto result_storage_class_index = 1;
+  const auto result_storage_class =
+      result_type->GetOperandAs<uint32_t>(result_storage_class_index);
```
alan-baker: Minor: template operand should be SpvStorageClass. Please update `storage_class` above too.

test/val/val_memory_test.cpp
zoddicus: Yup

dneto0:
FYI:  @johnkslang just landed the fixes for the glslang validator variances.  See https://github.com/KhronosGroup/glslang/commit/30cb924538eb5f3a15b378553b7933dafc1a67bc
And I pulled those into google/glslang from KhronosGroup/glslang.

The smoke test would pass now.

PR #2054, opened 2018-11-15
UniformConstant variables can have RuntimeArray, TypeAccelerationStructureNV

dneto0:
Part of the test for this is in the Glslang test suite. :-(


dneto0:
In the interest of getting an SDK out, I'm LGTM this.

PR #2055, opened 2018-11-15
Validate uniform variable type in Vulkan (#1949)

### LINKED ISSUE #1949: Validator does not check type of Uniform/StorageBuffer variables
The Vulkan spec requires that all variables in the `Uniform` and `StorageBuffer` storage classes be structures (or arrays of structures) that are decorated with `Block` or `BufferBlock` (see 14.5.2. Descriptor Set Interface).

The following shader declares a variable in the Uniform storage class that points directly to an integer, which should not validate. Same issue occurs when replacing `Uniform` with `StorageBuffer`.

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %kernel "main"
               OpExecutionMode %kernel LocalSize 1 1 1

               OpDecorate %var DescriptorSet 0
               OpDecorate %var Binding 0

     %voidty = OpTypeVoid
   %kernelty = OpTypeFunction %voidty
      %intty = OpTypeInt 32 0
      %varty = OpTypePointer Uniform %intty
      %value = OpConstant %intty 42

        %var = OpVariable %varty Uniform

     %kernel = OpFunction %voidty None %kernelty
      %label = OpLabel
               OpStore %var %value
               OpReturn
               OpFunctionEnd
```

fjhenigman:
From the Vulkan 1.1 spec 14.5.2:
  Variables identified with the Uniform storage class are used to access
  transparent buffer backed resources. Such variables must be typed as
  OpTypeStruct, or an array of this type.

Fixes #1949


source/val/validate_memory.cpp
```diff
@@ -325,35 +339,34 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
...
-
-    // If the variable is actually an array or runtime-array, extract the
-    // element type.
-    if (variable_type_opcode == SpvOpTypeArray ||
-        variable_type_opcode == SpvOpTypeRuntimeArray) {
-      variable_type = _.FindDef(variable_type->GetOperandAs<uint32_t>(1));
-      variable_type_opcode = variable_type->opcode();
-    }
+  // Vulkan 14.5.2: Check type of UniformConstant and Uniform variables.
+  if (1 && spvIsVulkanEnv(_.context()->target_env)) {
```
alan-baker: Drop the "1 &&"

test/val/val_builtins_test.cpp
```diff
@@ -1200,16 +1200,6 @@ INSTANTIATE_TEST_CASE_P(
         Values(TestResult(SPV_ERROR_INVALID_DATA,
                           "to be used only with Fragment execution model"))), );
 
-INSTANTIATE_TEST_CASE_P(
```
alan-baker: Can this test be salvaged? Would changing the storage class to Workgroup pass the test?

test/val/val_decoration_test.cpp
```diff
@@ -646,119 +646,6 @@ TEST_F(ValidateDecorations, RuntimeArrayOfDescriptorSetsIsAllowed) {
...
 
-// #version 440
-// #extension GL_EXT_nonuniform_qualifier : enable
-// layout(binding = 1) uniform sampler2D s2d[][2];
-// layout(location = 0) in nonuniformEXT int i;
-// void main()
-// {
-//     vec4 v = texture(s2d[i][i], vec2(0.3));
-// }
-TEST_F(ValidateDecorations, RuntimeArrayOfArraysOfDescriptorSetsIsDisallowed) {
```
alan-baker: Can you add some equivalents of these tests in val_memory_test.cpp that get caught be your check?

PR #2056, opened 2018-11-15
Initial commit for spirv-reduce.

afd:
This PR contributes the design of a new tool, spirv-reduce, for shrinking down a SPIR-V binary according to some user-defined "interestingness test" (inspired by the interestingness tests used by C-Reduce).

Currently only a couple of reduction passes are implemented - one that replaces a non-constant operand id with a constant, and another that removes an instruction whose id is not used.

I have re-used infrastructure from spirv-opt where possible, but have not tried to over-unify things; e.g. a reduction pass is fundamentally different from an optimization pass.

### DIFF
--- a/source/reduce/CMakeLists.txt
+++ b/source/reduce/CMakeLists.txt
@@ -1,0 +1,68 @@
+# Copyright (c) 2018 Google Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set(SPIRV_TOOLS_REDUCE_SOURCES
+        change_operand_reduction_opportunity.h
+        operand_to_const_reduction_pass.h
+        reducer.h
+        reduction_opportunity.h
+        reduction_pass.h
+        remove_instruction_reduction_opportunity.h
+        remove_unreferenced_instruction_reduction_pass.h
+
+        change_operand_reduction_opportunity.cpp
+        operand_to_const_reduction_pass.cpp
+        reducer.cpp
+        reduction_opportunity.cpp
+        reduction_pass.cpp
+        remove_instruction_reduction_opportunity.cpp
+        remove_unreferenced_instruction_reduction_pass.cpp
+        )
+
+if(MSVC)
+    # Enable parallel builds across four cores for this lib
+    add_definitions(/MP4)
+
+    # Enable precompiled header
+    if (CMAKE_GENERATOR MATCHES "^Visual Studio")
+        set(PCH_NAME "$(IntDir)\\pch.pch")
+    else()
+        set(PCH_NAME "pch.pch")
+    endif()
+    set_source_files_properties(${SPIRV_TOOLS_REDUCE_SOURCES} PROPERTIES COMPILE_FLAGS "/Yupch.h /FIpch.h /Fp${PCH_NAME} /Zm300" OBJECT_DEPENDS "${PCH_NAME}")
+    set_source_files_properties(pch.cpp PROPERTIES COMPILE_FLAGS "/Ycpch.h /Fp${PCH_NAME} /Zm300" OBJECT_OUTPUTS "${PCH_NAME}")
+    list(APPEND SPIRV_TOOLS_REDUCE_SOURCES "pch.cpp")
+endif()
+
+add_library(SPIRV-Tools-reduce ${SPIRV_TOOLS_REDUCE_SOURCES})
+
+spvtools_default_compile_options(SPIRV-Tools-reduce)
+target_include_directories(SPIRV-Tools-reduce
+        PUBLIC ${spirv-tools_SOURCE_DIR}/include
+        PUBLIC ${SPIRV_HEADER_INCLUDE_DIR}
+        PRIVATE ${spirv-tools_BINARY_DIR}
+        )
+# The reducer reuses a lot of functionality from the SPIRV-Tools library.
+target_link_libraries(SPIRV-Tools-reduce
+        PUBLIC ${SPIRV_TOOLS}
+        PUBLIC SPIRV-Tools-opt)
+
+set_property(TARGET SPIRV-Tools-reduce PROPERTY FOLDER "SPIRV-Tools libraries")
+spvtools_check_symbol_exports(SPIRV-Tools-reduce)
+
+if(ENABLE_SPIRV_TOOLS_INSTALL)
+    install(TARGETS SPIRV-Tools-reduce
+            RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+            LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+            ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+endif(ENABLE_SPIRV_TOOLS_INSTALL)
--- a/source/reduce/CMakeLists.txt
+++ b/source/reduce/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 Google Inc.
+# Copyright (c) 2018 Google LLC
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/source/reduce/CMakeLists.txt
+++ b/source/reduce/CMakeLists.txt
@@ -30,39 +30,31 @@ set(SPIRV_TOOLS_REDUCE_SOURCES
         )
 
 if(MSVC)
-    # Enable parallel builds across four cores for this lib
-    add_definitions(/MP4)
-
-    # Enable precompiled header
-    if (CMAKE_GENERATOR MATCHES "^Visual Studio")
-        set(PCH_NAME "$(IntDir)\\pch.pch")
-    else()
-        set(PCH_NAME "pch.pch")
-    endif()
-    set_source_files_properties(${SPIRV_TOOLS_REDUCE_SOURCES} PROPERTIES COMPILE_FLAGS "/Yupch.h /FIpch.h /Fp${PCH_NAME} /Zm300" OBJECT_DEPENDS "${PCH_NAME}")
-    set_source_files_properties(pch.cpp PROPERTIES COMPILE_FLAGS "/Ycpch.h /Fp${PCH_NAME} /Zm300" OBJECT_OUTPUTS "${PCH_NAME}")
-    list(APPEND SPIRV_TOOLS_REDUCE_SOURCES "pch.cpp")
+  # Enable parallel builds across four cores for this lib
+  add_definitions(/MP4)
 endif()
+
+spvtools_pch(SPIRV_TOOLS_REDUCE_SOURCES pch_source_reduce)
 
 add_library(SPIRV-Tools-reduce ${SPIRV_TOOLS_REDUCE_SOURCES})
 
 spvtools_default_compile_options(SPIRV-Tools-reduce)
 target_include_directories(SPIRV-Tools-reduce
-        PUBLIC ${spirv-tools_SOURCE_DIR}/include
-        PUBLIC ${SPIRV_HEADER_INCLUDE_DIR}
-        PRIVATE ${spirv-tools_BINARY_DIR}
-        )
+  PUBLIC ${spirv-tools_SOURCE_DIR}/include
+  PUBLIC ${SPIRV_HEADER_INCLUDE_DIR}
+  PRIVATE ${spirv-tools_BINARY_DIR}
+)
 # The reducer reuses a lot of functionality from the SPIRV-Tools library.
 target_link_libraries(SPIRV-Tools-reduce
-        PUBLIC ${SPIRV_TOOLS}
-        PUBLIC SPIRV-Tools-opt)
+  PUBLIC ${SPIRV_TOOLS}
+  PUBLIC SPIRV-Tools-opt)
 
 set_property(TARGET SPIRV-Tools-reduce PROPERTY FOLDER "SPIRV-Tools libraries")
 spvtools_check_symbol_exports(SPIRV-Tools-reduce)
 
 if(ENABLE_SPIRV_TOOLS_INSTALL)
-    install(TARGETS SPIRV-Tools-reduce
-            RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
-            LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-            ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+  install(TARGETS SPIRV-Tools-reduce
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 endif(ENABLE_SPIRV_TOOLS_INSTALL)
--- a/source/reduce/change_operand_reduction_opportunity.cpp
+++ b/source/reduce/change_operand_reduction_opportunity.cpp
@@ -20,7 +20,8 @@ namespace reduce {
 bool ChangeOperandReductionOpportunity::PreconditionHolds() {
   // Check that the instruction still has the original operand.
   return inst_->NumOperands() > operand_index_ &&
-         inst_->GetOperand(operand_index_).words[0] == original_id_;
+         inst_->GetOperand(operand_index_).words[0] == original_id_ &&
+         inst_->GetOperand(operand_index_).type == original_type_;
 }
 
 void ChangeOperandReductionOpportunity::Apply() {
--- a/source/reduce/change_operand_reduction_opportunity.cpp
+++ b/source/reduce/change_operand_reduction_opportunity.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Google Inc.
+// Copyright (c) 2018 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/source/reduce/change_operand_reduction_opportunity.cpp
+++ b/source/reduce/change_operand_reduction_opportunity.cpp
@@ -1,0 +1,31 @@
+// Copyright (c) 2018 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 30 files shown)


dnovillo:
This is awesome!  Thanks!  Sadly, I won't be able to look at this for the next couple of weeks, but I'm sure someone else will get the review started.

source/reduce/operand_to_const_reduction_pass.cpp
```diff
@@ -0,0 +1,64 @@
...
+std::vector<std::unique_ptr<ReductionOpportunity>>
+OperandToConstReductionPass::GetAvailableOpportunities(
+    opt::IRContext* context) const {
+  std::vector<std::unique_ptr<ReductionOpportunity>> result;
+  assert(result.empty());
+  for (const auto& constant : context->GetConstants()) {
+    for (auto& function : *context->module()) {
+      for (auto& block : function) {
+        for (auto& inst : block) {
+          for (uint32_t i = 0; i < inst.NumOperands(); i++) {
```
alan-baker: You can replace this loo and the check against operand type with:
```
inst.ForEachInId([&result,context](uint32_t* id_ptr) { /* code in the operand type if */});
```

source/reduce/operand_to_const_reduction_pass.cpp
```diff
@@ -0,0 +1,64 @@
...
+            const auto& operand = inst.GetOperand(i);
+            if (operand.type == SPV_OPERAND_TYPE_ID) {
+              const auto id = operand.words[0];
+              auto def = context->get_def_use_mgr()->GetDef(id);
+              if (context->get_constant_mgr()->GetConstantFromInst(def)) {
```
alan-baker: The constant manager is unable to represent all constants. For example, structs with some uninitialized members (OpUndef) cannot represented correctly. Depending on how you want to handle those you could instead query `spvOpcodeIsConstant(def->opcode())` or something similar (see source/opcode.h)

source/reduce/operand_to_const_reduction_pass.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace reduce {
+
+using namespace opt;
+
+std::vector<std::unique_ptr<ReductionOpportunity>>
+OperandToConstReductionPass::GetAvailableOpportunities(
+    opt::IRContext* context) const {
+  std::vector<std::unique_ptr<ReductionOpportunity>> result;
+  assert(result.empty());
+  for (const auto& constant : context->GetConstants()) {
```
alan-baker: It might be clearer to make this the innermost loop.

source/reduce/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) 2018 Google Inc.
```
alan-baker: These should be Google LLC.

source/reduce/reducer.cpp
```diff
@@ -0,0 +1,164 @@
...
+  impl_->current_pass = impl_->passes.begin();
+
+  uint32_t current_step = 0;
+
+  std::vector<uint32_t> current = std::move(binary_in);
+
+  // Initial state should be interesting.
+  assert(impl_->is_interesting(current));
+
+  for (;;) {
```
alan-baker: Instead of the following assert, why not enforce bounds on the loop?

source/reduce/reduction_pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+  std::vector<std::unique_ptr<ReductionOpportunity>> opportunities =
+      GetAvailableOpportunities(context.get());
+
+  if (!is_initialized_) {
+    assert(granularity_ == 0);
+    assert(index_ == 0);
+    is_initialized_ = true;
+    index_ = 0;
+    granularity_ = std::max((uint32_t)1, (uint32_t)opportunities.size());
```
alan-baker: If you remove the `assert(granularity_ > 0)` below (or move after the subsequent if) you can drop this max. It appears you would still always get a minimum granularity of 1.

source/reduce/remove_unreferenced_instruction_reduction_pass.cpp
```diff
@@ -0,0 +1,63 @@
...
+      for (auto& inst : block) {
+        if (context->get_def_use_mgr()->NumUses(&inst) == 0) {
+          switch (inst.opcode()) {
+            case SpvOpBranch:
+            case SpvOpBranchConditional:
+            case SpvOpLoopMerge:
+            case SpvOpSelectionMerge:
+            case SpvOpReturn:
+            case SpvOpSwitch:
+              // TODO: this should ultimately capture all opcodes that relate to
```
alan-baker: Missing OpKill, OpReturnValue and OpUnreachable.

Potential alternative code to check:
```
If (!(spvOpcodeIsBlockTerminator(inst.opcode() || inst.opcode() == SpvOpSelectionMerge || inst.opcode() == SpvOpLoopMerge))
```

alan-baker:
FYI @afd, I've launched a full set of builds.

source/reduce/operand_to_const_reduction_pass.cpp
afd: That's actually how I structured it originally, because that would make reduction opportunities for the same operand adjacent to one another.  If two reduction opportunities refer to the same operand, they can't both be applied (because one disables the other).  The reducer takes a greedy approach, as per delta debugging, where blocks of adjacent reduction opportunities are tried, with the block size exponentially shrinking.  Thus it's best to structure things so that adjacent opportunities are likely to be compatible, hence this loop structure.  I'll reflect this in the code comments.

source/reduce/CMakeLists.txt
afd: Thanks.  (Note that there seem to be quite a few instances of this in the existing code.)

source/reduce/remove_unreferenced_instruction_reduction_pass.cpp
afd: Done, thanks!

source/reduce/reducer.cpp
afd: Good plan - done.

source/reduce/reduction_pass.cpp
afd: Agreed, done.

source/reduce/CMakeLists.txt
afd: Fixed (the instances introduced in this PR).

source/reduce/operand_to_const_reduction_pass.cpp
afd: The problem I see here is that I use the loop counter "i" to know the index of the operand being changed when I create a ChangeOperandReductionOpportunity.

With the lambda approach I would get a pointer to the operand id, but wouldn't know (or at least not for free) which index this is at.

The choice seems to be to stick with the loop approach, or to identify an operand id in ChangeOperandReductionOpportunity by its pointer value.  I'm fine with either but slightly prefer the current index-based loop approach.  Let me know what you think.

source/reduce/operand_to_const_reduction_pass.cpp
afd: spvOpCodeIsConstant indeed looks good for this purpose - done.

source/reduce/operand_to_const_reduction_pass.cpp
afd: Resolving this by adding an explanatory comment in the code.

source/reduce/operand_to_const_reduction_pass.cpp
afd: Resolving for now having added a comment to the code to explain, but do please unresolve if there's a better approach; I'm keen on a functional style where possible.

alan-baker: In that case I would suggest checking code similar to the implementation of Instruction::WhileEachInId. Your current check misses SPV_OPERAND_TYPE_SCOPE_ID and SPV_OPERAND_TYPE_MEMORY_SEMANTICS_ID. These are required to be constants in the presence of the Shader capability, but they are effectively instruction ids.

source/reduce/reducer.cpp
```diff
@@ -0,0 +1,154 @@
...
+  return true;
+}
+
+void Reducer::AddReductionPass(
+    std::unique_ptr<ReductionPass>&& reduction_pass) {
+  impl_->passes.push_back(std::move(reduction_pass));
+}
+
+std::vector<uint32_t> Reducer::Impl::ApplyReduction(
+    const std::vector<uint32_t>& binary) {
```
s-perron: I find this function very hard to read.  The state changes that are stored in the class, and take place in `ApplyReduction` and in the caller `Run` mean I have to keep going back and forth between the two to understand what this function is doing.  I have not thought it through fully.  However, I could see something like:

```
bool something_chaged = false;
uint32_t num_changes = 0;
while(something_changed && num_changes < options->step_limit) {
  something_chaged = false;
  for( auto& pass : impl_->passes ) {
    auto maybe_result = (*current_pass)->ApplyReduction(binary);
    if (MadeProgress(maybe_result)) {
      something_changed = true;
      ++num_changes;
      // Plus any other book keeping.
      if (num_changes >= options->step_limit) break;
    }
  }
}
```

This might be easier to follow than have a function that iterates over the passes, but may leave and resume the iteration the next time it is called.

source/reduce/reduction_pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+#include "source/opt/build_module.h"
+
+namespace spvtools {
+namespace reduce {
+
+std::vector<uint32_t> ReductionPass::ApplyReduction(
+    const std::vector<uint32_t>& binary) {
+  std::unique_ptr<opt::IRContext> context =
+      BuildModule(target_env_, consumer_, binary.data(), binary.size());
```
s-perron: Is there a reason you want to build the module every time you apply a reduction pass?  If this is done a lot, it could end up being time consuming.  At the very least add a comment explaining why you rebuild the module and context every time.

source/reduce/reduction_pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+  index_ += granularity_;
+
+  std::vector<uint32_t> result;
+  context->module()->ToBinary(&result, false);
+  return result;
+}
+
+void ReductionPass::SetMessageConsumer(MessageConsumer consumer) {
+  assert (is_initialized_);
```
s-perron: Why can't we set the message consumer before it is initialized?  It is initialized during the first call to `ApplyReduction`.  That is after the build it built and the consumer might be needed there.  What if the build fails because the binary it invalid or it hits an internal compiler error?

source/reduce/reduction_pass.cpp
```diff
@@ -0,0 +1,77 @@
...
+  consumer_ = std::move(consumer);
+}
+
+bool ReductionPass::ReachedMinimumGranularity() const {
+  assert (is_initialized_);
```
s-perron: If it is not initialized couldn't you just return false?  Couldn't we think of the uninitialized granularity as infinite?  This way we don't put restrictions like `ApplyReduction` must be call before you can call `ReachedMinimumGranularity`.

source/reduce/remove_unreferenced_instruction_reduction_pass.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+std::vector<std::unique_ptr<ReductionOpportunity>>
+RemoveUnreferencedInstructionReductionPass::GetAvailableOpportunities(
+    opt::IRContext* context) const {
+  std::vector<std::unique_ptr<ReductionOpportunity>> result;
+
+  for (auto& function : *context->module()) {
+    for (auto& block : function) {
+      for (auto& inst : block) {
+        if (context->get_def_use_mgr()->NumUses(&inst) > 0) {
```
s-perron: Something to think about in the future.  Decoration and OpName instruction are considered uses of an id.  They could force lots of code to stick around that is not needed to reproduce the problem.

source/reduce/change_operand_reduction_opportunity.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "change_operand_reduction_opportunity.h"
+
+namespace spvtools {
+namespace reduce {
+
+bool ChangeOperandReductionOpportunity::PreconditionHolds() {
+  // Check that the instruction still has the original operand.
+  return inst_->NumOperands() > operand_index_ &&
+         inst_->GetOperand(operand_index_).words[0] == original_id_;
```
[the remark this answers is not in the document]
afd: Yes, that's a good safe-guard to add too - done.

source/reduce/reducer.cpp
```diff
@@ -0,0 +1,154 @@
...
+        Reducer::InterestingnessFunction interestingness_function) {
+  impl_->interestingness_function = std::move(interestingness_function);
+}
+
+bool Reducer::Run(std::vector<uint32_t>&& binary_in,
+                  std::vector<uint32_t>& binary_out,
+                  spv_const_reducer_options options) const {
+  impl_->current_pass = impl_->passes.begin();
+
+  std::vector<uint32_t> current = std::move(binary_in);
```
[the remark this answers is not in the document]
afd: Yes!  Especially since "current_pass" is also mentioned nearby.  Done.

source/reduce/reduction_pass.cpp
afd: I'll add this comment:

  // We represent modules as binaries because (a) attempts at reduction need to end up in binary form to be passed on
  // to SPIR-V-consuming tools, and (b) when we apply a reduction step we need to do it on a fresh version of the
  // module as if the reduction step proves to be uninteresting we need to backtrack; re-parsing from binary provides
  // a very clean way of cloning the module.

source/reduce/reduction_pass.cpp
afd: I did wonder about initializing granularity to be the max uint32_t value, and doing away with the need for is_initialized_.  I'll do that and see how it looks.

afd: OK; I think it's cleaner to keep the notion of "is_initialized_" after all, but indeed regard an uninitialized pass as having unbounded granularity.

source/reduce/remove_unreferenced_instruction_reduction_pass.cpp
afd: Great; noted for future (will file an issue if/when we make the reducer a standard component - noted locally for now.)

source/reduce/reduction_pass.cpp
afd: Good catch - I think my intention had been to require the consumer to be set before initialization (i.e., I had intended to assert !is_initialized_).  It turned out that message consumer setting was also being done in the wrong place in reduce.cpp.  Fixed both (deciding to actually remove this assertion, so that the message consumer could be changed if needed).

source/reduce/reducer.cpp
afd: In latest commit I have reworked this a lot to move all of the functionality into one method, getting rid of the member variables that were tracking state, and making the various loop exit conditions as explicit as possible.

source/reduce/reducer.cpp
```diff
@@ -54,100 +52,98 @@ void Reducer::SetInterestingnessFunction(
...
+          // Success!  The binary produced by this reduction step is
+          // interesting, so make it the binary of interest henceforth, and
+          // note that it's worth doing another round of reduction passes.
+          impl_->consumer(SPV_MSG_INFO, nullptr, {},
+                  "Reduction step succeeded.");
+          current_binary = std::move(maybe_result);
+          another_round_worthwhile = true;
+        }
+        // Bail out if the reduction step limit has been reached.
+      } while (!impl_->ReachedStepLimit(reductions_applied, options));
```
s-perron: Thanks for this change.  This is a lot easier to follow.

tools/reduce/reduce.cpp
```diff
@@ -0,0 +1,229 @@
...
+// Check that the std::system function can actually be used.
+bool CheckExecuteCommand() {
+  int res = std::system(nullptr);
+  return res != 0;
+}
+
+// Execute a command using the shell. Redirects to stdout.txt and stderr.txt in
+// |temp_dir|. |temp_dir| should end with a slash.
+int ExecuteCommand(const std::string& command) {
+  errno = 0;
```
s-perron: Seems like android does not like errno: https://source.cloud.google.com/results/invocations/31c3846c-3c7d-4b50-943c-94780d94d0f7/targets/graphics_shader_compiler%2Fspirv-tools%2Fandroid-release%2Fpresubmit/log

tools/reduce/reduce.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+  if (!CheckExecuteCommand()) {
+    std::cerr << "could not find shell interpreter for executing a command"
+              << std::endl;
+    return 2;
+  }
+
+  Reducer reducer(target_env);
+
+  reducer.SetInterestingnessFunction([&](std::vector<uint32_t> binary,
```
s-perron: We don't like the implicit capture.  List out the variables you need explicitly.

tools/reduce/reduce.cpp
```diff
@@ -0,0 +1,229 @@
...
+          uint32_t reductions_applied) -> bool {
+    std::stringstream ss;
+    ss << "temp_" << std::setw(4) << std::setfill('0') << reductions_applied
+        << ".spv";
+    const auto spv_file = ss.str();
+    const std::string command = std::string(interestingness_test) + " "
+            + spv_file;
+    assert(WriteFile(spv_file.c_str(), "wb", &binary[0], binary.size()));
```
s-perron: Asserts are removed in the release version, so the file will not be written in the release version.  Better to split this into two statements.  One that writes and saves the return values.  Then assert on the returned value.

tools/reduce/reduce.cpp
afd: Done.

tools/reduce/reduce.cpp
afd: Good catch!  Done.

tools/reduce/reduce.cpp
```diff
@@ -0,0 +1,231 @@
...
+  reducer.SetInterestingnessFunction(
+      [interestingness_test](std::vector<uint32_t> binary,
+                             uint32_t reductions_applied) -> bool {
+        std::stringstream ss;
+        ss << "temp_" << std::setw(4) << std::setfill('0') << reductions_applied
+           << ".spv";
+        const auto spv_file = ss.str();
+        const std::string command =
+            std::string(interestingness_test) + " " + spv_file;
+        auto write_file_succeeded =
```
s-perron: We also don't like unused variables.  You can add a dummy use for this just before the assert:

```
(void)(write_file_succeeded)
```

s-perron: Looks like this is the only issue that is still causing the builds to fail.  All of the debug builds have passed.

afd: Done.

tools/reduce/reduce.cpp
afd: Fixed (by including the right header file).

alan-baker:
Don't worry about those glslang failures in the smoketest. They are unrelated to this PR.

PR #2074, opened 2018-11-16
Validator: Support VK_EXT_scalar_block_layout

dneto0:
Adds validator option to specify scalar block layout rules.

Both VK_KHR_relax_block_layout and VK_EXT_scalar_block_layout can be
enabled at the same time.  But scalar block layout is as permissive
as relax block layout.

Also, scalar block layout does not require padding at the end of a
struct.

Cleanup: The internal getSize method does not need a round-up argument,
so remove it.

FYI:  This is part of the rollout of VK_EXT_scalar_block_layout: https://github.com/KhronosGroup/Vulkan-Docs/issues/854

### DIFF
--- a/source/spirv_validator_options.h
+++ b/source/spirv_validator_options.h
@@ -43,12 +43,14 @@ struct spv_validator_options_t {
         relax_struct_store(false),
         relax_logical_pointer(false),
         relax_block_layout(false),
+        scalar_block_layout(false),
         skip_block_layout(false) {}
 
   validator_universal_limits_t universal_limits_;
   bool relax_struct_store;
   bool relax_logical_pointer;
   bool relax_block_layout;
+  bool scalar_block_layout;
   bool skip_block_layout;
 };
 
--- a/source/val/validate_decorations.cpp
+++ b/source/val/validate_decorations.cpp
@@ -217,23 +217,60 @@ uint32_t getBaseAlignment(uint32_t member_id, bool roundUp,
   return baseAlignment;
 }
 
+// Returns scalar alignment of a type.
+uint32_t getScalarAlignment(uint32_t type_id, ValidationState_t& vstate) {
+  const auto inst = vstate.FindDef(type_id);
+  const auto& words = inst->words();
+  switch (inst->opcode()) {
+    case SpvOpTypeInt:
+    case SpvOpTypeFloat:
+      return words[2] / 8;
+    case SpvOpTypeVector:
+    case SpvOpTypeMatrix: {
+      const auto componentId = words[2];
+      return getScalarAlignment(componentId, vstate);
+    }
+    case SpvOpTypeArray:
+    case SpvOpTypeRuntimeArray: {
+      const auto elementTypeId = words[2];
+      return getScalarAlignment(elementTypeId, vstate);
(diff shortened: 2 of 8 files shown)


dneto0:
I expect the shaderc smoke test to fail because this MR changes the validation message to be more specific about which layout rules are in effect.

This is the one change to the Glslang validation results in its baseResults. (Basically "standard" -> "relaxed")

```
With diff:
@@ -147,5 +147,5 @@
 
 error: SPIRV-Tools Validation Errors
-error: Structure id 50 decorated as BufferBlock for variable in Uniform storage class must follow standard storage buffer layout rules: member 7 at offset 128 overlaps previous member ending at offset 171
+error: Structure id 50 decorated as BufferBlock for variable in Uniform storage class must follow relaxed storage buffer layout rules: member 7 at offset 128 overlaps previous member ending at offset 171
   %tbufName = OpTypeStruct %v4float %int %float %float %float %float %float %float %mat3v4float %mat3v4float %mat3v4float %mat3v4float
 

[  FAILED  ] ToSpirv/HlslCompileTest.FromFile/hlsl_buffer_frag, where GetParam() = 16-byte object <9B-A6 DB-D6 68-55 00-00 85-A5 DB-D6 68-55 00-00> (4 ms)
```

cc: @johnkslang This will be another update to the Glslang base results validation output.

source/val/validate_decorations.cpp
```diff
@@ -216,23 +216,60 @@ uint32_t getBaseAlignment(uint32_t member_id, bool roundUp,
...
+  switch (inst->opcode()) {
+    case SpvOpTypeInt:
+    case SpvOpTypeFloat:
+      return words[2] / 8;
+    case SpvOpTypeVector:
+    case SpvOpTypeMatrix: {
+      const auto componentId = words[2];
+      return getScalarAlignment(componentId, vstate);
+    }
+    case SpvOpTypeArray:
```
alan-baker: Same word as for vectors and matrices, I'd suggest just combining these cases.

source/spirv_validator_options.h
```diff
@@ -43,12 +43,14 @@ struct spv_validator_options_t {
...
         relax_logical_pointer(false),
         relax_block_layout(false),
+        scalar_block_layout(false),
         skip_block_layout(false) {}
 
   validator_universal_limits_t universal_limits_;
   bool relax_struct_store;
   bool relax_logical_pointer;
   bool relax_block_layout;
+  bool scalar_block_layout;
```
alan-baker: What do you think about turning this and relax_block_layout in a Layout enum? The setters could stay as is, but make it so if relaxed block layout won't downgrade from scalar block layout. I think it might make the checking code easier to read since you'd only have to compare against the one enum value instead of combining multiple boolean values.

dneto0: I took a look at doing this.  The trouble is that the current public API allows you to set each of these indivdal bits to _false_.  So if I've set   relax=true, then scalar=true, then scalar=false, then I need extra bits to store the fact that the final result should be relax=true instead of something else.

dneto0:
Rebased and pushed into master as 8e9be303b

PR #2077, opened 2018-11-19
Validate component literals for OpVectorShuffle in WebGPU environment

### LINKED ISSUE #2072: Validate OpVectorShuffle for WebGPU
> OpVectorShuffle may not have a component literal with value 0xFFFFFFFF.

zoddicus:
Fixes #2072

### DIFF
--- a/source/spirv_target_env.cpp
+++ b/source/spirv_target_env.cpp
@@ -220,3 +220,31 @@ bool spvIsOpenCLEnv(spv_target_env env) {
   }
   return false;
 }
+
+bool spvIsWebGPUEnv(spv_target_env env) {
+  switch (env) {
+    case SPV_ENV_UNIVERSAL_1_0:
+    case SPV_ENV_VULKAN_1_0:
+    case SPV_ENV_UNIVERSAL_1_1:
+    case SPV_ENV_OPENGL_4_0:
+    case SPV_ENV_OPENGL_4_1:
+    case SPV_ENV_OPENGL_4_2:
+    case SPV_ENV_OPENGL_4_3:
+    case SPV_ENV_OPENGL_4_5:
+    case SPV_ENV_UNIVERSAL_1_2:
+    case SPV_ENV_UNIVERSAL_1_3:
+    case SPV_ENV_VULKAN_1_1:
+    case SPV_ENV_OPENCL_1_2:
+    case SPV_ENV_OPENCL_EMBEDDED_1_2:
+    case SPV_ENV_OPENCL_2_0:
+    case SPV_ENV_OPENCL_EMBEDDED_2_0:
+    case SPV_ENV_OPENCL_EMBEDDED_2_1:
+    case SPV_ENV_OPENCL_EMBEDDED_2_2:
+    case SPV_ENV_OPENCL_2_1:
+    case SPV_ENV_OPENCL_2_2:
+      return false;
+    case SPV_ENV_WEBGPU_0:
+      return true;
+  }
+  return false;
+}
--- a/source/spirv_target_env.h
+++ b/source/spirv_target_env.h
@@ -27,6 +27,9 @@ bool spvIsVulkanEnv(spv_target_env env);
 // Returns true if |env| is an OPENCL environment, false otherwise.
 bool spvIsOpenCLEnv(spv_target_env env);
 
+// Returns true if |env| is an WEBGPU environment, false otherwise.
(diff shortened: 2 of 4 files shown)


source/spirv_target_env.cpp
```diff
@@ -220,3 +220,31 @@ bool spvIsOpenCLEnv(spv_target_env env) {
   }
   return false;
 }
+
+bool spvIsWebGPUEnv(spv_target_env env) {
+  switch (env) {
+    case SPV_ENV_UNIVERSAL_1_0:
```
dneto0: I guess it's deliberate that there is no "default" case.  That requires this code to be updated whenever a new environment is added.  Ok. I guess that makes sense.  (Otherwise I would make a catchall default clause.  But it's safer the way you have it here.)

dj2: spv_target_env isn't an enum class so you don't get covering warnings I believe. Although, in this case the default case would just be the same as the non-default return of false so I'm not sure if it matters.

source/spirv_target_env.cpp
```diff
@@ -220,3 +220,31 @@ bool spvIsOpenCLEnv(spv_target_env env) {
   }
   return false;
 }
+
+bool spvIsWebGPUEnv(spv_target_env env) {
+  switch (env) {
```
dj2: Couldn't we just simplify this to be:

return env == SPV_ENV_WEBGPU_0; ?

we may need to extend it later if we get more then just _0, but that seems like the simplest thing to do at this point?

source/spirv_target_env.h
```diff
@@ -27,6 +27,9 @@ bool spvIsVulkanEnv(spv_target_env env);
 // Returns true if |env| is an OPENCL environment, false otherwise.
 bool spvIsOpenCLEnv(spv_target_env env);
 
+// Returns true if |env| is an WEBGPU environment, false otherwise.
```
dj2: WebGPU ? I'm not sure the right capitalization here, what does the spec say?

source/val/validate_composites.cpp
```diff
@@ -464,17 +465,25 @@ spv_result_t ValidateVectorShuffle(ValidationState_t& _,
...
     auto literal = inst->GetOperandAs<uint32_t>(i);
     if (literal != 0xFFFFFFFF && literal >= N) {
       return _.diag(SPV_ERROR_INVALID_ID, inst)
              << "Component index " << literal << " is out of bounds for "
              << "combined (Vector1 + Vector2) size of " << N << ".";
     }
+
+    if (is_webgpu_env && literal == 0xFFFFFFFF) {
+      return _.diag(SPV_ERROR_INVALID_ID, inst)
+             << "Component literal @ operand " << i - firstLiteralIndex
```
dj2: What is that symbol after literal?

dj2: I think it's an @, but the font is terrible, heh. I'd just write at instead.

source/spirv_target_env.cpp
zoddicus: For this I was following the pattern established by the other spvIs*Env functions in the file. I agree with the assessment that this isn't throwing any warnings due to it being a C-style enum, so there isn't an enforcement that all possible values are tested. I assume the not default, but instead a final return statement is likely due to some combination of compiler and flags not complaining about no return statement when just using default.

I think the advantage of not just iterating the positive cases, is that it makes it easier for a dev to scan for the enum they care about. A find query will hit the enum regardless of it being a positive or negative case, as opposed to having to figure out it is caught by the final return.

source/spirv_target_env.cpp
zoddicus: I effectively responded to this on another comment :-/

From the other comment:
I think the advantage of not just iterating the positive cases, is that it makes it easier for a dev to scan for the enum they care about. A find query will hit the enum regardless of it being a positive or negative case, as opposed to having to figure out it is caught by the final return.

source/spirv_target_env.h
zoddicus: The other functions of this type in this header are documented using full caps. I think the intent is to indicate that this applies to environment enums that have the string WEBGPU (or VULKAN or OPENCL) in them.

source/val/validate_composites.cpp
zoddicus: Done

zoddicus:
Filed https://github.com/KhronosGroup/SPIRV-Tools/issues/2079 about the Appveyor check being flakey.

PR #2078, opened 2018-11-19
Catch invalid input type to OpConvertUToPtr

alan-baker:
Fixes https://crbug.com/906426

* Added a test


dj2:
Can you update the description to say what the issue was you're fixing?

Issue #2079, opened 2018-11-19
continuous-integration/appveyor/pr check is flakey


zoddicus:
This check is failing occasionally  due to a timeout in spirv-tools-test_val_limits.

An example failed run, https://ci.appveyor.com/project/Khronoswebmaster/spirv-tools/builds/20411836

dj2:
@s-perron started looking into this last week and identified the issue as a linear walk that the test executes ... a lot.

s-perron:
This is the loop that takes most of the time: https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/val/validate_cfg.cpp#L554.

This loops looks like it will be repeating a lot of work.  If a BB is contained in `n` nested constructs, then it predecessors will be checked `n` time.  It only needs to be checked for the inner-most containing construct.

PR #2081, opened 2018-11-19
Remove long running cfg limit test.

s-perron:
We had a test that checks that a spirv binary where the depth of the
deepest nested construct is just below the default limit.  This test
would take a long time causing a timeout in some builds.  We are
questioning the value of the test since we already have a test that can
tell us if we are within a custom limit.  We will remove the test.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/2079.


dneto0:
Rebased and pushed into master as 101c11

PR #2082, opened 2018-11-19
Add tests for matrix type data rule validation

### LINKED ISSUE #2065: Validate OpTypeMatrix for WebGPU
>Matrix types must have one of the following number of vectors:
> - 2
> - 3
> - 4

### LINKED ISSUE #2080: Validate Column Count for OpTypeMatrix
From SPIR-V spec, section 3.32.6 on OpTypeMatrix:
> Column Count is the number of columns in the new matrix type. It must be at least 2.

This is currently not checked by the validator from what I can tell.

zoddicus:
This covers the data rules for matrix types specified in the SPIR-V
spec, section 2.16.1, and the WebGPU SPIR-V Execution Environment
spec.

Fixes #2065 
Fixes #2080

### DIFF
--- a/test/val/val_id_test.cpp
+++ b/test/val/val_id_test.cpp
@@ -579,14 +579,44 @@ TEST_F(ValidateIdWithMessage, OpTypeMatrixGood) {
   CompileSuccessfully(spirv.c_str());
   EXPECT_EQ(SPV_SUCCESS, ValidateInstructions());
 }
-TEST_F(ValidateIdWithMessage, OpTypeMatrixColumnTypeBad) {
+
+TEST_F(ValidateIdWithMessage, OpTypeMatrixColumnTypeNonFloatBad) {
   std::string spirv = kGLSL450MemoryModel + R"(
-%1 = OpTypeInt 32 0
-%2 = OpTypeMatrix %1 3)";
+%1 = OpTypeInt 16 0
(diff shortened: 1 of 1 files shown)


test/val/val_id_test.cpp
```diff
@@ -579,14 +579,44 @@ TEST_F(ValidateIdWithMessage, OpTypeMatrixGood) {
...
   std::string spirv = kGLSL450MemoryModel + R"(
-%1 = OpTypeInt 32 0
-%2 = OpTypeMatrix %1 3)";
+%1 = OpTypeInt 16 0
+%2 = OpTypeVector %1 2
+%3 = OpTypeMatrix %2 2)";
   CompileSuccessfully(spirv.c_str());
-  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
-  EXPECT_THAT(getDiagnosticString(),
-              HasSubstr("Columns in a matrix must be of type vector."));
```
dj2: This error doesn't line up with the new requirement that the type be a float? Should the new error be vector of floating-point instead of just floating-point?

zoddicus: You are correct, I accidentally dropped a test for the column type being a vector. I will fix and upload a new patch.

zoddicus: Done

PR #2085, opened 2018-11-20
Update build instructions: add effcee, re2

hevrard:
Both effcee and re2 are now required dependencies, add their checkout in build instructions.


dj2:
Thank you. Looks like we got one of them but the build list is in there twice.

PR #2088, opened 2018-11-20
Xfb support

alan-baker:
Based on #2011, but fix up the tests.

@brett19, this will replace your PR.


dj2:
Can you update the description when you commit so folks know what this does? (Xfb support doesn't really mean much to me)

alan-baker:
Shaderc failures look like glslang will need an update of expected validation errors.

brett19:
Thanks @alan-baker!  Glad to see it being covered now.

PR #2089, opened 2018-11-20
Validate that debugging instructions are not present for WebGPU

### LINKED ISSUE #2063: Validate that all debug instructions have been stripped for WebGPU
> All debug instructions must be stripped.

From SPIR-V [spec]():
>### 2.15. Debug Information
>Debug information is supplied with:
> - Source-code text through OpString, OpSource, and OpSourceContinued.
> - Object names through OpName and OpMemberName.
> - Line numbers through OpLine.
>
>A module will not lose any semantics when all such instructions are removed.

zoddicus:
For WebGPU execution environments, check that all of the debug
instructions have already been stripped before validation.

Fixes #2063

### DIFF
--- a/source/val/validate_debug.cpp
+++ b/source/val/validate_debug.cpp
@@ -14,6 +14,8 @@
 
 #include "source/val/validate.h"
 
+#include "source/opcode.h"
+#include "source/spirv_target_env.h"
 #include "source/val/instruction.h"
 #include "source/val/validation_state.h"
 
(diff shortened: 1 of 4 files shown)


source/val/validate_debug.cpp
```diff
@@ -51,9 +53,33 @@ spv_result_t ValidateLine(ValidationState_t& _, const Instruction* inst) {
   return SPV_SUCCESS;
 }
 
+bool IsDebugInstruction(const Instruction* inst) {
```
alan-baker: This looks useful in general. I would suggest moving it into opcode.cpp/h and naming spvOpcodeIsDebugInst

source/val/validate_debug.cpp
```diff
@@ -51,9 +53,33 @@ spv_result_t ValidateLine(ValidationState_t& _, const Instruction* inst) {
...
+  switch (inst->opcode()) {
+    case SpvOpName:
+    case SpvOpMemberName:
+    case SpvOpSource:
+    case SpvOpSourceContinued:
+    case SpvOpSourceExtension:
+    case SpvOpString:
+    case SpvOpLine:
+    case SpvOpNoLine:
```
dj2: @alan-baker do we need to worry about things like OpDecorateStringGOOGLE here?

alan-baker: Those are decorations not debug instructions. So it would depend on what the env spec says for supported decorations.

zoddicus:
The other logical place I can think of putting this check would be up the parse around the callside of RegisterDebugInstruction in ProcessInstruction. I am not particularly wedded with one location over another, but you did point out offline that putting this check where it currently is, localizes it a bit better to the other debug validation related code.

source/val/validate_debug.cpp
zoddicus: Done, dropped the Inst suffix to match the style of other functions in the header.

source/val/validate_debug.cpp
zoddicus: The spec is defining a whitelist of decorations, so likely they will be handled by the default if not on this list, then not allowed. If/when we need to deal with these decorations they will need to be handled by something different then this opcode check.

PR #2092, opened 2018-11-21
Make sure that initialized variable have correct storage class

### LINKED ISSUE #2071: Validate Initializers for WebGPU
> Variable declarations that include initializers must be for one of the following storage classes:
> - Output
> - Private
> - Function

zoddicus:
For WebGPU environments variables must have the storage class Output,
Private, or Function if they have an initializer.

Fixes #2071

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -389,20 +389,32 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
     }
   }
 
-  // WebGPU: Check that if contains initializer, then storage class is Output,
-  // Private, or Function.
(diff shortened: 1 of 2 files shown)


source/val/validate_memory.cpp
```diff
@@ -389,6 +389,23 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
     }
   }
 
+  // WebGPU: Check that if contains initializer, then storage class is Output,
+  // Private, or Function.
+  if (spvIsWebGPUEnv(_.context()->target_env) && inst->operands().size() > 3) {
```
alan-baker: This the same restriction as Vulkan, can you enable (and modify the message) for that environment too please?

zoddicus:
smoketest failure unrelated to PR.

PR #2096, opened 2018-11-22
Validate gl_InstanceID for raytracing

alelenv:
Add precise check for allowing use of gl_InstanceID for specific vulkan raytracing stages .

### DIFF
--- a/source/val/validate_builtins.cpp
+++ b/source/val/validate_builtins.cpp
@@ -207,6 +207,11 @@ class BuiltInsValidator {
       const Instruction& referenced_inst,
       const Instruction& referenced_from_inst);
 
+  spv_result_t ValidateInstanceIdAtReference(
+      const Decoration& decoration, const Instruction& built_in_inst,
+      const Instruction& referenced_inst,
+      const Instruction& referenced_from_inst);
+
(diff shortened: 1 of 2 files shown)


alelenv:
@dneto0 @alan-baker : Can you please review
Thanks

source/val/validate_builtins.cpp
```diff
@@ -2098,6 +2103,43 @@ spv_result_t BuiltInsValidator::ValidateVertexIdOrInstanceIdAtDefinition(
...
 
+  if (label == SpvBuiltInInstanceId) {
+      return ValidateInstanceIdAtReference(decoration, inst, inst, inst);
+  }
+  return SPV_SUCCESS;
+}
+
+spv_result_t BuiltInsValidator::ValidateInstanceIdAtReference(
+    const Decoration& decoration, const Instruction& built_in_inst,
+    const Instruction& referenced_inst,
```
alan-baker: This unused parameter is causing the build failures.

alelenv:
Reverting to old format for break in switch case since CI formatter complained
I had ran through clang format with format file at root and thought this was some latent issue but seems like its not

PR #2097, opened 2018-11-22
Fix the missing pch files in spirv-reduce

gongminmin:
Trying to generate VS2017 project files, and cmake complains about can't find pch_source_reduce.cpp and .h. Somehow CI systems don't catch this. Maybe because spvtools_pch is only enabled in VS.

This change add these 2 files to fix this problem.


dneto0:
It's fine to put that copyright, I think.  I did that for a long time.

PR #2099, opened 2018-11-22
New spirv-reduce reduction pass: operand to dominating id.

afd:
If an instruction generating id x dominates and instruction generating y, and if the instructions have the same associated result type, this pass will try replacing uses of y with uses of x.  This is inspired by the idea in a more human-oriented language of replacing an expression with one of its sub-expressions, or with some available identifier of the right type.


source/reduce/operand_to_dominating_id_reduction_pass.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+void OperandToDominatingIdReductionPass::GetOpportunitiesForDominatingInst(
+    std::vector<std::unique_ptr<ReductionOpportunity>>* opportunities,
+    opt::Instruction* candidate_dominator, opt::Function* function,
+    opt::IRContext* context) const {
+  assert(candidate_dominator->HasResultId());
+  assert(candidate_dominator->type_id());
+  auto dominator_analysis = context->GetDominatorAnalysis(function);
+  for (auto& block : *function) {
```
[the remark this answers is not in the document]
afd: Thanks!  I've implemented this optimization in my latest commit.

afd:
Thanks for the idea for another reduction pass; I'll file issues for such ideas in due course.

source/reduce/operand_to_dominating_id_reduction_pass.cpp
```diff
@@ -0,0 +1,118 @@
...
+      // cannot be any of the desired reduction opportunities in this block.
+      continue;
+    }
+    for (auto& inst : *block) {
+      // We iterate through the operands using an explicit index (rather
+      // than using a lambda) so that we use said index in the construction
+      // of a ChangeOperandReductionOpportunity
+      for (uint32_t index = 0; index < inst.NumOperands(); index++) {
+        const auto& operand = inst.GetOperand(index);
+        if (spvIsIdType(operand.type)) {
```
alan-baker: This is starting to look like a useful helper function for you. I would suggest spvIsInIdType(). Please update WhileEachInId in addition to the other reduction passes.

afd: Doing a helper sounds good.  What's the significance of "In" in the name, though, in terms of distinguishing the function from the existing spvIsIdType()?

alan-baker: The ids it blacklists are related to the result of an instruction so this function returns true for any input operand id.

afd: OK, makes sense!

PR #2103, opened 2018-11-23
Add cmake option to turn off SPIRV_TIMER_ENABLED

baldurk:
Currently when building on linux, `SPIRV_TIMER_ENABLED` defaults to on which takes a dependency on `clock_gettime`. The `clock_*` functions were added to glibc in 2.17 though and prior to that you had to link `-lrt` to get it, so on old distributions the build fails due to an unresolved symbol. Since this is a rarely used feature and optional, I've added a cmake option to allow it (defaults to on, so no change).

I'm not sure if you want to link against `-lrt` to fix the older distro builds, as I don't know what that would do on new systems where libc provides the functions. It might resolve the libc symbol first and not declare a dependency on `librt`, or it might link anyway.

For context; I'm building a static linked version of the tools for distribution and so I'm building on ubuntu 12.04 as a lowest common denominator of compatibility. For that reason I'd want to avoid the extra library dependency anyway even if it was fixed properly.


alan-baker:
shaderc failure is known issue.

PR #2110, opened 2018-11-23
Prevent diagnostic memory leak

alan-baker:
Fixes https://crbug.com/906669

* Ensure allocated diagnostics are deleted before losing a reference to
them in the binary parsing

### DIFF
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -767,6 +767,13 @@ spv_result_t spvBinaryParse(const spv_const_context context, void* user_data,
                             spv_diagnostic* diagnostic) {
   spv_context_t hijack_context = *context;
   if (diagnostic) {
+    // Don't leak memory.
+    if (*diagnostic) {
+      if ((*diagnostic)->error) {
+        delete (*diagnostic)->error;
+      }
+      delete *diagnostic;
+    }
     *diagnostic = nullptr;
     spvtools::UseDiagnosticAsMessageConsumer(&hijack_context, diagnostic);
   }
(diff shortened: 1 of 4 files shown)


source/binary.cpp
```diff
@@ -767,6 +767,13 @@ spv_result_t spvBinaryParse(const spv_const_context context, void* user_data,
                             spv_diagnostic* diagnostic) {
   spv_context_t hijack_context = *context;
   if (diagnostic) {
+    // Don't leak memory.
```
dneto0: This arguably is a bug in the caller.
Please update the method comment in libspirv.h to describe this new behaviour.

dneto0: How do you know the pointer you're trying to delete was actually created with (scalar) new?    You don't.  I think this is the wrong fix.  I think the only safe thing is to update the contract so that the caller doesn't do this.

s-perron:
Do you have an details on which call to spvBinaryParse is causing the problem?  I tried looking for it, but I cannot find something that stands out.

alan-baker:
I think it was the call in the validation that used ProcessExtensions as the callback.

source/binary.cpp
alan-baker: I reverted this fix and went about it a different way.

alan-baker:
There are two parses in the validation flow that where we wish to ignore parse errors. One in the validation state constructor to count the number of instructions and one in the main validation driver where we process extensions. Before invoking either of those parses I override the message consumer to avoid polluting the state of the intended consumer. This felt more appropriate than overriding the consumer in spvBinaryParse, but I leave that for review.

PR #2117, opened 2018-11-26
Add validation for OpArrayLength.

s-perron:
The validation rules for OpArrayLength are not checked by the validator.
This with add them.

Fixes https://crbug.com/907451.

### DIFF
--- a/test/val/val_memory_test.cpp
+++ b/test/val/val_memory_test.cpp
@@ -543,6 +543,300 @@ OpFunctionEnd
           "%5 = OpVariable %_ptr_Input_float Input %float_1\n"));
 }
 
+TEST_F(ValidateMemory, ArrayLenCorrectResultType) {
+  std::string spirv = R"(
+               OpCapability Shader
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint Fragment %1 "main"
+               OpExecutionMode %1 OriginUpperLeft
+       %void = OpTypeVoid
+          %3 = OpTypeFunction %void
+      %float = OpTypeFloat 32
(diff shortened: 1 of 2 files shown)


test/val/val_memory_test.cpp
```diff
@@ -543,6 +543,300 @@ OpFunctionEnd
...
+       %void = OpTypeVoid
+          %3 = OpTypeFunction %void
+      %float = OpTypeFloat 32
+     %uint = OpTypeInt 32 0
+%_runtimearr_float = OpTypeRuntimeArray %float
+  %_struct_7 = OpTypeStruct %_runtimearr_float
+%_ptr_Function__struct_7 = OpTypePointer Function %_struct_7
+          %1 = OpFunction %void None %3
+          %9 = OpLabel
+         %10 = OpVariable %_ptr_Function__struct_7 Function
```
alan-baker: This is invalid in Vulkan, though we don't have a check for it yet. Suggest not using the vulkan environment. Same for subsequent tests.

From the Vulkan spec:
> OpTypeRuntimeArray must only be used for:
> * the last member of an OpTypeStruct that is in the StorageBuffer storage class decorated as Block, or that is in the Uniform storage class decorated as BufferBlock.
> * If the RuntimeDescriptorArrayEXT capability is supported, an array of variables with storage class Uniform, StorageBuffer, or UniformConstant, or for the outermost dimension of an array of arrays of such variables.

s-perron: Done.

alan-baker:
Did you file a glslang issue for the test failures?

PR #2119, opened 2018-11-26
Ensure that imported extended instructions for WebGPU are only "GLSL.std.450"

### LINKED ISSUE #2059: Validate OpExtInstImport for WebGPU

> All extended instruction sets declared by OpExtInstImport must be one of the following:
>"GLSL.std.450"

zoddicus:
Fixes #2059

### DIFF
--- a/source/val/validate_instruction.cpp
+++ b/source/val/validate_instruction.cpp
@@ -449,6 +449,17 @@ spv_result_t InstructionPass(ValidationState_t& _, const Instruction* inst) {
     CheckIfKnownExtension(_, inst);
   } else if (opcode == SpvOpCapability) {
     _.RegisterCapability(inst->GetOperandAs<SpvCapability>(0));
+  } else if (opcode == SpvOpExtInstImport) {
+    if (spvIsWebGPUEnv(_.context()->target_env)) {
+      const auto name_id = 1;
+      const std::string name(reinterpret_cast<const char*>(
+          inst->words().data() + inst->operands()[name_id].offset));
+      if (name != "GLSL.std.450") {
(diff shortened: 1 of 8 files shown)


source/val/validate_instruction.cpp
```diff
@@ -449,6 +449,17 @@ spv_result_t InstructionPass(ValidationState_t& _, const Instruction* inst) {
     CheckIfKnownExtension(_, inst);
   } else if (opcode == SpvOpCapability) {
     _.RegisterCapability(inst->GetOperandAs<SpvCapability>(0));
+  } else if (opcode == SpvOpExtInstImport) {
```
alan-baker: I would prefer this get into a new file, validate_extensions.cpp, and get invoked as ExtensionsPass between annotation and mode checks.

zoddicus: There is a validate_ext_inst.cpp that implements ExtInstPass. Should I add this code there instead of adding a new file and pass function?

zoddicus: Though on second look, it looks more like it is dealing with the specific calls into extension as opposed to the import.

zoddicus: Done

zoddicus:
Moved things to be name ExtInstImport based instead of Extensions based, since that is what is specifically being tested.

alan-baker:
I should have clarified how I envision the structure. Where possible I want instruction checks to be match specification sections. OpExtInstImport is within the Extensions sections of instructions so validate_extensions.cpp. We can leave a refactor of actual extended instructions for later.

You forgot to add the new file to the Android.mk

zoddicus:
Created issue #2123 to track the refactor.

PR #2129, opened 2018-11-28
Fixup BUILD.gn file

dj2:


dj2:
TBR'ing to fix the chrome roll. Has no impact on other code.

PR #2131, opened 2018-11-28
Merge two implementations of ValidateExecutionScope

zoddicus:


Android.mk
```diff
@@ -66,6 +66,7 @@ SPVTOOLS_SRC_FILES := \
 		source/val/validate_logicals.cpp \
 		source/val/validate_non_uniform.cpp \
 		source/val/validate_primitives.cpp \
+		source/val/validate_scopes.cpp \
```
dj2: BUILD.gn?

zoddicus: Done.

PR #2132, opened 2018-11-29
Merging two ValidateMemoryScope implementations

### LINKED ISSUE #2125: Merge multiple scope validator functions
There appears to be multiple implementations of ValidateMemoryScope, and ValidateExecutionScope local to specific validator files. These should be unified into a single canonical implementation for each scope type.

zoddicus:
Fixes #2125

### DIFF
--- a/source/val/validate_scopes.cpp
+++ b/source/val/validate_scopes.cpp
@@ -100,5 +100,56 @@ spv_result_t ValidateExecutionScope(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+spv_result_t ValidateMemoryScope(ValidationState_t& _, const Instruction* inst,
+                                 uint32_t scope) {
+  const SpvOp opcode = inst->opcode();
+  bool is_int32 = false, is_const_int32 = false;
+  uint32_t value = 0;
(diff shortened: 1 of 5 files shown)


source/val/validate_scopes.cpp
```diff
@@ -100,5 +100,56 @@ spv_result_t ValidateExecutionScope(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+spv_result_t ValidateMemoryScope(ValidationState_t& _, const Instruction* inst,
+                                 uint32_t scope) {
+  const SpvOp opcode = inst->opcode();
+  bool is_int32 = false, is_const_int32 = false;
+  uint32_t value = 0;
+  std::tie(is_int32, is_const_int32, value) = _.EvalInt32IfConst(scope);
```
alan-baker: EvalInt32IfConst returns that the value is constant for OpSpecConstant* opcodes. In that case, the value it returns is not reliable. In the worst case we reinterpret an opcode from an OpSpecConstantOp as a value. In other cases we interpret the default value as the real value. Both of these are wrong. I've filed #2133 to address this, since your PR doesn't make it any worse.

Issue #2005, opened 2018-10-23
Validate Vulkan PushConstant variables


alan-baker:
Variables in PushConstant storage class must:
* be a struct or array of structs
* decorated with `Block`
* be laid out explicitly (`Offset`, `ArrayStride` and `MatrixStride` decorations)

alan-baker:
Fixed via #2140

PR #2140, opened 2018-11-30
Validate PushConstants annotation and type

ale64bit:
Check that PushConstant OpVariable is identified by Block decoration and type is struct or array of structs.

### DIFF
--- a/source/val/validate_decorations.cpp
+++ b/source/val/validate_decorations.cpp
@@ -796,12 +796,15 @@ spv_result_t CheckDecorationsOfBuffers(ValidationState_t& vstate) {
         ComputeMemberConstraintsForStruct(&constraints, id, LayoutConstraints(),
                                           vstate);
 
-        if (push_constant && !hasDecoration(id, SpvDecorationBlock, vstate)) {
-          return vstate.diag(SPV_ERROR_INVALID_ID, vstate.FindDef(id))
-                 << "PushConstant id '" << id
-                 << "' is missing Block decoration.\n"
-                 << "From Vulkan spec, section 14.5.1:\n"
-                 << "Such variables must be identified with a Block decoration";
+        // Vulkan 14.5.1: Check Block annotation for PushConstant variables.
+        if (spvIsVulkanEnv(vstate.context()->target_env)) {
+          if (push_constant && !hasDecoration(id, SpvDecorationBlock, vstate)) {
+            return vstate.diag(SPV_ERROR_INVALID_ID, vstate.FindDef(id))
+                << "PushConstant id '" << id
+                << "' is missing Block decoration.\n"
+                << "From Vulkan spec, section 14.5.1:\n"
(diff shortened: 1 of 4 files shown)


alan-baker:
Addresses #2005

source/val/validate_decorations.cpp
```diff
@@ -795,6 +795,15 @@ spv_result_t CheckDecorationsOfBuffers(ValidationState_t& vstate) {
         MemberConstraints constraints;
         ComputeMemberConstraintsForStruct(&constraints, id, LayoutConstraints(),
                                           vstate);
+
+        if (push_constant && !hasDecoration(id, SpvDecorationBlock, vstate)) {
```
alan-baker: This check should also be gated by `spvIsVulkanEnv(vstate.context()->target_env)`. It does not apply generally to SPIR-V.

source/val/validate_memory.cpp
```diff
@@ -405,6 +405,17 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
                << "must be typed as OpTypeStruct, or an array of this type";
       }
     }
+
+    if (storage_class == SpvStorageClassPushConstant) {
```
alan-baker: Same here for the environment.

test/val/val_decoration_test.cpp
```diff
@@ -2219,6 +2219,38 @@ TEST_F(ValidateDecorations,
           "rules: member 1 at offset 4 is not aligned to 16"));
 }
 
+TEST_F(ValidateDecorations, PushConstantMissingBlockVulkanBad) {
```
alan-baker: This test should succeed with a universal environment. Please modify this test to use the Vulkan environment and add a new test for the universal environment.

test/val/val_memory_test.cpp
```diff
@@ -838,6 +838,34 @@ TEST_F(ValidateMemory, ArrayLenIndexNotPointerToStruct) {
           "The Struture's type in OpArrayLength <id> '12' must be a pointer to "
           "an OpTypeStruct.\n  %12 = OpArrayLength %uint %11 0\n"));
 }
+
+TEST_F(ValidateMemory, VulkanPushConstantNotStructBad) {
```
alan-baker: Similar treatment here. Add a passing version in the universal environment.

source/val/validate_memory.cpp
ale64bit: Isn't this already gated above?
https://github.com/KhronosGroup/SPIRV-Tools/blob/70fdd2e3a0bc71918a1d34a31584077279557890/source/val/validate_memory.cpp#L377

I added a comment and reorganized according to the spec, if that helps.

ale64bit:
PTAL

alan-baker:
Filed https://github.com/KhronosGroup/glslang/issues/1609 for the invalid glslang test this check exposes.

PR #2146, opened 2018-11-30
Validate uses of ids defined in unreachable blocks.

### LINKED ISSUE #2143: Validator does not complain about id used indifferent function than its definition.
Somehow OpUnreachable is causing the validator to accept the following code:

```
              OpCapability Shader
               OpMemoryModel Logical GLSL450
       %void = OpTypeVoid
          %5 = OpTypeFunction %void
      %float = OpTypeFloat 32
         %12 = OpTypeFunction %float
          %1 = OpFunction %void None %5
         %14 = OpLabel
         %18 = OpFunctionCall %float %19
               OpUnreachable
               OpFunctionEnd
         %19 = OpFunction %float None %12
         %21 = OpLabel
               OpReturnValue %18
               OpFunctionEnd
```

If we change the OpUnreachable to OpReturn, then the validator will complain that the definition of %18 does not dominate all of its uses. 

Fixes https://crbug.com/906329

s-perron:
For some reason we do not make sure the uses of ids that are defined
in unreachable blocks are dominated by their def.  This is causing
invalid code to pass the validator.

Fixes #2143

### DIFF
--- a/source/val/validate_id.cpp
+++ b/source/val/validate_id.cpp
@@ -65,7 +65,6 @@ spv_result_t CheckIdDefinitionDominateUse(ValidationState_t& _) {
     if (inst.id() == 0) continue;
     if (const Function* func = inst.function()) {
       if (const BasicBlock* block = inst.block()) {
-        if (!block->reachable()) continue;
         // If the Id is defined within a block then make sure all references to
         // that Id appear in a blocks that are dominated by the defining block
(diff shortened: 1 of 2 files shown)


s-perron:
I don't know why it was it was skipped, but I added a bunch of tests that I thought this change could affect.  They all work as they should.  Let me know if there are most thing I should try.

PR #2151, opened 2018-12-01
Check binding annotations in resource variables

### LINKED ISSUE #2007: Vulkan Resource variables require bindings
In Vulkan, all variables in Uniform, UniformConstant and StorageBuffer storage classes require `DescriptorSet` and `Binding` decorations.

ale64bit:
Fixes #2007


alan-baker:
The smoketest has a ton of failures. They look like cases of compilers not annotating descriptor set or binding as appropriate.

PR #2152, opened 2018-12-03
Fixup BUILD.gn from addtion of upgrade_memory_model

dj2:


dj2:
TBR=@alan-baker to unblock the Chrome roll.

alan-baker:
Sorry about that. I always forget the BUILD.gn.

PR #2155, opened 2018-12-03
Validate ArrayLength with a type for in place of the struct.

s-perron:
The current checks for OpArrayLength assumes that there is a type it in
the instruction that defines what is suppose to be pointer to the
structure.  However, if there is an id of a type in that position there
is not.  This can lead to a segmentation fault.

We just need to check for that case.

Fixes https://crbug.com/910013.


s-perron:
Fixed by d80259d35e667fda291adf32c5fd29efda976999.

PR #2156, opened 2018-12-03
Validate OpForwardPointer

s-perron:
The validator does not have a a check that OpForwardPointer is giving
a forward reference to a pointer type.  We add that check.

https://crbug.com/910852

### DIFF
--- a/source/val/validate_type.cpp
+++ b/source/val/validate_type.cpp
@@ -296,10 +296,25 @@ spv_result_t ValidateTypeFunction(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+spv_result_t ValidateTypeForwardPointer(ValidationState_t& _,
+                                        const Instruction* inst) {
+  const auto pointer_type_id = inst->GetOperandAs<uint32_t>(0);
+  const auto pointer_type_inst = _.FindDef(pointer_type_id);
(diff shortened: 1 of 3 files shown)


source/val/validate_type.cpp
```diff
@@ -296,10 +291,25 @@ spv_result_t ValidateTypeFunction(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+spv_result_t ValidateTypeForwardPointer(ValidationState_t& _,
```
alan-baker: I would also suggest validating that the storage classes match if the first check passes.

source/val/validate_type.cpp
```diff
@@ -296,10 +291,25 @@ spv_result_t ValidateTypeFunction(ValidationState_t& _,
...
+                                        const Instruction* inst) {
+  const auto pointer_type_id = inst->GetOperandAs<uint32_t>(0);
+  const auto pointer_type_inst = _.FindDef(pointer_type_id);
+  if (pointer_type_inst->opcode() != SpvOpTypePointer) {
+    return _.diag(SPV_ERROR_INVALID_ID, inst)
+           << "Pointer type in OpForwardPointer is not a pointer type.";
```
alan-baker: OpTypeForwardPointer

PR #2157, opened 2018-12-03
Remove undefined behaviour when folding shifts.

s-perron:
We currently simulate all shift operations when the two operand are
constants.  The problem is that if the shift amount is larger than
32, the result is undefined.

I'm changing the folder to return 0 if the shift value is too high.
That way, we will have defined behaviour.

https://crbug.com/910937.


s-perron:
@dnovillo Have I addressed all of your issues?

dnovillo:
> @dnovillo Have I addressed all of your issues?

Yes.  Thanks.

PR #2161, opened 2018-12-03
Check for recursion in Vulkan and WebGPU entry points

### LINKED ISSUE #2061: Validate OpEntryPoint for WebGPU
>There must be at least one OpEntryPoint in a module, and each one present must satisfy:
>
> - The OpFunction declared as the entry point must have no OpFunctionParameter and its Return Type must be OpTypeVoid.
> - The complete call graph rooted by the entry point must be present in the module (the Linkage capability is not valid, so there are no partially linked modules), and not contain any cycles (no recursion, even statically).
> - The Name operand must be a string of length [TBD, TBD] and cannot be used by any other OpEntryPoint in the same module. All entry-point names in a module must be unique module these rules. TBD do we want case sensitive?
>
> The execution model declared by each OpEntryPoint must be one of the following:
> - Vertex
> - Fragment
> - GLCompute

### LINKED ISSUE #2160: Validator should fail Vulkan code with entry points that have recursion in their call tree
From the Vulkan spec:
> Recursion: The static function-call graph for an entry point must not contain cycles.

The is a similar rule for WebGPU.

zoddicus:
Fixes #2061
Fixes #2160

### DIFF
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -14,9 +14,7 @@
 
 #include "source/val/validation_state.h"
 
-#include <algorithm>
 #include <cassert>
-#include <deque>
 #include <stack>
 #include <utility>
 
--- a/source/val/validation_state.cpp
+++ b/source/val/validation_state.cpp
@@ -14,7 +14,9 @@
 
 #include "source/val/validation_state.h"
 
+#include <algorithm>
 #include <cassert>
+#include <deque>
 #include <stack>
 #include <utility>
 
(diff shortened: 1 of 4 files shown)


zoddicus:
The smoketest failures I think are not related to my patch.

source/val/validation_state.cpp
```diff
@@ -926,6 +928,39 @@ void ValidationState_t::ComputeFunctionToEntryPointMapping() {
...
+      if (called_func_id == func.id()) {
+        for (const uint32_t entry_point :
+             function_to_entry_points_[called_func_id])
+          recursive_entry_points_.insert(entry_point);
+        break;
+      }
+
+      const Function* called_func = function(called_func_id);
+      if (called_func) {
+        // Other checks should error out on this invalid SPIR-V.
```
alan-baker: Is this comment meant to be above when `called_func_id == func.id()`?

zoddicus:
I don't think my PR has added anymore failures to the smoketest.

PR #2163, opened 2018-12-04
Validate that there is at most one push constant block

### LINKED ISSUE #2006: Vulkan requires there is at most one PushConstant interface per entry point
> There must be no more than one push constant block statically used per shader entry point.

ale64bit:
Fixes #2006

### DIFF
--- a/test/val/val_decoration_test.cpp
+++ b/test/val/val_decoration_test.cpp
@@ -2289,16 +2289,19 @@ TEST_F(ValidateDecorations, MultiplePushConstantInterfacesGood) {
             OpMemoryModel Logical GLSL450
             OpEntryPoint Fragment %1 "main"
             OpExecutionMode %1 OriginUpperLeft
-             OpDecorate %struct Block
+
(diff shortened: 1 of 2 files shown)


ale64bit:
PTAL

test/val/val_decoration_test.cpp
```diff
@@ -2283,6 +2283,69 @@ TEST_F(ValidateDecorations, VulkanPushConstantMissingBlockBad) {
...
+            OpReturn
+            OpFunctionEnd
+)";
+
+  CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_SUCCESS, ValidateAndRetrieveValidationState())
+      << getDiagnosticString();
+}
+
+TEST_F(ValidateDecorations, VulkanMultiplePushConstantInterfacesBad) {
```
alan-baker: This test should be valid because `%pc1` and `%pc2` are not used in `%1`.

ale64bit:
PTAL

PR #2172, opened 2018-12-05
Fix invalid OpPhi generated by merge-return.

s-perron:
When we create a new phi node for a value say %10, we have to replace
all of the uses of %10 that are no longer dominated by the def of %10
by the result id of the new phi.  However, if the use is in a phi node,
it is possible that the bb contains the use is not dominated by either.
In this case, needs to be handled differently.

### DIFF
--- a/source/opt/merge_return_pass.cpp
+++ b/source/opt/merge_return_pass.cpp
@@ -239,8 +239,21 @@ void MergeReturnPass::CreatePhiNodesForInst(BasicBlock* merge_block,
   if (inst.result_id() != 0) {
     std::vector<Instruction*> users_to_update;
     context()->get_def_use_mgr()->ForEachUser(
-        &inst, [&users_to_update, &dom_tree, inst_bb, this](Instruction* user) {
-          BasicBlock* user_bb = context()->get_instr_block(user);
+        &inst, [&users_to_update, &dom_tree, &inst, inst_bb, this](Instruction* user) {
+          BasicBlock* user_bb = nullptr;
+          if (user->opcode() != SpvOpPhi) {
+            user_bb = context()->get_instr_block(user);
+          } else {
+            // For OpPhi, the use should be considered to be in the predecessor.
+            for(uint32_t i = 0; i < user->NumInOperands(); i+=2) {
+              if (user->GetSingleWordInOperand(i) == inst.result_id()) {
+                uint32_t user_bb_id = user->GetSingleWordInOperand(i+1);
+                user_bb = context()->get_instr_block(user_bb_id);
+                break;
+              }
+            }
+          }
+
           // If |user_bb| is nullptr, then |user| is not in the function.  It is
           // something like an OpName or decoration, which should not be
           // replaced with the result of the OpPhi.
--- a/source/opt/merge_return_pass.cpp
+++ b/source/opt/merge_return_pass.cpp
@@ -204,7 +204,11 @@ void MergeReturnPass::BranchToBlock(BasicBlock* block, uint32_t target) {
     RecordReturned(block);
     RecordReturnValue(block);
   }
+
   BasicBlock* target_block = context()->get_instr_block(target);
+  if (target_block->GetLoopMergeInst()) {
+    cfg()->SplitLoopHeader(target_block);
+  }
   UpdatePhiNodes(block, target_block);
 
   Instruction* return_inst = block->terminator();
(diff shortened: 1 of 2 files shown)


source/opt/merge_return_pass.cpp
```diff
@@ -239,8 +239,21 @@ void MergeReturnPass::CreatePhiNodesForInst(BasicBlock* merge_block,
...
+          BasicBlock* user_bb = nullptr;
+          if (user->opcode() != SpvOpPhi) {
+            user_bb = context()->get_instr_block(user);
+          } else {
+            // For OpPhi, the use should be considered to be in the predecessor.
+            for(uint32_t i = 0; i < user->NumInOperands(); i+=2) {
+              if (user->GetSingleWordInOperand(i) == inst.result_id()) {
+                uint32_t user_bb_id = user->GetSingleWordInOperand(i+1);
+                user_bb = context()->get_instr_block(user_bb_id);
+                break;
```
alan-baker: What happens if the same value is merged across more than one predecessor? OpSwitch could generate an OpPhi with 3+ incoming edges with 2+ that use the same value.

s-perron: I thought of this.  We will replace all of the uses of that id.  The case where this can go wrong is if there is a case where we want to replace it in one but not the other.  I'll try harder to try to come up with a case like that.

s-perron: This part of the code should be fine.  I do have an informal proof of correctness.  We can discuss it in person.  However, I did find a bug in another part of the code.  I'll try to get that fixed up tomorrow and make it part of this PR.  It is similar.  We are not rewritting an OpPhi correctly.

source/opt/merge_return_pass.cpp
```diff
@@ -239,8 +243,22 @@ void MergeReturnPass::CreatePhiNodesForInst(BasicBlock* merge_block,
...
     context()->get_def_use_mgr()->ForEachUser(
-        &inst, [&users_to_update, &dom_tree, inst_bb, this](Instruction* user) {
-          BasicBlock* user_bb = context()->get_instr_block(user);
+        &inst,
+        [&users_to_update, &dom_tree, &inst, inst_bb, this](Instruction* user) {
+          BasicBlock* user_bb = nullptr;
+          if (user->opcode() != SpvOpPhi) {
+            user_bb = context()->get_instr_block(user);
+          } else {
+            // For OpPhi, the use should be considered to be in the predecessor.
```
dnovillo: But not just in the predecessor, right?  A Phi operand could come from any arbitrary CFG path that reaches this basic block.  I'm thinking a pseudocode like this:

```
BB0:
%4 = ...
BB1:
....
if (cond)
THEN_BB:
   ...
ELSE_BB:
   ...
   %5 = ...
ENDIF_BB:
%10 = OpPhi(%4@BB0, %5@ELSE_BB)
```

`BB0` is not a predecessor of `ENDIF_BB`, but it reaches it.

Or am I just confused by the phrasing of the comment?  The code seems to check just predecessors.

s-perron: I don't think your example is correct.  An OpPhi will mention the immediate predecessors of the node.  Your example should be:

```
BB0:
%4 = ...
BB1:
....
if (cond)
THEN_BB:
   ...
ELSE_BB:
   ...
   %5 = ...
ENDIF_BB:
%10 = OpPhi(%4, THEN_BB, %5, ELSE_BB)
```

We will pretend that the use of `%4` took place in `THEN_BB`, so we need to make sure that the def of `%4` dominates `THEN_BB`, not `ENDIF_BB`.

dnovillo: > I don't think your example is correct. An OpPhi will mention the immediate predecessors of the node. Your example should be:
> 
> ```
> BB0:
> %4 = ...
> BB1:
> ....
> if (cond)
> THEN_BB:
>    ...
> ELSE_BB:
>    ...
>    %5 = ...
> ENDIF_BB:
> %10 = OpPhi(%4, THEN_BB, %5, ELSE_BB)
> ```

Sorry, yes.  `%4` comes from `THEN_BB`

> We will pretend that the use of `%4` took place in `THEN_BB`, so we need to make sure that the def of `%4` dominates `THEN_BB`, not `ENDIF_BB`.

Ah, that predecessor.   Now I get it.  Thanks.

source/opt/merge_return_pass.cpp
s-perron: I've added a new test case and fixes for another case where OpPhis are getting messed up.  We can discuss if there are other cases that can go wrong.

PR #2173, opened 2018-12-05
Test for WorkgroupId not being a uvec3

### LINKED ISSUE #1545: spirv-val does not verify that WorkgroupId is a three-component vector of 32-bit integers
I recently ran into a crash with a shader on NVIDIA, the issue is documented in https://github.com/KhronosGroup/glslang/issues/1371

The problem is that the SPIR-V generated there contains a WorkgroupId as type uint2 and the spec says:

"The variable decorated with code:WorkgroupId must: be declared as a
three-component vector of 32-bit integers.”

This bug is to request that spirv-val check this requirement.

ale64bit:
Fixes #1545 

This check seems to be already handled in https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/val/validate_builtins.cpp#L2446.

This is an explicit test for such case. Feel free to discard if it doesn't add any value.

### DIFF
--- a/test/val/val_builtins_test.cpp
+++ b/test/val/val_builtins_test.cpp
@@ -1926,6 +1926,35 @@ OpStore %output_pos %pos
   ASSERT_EQ(SPV_SUCCESS, ValidateInstructions(SPV_ENV_VULKAN_1_0));
 }
 
+TEST_F(ValidateBuiltIns, WorkgroupIdNotVec3) {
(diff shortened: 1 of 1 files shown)


ale64bit:
PTAL

ale64bit:
After re-reading the relevant spec section, I realised that workgroup_id needs to be an Input variable, instead of a constant. I updated the test accordingly.

dneto0:
The AppVeyor failures have been fixed with #2174

PR #2174, opened 2018-12-05
Fix broken recurison validation tests

zoddicus:
This fixes tests that were broken by adding in validation of the order
of function definitions.


zoddicus:
Looks like the same set of prexisting failing tests in the smoketest as previous PRs.

PR #2175, opened 2018-12-05
Merge two implementations of ValidateMemorySemantics

### LINKED ISSUE #2170: Merge two implementations of ValidateMemorySemantics
Like ValidateMemoryScope and ValidateExecutionScope, there are two different implementations of this function that should be unified into a single canonical implementation.

zoddicus:
Fixes #2170


alan-baker:
The smoketest previously had 182 failures in glslang and now has 185. Do you mind investigating the new three and double checking they make sense?

zoddicus:
Thanks for catching that. Looked at all three, they are all rooted in that there is a missing capability that we are now catching. I will file a bug about these against glslang

PR #2176, opened 2018-12-06
 Restrict mask bits for memory semantics in WebGPU

### LINKED ISSUE #2070: Validate Memory Semantics for WebGPU
> #### Memory Semantics
> Among mask bits up to and including 0x10 (SequentiallyConsistent), only the following may be set:
> - Acquire
> - Release
> - AcquireRelease [how does this interact with the memory model?]
>
> The following mask bits may be used in any combination:
> - UniformMemory
> - WorkgroupMemory
> - ImageMemory
> - OutputMemoryKHR
> - MakeAvailableKHR
> - MakeVisibleKHR

zoddicus:
Fail to validate memory semantics value if it includes set bits that
are not on the whitelist from the spec.

Fixes #2070


zoddicus:
... I just realized that this is dependent on another pending PR I have, I will re-upload it once that one has landed.

PR #2181, opened 2018-12-06
Validate Uniform decoration

dneto0:


dneto0:
Looks like spurious link failure on VS2017

PR #2187, opened 2018-12-10
Add RemoveOpNameInstruction reduction pass

hevrard:
Add a spirv-reduce pass which removes OpName and OpMemberName instructions.

This is useful to enable other reduction passes, e.g. RemoveUnreferencedInstruction may not be able to remove an instruction creating an id whose only usage is an OpName for this id.

### DIFF
--- a/source/reduce/remove_opname_instruction_reduction_pass.cpp
+++ b/source/reduce/remove_opname_instruction_reduction_pass.cpp
@@ -28,7 +28,7 @@ RemoveOpNameInstructionReductionPass::GetAvailableOpportunities(
   std::vector<std::unique_ptr<ReductionOpportunity>> result;
 
   for (auto& inst : context->module()->debugs2()) {
-    if (inst.opcode() == SpvOpName) {
+    if (inst.opcode() == SpvOpName || inst.opcode() == SpvOpMemberName) {
       result.push_back(MakeUnique<RemoveInstructionReductionOpportunity>(&inst));
     }
   }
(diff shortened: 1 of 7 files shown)


source/reduce/remove_opname_instruction_reduction_pass.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+using namespace opt;
+
+std::vector<std::unique_ptr<ReductionOpportunity>>
+RemoveOpNameInstructionReductionPass::GetAvailableOpportunities(
+    opt::IRContext* context) const {
+  std::vector<std::unique_ptr<ReductionOpportunity>> result;
+
+  for (auto& inst : context->module()->debugs2()) {
+    if (inst.opcode() == SpvOpName || inst.opcode() == SpvOpMemberName) {
```
s-perron: Just something to watch out for.  Some tools rely on the names of variables to do reflection.  I don't think this will effect you.

hevrard: Thanks for the approval!

Regarding name of variable and reflection: I actually though we might remove an OpName only when it is the only usage of the given id. But then I though that we should not second-guess too much, and keep passes as simple as possible for now. We can still tweak this reduction pass later if needed!

afd: I am in favour of not second-guessing and keeping passes simple.

When using the reducer in practice, the user passes in an interestingness function to decide whether a reduced module should be considered interesting.  This can contain custom checks for the user's scenario, such as whether a given name is still declared in a module.

PR #2188, opened 2018-12-10
Add validation for VK_KHR_shader_float16_int8

JoseEmilio-ARM:


JoseEmilio-ARM:
Debugging the CI assertions

JoseEmilio-ARM:
The CI helped find and fix a couple of bugs, however I think it is lacking tests specifically targeted at the checks I introduced, working on adding one now.

source/val/validate_instruction.cpp
```diff
@@ -506,6 +506,24 @@ spv_result_t InstructionPass(ValidationState_t& _, const Instruction* inst) {
                   "outside of a function";
       }
     }
+    if (_.IncludesIntOrFloatOfWidth(inst->type_id(), SpvOpTypeInt, 8) &&
```
alan-baker: These checks belong in ValidateVariable in validate_memory.cpp; however, they seem incorrect. `OpCapability Int8` should allow full use of 8-bit integers. The checks seem too strict. 

Code exists in the validation state to control whether int8 or float16 can be declared and/or used. You should probably expand on that.

source/val/validation_state.cpp
```diff
@@ -831,6 +831,35 @@ bool ValidationState_t::IsPointerType(uint32_t id) const {
...
+  const Instruction* inst = FindDef(id);
+  assert(inst);
+  switch (inst->opcode()) {
+    case SpvOpTypeInt:
+    case SpvOpTypeFloat:
+    case SpvOpTypeBool:
+    case SpvOpTypeVector:
+    case SpvOpTypeMatrix:
+    case SpvOpTypeArray:
+      return GetComponentType(inst->id()) == check_type &&
```
alan-baker: GetComponentType returns the result id of the component type, not the opcode.

source/val/validate_instruction.cpp
JoseEmilio-ARM: Hi, thank you very much for looking into this. I am looking for a way of validating the following rule in the spec:

>`ifdef::VK_KHR_shader_float16_int8+!VK_KHR_8bit_storage[]`
Any OpVariable with the result type pointing to an 8-bit integer object or an object containing an 8-bit integer element must: not have one of the following as its Storage Class operand:
Uniform
PushConstant
`ifdef::VK_VERSION_1_1,VK_KHR_storage_buffer_storage_class[]`
StorageBuffer
`endif::VK_VERSION_1_1,VK_KHR_storage_buffer_storage_class[]
endif::VK_KHR_shader_float16_int8+!VK_KHR_8bit_storage[]`

Therefore my first guess was that my check would have to look something like:

```
if (_.IncludesIntOrFloatOfWidth(inst->type_id(), SpvOpTypeInt, 8) &&
*** something *** &&
(storage_class == SpvStorageClassUniform ||
storage_class == SpvStorageClassPushConstant ||
(_.HasCapability(SpvCapabilityStorageBuffer8BitAccess) &&
storage_class == SpvStorageClassStorageBuffer))) {
return _.diag(SPV_ERROR_INVALID_LAYOUT, inst)
<< "Variables can not point to an 8-bit integer "
"without 8-bit storage capabilities";
}
```

However I am not sure what this `something` should be. Originally I thought I had to check for what Vulkan extensions are enabled, but this is not possible from SPIRV-Tools?

alan-baker: No, the validator does not know what Vulkan extensions are enabled in general. There is enough information in SPIR-V to determine what is allowable in most cases. The `something` in your first version was too strict because the `Int8` capability allows full use of 8-bit integers.

s-perron:
@alan-baker Where do we stand with this PR?  I'd like to clean up some of these old PRs.

source/val/validate_instruction.cpp
JoseEmilio-ARM: > Code exists in the validation state to control whether int8 or float16 can be declared and/or used. You should probably expand on that.

I found
```
  switch (cap) {
...
    case SpvCapabilityInt8:
      features_.use_int8_type = true;
      features_.declare_int8_type = true;
      break;
```
But not sure how to implement this validation here, since we need to validate the StorageClass of each OpVariable.

>These checks belong in ValidateVariable in validate_memory.cpp; however, they seem incorrect. OpCapability Int8 should allow full use of 8-bit integers. The checks seem too strict.

I moved the checks to validate_memory.cpp removing the check for Int8 capability, but still check that 8bit_storate is _not_ enabled.

source/val/validation_state.cpp
JoseEmilio-ARM: True! I think I fixed this and a couple other issues in the latest version

source/val/validate_memory.cpp
```diff
@@ -427,6 +427,27 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
       }
     }
   }
+  // Vulkan Appendix A: Validation Rules within a Module
+  // Check when VK_KHR_shader_float16_int8 is enabled but
+  // VK_KHR_8bit_storage or VK_KHR_16bit_storage are not.
+  if (_.IncludesIntOrFloatOfWidth(result_type->word(3), SpvOpTypeInt, 8) &&
```
alan-baker: SPIR-V capabilities are not a direct mapping for the Vulkan functionality. In SPIR-V, 8-bit storage is accessible from the 8-bit storage capabilities *or* the Int8 capability. Similar for half floats. Some work likely belongs in the validation layers to check that the right Vulkan feature bits are enabled, but in SPIR-V if the Int8 capability is present, that is sufficient to be able to store and use 8-bit integers.

JoseEmilio-ARM: Hi @alan-baker , we merged https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/514/commits/c6c1561a632eda94c86fd5cb7d8b3aeea1ffca46 , would that be enough coverage for this extension?

alan-baker: Probably if only because adding more stringent checks will be quite difficult. The validation layers would have to inspect the SPIR-V with respect to the Vulkan feature bits. If there were any load/store containing an 8 or 16 bit type then the associated storage feature bit in Vulkan would have to be set. This would be fairly awkward to put into the validation layers is my understanding.

JoseEmilio-ARM: Agreed. I will close this PR then, thank you very much for your help!

PR #2189, opened 2018-12-10
Invalidate the decoration manager at the start of ADCE.

s-perron:
If the decoration manager is kept live the the contex will try to keep
it up to date.  ADCE deals with group decorations by changing the
operands in |OpGroupDecorate| instructions directly without informing
the decoration manager.  This puts it in an invalid state, which will
cause an error when the context tries to update it.  To Avoid this
problem, we will invalidate the decoration manager upfront.

At the same time, the decoration manager is now considered when checking
the consistency of the decoration manager.


dnovillo:
I just saw the failures.  Are they related to the change, or are they due to the shaderc issues we've been having?

s-perron:
No, the only one that should be failing is the shaderc-smoketest.

PR #2193, opened 2018-12-11
Disable checking that AtomicStorage capability is present

zoddicus:
There is inconsistencies between the different specs about whether or
not this capability is required/allowed, so tooling like glslang
currently ignores it. Once this is resolved the check and test can be
re-enabled.


zoddicus:
This is related to https://github.com/KhronosGroup/glslang/issues/1618

s-perron:
@zoddicus  Can you check the formatting?

Issue #2091, opened 2018-11-21
Validator and disassembler produce vague errors


Bidski:
I just installed the master branch so that I can validate a spirv module, but I am getting a very vague error. 

I was under the impression that there was functionality to display the offending instruction in the validator, but this is not happening with the master branch? Did this get removed?

```
$ spirv-val ./fetch.spv
error: line 122: Id is 0
```

The disassembler also doesn't give much insight. I **think** that the disassembler has stopped (just before?) the offending instruction, but from what I can see of the instructions that should follow none of them have an ID of 0.
```
$ spirv-dis --raw-id --offsets ./fetch.spv
; SPIR-V
; Version: 1.3
; Generator: Khronos; 0
; Bound: 144
; Schema: 0
               OpCapability Addresses ; 0x00000014
               OpCapability Linkage ; 0x0000001c
               OpCapability Kernel ; 0x00000024
               OpCapability ImageBasic ; 0x0000002c
               OpCapability LiteralSampler ; 0x00000034
          %1 = OpExtInstImport "OpenCL.std" ; 0x0000003c
               OpMemoryModel Physical32 OpenCL ; 0x00000050
          %2 = OpTypeInt 32 0 ; 0x0000005c
          %3 = OpTypeFloat 32 ; 0x0000006c
          %4 = OpTypeBool ; 0x00000078
          %5 = OpTypeVector %2 2 ; 0x00000080
          %6 = OpTypeVector %2 3 ; 0x00000090
          %7 = OpTypeVector %2 4 ; 0x000000a0
          %8 = OpTypeVector %3 2 ; 0x000000b0
          %9 = OpTypeVector %3 3 ; 0x000000c0
         %10 = OpTypeVector %3 4 ; 0x000000d0
         %15 = OpTypeImage %3 2D 0 0 0 1 Unknown ReadOnly ; 0x000000e0
         %16 = OpTypeSampler ; 0x00000108
         %17 = OpTypePointer UniformConstant %16 ; 0x00000110
         %18 = OpTypeSampledImage %15 ; 0x00000120
         %19 = OpTypeFunction %3 %15 %16 %8 ; 0x0000012c
         %28 = OpTypeFunction %10 %15 %16 %8 %8 ; 0x00000144
         %11 = OpConstant %2 0 ; 0x00000160
         %12 = OpConstant %3 0 ; 0x00000170
         %13 = OpConstant %2 1 ; 0x00000180
[...]

PR #2195, opened 2018-12-12
Output disassembly line number for binary parse errors.

dj2:
This Cl changes the binary parser to keep track of the instruction count
being processed. The parser will then use that instruction number as the
error number, instead of the binary word.

This should make it easier to match the error up to what the
disassembler would output for the error.

Issue #2091

### DIFF
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -123,8 +123,8 @@ class Parser {
   // returned object will be propagated to the current parse's diagnostic
   // object.
   spvtools::DiagnosticStream diagnostic(spv_result_t error) {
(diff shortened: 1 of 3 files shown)


dj2:
What is everyone's thoughts on this? It changes the index of the position to be the instruction number, instead of the word count, which is changing the external contract.

Using the line_number field doesn't work if the input was binary, it would only work if the input was text as the diagnostic does an isTextSource check before outputting line/column information.

dneto0:
The binary word count vs. line number difference was an unfortunate accidental inconsistency.  I think line number is more useful.  So I'm ok with this, provided a note eventually lands in the CHANGES file about it. (Thanks Steven!)

PR #2216, opened 2018-12-17
Add --target-env flag to spirv-opt

### LINKED ISSUE #2199: Unable to pass in target environment to spirv-opt
Currently the target env is hard coded to the default environment in the main function. As part of adding support for a WebGPU specific optimization, being able to set value needs to added.

Implementing this functionality can broken off into its own task and generalized by adding a flag, --target-env, which will work similarly to the same flag spirv-val.

zoddicus:
Fixes #2199

### DIFF
--- a/tools/opt/opt.cpp
+++ b/tools/opt/opt.cpp
@@ -341,7 +341,7 @@ Options (in lexicographical order):
                reflection information defined by SPV_GOOGLE_hlsl_functionality1.
   --target-env=<env>
                Set the target environment. Without this flag the target
-               enviroment defaults to spv1.0.
+               enviroment defaults to spv1.3.
                <env> must be one of vulkan1.0, vulkan1.1, opencl2.2, spv1.0,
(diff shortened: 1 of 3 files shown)


tools/opt/opt.cpp
```diff
@@ -338,6 +339,11 @@ Options (in lexicographical order):
   --strip-reflect
                Remove all reflection information.  For now, this covers
                reflection information defined by SPV_GOOGLE_hlsl_functionality1.
+  --target-env=<env>
+               Set the target environment. Without this flag the target
+               enviroment defaults to spv1.3.
+               <env> must be one of vulkan1.0, vulkan1.1, opencl2.2, spv1.0,
+               spv1.1, spv1.2, spv1.3, or webgpu0.
```
dnovillo: State the default value here?

dnovillo: Ah, never mind.  This is only done if the flag is passed, so the default is set elsewhere

PR #2217, opened 2018-12-17
Implement initial --webgpu-mode flag

### LINKED ISSUE #2166: Create high level recipe for WebGPU in optimizer
This should at first just do a dead code removal pass.

zoddicus:
Fixes #2166

### DIFF
--- a/source/opt/optimizer.cpp
+++ b/source/opt/optimizer.cpp
@@ -216,6 +216,12 @@ Optimizer& Optimizer::RegisterSizePasses() {
       .RegisterPass(CreateAggressiveDCEPass());
 }
 
+Optimizer& Optimizer::RegisterWebGPUPasses() {
+  return RegisterPass(CreateEliminateDeadFunctionsPass())
+      .RegisterPass(CreateAggressiveDCEPass())
+      .RegisterPass(CreateEliminateDeadConstantPass());
+}
+
 bool Optimizer::RegisterPassesFromFlags(const std::vector<std::string>& flags) {
   for (const auto& flag : flags) {
     if (!RegisterPassFromFlag(flag)) {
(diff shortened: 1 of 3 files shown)


zoddicus:
@s-perron, can you confirm that these are sensible flags to be setting for dead code elimation? I suspect agressive might already cover the other flags.

source/opt/optimizer.cpp
```diff
@@ -216,6 +216,12 @@ Optimizer& Optimizer::RegisterSizePasses() {
       .RegisterPass(CreateAggressiveDCEPass());
 }
 
+Optimizer& Optimizer::RegisterWebGPUPasses() {
+  return RegisterPass(CreateEliminateDeadFunctionsPass())
+      .RegisterPass(CreateAggressiveDCEPass())
+      .RegisterPass(CreateEliminateDeadConstantPass());
```
s-perron: If you really just want removal of dead code, and not something that will look to simplify obvious stuff, you should do AggressiveDCEPass and DeadBranchElimination.  The other two that you have are redundant when you have ADCE.

tools/opt/opt.cpp
```diff
@@ -359,6 +359,18 @@ Options (in lexicographical order):
...
+  --webgpu-mode
+               Turns on the prescribed passes for WebGPU and sets the target
+               environmet to webgpu0. Other passes may be turned on via
+               additional flags, but such combinations are not tested.
+               Using --target-env with this flag is not allowed.
+               This flag is the equivalent of setting the following flags:
+                 --target-env=webgpu0
+                 --eliminate-dead-functions
+                 --eliminate-dead-code-aggressive
+                 --eliminate-dead-const
```
s-perron: There is a way of getting the options for the optimization passes printed without hard coding the order here.  See how -O and -Os work.

source/opt/optimizer.cpp
zoddicus: Done

tools/opt/opt.cpp
zoddicus: Done

PR #2218, opened 2018-12-17
Added additional changes for the new AccelerationStructureNV type.

kholtnv:


kholtnv:
BTW the CI-check-format problem is in: source/opt/inline_pass.cpp. I didn't touch that file.

PR #2258, opened 2019-01-02
Handle CompositeInsert with no indices in VDCE

s-perron:
In the spec, there it nothing that forces an OpCompositeInsert to have
an index, but VDCE assumes there is at least 1 in a couple places.

This commit updates VDCE to handle these cases.

### DIFF
--- a/source/opt/vector_dce.cpp
+++ b/source/opt/vector_dce.cpp
@@ -113,30 +113,44 @@ void VectorDCE::MarkInsertUsesAsLive(
     std::vector<VectorDCE::WorkListItem>* work_list) {
   analysis::DefUseManager* def_use_mgr = context()->get_def_use_mgr();
 
-  uint32_t insert_position =
-      current_item.instruction->GetSingleWordInOperand(2);
+  if (current_item.instruction->NumInOperands() > 2) {
+    uint32_t insert_position =
+        current_item.instruction->GetSingleWordInOperand(2);
 
-  // Add the elements of the composite object that are used.
-  uint32_t operand_id =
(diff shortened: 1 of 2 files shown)


source/opt/vector_dce.cpp
```diff
@@ -324,13 +338,25 @@ bool VectorDCE::RewriteInsertInstruction(
...
   bool modified = false;
+
+  if (current_inst->NumInOperands() == 2) {
+    // If there are no indices, then this is the same as a copy.
+    modified = true;
+    context()->KillNamesAndDecorates(current_inst->result_id());
+    uint32_t object_id =
+        current_inst->GetSingleWordInOperand(kInsertObjectIdInIdx);
+    context()->ReplaceAllUsesWith(current_inst->result_id(), object_id);
+    return modified;
```
dnovillo: Just `return true` here?

source/opt/vector_dce.cpp
```diff
@@ -324,13 +338,25 @@ bool VectorDCE::RewriteInsertInstruction(
...
+  }
+
   uint32_t insert_index = current_inst->GetSingleWordInOperand(2);
   if (!live_components.Get(insert_index)) {
     modified = true;
     context()->KillNamesAndDecorates(current_inst->result_id());
     uint32_t composite_id =
         current_inst->GetSingleWordInOperand(kInsertCompositeIdInIdx);
     context()->ReplaceAllUsesWith(current_inst->result_id(), composite_id);
+    return modified;
```
dnovillo: Likewise.

PR #2259, opened 2019-01-02
Broader check for ids that require a type

alan-baker:
Fixes https://crbug.com/911700

* Adds a broader check for when id operands require a type
* updated a few tests
* added a test to catch the original issue

I'm not certain whether this check is too complicated in this location, but it does catch a large class of errors at a single location.

### DIFF
--- a/test/val/val_memory_test.cpp
+++ b/test/val/val_memory_test.cpp
@@ -1513,6 +1513,29 @@ OpFunctionEnd
   EXPECT_EQ(SPV_SUCCESS, ValidateInstructions(SPV_ENV_UNIVERSAL_1_3));
 }
 
+TEST_F(ValidateMemory, ArrayLengthStructIsLabel) {
+  const std::string spirv = R"(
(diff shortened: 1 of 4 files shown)


test/val/val_memory_test.cpp
```diff
@@ -1513,6 +1513,29 @@ OpFunctionEnd
...
+%24 = OpArrayLength %uint %20 0
+%25 = OpLoad %v4float %24
+OpReturnValue %25
+OpFunctionEnd
+)";
+
+  CompileSuccessfully(spirv);
+  EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("Operand 7[%7] requires a type"));
```
dneto0: It's confusing which Id is wrong.  I think it's the label %20?   Recommend adding an OpName to give it a clear name that matches up in the error.

PR #2262, opened 2019-01-03
[kokoro] Fix usage of ERRORLEVEL in Windows script.

ehsannas:


ehsannas:
is CI-windows-VS2017-debug supposed to be failing ?

dj2:
It isn't supposed to but it looks like the same thing @fjhenigman is looking at for the shaderc build. So, I'm guessing if we make this change for effcee we'll see it's broken as well .....

PR #2264, opened 2019-01-04
WIP: Permit UConvert spec-constant op for SPV_AMD_gpu_shader_int16

dneto0:
See https://github.com/KhronosGroup/glslang/issues/848

### DIFF
--- a/source/val/validate_constants.cpp
+++ b/source/val/validate_constants.cpp
@@ -329,11 +329,19 @@ spv_result_t ValidateSpecConstantOp(ValidationState_t& _,
       }
       break;
 
+    case SpvOpUConvert:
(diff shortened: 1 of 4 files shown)


dneto0:
This will fail one test in the Glslang tests, due to KhronosGroup/glslang#848 (See also KhronosGroup/glslang#1643)

dneto0:
Rebased and to work around GCC problem, don't use raw-string adjacent to ordinary string.

dneto0:
VS 2017 issue is a known RE2 issue.
The smoketest failure is known side effect of this change; will be fixed by a glslang known-good update.

dneto0:
Spoke to @johnkslang and he still thinks this is the most helpful thing for the ecosystem.

PR #2266, opened 2019-01-04
Upgrade modf and frexp

### LINKED ISSUE #2138: Upgrade Vulkan memory model should handle GLSL.std.450 Modf and Frexp
The Modf and Frexp instructions do a normal store through a pointer.  Those stores may have to be upgraded with MakeAvailable and NonPrivate flags. 

See https://github.com/KhronosGroup/SPIRV-Tools/pull/1904#pullrequestreview-180023433

alan-baker:
Fixes #2138

* Modf and frexp are upgraded to use the struct version of the
instruction and generate an explicit store whose flags can be upgraded
seperately
* Fixed major bug where availability and visibility were reversed for
non-copy memory instructions
* Upgraded all opt tests to use SPV_ENV_UNIVERSAL_1_3
* Started running the upgrade tests
 * was missing from CMakeLists.txt and had old define

### DIFF
--- a/source/opt/upgrade_memory_model.cpp
+++ b/source/opt/upgrade_memory_model.cpp
@@ -160,14 +160,14 @@ void UpgradeMemoryModel::UpgradeInstructions() {
       }
       // According to SPV_KHR_vulkan_memory_model, if both available and
       // visible flags are used the first scope operand is for availability
-      // (reads) and the second is for visibility (writes).
-      if (src_coherent) {
-        inst->AddOperand(
-            {SPV_OPERAND_TYPE_SCOPE_ID, {GetScopeConstant(src_scope)}});
-      }
+      // (writes) and the second is for visibility (reads).
       if (dst_coherent) {
         inst->AddOperand(
             {SPV_OPERAND_TYPE_SCOPE_ID, {GetScopeConstant(dst_scope)}});
+      }
+      if (src_coherent) {
+        inst->AddOperand(
+            {SPV_OPERAND_TYPE_SCOPE_ID, {GetScopeConstant(src_scope)}});
       }
     });
   }
--- a/source/opt/upgrade_memory_model.cpp
+++ b/source/opt/upgrade_memory_model.cpp
@@ -613,7 +613,7 @@ void UpgradeMemoryModel::UpgradeExtInst(Instruction* ext_inst) {
       context()->get_type_mgr()->GetTypeInstruction(&struct_type);
   // Change the operation
   GLSLstd450 new_op = is_modf ? GLSLstd450ModfStruct : GLSLstd450FrexpStruct;
-  ext_inst->SetOperand(3u, {new_op});
+  ext_inst->SetOperand(3u, {static_cast<uint32_t>(new_op)});
   // Remove the pointer argument
   ext_inst->RemoveOperand(5u);
   // Set the type id to the new struct.
--- a/source/opt/upgrade_memory_model.cpp
+++ b/source/opt/upgrade_memory_model.cpp
@@ -16,6 +16,7 @@
 
 #include <utility>
 
+#include "source/opt/ir_builder.h"
 #include "source/opt/ir_context.h"
 #include "source/util/make_unique.h"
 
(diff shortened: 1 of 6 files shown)


source/opt/upgrade_memory_model.cpp
```diff
@@ -68,6 +69,22 @@ void UpgradeMemoryModel::UpgradeInstructions() {
...
+  // Upgrade modf and frexp first since they generate new stores.
+  for (auto& func : *get_module()) {
+    func.ForEachInst([this](Instruction* inst) {
+      if (inst->opcode() == SpvOpExtInst) {
+        auto import =
+            get_def_use_mgr()->GetDef(inst->GetSingleWordInOperand(0u));
+        if (reinterpret_cast<char*>(import->GetInOperand(0u).words.data()) ==
+            std::string("GLSL.std.450")) {
+          auto ext_inst = inst->GetSingleWordInOperand(1u);
+          if (ext_inst == GLSLstd450Modf || ext_inst == GLSLstd450Frexp) {
```
dneto0: For speed reasons, I'd first test against ext_inst being Modf or Frexp.  Then do the string compare test. (The ext_inst opcode is always present)

test/opt/upgrade_memory_model_test.cpp
```diff
@@ -1430,6 +1429,288 @@ OpFunctionEnd
...
+; CHECK: [[float:%\w+]] = OpTypeFloat 32
+; CHECK: [[float_0:%\w+]] = OpConstant [[float]] 0
+; CHECK: [[ptr:%\w+]] = OpTypePointer StorageBuffer [[float]]
+; CHECK: [[var:%\w+]] = OpVariable [[ptr]] StorageBuffer
+; CHECK: [[struct:%\w+]] = OpTypeStruct [[float]] [[float]]
+; CHECK: [[modfstruct:%\w+]] = OpExtInst [[struct]] {{%\w+}} ModfStruct [[float_0]]
+; CHECK: [[ex0:%\w+]] = OpCompositeExtract [[float]] [[modfstruct]] 0
+; CHECK: [[ex1:%\w+]] = OpCompositeExtract [[float]] [[modfstruct]] 1
+; CHECK: OpStore [[var]] [[ex1]]
+; CHECK-NOT: NonPrivatePointerKHR
```
dneto0: Why shouldn't NonPrivatePointerKHR be there?  It seems it's the only safe thing to do, barring specific knowledge (and the compiler doesn't have that knowledge)

source/opt/upgrade_memory_model.cpp
```diff
@@ -581,5 +598,43 @@ bool UpgradeMemoryModel::IsDeviceScope(uint32_t scope_id) {
...
+      context(), where,
+      IRContext::kAnalysisDefUse | IRContext::kAnalysisInstrToBlockMapping);
+  auto extract_0 =
+      builder.AddCompositeExtract(element_type_id, ext_inst->result_id(), {0});
+  context()->ReplaceAllUsesWith(ext_inst->result_id(), extract_0->result_id());
+  // The extract's input was just changed to itself, so fix that.
+  extract_0->SetInOperand(0u, {ext_inst->result_id()});
+  auto extract_1 =
+      builder.AddCompositeExtract(pointee_type_id, ext_inst->result_id(), {1});
+  builder.AddStore(ptr_id, extract_1->result_id());
```
dneto0: I'm confused why we don't do anything at all to add NonPrivateKHR.
I see the tests are quite intentional on this (thank you).

At the very least I'd like a comment here saying that NonPrivateKHR and the availability op will be applied later as required.

test/opt/upgrade_memory_model_test.cpp
alan-baker: The pointer is not annotated with Coherent. All the upgrades currently require either coherent or being Workgroup storage class to get NonPrivate.

Do you think the default should be NonPrivate operations? That's a change that should apply widely to this pass.

source/opt/upgrade_memory_model.cpp
alan-baker: The store that gets added is then processed  with all the others. There's a comment in UpgradeInstructions() to that effect. If the pointer is identified as coherent the store will be flagged for NonPrivate and MakeAvailable.

alan-baker: Worth noting that's what the next tests check.

source/opt/upgrade_memory_model.cpp
alan-baker: Done

test/opt/upgrade_memory_model_test.cpp
dneto0: Makes sense. Thanks.

PR #2267, opened 2019-01-06
SPV_EXT_physical_storage_buffer

jeffbolznv:


jeffbolznv:
Depends on https://github.com/KhronosGroup/SPIRV-Headers/pull/88.

source/val/validate_decorations.cpp
```diff
@@ -331,6 +336,8 @@ uint32_t getSize(uint32_t member_id, const LayoutConstraints& inherited,
       const auto& constraint = constraints[std::make_pair(lastMember, lastIdx)];
       return offset + getSize(lastMember, constraint, constraints, vstate);
     }
+    case SpvOpTypePointer:
+      return 8;
```
dneto0: Ideally this would be a bit more abstracted, e.g. something set up early in the validation state.  A Physical32 addressing mode would have this as 4, but those environments doesn't run through this code.

source/val/validate_decorations.cpp
```diff
@@ -1127,12 +1135,14 @@ spv_result_t CheckFPRoundingModeForShaders(ValidationState_t& vstate,
     if (storage != SpvStorageClassStorageBuffer &&
         storage != SpvStorageClassUniform &&
         storage != SpvStorageClassPushConstant &&
-        storage != SpvStorageClassInput && storage != SpvStorageClassOutput) {
+        storage != SpvStorageClassInput &&
```
dneto0: We shouldn't have stores to Input.  But adding it is harmless here.

source/val/validate_memory.cpp
```diff
@@ -330,14 +342,23 @@ spv_result_t CheckMemoryAccess(ValidationState_t& _, const Instruction* inst,
...
                 "Workgroup, CrossWorkgroup, Generic, Image or StorageBuffer "
                 "storage classes.";
     }
   }
 
+  if (!(mask & SpvMemoryAccessAlignedMask)) {
+    if (src_sc == SpvStorageClassPhysicalStorageBufferEXT ||
+        dst_sc == SpvStorageClassPhysicalStorageBufferEXT) {
+      return _.diag(SPV_ERROR_INVALID_ID, inst)
+             << "Memory accesses with PhysicalStorageBufferEXT must use Aligned.";
```
dneto0: +1   I just found this in the Vulkan extension text.

source/val/validate_memory.cpp
```diff
@@ -277,7 +277,20 @@ uint32_t GetMakeVisibleScope(const Instruction* inst, uint32_t mask) {
 }
 
 spv_result_t CheckMemoryAccess(ValidationState_t& _, const Instruction* inst,
-                               uint32_t mask) {
+                               uint32_t index) {
+
+  SpvStorageClass dst_sc, src_sc;
+  std::tie(dst_sc, src_sc) = GetStorageClass(_, inst);
+  if (inst->operands().size() <= index) {
```
dneto0: Use of "index" is a nice cleanup of all callers.

source/operand.cpp
```diff
@@ -481,6 +481,9 @@ std::function<bool(unsigned)> spvOperandCanBeForwardDeclaredFunction(
     case SpvOpTypeForwardPointer:
       out = [](unsigned index) { return index == 0; };
       break;
+    case SpvOpTypeArray:
```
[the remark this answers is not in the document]
jeffbolznv: This is allowed. You may be confusing a forward pointer whose pointee type is an array type (not allowed, based on internal issue discussion) vs an array type whose element is a forward pointer type (allowed).

source/val/validate_decorations.cpp
jeffbolznv: I've added this.

source/operand.cpp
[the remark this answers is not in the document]
dneto0: Ah, the dangers of reading code too quickly!

PR #2268, opened 2019-01-07
[TEST]: Update VS2017 for kokoro builds.

s-perron:
A PR to test if I can update vs2017 and get a successful build in the kokoro environment.

### DIFF
--- a/kokoro/scripts/windows/build.bat
+++ b/kokoro/scripts/windows/build.bat
@@ -36,6 +36,9 @@ git clone --depth=1 https://github.com/google/re2                 external/re2
 if %VS_VERSION% == 2017 (
   call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" x64
   echo "Using VS 2017..."
+  echo "Updating VS 2017..."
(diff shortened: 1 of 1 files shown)


kokoro/scripts/windows/build.bat
```diff
@@ -36,6 +36,9 @@ git clone --depth=1 https://github.com/google/re2                 external/re2
 if %VS_VERSION% == 2017 (
   call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" x64
   echo "Using VS 2017..."
+  echo "Updating VS 2017..."
+  call "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vs_installer.exe" update --passive --norestart --installpath "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community"
```
dneto0: Looks like you need to us vs_enterprise.exe and you need two commands.
See https://docs.microsoft.com/en-us/visualstudio/install/command-line-parameter-examples?view=vs-2017#using---installpath

dneto0: And it says " To avoid a User Account Control dialog, run the command prompt as an Administrator."

s-perron: I took the command from https://stackoverflow.com/questions/48901633/how-to-update-visual-studio-2017-using-command-line.  `vs_installer.exe` seems okay (at least on my windows machine). 

You are right about the dialog.  I'll see how the bot responds to what I have.

Issue #1611, opened 2018-06-07
Opt: Implement Code Sinking?


greg-lunarg:
During recent comparison of SPIR-V to DX assembly, it seemed that fxc was performing an optimization that I will call "code sinking". The process of sinking is to move an instruction as deep as possible into conditional control flow. The reason to do this is that, for uniform control flow, execution of code under control flow may be avoided, thus potentially optimizing execution time of the shader. 

Analysis suggests that code sinking in spirv-opt could have performance benefit. And if code is not replicated, such sinking should generally do no harm to performance.

Evidence

Using renderdoc profiling of a SPIR-V shader in a real life Vulkan capture on a prominent graphics platform, I performed several different experiments.

First, I did sinking on the SPIR-V as much as possible manually. Even though the opportunities were not huge, about a 1% improvement was achieved in the shader execution time.

Then I reversed the optimization. I pulled as much code out of conditional control flow as I could and placed it just before the control flow divergence. The time for the shader increased by 20%.

These two results suggest two things: the platform I was using was performance sensitive to the placement of code relative to conditional control flow, and the driver's shader compiler was not doing this optimization.

Initial Restrictions

I think we can probably come up with a set of restrictions for this optimization that capture the lion's share of the benefit and keep the algorithm fairly simple. The algorithm can be made more sophisticated as is needed.

One almost obvious restriction is to not push code into loops. Most shader loops execute at least once so at best this will be a wash and at worst it will increase execution time.

Another restriction is to only push combinator operations (add, mul, etc) and non-indexed loads which shader compilers can hoist back out if they wish. In general pushing sample instructions is safe if the control flow is uniform, but it does require that analysis. For the several cases I looked at, most benefit could still be achieved without pushing samples. Removal of this restriction can be done at a later time as it is needed.

Finally, only push instructions whose results are used in just a single block. This trivially avoids duplication. This may mean we miss some opportunities, but it makes the algorithm much simpler. For the several cases I looked at, most benefit could still be achieved with this restriction. Removal of this restriction can be done at a later time as it is needed.

Other Algorithmic Details

A phi function use will count as a use for the block with which the value is associated through the phi function.

Other Optimization Implications

This optimization may reduce opportunities for if-conversion. But while this may slightly increase the shader size comparatively, its performance should be no worse and likely better as more code is under conditional execution. It is therefore recommended that this be done before if-conversion.

[...]

PR #2284, opened 2019-01-14
Sink

s-perron:
Add code sinking pass.  It will move OpLoad and OpAccessChain instructions as close as possible to their uses.

Part of #1611.

### DIFF
--- a/source/opt/code_sink.cpp
+++ b/source/opt/code_sink.cpp
@@ -27,8 +27,233 @@ namespace opt {
 
 Pass::Status CodeSinkingPass::Process() {
   bool modified = false;
+
+  for (Function& function : *get_module()) {
+    bool has_memory_sync = HasMemorySync(&function);
+    cfg()->ForEachBlockInPostOrder(
+        function.entry().get(), [&modified, has_memory_sync, this](BasicBlock* bb) {
+          if (SinkInstInBB(bb, has_memory_sync)) {
+            modified = true;
+          }
+        });
+  }
+
   return modified ? Status::SuccessWithChange : Status::SuccessWithoutChange;
 }
+
+bool CodeSinkingPass::SinkInstInBB(BasicBlock* bb, bool function_has_memory_sync) {
+  bool modified = false;
+  for( auto inst = bb->rbegin(); inst != bb->rend(); ++inst) {
+    if (SinkInst(&*inst, function_has_memory_sync)) {
+      inst = bb->rbegin();
+      modified = true;
+    }
+  }
+  return modified;
+}
+
+bool CodeSinkingPass::SinkInst(Instruction* inst, bool function_has_memory_sync) {
+  if (inst->result_id() == 0) {
+    return false;
+  }
+
+  if (!inst->IsOpcodeCodeMotionSafe()) {
+    return false;
+  }
+
+  if (ReferencesMutableMemory(inst, function_has_memory_sync)) {
+    return false;
+  }
+
+  if (BasicBlock* target_bb = FindNewBasicBlockFor(inst)) {
+
+    Instruction* pos = &*target_bb->begin();
+    while(pos->opcode() == SpvOpPhi) {
+      pos = pos->NextNode();
+    }
+
+    inst->InsertBefore(pos);
+    context()->set_instr_block(inst, target_bb);
+    std::cerr << "Moving instruction: " << inst->PrettyPrint(SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES) << std::endl;
+    std::cerr << "Moving before: " << pos->PrettyPrint(SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES) << std::endl;
+    return true;
+  }
+  return false;
+}
+
+BasicBlock* CodeSinkingPass::FindNewBasicBlockFor(Instruction* inst) {
+  assert(inst->result_id() != 0 && "Instruction should not have a result.");
+  BasicBlock* original_bb = context()->get_instr_block(inst);
+  BasicBlock* bb = original_bb;
+
+  std::cerr << "Looking for a basic block for: " << inst->PrettyPrint(SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES) << std::endl;
+
+  std::unordered_set<uint32_t> bbs_with_uses;
+  get_def_use_mgr()->ForEachUse(inst, [&bbs_with_uses, this](Instruction* use, uint32_t idx) {
+    if (use->opcode() != SpvOpPhi) {
+      bbs_with_uses.insert(context()->get_instr_block(use)->id());
+    } else {
+      bbs_with_uses.insert(use->GetSingleWordOperand(idx+1));
+    }
+  });
+
+  while(true) {
+    if (bbs_with_uses.count(bb->id())) {
+      break;
+    }
+
+    if (bb->terminator()->opcode() == SpvOpBranch) {
+      uint32_t succ_bb_id = bb->terminator()->GetSingleWordInOperand(0);
+      if (cfg()->preds(succ_bb_id).size() == 1) {
+        bb = context()->get_instr_block(succ_bb_id);
+        continue;
+      } else {
+        break;
+      }
+    }
+
+    Instruction* merge_inst = bb->GetMergeInst();
+    if (merge_inst == nullptr) {
+      break;
+    }
+
+    if (merge_inst->opcode() == SpvOpSelectionMerge) {
+      bool used_in_multiple_blocks = false;
+      uint32_t bb_used_in = 0;
+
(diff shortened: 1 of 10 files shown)


s-perron:
I have finished running this through a set of around 3,500 shaders.

First the bad news: the vgpr count on 86 of those shader go up.  Usually by just 1 or 2 registers.  One shader went from 61 vgprs used to 94.  This is because sinking allowed the ICD to unroll a loop, and unrolling/scheduling is what caused the register pressure to increase.  Other cases that went up are similair: it allowed the ICD to do more, and that is what increased register pressure.

The good news.  81 shaders reduced the vector register count by 1-4 registers, and 88 were reduced by 5 or more.  Many were reduced by as much as 26.

Overall, I believe this is a good change, but because it enables optimizations in the ICD, it could end up having a small negative effect.

source/opt/code_sink.h
```diff
@@ -0,0 +1,103 @@
...
+  // possible to the uses of |inst| without increasing the number of time |inst|
+  // will be executed.  Return |nullptr| if there is not need to move |inst|.
+  BasicBlock* FindNewBasicBlockFor(Instruction* inst);
+
+  // Return true if |inst| reference memory and it is possible that the data in
+  // the memory changes at some point.
+  bool ReferencesMutableMemory(Instruction* inst);
+
+  // Returns true if the module contains that instruction that is a memory sync.
+  bool HasUniformMemorySync();
```
dnovillo: "contains a memory sync instruction"?

source/opt/code_sink.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+  if (BasicBlock* target_bb = FindNewBasicBlockFor(inst)) {
+    Instruction* pos = &*target_bb->begin();
+    while (pos->opcode() == SpvOpPhi) {
+      pos = pos->NextNode();
+    }
+
+    inst->InsertBefore(pos);
+    context()->set_instr_block(inst, target_bb);
+    return true;
```
dnovillo: This does not seem to be removing the instruction from its original locus.  Did I miss something?

source/opt/code_sink.cpp
```diff
@@ -0,0 +1,316 @@
...
+bool CodeSinkingPass::IsSyncOnUniform(uint32_t mem_semantics_id) const {
+  const analysis::Constant* mem_semantics_const =
+      context()->get_constant_mgr()->FindDeclaredConstant(mem_semantics_id);
+  assert(mem_semantics_const != nullptr &&
+         "Expecting memory semantics id to be a constant.");
+  assert(mem_semantics_const->AsIntConstant() &&
+         "Memory semantics should be an integer.");
+  uint32_t mem_semantics_int = mem_semantics_const->GetU32();
+
+  // If it does not effect uniform memory, then it is does not apply to uniform
```
dnovillo: s/effect/affect/

source/opt/code_sink.cpp
```diff
@@ -0,0 +1,316 @@
...
+      break;
+    }
+
+    // Check all of the successors of |bb| it see which lead to a use of |inst|
+    // before reaching the merge node.
+    bool used_in_multiple_blocks = false;
+    uint32_t bb_used_in = 0;
+    bb->ForEachSuccessorLabel([this, bb, &bb_used_in, &used_in_multiple_blocks,
+                               &bbs_with_uses](uint32_t* succ_bb_id) {
+      if (IntersectsPath(*succ_bb_id, bb->MergeBlockIdIfAny(), bbs_with_uses)) {
```
dnovillo: Doesn't this fail if MergeBlockIdIfAny returns 0?

source/opt/code_sink.cpp
```diff
@@ -0,0 +1,316 @@
...
+          bb_used_in = *succ_bb_id;
+        } else {
+          used_in_multiple_blocks = true;
+        }
+      }
+    });
+
+    // If more than one successor, which is not the merge block, uses |inst|
+    // then we have to leave |inst| in bb because there is none of the
+    // successors dominate all uses of |inst".
```
dnovillo: s/|inst"/|inst|/

source/opt/code_sink.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+    if (bb_used_in == 0) {
+      // If |inst| is not used before reaching the merge node, then we can move
+      // |inst| to the merge node.
+      bb = context()->get_instr_block(bb->MergeBlockIdIfAny());
+    } else {
+      // If the only successor that leads to a used of |inst| has more than 1
+      // predecessor, then moving |inst| could cause it to be executed more
+      // often, so we cannot move it.
+      if (cfg()->preds(bb_used_in).size() != 1) {
```
dnovillo: Will this also cover the case where the load has 1 predecessor but it's still inside a loop?

source/opt/code_sink.cpp
s-perron: |InsertBefore| removes |inst| from the block it is in before inserting it into its new position.

source/opt/code_sink.cpp
s-perron: I won't be 0 because of the check at line 113.  `bb` has a merge instruction, and will therefore have a merge block.

source/opt/code_sink.cpp
s-perron: Since we are only moving 1 block at a time, if `bb_used_in` is in a loop but `bb` is not, then `bb_used_in` is the loop header.  The continue target for the loop header has to branch back to the loop header as well.  That means at least to predecessor for `bb_used_in`.

source/opt/code_sink.cpp
```diff
@@ -0,0 +1,316 @@
...
+  Instruction* base_ptr = inst->GetBaseAddress();
+  if (base_ptr->opcode() != SpvOpVariable) {
+    return true;
+  }
+
+  if (base_ptr->IsReadOnlyVariable()) {
+    return false;
+  }
+
+  if (HasUniformMemorySync()) {
```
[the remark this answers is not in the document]
s-perron: I've opened #2300 as a follow up issue for this.

tothambrus11:
For the record, I needed these passes to run beforehand to get any useful sinking behavior working:

```cpp
settings.optimizer->RegisterPass(spvtools::CreateInlineExhaustivePass());
            settings.optimizer->RegisterPass(spvtools::CreateScalarReplacementPass(0));
            settings.optimizer->RegisterPass(spvtools::CreateSSARewritePass());
            settings.optimizer->RegisterPass(spvtools::CreateLocalAccessChainConvertPass());
            settings.optimizer->RegisterPass(spvtools::CreateLocalSingleStoreElimPass());
            settings.optimizer->RegisterPass(spvtools::CreateAggressiveDCEPass());
```

There are still many limitations of the sinking, e.g. it doesn't sink in loops, which would benefit the most from this optimization.

I was also wondering what's the reason for allowing exactly these two instructions:
```cpp
if (inst->opcode() != spv::Op::OpLoad &&
      inst->opcode() != spv::Op::OpAccessChain) {
    return false;
  }
```
instead of using something more general, like
```cpp
if (!inst->IsOpcodeCodeMotionSafe()) {
    return false;
}
```
Even with only the above 2 allowed, I was able to get lots of sinks done, I'm not sure how e.g. `fAddOp` succeeded to get moved. Do you think changing it to the latter version would help?

s-perron:
> I was also wondering what's the reason for allowing exactly these two instructions:

Optimization is always a balance. In spirv-opt, we try to be conservative in what we do because we do not have any knowledge of the HW. For example, if you move a load into a loop, it will be executed every time through the loop, which could be expensive, especially in a small loop. On the other hand leaving it out of the loop requires an extra register, which could lower occupancy. 

For these case, we favour leaving the code as the user wrote it. Do not do too much code motion, giving the developer the power to decide if something should be done inside or outside a loop.

This specific pass was written because there are cases where the compiler generates loads in places the user might not expect. For example, in HLSL parameters are often loaded and stored before the function is called in order to do the copy-in. We wanted to experiment with moving those types of load as close as possible to their uses because the users could not write that code.

That is why we did loads and access chains specifically, and we chose not to sink the code into loops.

We didn't see any improvements in performance on the platform we were looking at, so we did not continue with it. That is why it is not part of any of the predefined sets of optimizations.

Do you see performance improvements from this pass? If so, which platforms are you targeting?

tothambrus11:
Hi, thanks for the detailed response!

Our use case was a bit obscure, so I don't think it's representative for the general performance optimizations. We have a shader graph editor where all nodes are translated to function calls, even the If/OpSelect block. There are some use cases where the different branches of an if would do very heavy computation, so we wanted to sink them as much as possible. We achieved pretty nice results with this pass, except where it really mattered - where the computations necessary for calculating a value had a loop that couldn't be unrolled. I guess it's more complicated to recognize such loops and lift them into another block as a whole.

Our approach now is to make our own, higher-level IR where we know which functions are pure and which ones are not, then we can use a similar algorithm as this one to sink pure functions into branches. At that level, functions will encapsulate all loops, so it will be much easier.

PR #2286, opened 2019-01-15
Remove use of deprecated googletest macro

s-perron:
INSTANTIATE_TEST_CASE_P has been deprecated.  We need to use
INSTANTIATE_TEST_SUITE_P instead.

### DIFF
--- a/test/opt/unify_const_test.cpp
+++ b/test/opt/unify_const_test.cpp
@@ -354,8 +354,14 @@ TEST_P(UnifyFrontEndConstantParamTest, TestCase) {
   Check(expected_builder, test_builder);
 }
 
-INSTANTIATE_TEST_CASE_P(Case, UnifyFrontEndConstantParamTest,
-                        ::testing::ValuesIn(std::vector<UnifyConstantTestCase>({
+INSTANTIATE_TEST_SUITE_P(
+    Case, UnifyFrontEndConstantParamTest,
+    ::
+        testing::
+            ValuesIn(
+                std::
+                    vector<UnifyConstantTestCase>(
+                        {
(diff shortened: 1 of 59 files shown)


dneto0:
If I accept this into SPIRV-Tools then it won't play nice with the NDK build,  which uses an older version of googletest.  See https://android.googlesource.com/platform/external/googletest/+log/master   
That is updated only periodically.

gennadiycivil:
Is there any reason not to upgrade to the newer googletest? ( Except that " it is work " :-) of course

s-perron:
There is an issue opened to get the Andriod NDK build to move up.  We have agreement that it will happen.  Just waiting for that to happen.

s-perron:
The Andriod NDK has been updated.  This should be good to go.

test/opt/unify_const_test.cpp
```diff
@@ -354,8 +354,14 @@ TEST_P(UnifyFrontEndConstantParamTest, TestCase) {
   Check(expected_builder, test_builder);
 }
 
-INSTANTIATE_TEST_CASE_P(Case, UnifyFrontEndConstantParamTest,
-                        ::testing::ValuesIn(std::vector<UnifyConstantTestCase>({
+INSTANTIATE_TEST_SUITE_P(
+    Case, UnifyFrontEndConstantParamTest,
```
dneto0: This looks .... wrong.

s-perron: clang-format does not change anything.

s-perron: The format check passed for this commit, so I think I will leave it.

PR #2288, opened 2019-01-15
Add in WebGPU specific memory scope validation

### LINKED ISSUE #2278: Add WebGPU Memory scope check
This was recently finalized, so should be implemented now.

From the [spec](https://github.com/gpuweb/spirv-execution-env/blob/master/execution-env.md):
> **Scope** when used for memory must be one of:
> 
>                                                                                                                                                                          
> [//]: # (No **Device** since that's optional for Vulkan memory model.  In single-device configurations QueueFamilyKHR is the same as Device)
> 
> *   **Workgroup**
> *   **Subgroup**
> *   **QueueFamilyKHR**

zoddicus:
Fixes #2278


zoddicus:
smoketest failure is due to it being unable to find spirv-cross

PR #2289, opened 2019-01-15
Add tests to validate allowed lengths of int and floats for WebGPU

### LINKED ISSUE #2277: Validate that OpTypeInt and OpTypeFloat are the correct size for WebGPU
I suspect this is ensured by other things like capability restrictions, but it would be nice to have some tests verifying.

From the spec:
> Scalar floating-point types (**OpTypeFloat**) must have one of the following widths:
> 
> 
> *   32 bits                                                                                                                                                                  
> 
> Scalar integer types (**OpTypeInteger**) must have one of the following widths:
> 
>                                                                                                                                                                           
> *   32 bits

zoddicus:
Fixes #2277

### DIFF
--- a/test/val/val_data_test.cpp
+++ b/test/val/val_data_test.cpp
@@ -36,6 +36,24 @@ std::string HeaderWith(std::string cap) {
          cap + " OpMemoryModel Logical GLSL450 ";
 }
 
+std::string WebGPUHeaderWith(std::string cap) {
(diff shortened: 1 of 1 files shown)


dneto0:
Smoketest failure due to inadvertent dependency on SPIRV-Cross https://github.com/google/shaderc/issues/542

Issue #2291, opened 2019-01-16
CI-shaderc-smoketest failing due to new dependency on spirv-cross


zoddicus:
Example of the failure: https://source.cloud.google.com/results/invocations/e397caa0-9ecf-4ef3-a5d6-48660adf0ce1/targets

I believe this failure is occuring due this commit in shaderc: https://github.com/google/shaderc/commit/a0a4695765d8641576c3e34294145c345b50dbdc

dneto0:
Should be addressed by https://github.com/google/shaderc/issues/542

dneto0:
https://github.com/google/shaderc/issues/542 was merged, and the smoketests are now passing (on the bots i'm watching)

PR #2293, opened 2019-01-16
Fix overflow when negating INT_MIN.

s-perron:
When doing (-INT_MIN) is considered overflow, so we cannot fold it by
actually performing the negation.

Fixes https://crbug.com/917991


dnovillo:
Is the shaderc failure known already?

```
CMake Error at third_party/CMakeLists.txt:89 (message):
  SPIRV-Cross was not found - required for compilation
```

dneto0:
The smoketest failure is https://github.com/KhronosGroup/SPIRV-Tools/issues/2291

dneto0:
Sorry, please rebase. There's a conflict.

PR #2294, opened 2019-01-16
Instrumentation: Fix vertex instrumentation to use VertexIndex and InstanceIndex

greg-lunarg:
...instead of VertexId and InstanceId


dneto0:
smoke test failure is due to https://github.com/KhronosGroup/SPIRV-Tools/issues/2291

PR #2295, opened 2019-01-16
Fix leaks in tests

dneto0:
Fixes all but 1 leak in test_spirv_unit_tests

Generally, a test fixture in a method that can generatea a binary
should release any previously cached binary.  Similarly for diagnostics.


dneto0:
Actually, it fixes all the leaks in that test executable.

dneto0:
Bot failure is independent, and due to https://github.com/KhronosGroup/SPIRV-Tools/issues/2291

PR #2298, opened 2019-01-16
spirv-val: Emit an error when an OpSwitch target is not an OpLabel

### LINKED ISSUE #1628: OpSwitch error should reference label requirement
The OpSwitch in this example is bad.  It uses a the uint value %7 as a case label, but that's not valid.

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %1 "foo"
       %uint = OpTypeInt 32 0
     %uint_0 = OpConstant %uint 0
       %void = OpTypeVoid
          %5 = OpTypeFunction %void
          %1 = OpFunction %void None %5
          %6 = OpLabel
          %7 = OpCopyObject %uint %uint_0
               OpSelectionMerge %8 None
               OpSwitch %uint_0 %8 0 %7
          %8 = OpLabel
               OpReturn
               OpFunctionEnd
```

The error message is wrong, and misleading:  
```
error: line 8: Block(s) {7} are referenced but not defined in function 1
  %1 = OpFunction %void None %5
```
The value %7 _is_ defined in the function.  The problem is it's not a _label_ that is defined in the function.

JasperNV:
Fixes #1628.

### DIFF
--- a/source/val/validate_cfg.cpp
+++ b/source/val/validate_cfg.cpp
@@ -112,6 +112,19 @@ spv_result_t ValidatePhi(ValidationState_t& _, const Instruction* inst) {
   return SPV_SUCCESS;
 }
 
+spv_result_t ValidateBranch(ValidationState_t& _, const Instruction* inst) {
+  // target operands must be OpLabel
+  const auto id = inst->GetOperandAs<uint32_t>(0);
+  const auto target = _.FindDef(id);
+  if (!target || SpvOpLabel != target->opcode()) {
+    return _.diag(SPV_ERROR_INVALID_ID, inst)
+           << "'Target Label' operands for OpBranch must be the ID "
+              "of an OpLabel instruction";
+  }
+
+  return SPV_SUCCESS;
+}
+
 spv_result_t ValidateBranchConditional(ValidationState_t& _,
                                        const Instruction* inst) {
   // num_operands is either 3 or 5 --- if 5, the last two need to be literal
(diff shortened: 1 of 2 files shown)


source/val/validate_cfg.cpp
```diff
@@ -155,6 +155,27 @@ spv_result_t ValidateBranchConditional(ValidationState_t& _,
...
+  // At least two operands (selector, default), any more than that are
+  // literal/target.
+
+  // target operands must be OpLabel
+  for (auto i = 2; i < num_operands; i += 2) {
+    const auto literal_ = inst->GetOperandAs<uint32_t>(i + 0);
+    const auto id = inst->GetOperandAs<uint32_t>(i + 1);
+
+    const auto target = _.FindDef(id);
+    if (!target || SpvOpLabel != target->opcode()) {
```
dneto0: Hey @alan-baker are there other places in the validator where branches are checked.  All branches need to target another block in the same function.  For example.

alan-baker: This look like the right spot. OpBranchConditional is checked in a similar spot. OpBranch seems to be unchecked currently.

JasperNV: I could add support for OpBranch as well if wanted. That would likely be a separate issue / PR though.

alan-baker: That's definitely an useful check if you don't mind adding it. Not a problem as a separate PR.

JasperNV:
CI-shaderc-smoketest fail does not look like my fault.

JasperNV:
Added similar check for OpBranch as well.

JasperNV:
That test is trying to branch to a function. Before, the error message was "Block blah are referenced but not defined in function.". My code changes it to "Target is not OpLabel instruction".

It sounds like the cleanest fix is to just delete that test, then?

alan-baker:
Its ok just to update the error. That is still an invalid test and the validator should catch it.

JasperNV:
Pushed an update. AppVeyor tests are now passing.

alan-baker:
@JasperNV PTAL at the bot failures. Formatting and unused variables it look like. I'll reach them when you push another commit.

JasperNV:
Sorry about that (on Windows here, can't test the other builds). Pushed one more update.

PR #2309, opened 2019-01-21
Add WebGPU specific validation for Position BuiltIn decoration

zoddicus:
This PR adds in the specific checks required for WebGPU, enables
running the builtin checks in WebGPU, and refactors the existing
testing infrastructure to support testing the new checks.

### DIFF
--- a/test/val/val_builtins_test.cpp
+++ b/test/val/val_builtins_test.cpp
@@ -24,6 +24,7 @@
 #include <vector>
 
 #include "gmock/gmock.h"
+#include "source/spirv_target_env.h"
 #include "test/unit_spirv.h"
 #include "test/val/val_fixtures.h"
 
@@ -51,6 +52,9 @@ using ::testing::ValuesIn;
 
 using ValidateBuiltIns = spvtest::ValidateBase<bool>;
 using ValidateVulkanCombineBuiltInExecutionModelDataTypeResult =
+    spvtest::ValidateBase<std::tuple<const char*, const char*, const char*,
+                                     const char*, TestResult>>;
+using ValidateWebGPUCombineBuiltInExecutionModelDataTypeResult =
     spvtest::ValidateBase<std::tuple<const char*, const char*, const char*,
                                      const char*, TestResult>>;
 using ValidateVulkanCombineBuiltInArrayedVariable = spvtest::ValidateBase<
@@ -121,6 +125,13 @@ OpCapability Float64
 OpCapability Int64
 OpCapability MultiViewport
 OpCapability SampleRateShading
+)";
+}
+
+std::string GetWebGPUShaderCapabilities() {
+  return R"(
+OpCapability Shader
+OpCapability VulkanMemoryModelKHR
 )";
 }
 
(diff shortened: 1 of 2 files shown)


zoddicus:
PTAL

test/val/val_builtins_test.cpp
```diff
@@ -416,6 +601,30 @@ TEST_P(ValidateVulkanCombineBuiltInExecutionModelDataTypeResult, Variable) {
...
+  const char* const built_in = std::get<0>(GetParam());
+  const char* const execution_model = std::get<1>(GetParam());
+  const char* const storage_class = std::get<2>(GetParam());
+  const char* const data_type = std::get<3>(GetParam());
+  const TestResult& test_result = std::get<4>(GetParam());
+
+  CodeGenerator generator = GetVariableCodeGenerator(
+      SPV_ENV_WEBGPU_0, built_in, execution_model, storage_class, data_type);
+
+  //  fprintf(stderr, "%s\n", generator.Build().c_str());
```
alan-baker: Dead code

zoddicus: Done

zoddicus:
> Does WebGPU allow a level of arraying on variables?

I am really unclear what you are asking here. Can you give an example?

alan-baker:
> > Does WebGPU allow a level of arraying on variables?
> 
> I am really unclear what you are asking here. Can you give an example?
In Vulkan, a single level of arraying is allowed on variables, so you could have an array of vec4 and the variable gets decorated with Position. I added some Vulkan tests related to it so you may want to check those tests. I'm not sure what WebGPU's stance is on the matter. @dneto0 do you know if this has been decided?

zoddicus:
> > > Does WebGPU allow a level of arraying on variables?
> > 
> > 
> > I am really unclear what you are asking here. Can you give an example?
> > In Vulkan, a single level of arraying is allowed on variables, so you could have an array of vec4 and the variable gets decorated with Position. I added some Vulkan tests related to it so you may want to check those tests. I'm not sure what WebGPU's stance is on the matter. @dneto0 do you know if this has been decided?

Looking at the code for Vulkan for this decoration in particular, it appears that arrayed variables are only supported for an execution mode that isn't supported by WebGPU, TesselationControl. For Vertex, Vulkan doesn't support arrayed variables. I have extended the test coverage for WebGPU to make sure it has the same behaviour as Vertex.

test/val/val_builtins_test.cpp
```diff
@@ -1906,6 +1950,14 @@ INSTANTIATE_TEST_CASE_P(
                               "needs to be a 4-component 32-bit float vector",
                               "is not a float vector"))), );
 
+INSTANTIATE_TEST_CASE_P(
+    PositionArrayedF32Vec4Vertex, ValidateWebGPUCombineBuiltInArrayedVariable,
+    Combine(Values("Position"), Values("Vertex"), Values("Output"),
+            Values("%f32"),
```
alan-baker: Shouldn't this be `%f32vec4`?

zoddicus: Done

PR #2310, opened 2019-01-21
Layout checks for nested structures and descriptor arrays

### LINKED ISSUE #2303: Validator: Memory layout validation problems
The recursive checking of memory layouts in spirv-val appears to have two issues:

1. Arrays are only checked recursively if they are arrays of structs, not of arrays.
2. Recursive struct checking doesn't take the offset of the parent struct into account, which means that relaxed block layouts can end up with improperly straddling vectors. For example:
```
struct {
    vec2 f;
    struct {
        float g;
        vec2 h;
    } s;
}
```
This gives f an alignment of 8 and an offset of 0, and s has an alignment of 8 so it gets offset 8. That leaves h improperly straddling from 12 to 20, but the validator thinks everything is fine.

alan-baker:
Fixes #2303 

* Accumulate offset through nested structures
* Perform layout checks through descriptor arrays
* New tests

### DIFF
--- a/source/val/validate_decorations.cpp
+++ b/source/val/validate_decorations.cpp
@@ -379,6 +379,7 @@ bool IsAlignedTo(uint32_t offset, uint32_t alignment) {
 // or row major-ness.
 spv_result_t checkLayout(uint32_t struct_id, const char* storage_class_str,
                          const char* decoration_str, bool blockRules,
+                         uint32_t incoming_offset,
                          MemberConstraints& constraints,
                          ValidationState_t& vstate) {
   if (vstate.options()->skip_block_layout) return SPV_SUCCESS;
@@ -429,7 +430,8 @@ spv_result_t checkLayout(uint32_t struct_id, const char* storage_class_str,
         }
       }
     }
-    member_offsets.push_back(MemberOffsetPair{memberIdx, offset});
+    member_offsets.push_back(
+        MemberOffsetPair{memberIdx, incoming_offset + offset});
   }
   std::stable_sort(
       member_offsets.begin(), member_offsets.end(),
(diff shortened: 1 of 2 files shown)


source/val/validate_decorations.cpp
```diff
@@ -514,7 +516,7 @@ spv_result_t checkLayout(uint32_t struct_id, const char* storage_class_str,
       if (SpvOpTypeStruct == arrayInst->opcode() &&
           SPV_SUCCESS != (recursive_status = checkLayout(
                               typeId, storage_class_str, decoration_str,
-                              blockRules, constraints, vstate)))
+                              blockRules, offset, constraints, vstate)))
```
gnl21: It looks like this is still only going to check the offset for the first struct in an array. I think that it can still go wrong where the ArrayStride is not a multiple of 4. For example:
```
struct {
    struct {
        float a;        // Offset 0
        vec2 b;         // Offset 4
        float c, d;     // Offsets 12, 16
    } p[2];
};
```
The size of the struct `p` is 20, and the alignment is 8, so the ArrayStride will be 24. That means that `p[1].b` will still improperly straddle from 28-36.

gnl21:
I haven't had a chance to work through the array changes in this MR but I don't think they are addressing the issue that I was seeing. The issue that I was having was with a struct member that contained a multi-level array, rather than checking the layout of an array of descriptors. Eg.
```
struct {
    float a[4][3];
};
```
The array stride on the inner array was not being checked. Should that be addressed by this change?

alan-baker:
@gnl21 thanks for the feedback. I've update the change to catch those cases, but there are some corners that will still slip through (like improper straddles of spec constant sized arrays and multidimensional arrays). I'll have to think more about how to best cover all those cases, but hopefully this will catch most errors now.

dneto0:
FYI.  I just merged #2312 which was the first set of changes.   Please rebase to start with that merged code.

PR #2312, opened 2019-01-22
Fix layout checks for nested struct in relaxed layout; and descriptor arrays

dneto0:
This has a formatting fix for #2310


dneto0:
I already approved 2310.  I'm using this to run it by bots.

PR #2314, opened 2019-01-22
Add utility to generate a logging string for a given environment

### LINKED ISSUE #2313: Add general way to get environment specific string for error messages
A significant number of the rules for WebGPU are the same as those for Vulkan. Thus in the validator the only major change that needs to occur is adding to the env test an "or WebGPU". This does cause issues with the debugging messages, since often they currently have the string "Vulkan" in them. I have previously been building a local string that says "Vulkan" or "WebGPU" and inserting that as appropriate. This works fine for a one off things, but there are more and more of these instances occuring, so I am going to implemnent a more generic way to get a debugging string for a given env and use that instead

zoddicus:
Fixes #2313

### DIFF
--- a/source/val/validate_type.cpp
+++ b/source/val/validate_type.cpp
@@ -110,11 +110,10 @@ spv_result_t ValidateTypeArray(ValidationState_t& _, const Instruction* inst) {
   if ((spvIsVulkanEnv(_.context()->target_env) ||
        spvIsWebGPUEnv(_.context()->target_env)) &&
       element_type->opcode() == SpvOpTypeRuntimeArray) {
-    const char* env_text =
-        spvIsVulkanEnv(_.context()->target_env) ? "Vulkan" : "WebGPU";
     return _.diag(SPV_ERROR_INVALID_ID, inst)
            << "OpTypeArray Element Type <id> '" << _.getIdName(element_type_id)
-           << "' is not valid in " << env_text << " environment.";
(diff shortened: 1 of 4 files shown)


source/val/validate_type.cpp
```diff
@@ -110,11 +110,10 @@ spv_result_t ValidateTypeArray(ValidationState_t& _, const Instruction* inst) {
   if ((spvIsVulkanEnv(_.context()->target_env) ||
        spvIsWebGPUEnv(_.context()->target_env)) &&
       element_type->opcode() == SpvOpTypeRuntimeArray) {
-    const char* env_text =
-        spvIsVulkanEnv(_.context()->target_env) ? "Vulkan" : "WebGPU";
     return _.diag(SPV_ERROR_INVALID_ID, inst)
            << "OpTypeArray Element Type <id> '" << _.getIdName(element_type_id)
-           << "' is not valid in " << env_text << " environment.";
+           << "' is not valid in "
+           << spvLogStringForEnv(_.context()->target_env) << " environment.";
```
s-perron: The message seems like it might be a bit ackward.  "is not valid in WebGPU environment."  Should it be something like "is not valid in a WebGPU environment" or "is not valid in WebGPU environments"?

zoddicus: Prexisting, but I agree is it wrong. Corrected and tests updated.

PR #2315, opened 2019-01-22
More layout check fixes

alan-baker:
* check array strides for multidimensional arrays
* check layouts of structs in arrays for multiple indices
* new tests

Next steps of #2310


zmike:
Hi, I've just hit one of the error messages added in this patch, namely `member 0 contains an array with stride 4 not satisfying alignment to 16`. I couldn't find any reference to this in the spec, and the commit log is also not very descriptive. Can someone point me to some references for this?

Thanks!

alan-baker:
That requirement comes from the Vulkan spec.

PR #2316, opened 2019-01-23
Expand bindless bounds checking to runtime-sized descriptor arrays

greg-lunarg:
Runtime-sized descriptor arrays are allowed by VK_EXT_descriptor_indexing as the final binding in a descriptor set layout. Since these arrays are not encapsulated in a struct type, OpArrayLength cannot be used to determine their length. Rather, the lengths for these arrays will be passed in a "debug input" buffer by the validation layer. The binding and format of the input buffer is documented in instrument.hpp.

To facilitate coordination with the validation layers, bounds checking of runtime arrays is off by default and must be enabled explicitly when creating the bindless bounds instrumentation pass.


source/opt/instrument_pass.cpp
```diff
@@ -298,19 +311,55 @@ uint32_t InstrumentPass::GetOutputBufferBinding() {
...
+
+analysis::Type* InstrumentPass::GetUintRuntimeArrayType(
+    analysis::DecorationManager* deco_mgr, analysis::TypeManager* type_mgr) {
+  if (uint_rarr_ty_ == nullptr) {
+    analysis::Integer uint_ty(32, false);
+    analysis::Type* reg_uint_ty = type_mgr->GetRegisteredType(&uint_ty);
+    analysis::RuntimeArray uint_rarr_ty_tmp(reg_uint_ty);
+    uint_rarr_ty_ = type_mgr->GetRegisteredType(&uint_rarr_ty_tmp);
+    uint32_t uint_arr_ty_id = type_mgr->GetTypeInstruction(uint_rarr_ty_);
+    deco_mgr->AddDecorationVal(uint_arr_ty_id, SpvDecorationArrayStride, 4u);
```
s-perron: What happens if an OpRuntimeArray of uints already exists?  What if the existing on already has an array stride of 8?  We might want to test for that.

source/opt/instrument_pass.cpp
```diff
@@ -331,23 +380,40 @@ uint32_t InstrumentPass::GetOutputBufferId() {
...
+    // If not created yet, create one
+    analysis::DecorationManager* deco_mgr = get_decoration_mgr();
+    analysis::TypeManager* type_mgr = context()->get_type_mgr();
+    analysis::Type* reg_uint_rarr_ty =
+        GetUintRuntimeArrayType(deco_mgr, type_mgr);
+    analysis::Struct ibuf_ty({reg_uint_rarr_ty});
+    analysis::Type* reg_ibuf_ty = type_mgr->GetRegisteredType(&ibuf_ty);
+    uint32_t ibufTyId = type_mgr->GetTypeInstruction(reg_ibuf_ty);
+    deco_mgr->AddDecoration(ibufTyId, SpvDecorationBlock);
+    deco_mgr->AddMemberDecoration(ibufTyId, 0, SpvDecorationOffset, 0);
```
s-perron: I think the same issue can come up here.  You may get an existing type used for something else.  Then adding the decorations could cause a decoration to be applied to something to which it should not be applied.

source/opt/instrument_pass.h
```diff
@@ -76,7 +76,7 @@ class InstrumentPass : public Pass {
            IRContext::kAnalysisInstrToBlockMapping |
            IRContext::kAnalysisDecorations | IRContext::kAnalysisCombinators |
            IRContext::kAnalysisNameMap | IRContext::kAnalysisBuiltinVarId |
-           IRContext::kAnalysisConstants | IRContext::kAnalysisTypes;
+           IRContext::kAnalysisConstants;
```
s-perron: I'm curious how the type manager gets out of date.  You use the type manager a lot during the pass.  Are you sure that it is still valid enough during the pass that it will not give you incorrect results?

source/opt/instrument_pass.cpp
greg-lunarg: Good question. I did think about this. I actually tried to add the decorations to the type here but the current interface did not allow me to do it and so I decided, partly for expediency, partly out of lack of confidence in my understanding of the TypeManager, to go this route and just invalidate the TypeManager at the end.

I agree I probably should go back and fix the TypeManager so I can do it this way. But I do think what I am doing is not creating the possibility of invalid code. This is because when I create an *undecorated* RuntimeArray of uint32, I am not going to get a pre-existing RuntimeArray because any pre-existing RuntimeArray of uint32 will be *decorated*. Since I will not be getting a pre-existing type, I am free to decorate it.

I agree this logic is pretty esoteric. If you are open to it, I could document this a little better. If you are not open to it, I can try to wrestle the TypeManager into submission.

source/opt/instrument_pass.cpp
greg-lunarg: I believe my logic from the previous comment also applies to this type. Any pre-existing struct type with a runtime array will have to be a block and thus will have to be decorated with at least one Offset. So I will not get a pre-existing type here so I will be free to decorate it.

source/opt/instrument_pass.h
greg-lunarg: It is out of date because of the same issue we have been discussing above: it does not account for the decorations I have added to two different types I created.

source/opt/instrument_pass.cpp
s-perron: You can document it, and leave it as is.  Reference the spec when pointing out that each array type has to have a stride.

source/opt/instrument_pass.cpp
s-perron: Okay.  Again document it, and reference the spec for any assumption you make.

source/opt/instrument_pass.cpp
```diff
@@ -298,23 +311,83 @@ uint32_t InstrumentPass::GetOutputBufferBinding() {
...
+    analysis::DecorationManager* deco_mgr, analysis::TypeManager* type_mgr) {
+  analysis::Struct buf_ty(element_types);
+  analysis::Type* reg_buf_ty = type_mgr->GetRegisteredType(&buf_ty);
+  uint32_t bufTyId = type_mgr->GetTypeInstruction(reg_buf_ty);
+  // By the Vulkan spec, a pre-existing struct containing a RuntimeArray
+  // must be a block, and will therefore be decorated with Block. Therefore
+  // the undecorated type returned here will not be pre-existing and can
+  // safely be decorated. Since this type is now decorated, it is out of
+  // sync with the TypeManager and therefore the TypeManager must be
+  // invalidated after this pass.
```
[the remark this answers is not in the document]
greg-lunarg: See comment for prototype.

source/opt/instrument_pass.h
```diff
@@ -211,15 +218,33 @@ class InstrumentPass : public Pass {
...
   uint32_t GetVoidId();
 
-  // Return id for output buffer uint type
-  uint32_t GetOutputBufferUintPtrId();
+  // Return pointer to type for runtime array of uint
+  analysis::Type* GetUintRuntimeArrayType(analysis::DecorationManager* deco_mgr,
+                                          analysis::TypeManager* type_mgr);
+
+  // Return id of buffer type of |element_types|.
+  uint32_t GetBufferTypeId(
```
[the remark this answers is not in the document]
greg-lunarg: You make a good point. I was trying to just be a good citizen and encapsulate common code, but the extra work needed to make this routine safe sort of defeats the purpose of encapsulating it in the first place and we will just end up with a routine that is not all that useful anyways. So I am just going to inline it back again.

greg-lunarg:
My latest changes have now passed CI.

greg-lunarg:
Thanks Jeff. Is (2) core or part of VK_EXT_descriptor_indexing?

jeffbolznv:
2 is also new to descriptor_indexing, it falls under the runtimeDescriptorArray feature. 3 falls under runtimeDescriptorArray and descriptorBindingVariableDescriptorCount.

greg-lunarg:
OK. Previously I had designed this with the assumption that there would be only one RuntimeArray of descriptors per descriptor set. Now we know there can be multiple.

So, given the debug input buffer:

struct {
   uint32_t data[];
};

and a RuntimeArray of descriptors at (set=s, binding=b),

previously its length was at

data[ s + 1 ]

but now its length will be at

data[ data[ s + 1 ] + b]

greg-lunarg:
@s-perron It looks like the CI build is failing because it is picking up a version of googletest where INSTANTIATE_TEST_CASE_P is deprecated, and spirv-tools is making heavy use of it. Do you have any quick thoughts on what to do here or should I track this down?

s-perron:
It should already be fixed.  Can you rebase?

greg-lunarg:
Doh! Thanks.

greg-lunarg:
@s-perron Any ideas on why the shaderc smoketest is failing at hlsl_structbuffer_fn2_comp?

s-perron:
A change in the validator is causing a unit test in glslang to fail.  Not a problem for you.  If other test pass, we can still merge.

@jeffbolznv Have all of your concerns been addressed?

jeffbolznv:
I'm not really familiar with this code and haven't reviewed in detail, but it looks like the design has been updated to handle multiple runtime arrays per set, so I don't object.

Issue #2319, opened 2019-01-23
Refactor reducer


afd:
The reducer currently has a number of different passes that share (unchanged) functionality for controlling the reduction process, and provide their own (more or less completely different) functionality for finding reduction opportunities.

The functionality for finding doesn't really belong with the functionality for controlling the reduction process.

It would be better to structure this so that we have a number of reduction opportunity finder classes, each of which captures different finding behaviour, and a single reduction pass class that has a reference to a particular finder.

paulthomson:
Fixed in #2321

PR #2321, opened 2019-01-23
Refactored reducer so that the 'finding' functionality of a reduction…

afd:
… pass is separated from the generic functionality for tracking progress of a pass.  With this change, we now have a ReductionOpportunityFinder abstract class, with many subclasses for each type of reduction, and just one ReductionPass class, which has an associated finder.

This addresses #2319


afd:
> I don't have a problem the code itself. However, I wonder what the real advantage will be.

The two advantages I see are:

- simplifies writing tests by avoiding the need to create a reduction pass in order to evaluate reduction opportunities (before this change we had a not very nice subclass of ReductionPass in the tests)

- reduces the potential for the functionality associated with finding reduction opportunities getting entangled with the functionality for reduction pass bookkeeping

PR #2322, opened 2019-01-23
Request for comment: extracting useful functionality from block merger to be reused by reducer

afd:
I'm opening this PR to start a discussion on how best to expose bits of functionality provided by the optimizer for use in the reducer.

The reducer should be able to try merging blocks.  It doesn't want to merge *all* blocks, as doing so might make the property of interest go away, so we'd like to be able to try mergeable pairs.

Here I have extracted the two key bits of functionality - checking whether a block can be merged with its successor, and actually merging a block with its successor - and made them public static methods of the block merger.  The reducer could then invoke them directly.  The price for this is that the block merger has to call these static methods and pass in its IRContext.

What do folks think of this approach in general (as there will be many other cases where the reducer could re-use analyses provided by the optimizer)?  It certainly seems better than redundantly reimplementing the same functionality in the reducer, but makes the reducer appear to depend on particular optimization pass classes, when really it just depends on helper functions associated with those passes.

### DIFF
--- a/source/opt/block_merge_pass.cpp
+++ b/source/opt/block_merge_pass.cpp
@@ -24,104 +24,118 @@
 namespace spvtools {
 namespace opt {
 
+bool BlockMergePass::canMergeWithSuccessor(IRContext* context, BasicBlock *block) {
+  // Find block with single successor which has no other predecessors.
+  auto ii = block->end();
+  --ii;
+  Instruction* br = &*ii;
+  if (br->opcode() != SpvOpBranch) {
+    return false;
+  }
+
+  const uint32_t lab_id = br->GetSingleWordInOperand(0);
+  if (context->cfg()->preds(lab_id).size() != 1) {
+    return false;
+  }
+
+  bool pred_is_merge = IsMerge(context, block);
+  bool succ_is_merge = IsMerge(context, lab_id);
+  if (pred_is_merge && succ_is_merge) {
+    // Cannot merge two merges together.
+    return false;
+  }
+
+  Instruction* merge_inst = block->GetMergeInst();
+  const bool pred_is_header = IsHeader(block);
+  if (pred_is_header && lab_id != merge_inst->GetSingleWordInOperand(0u)) {
+    bool succ_is_header = IsHeader(context, lab_id);
+    if (pred_is_header && succ_is_header) {
+      // Cannot merge two headers together when the successor is not the merge
+      // block of the predecessor.
+      return false;
+    }
+
+    // If this is a header block and the successor is not its merge, we must
+    // be careful about which blocks we are willing to merge together.
+    // OpLoopMerge must be followed by a conditional or unconditional branch.
+    // The merge must be a loop merge because a selection merge cannot be
+    // followed by an unconditional branch.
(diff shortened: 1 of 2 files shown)


dnovillo:
Re-factoring optimization analysis helpers out of the optimization passes themselves is a great idea.  Not sure I would use static functions.  Though I can see that it's easier than having to instantiate the whole pass just to use its analysis.  Or was there another reason for that?

dneto0:
The approach looks fine to me.  Passing the context into static functions is fine.
Being static means the compiler will check errant dependencies on member data of the pass, so yay.

Sometimes an analysis could be a method on the object being queried.  e.g. the IsHeader could have been Block::IsStructuredHeader or similar. But that's not symmetric with the IsMerge, so I'd leave it as you have it. 

And now I get to complain about:   
- first letter capitalize per Google C++ style guide
- method comments
- Call out why those are static, i.e. for use independently of the block-pass transform.

afd:
> Re-factoring optimization analysis helpers out of the optimization passes themselves is a great idea. Not sure I would use static functions. Though I can see that it's easier than having to instantiate the whole pass just to use its analysis. Or was there another reason for that?

Instantiating the whole pass would just feel wrong - the pass has a bunch of book-keeping stuff that would be irrelevant; there could be side-effects collecting in the pass class if helper methods are invoked again and again.  It just seems cleaner to separate out the pure functionality that can be re-used elsewhere.

afd:
> The approach looks fine to me. Passing the context into static functions is fine.
> Being static means the compiler will check errant dependencies on member data of the pass, so yay.
> 
> Sometimes an analysis could be a method on the object being queried. e.g. the IsHeader could have been Block::IsStructuredHeader or similar. But that's not symmetric with the IsMerge, so I'd leave it as you have it.
> 
> And now I get to complain about:
> 
> * first letter capitalize per Google C++ style guide
> * method comments
> * Call out why those are static, i.e. for use independently of the block-pass transform.

Cool, thanks.  I'll leave this PR open a bit longer in case there are more thoughts from others.  Then I'll abandon it and do a proper one paying attention to comments and style.

dnovillo:
> > Re-factoring optimization analysis helpers out of the optimization passes themselves is a great idea. Not sure I would use static functions. Though I can see that it's easier than having to instantiate the whole pass just to use its analysis. Or was there another reason for that?
> 
> Instantiating the whole pass would just feel wrong - the pass has a bunch of book-keeping stuff that would be irrelevant; there could be side-effects collecting in the pass class if helper methods are invoked again and again. It just seems cleaner to separate out the pure functionality that can be re-used elsewhere.

Ah, yes. Good point.  I'd say, go for it.  It's a nice cleanup and can be done incrementally.  Thanks!

afd:
> I think I would prefer separating it out into two classes. This pass is simple because it has no state. There are no side tables or other data structures. Many passes are the same way. However, some other passes might need the analysis phase to keep track of other information that needs to persist. I don't like the static functions.
> 
> This wold turn into something like one of the analysies, but because we do not want to keep it alive in subsequent passes, not need to add it to the IRContext.
> 
> Otherwise I am okay with this type of change.

OK, great.  I'll abandon this PR and do another one shortly with these methods in a different file.

PR #2324, opened 2019-01-23
Add helper for 'is Vulkan or WebGPU'

### LINKED ISSUE #2323: Fold checks for Vulkan or WebGPU into a single helper function
Lots of places in the validator the same rule applies to both Vulkan & WebGPU, so you have to test for either being the target env. The standard way of writting this is:

`
    if (spvIsVulkanEnv(_.context()->target_env) || spvIsWebGPUEnv(_.context()->target_env)) {
        ...
`

Using a helper it can be shortened to:

`
    if (spvIsVulkanOrWebGPUEnv(_.context()->target_env)) {
        ...
`

zoddicus:
Fixes #2323

### DIFF
--- a/source/spirv_target_env.cpp
+++ b/source/spirv_target_env.cpp
@@ -249,6 +249,10 @@ bool spvIsWebGPUEnv(spv_target_env env) {
   return false;
 }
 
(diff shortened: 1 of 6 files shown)


zoddicus:
PTAL

PR #2325, opened 2019-01-23
Extracted block merging functionality into its own utility file

afd:
The functionality to detect when it is safe to merge blocks, and to actually merge blocks, has been extracted into stand-alone methods in a new utility file.

The motivation for this is to make this functionality available to other analyses and passes, and in particular to the reducer.

I have opted for methods within a namespace, rather than having a class-with-only-static-methods, but do not feel strongly about this choice.


afd:
The ndk build is failing to link libspirvtools_test.so for mips.  It links fine for a bunch of other targets:

[arm64-v8a] SharedLibrary  : libspirvtools_test.so
[x86_64] SharedLibrary  : libspirvtools_test.so
[armeabi-v7a] SharedLibrary  : libspirvtools_test.so
[mips64] SharedLibrary  : libspirvtools_test.so
[armeabi] SharedLibrary  : libspirvtools_test.so
[x86] SharedLibrary  : libspirvtools_test.so
[mips] SharedLibrary  : libspirvtools_test.so
../source/opt/block_merge_pass.cpp:31: error: undefined reference to 'spvtools::opt::blockmergeutil::CanMergeWithSuccessor(spvtools::opt::IRContext*, spvtools::opt::BasicBlock*)'
../source/opt/block_merge_pass.cpp:32: error: undefined reference to 'spvtools::opt::blockmergeutil::MergeWithSuccessor(spvtools::opt::IRContext*, spvtools::opt::Function*, spvtools::opt::UptrVectorIterator<spvtools::opt::BasicBlock, false>)'
collect2: error: ld returned 1 exit status

I wonder why this could be.

source/opt/block_merge_util.cpp
```diff
@@ -0,0 +1,143 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "block_merge_util.h"
+
+namespace spvtools {
+namespace opt {
+namespace blockmergeutil {
+
+// Returns true if |block| contains a merge instruction.
```
s-perron: We may have done this wrong in the previous version.  If this function is meant to be used in this file only, then it should be in an anonymous namespace.  Same for other functions below.

afd: Thanks, fixed.

PR #2326, opened 2019-01-24
Add SpirvTools::IsValid().

fjhenigman:
Add a method to determine if a SpirvTools object was successfully
constructed and can be used.  It might not be depending on the parameter
to the constructor.
This is something a fuzzer wants to know before trying to use an
SpirvTools object constructed with a fuzzed parameter.


dnovillo:
You'll need to run clang-format on this PR (https://source.cloud.google.com/results/invocations/76bf43f9-3b74-4d6f-9161-3deab51c17fa/targets/graphics_shader_compiler%2Fspirv-tools%2Fcheck-format%2Fpresubmit_check_format/log)

PR #2339, opened 2019-01-27
Fixes #2338.  Added functionality to remove OpPhi instructions (replacing their uses) when merging blocks

### LINKED ISSUE #2338: merge-blocks should not merge block A with block B if B starts with OpPhi
Currently the --merge-blocks pass turns:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpSource ESSL 310
               OpName %4 "main"
               OpName %8 "x"
               OpName %10 "y"
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeInt 32 1
          %7 = OpTypePointer Function %6
          %9 = OpConstant %6 1
          %4 = OpFunction %2 None %3
          %5 = OpLabel
          %8 = OpVariable %7 Function
         %10 = OpVariable %7 Function
               OpStore %8 %9
         %11 = OpLoad %6 %8
               OpBranch %12
         %12 = OpLabel
         %13 = OpPhi %6 %11 %5
               OpStore %10 %13
               OpReturn
               OpFunctionEnd
```

into:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpSource ESSL 310
               OpName %4 "main"
               OpName %8 "x"
               OpName %10 "y"
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeInt 32 1
          %7 = OpTypePointer Function %6
          %9 = OpConstant %6 1
          %4 = OpFunction %2 None %3
          %5 = OpLabel
          %8 = OpVariable %7 Function
         %10 = OpVariable %7 Function
               OpStore %8 %9
         %11 = OpLoad %6 %8
         %13 = OpPhi %6 %11 %5
               OpStore %10 %13
               OpReturn
               OpFunctionEnd
```

I.e., it merges blocks %5 and %12.  But block %12 starts with OpPhi, so this leads to invalid SPIR-V.

afd:


afd:
> We can still merge these nodes. It is just that the OpPhi needs to be replaced by and OpCopyObject, or we do a call of replace all uses and delete the instruction. Since this affects the optimization pass as well and the change it not too difficult, I would prefer that we handle the case instead of punting on it.

OK, I have updated the PR to do the latter - replace all uses and delete the instruction, plus have added some tests for this.

source/opt/block_merge_util.cpp
```diff
@@ -49,14 +49,21 @@ bool IsMerge(IRContext* context, BasicBlock* block) {
...
-bool StartsWithOpPhi(IRContext* context, uint32_t id) {
-  bool result = false;
-  context->cfg()->block(id)->ForEachPhiInst([&result](Instruction* /*unused*/) {
-    // If any OpPhi is found then the block must start with OpPhi.
-    result = true;
+// Requires that any OpPhi in |successor| has |predecessor| as its sole parent.
+// Replaces all uses of such OpPhis with the id associated with |predecessor|,
+// and removes the OpPhi instructions.
+void EliminateOpPhiInstructions(IRContext* context, BasicBlock* successor,
+                                BasicBlock* predecessor) {
```
s-perron: You do not actually need `predecessor` in this function.  It is only used as a check to make sure that the OpPhi instruction was a valid OpPhi.  I would suggest removing it.  We check that there number of in operands is 2 and just assume that the only predecessor was the correct one.

If not, it would be invalid spirv and it does not matter what we generate.

PR #2343, opened 2019-01-28
Move CodeGenerator to be accessible by other validation tests

### LINKED ISSUE #2342: Refactor CodeGenerator from val_builtins_test.cpp to be generically usable
I am going to bove this test utility class out to where it can be access by other test files, since I am planning on writting new parameterized tests that would greatly benefit from it.

zoddicus:
Fixes #2342

### DIFF
--- a/test/val/CMakeLists.txt
+++ b/test/val/CMakeLists.txt
@@ -15,6 +15,7 @@
 set(VAL_TEST_COMMON_SRCS
   ${CMAKE_CURRENT_SOURCE_DIR}/../test_fixture.h
   ${CMAKE_CURRENT_SOURCE_DIR}/../unit_spirv.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/val_code_generator.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/val_code_generator.h
   ${CMAKE_CURRENT_SOURCE_DIR}/val_fixtures.h
 )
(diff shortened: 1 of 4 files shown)


test/val/val_code_generator.cpp
```diff
@@ -0,0 +1,222 @@
+// Copyright (c) 2018 Google LLC.
```
dj2: 2019

test/val/val_code_generator.h
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2018 Google LLC.
```
dj2: 2019

test/val/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@
 set(VAL_TEST_COMMON_SRCS
   ${CMAKE_CURRENT_SOURCE_DIR}/../test_fixture.h
   ${CMAKE_CURRENT_SOURCE_DIR}/../unit_spirv.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/val_code_generator.h
```
dj2: Where is the .cpp file listed in the CMakeLists file?

test/val/val_code_generator.h
zoddicus: Done

test/val/val_code_generator.cpp
zoddicus: Done

test/val/CMakeLists.txt
zoddicus: Done

PR #2344, opened 2019-01-28
Add source/opt/block_merge_util.* to BUILD.gn

dneto0:


dneto0:
I have not tested this.  Please satisfy yourself that this is sufficent and safe.

PR #2345, opened 2019-01-29
Use DEPS file for checking out third party code.

dj2:
This CL simplifies the DEPS file in SPIRV-Tools and removes all the Chrome stuff. The git-sync-deps script is also added which allows using the DEPS file and script to handle all the bot updates. The README.md is updated to make this the preferred way of syncing the external dependencies.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -45,10 +45,7 @@ install:
   - set PATH=C:\ninja;%PATH%
 
 before_build:
-  - git clone --depth=1 https://github.com/KhronosGroup/SPIRV-Headers.git external/spirv-headers
-  - git clone --depth=1 https://github.com/google/googletest.git external/googletest
-  - git clone --depth=1 https://github.com/google/effcee.git external/effcee
-  - git clone --depth=1 https://github.com/google/re2.git external/re2
+  - python utils\git-sync-deps
   # Set path and environment variables for the current Visual Studio version
   - if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2013" (call "C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\vcvarsall.bat" x86_amd64)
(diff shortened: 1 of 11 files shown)


s-perron:
I don't think I like this.  We are not living at head anymore.  The smoke test was specifically meant to be an early warning that the latest glslang does not work with the latest spirv-tools.  That has now been pegged to a specific version of glslang.

I don't like the idea of the builds using the DEPS file.  We want to always test with the latest.  This way we know as soon as possible that something is broken.  It has not created much of a problem for us because we do not have many dependencies.

I also don't like the responsibility of updating the DEPS file.  It is not clear who will do that and when.

dj2:
I think I mis-understood the smoketest. I thought, in this case, it was checking that current shaderc builds with the version of spirv-tools being committed? Is it checking both shaderc and glslang?

We've run into a few issues where we've had to peg re2 or googletest to specific revisions. The main benefit I see to the DEPS file is that the build is repeatable. You know all the bots built at the same version and you know what those revisions are. With the current system, I have no idea when the last green spriv-tools and spirv-headers as as spirv-headers could have been rolled but no bots triggered in spirv-tools.

The DEPS get rolled as needed. If you want to pull in new spirv-headers features into spirv-tools, roll deps. For googletest, re2 and effcee, we roll as needed.

dj2:
Speaking of, the deps for re2 and effcee need to be rolled in the DEPS file to make windows pass, heh.

fjhenigman:
I can't see why we wouldn't want this.  No surprise breakages to be investigated.  Green bots so we can land CLs.  Known good configurations for users.

fjhenigman:
If it's really important to catch breakage as early as possible, we could add a bot that builds with everything at ToT.  Then whoever cares can monitor that bot.

s-perron:
At the very least we have to leave the smoketest with the latest from every component.  If we do that, then that test can catch any failures early, and we will all see it.  It is not hidden away where nobody sees it like the asan test for shaderc were.

dj2:
Closing this for now. We can re-open when we investigate getting auto-rolls up for the dependencies.

PR #2346, opened 2019-01-29
Add whitelist for decorations in WebGPU

### LINKED ISSUE #2273: Add whitelist for decorations to WebGPU validation
Add a whitelist for the values being passed into OpDecoration and OpMemberDecoration, for the WebGPU env from the draft [spec](https://github.com/gpuweb/spirv-execution-env/blob/master/execution-env.md):

>Only the following decorations are valid:
>- SpecId
>- Block
>- RowMajor
>- ColMajor
>- ArrayStride (also see "Layouts" below)
>- MatrixStride (also see "Layouts" below)
>- Builtin (also see "Built-In Variables" below)
>- NoPerspective, only in the Input storage class and the Fragment execution model
>- Flat, only in the Input storage class and the Fragment execution model
>- Centroid
>- Restrict
>- Aliased
>- NonWritable
>- NonReadable
>- Uniform
>- Location, required for all Input and Output, see "Interface" section below
>- Component
>- Index
>- Binding
>- DescriptorSet
>- Offset (also see "Layouts" below)
>- NoContraction

zoddicus:
Fixes #2273

### DIFF
--- a/source/val/validate_annotation.cpp
+++ b/source/val/validate_annotation.cpp
@@ -23,6 +23,170 @@ namespace spvtools {
 namespace val {
 namespace {
 
+bool IsValidWebGPUDecoration(uint32_t decoration) {
+  switch (decoration) {
+    case SpvDecorationSpecId:
+    case SpvDecorationBlock:
+    case SpvDecorationRowMajor:
+    case SpvDecorationColMajor:
+    case SpvDecorationArrayStride:
+    case SpvDecorationMatrixStride:
+    case SpvDecorationBuiltIn:
+    case SpvDecorationNoPerspective:
+    case SpvDecorationFlat:
+    case SpvDecorationCentroid:
+    case SpvDecorationRestrict:
+    case SpvDecorationAliased:
+    case SpvDecorationNonWritable:
+    case SpvDecorationNonReadable:
+    case SpvDecorationUniform:
+    case SpvDecorationLocation:
+    case SpvDecorationComponent:
+    case SpvDecorationIndex:
+    case SpvDecorationBinding:
+    case SpvDecorationDescriptorSet:
+    case SpvDecorationOffset:
+    case SpvDecorationNoContraction:
+      return true;
+    default:
+      return false;
+  }
+}
+
+std::string LogStringForDecoration(uint32_t decoration) {
+  switch (decoration) {
+    case SpvDecorationRelaxedPrecision:
+      return "RelaxedPrecision";
+    case SpvDecorationSpecId:
+      return "SpecId";
+    case SpvDecorationBlock:
+      return "Block";
+    case SpvDecorationBufferBlock:
+      return "BufferBlock";
+    case SpvDecorationRowMajor:
+      return "RowMajor";
+    case SpvDecorationColMajor:
+      return "ColMajor";
+    case SpvDecorationArrayStride:
+      return "ArrayStride";
+    case SpvDecorationMatrixStride:
+      return "MatrixStride";
+    case SpvDecorationGLSLShared:
+      return "GLSLShared";
+    case SpvDecorationGLSLPacked:
+      return "GLSLPacked";
+    case SpvDecorationCPacked:
+      return "CPacked";
+    case SpvDecorationBuiltIn:
+      return "BuiltIn";
+    case SpvDecorationNoPerspective:
+      return "NoPerspective";
+    case SpvDecorationFlat:
+      return "Flat";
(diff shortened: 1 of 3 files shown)


zoddicus:
PTAL

source/val/validate_annotation.cpp
```diff
@@ -23,6 +23,170 @@ namespace spvtools {
...
+      return "NonUniformEXT";
+    case SpvDecorationRestrictPointerEXT:
+      return "RestrictPointerEXT";
+    case SpvDecorationAliasedPointerEXT:
+      return "AliasedPointerEXT";
+    case SpvDecorationHlslCounterBufferGOOGLE:
+      return "HlslCounterBufferGOOGLE";
+    case SpvDecorationHlslSemanticGOOGLE:
+      return "HlslSemanticGOOGLE";
+    case SpvDecorationMax:
```
alan-baker: Since Max is not a real decoration, I would prefer this to return unknown.

test/val/val_memory_test.cpp
```diff
@@ -2119,9 +2119,9 @@ OpFunctionEnd
   EXPECT_EQ(SPV_ERROR_INVALID_ID, ValidateInstructions(SPV_ENV_WEBGPU_0));
   EXPECT_THAT(
       getDiagnosticString(),
-      HasSubstr("For WebGPU, OpTypeStruct variables containing "
-                "OpTypeRuntimeArray must have storage class of StorageBuffer\n "
-                " %6 = OpVariable %_ptr_Uniform__struct_3 Uniform\n"));
+      HasSubstr(
```
alan-baker: This seems like the wrong update to make. I'd suggest instead changing `%struct_t` to be a Block in StorageBuffer instead.

test/val/val_decoration_test.cpp
```diff
@@ -20,16 +20,34 @@
...
+
+struct TestResult {
+  TestResult(spv_result_t in_validation_result = SPV_SUCCESS,
+             const char* in_pre_error_str = nullptr,
+             const char* in_post_error_str = nullptr)
+      : validation_result(in_validation_result),
+        pre_error_str(in_pre_error_str),
+        post_error_str(in_post_error_str) {}
+  spv_result_t validation_result;
+  const char* pre_error_str;
```
alan-baker: Is there a good reason to distinguish these strings? No test uses both. They also result in the same check getting performed.

I'd also prefer using std::string over char*, but its just a preference.

test/val/val_decoration_test.cpp
```diff
@@ -20,16 +20,30 @@
...
 namespace {
 
+using ::testing::Combine;
 using ::testing::Eq;
 using ::testing::HasSubstr;
+using ::testing::Values;
+
+struct TestResult {
+  TestResult(spv_result_t in_validation_result = SPV_SUCCESS,
+             const std::string in_error_str = "")
```
alan-baker: Nit: this should be a const reference.

test/val/val_decoration_test.cpp
```diff
@@ -20,16 +20,30 @@
...
+      : validation_result(in_validation_result), error_str(in_error_str) {}
+  spv_result_t validation_result;
+  const std::string error_str;
+  const char* post_error_str;
```
alan-baker: Unused member.

test/val/val_decoration_test.cpp
```diff
@@ -5741,6 +5755,67 @@ TEST_F(ValidateDecorations, NonWritableVarFunctionBad) {
...
+
+  EntryPoint entry_point;
+  entry_point.name = "main";
+  entry_point.execution_model = "Vertex";
+  generator.entry_points_.push_back(std::move(entry_point));
+
+  CompileSuccessfully(generator.Build(), SPV_ENV_WEBGPU_0);
+  ASSERT_EQ(test_result.validation_result,
+            ValidateInstructions(SPV_ENV_WEBGPU_0));
+  if (test_result.error_str != "") {
```
alan-baker: Nit: `!test_result.error_str.empty()`

test/val/val_decoration_test.cpp
zoddicus: Done

test/val/val_decoration_test.cpp
zoddicus: Done

test/val/val_decoration_test.cpp
zoddicus: Done

PR #2348, opened 2019-01-29
Add flatten-decorations flag to webgpu-mode flags

### LINKED ISSUE #2272: WebGPU transform: Should flatten decorations
The WebGPU env spec bans OpDecorationGroup, so the optimization recipe for transforming  Vulkan to WebGPU should run the --flatten-decorations flag.

zoddicus:
Fixes #2272


zoddicus:
PTAL

zoddicus:
Added a test to make sure all of the passes we are expecting are being registered, and another that verifies that the flatten-decoration flag is running.

zoddicus:
Squashed history to eliminate the commit from my chromium.org account and deleted the now invalid CLA warning.

PR #2358, opened 2019-02-06
Allow arrayed storage images for NonWritable decoration

### LINKED ISSUE #2354: Validator incorrectly complains about NonWritable decoration on storage image
From https://github.com/KhronosGroup/glslang/issues/1689:

The SPIR-V decoration targets a pointer to an array of storage images.

The SPIR-V specification says NonWritable can apply to storage images. Maybe there were too many levels of indirection to easily see it, but it seems like this should be allowed. Transferring to SPIR-V tools.

alan-baker:
Fixes #2354

* Storage image pointer registration allows optional level of arraying
* Added a test

FYI @alegal-arm

### DIFF
--- a/source/val/validate_type.cpp
+++ b/source/val/validate_type.cpp
@@ -262,17 +262,22 @@ spv_result_t ValidateTypeStruct(ValidationState_t& _, const Instruction* inst) {
 
 spv_result_t ValidateTypePointer(ValidationState_t& _,
                                  const Instruction* inst) {
-  const auto type_id = inst->GetOperandAs<uint32_t>(2);
-  const auto type = _.FindDef(type_id);
+  auto type_id = inst->GetOperandAs<uint32_t>(2);
+  auto type = _.FindDef(type_id);
   if (!type || !spvOpcodeGeneratesType(type->opcode())) {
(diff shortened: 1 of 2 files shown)


source/val/validate_type.cpp
```diff
@@ -262,17 +262,22 @@ spv_result_t ValidateTypeStruct(ValidationState_t& _, const Instruction* inst) {
...
            << "' is not a type.";
   }
   // See if this points to a storage image.
-  if (type->opcode() == SpvOpTypeImage) {
-    const auto storage_class = inst->GetOperandAs<uint32_t>(1);
-    if (storage_class == SpvStorageClassUniformConstant) {
+  const auto storage_class = inst->GetOperandAs<SpvStorageClass>(1);
+  if (storage_class == SpvStorageClassUniformConstant) {
+    // Unpack an optional level of arraying.
+    if (type->opcode() == SpvOpTypeArray) {
```
dneto0: Also need to allow for RuntimeArray here.

E.g. see uses of Capability RuntimeDescriptorArrayEXT in Vulkan, corresponding to runtimeDescriptorArray feature bit.

alan-baker: Yes, you're right. I'll do that.

PR #2362, opened 2019-02-07
Make spvParseTargetEnv public

dj2:
This CL moves the method to parse the SPIRV environment into the public headers. This will allow other applications to re-use the same parsing logic.


dj2:
Red bot should be fixed with CL to update glslang to use KhronosGroup.

PR #2365, opened 2019-02-07
Add better error message to ValidateInstructions

### LINKED ISSUE #2364: Make ValidateInstructions throw a useful error if you have failed to call CompileSuccessfully
Currently not calling CompileSuccessfully first will cause a SIGSEGV within the validation code the first time it tries access the binary input, because that will be null. This is somewhat annoying to root cause, since it isn't actually an issue with the validator, but with the test code. 

I am going to add an assert and error message in ValidateInstructions to catch this higher up the call stack and make it clear what the remedy likely is.

zoddicus:
Fixes #2364


zoddicus:
Pulling head fixed the previously failing smoketest and check-format test

PR #2370, opened 2019-02-11
Fix merge blocks opportunity to check if still enabled

### LINKED ISSUE #2369: reducer: merge_blocks pass fails to check for invalidation of merge block opportunity
`MergeBlocksReductionOpportunity` assumes it can never be invalidated/disabled by another `MergeBlocksReductionOpportunity` being applied, but this is not true. A loop header and a return block cannot be merged, and the return block might not start out as a return block; it can become a return block due to another merge.

paulthomson:
Fixes #2369. Added tests.

### DIFF
--- a/source/reduce/merge_blocks_reduction_opportunity.cpp
+++ b/source/reduce/merge_blocks_reduction_opportunity.cpp
@@ -34,9 +34,13 @@ MergeBlocksReductionOpportunity::MergeBlocksReductionOpportunity(
 }
 
 bool MergeBlocksReductionOpportunity::PreconditionHolds() {
-  // By construction, it is not possible for the merging of A->B to disable the
-  // merging of C->D, even when B and C are the same block.
-  return true;
+  const auto predecessors = context_->cfg()->preds(successor_block_->id());
+  assert(1 == predecessors.size() &&
+         "For a successor to be merged into its predecessor, exactly one "
+         "predecessor must be present.");
+  const uint32_t predecessor_id = predecessors[0];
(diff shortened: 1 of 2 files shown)


source/reduce/merge_blocks_reduction_opportunity.cpp
```diff
@@ -34,9 +34,13 @@ MergeBlocksReductionOpportunity::MergeBlocksReductionOpportunity(
 }
 
 bool MergeBlocksReductionOpportunity::PreconditionHolds() {
-  // By construction, it is not possible for the merging of A->B to disable the
-  // merging of C->D, even when B and C are the same block.
-  return true;
```
afd: It would be good to have a comment explaining that there do exist scenarios where one merge can disable another merge.  Giving one example would help.

test/reduce/merge_blocks_test.cpp
```diff
@@ -508,6 +508,144 @@ TEST(MergeBlocksReductionPassTest, MergeWithOpPhi) {
   CheckEqual(env, after, context.get());
 }
 
+void MergeBlocksReductionPassTest_LoopReturn_Helper(bool reverse) {
```
afd: Great that we have this example in the tests.  Still, a comment in the precondition checking code would be good.

source/reduce/merge_blocks_reduction_opportunity.cpp
paulthomson: Sounds good. Done.

Issue #1587, opened 2018-05-30
validate: (data) Block can't appear within a Block


dneto0:
Explicitly say (data) Block cannot be nested in another Block
SPIR-V internal issue 295.

sarahM0:
As mentioned in SPIR-V internal issue 295: To be more explicit...  Under
2.16.2. Validation Rules for Shader Capabilities
- Decorations
Add one more subbullet: 
Block or BufferBlock cannot decorate a structure type that is nested at any level inside another structure type decorated with Block or BufferBlock.

PR #2371, opened 2019-02-11
Validate: (data) Block can't appear within a Block

sarahM0:
To fix: https://github.com/KhronosGroup/SPIRV-Tools/issues/1587
Please review.

### DIFF
--- a/source/opt/block_merge_util.cpp
+++ b/source/opt/block_merge_util.cpp
@@ -86,6 +86,11 @@ bool CanMergeWithSuccessor(IRContext* context, BasicBlock* block) {
     return false;
   }
 
+  // Don't bother trying to merge unreachable blocks.
+  if (auto dominators = context->GetDominatorAnalysis(block->GetParent())) {
+    if (!dominators->IsReachable(block)) return false;
+  }
+
   Instruction* merge_inst = block->GetMergeInst();
   const bool pred_is_header = IsHeader(block);
   if (pred_is_header && lab_id != merge_inst->GetSingleWordInOperand(0u)) {
--- a/source/val/validate_instruction.cpp
+++ b/source/val/validate_instruction.cpp
@@ -361,7 +361,6 @@ spv_result_t LimitCheckStruct(ValidationState_t& _, const Instruction* inst) {
   // The nesting depth of a struct is 1+(largest depth of any member).
   // Scalars are at depth 0.
   uint32_t max_member_depth = 0;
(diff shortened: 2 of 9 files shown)


source/val/validate_instruction.cpp
```diff
@@ -379,6 +385,14 @@ spv_result_t LimitCheckStruct(ValidationState_t& _, const Instruction* inst) {
            << "Structure Nesting Depth may not be larger than " << depth_limit
            << ". Found " << cur_depth << ".";
   }
+
+  _.SetHasNestedBlockOrBufferBlockStruct(inst->id(), has_nested_blockOrBufferBlock_struct);
+  if(_.HasNestedBlockOrBufferBlockStruct(inst->id()) && _.HasDecoration(inst->id(), SpvDecorationBlock)){
```
sarahM0: This always returns false. Could tell me what am I doing wrong?

alan-baker: I think the decorations have not been registered at this point. Your check should be performed later in the flow.

test/val/val_decoration_test.cpp
```diff
@@ -2303,18 +2341,18 @@ TEST_F(ValidateDecorations, MultiplePushConstantsSingleEntryPointGood) {
                 OpMemoryModel Logical GLSL450
                 OpEntryPoint Fragment %1 "main"
                 OpExecutionMode %1 OriginUpperLeft
-
```
sarahM0: Do I need to be worried about these auto indentations which are done by my editor(Sublime)?

alan-baker: Not as long as the tests keep passing and the format checker is ok.

source/val/validate_instruction.cpp
sarahM0: That might be it. In that case I think we need to update  _HasDecoration_ definition to return an error, saying it doesn't know yet. Also I couldn't fully understand how this function works. I appreciate it if you could go through it with me when you were in office.

source/opt/block_merge_util.cpp
```diff
@@ -86,6 +86,11 @@ bool CanMergeWithSuccessor(IRContext* context, BasicBlock* block) {
     return false;
   }
 
+  // Don't bother trying to merge unreachable blocks.
```
alan-baker: This looks like a bad rebase.

sarahM0:
closed because of a bad rebase.

Issue #1298, opened 2018-02-14
It is invalid to apply both Restrict and Aliased to the same <id>.


dneto0:
From 2.18.2

PR #2377, opened 2019-02-12
It is invalid to apply both Restrict and Aliased to the same <id>

sarahM0:
To fix: https://github.com/KhronosGroup/SPIRV-Tools/issues/1298
Work in progress, do not review

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -547,6 +547,7 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
            << "PhysicalStorageBufferEXT must not be used with OpVariable.";
   }
 
+// dummy change
   auto pointee_base = pointee;
   while (pointee_base->opcode() == SpvOpTypeArray) {
     pointee_base = _.FindDef(pointee_base->GetOperandAs<uint32_t>(1u));
@@ -573,6 +574,7 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
       }
     }
   }
+// dummy change
 
   // Vulkan specific validation rules for OpTypeRuntimeArray
   if (spvIsVulkanEnv(_.context()->target_env)) {
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -399,6 +399,18 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
     }
   }
 
+  // 2.18.2. It is invalid to apply both Restrict and Aliased to the same <id>.
+  bool foundAliased2 =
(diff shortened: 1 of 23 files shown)


source/val/validate_memory.cpp
```diff
@@ -547,6 +547,7 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
            << "PhysicalStorageBufferEXT must not be used with OpVariable.";
```
sarahM0: Does _EXT_ mean decoration?
What is _PhysicalStorageBufferEXT_? is it the same as StorageBuffer, Shared externally, readable and writable, visible across all functions in all invocations in all work groups. Graphics storage buffers (buffer blocks)?
I think to fix this bug I need to simply remove the three nested if-conditions that is being applied to the code in lines 559 - 572, update the error message, tests and ... . Does it sound reasonable?

source/val/validate_memory.cpp
```diff
@@ -399,6 +399,18 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
     }
   }
 
+  // 2.18.2. It is invalid to apply both Restrict and Aliased to the same <id>.
```
sarahM0: I know this is a wrong place for this piece of code. It needs to be somewhere that we are validating IDs. I found this _spv_result_t spvValidateIDs(const spv_instruction_t* pInstructions,_ in validate.h. Where is the definition?

source/val/validate_memory.cpp
alan-baker: EXT is the suffix generally used with tokens associated with SPV_EXT_ extensions. In this case, the extension is SPV_EXT_physical_storage_buffer. PhysicalStorageBufferEXT is a storage buffer that has a physical64 addressing model.

source/val/validate_memory.cpp
alan-baker: I think this is left over from a refactor. That method was split up into the various ValidateMemory, etc methods.

sarahM0: Do I remove it or we'd like to keep there?
Any suggestion about where to place this code?

alan-baker: Its dead, go ahead and remove it. 

I would consider having a new function validate_decorations. If you see Aliased, then check other decorations of that id and issue an error if Restrict is one of them.

Alternative: Aliased and Restrict can only be applied to Memory Object Declarations (OpVariable or OpFunctionParameter), so you could add checks in ValidateVariable and ValidateFunctionParameter. This is less ideal since it is probably more code duplication.

sarahM0:
Closing this to implement a new approach using: CheckDecorationsCompatibility

PR #2378, opened 2019-02-12
Move usage detection to after all instructions are registered

### LINKED ISSUE #2373: Validator: unreachable loop with unreachable continue fails validation
The following SPIRV contains an unreachable loop with an unreachable continue target and an unreachable merge block. It fails validation:

```
error: line 17: Back-edges (11 -> 13) can only be formed between a block and a loop header.
  %11 = OpLabel
```

```asm
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %2 "main"
               OpExecutionMode %2 OriginUpperLeft
               OpSource ESSL 310
               OpName %2 "main"
          %3 = OpTypeVoid
          %4 = OpTypeFunction %3
          %5 = OpTypeInt 32 1
          %6 = OpTypePointer Function %5
          %7 = OpTypeBool
          %8 = OpConstantFalse %7

          %2 = OpFunction %3 None %4

          %9 = OpLabel
               OpBranch %10

         %11 = OpLabel                   ; header
               OpLoopMerge %12 %13 None
               OpBranch %14
         %13 = OpLabel                   ; continue
               OpBranch %11
         %14 = OpLabel
               OpReturn
         %12 = OpLabel                   ; merge
               OpBranch %10

         %10 = OpLabel
               OpReturn
               OpFunctionEnd
```

zoddicus:
This is required to properly handle uses of forward declared ids. Since forward
declared ids were not being properly covered by the validator this uncovered a
bunch of small issues that needed to be resolved to get tests passing again.

This needs to be in its own loop because some of the checks in the later loop depend on all of the usage information being calculated.

Fixes #2373

### DIFF
--- a/source/val/validate.cpp
+++ b/source/val/validate.cpp
@@ -398,6 +398,14 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
   // Catch undefined forward references before performing further checks.
   if (auto error = ValidateForwardDecls(*vstate)) return error;
 
+  // ID usage needs be handled in its own iteration of the instructions,
+  // between the two others. It depends on the first loop to have been
+  // finished, so that all instructions have been registered. And the following
+  // loop depends on all of the usage data being populated. Thus it cannot live
+  // in either of those iterations.
+  // It should also live after the forward declaration check, since it will
+  // have problems with missing forward declarations, but give less useful error
+  // messages.
   for (size_t i = 0; i < vstate->ordered_instructions().size(); ++i) {
     auto& instruction = vstate->ordered_instructions()[i];
     if (auto error = UpdateIdUse(*vstate, &instruction)) return error;
--- a/source/val/validate.cpp
+++ b/source/val/validate.cpp
@@ -385,7 +385,6 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
       Instruction* inst = const_cast<Instruction*>(&instruction);
       vstate->RegisterInstruction(inst);
     }
-    if (auto error = UpdateIdUse(*vstate, &instruction)) return error;
   }
 
   if (!vstate->has_memory_model_specified())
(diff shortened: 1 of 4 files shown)


source/val/validate.cpp
```diff
@@ -399,6 +398,11 @@ spv_result_t ValidateBinaryUsingContextAndValidationState(
   // Catch undefined forward references before performing further checks.
   if (auto error = ValidateForwardDecls(*vstate)) return error;
 
+  for (size_t i = 0; i < vstate->ordered_instructions().size(); ++i) {
```
alan-baker: Please add a comment to explain why this is here so we're not tempted to move it later.

source/val/validate_decorations.cpp
```diff
@@ -1185,6 +1185,9 @@ spv_result_t CheckFPRoundingModeForShaders(ValidationState_t& vstate,
   // Validates Object operand of an OpStore
   for (const auto& use : inst.uses()) {
     const auto store = use.first;
+    if (store->opcode() == SpvOpFConvert) continue;
+    if (store->opcode() == SpvOpDecorate) continue;
```
alan-baker: Should probably skip spvOpcodeIsDebug too. I would also consider adding spvOpcodeIsAnnotation and skipping all those instead of just the decorate and member decorate.

source/val/validate_type.cpp
```diff
@@ -324,10 +324,12 @@ spv_result_t ValidateTypeFunction(ValidationState_t& _,
            << num_args << " arguments.";
   }
 
-  // The only valid uses of OpTypeFunction are in an OpFunction instruction.
+  // The only valid uses of OpTypeFunction are in an OpFunction or debugging
+  // instruction.
   for (auto& pair : inst->uses()) {
     const auto* use = pair.first;
-    if (use->opcode() != SpvOpFunction) {
+    if (use->opcode() != SpvOpFunction && use->opcode() != SpvOpName &&
```
alan-baker: Same here for skipping debug and annotations.

source/val/validate.cpp
zoddicus: Done

source/val/validate_decorations.cpp
zoddicus: Done.
FYI, there is no spvOpcodeIsAnnotation, I am assuming you meant spvOpcodeIsDecoration.

source/val/validate_type.cpp
zoddicus: Done

PR #2379, opened 2019-02-12
Add pass to remove dead members.

s-perron:
Add a pass that looks for members of structs whose values do not affects
the output of the shader.  Those members are then removed and just
treated like padding in the struct.

### DIFF
--- a/source/opt/eliminate_dead_members_pass.cpp
+++ b/source/opt/eliminate_dead_members_pass.cpp
@@ -187,15 +187,22 @@ void EliminateDeadMembersPass::MarkMembersAsLiveForAccessChain(
                     : 2);
   for (; i < inst->NumInOperands(); ++i) {
     Instruction* type_inst = get_def_use_mgr()->GetDef(type_id);
-    const analysis::IntConstant* member_idx =
-        const_mgr->FindDeclaredConstant(inst->GetSingleWordInOperand(i))
-            ->AsIntConstant();
-    assert(member_idx);
     switch (type_inst->opcode()) {
-      case SpvOpTypeStruct:
-        used_members_[type_id].insert(member_idx->GetU32());
-        type_id = type_inst->GetSingleWordInOperand(member_idx->GetU32());
-        break;
+      case SpvOpTypeStruct: {
+        const analysis::IntConstant* member_idx =
+            const_mgr->FindDeclaredConstant(inst->GetSingleWordInOperand(i))
+                ->AsIntConstant();
+        assert(member_idx);
+        if (member_idx->type()->AsInteger()->width() == 32) {
+          used_members_[type_id].insert(member_idx->GetU32());
+          type_id = type_inst->GetSingleWordInOperand(member_idx->GetU32());
+        } else {
+          used_members_[type_id].insert(
+              static_cast<uint32_t>(member_idx->GetU64()));
+          type_id = type_inst->GetSingleWordInOperand(
+              static_cast<uint32_t>(member_idx->GetU64()));
+        }
+      } break;
       case SpvOpTypeArray:
       case SpvOpTypeRuntimeArray:
       case SpvOpTypeVector:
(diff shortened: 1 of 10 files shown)


source/opt/eliminate_dead_members_pass.cpp
```diff
@@ -0,0 +1,578 @@
...
+        modified |= UpdateCompsiteInsert(inst);
+        break;
+      case SpvOpArrayLength:
+        modified |= UpdateOpArrayLength(inst);
+        break;
+      case SpvOpSpecConstantOp:
+        assert(false && "Not yet implemented.");
+        // with OpCompositeExtract, OpCompositeInsert
+        // For kernels: OpAccessChain, OpInBoundsAccessChain, OpPtrAccessChain,
+        // OpInBoundsPtrAccessChain
```
dnovillo: This will silently pass on optimized builds.  Should it be more explicit and fail the pass?

source/opt/eliminate_dead_members_pass.h
```diff
@@ -0,0 +1,144 @@
+// Copyright (c) 2017 Google Inc.
```
dnovillo: s/2017/2019/

source/opt/eliminate_dead_members_pass.cpp
```diff
@@ -0,0 +1,578 @@
...
+  uint32_t i = (inst->opcode() == SpvOpAccessChain ||
+                        inst->opcode() == SpvOpInBoundsAccessChain
+                    ? 1
+                    : 2);
+  for (; i < inst->NumInOperands(); ++i) {
+    Instruction* type_inst = get_def_use_mgr()->GetDef(type_id);
+    const analysis::IntConstant* member_idx =
+        const_mgr->FindDeclaredConstant(inst->GetSingleWordInOperand(i))
+            ->AsIntConstant();
+    assert(member_idx);
```
alan-baker: Move this into the SpvOpTypeStruct case below. It is not guaranteed for other types that the index is constant.

source/opt/eliminate_dead_members_pass.cpp
```diff
@@ -0,0 +1,578 @@
...
+    switch (type_inst->opcode()) {
+      case SpvOpTypeStruct:
+        used_members_[type_id].insert(member_idx->GetU32());
```
alan-baker: I know its not common to see non-32 bit integers, but we should probably handle 64 too at least. `GetU32` asserts that the width in 32.

source/opt/eliminate_dead_members_pass.cpp
```diff
@@ -0,0 +1,578 @@
...
+    case SpvOpReturnValue:
+      // This should be an issue only if we are returning from the entry point.
+      // However, for now I will keep it more conservative because functions are
+      // often inlined leaving only the entry points.
+      MarkOperandTypeAsFullyUsed(inst, 0);
+      break;
+    case SpvOpArrayLength:
+      MarkMembersAsLiveForArrayLength(inst);
+      break;
+    default:
```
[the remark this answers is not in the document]
s-perron: I could add that for safety when new instructions are added.  I'll do that.

source/opt/eliminate_dead_members_pass.cpp
s-perron: Sure I could get that done.

source/opt/eliminate_dead_members_pass.cpp
```diff
@@ -0,0 +1,578 @@
...
+    new_operands.emplace_back(inst->GetInOperand(1));
+  }
+
+  for (uint32_t i = static_cast<uint32_t>(new_operands.size());
+       i < inst->NumInOperands(); ++i) {
+    Instruction* type_inst = get_def_use_mgr()->GetDef(type_id);
+    const analysis::IntConstant* member_idx =
+        const_mgr->FindDeclaredConstant(inst->GetSingleWordInOperand(i))
+            ->AsIntConstant();
+    assert(member_idx);
```
alan-baker: Same issue with constant requirement.

source/opt/eliminate_dead_members_pass.cpp
```diff
@@ -0,0 +1,578 @@
...
+
+  if (!modified) {
+    return false;
+  }
+  inst->SetInOperands(std::move(new_operands));
+  context()->UpdateDefUse(inst);
+  return true;
+}
+
+bool EliminateDeadMembersPass::UpdateCompsiteInsert(Instruction* inst) {
```
alan-baker: UpdateCompsiteInsert -> UpdateCompositeInsert

test/opt/eliminate_dead_member_test.cpp
```diff
@@ -0,0 +1,812 @@
...
+
+using EliminateDeadMemberTest = opt::PassTest<::testing::Test>;
+
+TEST_F(EliminateDeadMemberTest, RemoveMember1) {
+  // Test that the member "y" is removed.
+  // Update OpMemberName for |y| and |z|.
+  // Update OpMemberDecorate for |y| and |z|.
+  // Update OpAccessChain for access to |z|.
+  const std::string text = R"(
+; CHECK: OpMemberName %type__Globals 0 "x"
```
alan-baker: You should probably add CHECK-NOTs for the removed member.

test/opt/eliminate_dead_member_test.cpp
```diff
@@ -0,0 +1,812 @@
...
+; CHECK: %type__Globals = OpTypeStruct %float %float
+; CHECK: OpAccessChain %_ptr_Uniform_float %_Globals %int_0
+; CHECK: OpAccessChain %_ptr_Uniform_float %_Globals %uint_1
+               OpCapability Shader
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint Vertex %main "main" %in_var_Position %gl_Position
+               OpSource HLSL 600
+               OpName %type__Globals "type.$Globals"
+               OpName %_Globals "$Globals"
+        %gr1 = OpDecorationGroup
```
alan-baker: This code is invalid. The decorations targeting a decoration group must precede it in the binary.

source/opt/eliminate_dead_members_pass.cpp
s-perron: I left this, but in the live member idenfitification, I added code to mark they types as live for the result type of the OpSpecConstantOp.  There could still be some bugs there, but in general it should be okay because I don't expect them to be used all that often.

s-perron:
@dnovillo @alan-baker I believe I have addressed all of your issues.  Please take another look.

s-perron:
The NDK build passed if you look at the build logs, but github did not seem to be notified.

s-perron:
The NDK build passed if you look at the build logs, but github did not seem to be notified.

s-perron:
I addressed @dnovillo's issues, and he may not be able to get back to it for a couple days.  I'll merge.  If he has any other issues, I'll address them in another PR.

dnovillo:
> I addressed @dnovillo's issues, and he may not be able to get back to it for a couple days. I'll merge. If he has any other issues, I'll address them in another PR.

Thanks.  I finally took another look.  LGTM.

Issue #2168, opened 2018-12-04
In Vulkan, disallow BufferBlock on StorageBuffer variables


alan-baker:
The validator should not accept BufferBlock annotated structs for StorageClass variables.

PR #2380, opened 2019-02-12
In Vulkan, disallow BufferBlock on StorageBuffer variables 

sarahM0:
To fix: #2168
Please review.

### DIFF
--- a/source/val/validate_decorations.cpp
+++ b/source/val/validate_decorations.cpp
@@ -910,7 +910,7 @@ spv_result_t CheckDecorationsOfBuffers(ValidationState_t& vstate) {
 
         if (spvIsVulkanEnv(vstate.context()->target_env)) {
 
-          if(storage_buffer && hasDecoration(id, SpvDecorationBlock, vstate)){
+          if(storage_buffer && hasDecoration(id, SpvDecorationBufferBlock, vstate)){
             return vstate.diag(SPV_ERROR_INVALID_ID, vstate.FindDef(var_id))
                    << "Storage buffer id '" << var_id
                    << "In Vulkan, disallow BufferBlock on StorageBuffer"
--- a/source/val/validate_decorations.cpp
+++ b/source/val/validate_decorations.cpp
@@ -913,7 +913,8 @@ spv_result_t CheckDecorationsOfBuffers(ValidationState_t& vstate) {
               hasDecoration(id, SpvDecorationBufferBlock, vstate)) {
             return vstate.diag(SPV_ERROR_INVALID_ID, vstate.FindDef(var_id))
                    << "Storage buffer id '" << var_id
-                   << " In Vulkan, BufferBlock is disallowed on variables in the StorageBuffer storage class";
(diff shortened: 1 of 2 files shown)


source/val/validate_decorations.cpp
```diff
@@ -909,6 +909,13 @@ spv_result_t CheckDecorationsOfBuffers(ValidationState_t& vstate) {
                     : (push_constant ? "PushConstant" : "StorageBuffer");
 
         if (spvIsVulkanEnv(vstate.context()->target_env)) {
+
+          if(storage_buffer && hasDecoration(id, SpvDecorationBlock, vstate)){
```
sarahM0: In debug mode, we never get to this line. Why?

alan-baker: If you check for SpvDecorationBufferBlock here, the test fails as expected. Not sure what you mean by not getting here.

test/val/val_decoration_test.cpp
```diff
@@ -2841,8 +2872,6 @@ TEST_F(ValidateDecorations,
             OpEntryPoint Fragment %1 "main"
             OpExecutionMode %1 OriginUpperLeft
 
-            OpDecorate %struct BufferBlock
```
sarahM0: I've removed this line from this test since leaving it as is makes the test fail (it's BufferBlock inside a StorageBuffer). But removing it would not affect what it's testing. Am I right?

source/val/validate_decorations.cpp
```diff
@@ -909,6 +909,13 @@ spv_result_t CheckDecorationsOfBuffers(ValidationState_t& vstate) {
                     : (push_constant ? "PushConstant" : "StorageBuffer");
 
         if (spvIsVulkanEnv(vstate.context()->target_env)) {
+          if (storage_buffer &&
+              hasDecoration(id, SpvDecorationBufferBlock, vstate)) {
+            return vstate.diag(SPV_ERROR_INVALID_ID, vstate.FindDef(var_id))
+                   << "Storage buffer id '" << var_id
+                   << "In Vulkan, disallow BufferBlock on StorageBuffer"
```
alan-baker: I would change the message to:

> In Vulkan, BufferBlock is disallowed on variables in the StorageBuffer storage class

test/val/val_decoration_test.cpp
dneto0: Whoops.   
Agreed the test was incorrect, using BufferBlock on decoration for a struct used in a variable in StorageBuffer. It should have had a Block decoration, and then this test would actually test what it says it's going to test.

The valid combinations are:

*     Block decoration on struct in StorageBuffer storage class
*     BufferBlock decoration on struct in Uniform storage class.

Both indicate a storage buffer.  (Originally only the second one existed.)

dneto0: Ah.  There was an existing problem with the test:   The struct needed member decorations to set its layout . I'll fix this in a new MR

PR #2383, opened 2019-02-13
Added reduction passes for removing selection constructs and simplify…

afd:
…ing selection constructs.

### DIFF
--- a/source/reduce/remove_selection_reduction_opportunity.cpp
+++ b/source/reduce/remove_selection_reduction_opportunity.cpp
@@ -29,13 +29,44 @@ bool RemoveSelectionReductionOpportunity::PreconditionHolds() {
 }
 
 void RemoveSelectionReductionOpportunity::Apply() {
-  block_->GetMergeInst()->context()->KillInst(block_->GetMergeInst());
+  auto merge_instruction = block_->GetMergeInst();
+  auto merge_block_id = merge_instruction->GetSingleWordInOperand(0);
   auto terminator = block_->terminator();
+
   terminator->SetOpcode(SpvOpBranch);
   // Pick one of the former conditional branch operands depending on whether we
   // are choosing the LHS or RHS.
   terminator->SetInOperands({choose_lhs_ ? terminator->GetInOperand(1)
                                          : terminator->GetInOperand(2)});
+
+  // Go through the function looking for breaks out of the selection, i.e.
+  // conditional branches that can jump to the selection's merge block.  Rewrite
+  // these as unconditional branches.
+  for (auto& block : *function_) {
+    for (auto& inst : block) {
+      if (inst.opcode() != SpvOpBranchConditional) {
+        // This is not a conditional branch; move on.
+        continue;
+      }
+      const auto true_target = inst.GetSingleWordInOperand(1);
+      const auto false_target = inst.GetSingleWordInOperand(2);
+      if (true_target != merge_block_id && false_target != merge_block_id) {
+        // This conditional branch does not target the merge block for the
+        // selection; move on.
+        continue;
+      }
+      // Change the conditional branch to an unconditional branch; if one of the
+      // targets of the branch is not the merge block of the selection then
+      // favour that target.
+      inst.SetOpcode(SpvOpBranch);
+      uint32_t new_target =
+          true_target == merge_block_id ? false_target : true_target;
+      inst.SetInOperands({{SPV_OPERAND_TYPE_ID, {new_target}}});
+    }
+  }
+
+  // Remove the merge instruction.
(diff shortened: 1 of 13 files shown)


source/reduce/remove_selection_reduction_opportunity_finder.cpp
```diff
@@ -0,0 +1,68 @@
...
+  // branching to the RHS of the construct, to reduce the extent to which
+  // reduction opportunities that can disable one another are adjacent.
+  for (auto choose_lhs : {true, false}) {
+    // Consider every function.
+    for (auto& function : *context->module()) {
+      // Consider every block in the function.
+      for (auto& block : function) {
+        // Check whether the block heads a selection construct.
+        auto merge_inst = block.GetMergeInst();
+        if (!merge_inst || merge_inst->opcode() != SpvOpSelectionMerge) {
```
alan-baker: Given code like:
```
%1 = OpLabel
OpSelectionMerge %merge None
OpBranchConditional %cond %2 %3
%2 = OpLabel
OpBranch %merge
%3 = OpLabel
OpBranchConditional %cond2 %merge %4
%4 = OpLabel
OpBranch %merge
```
I feel like the reducer will produce an invalid cfg. Because of the "break" in %3 there is no selection merge instruction, but if you remove the selection merge in %1, you would now need one in %3.

afd: Interesting counterexample, thanks.

My thought on how to go forward is to have the pass look for OpBranchConditional instructions for which one target is the conditional's merge block, and to change each to be just an OpBranch to the target that is not the merge block.

I believe that this should work because the only case where an OpBranchConditional inside a selection construct can be intimately tied to the selection construct has this form.  If an OpBranchConditional doesn't target the selection construct's merge block, then removing the selection construct cannot make that OpBranchConditional invalid (it might jump to the merge block or continue target of a loop, say, but that's fine as the loop isn't being removed).

What do you think?

alan-baker: So for my example you would change the branch in %3 to just go to %4 if you remove the merge instruction in %1? That sounds look it should work.

afd: Yes.  I'll adapt the PR accordingly.

afd: OK, done.

source/reduce/remove_selection_reduction_opportunity.cpp
```diff
@@ -0,0 +1,73 @@
...
+  // are choosing the LHS or RHS.
+  terminator->SetInOperands({choose_lhs_ ? terminator->GetInOperand(1)
+                                         : terminator->GetInOperand(2)});
+
+  // Go through the function looking for breaks out of the selection, i.e.
+  // conditional branches that can jump to the selection's merge block.  Rewrite
+  // these as unconditional branches.
+  for (auto& block : *function_) {
+    for (auto& inst : block) {
+      if (inst.opcode() != SpvOpBranchConditional) {
```
alan-baker: I doubt it would be common, but you could have an OpSwitch with a default and one other target (one of which is the selection's merge).

afd: I don't get it, sorry.  Would you be able to sketch an example?

alan-baker:
```
OpSwitch %value %merge 1 %other
; or
OpSwitch %value %other 1 %merge
```

It really just an odd expression of a conditional branch, but I could also do things like:
```
OpSwitch %value %other 1 %merge 2 %merge 3 %merge
```

As long as there is only a single non-merge target I think it is valid to not have a selection merge precede it.

paulthomson: I suppose it might be better to look at the predecessors of the merge block; if a predecessor has more than one successor, then I think we need to consider it. The header might be a special case. There might be other issues.

On realizing the subtlety of these examples, it feels tempting to reconsider the previously discussed idea where we instead just try to simplify the conditional branch so both targets are the same, then eliminate dead code. The problematic branches inside the selection will still be valid, but they should then also get simplified in the same way, until the original selection construct can be trivially removed (we can check the predecessors of the merge block perhaps). 

We might reconsider limiting these reduction passes to OpSelectionMerge; we might be able to just consider the conditional branch instructions. Need to consider more examples though.

Another worry: what if we have a continue construct that contains a multi-way branch and we end up "simplifying" this so that the back-edge becomes unreachable; I think this will be invalid. On the other hand, if the validator can detect this it almost doesn't matter.

afd:
Closing this as it has bee superseded by other PRs.

Issue #2128, opened 2018-11-28
OpAtomicLoad, OpAtomicStore, OpAtomicExchange can operate on int or float value. Except for Vulkan environment that only operates on int value.


dneto0:
From Khronos internal issue 254.

This reverses #1091 which was implemented with #1229

PR #2385, opened 2019-02-13
 OpAtomicLoad, OpAtomicStore, OpAtomicExchange can operate on int or float value. Except for Vulkan environment that only operates on int value.

sarahM0:
To fix: #2128 
Please review.

### DIFF
--- a/source/val/validate_atomics.cpp
+++ b/source/val/validate_atomics.cpp
@@ -55,8 +55,7 @@ spv_result_t AtomicsPass(ValidationState_t& _, const Instruction* inst) {
       if (_.HasCapability(SpvCapabilityKernel) &&
           (opcode == SpvOpAtomicLoad || opcode == SpvOpAtomicExchange ||
            opcode == SpvOpAtomicCompareExchange)) {
-        if (!(_.IsFloatScalarType(result_type) &&
-              !spvIsVulkanEnv(_.context()->target_env)) &&
+        if (!_.IsFloatScalarType(result_type) &&
             !_.IsIntScalarType(result_type)) {
           return _.diag(SPV_ERROR_INVALID_DATA, inst)
                  << spvOpcodeString(opcode)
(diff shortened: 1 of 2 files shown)


source/val/validate_atomics.cpp
```diff
@@ -55,7 +55,8 @@ spv_result_t AtomicsPass(ValidationState_t& _, const Instruction* inst) {
       if (_.HasCapability(SpvCapabilityKernel) &&
           (opcode == SpvOpAtomicLoad || opcode == SpvOpAtomicExchange ||
            opcode == SpvOpAtomicCompareExchange)) {
-        if (!_.IsFloatScalarType(result_type) &&
+        if (!(_.IsFloatScalarType(result_type) &&
```
alan-baker: This code is gated by the Kernel capability. That should never be present in Vulkan.

test/val/val_atomics_test.cpp
```diff
@@ -225,6 +225,28 @@ TEST_F(ValidateAtomics, AtomicLoadKernelSuccess) {
...
+TEST_F(ValidateAtomics, AtomicLoadFloatVulkan) {
+  const std::string body = R"(
+%val1 = OpAtomicLoad %f32 %f32_var %device %relaxed
+%val2 = OpAtomicLoad %f32 %f32_var %workgroup %acquire
+)";
+
+  CompileSuccessfully(GenerateShaderCode(body), SPV_ENV_VULKAN_1_0);
+  ASSERT_EQ(SPV_ERROR_INVALID_DATA, ValidateInstructions(SPV_ENV_VULKAN_1_0));
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("expected Result Type to be int scalar type"));
```
alan-baker: This is not the error message for the code you changed. Did these tests work without your change in validate_atomics?

sarahM0: Your right this issue is already implemented in AtomicPass. I'm reverting my change in validate_atomics.cpp.

PR #2386, opened 2019-02-13
Implement WebGPU specific CFG validation

### LINKED ISSUE #2068: Validate CGFs for WebGPU
> #### CFGs
> Each block B in a function must satisfy one of the following rules:
> - B is the entry block (the first block listed in the function body), or there is a control flow path from the entry block to B. (We say B is reachable.)
> - There is no control flow path from the entry block to B, but:
>     * B is named as the merge-block for a merge instruction in a reachable block. Additionally there are no branches to B, and B contains only an OpLabel and an OpUnreachable instruction.
>     * B is named as the continue-target for an OpLoopMerge instruction in a reachable block H. Additionally there are no branches to B, and B contains only an OpLabel and an OpBranch instruction to H.

zoddicus:
In WebGPU all blocks are required to be reachable, unless they are one of two
specific degenerate cases for merge-block or continue-target. This PR adds in
checking for these conditions.

Fixes #2068


source/val/function.h
```diff
@@ -232,6 +232,14 @@ class Function {
     return function_call_targets_;
   }
 
+  BasicBlock* GetMergeHeader(BasicBlock* merge_block) {
```
alan-baker: Function comments. Especially useful would be indicating when these become usable.

source/val/validate_cfg.cpp
```diff
@@ -610,6 +611,122 @@ spv_result_t StructuredControlFlowChecks(
...
+               << "For WebGPU, unreachable continue-targe must be referenced "
+                  "by a reachable loop instruction.";
+      }
+
+      // 2. Confirm that continue-target has only a single back edge to loop
+      // block.
+      auto target_id = block->id();
+      auto header_id = continue_header->id();
+      bool seen_back_edge = false;
+      for (auto& edge : back_edges) {
```
alan-baker: Might be easier just to check that the terminator is an OpBranch that targets the header instead of looping through all back-edges?

source/val/validate_cfg.cpp
```diff
@@ -610,6 +611,122 @@ spv_result_t StructuredControlFlowChecks(
...
+      auto label_idx = label_inst - &_.ordered_instructions()[0];
+      auto terminator_idx = terminator_inst - &_.ordered_instructions()[0];
+      if (label_idx + 1 != terminator_idx) {
+        return _.diag(SPV_ERROR_INVALID_CFG, _.FindDef(block->id()))
+               << "For WebGPU, unreachable continue-target must only contain "
+                  "an OpLabel and an OpBranch instruction.";
+      }
+
+      // 4. Use OpLabel instruction to confirm there is only 1 use of the block.
+      if (label_inst->uses().size() != 1) {
```
alan-baker: What about?
```
%loop = OpLabel
OpBranch %body
%body = OpLabel
OpSelectionMerge %continue
OpBranchConditional %c %then %else
%then = OpLabel
OpReturn
%else = OpLabel
OpReturn
%continue = OpLabel
OpBranch %loop
%merge = OpLabel
OpUnreachable
```

`%continue` is referenced twice, but this strikes me as valid. Maybe this and the merge check should be that all uses must be merge instructions?

zoddicus: For you specific example is isn't valid, because %continue is being used a continue-target for OpLoopMerge and a merge-block for OpSelectionMerge, and those have mutually incompatible rules for termination of the block. That being said, there might an issue with if a block is referenced from multiple of the same type of header.

For selection merge, it seems likely that someone might want all of the degenerate merge-blocks collesced into a single example, so that defiantely needs to be investigated.

For loop merge, given that the target has to have a branch back to the header, I don't think it will be possible to encode multiple usages without breaking that rule.


... never mind, there is another rule in the validator that requires uniqueness of merge-blocks... I added related tests and refactored the code to more closely match what the spec says.

alan-baker: This strikes me as bug in the environment spec. @dneto0 what do you think. My example could be generated by a frontend easily enough. To make it legal would require introducing or changing the selection's merge target.

dneto0: Right, merge blocks must be unique (1-1 mapping back to header).

Yes, there is nothing stopping a continue target from also being a merge block.  That's @alan-baker's example just above here.

Another troublesome case is a continue target that is named by two different loop-merge instructions.  something like this:

```
%cond0 = ...
%cond1 = ...

%outer = OpLabel
OpLoopMerge %outermerge %continue None
OpBranch %inner

        %inner = OpLabel
        OpLoopMerge %innermerge %continue
        OpBranchConditional %cond0 %innermerge %continue

                %continue = OpLabel
                OpBranchConditional %cond1 %inner %outer ; branches to both loop headers

        %innermerge = OpLabel
        OpBranch %outermerge

%outermerge = OpLabel
OpReturn
```

source/val/function.cpp
```diff
@@ -86,6 +86,7 @@ spv_result_t Function::RegisterLoopMerge(uint32_t merge_id,
   continue_construct.set_corresponding_constructs({&loop_construct});
   loop_construct.set_corresponding_constructs({&continue_construct});
   merge_block_header_[&merge_block] = current_block_;
+  continue_target_header_[&continue_target_block] = current_block_;
```
dneto0: A continue target could be the declared continue target of multiple loop headers.  See example in https://github.com/KhronosGroup/SPIRV-Tools/pull/2386#discussion_r257400429

source/val/validate_cfg.cpp
zoddicus: I agree a block can be both a continue-target and a merge-block per @alan-baker's example. I talked with @dneto0 about the multiple loop header case. For the case of an unreachable continue-target, the rule about ending in an OpBranch to the header means that having multiple loops using the same target will fail either due to not ending in an OpBranch, or not being able to fulfill the rule about branching back to the referencing header for atleast one of them. Though the code still needs to properly handle multiple referencing headers.

source/val/function.cpp
zoddicus: Done

source/val/validate_cfg.cpp
zoddicus: I have added in some test cases for the block being both a merge and continue.

zoddicus:
Can I get a re-review on this?

PR #2390, opened 2019-02-14
Add utils/format script.

dj2:
This is a simple script to run clang-format over all the bits of the
repo. Makes it easier when submitting a CL to get everything formatted
as neede.

### DIFF
--- a/utils/format
+++ b/utils/format
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+find examples -name "*.h"  -exec clang-format -i {} \;
+find examples -name "*.cc"  -exec clang-format -i {} \;
+find source -name "*.h"  -exec clang-format -i {} \;
+find source -name "*.cc"  -exec clang-format -i {} \;
+find test -name "*.h"  -exec clang-format -i {} \;
+find test -name "*.cc"  -exec clang-format -i {} \;
+find tools -name "*.h"  -exec clang-format -i {} \;
+find tools -name "*.cc"  -exec clang-format -i {} \;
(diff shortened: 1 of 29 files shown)


utils/format
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+find examples -name "*.h"  -exec clang-format -i {} \;
+find examples -name "*.cc"  -exec clang-format -i {} \;
```
dneto0: SPIRV-Tools uses .cpp and .hpp files.

utils/format
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+find examples -name "*.h"  -exec clang-format -i {} \;
```
paulthomson: I believe clang-format can accept a list of files, which might be better, especially for those on Windows, where process creation can be slow. Perhaps not a big deal though.

utils/format
dneto0: I'd probably do a double loop

 for dir in source test tools include; do
  . ..
 ...

dneto0: or do something like this.

alias cgrep='find . -type f \( -name "*.c" -o -name "*.cc" -o -name "*.h" -o -name "*.hpp" -o -name "*.cpp" \) -print0 | xargs -0 grep'

dj2: done

utils/format
dj2: This seems like the easiest way to get the list of file names? I'd need to do something to loop over all dirs to find the files, yea?

dj2:
Closing this, it seems like git clang-format master would do what we want.

PR #2395, opened 2019-02-14
Refactor webgpu-mode pass ran tests to be parameterized

### LINKED ISSUE #2394: Refactor webgpu-mode pass ran tests to be parameterized
There is going to be a number of these tests in the future, so the boilerplate should be de-duplicated.

zoddicus:
Fixes #2394


zoddicus:
Accidently ammended to the commit when fixing formating, so had to force push.

PR #2398, opened 2019-02-15
reduce: remove unreferenced blocks pass

paulthomson:


source/reduce/remove_block_reduction_opportunity.cpp
```diff
@@ -0,0 +1,55 @@
...
+  // Removing other blocks cannot disable this opportunity.
+  return true;
+}
+
+void RemoveBlockReductionOpportunity::Apply() {
+  // We need an iterator pointing to the block, hence the loop.
+  for (auto bi = function_->begin(); bi != function_->end(); ++bi) {
+    if (bi->id() == block_->id()) {
+      bi->KillAllInsts(true);
+      bi.Erase();
```
paulthomson: This "RemoveBlock" code is seen elsewhere. Should it be refactored into some common function?

source/reduce/remove_block_reduction_opportunity.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+using namespace opt;
+
+RemoveBlockReductionOpportunity::RemoveBlockReductionOpportunity(
+    IRContext* context, Function* function, BasicBlock* block)
+    : context_(context),
+      function_(function),
+      block_(block) {
+  // precondition:
+  assert(context->get_def_use_mgr()->NumUsers(block->id()) == 0);
```
afd: I favour using && "message" to make it clear when the assertion fails what the cause is.

source/reduce/remove_block_reduction_opportunity.h
```diff
@@ -0,0 +1,47 @@
...
+#include "source/opt/basic_block.h"
+#include "source/opt/function.h"
+
+namespace spvtools {
+namespace reduce {
+
+// An opportunity to remove an unreferenced block.
+class RemoveBlockReductionOpportunity : public ReductionOpportunity {
+ public:
+  // Creates the opportunity to remove |block|.
```
afd: maybe mention that |block| is inside |function| within |context|

source/reduce/remove_block_reduction_opportunity_finder.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "source/reduce/remove_block_reduction_opportunity_finder.h"
+#include "source/reduce/remove_block_reduction_opportunity.h"
+
+namespace spvtools {
+namespace reduce {
+
+using namespace opt;
+
+std::string RemoveBlockReductionOpportunityFinder::GetName() const {
+  return "RemoveBlocksReductionOpportunityFinder";
```
afd: Blocks -> Block in string

source/reduce/remove_block_reduction_opportunity.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+bool RemoveBlockReductionOpportunity::PreconditionHolds() {
+  // Removing other blocks cannot disable this opportunity.
+  return true;
+}
+
+void RemoveBlockReductionOpportunity::Apply() {
+  // We need an iterator pointing to the block, hence the loop.
+  for (auto bi = function_->begin(); bi != function_->end(); ++bi) {
```
alan-baker: If I have blocks A and B that are unreachable and A branches to B and B uses some values defined in A, you may leave dangling uses in B I think. Since B is referenced it wouldn't be removed. Consider replacing uses in B of values defined in A with undef.

afd: Good point.  Rather than actually replacing uses, I think it would be better to rely on existing reduction passes that remove uses to get rid of them (I'm in favour of keeping passes as separate as possible unless we find that it's really necessary to have one pass do work that another pass can also do.)

alan-baker: So would you check that none of the instruction in A has any uses?

paulthomson: Yes, although I suppose uses from inside the block are OK. I have attempted to do this. Still need to add a unit test.

paulthomson:
OK I have added tests for Alan's dangling uses example that now pass

source/reduce/remove_block_reduction_opportunity.cpp
paulthomson: OK. Unit tests added.

PR #2401, opened 2019-02-19
Add some val options to reduce

### LINKED ISSUE #2396: [spirv-reduce] Accept options to pass to the validator.
DXC generates spir-v files that do not pass the validator unless certains options are pass to it, like `--skip-block-layout`.  However, I do not see how to tell spirv-reduce to use those options when calling the validator.  This means I am undable to use spirv-reduce on those spv files.

paulthomson:
Fix #2396 

* Change Reducer::Run to *not* immediately copy the input binary.
* Check that initial shader is valid.
* Fix RemoveOpnameAndRemoveUnreferenced test; turns out the original shader is invalid, but we never notice because we don't check this and the reduced shader is valid; fix original shader. Assert reduction status is kComplete.
* Always check return value from `Reducer::Run`.

### DIFF
--- a/tools/reduce/reduce.cpp
+++ b/tools/reduce/reduce.cpp
@@ -286,8 +286,8 @@ int main(int argc, const char** argv) {
   }
 
   std::vector<uint32_t> binary_out;
-  const auto reduction_status =
-      reducer.Run(binary_in, &binary_out, reducer_options, validator_options);
+  const auto reduction_status = reducer.Run(std::move(binary_in), &binary_out,
+                                            reducer_options, validator_options);
 
   if (reduction_status ==
           Reducer::ReductionResultStatus::kInitialStateNotInteresting ||
--- a/tools/reduce/reduce.cpp
+++ b/tools/reduce/reduce.cpp
@@ -286,8 +286,8 @@ int main(int argc, const char** argv) {
   }
 
   std::vector<uint32_t> binary_out;
-  const auto reduction_status = reducer.Run(std::move(binary_in), &binary_out,
-                                            reducer_options, validator_options);
+  const auto reduction_status =
+      reducer.Run(binary_in, &binary_out, reducer_options, validator_options);
 
   if (reduction_status ==
           Reducer::ReductionResultStatus::kInitialStateNotInteresting ||
--- a/tools/reduce/reduce.cpp
+++ b/tools/reduce/reduce.cpp
@@ -110,6 +110,22 @@ Options (in lexicographical order):
                reducer will take before giving up.
   --version
                Display reducer version information.
+
+Supported validator options:
+  --relax-logical-pointer          Allow allocating an object of a pointer type and returning
+                                   a pointer value from a function in logical addressing mode
+  --relax-block-layout             Enable VK_KHR_relaxed_block_layout when checking standard
+                                   uniform, storage buffer, and push constant layouts.
+                                   This is the default when targeting Vulkan 1.1 or later.
+  --scalar-block-layout            Enable VK_EXT_scalar_block_layout when checking standard
+                                   uniform, storage buffer, and push constant layouts.  Scalar layout
+                                   rules are more permissive than relaxed block layout so in effect
+                                   this will override the --relax-block-layout option.
+  --skip-block-layout              Skip checking standard uniform/storage buffer layout.
+                                   Overrides any --relax-block-layout or --scalar-block-layout option.
+  --relax-struct-store             Allow store from one struct type to a
+                                   different type with compatible layout and
+                                   members.
 )",
       program, program);
 }
@@ -127,7 +143,8 @@ void ReduceDiagnostic(spv_message_level_t level, const char* /*source*/,
 
 ReduceStatus ParseFlags(int argc, const char** argv, const char** in_file,
                         const char** interestingness_test,
-                        spvtools::ReducerOptions* reducer_options) {
+                        spvtools::ReducerOptions* reducer_options,
+                        spvtools::ValidatorOptions* validator_options) {
   uint32_t positional_arg_index = 0;
 
   for (int argi = 1; argi < argc; ++argi) {
@@ -164,6 +181,19 @@ ReduceStatus ParseFlags(int argc, const char** argv, const char** in_file,
       assert(!*interestingness_test);
       *interestingness_test = cur_arg;
       positional_arg_index++;
+
+      // TODO: Refactor. Repeated C++ code for parsing these options. Also
+      // consider removing repeated help text.
+    } else if (0 == strcmp(cur_arg, "--relax-logical-pointer")) {
+      validator_options->SetRelaxLogicalPointer(true);
+    } else if (0 == strcmp(cur_arg, "--relax-block-layout")) {
+      validator_options->SetRelaxBlockLayout(true);
+    } else if (0 == strcmp(cur_arg, "--scalar-block-layout")) {
+      validator_options->SetScalarBlockLayout(true);
+    } else if (0 == strcmp(cur_arg, "--skip-block-layout")) {
+      validator_options->SetSkipBlockLayout(true);
+    } else if (0 == strcmp(cur_arg, "--relax-struct-store")) {
+      validator_options->SetRelaxStructStore(true);
     } else {
       spvtools::Error(ReduceDiagnostic, nullptr, {},
                       "Too many positional arguments specified");
@@ -195,9 +225,10 @@ int main(int argc, const char** argv) {
 
   spv_target_env target_env = kDefaultEnvironment;
   spvtools::ReducerOptions reducer_options;
+  spvtools::ValidatorOptions validator_options;
 
-  ReduceStatus status =
-      ParseFlags(argc, argv, &in_file, &interestingness_test, &reducer_options);
+  ReduceStatus status = ParseFlags(argc, argv, &in_file, &interestingness_test,
+                                   &reducer_options, &validator_options);
 
   if (status.action == REDUCE_STOP) {
     return status.code;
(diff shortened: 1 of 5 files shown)


paulthomson:
Could add a test to `test/reduce/validation_during_reduction_test.cpp`. @s-perron Do you have a simple example of a SPIR-V module that would fail to validate without setting some validator options? Perhaps also with an unreferenced block that the reducer could remove? If so, that would work as a test.

s-perron:
Here is a small shader based on code generated by DXC:

```
               OpCapability Shader
               OpCapability SampledBuffer
               OpCapability StorageImageExtendedFormats
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %2 "SceneLuminance" %3
               OpExecutionMode %2 OriginUpperLeft
               OpSource HLSL 600
               OpDecorate %3 Location 0
               OpDecorate %4 DescriptorSet 1
               OpDecorate %4 Binding 0
               OpMemberDecorate %_struct_6 0 Offset 0
               OpMemberDecorate %_struct_6 1 Offset 1072
               OpMemberDecorate %_struct_7 0 Offset 0
               OpDecorate %_struct_7 Block
  %_struct_8 = OpTypeStruct
      %float = OpTypeFloat 32
        %int = OpTypeInt 32 1
      %int_0 = OpConstant %int 0
  %_struct_6 = OpTypeStruct %v4float %_struct_8
  %_struct_7 = OpTypeStruct %_struct_6
%_ptr_Uniform__struct_7 = OpTypePointer Uniform %_struct_7
%_ptr_Output_v4float = OpTypePointer Output %v4float
       %void = OpTypeVoid
         %17 = OpTypeFunction %void
          %4 = OpVariable %_ptr_Uniform__struct_7 Uniform
          %3 = OpVariable %_ptr_Output_v4float Output
          %2 = OpFunction %void None %17
         %19 = OpLabel
         %20 = OpAccessChain %_ptr_Uniform_v4float %4 %int_0 %int_0
         %21 = OpLoad %v4float %20
               OpStore %3 %21
               OpReturn
               OpFunctionEnd
```

paulthomson:
I made a few more changes here than expected. @afd FYI

* Change Reducer::Run signature; no need for r-value reference as argument is immediately copied.

* Check that initial shader is valid.

* Fix RemoveOpnameAndRemoveUnreferenced test; turns out the original shader is invalid, but we never notice because we don't check this and the reduced shader is valid; fix original shader. Assert reduction status is kComplete.

s-perron:
You can try this shader instead:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %2 "Main" %gl_Position
               OpSource HLSL 600
               OpDecorate %gl_Position BuiltIn Position
               OpDecorate %4 DescriptorSet 0
               OpDecorate %4 Binding 99
               OpDecorate %_arr_v4float_uint_2 ArrayStride 16
               OpMemberDecorate %_struct_6 0 Offset 0
               OpMemberDecorate %_struct_6 1 Offset 32
               OpMemberDecorate %_struct_6 1 MatrixStride 16
               OpMemberDecorate %_struct_6 1 ColMajor
               OpMemberDecorate %_struct_6 2 Offset 96
               OpMemberDecorate %_struct_6 3 Offset 100
               OpMemberDecorate %_struct_6 4 Offset 112
               OpMemberDecorate %_struct_6 4 MatrixStride 16
               OpMemberDecorate %_struct_6 4 ColMajor
               OpMemberDecorate %_struct_6 5 Offset 176
               OpDecorate %_struct_6 Block
      %float = OpTypeFloat 32
%mat4v4float = OpTypeMatrix %v4float 4
    %v2float = OpTypeVector %float 2
        %int = OpTypeInt 32 1
       %uint = OpTypeInt 32 0
     %uint_2 = OpConstant %uint 2
      %int_1 = OpConstant %int 1
      %int_5 = OpConstant %int 5
%_arr_v4float_uint_2 = OpTypeArray %v4float %uint_2
  %_struct_6 = OpTypeStruct %_arr_v4float_uint_2 %mat4v4float %uint %v2float %mat4v4float %float
%_ptr_Uniform__struct_6 = OpTypePointer Uniform %_struct_6
%_ptr_Output_v4float = OpTypePointer Output %v4float
       %void = OpTypeVoid
         %19 = OpTypeFunction %void
%_ptr_Uniform_float = OpTypePointer Uniform %float
          %4 = OpVariable %_ptr_Uniform__struct_6 Uniform
%gl_Position = OpVariable %_ptr_Output_v4float Output
          %2 = OpFunction %void None %19
         %21 = OpLabel
         %22 = OpAccessChain %_ptr_Uniform_float %4 %int_5
         %23 = OpLoad %float %22
         %24 = OpCompositeConstruct %v4float %23 %23 %23 %23
               OpStore %gl_Position %24
               OpReturn
               OpFunctionEnd
```

paulthomson:
Thanks! That works. Sorry for the delay. 

What do you mean by the old interface?

s-perron:
I was referring to this change: https://github.com/KhronosGroup/SPIRV-Tools/pull/2401/files#diff-88958976449829c9895e852ac25bcef2R85.  As I think about it more, all old code should still work, it just could be better.

paulthomson:
Ah I see. Thinking about it more, I prefer the original signature, so I have changed it back. I think my issue was that `Reducer::Run` immediately copied the `binary_in` parameter, which seemed pointless. So I have fixed that. I also found a few other calls to `Reducer::Run` in tests where the return value was not checked, so fixed those too.

tools/reduce/reduce.cpp
```diff
@@ -110,6 +110,22 @@ Options (in lexicographical order):
...
+                                   This is the default when targeting Vulkan 1.1 or later.
+  --scalar-block-layout            Enable VK_EXT_scalar_block_layout when checking standard
+                                   uniform, storage buffer, and push constant layouts.  Scalar layout
+                                   rules are more permissive than relaxed block layout so in effect
+                                   this will override the --relax-block-layout option.
+  --skip-block-layout              Skip checking standard uniform/storage buffer layout.
+                                   Overrides any --relax-block-layout or --scalar-block-layout option.
+  --relax-struct-store             Allow store from one struct type to a
+                                   different type with compatible layout and
+                                   members.
```
afd: ^ How painful would it be to factor out the text for these help options?  Their descriptions could well change in the validator, and it would be a shame for them to get out of sync.  I appreciate that with the validator being a separate project this might not be easy.

afd:
When I wrote "I'd almost rather say "run spirv-val --help for a description" than duplicate the text", I mean I would prefer to actually factor out the text so that it can be shown in both places in full.  But if that's not possible then it might be better to do what I propose here rather than duplicate it.

tools/val/val_strings.h
```diff
@@ -0,0 +1,40 @@
...
+                                   uniform, storage buffer, and push constant layouts.  Scalar layout
+                                   rules are more permissive than relaxed block layout so in effect
+                                   this will override the --relax-block-layout option.
+  --skip-block-layout              Skip checking standard uniform/storage buffer layout.
+                                   Overrides any --relax-block-layout or --scalar-block-layout option.
+  --relax-struct-store             Allow store from one struct type to a
+                                   different type with compatible layout and
+                                   members.)";
+
+}  // namespace val
```
afd: My concern about this is that it's not future-proof.  These options are shared with spirv-reduce, but they could end up getting shared with other tools.  And then maybe not all of them would be shared - perhaps a subset of them, or a partially-overlapping set that isn't a subset.  Also the name "shared_validator_options_string" doesn't indicate what they are shared with, yet it would be bad to have the validator refer explicitly to the reducer.

I suggest instead having one const char* per option:

const char* validator_option_relax_logical_pointer = ...;

const char* validator_option_relax_block_layout = ...;

etc.

This would make it possible to expose combinations of option help strings to other tools in a fine-grained manner and would avoid the problem of documenting vs. not documenting in the validator the fact that these options are shared.

Side note: are there naming conventions for string constants, such as starting their names with "k"?  Or is that just for numeric constants?

paulthomson:
I feel that the code is a hack either way and the factoring out is just the minimal approach to reduce repetition; the parsing of these args is still repeated, the strings in the parsing code are repeated, if more validator options are added, they need to be manually added to the reducer and both help texts, if appropriate. The help text is still separated from the arguments. If more sharing is needed, then there will need to be manual changes and checks either way. Having a string constant defined in a header is also pretty weird in itself; not sure if I should have done this, but using macros seemed unpleasant and adding another shared compilation unit seemed like overkill. If more sharing is needed *and* we care about making the code nice, we should use an argument parsing library so we can have a function that adds the shared arguments (with the help text) to an argument parser. But in summary, I made the changes :)

tools/val/val_strings.h
afd: @alan-baker I'd be curious to know what you think about this?  Options are:

- have relevant validator options text duplicated between validator and reducer
- factor out the text like Paul has done here
- avoid having the reducer explain those validator options at all, and instead have it say: "See spirv-val --help for details on the following validator options:"

alan-baker: I think for now I'd prefer something like the third option:
```
--relax-block-layout . Validate using Vulkan relaxed block layout. See spirv-val --help for a detailed description
```

Then we can consider how we should best share options separately from this PR.

paulthomson: I have done the following to avoid repeating the help text. See what you think. 

```
Supported validator options are as follows. See `spirv-val --help` for details.
  --relax-logical-pointer
  --relax-block-layout
  --scalar-block-layout
  --skip-block-layout
  --relax-struct-store
```

paulthomson:
I have reverted the help text refactoring and removed the repeated help text, pointing to user to `spirv-val --help` instead.

PR #2408, opened 2019-02-20
It is invalid to apply both Restrict and Aliased to the same <id>

sarahM0:
To fix: #1298

### DIFF
--- a/test/opt/aggressive_dead_code_elim_test.cpp
+++ b/test/opt/aggressive_dead_code_elim_test.cpp
@@ -5151,7 +5151,6 @@ OpFunctionEnd
 TEST_F(AggressiveDCETest, ParitallyDeadDecorationGroup) {
   const std::string text = R"(
 ; CHECK: OpDecorate [[grp:%\w+]] Restrict
-; CHECK: OpDecorate [[grp]] Aliased
 ; CHECK: [[grp]] = OpDecorationGroup
 ; CHECK: OpGroupDecorate [[grp]] [[output:%\w+]]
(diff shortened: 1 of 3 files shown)


test/opt/aggressive_dead_code_elim_test.cpp
```diff
@@ -5161,7 +5160,6 @@ OpMemoryModel Logical GLSL450
 OpEntryPoint Fragment %main "main" %output
 OpExecutionMode %main OriginUpperLeft
 OpDecorate %1 Restrict
-OpDecorate %1 Aliased
```
sarahM0: I deleted _OpDecorate %1 Aliased_ to stop it from failing. Would you double check if I need to replace with something else or the test is still valuable?

test/val/val_decoration_test.cpp
```diff
@@ -5097,6 +5097,37 @@ TEST_F(ValidateDecorations, NoUnsignedWrapExtInstGLSLGood) {
...
+TEST_F(ValidateDecorations, AliasedandRestrictBad) {
+  const std::string body = R"(
+OpCapability Shader
+%1 = OpExtInstImport "GLSL.std.450"
+OpMemoryModel Logical GLSL450
+OpEntryPoint GLCompute %main "main"
+OpExecutionMode %main LocalSize 1 1 1
+OpSource GLSL 430
+OpMemberDecorate %Output 0 Offset 0
+OpDecorate %Output Restrict
```
alan-baker: The Aliased and Restrict decorations should target %dataOutput since decorations can only be applied to memory object declarations.

PR #2410, opened 2019-02-20
Validate: (data) Block can't appear within a Block

sarahM0:
To fix: #1587


source/val/validate_type.cpp
```diff
@@ -237,6 +237,29 @@ spv_result_t ValidateTypeStruct(ValidationState_t& _, const Instruction* inst) {
...
+        has_nested_blockOrBufferBlock_struct = true;
+    }
+  }
+
+  _.SetHasNestedBlockOrBufferBlockStruct(inst->id(),
+                                         has_nested_blockOrBufferBlock_struct);
+  if (_.GetHasNestedBlockOrBufferBlockStruct(inst->id()) &&
+      (_.HasDecoration(inst->id(), SpvDecorationBufferBlock) ||
+       _.HasDecoration(inst->id(), SpvDecorationBlock))) {
+    return _.diag(SPV_ERROR_INVALID_ID, inst)
```
alan-baker: Please change to "A Block or BufferBlock cannot be nested within another Block or BufferBlock"

PR #2412, opened 2019-02-20
Don't change type of input and output var in dead member elim

s-perron:
The types of input and output variables must match for the pipeline.  We
cannot see the uses in all of the shader, so dead member
elimination cannot safely change the type of input and output variables.


s-perron:
@greg-lunarg FYI.

PR #2413, opened 2019-02-21
Add --strip-atomic-counter-memory

### LINKED ISSUE #2242: When converting from Vulkan to WebGPU strip AtomicCounterMemory
This bit is not actually used in Vulkan, but is allowed without its capability being declared. When targeting a WebGPU environment it should be stripped out.

zoddicus:
Adds an optimization pass to remove usages of AtomicCounterMemory
bit. This bit is ignored in Vulkan environments and outright forbidden
in WebGPU ones.

Fixes #2242


source/opt/strip_atomic_counter_memory_pass.h
```diff
@@ -22,11 +22,26 @@
...
 
-// DNS: Write documentation
+// Removes the AtomicCounterMemory bit from the value being passed into memory
+// semantics. This bit being set is ignored in Vulkan environments and
+// forbidden WebGPU ones.
 class StripAtomicCounterMemoryPass : public Pass {
  public:
   const char* name() const override { return "strip-atomic-counter-memory"; }
   Status Process() override;
+
```
zoddicus: Can someone with a deeper understanding of the optimizer confirm this list?
I removed kAnalysisDefUse and  kAnalysisConstants, since I am pretty sure this pass may change things related to those analyses.

source/opcode.h
```diff
@@ -133,4 +133,8 @@ bool spvOpcodeIsScalarizable(SpvOp opcode);
 // Returns true if the given opcode is a debug instruction.
 bool spvOpcodeIsDebug(SpvOp opcode);
 
+// Returns a vector the indices of the memory semantics <id> operands for
```
s-perron: Returns a vector **containing** the indices ...

source/opt/optimizer.cpp
```diff
@@ -219,6 +219,8 @@ Optimizer& Optimizer::RegisterSizePasses() {
 
 Optimizer& Optimizer::RegisterWebGPUPasses() {
   return RegisterPass(CreateStripDebugInfoPass())
+      .RegisterPass(CreateStripAtomicCounterMemoryPass())
+      .RegisterPass(CreateEliminateDeadConstantPass())
```
s-perron: The dead const pass is not needed.  The DCE pass below will remove the unused constants.

source/opt/strip_atomic_counter_memory_pass.cpp
```diff
@@ -0,0 +1,58 @@
...
+      mem_sem_val &= ~SpvMemorySemanticsAtomicCounterMemoryMask;
+
+      analysis::Integer int_type(32, false);
+      const analysis::Type* uint32_type =
+          type_mgr->GetRegisteredType(&int_type);
+      auto* new_const = constant_mgr->GetConstant(uint32_type, {mem_sem_val});
+      auto* new_const_inst = constant_mgr->GetDefiningInstruction(new_const);
+      auto new_const_id = new_const_inst->result_id();
+
+      inst->SetOperand(idx, {new_const_id});
```
s-perron: After this, you should call `context()->UpdateDefUse(inst)`.  This will keep the def-use manager up-to-date.  Invalidating it and rebuilding is an expensive operation.  We should let it go stale only if the cost of keeping it valid is too high.

source/opt/strip_atomic_counter_memory_pass.cpp
```diff
@@ -0,0 +1,58 @@
...
+        continue;
+      }
+      mem_sem_val &= ~SpvMemorySemanticsAtomicCounterMemoryMask;
+
+      analysis::Integer int_type(32, false);
+      const analysis::Type* uint32_type =
+          type_mgr->GetRegisteredType(&int_type);
+      auto* new_const = constant_mgr->GetConstant(uint32_type, {mem_sem_val});
+      auto* new_const_inst = constant_mgr->GetDefiningInstruction(new_const);
+      auto new_const_id = new_const_inst->result_id();
```
s-perron: This code should keep the constant manager up-to-date.  The constant manager knows about the potentially new constant.

source/opt/strip_atomic_counter_memory_pass.h
s-perron: The constant manager will still be valid.  The def-use manager requires a single line to keep it up-to-date.  I would suggest that you do that.  It can make a noticeable difference in compile time if you need to rebuild it a lot.

source/opt/strip_atomic_counter_memory_pass.cpp
```diff
@@ -0,0 +1,58 @@
...
+      // The spec explicitly says that this id must be an OpConstant
+      auto mem_sem_val = mem_sem_inst->GetSingleWordOperand(2);
+      if (!(mem_sem_val & SpvMemorySemanticsAtomicCounterMemoryMask)) {
+        continue;
+      }
+      mem_sem_val &= ~SpvMemorySemanticsAtomicCounterMemoryMask;
+
+      analysis::Integer int_type(32, false);
+      const analysis::Type* uint32_type =
+          type_mgr->GetRegisteredType(&int_type);
```
s-perron: This potentially creates a new type.  The original value may have been a signed integer.  The type manager will still be valid because you get the manager to create the new type.

test/opt/optimizer_test.cpp
```diff
@@ -346,7 +348,95 @@ INSTANTIATE_TEST_SUITE_P(
...
+         "%_ptr_Workgroup_uint = OpTypePointer Workgroup %uint\n"
+         "%4 = OpVariable %_ptr_Workgroup_uint Workgroup\n"
+         "%void = OpTypeVoid\n"
+         "%10 = OpTypeFunction %void\n"
+         "%1 = OpFunction %void None %10\n"
+         "%11 = OpLabel\n"
+         "OpReturn\n"
+         "OpFunctionEnd\n",
+         "eliminate-dead-const"},
+        // Strip Atomic Counter Memory
```
s-perron: There should probably be more unit tests for this pass.  You should make sure the uint type gets created if it is not present.  You should also check different values for the memory semantics.  You only use one value.  Things may work correctly for that one value, but not necessarily others.  Just in case somebody incorrectly changes the code later.

zoddicus: I agree there should be more tests. I have started writing them, but haven't gotten them working completely, so I have not included them in the most recent push.

zoddicus:
I have added in proper test coverage. The build failure is due to something unrelated to my PR. Something using SpvCapabilitySubgroupAvcMotionEstimationINTEL which doesn't exist.

zoddicus:
Looks like everything is working correctly. @s-perron PTAAL

PR #2415, opened 2019-02-22
Non memory object as parameters.

s-perron:
In relaxed addressing mode, we want to accept non memory objects
because this is a very natural translation of hlsl.  It should be fixed
by legalization by inlining the calls.

A sample hlsl code that generates this code pattern is:

```
struct S {
  float fn_1() { return 0.0; }
};

struct T {
  S s;
};

float main() : A {
  T t;
  t.s.fn_1();
  return 0.0;
}
➜  dxc cat t.hlsl
struct S {
  float fn_1() { return 0.0; }
};

struct T {
  S s;
};

float main() : A {
  T t;
  t.s.fn_1();
  return 0.0;
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,7 +176,7 @@ if(NOT COMMAND find_host_program)
   endmacro()
 endif()
 
-find_host_package(PythonInterp)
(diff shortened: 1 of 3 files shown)


s-perron:
@ehsannas FYI.

PR #2417, opened 2019-02-25
Handle back edges better in dead branch elim.

### LINKED ISSUE #2391: Opt: optimized code fails to validate
Given the following glsl:

```glsl
// shader.frag
#version 310 es

precision highp float;
precision highp int;

layout(set = 0, binding = 0) uniform buf0 {
  vec2 injectionSwitch;
};

layout(location=0) out vec4 _GLF_color;

bool s()
{
    if(true)
    {
        return false;
    }
}

float h(out vec4 p)
{
   float h = 0.0;
   for(
       int i = 0;
       i < 1;
       i ++
   )
    {
     if(gl_FragCoord.y >= 0.0)
      {
       do {
           h += 1.0;
           if (injectionSwitch.x >= 0.0) {
               break;
           }
       } while(s());
      }
     
    }
 if(h > 0.0)
  {
   p = vec4(1.0);
  }
  return 1.0;
}

void main()
{
 vec4 p = vec4(1.0);
 h(p);
 _GLF_color = p;
}
```

I do the following to get optimized and unoptimized spirv, and their disassembly:
```sh
glslangValidator -V shader.frag -o shader.frag.spv && \
spirv-val shader.frag.spv && \
spirv-dis shader.frag.spv -o shader.frag.asm && \
spirv-opt -O shader.frag.spv -o shader.frag.opt.spv && \
spirv-dis shader.frag.opt.spv -o shader.frag.opt.asm && \
spirv-val shader.frag.opt.spv
```

The final command fails:
```sh
spirv-val shader.frag.opt.spv 
error: line 60: Loop header 98[%98] is targeted by 0 back-edge blocks but the standard requires exactly one
  %98 = OpLabel
```

One of the loop headers does not have a back-edge.

I attach the files.
[shaders.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/2866264/shaders.zip)

s-perron:
Loop header must have exactly one back edge.  Sometimes the branch
with the back edge can be folded.  However, it should not be folded
if it removes the back edge.

The code to check this simply avoids folding the branch in the
continue block.  That needs to be changed to not fold the back edge,
wherever it is.

At the same time, the branch can be folded if it folds to a branch to
the header, because the back edge will still exist.

Fixes #2391.

### DIFF
--- a/source/opt/dead_branch_elim_pass.h
+++ b/source/opt/dead_branch_elim_pass.h
@@ -148,9 +148,14 @@ class DeadBranchElimPass : public MemPass {
                                                uint32_t merge_block_id,
                                                uint32_t loop_merge_id,
                                                uint32_t loop_continue_id);
-  void AddBlocksWithBackEdge(uint32_t cont_id, uint32_t header_id,
-                             uint32_t merge_id,
-                             std::unordered_set<BasicBlock*>* continues);
+
+  // Adds to |blocks_with_back_edges| all of the blocks on the path from the
+  // basic block |cont_id| to |header_id| and |merge_id|.  The intention is that
+  // |cond_id| is a the continue target of a loop, |header_id| is the header of
+  // the loop, and |merge_id| is the merge block of the loop.
+  void AddBlocksWithBackEdge(
+      uint32_t cont_id, uint32_t header_id, uint32_t merge_id,
+      std::unordered_set<BasicBlock*>* blocks_with_back_edges);
 };
 
 }  // namespace opt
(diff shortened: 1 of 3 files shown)


source/opt/dead_branch_elim_pass.h
```diff
@@ -148,6 +148,9 @@ class DeadBranchElimPass : public MemPass {
                                                uint32_t merge_block_id,
                                                uint32_t loop_merge_id,
                                                uint32_t loop_continue_id);
+  void AddBlocksWithBackEdge(uint32_t cont_id, uint32_t header_id,
```
alan-baker: Header comment.

PR #2419, opened 2019-02-26
Bindless Validation: Descriptor Initialization Check

greg-lunarg:
If SPV_EXT_descriptor_indexing is enabled, add check that for every
descriptor-based reference, the descriptor is initialized. Initialization
data is stored in the debug input buffer, added to the length information
already there. This feature must be separately enabled on the pass
creation routine.

### DIFF
--- a/source/opt/inst_bindless_check_pass.cpp
+++ b/source/opt/inst_bindless_check_pass.cpp
@@ -40,21 +40,74 @@ uint32_t InstBindlessCheckPass::GenDebugReadLength(
   uint32_t desc_set_idx =
       var2desc_set_[var_id] + kDebugInputBindlessOffsetLengths;
   uint32_t desc_set_idx_id = builder->GetUintConstantId(desc_set_idx);
-  uint32_t desc_set_offset_id = GenDebugDirectRead(desc_set_idx_id, builder);
-  Instruction* binding_idx_inst =
-      builder->AddBinaryOp(GetUintId(), SpvOpIAdd, desc_set_offset_id,
-                           builder->GetUintConstantId(var2binding_[var_id]));
-  return GenDebugDirectRead(binding_idx_inst->result_id(), builder);
+  uint32_t binding_idx_id = builder->GetUintConstantId(var2binding_[var_id]);
+  return GenDebugDirectRead({desc_set_idx_id, binding_idx_id}, builder);
 }
 
-void InstBindlessCheckPass::GenBindlessCheckCode(
-    BasicBlock::iterator ref_inst_itr,
-    UptrVectorIterator<BasicBlock> ref_block_itr, uint32_t instruction_idx,
-    uint32_t stage_idx, std::vector<std::unique_ptr<BasicBlock>>* new_blocks) {
-  // Look for reference through bindless descriptor. If not, return.
-  std::unique_ptr<BasicBlock> new_blk_ptr;
-  uint32_t image_id;
-  switch (ref_inst_itr->opcode()) {
+uint32_t InstBindlessCheckPass::GenDebugReadInit(uint32_t var_id,
+                                                 uint32_t desc_idx_id,
+                                                 InstructionBuilder* builder) {
+  uint32_t desc_set_base_id =
+      builder->GetUintConstantId(kDebugInputBindlessInitOffset);
+  uint32_t desc_set_idx_id = builder->GetUintConstantId(var2desc_set_[var_id]);
+  uint32_t binding_idx_id = builder->GetUintConstantId(var2binding_[var_id]);
+  uint32_t u_desc_idx_id = GenUintCastCode(desc_idx_id, builder);
+  return GenDebugDirectRead(
+      {desc_set_base_id, desc_set_idx_id, binding_idx_id, u_desc_idx_id},
+      builder);
+}
+
+uint32_t InstBindlessCheckPass::CloneOriginalReference(
+    ref_analysis* ref, InstructionBuilder* builder) {
+  // Clone descriptor load
+  Instruction* load_inst = get_def_use_mgr()->GetDef(ref->load_id);
+  Instruction* new_load_inst =
+      builder->AddLoad(load_inst->type_id(),
+                       load_inst->GetSingleWordInOperand(kSpvLoadPtrIdInIdx));
+  uid2offset_[new_load_inst->unique_id()] = uid2offset_[load_inst->unique_id()];
+  uint32_t new_load_id = new_load_inst->result_id();
+  get_decoration_mgr()->CloneDecorations(load_inst->result_id(), new_load_id);
+  uint32_t new_image_id = new_load_id;
+  // Clone Image/SampledImage with new load, if needed
+  if (ref->image_id != 0) {
+    Instruction* image_inst = get_def_use_mgr()->GetDef(ref->image_id);
+    if (image_inst->opcode() == SpvOp::SpvOpSampledImage) {
+      Instruction* new_image_inst = builder->AddBinaryOp(
+          image_inst->type_id(), SpvOpSampledImage, new_load_id,
+          image_inst->GetSingleWordInOperand(kSpvSampledImageSamplerIdInIdx));
+      uid2offset_[new_image_inst->unique_id()] =
+          uid2offset_[image_inst->unique_id()];
+      new_image_id = new_image_inst->result_id();
+    } else {
+      assert(image_inst->opcode() == SpvOp::SpvOpImage && "expecting OpImage");
+      Instruction* new_image_inst =
+          builder->AddUnaryOp(image_inst->type_id(), SpvOpImage, new_load_id);
+      uid2offset_[new_image_inst->unique_id()] =
+          uid2offset_[image_inst->unique_id()];
+      new_image_id = new_image_inst->result_id();
+    }
+    get_decoration_mgr()->CloneDecorations(ref->image_id, new_image_id);
+  }
+  // Clone original reference using new image code
+  std::unique_ptr<Instruction> new_ref_inst(ref->ref_inst->Clone(context()));
+  uint32_t ref_result_id = ref->ref_inst->result_id();
+  uint32_t new_ref_id = 0;
+  if (ref_result_id != 0) {
+    new_ref_id = TakeNextId();
+    new_ref_inst->SetResultId(new_ref_id);
+  }
+  new_ref_inst->SetInOperand(kSpvImageSampleImageIdInIdx, {new_image_id});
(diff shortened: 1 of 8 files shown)


s-perron:
@greg-lunarg Sorry I did not notice this earlier.  I'll take a look at it tomorrow morning.

greg-lunarg:
No problem. Here is a high-level view of what I did...

Essentially, I added an instrumentation phase to the bindless check pass that generates checks for uninitialized descriptors just after the phase that gens bounds check for descriptor indexing.

Essentially, here is the sequence:

Before any instrumentation:

...
         Do sample on indexed descriptor
...

After bounds instrumentation:

...
        If index less than bound
           Do sample on indexed descriptor
        Else
           Write bounds error record
...

After Initialization instrumentation:

 ...
        If index less than bound
           If descriptor initialized
              Do sample on indexed descriptor
           Else
              Write initialization error record
        Else
           Write bounds error record
...

This had a few advantages. I only had to write logic for the simple if-then-else sequence, but by running it twice (once for bounds, once for initialization), I get the more complex nested logic which also shortcuts the initialization check if the index is out of bounds.

I thought about making the initialization instrumentation its own pass, but by putting it into the same class/pass/file as the bounds check code, I was more easily able to refactor quite a bit of code between the bounds and initialization instrumentation routines. There is probably more refactoring that could be done, possibly pushing more of this code into instrument_pass.*, but this seemed sufficient for the moment.

I also had to create a map from an instruction's unique id to its index in the original SPIR-V. That's because we want to report the original index of the instruction, but after the first instrumentation phase, the SPIR-V is (potentially) perturbed, and so the instruction's current index is likely no longer the same as it originally was. Eventually, this map could probably be pushed into the context, but I decided to leave that as a future exercise :)

greg-lunarg:
Another big reason for combining both bounds and initialization checking into the same pass/class is that they are targeting the same instructions, so we are able to leverage the instruction analysis and cloning code across both instrumentation phases. It probably makes sense that this code will stay in InstBindlessCheckPass.

source/opt/inst_bindless_check_pass.h
```diff
@@ -84,24 +89,88 @@ class InstBindlessCheckPass : public InstrumentPass {
...
+                            InstructionBuilder* builder);
+
+  // Clone original reference
+  uint32_t CloneOriginalReference(InstructionBuilder* builder);
+
+  // If |inst| references through a descriptor, return the id of the value it
+  // references through. Else return 0.
+  uint32_t GetDescriptorValueId(Instruction* inst);
+
+  // Analyze descriptor reference |ref_inst| and save components for later use.
```
s-perron: You should describe the return value.  I'm not sure what true or false is suppose to mean.

source/opt/inst_bindless_check_pass.h
```diff
@@ -84,24 +89,88 @@ class InstBindlessCheckPass : public InstrumentPass {
...
+  // Descriptor pointer
+  uint32_t ptr_id_;
+  Instruction* ptr_inst_;
+  // Base descriptor variable
+  uint32_t var_id_;
+  Instruction* var_inst_;
+  // Index into descriptor array
+  uint32_t index_id_;
+  // Original reference
+  Instruction* ref_inst_;
```
s-perron: I'm wondering if you really need all of these values.  When I look at `AnalyzeDescriptorReference`, it seems like many of these are readily available from others.  For example, `load_id_` is the result id of `load_inst_`.  Is there a good reason to keep both?

Also `ptr_id_` is used in just a single function.  Does it need to be a member variable?

`ptr_inst_` is used in just in one other place?

It might be worth writing a separate member function that retrieves these values by looking at `load_inst_` instead of storing all of them.  I don't think it will make much difference in the compile time, and it because easier to make sense of the state of the class being carried around.

source/opt/inst_bindless_check_pass.cpp
```diff
@@ -87,70 +138,144 @@ void InstBindlessCheckPass::GenBindlessCheckCode(
...
-    return;
+  ref_inst_ = ref_inst;
+  return true;
+}
+
+void InstBindlessCheckPass::GenCheckCode(
+    uint32_t check_id, uint32_t error_id, uint32_t length_id,
+    uint32_t stage_idx, InstructionBuilder* builder,
+    std::unique_ptr<BasicBlock>* new_blk_ptr,
+    std::vector<std::unique_ptr<BasicBlock>>* new_blocks) {
```
s-perron: I have a preference for keeping the interface of a function as simple as possible.  In this case, you pass in `new_blk_ptr` only to add it to `new_blocks`.  You do nothing else with the input value.  You seem to want to do a number of things in sequence.  Add all of the instructions to a block, and then add it to `new_blocks`.  You want `new_blk_ptr` to always contain the current block our are working on.

Why couldn't you add the new blocks to `new_blocks` as soon as they are created?  Use `new_blocks->back()` in places you want to access `new_blk_ptr`.  This removes a parameter, and there is one less variable to think about.

Also, you can create the InstructionBuilder because you have to block available, right?  I know constructing and destructing can take extra time, but I don't think it is too expensive, and will simplify the interface.

source/opt/inst_bindless_check_pass.h
s-perron: Now that I have looked at more code, I am wondering if these should just be their own struct that is returned by `AnalyzeDescriptorReference`.  It seems like `AnalyzeDescriptorReference` is called, and then the member are essentially dead at the end of the function that calls it.  This just gets confusing make it part of the state of the the pass.  I would have expected them to be used in more places.

greg-lunarg: OK. I have cleaned up the "redundant" members and have encapsulated the remaining members into a struct whose logical and storage lifetimes will match.

source/opt/inst_bindless_check_pass.h
```diff
@@ -46,32 +48,35 @@ class InstBindlessCheckPass : public InstrumentPass {
...
-  uint32_t GenDebugReadLength(uint32_t var_id, InstructionBuilder* builder);
-
-  // Initialize state for instrumenting bindless checking
-  void InitializeInstBindlessCheck();
-
-  // This function does bindless checking instrumentation on a single
-  // instruction. It is designed to be passed to
+  // These functions do bindless checking instrumentation on a single
+  // instruction which references through a descriptor. GenBoundsCheckCode
+  // checks that a index into a descriptor array is in-bounds. GenInitCheckCode
```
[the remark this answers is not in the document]
greg-lunarg: :)

source/opt/inst_bindless_check_pass.h
```diff
@@ -84,24 +89,88 @@ class InstBindlessCheckPass : public InstrumentPass {
...
+  // array |var_id| from debug input buffer and return id of value.
+  uint32_t GenDebugReadLength(uint32_t var_id, InstructionBuilder* builder);
+
+  // Generate instructions into |builder| to read initialization status of
+  // descriptor array |image_id| at |index_id| from debug input buffer and
+  // return id of value.
+  uint32_t GenDebugReadInit(uint32_t image_id, uint32_t index_id,
+                            InstructionBuilder* builder);
+
+  // Clone original reference
```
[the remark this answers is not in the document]
greg-lunarg: Done.

source/opt/inst_bindless_check_pass.h
```diff
@@ -84,24 +89,88 @@ class InstBindlessCheckPass : public InstrumentPass {
...
+  uint32_t CloneOriginalReference(InstructionBuilder* builder);
+
+  // If |inst| references through a descriptor, return the id of the value it
```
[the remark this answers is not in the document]
greg-lunarg: I think this is an apt idiom. Ultimately, this pass is about descriptors, a well-defined, first-class concept in Vulkan. While SPIR-V does refer to descriptor sets, it relegates the meaning of "descriptor" and other terms to the overarching API. I have added comments to try to explain how this concept maps to SPIR-V. I have augmented this term with "ie references to images or buffers" here and at the top of the module in the comments for the two main routines. I also added a comment at the top that refers the reader to Vulkan for further information on descriptors.

source/opt/inst_bindless_check_pass.cpp
```diff
@@ -40,21 +40,72 @@ uint32_t InstBindlessCheckPass::GenDebugReadLength(
...
+  new_ref_inst->SetInOperand(kSpvImageSampleImageIdInIdx, {new_image_id});
+  // Register new reference and add to new block
+  Instruction* added_inst = builder->AddInstruction(std::move(new_ref_inst));
+  uid2offset_[added_inst->unique_id()] = uid2offset_[ref_inst_->unique_id()];
+  if (new_ref_id != 0)
+    get_decoration_mgr()->CloneDecorations(ref_result_id, new_ref_id);
+  return new_ref_id;
+}
+
+uint32_t InstBindlessCheckPass::GetDescriptorValueId(Instruction* inst) {
```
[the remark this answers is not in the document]
greg-lunarg: Good catch. Support for buffers is not yet in. I have now added TODO's in several places to add support for buffers. I actually have a branch with buffer support which I will be submitting next (after I merge in all these changes :P

source/opt/inst_bindless_check_pass.h
greg-lunarg: Fixed as part of general updating of this comment in light of other changes.

source/opt/inst_bindless_check_pass.cpp
```diff
@@ -87,70 +138,144 @@ void InstBindlessCheckPass::GenBindlessCheckCode(
     case SpvOp::SpvOpImageSparseFetch:
     case SpvOp::SpvOpImageSparseRead:
     case SpvOp::SpvOpImageWrite:
-      image_id =
-          ref_inst_itr->GetSingleWordInOperand(kSpvImageSampleImageIdInIdx);
+      return inst->GetSingleWordInOperand(kSpvImageSampleImageIdInIdx);
       break;
```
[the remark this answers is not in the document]
greg-lunarg: Fixed.

source/opt/inst_bindless_check_pass.cpp
greg-lunarg: Done.

greg-lunarg:
CI-ubuntu-clang-asan is failing but not showing me the log :(

greg-lunarg:
Rebased to latest master. Hoping that will fix ubuntu-clang-asan...

greg-lunarg:
I think I have addressed all previous issues...

s-perron:
Thanks.  I will try to get somebody else to look at it, since I won't get a chance before April.

PR #2420, opened 2019-02-27
Allow NonWritable to target struct members.

ehsannas:
It should be allowed for the `NonWritable` decoration to be applied to
structure type members.

### DIFF
--- a/source/val/validate_decorations.cpp
+++ b/source/val/validate_decorations.cpp
@@ -1269,17 +1269,6 @@ spv_result_t CheckNonWritableDecoration(ValidationState_t& vstate,
              << "Target of NonWritable decoration is invalid: must point to a "
                 "storage image, uniform block, or storage buffer";
     }
-  } else {
-    // The target is a struct member.  The annotations pass already checks that
-    // it is a structure.  So now fall through to ensure that the structure is
-    // used as a UBO or SSBO.
-    const auto type_id = inst.id();
-    if (!vstate.IsStructForUniformBlock(type_id) &&
-        !vstate.IsStructForStorageBuffer(type_id)) {
-      return vstate.diag(SPV_ERROR_INVALID_ID, &inst)
-             << "Target of NonWritable member decoration is invalid: must be "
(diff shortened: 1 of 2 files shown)


source/val/validate_decorations.cpp
```diff
@@ -1269,17 +1269,6 @@ spv_result_t CheckNonWritableDecoration(ValidationState_t& vstate,
              << "Target of NonWritable decoration is invalid: must point to a "
                 "storage image, uniform block, or storage buffer";
     }
-  } else {
```
alan-baker: This could allowed the decoration on struct members if the struct gets used in uniform or storage buffer storage classes. What is the DXC case you think is unhandled?

ehsannas: IIUC this `else` clause is executed if we have an OpMemberDecorate on a structure.
e.g. 
```
OpMemberDecorate %struct 0 NonWritable
%struct = OpTypeStruct %float %int ...
```
Correct me if I'm wrong, but I believe this should be allowed regardless of storage class. `%struct` is a type and does not have a storage class.
Stepping through the code showed that `!vstate.IsStructForUniformBlock(type_id) && !vstate.IsStructForStorageBuffer(type_id)` was true. I believe `IsStructForUniformBlock` and `IsStructForStorageBuffer` are meant to be used for result-id of variables or function parameters.

DXC generates the following decorations and type definitions for ByteAddressBuffers:
```
OpName %type_ByteAddressBuffer "type.ByteAddressBuffer"
OpMemberDecorate %type_ByteAddressBuffer 0 Offset 0
OpMemberDecorate %type_ByteAddressBuffer 0 NonWritable
OpDecorate %type_ByteAddressBuffer BufferBlock
%_runtimearr_uint = OpTypeRuntimeArray %uint
%type_ByteAddressBuffer = OpTypeStruct %_runtimearr_uint
```

alan-baker: This particular decoration is under discussion in Khronos internal issue #408. I will file an issue to revisit this upon its resolution and withdraw my objection for now.

PR #2424, opened 2019-03-01
Add required tools.

s-perron:
We added a section to the readme to mention the tools that are needed to
build and test spirv-tools.  For the compiler, the compilers used by the
bots are mentioned.

The CMakeFile is changed to search specifically for python2, since we
have python scripts that cannot be run using python3.


s-perron:
Does not work for the Andriod NDK because they do not have python2 installed at all.  Will work on another solution.

Issue #2423, opened 2019-03-01
Handling Workgroup function parameter and elimination of load / store for not updated variables


jaebaek:
This is related to [an issue of DXC](https://github.com/Microsoft/DirectXShaderCompiler/issues/1977).

```
struct testStruct {
	float Data[1 + TILE_GROUP_SIZE_X + 1][1 + TILE_GROUP_SIZE_X + 1];
	float3x3 testFunc(uint topLeftX, uint topLeftY) { ... }
};

groupshared testStruct sharedStruct;

void testMain() {
    float3x3 result = sharedStruct.testFunc( GroupThreadID.x, GroupThreadID.y );
    ...
}
```

When DXC compiles the above code, the first parameter of `testFunc()` must be `groupshared testStruct sharedStruct;` which has `Workgroup` scope. Since `Function` scope pointer is legal for the function parameter, DXC creates a temporal variable whose scope is `Function` and copies `groupshared testStruct sharedStruct;` to the temporal variable.

After spirv-opt optimizes it, it generates the following code:

```
; SPIR-V
; Version: 1.0
; Generator: Google spiregg; 0
; Bound: 43
; Schema: 0
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %testMain "testMain" %gl_GlobalInvocationID %gl_LocalInvocationID %gl_WorkGroupID
               OpExecutionMode %testMain LocalSize 8 8 1
               OpSource HLSL 600
               OpName %testStruct "testStruct"
               OpMemberName %testStruct 0 "Data"
               OpName %sharedStruct "sharedStruct"
               OpName %type_RWStructuredBuffer_float "type.RWStructuredBuffer.float"
               OpName %testOut "testOut"
               OpName %testMain "testMain"
               OpDecorate %gl_GlobalInvocationID BuiltIn GlobalInvocationId
               OpDecorate %gl_LocalInvocationID BuiltIn LocalInvocationId
               OpDecorate %gl_WorkGroupID BuiltIn WorkgroupId
               OpDecorate %testOut DescriptorSet 0
               OpDecorate %testOut Binding 0
               OpDecorate %_runtimearr_float ArrayStride 4
               OpMemberDecorate %type_RWStructuredBuffer_float 0 Offset 0
               OpDecorate %type_RWStructuredBuffer_float BufferBlock
        %int = OpTypeInt 32 1
      %int_0 = OpConstant %int 0
[...]

PR #2428, opened 2019-03-06
Validator: no Storage comparison for pointer param

jaebaek:
DXC changed its policy and now, for variables with Workgroup Storage
Class, DXC passes it directly to functions instead of creating
temporal variables with Function Storage Class.
If relax-logical-pointer is enabled, Validator must consider
variables whose pointer types are the same except Storage Class
as variables with the same type.

Related to #2423

### DIFF
--- a/source/val/validate_function.cpp
+++ b/source/val/validate_function.cpp
@@ -24,6 +24,17 @@ namespace spvtools {
 namespace val {
 namespace {
 
+// Returns true if |a| and |b| are instruction defining pointers that point to
+// the same type.
+bool ArePointersToSameType(val::Instruction* a, val::Instruction* b) {
+  if (a->opcode() != SpvOpTypePointer || b->opcode() != SpvOpTypePointer) {
+    return false;
+  }
+
+  uint32_t a_type = a->GetOperandAs<uint32_t>(2);
+  return a_type && (a_type == b->GetOperandAs<uint32_t>(2));
+}
+
 spv_result_t ValidateFunction(ValidationState_t& _, const Instruction* inst) {
   const auto function_type_id = inst->GetOperandAs<uint32_t>(3);
   const auto function_type = _.FindDef(function_type_id);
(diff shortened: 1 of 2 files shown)


source/val/validate_function.cpp
```diff
@@ -24,6 +24,23 @@ namespace spvtools {
 namespace val {
 namespace {
 
+// When --relax-logical-pointer argument is given, we assume that types
+// of function parameter and argument are the same if they are both
+// OpTypePointer and those pointers point the same type regardless of
+// their storage class.
```
s-perron: This function should not care if the types are from function parameters or not.  The comments could be simpler: 

> Returns true if a| and |b| are instruction defining pointers that point to the same type.

test/val/val_storage_test.cpp
```diff
@@ -186,6 +186,34 @@ TEST_F(ValidateStorage, GenericVariableInsideFunction) {
...
+          OpFunctionEnd
+%func   = OpFunction %voidt None %vifunct
+%arg    = OpFunctionParameter %ptrt
+%funcl  = OpLabel
+          OpReturn
+          OpFunctionEnd
+)";
+  CompileSuccessfully(str);
+  getValidatorOptions()->relax_logical_pointer = true;
+  ASSERT_EQ(SPV_SUCCESS, ValidateInstructions());
```
s-perron: You should add a test that uses relax-logical-pointer and fails because the types are different.  Just to be sure you do not accept too much.

source/val/validate_function.cpp
```diff
@@ -24,6 +24,23 @@ namespace spvtools {
...
+bool IsRelaxedLogicalPointerSameForFuncParam(ValidationState_t& _,
```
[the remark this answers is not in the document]
jaebaek: done

source/val/validate_function.cpp
```diff
@@ -24,6 +24,23 @@ namespace spvtools {
...
+                                             val::Instruction* a,
+                                             val::Instruction* b) {
+  if (!_.options()->relax_logical_pointer) return false;
+
+  if (a->opcode() != SpvOpTypePointer || b->opcode() != SpvOpTypePointer)
+    return false;
+
+  const auto type_a = _.FindDef(a->GetOperandAs<uint32_t>(1));
+  const auto type_b = _.FindDef(b->GetOperandAs<uint32_t>(1));
+  return type_a && type_a->id() && type_b->id();
```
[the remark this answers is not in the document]
jaebaek: You are right. Updated it.

test/val/val_storage_test.cpp
jaebaek: done

source/val/validate_function.cpp
```diff
@@ -24,6 +24,17 @@ namespace spvtools {
 namespace val {
 namespace {
 
+// Returns true if a| and |b| are instruction defining pointers that point to
```
s-perron: s/if a|/if |a|/

jaebaek: done

Issue #2253, opened 2018-12-21
validation: OpDecorateStringGOOGLE can only be used with decorations that take string arguments 


dneto0:
For example, the following should fail validation:

```
               OpCapability Shader
               OpExtension "SPV_GOOGLE_hlsl_functionality1"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %1 "main"
               OpExecutionMode %1 LocalSize 1 1 1
               OpSource GLSL 450
               OpDecorateStringGOOGLE %void ArrayStride 4 ;  this is nonsense
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
          %1 = OpFunction %void None %3
          %4 = OpLabel
               OpReturn
               OpFunctionEnd
```

PR #2438, opened 2019-03-11
OpDecorateStringGOOGLE can only be used with decorations that take string arguments

sarahM0:
To fix: #2253

### DIFF
--- a/source/val/validate_annotation.cpp
+++ b/source/val/validate_annotation.cpp
@@ -209,7 +209,7 @@ spv_result_t ValidateDecorate(ValidationState_t& _, const Instruction* inst) {
   return SPV_SUCCESS;
 }
 
-spv_result_t ValidateecorateStringGOOGLE(ValidationState_t& _,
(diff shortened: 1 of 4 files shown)


source/val/validate_annotation.cpp
```diff
@@ -427,6 +444,9 @@ spv_result_t AnnotationPass(ValidationState_t& _, const Instruction* inst) {
     case SpvOpGroupMemberDecorate:
       if (auto error = ValidateGroupMemberDecorate(_, inst)) return error;
       break;
+    case SpvOpDecorateStringGOOGLE:
```
sarahM0: This MR does not work. When we get to this switch, always "ValidateDecorate" is the next line. Could you help me understand why "SpvOpDecorateStringGOOGLE" is never the case?

alan-baker: I'm not sure what you mean here. When I test this method it works as its supposed to.

dneto0:
Hey @sarahM0 this PR is left dangling. Please revisit it.

PR #2440, opened 2019-03-11
Add missing DepthGreater case to Fragment only check

### LINKED ISSUE #2439: DepthGreater outside of Fragment not caught by spirv-val
White writing tests for a different PR, I wrote a parameterized test that put in a bunch of different execution modes for a GLCompute shader. The validator correctly caught the other Depth* modes being Fragment only, but it didn't error on DepthGreater. Checking the spec, I am pretty sure this should have thrown the same type of error, so I am going to investigate if this case was missed.

zoddicus:
Fixes #2439


zoddicus:
The change added to test/val/val_modes_test.cpp would have caught this issue.

zoddicus:
Sent up a PR, https://github.com/KhronosGroup/glslang/pull/1724, to fix the smoketest issue

Issue #2450, opened 2019-03-14
spirv-opt --eliminate-dead-branches generates code that spirv-val rejects


afd:
[This archive contains files necessary to reproduce the problem](https://github.com/KhronosGroup/SPIRV-Tools/files/2964204/files_to_reproduce.zip)

Generating SPIR-V from variant_136_reduced_final.frag using glslangValidator, and then optimizing it with the --eliminate-dead-branches option leads to SPIR-V that the validator rejects with:

```
error: line 34: block <ID> 23 branches to the selection construct, but not to the selection header <ID> 8
  %23 = OpLabel
```

Issue found using [GraphicsFuzz](https://github.com/google/graphicsfuzz).

alan-baker:
The input to SPIR-V opt is invalid, though the validator doesn't catch it. The branch from 22 -> 9 is an illegal exit from the construct headed by 8.

![in](https://user-images.githubusercontent.com/33432579/54383865-6a461900-4669-11e9-8a8c-9470929f8129.png)

alan-baker:
Marking this as a validation issue.

afd:
Thanks.  In that case there must be another issue that led to spirv-opt (or glslangValidator) getting the SPIR-V into an invalid state.  I'll have a look at the chain of shaders that led to this.

PR #2459, opened 2019-03-15
Validate construct exits

### LINKED ISSUE #2213: eliminate-dead-branches leads to code that the validator rejects
[1696.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/2687205/1696.zip)

"command" in the attachment gives the command necessary to repro the bug.

Found using [GraphicsFuzz](https://github.com/google/graphicsfuzz).

alan-baker:
Fixes #2213, #2450

This change validate exits from a construct. Valid exits are described in the comment for `IsStructuredExit()`.

### DIFF
--- a/source/val/construct.h
+++ b/source/val/construct.h
@@ -23,6 +23,7 @@
 
 namespace spvtools {
 namespace val {
+class ValidationState_t;
 
 /// Functor for ordering BasicBlocks. BasicBlock pointers must not be null.
 struct less_than_id {
@@ -108,6 +109,22 @@ class Construct {
   // be called before the exit block is set and dominators have been
   // calculated.
   ConstructBlockSet blocks(Function* function) const;
+
+  // Returns true if |dest| is structured exit from the construct. Structured
+  // exits depend on the construct type.
+  // Selection:
+  //  * branch to the associated merge
+  //  * branch to the merge or continue of the innermost loop containing the
+  //  selection
+  // Loop:
+  //  * branch to the associated merge or continue
+  // Continue:
+  //  * back-edge to the associated loop header
+  //  * branch to the associated loop merge
+  //
+  // Note: the validator does not generate case constructs. Switches are
+  // checked separately from other constructs.
+  bool IsStructuredExit(ValidationState_t& _, BasicBlock* dest) const;
 
  private:
   /// The type of the construct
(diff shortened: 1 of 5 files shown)


source/val/construct.h
```diff
@@ -109,6 +110,23 @@ class Construct {
...
+  // Selection:
+  //  * branch to the associated merge
+  //  * branch to the merge or continue of the innermost loop containing the
+  //  selection
+  // Loop:
+  //  * branch to the associated merge or continue
+  // Continue:
+  //  * back-edge to the associated loop header
+  //  * branch to the associated loop merge
+  // Case:
```
dj2: This lists case but the comment above says case never gets here?

alan-baker: Yes, there is separate code to check cases. I'll update this comment.

s-perron:
Is this PR just waiting for a review, or is there something else?

alan-baker:
Review

source/val/validation_state.h
```diff
@@ -653,6 +653,19 @@ class ValidationState_t {
   spv_result_t CooperativeMatrixShapesMatch(const Instruction* inst,
                                             uint32_t m1, uint32_t m2);
 
+  // Returns true if the edge |source| -> |dest| is a structured exit of the
```
dneto0: This looks mysteriously different from the other descriptions of structured exits.

alan-baker: It also has no definition. Thanks for noticing. It was from an early implementation before it evolved into what it is now. I'm removing it.

dneto0: Yay for code review!

null77:
Note that seems to have regressed some ANGLE tests. Not sure who is at fault here. Possibly glslang?

https://ci.chromium.org/p/chromium/builders/ci/Linux%20FYI%20dEQP%20Release%20(NVIDIA)/68569

dEQP-GLES3.functional.shaders.switch.conditional_fall_through_2_uniform_vertex
../../third_party/angle/src/tests/deqp_support/angle_deqp_gtest.cpp:44: Failure
Failed
DebugUtilsMessenger(291): [ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] SPIR-V module not valid: block <ID> 68[%68] exits the selection headed by <ID> 44[%44], but not via a structured exit
  %68 = OpLabel

alan-baker:
Same as #2604. The spec misses language for allowing breaks from switches.

PR #2460, opened 2019-03-15
Change prioritization of unreachable merge and continue

### LINKED ISSUE #2452: spirv-opt (with various flags) generates code that spirv-val rejects: "Loop header is targeted by 0 back-edge blocks but the standard requires exactly one"
[Files to reproduce the problem](https://github.com/KhronosGroup/SPIRV-Tools/files/2968697/to_file.zip)

Doing:

spirv-opt --redundancy-elimination --copy-propagate-arrays --merge-blocks --ccp 
--simplify-instructions --convert-local-access-chains --eliminate-dead-code-aggr
essive --eliminate-local-single-block --if-conversion --convert-local-access-cha
ins --eliminate-dead-branches --merge-return --scalar-replacement=100 --eliminat
e-local-multi-store spirv_reduced.spv -o temp.spv

spirv-val temp.spv

should trigger the problem.

alan-baker:
Fixes #2452

* Swaps priority of handling unreachable merge and continues so that the
back-edge is retained in the case a block is both a loop continue and
loop merge
* New test

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -336,21 +336,7 @@ bool DeadBranchElimPass::EraseDeadBlocks(
     const std::unordered_map<BasicBlock*, BasicBlock*>& unreachable_continues) {
   bool modified = false;
   for (auto ebi = func->begin(); ebi != func->end();) {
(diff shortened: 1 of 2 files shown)


alan-baker:
Appveyor failure is problem uploading artifacts. All tests pass.

PR #2462, opened 2019-03-15
Add option to validate after each pass

alan-baker:
* New command-line option to opt: --validate-after-all
 * Pass manager will validate after each pass it runs

Errors look like:
```
error: line 109: ID 54[%54] defined in block 43[%43] does not dominate its use in block 68[%68]
  %68 = OpLabel

error: line 0: Validation failed after pass merge-return
```

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -199,6 +199,9 @@ class Optimizer {
   // |out| output stream.
   Optimizer& SetTimeReport(std::ostream* out);
 
+  // Sets the option to validate the module after each pass.
+  Optimizer& SetValidateAfterAll(bool validate);
+
  private:
   struct Impl;                  // Opaque struct for holding internal data.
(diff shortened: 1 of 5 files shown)


alan-baker:
Especially useful when combined with -print-all.

paulthomson:
Perhaps the name should be slightly different to clarify that it runs after each pass, not after all passes have run. E.g. `--validate-every-pass` `--validate-after-each` `--validate-all` `--validate-passes` But this might just be me.

alan-baker:
> Perhaps the name should be slightly different to clarify that it runs after each pass, not after all passes have run. E.g. `--validate-every-pass` `--validate-after-each` `--validate-all` `--validate-passes` But this might just be me.

It was named after the llvm option -print-after-all. I guess we changed that to -print-all, but this option doesn't control the initial validation. Maybe `--validate-passes` is a better name.

paulthomson:
Ah maybe `--validate-after-all` is fine then.

Issue #2107, opened 2018-11-23
Validate variable pointer container restrictions


alan-baker:
> A variable pointer with the Logical addressing model cannot
> * be an operand to an OpArrayLength instruction
> * point to an object that is or contains any OpTypeMatrix types

sarahM0:
By the rule "be an operand to an OpArrayLength instruction" do we mean such this test should fail? if not could you give an example that is invalid according this rule.

TEST_F(ValidateMemory, ArrayLenCorrectResultType) {
  std::string spirv = R"(
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %1 "main"
               OpExecutionMode %1 OriginUpperLeft
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
     %uint = OpTypeInt 32 0
%_runtimearr_float = OpTypeRuntimeArray %float
  %_struct_7 = OpTypeStruct %_runtimearr_float
%_ptr_Function__struct_7 = OpTypePointer Function %_struct_7
          %1 = OpFunction %void None %3
          %9 = OpLabel
         %10 = OpVariable %_ptr_Function__struct_7 Function
         %11 = OpArrayLength %uint %10 0
               OpReturn
               OpFunctionEnd

)";

  CompileSuccessfully(spirv.c_str());
  EXPECT_EQ(SPV_SUCCESS, ValidateInstructions());
}

alan-baker:
The test you quote does pass a variable pointer to the OpArrayLength instruction so it is fine. An invalid example is:
```
OpCapability Shader
OpCapability VariablePointersStorageBuffer
OpMemoryModel Logical GLSL450
OpEntryPoint GLCompute %main "main"
OpExecutionMode LocalSize 1 1 1
OpDecorate %block Block
OpMemberDecorate %block 0 Offset 0
%void = OpTypeVoid
%int = OpTypeInt 32 0
%block = OpTypeStruct %int
%ptr_ssbo_block = OpTypePointer StorageBuffer %block
%null = OpConstantNull %ptr_ssbo_block
%void_fn = OpTypeFunction %void
%main = OpFunction %void None %void_fn
%entry = OpLabel
%length = OpArrayLength %int %null 0
OpReturn
OpFunctionEnd
```

That is the simplest variable pointer you can generate.

sarahM0:
[...]

PR #2465, opened 2019-03-18
Variable pointers cannot be an operand to OpArrayLength

sarahM0:
To fix: #2107

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -1181,6 +1181,21 @@ spv_result_t ValidateArrayLength(ValidationState_t& state,
            << state.getIdName(inst->id()) << "' must be an OpTypeRuntimeArray.";
   }
 
+  const bool uses_variable_pointers =
+      state.features().variable_pointers ||
+      state.features().variable_pointers_storage_buffer;
+
+  if (!pointer_type ||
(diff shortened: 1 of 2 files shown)


source/val/validate_memory.cpp
```diff
@@ -1181,6 +1181,21 @@ spv_result_t ValidateArrayLength(ValidationState_t& state,
            << state.getIdName(inst->id()) << "' must be an OpTypeRuntimeArray.";
   }
 
+  const bool uses_variable_pointers =
+      state.features().variable_pointers ||
+      state.features().variable_pointers_storage_buffer;
+
+  if (!pointer_type ||
```
alan-baker: This doesn't capture the spec correctly. If variable pointers are not enabled you will accept OpAccessChain as the pointer operand, but not if they are enabled.

The check needs to be more subtle. https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#VariablePointer defines the roots and transitivity of variable pointers. You will have to trace the generation of the pointer operand to determine if it is a variable pointer or not.

s-perron:
@sarahM0 Are you still working on this?

sarahM0:
> @sarahM0 Are you still working on this?

I'm busy with Amber-Dawn so I won't work on it for a while. Feel free to close this.

PR #2466, opened 2019-03-19
Fix merge return in the face of breaks

### LINKED ISSUE #2453: spirv-opt generates code that spirv-val rejects: ID ... defined in block ... does not dominate its us e in block ...
[Files to reproduce problem](https://github.com/KhronosGroup/SPIRV-Tools/files/2968969/to_file2.zip)

Doing:

spirv-opt spirv_reduced.spv --inline-entry-points-exhaustive --private-to-local 
--ccp --merge-blocks --eliminate-dead-branches --eliminate-dead-branches --combi
ne-access-chains --reduce-load-size --redundancy-elimination --convert-local-acc
ess-chains --convert-local-access-chains --eliminate-dead-inserts --private-to-l
ocal --reduce-load-size --ccp --eliminate-dead-branches --merge-return --elimina
te-dead-inserts --eliminate-local-multi-store --inline-entry-points-exhaustive -
-eliminate-dead-branches --eliminate-dead-code-aggressive --eliminate-local-sing
le-store --eliminate-local-multi-store --reduce-load-size --inline-entry-points-
exhaustive --copy-propagate-arrays --if-conversion --simplify-instructions --ccp
 --eliminate-local-multi-store -o temp.spv
spirv-val temp.spv

should trigger the problem:

error: line 71: ID 144[%144] defined in block 142[%142] does not dominate its us
e in block 152[%152]
  %152 = OpLabel

Found using [GraphicsFuzz](http://github.com/google/graphicsfuzz)

alan-baker:
Fixes #2453

* Enable addition of OpPhi instructions when the loop has multiple
predecessors of the merge due to a break
 * This can result in some values no longer dominating their uses
* Track return blocks in structured flow to produce OpPhis that have
multiple undef and non-undef arguments
* New tests to catch the bug

### DIFF
--- a/source/opt/merge_return_pass.cpp
+++ b/source/opt/merge_return_pass.cpp
@@ -408,8 +408,14 @@ bool MergeReturnPass::BreakFromConstruct(
   // Forget about the edges leaving block.  They will be removed.
   cfg()->RemoveSuccessorEdges(block);
 
-  BasicBlock* old_body = block->SplitBasicBlock(context(), TakeNextId(), iter);
+  auto old_body_id = TakeNextId();
+  BasicBlock* old_body = block->SplitBasicBlock(context(), old_body_id, iter);
   predicated->insert(old_body);
+  // If a return block is being split, mark the new body block also as a return
+  // block.
+  if (return_blocks_.count(block->id())) {
+    return_blocks_.insert(old_body_id);
+  }
 
   // If |block| was a continue target for a loop |old_body| is now the correct
   // continue target.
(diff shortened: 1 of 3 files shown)


alan-baker:
Linker crashed, seems unrelated to my change.

source/opt/merge_return_pass.cpp
```diff
@@ -232,11 +233,19 @@ void MergeReturnPass::UpdatePhiNodes(BasicBlock* new_source,
...
   if (target_pred.size() == 1) {
     MarkForNewPhiNodes(target, context()->get_instr_block(target_pred[0]));
+  } else {
+    // If the loop contained a break and a return, OpPhi instructions may be
+    // required starting from the dominator of the loop merge.
+    DominatorAnalysis* dom_tree =
+        context()->GetDominatorAnalysis(target->GetParent());
+    auto idom = dom_tree->ImmediateDominator(target);
+    if (idom) {
+      MarkForNewPhiNodes(target, idom);
```
s-perron: If the header is available here, we could check it only if `idom` is not the header for the merge block.  If it is not available, then this is fine.  The search will be short anyway.

PR #2467, opened 2019-03-19
Check OpSampledImage is only passed into valid instructions

### LINKED ISSUE #1528: Storing result of OpSampledImage
From my understanding of the spec, the result id of OpSampledImage can only be consumed by an instruction specified to take the type OpTypeSampledImage. However I can currently store the result of OpSampledImage into a variable with no validation errors. I assume this is missing validation logic and not actually a valid instruction?

zoddicus:
Fixes #1528

### DIFF
--- a/source/val/validate_image.cpp
+++ b/source/val/validate_image.cpp
@@ -748,6 +748,33 @@ spv_result_t ValidateTypeSampledImage(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+bool IsAllowedSampledImageOperand(SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpSampledImage:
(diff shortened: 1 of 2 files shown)


source/val/validate_image.cpp
```diff
@@ -830,6 +830,36 @@ spv_result_t ValidateSampledImage(ValidationState_t& _,
                << "' as an operand of <id> '"
                << _.getIdName(consumer_instr->id()) << "'.";
       }
+
+      if (consumer_opcode != SpvOpSampledImage &&
```
alan-baker: Can you turn this into a function and write it as a switch statement?

zoddicus: Done

PR #2471, opened 2019-03-20
Handle dead infinite loops in DCE

### LINKED ISSUE #2456: spirv-opt generates code that spirv-val rejects: Case construct that targets ... has invalid branch to block ... (not another case construct, corresponding merge, outer loop merge or outer loop continue)
[Files to reproduce problem](https://github.com/KhronosGroup/SPIRV-Tools/files/2968972/to_file5.zip)

Doing:

spirv-opt spirv_reduced.spv --combine-access-chains --redundancy-elimination --combine-access-chains --combine-access-chains --reduce-load-size --eliminate-dead-branches --merge-return --eliminate-dead-inserts --eliminate-local-single-store --eliminate-local-single-block --eliminate-dead-code-aggressive --eliminate-dead-inserts --scalar-replacement=100 --eliminate-dead-branches --merge-return --eliminate-dead-inserts --redundancy-elimination --eliminate-dead-inserts --eliminate-dead-inserts --copy-propagate-arrays --eliminate-local-single-block --if-conversion --eliminate-dead-code-aggressive -o temp.spv

spirv-val temp.spv

should trigger the problem:

error: line 26: Case construct that targets 39[%39] has invalid branch to block 44[%44] (not another case construct, corresponding merge, outer loop merge or outer loop continue)
  %39 = OpLabel

Found using [GraphicsFuzz](http://github.com/google/graphicsfuzz)

alan-baker:
Fixes #2456

* When eliminating a structured construct that has an unreachable merge,
replace that unreachable terminator with an appropriate return
* New tests

This gets into a case of exposing undefined behaviour.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -513,6 +513,26 @@ bool AggressiveDCEPass::AggressiveDCE(Function* func) {
       AddBranch(mergeBlockId, *bi);
       for (++bi; (*bi)->id() != mergeBlockId; ++bi) {
       }
+
+      auto merge_terminator = (*bi)->terminator();
+      if (merge_terminator->opcode() == SpvOpUnreachable) {
+        // The merge was unreachable. This is undefined behaviour so just
+        // return (or return an undef). Then mark the new return as live.
+        auto func_ret_type_inst = get_def_use_mgr()->GetDef(func->type_id());
+        if (func_ret_type_inst->opcode() == SpvOpTypeVoid) {
+          merge_terminator->SetOpcode(SpvOpReturn);
+        } else {
+          // Find an undef for the return value and make sure it gets kept by
+          // the pass.
+          auto undef_id = Type2Undef(func->type_id());
+          auto undef = get_def_use_mgr()->GetDef(undef_id);
+          live_insts_.Set(undef->unique_id());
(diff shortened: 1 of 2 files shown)


source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -513,6 +513,26 @@ bool AggressiveDCEPass::AggressiveDCE(Function* func) {
...
+      if (merge_terminator->opcode() == SpvOpUnreachable) {
+        // The merge was unreachable. This is undefined behaviour so just
+        // return (or return an undef). Then mark the new return as live.
+        auto func_ret_type_inst = get_def_use_mgr()->GetDef(func->type_id());
+        if (func_ret_type_inst->opcode() == SpvOpTypeVoid) {
+          merge_terminator->SetOpcode(SpvOpReturn);
+        } else {
+          // Find an undef for the return value and make sure it gets kept by
+          // the pass.
+          auto undef_id = Type2Undef(func->type_id());
```
s-perron: The original code is likely to be an error, if it gets executed.  I'm wondering if it might be better to return an OpConstantNull instead of an Undef.  It might make debugging easier if this is ever hit.

s-perron: Alan pointed out that not every type can be used in an OpConstantNull, so we should stick with the undef.

PR #2472, opened 2019-03-21
Improve reducer algorithm and other changes

### LINKED ISSUE #2475: reduce: improve core algorithm
The core reducer algorithm has some quirks and minor bugs. It has missed out on some changes that were made to the (almost identical) GraphicsFuzz core reducer algorithm.

### LINKED ISSUE #2476: reduce: add --fail-on-reduction-error and ensure tests fail on a reduction error

paulthomson:
Fix #2475. Fix #2476. 

* Improve reducer algorithm: shrink granularity, remove an early return, no lazy initialization, notify pass if binary is interesting, add comments.
* Add fail-on-validation-error option to fail a reduction if an invalid state is reached; useful for tests.
* Set fail-on-validation-error in tests.
* Improve some documentation comments.
* Add Reducer::AddDefaultReductionPasses so tests (and other library consumers) can add the default reduction passes.
* Add CLIMessageConsumer in test_reduce so we can see messages for tricky tests.
* Remove test RemoveUnreferencedInstructionReductionPassTest_ApplyReduction because it was indirectly testing the reduction algorithm, not the RemoveUnreferencedInstruction pass.
* Tweak tests where needed.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -557,14 +557,17 @@ SPIRV_TOOLS_EXPORT spv_reducer_options spvReducerOptionsCreate();
 // Destroys the given reducer options object.
 SPIRV_TOOLS_EXPORT void spvReducerOptionsDestroy(spv_reducer_options options);
 
-// Records the maximum number of reduction steps that should run before the
-// reducer gives up.
+// Sets the maximum number of reduction steps that should run before the reducer
+// gives up.
 SPIRV_TOOLS_EXPORT void spvReducerOptionsSetStepLimit(
     spv_reducer_options options, uint32_t step_limit);
 
-// Sets seed for random number generation.
-SPIRV_TOOLS_EXPORT void spvReducerOptionsSetSeed(spv_reducer_options options,
-                                                 uint32_t seed);
+// Sets the fail-on-validation-error option; if true, the reducer will return
+// kStateInvalid if a reduction step yields a state that fails SPIR-V
+// validation. Otherwise, an invalid state is treated as uninteresting and the
+// reduction backtracks and continues.
+SPIRV_TOOLS_EXPORT void spvReducerOptionsSetFailOnValidationError(
+    spv_reducer_options options, bool fail_on_validation_error);
 
 // Encodes the given SPIR-V assembly text to its binary representation. The
 // length parameter specifies the number of bytes for text. Encoded binary will
--- a/source/reduce/reducer.cpp
+++ b/source/reduce/reducer.cpp
@@ -15,10 +15,17 @@
 #include <cassert>
 #include <sstream>
 
+#include "source/reduce/merge_blocks_reduction_opportunity_finder.h"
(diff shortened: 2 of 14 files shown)


include/spirv-tools/libspirv.h
```diff
@@ -557,14 +557,17 @@ SPIRV_TOOLS_EXPORT spv_reducer_options spvReducerOptionsCreate();
...
 SPIRV_TOOLS_EXPORT void spvReducerOptionsDestroy(spv_reducer_options options);
 
-// Records the maximum number of reduction steps that should run before the
-// reducer gives up.
+// Sets the maximum number of reduction steps that should run before the reducer
+// gives up.
 SPIRV_TOOLS_EXPORT void spvReducerOptionsSetStepLimit(
     spv_reducer_options options, uint32_t step_limit);
 
-// Sets seed for random number generation.
```
paulthomson: This was left in accidentally I think.

source/reduce/reducer.cpp
```diff
@@ -153,6 +168,25 @@ Reducer::ReductionResultStatus Reducer::Run(
   return Reducer::ReductionResultStatus::kComplete;
 }
 
+void Reducer::AddDefaultReductionPasses() {
```
paulthomson: As explained in PR description, I have added this so tests (and other library consumers) can get a reducer with the default reduction passes. Otherwise, this code was only in `tool/reduce.cpp`

test/reduce/reduce_test_util.cpp
```diff
@@ -68,5 +70,28 @@ void NopDiagnostic(spv_message_level_t /*level*/, const char* /*source*/,
                    const spv_position_t& /*position*/,
                    const char* /*message*/) {}
 
+
```
paulthomson: This is not used yet, but can be used when debugging tests and could be used in larger, more complex tests to ensure there is some information.

test/reduce/remove_unreferenced_instruction_test.cpp
```diff
@@ -83,149 +87,14 @@ TEST(RemoveUnreferencedInstructionReductionPassTest, RemoveStores) {
   ASSERT_TRUE(ops[1]->PreconditionHolds());
   ops[1]->TryToApply();
 
-  CheckEqual(env, expected, context.get());
-}
-
-TEST(RemoveUnreferencedInstructionReductionPassTest, ApplyReduction) {
```
paulthomson: This test was essentially testing the reducer algorithm, not the remove unreferenced instruction pass, so I removed it. I extended the test above slightly too, which is basically doing the same thing without testing the reducer algorithm.

include/spirv-tools/libspirv.h
afd: Good catch.

source/reduce/reducer.cpp
afd: Good idea.

test/reduce/remove_unreferenced_instruction_test.cpp
afd: Thanks.

PR #2473, opened 2019-03-21
Validate that OpUnreacahble is not statically reachable

alan-baker:
* Adds a validator check that ensures no block reachable from the entry
block is terminated by OpUnreachable
* Updated tests
* Added new tests

This came about from #2456. Adds a better error than what the validator caught for that case.


alegal-arm:
Which validation rule did this PR implement? I don't see anything in the spec that refers to  "statically (un)reachable". The CTS uses OpUnreachable for dynamically unreachable blocks.

alan-baker:
Dynamically unreachable blocks are still reachable in the CFG. OpUnreachable says it declares the block cannot be reached in the CFG.

PR #2481, opened 2019-03-27
Add pass to generate needed initializers for WebGPU

### LINKED ISSUE #2387: When converting to WebGPU create needed initializers
Variables of the storages classes: Output, Private, and Function, in WebGPU must have initializers. The simplest thing to do is add initializing to null for those that are currently missing. This will be correct, but not necessarily optimal.

zoddicus:
Fixes #2387

### DIFF
--- a/source/opt/generate_webgpu_initializers_pass.cpp
+++ b/source/opt/generate_webgpu_initializers_pass.cpp
@@ -40,7 +40,6 @@ bool NeedsWebGPUInitializer(Instruction* inst) {
 }  // namespace
 
 Pass::Status GenerateWebGPUInitializersPass::Process() {
-  auto constant_mgr = context()->get_constant_mgr();
   auto* module = context()->module();
   bool changed = false;
 
@@ -50,8 +49,7 @@ Pass::Status GenerateWebGPUInitializersPass::Process() {
     Instruction* inst = &(*iter);
 
     if (inst->opcode() == SpvOpConstantNull) {
-      auto* type = constant_mgr->GetType(inst);
-      null_constant_type_map_[type] = inst;
+      null_constant_type_map_[inst->type_id()] = inst;
       seen_null_constants_.insert(inst);
       continue;
     }
@@ -63,10 +61,8 @@ Pass::Status GenerateWebGPUInitializersPass::Process() {
     auto* constant_inst = GetNullConstantForVariable(inst);
     if (seen_null_constants_.find(constant_inst) ==
         seen_null_constants_.end()) {
-      constant_inst->RemoveFromList();
       constant_inst->InsertBefore(inst);
-      auto* type = constant_mgr->GetType(constant_inst);
-      null_constant_type_map_[type] = inst;
+      null_constant_type_map_[inst->type_id()] = inst;
       seen_null_constants_.insert(inst);
     }
     AddNullInitializerToVariable(constant_inst, inst);
@@ -92,18 +88,16 @@ Pass::Status GenerateWebGPUInitializersPass::Process() {
 Instruction* GenerateWebGPUInitializersPass::GetNullConstantForVariable(
     Instruction* variable_inst) {
   auto constant_mgr = context()->get_constant_mgr();
-  auto* constant_type = context()
-                            ->get_constant_mgr()
-                            ->GetType(variable_inst)
-                            ->AsPointer()
-                            ->pointee_type();
+  auto* def_use_mgr = get_def_use_mgr();
 
-  if (null_constant_type_map_.find(constant_type) ==
-      null_constant_type_map_.end()) {
+  auto* ptr_inst = def_use_mgr->GetDef(variable_inst->type_id());
+  auto type_id = ptr_inst->GetInOperand(1).words[0];
+  if (null_constant_type_map_.find(type_id) == null_constant_type_map_.end()) {
+    auto* constant_type = context()->get_type_mgr()->GetType(type_id);
     auto* constant = constant_mgr->GetConstant(constant_type, {});
     return constant_mgr->GetDefiningInstruction(constant);
   } else {
-    return null_constant_type_map_[constant_type];
+    return null_constant_type_map_[type_id];
   }
 }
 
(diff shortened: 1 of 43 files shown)


zoddicus:
@alan-baker  I looked into the necessity of re-running iteration due to reordering. You were correct, it is not needed, since the nodes know their next/prev and the operations will correctly update them. I have removed that logic from the PR.

source/opt/generate_webgpu_initializers_pass.cpp
```diff
@@ -0,0 +1,100 @@
...
+  // Handle global/module scoped variables
+  for (auto inst = module->types_values_begin();
+       inst != module->types_values_end(); ++inst) {
+    if (!NeedsWebGPUInitializer(inst)) continue;
+
+    changed = true;
+    AddNullInitializer(context(), inst,
+                       [](Module* inner_module, uint32_t constant_id,
+                          InstructionList::iterator inner_inst) {
+                         inner_module->EnsureIdDefinedBeforeInstruction(
```
alan-baker: This could be done without the extra iteration from the module function. If you track OpConstantNulls as you iterate (e.g. map type id to result id), you'll know whether the constant you generate is ordered correctly based on whether it's in the map or not. When you generate an unmapped constant immediately move before the current variable and map it.

source/opt/generate_webgpu_initializers_pass.cpp
```diff
@@ -0,0 +1,100 @@
...
+                             constant_id, inner_inst);
+                       });
+  }
+
+  // Handle local/function scoped variables
+  for (auto func = module->begin(); func != module->end(); ++func) {
+    for (auto block = func->begin(); block != func->end(); ++block) {
```
alan-baker: Function variables will be the first instructions in the first basic block of a function. If you iterate through the entry block, you can stop once you hit a non-variable instruction (only other valid instructions are debug and stored differently).

source/opt/module.cpp
```diff
@@ -172,6 +172,21 @@ uint32_t Module::GetExtInstImportId(const char* extstr) {
   return 0;
 }
 
+void Module::EnsureIdDefinedBeforeInstruction(uint32_t id, inst_iterator inst) {
```
alan-baker: If you make the other changes, this function should be unnecessary.

source/opt/generate_webgpu_initializers_pass.cpp
zoddicus: Done

source/opt/generate_webgpu_initializers_pass.cpp
zoddicus: Done

source/opt/module.cpp
zoddicus: Done

source/opt/generate_webgpu_initializers_pass.cpp
```diff
@@ -0,0 +1,119 @@
...
+  auto* module = context()->module();
+  bool changed = false;
+
+  // Handle global/module scoped variables
+  for (auto iter = module->types_values_begin();
+       iter != module->types_values_end(); ++iter) {
+    Instruction* inst = &(*iter);
+
+    if (inst->opcode() == SpvOpConstantNull) {
+      auto* type = constant_mgr->GetType(inst);
```
alan-baker: I think you want to use the type_id. You can have two types with the same type but different type ids.

source/opt/generate_webgpu_initializers_pass.cpp
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    if (!NeedsWebGPUInitializer(inst)) continue;
+
+    changed = true;
+
+    auto* constant_inst = GetNullConstantForVariable(inst);
+    if (seen_null_constants_.find(constant_inst) ==
+        seen_null_constants_.end()) {
+      constant_inst->RemoveFromList();
```
alan-baker: Can just you MoveBefore I think

source/opt/generate_webgpu_initializers_pass.cpp
```diff
@@ -0,0 +1,119 @@
...
+  } else {
+    return null_constant_type_map_[constant_type];
+  }
+}
+
+void GenerateWebGPUInitializersPass::AddNullInitializerToVariable(
+    Instruction* constant_inst, Instruction* variable_inst) {
+  auto constant_id = constant_inst->result_id();
+  variable_inst->AddOperand(Operand(SPV_OPERAND_TYPE_ID, {constant_id}));
+  context()->UpdateDefUse(variable_inst);
```
alan-baker: I think you only need to call `get_def_use_mgr()->AnalyzeInstUses(variable_inst)`.

source/opt/generate_webgpu_initializers_pass.cpp
zoddicus: Done

source/opt/generate_webgpu_initializers_pass.cpp
zoddicus: Done

source/opt/generate_webgpu_initializers_pass.cpp
```diff
@@ -92,27 +88,24 @@ Pass::Status GenerateWebGPUInitializersPass::Process() {
...
-
-  if (null_constant_type_map_.find(constant_type) ==
-      null_constant_type_map_.end()) {
+  auto* def_use_mgr = get_def_use_mgr();
+
+  auto* ptr_inst = def_use_mgr->GetDef(variable_inst->type_id());
+  auto type_id = ptr_inst->GetInOperand(1).words[0];
+  if (null_constant_type_map_.find(type_id) == null_constant_type_map_.end()) {
+    auto* constant_type = context()->get_type_mgr()->GetType(type_id);
     auto* constant = constant_mgr->GetConstant(constant_type, {});
```
alan-baker: I missed this before, but you should provide the type_id here (its an optional parameter).

test/opt/generate_webgpu_initializers_test.cpp
```diff
@@ -105,11 +105,12 @@ std::string GetInitializedGlobalVariableTestString(std::string storage_type) {
 TEST_P(GlobalVariableTest, Check) {
   std::string storage_class = std::get<0>(GetParam());
   bool changed = std::get<1>(GetParam());
-
   std::string input = GetUninitializedGlobalVariableTestString(storage_class);
   std::string expected =
       changed ? GetInitializedGlobalVariableTestString(storage_class) : input;
 
+  fprintf(stderr, "%s\n", input.c_str());
```
alan-baker: Left some debug in?

PR #2485, opened 2019-03-29
reduce: add remove_selection_reduction_opportunity

### LINKED ISSUE #2484: reduce: remove selections reduction pass

paulthomson:
Fix #2484

### DIFF
--- a/source/reduce/remove_selection_reduction_opportunity_finder.cpp
+++ b/source/reduce/remove_selection_reduction_opportunity_finder.cpp
@@ -60,8 +60,8 @@ RemoveSelectionReductionOpportunityFinder::GetAvailableOpportunities(
           if (CanOpSelectionMergeBeRemoved(
                   context, block, merge_instruction,
                   merge_and_continue_blocks_from_loops)) {
-            result.push_back(MakeUnique<RemoveSelectionReductionOpportunity>(
-                &function, &block));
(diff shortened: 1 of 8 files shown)


source/reduce/remove_selection_reduction_opportunity_finder.cpp
```diff
@@ -0,0 +1,148 @@
...
+  {
+    uint32_t merge_block_id =
+        merge_instruction->GetSingleWordOperand(kMergeNodeIndex);
+    for (uint32_t predecessor_block_id :
+         context->cfg()->preds(merge_block_id)) {
+      BasicBlock* predecessor_block =
+          context->cfg()->block(predecessor_block_id);
+      assert(predecessor_block);
+      bool found_divergent_successor = false;
+      predecessor_block->ForEachSuccessorLabel(
```
s-perron: The version of `ForEachSuccessorLabel` that takes a `uint32_t` as apposed to a `uint32_t*` is a const function.  For some reason VS2013 does not let you call it using a non-const pointer.  There are a couple options.  You could change `predecessor_block` to a const pointer, or change to the other version of `ForEachSuccessorLabel`.

paulthomson: Thanks!

PR #2490, opened 2019-04-02
Handle variable pointer in some optimizations

### LINKED ISSUE #2458: spirv-opt assertion "GetBaseAddress should only be called on instructions that take a pointer or image"
spirv-opt --skip-validation -O -o out bug.spv

source/opt/instruction.cpp:163: spvtools::opt::Instruction* spvtools::opt::Instruction::GetBaseAddress() const: Assertion `(IsLoad() || opcode() == SpvOpStore || opcode() == SpvOpAccessChain || opcode() == SpvOpPtrAccessChain || opcode() == SpvOpInBoundsAccessChain || opcode() == SpvOpCopyObject || opcode() == SpvOpImageTexelPointer) && "GetBaseAddress should only be called on instructions that take a " "pointer or image."' failed.

[bug.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/2969159/bug.txt)

s-perron:
* Check var ptr capability for common uniform.

* Check var ptr capability in access chain convert.

Since we want this pass to run even if there are variable pointer on
storage buffers, we had to remove asserts that assumed there were no
variable pointers.  The functions with the asserts will now work, it
becomes the responsibility of the callers to deal with the output as
appropriate.

* Single block elimination and variable pointers.

It seems like the code in local single block elimination is able to
handle cases with variable pointers already.  This is because the
function `HasOnlySupportedRefs` ensures that variables that feed a
variable pointer are not candidates.

* Single store elimination and variable pointers.

It seems like the code in local single stroe elimination is able to
handle cases with variable pointers already.  This is because the
function `FindSingleStoreAndCheckUses` ensures that variables that feed
a variable pointer are not candidates.

* SSA rewriter and variable pointers.

It seems like the code in the two passes that call the SSA rewriter are
able to  handle cases with variable pointers already.  This is because the
function `HasOnlySupportedRefs` ensures that variables that feed
a variable pointer are not candidates.

Fixes #2458.

### DIFF
--- a/test/opt/local_access_chain_convert_test.cpp
+++ b/test/opt/local_access_chain_convert_test.cpp
@@ -700,6 +700,126 @@ OpFunctionEnd
                                                      true);
 }
 
+TEST_F(LocalAccessChainConvertTest, VariablePointersStorageBuffer) {
+  // A case with a storage buffer variable pointer.  We should still convert
+  // the access chain on the function scope symbol.
+  const std::string test =
+      R"(
+; CHECK: OpFunction
+; CHECK: [[var:%\w+]] = OpVariable {{%\w+}} Function
+; CHECK: [[ld:%\w+]] = OpLoad {{%\w+}} [[var]]
+; CHECK: OpCompositeExtract {{%\w+}} [[ld]] 0 0
+               OpCapability Shader
+               OpCapability VariablePointersStorageBuffer
+          %1 = OpExtInstImport "GLSL.std.450"
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint GLCompute %2 "main"
+               OpExecutionMode %2 LocalSize 1 1 1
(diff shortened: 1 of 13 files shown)


test/opt/local_access_chain_convert_test.cpp
```diff
@@ -700,6 +700,126 @@ OpFunctionEnd
...
+               OpFunctionEnd
+)";
+
+  SinglePassRunAndMatch<LocalAccessChainConvertPass>(test, true);
+}
+
+TEST_F(LocalAccessChainConvertTest, VariablePointers) {
+  // A case with variable pointer capability.  We should not convert
+  // the access chain on the function scope symbol because the variable pointer
+  // could the analysis to miss references to function scope symbols.
```
alan-baker: missing a word here I think:

> could cause the analysis ...

PR #2493, opened 2019-04-04
Fix webgpu header file name in BUILD.gn

ShabbyX:
A .h file was specified as .hpp file, causing `gn gen path/ --check` to
fail.


dj2:
This looks good but you need to accept the CLA (which looks like you need to associate the email address of the commit with your github account to do?)

ShabbyX:
@dj2, sorry I committed with the wrong email address.

dj2:
Apparently merging and reviewing are different. LGTM, heh.

PR #2494, opened 2019-04-04
reduce: miscellaneous fixes

paulthomson:
Various little fixes. Best reviewed one commit at a time.


.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,57 @@
...
+      <option name="SPACE_BEFORE_WHILE_PARENTHESES" value="false" />
+      <option name="SPACE_BEFORE_FOR_PARENTHESES" value="false" />
+      <indentOptions>
+        <option name="INDENT_SIZE" value="2" />
+        <option name="CONTINUATION_INDENT_SIZE" value="2" />
+        <option name="TAB_SIZE" value="2" />
+      </indentOptions>
+    </codeStyleSettings>
+  </code_scheme>
+</component>
```
s-perron: Why are you adding the files in `.idea`.  I don't think they should be.

tools/reduce/reduce.cpp
```diff
@@ -26,6 +27,7 @@
 #include "tools/util/cli_consumer.h"
 
 using namespace spvtools::reduce;
+using namespace spvtools::opt;
```
s-perron: Why are you using the namespace here?

.idea/codeStyles/Project.xml
paulthomson: The plan was to add a few specific files (that are not user-specific) so that anybody using CLion will have some good code style settings.

tools/reduce/reduce.cpp
paulthomson: I missed this one. Thanks.

.idea/codeStyles/Project.xml
s-perron: I don't know if we want to start supporting specific IDE.  It will create extra work to maintain as these files change with different versions of the tools.  Also, which IDE do we want to support.  Why CLion and not another?  

We use the google style.  For me, clion seems to default to that style, or something close to it. 
 That is probably because I have a google-centric version.  You can also add clang-format as an external tool, which is what I do.  Then I can format the current file by pressing Ctrl-L (or whatever you want).

Also, we want one source of truth for automatic formatting code: clang-format.  What if we can't get the two tools to do the exact same thing.

I don't think it is a profitable for us to maintain these files.

paulthomson: Yes, I wanted to see what you thought. Note that CLion does indeed defer to clang format for me. These settings just define some naming conventions, which are not covered by clang format. I also wanted to add my "new file" template, but then the copyright header would be "Google LLC" by default, which doesn't seem fair, and CLion doesn't yet support a way to use a "company name" macro in file templates. Regardless, I will remove these.

PR #2496, opened 2019-04-04
Add WebGPU->Vulkan and Vulkan->WebGPU flags in spirv-opt

### LINKED ISSUE #2495: Add flag to spirv-opt for converting SPIR-V from WebGPU to Vulkan
Should be called '--webgpu-to-vulkan', and the existing '--webgpu-mode' should be renamed '--vulkan-to-webgu'

zoddicus:
Renames the existing flag '--webgpu-mode' to '--vulkan-to-webgpu' for
the Vulkan->WebGPU operation, and adds a new flag '--webgpu-to-vulkan'
for the WebGPU->Vulkan operation.

Currently '--webgpu-to-vulkan' doesn't have any passes associated with
it, but further patches will implement them.

Fixes #2495

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -101,10 +101,17 @@ class Optimizer {
   // from time to time.
   Optimizer& RegisterSizePasses();
 
-  // Registers passes that have been prescribed for WebGPU environments.
+  // Registers passes that have been prescribed for converting from Vulkan to
+  // WebGPU.
   // This sequence of passes is subject to constant review and will change
   // from time to time.
-  Optimizer& RegisterWebGPUPasses();
+  Optimizer& RegisterVulkanToWebGPUPasses();
+
+  // Registers passes that have been prescribed for converting from WebGPU to
+  // VulkanU.
+  // This sequence of passes is subject to constant review and will change
+  // from time to time.
+  Optimizer& RegisterWebGPUToVulkanPasses();
 
   // Registers passes that attempt to legalize the generated code.
   //
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -102,15 +102,13 @@ class Optimizer {
   Optimizer& RegisterSizePasses();
 
   // Registers passes that have been prescribed for converting from Vulkan to
-  // WebGPU.
-  // This sequence of passes is subject to constant review and will change
-  // from time to time.
+  // WebGPU. This sequence of passes is subject to constant review and will
+  // change from time to time.
   Optimizer& RegisterVulkanToWebGPUPasses();
 
   // Registers passes that have been prescribed for converting from WebGPU to
-  // VulkanU.
-  // This sequence of passes is subject to constant review and will change
-  // from time to time.
+  // Vulkan. This sequence of passes is subject to constant review and will
+  // change from time to time.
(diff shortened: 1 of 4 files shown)


include/spirv-tools/optimizer.hpp
```diff
@@ -101,10 +101,17 @@ class Optimizer {
   // from time to time.
   Optimizer& RegisterSizePasses();
 
-  // Registers passes that have been prescribed for WebGPU environments.
+  // Registers passes that have been prescribed for converting from Vulkan to
+  // WebGPU.
   // This sequence of passes is subject to constant review and will change
   // from time to time.
```
s-perron: Could you format the comments better?  The line breaks look awkward.

include/spirv-tools/optimizer.hpp
```diff
@@ -101,10 +101,17 @@ class Optimizer {
...
-  Optimizer& RegisterWebGPUPasses();
+  Optimizer& RegisterVulkanToWebGPUPasses();
+
+  // Registers passes that have been prescribed for converting from WebGPU to
+  // VulkanU.
+  // This sequence of passes is subject to constant review and will change
+  // from time to time.
```
s-perron: Same as the comments above.

tools/opt/opt.cpp
```diff
@@ -623,16 +718,38 @@ OptStatus ParseFlags(int argc, const char** argv,
           return {OPT_STOP, 1};
         }
         optimizer->SetTargetEnv(target_env);
-      } else if (0 == strcmp(cur_arg, "--webgpu-mode")) {
```
s-perron: We generally don't like removing an option.  Are we sure it is not being used yet?

include/spirv-tools/optimizer.hpp
```diff
@@ -101,10 +101,17 @@ class Optimizer {
   // from time to time.
   Optimizer& RegisterSizePasses();
 
-  // Registers passes that have been prescribed for WebGPU environments.
+  // Registers passes that have been prescribed for converting from Vulkan to
+  // WebGPU.
   // This sequence of passes is subject to constant review and will change
   // from time to time.
-  Optimizer& RegisterWebGPUPasses();
+  Optimizer& RegisterVulkanToWebGPUPasses();
```
s-perron: This is a change to the public interface?  Are we sure nothing depends on this yet?

include/spirv-tools/optimizer.hpp
zoddicus: Done

include/spirv-tools/optimizer.hpp
zoddicus: I am not aware of anything that depends on this API at the moment, other then the tests in the repo it isn't integrated with anything. In the future I think spvc in shaderc will be using it, but I checked that hasn't been implemented yet.

Its implementation is currently incomplete and hasn't really been advertised as a thing to use, since the spec it is implementing is still being worked on.

include/spirv-tools/optimizer.hpp
zoddicus: Done

tools/opt/opt.cpp
zoddicus: I am not aware of anything that depends on this flag at the moment, other then the tests in the repo it isn't integrated with anything to my knowledge. The current implementation is incomplete at the moment and the spec isn't finalized.

PR #2509, opened 2019-04-10
Add pass to legalize OpVectorShuffle for WebGPU

### LINKED ISSUE #2349: WebGPU transform: Replace 0xFFFFFFFF literal on OpVectorShuffle by a 0 literal
The WebGPU usage of SPIR-V bans component literal 0xFFFFFFFF.   SPIR-V core spec says that in the 0xFFFFFFFF case the selected value is an undef.  So it should be fine to pretend that we had been given literal 0 instead.

zoddicus:
In WebGPU, the component operand 0xFFFFFFFF is forbidden, but in
Vulkan it is used to indicate a value is undefined. When converting to
WebGPU, 0xFFFFFFFF needs to converted to a legal value, though the
specific one does not matter, since it was used to indicate an
undefined entry in the original code. Choosing to use 0, since the
operands are required to be on [0, N-1], so 0 is guaranteed to always
be valid.

Fixes #2349

### DIFF
--- a/source/opt/legalize_vector_shuffle_pass.cpp
+++ b/source/opt/legalize_vector_shuffle_pass.cpp
@@ -21,7 +21,7 @@ namespace opt {
 
 Pass::Status LegalizeVectorShufflePass::Process() {
   bool changed = false;
-  context()->module()->ForEachInst([this, &changed](Instruction* inst) {
+  context()->module()->ForEachInst([&changed](Instruction* inst) {
     if (inst->opcode() != SpvOpVectorShuffle) return;
 
     bool inner_changed = false;
@@ -32,9 +32,6 @@ Pass::Status LegalizeVectorShufflePass::Process() {
       inner_changed = true;
       inst->SetInOperand(idx, {0});
     }
-
-    if (!inner_changed) return;
-    context()->UpdateDefUse(inst);
   });
 
   return changed ? Status::SuccessWithChange : Status::SuccessWithoutChange;
--- a/source/opt/legalize_vector_shuffle_pass.cpp
+++ b/source/opt/legalize_vector_shuffle_pass.cpp
@@ -1,0 +1,44 @@
+// Copyright (c) 2019 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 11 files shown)


source/opt/legalize_vector_shuffle_pass.cpp
```diff
@@ -0,0 +1,44 @@
...
+    for (uint32_t idx = 2; idx < inst->NumInOperands(); ++idx) {
+      auto literal = inst->GetSingleWordInOperand(idx);
+      if (literal != 0xFFFFFFFF) continue;
+      changed = true;
+      inner_changed = true;
+      inst->SetInOperand(idx, {0});
+    }
+
+    if (!inner_changed) return;
+    context()->UpdateDefUse(inst);
```
s-perron: Since you are only  changing the literals, you do not need to do this.  As long as ids don't change, the def-use manager does not need to be updated.

source/opt/legalize_vector_shuffle_pass.h
```diff
@@ -0,0 +1,53 @@
...
+#include "source/opt/module.h"
+#include "source/opt/pass.h"
+
+namespace spvtools {
+namespace opt {
+
+// Converts any usages of 0xFFFFFFFF for the literals in OpVectorShuffle to a
+// literal 0. This is needed because using OxFFFFFFFF is forbidden by the WebGPU
+// spec. 0xFFFFFFFF in the main spec indicates that the result for this
+// component has not source, thus is undefined. Since this is undefined
```
s-perron: s/has not source/has no source/

source/opt/legalize_vector_shuffle_pass.h
```diff
@@ -0,0 +1,53 @@
...
+// behaviour we are free to use that value 0.
```
s-perron: s/use that value 0/use 0

source/opt/legalize_vector_shuffle_pass.cpp
zoddicus: Done

source/opt/legalize_vector_shuffle_pass.h
zoddicus: Done

source/opt/legalize_vector_shuffle_pass.h
zoddicus: Done

s-perron:
Sounds like a good plan.

zoddicus:
Created #2511 and #2512 to track optimize after all of the passes are implemented.

Issue #2388, opened 2019-02-13
When converting from WebGPU expand out initializers


zoddicus:
WebGPU requires initiaiizers for a number of storage classes when creating variables. Iniitializers are known to be problematic in some situations, so when converting from WebGPU to another environment , initializers should be unrolled into a creation instruction followed by an immediate store.

PR #2513, opened 2019-04-12
Implement WebGPU->Vulkan initializer conversion for 'Function' variables

zoddicus:
WebGPU requires certain variables to be initialized, whereas there are
known issues with using initializers in Vulkan. This PR is the first
of three implementing a pass to decompose initialized variables into
a variable declaration followed by a store. This has been broken up
into multiple PRs, because there 3 distinct cases that need to be
handled, which require seperate implementations.

This first PR implements the basic infrastructure that is needed, and
handling of Function storage class variables. Private and Output will
be handled in future PRs.

This is part of resolving #2388


BUILD.gn
```diff
@@ -484,6 +484,8 @@ static_library("spvtools_opt") {
     "source/opt/dead_insert_elim_pass.h",
     "source/opt/dead_variable_elimination.cpp",
     "source/opt/dead_variable_elimination.h",
+    "source/opt/decompose_webgpu_initalizers_pass.cpp",
```
alan-baker: Wrong file names.

source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -0,0 +1,80 @@
...
+  return true;
+}
+
+}  // namespace
+
+Pass::Status DecomposeInitializedVariablesPass::Process() {
+  auto* module = context()->module();
+  bool changed = false;
+
+  // TODO(zoddicus): Handle 'Output' variables
```
alan-baker: For the next PRs, instead of just the two storage classes WebGPU (and Vulkan) require, I would handle all globals the same way and let the validator reject other cases.

source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -0,0 +1,80 @@
...
+      iter->RemoveOperand(3);
+    }
+
+    while (new_stores.size()) {
+      auto* store = new_stores.front();
+      new_stores.pop();
+      context()->AnalyzeDefUse(store);
+      context()->set_instr_block(store, block);
+      block->AddInstruction(std::unique_ptr<Instruction>(store));
+      if (iter != block->end()) {
```
alan-baker: Is it possible for this to be end on a valid function? All blocks must have a terminator instruction.

source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -0,0 +1,80 @@
...
+      if (!HasInitializer(inst)) continue;
+
+      changed = true;
+      auto var_id = inst->result_id();
+      auto val_id = inst->GetOperand(3).words[0];
+      Instruction* store_inst = new Instruction(
+          context(), SpvOpStore, 0, 0,
+          {{SPV_OPERAND_TYPE_ID, {var_id}}, {SPV_OPERAND_TYPE_ID, {val_id}}});
+      new_stores.push(store_inst);
+      iter->RemoveOperand(3);
```
alan-baker: I think you're missing updating the uses in `iter`.

include/spirv-tools/optimizer.hpp
```diff
@@ -768,6 +768,10 @@ Optimizer::PassToken CreateFixStorageClassPass();
 // converts those literals to 0.
 Optimizer::PassToken CreateLegalizeVectorShufflePass();
 
+// Create a pass to decomposed initialized variables into a seperate variable
```
s-perron: s/decomposed/decompose

source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2019 Google Inc.
```
s-perron: I think it is suppose to be Google LLC now not Google Inc.

source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -0,0 +1,79 @@
...
+  auto* module = context()->module();
+  bool changed = false;
+
+  // TODO(zoddicus): Handle 'Output' variables
+  // TODO(zoddicus): Handle 'Private' variables
+
+  // Handle 'Function' variables
+  for (auto func = module->begin(); func != module->end(); ++func) {
+    auto block = func->entry().get();
+    std::queue<Instruction*> new_stores;
```
s-perron: We don't like using std::queue.  It has a really poor implementation on windows.  std::vector is essentially always better.

source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -0,0 +1,79 @@
...
+      iter->RemoveOperand(3);
+      get_def_use_mgr()->UpdateDefUse(&*iter);
+    }
+
+    while (new_stores.size()) {
+      auto* store = new_stores.front();
+      new_stores.pop();
+      context()->AnalyzeDefUse(store);
+      context()->set_instr_block(store, block);
+      block->AddInstruction(std::unique_ptr<Instruction>(store));
```
s-perron: I don't think this call is needed.  It will insert the instruction at the end of the block, but you are moving it immediately after.  Did you try without this?

source/opt/decompose_initialized_variables_pass.h
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2019 Google Inc.
```
s-perron: Google LLC

test/opt/decompose_initialized_variables_test.cpp
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2019 Google Inc.
```
s-perron: Google LLC

include/spirv-tools/optimizer.hpp
zoddicus: Done

source/opt/decompose_initialized_variables_pass.cpp
zoddicus: That looks correct, I apparently have been using the wrong one.

source/opt/decompose_initialized_variables_pass.cpp
zoddicus: Done

source/opt/decompose_initialized_variables_pass.h
zoddicus: Done

test/opt/decompose_initialized_variables_test.cpp
zoddicus: Done

source/opt/decompose_initialized_variables_pass.cpp
zoddicus: Done

PR #2519, opened 2019-04-15
Change validation of memory semantics for OpAtomics* in WebGPU

### LINKED ISSUE #2499: Update WebGPU validation rules to encode change to OpAtomic*
https://github.com/gpuweb/spirv-execution-env/pull/21 is going to restriction the valid orderings bits to Relaxed, so the validation code will need to updated to reflect that.

zoddicus:
Recent change to the spec restricted the valid values for Memory
Semnatics in OpAtomics* in the WebGPU env. Implementing enforcing
these changes.

Fixes #2499

### DIFF
--- a/source/val/validate_memory_semantics.cpp
+++ b/source/val/validate_memory_semantics.cpp
@@ -48,21 +48,32 @@ spv_result_t ValidateMemorySemantics(ValidationState_t& _,
   }
 
   if (spvIsWebGPUEnv(_.context()->target_env)) {
-    uint32_t valid_bits = SpvMemorySemanticsAcquireMask |
-                          SpvMemorySemanticsReleaseMask |
-                          SpvMemorySemanticsAcquireReleaseMask |
-                          SpvMemorySemanticsUniformMemoryMask |
+    uint32_t valid_bits = SpvMemorySemanticsUniformMemoryMask |
                           SpvMemorySemanticsWorkgroupMemoryMask |
                           SpvMemorySemanticsImageMemoryMask |
                           SpvMemorySemanticsOutputMemoryKHRMask |
                           SpvMemorySemanticsMakeAvailableKHRMask |
                           SpvMemorySemanticsMakeVisibleKHRMask;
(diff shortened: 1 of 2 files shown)


zoddicus:
This change implements the spec change in https://github.com/gpuweb/spirv-execution-env/pull/21

source/val/validate_memory_semantics.cpp
```diff
@@ -48,21 +48,32 @@ spv_result_t ValidateMemorySemantics(ValidationState_t& _,
...
-             << "WebGPU spec disallows any bit masks in Memory Semantics that "
-                "are not Acquire, Release, AcquireRelease, UniformMemory, "
-                "WorkgroupMemory, ImageMemory, OutputMemoryKHR, "
-                "MakeAvailableKHR, or MakeVisibleKHR";
+      if (spvOpcodeIsAtomicOp(inst->opcode())) {
+        return _.diag(SPV_ERROR_INVALID_DATA, inst)
+               << "WebGPU spec disallows, for OpAtomic*, any bit masks in "
+                  "Memory Semantics that are not UniformMemory, "
+                  "WorkgroupMemory, ImageMemory, OutputMemoryKHR, "
+                  "MakeAvailableKHR, or MakeVisibleKHR";
```
alan-baker: MakeAvailableKHR and MakeVisibleKHR require Acquire, Release or AcquireRelease semantics so they can't be used trasnsitively.

zoddicus: There are checks lower down that catch those cases with specific errors about that rule being violated.

alan-baker: I get that there are other checks, but it feels odd that this error message indicates those bits can be used even though it would never be valid.

zoddicus: Would it be acceptable to remove the text from the error message mentioning them, but still let it fall through the later check for the specific error?

alan-baker: Yes

zoddicus: Done

PR #2526, opened 2019-04-17
Remove Acquire and Release from allowed Mem Sem bits for WebGPU

### LINKED ISSUE #2524: Restrict ControlBarrier and MemoryBarrier to AcquireRelease for WebGPU
per https://github.com/gpuweb/spirv-execution-env/pull/24

zoddicus:
Fixes #2524

### DIFF
--- a/test/val/val_barriers_test.cpp
+++ b/test/val/val_barriers_test.cpp
@@ -258,13 +258,15 @@ OpControlBarrier %workgroup %workgroup %acquire_release_uniform_workgroup
   ASSERT_EQ(SPV_SUCCESS, ValidateInstructions(SPV_ENV_WEBGPU_0));
 }
 
-TEST_F(ValidateBarriers, OpControlBarrierWebGPURelaxedSuccess) {
+TEST_F(ValidateBarriers, OpControlBarrierWebGPURelaxedFailure) {
   const std::string body = R"(
 OpControlBarrier %workgroup %workgroup %uniform_workgroup
 )";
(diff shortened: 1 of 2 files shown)


zoddicus:
Failing pre-submit is a linker crash when building an unrelated test binary. I don't think it is related to my PR.

test/val/val_barriers_test.cpp
```diff
@@ -245,16 +249,48 @@ OpControlBarrier %workgroup %workgroup %acquire_release_uniform_workgroup
...
 OpControlBarrier %workgroup %workgroup %acquire_release_uniform_workgroup
 )";
 
   CompileSuccessfully(GenerateWebGPUShaderCode(body), SPV_ENV_WEBGPU_0);
   ASSERT_EQ(SPV_SUCCESS, ValidateInstructions(SPV_ENV_WEBGPU_0));
 }
 
+TEST_F(ValidateBarriers, OpControlBarrierWebGPURelaxedSuccess) {
+  const std::string body = R"(
+OpControlBarrier %workgroup %workgroup %uniform_workgroup
```
alan-baker: When I read the links WebGPU issues, it sounds to me like only AcquireRelease should be supported.

From the 2019.04.08 notes:
>Consensus to have acquire/release barriers only, no relaxed ones.

zoddicus: Thanks for the catch on that. I will update the PR.

PR #2529, opened 2019-04-18

afd:
This is a simple prototype of the bare bones of spirv-fuzz, featuring a few simple transformations.


source/fuzz/transformation.h
```diff
@@ -0,0 +1,84 @@
...
+// The base class for all metamorphic transformations.
+//
+// Rules for transformations
+// -------------------------
+//
+// - Immutability: a transformation must be immutable.
+// - Ability to copy and serialize: to ensure that a copy of a transformation,
+//     possibly saved out to disk and read back again, is indistinguishable
+//     from the original transformation, thus a transformation must not depend
+//     only on well-defined pieces of state, such as instruction ids.  It must
```
paulthomson: "must not depend only on"->"must depend only one"?

source/fuzz/transformation.h
```diff
@@ -0,0 +1,84 @@
...
+  // precise English.
+  virtual bool IsApplicable(opt::IRContext* context) = 0;
+
+  // Requires that the transformation is applicable.  Applies the
+  // transformation, mutating the given SPIR-V module.
+  virtual void Apply(opt::IRContext* context) = 0;
+
+  // Returns true if and only if the module does not currently define the given
+  // id.
+  bool IsFreshId(opt::IRContext* context, uint32_t id);
```
paulthomson: Should `IsFreshId` and `UpdateModuleIdBound` be here? Should they be static, protected, or maybe even global functions, or in some other "utilities" class? I feel like the `Transformation` class is otherwise a pure interface, which I like.

source/fuzz/transformation.h
afd: Fixed.

source/fuzz/transformation.h
afd: I have opted for the global functions approach, creating a fuzzer_util.h/cpp pair.

source/fuzz/transformation_split_block.h
```diff
@@ -0,0 +1,69 @@
...
+
+ private:
+  // The result id of an instruction.
+  const uint32_t result_id_;
+  // An offset, such that the block containing |result_id_| should be split
+  // right before the instruction |offset_| instructions after |result_id_|.
+  const uint32_t offset_;
+  // An id that must not yet be used by the module to which this transformation
+  // is applied.
+  const uint32_t fresh_id_;
```
alan-baker: For the ids, consider whether you want the fuzzer to set a new id bound in the IRContext (the bound + 100) and then you could just use IRContext::TakeNextId() and not have to perform the checks that the id is still fresh in the transforms.

tools/fuzz/fuzz.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+The SPIR-V binary is read from <input>.
+
+NOTE: The fuzzer is a work in progress.
+
+Options (in lexicographical order):
+
+  -h, --help
+               Print this help.
+  --seed       Unsigned 32-bit integer seed to control random number
```
alan-baker: The descriptions are inconsistent about which line they should be on relative to the option

source/fuzz/fuzzer.cpp
```diff
@@ -0,0 +1,105 @@
...
+  // Make a PRNG, either from a given seed or from a random device.
+  PseudoRandomGenerator random_generator(
+      options->has_random_seed ? options->random_seed
+                               : (uint32_t)std::random_device()());
+
+  // The fuzzer will introduce new ids into the module.  The module's id bound
+  // gives the smallest id that can be used for this purpose.  We add an offset
+  // to this so that there is a sizeable gap between the ids used in the
+  // original module and the ids used for fuzzing, as a readability aid.
+  auto minimum_fresh_id = ir_context->module()->id_bound() + 100;
```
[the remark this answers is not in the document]
afd: I have filed issue #2541 for this and mentioned it in TODOs in the code, here and where we update the module id bound.

source/fuzz/fuzzer_pass_permute_blocks.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+  for (auto& function : *ir_context->module()) {
+    std::vector<uint32_t> block_ids;
+    // Collect all block ids for the function before messing with block
+    // ordering.
+    for (auto& block : function) {
+      block_ids.push_back(block.id());
+    }
+    // Now consider each block id.
+    for (auto id : block_ids) {
```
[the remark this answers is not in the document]
afd: Done, and added your example as a comment to explain the reason.

source/fuzz/transformation_split_block.h
afd: The reason for the current design is that future transformations will depend on the ids introduced by prior transformations, and we want to maximise the chances that future transformations are still applicable when we remove some prior transformations, to enable fuzzed shader reduction to work by re-generating using subsets of transformations.

I have added the following comment to clarify.  As this will probably apply elsewhere in future, I will consider a more prominent place to signpost it as a guideline for anyone adding a new transformation that will use fresh ids.

  // An id that must not yet be used by the module to which this transformation
  // is applied.  Rather than having the transformation choose a suitable id on
  // application, we require the id to be given upfront in order to facilitate
  // reducing fuzzed shaders by removing transformations.  The reason is that
  // future transformations may refer to the fresh id introduced by this
  // transformation, and if we end up changing what that id is, due to removing
  // earlier transformations, it may inhibit later transformations from
  // applying.
  const uint32_t fresh_id_;

tools/fuzz/fuzz.cpp
afd: Fixed, thanks.

source/fuzz/transformation_add_dead_break.cpp
```diff
@@ -0,0 +1,289 @@
...
+    // |from_block_| is not in a construct from which we can break.
+    return false;
+  }
+
+  // Consider case (2)
+  if (to_block_ ==
+      context->cfg()->block(containing_construct)->MergeBlockId()) {
+    // This looks like an instance of case (2).
+    // However, the structured CFG analysis regards the continue construct of a
+    // loop as part of the loop, but it is not legal to jump from a loop's
```
alan-baker: It is legal, but only from the back-edge block. I'm not sure if this is a worthwhile distinction here.

source/fuzz/transformation_add_dead_break.cpp
```diff
@@ -0,0 +1,289 @@
...
+      return false;
+    }
+    if (phi_extension->type_id() != inst.type_id()) {
+      // The instruction given to extend this OpPhi either does not have a type
+      // or its type does not match that of the OpPhi.
+      return false;
+    }
+    // Check whether the given id corresponds to a constant.  If so, that's
+    // fine.
+    bool found = false;
```
alan-baker: Instead of searching for |phi_extension| in all module scope values, I'd suggest getting the basic block that defines it. If it has no block, then it should be module scope and safe. Otherwise, you can verify that that basic block dominates |bb_from|.

source/fuzz/transformation_split_block.cpp
```diff
@@ -0,0 +1,152 @@
...
+  // (2) the transformation wants to split at some offset from a
+  //     non-label instruction inside the block.
+  // (3) the split associated with this transformation has nothing to do with
+  //     this block
+  if (result_id_ == block->id()) {
+    // Case (1).
+    if (offset_ == 0) {
+      // The offset is not allowed to be 0: this would mean splitting before the
+      // block's label.
+      return {true, block->end()};
```
alan-baker: This is a confusing result for the comment. Shouldn't it return {false, block-end()}?

source/fuzz/transformation_add_dead_break.cpp
afd: It would be good to allow this case eventually; I have filed #2577 to log it, and added a TODO(2577) comment.

source/fuzz/transformation_split_block.cpp
afd: I have clarified the reason as follows:

      // The offset is not allowed to be 0: this would mean splitting before the
      // block's label.
      // By returning (true, block->end()), we indicate that we did find the
      // instruction (so that it is not worth searching further for it), but
      // that splitting will not be possible.

and also there is a comment at the top of the method:

// Returns:
// - (true, block->end()) if the relevant instruction is in this block
//      but inapplicable
// - (true, it) if 'it' is an iterator for the relevant instruction
// - (false, _) otherwise.

source/fuzz/transformation_add_dead_break.cpp
afd: Done.

afd:
I am abandoning this PR as it has been superseded by #2581, which incorporates all the review comments associated with this PR.

PR #2536, opened 2019-04-23
Add validation specific to OpExecutionModeId

### LINKED ISSUE #1565: Validator should check that OpExecutionModeId only accepts SubgroupsPerWorkgroupId, LocalSizeId, and LocalSizeHintId

zoddicus:
Fixes #1565

### DIFF
--- a/source/val/validate_mode_setting.cpp
+++ b/source/val/validate_mode_setting.cpp
@@ -237,33 +237,25 @@ spv_result_t ValidateExecutionMode(ValidationState_t& _,
 
   const auto mode = inst->GetOperandAs<SpvExecutionMode>(1);
   if (inst->opcode() == SpvOpExecutionModeId) {
-    switch (mode) {
-      case SpvExecutionModeSubgroupsPerWorkgroupId:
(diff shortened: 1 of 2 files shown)


source/val/validate_mode_setting.cpp
```diff
@@ -236,6 +236,44 @@ spv_result_t ValidateExecutionMode(ValidationState_t& _,
...
+      case SpvExecutionModeLocalSizeHintId: {
+        const auto operand_id = inst->GetOperandAs<uint32_t>(2);
+        const auto* operand_inst = _.FindDef(operand_id);
+        if (!spvOpcodeIsConstant(operand_inst->opcode())) {
+          return _.diag(SPV_ERROR_INVALID_ID, inst)
+                 << "For OpExecutionModeId Extra Operand id must be "
+                    "constant instruction.";
+        }
+      } break;
+      case SpvExecutionModeLocalSizeId: {
```
alan-baker: I'd just combine this with the above and loop through all operands starting from index 2. We don't check that the instruction has the right number of operands here.

zoddicus: Done

PR #2537, opened 2019-04-24
Add support for Private & Output to initializer decompose flag

### LINKED ISSUE #2388: When converting from WebGPU expand out initializers

zoddicus:
Fixes #2388

### DIFF
--- a/source/opt/decompose_initialized_variables_pass.cpp
+++ b/source/opt/decompose_initialized_variables_pass.cpp
@@ -62,7 +62,7 @@ Pass::Status DecomposeInitializedVariablesPass::Process() {
     for (auto iter = first_block->begin();
          iter != first_block->end() && iter->opcode() == SpvOpVariable;
          ++iter) {
-      // For valid SPIRV-V, there is guaranteed to be atleast one instruction
+      // For valid SPIRV-V, there is guaranteed to be at least one instruction
       // after the OpVariable instructions.
       insert_point = (*iter).NextNode();
       Instruction* inst = &(*iter);
--- a/source/opt/decompose_initialized_variables_pass.cpp
+++ b/source/opt/decompose_initialized_variables_pass.cpp
@@ -34,7 +34,7 @@ bool HasInitializer(Instruction* inst) {
 
 Pass::Status DecomposeInitializedVariablesPass::Process() {
   auto* module = context()->module();
-  bool changed = false;
+  std::unordered_set<Instruction*> changed;
 
   std::vector<std::tuple<uint32_t, uint32_t>> global_stores;
   for (auto iter = module->types_values_begin();
@@ -42,11 +42,11 @@ Pass::Status DecomposeInitializedVariablesPass::Process() {
     Instruction* inst = &(*iter);
     if (!HasInitializer(inst)) continue;
 
-    changed = true;
     auto var_id = inst->result_id();
     auto val_id = inst->GetOperand(3).words[0];
     global_stores.push_back(std::make_tuple(var_id, val_id));
     iter->RemoveOperand(3);
+    changed.insert(&*iter);
   }
 
   std::unordered_set<uint32_t> entry_ids;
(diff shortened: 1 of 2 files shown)


source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -57,20 +72,41 @@ Pass::Status DecomposeInitializedVariablesPass::Process() {
...
+        uint32_t var_id;
+        uint32_t val_id;
+        std::tie(var_id, val_id) = store_ids;
+        auto* store_inst = new Instruction(
+            context(), SpvOpStore, 0, 0,
+            {{SPV_OPERAND_TYPE_ID, {var_id}}, {SPV_OPERAND_TYPE_ID, {val_id}}});
+        context()->AnalyzeDefUse(store_inst);
+        context()->set_instr_block(store_inst, &*first_block);
+        first_block->AddInstruction(std::unique_ptr<Instruction>(store_inst));
+        if (last_var == first_block->end()) {
```
s-perron: You can probably simplify a lot of the code related to `last_var` if you changed it from pointing to the last variable to point to the first instruction after the last variable.  That instruction will always exist, and then there is not special case for when there are no variable.  You can just always insert before.

source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -36,17 +36,32 @@ Pass::Status DecomposeInitializedVariablesPass::Process() {
...
+  for (auto iter = module->types_values_begin();
+       iter != module->types_values_end(); ++iter) {
+    Instruction* inst = &(*iter);
+    if (!HasInitializer(inst)) continue;
+
+    changed = true;
+    auto var_id = inst->result_id();
+    auto val_id = inst->GetOperand(3).words[0];
+    global_stores.push_back(std::make_tuple(var_id, val_id));
+    iter->RemoveOperand(3);
```
s-perron: Does this mess up the def-use manager?  You remove the operand which is the use of an id.

source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -57,20 +72,41 @@ Pass::Status DecomposeInitializedVariablesPass::Process() {
       Instruction* store_inst = new Instruction(
           context(), SpvOpStore, 0, 0,
           {{SPV_OPERAND_TYPE_ID, {var_id}}, {SPV_OPERAND_TYPE_ID, {val_id}}});
-      new_stores.push_back(store_inst);
+      function_stores.push_back(store_inst);
       iter->RemoveOperand(3);
```
s-perron: Does this mess up the def-use manager as well?

zoddicus: Yup, fixed it.

source/opt/decompose_initialized_variables_pass.cpp
zoddicus: Yup, fixed it.

source/opt/decompose_initialized_variables_pass.cpp
```diff
@@ -34,43 +34,78 @@ bool HasInitializer(Instruction* inst) {
...
-    for (auto iter = block->begin();
-         iter != block->end() && iter->opcode() == SpvOpVariable; ++iter) {
-      last_var = iter;
+    std::vector<Instruction*> function_stores;
+    auto first_block = func->entry().get();
+    inst_iterator insert_point = first_block->begin();
+    for (auto iter = first_block->begin();
+         iter != first_block->end() && iter->opcode() == SpvOpVariable;
+         ++iter) {
+      // For valid SPIRV-V, there is guaranteed to be atleast one instruction
```
s-perron: "atleast" should be two words.

zoddicus: Done

PR #2540, opened 2019-04-26
fix example.cpp

### LINKED ISSUE #2538: example won't run
showing the error:
```
error: No OpEntryPoint instruction was found. This is only allowed if the Linkage capability is being used.
```

according to the [specific](https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#_universal_validation_rules)

> There is at least one OpEntryPoint instruction, unless the Linkage capability is being used.

I think we could get the example update

clouds56:
this fixes #2538

### DIFF
--- a/examples/cpp-interface/main.cpp
+++ b/examples/cpp-interface/main.cpp
@@ -28,6 +28,7 @@
 
 int main() {
   const std::string source =
+      "         OpCapability Linkage "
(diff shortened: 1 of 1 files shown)


clouds56:
I'm not sure if this is the way to go, maybe we could construct an proper vulkan example.

PR #2542, opened 2019-04-26
Remove stale comment

### LINKED ISSUE #1121: OpEntryPoint: shader entry points have void(void) signature
Currently we check that it has no arguments.  We should also check the return type is void

zoddicus:
Fixes #1121


zoddicus:
The issue listed (#1121) is about making sure that shader entry point functions are void(). My understanding reading the existing code is that this is already being checked, just the comments are a bit out of date. I am just updating the comments and closing out the issue with this PR.

Issue #2535, opened 2019-04-23
HLSL legalization for isomorphic function parameter


jaebaek:
It would be great if HLSL legalization handled function parameter passing where the argument is actually isomorphic to the parameter but their types are not the same.

For the following HLSL code,

```
struct R {
  int a;
  void incr() { ++a; }
};

void decr(inout R b) {
  b.a--;
}

RWStructuredBuffer<R> rwsb;

void main() {
  rwsb[0].incr();
  decr(rwsb[0]);
}
```

DXC generates the following SPIR-V code for calling `rwsb[0].incr();` and `decr(rwsb[0]);`
```
               OpMemberDecorate %R 0 Offset 0
...
          %R = OpTypeStruct %int
        %R_0 = OpTypeStruct %int
%_ptr_Function_R_0 = OpTypePointer Function %R_0
%_ptr_Uniform_R = OpTypePointer Uniform %R
       %rwsb = OpVariable %_ptr_Uniform_type_RWStructuredBuffer_R Uniform
...
 %temp_var_R = OpVariable %_ptr_Function_R_0 Function
%param_var_b = OpVariable %_ptr_Function_R_0 Function
         %26 = OpAccessChain %_ptr_Uniform_R %rwsb %int_0 %uint_0
         %27 = OpLoad %R %26
         %28 = OpCompositeExtract %int %27 0
         %29 = OpCompositeConstruct %R_0 %28
               OpStore %temp_var_R %29
         %30 = OpFunctionCall %void %R_incr %temp_var_R
         %32 = OpLoad %R_0 %temp_var_R
         %33 = OpCompositeExtract %int %32 0
         %34 = OpCompositeConstruct %R %33
               OpStore %26 %34
         %35 = OpAccessChain %_ptr_Uniform_R %rwsb %int_0 %uint_0
         %36 = OpLoad %R %35
         %37 = OpCompositeExtract %int %36 0
         %38 = OpCompositeConstruct %R_0 %37
               OpStore %param_var_b %38
         %39 = OpFunctionCall %void %decr %param_var_b
         %41 = OpLoad %R_0 %param_var_b
         %42 = OpCompositeExtract %int %41 0
         %43 = OpCompositeConstruct %R %42
               OpStore %35 %43
```

After fixing #2430, the reason why we cannot directly pass
- `%26 = OpAccessChain %_ptr_Uniform_R %rwsb %int_0 %uint_0` to `%R_incr`
[...]

PR #2546, opened 2019-05-01
For Vulkan, disallow structures containing opaque types

jeffbolznv:
This change corresponds to a pending change to the Vulkan SPIR-V environment spec.

### DIFF
--- a/source/val/validate_type.cpp
+++ b/source/val/validate_type.cpp
@@ -190,6 +190,35 @@ spv_result_t ValidateTypeRuntimeArray(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+bool ContainsOpaqueType(ValidationState_t& _, const Instruction* str) {
+  const size_t elem_type_index = 1;
+  uint32_t elem_type_id;
+  Instruction* elem_type;
+
+  if (spvOpcodeIsBaseOpaqueType(str->opcode())) {
+    return true;
+  }
+
+  switch (str->opcode()) {
+    case SpvOpTypeArray:
+    case SpvOpTypeRuntimeArray:
+      elem_type_id = str->GetOperandAs<uint32_t>(elem_type_index);
+      elem_type = _.FindDef(elem_type_id);
+      return ContainsOpaqueType(_, elem_type);
+    case SpvOpTypeStruct:
+      for (size_t member_type_index = 1;
+           member_type_index < str->operands().size(); ++member_type_index) {
+        auto member_type_id = str->GetOperandAs<uint32_t>(member_type_index);
+        auto member_type = _.FindDef(member_type_id);
+        if (ContainsOpaqueType(_, member_type)) return true;
+      }
+      break;
+    default:
+      break;
+  }
+  return false;
+}
+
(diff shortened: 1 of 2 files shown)


jeffbolznv:
@johnkslang @alan-baker can you give me some advice on the shaderc-smoketests failures?

There appear to be two classes of failures. The first is new "Validation Failed" strings in some glslang gtests, which we agreed are expected, but how do we deal with the failures until glslang gold results are updated?

The second class of failures is two HlslLegalization tests that still have a struct with an opaque member, which at first glance seems like a bug in legalization. Any idea what's going on there?

alan-baker:
The validator runs before the optimizer. Generally there are some extra options sent to the validator for unlegalized code (e.g. --relax-logical-pointer). If legalization is needed to address this issue, we'll need another option for the validator to disable this check until during legalization.

johnkslang:
Yes, this is a common coding pattern for HLSL that the optimizer turns into Vulkan-legal SPIR-V, so it should be included in the set of things tolerated by the validator for pre-legalization validation.

s-perron:
I'm thinking we should create 1 new options for the validator that covers everything.  We have `--relax-logical-pointer`, but it does not seem appropriate to disable this check with that option.  I also may need to create a new option for https://github.com/KhronosGroup/SPIRV-Tools/issues/2535.  We should add an option say "--pre-hlsl-legalization" and turn off this check if that option is enabled.

Then we would need changes to DXC and glslang to set that option when appropriate.

@jeffbolznv I can add the option in my upcoming PR, and then you can use it in this PR.  That assumes we are not in a hurry to merge this one.

jeffbolznv:
That sounds good to me. It's not urgent to merge this.

alan-baker:
@jeffbolznv the option was merged if you want to rebase.

jeffbolznv:
Can somebody point me at that commit/option? Is anybody planning to add it to glslang, or should I do that?

alan-baker:
e935dac9ef8a01fdb6d0a4b7898563e6f62d1081 added the option. I'm not aware of anyone currently adding the option to glslang's HLSL flow.

jeffbolznv:
This finally passes CI, and is ready to merge.

bkaradzic:
This change causes failure to build HLSL shaders containing code like:

```
struct BgfxSampler2D
{
	SamplerState m_sampler;
	Texture2D m_texture;
};
```

Ace17:
This changes also causes failure to build HLSL shaders containing code like:
```
cbuffer MyCBuffer
{
    Texture2D myTexture;
    SamplerState mySampler;
}
```

s-perron:
@bkaradzic  @Ace17 Can you open new issues against the HLSL compiler that you are using with an example that can be compiled?  This change is not wrong because it is implementing the Vulkan spec.  The compiler generating the SPIR-V might need to change.

Ace17:
Here's the issue on the glslang project : https://github.com/KhronosGroup/glslang/issues/2754

greg-lunarg:
@s-perron FWIW, my current plan is to handle #2754 completely in glslang; no changes should be required in spirv-opt legalization. See the issue for more details.

PR #2549, opened 2019-05-06
Validate sign of int types.

s-perron:
Fixes https://crbug.com/959011.

### DIFF
--- a/source/val/validate_type.cpp
+++ b/source/val/validate_type.cpp
@@ -14,11 +14,10 @@
 
 // Ensures type declarations are unique unless allowed by the specification.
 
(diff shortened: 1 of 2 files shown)


source/val/validate_type.cpp
```diff
@@ -14,11 +14,10 @@
 
 // Ensures type declarations are unique unless allowed by the specification.
 
-#include "source/val/validate.h"
-
 #include "source/opcode.h"
 #include "source/spirv_target_env.h"
 #include "source/val/instruction.h"
+#include "source/val/validate.h"
```
s-perron: Clang-format keeps moving this here.

PR #2550, opened 2019-05-07
Support SPIR-V 1.4

dneto0:
Assembler, disassembler, and most validation.

Missing validation:
-  Loop control: conflicts between different controls, particularly unroll vs. max iters, min iters
- OpAtomicCompareExchangeWeak and BufferBlock are removed (use lastVersion: "1.3" in the grammar)

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -860,6 +860,26 @@ spv_result_t ValidateCopyMemoryMemoryAccess(ValidationState_t& _,
       if (_.features().copy_memory_permits_two_memory_accesses) {
         if (auto error = CheckMemoryAccess(_, inst, second_access_index))
           return error;
+
+        // In the two-access form:
+        //  - the first is the target (write) access and it can't have
+        //  make-visible.
+        //  - the second is the source (read) access and it can't have
+        //  make-available.
+        // This is a proposed rule. See
+        // https://gitlab.khronos.org/spirv/SPIR-V/issues/413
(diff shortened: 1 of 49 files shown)


source/val/validate_memory.cpp
```diff
@@ -863,6 +844,53 @@ spv_result_t ValidateStore(ValidationState_t& _, const Instruction* inst) {
...
+      if (_.features().copy_memory_permits_two_memory_accesses) {
+        if (auto error = CheckMemoryAccess(_, inst, second_access_index))
+          return error;
+
+        // In the two-access form:
+        //  - the first is the target (write) access and it can't have
+        //  make-visible.
+        //  - the second is the source (read) access and it can't have
+        //  make-available.
+        // This is a proposed rule. See
```
alan-baker: No longer proposed?

tools/val/val.cpp
```diff
@@ -61,16 +61,18 @@ NOTE: The validator is a work in progress.
                                    different type with compatible layout and
                                    members.
   --version                        Display validator version information.
-  --target-env                     {vulkan1.0|vulkan1.1|opencl2.2|spv1.0|spv1.1|spv1.2|spv1.3|webgpu0}
+  --target-env                     {vulkan1.0|vulkan1.1|opencl2.2|spv1.0|spv1.1|spv1.2|spv1.3|spv1.4
```
alan-baker: Missing vulkan1.1spv1.4

PR #2563, opened 2019-05-07
Add pass to fix some invalid unreachable blocks for WebGPU

### LINKED ISSUE #2429: WebGPU transform: split a block that is unreachable and both a merge block and continue target
This is required because unreachable merge must end with OpUnreachable, but unreachable continue target must end with OpBranch.

zoddicus:
Attempts to split up unreachable blocks that are used both as a
merge-block and a continue-target.

Fixes #2429

### DIFF
--- a/source/opt/optimizer.cpp
+++ b/source/opt/optimizer.cpp
@@ -225,6 +225,7 @@ Optimizer& Optimizer::RegisterVulkanToWebGPUPasses() {
       .RegisterPass(CreateStripAtomicCounterMemoryPass())
       .RegisterPass(CreateGenerateWebGPUInitializersPass())
       .RegisterPass(CreateLegalizeVectorShufflePass())
+      .RegisterPass(CreateSplitInvalidUnreachablePass())
       .RegisterPass(CreateEliminateDeadConstantPass())
       .RegisterPass(CreateFlattenDecorationPass())
       .RegisterPass(CreateAggressiveDCEPass())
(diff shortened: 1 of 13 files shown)


source/opt/optimizer.cpp
```diff
@@ -873,4 +874,9 @@ Optimizer::PassToken CreateDecomposeInitializedVariablesPass() {
       MakeUnique<opt::DecomposeInitializedVariablesPass>());
 }
 
+Optimizer::PassToken CreateSplitInvalidUnreachablePass() {
```
s-perron: Do you add command line options for the passes you are adding?  They are useful and should be added.

source/opt/split_invalid_unreachable_pass.cpp
```diff
@@ -0,0 +1,87 @@
...
+              std::unique_ptr<BasicBlock> new_block(
+                  new BasicBlock(std::move(new_label)));
+              auto* block_ptr = new_block.get();
+              InstructionBuilder builder(
+                  context(), new_block.get(),
+                  IRContext::kAnalysisDefUse |
+                      IRContext::kAnalysisInstrToBlockMapping);
+              builder.AddUnreachable();
+              cfg()->RegisterBlock(block_ptr);
+              (&*func)->AddBasicBlock(std::move(new_block));
```
s-perron: It will look awkward when people look at the spir-v because the merge block will be at the end of the function.  It is not necessary, but it might be best to insert the new block before `unreachable`.

zoddicus: Done

source/opt/optimizer.cpp
zoddicus: Created #2574 to track implementing this.

zoddicus:
Committed before the bots completed because the last patch was just running the formatter and all the other checks had passed.

PR #2570, opened 2019-05-08
Make pointers to isomorphic type interchangeable with option.

s-perron:
DXC will be generating code where the function parameters will be a more
generic type that the actual parameter.  They should be the same except
for the actual parameter will have extra decorations.

We want to accept this code with an options so that spirv-opt can then
inline and fix the type mismatch.  We will accept this under a new
options `--before-hlsl-legalization`.

The new option will also imply `relax-logical-pointer` so that HLSL
frontends will need to use just the one more generic option.

### DIFF
--- a/source/val/validate_function.cpp
+++ b/source/val/validate_function.cpp
@@ -12,27 +12,37 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "source/val/validate.h"
-
 #include <algorithm>
 
 #include "source/opcode.h"
 #include "source/val/instruction.h"
+#include "source/val/validate.h"
 #include "source/val/validation_state.h"
 
 namespace spvtools {
 namespace val {
 namespace {
 
-// Returns true if |a| and |b| are instruction defining pointers that point to
-// the same type.
-bool ArePointersToSameType(val::Instruction* a, val::Instruction* b) {
+// Returns true if |a| and |b| are instructions defining pointers that point to
+// types logically match and the decorations that apply to |b| are a subset
+// of the decorations that apply to |a|.
+bool DoPointeesLogicallyMatch(val::Instruction* a, val::Instruction* b,
+                              ValidationState_t& _) {
   if (a->opcode() != SpvOpTypePointer || b->opcode() != SpvOpTypePointer) {
     return false;
   }
 
   uint32_t a_type = a->GetOperandAs<uint32_t>(2);
-  return a_type && (a_type == b->GetOperandAs<uint32_t>(2));
+  uint32_t b_type = b->GetOperandAs<uint32_t>(2);
+
+  if (a_type == b_type) {
+    return true;
+  }
+
+  Instruction* a_type_inst = _.FindDef(a_type);
+  Instruction* b_type_inst = _.FindDef(b_type);
+
+  return _.LogicallyMatch(a_type_inst, b_type_inst, true);
 }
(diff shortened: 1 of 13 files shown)


source/val/validate_function.cpp
```diff
@@ -12,27 +12,102 @@
...
 namespace val {
 namespace {
 
-// Returns true if |a| and |b| are instruction defining pointers that point to
-// the same type.
-bool ArePointersToSameType(val::Instruction* a, val::Instruction* b) {
+// Returns true if |a| and |b| are instructions defining types that are
+// isomorphic and the decorations that apply to |b| are a subset of the
+// decorations that apply to |a|.
+bool AreIsomorphicTypes(val::Instruction* a, val::Instruction* b,
```
alan-baker: This is very similar to logically matching types for OpCopyLogical in SPIR-V 1.4. Take a look at `LogicallyMatch` in `validate_composites.cpp`. The SPIR-V spec also defines logically matching in OpCopyLogical's description.

s-perron: thanks that is very useful.  I have used it, but I had to change it a bit.  I also wanted to check decorations.

test/val/val_function_test.cpp
```diff
@@ -416,6 +416,369 @@ TEST_P(ValidateFunctionCall, NonMemoryObjectDeclarationVariablePointers) {
...
+               OpEntryPoint GLCompute %1 "main"
+               OpExecutionMode %1 LocalSize 1 1 1
+               OpSource HLSL 600
+               OpDecorate %2 DescriptorSet 0
+               OpDecorate %2 Binding 0
+               OpMemberDecorate %_struct_3 0 Offset 0
+               OpDecorate %_runtimearr__struct_3 ArrayStride 4
+               OpMemberDecorate %_struct_5 0 Offset 0
+               OpDecorate %_struct_5 BufferBlock
+               OpMemberDecorate %_struct_15 0 Offset 4
```
alan-baker: This is sort of problematic since this decoration should be ignored when instantiated in Function storage class. See SPIR internal issue 444. I would suggest using another another decoration, but that would only address this specific test.

s-perron: Since I am dealing with code that is illegal anyway, I'm not too worried about things like whether or not a decoration should be ignored.  We are creating our own superset of code that we are willing to accept.  I'll change it to another decoration, but I don't see the need to change the implementation.

test/val/val_function_test.cpp
```diff
@@ -714,6 +714,61 @@ TEST_F(ValidateFunctionCall,
...
+               OpCapability Shader
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint GLCompute %1 "main"
+               OpExecutionMode %1 LocalSize 1 1 1
+               OpSource HLSL 600
+               OpDecorate %2 DescriptorSet 0
+               OpDecorate %2 Binding 0
+               OpMemberDecorate %_struct_3 0 Offset 0
+               OpDecorate %_runtimearr__struct_3 ArrayStride 4
+               OpDecorate %_runtimearr__struct_3b ArrayStride 8
```
alan-baker: I wanted the ArrayStride decorations to target the pointers so that layout issues on the pointers are proven to be caught.

s-perron:
Smoketest fails without my change.  I'll merge anyway.

PR #2573, opened 2019-05-09
Update instrumentation passes to handle 1.4 interfaces

alan-baker:
When the input and output buffer ids are added as variables, also add them as interfaces to all entry points.


alan-baker:
FYI @greg-lunarg

PR #2576, opened 2019-05-09
Remove MarkV and Stats code.

dj2:
This Cl removes the MarkV and Stats code from SPIRV-Tools. This code was
unused and currently un-maintained.


dj2:
Done.

PR #2579, opened 2019-05-10
Validate loop merge

### LINKED ISSUE #2559: Validate loop control
Bits with operands are checked, but conflicting rules are not.

alan-baker:
Fixes #2559

* Validate OpLoopMerge including loop controls
  * add tests
  * fix some bad tests

### DIFF
--- a/source/val/validate_cfg.cpp
+++ b/source/val/validate_cfg.cpp
@@ -230,6 +230,78 @@ spv_result_t ValidateReturnValue(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+spv_result_t ValidateLoopMerge(ValidationState_t& _, const Instruction* inst) {
+  const auto merge_id = inst->GetOperandAs<uint32_t>(0);
+  const auto merge = _.FindDef(merge_id);
+  if (!merge || merge->opcode() != SpvOpLabel) {
+    return _.diag(SPV_ERROR_INVALID_ID, inst)
+           << "Merge Block " << _.getIdName(merge_id) << " must be an OpLabel";
+  }
+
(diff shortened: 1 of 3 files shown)


source/val/validate_cfg.cpp
```diff
@@ -230,6 +230,78 @@ spv_result_t ValidateReturnValue(ValidationState_t& _,
...
+
+  const auto continue_id = inst->GetOperandAs<uint32_t>(1);
+  const auto continue_target = _.FindDef(continue_id);
+  if (!continue_target || continue_target->opcode() != SpvOpLabel) {
+    return _.diag(SPV_ERROR_INVALID_ID, inst)
+           << "Continue Target " << _.getIdName(continue_id)
+           << " must be an OpLabel";
+  }
+
+  if (merge_id == continue_id) {
```
dneto0: I also tried an example where merge_id was the same as the header and that errored out elsewhere.
I tried other weird stuff and found https://github.com/KhronosGroup/SPIRV-Tools/issues/2588

But this is fine  in any case.

source/val/validate_cfg.cpp
```diff
@@ -230,6 +230,78 @@ spv_result_t ValidateReturnValue(ValidationState_t& _,
...
+    return _.diag(SPV_ERROR_INVALID_ID, inst)
+           << "Merge Block and Continue Target must be different ids";
+  }
+
+  const auto loop_control = inst->GetOperandAs<uint32_t>(2);
+  if ((loop_control >> SpvLoopControlUnrollShift) & 0x1 &&
```
dneto0: could use the ... Mask versions of these enums instead?

PR #2580, opened 2019-05-11
Linker: Better type comparison for OpTypeArray and OpTypeForwardPointer

### LINKED ISSUE #2442: Linker does not deduplicate equivalent array types
The linker does not deduplicate equivalent array types between modules, which causes type mismatch errors when sharing an array type (or types based on an array) between modules. Can verify fails for variables and for functions (when used as an argument, directly and pointer to).

Example, causes a type mismatch on symbol "x":
```
; a.spv
               OpCapability Linkage
               OpCapability Shader
               OpMemoryModel Logical Simple
               OpDecorate %1 LinkageAttributes "x" Import
       %void = OpTypeVoid
       %uint = OpTypeInt 32 0
     %uint_7 = OpConstant %uint 7
%_arr_uint_uint_7 = OpTypeArray %uint %uint_7
%_ptr_Workgroup__arr_uint_uint_7 = OpTypePointer Workgroup %_arr_uint_uint_7
          %1 = OpVariable %_ptr_Workgroup__arr_uint_uint_7 Workgroup

; b.spv, differs from a.spv only in changing Import -> Export
               OpCapability Linkage
               OpCapability Shader
               OpMemoryModel Logical Simple
               OpDecorate %1 LinkageAttributes "x" Export
       %void = OpTypeVoid
       %uint = OpTypeInt 32 0
     %uint_7 = OpConstant %uint 7
%_arr_uint_uint_7 = OpTypeArray %uint %uint_7
%_ptr_Workgroup__arr_uint_uint_7 = OpTypePointer Workgroup %_arr_uint_uint_7
          %1 = OpVariable %_ptr_Workgroup__arr_uint_uint_7 Workgroup
```
Removing the `OpDecorate` instructions produces duplicates of both the constant and types:
```
               OpCapability Shader
               OpMemoryModel Logical Simple
               OpModuleProcessed "Linked by SPIR-V Tools Linker"
       %void = OpTypeVoid
       %uint = OpTypeInt 32 0
     %uint_7 = OpConstant %uint 7
%_arr_uint_uint_7 = OpTypeArray %uint %uint_7
%_ptr_Workgroup__arr_uint_uint_7 = OpTypePointer Workgroup %_arr_uint_uint_7
          %6 = OpVariable %_ptr_Workgroup__arr_uint_uint_7 Workgroup
   %uint_7_0 = OpConstant %uint 7
%_arr_uint_uint_7_0 = OpTypeArray %uint %uint_7_0
%_ptr_Workgroup__arr_uint_uint_7_0 = OpTypePointer Workgroup %_arr_uint_uint_7_0
         %10 = OpVariable %_ptr_Workgroup__arr_uint_uint_7_0 Workgroup
```

pierremoreau:
This is an attempt to fix #2442 by re-implementing type comparisons for `OpTypeArray`. This new comparison checks that the values of the constants are the same, or if those are specialised constants, that they refer to the same `SpecId`.

### DIFF
--- a/source/opt/remove_duplicates_pass.cpp
+++ b/source/opt/remove_duplicates_pass.cpp
@@ -96,35 +96,67 @@ bool RemoveDuplicatesPass::RemoveDuplicateTypes() const {
     return modified;
   }
 
+  analysis::TypeManager type_manager(context()->consumer(), context());
+
   std::vector<Instruction*> visited_types;
+  std::vector<analysis::ForwardPointer> visited_forward_pointers;
   std::vector<Instruction*> to_delete;
   for (auto* i = &*context()->types_values_begin(); i; i = i->NextNode()) {
+    const bool is_i_forward_pointer = i->opcode() == SpvOpTypeForwardPointer;
+
     // We only care about types.
-    if (!spvOpcodeGeneratesType((i->opcode())) &&
-        i->opcode() != SpvOpTypeForwardPointer) {
+    if (!spvOpcodeGeneratesType(i->opcode()) && !is_i_forward_pointer) {
       continue;
     }
 
-    // Is the current type equal to one of the types we have aready visited?
-    SpvId id_to_keep = 0u;
-    // TODO(dneto0): Use a trie to avoid quadratic behaviour? Extract the
-    // ResultIdTrie from unify_const_pass.cpp for this.
-    for (auto j : visited_types) {
-      if (AreTypesEqual(*i, *j, context())) {
-        id_to_keep = j->result_id();
-        break;
+    if (!is_i_forward_pointer) {
+      // Is the current type equal to one of the types we have already visited?
+      SpvId id_to_keep = 0u;
+      analysis::Type* i_type = type_manager.GetType(i->result_id());
+      assert(i_type);
+      // TODO(dneto0): Use a trie to avoid quadratic behaviour? Extract the
+      // ResultIdTrie from unify_const_pass.cpp for this.
+      for (auto j : visited_types) {
+        analysis::Type* j_type = type_manager.GetType(j->result_id());
+        assert(j_type);
+        if (*i_type == *j_type) {
+          id_to_keep = j->result_id();
+          break;
+        }
       }
-    }
 
-    if (id_to_keep == 0u) {
-      // This is a never seen before type, keep it around.
-      visited_types.emplace_back(i);
+      if (id_to_keep == 0u) {
+        // This is a never seen before type, keep it around.
+        visited_types.emplace_back(i);
+      } else {
+        // The same type has already been seen before, remove this one.
+        context()->KillNamesAndDecorates(i->result_id());
+        context()->ReplaceAllUsesWith(i->result_id(), id_to_keep);
+        modified = true;
+        to_delete.emplace_back(i);
+      }
     } else {
-      // The same type has already been seen before, remove this one.
-      context()->KillNamesAndDecorates(i->result_id());
-      context()->ReplaceAllUsesWith(i->result_id(), id_to_keep);
-      modified = true;
-      to_delete.emplace_back(i);
+      analysis::ForwardPointer i_type(
+          i->GetSingleWordInOperand(0u),
+          (SpvStorageClass)i->GetSingleWordInOperand(1u));
+      i_type.SetTargetPointer(
+          type_manager.GetType(i_type.target_id())->AsPointer());
+
+      // TODO(dneto0): Use a trie to avoid quadratic behaviour? Extract the
+      // ResultIdTrie from unify_const_pass.cpp for this.
+      const bool found_a_match =
+          std::find(std::begin(visited_forward_pointers),
+                    std::end(visited_forward_pointers),
+                    i_type) != std::end(visited_forward_pointers);
+
+      if (!found_a_match) {
+        // This is a never seen before type, keep it around.
+        visited_forward_pointers.emplace_back(i_type);
+      } else {
+        // The same type has already been seen before, remove this one.
+        modified = true;
+        to_delete.emplace_back(i);
+      }
     }
   }
 
@@ -151,8 +183,8 @@ bool RemoveDuplicatesPass::RemoveDuplicateDecorations() const {
 
   analysis::DecorationManager decoration_manager(context()->module());
   for (auto* i = &*context()->annotation_begin(); i;) {
-    // Is the current decoration equal to one of the decorations we have aready
-    // visited?
+    // Is the current decoration equal to one of the decorations we have
+    // already visited?
     bool already_visited = false;
     // TODO(dneto0): Use a trie to avoid quadratic behaviour? Extract the
     // ResultIdTrie from unify_const_pass.cpp for this.
@@ -175,21 +207,6 @@ bool RemoveDuplicatesPass::RemoveDuplicateDecorations() const {
   }
 
   return modified;
-}
-
-bool RemoveDuplicatesPass::AreTypesEqual(const Instruction& inst1,
-                                         const Instruction& inst2,
-                                         IRContext* context) {
-  if (inst1.opcode() != inst2.opcode()) return false;
-  if (!IsTypeInst(inst1.opcode())) return false;
-
-  const analysis::Type* type1 =
-      context->get_type_mgr()->GetType(inst1.result_id());
-  const analysis::Type* type2 =
-      context->get_type_mgr()->GetType(inst2.result_id());
-  if (type1 && type2 && *type1 == *type2) return true;
-
-  return false;
 }
 
 }  // namespace opt
--- a/source/opt/type_manager.cpp
+++ b/source/opt/type_manager.cpp
@@ -66,7 +66,13 @@ uint32_t TypeManager::GetId(const Type* type) const {
 }
 
 void TypeManager::AnalyzeTypes(const Module& module) {
-  // First pass through the types.  Any types that reference a forward pointer
+  // First pass through the constants, as some will be needed when traversing
+  // the types in the next pass.
+  for (const auto* inst : module.GetConstants()) {
+    id_to_constant_inst_[inst->result_id()] = inst;
+  }
+
+  // Then pass through the types.  Any types that reference a forward pointer
   // (directly or indirectly) are incomplete, and are added to incomplete types.
   for (const auto* inst : module.GetTypes()) {
     RecordIfTypeDefinition(*inst);
@@ -154,7 +160,7 @@ void TypeManager::AnalyzeTypes(const Module& module) {
 
 #ifndef NDEBUG
   // Check if the type pool contains two types that are the same.  This
-  // is an indication that the hashing and comparision are wrong.  It
+  // is an indication that the hashing and comparison are wrong.  It
   // will cause a problem if the type pool gets resized and everything
   // is rehashed.
   for (auto& i : type_pool_) {
@@ -505,8 +511,15 @@ Type* TypeManager::RebuildType(const Type& type) {
     case Type::kArray: {
       const Array* array_ty = type.AsArray();
       const Type* ele_ty = array_ty->element_type();
-      rebuilt_ty =
-          MakeUnique<Array>(RebuildType(*ele_ty), array_ty->LengthId());
+      if (array_ty->length_spec_id() != 0u)
+        rebuilt_ty =
+            MakeUnique<Array>(RebuildType(*ele_ty), array_ty->LengthId(),
+                              array_ty->length_spec_id());
+      else
+        rebuilt_ty =
+            MakeUnique<Array>(RebuildType(*ele_ty), array_ty->LengthId(),
+                              array_ty->length_constant_type(),
+                              array_ty->length_constant_words());
       break;
     }
     case Type::kRuntimeArray: {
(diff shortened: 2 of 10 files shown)


pierremoreau:
I’m opening this as a WIP and only implemented it for `OpTypeArray`, as I am not sure this is the solution we want to go for. It would probably be best to have this as part of the regular type comparison done by the `Type::IsSameImpl()` functions, however I don’t know how this should be done. We could store the extra information when creating the types, but was the `DefUseManager` and `DecorationManager` created and up-to-date?

source/opt/type_manager.cpp
```diff
@@ -821,6 +866,60 @@ const Type* TypeManager::GetMemberType(
...
+      const analysis::Array* array_type1 = type1->AsArray();
+      assert(array_type1);
+
+      const analysis::Array* array_type2 = type2->AsArray();
+      assert(array_type2);
+
+      return AreConstantsCompatible(array_type1->LengthId(),
+                                    array_type2->LengthId(), context);
+    } break;
+    default:
```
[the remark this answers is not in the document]
pierremoreau: Vector and the others also need special support, not regarding constants but regarding IDs. If you have the following inputs:
```
  %uint = OpTypeInt 32 0
%uvec4 = OpTypeVector %uint 4
```
and
```
  %uint = OpTypeInt 32 0
%uvec4 = OpTypeVector %uint 4
```
then after linking you get
```
    %uint = OpTypeInt 32 0
  %uvec4 = OpTypeVector %uint 4
  %uint_0 = OpTypeInt 32 0
%uvec4_0 = OpTypeVector %uint_0 4
```
and `TYPE(%uvec4_0) == TYPE(%uvec4)` is currently going to return false as `ID(%uint) != ID(%uint0)`.

source/opt/type_manager.cpp
```diff
@@ -24,6 +24,51 @@
...
+
+    const Operand* constant2_spec_id;
+    decoration_manager.ForEachDecoration(
+        constant2_id, SpvDecorationSpecId,
+        [&constant2_spec_id](const Instruction& inst) {
+          assert(inst.opcode() == SpvOpDecorate);
+          constant2_spec_id = &inst.GetOperand(2u);
+        });
+
+    return *constant1_spec_id == *constant2_spec_id;
```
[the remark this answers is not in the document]
pierremoreau: Good point; I added an extra `if`-statement to early out in case one of the two pointers are null, before dereferencing them.

source/opt/type_manager.cpp
```diff
@@ -24,6 +24,51 @@
...
+  } else {
+    return type1_constant->GetOperand(2u) == type2_constant->GetOperand(2u);
```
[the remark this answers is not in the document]
pierremoreau: True, though I am not sure we want to accept such a case, at least not for linking, and in the regular case they should be considered different AFAICT.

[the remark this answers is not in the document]
blue42u: Then, should the types of the constants be compared in this case? At the moment 32, 16 and 8-bit constants (in both signedness) are compatible.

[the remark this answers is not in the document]
pierremoreau: You are correct, the types should be compared.

source/opt/type_manager.cpp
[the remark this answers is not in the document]
blue42u: `TYPE(%uvec4_0) == TYPE(%uvec4)` returns true, `Vector::IsSameImpl` is written as
```cpp
  return count_ == vt->count_ &&
         element_type_->IsSameImpl(vt->element_type_, seen) &&
         HasSameDecorations(that);
```
which doesn't compare the IDs but uses `Integer::IsSameImpl` in this case:
```cpp
  return it && width_ == it->width_ && signed_ == it->signed_ &&
         HasSameDecorations(that);
```
which compare equal as they have the same width and signedness. The other restricted types are similar, only Pointer, RuntimeArray, Struct and Function could have an Array as a (potentially embedded) component type, which is one of the two Types that compare by ID. (The other is ForwardPointer, but that has extra more difficult issues).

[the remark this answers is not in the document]
pierremoreau: I should have checked the implementation rather than just look at the spec; you are correct.

source/opt/type_manager.cpp
```diff
@@ -821,6 +866,60 @@ const Type* TypeManager::GetMemberType(
...
+    }
+    case analysis::Type::kArray: {
+      const analysis::Array* array_type1 = type1->AsArray();
+      assert(array_type1);
+
+      const analysis::Array* array_type2 = type2->AsArray();
+      assert(array_type2);
+
+      return AreConstantsCompatible(array_type1->LengthId(),
+                                    array_type2->LengthId(), context);
```
s-perron: Don't you need to check the element types as well?

source/opt/type_manager.cpp
```diff
@@ -821,6 +866,60 @@ const Type* TypeManager::GetMemberType(
...
+    case analysis::Type::kPointer: {
+      const analysis::Pointer* pointer_type1 = type1->AsPointer();
+      assert(pointer_type1);
+
+      const analysis::Pointer* pointer_type2 = type2->AsPointer();
+      assert(pointer_type2);
+
+      return pointer_type1->storage_class() == pointer_type2->storage_class() &&
+             AreTypesEqual(pointer_type1->pointee_type(),
+                           pointer_type2->pointee_type(), context);
```
s-perron: Is it possible to get an infinite loop if you have a link-list type structure?

pierremoreau:
I completely re-wrote the implementation to have all the verification done inside the type system, instead of having a separate function.
Sorry @s-perron, for not pushing it a bit earlier and avoiding for you to review an old version.

source/opt/type_manager.cpp
```diff
@@ -636,15 +648,39 @@ Type* TypeManager::RecordIfTypeDefinition(const Instruction& inst) {
     case SpvOpTypeSampledImage:
       type = new SampledImage(GetType(inst.GetSingleWordInOperand(0)));
       break;
-    case SpvOpTypeArray:
-      type = new Array(GetType(inst.GetSingleWordInOperand(0)),
-                       inst.GetSingleWordInOperand(1));
+    case SpvOpTypeArray: {
+      const uint32_t length_id = inst.GetSingleWordInOperand(1);
+      const Instruction* length_constant_insn = id_to_constant_insn_[length_id];
```
s-perron: I would prefer to use "inst" as the short form for "instruction".  That is already commonly used in spirv-tools.

source/opt/type_manager.h
```diff
@@ -209,6 +209,8 @@ class TypeManager {
 
   IdToTypeMap id_to_incomplete_type_;  // Maps ids to their type representations
                                        // for incomplete types.
+
+  std::unordered_map<uint32_t, const Instruction*> id_to_constant_insn_;
```
s-perron: This looks like it is duplicating information that is in the constant manager.  Isn't there another way to use that instead?

source/opt/types.cpp
```diff
@@ -383,17 +383,46 @@ void SampledImage::GetExtraHashWords(
...
+  // If it is a specialized constant
+  if (length_spec_id_ != 0u) {
+    // ensure they have the same SpecId
+    is_same = is_same && length_spec_id_ == at->length_spec_id_;
+  } else {
+    // else, ensure they have the same length literal number.
+    is_same =
+        is_same &&
+        length_constant_type_->IsSameImpl(at->length_constant_type_, seen) &&
+        length_constant_words_ == at->length_constant_words_;
```
s-perron: This seems to be the only place where these two new members are used.  In general, the constant manager will be available right?  Couldn't you use the constant manager instead?  Say the Array class hold a pointer to a constant and this check would just make sure they are the same constant.

source/opt/remove_duplicates_pass.cpp
```diff
@@ -151,8 +183,8 @@ bool RemoveDuplicatesPass::RemoveDuplicateDecorations() const {
 
   analysis::DecorationManager decoration_manager(context()->module());
   for (auto* i = &*context()->annotation_begin(); i;) {
-    // Is the current decoration equal to one of the decorations we have aready
-    // visited?
+    // Is the current decoration equal to one of the decorations we have
+    // aready visited?
```
s-perron: Since you are fixing the formating here, could you fix the typo as well? Should be "already visited".

source/opt/types.cpp
blue42u: AFAIK using the constant manager here will cause some issues. The constant manager uses the type manager to unify Constants, so using it in here (should) cause an infinite loop between `IRContext::BuildTypeManager` and `BuildConstantManager`. On the other hand, valid SPIR-V should present types and constants in a forward-only order, so this might not be an issue if instruction parsing is interleaved between the two managers (requiring a more significant refactor).

source/opt/type_manager.h
pierremoreau: I am not entirely sure: the constant manager depends on the type manager (i.e. creating a constant requires passing in the corresponding type object), so the constant manager is built after the type manager. But now, for creating some of the types, we would need constants to already have been created, so we have a circular dependency.

source/opt/types.cpp
pierremoreau: Ideally yes, but we might have to create the constant object manually rather than going through the constant manager, since `IntConstant` requires an `Integer` object, and then `Array` would require an `IntConstant`. So we would need to analyse some of the types to construct some of the types object, that can then be used to create the constant objects, which in turn will be used to create more types object, which will allow creating more constant objects…

source/opt/remove_duplicates_pass.cpp
pierremoreau: Fixed

source/opt/type_manager.cpp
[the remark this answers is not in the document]
pierremoreau: The types are now compared in the new version.

pierremoreau:
@blue42u I am not really familiar with re2 & co. Do you have any idea why this fails?
```
[ RUN      ] TypeMatch.ArrayOfTypeFloat
/test/link/linker_fixture.h:173: Failure
Expected equality of these values:
  effcee::Result::Status::Ok
    Which is: 4-byte object <00-00 00-00>
  match_res.status()
    Which is: 4-byte object <01-00 00-00>
<stdin>:16:10: error: expected string not found in input
; CHECK: [[type:%\w+]] = OpTypeArray [[a:%\w+]] %const
         ^
<stdin>:11:24: note: scanning from here
%float = OpTypeFloat 32
                       ^

Expanded from:
OpCapability Linkage
OpCapability NamedBarrier
OpCapability PipeStorage
OpCapability Pipes
OpCapability DeviceEnqueue
OpCapability Kernel
OpCapability Shader
OpCapability Addresses
OpDecorate %var LinkageAttributes "foo" {Import,Export}
; CHECK: [[baseint:%\w+]] = OpTypeInt 32 1
%baseint = OpTypeInt 32 1
; CHECK: [[const:%\w+]] = OpConstant [[baseint]] 3
%const = OpConstant %baseint 3
; CHECK: [[a:%\w+]] = OpTypeFloat 32
%a = OpTypeFloat 32
; CHECK: [[type:%\w+]] = OpTypeArray [[a:%\w+]] %const
%type = OpTypeArray %a %const
; CHECK: OpVariable [[type]] Uniform
%var = OpVariable %type Uniform
Checking result:
OpCapability NamedBarrier
OpCapability PipeStorage
OpCapability Pipes
OpCapability DeviceEnqueue
OpCapability Kernel
OpCapability Shader
OpCapability Addresses
OpModuleProcessed "Linked by SPIR-V Tools Linker"
%int = OpTypeInt 32 1
%int_3 = OpConstant %int 3
%float = OpTypeFloat 32
%_arr_float_int_3 = OpTypeArray %float %int_3
%int_3_0 = OpConstant %int 3
%6 = OpVariable %_arr_float_int_3 Uniform

[  FAILED  ] TypeMatch.ArrayOfTypeFloat (15 ms)
```
The `%_arr_float_int_3 = OpTypeArray %float %int_3` should pass the regex test IMHO.

pierremoreau:
All the tests are passing now, except for `TypeMatch.ForwardPointerOfArray`, which fails with
```
test_link: source/opt/type_manager.cpp:171: void spvtools::opt::analysis::TypeManager::AnalyzeTypes(const spvtools::opt::Module &): Assertion `(ti == tj || !ti->IsSame(tj)) && "Type pool contains two types that are the same."' failed.
```

blue42u:
Ah, apologies, that's a remnant from an old test. It should really be
```
; CHECK: [[type:%\w+]] = OpTypeArray [[a:%\w+]] [[const]]
```
to match with the `const` defined by the `OpConstant` check line (`[[const:%\w+]]` will match any ID). Also, thanks for fixing the ForwardPointer tests, I must have entirely misread that part of the spec when writing them.

I'll give a more thorough look next week when I have more time to spare, the error appears to be fairly subtle at a glance (maybe something if/when the result of `ForwardPointer::IsSameImpl` changes?)

pierremoreau:
If I understand correctly, it could even be
```
; CHECK: [[type:%\w+]] = OpTypeArray [[a]] [[const]]
```
so that `a` only accepts an already defined type rather than ID.

blue42u:
That is correct, it looks like there was a typo in my original tests (should have been `"[[" #T "]]"` in `MatchPart2`). Maybe something like [this](https://github.com/KhronosGroup/SPIRV-Tools/files/3219403/x.txt) would suffice, although its a bit messy.

pierremoreau:
We went with the same solution. :smile: 

Regarding the remaining error, it seems to come from the hashing of `Pointer`: when adding a `Type` to the `type_pool_`, they are automatically tested for uniqueness using a provided hash (computed via the `GetExtraHashWords()` method) since an `std::unordered_set` is used. But the check later on (that fails) uses `SameImpl()` for the comparison.

So we have two types for which `SameImpl()` returns `true`, but their hashes are actually different; I’ll continue debugging it.

pierremoreau:
I needed to update `Array::GetExtraHashWords()` to fix the last test; this has now been fixed and all tests are enabled and passing.

s-perron:
Please fix the failures in the bots.  Looks like "shadow variables":

```
/usr/bin/c++   -DSPIRV_CHECK_CONTEXT -DSPIRV_COLOR_TERMINAL -DSPIRV_LINUX -DSPIRV_TIMER_ENABLED -I../ -I../include -I../external/spirv-headers/include -I. -g -fPIC   -Wall -Wextra -Wnon-virtual-dtor -Wno-missing-field-initializers -Werror -std=c++11 -fno-exceptions -fno-rtti -Wno-long-long -Wshadow -Wundef -Wconversion -Wno-sign-conversion -std=gnu++11 -MD -MT source/opt/CMakeFiles/SPIRV-Tools-opt.dir/types.cpp.o -MF source/opt/CMakeFiles/SPIRV-Tools-opt.dir/types.cpp.o.d -o source/opt/CMakeFiles/SPIRV-Tools-opt.dir/types.cpp.o -c /tmpfs/src/github/SPIRV-Tools/source/opt/types.cpp
461
/tmpfs/src/github/SPIRV-Tools/source/opt/types.cpp: In constructor ‘spvtools::opt::analysis::Array::Array(spvtools::opt::analysis::Type*, uint32_t, const spvtools::opt::analysis::Type*, spvtools::opt::Operand::OperandData)’:
462
/tmpfs/src/github/SPIRV-Tools/source/opt/types.cpp:398:56: error: declaration of ‘length_constant_type’ shadows a member of 'this' [-Werror=shadow]
463
              Operand::OperandData length_constant_words)
```

pierremoreau:
@s-perron The variable shadowing should be fixed, so the kokoro run could be restarted.

pierremoreau:
@s-perron Are you able to restart the AppVeyor test? It did pass but GitHub isn’t picking up the results. I fixed the kokoro failure for android and gcc, so hopefully all of them should now be passing.

pierremoreau:
> Everything looks good on my end, thank you very much for fixing this.

Thank you for reporting the issue, submitting the tests and trying out the fix.

pierremoreau:
I’m fine with the current version, though I am sure that some parts could be improved.

PR #2581, opened 2019-05-13

afd:
This PR superseded #2529 (which I will close), and incorporates the review feedback given by @alan-baker and @paulthomson on that PR.

This PR is pretty large because it was difficult to make good decisions on some of the design aspects of spirv-fuzz without getting quite a bit of stuff up and running.  Future PRs to this tool should be a lot smaller if they can build on the framework presented in this PR.

A major design decision in this PR is to use protobufs to represent transformations, and to have transformation functions operate directly on C++ protobuf objects, rather than parsing said objects into some other representation and turning that representation back into protobufs.

With this design, we add a dependency on the protobuf library, and I would like feedback on whether that dependency is adequately handled at present, which is by invoking include(FindProtobuf) in source/fuzz/CMakeLists.txt.  This requires that the protobuf library is already installed on the user's system.


afd:
Abandoning this PR due to a rebasing error.  New PR is #2584.

PR #2582, opened 2019-05-14
Handle id overflow in sroa better.

s-perron:
There is a case where sroa is not handling id overflow gracefully.  It
is handled and an error message is output when the ids overflow.

Fixes https://crbug.com/961030.


s-perron:
I first say the shaderc failure yesterday.  It is unrelated to my change.

dnovillo:
> I first saw the shaderc failure yesterday. It is unrelated to my change.

Oh, good.  Do you know if it's been reported in shaderc?

PR #2583, opened 2019-05-14
Bindless Validation: Instrument descriptor-based loads and stores

greg-lunarg:
Essentially, support UBOs and SSBOs, scalar and array (sized and unsized).


greg-lunarg:
The shaderc smoketest was the only CI failure and I am pretty sure it was not due to my change which was isolated to --inst-bindless-check which is not (at least should not) be part of any compilation outside of the GPU Validation Layer.

dneto0:
Agreed, the smoketest failure is a symptom of the fact that https://github.com/KhronosGroup/SPIRV-Headers/pull/103 was merged but the Glslang had not yet updated SPIRV-Headers for its own tests, so validation results are different.

PR #2584, opened 2019-05-15

afd:
This PR superseded #2529, and incorporates the review feedback given by @alan-baker and @paulthomson on that PR.

This PR is pretty large because it was difficult to make good decisions on some of the design aspects of spirv-fuzz without getting quite a bit of stuff up and running. Future PRs to this tool should be a lot smaller if they can build on the framework presented in this PR.

A major design decision in this PR is to use protobufs to represent transformations, and to have transformation functions operate directly on C++ protobuf objects, rather than parsing said objects into some other representation and turning that representation back into protobufs.

With this design, we add a dependency on the protobuf library, which is built from source.  In the CI, it is disabled for VS2013 (which cannot build protobufs) and Android (which I'm sure can, but it did not work out of the box, and while fuzzing Android graphics drivers is very important, the SPIR-V transformation tool would not be expected to run directly on Android.)


source/fuzz/fact_manager.cpp
```diff
@@ -0,0 +1,274 @@
+#include <utility>
```
dj2: This should be after the copyright.

afd:
@dj2 Yes, I think that splitting it up like that as you suggest would be a good idea.

@alan-baker and @s-perron any thoughts on that?

alan-baker:
@dj2's suggestions for the split seem reasonable to me. Should make it considerably easier to review.

s-perron:
Yes, I agree on dividing it as Dan suggested.

afd:
OK, I have opened #2616 as the first of this proposed series of PRs.  I would be grateful for a review there, and will close this one shortly.

source/fuzz/fact_manager.cpp
afd: Fixed, thanks.

PR #2587, opened 2019-05-15
Add generated instruction API.

fjhenigman:
Adds class spvtools::Instruction and an automatically generated derived
class for each SPIR-V instruction.  The derived class have getter methods
for each operand.

spvtools::val::Instruction now contains a spvtools::Instruction instead
of word and operand vectors.

The file validate_annotation.cpp is converted to use the new API.


include/spirv-tools/instructions.hpp.inc
```diff
@@ -0,0 +1,5011 @@
+// THIS FILE IS GENERATED
```
dj2: Can we generate this during the build and not check it into the repo?

source/instructions.cpp
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "spirv-tools/instructions.hpp"
+
+namespace spvtools {
+#include "instructions.cpp.inc"
+}
```
dj2: nit:  // namespace spvtools

source/instructions.cpp.inc
```diff
@@ -0,0 +1,512 @@
+// THIS FILE IS GENERATED
```
dj2: ditto

fjhenigman:
design doc  https://docs.google.com/document/d/14lu5C97oGuIwwiL-qtYkWHWo0ec-tlwHt7ttN3NXyQ4/

source/val/validate_annotation.cpp
```diff
@@ -200,278 +202,240 @@ bool DecorationTakesIdParameters(uint32_t type) {
...
-  const auto decoration = inst->GetOperandAs<uint32_t>(1);
-  if (decoration == SpvDecorationSpecId) {
-    const auto target_id = inst->GetOperandAs<uint32_t>(0);
-    const auto target = _.FindDef(target_id);
-    if (!target || !spvOpcodeIsScalarSpecConstant(target->opcode())) {
-      return _.diag(SPV_ERROR_INVALID_ID, inst)
-             << "OpDecorate SpecId decoration target <id> '"
-             << _.getIdName(target_id)
-             << "' is not a scalar specialization constant.";
+struct AnnotationPassHandler
```
dj2: What is an AnnotationPassHandler?

source/val/validate_annotation.cpp
```diff
@@ -200,278 +202,240 @@ bool DecorationTakesIdParameters(uint32_t type) {
...
-    case SpvOpGroupDecorate:
-      if (auto error = ValidateGroupDecorate(_, inst)) return error;
-      break;
-    case SpvOpGroupMemberDecorate:
-      if (auto error = ValidateGroupMemberDecorate(_, inst)) return error;
-      break;
-    default:
-      break;
-  }
+  if (auto error = AnnotationPassHandler{}(inst->inst(), _, inst)) return error;
```
dj2: Ugh, AnnotationPassHandler{}(...) is nasty to read

include/spirv-tools/instructions.hpp.inc
fjhenigman: The build generates it, and I guess it doesn't need to be checked in.  This file does form part of the API, but now that I think about it again, I can't see a reasonable usage where it doesn't get generated before it's needed.

dj2:
Can we split the code which adds the visitor pattern to a separate CL and just add the instruction generation in this CL?

s-perron:
AFAIK @fjhenigman is not longer working on this.  I don't think anybody is picking up this work.  We should just close the PR.  Reopen if I am wrong.

PR #2597, opened 2019-05-17
IGNORE JUST TESTING Add generated instruction API.

fjhenigman:
Adds class spvtools::Instruction and an automatically generated derived
class for each SPIR-V instruction.  The derived class have getter methods
for each operand.

spvtools::val::Instruction now contains a spvtools::Instruction instead
of word and operand vectors.

The file validate_annotation.cpp is converted to use the new API.


s-perron:
AFAIK @fjhenigman is not longer working on this. I don't think anybody is picking up this work. We should just close the PR. Reopen if I am wrong.

PR #2598, opened 2019-05-17
Using the instruction folder to fold OpSpecConstantOp

### LINKED ISSUE #2585: Fold OpSpecConstantOp+OpCompositeInsert

s-perron:
In order to try to reduce code duplication and to be able
to fold more cases, we want to use the instruction folder
when folding an OpSpecConstantOp with constant operands.

A couple other changes are need to make this work.  First
GetDefiningInstruction| in the constant manager is able
to handle |type_id| being logically equivalent to another
type, so we updated the interface, and removed the assert.

Some tests were also updated because we not generate
better code because constants are not duplicated as much
as before.

No need for new tests.  The functionality of the instruction folder is
already tested.  There are tests check that the instruction folder is
being used correctly for OpCompositeExtract and OpVectorShuffle in the
existing test cases.

Fixes #2585.

### DIFF
--- a/source/opt/constants.cpp
+++ b/source/opt/constants.cpp
@@ -185,8 +185,6 @@ Instruction* ConstantManager::BuildInstructionAndAddToModule(
 
 Instruction* ConstantManager::GetDefiningInstruction(
     const Constant* c, uint32_t type_id, Module::inst_iterator* pos) {
-  assert(type_id == 0 ||
-         context()->get_type_mgr()->GetType(type_id) == c->type());
   uint32_t decl_id = FindDeclaredConstant(c, type_id);
   if (decl_id == 0) {
     auto iter = context()->types_values_end();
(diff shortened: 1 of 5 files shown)


source/opt/constants.cpp
```diff
@@ -185,8 +185,6 @@ Instruction* ConstantManager::BuildInstructionAndAddToModule(
 
 Instruction* ConstantManager::GetDefiningInstruction(
     const Constant* c, uint32_t type_id, Module::inst_iterator* pos) {
-  assert(type_id == 0 ||
-         context()->get_type_mgr()->GetType(type_id) == c->type());
```
dnovillo: Won't this start failing if `type_id` is invalid?  Or is that now not possible?

s-perron: If |type_id| is 0, then we just want to use whatever type is found.  There are no invalid values for |type_id|.

dnovillo: > If |type_id| is 0, then we just want to use whatever type is found. There are no invalid values for |type_id|.

Ah, quite right.  Thanks.

s-perron: Also if the type for |c| is not the same, then that will be handled by |FindDeclaredConstant|.

PR #2607, opened 2019-05-21
Enable the base branch to be overridden in the code format check script

kpet:
Default to master.

Signed-off-by: Kevin Petit <kevin.petit@arm.com>

### DIFF
--- a/utils/check_code_format.sh
+++ b/utils/check_code_format.sh
@@ -18,14 +18,16 @@
 #
 # This script assumes to be invoked at the project root directory.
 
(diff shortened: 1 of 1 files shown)


utils/check_code_format.sh
```diff
@@ -18,14 +18,16 @@
 #
 # This script assumes to be invoked at the project root directory.
 
-FILES_TO_CHECK=$(git diff --name-only master | grep -E ".*\.(cpp|cc|c\+\+|cxx|c|h|hpp)$")
+BASE_BRANCH=${BASE_BRANCH:-master}
```
dneto0: Seems ok, but why not make this an argument to the script?

kpet: I did consider it. No real reason beyond wondering whether this would be the preferred way. I've updated the PR.

PR #2613, opened 2019-05-22
linker: Add tests for various type comparisons

blue42u:
This adds a number of tests that check that all types will match to identically written clones during linking, including nearly every Type and some combinations (e.g. Functions of Arrays of Floats). Intent is for use with https://github.com/KhronosGroup/SPIRV-Tools/pull/2580, however that PR focuses on issues with TypeArray whereas these tests are (more) comprehensive and test more subtle (and possibly incorrect) cases.

A number of these tests fail (and are disabled to pass CI checks), many are fixed by the aforementioned PR. Some additional tests involving TypeForwardPointer cause assertion failures, although the cases they describe are more rare.

### DIFF
--- a/test/link/CMakeLists.txt
+++ b/test/link/CMakeLists.txt
@@ -23,5 +23,6 @@ add_spvtools_unittest(TARGET link
        memory_model_test.cpp
        partial_linkage_test.cpp
        unique_ids_test.cpp
+       type_match_test.cpp
   LIBS SPIRV-Tools-opt SPIRV-Tools-link
 )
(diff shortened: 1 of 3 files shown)


blue42u:
@pierremoreau please have a look and let me know if there are any mistakes. The tests are cases where the matched types are identical (byte-for-byte, but not just equivalent where for e.g. the constant’s types differ), so they should be helpful for development on #2580 etc.

pierremoreau:
Thanks, I’ll have a look at it tomorrow.

I have been re-writing the code in #2580 to have all the checking done in `Array::IsSameImpl()` rather than creating a separate function. I finally have something working, so this is going to be quite useful to test it out! (I’ll need to also add the `ForwardPointer` handling, but it should be relatively straightforward compared to the `Array` handling.)

PR #2616, opened 2019-05-22
Added an external dependency on protobufs, to be used by the upcoming spirv-fuzz tool.

afd:
Added an external dependency on protobufs, included when SPIRV_BUILD_FUZZER is defined, so that they can be used by the (upcoming) spirv-fuzz tool.  Also updated the kokoro build scripts, for relevant targets, to clone an appropriate tag of the protobufs repo, and to pass -DSPIRV_BUILD_FUZZER to the configurations for which we intend to ultimately build spirv-fuzz.


external/CMakeLists.txt
```diff
@@ -102,3 +102,15 @@ if (NOT ${SPIRV_SKIP_TESTS})
     endif()
   endif()
 endif()
+
+if(SPIRV_BUILD_FUZZER)
+  set(PROTOBUF_DIR ${CMAKE_CURRENT_SOURCE_DIR}/protobuf/cmake)
+  set(protobuf_BUILD_TESTS OFF CACHE BOOL "Disable protobuf tests")
```
dj2: these protobufs are lower case? weird.

afd: Weird indeed but yes, they are - thanks for approving!

PR #2617, opened 2019-05-22
Allow arrays of out per-primitive builtins for mesh shaders

sparmarNV:
- Allow arrays of out per-primitive builtins for mesh shaders - PrimitiveID, Layer, ViewportIndex.
- We already allow arrays of per-vertex builtins such as Position, PointSize, etc. (see use of similar function ValidateOptionalArrayedF32 for reference).


sparmarNV:
> Please add some tests for the new coverage.

Added a positive and a negative test for mesh builtins

Issue #2612, opened 2019-05-21
SPIR-V optimizer breaks multiple ANGLE tests


ShabbyX:
See this change enabling SPIR-V optimization in ANGLE (currently using SPIRV-Tools commit 47741f05043ad880f1964e0c2a8b6733232feaab):

https://chromium-review.googlesource.com/c/angle/angle/+/1616963

Notice the various failures on different platforms. Some of these failures are due to invalid SPIR-V generated (see e.g. `dEQP.GLES3/functional_shaders_switch_conditional_fall_through_2_static_vertex` failing [here](https://chromium-swarm.appspot.com/task?id=44f3e9d908a3c610)), some are just failures due to SPIR-V change in functionality.

I debugged one such case, `CopyTexImageTest.RGBAToL/ES2_Vulkan` end2end test failing on Android and see this:

Original shader (in `src/libANGLE/renderer/vulkan/shaders/src/ImageCopy.frag`) after preprocessing is the following:

```
#version 450 core

#extension GL_EXT_samplerless_texture_functions : require

layout(set = 0, binding = 0)uniform texture2D src;
layout(location = 0)out vec4 dest;

layout(push_constant)uniform PushConstants {

    ivec2 srcOffset;
    ivec2 destOffset;
    int srcMip;
    int srcLayer;

    bool flipY;

    bool premultiplyAlpha;
    bool unmultiplyAlpha;

    bool destHasLuminance;
    bool destIsAlpha;

    int destDefaultChannelsMask;
} params;

void main()
{
    ivec2 destSubImageCoords = ivec2(gl_FragCoord . xy)- params . destOffset;

    ivec2 srcSubImageCoords = destSubImageCoords;

    if(params . flipY)
        srcSubImageCoords . y = - srcSubImageCoords . y;

    vec4 srcValue = texelFetch(src, params . srcOffset + srcSubImageCoords, params . srcMip);

    if(params . premultiplyAlpha)
    {
        srcValue . rgb *= srcValue . a;
    }
    else if(params . unmultiplyAlpha && srcValue . a > 0)
    {
        srcValue . rgb /= srcValue . a;
    }

    vec4 destValue = vec4(srcValue);

    if(params . destHasLuminance)
    {
        destValue . rg = destValue . ra;
    }
    else if(params . destIsAlpha)
    {
        destValue . r = destValue . a;
    }
    else
    {
        int defaultChannelsMask = params . destDefaultChannelsMask;
[...]

PR #2624, opened 2019-05-24
Handle nested breaks from switches.

s-perron:
There was a recent decision made to allow branches to the merge node of
a switch even if the switch is not the first enclosing construct.  They
can be generated by glslang from break statements in switches.

Dead branch elimination seems to be the only optimization that will
break because of this change, so I will update that optimizations.

The change made are:

- Track switches in structured cfg analysis.
- In Dead branch elimination:
  - Look for nested breaks that will require a switch instruction.
  - Rewrite, but don't delete, switchs that are required even if it
    could be replaced by an unconditional branch.
  - When looking for the first break, consider the merge of a switch
    as well.

See #2612.

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -169,24 +169,43 @@ bool DeadBranchElimPass::MarkLiveBlocks(
 
     if (simplify) {
       modified = true;
-      // Replace with unconditional branch.
-      // Remove the merge instruction if it is a selection merge.
-      AddBranch(live_lab_id, block);
-      context()->KillInst(terminator);
+      // Replace branch with a simpler branch.
+      // Fix up the merge instruction if it is a selection merge.
       Instruction* mergeInst = block->GetMergeInst();
       if (mergeInst && mergeInst->opcode() == SpvOpSelectionMerge) {
-        Instruction* first_break = FindFirstExitFromSelectionMerge(
-            live_lab_id, mergeInst->GetSingleWordInOperand(0),
-            cfgAnalysis->LoopMergeBlock(live_lab_id),
-            cfgAnalysis->LoopContinueBlock(live_lab_id));
-        if (first_break == nullptr) {
-          context()->KillInst(mergeInst);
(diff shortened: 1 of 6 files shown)


source/opt/dead_branch_elim_pass.cpp
```diff
@@ -169,24 +169,43 @@ bool DeadBranchElimPass::MarkLiveBlocks(
...
+          new_operands.push_back({SPV_OPERAND_TYPE_ID, {live_lab_id}});
+          terminator->SetInOperands(std::move(new_operands));
+          context()->UpdateDefUse(terminator);
         } else {
-          mergeInst->RemoveFromList();
-          first_break->InsertBefore(std::unique_ptr<Instruction>(mergeInst));
-          context()->set_instr_block(mergeInst,
-                                     context()->get_instr_block(first_break));
+          // Check if the merge instruction is still needed because of a non
+          // nested break from the construct.  More the merge instruction if it
```
alan-baker: More -> Move

source/opt/struct_cfg_analysis.h
```diff
@@ -68,6 +73,21 @@ class StructuredCFGAnalysis {
   // construct.
   uint32_t LoopContinueBlock(uint32_t bb_id);
 
+  // Returns the id of the header of the innermost switch construct
+  // that contains |bb_id| as long as there is no intervening loop.  Returns |0|
+  // if construct exists.
```
alan-baker: if no construct exists

test/opt/dead_branch_elim_test.cpp
```diff
@@ -2993,6 +2993,94 @@ OpFunctionEnd
...
+               OpBranch %11
+         %11 = OpLabel
+               OpReturn
+               OpFunctionEnd
+)";
+
+  SinglePassRunAndMatch<DeadBranchElimPass>(spirv, true);
+}
+
+TEST_F(DeadBranchElimTest, FoldBrachWithBreakToSwitch) {
```
alan-baker: Brach -> Branch

PR #2630, opened 2019-05-27
Instrument: Add version 2 of record formats

greg-lunarg:
New version has additional word in stage-specific section. Also
some changes in content for tesselation and compute shaders. Either
version can be invoked at pass creation. This is done to ease integration
and updating of validation layers. Version 1 is deprecated and eventually
will go away.

Also sneaking in fix to version 1 compute shaders.


greg-lunarg:
The one CI failure is in a file I did not change. Not sure how that happened. Can you handle that on your end?

s-perron:
Don't worry about that.  The clang format checker change every file that is different from the master branch, even if you change is behind.  If someone else fixed the format for another file in the meantime, then it look like you change broke that file.  It is fine.

source/opt/instrument_pass.cpp
```diff
@@ -169,37 +167,97 @@ void InstrumentPass::GenStageStreamWriteCode(uint32_t stage_idx,
...
+          builder);
+      Instruction* x_inst = builder->AddIdLiteralOp(
+          GetUintId(), SpvOpCompositeExtract, load_id, 0);
+      Instruction* y_inst = builder->AddIdLiteralOp(
+          GetUintId(), SpvOpCompositeExtract, load_id, 1);
+      Instruction* z_inst = builder->AddIdLiteralOp(
+          GetUintId(), SpvOpCompositeExtract, load_id, 2);
+      if (version_ == 1) {
+        // For version 1 format, as a stopgap, pack uvec3 into first word:
+        // x << 21 | y << 10 | z. Second word is unused. (DEPRECATED)
```
s-perron: It is hard to tell what is deprecated.  Is it version 1 or the second word?

include/spirv-tools/optimizer.hpp
```diff
@@ -738,9 +738,10 @@ Optimizer::PassToken CreateCombineAccessChainsPass();
 // |input_length_enable| controls instrumentation of runtime descriptor array
 // references, and |input_init_enable| controls instrumentation of descriptor
 // initialization checking, both of which require input buffer support.
+// |version| specifies the buffer record format.
 Optimizer::PassToken CreateInstBindlessCheckPass(
     uint32_t desc_set, uint32_t shader_id, bool input_length_enable = false,
-    bool input_init_enable = false);
+    bool input_init_enable = false, uint32_t version = 1);
```
s-perron: If version 1 is deprecated, why do we make that the default?  This will encourage people to generate the deprecated version.  We should default to the latest version if we default at all.

source/opt/instrument_pass.cpp
greg-lunarg: I "fixed" the version 1 code which was actually causing a validation error. I then deprecated version 1, announcing that any further fixes would not be done. I will try to make that clearer.

include/spirv-tools/optimizer.hpp
greg-lunarg: This entry point is only used for testing, which is still generally done with version 1 since that is what the validation layer is still using. I did add one test using version 2.

Once the layer switches to 2, then I will update all the tests to 2. Maybe I am slightly premature in deprecating version 1, but I thought everything would be clearer to you if you understood that version 1 was going away eventually.

greg-lunarg:
Thanks for the review!

As for --inst-bind less-check, that is not really intended as a "true" command-line option. I put it in as a convenience for developers who want to call it easily. So I would prefer not to elaborate on it or make it public.

As for a date for removing version 1, it will happen as soon as inst-bindless-check quits using it. There are no other users of instrument_pass at the moment so I don't think we need to warn anyone.

PR #2631, opened 2019-05-27
Add spirv-fuzz tool.

afd:
The current tool can parse basic command-line argument, but generates
a binary identical to the input binary, since no transformations are
yet implemented.

### DIFF
--- a/tools/fuzz/fuzz.cpp
+++ b/tools/fuzz/fuzz.cpp
@@ -1,0 +1,253 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


tools/fuzz/fuzz.cpp
```diff
@@ -0,0 +1,253 @@
...
+  if (json_generation_status != google::protobuf::util::Status::OK) {
+    spvtools::Error(FuzzDiagnostic, nullptr, {},
+                    "Error writing out transformations in JSON format");
+    return 1;
+  }
+
+  std::ofstream transformations_json_file(output_file_prefix + ".json");
+  transformations_json_file << json_string;
+  transformations_json_file.close();
+}
```
dj2: nit: return 0; at the end of the method.

afd: Fixed, thanks.

PR #2636, opened 2019-05-29
Add spirv-fuzz pass to permute blocks.

afd:
The blocks within each function in the module will be permuted in a
randomized manner that respects dominance.

### DIFF
--- a/source/fuzz/fuzzer_pass_permute_blocks.cpp
+++ b/source/fuzz/fuzzer_pass_permute_blocks.cpp
@@ -1,0 +1,87 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_permute_blocks.h"
+
+#include "source/fuzz/transformation_move_block_down.h"
+
+namespace spvtools {
+namespace fuzz {
+
+using opt::IRContext;
+
+FuzzerPassPermuteBlocks::FuzzerPassPermuteBlocks(
+    opt::IRContext* ir_context, FactManager* fact_manager,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, fact_manager, fuzzer_context, transformations) {}
+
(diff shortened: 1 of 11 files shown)


source/fuzz/fuzzer_pass_permute_blocks.cpp
```diff
@@ -0,0 +1,87 @@
...
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_permute_blocks.h"
+
+#include "source/fuzz/transformation_move_block_down.h"
+
+namespace spvtools {
+namespace fuzz {
+
+using opt::IRContext;
```
dj2: nit: remove, always fully qualified.

source/fuzz/transformation_move_block_down.h
```diff
@@ -0,0 +1,48 @@
+#include <utility>
```
dj2: nit: move down

source/fuzz/transformation_move_block_down.h
```diff
@@ -0,0 +1,48 @@
...
+
+namespace spvtools {
+namespace fuzz {
+namespace transformation {
+
+// - |block_id| must be the id of a block b in the given module.
+// - b must not be the first nor last block appearing, in program order,
+//   in a function.
+// - b must not dominate the block that follows it in program order.
+bool IsApplicable(const protobufs::TransformationMoveBlockDown& message,
```
dj2: Isn't this going to conflict with the transform_split_block code? I think you need to add another namespace, or wrap these things in classes. Otherwise you've got two spvtools::fuzz::transform::IsApplicable and Apply methods.

source/fuzz/transformation_move_block_down.h
afd: Ug, sorry - I thought I had fixed all those (I think I made the mistake once and then happened to copy the license from the file in which i'd made it.)

Fixed.

source/fuzz/fuzzer_pass_permute_blocks.cpp
afd: Ah OK - I thought it was alright to do this in a .cpp file.  Will fully qualify.

afd: (Turns out I wasn't even relying on having that using statement...)

source/fuzz/transformation_move_block_down.h
afd: It doesn't conflict because the types of their first arguments are different.  In the case of transform_split_block the first argument is protobufs::TransformationSplitBlock; in this case it is protobufs::TransformMoveBlockDown, and other transformations will IsApplicable and Apply methods but with their own protobuf message type as the first argument.  I was planning to use a single namespace, "transformation", for these methods.

What we really want is a Transformation interface with Apply and IsApplicable methods, and to then have an implementation of this interface for each kind of transformation.  Protobufs doesn't generate code that way, though, hence instead overloading on the first argument.

afd:
Abandoning this PR as I made it from the wrong branch.  Will open another one from the correct branch shortly.

PR #2642, opened 2019-05-30
Add spirv-fuzz pass to permute blocks.

afd:
The blocks within each function in the module will be permuted in a randomized manner that respects dominance.

### DIFF
--- a/source/fuzz/transformation_move_block_down.cpp
+++ b/source/fuzz/transformation_move_block_down.cpp
@@ -1,0 +1,102 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


source/fuzz/transformation_move_block_down.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "source/opt/basic_block.h"
+
+namespace spvtools {
+namespace fuzz {
+namespace transformation {
+
+using opt::BasicBlock;
+using opt::IRContext;
+
+bool IsApplicable(const protobufs::TransformationMoveBlockDown& message,
```
afd: Transcribing from previous conversation:

@dj2: "Isn't this going to conflict with the transform_split_block code? I think you need to add another namespace, or wrap these things in classes. Otherwise you've got two spvtools::fuzz::transform::IsApplicable and Apply methods."

@afd: "It doesn't conflict because the types of their first arguments are different. In the case of transform_split_block the first argument is protobufs::TransformationSplitBlock; in this case it is protobufs::TransformMoveBlockDown, and other transformations will IsApplicable and Apply methods but with their own protobuf message type as the first argument. I was planning to use a single namespace, "transformation", for these methods.  What we really want is a Transformation interface with Apply and IsApplicable methods, and to then have an implementation of this interface for each kind of transformation. Protobufs doesn't generate code that way, though, hence instead overloading on the first argument."

PR #2647, opened 2019-05-31
Add fuzzer pass that adds useful constructs to a module

afd:
This new pass adds some basic ingredients to a module on which future passes are likely to depend, such as boolean constants and some specfic integer and floating-point values.  This is not a fuzzer pass in the true sense in that it does not employ randomization, but it makes sense to define it as a fuzzer pass since it is the first of a number of transformations passes that the fuzzer will run on a module.

### DIFF
--- a/source/fuzz/fuzzer_pass_add_useful_constructs.cpp
+++ b/source/fuzz/fuzzer_pass_add_useful_constructs.cpp
@@ -1,0 +1,182 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_add_useful_constructs.h"
+
+#include "source/fuzz/transformation_add_constant_boolean.h"
+#include "source/fuzz/transformation_add_constant_scalar.h"
+#include "source/fuzz/transformation_add_type_boolean.h"
+#include "source/fuzz/transformation_add_type_float.h"
+#include "source/fuzz/transformation_add_type_int.h"
+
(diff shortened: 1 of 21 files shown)


source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -35,18 +35,86 @@ message TransformationSequence {
 
 message Transformation {
   oneof transformation {
-    TransformationMoveBlockDown move_block_down = 1;
-    TransformationSplitBlock split_block = 2;
+    TransformationAddConstantBoolean add_constant_boolean = 1;
```
alan-baker: I don't know much about protobufs, but are creating a versioning problem by renumbering these?

afd: Yes, I believe so - renumbering means that data that was previously serialised won’t be deserialisable. My thinking was that this could be ignored until the tool is fleshed out with a basic set of transformations (because before then there will not exist any valuable serialised data to be preserved), after which we could do a renumbering leaving some gaps for future transformations.

alan-baker: Are the numbers important or are you just keeping a lexicographical ordering? Maybe don't change them and add a comment stating that new transforms should take the next available number at the end?

afd: I was just keeping a lexicographical ordering.  But really this isn't super important.  So in my latest commit I have changed things so that numerical ordering is used, with new transformations taking on successive numeric ids.  That will make it trivial to see what the next id should be when we have tens of transformation kinds, which is good.

source/fuzz/fuzzer_pass_add_useful_constructs.cpp
```diff
@@ -0,0 +1,182 @@
...
+                                   ->GetRegisteredType(&temp_float_type)
+                                   ->AsFloat();
+  auto float_type_id =
+      GetIRContext()->get_type_mgr()->GetId(registered_float_type);
+  assert(
+      float_type_id &&
+      "The relevant float type should have been added to the module already.");
+  opt::analysis::FloatConstant float_constant(registered_float_type, data);
+  if (!GetIRContext()->get_constant_mgr()->FindConstant(&float_constant)) {
+    protobufs::TransformationAddConstantScalar add_constant_int =
```
alan-baker: Bad variable name for this function.

afd: Fixed, thanks.

PR #2651, opened 2019-06-03
Disallow stores to UBOs

### LINKED ISSUE #2638: validator does not catch stores to UBO
UBOs are read-only.

(Does it catch stores to PushConstant, Input, .... ?)

alan-baker:
Fixes #2638

* Adds a check that errors out if there is a store to a UBO in the
Vulkan environment
  * tests

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -811,10 +811,7 @@ spv_result_t ValidateStore(ValidationState_t& _, const Instruction* inst) {
 
     if (spvIsVulkanEnv(_.context()->target_env) &&
         storage_class == SpvStorageClassUniform) {
-      auto base_ptr = pointer;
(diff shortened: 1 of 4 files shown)


source/val/validate_memory.cpp
```diff
@@ -804,6 +803,27 @@ spv_result_t ValidateStore(ValidationState_t& _, const Instruction* inst) {
              << "OpStore Pointer <id> '" << _.getIdName(pointer_id)
              << "' storage class is read-only";
     }
+
+    if (spvIsVulkanEnv(_.context()->target_env) &&
+        storage_class == SpvStorageClassUniform) {
+      auto base_ptr = pointer;
+      while (base_ptr->opcode() == SpvOpAccessChain) {
```
dneto0: Also need to trace through   OpInBoundsAccessChain and OpCopyObject.

PR #2652, opened 2019-06-03
Optimizer: Handle array type with OpSpecConstantOp length

### LINKED ISSUE #2649: optimzer assert faliure when array size is an OpSpecConstantOp
Example GLSL shader:
#version 450

layout(constant_id = 0) const uint a = 12;
layout(location = 0) out uint o;

void main() {
  uint arr[a+a];
  o = arr[a];
}

producing this SPIR-V:

```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Vertex %main "main" %o
               OpSource GLSL 450
               OpName %main "main"
               OpName %o "o"
               OpName %a "a"
               OpName %arr "arr"
               OpDecorate %o Location 0
               OpDecorate %a SpecId 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
       %uint = OpTypeInt 32 0
%_ptr_Output_uint = OpTypePointer Output %uint
          %o = OpVariable %_ptr_Output_uint Output
          %a = OpSpecConstant %uint 12
         %10 = OpSpecConstantOp %uint IAdd %a %a
%_arr_uint_10 = OpTypeArray %uint %10
%_ptr_Function__arr_uint_10 = OpTypePointer Function %_arr_uint_10
%_ptr_Function_uint = OpTypePointer Function %uint
          %5 = OpLabel
        %arr = OpVariable %_ptr_Function__arr_uint_10 Function
         %15 = OpAccessChain %_ptr_Function_uint %arr %a
         %16 = OpLoad %uint %15
               OpStore %o %16
               OpReturn
               OpFunctionEnd
```

With latest SPIRV-Toosl, and running:

       spirv-opt -O c.spv -o d.spv

I get this assert from a debug build.

spirv-opt: spirv-tools/source/opt/types.cpp:406: spvtools::opt::analysis::Array::Array(spvtools::opt::analysis::Type*, uint32_t, const spvtools::opt::analysis::Type*, spvtools::opt::Operand::OperandData): Assertion `constant_type && constant_type->AsInteger()' failed.

dneto0:
In the OpSpecConstantOp array length case, consider the lengths
equal if they have the same result id.

Preserve the previous behaviour for other cases:
- when it's an OpSpecConstant, only the SpecId decorations are compared.
- when it's an OpConstant, the literal bit values are compared.

Fixes #2649

### DIFF
--- a/source/opt/type_manager.cpp
+++ b/source/opt/type_manager.cpp
@@ -510,16 +510,8 @@ Type* TypeManager::RebuildType(const Type& type) {
     }
     case Type::kArray: {
       const Array* array_ty = type.AsArray();
-      const Type* ele_ty = array_ty->element_type();
-      if (array_ty->length_spec_id() != 0u)
-        rebuilt_ty =
-            MakeUnique<Array>(RebuildType(*ele_ty), array_ty->LengthId(),
-                              array_ty->length_spec_id());
-      else
-        rebuilt_ty =
-            MakeUnique<Array>(RebuildType(*ele_ty), array_ty->LengthId(),
-                              array_ty->length_constant_type(),
-                              array_ty->length_constant_words());
+      rebuilt_ty =
+          MakeUnique<Array>(array_ty->element_type(), array_ty->length_info());
       break;
     }
     case Type::kRuntimeArray: {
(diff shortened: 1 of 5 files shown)


source/opt/type_manager.cpp
```diff
@@ -654,28 +646,32 @@ Type* TypeManager::RecordIfTypeDefinition(const Instruction& inst) {
       const Instruction* length_constant_inst = id_to_constant_inst_[length_id];
       assert(length_constant_inst);
 
-      // If it is a specialised constants, retrieve its SpecId.
+      // If it is a specialised constant, retrieve its SpecId.
+      // Only OpSpecConstantOp has a SpecId.
```
alan-baker: OpSpecConstantOp -> OpSpecConstant

source/opt/types.cpp
```diff
@@ -383,63 +384,72 @@ void SampledImage::GetExtraHashWords(
...
-      length_id_(length_id),
-      length_spec_id_(0u),
-      length_constant_type_(constant_type),
-      length_constant_words_(constant_words) {
-  assert(!type->AsVoid());
-  assert(constant_type && constant_type->AsInteger());
+  assert(length_info.opcode == SpvOpSpecConstant ||
+         length_info.opcode == SpvOpSpecConstantOp ||
+         length_info.opcode == SpvOpConstant);
+  // A spec constant leaf value must have a spec id.
```
alan-baker: This sounds like a strange requirement. Maybe clarify that the spec constant must have a spec id word even if it was not decorated.

source/opt/type_manager.cpp
```diff
@@ -654,28 +646,32 @@ Type* TypeManager::RecordIfTypeDefinition(const Instruction& inst) {
...
       uint32_t spec_id = 0u;
```
alan-baker: Maybe this default should be `~0u`? 0 is reasonable spec id that would be used. This makes undecorated constants look the same as some decorated constants. Another option might be to use two words: one for the spec id and one to indicate if it was decorated.

source/opt/type_manager.cpp
```diff
@@ -654,28 +646,32 @@ Type* TypeManager::RecordIfTypeDefinition(const Instruction& inst) {
...
+      // Determine extra words required to distinguish this array length
+      // from another.
+      std::vector<uint32_t> extra_words;
+      const auto opcode = length_constant_inst->opcode();
+      if (opcode == SpvOpConstant) {
+        extra_words.insert(extra_words.end(),
+                           length_constant_inst->GetOperand(2).words.begin(),
+                           length_constant_inst->GetOperand(2).words.end());
+      } else if (opcode == SpvOpSpecConstant) {
+        extra_words.push_back(spec_id);
```
alan-baker: Can the same spec id be used in multiple decoration? I wonder about the case where the linker is attempting to link two modules with the same spec id, but different default values. What happens if such a module is not specialized?

Should we also include the default value in the extra words?

source/opt/types.cpp
```diff
@@ -383,63 +384,72 @@ void SampledImage::GetExtraHashWords(
...
-        is_same &&
-        length_constant_type_->IsSameImpl(at->length_constant_type_, seen) &&
-        length_constant_words_ == at->length_constant_words_;
+  bool is_same = element_type_->IsSameImpl(at->element_type_, seen);
+  is_same = is_same && HasSameDecorations(that);
+
+  // A valid array size is a numeric constant.
+  // So the cases are:
+  //  OpConstant, or
+  //  OpSpecConstant, with a SpecId decoration, or
```
alan-baker: Is just spec id enough? Related to earlier comments.

source/opt/types.cpp
dneto0: I assumed an OpSpecConstant is always decorated with a SpecId.  I realize that doesn't have to be the case.  I'll have to rethink the logic in Array::IsSameImpl (and related)

dneto0: Addressed in new commit

source/opt/type_manager.cpp
dneto0: I added a boolean flag to say if it was initialized, and distinguished the case based on the flag instead.

source/opt/type_manager.cpp
dneto0: Good point.  I've adde the literal value words in this case too.

source/opt/types.cpp
dneto0: Fixed.

dneto0:
I believe all feedback has been addressed.

dneto0:
Rebased and fixed the assert failure.  Actually tested on a Debug build.

source/opt/types.cpp
dneto0: Hey:  @pierremoreau 
This might partly disable the effect you were looking for in the linker.  If two spec constants have the same spec id but different literal values, then they should compare as unequal.

dneto0: cc: @s-perron who reviewed the original change.

dneto0:
Added more type manager tests.
Fixed GCC warning about lenght_info argument name masking the length_info method name.

source/opt/type_manager.cpp
dneto0: done

dneto0:
@alan-baker tests passed on the last round.

PR #2655, opened 2019-06-04
Instrument: Fix code for version 2 output format.

greg-lunarg:
Correct record size. Also bring version 2 tests up to version 1
equivalence.


greg-lunarg:
Version 1 output format is deprecated and will go away as soon as possible. That includes version 1 tests.

PR #2659, opened 2019-06-06
Fuzzer transformation replace boolean constant

afd:
Add transformation to replace a boolean constant with a numeric comparison.
    
The transformation can, for example, replace "true" with "12.0 > 6.0", if constants for those floating-point values are available.
    
This introduces a new 'id use descriptor' structure, which provides a way to describe a particular use of an id, and which will be heavily used in future transformations.  Describing an id use is trivial if the use occurs in an instruction that itself generates an id, but is less straightforward if the id of interest is used by an instruction such as OpStore that does not have a result id.  The 'id use descriptor' structure caters for such cases.

### DIFF
--- a/source/fuzz/id_use_descriptor.cpp
+++ b/source/fuzz/id_use_descriptor.cpp
@@ -1,0 +1,82 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/id_use_descriptor.h"
+
+namespace spvtools {
+namespace fuzz {
+
+opt::Instruction* transformation::FindInstruction(
+    const protobufs::IdUseDescriptor& descriptor,
+    spvtools::opt::IRContext* context) {
+  for (auto& function : *context->module()) {
+    for (auto& block : function) {
+      bool found_base = block.id() == descriptor.base_instruction_result_id();
+      uint32_t num_ignored = 0;
+      for (auto& instruction : block) {
+        if (instruction.HasResultId() &&
+            instruction.result_id() ==
+                descriptor.base_instruction_result_id()) {
+          assert(!found_base &&
+                 "It should not be possible to find the base instruction "
+                 "multiple times.");
+          found_base = true;
+          assert(num_ignored == 0 &&
+                 "The skipped instruction count should only be incremented "
+                 "after the instruction base has been found.");
+        }
+        if (found_base &&
+            instruction.opcode() == descriptor.target_instruction_opcode()) {
+          if (num_ignored == descriptor.num_opcodes_to_ignore()) {
+            if (descriptor.in_operand_index() >= instruction.NumInOperands()) {
+              return nullptr;
+            }
+            auto in_operand =
+                instruction.GetInOperand(descriptor.in_operand_index());
+            if (in_operand.type != SPV_OPERAND_TYPE_ID) {
+              return nullptr;
+            }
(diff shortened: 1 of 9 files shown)


source/fuzz/id_use_descriptor.cpp
```diff
@@ -0,0 +1,82 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/id_use_descriptor.h"
+
+namespace spvtools {
+namespace fuzz {
+
+opt::Instruction* transformation::FindInstruction(
```
dj2: Can you add a comment here? What instruction are you trying to find?

source/fuzz/transformation_replace_boolean_constant_with_constant_binary.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+  // The id use descriptor must identify some instruction
+  return transformation::FindInstruction(message.id_use_descriptor(),
+                                         context) != nullptr;
+}
+
+opt::Instruction* Apply(
+    const protobufs::TransformationReplaceBooleanConstantWithConstantBinary&
+        message,
+    spvtools::opt::IRContext* context,
```
dj2: I don't think the spvtools:: is needed on these lines

source/fuzz/transformation_replace_boolean_constant_with_constant_binary.cpp
```diff
@@ -0,0 +1,284 @@
...
+    spvtools::fuzz::FactManager* /*unused*/) {
+  opt::analysis::Bool bool_type;
+  opt::Instruction::OperandList operands = {
+      {SPV_OPERAND_TYPE_ID, {message.lhs_id()}},
+      {SPV_OPERAND_TYPE_ID, {message.rhs_id()}}};
+  auto binary_instruction = MakeUnique<opt::Instruction>(
+      context, (SpvOp)message.opcode(),
```
dj2: nit: static_cast<SpvOp>(message.opcode())

source/fuzz/transformation_replace_boolean_constant_with_constant_binary.cpp
```diff
@@ -0,0 +1,284 @@
...
+      transformation::FindInstruction(message.id_use_descriptor(), context);
+
+  // We want to insert the new instruction before the instruction that contains
+  // the use of the boolean, but we need to go backwards one more instruction if
+  // the using instruction is preceded by a merge instruction.
+  auto instruction_before_which_to_insert = instruction_containing_constant_use;
+  {
+    opt::Instruction* previous_node =
+        instruction_before_which_to_insert->PreviousNode();
+    if (previous_node) {
```
dj2: if (previous_node && (previous_node->opcode() == SpvOpLoopMerge || previous_node->opcode() == SpvOpSelectionMerge)) ?

source/fuzz/transformation_replace_boolean_constant_with_constant_binary.h
```diff
@@ -0,0 +1,61 @@
+#include <utility>
```
dj2: move down

source/fuzz/transformation_replace_boolean_constant_with_constant_binary.cpp
afd: Fixed.

source/fuzz/transformation_replace_boolean_constant_with_constant_binary.cpp
afd: Fixed all of these, thanks.

source/fuzz/transformation_replace_boolean_constant_with_constant_binary.h
afd: Fixed.

source/fuzz/transformation_replace_boolean_constant_with_constant_binary.cpp
afd: Done.

source/fuzz/id_use_descriptor.cpp
afd: I had commented the function's prototype in the .h file.  I've elaborated that comment a bit more.  But I think it's better that it remains in the .h file rather than being in the .cpp file.

PR #2660, opened 2019-06-07
Add constant == uniform facts.

afd:
Adds a new (and first) kind of fact to the fact manager, which is that
a specific uniform value is guaranteed to be equal to a specific
constant.  The point of this is that such information (if known to be
true by some external source) can be used by spirv-fuzz to transform
the module in interesting ways that a static compiler cannot reverse
via compile-time analysis.

This change introduces protobuf messages for the fact, and adds
capabilities to the fact manager to store this kind of fact and
provide information about it.

### DIFF
--- a/source/fuzz/fact_manager.cpp
+++ b/source/fuzz/fact_manager.cpp
@@ -12,15 +12,217 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <utility>
-
 #include "source/fuzz/fact_manager.h"
+
+#include "source/fuzz/uniform_buffer_element_descriptor.h"
 #include "source/opt/ir_context.h"
 
 namespace spvtools {
 namespace fuzz {
 
-FactManager::FactManager() = default;
+// The purpose of this struct is to group the fields and data used to represent
+// facts about uniform constants.
+struct FactManager::ConstantUniformFacts {
+  // See method in FactManager which delegates to this method.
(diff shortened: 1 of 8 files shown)


source/fuzz/fact_manager.cpp
```diff
@@ -12,15 +12,217 @@
...
+    data_in_fact.push_back(word);
+  }
+  return data_in_constant == data_in_fact;
+}
+
+std::vector<uint32_t>
+FactManager::ConstantUniformFacts::GetConstantsAvailableFromUniformsForType(
+    opt::IRContext* ir_context, uint32_t type_id) const {
+  std::vector<uint32_t> result;
+  for (auto& type_or_value : ir_context->module()->types_values()) {
```
alan-baker: I think you could avoid the nested loop if you're willing to generate `Constants`, but not instructions.

```
for (auto& fact_and_type_id : facts_and_type_ids) {
  auto type = ir_context()->get_type_mgr()->GetType(fact_and_type_id.second);
  auto constant = ir_context()->get_constant_mgr()->GetConstant(type, fact_and_type_id.first);
  if (auto const_id = ir_context()->get_constant_mgr()->FindDeclaredConstant(constant, fact_and_type_id.first)) {
     result.push_back(const_id);
  }
}
```

afd: Thanks - I have made a change similar to this, and applied a similar change in a few other methods.

PR #2664, opened 2019-06-11
Add replayer tool for spirv-fuzz.

afd:
The replayer takes an existing sequence of transformations and applies them to a module.  Replaying a sequence of transformations that were obtained via fuzzing should lead to an identical module to the module that was fuzzed.  Tests have been added to check for this.

### DIFF
--- a/source/fuzz/replayer.cpp
+++ b/source/fuzz/replayer.cpp
@@ -140,8 +140,6 @@ struct Replayer::Impl {
 
 Replayer::Replayer(spv_target_env env) : impl_(MakeUnique<Impl>(env)) {}
 
-Replayer::~Replayer() = default;
-
 void Replayer::SetMessageConsumer(MessageConsumer c) {
   impl_->consumer = std::move(c);
 }
--- a/source/fuzz/replayer.cpp
+++ b/source/fuzz/replayer.cpp
@@ -1,0 +1,200 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/replayer.h"
+
+#include <utility>
+
+#include "source/fuzz/fact_manager.h"
+#include "source/fuzz/protobufs/spirvfuzz_protobufs.h"
+#include "source/fuzz/transformation_add_constant_boolean.h"
+#include "source/fuzz/transformation_add_constant_scalar.h"
+#include "source/fuzz/transformation_add_dead_break.h"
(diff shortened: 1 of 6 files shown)


source/fuzz/replayer.h
```diff
@@ -0,0 +1,74 @@
...
+  // Constructs a replayer from the given target environment.
+  explicit Replayer(spv_target_env env);
+
+  // Disables copy/move constructor/assignment operations.
+  Replayer(const Replayer&) = delete;
+  Replayer(Replayer&&) = delete;
+  Replayer& operator=(const Replayer&) = delete;
+  Replayer& operator=(Replayer&&) = delete;
+
+  ~Replayer();
```
alan-baker: Is this necessary? The implementation is just the default. Probably best not to include unless it's different.

source/fuzz/replayer.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include "source/util/make_unique.h"
+
+namespace spvtools {
+namespace fuzz {
+
+namespace {
+
+// Returns true if and only if the precondition for |transformation| holds, with
+// respect to the given |context| and |fact_manager|.
+bool IsApplicable(const protobufs::Transformation& transformation,
```
alan-baker: It would likely be more convenient if this and Apply() were functions of a generic Transform class. That would suggest a larger refactoring. I'm not opposed to that occurring at a later date.

tools/fuzz/fuzz.cpp
```diff
@@ -205,15 +220,40 @@ int main(int argc, const char** argv) {
...
-  fuzzer.SetMessageConsumer(spvtools::utils::CLIMessageConsumer);
-  auto fuzz_result_status =
-      fuzzer.Run(binary_in, initial_facts, &binary_out,
-                 &transformations_applied, fuzzer_options);
-  if (fuzz_result_status !=
-      spvtools::fuzz::Fuzzer::FuzzerResultStatus::kComplete) {
-    spvtools::Error(FuzzDiagnostic, nullptr, {}, "Error running fuzzer");
-    return 1;
+  if (status.action == FuzzActions::REPLAY) {
+    std::ifstream existing_transformations_file;
```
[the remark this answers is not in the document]
afd: Done (with slightly different names).

source/fuzz/replayer.h
afd: It does seem to be necessary, yes - if I remove the destructor then I get compile errors due to the use of unique_ptr in the class.  I confess that I do not fully understand the reason for these errors, but I believe explicitly declaring a default destructor really is required here.

source/fuzz/replayer.cpp
afd: Noted, and will consider a refactoring along these lines once the final PRs for the spirv-fuzz prototype have landed.

source/fuzz/replayer.h
alan-baker: Can you move the implementation into the header in that case? Good to know it's not doing anything complicated.

afd: Done.

afd: Actually moving the implementation into the header doesn't work.  This is to do with the use of std::unique_ptr and forward declaration.  It seems that by being inline the destructor gets deleted.

I notice that opt/optimizer.hpp and opt/optimizer.cpp suffer from a similar issue - if the default destructor for Optimizer is moved in to the .hpp file, tools/opt.cpp fails to compile.

PR #2672, opened 2019-06-13
Validate Volatile memory semantics bit

alan-baker:
* Can only be used with Vulkan memory model
* Can only be used with atomics
* Bit setting must match for compare exchange opcodes
* Updated memory semantics checks to allow constant instructions
generally with CooperativeMatrixNV

@dneto0 and @zoddicus I left it disallowed for WebGPU. Should this be changed? 

FYI @jeffbolznv

### DIFF
--- a/source/val/validate_atomics.cpp
+++ b/source/val/validate_atomics.cpp
@@ -175,13 +175,37 @@ spv_result_t AtomicsPass(ValidationState_t& _, const Instruction* inst) {
         return error;
       }
 
-      if (auto error = ValidateMemorySemantics(_, inst, operand_index++))
+      const auto equal_semantics_index = operand_index++;
+      if (auto error = ValidateMemorySemantics(_, inst, equal_semantics_index))
         return error;
 
       if (opcode == SpvOpAtomicCompareExchange ||
           opcode == SpvOpAtomicCompareExchangeWeak) {
-        if (auto error = ValidateMemorySemantics(_, inst, operand_index++))
+        const auto unequal_semantics_index = operand_index++;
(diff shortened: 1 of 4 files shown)


jeffbolznv:
LGTM

source/val/validate_atomics.cpp
```diff
@@ -175,13 +175,37 @@ spv_result_t AtomicsPass(ValidationState_t& _, const Instruction* inst) {
...
+        bool is_unequal_const = false;
+        uint32_t equal_value = 0;
+        uint32_t unequal_value = 0;
+        std::tie(is_int32, is_equal_const, equal_value) = _.EvalInt32IfConst(
+            inst->GetOperandAs<uint32_t>(equal_semantics_index));
+        std::tie(is_int32, is_unequal_const, unequal_value) =
+            _.EvalInt32IfConst(
+                inst->GetOperandAs<uint32_t>(unequal_semantics_index));
+        if (is_equal_const && is_unequal_const &&
+            ((equal_value & SpvMemorySemanticsVolatileMask) ^
```
jeffbolznv: I think the new flag will end with "KHRMask"

dneto0: We put it in without the KHR part because the semantics won't change. 
https://github.com/KhronosGroup/SPIRV-Headers/pull/119#discussion_r293626633

dneto0:
I'm trying to fix the bot failures with https://github.com/KhronosGroup/SPIRV-Tools/pull/2675

dneto0:
Please rebase to get bots to pass.

PR #2673, opened 2019-06-13
Whitelist SPV_GOOGLE_user_type extension

ehsannas:


dneto0:
The smoketest failure in glslc-tests is fixed upstream by now.
The MSVC failures is to be fixed in https://github.com/KhronosGroup/SPIRV-Tools/pull/2675

ehsannas:
Thanks @dneto0

dneto0:
Sorry, this is going to need another rebase to pick up #2676

PR #2675, opened 2019-06-14
Fix uint -> uint32_t in fuzz.cpp

dneto0:
Fixes MSVC builds


dneto0:
Where did all the bots go?  Why are all the kokoro MSVC bots gone?  I'm merging this because this should unblock other builds.

PR #2678, opened 2019-06-18
AggressiveDCE: Retain all bindings

troughton:
Some background: my shader render pipeline uses SPIRV-Cross to convert SPIR-V into Metal shaders. For resource binding, argument buffers are used, mapping directly to Vulkan descriptor sets. Crucially, these argument buffers are shared across pipeline stages; each argument buffer may contain resources relevant to both the vertex and fragment stage.

As a result of this, I need all bindings to be visible to all shader stages, even if the binding may be unused in that stage: the argument buffer definition must be complete. To achieve this, the AggressiveDCE pass must not strip instructions with binding decorations. (Additional changes are also needed for SPIRV-Cross to support this; see https://github.com/troughton/SPIRV-Cross/commit/c3daef79b89caf3a85a7e7d7d5775dfb86babad6)

Note that simply not running the pass isn't an option; my source shaders are in HLSL so need to undergo legalisation.

I don't expect this to be merged as-is (since that would obviously be undesirable for code-size in a number of shaders), but I welcome suggestions for how retaining bindings should be added as an option – should it be added an option to AggressiveDCE, or should a separate pass be used? How should the option be exposed in the driver?

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -570,11 +570,18 @@ void AggressiveDCEPass::InitializeModuleScopeLiveInstructions() {
       AddToWorklist(&entry);
     }
   }
-  // Keep workgroup size.
+  
   for (auto& anno : get_module()->annotations()) {
     if (anno.opcode() == SpvOpDecorate) {
+      // Keep workgroup size.
       if (anno.GetSingleWordInOperand(1u) == SpvDecorationBuiltIn &&
           anno.GetSingleWordInOperand(2u) == SpvBuiltInWorkgroupSize) {
+        AddToWorklist(&anno);
+      }
+      
+      // Keep all bindings.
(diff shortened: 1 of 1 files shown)


s-perron:
Sorry, I though I had responded earlier.  For the design, I would suggest adding a flag that is passed as a parameter to the constructor for [`AggressiveDCEPass`](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/aggressive_dead_code_elim_pass.h#L47).  Your code should be guarded by that flag.

For the public interface, [`CreateAggressiveDCEPass`](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/include/spirv-tools/optimizer.hpp#L529) should take the flag as a parameter as well, and it should default to the original behaviour.  We do not want to force others to have to change their code.

Lastly add a new suboption to [eliminate-dead-code-aggressive](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/optimizer.cpp#L321).  See scalar replacement for how to do that.  Also, update the documentation for the option in `tools/opt/opt.cpp`.

PR #2679, opened 2019-06-18
Basic validation for Component decorations

jozefkucia:


source/val/validate_decorations.cpp
```diff
@@ -1393,6 +1393,76 @@ spv_result_t CheckIntegerWrapDecoration(ValidationState_t& vstate,
...
+// decoration.  Otherwise emits a diagnostic and returns something other than
+// SPV_SUCCESS.
+spv_result_t CheckComponentDecoration(ValidationState_t& vstate,
+                                      const Instruction& inst,
+                                      const Decoration& decoration) {
+  assert(inst.id() && "Parser ensures the target of the decoration has an ID");
+
+  uint32_t type_id;
+  if (decoration.struct_member_index() == Decoration::kInvalidMember) {
+    // The target must be a memory object declaration.
```
dneto0: That can include an OpFunctionParameter of pointer type.
(See similar example for NonWritable validation).

source/val/validate_decorations.cpp
```diff
@@ -1393,6 +1393,76 @@ spv_result_t CheckIntegerWrapDecoration(ValidationState_t& vstate,
...
+        !vstate.IsFloatScalarOrVectorType(type_id)) {
+      return vstate.diag(SPV_ERROR_INVALID_ID, &inst)
+             << "Component decoration specified for type "
+             << vstate.getIdName(type_id) << " that is not a scalar or vector";
+    }
+
+    // For 16-, and 32-bit types, it is invalid if this sequence of components
+    // gets larger than 3.
+    const auto bit_width = vstate.GetBitWidth(type_id);
+    if (bit_width != 16 || bit_width != 32) {
```
dneto0: This condition does not match the comment.  
In fact, this condition always evaluates to true.

source/val/validate_decorations.cpp
jozefkucia: I think it should be fixed by ce8e502

test/val/val_decoration_test.cpp
```diff
@@ -6400,6 +6400,315 @@ OpDecorate %1 BufferBlock
...
+
+  CompileSuccessfully(spirv, env);
+  EXPECT_EQ(SPV_SUCCESS, ValidateAndRetrieveValidationState(env));
+  EXPECT_THAT(getDiagnosticString(), Eq(""));
+}
+
+TEST_F(ValidateDecorations, ComponentDecorationIntGood3Vulkan) {
+  const spv_target_env env = SPV_ENV_VULKAN_1_0;
+  std::string spirv =
+      ShaderWithComponentDecoration("OpTypeInt 32 0", "Component 2");
```
alan-baker: This is the same as the previous test.

jozefkucia: I changed to `Component 3`

PR #2683, opened 2019-06-19
Add infrastructure for maintaining and using DEPS

### LINKED ISSUE #2665: Either remove DEPS or make it the one true way
My understanding is that the current expectation for spirv-tools is to use HEAD from all dependencies and fix things as they break. This is done via manually git pulling all the repos periodically.

There is no automated mechanism to use or update the DEPS file provided, so the values I believe have gone stale. I think we should either remove it to avoid confusing people or put in the place the tooling to make it the preferred path.

zoddicus:
This adds in a script for updating DEPS and another for pulling in the
specified versions in the DEPS file.

DEPS entries that are not part of the documented build processs are
removed. Using the documented cmake process with them made for a very
sad state of one's checkout. If this removal breaks workflows we can
revist this, since the state is saved in git.

DEPS are rolled forward, so that this is taking use to a known good
state. These look like scary big rolls, but in reality the bots and
most people are running near HEAD for external/, so this integration
should already have been tested.

Summary of DEPS roll:
+ Rolling 3 dependencies
+ Roll external/effcee/ 04b624799..b83b58d17 (14 commits)
+ Roll external/googletest/ 98a0d007d..2f42d769a (576 commits)
+ Roll external/re2/ 6cf8ccd82..848dfb7e1 (90 commits)
+ Created with:
+   roll-dep external/effcee external/googletest external/re2 external/spirv-headers

Fixes #2665

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -1,174 +1,26 @@
 use_relative_paths = True
 
 vars = {
-  'chromium_git': 'https://chromium.googlesource.com',
   'github': 'https://github.com',
 
-  'build_revision': '037f38ae0fe5e11b4f7c33b750fd7a1e9634a606',
-  'buildtools_revision': 'ab7b6a7b350dd15804c87c20ce78982811fdd76f',
-  'clang_revision': 'abe5e4f9dc0f1df848c7a0efa05256253e77a7b7',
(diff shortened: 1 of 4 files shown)


zoddicus:
Woops, bad config in my checkout, I will fix and re-upload

PR #2685, opened 2019-06-20
Validate that in OpenGL env block variables have Binding

cmarcelo:


cmarcelo:
I've updated fixing the formatting issue.

PR #2690, opened 2019-06-20
Avoid using a global static map.

ehsannas:
It is better to use a "getter" function I believe according to the C++
coding guide.

The existence of the global static caused a weird startup issue in DXC (which includes SPIRV-Tools as
an external project).

### DIFF
--- a/source/spirv_target_env.cpp
+++ b/source/spirv_target_env.cpp
@@ -104,39 +104,42 @@ uint32_t spvVersionForTargetEnv(spv_target_env env) {
   return SPV_SPIRV_VERSION_WORD(0, 0);
 }
 
-static const std::map<std::string, spv_target_env> spvTargetEnvNameMap = {
-    {"vulkan1.1spv1.4", SPV_ENV_VULKAN_1_1_SPIRV_1_4},
-    {"vulkan1.0", SPV_ENV_VULKAN_1_0},
-    {"vulkan1.1", SPV_ENV_VULKAN_1_1},
-    {"spv1.0", SPV_ENV_UNIVERSAL_1_0},
-    {"spv1.1", SPV_ENV_UNIVERSAL_1_1},
-    {"spv1.2", SPV_ENV_UNIVERSAL_1_2},
-    {"spv1.3", SPV_ENV_UNIVERSAL_1_3},
-    {"spv1.4", SPV_ENV_UNIVERSAL_1_4},
-    {"opencl1.2embedded", SPV_ENV_OPENCL_EMBEDDED_1_2},
-    {"opencl1.2", SPV_ENV_OPENCL_1_2},
-    {"opencl2.0embedded", SPV_ENV_OPENCL_EMBEDDED_2_0},
-    {"opencl2.0", SPV_ENV_OPENCL_2_0},
-    {"opencl2.1embedded", SPV_ENV_OPENCL_EMBEDDED_2_1},
-    {"opencl2.1", SPV_ENV_OPENCL_2_1},
-    {"opencl2.2embedded", SPV_ENV_OPENCL_EMBEDDED_2_2},
-    {"opencl2.2", SPV_ENV_OPENCL_2_2},
-    {"opengl4.0", SPV_ENV_OPENGL_4_0},
-    {"opengl4.1", SPV_ENV_OPENGL_4_1},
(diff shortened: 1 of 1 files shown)


source/spirv_target_env.cpp
```diff
@@ -104,39 +104,42 @@ uint32_t spvVersionForTargetEnv(spv_target_env env) {
...
-    {"opencl2.2embedded", SPV_ENV_OPENCL_EMBEDDED_2_2},
-    {"opencl2.2", SPV_ENV_OPENCL_2_2},
-    {"opengl4.0", SPV_ENV_OPENGL_4_0},
-    {"opengl4.1", SPV_ENV_OPENGL_4_1},
-    {"opengl4.2", SPV_ENV_OPENGL_4_2},
-    {"opengl4.3", SPV_ENV_OPENGL_4_3},
-    {"opengl4.5", SPV_ENV_OPENGL_4_5},
-    {"webgpu0", SPV_ENV_WEBGPU_0},
-};
+const std::map<std::string, spv_target_env> getSpvTargetEnvNameMap() {
```
s-perron: This should return a reference to the map.

source/spirv_target_env.cpp
```diff
@@ -104,39 +104,42 @@ uint32_t spvVersionForTargetEnv(spv_target_env env) {
...
+      {"opencl2.1embedded", SPV_ENV_OPENCL_EMBEDDED_2_1},
+      {"opencl2.1", SPV_ENV_OPENCL_2_1},
+      {"opencl2.2embedded", SPV_ENV_OPENCL_EMBEDDED_2_2},
+      {"opencl2.2", SPV_ENV_OPENCL_2_2},
+      {"opengl4.0", SPV_ENV_OPENGL_4_0},
+      {"opengl4.1", SPV_ENV_OPENGL_4_1},
+      {"opengl4.2", SPV_ENV_OPENGL_4_2},
+      {"opengl4.3", SPV_ENV_OPENGL_4_3},
+      {"opengl4.5", SPV_ENV_OPENGL_4_5},
+      {"webgpu0", SPV_ENV_WEBGPU_0},
```
s-perron: Sorry I did not check this closely at first.  The google style guide does not allow static variables that are not trivially destructible.  https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

This static is not valid.  Also, you cannot make the static variable a pointer to an std::map unless you make sure it gets deleted.

I think the best option would be to make it a sorted array of pairs, and do a binary search to find the correct element.

ehsannas: My understanding of the google style guide was that it is OK to use static inside a function, even if it is not trivially destructible 🤔

ehsannas:
looks like this change fails the `spirv-tools-symbol-exports-SPIRV-Tools` test only on Linux

source/spirv_target_env.cpp
ehsannas: it has this as an example:
```
void foo() {
  static const char* const kMessages[] = {"hello", "world"};  // allowed
}
```

ehsannas: maybe array of `const char*` is considered trivially destructible? :)

s-perron: Yes, the pointer is trivially destructible.

dneto0: whoops, my bad on letting this pass original review.
Can we make it a std::array to still get the benefit of auto-iteration?

source/spirv_target_env.cpp
dneto0: This is only used in this file, so put it in the anonymous namespace, or declare the _function_ static.

source/spirv_target_env.cpp
dneto0: Also I std::array is trivially destructible (its desctructor is implicitly declared).
So the style guide permits a static-duration std::array . 

So the only extra thing you have to do is the lookup to replace the logic in lines 137-140

kpet:
Sorry for the disruption, I believe https://github.com/KhronosGroup/SPIRV-Tools/pull/2691 should solve the issues.

dneto0:
#2691 fixed the issue. Closing.

ehsannas:
Thanks @kpet for your fast response and the fix :)

PR #2692, opened 2019-06-21
Fix BUILD.gn for Fuchsia platform build.

digit-google:
In order for the Fuchsia source tree to update its
version of SPIRV-Tools to a newer upstream, the
BUILD.gn needs to be slightly altered to take care
of the fact that it can be used with a different
GN //build set of rules and configs than the
Chromium one.

This is done by using |build_with_chromium|, already
defined by //build_overrides/build.gni, to guard
Chromium-specific statements.

+ Add a target to generate spirv-opt which is used by
  Fuchsia to optimize shaders at build time for some of
  its graphics libraries.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 import("//build_overrides/spirv_tools.gni")
-
-import("//testing/test.gni")
 import("//build_overrides/build.gni")
+if (build_with_chromium) {
+  import("//testing/test.gni")
+}
 
 spirv_headers = spirv_tools_spirv_headers_dir
 
@@ -379,9 +380,11 @@ static_library("spvtools") {
     ":spvtools_headers",
   ]
 
-  configs -= [ "//build/config/compiler:chromium_code" ]
+  if (build_with_chromium) {
+    configs -= [ "//build/config/compiler:chromium_code" ]
+    configs += [ "//build/config/compiler:no_chromium_code" ]
+  }
   configs += [
-    "//build/config/compiler:no_chromium_code",
     ":spvtools_internal_config",
   ]
 }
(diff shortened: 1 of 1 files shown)


digit-google:
For the record, I've signed the CLA (multiple times). Does anyone knows why this is ignored? Thanks.

BUILD.gn
```diff
@@ -771,3 +771,21 @@ executable("spirv-as") {
   ]
   configs += [ ":spvtools_internal_config" ]
 }
+
+}
```
digit-google: mmm, this should not be here !? I'll upload a fix.

zoddicus:
> For the record, I've signed the CLA (multiple times). Does anyone knows why this is ignored? Thanks.

If you have signed it since it originally posed, you may have to click the link about re-checking it.

For me atleast, this often this occurs when I have set the patch author in git as a different account then my GitHub one. Make sure that the user.email in your .git/config is set to the account associated with your GitHub account. You may have to amend the patch to update the author.

digit-google:
> > For the record, I've signed the CLA (multiple times). Does anyone knows why this is ignored? Thanks.
> 
> If you have signed it since it originally posed, you may have to click the link about re-checking it.
> 
> For me atleast, this often this occurs when I have set the patch author in git as a different account then my GitHub one. Make sure that the user.email in your .git/config is set to the account associated with your GitHub account. You may have to amend the patch to update the author.

Thank you, that's very likely to be the cause for this problem. My github email uses a +github@google.com suffix to help filtering emails :-/ This explains why the rechecks didn't do anything.

I will have to to amend my patch then. Wow!

digit-google:
I totally understand and agree. Thanks and sorry.

zoddicus:
> I totally understand and agree. Thanks and sorry.

Does this PR need any changes, like the shaderc one did?

digit-google:
Yes, I'm preparing a patch. Sorry about that...

zoddicus:
I think this still needs a fresh version with the correct author for the CLA check.

digit-google:
Ah yes, there was a typo in the ameneded author name in the latest commit (I used @google.cm instead of @google.com) and couldn't see it clearly. I was really puzzled why this wasn't working. I'll fix it. Thanks!

PR #2693, opened 2019-06-22
Add —preserve-bindings and —preserve-spec-constants

troughton:
Continuing from https://github.com/KhronosGroup/SPIRV-Tools/pull/2678, a more complete implementation of flags to preserve all bindings and/or specialisation constants within a SPIR-V module, which can be useful/necessary for reflection or for cross-compilation into other shader languages (e.g. for Metal argument buffers).

@s-perron I've implemented this as an optimiser option rather than a flag specifically for AggressiveDCE. At the moment, I think AggressiveDCE is the only pass that _can_ remove unused bindings/spec-constants; however, if future passes started to do the same I'd imagine we'd want a single flag to control it. Does that sound okay?

I've also enabled the flags by default for HLSL legalisation, since some tools (e.g. `dxc` perform legalisation by default on HLSL source even when optimisation is disabled, and preserving the bindings/spec-constants in an optimisation-disabled environment seems the right option.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -570,12 +570,27 @@ void AggressiveDCEPass::InitializeModuleScopeLiveInstructions() {
       AddToWorklist(&entry);
     }
   }
-  // Keep workgroup size.
   for (auto& anno : get_module()->annotations()) {
     if (anno.opcode() == SpvOpDecorate) {
+      // Keep workgroup size.
       if (anno.GetSingleWordInOperand(1u) == SpvDecorationBuiltIn &&
           anno.GetSingleWordInOperand(2u) == SpvBuiltInWorkgroupSize) {
         AddToWorklist(&anno);
+      }
+
+      if (context()->preserve_bindings()) {
+        // Keep all bindings.
+        if ((anno.GetSingleWordInOperand(1u) == SpvDecorationDescriptorSet) ||
+            (anno.GetSingleWordInOperand(1u) == SpvDecorationBinding)) {
+          AddToWorklist(&anno);
+        }
+      }
+
+      if (context()->preserve_spec_constants()) {
+        // Keep all specialization constant instructions
+        if (anno.GetSingleWordInOperand(1u) == SpvDecorationSpecId) {
+          AddToWorklist(&anno);
+        }
       }
     }
   }
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -100,7 +100,9 @@ class IRContext {
         constant_mgr_(nullptr),
         type_mgr_(nullptr),
         id_to_name_(nullptr),
-        max_id_bound_(kDefaultMaxIdBound) {
+        max_id_bound_(kDefaultMaxIdBound),
+        preserve_bindings_(false),
+        preserve_spec_constants_(false) {
     SetContextMessageConsumer(syntax_context_, consumer_);
     module_->SetContext(this);
   }
@@ -115,7 +117,9 @@ class IRContext {
         valid_analyses_(kAnalysisNone),
         type_mgr_(nullptr),
         id_to_name_(nullptr),
-        max_id_bound_(kDefaultMaxIdBound) {
+        max_id_bound_(kDefaultMaxIdBound),
+        preserve_bindings_(false),
+        preserve_spec_constants_(false) {
     SetContextMessageConsumer(syntax_context_, consumer_);
     module_->SetContext(this);
     InitializeCombinators();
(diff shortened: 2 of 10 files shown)


source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -826,9 +841,19 @@ bool AggressiveDCEPass::ProcessGlobalValues() {
...
+                                     preserve_bindings_(preserve_bindings),
+                                     preserve_spec_constants_(preserve_spec_constants) {}
 
 Pass::Status AggressiveDCEPass::Process() {
+  // Copy preserve_bindings and preserve_spec_constants from the context.
+  // We OR the arguments together with the arguments passed into the constructor
+  // since the arguments in the constructor are per-pass overrides
+  // (e.g. we shouldn't strip bindings or spec constants for HLSL legalization).
+  preserve_bindings_ = preserve_bindings_ || context()->preserve_bindings();
+  preserve_spec_constants_ = preserve_spec_constants_ || context()->preserve_spec_constants();
```
s-perron: I don't like having two ways of doing the same thing.  Either have the options to ADCE or have a flag in the context.  I think the flag in the context is a good idea.

source/opt/aggressive_dead_code_elim_pass.h
```diff
@@ -44,7 +44,12 @@ class AggressiveDCEPass : public MemPass {
   using GetBlocksFunction =
       std::function<std::vector<BasicBlock*>*(const BasicBlock*)>;
 
-  AggressiveDCEPass();
+  // preserve_bindings and preserve_spec_constants here serve as overrides
+  // for context()->preserve_bindings() and context->preserve_spec_constants(),
+  // since e.g. when legalizing HLSL we never want to strip bindings or
+  // specialization constants.
```
s-perron: When we did all of the work for legalization, we had the specific requirement that we were to remove unused resources.  It takes time to bind resources, so if the resource is unused, you can save time by not binding the resource.

Correct me if I am wrong, but removing the binding was also the default behaviour for FXC and DXC when generating DXIL.  That should be the default behaviour.

source/opt/optimizer.cpp
```diff
@@ -122,16 +122,16 @@ Optimizer& Optimizer::RegisterLegalizationPasses() {
           // Propagate the value stored to the loads in very simple cases.
           .RegisterPass(CreateLocalSingleBlockLoadStoreElimPass())
           .RegisterPass(CreateLocalSingleStoreElimPass())
-          .RegisterPass(CreateAggressiveDCEPass())
+          .RegisterPass(CreateAggressiveDCEPass(true, true))
```
s-perron: Please revert this and all of the similar changes.

ehsannas:
@s-perron is right. Please don't change the default behavior. Unused bindings are removed by DXC when generating DXIL. And we want to have the similar behavior for the SPIR-V backend. Here is an [example](http://shader-playground.timjones.io/511e72847b4ac9391f5e8b8a2e82dcaa). (Notice that only `sArr` and `tArr` have bindings regardless of whether or not optimization is performed and regardless of optimization level). I believe this is mainly for efficiency/performance reasons (and I'm not sure if all drivers particularly play nice when there are unused bindings).

troughton:
To be clear, what I’ve changed here only disables removing bindings during _legalisation_. This matches the behaviour of passing `-Od` to dxc when legalisation is not required, which causes all bindings to be preserved. Optimisation will still strip bindings, and the default behaviour doesn’t change with this patch.

Regardless, I can make the changes you’ve suggested, but it does mean I’ll also need to add the same options to dxc.

ehsannas:
@troughton Oh I may have misunderstood your solution then. Let me take a second look. Thanks

ehsannas:
@troughton I still believe it'd be best not to change the default behavior of the legalization phase due to flows that expect unused bindings to be removed by the legalizer.

In terms of command line options for DXC: We strongly try not to add command line options to DXC unless absolutely necessary. DXC does have a command line option named `-Oconfig=` which can control the flags that DXC passes to spirv-opt. We should try to leverage this option rather than adding new command line options.

s-perron:
I think we could use the -Oconfig option.  It would be important to make sure the function `Optimizer::RegisterPassesFromFlags` is able to parse this spirv-opt option that is added.

troughton:
I've made most of the changes you requested. I haven't modified `Optimizer:: RegisterPassesFromFlags`; to me, that feels like the wrong place to detect these flags; the optimiser doesn't currently have any knowledge of the optimiser flags until later in the pipeline. 

I'm not sure of the best approach for this; I can work around it locally by disabling validation and then using a separate invocation to `spirv-opt`, but we probably don't want people to have to do that in general.

source/opt/ir_context.h
```diff
@@ -490,6 +494,16 @@ class IRContext {
 
   uint32_t max_id_bound() const { return max_id_bound_; }
   void set_max_id_bound(uint32_t new_bound) { max_id_bound_ = new_bound; }
+    
+  bool preserve_bindings() const { return preserve_bindings_; }
+  void set_preserve_bindings(bool preserve_bindings) {
```
s-perron: GCC complains because `preserve_bindings` is the name of this parameter and the function above.  Can you fix that?

source/opt/ir_context.h
```diff
@@ -490,6 +494,16 @@ class IRContext {
...
+    preserve_bindings_ = preserve_bindings;
+  }
+  
+  bool preserve_spec_constants() const { return preserve_spec_constants_; }
+  void set_preserve_spec_constants(bool preserve_spec_constants) {
```
s-perron: Shadow variable here too.

s-perron:
I'll fix up the formatting and merge it.

PR #2704, opened 2019-06-26
Validate usage of 8- and 16-bit types with only storage capabilities

### LINKED ISSUE #2669: Better validation of 8- and 16- bit storage capabilities
Validate the limitations related to 8- and 16-bit storage capabilities.

Refer to internal SPIR issue 319.

alan-baker:
Fixes #2669 

* Validates use cases for 8- and 16-bit types when only storage capabilities are present
  * OpVariable - specific capability check for storage class
  * OpLoad - only scalar vector and matrix (and pointers)
  * OpStore - only scalar vector and matrix (and pointers)
  * Conversions - width-only conversions
  * Constants - only pointer null
  * Composites - none
* Unified some conversions tests to cover more cases
* Some new helper functions in ValidationState_t to identify the right types and use cases

### DIFF
--- a/source/val/validate_memory.cpp
+++ b/source/val/validate_memory.cpp
@@ -428,8 +428,7 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
     }
   }
 
-  auto storage_class =
-      inst->GetOperandAs<SpvStorageClass>(storage_class_index);
+  auto storage_class = inst->GetOperandAs<SpvStorageClass>(storage_class_index);
   if (storage_class != SpvStorageClassWorkgroup &&
       storage_class != SpvStorageClassCrossWorkgroup &&
       storage_class != SpvStorageClassPrivate &&
@@ -719,7 +718,8 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
       auto underlying_type = value_type;
       while (underlying_type->opcode() == SpvOpTypePointer) {
         storage_class = underlying_type->GetOperandAs<SpvStorageClass>(1u);
-        underlying_type = _.FindDef(underlying_type->GetOperandAs<uint32_t>(2u));
+        underlying_type =
+            _.FindDef(underlying_type->GetOperandAs<uint32_t>(2u));
       }
       bool storage_class_ok = true;
       std::string sc_name = _.grammar().lookupOperandName(
@@ -773,7 +773,8 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
       auto underlying_type = value_type;
       while (underlying_type->opcode() == SpvOpTypePointer) {
         storage_class = underlying_type->GetOperandAs<SpvStorageClass>(1u);
-        underlying_type = _.FindDef(underlying_type->GetOperandAs<uint32_t>(2u));
+        underlying_type =
+            _.FindDef(underlying_type->GetOperandAs<uint32_t>(2u));
       }
       bool storage_class_ok = true;
       std::string sc_name = _.grammar().lookupOperandName(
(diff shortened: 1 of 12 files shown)


alan-baker:
The failing glslang tests are all generating 8- or 16-bit constants with just the storage capabilities.

source/val/validate_memory.cpp
```diff
@@ -712,6 +708,115 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
...
+        case SpvStorageClassStorageBuffer:
+        case SpvStorageClassPhysicalStorageBufferEXT:
+          if (!_.HasCapability(SpvCapabilityStorageBuffer16BitAccess)) {
+            storage_class_ok = false;
+          }
+          break;
+        case SpvStorageClassUniform:
+          if (!_.HasCapability(
+                  SpvCapabilityUniformAndStorageBuffer16BitAccess)) {
+            if (value_type->opcode() == SpvOpTypeArray ||
```
dneto0: This looks wrong.  You're changing the value type from enclosing struct to the contained array/RTArray. Then checking for the decoration on the array type.  But the BufferBlock decoration goes on the struct.
Seems like this whole "if" clause should be removed.

source/val/validate_memory.cpp
```diff
@@ -712,6 +708,115 @@ spv_result_t ValidateVariable(ValidationState_t& _, const Instruction* inst) {
...
+        case SpvStorageClassStorageBuffer:
+        case SpvStorageClassPhysicalStorageBufferEXT:
+          if (!_.HasCapability(SpvCapabilityStorageBuffer8BitAccess)) {
+            storage_class_ok = false;
+          }
+          break;
+        case SpvStorageClassUniform:
+          if (!_.HasCapability(
+                  SpvCapabilityUniformAndStorageBuffer8BitAccess)) {
+            if (value_type->opcode() == SpvOpTypeArray ||
```
dneto0: same here.

source/val/validate_memory.cpp
```diff
@@ -1017,7 +1146,20 @@ spv_result_t ValidateCopyMemory(ValidationState_t& _, const Instruction* inst) {
 
     if (auto error = CheckMemoryAccess(_, inst, 3)) return error;
   }
-  return ValidateCopyMemoryMemoryAccess(_, inst);
+  if (auto error = ValidateCopyMemoryMemoryAccess(_, inst)) return error;
+
+  // Get past the pointers to avoid checking a pointer copy.
+  auto sub_type = _.FindDef(target_pointer_type->GetOperandAs<uint32_t>(2));
+  while (sub_type->opcode() == SpvOpTypePointer) {
```
dneto0: Arguable this bit of logic -- to trace down past pointers -- could be a common method.  But no request to change this now.

source/val/validation_state.cpp
```diff
@@ -1226,5 +1226,55 @@ const Instruction* ValidationState_t::TracePointer(
...
+    return inst->GetOperandAs<uint32_t>(1u) == width;
+  }
+
+  switch (inst->opcode()) {
+    case SpvOpTypeArray:
+    case SpvOpTypeRuntimeArray:
+    case SpvOpTypeVector:
+    case SpvOpTypeMatrix:
+    case SpvOpTypeImage:
+    case SpvOpTypeSampledImage:
```
dneto0: This sampled image case looks dodgy.   But "Contains" here really means "references or contains"

dneto0: Add the case for OpTypeCooperativeMatrixNV

source/val/validation_state.h
```diff
@@ -569,6 +569,14 @@ class ValidationState_t {
   bool IsIntCooperativeMatrixType(uint32_t id) const;
   bool IsUnsignedIntCooperativeMatrixType(uint32_t id) const;
 
+  // Returns true if |id| contains |type| (or integer or floating point type)
```
dneto0: Should say that |id| is the id of a type.

source/val/validation_state.h
```diff
@@ -569,6 +569,14 @@ class ValidationState_t {
...
+  // of |width| bits.
+  bool ContainsSizedIntOrFloatType(uint32_t id, SpvOp type,
+                                   uint32_t width) const;
+  // Returns true if |id| contains a 8- or 16-bit int or 16-bit float that is
```
dneto0: Should say |id| is a type id.

source/val/validate_memory.cpp
alan-baker: Right, I think it should `underlying_type` instead. That starts as a copy of `value_type`, which is the first pointed to type (i.e. the variable's allocated type). To find the Block/BufferBlock decorated struct we have to resolve the potential array-ness of the variable. I will add a few tests for this.

PR #2706, opened 2019-06-27
Add SPIRV-Opt pass to isolate OpFunctionCall instructions

chrisforbes:
When a callee is large and contains many function calls, the inliner spends
a significant amount of time repeatedly walking the instructions looking
for inlining opportunities. This accounts for approx 10% of my profile
in dEQP-VK.ubo.random.all_shared_buffer.48 [which has 2 shaders, each of
which contains about 2K inlining opportunities] run on Swiftshader.

By splitting the callee block on either side of the call instruction, we
ensure that there is no large body of non-call instructions to repeatedly consider.
The resulting branches can be cleaned up by a later block merging pass.

Change-Id: Ib8c1420ac452aa591496f6bfedf8cc2ba7bc7416

### DIFF
--- a/source/opt/isolate_function_calls_pass.h
+++ b/source/opt/isolate_function_calls_pass.h
@@ -1,0 +1,55 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SOURCE_OPT_ISOLATE_FUNCTION_CALLS_PASS_H_
+#define SOURCE_OPT_ISOLATE_FUNCTION_CALLS_PASS_H_
+
+#include <algorithm>
+#include <map>
+#include <queue>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+
+#include "source/opt/basic_block.h"
+#include "source/opt/def_use_manager.h"
+#include "source/opt/ir_context.h"
+#include "source/opt/module.h"
+#include "source/opt/pass.h"
+
+namespace spvtools {
+namespace opt {
+
+// See optimizer.hpp for documentation.
+class IsolateFunctionCallsPass : public Pass {
+ public:
+  IsolateFunctionCallsPass();
+  const char* name() const override { return "isolate-function-calls"; }
+  Status Process() override;
+
+  IRContext::Analysis GetPreservedAnalyses() override {
+    return IRContext::kAnalysisInstrToBlockMapping |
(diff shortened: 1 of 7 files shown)


chrisforbes:
Ah, yes. I had checked that phis would be OK, but you're right -- the back edge is a problem.

source/opt/isolate_function_calls_pass.h
```diff
@@ -0,0 +1,55 @@
...
+
+// See optimizer.hpp for documentation.
+class IsolateFunctionCallsPass : public Pass {
+ public:
+  IsolateFunctionCallsPass();
+  const char* name() const override { return "isolate-function-calls"; }
+  Status Process() override;
+
+  IRContext::Analysis GetPreservedAnalyses() override {
+    return IRContext::kAnalysisInstrToBlockMapping |
```
s-perron: I don't think the instr-to-block mapping is kept up-to-date.  I might be a good idea to keep the def-use manager up-to-date as well.  That is expensive to rebuild.  The gains you make in shaders with small basic block might be lost if the def-use chain need to be rebuilt.

alan-baker:
> I have nothing to add beyond what Alan already mentioned. Add tests. For the OpLoopMerge, you can just move it from the post call block to the pre call block. I do not see a reason to not handle that case.
> 
> Also, do you want to add it to the standard set of optimizations [here](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/optimizer.cpp#L114), [here](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/optimizer.cpp#L158), and [here](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/optimizer.cpp#L196)? I can run some tests to see if it helps on some real games, which I think it will.

My concern is the following:
```
%header = OpLabel
OpBranchConditional %c %body1 %body2
%body1 = OpLabel
OpBranch %continue
%body2 = OpLabel
OpBranch %continue
%continue = OpLabel
OpBranchConditional %c1 %header %merge
%merge = OpLabel
...

Splitting this header would require introducing a selection merge. I don't see anything that forbids this loop, but I also doubt drivers expect to see it.
```

dneto0:
I haven't looked at the code.  But this might also break this rule for OpSampledImage:

> All OpSampledImage instructions must be in the same block in which their Result <id> are consumed.

The inliner has special logic to ensure this rule is followed when we do actually inline a function call that is between the OpSampledImage instruction and the use of its result Id.   Pre-emptively splitting blocks at function calls will have to do the same.

dneto0:
There's a similar rule for when you get a scalar image or sampler from an aggregate:

> Instructions for extracting a scalar image or scalar sampler out of a composite must only use dynamically-uniform indexes. They must be in the same block in which their Result id are consumed.

chrisforbes:
Just circling back to this now -- we found other ways to get the performance we need for now.

Issues with validity of the output may make this less useful for other clients of the optimizer, but for Swiftshader's case we don't actually need some of those properties. What is SPIRV-Tools' position on passes which don't preserve all the validity requirements, but are OK for their intended purpose?

s-perron:
We do not allow transformations that do not preserve validity.

s-perron:
@chrisforbes Has this been abandoned?  If so, we should close the PR.  We can open a new PR if we determine it is needed later.

chrisforbes:
Yes, we found the performance we needed a different way.

PR #2707, opened 2019-07-01
Add validation for SPV_EXT_demote_to_helper_invocation

jeffbolznv:
This depends on https://github.com/KhronosGroup/SPIRV-Headers/pull/121 and won't pass CI until that gets merged and I can update DEPS.


jeffbolznv:
I've updated DEPS to pick up the new spirv-headers. The glslang change got merged already, and picking up spirv-headers makes it not fail validation in one of its self-tests, which causes CI to fail. How am I supposed to resolve this? I don't think I can change either repo without breaking the other one. Do I need to disable the test in glslang?

jeffbolznv:
Oh wait, glslang has its own version of spirv-headers, I think I can just fix it in glslang first.

jeffbolznv:
Removed WIP, I think this is ready.

jeffbolznv:
Added the check for bool.

alan-baker:
Please rebase and fix the formatting.

PR #2708, opened 2019-07-02
Shrinker for spirv-fuzz

afd:
Adds to spirv-fuzz the option to shrink a sequence of transformations that lead to an interesting binary to be generated, to find a smaller sub-sequence of transformations that still lead to an interesting (but hopefully simpler) binary being generated.  The notion of what counts as "interesting" comes from a user-provided script, the "interestingness function", similar to the way the spirv-reduce tool works.  The shrinking process will give up after a maximum number of steps, which can be configured on the command line.

Tests for the combination of fuzzing and shrinking are included, using a variety of interestingness functions.

### DIFF
--- a/source/fuzz/shrinker.cpp
+++ b/source/fuzz/shrinker.cpp
@@ -14,6 +14,8 @@
 
 #include "source/fuzz/shrinker.h"
 
+#include <sstream>
+
 #include "source/fuzz/pseudo_random_generator.h"
 #include "source/fuzz/replayer.h"
 #include "source/spirv_fuzzer_options.h"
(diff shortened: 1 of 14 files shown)


source/fuzz/shrinker.cpp
```diff
@@ -0,0 +1,223 @@
...
+                // remove transformations finer.  If we managed to remove at
+                // least one chunk of transformations at a particular
+                // granularity, we set this flag so that we do not yet decrease
+                // granularity.
+
+    // We go through the transformations in reverse, in chunks of size
+    // |granularity|. This is encoded here via an index, |index|, recording how
+    // many chunks of size |granularity| have been tried. The loop exits early
+    // if we reach the shrinking step limit.
+    for (uint32_t index = 0;
```
paulthomson: Could call it `chunk_index` for clarity.

source/fuzz/shrinker.cpp
```diff
@@ -0,0 +1,223 @@
...
+                            // high and decreases during the shrinking process.
+
+  // Keep shrinking until we:
+  // - reach the step limit,
+  // - run out of transformations to remove, or
+  // - cannot make the granularity of shrinking any finer.
+  while (attempt < impl_->step_limit &&
+         !current_best_transformations.transformation().empty() &&
+         granularity > 0) {
+    bool progress_this_round =
```
paulthomson: In other versions of the algorithms: we had a check at the start of the loop that reduces `granularity` if it is more than the size of the number of transformations/opportunities. Here, we could do:
if granularity >= len(transformations)): granularity = len(transformations) / 2

source/fuzz/shrinker.cpp
```diff
@@ -0,0 +1,223 @@
...
+    impl_->consumer(SPV_MSG_INFO, nullptr, {},
+                    "Initial binary is not interesting; stopping.");
+    return ShrinkerResultStatus::kInitialBinaryNotInteresting;
+  }
+
+  uint32_t attempt = 0;  // Keeps track of the number of shrink attempts that
+                         // have been tried, whether successful or not.
+
+  uint32_t granularity =
+      std::max(1u, NumRemainingTransformations(current_best_transformations) /
```
paulthomson: Could call it `chunk_size` for clarity.

source/fuzz/shrinker.cpp
paulthomson: Wrong place

source/fuzz/shrinker.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    // We go through the transformations in reverse, in chunks of size
+    // |granularity|. This is encoded here via an index, |index|, recording how
+    // many chunks of size |granularity| have been tried. The loop exits early
+    // if we reach the shrinking step limit.
+    for (uint32_t index = 0;
+         attempt < impl_->step_limit &&
+         index * granularity <
+             NumRemainingTransformations(current_best_transformations);) {
+      // Remove a chunk of transformations according to the current index and
```
paulthomson: In other versions of the reducer algorithm: we had a check at the start of the loop that reduces `granularity` if it is more than the number of transformations/opportunities. Here, we could do:
if granularity >= len(transformations): granularity = max(1, len(transformations) / 2)

paulthomson: We also usually reset the index in this case.

source/fuzz/shrinker.cpp
afd: Done.

source/fuzz/shrinker.cpp
afd: Done.

source/fuzz/shrinker.cpp
afd: I have reworked the shrinker algorithm quite a bit in this revision.  I believe it is somewhat simpler now, and it's now more evident that if we remove chunks in reverse order, removing a later chunk cannot affect whether an earlier chunk still applies, so long as the points at which chunks are deemed to start is not affected.

PR #2717, opened 2019-07-08
Remove extra semis

alan-baker:
Remove extra semi-colons found through (-Werror -Wextra-semi which is a Clang-only option). Also update re2 to newer version which avoids extra semi-colons.


zoddicus:
Can you throw, // clang-format off ... // clang-format on, around the calls to DeclareCastMethod, so the formatter doesn't misbehave on them?

PR #2718, opened 2019-07-08
Invalidate StructuredCFGAnalysis in KillInst

### LINKED ISSUE #2667: opt: segfault in StructuredCFGAnalysis::SwitchMergeBlock (GraphicsFuzz reference: spirv-tools-segfault-june-12a)
GLSL:

```glsl
#version 310 es

precision highp float;

void main()
{
 switch(0)
  {
   case 0:
   if(false)
    {
    }
   else
    {
    }
  }
}
```
SPIR-V:

```ini
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 7
; Bound: 15
; Schema: 0
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpSource ESSL 310
               OpName %4 "main"
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeInt 32 1
          %7 = OpConstant %6 0
         %10 = OpTypeBool
         %11 = OpConstantFalse %10
          %4 = OpFunction %2 None %3
          %5 = OpLabel
               OpSelectionMerge %9 None
               OpSwitch %7 %9 0 %8
          %8 = OpLabel
               OpSelectionMerge %13 None
               OpBranchConditional %11 %12 %13
         %12 = OpLabel
               OpBranch %13
         %13 = OpLabel
               OpBranch %9
          %9 = OpLabel
               OpReturn
               OpFunctionEnd

```

See zip below. To reproduce:

```
# Optional.
glslangValidator -V shader.frag -o shader.spv

catchsegv spirv-opt -O shader.spv -o shader_opt.spv
```

[reproduce.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/3281590/reproduce.zip)

paulthomson:
Fixes issue in dead_branch_elim_pass.cpp where a stale
StructuredCFGAnalysis leads to dereferencing a null pointer.

Fixes #2667

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -93,8 +93,6 @@ BasicBlock* DeadBranchElimPass::GetParentBlock(uint32_t id) {
 
 bool DeadBranchElimPass::MarkLiveBlocks(
     Function* func, std::unordered_set<BasicBlock*>* live_blocks) {
-  StructuredCFGAnalysis* cfgAnalysis = context()->GetStructuredCFGAnalysis();
-
   std::unordered_set<BasicBlock*> blocks_with_backedge;
   std::vector<BasicBlock*> stack;
   stack.push_back(&*func->begin());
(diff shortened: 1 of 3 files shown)


paulthomson:
This might not be the best approach; in this case, I am happy to close this.

s-perron:
The design we went for was that analyses would be invalidated after a pass has been finished.  It is up to the pass to deal with a potentially invalid analysis when that pass has caused it to go stale.  Very often the most valid analysis is still useful, and does not need to be rebuilt.  That is why I do not want `KillInst` to invalidate analyses.

In `KillInst`, we will update analyses to keep them valid, when it is not too much work.

I have implemented a different solution for this issue.

paulthomson:
> The design we went for was that analyses would be invalidated after a pass has been finished. It is up to the pass to deal with a potentially invalid analysis when that pass has caused it to go stale. 

Ah yes, I wondered if this was the case. Sounds good to me.

PR #2734, opened 2019-07-11
Fix bug in merge return

### LINKED ISSUE #2702: opt: invalid SPIR-V produced with --eliminate-dead-branches --merge-return. In OpPhi instruction, ID definition does not dominate its parent. (GraphicsFuzz ref: spirv-tools-use-not-dominated-from-swiftshader)
To reproduce:

```sh
spirv-opt --eliminate-dead-branches --merge-return shader_reduced_final.comp.spv -o temp.comp.spv
spirv-val temp.comp.spv
```

[spirv-tools-use-not-dominated-from-swiftshader.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/3330741/spirv-tools-use-not-dominated-from-swiftshader.zip)

Found by GraphicsFuzz.

s-perron:
The merge return pass seems to assume that the only new edges in the cfg
are from return block to merge blocks.  However, it is possible that a
merge block branches to a merge block when it did not before.

This change add a new variable to track all of the new edges.  It also
renames some other variables and cleans us the code to make it a bit
easier to read.

Fixes #2702.

### DIFF
--- a/source/opt/merge_return_pass.cpp
+++ b/source/opt/merge_return_pass.cpp
@@ -224,6 +224,7 @@ void MergeReturnPass::BranchToBlock(BasicBlock* block, uint32_t target) {
   return_inst->SetOpcode(SpvOpBranch);
   return_inst->ReplaceOperands({{SPV_OPERAND_TYPE_ID, {target}}});
   context()->get_def_use_mgr()->AnalyzeInstDefUse(return_inst);
+  new_edges_[target_block].insert(block->id());
   cfg()->AddEdge(block->id(), target);
 }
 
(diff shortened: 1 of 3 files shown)


source/opt/merge_return_pass.cpp
```diff
@@ -659,36 +654,49 @@ void MergeReturnPass::MergeReturnBlocks(
...
-      AddNewPhiNodes(bb, new_merge_nodes_[bb], dominator->id());
-    }
+    AddNewPhiNodes(bb);
   }
 }
 
-void MergeReturnPass::AddNewPhiNodes(BasicBlock* bb, BasicBlock* pred,
-                                     uint32_t header_id) {
+void MergeReturnPass::AddNewPhiNodes(BasicBlock* bb) {
+  // New phi nodes are needed for id whose definition use to dominate |bb|, but
```
dnovillo: s/use/used/
s/id/any id/

source/opt/merge_return_pass.cpp
```diff
@@ -659,36 +654,49 @@ void MergeReturnPass::MergeReturnBlocks(
...
+  // no longer dominates |bb|.  These are found by walking the dominator tree
+  // starting at the original immediate dominator of |bb| and ending at its
+  // current dominator.
+
+  // Because are walking the updated dominator tree it is important that the new
```
dnovillo: s/are/we are/

source/opt/merge_return_pass.cpp
```diff
@@ -659,36 +654,49 @@ void MergeReturnPass::MergeReturnBlocks(
...
+  // phi nodes for the original dominators of |bb| must have already been added.
```
dnovillo: s/must have/have/

PR #2739, opened 2019-07-12
Remove unneeded future imports

### LINKED ISSUE #2738: Remove Python 2isms
There are a bunch of 'from __future__ import print_function' occurrences in the repo which are no longer needed, since we move to requiring Python3

zoddicus:
Fixes #2738


zoddicus:
Looks like the the NDK bot is building using Python2 instead of Python3. I am not sure if that is because it doesn't have Python3, or if it is PATH order or some other issue.

alan-baker:
> Looks like the the NDK bot is building using Python2 instead of Python3. I am not sure if that is because it doesn't have Python3, or if it is PATH order or some other issue.

The ndk build is the same bot type as the linux builds. You should just need to update the path in kokoro/ndk-build/build.sh to use python3.

PR #2742, opened 2019-07-16
Add descriptor array scalar replacement

### LINKED ISSUE #2740: Scalar replacement on descriptor arrays
Consider the following hlsl code,

```
#define COUNT 5

Texture2D    MyTextures[COUNT] : register(t0);
Texture2D    NextTexture;

SamplerState MySampler;

float4 main(float2 TexCoord : TexCoord) : SV_Target0
{
  float4 result =
    MyTextures[0].Sample(MySampler, TexCoord) +
    MyTextures[1].Sample(MySampler, TexCoord) +
    MyTextures[2].Sample(MySampler, TexCoord) +
    MyTextures[3].Sample(MySampler, TexCoord) +
    MyTextures[4].Sample(MySampler, TexCoord) +
    NextTexture.Sample(MySampler, TexCoord);
  return result;
}
```

If you look at the output from dxc when generating [dxil](http://shader-playground.timjones.io/0dd78246a5ceed3c3b1ecc85f006bd68), you will notice that `MyTextures` takes up 5 binding t0 to t4. Then `NextTexture` is placed in t5.  The [dxbc](http://shader-playground.timjones.io/f9fbbfa4bf0de7d320bf655e4dd1dcf3) from FXC is similar.

For the [spirv](http://shader-playground.timjones.io/b1db8aa140d124148025272151b71fc5), `MyTextures` is a single array at binding 0, and `NextTexture` is at binding 1.

This can make it conceptually difficult for code that wants to work with Vulkan and DirectX.  It becomes hard to keep track of where everything is bound.

To help with this DXC will be adding an option to have texture map the same way.  However, the front-end cannot convert the array of textures into separate variables until after legalization because it may require optimizations to determine fold the index in the use to a literal.

To this end, we will add a pass that will do scalar replacement on descriptor arrays.  It will be an error it the pass is unable to the scalar replacement.

s-perron:
Creates a pass that will replace a descriptor array with individual variables.  See #2740 for details.

Fixes #2740.

### DIFF
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -788,8 +788,10 @@ Optimizer::PassToken CreateGraphicsRobustAccessPass();
 // This pass replaces every array variable |desc| that has a DescriptorSet and
 // Binding decorations with a new variable for each element of the array.
 // Suppose |desc| was bound at binding |b|.  Then the variable corresponding to
-// |desc[i]| will have binding |b+i|.  The descriptor set will be the same.  All
-// accesses to |desc| must be in OpAccessChain instructions with a literal index
+// |desc[i]| will have binding |b+i|.  The descriptor set will be the same.  It
+// is assumed that no other variable already has a binding that will used by one
+// of the new variables.  If not, the pass will generate invalid Spir-V.  All
+// accesses to |desc| must be OpAccessChain instructions with a literal index
 // for the first index.
 Optimizer::PassToken CreateDescriptorScalarReplacementPass();
 
--- a/include/spirv-tools/optimizer.hpp
+++ b/include/spirv-tools/optimizer.hpp
@@ -784,6 +784,15 @@ Optimizer::PassToken CreateSplitInvalidUnreachablePass();
 // wide.
 Optimizer::PassToken CreateGraphicsRobustAccessPass();
 
+// Create descriptor scalar replacement pass.
+// This pass replaces every array variable |desc| that has a DescriptorSet and
+// Binding decorations with a new variable for each element of the array.
+// Suppose |desc| was bound at binding |b|.  Then the variable corresponding to
+// |desc[i]| will have binding |b+i|.  The descriptor set will be the same.  All
+// accesses to |desc| must be in OpAccessChain instructions with a literal index
+// for the first index.
+Optimizer::PassToken CreateDescriptorScalarReplacementPass();
+
 }  // namespace spvtools
 
 #endif  // INCLUDE_SPIRV_TOOLS_OPTIMIZER_HPP_
--- a/source/opt/desc_sroa.cpp
+++ b/source/opt/desc_sroa.cpp
@@ -101,7 +101,7 @@ bool DescriptorScalarReplacement::ReplaceCandidate(Instruction* var) {
             work_list.push_back(use);
             return true;
           default:
-            EmmitErrorMessage(
+            context()->EmitErrorMessage(
                 "Variable cannot be replaced: invalid instruction", use);
             return false;
         }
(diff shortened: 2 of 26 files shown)


s-perron:
@ehsannas Let me know if the will work with DXC.  We should not merge until we know this will work.

s-perron:
I still need to add tests.

source/opt/desc_sroa.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+  uint32_t ptr_element_type_id = context()->get_type_mgr()->FindPointerToType(
+      element_type_id, storage_class);
+
+  // Create the variable.
+  uint32_t id = TakeNextId();
+  std::unique_ptr<Instruction> variable(
+      new Instruction(context(), SpvOpVariable, ptr_element_type_id, id,
+                      std::initializer_list<Operand>{
+                          {SPV_OPERAND_TYPE_STORAGE_CLASS, {storage_class}}}));
```
ehsannas: I got "requires narrowing conversion" compilation warning (and/or error) for this. Had to use `static_cast<uint32_t>(storage_class)` instead to get it working.

source/opt/desc_sroa.cpp
```diff
@@ -0,0 +1,290 @@
...
+  for (uint32_t i = 4; i < use->NumOperands(); i++) {
+    new_operands.emplace_back(use->GetOperand(i));
+  }
+
+  use->ReplaceOperands(new_operands);
+  context()->UpdateDefUse(use);
+  return true;
+}
+
+void DescriptorScalarReplacement::EmmitErrorMessage(std::string message,
```
dnovillo: s/Emmit/Emit/

dnovillo: I wonder if this function should be floated up to class `Instruction`.  Seems very useful.

source/opt/desc_sroa.cpp
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  if (replacement_vars->second[idx] == 0) {
+    replacement_vars->second[idx] = CreateReplacementVariable(var, idx);
+  }
+
+  return replacement_vars->second[idx];
+}
+
+uint32_t DescriptorScalarReplacement::CreateReplacementVariable(
```
dnovillo: Can we reuse similar functionality from SROA itself?

include/spirv-tools/optimizer.hpp
```diff
@@ -784,6 +784,15 @@ Optimizer::PassToken CreateSplitInvalidUnreachablePass();
 // wide.
 Optimizer::PassToken CreateGraphicsRobustAccessPass();
 
+// Create descriptor scalar replacement pass.
+// This pass replaces every array variable |desc| that has a DescriptorSet and
+// Binding decorations with a new variable for each element of the array.
+// Suppose |desc| was bound at binding |b|.  Then the variable corresponding to
```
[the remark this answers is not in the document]
s-perron: Okay.  I've added 

> It is assumed that no other variable already has a the bindings that will used by the new variables.  If not, the pass will generate invalid Spir-V.

source/opt/desc_sroa.cpp
s-perron: I've moved it to the context.  I could see it being extended to have a "nullptr" for the instruction when the error is not related to a specific instruction.

source/opt/desc_sroa.cpp
s-perron: There might be a small opportunity for reuse, but I don't know if it will be worth it.  SROA uses some extra data when creating variable.  It tries to determine which elements get used, and only the variables that are used get created.  This analysis is done to improve compile time, but does not apply to this pass.

The decorations need to be handled differently because different decorations apply to the global variables vs the function scope variables.

That just leaves some of the code from lines 245 to 267.  If you hide the  code that gets the type into more general functions, then that code will need to be repeated in SROA, so we won't end up reducing the size.
That leaves only 261-266.  Those line are mostly the parameters that need to be passed in.  Making a function out of that code will not reduce the size of the code.  Line 267 is different form sroa because the OpVariable instruction needs to be placed in a different location.

dnovillo: OK, thanks.  It doesn't seem worth it.  I agree.

Issue #2595, opened 2019-05-16
[OpenCL validation] Atomic instructions


kpet:
From the environment spec:
```
For all Atomic Instructions:
- Only 32-bit integer types are supported for the Result Type and/or type of Value.
- The Pointer operand must be a pointer to the Function, Workgroup, or CrossWorkGroup
  Storage Classes. Note that an Atomic Instruction on a pointer to the Function Storage Class
  is valid, but does not have defined behavior.
```

PR #2750, opened 2019-07-18
Validate storage class OpenCL environment rules for atomics

kpet:
This change refactors all storage class validation for atomics
to reflect the similar refactoring in the specification.

It is currently not possible to write a test for the check
rejecting Generic in an OpenCL 1.2 environment as the required
GenericPointer capability isn't allowed there. I've decided
to keep the check nonetheless to guard against the capability
becoming available without the rules for atomics being updated.

The ID changes in existing tests aren't ideal but introducing
names drags in a substantial refactoring of this file.

Contributes to #2595.

Signed-off-by: Kevin Petit <kevin.petit@arm.com>

### DIFF
--- a/source/val/validate_atomics.cpp
+++ b/source/val/validate_atomics.cpp
@@ -25,10 +25,9 @@
 #include "source/val/validate_scopes.h"
 #include "source/val/validation_state.h"
 
-namespace spvtools {
-namespace val {
+namespace {
 
-static bool IsStorageClassAllowedByUniversalRules(uint32_t storage_class) {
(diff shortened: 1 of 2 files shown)


source/val/validate_atomics.cpp
```diff
@@ -28,6 +28,24 @@
 namespace spvtools {
 namespace val {
 
+static bool IsStorageClassAllowedByUniversalRules(uint32_t storage_class) {
```
alan-baker: Please drop the static and put this function in an anonymous namespace to match the style of the rest of the validator.

kpet: Donw

PR #2751, opened 2019-07-18
Document opt::Instruction::InsertBefore methods

dneto0:


source/opt/instruction.cpp
```diff
@@ -372,6 +372,11 @@ uint32_t Instruction::GetTypeComponent(uint32_t element) const {
   return subtype;
 }
 
+Instruction* Instruction::InsertBefore(std::unique_ptr<Instruction>&& i) {
```
dneto0: I moved this up to match the declaration order in the header.
(That's a subtle rule/guidance in the coding style guide.)

source/opt/instruction.h
```diff
@@ -398,8 +398,13 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   inline bool operator!=(const Instruction&) const;
   inline bool operator<(const Instruction&) const;
 
-  Instruction* InsertBefore(std::vector<std::unique_ptr<Instruction>>&& list);
+  // Takes ownership of the instruction owned by |i| and inserts it immediately
```
dneto0: I found it conceptually easier to describe the one-instruction case first.

PR #2761, opened 2019-07-24
Process OpDecorateId in ADCE

s-perron:
When there is an OpDecorateId instruction that is live,
the ids that is references must be kept live.  This change
adds them to the worklist.

I've also updated a validator check to allow OpDecorateId
to be able to apply to decoration groups.

Fixes #1759.


source/opt/aggressive_dead_code_elim_pass.cpp
```diff
@@ -736,6 +761,13 @@ bool AggressiveDCEPass::ProcessGlobalValues() {
               context()->KillInst(annotation);
               modified = true;
             }
+          } else {
+            /*
```
alan-baker: Dead code?

PR #2766, opened 2019-07-26
Add 'copy object' transformation

afd:
This transformation can introduce an instruction that uses
OpCopyObject to make a copy of some other result id.  This change
introduces the transformation, but does not yet introduce a fuzzer
pass to actually apply it.

### DIFF
--- a/source/fuzz/fuzzer_util.cpp
+++ b/source/fuzz/fuzzer_util.cpp
@@ -202,6 +202,7 @@ opt::BasicBlock::iterator GetIteratorForBaseInstructionAndOffset(
     }
   }
   assert(false && "The base instruction was not found.");
+  return nullptr;
 }
 
 }  // namespace fuzzerutil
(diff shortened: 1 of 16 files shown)


source/fuzz/fuzzer_util.cpp
```diff
@@ -170,6 +170,40 @@ bool BlockIsInLoopContinueConstruct(opt::IRContext* context, uint32_t block_id,
...
+    if (base_inst == &*inst_it) {
+      // We have found the base instruction; we now apply the offset.
+      for (uint32_t i = 0; i < offset && inst_it != block->end(); i++) {
+        ++inst_it;
+      }
+      // This is either the desired instruction, or the end of the block.
+      return inst_it;
+    }
+  }
+  assert(false && "The base instruction was not found.");
```
dneto0: Need to return something to satisfy some compilers.
Suggest returning nullptr.

PR #2767, opened 2019-07-26
Fix #2609 - Handle out-of-bounds scalar replacements.

### LINKED ISSUE #2609: Segmentation fault when applying spirv-opt to a shader that performs and out-of-bounds access
I expect this is closely related to #2608 but am filing it in addition as the symptoms are different.

[This archive](https://github.com/KhronosGroup/SPIRV-Tools/files/3201632/files.zip) contains the files required to reproduce the problem.

To reproduce, do:

spirv-opt -O small.spv -o temp.spv

I get a segmentation fault in `spvtools::opt::Instruction::result_id()`.

The .frag file exhibits and out-of-bounds access: a is an int[1], and a[i] attempts to index v at 1.

As in #2608, I am unsure what the semantics of out-of-bounds accesses are in SPIR-V, but I note that an OpAccessChain, rather than an OpInBoundsAccessChain, is used; I would assume that the existence of OpInBoundsAccessChain means that OpAccessChain is supposed to account for the out-of-bounds case.

Found via [GraphicsFuzz](https://github.com/google/graphicsfuzz).

dnovillo:
When SROA tries to do a replacement for an OpAccessChain that is exactly
one element out of bounds, the code was trying to access its internal
array of replacements and segfaulting.

This protects the code from doing this, and it additionally fixes the
way SROA works by not returning failure when it refuses to do a
replacement.  Instead of failing the optimization pass, SROA will now
simply refuse to do the replacement and keep going.

### DIFF
--- a/source/opt/scalar_replacement_pass.cpp
+++ b/source/opt/scalar_replacement_pass.cpp
@@ -60,20 +60,21 @@ Pass::Status ScalarReplacementPass::ProcessFunction(Function* function) {
     Instruction* varInst = worklist.front();
     worklist.pop();
 
-    if (!ReplaceVariable(varInst, &worklist))
-      return Status::Failure;
-    else
-      status = Status::SuccessWithChange;
+    Status var_status = ReplaceVariable(varInst, &worklist);
+    if (var_status == Status::Failure)
+      return var_status;
+    else if (var_status == Status::SuccessWithChange)
+      status = var_status;
   }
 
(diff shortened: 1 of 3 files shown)


dnovillo:
@alan-baker thanks for spotting the bad test I had pasted.  I've done the constant propagation manually and now the test does the proper indexing that I was getting with `-O`.  PTAL.

Issue #2724, opened 2019-07-09
Ensure that restrictions on Barrier instructions in WebGPU are enforced


zoddicus:
Wording of these restrictions was clarified in gpuweb/spirv-execution-env@131313e. Need to confirm that the implementation matches the new text.

> *  **OpControlBarrier**\* restrictions:
>    *   The _Execution_ scope operand must be **Workgroup**
>    *   The _Memory_ scope operand must be **Workgroup**
>    *   The _Semantics_ operand:
>       * Memory Semantics Order bits: Must set the **AcquireRelease** bit, and no other bits.
>       * Memory Semantics Storage Class bits: Must set the **WorkgroupMemory** bit, and no
> other bits.
>       * Must not set any Memory Semantics Propagation bits.
>
> Note: To ensure propagation of **Workgroup** writes to readers,
> individual accesses of **Workgroup** memory should use availability
> and visibility semantics:
> **MakePointerAvailableKHR** with **Workgroup** scope on writes,
> **MakePointerVisibleKHR** with **Workgroup** scope on reads,
> and **NonPrivatePointerKHR** on both.
>
> *  **OpMemoryBarrier**\* restrictions:
>    *   The _Memory_  scope operand must be **Workgroup**
>    *   The _Semantics_ operand:
>       * Must not set any Memory Semantics Order bits.
>       * Memory Semantics Storage Class bits: Must set the **ImageMemory** bit, and no other
> bits.
>       * Must not set any Memory Semantics Propagation bits.
> 
> Note: **OpMemoryBarrier** is only used to order reads and writes by the same
> invocation to the same locations in image memory.

zoddicus:
This is actually resolved by #2775 . I thought there was more work needed, but that is in a different issue.

PR #2769, opened 2019-07-26
Update OpControlBarriers rules for WebGPU

zoddicus:
Part of #2724

### DIFF
--- a/test/val/val_barriers_test.cpp
+++ b/test/val/val_barriers_test.cpp
@@ -297,6 +297,18 @@ OpControlBarrier %workgroup %workgroup %acquire_release_uniform_workgroup
                 "for Memory Semantics of OpControlBarrier."));
 }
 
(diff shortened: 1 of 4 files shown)


test/val/val_barriers_test.cpp
```diff
@@ -251,7 +253,7 @@ OpControlBarrier %workgroup %workgroup %acquire_release_uniform_workgroup
 
 TEST_F(ValidateBarriers, OpControlBarrierWebGPUAcquireReleaseSuccess) {
   const std::string body = R"(
-OpControlBarrier %workgroup %workgroup %acquire_release_uniform_workgroup
+OpControlBarrier %workgroup %workgroup %acquire_release_workgroup
```
alan-baker: I think you should duplicate this test as a failure for including uniform

zoddicus: Apparently I moved the Uniform test lower down, and removed a different test :-S

I have added back in the missing test, and the test you asked for should be at around 286

PR #2770, opened 2019-07-26
Add opt test fixture method SinglePassRunAndFail

dneto0:
Checks for failure status code and matches against
the expected error message.

### DIFF
--- a/test/opt/pass_fixture.h
+++ b/test/opt/pass_fixture.h
@@ -188,6 +188,32 @@ class PassTest : public TestT {
         << disassembly;
   }
 
+  // Runs a single pass of class |PassT| on the binary assembled from the
+  // |original| assembly. Check for failure and expect an Effcee matcher
+  // to pass when run on the diagnostic messages. This does *not* involve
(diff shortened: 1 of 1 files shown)


dneto0:
I'm going to use this in tests for a pass I'm currently developing.

dneto0:
> Would it make sense to re-factor the existing SinglePassRunAndCheck() to add a parameter for the expected result from the pass? To avoid duplicating code.

I think you mean SinglePassRunAndMatch.  That's the one that does pattern-matching with Effcee.   That one has a bool do_validation parameter which gets in the way of passing the expected result status in this known-failure usage.

dnovillo:
> > Would it make sense to re-factor the existing SinglePassRunAndCheck() to add a parameter for the expected result from the pass? To avoid duplicating code.
> 
> I think you mean SinglePassRunAndMatch. That's the one that does pattern-matching with Effcee. That one has a bool do_validation parameter which gets in the way of passing the expected result status in this known-failure usage.

Ah, yes.  OK, I guess we want this one then.  Thanks.

PR #2772, opened 2019-07-29
Don't move debug or decorations when folding

### LINKED ISSUE #2764: spirv-opt moves RelaxedPrecision decoration from OpIAdd to FindSMsb
[spirvoptissue.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/3432486/spirvoptissue.zip)

The attached zip contains this little fragment shader:

```
#version 310 es

precision highp float;

layout(location = 0) out vec4 _GLF_color;

void main() {
  highp int a = 16777216;
  _GLF_color = vec4(24 == (0 + findMSB(a)) ? 1.0 : 0.0, 0.0, 0.0, 1.0);
}
```

glslang treats the result of "0 + findMSB(a)" as having relaxed precision, decorating `%18` in example.asm with RelaxedPrecision.

```
               OpDecorate %18 RelaxedPrecision
                       ...
         %18 = OpIAdd %6 %15 %17
```

This is OK, because `findMSB` has lowp on its return type (which makes sense because the most significant bit of an integer cannot be a large number), and 0 does not have a precision, so lowp is the lowest common denominator.

Running spirv-opt -O on the generated SPIR-V gives example.opt.asm, in which the "0 +" has been removed.

The problem is that the RelaxedPrecision decoration has now been changed to refer to `%17`, which is the call to findMSB:

```
               OpDecorate %17 RelaxedPrecision
                       ...
         %17 = OpExtInst %6 %1 FindSMsb %9
```

According to the SPIR-V specification, this leads not just to the result of FindSMsb, but also its argument, `%9`, being treated as having relaxed precision.

On some implementations this leads to `%9` being truncated to a smaller number of bits, changing what is rendered.  (In this test, `%9` contains 16777216, which has 24 as its MSB, but if it is truncated to e.g. 16 bits it has 0 as its MSB.)

Issue found using [GraphicsFuzz](https://github.com/google/graphicsfuzz)

alan-baker:
Fixes #2764

* Don't replace all uses when simplifying instructions, instead only
update non-debug, non-decoration uses
  * added a test

Decorations shouldn't be pulled into generators when simplifying instructions.

### DIFF
--- a/source/opt/simplification_pass.cpp
+++ b/source/opt/simplification_pass.cpp
@@ -71,8 +71,19 @@ bool SimplificationPass::SimplifyFunction(Function* function) {
               }
             });
             if (inst->opcode() == SpvOpCopyObject) {
-              context()->ReplaceAllUsesWith(inst->result_id(),
-                                            inst->GetSingleWordInOperand(0));
+              const auto replacement = inst->GetSingleWordInOperand(0);
+              std::unordered_set<Instruction*> uses_to_update;
+              get_def_use_mgr()->ForEachUse(
+                  inst, [&uses_to_update, replacement](Instruction* user,
+                                                       uint32_t operand) {
+                    const auto opcode = user->opcode();
+                    if (!spvOpcodeIsDebug(opcode) &&
+                        !spvOpcodeIsDecoration(opcode)) {
+                      user->SetOperand(operand, {replacement});
+                      uses_to_update.insert(user);
+                    }
+                  });
+              for (auto user : uses_to_update) context()->AnalyzeUses(user);
               inst_to_kill.insert(inst);
               in_work_list.insert(inst);
             } else if (inst->opcode() == SpvOpNop) {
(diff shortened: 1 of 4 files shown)


source/opt/simplification_pass.cpp
```diff
@@ -71,8 +71,19 @@ bool SimplificationPass::SimplifyFunction(Function* function) {
...
             if (inst->opcode() == SpvOpCopyObject) {
-              context()->ReplaceAllUsesWith(inst->result_id(),
-                                            inst->GetSingleWordInOperand(0));
+              const auto replacement = inst->GetSingleWordInOperand(0);
+              std::unordered_set<Instruction*> uses_to_update;
+              get_def_use_mgr()->ForEachUse(
+                  inst, [&uses_to_update, replacement](Instruction* user,
+                                                       uint32_t operand) {
+                    const auto opcode = user->opcode();
+                    if (!spvOpcodeIsDebug(opcode) &&
```
dnovillo: Suggest factoring into `ReplaceAllCodeUses` (or some better name).  We may need to do more of these.

alan-baker: I refactored this into a RAUW version that takes a predicate. I left the predicate specific to this pass for now. I'm not sure we want to create a whole bunch of variants right away. Let me know what you think.

dnovillo: Yup, this is good.  We can always add more predicates or make them globally available.  Thanks.

PR #2776, opened 2019-07-31
Treat access chain indexes as signed in SROA

### LINKED ISSUE #2768: Scalar replacement should interpret constant indexes as signed counts
This was clarified in the spec for OpAccessChain, indexes are signed counts.

Not sure if other passes also need updated.

alan-baker:
Fixes #2768

* In scalar replacement, interpret access chain indexes as signed counts
* Use Constant::GetSignExtendedValue and Constant::GetZeroExtendedValue
where appropriate
* new tests

Previously, sroa didn't handle 8 or 16 bit index constants.

### DIFF
--- a/source/opt/scalar_replacement_pass.cpp
+++ b/source/opt/scalar_replacement_pass.cpp
@@ -232,8 +232,12 @@ bool ScalarReplacementPass::ReplaceAccessChain(
   // indexes) or a direct use of the replacement variable.
   uint32_t indexId = chain->GetSingleWordInOperand(1u);
   const Instruction* index = get_def_use_mgr()->GetDef(indexId);
-  uint64_t indexValue = GetConstantInteger(index);
-  if (indexValue >= replacements.size()) {
+  int64_t indexValue = context()
+                           ->get_constant_mgr()
+                           ->GetConstantFromInst(index)
+                           ->GetSignExtendedValue();
+  if (indexValue < 0 ||
+      static_cast<size_t>(indexValue) >= replacements.size()) {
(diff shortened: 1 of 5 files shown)


source/opt/scalar_replacement_pass.cpp
```diff
@@ -232,8 +232,12 @@ bool ScalarReplacementPass::ReplaceAccessChain(
   // indexes) or a direct use of the replacement variable.
   uint32_t indexId = chain->GetSingleWordInOperand(1u);
   const Instruction* index = get_def_use_mgr()->GetDef(indexId);
-  uint64_t indexValue = GetConstantInteger(index);
```
dnovillo: Would it make sense to templatize `GetConstantInteger` so it does all this work based on the requested type?  I'm thinking moving `GetConstantInteger` to `constants.cpp` and write it in terms of `GetConstantFromInst` and `GetSignExtendedValue`.

alan-baker: I left GetConstantInteger around to minimize the total diff, but it doesn't serve much purpose anymore. It could be removed. I prefer authors having to be explicit about getting the sign or zero extended versions. This matches up with LLVM.

dnovillo: > I left GetConstantInteger around to minimize the total diff, but it doesn't serve much purpose anymore. It could be removed. 

Yeah, I'd rather remove it.

> I prefer authors having to be explicit about getting the sign or zero extended versions. This matches up with LLVM.

Wouldn't sign/zero extension be implicit with the template type?  But I understand the preference for being explicit.  OK, we can do that later, if needed.

PR #2780, opened 2019-08-01
Remove extra ';' after member function definition.

vonture:
This fixes a clang compiler warning about extra semicolons that blocks ANGLE's integration.


zoddicus:
I have filed #2781 about getting the right flags enabled to prevent this in the future. I will implement it after I get back from the long weekend.

PR #2782, opened 2019-08-02
Instrument: Fix version 2 output record write for tess eval shaders.

greg-lunarg:
Also change command line for bindless instrumentation to use use
output record version 2.


s-perron:
The appveyor failures looks like a connectivity issue.  I'll merge anyway.

PR #2784, opened 2019-08-06
Install CMake Config, raise minimum required version to v3.0

mmha:
* Raise minimum CMake version to 3.0
* Install a Config module to make the tools findable by CMake find_package()
* Use find_package() for the headers
	* Implement a fallback for older versions of the header
* Minor cleanups throughout the build scripts

I'd say it closes #2662 since it does install a Config file, but there's still a lot of cleanup possible.

PR #2785, opened 2019-08-06
Export SPIRV-Tools targets on installation

rumblehhh:
This allows the targets to be used in other cmake projects. See the following for more details:
https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-packages
https://foonathan.net/blog/2016/07/07/cmake-dependency-handling.html


s-perron:
I would like to get https://github.com/KhronosGroup/SPIRV-Tools/pull/2784 merged first.  They both touch the cmake files.

jamesRumbleh:
Sure, there's no rush for mine

rumblehhh:
Thanks for merging this through

PR #2791, opened 2019-08-13
Update remquo validation to match the OpenCL Extended Instruction Set Specification

tremmelg:
The `remquo` builtin was validated as if the last argument was supposed to be to pointer to a float instead of a pointer to a 32-bit integer, which doesn't match the OpenCL Extended Instruction Set Specification (https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.ExtendedInstructionSet.100.html#remquo).

Move `remquo` specific checks to the same case as `frexp` and `lgamma_r` which require a pointer to a 32-bit integer as the last argument and update the tests.


tremmelg:
Tried to fix the formatting error

PR #2795, opened 2019-08-14
Change the way to include header

jaebaek:
Including source/util/string_utils.h always uses
"source/util/string_utils.h" instead of
<source/util/string_utils.h>.


jaebaek:
For the future reference, `#include <source/util/string_utils.h>` works only when we specify `include_directories(${CMAKE_CURRENT_SOURCE_DIR}/)`. I am not sure how easily we can set it in bazel, but it is hard to set the source directory as a include path in some build systems. Using the relative path easily solves this issue.

PR #2797, opened 2019-08-15
Extend reducer to remove relaxed precision decorations

afd:
Adds a reduction pass that removes OpDecorate and OpMemberDecorate
instructions that annotate instructions and members with
RelaxedPrecision.  As well as being useful in its own right, removing
such references allows other passes to remove further instructions.

### DIFF
--- a/test/reduce/remove_relaxed_precision_decoration_test.cpp
+++ b/test/reduce/remove_relaxed_precision_decoration_test.cpp
@@ -1,0 +1,177 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


test/reduce/remove_relaxed_precision_decoration_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+         %22 = OpLoad %10 %12
+         %23 = OpLoad %6 %8
+         %24 = OpLoad %14 %16
+         %25 = OpCompositeConstruct %19 %22 %23 %24
+               OpStore %21 %25
+               OpReturn
+               OpFunctionEnd
+  )";
+
+  CheckEqual(env, expected, context.get());
```
dneto0: This is all good.

Depending on your goals, you might want to test with the [Effcee](url) pattern matcher (inspired by LLVM's FileCheck).   It can make tests easier to read / more focused because you only test against the disassembly you care about.  See optimizer tests using SinglePassRunAndMatch for examples.

afd: Thanks - I will look into that for future tests (and will consider refactoring these if it makes sense to do so).

PR #2799, opened 2019-08-16
spvtools::Optimizer - don't assume original_binary and optimized_binary are aliased

### LINKED ISSUE #2798: spvtools::Optimizer::Run() assumes original_binary and optimized_binary are aliased
Despite the documentation for `spvtools::Optimizer::Run(const uint32_t* original_binary, size_t original_binary_size, std::vector<uint32_t>* optimized_binary) const` saying:

`// It's allowed to alias |original_binary| to the start of |optimized_binary|.`

The actual logic seems to make assumptions that they are aliased:

```C++
  // Optimization logic. Does not assign to optimized_binary.

  bool binary_changed = false;
  if (status == opt::Pass::Status::SuccessWithChange) {
    binary_changed = true;
  } else if (status == opt::Pass::Status::SuccessWithoutChange) {
    // -- optimized_binary has not been assigned to so this if will always be true! --
    if (optimized_binary->size() != original_binary_size ||
        (memcmp(optimized_binary->data(), original_binary,
                original_binary_size) != 0)) {
      binary_changed = true;
      Log(consumer(), SPV_MSG_WARNING, nullptr, {},
          "Binary unexpectedly changed despite optimizer saying there was no "
          "change");
    }
  }

  if (binary_changed) {
    optimized_binary->clear();
    context->module()->ToBinary(optimized_binary, /* skip_nop = */ true);
  }

  return status != opt::Pass::Status::Failure;
}
```

As `optimized_binary` is never actually assigned to before doing the size and data check, we will always get the message printed:
  `"Binary unexpectedly changed despite optimizer saying there was no change"`
And because `binary_changed` is set to true, we get the `optimized_binary` written.

ben-clayton:
If they are not aliased, the function will often print the message:

     "Binary unexpectedly changed despite optimizer saying there was no change"

Which is (usually) totally bogus.

Fixes #2798

### DIFF
--- a/source/opt/optimizer.cpp
+++ b/source/opt/optimizer.cpp
@@ -540,26 +540,24 @@ bool Optimizer::Run(const uint32_t* original_binary,
   impl_->pass_manager.SetTargetEnv(impl_->target_env);
   auto status = impl_->pass_manager.Run(context.get());
 
-  bool binary_changed = false;
-  if (status == opt::Pass::Status::SuccessWithChange) {
-    binary_changed = true;
-  } else if (status == opt::Pass::Status::SuccessWithoutChange) {
+  if (status == opt::Pass::Status::Failure) {
+    return false;
+  }
+
+  optimized_binary->clear();
+  context->module()->ToBinary(optimized_binary, /* skip_nop = */ true);
+
+  if (status == opt::Pass::Status::SuccessWithoutChange) {
     if (optimized_binary->size() != original_binary_size ||
         (memcmp(optimized_binary->data(), original_binary,
                 original_binary_size) != 0)) {
-      binary_changed = true;
       Log(consumer(), SPV_MSG_WARNING, nullptr, {},
           "Binary unexpectedly changed despite optimizer saying there was no "
           "change");
     }
   }
 
-  if (binary_changed) {
-    optimized_binary->clear();
-    context->module()->ToBinary(optimized_binary, /* skip_nop = */ true);
-  }
-
-  return status != opt::Pass::Status::Failure;
+  return true;
 }
 
 Optimizer& Optimizer::SetPrintAll(std::ostream* out) {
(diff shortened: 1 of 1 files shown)


source/opt/optimizer.cpp
```diff
@@ -540,26 +540,24 @@ bool Optimizer::Run(const uint32_t* original_binary,
...
+
+  optimized_binary->clear();
+  context->module()->ToBinary(optimized_binary, /* skip_nop = */ true);
+
+  if (status == opt::Pass::Status::SuccessWithoutChange) {
     if (optimized_binary->size() != original_binary_size ||
         (memcmp(optimized_binary->data(), original_binary,
                 original_binary_size) != 0)) {
-      binary_changed = true;
       Log(consumer(), SPV_MSG_WARNING, nullptr, {},
```
s-perron: This strikes me as odd.  If the optimizer reports that there was no change, but it did change, then that is a bug in the optimizer.  Why would the user need to be warned about that.  They can do nothing about it. Warning and error should be warning or error about something the user did.  Not bugs in the compiler.

I think we should change this to only do the comparison an assert, which is only enabled in a debug build.  Then for a release build, we only need to write to `optimized_binary` if the optimizer reported a change or `optimized_binary != original_binary`.

ben-clayton: Couldn't the user be providing their own optimization passes?

s-perron: They  could, but at that point they are now compiler developers.  The error is in their role as a compiler developer, not as a user.

This warning was added in #2487.  I should have picked up on it in the code review for that PR.

ben-clayton: Okay, migrated to an assert.

ben-clayton:
I do not have permissions to land this change myself.
If you're happy with it, please can do the honours?

PR #2806, opened 2019-08-21
Even more id overflow in sroa

s-perron:
Now we need to handle id overflow when we overflow while replacing uses of the variable.  While looking at this code, I noticed an error in the way we handle access chains that cannot be replaced because of overflow.  Name it will make some change, and then give up by returning SuccessWithoutChange.  But it was changed.

This is fixed up by returning Failure if we notice the error at the time of rewriting the users.  This is for both id overflow or out-of-bounds accesses.

Code is added to "CheckUses" to remove variables that have out-of-bounds accesses from the candidate list, so we don't even try to rewrite its uses.

Fixes https://crbug.com/995032

### DIFF
--- a/source/opt/scalar_replacement_pass.cpp
+++ b/source/opt/scalar_replacement_pass.cpp
@@ -78,36 +78,47 @@ Pass::Status ScalarReplacementPass::ReplaceVariable(
   }
 
   std::vector<Instruction*> dead;
-  if (get_def_use_mgr()->WhileEachUser(
-          inst, [this, &replacements, &dead](Instruction* user) {
-            if (!IsAnnotationInst(user->opcode())) {
-              switch (user->opcode()) {
-                case SpvOpLoad:
-                  ReplaceWholeLoad(user, replacements);
-                  dead.push_back(user);
-                  break;
-                case SpvOpStore:
-                  ReplaceWholeStore(user, replacements);
-                  dead.push_back(user);
-                  break;
-                case SpvOpAccessChain:
-                case SpvOpInBoundsAccessChain:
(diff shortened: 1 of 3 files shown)


source/opt/scalar_replacement_pass.cpp
```diff
@@ -78,36 +78,47 @@ Pass::Status ScalarReplacementPass::ReplaceVariable(
...
+              break;
+            default:
+              assert(false && "Unexpected opcode");
+              break;
+          }
+        }
+        return true;
+      });
+
+  if (replaced_all_uses) {
```
dnovillo: Won't this undo the semantics that got implemented in https://github.com/KhronosGroup/SPIRV-Tools/pull/2767.  In that case, we wanted to silently not do the update but also not fail.

s-perron: That is why I mentioned the changes to [CheckUses](https://github.com/KhronosGroup/SPIRV-Tools/pull/2806/files#diff-a072caf4776a304d1df3bf4e0ba7ebb0R726).  For the out-of-bounds accesses, we will gracefully bail out before we hit this code.  

The design for sroa and a lot of other optimizations, is to first check if we can do the optimization.  If we can, do the changes.  We are now more inline with that design.

s-perron: You can see that the test that was added in #2767 still passes.

dnovillo: Yes.  Given your change and my memory of #2767, I was not trusting my test.  Hence the question ;)

PR #2807, opened 2019-08-21
Handle Id overflow in private-to-local

s-perron:
We need to handle id overflow in the private to local pass.

Fixes https://crbug.com/962295


s-perron:
the smoketest failure is not due to this change.

PR #2809, opened 2019-08-23
Add missing files to BUILD.gn

jonahryandavis:
New files missing from BUILD.gn caused build failures in Chromium and
ANGLE:
remove_relaxed_precision_decoration_opportunity_finder.cpp
remove_relaxed_precision_decoration_opportunity_finder.h


zoddicus:
Also filed #2810 about having the bots catch this type of issue at review time.

PR #2821, opened 2019-08-30
Fix gn check

Kangz:
spriv-opt was missing a dependency on the AMD ballot extension that was
needed because it uses a header in the AMD ext to KHR ext pass.


Kangz:
Can you merge them for me? Thanks!

Issue #2814, opened 2019-08-26
Add passes to replace uses of AMD extensions


s-perron:
Some extensions have become obsolete because they have equivalent functionality in new versions of spir-v.  For portability developers may want to change their spir-v module to remove the extensions.

For this issue we specifically want a pass to replaces uses of three extensions:
  - VK_AMD_shader_ballot
  - VK_AMD_gcn_shader
  - VK_AMD_shader_trinary_minmax

s-perron:
There are still two issues that need to be addressed for the three extension.  We cannot replace `TimeAMD` instructions because the Vulkan extension that enables SPV_KHR_shader_clock has not been released yet.

Also, the extension that allows `OpGroup*` instructions to work on 16-bit integers has not been release, as far as I know.

dnovillo:
I'll take a look at adding `TimeAMD` and `OpGroup*` next.

PR #2822, opened 2019-08-30
Replace SwizzleInvocationsMaskedAMD extended instruction.

s-perron:
Part of #2814

### DIFF
--- a/source/opt/type_manager.h
+++ b/source/opt/type_manager.h
@@ -139,6 +139,61 @@ class TypeManager {
   const Type* GetMemberType(const Type* parent_type,
                             const std::vector<uint32_t>& access_chain);
 
+  Type* GetUIntType() {
+    Integer int_type(32, false);
+    return GetRegisteredType(&int_type);
+  }
+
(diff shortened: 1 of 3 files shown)


source/opt/type_manager.h
```diff
@@ -139,6 +139,61 @@ class TypeManager {
...
+    return GetRegisteredType(&vec_type);
+  }
+
+  uint32_t GetUIntVectorTypeId(uint32_t size) {
+    return GetTypeInstruction(GetUIntVectorType(size));
+  }
+
+  Type* GetSIntVectorType(uint32_t size) {
+    Vector vec_type(GetSIntType(), size);
+    return GetRegisteredType(&vec_type);
```
dnovillo: Do we want a few GetType<T, W>() for all the scalar types T of width W?  Not sure how much this would clean things up in the current code, though.

s-perron:
Anytime I write or review code where we need a specific type, it is always 32-bit integers and floats or vectors of them.

PR #2823, opened 2019-08-30
Replace SwizzleInvocationsAMD extended instruction.

s-perron:
Part of #2814


source/opt/amd_ext_to_khr.cpp
```diff
@@ -79,14 +79,118 @@ FoldingRule ReplaceGroupNonuniformOperationOpCode(SpvOp new_opcode) {
   };
 }
 
-FoldingRule NotImplementedYet() {
-  return [](IRContext*, Instruction*,
+// Returns a folding rule that will replace the WriteInvocationAMD extended
```
dnovillo: s/WriteInvocationAMD/SwizzleInvocationAMD/

PR #2828, opened 2019-08-30
GN: Make SPIRV-Tools target use public_deps.

null77:
Should prevent invalid header usage warnings in dependent targets.

See http://anglebug.com/3876 for context.

@dj2 PTAL


null77:
@dneto0 PTAL if Dan is out today, not sure.

PR #2836, opened 2019-09-04
Fold Min, Max, and Clamp instructions.

### LINKED ISSUE #2830: FClamp/UClamp/SClamp opt: Evaluate as constant where possible
The clamp instructions in the [GLSL extended instruction set](https://www.khronos.org/registry/spir-v/specs/1.0/GLSL.std.450.html) implement the following for a given value of `x`:
```
min(max(x, minVal), maxVal)
```
so if `x <= minVal`, the result of the clamp is `minVal`. And if `x >= maxVal`, the result of the clamp is `maxVal`.

If the optimizer recognizes this, it can potentially simplify large portions of shader code by constant propagation.

For example, the result of
```
         %26 = OpExtInst %float %1 FClamp %float_0 %float_0 %float_1
```
is in fact `%float_0`.

This manifested itself in https://github.com/microsoft/DirectXShaderCompiler/issues/2437.

s-perron:
Fixes #2830.

### DIFF
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -353,6 +353,10 @@ ConstantFoldingRule FoldFPBinaryOp(BinaryScalarFoldingRule scalar_rule) {
     if (!inst->IsFloatingPointFoldingAllowed()) {
       return nullptr;
     }
+    if (inst->opcode() == SpvOpExtInst) {
+      return FoldFPBinaryOp(scalar_rule, inst->type_id(),
+                            {constants[1], constants[2]}, context);
+    }
     return FoldFPBinaryOp(scalar_rule, inst->type_id(), constants, context);
   };
 }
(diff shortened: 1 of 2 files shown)


source/opt/const_folding_rules.cpp
```diff
@@ -893,6 +897,124 @@ ConstantFoldingRule FoldFMix() {
...
+  return [](IRContext* context, Instruction* inst,
+            const std::vector<const analysis::Constant*>& constants)
+             -> const analysis::Constant* {
+    assert(inst->opcode() == SpvOpExtInst &&
+           "Expecting an extended instruction.");
+    assert(inst->GetSingleWordInOperand(0) ==
+               context->get_feature_mgr()->GetExtInstImportId_GLSLstd450() &&
+           "Expecting a GLSLstd450 extended instruction.");
+
+    // Make sure all Clamp operands are constants.
```
alan-baker: This seems more restrictive than necessary.

```
clamp(1.0f, 0.0f, x) -> 1.0f
clamp(x, 2.0f, 1.0f) -> 1.0f
```

If this were a regular folding rule (not just a const rule) you could also fold:

```
clamp(1.0f, 2.0f, x) -> max(2.0f, x)
clamp(x, 1.0f, 2.0f) -> min(x, 1.0f)
```

Do you think it is worth handling these cases too?

s-perron:
```
clamp(1.0f, 0.0f, x) -> 1.0f
```

This is not correct, This can be simplified to `min(1.0f, x)`, but no further.  However, I do not expect this to be very common.

```
clamp(x, 2.0f, 1.0f) -> 1.0f
```

This is undefined behaviour according to the instruction: "Result is undefined if minVal > maxVal." I could fold it, but it is not suppose to happen.

```
clamp(1.0f, 2.0f, x) -> max(2.0f, x)
```

We could add a folding rule for when the first two values are constants.  However the result is `min(2.0f, x)`.  Same as above.  I do not expect this to be common.

```
clamp(x, 1.0f, 2.0f) -> min(x, 1.0f)
```

This cannot be simplified.  If `x > 2`, then the result is `2`.  This will be the most common situation.

alan-baker: I think I had my operands in the wrong order. Above I was using (incorrectly), min, val, max.

s-perron: I added folding rules for the first two.  We can leave the others for another PR if we see them showing up.

PR #2838, opened 2019-09-06
Fix add-dead-break and add-dead-continue passes to respect dominance

afd:
The implementation of these passes had overlooked the fact that adding
a new edge to a control flow graph can change dominance information.
Adding a dead break/continue risks causing uses to no longer be
dominated by their definitions.  This change introduces various tests
to expose such scenarios, and augments the preconditions for these
transformations with checks to guard against the situation.

### DIFF
--- a/source/fuzz/fuzzer_util.cpp
+++ b/source/fuzz/fuzzer_util.cpp
@@ -246,37 +246,74 @@ bool CanReachWithoutTraversingLoopBackEdge(opt::IRContext* context,
 bool NewEdgeLeavingConstructBodyRespectsUseDefDominance(
     opt::IRContext* context, opt::BasicBlock* bb_from, opt::BasicBlock* bb_to,
     opt::BasicBlock* construct_merge_block) {
-  // Check that the dominators for id uses in the loop's continue target are not
-  // spoiled by the new edge.
   auto enclosing_function = bb_from->GetParent();
   auto dominator_analysis = context->GetDominatorAnalysis(enclosing_function);
+
+  // We are concerned about definitions occurring in the construct being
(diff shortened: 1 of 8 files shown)


source/fuzz/fuzzer_util.cpp
```diff
@@ -205,6 +205,123 @@ opt::BasicBlock::iterator GetIteratorForBaseInstructionAndOffset(
   return nullptr;
 }
 
+// Helper method to figure out whether there is a back-edge-free path from
+// |source| to |dest|.
+bool CanReachWithoutTraversingLoopBackEdge(opt::IRContext* context,
+                                           opt::BasicBlock* source,
+                                           opt::BasicBlock* dest) {
+  // First, check that the destination block occurs after the source block in
```
alan-baker: This isn't true

```
for (;;) {
  // a
  if (cond) {
    // b
  } else {
    // c
  }
  // d
}
// e
```
Blocks `b` and `c` may appear after the loop merge `e` in the module.  So `b` and `c` could reach `d` and `e` without traversing a back-edge, but not be found in the iteration.

There is an optimizer utility to order the blocks in "structured order" so that all the blocks contained in a construct are ordered before the merge.

There is a similar issue in the iteration in the other function, but I think it will generally result in overly conservative decisions.

afd: Thanks - you're right of course!

I have done a commit that adds a test exposing the problem you have identified, and that includes a new algorithm, designed in collaboration with @paulthomson, which we believe does the right thing.

It is inspired by the structured order algorithms in the optimizer - in particular the idea of doing a depth-first search where merge blocks are visited before successors.

I personally don't think it's worth trying to directly re-use code here.

Issue #2454, opened 2019-03-15
spirv-opt generates code that spirv-val rejects: The continue construct with the continue target ... is not post dominated by the back-edge block ...


afd:
[Files to reproduce problem](https://github.com/KhronosGroup/SPIRV-Tools/files/2968970/to_file3.zip)

glsl_reduced.frag is a minimized GLSL shader.  When compiled to SPIR-V with glslangValidator we get glsl_reduced.spv.  I tried to further reduce this using spirv-reduce, but it failed due to a spirv-reduce bug, which I will report separately.

Doing:

spirv-opt glsl_reduced.spv --convert-local-access-chains --eliminate-local-singl
e-store --eliminate-local-single-store --eliminate-dead-code-aggressive --copy-p
ropagate-arrays --inline-entry-points-exhaustive --if-conversion --eliminate-loc
al-single-store --scalar-replacement=100 --inline-entry-points-exhaustive --priv
ate-to-local --ccp --eliminate-dead-branches --private-to-local --vector-dce --s
calar-replacement=100 --merge-blocks --eliminate-dead-branches --eliminate-dead-
branches --merge-return --scalar-replacement=100 --simplify-instructions --merge
-blocks --private-to-local --inline-entry-points-exhaustive --eliminate-local-si
ngle-block -o temp.spv

spirv-val temp.spv

should trigger the problem:

error: line 77: The continue construct with the continue target 46[%46] is not p
ost dominated by the back-edge block 58[%58]
  %58 = OpLabel

I did not try to reduce the spirv-opt flags, in case (as with #2450) there may be an issue both with the optimizer generating invalid code and the validator not picking this up soon enough.

Found using [GraphicsFuzz](http://github.com/google/graphicsfuzz)

alan-baker:
Same underlying issue as #2433.

PR #2842, opened 2019-09-09
Don't inline function containing OpKill

### LINKED ISSUE #2433: spirv-opt improper inlining OpKill into continue construct
@alan-baker asked me to open this issue here from https://github.com/KhronosGroup/SPIRV-Headers/issues/86

Original issue:
Is a continue construct that consists of OpKill valid? If so, the rules on the header block post-dominating the continue construct need to be fixed. Otherwise, there's a bug in spirv-opt where `kill()` is improperly inlined.

http://shader-playground.timjones.io/1323167d9be3ec6d7a54996e098612de

spirv-opt output:
```
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 7
; Bound: 13
; Schema: 0
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpSource GLSL 330
               OpName %main "main"
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
          %5 = OpLabel
               OpLoopMerge %11 %12 None
               OpBranch %12
         %12 = OpLabel
               OpKill
         %11 = OpLabel
               OpUnreachable
               OpFunctionEnd
```

spirv-opt input:
```
// Module Version 10000
// Generated by (magic number): 80007
// Id's are bound by 17

                              Capability Shader
               1:             ExtInstImport  "GLSL.std.450"
                              MemoryModel Logical GLSL450
                              EntryPoint Fragment 4  "main"
                              ExecutionMode 4 OriginUpperLeft
                              Source GLSL 330
                              Name 4  "main"
                              Name 6  "kill("
               2:             TypeVoid
               3:             TypeFunction 2
              14:             TypeBool
              15:    14(bool) ConstantTrue
         4(main):           2 Function None 3
               5:             Label
                              Branch 9
               9:             Label
                              LoopMerge 11 12 None
                              Branch 13
              13:             Label
                              BranchConditional 15 10 11
              10:               Label
                                Branch 12
              12:               Label
              16:           2   FunctionCall 6(kill()
                                Branch 9
              11:             Label
                              Return
                              FunctionEnd
        6(kill():           2 Function None 3
               7:             Label
                              Kill
                              FunctionEnd
```

### LINKED ISSUE #2554: Folding rules for 1.4 selects
SPIR-V 1.4 allows selects between composites with a scalar bool (even for vectors). Folding rules should reflect this.

s-perron:
If an OpKill instruction is inlined into a continue construct, then the
spir-v is no longer valid.  To avoid this issue, we do inline into an
OpKill at all.  This method was chosen because it is difficult to keep
track of whether or not you are in a continue construct while changing
the function that is being inlined into.  This will work well with wrap
OpKill because every will still be inlined except for the OpKill
instruction itself.

Fixes #2554
Fixes #2433

This reverts commit aa9e8f538041db3055ea443080e0ccc315fa114f.

### DIFF
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -720,6 +720,15 @@ bool InlinePass::IsInlinableFunction(Function* func) {
     return false;
   }
 
+  // Do not inline functions with an OpKill because they may be inlined into a
+  // continue construct.
+  bool has_opkill = !func->WhileEachInst(
+      [](Instruction* inst) { return inst->opcode() != SpvOpKill; });
+
+  if (has_opkill) {
+    return false;
+  }
+
   return true;
 }
 
--- a/test/opt/inline_test.cpp
+++ b/test/opt/inline_test.cpp
@@ -2792,182 +2792,6 @@ TEST_F(InlineTest, SetParent) {
   }
 }
 
-TEST_F(InlineTest, OpKill) {
-  const std::string text = R"(
-; CHECK: OpFunction
-; CHECK-NEXT: OpLabel
-; CHECK-NEXT: OpKill
-; CHECK-NEXT: OpLabel
-; CHECK-NEXT: OpReturn
-; CHECK-NEXT: OpFunctionEnd
-OpCapability Shader
-OpMemoryModel Logical GLSL450
-OpEntryPoint Fragment %main "main"
-OpExecutionMode %main OriginUpperLeft
-%void = OpTypeVoid
-%voidfuncty = OpTypeFunction %void
-%main = OpFunction %void None %voidfuncty
-%1 = OpLabel
-%2 = OpFunctionCall %void %func
-OpReturn
-OpFunctionEnd
-%func = OpFunction %void None %voidfuncty
-%3 = OpLabel
-OpKill
-OpFunctionEnd
-)";
-
-  SinglePassRunAndMatch<InlineExhaustivePass>(text, true);
-}
-
-TEST_F(InlineTest, OpKillWithTrailingInstructions) {
-  const std::string text = R"(
-; CHECK: OpFunction
-; CHECK-NEXT: OpLabel
-; CHECK-NEXT: [[var:%\w+]] = OpVariable
-; CHECK-NEXT: OpKill
-; CHECK-NEXT: OpLabel
-; CHECK-NEXT: OpStore [[var]]
-; CHECK-NEXT: OpReturn
-; CHECK-NEXT: OpFunctionEnd
-OpCapability Shader
-OpMemoryModel Logical GLSL450
-OpEntryPoint Fragment %main "main"
-OpExecutionMode %main OriginUpperLeft
-%void = OpTypeVoid
-%bool = OpTypeBool
-%true = OpConstantTrue %bool
-%bool_func_ptr = OpTypePointer Function %bool
-%voidfuncty = OpTypeFunction %void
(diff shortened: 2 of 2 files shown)


greg-lunarg:
> To avoid this issue, we do inline into an OpKill at all.

I am having trouble parsing this sentence.

> This will work well with wrap OpKill because every will still be inlined except for the OpKill instruction itself.

I am having trouble parsing this sentence too.

s-perron:
Sorry, I should proof read better.  I've updated it.

greg-lunarg:
I am sorry I didn't jump in on the triggering issue sooner.

My biggest concern with this solution is that we are going to require folks who roll their own optimization to know about this restriction, a restriction which to me seems fairly unintuitive. It will also require them to invoke a separate function wrapping pass to get the optimization that they would normally expect from an inliner.

It seems strange to me that a valid program can be made invalid through inlining, and that an invalid programming can be made valid by wrapping an instruction in a function. Judging by the threads coming from the triggering issue, I think everyone involved shares this concern to a certain extent.

The current situation suggests to me that our definition of what a valid program is a little out of whack. As such, it seems to me that it might be better to fix the problem by changing the definition of a valid program rather than changing the optimizer.

Given that, with this solution, we are allowing OpKill to be executed within a continue construct, albeit wrapped in its own function, it seems that outlawing OpKill within the call tree of a continue construct is not the way to go.

I would rather suggest the following: where the spec says the merge block must post-dominate the continue target, we add "ignoring OpKill". That is, for the purposes of continue target validation, an OpKill in a continue construct is treated as, for example, an unconditional branch to the merge block.

Given this, we would no longer need people to be aware that we don't inline functions with OpKill, which is somewhat unintuitive, and require them to preventatively call a separate pass to wrap OpKill in a function.

If we think this all makes sense, for consistency we will probably want to actually add "ignoring OpKill and OpReturn", since OpReturn has a similar effect on a continue construct, though it lacks the interaction with inlining and function wrapping that OpKill has.

Thoughts?

s-perron:
I already brought this up with the spir-v committee.  See the gitlab issue https://gitlab.khronos.org/spirv/SPIR-V/issues/459.

greg-lunarg:
Thanks for the link. It is very helpful to know what has already been discussed.

FWIW, my proposal was not one of the alternatives that were considered, so I still feel it might be worth bringing up. While I agree allowing OpKill to appear in a continue construct is not a highly used capability, it would allow us to easily remove the unintuitive and draconian restriction that functions with OpKill cannot be inlined, so that is something.

Still, there may be another alternative which is not quite as good but would probably be acceptable in terms of code quality, would let us remove this restriction and wouldn't require us to go back through the SPIR-V WG:

Have the inliner do the OpKill wrapping itself. It would do the wrapping pass as the first thing before any inlining. A slight variation would be to have the inliner do the wrapping as part of the inlining process.

The advantage here is that the user does not have to know anything about OpKill, inlining, and wrapping.

Thoughts?

s-perron:
Yes that could work.  In general, I prefer keeping each pass as simple as possible.  That is why I separated them.  So I don't know how you were thinking of implementing this, but if we were to do something like that, I would want the options and api calls to imply both passes instead of having a single pass.

If you have concerns about this, open another issue to discuss it.

greg-lunarg:
How does this PR fix #2554 ?

s-perron:
Typo.  Updated to #2454.

PR #2852, opened 2019-09-11
Assembler: Can't set an ID in instruction without result ID

### LINKED ISSUE #2257: Assembler should fail if input assigns to Id but instruction does not
For example, this should fail to assemble:

          %dud =  OpCopyMemory %2 %3

Instead I get a binary that looks like this:

```
$ od -t x4 a.spv
0000000 07230203 00010400 00070000 00000003
0000020 00000000 0003003f 00000001 00000002
0000040
```

 and which disassembles to:

      OpCopyMemory %1 %2

dneto0:
Fix tests that violated this rule.

Fixes #2257


dneto0:
cc: @baldurk : I'm closing this loophole.

PR #2857, opened 2019-09-12
Add fuzzer pass to replace ids with synonyms

afd:
If the fuzzer's fact manager knows that ids A and B are synonymous, it
can replace a use of A with a use of B, so long as various conditions
hold (e.g. the definition of B must dominate the use of A, and it is
not legal to replace a use of an OpConstant in a struct's access chain
with a synonym that is not an OpConstant).

This change adds a fuzzer pass to sprinke such synonym replacements
through the module.

### DIFF
--- a/source/fuzz/fuzzer_pass_apply_id_synonyms.cpp
+++ b/source/fuzz/fuzzer_pass_apply_id_synonyms.cpp
@@ -14,7 +14,8 @@
 
 #include "source/fuzz/fuzzer_pass_apply_id_synonyms.h"
 
-#include "source/opt/ir_context.h"
+#include "source/fuzz/id_use_descriptor.h"
+#include "source/fuzz/transformation_replace_id_with_synonym.h"
 
 namespace spvtools {
 namespace fuzz {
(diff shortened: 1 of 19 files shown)


source/fuzz/fuzzer_pass_apply_id_synonyms.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "source/fuzz/fuzzer_pass.h"
+
+#include "source/opt/ir_context.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// An initial pass for adding useful ingredients to the module, such as boolean
+// constants, if they are not present.
```
paulthomson: Fix comment?

source/fuzz/fuzzer_pass_apply_id_synonyms.cpp
```diff
@@ -0,0 +1,108 @@
...
+          for (auto& data_descriptor :
+               GetFactManager()->GetSynonymsForId(id_with_known_synonyms)) {
+            synonyms_to_try.push_back(&data_descriptor);
+          }
+          while (!synonyms_to_try.empty()) {
+            auto synonym_index =
+                GetFuzzerContext()->RandomIndex(synonyms_to_try);
+            auto synonym_to_try = synonyms_to_try[synonym_index];
+            synonyms_to_try.erase(synonyms_to_try.begin() + synonym_index);
+            assert(synonym_to_try->index().size() == 0 &&
```
paulthomson: Use `.empty()` (says clang-tidy)

source/fuzz/transformation_replace_id_with_synonym.cpp
```diff
@@ -0,0 +1,150 @@
...
+    if (!dominator_analysis->Dominates(
+            context->get_instr_block(defining_instruction)->id(),
+            parent_block)) {
+      return false;
+    }
+  } else if (!dominator_analysis->Dominates(defining_instruction,
+                                            use_instruction)) {
+    return false;
+  }
+  return true;
```
paulthomson: The last case is pretty clear and I know redundant comments are silly, but for consistency, perhaps you could add a comment explaining what you are checking in the last case.

source/fuzz/fuzzer_pass_apply_id_synonyms.h
afd: Good catch!

afd: Good catch!

source/fuzz/transformation_replace_id_with_synonym.cpp
afd: Agreed - done.

PR #2862, opened 2019-09-12
Adding OpEntryPoint validation

raunraun:
OpEntryPoint

Declare an entry point, its execution model, and its interface.

Execution Model is the execution model for the entry point and its static call tree. See Execution Model.

Entry Point must be the Result <id> of an OpFunction instruction.

**Name is a name string for the entry point. A module cannot have two OpEntryPoint instructions with the same Execution Model and the same Name string.**


raunraun:
Alright, take 2.  Thansk @alan-baker

alan-baker:
Can you please rebase to pick up the 1.5 tooling updates?

Issue #2863, opened 2019-09-13
ArrayStride decoration check too strict


asuonpaa:
The SPIR-V code below gets a validation error saying %outputArray is missing an ArrayStride decoration.

```
OpDecorate %struct Block
%struct = OpTypeStruct %matrix %v4f32 %f32 %f32 %f32 %f32
%numElements = OpConstant %u32 8
%outputArray = OpTypeArray %struct %numElements
%Output = OpTypeStruct %outputArray
%_ptr_Output = OpTypePointer StorageBuffer %Output
```

But SPIR-V spec says 

> Each array type must have an ArrayStride decoration, unless it is an array that contains a structure decorated with Block or BufferBlock, in which case it must not have an ArrayStride decoration.
> 

I suppose this exception is not handled correctly.

alan-baker:
Could you attachment the whole shader? I have not been able to reproduce this based on the fragment you provided; however, the fragment looks invalid because the array is within a struct. That outer struct should be decorated with Block to be valid. If you want to generate an array of descriptors, drop the outer struct and the shader should be ok.

asuonpaa:
> Could you attachment the whole shader? I have not been able to reproduce this based on the fragment you provided; however, the fragment looks invalid because the array is within a struct. That outer struct should be decorated with Block to be valid. If you want to generate an array of descriptors, drop the outer struct and the shader should be ok.

Here is the whole shader. This might be from different case than the previous fragments, but at least it has the outer struct decorated as block.

```
Validation FAILED: Structure id 4 decorated as Block must be explicitly laid out with ArrayStride decorations.
  %_struct_4 = OpTypeStruct %_arr__struct_7_uint_8_0
; SPIR-V
; Version: 1.0
; Generator: Khronos SPIR-V Tools Assembler; 0
; Bound: 42
; Schema: 0
               OpCapability Shader
               OpCapability VariablePointers
               OpExtension "SPV_KHR_storage_buffer_storage_class"
               OpExtension "SPV_KHR_variable_pointers"
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %2 "main" %gl_GlobalInvocationID
               OpExecutionMode %2 LocalSize 1 1 1
               OpSource GLSL 430
               OpDecorate %gl_GlobalInvocationID BuiltIn GlobalInvocationId
               OpMemberDecorate %_struct_4 0 Offset 0
[...]

PR #2864, opened 2019-09-13
Extra resource interface validation

alan-baker:
* Vulkan specific checks
  * storage buffer variables must be structs or arrays of structs
  * storage buffer struct must be Block decorated
  * uniform struct must be Block or BufferBlock decorated
* new tests

Also tested against CTS.


alan-baker:
This picked up the 1.5 headers, which require other fixes to SPIRV-Tools. I'll get a new bot run after those fixes.

Separately this helps to address what I think is the underlying issue in #2863.

alan-baker:
Rebased on 1.5

PR #2867, opened 2019-09-16
Fix description of inst interators in opt::Function

dneto0:


source/opt/function.h
```diff
@@ -104,19 +104,26 @@ class Function {
     });
   }
 
-  // Runs the given function |f| on each instruction in this function, and
-  // optionally on debug line instructions that might precede them.
+  // Runs the given function |f| on instructions in this function, in order,
+  // and optionally on debug line instructions that might precede them.
+  // If |f| returns false, iteration is terminated.
```
s-perron: This is not true for the "ForEachInst" version.  |f| has no return value.

source/opt/function.h
```diff
@@ -104,19 +104,26 @@ class Function {
...
   bool WhileEachInst(const std::function<bool(Instruction*)>& f,
                      bool run_on_debug_line_insts = false);
   bool WhileEachInst(const std::function<bool(const Instruction*)>& f,
                      bool run_on_debug_line_insts = false) const;
 
   // Runs the given function |f| on each parameter instruction in this function,
-  // and optionally on debug line instructions that might precede them.
+  // in order, and optionally on debug line instructions that might precede
+  // them. If |f| returns false, iteration is terminated and this function
+  // returns false.
```
s-perron: |f| is a void function, and does not return true or false.

s-perron:
@dneto0 Will you be updating this?

dneto0:
whoops.  yeah, I'll fix. thanks for the review. ;-)

source/opt/function.h
dneto0: Fixed in next commit

source/opt/function.h
dneto0: Fixed in next commit.

dneto0:
Hey @s-perron, this is ready for review again

PR #2873, opened 2019-09-17
Allow validation during spirv-fuzz replay

afd:
To aid in debugging issues in spirv-fuzz, this change adds an option whereby the SPIR-V module is validated after each transformation is applied during replay.  This can assist in finding a transformation that erroneously makes the module invalid, so that said transformation can be debugged.

### DIFF
--- a/source/fuzz/replayer.cpp
+++ b/source/fuzz/replayer.cpp
@@ -85,11 +85,11 @@ Replayer::ReplayerResultStatus Replayer::Run(
       impl_->target_env, impl_->consumer, binary_in.data(), binary_in.size());
   assert(ir_context);
 
-  std::vector<uint32_t> last_good_binary;
-  std::vector<uint32_t> first_bad_binary;
-
+  // For replay validation, we track the last valid SPIR-V binary that was
+  // observed. Initially this is the input binary.
+  std::vector<uint32_t> last_valid_binary;
   if (impl_->validate_during_replay) {
-    ir_context->module()->ToBinary(&last_good_binary, false);
+    last_valid_binary = binary_in;
   }
 
   FactManager fact_manager;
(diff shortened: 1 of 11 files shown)


source/fuzz/replayer.cpp
```diff
@@ -80,6 +85,13 @@ Replayer::ReplayerResultStatus Replayer::Run(
       impl_->target_env, impl_->consumer, binary_in.data(), binary_in.size());
   assert(ir_context);
 
+  std::vector<uint32_t> last_good_binary;
+  std::vector<uint32_t> first_bad_binary;
```
paulthomson: Is this used?

source/fuzz/shrinker.cpp
```diff
@@ -180,7 +180,7 @@ Shrinker::ShrinkerResultStatus Shrinker::Run(
       // transformations inapplicable.
       std::vector<uint32_t> next_binary;
       protobufs::TransformationSequence next_transformation_sequence;
-      if (Replayer(impl_->target_env)
+      if (Replayer(impl_->target_env, false)
```
paulthomson: Should we allow validation in the shrinker too? Can you make the option affect the shrinker?

source/fuzz/replayer.cpp
afd: No - good catch.  Initially I had been imagining saving out the last good and last bad binaries, but then I realised it would probably be more convenient to just have a place where a breakpoint could be set, so I didn't end up needing "first_bad_binary" (but forgot to remove it).

source/fuzz/shrinker.cpp
afd: Yes - good idea, done.

I have also enabled this in the replayer and shrinker tests (might make them take too long though).

PR #2879, opened 2019-09-18
Add SPV_KHR_shader_clock validation

AaronHaganAMD:
Add checks for SPV_KHR_shader_clock

1) Check that the Result Type must be a 64-bit unsigned integer type or a vector of two-components of 32-bit unsigned integer.

2) Check for a valid execution scope.

### DIFF
--- a/source/val/validate_misc.cpp
+++ b/source/val/validate_misc.cpp
@@ -18,6 +18,7 @@
 #include "source/opcode.h"
 #include "source/spirv_target_env.h"
 #include "source/val/instruction.h"
+#include "source/val/validate_scopes.h"
 #include "source/val/validation_state.h"
 
 namespace spvtools {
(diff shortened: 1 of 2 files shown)


DEPS
```diff
@@ -6,7 +6,7 @@ vars = {
   'effcee_revision': 'b83b58d177b797edd1f94c5f10837f2cc2863f0a',
   'googletest_revision': '2f42d769ad1b08742f7ccb5ad4dd357fc5ff248c',
   're2_revision': 'e356bd3f80e0c15c1050323bb5a2d0f8ea4845f4',
-  'spirv_headers_revision': '123dc278f204f8e833e1a88d31c46d0edf81d4b2',
+  'spirv_headers_revision': 'e4322e3be589e1ddd44afb20ea842a977c1319b8',
```
dneto0: This is now obsolete 
See https://github.com/KhronosGroup/SPIRV-Tools/pull/2880

source/val/validate_misc.cpp
```diff
@@ -35,6 +35,37 @@ spv_result_t ValidateUndef(ValidationState_t& _, const Instruction* inst) {
...
+      return _.diag(SPV_ERROR_INVALID_DATA, inst) << "ExecutionScope should be "
+                                                     "ScopeSubgroup or ScopeDevice";
+    }
+  }
+
+  // Result Type must be a 64 - bit unsigned integer type or
+  // a vector of two - components of 32 -
+  // bit unsigned integer type 
+  const uint32_t result_type = inst->type_id();
+  if (((!_.IsUnsignedIntVectorType(result_type) ||
```
dneto0: should also check that the components of the vector are 32 bits wide.

alan-baker:
```
if (!(_.IsUnsignedIntScalarType(result_type) && _.GetBitWidth(result_type) == 64) &&
    !(_.IsUnsignedIntVectorType(result_type) && _.GetDimension(result_type) == 2 && _.GetBitWidth(result_type) == 32)) {
  // error
}
```

Then we don't need the new method.

AaronHaganAMD:
Added tests..

alan-baker:
When you fix the formatting issues, could you also rebase onto master?

PR #2881, opened 2019-09-18
Update DEPS: effcee, re2, googletest

dneto0:


alan-baker:
Your change seems unrelated to a fuzzer test. @afd and @paulthomson is this a known issue?

afd:
I will take a look.

afd:
When I checkout @dneto0's branch and run it locally the tests pass.  Thus this is probably due to the tests employing randomization - although we use fixed seeds, there could still be differences across platforms.

@dneto0: #2874 included a fix for a bug in the fuzzer I found earlier this week; I think that probably wasn't in by the time the CI ran for this PR - could you rebase against master and try again?  It might have been that bug manifesting.

paulthomson:
@afd Looks like the change was in.

dneto0:
> Thus this is probably due to the tests employing randomization - although we use fixed seeds, there could still be differences across platforms.

Uh, yeah.  Sounds quite plausible.  Things like conformance tests use their own copy of some "standard" open source random number generator exactly for their reproducibility.
E.g. as I recall OpenCL conformance uses a mersenne twister  PRNG.

afd:
OK, @dneto0 - please rebase w.r.t. master and try again - the problematic tests won't run during CI now.

PR #2882, opened 2019-09-18
Fix header include syntax.

ehsannas:


dneto0:
Now this fails format because clang-format wants to move the headers down.

ehsannas:
Ah ... sorry forgot to format. fix coming up.

source/opt/function.cpp
```diff
@@ -16,9 +16,9 @@
 #include "function.h"
 #include "ir_context.h"
```
dj2: These two should move down with bit_vector.

source/opt/function.cpp
```diff
@@ -16,9 +16,9 @@
...
 
-#include "source/util/bit_vector.h"
 #include <ostream>
 #include <sstream>
+#include "source/util/bit_vector.h"
```
dj2: Should be a new line before this to separate system headers.

s-perron: Line 19 is a space between `function.h` and `sstream`.  I don't think there should be space between `ir_context.h` and `bit_vector.h` because they are both part of the spirv-tools repo.  The first two are in the opt directory, the other is in the util.

I'm wondering if we should have a full path for all of the includes to be consistent.  For example, `function.h` becomes `source/opt/function.h`.

dj2: +1 for full paths, but I'd do it in a different CL. Then turn on the linter, heh.

ehsannas: +1 for full paths.

PR #2883, opened 2019-09-18
Android.mk: Add dependency from optimizer file to amd-shader-ballot-insts.inc

dneto0:
This is required to fix the NDK build.


dneto0:
LGTM tested locally.
Merging this to fix the NDK build.

PR #2884, opened 2019-09-18
Fix comment typo found by protobufs linter

dneto0:


dneto0:
This was found by the protobufs linter when I integrated recent changes into the Android NDK build.

PR #2885, opened 2019-09-19
Add OpCopyMemory test to SVA.

dj2:
This CL adds a test for OpCopyMemory and adds the needed IdMemorySemantics and
IdScope in order to extract the parameters.


dj2:
This ended up mostly just working. I store the operands slightly differently in SVA vs how spirv-as does it. Instead of a list of all operands, the operands have parameters. So, for each entry which requires extra parameters there is a params() list on the Operand.

In this case, we end up with the "volatile|..." operand with a params of [3] and the "Aligned|..." operand with a params of [16, 4]. The overall instruction has 4 operands attached to it.

PR #2891, opened 2019-09-20
Add Bazel build configuration.

ehsannas:
I will add a Kokoro bot to exercise this once it has landed (it works when I use it locally).

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -420,28 +420,35 @@ cc_library(
     deps = [":test_common"],
 )
 
+# PCH (precompiled header) tests only work when using CMake and MSVC on Windows,
+# so they will be skipped in the Bazel builds.
+
 [base_test(
-    name = f[5:-9],  # strip test/, _test.cpp
+    name = f[5:-4],  # strip test/, .cpp
     srcs = [f],
 ) for f in glob(
-    ["test/*_test.cpp"],
+    ["test/*.cpp"],
     exclude = [
-        "test/cpp_interface_test.cpp",
-        "test/log_test.cpp",
-        "test/pch_test.cpp",
-        "test/timer_test.cpp",
+        "test/cpp_interface_test.cpp", # has its own base_test below.
+        "test/log_test.cpp", # has its own base_test below.
+        "test/pch_test.cpp", # pch tests are skipped.
+        "test/timer_test.cpp", # has its own base_test below.
     ],
 )]
+
+# This test uses unistd.h and does not run on Windows.
+base_test(
+    name = "timer_test",
+    srcs = select({
+        "@bazel_tools//src/conditions:windows": [],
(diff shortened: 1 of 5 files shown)


BUILD.bazel
```diff
@@ -0,0 +1,523 @@
...
+    srcs = [f],
+) for f in glob(
+    ["test/link/*_test.cpp"],
+    exclude = ["test/link/type_match_test.cpp"],
+)]
+
+[opt_test(
+    name = f[9:-9],  # strip test/opt/, _test.cpp
+    srcs = [f],
+) for f in glob(["test/opt/*_test.cpp"])]
```
s-perron: This will miss some tests like `ir_builder.cpp`.  It forces a particular naming convention that we general keep, but we have not enforced.  We should try to make it more general.

s-perron: The same goes for other places we look for `*_test.cpp`.

README.md
```diff
@@ -286,6 +301,8 @@ installed regardless of your OS:
   2.8.12 or later.
```
s-perron: You should add a phrase similar to what you have with Bazel, "if using CMake for generating compilation targets."

README.md
```diff
@@ -286,6 +301,8 @@ installed regardless of your OS:
...
 - [Python 3](http://www.python.org/): for utility scripts and running the test
 suite.
+- [Bazel](https://baze.build/) (optional): if building the source with Bazel,
+you need to install Bazel on your machine.
```
s-perron: Is there a particular version requirement?

BUILD.bazel
ehsannas: Done

BUILD.bazel
```diff
@@ -0,0 +1,523 @@
...
+) for f in glob(["test/util/*_test.cpp"])]
+
+[val_test(
+    name = f[13:-9],  # strip test/val/val_, _test.cpp
+    srcs = [f],
+) for f in glob(
+    ["test/val/val_*_test.cpp"],
+    exclude = [
+        "test/val/val_capability_test.cpp",
+        "test/val/val_limits_test.cpp",
```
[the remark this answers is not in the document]
ehsannas: Done

README.md
ehsannas: Done

README.md
ehsannas: Done

BUILD.bazel
```diff
@@ -420,28 +420,35 @@ cc_library(
...
+    name = "timer_test",
+    srcs = select({
+        "@bazel_tools//src/conditions:windows": [],
+        "//conditions:default": ["test/timer_test.cpp"],
+    }),
 )
 
 base_test(
-    name = "ilist_test",
-    srcs = ["test/util/ilist_test.cpp"],
```
ehsannas: this test is now covered under the util_test

PR #2904, opened 2019-09-24
spirv-fuzz: Fixes to preconditions for adding dead break/continue edges

### LINKED ISSUE #2898: spirv-fuzz: add-dead-continue can erroneously add an edge from a continue target to itself
spirv-fuzz regards adding a dead continue from a continue target to itself as legal - this creates a self-loop from the continue target to itself, which is invalid.

We want to allow a dead continue to be added to a loop that is itself in some other loop's continue construct.  But we do not want to allow a dead continue from the continue construct of a loop L to the continue target for L.

I suspect the problem may be that we depend on an analysis that regards a loop's continue target as part of the loop.

### LINKED ISSUE #2900: spirv-fuzz: add dead continue can break dominance rules
If a loop's continue target C is unreachable and appears before block B in the loop, then adding a dead continue edge from B to the C will make B dominate C, and violate the rule that block ordering must respect dominance.

Adding of dead continues needs to be refined to avoid this scenario.

### LINKED ISSUE #2905: spirv-fuzz: transformations can lead to invalid shader: "ID ... defined in block ... does not dominate its use in block ..."
[tofile.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/3647813/tofile.zip)

Command to reproduce:

spirv-fuzz shader.frag.spv --replay=shrunk.transformations -o temp.spv && spirv-val temp.spv

afd:
Issues #2898 and #2900 identify some cases where adding a dead
continue would lead to an invalid module, and these turned out to be
due to the lack of sensible dominance information when a continue
target is unreachable.  This change requires that the header of a loop
dominates the loop's continue target if a dead continue is to be
added.

Furthermore, issue #2905 identified a shortcoming in the algorithm
being used to identify when it is OK, from a dominance point of view,
to add a new break/continue edge to a control flow graph.  This change
replaces that algorithm with a simpler and more obviously correct
algorithm (that incidentally does not require the new edge to be a
break/continue edge in particular).

Fixes #2898.
Fixes #2900.
Fixes #2905.

### DIFF
--- a/source/fuzz/fuzzer_util.cpp
+++ b/source/fuzz/fuzzer_util.cpp
@@ -205,156 +205,63 @@ opt::BasicBlock::iterator GetIteratorForBaseInstructionAndOffset(
   return nullptr;
 }
 
-// Returns the ids of all successors of |block|
-std::vector<uint32_t> GetSuccessors(opt::BasicBlock* block) {
-  std::vector<uint32_t> result;
-  switch (block->terminator()->opcode()) {
-    case SpvOpBranch:
-      result.push_back(block->terminator()->GetSingleWordInOperand(0));
-      break;
-    case SpvOpBranchConditional:
-      result.push_back(block->terminator()->GetSingleWordInOperand(1));
-      result.push_back(block->terminator()->GetSingleWordInOperand(2));
-      break;
-    case SpvOpSwitch:
-      for (uint32_t i = 1; i < block->terminator()->NumInOperands(); i += 2) {
-        result.push_back(block->terminator()->GetSingleWordInOperand(i));
-      }
-      break;
-    default:
-      break;
+bool NewEdgeRespectsUseDefDominance(opt::IRContext* context,
+                                    opt::BasicBlock* bb_from,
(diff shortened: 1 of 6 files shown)


source/fuzz/fuzzer_util.cpp
```diff
@@ -205,156 +205,63 @@ opt::BasicBlock::iterator GetIteratorForBaseInstructionAndOffset(
...
-void FindBypassedBlocks(opt::IRContext* context, opt::BasicBlock* bb_from,
-                        opt::BasicBlock* bb_to,
-                        std::set<opt::BasicBlock*>* bypassed_blocks) {
-  // This algorithm finds all blocks different from |bb_from| that:
-  // - are in the innermost structured control flow construct containing
-  // |bb_from|
-  // - can be reached from |bb_from| without traversing a back-edge or going
-  // through |bb_to|
+  // Suppose that some block Y is dominated by |bb_to| (which includes the case
+  // where X = |bb_to|).
```
paulthomson: Just checking: do you mean Y above?

PR #2910, opened 2019-09-24
Handle extract with no indexes

s-perron:
It is possible that OpCompositeExtract instructions will not have any
indexes.  This is not handled well by scalar replacement and instruction
folding.

Fixes https://crbug.com/1006435


source/opt/folding_rules.cpp
```diff
@@ -1401,6 +1407,20 @@ FoldingRule CompositeExtractFeedingConstruct() {
   };
 }
 
+// Folds an OpCompositeExtact instruction with no indexes into an OpCopyObject.
```
dj2: s/OpCompositeExtact/OpCompositeExtract

s-perron: done

PR #2911, opened 2019-09-24
Remove validate_datarules.cpp

alan-baker:
The checks are now done in a more consistent location with the rest of the validator. Checks are now pulled into specific opcode validation to be more complete and reduce duplication.

* Checks moved into individual opcode validation
  * removes duplicated checks
* Add check that forward pointer points to struct

I'll update this after the bazel build is submitted if necessary.


source/val/validate_type.cpp
```diff
@@ -246,19 +353,12 @@ spv_result_t ValidateTypeStruct(ValidationState_t& _, const Instruction* inst) {
...
     }
     if (_.IsForwardPointer(member_type_id)) {
-      // If we're dealing with a forward pointer:
-      // Find out the type that the pointer is pointing to (must be struct)
-      // word 3 is the <id> of the type being pointed to.
-      auto type_pointing_to = _.FindDef(member_type->words()[3]);
-      if (type_pointing_to && type_pointing_to->opcode() != SpvOpTypeStruct) {
-        // Forward declared operands of a struct may only point to a struct.
+      const auto inst_index = inst - &_.ordered_instructions()[0];
+      const auto member_index = inst - &_.ordered_instructions()[0];
```
dj2: Is one of these indexes wrong? Otherwise, won't these always be the same value?

alan-baker: Yes, thanks

alan-baker: I removed this check. It gets covered by checking forward references.

Issue #2912, opened 2019-09-24
#2790, #2842 OpKill changes causing size regressions


greg-lunarg:
PRs #2790 and #2842 were done to fix a problem: stop inlining an OpKill into a loop's continue construct.

Unfortunately, these changes are now causing size regressions for real shaders that do not call functions with OpKill from a continue construct.

Case in point: the attached SPIR-V shader previously optimized to 940 bytes, now optimizes to 1148, an increase in file size of 20% and nearly 100% increase in function size.

I understand that @s-perron attempted to only "fix" problematic shaders but wasn't able to. Unfortunately, a solution was devised that caused **all** shaders with OpKill to regress in optimized size.

I propose a possible solution: create an analysis which determines if a shader contains a call to a function containing an OpKill from a continue construct. If it does, WrapOpKill and Inlining behave as they do now. If it does not, WrapOpKill returns without wrapping anything and Inlining performs without worrying about OpKill.

The analysis would be run at the beginning of both WrapOpKill and Inlining passes. There might even be a way to preserve the analysis from WrapOpKill to Inlining.

While, admittedly, the increase in size will be less dramatic for larger shaders, the case that this fix is guarding against is so rare, any penalty incurred by non-problematic shaders is difficult to justify.

greg-lunarg:
Attn: @danginsburg

greg-lunarg:
Here is the HLSL source for the shader:

[glsh.frag.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/3649247/glsh.frag.txt)

greg-lunarg:
@s-perron Can you please review the proposal above?

s-perron:
We could do something like that.  That gives me another idea.  We could write an analysis that determines which functions are call from within a continue construct.  This seems like it would be a simple analysis.  Then that could be used in wrap-opkill and inlining.

In wrap-opkill, we apply the transformation only if a function is called from a continue construct.  In most cases, this will mean that wrap-opkill does not do anything.  Most of the work will be in updating the tests.

In inlinine, we say that if a function is called from a continue construct and it contains an opkill, we do not inline.  It could be the function will not be inlined in every place it could.  If it is call from multiple places, but only one is in a continue construct, then we will never inline it.  However, in the most common cases, everything will be inlined.

I would be okay with either idea.  Whatever you think is best.

greg-lunarg:
Yes, Your idea sounds very good.

[...]

PR #2922, opened 2019-09-26
Add continue construct analysis to struct cfg analysis

s-perron:
Add the ability to identify which blocks are in the continue construct for a
loop, and to get functions that are called from those blocks, directly or
indirectly.

Part of https://github.com/KhronosGroup/SPIRV-Tools/issues/2912.

### DIFF
--- a/source/opt/struct_cfg_analysis.cpp
+++ b/source/opt/struct_cfg_analysis.cpp
@@ -45,6 +45,7 @@ void StructuredCFGAnalysis::AddBlocksInFunction(Function* func) {
   struct TraversalInfo {
     ConstructInfo cinfo;
     uint32_t merge_node;
+    uint32_t continue_node;
   };
 
   // Set up a stack to keep track of currently active constructs.
@@ -53,7 +54,9 @@ void StructuredCFGAnalysis::AddBlocksInFunction(Function* func) {
   state[0].cinfo.containing_construct = 0;
   state[0].cinfo.containing_loop = 0;
   state[0].cinfo.containing_switch = 0;
+  state[0].cinfo.in_continue = false;
   state[0].merge_node = 0;
+  state[0].continue_node = 0;
 
   for (BasicBlock* block : order) {
     if (context_->cfg()->IsPseudoEntryBlock(block) ||
@@ -63,6 +66,12 @@ void StructuredCFGAnalysis::AddBlocksInFunction(Function* func) {
 
     if (block->id() == state.back().merge_node) {
       state.pop_back();
+    }
+
+    // This works because the structured order is designed to keep the blocks in
+    // the continue construct between the continue header and the merge node.
+    if (block->id() == state.back().continue_node) {
+      state.back().cinfo.in_continue = true;
     }
 
     bb_to_construct_.emplace(std::make_pair(block->id(), state.back().cinfo));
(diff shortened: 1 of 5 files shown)


source/opt/struct_cfg_analysis.h
```diff
@@ -103,6 +119,7 @@ class StructuredCFGAnalysis {
     uint32_t containing_construct;
     uint32_t containing_loop;
     uint32_t containing_switch;
+    bool in_continue;
```
ehsannas: The comment hasn't been updated for `|containing_switch|` and `|in_continue|`

source/opt/struct_cfg_analysis.cpp
```diff
@@ -146,9 +161,59 @@ bool StructuredCFGAnalysis::IsContinueBlock(uint32_t bb_id) {
...
+    }
+  }
+
+  // Now collect all of the functions that are indirectly called as well.
+  while (!funcs_to_process.empty()) {
+    uint32_t func_id = funcs_to_process.front();
+    funcs_to_process.pop();
+    Function* func = context_->GetFunction(func_id);
+    if (called_from_continue.insert(func_id).second) {
+      context_->AddCalls(func, &funcs_to_process);
```
ehsannas: Does `AddCalls` go recursively deep into functions called by functions inside `func` (i.e. functions indirectly called by `func`) ?

source/opt/struct_cfg_analysis.cpp
```diff
@@ -146,9 +161,59 @@ bool StructuredCFGAnalysis::IsContinueBlock(uint32_t bb_id) {
...
+  }
+  return it->second.in_continue;
+}
+
+bool StructuredCFGAnalysis::IsInContinueConstruct(uint32_t bb_id) {
+  while (bb_id != 0) {
+    if (IsInContainingLoopsContinueConstruct(bb_id)) {
+      return true;
+    }
+    bb_id = ContainingLoop(bb_id);
```
ehsannas: is this for a case where you have a loop inside a continue block?

test/opt/struct_cfg_analysis_test.cpp
```diff
@@ -835,8 +1036,334 @@ OpFunctionEnd
...
+         %15 = OpFunctionCall %void %16
+               OpReturn
+               OpFunctionEnd
+         %14 = OpFunction %void None %7
+         %17 = OpLabel
+         %19 = OpFunctionCall %void %16
+               OpReturn
+               OpFunctionEnd
+         %16 = OpFunction %void None %7
+         %18 = OpLabel
```
ehsannas: Can we have `%16` also call one more level? (e.g.` OpFunctionCall %void %19`), and expect
`UnorderedElementsAre(14u, 16u, 19u)` ?

source/opt/struct_cfg_analysis.cpp
s-perron: Yes, that is correct.  It walking loop next checking if a loop is in the continue construct of the next loop.

source/opt/struct_cfg_analysis.cpp
s-perron: `AddCalls` only add the functions called directly by `func`.  The comment for `AddCalls` could be clearer.

test/opt/struct_cfg_analysis_test.cpp
s-perron: done.

source/opt/struct_cfg_analysis.h
s-perron: Done

source/opt/struct_cfg_analysis.cpp
s-perron: I updated the comments.

PR #2923, opened 2019-09-26
Add removing references to debug instructions when removing them

### LINKED ISSUE #2921: Decorations referring to OpString not being removed by StripDebugInstructionPass
Need to call KillNameAndDecorates on the instructions before removing them, to make sure that references are being cleaned up.

zoddicus:
Fixes #2921

### DIFF
--- a/source/opt/strip_debug_info_pass.cpp
+++ b/source/opt/strip_debug_info_pass.cpp
@@ -22,14 +22,11 @@ Pass::Status StripDebugInfoPass::Process() {
   bool modified = !context()->debugs1().empty() ||
                   !context()->debugs2().empty() ||
                   !context()->debugs3().empty();
-  for (auto& dbg : context()->debugs1())
-    context()->KillNamesAndDecorates(&dbg);
+  for (auto& dbg : context()->debugs1()) context()->KillNamesAndDecorates(&dbg);
 
-  for (auto& dbg : context()->debugs2())
-    context()->KillNamesAndDecorates(&dbg);
+  for (auto& dbg : context()->debugs2()) context()->KillNamesAndDecorates(&dbg);
 
-  for (auto& dbg : context()->debugs3())
-    context()->KillNamesAndDecorates(&dbg);
+  for (auto& dbg : context()->debugs3()) context()->KillNamesAndDecorates(&dbg);
 
   context()->debug_clear();
 
--- a/source/opt/strip_debug_info_pass.cpp
+++ b/source/opt/strip_debug_info_pass.cpp
@@ -22,10 +22,22 @@ Pass::Status StripDebugInfoPass::Process() {
   bool modified = !context()->debugs1().empty() ||
                   !context()->debugs2().empty() ||
                   !context()->debugs3().empty();
+  for (auto& dbg : context()->debugs1())
+    context()->KillNamesAndDecorates(&dbg);
+
+  for (auto& dbg : context()->debugs2())
+    context()->KillNamesAndDecorates(&dbg);
+
+  for (auto& dbg : context()->debugs3())
+    context()->KillNamesAndDecorates(&dbg);
+
   context()->debug_clear();
 
-  context()->module()->ForEachInst([&modified](Instruction* inst) {
+  context()->module()->ForEachInst([&modified, this](Instruction* inst) {
     modified |= !inst->dbg_line_insts().empty();
+    for (auto& dbg : inst->dbg_line_insts())
+      this->context()->KillNamesAndDecorates(&dbg);
+
     inst->dbg_line_insts().clear();
   });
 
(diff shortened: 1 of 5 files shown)


zoddicus:
> You should also add a test for an OpName targeting an OpString.

Done

test/opt/strip_debug_info_test.cpp
```diff
@@ -74,6 +74,82 @@ TEST_F(StripLineDebugInfoTest, LineNoLine) {
...
+                "%2 = OpFunction %void None %5",
+                "%6 = OpLabel",
+                     "OpReturn",
+                     "OpFunctionEnd",
+      // clang-format on
+  };
+  SinglePassRunAndCheck<StripDebugInfoPass>(JoinAllInsts(input),
+                                            JoinAllInsts(output),
+                                            /* skip_nop = */ false,
+                                            /* do_validation */ false);
```
alan-baker: I would turn validation on if possible.

source/opt/strip_debug_info_pass.cpp
```diff
@@ -22,10 +22,19 @@ Pass::Status StripDebugInfoPass::Process() {
   bool modified = !context()->debugs1().empty() ||
                   !context()->debugs2().empty() ||
                   !context()->debugs3().empty();
+  for (auto& dbg : context()->debugs1()) context()->KillNamesAndDecorates(&dbg);
```
s-perron: I think it would be best to just call `KillInst`.  I know this pass does not keep any of the analyses up-to-date, but it is still the standard way of deleting an instruction.

source/opt/strip_debug_info_pass.cpp
```diff
@@ -22,10 +22,19 @@ Pass::Status StripDebugInfoPass::Process() {
...
+
+  for (auto& dbg : context()->debugs2()) context()->KillNamesAndDecorates(&dbg);
+
+  for (auto& dbg : context()->debugs3()) context()->KillNamesAndDecorates(&dbg);
+
   context()->debug_clear();
```
s-perron: If you call `KillInst` above, then this is not needed.  In fact, I would suggest removing `debug_clear` from the context and module classes.  It is a single purpose function and it does not keep the analyses valid, so they can be dangerous.

source/opt/strip_debug_info_pass.cpp
```diff
@@ -22,10 +22,19 @@ Pass::Status StripDebugInfoPass::Process() {
...
 
-  context()->module()->ForEachInst([&modified](Instruction* inst) {
+  context()->module()->ForEachInst([&modified, this](Instruction* inst) {
     modified |= !inst->dbg_line_insts().empty();
+    for (auto& dbg : inst->dbg_line_insts())
+      this->context()->KillNamesAndDecorates(&dbg);
```
s-perron: OpLine instructions do not have a result id, so cannot have names or decorations.

source/opt/strip_debug_info_pass.cpp
```diff
@@ -22,10 +22,19 @@ Pass::Status StripDebugInfoPass::Process() {
...
+
     inst->dbg_line_insts().clear();
```
s-perron: OpLine instruction are different.  Many of the analysies do not even consider the OpLine instructions.  For now, we can leave this clear.  Eventually we may change the way we handle debug info, then everything will have to change.

source/opt/strip_debug_info_pass.cpp
zoddicus: Done

source/opt/strip_debug_info_pass.cpp
zoddicus: Done

source/opt/strip_debug_info_pass.cpp
zoddicus: Done

source/opt/strip_debug_info_pass.cpp
zoddicus: Done

test/opt/strip_debug_info_test.cpp
zoddicus: Done

zoddicus:
I am unclear on why I am getting an OOB error in this test only on MSVC 2015 & 2017

s-perron:
I think I see why you get the failure on Windows, and it is a legit error.

The test StripDebugStringTest.OpNameRemoved has two instructions that need to be removed:

```
                "%3 = OpString \"minimal.vert\"",
                     "OpName %3 \"bob\"",
```

If you kill the `OpString` first, it will remove the name when you call `KillNameAndDecorates` in `KillInst`.  So the name is kill, but it will still be in the set of instructions to kill.  After killing the `OpString` instruction, it will try to kill the OpName instruction, which is a dangling pointer.

This error is not deterministic because you are using an unorder_set.  If you kill the OpName first, then there is no problem.

PR #2926, opened 2019-09-27
Record trailing line dbg instructions

s-perron:
There is nothing in the spir-v spec that says the last
instructions in a module cannot be OpLine or OpNoLine.
However, the code that parses the module will simply drop
these instructions.

We add code that will preserve these instructions.

Strip-debug-info is updated to remove these instructions.

Fixes https://crbug.com/1000689.


s-perron:
This should get rebased onto https://github.com/KhronosGroup/SPIRV-Tools/pull/2923 after it is merged.

Issue #1166, opened 2018-01-04
For loop incorrectly optimized away


antiagainst:
For the following HLSL code:
```hlsl
SamplerState gSS  :register(s2, space0);
Texture2D    gTex :register(t0, space0);

cbuffer MyCBuffer :register(b0, space2)
{
  float2 scale;
  float3 size;
};

float4 GetTexture2DLod(Texture2D tex, SamplerState samplerState, float2 uv, float lod)
{
  return tex.SampleLevel(samplerState, uv.xy, lod);
}

float4 PS_BlurFilter_Box1() : SV_Target
{
  float4 color = 0.0f;
  float i = 0.0f;

  for (i = -size.x; i <= size.x; i++)
    color += GetTexture2DLod(gTex, gSS, i * scale, 0.f);

  return color;
}
```
DXC/Spiregg generates the following raw SPIR-V:
```llvm
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %PS_BlurFilter_Box1 "PS_BlurFilter_Box1" %out_var_SV_Target
               OpExecutionMode %PS_BlurFilter_Box1 OriginUpperLeft
               OpName %bb_entry "bb.entry"
               OpName %for_check "for.check"
               OpName %for_body "for.body"
               OpName %for_continue "for.continue"
               OpName %for_merge "for.merge"
               OpName %bb_entry_0 "bb.entry"
               OpName %type_sampler "type.sampler"
               OpName %gSS "gSS"
               OpName %type_2d_image "type.2d.image"
               OpName %gTex "gTex"
               OpName %type_MyCBuffer "type.MyCBuffer"
               OpMemberName %type_MyCBuffer 0 "scale"
               OpMemberName %type_MyCBuffer 1 "size"
               OpName %var_MyCBuffer "var.MyCBuffer"
               OpName %src_PS_BlurFilter_Box1 "src.PS_BlurFilter_Box1"
               OpName %PS_BlurFilter_Box1 "PS_BlurFilter_Box1"
               OpName %out_var_SV_Target "out.var.SV_Target"
               OpName %color "color"
               OpName %i "i"
               OpName %param_var_tex "param.var.tex"
               OpName %param_var_samplerState "param.var.samplerState"
               OpName %param_var_uv "param.var.uv"
               OpName %param_var_lod "param.var.lod"
               OpName %GetTexture2DLod "GetTexture2DLod"
               OpName %tex "tex"
[...]

PR #2927, opened 2019-09-30
Enable OpTypeCooperativeMatrix specialization

jeremy-lunarg:
spirv-opt was failing during constant specialization of OpTypeCooperativeMatrix (see Vulkan-ValidationLayers #1166).

### DIFF
--- a/source/opt/types.cpp
+++ b/source/opt/types.cpp
@@ -127,6 +127,7 @@ std::unique_ptr<Type> Type::Clone() const {
     DeclareKindCase(PipeStorage);
     DeclareKindCase(NamedBarrier);
     DeclareKindCase(AccelerationStructureNV);
+    DeclareKindCase(CooperativeMatrixNV);
 #undef DeclareKindCase
     default:
       assert(false && "Unhandled type");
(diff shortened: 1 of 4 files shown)


jeremy-lunarg:
@alan-baker @s-perron Can you please review? Thanks.

source/opt/types.cpp
```diff
@@ -654,6 +657,41 @@ void ForwardPointer::GetExtraHashWords(
...
+std::string CooperativeMatrixNV::str() const {
+  std::ostringstream oss;
+  oss << "<" << component_type_->str() << ", " << scope_id_ << ", " << rows_id_
+      << ", " << columns_id_ << ">";
+  return oss.str();
+}
+
+void CooperativeMatrixNV::GetExtraHashWords(
+    std::vector<uint32_t>* words, std::unordered_set<const Type*>* pSet) const {
+  component_type_->GetHashWords(words, pSet);
```
s-perron: You probably want to include the scope, rows and columns in the hash words.

jeremy-lunarg: Ah yes. Thanks.

PR #2928, opened 2019-10-01
fuzz: add shrinker-temp-file-prefix

paulthomson:
And relax naming requirements of input and output files.

### DIFF
--- a/tools/fuzz/fuzz.cpp
+++ b/tools/fuzz/fuzz.cpp
@@ -110,6 +110,12 @@ Options (in lexicographical order):
                Unsigned 32-bit integer specifying maximum number of steps the
                shrinker will take before giving up.  Ignored unless --shrink
                is used.
+  --shrinker-temp-file-prefix
+               Specifies a temporary file prefix that will be used to output
+               temporary shader files during shrinking.  A number and .spv
+               extension will be added.  The default is "temp_", which will
+               cause files like "temp_0001.spv" to be output to the current
+               directory.  Ignored unless --shrink is used.
   --replay-validation
                Run the validator after applying each transformation during
                replay (including the replay that occurs during shrinking).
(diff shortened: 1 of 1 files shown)


tools/fuzz/fuzz.cpp
```diff
@@ -215,23 +219,11 @@ FuzzStatus ParseFlags(int argc, const char** argv, std::string* in_binary_file,
     return {FuzzActions::STOP, 1};
   }
```
afd: By removing the following, it looks like we make it possible to pass an input file that does not end in .spv.  Am I correct, and if so is it intentional?  The help for the tool says that the input file must be a .spv file.

paulthomson: Yes it is intentional. I'd like to pass a file with extension ".spv_orig". I will update the help text.

tools/fuzz/fuzz.cpp
```diff
@@ -75,9 +75,8 @@ USAGE: %s [options] <input.spv> -o <output.spv>
 USAGE: %s [options] <input.spv> -o <output.spv> \
   --shrink=<input.transformations> -- <interestingness_function> <args...>
```
afd: I think it would be clearer to say:

The SPIR-V binary is read from <input[.ext]>, e.g. 'shader.spv', or 'shader'.  If <input.facts> ('shader.facts' in both example cases) is also present, facts about the SPIR-V binary are read from this file.

But perhaps this is a bit too nit-picky.  What do you think?

paulthomson: I personally find it easier to use input.spv throughout as it gives me a reminder of what the input is. I think the use of the .spv extension should appear to be almost required, but we should support other extensions for convenience.

afd: That's fine by me.

PR #2931, opened 2019-10-03
Keep NOPs when comparing with original binary

s-perron:
We have a check that ensures that the optimizer did not change the
binary when it says that it did not.  However, when the binary is
converted back to a binary, we made a decision to remove OpNop
instructions.  This means that any spv file that contains a NOP
originally will fail this check.

To get around this, we convert the module to a second binary that keeps
the OpNop instructions.  That binary is compared against the original.

Fixes https://crbug.com/1010191

### DIFF
--- a/source/opt/optimizer.cpp
+++ b/source/opt/optimizer.cpp
@@ -565,8 +565,11 @@ bool Optimizer::Run(const uint32_t* original_binary,
 
 #ifndef NDEBUG
   if (status == opt::Pass::Status::SuccessWithoutChange) {
-    auto changed = optimized_binary->size() != original_binary_size ||
-                   memcmp(optimized_binary->data(), original_binary,
+    std::vector<uint32_t> optimized_binary_with_nop;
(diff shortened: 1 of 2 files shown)


dj2:
Why doesn't opt say it changed the binary when it removed NOPs since, the binary is changed?

s-perron:
Because the optimizer did not change the binary.  the NOPs are still there.  The NOPs are removed [here](https://github.com/KhronosGroup/SPIRV-Tools/pull/2931/files#diff-9c001dfbb726090f6c236ad2506c10efR564) when the module is converted back to a binary.

We could not skip the NOPs at that point.  It was originally added because the optimizer could not easily remove instructions, so they were all converted to NOPs instead.  Now that we delete instructions in the optimizer instead of converting them to NOPs, we do not need to remove them.  However, I did not want to change behaviour for this.

s-perron:
@dneto0 Could you take a look at this?

PR #2935, opened 2019-10-03
Generate null pointer by converting uint64 zero to pointer.

### LINKED ISSUE #2929: InstBuffAddrCheckPass should not use null physical storage buffer pointers
https://github.com/KhronosGroup/SPIRV-Registry/pull/51 updated the EXT version of the extensions and https://github.com/KhronosGroup/SPIRV-Registry/pull/50 updated the KHR version. Physical storage buffer pointers cannot be used with OpConstantNull (or OpPtrEqual, OpPtrNotEqual or OpPtrDiff).

The pass should instead convert the pointer before generating the OpPhi and use a 64-bit integer or uvec2 0 value.

@greg-lunarg, I am disabling validation in one of the tests until this is resolved.

greg-lunarg:
Fixes #2929.


greg-lunarg:
@s-perron ? @alan-baker ?

PR #2941, opened 2019-10-04
spirv-fuzz: add transformation and pass to construct composites

### LINKED ISSUE #2858: fuzz: use OpCompositeConstruct to create id synonyms
The fuzzer should find ids %a, %b, ... that are all available at a particular insertion point, and create:

%fresh = OpCompositeConstruct ... %a %b

then tell the fact manager that %a is synonymous with %fresh[0], %b with %fresh[1], etc.

These synonyms will then be exploitable by the pass that rewrites ids with known synonyms.

afd:
Adds a fuzzer pass and transformation to create a composite (array,
matrix, struct or vector) from available constituent components, and
inform the fact manager that each component of the new composite is
synonymous with the id that was used to construct it.  This allows the
"replace id with synonym" pass to then replace uses of said ids with
uses of elements extracted from the composite.

Fixes #2858.


source/fuzz/fuzzer_pass.cpp
```diff
@@ -27,5 +27,97 @@ FuzzerPass::FuzzerPass(opt::IRContext* ir_context, FactManager* fact_manager,
...
+        // Its return value informs us of how many instructions it inserted.
+        // (This will be 0 if no transformation was applied.)
+        uint32_t num_instructions_inserted =
+            maybe_apply_transformation(function, &block, inst_it, base, offset);
+
+        if (!inst_it->HasResultId()) {
+          // We are tracking the current id-less instruction via an offset,
+          // |offset|, from a previous instruction, |base|, that has an id. We
+          // increment |offset| to reflect any newly-inserted instructions.
+          offset += num_instructions_inserted;
```
paulthomson: Could add a comment explaining that, even though the inserted instructions might include one with a result id, you are not going to use that result id.

afd: Done.

PR #2945, opened 2019-10-04
reduce: improve remove unref instr pass

paulthomson:
* Remove Impl struct in Reducer; we can re-add it later (in a cleaner fashion) if we need to. 
* Add cleanup passes in Reducer; needed so that we removal of constants can be disabled during the main passes, and then enabled during cleanup passes, otherwise some main passes can perform worse due to lack of available constants. 
* Delete passes: remove op name, remove relaxed precision. And delete associated tests. 
* Add more tests for remove unreferenced instructions. 
* Always return and write the output file, even if there was a reduction failure. 
* Only exit with 0 if the reduction completed or we hit the reduction step limit.


source/reduce/reducer.h
```diff
@@ -86,17 +86,34 @@ class Reducer {
...
+  void AddCleanupReductionPass(
+      std::unique_ptr<ReductionOpportunityFinder>&& finder);
+
   // Reduces the given SPIR-V module |binary_out|.
   // The reduced binary ends up in |binary_out|.
   // A status is returned.
   ReductionResultStatus Run(std::vector<uint32_t>&& binary_in,
                             std::vector<uint32_t>* binary_out,
                             spv_const_reducer_options options,
-                            spv_validator_options validator_options) const;
```
afd: Is there a good reason to remove const here?

paulthomson: I believe it was necessary. This function now (sort of) modifies `passes_`.

afd: Thanks for confirming.

PR #2947, opened 2019-10-08
reduce: add large tests and fix

paulthomson:
* Add larger reducer tests.
* Fix conditional_branch_to_simple_conditional_branch_opportunity pass.


source/reduce/conditional_branch_to_simple_conditional_branch_reduction_opportunity.cpp
```diff
@@ -19,12 +19,15 @@
 namespace spvtools {
 namespace reduce {
 
+using opt::IRContext;
```
afd: In general my preference is (these days) to move away from "using", and instead fully qualify things; this is based on code review from @dj2.  No need to address for this PR necessarily but worth bearing in mind.

PR #2948, opened 2019-10-08
CMake: Add support for building with emscripten

Kangz:


Kangz:
Looking at the kokoro logs the issue isn't related to this patch. How do we move forward?

alan-baker:
Can you please rebase on master? That should pick up the bazel scripts.

PR #2949, opened 2019-10-08
spirv-fuzz: Add fuzzer pass to change loop controls

### LINKED ISSUE #2938: spirv-fuzz: Change loop control
The loop control operand of OpLoopMerge is a hint to the shader compiler; spirv-fuzz should try playing with this hint.

### LINKED ISSUE #2943: spirv-fuzz: Play with PeelCount, PartialCount, and other new loop selection control options
Peel and partial counts could be added, removed or changed, as they are just hints.

Controls such as MinIterations could be removed, as they communicate user-provided information that might help an optimizer, but cannot be added without special knowledge.

afd:
A new pass that allows the fuzzer to change the 'loop control' operand
(and associated literal operands) of OpLoopMerge instructions.
    
Fixes #2938.
Fixes #2943.

### DIFF
--- a/source/fuzz/fuzzer_pass_adjust_loop_controls.cpp
+++ b/source/fuzz/fuzzer_pass_adjust_loop_controls.cpp
@@ -39,36 +39,49 @@ void FuzzerPassAdjustLoopControls::Apply() {
           continue;
         }
 
-        assert(false);
+        if (!GetFuzzerContext()->ChoosePercentage(
+                GetFuzzerContext()->GetChanceOfAdjustingLoopControl())) {
+          continue;
+        }
 
-//        // Choose randomly whether to change the loop control for this
-//        // instruction.
-//        if (!GetFuzzerContext()->ChoosePercentage(
-//                GetFuzzerContext()->GetChanceOfAdjustingLoopControl())) {
-//          continue;
-//        }
-//
-//        // The choices to change the loop control to are the set of valid
-//        // controls, minus the current control.
-        std::vector<uint32_t> choices;
-//        for (auto control :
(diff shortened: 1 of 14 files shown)


source/fuzz/fuzzer_pass_adjust_loop_controls.cpp
```diff
@@ -0,0 +1,115 @@
...
+                                             SpvLoopControlUnrollMask,
+                                             SpvLoopControlDontUnrollMask};
+        uint32_t new_mask =
+            basic_masks[GetFuzzerContext()->RandomIndex(basic_masks)];
+
+        // For the loop controls that depend on guarantees about what the loop
+        // does, check which of these were present in the existing mask and
+        // randomly decide whether to keep them.  They are just hints, so
+        // removing them should not change the semantics of the module.
+        for (auto mask :
```
alan-baker: Some of these require SPIR-V 1.4, you should check the binary version before adding these.

afd: @alan-baker I thought about this, but the code only adds them if they were present already - and an assumption throughout spirv-fuzz is that the initial module was valid (so the tool needs to keep it valid, but can assume that it was already valid).  So on that basis I think the below is OK, do you agree?  (Worth adding a comment to explain the rationale, though, which I will do.)

alan-baker: Ok, I guess I didn't read closely enough. If they were already present that's ok.

afd: I think it's a point worth clarifying, so I have pushed a comment.

Issue #2952, opened 2019-10-09
ValidateExecutionScope is incorrect for OpReadClockKHR


alegal-arm:
spirv-val subjects OpReadClockKHR to generic Vulkan rules regarding execution scope, i.e. 
Scope for execution must be limited to:
◦ Workgroup
◦ Subgroup

However the spec actually says that 
Scope for OpReadClockKHR must be limited to:
◦ Subgroup ...
◦ Device ...

alegal-arm:
@AaronHaganAMD @alan-baker can you have a look?

alan-baker:
The currently published spec says:
> All invocations within the Execution read from the same source clock. Execution must be a valid execution scope.

And Vulkan limits execution scopes to workgroup and subgroup.

alegal-arm:
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/pdf/vkspec.pdf v 1.1.124
p. 1574
```
Scope for OpReadClockKHR must be limited to:
◦ Subgroup
▪ if shaderSubgroupClock is not enabled, the Subgroup scope must not be used.
◦ Device
▪ if shaderDeviceClock is not enabled, the Device scope must not be used
```

alan-baker:
The Vulkan spec is inconsistent then IMO and needs fixed. In the SPIR-V extension we intentionally made it an execution scope to fall in with what already existed.

alan-baker:
I will disable scope validation until this gets resolved

alan-baker:
Vulkan internal issue 1854

alegal-arm:
Just to give the full picture. glslang actually generates Device scope for OpReadClockKHR

https://github.com/KhronosGroup/glslang/issues/1923

PR #2953, opened 2019-10-09
Disable scope validation for OpReadClockKHR

alan-baker:
See #2952

Disabled until specification is clarified

### DIFF
--- a/source/val/validate_misc.cpp
+++ b/source/val/validate_misc.cpp
@@ -42,10 +42,13 @@ spv_result_t ValidateUndef(ValidationState_t& _, const Instruction* inst) {
 
 spv_result_t ValidateShaderClock(ValidationState_t& _,
                                  const Instruction* inst) {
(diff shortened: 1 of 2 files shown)


source/val/validate_misc.cpp
```diff
@@ -42,10 +42,10 @@ spv_result_t ValidateUndef(ValidationState_t& _, const Instruction* inst) {
 
 spv_result_t ValidateShaderClock(ValidationState_t& _,
                                  const Instruction* inst) {
-  const uint32_t execution_scope = inst->word(3);
-  if (auto error = ValidateExecutionScope(_, inst, execution_scope)) {
-    return error;
-  }
+  // const uint32_t execution_scope = inst->word(3);
```
dneto0: I'd prefer #if 0..#endif
But c'est la vie

alan-baker: I'll change it, no problem.

Issue #2557, opened 2019-05-07
Fold OpSpecConstantOp UConvert


alan-baker:
Update FoldSpecConstantOpAndCompositePass to support this operation.

alan-baker:
Constant folder only seems to handle 32-bit numbers currently.

s-perron:
Here is an example that is not currently handled.

```
               OpCapability Shader
               OpCapability Int64
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %1 "main"
               OpExecutionMode %1 OriginUpperLeft
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
       %uint = OpTypeInt 32 0
      %ulong = OpTypeInt 64 0
     %uint_1 = OpConstant %uint 1
          %7 = OpSpecConstantOp %ulong UConvert %uint_1
          %1 = OpFunction %void None %3
          %8 = OpLabel
               OpReturn
               OpFunctionEnd
```

PR #2960, opened 2019-10-10
Support constant-folding UConvert and SConvert

kuhar:
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/2557.

### DIFF
--- a/source/opt/fold_spec_constant_op_and_composite_pass.cpp
+++ b/source/opt/fold_spec_constant_op_and_composite_pass.cpp
@@ -316,6 +316,59 @@ bool IsValidTypeForComponentWiseOperation(const analysis::Type* type) {
   }
   return false;
 }
+
+// Encodes the integer |value| of in a word vector format appropriate for
+// representing this value as a operands for a constant definition. Performs
+// zero-extension/sign-extension/truncation when needed, based on the signess of
+// the given target type.
+//
+// Note: type |type| argument must be either Integer or Bool.
+utils::SmallVector<uint32_t, 2> EncodeIntegerAsWords(const analysis::Type& type,
+                                                     uint32_t value) {
+  const uint32_t all_ones = ~0;
+  uint32_t bit_width = 0;
+  uint32_t pad_value = 0;
+  bool result_type_signed = false;
+  if (auto* int_ty = type.AsInteger()) {
+    bit_width = int_ty->width();
+    result_type_signed = int_ty->IsSigned();
+    if (result_type_signed && static_cast<int32_t>(value) < 0) {
+      pad_value = all_ones;
+    }
+  } else if (type.AsBool()) {
+    bit_width = 1;
+  } else {
+    assert(false && "type must be Integer or Bool");
+  }
+
+  assert(bit_width > 0);
+  uint32_t first_word = value;
+  const uint32_t bits_per_word = 32;
+
+  // Truncate first_word if the |type| has width less than uint32.
+  if (bit_width < bits_per_word) {
+    const uint32_t num_high_bits_to_mask = bits_per_word - bit_width;
+    const bool is_negative_after_truncation =
+        result_type_signed &&
+        utils::IsBitAtPositionSet(first_word, bit_width - 1);
+
+    if (is_negative_after_truncation) {
+      // Truncate and sign-extend |first_word|. No padding words will be
+      // added and |pad_value| can be left as-is.
+      first_word = utils::SetHighBits(first_word, num_high_bits_to_mask);
+    } else {
+      first_word = utils::ClearHighBits(first_word, num_high_bits_to_mask);
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


test/opt/fold_spec_const_op_composite_test.cpp
```diff
@@ -851,7 +941,7 @@ INSTANTIATE_TEST_SUITE_P(
               {
                 "%int_n1 = OpConstant %int -1",
```
s-perron: Since this is no longer used, it should be removed.

source/opt/fold_spec_constant_op_and_composite_pass.cpp
```diff
@@ -316,6 +316,31 @@ bool IsValidTypeForComponentWiseOperation(const analysis::Type* type) {
   }
   return false;
 }
+
+utils::SmallVector<uint32_t, 2> EncodeIntegerAsWords(const analysis::Type& type,
```
s-perron: Add a comment explaining what the function does, and any assumptions it makes .  For example, what happens if the type of a 64-bit floating point type?

test/opt/fold_spec_const_op_composite_test.cpp
kuhar: This is expected code, and it does appear in the output.

kuhar:
Thanks for the comments, @s-perron. I forgot about the truncation case.

This revision should handle all of zext/sext/truncate.

source/opt/fold_spec_constant_op_and_composite_pass.cpp
```diff
@@ -316,6 +316,50 @@ bool IsValidTypeForComponentWiseOperation(const analysis::Type* type) {
...
+
+  assert(bit_width > 0);
+  uint32_t first_word = value;
+  const uint32_t bits_per_word = 32;
+
+  // Truncate first_word if the |type| has width less than uint32.
+  if (bit_width < bits_per_word) {
+    const uint32_t num_high_bits_to_mask = bits_per_word - bit_width;
+    const uint32_t all_ones = ~0;
+    // Set the most significant to zeros.
```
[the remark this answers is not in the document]
s-perron: Also the bits should be sign extended if it is a say 16-bit signed value.  See https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#Literal_Number for encodings of liter numbers.

kuhar:
Sign-extension after truncation is now handled. I checked and this should is in-line with the runtime behavior of OpSConvert.
@s-perron

source/opt/fold_spec_constant_op_and_composite_pass.cpp
```diff
@@ -316,6 +316,66 @@ bool IsValidTypeForComponentWiseOperation(const analysis::Type* type) {
...
+    assert(false && "type must be Integer or Bool");
+  }
+
+  assert(bit_width > 0);
+  uint32_t first_word = value;
+  const uint32_t bits_per_word = 32;
+
+  // Truncate first_word if the |type| has width less than uint32.
+  if (bit_width < bits_per_word) {
+    const uint32_t num_high_bits_to_mask = bits_per_word - bit_width;
```
s-perron: This section is a little confusing.  It might be clearer if you just has some utility functions.  You also so extra work.  For example, you clear the bits even if you will later set all of them.  Another is fall through and doing the loop at the end.  I would try to get this down to something like:

```
if (result_type_signed && IsNegativeValue(value, bit_width)) { return { SetBits(value, 32 - bit_width); } }
else { return { ClearBits(value, 32 - bit_width); }}
```

I removes the need for many of the comments because the function names tell us what is happening.

source/opt/fold_spec_constant_op_and_composite_pass.cpp
```diff
@@ -316,6 +316,66 @@ bool IsValidTypeForComponentWiseOperation(const analysis::Type* type) {
   }
   return false;
 }
+
+// Encodes the integer |value| of in a word vector format appropriate for
+// representing this value as a operands for a constant definition. Performs
+// zero-extension/sign-extension/truncation when needed, based on the signess of
+// the given target type.
+//
+// Note: type |type| argument must be either Integer or Bool.
```
s-perron: Should document the bit widths that are handled as well.

source/opt/fold_spec_constant_op_and_composite_pass.cpp
kuhar: Yeah, I don't like bit manipulations either. Does it make more sense add these helper functions to this file or to `source/util/bitutils.h`? Not sure whether other (/future) code will need them.

s-perron: Yes, bitutil is a good place for them.

kuhar:
Moved the bit manipulation code to `bitutils` and added unit tests.

PR #2961, opened 2019-10-11
spirv-fuzz: Refactor 'split blocks' to identify instructions differently

afd:
This change refactors the 'split blocks' transformation so that an
instruction is identified via a base, opcode, and number of those
opcodes to be skipped when searching from the base, as opposed to the
previous design which used a base and offset.

### DIFF
--- a/source/fuzz/fuzzer_pass_split_blocks.cpp
+++ b/source/fuzz/fuzzer_pass_split_blocks.cpp
@@ -14,7 +14,6 @@
 
 #include "source/fuzz/fuzzer_pass_split_blocks.h"
 
-#include <tuple>
 #include <vector>
 
 #include "source/fuzz/instruction_descriptor.h"
@@ -50,15 +49,16 @@ void FuzzerPassSplitBlocks::Apply() {
       // We are not going to try to split this block.
       continue;
     }
+
+    // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/2964): consider
+    //  taking a simpler approach to identifying the instruction before which
+    //  to split a block.
+
     // We are going to try to split this block.  We now need to choose where
     // to split it.  We describe the instruction before which we would like to
-    // split a block via the opcode 'opc' of the relevant instruction, a base
-    // instruction 'base' that has a result id, and the number of instructions
-    // with opcode 'opc' that we should skip when searching from 'base' for the
-    // desired instruction.  (When the instruction before which we would like to
-    // split the block actually has a result id, the instruction is used as
-    // base, 'opc' is its opcode, and there are 0 instructions to skip.)
-    std::vector<std::tuple<uint32_t, SpvOp, uint32_t>> base_opcode_skip_triples;
+    // split a block via an InstructionDescriptor, details of which are
+    // commented in the protobufs definition file.
+    std::vector<protobufs::InstructionDescriptor> instruction_descriptors;
 
     // The initial base instruction is the block label.
     uint32_t base = block->id();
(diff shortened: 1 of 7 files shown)


source/fuzz/fuzzer_pass_split_blocks.cpp
```diff
@@ -50,39 +51,51 @@ void FuzzerPassSplitBlocks::Apply() {
...
+    // base, 'opc' is its opcode, and there are 0 instructions to skip.)
+    std::vector<std::tuple<uint32_t, SpvOp, uint32_t>> base_opcode_skip_triples;
+
     // The initial base instruction is the block label.
     uint32_t base = block->id();
-    uint32_t offset = 0;
+
+    // Counts the number of times we have seen each opcode since we reset the
+    // base instruction.
+    std::map<SpvOp, uint32_t> skip_count;
```
paulthomson: As discussed offline, I would first find the instruction you want and then use a utility function that works backwards to find the nearest result id. This way, we don't need to interleave the logic for finding a suitable instruction and the logic that keeps track of the nearest result id. Plus, the utility function can be trivially re-used.

source/fuzz/fuzzer_pass_split_blocks.cpp
```diff
@@ -50,39 +51,51 @@ void FuzzerPassSplitBlocks::Apply() {
...
-    // OpStore instructions cannot).
-    std::vector<std::pair<uint32_t, uint32_t>> base_offset_pairs;
+    // to split it.  We describe the instruction before which we would like to
+    // split a block via the opcode 'opc' of the relevant instruction, a base
+    // instruction 'base' that has a result id, and the number of instructions
+    // with opcode 'opc' that we should skip when searching from 'base' for the
+    // desired instruction.  (When the instruction before which we would like to
+    // split the block actually has a result id, the instruction is used as
+    // base, 'opc' is its opcode, and there are 0 instructions to skip.)
+    std::vector<std::tuple<uint32_t, SpvOp, uint32_t>> base_opcode_skip_triples;
```
paulthomson: If we plan to change this later anyway, then never mind, But otherwise, could we just use `InstructionDescriptor`s here.

source/fuzz/transformation_split_block.cpp
```diff
@@ -40,31 +40,28 @@ bool TransformationSplitBlock::IsApplicable(
     // We require the id for the new block to be unused.
     return false;
   }
-  auto base_instruction =
-      context->get_def_use_mgr()->GetDef(message_.base_instruction_id());
-  if (!base_instruction) {
+  auto instruction_to_split_before =
+      FindInstruction(message_.instruction_to_split_before(), context);
```
paulthomson: Perhaps `FindInstruction` should return an instruction iterator for maximum flexibility.

source/fuzz/fuzzer_pass_split_blocks.cpp
afd: I have opened https://github.com/KhronosGroup/SPIRV-Tools/issues/2964 to track this - I agree we should do this a simpler way (and quite possibly the way you propose), but indeed in a future PR.

source/fuzz/fuzzer_pass_split_blocks.cpp
afd: Wow, of course!!

source/fuzz/transformation_split_block.cpp
afd: Unfortunately this won't work because there may not be a basic block associated with the instruction (if the instruction does not exist).

PR #2962, opened 2019-10-11
Validate that selections are structured

alan-baker:
Validate that all selections are structured (i.e. are preceeded by a merge instruction) where necessary.

Checks that switches and conditional branches are preceeded by a merge instruction if two or more  unique targets have not yet been encountered in a reverse post order traversal. This intentional does not flag code like:
```
%b0 = OpLabel
OpSelectionMerge %b3 None
OpBranchConditional %c1 %b1 %b2
%b1 = OpLabel
OpBranchConditional %c2 %b2 %b3 ; not an error
%b2 = OpLabel
OpBranch %b3
%b3 = OpLabel
```

Updated a few offending unit tests. Tested against CTS.

### DIFF
--- a/test/val/val_cfg_test.cpp
+++ b/test/val/val_cfg_test.cpp
@@ -3750,7 +3750,149 @@ OpFunctionEnd
   EXPECT_EQ(SPV_SUCCESS, ValidateInstructions());
 }
 
-/// TODO(umar): Nested CFG constructs
+TEST_F(ValidateCFG, MissingMergeConditionalBranchBad) {
+  const std::string text = R"(
+OpCapability Shader
+OpCapability Linkage
+OpMemoryModel Logical GLSL450
+%void = OpTypeVoid
+%void_fn = OpTypeFunction %void
+%bool = OpTypeBool
+%undef = OpUndef %bool
+%func = OpFunction %void None %void_fn
+%entry = OpLabel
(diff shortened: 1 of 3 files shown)


test/val/val_cfg_test.cpp
```diff
@@ -3026,6 +3026,7 @@ OpMemoryModel Logical GLSL450
 %undef = OpUndef %bool
 %func = OpFunction %void None %void_fn
 %entry = OpLabel
+OpSelectionMerge %block None
```
dneto0: sweet

dneto0:
FYI, for this case that is still permitted:

```
%b0 = OpLabel
OpSelectionMerge %b3 None
OpBranchConditional %c1 %b1 %b2
%b1 = OpLabel
OpBranchConditional %c2 %b2 %b3 ; not an error
%b2 = OpLabel
OpBranch %b3
%b3 = OpLabel
```
consider this translation to high level language, without duplicating blocks.

```
b0;
if (c1) { b1; }
if ( (c1&&c2) || (!c1) ) { b2; }
b3;
```
That is, the CFG is a DAG, so emit in topological order, but each block X is predicated by the union of conditions implied by the paths to that block X.  (Imagine you can hoist the . conditions, which is a matter of bookkeeping....)

PR #2963, opened 2019-10-14
CMake fixes.

akb825:
Only search for Python when tests are enabled. This avoids a requirement
for Python 3 when tests aren't needed.

Added exports for libraries. External libraries that themselves use
libraries require all dependencies have exports, so not having exports can
cause major problems when used within other projects.

Install paths for exports are now placed in the proper directories expected
by Windows and *nix systems.


CMakeLists.txt
```diff
@@ -185,12 +185,14 @@ if(NOT COMMAND find_host_program)
 endif()
 
 # Tests require Python3
-find_host_package(PythonInterp 3 REQUIRED)
+if(NOT SPIRV_SKIP_TESTS)
```
s-perron: Python is used to build the grammar tables.  The reason this works even if python is used is because `${PYTHON_EXECUTABLE}` expands to the empty string, and the python scripts in `utils` are all have execute permission.  When they run, they just look for `python` on your path.

We should be searching for python in all cases.  If there is a fix needed, it is to change the permission on those files.

CMakeLists.txt
```diff
@@ -185,12 +185,14 @@ if(NOT COMMAND find_host_program)
...
+  find_host_package(PythonInterp 3 REQUIRED)
+endif()
 
 # Check for symbol exports on Linux.
 # At the moment, this check will fail on the OSX build machines for the Android NDK.
 # It appears they don't have objdump.
-if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
+if(NOT SPIRV_SKIP_TESTS AND "${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
```
s-perron: I'm not sure what you are trying to accomplish this this change.  If tests are being skipped, then the macro `spvtools_check_symbol_exports` will do nothing in both cases.

CMakeLists.txt
akb825: This originally came up because I tried to get it running on an automated build system where it didn't have Python 3 available and it appeared to not be required. I missed the requirement for the tables, and it must have just worked for Python 2.

I'll revert these changes and just have the export/install fixes then.

CMakeLists.txt
akb825: You're right, I just didn't look at that close enough. I think at some point I saw an error that prompted me to make that change, but it may have been caused by something unrelated.

akb825:
I've updated my patch and made a couple other fixes to make sure it works properly with find_package().

s-perron:
@akb825 The spir-v header have been updated.  Could you please rebase your change on top of the latest master.  This will allow the tests to pass.

PR #2968, opened 2019-10-16
spirv-fuzz: fuzzer pass to adjust memory access operands

### LINKED ISSUE #2940: spirv-fuzz: add memory access operand
It is always harmless, semantics-wise, to apply the "Volatile" or "Nontemporal" memory access operands to instructions that manipulate memory (e.g. OpLoad).

"Volatile" should not be removed - it could be needed.

"Nontemporal" could be removed - it's just a hint.

The fuzzer should play with these.

afd:
A new pass that gives spirv-fuzz the ability to adjust the memory
operand masks associated with memory access instructions (such as
OpLoad and OpCopy Memory).

Fixes #2940.

### DIFF
--- a/source/fuzz/fuzzer_pass_adjust_memory_operands_masks.cpp
+++ b/source/fuzz/fuzzer_pass_adjust_memory_operands_masks.cpp
@@ -1,0 +1,113 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_adjust_memory_operands_masks.h"
+
+#include "source/fuzz/instruction_descriptor.h"
+#include "source/fuzz/transformation_set_memory_operands_mask.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassAdjustMemoryOperandsMasks::FuzzerPassAdjustMemoryOperandsMasks(
+    opt::IRContext* ir_context, FactManager* fact_manager,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, fact_manager, fuzzer_context, transformations) {}
+
+FuzzerPassAdjustMemoryOperandsMasks::~FuzzerPassAdjustMemoryOperandsMasks() =
+    default;
+
+void FuzzerPassAdjustMemoryOperandsMasks::Apply() {
+  // Consider every block in every function.
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      // Consider every instruction in this block, using an explicit iterator so
+      // that when we find an instruction of interest we can search backwards to
+      // create an id descriptor for it.
+      for (auto inst_it = block.cbegin(); inst_it != block.cend(); ++inst_it) {
+        if (!TransformationSetMemoryOperandsMask::IsMemoryAccess(*inst_it)) {
+          // We are only interested in memory access instructions.
+          continue;
+        }
+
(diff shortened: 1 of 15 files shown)


source/fuzz/fuzzer_pass_adjust_memory_operands_masks.cpp
```diff
@@ -0,0 +1,113 @@
...
+          bool toggle_nontemporal =
+              !add_volatile || GetFuzzerContext()->ChooseEven();
+
+          // These bitwise operations use '|' to add Volatile if desired, and
+          // '^' to toggle Nontemporal if desired.
+          uint32_t new_mask =
+              (existing_mask | (add_volatile ? SpvMemoryAccessVolatileMask
+                                             : SpvMemoryAccessMaskNone)) ^
+              (toggle_nontemporal ? SpvMemoryAccessNontemporalMask
+                                  : SpvMemoryAccessMaskNone);
```
paulthomson: I suppose you could extract this into a public static method and then add a test for it?

source/fuzz/instruction_descriptor.cpp
```diff
@@ -66,5 +66,35 @@ protobufs::InstructionDescriptor MakeInstructionDescriptor(
...
+       backwards_iterator != block.begin(); --backwards_iterator) {
+    if (backwards_iterator->HasResultId()) {
+      // As soon as we find an instruction with a result id, we can return a
+      // descriptor for |inst_it|.
+      return MakeInstructionDescriptor(backwards_iterator->result_id(), opcode,
+                                       skip_count);
+    }
+    if (backwards_iterator != inst_it &&
+        backwards_iterator->opcode() == opcode) {
+      // We are skipping over an with the same opcode as |inst_it|; we increase
```
paulthomson: typo: an instruction

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -464,6 +465,25 @@ message TransformationSetLoopControl {
...
+message TransformationSetMemoryOperandsMask {
+
+  // A transformation that sets the memory operands mask of a memory access
+  // instruction.
+
+  // A descriptor for a memory access instruction, e.g. an OpLoad
+  InstructionDescriptor memory_access_instruction = 1;
+
+  // A mask of memory operands to be applied to the instruction.  It must be the
+  // same as the original mask, except that Volatile can be removed, and
```
paulthomson: Volatile can be added

source/fuzz/transformation_set_memory_operands_mask.cpp
```diff
@@ -0,0 +1,195 @@
...
+  }
+  if (!IsMemoryAccess(*instruction)) {
+    return false;
+  }
+
+  if (message_.memory_operands_mask_index() != 0) {
+    assert(message_.memory_operands_mask_index() == 1);
+    assert(instruction->opcode() == SpvOpCopyMemory ||
+           instruction->opcode() == SpvOpCopyMemorySized);
+    assert(MultipleMemoryOperandMasksAreSupported(context));
```
paulthomson: I wonder if these should be more than just debug assertions. Do we assume we won't accidentally find a different instruction?

source/fuzz/transformation_set_memory_operands_mask.cpp
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  auto original_mask_in_operand_index = GetInOperandIndexForMask(
+      *instruction, message_.memory_operands_mask_index());
+  assert(original_mask_in_operand_index != 0 &&
+         "The given mask index is not valid.");
```
paulthomson: Is this right?

source/fuzz/instruction_descriptor.cpp
afd: Fixed.

source/fuzz/protobufs/spvtoolsfuzz.proto
afd: Good catch.

source/fuzz/transformation_set_memory_operands_mask.cpp
afd: They really should never be violated.  I have moved them to the top of the method and made the accesses to 'opcode' refer to the message's instruction descriptor, and added a comment to explain.

source/fuzz/fuzzer_pass_adjust_memory_operands_masks.cpp
afd: I will leave this as is, since the code involves choosing randomly.

PR #2979, opened 2019-10-16
Check binary->code existence before destroying

### LINKED ISSUE #2975: Need to check that code is set before freeing in spvBinaryDestroy
New fuzzer for spirv-as I am working on found this.

zoddicus:
Fixes #2975

### DIFF
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -781,7 +781,7 @@ spv_result_t spvBinaryParse(const spv_const_context context, void* user_data,
 // TODO(dneto): This probably belongs in text.cpp since that's the only place
 // that a spv_binary_t value is created.
 void spvBinaryDestroy(spv_binary binary) {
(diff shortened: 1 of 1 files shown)


source/binary.cpp
```diff
@@ -781,9 +781,10 @@ spv_result_t spvBinaryParse(const spv_const_context context, void* user_data,
 // TODO(dneto): This probably belongs in text.cpp since that's the only place
 // that a spv_binary_t value is created.
 void spvBinaryDestroy(spv_binary binary) {
-  if (!binary) return;
-  delete[] binary->code;
-  delete binary;
+  if (!binary) {
```
alan-baker: This is now inverted and only deletes null pointers.

zoddicus: Thanks for the catch

PR #2983, opened 2019-10-17
Check that derivatives operate on 32-bit values

alan-baker:
* Add a check that derivative functions only operate on scalar or vector
32-bit floating point values
* Added tests to disallow half derivatives

### DIFF
--- a/test/val/val_derivatives_test.cpp
+++ b/test/val/val_derivatives_test.cpp
@@ -173,7 +173,7 @@ TEST_P(ValidateHalfDerivatives, ScalarFailure) {
               HasSubstr("Result type component width must be 32 bits"));
 }
 
-TEST_P(ValidateHalfDerivatives, VectorSuccess) {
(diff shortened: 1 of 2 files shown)


test/val/val_derivatives_test.cpp
```diff
@@ -150,6 +160,36 @@ TEST_F(ValidateDerivatives, OpDPdxWrongExecutionModel) {
...
+  const std::string body = "%val = " + op + " %f16 %f16_0\n";
+
+  CompileSuccessfully(
+      GenerateShaderCode(body, "OpCapability Float16\n").c_str());
+  ASSERT_EQ(SPV_ERROR_INVALID_DATA, ValidateInstructions());
+  EXPECT_THAT(getDiagnosticString(),
+              HasSubstr("Result type component width must be 32 bits"));
+}
+
+TEST_P(ValidateHalfDerivatives, VectorSuccess) {
```
s-perron: I'm guessing this should be `VectorFailure` and not `VectorSuccess`.

PR #2988, opened 2019-10-23
spirv-fuzz: add class to represent equivalence relation

afd:
Adds a templated class for representing an equivalence relation on a
value data type.  This will be used by spirv-fuzz for representing
sets of distinct pieces of data in a shader that are known to have
equal values.

### DIFF
--- a/source/fuzz/equivalence_relation.h
+++ b/source/fuzz/equivalence_relation.h
@@ -31,6 +31,41 @@ namespace fuzz {
 // constructor, and the |PointerHashT| and |PointerEqualsT| must be functors
 // providing hashing and equality testing functionality for pointers to objects
 // of type |T|.
+//
+// A disjoint-set (a.k.a. union-find or merge-find) data structure is used to
+// represent the equivalence relation.  Path compression is used.  Union by
+// rank/size is not used.
+//
+// Each disjoint set is represented as a tree, rooted at the representative
+// of the set.
+//
+// Getting the representative of a value simply requires chasing parent pointers
+// from the value until you reach the root.
+//
+// Checking equivalence of two elements requires checking that the
+// representatives are equal.
+//
+// Traversing the tree rooted at a value's representative visits the value's
+// equivalence class.
+//
+// |PointerHashT| and |PointerEqualsT| are used to define *equality* between
+// values, and otherwise are *not* used to define the equivalence relation
+// (except that equal values are equivalent).  The equivalence relation is
+// constructed by repeatedly adding pairs of (typically non-equal) values that
+// are deemed to be equivalent.
+//
+// For example in an equivalence relation on integers, 1 and 5 might be added
+// as equivalent, so that IsEquivalent(1, 5) holds, because they represent
+// IDs in a SPIR-V binary that are known to contain the same value at run time,
+// but clearly 1 != 5.  Since 1 and 1 are equal, IsEquivalent(1, 1) will also
+// hold.
+//
+// Each unique (up to equality) value added to the relation is copied into
+// |owned_values_|, so there is one canonical memory address per unique value.
+// Uniqueness is ensured by storing (and checking) a set of pointers to these
+// values in |value_set_|, which uses |PointerHashT| and |PointerEqualsT|.
+//
+// |parent_| and |children_| encode the equivalence relation, i.e., the trees.
 template <typename T, typename PointerHashT, typename PointerEqualsT>
 class EquivalenceRelation {
  public:
--- a/source/fuzz/equivalence_relation.h
+++ b/source/fuzz/equivalence_relation.h
@@ -27,7 +27,7 @@ namespace fuzz {
 
 // A class for representing an equivalence relation on objects of type |T|,
 // which should be a value type.  The type |T| is required to have a copy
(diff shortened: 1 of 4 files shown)


source/fuzz/equivalence_relation.h
```diff
@@ -0,0 +1,181 @@
...
+
+namespace spvtools {
+namespace fuzz {
+
+// A class for representing an equivalence relation on objects of type |T|,
+// which should be a value type.  The type |T| is required to have a copy
+// constructor, and the |PointerHashT| and |PointerEqualsT| must be functors
+// providing hashing and equality testing functionality for pointers to objects
+// of type |T|.
+template <typename T, typename PointerHashT, typename PointerEqualsT>
```
paulthomson: Somewhere (maybe not here) it would be great to have a comment (or comments) describing a few more details, including the implementation a bit. It is a disjoint-set (a.k.a. union-find, merge-find) data structure. There are no weight/rank/size values. There is path compression. The disjoint sets are represented as trees, where the root is the representative of the set. [Maybe stop here?] Getting the representative element simply requires following the parent pointer of an element until you reach the root. Checking equivalence of two elements requires checking that the representatives are equal.

This paragraph is perhaps more for the user of the class: PointerHashT and PointerEqualsT are used to detect trivial equality between elements and otherwise are NOT used to define the equivalence relation (although elements that are trivially equal and are in the EquivalenceRelation WILL be deemed equivalent). Perhaps this is obvious. The equivalence relation is given by manually adding pairs of T that are deemed to be equivalent (and are most likely NOT trivially equal, although they can be). For example, 1 and 5 might be added as equivalent, so that IsEquivalent(1,5) holds, because they represent IDs in a SPIR-V binary that contain the same value at run time, but clearly 1 != 5. Yet, 1 and 1 are trivially equal, so IsEquivalent(1,1) will also hold.

Each unique (in terms of equality) value added is copied into `owned_values_`, so there is one canonical memory address per unique value. Uniqueness is ensured by storing (and checking) a set of pointers to these values in `value_set_`, which uses `PointerHashT` and `PointerEqualsT`. `parent_` and `children_` encode the equivalence relation, i.e., the trees.

afd: Done, thanks!

test/fuzz/equivalence_relation_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+  }
+};
+
+struct UInt32Hash {
+  size_t operator()(const uint32_t* element) const {
+    return static_cast<size_t>(*element);
+  }
+};
+
+std::set<uint32_t> ChasePointers(
```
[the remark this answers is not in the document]
afd: Done.  (ToUIntSet).

source/fuzz/equivalence_relation.h
```diff
@@ -0,0 +1,181 @@
...
+  // compression.
+  mutable std::map<const T*, const T*> parent_;
+
+  // Stores the children of each value.  This allows the equivalence class of
+  // a value to be calculated by traversing all descendents of the class's
+  // representative.
+  //
+  // Mutable because the intuitively const method, 'Find', performs path
+  // compression.
+  mutable std::map<const T*, std::set<const T*>> children_;
```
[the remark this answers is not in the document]
afd: Done.

test/fuzz/equivalence_relation_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+  relation.MakeEquivalent(78, 80);
+
+  std::set<uint32_t> class1;
+  for (uint32_t element = 0; element < 98; element += 2) {
+    ASSERT_TRUE(relation.IsEquivalent(0, element));
+    ASSERT_TRUE(relation.IsEquivalent(element, element + 2));
+    class1.insert(element);
+  }
+  class1.insert(98);
+  ASSERT_TRUE(class1 == ToUIntSet(relation.GetEquivalenceClass(0)));
```
s-perron: There are better ways of checking if two containers are the same.  See https://github.com/google/googletest/blob/master/googlemock/docs/cheat_sheet.md#container-matchers.

test/fuzz/equivalence_relation_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+  size_t operator()(const uint32_t* element) const {
+    return static_cast<size_t>(*element);
+  }
+};
+
+std::set<uint32_t> ToUIntSet(
+    EquivalenceRelation<uint32_t, UInt32Hash, UInt32Equals>::ValueSet
+        pointers) {
+  std::set<uint32_t> result;
+  for (auto pointer : pointers) {
```
s-perron: Traversing any set or unordered set where the key is a pointer is non-deterministic.  It does not seem to be an issue in this case.

test/fuzz/equivalence_relation_test.cpp
afd: Thanks - I have incorporated some ideas from here in #3006.

test/fuzz/equivalence_relation_test.cpp
afd: In #3006 I have changed the interface to EquivalenceRelation so that it does not expose unordered sets.

PR #2989, opened 2019-10-25
spirv-fuzz: Rework management of data synonyms

afd:
This change uses the recently-added equivalence relation class to
re-work the way synonyms between data values are managed by the fact
manager.

The tests for 'transformation_replace_id_with_synonym' have been
temporarily removed.  This is because those tests are going to be
split into a number of test classes in an upcoming PR, once some other
refactorings have been applied, and it would be burdensome to
temporarily refactor all the tests to be in a working state for this
intermediate change.


source/fuzz/fact_manager.h
```diff
@@ -52,6 +53,10 @@ class FactManager {
   // fact manager.
   bool AddFact(const protobufs::Fact& fact, opt::IRContext* context);
 
+  // TODO comment
```
paulthomson: Comment :)

paulthomson: Comment

afd: Fixed.

PR #2991, opened 2019-10-25
spirv-fuzz: Transformation to extract from a composite object

afd:
At present, TransformationReplaceIdWithSynonym both extracts elements
from composite objects and replaces uses of ids with synonyms.  This
new TransformationCompositeExtract class will allow that
transformation to be broken into smaller transformations.

### DIFF
--- a/source/fuzz/transformation_composite_extract.cpp
+++ b/source/fuzz/transformation_composite_extract.cpp
@@ -76,7 +76,7 @@ bool TransformationCompositeExtract::IsApplicable(
     return false;
   }
 
-  return fuzzerutil::WalkCompositeIndices(
+  return fuzzerutil::WalkCompositeTypeIndices(
              context, composite_instruction->type_id(), message_.index()) != 0;
 }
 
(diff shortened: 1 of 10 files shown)


source/fuzz/transformation_composite_extract.cpp
```diff
@@ -0,0 +1,123 @@
...
+  if (!composite_type) {
+    return false;
+  }
+
+  if (!fuzzerutil::CanInsertOpcodeBeforeInstruction(
+          SpvOpCompositeExtract, instruction_to_insert_before)) {
+    return false;
+  }
+
+  return fuzzerutil::WalkCompositeIndices(
```
paulthomson: Maybe we should rename this to something like: `WalkCompositeType`, `WalkCompositeTypeViaIndices`, `WalkCompositeTypeIndices`, `GetCompositePartType`, `CompositePartType`, `CompositeExtractType`, ...

afd: Done (WalkCompositeTypeIndices).

PR #2996, opened 2019-10-29
spirv-fuzz: transformation to insert OpVectorShuffle instructions

afd:
A new transformation that can insert an OpVectorShuffle instruction
into the module, and generate facts relating the result id of the
shuffle instruction to the vectors and indices being shuffled.


afd:
Abandoning this due to a re-think about how to handle management of data synonym facts.

PR #2998, opened 2019-10-29
Fix some clang-tidy issues in graphics_robust_access_pass

dneto0:
One remains: the fact that the image-texel-pointer modification
is mostly dead code. But that's intentional for now.

### DIFF
--- a/source/opt/graphics_robust_access_pass.h
+++ b/source/opt/graphics_robust_access_pass.h
@@ -81,8 +81,8 @@ class GraphicsRobustAccessPass : public Pass {
   // sign extension or zero extension.  The new instruction is inserted
   // immediately before |before_inst|, and is analyzed for definitions and uses.
   // Returns the newly inserted instruction.  Assumes the |value| is an integer
-  // scalar of a narrower type than |bitwidth| bits.
-  Instruction* WidenInteger(bool sign_extend, uint32_t bitwidth,
+  // scalar of a narrower type than |bit_width| bits.
+  Instruction* WidenInteger(bool sign_extend, uint32_t bit_width,
                             Instruction* value, Instruction* before_inst);
 
   // Returns a new instruction that invokes the UClamp GLSL.std.450 extended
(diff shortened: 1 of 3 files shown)


dneto0:
Sent to dnovillo@ because he told me about the clang-tidy complaints.

source/opt/graphics_robust_access_pass.h
```diff
@@ -81,8 +81,8 @@ class GraphicsRobustAccessPass : public Pass {
   // sign extension or zero extension.  The new instruction is inserted
   // immediately before |before_inst|, and is analyzed for definitions and uses.
   // Returns the newly inserted instruction.  Assumes the |value| is an integer
-  // scalar of a narrower type than |bitwidth| bits.
-  Instruction* WidenInteger(bool sign_extend, uint32_t bitwidth,
+  // scalar of a narrower type than |bit_width| bits.
+  Instruction* WidenInteger(bool sign_extend, uint32_t bit_width,
```
dneto0: Make the parameter name match the one used on the definition of this method.

source/opt/graphics_robust_access_pass.h
```diff
@@ -109,7 +109,8 @@ class GraphicsRobustAccessPass : public Pass {
   // the module is modified.  Returns a status code to indicate success
   // or failure.  If assumptions are not met, returns an error status code
   // and emits a diagnostic.
-  spv_result_t ClampCoordinateForImageTexelPointer(opt::Instruction* itp);
+  spv_result_t ClampCoordinateForImageTexelPointer(
+      opt::Instruction* image_texel_pointer);
```
dneto0: Make the parameter name match the one used on the definition of this method.

test/opt/graphics_robust_access_test.cpp
```diff
@@ -199,23 +199,23 @@ std::string ACCheck(const std::string& access_chain_inst,
                     const std::string& original,
                     const std::string& transformed) {
   return "\n ; CHECK: %ac = " + access_chain_inst + " %ptr_ty %var" +
-         (transformed.size() ? " " : "") + transformed +
+         (transformed.empty() ? "" : " ") + transformed +
```
dneto0: Prefer empty() to using size() to see if a container is empty.

PR #3001, opened 2019-10-30
Add iOS as a supported platform

### LINKED ISSUE #2994: Enable iOS platform
CMake 3.15 added [first-class support for iOS builds](https://cmake.org/cmake/help/v3.15/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-or-watchos).
Xcode projects targeting iOS can be generated by specifying `-DCMAKE_SYSTEM_NAME=iOS` in the command line, but CMake fails with the following error:

```
CMake Error at ThirdParty/SPIRV-Tools/CMakeLists.txt:56 (message):
  Your platform 'iOS' is not supported!
```

There should be no major problems adding iOS support as I can already successfully build the project for iOS by specifying custom CMake toolchain. As a matter of fact, just treating iOS similar to Darwin makes the project build OK.

s-perron:
Treating it the same as Darwin.  Fixes #2994.


DiligentGraphics:
> @DiligentGraphics Can you confirm that this works for you?

Will try it today

CMakeLists.txt
```diff
@@ -47,6 +47,8 @@ elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "CYGWIN")
   add_definitions(-DSPIRV_WINDOWS)
 elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
   add_definitions(-DSPIRV_MAC)
+elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "iOS")
+  add_definitions(-DSPIRV_MAC)
```
DiligentGraphics: It is probably better to define `SPIRV_IOS` macro to be consistent with other platforms and in case you will need to implement something iOS-specific in the future.
There is [only single place in the code](https://github.com/KhronosGroup/SPIRV-Tools/search?q=SPIRV_MAC&unscoped_q=SPIRV_MAC) where  `SPIRV_MAC` macro is used, and where `SPIRV_IOS` will need to be added:

```cpp
#if defined(SPIRV_ANDROID) || defined(SPIRV_LINUX) || defined(SPIRV_MAC) || \
    defined(SPIRV_IOS) || defined(SPIRV_FREEBSD) || defined(SPIRV_EMSCRIPTEN)
```

PR #3010, opened 2019-11-03
spirv-fuzz: better computation of data synonym facts

afd:
When a data synonym fact about two composites is added, data synonym
facts between all sub-components of the composites are also added.

Furthermore, when data synonym facts been all sub-components of two
composites are known, a data synonym fact relating the two composites
is added.  Identification of this case is done in a lazy manner, when
questions about data synonym facts are asked.

The change introduces helper methods to get the size of an array type
and the number of elements of a struct type, and fixes
TransformationCompositeExtract to invalidate analyses appropriately.

### DIFF
--- a/source/fuzz/fact_manager.cpp
+++ b/source/fuzz/fact_manager.cpp
@@ -14,8 +14,9 @@
 
 #include "source/fuzz/fact_manager.h"
 
-#include <map>
 #include <sstream>
+#include <unordered_map>
+#include <unordered_set>
 
 #include "source/fuzz/equivalence_relation.h"
 #include "source/fuzz/fuzzer_util.h"
@@ -29,7 +30,7 @@ namespace {
 
 std::string ToString(const protobufs::Fact& fact) {
   assert(fact.fact_case() == protobufs::Fact::kConstantUniformFact &&
-         "Right now this is the only fact.");
+         "Right now this is the only fact we know how to stringify.");
   std::stringstream stream;
   stream << "("
          << fact.constant_uniform_fact()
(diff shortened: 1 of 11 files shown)


source/fuzz/fact_manager.cpp
```diff
@@ -309,44 +316,482 @@ bool FactManager::ConstantUniformFacts::AddFact(
...
+             DataDescriptorEquals()(&first.second, &second.second);
+    }
+  };
+
+  // This map records, for a given pair of composite data descriptors of the
+  // same type, all the indices at which the data descriptors are known to be
+  // synonymous.  A pair is a key to this map only if we have observed that
+  // the pair are synonymous at *some* index, but not at *all* indices.
+  // Once we find that a pair of data descriptors are equivalent at all indices
+  // indices we record the fact that they are synonymous and remove them from
```
paulthomson: typo: indices repeated

source/fuzz/fact_manager.cpp
```diff
@@ -309,44 +316,482 @@ bool FactManager::ConstantUniformFacts::AddFact(
...
+        auto dd2_it = dd1_it;
+        for (++dd2_it; dd2_it != equivalence_class.end(); ++dd2_it) {
+          auto dd2 = *dd2_it;
+          // If this data descriptor has no indices then it does not have the
+          // form obj_2[b_1, ..., b_n, i], so move on.
+          if (dd2->index_size() == 0) {
+            continue;
+          }
+
+          // At this point we know that:
```
paulthomson: Could assert something here just to be extra safe.

source/fuzz/fact_manager.cpp
afd: Fixed.

source/fuzz/fact_manager.cpp
afd: Yes - I can assert that both dd1 and dd2 have at least one index - done.

source/fuzz/fact_manager.cpp
```diff
@@ -309,44 +316,482 @@ bool FactManager::ConstantUniformFacts::AddFact(
...
+            std::vector<bool> entry;
+            for (uint32_t i = 0; i < num_components_in_composite; i++) {
+              entry.push_back(i == common_final_index);
+            }
+            candidate_composite_synonyms[candidate_composite_synonym] = entry;
+            existing_entry =
+                candidate_composite_synonyms.find(candidate_composite_synonym);
+          } else {
+            // We have seen this pair of data descriptors before, and we know
+            // know that they are synonymous at one further index, so we
```
paulthomson: typo: repeated "know"...or perhaps "now know"

PR #3023, opened 2019-11-08
Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/2997

Tobski:
Also validates the type parameters of Result Type and Value for
OpGroupNonUniformBroadcast, BroadcastFirst, and QuadBroadcast

### DIFF
--- a/source/val/validate_non_uniform.cpp
+++ b/source/val/validate_non_uniform.cpp
@@ -50,6 +50,97 @@ spv_result_t ValidateGroupNonUniformBallotBitCount(ValidationState_t& _,
   return SPV_SUCCESS;
 }
 
+spv_result_t ValidateGroupNonUniformBroadcast(ValidationState_t& _,
+                                              const Instruction* inst) {
+  // Scope is already checked by ValidateExecutionScope() above.
+
+  const uint32_t result_type = inst->type_id();
+  if (!(_.IsFloatScalarOrVectorType(result_type) ||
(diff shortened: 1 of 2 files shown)


Tobski:
Note for reviewers: spir-v tools don't yet support SPIR-V 1.5, so I couldn't figure out a (non-hacky) way to write a test to check the behaviour under 1.5 semantics (i.e. that constants aren't required). It seems to work, but I can't write tests for the 1.5 portion until that support lands. It wasn't clear to me if I should even handle the 1.5 semantics in the validation code because of this, so I put it into a separate commit.

alan-baker:
SPIRV-Tools does support SPIR-V 1.5 as of 5a48c0da15628d69b176f152fcddb1d67661f61e (Sept 13). Validation tests can be written by compiling and validating using the SPV_ENV_UNVERSAL_1_5 target environment. There are tests in test/val/val_capability_tests.cpp that use it.

Tobski:
That's weird... it didn't show up anywhere when I tried to use it, checked for that exact environment variable, in fact. I definitely fetched the latest head too 🤔 

Something screwy probably going on with my local repo, I'll grab it fresh again and see what happens.

s-perron:
@Tobski Is this still being worked on?

s-perron:
This seems to be stale.  I'll close it.  Reopen if it is still being worked on.

PR #3024, opened 2019-11-08
Update Offset to ConstOffset bitmask if operand is constant.

ehsannas:


source/opt/folding_rules.cpp
```diff
@@ -2218,6 +2257,68 @@ FoldingRule RemoveRedundantOperands() {
...
+    if (operand_index >= 0) {
+      auto image_operands = inst->GetSingleWordOperand(operand_index);
+      if (image_operands & SpvImageOperandsOffsetMask) {
+        uint32_t offset_operand_index = operand_index + 1;
+        if (image_operands & SpvImageOperandsBiasMask) offset_operand_index++;
+        if (image_operands & SpvImageOperandsLodMask) offset_operand_index++;
+        if (image_operands & SpvImageOperandsGradMask)
+          offset_operand_index += 2;
+        // Note that it is illegal to have both Offset and ConstOffset bits
+        // enabled in the bitmask.
```
s-perron: No need for the comment.  The text in the assert says essentially the same thing.

source/opt/folding_rules.cpp
```diff
@@ -2218,6 +2257,68 @@ FoldingRule RemoveRedundantOperands() {
...
+        assert(((image_operands & SpvImageOperandsConstOffsetMask) == 0) &&
+               "Offset and ConstOffset may not be used together");
+        if (offset_operand_index < inst->NumOperands()) {
+          Instruction* offset_inst = context->get_def_use_mgr()->GetDef(
+              inst->GetSingleWordOperand(offset_operand_index));
+          if (offset_inst && offset_inst->IsConstant()) {
```
s-perron: You should check if the operand is constant by looking at the unnamed parameter at line 2264.

test/opt/fold_test.cpp
```diff
@@ -6941,6 +6941,136 @@ ::testing::Values(
                                  1, false)
 ));
 
+// clang-format on
+using ImageOperandsBitmaskFoldingTest =
+    ::testing::TestWithParam<InstructionFoldingCase<bool>>;
+
+TEST_P(ImageOperandsBitmaskFoldingTest, Case) {
```
s-perron: Why did you have to rewrite this?  I think you could have reused `MatchingInstructionWithNoResultFoldingTest`.  None of your tests test what happens to uses of the result id.  They all want to fold the instruction before the return. 
 Even if the instruction before the return has a result id, it should still get folded by `MatchingInstructionWithNoResultFoldingTest`.  It might just need to be renamed.

source/opt/folding_rules.cpp
ehsannas: Done

source/opt/folding_rules.cpp
ehsannas: Done

test/opt/fold_test.cpp
ehsannas: Ah, yes. I hadn't seen this `MatchingInstructionWithNoResultFoldingTest`.  Done. I reused the same name.

PR #3055, opened 2019-11-19
Kill the id-to-func map after wrap-opkill

### LINKED ISSUE #3038: opt: dEQP-VK.graphicsfuzz.two-nested-infinite-loops-discard
This test fails. To reproduce: 

```
spirv-opt -O -o temp.spv *.spv
spirv-val temp.spv
```

[two-nested-infinite-loops-discard.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/3833348/two-nested-infinite-loops-discard.zip)

s-perron:
Wrap-opkill will create a new function, invalidating the id-to-func map.
The preserved analyses for the pass have been updated to reflect that.

Also adding consistency check for the id-to-func map.  With this new
check, old tests identify this problem.  No new tests are needed.

Fixes #3038

### DIFF
--- a/source/opt/ir_context.cpp
+++ b/source/opt/ir_context.cpp
@@ -273,6 +273,14 @@ bool IRContext::IsConsistent() {
     }
   }
 
+  if (AreAnalysesValid(kAnalysisIdToFuncMapping)) {
+    for (auto& fn : *module_) {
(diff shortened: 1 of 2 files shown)


s-perron:
The macos failure is related to the nosetest that is being looked into.

PR #3064, opened 2019-11-22
Add test with explicit example of stripping reflection info

dneto0:


s-perron:
What is the purpose of adding this test?  It looks like it is the same as the already existing test just below it.

dneto0:
I wrote this to give a complete code example to a developer. I have improved the comment to clarify its purpose.

PR #3068, opened 2019-11-25
Validate nested constructs

### LINKED ISSUE #3031: opt: dEQP-VK.graphicsfuzz.discard-continue-return
This test fails. To reproduce: 

```
spirv-opt -O -o temp.spv *.spv
spirv-val temp.spv
```

[discard-continue-return.zip](https://github.com/KhronosGroup/SPIRV-Tools/files/3833315/discard-continue-return.zip)

alan-baker:
Validate the spec rule:
> when a construct contains another header block, it also contains that header’s corresponding merge block if that merge block is reachable in the CFG

Fixes #3031 

Identified problems in block merging and inlining. Block merging will no longer merge a continue target. Inlining a single block loop will not modify the continue target. Had to fix up many invalid tests.

### DIFF
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -284,12 +284,11 @@ bool InlinePass::GenInlineCode(
     if (rid != 0) callee_result_ids.insert(rid);
   });
 
-  // If the caller is in a single-block loop, and the callee has multiple
-  // blocks, then the normal inlining logic will place the OpLoopMerge in
-  // the last of several blocks in the loop.  Instead, it should be placed
-  // at the end of the first block.  First determine if the caller is in a
-  // single block loop.  We'll wait to move the OpLoopMerge until the end
-  // of the regular inlining logic, and only if necessary.
+  // If the caller is a loop header and the callee has multiple blocks, then the
+  // normal inlining logic will place the OpLoopMerge in the last of several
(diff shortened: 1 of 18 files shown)


alan-baker:
This PR was tested against CTS and found issues with several tests. Those are being addressed separately, but this PR should not be integrated into CTS until those changes are made.

source/opt/inline_pass.cpp
```diff
@@ -291,13 +290,9 @@ bool InlinePass::GenInlineCode(
   // at the end of the first block.  First determine if the caller is in a
   // single block loop.  We'll wait to move the OpLoopMerge until the end
   // of the regular inlining logic, and only if necessary.
-  bool caller_is_single_block_loop = false;
```
dneto0: The comment immediately above this calls out the single-block loop case.
But what's really salient about the situation is that the caller is in the loop header, and the callee has multiple blocks.
You should update the comment to reflect the more accurate/salient aspect of the situation.

test/val/val_cfg_test.cpp
```diff
@@ -2038,12 +2039,9 @@ TEST_P(ValidateCFG,
 }
 
 TEST_P(ValidateCFG, ContinueTargetCanBeMergeBlockForNestedStructureGood) {
```
dneto0: Please update the test name.  Shouldn't say Good unconditionally at the end.

PR #3073, opened 2019-11-27
graphics robust access: use signed clamp

### LINKED ISSUE #3072: graphics-robust-access clamping incorrectly
When the graphics-robust-access pass runs it inserts `UClamp` calls for each of the bounds, it should be using `SClamp`. This is because negative indices will end up being very large and clamping to the end instead of clamping to the beginning.

This can be see in the `compute_robust_buffer_access_ssbo.amber` test case.

dneto0:
Access chain indices are always interpreted as signed integers.
So use signed clamp instead of unsigned clamp.  We must also
clamp to the max signed int for the index type.

Fixes #3072

### DIFF
--- a/source/opt/graphics_robust_access_pass.cpp
+++ b/source/opt/graphics_robust_access_pass.cpp
@@ -121,6 +121,10 @@
 //      the result of an OpArrayLength instruction acting on the pointer of
 //      the containing structure as noted above.
 //
+//      - Access chain indicies are always treated as signed, so:
+//        - Clamp the upper bound at the signed integer maximum.
+//        - Use SClamp for all clamping.
+//
 //    - TODO(dneto): OpImageTexelPointer:
 //      - Clamp coordinate to the image size returned by OpImageQuerySize
 //      - If multi-sampled, clamp the sample index to the count returned by
@@ -141,6 +145,7 @@
 #include <cstring>
 #include <functional>
 #include <initializer_list>
+#include <limits>
 #include <utility>
 
 #include "constants.h"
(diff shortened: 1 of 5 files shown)


dneto0:
Testing this with swiftshader and amber

dneto0:
yay, it passes the previously failing Amber test.

$ VK_ICD_FILENAMES=`pwd`/Linux/vk_swiftshader_icd.json ./amber ../amber/tests/cases/compute_robust_buffer_access_ssbo.amber 

Summary: 1 pass, 0 fail

dj2:
Should we create an Amber test with the SPIR-V ASM for the UClamp version and upload it to the CTS?

source/opt/graphics_robust_access_pass.cpp
```diff
@@ -121,6 +121,10 @@
 //      the result of an OpArrayLength instruction acting on the pointer of
 //      the containing structure as noted above.
 //
+//      - Access chain indicies are always treated as signed, so:
```
alan-baker: indicies -> indices

dneto0:
> Should we create an Amber test with the SPIR-V ASM for the UClamp version and upload it to the CTS?

I don't think so.  That would in effect only be testing the functionality of UClamp.  That's covered elsewhere by targeted tests.
I believe I was mistaken yesterday thinking UClamp was wrong; instead I another explanation is the driver on my desktop might automatically be clipping the index.

test/opt/graphics_robust_access_test.cpp
```diff
@@ -937,11 +940,13 @@ TEST_F(GraphicsRobustAccessTest, ACRTArrayGeneralShortIndexClamped) {
        ; CHECK: %uint = OpTypeInt 32 0
        ; CHECK-DAG: %uint_1 = OpConstant %uint 1
        ; CHECK-DAG: %uint_0 = OpConstant %uint 0
+       ; CHECK-DAG: %[[intmax:\w+]] = OpConstant %uint 2147483647
        ; CHECK: OpLabel
        ; CHECK: %[[arrlen:\w+]] = OpArrayLength %uint %var 2
        ; CHECK-DAG: %[[max:\w+]] = OpISub %uint %[[arrlen]] %uint_1
        ; CHECK-DAG: %[[i_ext:\w+]] = OpSConvert %uint %i
-       ; CHECK: %[[clamp:\w+]] = OpExtInst %uint %[[GLSLSTD450]] UClamp %[[i_ext]] %uint_0 %[[max]]
+       ; CHECK: %[[smin:\w+]] = OpExtInst %uint %[[GLSLSTD450]] UMin %[[max]] %[[intmax]]
```
dneto0: Answering @alan-baker:   This is an example where a 16bit index i is widened to 32 bits, and we make sure its widened version is still within 0.. signed-int32-max

dneto0:
> Do you have a test where you increase the width to fit the max signed number? It tricky to dissect all the tests from the diff.

Yes, e.g. ACRTArrayGeneralShortIndexClamped

The old version of the test:
https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/opt/graphics_robust_access_test.cpp#L923

I commented on the updated test.

dneto0:
https://github.com/KhronosGroup/SPIRV-Tools/pull/3077 updates validation to justify not having to add the Int64 capability when widening the max permitted value.

dneto0:
> #3077 updates validation to justify not having to add the Int64 capability when widening the max permitted value.

Actually, there is another reason to not have to add Int64 capability.  See the update to optimizer.hpp for details.

Ready for review again.

PR #3074, opened 2019-11-27
spirv-fuzz: Improve debugging facilities

afd:
Adds an option to run the validator on the SPIR-V binary after each
fuzzer pass has been applied, to help identify when the fuzzer has
made the module invalid.  Also adds a helper method to allow dumping
of the sequence of transformations that have been applied to a JSON
file.


dneto0:
> Also adds a helper method to allow dumping
of the sequence of transformations that have been applied to a JSON
file.

FYI. SPIR-V 1.1 added OpModuleProcessed to record this kind of info in the module itself.  But I don't know that any tool added support for it.

afd:
Interesting regarding OpModuleProcessed @dneto0 - I'll take a look.

PR #3077, opened 2019-11-27
WebGPU: Array size at most max signed int + 1

dneto0:
Khronos SPIR-V internal issue 524


dneto0:
The testing infra for the existing array length tests were quite awkward, but I didn't want to rock the boat too much there.

dneto0:
After discussing with @alan-baker I'll make this only for WebGPU.

dneto0:
Read for review again.

dneto0:
Arg. The [0-9] regular expression is not recognized by the googletest regexp parser.

PR #3078, opened 2019-11-29
spirv-fuzz: function outlining fuzzer pass

afd:
A new transformation and associated fuzzer pass in spirv-fuzz that
selects single-entry single-exit control flow graph regions and for
each selected region outlines the region into a new function and
replaces the original region with a call to this function.


source/fuzz/transformation_outline_function.cpp
```diff
@@ -0,0 +1,1011 @@
...
+  }
+  return result;
+}
+
+std::vector<uint32_t> TransformationOutlineFunction::GetRegionOutputIds(
+    opt::IRContext* context, const std::set<opt::BasicBlock*>& region_set,
+    opt::BasicBlock* region_entry_block, opt::BasicBlock* region_exit_block) {
+  std::vector<uint32_t> result;
+
+  // Consider each block in the function containing the region.
```
paulthomson: contained in the region?

source/fuzz/transformation_outline_function.cpp
```diff
@@ -0,0 +1,1011 @@
...
+    if (inst.opcode() == SpvOpPhi) {
+      continue;
+    }
+    outlined_region_entry_block->AddInstruction(
+        std::unique_ptr<opt::Instruction>(inst.Clone(context)));
+  }
+  outlined_function->AddBasicBlock(std::move(outlined_region_entry_block));
+
+  // We now go through the single-entry single-exit region defined by the entry
+  // and exit blocks, and clones of all such blocks to the new function.
```
paulthomson: typo: and clones

test/fuzz/transformation_outline_function_test.cpp
```diff
@@ -0,0 +1,2333 @@
...
+
+  FactManager fact_manager;
+
+  TransformationOutlineFunction transformation(6, 8, 100, 101, 102, 300, 103,
+                                               104, 105, {}, {});
+  ASSERT_FALSE(transformation.IsApplicable(context.get(), fact_manager));
+}
+
+TEST(TransformationOutlineFunctionTest,
+     DoNotOutlineIfLoopHeadIsOutsiderRegion) {
```
paulthomson: Type? Outsider?

afd:
Hmm, that's an interesting idea regarding doing some splitting to just avoid the complex cases.

If we do that, I wonder whether we should have the outliner pass emit the splitting transformations - i.e. decide what it would like to outline, do any splitting if necessary to make outlining possible, and then do the outlining.  Or instead we could just leave it to chance that "split blocks" will sometimes have been run by the fuzzer by chance in a way that makes more outlining opportunities possible.

What do you think?

source/fuzz/transformation_outline_function.cpp
afd: Fixed.

test/fuzz/transformation_outline_function_test.cpp
afd: Fixed, thanks.

paulthomson:
I assume we would want the outliner pass to emit the splitting transformations. The splits need to happen in a certain way (I think): the original entry block will get a fresh successor block (that becomes the new entry block), while the exit block will probably need a fresh predecessor block (that becomes the new exit block).

paulthomson:
Not sure about the case of inlining a single block.

PR #3079, opened 2019-11-29
spirv-fuzz: Fix invalid tests

afd:
This change fixes some tests that turned out to be invalid; recent
improvements to the validator in #3068 exposed them as such.


afd:
Those tests are disabled unless you configure with -DSPIRV_ENABLE_LONG_FUZZER_TESTS=ON.

PR #3081, opened 2019-12-02
Remove names and decorations of imported symbols

aqnuep:
There seems to be no reason to keep the original variable/function/parameter decorations around for import symbols, as the corresponding exports should either match them or should override them. At least that's my understanding, but if there are any exceptions to that please let me know.

Also, with this change the generated module won't have stale decoration and name instructions referring to ids which no longer exist after linking.

### DIFF
--- a/source/link/linker.cpp
+++ b/source/link/linker.cpp
@@ -137,8 +137,6 @@ spv_result_t CheckImportExportCompatibility(const MessageConsumer& consumer,
 // TODO(pierremoreau): Linkage attributes applied by a group decoration are
 //                     currently not handled. (You could have a group being
 //                     applied to a single ID.)
-// TODO(pierremoreau): Run a pass for removing dead instructions, for example
-//                     OpName for prototypes of imported funcions.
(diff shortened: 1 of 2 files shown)


dneto0:
Please add a test to check this change in functionality.

Also I'd like @pierremoreau's opinion.

aqnuep:
I've added a test case to the PR. Let me know if anything else needs to be modified.

pierremoreau:
And sorry for never sending the patch. :-/

pierremoreau:
@aqnuep Could you please rebase the series on the latest master? It includes the fix for the check-format test, which is currently failing on code unrelated to your MR.

aqnuep:
Done!

pierremoreau:
Thank you!
Hopefully @dneto0 or @s-perron can schedule a new kokoro run whenever they have time.

pierremoreau:
If there are no other comments on this merge request, could someone please merge it (as I do not have write access to the repository myself)? Thanks in advance.

pierremoreau:
Thank you Steven for merging. :slightly_smiling_face:

PR #3082, opened 2019-12-02
Add unrolling to performance passes

### LINKED ISSUE #3067: unnecessary copy when using matrices
The high-level shader for this issue was reported on https://github.com/microsoft/DirectXShaderCompiler/issues/2598.

The issue comes down to a matrix variable that seems unnecessary to have. It is initialized by copying the matrix from another variable and only used once. (hence the original matrix variable could have been used). The user is reporting a 17x slowdown (the duration in RenderDoc for the draw using this shader is ~1300μs with the copy and ~75μs without the copy).

The HLSL shader is:
```hlsl
struct PSInput
{
    float4 color : COLOR;
};

cbuffer bob
{
    float4x3 matrices[3];  
};

struct compute_params
{
    uint index;
    float4x3 input_matrices[3];  
};

inline float3 compute_result(float4x3 input_matrices[3], uint index)
{
    return input_matrices[index][0];
}

float4 main(PSInput input) : SV_TARGET
{
    float4 result = input.color;
    
    compute_params p;
    for (uint i = 0; i < 3; i++) {
        p.input_matrices[i] = matrices[i];
    }
    p.index = 0;
    
    result.xyz += compute_result(p.input_matrices, p.index);
    return result;
}

```

Here is the compilation result to SPIR-V:
http://shader-playground.timjones.io/6159d5ca53c9ddbfcaab4199fac479ec

specifically note:
```spir-v
         %29 = OpVariable %_ptr_Function__arr_mat4v3float_uint_3_0 Function
         %30 = OpVariable %_ptr_Function__arr_mat4v3float_uint_3_0 Function
         %41 = OpLoad %_arr_mat4v3float_uint_3_0 %29
               OpStore %30 %41

       ; uses %30 for the remainder of the shader (could have used %29 instead)
```

@s-perron can you share your thoughts on this?
Thanks

s-perron:
Unroll loops that are marked as unroll when doing -O.  Add cleanup
optimizations after unrolling.

Fixes #3067


s-perron:
@ehsannas I've made changes to the performance optimization to be able to get this case when the loop is marked as unroll.  Can you make sure this is acceptable for DXC?

ehsannas:
Thanks. I'll check and provide an update.

s-perron:
I was told this caused some regressions in DXC.  I'm looking into them now.

s-perron:
This is exposing a bug in local-access-chain-convert.  There is an OpAccessChain that creates a pointer that over flows the array.  That is not considered invalid as long as it does not execute.  That access chain in getting converted into an OpCompositeExtract that is out-of-bounds.  This fails validation.  I'll open another PR to fix that issue, then we can merge this one.

ehsannas:
Thanks for finding the issue @s-perron

s-perron:
I believe the validation error related to OpCompositeExtract has been fixed.  This is good to merge now.  @ehsannas can you check that this is works well for DXC?  I'll be merging it.

ehsannas:
I'll test it later today or early tomorrow.

ehsannas:
@s-perron passes all tests. LGTM

Issue #2108, opened 2018-11-23
Validate variable pointer storage class restrictions


alan-baker:
Wait for the resolution of Khronos internal issue 384 to be resolved.

alan-baker:
All variable pointers must be in either StorageBuffer or Workgroup (requires full VariablePointers capability) storage classes.

PR #3091, opened 2019-12-09
Fix and add more checks for logical and variable pointers

nhaehnle:
Factor out the check of some generic validation rules about generating
logical and variable pointers to ensure that:

1. Variable pointers rules are checked also in the PhysicalStorageBuffer64
   addressing model.

2. The storage class of variable pointers is checked as per the spec.

3. Only the explicitly white-listed instructions can generate logical
   pointers.

This change assumes that OpInBoundsPtrAccessChain should be treated
the same as OpPtrAccessChain for the purpose of these validation
rules.

Fixes issue #2108.
See also: https://github.com/KhronosGroup/SPIRV-Tools/milestone/7
See also: Khronos SPIR-V issues #526 and #528

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -306,6 +306,7 @@ set(SPIRV_SOURCES
   ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_layout.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_literals.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_logicals.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_logical_pointers.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_memory.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_memory_semantics.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/val/validate_misc.cpp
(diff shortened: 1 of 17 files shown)


alan-baker:
You should also test this against CTS as heads up for any necessary changes there.

s-perron:
@nhaehnle Will you be fixing up this PR so the build no longer fail?  If not, please close the PR.

nhaehnle:
> @nhaehnle Will you be fixing up this PR so the build no longer fail? If not, please close the PR.

Yes, the idea is to get around to it. Sorry it's taking so long... I did just rebase, but won't be able to look at anything else before next week.

nhaehnle:
There are now some validation failures on the CTS, and I'm going to work through those first.

In the meantime, is there a reliable way to reproduce the formatting checks of the builders locally? The utils/check_code_format script doesn't work by default, and if I use the same `curl` line to download the required Python script first, I get some formatting complaints, but they're different from the ones in the builder log...

alan-baker:
I've noticed I sometimes get different formatting depending on my platform too. Sometimes I just download the errors as a patch and apply it.

s-perron:
In general, you can fix the formatting by running clang-format with the `.clang-format` in the repo on the files that you have changed.  As Alan points out, sometimes clang-format will do different things if you have a different version of clang-format.

nhaehnle:
Note: I'm keeping that last commit separate for now on purpose until the SPIR-V issue and VK-GL-CTS changes have settled.

nhaehnle:
Note: I'm keeping that last commit separate for now on purpose until the SPIR-V issue and VK-GL-CTS changes have settled.

s-perron:
@nhaehnle Can you rebase this PR and fix the build failures?

s-perron:
This looks stale.  I'll close it.  Reopen if it is still being worked on.

PR #3092, opened 2019-12-09
Don't crash when folding construct of empty struct

s-perron:
An OpCompositeConstruct of an empty struct will be folded to a constant
under normal circumstances.  However, if the id limit has been reached
and the constant cannot be generated, then other folding rules will be
tried.

These rules do not handle the case of an empty struct.  We add allow it
to be handled.

Fixes http://crbug/1030194

### DIFF
--- a/source/opt/folding_rules.cpp
+++ b/source/opt/folding_rules.cpp
@@ -1524,7 +1524,7 @@ bool CompositeExtractFeedingConstruct(
   // - extracting the same position they are inserting
   // - all extract from the same id.
   for (uint32_t i = 0; i < inst->NumInOperands(); ++i) {
-    uint32_t element_id = inst->GetSingleWordInOperand(i);
+    const uint32_t element_id = inst->GetSingleWordInOperand(i);
     Instruction* element_inst = def_use_mgr->GetDef(element_id);
 
     if (element_inst->opcode() != SpvOpCompositeExtract) {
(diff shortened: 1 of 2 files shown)


source/opt/folding_rules.cpp
```diff
@@ -1501,60 +1501,64 @@ FoldingRule CompositeConstructFeedingExtract() {
...
+  }
 
-      if (element_inst->opcode() != SpvOpCompositeExtract) {
-        return false;
-      }
+  // Check each element to make sure they are:
+  // - extractions
+  // - extracting the same position they are inserting
+  // - all extract from the same id.
+  for (uint32_t i = 0; i < inst->NumInOperands(); ++i) {
```
kuhar: not sure what the convention is, but this could be
`for (uint32_t i = 0, e = inst->NumInOperands(); i != e; ++i)`

source/opt/folding_rules.cpp
```diff
@@ -1501,60 +1501,64 @@ FoldingRule CompositeConstructFeedingExtract() {
...
+    uint32_t element_id = inst->GetSingleWordInOperand(i);
```
kuhar: `const uint32_t element_id`?

source/opt/folding_rules.cpp
s-perron: We use the idiom that I wrote.

PR #3093, opened 2019-12-09
Disallow forward references in arrays

alan-baker:
Fixes https://crbug.com/1031049

* Disallow forward references in arrays
* Add a test


alan-baker:
I believe glslang is generating incorrect code in this case.

dneto0:
@alan-baker Please make sure there is a fix or tracking issue for the Glslang codegen issue.

PR #3096, opened 2019-12-10
Make Instrumentation format version 2 the default (Step 1)

greg-lunarg:
Add new interfaces without version number argument. Remove version 1
logic and tests. Version interfaces will be removed in step 2 after
layers have transitioned to new interface.


include/spirv-tools/instrument.hpp
```diff
@@ -160,6 +159,10 @@ static const int kInst2BindlessUninitOutCnt = kInst2StageOutCnt + 3;
 
 // A buffer address unalloc error will output the 64-bit pointer in
 // two 32-bit pieces, lower bits first.
+static const int kInstBuffAddrUnallocOutDescPtrLo = kInstStageOutCnt + 1;
+static const int kInstBuffAddrUnallocOutDescPtrHi = kInstStageOutCnt + 2;
+static const int kInstBuffAddrUnallocOutCnt = kInstStageOutCnt + 3;
+
 static const int kInst2BuffAddrUnallocOutDescPtrLo = kInst2StageOutCnt + 1;
 static const int kInst2BuffAddrUnallocOutDescPtrHi = kInst2StageOutCnt + 2;
 static const int kInst2BuffAddrUnallocOutCnt = kInst2StageOutCnt + 3;
```
s-perron: If the kInst2* values are suppose to equal the corresponding kInst* value, it might be better to make them equal:

```
kInst2BuffAddrUnallocOutDescPtrLo = kInstBuffAddrUnallocOutDescPtrLo;
```

source/opt/inst_bindless_check_pass.h
```diff
@@ -28,18 +28,19 @@ namespace opt {
...
-        input_length_enabled_(true),
-        input_init_enabled_(true) {}
-  // For all other interfaces
+  // Deprecated interface
   InstBindlessCheckPass(uint32_t desc_set, uint32_t shader_id,
                         bool input_length_enable, bool input_init_enable,
                         uint32_t version)
       : InstrumentPass(desc_set, shader_id, kInstValidationIdBindless, version),
         input_length_enabled_(input_length_enable),
         input_init_enabled_(input_init_enable) {}
```
s-perron: Should we issue an error if the version is not 2?  If somebody is still trying to get version 1, they should be told.

source/opt/inst_buff_addr_check_pass.h
```diff
@@ -28,13 +28,13 @@ namespace opt {
...
 class InstBuffAddrCheckPass : public InstrumentPass {
  public:
-  // For test harness only
-  InstBuffAddrCheckPass()
-      : InstrumentPass(7, 23, kInstValidationIdBuffAddr, 1) {}
-  // For all other interfaces
+  // Deprecated interface
   InstBuffAddrCheckPass(uint32_t desc_set, uint32_t shader_id, uint32_t version)
       : InstrumentPass(desc_set, shader_id, kInstValidationIdBuffAddr,
                        version) {}
```
s-perron: Same thing here.  Issue an error if they ask for an unsupported version.

include/spirv-tools/instrument.hpp
greg-lunarg: All the kInst2* values are deprecated and will be removed as soon as the layers move away from them, which will happen as soon as this gets checked in, which will be a matter of days. The current way they are expressed represents their "true" nature, so I am not sure I want to change that. Finally, I was trying not to disrupt things too much at this point. So if we can leave them as they are for the moment, I think that would be best.

source/opt/inst_bindless_check_pass.h
greg-lunarg: I did add an assert in instrument_pass.cpp at the top of InstrumentInitialize() which should be sufficient to catch any uses of version 1 when they test the layers. Again, once the layers transition away from specifying the version, the "version" interfaces will go away as well as version_ and the assert. So I think we are covered. Let me know if you still see a hole.

source/opt/inst_buff_addr_check_pass.h
greg-lunarg: See my response above.

source/opt/inst_bindless_check_pass.h
s-perron: Okay, now I see the assert.  The assert will not trigger in a release build.  Could we log an error message, which will be seen with release builds?

include/spirv-tools/instrument.hpp
s-perron: Okay.

source/opt/inst_bindless_check_pass.h
greg-lunarg: OK. Will do.

greg-lunarg:
I believe I have addressed all concerns to this point. Thanks!

PR #3097, opened 2019-12-11
spirv-fuzz: Fuzzer pass to merge blocks

afd:
This change adds a fuzzer pass that randomly merges blocks in a module.


source/fuzz/fuzzer_pass_merge_blocks.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+void FuzzerPassMergeBlocks::Apply() {
+  // First we populate a sequence of transformations that we might consider
+  // applying.
+  std::vector<TransformationMergeBlocks> potential_transformations;
+  // We do this by considering every block of every function.
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      // We probabilistically decide to ignore some blocks.
+      if (!GetFuzzerContext()->ChoosePercentage(
```
paulthomson: I wonder whether this should be done *after* checking if each potential transformation is applicable.

source/fuzz/transformation_merge_blocks.cpp
```diff
@@ -0,0 +1,78 @@
...
+    opt::IRContext* context, spvtools::fuzz::FactManager* /*unused*/) const {
+  auto second_block = fuzzerutil::MaybeFindBlock(context, message_.block_id());
+  auto first_block =
+      context->cfg()->block(context->cfg()->preds(second_block->id()).at(0));
+
+  auto function = first_block->GetParent();
+  // We need an iterator pointing to the predecessor, hence the loop.
+  for (auto bi = function->begin(); bi != function->end(); ++bi) {
+    if (bi->id() == first_block->id()) {
+      opt::blockmergeutil::MergeWithSuccessor(context, function, bi);
```
paulthomson: Maybe add a comment noting that `CanMergeWithSuccessor` was already called in `IsApplicable`, so we know the precondition holds.

source/fuzz/fuzzer_pass_merge_blocks.cpp
afd: I think the probability of choosing an applicable transformation is the same either way, and it seems preferable to only do the work of checking applicability if we actually want to do take the transformation.

source/fuzz/transformation_merge_blocks.cpp
afd: Done - via an assertion actually.

PR #3101, opened 2019-12-13
spirv-fuzz: Transformations to add types, constants and variables

afd:
This change adds several transformations that allow types, constants,
undefined values and global variables to be added to a module.

### DIFF
--- a/source/fuzz/transformation_add_global_variable.cpp
+++ b/source/fuzz/transformation_add_global_variable.cpp
@@ -14,6 +14,8 @@
 
 #include "source/fuzz/transformation_add_global_variable.h"
 
+#include "source/fuzz/fuzzer_util.h"
+
 namespace spvtools {
 namespace fuzz {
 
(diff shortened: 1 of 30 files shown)


source/fuzz/transformation_add_global_variable.cpp
```diff
@@ -0,0 +1,104 @@
...
+  if (message_.initializer_id()) {
+    auto initilizer_instruction =
+        context->get_def_use_mgr()->GetDef(message_.initializer_id());
+    // If an initializer id is provided it must correspond to an instruction.
+    if (!initilizer_instruction) {
+      return false;
+    }
+    // The instruction must be a constant.
+    switch (initilizer_instruction->opcode()) {
+      case SpvOpConstant:
```
paulthomson: Is there not also `SpvOpConstantTrue`, `SpvOpConstantNull`? In a later transformation, you use `ConstantManager::GetConstantsFromIds` to check if things are constant. Could you do that here as well?

source/fuzz/transformation_add_type_matrix.cpp
```diff
@@ -0,0 +1,69 @@
...
+  if (!fuzzerutil::IsFreshId(context, message_.fresh_id())) {
+    return false;
+  }
+  auto column_type =
+      context->get_type_mgr()->GetType(message_.column_type_id());
+  if (!column_type) {
+    return false;
+  }
+  return column_type->AsVector() &&
+         column_type->AsVector()->element_type()->AsFloat();
```
paulthomson: Does it need to be a float? Maybe it obviously has to be but if not could add a comment in the proto file?

source/fuzz/transformation_add_global_variable.cpp
afd: Good point - addressed in the manner suggested.

source/fuzz/transformation_add_type_matrix.cpp
afd: Yes it does - I have added comments in this file and in the .proto file to clarify that the SPIR-V spec requires this.

PR #3104, opened 2019-12-16
Handle conflict between debug info and existing validation rule

### LINKED ISSUE #3086: Contradiction between OpExtInst and debug extension spec
OpenCL.100.debug extension (and [its old version](https://www.khronos.org/registry/spir-v/specs/unified1/DebugInfo.html)) says

"DebugScope, DebugNoScope,
DebugDeclare, DebugValue
instructions can interleave with instructions within a function body.
All other debugging instructions should be located between section 9
(All type declarations (OpTypeXXX instructions), all constant instructions, and
all global variable declarations …​) and section 10 (All function declaration)
per the core SPIR-V specification."

However, when I use a debug instruction such as `%dbg_src = OpExtInst %void %DbgExt DebugSource %6 %src`, it says "ExtInst must be used inside a block", which means OpExtInst is a function scope variable.

In order to avoid this, I think we should update the validation rule for `OpExtInst` when we use it for debug instructions.

### LINKED ISSUE #3102: DebugFunction has a function as its operand which results in spirv-val "not defined ID" error
DebugFunction must be located before a function definition based on the layout rule in the debug info spec:

"All other debugging instructions should be located between section 9
(All type declarations (OpTypeXXX instructions), all constant instructions, and
all global variable declarations …​) and section 10 (All function declaration)
per the core SPIR-V specification."

However, DebugFunction has a function id as one of its operands which results in "not defined ID" error when we run spirv-val

```
; Function information
%main_info = OpExtInst %void %DbgExt DebugFunction %main_name %main_type_info %dbg_src 6 1 %comp_unit %main_linkage_name FlagIsPublic 7 %main

       %main = OpFunction %void None %16
```

outputs

```
error: line 54: ID 2[%main] has not been defined
  %44 = OpExtInst %void %1 DebugFunction %14 %42 %35 6 1 %36 %15 FlagIsProtected|FlagIsPrivate 7 %main
```

### LINKED ISSUE #3103: OpVariable whose storage class is Function right after debug info instructions causes a spirv-val error
Putting OpVariable instructions after debug instructions causes spirv-val error

```
error: line 73: All OpVariable instructions in a function must be the first instructions in the first block.
  %vout = OpVariable %_ptr_Function_VS_OUTPUT Function
```

For example,

```
       %main = OpFunction %void None %16
         %20 = OpLabel

; Start the scope of function "main()"
%main_scope = OpExtInst %void %DbgExt DebugScope %main_info

               OpLine %src 8 13
       %vout = OpVariable %_ptr_Function_VS_OUTPUT Function
```

jaebaek:
This commit includes several changes to resolve conflict between
OpenCL.100.debug extension instructions and existing validation
rule.
1. Allow us to put `OpExtInst` between section 9 (type, const) and
section 10 (function) when it is used for debug info (fixes #3086).
2. Allow a forward reference to a result of `OpFunction` from
`DebugFunction` (fixes #3102).
3. Allow `OpExtInst` for `DebugScope`, `DebugNoScope`,
`DebugDeclare`, and `DebugValue` between `OpLabel` and `OpVariable`
for function local variable (fixes #3103).
4. Update IR loader of spirv-opt to make the result of round trip
running spirv-opt the same with the initial SPIR-V code with debug
info.


jaebaek:
After submitting this CL, I will create a separate PR for the validation of debug info itself.

source/opt/ir_context.h
```diff
@@ -187,6 +187,14 @@ class IRContext {
   inline IteratorRange<Module::inst_iterator> debugs3();
   inline IteratorRange<Module::const_inst_iterator> debugs3() const;
 
+  // Iterators for debug info instructions (excluding OpLine & OpNoLine)
+  // contained in this module.  These are OpExtInst for OpenCL.DebugInfo.100
+  // or DebugInfo extension placed between section 9 and 10.
+  inline Module::inst_iterator debuginfo_begin();
```
dneto0: The name is confusing because there are so many kinds of debug info.
Please rename this to ext_inst_debuginfo or ext_inst_debug_info

source/opt/ir_context.h
```diff
@@ -1018,6 +1044,10 @@ void IRContext::AddDebug3Inst(std::unique_ptr<Instruction>&& d) {
   module()->AddDebug3Inst(std::move(d));
 }
 
+void IRContext::AddDebugInfoInst(std::unique_ptr<Instruction>&& d) {
```
dneto0: Please rename this method to AddExtInstDebugInfo

source/opt/ir_loader.cpp
```diff
@@ -136,6 +140,23 @@ bool IrLoader::AddInstruction(const spv_parsed_instruction_t* inst) {
         }
         function_->AddParameter(std::move(spv_inst));
       } else {
+        if (opcode == SpvOpExtInst &&
+            spvExtInstIsDebugInfo(inst->ext_inst_type)) {
+          const uint32_t ext_inst_index = inst->words[4];
+          const OpenCLDebugInfo100Instructions ext_inst_key =
+              OpenCLDebugInfo100Instructions(ext_inst_index);
+          if (ext_inst_key != OpenCLDebugInfo100DebugScope &&
```
dneto0: What happens if this code is executed with a module using the older DebugInfo extended instruction set? 
Does it work by accident? fail by accident?
Please add a comment about what is expected to work.  And if it's easy to make work with DebugInfo, then please do so.

source/reduce/remove_unreferenced_instruction_reduction_opportunity_finder.cpp
```diff
@@ -52,6 +52,13 @@ RemoveUnreferencedInstructionReductionOpportunityFinder::
     result.push_back(MakeUnique<RemoveInstructionReductionOpportunity>(&inst));
   }
 
+  for (auto& inst : context->module()->debuginfo()) {
```
dneto0: It's a nice touch to add this to the fuzzer!

jaebaek: :+1:

source/val/validate_layout.cpp
```diff
@@ -46,6 +47,19 @@ spv_result_t ModuleScopedInstructions(ValidationState_t& _,
...
                  << "section";
         }
+      } else if (spvExtInstIsDebugInfo(inst->ext_inst_type())) {
+        // Debug info extinst opcodes other than DebugScope, DebugNoScope,
+        // DebugDeclare, DebugValue must be placed between section 9 (types,
+        // constants, global variables) and section 10 (function declarations).
+        if (_.current_layout_section() < kLayoutTypes ||
+            _.current_layout_section() >= kLayoutFunctionDeclarations) {
+          return _.diag(SPV_ERROR_INVALID_LAYOUT, inst)
+                 << "OpenCL.DebugInfo.100 instructions other than "
```
dneto0: Nit:  This applies to both OpenCL.DebugInfo.100 and DebugInfo

source/val/validate_layout.cpp
```diff
@@ -182,6 +196,53 @@ spv_result_t FunctionScopedInstructions(ValidationState_t& _,
...
+              local_debug_info = true;
+            }
+          }
+
+          if (local_debug_info) {
+            if (_.in_function_body() == false) {
+              // DebugScope, DebugNoScope, DebugDeclare, DebugValue must
+              // appear in a function body.
+              return _.diag(SPV_ERROR_INVALID_LAYOUT, inst)
+                     << "DebugScope, DebugNoScope, DebugDeclare, DebugValue "
```
dneto0: This failure is not tested.

source/val/validate_layout.cpp
jaebaek: I changed the wording "Debug info extension instructions other than .. "

source/val/validate_layout.cpp
jaebaek: Added

Issue #3099, opened 2019-12-12
spirv-opt: The convert-relaxed-to-half optimization pass generates invalid SPIR-V.


bejado:
Steps to repro:

With the following GLSL frag shader:

```
#version 410 core

precision mediump float;

layout(location = 1) in vec3 foo;
layout(location = 2) in mat2 bar;

vec3 func(vec3 tap, mat2 M) {
    return vec3(M * tap.xy, 1.0);

    // Doing this instead solves the issue:
    // vec2 temp = M * tap.xy;
    // return vec3(temp, 1.0);
}

void main() {
    vec3 result = func(foo, bar);
}
```

Compile to SPIR-V, optimize, and validate:

```
$ glslangValidator -V -o test.spv test.frag
$ spirv-opt --convert-relaxed-to-half -o half.spv test.spv
$ spirv-val half.spv
```

The validator errors with the following:

```
error: line 71: Expected Constituents to be scalars or vectors of the same type as Result Type components
  %24 = OpCompositeConstruct %v3float %22 %23 %float_1
```

The problematic SPIR-V:

```
%22 = OpCompositeExtract %half %20 0
%23 = OpCompositeExtract %half %20 1
%24 = OpCompositeConstruct %v3float %22 %23 %float_1
```

alan-baker:
@greg-lunarg PTAL.

greg-lunarg:
I am adding a phase to the algorithm that will perform a formal closure on all additional instructions that we want to mark "relaxed" before doing the conversion to half. The current algorithm catches these in an ad hoc manner during the conversion which is causing some instructions to be missed.

Specifically, this is code that tries to *relax* OpPhi instructions and composite instructions such as OpCompositeExtract which are embedded in a larger relaxed compuation in order to eliminate needless conversions.

Pinging @danginsburg who is interested in this pass and will want to test the resulting code.

s-perron:
I believe this is fixed by #3106.

PR #3106, opened 2019-12-17
Fix convert-relax-to-half invalid code (#3099)

greg-lunarg:
Rewrite to do formal closure of relaxed precision on composite and phi instructions contiguous to relaxed computations to "minimize" convert operations.


test/opt/convert_relaxed_to_half_test.cpp
```diff
@@ -1222,6 +1222,147 @@ OpFunctionEnd
                                            defs_after + func_after, true, true);
 }
 
+TEST_F(ConvertToHalfTest, ConvertToHalfWithClosure) {
```
s-perron: Please use `SinglePassRunAndMatch` for tests as much as possible.  When done well, the make it obvious what the test is testing.

greg-lunarg:
CI is showing a format error in a file I didn't even change :P

greg-lunarg:
Rebased to latest master and re-pushing

test/opt/convert_relaxed_to_half_test.cpp
```diff
@@ -1222,6 +1222,115 @@ OpFunctionEnd
...
+; CHECK-NOT: OpDecorate %bar RelaxedPrecision
+OpDecorate %bar Location 2
+OpDecorate %34 RelaxedPrecision
+OpDecorate %36 RelaxedPrecision
+OpDecorate %41 RelaxedPrecision
+OpDecorate %42 RelaxedPrecision
+; CHECK-NOT: OpDecorate %34 RelaxedPrecision
+; CHECK-NOT: OpDecorate %36 RelaxedPrecision
+; CHECK-NOT: OpDecorate %41 RelaxedPrecision
+; CHECK-NOT: OpDecorate %42 RelaxedPrecision
```
s-perron: This nice part about this is that you could do 1 line like:

```
; CHECK-NOT: OpDecorate {%\w+} RelaxedPrecision
```

to say there are no relaxed precision decorations.  A test could still pass if one remains but has a different number.

test/opt/convert_relaxed_to_half_test.cpp
```diff
@@ -1222,6 +1222,115 @@ OpFunctionEnd
...
+%bar = OpVariable %_ptr_Input_mat2v2float Input
+)";
+
+  const std::string func =
+      R"(%main = OpFunction %void None %3
+%5 = OpLabel
+%34 = OpLoad %v3float %foo
+%36 = OpLoad %mat2v2float %bar
+; CHECK: %48 = OpFConvert %v3half %34
+; CHECK: %49 = OpFConvert %v3half %34
```
s-perron: You can use regular expression for the ids so that even if the ids change the tests can still pass.  It make them more rebust of you have to change the pass.

PR #3114, opened 2019-12-19
spirv-fuzz: Transformation to add a new function to a module

afd:
This adds a large transformation that can add a new function to a
SPIR-V module.  This paves the way for donation of code from one
module to another.

### DIFF
--- a/source/fuzz/transformation_add_function.h
+++ b/source/fuzz/transformation_add_function.h
@@ -45,7 +45,7 @@ class TransformationAddFunction : public Transformation {
  private:
   // Attempts to create a function from the series of instructions in
   // |message_.instruction| and add it to |context|.  Returns false if this is
-  // not possible due to the messages not respect the basic structure of a
+  // not possible due to the messages not respecting the basic structure of a
   // function, e.g. if there is no OpFunction instruction or no blocks; in this
   // case |context| is left in an indeterminate state.
   //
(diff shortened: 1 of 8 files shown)


source/fuzz/transformation_add_function.h
```diff
@@ -0,0 +1,70 @@
...
+
+  // Adds the function defined by |message_.instruction| to the module
+  void Apply(opt::IRContext* context, FactManager* fact_manager) const override;
+
+  protobufs::Transformation ToMessage() const override;
+
+ private:
+  // Attempts to create a function from the series of instructions in
+  // |message_.instruction| and add it to |context|.  Returns false if this is
+  // not possible due to the messages not respect the basic structure of a
```
paulthomson: respect -> respecting

test/fuzz/transformation_add_function_test.cpp
```diff
@@ -0,0 +1,447 @@
...
+
+  const auto env = SPV_ENV_UNIVERSAL_1_4;
+  const auto consumer = nullptr;
+  const auto context = BuildModule(env, consumer, shader, kFuzzAssembleOption);
+  ASSERT_TRUE(IsValid(env, context.get()));
+
+  FactManager fact_manager;
+
+  TransformationAddFunction transformation1(std::vector<protobufs::Instruction>(
+      {MakeInstructionMessage(
```
paulthomson: Up to you, but could use `// clang-format off` and `// clang-format on`. And then keep each message on its own line.

source/fuzz/transformation_add_function.h
afd: Fixed.

test/fuzz/transformation_add_function_test.cpp
afd: Thanks for the tip, though I'm OK with what clang-format has done here.

PR #3117, opened 2019-12-20
spirv-fuzz: Add fuzzer pass to perform module donation

afd:
This change adds a fuzzer pass that allows code from other SPIR-V
modules to be donated into the module under transformation.  It also
changes the command-line options of the tools so that, in fuzzing
mode, a file must be specified that contains the names of available
donor modules.

### DIFF
--- a/source/fuzz/fuzzer_pass_donate_modules.cpp
+++ b/source/fuzz/fuzzer_pass_donate_modules.cpp
@@ -1,0 +1,584 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_donate_modules.h"
+
+#include <map>
+#include <queue>
+#include <set>
+
+#include "source/fuzz/instruction_message.h"
+#include "source/fuzz/transformation_add_constant_boolean.h"
+#include "source/fuzz/transformation_add_constant_composite.h"
+#include "source/fuzz/transformation_add_constant_scalar.h"
+#include "source/fuzz/transformation_add_function.h"
+#include "source/fuzz/transformation_add_global_undef.h"
+#include "source/fuzz/transformation_add_global_variable.h"
+#include "source/fuzz/transformation_add_type_array.h"
+#include "source/fuzz/transformation_add_type_boolean.h"
+#include "source/fuzz/transformation_add_type_float.h"
+#include "source/fuzz/transformation_add_type_function.h"
+#include "source/fuzz/transformation_add_type_int.h"
+#include "source/fuzz/transformation_add_type_matrix.h"
+#include "source/fuzz/transformation_add_type_pointer.h"
+#include "source/fuzz/transformation_add_type_struct.h"
+#include "source/fuzz/transformation_add_type_vector.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassDonateModules::FuzzerPassDonateModules(
+    opt::IRContext* ir_context, FactManager* fact_manager,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations,
+    const std::vector<fuzzerutil::ModuleSupplier>& donor_suppliers)
+    : FuzzerPass(ir_context, fact_manager, fuzzer_context, transformations),
+      donor_suppliers_(donor_suppliers) {}
+
+FuzzerPassDonateModules::~FuzzerPassDonateModules() = default;
+
+void FuzzerPassDonateModules::Apply() {
+  // If there are no donor suppliers, this fuzzer pass is a no-op.
+  if (donor_suppliers_.empty()) {
+    return;
+  }
+
+  // Donate at least one module, and probabilistically decide when to stop
+  // donating modules.
+  do {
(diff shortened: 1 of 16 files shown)


source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -0,0 +1,584 @@
...
+    assert(function_to_donate && "Function to be donated was not found.");
+
+    // We will collect up protobuf messages representing the donor function's
+    // instructions here, and use them to create an AddFunction transformation.
+    std::vector<protobufs::Instruction> donated_instructions;
+
+    // Scan through the function, remapping each result id that it generates to
+    // a fresh id.  This is necessary because functions include forward
+    // references, e.g. to labels.
+    function_to_donate->ForEachInst([this, &original_id_to_donated_id](
```
paulthomson: Could we just use a for loop here? And below too?

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -0,0 +1,584 @@
...
+                                        const opt::Instruction* instruction) {
+      if (instruction->result_id()) {
+        original_id_to_donated_id->operator[](instruction->result_id()) =
```
paulthomson: Could we use insert here since we know the key should not already exist? This might also apply elsewhere.

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -0,0 +1,584 @@
...
+          opt::Instruction::OperandList input_operands;
+
+          // Consider each input operand in turn.
+          for (uint32_t in_operand_index = 0;
+               in_operand_index < instruction->NumInOperands();
+               in_operand_index++) {
+            std::vector<uint32_t> operand_data;
+            const opt::Operand& in_operand =
+                instruction->GetInOperand(in_operand_index);
+            if (in_operand.type == SPV_OPERAND_TYPE_ID) {
```
paulthomson: It looks like there is also `SPV_OPERAND_TYPE_TYPE_ID` and a few more.

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -0,0 +1,584 @@
...
+                  original_id_to_donated_id->at(in_operand.words[0]));
+            } else {
+              // For non-id operands, we just add each of the data words.
+              for (auto word : in_operand.words) {
+                operand_data.push_back(word);
+              }
+            }
+            input_operands.push_back({in_operand.type, operand_data});
+          }
+          // Transform the
```
paulthomson: Unfinished comment?

source/fuzz/fuzzer_util.h
```diff
@@ -25,9 +25,12 @@
 namespace spvtools {
 namespace fuzz {
 
-// Provides global utility methods for use by the fuzzer
+// Provides types and global utility methods for use by the fuzzer
 namespace fuzzerutil {
 
+// TODO comment
```
paulthomson: Comment

test/fuzz/fuzzer_replayer_test.cpp
```diff
@@ -23,109 +24,35 @@ namespace {
...
-  //     while(j > 0) {
-  //       foo();
-  //       j--;
-  //     }
-  //     do {
-  //       i++;
-  //     } while(i < 4);
-  //   }
-  // }
+// The SPIR-V came from this GLSL:
```
paulthomson: Maybe double-check this file; it is probably fine but it is hard for me to check it due to the weird diff!

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -0,0 +1,584 @@
...
+        // A matching import has found.  Map the result id for the donor import
+        // to the id of the existing import, so that when donor instructions
+        // rely on the import they will be rewritten to use the existing import.
+        original_id_to_donated_id->operator[](donor_import.result_id()) =
+            existing_import.result_id();
+        break;
+      }
+    }
+    // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3116): At present
+    //  we do not handle donation of instruction imports that is,allowing the
```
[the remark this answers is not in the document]
afd: Fixed.

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -0,0 +1,584 @@
...
+    std::map<uint32_t, uint32_t>* original_id_to_donated_id) {
+  // Consider every type/global/constant/undef in the module.
+  for (auto& type_or_value : donor_ir_context->module()->types_values()) {
+    // Each such instruction generates a result id, and as part of donation we
+    // need to associate the donor's result id with a new result id.  That new
+    // result id will either be the id of some existing instruction, or a fresh
+    // id.  This variable captures it.
+    uint32_t new_result_id;
+
+    // Decide how to handle each kind of instruction on a case-by-case basis.
```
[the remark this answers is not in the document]
afd: Done.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Iterating over a function yields its basic blocks.  We want to go through every instruction involved in a function, including OpFunction, OpParameter, etc.  It does not seem that there is a suitable iterator in Function's interface to allow this to be achieved via a for loop.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: You're right - done.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Good catch.

source/fuzz/fuzzer_util.h
afd: Fixed, thanks.

test/fuzz/fuzzer_replayer_test.cpp
afd: Double-checked.

PR #3130, opened 2020-01-06
Fix OpLine bug of merge-blocks pass

### LINKED ISSUE #3118: spirv-opt put OpLine between OpLoopMerge and OpBranchConditional
This issue is related to [a DXC issue](https://github.com/microsoft/DirectXShaderCompiler/issues/2625).

After we run spirv-opt on the attached SPIR-V code, it has a validation error from spirv-val
```
fatal error: generated SPIR-V is invalid: OpLoopMerge must immediately precede either an OpBranch or OpBranchConditional instruction. OpLoopMerge must be the second-to-last instruction in its block.
OpLoopMerge %24 %25 None
```
[a.noopt.spv.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/3990019/a.noopt.spv.txt)

jaebaek:
As explained in #3118, spirv-opt merge-blocks pass causes a
spirv-val error when an OpBranch has an OpLine in front of it.

OpLoopMerge
OpBranch ; Will be killed by merge-blocks pass
OpLabel  ; Will be killed by merge-blocks pass
OpLine   ; will be placed between OpLoopMerge and OpBranch - error!
OpBranch

To fix this issue, this commit moves line info of OpBranch to
OpLoopMerge.

Fixes #3118

### DIFF
--- a/source/opt/block_merge_util.cpp
+++ b/source/opt/block_merge_util.cpp
@@ -175,7 +175,9 @@ void MergeWithSuccessor(IRContext* context, Function* func,
       // the validation error that OpLine is placed between OpLoopMerge
       // and OpBranchConditional.
       auto terminator = bi->terminator();
-      merge_inst->dbg_line_insts() = terminator->dbg_line_insts();
+      auto& vec = terminator->dbg_line_insts();
+      auto& new_vec = merge_inst->dbg_line_insts();
+      new_vec.insert(new_vec.end(), vec.begin(), vec.end());
       terminator->clear_dbg_line_insts();
 
       // Move the merge instruction to just before the terminator.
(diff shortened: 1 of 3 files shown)


source/opt/block_merge_util.cpp
```diff
@@ -171,8 +171,15 @@ void MergeWithSuccessor(IRContext* context, Function* func,
       // flow declaration.
       context->KillInst(merge_inst);
     } else {
+      // Move OpLine/OpNoLine information to merge_inst. This solves
+      // the validation error that OpLine is placed between OpLoopMerge
+      // and OpBranchConditional.
+      auto terminator = bi->terminator();
+      merge_inst->dbg_line_insts() = terminator->dbg_line_insts();
```
zoddicus: This is going to replace the merge instruction debug line instructions, should it be appending instead?

jaebaek: Good point. Updated.

PR #3133, opened 2020-01-08
Add validation rules for OpenCL.DebugInfo.100 extension

jaebaek:
Add validation rules for DebugCompilationUnit, DebugSource,
DebugTypeBasic, DebugTypeVector, DebugTypeArray, DebugTypedef,
DebugTypeFunction, DebugTypeEnum, DebugTypeComposite,
DebugTypeMember, DebugTypeInheritance, DebugFunction,
DebugFunctionDeclaration, DebugLexicalBlock, DebugScope,
DebugLocalVariable, DebugDeclare, DebugExpression.

### DIFF
--- a/source/val/validate_extensions.cpp
+++ b/source/val/validate_extensions.cpp
@@ -2184,7 +2184,7 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
       case OpenCLDebugInfo100DebugOperation:
         // The binary parser validates the opcode for DebugInfoNone,
         // DebugNoScope, DebugOperation, and the literal values don't need
-        // furhter checks.
+        // further checks.
         break;
       case OpenCLDebugInfo100DebugCompilationUnit: {
         CHECK_DEBUG_OPERAND("Source", OpenCLDebugInfo100DebugSource, 7);
--- a/source/val/validate_extensions.cpp
+++ b/source/val/validate_extensions.cpp
@@ -20,6 +20,7 @@
 #include <string>
 #include <vector>
 
+#include "OpenCLDebugInfo100.h"
 #include "source/diagnostic.h"
 #include "source/enum_string_mapping.h"
 #include "source/extensions.h"
--- a/source/val/validate_extensions.cpp
+++ b/source/val/validate_extensions.cpp
@@ -42,6 +42,23 @@ uint32_t GetSizeTBitWidth(const ValidationState_t& _) {
   return 0;
 }
 
+spv_result_t ValidateOperandBaseType(ValidationState_t& _,
+                                     const Instruction* inst,
+                                     uint32_t operand_index,
+                                     std::function<std::string()> ext_inst_name) {
+  auto* base_type = _.FindDef(inst->word(operand_index));
+  if (base_type->opcode() != SpvOpExtInst ||
+      base_type->ext_inst_type() != SPV_EXT_INST_TYPE_OPENCL_DEBUGINFO_100 ||
+      OpenCLDebugInfo100Instructions(base_type->word(4)) !=
+          OpenCLDebugInfo100DebugTypeBasic) {
+    return _.diag(SPV_ERROR_INVALID_DATA, inst)
+           << ext_inst_name() << ": "
+           << "expected operand Base Type must be a result id of "
+           << "DebugTypeBasic";
+  }
+  return SPV_SUCCESS;
+}
+
 }  // anonymous namespace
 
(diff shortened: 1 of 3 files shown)


source/val/validate_extensions.cpp
```diff
@@ -2028,6 +2131,377 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
...
+        auto validate_src = ValidateDebugInfoOperand(
+            _, "Source", OpenCLDebugInfo100DebugSource, inst, 7, ext_inst_name);
+        if (validate_src != SPV_SUCCESS) return validate_src;
+        break;
+      }
+      case OpenCLDebugInfo100DebugSource: {
+        auto validate_file = ValidateOperandForDebugInfo(
+            _, "File", SpvOpString, inst, 5, ext_inst_name);
+        if (validate_file != SPV_SUCCESS) return validate_file;
+        if (num_operands == 6) {
```
dneto0: This is a subtle problem.  This code uses num_operands, which is  inst->operands().size().
But then the helper function accesses inst->words(index).
The two vectors operands_ and words_ are different things. It's a happy accident that they have the same number of elements.  But it's confusing to use the count of one to control the indexing of the other.

If you're going to be accessing the word vector, then get a num_words variable from inst->words().size(), and use num_words consistently below.

The operands_ member if Instruction is not the same as the words

source/val/validate_extensions.cpp
```diff
@@ -2028,6 +2131,377 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
...
+        auto validate_src = ValidateDebugInfoOperand(
+            _, "Source", OpenCLDebugInfo100DebugSource, inst, 7, ext_inst_name);
+        if (validate_src != SPV_SUCCESS) return validate_src;
+        auto validate_parent =
+            ValidateOperandLexicalScope(_, "Parent", inst, 10, ext_inst_name);
+        if (validate_parent != SPV_SUCCESS) return validate_parent;
+        auto validate_size = ValidateOperandForDebugInfo(
+            _, "Size", SpvOpConstant, inst, 11, ext_inst_name);
+        if (validate_size != SPV_SUCCESS) return validate_size;
+        for (uint32_t word_index = 13; word_index <= num_operands;
```
dneto0: This is a buffer overflow risk.   This checks word_index <= num_operands but then the body accesses word_index+1.
The loop test should be word_index + 1 < num_words.

source/val/validate_extensions.cpp
```diff
@@ -2028,6 +2131,377 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
...
+             word_index += 2) {
+          auto validate_value = ValidateOperandForDebugInfo(
+              _, "Value", SpvOpConstant, inst, word_index, ext_inst_name);
+          if (validate_value != SPV_SUCCESS) return validate_value;
+          assert(word_index + 1 <= num_operands && "Missing \"Name\" operand");
```
dneto0: Don't assert.  Asserts are only checked in debug builds.  Instead avoid this case via the loop test.

source/val/validate_extensions.cpp
```diff
@@ -2028,6 +2131,377 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
...
+          validate_name = ValidateOperandForDebugInfo(
+              _, "Name", SpvOpString, inst, word_index + 1, ext_inst_name);
+          if (validate_name != SPV_SUCCESS) return validate_name;
+        }
+        break;
+      }
+      case OpenCLDebugInfo100DebugTypeComposite: {
+        auto validate_name = ValidateOperandForDebugInfo(
```
dneto0: This is a very repetitive code sequence.  The same things are repeatedly done:
  - call a helper method to check
  - check return status, and return the status if it's not SPV_SUCCESS
  - in these things, reuse inst and ext_inst_name.

I think you can shorten and clarify by using macros, e.g. near the top, define:
   #define CHECK_OPERAND(NAME, opcode, index)  { \
      auto result = ValidateperandForDebugInfo(_, NAME, opcode, inst, index, ext_inst_name); if (result != SPV_SUCCESS) return result;}

Then at this point you would only write this:
  CHECK_OPERAND("Name", SpvOpString, 5)


Please shorten all your code sequences. It will make the code far more readable and easier to verify.

jaebaek: done

dneto0:
Approved.  Thanks for your patience.

PR #3135, opened 2020-01-13
spirv-fuzz: add dead blocks

afd:
This adds a new kind of fact to the fact manager that knows whether a
block is dead - i.e. guaranteed to be statically unreachable - and a
new transformation for adding a selection construct to a CFG that
conditionally branches to a fresh, dead block, such that the branch
will never be dynamically taken.  Transformations that may create new
blocks ('split block' and 'outline function') are updated to propagate
dead block facts to newly-created blocks where appropriate.  A fuzzer
pass randomly adds dead blocks to the module.

Future transformations will be able to exploit the fact that such
blocks are known to be dead.

### DIFF
--- a/source/fuzz/fuzzer_pass_add_dead_blocks.cpp
+++ b/source/fuzz/fuzzer_pass_add_dead_blocks.cpp
@@ -1,0 +1,33 @@
+// Copyright (c) 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_add_dead_blocks.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassAddDeadBlocks::FuzzerPassAddDeadBlocks(
+    opt::IRContext* ir_context, FactManager* fact_manager,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, fact_manager, fuzzer_context, transformations) {}
+
+FuzzerPassAddDeadBlocks::~FuzzerPassAddDeadBlocks() = default;
+
+void FuzzerPassAddDeadBlocks::Apply() {
+  assert(false && "Implement");
+}
+
+}  // namespace fuzz
+}  // namespace spvtools
--- a/source/fuzz/fuzzer_pass_add_dead_blocks.cpp
+++ b/source/fuzz/fuzzer_pass_add_dead_blocks.cpp
@@ -14,6 +14,9 @@
 
 #include "source/fuzz/fuzzer_pass_add_dead_blocks.h"
 
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/transformation_add_dead_block.h"
+
 namespace spvtools {
 namespace fuzz {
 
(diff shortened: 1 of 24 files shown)


source/fuzz/fuzzer_pass_add_dead_blocks.cpp
```diff
@@ -0,0 +1,67 @@
...
+              GetIRContext(), block.terminator()->GetSingleWordInOperand(0))) {
+        continue;
+      }
+      candidate_blocks.push_back(&block);
+    }
+  }
+  while (!candidate_blocks.empty()) {
+    uint32_t index = GetFuzzerContext()->RandomIndex(candidate_blocks);
+    auto block = candidate_blocks.at(index);
+    candidate_blocks.erase(candidate_blocks.begin() + index);
```
paulthomson: Add a comment explaining why you do this in a random order?

source/fuzz/fuzzer_pass_add_dead_blocks.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void FuzzerPassAddDeadBlocks::Apply() {
+  std::vector<opt::BasicBlock*> candidate_blocks;
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      if (!GetFuzzerContext()->ChoosePercentage(
+              GetFuzzerContext()->GetChanceOfAddingDeadBlock())) {
+        continue;
+      }
+      if (block.IsLoopHeader()) {
```
paulthomson: Bit of a nit, but these checks are repeated in `IsApplicable` (with comments!). Maybe you could create the transformation (the `condition` doesn't matter at this stage; `fresh_id` would have to pass the check somehow...not ideal), check if it `IsApplicable`, then fix up the transformation with a randomly chosen condition and with a fresh id. OK, maybe this is more confusing. Could extract the common checks, but again maybe this is more confusing than a bit of repetition.

test/fuzz/fuzz_test_util.cpp
```diff
@@ -76,6 +76,7 @@ bool IsValid(spv_target_env env, const opt::IRContext* ir) {
   std::vector<uint32_t> binary;
   ir->module()->ToBinary(&binary, false);
   SpirvTools t(env);
+  t.SetMessageConsumer(kConsoleMessageConsumer);
```
paulthomson: This seems fine but I just want to check it is intentional.

source/fuzz/fuzzer_pass_add_dead_blocks.cpp
afd: I rewrote this to avoid the random order.

source/fuzz/fuzzer_pass_add_dead_blocks.cpp
afd: Actually I think you're right - we don't mind wasting some fresh ids, so I have written the code now so that it populates a list of transformations (speculatively), and then goes ahead and applies the ones that turn out to be applicable.  That indeed eliminates all this re-checking of applicability.  I have done that for some other fuzzer passes, and I think it's a good plan unless checking full applicability is particularly expensive.

test/fuzz/fuzz_test_util.cpp
afd: Thanks - yes it is.

source/fuzz/fuzzer_pass_add_dead_blocks.cpp
```diff
@@ -0,0 +1,67 @@
...
+        continue;
+      }
+      if (block.terminator()->opcode() != SpvOpBranch) {
+        continue;
+      }
+      if (fuzzerutil::IsMergeOrContinue(
+              GetIRContext(), block.terminator()->GetSingleWordInOperand(0))) {
```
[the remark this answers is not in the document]
afd: I'd say that the comment "Its successor ..." spells this out enough.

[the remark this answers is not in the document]
paulthomson: That comment wasn't there when I wrote this :-)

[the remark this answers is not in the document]
afd: Haha, sorry!

PR #3142, opened 2020-01-16
Add spvParseVulkanEnv

dneto0:
This new API lets clients request a minimal spv_target_env value
that supports a given Vulkan and SPIR-V version, by a generic
numbering scheme (as already defined by Vulkan and SPIR-V specs).

This breaks a formal source dependency from Glslang to SPIRV-Tools.
When a new API is rolled out, such as Vulkan 1.2, Glslang currently
needs to reference a specific SPIRV-Tools enum by name.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -473,6 +473,19 @@ SPIRV_TOOLS_EXPORT const char* spvTargetEnvDescription(spv_target_env env);
 // false and sets *env to SPV_ENV_UNIVERSAL_1_0.
 SPIRV_TOOLS_EXPORT bool spvParseTargetEnv(const char* s, spv_target_env* env);
 
+// Determines the target env value with the least features but which enables
+// the given Vulkan and SPIR-V versions. If such a target is supported, returns
+// true and writes the value to |env|, otherwise returns false.
+//
+// The Vulkan version is given as an unsigned 32-bit number as specified in
+// Vulkan section "29.2.1 Version Numbers": the major version number appears
+// in bits 22 to 21, and the minor version is in bits 12 to 21.  The SPIR-V
+// version is given in the SPIR-V version header word: major version in bits
+// 16 to 23, and minor version in bits 8 to 15.
+SPIRV_TOOLS_EXPORT bool spvParseVulkanEnv(uint32_t vulkan_ver,
+                                          uint32_t spirv_ver,
+                                          spv_target_env* env);
+
 // Creates a context object.  Returns null if env is invalid.
 SPIRV_TOOLS_EXPORT spv_context spvContextCreate(spv_target_env env);
 
--- a/source/spirv_target_env.cpp
+++ b/source/spirv_target_env.cpp
@@ -167,7 +167,6 @@ struct VulkanEnv {
 static const VulkanEnv ordered_vulkan_envs[] = {
     {SPV_ENV_VULKAN_1_0, VULKAN_VER(1, 0), SPIRV_VER(1, 0)},
     {SPV_ENV_VULKAN_1_1, VULKAN_VER(1, 1), SPIRV_VER(1, 3)},
-    {SPV_ENV_VULKAN_1_1, VULKAN_VER(1, 1), SPIRV_VER(1, 3)},
     {SPV_ENV_VULKAN_1_1_SPIRV_1_4, VULKAN_VER(1, 1), SPIRV_VER(1, 4)},
     {SPV_ENV_VULKAN_1_2, VULKAN_VER(1, 2), SPIRV_VER(1, 5)}};
 
--- a/source/spirv_target_env.cpp
+++ b/source/spirv_target_env.cpp
@@ -153,6 +153,35 @@ bool spvParseTargetEnv(const char* s, spv_target_env* env) {
   return false;
 }
 
+#define VULKAN_VER(MAJOR, MINOR) ((MAJOR << 22) | (MINOR << 12))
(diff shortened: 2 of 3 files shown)


dneto0:
cc: @johnkslang 
No need to rush this into Glslang.  But this should enable breaking the formal source dependency from Glslang to SPIRV-Tools.

source/spirv_target_env.cpp
```diff
@@ -153,6 +153,35 @@ bool spvParseTargetEnv(const char* s, spv_target_env* env) {
...
+  spv_target_env vulkan_env;
+  uint32_t vulkan_ver;
+  uint32_t spirv_ver;
+};
+// Maps each Vulkan target environment enum to the Vulkan version, and the
+// maximum supported SPIR-V version for that Vulkan environment.
+// Keep this ordered from least capable to most capable.
+static const VulkanEnv ordered_vulkan_envs[] = {
+    {SPV_ENV_VULKAN_1_0, VULKAN_VER(1, 0), SPIRV_VER(1, 0)},
+    {SPV_ENV_VULKAN_1_1, VULKAN_VER(1, 1), SPIRV_VER(1, 3)},
```
alan-baker: This row is duplicated.

dneto0: Fixed, thanks

test/target_env_test.cpp
```diff
@@ -103,5 +105,61 @@ INSTANTIATE_TEST_SUITE_P(
...
+        {VK(1, 1), SPV(1, 4), true, SPV_ENV_VULKAN_1_1_SPIRV_1_4},
+        {VK(1, 1), SPV(1, 5), true, SPV_ENV_VULKAN_1_2},
+        {VK(1, 1), SPV(1, 6), false, SPV_ENV_UNIVERSAL_1_0},
+        // Vulkan 1.2 cases
+        {VK(1, 2), SPV(1, 0), true, SPV_ENV_VULKAN_1_2},
+        {VK(1, 2), SPV(1, 1), true, SPV_ENV_VULKAN_1_2},
+        {VK(1, 2), SPV(1, 2), true, SPV_ENV_VULKAN_1_2},
+        {VK(1, 2), SPV(1, 3), true, SPV_ENV_VULKAN_1_2},
+        {VK(1, 2), SPV(1, 4), true, SPV_ENV_VULKAN_1_2},
+        {VK(1, 2), SPV(1, 5), true, SPV_ENV_VULKAN_1_2},
```
johnkslang: So, what I was confused about before was that sometimes the SPIR-V version is in the context enum and sometimes it is not.

If the validation/etc. re-adapts to whatever version is actually within the SPIR-V module, then I don't understand why it's ever part of the context enum.

On the other hand, if the validation/etc. does not adapt to the version in the SPIR-V module, then I don't understand how the validator will operate differently for the different combinations of a given Vulkan release but with different SPIR-V versions.

johnkslang: From an interface perspective, I guess what's important to know is which path triggers the SPIR-V version-specific validation: the value passed in the this new entry point, or the content of the module.

include/spirv-tools/libspirv.h
```diff
@@ -473,6 +473,19 @@ SPIRV_TOOLS_EXPORT const char* spvTargetEnvDescription(spv_target_env env);
...
+// Determines the target env value with the least features but which enables
+// the given Vulkan and SPIR-V versions. If such a target is supported, returns
+// true and writes the value to |env|, otherwise returns false.
+//
+// The Vulkan version is given as an unsigned 32-bit number as specified in
+// Vulkan section "29.2.1 Version Numbers": the major version number appears
+// in bits 22 to 21, and the minor version is in bits 12 to 21.  The SPIR-V
+// version is given in the SPIR-V version header word: major version in bits
+// 16 to 23, and minor version in bits 8 to 15.
+SPIRV_TOOLS_EXPORT bool spvParseVulkanEnv(uint32_t vulkan_ver,
```
johnkslang: What should glslang do for OpenGL contexts?

Currently, glslang also has paths for getting a context for non-Vulkan contexts.  Keep that code as is?  Or, generalize this to something like (pseudo code):

```
enum clientAPI {
    Vulkan,
    OpenGL
};
SPIRV_TOOLS_EXPORT bool spvParseEnv(clientAPI client, ...
```

test/target_env_test.cpp
alan-baker: If we could go back in time, separating the binary version from the target environment would be a good thing. Validation should be considering the binary version of the module for SPIR-V rules and using the target environment for the client API rules.

include/spirv-tools/libspirv.h
dneto0: Yeah, I wondered about OpenGL (and didn't look!).

I can't predict how versions will be described for upcoming APIs. It happens that Vulkan and OpenGL have conventions that fit into a 32-bit int.  I have no idea how WebGPU would fare.  So my preference is to have a separate entry point for each API.  Adding one more for OpenGL is low cost and keeps the parameter descriptions clear.

I'll ruminate over the weekend over this.

johnkslang: Yes, separate entry points is better.

test/target_env_test.cpp
johnkslang: So, don't we have a way forward by deprecating binary version from the context enum, and saying from now on that comes from the SPIR-V module?  In which case, the entry point just becomes one of saying which version of the client target to use.

dneto0:
Ok.  It seems this API is a net win.  I can do the OpenGL entry point separately.

PR #3143, opened 2020-01-20
Add GN public config for spirv-headers

ShabbyX:
Allows using spirv-headers without necessarily using spirv-tools.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -324,12 +324,19 @@ config("spvtools_internal_config") {
   }
 }
 
+config("spv_headers_public_config") {
+  include_dirs = [ "${spirv_headers}/include" ]
+}
+
 source_set("spv_headers") {
   sources = [
     "$spirv_headers/include/spirv/1.2/GLSL.std.450.h",
     "$spirv_headers/include/spirv/unified1/OpenCL.std.h",
     "$spirv_headers/include/spirv/unified1/spirv.h",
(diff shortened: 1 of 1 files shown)


dj2:
This seems like it should be in a spirv-headers build.gn instead of spirv-tools?

ShabbyX:
@dj2, There is no spirv-headers/BUILD.gn. It's true that I'm hijacking spirv-tools for this purpose. But currently there is no reason to implement a full-fledged BUILD.gn for spirv-headers just to get this out of it (and I don't think writing a non functional BUILD.gn for spirv-headers that just exposes this is a great idea).

dj2:
Why would the build.gn for spirv-headers be non-functional? I'm guessing what we have for headers in here would move over to that build.gn file?

ShabbyX:
Because I don't intend to write a BUILD.gn for spirv-headers that actually builds things.

Or maybe there's nothing to build in spirv-headers?

dj2:
I don't think headers would have to actually generate anything, just expose the deps. The CMake file is similar in that it provides the include directories and a bunch of stuff for install, but build.gn doesn't need the install stuff. We also wouldn't care about the examples. So, just the deps export.

ShabbyX:
Ok, done: https://github.com/KhronosGroup/SPIRV-Headers/pull/140

Once that lands, you are right, we can simplify things in `BUILD.gn` of spirv-tools.

PR #3144, opened 2020-01-21
spirv-fuzz: Refactoring and type-related fixes

afd:
This change refactors some code for walking access chain indexes to
make it mirror the structure of other code (to improve readability in
the first instance and potentially enable a future refactoring to
extract common code), and fixes a problem related to module donation
and function types.

### DIFF
--- a/source/fuzz/fuzzer_util.cpp
+++ b/source/fuzz/fuzzer_util.cpp
@@ -362,7 +362,7 @@ uint32_t FindFunctionType(opt::IRContext* ir_context,
       // Not a match: different numbers of arguments.
       continue;
     }
-    // Check whether the return type and argument types match (after remapping).
+    // Check whether the return type and argument types match.
     bool input_operands_match = true;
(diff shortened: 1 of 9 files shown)


source/fuzz/fuzzer_util.cpp
```diff
@@ -347,6 +350,35 @@ bool IsMergeOrContinue(opt::IRContext* ir_context, uint32_t block_id) {
...
+  for (auto& type_or_value : ir_context->types_values()) {
+    if (type_or_value.opcode() != SpvOpTypeFunction) {
+      // We are only interested in function types.
+      continue;
+    }
+    if (type_or_value.NumInOperands() != type_ids.size()) {
+      // Not a match: different numbers of arguments.
+      continue;
+    }
+    // Check whether the return type and argument types match (after remapping).
```
paulthomson: Remove "after remapping"?

afd: Good catch, thanks!  (Factored this out from some code where there was remapping.)

PR #3150, opened 2020-01-23
Clarify mapping of target env to SPIR-V version

dneto0:
It depends on the API method.

### DIFF
--- a/include/spirv-tools/libspirv.hpp
+++ b/include/spirv-tools/libspirv.hpp
@@ -312,7 +312,12 @@ class SpirvTools {
   // client API (such as Vulkan 1.1), then validate for that client API version,
   // to the extent that it is verifiable from data in the binary itself.
   bool Validate(const std::vector<uint32_t>& binary) const;
+  // Like the previous overload, but provides the binary as a pointer and size:
   // |binary_size| specifies the number of words in |binary|.
+  // Validates for SPIR-V spec rules for the SPIR-V version named in the binary's
+  // header (at word offset 1).  Additionally, if the target environment is a
+  // client API (such as Vulkan 1.1), then validate for that client API version,
+  // to the extent that it is verifiable from data in the binary itself.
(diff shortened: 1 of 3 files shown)


dneto0:
cc: @johnkslang

include/spirv-tools/libspirv.hpp
```diff
@@ -300,10 +307,19 @@ class SpirvTools {
...
   // registered.
+  // Validates for SPIR-V spec rules for the SPIR-V version named in the binary's
+  // header (at word offset 1).  Additionally, if the target environment is a
+  // client API (such as Vulkan 1.1), then validate for that client API version,
+  // to the extent that it is verifiable from data in the binary itself.
   bool Validate(const std::vector<uint32_t>& binary) const;
   // |binary_size| specifies the number of words in |binary|.
   bool Validate(const uint32_t* binary, size_t binary_size) const;
   // Like the previous overload, but takes an options object.
+  // Validates for SPIR-V spec rules for the SPIR-V version named in the binary's
```
alan-baker: A little odd to repeat this here, but not in the previous overload.

dneto0: Right. The previous one should say "like the previous one..." as well.  Anyway, I duplicated the text.

dneto0:
> No changes were necessary in optimizer.hpp? The optimizer runs validation by default.

Right, I skipped that.  I added a commit that I think captures the optimizer semantics.  Hey @zoddicus I'm fuzzy on the behaviour of which validations change when setting the environment after optimizer object creation (used for WebGPU to Vulkan and vice versa).  Please verify that the comments are right.

zoddicus:
I assume this blocked by https://github.com/google/shaderc/pull/973?

dneto0:
> I assume this blocked by google/shaderc#973?

No, this is an independent stream.  This is only a doc update to (better) explain handling of target env.

PR #3151, opened 2020-01-24
Use dummy switch instead of dummy loop in MergeReturn pass.

### LINKED ISSUE #3127: Improve MergeReturn to not add loop when function does not contain one
Currently MergeReturn adds a loop to the function. This can be disadvantageous for loops containing such (inlined) functions since LoopUnrolling will not unroll loops which contain inner loops. 

Propose that the MergeReturn algorithm be improved so that it does not add a loop to functions which do not already contain a loop.

greg-lunarg:
Fixes #3127


greg-lunarg:
The shaderc-smoketest seems to not like SPV_ENV_VULKAN_1_2

PR #3160, opened 2020-01-28
Dead branch elim fix

s-perron:
We must treat a branch to the merge node of a switch that is in the
header of a construct as a nested construced.  The original merge
instruction is still needed in that case.

### DIFF
--- a/source/opt/dead_branch_elim_pass.cpp
+++ b/source/opt/dead_branch_elim_pass.cpp
@@ -167,7 +167,6 @@ bool DeadBranchElimPass::MarkLiveBlocks(
     }
 
     if (simplify) {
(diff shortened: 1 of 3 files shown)


alan-baker:
@ShabbyX I think this should address the underlying issue #3151 exposed.

ShabbyX:
I tested it locally, and yes the test is passing now. I don't know if this would break anything else.

ShabbyX:
Could we land this then so I can roll it in Chromium?

alan-baker:
@ShabbyX thanks for your patience, sorry for the break.

ShabbyX:
Thank you for the fix!

PR #3165, opened 2020-01-29
spirv-fuzz: Arbitrary variable facts

afd:
This change adds a new kind of fact to the fact manager, which records
when a variable (or pointer parameter) refers to an arbitrary value,
so that anything can be stored to it, without affecting the observable
behaviour of the module, and nothing can be guaranteed about values
loaded from it.  Donated modules are the current source of such
variables, and other transformations, such as outlining, have been
adapted to propagate these facts appropriately.

### DIFF
--- a/test/fuzz/transformation_add_function_test.cpp
+++ b/test/fuzz/transformation_add_function_test.cpp
@@ -58,6 +58,56 @@ std::vector<protobufs::Instruction> GetInstructionsForFunction(
   return result;
 }
 
+// Returns true if and only if every pointer parameter and variable associated
+// with |function_id| in |context| is known by |fact_manager| to be arbitrary,
+// with the exception of |loop_limiter_id|, which must not be arbitrary.  (It
+// can be 0 if no loop limiter is expected, and 0 should not be deemed
+// arbitrary).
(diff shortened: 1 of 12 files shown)


test/fuzz/transformation_add_function_test.cpp
```diff
@@ -58,6 +58,56 @@ std::vector<protobufs::Instruction> GetInstructionsForFunction(
...
 }
 
+// Returns true if and only if every pointer parameter and variable associated
+// with |function_id| in |context| is known by |fact_manager| to be arbitrary,
+// with the exception of |loop_limiter_id|, which must not be arbitrary.  (It
+// can be 0 if no loop limiter is expected, and 0 should not be deemed
+// arbitrary).
+bool AllVariablesAndParametersExceptLoopLimiterAreArbitrary(
+    opt::IRContext* context, const FactManager& fact_manager,
+    uint32_t function_id, uint32_t loop_limiter_id) {
```
paulthomson: Could mention that you don't have tests with multiple loops? Hence you only need one loop_limiter_id

afd: There is actually only one loop limiter per function.

PR #3166, opened 2020-01-29
Implement constant folding for many transcendentals

### LINKED ISSUE #1390: Missing fp const folding for std.450 extended instructions
We have some shaders that have instructions from the extended set that could be folded but aren't, specifically:

- FMin/FMax
- Pow/Exp/Log/Log2
- Sin/Cos

(I won't have time to look into this up until after GDC so filing an issue for now, maybe somebody gets to this first!)

zeux:
This change adds support for folding of sin/cos/tan/asin/acos/atan,
exp/log/exp2/log2, sqrt, atan2 and pow.

The mechanism allows to use any C function to implement folding in the
future; for now I limited the actual additions to the most commonly used
intrinsics in the shaders.

Unary folder had to be tweaked to work with extended instructions - for
extended instructions, constants.size() == 2 and constants[0] ==
nullptr. This adjustment is similar to the one binary folder already
performs.

Fixes #1390.

### DIFF
--- a/source/opt/const_folding_rules.cpp
+++ b/source/opt/const_folding_rules.cpp
@@ -265,7 +265,10 @@ ConstantFoldingRule FoldFPUnaryOp(UnaryScalarFoldingRule scalar_rule) {
       return nullptr;
     }
 
-    if (constants[0] == nullptr) {
+    const analysis::Constant* arg =
+        (inst->opcode() == SpvOpExtInst) ? constants[1] : constants[0];
+
+    if (arg == nullptr) {
       return nullptr;
     }
 
@@ -273,7 +276,7 @@ ConstantFoldingRule FoldFPUnaryOp(UnaryScalarFoldingRule scalar_rule) {
       std::vector<const analysis::Constant*> a_components;
       std::vector<const analysis::Constant*> results_components;
 
-      a_components = constants[0]->GetVectorComponents(const_mgr);
+      a_components = arg->GetVectorComponents(const_mgr);
 
       // Fold each component of the vector.
       for (uint32_t i = 0; i < a_components.size(); ++i) {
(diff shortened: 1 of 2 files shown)


zeux:
Apparently Android's standard library implementation used by the build doesn't have std::exp2 :-/ I think this is because it's using a (deprecated) libstdc++ library instead of libc++...

I'm going to change exp2/log2 to use global (C99) versions.

zeux:
This didn't solve this completely :(

The build looks like it's using r15c NDK and targets Android 15 ABI (4.1).
I've downloaded r15c NDK and somehow it has exp2 but not log2 in that ABI. It only has log2 starting from Android 18 (4.3).

I'm surprised this still targets the toolchain/ABI this old, but I could implement a shim that uses `log(value) / log(2.0)` for Android which should hopefully fix this.

zeux:
clang-debug build timed out *during building* so probably unrelated to this change; at least Android is green now 😅

s-perron:
macos builds passed earlier, so that is not an issue.  I just want to take a little time to see if it is possible to avoid the if-def that was added in the middle of `AddFoldingRules`.  It is a bit of an eyesore.

zeux:
> macos builds passed earlier, so that is not an issue. I just want to take a little time to see if it is possible to avoid the if-def that was added in the middle of AddFoldingRules. It is a bit of an eyesore.

Sure, but it's a necessity. I don't know why the Android ABI that is used is so low - if the ABI can be bumped, we can use global exp2/log2 unconditionally. If the Android build stops using deprecated STL implementation and becomes C++11 compliant, we can use C++11 std::exp2/std::log2.

Alternatively we can *always* use ::exp2 and log2 emulation but it seems wrong to do this because surely, eventually Android builds will migrate to modern versions of the toolchain.

s-perron:
I'll fix this up later.  I know we cannot remove the if-def completely.  However, it could be moved somewhere out of the way.

PR #3169, opened 2020-02-03
vscode: Add missing fields to schema.Opcode

ben-clayton:
These were declared in the `Opcode` struct type, but were never assigned.
They're not actually used by the language server, but I reused this go schema package for a local experiment and found they were missing.


ben-clayton:
Note: I do not have permissions to merge this myself.

PR #3171, opened 2020-02-04
spirv-fuzz: Fuzzer pass to add composite types

afd:
Adds a fuzzer pass that randomly adds vector and matrix types not
already present in the module, and randomly adds structs with random
field types and arrays with random base types and sizes.  Other passes
will be able to create variables and ids using these types.

### DIFF
--- a/source/fuzz/fuzzer_pass_add_composite_types.cpp
+++ b/source/fuzz/fuzzer_pass_add_composite_types.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Google LLC
+// Copyright (c) 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -33,7 +33,7 @@ void FuzzerPassAddCompositeTypes::Apply() {
   MaybeAddMissingVectorTypes();
   MaybeAddMissingMatrixTypes();
 
-  // (3) Randomly interleave between adding struct and array composite types
+  // Randomly interleave between adding struct and array composite types
   while (GetFuzzerContext()->ChoosePercentage(
       GetFuzzerContext()->GetChanceOfAddingArrayOrStructType())) {
     if (GetFuzzerContext()->ChoosePercentage(
@@ -119,6 +119,7 @@ uint32_t FuzzerPassAddCompositeTypes::ChooseScalarOrCompositeType() {
       case SpvOpTypeFloat:
       case SpvOpTypeInt:
       case SpvOpTypeMatrix:
+      case SpvOpTypeStruct:
       case SpvOpTypeVector:
         candidates.push_back(inst.result_id());
         break;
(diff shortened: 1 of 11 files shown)


source/fuzz/fuzzer_pass_add_composite_types.cpp
```diff
@@ -0,0 +1,137 @@
+// Copyright (c) 2019 Google LLC
```
paulthomson: Date?

source/fuzz/fuzzer_pass_add_composite_types.h
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2019 Google LLC
```
paulthomson: Date?

source/fuzz/fuzzer_pass_add_composite_types.cpp
```diff
@@ -0,0 +1,137 @@
...
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, fact_manager, fuzzer_context, transformations) {}
+
+FuzzerPassAddCompositeTypes::~FuzzerPassAddCompositeTypes() = default;
+
+void FuzzerPassAddCompositeTypes::Apply() {
+  MaybeAddMissingVectorTypes();
+  MaybeAddMissingMatrixTypes();
+
+  // (3) Randomly interleave between adding struct and array composite types
```
paulthomson: (3) ?

source/fuzz/fuzzer_pass_add_composite_types.cpp
```diff
@@ -0,0 +1,137 @@
...
+      GetFuzzerContext()->GetChanceOfAddingAnotherStructField()));
+  ApplyTransformation(TransformationAddTypeStruct(
+      GetFuzzerContext()->GetFreshId(), field_type_ids));
+}
+
+uint32_t FuzzerPassAddCompositeTypes::ChooseScalarOrCompositeType() {
+  // Gather up all the possibly-relevant types.
+  std::vector<uint32_t> candidates;
+  for (auto& inst : GetIRContext()->types_values()) {
+    switch (inst.opcode()) {
```
paulthomson: Could you have struct types here?

source/fuzz/fuzzer_pass_add_composite_types.h
afd: Fixed, thanks.

source/fuzz/fuzzer_pass_add_composite_types.cpp
afd: Fixed, thanks.

source/fuzz/fuzzer_pass_add_composite_types.cpp
afd: Oops (a remnant from before I refactored) - fixed.

source/fuzz/fuzzer_pass_add_composite_types.cpp
afd: Amazing catch, thanks - it was absolutely my intention to have structs.

PR #3174, opened 2020-02-04
Make spvOpcodeString part of the public API

### LINKED ISSUE #3138: Make spvOpcodeString() part of the public API
In https://swiftshader-review.googlesource.com/c/SwiftShader/+/38820 we added a forward declaration of spvOpcodeString() to our code base. Adding it to the public API of SPIRV-Tools would allow it to be used more widely and help avoid breakage.

s-perron:
Fixes #3138

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -766,6 +766,9 @@ SPIRV_TOOLS_EXPORT void spvDiagnosticDestroy(spv_diagnostic diagnostic);
 SPIRV_TOOLS_EXPORT spv_result_t
 spvDiagnosticPrint(const spv_diagnostic diagnostic);
 
+// Gets the name of an instruction, without the "Op" prefix.
+SPIRV_TOOLS_EXPORT const char* spvOpcodeString(const uint32_t opcode);
+
 // The binary parser interface.
 
 // A pointer to a function that accepts a parsed SPIR-V header.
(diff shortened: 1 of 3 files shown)


include/spirv-tools/libspirv.h
```diff
@@ -24,6 +24,8 @@ extern "C" {
 #include <stddef.h>
 #include <stdint.h>
 
+#include <spirv/unified1/spirv.h>
```
dneto0: This part makes me sad.  It's suboptimal to add a new project / include-dir dependency for this one feature.

include/spirv-tools/libspirv.h
```diff
@@ -766,6 +768,9 @@ SPIRV_TOOLS_EXPORT void spvDiagnosticDestroy(spv_diagnostic diagnostic);
 SPIRV_TOOLS_EXPORT spv_result_t
 spvDiagnosticPrint(const spv_diagnostic diagnostic);
 
+// Gets the name of an instruction, without the "Op" prefix.
+SPIRV_TOOLS_EXPORT const char* spvOpcodeString(const SpvOp opcode);
```
dneto0: To break the dependency on SpvOp, can we fudge it with uint32_t like this:
 
 // ...
 // Use a uint32_t to avoid a public header dependency on spirv.h
 ....  spvOpcodeStringFromOpcode(uint32_t opcode_value);

s-perron: Good idea.  I've tried it and it works.

PR #3175, opened 2020-02-05
spirv-fuzz: Fuzzer passes to add local and global variables

afd:
Adds two new fuzzer passes to add variables to a module: one that adds
Private storage class global variables, another that adds Function
storage class local variables.

### DIFF
--- a/source/fuzz/fuzzer_pass.cpp
+++ b/source/fuzz/fuzzer_pass.cpp
@@ -297,11 +297,11 @@ std::pair<std::vector<uint32_t>, std::map<uint32_t, std::vector<uint32_t>>>
 FuzzerPass::GetAvailableBaseTypesAndPointers(
     SpvStorageClass storage_class) const {
   // Records all of the base types available in the module.
-  std::vector<uint32_t> base_type_ids;
+  std::vector<uint32_t> base_types;
 
   // For each base type, records all the associated pointer types that target
   // that base type and that have |storage_class| as their storage class.
-  std::map<uint32_t, std::vector<uint32_t>> base_type_to_pointer;
+  std::map<uint32_t, std::vector<uint32_t>> base_type_to_pointers;
 
   for (auto& inst : GetIRContext()->types_values()) {
     switch (inst.opcode()) {
@@ -315,14 +315,14 @@ FuzzerPass::GetAvailableBaseTypesAndPointers(
         // These types are suitable as pointer base types.  Record the type,
         // and the fact that we cannot yet have seen any pointers that use this
         // as its base type.
-        base_type_ids.push_back(inst.result_id());
-        base_type_to_pointer.insert({inst.result_id(), {}});
+        base_types.push_back(inst.result_id());
+        base_type_to_pointers.insert({inst.result_id(), {}});
         break;
       case SpvOpTypePointer:
         if (inst.GetSingleWordInOperand(0) == storage_class) {
           // The pointer has the desired storage class, so we are interested in
           // it.  Associate it with its base type.
-          base_type_to_pointer.at(inst.GetSingleWordInOperand(1))
+          base_type_to_pointers.at(inst.GetSingleWordInOperand(1))
               .push_back(inst.result_id());
         }
         break;
@@ -330,7 +330,7 @@ FuzzerPass::GetAvailableBaseTypesAndPointers(
         break;
     }
   }
-  return {base_type_ids, base_type_to_pointer};
+  return {base_types, base_type_to_pointers};
 }
 
 uint32_t FuzzerPass::FindOrCreateZeroConstant(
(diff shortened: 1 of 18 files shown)


source/fuzz/fuzzer_pass.cpp
```diff
@@ -254,5 +293,152 @@ uint32_t FuzzerPass::FindOrCreateGlobalUndef(uint32_t type_id) {
...
+            GetIRContext()->get_constant_mgr()->GetConstantFromInst(
+                GetIRContext()->get_def_use_mgr()->GetDef(field_constant_id)));
+      }
+      return FindOrCreateCompositeConstant(
+          *type_instruction, field_zero_constants, field_zero_ids);
+    }
+    case SpvOpTypeVector: {
+      return GetZeroConstantForHomogeneousComposite(
+          *type_instruction, type_instruction->GetSingleWordInOperand(0),
+          type_instruction->GetSingleWordInOperand(1));
```
paulthomson: Nit: this looks identical to `case SpvOpTypeMatrix:` above. Could combine?

source/fuzz/fuzzer_pass.cpp
```diff
@@ -254,5 +293,152 @@ uint32_t FuzzerPass::FindOrCreateGlobalUndef(uint32_t type_id) {
...
+
+uint32_t FuzzerPass::GetZeroConstantForHomogeneousComposite(
+    const opt::Instruction& composite_type_instruction,
+    uint32_t component_type_id, uint32_t num_components) {
+  std::vector<const opt::analysis::Constant*> zero_constants;
+  std::vector<uint32_t> zero_ids;
+  uint32_t zero_component = FindOrCreateZeroConstant(component_type_id);
+  const opt::analysis::Constant* registered_zero_component =
+      GetIRContext()->get_constant_mgr()->GetConstantFromInst(
+          GetIRContext()->get_def_use_mgr()->GetDef(zero_component));
```
paulthomson: I don't quite get why GetDef is needed here, but maybe I would if I had the code checked out.

source/fuzz/fuzzer_pass_add_local_variables.cpp
```diff
@@ -0,0 +1,63 @@
...
+    while (GetFuzzerContext()->ChoosePercentage(
+        GetFuzzerContext()->GetChanceOfAddingLocalVariable())) {
+      uint32_t base_type_id =
+          base_type_ids[GetFuzzerContext()->RandomIndex(base_type_ids)];
+      uint32_t pointer_type_id;
+      std::vector<uint32_t>& available_pointers =
+          base_type_to_pointer.at(base_type_id);
+      if (available_pointers.empty()) {
+        pointer_type_id = GetFuzzerContext()->GetFreshId();
+        available_pointers.push_back(pointer_type_id);
```
paulthomson: Might want to add a comment that this modifies the base_type_to_pointer map. This ensure that future loop iterations know that this type exists; it is not useful for this iteration. Could also rename it to available_pointers_to_base_type.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -516,6 +517,31 @@ message TransformationAddGlobalVariable {
...
+  // The id of the function to which the local variable should be added
+  uint32 function_id = 3;
+
+  // Initial value of the variable
+  uint32 initializer_id = 4;
+
+  // True if and only if the value of the variable should be regarded, in
+  // general, as totally unknown and subject to change.  Stores to such
+  // variables can be performed in an arbitrary fashion.
+  bool value_is_arbitrary = 5;
```
paulthomson: Could clarify that this is not like OpUndef (I think). It is more like: the fuzzer may add writes to this variable at any point.

source/fuzz/fuzzer_pass.cpp
paulthomson: Oh it is because you need an Instruction*, right?

source/fuzz/transformation_add_local_variable.cpp
```diff
@@ -0,0 +1,105 @@
...
+  switch (initializer_instruction->opcode()) {
+    case SpvOpConstant:
+    case SpvOpConstantComposite:
+    case SpvOpConstantFalse:
+    case SpvOpConstantNull:
+    case SpvOpConstantTrue:
+      break;
+    default:
+      return false;
+  }
```
paulthomson: spvOpcodeIsConstant exists and has more cases, but perhaps that is fine.

source/fuzz/fuzzer_pass.cpp
afd: Done.

source/fuzz/protobufs/spvtoolsfuzz.proto
afd: Done.

source/fuzz/transformation_add_local_variable.cpp
afd: Done (there is also SpvOpcodeIsComposite, which I should probably use elsewhere in the code base ...).

source/fuzz/fuzzer_pass_add_local_variables.cpp
afd: Good idea - done (and similar in fuzzer_pass_add_global_variables.cpp)

PR #3178, opened 2020-02-07
spirv-fuzz: Fuzzer pass to add function calls

afd:
Adds a fuzzer pass that inserts function calls into the module at
random.  Calls from dead blocks can be arbitrary (so long as they do
not introduce recursion), while calls from other blocks can only be to
livesafe functions.

The change fixes some oversights in transformations to replace
constants with uniforms and to obfuscate constants which testing of
this fuzzer pass identified.

### DIFF
--- a/source/fuzz/fuzzer_pass_add_function_calls.cpp
+++ b/source/fuzz/fuzzer_pass_add_function_calls.cpp
@@ -16,7 +16,8 @@
 
 #include "source/fuzz/call_graph.h"
 #include "source/fuzz/fuzzer_util.h"
-#include "source/fuzz/transformation_add_global_undef.h"
+#include "source/fuzz/transformation_add_global_variable.h"
+#include "source/fuzz/transformation_add_local_variable.h"
 #include "source/fuzz/transformation_function_call.h"
 
 namespace spvtools {
@@ -38,8 +39,8 @@ void FuzzerPassAddFunctionCalls::Apply() {
           -> void {
         // Check whether it is legitimate to insert a function call before the
         // instruction.
-        if (!fuzzerutil::CanInsertOpcodeBeforeInstruction(
-                SpvOpFunctionCall, inst_it)) {
+        if (!fuzzerutil::CanInsertOpcodeBeforeInstruction(SpvOpFunctionCall,
+                                                          inst_it)) {
           return;
         }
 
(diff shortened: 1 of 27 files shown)


source/fuzz/fuzzer_pass_add_function_calls.cpp
```diff
@@ -0,0 +1,246 @@
+// Copyright (c) 2019 Google LLC
```
paulthomson: Date?

source/fuzz/fuzzer_pass_add_function_calls.h
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) 2019 Google LLC
```
paulthomson: Date?

source/fuzz/transformation_function_call.cpp
```diff
@@ -0,0 +1,207 @@
+// Copyright (c) 2019 Google LLC
```
paulthomson: Date?

source/fuzz/transformation_function_call.cpp
```diff
@@ -0,0 +1,207 @@
...
+    if (!arg_inst->type_id()) {
+      // The given argument does not have a type; it is thus not suitable.
+    }
+    opt::Instruction* arg_type_inst =
+        context->get_def_use_mgr()->GetDef(arg_inst->type_id());
+    if (arg_type_inst->result_id() !=
+        callee_type_inst->GetSingleWordInOperand(arg_index + 1)) {
+      // Argument type mismatch.
+      return false;
+    }
```
paulthomson: Nit: Could you get the `arg_type_inst` later and just use `arg_inst->type_id()` here

source/fuzz/transformation_function_call.cpp
```diff
@@ -0,0 +1,207 @@
...
+        default:
+          // Other pointer ids cannot be passed as parameters
+          return false;
+      }
+      if (!block_is_dead &&
+          !fact_manager.PointeeValueIsIrrelevant(arg_inst->result_id())) {
+        // This is not a dead block, so pointer parameters passed to the called
+        // function might really have their contents modified. We thus require
+        // such pointers to be to arbitrary-valued variables, which this is not.
+        return false;
```
paulthomson: I keep forgetting exactly when we should use assert. Should we assert(false) here to catch cases where we violate this due to bugs in spirv-fuzz? Should we add a special type of assert for testing?

source/fuzz/transformation_function_call.cpp
```diff
@@ -0,0 +1,207 @@
...
+      }
+    }
+
+    if (arg_inst->opcode() == SpvOpFunctionParameter) {
+      bool found = false;
```
paulthomson: Could add a comment for this case.

source/fuzz/transformation_function_call.h
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2019 Google LLC
```
paulthomson: Date

source/fuzz/transformation_replace_constant_with_uniform.cpp
```diff
@@ -154,6 +154,12 @@ bool TransformationReplaceConstantWithUniform::IsApplicable(
     return false;
   }
 
+  // The use must not be a variable inializer; these are required to be
```
paulthomson: Typo: inializer -> initializer

test/fuzz/transformation_function_call_test.cpp
```diff
@@ -0,0 +1,443 @@
+// Copyright (c) 2019 Google LLC
```
paulthomson: Date

source/fuzz/fuzzer_pass_add_function_calls.cpp
```diff
@@ -0,0 +1,246 @@
...
+  std::vector<uint32_t> result;
+  for (uint32_t arg_index = 1; arg_index < function_type->NumInOperands();
+       arg_index++) {
+    auto arg_type_id =
+        GetIRContext()
+            ->get_def_use_mgr()
+            ->GetDef(function_type->GetSingleWordInOperand(arg_index))
+            ->result_id();
+    if (type_to_available_instructions.count(arg_type_id)) {
+      std::vector<opt::Instruction*> candidate_arguments =
```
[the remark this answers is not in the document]
afd: Done.

source/fuzz/fuzzer_pass_add_function_calls.cpp
```diff
@@ -0,0 +1,246 @@
...
+          type_to_available_instructions.at(arg_type_id);
+      // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3177) The value
+      //  selected here is arbitrary.  We should consider adding a fact to log
+      //  that it could be transformed.
+      result.push_back(candidate_arguments[GetFuzzerContext()->RandomIndex(
+                                               candidate_arguments)]
+                           ->result_id());
```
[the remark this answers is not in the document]
afd: Correct, and I've used your text in the comment.

source/fuzz/fuzzer_pass_add_function_calls.cpp
afd: Fixed, thanks.

source/fuzz/transformation_function_call.cpp
afd: Fixed, thanks.

source/fuzz/fuzzer_pass_add_function_calls.h
afd: Fixed, thanks.

source/fuzz/transformation_function_call.cpp
afd: Good idea, done.

source/fuzz/transformation_function_call.cpp
afd: We should only assert false in case that we are sure should never occur, even when transformations are removed by the shrinker.  In general I am keen to allow the shrinker to be a bit aggressive, perhaps leaving open the possibility for it to try changing transformations, so I am conservative about assert false.  But it would be good to discuss this more.

source/fuzz/transformation_function_call.h
afd: Fixed, thanks.

source/fuzz/transformation_function_call.cpp
afd: Actually it turned out that this and the code around it could be replaced with a call to fuzzerutil::IdIsAvailableBeforeInstruction() - thanks for flagging this up.

source/fuzz/transformation_replace_constant_with_uniform.cpp
afd: Thanks.

test/fuzz/transformation_function_call_test.cpp
afd: Fixed, thanks.

PR #3182, opened 2020-02-09
spirv-fuzz: Fuzzer pass that adds access chains

afd:
This change adds a fuzzer pass that sprinkles access chain
instructions into a module at random.  This allows other passes to
have a richer set of pointers available to them, in particular the
passes that add loads and stores.

### DIFF
--- a/source/fuzz/fuzzer_pass_add_access_chains.cpp
+++ b/source/fuzz/fuzzer_pass_add_access_chains.cpp
@@ -111,9 +111,13 @@ void FuzzerPassAddAccessChains::Apply() {
               break;
             default:
               assert(false && "Not a composite type opcode.");
+              // Set the bound to a value in order to keep release compilers
+              // happy.
+              bound = 0;
+              break;
           }
           if (bound == 0) {
-            // It is possible for a composite type to have zero
+            // It is possible for a composite type to legitimately have zero
             // sub-components, at least in the case of a struct, which
             // can have no fields.
             break;
--- a/source/fuzz/fuzzer_pass_add_access_chains.cpp
+++ b/source/fuzz/fuzzer_pass_add_access_chains.cpp
@@ -1,0 +1,31 @@
+// Copyright (c) 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_add_access_chains.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassAddAccessChains::FuzzerPassAddAccessChains(
+    opt::IRContext* ir_context, FactManager* fact_manager,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, fact_manager, fuzzer_context, transformations) {}
+
+FuzzerPassAddAccessChains::~FuzzerPassAddAccessChains() = default;
+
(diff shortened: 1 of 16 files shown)


source/fuzz/fuzzer_pass_add_access_chains.cpp
```diff
@@ -0,0 +1,162 @@
...
+FuzzerPassAddAccessChains::FuzzerPassAddAccessChains(
+    opt::IRContext* ir_context, FactManager* fact_manager,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, fact_manager, fuzzer_context, transformations) {}
+
+FuzzerPassAddAccessChains::~FuzzerPassAddAccessChains() = default;
+
+void FuzzerPassAddAccessChains::Apply() {
+  MaybeAddTransformationBeforeEachInstruction(
```
paulthomson: Maybe not in this PR, but should this method be renamed? Is it more like ForEachInstructionWithInstructionDescriptor?

source/fuzz/fuzzer_pass_add_access_chains.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+        std::vector<opt::Instruction*> relevant_pointer_instructions =
+            FindAvailableInstructions(
+                function, block, inst_it,
+                [](opt::IRContext* context,
+                   opt::Instruction* instruction) -> bool {
+                  if (!instruction->result_id() || !instruction->type_id()) {
+                    return false;
+                  }
+                  switch (instruction->result_id()) {
```
paulthomson: Should this be opcode?

source/fuzz/fuzzer_pass_add_access_chains.cpp
```diff
@@ -0,0 +1,162 @@
...
+                    case SpvOpUndef:
+                      // Do not allow making an access chain from a null or
+                      // undefined pointer.
+                      return false;
+                    default:
+                      break;
+                  }
+                  return context->get_def_use_mgr()
+                             ->GetDef(instruction->type_id())
+                             ->opcode() == SpvOpTypePointer;
```
paulthomson: Could add comment

source/fuzz/transformation_access_chain.cpp
```diff
@@ -0,0 +1,212 @@
...
+    return false;
+  }
+
+  // Do not allow making an access chain from a null or undefined pointer, as
+  // we do not want to allow accessing such pointers.  This might be acceptable
+  // in dead blocks, but we conservatively avoid it.
+  switch (pointer->opcode()) {
+    case SpvOpConstantNull:
+    case SpvOpUndef:
+      return false;
```
paulthomson: Perhaps this is another example where having some kind of "strict assert false" might help in finding bugs, like how I think you accidentally used ->result_id() earlier.

paulthomson: But I suppose a reducer could simplify pointers...but then we could disable such assertions.

test/fuzz/transformation_access_chain_test.cpp
```diff
@@ -0,0 +1,377 @@
+// Copyright (c) 2019 Google LLC
```
paulthomson: Date?

source/fuzz/fuzzer_pass_add_access_chains.cpp
afd: https://github.com/KhronosGroup/SPIRV-Tools/issues/3184

source/fuzz/fuzzer_pass_add_access_chains.cpp
afd: Done.

source/fuzz/fuzzer_pass_add_access_chains.cpp
afd: Good catch!

source/fuzz/transformation_access_chain.cpp
afd: For now I think we should assert false in cases where we don't want the reducer or fuzzer to generate certain transformations.  This is one of them, so I have added an assert false.

test/fuzz/transformation_access_chain_test.cpp
afd: Fixed.

PR #3183, opened 2020-02-10
Add data structures for DebugScope, DebugNoScope, DebugDeclare, DebugValue

jaebaek:
This commit adds data structures for rich debug info and lets ir_loader
and module correctly load/dump IR instructions.
- For debug instructions other than `DebugScope`, `DebugNoScope`,
`DebugDeclare`, `DebugValue`, we already added
`InstructionList ext_inst_debuginfo_` in `Module` class.
- For `DebugScope`/`DebugNoScope`, this commit adds
`struct DebugScope` in `Instruction` class.
  - When ir_loader processes `DebugScope`/`DebugNoScope`, we update
`last_dbg_scope_` of ir_loader to update the last `DebugScope`
information. When ir_loader processes a terminator instruction or merge
instruction, it sets `last_dbg_scope_` as `DebugNoScope` because it is the
end of the scope. When module dumps IR, if the next `DebugScope` is new
scope information, it generates a `DebugScope` instruction. If the next
`DebugScope` denotes the end of a scope, it generates a `DebugNoScope`
instruction.
- For `DebugDeclare`, this commit adds
`std::map<uint32_t, std::unique_ptr<Instruction>> local_var_to_dbg_decl_`
that is a map between each result id of local variable and its
corresponding `DebugDeclare`. When we remove a local variable, we can
remove `DebugDeclare` and add a new `DebugValue` in `InstructionList`
of basic block like a normal instruction.

### DIFF
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -116,7 +116,11 @@ Instruction::Instruction(Instruction&& that)
       unique_id_(that.unique_id_),
       operands_(std::move(that.operands_)),
       dbg_line_insts_(std::move(that.dbg_line_insts_)),
-      dbg_scope_(that.dbg_scope_) {}
+      dbg_scope_(that.dbg_scope_) {
+  for (auto& i : dbg_line_insts_) {
+    i.dbg_scope_ = that.dbg_scope_;
+  }
+}
 
 Instruction& Instruction::operator=(Instruction&& that) {
   opcode_ = that.opcode_;
@@ -126,6 +130,9 @@ Instruction& Instruction::operator=(Instruction&& that) {
   operands_ = std::move(that.operands_);
   dbg_line_insts_ = std::move(that.dbg_line_insts_);
   dbg_scope_ = that.dbg_scope_;
+  for (auto& i : dbg_line_insts_) {
+    i.dbg_scope_ = that.dbg_scope_;
+  }
   return *this;
 }
 
@@ -138,6 +145,9 @@ Instruction* Instruction::Clone(IRContext* c) const {
   clone->operands_ = operands_;
   clone->dbg_line_insts_ = dbg_line_insts_;
   clone->dbg_scope_ = dbg_scope_;
+  for (auto& i : clone->dbg_line_insts_) {
+    i.dbg_scope_ = dbg_scope_;
+  }
   return clone;
 }
 
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -130,9 +130,6 @@ Instruction& Instruction::operator=(Instruction&& that) {
   operands_ = std::move(that.operands_);
   dbg_line_insts_ = std::move(that.dbg_line_insts_);
   dbg_scope_ = that.dbg_scope_;
-  for (auto& i : dbg_line_insts_) {
-    i.dbg_scope_ = that.dbg_scope_;
-  }
   return *this;
 }
 
@@ -145,9 +142,6 @@ Instruction* Instruction::Clone(IRContext* c) const {
   clone->operands_ = operands_;
   clone->dbg_line_insts_ = dbg_line_insts_;
   clone->dbg_scope_ = dbg_scope_;
-  for (auto& i : clone->dbg_line_insts_) {
-    i.dbg_scope_ = dbg_scope_;
-  }
   return clone;
 }
 
@@ -767,6 +761,29 @@ bool Instruction::IsOpcodeSafeToDelete() const {
     default:
       return false;
   }
+}
+
+void DebugScope::ToBinary(uint32_t type_id, uint32_t result_id,
+                          uint32_t ext_set,
+                          std::vector<uint32_t>* binary) const {
+  uint32_t num_words = 7;
+  uint32_t dbg_opcode = 23;
+  if (!lexical_scope_) {
+    num_words = 5;
+    dbg_opcode = 24;
+  } else if (!inlined_at_) {
+    num_words = 6;
+  }
+  std::vector<uint32_t> operands = {
+      (num_words << 16) | static_cast<uint16_t>(SpvOpExtInst),
+      type_id,
+      result_id,
+      ext_set,
+      dbg_opcode,
+  };
+  binary->insert(binary->end(), operands.begin(), operands.end());
+  if (lexical_scope_) binary->push_back(lexical_scope_);
+  if (inlined_at_) binary->push_back(inlined_at_);
 }
 
 }  // namespace opt
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -16,6 +16,7 @@
 
 #include <initializer_list>
 
+#include "OpenCLDebugInfo100.h"
 #include "source/disassemble.h"
 #include "source/opt/fold.h"
 #include "source/opt/ir_context.h"
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -39,7 +39,7 @@ Instruction::Instruction(IRContext* c)
       has_type_id_(false),
       has_result_id_(false),
       unique_id_(c->TakeNextUniqueId()),
-      dbg_scope_(0, 0) {}
+      dbg_scope_(kNoDebugScope, kNoInlinedAt) {}
 
 Instruction::Instruction(IRContext* c, SpvOp op)
     : utils::IntrusiveNodeBase<Instruction>(),
@@ -48,7 +48,7 @@ Instruction::Instruction(IRContext* c, SpvOp op)
       has_type_id_(false),
       has_result_id_(false),
       unique_id_(c->TakeNextUniqueId()),
-      dbg_scope_(0, 0) {}
+      dbg_scope_(kNoDebugScope, kNoInlinedAt) {}
 
 Instruction::Instruction(IRContext* c, const spv_parsed_instruction_t& inst,
                          std::vector<Instruction>&& dbg_line)
@@ -58,7 +58,7 @@ Instruction::Instruction(IRContext* c, const spv_parsed_instruction_t& inst,
       has_result_id_(inst.result_id != 0),
       unique_id_(c->TakeNextUniqueId()),
       dbg_line_insts_(std::move(dbg_line)),
-      dbg_scope_(0, 0) {
+      dbg_scope_(kNoDebugScope, kNoInlinedAt) {
   assert((!IsDebugLineInst(opcode_) || dbg_line.empty()) &&
          "Op(No)Line attaching to Op(No)Line found");
   for (uint32_t i = 0; i < inst.num_operands; ++i) {
@@ -96,7 +96,7 @@ Instruction::Instruction(IRContext* c, SpvOp op, uint32_t ty_id,
       has_result_id_(res_id != 0),
       unique_id_(c->TakeNextUniqueId()),
       operands_(),
-      dbg_scope_(0, 0) {
+      dbg_scope_(kNoDebugScope, kNoInlinedAt) {
   if (has_type_id_) {
     operands_.emplace_back(spv_operand_type_t::SPV_OPERAND_TYPE_TYPE_ID,
                            std::initializer_list<uint32_t>{ty_id});
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -38,7 +38,8 @@ Instruction::Instruction(IRContext* c)
       opcode_(SpvOpNop),
       has_type_id_(false),
       has_result_id_(false),
-      unique_id_(c->TakeNextUniqueId()) {}
+      unique_id_(c->TakeNextUniqueId()),
+      dbg_scope_(0, 0) {}
 
 Instruction::Instruction(IRContext* c, SpvOp op)
     : utils::IntrusiveNodeBase<Instruction>(),
@@ -46,7 +47,8 @@ Instruction::Instruction(IRContext* c, SpvOp op)
       opcode_(op),
       has_type_id_(false),
       has_result_id_(false),
-      unique_id_(c->TakeNextUniqueId()) {}
+      unique_id_(c->TakeNextUniqueId()),
+      dbg_scope_(0, 0) {}
 
 Instruction::Instruction(IRContext* c, const spv_parsed_instruction_t& inst,
                          std::vector<Instruction>&& dbg_line)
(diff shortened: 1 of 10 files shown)


source/opt/instruction.h
```diff
@@ -473,6 +495,11 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   // empty.
   std::vector<Instruction> dbg_line_insts_;
 
+  // Debug scope information. If this Instruction is included in a scope,
+  // |dbg_scope_| must have non-zero |lexical_scope|. Otherwise,
+  // |dbg_scope_.lexical_scope| must be 0.
```
s-perron: The comments of the special values for debug scope are best placed on the definition of `DebugScope`.

source/opt/instruction.h
```diff
@@ -91,6 +91,21 @@ inline bool operator!=(const Operand& o1, const Operand& o2) {
   return !(o1 == o2);
 }
 
+// Source level scope information. When a SPIR-V module contains rich
+// debug information written in OpenCL.100.DebugInfo extension, each
+// instruction in a function is wrapped by DebugScope/DebugNoScope,
+// which denotes a lexical scope and/or inlining information for the
+// scope.
+struct DebugScope {
```
s-perron: You should explain why you are not storing the result id.

source/opt/ir_loader.cpp
```diff
@@ -23,23 +23,72 @@
 #include "source/opt/reflect.h"
 #include "source/util/make_unique.h"
 
+static const uint32_t kExtensionInstructionIndex = 4;
```
s-perron: These should try to match the operand names in the spec.  A better name might be `kExtInstSetIndex`. 
 See https://www.khronos.org/registry/spir-v/specs/1.2/SPIRV.html#OpExtInst

source/opt/module.h
```diff
@@ -118,6 +119,16 @@ class Module {
...
   inline void AddFunction(std::unique_ptr<Function> f);
 
+  // Adds a mapping between a result id of local variable OpVariable or
+  // a value of a local variable and a Instruction object for a
+  // DebugDeclare or a DebugValue whose operand is the local variable.
+  inline void AddDebugLocalVariableInfo(uint32_t id,
+                                        std::unique_ptr<Instruction> di);
+
+  // Sets |contains_debug_scope_| as true.
+  inline void SetContainDebugScope();
```
s-perron: Should be `SetContainsDebugScope` and `ContainsDebugScope`.

source/opt/ir_loader.h
```diff
@@ -78,6 +78,12 @@ class IrLoader {
   std::unique_ptr<BasicBlock> block_;
   // Line related debug instructions accumulated thus far.
   std::vector<Instruction> dbg_line_info_;
+
+  // The last DebugScope information. If |last_dbg_scope_.lexical_scope|
+  // is not zero, AddInstruction() lets Instruction::dbg_scope_ be the
+  // same with |last_dbg_scope_|. Otherwise, AddInstruction() lets
+  // Instruction::dbg_scope_ have DebugScope(0, 0) which means DebugNoScope.
```
s-perron: The special values of 0 only need to be documented in the `DebugScope` class.

source/opt/module.cpp
```diff
@@ -137,10 +137,54 @@ void Module::ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const {
...
+        i->ToBinaryWithoutAttachedDebugInsts(binary);
+    };
+    ForEachInst(write_inst, true);
+  } else {
+    DebugScope last_scope(0, 0);
+    auto write_inst = [binary, skip_nop, &last_scope,
+                       this](const Instruction* i) {
+      if (!(skip_nop && i->IsNop())) {
+        const auto& scope = i->GetDebugScope();
+        if (scope != last_scope) {
```
s-perron: Everything in this if statement should be moved to a helper function.

source/opt/module.cpp
```diff
@@ -137,10 +137,54 @@ void Module::ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const {
   binary->push_back(header_.bound);
   binary->push_back(header_.reserved);
 
-  auto write_inst = [binary, skip_nop](const Instruction* i) {
-    if (!(skip_nop && i->IsNop())) i->ToBinaryWithoutAttachedDebugInsts(binary);
-  };
-  ForEachInst(write_inst, true);
+  if (ext_inst_debuginfo_.empty()) {
```
s-perron: For simplicity, I would prefer to just have one version of the lambda that writes an instruction.  We can see it takes up too much time, then we can reconsider.

source/opt/module.cpp
```diff
@@ -137,10 +137,54 @@ void Module::ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const {
...
+  } else {
+    DebugScope last_scope(0, 0);
+    auto write_inst = [binary, skip_nop, &last_scope,
+                       this](const Instruction* i) {
+      if (!(skip_nop && i->IsNop())) {
+        const auto& scope = i->GetDebugScope();
+        if (scope != last_scope) {
+          // Create a new DebugScope instruction
+          std::unique_ptr<Instruction> dscope(
+              ext_inst_debuginfo_.begin()->Clone(context()));
```
s-perron: Why are you cloning?  This just seems to hide what you are doing making the code harder to read.

source/opt/module.cpp
```diff
@@ -137,10 +137,54 @@ void Module::ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const {
...
+          std::vector<Operand> operands(dscope->begin(), dscope->begin() + 4);
+          operands[1] = Operand(
+              SPV_OPERAND_TYPE_RESULT_ID,
+              utils::SmallVector<uint32_t, 2>({context()->TakeNextId()}));
+          if (scope.lexical_scope) {
+            operands[3] = Operand(SPV_OPERAND_TYPE_LITERAL_INTEGER,
+                                  utils::SmallVector<uint32_t, 2>({23}));
+            operands.emplace_back(
+                SPV_OPERAND_TYPE_RESULT_ID,
```
s-perron: I don't think this is correct.  I think this should be SPV_OPERAND_TYPE_ID.  It is not a result in this instruction.  SPV_OPERAND_TYPE_RESULT_ID is only for the result of the current instruction.

source/opt/module.cpp
```diff
@@ -137,10 +137,54 @@ void Module::ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const {
...
+                utils::SmallVector<uint32_t, 2>({scope.lexical_scope}));
+            if (scope.inlined_at) {
+              operands.emplace_back(
+                  SPV_OPERAND_TYPE_RESULT_ID,
```
s-perron: Same here.

source/opt/module.cpp
```diff
@@ -137,10 +137,54 @@ void Module::ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const {
...
+                  utils::SmallVector<uint32_t, 2>({scope.inlined_at}));
```
s-perron: I'm guessing that `{scope.inline_at}` does not work.  It is does please use that instead.

source/opt/module.cpp
```diff
@@ -137,10 +137,54 @@ void Module::ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const {
...
+          dscope->ReplaceOperands(operands);
+          dscope->ToBinaryWithoutAttachedDebugInsts(binary);
+        }
+        last_scope = scope;
+
+        i->ToBinaryWithoutAttachedDebugInsts(binary);
+        auto it = local_var_info_.find(i->result_id());
+        if (it != local_var_info_.end())
+          it->second->ToBinaryWithoutAttachedDebugInsts(binary);
+      }
```
s-perron: Why don't you just output all of the instructions in the local var info?  I don't think that writing the binary should be deleting instructions.  We might want to see the module as it actually is.  We can rely on dead code elimination to remove unused instructions.

source/opt/module.h
```diff
@@ -295,6 +306,13 @@ class Module {
   // If the module ends with Op*Line instruction, they will not be attached to
   // any instruction.  We record them here, so they will not be lost.
   std::vector<Instruction> trailing_dbg_line_info_;
+
+  // This module contains DebugScope or DebugNoScope.
+  bool contains_debug_scope_;
+
+  // A map between a result id of a local variable or a value of a local
+  // variable and its corresponding DebugDeclare or DebugValue.
+  std::unordered_map<uint32_t, std::unique_ptr<Instruction>> local_var_info_;
```
s-perron: Why do you need a map and not just a vector of instructions?

s-perron: You will also need to update `ForEachInst` to traverse these instructions.

jaebaek:
Because of a test failure, I rebased it with the upstream master

source/opt/instruction.h
```diff
@@ -91,6 +91,46 @@ inline bool operator!=(const Operand& o1, const Operand& o2) {
...
 }
 
+// This structure is used to represent a DebugScope instruction from
+// the OpenCL.100.DebugInfo extened instruction set. Since this
+// structure is added to Instruction class as a member variable, we
+// want to reduce the size of structure. Hence, we keep only its Scope
+// and InlinedAt operands but not its result id. Note that losing its
+// result id will result in a different binary even when there is no
+// optimization. We skip the sanity check in Optimizer::Run(..) when
+// the module contains a DebugScope instruction.
```
s-perron: You should explain that the result id is not suppose to be used for anything, which is why we can ignore it.

Do we have a validator check that the result id of a debug scope instruction is not used?

source/opt/instruction.h
```diff
@@ -91,6 +91,46 @@ inline bool operator!=(const Operand& o1, const Operand& o2) {
...
+  void SetInlinedAt(uint32_t at) { inlined_at_ = at; }
+
+  // Pushes the binary segments for this DebugScope instruction into
+  // the back of *|binary|.
+  void ToBinary(uint32_t type_id, uint32_t result_id, uint32_t ext_set,
+                std::vector<uint32_t>* binary) const;
+
+ private:
+  // The result id of the lexical scope in which this debug scope is
+  // contained. The value is |0| if there is no scope.
```
s-perron: Sorry, I should have said this earlier.  I'm seeing too many `0` around the code.  Should be have a named constant for them: `kNoDebugScope` and `kNoInlinedAt`?  It might make the code more readable.

source/opt/ir_loader.cpp
```diff
@@ -23,23 +23,72 @@
...
   }
 
+  // If it is a DebugScope or DebugNoScope of debug extension, we do not
+  // create a new instruction, but simply keep the information in
+  // struct DebugScope.
+  if (opcode == SpvOpExtInst && spvExtInstIsDebugInfo(inst->ext_inst_type)) {
+    const uint32_t ext_inst_index = inst->words[kExtInstSetIndex];
+    if (inst->ext_inst_type == SPV_EXT_INST_TYPE_OPENCL_DEBUGINFO_100) {
+      const OpenCLDebugInfo100Instructions ext_inst_key =
+          OpenCLDebugInfo100Instructions(ext_inst_index);
```
s-perron: Can you simple do: `const OpenCLDebugInfo100Instructions ext_inst_key(ext_inst_index);`?  This happens in two other places as well.

source/opt/ir_loader.cpp
```diff
@@ -23,23 +23,72 @@
...
+        return true;
+      }
+      if (ext_inst_key == OpenCLDebugInfo100DebugNoScope) {
+        last_dbg_scope_ = DebugScope(0, 0);
+        module()->SetContainsDebugScope();
+        return true;
+      }
+    } else {
+      const DebugInfoInstructions ext_inst_key =
+          DebugInfoInstructions(ext_inst_index);
```
s-perron: Same as above.  Can we do `const DebugInfoInstructions ext_inst_key(ext_inst_index);`?

source/opt/ir_loader.cpp
```diff
@@ -131,40 +181,53 @@ bool IrLoader::AddInstruction(const spv_parsed_instruction_t* inst) {
...
+      if (opcode == SpvOpLoopMerge || opcode == SpvOpSelectionMerge)
+        last_dbg_scope_ = DebugScope(0, 0);
+      if (last_dbg_scope_.GetLexicalScope())
+        spv_inst->SetDebugScope(last_dbg_scope_);
+      if (opcode == SpvOpExtInst &&
+          spvExtInstIsDebugInfo(inst->ext_inst_type)) {
+        const uint32_t ext_inst_index = inst->words[kExtInstSetIndex];
+        if (inst->ext_inst_type == SPV_EXT_INST_TYPE_OPENCL_DEBUGINFO_100) {
+          const OpenCLDebugInfo100Instructions ext_inst_key =
+              OpenCLDebugInfo100Instructions(ext_inst_index);
```
s-perron: Same as above.

source/opt/ir_loader.cpp
```diff
@@ -131,40 +181,53 @@ bool IrLoader::AddInstruction(const spv_parsed_instruction_t* inst) {
...
+          switch (ext_inst_key) {
+            case OpenCLDebugInfo100DebugDeclare: {
+              const uint32_t local_var_id = inst->words[kLocalVariableIdIndex];
+              module_->AddDebugLocalVariableInfo(local_var_id,
+                                                 std::move(spv_inst));
+              break;
+            }
+            case OpenCLDebugInfo100DebugValue: {
+              const uint32_t value_id = inst->words[kLocalVariableIdIndex];
+              module_->AddDebugLocalVariableInfo(value_id, std::move(spv_inst));
```
s-perron: For debug value, you have to leave that in the body of the function at the position that it appears.  That is because it says "starting here the value of the variable is obtained this way".  If there are multiple debug values for the same variable, how do you know which one to applies?  

I think that debug declare should be in the function body as well.  We  can discuss this again.  If needed, we can build a map from a variable to it debug declare instruction as an analysis.

source/opt/module.cpp
```diff
@@ -137,10 +137,27 @@ void Module::ToBinary(std::vector<uint32_t>* binary, bool skip_nop) const {
   binary->push_back(header_.bound);
   binary->push_back(header_.reserved);
 
-  auto write_inst = [binary, skip_nop](const Instruction* i) {
-    if (!(skip_nop && i->IsNop())) i->ToBinaryWithoutAttachedDebugInsts(binary);
+  size_t bound_idx = binary->size() - 2;
+  DebugScope last_scope(0, 0);
```
s-perron: Should use the names constants instead of `0, 0`.

source/opt/module.h
s-perron: We need to discuss this more.  The debug declare should be in the body of the function (in the basic block), and the debug value has to be the same.  I'll try  coming up with an example to demonstrate.

source/opt/instruction.cpp
```diff
@@ -735,5 +763,28 @@ bool Instruction::IsOpcodeSafeToDelete() const {
   }
 }
 
+void DebugScope::ToBinary(uint32_t type_id, uint32_t result_id,
+                          uint32_t ext_set,
+                          std::vector<uint32_t>* binary) const {
+  uint32_t num_words = 7;
+  uint32_t dbg_opcode = 23;
```
s-perron: Isn't there a named constant for this?  Where does 23 come from?

source/opt/instruction.cpp
```diff
@@ -735,5 +763,28 @@ bool Instruction::IsOpcodeSafeToDelete() const {
...
+  if (!lexical_scope_) {
```
s-perron: For this and the next check, it would be easier to read of you compare with the named constants.  Also use the accessor functions.

```
if (GetLexicalScope() == kNoDebugScope) {
```

source/opt/instruction.h
```diff
@@ -91,6 +94,44 @@ inline bool operator!=(const Operand& o1, const Operand& o2) {
...
 
+// This structure is used to represent a DebugScope instruction from
+// the OpenCL.100.DebugInfo extened instruction set. Note that we can
+// ignore the result id of DebugScope instruction because it is not
+// used for anything. We do not keep it to reduce the size of
+// structure.
+// TODO: Let validator check that the result id is not used anywhere.
+class DebugScope {
+ public:
+  DebugScope(uint32_t id0, uint32_t id1)
```
s-perron: please give more meaningful names to `id0` and `id1`.

source/opt/module.cpp
jaebaek: done

jaebaek:
Rebased

PR #3190, opened 2020-02-14
Brief guide to writing a spirv-fuzz fuzzer pass

afd:


docs/spirv-fuzz.md
```diff
@@ -0,0 +1,83 @@
+# Guide to writing a spirv-fuzz fuzzer pass
+
+Writing a spirv-fuzz fuzzer pass usually requires two main contributions:
+
+- A *transformation*, capturing a small semantics-preserving change that can be made to a SPIR-V module.  This requires adding a protobuf message representing the transformation, and a corresponding class that implements the `Transformation` interface.
+- A new *fuzzer pass* class, implementing the `FuzzerPass` interface, that knows how to walk a SPIR-V module and apply the new transformation in a randomized fashion.
+
+In some cases, more than one kind of transformation is required for a single fuzzer pass, and in some cases the transformations that a new fuzzer pass requires have already been introduced by existing passes.  But the most common case is to introduce a transformation and fuzzer pass together.
+
```
paulthomson: Note that all fuzzer passes _record_ the transformations as they are applied, resulting in a list of transformation protobuf messages. If we take the original SPIR-V module (without any transformations applied), we can re-apply all of the transformations in the list (in the same order) to get to the same transformed SPIR-V module.

Crucially, we can _replay_ a _subset_ of the transformations. Thus, when replaying an individual transformation, the SPIR-V module might be _different_ from when the transformation was created because earlier transformations were not replayed. As we will see, transformations much be designed with this in mind.

docs/spirv-fuzz.md
```diff
@@ -0,0 +1,83 @@
+# Guide to writing a spirv-fuzz fuzzer pass
+
+Writing a spirv-fuzz fuzzer pass usually requires two main contributions:
+
+- A *transformation*, capturing a small semantics-preserving change that can be made to a SPIR-V module.  This requires adding a protobuf message representing the transformation, and a corresponding class that implements the `Transformation` interface.
+- A new *fuzzer pass* class, implementing the `FuzzerPass` interface, that knows how to walk a SPIR-V module and apply the new transformation in a randomized fashion.
+
+In some cases, more than one kind of transformation is required for a single fuzzer pass, and in some cases the transformations that a new fuzzer pass requires have already been introduced by existing passes.  But the most common case is to introduce a transformation and fuzzer pass together.
```
paulthomson: Remove: "and in some cases the transformations that a new fuzzer pass requires have already been introduced by existing passes."

A fuzzer pass does not really require transformations; it requires stuff in the SPIR-V module. So I think this is just confusing. Perhaps it would be better to replace it with a mini-example.

"In some cases, more than one kind of transformation is required for a single fuzzer pass; e.g. transformation that adds a type declaration, transformation that adds a use of the type declaration."

docs/spirv-fuzz.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Adding a new protobuf message
+
+Take a look at the `Transformation` message in `spvtoolsfuzz.proto`.  This has a `oneof` field that can be any one of the different spirv-fuzz transformations.  Observe that one of the options is `TransformationSetSelectionControl`.  When adding a transformation you first need to add an option for your transformation to the end of the `oneof` declaration.
+
+Now look at the `TransformationSetSelectionControl` message.  If adding your own transformation you need to add a new message for your transformation, and it should be placed alphabetically with respect to other transformations.
+
+The fields of `TransformationSetSelectionControl` provide just enough information to (a) determine whether a given example of this transformation is actually applicable, and (b) apply the transformation in the case that it is applicable.  The details of the transformation message will vary a lot between transformations.  In this case, the message has a `block_id` field, specifying a block that must end with `OpSelectionMerge`, and a `selection_control` field, which is the new value for the selection control mask of the `OpSelectionMerge` instruction.
```
paulthomson: (a) and (b) might just be confusing at this stage because we have not really introduced much about why a transformation would not be applicable. This will also be covered when discussing the transformation class anyway. Could say: 

The fields of `TransformationSetSelectionControl` provide the information needed to find the `OpSelectionMerge` instruction and to change it. As explained earlier, the SPIR-V module may have changed, which is why we say that we must _find_ the instruction. The details of the transformation message will vary...

docs/spirv-fuzz.md
```diff
@@ -0,0 +1,83 @@
...
+
+For example, the tests for `TransformationSetSelectionControl` check that a transformation of this kind is inapplicable if the `block_id` field of the transformation is not a block, or does not end in `OpSelectionMerge`, or if the `selection_control` mask has an illegal value.  It also checks that applying a sequence of valid transformations to a SPIR-V shader leads to a shader with appropriately modified selection controls.
+
+## Adding a new fuzzer pass class
+
+A *fuzzer pass* traverses a SPIR-V module looking for places to apply a certain kind of transformation, and randomly decides at which of these points to actually apply the transformation.  It might be necessary to apply other transformations in order to apply a given transformation (for example, if a transformation requires a certain type to be present in the module, said type can be added if not already present via another transformation).
```
paulthomson: A *fuzzer pass* traverses a SPIR-V module looking for locations where a certain kind of transformation could be applied, and randomly decides at which of these points to actually apply the transformation.

PR #3198, opened 2020-02-21
Avoid ignored const in static_cast

dneto0:
Fixes a warning (and hence error) on GCC 9.2


zoddicus:
https://github.com/KhronosGroup/SPIRV-Tools/pull/3197 already fixes this

dneto0:
I merged #3198. So I'm abandoning this one.

PR #3200, opened 2020-02-21
Combine extinst-name and extinst-output-base into one arg.

vonture:
Some build systems such as Android blueprints require that the inputs
and outputs of generator scripts are all provided as arguments.  These
two arguments to generate_language_headers.py are combined to form the
output path in the script.  This change simply lets the user provide the
whole output path as an argument.

I wasn't able to test to bazel build files locally.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -235,9 +235,8 @@ $(1)/$(2).h : \
         $(LOCAL_PATH)/utils/generate_language_headers.py \
         $(3)
 		@$(HOST_PYTHON) $(LOCAL_PATH)/utils/generate_language_headers.py \
-		    --extinst-name=$(2) \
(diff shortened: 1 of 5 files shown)


vonture:
David, PTAL.

dneto0:
I had ambitions for this script to generate multiple language headers for each grammar file. But in the end I only made it for a single output C/C++.

This is a sensible change and I can definitely see how it makes it easier to blend into various build systems.  If I ever get to the other language outputs, I can always add new options for each new output.

vonture:
Hey David, I think I fixed the issues with ndk and bazel builds, can you re-run kokoro?

vonture:
Awesome, thanks.  Is this good to land?

PR #3201, opened 2020-02-23
Fix Wrange-loop-analysis warnings in SPIRV-Tools.

nico:
Part of https://bugs.chromium.org/p/chromium/issues/detail?id=1039697


nico:
Thanks! Can you hit "merge" as well?

PR #3203, opened 2020-02-28
Avoid use of Python distutils.dir_util

dneto0:
Required for compatibility with (some?) Python3 installations.


dneto0:
This copies the solution from update_build_version.py

PR #3205, opened 2020-03-03
spirv-fuzz: Add swap commutable operands transformation

andreperezmaselco:
In this PR, the classes that represent the swap commutable operands transformation and the fuzzer pass were implemented.

### DIFF
--- a/source/fuzz/fuzzer_pass_swap_commutable_operands.cpp
+++ b/source/fuzz/fuzzer_pass_swap_commutable_operands.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 André Perez Maselco
+// Copyright (c) 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/source/fuzz/fuzzer_pass_swap_commutable_operands.cpp
+++ b/source/fuzz/fuzzer_pass_swap_commutable_operands.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Google LLC
+// Copyright (c) 2020 André Perez Maselco
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/source/fuzz/fuzzer_pass_swap_commutable_operands.cpp
+++ b/source/fuzz/fuzzer_pass_swap_commutable_operands.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Google LLC
+// Copyright (c) 2020 André Perez Maselco
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -22,26 +22,28 @@ namespace spvtools {
 namespace fuzz {
 
 FuzzerPassSwapCommutableOperands::FuzzerPassSwapCommutableOperands(
-  opt::IRContext* ir_context,
-  FactManager* fact_manager,
-  FuzzerContext* fuzzer_context,
-  protobufs::TransformationSequence* transformations
-) : FuzzerPass(ir_context, fact_manager, fuzzer_context, transformations) {}
+    opt::IRContext* ir_context, FactManager* fact_manager,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, fact_manager, fuzzer_context, transformations) {}
 
 FuzzerPassSwapCommutableOperands::~FuzzerPassSwapCommutableOperands() = default;
 
 void FuzzerPassSwapCommutableOperands::Apply() {
   auto context = GetIRContext();
-  // Iterates over the module's instructions and checks whether it is commutative.
-  // In this case, the transformation is probabilistically applied.
-  context->module()->ForEachInst([this, context](opt::Instruction* instruction) {
-    if (spvOpcodeIsCommutative(instruction->opcode()) &&
-        GetFuzzerContext()->ChooseEven()) {
-      auto instructionDescriptor = MakeInstructionDescriptor(context, instruction);
-      auto transformation = TransformationSwapCommutableOperands(instructionDescriptor);
-      ApplyTransformation(transformation);
-    }
-  });
+  // Iterates over the module's instructions and checks whether it is
+  // commutative. In this case, the transformation is probabilistically applied.
+  context->module()->ForEachInst(
+      [this, context](opt::Instruction* instruction) {
+        if (spvOpcodeIsCommutativeBinaryOperator(instruction->opcode()) &&
+            GetFuzzerContext()->ChooseEven()) {
+          auto instructionDescriptor =
+              MakeInstructionDescriptor(context, instruction);
+          auto transformation =
+              TransformationSwapCommutableOperands(instructionDescriptor);
+          ApplyTransformation(transformation);
+        }
+      });
 }
 
 }  // namespace fuzz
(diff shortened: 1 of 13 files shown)


source/fuzz/fuzzer_pass_swap_commutable_operands.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2020 Google LLC
```
afd: For any files that you contribute from scratch, please use:

// Copyright (c) 2020 [your name]

source/fuzz/fuzzer_pass_swap_commutable_operands.h
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2020 Google LLC
```
afd: Ditto.

source/fuzz/transformation_swap_commutable_operands.cpp
```diff
@@ -0,0 +1,55 @@
...
+) {
+  *message_.mutable_instruction_descriptor() = instruction_descriptor;
+}
+
+bool TransformationSwapCommutableOperands::IsApplicable(
+  opt::IRContext* /*unused*/,
+  const spvtools::fuzz::FactManager& /*unused*/
+) const {
+  SpvOp opcode = static_cast<SpvOp>(message_.instruction_descriptor().target_instruction_opcode());
+  return spvOpcodeIsCommutative(opcode);
```
afd: This needs a bit more work: you should check that the instruction descriptor refers to an instruction that actually exists (find the instruction and check that the pointer you get is not null, returning false if so).

As a sanity check you could also assert that the located instruction has the same opcode as in the descriptor.

source/fuzz/transformation_swap_commutable_operands.cpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+void TransformationSwapCommutableOperands::Apply(
+  opt::IRContext* context, spvtools::fuzz::FactManager* /*unused*/
+) const {
+  auto instruction = FindInstruction(message_.instruction_descriptor(), context);
+  std::swap(instruction->GetOperand(2), instruction->GetOperand(3));
```
afd: Would it work to use GetInOperand(0) and GetInOperand(1) here?  I prefer to work with the "In" versions of operand getter methods when manipulating the input operands of an instruction.

Either way, please add a comment explaining that, by design, the instructions defined to be commutative have exactly two input parameters.

source/fuzz/fuzzer_pass_swap_commutable_operands.h
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef SOURCE_FUZZ_FUZZER_PASS_SWAP_COMMUTABLE_OPERANDS_H_
+#define SOURCE_FUZZ_FUZZER_PASS_SWAP_COMMUTABLE_OPERANDS_H_
+
+#include "source/fuzz/fuzzer_pass.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// Fuzzer pass that swaps the operands of a commutative instruction, at random.
```
afd: Rephrase to make it clear that the instructions are chosen at random (the swapping itself is, of course, deterministic).

source/fuzz/transformation_swap_commutable_operands.h
```diff
@@ -0,0 +1,53 @@
...
+  explicit TransformationSwapCommutableOperands(
+    const protobufs::TransformationSwapCommutableOperands& message
+  );
+
+  TransformationSwapCommutableOperands(
+    const protobufs::InstructionDescriptor& instruction_descriptor
+  );
+
+  // The transformation is applicable if the opcode represents
+  // a commutative instruction.
```
afd: Please look at the comments associated with other transformation classes and try to follow that style, in particular adding a description of the precondition I mention above about the described instruction having to exist.

source/opcode.cpp
```diff
@@ -612,6 +612,39 @@ bool spvOpcodeIsDebug(SpvOp opcode) {
   }
 }
 
+bool spvOpcodeIsCommutative(SpvOp opcode) {
```
afd: How about `spvOpcodeIsCommutativeBinaryOperator` as a name?

source/opcode.h
```diff
@@ -130,6 +130,9 @@ bool spvOpcodeIsScalarizable(SpvOp opcode);
 // Returns true if the given opcode is a debug instruction.
 bool spvOpcodeIsDebug(SpvOp opcode);
 
+// Returns true if the given opcode is a commutative instruction.
```
afd: I would expand the comment a bit to say that it returns true for exactly those opcodes that are binary operators where the order of the operands is irrelevant.

test/fuzz/transformation_swap_commutable_operands_test.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+#include "source/fuzz/transformation_swap_commutable_operands.h"
+#include "source/fuzz/instruction_descriptor.h"
+#include "test/fuzz/fuzz_test_util.h"
+
+namespace spvtools {
+namespace fuzz {
+namespace {
+
+TEST(TransformationSwapCommutableOperandsTest, IsApplicableTest) {
```
afd: Please update this test with some cases that fail due to the described instruction not existing, including:

- the base instruction id not existing
- there being no instruction with the desired opcode after the base instruction id
- there being an instruction with the desired opcode after the base instruction id, but the skip count associated with the instruction descriptor being so high.

source/opcode.cpp
andreperezmaselco: This name `spvOpcodeIsCommutativeBinaryOperator` is better, since the operator is binary.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1019,6 +1020,18 @@ message TransformationStore {
...
 
+message TransformationSwapCommutableOperands {
+
+  // A transformation that swaps the operands of a commutative instruction.
+
+  // A descriptor for a commutative instruction
+  InstructionDescriptor instruction_descriptor = 1;
+
+  // The commutable operands
+  repeated uint32 operands = 2;
```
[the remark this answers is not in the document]

source/fuzz/fuzzer_pass_swap_commutable_operands.cpp
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_swap_commutable_operands.h
andreperezmaselco: Done.

source/fuzz/transformation_swap_commutable_operands.cpp
andreperezmaselco: Done.

source/fuzz/transformation_swap_commutable_operands.cpp
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_swap_commutable_operands.h
andreperezmaselco: Done.

source/fuzz/transformation_swap_commutable_operands.h
andreperezmaselco: Done.

source/opcode.h
andreperezmaselco: Done.

test/fuzz/transformation_swap_commutable_operands_test.cpp
andreperezmaselco: Done.

PR #3233, opened 2020-03-17
Update validator for SPV_KHR_ray_tracing.

dgkoch:
Minimal support only at this point. 

CC @s-perron @johnkslang


s-perron:
This PR will not work without changes from #3234.  We need to either merge those two PR (ie close this one and just use #3234), or reorder them so we can merge #3234 first.

dgkoch:
ok - we should just close this one and use the other #3234 then.

PR #3235, opened 2020-03-17
Add support for KHR_ray_{query,tracing} extensions

dgkoch:

PR #3236, opened 2020-03-17
Whitelist SPV_EXT_demote_to_helper_invocation for opt passes

ehsannas:


s-perron:
We'll need to rebase on top of #3235 once it has been merged.  That way we can run the tests.

ehsannas:
Rebased

PR #3246, opened 2020-03-23
Disallow phis of images, samplers and sampled images

alan-baker:
* Disallow phis of sampled images always
* Disallow phis of samplers and images for shaders
* Add tests

### DIFF
--- a/source/val/validate_cfg.cpp
+++ b/source/val/validate_cfg.cpp
@@ -62,6 +62,13 @@ spv_result_t ValidatePhi(ValidationState_t& _, const Instruction* inst) {
     }
   }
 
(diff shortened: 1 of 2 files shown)


s-perron:
I'm not sure how we should handle this.  It won't affect DXC directly.  However, when legalizing the code from DXC, we will probably create OpPhi instructions on images and samplers as an intermediate step.

I'm not asking for any changes, because I think it will still works.  However, it is something to be aware of.

alan-baker:
I'm happy to gate this with the legalization option if you prefer.

s-perron:
> I'm happy to gate this with the legalization option if you prefer.

I think that would be a good idea.  Thanks.

PR #3251, opened 2020-03-25
Preserve debug info in eliminate-dead-functions

jaebaek:
The elimination of dead functions makes OpFunction operand of
DebugFunction invalid. This commit replaces the operand with
DebugInfoNone.

### DIFF
--- a/source/opt/eliminate_dead_functions_util.cpp
+++ b/source/opt/eliminate_dead_functions_util.cpp
@@ -24,6 +24,7 @@ Module::iterator EliminateFunction(IRContext* context,
   (*func_iter)
       ->ForEachInst([context](Instruction* inst) { context->KillInst(inst); },
                     true);
+  // TODO: Kill DebugInlinedAt used only by DebugScope in a killed function.
   return func_iter->Erase();
 }
 
--- a/source/opt/eliminate_dead_functions_util.cpp
+++ b/source/opt/eliminate_dead_functions_util.cpp
@@ -24,7 +24,6 @@ Module::iterator EliminateFunction(IRContext* context,
   (*func_iter)
       ->ForEachInst([context](Instruction* inst) { context->KillInst(inst); },
                     true);
-  // TODO: Kill DebugInlinedAt used only by DebugScope in a killed function.
   return func_iter->Erase();
 }
 
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -22,14 +22,14 @@
 #include <utility>
 #include <vector>
 
-#include "source/opcode.h"
-#include "source/operand.h"
-#include "source/util/ilist_node.h"
-#include "source/util/small_vector.h"
-
+#include "OpenCLDebugInfo100.h"
 #include "source/latest_version_glsl_std_450_header.h"
 #include "source/latest_version_spirv_header.h"
+#include "source/opcode.h"
+#include "source/operand.h"
 #include "source/opt/reflect.h"
+#include "source/util/ilist_node.h"
+#include "source/util/small_vector.h"
 #include "spirv-tools/libspirv.h"
 
 const uint32_t kNoDebugScope = 0;
@@ -162,7 +162,8 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
         has_type_id_(false),
         has_result_id_(false),
         unique_id_(0),
-        dbg_scope_(kNoDebugScope, kNoInlinedAt) {}
+        dbg_scope_(kNoDebugScope, kNoInlinedAt),
+        is_cl100_dbg_inst_(false) {}
 
   // Creates a default OpNop instruction.
   Instruction(IRContext*);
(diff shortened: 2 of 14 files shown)


source/opt/eliminate_dead_functions_util.cpp
```diff
@@ -24,6 +24,7 @@ Module::iterator EliminateFunction(IRContext* context,
   (*func_iter)
       ->ForEachInst([context](Instruction* inst) { context->KillInst(inst); },
                     true);
+  // TODO: Kill DebugInlinedAt used only by DebugScope in a killed function.
```
s-perron, line 27: This can be taken care of by ADCE.  Nothing needs to be done in this pass for this.

test/opt/eliminate_dead_functions_test.cpp
```diff
@@ -204,6 +204,132 @@ OpFunctionEnd
...
+%33 = OpLoad %v4float %3
+%34 = OpFAdd %v4float %32 %33
+OpStore %4 %34
+%47 = OpExtInst %void %1 DebugNoScope
+OpReturn
+OpFunctionEnd
+)";
+
+  SetAssembleOptions(SPV_TEXT_TO_BINARY_OPTION_PRESERVE_NUMERIC_IDS);
+  SinglePassRunAndCheck<EliminateDeadFunctionsPass>(text, expected_output,
```
s-perron, line 329: Do a match test.  The make it easier to see what the test is actually trying to test.  See https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/opt/strength_reduction_test.cpp#L70

source/opt/instruction.h
```diff
@@ -162,7 +162,8 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
         has_type_id_(false),
         has_result_id_(false),
         unique_id_(0),
-        dbg_scope_(kNoDebugScope, kNoInlinedAt) {}
+        dbg_scope_(kNoDebugScope, kNoInlinedAt),
+        is_cl100_dbg_inst_(false) {}
```
s-perron, line 166: I don't like having this.  We are burning 32-bit for every instruction, and all that we gain is to avoid checking that this is an instruction in the appropriate extended instruction set.  We can cache the id of the cl100_dbg extended instruction set like we do for the glslstd450 (https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/feature_manager.h#L86).

source/opt/ir_context.cpp
```diff
@@ -365,6 +373,49 @@ void IRContext::KillNamesAndDecorates(Instruction* inst) {
   KillNamesAndDecorates(rId);
 }
 
+void IRContext::KillOperandFromDebugInstructions(Instruction* inst) {
```
s-perron, line 376: You should have unit tests for this function in https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/opt/ir_context_test.cpp

source/opt/module.h
```diff
@@ -119,6 +120,10 @@ class Module {
   // Appends a function to this module.
   inline void AddFunction(std::unique_ptr<Function> f);
 
+  // Get the existing DebugInfoNone. If it is null, create one and keep it.
+  Instruction* GetDebugInfoNone(uint32_t type_id, uint32_t result_id,
+                                uint32_t ext_set);
+
```
s-perron, line 126: This does not belong on the module, move it to the IRContext.  Also it should not take a result id since it may not create a new instruction.  It should take the next id if it needs to create a new instruction.

jaebaek: done

test/opt/eliminate_dead_functions_test.cpp
jaebaek: done

source/opt/ir_context.h
```diff
@@ -798,6 +806,10 @@ class IRContext {
   // Whether all specialization constants within |module_|
   // should be preserved.
   bool preserve_spec_constants_;
+
+  // DebugInfoNone instruction. We need only a single DebugInfoNone.
+  // To reuse the existing one, we keep it using this member variable.
+  Instruction* debug_info_none_;
```
s-perron, line 812: The name of this variable sounds like it should be a boolean that says if there is debug info or not.  Can you rename this and the related getter to `debug_info_none_inst_`?

source/opt/ir_context.cpp
```diff
@@ -365,6 +372,62 @@ void IRContext::KillNamesAndDecorates(Instruction* inst) {
...
+        continue;
+      auto& operand = it->GetOperand(kDebugFunctionOperandFunctionIndex);
+      if (operand.words[0] == id) {
+        operand.words[0] = GetOpenCL100DebugInfoNone()->result_id();
+      }
+    }
+  }
+  // Kill id of OpVariable for global variable from DebugGlobalVariable.
+  if (opcode == SpvOpVariable ||
+      (SpvOpConstantTrue <= opcode && opcode <= SpvOpConstantNull)) {
```
s-perron, line 413: I think you should be able to use the "IsConstant" function in the instruction class instead of this check.  It will read better.

jaebaek: done

source/opt/ir_context.h
jaebaek: done

PR #3256, opened 2020-03-27
Prevent Effcee install his things when build SIRV-Tools with testing enabled

sl1pkn07:
When build SPIRV-Tools with testing enabled and install it, also install Effcee libs and headers

This PR adds `EXCLUDE_FROM_ALL` in the external/CMakeLists.txt in `add_subdirectory(effcee)`

also prevent build his test and examples


external/CMakeLists.txt
```diff
@@ -88,13 +88,15 @@ if (NOT ${SPIRV_SKIP_TESTS})
...
+        set(RE2_BUILD_TESTING OFF)
       endif()
       if (MSVC)
-	# SPIRV-Tools uses the shared CRT with MSVC.  Tell Effcee to do the same.
-	set(EFFCEE_ENABLE_SHARED_CRT ON)
+        # SPIRV-Tools uses the shared CRT with MSVC.  Tell Effcee to do the same.
+        set(EFFCEE_ENABLE_SHARED_CRT ON)
       endif()
-      add_subdirectory(effcee)
+      set(EFFCEE_BUILD_SAMPLES OFF CACHE BOOL "Do not build Effcee examples")
```
dneto0, line 97: I suggest making this only take effect if the effcee target is not already configured.
Doing so will avoid breaking a build configuration that had effcee already configured at an outer layer.

I did a similar thing at line 90 by checking for the re2 target first.

sl1pkn07: like this snippet you mean?

https://github.com/KhronosGroup/SPIRV-Tools/blob/538512e8e8980cc01ab6501cd1cbd402d54f8491/external/CMakeLists.txt#L90-L92

PR #3259, opened 2020-03-30
Validate Buffer and BufferBlock apply only to struct types

cmarcelo:


source/val/validate_decorations.cpp
```diff
@@ -1570,6 +1586,9 @@ spv_result_t CheckDecorationsFromDecoration(ValidationState_t& vstate) {
         case SpvDecorationNoUnsignedWrap:
           PASS_OR_BAIL(CheckIntegerWrapDecoration(vstate, *inst, decoration));
           break;
+        case SpvDecorationBlock:
```
alan-baker: Based on the error message, this should also include SpvDecorationBufferBlock it seems.

cmarcelo: You are correct. Fixed.

PR #3262, opened 2020-03-31
spirv-fuzz: Handle isomorphic types property in composite construction

afd:
The fuzzer pass that constructs composites had an issue where it would
regard isomorphic but distinct structs (similarly arrays) as being
interchangeable when constructing composites.  This change fixes the
problem by relying less on the type manager.

### DIFF
--- a/source/fuzz/fuzzer_pass_construct_composites.cpp
+++ b/source/fuzz/fuzzer_pass_construct_composites.cpp
@@ -88,7 +88,7 @@ void FuzzerPassConstructComposites::Apply() {
         // for constructing a composite of that type.  Otherwise these variables
         // will remain 0 and null respectively.
         uint32_t chosen_composite_type = 0;
-        std::unique_ptr<std::vector<uint32_t>> constructor_arguments = nullptr;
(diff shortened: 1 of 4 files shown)


source/fuzz/fuzzer_pass_construct_composites.cpp
```diff
@@ -164,18 +161,15 @@ void FuzzerPassConstructComposites::RecordAvailableInstruction(
 
 std::unique_ptr<std::vector<uint32_t>>
 FuzzerPassConstructComposites::TryConstructingArrayComposite(
-    const opt::analysis::Array& array_type,
+    const opt::Instruction& array_type_instruction,
```
paulthomson, line 164: This is not really part of this PR but...I find the function name confusing. Shouldn't it be more like `FindSuitableArrayComponents` or something similar? And does it need to return a unique_ptr?

afd: I addressed the name issue and made this (and the other similar functions) return a vector, not a unique_ptr, with an empty vector (rather than nullptr) representing failure.

PR #3271, opened 2020-04-01
Update dominates to check for null nodes

### LINKED ISSUE #3270: Loop descriptors segfault if the loop has an unreachable merge
```
               OpCapability Shader
               OpMemoryModel Logical GLSL450
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
          %1 = OpFunction %void None %3
          %4 = OpLabel
               OpBranch %5
          %5 = OpLabel
               OpLoopMerge %6 %7 None
               OpBranch %8
          %8 = OpLabel
               OpBranch %9
          %9 = OpLabel
               OpBranch %7
          %7 = OpLabel
               OpBranch %5
          %6 = OpLabel
               OpUnreachable
               OpFunctionEnd
```

Running `--loop-unroll` will cause the optimizer to segfault. The construction of loop descriptors attempts to dereference a null dominator tree node.

alan-baker:
Fixes #3270

* Dominates using tree nodes now checks for null nodes
* Added a test

### DIFF
--- a/source/opt/dominator_tree.cpp
+++ b/source/opt/dominator_tree.cpp
@@ -242,6 +242,7 @@ bool DominatorTree::Dominates(uint32_t a, uint32_t b) const {
 bool DominatorTree::Dominates(const DominatorTreeNode* a,
                               const DominatorTreeNode* b) const {
   // Node A dominates node B if they are the same.
+  if (!a || !b) return false;
   if (a == b) return true;
 
   return a->dfs_num_pre_ < b->dfs_num_pre_ &&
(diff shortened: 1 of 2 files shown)


source/opt/dominator_tree.cpp
```diff
@@ -242,6 +242,7 @@ bool DominatorTree::Dominates(uint32_t a, uint32_t b) const {
 bool DominatorTree::Dominates(const DominatorTreeNode* a,
                               const DominatorTreeNode* b) const {
   // Node A dominates node B if they are the same.
+  if (!a || !b) return false;
```
s-perron, line 245: Can you move the comment at line 244 just before the condition that it is describing?

PR #3277, opened 2020-04-06
spirv-fuzz: Improve support for compute shaders in donation

afd:
(1) Runtime arrays are turned into fixed-size arrays, by turning
    OpTypeRuntimeArray into OpTypeArray and uses of OpArrayLength into
    uses of the constant used for the length of the fixed-size array.

(2) Atomic instructions are not donated, and uses of their results are
    replaced with uses of constants of the result type.

### DIFF
--- a/source/fuzz/fuzzer_pass_donate_modules.cpp
+++ b/source/fuzz/fuzzer_pass_donate_modules.cpp
@@ -116,6 +116,7 @@ SpvStorageClass FuzzerPassDonateModules::AdaptStorageClass(
   switch (donor_storage_class) {
     case SpvStorageClassFunction:
     case SpvStorageClassPrivate:
+    case SpvStorageClassWorkgroup:
       // We leave these alone
       return donor_storage_class;
     case SpvStorageClassInput:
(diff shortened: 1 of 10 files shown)


source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -280,36 +281,71 @@ void FuzzerPassDonateModules::HandleTypesAndValues(
...
-              member_type_ids.push_back(
-                  original_id_to_donated_id->at(*component_type_id));
-            });
+        for (uint32_t i = 0; i < type_or_value.NumInOperands(); i++) {
+          auto component_type_id = type_or_value.GetSingleWordInOperand(i);
+          if (!original_id_to_donated_id->count(component_type_id)) {
+            // We did not donate every member type for this struct type, so we
+            // cannot donate the struct type.
+            continue;
+          }
```
paulthomson, line 327: Are these changes missing from the description of the PR?

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -426,21 +486,43 @@ void FuzzerPassDonateModules::HandleTypesAndValues(
...
         }
         ApplyTransformation(TransformationAddGlobalVariable(
-            new_result_id, remapped_pointer_type, initializer_id, true));
+            new_result_id, remapped_pointer_type, storage_class, initializer_id,
+            true));
       } break;
       case SpvOpUndef: {
+        if (!original_id_to_donated_id->count(type_or_value.type_id())) {
+          // We did not denote the type associated wit this undef, so we cannot
+          // donate the undef.
```
paulthomson, line 522: denote->donate wit->with

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -493,9 +575,79 @@ void FuzzerPassDonateModules::HandleFunctions(
...
+                  donated_struct_type_instruction->NumInOperands() - 1);
+          auto fixed_size_array_type_instruction =
+              GetIRContext()->get_def_use_mgr()->GetDef(
+                  fixed_size_array_type_id);
+          assert(fixed_size_array_type_instruction->opcode() ==
+                     SpvOpTypeArray &&
+                 "The donated array type must be fixed-size.");
+          original_id_to_donated_id->insert(
+              {instruction->result_id(),
+               fixed_size_array_type_instruction->GetSingleWordInOperand(1)});
```
paulthomson, line 623: Could you may add yet another local variable here to clarify that this is the id corresponding to the array size?

source/fuzz/transformation_add_global_variable.cpp
```diff
@@ -38,6 +39,16 @@ bool TransformationAddGlobalVariable::IsApplicable(
...
+
+  // The storage class must be Private or Workgroup.
+  auto storage_class = static_cast<SpvStorageClass>(message_.storage_class());
+  switch (storage_class) {
+    case SpvStorageClassPrivate:
+    case SpvStorageClassWorkgroup:
+      break;
+    default:
+      return false;
+  }
```
paulthomson, line 51: Is this a case where we should assert false?

source/fuzz/transformation_add_global_variable.cpp
```diff
@@ -48,23 +59,29 @@ bool TransformationAddGlobalVariable::IsApplicable(
...
-         "We asked for the constant associated with a single id; we should "
-         "get a single constant.");
-  // The type of the constant must match the pointee type of the pointer.
-  if (pointer_type->pointee_type() != constant_id[0]->type()) {
-    return false;
+  if (message_.initializer_id()) {
+    // An initializer is not allowed if the storage class is Workgroup.
+    if (storage_class == SpvStorageClassWorkgroup) {
+      return false;
+    }
```
paulthomson, line 70: Maybe assert false

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Thanks!

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Done (though I hate local variables :-))

source/fuzz/transformation_add_global_variable.cpp
afd: Yes, done.

source/fuzz/transformation_add_global_variable.cpp
afd: Yes, done.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: I had thought these changes (or at least some of them) were necessary due to the way OpRuntimeArrays are handled, but actually they are not, so I have removed them from the PR.

PR #3280, opened 2020-04-07
spirv-fuzz: Handle more general SPIR-V in donation

afd:
This change increases the extent to which arbitrary SPIR-V can be used
by the fuzzer pass that donates modules.  It handles the case where
various ingredients (such as types, variables and particular
instructions) cannot be donated by omitting them, and then either
omitting their dependencies or replacing their dependencies with
alternative instructions.

The change pays particular attention to allowing code that manipulates
image types to be handled (by skipping anything image-specific).

### DIFF
--- a/source/fuzz/fuzzer_pass_donate_modules.cpp
+++ b/source/fuzz/fuzzer_pass_donate_modules.cpp
@@ -585,7 +585,8 @@ void FuzzerPassDonateModules::HandleFunctions(
     }
     assert(function_to_donate && "Function to be donated was not found.");
 
-    if (!original_id_to_donated_id->count(function_to_donate->DefInst().GetSingleWordInOperand(1))) {
+    if (!original_id_to_donated_id->count(
+            function_to_donate->DefInst().GetSingleWordInOperand(1))) {
       // We were not able to donate this function's type, so we cannot donate
       // the function.
       continue;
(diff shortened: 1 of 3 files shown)


source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -308,18 +325,33 @@ void FuzzerPassDonateModules::HandleTypesAndValues(
...
         // Similar to SpvOpTypeArray.
         std::vector<uint32_t> member_type_ids;
+        bool can_donate = true;
         for (uint32_t i = 0; i < type_or_value.NumInOperands(); i++) {
           auto component_type_id = type_or_value.GetSingleWordInOperand(i);
+          if (!original_id_to_donated_id->count(component_type_id)) {
+            // We did not donate every member type for this struct type, so we
+            // cannot donate the struct type.
+            can_donate = false;
+            break;
```
paulthomson, line 335: I wonder if it would be better to add another function and change "continue"->"return" (elsewhere). Then this could also be a "return".

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -865,6 +668,86 @@ bool FuzzerPassDonateModules::IgnoreInstruction(
...
+        if (!original_id_to_donated_id.count(*in_id)) {
+          if (skipped_instructions.count(*in_id) ||
+              donor_ir_context->get_def_use_mgr()->GetDef(*in_id)->opcode() ==
+                  SpvOpFunction) {
+            result = false;
+            return false;
+          }
+        }
+        return true;
+      });
```
paulthomson, line 738: Could you add a comment for this loop?

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -917,5 +800,331 @@ FuzzerPassDonateModules::GetFunctionsInCallGraphTopologicalOrder(
...
+    // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3279): We could
+    //  instead look for an available id of the right type and generate an
+    //  OpCopyObject of that id.
+    skipped_instructions->insert(instruction.result_id());
+    return;
+  }
+
+  // We are going to donate an OpCopyObject instruction.  Create a result id for
+  // the donated instruction if it does not already exist (it may exist in the
+  // case that it has been forward-referenced).
```
paulthomson, line 886: Bit of a nit: could put "donate" in quotation marks, as we are not exactly donating it. Could instead say "going to add".

Also: "Create a result id..." -> "Add a mapping for the result id of the original instruction if it..." ?

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -917,5 +800,331 @@ FuzzerPassDonateModules::GetFunctionsInCallGraphTopologicalOrder(
...
+  if (!original_id_to_donated_id->count(instruction.result_id())) {
+    original_id_to_donated_id->insert(
+        {instruction.result_id(), GetFuzzerContext()->GetFreshId()});
+  }
+
+  // We donate a copy of the zero constant for the type in question.
```
paulthomson, line 892: "We find or add a zero constant to the receiving module for the type in question." (<- or whatever a good phrase is for the "receiving module"). I.e. it is not added to `donated_instructions`.

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -917,5 +800,331 @@ FuzzerPassDonateModules::GetFunctionsInCallGraphTopologicalOrder(
...
+  // Get the instruction's input operands into donation-ready form,
+  // remapping any id uses in the process.
+  opt::Instruction::OperandList input_operands;
+
+  // Consider each input operand in turn.
+  for (uint32_t in_operand_index = 0;
+       in_operand_index < instruction.NumInOperands(); in_operand_index++) {
+    std::vector<uint32_t> operand_data;
+    const opt::Operand& in_operand = instruction.GetInOperand(in_operand_index);
+    if (spvIsIdType(in_operand.type)) {
```
paulthomson, line 917: The name of this function is _so_ confusing that maybe it is worth clarifying every time we call it that it means "Is this operand an id?" (as opposed to any other _type_ of operand!)

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Good idea.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Yes, will do.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Done.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Done.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Haha, done.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Done.

source/fuzz/fuzzer_pass_donate_modules.cpp
afd: Done (and it is now much simpler!)

PR #3281, opened 2020-04-07
spirv-fuzz: Improve the handling of equation facts

afd:
The management of equation facts suffered from two problems:

(1) The processing of an equation fact required the data descriptors
    used in the equation to be in canonical form.  However, during
    fact processing it can be deduced that certain data descriptors
    are equivalent, causing their equivalence classes to be merged,
    and that could cause previously canonical data descriptors to no
    longer be canonical.

(2) Related to this, if id equations were known about a canonical data
    descriptor dd1, and other id equations known about a different
    canonical data descriptor dd2, the equation facts about these data
    descriptors were not being merged in the event that dd1 and dd2
    were deduced to be equivalent.

This changes solves (1) by not requiring equation facts to be in
canonical form while processing them, but instead always checking
whether (not necessary canonical) data descriptors are equivalent when
looking for corollaries of equation facts, rather than comparing them
using ==.

Problem (2) is solved by adding logic to merge sets of equations when
data descriptors are made equivalent.

In addition, the change also requires elements to be registered in an
equivalence relation before they can be made equivalent, rather than
being added (if not already present) at the point of being made
equivalent.

### DIFF
--- a/source/fuzz/fact_manager.cpp
+++ b/source/fuzz/fact_manager.cpp
@@ -460,16 +460,15 @@ class FactManager::DataSynonymAndIdEquationFacts {
       opt::IRContext* context, const protobufs::DataDescriptor& dd1,
       const protobufs::DataDescriptor& dd2) const;
 
-  // Requires that |lhs_dd_non_canonical| and every element of
-  // |rhs_dds_non_canonical| is present in the |synonymous_| equivalence,
-  // but is not necessarily its own representative.  Records the fact that the
-  // equation "|lhs_dd_non_canonical| |opcode| |rhs_dds_non_canonical|" holds,
-  // and adds any corollaries, in the form of data synonym or equation facts,
-  // that follow from this and other known facts.
+  // Requires that |lhs_dd| and every element of |rhs_dds| is present in the
+  // |synonymous_| equivalence relation, but is not necessarily its own
+  // representative.  Records the fact that the equation
+  // "|lhs_dd| |opcode| |rhs_dds_non_canonical|" holds, and adds any
+  // corollaries, in the form of data synonym or equation facts, that follow
+  // from this and other known facts.
   void AddEquationFactRecursive(
-      const protobufs::DataDescriptor& lhs_dd_non_canonical, SpvOp opcode,
-      const std::vector<const protobufs::DataDescriptor*>&
-          rhs_dds_non_canonical,
+      const protobufs::DataDescriptor& lhs_dd, SpvOp opcode,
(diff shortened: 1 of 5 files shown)


source/fuzz/fact_manager.cpp
```diff
@@ -529,38 +543,41 @@ void FactManager::DataSynonymAndIdEquationFacts::AddFact(
     }
     rhs_dd_ptrs.push_back(synonymous_.Find(&rhs_dd));
   }
-  // We now have the equation in a form where it refers exclusively to
-  // equivalence class representatives.  Add it to our set of facts and work
-  // out any follow-on facts.
-  AddEquationFactRecursive(*lhs_dd_ptr, static_cast<SpvOp>(fact.opcode()),
+  AddEquationFactRecursive(lhs_dd, static_cast<SpvOp>(fact.opcode()),
                            rhs_dd_ptrs, context);
```
paulthomson, line 547: Perhaps you could still comment that we now add the fact, or just add some new lines to separate it from the code above?

source/fuzz/fact_manager.cpp
```diff
@@ -529,38 +543,41 @@ void FactManager::DataSynonymAndIdEquationFacts::AddFact(
...
+  for (auto rhs_dd_non_canonical : rhs_dds_non_canonical) {
+    assert(synonymous_.Exists(*rhs_dd_non_canonical) &&
+           "The RHS operands must be known to the equivalence relation.");
+    rhs_dds.push_back(synonymous_.Find(rhs_dd_non_canonical));
   }
 
-  if (id_equations_.count(&lhs_dd) == 0) {
+  // |lhs_dd_non_canonical| and |rhs_dds_non_canonical| should not be used from
+  // here onwards; their canonicalized forms, |lhs_dd| and |rhs_dds|, should be
+  // used instead.
```
paulthomson, line 569: Is it the case that we could just use `lhs_dd_non_canonical` and `rhs_dds_non_canonical`? But it is likely more efficient to do what you did? And is it true that `lhs_dd` and `rhs_dds` might become "non-canonical" part-way through this function? Could be worth noting this, if true.

source/fuzz/fact_manager.cpp
afd: Done.

source/fuzz/fact_manager.cpp
afd: Actually you are correct that the non-canonical versions can be used, which makes things simpler.  Done.

PR #3302, opened 2020-04-14
spirv-fuzz: Fix to outliner

afd:
Adds an extra condition on when a region can be outlined to avoid the
case where a region ends with a loop head but such that the loop's
continue target is in the region.  (Outlining such a region would mean
that the loop merge is in the original function and the continue target
in the outlined function.)


paulthomson:
Typo in PR description: "loop mere"?

afd:
Typo fixed, thanks.

PR #3303, opened 2020-04-15
BUILD.gn: Don't warn on unreachable code

Kangz:

PR #3304, opened 2020-04-15
Remove unreachable code.

s-perron:
@dneto0 Are you okay with this change or should we just delete the code in the robust buffer access code?  This is the alternative solution to https://github.com/KhronosGroup/SPIRV-Tools/pull/3303.


s-perron:
Merging now.  I can see the appveyor build passed, even if it is not showing it here.

PR #3305, opened 2020-04-15
Add debug information analysis

jaebaek:
We need an analysis for OpenCL.DebugInfo.100 extension instructions such
as a map between function id and its DebugFunction. This commit add an
analysis for it.


source/opt/debug_info_manager.cpp
```diff
@@ -0,0 +1,198 @@
...
+      auto var_id =
+          cpi->GetSingleWordOperand(kDebugDeclareOperandVariableIndex);
+      if (local_var_id_to_dbgdecl_.find(var_id) !=
+          local_var_id_to_dbgdecl_.end()) {
+        return;
+      }
+      local_var_id_to_dbgdecl_[var_id] = cpi;
+    }
+  });
+}
```
s-perron, line 194: You will need a way for the users of the analysis to update it.  When the ssa-rewriter creates new debug instruction, the analysis may have to be updated.  What kind of instructions will inlining create?  It could create new debug scope instructions, but the analysis won't reflect this change.

I would suggest something like the def-use manager.  Write one function that handles a single instruction.  Make that public.  Then have the version that looks at the entire module call the single instruction version on every instruction that matters.

source/opt/debug_info_manager.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+  id_to_dbg_inst_[result_id] = dbg_info_none_inst.get();
+
+  // Add to the front of |ext_inst_debuginfo_|.
+  return context()->module()->ext_inst_debuginfo_begin()->InsertBefore(
+      std::move(dbg_info_none_inst));
+}
+
+Instruction* DebugInfoManager::CloneDebugDeclare(uint32_t from, uint32_t to) {
+  auto debugdecl_it = local_var_id_to_dbgdecl_.find(from);
```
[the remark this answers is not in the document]
jaebaek: done

source/opt/debug_info_manager.cpp
s-perron: I still want to see what I suggested above.  Have one function called `AnalyzeDebugInst(Intruction* dbg_inst)` that is called by this function.  This is a single generic function that is needed for inlining.  How will inlining handle a DebugValue instruction that needs to be inlined?

source/opt/debug_info_manager.h
```diff
@@ -0,0 +1,132 @@
...
+
+// A class for analyzing, managing, and creating OpenCL.DebugInfo.100 extension
+// instructions.
+class DebugInfoManager {
+ public:
+  DebugInfoManager(IRContext* context);
+  // Constructs a def-use manager from the given |module|. All internal messages
+  // will be communicated to the outside via the given message |consumer|. This
+  // instance only keeps a reference to the |consumer|, so the |consumer| should
+  // outlive this instance.
```
s-perron, line 35: Copy-paste error?  This comment does not make sense.

source/opt/debug_info_manager.h
```diff
@@ -0,0 +1,132 @@
...
+    return !(lhs == rhs);
+  }
+
+  // Clones DebugDeclare or DebugValue for a local variable whose result id is
+  // |orig_var_id|. Set Variable operand of the new DebugDeclare as
+  // |new_var_id|. Return the new DebugDeclare or DebugValue. If this is the
+  // first DebugDeclare or DebugValue, keep it in |local_var_id_to_dbgdecl_|.
+  Instruction* CloneDebugDeclare(uint32_t orig_var_id, uint32_t new_var_id);
+
+  // Returns id of new DebugInlinedAt. Its Line operand is the line number
```
s-perron, line 54: Use a lower case "L" for "line".

source/opt/debug_info_manager.h
```diff
@@ -0,0 +1,132 @@
...
+  // of |line| if |line| is not nullptr. Otherwise, its Line operand
+  // is the line number of lexical scope of |scope|. Its Scope and Inlined
+  // operands are Scope and Inlined of |scope|. Note that this function puts
+  // the new DebugInlinedAt into the tail of debug instructions.
```
s-perron, line 58: This note should be worded as a promise "The returned instruction will be at the end of the debug instructions."  However, I think we should delete it.  In the future, if we want to use an already existing instruction, we may return one that is not at the end of the debug instructions.  So users should not rely on that.

source/opt/debug_info_manager.h
```diff
@@ -0,0 +1,132 @@
...
+  uint32_t CreateDebugInlinedAt(const Instruction* line,
+                                const DebugScope& scope);
+
+  // Creates a new DebugInfoNone instruction and returns it. In addition,
+  // insert the new DebugInfoNone instruction before the head of debug
+  // instructions.
```
s-perron, line 64: I still this should return an already existing instruction.  

The promise of the location should be removed unless you have a specific reason.  I could see a requirement that it be at the start to be useful to ensure that it can be referenced in any other debug instruction.

If you keep it, please reword to "The returned instruction will be at the start of the debug instructions."

source/opt/debug_info_manager.cpp
```diff
@@ -0,0 +1,213 @@
...
+  if (inlined_at == nullptr) return nullptr;
+  auto* new_inlined_at = inlined_at->Clone(context());
+  new_inlined_at->SetResultId(context()->TakeNextId());
+  RegisterDbgInst(new_inlined_at);
+  return new_inlined_at;
+}
+
+void DebugInfoManager::AnalyzeDebugInsts(Module& module) {
+  for (auto& inst : module.ext_inst_debuginfo()) {
+    RegisterDbgInst(&inst);
```
s-perron, line 194: Somewhere here, you can identify a DebugInfoNone instruction and cache it.  This way you will not have to create a new one.  If there is a requirment that it be at the start, then you can simple check the first instruction to see if it is a DebugInfoNone and cache that value.

source/opt/debug_info_manager.cpp
```diff
@@ -0,0 +1,213 @@
...
+  }
+  return inlined_at;
+}
+
+Instruction* DebugInfoManager::CloneDebugInlinedAt(uint32_t dbg_inlined_at_id) {
+  auto* inlined_at = GetDebugInlinedAt(dbg_inlined_at_id);
+  if (inlined_at == nullptr) return nullptr;
+  auto* new_inlined_at = inlined_at->Clone(context());
+  new_inlined_at->SetResultId(context()->TakeNextId());
+  RegisterDbgInst(new_inlined_at);
```
s-perron, line 188: Should be be registering a debug instruction that is not in the module yet?  Also, to better keep track of ownership (ie who is responsible for deleting the instruction, you should return a unique pointer.  Lastly, what is the reason for not putting it in the module?

source/opt/debug_info_manager.h
```diff
@@ -0,0 +1,132 @@
...
+  // exist, return nullptr.
+  Instruction* GetDebugFunction(uint32_t fn_id) {
+    auto dbg_fn_it = fn_id_to_dbg_fn_.find(fn_id);
+    return dbg_fn_it == fn_id_to_dbg_fn_.end() ? nullptr : dbg_fn_it->second;
+  }
+
+  // Clones DebugInlinedAt whose id is |dbg_inlined_at_id|. If
+  // |dbg_inlined_at_id| is not an id of DebugInlinedAt, returns nullptr.
+  // Note that this function does not insert the new DebugInlinedAt into
+  // debug instruction list of the module.
```
s-perron, line 81: The comment should be more definitive.  They are either hard requirements for the inputs or promises about the output.  When you use a term like "note", you drawing attention to something that they should be able to see somewhere else.  There is a photo and I ask you to note the person in the red coat.  You can look at the photo to see the person in the red coat.  We do not want people looking at the implementation or anything else.  We are making a promise: "This is true."  I've probably made this myself, and I want to try to get away from it.

In this case, something like "The new instruction will not be in the debug instruction of the module yet."

source/opt/debug_info_manager.h
```diff
@@ -0,0 +1,132 @@
...
+  void AnalyzeDebugInsts(Module& module);
+
+  // Returns the DebugDeclare instruction that corresponds to the variable
+  // with id |var_id|. Returns |nullptr| if one does not exists.
+  Instruction* GetDbgDeclareForVar(uint32_t var_id);
+
+  // Registers the DebugDeclare instruction |inst| into
+  // |local_var_id_to_dbgdecl_| using the variable operand of |inst| as a key.
+  // If |local_var_id_to_dbgdecl_| already has a key that corresponds to the
+  // variable operand of |inst|, it just returns without doing anything.
```
s-perron, line 98: Shouldn't it be an error if the key already exists with a different instruction?  Why would this not be a sign that something is wrong?

I believe that you do this because you want to clone debug declare instruction and register them right away.  I think this shows that it might be the wrong direction.  It should be up to inlining to first clone the function.  Update the variable to the new variable, and then you register to the new debug declare instruction.

This shows why we need a public function to register new debug instructions.  We do not have all of the necessary information when doing the cloning.

source/opt/debug_info_manager.h
```diff
@@ -0,0 +1,132 @@
...
+  // Returns the debug instruction whose id is |id|. Returns |nullptr| if one
+  // does not exists.
+  Instruction* GetDbgInst(uint32_t id);
+
+  // Registers the debug instruction |inst| into |id_to_dbg_inst_| using id of
+  // |inst| as a key.
+  void RegisterDbgInst(Instruction* inst);
+
+  // Registers the DebugFunction |inst| into |fn_id_to_dbg_fn_| using the
+  // function operand of |inst| as a key.
```
s-perron, line 110: Any special cases that need to be documented?  What happens if we have two debug function instructions for the same function?

jaebaek: I am also worried about the case. I added an `assert()` in the function that crashes if the function was already registered with a different `DebugFunction`.

I am not sure what comment it must be but I explained it "If the function operand of |inst| is already registered with a different DebugFunction, it will result in an assertion failure"

jaebaek:
Thank you for your comments! I updated the code based on the comments. PTAL

source/opt/debug_info_manager.h
s-perron: Or just say "The function referenced in |inst| must not already be registered."  don't mention the assert because assert are compiled out in release builds.

source/opt/debug_info_manager.h
```diff
@@ -45,24 +42,21 @@ class DebugInfoManager {
...
-  // the new DebugInlinedAt into the tail of debug instructions.
+  // operands are Scope and Inlined of |scope|.
   uint32_t CreateDebugInlinedAt(const Instruction* line,
                                 const DebugScope& scope);
 
-  // Creates a new DebugInfoNone instruction and returns it. In addition,
-  // insert the new DebugInfoNone instruction before the head of debug
-  // instructions.
-  Instruction* CreateDebugInfoNone();
+  // If |debug_info_none_inst_| is not a nullptr, returns it. Otherwise,
```
s-perron, line 55: No need to mention the internal implementation of the class.  You can simply say, "Returns a DebugInfoNone instruction."

source/opt/debug_info_manager.h
```diff
@@ -107,7 +93,9 @@ class DebugInfoManager {
   void RegisterDbgInst(Instruction* inst);
 
   // Registers the DebugFunction |inst| into |fn_id_to_dbg_fn_| using the
-  // function operand of |inst| as a key.
+  // function operand of |inst| as a key. If the function operand of |inst| is
+  // already registered with a different DebugFunction, this function will
+  // result in an assertion failure.
```
s-perron, line 98: You should not be talking about the internals of the class in the external interface.  You can simply say "Register the DebugFunction instruction |inst|.  The function referenced in |inst| must not already be registered."

source/opt/ir_context.cpp
```diff
@@ -412,7 +405,8 @@ void IRContext::KillOperandFromDebugInstructions(Instruction* inst) {
         continue;
       auto& operand = it->GetOperand(kDebugGlobalVariableOperandVariableIndex);
       if (operand.words[0] == id) {
-        operand.words[0] = GetOpenCL100DebugInfoNone()->result_id();
+        operand.words[0] =
+            get_debug_info_mgr()->GetDebugInfoNone()->result_id();
```
s-perron, line 409: Are we sure that the DebugInfoNone will be defined before |inst|, or are forward references allowed here?

jaebaek: Oh nice catch. If the input SPIR-V has `DebugInfoNone` after `DebugFunction` or `DebugGlobalVariable`, it causes illegal forward references. I updated `DebugInfoManager` so that it moves the exisitng `DebugInfoNone` to the beginning of the debug instruction list.

PR #3312, opened 2020-04-22
spirv-fuzz: Do not outline regions that end with a loop header

afd:
The outliner would outline regions ending with a loop header, making
the block containing the call to the outlined function serve as the
loop header.  This, however, is incorrect in general, since the whole
outlined function -- rather than just the exit block for the region --
would end up getting called every time the loop would iterate.

This change restricts the outliner so that the last block in a region
cannot be a loop header.

### DIFF
--- a/source/fuzz/transformation_outline_function.cpp
+++ b/source/fuzz/transformation_outline_function.cpp
@@ -200,21 +200,13 @@ bool TransformationOutlineFunction::IsApplicable(
       // It is OK (and typically expected) for the exit block of the region to
       // have successors outside the region.
       //
-      // It is also OK for the exit block to head a structured control flow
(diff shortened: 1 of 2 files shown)


paulthomson:
Maybe we should have one GitHub issue that lists several limitations of the outliner (and similarly for other groups of limitations). I suppose one issue per limitation can also work, but it might be nice to group them and then create individual issues only when we want to start or assign them.

afd:
Maybe.  However, we don't actually want to enhance the outliner to handle these situations.  Rather, we would like other transformations to get code into an outline-ready form.  I'm undecided as to whether we want to explicitly direct that process, or just have enough transformations in the mix that regions of code have a good chance of becoming outlinable.

PR #3316, opened 2020-04-23
spirv-fuzz: Do not allow adding stores to read-only pointers

afd:
Adds checks for PushConstant and UniformConstant storage classes (in
addition to the existing check for Input storage class).

### DIFF
--- a/source/fuzz/transformation_store.cpp
+++ b/source/fuzz/transformation_store.cpp
@@ -50,7 +50,8 @@ bool TransformationStore::IsApplicable(
   }
 
   // The pointer must not be read only.
-  if (pointer_type->GetSingleWordInOperand(0) == SpvStorageClassInput) {
+  if (TransformationStore::StorageClassIsReadOnly(static_cast<SpvStorageClass>(
+          pointer_type->GetSingleWordInOperand(0)))) {
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


source/fuzz/transformation_store.cpp
```diff
@@ -125,5 +126,18 @@ protobufs::Transformation TransformationStore::ToMessage() const {
...
+  switch (storage_class) {
+    case SpvStorageClassInput:
+    case SpvStorageClassPushConstant:
+    case SpvStorageClassUniformConstant:
+      // Read-only: cannot store to it.
+      return true;
+    default:
+      return false;
+  }
+}
```
paulthomson, line 140: There seems to already be `IsReadOnlyVariableShaders`. Could that be used? It looks more complex though.

afd: It appears that it only applies to OpVariable instructions.  I need to know whether a pointer is suitable for a store, and that pointer could have come from an access chain.

Looks like the code should be adaptable, though.  I have opened an issue (#3324) which I will reference from this code, and will refactor later if it turns out to be reusable.

afd: I have made a PR - #3325 - that should allow us to re-use this functionality.

afd: Adapted this PR to re-use that functionality.

PR #3320, opened 2020-04-24
Add spvtools::opt::Operand::AsLiteralUint64

dneto0:


dneto0:
This is handy to have.  I plan to use it in the SPIR-V to Tint AST conversion, for collecting case values.

source/opt/instruction.h
```diff
@@ -92,6 +92,19 @@ struct Operand {
   // Returns a string operand as a std::string.
   std::string AsString() const { return AsCString(); }
 
+  // Returns a literal integer operand as a uint64_t
+  uint64_t AsLiteralUint64() const {
+    assert(type == SPV_OPERAND_TYPE_TYPED_LITERAL_NUMBER);
+    assert(1 <= words.size());
+    assert(words.size() <= 2);
```
dj2, line 99: Doesn't this mean that words.size() == 1 is valid, which means words[1] will be out of bounds?

PR #3321, opened 2020-04-25
spirv-reduce: Add the capability to remove unused uniforms and similar

afd:
Adds a reduction opportunity to get rid of a resource variable (such
as a uniform or image) that is unreferenced except by descriptor set,
binding, readonly or writeonly decorations, getting rid of said
decorations at the same time.

### DIFF
--- a/source/reduce/remove_unused_instruction_reduction_opportunity_finder.cpp
+++ b/source/reduce/remove_unused_instruction_reduction_opportunity_finder.cpp
@@ -140,7 +140,8 @@ bool RemoveUnusedInstructionReductionOpportunityFinder::
       decoration = inst.GetSingleWordInOperand(2u);
       break;
     default:
-      // The instruction is not a decoration.
+      // The instruction is not a decoration.  It is legitimate for this to be
+      // reached: it allows the method to be invoked on arbitrary instructions.
       return false;
   }
 
--- a/source/reduce/remove_unused_instruction_reduction_opportunity_finder.cpp
+++ b/source/reduce/remove_unused_instruction_reduction_opportunity_finder.cpp
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "source/reduce/remove_unreferenced_instruction_reduction_opportunity_finder.h"
+#include "source/reduce/remove_unused_instruction_reduction_opportunity_finder.h"
 
 #include "source/opcode.h"
 #include "source/opt/instruction.h"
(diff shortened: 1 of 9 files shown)


source/reduce/remove_resource_variable_reduction_opportunity.cpp
```diff
@@ -0,0 +1,70 @@
...
+        use_ids_to_kill.push_back(user->result_id());
+      });
+
+  // Kill each of these decorations.
+  for (auto id : use_ids_to_kill) {
+    context_->KillInst(context_->get_def_use_mgr()->GetDef(id));
+  }
+
+  // Kill the variable.
+  context_->KillInst(variable_);
```
paulthomson, line 51: I think this already kills the corresponding names and decorations. Also, it may do so more carefully than the above.

afd: Good catch - and by exploiting that fact, this reduction opportunity becomes identical to remove_instruction_reduction_opportunity - i.e., there's nothing special to be done.  I have simplified the PR on that basis.

afd:
@paulthomson I changed this PR a lot so that the new functionality to remove decorations and related variables/types together works with the existing infrastructure to remove instructions.

I opted for an approach that treats a decoration as either being independently removable or not, so that independently-removable declarations will be removed as individual instructions, and other decorations will only be removed when getting rid of the decorated instruction.

It means that, for instance, if a variable is decorated with Relaxed and also with some other more intimate things, the Relaxed will need to be gotten rid of before the variable can be gotten rid of.

source/reduce/remove_unused_instruction_reduction_opportunity_finder.cpp
```diff
@@ -139,9 +111,53 @@ RemoveUnreferencedInstructionReductionOpportunityFinder::
...
+    case SpvOpDecorateString:
+      decoration = inst.GetSingleWordInOperand(1u);
+      break;
+    case SpvOpMemberDecorate:
+    case SpvOpMemberDecorateString:
+      decoration = inst.GetSingleWordInOperand(2u);
+      break;
+    default:
+      // The instruction is not a decoration.
+      return false;
```
paulthomson, line 144: I suppose this could be an assert false?

afd: No, actually - we rely on this path to allow asking the question on arbitrary instructions.  I will add a comment.

PR #3325, opened 2020-04-28
Generalize IsReadOnlyVariable() to apply to pointers

### LINKED ISSUE #3324: Could IsReadOnlyVariable() be generalized to apply to pointer instructions?
I would like to be able to check whether an instruction of pointer type should be considered read-only.

`Instruction::IsReadOnlyVariable()` seems to do what I want, except that it is designed to work on OpVariable instructions.  However, looking at the code it seems like it should be suitable for application to any instruction whose result type is OpTypePointer.  Some of the checks would not be relevant for non-variables, e.g. checking for `SpvDecorationNonWritable`, but it looks like it would be harmless to do such checks for pointers that are not from `SpvOpVariable`.

@s-perron can you think of any reasons why this wouldn't work?  If not then I'll do a PR.

afd:
Generalizes the IsReadOnlyVariable() method, and related methods, so
that they can be used to ask whether pointer result ids are read-only.

Fixes #3324.

### DIFF
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -415,7 +415,7 @@ bool Instruction::IsVulkanUniformBuffer() const {
 }
 
 bool Instruction::IsReadOnlyPointerShaders() const {
-  if (!type_id()) {
+  if (type_id() == 0) {
     return false;
   }
 
@@ -454,7 +454,7 @@ bool Instruction::IsReadOnlyPointerShaders() const {
 }
 
 bool Instruction::IsReadOnlyPointerKernel() const {
-  if (!type_id()) {
+  if (type_id() == 0) {
     return false;
   }
 
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -29,7 +29,7 @@ namespace {
 // Indices used to get particular operands out of instructions using InOperand.
 const uint32_t kTypeImageDimIndex = 1;
 const uint32_t kLoadBaseIndex = 0;
-const uint32_t kVariableStorageClassIndex = 0;
+const uint32_t kPointerTypeStorageClassIndex = 0;
 const uint32_t kTypeImageSampledIndex = 5;
 
 // Constants for OpenCL.DebugInfo.100 extension instructions.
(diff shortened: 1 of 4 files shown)


source/opt/instruction.cpp
```diff
@@ -409,9 +414,18 @@ bool Instruction::IsVulkanUniformBuffer() const {
   return is_block;
 }
 
-bool Instruction::IsReadOnlyVariableShaders() const {
-  uint32_t storage_class = GetSingleWordInOperand(kVariableStorageClassIndex);
+bool Instruction::IsReadOnlyPointerShaders() const {
+  if (!type_id()) {
```
s-perron, line 418: type_id is not a bool, so I prefer a comparison with 0.

```
if (type_id == 0) {
```

source/opt/instruction.cpp
```diff
@@ -439,8 +453,19 @@ bool Instruction::IsReadOnlyVariableShaders() const {
   return is_nonwritable;
```
s-perron, line 453: When this was suppose to be for variables only, this check would be appropriate because the decoration is on the variable.  However, will there be a need for this to be expanded?  For example, if you ask if the result of an OpAccessChain is read-only, should it recursively check if the input pointer is read-only?

You don't need to implement it.  This function was only intended to return turn if is definitely know that it is read only.  However, returning false does not guarantee that it is not read-only.  How please leave a comment if you are not going to implement it.

afd: Good point!

I'll leave a comment to this effect, because the general solution could be fairly complex (for example, knowing whether a function parameter is definitely read only would I suppose require analysing all calls to the function).

source/opt/instruction.cpp
afd: Done.

source/opt/instruction.cpp
afd: Done.

afd:
@s-perron I have made the changes, added tests, and extended some existing tests a bit too.

s-perron:
I approve.  Fix up the CI failures, and you can merge.

afd:
@s-perron Thanks!  Note that the CI failed because I'd gotten a test wrong: I had meant to check that if a variable is regarded as read-only due to being decorated with NonWritable, an object copy of this variable will not be regarded as read-only, due to not having this decoration.  This illustrates the point you made that in your first batch of comments.  (I had asserted the wrong thing, and had run the wrong tests locally!)

I'll indeed merge if the CI passes.

PR #3328, opened 2020-04-30
[spirv-opt] refactor inlining pass

jaebaek:
In the function inlining pass, a lamda function contains
lots of logic and it is too complicated. This commit breaks
it down into small functions to be more readable.

### DIFF
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -312,61 +312,6 @@ std::unique_ptr<BasicBlock> InlinePass::InsertHeaderBlockForSingleTripLoop(
   return new_blk_ptr;
 }
 
-std::unique_ptr<BasicBlock> InlinePass::InlineInstsBeforeEntryBlock(
-    std::vector<std::unique_ptr<BasicBlock>>* new_blocks,
-    std::unordered_map<uint32_t, uint32_t>* callee2caller,
-    std::unordered_map<uint32_t, Instruction*>* preCallSB,
-    std::unique_ptr<BasicBlock>* single_trip_loop_cont_blk,
-    uint32_t* returnLabelId, BasicBlock::iterator call_inst_itr,
-    UptrVectorIterator<BasicBlock> call_block_itr, bool caller_is_loop_header,
-    Function* calleeFn) {
-  // First block needs to use label of original block
-  // but map callee label in case of phi reference.
-  uint32_t labelId = call_block_itr->id();
-  uint32_t entry_blk_label_id = calleeFn->begin()->GetLabelInst()->result_id();
-  (*callee2caller)[entry_blk_label_id] = labelId;
-  std::unique_ptr<BasicBlock> new_blk_ptr =
-      MakeUnique<BasicBlock>(NewLabel(labelId));
-
-  // Copy instructions of original caller block up to call instruction.
-  CopyInstsBeforeEntryBlock(preCallSB, new_blk_ptr.get(), call_inst_itr,
-                            call_block_itr);
-
-  if (caller_is_loop_header &&
-      (*(calleeFn->begin())).GetMergeInst() != nullptr) {
-    // We can't place both the caller's merge instruction and
-    // another merge instruction in the same block.  So split the
-    // calling block. Insert an unconditional branch to a new guard
-    // block.  Later, once we know the ID of the last block,  we
-    // will move the caller's OpLoopMerge from the last generated
-    // block into the first block. We also wait to avoid
-    // invalidating various iterators.
-    new_blk_ptr = AddGuardBlock(new_blocks, callee2caller,
-                                std::move(new_blk_ptr), entry_blk_label_id);
-    if (new_blk_ptr == nullptr) return nullptr;
-  }
-
-  if (early_return_funcs_.find(calleeFn->result_id()) !=
-      early_return_funcs_.end()) {
-    // If callee has early return, insert a header block for
-    // single-trip loop that will encompass callee code.  Start
-    // postheader block.
-    //
-    // Note: Consider the following combination:
-    //  - the caller is a single block loop
-    //  - the callee does not begin with a structure header
-    //  - the callee has multiple returns.
-    // We still need to split the caller block and insert a guard
-    // block. But we only need to do it once. We haven't done it yet,
-    // but the single-trip loop header will serve the same purpose.
-    new_blk_ptr = InsertHeaderBlockForSingleTripLoop(
-        new_blocks, callee2caller, single_trip_loop_cont_blk, returnLabelId,
-        std::move(new_blk_ptr), entry_blk_label_id);
-    if (new_blk_ptr == nullptr) return nullptr;
-  }
-  return new_blk_ptr;
-}
-
 InstructionList::iterator InlinePass::AddStoresForVariableInitializers(
     std::unordered_map<uint32_t, uint32_t>* callee2caller,
     std::unique_ptr<BasicBlock>* new_blk_ptr,
@@ -660,13 +605,50 @@ bool InlinePass::GenInlineCode(
     return false;
   }
 
-  // Move instructions of the caller function up to the call instruction.
+  // First block needs to use label of original block
+  // but map callee label in case of phi reference.
+  uint32_t entry_blk_label_id = calleeFn->begin()->GetLabelInst()->result_id();
+  callee2caller[entry_blk_label_id] = call_block_itr->id();
+  std::unique_ptr<BasicBlock> new_blk_ptr =
+      MakeUnique<BasicBlock>(NewLabel(call_block_itr->id()));
+
+  // Copy instructions of original caller block up to call instruction.
+  CopyInstsBeforeEntryBlock(&preCallSB, new_blk_ptr.get(), call_inst_itr,
+                            call_block_itr);
+
+  if (caller_is_loop_header &&
+      (*(calleeFn->begin())).GetMergeInst() != nullptr) {
+    // We can't place both the caller's merge instruction and
+    // another merge instruction in the same block.  So split the
+    // calling block. Insert an unconditional branch to a new guard
+    // block.  Later, once we know the ID of the last block,  we
+    // will move the caller's OpLoopMerge from the last generated
+    // block into the first block. We also wait to avoid
+    // invalidating various iterators.
+    new_blk_ptr = AddGuardBlock(new_blocks, &callee2caller,
+                                std::move(new_blk_ptr), entry_blk_label_id);
+    if (new_blk_ptr == nullptr) return false;
+  }
+
   uint32_t returnLabelId = 0;
-  std::unique_ptr<BasicBlock> new_blk_ptr = InlineInstsBeforeEntryBlock(
-      new_blocks, &callee2caller, &preCallSB, &single_trip_loop_cont_blk,
-      &returnLabelId, call_inst_itr, call_block_itr, caller_is_loop_header,
-      calleeFn);
-  if (new_blk_ptr == nullptr) return false;
+  if (early_return_funcs_.find(calleeFn->result_id()) !=
+      early_return_funcs_.end()) {
+    // If callee has early return, insert a header block for
+    // single-trip loop that will encompass callee code.  Start
+    // postheader block.
+    //
+    // Note: Consider the following combination:
+    //  - the caller is a single block loop
+    //  - the callee does not begin with a structure header
+    //  - the callee has multiple returns.
+    // We still need to split the caller block and insert a guard
+    // block. But we only need to do it once. We haven't done it yet,
+    // but the single-trip loop header will serve the same purpose.
+    new_blk_ptr = InsertHeaderBlockForSingleTripLoop(
+        new_blocks, &callee2caller, &single_trip_loop_cont_blk, &returnLabelId,
+        std::move(new_blk_ptr), entry_blk_label_id);
+    if (new_blk_ptr == nullptr) return false;
+  }
 
   // Create return var if needed.
   const uint32_t calleeTypeId = calleeFn->type_id();
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -232,6 +232,166 @@ bool InlinePass::CloneSameBlockOps(
   });
 }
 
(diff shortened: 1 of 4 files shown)


test/opt/inline_test.cpp
```diff
@@ -1442,33 +1442,33 @@ OpFunctionEnd
       R"(%false = OpConstantFalse %bool
```
jaebaek, line 1442: I changed the test but it only changes the result id because of the id assignment order.

source/opt/inline_pass.cpp
```diff
@@ -284,313 +637,74 @@ bool InlinePass::GenInlineCode(
...
-            }
-            new_blk_ptr->AddInstruction(std::move(cp_inst));
-          } break;
+  bool multiBlocks = earlyReturn;
+  new_blk_ptr =
+      InlineBasicBlocks(new_blocks, &callee2caller, std::move(new_blk_ptr),
+                        &returnLabelId, &prevInstWasReturn, &multiBlocks,
+                        calleeFn, callee_result_ids, returnVarId);
+  if (new_blk_ptr == nullptr) return false;
+
```
jaebaek, line 658: I will refactor the following lines of code until the end of `InlinePass::GenInlineCode()`.

source/opt/inline_pass.cpp
```diff
@@ -266,6 +611,14 @@ bool InlinePass::GenInlineCode(
     return false;
   }
 
+  // Inline instructions of the caller function up to the call instruction.
```
s-perron, line 614: Don't call moving instructions in the caller "inline".  Say you are moving them instead.

source/opt/inline_pass.cpp
```diff
@@ -266,6 +611,14 @@ bool InlinePass::GenInlineCode(
...
+  uint32_t returnLabelId = 0;
+  std::unique_ptr<BasicBlock> new_blk_ptr = InlineInstsBeforeEntryBlock(
+      new_blocks, &callee2caller, &preCallSB, &single_trip_loop_cont_blk,
+      &returnLabelId, call_inst_itr, call_block_itr, caller_is_loop_header,
+      calleeFn);
```
s-perron, line 619: I like this, but it does have a lot of parameters.  We should see if we can reduce the number of parameters.

source/opt/inline_pass.cpp
jaebaek: done

source/opt/inline_pass.cpp
jaebaek: I removed this function to reduce the number of arguments.
Instead, I inlined the contents of this function.

jaebaek:
@s-perron It's weird. It seems appveyor execution has some issues.
Even though the testing is not done yet, could you please review this PR?

source/opt/inline_pass.cpp
```diff
@@ -284,333 +670,60 @@ bool InlinePass::GenInlineCode(
     if (rid != 0) callee_result_ids.insert(rid);
   });
```
s-perron, line 671: I think we can change this to immediately assign new ids to `callee_result_ids`.  Then we can delete the references to `callee_result_ids`, since it will just be redundant information.  This could also help us simplify the code.  The only reason inline fails is because we run out of ids.  If we limit the places were we take the next id, then we do not have to check for failures in as many places.

If for some reason we cannot assign new ids right away, we can map all of the result ids to 0 initially.  Then that turns into the clue that it is a forward reference.  Either way, we do not need `callee_result_ids`.

source/opt/inline_pass.cpp
```diff
@@ -284,333 +670,60 @@ bool InlinePass::GenInlineCode(
...
-  // blocks in the loop.  Instead, it should be placed at the end of the first
-  // block.  We'll wait to move the OpLoopMerge until the end of the regular
-  // inlining logic, and only if necessary.
-  bool caller_is_loop_header = false;
-  if (call_block_itr->GetLoopMergeInst()) {
-    caller_is_loop_header = true;
+  // Inline the entry block of the callee function.
+  bool prevInstWasReturn = false;
+  if (!InlineEntryBlock(&callee2caller, &new_blk_ptr, &returnLabelId,
+                        &prevInstWasReturn, calleeFn, callee_result_ids,
```
s-perron, line 676: You can probably pass in the first basic block instead of `calleeFn`.

source/opt/inline_pass.cpp
```diff
@@ -284,333 +670,60 @@ bool InlinePass::GenInlineCode(
...
-      (*(calleeFn->begin())).GetMergeInst() != nullptr;
-
-  // Clone and map callee code. Copy caller block code to beginning of
-  // first block and end of last block.
-  bool prevInstWasReturn = false;
-  uint32_t singleTripLoopHeaderId = 0;
-  uint32_t singleTripLoopContinueId = 0;
-  uint32_t returnLabelId = 0;
+  // Inline blocks of the callee function other than the entry block.
   bool multiBlocks = false;
```
s-perron, line 682: I think we can get rid of this as well.  All it is interested in is if inlining created more than 1 basic block.  This can happen in two ways: the callee has multiple basic block or has an early return.  However, these turn into the same thing: if there is an early return, then there must be more than one basic block.  All you need to check is if the callee has more than 1 basic block.

There is no need for the variable.  Would replace the use in the call to `MoveCallerInstsAfterFunctionCall` by checking the size of `new_blocks`.

jaebaek: done

source/opt/inline_pass.cpp
```diff
@@ -232,6 +233,338 @@ bool InlinePass::CloneSameBlockOps(
...
+  (*callee2caller)[entry_blk_label_id] = postHeaderId;
+
+  *single_trip_loop_cont_blk =
+      MakeUnique<BasicBlock>(NewLabel(singleTripLoopContinueId));
+  uint32_t false_id = GetFalseId();
+  if (false_id == 0) {
+    return nullptr;
+  }
+  AddBranchCond(false_id, singleTripLoopHeaderId, *returnLabelId,
+                single_trip_loop_cont_blk);
```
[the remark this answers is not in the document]
jaebaek: I removed the code handling early returns

source/opt/inline_pass.cpp
```diff
@@ -266,6 +605,53 @@ bool InlinePass::GenInlineCode(
...
+    // invalidating various iterators.
+    new_blk_ptr = AddGuardBlock(new_blocks, &callee2caller,
+                                std::move(new_blk_ptr), entry_blk_label_id);
+    if (new_blk_ptr == nullptr) return false;
+  }
+
+  uint32_t returnLabelId = 0;
+  bool earlyReturn = false;
+  if (early_return_funcs_.find(calleeFn->result_id()) !=
+      early_return_funcs_.end()) {
```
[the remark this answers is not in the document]
jaebaek: updated.

source/opt/inline_pass.cpp
```diff
@@ -284,333 +670,60 @@ bool InlinePass::GenInlineCode(
...
-  // If the caller is a loop header and the callee has multiple blocks, then the
-  // normal inlining logic will place the OpLoopMerge in the last of several
-  // blocks in the loop.  Instead, it should be placed at the end of the first
-  // block.  We'll wait to move the OpLoopMerge until the end of the regular
-  // inlining logic, and only if necessary.
-  bool caller_is_loop_header = false;
-  if (call_block_itr->GetLoopMergeInst()) {
-    caller_is_loop_header = true;
+  // Inline the entry block of the callee function.
+  bool prevInstWasReturn = false;
```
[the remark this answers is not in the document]
jaebaek: I agree with you but simply removing `prevInstWasReturn` caused wrong results for early return cases that are not handled by the merge-return pass. For example, see the following code:

```
OpCapability Shader
OpMemoryModel Logical GLSL450
OpSource GLSL 450
OpName %main "main"
OpName %foo_ "foo("
%void = OpTypeVoid
%4 = OpTypeFunction %void
%bool = OpTypeBool
%true = OpConstantTrue %bool
%foo_ = OpFunction %void None %4
%7 = OpLabel
OpSelectionMerge %8 None
OpBranchConditional %true %9 %8
%8 = OpLabel
OpReturn
%9 = OpLabel
OpBranch %8
OpFunctionEnd
%main = OpFunction %void None %4
%10 = OpLabel
%11 = OpFunctionCall %void %foo_
OpReturn
OpFunctionEnd
```

`%foo_` has `OpReturn` before the tail block. I ran the merge-return pass on it but the merge-return pass does not move the `OpReturn`. In this case, we need `prevInstWasReturn`.

However, I agree with you and I have a different idea to simplify this code. PTAL.

source/opt/inline_pass.cpp
```diff
@@ -277,341 +551,69 @@ bool InlinePass::GenInlineCode(
...
-          } break;
+  calleeFn->WhileEachInst(
+      [&callee2caller, &returnLabelId, this](const Instruction* cpi) {
+        // Create set of callee result ids. Used to detect forward references
+        const uint32_t rid = cpi->result_id();
+        if (rid != 0 && callee2caller.find(rid) == callee2caller.end()) {
+          const uint32_t nid = context()->TakeNextId();
+          if (nid == 0) return false;
+          callee2caller[rid] = nid;
+        }
```
jaebaek, line 563: I added this part to handle the cases that include `OpReturn` before the last basic block, which we cannot simply reply on the merge-return pass.

For example,
```
OpCapability Shader
OpMemoryModel Logical GLSL450
OpSource GLSL 450
OpName %main "main"
OpName %foo_ "foo("
%void = OpTypeVoid
%4 = OpTypeFunction %void
%bool = OpTypeBool
%true = OpConstantTrue %bool
%foo_ = OpFunction %void None %4
%7 = OpLabel
OpSelectionMerge %8 None
OpBranchConditional %true %9 %8
%8 = OpLabel
OpReturn
%9 = OpLabel
OpBranch %8
OpFunctionEnd
%main = OpFunction %void None %4
%10 = OpLabel
%11 = OpFunctionCall %void %foo_
OpReturn
OpFunctionEnd
```

`%foo_` has `OpReturn` in the middle but running the merge-return pass on it does not update it.

This new code simplifies this function inling pass.

source/opt/inline_pass.cpp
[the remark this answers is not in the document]
s-perron: We can get merge-return to handle this by having it reorder the block, or we can change the order in which inlining traverses the basic block.  I'll take a look.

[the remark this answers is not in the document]
s-perron: You can try going through the block in reverse post order.  You can use `ForEachBlockInReversePostOrder` in cfg.h for that I believe.

jaebaek:
Remove the code for handling early return corner cases. PTAL.

source/opt/inline_pass.cpp
```diff
@@ -232,6 +233,239 @@ bool InlinePass::CloneSameBlockOps(
...
+    Function* calleeFn, uint32_t returnVarId) {
+  auto callee_block_itr = calleeFn->begin();
+  ++callee_block_itr;
+
+  while (callee_block_itr != calleeFn->end()) {
+    const auto mapItr =
+        callee2caller->find(callee_block_itr->GetLabelInst()->result_id());
+    uint32_t labelId = (mapItr != callee2caller->end())
+                           ? mapItr->second
+                           : context()->TakeNextId();
```
s-perron, line 398: Shouldn't these have already been given a new id in the code at line 555.

source/opt/inline_pass.cpp
```diff
@@ -266,6 +506,40 @@ bool InlinePass::GenInlineCode(
...
+                                std::move(new_blk_ptr), entry_blk_label_id);
+    if (new_blk_ptr == nullptr) return false;
+  }
+
+  if (early_return_funcs_.find(calleeFn->result_id()) !=
+      early_return_funcs_.end()) {
+    // We rely on the merge-return pass to handle the early return case
+    // in advance.
+    std::string message =
+        "Early return. Handle it using merge-return pass in advance.";
```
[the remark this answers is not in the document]
s-perron: This check should also be moved earlier before you actaully do any changes to the functions?  It might even belong in a different function.  Try to find a function related to "Can be inlined".  Also note that it outputs a message, but still tried to do the inlining.  That is wrong.

source/opt/inline_pass.cpp
```diff
@@ -277,341 +551,64 @@ bool InlinePass::GenInlineCode(
...
-                  return false;
-                }
-                callee2caller[rid] = nid;
-              }
-              cp_inst->SetResultId(nid);
-              get_decoration_mgr()->CloneDecorations(rid, nid);
-            }
-            new_blk_ptr->AddInstruction(std::move(cp_inst));
-          } break;
+  calleeFn->WhileEachInst(
```
s-perron, line 555: This should probably be moved earlier so that any uses of callee2caller can assume that all ids are already in the map.  This will reduce the amount of error checking code that is needed.

s-perron:
Looking at it now.  I think we should just fix up the message and the issues related to this comment https://github.com/KhronosGroup/SPIRV-Tools/pull/3328#discussion_r424555527.

Then we could merge this and then make the other changes in a separate PR.  This is already a big improvement.

source/opt/inline_pass.cpp
```diff
@@ -277,341 +551,64 @@ bool InlinePass::GenInlineCode(
...
+      [&callee2caller, &returnLabelId, this](const Instruction* cpi) {
+        // Create set of callee result ids. Used to detect forward references
+        const uint32_t rid = cpi->result_id();
+        if (rid != 0 && callee2caller.find(rid) == callee2caller.end()) {
+          const uint32_t nid = context()->TakeNextId();
+          if (nid == 0) return false;
+          callee2caller[rid] = nid;
+        }
+        // Create returnLabelId if we have OpUnreachable, OpKill, or return
+        // before the tail block.
```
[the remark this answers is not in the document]
jaebaek: I just simplified the code related to `returnLabelId`. I simply put all related code to `InlineReturn()`. PTAL.

jaebaek:
Ah .. give me a second. I forgot to update the early return case check. I will send one more commit.

jaebaek:
Updated. PTAL!

s-perron:
Sorry I merged too quick.  The android build still looked like it was failing.

PR #3329, opened 2020-04-30
spirv-reduce: Remove unused struct members

afd:
Adds a reduction pass to remove unused members from structs.


source/reduce/remove_unused_struct_member_reduction_opportunity_finder.cpp
```diff
@@ -0,0 +1,191 @@
...
+        unused_member_to_structs.insert(
+            {member, std::set<opt::Instruction*>()});
+      }
+      unused_member_to_structs.at(member).insert(&type_or_value);
+    }
+  }
+
+  // We now go through every instruction that might index into a struct, and
+  // refine our tracking of which struct members are used based on the struct
+  // indexing we observe.
```
paulthomson, line 75: Maybe add a comment: we cannot just go through all uses of a struct type because the type is not necessarily even referenced when walking arrays of structs of structs etc.

afd: Done.

paulthomson:
Formatting?

PR #3331, opened 2020-05-01
Preserve debug info for wrap-opkill

jaebaek:


source/opt/wrap_opkill.cpp
```diff
@@ -70,6 +73,7 @@ bool WrapOpKill::ReplaceWithFunctionCall(Instruction* inst) {
     if (undef == nullptr) {
       return false;
     }
+    undef->UpdateDebugInfo(inst);
```
s-perron, line 76: This should not been needed.  The undef and the return that follow are compiler generated instructions.  They do not represent code in the HLSL (or whatever the original source was).  They are not in any scope, and cannot actaully be reached.

source/opt/wrap_opkill.cpp
```diff
@@ -105,7 +110,7 @@ uint32_t WrapOpKill::GetVoidFunctionTypeId() {
   return type_mgr->GetTypeInstruction(&func_type);
 }
 
-uint32_t WrapOpKill::GetOpKillFuncId() {
+uint32_t WrapOpKill::GetOpKillFuncId(Instruction* dbg_info) {
```
s-perron, line 113: I don't think we should do this change.  It is a tough one because we want a single OpKill Function, which can be called in many places.  In this case, you will use the debug info for the first location of an OpKill, but that could be completely wrong for the next instance.

I would prefer to leave the OpKill with no debug info.  The debug info on the call is more informative.

source/opt/wrap_opkill.cpp
jaebaek: removed

source/opt/wrap_opkill.cpp
jaebaek: Ok. I add the debug information only for `OpFunctionCall` for the new function that executes `OpKill`.

PR #3336, opened 2020-05-08
Add adjust branch weights transformation

andreperezmaselco:
In this PR, the classes that represent the adjust branch weights transformation and fuzzer pass were implemented. This transformation adjusts the branch weights of a `OpBranchConditional` instruction.

### DIFF
--- a/source/fuzz/fuzzer_context.h
+++ b/source/fuzz/fuzzer_context.h
@@ -192,8 +192,8 @@ class FuzzerContext {
   uint32_t GetMaximumEquivalenceClassSizeForDataSynonymFactClosure() {
     return max_equivalence_class_size_for_data_synonym_fact_closure_;
   }
-  uint32_t GetRandomUint32(uint32_t bound) {
-    return random_generator_->RandomUint32(bound);
+  uint32_t GetRandomIndexForAccessChain(uint32_t composite_size_bound) {
+    return random_generator_->RandomUint32(composite_size_bound);
   }
   uint32_t GetRandomLoopControlPartialCount() {
     return random_generator_->RandomUint32(max_loop_control_partial_count_);
@@ -203,6 +203,18 @@ class FuzzerContext {
   }
   uint32_t GetRandomLoopLimit() {
     return random_generator_->RandomUint32(max_loop_limit_);
+  }
+  std::pair<uint32_t, uint32_t> GetRandomBranchWeights() {
+    std::pair<uint32_t, uint32_t> branch_weights = {0, 0};
+
+    while (branch_weights.first == 0 && branch_weights.second == 0) {
+      // Using INT32_MAX to do not overflow UINT32_MAX when the branch weights
+      // are added together.
+      branch_weights.first = random_generator_->RandomUint32(INT32_MAX);
+      branch_weights.second = random_generator_->RandomUint32(INT32_MAX);
+    }
+
+    return branch_weights;
   }
   uint32_t GetRandomSizeForNewArray() {
     // Ensure that the array size is non-zero.
--- a/source/fuzz/fuzzer_context.h
+++ b/source/fuzz/fuzzer_context.h
@@ -136,6 +136,9 @@ class FuzzerContext {
   uint32_t GetChanceOfAddingVectorType() {
     return chance_of_adding_vector_type_;
   }
+  uint32_t GetChanceOfAdjustingBranchWeights() {
+    return chance_of_adjusting_branch_weights_;
+  }
   uint32_t GetChanceOfAdjustingFunctionControl() {
     return chance_of_adjusting_function_control_;
   }
@@ -189,8 +192,8 @@ class FuzzerContext {
   uint32_t GetMaximumEquivalenceClassSizeForDataSynonymFactClosure() {
     return max_equivalence_class_size_for_data_synonym_fact_closure_;
   }
-  uint32_t GetRandomIndexForAccessChain(uint32_t composite_size_bound) {
-    return random_generator_->RandomUint32(composite_size_bound);
+  uint32_t GetRandomUint32(uint32_t bound) {
+    return random_generator_->RandomUint32(bound);
   }
   uint32_t GetRandomLoopControlPartialCount() {
     return random_generator_->RandomUint32(max_loop_control_partial_count_);
@@ -231,6 +234,7 @@ class FuzzerContext {
   uint32_t chance_of_adding_no_contraction_decoration_;
   uint32_t chance_of_adding_store_;
   uint32_t chance_of_adding_vector_type_;
+  uint32_t chance_of_adjusting_branch_weights_;
   uint32_t chance_of_adjusting_function_control_;
   uint32_t chance_of_adjusting_loop_control_;
   uint32_t chance_of_adjusting_memory_operands_mask_;
--- a/source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
+++ b/source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
@@ -41,12 +41,12 @@ void FuzzerPassAdjustBranchWeights::Apply() {
         instruction->NumOperands() == 5 &&
         fuzzer_context->ChoosePercentage(
             GetFuzzerContext()->GetChanceOfAdjustingBranchWeights())) {
-      uint32_t weight_for_true_label =
-          fuzzer_context->GetRandomUint32(INT32_MAX);
-      uint32_t weight_for_false_label =
-          fuzzer_context->GetRandomUint32(INT32_MAX - weight_for_true_label);
-      std::vector<uint32_t> branch_weights = {weight_for_true_label,
-                                              weight_for_false_label};
+      std::pair<uint32_t, uint32_t> branch_weights = {0, 0};
+      while (branch_weights.first == 0 && branch_weights.second == 0) {
+        branch_weights.first = fuzzer_context->GetRandomUint32(INT32_MAX);
+        branch_weights.second =
+            fuzzer_context->GetRandomUint32(INT32_MAX - branch_weights.first);
+      }
 
       auto instruction_descriptor =
           MakeInstructionDescriptor(ir_context, instruction);
--- a/source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
+++ b/source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
@@ -1,0 +1,61 @@
+// Copyright (c) 2020 André Perez Maselco
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_adjust_branch_weights.h"
+
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/instruction_descriptor.h"
+#include "source/fuzz/transformation_adjust_branch_weights.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassAdjustBranchWeights::FuzzerPassAdjustBranchWeights(
+    opt::IRContext* ir_context, TransformationContext* transformation_context,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassAdjustBranchWeights::~FuzzerPassAdjustBranchWeights() = default;
+
+void FuzzerPassAdjustBranchWeights::Apply() {
+  auto ir_context = GetIRContext();
+  auto fuzzer_context = GetFuzzerContext();
+  // For all OpBranchConditional instructions with branch weights,
+  // randomly applies the transformation.
+  ir_context->module()->ForEachInst([this, ir_context, fuzzer_context](
+                                        opt::Instruction* instruction) {
+    if (instruction->opcode() == SpvOpBranchConditional &&
+        instruction->NumOperands() == 5 &&
+        fuzzer_context->ChoosePercentage(
+            GetFuzzerContext()->GetChanceOfAdjustingBranchWeights())) {
+      uint32_t weight_for_true_label =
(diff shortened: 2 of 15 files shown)


source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
```diff
@@ -0,0 +1,61 @@
...
+    if (instruction->opcode() == SpvOpBranchConditional &&
+        instruction->NumOperands() == 5 &&
+        fuzzer_context->ChoosePercentage(
+            GetFuzzerContext()->GetChanceOfAdjustingBranchWeights())) {
+      std::pair<uint32_t, uint32_t> branch_weights = {0, 0};
+      while (branch_weights.first == 0 && branch_weights.second == 0) {
+        branch_weights.first = fuzzer_context->GetRandomUint32(INT32_MAX);
+        branch_weights.second =
+            fuzzer_context->GetRandomUint32(INT32_MAX - branch_weights.first);
+      }
```
afd, line 49: Looking at the SPIR-V spec: branch weights don't need to sum to anything particular - it's fine to choose any pair of uint32_t values, so long as they are not both zero and their sum doesn't overflow.

I suggest you add a method, GetRandomBranchWeights, to FuzzerContext, and have it return a std::pair<uint32_t>.

I like the way you use INT32_MAX, not UIN32_MAX, to avoid overflow problems - please use a code comment to explain that.

source/fuzz/transformation_adjust_branch_weights.cpp
```diff
@@ -0,0 +1,90 @@
...
+  SpvOp opcode = static_cast<SpvOp>(
+      message_.instruction_descriptor().target_instruction_opcode());
+
+  assert(instruction->opcode() == opcode &&
+         "The located instruction must have the same opcode as in the "
+         "descriptor.");
+
+  // Must be an OpBranchConditional instruction
+  // with all operands.
+  if (opcode != SpvOpBranchConditional || instruction->NumOperands() != 5) {
```
afd, line 52: As mentioned earlier, drop the requirement for all operands to be present (as we can add weights if they are missing).

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -742,6 +743,19 @@ message TransformationAddTypeVector {
...
+message TransformationAdjustBranchWeights {
+
+  // A transformation that adjusts the branch weights
+  // of a branch conditional instruction.
+
+  // A descriptor for a branch conditional instruction.
+  InstructionDescriptor instruction_descriptor = 1;
+
+  // Branch weights of a branch conditional instruction.
+  repeated uint32 branch_weights = 2;
```
afd, line 755: Use UInt32Pair instead of a repeated field here.

source/fuzz/transformation_adjust_branch_weights.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return false;
+  }
+
+  // There must be two branch weights.
+  if (message_.branch_weights_size() != 2) {
+    return false;
+  }
```
afd, line 59: This can go if you use UInt32Pair in the protobuf message.

source/fuzz/transformation_adjust_branch_weights.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  // At least one weight must be non-zero.
+  if (message_.branch_weights(0) == 0 && message_.branch_weights(1) == 0) {
+    return false;
+  }
```
afd, line 64: Let's change this to *assert* that at least one is non-zero, as we don't want a transformation where they are both zero to ever get created.

You can use death tests in your tests to check that the assert fails if such transformations are created.

source/fuzz/transformation_adjust_branch_weights.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  // The two weights must not overflow a 32-bit unsigned integer when added
+  // together.
+  if (message_.branch_weights(0) > UINT32_MAX - message_.branch_weights(1)) {
+    return false;
+  }
```
afd, line 70: Again, this could become an assertion.

source/fuzz/transformation_adjust_branch_weights.cpp
```diff
@@ -0,0 +1,90 @@
...
+  return true;
+}
+
+void TransformationAdjustBranchWeights::Apply(
+    opt::IRContext* ir_context, TransformationContext* /*unused*/) const {
+  auto instruction =
+      FindInstruction(message_.instruction_descriptor(), ir_context);
+  instruction->SetOperand(3, {message_.branch_weights(0)});
+  instruction->SetOperand(4, {message_.branch_weights(1)});
+}
```
afd, line 81: You could split this into two cases: the case where there are not existing branch weights (in which case you need to add two new input operands), and the case where there are existing branch weights (in which you need to change existing input operands as you are doing here).

Can you please introduce constants in an an anonymous namespace at the top of this CPP file for the magic numbers 3, 4 and 5, with descriptive names?

(I'm guilty of not always doing this in my code but should get more disciplined about it.)

source/fuzz/transformation_adjust_branch_weights.h
```diff
@@ -0,0 +1,55 @@
...
+  explicit TransformationAdjustBranchWeights(
+      const protobufs::TransformationAdjustBranchWeights& message);
+
+  TransformationAdjustBranchWeights(
+      const protobufs::InstructionDescriptor& instruction_descriptor,
+      const std::pair<uint32_t, uint32_t>& branch_weights);
+
+  // - |message_.instruction_descriptor| must identify an existing
+  //   branch conditional instruction
+  // - |branch_weights| will replace the current instruction branch weights
```
afd, line 37: This is not a condition that's checked inside IsApplicable.  The comments on IsApplicable should really match the various if statements in its implementation.

test/fuzz/transformation_adjust_branch_weights_test.cpp
```diff
@@ -0,0 +1,342 @@
...
+      TransformationAdjustBranchWeights(instruction_descriptor, {0, 1});
+  ASSERT_TRUE(
+      transformation.IsApplicable(context.get(), transformation_context));
+
+  instruction_descriptor =
+      MakeInstructionDescriptor(33, SpvOpBranchConditional, 0);
+  transformation =
+      TransformationAdjustBranchWeights(instruction_descriptor, {0, 0});
+  ASSERT_FALSE(
+      transformation.IsApplicable(context.get(), transformation_context));
```
afd, line 121: Before each ASSERT_FALSE, write a brief comment to explain why IsApplicable is false.

You may need to change some of these asserts to death tests if they will now trigger assertion failures in IsApplicable.

source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+void FuzzerPassAdjustBranchWeights::Apply() {
+  auto ir_context = GetIRContext();
+  auto fuzzer_context = GetFuzzerContext();
+  // For all OpBranchConditional instructions with branch weights,
+  // randomly applies the transformation.
+  ir_context->module()->ForEachInst([this, ir_context, fuzzer_context](
+                                        opt::Instruction* instruction) {
+    if (instruction->opcode() == SpvOpBranchConditional &&
+        instruction->NumOperands() == 5 &&
```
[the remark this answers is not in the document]
andreperezmaselco: I thought about it. But if they were not present, adding them will preserve the semantics?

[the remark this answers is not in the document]
afd: Yes, I believe so.  The spec says "branch weights are only hints".  So while they may change the performance of the code they should not change what it computes.

source/fuzz/fuzzer_context.h
```diff
@@ -189,8 +192,8 @@ class FuzzerContext {
   uint32_t GetMaximumEquivalenceClassSizeForDataSynonymFactClosure() {
     return max_equivalence_class_size_for_data_synonym_fact_closure_;
   }
-  uint32_t GetRandomIndexForAccessChain(uint32_t composite_size_bound) {
-    return random_generator_->RandomUint32(composite_size_bound);
+  uint32_t GetRandomUint32(uint32_t bound) {
+    return random_generator_->RandomUint32(bound);
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
```diff
@@ -0,0 +1,61 @@
...
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassAdjustBranchWeights::~FuzzerPassAdjustBranchWeights() = default;
+
+void FuzzerPassAdjustBranchWeights::Apply() {
+  auto ir_context = GetIRContext();
+  auto fuzzer_context = GetFuzzerContext();
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_adjust_branch_weights.cpp
andreperezmaselco: Done.

source/fuzz/transformation_adjust_branch_weights.cpp
andreperezmaselco: Done.

source/fuzz/transformation_adjust_branch_weights.cpp
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
andreperezmaselco: Done.

source/fuzz/transformation_adjust_branch_weights.cpp
andreperezmaselco: Done.

source/fuzz/transformation_adjust_branch_weights.h
andreperezmaselco: Done.

source/opt/instruction.h
```diff
@@ -806,6 +810,14 @@ inline void Instruction::ForEachInOperand(
   });
 }
 
+inline bool Instruction::HasBranchWeights() const {
```
afd, line 813: Don't make this 'inline' - make it a normal method and put the implementation in the .cpp file.  (We can promote it to inline in due course if profiling suggests that it is a hotspot.)

source/opt/instruction.h
```diff
@@ -806,6 +810,14 @@ inline void Instruction::ForEachInOperand(
...
+  if (opcode_ == SpvOpBranchConditional && NumOperands() == 5) {
```
afd, line 814: Once this is in the cpp file, declare a constant in the anonymous namespace at the top of the file to avoid using the magic number 5.  Give the constant a suitable name to indicate that it represents the number of operands of a branch with weights.

source/fuzz/transformation_adjust_branch_weights.h
```diff
@@ -34,7 +34,9 @@ class TransformationAdjustBranchWeights : public Transformation {
 
   // - |message_.instruction_descriptor| must identify an existing
   //   branch conditional instruction
-  // - |branch_weights| will replace the current instruction branch weights
+  // - |branch_weights| at least one weight must be non-zero and
+  //   the two weights must not overflow a 32-bit unsigned integer when added
+  //   together
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_adjust_branch_weights.cpp
```diff
@@ -31,25 +31,15 @@ FuzzerPassAdjustBranchWeights::FuzzerPassAdjustBranchWeights(
...
-            fuzzer_context->GetRandomUint32(INT32_MAX - branch_weights.first);
-      }
-
       auto instruction_descriptor =
-          MakeInstructionDescriptor(ir_context, instruction);
+          MakeInstructionDescriptor(GetIRContext(), instruction);
+      auto branch_weights = GetFuzzerContext()->GetRandomBranchWeights();
       auto transformation = TransformationAdjustBranchWeights(
           instruction_descriptor, branch_weights);
       ApplyTransformation(transformation);
```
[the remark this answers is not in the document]
andreperezmaselco: Good. Avoiding temporaries can save memory if the argument is not passed by reference.
Done.

source/opt/instruction.h
```diff
@@ -366,6 +366,10 @@ class Instruction : public utils::IntrusiveNodeBase<Instruction> {
   inline bool WhileEachInOperand(
       const std::function<bool(const uint32_t*)>& f) const;
 
+  // Returns true if it's an OpBranchCondition instruction
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/opt/instruction.h
andreperezmaselco: Done.

source/opt/instruction.h
andreperezmaselco: Done.

source/fuzz/protobufs/spvtoolsfuzz.proto
afd: Looks like you missed this one @andreperezmaselco

afd: This will require a couple of corresponding changes.

test/fuzz/transformation_adjust_branch_weights_test.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+  // Tests the two branch weights equal to 0.
+  instruction_descriptor =
+      MakeInstructionDescriptor(33, SpvOpBranchConditional, 0);
+  transformation =
+      TransformationAdjustBranchWeights(instruction_descriptor, {0, 0});
+#ifndef NDEBUG
+  ASSERT_DEATH(
+      transformation.IsApplicable(context.get(), transformation_context),
+      "At least one weight must be non-zero.");
```
paulthomson, line 124: I recommend removing the "." character or escaping it. Same thing below.

andreperezmaselco: Done.

PR #3337, opened 2020-05-12
Do merge return if the return is not at the end of the function.

s-perron:
We will remove the code in inlining to handle a return in the middle of
a function.  To inline those functions, we need to run merge return to
move the return to the end of the function.

### DIFF
--- a/source/opt/function.h
+++ b/source/opt/function.h
@@ -89,8 +89,8 @@ class Function {
   const std::unique_ptr<BasicBlock>& entry() const { return blocks_.front(); }
 
   // Returns the last basic block in this function.
-  BasicBlock* back() { return blocks_.back().get(); }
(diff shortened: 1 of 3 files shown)


source/opt/function.h
```diff
@@ -88,6 +88,10 @@ class Function {
   // Returns the entry basic block for this function.
   const std::unique_ptr<BasicBlock>& entry() const { return blocks_.front(); }
 
+  // Returns the last basic block in this function.
+  BasicBlock* back() { return blocks_.back().get(); }
+  const BasicBlock* back() const { return blocks_.back().get(); }
```
jaebaek, line 93: `BasicBlock` uses `tail()` instead of `back()`.
For the consistency of the name, why don't we use `tail()`?

s-perron: Done.  Good call.

PR #3339, opened 2020-05-13
spirv-fuzz: Fix to fact manager

### LINKED ISSUE #3338: spirv-fuzz: Fact manager can mutate the constant manager when querying constant facts
The fact manager's "GetConstantId" method calls the constant manager's "GetConstant" method, which adds a constant to the constant pool if not already present.  This can lead to a constant that doesn't have a corresponding id in the module being added to the constant manager.

While this hasn't yet led to a functional error, we should maintain the invariant (in spirv-fuzz at least) that the constant manager only knows about constants that exist in the module.

@Vasniktel @paulthomson FYI.

afd:
Reworks the way uniforms with known constants are queried to avoid
unintended side effects to the constant manager.

Fixes #3338.

### DIFF
--- a/source/fuzz/fact_manager.cpp
+++ b/source/fuzz/fact_manager.cpp
@@ -159,9 +159,26 @@ uint32_t FactManager::ConstantUniformFacts::GetConstantId(
     uint32_t type_id) const {
   auto type = context->get_type_mgr()->GetType(type_id);
   assert(type != nullptr && "Unknown type id.");
-  auto constant = context->get_constant_mgr()->GetConstant(
-      type, GetConstantWords(constant_uniform_fact));
-  return context->get_constant_mgr()->FindDeclaredConstant(constant, type_id);
+  const opt::analysis::Constant* known_constant;
+  if (type->AsInteger()) {
+    opt::analysis::IntConstant candidate_constant(
+        type->AsInteger(), GetConstantWords(constant_uniform_fact));
+    known_constant =
+        context->get_constant_mgr()->FindConstant(&candidate_constant);
+  } else {
+    assert(
+        type->AsFloat() &&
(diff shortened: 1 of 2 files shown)


source/fuzz/fact_manager.cpp
```diff
@@ -159,9 +159,26 @@ uint32_t FactManager::ConstantUniformFacts::GetConstantId(
...
+  const opt::analysis::Constant* known_constant;
+  if (type->AsInteger()) {
+    opt::analysis::IntConstant candidate_constant(
+        type->AsInteger(), GetConstantWords(constant_uniform_fact));
+    known_constant =
+        context->get_constant_mgr()->FindConstant(&candidate_constant);
+  } else {
+    assert(
+        type->AsFloat() &&
+        "Uniform constant facts are only supported for int and float types.");
```
Vasniktel, line 171: I'm curious whether uniform variables that have types different from `int` or `float` can have `FactConstantUniform` applied.

Vasniktel: Or why can't they in this case.

afd: spirv-fuzz accepts facts of the form: "uniform + indices" == "integer or float constant".

If you have a uniform u that is a struct consisting of a vec4 and an int2, you could say that the vec4 component is equal to (0.0, 0.5, 1.0, 1.5) by asserting four facts:

u[0][0] == 0.0
u[0][1] == 0.5
u[0][2] == 1.0
u[0][3] == 1.5

As almost everything in SPIR-V boils down to float or int (with the exception of bool) pretty much anything can be expressed this way (albeit in a somewhat cumbersome manner).

Vasniktel: Thanks, that explains it.

PR #3349, opened 2020-05-15
Preserve debug info in inline pass

jaebaek:


jaebaek:
@s-perron I closed https://github.com/KhronosGroup/SPIRV-Tools/pull/3250 and created a new PR. PTAL!

source/opt/inline_pass.cpp
```diff
@@ -270,30 +319,120 @@ std::unique_ptr<BasicBlock> InlinePass::AddGuardBlock(
...
+    uint32_t callee_inlined_at, DebugInlinedAtContext& inlined_at_ctx) {
+  if (inlined_at_ctx.FunctionCallScope().GetLexicalScope() == kNoDebugScope)
+    return kNoInlinedAt;
+
+  // Reuse the already generated DebugInlinedAt chain if exists.
+  uint32_t already_generated_chain_head_id =
+      inlined_at_ctx.InlinedAtChainHead(callee_inlined_at);
+  if (already_generated_chain_head_id != kNoInlinedAt) {
+    return already_generated_chain_head_id;
+  }
```
s-perron: The function and variable names are not very meaningful.  I have no idea what this is looking for or in what circumstances it will happen.  `InlinedAtChainHead` sounds like it should return a bool that says whether or not something has already been inlined at the chain head.

You don't have any comments at the declaration of the function.

source/opt/inline_pass.cpp
```diff
@@ -270,30 +319,120 @@ std::unique_ptr<BasicBlock> InlinePass::AddGuardBlock(
...
+
+  const uint32_t inlined_to_caller =
```
s-perron: This is another case where the variable name is not very useful.  Something like `new_dbg_inlined_at_id` might be better.  Again `inlined_to_caller` reads like it should be a boolean value.

source/opt/inline_pass.cpp
```diff
@@ -270,30 +319,120 @@ std::unique_ptr<BasicBlock> InlinePass::AddGuardBlock(
...
+      context()->get_debug_info_mgr()->CreateDebugInlinedAt(
+          inlined_at_ctx.FunctionCallLine(),
+          inlined_at_ctx.FunctionCallScope());
+  if (inlined_to_caller == kNoInlinedAt) return kNoInlinedAt;
+
+  // Simply specify that this instruction was inlined to the caller
+  // if it has no DebugInlinedAt in advance.
```
s-perron: This comment is hard to understand.  To write better comment be direct.  For example saying "Simply specify ..." adds nothing over "Specify ...".  Also the comments should add information that is not obvious from the code.  In this case, I don't think that this comment would be needed if the variable names were more descriptive.

source/opt/inline_pass.cpp
```diff
@@ -270,30 +319,120 @@ std::unique_ptr<BasicBlock> InlinePass::AddGuardBlock(
...
+  if (callee_inlined_at == kNoInlinedAt) {
+    // Keep the new chain information that will be reused it.
```
s-perron: This comment would not be needed if the `UpdateInlinedAtContext` was named better.

source/opt/inline_pass.cpp
```diff
@@ -270,30 +319,120 @@ std::unique_ptr<BasicBlock> InlinePass::AddGuardBlock(
...
+    inlined_at_ctx.UpdateInlinedAtContext(kNoInlinedAt, inlined_to_caller);
+    return inlined_to_caller;
+  }
+
+  // Create a new recursive DebugInlinedAt chain.
```
s-perron: does this mean the debug inlined at will refer to itself?

source/opt/inline_pass.cpp
```diff
@@ -29,8 +29,35 @@ static const int kSpvFunctionCallFunctionId = 2;
...
+  assert(dbg_inlined_at->GetOpenCL100DebugOpcode() ==
+         OpenCLDebugInfo100DebugInlinedAt);
+  return dbg_inlined_at->NumOperands() > kDebugInlinedAtOperandInlinedIndex;
+}
+
+void UpdateInlinedOperand(Instruction* dbg_inlined_at,
+                          uint32_t recursive_inlined_at_id) {
+  assert(HasInlinedOperand(dbg_inlined_at));
+  dbg_inlined_at->GetOperand(kDebugInlinedAtOperandInlinedIndex).words[0] =
+      recursive_inlined_at_id;
```
s-perron: There are other functions you can use to set the operand: `SetOperand` or `SetInOperand`.  Do not do it this way.  Also this is just a single line.  This should be inlined.  It will be easy enough to read.

Also it updates the "InlinedAt" operand not the "Inlined" operand.  The name should have reflected that.

source/opt/inline_pass.cpp
```diff
@@ -270,30 +319,120 @@ std::unique_ptr<BasicBlock> InlinePass::AddGuardBlock(
...
+    if (last_inlined_at_in_chain != nullptr) {
+      UpdateInlinedOperand(last_inlined_at_in_chain,
+                           new_inlined_at_in_chain->result_id());
+    }
+    last_inlined_at_in_chain = new_inlined_at_in_chain;
+
+    if (!HasInlinedOperand(new_inlined_at_in_chain)) {
+      break;
+    }
+    chain_iter_id = GetInlinedOperand(new_inlined_at_in_chain);
```
s-perron: Would be a better name `GetInlinedAtOperand`.  Also, why not include the check if the operand is present in this function?  It can return 0 if it does not exist.

source/opt/inline_pass.cpp
```diff
@@ -270,30 +319,120 @@ std::unique_ptr<BasicBlock> InlinePass::AddGuardBlock(
...
+  } while (chain_iter_id);
```
s-perron: I prefer comparing with 0 explicitly when the type of the variable is an int.  I find it easier to read.

source/opt/inline_pass.cpp
```diff
@@ -270,30 +319,120 @@ std::unique_ptr<BasicBlock> InlinePass::AddGuardBlock(
...
+
+  // Put |inlined_to_caller| into the end of the chain.
+  if (last_inlined_at_in_chain != nullptr) {
+    if (HasInlinedOperand(last_inlined_at_in_chain)) {
+      UpdateInlinedOperand(last_inlined_at_in_chain, inlined_to_caller);
```
s-perron: It might be a good idea to have this function add the InlinedAt operand if it does not exist instead of cluttering this part of the code.

source/opt/inline_pass.cpp
```diff
@@ -270,30 +319,120 @@ std::unique_ptr<BasicBlock> InlinePass::AddGuardBlock(
...
-  return callee_var_itr;
+  return callee_itr;
 }
 
-bool InlinePass::InlineInstructionInBB(
+bool InlinePass::InlineSingleInstruction(
     const std::unordered_map<uint32_t, uint32_t>& callee2caller,
-    BasicBlock* new_blk_ptr, const Instruction* inst) {
+    BasicBlock* new_blk_ptr, const Instruction* inst, uint32_t dbg_inlined_at,
+    bool use_new_id) {
```
s-perron, line 334: I don't get why you need `use_new_id`.  Is there a reason the result ids of the debug instruction should not be mapped?  Is there a test that points out the reason for this?

source/opt/inline_pass.h
```diff
@@ -169,6 +166,52 @@ class InlinePass : public Pass {
   std::unordered_set<uint32_t> funcs_called_from_continue_;
 
  private:
+  // DebugInlinedAtContext keeps data used for generating DebugInlinedAt
+  // instructions.
```
s-perron: Try to make this more meaningful.  What data is kept?  Also add comments describing the functions that are not getters and setters.

source/opt/inline_pass.h
```diff
@@ -169,6 +166,52 @@ class InlinePass : public Pass {
...
+    const DebugScope& FunctionCallScope() { return call_instr_scope; }
+    uint32_t InlinedAtChainHead(uint32_t callee_instr_inlined_at) {
+      auto chain_itr = callee_inlined_at2chain.find(callee_instr_inlined_at);
+      if (chain_itr != callee_inlined_at2chain.end()) return chain_itr->second;
+      return kNoInlinedAt;
+    }
+
+   private:
+    const Instruction*
+        call_instr_line;  // Line instruction of function call instruction
```
s-perron: Move the comment to the line before the declaration.  It will look cleaner.  The same for the next line.

source/opt/inline_pass.h
```diff
@@ -169,6 +166,52 @@ class InlinePass : public Pass {
...
+    const DebugScope&
+        call_instr_scope;  // Debug scope of function call instruction
+
+    // Map from DebugInlinedAt ids of callee to head ids of new generated
+    // DebugInlinedAt chain.
+    std::unordered_map<uint32_t, uint32_t> callee_inlined_at2chain;
```
s-perron: class member name should end with an underscore.

source/opt/inline_pass.h
```diff
@@ -169,6 +166,52 @@ class InlinePass : public Pass {
...
+  };
+
+  // Clone and map callee locals.  Return true if successful.
+  bool CloneAndMapLocals(Function* calleeFn,
+                         std::vector<std::unique_ptr<Instruction>>* new_vars,
+                         std::unordered_map<uint32_t, uint32_t>* callee2caller,
+                         DebugInlinedAtContext& inlined_at_ctx);
+
+  // Builds DebugScope for the callee instruction.
```
s-perron: You should be describing what is returned and any side effect.  You just say it builds something.  That tells me nothing about the return value or were the built debug scope will be.  You could try something like:

> Returns the debug scope corresponding to inlining |inlined_at_ctx| into the scope |callee_instr_scope|.  Generates all new debug instruction needed to represent the scope.

source/opt/inline_pass.h
```diff
@@ -169,6 +166,52 @@ class InlinePass : public Pass {
...
+  DebugScope BuildDebugScope(const DebugScope& callee_instr_scope,
+                             DebugInlinedAtContext& inlined_at_ctx);
+
+  // Builds DebugInlinedAt chain for the callee instruction. Returns the
```
s-perron: This is not clear on what is actually built.  You reference a "callee instruction", but this function does not take a callee instruction.  The function should not be describing the context in which you use it.  It should describe what this function does.

source/opt/inline_pass.cpp
jaebaek: Ah .. I was confused. I reverted it.

PR #3356, opened 2020-05-19
[spirv-opt] debug info preservation in ssa-rewrite

jaebaek:
OpenCL.DebugInfo.100 spec mimics
[llvm.dbg.declare](https://llvm.org/docs/SourceLevelDebugging.html#llvm-dbg-declare) and [llvm.dbg.value](https://llvm.org/docs/SourceLevelDebugging.html#llvm-dbg-value) of LLVM IR
using [DebugDeclare](https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html#DebugDeclare) and [DebugValue](https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html#DebugValue).

For example, we can describe an `OpVariable` for a local
variable using `DebugDeclare` and when it is updated by
`OpStore` we can specify the new value for the local variable
using `DebugValue`.

As what LLVM ssa-rewrite does, we want to add
`DebugValue` for variable updates and phi instructions that
will be used by a shader debugger. This commit lets
ssa-rewrite pass add `DebugValue` instructions for store
and a phi instructions to specify a result id of a value is
corresponding to the value of a local variable. Note that
ssa-rewrite will remove `DebugDeclare` after adding all
`DebugValue` instructions properly.

### DIFF
--- a/source/opt/debug_info_manager.cpp
+++ b/source/opt/debug_info_manager.cpp
@@ -156,7 +156,8 @@ uint32_t DebugInfoManager::CreateDebugInlinedAt(const Instruction* line,
         {spv_operand_type_t::SPV_OPERAND_TYPE_ID, {scope.GetInlinedAt()}});
   }
   RegisterDbgInst(inlined_at.get());
-  context()->get_def_use_mgr()->AnalyzeInstDefUse(inlined_at.get());
+  if (context()->AreAnalysesValid(IRContext::Analysis::kAnalysisDefUse))
+    context()->get_def_use_mgr()->AnalyzeInstDefUse(inlined_at.get());
   context()->module()->AddExtInstDebugInfo(std::move(inlined_at));
   return result_id;
 }
@@ -246,7 +247,8 @@ Instruction* DebugInfoManager::GetDebugInfoNone() {
           std::move(dbg_info_none_inst));
 
   RegisterDbgInst(debug_info_none_inst_);
-  context()->get_def_use_mgr()->AnalyzeInstDefUse(debug_info_none_inst_);
+  if (context()->AreAnalysesValid(IRContext::Analysis::kAnalysisDefUse))
+    context()->get_def_use_mgr()->AnalyzeInstDefUse(debug_info_none_inst_);
   return debug_info_none_inst_;
 }
 
@@ -272,7 +274,8 @@ Instruction* DebugInfoManager::GetEmptyDebugExpression() {
           std::move(empty_debug_expr));
 
   RegisterDbgInst(empty_debug_expr_inst_);
-  context()->get_def_use_mgr()->AnalyzeInstDefUse(empty_debug_expr_inst_);
+  if (context()->AreAnalysesValid(IRContext::Analysis::kAnalysisDefUse))
+    context()->get_def_use_mgr()->AnalyzeInstDefUse(empty_debug_expr_inst_);
   return empty_debug_expr_inst_;
 }
 
@@ -293,7 +296,8 @@ Instruction* DebugInfoManager::CloneDebugInlinedAt(uint32_t clone_inlined_at_id,
   std::unique_ptr<Instruction> new_inlined_at(inlined_at->Clone(context()));
   new_inlined_at->SetResultId(context()->TakeNextId());
   RegisterDbgInst(new_inlined_at.get());
-  context()->get_def_use_mgr()->AnalyzeInstDefUse(new_inlined_at.get());
+  if (context()->AreAnalysesValid(IRContext::Analysis::kAnalysisDefUse))
+    context()->get_def_use_mgr()->AnalyzeInstDefUse(new_inlined_at.get());
   if (insert_before != nullptr)
     return insert_before->InsertBefore(std::move(new_inlined_at));
   return context()->module()->ext_inst_debuginfo_end()->InsertBefore(
@@ -347,7 +351,8 @@ Instruction* DebugInfoManager::AddDebugValue(Instruction* instr,
   Instruction* added_dbg_value =
       insert_before->InsertBefore(std::move(new_dbg_value));
   AnalyzeDebugInst(added_dbg_value);
-  context()->get_def_use_mgr()->AnalyzeInstDefUse(added_dbg_value);
+  if (context()->AreAnalysesValid(IRContext::Analysis::kAnalysisDefUse))
+    context()->get_def_use_mgr()->AnalyzeInstDefUse(added_dbg_value);
   return added_dbg_value;
 }
 
@@ -370,6 +375,12 @@ uint32_t DebugInfoManager::GetVariableIdOfDebugValueUsedForDeclare(
 
   uint32_t var_id =
       inst->GetSingleWordOperand(kDebugDeclareOperandVariableIndex);
+  if (!context()->AreAnalysesValid(IRContext::Analysis::kAnalysisDefUse)) {
+    assert(false &&
+           "Checking a DebugValue can be used for declare needs DefUseManager");
+    return 0;
+  }
+
   auto* var = context()->get_def_use_mgr()->GetDef(var_id);
   if (var->opcode() == SpvOpVariable &&
       SpvStorageClass(var->GetSingleWordOperand(
--- a/source/opt/debug_info_manager.cpp
+++ b/source/opt/debug_info_manager.cpp
@@ -68,15 +68,6 @@ Instruction* DebugInfoManager::GetDbgInst(uint32_t id) {
   return dbg_inst_it == id_to_dbg_inst_.end() ? nullptr : dbg_inst_it->second;
 }
 
-Instruction* DebugInfoManager::GetDebugDeclareOrValueForLoadOrStore(
-    Instruction* load_or_store) {
-  auto dbg_decl_itr = var_load_store_to_dbg_decl_.find(load_or_store);
-  if (dbg_decl_itr != var_load_store_to_dbg_decl_.end()) {
-    return dbg_decl_itr->second;
-  }
-  return nullptr;
-}
-
 void DebugInfoManager::RegisterDbgInst(Instruction* inst) {
   assert(
       inst->NumInOperands() != 0 &&
@@ -296,10 +287,11 @@ Instruction* DebugInfoManager::CloneDebugInlinedAt(uint32_t clone_inlined_at_id,
       std::move(new_inlined_at));
 }
 
-void DebugInfoManager::AddDebugValue(Instruction* load_or_store,
-                                     uint32_t variable_id, uint32_t value_id) {
+Instruction* DebugInfoManager::AddDebugValue(Instruction* instr,
+                                             uint32_t variable_id,
+                                             uint32_t value_id) {
   auto dbg_decl_itr = var_id_to_dbg_decl_.find(variable_id);
-  if (dbg_decl_itr == var_id_to_dbg_decl_.end()) return;
+  if (dbg_decl_itr == var_id_to_dbg_decl_.end()) return nullptr;
 
   uint32_t result_id = context()->TakeNextId();
   std::unique_ptr<Instruction> new_dbg_value(new Instruction(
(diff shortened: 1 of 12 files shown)


jaebaek:
As far as I understand, the only thing we must do is updating operands of `DebugDeclare` and `DebugValue` according to the load/store/variable updates in the ssa-rewrite.

Actually, since `bool IRContext::ReplaceAllUsesWith(uint32_t before, uint32_t after)` does the replacement correctly, the only thing we have to do is to include result ids in the targets if they are used by `DebugDeclare` and `DebugValue` in addition to `OpLoad`, `OpStore`, and `OpVariable`.

I will add more test cases later but I first want to doublecheck the idea before taking long time for writing test cases.

s-perron:
This does not seem right.  The ssa rewriter needs to create new DebugValue instructions every time it removes a store.  We can talk offline, but the simple example we should walk through is

```
#version 140
  
in vec4 BC;
out float fo;
  
void main()
{
  float f = 0.0;
  if (BD.x > 0.0) {
    f = 1.0;
  }
  fo = f;
}
```

jaebaek:
I updated the code based on the offline discussion. PTAL!
I did not add many tests yet. Once we both agree the direction, I will add more tests.

s-perron:
Two general comments.  First, remember that the analysis should represent information that can be rebuild from the spir-v, and nothing more.  It should not contain any state specific to a particular pass.  Something like [this](https://github.com/KhronosGroup/SPIRV-Tools/pull/3356/files#diff-2b2907e1dc90aeb85b28cc414990a4daR133) is not right because linking a debug value instruction to a particular load or store is not possible when looking at the spir-v after the pass.

Second, it does not look like you are adding a debug value after an OpPhi is generated.  The DebugValue instructions should be use to say which expression can be used to get the value for a given variable.  That expression must be valid if the debugger is going to work.  Loads are unimportant because they do not affect the value of the variable.  Stores are important because they change the value of a variable.  Merge blocks are important because the old value may not be accurate anymore. 
 If a OpPhi is added for that variable, then you need a new DebugValue.  If you want to see how this is suppose to work look at clang.

If the cpp file is

```
int foo(bool b)                                   
{ 
  int i = 0;
  if (b) {
    i = foo(false); 
  }
  if (!b) {
    i = 1;
  }                                                                                                                                 
  return i; 
} 
```

Compile using `clang++ t.cpp -g -O  -c -mllvm -print-after-all`.  Notice that when the debug value instructions are added, there is a debug value after every phi instruction.

jaebaek:
Now it adds `DebugValue` only for applicable `OpStore` and `OpPhi`. PTAL.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -2024,6 +2024,788 @@ OpFunctionEnd
...
+; CHECK: [[i_name:%\w+]] = OpString "i"
+; CHECK: [[dbg_f:%\w+]] = OpExtInst %void [[ext:%\d+]] DebugLocalVariable [[f_name]]
+; CHECK: [[dbg_i:%\w+]] = OpExtInst %void [[ext]] DebugLocalVariable [[i_name]]
+
+; CHECK:      OpStore %f %float_0
+; CHECK:      OpStore %i %int_0
+; CHECK:      OpExtInst %void [[ext]] DebugDeclare [[dbg_f]] %f
+; CHECK:      OpExtInst %void [[ext]] DebugDeclare [[dbg_i]] %i
+; CHECK:      [[phi0:%\w+]] = OpPhi %float %float_0
+; CHECK-NEXT: OpExtInst %void [[ext]] DebugValue [[dbg_f]] [[phi0]]
```
s-perron, line 2053: Can you have debug value instruction mixed in with the OpPhi instructions?

test/opt/local_ssa_elim_test.cpp
```diff
@@ -2024,6 +2024,788 @@ OpFunctionEnd
...
+; CHECK:      [[phi1:%\w+]] = OpPhi %int %int_0
+; CHECK-NEXT: OpExtInst %void [[ext]] DebugValue [[dbg_i]] [[phi1]]
+; CHECK:      OpStore %f [[f_val:%\w+]]
+; CHECK-NEXT: OpExtInst %void [[ext]] DebugValue [[dbg_f]] [[f_val]]
+; CHECK:      OpStore %i [[i_val:%\w+]]
+; CHECK-NEXT: OpExtInst %void [[ext]] DebugValue [[dbg_i]] [[i_val]]
```
s-perron, line 2059: It might be good to provide some context by checking that these are all in the loop.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -2024,6 +2024,788 @@ OpFunctionEnd
...
+  const std::string text = R"(
+; CHECK: [[f_name:%\w+]] = OpString "f"
+; CHECK: [[i_name:%\w+]] = OpString "i"
+; CHECK: [[dbg_f:%\w+]] = OpExtInst %void [[ext:%\d+]] DebugLocalVariable [[f_name]]
+; CHECK: [[dbg_i:%\w+]] = OpExtInst %void [[ext]] DebugLocalVariable [[i_name]]
+
+; CHECK:      OpStore %f %float_0
+; CHECK:      OpStore %i %int_0
+; CHECK:      OpExtInst %void [[ext]] DebugDeclare [[dbg_f]] %f
+; CHECK:      OpExtInst %void [[ext]] DebugDeclare [[dbg_i]] %i
```
s-perron, line 2051: Is there a DebugValue after the initialization of `f` and `i`?  If not, there should be.  Otherwise, we will have to keep the variable around otherwise the debugger won't know what the value is until it reaches the first debug value.

I'm wondering if we should be keeping the debug delcare around if the ssa rewriter removes the use of the variable.  Do you know what happens in LLVM?

test/opt/local_ssa_elim_test.cpp
```diff
@@ -2024,6 +2024,788 @@ OpFunctionEnd
...
+%37 = OpLoad %v4float %v
+OpStore %gl_FragColor %37
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndMatch<SSARewritePass>(text, true);
+}
+
+TEST_F(LocalSSAElimTest, DebugLostCopyProblem) {
```
s-perron, line 2580: Can you add a comment explaining what the lost copy problem is?  It does not mean much to me.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -2024,6 +2024,788 @@ OpFunctionEnd
...
+%44 = OpLoad %float %t
+OpStore %fo %44
+OpReturn
+OpFunctionEnd
+)";
+
+  SinglePassRunAndMatch<SSARewritePass>(text, true);
+}
+
+TEST_F(LocalSSAElimTest, DebugSwapProblem) {
```
s-perron, line 2691: Same as before.  I don't know what that problem is.

test/opt/local_ssa_elim_test.cpp
jaebaek: done

test/opt/local_ssa_elim_test.cpp
jaebaek: I added this because the `LostCopyProblem` test case exists. However, it does not mean much to me as well. Let me remove it.

test/opt/local_ssa_elim_test.cpp
jaebaek: I added a comment. When I checked this case, I saw the following two phi instructions
```
%65 = OpPhi %float %float_1 %39 %64 %47
%64 = OpPhi %float %float_0 %39 %65 %47
```
They are operands of each other i.e., `%64` and `%65`. I want to check we correctly add `DebugValue` for those phi instructions of `%f1` and `%f2`.

test/opt/local_ssa_elim_test.cpp
s-perron: The spec says:

> Within a block, this instruction must appear before all non-OpPhi instructions (except for OpLine and OpNoLine, which can be mixed with OpPhi).

I do not see anything in the debug instruction spec that changes this.  Personally, it does not make sense that a debug value or debug declare should be mixed with the OpPhi instructions.  Technically the spec says they are not allowed (regardless of what spirv-val does).

s-perron: We should double check this.

test/opt/local_ssa_elim_test.cpp
s-perron: LLVM gets rid of the alloca as well, which is a different design decision that what we did.  Diego decided to keep the stores, so the OpVariable instruction must stick around.

I'm leaning towards removing the debug declare in ssa-rewrite, but we should have a quick discussion about it tomorrow with Diego.

test/opt/local_ssa_elim_test.cpp
jaebaek: Oh I see. Now I checked it in the description of `OpPhi`. I fixed it.

dnovillo:
> _No description provided._

Please provide a description for this change.

source/opt/debug_info_manager.h
```diff
@@ -128,6 +129,14 @@ class DebugInfoManager {
   uint32_t BuildDebugInlinedAtChain(uint32_t callee_inlined_at,
                                     DebugInlinedAtContext* inlined_at_ctx);
 
+  // Creates a DebugValue whose 'Local Variable' and 'Value' operands are
+  // |variable_id| and |value_id| and inserts it after |instr|.
```
s-perron, line 133: This needs to be updated.  It might create more than one, and is not clear on what the function is doing.

Try something like:

Generates a DebugValue instruction with value |value_id| for every local variable that is in scope and whose memory is |variable_id|.

source/opt/debug_info_manager.cpp
```diff
@@ -252,12 +314,157 @@ Instruction* DebugInfoManager::CloneDebugInlinedAt(uint32_t clone_inlined_at_id,
...
+  }
+  return parent_scope;
+}
+
+bool DebugInfoManager::IsDeclareVisibleToInstr(Instruction* dbg_declare,
+                                               uint32_t instr_scope_id) {
+  if (instr_scope_id == kNoDebugScope) return false;
+
+  uint32_t dbg_local_var_id =
+      dbg_declare->GetSingleWordOperand(kDebugValueOperandLocalVariableIndex);
```
s-perron, line 362: Is this a debug declare or debugValue?  You seem to mix the two.

jaebaek: Changed it to `kDebugDeclareOperandLocalVariableIndex`.

source/opt/debug_info_manager.cpp
```diff
@@ -252,12 +314,157 @@ Instruction* DebugInfoManager::CloneDebugInlinedAt(uint32_t clone_inlined_at_id,
...
+}
+
+void DebugInfoManager::AddDebugValue(Instruction* instr, uint32_t variable_id,
+                                     uint32_t value_id) {
+  auto dbg_decl_itr = var_id_to_dbg_decl_.find(variable_id);
+  if (dbg_decl_itr == var_id_to_dbg_decl_.end()) return;
+
+  uint32_t instr_scope_id = instr->GetDebugScope().GetLexicalScope();
+  for (auto* dbg_decl_or_val : dbg_decl_itr->second) {
+    if (!IsDeclareVisibleToInstr(dbg_decl_or_val, instr_scope_id)) continue;
```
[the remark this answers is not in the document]
jaebaek: It's a little confusing. I added one more check `IsVariableVisibleToInstr()` right after `IsDeclareVisibleToInstr()`, which returns true if the variable is visible in the scope of the instruction. Please let me know if I understand your intention incorrectly.

source/opt/debug_info_manager.h
jaebaek: done

source/opt/debug_info_manager.cpp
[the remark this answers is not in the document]
s-perron: That is not what I meant.  I meant that you use |instr| for two purposes: 

1. Identify the position to place the new instructions: The first legal position after |intsr|.
1. As the debug scope of the new instruction.

I don't think that they would have to be the same.  How hard it is to split these two things into two separate parameters: `Instruction* insert_pos, uint32_t debug_scope_id`.

Remove  `IsVariableVisibleToInstr`.  That is not useful.  The debug scope of the instruction allocating the memory (`OpVariable`) is irrelevant.  You had it right the first time.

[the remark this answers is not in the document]
jaebaek: > Remove IsVariableVisibleToInstr. That is not useful. The debug scope of the instruction allocating the memory (OpVariable) is irrelevant. You had it right the first time.

I agree.
It makes sense. I updated it.

PR #3359, opened 2020-05-19
spirv-fuzz: Add push id through variable transformation

### LINKED ISSUE #3192: spirv-fuzz transformation: create id synonym by storing through a variable
Create a new transformation, TransformationPushIdThroughVariable, that:

- takes an id x, an instruction descriptor inst and two fresh ids, v and t
- adds either a global variable to the module, or a local variable to the function containing inst, with result id v and with a pointer to whatever type id has
- inserts, before inst:
    - a store of x to v
    - a load from v to t
- adds the fact that ids x and t are synonymous

Create a fuzzer pass, FuzzerPassPushIdsThroughVariables, that considers at each point in the module where such a transformation could be applied, randomly decides whether to apply one, and for those places where the transformation should be applied randomly chooses an id that is in scope and creates and applies the transformation.

andreperezmaselco:
Fixes #3192.

### DIFF
--- a/source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
+++ b/source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
@@ -92,6 +92,9 @@ void FuzzerPassPushIdsThroughVariables::Apply() {
           return;
         }
 
+        // If the pointer type does not exist, then create it.
+        FindOrCreatePointerType(basic_type_id, variable_storage_class);
+
         // Applies the push id through variable transformation.
         ApplyTransformation(TransformationPushIdThroughVariable(
             value_instructions[GetFuzzerContext()->RandomIndex(
--- a/source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
+++ b/source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
@@ -38,25 +38,30 @@ void FuzzerPassPushIdsThroughVariables::Apply() {
   ForEachInstructionWithInstructionDescriptor(
       [this](opt::Function* function,
              opt::BasicBlock* block,
-             opt::BasicBlock::iterator inst_it,
+             opt::BasicBlock::iterator instruction_iterator,
              const protobufs::InstructionDescriptor& instruction_descriptor) -> void {
-        assert(inst_it->opcode() == instruction_descriptor.target_instruction_opcode() &&
+        assert(instruction_iterator->opcode() == instruction_descriptor.target_instruction_opcode() &&
                "The opcode of the instruction we might insert before must be "
                "the same as the opcode in the descriptor for the instruction");
-
-        // Check whether the store and load instructions can be
-        // inserted before this instruction.
-        if (!fuzzerutil::CanInsertOpcodeBeforeInstruction(SpvOpStore, inst_it) ||
-            !fuzzerutil::CanInsertOpcodeBeforeInstruction(SpvOpLoad, inst_it)) {
-          return;
-        }
 
         // Randomly decide whether to try pushing an id through a variable.
         if (!GetFuzzerContext()->ChoosePercentage(GetFuzzerContext()->GetChanceOfPushingIdThroughVariable())) {
           return;
         }
 
-        // Storage class for a global or local variable.
+        // The instruction to insert before containing block must be reachable.
+        if (!fuzzerutil::BlockIsReachableInItsFunction(GetIRContext(), block)) {
+          return;
+        }
+
+        // It must be valid to insert OpStore and OpLoad instructions
+        // before the instruction to insert before.
+        if (!fuzzerutil::CanInsertOpcodeBeforeInstruction(SpvOpStore, instruction_iterator) ||
+            !fuzzerutil::CanInsertOpcodeBeforeInstruction(SpvOpLoad, instruction_iterator)) {
+          return;
+        }
+
+        // Randomly decides whether a global or local variable will be added.
         SpvStorageClass variable_storage_class;
         if (GetFuzzerContext()->ChooseEven()) {
           variable_storage_class = SpvStorageClassPrivate;
@@ -64,15 +69,14 @@ void FuzzerPassPushIdsThroughVariables::Apply() {
           variable_storage_class = SpvStorageClassFunction;
         }
 
-        // Gets the basic type and pointers to basic types.
+        // Gets the available basic and pointer types.
         auto basic_type_ids_and_pointers = GetAvailableBasicTypesAndPointers(variable_storage_class);
         auto& basic_types = basic_type_ids_and_pointers.first;
         uint32_t basic_type_id = basic_types[GetFuzzerContext()->RandomIndex(basic_types)];
 
-        // Look for values we might consider to store.
+        // Looks for values we might consider to store.
         std::vector<opt::Instruction*> value_instructions =
-            FindAvailableInstructions(
-                function, block, inst_it,
+            FindAvailableInstructions(function, block, instruction_iterator,
                 [basic_type_id](opt::IRContext* /*unused*/,
                           opt::Instruction* instruction) -> bool {
                   if (!instruction->result_id() || !instruction->type_id()) {
@@ -85,20 +89,21 @@ void FuzzerPassPushIdsThroughVariables::Apply() {
           return;
         }
 
+        // Gets type pointer ids to the basic type.
         auto& basic_type_to_pointers = basic_type_ids_and_pointers.second;
         std::vector<uint32_t>& type_pointer_ids = basic_type_to_pointers.at(basic_type_id);
         uint32_t type_pointer_id;
 
+        // Gets the type pointer id.
         if (type_pointer_ids.empty()) {
           type_pointer_id = GetFuzzerContext()->GetFreshId();
-          type_pointer_ids.push_back(type_pointer_id);
           ApplyTransformation(TransformationAddTypePointer(type_pointer_id, variable_storage_class, basic_type_id));
+          type_pointer_ids.push_back(type_pointer_id);
         } else {
           type_pointer_id = type_pointer_ids[GetFuzzerContext()->RandomIndex(type_pointer_ids)];
         }
 
-        // Create whether a global variable
-        // or a local variable.
+        // Adds whether a global or local variable.
         uint32_t variable_id = GetFuzzerContext()->GetFreshId();
         if (variable_storage_class == SpvStorageClassPrivate) {
           ApplyTransformation(TransformationAddGlobalVariable(
@@ -110,10 +115,11 @@ void FuzzerPassPushIdsThroughVariables::Apply() {
               FindOrCreateZeroConstant(basic_type_id), false));
         }
 
+        // Applies the push id through variable transformation.
         ApplyTransformation(TransformationPushIdThroughVariable(
             GetFuzzerContext()->GetFreshId(),
-            variable_id,
             value_instructions[GetFuzzerContext()->RandomIndex(value_instructions)]->result_id(),
+            variable_id,
             instruction_descriptor));
       });
 }
--- a/source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
+++ b/source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
@@ -48,7 +48,8 @@ void FuzzerPassPushIdsThroughVariables::Apply() {
           return;
         }
 
-        // The instruction to insert before containing block must be reachable.
+        // The block containing the instruction we are going to insert before
+        // must be reachable.
         if (!fuzzerutil::BlockIsReachableInItsFunction(GetIRContext(), block)) {
           return;
         }
@@ -63,12 +64,9 @@ void FuzzerPassPushIdsThroughVariables::Apply() {
         }
 
         // Randomly decides whether a global or local variable will be added.
-        SpvStorageClass variable_storage_class;
-        if (GetFuzzerContext()->ChooseEven()) {
-          variable_storage_class = SpvStorageClassPrivate;
-        } else {
-          variable_storage_class = SpvStorageClassFunction;
-        }
+        auto variable_storage_class = GetFuzzerContext()->ChooseEven()
+                                          ? SpvStorageClassPrivate
+                                          : SpvStorageClassFunction;
 
         // Gets the available basic and pointer types.
         auto basic_type_ids_and_pointers =
@@ -77,7 +75,8 @@ void FuzzerPassPushIdsThroughVariables::Apply() {
         uint32_t basic_type_id =
             basic_types[GetFuzzerContext()->RandomIndex(basic_types)];
 
-        // Looks for values we might consider to store.
+        // Looks for ids that we might wish to consider pushing through a
+        // variable.
         std::vector<opt::Instruction*> value_instructions =
             FindAvailableInstructions(
                 function, block, instruction_iterator,
@@ -99,7 +98,6 @@ void FuzzerPassPushIdsThroughVariables::Apply() {
                                    value_instructions)]
                 ->result_id(),
             GetFuzzerContext()->GetFreshId(), GetFuzzerContext()->GetFreshId(),
-            FindOrCreatePointerType(basic_type_id, variable_storage_class),
             variable_storage_class, instruction_descriptor));
       });
 }
--- a/source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
+++ b/source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
@@ -1,0 +1,122 @@
+// Copyright (c) 2020 André Perez Maselco
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  // The instruction to insert before must be defined.
+  auto instruction_to_insert_before =
+      FindInstruction(message_.instruction_descriptor(), ir_context);
+  if (!instruction_to_insert_before) {
+    return false;
+  }
+
+  // The instruction to insert before must belongs to a reachable block.
```
afd, line 74: belongs -> belong

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,137 @@
...
+void TransformationPushIdThroughVariable::Apply(
+    opt::IRContext* ir_context,
+    TransformationContext* transformation_context) const {
+  TransformationStore(message_.variable_id(), message_.value_id(),
+                      message_.instruction_descriptor())
+      .Apply(ir_context, transformation_context);
+
+  TransformationLoad(message_.value_synonym_id(), message_.variable_id(),
+                     message_.instruction_descriptor())
+      .Apply(ir_context, transformation_context);
```
afd, line 122: @paulthomson What do you think of this?  We haven't done this before - used Apply methods from other transformations in a larger transformation.

test/fuzz/transformation_push_id_through_variable_test.cpp
```diff
@@ -0,0 +1,389 @@
...
+)";
+
+const auto env = SPV_ENV_UNIVERSAL_1_4;
+const auto consumer = nullptr;
+const auto context =
+    BuildModule(env, consumer, reference_shader, kFuzzAssembleOption);
+
+FactManager fact_manager;
+spvtools::ValidatorOptions validator_options;
+TransformationContext transformation_context(&fact_manager, validator_options);
```
afd, line 97: I am unsure about making these global.  I'm not sure how gtest works but I would not be surprised if the state of globals persists between individual tests.  We do not want to require the tests to run in a particular order.

I suggest you follow the style used by other tests, where each test creates its own IR context locally.  Alternatively you could look at other facilities that gtest might offer, such as test fixtures - they could be useful if you want to reduce redundancy between tests.

That said, I quite like each test to be pretty simple and self-contained, even if it ends up being a bit long.

@paulthomson your feedback here would be great.

source/fuzz/fuzzer_pass_add_loads.cpp
```diff
@@ -60,7 +60,7 @@ void FuzzerPassAddLoads::Apply() {
                   if (!instruction->result_id() || !instruction->type_id()) {
                     return false;
                   }
-                  switch (instruction->result_id()) {
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_add_stores.cpp
```diff
@@ -72,7 +72,7 @@ void FuzzerPassAddStores::Apply() {
                     // Read only: cannot store to it.
                     return false;
                   }
-                  switch (instruction->result_id()) {
+                  switch (instruction->opcode()) {
                     case SpvOpConstantNull:
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
andreperezmaselco: Done.

test/fuzz/transformation_push_id_through_variable_test.cpp
paulthomson: I agree with Ally here.

source/fuzz/transformation_push_id_through_variable.cpp
paulthomson: This seems OK. Perhaps it is slightly confusing, but a comment will make it clear that the transformations are not being added to the list of transformations. Apply might update the fact manager, but that is OK too I think.

paulthomson: But actually, in this particular case, I think it is clearer to just add the code that adds the load and store instructions because there is not much code.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -983,6 +984,27 @@ message TransformationPermuteFunctionParameters {
...
+  // synonymous by storing |value_id| into |variable_id| and
+  // loading |variable_id| to |value_synonym_id|.
+
+  // A fresh id for the result of the load instruction.
+  uint32 value_synonym_id = 1;
+
+  // The value to be stored.
+  uint32 value_id = 2;
+
+  // The variable to be stored to.
```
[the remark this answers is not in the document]

source/fuzz/transformation_push_id_through_variable.cpp
andreperezmaselco: Done.

test/fuzz/transformation_push_id_through_variable_test.cpp
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  // |message_.value_id| must be available at the insertion point.
+  return fuzzerutil::IdIsAvailableBeforeInstruction(
+      ir_context, instruction_to_insert_before, message_.value_id());
+}
+
+void TransformationPushIdThroughVariable::Apply(
+    opt::IRContext* ir_context,
+    TransformationContext* transformation_context) const {
+  // Adds whether a global or local variable.
```
afd, line 116: Instead of using pointer_type_id, you'll need to use the helper function I've suggested above to get that id.

source/fuzz/transformation_push_id_through_variable.h
```diff
@@ -0,0 +1,68 @@
...
+  //   which it is valid to insert the OpStore and OpLoad instructions before it
+  //   and must be belongs to a reachable block.
+  bool IsApplicable(
+      opt::IRContext* ir_context,
+      const TransformationContext& transformation_context) const override;
+
+  // Stores |value_id| to |variable_id|,
+  // loads |variable_id| to |value_synonym_id| and
+  // adds the fact that |value_synonym_id|
+  // and |value_id| are synonymous.
```
afd, line 55: Nit: I would use fewer lines and more of your 80-character width.  (Not important though.)

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,174 @@
...
+  if (pointer_type_instruction->GetSingleWordInOperand(1) !=
+      value_instruction->type_id()) {
+    return false;
+  }
+
+  // |message_.variable_storage_class| must be private or function.
+  if (message_.variable_storage_class() != SpvStorageClassPrivate &&
+      message_.variable_storage_class() != SpvStorageClassFunction) {
+    return false;
+  }
```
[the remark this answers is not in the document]
afd: You may need to change some of your test assertions to death tests, of course.

source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
```diff
@@ -0,0 +1,108 @@
...
+          return;
+        }
+
+        // Randomly decides whether a global or local variable will be added.
+        SpvStorageClass variable_storage_class;
+        if (GetFuzzerContext()->ChooseEven()) {
+          variable_storage_class = SpvStorageClassPrivate;
+        } else {
+          variable_storage_class = SpvStorageClassFunction;
+        }
```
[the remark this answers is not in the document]
andreperezmaselco: You're right.
Done.

source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
```diff
@@ -0,0 +1,108 @@
...
+               "The opcode of the instruction we might insert before must be "
+               "the same as the opcode in the descriptor for the instruction");
+
+        // Randomly decide whether to try pushing an id through a variable.
+        if (!GetFuzzerContext()->ChoosePercentage(
+                GetFuzzerContext()->GetChanceOfPushingIdThroughVariable())) {
+          return;
+        }
+
+        // The instruction to insert before containing block must be reachable.
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_push_ids_through_variables.cpp
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        // Gets the available basic and pointer types.
+        auto basic_type_ids_and_pointers =
+            GetAvailableBasicTypesAndPointers(variable_storage_class);
+        auto& basic_types = basic_type_ids_and_pointers.first;
+        uint32_t basic_type_id =
+            basic_types[GetFuzzerContext()->RandomIndex(basic_types)];
+
+        // Looks for values we might consider to store.
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+#include "source/fuzz/transformation_push_id_through_variable.h"
+
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/instruction_descriptor.h"
+#include "source/fuzz/transformation_add_global_variable.h"
+#include "source/fuzz/transformation_add_local_variable.h"
+#include "source/fuzz/transformation_add_type_pointer.h"
+#include "source/fuzz/transformation_load.h"
+#include "source/fuzz/transformation_store.h"
```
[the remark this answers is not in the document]
andreperezmaselco: You're right.
I forgot to remove it.
Done.

source/fuzz/fuzzer_pass_push_ids_through_variables.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "source/fuzz/fuzzer_pass.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// For each instruction with a descriptor, checks if OpStore and
+// OpLoad instructions can be inserted before it. So, randomly,
+// looks for values, adds whether a global or local variable
+// and applies the push id through variable transformation.
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,174 @@
...
+  if (!pointer_type_instruction ||
+      pointer_type_instruction->opcode() != SpvOpTypePointer) {
+    return false;
+  }
+
+  // The pointer type storage class must be equal to the variable storage class.
+  if (pointer_type_instruction->GetSingleWordInOperand(0) !=
+      message_.variable_storage_class()) {
+    return false;
+  }
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+  // The pointee type must be equal to the value type.
+  if (pointer_type_instruction->GetSingleWordInOperand(1) !=
+      value_instruction->type_id()) {
+    return false;
+  }
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.h
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,174 @@
...
+      FindInstruction(message_.instruction_descriptor(), ir_context);
+  if (!instruction_to_insert_before) {
+    return false;
+  }
+
+  // The instruction to insert before must belong to a reachable block.
+  auto basic_block = ir_context->get_instr_block(instruction_to_insert_before);
+  if (!fuzzerutil::BlockIsReachableInItsFunction(ir_context, basic_block)) {
+    return false;
+  }
```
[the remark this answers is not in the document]
afd: I'm gonna resolve this, assuming that it is required.

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,154 @@
...
+      ir_context->get_def_use_mgr()->GetDef(message_.value_id());
+  if (!value_instruction || !value_instruction->type_id()) {
+    return false;
+  }
+
+  // A pointer type instruction pointing to the value type must be defined.
+  auto pointer_type_id = fuzzerutil::MaybeGetPointerType(
+      ir_context, value_instruction->type_id(),
+      static_cast<SpvStorageClass>(message_.variable_storage_class()));
+  assert(pointer_type_id && "The required pointer type must be available.");
```
afd, line 79: This is not right: we are in IsApplicable here.  If the required pointer type is not available we should return false.

Please make sure there is a test for this case - a test where the transformation is only inapplicable because the pointer type is not present.  There should be no assertion failure - just a return of false.

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+  // |message_.variable_storage_class| must be private or function.
+  assert((message_.variable_storage_class() == SpvStorageClassPrivate ||
+          message_.variable_storage_class() == SpvStorageClassFunction) &&
+         "The variable storage class must be private or function.");
```
afd, line 84: In contrast it is correct to use assertions here because - no matter what the shrinker does - we should never see a different storage class being passed as `variable_storage_class`.

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+void TransformationPushIdThroughVariable::Apply(
+    opt::IRContext* ir_context,
+    TransformationContext* transformation_context) const {
+  auto value_instruction =
+      ir_context->get_def_use_mgr()->GetDef(message_.value_id());
+  auto pointer_type_id = fuzzerutil::MaybeGetPointerType(
+      ir_context, value_instruction->type_id(),
+      static_cast<SpvStorageClass>(message_.variable_storage_class()));
+
```
afd, line 99: It would be good to assert that pointer_type_id is non-zero here - basically the assert you have in IsApplicable that I complain about above :)

test/fuzz/transformation_push_id_through_variable_test.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+#include "source/fuzz/transformation_push_id_through_variable.h"
+#include "source/fuzz/instruction_descriptor.h"
+#include "test/fuzz/fuzz_test_util.h"
+
+namespace spvtools {
+namespace fuzz {
+namespace {
+
+TEST(TransformationPushIdThroughVariableTest, IsApplicable) {
```
afd, line 23: I did not scrutinise the tests carefully, but in your tests please make sure you have covered all the cases where IsApplicable can return false.

source/fuzz/transformation_push_id_through_variable.cpp
andreperezmaselco: Done.

andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
andreperezmaselco: Done.

test/fuzz/transformation_push_id_through_variable_test.cpp
andreperezmaselco: Done.

source/fuzz/transformation_push_id_through_variable.cpp
```diff
@@ -76,7 +76,9 @@ bool TransformationPushIdThroughVariable::IsApplicable(
   auto pointer_type_id = fuzzerutil::MaybeGetPointerType(
```
afd, line 76: Nit: I like the way you have a comment before each condition - I would take the message from the assertion you removed and use it as a comment:

// The required pointer type must be available.

afd: My bad - I don't know how but I didn't see the comment (the diff must have confused me).

PR #3363, opened 2020-05-20
Add validation support for ImageGatherBiasLodAMD

rg3igalia:
ImageGatherBiasLodAMD makes it possible to use the Bias and Lod image
operands with OpImageGather and OpImageSparseGather. This commit makes
sure the validator checks for that capability before reporting errors
and adds a few positive tests.


s-perron:
Looks good to me, but Alan knows the validator better than I do.

PR #3379, opened 2020-05-25
Avoid operand type range checks

dneto0:
Deprecates the SPV_OPERAND_TYPE_FIRST_* and SPV_OPERAND_TYPE_LAST_*
macros.

The "variable" and "optional" operand types are only for internal use.
Export spvOperandIsConcrete instead, as that should cover intended
external uses.

Test that each operand type is classified either as one of:
- a sentinel value
- a concrete operand type
- an optional operand type (which includes variable-expansion types)

Test that each concrete and optional non-variable operand type
has a name for use internally when generating messages.

### DIFF
--- a/include/spirv-tools/libspirv.h
+++ b/include/spirv-tools/libspirv.h
@@ -48,8 +48,8 @@ extern "C" {
 
 #define SPV_BIT(shift) (1 << (shift))
 
-#define SPV_FORCE_16_BIT_ENUM(name) _##name = 0x7fff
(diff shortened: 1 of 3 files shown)


dneto0:
I did this because having range enum values in a public header works against binary compatibility, and binary compatibility should not be broken lightly.

dneto0:
Fixed formatting and rebased against master

s-perron:
I've rebased this and I believe I have fixed the test failure.

PR #3385, opened 2020-06-01
spirv-fuzz: Fix assertion failure related to transformation applicability

andreperezmaselco:
This change fixes an assertion failure related to the push
id through variable transformation. In the fuzzer pass class
it was missing the `IdIsAvailableBeforeInstruction` condition
that is checked in the `IsApplicable` function.


paulthomson:
This looks good to me. We don't have a good way to test fuzzer passes right now, so I don't think a test is required.

PR #3390, opened 2020-06-02
spirv-fuzz: Fix instruction function use

andreperezmaselco:
This change replaces the instruction function `GetSingleWordOperand`
with `GetSingleWordInOperand`. Debugging spirv-fuzz, `GetSingleWordOperand`
was returning the result id of an OpLoad instruction with memory operands.


paulthomson:
Looks good to me! Let's wait for the CI to be fixed before testing.

PR #3402, opened 2020-06-05
spirv-fuzz: Add replace linear algebra instruction transformation

andreperezmaselco:
This PR implements a transformation that replaces
a linear algebra instruction with its mathematical definition.

### DIFF
--- a/source/fuzz/protobufs/spvtoolsfuzz.proto
+++ b/source/fuzz/protobufs/spvtoolsfuzz.proto
@@ -1091,6 +1091,17 @@ message TransformationReplaceLinearAlgebraInstruction {
   repeated uint32 fresh_ids = 1;
 
   // A descriptor for a linear algebra instruction.
+  // This transformation is only applicable if the described instruction has one of the following opcodes.
+  // Supported:
+  //   OpVectorTimesScalar
+  //   OpDot
+  // Unsupported:
+  //   OpTranspose
+  //   OpMatrixTimesScalar
+  //   OpVectorTimesMatrix
+  //   OpMatrixTimesVector
+  //   OpMatrixTimesMatrix
+  //   OpOuterProduct
   InstructionDescriptor instruction_descriptor = 2;
 
 }
--- a/source/fuzz/protobufs/spvtoolsfuzz.proto
+++ b/source/fuzz/protobufs/spvtoolsfuzz.proto
@@ -1095,7 +1095,10 @@ message TransformationReplaceLinearAlgebraInstruction {
   // Supported:
   //   OpVectorTimesScalar
   //   OpDot
-  // Unsupported:
+  // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3354):
+  // Right now we only support certain operations. When this issue is addressed
+  // the supporting comments can be removed.
+  // To be supported in the future:
   //   OpTranspose
   //   OpMatrixTimesScalar
   //   OpVectorTimesMatrix
--- a/source/fuzz/protobufs/spvtoolsfuzz.proto
+++ b/source/fuzz/protobufs/spvtoolsfuzz.proto
@@ -377,6 +377,7 @@ message Transformation {
     TransformationAdjustBranchWeights adjust_branch_weights = 46;
     TransformationPushIdThroughVariable push_id_through_variable = 47;
     TransformationAddSpecConstantOp add_spec_constant_op = 48;
+    TransformationReplaceLinearAlgebraInstruction replace_linear_algebra_instruction = 49;
     // Add additional option using the next available number.
   }
 }
@@ -1078,6 +1079,19 @@ message TransformationReplaceIdWithSynonym {
 
   // The synonymous id
   uint32 synonymous_id = 2;
+
+}
+
+message TransformationReplaceLinearAlgebraInstruction {
+
+  // Replaces a linear algebra instruction with its
+  // mathematical definition.
+
+  // The fresh ids needed to apply the transformation.
+  repeated uint32 fresh_ids = 1;
+
+  // A descriptor for a linear algebra instruction.
+  InstructionDescriptor instruction_descriptor = 2;
 
 }
 
--- a/source/fuzz/transformation_replace_linear_algebra_instruction.cpp
+++ b/source/fuzz/transformation_replace_linear_algebra_instruction.cpp
@@ -1,0 +1,268 @@
+// Copyright (c) 2020 André Perez Maselco
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/transformation_replace_linear_algebra_instruction.h"
+
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/instruction_descriptor.h"
+
+namespace spvtools {
+namespace fuzz {
+
+TransformationReplaceLinearAlgebraInstruction::
+    TransformationReplaceLinearAlgebraInstruction(
+        const spvtools::fuzz::protobufs::
+            TransformationReplaceLinearAlgebraInstruction& message)
+    : message_(message) {}
+
+TransformationReplaceLinearAlgebraInstruction::
+    TransformationReplaceLinearAlgebraInstruction(
+        const std::vector<uint32_t>& fresh_ids,
+        const protobufs::InstructionDescriptor& instruction_descriptor) {
+  for (auto fresh_id : fresh_ids) {
+    message_.add_fresh_ids(fresh_id);
+  }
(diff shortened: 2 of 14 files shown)


source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1062,6 +1063,19 @@ message TransformationReplaceIdWithSynonym {
...
 
+message TransformationReplaceLinearAlgebraInstruction {
+
+  // Replaces a linear algebra instruction with its
+  // mathematical definition.
+
+  // The fresh ids needed to apply the transformation.
+  repeated uint32 fresh_ids = 1;
+
+  // A descriptor for a linear algebra instruction.
```
afd, line 1074: I would add a note explaining that the transformation is only applicable if the described instruction has one of a specific set of opcodes, and then list the opcodes that are currently supported (to be updated with more opcodes in future PRs).

source/opcode.cpp
```diff
@@ -650,6 +650,19 @@ bool spvOpcodeIsCommutativeBinaryOperator(SpvOp opcode) {
   }
 }
 
+bool spvOpcodeIsLinearAlgebra(SpvOp opcode) {
+  // TODO: implement the cases OpMatrixTimesScalar, OpVectorTimesMatrix,
+  // OpMatrixTimesVector, OpMatrixTimesMatrix and OpOuterProduct in the
+  // TransformationReplaceLinearAlgebraInstruction class.
```
afd, line 656: We don't really want this part of the code base, which has nothing to do with spirv-fuzz, make reference to spirv-fuzz-specific stuff.

I suggest you make this function complete - fill out the switch cases.  And then in the transformation add some code that restricts to the cases that we support.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
```diff
@@ -0,0 +1,268 @@
...
+bool TransformationReplaceLinearAlgebraInstruction::IsApplicable(
+    opt::IRContext* ir_context, const TransformationContext& /*unused*/) const {
+  auto instruction =
+      FindInstruction(message_.instruction_descriptor(), ir_context);
+
+  // It must be a linear algebra instruction.
+  if (!spvOpcodeIsLinearAlgebra(instruction->opcode())) {
+    return false;
+  }
+
```
afd, line 48: Here, add:

```
// TODO(url for bug for completing the other cases): ...
if (instruction->opcode != ... && instruction->opcode != ...) {
    return false;
}
```

Then once all the cases covered by spvOpCodeIsLinearAlgebra are dealt with, the TODO and the conditional can go.

Also see my comment on spvOpcodeIsLinearAlgebra.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
```diff
@@ -0,0 +1,268 @@
...
+  // |message_.fresh_ids.size| must be the exact number of fresh ids needed to
+  // apply the transformation.
+  if ((uint32_t)message_.fresh_ids().size() !=
```
afd, line 51: use static_cast<uint32_t>(...)

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
```diff
@@ -0,0 +1,268 @@
...
+      GetFreshIdCount(ir_context, instruction)) {
+    return false;
+  }
+
+  // All ids in |message_.fresh_ids| must be fresh.
+  for (uint32_t i = 0; i < (uint32_t)message_.fresh_ids().size(); i++) {
```
afd, line 57: static_cast

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+  switch (linear_algebra_instruction->opcode()) {
+    case SpvOpVectorTimesScalar:
+      ReplaceOpVectorTimesScalar(ir_context, linear_algebra_instruction);
+      break;
+    case SpvOpDot:
+      ReplaceOpDot(ir_context, linear_algebra_instruction);
+      break;
+    default:
+      break;
```
afd, line 79: I suggest adding an assert(false && "Should be unreachable."); before the break.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
```diff
@@ -0,0 +1,268 @@
...
+}
+
+protobufs::Transformation
+TransformationReplaceLinearAlgebraInstruction::ToMessage() const {
+  protobufs::Transformation result;
+  *result.mutable_replace_linear_algebra_instruction() = message_;
+  return result;
+}
+
+uint32_t TransformationReplaceLinearAlgebraInstruction::GetFreshIdCount(
```
afd, line 92: I suggest calling it GetRequiredFreshIdCount.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
```diff
@@ -0,0 +1,268 @@
...
+    opt::IRContext* ir_context, opt::Instruction* instruction) {
+  // TODO for OpMatrixTimesScalar, OpVectorTimesMatrix, OpMatrixTimesVector,
+  // OpMatrixTimesMatrix and OpOuterProduct.
```
afd, line 95: Change to a TODO(bug url) that links to the GitHub issue for finishing this off.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
```diff
@@ -0,0 +1,268 @@
...
+                     ->GetType(
+                         ir_context->get_def_use_mgr()
+                             ->GetDef(instruction->GetSingleWordInOperand(0))
+                             ->type_id())
+                     ->AsVector()
+                     ->element_count() -
+             2;
+    }
+    default:
+      return 0;
```
afd, line 123: Add an "unreachable" assertion before the return.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
```diff
@@ -0,0 +1,268 @@
...
+        ir_context, SpvOpFMul, linear_algebra_instruction->type_id(),
+        float_multiplication_ids[i],
+        opt::Instruction::OperandList(
+            {{SPV_OPERAND_TYPE_ID, {vector_1_extract_id}},
+             {SPV_OPERAND_TYPE_ID, {vector_2_extract_id}}})));
+  }
+
+  // If the vector has 2 components, then there will be 2 float multiplication
+  // instructions.
+  if (vectors_component_count == 2) {
```
afd, line 225: I feel like there should be a way to harmonize this so that the two blocks of code that adjust the existing instruction are not both needed.

Give it some thought but OK if not.

source/fuzz/transformation_replace_linear_algebra_instruction.h
```diff
@@ -0,0 +1,67 @@
...
+  static uint32_t GetFreshIdCount(opt::IRContext* ir_context,
+                                  opt::Instruction* instruction);
+
+  // Replaces an OpVectorTimesScalar instruction.
+  void ReplaceOpVectorTimesScalar(opt::IRContext* ir_context,
+                                  opt::Instruction* instruction) const;
+
+  // Replaces an OpDot instruction.
+  void ReplaceOpDot(opt::IRContext* ir_context,
+                    opt::Instruction* instruction) const;
```
afd, line 58: These can be private, I believe.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
andreperezmaselco: I was thinking about it. But for a 2-dimensional vector we make just an instruction replacement and not an insertion. For now, that is the way it works better.

source/fuzz/protobufs/spvtoolsfuzz.proto
andreperezmaselco: Done.

source/opcode.cpp
andreperezmaselco: Done.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
andreperezmaselco: Done.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
andreperezmaselco: Done.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
andreperezmaselco: Done.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
andreperezmaselco: Done.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
andreperezmaselco: Done.

source/fuzz/transformation_replace_linear_algebra_instruction.cpp
andreperezmaselco: Done.

source/fuzz/transformation_replace_linear_algebra_instruction.h
andreperezmaselco: Done.

PR #3404, opened 2020-06-05
Avoid confusing short-circuiting

dneto0:
When checking the OpBranchConditional for selection headers,
we intend to register both the true and false targets.
Short circuiting was getting in the way.


test/val/val_cfg_test.cpp
```diff
@@ -4503,6 +4503,67 @@ OpFunctionEnd
   ASSERT_EQ(SPV_SUCCESS, ValidateInstructions());
 }
 
+TEST_F(ValidateCFG, StructuredSelections_RegisterBothTrueAndFalse) {
```
alan-baker, line 4506: Is there anyway to actually trigger this case?

PR #3405, opened 2020-06-05
WIP: Ifbreak is divergence

dneto0:
FYI:  Prototype of proposed rule that a branch to the merge of an if-selection should count as divergence.

I've adjusted a couple of other tests that run afoul of this rule.
I haven't updated anything in the fuzzer.


dneto0:
Lowering the priority on this.
I've figured out the control flow reconstruction algorithm for it. (https://dawn-review.googlesource.com/c/tint/+/23182)

PR #3411, opened 2020-06-07
spirv-val: Extend variable pointer return cases

### LINKED ISSUE #3410: spirv-val: structure with variable pointer member
`spirv-val` invalidated a fuzzed shader because  `OpCompositeExtract` was extracting a variable pointer from a structure. The result id was going to be used in a `OpStore` instruction. So, `spir-val` logged `OpStore Pointer <id> is not a logical pointer.`

andreperezmaselco:
Fixes #3410.

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -289,6 +289,7 @@ bool spvOpcodeReturnsLogicalVariablePointer(const SpvOp opcode) {
     case SpvOpPtrAccessChain:
     case SpvOpLoad:
     case SpvOpConstantNull:
+    case SpvOpCompositeExtract:
       return true;
     default:
       return false;


andreperezmaselco:
@alan-baker 
But what if the variable pointer is a structure member going to be used in a `OpStore` instruction?

alan-baker:
> @alan-baker
> But what if the variable pointer is a structure member going to be used in a `OpStore` instruction?

Probably you have something like a function or private variable containing an aggregate of pointers. In that case you should use OpAccessChain to resolve the correct member.

andreperezmaselco:
> > @alan-baker
> > But what if the variable pointer is a structure member going to be used in a `OpStore` instruction?
> 
> Probably you have something like a function or private variable containing an aggregate of pointers. In that case you should use OpAccessChain to resolve the correct member.

Ok.
Thanks!

PR #3412, opened 2020-06-08
spirv-fuzz: Implement vector shuffle fuzzer pass

### LINKED ISSUE #3108: spirv-fuzz: Write fuzzer pass to apply vector shuffles
We have a transformation to add a vector shuffle to a module, generating corresponding data synonyms.  We need a pass that will actually generate these transformations.

andreperezmaselco:
Fixes #3108.

### DIFF
--- a/source/fuzz/fuzzer_pass_add_vector_shuffle_instructions.cpp
+++ b/source/fuzz/fuzzer_pass_add_vector_shuffle_instructions.cpp
@@ -100,7 +100,9 @@ void FuzzerPassAddVectorShuffleInstructions::Apply() {
                                  ->AsVector();
 
         // |vector_1| and |vector_2| must have the same element type as each
-        // other.
+        // other. The loop is guaranteed to terminate because each iteration
+        // removes on possible choice for |vector_2|, and there is at least one
+        // choice that will cause the loop to exit - namely |vector_1|.
         while (vector_1_type->element_type() != vector_2_type->element_type()) {
           vector_2_instruction =
               GetFuzzerContext()->RemoveAtRandomIndex(&vector_instructions);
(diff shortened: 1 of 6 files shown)


source/fuzz/fuzzer_pass_add_vector_shuffle_instructions.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+        auto vector_1_instruction =
+            vector_instructions[GetFuzzerContext()->RandomIndex(
+                vector_instructions)];
+        auto vector_1_type = GetIRContext()
+                                 ->get_type_mgr()
+                                 ->GetType(vector_1_instruction->type_id())
+                                 ->AsVector();
+
+        auto vector_2_instruction =
```
afd, line 95: At first I was worried that there might not exist a compatible vector, and that the loop might run forever.

Of course, that's not possible - because vector_1 is a choice that will definitely work.

However, I suggest you use FuzzerContext::RemoveAtRandomIndex each time you consider a choice of vector2.  That way you gradually eliminate all the choices that won't work.

This could be really good if e.g. we need an int vector and there are thousands of float vectors available but not many int vectors.  The chance of choosing a float vector will decrease gradually as we remove them.

source/fuzz/fuzzer_pass_add_vector_shuffle_instructions.cpp
```diff
@@ -0,0 +1,133 @@
...
+            vector_instructions[GetFuzzerContext()->RandomIndex(
+                vector_instructions)];
+        auto vector_2_type = GetIRContext()
+                                 ->get_type_mgr()
+                                 ->GetType(vector_2_instruction->type_id())
+                                 ->AsVector();
+
+        // |vector_1| and |vector_2| must have the same element type with each
+        // other.
```
afd, line 104: "with" -> "as"

source/fuzz/fuzzer_pass_add_vector_shuffle_instructions.cpp
andreperezmaselco: Great.
Done.

source/fuzz/fuzzer_pass_add_vector_shuffle_instructions.cpp
andreperezmaselco: Done.

PR #3420, opened 2020-06-10
Debug info preservation in ccp pass

jaebaek:
For constant propagation, the ccp pass already replaces the result id of
a value with a result id of the corresponding constant value. As a part
of the replacement, it correctly updates the operands of
DebugValue/DebugDeclare as well. Since we do not have to any addition
work other than the ccp pass itself, this commit just adds unit tests to
check the debug information preservation.

### DIFF
--- a/test/opt/ccp_test.cpp
+++ b/test/opt/ccp_test.cpp
@@ -927,13 +927,6 @@ TEST_F(CCPTest, FoldWithDecoration) {
 
 TEST_F(CCPTest, DebugSimpleFoldConstant) {
   const std::string text = R"(
-; CHECK: [[float1:%\w+]] = OpConstant {{%\w+}} 1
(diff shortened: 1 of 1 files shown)


test/opt/ccp_test.cpp
```diff
@@ -925,6 +925,133 @@ TEST_F(CCPTest, FoldWithDecoration) {
   SinglePassRunAndMatch<CCPPass>(text, true);
 }
 
+TEST_F(CCPTest, DebugSimpleFoldConstant) {
+  const std::string text = R"(
+; CHECK: [[float1:%\w+]] = OpConstant {{%\w+}} 1
```
dnovillo, line 930: Can you move the checks next to the lines they are checking?  Much easier to read that way.  OK with that change.

jaebaek: done

PR #3433, opened 2020-06-15
Update access control lists.

dj2:
This CL updates the access control lists used in SPIRV-Tools to the more
descriptive allow/deny naming.


s-perron:
What is wrong with whilelist and blacklist?  Aren't those common enough terms that are well defined?  Can you point to something that explains why these terms are not good enough?

dj2:
They maybe well defined, but allow/deny are clearer terms for what the list is doing.

PR #3439, opened 2020-06-16
spirv-fuzz: Add image sample unused components transformation

### LINKED ISSUE #3375: spirv-fuzz: Add unused components to image instructions
We had a bug report that ended up being triggered by an image instruction having unused components in the coordinate. The SPIR-V spec for image instructions has language like:

> Coordinate must be a scalar or vector of floating-point type or integer type. It contains (u[, v] … [, array layer]) as needed by the definition of Sampled Image. Unless the Kernel capability is being used, it must be floating point. **It may be a vector larger than needed, but all unused components will appear after all used components.**

Adding unused components seems ideally suited for the fuzzer.

andreperezmaselco:
Fixes #3375.

### DIFF
--- a/source/fuzz/fuzzer_pass_add_image_sample_unused_components.cpp
+++ b/source/fuzz/fuzzer_pass_add_image_sample_unused_components.cpp
@@ -1,0 +1,142 @@
+// Copyright (c) 2020 André Perez Maselco
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_add_image_sample_unused_components.h"
+
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/instruction_descriptor.h"
+#include "source/fuzz/transformation_add_image_sample_unused_components.h"
+#include "source/fuzz/transformation_composite_construct.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassAddImageSampleUnusedComponents::
+    FuzzerPassAddImageSampleUnusedComponents(
+        opt::IRContext* ir_context,
+        TransformationContext* transformation_context,
+        FuzzerContext* fuzzer_context,
+        protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassAddImageSampleUnusedComponents::
+    ~FuzzerPassAddImageSampleUnusedComponents() = default;
+
+void FuzzerPassAddImageSampleUnusedComponents::Apply() {
+  GetIRContext()->module()->ForEachInst([this](opt::Instruction* instruction) {
+    if (!spvOpcodeIsImageSample(instruction->opcode())) {
+      return;
+    }
+
+    if (!GetFuzzerContext()->ChoosePercentage(
+            GetFuzzerContext()
+                ->GetChanceOfAddingImageSampleUnusedComponents())) {
+      return;
+    }
+
+    // Gets image sample coordinate information.
+    uint32_t coordinate_id = instruction->GetSingleWordInOperand(1);
+    auto coordinate_instruction =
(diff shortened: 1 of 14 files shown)


source/fuzz/fuzzer_pass_add_image_sample_unused_components.cpp
```diff
@@ -0,0 +1,142 @@
...
+                : FindOrCreateVectorType(
+                      GetIRContext()->get_type_mgr()->GetId(
+                          coordinate_type->AsVector()->element_type()),
+                      unused_component_count);
+        break;
+      default:
+        break;
+    }
+
+    // Gets |coordinate_type| again because the module has changed.
```
afd, line 107: "because the module has changed." -> "because the module may have changed due to the use of FindOrCreateVectorType above."

source/fuzz/fuzzer_pass_add_image_sample_unused_components.cpp
```diff
@@ -0,0 +1,142 @@
...
+    if (!spvOpcodeIsImageSample(instruction->opcode())) {
+      return;
+    }
+
+    if (!GetFuzzerContext()->ChoosePercentage(
+            GetFuzzerContext()
+                ->GetChanceOfAddingImageSampleUnusedComponents())) {
+      return;
+    }
+
```
afd, line 48: The following looks very good.  However, it is intricate and will be hard to understand for someone who doesn't already know this instruction.

I suggest you add two worked examples, one where the existing type is int, the other where it is vec2 (i.e., a 2d floating point vector).  Then at each stage of the algorithm you show using psesudocode what would be computed in each case.

Take a look at what I've done in transformation_add_function.cpp from lines 586-614 - there are little snippets of illustrative SPIR-V.  I think something along those lines could be useful here.  It will be different as you'll be describing existing SPIR-V and the types that will be needed based on it.

source/fuzz/transformation_add_image_sample_unused_components.cpp
```diff
@@ -0,0 +1,115 @@
...
+  auto coordinate_with_unused_components_instruction =
+      ir_context->get_def_use_mgr()->GetDef(
+          message_.coordinate_with_unused_components_id());
+
+  // The coordinate with unused components instruction must be defined.
+  if (coordinate_with_unused_components_instruction == nullptr) {
+    return false;
+  }
+
+  // It must be an OpCompositeConstruct instruction.
```
afd, line 74: Am I correct that it doesn't have to be OpCompositeConstruct in order for the module to be valid, but rather we need it to be this instruction because we need to look inside the construction to check that the original components are there?  I think it would be good to clarify that slightly in the comment.

source/fuzz/transformation_add_image_sample_unused_components.cpp
```diff
@@ -0,0 +1,115 @@
...
+  return true;
+}
+
+void TransformationAddImageSampleUnusedComponents::Apply(
+    opt::IRContext* ir_context, TransformationContext* /*unused*/) const {
+  // Sets the coordinate operand.
+  auto image_sample_instruction =
+      FindInstruction(message_.instruction_descriptor(), ir_context);
+  image_sample_instruction->SetInOperand(
+      1, {message_.coordinate_with_unused_components_id()});
```
afd, line 104: After this we should invalidate analyses since the module has changed.  I recommend invalidating all analyses - as is done in most other transformations - out of caution.

source/fuzz/fuzzer_pass_add_image_sample_unused_components.cpp
andreperezmaselco: Done.

source/fuzz/transformation_add_image_sample_unused_components.cpp
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_add_image_sample_unused_components.cpp
andreperezmaselco: Don't you think the comments explain by themselves?
I'm trying to explain with code, but I guess it is redundant.

source/fuzz/transformation_add_image_sample_unused_components.cpp
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_add_image_sample_unused_components.cpp
afd: For me, a worked example alongside the comments would be really useful.  The comments explain what the code does *in general*.   A worked example would show what the code does to a specific fragment of SPIR-V.  I think the two together would be great.

Can you have a go?

andreperezmaselco: Done.

PR #3442, opened 2020-06-17
Add support for different default branches in roll-deps

zoddicus:
This in anticipation of some repos changing their default branch name. This is based off to the change I made in https://github.com/google/shaderc/pull/1086

### DIFF
--- a/utils/roll_deps.sh
+++ b/utils/roll_deps.sh
@@ -13,19 +13,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Attempts to roll all entries in DEPS to origin/master and creates a
-# commit.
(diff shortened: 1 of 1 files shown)


utils/roll_deps.sh
```diff
@@ -13,19 +13,34 @@
...
 # limitations under the License.
 
-# Attempts to roll all entries in DEPS to origin/master and creates a
-# commit.
+# Attempts to roll all entries in DEPS to tip-of-tree and create a commit.
 #
 # Depends on roll-dep from depot_path being in PATH.
 
+effcee_dir="third_party/effcee/"
+effcee_trunk="origin/master"
```
dj2, line 21: I believe effcee should be origin/main

zoddicus: Correct

PR #3444, opened 2020-06-17
Debug info preservation in copy-prop-array pass

### LINKED ISSUE #3429: OpenCL.DebugInfo.100 information preservation in copy-propagate-arrays pass

jaebaek:
When the pass replaces the local variable `OpVariable` ids to their
corresponding pointers, we have to update operands of DebugValue or
DebugDeclare instructions that are the local variable ids.

Fixes #3429

### DIFF
--- a/source/opt/copy_prop_arrays.cpp
+++ b/source/opt/copy_prop_arrays.cpp
@@ -29,6 +29,12 @@ const uint32_t kCompositeExtractObjectInOperand = 0;
 const uint32_t kTypePointerStorageClassInIdx = 0;
 const uint32_t kTypePointerPointeeInIdx = 1;
 
+bool IsOpenCL100DebugInstructionReferencingOpVariable(Instruction* di) {
+  auto dbg_opcode = di->GetOpenCL100DebugOpcode();
+  return dbg_opcode == OpenCLDebugInfo100DebugDeclare ||
+         dbg_opcode == OpenCLDebugInfo100DebugValue;
+}
+
 }  // namespace
 
 Pass::Status CopyPropagateArrays::Process() {
@@ -188,6 +194,8 @@ bool CopyPropagateArrays::HasValidReferencesOnly(Instruction* ptr_inst,
           return ptr_inst->opcode() == SpvOpVariable &&
                  store_inst->GetSingleWordInOperand(kStorePointerInOperand) ==
                      ptr_inst->result_id();
+        } else if (IsOpenCL100DebugInstructionReferencingOpVariable(use)) {
+          return true;
         }
         // Some other instruction.  Be conservative.
         return false;
@@ -492,6 +500,8 @@ bool CopyPropagateArrays::CanUpdateUses(Instruction* original_ptr_inst,
                                                        const_mgr,
                                                        type](Instruction* use,
                                                              uint32_t) {
+    if (IsOpenCL100DebugInstructionReferencingOpVariable(use)) return true;
+
     switch (use->opcode()) {
       case SpvOpLoad: {
         analysis::Pointer* pointer_type = type->AsPointer();
@@ -565,6 +575,7 @@ bool CopyPropagateArrays::CanUpdateUses(Instruction* original_ptr_inst,
     }
   });
 }
+
 void CopyPropagateArrays::UpdateUses(Instruction* original_ptr_inst,
                                      Instruction* new_ptr_inst) {
   analysis::TypeManager* type_mgr = context()->get_type_mgr();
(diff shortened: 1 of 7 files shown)


source/opt/copy_prop_arrays.cpp
```diff
@@ -580,6 +591,39 @@ void CopyPropagateArrays::UpdateUses(Instruction* original_ptr_inst,
   for (auto pair : uses) {
     Instruction* use = pair.first;
     uint32_t index = pair.second;
+
+    if (use->GetOpenCL100DebugOpcode() != OpenCLDebugInfo100InstructionsMax) {
+      switch (use->GetOpenCL100DebugOpcode()) {
+        case OpenCLDebugInfo100DebugDeclare: {
+          context()->ForgetUses(use);
```
s-perron, line 598: Might be worth having a comment on what you are doing and why.  It might not be clear to other why you have to change the DebugDeclare to a DebugValue.

source/opt/pass.h
```diff
@@ -71,6 +71,10 @@ class Pass {
     return context()->get_def_use_mgr();
   }
 
+  analysis::DebugInfoManager* get_debug_info_mgr() const {
+    return context()->get_debug_info_mgr();
+  }
+
```
s-perron, line 77: The functions like this that already exist are a relic.  I do not like getting the analysis from the pass because it gives the impression they belong to the pass.  I would prefer if this was not added.

source/opt/debug_info_manager.cpp
```diff
@@ -210,6 +212,52 @@ uint32_t DebugInfoManager::BuildDebugInlinedAtChain(
...
+      context()->module()->ext_inst_debuginfo_begin()->InsertBefore(
+          std::move(deref_operation));
+
+  RegisterDbgInst(deref_operation_);
+  if (context()->AreAnalysesValid(IRContext::Analysis::kAnalysisDefUse))
+    context()->get_def_use_mgr()->AnalyzeInstDefUse(deref_operation_);
+  return deref_operation_;
+}
+
+Instruction* DebugInfoManager::CloneDebugExpressionAndPushDerefOperation(
```
s-perron, line 244: The name does not roll off the tongue.  I think something like "DerefDebugExpression" would be better.

source/opt/debug_info_manager.cpp
```diff
@@ -210,6 +212,52 @@ uint32_t DebugInfoManager::BuildDebugInlinedAtChain(
...
+    Instruction* dbg_expr) {
+  assert(dbg_expr->GetOpenCL100DebugOpcode() ==
+         OpenCLDebugInfo100DebugExpression);
+  std::unique_ptr<Instruction> deref_expr(dbg_expr->Clone(context()));
+  deref_expr->SetResultId(context()->TakeNextId());
+  deref_expr->InsertOperand(
+      kDebugExpressOperandOperationIndex,
+      {SPV_OPERAND_TYPE_ID, {GetDebugOperationWithDeref()->result_id()}});
```
s-perron, line 252: Did you check the order in which the expression should be applied?  How do you know it should be the start of the expression instead of the end?

jaebaek:
```
struct S {
  bool a;
  char b;
} foo;

// When "%1" is "foo.a"
call void @llvm.dbg.value(metadata i8* %1, metadata !11, metadata !DIExpression(DW_OP_deref, DW_OP_LLVM_fragment, 0, 8)), !dbg !18
```
In the above code, `%1` is `foo.a` and
`DIExpression(DW_OP_deref, DW_OP_LLVM_fragment, 0, 8)` means
1. apply `DW_OP_deref`
2. apply `DW_OP_LLVM_fragment, 0, 8` that mean the member in the offset `0` whose size is `8`

If I change it to `call void @llvm.dbg.value(metadata i8* %1, metadata !11, metadata !DIExpression(DW_OP_LLVM_fragment, 0, 8, DW_OP_deref)), !dbg !18` and generate the assembly code, LLVM prints error
```
invalid expression
!DIExpression(4096, 0, 8, 6)
```
`4096` is the opcode for `DW_OP_LLVM_fragment` and `6` is the opcode for `DW_OP_deref`.

Therefore, in LLVM IR, the order of applying operations of `DIExpression` is from the lower indexed argument to the higher indexed argument.

I quickly checked https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/master/lib/SPIRV/SPIRVToLLVMDbgTran.cpp#L781 that is SPIRV-to-LLVM translation for `DebugExpression`.
It uses the same order when converting it to LLVM IR.

In my opinion, we can consider the order must be from the first operation to the last operation.

However, we have to **update the spec for the clarification**.

source/opt/copy_prop_arrays.cpp
s-perron: I don't think the explanation you added makes sense.  Someone might ask why you cannot replace one pointer with another?  `new_ptr_inst` is not a value it is a pointer.  Please try again.

jaebaek: After considering this more, I am confused what will be the correct approach to handle this case.
```
ImageArray input;
void main() {
  ImageArray foo;
  foo = input;
  // assume that value(foo) == value(input) == X
  ... use foo ...
  update value(input) = Y;
}

// Before copy-prop-array
%input = OpVariable %ptr_image_array_Input Input
...
%main = OpFunction
%foo = OpVariable %ptr_image_array_Function Function
OpStore %foo %input
DebugDeclare %dbg_foo %foo    --> value(foo) == value(input) == X here
... use %foo ...
... update %input to Y ...
// debugger_print(%dbg_foo) must be X here


// After copy-prop-array replaces %foo to %input
%input = OpVariable %ptr_image_array_Input Input
...
%main = OpFunction
DebugValue %dbg_foo %input Deref   // or DebugDeclare %dbg_foo %input
... use %foo ...
... update %input to Y ...
// debugger_print(%dbg_foo) must be Y, which is the value(input) not value(foo)
```

In the above example, since copy-prop-array pass replaces `%foo` to `%input`, the value inspection using a debugger at then end will be `Y` instead of `X`, where `Y` must be the value of `%input` and `X` must be the value of `%foo`.

In my opinion, if the `OpVariable` to be replaced has `DebugDeclare` or `DebugValue` with `Deref`, we have to skip replacing it with the other pointer because it will result in a wrong debugger value inspection result.

What do you think?

jaebaek:
Since a pointer other than `OpVariable` and `OpFunctionParameter` cannot be an operand of `DebugDeclare`, we have to use `DebugValue` for the case.

The code keeps `DebugDeclare` if the new pointer is `OpVariable` or `OpFunctionParameter`. Otherwise, it changes `DebugDeclare` to `DebugValue` with `Deref`.

PTAL!

PR #3451, opened 2020-06-19
Upgrade elim-local-single-block for OpenCL.DebugInfo.100

greg-lunarg:
Essentially creates a DebugValue when removing a store to a local variable.

### DIFF
--- a/source/opt/local_single_block_elim_pass.cpp
+++ b/source/opt/local_single_block_elim_pass.cpp
@@ -31,6 +31,11 @@ const uint32_t kStoreValIdInIdx = 1;
 bool LocalSingleBlockLoadStoreElimPass::HasOnlySupportedRefs(uint32_t ptrId) {
   if (supported_ref_ptrs_.find(ptrId) != supported_ref_ptrs_.end()) return true;
   if (get_def_use_mgr()->WhileEachUser(ptrId, [this](Instruction* user) {
+        auto dbg_op = user->GetOpenCL100DebugOpcode();
+        if (dbg_op == OpenCLDebugInfo100DebugDeclare ||
+            dbg_op == OpenCLDebugInfo100DebugValue) {
+          return true;
+        }
         SpvOp op = user->opcode();
         if (IsNonPtrAccessChain(op) || op == SpvOpCopyObject) {
           if (!HasOnlySupportedRefs(user->result_id())) {
@@ -73,10 +78,12 @@ bool LocalSingleBlockLoadStoreElimPass::LocalSingleBlockLoadStoreElim(
           // variable.
           if (ptrInst->opcode() == SpvOpVariable) {
             // If a previous store to same variable, mark the store
-            // for deletion if not still used.
+            // for deletion if not still used. Don't delete store
+            // if debugging; let ssa-rewrite and DCE handle it
             auto prev_store = var2store_.find(varId);
             if (prev_store != var2store_.end() &&
-                instructions_to_save.count(prev_store->second) == 0) {
+                instructions_to_save.count(prev_store->second) == 0 &&
+                !get_debug_info_mgr()->IsDebugDeclared(varId)) {
               instructions_to_kill.push_back(prev_store->second);
               modified = true;
             }
--- a/test/opt/local_single_block_elim.cpp
+++ b/test/opt/local_single_block_elim.cpp
@@ -1118,6 +1118,340 @@ TEST_F(LocalSingleBlockLoadStoreElimTest, VariablePointerTest) {
   )";
   SinglePassRunAndMatch<LocalSingleBlockLoadStoreElimPass>(text, false);
 }
+
+TEST_F(LocalSingleBlockLoadStoreElimTest, DebugDeclareTest) {
+  // If OpenCL.DebugInfo.100 enabled, check that store/load is still
+  // optimized, but stores are not deleted for store/store.
+  //
+  // struct PS_INPUT {
+  //   float4 c0 : COLOR0;
+  //   float4 c1 : COLOR1;
+  // };
+  //
+  // struct PS_OUTPUT {
+  //   float4 vColor : SV_Target0;
+  // };
+  //
+  // PS_OUTPUT MainPs(PS_INPUT i) {
+  //   PS_OUTPUT ps_output;
+  //   float4 c;
+  //   c = i.c0;
+  //   c += i.c1;
+  //   c /= 2.0;
+  //   ps_output.vColor = c;
+  //   return ps_output;
+  // }
+  const std::string text = R"(
+               OpCapability Shader
+          %1 = OpExtInstImport "OpenCL.DebugInfo.100"
+               OpMemoryModel Logical GLSL450
+               OpEntryPoint Fragment %MainPs "MainPs" %in_var_COLOR0 %in_var_COLOR1 %out_var_SV_Target0
+               OpExecutionMode %MainPs OriginUpperLeft
+          %6 = OpString "foo3.frag"
+          %7 = OpString "PS_OUTPUT"
+          %8 = OpString "float"
+          %9 = OpString "vColor"
+         %10 = OpString "PS_INPUT"
+         %11 = OpString "c1"
+         %12 = OpString "c0"
+         %13 = OpString "src.MainPs"
+         %14 = OpString "c"
+         %15 = OpString "ps_output"
+         %16 = OpString "i"
+               OpName %in_var_COLOR0 "in.var.COLOR0"
+               OpName %in_var_COLOR1 "in.var.COLOR1"
+               OpName %out_var_SV_Target0 "out.var.SV_Target0"
+               OpName %MainPs "MainPs"
+               OpName %PS_INPUT "PS_INPUT"
+               OpMemberName %PS_INPUT 0 "c0"
+               OpMemberName %PS_INPUT 1 "c1"
+               OpName %param_var_i "param.var.i"
+               OpName %PS_OUTPUT "PS_OUTPUT"
+               OpMemberName %PS_OUTPUT 0 "vColor"
+               OpName %src_MainPs "src.MainPs"
+               OpName %i "i"
+               OpName %bb_entry "bb.entry"
+               OpName %ps_output "ps_output"
+               OpName %c "c"
+               OpDecorate %in_var_COLOR0 Location 0
+               OpDecorate %in_var_COLOR1 Location 1
+               OpDecorate %out_var_SV_Target0 Location 0
+        %int = OpTypeInt 32 1
+      %int_0 = OpConstant %int 0
+      %int_1 = OpConstant %int 1
+      %float = OpTypeFloat 32
+    %float_2 = OpConstant %float 2
+    %v4float = OpTypeVector %float 4
+         %31 = OpConstantComposite %v4float %float_2 %float_2 %float_2 %float_2
+       %uint = OpTypeInt 32 0
+    %uint_32 = OpConstant %uint 32
+%_ptr_Input_v4float = OpTypePointer Input %v4float
+%_ptr_Output_v4float = OpTypePointer Output %v4float
+       %void = OpTypeVoid
+   %uint_128 = OpConstant %uint 128
+     %uint_0 = OpConstant %uint 0
+   %uint_256 = OpConstant %uint 256
+         %40 = OpTypeFunction %void
+   %PS_INPUT = OpTypeStruct %v4float %v4float
+%_ptr_Function_PS_INPUT = OpTypePointer Function %PS_INPUT
+  %PS_OUTPUT = OpTypeStruct %v4float
+         %42 = OpTypeFunction %PS_OUTPUT %_ptr_Function_PS_INPUT
+%_ptr_Function_PS_OUTPUT = OpTypePointer Function %PS_OUTPUT
+%_ptr_Function_v4float = OpTypePointer Function %v4float
+%in_var_COLOR0 = OpVariable %_ptr_Input_v4float Input
+%in_var_COLOR1 = OpVariable %_ptr_Input_v4float Input
+%out_var_SV_Target0 = OpVariable %_ptr_Output_v4float Output
+         %45 = OpExtInst %void %1 DebugSource %6
+         %46 = OpExtInst %void %1 DebugCompilationUnit 1 4 %45 HLSL
+         %47 = OpExtInst %void %1 DebugTypeComposite %7 Structure %45 8 1 %46 %7 %uint_128 FlagIsProtected|FlagIsPrivate %48
+         %49 = OpExtInst %void %1 DebugTypeBasic %8 %uint_32 Float
+         %50 = OpExtInst %void %1 DebugTypeVector %49 4
+         %48 = OpExtInst %void %1 DebugTypeMember %9 %50 %45 10 5 %47 %uint_0 %uint_128 FlagIsProtected|FlagIsPrivate
+         %51 = OpExtInst %void %1 DebugTypeComposite %10 Structure %45 2 1 %46 %10 %uint_256 FlagIsProtected|FlagIsPrivate %52 %53
+         %53 = OpExtInst %void %1 DebugTypeMember %11 %50 %45 5 5 %51 %uint_128 %uint_128 FlagIsProtected|FlagIsPrivate
+         %52 = OpExtInst %void %1 DebugTypeMember %12 %50 %45 4 5 %51 %uint_0 %uint_128 FlagIsProtected|FlagIsPrivate
+         %54 = OpExtInst %void %1 DebugTypeFunction FlagIsProtected|FlagIsPrivate %47 %51
+         %55 = OpExtInst %void %1 DebugFunction %13 %54 %45 13 1 %46 %13 FlagIsProtected|FlagIsPrivate 14 %src_MainPs
+         %56 = OpExtInst %void %1 DebugLexicalBlock %45 14 1 %55
+         %57 = OpExtInst %void %1 DebugLocalVariable %14 %50 %45 16 12 %56 FlagIsLocal
+         %58 = OpExtInst %void %1 DebugLocalVariable %15 %47 %45 15 15 %56 FlagIsLocal
+         %59 = OpExtInst %void %1 DebugExpression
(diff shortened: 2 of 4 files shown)


greg-lunarg:
This is WIP as I have not yet written a test. Once the reviewers have agreed to the approach I have used I will write a test.

The approach I have used is to generate DebugValue instructions without removing the associated DebugDeclare. I did not want to remove the DebugDeclare since this pass will always leave at least one Store to the variable.  So this pass will leave SPIR-V such that there are both DebugDeclares and DebugValues for the same variable.

This is not expressly ruled out in any spec I could find, although it probably isn't ideal for tools. The good news is that this is only done when it is quite certain that a subsequent pass of ssa-rewrite will clean up the DebugDeclare. Specifically, we will only replace a Store with a DebugValue for scalar variables.

For additional consideration, this pass was written solely to improve the performance of ssa-rewrite, so we can generally expect a pass of ssa-rewrite to follow. It does for all the optimization sequences currently used by glslang and dxc.

Also for consideration: ssa-rewrite and other analyses that remove stores should be able to deal with this hybrid situation without difficulty, just concentrating on the DebugDeclare and Stores and ignoring any associated DebugValues.

greg-lunarg:
@jaebaek Please review

source/opt/local_single_block_elim_pass.cpp
```diff
@@ -73,11 +90,21 @@ bool LocalSingleBlockLoadStoreElimPass::LocalSingleBlockLoadStoreElim(
...
-            // for deletion if not still used.
+            // for deletion if not still used. Don't delete store
+            // for aggregate if debugging; the DebugValue will complicate
+            // scalar replacement.
             auto prev_store = var2store_.find(varId);
             if (prev_store != var2store_.end() &&
-                instructions_to_save.count(prev_store->second) == 0) {
+                instructions_to_save.count(prev_store->second) == 0 &&
+                !(IsArrayOrStruct(varId) &&
+                  get_debug_info_mgr()->IsDebugDeclared(varId))) {
```
jaebaek, line 100: In my opinion, we should allow delete store for this case. Consider the following example:
```
%empty_expr = OpExtInst %void %ext DebugExpression
...
%foo = OpVariable %ptr_Function_int Function
...
OpStore %foo %int_13  // will be deleted
OpStore %foo %int_17
%decl = OpExtInst %void %ext DebugDeclare %dbg_foo %foo %empty_expr
```
We have to delete `OpStore %foo %int_13`.

The following case looks non-trivial:
```
%empty_expr = OpExtInst %void %ext DebugExpression
...
%foo = OpVariable %ptr_Function_int Function
...
OpStore %foo %int_13  // will be deleted
%decl = OpExtInst %void %ext DebugDeclare %dbg_foo %foo %empty_expr
...
OpStore %foo %int_17
```

IMO, we have to change it to
```
%empty_expr = OpExtInst %void %ext DebugExpression
...
%foo = OpVariable %ptr_Function_int Function
...
%val = OpExtInst %void %ext DebugValue %dbg_foo %int_13 %empty_expr
...
OpStore %foo %int_17
%decl = OpExtInst %void %ext DebugDeclare %dbg_foo %foo %empty_expr
```

What do you think?

s-perron: I think we should just not delete the store for any variable (scalars as well) that have a debug declare.  If DebugDeclare is suppose to match llvm.dbg.declare, the address has to always contains the correct value.  However, the spir-v extension is not as clear as the llvm-ir spec, this might be fine.  

I think that as long as we still remove the loads, this pass will still do what it needs to do.

s-perron:
Thanks for this Greg.  This looks good in general.  The problem I am having with all of this debug stuff is how imprecise the specification of the debug extension is.

source/opt/local_single_block_elim_pass.cpp
greg-lunarg: For clarification, do you believe that we should not remove the store even if we replace it with a DebugValue, which is being done here?

More generally, can we allow DebugDeclare and DebugValue to coexist for the same DebugLocalVariable? I think we can temporarily within spirv-opt as long as we ultimately remove the DebugDeclares before outputting the SPIR-V.

s-perron: I don't think we should remove the store.  The DebugValue is not needed, but I don't think it hurts.

> can we allow DebugDeclare and DebugValue to coexist for the same DebugLocalVariable?

The spec is so unclear, I am not sure what we can or cannot do.  I believe it would be okay to have both as long as they give the same information.

>  I think we can temporarily within spirv-opt as long as we ultimately remove the DebugDeclares before outputting the SPIR-V.

If we could guarantee that users always run ssa-rewrite then I could agree with this: temporarily invalid code is fine.  However, our passes are meant to be used in potentially any order.  We have tried to ensure that no pass requires another pass to run in order to be correct.  The only exception so far is merge-return and dead-branch elimination.  In that case an error is emitted (https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/merge_return_pass.cpp#L108).

jaebaek: Ah ok, my comment was wrong.
[llvm-dbg-declare](https://llvm.org/docs/SourceLevelDebugging.html#llvm-dbg-declare) explains "It is not control-dependent ... that address is considered to be the true home of the variable across its entire lifetime."

We must not delete the store if the variable is used for `DebugDeclare`.

greg-lunarg: Not sure why you think we should not remove the OpStore.

Do we agree that for a scalar local (function scope) OpVariable with a DebugLocalVariable and a DebugDeclare, it is ok , and actually a good thing (a goal for spirv-opt) to replace all its OpStores with DebugValues, remove the DebugDeclare and the actual OpVariable, but leave the DebugLocalVariable in place?

What I would like to establish is what we want the legalization passes to do and not do in the presense of OpenCL.DebugInfo.100.

Can/should the legalization passes do less optimization in the presense of OpenCL.DebugInfo.100? For example, can/should the passes not eliminate OpStores? I think it is possible to eliminate OpStores as long as we replace them with DebugValues.

greg-lunarg: I think it might be possible to legalize without removing OpStores. I will have to go back to our original legalization shaders and see. I just thought, to be safe, we should try to stick to the original passes as closely as possible but still retain the debug information.

greg-lunarg: I am asking because I am looking at DCE right now which eliminates a lot of OpStores. If you think that is bad, it would be good to know now.

s-perron: > Not sure why you think we should not remove the OpStore.

If you look up the spec for llvm.dbg.declar (which Jaebaek referenced above), it clearly says that if you have an llvm.dbg.declare for a local variable, then that is the address for the variable for its entire lifetime.  If we remove a store, while leaving a debug declare, then that will have been violated.  That is why I don't think we can remove the store.

s-perron: > Do we agree that for a scalar local (function scope) OpVariable with a DebugLocalVariable and a DebugDeclare, it is ok , and actually a good thing (a goal for spirv-opt) to replace all its OpStores with DebugValues, remove the DebugDeclare and the actual OpVariable, but leave the DebugLocalVariable in place?

Yes, I agree.  The ssa-rewriter should be able to do that, and it will get done.  The ssa-rewritter will replace the DebugDeclare with all of the apporpriate DebugValue instruction, and then we can remove the stores.  If the DebugDeclare is still present, we cannot remove the stores.

s-perron: > Can/should the legalization passes do less optimization in the presense of OpenCL.DebugInfo.100? For example, can/should the passes not eliminate OpStores? I think it is possible to eliminate OpStores as long as we replace them with DebugValues.

If the OpVariable is used by a DebugDeclare or a DebugValue then we should not remove the store unless you can also remove the use in the DebugDeclare or DebugValue.  ssa-rewrite if removing the DebugDeclare and DebugValue, so later passes like ADCE can remove the stores.  We can still remove the loads at any time, and do the copy propagation etc.

s-perron: > I think it might be possible to legalize without removing OpStores. I will have to go back to our original legalization shaders and see. I just thought, to be safe, we should try to stick to the original passes as closely as possible but still retain the debug information.

I believe so as well, as long at the loads are removed.  That is what we really wanted from elim-local-single-block and elim-local-single-store anyway.

s-perron: > I am asking because I am looking at DCE right now which eliminates a lot of OpStores. If you think that is bad, it would be good to know now.

As I said before DCE should not remove the store if the OpVariable is used in a DebugDeclare or DebugValue.  The ssa rewriter should remove those uses in cases where we can remove the stores, so I don't think it will hinder optimization.

greg-lunarg: I think it is ok to replace a single OpStore with a DebugValue as long as we eventually replace them all and get rid of the DebugDeclare before outputting the SPIR-V.

I agree with the general philosophy that each pass should be able to stand alone, but elim-local-single-block is not really a stand-alone pass; it was written to be a helper to ssa-rewrite, so I think it is ok to assume that ssa-rewrite will follow.

That said, I don't think its a huge deal if we don't eliminate OpStores in elim-local-single-block. I was just trying to create as little new logic and exceptions and different behavior for DebugInfo as possible in the interest of reducing complexity.

greg-lunarg: OK. Just looked at output from ssa-rewrite. Now I understand your approach better.

ssa-rewrite is adding DebugValue after each Store and removing the DebugDeclare. So essentially ssa-rewrite is preparing the OpStores and OpVariable for deletion.

I can work with this model now that I understand it.

greg-lunarg:
I have changed this to only avoiding deleting a store in the case that the variable is has a DebugDeclare, deferring to ssa-rewrite to take care of it. I have also added a test and thus have removed the WIP.

source/opt/local_single_block_elim_pass.cpp
```diff
@@ -31,6 +31,10 @@ const uint32_t kStoreValIdInIdx = 1;
 bool LocalSingleBlockLoadStoreElimPass::HasOnlySupportedRefs(uint32_t ptrId) {
   if (supported_ref_ptrs_.find(ptrId) != supported_ref_ptrs_.end()) return true;
   if (get_def_use_mgr()->WhileEachUser(ptrId, [this](Instruction* user) {
+        auto dbg_op = user->GetOpenCL100DebugOpcode();
+        if (dbg_op == OpenCLDebugInfo100DebugDeclare) {
```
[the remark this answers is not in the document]
s-perron: @greg-lunarg I believe that jaebaek is right.  You should look for a DebugValue here as well.  Scalar replacement will be replacing DebugDeclare instructions with DebugValue instruction as shown above.  This would still be a supported ref.

test/opt/local_single_block_elim.cpp
```diff
@@ -1118,6 +1118,182 @@ TEST_F(LocalSingleBlockLoadStoreElimTest, VariablePointerTest) {
   )";
   SinglePassRunAndMatch<LocalSingleBlockLoadStoreElimPass>(text, false);
 }
+
+TEST_F(LocalSingleBlockLoadStoreElimTest, DebugInfoTest) {
```
s-perron, line 1122: This is a good test.  We should add another one where a DebugValue with a dref expression is used in place of the DebugDeclare.  We would expect the same resutls.

greg-lunarg: I admit that the OpenCL.DebugInfo.100 spec is vague, but after some research and thought, I do not believe that a DebugValue with a deref of a pointer Value is the same as a DebugDeclare. A DebugValue represents an assignment of the Value and Expression to the **rvalue** of the LocalVariable. A DebugDeclare represents an "equivalencing" of the Expression applied to lvalue of the Variable to the **lvalue** of the LocalVariable. I am fairly certain these are not the same thing.

I am also fairly certain that we will never have cause to have a deref in an Expression in a DebugValue in a shader. The Variable would have to be of type pointer into function scope memory which don't exist in shaders.

greg-lunarg: After some thought, while I am still fairly certain that one cannot use a DebugValue to create an equivalent to a DebugDeclare in a shader, I will grant that one could create a legitimate DebugValue that contains a deref of a local pointer for a shader, However, I think that such a reference would be so highly unlikely that it would not be worth creating the logic to add it to the supported references for this optimization. If such a thing did appear, it would merely disable optimization and not generate incorrect code. So I think the existing logic is reasonable and acceptable at this time. If someone can make an argument for why such a construct would be prevalent enough to justify optimization here, we can reconsider this logic at that time.

s-perron: > A DebugValue represents an assignment of the Value and Expression to the rvalue of the LocalVariable.

I have thought about this a lot as well when trying to figure out the design for scalar replacement.  I understand your point of view.  At first, I was thinking that way as well.  However, I started thinking about how this would actually be implemented by a debugger, and in dwarf.  In DWARF, you don't give the value of a variable.  DWARF say where you go to get the value (register or memory).  The important point to realize is that the DebugValue is not an assignment of a value.  This can be easily confused when rerefs are not used because, in llvm-ir and in spir-v, the ssa value cannot be changed.

The other reason that I believed the llvm.dbg.value could not be viewed as an assignment of a value is the extra work this would put on debuggers, and how it would slow down debugging.  The debugging info is not present in the machine code.  If the debugger was required to capture the value of the expression at every llvm.dbg.value, the debugger would have to stop the normal execution of the program at every position of an llvm.dbg.value to read the value.  This would cause a ridiculous amount of overhead in debugging.  We wouldn't want a spir-v debugger to have to do the same.

Granted this was just guesses, which is why I ask Jaebaek to look at what code actually gets generated by llvm-ir when a debug value is used with a dref expression, and to examine what a real debugger does with it.  We hand wrote llvm-ir that has exactly one llvm.dbg.value where the address of the memory location is the value and there is a deref expression.  The instruction was placed after the first store to the memory location.  There was a second store to the same memory location, but there was not a second llvm.dbg.value.  When running in the debugger, you see the value of the first store when you break after the first store.  However, when you break after the second store, you see the result of the second store.  The debugger did not capture the value of the first store to be used anytime the developer asks for the value of the variable.  The debugger checks the value in the memory location.

greg-lunarg: OK. Thanks for the detailed response. I do now agree that a DebugDeclare can be replaced with a DebugValue with a deref, and I apologize that it took me this long to come to this understanding.

So I guess now the question in my mind is: is this really what we want to do?

Essentially, to do this, we have to keep backing memory for all locals (function scope variables) and keep all Stores to that memory that we normally remove. So we will have to add extra logic to all optimizations that eliminate stores. In one way it will be simpler for the optimizer since it only has to generate one DebugValue at the top of the program, but there will be performance cost from the extra Stores.

This design will also require a whole additional and redundant memory analysis of derefs in DebugValues.

The next question is: do we need to do this? You opined above that DebugValue would be intractable for debuggers as I imagined them. But if you look at https://llvm.org/docs/SourceLevelDebugging.html#object-lifetime-in-optimized-code it appears that llvm.dbg.value is being used in just this manner.

As far as debugger implementation, I imagine the debugger would be able to create a fairly simple map for each local variable, mapping a line number range to a register.

Still, your design will functionally work, albeit with probable performance costs, and I can work with it.

greg-lunarg: I think I can just look at the Value field of the DebugValue and see if its a pointer, so that should be fairly easy.

jaebaek: Your questions are very important questions to remove the technical debt and make a reasonable design. I appreciate it. I am conducting more experiments to figure out the relationship between `llvm.dbg.declare` and `llvm.dbg.value` particularly with `DW_OP_LLVM_fragment`. I am also checking `DW_OP_bit_piece`, which is similar to `DW_OP_LLVM_fragment`.

IIRC, we decided to use `DebugValue` with `Deref` to express the declaration because we thought it is the only way to express the declaration in the scalar replacement. In LLVM world, SROA conducts both the scalar replacement and ssa-rewrite in a single pass, so it does not have to consider "splitting a composite type while keeping the information of declaration". On the other hand, spirv-opt scalar replacement must "split a composite type while keeping the information of declaration". Since `DebugDeclare` does not have `Indexes` operand while `DebugValue` does, we had no idea how to use `DebugDeclare` for each member of a composite. See [DebugDeclare](https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html#DebugDeclare).

The main questions I currently have are
- **Q1**: Does LLVM SROA pass propagate `llvm.dbg.value` for store and phi instructions when we give `llvm.dbg.value` with `DW_OP_deref`?
- **Q2**: Can we use `DW_OP_LLVM_fragment` with `llvm.dbg.declare` for the scalar replacement?
- **Q3**: Is there a way to express a member of a composite using `DebugDeclare` and `Deref` operation?

If **Q2** and **Q3** are true, I think we do not have to use `DebugValue` with `Deref` for the declaration in the scalar replacement.
In addition, if **Q2** and **Q3** are true and **Q1** is false, it would be better to follow what LLVM does i.e., we must not use `DebugValue` with `Deref` as a declaration.

In my brief experiments, **Q1** and **Q2** are true, but I am doublechecking it if I made some mistakes. I am also digging into LLVM source code.
As far as I understand, `DW_OP_LLVM_fragment` is applied to the variable not its value of `llvm.dbg.value` nor its storage of `llvm.dbg.declare`.
I am curious what `DW_OP_bit_piece` means because I guess (I do not have any evidence yet) that it is applied to the value of `llvm.dbg.value` or the storage of `llvm.dbg.declare`.

For **Q3**, I am discussing the spec change with the initial author of the spec.

greg-lunarg:
@s-perron I think have provided all of your requested changes. Is there anything else that is needed?

jaebaek:
@greg-lunarg Steven is on vacation and he will not work for a while. I think this PR is ready to be submitted.

I will open a separate bug if we have to update the part related to `DebugValue`.

s-perron:
This looks good to me.  Thanks.

PR #3456, opened 2020-06-22
Support OpCompositeExtract pattern in desc_sroa

ehsannas:


ehsannas:
shaderc-smoketest failure seems related to glslang, not this CL.

source/opt/desc_sroa.cpp
```diff
@@ -114,7 +117,13 @@ bool DescriptorScalarReplacement::ReplaceCandidate(Instruction* var) {
...
           case SpvOpAccessChain:
           case SpvOpInBoundsAccessChain:
-            work_list.push_back(use);
+            access_chain_work_list.push_back(use);
+            return true;
+          case SpvOpLoad:
+            load_work_list.push_back(use);
+            return true;
+          case SpvOpCompositeExtract:
+            extract_work_list.push_back(use);
```
s-perron, line 126: Can this happen?  The value would have to be loaded first, and then extracted right?

source/opt/desc_sroa.cpp
```diff
@@ -360,5 +388,66 @@ uint32_t DescriptorScalarReplacement::GetNumBindingsUsedByType(
...
+}
+
+bool DescriptorScalarReplacement::ReplaceCompositeExtract(
+    Instruction* var, Instruction* extract) {
+  assert(extract->opcode() == SpvOpCompositeExtract);
+  if (extract->NumInOperands() != 2) {
+    context()->EmitErrorMessage(
+        "Variable cannot be replaced: invalid instruction", extract);
+    return false;
+  }
```
s-perron, line 433: Add a comment saying this it is possible to handle this case if we need to, but we won't for now.

source/opt/desc_sroa.cpp
```diff
@@ -360,5 +388,66 @@ uint32_t DescriptorScalarReplacement::GetNumBindingsUsedByType(
...
+      GetReplacementVariable(var, extract->GetSingleWordInOperand(1));
+
+  // The result type of the OpLoad is the same as the result type of the
+  // OpCompositeExtract.
+  uint32_t load_id = TakeNextId();
+  std::unique_ptr<Instruction> load(
+      new Instruction(context(), SpvOpLoad, extract->type_id(), load_id,
+                      std::initializer_list<Operand>{
+                          {SPV_OPERAND_TYPE_ID, {replacement_var}}}));
+  get_def_use_mgr()->AnalyzeInstDefUse(load.get());
```
s-perron, line 445: You'll need to update the inst-to-block mapping as well.  See IrContext::set_inst_block.

source/opt/desc_sroa.cpp
ehsannas: oh yes that's true

source/opt/desc_sroa.cpp
ehsannas: Done

source/opt/desc_sroa.cpp
ehsannas: Done

PR #3459, opened 2020-06-23
Clear debug information for kill and replacement

jaebaek:
For many spirv-opt passes such as simplify-instructions pass, we have to
correctly clear the OpenCL.DebugInfo.100 debug information for
KillInst() and ReplaceAllUses(). If we keep some debug information that
disappeared because of KillInst() and ReplaceAllUses(), adding new
DebugValue instructions will generate incorrect information. This CL
update DebugInfoManager and IRContext to correctly clear debug
information.

### DIFF
--- a/source/opt/debug_info_manager.h
+++ b/source/opt/debug_info_manager.h
@@ -15,8 +15,8 @@
 #ifndef SOURCE_OPT_DEBUG_INFO_MANAGER_H_
 #define SOURCE_OPT_DEBUG_INFO_MANAGER_H_
 
+#include <list>
 #include <unordered_map>
-#include <vector>
 
 #include "source/opt/instruction.h"
 #include "source/opt/module.h"
@@ -138,6 +138,12 @@ class DebugInfoManager {
   // Erases |instr| from data structures of this class.
   void ClearDebugInfo(Instruction* instr);
 
+  // Runs the given function |f| on each DebugDeclare or DebugValue
+  // instruction whose variable or value is |var_or_value_id|.
+  void ForEachDebugDeclareOrValue(
+      uint32_t var_or_value_id,
+      const std::function<void(Instruction*)>& f) const;
+
  private:
   IRContext* context() { return context_; }
 
@@ -157,17 +163,18 @@ class DebugInfoManager {
   // in |inst| must not already be registered.
   void RegisterDbgFunction(Instruction* inst);
 
-  // Register the DebugDeclare instruction |dbg_declare| into
-  // |var_id_to_dbg_decl_| using OpVariable id |var_id| as a key.
-  void RegisterDbgDeclare(uint32_t var_id, Instruction* dbg_declare);
+  // Register the DebugDeclare or DebugValue instruction |dbg_inst| into
+  // |var_id_to_dbg_decl_| using |id| as a key.
+  void RegisterDbgDeclareOrValue(uint32_t var_or_value_id,
+                                 Instruction* dbg_inst);
 
   // Returns a DebugExpression instruction without Operation operands.
   Instruction* GetEmptyDebugExpression();
 
-  // Returns the id of Value operand if |inst| is DebugValue who has Deref
-  // operation and its Value operand is a result id of OpVariable with
-  // Function storage class. Otherwise, returns 0.
-  uint32_t GetVariableIdOfDebugValueUsedForDeclare(Instruction* inst);
+  // Returns true if |inst| is DebugValue who has Deref operation and
+  // its Value operand is a result id of OpVariable with Function
+  // storage class. Otherwise, returns 0.
+  bool IsDebugValueUsedForDeclare(Instruction* inst);
 
   // Returns true if a scope |ancestor| is |scope| or an ancestor scope
   // of |scope|.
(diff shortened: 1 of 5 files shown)


jaebaek:
@greg-lunarg I think you need this PR for the debug information preservation in ADCE. Could you please review the PR if it has some issues with ADCE?

source/opt/debug_info_manager.h
```diff
@@ -157,17 +163,18 @@ class DebugInfoManager {
   // in |inst| must not already be registered.
   void RegisterDbgFunction(Instruction* inst);
 
-  // Register the DebugDeclare instruction |dbg_declare| into
-  // |var_id_to_dbg_decl_| using OpVariable id |var_id| as a key.
-  void RegisterDbgDeclare(uint32_t var_id, Instruction* dbg_declare);
+  // Register the DebugDeclare or DebugValue instruction |dbg_inst| into
+  // |var_id_to_dbg_decl_| using |id| as a key.
```
s-perron, line 167: |id| is not a variable.  I think you meant |var_or_value_id|.

source/opt/debug_info_manager.h
```diff
@@ -191,9 +198,9 @@ class DebugInfoManager {
   // operand is the function.
   std::unordered_map<uint32_t, Instruction*> fn_id_to_dbg_fn_;
 
-  // Mapping from local variable ids to DebugDeclare instructions whose
-  // operand is the local variable.
-  std::unordered_map<uint32_t, std::vector<Instruction*>> var_id_to_dbg_decl_;
+  // Mapping from variable or value ids to DebugDeclare or DebugValue
+  // instructions whose operand is the variable or value.
+  std::unordered_map<uint32_t, std::list<Instruction*>> var_id_to_dbg_decl_;
```
s-perron, line 203: Why did you change to a list?

source/opt/ir_context.cpp
```diff
@@ -424,10 +430,12 @@ void IRContext::KillOperandFromDebugInstructions(Instruction* inst) {
...
-  // Notice that we do not need anythings to do for local variables.
-  // DebugLocalVariable does not have an OpVariable operand. Instead,
-  // DebugDeclare/DebugValue has an OpVariable operand for a local
-  // variable. The function inlining pass handles it properly.
+  // Kill DebugValue or DebugDeclare whose operand is |id|.
+  std::vector<Instruction*> dbg_decl_or_value;
+  get_debug_info_mgr()->ForEachDebugDeclareOrValue(
+      id,
+      [&dbg_decl_or_value](Instruction* i) { dbg_decl_or_value.push_back(i); });
+  for (auto* i : dbg_decl_or_value) KillInst(i);
```
s-perron, line 438: I'm not sure about this.  If the DebugDeclare or DebugValue are still in the code when an instruction is killed, then it is likely that the optimization is wrong.  The value should have been replaced or the debug instruction explicitly removed.

s-perron: Also, don't add any type of assert.  We want to allow the passes to delete instructions in any order the want.  An error will be picked by by the validator it the pass does not delete the instruction.

source/opt/debug_info_manager.h
s-perron: unorder_set is fine.  Avoid std::list, it is known to have poor implementations.  In general, if you know things will be small, you can never go wrong with std::vector.  It has so much less overhead than the other containers, that in practice it can still be faster.

>  std::vector::erase calls the destructor and the constructor O(N) times

A pointer does not have a constructor or destructor, so it will not be called.

test/opt/local_ssa_elim_test.cpp
```diff
@@ -2185,13 +2185,13 @@ TEST_F(LocalSSAElimTest, DebugValueForReferenceVariable) {
...
+; CHECK: OpExtInst %void [[ext:%\d+]] DebugLocalVariable [[f_name]]
 ; CHECK: [[dbg_i:%\w+]] = OpExtInst %void [[ext]] DebugLocalVariable [[i_name]]
-; CHECK: [[dbg_x:%\w+]] = OpExtInst %void [[ext]] DebugLocalVariable [[x_name]]
+; CHECK: OpExtInst %void [[ext]] DebugLocalVariable [[x_name]]
 
 ; CHECK:      OpStore %f %float_0
-; CHECK-NEXT: OpExtInst %void [[ext]] DebugValue [[dbg_f]] %float_0
-; CHECK-NEXT: OpExtInst %void [[ext]] DebugValue [[dbg_x]] %float_0
+; CHECK-NEXT: OpExtInst %void [[ext]] DebugValue {{%\w+}} %float_0
+; CHECK-NEXT: OpExtInst %void [[ext]] DebugValue {{%\w+}} %float_0
```
s-perron, line 2194: Why are you not checking the local variables anymore?  If the issue is order, you can use CHECK-DAG.

jaebaek: Thanks. CHECK-DAG is what I needed.

source/opt/debug_info_manager.h
jaebaek: I see. Thank you for the explanation!

PR #3461, opened 2020-06-25
Preserve debug info in scalar replacement pass

jaebaek:
Preserve OpenCL.DebugInfo.100 and line information in scalar
replacement pass.

1. Set the debug scope and line information for the new replacement
   instructions.
2. Replace DebugDeclare and DebugValue if their OpVariable or value
   operands are replaced by scalars.
  - If the initial aggregate has DebugValue, we clone a DebugValue
    for each scalar and add one more index for the scalar to 'Indexes'
    operand of the cloned DebugValue as we add an index to 'Indexes'
    operand of OpAccessChain.
  - If the initial aggregate has DebugDeclare, we create a DebugValue
    with Deref operation for each scalar and add an index for the scalar
    to 'Indexes' operand.
  - For example,
```
   struct S { int a; int b;}
   S foo; // before scalar replacement

   int foo_a; // after scalar replacement
   int foo_b;

   DebugDeclare %dbg_foo %foo %null_expr // before

   DebugValue %dbg_foo %foo_a %Deref_expr 0 // after
   DebugValue %dbg_foo %foo_b %Deref_expr 1 // means Value(foo.members[1]) == Deref(%foo_b)
```
Note that the DebugValue instructions for the SROA'd structure
elements are linked to the original debug info for `S foo` via the
`%dbg_foo`. Each DebugValue has Deref and 'Indexes' i.e., 0
for `%foo_a` and 1 for `%foo_b` in the above example.
`DebugValue %dbg_foo %foo_a %Deref_expr 0` means that
`Deref(%foo_a)` is the value of `0`th member of `%dbg_foo`.

### DIFF
--- a/source/opt/constants.h
+++ b/source/opt/constants.h
@@ -630,7 +630,7 @@ class ConstantManager {
   // Returns the id of a 32-bit floating point constant with value |val|.
   uint32_t GetFloatConst(float val);
 
-  // Returns the id of a 32-bit singed integer constant with value |val|.
+  // Returns the id of a 32-bit signed integer constant with value |val|.
   uint32_t GetSIntConst(int32_t val);
 
  private:
--- a/source/opt/constants.h
+++ b/source/opt/constants.h
@@ -630,6 +630,9 @@ class ConstantManager {
   // Returns the id of a 32-bit floating point constant with value |val|.
   uint32_t GetFloatConst(float val);
 
+  // Returns the id of a 32-bit singed integer constant with value |val|.
+  uint32_t GetSIntConst(int32_t val);
+
  private:
   // Creates a Constant instance with the given type and a vector of constant
   // defining words. Returns a unique pointer to the created Constant instance
--- a/source/opt/debug_info_manager.cpp
+++ b/source/opt/debug_info_manager.cpp
@@ -448,6 +448,42 @@ bool DebugInfoManager::IsDeclareVisibleToInstr(Instruction* dbg_declare,
   return IsAncestorOfScope(instr_scope_id, decl_scope_id);
 }
 
+void DebugInfoManager::AddDebugValueWithIndex(uint32_t dbg_local_var_id,
+                                              uint32_t value_id,
+                                              uint32_t expr_id,
+                                              uint32_t index_id,
+                                              Instruction* insert_before) {
+  std::unique_ptr<Instruction> new_dbg_value(new Instruction(
+      context(), SpvOpExtInst, context()->get_type_mgr()->GetVoidTypeId(),
+      context()->TakeNextId(),
+      {
+          {spv_operand_type_t::SPV_OPERAND_TYPE_ID,
+           {context()
+                ->get_feature_mgr()
+                ->GetExtInstImportId_OpenCL100DebugInfo()}},
+          {spv_operand_type_t::SPV_OPERAND_TYPE_EXTENSION_INSTRUCTION_NUMBER,
+           {static_cast<uint32_t>(OpenCLDebugInfo100DebugValue)}},
+          {spv_operand_type_t::SPV_OPERAND_TYPE_ID, {dbg_local_var_id}},
+          {spv_operand_type_t::SPV_OPERAND_TYPE_ID, {value_id}},
+          {spv_operand_type_t::SPV_OPERAND_TYPE_ID,
+           {expr_id == 0 ? GetEmptyDebugExpression()->result_id() : expr_id}},
+      }));
+  if (index_id) {
+    new_dbg_value->AddOperand(
+        {spv_operand_type_t::SPV_OPERAND_TYPE_ID, {index_id}});
+  }
+
+  Instruction* added_dbg_value =
+      insert_before->InsertBefore(std::move(new_dbg_value));
+  if (context()->AreAnalysesValid(IRContext::Analysis::kAnalysisDefUse))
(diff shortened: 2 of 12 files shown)


jaebaek:
FYI, @greg-lunarg This is the PR for the scalar replacement.

jaebaek:
We currently discuss some SPIR-V spec issues related to this PR. I will update this PR after that.

source/opt/constants.h
```diff
@@ -630,6 +630,9 @@ class ConstantManager {
   // Returns the id of a 32-bit floating point constant with value |val|.
   uint32_t GetFloatConst(float val);
 
+  // Returns the id of a 32-bit singed integer constant with value |val|.
```
dnovillo, line 633: s/singed/signed/

source/opt/debug_info_manager.cpp
```diff
@@ -446,6 +446,42 @@ bool DebugInfoManager::IsDeclareVisibleToInstr(Instruction* dbg_declare,
   return IsAncestorOfScope(instr_scope_id, decl_scope_id);
 }
 
+void DebugInfoManager::AddDebugValueWithIndex(uint32_t dbg_local_var_id,
+                                              uint32_t value_id,
+                                              uint32_t expr_id,
+                                              uint32_t index_id,
+                                              Instruction* insert_before) {
+  std::unique_ptr<Instruction> new_dbg_value(new Instruction(
```
dnovillo, line 454: Why doesn't this re-use `AddDebugValue`?  Shouldn't it need to check the same things that `AddDebugValue` checks?  And now in #3511, `AddDebugValue` returns a boolean flag indicating success.  It seems odd that this would not be also expected in this function.

source/opt/constants.h
jaebaek: done

source/opt/debug_info_manager.cpp
jaebaek: `AddDebugValueWithIndex()` simply generates and inserts a single `DebugValue` while `AddDebugValue()` does the same thing for the multiple `DebugValue`. I updated `AddDebugValue()` to use `AddDebugValueWithIndex()`.

In terms of the boolean return of `AddDebugValue()`, it does not matter that much of this PR. It will follow #3511 when I merge it.

dnovillo:
> Preserve OpenCL.DebugInfo.100 and line information in scalar
> replacement pass.
> 
> 1. Set the debug scope and line information for the new replacement
>    instructions.
> 2. Replace DebugDeclare and DebugValue if their OpVariable or value
>    operands are replaced by scalars. It uses 'Indexes' operand of
>    DebugValue. For example,
>    struct S { int a; int b;}
>    S foo; // before scalar replacement
>    int foo_a; // after scalar replacement
>    int foo_b;
>    DebugDeclare %dbg_foo %foo %null_expr // before
>    DebugValue %dbg_foo %foo_a %Deref_expr 0 // after
>    DebugValue %dbg_foo %foo_b %Deref_expr 1 // means Value(foo.members[1]) == Deref(%foo_b)

Add some commentary on why we have to replace `DebugDeclare` with `DebugValue`?  My understanding from reading https://www.khronos.org/registry/spir-v/specs/unified1/OpenCL.DebugInfo.100.html#DebugDeclare is used to link debug information to a local declaration.  When we scalarize the variable, this declaration disappears because the symbol `foo` disappears.  What is not completely clear in this description, is how will a debugger request for `foo.a` be linked to the `dbg_foo` debug symbol.

I would add something along the lines of "the `DebugValue` expressions for the SROA'd structure elements are linked to the original debug info for `foo` via the `dbg_foo` symbol.  Each `DebugValue` will have instructions on what field is being accessed via the `Deref_*` expressions."

Does that sound like I'm understanding the difference between the two?

source/opt/debug_info_manager.cpp
dnovillo: OK, so now I have a question about `AddDebugValue()`.  It seems to be a low-level primitive, and yet it may (under some conditions) not actually add a `DebugValue()`.  This is confusing, as it would mean that a low-level helper decides whether or not to do something (despite the name of the function).  Perhaps rename the function to reflect this?  Something like `MaybeAdd...` or (better) make the caller responsible for calling `AddDebugValue` when it really means to add a value.

jaebaek: I agree that these names are confusing. I just update `AddDebugValue()` to `AddDebugValueIfVarDeclIsVisible()`.

source/opt/scalar_replacement_pass.cpp
```diff
@@ -144,6 +162,55 @@ Pass::Status ScalarReplacementPass::ReplaceVariable(
...
+
+  // Add DebugValue instruction with Indexes operand and Deref operation.
+  int32_t idx = 0;
+  for (const auto* var : replacements) {
+    Instruction* added_dbg_value =
+        context()->get_debug_info_mgr()->AddDebugValueWithIndex(
+            dbg_decl->GetSingleWordOperand(
+                kDebugDeclareOperandLocalVariableIndex),
+            var->result_id(), deref_expr->result_id(),
+            context()->get_constant_mgr()->GetSIntConst(idx), var->NextNode());
```
s-perron, line 181: This is really hard to ready.  Each parameter is long, and it is not clear what each one is.  Could you find a way to make this clearer?  Maybe create some local variables that will be passed in the parameters.

jaebaek: done

PR #3463, opened 2020-06-25
Permit Simple and GLSL450 memory model in WEBGPU_0

dneto0:
Now that WebGPU ingests WGSL instead of SPIR-V,
there is no need to be so strict about the memory model.

Allow any memory model that is already allowed by Vulkan 1.0,
either directly or via an existing.


dneto0:
Rebasing to see if the smoketest bot works.

PR #3470, opened 2020-06-26
Fix ADCE pass bug for mulitple entries

jaebaek:
When there are multiple entries and the shader has a variable with
WorkGroup storage class, those multiple entry functions store values to
the variable. Since ADCE pass uses def-use chains to propagate the work
list, some of instructions in the work list are not actually a part of
the currently processed function. As a result, it adds instructions in
other functions and put them in `live_insts_`. However, it does not
have the control flow information for those instructions in other
functions i.e., `block2headerBranch_` and `header2nextHeaderBranch_`.
It skips handling those instructions because they are already in
`live_insts_` and does not check their control flow information e.g.,
`block2headerBranch_` and `header2nextHeaderBranch_`, which results
in skipping some branches. Even though those branches are live
branches, it considers they are dead branches.

It happens because multiple functions can store values to a variable with
WorkGroup storage class and the code handling store instructions add
them into `live_insts_` without considering they are in a function currently
processed. This commit lets it add an instruction to `live_insts_` only
when the function is currently processed one.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -105,13 +105,17 @@ bool AggressiveDCEPass::IsLocalVar(uint32_t varId) {
          IsVarOfStorage(varId, SpvStorageClassWorkgroup);
 }
 
-void AggressiveDCEPass::AddStores(uint32_t ptrId) {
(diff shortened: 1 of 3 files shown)


jaebaek:
This issue is initially reported in DXC github repo https://github.com/microsoft/DirectXShaderCompiler/issues/2828

jaebaek:
Yeah. Please doublecheck it. I am not 100% sure I understand all the relations between global variables.

PR #3485, opened 2020-07-01
spirv-fuzz: Add variables with workgroup storage class

### LINKED ISSUE #3274: spirv-fuzz: Add variables with workgroup storage class
The fuzzer currently has a pass for adding global variables that only adds Private storage class variables.

It could be enhanced to be able to add Workgroup storage class variables if the shader being transformed is a compute shader.

andreperezmaselco:
Fixes #3274.

### DIFF
--- a/source/fuzz/fuzzer_pass_add_global_variables.cpp
+++ b/source/fuzz/fuzzer_pass_add_global_variables.cpp
@@ -35,9 +35,11 @@ void FuzzerPassAddGlobalVariables::Apply() {
     // If the execution model of some entry point is GLCompute,
     // then the variable storage class may be Workgroup.
     if (entry_point.GetSingleWordInOperand(0) == SpvExecutionModelGLCompute) {
-      variable_storage_class = GetFuzzerContext()->ChooseEven()
-                                   ? SpvStorageClassPrivate
-                                   : SpvStorageClassWorkgroup;
+      variable_storage_class =
+          GetFuzzerContext()->ChoosePercentage(
+              GetFuzzerContext()->GetChanceOfChoosingWorkgroupStorageClass())
+              ? SpvStorageClassWorkgroup
+              : SpvStorageClassPrivate;
       break;
     }
   }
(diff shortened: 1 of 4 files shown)


source/fuzz/fuzzer_pass_add_global_variables.cpp
```diff
@@ -30,8 +30,20 @@ FuzzerPassAddGlobalVariables::FuzzerPassAddGlobalVariables(
 FuzzerPassAddGlobalVariables::~FuzzerPassAddGlobalVariables() = default;
 
 void FuzzerPassAddGlobalVariables::Apply() {
+  SpvStorageClass variable_storage_class = SpvStorageClassPrivate;
+  for (auto& entry_point : GetIRContext()->module()->entry_points()) {
+    // If the execution model of some entry point is GLCompute,
+    // then the variable storage class may be Workgroup.
+    if (entry_point.GetSingleWordInOperand(0) == SpvExecutionModelGLCompute) {
+      variable_storage_class = GetFuzzerContext()->ChooseEven()
```
afd, line 38: Instead of hard-coding this as 50:50, could you add a constant to FuzzerContext for the probability of choosing workgroup storage class for a variable?  I'm fine with it being 50:50 by default, but I'd like to give it a name so that we can control it later.

source/fuzz/fuzzer_pass_add_global_variables.cpp
```diff
@@ -30,8 +30,20 @@ FuzzerPassAddGlobalVariables::FuzzerPassAddGlobalVariables(
...
+                                   ? SpvStorageClassPrivate
+                                   : SpvStorageClassWorkgroup;
+      break;
```
afd, line 41: Nit: I prefer "return" since there's no more work to do.

source/fuzz/fuzzer_pass_add_global_variables.cpp
andreperezmaselco: Done.

Issue #3132, opened 2020-01-08
spirv-fuzz: Interchange zero-like constants


afd:
OpConstantNull is in at least some cases equivalent to zero.  An interesting transformation would be to change zeros to null, or vice versa.

paulthomson:
@stefanomil Please reply and I will officially assign this to you :-)

stefanomil:
Sure, thanks!

afd:
@stefanomil We are done with this one, correct?  (I guess we didn't include "Fixes ..." in the final PR related to it.)

stefanomil:
Yes

PR #3486, opened 2020-07-01
Implement transformation to interchange zero-like constants

stefanomil:
#3132 
I have implemented the TransformationToggleConstantNull transformation, tests and fuzzer pass, which transforms zero-valued constants of type int, float or bool into null constants of the same type (according to the specification, a null constant of these types is equivalent to a constant with value zero - false for booleans).

Is the code generally ok? Are there other cases that I should have considered?
I chose the upper and lower bounds for the probability of applying the transformation to be 20 and 90 a bit randomly.
I am also not sure if the name is explicative enough.


source/fuzz/fuzzer_pass_toggle_constant_null.cpp
```diff
@@ -0,0 +1,59 @@
...
+    if (!transformation.IsApplicable(GetIRContext(),
+                                     *GetTransformationContext())) {
+      continue;
+    }
+
+    // Choose randomly whether to apply the transformation
+    if (!GetFuzzerContext()->ChoosePercentage(
+            GetFuzzerContext()->GetChanceOfTogglingConstantNull())) {
+      continue;
+    }
```
afd, line 53: My overall feedback is that we should do this PR a different way.

However, before I realised that I had written this (which is sound advice in general even if it no longer applies here.)

It's best to do this check earlier - this is a super-cheap check, and there's no point making the transformation and checking whether it is applicable if we may probabilistically decide not to do the transformation.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1314,6 +1315,15 @@ message TransformationToggleAccessChainInstruction {
 
 }
 
+message TransformationToggleNullConstant {
+
+  // A transformation that replaces a null boolean, int or floating-point null constant (OpConstantNull) with
+  // a zero-valued constant of the same type (OpConstant), and vice versa
+
+  // The result id of the constant declaration instruction
+  uint32 constant_id = 1;
```
afd, line 1324: Again, this is sort of obsolete given my overall comment on the PR, but is perhaps worth digesting anyway.

I think it would be better for us to do this on a per-use basis.

For example, if %a is OpConstantFalse of type bool, and %b is OpConstantNull of type bool, and if the module contains many uses of %a then it should be possible to transform just one of those uses to %b.

To do this, use an IdUseDescriptor.

afd:
In spirv-fuzz there is a "fact manager" that tracks facts about the module being transformed.

One kind of fact is called a "data synonym" fact, and it records that two pieces of data are known to be equal.  In the simplest case - which suffices for this discussion - it tracks that two ids are equal.

There is a transformation called TransformationReplaceIdWithSynonym that takes some use of an id %a, and an id %b that is synonymous with %a, and replaces the use of %a with a use of %b.

For example, there might be an instruction:

%c = OpIAdd %int %a %a ; This is like c = a + a

If the fact manager knows that %a and %b are synonymous (where %b is some other id) then TransformationReplaceIdWithSynonym could say "replace the first use of %a in '%c = OpIAdd %int %a %a' with %b".

That would turn the above instruction into:

%c = OpIAdd %int %b %a ; This is like c = b + a

and this is semantics-preserving if the fact that %a and %b are synonymous really is true.

Now, what does this have to do with interchanging zero-like constants?

Well: we could write a fuzzer pass that does the following:

- For each zero-like constant %c
    - Add an equivalent, but toggled, zero-like constant %d to the module*
    - Record the fact that %c and %d are synonymous**
    - For each use of %c
          - Probabilistically decide whether to leave %c alone or replace %c with %d***

* This will involve using TransformationAddConstantNull, TransformationAddConstantBoolean, or similar

** This will need a new transformation, TransformationRecordSynonymousConstants, which should be applicable if two given constants really are equivalent and should have the effect of adding the fact that they are synonymous to the fact manager - it should not actually change the module

*** This will require applying a TransformationReplaceIdWithSynonym

I suggest you take a look at FuzzerPassApplyIdSynonyms to see some related code.

Let me know if this makes sense - keen to discuss.

stefanomil:
I have implemented the fuzzer pass and the transformation as described.

For each zero-like constant %c, the fuzzer pass looks for an existing toggled instruction and adds a new one only if it doesn't find it, otherwise it uses the already-existing one (there were some functions in the fuzzer_pass file that I could reuse to do that).

I was unsure whether the IsApplicable function of the transformation class should return true with two constants that are equal (same opcode and same operands). I opted for also returning true in this case (even though this is not used by the pass) because in my mind it seemed to make more sense this way.

Let me know about any observations or corrections to make.

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  // If the constants are equal, then they are equivalent
+  if (constant == synonym) {
+    return true;
+  }
+
+  // The types must be the same
+  if (!constant->type()->IsSame(synonym->type())) {
+    return false;
+  }
```
afd, line 65: A refinement to this: if the types are both integer, with the same width, then we don't care if their signedness is different.  In that case we'd want to check that their associated data words are the same.

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -0,0 +1,97 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "transformation_record_synonymous_constants.h"
+
+namespace spvtools {
+namespace fuzz {
+
+namespace {
+static inline bool IsStaticZeroConstant(
```
afd, line 21: Remove `static` - it's redundant since you're in an anonymous namespace.

Remove `inline` - the code isn't performance critical.

Change name to `IsScalarZeroConstant`.

stefanomil:
I have opened a new pull request (#3494 ) that just introduces the transformation class and tests, so I am closing this one.

PR #3492, opened 2020-07-06
Preserve OpenCL.100.DebugInfo in reduce-load-size pass

jaebaek:
The decision to reduce the load must be not affected by debug
instructions. For example, even when a DebugValue references a
result id of a loaded composite value, this change lets the
reduce-load-size pass reduce the load if the full composite value is not
used anywhere other than the DebugValue.

### DIFF
--- a/source/opt/reduce_load_size.cpp
+++ b/source/opt/reduce_load_size.cpp
@@ -112,10 +112,10 @@ bool ReduceLoadSize::ReplaceExtract(Instruction* inst) {
   Instruction* new_access_chain = ir_builder.AddAccessChain(
       pointer_to_result_type_id,
       composite_inst->GetSingleWordInOperand(kLoadPointerInIdx), ids);
-  Instruction* new_laod =
+  Instruction* new_load =
       ir_builder.AddLoad(inst->type_id(), new_access_chain->result_id());
 
-  context()->ReplaceAllUsesWith(inst->result_id(), new_laod->result_id());
+  context()->ReplaceAllUsesWith(inst->result_id(), new_load->result_id());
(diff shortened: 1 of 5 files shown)


jaebaek:
@greg-lunarg I cannot add you as a code reviewer, but please give me your feedback if you find something weird for this PR. Thanks!

source/opt/reduce_load_size.cpp
```diff
@@ -139,6 +139,8 @@ bool ReduceLoadSize::ShouldReplaceExtract(Instruction* inst) {
 
   all_elements_used =
       !def_use_mgr->WhileEachUser(op_inst, [&elements_used](Instruction* use) {
+        if (use->GetOpenCL100DebugOpcode() != OpenCLDebugInfo100InstructionsMax)
+          return true;
```
dnovillo, line 143: I think this predicate will be used quite a bit.  During traversal, we may need to skip debug instructions or ignore them for reasons similar to this.  Perhaps factor this into a IsDebugInstr() predicate?

jaebaek: Do you mean adding some code similar to the following function in a common header like `source/opt/reflect.h`
```
bool IsDebugInstr(Instruction *instr) {
  return instr->GetOpenCL100DebugOpcode() != OpenCLDebugInfo100InstructionsMax;
}
```
?

dnovillo: Yes, though not as a free function.  As a member of `class Instruction`.

jaebaek: done

PR #3494, opened 2020-07-07
Implement transformation to record synonymous constants.

stefanomil:
#3132 
This pull request only includes the code related to TransformationRecordSynonymousConstants and related tests (after feedback for #3486 , with changes based on the feedback from code review).


source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -337,7 +337,7 @@ message Transformation {
     TransformationAddTypeInt add_type_int = 7;
     TransformationAddDeadBreak add_dead_break = 8;
     TransformationReplaceBooleanConstantWithConstantBinary
-      replace_boolean_constant_with_constant_binary = 9;
+        replace_boolean_constant_with_constant_binary = 9;
```
afd, line 340: Please avoid unnecessary whitespace changes.

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -0,0 +1,104 @@
...
+          constant2->type()->AsInteger()->width() &&
+      constant1->AsIntConstant()->words() ==
+          constant2->AsIntConstant()->words()) {
+    return true;
+  }
+
+    // The types must be the same
+    if (!constant1->type()->IsSame(constant2->type())) {
+      return false;
+    }
```
afd, line 72: Indentation looks wrong.

test/fuzz/transformation_record_synonymous_constants_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+  spvtools::ValidatorOptions validator_options;
+  TransformationContext transformation_context(&fact_manager,
+                                               validator_options);
+  ASSERT_TRUE(IsValid(env, context.get()));
+
+  // %3 is not a constant declaration
+  ASSERT_FALSE(TransformationRecordSynonymousConstants(3, 9).IsApplicable(
+      context.get(), transformation_context));
+
+  ASSERT_FALSE(TransformationRecordSynonymousConstants(9, 3).IsApplicable(
```
afd, line 81: It's nice to have a brief comment before each assert false explaining why it should fail (like you've done with the others).

test/fuzz/transformation_record_synonymous_constants_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+  // The two constants must be different
+  ASSERT_FALSE(TransformationRecordSynonymousConstants(9, 9).IsApplicable(
+      context.get(), transformation_context));
+
+  // %9 and %13 are not equivalent
+  ASSERT_FALSE(TransformationRecordSynonymousConstants(9, 13).IsApplicable(
+      context.get(), transformation_context));
+
+  ASSERT_FALSE(TransformationRecordSynonymousConstants(13, 9).IsApplicable(
```
afd, line 92: Similar - nice to have a comment.

test/fuzz/transformation_record_synonymous_constants_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+      context.get(), transformation_context));
+
+  // %11 and %13 are not equivalent
+  ASSERT_FALSE(TransformationRecordSynonymousConstants(11, 13).IsApplicable(
+      context.get(), transformation_context));
+
+  ASSERT_FALSE(TransformationRecordSynonymousConstants(13, 11).IsApplicable(
```
afd, line 99: Similar (and please add comments for any further cases).

test/fuzz/transformation_record_synonymous_constants_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+  // %9 and %11 are equivalent
+  ASSERT_TRUE(TransformationRecordSynonymousConstants(9, 11).IsApplicable(
+      context.get(), transformation_context));
+
+  ASSERT_TRUE(TransformationRecordSynonymousConstants(11, 9).IsApplicable(
+      context.get(), transformation_context));
+
+  // Apply transformation
+  TransformationRecordSynonymousConstants(9, 11).Apply(context.get(),
```
afd, line 122: Please apply all the applicable cases so that we get facts due to:

- null and zero for integer
- two constants of the same integer type
- two constants of the same integer type but with different type

test/fuzz/transformation_record_synonymous_constants_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+  ASSERT_TRUE(TransformationRecordSynonymousConstants(9, 20).IsApplicable(
+      context.get(), transformation_context));
+
+  // Apply transformation
+  TransformationRecordSynonymousConstants(9, 20).Apply(context.get(),
+                                                       &transformation_context);
+
+  // Check that the ids are now synonyms
+  transformation_context.GetFactManager()->IsSynonymous(
+      MakeDataDescriptor(9, {}), MakeDataDescriptor(20, {}));
```
afd, line 195: Can you enhance the test to check that:

- if there are multiple OpConstantTrues they can be regarded as synonymous
- Same for multiple OpConstantFalses?

test/fuzz/transformation_record_synonymous_constants_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+  ASSERT_TRUE(TransformationRecordSynonymousConstants(9, 11).IsApplicable(
+      context.get(), transformation_context));
+
+  // Apply transformation
+  TransformationRecordSynonymousConstants(9, 11).Apply(context.get(),
+                                                       &transformation_context);
+
+  // Check that the ids are now synonyms
+  transformation_context.GetFactManager()->IsSynonymous(
+      MakeDataDescriptor(9, {}), MakeDataDescriptor(11, {}));
```
afd, line 281: Please add a case to show that a fact is added when we apply the transformation to two identical float constants.

afd:
@stefanomil The CI has failed due to formatting.  Please search for "Code should be formatted with `clang-format`." in README.md for instructions on how to get that set up.

stefanomil:
I have implemented the suggested changes. In the test file, since the transformation is applied many times, I wrote a function to apply the transformation and check the results. Let me know if that is ok or if there are any downsides in terms of readability or other.

source/fuzz/protobufs/spvtoolsfuzz.proto
stefanomil: Fixed

source/fuzz/transformation_record_synonymous_constants.cpp
stefanomil: Fixed

test/fuzz/transformation_record_synonymous_constants_test.cpp
stefanomil: Done

test/fuzz/transformation_record_synonymous_constants_test.cpp
stefanomil: Done

test/fuzz/transformation_record_synonymous_constants_test.cpp
stefanomil: Done

test/fuzz/transformation_record_synonymous_constants_test.cpp
stefanomil: Done. Since the transformation is applied many times, I wrote a function to apply the transformation and check the results. Let me know if that is ok or if there are any downsides in terms of readability or other.

test/fuzz/transformation_record_synonymous_constants_test.cpp
stefanomil: Done

test/fuzz/transformation_record_synonymous_constants_test.cpp
stefanomil: Done

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) 2020 Stefano Milizia
+// Modifications Copyright (c) 2020 Google LLC
```
afd, line 2: I think this will fail the project's formatting checks, and also you typically just list all the copyright holders - you don't tend to say "Modifications".

stefanomil: Thanks, I was unsure about how to do this so I looked in the repository and found this way. Should I change it? Do I list all the copyright holders just by putting a comma between them?

stefanomil: (for example in source/operand.cpp)

stefanomil: Fixed

afd:
@Vasniktel please see the transformation added here.  We will need to adapt this so that - when we have IdIsIrrelevant facts - we do not allow facts that make constants synonymous where either one is irrelevant.

Vasniktel:
Thanks, I'll take a look.

PR #3497, opened 2020-07-07
Preserve debug info in vector DCE pass

jaebaek:
This commit lets the vector DCE pass preserve the OpenCL.DebugInfo.100
information properly. When the vector DCE pass determines the liveness
of instructions, the debug instructions must not affect the decision. In
addition, when it kills some instructions, it has to kill DebugValue
instructions that use the killed instructions. When it updates some
composite values to meaningful values (not undef), it has to remove
DebugValue because the value information becomes incorrect.

### DIFF
--- a/source/opt/vector_dce.cpp
+++ b/source/opt/vector_dce.cpp
@@ -36,7 +36,6 @@ Pass::Status VectorDCE::Process() {
 
 bool VectorDCE::VectorDCEFunction(Function* function) {
   LiveComponentMap live_components;
-  dead_dbg_value_.clear();
   FindLiveComponents(function, &live_components);
   return RewriteInstructions(function, live_components);
 }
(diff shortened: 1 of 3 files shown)


jaebaek:
@greg-lunarg I cannot add you as a code reviewer, but please give me your feedback if you find something weird for this PR. Thanks!

source/opt/vector_dce.cpp
```diff
@@ -52,6 +53,10 @@ void VectorDCE::FindLiveComponents(Function* function,
   // components are live because of arbitrary nesting of structs.
   function->ForEachInst(
       [&work_list, this, live_components](Instruction* current_inst) {
+        if (current_inst->GetOpenCL100DebugOpcode() !=
+            OpenCLDebugInfo100InstructionsMax) {
+          return;
```
dnovillo, line 58: Similar comment to #3492.  Add a predicate member to class `Instruction`.  Maybe `Instruction::IsDebug()`?

jaebaek: Let me update this code after submitting #3492.

jaebaek: done

source/opt/vector_dce.cpp
```diff
@@ -377,6 +384,14 @@ bool VectorDCE::RewriteInsertInstruction(
   return false;
 }
 
+void VectorDCE::MarkDebugValueUsesAsDead(Instruction* composite) {
+  context()->get_def_use_mgr()->ForEachUser(
+      composite, [this](Instruction* use) {
+        if (use->GetOpenCL100DebugOpcode() == OpenCLDebugInfo100DebugValue)
```
dnovillo, line 390: Change to `if (use->IsOpenCL100DebugInstr())`?

jaebaek: No. I want to check if this is exactly `DebugValue` not other debug instructions.

jaebaek:
I replaced the class member variable `std::vector<Instruction*> dead_dbg_value_` to `std::vector<Instruction*> dead_dbg_value`.
I followed how it uses `live_components` and `work_list`, which are not class member variables but just local variables.

source/opt/vector_dce.cpp
dnovillo: Ah, yes.

jaebaek:
We iterate them in `function->ForEachInst( ... )` or `ForEachUse`.
If we directly remove an instruction in the middle of the instruction iteration loop, it will have a dangling pointer.
The only purpose of it is to avoid dangling pointers.
Let me add a comment for it.

PR #3498, opened 2020-07-07
Preserve OpenCL.DebugInfo.100 through elim-local-single-store

greg-lunarg:
This pass basically follows the same process as ssa-rewrite: it adds a DebugValue after each Store and removes the DebugDeclare or DebugValue Deref. It only does this if all instructions that are dependent on the Store are Loads and are replaced.

### DIFF
--- a/source/opt/local_single_store_elim_pass.cpp
+++ b/source/opt/local_single_store_elim_pass.cpp
@@ -133,7 +133,27 @@ bool LocalSingleStoreElimPass::ProcessVariable(Instruction* var_inst) {
     return false;
   }
 
-  return RewriteLoads(store_inst, users);
+  bool all_rewritten;
+  bool modified = RewriteLoads(store_inst, users, &all_rewritten);
+
+  // If all uses are rewritten and the variable has a DebugDeclare and the
+  // variable is not an aggregate, add a DebugValue after the store and remove
+  // the DebugDeclare.
+  uint32_t var_id = var_inst->result_id();
+  if (all_rewritten &&
+      context()->get_debug_info_mgr()->IsDebugDeclared(var_id)) {
+    const analysis::Type* var_type =
+        context()->get_type_mgr()->GetType(var_inst->type_id());
+    const analysis::Type* store_type = var_type->AsPointer()->pointee_type();
+    if (!(store_type->AsStruct() || store_type->AsArray())) {
+      context()->get_debug_info_mgr()->AddDebugValue(
+          store_inst, var_id, store_inst->GetSingleWordInOperand(1),
+          store_inst);
+      context()->get_debug_info_mgr()->KillDebugDeclares(var_id);
+    }
+  }
+
+  return modified;
 }
 
 Instruction* LocalSingleStoreElimPass::FindSingleStoreAndCheckUses(
@@ -172,6 +192,14 @@ Instruction* LocalSingleStoreElimPass::FindSingleStoreAndCheckUses(
       case SpvOpName:
       case SpvOpCopyObject:
         break;
+      case SpvOpExtInst: {
+        auto dbg_op = user->GetOpenCL100DebugOpcode();
+        if (dbg_op == OpenCLDebugInfo100DebugDeclare ||
+            dbg_op == OpenCLDebugInfo100DebugValue) {
+          break;
+        }
+        return nullptr;
+      }
       default:
         if (!user->IsDecoration()) {
           // Don't know if this instruction modifies the variable.
@@ -218,7 +246,8 @@ bool LocalSingleStoreElimPass::FeedsAStore(Instruction* inst) const {
 }
 
 bool LocalSingleStoreElimPass::RewriteLoads(
-    Instruction* store_inst, const std::vector<Instruction*>& uses) {
+    Instruction* store_inst, const std::vector<Instruction*>& uses,
+    bool* all_rewritten) {
   BasicBlock* store_block = context()->get_instr_block(store_inst);
   DominatorAnalysis* dominator_analysis =
       context()->GetDominatorAnalysis(store_block->GetParent());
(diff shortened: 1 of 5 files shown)


greg-lunarg:
@jaebaek @s-perron Please review.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -133,7 +133,27 @@ bool LocalSingleStoreElimPass::ProcessVariable(Instruction* var_inst) {
...
+    const analysis::Type* var_type =
+        context()->get_type_mgr()->GetType(var_inst->type_id());
+    const analysis::Type* store_type = var_type->AsPointer()->pointee_type();
+    if (!(store_type->AsStruct() || store_type->AsArray())) {
+      context()->get_debug_info_mgr()->AddDebugValue(
+          store_inst, var_id, store_inst->GetSingleWordInOperand(1),
+          store_inst);
+      context()->get_debug_info_mgr()->KillDebugDeclares(var_id);
+    }
+  }
```
jaebaek, line 154: If I understand it correctly, it works because the `var_inst` at this point has only a single `OpStore`.

### tl;dr
The following looks correct for me
```
if (the declare is DebugValue with Deref that does not dominate store_inst) return;
else
   // the logic you wrote
```

### Reasoning
I am sorry for bothering you but I am still considering how we must handle `DebugValue` with `Deref`.
I opened a bug #3504. I will update the ssa-rewrite pass for it to consider the dominance relation when we use `DebugValue` with `Deref` for the declaration.

For example, we have to consider the dominance relation for the following case:
```
%x = alloca i32
store i32 13, %x   // debugger says %dbg_x = uninitialized
store i32 7, %x   // debugger says %dbg_x = uninitialized
llvm.dbg.value(%x, %dbg_x, DW_OP_deref)
store i32 2, %x   // debugger says %dbg_x = 2  --> because this store is dominated by the above llvm.dbg.value
store i32 4, %x   // debugger says %dbg_x = 4
```
When we remove `llvm.dbg.value` with `DW_OP_deref`, we have to propagate `llvm.dbg.value` for the `store i32 2` and `store i32 4`, but not for the `store i32 13` and `store i32 7`.
This is mainly because `llvm.dbg.value` is control-dependent while `llvm.dbg.declare` is not.

Consider the following case
```
%foo = OpVariable %ptr_Function_int
...
OpStore %foo %int_13
// at this point, the compiler must print "undefined" for %dbg_foo
...
DebugValue %dbg_foo %foo Deref    // from here, the compiler must print the value of %dbg_foo as Deref(%foo)
```

I think we have to remove the declare and add `DebugValue` **only when** the declare is `DebugDeclare` or it is **`DebugValue` with `Deref` that dominates the `store_inst`**.

jaebaek: Briefly,
```
auto *dbg_decl = context()->get_debug_info_mgr()->GetDebugDeclare(var_id);
if (dbg_decl == nullptr) return;

if (dbg_decl->GetOpenCL100DebugOpCode() != DebugDeclare) {
  BasicBlock* dbg_decl_block = context()->get_instr_block(dbg_decl);
  DominatorAnalysis* dominator_analysis =
      context()->GetDominatorAnalysis(dbg_decl_block->GetParent());

  if (!dominator_analysis->Dominates(dbg_decl, store_inst)) return;
}

// the code you wrote
    const analysis::Type* var_type =
        context()->get_type_mgr()->GetType(var_inst->type_id());
    const analysis::Type* store_type = var_type->AsPointer()->pointee_type();
    if (!(store_type->AsStruct() || store_type->AsArray())) {
      context()->get_debug_info_mgr()->AddDebugValue(
          store_inst, var_id, store_inst->GetSingleWordInOperand(1),
          store_inst);
      context()->get_debug_info_mgr()->KillDebugDeclares(var_id);
    }
```

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -133,7 +133,27 @@ bool LocalSingleStoreElimPass::ProcessVariable(Instruction* var_inst) {
...
+  // If all uses are rewritten and the variable has a DebugDeclare and the
+  // variable is not an aggregate, add a DebugValue after the store and remove
+  // the DebugDeclare.
+  uint32_t var_id = var_inst->result_id();
+  if (all_rewritten &&
+      context()->get_debug_info_mgr()->IsDebugDeclared(var_id)) {
+    const analysis::Type* var_type =
+        context()->get_type_mgr()->GetType(var_inst->type_id());
+    const analysis::Type* store_type = var_type->AsPointer()->pointee_type();
+    if (!(store_type->AsStruct() || store_type->AsArray())) {
```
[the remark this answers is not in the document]
greg-lunarg: Based partly on a concern from @s-perron who thought we should not have DebugValue of aggregate (array or struct) type. The result of this logic is that arrays and structs will stay in memory and keep their DebugDeclare. I think this is a completely acceptable result.

Really, with loop unrolling and scalar renaming, we should almost never see aggregate types at this point. If we do, as I said, I think this is a completely acceptable outcome.

source/opt/local_single_store_elim_pass.cpp
```diff
@@ -172,6 +192,14 @@ Instruction* LocalSingleStoreElimPass::FindSingleStoreAndCheckUses(
       case SpvOpName:
       case SpvOpCopyObject:
         break;
+      case SpvOpExtInst: {
+        auto dbg_op = user->GetOpenCL100DebugOpcode();
+        if (dbg_op == OpenCLDebugInfo100DebugDeclare ||
+            dbg_op == OpenCLDebugInfo100DebugValue) {
+          break;
+        }
```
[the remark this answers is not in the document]
greg-lunarg: Because we have a validation pass that is run by default at the beginning of spirv-opt, we can and do assume that every SPIR-V instruction is valid. This keeps our code from being cluttered and slowed down with repetitive validity checks. In light of this, could we agree this assert is not necessary? Or if it is necessary, a better place for it is in the validation pass?

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: I think I have a pretty good reason that we don't have to check for dominance. I will write up my thinking tomorrow morning.

source/opt/local_single_store_elim_pass.cpp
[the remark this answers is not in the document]
jaebaek: Ah I see. Never mind it.

source/opt/local_single_store_elim_pass.cpp
greg-lunarg: Yes, you can hand-write SPIR-V such that a DebugValue/Deref instruction does not dominate a Store to the associated variable, but will it ever happen for valid and useful transformations applied to a valid shader generated for a GLSL or HLSL shader?

Asked another way: is there a legitimate and useful reason that a DebugValue/Deref instruction would not dominate the full live range of its associated variable?

I would propose that the answer is no.

For local variables in shader languages, there is a 1-1 correspondence between a variable and its memory. So there is no reason for a DebugValue/Deref to not dominate all Stores of its associated variable.

I am presuming that any pass that generates a DebugValue/Deref will generate it so that it dominates all its associated Stores and no legitimate subsequent pass will break that. So I don't see a reason to check for that, in the same way we do not need to check for validity of any other code.

greg-lunarg: In summary, I posit that it is not necessary to check for dominance, here or in ssa-rewrite.

greg-lunarg: Adding dominance checks would greatly and needlessly increase the complexity of these transformations.

jaebaek: Ok. I agree with you that we would not have the case. Let's revisit it when we have an actual example.
Thank you for reviewing my opinion!

PR #3511, opened 2020-07-10
Remove DebugDeclare only for target variables in ssa-rewrite

jaebaek:
For each local variable, ssa-rewrite should remove its DebugDeclare
if and only if it is replaced by any number of DebugValues for store
and phi instructions.

For example, when we have two variables `a` whose DebugDeclare
will be replaced to DebugValues by ssa-rewrite pass and `b` whose
DebugDeclare will not be replaced, we have to remove only DebugDeclare
for `a`, not `b`.

### DIFF
--- a/source/opt/debug_info_manager.cpp
+++ b/source/opt/debug_info_manager.cpp
@@ -487,12 +487,13 @@ Instruction* DebugInfoManager::AddDebugValueWithIndex(
   return added_dbg_value;
 }
 
-void DebugInfoManager::AddDebugValueIfVarDeclIsVisible(
+bool DebugInfoManager::AddDebugValueIfVarDeclIsVisible(
     Instruction* scope_and_line, uint32_t variable_id, uint32_t value_id,
     Instruction* insert_pos) {
   auto dbg_decl_itr = var_id_to_dbg_decl_.find(variable_id);
-  if (dbg_decl_itr == var_id_to_dbg_decl_.end()) return;
+  if (dbg_decl_itr == var_id_to_dbg_decl_.end()) return false;
 
+  bool added = false;
   uint32_t instr_scope_id = scope_and_line->GetDebugScope().GetLexicalScope();
   for (auto* dbg_decl_or_val : dbg_decl_itr->second) {
     if (!IsDeclareVisibleToInstr(dbg_decl_or_val, instr_scope_id)) continue;
@@ -517,7 +518,9 @@ void DebugInfoManager::AddDebugValueIfVarDeclIsVisible(
                                value_id, 0, index_id, insert_before);
     assert(added_dbg_value != nullptr);
     added_dbg_value->UpdateDebugInfoFrom(scope_and_line);
+    added = true;
   }
+  return added;
 }
 
 uint32_t DebugInfoManager::GetVariableIdOfDebugValueUsedForDeclare(
--- a/source/opt/debug_info_manager.cpp
+++ b/source/opt/debug_info_manager.cpp
@@ -487,13 +487,12 @@ Instruction* DebugInfoManager::AddDebugValueWithIndex(
   return added_dbg_value;
 }
 
-bool DebugInfoManager::AddDebugValueIfVarDeclIsVisible(
+void DebugInfoManager::AddDebugValueIfVarDeclIsVisible(
     Instruction* scope_and_line, uint32_t variable_id, uint32_t value_id,
     Instruction* insert_pos) {
   auto dbg_decl_itr = var_id_to_dbg_decl_.find(variable_id);
-  if (dbg_decl_itr == var_id_to_dbg_decl_.end()) return false;
+  if (dbg_decl_itr == var_id_to_dbg_decl_.end()) return;
 
-  bool added = false;
   uint32_t instr_scope_id = scope_and_line->GetDebugScope().GetLexicalScope();
   for (auto* dbg_decl_or_val : dbg_decl_itr->second) {
     if (!IsDeclareVisibleToInstr(dbg_decl_or_val, instr_scope_id)) continue;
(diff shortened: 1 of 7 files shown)


jaebaek:
I missed this case in #3356.

Basically, the existing ssa-rewrite pass removes all DebugDeclare even when it should not.
If some DebugDeclare instructions are not propagated as DebugValue for store and phi instructions, we must not remove them.
For example,
```
 #version 140
 
 in vec4 BC;
 out float fo;
 
 void main()
 {
     float f = 0.0; 
     for (int j=0; j<4; j++) {
       // BB_inside
       int &i = j; 
       f = f + BC[i];
     }    
     fo = f; 
 }

 %f = OpVariable %_ptr_Function_float Function
 %j = OpVariable %_ptr_Function_int Function
 OpStore %f %float_0
 OpStore %j %int_0
 %decl0 = OpExtInst %void %ext DebugDeclare %dbg_f %f %null_expr
 %decl1 = OpExtInst %void %ext DebugDeclare %dbg_i %j %null_expr   ; %dbg_i == "int &i"
```

If we use `%j` for `int &i` in `BB_inside`, we must not add a `DebugValue` for the `j++`.
In this case, we must not remove `%decl1 = OpExtInst %void %ext DebugDeclare %dbg_i %j %null_expr   ; %dbg_i == "int &i"`.

source/opt/ssa_rewrite_pass.cpp
```diff
@@ -615,7 +628,17 @@ Pass::Status SSARewriter::RewriteFunctionIntoSSA(Function* fp) {
             << fp->PrettyPrint(0) << "\n";
 #endif
 
-  if (modified) pass_->context()->KillDebugDeclareInsts(fp);
+  if (modified) {
+    // Kill DebugDeclare only if it is added as DebugValue for each store
+    // or phi instruction and there is no store or phi instruction that it
+    // cannot be added as DebugValue because the out-of-scope issue.
```
dnovillo, line 634: Could you clarify this comment?  I don't understand what it's trying to state.

dnovillo: Fundamentally, I don't think I'm understanding why `DebugValue` would have to remove a `DebugDeclare`.  I'm missing something, but I'm not sure what.

source/opt/ssa_rewrite_pass.h
```diff
@@ -284,6 +284,14 @@ class SSARewriter {
   // Set of blocks that have been sealed already.
   std::unordered_set<BasicBlock*> sealed_blocks_;
 
+  // Set of OpVariable ids whose DebugValue have been added for store or
+  // phi instructions.
+  std::unordered_set<uint32_t> var_ids_added_dbg_value_;
+
+  // Set of OpVariable ids that tries to add DebugValue but failed because
+  // a store or phi instruction is not under its declaration scope.
```
dnovillo, line 292: Same here.  The comment does not really clarify anything for me.

source/opt/debug_info_manager.cpp
```diff
@@ -449,12 +449,13 @@ bool DebugInfoManager::IsDeclareVisibleToInstr(Instruction* dbg_declare,
   return IsAncestorOfScope(instr_scope_id, decl_scope_id);
 }
 
-void DebugInfoManager::AddDebugValue(Instruction* scope_and_line,
+bool DebugInfoManager::AddDebugValue(Instruction* scope_and_line,
                                      uint32_t variable_id, uint32_t value_id,
```
dnovillo, line 453: See my comment in #3461.  It's confusing that `AddDebugValue` doesn't always does what the caller meant (https://github.com/KhronosGroup/SPIRV-Tools/pull/3461#discussion_r456500705).

s-perron:
I cannot remember if we have talked about this or not.  This does not look right.  Replacing a debug declare should always be an all or nothing.  You can either replace the debug declare with a series of debug value instructions, or you do not add any debug value instructions.

More over, if ssa rewrite is able to rewrite the variable, then it should always be able to replace the debug declare with debug value instructions.  So the logic should be: "if ssa-rewrite chooses to rewrite the variable, then the DebugDeclare should be removed."  Nothing more complicated should be needed.

jaebaek:
@s-perron Thank you for your comment.

Let me clearly understand what the expected behavior is.
Let's assume that
- There are variables `a, b`
- We can replace DebugDeclare of `a` to DebugValue, but we cannot do that for `b`

In this case, what we should do among the following **E1** - **E4**?
- **E1**: Replace only DebugDeclare of `a` to DebugValue and remove DebugDeclare of `b`
- **E2**: Replace only DebugDeclare of `a` to DebugValue and keep DebugDeclare of `b`
- **E3**: do not replace any DebugDeclare to DebugValue and just keep both DebugDeclares of `a` and `b`
- **E4**: Replace DebugDeclare of `a` to DebugValue and replace DebugDeclare of `b` to DebugValue with Deref?
  - if `b` is already declared by DebugValue with Deref, do not update the DebugValue for `b`

s-perron:
E2: Replace only DebugDeclare of a to DebugValue and keep DebugDeclare of b

Sorry that I was not clear.  I was thinking of a single variable at a time, because logic of ssa-rewrite acts on each variable individually.

jaebaek:
Thank you for clarifying it. I updated the description of this PR and the code. PTAL.

source/opt/ssa_rewrite_pass.cpp
```diff
@@ -615,7 +622,11 @@ Pass::Status SSARewriter::RewriteFunctionIntoSSA(Function* fp) {
             << fp->PrettyPrint(0) << "\n";
 #endif
 
-  if (modified) pass_->context()->KillDebugDeclareInsts(fp);
+  if (modified) {
+    for (auto var_id : var_ids_debugdeclare_replaced_) {
```
s-perron, line 626: I believe that the target variables (those in `seen_target_vars_`) should be the ones you need to go through.  No need to keep a separate list.  If there is a debug declare for a target variable that is not removed, then I would expect that the be an error.

Could we try to use the target variable list here instead?

jaebaek: Thank you for the suggestion.

Initially, I concerned the following case but I agree with you that we can simple check if it is a target variable.
```
 int j; // global variable
 void main()
 {
     float f = 0.0;
     for (j=0; j<4; j++) {
       int& i = j; // I concerned DebugDeclare for i.
       f = f + BC[i];
     }
     fo = f;
 }
```

jaebaek:
I agree. Thank you for the suggestion. I updated the code to simply kill DebugDeclares for target variables. PTAL

PR #3517, opened 2020-07-11
spirv-fuzz: Add inline function transformation

### LINKED ISSUE #3505: Add inlining transformation to spirv-fuzz
Inlining is an interesting transformation to apply in spirv-fuzz.  For technical reasons it makes sense to implement our own simple inliner rather than to re-use the opt inliner.

Some technical details discussed with @andreperezmaselco:

FUZZING

Foo:
  %1 = some_instruction
  %2 = some_instruction

Main:
  %3 = OpFunctionCall Foo
  
----- Fuzzer pass add loads ----

Foo:
  %1 = some_instruction
  %4 = OpLoad ...
  %2 = some_instruction

Main:
  %3 = OpFunctionCall Foo

------ Inline foo -----

Foo:
  %1 = some_instruction
  %4 = OpLoad ...
  %2 = some_instruction

Main:
  %5 = some_instruction
  %6 = OpLoad
  %7 = some_instruction
  
------- Transform instruction %7 ----

Foo:
  %1 = some_instruction
  %4 = OpLoad ...
  %2 = some_instruction

Main:
  %5 = some_instruction
  %6 = OpLoad
  %7 = transformed(some_instruction)
  
  
  
  
SHRINKING

Foo:
  %1 = some_instruction
  %2 = some_instruction

Main:
  %3 = OpFunctionCall Foo
  
------ Inline foo -----

Foo:
  %1 = some_instruction
  %2 = some_instruction

Main:
  %5 = some_instruction
  %6 = some_instruction
  
------- Transform instruction %7 ----

NOT APPLICABLE - there is no instruction %7

ALTERNATIVE

FUZZING

Foo:
  %1 = some_instruction
  %2 = some_instruction

Main:
  %3 = OpFunctionCall Foo
  
----- Fuzzer pass add loads ----

Foo:
  %1 = some_instruction
  %4 = OpLoad ...
  %2 = some_instruction

Main:
  %3 = OpFunctionCall Foo

------ Inline foo: [%1 -> %5, %4 -> %6, %2 -> %7] -----

Foo:
  %1 = some_instruction
  %4 = OpLoad ...
  %2 = some_instruction

Main:
  %5 = some_instruction
  %6 = OpLoad
  %7 = some_instruction
  
------- Transform instruction %7 ----

Foo:
  %1 = some_instruction
  %4 = OpLoad ...
  %2 = some_instruction

Main:
  %5 = some_instruction
  %6 = OpLoad
  %7 = transformed(some_instruction)
  

SHRINKING

Foo:
  %1 = some_instruction
  %2 = some_instruction

Main:
  %3 = OpFunctionCall Foo
  
------ Inline foo: [%1 -> %5, %4 -> %6, %2 -> %7] -----

Foo:
  %1 = some_instruction
  %2 = some_instruction

Main:
  %5 = some_instruction
  %7 = some_instruction
  
------- Transform instruction %7 ----
IS applicable as %7 still exists

Foo:
  %1 = some_instruction
  %2 = some_instruction

Main:
  %5 = some_instruction
  %7 = transformed(some_instruction)

andreperezmaselco:
Fixes #3505.

### DIFF
--- a/source/fuzz/fuzzer_pass_inline_functions.cpp
+++ b/source/fuzz/fuzzer_pass_inline_functions.cpp
@@ -17,6 +17,7 @@
 #include "source/fuzz/fuzzer_util.h"
 #include "source/fuzz/instruction_descriptor.h"
 #include "source/fuzz/transformation_inline_function.h"
+#include "source/fuzz/transformation_split_block.h"
 
 namespace spvtools {
 namespace fuzz {
@@ -31,6 +32,9 @@ FuzzerPassInlineFunctions::FuzzerPassInlineFunctions(
 FuzzerPassInlineFunctions::~FuzzerPassInlineFunctions() = default;
 
 void FuzzerPassInlineFunctions::Apply() {
+  // |function_call_instructions| are the instructions that will be inlined.
+  // First, they will be collected and then do the inlining in another loop.
+  // This avoids changing the module while it is being inspected.
   std::vector<opt::Instruction*> function_call_instructions;
 
   for (auto& function : *GetIRContext()->module()) {
@@ -41,34 +45,21 @@ void FuzzerPassInlineFunctions::Apply() {
           continue;
         }
 
-        // |instruction| must be OpFunctionCall to consider applying the
-        // transformation.
-        if (instruction.opcode() != SpvOpFunctionCall) {
+        // |instruction| must be suitable for inlining.
+        if (!TransformationInlineFunction::IsSuitableForInlining(
+                GetIRContext(), &instruction)) {
           continue;
         }
 
-        // |called_function| must not have an early return.
-        auto called_function = fuzzerutil::FindFunction(
-            GetIRContext(), instruction.GetSingleWordInOperand(0));
-        if (called_function->HasEarlyReturn()) {
-          continue;
-        }
-
-        // Checks if some block in |called_function| has OpKill or OpUnreachable
-        // as its termination instruction.
-        bool hasKillOrUnreachableInstruction = false;
-        for (auto& called_function_block : *called_function) {
-          if (called_function_block.terminator()->opcode() == SpvOpKill ||
-              called_function_block.terminator()->opcode() ==
-                  SpvOpUnreachable) {
-            hasKillOrUnreachableInstruction = true;
-            break;
-          }
-        }
-
-        // |called_function| must not have a block with OpKill or OpUnreachable
-        // as its termination instruction.
-        if (hasKillOrUnreachableInstruction) {
+        // If |instruction| is not the penultimate instruction in |block| or
+        // |block| termination instruction is not OpBranch, then try to split
+        // |block| such that the conditions are met.
+        if ((&instruction != &*--block.tail() ||
+             block.terminator()->opcode() != SpvOpBranch) &&
+            !MaybeApplyTransformation(TransformationSplitBlock(
+                MakeInstructionDescriptor(GetIRContext(),
+                                          instruction.NextNode()),
+                GetFuzzerContext()->GetFreshId()))) {
           continue;
         }
 
@@ -77,6 +68,8 @@ void FuzzerPassInlineFunctions::Apply() {
     }
   }
 
+  // Once the function calls have been collected, it's time to actually create
+  // and apply the inlining transformations.
   for (auto& instruction : function_call_instructions) {
     auto called_function = fuzzerutil::FindFunction(
         GetIRContext(), instruction->GetSingleWordInOperand(0));
--- a/source/fuzz/fuzzer_pass_inline_functions.cpp
+++ b/source/fuzz/fuzzer_pass_inline_functions.cpp
@@ -37,8 +37,20 @@ void FuzzerPassInlineFunctions::Apply() {
       while (instruction != block.end()) {
         // |instruction| must be OpFunctionCall to consider applying the
         // transformation.
-        if (instruction->opcode() != SpvOpFunctionCall ||
-            !GetFuzzerContext()->ChoosePercentage(
+        if (instruction->opcode() != SpvOpFunctionCall) {
+          ++instruction;
+          continue;
+        }
+
+        // |called_function| must not have an early return.
+        auto called_function = fuzzerutil::FindFunction(
+            GetIRContext(), instruction->GetSingleWordInOperand(0));
+        if (called_function->HasEarlyReturn()) {
+          ++instruction;
+          continue;
+        }
+
+        if (!GetFuzzerContext()->ChoosePercentage(
                 GetFuzzerContext()->GetChanceOfInliningFunction())) {
           ++instruction;
           continue;
@@ -46,8 +58,6 @@ void FuzzerPassInlineFunctions::Apply() {
 
         // Mapping the called function instructions.
         std::map<uint32_t, uint32_t> result_id_map;
-        auto called_function = fuzzerutil::FindFunction(
-            GetIRContext(), instruction->GetSingleWordInOperand(0));
         for (auto& called_function_block : *called_function) {
           // The called function entry block label will not be inlined.
           if (&called_function_block != &*called_function->entry()) {
@@ -56,15 +66,10 @@ void FuzzerPassInlineFunctions::Apply() {
           }
 
           for (auto& instruction_to_inline : called_function_block) {
-            // If the |instruction_to_inline| result id is the returned value,
-            // then it will be mapped to the function call result id.
-            if (called_function->tail()->tail()->opcode() == SpvOpReturnValue &&
-                instruction_to_inline.HasResultId() &&
+            // The return value id will not be mapped.
+            if (instruction_to_inline.HasResultId() &&
                 instruction_to_inline.result_id() ==
-                    called_function->tail()->tail()->GetSingleWordInOperand(
-                        0)) {
-              result_id_map[instruction_to_inline.result_id()] =
-                  instruction->result_id();
+                    called_function->GetReturnValueId()) {
               continue;
             }
 
--- a/source/fuzz/fuzzer_pass_inline_functions.cpp
+++ b/source/fuzz/fuzzer_pass_inline_functions.cpp
@@ -31,64 +31,60 @@ FuzzerPassInlineFunctions::FuzzerPassInlineFunctions(
 FuzzerPassInlineFunctions::~FuzzerPassInlineFunctions() = default;
 
 void FuzzerPassInlineFunctions::Apply() {
+  std::vector<opt::Instruction*> function_call_instructions;
+
   for (auto& function : *GetIRContext()->module()) {
     for (auto& block : function) {
-      auto instruction = block.begin();
-      while (instruction != block.end()) {
+      for (auto& instruction : block) {
         // |instruction| must be OpFunctionCall to consider applying the
         // transformation.
-        if (instruction->opcode() != SpvOpFunctionCall) {
-          ++instruction;
+        if (instruction.opcode() != SpvOpFunctionCall) {
           continue;
         }
 
         // |called_function| must not have an early return.
         auto called_function = fuzzerutil::FindFunction(
-            GetIRContext(), instruction->GetSingleWordInOperand(0));
+            GetIRContext(), instruction.GetSingleWordInOperand(0));
         if (called_function->HasEarlyReturn()) {
-          ++instruction;
           continue;
         }
 
         if (!GetFuzzerContext()->ChoosePercentage(
                 GetFuzzerContext()->GetChanceOfInliningFunction())) {
-          ++instruction;
           continue;
         }
 
-        // Mapping the called function instructions.
-        std::map<uint32_t, uint32_t> result_id_map;
-        for (auto& called_function_block : *called_function) {
-          // The called function entry block label will not be inlined.
-          if (&called_function_block != &*called_function->entry()) {
-            result_id_map[called_function_block.GetLabelInst()->result_id()] =
-                GetFuzzerContext()->GetFreshId();
-          }
-
-          for (auto& instruction_to_inline : called_function_block) {
-            // The return value id will not be mapped.
-            if (instruction_to_inline.HasResultId() &&
-                instruction_to_inline.result_id() ==
-                    called_function->GetReturnValueId()) {
-              continue;
-            }
-
-            // The remaining instructions are mapped to fresh ids.
-            if (instruction_to_inline.HasResultId()) {
-              result_id_map[instruction_to_inline.result_id()] =
-                  GetFuzzerContext()->GetFreshId();
-            }
-          }
-        }
-
-        // Applies the inline function transformation.
-        ApplyTransformation(TransformationInlineFunction(
-            result_id_map, instruction->result_id()));
-
+        function_call_instructions.push_back(&instruction);
         // Erases the function call instruction from the caller function.
-        instruction = instruction.Erase();
       }
     }
+  }
+
+  for (auto& instruction : function_call_instructions) {
+    auto called_function = fuzzerutil::FindFunction(
+        GetIRContext(), instruction->GetSingleWordInOperand(0));
+
+    // Mapping the called function instructions.
+    std::map<uint32_t, uint32_t> result_id_map;
+    for (auto& called_function_block : *called_function) {
+      // The called function entry block label will not be inlined.
+      if (&called_function_block != &*called_function->entry()) {
+        result_id_map[called_function_block.GetLabelInst()->result_id()] =
+            GetFuzzerContext()->GetFreshId();
+      }
+
+      for (auto& instruction_to_inline : called_function_block) {
+        // The instructions are mapped to fresh ids.
+        if (instruction_to_inline.HasResultId()) {
+          result_id_map[instruction_to_inline.result_id()] =
+              GetFuzzerContext()->GetFreshId();
+        }
+      }
+    }
+
+    // Applies the inline function transformation.
+    ApplyTransformation(
+        TransformationInlineFunction(result_id_map, instruction->result_id()));
   }
 }
 
--- a/source/fuzz/fuzzer_pass_inline_functions.cpp
+++ b/source/fuzz/fuzzer_pass_inline_functions.cpp
@@ -51,18 +51,6 @@ void FuzzerPassInlineFunctions::Apply() {
           continue;
         }
 
-        // If |instruction| is not the penultimate instruction in |block| or
-        // |block| termination instruction is not OpBranch, then try to split
-        // |block| such that the conditions are met.
-        if ((&instruction != &*--block.tail() ||
-             block.terminator()->opcode() != SpvOpBranch) &&
-            !MaybeApplyTransformation(TransformationSplitBlock(
-                MakeInstructionDescriptor(GetIRContext(),
-                                          instruction.NextNode()),
-                GetFuzzerContext()->GetFreshId()))) {
-          continue;
-        }
-
         function_call_instructions.push_back(&instruction);
       }
     }
@@ -70,9 +58,23 @@ void FuzzerPassInlineFunctions::Apply() {
 
   // Once the function calls have been collected, it's time to actually create
   // and apply the inlining transformations.
-  for (auto& instruction : function_call_instructions) {
-    auto called_function = fuzzerutil::FindFunction(
-        GetIRContext(), instruction->GetSingleWordInOperand(0));
+  for (auto& function_call_instruction : function_call_instructions) {
+    // If |function_call_instruction| is not the penultimate instruction in its
+    // block or its block termination instruction is not OpBranch, then try to
+    // split |function_call_block| such that the conditions are met.
+    auto* function_call_block =
+        GetIRContext()->get_instr_block(function_call_instruction);
+    if ((function_call_instruction != &*--function_call_block->tail() ||
+         function_call_block->terminator()->opcode() != SpvOpBranch) &&
+        !MaybeApplyTransformation(TransformationSplitBlock(
+            MakeInstructionDescriptor(GetIRContext(),
+                                      function_call_instruction->NextNode()),
+            GetFuzzerContext()->GetFreshId()))) {
+      continue;
+    }
+
+    auto* called_function = fuzzerutil::FindFunction(
+        GetIRContext(), function_call_instruction->GetSingleWordInOperand(0));
 
     // Mapping the called function instructions.
     std::map<uint32_t, uint32_t> result_id_map;
@@ -93,8 +95,8 @@ void FuzzerPassInlineFunctions::Apply() {
     }
 
     // Applies the inline function transformation.
-    ApplyTransformation(
-        TransformationInlineFunction(instruction->result_id(), result_id_map));
+    ApplyTransformation(TransformationInlineFunction(
+        function_call_instruction->result_id(), result_id_map));
   }
 }
 
--- a/source/fuzz/fuzzer_pass_inline_functions.cpp
+++ b/source/fuzz/fuzzer_pass_inline_functions.cpp
@@ -36,6 +36,11 @@ void FuzzerPassInlineFunctions::Apply() {
   for (auto& function : *GetIRContext()->module()) {
     for (auto& block : function) {
       for (auto& instruction : block) {
+        if (!GetFuzzerContext()->ChoosePercentage(
+                GetFuzzerContext()->GetChanceOfInliningFunction())) {
+          continue;
+        }
+
         // |instruction| must be OpFunctionCall to consider applying the
         // transformation.
         if (instruction.opcode() != SpvOpFunctionCall) {
@@ -49,13 +54,25 @@ void FuzzerPassInlineFunctions::Apply() {
           continue;
         }
 
-        if (!GetFuzzerContext()->ChoosePercentage(
-                GetFuzzerContext()->GetChanceOfInliningFunction())) {
+        // Checks if some block in |called_function| has OpKill or OpUnreachable
+        // as its termination instruction.
+        bool hasKillOrUnreachableInstruction = false;
+        for (auto& called_function_block : *called_function) {
+          if (called_function_block.terminator()->opcode() == SpvOpKill ||
+              called_function_block.terminator()->opcode() ==
+                  SpvOpUnreachable) {
+            hasKillOrUnreachableInstruction = true;
+            break;
+          }
+        }
+
+        // |called_function| must not have a block with OpKill or OpUnreachable
+        // as its termination instruction.
+        if (hasKillOrUnreachableInstruction) {
           continue;
         }
 
         function_call_instructions.push_back(&instruction);
-        // Erases the function call instruction from the caller function.
       }
     }
   }
@@ -84,7 +101,7 @@ void FuzzerPassInlineFunctions::Apply() {
 
     // Applies the inline function transformation.
     ApplyTransformation(
-        TransformationInlineFunction(result_id_map, instruction->result_id()));
+        TransformationInlineFunction(instruction->result_id(), result_id_map));
   }
 }
 
--- a/source/fuzz/fuzzer_pass_inline_functions.cpp
+++ b/source/fuzz/fuzzer_pass_inline_functions.cpp
@@ -1,0 +1,91 @@
+// Copyright (c) 2020 André Perez Maselco
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_inline_functions.h"
+
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/instruction_descriptor.h"
+#include "source/fuzz/transformation_inline_function.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassInlineFunctions::FuzzerPassInlineFunctions(
+    opt::IRContext* ir_context, TransformationContext* transformation_context,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassInlineFunctions::~FuzzerPassInlineFunctions() = default;
+
+void FuzzerPassInlineFunctions::Apply() {
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      auto instruction = block.begin();
+      while (instruction != block.end()) {
+        // |instruction| must be OpFunctionCall to consider applying the
+        // transformation.
+        if (instruction->opcode() != SpvOpFunctionCall ||
+            !GetFuzzerContext()->ChoosePercentage(
+                GetFuzzerContext()->GetChanceOfInliningFunction())) {
+          ++instruction;
+          continue;
+        }
+
+        // Mapping the called function instructions.
+        std::map<uint32_t, uint32_t> result_id_map;
+        auto called_function = fuzzerutil::FindFunction(
+            GetIRContext(), instruction->GetSingleWordInOperand(0));
+        for (auto& called_function_block : *called_function) {
+          // The called function entry block label will not be inlined.
+          if (&called_function_block != &*called_function->entry()) {
+            result_id_map[called_function_block.GetLabelInst()->result_id()] =
+                GetFuzzerContext()->GetFreshId();
+          }
+
+          for (auto& instruction_to_inline : called_function_block) {
+            // If the |instruction_to_inline| result id is the returned value,
+            // then it will be mapped to the function call result id.
+            if (called_function->tail()->tail()->opcode() == SpvOpReturnValue &&
+                instruction_to_inline.HasResultId() &&
+                instruction_to_inline.result_id() ==
+                    called_function->tail()->tail()->GetSingleWordInOperand(
+                        0)) {
+              result_id_map[instruction_to_inline.result_id()] =
+                  instruction->result_id();
+              continue;
+            }
+
+            // The remaining instructions are mapped to fresh ids.
+            if (instruction_to_inline.HasResultId()) {
+              result_id_map[instruction_to_inline.result_id()] =
+                  GetFuzzerContext()->GetFreshId();
+            }
+          }
+        }
+
+        // Applies the inline function transformation.
+        ApplyTransformation(TransformationInlineFunction(
+            result_id_map, instruction->result_id()));
+
+        // Erases the function call instruction from the caller function.
+        instruction = instruction.Erase();
+      }
+    }
+  }
+}
+
+}  // namespace fuzz
+}  // namespace spvtools
--- a/source/fuzz/fuzzer_pass_inline_functions.h
+++ b/source/fuzz/fuzzer_pass_inline_functions.h
@@ -1,0 +1,42 @@
+// Copyright (c) 2020 André Perez Maselco
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SOURCE_FUZZ_FUZZER_PASS_INLINE_FUNCTIONS_H_
+#define SOURCE_FUZZ_FUZZER_PASS_INLINE_FUNCTIONS_H_
+
+#include "source/fuzz/fuzzer_pass.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// Looks for OpFunctionCall instructions and randomly decides whether to apply
+// the transformation. If the instructions of the called function are going to
+// be inlined, then a mapping, between their result ids and suitable ids, is
+// done.
+class FuzzerPassInlineFunctions : public FuzzerPass {
+ public:
+  FuzzerPassInlineFunctions(opt::IRContext* ir_context,
+                            TransformationContext* transformation_context,
+                            FuzzerContext* fuzzer_context,
+                            protobufs::TransformationSequence* transformations);
+
+  ~FuzzerPassInlineFunctions() override;
+
+  void Apply() override;
+};
+
+}  // namespace fuzz
+}  // namespace spvtools
+
+#endif  // SOURCE_FUZZ_FUZZER_PASS_INLINE_FUNCTIONS_H_
--- a/source/fuzz/fuzzer_pass_inline_functions.h
+++ b/source/fuzz/fuzzer_pass_inline_functions.h
@@ -20,10 +20,9 @@
 namespace spvtools {
 namespace fuzz {
 
-// Looks for OpFunctionCall instructions and randomly decides whether to apply
-// the transformation. If the instructions of the called function are going to
-// be inlined, then a mapping, between their result ids and suitable ids, is
-// done.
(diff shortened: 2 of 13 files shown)


source/opt/function.cpp
```diff
@@ -193,6 +193,15 @@ BasicBlock* Function::InsertBasicBlockBefore(
...
 }
 
+bool Function::HasEarlyReturn() const {
+  for (auto& block : blocks_) {
+    if (spvOpcodeIsReturn(block->tail()->opcode()) && &*block != tail()) {
+      return true;
+    }
+  }
+  return false;
+}
```
afd, line 203: This takes a syntactic approach and identifies a return that is not in the last syntactic block.

Can we change it so that it identifies early returns regardless of block layout?

Here are some tests that illustrate what I mean:

[patch.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/4918650/patch.txt)

And here is a proposed implementation of HasEarlyReturn that I think should give us what I think we need here:

```
bool Function::HasEarlyReturn() const {
  auto postdominator_analysis = blocks_.front()->GetLabel()->context()->GetPostDominatorAnalysis(this);
  for (auto& block : blocks_) {
    if (spvOpcodeIsReturn(block->tail()->opcode()) &&
        !postdominator_analysis->Dominates(block.get(), entry().get())) {
      return true;
    }
  }
  return false;
}
```

What this does is checks that if a block ends in return then it must be the case that every path from the entry block ends up at this block.  That captures what we want independently from syntactic layout.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,166 @@
...
+  *message_.mutable_result_id_map() =
+      google::protobuf::Map<google::protobuf::uint32, google::protobuf::uint32>(
+          result_id_map.begin(), result_id_map.end());
+  message_.set_function_call_id(function_call_id);
+}
+
+bool TransformationInlineFunction::IsApplicable(
+    opt::IRContext* ir_context, const TransformationContext& /*unused*/) const {
+  // |function_call_instruction| must be defined and must be an OpFunctionCall
+  // instruction.
```
afd, line 39: First of all: check that the values in the map are (a) all fresh, and (b) all distinct.  See TransformationOutlineFunction for an example.

I think currently you don't check distinctness (later).  And it's good to do these checks first as they are cheap.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  // The called function must not have an early return.
+  auto called_function = fuzzerutil::FindFunction(
+      ir_context, function_call_instruction->GetSingleWordInOperand(0));
+  if (called_function->HasEarlyReturn()) {
+    return false;
+  }
+
+  for (auto& pair : message_.result_id_map()) {
```
afd, line 54: This loop checks that every key to the map does correspond to an instruction in the function being inlined, and complains if some key is not in the function.

I believe it *does not* currently check whether every instruction in the function has some associated key.

We should turn this around:

- It doesn't matter if the map contains redundant entries, and in fact it might happen during shrinking so we want to allow it

- It *does* matter if the map doesn't contain enough entries - it has to have an entry for every result id in the function that we are inlining

Can you change things to reflect this?

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,166 @@
...
+    // |value_instruction| must be defined and must be a |called_function|
+    // instruction.
+    auto value_instruction = ir_context->get_def_use_mgr()->GetDef(pair.first);
+    if (value_instruction == nullptr ||
+        ir_context->get_instr_block(value_instruction)->GetParent() !=
+            called_function) {
+      return false;
+    }
+
+    // If it exists, the return value must map to the function call result id.
```
afd, line 64: Since we require the return value to map to the function result id, let's instead not require the return value to be captured in the map.  I.e., the map should contain entries for all of the result ids generated by the function, except for an id coming from a return instruction (if any).

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -958,6 +959,18 @@ message TransformationFunctionCall {
 
 }
 
+message TransformationInlineFunction {
+
+  // This transformation inlines a function by mapping the function instructions to suitable ids.
+
+  // Map field to map the called function instructions to suitable ids.
+  map<uint32, uint32> result_id_map = 1;
```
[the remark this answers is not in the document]
andreperezmaselco: But not all values are fresh ids. If it exists, the return value is mapped to the function call result id.
I'm going to describe it better.

source/opt/function.cpp
andreperezmaselco: You're right. The block must have a return instruction and do not dominate the entry block.
Thanks!

source/fuzz/protobufs/spvtoolsfuzz.proto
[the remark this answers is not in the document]
andreperezmaselco: Yes, this is a better explanation. I'm going to change it.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: Done.

source/opt/function.cpp
andreperezmaselco: Done.

source/fuzz/protobufs/spvtoolsfuzz.proto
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/opt/function.cpp
```diff
@@ -193,6 +193,30 @@ BasicBlock* Function::InsertBasicBlockBefore(
...
+  auto post_dominator_analysis =
+      blocks_.front()->GetLabel()->context()->GetPostDominatorAnalysis(this);
+  for (auto& block : blocks_) {
+    if (post_dominator_analysis->Dominates(block.get(), entry().get()) &&
+        block->tail()->opcode() == SpvOpReturnValue) {
+      return block->tail()->GetSingleWordOperand(0);
+    }
+  }
+  return 0;
+}
```
afd, line 206: Please move this into the transformation class.  It only makes sense in the case where there is a single return instruction in the function.  That's something the transformation assumes, but not something that is true in general, so the function doesn't belong in function.cpp.

source/opt/function.h
```diff
@@ -148,7 +148,14 @@ class Function {
   BasicBlock* InsertBasicBlockBefore(std::unique_ptr<BasicBlock>&& new_block,
                                      BasicBlock* position);
 
-  // Return true if the function calls itself either directly or indirectly.
+  // If this function has only one OpReturnValue instruction,
+  // then returns the value id. Otherwise, returns 0.
```
afd, line 152: As mentioned above, please move this into the transformation.  Then the comment should be described as a precondition:

- Requires that the function contains at most one OpReturnValue instruction.  Returns the id associated with this instruction if present, and 0 otherwise.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,176 @@
...
+  for (auto& block : *called_function) {
+    // Since the entry block label will not be inlined, only the remaining
+    // labels must have a corresponding value in the map.
+    if (&block != &*called_function->entry() &&
+        !message_.result_id_map().count(block.GetLabel()->result_id())) {
+      return false;
+    }
+    for (auto& instruction : block) {
+      if (instruction.HasResultId() &&
+          instruction.result_id() != called_function->GetReturnValueId() &&
```
afd, line 75: I would store the result of `called_function->GetReturnValueId()` in a temporary, computed before the outer loop, to avoid computing it again and again for every instruction that has a result id.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,176 @@
...
+  // |message_.result_id_map| must have an entry for every result id in the
+  // called function.
+  for (auto& block : *called_function) {
+    // Since the entry block label will not be inlined, only the remaining
+    // labels must have a corresponding value in the map.
+    if (&block != &*called_function->entry() &&
+        !message_.result_id_map().count(block.GetLabel()->result_id())) {
+      return false;
+    }
+    for (auto& instruction : block) {
```
afd, line 73: Please comment the statements in this loop more carefully to explain what is going on.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,176 @@
...
+          message_.result_id_map().at(cloned_label_instruction->result_id());
+      cloned_label_instruction->SetResultId(fresh_id);
+      function_call_instruction->InsertBefore(
+          std::unique_ptr<opt::Instruction>(cloned_label_instruction));
+      fuzzerutil::UpdateModuleIdBound(ir_context, fresh_id);
+    }
+
+    for (auto& instruction : block) {
+      // The function return instruction will not be inlined.
+      if (&instruction == &*called_function->tail()->tail()) {
```
afd, line 108: This isn't the right way to identify the function return instruction because it might not be the final instruction in the function.  Remember that it can be anywhere in the function so long as dominance rules are respected.

I assume you're going to change the single "OpReturn" or "OpReturnValue" in the function into an "OpBranch" to the basic block that follows the block containing the function call?

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,176 @@
...
+        fuzzerutil::UpdateModuleIdBound(ir_context, result_id);
+      }
+
+      // If the called function return instruction is OpReturnValue and the
+      // |cloned_instruction| result id is the returned value, then sets the
+      // |cloned_instruction| result id to the function call result id.
+      if (cloned_instruction->result_id() ==
+          called_function->GetReturnValueId()) {
+        cloned_instruction->SetResultId(function_call_instruction->result_id());
+      }
```
afd, line 152: This code should be moved to inside the "if (cloned_instruction->HasResultId())" above, otherwise you end up inspecting the result ids of instructions that don't have them and they may be spurious.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,176 @@
...
+            std::unique_ptr<opt::Instruction>(cloned_instruction));
+      } else {
+        function_call_instruction->InsertBefore(
+            std::unique_ptr<opt::Instruction>(cloned_instruction));
+      }
+    }
+  }
+
+  ir_context->InvalidateAnalysesExceptFor(opt::IRContext::kAnalysisNone);
+}
```
afd, line 167: What happens to function_call_instruction - does it get removed from the module?  (It should right?  I didn't see the code to do it, though.)

From your tests, the instruction appears to be gone, but I don't see how this happens.

Oh - in the test you kill the instruction!  That should be part of the transformation.

test/fuzz/transformation_inline_function_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+                                                      {31, 53},
+                                                      {32, 54},
+                                                      {33, 55},
+                                                      {34, 56},
+                                                      {35, 57},
+                                                      {36, 58},
+                                                      {37, 59}},
+                                                     43);
+  transformation.Apply(context.get(), &transformation_context);
+  context.get()->KillInst(function_call_instruction);
```
afd, line 259: This should be done as part of Apply.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,176 @@
...
+    const std::map<uint32_t, uint32_t>& result_id_map,
+    uint32_t function_call_id) {
+  *message_.mutable_result_id_map() =
+      google::protobuf::Map<google::protobuf::uint32, google::protobuf::uint32>(
+          result_id_map.begin(), result_id_map.end());
+  message_.set_function_call_id(function_call_id);
+}
+
+bool TransformationInlineFunction::IsApplicable(
+    opt::IRContext* ir_context, const TransformationContext& /*unused*/) const {
```
afd, line 37: Some other things to be checked:
- The OpFunctionCall instruction should be the penultimate instruction in its block
- The block terminator should be OpBranch

This is important because:

(a) it means that the block is not a loop header or selection header, which avoids various validity problems
(b) it means that we can handle OpReturn / OpReturnValue instructions by turning them into OpBranch instructions to the successor of the original block

source/opt/function.cpp
andreperezmaselco: Done.

source/opt/function.h
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: You're right.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: Done.

test/fuzz/transformation_inline_function_test.cpp
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: I guess the return instruction must be changed to the last instruction in the block containing the function call instruction.
I'll leave this changes to the next commit so that we can talk about this and the last comment.

afd: @andreperezmaselco:

- OpReturn should turn into OpBranch (to the block following the original function call)

- OpReturnValue %v should turn into %w = OpCopyObject %v; OpBranch (to the block following the original function call)
where %w is the receiving variable for the original OpFunctionCall instruction.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: Done.

paulthomson:
I have not reviewed this yet, but this is a reminder to ensure that inlined dead blocks are also marked as dead.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,239 @@
...
+                });
+          });
+
+      // If |cloned_instruction| has a result id, then set it to its mapped
+      // value.
+      if (cloned_instruction->HasResultId() &&
+          message_.result_id_map().count(cloned_instruction->result_id())) {
+        assert(message_.result_id_map().contains(
+                   cloned_instruction->result_id()) &&
+               "Result id must be mapped to a fresh id.");
```
Vasniktel, line 156: The assertion will never fail.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,239 @@
...
+  for (auto& pair : message_.result_id_map()) {
+    if (!CheckIdIsFreshAndNotUsedByThisTransformation(
+            pair.second, ir_context, &ids_used_by_this_transformation)) {
+      return false;
+    }
+  }
+
+  // |function_call_instruction| must be defined, must be an OpFunctionCall
+  // instruction, must be the penultimate instruction in its block and its block
+  // termination instruction must be an OpBranch.
```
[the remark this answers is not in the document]
Vasniktel: I see that we don't and that's good. One particular issue with this case is that once we inline the function, we might end up with a situation where a back-edge of the loop does not point to the loop's header.

It might be good to make an explicit comment about this and add a test for it.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,239 @@
...
+  message_.set_function_call_id(function_call_id);
+}
+
+bool TransformationInlineFunction::IsApplicable(
+    opt::IRContext* ir_context, const TransformationContext& /*unused*/) const {
+  // The values in the |message_.result_id_map| must be all fresh and all
+  // distinct.
+  std::set<uint32_t> ids_used_by_this_transformation;
+  for (auto& pair : message_.result_id_map()) {
+    if (!CheckIdIsFreshAndNotUsedByThisTransformation(
```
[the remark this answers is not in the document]
andreperezmaselco: It is in the `Transformation` class: `source/fuzz/transformation.cpp:304`.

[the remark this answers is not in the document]
Vasniktel: Right, thanks! I've never used it myself :)

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,239 @@
...
+    }
+  }
+
+  // |function_call_instruction| must be defined, must be an OpFunctionCall
+  // instruction, must be the penultimate instruction in its block and its block
+  // termination instruction must be an OpBranch.
+  auto function_call_instruction =
+      ir_context->get_def_use_mgr()->GetDef(message_.function_call_id());
+  auto function_call_instruction_block =
+      ir_context->get_instr_block(function_call_instruction);
```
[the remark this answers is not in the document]
andreperezmaselco: I don't think so, because I've already checked that `function_call_instruction` is not `nullptr` and it is an `SpvOpFunctionCall` instruction. Therefore, `function_call_instruction_block` can not be `nullptr` because an `SpvOpFunctionCall` instruction is only valid inside a block.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: It will fail if some instruction, inside `called_function`, is not mapped.

Vasniktel: I've just noticed that condition in the `if` statement is the same (almost) as the one in the assertion (I believe `count` function returns `0` if `contains` returns `false`).

source/fuzz/transformation_inline_function.cpp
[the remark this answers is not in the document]
paulthomson: `message_.function_call_id()` might not exist in the module. So `function_call_instruction = nullptr`. Then we call `ir_context->get_instr_block(nullptr);`. I believe this won't crash, but it is not ideal. I suggest you check that `function_call_instruction` is non-null immediately, like so:

```
auto function_call_instruction =
      ir_context->get_def_use_mgr()->GetDef(message_.function_call_id());
if (!function_call_instruction) {
  return false;
}
```

Just because this is clearer.

source/fuzz/fuzzer_pass_inline_functions.cpp
```diff
@@ -0,0 +1,92 @@
...
+          continue;
+        }
+
+        if (!GetFuzzerContext()->ChoosePercentage(
+                GetFuzzerContext()->GetChanceOfInliningFunction())) {
+          continue;
+        }
+
+        function_call_instructions.push_back(&instruction);
+        // Erases the function call instruction from the caller function.
```
paulthomson, line 58: I suggest removing this comment because it is unclear (_what_ erases the function instruction and _why_ is that information relevant here).

source/fuzz/transformation_inline_function.cpp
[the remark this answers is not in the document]
paulthomson: @Vasniktel I don't quite understand. Can you give an example?

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,239 @@
...
+                ->terminator()
+                ->GetSingleWordInOperand(0);
+        switch (cloned_instruction->opcode()) {
+          case SpvOpReturn:
+            cloned_instruction->AddOperand(
+                {SPV_OPERAND_TYPE_ID, {following_block_id}});
+            break;
+          case SpvOpReturnValue: {
+            auto returned_value_id =
+                GetReturnValueId(ir_context, called_function);
```
paulthomson, line 177: Could you just get the returned value id from `instruction`? You could also rename it `original_instruction` for clarity and less chance of name clashes.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,239 @@
...
+    if (&block != &*called_function->entry()) {
+      auto cloned_label_instruction = block.GetLabelInst()->Clone(ir_context);
+      assert(message_.result_id_map().contains(
+                 cloned_label_instruction->result_id()) &&
+             "Label id must be mapped to a fresh id.");
+      uint32_t fresh_id =
+          message_.result_id_map().at(cloned_label_instruction->result_id());
+      cloned_label_instruction->SetResultId(fresh_id);
+      function_call_instruction->InsertBefore(
+          std::unique_ptr<opt::Instruction>(cloned_label_instruction));
```
paulthomson, line 121: Actually, I don't think this will work. I think your tests might work because the functions always only have one basic block. But if you have additional basic blocks, you will need to add basic blocks, not just add label instructions. I _think_ adding a label instruction does not create a basic block automatically. See this example:

https://github.com/KhronosGroup/SPIRV-Tools/blob/150be20d4334b66c31a8ed81230f85b814b14a9f/source/fuzz/transformation_add_function.cpp#L276

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,239 @@
...
+      ir_context->get_def_use_mgr()->GetDef(message_.function_call_id());
+  auto caller_function =
+      ir_context->get_instr_block(function_call_instruction)->GetParent();
+  auto called_function = fuzzerutil::FindFunction(
+      ir_context, function_call_instruction->GetSingleWordInOperand(0));
+
+  for (auto& block : *called_function) {
+    // The called function entry block label will not be inlined.
+    if (&block != &*called_function->entry()) {
+      auto cloned_label_instruction = block.GetLabelInst()->Clone(ir_context);
```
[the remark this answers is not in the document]
paulthomson: But see a later comment because I think this loop might not work.

source/fuzz/transformation_inline_function.cpp
Vasniktel: Yes, you would need to invalidate analyses to make IRContext aware of the new basic blocks. But in this case, you will potentially invalidate all pointers as well.

paulthomson: Actually, I think invaliding analyzes is not even enough. I think the opt code base stores a module, not as a list of instructions, but as a list of functions, each with a list of basic blocks, each with a list of instructions. Thus, adding a label will not create a new basic block in the in-memory data structures, unless maybe you output the module as bytes and re-parsed it.

source/fuzz/transformation_inline_function.cpp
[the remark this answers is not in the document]
Vasniktel: I've been thinking about this situation:
```
%1 = OpFunction %int None %type
%2 = OpLabel
OpBranch %3
%3 = OpLabel
OpReturnValue %int_constant
OpFunctionEnd
...
%4 = OpLabel
%5 = OpFunctionCall %int %1
OpLoopMerge ....
OpBranch %6
...
```
If we inline `%1`, we might get:
```
%4 = OpLabel
OpBranch %7
%7 = OpLabel
%8 = OpCopyObject %int %inst_constant
OpLoopMerge ...
OpBranch %6
...
```
In this situation, there is a back-edge block somewhere branching to `%4` which is no longer a loop header.

source/fuzz/transformation_inline_function.cpp
Vasniktel: Yes, right. OpLabels are not even stored in the instruction list. They are stored separately in each basic block.

source/fuzz/transformation_inline_function.cpp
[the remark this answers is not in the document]
paulthomson: Oh I see. You mean if we did not require the call to be the penultimate instruction then we could get the above situation. Yes, it is good that we avoid this. No action is needed for this comment chain, right?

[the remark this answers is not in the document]
Vasniktel: Yeah, it would be good to comment and test it, though.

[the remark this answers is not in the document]
paulthomson: Ah right, yes you said that. Sounds good. E.g. a comment like "This avoids the case where..." and also a test that tries to inline a function call at the end of a loop header.

source/fuzz/transformation_inline_function.cpp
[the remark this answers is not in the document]
andreperezmaselco: But this is already done in the following `if` statement.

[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_inline_functions.cpp
andreperezmaselco: Done.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1074,6 +1075,20 @@ message TransformationFunctionCall {
...
 }
 
+message TransformationInlineFunction {
+
+  // This transformation inlines a function by mapping the function instructions to fresh ids.
+
+  // For each result id defined by the called function,
+  // this map provides an associated fresh id that can
+  // be used in the inlined version of the function call.
+  map<uint32, uint32> result_id_map = 1;
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+            uint32_t parameter_index = 0;
+            called_function->ForEachParam(
+                [id, function_call_instruction,
+                 &parameter_index](opt::Instruction* parameter_instruction) {
+                  // If the id is a function parameter, then set it to the
+                  // parameter value passed in the function call instruction.
+                  if (*id == parameter_instruction->result_id()) {
+                    *id = function_call_instruction->GetSingleWordInOperand(
+                        parameter_index + 1);
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: You're right.
Done.

source/opt/function.cpp
```diff
@@ -193,6 +193,18 @@ BasicBlock* Function::InsertBasicBlockBefore(
...
+  auto post_dominator_analysis =
+      blocks_.front()->GetLabel()->context()->GetPostDominatorAnalysis(this);
+  for (auto& block : blocks_) {
+    if (spvOpcodeIsReturn(block->tail()->opcode()) &&
+        !post_dominator_analysis->Dominates(block.get(), entry().get())) {
+      return true;
+    }
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/opt/function.cpp
```diff
@@ -227,6 +227,18 @@ BasicBlock* Function::InsertBasicBlockBefore(
   return nullptr;
 }
 
+bool Function::HasEarlyReturn() const {
```
Vasniktel, line 230: Don't forget to remove this change once you've refactored it into a separate PR or a static method.

source/fuzz/fuzzer_pass_inline_functions.cpp
```diff
@@ -0,0 +1,109 @@
...
+        // as its termination instruction.
+        bool hasKillOrUnreachableInstruction = false;
+        for (auto& called_function_block : *called_function) {
+          if (called_function_block.terminator()->opcode() == SpvOpKill ||
+              called_function_block.terminator()->opcode() ==
+                  SpvOpUnreachable) {
+            hasKillOrUnreachableInstruction = true;
+            break;
+          }
+        }
```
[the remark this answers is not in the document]
andreperezmaselco: Actually, a recent PR added `fuzzerutil::FunctionContainsOpKillOrUnreachable` that I'm doing to use.

source/opt/function.cpp
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_inline_functions.cpp
```diff
@@ -0,0 +1,109 @@
...
+        // |called_function| must not have an early return.
+        auto called_function = fuzzerutil::FindFunction(
+            GetIRContext(), instruction.GetSingleWordInOperand(0));
+        if (called_function->HasEarlyReturn()) {
+          continue;
+        }
+
+        // Checks if some block in |called_function| has OpKill or OpUnreachable
+        // as its termination instruction.
+        bool hasKillOrUnreachableInstruction = false;
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
andreperezmaselco: Done.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,234 @@
...
+      if (spvOpcodeIsReturn(cloned_instruction->opcode())) {
+        uint32_t following_block_id =
+            ir_context->get_instr_block(function_call_instruction)
+                ->terminator()
+                ->GetSingleWordInOperand(0);
+        switch (cloned_instruction->opcode()) {
+          case SpvOpReturn:
+            cloned_instruction->AddOperand(
+                {SPV_OPERAND_TYPE_ID, {following_block_id}});
+            break;
```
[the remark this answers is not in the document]
Vasniktel: One solution I've just recalled is to use `OpUndef` of void type. However, this is controversial and there is a discussion (#3716) on whether this should be allowed: currently, it is not according to the spec but spirv-val doesn't mind.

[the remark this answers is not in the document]
Vasniktel: Prohibiting this case is probably the best option.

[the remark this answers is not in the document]
andreperezmaselco: Yes, you're right.
I've added a condition that prohibits this case.

source/fuzz/fuzzer_pass_inline_functions.cpp
```diff
@@ -0,0 +1,109 @@
...
+        for (auto& called_function_block : *called_function) {
```
afd, line 60: See [Google C++ style](https://github.com/KhronosGroup/SPIRV-Tools/pull/3728) - variable names should be lower case with underscores between words.

source/fuzz/fuzzer_pass_inline_functions.cpp
```diff
@@ -0,0 +1,109 @@
...
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassInlineFunctions::~FuzzerPassInlineFunctions() = default;
+
+void FuzzerPassInlineFunctions::Apply() {
+  std::vector<opt::Instruction*> function_call_instructions;
+
```
afd, line 35: Add a comment that these are the instructions we are going to inline, and explain that we find them first, and then do the inlining, to avoid changing the module while we are inspecting it.

source/fuzz/fuzzer_pass_inline_functions.cpp
```diff
@@ -0,0 +1,109 @@
...
+        if (hasKillOrUnreachableInstruction) {
+          continue;
+        }
+
+        function_call_instructions.push_back(&instruction);
+      }
+    }
+  }
+
+  for (auto& instruction : function_call_instructions) {
```
afd, line 80: I would add a comment at this point to clarify that we have now decided which calls to inline, and it's time to actually create and apply the inlining transformations.

source/fuzz/fuzzer_pass_inline_functions.h
```diff
@@ -0,0 +1,42 @@
...
+#define SOURCE_FUZZ_FUZZER_PASS_INLINE_FUNCTIONS_H_
+
+#include "source/fuzz/fuzzer_pass.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// Looks for OpFunctionCall instructions and randomly decides whether to apply
+// the transformation. If the instructions of the called function are going to
+// be inlined, then a mapping, between their result ids and suitable ids, is
```
afd, line 25: "... whether to apply the transformation." -> "... which ones to inline."

source/fuzz/transformation_inline_function.h
```diff
@@ -0,0 +1,64 @@
...
+  explicit TransformationInlineFunction(
+      const protobufs::TransformationInlineFunction& message);
+
+  TransformationInlineFunction(
+      uint32_t function_call_id,
+      const std::map<uint32_t, uint32_t>& result_id_map);
+
+  // - |message_.result_id_map| must map the instructions of the called function
+  // to fresh ids.
+  // - |message_.function_call_id| must be an OpFunctionCall instruction. The
```
afd, line 37: Indentation: the "to" should line up underneath "|message_"

afd, line 37: Indentation: the "to" should line up underneath "|message_"

afd: (This is meant to come under the line "to fresh ids", but the CLion code review isn't letting me put it there.)

source/fuzz/transformation_inline_function.h
```diff
@@ -0,0 +1,64 @@
...
+  // called function must not have an early return.
+  bool IsApplicable(
```
afd, line 39: Similarly, fix indentation here.

Also add a note that the called function must not use OpUnreachable or OpKill, and briefly say that this is to guard against making hte module invalid when the caller is inside a continue construct.

With respect to this new bullet that I'm proposing: can you open an issue to note that we could make the spirv-fuzz inliner more aggressive by allowing functions that use OpKill / OpUnreachable to be inlined if the function call is *not* part of a continue construct, and do a TODO(issue URL) comment here?

source/fuzz/transformation_inline_function.h
```diff
@@ -0,0 +1,64 @@
...
+      opt::IRContext* ir_context,
+      const TransformationContext& transformation_context) const override;
+
+  // Iterates over the called function blocks and clones each instruction in the
+  // blocks. The cloned instruction result id and its operand ids are set to the
+  // corresponding value in |message_.result_id_map|. Finally, the cloned
+  // instructions are inserted into the caller function.
+  void Apply(opt::IRContext* ir_context,
```
afd, line 47: This comment describes the *implementation* of Apply.  It's better to try to express the overall effect of a function in a comment.

For example:

"Replaces the OpFunctionCall instruction identified by |message_.function_call_id| with a copy of the function's body.  |message_.result_id_map| is used to provide fresh ids for duplicated instructions."

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,268 @@
...
+  if (!function_call_instruction ||
+      function_call_instruction->opcode() != SpvOpFunctionCall) {
+    return false;
+  }
+
+  // |function_call_instruction| must be the penultimate instruction in its
+  // block and its block termination instruction must be an OpBranch. This
+  // avoids the case where the penultimate instruction is an OpLoopMerge, which
+  // would make the back-edge block not branch to the loop header.
+  auto function_call_instruction_block =
```
afd, line 62: This is good, but does the fuzzer pass guarantee that this is the case before applying a transformation?

If the fuzzer pass finds an OpFunctionCall instruction that can be inlined but doese *not* meet these conditions, it should apply TransformationSplitBlock so that the function call instruction now does meet the required conditions.

source/fuzz/fuzzer_pass_inline_functions.cpp
```diff
@@ -0,0 +1,109 @@
...
+          result_id_map[instruction_to_inline.result_id()] =
+              GetFuzzerContext()->GetFreshId();
+        }
+      }
+    }
+
+    // Applies the inline function transformation.
+    ApplyTransformation(
+        TransformationInlineFunction(instruction->result_id(), result_id_map));
+  }
```
afd, line 105: It looks to me like the transformation might not be applicable by construction, in particular if the function call is not the penultimate instruction in its block and if the block terminator is not OpBranch.

To avoid this problem you should detect when this is the case, and apply a TransformationSplitBlock before applying the inlining transformation.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+  // If |function_call_instruction| type is void, then assert there is no uses.
+  if (ir_context->get_type_mgr()
+          ->GetType(function_call_instruction->type_id())
+          ->AsVoid()) {
+    assert(ir_context->get_def_use_mgr()->NumUses(function_call_instruction) ==
+               0 &&
+           "Function call with void return must not be used");
+  }
+
```
afd, line 78: I don't think we can assume that this never happens.  For example, OpName is allowed to target any result id, so it could target the result id of a function call.

Please:

(1) Change this so that it returns false if there are uses
(2) Add a test that confirms that a transformation will be inapplicable if the return type is void and there is a use (e.g. via OpName)
(3) Adapt the fuzzer pass so that it skips function call instructions that would fail to be applicable due to this issue

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,268 @@
...
+  // The called function must not have an early return.
+  auto called_function = fuzzerutil::FindFunction(
+      ir_context, function_call_instruction->GetSingleWordInOperand(0));
+  if (called_function->HasEarlyReturn()) {
+    return false;
+  }
+
```
afd, line 85: To limit duplication between the fuzzer pass and the transformation, how about you introduce a static helper function in the transformation class, IsSuitableForInlining.

This function could:

- Check the void return type constraints
- Ensure there are no early returns
- Ensure that there are no uses of OpKill or OpUnreachable

In both the fuzzer pass and the transformation, those checks could be accounted for by calling the helper function.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,268 @@
...
+  for (auto& block : *called_function) {
+    // |called_function| must not have a block with OpKill or OpUnreachable as
+    // its termination instruction.
+    if (block.terminator()->opcode() == SpvOpKill ||
+        block.terminator()->opcode() == SpvOpUnreachable) {
+      return false;
+    }
+
```
afd, line 93: This would be come unnecessary if you use the helper function that I'm proposing.

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,268 @@
...
+    // function.
+    for (auto& instruction : block) {
+      // If |instruction| has result id, then it must have a mapped id in
+      // |result_id_map|.
+      if (instruction.HasResultId() &&
+          !result_id_map.count(instruction.result_id())) {
+        return false;
+      }
+    }
+  }
```
afd, line 111: Can you open an issue for the fact that we could use overflow ids to be less conservative here, and add a TODO(comment)?  (The overflow ids stuff is not yet merged but it's coming soon.)

source/fuzz/transformation_inline_function.cpp
```diff
@@ -0,0 +1,268 @@
...
+}
+
+protobufs::Transformation TransformationInlineFunction::ToMessage() const {
+  protobufs::Transformation result;
+  *result.mutable_inline_function() = message_;
+  return result;
+}
+
+void TransformationInlineFunction::InlineInstruction(
+    opt::IRContext* ir_context,
```
afd, line 192: I like this helper function but I do not like its name.

The reason I don't like its name is that you have *already* inserted the instruction by the time you call this helper, so in a sense the inlining has already been done, and now you are adapting the instruction to make it fit.

How about AdaptInlinedInstruction?

source/fuzz/transformation_inline_function.cpp
afd: I will take care of this.

afd (approved):
Thanks for making the changes.

This looks really good, and I'm happy to merge it!

PR #3519, opened 2020-07-12
spirv-fuzz: Support adding dead break from back-edge block

### LINKED ISSUE #2577: Allow spirv-fuzz to generate a dead break from the back-edge block of a loop
At present SPIR-V fuzz does not consider generating a dead break from a loop's continue construct to the loop's merge block, as this is usually illegal.  But it is legal if the block being broken from is the loop's back edge block.  It would be good to extend the tool to allow for this case.

andreperezmaselco:
Fixes #2577.

### DIFF
--- a/source/fuzz/fuzzer_util.h
+++ b/source/fuzz/fuzzer_util.h
@@ -64,6 +64,12 @@ void AddUnreachableEdgeAndUpdateOpPhis(
     bool condition_value,
     const google::protobuf::RepeatedField<google::protobuf::uint32>& phi_ids);
 
+// Returns true if and only if |loop_header_id| is in fact loop header and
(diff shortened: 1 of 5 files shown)


source/fuzz/fuzzer_util.h
```diff
@@ -64,6 +64,12 @@ void AddUnreachableEdgeAndUpdateOpPhis(
     bool condition_value,
     const google::protobuf::RepeatedField<google::protobuf::uint32>& phi_ids);
 
+// Returns true if and only if |loop_header_id| is in fact loop header and
```
afd, line 67: "is in fact loop header" -> "is a loop header"

andreperezmaselco: Done.

PR #3522, opened 2020-07-13
Added TransformationAddRelaxedDecoration

antnkarp:
Added this transformation for now without related fuzzer pass. This transformation is very closesly related to TransformationAddNoContractionDecoration, so I reused some of its code. Operations which are applicable are taken from SPIR-V documentation. These are all operations that return 32-bit integers or floats.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -345,7 +345,7 @@ set_source_files_properties(
 
 spvtools_pch(SPIRV_SOURCES pch_source)
 
-add_library(${SPIRV_TOOLS} ${SPIRV_SOURCES} ../test/fuzz/transformation_add_relaxed_decoration_test.cpp)
+add_library(${SPIRV_TOOLS} ${SPIRV_SOURCES})
 spvtools_default_compile_options(${SPIRV_TOOLS})
 target_include_directories(${SPIRV_TOOLS}
   PUBLIC
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -345,7 +345,7 @@ set_source_files_properties(
 
 spvtools_pch(SPIRV_SOURCES pch_source)
 
-add_library(${SPIRV_TOOLS} ${SPIRV_SOURCES})
+add_library(${SPIRV_TOOLS} ${SPIRV_SOURCES} ../test/fuzz/transformation_add_relaxed_decoration_test.cpp)
 spvtools_default_compile_options(${SPIRV_TOOLS})
 target_include_directories(${SPIRV_TOOLS}
   PUBLIC
--- a/source/fuzz/fuzzer_pass_add_relaxed_decorations.cpp
+++ b/source/fuzz/fuzzer_pass_add_relaxed_decorations.cpp
@@ -1,0 +1,57 @@
+// Copyright (c) 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_add_relaxed_decorations.h"
+
+#include "source/fuzz/transformation_add_relaxed_decoration.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassAddRelaxedDecorations::FuzzerPassAddRelaxedDecorations(
+    opt::IRContext* ir_context, TransformationContext* transformation_context,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassAddRelaxedDecorations::
+~FuzzerPassAddRelaxedDecorations() = default;
+
+void FuzzerPassAddRelaxedDecorations::Apply() {
+  // Consider every instruction in every block in every function.
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      for (auto& inst : block) {
+        // Restrict attention to numeric instructions (returning 32-bit floats
+        // or ints according to SPIR-V documentation).
+        if (TransformationAddRelaxedDecoration::IsNumeric(
+            inst.opcode())) {
+          // Randomly choose whether to apply the RelaxedPrecision decoration to
+          // this arithmetic instruction.
+          if (GetFuzzerContext()->ChoosePercentage(
+              GetFuzzerContext()
+                  ->GetChanceOfAddingRelaxedDecoration())) {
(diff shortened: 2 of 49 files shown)


source/fuzz/transformation_add_relaxed_decoration.cpp
```diff
@@ -0,0 +1,143 @@
...
+bool TransformationAddRelaxedDecoration::IsApplicable(
+    opt::IRContext* ir_context, const TransformationContext& transformation_context) const {
+  // |message_.result_id| must be the id of an instruction.
+  auto instr = ir_context->get_def_use_mgr()->GetDef(message_.result_id());
+  if (!instr) {
+    return false;
+  }
+  opt::BasicBlock* cur_block =
+      ir_context->get_instr_block(instr);
+  // current block must be a dead block
```
paulthomson, line 42: Please start comments with a capital letter and end with a full stop (period). Prefer proper sentences. E.g. Prefer "The id must be..." vs. "Id must be...". Referring to variables is fine though: e.g. "|id| must be..." (and no capital letter is needed). 

You have done this fine but also notice: comments inside functions use the style:

"Add foo to bar."

but documentation comments before a function declaration always use the style:

"Adds..."
"Returns..."
"Transforms...", etc., 

at least for the first sentence.

source/fuzz/transformation_add_relaxed_decoration.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+bool TransformationAddRelaxedDecoration::IsNumeric(uint32_t opcode) {
+  switch (opcode) {
+    case SpvOpConvertFToU:
+    case SpvOpConvertFToS:
+    case SpvOpConvertSToF:
+    case SpvOpConvertUToF:
+    case SpvOpUConvert:
+    case SpvOpSConvert:
+    case SpvOpFConvert:
```
paulthomson, line 73: I don't think we want this function because I don't think RelaxedPrecision is only applied to numeric instructions. But if we did want this function, I think we should move the original function to fuzzer_util.h (or elsewhere) and re-use it, instead of duplicating it.

paulthomson:
Two more things I missed:

 - the format check failed. Please see our hello doc for one approach to setting up clang-format-5 (it must be version 5) with CLion.
- We will need some tests for this transformation!

antnkarp:
I'll fix the code, add required tests and also add fuzzer-pass (because it's straightforward to me) and submit next PR.

paulthomson:
Update to my suggestion to check spirv-val: I don't think spirv-val checks RelaxedPrecision decorations. So we can't copy spirv-val.

antnkarp:
Now it's building properly. Thank you.
I'm not sure how this line got there. Is there any specific command to avoid in CLion/or maybe there is a way to disable this kind of modifications?

source/CMakeLists.txt
```diff
@@ -345,7 +345,7 @@ set_source_files_properties(
 
 spvtools_pch(SPIRV_SOURCES pch_source)
 
-add_library(${SPIRV_TOOLS} ${SPIRV_SOURCES})
+add_library(${SPIRV_TOOLS} ${SPIRV_SOURCES} ../test/fuzz/transformation_add_relaxed_decoration_test.cpp)
```
paulthomson, line 348: I think this is the problem!

paulthomson:
I think when adding a file, CLion will ask if you want to update the CMakeLists.txt files. You should choose "no" :-)

You can check your changes locally with `git diff`. Then it will be obvious. CLion can also show you your local changes.

source/fuzz/transformation_add_no_contraction_decoration.cpp
```diff
@@ -37,6 +37,11 @@ bool TransformationAddNoContractionDecoration::IsApplicable(
   if (!instr) {
     return false;
   }
+  opt::BasicBlock* cur_block = ir_context->get_instr_block(instr);
+  // The instruction must have a block.
+  if (cur_block == nullptr) {
+    return false;
+  }
```
paulthomson, line 44: I don't think this is actually needed, as the IsArithmetic should restrict to the valid cases. My previous comment was mainly about checking that there was not another transformation getting the block of an instruction and then using it without checking for nullptr.

source/fuzz/transformation_add_relaxed_decoration.cpp
paulthomson: Actually, I think we do want this function. But we could maybe avoid the duplication.

source/fuzz/fuzzer_pass_add_relaxed_decorations.cpp
```diff
@@ -0,0 +1,55 @@
...
+      for (auto& inst : block) {
+        // Restrict attention to numeric instructions (returning 32-bit floats
+        // or ints according to SPIR-V documentation) in dead blocks.
+        TransformationAddRelaxedDecoration transformation(inst.result_id());
+        if (transformation.IsApplicable(GetIRContext(),
+                                        *GetTransformationContext())) {
+          // Randomly choose whether to apply the RelaxedPrecision decoration to
+          // this numeric instruction.
+          if (GetFuzzerContext()->ChoosePercentage(
+                  GetFuzzerContext()->GetChanceOfAddingRelaxedDecoration())) {
```
afd, line 44: Please move this check to the start of the for loop, so that we don't redundantly make the transformation and check its applicability if we turn out not to be in the mood to apply it once done.  That's the patterns we use in other fuzzer passes and it makes not difference from a probabilistic perspective.

source/fuzz/fuzzer_pass_add_relaxed_decorations.cpp
```diff
@@ -0,0 +1,55 @@
...
+            transformation.Apply(GetIRContext(), GetTransformationContext());
+            *GetTransformations()->add_transformation() =
+                transformation.ToMessage();
+          }
```
afd, line 48: Once #3540 we should re-use the method that it is going to add.

source/fuzz/transformation_add_no_contraction_decoration.cpp
afd: I agree - please revert this, @antnkarp.

source/fuzz/transformation_add_relaxed_decoration.cpp
afd: @alan-baker The validator doesn't seem to complain if I apply RelaxedPrecision to e.g. OpConstantTrue, or the result of a load.

Is it legitimate to apply RelaxedPrecision to ids that are non-numeric, with the expectation that it would just be ignored?

alan-baker: I think the validator doesn't check what it is applied to. I agree it doesn't really make sense to apply to a bool.

source/fuzz/fuzzer_pass_add_relaxed_decorations.cpp
antnkarp: Done

source/fuzz/transformation_add_no_contraction_decoration.cpp
antnkarp: Done

antnkarp: Done

afd: @antnkarp Thanks - please remember to resolve a comment once you've addressed it - only leave it open if you are unsure how to address it or if you have posted a follow-up question.  So you can resolve this one, and please go up and resolve any above comments that you have addressed.

PR #3525, opened 2020-07-13
Fix Wunreachable-code-loop-increment error in clang++

jonahryandavis:
Clang complains about this for loop being exited on the first iteration.
Instead, just access the first member of this unordered_set if it's not
empty.


jonahryandavis:
@greg-lunarg Hey, I think this is functionally equivalent, and it stops clang from complaining in ANGLE's builds.

zoddicus:
This is superseded by #3513

Issue #3185, opened 2020-02-11
spirv-fuzz: Strict assertions for testing


afd:
When checking the precondition of a transformation, there are cases where the transformation is potentially valid but does not apply, and cases where the transformation is completely bogus and should not have been created in the first place.

For the second case, when writing transformation unit tests we would like to be able to have negative tests that check that such transformations are not applicable.

But when actually running the fuzzer, we would like an assertion to fail when a completely invalid transformation is created.

Some thought is required to figure out how best to allow both cases.

afd:
Death tests provide a solution for this.

paulthomson:
There are still TODOs that reference this issue.

PR #3528, opened 2020-07-14
Assert false in IsApplicable method of TransformationAccessChain

stefanomil:
… if it is invalid (with null or undefined pointer).

In this case, assert false instead of returning false so that the fuzzer fails.

(See issue #3185 )

### DIFF
--- a/source/fuzz/transformation_access_chain.cpp
+++ b/source/fuzz/transformation_access_chain.cpp
@@ -74,10 +74,7 @@ bool TransformationAccessChain::IsApplicable(
   switch (pointer->opcode()) {
     case SpvOpConstantNull:
     case SpvOpUndef:
-      // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3185): When
-      //  fuzzing for real we would like an 'assert(false)' here.  But we also
-      //  want to be able to write negative unit tests.
-      return false;
+      assert(false && "Null or undefined pointer");
     default:
       break;
   }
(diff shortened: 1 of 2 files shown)


source/fuzz/transformation_access_chain.cpp
```diff
@@ -74,10 +74,7 @@ bool TransformationAccessChain::IsApplicable(
   switch (pointer->opcode()) {
     case SpvOpConstantNull:
     case SpvOpUndef:
-      // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3185): When
-      //  fuzzing for real we would like an 'assert(false)' here.  But we also
-      //  want to be able to write negative unit tests.
-      return false;
+      assert(false && "Null or undefined pointer");
```
afd, line 77: Please change message to: "Access chains should not be created from null/undefined pointers".

And please add "return false" after the assertion.  That way in release mode - when there are no assertions - the function will return false, which is the right thing to do.

test/fuzz/transformation_access_chain_test.cpp
```diff
@@ -184,14 +184,16 @@ TEST(TransformationAccessChainTest, BasicTest) {
                    .IsApplicable(context.get(), transformation_context));
 
   // Bad: pointer is null
-  ASSERT_FALSE(TransformationAccessChain(
+  ASSERT_DEATH(TransformationAccessChain(
                    100, 45, {80}, MakeInstructionDescriptor(24, SpvOpLoad, 0))
```
afd, line 188: Uses of ASSERT_DEATH need to be wrapped in:

```
#ifndef NDEBUG
// comment
ASSERT_DEATH(...);
#endif
```

See transformation_permute_phi_operands_test.cpp for an example.

This is to keep release compilers happy - death tests don't work in release mode.

stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Done

PR #3532, opened 2020-07-14
Remove single iteration loop

zoddicus:
This loop is intentionally designed to only ever execute, which is non-obvious
on first glance. Changing the code to be clearer that the entire data structure
isn't being iterated.

This will also fix importing SPIRV-Tools into Chromium, since their style
checkers are failing the build due to this iterate once construct.


jaebaek:
LGTM but it seems that #3513 will solve this issue.

zoddicus:
Superseded by #3513

PR #3535, opened 2020-07-15
spirv-fuzz: Add replay range option

afd:
This change adds a --replay-range argument to spirv-fuzz that
facilitates applying only a prefix of transformations.

### DIFF
--- a/tools/fuzz/fuzz.cpp
+++ b/tools/fuzz/fuzz.cpp
@@ -121,6 +121,18 @@ Options (in lexicographical order):
   --replay
                File from which to read a sequence of transformations to replay
                (instead of fuzzing)
+  --replay-range=
+               Signed 32-bit integer.  If set to a positive value N, only the
+               first N transformations will be applied during replay.  If set to
+               a negative value -N, all but the final N transformations will be
+               applied during replay.  If set to 0 (the default), all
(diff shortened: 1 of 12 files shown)


tools/fuzz/fuzz.cpp
```diff
@@ -208,6 +215,15 @@ FuzzStatus ParseFlags(int argc, const char** argv, std::string* in_binary_file,
...
         const auto split_flag = spvtools::utils::SplitFlagArgs(cur_arg);
         *replay_transformations_file = std::string(split_flag.second);
+      } else if (0 == strncmp(cur_arg, "--replay-range=",
+                              sizeof("--replay-range=") - 1)) {
+        const auto split_flag = spvtools::utils::SplitFlagArgs(cur_arg);
+        char* end = nullptr;
+        errno = 0;
+        const auto replay_range =
+            static_cast<int32_t>(strtol(split_flag.second.c_str(), &end, 10));
+        assert(end != split_flag.second.c_str() && errno == 0);
```
paulthomson, line 225: Although we do it elsewhere, I am not sure that using assert is a good idea.

afd: I guess this is something we should revisit globally.  I'll leave it for now.

Issue #3533, opened 2020-07-15
spirv-fuzz: Extend TransformationRecordSynonymousConstants to accept zero-like composite constants


stefanomil:
Right now, TransformationRecordSynonymousConstants accepts constants if they are equal or if they represent zero-like scalar values of the same type. This could be extended to zero composite constants.

stefanomil:
This requires the IsApplicable method to be rethought. There should be a AreEquivalentConstants (so similar) recursive function that checks the equivalency of components.

stefanomil:
Fixed by #3537

stefanomil:
#3537 still needs to be merged

PR #3537, opened 2020-07-15
spirv-fuzz: Extend TransformationRecordSynonymousConstants to allow composite constants

stefanomil:
Implemented AreEquivalentConstants method to check equivalency of constants, changing IsApplicable method of TransformationRecordSynonymousConstants to allow recording equivalence of composite constants; added some tests to check this.

Tests with arrays and matrices still need to be added.

Related issue: #3533

### DIFF
--- a/source/fuzz/transformation_record_synonymous_constants.cpp
+++ b/source/fuzz/transformation_record_synonymous_constants.cpp
@@ -23,44 +23,50 @@ namespace {
 // Returns true if the two given constants are equivalent
 // (the description of IsApplicable specifies the conditions they must satisfy
 // to be considered equivalent)
-bool AreEquivalentConstants(opt::IRContext* ir_context,
-                            const opt::analysis::Constant& constant1,
-                            const opt::analysis::Constant& constant2) {
-  // Check that the type ids are the same
+bool AreEquivalentConstants(opt::IRContext* ir_context, uint32_t constant_id1,
+                            uint32_t constant_id2) {
+  opt::Instruction* def_1 = ir_context->get_def_use_mgr()->GetDef(constant_id1);
+  opt::Instruction* def_2 = ir_context->get_def_use_mgr()->GetDef(constant_id2);
+
+  // Check that the definitions exist
+  assert(def_1 && def_2 && "The constant ids must exist in the module.");
+
+  // The type ids must be the same
   // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3536): Somehow
   // relax this for integers (so that unsigned integer and signed integer are
   // considered the same type)
-  uint32_t type_id1 = ir_context->get_type_mgr()->GetId(constant1.type());
-  uint32_t type_id2 = ir_context->get_type_mgr()->GetId(constant2.type());
-  if (type_id1 != type_id2) {
+  if (def_1->type_id() != def_2->type_id()) {
     return false;
   }
 
+  auto constant1 = ir_context->get_constant_mgr()->GetConstantFromInst(def_1);
+  auto constant2 = ir_context->get_constant_mgr()->GetConstantFromInst(def_2);
+
+  assert(constant1 && constant2 && "The ids must refer to constants.");
+
   // If either constant is null, the other is equivalent iff it is zero-like
-  if (constant1.AsNullConstant()) {
-    return constant2.IsZero();
+  if (constant1->AsNullConstant()) {
+    return constant2->IsZero();
   }
 
-  if (constant2.AsNullConstant()) {
-    return constant1.IsZero();
+  if (constant2->AsNullConstant()) {
+    return constant1->IsZero();
   }
 
   // If the constants are scalar, they are equal iff their words are the same
-  if (auto scalar1 = constant1.AsScalarConstant()) {
-    return scalar1->words() == constant2.AsScalarConstant()->words();
+  if (auto scalar1 = constant1->AsScalarConstant()) {
+    return scalar1->words() == constant2->AsScalarConstant()->words();
   }
 
   // If the constants are composite, they are equivalent iff their components
   // match
-  if (constant1.AsCompositeConstant()) {
-    auto components1 = constant1.AsCompositeConstant()->GetComponents();
-    auto components2 = constant2.AsCompositeConstant()->GetComponents();
-
+  if (constant1->AsCompositeConstant()) {
     // Since the types match, we already know that the number of components is
-    // the same, so we just need to check equivalence for each one
-    for (size_t i = 0; i < components1.size(); i++) {
-      if (!AreEquivalentConstants(ir_context, *components1[i],
-                                  *components2[i])) {
+    // the same. We check that the input operands of the definitions are all
+    // constants and that they are pairwise equivalent.
+    for (uint32_t i = 0; i < def_1->NumInOperands(); i++) {
+      if (!AreEquivalentConstants(ir_context, def_1->GetSingleWordInOperand(i),
+                                  def_2->GetSingleWordInOperand(i))) {
         return false;
       }
     }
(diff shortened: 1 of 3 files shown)


source/fuzz/transformation_record_synonymous_constants.h
```diff
@@ -33,11 +33,14 @@ class TransformationRecordSynonymousConstants : public Transformation {
   //   of constants
   // - |message_.constant_id| and |message_.synonym_id| refer to constants
   //   that are equal or equivalent.
-  //   Two integers with the same width and value are equal, even if one is
-  //   signed and the other is not.
-  //   Constants are equivalent if both of them represent zero-like scalar
-  //   values of the same type (for example OpConstant of type int and value
-  //   0 and OpConstantNull of type int).
+  //   TODO: Fix this (Signed and unsigned integers are currently considered
+  //   non-equivalent)
```
afd, line 37: Please change to:

// TODO(url of GitHub issue): Signed and unsigned integers are currently considered non-equivalent

source/fuzz/transformation_record_synonymous_constants.h
```diff
@@ -33,11 +33,14 @@ class TransformationRecordSynonymousConstants : public Transformation {
...
+  //   Two integers with the same width and value are equal,
+  //   even if one is signed and the other is not. Constants are equivalent if
+  //   both of them represent zero-like values of the same type (for example
+  //   OpConstant of type int and value 0 and OpConstantNull of type int).
+  //   Composite constants are equivalent if they have the same type and their
+  //   components are pairwise equivalent.
```
afd, line 43: Can you work on the phrasing here - I couldn't understand it, partly because you don't close the "(".

test/fuzz/transformation_record_synonymous_constants_test.cpp
```diff
@@ -129,12 +129,14 @@ TEST(TransformationRecordSynonymousConstantsTest, IntConstants) {
   ApplyTransformationAndCheckFactManager(13, 22, context.get(),
                                          &transformation_context);
 
+  // TODO (https://github.com/KhronosGroup/SPIRV-Tools/issues/3536):
+  // Relax type check for integers.
```
afd, line 133: "TODO (" -> "TODO("

afd: Also say: "Uncomment this code once the issue is fixed."

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -19,8 +19,61 @@ namespace spvtools {
...
 namespace {
-bool IsScalarZeroConstant(const opt::analysis::Constant* constant) {
-  return constant->AsScalarConstant() && constant->IsZero();
+
+// Returns true if the two given constants are equivalent
+// (the description of IsApplicable specifies the conditions they must satisfy
+// to be considered equivalent)
+bool AreEquivalentConstants(opt::IRContext* ir_context,
+                            const opt::analysis::Constant& constant1,
+                            const opt::analysis::Constant& constant2) {
```
afd, line 28: Due to limitations of the constant manager and type manager, it is safer if we use the ids of the constants here - i.e. change constant1 and constant2 to be uint32_t, named constant_id1 and constant_id2, representing the ids of the constants.

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -19,8 +19,61 @@ namespace spvtools {
...
+  // Check that the type ids are the same
+  // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3536): Somehow
+  // relax this for integers (so that unsigned integer and signed integer are
+  // considered the same type)
+  uint32_t type_id1 = ir_context->get_type_mgr()->GetId(constant1.type());
+  uint32_t type_id2 = ir_context->get_type_mgr()->GetId(constant2.type());
```
afd, line 34: Change this so that it uses the def-use manager to get the instruction for each constant id, and then the type from that instruction.

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -19,8 +19,61 @@ namespace spvtools {
...
+  if (type_id1 != type_id2) {
+    return false;
+  }
+
+  // If either constant is null, the other is equivalent iff it is zero-like
```
afd, line 39: At this point, use the constant manager to get declared constants, constant1 and constant2, from the constant ids.

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -19,8 +19,61 @@ namespace spvtools {
...
+    auto components2 = constant2.AsCompositeConstant()->GetComponents();
+
+    // Since the types match, we already know that the number of components is
+    // the same, so we just need to check equivalence for each one
+    for (size_t i = 0; i < components1.size(); i++) {
+      if (!AreEquivalentConstants(ir_context, *components1[i],
+                                  *components2[i])) {
+        return false;
+      }
+    }
```
afd, line 66: At this point, revert back to looking at the constant instructions and check whether their input operands (all of which must be constant ids) are pairwise equivalent.

source/fuzz/transformation_record_synonymous_constants.h
stefanomil: Done

source/fuzz/transformation_record_synonymous_constants.h
stefanomil: I have changed it. See if it's clearer

test/fuzz/transformation_record_synonymous_constants_test.cpp
stefanomil: Done

source/fuzz/transformation_record_synonymous_constants.cpp
stefanomil: Done

source/fuzz/transformation_record_synonymous_constants.cpp
stefanomil: Done

source/fuzz/transformation_record_synonymous_constants.cpp
stefanomil: Done

source/fuzz/transformation_record_synonymous_constants.cpp
stefanomil: Done

stefanomil:
What should happen if the ids given to the transformations don't correspond to actual ids or to constant declarations? Right now I assert that they should, but I'm not sure.

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -19,8 +19,67 @@ namespace spvtools {
 namespace fuzz {
 
 namespace {
-bool IsScalarZeroConstant(const opt::analysis::Constant* constant) {
-  return constant->AsScalarConstant() && constant->IsZero();
+
+// Returns true if the two given constants are equivalent
+// (the description of IsApplicable specifies the conditions they must satisfy
+// to be considered equivalent)
```
afd, line 25: Move comment to the .h file - that's the place for top-level method comments.

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -19,8 +19,67 @@ namespace spvtools {
...
+  }
+
+  // If the constants are scalar, they are equal iff their words are the same
+  if (auto scalar1 = constant1->AsScalarConstant()) {
+    return scalar1->words() == constant2->AsScalarConstant()->words();
+  }
+
+  // If the constants are composite, they are equivalent iff their components
+  // match
+  if (constant1->AsCompositeConstant()) {
```
afd, line 63: Move the assertion from the bottom of the method to here, so that you assert constant1->AsCompositeConstant() holds.  Then get rid of the if statement and the need for the dummy return at the end of the method - i.e., following the assertion proceed as if the constant is a composite.

source/fuzz/transformation_record_synonymous_constants.cpp
```diff
@@ -19,8 +19,67 @@ namespace spvtools {
...
+      }
+    }
+
+    // If we get here, all the components are equivalent
+    return true;
+  }
+
+  // If we get here, the type is not sensible to check for equivalence
+  assert(false &&
+         "Equivalency of constants can only be checked with scalar, composite "
```
afd, line 80: Equivalency -> Equivalence

Also see comment above about moving this assertion.

source/fuzz/transformation_record_synonymous_constants.cpp
stefanomil: Done

source/fuzz/transformation_record_synonymous_constants.cpp
stefanomil: Done

source/fuzz/transformation_record_synonymous_constants.cpp
stefanomil: Done. I made the method a static member method and moved the comment to the declaration in the header file.

Issue #3530, opened 2020-07-14
spirv-fuzz: Helper function to apply transformations only if they are applicable


stefanomil:
Sometimes, fuzzer passes need to apply a transformation only after checking that they are applicable (rather than asserting that they are, as happens in ApplyTransformation).

It would be useful to have a helper function MaybeApplyTransformation to do that and use it in all places where this pattern is used.

PR #3540, opened 2020-07-15
spirv-fuzz: Implement MaybeApplyTransformation helper function in fuzzer_pass.h

stefanomil:
This function can be used to apply a transformation only if it is applicable and use it wherever this pattern is used (see issue #3530 ).

### DIFF
--- a/source/fuzz/fuzzer_pass_merge_blocks.cpp
+++ b/source/fuzz/fuzzer_pass_merge_blocks.cpp
@@ -56,11 +56,7 @@ void FuzzerPassMergeBlocks::Apply() {
     uint32_t index = GetFuzzerContext()->RandomIndex(potential_transformations);
     auto transformation = potential_transformations.at(index);
     potential_transformations.erase(potential_transformations.begin() + index);
-    if (transformation.IsApplicable(GetIRContext(),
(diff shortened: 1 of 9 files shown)


stefanomil:
Some parts of the code implement a similar pattern, but then go on to perform other actions if the transformation was applicable.

Would it be useful to make the MaybeApplyTransformation function return true if the transformation was actually applied so that it can also be used in these situations?

source/fuzz/fuzzer_pass_merge_blocks.cpp
```diff
@@ -56,11 +56,7 @@ void FuzzerPassMergeBlocks::Apply() {
     uint32_t index = GetFuzzerContext()->RandomIndex(potential_transformations);
     auto transformation = potential_transformations.at(index);
     potential_transformations.erase(potential_transformations.begin() + index);
```
afd, line 58: @stefanomil separate from this PR, could you see whether "RemoveAtRandomIndex" from "FuzzerContext" could be used to make this piece of code simpler, and do a little PR for it if so?

stefanomil: I will take a look

PR #3541, opened 2020-07-15
Fix reachability in the validator

### LINKED ISSUE #3529: spirv-fuzz: TransformationMoveBlockDown invalidates the module
Consider the following file.
[shader.txt](https://github.com/KhronosGroup/SPIRV-Tools/files/4919001/shader1.txt)

It contains a perfectly valid SPIR-V (i.e. the following test succeeds).
```cpp
TEST(A, a) {
  std::string shader; // code from the file

  const auto env = SPV_ENV_UNIVERSAL_1_3;
  const auto consumer = nullptr;
  const auto context = BuildModule(env, consumer, shader, kFuzzAssembleOption);
  ASSERT_TRUE(IsValid(env, context.get()));
}
```

However, if we try to move a basic block with id `370` one block down, the transformation will be applied correctly (i.e. `TransformationMoveBlockDown::IsApplicable` returns `true`) but the `spirv-val` will fail with with a somewhat ambiguous message:
```
error: line 121: Case construct that targets 310[%310] has invalid branch to block 414[%414] (not another case construct, corresponding merge, outer loop merge or outer loop continue)
  %310 = OpLabel
```

Commit id is 7afbc0c8.

alan-baker:
Fixes #3529

* Make BasicBlock::reachable() only consider static reachability
* Fix reachability calculation to be independent of block order
* add tests

Also tested against CTS.

### DIFF
--- a/source/val/function.h
+++ b/source/val/function.h
@@ -97,7 +97,6 @@ class Function {
   /// Registers the end of the block
   ///
   /// @param[in] successors_list A list of ids to the block's successors
-  /// @param[in] branch_instruction the branch instruction that ended the block
   void RegisterBlockEnd(std::vector<uint32_t> successors_list);
 
   /// Registers the end of the function.  This is idempotent.
--- a/source/val/function.h
+++ b/source/val/function.h
@@ -98,8 +98,7 @@ class Function {
   ///
   /// @param[in] successors_list A list of ids to the block's successors
   /// @param[in] branch_instruction the branch instruction that ended the block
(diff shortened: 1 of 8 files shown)


source/val/function.h
```diff
@@ -98,8 +98,7 @@ class Function {
   ///
   /// @param[in] successors_list A list of ids to the block's successors
   /// @param[in] branch_instruction the branch instruction that ended the block
```
dneto0, line 100: nit: Delete the branch_instruction comment too.

PR #3542, opened 2020-07-15
Preserve OpenCL.DebugInfo.100 through elim-dead-code-aggressive

greg-lunarg:
Essentially, it marks all DebugInfo instructions in functions (and their operands) as live. It treats DebugDeclare and DebugValue with Deref as loads and so marks Stores of their variables as live.

It marks each DebugGlobalVariables as live except for its variable. After closure, it rechecks if the variable is live. If not, the DebugGlobalVariable instruction's variable operand is set to DebugInfoNone, per the DebugInfo spec.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.cpp
+++ b/source/opt/aggressive_dead_code_elim_pass.cpp
@@ -38,6 +38,8 @@ const uint32_t kLoopMergeMergeBlockIdInIdx = 0;
 const uint32_t kLoopMergeContinueBlockIdInIdx = 1;
 const uint32_t kCopyMemoryTargetAddrInIdx = 0;
 const uint32_t kCopyMemorySourceAddrInIdx = 1;
+const uint32_t kDebugDeclareOperandVariableIndex = 5;
+const uint32_t kGlobalVariableVariableIndex = 12;
 
 // Sorting functor to present annotation instructions in an easy-to-process
 // order. The functor orders by opcode first and falls back on unique id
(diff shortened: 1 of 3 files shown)


jaebaek:
Thanks @greg-lunarg. I will take a look!

greg-lunarg:
FWIW, I did spend some time thinking about legalization.

There are two different HLSL constructs that are problematic: local structured buffers and local structs with opaque types (eg textures and/or samplers). You can find actual problematic HLSL shaders in this GDC talk I presented in:

https://www.khronos.org/assets/uploads/developers/library/2018-gdc-webgl-and-gltf/2-Vulkan-HLSL-There-and-Back-Again_Mar18.pdf

I couldn't get dxc to cooperate* in creating a completely perfect example of what ADCE will see for these, so I did a thought experiment instead.

I came to the conclusion that if the passes that precede ADCE in the legalization recipe did what I expected them to do for these shaders, that ADCE would eliminate the problematic local construct from the SPIR-V. But it would still leave the DebugInfo for the original HLSL construct so that the user could query in terms of the original shader.

*I could not force dxc to create the invalid SPIR-V for these shaders; it would just not generate a SPIR-V file and would give no error message. Ultimately we will need to double back and make sure these shaders are handled correctly when all the pieces are in place.

jaebaek:
It looks mostly good but I want to check it with your legalization example.

```
struct os {
SamplerState o_s2D;
Texture2D o_tex;
};
Texture2D tex;
SamplerState s2D;
float4 osCall(os s, float2 f2)
{
return s.o_tex.Sample(s.o_s2D, f2);
}
float4 main() : SV_TARGET0
{
os s;
s.o_tex = tex;
s.o_s2D = s2D;
return osCall(s, float2(0.2, 0.3));
}
```

I found a few bugs for the DXC's debug info generation while testing it.
I will get back to you after fixing them.

greg-lunarg:
The desirable result for this legalization example will depend on what scalar-replacement does.

jaebaek:
I agree. I just tested the above code with the consideration that we do not have the scalar-replacement. It works fine even without the scalar-replacement.

The case I am worried is that we have DebugDeclare with invalid OpVariable or OpFunctionParameter. I think it would be fine because ssa-rewrite will replace the OpVariable to its value correctly using DebugValue. In addition, after function inlining pass, we will not have OpFunctionParameter.

It looks good to me.

PR #3546, opened 2020-07-16
spirv-fuzz: Generalise transformation access chain

### LINKED ISSUE #3179: spirv-fuzz: Transformation to add access chain initially restricted to use literal indices
For simplicity, the original version of the transformation to add access chain instructions to a module will be restricted to using access chain indices that use in-bounds literals.  In the future it would be nice to generalise this so that non-struct indices can be dynamic and clamped to be in-bounds.

stefanomil:
This PR generalises TransformationAddAccessChain so that dynamic
indices for non-struct composites (with clamping to ensure that
accesses are in-bound) are allowed.

The transformation will add instructions to clamp any index to
a non-struct composite, regardless of whether it is a constant
or not.

Fixes #3179

### DIFF
--- a/source/fuzz/fuzzer_pass.h
+++ b/source/fuzz/fuzzer_pass.h
@@ -179,7 +179,7 @@ class FuzzerPass {
   // on the |is_irrelevant| parameter.
   uint32_t FindOrCreateIntegerConstant(const std::vector<uint32_t>& words,
                                        uint32_t width, bool is_signed,
-                                       bool is_irrelevant);
+                                       bool is_irrelevant = false);
 
   // Returns the id of an OpConstant instruction, with a floating-point
   // type of width specified by |width|, with |words| as its value.  If either
--- a/source/fuzz/fuzzer_pass.h
+++ b/source/fuzz/fuzzer_pass.h
@@ -175,11 +175,10 @@ class FuzzerPass {
   // with |words| as its value.  If either the required integer type or the
   // constant do not exist, transformations are applied to add them.
   // The returned id either participates in IdIsIrrelevant fact or not,
-  // depending
-  // on the |is_irrelevant| parameter.
+  // depending on the |is_irrelevant| parameter.
   uint32_t FindOrCreateIntegerConstant(const std::vector<uint32_t>& words,
                                        uint32_t width, bool is_signed,
-                                       bool is_irrelevant = false);
+                                       bool is_irrelevant);
 
   // Returns the id of an OpConstant instruction, with a floating-point
   // type of width specified by |width|, with |words| as its value.  If either
--- a/source/fuzz/fuzzer_pass_add_access_chains.cpp
+++ b/source/fuzz/fuzzer_pass_add_access_chains.cpp
@@ -140,9 +140,6 @@ void FuzzerPassAddAccessChains::Apply() {
           uint32_t index_value =
               GetFuzzerContext()->GetRandomIndexForAccessChain(bound);
 
-          index_ids.push_back(FindOrCreateIntegerConstant(
-              {index_value}, 32, GetFuzzerContext()->ChooseEven(), false));
-
           switch (subobject_type->opcode()) {
             case SpvOpTypeArray:
             case SpvOpTypeMatrix:
--- a/source/fuzz/fuzzer_pass_add_access_chains.cpp
+++ b/source/fuzz/fuzzer_pass_add_access_chains.cpp
@@ -146,12 +146,28 @@ void FuzzerPassAddAccessChains::Apply() {
           switch (subobject_type->opcode()) {
             case SpvOpTypeArray:
             case SpvOpTypeMatrix:
-            case SpvOpTypeVector:
+            case SpvOpTypeVector: {
               // The index will be clamped
-              index_ids.push_back(FindOrCreateIntegerConstantReadyForClamping(
-                  index_value, bound, &fresh_ids_for_clamping));
+
+              bool is_signed = GetFuzzerContext()->ChooseEven();
+
+              // Make the constant ready for clamping. We need:
+              // - an OpTypeBool to be present in the module
+              // - an OpConstant with the same type as the index and value
+              //   the maximum value for an index
+              // - a new pair of fresh ids for the clamping instructions
+              FindOrCreateBoolType();
+              FindOrCreateIntegerConstant({bound - 1}, 32, is_signed);
+              std::pair<uint32_t, uint32_t> fresh_pair_of_ids = {
+                  GetFuzzerContext()->GetFreshId(),
+                  GetFuzzerContext()->GetFreshId()};
+              fresh_ids_for_clamping.emplace_back(fresh_pair_of_ids);
+
+              index_ids.push_back(
+                  FindOrCreateIntegerConstant({index_value}, 32, is_signed));
               subobject_type_id = subobject_type->GetSingleWordInOperand(0);
-              break;
+
+            } break;
             case SpvOpTypeStruct:
               index_ids.push_back(FindOrCreateIntegerConstant(
                   {index_value}, 32, GetFuzzerContext()->ChooseEven()));
@@ -173,22 +189,6 @@ void FuzzerPassAddAccessChains::Apply() {
             GetFuzzerContext()->GetFreshId(), chosen_pointer->result_id(),
             index_ids, instruction_descriptor, fresh_ids_for_clamping));
       });
-}
-
-uint32_t FuzzerPassAddAccessChains::FindOrCreateIntegerConstantReadyForClamping(
-    uint32_t value, uint32_t bound,
-    std::vector<std::pair<uint32_t, uint32_t>>* fresh_ids_for_clamping) {
-  // Choose the signedness of the constant
-  bool is_signed = GetFuzzerContext()->ChooseEven();
-
-  // Make the constant ready for clamping
-  FindOrCreateBoolType();
-  FindOrCreateIntegerConstant({bound - 1}, 32, is_signed);
-  std::pair<uint32_t, uint32_t> fresh_pair_of_ids = {
-      GetFuzzerContext()->GetFreshId(), GetFuzzerContext()->GetFreshId()};
-  fresh_ids_for_clamping->emplace_back(fresh_pair_of_ids);
-
-  return FindOrCreateIntegerConstant({value}, 32, is_signed);
 }
 
 }  // namespace fuzz
--- a/source/fuzz/fuzzer_pass_add_access_chains.cpp
+++ b/source/fuzz/fuzzer_pass_add_access_chains.cpp
@@ -154,20 +154,20 @@ void FuzzerPassAddAccessChains::Apply() {
               //   the maximum value for an index
               // - a new pair of fresh ids for the clamping instructions
               FindOrCreateBoolType();
-              FindOrCreateIntegerConstant({bound - 1}, 32, is_signed);
+              FindOrCreateIntegerConstant({bound - 1}, 32, is_signed, false);
               std::pair<uint32_t, uint32_t> fresh_pair_of_ids = {
                   GetFuzzerContext()->GetFreshId(),
                   GetFuzzerContext()->GetFreshId()};
               fresh_ids_for_clamping.emplace_back(fresh_pair_of_ids);
 
-              index_ids.push_back(
-                  FindOrCreateIntegerConstant({index_value}, 32, is_signed));
+              index_ids.push_back(FindOrCreateIntegerConstant(
+                  {index_value}, 32, is_signed, false));
               subobject_type_id = subobject_type->GetSingleWordInOperand(0);
 
             } break;
             case SpvOpTypeStruct:
               index_ids.push_back(FindOrCreateIntegerConstant(
-                  {index_value}, 32, GetFuzzerContext()->ChooseEven()));
+                  {index_value}, 32, GetFuzzerContext()->ChooseEven(), false));
               subobject_type_id =
                   subobject_type->GetSingleWordInOperand(index_value);
               break;
--- a/source/fuzz/fuzzer_pass_add_access_chains.cpp
+++ b/source/fuzz/fuzzer_pass_add_access_chains.cpp
@@ -92,6 +92,11 @@ void FuzzerPassAddAccessChains::Apply() {
             relevant_pointer_instructions[GetFuzzerContext()->RandomIndex(
                 relevant_pointer_instructions)];
         std::vector<uint32_t> index_ids;
+
+        // Each index accessing a non-struct composite will be clamped, thus
+        // needing a pair of fresh ids
+        std::vector<std::pair<uint32_t, uint32_t>> fresh_ids_for_clamping;
+
         auto pointer_type = GetIRContext()->get_def_use_mgr()->GetDef(
             chosen_pointer->type_id());
         uint32_t subobject_type_id = pointer_type->GetSingleWordInOperand(1);
@@ -132,20 +137,24 @@ void FuzzerPassAddAccessChains::Apply() {
             break;
           }
 
-          // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3179) We
-          //  could allow non-constant indices when looking up non-structs,
-          //  using clamping to ensure they are in-bounds.
           uint32_t index_value =
               GetFuzzerContext()->GetRandomIndexForAccessChain(bound);
+
           index_ids.push_back(FindOrCreateIntegerConstant(
               {index_value}, 32, GetFuzzerContext()->ChooseEven(), false));
+
           switch (subobject_type->opcode()) {
             case SpvOpTypeArray:
             case SpvOpTypeMatrix:
             case SpvOpTypeVector:
+              // The index will be clamped
+              index_ids.push_back(FindOrCreateIntegerConstantReadyForClamping(
+                  index_value, bound, &fresh_ids_for_clamping));
               subobject_type_id = subobject_type->GetSingleWordInOperand(0);
               break;
             case SpvOpTypeStruct:
+              index_ids.push_back(FindOrCreateIntegerConstant(
+                  {index_value}, 32, GetFuzzerContext()->ChooseEven()));
               subobject_type_id =
                   subobject_type->GetSingleWordInOperand(index_value);
               break;
@@ -162,8 +171,24 @@ void FuzzerPassAddAccessChains::Apply() {
         // Apply the transformation to add an access chain.
         ApplyTransformation(TransformationAccessChain(
             GetFuzzerContext()->GetFreshId(), chosen_pointer->result_id(),
-            index_ids, instruction_descriptor));
+            index_ids, instruction_descriptor, fresh_ids_for_clamping));
       });
+}
+
+uint32_t FuzzerPassAddAccessChains::FindOrCreateIntegerConstantReadyForClamping(
+    uint32_t value, uint32_t bound,
+    std::vector<std::pair<uint32_t, uint32_t>>* fresh_ids_for_clamping) {
+  // Choose the signedness of the constant
+  bool is_signed = GetFuzzerContext()->ChooseEven();
+
+  // Make the constant ready for clamping
+  FindOrCreateBoolType();
+  FindOrCreateIntegerConstant({bound - 1}, 32, is_signed);
+  std::pair<uint32_t, uint32_t> fresh_pair_of_ids = {
+      GetFuzzerContext()->GetFreshId(), GetFuzzerContext()->GetFreshId()};
+  fresh_ids_for_clamping->emplace_back(fresh_pair_of_ids);
+
+  return FindOrCreateIntegerConstant({value}, 32, is_signed);
 }
 
 }  // namespace fuzz
--- a/source/fuzz/fuzzer_pass_add_access_chains.h
+++ b/source/fuzz/fuzzer_pass_add_access_chains.h
@@ -33,21 +33,6 @@ class FuzzerPassAddAccessChains : public FuzzerPass {
   ~FuzzerPassAddAccessChains();
 
   void Apply() override;
-
- private:
-  // Returns the id of a 32-bit integer constant of value |value|, which is
-  // randomly chosen to be signed or unsigned. If such constant already exists
-  // in the module, it is found, otherwise a new one is created.
-  // Adds instructions to the module so that clamping can be correctly
-  // performed. In particular, it makes sure (by adding them if not present)
-  // that the module has:
-  // - An OpTypeBool instruction
-  // - An OpConstant of the same type as the one whose id is returned, and value
-  //   |bound| - 1
-  // Adds a pair of fresh ids to |fresh_ids_for_clamping|.
-  uint32_t FindOrCreateIntegerConstantReadyForClamping(
-      uint32_t value, uint32_t bound,
-      std::vector<std::pair<uint32_t, uint32_t>>* fresh_ids_for_clamping);
 };
 
 }  // namespace fuzz
--- a/source/fuzz/fuzzer_pass_add_access_chains.h
+++ b/source/fuzz/fuzzer_pass_add_access_chains.h
@@ -33,6 +33,21 @@ class FuzzerPassAddAccessChains : public FuzzerPass {
   ~FuzzerPassAddAccessChains();
 
   void Apply() override;
+
+ private:
+  // Returns the id of a 32-bit integer constant of value |value|, which is
+  // randomly chosen to be signed or unsigned. If such constant already exists
+  // in the module, it is found, otherwise a new one is created.
+  // Adds instructions to the module so that clamping can be correctly
+  // performed. In particular, it makes sure (by adding them if not present)
+  // that the module has:
+  // - An OpTypeBool instruction
+  // - An OpConstant of the same type as the one whose id is returned, and value
+  //   |bound| - 1
+  // Adds a pair of fresh ids to |fresh_ids_for_clamping|.
+  uint32_t FindOrCreateIntegerConstantReadyForClamping(
+      uint32_t value, uint32_t bound,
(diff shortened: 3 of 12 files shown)


source/fuzz/transformation_access_chain.h
```diff
@@ -68,11 +79,36 @@ class TransformationAccessChain : public Transformation {
...
+  // result is smaller than the given bound. The |fresh_ids| are used to
+  // add new instructions to the module, to perform such operations.
+  // The instructions are inserted just before
+  // |message_.instruction_to_insert_before|.
+  // Returns false if a constant with value |bound|-1 is not found in the
+  // module or if the bool type is not found in the module, true otherwise.
+  bool TryToClampIntVariable(opt::IRContext* ir_context,
+                             const opt::Instruction& int_inst, uint32_t bound,
+                             std::pair<uint32_t, uint32_t> fresh_ids,
+                             bool add_clamping_instructions) const;
```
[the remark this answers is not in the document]
paulthomson: Also, this function documentation comment should start with "Tries...", "Clamps...", "Returns..." or similar. Here is how the Google C++ style guide says it:

These comments should open with descriptive verbs in the indicative mood ("Opens the file") rather than verbs in the imperative ("Open the file").

source/fuzz/transformation_access_chain.h
```diff
@@ -68,11 +79,36 @@ class TransformationAccessChain : public Transformation {
...
-                                          uint32_t index_id) const;
+  // Returns {false, 0} if:
+  // - |index_id| does not correspond to a 32-bit integer
+  // - the object being indexed is not a composite type
+  // - the object type is a struct and the index is not a constant or
+  //   the index is out of bounds
+  // - it is not possible to clamp the index variable to the bound
+  // Otherwise, returns:
+  // - {true, value} if the object being indexed is a struct (and the index
+  //   is thus a constant), where value is the value of the constant
```
paulthomson, line 90: I would remove "thus" and the parentheses because the function does not assume the index is a constant; it checks this and will return false otherwise. I guess this is already implied by the above.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -195,22 +339,118 @@ protobufs::Transformation TransformationAccessChain::ToMessage() const {
...
+
+    if (value >= bound) {
+      return {false, 0};
+    }
+
+    return {true, value};
+  }
+
+  // The object being traverse is a struct. We must clamp the index.
+
```
paulthomson, line 387: I think this is incorrect?

source/fuzz/transformation_access_chain.h
```diff
@@ -68,11 +79,36 @@ class TransformationAccessChain : public Transformation {
...
+  // - {true, clamped_id} if the object being indexed is not a struct, where
+  //   clamped_id is the id at which to find the clamped index id
+  // This method only modifies the module if add_clamping_instructions is true.
+  // |fresh_ids| contains the fresh ids needed for clamping, and it is ignored
+  // if clamping is not needed (i.e. the object is a struct).
+  std::pair<bool, uint32_t> GetIndexValueOrId(
+      opt::IRContext* ir_context, uint32_t index_id, uint32_t object_type_id,
+      bool add_clamping_instructions = false,
+      std::pair<uint32_t, uint32_t> fresh_ids = {0, 0}) const;
```
paulthomson, line 99: I suggest splitting this into two functions; one for the clamping case, one for the non-clamping case. The clamping case could be renamed to something like `CreateAndGetClampedIndexId` so that it is obvious that we are not just "getting" something that already exists.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -409,6 +409,9 @@ message TransformationAccessChain {
   // OpAccessChain instruction should be inserted
   InstructionDescriptor instruction_to_insert_before = 4;
 
+  // Additional fresh ids, required to clamp index variables
+  repeated UInt32Pair fresh_ids_for_clamping = 5;
```
paulthomson, line 413: I would clarify what each id is used for. In fact, I would add a large comment either here, or just below `message TransformationAccessChain` explaining how clamping instructions may also be added.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -22,21 +22,82 @@
...
+  auto int_type_inst = ir_context->get_def_use_mgr()->GetDef(int_type_id);
+
+  assert(int_type_inst && "The given type id must exist.");
+
+  auto int_type = ir_context->get_type_mgr()
+                      ->GetType(int_type_inst->result_id())
+                      ->AsInteger();
+
+  assert(int_type && "The given type id must correspond to an integer type.");
+
```
paulthomson, line 40: Could put this in `fuzzer_util.cpp`? If so, change the name of `bound_minus_one`. Documentation comment should start with "Returns...", "Finds...", or similar. Make sure you also document the assertions; i.e. the "requirements" of this function. E.g.

Returns the id of an integer constant with type |int_type_id| and value |value|, or 0 if no such constant exists in the module. |int_type_id| must exist in the module and must correspond to an integer type id.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -22,21 +22,82 @@
...
+      ->GetDefiningInstruction(&bound_minus_one)
+      ->result_id();
+}
+
+// TODO: This is copied from TransformationAddFunction. Move it somewhere
+// where it can be accessed by both.
+uint32_t GetBoundForCompositeIndex(
+    opt::IRContext* ir_context, const opt::Instruction& composite_type_inst) {
+  switch (composite_type_inst.opcode()) {
+    case SpvOpTypeArray:
```
paulthomson, line 58: I think fuzzer_util.cpp would be a good place for this as well.

source/fuzz/transformation_access_chain.h
```diff
@@ -68,11 +79,36 @@ class TransformationAccessChain : public Transformation {
...
-  // constant.  Otherwise, returns {true, value}, where value is the value of
-  // the 32-bit integer constant to which |index_id| corresponds.
-  std::pair<bool, uint32_t> GetIndexValue(opt::IRContext* ir_context,
-                                          uint32_t index_id) const;
+  // Returns {false, 0} if:
+  // - |index_id| does not correspond to a 32-bit integer
+  // - the object being indexed is not a composite type
+  // - the object type is a struct and the index is not a constant or
+  //   the index is out of bounds
+  // - it is not possible to clamp the index variable to the bound
```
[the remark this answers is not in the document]
stefanomil: Fixed

source/fuzz/transformation_access_chain.cpp
```diff
@@ -96,23 +157,72 @@ bool TransformationAccessChain::IsApplicable(
   // Start from the base type of the pointer.
   uint32_t subobject_type_id = pointer_type->GetSingleWordInOperand(1);
 
+  int id_pairs_used = 0;
+
+  // Keep track of the fresh ids used to make sure that they are distinct
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_access_chain.h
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_access_chain.h
stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Yes!

Fixed

source/fuzz/transformation_access_chain.h
stefanomil: Done

source/fuzz/protobufs/spvtoolsfuzz.proto
stefanomil: Done. Check that it's clear.

source/fuzz/transformation_access_chain.cpp
stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Moved it there, removed it from both TransformationAccessChain and TransformationAddFunction and renamed all the references.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -413,6 +413,14 @@ message TransformationAccessChain {
 
   // Adds an access chain instruction based on a given pointer and indices.
 
+  // When accessing a struct, the corresponding indices must be 32-bit integer constants.
+  // For any other composite, the indices can be any 32-bit integer, and the transformation
+  // adds two instructions for each such index to clamp it to the bound, as follows:
+  //
+  // %t1 = OpULessThanEqual %bool %index %bound_minus_one
+  // %t2 = OpSelect %int_type %t1 %index %bound_minus_one
+
```
afd, line 422: Remove extraneous blank line (one blank line after comment for the transformation and before comment for the first field).

source/fuzz/fuzzer_util.h
```diff
@@ -367,6 +374,14 @@ uint32_t MaybeGetIntegerConstant(
     const std::vector<uint32_t>& words, uint32_t width, bool is_signed,
     bool is_irrelevant = false);
 
+// Returns the id of an integer constant in the module with type |int_type_id|
+// and value |value|, or 0 if no such constant exists in the module.
+// |int_type_id| must exist in the module and it must correspond to an integer
+// type.
```
afd, line 380: I suppose it should correspond to a 32-bit integer type for this to make sense; can you adapt the comment to clarify that?

source/fuzz/fuzzer_pass_add_access_chains.cpp
```diff
@@ -132,20 +137,21 @@ void FuzzerPassAddAccessChains::Apply() {
...
-          index_ids.push_back(FindOrCreateIntegerConstant(
-              {index_value}, 32, GetFuzzerContext()->ChooseEven()));
+
           switch (subobject_type->opcode()) {
             case SpvOpTypeArray:
             case SpvOpTypeMatrix:
             case SpvOpTypeVector:
+              // The index will be clamped
+              index_ids.push_back(FindOrCreateIntegerConstantReadyForClamping(
+                  index_value, bound, &fresh_ids_for_clamping));
```
afd, line 149: I actually think it's a bad idea to introduce the function FindOrCreateIntegerConstantReadyForClamping.

- The function is not very cohesive.  It adds a boolean type, adds an integer constant, returns the id of that constant, and then - on the side - creates a pair of fresh ids.

- The function is only called once.  Thus there's no argument for avoiding code duplication in the fuzzer pass.

- The function is very unlikely to be reusable, due to it being very specific to this fuzzer pass and not very cohesive overall.

- The body of the function would arguably be easier to understand if presented inline, here, rather than on its own.

Can you present the code inline and get rid of the function?

source/fuzz/fuzzer_pass_add_access_chains.cpp
```diff
@@ -162,9 +168,25 @@ void FuzzerPassAddAccessChains::Apply() {
...
-            index_ids, instruction_descriptor));
+            index_ids, instruction_descriptor, fresh_ids_for_clamping));
       });
 }
 
+uint32_t FuzzerPassAddAccessChains::FindOrCreateIntegerConstantReadyForClamping(
+    uint32_t value, uint32_t bound,
+    std::vector<std::pair<uint32_t, uint32_t>>* fresh_ids_for_clamping) {
+  // Choose the signedness of the constant
+  bool is_signed = GetFuzzerContext()->ChooseEven();
```
afd, line 179: When you make this inline can you also avoid this temporary and use the "ChooseEven" experession as the actual parameter for FindOrCreateIntegerConstant?

source/fuzz/fuzzer_pass_add_access_chains.h
```diff
@@ -33,6 +33,21 @@ class FuzzerPassAddAccessChains : public FuzzerPass {
...
+  // Returns the id of a 32-bit integer constant of value |value|, which is
+  // randomly chosen to be signed or unsigned. If such constant already exists
+  // in the module, it is found, otherwise a new one is created.
+  // Adds instructions to the module so that clamping can be correctly
+  // performed. In particular, it makes sure (by adding them if not present)
+  // that the module has:
+  // - An OpTypeBool instruction
+  // - An OpConstant of the same type as the one whose id is returned, and value
+  //   |bound| - 1
+  // Adds a pair of fresh ids to |fresh_ids_for_clamping|.
```
afd, line 47: The fact that there's a lot going on in this method comment suggests that the method is perhaps not very cohesive enough.  Sometimes complex methods with a comment like this can't be avoided, but if you're extracting code into a method and find yourself writing such a detailed description of what the method does that might be a sign that the code would be clearer if left inline.

source/fuzz/fuzzer_util.cpp
```diff
@@ -1002,6 +1024,31 @@ uint32_t MaybeGetIntegerConstant(
...
+                                                 uint32_t int_type_id) {
+  auto int_type_inst = ir_context->get_def_use_mgr()->GetDef(int_type_id);
+
+  assert(int_type_inst && "The given type id must exist.");
+
+  auto int_type = ir_context->get_type_mgr()
+                      ->GetType(int_type_inst->result_id())
+                      ->AsInteger();
+
+  assert(int_type && "The given type id must correspond to an integer type.");
```
afd, line 1038: If you do change the method contract to require that the integer is 32-bit, it could be asserted here too.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -196,21 +280,149 @@ protobufs::Transformation TransformationAccessChain::ToMessage() const {
...
+
+  if (value >= bound) {
+    return {false, 0};
+  }
+
+  return {true, value};
+}
+
+std::pair<bool, uint32_t> TransformationAccessChain::CreateAndGetClampedIndexId(
+    opt::IRContext* ir_context, uint32_t index_id, uint32_t object_type_id,
```
afd, line 309: I am not keen on this function, nor on TryToClampInteger.  They are complex because they capture both the *checking* that is necessary in IsApplicable, and the *doing* that should only happen in Apply.

I suggest you approach this as follows:

- In "IsApplicable", instead of calling CreateAndGetClampedIndexId, inline all of the checking code so that "IsApplicable" does all of the checks inline.  Similarly, don't call TryToClampInteger - write the code out inline, even if it gets quite long.

- Do the same for "Apply", but only include the "doing" code.  Remember that "Apply" can assume that "IsApplicable" returned true, so e.g. if you need bool type you can use "MaybeGet" to get it, and assert that the result of this call is non-zero; you don't need to check whether it could be found and return false if it wasn't.

Then see how the code looks.  There may well be a bit more code overall, but I expect it will be more readable since in the "IsApplicable" code there will be no logic for changing the module - and no need for a "add_clamping_instructions" to control whether we want to make changes.  Similarly in "Apply" there will be no checking code (just some sanity-checking assertions perhaps).

You might then decide you want to factor some code out into helper functions for readability - but no need to share the helpers between "IsApplicable" and "Apply" unless you think it really makes sense to do so.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -196,21 +280,149 @@ protobufs::Transformation TransformationAccessChain::ToMessage() const {
...
+      fuzzerutil::GetBoundForCompositeIndex(*object_type_def, ir_context);
+
+  if (!TryToClampInteger(ir_context, *index_instruction, bound, fresh_ids,
+                         add_clamping_instructions)) {
+    // It was not possible to clamp the integer
+    return {false, 0};
+  }
+
+  // The clamped integer will be at id |fresh_ids[1]|.
+  return {true, fresh_ids.second};
```
afd, line 342: The fact that - regardless of the rest of the computation - this function returns something deterministic as the second component in its pair - isn't good.

I suggest (as per the above comment) that you get rid of this function in any case.

But if you were to keep it, it might be better to make it simply return a boolean, and to have the "Apply" code that invokes it know that fresh_ids.second is the result id.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -196,21 +280,149 @@ protobufs::Transformation TransformationAccessChain::ToMessage() const {
...
+
+  // The composite must not be a struct
+  auto object_type_def = ir_context->get_def_use_mgr()->GetDef(object_type_id);
+
+  if (object_type_def->opcode() == SpvOpTypeStruct) {
+    return {false, 0};
+  }
+
+  // Valid fresh ids need to have been given
+  if (fresh_ids.first <= 0 || fresh_ids.second <= 0) {
```
afd, line 324: I don't like comparing unsigned values against 0 using <=, since the < part of <= can never be true.  I suggest you use ==.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -196,21 +280,149 @@ protobufs::Transformation TransformationAccessChain::ToMessage() const {
...
+  if (!ValidIndexToComposite(ir_context, index_id, object_type_id)) {
     return {false, 0};
   }
+
+  // The composite must not be a struct
+  auto object_type_def = ir_context->get_def_use_mgr()->GetDef(object_type_id);
+
+  if (object_type_def->opcode() == SpvOpTypeStruct) {
+    return {false, 0};
+  }
```
afd, line 321: Example: in the "Apply" path this could be left out or asserted.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -196,21 +280,149 @@ protobufs::Transformation TransformationAccessChain::ToMessage() const {
...
+
+  // Valid fresh ids need to have been given
+  if (fresh_ids.first <= 0 || fresh_ids.second <= 0) {
+    return {false, 0};
+  }
```
afd, line 326: Again, in the "Apply" path this could be left out or asserted.

source/fuzz/fuzzer_pass_add_access_chains.cpp
stefanomil: I think this would be a problem, because the constant of bounds (bound - 1) needs to have the same type as the index constant (and so the same signedness), so we need a way to make sure that the two are the same (the temporary is used twice)

afd: Oh - I didn't spot that it was used twice; please ignore the comment then.

source/fuzz/fuzzer_pass_add_access_chains.h
stefanomil: I am worried that, since this function is used both in the IsApplicable method and in the Apply method, this would create quite a bit of duplication. Do you think this amount of duplication would be fine?

source/fuzz/protobufs/spvtoolsfuzz.proto
stefanomil: Done

source/fuzz/fuzzer_util.h
stefanomil: Done

source/fuzz/fuzzer_pass_add_access_chains.cpp
stefanomil: Done

source/fuzz/fuzzer_pass_add_access_chains.h
stefanomil: I inlined it (and CreateAndGetClampedIndexId)

source/fuzz/fuzzer_util.cpp
stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Done. The duplication in the code actually turned out to be much less than expected, because of all the checks that only IsApplicable needs to do and the actions that only Apply needs to perform.

source/fuzz/transformation_access_chain.cpp
stefanomil: Good point. I inlined the function so this is no longer relevant in this specific case.

source/fuzz/transformation_access_chain.cpp
stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Left out

source/fuzz/transformation_access_chain.cpp
stefanomil: Left out

source/fuzz/transformation_access_chain.cpp
```diff
@@ -136,24 +136,50 @@ bool TransformationAccessChain::IsApplicable(
...
-                                      false, {pair.first(), pair.second()})
-               .first) {
+      if (!ValidIndexToComposite(ir_context, index_id, subobject_type_id)) {
         return false;
       }
 
+      // Valid fresh ids need to have been given
+      if (fresh_ids.first() == 0 || fresh_ids.second() == 0) {
+        return false;
+      }
```
afd, line 158: I suggest moving this up to line 141 - right after you get the fresh ids, and change the comment to "Valid ids need to have been given".

source/fuzz/transformation_access_chain.cpp
```diff
@@ -136,24 +136,50 @@ bool TransformationAccessChain::IsApplicable(
...
+
+      // Perform the clamping using the fresh ids at our disposal.
+      // The module will not be changed if |add_clamping_instructions| is not
+      // set.
```
afd, line 162: Change / get rid of this comment; it still talks about |add_clamping_instructions|.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -136,24 +136,50 @@ bool TransformationAccessChain::IsApplicable(
...
+      if (!fuzzerutil::MaybeGetIntegerConstantFromValueAndType(
+              ir_context, bound - 1, index_instruction->type_id())) {
+        return false;
+      }
+
+      // The module must have the definition of bool type to make a comparison.
+      opt::analysis::Bool bool_type;
+      if (!ir_context->get_type_mgr()->GetId(&bool_type)) {
+        return false;
+      }
```
afd, line 180: Use fuzzerutil::MaybeGetBoolType here.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -136,24 +136,50 @@ bool TransformationAccessChain::IsApplicable(
...
+
+      // Use value 0 to traverse the composite type
```
afd, line 182: Can you update this comment to say that we need some index in order to walk into the composite type, and because this is not a struct we don't necessarily have a known constant index.  We thus pretend the index is zero because that will always be in bounds, and the specific index is irrelevant for a non-struct type.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -225,16 +251,55 @@ void TransformationAccessChain::Apply(
...
+      auto index_instruction = ir_context->get_def_use_mgr()->GetDef(index_id);
+
+      uint32_t bound = fuzzerutil::GetBoundForCompositeIndex(
+          *ir_context->get_def_use_mgr()->GetDef(subobject_type_id),
+          ir_context);
+
+      auto bound_minus_one_id =
+          fuzzerutil::MaybeGetIntegerConstantFromValueAndType(
+              ir_context, bound - 1, index_instruction->type_id());
+
```
afd, line 269: Add an assertion (with a '&& "..."' comment) to say that this must exist as a precondition of the transformation.

source/fuzz/transformation_access_chain.cpp
```diff
@@ -225,16 +251,55 @@ void TransformationAccessChain::Apply(
...
+      opt::analysis::Bool bool_type;
+      uint32_t bool_type_id = ir_context->get_type_mgr()->GetId(&bool_type);
```
afd, line 271: Use MaybeGetBoolType, and add a precondition assertion.

source/fuzz/transformation_access_chain.cpp
stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Done

source/fuzz/transformation_access_chain.cpp
stefanomil: Done

source/fuzz/fuzzer_pass.h
```diff
@@ -179,7 +179,7 @@ class FuzzerPass {
   // on the |is_irrelevant| parameter.
   uint32_t FindOrCreateIntegerConstant(const std::vector<uint32_t>& words,
                                        uint32_t width, bool is_signed,
-                                       bool is_irrelevant);
+                                       bool is_irrelevant = false);
```
afd, line 182: Looks like you may have merged incorrectly here; @Vasniktel changed this so that it does not use default parameters anymore, so we should keep it that way.

stefanomil: I have now removed the default parameter (I also noticed that the last line in the comment above seemed to break pointlessly, so I fixed it)

PR #3548, opened 2020-07-16
Debug info preservation in loop-unroll pass

jaebaek:
When we copy the loop body to unroll it, we have to copy its
instructions but DebugDeclare or DebugValue used for the declaration
i.e., DebugValue with Deref must not be copied and only the first block
can contain those instructions.

### DIFF
--- a/source/opt/debug_info_manager.h
+++ b/source/opt/debug_info_manager.h
@@ -133,10 +133,12 @@ class DebugInfoManager {
   uint32_t BuildDebugInlinedAtChain(uint32_t callee_inlined_at,
                                     DebugInlinedAtContext* inlined_at_ctx);
 
-  // Returns true if |variable_id| has DebugDeclare or DebugVal.
+  // Returns true if there is a debug declaration instruction whose
+  // 'Local Variable' operand is |variable_id|.
   bool IsVariableDebugDeclared(uint32_t variable_id);
 
-  // Kills all DebugDeclares for |variable_id|
+  // Kills all debug declaration instructions with Deref whose 'Local Variable'
+  // operand is |variable_id|.
   void KillDebugDeclares(uint32_t variable_id);
 
   // Generates a DebugValue instruction with value |value_id| for every local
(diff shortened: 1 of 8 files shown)


dnovillo:
> When we copy the loop body to unroll it, we have to copy its
> instructions but DebugDeclare or DebugValue used for the declaration
> i.e., DebugValue with Deref must not be copied and only the first block
> can contain those instructions.

Can you add an example?  What is a `DebugValue` with `Deref`?  And why can't it be copied?

Sorry for all the questions!  I'm still trying to wrap my head around some of these changes.

source/opt/loop_unroller.cpp
```diff
@@ -286,6 +286,10 @@ class LoopUnrollerUtilsImpl {
   // to be the actual value of the phi at that point.
   void LinkLastPhisToStart(Loop* loop) const;
 
+  // Kill DebugDeclare or DebugValue that is used for the declaration with
+  // Deref from |bb|.
```
dnovillo, line 290: What does 'DebugValue that is used for the declaration' mean?  May be: "Kill all debug declaration instructions from |bb|".  It seems that there are two ways to tie debug info to a variable:

* `DebugDeclare`
* `DebugValue` with a dereference expression.  Used when a compound type is scalarized (e.g. by SROA).

We seem to be overloading terminology.  Perhaps call these something like "debug info declaration" instructions?  It uses the word "declare" again, but I can't think of anything better atm.  It should convey that there are more than one way to associate debug info with a memory location or part of a memory location.  Names are hard ;)

source/opt/debug_info_manager.h
```diff
@@ -148,6 +148,10 @@ class DebugInfoManager {
   // Erases |instr| from data structures of this class.
   void ClearDebugInfo(Instruction* instr);
 
+  // Returns true if |instr| is a DebugDeclare or a DebugValue with Deref
+  // that is used for a declaration.
+  bool IsUsedForDebugDeclare(Instruction* instr);
```
dnovillo, line 153: I don't follow.  Does this mean that the instruction acts as a declaration point?  Used for `DebugDeclare` means to me a `DebugDeclare`.  Perhaps `IsDebugInfoDeclaration`?  See my comment below.

source/opt/loop_unroller.cpp
jaebaek: "Kill all debug declaration instructions from |bb|" sounds good. Thanks!

source/opt/debug_info_manager.h
jaebaek: I just changed it to `IsDebugDeclare` that returns true if `instr` is a debug declaration instruction.

PR #3549, opened 2020-07-16
OpenCL.DebugInfo.100 DebugTypeArray with variable size

jaebaek:
The updated OpenCL.DebugInfo.100 spec says that we can use
variable for "Component Count" operand of DebugTypeArray i.e.,
DebugLocalVariable or DebugGlobalVariable. This commit updates spirv-val
based on the spec update.

### DIFF
--- a/source/val/validate_extensions.cpp
+++ b/source/val/validate_extensions.cpp
@@ -2730,8 +2730,8 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
           if (invalid) {
             return _.diag(SPV_ERROR_INVALID_DATA, inst)
                    << ext_inst_name() << ": Component Count must be "
-                   << "OpConstant with a 32 or 64-bits integer scalar type or "
-                   << "DebugGlobalVariable or DebugLocalVariable with a 32 or "
+                   << "OpConstant with a 32- or 64-bits integer scalar type or "
+                   << "DebugGlobalVariable or DebugLocalVariable with a 32- or "
                    << "64-bits unsigned integer scalar type";
           }
         }
--- a/source/val/validate_extensions.cpp
+++ b/source/val/validate_extensions.cpp
@@ -2697,27 +2697,32 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
             if (!component_count->word(3)) {
               invalid = true;
             }
-          } else if (OpenCLDebugInfo100Instructions(component_count->word(4)) ==
-                         OpenCLDebugInfo100DebugLocalVariable ||
-                     OpenCLDebugInfo100Instructions(component_count->word(4)) ==
-                         OpenCLDebugInfo100DebugGlobalVariable) {
+          } else if (component_count->words().size() > 6 &&
+                     (OpenCLDebugInfo100Instructions(component_count->word(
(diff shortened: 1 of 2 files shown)


source/val/validate_extensions.cpp
```diff
@@ -2226,13 +2226,36 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
             ValidateOperandDebugType(_, "Base Type", inst, 5, ext_inst_name);
         if (validate_base_type != SPV_SUCCESS) return validate_base_type;
         for (uint32_t i = 6; i < num_words; ++i) {
-          CHECK_OPERAND("Component Count", SpvOpConstant, i);
+          bool invalid = false;
           auto* component_count = _.FindDef(inst->word(i));
-          if (!_.IsIntScalarType(component_count->type_id()) ||
-              !component_count->word(3)) {
+          if (component_count->opcode() == SpvOpConstant) {
+            if (!_.IsIntScalarType(component_count->type_id()) ||
```
alan-baker, line 2232: This isn't checking the 32- or 64-bit size requirement. Didn't before either, but that is spelled out in the spec.

source/val/validate_extensions.cpp
```diff
@@ -2226,13 +2226,36 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
...
+          } else if (OpenCLDebugInfo100Instructions(component_count->word(4)) ==
+                         OpenCLDebugInfo100DebugLocalVariable ||
+                     OpenCLDebugInfo100Instructions(component_count->word(4)) ==
+                         OpenCLDebugInfo100DebugGlobalVariable) {
+            auto* component_count_type = _.FindDef(component_count->word(6));
+            if (OpenCLDebugInfo100Instructions(component_count_type->word(4)) !=
+                    OpenCLDebugInfo100DebugTypeBasic ||
+                (OpenCLDebugInfo100DebugBaseTypeAttributeEncoding(
+                     component_count_type->word(7)) !=
+                     OpenCLDebugInfo100Signed &&
```
alan-baker, line 2245: > Otherwise its type must be
 a <<DebugTypeBasic,*DebugTypeBasic*>> whose 'Size' is 32 or 64 and whose
 'Encoding' is *Unsigned*.

Doesn't seem that Signed encoding is valid. Also the sizes are not checked.

jaebaek: done

source/val/validate_extensions.cpp
```diff
@@ -2682,13 +2690,49 @@ spv_result_t ValidateExtInst(ValidationState_t& _, const Instruction* inst) {
...
+          } else {
+            invalid = true;
+          }
+          if (invalid) {
             return _.diag(SPV_ERROR_INVALID_DATA, inst)
-                   << ext_inst_name() << ": Component Count must be positive "
-                   << "integer";
+                   << ext_inst_name() << ": Component Count must be "
+                   << "OpConstant with a 32 or 64-bits integer scalar type or "
+                   << "DebugGlobalVariable or DebugLocalVariable with a 32 or "
```
alan-baker, line 2734: nit: prefer `32- or 64-bit unsigned integer scalar type`

jaebaek: done

Issue #3526, opened 2020-07-13
current gate tip fails to build with mingw64 gcc 10.1.0 due to maybe uninitialized error


binarycrusader:
Current spirv-tools gate tip fails to build with mingw64 gcc 10.0 due to maybe uninitialized error:

```
In file included from G:/Devel/scopes-md/SPIRV-Tools/test/opt/instruction_test.cpp:21:
G:/Devel/scopes-md/SPIRV-Tools/source/opt/instruction.h: In member function 'virtual void spvtools::opt::{anonymous}::InstructionTest_OperandAsLiteralUint64_64bits_Test::TestBody()':
G:/Devel/scopes-md/SPIRV-Tools/source/opt/instruction.h:101:23: error: '*((void*)& operand +32)' may be used uninitialized in this function [-Werror=maybe-uninitialized]
                        101 |     uint64_t result = uint64_t(words[0]);
                            |                       ^~~~~~~~~~~~~~~~~~
cc1plus.exe: all warnings being treated as errors
make[2]: *** [test/opt/CMakeFiles/test_opt.dir/build.make:629: test/opt/CMakeFiles/test_opt.dir/instruction_test.cpp.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:2656: test/opt/CMakeFiles/test_opt.dir/all] Error 2
make: *** [Makefile:161: all] Error 2 
```

binarycrusader:
Looks like this only happens when the optional googletest component is checked out per the build instructions:

`git clone https://github.com/google/googletest.git external/googletest # optional`

dneto0:
The normal development flow is to have googletest checked out and to build and run tests. So this is intended to build without issue.

The method in question is:
  // Returns a literal integer operand as a uint64_t
  uint64_t AsLiteralUint64() const {
    assert(type == SPV_OPERAND_TYPE_TYPED_LITERAL_NUMBER);
    assert(1 <= words.size());
    assert(words.size() <= 2);
    // Load the low word.
    uint64_t result = uint64_t(words[0]);
    if (words.size() > 1) {
      result = result | (uint64_t(words[1]) << 32);
    }
    return result;
  }

The data is protected by contract, but the compiler can't tell that.  And in a debug build the bad case would trigger the assert.
It would be harmless to add a member initializations.  I can try that.

dneto0:
[...]

PR #3556, opened 2020-07-17
WIP: Add member initializations to fix GCC10 warning

dneto0:
See #3526

I don't have MinGW GCC 10.0 handy so I can't verify this actually fixes the problem.  It's just a guess.

### DIFF
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -80,8 +80,10 @@ struct Operand {
 
   Operand(spv_operand_type_t t, const OperandData& w) : type(t), words(w) {}
 
-  spv_operand_type_t type;  // Type of this logical operand.
-  OperandData words;        // Binary segments of this logical operand.
+  // Type of this logical operand.
+  spv_operand_type_t type = SPV_OPERAND_TYPE_NONE;
+  // Binary segments of this logical operand.
+  OperandData words = {};
 
   // Returns a string operand as a C-style string.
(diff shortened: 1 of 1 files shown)


Biswa96:
In msys2/mingw64 environment, GCC 10 is used. There is a way to setup GitHub Actions with msys2. If allowed I can talk with someone who is familiar with it.

Biswa96:
I have tested with mingw-w64-gcc 10.1.0. It shows same warning as follows:

```c
In file included from ../test/opt/instruction_test.cpp:21:
../source/opt/instruction.h: In member function 'virtual void spvtools::opt::{anonymous}::InstructionTest_OperandAsLiteralUint64_64bits_Test::TestBody()':
../source/opt/instruction.h:101:23: warning: '*((void*)& operand +32)' may be used uninitialized in this function [-Wmaybe-uninitialized]
  101 |     uint64_t result = uint64_t(words[0]);
      |                       ^~~~~~~~~~~~~~~~~~
```

There is a way to setup GitHub Actions with msys2 so that devs don't have to install the environment. If allowed I can talk with someone who is familiar with it.

s-perron:
I'm guessing this solution does not work because the `SmallVector` class does not write to the memory in the class even if you initialize the vector to `{}`.  It probably just sets the size.

s-perron:
Are we able to verify if this is still a problem?

PR #3568, opened 2020-07-21
Support SPV_KHR_terminate_invocation

alan-baker:
Covers:
- assembler
- disassembler
- validator
- optimizer

### DIFF
--- a/source/opt/wrap_opkill.cpp
+++ b/source/opt/wrap_opkill.cpp
@@ -27,7 +27,8 @@ Pass::Status WrapOpKill::Process() {
   for (uint32_t func_id : func_to_process) {
     Function* func = context()->GetFunction(func_id);
     bool successful = func->WhileEachInst([this, &modified](Instruction* inst) {
-      if (inst->opcode() == SpvOpKill) {
+      const auto opcode = inst->opcode();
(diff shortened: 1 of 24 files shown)


source/opt/wrap_opkill.cpp
```diff
@@ -108,13 +115,22 @@ uint32_t WrapOpKill::GetVoidFunctionTypeId() {
...
 }
 
-uint32_t WrapOpKill::GetOpKillFuncId() {
-  if (opkill_function_ != nullptr) {
-    return opkill_function_->result_id();
+uint32_t WrapOpKill::GetKillingFuncId(SpvOp opcode) {
+  //  Parameterize by opcode
+  assert(opcode == SpvOpKill
+
+         || opcode == SpvOpTerminateInvocation);
```
s-perron, line 122: The formatting here looks odd.  I don't know if clang-format will fix this up or not.

test/val/val_extension_spv_khr_terminate_invocation.cpp
```diff
@@ -0,0 +1,150 @@
+// Copyright (c) 2017 Google Inc.
```
s-perron, line 1: Wrong year.

PR #3571, opened 2020-07-21
Preserve OpenCL.DebugInfo.100 through private-to-local pass

jaebaek:
1. A debug instruction must not have any impact on the private-to-local
optimization.
2. DebugGlobalVariable for the OpVariable optimized by private-to-local
must be converted to a DebugLocalVariable because we want to run
the ssa-rewrite to propagate DebugValue properly, which correctly
delivers the updates of local variables to the debugger.

### DIFF
--- a/source/opt/private_to_local_pass.cpp
+++ b/source/opt/private_to_local_pass.cpp
@@ -157,6 +157,12 @@ uint32_t PrivateToLocalPass::GetNewType(uint32_t old_type_id) {
 bool PrivateToLocalPass::IsValidUse(const Instruction* inst) const {
   // The cases in this switch have to match the cases in |UpdateUse|.
   // If we don't know how to update it, it is not valid.
+  if (inst->GetOpenCL100DebugOpcode() ==
+          OpenCLDebugInfo100DebugGlobalVariable ||
+      inst->GetOpenCL100DebugOpcode() == OpenCLDebugInfo100DebugDeclare ||
+      inst->GetOpenCL100DebugOpcode() == OpenCLDebugInfo100DebugValue) {
+    return true;
+  }
   switch (inst->opcode()) {
     case SpvOpLoad:
     case SpvOpStore:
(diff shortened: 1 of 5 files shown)


test/opt/private_to_local_test.cpp
```diff
@@ -452,6 +452,48 @@ TEST_F(PrivateToLocalTest, IdBoundOverflow1) {
...
+; CHECK: [[newtype:%[a-zA-Z_\d]+]] = OpTypePointer Function [[float]]
+          %6 = OpTypePointer Private %5
+; CHECK-NOT: OpVariable [[.+]] Private
+          %8 = OpVariable %6 Private
+
+         %12 = OpExtInst %3 %10 DebugTypeBasic %11 %14 Float
+         %15 = OpExtInst %3 %10 DebugSource %11
+         %16 = OpExtInst %3 %10 DebugCompilationUnit 1 4 %15 GLSL
+; CHECK: DebugGlobalVariable {{%\d+}} {{%\d+}} {{%\d+}} 0 0 {{%\d+}} {{%\d+}} [[newvar:%[a-zA-Z_\d]+]]
+         %17 = OpExtInst %3 %10 DebugGlobalVariable %11 %12 %15 0 0 %16 %11 %8 FlagIsDefinition
```
s-perron, line 482: Once the private variable become a function scope variable, should it still be using a DebugGlobalVariable?  That does not seem right to me.

source/opt/private_to_local_pass.cpp
```diff
@@ -157,6 +157,12 @@ uint32_t PrivateToLocalPass::GetNewType(uint32_t old_type_id) {
 bool PrivateToLocalPass::IsValidUse(const Instruction* inst) const {
   // The cases in this switch have to match the cases in |UpdateUse|.
   // If we don't know how to update it, it is not valid.
+  if (inst->GetOpenCL100DebugOpcode() ==
+          OpenCLDebugInfo100DebugGlobalVariable ||
+      inst->GetOpenCL100DebugOpcode() == OpenCLDebugInfo100DebugDeclare ||
+      inst->GetOpenCL100DebugOpcode() == OpenCLDebugInfo100DebugValue) {
```
s-perron, line 163: Can a DebugDeclare and DebugValue reference a private varaible?  I thought these were only used for a local variable, and a private variable is a global variable.

test/opt/private_to_local_test.cpp
jaebaek: I converted it to a DebugLocalVariable and creates a new DebugDeclare for it.

PR #3575, opened 2020-07-23
spirv-fuzz: adds TransformationReplaceCopyMemoryWithLoadStore

### LINKED ISSUE #3352: spirv-fuzz: Replace OpCopyMemory with the pair OpLoad and OpStore instructions
If we have `OpCopyMemory %target %source`, we could load `%source` to `%source_value` and store `%source_value` to `%target`. The following is an example.
```
%source_value = OpLoad %type %source
OpStore %target %source_value
```
This transformation would receive a fresh id and an instruction descriptor.
The fuzzer pass would look for `OpCopyMemory` instructions and, randomly, apply the transformation.

antnkarp:
Adds a transformation that replaces instruction OpCopyMemory with
loading the source variable to an intermediate value and storing this
value into the target variable of the original OpCopyMemory instruction.

Fixes #3352

### DIFF
--- a/source/fuzz/fuzzer_pass_replace_copy_memories_with_loads_stores.cpp
+++ b/source/fuzz/fuzzer_pass_replace_copy_memories_with_loads_stores.cpp
@@ -41,7 +41,7 @@ void FuzzerPassReplaceCopyMemoriesWithLoadsStores::Apply() {
                 ->GetChanceOfReplacingCopyMemoryWithLoadStore())) {
       return;
     }
-    // The instruction must be OpCopyObject.
+    // The instruction must be OpCopyMemory.
     if (instruction->opcode() != SpvOpCopyMemory) {
       return;
     }
@@ -56,6 +56,7 @@ void FuzzerPassReplaceCopyMemoriesWithLoadsStores::Apply() {
 
     auto instruction_descriptor =
         MakeInstructionDescriptor(GetIRContext(), instruction);
+
     // Apply the transformation replacing OpCopyMemory with OpLoad and OpStore.
     ApplyTransformation(TransformationReplaceCopyMemoryWithLoadStore(
         GetFuzzerContext()->GetFreshId(), instruction_descriptor));
--- a/source/fuzz/fuzzer_pass_replace_copy_memories_with_loads_stores.cpp
+++ b/source/fuzz/fuzzer_pass_replace_copy_memories_with_loads_stores.cpp
@@ -1,0 +1,66 @@
+// Copyright (c) 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_replace_copy_memories_with_loads_stores.h"
+
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/instruction_descriptor.h"
+#include "source/fuzz/transformation_replace_copy_memory_with_load_store.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassReplaceCopyMemoriesWithLoadsStores::
+    FuzzerPassReplaceCopyMemoriesWithLoadsStores(
+        opt::IRContext* ir_context,
+        TransformationContext* transformation_context,
+        FuzzerContext* fuzzer_context,
+        protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassReplaceCopyMemoriesWithLoadsStores::
+    ~FuzzerPassReplaceCopyMemoriesWithLoadsStores() = default;
+
+void FuzzerPassReplaceCopyMemoriesWithLoadsStores::Apply() {
+  GetIRContext()->module()->ForEachInst([this](opt::Instruction* instruction) {
+    // Randomly decide whether to replace the OpCopyMemory.
+    if (!GetFuzzerContext()->ChoosePercentage(
+            GetFuzzerContext()
+                ->GetChanceOfReplacingCopyMemoryWithLoadStore())) {
+      return;
+    }
+    // The instruction must be OpCopyObject.
+    if (instruction->opcode() != SpvOpCopyMemory) {
+      return;
+    }
+
+    // It must be valid to insert OpStore and OpLoad instructions
+    // before the instruction OpCopyMemory.
+    if (!fuzzerutil::CanInsertOpcodeBeforeInstruction(SpvOpStore,
+                                                      instruction) ||
+        !fuzzerutil::CanInsertOpcodeBeforeInstruction(SpvOpLoad, instruction)) {
+      return;
+    }
+
+    auto instruction_descriptor =
+        MakeInstructionDescriptor(GetIRContext(), instruction);
+    // Apply the transformation replacing OpCopyMemory with OpLoad and OpStore.
+    ApplyTransformation(TransformationReplaceCopyMemoryWithLoadStore(
+        GetFuzzerContext()->GetFreshId(), instruction_descriptor));
+  });
+}
+
+}  // namespace fuzz
+}  // namespace spvtools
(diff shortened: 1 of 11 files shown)


source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -404,7 +404,8 @@ message Transformation {
     TransformationAddSynonym add_synonym = 57;
     TransformationAddRelaxedDecoration add_relaxed_decoration = 58;
     TransformationReplaceParamsWithStruct replace_params_with_struct = 59;
-    TransformationReplaceCopyObjectWithStoreLoad replace_copy_object_with_store_load = 60;
+    TransformationReplaceCopyMemoryWithLoadStore replace_copy_memory_with_load_store = 60;
+    TransformationReplaceCopyObjectWithStoreLoad replace_copy_object_with_store_load = 61;
```
afd, line 408: When you rebase on top of the change I just merged, keep the copy object transformation as 60 and make this one 61.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1217,6 +1218,19 @@ message TransformationRecordSynonymousConstants {
...
 }
 
+message TransformationReplaceCopyMemoryWithLoadStore {
+
+   // A transformation that replaces instructions OpCopyMemory with loading
+   // the source variable to an intermediate value and storing this value into the
+   // target variable of the original OpCopyMemory instruction.
+
+   // The intermediate value
+   uint32 source_value = 1;
```
afd, line 1228: For transformations with a single fresh id, I tend to name the field "fresh_id".  Having "fresh" in the name is good for readability.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1217,6 +1218,19 @@ message TransformationRecordSynonymousConstants {
...
+
+   // The instruction descriptor
+   InstructionDescriptor instruction_descriptor = 2;
```
afd, line 1231: "instruction_descriptor" is a good name when we want to describe any old instruction.  Here, though, we want to describe an OpCopyMemory, and we're only using an instruction descriptor because OpCopyMemory doesn't have a result id.  Thus I would call this "copy_memory_instruction_descriptor", or just "copy_memory_instruction" to indicate that it is indented to be one of these (I would update the comment a little to reflect this too).

source/fuzz/transformation_replace_copy_memory_with_load_store.h
```diff
@@ -0,0 +1,57 @@
...
+namespace fuzz {
+
+class TransformationReplaceCopyMemoryWithLoadStore : public Transformation {
+ public:
+  explicit TransformationReplaceCopyMemoryWithLoadStore(
+      const protobufs::TransformationReplaceCopyMemoryWithLoadStore& message);
+
+  TransformationReplaceCopyMemoryWithLoadStore(
+      uint32_t source_value,
+      const protobufs::InstructionDescriptor& instruction_descriptor);
```
afd, line 33: If you change the field names in the .proto file, be sure to change these parameter names (in the .cpp file too).

source/fuzz/transformation_replace_copy_memory_with_load_store.cpp
```diff
@@ -0,0 +1,118 @@
...
+  *message_.mutable_instruction_descriptor() = instruction_descriptor;
+}
+
+bool TransformationReplaceCopyMemoryWithLoadStore::IsApplicable(
+    opt::IRContext* ir_context, const TransformationContext& /*unused*/) const {
+  // |message_.source_value| must be fresh.
+  if (!fuzzerutil::IsFreshId(ir_context, message_.source_value())) {
+    return false;
+  }
+  // The instruction to be replaced must be defined and of opcode
```
afd, line 43: "and of opcode" -> "and have opcode"

source/fuzz/transformation_replace_copy_memory_with_load_store.cpp
```diff
@@ -0,0 +1,118 @@
...
+      FindInstruction(message_.instruction_descriptor(), ir_context);
+  if (!copy_memory_instruction ||
+      copy_memory_instruction->opcode() != SpvOpCopyMemory) {
+    return false;
+  }
+  // It must be valid to insert the OpStore and OpLoad instruction before it.
+  return (fuzzerutil::CanInsertOpcodeBeforeInstruction(
+              SpvOpStore, copy_memory_instruction) &&
+          fuzzerutil::CanInsertOpcodeBeforeInstruction(
+              SpvOpLoad, copy_memory_instruction));
```
afd, line 55: I think we can get rid of this.  OpCopyMemory is of the same nature as OpLoad and OpStore, so I believe it's always legitimate to insert the latter in place of the former in terms of where in the module they can appear.

If you can think of an example to the contrary then keep this and add a test for that example that demonstrates that "false" can actually be returned here under the right conditions.  Otherwise drop the code and just return true here (or change the previous if statement to a return).

source/fuzz/transformation_replace_copy_memory_with_load_store.cpp
```diff
@@ -0,0 +1,118 @@
...
+      copy_memory_instruction->GetSingleWordInOperand(1));
+  auto target_type_opcode =
+      ir_context->get_def_use_mgr()->GetDef(target->type_id())->opcode();
+  auto source_type_opcode =
+      ir_context->get_def_use_mgr()->GetDef(source->type_id())->opcode();
+
+  // Both operands must be pointers.
+  assert(target_type_opcode == SpvOpTypePointer &&
+         source_type_opcode == SpvOpTypePointer &&
+         "Operands must be of type OpTypePointer");
```
afd, line 80: All this is merely a coherence check.

I like coherence checks in general, so happy to keep this, but please indicate with a comment that it's a coherence check - because it introduces various temporary variables a reader of the code might at first think they were going to be used in the functionality of Apply.

Also, you may need to do:

(void)temporary_name; // Keep release-mode compilers happy

to avoid compiler warnings - in release mode the assertions don't exist so you might get messages about unused variables.

source/fuzz/transformation_replace_copy_memory_with_load_store.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  // And they must point to the same type.
+  uint32_t target_pointee_type = fuzzerutil::GetPointeeTypeIdFromPointerType(
+      ir_context, target->type_id());
+  uint32_t source_pointee_type = fuzzerutil::GetPointeeTypeIdFromPointerType(
+      ir_context, source->type_id());
+  assert(target_pointee_type == source_pointee_type &&
+         "Operands must have the same type to which they point to.");
```
afd, line 88: Actually I think this is also part of the coherence check.

source/fuzz/fuzzer_pass_replace_copy_memories_with_loads_stores.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void FuzzerPassReplaceCopyMemoriesWithLoadsStores::Apply() {
+  GetIRContext()->module()->ForEachInst([this](opt::Instruction* instruction) {
+    // Randomly decide whether to replace the OpCopyMemory.
+    if (!GetFuzzerContext()->ChoosePercentage(
+            GetFuzzerContext()
+                ->GetChanceOfReplacingCopyMemoryWithLoadStore())) {
+      return;
+    }
+    // The instruction must be OpCopyMemory.
```
afd, line 44: Nit: you've not left a space between "}" and the following "//" here, but below you do.  It doesn't really matter, but it would be mildly preferable to do one or the other consistently.  (I don't mind which.)

BTW, just to give a bit of context about code review: this is the kind of comment I only make when I am requesting other changes; if this were the only thing I noticed I wouldn't request changes based on it.

source/fuzz/fuzzer_pass_replace_copy_memories_with_loads_stores.cpp
```diff
@@ -0,0 +1,67 @@
...
+      return;
+    }
+
+    // It must be valid to insert OpStore and OpLoad instructions
+    // before the instruction OpCopyMemory.
+    if (!fuzzerutil::CanInsertOpcodeBeforeInstruction(SpvOpStore,
+                                                      instruction) ||
+        !fuzzerutil::CanInsertOpcodeBeforeInstruction(SpvOpLoad, instruction)) {
+      return;
+    }
```
afd, line 55: Please see my remark at the end of "IsApplicable" below.  In short: I think this check should go.

source/fuzz/fuzzer_pass_replace_copy_memories_with_loads_stores.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+    auto instruction_descriptor =
+        MakeInstructionDescriptor(GetIRContext(), instruction);
```
afd, line 58: I would personally get rid of the temporary and use MakeInstructionDescriptor inline in the call below.  I know @paulthomson thinks I do this too much sometimes, though :)

source/fuzz/transformation_replace_copy_memory_with_load_store.cpp
antnkarp: Thanks for pointing this out.

PR #3586, opened 2020-07-24
spirv-fuzz: adds TransformationReplaceLoadStoreWithCopyMemory

### LINKED ISSUE #3353: spirv-fuzz: Replace the pair OpLoad and OpStore with OpCopyMemory
If we have a pair of instructions like `%source_value = OpLoad %type %source` and `OpStore %target %source_value`, we could replace this pair with `OpCopyMemory %target %source`.

This transformation would receive `%target`, `%source` and a pair of instruction descriptors.
The fuzzer pass would look for pairs `OpLoad` and `OpStore`, like described above, and, randomly apply the transformation.

antnkarp:
Adds a transformation that takes a pair of instruction descriptors to
OpLoad and OpStore that have the same intermediate value and replaces
the OpStore with an equivalent OpCopyMemory.

Fixes #3353


source/fuzz/fuzzer_pass_replace_loads_stores_with_copy_memories.h
```diff
@@ -0,0 +1,41 @@
...
+#define SPIRV_TOOLS_FUZZER_PASS_REPLACE_LOADS_STORES_WITH_COPY_MEMORIES_H
+
+#include "source/fuzz/fuzzer_pass.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// Takes a pair of instruction descriptors to OpLoad and OpStore that have
+// the same intermediate value and replaces the OpStore with an equivalent
+// OpCopyMemory.
```
paulthomson, line 25: This sounds like the documentation for the transformation, not the fuzzer pass. But thanks for creating this comment; we are lacking such comments in most transformation classes! On the other hand, perhaps the best place to document a transformation is in the protobuf file.

source/fuzz/transformation_replace_load_store_with_copy_memory.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  assert(target_pointee_type == source_pointee_type &&
+         "The target of OpStore and the source of OpLoad must point to the "
+         "same type.");
+
+  // Coherence check: The first operand of the OpLoad must match the type to
+  // which the source of OpLoad points to.
+  assert(load_instruction->GetSingleWordOperand(0) == source_pointee_type &&
+         "First operand of the OpLoad must match the type to which the source "
+         "of OpLoad points to.");
```
paulthomson, line 115: I recommend removing these coherence checks because they take up a lot of lines and they don't add much. Adding a redundant but simple `assert` seems fine, but adding additional code to check that the module is valid is unnecessary.

source/fuzz/transformation_replace_load_store_with_copy_memory.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+  // Insert the OpCopyMemory instruction before the OpStore instruction.
```
paulthomson, line 117: Extra checks: I think we may need to check that there are no memory-related instructions between the OpLoad and OpStore, similar to the check in the fuzzer pass (but this time focused on just the specific OpLoad and OpStore). Or perhaps we can just assume that this is very unlikely. The check will be fairly simple though.

source/fuzz/fuzzer_pass_replace_loads_stores_with_copy_memories.cpp
```diff
@@ -0,0 +1,86 @@
...
+          // interfering OpStore instructions. Consider for example:
+          // %a = OpLoad %ptr1
+          // OpStore %ptr2 %a <-- we haven't clear the map
+          // OpStore %ptr3 %a <-- if %ptr2 points to the same variable as
+          //    %ptr1, then replacing this instruction with OpCopyMemory %ptr3
+          //    %ptr1 is unsafe.
+          current_op_loads.clear();
+        }
+      }
+    }
```
paulthomson, line 72: Should we check if the pointee value is irrelevant? I suppose we don't need to. @afd

source/fuzz/fuzzer_pass_replace_loads_stores_with_copy_memories.cpp
```diff
@@ -0,0 +1,86 @@
...
+                &instruction));
+          }
+          // We need to clear the hash map. If we don't, there might be some
+          // interfering OpStore instructions. Consider for example:
+          // %a = OpLoad %ptr1
+          // OpStore %ptr2 %a <-- we haven't clear the map
+          // OpStore %ptr3 %a <-- if %ptr2 points to the same variable as
+          //    %ptr1, then replacing this instruction with OpCopyMemory %ptr3
+          //    %ptr1 is unsafe.
+          current_op_loads.clear();
```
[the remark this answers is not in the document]
antnkarp: Thanks for the explanation. I have fixed it.

source/fuzz/transformation_replace_load_store_with_copy_memory.cpp
antnkarp: I am not sure how to implement this. Should I put checks for opcodes in a for loop and iterate from OpLoad to OpStore? How to iterate over instructions in this way? In case of MemoryBarrier/MemoryNamedBarrier should I check for storage class of the variable as well?

antnkarp: Can I assume that the matching OpLoad and OpStore are always in the same block?

paulthomson: I think you can get an instruction iterator, or just use `instruction->NextNode()`. I think the way that you check opcodes (including memory barriers) should match the approach used in the fuzzer pass. So if you check the storage class in the fuzzer pass, you should do it here as well. You may want to add some static functions to the transformation: one that returns true if the given opcode is a memory barrier, one that returns true if the given opcode is some other memory instruction that we care about, one that returns true if a given storage is not affected by memory barriers. Or just one function (for opcodes) if you don't want to handle storage classes yet.

antnkarp: @paulthomson  Does iterating using instruction iterator work when matching OpLoad and OpStore instructions are in different blocks? Is it just simply taking instruction->NextNode() until we reach the OpStore? In the fuzzer pass we assume that they are in the same block and use for(auto& instruction: block).

paulthomson: I think that if the instructions are not in the same block then IsApplicable should return false. 

Thus, you can iterate from the load to store, but if you reach the end of the block, you will get nullptr and you should return false (from IsApplicable).

paulthomson: Sorry, I should have put this comment under `IsApplicable`, not under `Apply`. No need to check this in Apply.

source/fuzz/fuzzer_pass_replace_loads_stores_with_copy_memories.cpp
```diff
@@ -0,0 +1,108 @@
...
+        } else if (instruction.opcode() == SpvOpStore) {
+          if (current_op_loads.find(instruction.GetSingleWordOperand(1)) !=
+              current_op_loads.end()) {
+            // We have found the matching OpLoad instruction to the current
+            // OpStore instruction.
+            op_load_store_pairs.push_back(std::make_pair(
+                current_op_loads[instruction.GetSingleWordOperand(1)],
+                &instruction));
+          }
+          current_op_loads.clear();
```
paulthomson, line 59: This looks good. I was thinking of something more like this:

```
if opcode == SpvOpLoad:
  ...
else if opcode == SpvOpStore
  // add pair.

if IsMemoryWritingOpcode(opcode):
  current_op_loads.clear()
else if IsMemoryBarrierOpcode(opcode):
  for load in current_op_loads:
    storage_class = ...load...
    if !IsStorageClassSafeAcrossMemoryBarriers(storage_class):
      erase load
```

Where `IsMemoryWritingOpcode` would return true for `SpvOpStore`, `SpvOpCopyMemory`, `SpvOpCopyMemorySized`, and all atomic instructions **except** for `OpAtomicLoad` (and we might exclude other instructions some day). 

Then in IsApplicable, you could re-use most of this from within your loop:

```
if IsMemoryWritingOpcode(opcode):
  return false
else if IsMemoryBarrierOpcode(opcode) and !IsStorageClassSafeAcrossMemoryBarriers(storage_class):
  return false
```

antnkarp: Yes, this solution looks much cleaner and it doesn't reuse code. Thanks. I'll implement it.

test/fuzz/transformation_replace_load_store_with_copy_memory_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+  auto load_instruction_descriptor_1 =
+      MakeInstructionDescriptor(5, SpvOpLoad, 0);
+  auto load_instruction_descriptor_2 =
+      MakeInstructionDescriptor(5, SpvOpLoad, 1);
+  auto store_instruction_descriptor_1 =
+      MakeInstructionDescriptor(18, SpvOpStore, 0);
+  auto store_instruction_descriptor_2 =
+      MakeInstructionDescriptor(19, SpvOpStore, 0);
+
+  // Invalid: |load_instruction_descriptor| is incorrect.
```
paulthomson, line 88: It might be best to avoid using the word "invalid" because it is often used to mean invalid SPIR-V. Ally often uses // Bad: ...

source/fuzz/transformation_replace_load_store_with_copy_memory.h
```diff
@@ -0,0 +1,63 @@
...
+      const protobufs::TransformationReplaceLoadStoreWithCopyMemory& message);
+
+  TransformationReplaceLoadStoreWithCopyMemory(
+      const protobufs::InstructionDescriptor& load_instruction_descriptor,
+      const protobufs::InstructionDescriptor& store_instruction_descriptor);
+
+  // - |message_.load_instruction_descriptor| must refer to an OpLoad
+  // instruction.
+  // - |message_.store_instruction_descriptor| must refer to an OpStore
+  // instruction.
```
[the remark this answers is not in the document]
paulthomson: Having said this, you don't need to include _all_ details because the body of the function has good comments. But mention that the OpStore must write the intermediate value loaded by the OpLoad and the OpLoad and OpStore must not have certain memory instructions in between.

source/fuzz/transformation_replace_load_store_with_copy_memory.cpp
```diff
@@ -0,0 +1,150 @@
...
+  assert(load_instruction->result_id() ==
+             store_instruction->GetSingleWordOperand(1) &&
+         "OpLoad and OpStore must refer to the same value.");
+
+  // Get the ids of the source operand of the OpLoad and the target operand of
+  // the OpStore.
+  auto source = ir_context->get_def_use_mgr()->GetDef(
+      load_instruction->GetSingleWordOperand(2));
+  auto target = ir_context->get_def_use_mgr()->GetDef(
+      store_instruction->GetSingleWordInOperand(0));
```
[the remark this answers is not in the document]
antnkarp: Oh right, I see, in my code it was unnecessary. Thanks for pointing this out.

PR #3587, opened 2020-07-24
spirv-fuzz: Add condition to make functions livesafe

### LINKED ISSUE #3162: spirv-fuzz: Account for unreachable loop merge blocks when making functions livesafe
Making a function livesafe involves adding a branch from a loop's back-edge block to the loop's merge if the number of loop iterations exceeds some limit.  If the loop's merge block was unreachable, the new edge could be a problem, e.g. because the previously unreachable block might now be dominated, and might not be in the right position in the module for this.

A possible solution to this is to branch to a new block and have that new block return.

andreperezmaselco:
Fixes #3162.

### DIFF
--- a/source/fuzz/transformation_add_function.cpp
+++ b/source/fuzz/transformation_add_function.cpp
@@ -488,7 +488,16 @@ bool TransformationAddFunction::TryToAddLoopLimiters(
       // move on from this loop.
       continue;
     }
-    auto back_edge_block = ir_context->cfg()->block(back_edge_block_id);
+
+    // If the loop's merge block is unreachable, then it is valid to be before
+    // the loop header. In this case, if the back-edge block branch to the merge
+    // block, then the module will not be valid because blocks must appear
+    // before all blocks they dominate. Therefore, the loop header must
+    // dominates its merge block.
+    if (!ir_context->GetDominatorAnalysis(added_function)
+             ->Dominates(loop_header->id(), loop_header->MergeBlockId())) {
+      return false;
+    }
 
     // Go through the sequence of loop limiter infos and find the one
     // corresponding to this loop.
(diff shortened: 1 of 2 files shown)


test/fuzz/transformation_add_function_test.cpp
```diff
@@ -2337,54 +2337,8 @@ TEST(TransformationAddFunctionTest, InfiniteLoop1) {
...
-        %104 = OpUGreaterThanEqual %19 %102 %32
-               OpLoopMerge %14 %12 None
-               OpBranchConditional %104 %14 %12
-         %14 = OpLabel
-               OpReturn
-               OpFunctionEnd
-  )";
-  ASSERT_TRUE(IsEqual(env, expected, context.get()));
+  ASSERT_FALSE(add_livesafe_function.IsApplicable(context.get(),
+                                                  transformation_context));
```
afd, line 2341: Can you add a brief comment above this assertion explaining (in terms of unreachability) why we expect IsApplicable to fail?

andreperezmaselco: Done.

source/fuzz/transformation_add_function.cpp
```diff
@@ -488,7 +488,16 @@ bool TransformationAddFunction::TryToAddLoopLimiters(
       // move on from this loop.
       continue;
     }
-    auto back_edge_block = ir_context->cfg()->block(back_edge_block_id);
+
+    // If the loop's merge block is unreachable, then it is valid to be before
+    // the loop header. In this case, if the back-edge block branch to the merge
+    // block, then the module will not be valid because blocks must appear
+    // before all blocks they dominate. Therefore, the loop header must
+    // dominates its merge block.
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

PR #3590, opened 2020-07-26
CMake: install cmake files in libdir/cmake if not compiled with msvc

Biswa96:
when compiled with GCC or Clang in Windows the cmake files are installed
in individual folder. But with this change cmake files are placed in lib
folder which follows the GNUInstallDirs rules.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,7 +210,7 @@ else()
 endif()
 
 if(ENABLE_SPIRV_TOOLS_INSTALL)
-  if(WIN32)
+  if(MSVC)
     macro(spvtools_config_package_dir TARGET PATH)
       set(${PATH} ${TARGET}/cmake)
     endmacro()


dj2:
Do you have a link to the  install dir rules? What is the issue with using a subdirectory?

Biswa96:
> Do you have a link to the install dir rules?

[Here](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html).

> What is the issue with using a subdirectory?

As I mentioned in the commit message, without this change the .cmake files are installed in separate folder instead of lib/cmake folder.

dj2:
It looks like the files would have been in a subfolder of `lib/cmake`? So, my question is why is using a subfolder an issue? Will moving to a parent directly cause issues for folks using the cmake files now? Like, can this potentially break existing users as they won't be able to find the cmake files?

Biswa96:
In msys2 and cygwin environment, it follows Unix like directory hierarchy (typical bin, lib, usr, etc). Without this change,  SPIRV-Tools-* folders (containing .cmake files) are created in `CMAKE_INSTALL_PREFIX` e.g. `/usr`. According to the [search procedure][1] of [`find_package()`][2] function, for all the Unix cases, CMake searches `<prefix>/(lib/<arch>|lib*|share)` style path e.g. `/usr/lib`.

[1]: https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure
[2]: https://cmake.org/cmake/help/latest/command/find_package.html

dj2:
They go in `/usr`? Don't they hit the ` set(${PATH} ${CMAKE_INSTALL_LIBDIR}/cmake/${TARGET})` and go in the `/usr/cmake` folder? Which would get found by those search paths.  I'm not sure I understand what this is fixing?

Biswa96:
It's not `/usr/cmake`. It's `/usr/lib/cmake`. I am not cmake expert. This is as far as I know.

Biswa96:
It's not `/usr/cmake`. It's `/usr/lib/cmake`. I am not cmake expert. The previous logic is as far as I know.

Biswa96:
It's not `/usr/cmake`. It's `/usr/lib/cmake`. 

> Don't they hit the `set(${PATH} ${CMAKE_INSTALL_LIBDIR}/cmake/${TARGET})`

In mingw environment, `WIN32` is true. So, CMake hits `set(${PATH} ${TARGET}/cmake)`. I am not cmake expert. The previous logic is as far as I know.

dj2:
Right, but even that is a valid find_package path isn't it `<prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/`.

Are you encountering an issue where the cmake files aren't being found by find_package?

Biswa96:
Maybe this https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-spirv-tools/001-cmake-files-location.patch

dj2:
I don't see any explanation in that issue of what this is fixing or working around. Without an actual issue, I'm not sure if we want to change this just for the sake of change.

Biswa96:
OK, as you wish. Thank you for your time ❤️

dneto0:
Thanks for the discussion. 

The MinGW toolchains are also used for people on Windows who don't use msys or cygwin-like environments.

PR #3597, opened 2020-07-27
spirv-fuzz: Add TransformationMakeVectorOperationDynamic

### LINKED ISSUE #3588: spirv-fuzz: Dynamically extract and insert vector components
A vector component can be extracted or inserted using the `OpCompositeExtract` and `OpCompositeInsert` instructions.
Although, these instructions use literals to index the component, meaning that the component is known by the SPIR-V compiler.
It would be interesting to replace this static approach with a dynamic using the  `OpVectorExtractDynamic` and `OpVectorInsertDynamic` instructions.

Reference shader:
```
      OpCapability Shader
      OpMemoryModel Logical GLSL450
      OpEntryPoint Vertex %11 "main"
 
; Types
 %2 = OpTypeVoid
 %3 = OpTypeFunction %2
 %4 = OpTypeFloat 32
 %5 = OpTypeVector %4 4

; Constants
 %6 = OpConstant %4 0
 %7 = OpConstant %4 1
 %8 = OpConstant %4 2
 %9 = OpConstant %4 3
%10 = OpConstantComposite %5 %6 %7 %8 %9

; main function
%11 = OpFunction %2 None %3
%12 = OpLabel

; Statically extracting vector components
%13 = OpCompositeExtract %4 %10 0
%14 = OpCompositeExtract %4 %10 1
%15 = OpCompositeExtract %4 %10 2
%16 = OpCompositeExtract %4 %10 3

; Statically inserting vector components
%17 = OpCompositeInsert %5 %6 %10 0
%18 = OpCompositeInsert %5 %7 %10 1
%19 = OpCompositeInsert %5 %8 %10 2
%20 = OpCompositeInsert %5 %9 %10 3

      OpReturn
      OpFunctionEnd
```

Variant shader:
```
      OpCapability Shader
      OpMemoryModel Logical GLSL450
      OpEntryPoint Vertex %11 "main"
 
; Types
 %2 = OpTypeVoid
 %3 = OpTypeFunction %2
 %4 = OpTypeFloat 32
 %5 = OpTypeVector %4 4

; Constants
 %6 = OpConstant %4 0
 %7 = OpConstant %4 1
 %8 = OpConstant %4 2
 %9 = OpConstant %4 3
%10 = OpConstantComposite %5 %6 %7 %8 %9

; main function
%11 = OpFunction %2 None %3
%12 = OpLabel

; Dynamically extracting vector components
%13 = OpVectorExtractDynamic %4 %10 %6
%14 = OpVectorExtractDynamic %4 %10 %7
%15 = OpVectorExtractDynamic %4 %10 %8
%16 = OpVectorExtractDynamic %4 %10 %9

; Dynamically inserting vector components
%17 = OpVectorInsertDynamic %5 %10 %6 %6
%18 = OpVectorInsertDynamic %5 %10 %7 %7
%19 = OpVectorInsertDynamic %5 %10 %8 %8
%20 = OpVectorInsertDynamic %5 %10 %9 %9

      OpReturn
      OpFunctionEnd
```

andreperezmaselco:
Fixes #3588.


source/fuzz/fuzzer_pass_vector_dynamic.h
```diff
@@ -0,0 +1,41 @@
...
+#define SOURCE_FUZZ_FUZZER_PASS_VECTOR_DYNAMIC_H_
+
+#include "source/fuzz/fuzzer_pass.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// Looks for OpCompositeExtract and OpCompositeInsert instructions, checks if
+// the composite operand is a vector, creates constants for their literal
+// operands and randomly applies the transformation.
```
afd, line 25: I would abbreviate a bit: "Looks for OpCompositeExtract/Inser instructions on vectors, and replaces them with OpVectorExtract/InsertDynamic."  The stuff about adding constants is more of an implementation detail.

source/fuzz/fuzzer_pass_vector_dynamic.cpp
```diff
@@ -0,0 +1,77 @@
...
+                 ->GetType(composite_instruction->type_id())
+                 ->AsVector()) {
+          continue;
+        }
+
+        // Randomly decide whether to try applying the transformation.
+        if (!GetFuzzerContext()->ChoosePercentage(
+                GetFuzzerContext()->GetChanceOfVectoringDynamic())) {
+          continue;
+        }
```
afd, line 59: I suggest moving this to the start of the innermost loop.  It's equivalent, and I think slightly nicer to do this check first, to avoid wasted work if in the end we decide not to proceed.

source/fuzz/fuzzer_pass_vector_dynamic.cpp
```diff
@@ -0,0 +1,77 @@
...
+
```
afd, line 60: Because in general OpCompositeExtract and OpCompositeInsert can have many indices, I think it would be nice to have an assertion to document the fact that - given that we know this is a vector - it is guaranteed to have just one such index - i.e., an assertion that the number of in operands is (whatever it is guaranteed to be for the two cases).

source/fuzz/fuzzer_context.cpp
```diff
@@ -95,6 +95,7 @@ const std::pair<uint32_t, uint32_t> kChanceOfSwappingConditionalBranchOperands =
     {10, 70};
 const std::pair<uint32_t, uint32_t> kChanceOfTogglingAccessChainInstruction = {
     20, 90};
+const std::pair<uint32_t, uint32_t> kChanceOfVectoringDynamic = {20, 90};
```
afd, line 98: How about `kChanceOfMakingVectorOperationDynamic`?  I don't like `VectoringDynamic`.  (Propagate change to getter method, etc.)

source/fuzz/transformation_vector_dynamic.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  // The composite must be a vector.
+  auto composite_instruction =
+      ir_context->get_def_use_mgr()->GetDef(instruction->GetSingleWordInOperand(
+          instruction->opcode() == SpvOpCompositeExtract ? 0 : 1));
+  if (!ir_context->get_type_mgr()
+           ->GetType(composite_instruction->type_id())
+           ->AsVector()) {
+    return false;
+  }
```
afd, line 52: I suggest you introduce a public static helper method in TransformationVectorDynamic that checks whether an instruction is suitable for being made dynamic - then you can avoid the duplication between here and the fuzzer pass.

source/fuzz/transformation_vector_dynamic.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  // The |instruction| literal operand must be defined as constant.
+  if (!fuzzerutil::MaybeGetIntegerConstant(
+          ir_context, transformation_context,
+          {instruction->GetSingleWordInOperand(
+              instruction->opcode() == SpvOpCompositeExtract ? 1 : 2)},
+          32, false, false)) {
+    return false;
+  }
```
afd, line 61: Here you could then call MaybeGetConstantForIndex (the static method I've suggested you add).

source/fuzz/fuzzer_pass_vector_dynamic.cpp
```diff
@@ -0,0 +1,77 @@
...
+        // Make sure the |instruction| literal operand is defined as constant.
+        // It will be used as operand of the vector dynamic instruction.
+        FindOrCreateIntegerConstant(
+            {instruction.GetSingleWordInOperand(
+                instruction.opcode() == SpvOpCompositeExtract ? 1 : 2)},
+            32, false, false);
```
afd, line 66: The index doesn't have to be unsigned.  I would like us to play more with mixing up sign.  Therefore I suggest you add a static method to the TransformationVectorDynamic class called MaybeGetConstantForIndex.  It should take the instruction as a const-ref, and:
The integer constant is allowed to be either signed or unsigned.

- Use fuzzerutil::MaybeGet to look for an unsigned version of the constant
- If that doesn't exist, use fuzzerutil::MaybeGet to look for a signed version
- Return 0 if that doesn't exist either.

Then if 0 gets returned, use FindOrCreate with ChooseEven() as the sign parameter, so signedness is random.

See corresponding review comment in the transformation class.

source/fuzz/transformation_vector_dynamic.cpp
```diff
@@ -0,0 +1,101 @@
...
+  // in reverse order in relation to the OpCompositeInsert corresponding
+  // operands.
+  if (instruction->opcode() == SpvOpCompositeInsert) {
+    std::swap(instruction->GetInOperand(0), instruction->GetInOperand(1));
+  }
+
+  // Sets the literal operand to the equivalent constant.
+  instruction->SetInOperand(
+      instruction->opcode() == SpvOpCompositeExtract ? 1 : 2,
+      {fuzzerutil::MaybeGetIntegerConstant(
```
afd, line 82: Again, you'd want to use the MaybeGetConstantForIndex function here.

source/fuzz/fuzzer_pass_vector_dynamic.h
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_vector_dynamic.cpp
andreperezmaselco: Yes, I think about it. But sometimes it will randomly choose inapplicable instructions. This changes the conditional probability. I guess it makes sense randomly choose applicable instructions.

andreperezmaselco: Done.

source/fuzz/fuzzer_context.cpp
andreperezmaselco: Do you think we could change the transformation name to `TransformationMakeVectorOperationDynamic`?

andreperezmaselco: Done.

source/fuzz/transformation_vector_dynamic.cpp
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_vector_dynamic.cpp
andreperezmaselco: Done.

source/fuzz/transformation_vector_dynamic.cpp
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_vector_dynamic.cpp
andreperezmaselco: Done.

source/fuzz/transformation_vector_dynamic.cpp
andreperezmaselco: Done.

source/fuzz/transformation_make_vector_operation_dynamic.h
```diff
@@ -0,0 +1,59 @@
...
+class TransformationMakeVectorOperationDynamic : public Transformation {
+ public:
+  explicit TransformationMakeVectorOperationDynamic(
+      const protobufs::TransformationMakeVectorOperationDynamic& message);
+
+  TransformationMakeVectorOperationDynamic(uint32_t instruction_result_id,
+                                           uint32_t constant_index_id);
+
+  // - |message_.instruction_result_id| must be an instruction result id of an
+  // OpCompositeExtract or OpCompositeInsert instruction.
```
Vasniktel, line 35: I would also add that an operand to this instruction must be a vector and that `constant_index_id` must be an integral instruction.

source/fuzz/fuzzer_context.cpp
```diff
@@ -98,6 +98,8 @@ const std::pair<uint32_t, uint32_t> kChanceOfSwappingConditionalBranchOperands =
     {10, 70};
 const std::pair<uint32_t, uint32_t> kChanceOfTogglingAccessChainInstruction = {
     20, 90};
+const std::pair<uint32_t, uint32_t> kChanceOfMakingVectorOperationDynamic = {
+    20, 90};
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/fuzzer_context.cpp
```diff
@@ -242,6 +244,8 @@ FuzzerContext::FuzzerContext(RandomGenerator* random_generator,
       ChooseBetweenMinAndMax(kChanceOfSwappingConditionalBranchOperands);
   chance_of_toggling_access_chain_instruction_ =
       ChooseBetweenMinAndMax(kChanceOfTogglingAccessChainInstruction);
+  chance_of_making_vector_operation_dynamic_ =
+      ChooseBetweenMinAndMax(kChanceOfMakingVectorOperationDynamic);
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/fuzzer_context.h
```diff
@@ -246,6 +246,9 @@ class FuzzerContext {
   uint32_t GetChanceOfTogglingAccessChainInstruction() {
     return chance_of_toggling_access_chain_instruction_;
   }
+  uint32_t GetChanceOfMakingVectorOperationDynamic() {
+    return chance_of_making_vector_operation_dynamic_;
+  }
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/fuzzer_context.h
```diff
@@ -381,6 +384,7 @@ class FuzzerContext {
   uint32_t chance_of_splitting_block_;
   uint32_t chance_of_swapping_conditional_branch_operands_;
   uint32_t chance_of_toggling_access_chain_instruction_;
+  uint32_t chance_of_making_vector_operation_dynamic_;
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1561,6 +1562,19 @@ message TransformationToggleAccessChainInstruction {
...
+message TransformationMakeVectorOperationDynamic {
+
+  // A transformation that replaces the OpCompositeExtract and OpCompositeInsert
+  // instructions with the OpVectorExtractDynamic and OpVectorInsertDynamic instructions.
+
+  // The composite instruction result id.
+  uint32 instruction_result_id = 1;
+
+  // The constant index instruction result id.
+  uint32 constant_index_id = 2;
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/fuzz/transformation_make_vector_operation_dynamic.cpp
```diff
@@ -0,0 +1,108 @@
...
+  if (!IsVectorOperation(ir_context, instruction)) {
+    return false;
+  }
+
+  // |constant_index| must be defined as an integer constant.
+  auto constant_index = ir_context->get_constant_mgr()->FindDeclaredConstant(
+      message_.constant_index_id());
+  if (!constant_index || !constant_index->AsIntConstant()) {
+    return false;
+  }
```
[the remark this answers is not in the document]
andreperezmaselco: You're right.
Done.

source/fuzz/transformation_make_vector_operation_dynamic.h
andreperezmaselco: Done.

source/fuzz/transformation_make_vector_operation_dynamic.cpp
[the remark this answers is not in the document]
paulthomson: It is interesting because this transformation is designed assuming that "constant_index" is actually a constant, but I guess we indeed don't need to enforce that.

source/fuzz/transformation_make_vector_operation_dynamic.cpp
```diff
@@ -0,0 +1,111 @@
...
+  if (!IsVectorOperation(ir_context, instruction)) {
+    return false;
+  }
+
+  // |constant_index_instruction| must be defined as an integer instruction.
+  auto constant_index_instruction =
+      ir_context->get_def_use_mgr()->GetDef(message_.constant_index_id());
+  if (!constant_index_instruction ||
+      !ir_context->get_type_mgr()
+           ->GetType(constant_index_instruction->type_id())
```
paulthomson, line 50: Unlikely, but I suppose `constant_index_id` might indeed refer to an instruction, but it might not have a type id. So we should check that in the `if` first.

test/fuzz/transformation_make_vector_operation_dynamic_test.cpp
```diff
@@ -0,0 +1,328 @@
...
+      transformation.IsApplicable(context.get(), transformation_context));
+
+  transformation = TransformationMakeVectorOperationDynamic(38, 11);
+  ASSERT_TRUE(
+      transformation.IsApplicable(context.get(), transformation_context));
+
+  transformation = TransformationMakeVectorOperationDynamic(39, 10);
+  ASSERT_TRUE(
+      transformation.IsApplicable(context.get(), transformation_context));
+}
```
paulthomson, line 144: Interesting. We normally apply the transformations and then compare the module to some SPIR-V assembly and check that it is valid. I feel like you might as well do this? But I can see that you do this kind of thing below in a separate test.

test/fuzz/transformation_make_vector_operation_dynamic_test.cpp
```diff
@@ -0,0 +1,328 @@
...
+  FactManager fact_manager;
+  spvtools::ValidatorOptions validator_options;
+  TransformationContext transformation_context(&fact_manager,
+                                               validator_options);
+
+  auto transformation = TransformationMakeVectorOperationDynamic(22, 9);
+  transformation.Apply(context.get(), &transformation_context);
+
+  transformation = TransformationMakeVectorOperationDynamic(23, 10);
+  transformation.Apply(context.get(), &transformation_context);
```
paulthomson, line 217: I think it probably makes sense to ASSERT that each transformation is applicable. That way, we might get a test failure sooner, closer to the actual issue.

source/fuzz/transformation_make_vector_operation_dynamic.cpp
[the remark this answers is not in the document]
andreperezmaselco: Yes. Initially I assumed it was a constant because it is related to a literal, but it can be the result id of any integer instruction.

source/fuzz/transformation_make_vector_operation_dynamic.cpp
andreperezmaselco: You're right.
Done.

test/fuzz/transformation_make_vector_operation_dynamic_test.cpp
andreperezmaselco: It is in the `Apply` test.

test/fuzz/transformation_make_vector_operation_dynamic_test.cpp
andreperezmaselco: Done.

PR #3598, opened 2020-07-27
spirv-fuzz: TransformationReplaceAddSubMulWithCarryingExtended

### LINKED ISSUE #3577: spirv-fuzz: replace add, sub and mul with carrying / extended versions
Instructions:

- OpIAddCarry
- OpISubBorrow
- OpUMulExtended
- OpSMulExtended

behave identically to OpIAdd, OpISub, OpUMul and OpSMul, except that they return a pair of results; the first element of the pair is the same result that would be obtained from the normal operation.

Thus we could e.g. replace OpIAdd with OpIAddCarry into a fresh id + an extract from that fresh id into the original result of OpIAdd, and similar for the other instructions.

antnkarp:
Replaces OpIAdd with OpIAddCarry, OpISub with OpISubBorrow, OpIMul with
OpUMulExtended or OpSMulExtended and stores the result into a fresh_id
representing a structure. Extracts the first element of the result into
the original result_id. This value is the same as the result of the
original instruction.

Fixes #3577

### DIFF
--- a/source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
+++ b/source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
@@ -51,17 +51,14 @@ void FuzzerPassReplaceAddsSubsMulsWithCarryingExtended::Apply() {
         // of the instruction.
         auto instruction_opcode = instruction.opcode();
 
-        switch (instruction_opcode) {
-          case SpvOpIAdd:
-          case SpvOpISub:
-          case SpvOpIMul:
-            if (!TransformationReplaceAddSubMulWithCarryingExtended::
-                    IsInstructionSuitable(GetIRContext(), instruction)) {
-              continue;
-            }
-            break;
-          default:
-            continue;
+        if (instruction_opcode != SpvOpIAdd &&
+            instruction_opcode != SpvOpISub &&
+            instruction_opcode != SpvOpIMul) {
+          continue;
+        }
+        if (!TransformationReplaceAddSubMulWithCarryingExtended::
+                IsInstructionSuitable(GetIRContext(), instruction)) {
+          continue;
         }
 
         // Get the id of the required struct type. We know that the the types of
--- a/source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
+++ b/source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
@@ -32,17 +32,40 @@ FuzzerPassReplaceAddsSubsMulsWithCarryingExtended::
     ~FuzzerPassReplaceAddsSubsMulsWithCarryingExtended() = default;
 
 void FuzzerPassReplaceAddsSubsMulsWithCarryingExtended::Apply() {
-  for (auto& function : *GetIRContext()->module()) {
-    for (auto& block : function) {
-      for (auto& inst : block) {
-        (void)inst;
-        if (GetFuzzerContext()->ChoosePercentage(
-                GetFuzzerContext()
-                    ->GetChanceOfReplacingAddSubMulWithCarryingExtended())) {
-        }
-      }
+  GetIRContext()->module()->ForEachInst([this](opt::Instruction* instruction) {
+    if (!GetFuzzerContext()->ChoosePercentage(
+            GetFuzzerContext()
+                ->GetChanceOfReplacingAddSubMulWithCarryingExtended())) {
+      return;
     }
-  }
+    /*
+     * std::vector<uint32_t> component_type_ids;
+uint32_t component_1_type_id =
+ir_context->get_def_use_mgr()
+    ->GetDef(instruction->GetSingleWordOperand(2))
+    ->type_id();
+component_type_ids.push_back(component_1_type_id);
+uint32_t component_2_type_id =
+ir_context->get_def_use_mgr()
+    ->GetDef(instruction->GetSingleWordOperand(3))
+    ->type_id();
+component_type_ids.push_back(component_2_type_id);
+
+fuzzerutil::UpdateModuleIdBound(ir_context, message_.struct_fresh_id());
+
+uint32_t struct_type_id =
+fuzzerutil::MaybeGetStructType(ir_context, component_type_ids);
+if (struct_type_id == 0) {
+fuzzerutil::AddStructType(ir_context, message_.struct_type_fresh_id(),
+                        component_type_ids);
+struct_type_id = message_.struct_type_fresh_id();
+}
+     *
+     */
+          }
 }
 }  // namespace fuzz
 }  // namespace spvtools
+}
+}  // namespace fuzz
+}  // namespace fuzz
--- a/source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
+++ b/source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
@@ -20,6 +20,10 @@
 namespace spvtools {
 namespace fuzz {
 
+namespace {
+const uint32_t kArithmeticInstructionIndexLeftInOperand = 0;
+}
+
 FuzzerPassReplaceAddsSubsMulsWithCarryingExtended::
     FuzzerPassReplaceAddsSubsMulsWithCarryingExtended(
         opt::IRContext* ir_context,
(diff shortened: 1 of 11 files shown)


source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,227 @@
...
+  auto instruction_composite_extract =
+      original_instruction->InsertBefore(MakeUnique<opt::Instruction>(
+          ir_context, SpvOpCompositeExtract, operand_1_type_id,
+          message_.result_id(),
+          opt::Instruction::OperandList(
+              {{SPV_OPERAND_TYPE_ID, {message_.struct_fresh_id()}},
+               {SPV_OPERAND_TYPE_LITERAL_INTEGER,
+                {kOpCompositeExtractIndexLowOrderBits}}})));
+
+  switch (original_instruction_opcode) {
```
afd, line 152: I didn't fully check, but I *think* you could replace all this with:

SpvOpcode new_instruction_opcode;
switch (original_instruction_opcode) {
  // work out what the new opcode is depending on the existing opcode and the signedness, if relevant.
  // assert false in a default case, and assign new_instruction_opcode to SpvOpMax in that case (to keep compilers happy)
}

instruction_composite_extract->InsertBefore(....); <--- A single constructor call that takes new_instruction_opcode as an argument.

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.h
```diff
@@ -0,0 +1,63 @@
...
+  explicit TransformationReplaceAddSubMulWithCarryingExtended(
+      const protobufs::TransformationReplaceAddSubMulWithCarryingExtended&
+          message);
+
+  explicit TransformationReplaceAddSubMulWithCarryingExtended(
+      uint32_t struct_fresh_id, uint32_t struct_type_id, uint32_t result_id);
+
+  // - |message_.struct_fresh_id| must be fresh.
+  // - |message_.struct_type_id| must refer to the proper type of a struct
+  //   holding the intermediate result.
```
afd, line 38: Not sure what you mean by "proper type", and change "holding the intermediate result" to something like "suitable for holding the intermediate result".

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,194 @@
...
+        new_instruction_opcode = SpvOpUMulExtended;
+      } else {
+        new_instruction_opcode = SpvOpSMulExtended;
+      }
+      break;
+    default:
+      assert(false);
+      new_instruction_opcode = SpvOpMax;
+      // No uninitialized variables left (to keep compilers happy).
+      break;
```
paulthomson, line 114: I think you can just replace this with:

```
assert(false);
return;
```

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  // Insert the OpCompositeExtract. This instruction takes the first component
+  // of the struct which represents low-order bits of the operation. This the
+  // original result.
+  auto instruction_composite_extract =
```
paulthomson, line 120: Nit: consider inserting the new Add/Sub/Mul instruction first, before the original instruction. Then insert the extract, also before the original instruction.

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,194 @@
...
+void TransformationReplaceAddSubMulWithCarryingExtended::Apply(
+    opt::IRContext* ir_context, TransformationContext* /*unused*/) const {
+  // |message_.struct_fresh_id| must be fresh.
+  assert(fuzzerutil::IsFreshId(ir_context, message_.struct_fresh_id()) &&
+         "|message_.struct_fresh_id| must be fresh");
+
+  auto original_instruction =
+      ir_context->get_def_use_mgr()->GetDef(message_.result_id());
+  auto original_instruction_opcode = original_instruction->opcode();
+  auto operand_type_id = original_instruction->type_id();
```
paulthomson, line 84: Consider removing these, since `original_instruction->opcode()` and `original_instruction->type_id()` are fairly simple.

test/fuzz/transformation_replace_add_sub_mul_with_carrying_extended_test.cpp
```diff
@@ -0,0 +1,267 @@
...
+                                               validator_options);
+
+  // Bad: |struct_fresh_id| must be fresh.
+  auto transformation_bad_1 =
+      TransformationReplaceAddSubMulWithCarryingExtended(34, 38, 28);
+
+  ASSERT_FALSE(
+      transformation_bad_1.IsApplicable(context.get(), transformation_context));
+
+  // Bad: The transformation cannot be applied to an instruction OpSDiv.
```
paulthomson, line 118: Add test for when instruction id is not found.

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include "source/fuzz/fuzzer_util.h"
+
+namespace spvtools {
+namespace fuzz {
+
+namespace {
+const uint32_t kOpCompositeExtractIndexLowOrderBits = 0;
+const uint32_t kArithmeticInstructionIndexLeftInOperand = 0;
+const uint32_t kArithmeticInstructionIndexRightInOperand = 1;
+const uint32_t kOpTypeIndexSignedness = 2;
```
paulthomson, line 26: Nice

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,204 @@
...
+      message_.result_id(),
+      opt::Instruction::OperandList(
+          {{SPV_OPERAND_TYPE_ID, {message_.struct_fresh_id()}},
+           {SPV_OPERAND_TYPE_LITERAL_INTEGER,
+            {kOpCompositeExtractIndexLowOrderBits}}})));
+
+  // Remove the original instruction.
+  ir_context->KillInst(original_instruction);
+
+  ir_context->InvalidateAnalysesExceptFor(opt::IRContext::kAnalysisNone);
```
Vasniktel, line 139: Add a comment here saying "We have modified the module so most analyzes are now invalid."

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,204 @@
...
+}
+
+bool TransformationReplaceAddSubMulWithCarryingExtended::IsInstructionSuitable(
+    opt::IRContext* ir_context, const opt::Instruction* instruction) {
```
Vasniktel, line 143: Google C++ style guide recommends using a const reference if the variable is not modified. https://google.github.io/styleguide/cppguide.html

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,204 @@
...
+      ir_context->get_def_use_mgr()
+          ->GetDef(instruction->GetSingleWordInOperand(
+              kArithmeticInstructionIndexLeftInOperand))
+          ->type_id();
+
+  uint32_t operand_2_type_id =
+      ir_context->get_def_use_mgr()
+          ->GetDef(instruction->GetSingleWordInOperand(
+              kArithmeticInstructionIndexRightInOperand))
+          ->type_id();
```
Vasniktel, line 165: Operands can be vectors.

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,204 @@
...
+    case SpvOpIAdd:
+    case SpvOpISub:
+      // Both types of the operands and the result type must be unsigned.
+      instruction_signedness =
+          ir_context->get_def_use_mgr()
+              ->GetDef(instruction->type_id())
+              ->GetSingleWordOperand(kOpTypeIndexSignedness);
+      if (instruction_signedness != 0) {
+        return false;
+      }
```
Vasniktel, line 188: You can move `instruction_signedness` into the `switch` if you use this pattern.
```cpp
switch(...) {
  case ...: {
    ...
  } break;
}
```

test/fuzz/CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ if (${SPIRV_BUILD_FUZZER})
           transformation_permute_function_parameters_test.cpp
           transformation_permute_phi_operands_test.cpp
           transformation_push_id_through_variable_test.cpp
+          transformation_replace_add_sub_mul_with_carrying_extended_test.cpp
```
Vasniktel, line 71: I would be grateful if you could sort these filenames alphabetically. Someone (probably me :) did something wrong.

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
Vasniktel: Return type of the instruction can be a vector too.

source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        // Get the id of the required struct type. We know that the the types of
+        // the operands are the same, so we can use one operand type id twice.
+        uint32_t operand_type_id =
+            GetIRContext()
+                ->get_def_use_mgr()
+                ->GetDef(instruction.GetSingleWordInOperand(
+                    kArithmeticInstructionIndexLeftInOperand))
+                ->type_id();
```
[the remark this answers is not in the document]
antnkarp: If the operand is a vector, we create a struct that has two members each of them being a vector and results are computed per component (according to the specification). FindOrCreateStructType uses AddStructType that can operate on vectors. It requires type_ids, so I think that extracting type_id, instead of using the type manager is a correct way to add a struct. Please correct me if I'm wrong.

source/fuzz/transformation.cpp
```diff
@@ -215,6 +216,10 @@ std::unique_ptr<Transformation> Transformation::FromMessage(
         kRecordSynonymousConstants:
       return MakeUnique<TransformationRecordSynonymousConstants>(
           message.record_synonymous_constants());
+    case protobufs::Transformation::TransformationCase::
+        kReplaceAddSubMulWithCarryingExtended:
+      return MakeUnique<TransformationReplaceAddSubMulWithCarryingExtended>(
+          message.replace_add_sub_mul_with_carrying_extended());
```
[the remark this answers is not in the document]
antnkarp: No problem :)

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,204 @@
...
+    case SpvOpISub:
+    case SpvOpIMul:
+      if (!TransformationReplaceAddSubMulWithCarryingExtended::
+              IsInstructionSuitable(ir_context, instruction))
+        return false;
+      break;
+    default:
+      return false;
+  }
+  return true;
```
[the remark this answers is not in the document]
antnkarp: Thanks, I fixed it.

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,204 @@
...
+          {{SPV_OPERAND_TYPE_ID,
+            {original_instruction->GetSingleWordInOperand(
+                kArithmeticInstructionIndexLeftInOperand)}},
+           {SPV_OPERAND_TYPE_ID,
+            {original_instruction->GetSingleWordInOperand(
+                kArithmeticInstructionIndexRightInOperand)}}})));
+
+  // Insert the OpCompositeExtract after the added instruction. This instruction
+  // takes the first component of the struct which represents low-order bits of
+  // the operation. This the original result.
```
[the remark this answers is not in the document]
antnkarp: I am not sure what you mean by that.

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
antnkarp: If operands are vectors then the result is also a vector. And both type_ids of operands and the result type_id must be equal.

source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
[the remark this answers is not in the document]
Vasniktel: Yes, you are right. It seems I've added this comment a bit prematurely.

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
Vasniktel: Yes, you are right! Sorry for the confusion.

Vasniktel:
Builds fail because of the unused `kOpTypeIndexSignedness` variable.

antnkarp:
> Builds fail because of the unused `kOpTypeIndexSignedness` variable.

Thanks, I'll fix it.

source/fuzz/transformation_replace_add_sub_mul_with_carrying_extended.cpp
```diff
@@ -0,0 +1,242 @@
...
+    case SpvOpISub:
+    case SpvOpIMul:
+      if (!TransformationReplaceAddSubMulWithCarryingExtended::
+              IsInstructionSuitable(ir_context, *instruction)) {
+        return false;
+      }
+      break;
+    default:
+      return false;
+  }
```
paulthomson, line 69: I don't think you need the switch.

source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
```diff
@@ -0,0 +1,85 @@
...
+          case SpvOpISub:
+          case SpvOpIMul:
+            if (!TransformationReplaceAddSubMulWithCarryingExtended::
+                    IsInstructionSuitable(GetIRContext(), instruction)) {
+              continue;
+            }
+            break;
+          default:
+            continue;
+        }
```
paulthomson, line 65: Same here; I don't think you need the switch.

Issue #3095, opened 2019-12-10
spirv-fuzz: have outliner use some auxiliary transformations to handle more complex control flow


afd:
Following dicussion on #3078, the outlining transformation has been made more restrictive with respect to the control flow that it handles.

[These tests](https://github.com/KhronosGroup/SPIRV-Tools/files/3944974/disabled_tests.txt) were removed.  The outliner should be enhanced so that it can use other transformations - such as block splitting and merging - to be capable of applying outlining in the contexts captured by the tests what were removed.

afd:
@stefanomil Can you take a look at this one?  Please reply and I can assign it.  We'll discuss it with you in some detail.

stefanomil:
Yes

afd:
Comments following our discussion about this:

In the fuzzer pass:
- If the entry block starts with OpVariable, split the entry block after the final OpVariable and make the new block be the entry block.
- If the entry block is a loop header, check whether there is already a loop pre-header - i.e., a single predecessor not in the loop that branches unconditionally to the loop header.  If so, use this.  If not, apply a (new) transformation to add a pre-header to the loop and make the entry block be the pre-header.
- If the exit block is a merge block, split the merge block and make the exit block be the second block in the pair.

When populating "candidate exit blocks", skip any blocks that are continue targets.

PR #3599, opened 2020-07-27
spirv-fuzz: Transformation to add loop preheader

stefanomil:
This PR introduces TransformationAddLoopPreheader, which, given
a loop header and enough fresh ids, adds a loop preheader, updating
all the references so that this new block is the only out-of-loop
predecessor of the header, which branches unconditionally to the
header.

See the discussion in #3095 .

### DIFF
--- a/source/fuzz/transformation_add_loop_preheader.cpp
+++ b/source/fuzz/transformation_add_loop_preheader.cpp
@@ -35,7 +35,7 @@ TransformationAddLoopPreheader::TransformationAddLoopPreheader(
 
 bool TransformationAddLoopPreheader::IsApplicable(
     opt::IRContext* ir_context,
-    const TransformationContext& transformation_context) const {
+    const TransformationContext& /* unused */) const {
   // |message_.loop_header_block()| must be the id of a loop header block.
   opt::BasicBlock* loop_header_block =
       fuzzerutil::MaybeFindBlock(ir_context, message_.loop_header_block());
@@ -52,7 +52,7 @@ bool TransformationAddLoopPreheader::IsApplicable(
 
   // If the block only has one predecessor outside of the loop (and thus 2 in
   // total), then no additional fresh ids are necessary.
-  if (ir_context->cfg()->preds(message_.fresh_id()).size() == 2) {
+  if (ir_context->cfg()->preds(message_.loop_header_block()).size() == 2) {
     return true;
   }
 
--- a/source/fuzz/transformation_add_loop_preheader.cpp
+++ b/source/fuzz/transformation_add_loop_preheader.cpp
@@ -50,9 +50,18 @@ bool TransformationAddLoopPreheader::IsApplicable(
     return false;
   }
 
+  size_t num_predecessors =
+      ir_context->cfg()->preds(message_.loop_header_block()).size();
+
+  // The block must have at least 2 predecessors (the back-edge block and
+  // another predecessor outside of the loop)
+  if (num_predecessors < 2) {
+    return false;
+  }
+
   // If the block only has one predecessor outside of the loop (and thus 2 in
   // total), then no additional fresh ids are necessary.
-  if (ir_context->cfg()->preds(message_.loop_header_block()).size() == 2) {
+  if (num_predecessors == 2) {
     return true;
   }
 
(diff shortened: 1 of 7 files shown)


test/fuzz/transformation_add_loop_preheader_test.cpp
```diff
@@ -0,0 +1,273 @@
...
+               OpReturn
+               OpFunctionEnd
+  )";
+
+  ASSERT_TRUE(IsEqual(env, after_transformations, context.get()));
+}
+
+}  // namespace
+}  // namespace fuzz
+}  // namespace spvtools
```
afd, line 273: Newline at end of file always, please.

source/fuzz/transformation_add_loop_preheader.cpp
```diff
@@ -0,0 +1,218 @@
...
+            use_inst->opcode() != SpvOpLoopMerge) {
+          return;
+        }
+
+        // Update the reference.
+        use_inst->SetOperand(use_index, {message_.fresh_id()});
+      });
+
+  // The back edge block should have been found.
+  assert(back_edge_block_id);
```
afd, line 124: Please turn the comment into a `&& "..."` string for the assertion, so that when it fails you see the message.  I like to have a `&& "..."` string for every assertion.

source/fuzz/transformation_add_loop_preheader.cpp
```diff
@@ -0,0 +1,218 @@
...
+      fuzzerutil::MaybeFindBlock(ir_context, message_.loop_header_block());
+
+  auto dominator_analysis =
+      ir_context->GetDominatorAnalysis(loop_header->GetParent());
+
+  uint32_t back_edge_block_id = 0;
+
+  // Update the branching instructions of the out-of-loop predecessors of the
+  // header. Set |back_edge_block_id| to be the id of the block from which the
+  // back edge starts.
```
afd, line 94: "the id of the block from which the back edge starts" -> "the id of the back-edge block" (that's its technical name).

source/fuzz/transformation_add_loop_preheader.cpp
```diff
@@ -0,0 +1,218 @@
...
+      loop_header->id(),
+      [this, &ir_context, &dominator_analysis, &loop_header,
+       &back_edge_block_id](opt::Instruction* use_inst, uint32_t use_index) {
+
+        if (dominator_analysis->Dominates(loop_header->GetLabelInst(),
+                                          use_inst)) {
+          // If |use_inst| is a branch instruction dominated by the header, the
+          // block containing it is the back-edge block.
+          if (use_inst->IsBranch()) {
+            back_edge_block_id = ir_context->get_instr_block(use_inst)->id();
```
afd, line 105: I suggest preceding this by an assertion that back_edge_block_id is 0, since there can only be one back-edge block.  This is an example of something that we don't really need to assert about - since we assume valid SPIR-V - but where I feel that the assertion would help someone reading the code understand that indeed this assignment should happen exactly once.

source/fuzz/transformation_add_loop_preheader.cpp
afd: ... and it's great that you have this assertion BTW.

source/fuzz/transformation_add_loop_preheader.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+  // Update the OpPhi instructions and, if there is more than one out-of-loop
+  // predecessor, add necessary OpPhi instructions so the preheader.
+  loop_header->ForEachPhiInst([this, &ir_context, &preheader,
+                               &back_edge_block_id,
+                               &phi_ids_used](opt::Instruction* phi_inst) {
+
+    // The loop header must have at least 2 incoming edges (the back edge, and
+    // at least one from outside the loop).
+    assert(phi_inst->NumInOperands() >= 4);
```
afd, line 141: Can you ensure you have test showing that the transformation is not applicable if you apply it to a loop with only the back-edge as a predecessor - i.e. a loop that is unreachable in the control flow graph?

test/fuzz/transformation_add_loop_preheader_test.cpp
stefanomil: Added

source/fuzz/transformation_add_loop_preheader.cpp
stefanomil: Done

source/fuzz/transformation_add_loop_preheader.cpp
stefanomil: Changed

source/fuzz/transformation_add_loop_preheader.cpp
stefanomil: Done

source/fuzz/transformation_add_loop_preheader.cpp
stefanomil: Added a test; I found out that I did not check for this in the IsApplicable method, so I added a check to it.

PR #3606, opened 2020-07-28
spirv-fuzz: add FuzzerPassAddCompositeInserts

### LINKED ISSUE #2859: spirv-fuzz: Add OpCompositeInsert instructions to create synonyms.
Similar to the fuzzer pass that adds extracts or vector shuffles, this pass should add OpCompositeInsert instructions to the module.

Each insert yields several synonym facts: every component of the result is synonymous with the components of the original id, except for the inserted component which is synonymous with the given operand.

antnkarp:
Adds FuzzerPassAddCompositeInserts, which randomly adds new
OpCompositeInsert instructions. Each OpCompositeInsert instruction
yields a copy of an original composite with one subcomponent replaced
with an existing or newly added object. Synonym facts are added for the
unchanged components in the original and added composite, and for the
replaced subcomponent and the object, if possible.

Fixes #2859


source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+        // It must be valid to insert an OpCompositeInsert instruction
+        // before |instruction_iterator|.
+        if (!fuzzerutil::CanInsertOpcodeBeforeInstruction(
+                SpvOpCompositeInsert, instruction_iterator)) {
+          return;
+        }
+
+        // Look for available constants that have the type OpCompositeConstant.
+        std::vector<opt::Instruction*> available_composite_constants =
```
paulthomson, line 58: Could we consider _any_ id that has a composite type?

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,164 @@
...
+                    opt::IRContext* ir_context,
+                    opt::Instruction* instruction) -> bool {
+
+                  if (instruction->opcode() != SpvOpConstantComposite) {
+                    return false;
+                  }
+
+                  return fuzzerutil::IdIsAvailableBeforeInstruction(
+                      ir_context,
+                      FindInstruction(instruction_descriptor, ir_context),
```
paulthomson, line 71: I suggest that you avoid using `instruction_descriptor` throughout; you don't need to use it until you construct your transformation. You have `instruction_iterator` already. The instruction descriptor's only purpose is to allow the instruction to be found within a transformation so that the transformation can still be applied later even if the original module has changed slightly.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,164 @@
...
+        auto constant_to_replace_id =
+            available_composite_constant->GetSingleWordInOperand(
+                index_to_replace);
+        auto constant_to_replace_type_id = GetIRContext()
+                                               ->get_def_use_mgr()
+                                               ->GetDef(constant_to_replace_id)
+                                               ->type_id();
+
+        // Look for available constants that have the same type id as the
+        // |constant_to_replace|.
```
paulthomson, line 98: Do you need to only consider constants? Could you consider any id of the appropriate type?Actually, it looks like this code does not enforce that the type is a constant, so just be sure to update this comment and the variable names so that you do not imply that everything is a constant.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,164 @@
...
+        // replaces |constant_to_replace| with |available_constant|.
+        FindInstruction(instruction_descriptor, GetIRContext())
+            ->InsertBefore(MakeUnique<opt::Instruction>(
+                GetIRContext(), SpvOpCompositeInsert,
+                available_composite_constant->type_id(), new_result_id,
+                opt::Instruction::OperandList(
+                    {{SPV_OPERAND_TYPE_ID, {available_constant->result_id()}},
+                     {SPV_OPERAND_TYPE_ID,
+                      {available_composite_constant->result_id()}},
+                     {SPV_OPERAND_TYPE_LITERAL_INTEGER, {index_to_replace}}})));
```
paulthomson, line 134: Fuzzer passes must not directly change the module. They can only change the module via transformations. The output from spirv-fuzz is a list of transformations that change the module. We also output the final modified module, but the most important output is the list of transformations. We need to be able to replay the list of transformations and replay a subset of the list of transformations.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+        fuzzerutil::UpdateModuleIdBound(GetIRContext(), new_result_id);
+        GetIRContext()->InvalidateAnalysesExceptFor(
+            opt::IRContext::kAnalysisNone);
+
+        // Every every element which hasn't been changed in the copy is
```
paulthomson, line 140: As above, the following code must be in a transformation.

test/fuzz/fuzzer_pass_add_composite_inserts_test.cpp
```diff
@@ -0,0 +1,146 @@
...
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_add_composite_inserts.h"
+#include "source/fuzz/pseudo_random_generator.h"
+#include "test/fuzz/fuzz_test_util.h"
+
+namespace spvtools {
+namespace fuzz {
+namespace {
+TEST(FuzzerPassAddCompositeInsertsTest, BasicScenarios) {
```
paulthomson, line 22: This is a nice idea. Unfortunately, as explained above already, you will need to use a transformation and so you will need to add tests for the transformation.

test/fuzz/fuzzer_pass_add_composite_inserts_test.cpp
```diff
@@ -0,0 +1,146 @@
...
+  const auto consumer = nullptr;
+  const auto context = BuildModule(env, consumer, shader, kFuzzAssembleOption);
+  ASSERT_TRUE(IsValid(env, context.get()));
+
+  FactManager fact_manager;
+  spvtools::ValidatorOptions validator_options;
+  TransformationContext transformation_context(&fact_manager,
+                                               validator_options);
+  auto prng = MakeUnique<PseudoRandomGenerator>(12);
+  FuzzerContext fuzzer_context(prng.get(), 100);
```
paulthomson, line 83: Unfortunately, the RNG is not guaranteed to be deterministic across all platforms. The exact behavior is implementation defined. Thus, we can't (currently) make tests like these. Plus, tests like these often end up being too sensitive, so if we change something, a lot of tests start failing. We will shortly improve the RNG so that we can override it deterministically and then we can start writing fuzzer pass tests.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,116 @@
...
+      ir_context->get_type_mgr()->GetType(composite->type_id());
+  if (!fuzzerutil::IsCompositeType(composite_type)) {
+    return false;
+  }
+
+  // |message_.index| must refer to a valid index.
+  if (fuzzerutil::WalkCompositeTypeIndices(ir_context, composite->type_id(),
+                                           message_.index()) == 0) {
+    return false;
+  }
```
Vasniktel, line 61: This duplicates the functionality below.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+  // The type id of the object at |message_.object_id| and the type id of the
+  // component of the composite at index |message_.index| must be the same.
+  auto component_to_be_replaced_type_id = fuzzerutil::WalkCompositeTypeIndices(
+      ir_context, composite->type_id(), message_.index());
+  if (component_to_be_replaced_type_id == 0) {
+    return false;
+  }
+  auto object_type_id =
+      ir_context->get_def_use_mgr()->GetDef(message_.object_id())->type_id();
```
Vasniktel, line 71: Make sure to check that `message_.object_id` is valid.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,116 @@
...
+  if (component_to_be_replaced_type_id != object_type_id) {
+    return false;
+  }
+
+  // |message_.instruction_to_insert_before| must refer to a valid
+  // instruction.
+  auto instruction_to_insert_before =
+      FindInstruction(message_.instruction_to_insert_before(), ir_context);
+
+  return instruction_to_insert_before != nullptr;
```
Vasniktel, line 82: Also, check that we can insert `OpCompositeInsert` before `instruction_descriptor`.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,116 @@
...
+      ir_context, composite->type_id(), message_.index());
+  if (component_to_be_replaced_type_id == 0) {
+    return false;
+  }
+  auto object_type_id =
+      ir_context->get_def_use_mgr()->GetDef(message_.object_id())->type_id();
+
+  if (component_to_be_replaced_type_id != object_type_id) {
+    return false;
+  }
```
Vasniktel, line 75: It might be also better to assert that `object_type_id` is not a result id of a pointer type if you decide to ignore pointers for now. Also, add a TODO referring the issue and explaining that we ignore pointer components in structs.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,116 @@
...
+         "|message_.fresh_id| must be fresh");
+
+  opt::Instruction::OperandList in_operands;
+  in_operands.push_back({SPV_OPERAND_TYPE_ID, {message_.object_id()}});
+  in_operands.push_back({SPV_OPERAND_TYPE_ID, {message_.composite_id()}});
+  for (auto an_index : message_.index()) {
+    in_operands.push_back({SPV_OPERAND_TYPE_LITERAL_INTEGER, {an_index}});
+  }
+  auto composite =
+      ir_context->get_def_use_mgr()->GetDef(message_.composite_id());
```
Vasniktel, line 98: As a suggestion, you can use `fuzzerutil::GetTypeId`.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+  FindInstruction(message_.instruction_to_insert_before(), ir_context)
+      ->InsertBefore(MakeUnique<opt::Instruction>(
+          ir_context, SpvOpCompositeInsert, composite->type_id(),
+          message_.fresh_id(), in_operands));
```
Vasniktel, line 103: Not crucial but we can `std::move(in_operands)` here.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+  fuzzerutil::UpdateModuleIdBound(ir_context, message_.fresh_id());
+  ir_context->InvalidateAnalysesExceptFor(opt::IRContext::kAnalysisNone);
```
Vasniktel, line 106: Add a comment here saying "We have modified the module so most analyzes are now invalid."

source/fuzz/transformation_composite_insert.h
```diff
@@ -0,0 +1,62 @@
...
+  TransformationCompositeInsert(
+      const protobufs::InstructionDescriptor& instruction_to_insert_before,
+      uint32_t fresh_id, uint32_t composite_id, uint32_t object_id,
+      std::vector<uint32_t>&& index);
+
+  // - |message_.fresh_id| must be fresh.
+  // - |message_.composite_id| must refer to an existing composite value.
+  // - |message_.index| must refer to a valid index in the composite.
+  // - The type id of the object and the type id of the component of the
+  //   composite.
```
Vasniktel, line 40: The dot looks like a typo.

source/fuzz/transformation_composite_insert.h
```diff
@@ -0,0 +1,62 @@
...
+  //   at index |message_.index| must be the same.
+  // - |message_.instruction_to_insert_before| must refer to a valid
+  //   instruction.
```
Vasniktel, line 43: Add "It must be possible to insert OpCompositeInsert before |instruction_to_insert_before|.

source/fuzz/transformation_composite_insert.h
```diff
@@ -0,0 +1,62 @@
...
+      const protobufs::InstructionDescriptor& instruction_to_insert_before,
+      uint32_t fresh_id, uint32_t composite_id, uint32_t object_id,
+      std::vector<uint32_t>&& index);
+
+  // - |message_.fresh_id| must be fresh.
+  // - |message_.composite_id| must refer to an existing composite value.
+  // - |message_.index| must refer to a valid index in the composite.
+  // - The type id of the object and the type id of the component of the
+  //   composite.
+  //   at index |message_.index| must be the same.
```
Vasniktel, line 41: Add a TODO with an issue saying that pointer types are not supported right now if you decide that the approach of ignoring pointers is better right now.

source/fuzz/transformation_composite_insert.h
```diff
@@ -0,0 +1,62 @@
...
+  // - |message_.instruction_to_insert_before| must refer to a valid
+  //   instruction.
+  bool IsApplicable(
+      opt::IRContext* ir_context,
+      const TransformationContext& transformation_context) const override;
+
+  // Adds a instruction OpCompositeInsert which creates a new composite from an
+  // existing composite, with an element inserted.
```
Vasniktel, line 49: It might be good to mention the arguments of the constructor explicitly here. Something like "Adds an instruction OpCompositeInsert before |instruction_to_insert_before| which creates a new composite from |composite_id| by inserting |object_id| at the specified |index|."

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+        // Look for available values that have composite type.
+        std::vector<opt::Instruction*> available_composites =
+            FindAvailableInstructions(
+                function, block, instruction_iterator,
+                [instruction_descriptor](
+                    opt::IRContext* ir_context,
+                    opt::Instruction* instruction) -> bool {
+                  auto instruction_type = ir_context->get_type_mgr()->GetType(
+                      instruction->type_id());
```
Vasniktel, line 68: Be sure to check that the instruction has a type id and a result id. Also, add similar checks to the transformation class.

Vasniktel: You might want to refactor all these conditions into a static method of the transformation class (something like `IsInstructionSupported` or `IsCompositeInstruction` etc) if the code becomes very similar to the `IsApplicable` method. You can then call the created function from the `IsApplicable` method.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,256 @@
...
+        // |current_node_type_id| and can be inserted.
+        std::vector<opt::Instruction*> available_objects =
+            FindAvailableInstructions(
+                function, block, instruction_iterator,
+                [instruction_descriptor, current_node_type_id](
+                    opt::IRContext* ir_context,
+                    opt::Instruction* instruction) -> bool {
+                  if (instruction->type_id() != current_node_type_id) {
+                    return false;
+                  }
```
Vasniktel, line 135: It might be better to be explicit here about `instruction` having a result id and a type id. Add similar checks to the `IsApplicable` method.

source/fuzz/transformation_composite_insert.h
Vasniktel: Also, mention the created synonyms.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,256 @@
...
+            }
+          }
+        }
+
+        // The element which has been changed is synonymous to the found
+        // |available_object| itself.
+        GetTransformationContext()->GetFactManager()->AddFactDataSynonym(
+            MakeDataDescriptor(available_object_id, {}),
+            MakeDataDescriptor(new_result_id, std::move(index_to_replace)),
+            GetIRContext());
```
Vasniktel, line 208: Add synonyms only if neither `composite_id` nor `object_id` are irrelevant (`FactManager::IdIsIrrelevant`).

test/fuzz/transformation_composite_insert_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+#include "source/fuzz/instruction_descriptor.h"
+#include "test/fuzz/fuzz_test_util.h"
+
+namespace spvtools {
+namespace fuzz {
+namespace {
+
+TEST(TransformationCompositeInsertTest, BasicScenarios) {
+  // This is a simple transformation and this test handles the main cases.
```
Vasniktel, line 25: Add tests for the following cases:
1. `composite_id` is invalid (points to nonexistent instruction or instruction without a type id)
2. `object_id` is invalid (same conditions as above).
3. You can add an ASSERT_DEATH test for the case when the inserted component is a pointer if you add a corresponding assertion into `IsApplicable` method.
4. The case when it's not possible to insert `OpCompositeInsert` before the instruction descriptor.
5. Check that applied transformations create required synonyms.
6. Check the case when either `composite_id` or `object_id` is irrelevant (synonyms shouldn't be created in that case).
7. The composite has no elements.

This test case will become huge and hard to understand if you attempt to add all of the situations into it. Thus, it might be better to create separate test cases for some of them.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,256 @@
...
+             current_level < index_to_replace.size(); current_level++) {
+          index_to_skip = index_to_replace[current_level];
+          num_of_components =
+              GetNumberOfComponents(GetIRContext(), current_node_type_id);
+
+          // Store the prefix of the |index_to_replace|.
+          if (current_level != 0) {
+            current_index.push_back(index_to_replace[current_level - 1]);
+          }
+          for (uint32_t i = 0; i < num_of_components; i++) {
```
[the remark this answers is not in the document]
paulthomson: I haven't read this in detail, but there is a subtlety that might be relevant; we must mark the _components on the same level as the component that was replaced_ (except for the component that was replaced) as synonymous with the corresponding component in the original. And then mark the replaced component as synonymous with the replacement object. Thus, some descending into the composite is necessary.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+        // Take a random component of the chosen composite value. If the chosen
+        // component is itself a composite, then randomly decide whether to take
+        // its component and repeat. Use OpCompositeExtract to get the
+        // component.
+        bool reached_end_node = false;
+        uint32_t current_node_type_id = available_composite->type_id();
+        uint32_t one_selected_index;
+        uint32_t num_of_components;
+        std::vector<uint32_t> index_to_replace;
```
paulthomson, line 101: Perhaps we should rename this to something like "path_to_replaced" to make it clearer.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,256 @@
...
+        auto new_result_id = GetFuzzerContext()->GetFreshId();
+
+        // Insert an OpCompositeInsert instruction which copies
+        // |available_composite| and in the copy inserts the object
+        // of type |available_object_id| at index |index_to_replace|.
+        TransformationCompositeInsert transformation =
+            TransformationCompositeInsert(instruction_descriptor, new_result_id,
+                                          available_composite->result_id(),
+                                          available_object_id,
+                                          std::move(index_to_replace));
```
paulthomson, line 164: You have moved `index_to_replace` but then you use it below. I don't think you want to move it here. I guess copying is fine.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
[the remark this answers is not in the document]
Vasniktel: Indeed. I've missed this point. Consider the following as an example:
```glsl
struct S {
  struct{
    vec2 a;
  } b;
  struct {
    struct {
      vec4 c;
      int d;
    } e;
    vec4 f;
  } g;
};
```

Imagine we have two objects `A` and `B` of type `S`. If we decide to replace `A.g.e.c` with some value `v`, we would only need to add the following synonyms:
- `A.b` ~ `B.b`
- `A.g.e.d` ~ `B.g.e.d`
- `A.g.f` ~ `B.g.f`
- `A.g.e.c` ~ `v`

That is, we only recur in the component, whose element we are replacing.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,256 @@
...
+                      FindInstruction(instruction_descriptor, ir_context),
+                      instruction->result_id());
+                });
+
+        // If there are no objects of the specific type available, create a zero
+        // constant of this type.
+        uint32_t available_object_id;
+        if (available_objects.empty()) {
+          available_object_id =
+              FindOrCreateZeroConstant(current_node_type_id, false);
```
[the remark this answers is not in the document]
antnkarp: Thanks for your explanation. I chose solution 2.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,256 @@
...
+        bool reached_end_node = false;
+        uint32_t current_node_type_id = available_composite->type_id();
+        uint32_t one_selected_index;
+        uint32_t num_of_components;
+        std::vector<uint32_t> index_to_replace;
+        while (!reached_end_node) {
+          num_of_components =
+              GetNumberOfComponents(GetIRContext(), current_node_type_id);
+          one_selected_index = GetFuzzerContext()->GetRandomIndexForAccessChain(
+              num_of_components);
```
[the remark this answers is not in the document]
antnkarp: I was thinking about solving this problem by checking in lambda if length is ==0 and return false if so. I don't think that according to spec you can perform an OpCompositeInsert on an empty composite, because at least one index literal must be defined.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,256 @@
...
+          }
+          for (uint32_t i = 0; i < num_of_components; i++) {
+            if (i == index_to_skip) {
+              continue;
+            } else {
+              current_index.push_back(i);
+              GetTransformationContext()->GetFactManager()->AddFactDataSynonym(
+                  MakeDataDescriptor(new_result_id, std::move(current_index)),
+                  MakeDataDescriptor(available_composite->result_id(),
+                                     std::move(current_index)),
```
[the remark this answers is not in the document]
antnkarp: I have fixed it by constructing a vector in place as a r-value using copy constructor. Is it correct?

source/fuzz/transformation.cpp
```diff
@@ -93,6 +94,9 @@ std::unique_ptr<Transformation> Transformation::FromMessage(
   switch (message.transformation_case()) {
     case protobufs::Transformation::TransformationCase::kAccessChain:
       return MakeUnique<TransformationAccessChain>(message.access_chain());
+    case protobufs::Transformation::TransformationCase::kCompositeInsert:
+      return MakeUnique<TransformationCompositeInsert>(
+          message.composite_insert());
```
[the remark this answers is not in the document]
antnkarp: Thanks, I missed that.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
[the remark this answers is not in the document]
Vasniktel: Sorry for the confusion, I meant that in the following case:
```glsl
struct S {
  struct F {
  } a;
};
```
when you reach `a`, you can insert a component of type `F` into `S.a`. That being said, you can do:
```
%f = OpConstantComposite %S_F
%f1 = OpCopyObject %S_F %f
%s = OpConstantComposite %S %f
%s1 = OpCompositeInsert %S %f1 %s 0
```

Although, returning false is fine with me too.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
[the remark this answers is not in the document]
Vasniktel: Can you provide an example?

source/fuzz/fuzzer_pass_add_composite_inserts.h
```diff
@@ -0,0 +1,46 @@
...
+
+  ~FuzzerPassAddCompositeInserts();
+  void Apply() override;
+
+  // Gets number of components in composites (array, matrix, vector or struct).
+  static uint32_t GetNumberOfComponents(opt::IRContext* ir_context,
+                                        uint32_t composite_id);
+
+  // Checks if any component of a composite has type OpTypeRuntimeArray.
+  static bool ContainsRuntimeArray(const opt::analysis::Type* type);
```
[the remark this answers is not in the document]
antnkarp: I would like to use GetNumberOfComponents() in the transformation. Is it ok if I leave it as public?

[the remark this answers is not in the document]
Vasniktel: Yes, I think so.

test/fuzz/transformation_composite_insert_test.cpp
antnkarp: Could you give an example when it is not possible to insert OpCompositeInsert before the instruction descriptor?

Vasniktel: For example, when an instruction descriptor is an OpBranchConditional with OpSelectionMerge above it. Or when instruction descriptor is an OpPhi.

Vasniktel: The reason why we need to test these cases is that, even though you are applying your transformation correctly in the fuzzer pass, we want to be sure that it will never apply incorrectly in the shrinker.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,209 @@
...
+    return false;
+  }
+
+  // We ignore pointers for now. Consider adding support for pointer types.
+  auto object_instruction_type =
+      ir_context->get_type_mgr()->GetType(object_instruction->type_id());
+  // No unused variables in release mode (to keep compilers happy).
+  (void)object_instruction_type;
+  assert(object_instruction_type->kind() != opt::analysis::Type::kPointer &&
+         "The object to be inserted cannot be a pointer.");
```
paulthomson, line 77: I believe we currently like to imagine that transformations could be applied to ANY module, and they should never crash. This means the `IsApplicable` method should not crash, but may return false (and so the `Apply` method will never be called). Theoretically, if `Apply` was somehow called even though `IsApplicable` would return false, we are allowed to crash (assert false). 

Thus, the instruction found from `message_.object_id` might end up not having a type id. So you should do `if (!object_instruction->type_id()) return false;`.

Similarly, I think the assert should not be an assert, and we should instead just return false if `object_instruction_type` happens to be a pointer type. Also, I think we prefer to use `->AsPointer()` (or `->AsInteger()`, `->AsFloat()`, etc.) to check the type; we will get nullptr if the type is not a pointer (integer, float, etc.).

paulthomson: Actually, I guess you technically don't need to check that `object_instruction->type_id()` is non-zero because it is compared with `component_to_be_replaced_type_id`, which is non-zero. But in general, it does not hurt to check this.

Vasniktel: Indeed. I agree. It was me who requested that we add an assertion here (sorry, @antnkarp).

But now that I think about it, it might indeed be better to use an `if` statement here. I guess the rule of thumb here is:
> if some condition is valid but not supported (like the pointer case above) - use `if`s; otherwise (if the condition is not possible in the current context) - use asserts. 

The last part might seem like a waste of time but it helps the reader to understand writer's assumptions about the code.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+  // We have modified the module so most analyzes are now invalid.
+  ir_context->InvalidateAnalysesExceptFor(opt::IRContext::kAnalysisNone);
+
+  // Add facts about synonyms. Every element which hasn't been changed in
+  // the copy is synonymous to the corresponding element in the original
+  // |available_composite|. For every index that is a prefix of
+  // |index_to_replace| the components different from the one that
+  // contains the inserted object are synonymous with corresponding
+  // elements in the |available_composite|.
```
Vasniktel, line 130: Don't forget to adjust this comment once you've moved it from the fuzzer pass. For example, `available_composite` is not available :) in the current scope.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+    for (uint32_t i = 0; i < num_of_components; i++) {
+      if (i == index_to_skip) {
+        continue;
+      } else {
+        current_index.push_back(i);
+        // TODO: Google C++ guide restricts the use of r-value references.
+        //       https://google.github.io/styleguide/cppguide.html#Rvalue_references
+        //       Consider changing the signature of MakeDataDescriptor()
+        //       https://github.com/KhronosGroup/SPIRV-Tools/issues/3659
```
Vasniktel, line 161: Use this style for TODOs
```
TODO(issue link):
 description
```

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,209 @@
...
+  message_.set_fresh_id(fresh_id);
+  message_.set_composite_id(composite_id);
+  message_.set_object_id(object_id);
+  for (auto an_index : index) {
+    message_.add_index(an_index);
+  }
+}
+
+bool TransformationCompositeInsert::IsApplicable(
+    opt::IRContext* ir_context, const TransformationContext& /*unused*/) const {
```
Vasniktel, line 43: Add checks for cases when either `composite_id` or `object_id` is not available before `instruction_to_insert_before`. Check the corner case when `instruction_to_insert_before` is equal to either `composite_id` or `object_id`.

Don't forget to add tests about this.

source/fuzz/transformation_composite_insert.cpp
Vasniktel, line 77: I've noticed that you've marked the comment https://github.com/KhronosGroup/SPIRV-Tools/pull/3606/files#r466504064 as resolved but the issue is still present.

I'm sorry I've requested to add an assertion here initially :) It's sometimes easy to lose track of things.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,209 @@
...
+  // |message_.index| must refer to a valid index. The type id of the object at
+  // |message_.object_id| and the type id of the component of the composite at
+  // index |message_.index| must be the same.
+  auto component_to_be_replaced_type_id = fuzzerutil::WalkCompositeTypeIndices(
+      ir_context, composite->type_id(), message_.index());
+  if (component_to_be_replaced_type_id == 0) {
+    return false;
+  }
+  auto object_instruction =
+      ir_context->get_def_use_mgr()->GetDef(message_.object_id());
```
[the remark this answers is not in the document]
Vasniktel: I've requested this change since it might cause a segfault when you try to use a type manager below (when you are checking that the type is not a pointer).

source/fuzz/transformation_composite_insert.cpp
Vasniktel: You can do all of this with `fuzzerutil::IdIsAvailableBeforeInstruction`. Simply move those two checks from the fuzzer pass into the `IsApplicable` method.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,228 @@
...
+        bool reached_end_node = false;
+        uint32_t current_node_type_id = available_composite->type_id();
+        uint32_t one_selected_index;
+        uint32_t num_of_components;
+        std::vector<uint32_t> path_to_replaced;
+        while (!reached_end_node) {
+          num_of_components =
+              GetNumberOfComponents(GetIRContext(), current_node_type_id);
+          one_selected_index =
+              GetFuzzerContext()->GetRandomIndexForComposite(num_of_components);
```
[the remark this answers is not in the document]
antnkarp: Thanks for pointing this out.

source/fuzz/fuzzer_context.h
```diff
@@ -287,6 +293,9 @@ class FuzzerContext {
 
     return components;
   }
+  uint32_t GetRandomIndexForComposite(uint32_t number_of_components) {
+    return random_generator_->RandomUint32(number_of_components);
+  }
```
Vasniktel, line 298: Make sure these are sorted.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+          // If the component is not a composite or if the composite is empty or
+          // if we decide not to go deeper, then end the iteration.
+          if (!fuzzerutil::IsCompositeType(
+                  GetIRContext()->get_type_mgr()->GetType(
+                      current_node_type_id))) {
+            reached_end_node = true;
+          } else if (GetNumberOfComponents(GetIRContext(),
+                                           current_node_type_id) == 0) {
+            reached_end_node = true;
```
Vasniktel, line 123: If `GetNumberOfComponents(GetIRContext(), current_node_type_id) == 0`, then `GetFuzzerContext()->GetRandomIndexForComposite` has already fired an assertion above.

Vasniktel: Also, I think it is better to get rid of `reached_end_node` and use `while(true)` and `break` here.

Vasniktel: Or maybe `do {...} while (...);`. Whichever you like most.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,228 @@
...
+        uint32_t one_selected_index;
+        uint32_t num_of_components;
+        std::vector<uint32_t> path_to_replaced;
+        while (true) {
+          num_of_components =
+              GetNumberOfComponents(GetIRContext(), current_node_type_id);
+
+          // If the composite is empty, then end the iteration.
+          if (GetNumberOfComponents(GetIRContext(), current_node_type_id) ==
+              0) {
```
Vasniktel, line 109: You can use `num_of_components` here.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,225 @@
...
+  }
+  // The instruction having the id of |message_.object_id| must be valid.
+  auto object_instruction =
+      ir_context->get_def_use_mgr()->GetDef(message_.object_id());
+  if (object_instruction == nullptr) {
+    return false;
+  }
+  if (object_instruction->type_id() == 0) {
+    return false;
+  }
```
Vasniktel, line 71: Minor issue: you can merge these two.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+  // We ignore pointers for now. Consider adding support for pointer types.
+  auto object_instruction_type =
+      ir_context->get_type_mgr()->GetType(object_instruction->type_id());
+  // No unused variables in release mode (to keep compilers happy).
+  (void)object_instruction_type;
```
Vasniktel, line 77: This is no longer needed.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,225 @@
...
+         "|message_.fresh_id| must be fresh");
+
+  std::vector<uint32_t> index =
+      fuzzerutil::RepeatedFieldToVector(message_.index());
+  opt::Instruction::OperandList in_operands;
+  in_operands.push_back({SPV_OPERAND_TYPE_ID, {message_.object_id()}});
+  in_operands.push_back({SPV_OPERAND_TYPE_ID, {message_.composite_id()}});
+  for (uint32_t i = 0; i < index.size(); i++) {
+    in_operands.push_back({SPV_OPERAND_TYPE_LITERAL_INTEGER, {index[i]}});
+  }
```
Vasniktel, line 127: Consider using a range-for loop here (`for (auto i : index)`).

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+  // If |composite_id| or |object_id| is irrelevant then don't add any synonyms.
+  if (transformation_context->GetFactManager()->IdIsIrrelevant(
+          message_.composite_id()) ||
+      transformation_context->GetFactManager()->IdIsIrrelevant(
+          message_.object_id())) {
+    return;
+  }
+  uint32_t current_node_type_id = composite_type_id;
+  uint32_t index_to_skip, num_of_components;
```
Vasniktel, line 156: It is better to reduce the scope in which the variable can be used. In this case, you can move these two into the loop.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,225 @@
...
+  // elements in the original composite.
+
+  // If |composite_id| or |object_id| is irrelevant then don't add any synonyms.
+  if (transformation_context->GetFactManager()->IdIsIrrelevant(
+          message_.composite_id()) ||
+      transformation_context->GetFactManager()->IdIsIrrelevant(
+          message_.object_id())) {
+    return;
+  }
+  uint32_t current_node_type_id = composite_type_id;
```
Vasniktel, line 155: I don't believe this variable is not changed in the loop.

Vasniktel, line 155: I don't believe this variable is changed in the loop.

test/fuzz/transformation_composite_insert_test.cpp
```diff
@@ -0,0 +1,815 @@
...
+                                         MakeDataDescriptor(51, {0, 0})));
+  ASSERT_FALSE(fact_manager.IsSynonymous(MakeDataDescriptor(50, {0, 1, 0}),
+                                         MakeDataDescriptor(51, {0, 1, 0})));
+  ASSERT_FALSE(fact_manager.IsSynonymous(MakeDataDescriptor(51, {0, 1, 1}),
+                                         MakeDataDescriptor(11, {})));
+}
+
+TEST(TransformationCompositeInsertTest, ApplicableCreatedSynonyms) {
+  // This test handles cases where neither |composite| nor |object| is
+  // irrelevant. The transformation should create synonyms.
```
Vasniktel, line 473: Add some cases when you are inserting into a subcomponent that has a different number of elements than its parent. This should reveal the bug that can be caused by the variable not being changed in the loop (see my comment in the transformation class).

Vasniktel:
I've just noticed that you haven't added your fuzzer pass to `fuzzer.cpp` file. Without it, your transformation will never be applied.

This mistake is easy to make so be sure that your other merged transformations don't have it.

source/fuzz/transformation_composite_insert.cpp
antnkarp: I've tried it and it no longer preserved the increasing order of index i (0->1->2 ...). I have left it as it is.

source/fuzz/transformation_composite_insert.cpp
antnkarp: Thanks for pointing it out, I have fixed it.

source/fuzz/transformation_composite_insert.cpp
Vasniktel: I can be a little bit confusing but you would have to use `i` instead of `index[i]` once you've replaced the loop. Range-for loop iterates through the values in the vector, not indices into it.

antnkarp: Yes, you're right, I missed that. Thanks.

Vasniktel: :+1:

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,227 @@
...
+            available_composites[GetFuzzerContext()->RandomIndex(
+                available_composites)];
+
+        // Take a random component of the chosen composite value. If the chosen
+        // component is itself a composite, then randomly decide whether to take
+        // its component and repeat. Use OpCompositeExtract to get the
+        // component.
+        uint32_t current_node_type_id = available_composite->type_id();
+        uint32_t one_selected_index;
+        uint32_t num_of_components;
```
Vasniktel, line 101: I think these can be moved into the loop.

test/fuzz/transformation_composite_insert_test.cpp
```diff
@@ -0,0 +1,906 @@
...
+  fresh_id = 50;
+  composite_id = 32;
+  object_id = 9;
+  index = {2, 1};
+  auto transformation_good_1 = TransformationCompositeInsert(
+      instruction_to_insert_before, fresh_id, composite_id, object_id, index);
+  ASSERT_TRUE(transformation_good_1.IsApplicable(context.get(),
+                                                 transformation_context));
+  transformation_good_1.Apply(context.get(), &transformation_context);
+  ASSERT_TRUE(IsValid(env, context.get()));
```
Vasniktel, line 901: Please, add assertions on the synonyms, created by the transformation.

antnkarp: I have added them: see line 600.

Vasniktel: You should then merge this test case with the ones that test synonymous facts. This is to make sure that the fix for the variable not being changed in the loop works.

antnkarp: Done, thanks.

test/fuzz/transformation_composite_insert_test.cpp
```diff
@@ -0,0 +1,826 @@
...
+  // Bad: |fresh_id| is not fresh.
+  instruction_to_insert_before = MakeInstructionDescriptor(29, SpvOpStore, 0);
+  fresh_id = 20;
+  composite_id = 29;
+  object_id = 11;
+  index = {1, 0, 0};
+  auto transformation_bad_1 = TransformationCompositeInsert(
+      instruction_to_insert_before, fresh_id, composite_id, object_id, index);
+  ASSERT_FALSE(
+      transformation_bad_1.IsApplicable(context.get(), transformation_context));
```
paulthomson, line 115: Google C++ style: Place a function's variables in the narrowest scope possible, and initialize variables in the declaration.

Ally would probably say to just initialize stuff inline:

```
TransformationCompositeInsert(MakeInstructionDescriptor(29, SpvOpStore, 0), 20, ...)
```

If you want to make it easier to read, you can add comments (but this is not required):

```
TransformationCompositeInsert(
   MakeInstructionDescriptor(29, SpvOpStore, 0), ...), // instruction to insert before
   20,                                                 // fresh id
...
```

Instead of adding comments like above, you can just add more information to the original comment, so it is still fairly clear what is happening:

```
// Bad: |fresh_id| %20 is not fresh.
```

Or, if you would rather use variables, you should do:

```
  // Bad: |fresh_id| is not fresh.
  {
    protobufs::InstructionDescriptor instruction_to_insert_before = MakeInstructionDescriptor(29, SpvOpStore, 0);
    uint32_t fresh_id = 20;
    uint32_t composite_id = 29;
    uint32_t object_id = 11;
    std::vector<uint32_t> index = {1, 0, 0};
    auto transformation_bad_1 = TransformationCompositeInsert(
      instruction_to_insert_before, fresh_id, composite_id, object_id, index);
  }
```

But I don't think I would recommend using variables, as most of the existing tests don't do this.

A big problem with your current approach is: you may have missed assigning to one of the existing variables, but then still use the variable when you construct the transformation. Nobody will notice this during code review, so we prefer to avoid it by instead using one of the above approaches :-)

test/fuzz/transformation_composite_insert_test.cpp
```diff
@@ -0,0 +1,826 @@
...
+               OpFunctionEnd
+  )";
+  ASSERT_TRUE(IsEqual(env, after_transformations, context.get()));
+}
+
+TEST(TransformationCompositeInsertTest, IrrelevantNoSynonyms) {
+  // This test handles cases where either |composite| or |object| is irrelevant.
+  // The transformation shouldn't create any synonyms.
+  // The member composite has a different number of elements than the parent
+  // composite.
```
paulthomson, line 343: I guess this test will need some updates, assuming that my previous comment is correct (about still adding most synonyms when the composite is not irrelevant).

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,228 @@
...
+      return composite_type->GetSingleWordInOperand(1);
+    case SpvOpTypeStruct:
+      return fuzzerutil::GetNumberOfStructMembers(*composite_type);
+    default:
+      assert(false &&
+             "|composite_type_id| must be a result id of a composite "
+             "type.");
+      return 0;
+  }
+}
```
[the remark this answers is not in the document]
antnkarp: Thanks for pointing this out. Turns out the function with this functionality was already implemented.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+  // |message_.composite_id| must refer to an existing composite value.
+  auto composite =
+      ir_context->get_def_use_mgr()->GetDef(message_.composite_id());
+
+  if (!IsCompositeInstructionSupported(ir_context, composite)) {
+    return false;
+  }
+
+  // |message_.index| must refer to a correct index.
```
paulthomson, line 57: I think you mean: The indices in |message_.index| must be suitable for indexing into |composite->type_id()|.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,226 @@
...
+      fuzzerutil::GetTypeId(ir_context, message_.composite_id());
+
+  FindInstruction(message_.instruction_to_insert_before(), ir_context)
+      ->InsertBefore(MakeUnique<opt::Instruction>(
+          ir_context, SpvOpCompositeInsert, composite_type_id,
+          message_.fresh_id(), std::move(in_operands)));
+
+  fuzzerutil::UpdateModuleIdBound(ir_context, message_.fresh_id());
+
+  // Make sure our changes are analyzed.
```
paulthomson, line 133: I think it is best to change it back to: We have modified the module so most analyzes are now invalid.

source/fuzz/transformation_composite_insert.cpp
```diff
@@ -0,0 +1,226 @@
...
+        *current_node_type_inst, ir_context);
+
+    // Update the current_node_type_id.
+    current_node_type_id = fuzzerutil::WalkOneCompositeTypeIndex(
+        ir_context, current_node_type_id, index_to_skip);
+
+    for (uint32_t i = 0; i < num_of_components; i++) {
+      if (i == index_to_skip) {
+        continue;
+      } else {
```
paulthomson, line 166: Nit: remove "else" (and do this whenever it is not needed because the if statement continues or breaks or returns).

source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
```diff
@@ -37,40 +37,39 @@ FuzzerPassReplaceAddsSubsMulsWithCarryingExtended::
...
-                    ->GetChanceOfReplacingAddSubMulWithCarryingExtended())) {
-          continue;
-        }
+  GetIRContext()->module()->ForEachInst([this](opt::Instruction* instruction) {
+    // Randomly decide whether to apply the transformation.
+    if (!GetFuzzerContext()->ChoosePercentage(
+            GetFuzzerContext()
+                ->GetChanceOfReplacingAddSubMulWithCarryingExtended())) {
+      return;
+    }
```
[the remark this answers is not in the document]
antnkarp: These changes are necessary for the FuzzerPassReplaceAddsMulsWithCarryingExtended, because it turned out that all of my passes (including this one) were not included in fuzzer.cpp. Hence they were untested. I have tested this fuzzer pass and it crashed, because of broken iterator in the loop. These changes fix all the crashes that happened after I have included all of my fuzzer passes. Should I revert these changes now and open a new PR for them?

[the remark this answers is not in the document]
Vasniktel: A possible solution is to store all the necessary instructions (from a block or function etc) in the vector so that the interators are not invalidated.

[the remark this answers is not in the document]
Vasniktel: i. e. use the auxiliary space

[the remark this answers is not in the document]
paulthomson: OK we can just leave these changes this time.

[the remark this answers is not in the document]
paulthomson: Oh, but I don't think this particular change is a good one. I think Vasyl's suggestion is a good one.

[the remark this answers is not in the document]
antnkarp: I have changed the pass according to this suggestion.

source/fuzz/fuzzer_pass_add_composite_inserts.cpp
```diff
@@ -0,0 +1,207 @@
...
+        //       FindOrCreateZeroConstant cannot be called on a pointer. We
+        //       ignore pointers for now. Consider adding support for pointer
+        //       types.
+        uint32_t available_object_id;
+        if (available_objects.empty()) {
+          auto current_node_type =
+              GetIRContext()->get_type_mgr()->GetType(current_node_type_id);
+          if (current_node_type->kind() == opt::analysis::Type::kPointer) {
+            return;
+          }
```
paulthomson, line 167: You only ignore pointer types in the case where you try to create a zero constant. In the other case (an existing, available id), the id might have a pointer type. However, in IsApplicable, you return false if the id has a pointer type. Thus, this fuzzer pass could fail if it finds a struct with a member with pointer type and an existing id to insert of the required type. I suggest that you (for now) ignore any composites with pointer types (like you do for OpRuntimeArray). Or you could remove the check from IsApplicable, but it might be hard to write a test case for a struct with a pointer type (and I doubt we will see this often).

source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
```diff
@@ -39,7 +39,10 @@ FuzzerPassReplaceAddsSubsMulsWithCarryingExtended::
 void FuzzerPassReplaceAddsSubsMulsWithCarryingExtended::Apply() {
   for (auto& function : *GetIRContext()->module()) {
     for (auto& block : function) {
-      for (auto& instruction : block) {
+      // Copy instructions since the transformation invalidates iterators.
+      std::vector<opt::Instruction> instr_from_block =
+          std::vector<opt::Instruction>(block.begin(), block.end());
+      for (auto& instruction : instr_from_block) {
```
paulthomson, line 45: Similar to other fuzzer passes, I suggest that you instead store chosen OpIAdd/Sub/Mul instructions to a vector. Then add a second loop over the vector that gets the operand type id, finds or creates the struct type, and applies the transformation.

source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
```diff
@@ -73,4 +76,4 @@ void FuzzerPassReplaceAddsSubsMulsWithCarryingExtended::Apply() {
   }
 }
 }  // namespace fuzz
-}  // namespace spvtools
+}  // namespace spvtools
```
paulthomson, line 79: Make sure the file ends in a new line. There is a CLion option to add this on file save.

source/fuzz/fuzzer_pass_replace_copy_objects_with_stores_loads.cpp
```diff
@@ -66,7 +66,15 @@ void FuzzerPassReplaceCopyObjectsWithStoresLoads::Apply() {
...
-    // Find or create a constant to initialize the variable from.
+    // Find or create a constant to initialize the variable from. The type of
+    // |instruction| must be scalar or composite.
+    auto instruction_type =
+        GetIRContext()->get_type_mgr()->GetType(instruction->type_id());
+    if (!instruction_type->AsBool() && !instruction_type->AsInteger() &&
+        !instruction_type->AsFloat() &&
+        !fuzzerutil::IsCompositeType(instruction_type)) {
+      return;
+    }
```
paulthomson, line 77: Is this essentially saying that the type must not be a pointer type? If so, please just check that instead.

antnkarp: It is to assert that we can execute FindOrCreateZeroConstant on line 79 so I think that we need to exclude more cases than just pointer types.

source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.cpp
antnkarp: @paulthomson Could you point me to a specific fuzzer pass that uses this approach?

paulthomson: I think an example is FuzzerPassAddDeadBlocks

source/fuzz/fuzzer_pass_replace_copy_objects_with_stores_loads.cpp
paulthomson: Ah yes you are right.

PR #3610, opened 2020-07-29
spirv-opt: Improve the code of the Instruction class

andreperezmaselco:
This PR makes some improvements to the source code of the Instruction class,
choosing better names for variables and avoiding unnecessary function calls.

### DIFF
--- a/source/opt/instruction.cpp
+++ b/source/opt/instruction.cpp
@@ -171,20 +171,17 @@ uint32_t Instruction::NumInOperandWords() const {
 }
 
 bool Instruction::HasBranchWeights() const {
-  if (opcode_ == SpvOpBranchConditional &&
-      NumOperands() == kOpBranchConditionalWithWeightsNumOperands) {
-    return true;
-  }
-
-  return false;
+  return opcode_ == SpvOpBranchConditional &&
+         NumOperands() == kOpBranchConditionalWithWeightsNumOperands;
 }
 
 void Instruction::ToBinaryWithoutAttachedDebugInsts(
     std::vector<uint32_t>* binary) const {
   const uint32_t num_words = 1 + NumOperandWords();
   binary->push_back((num_words << 16) | static_cast<uint16_t>(opcode_));
-  for (const auto& operand : operands_)
+  for (const auto& operand : operands_) {
     binary->insert(binary->end(), operand.words.begin(), operand.words.end());
+  }
 }
 
 void Instruction::ReplaceOperands(const OperandList& new_operands) {
@@ -283,8 +280,7 @@ bool Instruction::IsVulkanStorageImage() const {
 
   // Check if the image is sampled.  If we do not know for sure that it is,
   // then assume it is a storage image.
-  auto s = base_type->GetSingleWordInOperand(kTypeImageSampledIndex);
-  return s != 1;
+  return base_type->GetSingleWordInOperand(kTypeImageSampledIndex) != 1;
 }
 
 bool Instruction::IsVulkanSampledImage() const {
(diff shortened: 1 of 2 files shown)


source/opt/instruction.h
```diff
@@ -821,16 +819,16 @@ inline bool Instruction::WhileEachInOperand(
 
 inline void Instruction::ForEachInOperand(
     const std::function<void(uint32_t*)>& f) {
-  WhileEachInOperand([&f](uint32_t* op) {
-    f(op);
+  WhileEachInOperand([&f](uint32_t* operand_id) {
+    f(operand_id);
```
afd, line 823: The argument is not necessarily an id.  It could be a storage class, a literal, or various other things.

source/opt/instruction.h
```diff
@@ -821,16 +819,16 @@ inline bool Instruction::WhileEachInOperand(
...
     return true;
   });
 }
 
 inline void Instruction::ForEachInOperand(
     const std::function<void(const uint32_t*)>& f) const {
-  WhileEachInOperand([&f](const uint32_t* op) {
-    f(op);
+  WhileEachInOperand([&f](const uint32_t* operand_id) {
```
s-perron, line 830: Same here.  This is not always an id.

source/opt/instruction.cpp
```diff
@@ -171,20 +171,17 @@ uint32_t Instruction::NumInOperandWords() const {
 }
 
 bool Instruction::HasBranchWeights() const {
-  if (opcode_ == SpvOpBranchConditional &&
-      NumOperands() == kOpBranchConditionalWithWeightsNumOperands) {
-    return true;
-  }
-
-  return false;
```
s-perron, line 179: We might sometimes end up with code like this because it is easier to add a break point that breaks anytime this function returns true.  You break on "return true".  With your code, we would have to add a conditional breakpoint, which is much more tedious to add.

The change looks cleaner, so this is another example of something that we treat as a personal preference.

andreperezmaselco:
@s-perron, I was taking a look at the `Instruction` class and I thought some changes would improve the quality of the code.
Do you think I should fix the issues in the comments or I should close this PR?

s-perron:
Fix some of the issue.  Undo those the changes that I specifically mentioned.

source/opt/instruction.cpp
andreperezmaselco: Done.

source/opt/instruction.cpp
```diff
@@ -318,12 +314,11 @@ bool Instruction::IsVulkanSampledImage() const {
...
   // Check if the image is sampled.  If we know for sure that it is,
   // then return true.
-  auto s = base_type->GetSingleWordInOperand(kTypeImageSampledIndex);
-  return s == 1;
+  return base_type->GetSingleWordInOperand(kTypeImageSampledIndex) == 1;
 }
 
 bool Instruction::IsVulkanStorageTexelBuffer() const {
-  if (opcode() != SpvOpTypePointer) {
+  if (opcode_ != SpvOpTypePointer) {
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/opt/instruction.cpp
```diff
@@ -502,16 +497,17 @@ uint32_t Instruction::GetTypeComponent(uint32_t element) const {
   return subtype;
 }
 
-Instruction* Instruction::InsertBefore(std::unique_ptr<Instruction>&& i) {
-  i.get()->InsertBefore(this);
-  return i.release();
+Instruction* Instruction::InsertBefore(
+    std::unique_ptr<Instruction>&& instruction) {
```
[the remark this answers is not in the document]
andreperezmaselco: Done.

source/opt/instruction.h
andreperezmaselco: Done.

source/opt/instruction.h
andreperezmaselco: Done.

andreperezmaselco:
@s-perron, I made the changes. Could you review them?

andreperezmaselco:
@s-perron, I guess it is ready to be merged.

PR #3618, opened 2020-07-29
Validator support for non-semantic clspv reflection

alan-baker:
Add validation for clspv's embedded reflection instructions.

* Remove SPIRV-Tools local duplicated copies of extended instruction set grammars
* Change the build to use extended instruction grammars from SPIRV-Headers instead
* Add a build for clspv reflection extended instruction grammar


source/CMakeLists.txt
```diff
@@ -19,8 +19,8 @@ set(LANG_HEADER_PROCESSING_SCRIPT "${spirv-tools_SOURCE_DIR}/utils/generate_lang
 
 # For now, assume the DebugInfo grammar file is in the current directory.
 # It might migrate to SPIRV-Headers.
-set(DEBUGINFO_GRAMMAR_JSON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/extinst.debuginfo.grammar.json")
-set(CLDEBUGINFO100_GRAMMAR_JSON_FILE "${CMAKE_CURRENT_SOURCE_DIR}/extinst.opencl.debuginfo.100.grammar.json")
+set(DEBUGINFO_GRAMMAR_JSON_FILE "${SPIRV_HEADER_INCLUDE_DIR}/spirv/unified1/extinst.debuginfo.grammar.json")
```
dneto0, line 22: Also need corresponding changes in the Android.mk file

source/val/validate_extensions.cpp
```diff
@@ -180,6 +181,443 @@ spv_result_t ValidateOperandDebugType(
...
+           << "DescriptorSet must be a 32-bit unsigned integer OpConstant";
+  }
+
+  if (!IsUint32Constant(_, inst->GetOperandAs<uint32_t>(7))) {
+    return _.diag(SPV_ERROR_INVALID_ID, inst)
+           << "Binding must be a 32-bit unsigned integer OpConstant";
+  }
+
+  if (num_operands == 9) {
+    if (auto error = ValidateArgInfo(_, inst, 8)) {
```
dneto0, line 349: You're passing in an argument index, but the function uses that third parameter as a SPIR-V ID.

The same occurs for the other uses of ValidateArgInfo

source/CMakeLists.txt
alan-baker: And BUILD.bazel and BUILD.gn... so many build systems.

source/val/validate_extensions.cpp
alan-baker: I think this actually a bad parameter name. The first line of ValidateArgInfo is
```
  auto info = _.FindDef(inst->GetOperandAs<uint32_t>(id));
```

So it should be an index. I'll update it to be clearer.

alan-baker:
@dnovillo please take a look at the BUILD.bazel changes.

@zoddicus please take a look at the BUILD.gn changes.

source/val/validate_extensions.cpp
dneto0: Oh, I was misled and misread the code. This works.

BUILD.gn
```diff
@@ -110,8 +110,8 @@ template("spvtools_glsl_tables") {
     core_json_file =
         "${spirv_headers}/include/spirv/$version/spirv.core.grammar.json"
     glsl_json_file = "${spirv_headers}/include/spirv/${version}/extinst.glsl.std.450.grammar.json"
-    debuginfo_insts_file = "source/extinst.debuginfo.grammar.json"
-    cldebuginfo100_insts_file = "source/extinst.opencl.debuginfo.100.grammar.json"
+    debuginfo_insts_file = "${spirv_headers/include/spirv/unified1/extinst.debuginfo.grammar.json"
```
zoddicus, line 113: Missing closing }s

PR #3619, opened 2020-07-29
Favour 'integrity' over 'coherence' as a replacement for 'sanity'.

afd:


dneto0 (approved):
Thanks! this is an improvement.
Thanks for the thoughtful followup to my previous (late) feedback.

PR #3627, opened 2020-07-31
spirv-fuzz: Support pointer types in FuzzerPassAddParameters

### LINKED ISSUE #3403: spirv-fuzz: Add support for new types to FindOrCreateZeroConstant
`FindOrCreateZeroConstant` function only supports a limited number of type instructions. However, we are using this function extensively in many different transformations. Thus, it would be beneficial to add support for some additional types (e.g. `TransformationAddParameters` can't add `OpTypePointer` parameters to functions because `FindOrCreateZeroConstant` function does not support them).

antnkarp:
For FuzzerPassAddParameters, adds pointer types (that have the storage
class Function or Private) to the pool of available types for new
parameters. If there are no variables of the chosen pointer type, it
invokes TransformationAddLocalVariable / TransformationAddGlobalVariable
to add one.

Part of #3403


source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -38,8 +40,23 @@ void FuzzerPassAddParameters::Apply() {
...
+                                                     type_inst->result_id());
+      // Check if the pointee type is supported.
+      auto pointee_type = GetIRContext()->get_type_mgr()->GetType(
+          fuzzerutil::GetPointeeTypeIdFromPointerType(GetIRContext(),
+                                                      type_inst->result_id()));
+      if (TransformationAddParameter::IsParameterTypeSupported(*pointee_type) &&
+          (storage_class == SpvStorageClassPrivate ||
+           storage_class == SpvStorageClassFunction)) {
+        type_candidates.push_back(type_inst->result_id());
+      }
```
Vasniktel, line 59: I think we can move this check into `TransformationAddParameter::IsParameterTypeSupported`.

Vasniktel: We could also add support for `Workgroup` storage class. It's the same as `Private` storage class but the initializer is `0`. See `TransformationAddGlobalVariable` docs for more details.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +87,62 @@ void FuzzerPassAddParameters::Apply() {
...
+            [this, &available_variable_ids,
+             current_type_id](opt::Instruction* instruction) {
+              if (instruction->opcode() != SpvOpVariable) {
+                return;
+              }
+              if (instruction->type_id() != current_type_id) {
+                return;
+              }
+              available_variable_ids.push_back(instruction->result_id());
+            });
```
Vasniktel, line 116: This part needs some refactoring. See the general review comment for more details.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +87,62 @@ void FuzzerPassAddParameters::Apply() {
...
+        uint32_t initializer_id =
+            FindOrCreateZeroConstant(current_pointee_type_id, true);
```
Vasniktel, line 118: I'm not sure whether this is a good or a bad practice to mark a variable initializer as irrelevant. @paulthomson, what do you think? In case we decide to make it irrelevant, it would be good to add a comment similar to
> We mark the constant as irrelevant so that we can replace it with a more interesting value later.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +87,62 @@ void FuzzerPassAddParameters::Apply() {
...
+        // irrelevant.
+        if (available_variable_ids.empty()) {
+          if (storage_class == SpvStorageClassPrivate) {
+            ApplyTransformation(TransformationAddGlobalVariable(
+                GetFuzzerContext()->GetFreshId(), current_type_id,
+                SpvStorageClassPrivate, initializer_id, true));
+          } else if (storage_class == SpvStorageClassFunction) {
+            ApplyTransformation(TransformationAddLocalVariable(
+                GetFuzzerContext()->GetFreshId(), current_type_id,
+                function.result_id(), initializer_id, true));
```
Vasniktel, line 130: This will create a local `OpVariable` instruction in the `function`. This is somewhat tricky.

The initial idea is that the created variable will be used as an operand to `OpFunctionCall` instructions. These instructions can belong to different functions - we'll call them callers - (since we can call any function from any other as long as there are no recursion - SPIR-V doesn't support that). Thus, this piece of code should create a local variable in every caller of `function.result_id()`. This is somewhat tricky because some callers might already have a suitable local variable.

I propose to refactor this fuzzer pass as follows:
1. Check the storage class of the variable. If it's `Private` or `Workgroup`, either find a global variable (`IRContext` must have a method that returns only global instructions) or create a new one (as you do above). Otherwise...
2. Iterate over all callers of `function.result_id()` (see `fuzzerutil::GetCallers`) and for each one either find a suitable local variable or create one.

I would also suggest to change the protobuf file to pass a repeated field of `UInt32Pair` (see `fuzzerutil::Repeated*ToMap` and `fuzzerutil::MapToRepeated*`). This field would represent a map from the result id of the caller to the result id of the local variable in that caller.

You would also need to adjust `IsApplicable` method in the transformation class to check that each caller has a local variable (if the added parameter is indeed a pointer with `Function` storage class). Hope this makes sense.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +87,62 @@ void FuzzerPassAddParameters::Apply() {
...
+          }
+        }
+        // Add a parameter with the created initializer.
+        ApplyTransformation(TransformationAddParameter(
+            function.result_id(), GetFuzzerContext()->GetFreshId(),
+            initializer_id, GetFuzzerContext()->GetFreshId()));
```
Vasniktel, line 136: `initializer_id` is used to initialize the `OpVariable` instruction. There is also an `initializer_id` argument in the transformation's constructor that is used to initialize an operand in `OpFunctionCall` instruction. In our case, the latter is the result id of the `OpVariable`. 

Sorry, it's a bit confusing but the bottom-line is that you should use the result id of the variable here instead of the `initializer_id`.

Vasniktel:
@antnkarp, you'll probably need to rebase since I merged some changes to the `TransformationAddParameters`. Let me know if you have any questions.

source/fuzz/fuzzer_pass_add_parameters.cpp
antnkarp: I was thinking about the adding a map to a protobuf file and for me it seems unneccessary. I came up with two reasons:

1. It would mean refactoring the code in cases where AddParameter doesn't add a pointer with Function storage class.

2. In the case where AddParameter is about to add the pointer with Function storage class, the information whether the transformation is applicable or not is already accessible, since we pass function_id. And as you said, we could iterate over all callers and check if there is a local variable in each calling function. Even if we pass the map in the protobuf we still have to _recreate_ it in the IsApplicable method by iterating over all callers to check whether the resulting two maps are equal. For example, even if we have checked that each key in the passed protobuf map refers to a valid local variable, we might have missed some callers. The actual check is in iterating over the callers in IsApplicable method.

Vasniktel: Yes, you are right. It is indeed might be better to simply check in the transformation that the required parameter is available and in the fuzzer pass to ensure that the parameter is available. In this case, I would've created two functions in the fuzzerutil: `MaybeGetLocalVariable` and `MaybeGetGlobalVariable` (or some equivalent of these) that return result id of the local/global variable or 0 if no such variable exists.

You could then pass the type of the newly created parameter to the transformation and check that the variable (or the constant) of the required type exists.

There is one caveat, though. Currently, we do the opposite: we pass the value of the parameter and infer parameter's type out of it. You will need to change that.

Vasniktel: The function-creation part is up to you, though.

source/fuzz/fuzzer_util.h
```diff
@@ -493,6 +493,16 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
...
 
+// Returns the id of an available local variable that has the type |type_id| and
+// is declared in the function that has the id |function_id|. If no such
+// variable is available, returns 0.
+uint32_t MaybeGetLocalVariable(opt::IRContext* ir_context, uint32_t type_id,
+                               uint32_t function_id);
+
+// Returns the id of an available global variable that has the type |type_id|.
+// If no such variable is available, returns 0.
+uint32_t MaybeGetGlobalVariable(opt::IRContext* ir_context, uint32_t type_id);
```
Vasniktel, line 504: It would be good to add two boolean parameters to these functions:
1. `pointee_value_is_irrelevant`
2. `id_is_irrelevant`

and make sure the the returned id is marked (or not marked depending on the value of parameters) with those two facts accordingly.

source/fuzz/fuzzer_util.cpp
```diff
@@ -1342,7 +1342,47 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
...
+    if (instruction.type_id() != type_id) {
+      continue;
+    }
+    // We have found the matching variable declaration.
+    return_id = instruction.result_id();
+    break;
+  }
+  return return_id;
+}
+}  // namespace fuzzerutil
```
Vasniktel, line 1386: Missing a new line.

source/fuzz/fuzzer_util.h
```diff
@@ -493,6 +493,16 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
                                                  uint32_t block_id,
                                                  SpvOp opcode);
 
+// Returns the id of an available local variable that has the type |type_id| and
+// is declared in the function that has the id |function_id|. If no such
+// variable is available, returns 0.
+uint32_t MaybeGetLocalVariable(opt::IRContext* ir_context, uint32_t type_id,
+                               uint32_t function_id);
```
Vasniktel, line 500: Mention that `type_id` is required to be a result id of `OpTypePointer` instruction with `Function` storage class.

source/fuzz/fuzzer_util.h
Vasniktel: Also, mention these parameters in the doc.

Vasniktel, line 504: Mention that `type_id` is a result id of an `OpTypePointer` instruction with `Private` or `Workgroup` storage class.

source/fuzz/fuzzer_util.cpp
```diff
@@ -1342,7 +1342,47 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
   return CanInsertOpcodeBeforeInstruction(opcode, &*it) ? &*it : nullptr;
 }
 
-}  // namespace fuzzerutil
+uint32_t MaybeGetLocalVariable(opt::IRContext* ir_context, uint32_t type_id,
+                               uint32_t function_id) {
+  uint32_t return_id = 0;
+  auto function = FindFunction(ir_context, function_id);
```
Vasniktel, line 1348: Make sure to assert that `type_id` and `function_id` have valid values. That is:
1. `function != nullptr`
2. see my comment in the header file regarding the conditions on the `type_id`.

source/fuzz/fuzzer_util.cpp
```diff
@@ -1342,7 +1342,47 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
...
+
+  // All of the local variable declarations are located in the first block
+  auto block = function->begin();
+  for (auto& instruction : *block) {
+    if (instruction.opcode() != SpvOpVariable) {
+      continue;
```
Vasniktel, line 1354: Add a comment saying that `block` might have `OpLine` and/or `OpNoLine` instructions that can be intermixed with `OpVariable`.

source/fuzz/fuzzer_util.cpp
```diff
@@ -1342,7 +1342,47 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
...
 }
 
-}  // namespace fuzzerutil
+uint32_t MaybeGetLocalVariable(opt::IRContext* ir_context, uint32_t type_id,
+                               uint32_t function_id) {
+  uint32_t return_id = 0;
+  auto function = FindFunction(ir_context, function_id);
+
+  // All of the local variable declarations are located in the first block
+  auto block = function->begin();
```
Vasniktel, line 1351: Make sure to have a reference here (or a constant reference which is better). Otherwise, we are creating a copy of an entry block.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +72,76 @@ void FuzzerPassAddParameters::Apply() {
...
-              type_candidates[GetFuzzerContext()->RandomIndex(type_candidates)],
-              true),
-          GetFuzzerContext()->GetFreshId()));
+      auto current_type_id =
+          type_candidates[GetFuzzerContext()->RandomIndex(type_candidates)];
+      auto current_type =
+          GetIRContext()->get_type_mgr()->GetType(current_type_id);
+
+      // Consider pointer types separately.
+      if (current_type->kind() == opt::analysis::Type::kPointer) {
```
Vasniktel, line 81: You could use:
```cpp
if (const auto* pointer_type = current_type->AsPointer()) {
  auto storage_class = pointer_type->storage_class();
  ...
}
```

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +72,76 @@ void FuzzerPassAddParameters::Apply() {
...
+        auto storage_class = fuzzerutil::GetStorageClassFromPointerType(
+            GetIRContext(), current_type_id);
+        uint32_t pointee_type_id =
+            (GetIRContext()->get_def_use_mgr()->GetDef(current_type_id))
```
Vasniktel, line 85: There is a function in `fuzzerutil` for retrieving pontee type id from the pointer type id.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +72,76 @@ void FuzzerPassAddParameters::Apply() {
...
+                ->GetSingleWordInOperand(1);
+        uint32_t variable_id = 0;
```
Vasniktel, line 87: Consider moving this into the switch.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +72,76 @@ void FuzzerPassAddParameters::Apply() {
...
+            // PointeeValueIsIrrelevant.
+            variable_id = fuzzerutil::MaybeGetGlobalVariable(GetIRContext(),
+                                                             current_type_id);
+            if (!variable_id || !GetTransformationContext()
+                                     ->GetFactManager()
+                                     ->PointeeValueIsIrrelevant(variable_id)) {
+              ApplyTransformation(TransformationAddGlobalVariable(
+                  GetFuzzerContext()->GetFreshId(), current_type_id,
+                  storage_class,
+                  FindOrCreateZeroConstant(pointee_type_id, true), true));
```
Vasniktel, line 125: `initializer_id` must be 0 if the storage class is `Workgroup`.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +72,76 @@ void FuzzerPassAddParameters::Apply() {
...
+            }
+          } break;
+          default:
+            break;
```
Vasniktel, line 129: It might be better to add an assert here saying that storage class is not supported. This will make it easier to identify all the places we need to adjust if we decide to add support for more storage classes.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -70,14 +72,76 @@ void FuzzerPassAddParameters::Apply() {
...
+        }
+
+        ApplyTransformation(TransformationAddParameter(
+            function.result_id(), GetFuzzerContext()->GetFreshId(),
+            current_type_id, GetFuzzerContext()->GetFreshId()));
+
+      } else  // Consider non-pointer parameters.
```
Vasniktel, line 136: You can move the comment into the scope like so:
```cpp
} else {
  // comment
```

source/fuzz/transformation_add_parameter.cpp
```diff
@@ -24,54 +24,113 @@ TransformationAddParameter::TransformationAddParameter(
...
   if (!initializer_type || !IsParameterTypeSupported(*initializer_type)) {
     return false;
   }
 
+  // Consider pointer types separately.
+  bool is_valid = true;
+  if (initializer_type->kind() == opt::analysis::Type::kPointer) {
+    uint32_t pointer_type_id = message_.initializer_id_or_pointer_type_id();
+    auto storage_class =
+        fuzzerutil::GetStorageClassFromPointerType(ir_context, pointer_type_id);
```
Vasniktel, line 77: Consider using `opt::analysis::Type::AsPointer` as I've described in the fuzzer pass.

source/fuzz/transformation_add_parameter.cpp
```diff
@@ -24,54 +24,113 @@ TransformationAddParameter::TransformationAddParameter(
...
+        for (auto* instr :
+             fuzzerutil::GetCallers(ir_context, message_.function_id())) {
+          auto block = ir_context->get_instr_block(instr);
+          auto function_id = block->GetParent()->result_id();
+          // If there is no available local variable in at least one caller,
+          // the transformation is invalid.
+          if (!fuzzerutil::MaybeGetLocalVariable(ir_context, pointer_type_id,
+                                                 function_id)) {
+            is_valid = false;
+            break;
```
Vasniktel, line 89: You can get rid of `is_valid` and return false directly.

source/fuzz/transformation_add_parameter.cpp
```diff
@@ -145,6 +249,19 @@ bool TransformationAddParameter::IsParameterTypeSupported(
...
+        case SpvStorageClassPrivate:
+        case SpvStorageClassFunction:
+        case SpvStorageClassWorkgroup: {
+          auto pointee_type = type.AsPointer()->pointee_type();
+          return IsParameterTypeSupported(*pointee_type);
+        }
+        default:
+          return false;
+      }
+    }
```
Vasniktel, line 264: Nice!

source/fuzz/transformation_add_parameter.cpp
```diff
@@ -24,54 +24,113 @@ TransformationAddParameter::TransformationAddParameter(
...
+      case SpvStorageClassPrivate:
+      case SpvStorageClassWorkgroup:
+        // If there is no available global variable, the transformation is
+        // invalid.
+        if (!fuzzerutil::MaybeGetGlobalVariable(ir_context, pointer_type_id)) {
+          is_valid = false;
+        }
+        break;
+      default:
+        break;
```
Vasniktel, line 102: It would be good to make an assert here saying that storage class is not supported.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -755,8 +755,9 @@ message TransformationAddParameter {
 
   // Result id of the instruction, used to initializer new parameter
   // in function calls. Type id of that instruction is the type id of the parameter.
-  // It may not be OpTypeVoid.
-  uint32 initializer_id = 3;
+  // It may not be OpTypeVoid. In case of new parameter being a pointer, this is the type id
+  // of the pointer type. Initializer id and pointer type id are mutually exclusive.
+  uint32 initializer_id_or_pointer_type_id = 3;
```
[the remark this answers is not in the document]
Vasniktel: Or maybe `parameter_type_id` to be consistent with the naming.

source/fuzz/fuzzer_pass_add_parameters.cpp
paulthomson: I think it works.

paulthomson:
I think it will be good to make a design doc for this change and we can all clarify some things there before we proceed!

Vasniktel:
@antnkarp, please change `Fixes ...` to `Part of` in the PR description so that the related issue is not closed when the PR is merged.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -69,15 +72,72 @@ void FuzzerPassAddParameters::Apply() {
...
+      std::map<uint32_t, uint32_t> call_parameter_id;
+
+      // Consider the case when a pointer type was selected.
+      if (current_type->kind() == opt::analysis::Type::kPointer) {
+        auto storage_class = fuzzerutil::GetStorageClassFromPointerType(
+            GetIRContext(), current_type_id);
+        switch (storage_class) {
+          case SpvStorageClassFunction: {
+            // In every caller find or create a local variable with that has the
+            // selected type.
```
Vasniktel, line 90: The comment needs refactoring ('with that' issue).

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -69,15 +72,72 @@ void FuzzerPassAddParameters::Apply() {
...
+                call_parameter_id[instr->result_id()] = variable_id;
+              }
+            }
+          } break;
+          default:
+            break;
+        }
+      } else {
+        // If there exists at least one caller, find or create a zero constant
+        // with that has the selected type.
```
Vasniktel, line 119: The comment needs refactoring ('with that' issue).

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -69,15 +72,72 @@ void FuzzerPassAddParameters::Apply() {
...
+          for (auto* instr :
+               fuzzerutil::GetCallers(GetIRContext(), function.result_id())) {
+            call_parameter_id[instr->result_id()] = constant_id;
+          }
+        }
+      }
+
+      // If the function has no callers, and a selected type to the key 0. It is
+      // necessary to pass information of the new type to the transformation.
+      // Since there are no callers, we pass the type id, not the value id.
```
Vasniktel, line 134: Thanks for spotting this issue. I'm not sure about the current approach to resolving it, though. Maybe it is better to have a separate field with the parameter's type id. I haven't used the type id initially because it was always possible to compute it. In this case, however, using an explicit type id will make the code cleaner. Besides, we still need to check that all values in the `call_parameter_id` have the same type even without an explicit type id field.

Vasniktel: @antnkarp, @paulthomson, what do you think?

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
 }
 
+uint32_t FuzzerPassAddParameters::FindOrCreateLocalVariable(
+    uint32_t pointer_type_id, uint32_t function_id,
+    bool pointee_value_is_irrelevant) {
+  uint32_t result_id = 0;
+  auto pointer_type = GetIRContext()->get_type_mgr()->GetType(pointer_type_id);
+  // No unused variables in release mode.
+  (void)pointer_type;
+  assert(pointer_type->AsPointer() &&
```
Vasniktel, line 161: You can use `assert(pointer_type && pointer_type->AsPointer() && pointer_type->AsPointer()->storage_class() == Function...`

Vasniktel: i.e. don't forget to check that `pointer_type` is not null.

Vasniktel: However, maybe it's OK if this function is only used by the `FuzzerPassAddParameters`. @antnkarp, @paulthomson, are we planning on adding it to the `FuzzerPass`?

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
+  auto function = fuzzerutil::FindFunction(GetIRContext(), function_id);
+  assert(function && "The function must be defined.");
+
+  // All of the local variable declarations are located in the first block.
+  auto block = function->begin();
+  for (auto& instruction : *block) {
+    if (instruction.opcode() != SpvOpVariable) {
+      continue;
+    }
+    if (!instruction.type_id() || instruction.type_id() != pointer_type_id) {
```
Vasniktel, line 172: `OpVariable` always has a type id so I think this check (`!instruction.type_id()`) is unnecessary. But you can leave it if you want to.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
+      continue;
+    }
+
+    // Check if the found variable is marked with PointeeValueIsIrrelevant.
+    if (!GetTransformationContext()->GetFactManager()->PointeeValueIsIrrelevant(
+            instruction.result_id())) {
```
Vasniktel, line 178: I think the correct condition is 
```cpp
if (GetTransformationContext()->GetFactManager()->PointeeValueIsIrrelevant(
            instruction.result_id()) != pointee_value_is_irrelevant) {
```

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
+      continue;
+    }
+    result_id = instruction.result_id();
+    break;
```
Vasniktel, line 182: Consider returning `instruction.result_id()` directly. It is always a good idea to restrict the scope in which the variable can be used.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
+  }
+
+  // No such variable was found. Apply a transformation to get one.
+  if (!result_id) {
```
Vasniktel, line 186: This check is unnecessary if you return from the loop right after the variable is found.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
+  return result_id;
+}
+
+uint32_t FuzzerPassAddParameters::FindOrCreateGlobalVariable(
+    uint32_t pointer_type_id, bool pointee_value_is_irrelevant) {
+  uint32_t result_id = 0;
+  auto pointer_type = GetIRContext()->get_type_mgr()->GetType(pointer_type_id);
+  // No unused variables in release mode.
+  (void)pointer_type;
+  assert(pointer_type->AsPointer() &&
```
Vasniktel, line 204: This the comment about a similar assertion in the other method.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
+         "The pointer_type_id must refer to a pointer type");
+  for (auto& instruction : GetIRContext()->module()->types_values()) {
+    if (instruction.opcode() != SpvOpVariable) {
+      continue;
+    }
+    if (!instruction.type_id() || instruction.type_id() != pointer_type_id) {
```
Vasniktel, line 210: See the comment on the similar check in the other method.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
+      continue;
+    }
+
+    // Check if the found variable is marked with PointeeValueIsIrrelevant.
+    if (!GetTransformationContext()->GetFactManager()->PointeeValueIsIrrelevant(
```
Vasniktel, line 215: See the comment on the similar check in the other method.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
+            instruction.result_id())) {
+      continue;
+    }
+    result_id = instruction.result_id();
```
Vasniktel, line 219: Consider returning directly.

source/fuzz/fuzzer_pass_add_parameters.cpp
```diff
@@ -91,5 +151,91 @@ uint32_t FuzzerPassAddParameters::GetNumberOfParameters(
...
+        GetIRContext(), pointer_type_id);
+    auto storage_class = fuzzerutil::GetStorageClassFromPointerType(
+        GetIRContext(), pointer_type_id);
+    assert((storage_class == SpvStorageClassPrivate ||
+            storage_class == SpvStorageClassWorkgroup) &&
+           "The storage class must be Private or Workgroup");
+    result_id = GetFuzzerContext()->GetFreshId();
+    ApplyTransformation(TransformationAddGlobalVariable(
+        result_id, pointer_type_id, storage_class,
+        FindOrCreateZeroConstant(pointee_type_id, true),
```
Vasniktel, line 235: If I recall correctly, `TransformationAddGlobalVariable` requires initializer to be `0` if the storage class is `Workgroup`.

source/fuzz/transformation_add_parameter.cpp
```diff
@@ -24,37 +24,84 @@ TransformationAddParameter::TransformationAddParameter(
     : message_(message) {}
 
 TransformationAddParameter::TransformationAddParameter(
-    uint32_t function_id, uint32_t parameter_fresh_id, uint32_t initializer_id,
+    uint32_t function_id, uint32_t parameter_fresh_id,
+    std::map<uint32_t, uint32_t>&& call_parameter_id,
```
Vasniktel, line 28: As per our discussion the other day, consider using `std::map<uint32_t, uint32_t> call_parameter_id` if you want to use `std::move`. The effect is the same but API is better (recall `MakeDataDescriptor` and the fact that we are unable to pass anything other than rvalue).

source/fuzz/transformation_add_parameter.cpp
```diff
@@ -24,37 +24,84 @@ TransformationAddParameter::TransformationAddParameter(
...
+  std::map<uint32_t, uint32_t> call_parameter_id_map =
+      fuzzerutil::RepeatedUInt32PairToMap(message_.call_parameter_id());
+  uint32_t new_parameter_type_id = 0;
+  // Iterate over all callers.
+  for (auto* instr :
+       fuzzerutil::GetCallers(ir_context, message_.function_id())) {
+    uint32_t caller_id = instr->result_id();
+
+    // If there is no entry for this caller, return false.
+    if (call_parameter_id_map.find(caller_id) == call_parameter_id_map.end()) {
```
Vasniktel, line 55: Alternatively, you can use `call_parameter_id_map.count(caller_id)`. Its advantage is a smaller amount of code but it's not an ideal solution either. In any way, I'm OK with this approach too.

test/fuzz/transformation_add_parameter_test.cpp
```diff
@@ -108,53 +107,53 @@ TEST(TransformationAddParameterTest, BasicTest) {
                                                validator_options);
 
   // Can't modify entry point function.
-  ASSERT_FALSE(TransformationAddParameter(4, 60, 12, 61)
+  ASSERT_FALSE(TransformationAddParameter(4, 60, {{0, 7}}, 61)
```
Vasniktel, line 110: I think one of the compilers produces strange errors in release mode if we use maps this way. A solution is to wrap it in an additional pair of braces. `{{{0, 7}}}`.

source/fuzz/transformation_add_parameter.cpp
```diff
@@ -81,17 +136,22 @@ void TransformationAddParameter::Apply(
...
+  // irrelevant so that we can replace its use with some other id. If the
+  // |new_parameter_type_id| is not a pointer type, we cannot mark it, because
+  // this pointer might be replaced by a pointer from original shader. This
+  // would change the semantics of the module.
+  auto new_parameter_type =
+      ir_context->get_type_mgr()->GetType(new_parameter_type_id);
+  if (new_parameter_type->kind() != opt::analysis::Type::kPointer) {
+    transformation_context->GetFactManager()->AddFactIdIsIrrelevant(
+        message_.parameter_fresh_id());
+  }
```
paulthomson, line 149: If we are adding a pointer parameter, we should mark it as PointeeValueIsIrrelevant.

source/fuzz/fuzzer_pass_add_parameters.cpp
antnkarp: In this fragment of code, I've used FindOrCreateZeroConstant so it seems to me that it creates a valid Workgroup variable. Do you mean 0 as in integer constant of 0?

Vasniktel: No, sorry for the confusion. I meant that an `OpVariable` instruction shouldn't have an initializer. Thus, we pass `0` instead of the result of `FindOrCreateZero`.

antnkarp: Fixed in 007db36

source/fuzz/fuzzer_pass.cpp
```diff
@@ -595,5 +597,99 @@ opt::BasicBlock* FuzzerPass::GetOrCreateSimpleLoopPreheader(
...
+      "The pointer_type_id must refer to a defined pointer type with storage "
+      "class Private or Workgroup");
+  for (auto& instruction : GetIRContext()->module()->types_values()) {
+    if (instruction.opcode() != SpvOpVariable) {
+      continue;
+    }
+    // Every OpVariable has its type id defined.
+    if (instruction.type_id() != pointer_type_id) {
+      continue;
+    }
```
paulthomson, line 663: It sounds like this should be an assert?

test/fuzz/transformation_add_parameter_test.cpp
```diff
@@ -240,10 +239,343 @@ TEST(TransformationAddParameterTest, BasicTest) {
...
+      transformation_bad_4.IsApplicable(context.get(), transformation_context));
+
+  // Function with id 14 does not have any callers.
+  // Bad:  Id 3 refers to OpTypeVoid, which is not supported.
+  TransformationAddParameter transformation_bad_6(14, 50, 3, {{}}, 51);
+  ASSERT_FALSE(
+      transformation_bad_6.IsApplicable(context.get(), transformation_context));
+}
+
+TEST(TransformationAddParameterTest, PointerTypeTest) {
```
paulthomson, line 362: The tests look good, but please add global pointer parameters with storage class Private and Workgroup. Feel free to do this in a new test (or 2 new tests), and no need to cover the many different failure cases. But we at least need the success cases, as this would have caught the mistake of trying to give a workgroup variable an initializer. 

You might want a compute shader example from here:

https://github.com/google/graphicsfuzz/tree/master/shaders/src/main/glsl/samples/compute/320es

source/fuzz/fuzzer_pass.cpp
```diff
@@ -595,5 +597,99 @@ opt::BasicBlock* FuzzerPass::GetOrCreateSimpleLoopPreheader(
...
+
+  // A variable with storage class Workgroup shouldn't have an initializer.
+  if (storage_class == SpvStorageClassWorkgroup) {
+    ApplyTransformation(TransformationAddGlobalVariable(
+        result_id, pointer_type_id, SpvStorageClassWorkgroup, 0,
+        pointee_value_is_irrelevant));
+  } else {
+    ApplyTransformation(TransformationAddGlobalVariable(
+        result_id, pointer_type_id, SpvStorageClassPrivate,
+        FindOrCreateZeroConstant(pointee_type_id, true),
```
paulthomson, line 688: I am unsure if this should be "true" or pointee_value_is_irrelevant or "false". I think pointee_value_is_irrelevant should work.

source/fuzz/fuzzer_pass.cpp
antnkarp: I don't think so, because there could many local/global variables that have instruction.type_id() different than the supplied pointer_type_id. In that case we just continue to the next instruction.

source/fuzz/fuzzer_pass.cpp
```diff
@@ -595,5 +597,99 @@ opt::BasicBlock* FuzzerPass::GetOrCreateSimpleLoopPreheader(
...
+  assert(pointer_type && pointer_type->AsPointer() &&
+         pointer_type->AsPointer()->storage_class() ==
+             SpvStorageClassFunction &&
+         "The pointer_type_id must refer to a defined pointer type with "
+         "storage class Function");
+  auto function = fuzzerutil::FindFunction(GetIRContext(), function_id);
+  assert(function && "The function must be defined.");
+
+  // All of the local variable declarations are located in the first block.
+  auto block = function->begin();
```
Vasniktel, line 615: Consider moving this into the loop:
```cpp
for (auto& ... : *function->begin())
```

source/fuzz/transformation_add_parameter.cpp
```diff
@@ -24,40 +24,74 @@ TransformationAddParameter::TransformationAddParameter(
...
+    auto value_instr = ir_context->get_def_use_mgr()->GetDef(value_id);
+    if (!value_instr) {
+      return false;
+    }
+    // If the id of the value of the map is not available before the caller,
+    // return false.
+    if (!fuzzerutil::IdIsAvailableBeforeInstruction(ir_context, instr,
+                                                    value_id)) {
+      return false;
+    }
```
Vasniktel, line 82: This can potentially trigger a bug if the caller is in an unreachable block. `fuzzerutil::IdIsAvailableBeforeInstruction` uses dominator analysis to check that `value_id` is available and the domination rules are not defined for unreachable blocks.

A solution is to adjust `fuzzerutil::IdIsAvailableBeforeInstruction` so that it uses DFS instead of the DominatorAnalysis. Or we could just check that the block is not unreachable.

Either way, I think it is better to leave it as it is for now and create a separate issue for this. @paulthomson @antnkarp , what do you think?

Vasniktel: I think this issue is common for many transformations. Maybe we could implement our own alternative to DominatorAnalysis or adjust the existing one?

source/fuzz/transformation_add_parameter.h
```diff
@@ -46,8 +51,9 @@ class TransformationAddParameter : public Transformation {
   // - Creates a new OpFunctionParameter instruction with result id
   //   |parameter_fresh_id| for the function with |function_id|.
   // - Adjusts function's type to include a new parameter.
-  // - Adds |initializer_id| as a new operand to every OpFunctionCall
-  //   instruction that calls the function.
+  // - There could be no callers. Otherwise, for every caller adds as a new
+  //   operand to every caller the value of the map |call_parameter_id| for the
+  //   key of the id of this caller.
```
Vasniktel, line 56: I would write it as follows: 
> Adds an argument to every caller of the function to account for the added parameter. The argument is the value in |call_parameter_id| map.

test/fuzz/transformation_add_parameter_test.cpp
```diff
@@ -240,7 +239,576 @@ TEST(TransformationAddParameterTest, BasicTest) {
...
+  const auto consumer = nullptr;
+  const auto context = BuildModule(env, consumer, shader, kFuzzAssembleOption);
+  ASSERT_TRUE(IsValid(env, context.get()));
+
+  FactManager fact_manager;
+  spvtools::ValidatorOptions validator_options;
+  TransformationContext transformation_context(&fact_manager,
+                                               validator_options);
+
+  // Bad: Id 19 is not available in the caller that has id 25.
```
Vasniktel, line 329: I think this comment needs refactoring.

test/fuzz/transformation_add_parameter_test.cpp
```diff
@@ -240,7 +239,576 @@ TEST(TransformationAddParameterTest, BasicTest) {
                OpReturn
                OpFunctionEnd
   )";
+  ASSERT_TRUE(IsEqual(env, expected_shader, context.get()));
+}
+
+TEST(TransformationAddParameterTest, NonPointerNotApplicableTest) {
+  // This types handles case of adding a new parameter of a non-pointer type
+  // where the transformation is not applicable.
```
Vasniktel, line 247: Add a test case for the fact that `call_parameter_id` should have an entry for _at least_ every caller (e.i. add some non-caller entries to the map).

test/fuzz/transformation_add_parameter_test.cpp
Vasniktel: Id 25 is not a caller.

source/fuzz/transformation_add_parameter.cpp
antnkarp: If it is common for many transformations, I agree that we need to create a sperate issue.

Vasniktel: Please, add a TODO(#3722) comment explaining that we need to refactor this. Use an original issue URL instead of #...

source/fuzz/transformation_add_parameter.cpp
```diff
@@ -81,17 +125,24 @@ void TransformationAddParameter::Apply(
...
+  // |new_parameter_type_id| is a pointer type, we cannot mark it with
+  // IdIsIrrelevant, because this pointer might be replaced by a pointer from
+  // original shader. This would change the semantics of the module. In the case
+  // of a pointer type we mark it with PointeeValueIsIrrelevant.
+  if (new_parameter_type->kind() != opt::analysis::Type::kPointer) {
+    transformation_context->GetFactManager()->AddFactIdIsIrrelevant(
+        message_.parameter_fresh_id());
+  } else {
+    transformation_context->GetFactManager()->AddFactValueOfPointeeIsIrrelevant(
+        message_.parameter_fresh_id());
```
Vasniktel, line 139: Please, add a couple of tests for this.

antnkarp: Fixed in 6b25b44

source/fuzz/fuzzer_pass.cpp
paulthomson: Ah sorry you are right! I think the comment confused me. I somehow just assumed the code was checking that the type id was non-zero.

Maybe consider changing the comment to: 

`The existing OpVariable must have type |pointer_type_id|.`

You could also add a comment at the start of the loop (perhaps before your "All of the local variable..." comment) saying something like: "First, we try to find a suitable existing variable."

paulthomson (approved):
Awesome! Thanks a lot for your work on this.

Issue #3107, opened 2019-12-18
spirv-dis: Invalid extended instruction import 'SPIRV.debug'


jinz2014:
Can you please explain what I can do after running the spirv disassembler  ?

Thanks

; SPIR-V
; Version: 1.0
; Generator: Khronos LLVM/SPIR-V Translator; 14
; Bound: 1362
; Schema: 0
               OpCapability Addresses
               OpCapability Linkage
               OpCapability Kernel
               OpCapability Float64
               OpCapability Int64
          %1 = OpExtInstImport "OpenCL.std"
error: 22: Invalid extended instruction import 'SPIRV.debug'

dj2:
It looks like the disassembly failed, it found an import instruction it doesn't understand, so there is something in your SPIR-V binary file it doesn't like.

Normally, you can send the disassembly back through spirv-as to re-build it into a binary.

jinz2014:
Will SPIRV.debug be supported in the future ?   The disassember does not generate an assembly file after the error. I am not sure how  I can send the disassembly back through spirv-as to re-build it into a binary.

Thanks

Ryp:
I'm hitting this today while trying to run `spirv-val` on a HLSL compute shader with a `printf` statement.

PR #3628, opened 2020-07-31
Connect OpenCL.DebugInfo.100 with "SPIRV.debug"

NikitaRudenkoIntel:
The SPIRV-LLVM-Translator by default uses "SPIRV.debug" instead of "OpenCL.DebugInfo.100". This commit allows to tolerate this behavior.

Resolves issue #3107
https://github.com/KhronosGroup/SPIRV-LLVM-Translator

### DIFF
--- a/source/ext_inst.cpp
+++ b/source/ext_inst.cpp
@@ -127,6 +127,9 @@ spv_ext_inst_type_t spvExtInstImportTypeGet(const char* name) {
   if (!strcmp("OpenCL.DebugInfo.100", name)) {
     return SPV_EXT_INST_TYPE_OPENCL_DEBUGINFO_100;
   }
+  if (!strcmp("SPIRV.debug", name)) {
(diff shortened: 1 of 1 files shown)


alan-baker:
This seems like a bug in the SPIRV-LLVM-Translator. Why doesn't the translator output the published name?

jaebaek:
I agree with alan. I tested SPIRV-LLVM-Translator for the SPIR-V code with "OpenCL.DebugInfo.100" but it does not work. I did not send a PR to SPIRV-LLVM-Translator, but I believe it is SPIRV-LLVM-Translator's bug that it says "OpenCL.DebugInfo.100" is unknown.

NikitaRudenkoIntel:
Ok, will try on the other side...
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/659

NikitaRudenkoIntel:
KhronosGroup/SPIRV-LLVM-Translator#659 merged

PR #3652, opened 2020-08-04
Preserve debug info in dead-insert-elim pass

jaebaek:
When we determine the liveness of an insert, we have to ignore the debug
instruction (it must not have any impact on the decision).

### DIFF
--- a/test/opt/dead_insert_elim_test.cpp
+++ b/test/opt/dead_insert_elim_test.cpp
@@ -653,6 +653,7 @@ OpDecorate %_ Binding 0
 %28 = OpLoad %float %In1
 %29 = OpFAdd %float %27 %28
 
+; CHECK:      [[repl:%\w+]] = OpLoad %v2float %In2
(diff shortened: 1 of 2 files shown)


test/opt/dead_insert_elim_test.cpp
```diff
@@ -563,6 +563,112 @@ OpFunctionEnd
...
+
+%main = OpFunction %void None %11
+%25 = OpLabel
+%26 = OpLoad %v2float %In2
+%27 = OpLoad %float %In0
+%28 = OpLoad %float %In1
+%29 = OpFAdd %float %27 %28
+
+; CHECK-NOT:  OpCompositeInsert
+; CHECK:      DebugValue {{%\w+}} [[repl:%\w+]]
```
s-perron, line 657: You should check that "repl" is the load that you expect it to be.

jaebaek: done

jaebaek:
Thanks!

PR #3660, opened 2020-08-06
spirv-fuzz: Add DecisionMaker

### LINKED ISSUE #2849: fuzz: centralize probabilities so we can easily bias random choices
For example, the following allows the caller to specify probabilities:

`rng->ChoosePercentage(GetChanceOfMovingBlockDown())`

Instead, the rng caller should provide the number of choices it has available, plus some hints as to why it needs to make the choice and any other information that could be useful:

```c++
rng->ChooseIntUpTo(
  15, // the number of ways to fuzz a constant 
      // (would probably be something like WaysToFuzz.SIZE, assuming this was an enum)
  CHOOSING_HOW_TO_FUZZ_CONSTANT,  // a hint at why we are making this choice
  {DEPTH_HINT, 5} // further pairs adding extra, optional information, that won't necessarily be used or applicable
)
```

Thus, the implementation of `ChooseIntUpTo` can always just return a uniform random choice. But, we can easily add additional checks as needed (a switch statement) to bias probabilities from a centralized place and even tweak these at run time. Furthermore, callers can easily add as much additional information as they like, even if it won't be used yet.

When the implementation sees `CHOOSING_HOW_TO_FUZZ_CONSTANT`, it may know that the choice is being made over all enum values of `WaysToFuzz`. This won't work if the number of choices varies depending on what is possible in the given context; in this case, perhaps the meaning of each choice should be clarified via hints, or perhaps we should have another `Choose*` method that takes the list of choices (as an iterable list of size_t, assuming they will always be enums).

```c++
rng->ChooseEnum(
  ways_to_fuzz, // vector of enum values; possibly need to cast it or could
                // template this function based on enum type
  CHOOSING_HOW_TO_FUZZ_CONSTANT,  // a hint at why we are making this choice
  {DEPTH_HINT, 5} // further pairs adding extra, optional information, that won't necessarily be used or applicable
)
```

The downside to templating is perhaps it would add more effort if we just want to, at run time, decide to switch to code that doesn't distinguish between the type of the enum. Not completely sure how this would look though.

paulthomson:
Introduces a prototype of the DecisionMaker, a class that resolves all random decisions made by
 fuzzer passes. It can also be mocked to allow testing of fuzzer passes.

This is just a prototype and we probably don't want to merge this yet.

Fixes #2849.

### DIFF
--- a/source/fuzz/CMakeLists.txt
+++ b/source/fuzz/CMakeLists.txt
@@ -31,6 +31,7 @@ if(SPIRV_BUILD_FUZZER)
   set(SPIRV_TOOLS_FUZZ_SOURCES
         call_graph.h
         data_descriptor.h
+        decision_maker.h
         equivalence_relation.h
         fact_manager.h
         force_render_red.h
@@ -165,6 +166,7 @@ if(SPIRV_BUILD_FUZZER)
 
         call_graph.cpp
         data_descriptor.cpp
+        decision_maker.cpp
         fact_manager.cpp
         force_render_red.cpp
         fuzzer.cpp
--- a/source/fuzz/decision_maker.cpp
+++ b/source/fuzz/decision_maker.cpp
@@ -1,0 +1,98 @@
+// Copyright (c) 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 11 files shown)


Vasniktel:
That's nice! I have a couple of thoughts regarding the design of the component.

With the current design, `DecisionMaker` will end up having lots of functions related to various fuzzer passes (this is pretty much the situation the `FuzzerContext` is currently in). Maybe it is better for the `DecisionMaker` to have only probability related functions (like `ChooseEven` etc) and refactor all other functions into a special class, unique for every fuzzer pass.
That class would have a pointer to the decision-maker. This approach will make the code cleaner.

PR #3664, opened 2020-08-06
spirv-fuzz: Ignore specialization constants

### LINKED ISSUE #3663: spirv-fuzz: Segmentation fault in FuzzerPassInterchangeZeroLikeConstants::FindOrCreateToggledConstant
```
Program received signal SIGSEGV, Segmentation fault.
0x0000555555daae30 in
spvtools::fuzz::FuzzerPassInterchangeZeroLikeConstants::FindOrCreateToggledConstant
at source/fuzz/fuzzer_pass_interchange_zero_like_constants.cpp:41
```

Debugging `spirv-fuzz`, I found that the segmentation fault is caused because `FindDeclaredConstant` returns a null pointer that is used to call a function.

```
auto constant = GetIRContext()->get_constant_mgr()->FindDeclaredConstant(declaration->result_id());

if (!constant->IsZero()) {
  return 0;
}
```

`declaration` is `%10 = OpSpecConstant %5 1`, but `opt::analysis::ConstantManager` does not support `OpSpecConstant`.

andreperezmaselco:
`FuzzerPassInterchangeSignednessOfIntegerOperands` and `FuzzerPassInterchangeZeroLikeConstants` both included specialization constants when trying to find integer constants with known values. However, this is incorrect behavior because we do not know the value of specialization constants. Furthermore, ConstantManager does not support them, and this led to crashes where we assumed we could look up specialization constants via the ConstantManager.

This change fixes both passes to ignore specialization constants.

Fixes #3663.

### DIFF
--- a/source/fuzz/fuzzer_pass_interchange_signedness_of_integer_operands.cpp
+++ b/source/fuzz/fuzzer_pass_interchange_signedness_of_integer_operands.cpp
@@ -91,8 +91,8 @@ void FuzzerPassInterchangeSignednessOfIntegerOperands::Apply() {
 
 uint32_t FuzzerPassInterchangeSignednessOfIntegerOperands::
     FindOrCreateToggledIntegerConstant(uint32_t id) {
-  // |id| must represent a non-specialization constant because the constant
-  // manager does not declare specialization constants.
+  // |id| must not be a specialization constant because we do not know the value
+  // of specialization constants.
   if (opt::IsSpecConstantInst(
           GetIRContext()->get_def_use_mgr()->GetDef(id)->opcode())) {
     return 0;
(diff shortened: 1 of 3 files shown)


Vasniktel:
I think it would be better to mark it as `spirv-opt` PR and ask guys from `spirv-opt` team to review it (since the change to the constant manager might break some code in the `spirv-opt` tool).

Also, the constant manager is somewhat nasty. Maybe there is a way to get rid of it in the fuzzer. @andreperezmaselco, what o you think?

andreperezmaselco:
Yes. There is a way to fix it in the `spirv-fuzz` code.
I'm going to fix it.

source/fuzz/fuzzer_pass_interchange_signedness_of_integer_operands.cpp
```diff
@@ -91,6 +91,13 @@ void FuzzerPassInterchangeSignednessOfIntegerOperands::Apply() {
 
 uint32_t FuzzerPassInterchangeSignednessOfIntegerOperands::
     FindOrCreateToggledIntegerConstant(uint32_t id) {
+  // |id| must represent a non-specialization constant because the constant
+  // manager does not declare specialization constants.
```
paulthomson, line 95: Sorry for missing this before but I think the more accurate comment is:

|id| must not be a specialization constant because we do not know the value of specialization constants.

paulthomson: Even if the constant manager could handle them, I don't think we could use them here.

andreperezmaselco: Done.

PR #3666, opened 2020-08-06
spirv-fuzz: Fix in operand type assertion

### LINKED ISSUE #3665: spirv-fuzz: Assertion 'in_operand.type == SPV_OPERAND_TYPE_ID' failed
```
spirv-fuzz: source/fuzz/id_use_descriptor.cpp:55: spvtools::fuzz::protobufs::IdUseDescriptor
spvtools::fuzz::MakeIdUseDescriptorFromUse(spvtools::opt::IRContext*, spvtools::opt::Instruction*, uint32_t):
Assertion `in_operand.type == SPV_OPERAND_TYPE_ID' failed.
Aborted (core dumped)
```
Debugging `spirv-fuzz`, I found that the assertion failed because `in_operand` was the first operand of an `OpControlBarrier` instruction and `in_operand.type = SPV_OPERAND_TYPE_SCOPE_ID`.

andreperezmaselco:
`spvtools::fuzz::MakeIdUseDescriptorFromUse` asserts if the id use type is
`SPV_OPERAND_TYPE_ID`. The problem is that not all id types are covered by
this condition. The bug was found because the first operand of an `OpControlBarrier`
instruction has `SPV_OPERAND_TYPE_SCOPE_ID` as type. Therefore, to cover
all cases the `spvIsInIdType` function is used.

Fixes #3665.

### DIFF
--- a/source/fuzz/id_use_descriptor.cpp
+++ b/source/fuzz/id_use_descriptor.cpp
@@ -52,7 +52,7 @@ protobufs::IdUseDescriptor MakeIdUseDescriptorFromUse(
     opt::IRContext* context, opt::Instruction* inst,
     uint32_t in_operand_index) {
   const auto& in_operand = inst->GetInOperand(in_operand_index);
-  assert(in_operand.type == SPV_OPERAND_TYPE_ID);
+  assert(spvIsInIdType(in_operand.type));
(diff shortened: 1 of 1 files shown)


andreperezmaselco:
Thanks, @paulthomson!

Issue #3544, opened 2020-07-16
spirv-fuzz: Transformation to convert OpSelect to conditional branch (and vice versa?)


paulthomson:

afd:
@stefanomil I think this would be a good one for you to look at in due course.

Turning OpSelect into a conditional branch should be relatively straightforward.

The "vice versa" part might be a lot harder but is potentially very interesting.

stefanomil:
@afd Thanks, I'll look into it!

PR #3667, opened 2020-08-07
spirv-fuzz: Transformation to flatten conditional branch

stefanomil:
This transformation, given the header of a selection construct with
branching instruction OpBranchConditional, flattens it.
Side-effecting operations such as OpLoad, OpStore and OpFunctionCall
are enclosed within smaller conditionals.
It is applicable if the construct does not contain inner selection
constructs or loops, or atomic or barrier instructions.

The corresponding fuzzer pass looks for selection headers and
tries to flatten them.

Needed for the issue #3544 , but it does not fix it completely.


source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+              // instruction from the number of overflow ids needed.
+              overflow_ids_needed =
+                  fresh_ids.size() > overflow_ids_needed
+                      ? 0
+                      : overflow_ids_needed - (uint32_t)fresh_ids.size();
+            }
+
+            // We need |overflow_ids_needed| overflow ids
+            if (overflow_ids_used + overflow_ids_needed >
+                (uint32_t)message_.overflow_id_size()) {
```
Vasniktel, line 216: Consider using `static_cast`.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+
+            // The ids must all be distinct and unused.
+            for (uint32_t fresh_id : fresh_ids) {
+              if (!CheckIdIsFreshAndNotUsedByThisTransformation(
+                      fresh_id, ir_context, &used_fresh_ids)) {
+                return false;
+              }
+            }
+
+            // All OpSampledImage ids defined before these point should not be
```
Vasniktel, line 234: these -> this

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+              instruction->opcode() == SpvOpStore ||
+              instruction->opcode() == SpvOpFunctionCall) {
+            // Keep a vector of the fresh ids needed by this instruction.
+            std::vector<uint32_t> fresh_ids;
+            uint32_t overflow_ids_needed;
+
+            // Initialise overflow_ids_needed to the total number of ids
+            // needed.
+            overflow_ids_needed =
+                NumOfFreshIdsNeededByOpcode(instruction->opcode());
```
Vasniktel, line 199: The code for computing a precise number of fresh ids used by each memory instruction is smart but I find it hard to understand. 

It might be better to simply have a counter and increment it every time we find a memory instruction. If the counter is greater than the size of the `overflow_ids` - return false.

Then, as a separate check, we could ensure that all ids in `overflow_ids` are unique and fresh. We might check more ids than we will actually use in the transformation but I don't think this is a problem. @paulthomson, what do you think regarding this check causing any problems in the shrinker?

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+  // Keep track of the number of overflow ids used.
+  uint32_t overflow_ids_used = 0;
+
+  auto branch_instruction = header_block->terminator();
+
+  opt::BasicBlock* last_true_block = nullptr;
+
+  // Adjust the conditional branches by enclosing problematic instructions are
+  // enclosed by conditionals and get references to the last block in each
+  // branch.
```
Vasniktel, line 287: This comment needs refactoring :)

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+            NumOfFreshIdsNeededByOpcode(instruction->opcode());
+        std::vector<uint32_t> fresh_ids;
+
+        // Get them from the map.
+        if (instructions_to_fresh_ids.count(instruction) != 0) {
+          fresh_ids = instructions_to_fresh_ids[instruction];
+        }
+
+        // Get the ones still needed from the overflow ids.
+        for (int still_needed = ids_needed - (int)fresh_ids.size();
```
Vasniktel, line 339: I'm not sure about this minus operator. If I'm not mistaken, `(int)fresh_ids.size()` will be cast to `uint32_t` which might cause an overflow when subtracted.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1053,6 +1066,23 @@ message TransformationEquationInstruction {
...
+  // The label id of the header block
+  uint32 header_block_id = 1;
+
+  // A list of pairs mapping instruction descriptors to fresh ids.
+  // Necessary for store, load and function call instructions.
+  repeated InstructionUint32ListPair instruction_to_fresh_ids = 2;
+
+  // A list of fresh ids to be used for instructions that need them, if
+  // they are not in instruction_to_fresh_ids.
+  repeated uint32 overflow_id = 3;
```
Vasniktel, line 1083: I wonder if we need this if all ids are accounted for in the previous field.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+  switch (opcode) {
+    case SpvOpStore:
+      return 2;
+    case SpvOpLoad:
+    case SpvOpFunctionCall:
+      return 5;
+    default:
+      assert(false && "This line should never be reached");
+      return 0;
+  }
```
Vasniktel, line 444: It would be good to leave at least a small comment regarding these numbers. Although, I realize it is hard to do so without diagrams.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+}
+
+opt::BasicBlock*
+TransformationFlattenConditionalBranch::EncloseInstructionInConditional(
+    opt::IRContext* ir_context, TransformationContext* transformation_context,
+    opt::BasicBlock* block, opt::Instruction* instruction,
+    std::vector<uint32_t> fresh_ids, uint32_t condition_id,
```
Vasniktel, line 451: Consider using `const std::vector<uint32_t>&` so that we don't make an unnecessary copy of the argument.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+  }
+
+  // Initially, consider the merge block as the alternative block to branch to
+  // if the instruction should not be executed.
+  auto alternative_block = merge_block;
+
+  // Add an unconditional branch from |execute_block| to |merge_block|.
+  execute_block->AddInstruction(MakeUnique<opt::Instruction>(
+      ir_context, SpvOpBranch, 0, 0,
+      std::initializer_list<opt::Operand>{opt::Operand(
```
Vasniktel, line 498: Consider using `opt::Instruction::OperandList` instead of `std::initializer_list<opt::Operand>`. If you do so, you can get rid of `opt::Operand` and replace it with `{}`.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+          spv_operand_type_t::SPV_OPERAND_TYPE_ID, {merge_block->id()})}));
```
Vasniktel, line 499: You can get rid of `spv_operand_type_t::`.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+  //   instruction, in the merge block.
+  if (instruction->HasResultId()) {
+    // We need 3 more fresh ids for 1 additional block and 2 additional
+    // instructions.
+    assert(fresh_ids.size() >= 5 && "Not enough fresh ids.");
+
+    // Create a new block using a fresh id for its label.
+    auto alternative_block_temp = MakeUnique<opt::BasicBlock>(
+        MakeUnique<opt::Instruction>(ir_context, SpvOpLabel, 0, fresh_ids[2],
+                                     std::initializer_list<opt::Operand>{}));
```
Vasniktel, line 515: `std::initializer_list<opt::Operand>` -> `opt::Instruction::OperandList`.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+    uint32_t original_result_id = instruction->result_id();
+
+    // Set the result id of the instruction to a fresh id.
+    instruction->SetResultId(fresh_ids[3]);
+
+    // Add an OpUndef instruction, with the same type as the original
+    // instruction and a fresh id, to the new block.
+    alternative_block_temp->AddInstruction(MakeUnique<opt::Instruction>(
+        ir_context, SpvOpUndef, instruction->type_id(), fresh_ids[4],
+        std::initializer_list<opt::Operand>{}));
```
Vasniktel, line 527: Same comment about initializer list.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+
+    // Add an unconditional branch from the new block to the merge block.
+    alternative_block_temp->AddInstruction(MakeUnique<opt::Instruction>(
+        ir_context, SpvOpBranch, 0, 0,
+        std::initializer_list<opt::Operand>{opt::Operand(
```
Vasniktel, line 532: Same comment about initializer list.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+    // Insert the block before the merge block.
+    alternative_block = block->GetParent()->InsertBasicBlockBefore(
+        std::move(alternative_block_temp), merge_block);
+
+    // Using the original instruction result id, add an OpPhi instruction to the
+    // merge block, which will either take the value of the result of the
+    // instruction or the dummy value defined in the alternative block.
+    merge_block->begin().InsertBefore(MakeUnique<opt::Instruction>(
+        ir_context, SpvOpPhi, instruction->type_id(), original_result_id,
+        std::initializer_list<opt::Operand>{
```
Vasniktel, line 544: Same comment about initializer list.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+                         ->GetLabel()
+                         ->result_id();
+  auto else_block_id = (exec_if_cond_true ? alternative_block : execute_block)
+                           ->GetLabel()
+                           ->result_id();
+
+  // Add an OpSelectionMerge instruction to the block.
+  block->AddInstruction(MakeUnique<opt::Instruction>(
+      ir_context, SpvOpSelectionMerge, 0, 0,
+      std::initializer_list<opt::Operand>{
```
Vasniktel, line 573: Same comment about initializer list.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+                       {merge_block->id()}),
+          opt::Operand(spv_operand_type_t::SPV_OPERAND_TYPE_SELECTION_CONTROL,
+                       {0})}));
+
+  // Add an OpBranchConditional, to the block, using |condition_id| as the
+  // condition and branching to |if_block_id| if the condition is true and to
+  // |else_block_id| if the condition is false.
+  block->AddInstruction(MakeUnique<opt::Instruction>(
+      ir_context, SpvOpBranchConditional, 0, 0,
+      std::initializer_list<opt::Operand>{
```
Vasniktel, line 584: Same comment about initializer list.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+  for (auto const& pair : instructions_to_fresh_ids) {
+    protobufs::InstructionUint32ListPair mapping;
+    *mapping.mutable_instruction_descriptor() = pair.first;
+    for (auto id : pair.second) {
+      mapping.add_id(id);
+    }
+    *message_.add_instruction_to_fresh_ids() = mapping;
+    for (auto id : overflow_ids) {
+      message_.add_overflow_id(id);
+    }
```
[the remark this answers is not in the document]
stefanomil: Yes. Fixed

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+
+bool TransformationFlattenConditionalBranch::IsApplicable(
+    opt::IRContext* ir_context,
+    const TransformationContext& /* unused */) const {
+  uint32_t header_block_id = message_.header_block_id();
+
+  // |header_block_id| must refer to a block label.
+  {
+    auto label = ir_context->get_def_use_mgr()->GetDef(header_block_id);
+    if (label->opcode() != SpvOpLabel) {
```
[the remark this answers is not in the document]
stefanomil: True. Fixed

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+  // block).
+  uint32_t convergence_block_id = merge_block_id;
+  while (ir_context->cfg()->preds(convergence_block_id).size() == 1) {
+    if (convergence_block_id == header_block_id) {
+      // There is a chain of blocks with one predecessors from the header block
+      // to the merge block. This means that the region is not single-entry,
+      // single-exit (because the merge block is only reached by one of the two
+      // branches).
+      return false;
+    }
```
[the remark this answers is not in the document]
stefanomil: Yes, I have added this case to the tests

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+  if (header_block->terminator()->opcode() != SpvOpBranchConditional) {
+    return false;
+  }
+
+  // Find the first block where flow converges (it is not necessarily the merge
+  // block).
+  uint32_t convergence_block_id = merge_block_id;
+  while (ir_context->cfg()->preds(convergence_block_id).size() == 1) {
+    if (convergence_block_id == header_block_id) {
+      // There is a chain of blocks with one predecessors from the header block
```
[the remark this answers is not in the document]
stefanomil: Fixed

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+  for (auto pair : message_.instruction_to_fresh_ids()) {
+    std::vector<uint32_t> fresh_ids;
+    for (uint32_t id : pair.id()) {
+      fresh_ids.push_back(id);
+    }
+
+    auto instruction =
+        FindInstruction(pair.instruction_descriptor(), ir_context);
+    if (instruction) {
+      instructions_to_fresh_ids.emplace(instruction, fresh_ids);
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,600 @@
...
+              // There is a mapping from this instruction to a list of fresh
+              // ids.
+
+              fresh_ids = instructions_to_fresh_ids[instruction];
+              // We can deduct the number of fresh ids specific to this
+              // instruction from the number of overflow ids needed.
+              overflow_ids_needed =
+                  fresh_ids.size() > overflow_ids_needed
+                      ? 0
+                      : overflow_ids_needed - (uint32_t)fresh_ids.size();
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: Oops. Done

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: Yes, I agree that it was a bit convoluted so I changed it so that the inner loop only checks if there are enough fresh ids, and the check that they are distinct happens outside.

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: Fixed

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: I think that shouldn't be a problem, since the result would then be cast to int and will thus be interpreted as signed. The subtraction operation does not operate differently on signed and unsigned integers.

source/fuzz/protobufs/spvtoolsfuzz.proto
stefanomil: We might need this because shrinking may cause some instruction that would otherwise be removed to remain, requiring additional ids

stefanomil: Right, @paulthomson ?

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: I have added a comment

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: Done

source/fuzz/protobufs/spvtoolsfuzz.proto
Vasniktel: I'm not sure but I don't think the shrinker can create additional instructions that weren't present during the fuzzing. But I'm not sure. @paulthomson, what do you think?

Vasniktel: Even if this is the case, we could probably not apply the transformation at all. This requires some thought, though.

Vasniktel: A clarification on the last comment: 
Even if the shrinker can create new instructions, I think it would be better not to apply the transformation if we don't have a sufficient number of fresh ids.

P.S. We could have a transformation that changes (for whatever reasons) one memory instruction into another (e.g. OpStore into OpLoad). In this case, if the shrinker removes this transformation from the sequence, we could be lacking fresh ids (because OpLoad requires more of them than OpStore).

stefanomil: That is the purpose of the overflow ids: to have fresh ids even when we see "unexpected" instructions so we are still able to apply the transformation. Do you think that we should not apply the transformation in this case? (In your example, there would be no match in the map, since the instruction descriptor for the instruction would not be valid anymore)

Vasniktel: Yeah, you are right. The transformation might be inapplicable any time one changes some instructions since instruction descriptor depends on the relative order of instructions. That's not the best approach.

Still, it looks odd to me that we have two fields in the message that serve the same purpose (even if it's probably not a big deal).

@stefanomil, what do you think about leaving `overflow_id` and removing the other one?

stefanomil: Having a map helps preserve stability of fresh ids: the same instructions will get the same ids even if some previous instructions are removed (while if we just had a list, it would presumably get ids that come sooner in the list), so possible subsequent transformation would still be applicable. But just having a map would make the transformation less flexible, in that some changes could easily make it inapplicable.

So there are two fields so that the transformation is both stable and flexible.

@paulthomson can probably clarify better

Vasniktel: Indeed. Thanks for the clarification!

In that case, it might be better to mention some points from this discussion in the comments. Something like:
1. for the map field: "This field helps to preserve the stability of the transformation's result since every instruction has a specific set of ids that it can use"
2. for the `overflow_id`: "This field makes the transformation more flexible during the shrinking since every entry (the instruction descriptor) in the previous field depends on the relative position of the instructions in the module".

Nice idea BTW.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,622 @@
...
+
+    // All result ids of an OpSampledImage instruction occurring before the last
+    // point where the block will need to be split.
+    std::set<uint32_t> sampled_image_result_ids_before_split;
+
+    // All result ids of an OpSampledImage instruction occurring after the last
+    // point where the block will need to be split. They can still be used.
+    std::set<uint32_t> sampled_image_result_ids_after_split;
+
+    // Check all of the instructions in the block.
```
paulthomson, line 368: This all looks really good and I am impressed with your approach and the code clarity.

However, I think we might need to be even more careful. There are a lot of instructions that might cause trouble if they execute unconditionally because they have side-effects or undefined behavior. E.g. 

OpImageFetch
OpImageRead
OpImageWrite
OpArrayLength

OpAccessChain (does a non-constant index need to be in bounds? I think probably not, despite what the spec says (the spec will be updated eventually I think), so this instruction might be OK)


These are a bit like OpLoad and OpStore:
OpCopyMemory
OpCopyMemorySized

There are some instructions from the [GLSL extended instructions](https://www.khronos.org/registry/spir-v/specs/1.0/GLSL.std.450.html) that write their result to a pointer (but most don't):

Modf
Frexp

InterpolateAtSample (requires a capability)
...

And there might be other instructions that we have missed or that get added.

One approach is to explicitly list all instructions that we want to handle. I believe @Vasniktel did this for his PropagateInstructionUp transformation. See TransformationPropagateInstructionUp::IsOpcodeSupported. I just realised that this misses a lot of GLSL extended instructions (opcode is always OpExtInst, but then there is a second opcode for the GLSL instruction), so we should be sure to include those if we use this approach. 

Another approach is to conservatively assume that all instructions that have pointer or image or sampler in-operands are tricky. Also, all barrier instructions. We can then explicitly list the instructions that have pointer in-operands that are know are fine (e.g. OpAccessChain...I think...there might be a better example). Most of the tricky instructions can still be handled though; we just need to add a conditional branch like you do for OpLoad, OpStore, OpFunctionCall. For example, I think we could handle atomic instructions (and there would be no special code for this; it would be the same code that handles OpLoad, OpStore, ...[anything with a pointer in-operand]).

source/fuzz/fuzzer_pass_flatten_conditional_branches.cpp
```diff
@@ -0,0 +1,92 @@
...
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      // Randomly decide whether to consider this block.
+      if (!GetFuzzerContext()->ChoosePercentage(
+              GetFuzzerContext()->GetChanceOfFlatteningConditionalBranch())) {
+        continue;
+      }
+
+      // Only consider this block if it is the header of a conditional.
+      if (block.MergeBlockIdIfAny() &&
```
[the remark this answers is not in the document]
stefanomil: You are right, thanks!

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,622 @@
...
+    opt::IRContext* ir_context, opt::BasicBlock* header,
+    std::set<opt::Instruction*>* instructions_that_need_ids) {
+  uint32_t merge_block_id = header->MergeBlockIdIfAny();
+  assert(merge_block_id &&
+         header->GetMergeInst()->opcode() == SpvOpSelectionMerge &&
+         header->terminator()->opcode() == SpvOpBranchConditional &&
+         "|header| must be the header of a conditional.");
+
+  // Find the first block where flow converges (it is not necessarily the merge
+  // block).
```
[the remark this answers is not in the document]
stefanomil: I have added a description of this to the design doc

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,622 @@
...
+    convergence_block_id = ir_context->cfg()->preds(convergence_block_id)[0];
+  }
+
+  // Get all the blocks reachable by the header block before reaching the
+  // convergence block and check that, for each of these blocks:
+  //  - the header dominates it and the convergence block postdominates it (so
+  //    that the header and merge block form a single-entry, single-exit
+  //    region)
+  //  - it does not contain merge instructions
+  //  - it branches unconditionally to another block
```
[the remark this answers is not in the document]
stefanomil: Thanks, I have now added it

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,622 @@
...
+      // successors.
+      continue;
+    }
+
+    // If the block is not dominated by the header or it is not postdominated by
+    // the convergence_block, this is not a single-entry, single-exit region.
+    if (!dominator_analysis->Dominates(header->id(), block_id) ||
+        !postdominator_analysis->Dominates(convergence_block_id, block_id)) {
+      return false;
+    }
```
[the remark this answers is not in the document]
stefanomil: You are right, thanks!

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: I have now changed it so that:
- a few instructions are listed as having no side effects. They can always be executed
- barrier and OpSampledImage instructions are forbidden
- other instructions are allowed but require fresh ids (5 if they have a result id, 2 otherwise)

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+
+          // We cannot go ahead if we encounter an instruction that cannot be
+          // handled.
+          if (!InstructionCanBeHandled(instruction)) {
+            return false;
+          }
+
+          // If the instruction has side effects, add it to the
+          // |instructions_that_need_ids| set.
+          if (!fuzzerutil::InstructionHasNoSideEffects(instruction)) {
```
Vasniktel, line 382: Nice! Thanks for adding this to fuzzerutil.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+  auto condition_operand = branch_instruction->GetInOperand(0);
+  uint32_t first_true_block_id = branch_instruction->GetSingleWordInOperand(1);
+  uint32_t first_false_block_id = branch_instruction->GetSingleWordInOperand(2);
+
+  // The current header should unconditionally branch to the first block in the
+  // true branch, if there exists a true branch, and to the first block in the
+  // false branch if there is no true branch.
+  uint32_t after_header = first_true_block_id != convergence_block_id
+                              ? first_true_block_id
+                              : first_false_block_id;
```
Vasniktel, line 248: SPIR-V has a rule that all blocks should appear syntactically before any blocks they dominate. You've inserted all blocks from the false branch right before the blocks from the true branch. Thus, I think should branch unconditionally to the first false block if there is one or to the first true block otherwise. @stefanomil, what do you think?

Vasniktel: Or you could've inserted the blocks from the true branch first.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+  auto postdominator_analysis =
+      ir_context->GetPostDominatorAnalysis(enclosing_function);
+
+  // Check that this is a single-entry, single-exit region, by checking that the
+  // header dominates the convergence block and that the convergence block
+  // post-dominates the header.
+  if (!dominator_analysis->Dominates(header->id(), convergence_block_id) ||
+      !postdominator_analysis->Dominates(convergence_block_id, header->id())) {
+    return false;
+  }
```
Vasniktel, line 330: It should be OK but make sure to test that the transformation can't be applied to the unreachable selection construct.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+    uint32_t original_result_id = instruction->result_id();
+
+    // Set the result id of the instruction to a fresh id.
+    instruction->SetResultId(fresh_ids[3]);
+
+    // Add an OpUndef instruction, with the same type as the original
+    // instruction and a fresh id, to the new block.
+    alternative_block_temp->AddInstruction(MakeUnique<opt::Instruction>(
+        ir_context, SpvOpUndef, instruction->type_id(), fresh_ids[4],
+        opt::Instruction::OperandList{}));
```
Vasniktel, line 514: This is related to #3716. If `instruction` is an `OpFunctionCall` to the function that doesn't return anything, we might not be able to create an `OpUndef` of `OpTypeVoid`. This means that we won't be able to create an `OpPhi` as well.

I think this should be OK if we prohibit all usages of void ids as well as `OpUndef` of void type. Otherwise (if we only prohibit `OpUndef`), we might have some trouble with `TransformationAddSynonym` that can create `OpCopyObject` of void type.

@paulthomson for info.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+  auto else_block_id = (exec_if_cond_true ? alternative_block : execute_block)
+                           ->GetLabel()
+                           ->result_id();
+
+  // Add an OpSelectionMerge instruction to the block.
+  block->AddInstruction(MakeUnique<opt::Instruction>(
+      ir_context, SpvOpSelectionMerge, 0, 0,
+      opt::Instruction::OperandList{
+          {SPV_OPERAND_TYPE_ID, {merge_block->id()}},
+          {SPV_OPERAND_TYPE_SELECTION_CONTROL, {0}}}));
```
Vasniktel, line 558: Consider using an SpvSelectionControlMaskNone.

Vasniktel (changes requested):
This looks great. Code quality is amazing.

I left some review comments.

source/fuzz/transformation_flatten_conditional_branch.cpp
Vasniktel: Additionally, you could create an `OpUndef` in the global scope. This should reduce some amount of code.

paulthomson: That could work. I do wonder if keeping things local is more interesting in this case though. Hard to know for sure. I worry that spirv-fuzz might generally add too many uses of globals.

I guess if we see a function call with a void return type and if the void result is used (which seems unlikely) we could give up trying to do this transformation.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+            switch (instruction->opcode()) {
+              case SpvOpStore:
+              case SpvOpLoad:
+              case SpvOpFunctionCall:
+                problematic_instructions.push_back(instruction);
+                break;
+              default:
+                break;
+            }
+          });
```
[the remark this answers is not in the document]
stefanomil: Thanks for spotting this, I forgot to change it! I think using ``if (!InstructionHasNoSideEffects(...))`` would be fine, because we should be able to handle all the functions that return true when given to ``InstructionCanBeHandled`` and, if there is some instruction that cannot be handled, it will have already been spotted by IsApplicable.

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: I have inserted all blocks from the true branch right before the false branch (I first add the false branch right after the header and then the true branch right after the header). So it should branch unconditionally to the true branch, if it exists

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: What do you mean? Should it not be applicable even if a selection construct is unreachable?

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: (The comment saying otherwise was wrong)

source/fuzz/transformation_flatten_conditional_branch.cpp
Vasniktel: Maybe we could apply this transformation to an unreachable block but that's a bit tricky. The reason is that domination rules are undefined for them.

If you recall, block A dominates block B if all paths from the entry block to B go through A. In case of unreachable blocks, there is no path from the entry block.

Vasniktel: This implies that there is no rule that an instruction should dominate all its users.

stefanomil: I assume it doesn't get applied now, because nothing dominates anything, right?

Vasniktel: Yeah. It would still be good to add a test for it, though.

source/fuzz/fuzzer_util.cpp
```diff
@@ -1342,6 +1342,117 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
   return CanInsertOpcodeBeforeInstruction(opcode, &*it) ? &*it : nullptr;
 }
 
+bool InstructionHasNoSideEffects(opt::Instruction* instruction) {
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+
+  // |header_block_id| must refer to a block label.
+  {
+    auto label = ir_context->get_def_use_mgr()->GetDef(header_block_id);
+    if (!label || label->opcode() != SpvOpLabel) {
+      return false;
+    }
+  }
+
+  auto header_block = ir_context->cfg()->block(header_block_id);
```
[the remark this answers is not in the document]
stefanomil: Thanks, done

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+    return 2;
+  }
+}
+
+std::unordered_map<opt::Instruction*, std::vector<uint32_t>>
+TransformationFlattenConditionalBranch::GetInstructionsToFreshIdsMapping(
+    opt::IRContext* ir_context) const {
+  std::unordered_map<opt::Instruction*, std::vector<uint32_t>>
+      instructions_to_fresh_ids;
+  for (auto pair : message_.instruction_to_fresh_ids()) {
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+    std::vector<uint32_t> fresh_ids;
+    for (uint32_t id : pair.id()) {
+      fresh_ids.push_back(id);
+    }
```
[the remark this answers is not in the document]
stefanomil: I find this easier to understand

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+    // Check the overflow ids.
+    for (uint32_t id : message_.overflow_id()) {
+      if (!CheckIdIsFreshAndNotUsedByThisTransformation(id, ir_context,
+                                                        &used_fresh_ids)) {
+        return false;
+      }
+    }
+
+    // Check the ids in the map.
+    for (auto pair : instructions_to_fresh_ids) {
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+}
+
+protobufs::Transformation TransformationFlattenConditionalBranch::ToMessage()
+    const {
+  protobufs::Transformation result;
+  *result.mutable_flatten_conditional_branch() = message_;
+  return result;
+}
+
+bool TransformationFlattenConditionalBranch::ConditionalCanBeFlattened(
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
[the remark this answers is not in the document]
stefanomil: I have now updated this function

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,595 @@
...
+        // Get them from the map.
+        if (instructions_to_fresh_ids.count(instruction) != 0) {
+          fresh_ids = instructions_to_fresh_ids[instruction];
+        }
+
+        // Get the ones still needed from the overflow ids.
+        for (int still_needed = ids_needed - static_cast<int>(fresh_ids.size());
+             still_needed > 0; still_needed--) {
+          fresh_ids.push_back(message_.overflow_id(overflow_ids_used++));
+        }
```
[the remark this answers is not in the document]
stefanomil: You are right; I have changed the code to make it simpler using this fact

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: I added new tests

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: I have changed the code so that, whenever an instruction has side effects and returns a void type:
- is treated like an instruction that does not return anything if its result is not used in the module
- causes the transformation to be inapplicable otherwise

For now, the transformation still adds ``OpUndef`` locally. @paulthomson What advantages/disadvantages are there to have it global? Is there a transformation to add a global OpUndef?

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
paulthomson: Sounds good! I suggest keeping it local for now. Replacing it with a global undef is an easy transformation that could occur later.

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+    // Set the result id of the instruction to a fresh id.
+    instruction->SetResultId(fresh_ids[3]);
+
+    // Add an OpUndef instruction, with the same type as the original
+    // instruction and a fresh id, to the new block.
+    alternative_block_temp->AddInstruction(MakeUnique<opt::Instruction>(
+        ir_context, SpvOpUndef, instruction->type_id(), fresh_ids[4],
+        opt::Instruction::OperandList{}));
+
```
afd, line 501: Rather than creating an OpUndef here, I would prefer it if the transformation protobuf would contain an id of the right type that should be used.

The fuzzer pass can provide an irrelevant zero whenever possible, and an OpUndef when an irrelevant zero is not possible.

(We might create an OpUndef here as a last resort, in the case where we have to use overflow ids.)

source/fuzz/fuzzer_pass_flatten_conditional_branches.cpp
```diff
@@ -0,0 +1,93 @@
...
+    default;
+
+void FuzzerPassFlattenConditionalBranches::Apply() {
+  // Get all the selection headers that we want to flatten. We need to collect
+  // all of them first, because, since we are changing the structure of the
+  // module, it's not safe to modify them while iterating.
+  std::vector<opt::BasicBlock*> selection_headers;
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      // Randomly decide whether to consider this block.
```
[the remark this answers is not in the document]
stefanomil: Good idea, I will do this once #3753 is merged, and will prioritise headers with a higher depth, so the transformation will be applicable to as many conditionals as possible

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1080,6 +1093,28 @@ message TransformationEquationInstruction {
 
 }
 
+message TransformationFlattenConditionalBranch {
+
+  // A transformation that takes a selection construct with a header
+  // containing an OpBranchConditional instruction and flattens it.
```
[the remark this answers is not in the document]
stefanomil: I have added an explanation

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -49,6 +49,18 @@ message InstructionDescriptor {
...
+  // Useful to represent a mapping from instructions to a list of ids.
+
+  // An instruction descriptor to identify an instruction.
+  InstructionDescriptor instruction_descriptor = 1;
+
+  // A list of ids.
+  repeated uint32 id = 2;
+
+}
+
```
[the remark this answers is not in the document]
stefanomil: I have renamed it, added an explanation and changed it so that it now has a named field for each possible id that is needed, instead of a list

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1080,6 +1093,28 @@ message TransformationEquationInstruction {
...
+  // A transformation that takes a selection construct with a header
+  // containing an OpBranchConditional instruction and flattens it.
+
+  // The label id of the header block
+  uint32 header_block_id = 1;
+
+  // A list of pairs mapping instruction descriptors to fresh ids.
+  // Necessary for store, load and function call instructions.
+  // This helps preserve the stability of the transformation, so
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.h
```diff
@@ -0,0 +1,114 @@
...
+  // Returns the number of fresh ids needed to enclose the given instruction
+  // inside a conditional. That is:
+  // - 2 if the instruction does not have a result id or has a void result id,
+  //   needed for 2 new blocks
+  // - 5 if the instruction has a non-void result id: 3 for new blocks, 1 for a
+  //   new OpUndef instruction, 1 for the instruction itself
+  // Assumes that if the instruction has a non-void result type, its result id
+  // is not used in the module.
+  static uint32_t NumOfFreshIdsNeededByInstruction(
+      opt::IRContext* ir_context, const opt::Instruction& instruction);
```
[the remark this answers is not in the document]
stefanomil: I have changed this method to return a boolean, depending on whether enclosing the instruction would require to have a placeholder instruction, thereby also changing the name

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: Could the transformation, instead, try to find a suitable zero constant? The fuzzer pass would ensure that it is in the module. This way, we minimise the amount of times that we need to use OpUndef (only for types that cannot be zero)

source/fuzz/fuzzer_pass_flatten_conditional_branches.cpp
```diff
@@ -0,0 +1,93 @@
...
+          NumOfFreshIdsNeededByInstruction(GetIRContext(), *instruction);
+
+      instructions_to_fresh_ids.push_back(
+          {MakeInstructionDescriptor(GetIRContext(), instruction),
+           GetFuzzerContext()->GetFreshIds(num_fresh_ids_needed)});
+    }
+
+    // Add 10 overflow ids to account for possible changes in the module.
+    auto overflow_ids = GetFuzzerContext()->GetFreshIds(10);
+
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,604 @@
...
+  while (ir_context->cfg()->preds(convergence_block_id).size() == 1) {
+    convergence_block_id = ir_context->cfg()->preds(convergence_block_id)[0];
+  }
+
+  // Get the mapping from instructions to fresh ids.
+  auto instructions_to_fresh_ids = GetInstructionsToFreshIdsMapping(ir_context);
+
+  // Keep track of the number of overflow ids used.
+  uint32_t overflow_ids_used = 0;
+
```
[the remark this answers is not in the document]
stefanomil: I have changed the code to use the new OverflowIdSource

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,604 @@
...
+          fresh_ids = instructions_to_fresh_ids[instruction];
+        }
+
+        // If we could not get it from the map, use overflow ids.
+        if (fresh_ids.empty()) {
+          for (uint32_t i = 0; i < ids_needed; i++) {
+            fresh_ids.push_back(message_.overflow_id(overflow_ids_used++));
+          }
+        }
+
```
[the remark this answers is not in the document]
stefanomil: Fixed

source/fuzz/fuzzer_pass_flatten_conditional_branches.cpp
```diff
@@ -0,0 +1,93 @@
...
+        instructions_to_fresh_ids;
+
+    for (auto instruction : instructions_that_need_ids) {
+      uint32_t num_fresh_ids_needed = TransformationFlattenConditionalBranch::
+          NumOfFreshIdsNeededByInstruction(GetIRContext(), *instruction);
+
+      instructions_to_fresh_ids.push_back(
+          {MakeInstructionDescriptor(GetIRContext(), instruction),
+           GetFuzzerContext()->GetFreshIds(num_fresh_ids_needed)});
+    }
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/fuzzer_pass_flatten_conditional_branches.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    // Do not consider this header if the conditional cannot be flattened.
+    if (!TransformationFlattenConditionalBranch::ConditionalCanBeFlattened(
+            GetIRContext(), header, &instructions_that_need_ids)) {
+      continue;
+    }
+
+    // Generate entries (instruction descriptor, fresh ids) for all the
+    // instructions that require fresh ids.
+    std::vector<
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.h
```diff
@@ -0,0 +1,114 @@
...
+  // they are only executed when they should.
+  void Apply(opt::IRContext* ir_context,
+             TransformationContext* transformation_context) const override;
+
+  protobufs::Transformation ToMessage() const override;
+
+  // Returns true if the conditional headed by |header| can be flattened,
+  // according to the conditions of the IsApplicable method, assuming that
+  // enough fresh ids are given. In this case, it fills the
+  // |instructions_that_need_ids| set with all the instructions that would
```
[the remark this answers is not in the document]
stefanomil: Fixed

source/fuzz/transformation_flatten_conditional_branch.h
```diff
@@ -0,0 +1,114 @@
...
+      std::set<opt::Instruction*>* instructions_that_need_ids);
+
+  // Returns the number of fresh ids needed to enclose the given instruction
+  // inside a conditional. That is:
+  // - 2 if the instruction does not have a result id or has a void result id,
+  //   needed for 2 new blocks
+  // - 5 if the instruction has a non-void result id: 3 for new blocks, 1 for a
+  //   new OpUndef instruction, 1 for the instruction itself
+  // Assumes that if the instruction has a non-void result type, its result id
+  // is not used in the module.
```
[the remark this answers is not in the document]
stefanomil: Yes. I have added an assertion to check this and specified it in the comment

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,604 @@
...
+  // Get the mapping from instructions to fresh ids.
+  auto instructions_to_fresh_ids = GetInstructionsToFreshIdsMapping(ir_context);
+
+  // Keep track of the number of overflow ids used.
+  uint32_t overflow_ids_used = 0;
+
+  auto branch_instruction = header_block->terminator();
+
+  opt::BasicBlock* last_true_block = nullptr;
+
```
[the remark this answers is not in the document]
stefanomil: I have added a comment

source/fuzz/fuzzer_pass_flatten_conditional_branches.cpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  // Apply the transformation to the headers which can be flattened.
+  for (auto header : selection_headers) {
+    // Make a set to keep track of the instructions that need fresh ids.
+    std::set<opt::Instruction*> instructions_that_need_ids;
+
+    // Do not consider this header if the conditional cannot be flattened.
+    if (!TransformationFlattenConditionalBranch::ConditionalCanBeFlattened(
```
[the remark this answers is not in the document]
stefanomil: I renamed the function and explained in the comment

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,604 @@
...
+    auto block = header_block;
+    // Get the id of the first block in this branch.
+    uint32_t block_id = branch_instruction->GetSingleWordInOperand(branch);
+
+    // Consider all blocks in the branch until the convergence block is reached.
+    while (block_id != convergence_block_id) {
+      // Move the block to right after the previous one.
+      block->GetParent()->MoveBasicBlockToAfter(block_id, block);
+
```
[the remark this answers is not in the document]
stefanomil: I have changed the names to "current_block" and "next_block_id"

source/fuzz/fuzzer_util.cpp
```diff
@@ -1352,6 +1352,118 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
   return CanInsertOpcodeBeforeInstruction(opcode, &*it) ? &*it : nullptr;
 }
 
+bool InstructionHasNoSideEffects(const opt::Instruction& instruction) {
+  switch (instruction.opcode()) {
```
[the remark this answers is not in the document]
stefanomil: I have added a TODO(...) comment and opened a related issue

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,604 @@
...
+  opt::BasicBlock* last_true_block = nullptr;
+
+  // Adjust the conditional branches by enclosing problematic instructions
+  // within conditionals and get references to the last block in each branch.
+  for (int branch = 2; branch >= 1; branch--) {
+    // branch = 1 corresponds to the true branch, branch = 2 corresponds to the
+    // false branch. Consider the false branch first so that the true branch is
+    // laid out right after the header.
+
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,604 @@
...
+  uint32_t overflow_ids_used = 0;
+
+  auto branch_instruction = header_block->terminator();
+
+  opt::BasicBlock* last_true_block = nullptr;
+
+  // Adjust the conditional branches by enclosing problematic instructions
+  // within conditionals and get references to the last block in each branch.
+  for (int branch = 2; branch >= 1; branch--) {
+    // branch = 1 corresponds to the true branch, branch = 2 corresponds to the
```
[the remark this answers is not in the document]
stefanomil: I agree. Done (and then changed to allow choosing the order in which the two branches are laid out)

source/fuzz/transformation_flatten_conditional_branch.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+void TransformationFlattenConditionalBranch::Apply(
+    opt::IRContext* ir_context,
+    TransformationContext* transformation_context) const {
+  uint32_t header_block_id = message_.header_block_id();
+  auto header_block = ir_context->cfg()->block(header_block_id);
+
+  // Find the first block where flow converges (it is not necessarily the merge
+  // block).
+  uint32_t convergence_block_id = header_block->MergeBlockId();
```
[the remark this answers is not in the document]
stefanomil: I have commented on (1) - this is necessary because we cannot duplicate common blocks, and changed the algorithm so that the first point where flow diverges is chosen, as I have found out that it is possible to have an inner construct after flow has converged but before the merge block. So I don't think that I need to say anything about (2) because that case is now avoided

source/fuzz/transformation_flatten_conditional_branch.cpp
stefanomil: The fuzzer pass will provide a suitable id. In the case where we have to use overflow ids, the transformation will try to find a suitable zero constant and it will use an OpUndef only if it fails to do so

stefanomil: Done

source/fuzz/fuzzer_pass_flatten_conditional_branches.cpp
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -306,6 +306,81 @@ message AccessChainClampingInfo {
 
 }
 
+message InstToIdsForEnclosing {
```
afd, line 309: I really like the way you have factored this out, and the comments are great.  However I don't like this name.

I'm finding it hard to think of a good name, but I think it should be of the form SomethingInfo (to mirror AccessChainClampingInfo and LoopLimiterInfo).

How about SideEffectEnclosureInfo, or SideEffectWrapperInfo?

source/fuzz/fuzzer_pass_flatten_conditional_branches.cpp
```diff
@@ -0,0 +1,144 @@
...
+      if (TransformationFlattenConditionalBranch::InstructionNeedsPlaceholder(
+              GetIRContext(), *instruction)) {
+        info.actual_result_id = GetFuzzerContext()->GetFreshId();
+        info.alternative_block_id = GetFuzzerContext()->GetFreshId();
+        info.placeholder_result_id = GetFuzzerContext()->GetFreshId();
+
+        // The id will be a zero constant if the type allows it, and an OpUndef
+        // otherwise. We want to avoid using OpUndef if possible because they
+        // limit the fuzzer's ability of applying further transformations,
+        // having undefined behaviour.
```
afd, line 103: it's not because they limit the fuzzer's ability to apply further transformations; it's simply because we want to avoid undefined behaviour in the module as much as possible.

source/fuzz/fuzzer_pass_flatten_conditional_branches.h
```diff
@@ -0,0 +1,58 @@
...
+  // not as deep.
+  class LessIfNestedMoreDeeply {
+   public:
+    explicit LessIfNestedMoreDeeply(opt::IRContext* ir_context)
+        : ir_context_(ir_context) {}
+
+    bool operator()(const opt::BasicBlock* bb1, opt::BasicBlock* bb2) const {
+      return NestingDepth(ir_context_, bb1->id()) >
+             NestingDepth(ir_context_, bb2->id());
+    }
```
afd, line 50: Could this be in the private section?

source/fuzz/fuzzer_pass_flatten_conditional_branches.h
```diff
@@ -0,0 +1,58 @@
...
+      protobufs::TransformationSequence* transformations);
+
+  ~FuzzerPassFlattenConditionalBranches() override;
+
+  void Apply() override;
+
+  // Returns the nesting depth of the given block, i.e. the number of constructs
+  // containing it. Headers and merge blocks of constructs are not considered
+  // part of a construct.
+  static uint32_t NestingDepth(opt::IRContext* ir_context, uint32_t block_id);
```
afd, line 37: Could this be in the private section?

afd: OK, I see it is public because you have tests for it.

I think it would be best for this to live in structured cfg analysis, as it's generally useful.

You could either do a PR there, first, or you could open an issue and put a TODO comment here, get this PR merged, and then do a refactoring.

The tests should migrate to the tests for structured CFG analysis.

source/fuzz/fuzzer_pass_flatten_conditional_branches.h
afd: Again, I see it's public because you have tests for it.

I can imagine this being a generally useful comparator.

I suggest you either:

(1) Put it in its own header file (I guess there won't be a corresponding C++ file), and have an associated test file

or

(2) Put it into fuzzer_util, and make a test file for fuzzer_util with your tests there

I have a mild preference for (1) as I quite like the Java approach of every class being in its own file.

test/fuzz/fuzzer_pass_flatten_conditional_branches_test.cpp
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) 2020 Google LLC
```
afd, line 1: My suggestions for moving out the comparator and the helper function would mean that all the content of this file would migrate away elsewhere.

source/fuzz/transformation_flatten_conditional_branch.h
```diff
@@ -0,0 +1,145 @@
...
+#define SOURCE_FUZZ_TRANSFORMATION_FLATTEN_CONDITIONAL_BRANCH_H
+
+#include "source/fuzz/transformation.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// A structure declaration for all the fresh ids needed to enclose a
+// side-effecting instruction inside a conditional.
+struct IdsForEnclosingInst {
```
afd, line 25: I suggest you do away with this class altogether and instead do what I've done for AccessChainClampingInfo:

- In the fuzzer pass it's no trouble to build up the protobuf object directly
- In the tests it is a pain to have to do this repeatedly per test, but you can make a factory method - see protobufs::AccessChainClampingInfo MakeAccessClampingInfo(...) in transformation_add_function_test.

I like to avoid classes that mirror the protobuf messages unless there's a really good reason.

afd (changes requested):
Thanks for making the changes!  This is looking excellent.  My probably-final feedback is to do with where the new comparator and nesting depth method should live, what you should call the new helper protobuf message, and whether you should have a helper class that mirrors the protobuf message.

Happy to discuss these things live if that would help, of course.

source/fuzz/protobufs/spvtoolsfuzz.proto
stefanomil: I renamed it to SideEffectWrapperInfo

source/fuzz/fuzzer_pass_flatten_conditional_branches.cpp
stefanomil: I changed the comment

source/fuzz/transformation_flatten_conditional_branch.h
stefanomil: Done

source/fuzz/fuzzer_pass_flatten_conditional_branches.h
stefanomil: I have opened PR #3771 .

stefanomil:
@afd There is a case that I previously had not considered: the first block of each branch could itself have OpPhi instruction, which needs to be updated in the second branch. Also, if the header branches to the same block (there is not an actual conditional), the OpPhi instructions in it do not need to be updated. Please look at the changes I made about this.

source/fuzz/fuzzer_pass_flatten_conditional_branches.h
stefanomil: I have put it in its own header file

source/fuzz/transformation_flatten_conditional_branch.h
```diff
@@ -0,0 +1,113 @@
...
+
+class TransformationFlattenConditionalBranch : public Transformation {
+ public:
+  explicit TransformationFlattenConditionalBranch(
+      const protobufs::TransformationFlattenConditionalBranch& message);
+
+  TransformationFlattenConditionalBranch(
+      uint32_t header_block_id, bool true_branch_first,
+      const std::vector<protobufs::SideEffectWrapperInfo>&
+          side_effect_wrappers_info = {});
```
afd, line 31: I'm not keen on default parameters unless there is a really strong case for them, so can you make this non-default, and pass in {} for the cases where no wrappers are required?

I'm guessing that in your case this merely makes the test code a tad easier to write, and I don't think that's a strong case.

afd (changes requested):
Thanks!  Some small things and then I'm happy to merge.

source/fuzz/comparator_deep_blocks_first.h
```diff
@@ -0,0 +1,47 @@
...
+
+// Comparator for blocks, comparing them based on how deep they are nested
+// inside selection or loop constructs. Deeper blocks are considered less than
+// ones that are not as deep.
+class ComparatorDeepBlocksFirst {
+ public:
+  explicit ComparatorDeepBlocksFirst(opt::IRContext* ir_context)
+      : ir_context_(ir_context) {}
+
+  bool operator()(uint32_t bb1, uint32_t bb2) const {
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/comparator_deep_blocks_first.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef SOURCE_FUZZ_COMPARATOR_BLOCKS_DEEP_FIRST_H_
+#define SOURCE_FUZZ_COMPARATOR_BLOCKS_DEEP_FIRST_H_
+
+#include "source/opt/ir_context.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// Comparator for blocks, comparing them based on how deep they are nested
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_flatten_conditional_branch.h
stefanomil: I have made it non-default and updated the tests

afd (approved):
Thanks for this contribution!

PR #3668, opened 2020-08-07
spirv-fuzz: Fuzzer pass to randomly apply loop preheaders

### LINKED ISSUE #3621: spirv-fuzz: Fuzzer pass to randomly apply pre-headers to loops
There should be a pass that randomly selects loop headers which do
not have a preheader and uses TransformationAddLoopPreheader to add
one.

stefanomil:
This PR introduces FuzzerPassAddLoopPreheaders, which:
- Finds existing loop headers
- If they have no simple preheader (where simple means that the
  preheader itself is not a loop header), randomly decides whether
  to add one.

Fixes #3621 .

### DIFF
--- a/source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
+++ b/source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
@@ -32,7 +32,7 @@ FuzzerPassAddLoopPreheaders::~FuzzerPassAddLoopPreheaders() = default;
 void FuzzerPassAddLoopPreheaders::Apply() {
   for (auto& function : *GetIRContext()->module()) {
     // Keep track of all the loop headers we want to add a preheader to.
-    std::vector<opt::BasicBlock*> loop_headers_to_consider;
+    std::vector<uint32_t> loop_headers_to_consider;
     for (auto& block : function) {
       // We only care about loop headers.
       if (!block.IsLoopHeader()) {
@@ -47,18 +47,17 @@ void FuzzerPassAddLoopPreheaders::Apply() {
 
       // We exclude loop headers with just one predecessor (the back-edge block)
       // because they are unreachable.
-      if (GetIRContext()->cfg()->preds(block.GetLabel()->result_id()).size() <
-          2) {
+      if (GetIRContext()->cfg()->preds(block.id()).size() < 2) {
         continue;
       }
 
-      loop_headers_to_consider.push_back(&block);
+      loop_headers_to_consider.push_back(block.id());
     }
 
-    for (auto& header : loop_headers_to_consider) {
+    for (uint32_t header_id : loop_headers_to_consider) {
       // If not already present, add a preheader which is not also a loop
       // header.
-      GetOrCreateSimpleLoopPreheader(header->GetLabel()->result_id());
+      GetOrCreateSimpleLoopPreheader(header_id);
     }
   }
 }
--- a/source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
+++ b/source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
@@ -32,7 +32,7 @@ FuzzerPassAddLoopPreheaders::~FuzzerPassAddLoopPreheaders() = default;
 void FuzzerPassAddLoopPreheaders::Apply() {
   for (auto& function : *GetIRContext()->module()) {
     // Keep track of all the loop headers we want to add a preheader to.
-    std::vector<uint32_t> loop_headers_to_consider;
+    std::vector<uint32_t> loop_header_ids_to_consider;
     for (auto& block : function) {
       // We only care about loop headers.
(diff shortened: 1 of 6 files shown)


source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
```diff
@@ -0,0 +1,67 @@
...
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassAddLoopPreheaders::~FuzzerPassAddLoopPreheaders() = default;
+
+void FuzzerPassAddLoopPreheaders::Apply() {
+  for (auto& function : *GetIRContext()->module()) {
+    // Keep track of all the loop headers we want to add a preheader to.
+    std::vector<opt::BasicBlock*> loop_headers_to_consider;
```
paulthomson, line 35: Nit: since you just want the block id, you could just use a vector of uint32_t (loop_header_block_ids_to_consider).

source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
```diff
@@ -0,0 +1,67 @@
...
+    for (auto& header : loop_headers_to_consider) {
+      // If not already present, add a preheader which is not also a loop
+      // header.
+      GetOrCreateSimpleLoopPreheader(header->GetLabel()->result_id());
+    }
+  }
+}
+
+}  // namespace fuzz
+}  // namespace spvtools
```
paulthomson, line 67: Please end file with a new line. In CLion, there is a setting: "Ensure an empty line at the end of a file on Save". "Strip trailing spaces on Save for:" is also good.

source/fuzz/fuzzer_pass_add_loop_preheaders.h
```diff
@@ -0,0 +1,43 @@
...
+#include "source/fuzz/fuzzer_pass.h"
+
+namespace spvtools {
+namespace fuzz {
+
+// A fuzzer pass that randomly adds loop preheaders to headers that do not have
+// one that satisfies all of the following conditions:
+// - it is the only out-of-loop predecessor of the header
+// - it branches unconditionally to the header
+// - it is not a loop header itself
```
paulthomson, line 27: I feel that we have made up the term "loop preheader", and a loop preheader must satisfy all of these conditions. Thus, you cannot have a loop preheader that does not satisfy these conditions. Thus, I would say:


A fuzzer pass that randomly adds loop preheaders to loops that do not already have a loop preheader. A loop preheader is simply a predecessor block of the loop header that:
- is the only out-of-loop predecessor of the header
- branches unconditionally to the header
- is not a loop header itself

source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
```diff
@@ -0,0 +1,67 @@
...
+      // Randomly decide whether to consider this header.
+      if (!GetFuzzerContext()->ChoosePercentage(
+              GetFuzzerContext()->GetChanceOfAddingLoopPreheader())) {
+        continue;
+      }
+
+      // We exclude loop headers with just one predecessor (the back-edge block)
+      // because they are unreachable.
+      if (GetIRContext()->cfg()->preds(block.GetLabel()->result_id()).size() <
+          2) {
```
paulthomson, line 51: I think you can just do `block.id()`

source/fuzz/fuzzer_pass_add_loop_preheaders.h
stefanomil: The preheader block definition given [here](https://llvm.org/docs/LoopTerminology.html) only imposes the first two conditions, could this cause any confusion?

source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
stefanomil: Yes, I agree that this is better. I called it "loop_header_ids_to_consider"

source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
stefanomil: Thanks! Fixed

source/fuzz/fuzzer_pass_add_loop_preheaders.h
stefanomil: I changed it to:

A fuzzer pass that randomly adds simple loop preheaders to loops that do not have one. A simple loop preheader is a block that:
- is the only out-of-loop predecessor of the header
- branches unconditionally to the header
- is not a loop header itself

source/fuzz/fuzzer_pass_add_loop_preheaders.cpp
stefanomil: Yes!

PR #3670, opened 2020-08-07
spirv-fuzz: Check integer and float width capabilities

### LINKED ISSUE #3669: spirv-fuzz: Fuzzed shader invalidation due missing Float64 capability
`spirv-val` invalidated a fuzzed shader because a 64-bit float type was declared but the `Float64` capability was missing.
```
error: line 67: Using a 64-bit floating point type requires the Float64 capability.
  %double = OpTypeFloat 64
```
Debugging `spirv-fuzz`, I found that `TransformationAddTypeFloat` adds 16-bit and 64-bit float types without adding the corresponding `Float16` and `Float64` capabilities.

andreperezmaselco:
`TransformationAddTypeFloat` and `TransformationAddTypeInt` did not check whether the required capabilities were present when adding 16-bit, 64-bit, and 8-bit types.

This change adds these checks in the `IsApplicable` method of each transformation.

Fixes #3669.

### DIFF
--- a/source/fuzz/transformation_add_type_float.cpp
+++ b/source/fuzz/transformation_add_type_float.cpp
@@ -36,6 +36,16 @@ bool TransformationAddTypeFloat::IsApplicable(
     return false;
   }
 
+  // If the float type width is 16 or 64, then the corresponding Float16 or
+  // Float64 capability must be present.
+  if (message_.width() == 16) {
+    assert(ir_context->get_feature_mgr()->HasCapability(SpvCapabilityFloat16) &&
+           "Missing Float16 capability.");
+  } else if (message_.width() == 64) {
+    assert(ir_context->get_feature_mgr()->HasCapability(SpvCapabilityFloat64) &&
+           "Missing Float64 capability.");
+  }
+
   // Applicable if there is no float type with this width already declared in
   // the module.
   return fuzzerutil::MaybeGetFloatType(ir_context, message_.width()) == 0;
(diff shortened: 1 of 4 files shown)


Vasniktel:
It would be good to add simple tests.

source/fuzz/transformation_add_type_float.cpp
```diff
@@ -36,6 +36,15 @@ bool TransformationAddTypeFloat::IsApplicable(
...
   }
 
+  // If the float type width is 16 or 64, then the corresponding Float16 or
+  // Float64 capability must be present.
+  if ((message_.width() == 16 &&
+       !ir_context->get_feature_mgr()->HasCapability(SpvCapabilityFloat16)) ||
+      (message_.width() == 64 &&
+       !ir_context->get_feature_mgr()->HasCapability(SpvCapabilityFloat64))) {
+    return false;
+  }
```
paulthomson, line 46: This is really nice. Sorry to not have suggested this sooner, but could you do a switch statement on the width, include cases for all expected widths (16, 32, 64, (and 8 for int)) where the 32 case just has a break with a comment saying something like `// No capabilities needed.` And then in the default case do:

```
assert(false && "Unexpected float type width");
return false;
```

Or something like that. And do the same for TransformationAddTypeInt.

andreperezmaselco: Done.

PR #3672, opened 2020-08-07
spirv-fuzz: consider additional access chain instructions

### LINKED ISSUE #3671: spirv-fuzz: Invalidation when indexing into a structure
`spirv-val` invalidated a fuzzed shader because the result id of an `OpConstantNull` instruction was being used to index into a structure. According to SPIR-V specification, when indexing into a structure, each index must be the result id of an `OpConstant` instruction.
```
error: line 79: The <id> passed to OpInBoundsAccessChain to index into a structure must be an OpConstant.
  %399 = OpConstantNull %uint
```
Debugging `spirv-fuzz`, I found that `TransformationReplaceIdWithSynonym` checks only if the index instruction is `OpAccessChain`, while the same is true for `OpInBoundsAccessChain` and `OpPtrAccessChain`.

andreperezmaselco:
`TransformationReplaceIdWithSynonym` is careful to avoid replacing id uses that index into a struct with synonyms because the indices must only be `OpConstant` instructions. However, the check only considered `OpAccessChain` instructions, even though the same restriction applies to `OpInBoundsAccessChain`, `OpPtrAccessChain`, etc. 

This change extends the check to include all access chain instructions.

Fixes #3671.

### DIFF
--- a/source/fuzz/transformation_replace_id_with_synonym.cpp
+++ b/source/fuzz/transformation_replace_id_with_synonym.cpp
@@ -109,7 +109,7 @@ protobufs::Transformation TransformationReplaceIdWithSynonym::ToMessage()
 bool TransformationReplaceIdWithSynonym::UseCanBeReplacedWithSynonym(
     opt::IRContext* ir_context, opt::Instruction* use_instruction,
     uint32_t use_in_operand_index) {
-  if (use_instruction->opcode() == SpvOpAccessChain &&
+  if (spvOpcodeIsAccessChain(use_instruction->opcode()) &&
       use_in_operand_index > 0) {
     // This is an access chain index.  If the (sub-)object being accessed by the
     // given index has struct type then we cannot replace the use with a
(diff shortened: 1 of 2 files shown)


source/opcode.cpp
```diff
@@ -719,3 +719,14 @@ std::vector<uint32_t> spvOpcodeMemorySemanticsOperandIndices(SpvOp opcode) {
...
+bool spvOpcodeIsAccessChain(SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpAccessChain:
+    case SpvOpInBoundsAccessChain:
+    case SpvOpPtrAccessChain:
+      return true;
+    default:
+      return false;
+  }
+}
```
Vasniktel, line 732: It might be better to make it a separate PR since it changes some code outside of the spirv-fuzz tool. I'm not sure whether this is critical, though. @paulthomson, what do you think?

Vasniktel:
It is also good to add some tests to cover the bug.

source/opcode.cpp
paulthomson: I agree. We prefer to let the spirv-opt folks approve changes in their code.

paulthomson: But it looks like a good change so I am guess they will approve.

andreperezmaselco: Done.

PR #3681, opened 2020-08-11
spirv-fuzz: Transformation to convert OpSelect to conditional branch

stefanomil:
This transformation takes an OpSelect instruction and replaces it with
a conditional branch, selecting the correct value using an OpPhi
instruction.

Fixes part of the issue #3544 .

### DIFF
--- a/source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.cpp
+++ b/source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.cpp
@@ -15,7 +15,9 @@
 #include "source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.h"
 
 #include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/instruction_descriptor.h"
 #include "source/fuzz/transformation_replace_opselect_with_conditional_branch.h"
+#include "source/fuzz/transformation_split_block.h"
 
 namespace spvtools {
 namespace fuzz {
@@ -42,8 +44,11 @@ void FuzzerPassReplaceOpSelectsWithConditionalBranches::Apply() {
   for (auto& function : *GetIRContext()->module()) {
     for (auto& block : function) {
       // We cannot split loop headers, so we don't need to consider instructions
-      // in loop headers.
-      if (block.IsLoopHeader()) {
+      // in loop headers that are also merge blocks (since they would need to be
+      // split).
+      if (block.IsLoopHeader() &&
+          GetIRContext()->GetStructuredCFGAnalysis()->IsMergeBlock(
+              block.id())) {
         continue;
       }
 
@@ -58,6 +63,14 @@ void FuzzerPassReplaceOpSelectsWithConditionalBranches::Apply() {
                 GetFuzzerContext()
                     ->GetChanceOfReplacingOpselectWithConditionalBranch())) {
           continue;
+        }
+
+        // If the block is a loop header and we need to split it, the
+        // transformation cannot be applied because loop headers cannot be
+        // split. We can break out of this loop because the transformation can
+        // only be applied to at most the first instruction in a loop header.
+        if (block.IsLoopHeader() && InstructionNeedsSplitBefore(&instruction)) {
+          break;
         }
 
         // If the instruction separates an OpSampledImage from its use, the
@@ -75,12 +88,74 @@ void FuzzerPassReplaceOpSelectsWithConditionalBranches::Apply() {
     }
   }
 
-  // Apply the transformations.
+  // Apply the transformations, splitting the blocks containing the
+  // instructions, if necessary.
   for (uint32_t instruction_id : replaceable_opselect_instruction_ids) {
+    auto instruction =
+        GetIRContext()->get_def_use_mgr()->GetDef(instruction_id);
+
+    // If the instruction requires the block containing it to be split before
+    // it, split the block.
+    if (InstructionNeedsSplitBefore(instruction)) {
+      ApplyTransformation(TransformationSplitBlock(
+          MakeInstructionDescriptor(GetIRContext(), instruction),
+          GetFuzzerContext()->GetFreshId()));
+    }
+
+    // Decide whether to have two branches or just one.
+    bool two_branches = GetFuzzerContext()->ChoosePercentage(
+        GetFuzzerContext()
+            ->GetChanceOfAddingBothBranchesWhenReplacingOpSelect());
+
+    // If there will be only one branch, decide whether it will be the true
+    // branch or the false branch.
+    bool true_branch_id_zero =
+        !two_branches &&
+        GetFuzzerContext()->ChoosePercentage(
+            GetFuzzerContext()
+                ->GetChanceOfAddingTrueBranchWhenReplacingOpSelect());
+    bool false_branch_id_zero = !two_branches && !true_branch_id_zero;
+
+    uint32_t true_branch_id =
+        true_branch_id_zero ? 0 : GetFuzzerContext()->GetFreshId();
+    uint32_t false_branch_id =
+        false_branch_id_zero ? 0 : GetFuzzerContext()->GetFreshId();
+
     ApplyTransformation(TransformationReplaceOpSelectWithConditionalBranch(
-        instruction_id, GetFuzzerContext()->GetFreshId(),
-        GetFuzzerContext()->GetFreshId()));
+        instruction_id, true_branch_id, false_branch_id));
   }
+}
+
+bool FuzzerPassReplaceOpSelectsWithConditionalBranches::
+    InstructionNeedsSplitBefore(opt::Instruction* instruction) {
+  assert(instruction && instruction->opcode() == SpvOpSelect &&
+         "The instruction must be OpSelect.");
+
+  auto block = GetIRContext()->get_instr_block(instruction);
+  assert(block && "The instruction must be contained in a block.");
+
+  // We need to split the block if the instruction is not the first in its
+  // block.
+  if (instruction->unique_id() != block->begin()->unique_id()) {
+    return true;
+  }
+
+  // We need to split the block if it is a merge block.
+  if (GetIRContext()->GetStructuredCFGAnalysis()->IsMergeBlock(block->id())) {
+    return true;
+  }
+
+  // We need to split the block if it has more than one predecessor.
+  if (GetIRContext()->cfg()->preds(block->id()).size() != 1) {
+    return true;
+  }
+
+  // We need to split the block if its predecessor is a header or it does not
+  // branch unconditionally to the block.
+  auto predecessor = GetIRContext()->get_instr_block(
+      GetIRContext()->cfg()->preds(block->id())[0]);
+  return predecessor->MergeBlockIdIfAny() ||
+         predecessor->terminator()->opcode() != SpvOpBranch;
 }
 
 }  // namespace fuzz
--- a/source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.cpp
+++ b/source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.cpp
@@ -12,4 +12,76 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "fuzzer_pass_replace_opselects_with_conditional_branches.h"
+#include "source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.h"
+
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/transformation_replace_opselect_with_conditional_branch.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassReplaceOpselectsWithConditionalBranches::
+    FuzzerPassReplaceOpselectsWithConditionalBranches(
+        opt::IRContext* ir_context,
+        TransformationContext* transformation_context,
+        FuzzerContext* fuzzer_context,
+        protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassReplaceOpselectsWithConditionalBranches::
+    ~FuzzerPassReplaceOpselectsWithConditionalBranches() = default;
+
+void FuzzerPassReplaceOpselectsWithConditionalBranches::Apply() {
+  // Keep track of the instructions that we want to replace. We need to collect
+  // them in a vector, since it's not safe to modify the module while iterating
+  // over it.
+  std::vector<uint32_t> replaceable_opselect_instruction_ids;
+
+  // Loop over all the instructions in the module.
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      // We cannot split loop headers, so we don't need to consider instructions
+      // in loop headers.
+      if (block.IsLoopHeader()) {
+        continue;
+      }
+
+      for (auto& instruction : block) {
+        // We only care about OpSelect instructions.
+        if (instruction.opcode() != SpvOpSelect) {
+          continue;
+        }
+
+        // Randomly choose whether to consider this instruction for replacement.
+        if (!GetFuzzerContext()->ChoosePercentage(
+                GetFuzzerContext()
+                    ->GetChanceOfReplacingOpselectWithConditionalBranch())) {
+          continue;
+        }
+
+        // If the instruction separates an OpSampledImage from its use, the
+        // block cannot be split around it and the instruction cannot be
+        // replaced.
+        if (fuzzerutil::
+                SplitBeforeInstructionSeparatesOpSampledImageDefinitionFromUse(
+                    &block, &instruction)) {
+          continue;
+        }
+
+        // We can apply the transformation to this instruction.
+        replaceable_opselect_instruction_ids.push_back(instruction.result_id());
+      }
+    }
+  }
+
+  // Apply the transformations.
+  for (uint32_t instruction_id : replaceable_opselect_instruction_ids) {
+    ApplyTransformation(TransformationReplaceOpselectWithConditionalBranch(
+        instruction_id,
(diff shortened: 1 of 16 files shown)


source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1454,6 +1455,18 @@ message TransformationReplaceLoadStoreWithCopyMemory {
...
+message TransformationReplaceOpSelectWithConditionalBranch {
+
+  // A transformation that takes an OpSelect instruction and replaces
+  // it with a conditional branch and an OpPhi instruction.
+
+  // The result id of the OpSelect instruction
+  uint32 select_id = 1;
+
+  // Fresh ids for the new blocks
+  UInt32Pair new_block_ids = 2;
```
Vasniktel, line 1467: Not a strict requirement but it would probably be better to use two separate fields. Something like `true_block_fresh_id` and `merge_block_fresh_id`.

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  // Check that the new block ids are fresh and distinct.
+  std::set<uint32_t> used_ids;
+  for (uint32_t id :
+       {message_.new_block_ids().first(), message_.new_block_ids().second()}) {
+    if (!CheckIdIsFreshAndNotUsedByThisTransformation(id, ir_context,
+                                                      &used_ids)) {
+      return false;
+    }
+  }
```
Vasniktel, line 58: Nice!

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
```diff
@@ -0,0 +1,146 @@
...
+    const TransformationContext& /* unused */) const {
+  auto instruction =
+      ir_context->get_def_use_mgr()->GetDef(message_.select_id());
+
+  // The instruction must exist and it must be an OpSelect instruction.
+  if (!instruction || instruction->opcode() != SpvOpSelect) {
+    return false;
+  }
+
+  auto block = ir_context->get_instr_block(instruction);
```
Vasniktel, line 48: I would've checked that `block != nullptr` just to be safe.

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
```diff
@@ -0,0 +1,146 @@
...
+      ir_context, SpvOpBranch, 0, 0,
+      opt::Instruction::OperandList{
+          {SPV_OPERAND_TYPE_ID, {merge_block->id()}}}));
+
+  // Add an OpSelectionMerge instruction to the original block.
+  block->AddInstruction(MakeUnique<opt::Instruction>(
+      ir_context, SpvOpSelectionMerge, 0, 0,
+      opt::Instruction::OperandList{
+          {SPV_OPERAND_TYPE_ID, {merge_block->id()}},
+          {SPV_OPERAND_TYPE_SELECTION_CONTROL, {0}}}));
```
Vasniktel, line 108: I believe there is an enum member for `None` operand.

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  // Add a conditional branching instruction to the original block, using the
+  // same conditional as OpSelect and branching to the new block if the
+  // condition is true, to the merge block otherwise.
+  block->AddInstruction(MakeUnique<opt::Instruction>(
+      ir_context, SpvOpBranchConditional, 0, 0,
+      opt::Instruction::OperandList{
+          {SPV_OPERAND_TYPE_ID, {instruction->GetSingleWordInOperand(0)}},
```
Vasniktel, line 116: Condition in OpSelect can be a scalar or a vector of boolean components. I think we should check the latter case in the IsApplicable since we don't support it.

source/fuzz/transformation_replace_opselect_with_conditional_branch.h
```diff
@@ -0,0 +1,53 @@
...
+             TransformationContext* transformation_context) const override;
+
+  protobufs::Transformation ToMessage() const override;
+
+ private:
+  protobufs::TransformationReplaceOpSelectWithConditionalBranch message_;
+};
+}  // namespace fuzz
+}  // namespace spvtools
+#endif  // SOURCE_FUZZ_TRANSFORMATION_REPLACE_OPSELECT_WITH_CONDITIONAL_BRANCH_H
```
Vasniktel, line 53: `\n` are missing.

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
```diff
@@ -0,0 +1,146 @@
...
+          {SPV_OPERAND_TYPE_ID, {new_block->id()}},
+          {SPV_OPERAND_TYPE_ID, {merge_block->id()}}}));
+
+  // Replace the OpSelect instruction in the merge block with an OpPhi.
+  // This:          OpSelect %type %cond %if %else
+  // will become:   OpPhi %type %if %new_block_id %else %block_id
```
Vasniktel, line 122: Nice comments.

test/fuzz/fuzzer_pass_construct_composites_test.cpp
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "source/fuzz/fuzzer_pass_construct_composites.h"
+#include "source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.h"
```
Vasniktel, line 16: Do we need this?

test/fuzz/transformation_replace_opselect_with_conditional_branch_test.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+#include "source/fuzz/transformation_replace_opselect_with_conditional_branch.h"
+
+#include "test/fuzz/fuzz_test_util.h"
+
+namespace spvtools {
+namespace fuzz {
+namespace {
+
+TEST(TransformationReplaceOpSelectWithConditionalBranchTest, Inapplicable) {
```
Vasniktel, line 23: Add a test for the case when OpSelect's condition is a vector of booleans.

source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.cpp
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        // If the instruction separates an OpSampledImage from its use, the
+        // block cannot be split around it and the instruction cannot be
+        // replaced.
+        if (fuzzerutil::
+                SplitBeforeInstructionSeparatesOpSampledImageDefinitionFromUse(
+                    &block, &instruction)) {
+          continue;
+        }
```
Vasniktel, line 70: This loop essentially repeats almost all the conditions from the `IsApplicable` method. You could've refactored these repeatable conditions into a separate static method in the transformation class (something like `CanSplitBlockAt` or similar). It's up to you, though.

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
Vasniktel: `SpvSelectionControlMaskNone`.

source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.h
```diff
@@ -0,0 +1,36 @@
...
+
+#ifndef SOURCE_FUZZ_FUZZER_PASS_REPLACE_OPSELECTS_WITH_CONDITIONAL_BRANCHES_
+#define SOURCE_FUZZ_FUZZER_PASS_REPLACE_OPSELECTS_WITH_CONDITIONAL_BRANCHES_
+
+#include "source/fuzz/fuzzer_pass.h"
+
+namespace spvtools {
+namespace fuzz {
+
+class FuzzerPassReplaceOpSelectsWithConditionalBranches : public FuzzerPass {
```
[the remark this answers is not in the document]
stefanomil: Thanks, I added one

source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.h
```diff
@@ -0,0 +1,36 @@
...
+      FuzzerContext* fuzzer_context,
+      protobufs::TransformationSequence* transformations);
+
+  ~FuzzerPassReplaceOpSelectsWithConditionalBranches() override;
+
+  void Apply() override;
+};
+}  // namespace fuzz
+}  // namespace spvtools
+#endif  // SOURCE_FUZZ_FUZZER_PASS_REPLACE_OPSELECTS_WITH_CONDITIONAL_BRANCHES_
```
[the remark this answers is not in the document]
stefanomil: Added

source/fuzz/protobufs/spvtoolsfuzz.proto
stefanomil: Yes, I agree

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
stefanomil: I have added an assertion

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
stefanomil: Thanks

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
stefanomil: Yes, thanks for spotting this!

source/fuzz/transformation_replace_opselect_with_conditional_branch.h
stefanomil: Added

test/fuzz/fuzzer_pass_construct_composites_test.cpp
stefanomil: No, I forgot to delete it after I quickly tried running the new fuzzer pass

test/fuzz/transformation_replace_opselect_with_conditional_branch_test.cpp
stefanomil: Done

source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.cpp
stefanomil: I think it's simple enough to stay this way (I think it's just 2 checks that could be refactored into a function)

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1480,6 +1481,20 @@ message TransformationReplaceLoadStoreWithCopyMemory {
...
+  // A transformation that takes an OpSelect instruction with a
+  // scalar boolean condition and replaces it with a conditional
+  // branch and an OpPhi instruction.
+
+  // The result id of the OpSelect instruction
+  uint32 select_id = 1;
+
+  // Fresh ids for the new blocks
+  uint32 true_block_fresh_id = 2;
+  uint32 merge_block_fresh_id = 3;
```
afd, line 1495: I think we tend to have a comment per field and line of whitespace between field for protobuf messages, so I suggest commenting in a little more detail - e.g. explaining that in the true block there will be a copy object of the "then" id of the select, and that the merge block with choose between this and the "else" id of the select using an OpPhi.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -1480,6 +1481,20 @@ message TransformationReplaceLoadStoreWithCopyMemory {
   InstructionDescriptor store_instruction_descriptor = 2;
 }
 
+message TransformationReplaceOpSelectWithConditionalBranch {
+
+  // A transformation that takes an OpSelect instruction with a
+  // scalar boolean condition and replaces it with a conditional
+  // branch and an OpPhi instruction.
```
afd, line 1488: I wonder whether it might be nice to have two block fields: true_block and false_block, and to allow at most one of them to be 0.  If neither are 0 we end up with a proper diamond shape, and otherwise we get a half diamond, and it can be in either direction -- see what I mean?  The fuzzer pass could decide which to create.  This might add a bit more diversity.

afd (commented):
A couple of comments - didn't have time to do a full review yet and need to stop for the day.

As mentioned in the comment on the protobuf, it would be nice to allow this transformation to be able to generate a few different forms of the control flow subgraph.

afd:
@stefanomil I was thinking a bit more about this transformation and had the following thought: rather than having the *transformation* do any block splitting, it would be fine and arguably better to have the transformation assume something straightforward, e.g. that the OpSelect instruction is the first instruction in its block.  Or - if it would make things more straightforward still - the OpSelect could be a singleton instruction in its own block.

Then the fuzzer pass could identify the instructions that it would like to turn into conditionals, and could do the necessary block splitting.

Do you see what I mean?

stefanomil:
@afd So, the transformation could check that:
- the OpSelect is the first instruction in its block
- the block that OpSelect is in has exactly one predecessor, which branches unconditionally to it

Then, when applied, the transformation would change the branching instruction of the predecessor, introduce a new empty block and replace OpSelect with OpPhi.

Is this what you meant?

afd:
Yes.  But I stick with [my other suggestion](https://github.com/KhronosGroup/SPIRV-Tools/pull/3681#discussion_r477432797) as well.

source/fuzz/protobufs/spvtoolsfuzz.proto
stefanomil: Done (after making the other changes)

source/fuzz/protobufs/spvtoolsfuzz.proto
stefanomil: Done

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
```diff
@@ -0,0 +1,201 @@
...
+  if (!instruction || instruction->opcode() != SpvOpSelect) {
+    return false;
+  }
+
+  // Check that the condition is a scalar boolean.
+  auto condition = ir_context->get_def_use_mgr()->GetDef(
+      instruction->GetSingleWordInOperand(0));
+  if (!condition) {
+    return false;
+  }
```
afd, line 63: We can assume that the module is valid.  As a result, "condition" *must* be an instruction.  You can thus either omit this check or change it to an assertion with appropriate '&& "..."' string.

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
```diff
@@ -0,0 +1,201 @@
...
+  auto condition_type =
+      ir_context->get_type_mgr()->GetType(condition->type_id());
+  if (!condition_type || !condition_type->AsBool()) {
+    return false;
+  }
```
afd, line 68: Again, condition_type must exist and must be in the type manager - it cannot be zero.  So you could just go ahead and check AsBool holds.

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
```diff
@@ -0,0 +1,201 @@
...
+  // |if_block| will be the true block, if it has been created, the instruction
+  // block otherwise.
+  uint32_t if_block =
+      message_.true_block_id() ? message_.true_block_id() : block->id();
+
+  // |else_block| will be the false block, if it has been created, the
+  // instruction block otherwise.
+  uint32_t else_block =
+      message_.false_block_id() ? message_.false_block_id() : block->id();
+
```
afd, line 155: Up to you but I would probably assert here that if_block and else_block are different (as an integrity check that they can't both end up as block->id()).

source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.cpp
```diff
@@ -0,0 +1,162 @@
...
+        // only be applied to at most the first instruction in a loop header.
+        if (block.IsLoopHeader() && InstructionNeedsSplitBefore(&instruction)) {
+          break;
+        }
+
+        // If the instruction separates an OpSampledImage from its use, the
+        // block cannot be split around it and the instruction cannot be
+        // replaced.
+        if (fuzzerutil::
+                SplitBeforeInstructionSeparatesOpSampledImageDefinitionFromUse(
```
afd, line 80: I think it would be better for this function name to start "Splitting", i.e. "SplittingBeforeInstructionSeparatesOpSampledImageDefinitionFromUse"; could you make that change as part of this PR?

afd (changes requested):
Looks great!  Some minor final changes requested.

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
stefanomil: I turned this into an assertion

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
stefanomil: Done

source/fuzz/transformation_replace_opselect_with_conditional_branch.cpp
stefanomil: I added an assertion

source/fuzz/fuzzer_pass_replace_opselects_with_conditional_branches.cpp
stefanomil: Done

afd (approved):
LGTM - please rebase and I will merge.

stefanomil:
@afd I have rebased

PR #3682, opened 2020-08-11
spirv-opt: Add spvOpcodeIsAccessChain

andreperezmaselco:
This PR implements the `spvOpcodeIsAccessChain` function
that checks if the given opcode represents an access chain
instruction.

### DIFF
--- a/source/opcode.cpp
+++ b/source/opcode.cpp
@@ -719,3 +719,14 @@ std::vector<uint32_t> spvOpcodeMemorySemanticsOperandIndices(SpvOp opcode) {
       return {};
   }
 }
+
+bool spvOpcodeIsAccessChain(SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpAccessChain:
+    case SpvOpInBoundsAccessChain:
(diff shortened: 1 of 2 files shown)


andreperezmaselco:
@s-perron, this implementation will be useful in `spirv-fuzz`.

source/opcode.cpp
```diff
@@ -719,3 +719,14 @@ std::vector<uint32_t> spvOpcodeMemorySemanticsOperandIndices(SpvOp opcode) {
       return {};
   }
 }
+
+bool spvOpcodeIsAccessChain(SpvOp opcode) {
+  switch (opcode) {
+    case SpvOpAccessChain:
+    case SpvOpInBoundsAccessChain:
+    case SpvOpPtrAccessChain:
```
s-perron, line 727: Why don't you include "OpInBoundsPtrAccessChain"?

andreperezmaselco: I just missed it because it was not in sequence with the others in the specification.
Done.

PR #3693, opened 2020-08-13
Improve non-semantic instruction handling in the optimizer

alan-baker:
This does not aim to be complete support for non-semantic instructions in the optimizer. It is enough to handle basics properly and meet the needs of https://github.com/kpet/clvk/pull/250. In the future, `KillInst()` should probably call `KillNonSemanticInfo()` automatically.

* No longer blindly add global non-semantic info instructions to global
  types and values
  * functions now have a list of non-semantic instructions that succeed
    them in the global scope
  * global non-semantic instructions go in global types and values if
    they appear before any function, otherwise they are attached to the
    immediate function predecessor in the module
* changed ADCE to use the function removal utility
* Modified EliminateFunction to have special handling for non-semantic
  instructions in the global scope
  * non-semantic instructions are moved to an earlier function (or full
    global set) if the function they are attached to is eliminated
  * Added IRContext::KillNonSemanticInfo to remove the tree of
    non-semantic instructions that use an instruction
  * this is used in function elimination
* There is still significant work in the optimizer to handle
  non-semantic instructions fully in the optimizer

### DIFF
--- a/source/opt/eliminate_dead_functions_util.cpp
+++ b/source/opt/eliminate_dead_functions_util.cpp
@@ -21,9 +21,32 @@ namespace eliminatedeadfunctionsutil {
 
 Module::iterator EliminateFunction(IRContext* context,
                                    Module::iterator* func_iter) {
+  bool first_func = *func_iter == context->module()->begin();
+  bool seen_func_end = false;
   (*func_iter)
-      ->ForEachInst([context](Instruction* inst) { context->KillInst(inst); },
-                    true);
+      ->ForEachInst(
+          [context, first_func, func_iter, &seen_func_end](Instruction* inst) {
+            if (inst->opcode() == SpvOpFunctionEnd) {
+              seen_func_end = true;
+            }
+            // Move non-semantic instructions to the previous function or
+            // global values if this is the first function.
+            if (seen_func_end && inst->opcode() == SpvOpExtInst) {
+              std::unique_ptr<Instruction> clone(inst->Clone(context));
+              if (first_func) {
+                context->AddGlobalValue(std::move(clone));
+              } else {
+                auto prev_func_iter = *func_iter;
+                --prev_func_iter;
+                prev_func_iter->AddNonSemanticInstruction(std::move(clone));
+              }
+              inst->ToNop();
+            } else {
+              context->KillNonSemanticInfo(inst);
+              context->KillInst(inst);
+            }
+          },
+          true, true);
   return func_iter->Erase();
 }
 
--- a/source/opt/eliminate_dead_functions_util.cpp
+++ b/source/opt/eliminate_dead_functions_util.cpp
@@ -32,7 +32,10 @@ Module::iterator EliminateFunction(IRContext* context,
             // Move non-semantic instructions to the previous function or
             // global values if this is the first function.
             if (seen_func_end && inst->opcode() == SpvOpExtInst) {
+              assert(inst->IsNonSemanticInstruction());
               std::unique_ptr<Instruction> clone(inst->Clone(context));
+              context->ForgetUses(inst);
+              context->AnalyzeDefUse(clone.get());
               if (first_func) {
                 context->AddGlobalValue(std::move(clone));
               } else {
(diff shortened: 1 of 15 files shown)


source/opt/function.h
```diff
@@ -79,6 +79,10 @@ class Function {
   // Saves the given function end instruction.
   inline void SetFunctionEnd(std::unique_ptr<Instruction> end_inst);
 
+  // Add a non-semantic instruction that succeeds this function in the module.
```
s-perron, line 82: Should you say where the new instruction will be relative to the previously existing non-semantic instructions succeeding this function?  Does it come before all of them or after all of them?

source/opt/function.h
```diff
@@ -115,19 +119,24 @@ class Function {
   }
 
   // Runs the given function |f| on instructions in this function, in order,
-  // and optionally on debug line instructions that might precede them.
+  // and optionally on debug line instructions that might precede them and
+  // non-semantic instructions that succceed them.
```
s-perron, line 123: Small nit.  For the debug line instructions the "them" refers to the instructions in the function.  So it sounds like the non-semantic instructions that will be check are those that succeed every individual instruction in the function.  It might be more precise to say "non-semantic instruction that succeed the function."

source/opt/function.h
```diff
@@ -115,19 +119,24 @@ class Function {
...
   void ForEachInst(const std::function<void(const Instruction*)>& f,
-                   bool run_on_debug_line_insts = false) const;
+                   bool run_on_debug_line_insts = false,
+                   bool run_on_non_semantic_insts = false) const;
   // Runs the given function |f| on instructions in this function, in order,
-  // and optionally on debug line instructions that might precede them.
-  // If |f| returns false, iteration is terminated and this function returns
-  // false.
+  // and optionally on debug line instructions that might precede them and
+  // non-semantic instructions that succeed them.  If |f| returns false,
```
s-perron, line 132: Same as above.

source/opt/function.h
```diff
@@ -172,6 +181,8 @@ class Function {
   std::vector<std::unique_ptr<BasicBlock>> blocks_;
   // The OpFunctionEnd instruction.
   std::unique_ptr<Instruction> end_inst_;
+  // Non-semantic instructions preceded by this function.
```
s-perron, line 184: Should is be the instructions that "succeed" this function?

source/opt/ir_context.cpp
```diff
@@ -213,6 +213,21 @@ Instruction* IRContext::KillInst(Instruction* inst) {
...
 }
 
+void IRContext::KillNonSemanticInfo(Instruction* inst) {
+  if (!inst->HasResultId()) return;
+  get_def_use_mgr()->ForEachUser(inst, [this](Instruction* user) {
+    if (user->opcode() == SpvOpExtInst) {
+      auto import_inst =
+          get_def_use_mgr()->GetDef(user->GetSingleWordInOperand(0));
+      std::string import_name = import_inst->GetInOperand(0).AsString();
+      if (import_name.find("NonSemantic.") == 0) {
```
s-perron, line 223: It might be nice having this check in its own function.  So this code would be something like:

```
if (user->IsNonSemantic()) { KillNonSemanticInfo(user); KillInst(user); }
```

source/opt/eliminate_dead_functions_util.cpp
```diff
@@ -21,9 +21,32 @@ namespace eliminatedeadfunctionsutil {
...
-                    true);
+      ->ForEachInst(
+          [context, first_func, func_iter, &seen_func_end](Instruction* inst) {
+            if (inst->opcode() == SpvOpFunctionEnd) {
+              seen_func_end = true;
+            }
+            // Move non-semantic instructions to the previous function or
+            // global values if this is the first function.
+            if (seen_func_end && inst->opcode() == SpvOpExtInst) {
+              std::unique_ptr<Instruction> clone(inst->Clone(context));
```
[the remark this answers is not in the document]
alan-baker: Done with the new function IsNonSemanticInfo.

source/opt/ir_context.cpp
alan-baker: Added IsNonSemanticInstruction().

PR #3694, opened 2020-08-13
spirv-fuzz: Check header dominance when adding dead block

### LINKED ISSUE #3690: spirv-fuzz: Invalidation when adding dead block
`spirv-val` invalidated a fuzzed shader because a selection header does not dominates its merge block. According to SPIR-V specification, each header block must strictly dominate its merge block, unless the merge block is unreachable in the CFG.
```
error: line 172: The selection construct with the selection header 487[%487] does not dominate the merge block 59[%59]
  %59 = OpLabel
```
Debugging `spirv-fuzz`, I found that `TransformationAddDeadBlock::IsApplicable` does not check if `existing_block` strictly dominate `successor_block_id`.

andreperezmaselco:
`TransformationAddDeadBlock` did not check whether the existing block
(that will become a selection header) dominates its successor block (that
will become its merge block).

This change adds the check.

Fixes #3690.

### DIFF
--- a/source/fuzz/transformation_add_dead_block.cpp
+++ b/source/fuzz/transformation_add_dead_block.cpp
@@ -78,6 +78,16 @@ bool TransformationAddDeadBlock::IsApplicable(
     return false;
   }
 
+  // If |successor_block_id| is reachable, then it must be strictly dominated by
+  // |existing_block|.
+  opt::DominatorAnalysis* dominator_analysis =
+      ir_context->GetDominatorAnalysis(existing_block->GetParent());
+  if (dominator_analysis->IsReachable(successor_block_id) &&
+      !dominator_analysis->StrictlyDominates(existing_block->id(),
+                                             successor_block_id)) {
+    return false;
+  }
+
   return true;
 }
 
--- a/source/fuzz/transformation_add_dead_block.cpp
+++ b/source/fuzz/transformation_add_dead_block.cpp
@@ -78,13 +78,24 @@ bool TransformationAddDeadBlock::IsApplicable(
     return false;
   }
 
-  // If |successor_block_id| is reachable, then it must be strictly dominated by
-  // |existing_block|.
+  // |existing_block| must be reachable.
   opt::DominatorAnalysis* dominator_analysis =
       ir_context->GetDominatorAnalysis(existing_block->GetParent());
-  if (dominator_analysis->IsReachable(successor_block_id) &&
-      !dominator_analysis->StrictlyDominates(existing_block->id(),
-                                             successor_block_id)) {
+  if (!dominator_analysis->IsReachable(existing_block->id())) {
+    return false;
+  }
+
+  assert(existing_block->id() != successor_block_id &&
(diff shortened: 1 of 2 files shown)


source/fuzz/transformation_add_dead_block.cpp
```diff
@@ -78,6 +78,16 @@ bool TransformationAddDeadBlock::IsApplicable(
     return false;
   }
 
+  // If |successor_block_id| is reachable, then it must be strictly dominated by
+  // |existing_block|.
+  opt::DominatorAnalysis* dominator_analysis =
+      ir_context->GetDominatorAnalysis(existing_block->GetParent());
+  if (dominator_analysis->IsReachable(successor_block_id) &&
```
Vasniktel, line 85: I think in this case `successor_block_id` is always reachable _from the header block_. I think the spec meant if there is no path from the header block to the merge block.

Thus, I think it is better to remove `IsReachable` check since we might be creating a dead block inside some other dead block. That is, both `existing_block` and `successor_block_id` might be unreahcble from the entry point but there is still a path from the former to the latter. Thus, the former should always dominate the latter.

@paulthomson, what do you think?

andreperezmaselco: But if `existing_block` is unreachable, then it does not dominates `successor_block_id`.

Vasniktel: Nice catch! Indeed. I wonder what the dominator analysis has to say about this situation.

Vasniktel: That's a bit strange, though. The fact that there are no domination rules if the block is unreachable doesn't make any sense if you think about some instruction, its users and dependencies.

paulthomson: Yes, blocks that are statically unreachable are tricky. I often think we should favour iterating over _reachable_ blocks of a function, rather than all blocks. 

Andre: I think we should require the following:


 - `existing_block` is reachable: if not, return false
 - assert(existing_block->id() != successor_block_id);
 - existing_block->id() dominates successor_block_id: if not, return false. 

I suggest "dominates" instead of strictly dominates, otherwise we are essentially combining two checks, which can be confusing. I am fairly certain that the blocks are guaranteed to be different as they must both be reachable and we already check that the successor is not a loop header.

For the dominates check: I assume this is needed for the case where control flow re-converges before a merge block. E.g. existing_block=a, successor_block=c

```
 selection header
   /      \
  a       b
   \     /
      c
      |
 merge block

```

If so, please could you add a comment explaining why we need this check:

Even though we know |successor_block_id| is not a merge block, it might still have multiple predecessors because divergent control flow is allowed to converge early (before the merge block). In this case, when we create the selection construct, its header |existing_block| will not dominate the merge block |successor_block_id|, which is invalid. Thus, |existing_block| must dominate |successor_block_id|.

andreperezmaselco: Done.

PR #3697, opened 2020-08-13
spirv-fuzz: Transformation to replace the use of an irrelevant id

### LINKED ISSUE #3503: spirv-fuzz: a transformation that replaces an "irrelevant" id with some other id of the required type

stefanomil:
A transformation that replaces the use of an irrelevant id with
another id of the same type.

The related fuzzer pass, for every use of an irrelevant id,
checks whether the id can be replaced in that use by another
id of the same type and randomly decides whether to replace
it.

Fixes #3503 .

### DIFF
--- a/source/fuzz/fuzzer.cpp
+++ b/source/fuzz/fuzzer.cpp
@@ -194,6 +194,8 @@ Fuzzer::FuzzerResultStatus Fuzzer::Run(
                                                validator_options_);
 
   // Apply some semantics-preserving passes.
+  // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3764): Enable
+  //  certain passes to run with a higher priority than the others.
   std::vector<std::unique_ptr<FuzzerPass>> passes;
   while (passes.empty()) {
     MaybeAddPass<FuzzerPassAddAccessChains>(
--- a/source/fuzz/fuzzer.cpp
+++ b/source/fuzz/fuzzer.cpp
@@ -67,6 +67,7 @@
 #include "source/fuzz/fuzzer_pass_replace_adds_subs_muls_with_carrying_extended.h"
 #include "source/fuzz/fuzzer_pass_replace_copy_memories_with_loads_stores.h"
 #include "source/fuzz/fuzzer_pass_replace_copy_objects_with_stores_loads.h"
+#include "source/fuzz/fuzzer_pass_replace_irrelevant_ids.h"
 #include "source/fuzz/fuzzer_pass_replace_linear_algebra_instructions.h"
 #include "source/fuzz/fuzzer_pass_replace_loads_stores_with_copy_memories.h"
 #include "source/fuzz/fuzzer_pass_replace_parameter_with_global.h"
@@ -375,6 +376,9 @@ Fuzzer::FuzzerResultStatus Fuzzer::Run(
       transformation_sequence_out);
   MaybeAddPass<FuzzerPassPermutePhiOperands>(
       &final_passes, ir_context.get(), &transformation_context, &fuzzer_context,
+      transformation_sequence_out);
+  MaybeAddPass<FuzzerPassReplaceIrrelevantIds>(
+      &passes, ir_context.get(), &transformation_context, &fuzzer_context,
       transformation_sequence_out);
   MaybeAddPass<FuzzerPassSwapCommutableOperands>(
       &final_passes, ir_context.get(), &transformation_context, &fuzzer_context,
--- a/source/fuzz/fuzzer_context.cpp
+++ b/source/fuzz/fuzzer_context.cpp
@@ -98,6 +98,7 @@ const std::pair<uint32_t, uint32_t> kChanceOfReplacingCopyMemoryWithLoadStore =
 const std::pair<uint32_t, uint32_t> kChanceOfReplacingCopyObjectWithStoreLoad =
     {20, 90};
 const std::pair<uint32_t, uint32_t> kChanceOfReplacingIdWithSynonym = {10, 90};
+const std::pair<uint32_t, uint32_t> kChanceOfReplacingIrrelevantId = {10, 90};
 const std::pair<uint32_t, uint32_t>
     kChanceOfReplacingLinearAlgebraInstructions = {10, 90};
 const std::pair<uint32_t, uint32_t> kChanceOfReplacingLoadStoreWithCopyMemory =
@@ -260,6 +261,8 @@ FuzzerContext::FuzzerContext(RandomGenerator* random_generator,
       ChooseBetweenMinAndMax(kChanceOfReplacingCopyObjectWithStoreLoad);
   chance_of_replacing_id_with_synonym_ =
       ChooseBetweenMinAndMax(kChanceOfReplacingIdWithSynonym);
+  chance_of_replacing_irrelevant_id_ =
+      ChooseBetweenMinAndMax(kChanceOfReplacingIrrelevantId);
   chance_of_replacing_linear_algebra_instructions_ =
       ChooseBetweenMinAndMax(kChanceOfReplacingLinearAlgebraInstructions);
   chance_of_replacing_load_store_with_copy_memory_ =
--- a/source/fuzz/fuzzer_context.cpp
+++ b/source/fuzz/fuzzer_context.cpp
@@ -98,7 +98,7 @@ const std::pair<uint32_t, uint32_t> kChanceOfReplacingCopyMemoryWithLoadStore =
 const std::pair<uint32_t, uint32_t> kChanceOfReplacingCopyObjectWithStoreLoad =
     {20, 90};
 const std::pair<uint32_t, uint32_t> kChanceOfReplacingIdWithSynonym = {10, 90};
-const std::pair<uint32_t, uint32_t> kChanceOfReplacingIrrelevantId = {10, 90};
+const std::pair<uint32_t, uint32_t> kChanceOfReplacingIrrelevantId = {35, 95};
 const std::pair<uint32_t, uint32_t>
     kChanceOfReplacingLinearAlgebraInstructions = {10, 90};
 const std::pair<uint32_t, uint32_t> kChanceOfReplacingLoadStoreWithCopyMemory =
--- a/source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
+++ b/source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
@@ -14,9 +14,16 @@
 
 #include "source/fuzz/fuzzer_pass_replace_irrelevant_ids.h"
 
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/id_use_descriptor.h"
+#include "source/fuzz/transformation_replace_irrelevant_id.h"
+
 namespace spvtools {
 namespace fuzz {
 
+// A fuzzer pass that, for every use of an id that has been recorded as
+// irrelevant, randomly decides whether to replace it with another id of the same
+// type.
 FuzzerPassReplaceIrrelevantIds::FuzzerPassReplaceIrrelevantIds(
     opt::IRContext* ir_context, TransformationContext* transformation_context,
     FuzzerContext* fuzzer_context,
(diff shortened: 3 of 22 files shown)


source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
```diff
@@ -0,0 +1,161 @@
...
+          transformations_to_apply.emplace_back(
+              TransformationReplaceIrrelevantId(
+                  MakeIdUseDescriptorFromUse(GetIRContext(), use_inst,
+                                             in_index),
+                  replacement_id));
+        });
+  }
+
+  // Apply all the transformations.
+  for (auto transformation : transformations_to_apply) {
```
Vasniktel, line 155: Consider using `const auto&` to remove the needless call to the copy constructor.

source/fuzz/fuzzer_util.cpp
```diff
@@ -1342,6 +1342,97 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
...
+      use_in_operand_index > 0) {
+    // This is a function call argument.  It is not allowed to have pointer
+    // type.
+
+    // Get the definition of the function being called.
+    auto function = ir_context->get_def_use_mgr()->GetDef(
+        use_instruction->GetSingleWordInOperand(0));
+    // From the function definition, get the function type.
+    auto function_type = ir_context->get_def_use_mgr()->GetDef(
+        function->GetSingleWordInOperand(1));
```
Vasniktel, line 1413: I think there must be a function in fuzzerutil that can simplify this process.

source/fuzz/transformation_replace_irrelevant_id.cpp
```diff
@@ -0,0 +1,104 @@
...
+namespace spvtools {
+namespace fuzz {
+
+TransformationReplaceIrrelevantId::TransformationReplaceIrrelevantId(
+    const protobufs::TransformationReplaceIrrelevantId& message)
+    : message_(message) {}
+
+TransformationReplaceIrrelevantId::TransformationReplaceIrrelevantId(
+    protobufs::IdUseDescriptor id_use_descriptor, uint32_t replacement_id) {
+  *message_.mutable_id_use_descriptor() = id_use_descriptor;
```
Vasniktel, line 29: Not a strict requirement but consider using either `const protobufs::IdUseDescriptor&` or `std::move(id_use_descriptor)`.

source/fuzz/transformation_replace_irrelevant_id.cpp
```diff
@@ -0,0 +1,104 @@
...
+      FindInstructionContainingUse(message_.id_use_descriptor(), ir_context);
+  if (!use_instruction) {
+    return false;
+  }
+
+  // The type of the id of interest and of the replacement id must be the same.
+  uint32_t type_id_of_interest =
+      ir_context->get_def_use_mgr()->GetDef(id_of_interest)->type_id();
+  uint32_t type_replacement_id = ir_context->get_def_use_mgr()
+                                     ->GetDef(message_.replacement_id())
```
Vasniktel, line 55: Don't forget to check that `replacement_id` exists.

Vasniktel: Also, add a test for it.

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
stefanomil: Done

source/fuzz/fuzzer_util.cpp
stefanomil: I could not find it. There is GetFunctionType, but it requires a function, so we would need to get it and the code would not result simpler. Do you remember using a function that would simplify this?

source/fuzz/transformation_replace_irrelevant_id.cpp
stefanomil: Done (using `const protobufs::IdUseDescriptor&`)

source/fuzz/transformation_replace_irrelevant_id.cpp
stefanomil: Thanks for spotting this, done

source/fuzz/fuzzer_util.cpp
Vasniktel: My idea was to use `fuzzerutil::FindFunction` with `fuzzerutil::GetFunctionType`. Code indeed won't be shorter so it's up to you to replace it if you want to.

source/fuzz/fact_manager.cpp
```diff
@@ -1401,6 +1401,9 @@ class FactManager::IrrelevantValueFacts {
   // See method in FactManager which delegates to this method.
   bool IdIsIrrelevant(uint32_t pointer_id) const;
 
+  // See method in FactManager which delegates to this method.
+  std::unordered_set<uint32_t> GetIrrelevantIds() const;
```
Vasniktel, line 1405: I think it's better to make it a const lvalue reference (`const std::unordered_set<uint32_t>&`).

source/fuzz/fact_manager.cpp
```diff
@@ -1565,6 +1573,10 @@ bool FactManager::IdIsIrrelevant(uint32_t result_id) const {
   return irrelevant_value_facts_->IdIsIrrelevant(result_id);
 }
 
+std::unordered_set<uint32_t> FactManager::GetIrrelevantIds() const {
```
Vasniktel, line 1576: Same here - use a reference.

Vasniktel: Also, maybe add some very trivial tests for this method to `fact_manager_test.cpp`

source/fuzz/fuzzer_util.cpp
```diff
@@ -1342,6 +1342,97 @@ opt::Instruction* GetLastInsertBeforeInstruction(opt::IRContext* ir_context,
...
+
+    // We have found the composite type being accessed by the index we are
+    // considering replacing. If it is a struct, then we cannot do the
+    // replacement as struct indices must be constants.
+    if (composite_type_being_accessed->AsStruct()) {
+      return false;
+    }
+  }
+
+  if (use_instruction->opcode() == SpvOpFunctionCall &&
```
Vasniktel, line 1403: Do we want to disallow the replacement of pointer operands in `OpFunctionCall` with their synonyms (since this function is used in `TransformationReplaceIdWithSynonym` as well)?

source/fuzz/transformation_replace_irrelevant_id.cpp
```diff
@@ -0,0 +1,110 @@
...
+          message_.id_use_descriptor().in_operand_index())) {
+    return false;
+  }
+
+  // The id must be available to use at the use point.
+  return fuzzerutil::IdIsAvailableAtUse(
+      ir_context, use_instruction,
+      message_.id_use_descriptor().in_operand_index(),
+      message_.replacement_id());
+}
```
Vasniktel, line 84: Please, add a TODO(#3722) mentioning that we need to refactor this to account for unreachable blocks.

Vasniktel: As always, use a URL of the issue.

source/fuzz/fact_manager.cpp
stefanomil: Thanks!

source/fuzz/fact_manager.cpp
stefanomil: Done

source/fuzz/fuzzer_util.cpp
stefanomil: This function was originally in TransformationReplaceIdWithSynonym so I think that we want that (unless the transformation was then changed)

source/fuzz/transformation_replace_irrelevant_id.cpp
stefanomil: Done

source/fuzz/fuzzer_util.cpp
Vasniktel: What are the dangers of replacing a pointer parameter with a synonym?

stefanomil: I don't know, it was already in TransformationReplaceIdWithSynonym and doing that check doesn't make a difference to this transformation, since pointers cannot be irrelevant, so they wouldn't be considered anyway

stefanomil: @afd @paulthomson do you know about this?

Vasniktel: Or, I see. I thought you've added this bit. I must've forgotten to rebase my local repo. Thanks for explaining.

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
```diff
@@ -0,0 +1,160 @@
...
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassReplaceIrrelevantIds::~FuzzerPassReplaceIrrelevantIds() = default;
+
+void FuzzerPassReplaceIrrelevantIds::Apply() {
+  // Keep track of the irrelevant ids.
+  std::vector<uint32_t> irrelevant_ids;
```
afd, line 38: Extend the comment a bit to say that this is all ids the fact manager knows about that still exist in the module (noting that some transformations can eliminate ids).

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+  // Keep a map from the type ids of irrelevant ids to all the ids with that
+  // type.
+  std::unordered_map<uint32_t, std::vector<uint32_t>> types_to_ids;
+
+  for (auto id :
+       GetTransformationContext()->GetFactManager()->GetIrrelevantIds()) {
```
afd, line 45: Give this loop a top-level comment, indicating that it finds those irrelevant ids that still exist, and that it identifies all types for which irrelevant ids exist.

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
```diff
@@ -0,0 +1,160 @@
...
+      continue;
+    }
+
+    irrelevant_ids.push_back(id);
+
+    // If the type of this id has not been seen before, add a mapping from this
+    // type id to an empty list in |types_to_ids|. The list will be filled later
+    // on.
+    if (types_to_ids.count(declaration->type_id()) == 0) {
+      types_to_ids[declaration->type_id()] = {};
```
afd, line 58: I have a mild preference for using insert over [], but I don't have a strong argument for why.

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+
+  // If no irrelevant ids were found, return.
+  if (irrelevant_ids.empty()) {
+    return;
+  }
+
+  // Loop through all the ids in the module and fill all the empty lists in
+  // |types_to_ids|.
```
afd, line 68: I would change this to say "For every type for which we have at least one irrelevant id, record all ids in the module that have that type."

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
```diff
@@ -0,0 +1,160 @@
...
+  for (const auto& pair : GetIRContext()->get_def_use_mgr()->id_to_defs()) {
+    uint32_t type_id = pair.second->type_id();
+    if (types_to_ids.count(type_id)) {
```
afd, line 71: I guess you are relying on the fact that if there is no type, type_id() will return 0 and 0 will not be a key for your map.  I suppose that's fine.  I would probably write:

if(type_id && types_to_ids.count(type_id))

personally.

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
```diff
@@ -0,0 +1,160 @@
...
+            // The id is used as an output operand, so we cannot replace this
+            // usage.
+            return;
+          }
+
+          // Randomly decide whether to consider this use.
+          if (!GetFuzzerContext()->ChoosePercentage(
+                  GetFuzzerContext()->GetChanceOfReplacingIrrelevantId())) {
+            return;
+          }
```
afd, line 103: We tend to do this check first - probabilistically I don't think there's any disadvantage to doing so.  I.e., I'd move it to be the first thing that happens in the lambda.

source/fuzz/transformation_replace_irrelevant_id.cpp
```diff
@@ -0,0 +1,113 @@
...
+    const TransformationContext& transformation_context) const {
+  auto id_of_interest = message_.id_use_descriptor().id_of_interest();
+
+  // The id must be irrelevant.
+  if (!transformation_context.GetFactManager()->IdIsIrrelevant(
+          id_of_interest)) {
+    return false;
+  }
+
+  // Find the instruction containing the id use.
```
afd, line 45: "containing the id use." -> "containing the id use; it must exist."

source/fuzz/transformation_replace_irrelevant_id.cpp
```diff
@@ -0,0 +1,113 @@
...
+  if (!fuzzerutil::IdUseCanBeReplaced(
+          ir_context, use_instruction,
+          message_.id_use_descriptor().in_operand_index())) {
+    return false;
+  }
+
+  // The id must be available to use at the use point.
+  // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3722):
+  // IdIsAvailable at use always returns false if the instruction is in an
+  // unreachable block, so it might need to be fixed.
```
afd, line 82: Indent comment text by 1 so that these lines are highlighted as part of the TODO.

source/fuzz/transformation_replace_irrelevant_id.h
```diff
@@ -0,0 +1,62 @@
...
+  //   according to the fact manager.
+  // - The types of the original id and of the replacement ids are the same.
+  // - |message_.replacement_id| is available to use at the enclosing
+  //   instruction of |message_.id_use_descriptor|.
+  // - The original id is in principle replaceable with any other id of the same
+  //   type. In particular:
+  //    - it is not used to index a struct in an access chain (such ids must be
+  //      constants).
+  //    - it is not used as the Sample parameter of an OpImageTexelPointer
+  //      instruction (this parameter must satisfy particular requirements).
```
afd, line 42: Perhaps this comment and the corresponding comment in transformation_replace_id_with_synonym should be shortened so that, instead of "In particular:" onwards they say "See ... for details.", where "..." is the name of the helper function.

afd: The motivation is to reduce comment maintenance burden / the chance of comments becoming out of sync or stale.

afd (changes requested):
This looks very good!

Some changes and issue filing requested.

You'll need to rebase and your changes to the fact manager will be non-trivial since @Vasniktel split the fact manager into multiple classes - hopefully it should be clear what to do.

afd:
@stefanomil Actually please don't rebase until I have re-reviewed and approved.  It's much easier to re-review if I can ask GitHub to show changes since my last review, but rebasing spoils that.

source/fuzz/fuzzer_context.cpp
```diff
@@ -95,6 +95,7 @@ const std::pair<uint32_t, uint32_t> kChanceOfReplacingCopyMemoryWithLoadStore =
 const std::pair<uint32_t, uint32_t> kChanceOfReplacingCopyObjectWithStoreLoad =
     {20, 90};
 const std::pair<uint32_t, uint32_t> kChanceOfReplacingIdWithSynonym = {10, 90};
+const std::pair<uint32_t, uint32_t> kChanceOfReplacingIrrelevantId = {10, 90};
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/fuzzer.cpp
```diff
@@ -385,6 +386,9 @@ Fuzzer::FuzzerResultStatus Fuzzer::Run(
   MaybeAddPass<FuzzerPassPermutePhiOperands>(
       &final_passes, ir_context.get(), &transformation_context, &fuzzer_context,
       transformation_sequence_out);
+  MaybeAddPass<FuzzerPassReplaceIrrelevantIds>(
+      &passes, ir_context.get(), &transformation_context, &fuzzer_context,
+      transformation_sequence_out);
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
stefanomil: Done

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
stefanomil: Done

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
stefanomil: I have changed it to use insert

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
stefanomil: Done

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
stefanomil: I agree that this makes it clearer

source/fuzz/fuzzer_pass_replace_irrelevant_ids.cpp
stefanomil: Done

source/fuzz/transformation_replace_irrelevant_id.cpp
stefanomil: Changed

source/fuzz/transformation_replace_irrelevant_id.cpp
stefanomil: Done

source/fuzz/transformation_replace_irrelevant_id.h
stefanomil: Done, both here and in transformation_replace_id_with_synonym

afd (approved):
Thanks for making the changes.  Some of the bots are failing, but I expect that rebasing will solve this.

Please rebase and then I can take a quick final look and merge if all is well.

PR #3701, opened 2020-08-14
spirv-fuzz: Create synonym via OpPhi and existing synonyms

### LINKED ISSUE #3592: spirv-fuzz: Create synonyms using OpPhi and existing synonyms
Suppose a block B has n predecessors, where n >= 1.  Call the predecessors pred_1, ..., pred_n.

Suppose there are ids, id_1, ... id_n (not necessarily distinct), such that:

- For each i, id_i is available at the end of pred_i
- The id_i's are all synonymous and all have the same type

Then we can add a new instruction to the start of B, of the form:

```
%fresh_id = OpPhi %type %pred_1 %id_1 %pred_2 %id_2 ... %pred_n %id_n
```

and we can add the fact that %fresh_id is synonymous with each id_i (this can be achieved by recording that it is synonymous with one of them, as it will join their equivalence class).

I think this deserves its own transformation class and fuzzer pass, e.g. TransformationAddOpPhiSynonym and FuzzerPassAddOpPhiSynonyms.

stefanomil:
A transformation that adds new OpPhi instructions to blocks with >=1
predecessors, so that its value depends on previously-defined ids of
the right type, which are all synonymous. This instruction is also
recorded as synonymous to the others.

The related fuzzer pass still needs to be implemented.

Fixes #3592 .

### DIFF
--- a/source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
+++ b/source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
@@ -34,10 +34,10 @@ void FuzzerPassAddOpPhiSynonyms::Apply() {
   // same OpPhi instruction.
   auto equivalence_classes = GetIdEquivalenceClasses();
 
-  // Make a list of references to the sets.
-  std::vector<std::set<uint32_t>*> sets;
+  // Make a list of references, to avoid copying sets unnecessarily.
+  std::vector<std::set<uint32_t>*> equivalence_class_pointers;
   for (auto& set : equivalence_classes) {
-    sets.push_back(&set);
+    equivalence_class_pointers.push_back(&set);
   }
 
   // Keep a list of transformations to apply at the end.
@@ -57,23 +57,25 @@ void FuzzerPassAddOpPhiSynonyms::Apply() {
         continue;
       }
 
-      std::set<uint32_t>* chosen_set = nullptr;
+      std::set<uint32_t>* chosen_equivalence_class = nullptr;
 
       if (num_preds > 1) {
         // If the block has more than one predecessor, prioritise sets with at
         // least 2 ids available at some predecessor.
-        chosen_set = MaybeFindSuitableSetRandomly(sets, block.id(), 2);
+        chosen_equivalence_class = MaybeFindSuitableEquivalenceClassRandomly(
+            equivalence_class_pointers, block.id(), 2);
       }
 
       // If a set was not already chosen, choose one with at least one available
       // id.
-      if (!chosen_set) {
-        chosen_set = MaybeFindSuitableSetRandomly(sets, block.id(), 1);
+      if (!chosen_equivalence_class) {
+        chosen_equivalence_class = MaybeFindSuitableEquivalenceClassRandomly(
+            equivalence_class_pointers, block.id(), 1);
       }
 
       // If no suitable set was found, we cannot apply the transformation to
       // this block.
-      if (!chosen_set) {
+      if (!chosen_equivalence_class) {
         continue;
       }
 
@@ -88,10 +90,10 @@ void FuzzerPassAddOpPhiSynonyms::Apply() {
 
       // Choose an id for each predecessor.
       for (uint32_t pred_id : GetIRContext()->cfg()->preds(block.id())) {
-        auto suitable_ids = GetSuitableIds(*chosen_set, pred_id);
+        auto suitable_ids = GetSuitableIds(*chosen_equivalence_class, pred_id);
         assert(!suitable_ids.empty() &&
                "We must be able to find at least one suitable id because the "
-               "set was chosen among suitable sets.");
+               "equivalence class was chosen among suitable ones.");
 
         // If this predecessor has more than one id to choose from and it is the
         // first one of this kind that we found, remember its id.
@@ -117,7 +119,8 @@ void FuzzerPassAddOpPhiSynonyms::Apply() {
       // alternative.
       if (num_preds > 1 && pred_with_alternatives != 0 &&
           ids_chosen.size() == 1) {
-        auto suitable_ids = GetSuitableIds(*chosen_set, pred_with_alternatives);
+        auto suitable_ids =
+            GetSuitableIds(*chosen_equivalence_class, pred_with_alternatives);
         uint32_t chosen_id =
             GetFuzzerContext()->RemoveAtRandomIndex(&suitable_ids);
         if (chosen_id == preds_to_ids[pred_with_alternatives]) {
@@ -175,7 +178,8 @@ FuzzerPassAddOpPhiSynonyms::GetIdEquivalenceClasses() {
     for (auto synonym :
          GetTransformationContext()->GetFactManager()->GetSynonymsForId(
              pair.first)) {
-      // The synonym must not be an indexed access into a composite.
+      // The synonym must be a plain id - it cannot be an indexed access into a
+      // composite.
       if (synonym->index_size() > 0) {
         continue;
       }
@@ -206,8 +210,8 @@ FuzzerPassAddOpPhiSynonyms::GetIdEquivalenceClasses() {
   return id_equivalence_classes;
 }
 
-bool FuzzerPassAddOpPhiSynonyms::SetIsSuitableForBlock(
-    const std::set<uint32_t>& set, uint32_t block_id,
+bool FuzzerPassAddOpPhiSynonyms::EquivalenceClassIsSuitableForBlock(
+    const std::set<uint32_t>& equivalence_class, uint32_t block_id,
     uint32_t distinct_ids_required) {
   bool at_least_one_id_for_each_pred = true;
 
@@ -222,7 +226,7 @@ bool FuzzerPassAddOpPhiSynonyms::SetIsSuitableForBlock(
 
     // Initially assume that there is not a suitable id for this predecessor.
     bool at_least_one_suitable_id_found = false;
-    for (uint32_t id : set) {
+    for (uint32_t id : equivalence_class) {
       if (fuzzerutil::IdIsAvailableBeforeInstruction(GetIRContext(),
                                                      last_instruction, id)) {
         // We have found a suitable id.
@@ -236,17 +240,17 @@ bool FuzzerPassAddOpPhiSynonyms::SetIsSuitableForBlock(
         }
       }
     }
-    // If no suitable id was found for this predecessor, this set is not
-    // suitable and we don't need to check the other predecessors.
+    // If no suitable id was found for this predecessor, this equivalence class
+    // is not suitable and we don't need to check the other predecessors.
     if (!at_least_one_suitable_id_found) {
       at_least_one_id_for_each_pred = false;
       break;
     }
   }
 
-  // The set is suitable if at least one suitable id was found for each
-  // predecessor and we have found at least |distinct_ids_required| distinct
-  // suitable ids in general.
+  // The equivalence class is suitable if at least one suitable id was found for
+  // each predecessor and we have found at least |distinct_ids_required|
+  // distinct suitable ids in general.
   return at_least_one_id_for_each_pred &&
          suitable_ids_found.size() >= distinct_ids_required;
 }
@@ -270,14 +274,17 @@ std::vector<uint32_t> FuzzerPassAddOpPhiSynonyms::GetSuitableIds(
   return suitable_ids;
 }
 
-std::set<uint32_t>* FuzzerPassAddOpPhiSynonyms::MaybeFindSuitableSetRandomly(
+std::set<uint32_t>*
+FuzzerPassAddOpPhiSynonyms::MaybeFindSuitableEquivalenceClassRandomly(
     const std::vector<std::set<uint32_t>*>& candidates, uint32_t block_id,
     uint32_t distinct_ids_required) {
-  auto candidate_sets = candidates;
-  while (!candidate_sets.empty()) {
+  auto remaining_candidates = candidates;
+  while (!remaining_candidates.empty()) {
     // Choose one set randomly and return it if it is suitable.
-    auto chosen = GetFuzzerContext()->RemoveAtRandomIndex(&candidate_sets);
-    if (SetIsSuitableForBlock(*chosen, block_id, distinct_ids_required)) {
+    auto chosen =
+        GetFuzzerContext()->RemoveAtRandomIndex(&remaining_candidates);
+    if (EquivalenceClassIsSuitableForBlock(*chosen, block_id,
+                                           distinct_ids_required)) {
       return chosen;
     }
   }
--- a/source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
+++ b/source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
@@ -14,6 +14,7 @@
 
 #include "source/fuzz/fuzzer_pass_add_opphi_synonyms.h"
 
+#include "source/fuzz/fuzzer_util.h"
 #include "source/fuzz/transformation_add_opphi_synonym.h"
 
 namespace spvtools {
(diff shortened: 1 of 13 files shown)


source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,162 @@
...
+  // There must be at least one predecessor.
+  if (predecessors.empty()) {
+    return false;
+  }
+
+  // There must be exactly a mapping for each predecessor.
+  if (predecessors.size() != static_cast<size_t>(message_.pred_to_id_size()) ||
+      predecessors.size() != preds_to_ids.size()) {
+    return false;
+  }
```
Vasniktel, line 63: I think it would be good to check that there is an id for _at least_ every predecessor. That is, the map can have more entries than there are predecessors. This can be useful during the shrinking.

source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,162 @@
...
+  // Check that each predecessor has a corresponding mapping.
+  for (uint32_t pred : predecessors) {
+    if (preds_to_ids.count(pred) == 0) {
+      return false;
+    }
+  }
+
+  // Check that all the ids exist in the module.
+  for (auto& pair : message_.pred_to_id()) {
+    if (!ir_context->get_def_use_mgr()->GetDef(pair.second())) {
```
Vasniktel, line 74: I would've checked that the type id exists (just to be safe :).

source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,162 @@
...
+    TransformationContext* transformation_context) const {
+  // Get the type id from one of the ids.
+  uint32_t first_id = message_.pred_to_id(0).second();
+  uint32_t type_id = ir_context->get_def_use_mgr()->GetDef(first_id)->type_id();
+
+  // Define the operand list.
+  opt::Instruction::OperandList operand_list;
+
+  // For each predecessor, add the corresponding operands.
+  for (auto pair : message_.pred_to_id()) {
```
Vasniktel, line 126: Consider using a reference here.

Vasniktel, line 126: Consider the case when a block has multiple identical predecessors (e.g. `OpBranchConditional %true %1 %1`). SPIR-V spec requires OpPhi instruction to have exactly one operand for each unique predecessor.

Vasniktel: Also, add a test for this case.

test/fuzz/transformation_add_opphi_synonym_test.cpp
```diff
@@ -0,0 +1,298 @@
...
+  // %13 is not a block label.
+  ASSERT_FALSE(TransformationAddOpPhiSynonym(13, {}, 100)
+                   .IsApplicable(context.get(), transformation_context));
+
+  // Block %12 does not have a predecessor.
+  ASSERT_FALSE(TransformationAddOpPhiSynonym(12, {}, 100)
+                   .IsApplicable(context.get(), transformation_context));
+
+  // Not all predecessors of %16 (%17 and %18) are considered in the map.
+  ASSERT_FALSE(TransformationAddOpPhiSynonym(16, {{17, 19}}, 100)
```
Vasniktel, line 101: I recall one of the compilers produces errors in release mode if the map is used this way. The solution is to add another pair of braces `{{{17, 19}}}`.

source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+bool TransformationAddOpPhiSynonym::IsApplicable(
+    opt::IRContext* ir_context,
+    const TransformationContext& transformation_context) const {
+  // Check that |message_.block_id| is a block label id.
+  auto block = ir_context->get_instr_block(message_.block_id());
+  if (!block) {
+    return false;
+  }
```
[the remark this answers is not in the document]
Vasniktel: However, `block_id` might not be an `OpLabel` instruction. You can use `fuzzerutil::MaybeGetBlock` instead.

source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include "source/fuzz/fuzzer_util.h"
+
+namespace spvtools {
+namespace fuzz {
+TransformationAddOpPhiSynonym::TransformationAddOpPhiSynonym(
+    const protobufs::TransformationAddOpPhiSynonym& message)
+    : message_(message) {}
+
+TransformationAddOpPhiSynonym::TransformationAddOpPhiSynonym(
+    uint32_t block_id, std::map<uint32_t, uint32_t>&& preds_to_ids,
```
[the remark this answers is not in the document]
stefanomil: Thanks! I've changed it to use a const lvalue reference

source/fuzz/transformation_add_opphi_synonym.cpp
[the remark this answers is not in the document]
stefanomil: Thanks, fixed.

source/fuzz/transformation_add_opphi_synonym.cpp
stefanomil: Yeah, you're right

source/fuzz/transformation_add_opphi_synonym.cpp
stefanomil: I don't think that's necessary, since we are assuming that the module is valid

source/fuzz/transformation_add_opphi_synonym.cpp
stefanomil: Done

stefanomil: I don't think this is a special case, since a block that branches to a second one is counted as one predecessor, even if it branches via multiple branches. I have added a test for this anyway

test/fuzz/transformation_add_opphi_synonym_test.cpp
stefanomil: Thanks for letting me know

source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,179 @@
...
+bool TransformationAddOpPhiSynonym::CheckTypeIsAllowed(
+    opt::IRContext* ir_context, uint32_t type_id) {
+  auto type = ir_context->get_type_mgr()->GetType(type_id);
+  if (!type) {
+    return false;
+  }
+
+  return type->AsBool() || type->AsInteger() || type->AsFloat() ||
+         type->AsVector() || type->AsMatrix() || type->AsArray() ||
+         type->AsRuntimeArray() || type->AsStruct();
```
Vasniktel, line 175: Maybe it's better to remove support for runtime arrays? @paulthomson, what do you think?

Also, I don't think it's possible to create an object of OpTypeRuntimeArray explicitly. It only should be the last member of a struct.

source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+    auto pred_block = ir_context->get_instr_block(pair.first);
+    // We should always be able to find the predecessor block, since it is in
+    // the predecessors list of |block|.
+    assert(pred_block && "Could not find one of the predecessor blocks.");
+
+    if (!fuzzerutil::IdIsAvailableBeforeInstruction(
+            ir_context, pred_block->terminator(), pair.second)) {
+      return false;
+    }
```
Vasniktel, line 115: This is a common issue. This function always returns false if `pred_block` is unreachable.

Vasniktel: Please, add a TODO(#3722) comment explaining that we need to refactor this. Use an original issue URL instead of #...

test/fuzz/transformation_add_opphi_synonym_test.cpp
```diff
@@ -0,0 +1,320 @@
...
+void SetUpIdSynonyms(FactManager* fact_manager, opt::IRContext* context) {
+  fact_manager->AddFact(MakeSynonymFact(11, 9), context);
+  fact_manager->AddFact(MakeSynonymFact(13, 9), context);
+  fact_manager->AddFact(MakeSynonymFact(14, 9), context);
+  fact_manager->AddFact(MakeSynonymFact(19, 9), context);
+  fact_manager->AddFact(MakeSynonymFact(20, 9), context);
+  fact_manager->AddFact(MakeSynonymFact(10, 21), context);
+}
+
+TEST(TransformationAddOpPhiSynonymTest, Inapplicable) {
```
Vasniktel, line 42: Do we require `preds_to_ids` to have a synonymous id for _at least_ every predecessor? If we do, add a test for it.

Also, if we do, change the code in the IsApplicable to iterate over all _existing_ predecessors (not all entries in the map) - use `ir_context->cfg()->preds()`.

source/fuzz/transformation_add_opphi_synonym.cpp
stefanomil: So can there be no ids with type OpTypeRuntimeArray?

stefanomil: I'll remove it just to be safe

Vasniktel: > So can there be no ids with type OpTypeRuntimeArray?

I think so. AFAIK, runtime arrays are only used to store things like the shader's (or the kernel's in case of OpenCL) output.

source/fuzz/transformation_add_opphi_synonym.cpp
stefanomil: Done

test/fuzz/transformation_add_opphi_synonym_test.cpp
stefanomil: There is already a test for this:

```
  // Not all predecessors of %16 (%17 and %18) are considered in the map.
  ASSERT_FALSE(TransformationAddOpPhiSynonym(16, {{{17, 19}}}, 100)
                   .IsApplicable(context.get(), transformation_context));

```

Vasniktel: No, sorry for the confusion. I meant the case when the map has an entry for every predecessor + some additional entries.

source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,192 @@
...
+  if (type->AsBool() || type->AsInteger() || type->AsFloat() ||
+      type->AsVector() || type->AsMatrix() || type->AsArray() ||
+      type->AsStruct()) {
+    return true;
+  }
+
+  // If the VariablePointers capability is enabled, we can also allow pointers.
+  return ir_context->get_feature_mgr()->HasCapability(
+             SpvCapabilityVariablePointers) &&
+         type->AsPointer();
```
Vasniktel, line 188: According to SPIR-V spec:
> A variable pointer must point to one of the following storage classes:
> 1. StorageBuffer
> 2. Workgroup (if the VariablePointers capability is declared)

Thus, it would be better to check that the pointer indeed has the required storage class.

Alternatively, we could ignore `VariablePointers` capability for now and add a TODO to fix it later.

stefanomil: Thanks! Pointers will be ignored for now

source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
```diff
@@ -0,0 +1,290 @@
...
+FuzzerPassAddOpPhiSynonyms::~FuzzerPassAddOpPhiSynonyms() = default;
+
+void FuzzerPassAddOpPhiSynonyms::Apply() {
+  // Get a list of synonymous ids with the same type that can be used in the
+  // same OpPhi instruction.
+  auto equivalence_classes = GetIdEquivalenceClasses();
+
+  // Make a list of references to the sets.
+  std::vector<std::set<uint32_t>*> sets;
+  for (auto& set : equivalence_classes) {
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
```diff
@@ -0,0 +1,290 @@
...
+      }
+
+      // The block must have at least one predecessor.
+      size_t num_preds = GetIRContext()->cfg()->preds(block.id()).size();
+      if (num_preds == 0) {
+        continue;
+      }
+
+      std::set<uint32_t>* chosen_set = nullptr;
+
```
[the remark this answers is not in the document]
stefanomil: You are right, I changed it

source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
```diff
@@ -0,0 +1,290 @@
...
+    // Add the new equivalence class to the list of equivalence classes.
+    id_equivalence_classes.emplace_back(std::move(new_equivalence_class));
+  }
+
+  return id_equivalence_classes;
+}
+
+bool FuzzerPassAddOpPhiSynonyms::SetIsSuitableForBlock(
+    const std::set<uint32_t>& set, uint32_t block_id,
+    uint32_t distinct_ids_required) {
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
```diff
@@ -0,0 +1,290 @@
...
+            GetIRContext(), predecessor->terminator(), id)) {
+      suitable_ids.push_back(id);
+    }
+  }
+
+  return suitable_ids;
+}
+
+std::set<uint32_t>* FuzzerPassAddOpPhiSynonyms::MaybeFindSuitableSetRandomly(
+    const std::vector<std::set<uint32_t>*>& candidates, uint32_t block_id,
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+protobufs::Transformation TransformationAddOpPhiSynonym::ToMessage() const {
+  protobufs::Transformation result;
+  *result.mutable_add_opphi_synonym() = message_;
+  return result;
+}
+
+bool TransformationAddOpPhiSynonym::CheckTypeIsAllowed(
+    opt::IRContext* ir_context, uint32_t type_id) {
+  auto type = ir_context->get_type_mgr()->GetType(type_id);
```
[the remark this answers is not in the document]
stefanomil: I have allowed pointers if the right capability is enabled and the right storage classes used

source/fuzz/transformation_add_opphi_synonym.cpp
```diff
@@ -0,0 +1,185 @@
...
+    auto pred_block = ir_context->get_instr_block(pred);
+
+    // We should always be able to find the predecessor block, since it is in
+    // the predecessors list of |block|.
+    assert(pred_block && "Could not find one of the predecessor blocks.");
+
+    // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3722): This
+    // function always returns false if the block is unreachable, so it may need
+    // to be refactored.
+    if (!fuzzerutil::IdIsAvailableBeforeInstruction(
```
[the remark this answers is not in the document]
stefanomil: Thanks!

source/fuzz/fuzzer_pass_add_opphi_synonyms.h
```diff
@@ -0,0 +1,70 @@
...
+
+  void Apply() override;
+
+  // Computes the equivalence classes for the non-pointer and non-irrelevant ids
+  // in the module, where two ids are considered equivalent iff they have been
+  // declared synonymous and they have the same type.
+  std::vector<std::set<uint32_t>> GetIdEquivalenceClasses();
+
+  // Returns true iff |set| contains at least |distinct_ids_required| ids so
+  // that all of these ids are available at the end of at least one predecessor
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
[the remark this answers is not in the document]
stefanomil: Done

afd (approved):
Thanks for making the changes - LGTM!  Please rebase and then I will merge.

stefanomil:
@afd I have rebased

PR #3703, opened 2020-08-14
Let ADCE pass check DebugScope

jaebaek:
In the existing code, ADCE pass does not check DebugScope of an
instruction when it checks the users of each instruction, which results
in removing OpenCL.Debug.100 instructions that are only used by
DebugScope. This commit lets ADCE pass add DebugScope of an instruction
to the live instruction set when the instruction is added to the live
instruction set.

### DIFF
--- a/source/opt/aggressive_dead_code_elim_pass.h
+++ b/source/opt/aggressive_dead_code_elim_pass.h
@@ -81,15 +81,6 @@ class AggressiveDCEPass : public MemPass {
     if (!live_insts_.Set(inst->unique_id())) {
       worklist_.push(inst);
     }
-    if (inst->GetDebugScope().GetLexicalScope() != kNoDebugScope) {
-      auto* scope =
-          get_def_use_mgr()->GetDef(inst->GetDebugScope().GetLexicalScope());
-      AddToWorklist(scope);
-    }
-    if (inst->GetDebugInlinedAt() != kNoInlinedAt) {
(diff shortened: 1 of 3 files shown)


source/opt/aggressive_dead_code_elim_pass.h
```diff
@@ -81,6 +81,15 @@ class AggressiveDCEPass : public MemPass {
...
     }
+    if (inst->GetDebugScope().GetLexicalScope() != kNoDebugScope) {
+      auto* scope =
+          get_def_use_mgr()->GetDef(inst->GetDebugScope().GetLexicalScope());
+      AddToWorklist(scope);
+    }
+    if (inst->GetDebugInlinedAt() != kNoInlinedAt) {
+      auto* inlined_at = get_def_use_mgr()->GetDef(inst->GetDebugInlinedAt());
+      AddToWorklist(inlined_at);
+    }
```
s-perron, line 92: This is not the right place for this.  This should be in the loop where we process and instruction.  Probably somewhere around [here](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/source/opt/aggressive_dead_code_elim_pass.cpp#L535).  The design is that the work list is the list of instructions that need to be processed.  We do not process it when adding it to the work list, but when we remove it from the work list.

jaebaek: done

jaebaek:
Thank you for the code review! PTAL

PR #3710, opened 2020-08-15
spirv-fuzz: Check termination instructions when donating modules

### LINKED ISSUE #3709: spirv-fuzz: FuzzerPassDonateModules does not check termination instructions
At `FuzzerPassDonateModules::MaybeAddLivesafeFunction`, the `|function_to_donate|` termination instructions must be checked if they are `OpKill` or `OpUnreachable` instructions. This check is not currently being performed.

andreperezmaselco:
The `FuzzerPassDonateModules` was not checking if the function to donate had
a block with `OpKill` or `OpUnreachable` as its termination instruction.

Fixes #3709.

### DIFF
--- a/source/fuzz/fuzzer_pass_donate_modules.cpp
+++ b/source/fuzz/fuzzer_pass_donate_modules.cpp
@@ -1234,23 +1234,30 @@ bool FuzzerPassDonateModules::MaybeAddLivesafeFunction(
     }
   }
 
-  // If the function contains OpKill or OpUnreachable instructions, and has
-  // non-void return type, then we need a value %v to use in order to turn
+  // If |function_to_donate| contains OpKill or OpUnreachable instructions, and
+  // has non-void return type, then we need a value %v to use in order to turn
   // these into instructions of the form OpReturn %v.
-  uint32_t kill_unreachable_return_value_id;
-  auto function_return_type_inst =
-      donor_ir_context->get_def_use_mgr()->GetDef(function_to_donate.type_id());
-  if (function_return_type_inst->opcode() == SpvOpTypeVoid) {
-    // The return type is void, so we don't need a return value.
-    kill_unreachable_return_value_id = 0;
-  } else {
-    // We do need a return value; we use zero.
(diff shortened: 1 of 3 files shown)


andreperezmaselco:
Yes, it supports donating functions with `OpKill` and `OpUnreachable` instructions.
But if such instructions are present, the transformation needs `kill_unreachable_return_value_id` to use in order to turn these into instructions of the form `OpReturnValue %kill_unreachable_return_value_id`.

source/fuzz/fuzzer_pass_donate_modules.cpp
```diff
@@ -1234,23 +1234,30 @@ bool FuzzerPassDonateModules::MaybeAddLivesafeFunction(
...
+            function_to_donate.type_id());
+    if (function_return_type_inst->opcode() != SpvOpTypeVoid) {
+      // We do need a return value; we use zero.
+      assert(function_return_type_inst->opcode() != SpvOpTypePointer &&
+             "Function return type must not be a pointer.");
+      kill_unreachable_return_value_id = FindOrCreateZeroConstant(
+          original_id_to_donated_id.at(function_return_type_inst->result_id()),
+          false);
+      break;
+    }
```
paulthomson, line 1258: I understand now. You are checking if the function really contains OpKill or OpUnreachable. If not, then there is no need to create the zero constant.

I think this might be a case where having a function in fuzzer util to check if a function contains OpKill or OpUnreachable. Then you could just do something like:

```
uint32_t kill_unreachable_return_value_id = 0;
auto function_return_type_inst =
    donor_ir_context->get_def_use_mgr()->GetDef(
        function_to_donate.type_id());
if (function_return_type_inst->opcode() != SpvOpTypeVoid 
     && fuzzerutil::FunctionContainsOpKillOrUnreachable(function_to_donate)) {
  kill_unreachable_return_value_id = FindOrCreateZeroConstant(
    original_id_to_donated_id.at(function_return_type_inst->result_id()),
    false);
}


```

andreperezmaselco: Done.

PR #3718, opened 2020-08-17
Correctly replace debug lexical scope of instruction

jaebaek:
When we update OpenCL.DebugInfo.100 lexical scopes e.g., DebugFunction,
we have to replace DebugScope of each instruction that uses the lexical
scope correctly.

### DIFF
--- a/source/opt/ir_context.h
+++ b/source/opt/ir_context.h
@@ -418,13 +418,13 @@ class IRContext {
   bool ReplaceAllUsesWith(uint32_t before, uint32_t after);
 
   // Replace all uses of |before| id with |after| id if those uses
-  // (instruction, operand pair) return true for |predicate|. Returns true if
+  // (instruction) return true for |predicate|. Returns true if
   // any replacement happens. This method does not kill the definition of the
   // |before| id. If |after| is the same as |before|, does nothing and return
   // false.
   bool ReplaceAllUsesWithPredicate(
       uint32_t before, uint32_t after,
-      const std::function<bool(Instruction*, uint32_t)>& predicate);
+      const std::function<bool(Instruction*)>& predicate);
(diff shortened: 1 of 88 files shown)


source/opt/ir_context.h
```diff
@@ -576,6 +576,24 @@ class IRContext {
...
+  // Updates DebugInlinedAt of |inst| and analyzes use information
+  // of DebugInlinedAt.
+  void UpdateDebugInlinedAt(Instruction* inst, uint32_t dbg_inlined_at) {
+    inst->UpdateDebugInlinedAt(dbg_inlined_at);
+    if (AreAnalysesValid(kAnalysisDebugInfo)) {
+      get_debug_info_mgr()->AnalyzeDebugInst(inst);
+    }
+  }
+
+  // Updates DebugScope and line of |inst| and analyzes use information
```
s-perron, line 588: You can update the analysis in `Instruction::UpdateDebugInfoFrom`.  The instruction class has access to the context.

source/opt/ir_context.h
```diff
@@ -576,6 +576,24 @@ class IRContext {
     return GetFunction(inst->result_id());
   }
 
+  // Updates DebugInlinedAt of |inst| and analyzes use information
+  // of DebugInlinedAt.
+  void UpdateDebugInlinedAt(Instruction* inst, uint32_t dbg_inlined_at) {
```
s-perron, line 581: I don't think you need this function.  You can move the update of the debug info manager to the instruction classes version of the function.

jaebaek: done

source/opt/ir_context.h
jaebaek: done

PR #3719, opened 2020-08-17
Fix SSA-rewrite to remove DebugDeclare for variables without loads

greg-lunarg:
Previously, the removal of DebugDeclares was predicated on the module "changing". In this case, "changing" actually meant loads being removed. This omitted the case where a variable had all its loads removed by a previous pass but still needed its DebugDeclare to be replaced with DebugValues by SSA-rewrite. The test case included in this PR gives further details.

This commit now has SSA-rewrite do  the DebugDeclare removal unconditionally. It appears that the purpose of the previous condition was just one of optimization, so it seems that its removal will not cause a functional problem. It is imagined that few functions will not require SSA-Rewrite, so few functions will suffer compile-time degradation from this change, and for those that do, the degradation will likely be nearly insignificant.


greg-lunarg:
@jaebaek Can you please review?

Issue #3704, opened 2020-08-15
spirv-opt: InlineExhaustivePass doesn't handle OpCopyObject of void


Vasniktel:
spirv-opt produces the following error
> error: line 11: ID 8[%8] has not been defined
>    %9 = OpCopyObject %void %8
>
> error: line 0: Validation failed after pass inline-entry-points-exhaustive

when executed with the following command `spirv-opt <attached shader> -o out.spv --inline-entry-points-exhaustive --validate-after-all`.

The following shader causes the trouble
```
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %4 "main"
               OpExecutionMode %4 OriginUpperLeft
               OpSource ESSL 320
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %4 = OpFunction %2 None %3
          %5 = OpLabel
          %8 = OpFunctionCall %2 %6
          %9 = OpCopyObject %2 %8
               OpReturn
               OpFunctionEnd
          %6 = OpFunction %2 None %3
          %7 = OpLabel
               OpReturn
               OpFunctionEnd
```
Can be reproduced on c20995ef.

Might be related to #3547.

Vasniktel:
@s-perron @alan-baker ping

PR #3720, opened 2020-08-17
Add undef for inlined void function

s-perron:
It is possible that the result of a void function call is used.  In case
it is used, we need something that still defines its id after inlining.
We add an undef for that purpose.

Fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/3704

### DIFF
--- a/source/opt/inline_pass.cpp
+++ b/source/opt/inline_pass.cpp
@@ -619,6 +619,14 @@ bool InlinePass::GenInlineCode(
     assert(resId != 0);
     AddLoad(calleeTypeId, resId, returnVarId, &new_blk_ptr,
             call_inst_itr->dbg_line_inst(), call_inst_itr->GetDebugScope());
+  } else {
+    // Even though it is very unlikely, it is possible that the result id of
+    // the void-function call is used, so we need to generate an instruction
+    // with that result id.
+    std::unique_ptr<Instruction> undef_inst(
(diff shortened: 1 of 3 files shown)


source/opt/inline_pass.cpp
```diff
@@ -619,6 +619,14 @@ bool InlinePass::GenInlineCode(
     assert(resId != 0);
     AddLoad(calleeTypeId, resId, returnVarId, &new_blk_ptr,
             call_inst_itr->dbg_line_inst(), call_inst_itr->GetDebugScope());
+  } else {
+    // Even though it is very unlikely, it is possible that the result id of
+    // the void-function call is used, so we need to generate an instruction
+    // with that result id.
+    std::unique_ptr<Instruction> undef_inst(
```
dneto0, line 626: OpUndef's are preferred to be in the constants section, not in function bodies.
This is ok. But we should consider this the deprecated way.

dneto0:
Tests are failing because some expectations were not updated.

For an improved MR: 
- only generate the undef value if the return value has a use
- generate the undef value in the constants section of the module, rather than in the function body.

s-perron:
> only generate the undef value if the return value has a use

That is hard because we do not keep the def-use changes valid while inlining.  We would have to keep rebuilding the def-use chains to do that, which would cost more compile time that I would like to use.  I prefer to get get DCE to delete the undef if it is not used.

> generate the undef value in the constants section of the module, rather than in the function body.

That is doable.

PR #3724, opened 2020-08-18
Handle DebugScope in compact-ids pass

jaebaek:
We have to update OpenCL.DebugInfo.100 DebugScope for instructions if we
change its lexical scope or DebugInlinedAt.

### DIFF
--- a/source/opt/compact_ids_pass.cpp
+++ b/source/opt/compact_ids_pass.cpp
@@ -56,6 +56,17 @@ Pass::Status CompactIdsPass::Process() {
           }
           ++operand;
         }
+        if (inst->GetDebugScope().GetLexicalScope() != kNoDebugScope) {
+          auto it =
+              result_id_mapping.find(inst->GetDebugScope().GetLexicalScope());
+          if (it != result_id_mapping.end())
+            inst->UpdateLexicalScope(it->second);
+        }
+        if (inst->GetDebugInlinedAt() != kNoInlinedAt) {
+          auto it = result_id_mapping.find(inst->GetDebugInlinedAt());
+          if (it != result_id_mapping.end())
+            inst->UpdateDebugInlinedAt(it->second);
+        }
       },
       true);
 
--- a/source/opt/compact_ids_pass.cpp
+++ b/source/opt/compact_ids_pass.cpp
@@ -21,6 +21,22 @@
 
 namespace spvtools {
 namespace opt {
+namespace {
+
+uint32_t GetRemappedId(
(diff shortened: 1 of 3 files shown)


source/opt/compact_ids_pass.cpp
```diff
@@ -56,6 +56,17 @@ Pass::Status CompactIdsPass::Process() {
           }
           ++operand;
         }
+        if (inst->GetDebugScope().GetLexicalScope() != kNoDebugScope) {
+          auto it =
+              result_id_mapping.find(inst->GetDebugScope().GetLexicalScope());
+          if (it != result_id_mapping.end())
```
s-perron, line 62: In what situations will you not find the lexical scope in the map?

s-perron: I would suggestion you extract the code in the loop that from lines 37-45 into a new function "GetRemappedId" or something like that.  Then you should call that function here as well.

source/opt/compact_ids_pass.cpp
```diff
@@ -56,6 +56,17 @@ Pass::Status CompactIdsPass::Process() {
...
+            inst->UpdateLexicalScope(it->second);
```
s-perron, line 63: You should be setting modified to true when you change something.  You should only update the value if it is actually a different value.

source/opt/compact_ids_pass.cpp
```diff
@@ -56,6 +56,17 @@ Pass::Status CompactIdsPass::Process() {
...
+        }
+        if (inst->GetDebugInlinedAt() != kNoInlinedAt) {
+          auto it = result_id_mapping.find(inst->GetDebugInlinedAt());
```
s-perron, line 66: All of the same comments as above.

source/opt/compact_ids_pass.cpp
jaebaek: done

source/opt/compact_ids_pass.cpp
jaebaek: done

source/opt/compact_ids_pass.cpp
jaebaek: done

source/opt/compact_ids_pass.cpp
```diff
@@ -21,6 +21,22 @@
 
 namespace spvtools {
 namespace opt {
+namespace {
+
+uint32_t GetRemappedId(
```
s-perron, line 26: Please add comments describing this function.

jaebaek: Ah right. Done.

PR #3728, opened 2020-08-20
spirv-fuzz: Add words instead of logical operands

### LINKED ISSUE #3727: spirv-fuzz: TransformationAddConstantScalar adds operands instead of words
In the `TransformationAddConstantScalar`, when the scalar has more than 1 `word`, the `Apply` function is adding 1 `operand` per `word` instead of adding 1 `operand` with all `words`.

andreperezmaselco:
Fixes #3727.


Vasniktel (approved):
LGTM.

afd (approved):
LGTM.

PR #3734, opened 2020-08-21
spirv-fuzz: Overflow ids

afd:
This change adds the notion of "overflow ids", which can be used
during shrinking to facilitate applying transformations that would
otherwise have become inapplicable due to earlier transformations
being removed.


paulthomson (approved):
LGTM!

PR #3744, opened 2020-08-25
spirv-fuzz: Replace id in OpPhi coming from a dead predecessor

### LINKED ISSUE #3726: spirv-fuzz: Transformation to replace id in OpPhi when predecessor is dead
The transformation to add an OpPhi synonym (issue: #3592 , PR: #3701 )
accepts a mapping from the predecessors of a block to ids that must
all be synonymous. If an id corresponds to a predecessor that is a
dead block (it will never be executed), this rule can be relaxed and
we could accept any available id of the right type.

After the discussion: we could actually make a new transformation for
replacing ids in OpPhi instructions.

stefanomil:
This transformation takes the id of an OpPhi instruction, of a dead
predecessor of the block containing it and a replacement id of
available to use and of the same type as the OpPhi, and changes
the id in the OpPhi corresponding to the given predecessor.

For example, %id = OpPhi %type %v1 %p1 %v2 %p2
becomes %id = OpPhi %type %v3 %p1 %v2 %p2
if the transformation is given %id, %p1 and %v3, %p1 is a dead block,
%v3 is type type and it is available to use at the end of %p1.

The fuzzer pass randomly decides to apply the transformation to OpPhi
instructions for which at least one of the predecessors is dead

Fixes #3726 .

### DIFF
--- a/source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
+++ b/source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
@@ -1,0 +1,115 @@
+// Copyright (c) 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.h"
+#include "source/fuzz/transformation_replace_opphi_id_from_dead_predecessor.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassReplaceOpPhiIdsFromDeadPredecessors::
+    FuzzerPassReplaceOpPhiIdsFromDeadPredecessors(
+        opt::IRContext* ir_context,
+        TransformationContext* transformation_context,
+        FuzzerContext* fuzzer_context,
+        protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassReplaceOpPhiIdsFromDeadPredecessors::
+    ~FuzzerPassReplaceOpPhiIdsFromDeadPredecessors() = default;
+
+void FuzzerPassReplaceOpPhiIdsFromDeadPredecessors::Apply() {
+  // Keep a vector of the transformations to apply.
+  std::vector<TransformationReplaceOpPhiIdFromDeadPredecessor> transformations;
+
+  // Loop through the blocks in the module.
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      // Only consider dead blocks.
+      if (!GetTransformationContext()->GetFactManager()->BlockIsDead(
+              block.id())) {
+        continue;
+      }
+
+      // Find all the uses of the label id of the block inside OpPhi
+      // instructions.
+      GetIRContext()->get_def_use_mgr()->ForEachUse(
+          block.id(), [this, &function, &block, &transformations](
+                          opt::Instruction* instruction, uint32_t) {
+            // Only consider OpPhi instructions.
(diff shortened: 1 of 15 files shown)


source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
```diff
@@ -0,0 +1,115 @@
...
+            // Get the current id corresponding to the predecessor.
+            uint32_t current_id = 0;
+            for (uint32_t i = 1; i < instruction->NumInOperands(); i += 2) {
+              if (instruction->GetSingleWordInOperand(i) == block.id()) {
+                // The corresponding id is at the index of the block - 1.
+                current_id = instruction->GetSingleWordInOperand(i - 1);
+                break;
+              }
+            }
+
```
afd, line 72: It would be good to assert here that current_id is non-zero - i.e. that a predecessor was always found.

source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
```diff
@@ -0,0 +1,115 @@
...
+            // Get the type id.
```
afd, line 73: Fine to leave this comment in if you like -- and I know in my reviews I tend to ask for more comments -- but arguably this comment is redundant as it basically parrots what the statement does.

I tend to favour descriptive comments about blocks of code, or per-line comments about particularly tricky lines of code.

source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
```diff
@@ -0,0 +1,115 @@
...
+            uint32_t type_id = instruction->type_id();
+
+            // If the list of candidates for this type was not already computed,
+            // compute it.
```
afd, line 77: What is conditional about the following?  It looks like you compute a list of candidates unconditionally.

source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
```diff
@@ -0,0 +1,115 @@
...
+            const auto& candidates = FindAvailableInstructions(
+                &function, &block, block.end(),
+                [type_id, current_id](opt::IRContext* /* unused */,
+                                      opt::Instruction* candidate) -> bool {
+                  // Only consider instructions with a result id, with the
+                  // right type and different from the currently-used one.
```
afd, line 83: I'd rephrase to say "with a result id different from the currently-used one, and with the right type"

source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
```diff
@@ -0,0 +1,115 @@
...
+                         candidate->result_id() != current_id;
+                });
+
+            // If there is no possible replacement, we cannot apply any
+            // transformation.
+            if (candidates.empty()) {
+              return;
+            }
+
+            // Choose one of the candidates.
```
afd, line 95: I like this comment BTW.  Not that the line of code is particularly complex, but it's complex enough that you don't see at a glance what it does.

afd (changes requested):
Looks great!  Very minor changes requested.

source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
stefanomil: Done

source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
stefanomil: I agree that this comment was redundant, I have removed it

source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
stefanomil: I think this is a comment that I had written when I thought about doing this in a slightly different way, and then forgot to change it. I have now changed it.

source/fuzz/fuzzer_pass_replace_opphi_ids_from_dead_predecessors.cpp
stefanomil: Done

afd (approved):
Thanks for making the changes.  Please rebase - though note that every PR that adds a new transformation requires a rebase (due to changes in the .proto file), so it will be most efficient if you decide on an order to rebase your pending PRs and have me merge them.

Issue #3743, opened 2020-08-24
spirv-opt: OpenCL.DebugInfo.100: Assert: Two DebugFunction instruction exists for a single OpFunction.


greg-lunarg:
Compiling a Sascha Willems shader with dxc then passing through spirv-opt legalization gives:

Assertion failed: GetDebugFunction(dbg_inst->GetSingleWordOperand( kDebugFunctionOperandFunctionIndex)) == nullptr && "Two DebugFunction instruction exists for a single OpFunction.", file C:\Users\Greg\valve5\spirv-tools\source\opt\debug_info_manager.cpp, line 571

The dxc compile was:

dxc.exe -spirv -E main -T ps_6_1 -fspv-target-env=vulkan1.2 -fcgl -fspv-debug=rich -Vd -Fo genbrdflut.frag.rdbg.Od.spv genbrdflut.frag

The spirv-opt command is:

spirv-opt.exe --skip-validation --wrap-opkill --eliminate-dead-branches --merge-return --inline-entry-points-exhaustive --eliminate-dead-functions --private-to-local --fix-storage-class --eliminate-local-single-block --eliminate-local-single-store --eliminate-dead-code-aggressive --scalar-replacement --eliminate-local-single-block --eliminate-local-single-store --eliminate-dead-code-aggressive --eliminate-local-multi-store --eliminate-dead-code-aggressive --ccp --loop-unroll --eliminate-dead-branches --simplify-instructions --eliminate-dead-code-aggressive --copy-propagate-arrays --vector-dce --eliminate-dead-inserts --reduce-load-size --eliminate-dead-code-aggressive -o genbrdflut.frag.rdbg.lgl.spv genbrdflut.frag.rdbg.Od.spv

FWIW, I have discovered that the inlining is pass is introducing invalid code. Here is the spirv-opt command:

spirv-opt.exe --wrap-opkill --eliminate-dead-branches --merge-return --inline-entry-points-exhaustive -o genbrdflut.frag.rdbg.inl.spv genbrdflut.frag.rdbg.Od.spv

Here is the spirv-val command:

spirv-val.exe genbrdflut.frag.rdbg.inl.spv
error: line 390: Invalid instruction word count: 0

I am also seeing this with the Sascha Willems shader pbrtexture.frag. These are the only two I have looked at so far.

@jaebaek I am assigning this to myself.

jaebaek:
Ah, I just realized this bug today and opened it in #3750.
[...]

PR #3748, opened 2020-08-25
Fix DebugNoScope to not output InlinedAt operand.

greg-lunarg:
This problem was seen as part of https://github.com/KhronosGroup/SPIRV-Tools/issues/3743 although it does not fix this issue, so the issue will stay open.

Essentially, the logic for outputting DebugNoScope was printing an InlinedAt operand when "available", which is not valid. This prevents this from happening.

The extra operand at the end of DebugNoScope created the appearance of an instruction with zero words which confused validation.


greg-lunarg:
@jaebaek Please review

jaebaek:
Thank you for working on it!

PR #3749, opened 2020-08-25
Do not register DebugFunction for functions optimized away by inlining.

greg-lunarg:
This fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/3743

The debug info manager was attempting to register functions that had been optimized away by inlining. They all have a function id of DebugInfoNone which caused a problem with the DebugFunction map which keys on the function id, resulting in conflict.

The solution is to not register functions which map to DebugInfoNone.

### DIFF
--- a/source/opt/debug_info_manager.cpp
+++ b/source/opt/debug_info_manager.cpp
@@ -99,6 +99,13 @@ void DebugInfoManager::RegisterDbgFunction(Instruction* inst) {
   assert(inst->GetOpenCL100DebugOpcode() == OpenCLDebugInfo100DebugFunction &&
          "inst is not a DebugFunction");
   auto fn_id = inst->GetSingleWordOperand(kDebugFunctionOperandFunctionIndex);
+  // Do not register function that has been optimized away
+  auto fn_inst = GetDbgInst(fn_id);
+  if (fn_inst != nullptr) {
+    assert(GetDbgInst(fn_id)->GetOpenCL100DebugOpcode() ==
+           OpenCLDebugInfo100DebugInfoNone);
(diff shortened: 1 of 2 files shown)


greg-lunarg:
@jaebaek Please review

source/opt/debug_info_manager.cpp
```diff
@@ -99,6 +99,8 @@ void DebugInfoManager::RegisterDbgFunction(Instruction* inst) {
   assert(inst->GetOpenCL100DebugOpcode() == OpenCLDebugInfo100DebugFunction &&
          "inst is not a DebugFunction");
   auto fn_id = inst->GetSingleWordOperand(kDebugFunctionOperandFunctionIndex);
+  // Do not register function that has been optimized away
+  if (fn_id == GetDebugInfoNone()->result_id()) return;
```
jaebaek, line 103: Since we can add another `DebugInfoNone` and do not reuse it for multiple `DebugFunction`, it can be different from `GetDebugInfoNone()->result_id()` even when it is actually `DebugInfoNone`.

Why don't we change this to
```
// Skip if Function operand of DebugFunction is DebugInfoNone
if (GetDbgInst(fn_id) != nullptr) {
  assert(GetDbgInst(fn_id)->GetOpenCL100DebugOpcode() == OpenCLDebugInfo100DebugInfoNone);
  return;
}
```
?

greg-lunarg: Change made and test added. I believe I have addressed all concerns.

jaebaek:
Thank you for working on it!

PR #3752, opened 2020-08-26
Add Wasm build

pjoe:
This adds build for Wasm (JavaScript). Based on https://github.com/pjoe/SPIRV-Tools.js (see also on npm: https://www.npmjs.com/package/spirv-tools).

This consists of:
- A small C++ wrapper in `source/wasm/spirv-tools.cpp` using embind
  - Currently only assembly and disassembly is wrapped
- Typescript definitions in `source/wasm/spirv-tools.d.ts`
- A small JS test in `test/wasm/test.js`
- Emscripten based build using docker and docker-compose. To build just run `docker-compose up`
- GitHub actions CI setup to build this
  - For now publishing to npm is commented out, need to figure out credentials (a npm token needs to be set as a secret on the repo)
  - See example build output here: https://github.com/pjoe/SPIRV-Tools/runs/1030409502?check_suite_focus=true


source/wasm/build.sh
```diff
@@ -0,0 +1,75 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+NUM_CORES=$(nproc)
```
dneto0, line 19: I'm a little concerned that this is not quite right for the cloud.  But I don't have data to back up that concern.
Presumably a cloud instance will have an appropriate limit.

source/wasm/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "spirv-tools",
+  "version": "20.4.2",
```
dneto0, line 3: What does this version come from?
Normally the SPIRV-Tools version comes from the CHANGES file.

source/wasm/spirv-tools.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include <emscripten/bind.h>
+#include <emscripten/val.h>
+using namespace emscripten;
+
+auto print_msg_to_stderr = [](spv_message_level_t, const char*,
```
dneto0, line 25: You should be able to declare this as a regular function.
The constructor for the  std::function will auto-convert it, I think.

Also, the style guide recommends not creating static-duration objects with non-trivial destructors.  https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables 
 I think the lambda falls into that category.

dneto0 (changes requested):
My main concern is how you set the version.

(This is very cool!)

source/wasm/build.sh
pjoe: I think this 'should just work'™, as it should just se the virtual CPU's of the VM it's running in, but I'll do some experimentation to make sure.

source/wasm/spirv-tools.cpp
pjoe: I think I got this from the example: https://github.com/KhronosGroup/SPIRV-Tools/blob/master/examples/cpp-interface/main.cpp#L43, Anyway will take a look.

pjoe:
While waiting for me fixing based on comments, you can check it out running in browser here (naga with spv output -> spirv-tools dis): https://wasm-shaders.vercel.app/.

Still very much WIP 😄

source/wasm/spirv-tools.cpp
pjoe: Guess I had just blindly copy/pasted from the example - regular function works just fine here.

source/wasm/build.sh
pjoe: I added build output showing number of detected cores. For GitHub actions it shows 2, which AFAIK matches the published spec for the GitHub action runners.

source/wasm/package.json
pjoe: Added logic (bash, sed) to take this from the CHANGES file and at the GITHUB_RUN_ID as patch part of.

So `v2020.5` from changes becomes e.g. `2020.5.2`

I also changed the publish step to use: https://github.com/marketplace/actions/npm-publish (for now with dry-run)

pjoe:
Had to rebase, because of changes to spirv-headers.

Otherwise, I think it's almost there, you can see build output here: https://github.com/pjoe/SPIRV-Tools/actions/runs/364149807

Remaining thing to figure out is the actual NPM publish step:
- When should it publish, e.g. only on push to master or only when tagged?
- Where should it publish to? For my testing I setup `spirv-tools` on NPM, can of course keep using this and transfer ownership, but maybe better would be something like `@khronos/spirv-tools`?
- For GitHub Actions to do the publish, a repository secret (NPM_TOKEN) with a NPM automation access token need to be set.
- Remove the `dry-run` option in the GitHub Actions config yaml.

source/wasm/build.sh
```diff
@@ -0,0 +1,78 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+NUM_CORES=$(nproc)
+echo "Detected $NUM_CORES cores for building"
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+VERSION=$(sed -n '0,/^v20/ s/^v\(20[0-9.]*\).*/\1/p' $DIR/../../CHANGES).${GITHUB_RUN_NUMBER:-0}
```
dneto0, line 23: This ignores any `-dev` suffix which may be present.
But npm assumes semantic versioning.  (https://docs.npmjs.com/about-semantic-versioning )

So we could have:
    

           CHANGES file says                npm sees
           2021.5-dev                             2021.5.##
           2021.5                                     2021.5.##  (## incremeents by built number)
           2021.6-dev                             2021.6.#
           2021.6                                     2021.6

In principle, breaking changes could occur in the -dev series, which could break assumptions downstream.
On the other hand, the current JS interface is very narrow: assemble and disassemble, and the environment enums.   So there's essentially no risk this will break.

dneto0 (commented):
Thanks for this contribution.
My apologies for the long delay in looking at this.  I really didn't know enough JS/TS/node, etc.

It works when I rebase against master.  I plan to rebase this and make a new PR.

dneto0:
> Had to rebase, because of changes to spirv-headers.

Same.

> 
> Otherwise, I think it's almost there, you can see build output here: https://github.com/pjoe/SPIRV-Tools/actions/runs/364149807
> 
> Remaining thing to figure out is the actual NPM publish step:
> 
> * When should it publish, e.g. only on push to master or only when tagged?

Publishing only when specific tags are created would be the ideal.
That would (mostly) avoid the concerns I had about version extraction.

> * Where should it publish to? For my testing I setup `spirv-tools` on NPM, can of course keep using this and transfer ownership, but maybe better would be something like `@khronos/spirv-tools`?

@khronos would be the natural place.
I see https://www.npmjs.com/~khronos exists and has some GLTF content.
I'd have to coordinate with someone inside Khronos to enable publishing there.


> * For GitHub Actions to do the publish, a repository secret (NPM_TOKEN) with a NPM automation access token need to be set.

I suggest removing the publish step for now  (comment it out). Then at least we can land the flow, folks can make the package themselves from the source tree.

> * Remove the `dry-run` option in the GitHub Actions config yaml.

dneto0:
> I suggest removing the publish step for now (comment it out). Then at least we can land the flow, folks can make the package themselves from the source tree.
> 
> > * Remove the `dry-run` option in the GitHub Actions config yaml.

Heh.  Keeping the dry-run is sufficient.  :-)    I'm learning, I really am.

dneto0:
Ok, as I understand it:
- since dry-run is used, this won't actually publish to the registry
- but the flow would exist for someone to make the package locally.

So I would be happy to land this after a rebase and squash.

dneto0:
> 
> So I would be happy to land this after a rebase and squash.

This PR only adds files.  So squash-and-merge will work as I want (because GitHub squash-and-merge uses the fast-forward option).

dneto0 (approved):
Thanks!

dneto0:
Landed this to land the mechanism. 
Pending work is about policy of when to release, and where.

pjoe:
No worries on the delay, just happy this was still useful 😄

PR #3753, opened 2020-08-26
spirv-fuzz: Compute interprocedural loop nesting depth of blocks

stefanomil:
This PR extends CallGraph with functions to:
- return a list of functions in lexicographical order, with respect to
  function calls
- the maximum loop nesting depth that a function can be called from
  (computed interprocedurally, e.g. if foo() calls bar() at depth 2
  and bar() calls baz() at depth 1, the maximum depth of baz() will
  be 3).

### DIFF
--- a/source/fuzz/call_graph.cpp
+++ b/source/fuzz/call_graph.cpp
@@ -33,95 +33,14 @@ CallGraph::CallGraph(opt::IRContext* context) {
   // to the corresponding maximum depth.
   std::map<std::pair<uint32_t, uint32_t>, uint32_t> call_to_max_depth;
 
-  // Consider every function.
-  for (auto& function : *context->module()) {
-    // Avoid considering the same callee of this function multiple times by
-    // recording known callees.
-    std::set<uint32_t> known_callees;
-    // Consider every function call instruction in every block.
-    for (auto& block : function) {
-      for (auto& instruction : block) {
-        if (instruction.opcode() != SpvOpFunctionCall) {
-          continue;
-        }
-        // Get the id of the function being called.
-        uint32_t callee = instruction.GetSingleWordInOperand(0);
-
-        // Get the loop nesting depth of this function call.
-        uint32_t loop_nesting_depth =
-            context->GetStructuredCFGAnalysis()->LoopNestingDepth(block.id());
-        // If inside a loop header, consider the function call nested inside the
-        // loop headed by the block.
-        if (block.IsLoopHeader()) {
-          loop_nesting_depth++;
-        }
-
-        // Update the map if we have not seen this pair (caller, callee)
-        // before or if this function call is from a greater depth.
-        if (!known_callees.count(callee) ||
-            call_to_max_depth[{function.result_id(), callee}] <
-                loop_nesting_depth) {
-          call_to_max_depth[{function.result_id(), callee}] =
-              loop_nesting_depth;
-        }
-
-        if (known_callees.count(callee)) {
-          // We have already considered a call to this function - ignore it.
-          continue;
-        }
-        // Increase the callee's in-degree and add an edge to the call graph.
-        function_in_degree_[callee]++;
-        call_graph_edges_[function.result_id()].insert(callee);
-        // Mark the callee as 'known'.
-        known_callees.insert(callee);
-      }
-    }
-  }
+  // Compute |function_in_degree_|, |call_graph_edges_| and |call_to_max_depth|.
+  BuildGraphAndGetDepthOfFunctionCalls(context, &call_to_max_depth);
 
   // Compute |functions_in_topological_order_|.
   ComputeTopologicalOrderOfFunctions();
 
-  // Find the maximum loop nesting depth that each function can be called from
-  // by considering them in topological order.
-  for (uint32_t function_id : functions_in_topological_order_) {
-    const auto& callees = call_graph_edges_[function_id];
-
-    // For each callee, update its maximum loop nesting depth, if a call from
-    // |function_id| increases it.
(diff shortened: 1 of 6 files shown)


source/opt/struct_cfg_analysis.h
```diff
@@ -74,6 +74,9 @@ class StructuredCFGAnalysis {
   // construct.
   uint32_t LoopContinueBlock(uint32_t bb_id);
 
+  // Returns the loop nesting depth of |bb_id|.
```
afd, line 77: I think it would be good to elaborate slightly more on what this means, and be clear that it is the depth with respect to the current function.

"Returns the loop nesting depth of |bb_id| within its function.  That is, the number of loop constructs in which |bb_id| is contained.  Note that, as per other functions in StructuredCFGAnalysis, a loop header is *not* regarded as being part of the loop that it heads, so that e.g. the nesting depth of an outer-most loop header is 0."

source/opt/struct_cfg_analysis.cpp
```diff
@@ -150,6 +150,19 @@ uint32_t StructuredCFGAnalysis::LoopContinueBlock(uint32_t bb_id) {
   return merge_inst->GetSingleWordInOperand(kContinueNodeIndex);
 }
 
+uint32_t StructuredCFGAnalysis::LoopNestingDepth(uint32_t bb_id) {
+  uint32_t depth = 0;
```
afd, line 154: Up to you, but I like to use `result` for the name of the thing that will ultimately be returned by a method.

source/fuzz/call_graph.h
```diff
@@ -43,6 +43,18 @@ class CallGraph {
...
   std::set<uint32_t> GetIndirectCallees(uint32_t function_id) const;
 
+  // Returns the ids of all the functions in the graph in a topological order,
+  // in relation to the function calls, which are assumed to be recursion-free.
+  const std::vector<uint32_t>& GetFunctionsInTopologicalOrder() const {
+    return functions_in_topological_order_;
+  }
+
+  // Returns the maximum loop nesting depth from which |function_id| can be
+  // called from.
```
afd, line 53: "from which |function_id| can be called from." -> "from which |function_id| can be called."

Also be clear on the fact that this is inter-procedural, and be clear that it's a static analysis, i.e. the reported nesting depth might not actually be possible at runtime.

afd: Can you please augment the top-level comment for this class with two things:

(1) The fact that the module is assumed to be recursion-free
(2) The fact that the call graph is immutable, and thus needs to be computed afresh if the module changes

source/fuzz/call_graph.cpp
```diff
@@ -20,12 +20,19 @@ namespace spvtools {
...
+    function_max_loop_nesting_depth_[function.result_id()] = 0;
   }
 
+  // Record the maximum loop nesting depth for each edge, by keeping a map from
+  // pairs of function ids, where (A, B) represents a function call from A to B,
+  // to the corresponding maximum depth.
+  std::map<std::pair<uint32_t, uint32_t>, uint32_t> call_to_max_depth;
+
   // Consider every function.
   for (auto& function : *context->module()) {
```
afd, line 37: Up to you, but it might be nice to extract this whole loop into a private helper function with an appropriate name, as well as the code that comes after the call to ComputeTopologicalOrderOfFunctions.

That way the constructor would be fairly short and would consist mainly of calls to individually cohesive helpers.

source/fuzz/call_graph.h
```diff
@@ -54,6 +66,19 @@ class CallGraph {
...
   std::map<uint32_t, uint32_t> function_in_degree_;
+
+  // Stores the ids of the functions in a topological order,
+  // in relation to the function calls, which are assumed to be recursion-free.
+  std::vector<uint32_t> functions_in_topological_order_;
+
+  // Computes a topological order of the functions in the graph, writing the
+  // result to |functions_in_topological_order_|. The function calls are assumed
+  // to be recursion-free.
+  void ComputeTopologicalOrderOfFunctions();
```
afd, line 77: See Google C++ style: https://google.github.io/styleguide/cppguide.html#Declaration_Order

We prefer to have methods before data members.

afd (changes requested):
This looks excellent!  I have suggested some small changes.

Contrary to my previous advice, I think it would be good to separate this into two PRs, with the first PR contributing the changes to structured CFG analysis.  I would like @alan-baker or @s-perron to review those, and I think it would make sense for them to review the parts that affect opt in a PR that doesn't also affect fuzz.

stefanomil:
@afd I have opened #3754 for the spirv-opt part. This PR will be for spirv-fuzz

source/opt/struct_cfg_analysis.h
stefanomil: Done (also in #3754)

source/opt/struct_cfg_analysis.cpp
stefanomil: Done (also in #3754 )

source/fuzz/call_graph.h
stefanomil: Done

source/fuzz/call_graph.h
stefanomil: Thanks, I moved method declarations to before data members

source/fuzz/call_graph.cpp
stefanomil: Done

afd (approved):
Thanks for making the changes (and for splitting this into multiple PRs).

PR #3754, opened 2020-08-27
spirv-opt: Add function to compute loop nesting depth of a block in StructuredCFGAnalysis

stefanomil:
This PR adds the LoopNestingDepth function to StructuredCFGAnalysis.
This function, given a block id, returns the loop nesting depth of
such block within its function (the number of loop constructs
containing it).


afd (commented):
I am happy with this addition.

@s-perron could you also review?

It's part of something we're doing in spirv-fuzz, but it makes sense for this functionality to be added to StructuredCFGAnalaysis I think.

PR #3756, opened 2020-08-27
spirv-val: Add Vulkan VUID labels to BuiltIn

sfricke-samsung:
The Vulkan 1.2.152 headers/spec now include Valid Usage IDs (VUID)
for every BuiltIn. This change adds labeling to help aid tracking
the coverage gap in Vulkan targeted validation. This is done with
a script in the Validation Layers that parses the source for VUID
strings, both that there is an implementation and a test for it.

There are 2 main changes:

1. A VUID string is added where applicable. It is wrapped in a function
that at runtimes checks if Vulkan is the target so that other targets
will not be effected as the output error only changes for Vulkan and
the overhead only occurs if there is an error at all.

2. For unit test, a new parameter value was added to allow make sure
that for each VUID there is a matching test. There are cases where the
same unit test checks multiple VUs via the parameter feature of gtest.
For this, I added a custom gMock Matcher to simply loop over a list
of VUID strings

### DIFF
--- a/source/val/validate_builtins.cpp
+++ b/source/val/validate_builtins.cpp
@@ -365,7 +365,7 @@ class BuiltInsValidator {
   // |referenced_from_inst| - instruction which references id defined by
   //                          |referenced_inst| from within a function.
   spv_result_t ValidateNotCalledWithExecutionModel(
-      const char* comment, SpvExecutionModel execution_model,
+      std::string comment, SpvExecutionModel execution_model,
       const Decoration& decoration, const Instruction& built_in_inst,
       const Instruction& referenced_inst,
       const Instruction& referenced_from_inst);
@@ -864,7 +864,7 @@ spv_result_t BuiltInsValidator::ValidateF32ArrHelper(
 }
 
 spv_result_t BuiltInsValidator::ValidateNotCalledWithExecutionModel(
-    const char* comment, SpvExecutionModel execution_model,
+    std::string comment, SpvExecutionModel execution_model,
     const Decoration& decoration, const Instruction& built_in_inst,
     const Instruction& referenced_inst,
     const Instruction& referenced_from_inst) {
@@ -903,6 +903,7 @@ spv_result_t BuiltInsValidator::ValidateClipOrCullDistanceAtReference(
     const Decoration& decoration, const Instruction& built_in_inst,
     const Instruction& referenced_inst,
     const Instruction& referenced_from_inst) {
+  uint32_t operand = decoration.params()[0];
   if (spvIsVulkanEnv(_.context()->target_env)) {
     const SpvStorageClass storage_class = GetStorageClass(referenced_from_inst);
     if (storage_class != SpvStorageClassMax &&
@@ -911,7 +912,7 @@ spv_result_t BuiltInsValidator::ValidateClipOrCullDistanceAtReference(
       return _.diag(SPV_ERROR_INVALID_DATA, &referenced_from_inst)
              << "Vulkan spec allows BuiltIn "
              << _.grammar().lookupOperandName(SPV_OPERAND_TYPE_BUILT_IN,
-                                              decoration.params()[0])
+                                              operand)
              << " to be only used for variables with Input or Output storage "
                 "class. "
              << GetReferenceDesc(decoration, built_in_inst, referenced_inst,
@@ -949,7 +950,12 @@ spv_result_t BuiltInsValidator::ValidateClipOrCullDistanceAtReference(
                   decoration, built_in_inst, /* Any number of components */ 0,
                   [this, &decoration, &referenced_from_inst](
                       const std::string& message) -> spv_result_t {
+                    uint32_t vuid =
+                        (decoration.params()[0] == SpvBuiltInClipDistance)
+                            ? 4191
+                            : 4200;
                     return _.diag(SPV_ERROR_INVALID_DATA, &referenced_from_inst)
+                           << _.VkErrorID(vuid)
                            << "According to the Vulkan spec BuiltIn "
                            << _.grammar().lookupOperandName(
                                   SPV_OPERAND_TYPE_BUILT_IN,
@@ -971,8 +977,13 @@ spv_result_t BuiltInsValidator::ValidateClipOrCullDistanceAtReference(
                     decoration, built_in_inst, /* Any number of components */ 0,
                     [this, &decoration, &referenced_from_inst](
                         const std::string& message) -> spv_result_t {
+                      uint32_t vuid =
+                          (decoration.params()[0] == SpvBuiltInClipDistance)
+                              ? 4191
+                              : 4200;
                       return _.diag(SPV_ERROR_INVALID_DATA,
                                     &referenced_from_inst)
+                             << _.VkErrorID(vuid)
                              << "According to the Vulkan spec BuiltIn "
                              << _.grammar().lookupOperandName(
                                     SPV_OPERAND_TYPE_BUILT_IN,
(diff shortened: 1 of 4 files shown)


source/val/validate_builtins.cpp
```diff
@@ -2554,12 +2651,21 @@ spv_result_t BuiltInsValidator::ValidateComputeShaderI32Vec3InputAtDefinition(
             decoration, inst, 3,
             [this, &decoration,
              &inst](const std::string& message) -> spv_result_t {
+              uint32_t operand = decoration.params()[0];
+              const char* vuid =
```
alan-baker, line 2655: At this point I find it much easier to read as a switch statement.

source/val/validation_state.cpp
```diff
@@ -1319,5 +1319,9 @@ bool ValidationState_t::IsValidStorageClass(
   return true;
 }
 
+std::string ValidationState_t::VkErrorID(std::string id) const {
```
alan-baker, line 1322: I think this should just take a uint32_t and construct the message here.

```
uint32_t ValidationState_t::VkErrorID(uint32_t id) const {
  if (!spvIsVulkanEnv(context_->target_env))
    return "";

  switch (id) {
    case 4425:
      return "VUID-WorkgroupSize-WorkgroupSize-04425";
    ...
    case 0: // an invalid id.
    default:
     return "";
}
```

source/val/validate_builtins.cpp
```diff
@@ -2577,12 +2683,21 @@ spv_result_t BuiltInsValidator::ValidateComputeShaderI32Vec3InputAtReference(
     const Decoration& decoration, const Instruction& built_in_inst,
     const Instruction& referenced_inst,
     const Instruction& referenced_from_inst) {
+  uint32_t operand = decoration.params()[0];
   if (spvIsVulkanOrWebGPUEnv(_.context()->target_env)) {
     const SpvStorageClass storage_class = GetStorageClass(referenced_from_inst);
     if (storage_class != SpvStorageClassMax &&
         storage_class != SpvStorageClassInput) {
+      const char* vuid =
```
alan-baker, line 2691: same here

source/val/validate_builtins.cpp
```diff
@@ -2599,7 +2714,16 @@ spv_result_t BuiltInsValidator::ValidateComputeShaderI32Vec3InputAtReference(
       bool has_webgpu_model = execution_model == SpvExecutionModelGLCompute;
       if ((spvIsVulkanEnv(_.context()->target_env) && !has_vulkan_model) ||
           (spvIsWebGPUEnv(_.context()->target_env) && !has_webgpu_model)) {
+        const char* vuid =
```
alan-baker, line 2717: and here

source/val/validation_state.h
```diff
@@ -707,6 +707,10 @@ class ValidationState_t {
   // Validates the storage class for the target environment.
   bool IsValidStorageClass(SpvStorageClass storage_class) const;
 
+  // Takes a Vulkan Valid Usage ID (VUID) and return the string only if
+  // targeting Vulkan
+  std::string VkErrorID(std::string id) const;
```
alan-baker, line 712: This deserves a longer comment about how to find VUIDs in various places (e.g. html spec or from Vulkan-Docs).

source/val/validation_state.cpp
sfricke-samsung: 1. my worry is the switch statement becomes this awful 200 case long, the indirect lookup table
2. It is technically possible to have `VUID-x-0001` and `VUID-y-0001` if both are the same VU in the Vulkan spec using the "common validity" generation. Not an issue for the Built-in, but worry about when one case of that happens

sfricke-samsung:
> Who is responsible for updating for the VUIDs if they change in the spec?
>
> Are you signing up to add the VUIDs for everything we check under the Vulkan environment?

Currently, I don't see many changes in the built-in occurring in the Vulkan spec, I foresee additions, but that's not an issue

My idea is that all this tracking can be done outside of SPIR-V tool since the [VVL has a script](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/2141), which I plan to modify, has all the infrastructure already. The main goal of this PR/effort is to "track validation coverage" and that is going to be done atm in the script in the PR. If the scripts does realize that ToT SPIR-V Tools has an outdated VUID, I would imagine it would print an alert for someone (which I am willing to do for the foreseeable future) to come here and make a PR to update it

source/val/validation_state.h
sfricke-samsung: fair, I wanted to put a "here is the long detailed explanation" somewhere, but wasn't sure where, if you think this declaration is the spot to add it, happy to append this

source/val/validation_state.cpp
alan-baker: To 1, there are solutions. The speed is mostly irrelevant since we're only doing this if we're stopping because we found an error.

To 2, that sounds like a bad design, but I can't really change it. Mostly I would like the code to be briefer as these are rather long (and seemingly redundant) tags.

sfricke-samsung: 1. agree

2. The reason is for things like all the `vkCmdDraw*` commands it was easier to write the VU once and was simpler for the VU generation logic to us the same number as the string was the unique part, not the number... but again, out of scope

Thoughts on

```
uint32_t ValidationState_t::VkErrorID(uint32_t id, const char* reference = nullptr) const;
```

This way if there is a overlap of the id number, there is something it if/else in the switch case

alan-baker: The extra parameter seems reasonable. We can always revisit this later.

validation_info.json
```diff
@@ -0,0 +1,15 @@
+{
```
sfricke-samsung, line 1: @alan-baker This file is needed in case the file locations would ever change it would be a pain to have to track various locations to look. Debated about putting file in `utils` folder, but figured if put at root there will never be a guess where this file is (which is important)

alan-baker: It seems like we know which VUIDs are checked by looking solely at validation_state.cpp. I'm not clear on the use of this file.

sfricke-samsung: If for whatever reason that location changes, then the Validation Layers have to keep a list of every location that it might have been

alan-baker: It seems like the name of the file should be `vulkan_validation_info.json` at least. Also, I'm not clear on why the test files are listed. What is the validation layer script doing exactly?

sfricke-samsung: 1. I don't know the plan for WebGPU/CL/futureApis are and if it would care for something similar, so that's what I have the `target : Vulkan` section... happy to rename the file, but then in the future there should be a `api_validation_info.json` if someone else does need one

2.  [Here is the current code](https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/2145/files#diff-d648cc0d8fa02012cc79869be6d282e5R762). When you run the Validation Layer stats script it looks from non-commented strings in the code and then builds this [database](https://vulkan.lunarg.com/doc/sdk/1.2.148.1/windows/validation_error_database.html) of info of what is done and what isn't. The reason for the test is it nice to know if there is an actual test testing the validation

So sorry in making you become a Vulkan validation expert with this PR, but hopefully this answers your questions

dj2: Wouldn't it be better to add this file in the event we move the strings in the future?

alan-baker: I think Dan meant, wait to add this file until it is required and I tend to agree with that.

sfricke-samsung: I assume that the 2 paths I have listed in the file currently won't change unless there is a large refactor of the repo, but if that event occurs, feel adding a file then doesn't do much as there is always this "need to check the range from first adding VUs to the file" check that will always occur

Regardless, thanks for weighing in, will remove file, hopefully everything is else is good to go otherwise

source/val/validation_state.cpp
```diff
@@ -1319,5 +1319,183 @@ bool ValidationState_t::IsValidStorageClass(
...
+#define VUID_WRAP(vuid) "[" #vuid "] "
+
+std::string ValidationState_t::VkErrorID(uint32_t id,
+                                         const char* reference) const {
+  if (!spvIsVulkanEnv(context_->target_env)) {
+    return "";
+  }
+
+  // Currently no 2 VUID shared id so no need for reference
+  (void)reference;
```
alan-baker, line 1331: Nit: Add this comment at the start of the function and remove the parameter name.

PR #3767, opened 2020-09-01
spirv-fuzz: Fix GetIdEquivalenceClasses

### LINKED ISSUE #3761: spirv-fuzz: Invalid variant shader
A variant shader was invalidated due an issue in `FuzzerPassAddOpPhiSynonyms`.
Debugging `spirv-fuzz`, I found that, in `FuzzerPassAddOpPhiSynonyms::GetIdEquivalenceClasses` at line 165, `pair` represents an `OpFunction` instruction with float return type. But this instruction can not be synonymous with anything. In this case the synonym to be considered is an `OpFunctionCall` instruction that calls `pair.first`.

```
error: line 113: Invalid use of function result id 10[%fx_f1_].
  %262 = OpPhi %float %fx_f1_ %56
```

stefanomil:
FuzzerPassAddOpPhiSynonyms::GetIdEquivalenceClasses considers every id
with type among Bool, Integer, Float, Vector, Matrix, Array, Struct,
Pointer (if VariablePointers is enabled) to find sets of potential
synonyms.

However, some instructions with these types cannot be used in an OpPhi:
- OpFunction cannot be used as a value
- OpUndef should not be used, because it yields an undefined value for
  each use

Fixes #3761 .

### DIFF
--- a/source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
+++ b/source/fuzz/fuzzer_pass_add_opphi_synonyms.cpp
@@ -167,6 +167,12 @@ FuzzerPassAddOpPhiSynonyms::GetIdEquivalenceClasses() {
       continue;
     }
 
+    // Exclude OpFunction and OpUndef instructions.
+    if (pair.second->opcode() == SpvOpFunction ||
+        pair.second->opcode() == SpvOpUndef) {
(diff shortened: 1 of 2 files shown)


afd (approved):
LGTM.

PR #3770, opened 2020-09-02
spirv-opt: Extend SetResultType and SetResultId functions

### LINKED ISSUE #3769: spirv-opt: Extend SetResultType and SetResultId functions in the Instruction class
The `SetResultType` and `SetResultId` functions, in the `Instruction` class, do not handle the case where the result type and result id are not yet in `operands_`. Therefore, the functions can be extended to cover that case.

andreperezmaselco:
Fixes #3769.

### DIFF
--- a/source/opt/instruction.h
+++ b/source/opt/instruction.h
@@ -648,27 +648,26 @@ inline void Instruction::SetInOperands(OperandList&& new_operands) {
   operands_.insert(operands_.end(), new_operands.begin(), new_operands.end());
 }
 
-inline void Instruction::SetResultId(uint32_t result_id) {
+inline void Instruction::SetResultId(uint32_t res_id) {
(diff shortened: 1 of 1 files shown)


andreperezmaselco:
@s-perron and @alan-baker, could you review it?

s-perron (changes requested):
I've added a general comment on the strategy to the issue.  I think this is unsafe behaviour, and too easy for some one to make a mistake.

afd:
I couldn't see @s-perron's comment, but I'm also not keen on this change.  @andreperezmaselco - I think you wanted it to make one of your spirv-fuzz PRs work a bit differently regarding how instructions are created, but as we discussed I'd prefer you to avoid that in the spirv-fuzz PR in any case.  Would you thus be OK to close this (and associated issue) @andreperezmaselco?

andreperezmaselco:
OK, @afd. I'll close them.

PR #3771, opened 2020-09-03
spirv-opt: Add function to compute nesting depth of a block

stefanomil:
This PR adds the NestingDepth function to StructuredCFGAnalysis.
This function, given a block id, returns the number of merge
constructs containing it.

This is needed by spirv-fuzz, but it makes sense to add it to
StructuredCFGAnalaysis, which contains related functionalities.


afd (approved):
LGTM.

PR #3775, opened 2020-09-04
spirv-fuzz: Add bit instruction synonym transformation

andreperezmaselco:
This PR implements part of the add bit instruction synonym transformation.
For now, the implementation covers the `OpBitwiseOr`, `OpBitwiseXor` and
`OpBitwiseAnd` cases.

### DIFF
--- a/source/fuzz/fuzzer_pass_add_bit_instruction_synonyms.cpp
+++ b/source/fuzz/fuzzer_pass_add_bit_instruction_synonyms.cpp
@@ -41,8 +41,13 @@ void FuzzerPassAddBitInstructionSynonyms::Apply() {
           continue;
         }
 
-        // |instruction| must be a bit instruction.
-        if (!spvOpcodeIsBit(instruction.opcode())) {
+        // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3557):
+        //  Right now we only support certain operations. When this issue is
+        //  addressed the following conditional can use the function
+        //  |spvOpcodeIsBit|.
+        if (instruction.opcode() != SpvOpBitwiseOr &&
+            instruction.opcode() != SpvOpBitwiseXor &&
+            instruction.opcode() != SpvOpBitwiseAnd) {
           continue;
         }
 
@@ -64,7 +69,7 @@ void FuzzerPassAddBitInstructionSynonyms::Apply() {
           FindOrCreateIntegerConstant({i}, 32, false, false);
         }
 
-        // Tries to apply the transformation.
+        // Applies the add bit instruction synonym transformation.
         ApplyTransformation(TransformationAddBitInstructionSynonym(
             instruction.result_id(),
             GetFuzzerContext()->GetFreshIds(
--- a/source/fuzz/fuzzer_pass_add_bit_instruction_synonyms.cpp
+++ b/source/fuzz/fuzzer_pass_add_bit_instruction_synonyms.cpp
@@ -1,0 +1,79 @@
+// Copyright (c) 2020 André Perez Maselco
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "source/fuzz/fuzzer_pass_add_bit_instruction_synonyms.h"
+
+#include "source/fuzz/fuzzer_util.h"
+#include "source/fuzz/instruction_descriptor.h"
+#include "source/fuzz/transformation_add_bit_instruction_synonym.h"
+
+namespace spvtools {
+namespace fuzz {
+
+FuzzerPassAddBitInstructionSynonyms::FuzzerPassAddBitInstructionSynonyms(
+    opt::IRContext* ir_context, TransformationContext* transformation_context,
+    FuzzerContext* fuzzer_context,
+    protobufs::TransformationSequence* transformations)
+    : FuzzerPass(ir_context, transformation_context, fuzzer_context,
+                 transformations) {}
+
+FuzzerPassAddBitInstructionSynonyms::~FuzzerPassAddBitInstructionSynonyms() =
+    default;
+
+void FuzzerPassAddBitInstructionSynonyms::Apply() {
+  for (auto& function : *GetIRContext()->module()) {
+    for (auto& block : function) {
+      for (auto& instruction : block) {
+        // Randomly decides whether the transformation will be applied.
+        if (!GetFuzzerContext()->ChoosePercentage(
+                GetFuzzerContext()->GetChanceOfAddingBitInstructionSynonym())) {
+          continue;
+        }
+
+        // |instruction| must be a bit instruction.
+        if (!spvOpcodeIsBit(instruction.opcode())) {
+          continue;
+        }
+
+        // Right now, only integer operands are supported.
+        if (GetIRContext()
+                ->get_type_mgr()
+                ->GetType(instruction.type_id())
+                ->AsVector()) {
+          continue;
+        }
+
+        // Make sure all bit indexes are defined as 32-bit unsigned integers.
+        uint32_t width = GetIRContext()
+                             ->get_type_mgr()
+                             ->GetType(instruction.type_id())
+                             ->AsInteger()
+                             ->width();
+        for (uint32_t i = 0; i < width; i++) {
+          FindOrCreateIntegerConstant({i}, 32, false, false);
+        }
+
+        // Tries to apply the transformation.
+        ApplyTransformation(TransformationAddBitInstructionSynonym(
+            instruction.result_id(),
+            GetFuzzerContext()->GetFreshIds(
+                TransformationAddBitInstructionSynonym::GetRequiredFreshIdCount(
+                    GetIRContext(), &instruction))));
+      }
+    }
+  }
+}
+
+}  // namespace fuzz
+}  // namespace spvtools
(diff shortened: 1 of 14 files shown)


source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -455,6 +456,30 @@ message TransformationAccessChain {
 
 }
 
+message TransformationAddBitInstructionSynonym {
```
afd, line 459: Can we change "Bit" to "Bitwise"?

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -455,6 +456,30 @@ message TransformationAccessChain {
...
+
+  // A transformation that adds synonyms for bit instructions by evaluating each pair of bits.
```
afd, line 461: "bit" -> "bitwise"

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -455,6 +456,30 @@ message TransformationAccessChain {
...
+
+  // The bit instruction result id.
```
afd, line 463: Same (and similar elsewhere - please change the names of the transformation and fuzzer pass classes too).

source/fuzz/protobufs/spvtoolsfuzz.proto
afd, line 461: Can you give a bit more detail?  A small example of how this would work on e.g. two 4-bit integers would be good.

source/opcode.cpp
```diff
@@ -731,3 +731,16 @@ bool spvOpcodeIsAccessChain(SpvOp opcode) {
       return false;
   }
 }
+
+bool spvOpcodeIsBit(SpvOp opcode) {
```
afd, line 735: Change name to spvOpcodeIsBitwise.

source/opcode.cpp
```diff
@@ -731,3 +731,16 @@ bool spvOpcodeIsAccessChain(SpvOp opcode) {
...
+  switch (opcode) {
+    // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3557):
+    // Right now, only certain operations are supported.
+    case SpvOpBitwiseOr:
+    case SpvOpBitwiseXor:
+    case SpvOpBitwiseAnd:
+      return true;
+    default:
+      return false;
+  }
```
afd, line 745: I don't feel comfortable merging this in a TODO state, as there aren't that many bitwise operations and someone might decide to use this function without looking at the implementation (assuming it works for all of them).

Would you be happy to go ahead and get this working for all opcodes?

An alternative is to have this function handle all opcodes, but then in spirv-fuzz have a TODO to explain that the transformation doesn't yet support all of them.

source/opcode.h
```diff
@@ -134,17 +134,20 @@ bool spvOpcodeIsDebug(SpvOp opcode);
...
 
 // Returns a vector containing the indices of the memory semantics <id>
 // operands for |opcode|.
 std::vector<uint32_t> spvOpcodeMemorySemanticsOperandIndices(SpvOp opcode);
 
-// Returns true for opcodes that represents access chain instructions.
+// Returns true for opcodes that represent access chain instructions.
 bool spvOpcodeIsAccessChain(SpvOp opcode);
 
+// Returns true for opcodes that represent bit instructions.
```
afd, line 150: bitwise.

source/fuzz/transformation_add_bit_instruction_synonym.h
```diff
@@ -0,0 +1,64 @@
...
+class TransformationAddBitInstructionSynonym : public Transformation {
+ public:
+  explicit TransformationAddBitInstructionSynonym(
+      const protobufs::TransformationAddBitInstructionSynonym& message);
+
+  TransformationAddBitInstructionSynonym(
+      const uint32_t instruction_result_id,
+      const std::vector<uint32_t>& fresh_ids);
+
+  // - |message_.instruction_result_id| must be a bit instruction.
```
afd, line 35: bitwise

source/fuzz/transformation_add_bit_instruction_synonym.h
```diff
@@ -0,0 +1,64 @@
...
+  // - |message_.fresh_ids| must be fresh ids needed to apply the
+  // transformation.
```
afd, line 37: Indent "transformation" to be in line with "|message_"

source/fuzz/transformation_add_bit_instruction_synonym.cpp
```diff
@@ -0,0 +1,220 @@
...
+    const {
+  protobufs::Transformation result;
+  *result.mutable_add_bit_instruction_synonym() = message_;
+  return result;
+}
+
+uint32_t TransformationAddBitInstructionSynonym::GetRequiredFreshIdCount(
+    opt::IRContext* ir_context, opt::Instruction* bit_instruction) {
+  // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3557):
+  // Right now, only certain operations are supported.
```
afd, line 113: Add one space before "Right" so that it is slightly indented after "TODO" (this means that it shows up in CLion's highlighting as part of the TODO).

source/fuzz/transformation_add_bit_instruction_synonym.cpp
```diff
@@ -0,0 +1,220 @@
...
+    return false;
+  }
+
+  // Right now, only integer operands are supported.
+  if (ir_context->get_type_mgr()->GetType(instruction->type_id())->AsVector()) {
+    return false;
+  }
+
+  // TODO: This condition could be relaxed if the index exists as another
+  // integer type. All bit indexes must be defined as 32-bit unsigned integers.
```
afd, line 54: We avoid raw TODOs in the project, so please open an issue for extending this to other types, and turn it into TODO(issue URL).

source/fuzz/fuzzer_pass_add_bit_instruction_synonyms.cpp
```diff
@@ -0,0 +1,79 @@
...
+        uint32_t width = GetIRContext()
+                             ->get_type_mgr()
+                             ->GetType(instruction.type_id())
+                             ->AsInteger()
+                             ->width();
+        for (uint32_t i = 0; i < width; i++) {
+          FindOrCreateIntegerConstant({i}, 32, false, false);
+        }
+
+        // Tries to apply the transformation.
```
afd, line 67: I wouldn't say "Tries", because this is the way we add transformations when they are guaranteed to be applicable by construction.

afd (changes requested):
Looks good!  Some changes requested - and in particular I wonder how hard it would be to handle all the bitwise operations in this PR?

source/opcode.cpp
andreperezmaselco: Done.

source/fuzz/transformation_add_bit_instruction_synonym.h
andreperezmaselco: Done.

source/fuzz/transformation_add_bit_instruction_synonym.cpp
andreperezmaselco: Done.

source/fuzz/fuzzer_pass_add_bit_instruction_synonyms.cpp
andreperezmaselco: Done.

source/fuzz/transformation_add_bit_instruction_synonym.cpp
andreperezmaselco: Done.

source/fuzz/protobufs/spvtoolsfuzz.proto

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -455,6 +456,107 @@ message TransformationAccessChain {
...
+  // |                                             | %29 = OpBitFieldUExtract %2 %6 %12 %6 ; extracts bit 4 from %6 |
+  // |                                             | %30 = OpBitwiseOr %2 %28 %29                                   |
+  // |                                             |                                                                |
+  // |                                             | %31 = OpBitFieldUExtract %2 %5 %13 %6 ; extracts bit 5 from %5 |
+  // |                                             | %32 = OpBitFieldUExtract %2 %6 %13 %6 ; extracts bit 5 from %6 |
+  // |                                             | %33 = OpBitwiseOr %2 %31 %32                                   |
+  // |                                             |                                                                |
+  // |                                             | %34 = OpBitFieldUExtract %2 %5 %14 %6 ; extracts bit 6 from %5 |
+  // |                                             | %35 = OpBitFieldUExtract %2 %6 %14 %6 ; extracts bit 6 from %6 |
+  // |                                             | %36 = OpBitwiseOr %2 %34 %35                                   |
```
afd, line 519: The example is great but is a bit long and might be annoying for folks browsing the file looking for particular protobuf messages.

How about you replace all this with:

```
// |                                       |  ...                         |
```

That would make it clear that the theme continues similarly.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -455,6 +456,107 @@ message TransformationAccessChain {
...
+  // |                                             | %37 = OpBitFieldUExtract %2 %5 %15 %6 ; extracts bit 7 from %5 |
+  // |                                             | %38 = OpBitFieldUExtract %2 %6 %15 %6 ; extracts bit 7 from %6 |
+  // |                                             | %39 = OpBitwiseOr %2 %37 %38                                   |
+  // |                                             |                                                                |
+  // |                                             | %40 = OpBitFieldInsert %2 %18 %21 %6 %6 ; inserts bit 1        |
+  // |                                             | %41 = OpBitFieldInsert %2 %40 %24 %10 %6 ; inserts bit 2       |
+  // |                                             | %42 = OpBitFieldInsert %2 %41 %27 %11 %6 ; inserts bit 3       |
+  // |                                             | %43 = OpBitFieldInsert %2 %42 %30 %12 %6 ; inserts bit 4       |
+  // |                                             | %44 = OpBitFieldInsert %2 %43 %33 %13 %6 ; inserts bit 5       |
+  // |                                             | %45 = OpBitFieldInsert %2 %44 %36 %14 %6 ; inserts bit 6       |
```
afd, line 530: You could also replace this with a `...`.

afd: An alternative (or you could do it as well) would be to move the example to a comment in the header file of the transformation class, and remark here that the reader should see there for an example.

afd (commented):
A suggestion on how to make the example a bit more space efficient, but otherwise looks good.

Let me know what you decide to do here, and rebase, and then I can merge.

afd (approved):
LGTM.

afd:
@andreperezmaselco apologies but can you rebase again please?

PR #3781, opened 2020-09-08
Improve spirv-fuzz CMake code

paulthomson:
This change improves spirv-fuzz CMake code to be more compatible with other projects that might want to include spirv-fuzz as a sub-project.

* Add a CMake option for building spirv-fuzz. 
* We now check if protobuf targets are already available. 
* We no longer specify `-DGOOGLE_PROTOBUF_NO_RTTI -DGOOGLE_PROTOBUF_USE_UNALIGNED=0`; a newer version of protobuf does not require this. Note that we probably should have specified this for protobuf targets as well, but this is no longer needed. 
* Updated protobuf version in Kokoro scripts and README.md.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,8 @@ if(SPIRV_BUILD_COMPRESSION)
           "Please remove SPIRV_BUILD_COMPRESSION from your build options.")
 endif(SPIRV_BUILD_COMPRESSION)
 
+option(SPIRV_BUILD_FUZZER "Build spirv-fuzz" OFF)
+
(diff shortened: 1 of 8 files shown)


afd (approved):
This looks excellent.

Before you merge, can you update the top-level README under "Note about the fuzzer", with the improved command for cloning the protobuf dependency?

paulthomson:
> This looks excellent.
> 
> Before you merge, can you update the top-level README under "Note about the fuzzer", with the improved command for cloning the protobuf dependency?

I believe I already did that.

afd:
Thanks - sorry, must have missed that in my review!

Issue #3779, opened 2020-09-08
Removing SpvDecorationUserTypeGOOGLE in StripReflectInfoPass 


TheMostDiligent:
Hello!

We are trying to use shaders compiled from HLSL by DXC and are getting the following validation error:

```
SPIR-V module not valid: 2nd operand of DecorateString: operand UserTypeGOOGLE(5636) requires one of these extensions: SPV_GOOGLE_user_type
```

We made the following changes to [StripReflectInfoPass](https://github.com/KhronosGroup/SPIRV-Tools/blob/dd37d73c5ea01c7f401b225dc15e143e502513a3/source/opt/strip_reflect_info_pass.cpp#L26) to remove the offending operands and everything seems to be working fine now. Lines [35](https://github.com/KhronosGroup/SPIRV-Tools/blob/dd37d73c5ea01c7f401b225dc15e143e502513a3/source/opt/strip_reflect_info_pass.cpp#L35) and [43](https://github.com/KhronosGroup/SPIRV-Tools/blob/dd37d73c5ea01c7f401b225dc15e143e502513a3/source/opt/strip_reflect_info_pass.cpp#43)

```cpp
      case SpvOpDecorateStringGOOGLE:
        if (inst.GetSingleWordInOperand(1) == SpvDecorationHlslSemanticGOOGLE ||
            inst.GetSingleWordInOperand(1) == SpvDecorationUserTypeGOOGLE) {
          to_remove.push_back(&inst);
        } else {
          other_uses_for_decorate_string = true;
        }
        break;

      case SpvOpMemberDecorateStringGOOGLE:
        if (inst.GetSingleWordInOperand(2) == SpvDecorationHlslSemanticGOOGLE ||
            inst.GetSingleWordInOperand(2) == SpvDecorationUserTypeGOOGLE) {
          to_remove.push_back(&inst);
        } else {
          other_uses_for_decorate_string = true;
        }
        break;
```

[Line 67](https://github.com/KhronosGroup/SPIRV-Tools/blob/dd37d73c5ea01c7f401b225dc15e143e502513a3/source/opt/strip_reflect_info_pass.cpp#L67):

```cpp
    if (0 == std::strcmp(ext_name, "SPV_GOOGLE_hlsl_functionality1")) {
      to_remove.push_back(&inst);
    } else if (0 == std::strcmp(ext_name, "SPV_GOOGLE_user_type")) {
      to_remove.push_back(&inst);
[...]

PR #3783, opened 2020-09-08
StripReflectInfoPass: removing SpvDecorationUserTypeGOOGLE (fixes https://github.com/KhronosGroup/SPIRV-Tools/issues/3779)

TheMostDiligent:
This fix removes SpvDecorationUserTypeGOOGLE operands when stripping reflection info. See https://github.com/KhronosGroup/SPIRV-Tools/issues/3779


s-perron:
@TheMostDiligent Will you be able to add tests?

TheMostDiligent:
@s-perron Can you point me to an example of the required test?

s-perron:
This is a good example of the test.  Provide a small spir-v module that has the instructions that should be removed, and check that they are not there in the output.

https://github.com/KhronosGroup/SPIRV-Tools/blob/master/test/opt/strip_reflect_info_test.cpp#L136

PR #3784, opened 2020-09-09
spirv-fuzz: Fix header guards in transformations/fuzzer passes

antnkarp:
I have corrected header guards in my transformations and fuzzer passes: I have replaced "SPIRV_TOOLS_..." with "SOURCE_FUZZ_..."

### DIFF
--- a/source/fuzz/fuzzer_pass_add_composite_inserts.h
+++ b/source/fuzz/fuzzer_pass_add_composite_inserts.h
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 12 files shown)


stefanomil (changes requested):
In the existing fuzzer passes and transformations, there is also a _ at the end, like SOURCE_FUZZ_FUZZER_PASS_ADD_COMPOSITE_INSERTS_H_.

Could you change your ones to reflect this?

antnkarp:
> In the existing fuzzer passes and transformations, there is also a _ at the end, like SOURCE_FUZZ_FUZZER_PASS_ADD_COMPOSITE_INSERTS_H_.
> 
> Could you change your ones to reflect this?

Done, thanks.

stefanomil (approved):

afd (approved):
Thanks for fixing this!

PR #3790, opened 2020-09-10
spirv-fuzz: Create synonym of int constant using a loop

### LINKED ISSUE #3616: spirv-fuzz: Introduce a loop to create a synonym of an integer constant
**Motivation for transformation:*
- Increase the extent to which loops are used in the SPIR-V module
- Provide the possibility to replace constant uses with the results of computations coming from loops
- The new loops can be transformed further by other fuzzer passes to make them more interesting.

Suppose that C is an integer constant (possibly a vector).

Decide on:
- A number of loop iterations, N
- A step count vector S
- An initial value vector I such that C = I - N*S

Introduce a loop that does:

```
temp = I // A constant containing value I
for (int new_counter = 0; new_counter < N; new_counter++) {
  temp = temp - S;
}
```

The number of loop iterations should be fairly small, to avoid very long-running loops.
The step count can vary between components of the vector, but overall should be small enough that we don't have overflow problems.
We should be careful about the extent to which we nest these - that could be taken care of in the fuzzer pass, not the transformation, though.

Then add a synonym between C and the final value of temp.

stefanomil:
This transformation, given a constant integer (scalar or vector) C,
constants I and S of compatible type and scalar 32-bit integer constant
N, such that C = I - S*N, adds a loop which subtracts S from I, N
times, creating a synonym for C.

The related fuzzer pass randomly chooses constants to which to add
synonyms using this transformation, and the location where they should
be added.

Fixes #3616 .

### DIFF
--- a/source/fuzz/fuzzer_context.h
+++ b/source/fuzz/fuzzer_context.h
@@ -348,7 +348,7 @@ class FuzzerContext {
   uint32_t GetRandomIndexForCompositeInsert(uint32_t number_of_components) {
     return random_generator_->RandomUint32(number_of_components);
   }
-  int64_t GetRandomLongInteger() {
+  int64_t GetRandomValueForStepConstantInLoop() {
     // INT32_MAX is the maximum value that can be used with RandomUint32. Each
     // call to RandomUint32 will return a 31-bit number (the 32-th bit - the
     // sign bit - is always 0). By appending them together, we get a number
--- a/source/fuzz/fuzzer_context.h
+++ b/source/fuzz/fuzzer_context.h
@@ -154,6 +154,9 @@ class FuzzerContext {
   uint32_t GetChanceOfAddingLoopPreheader() {
     return chance_of_adding_loop_preheader_;
   }
+  uint32_t GetChanceOfAddingLoopToCreateIntSynonym() {
+    return chance_of_adding_loop_to_create_int_synonym_;
+  }
   uint32_t GetChanceOfAddingMatrixType() {
     return chance_of_adding_matrix_type_;
   }
@@ -204,6 +207,9 @@ class FuzzerContext {
     return chance_of_constructing_composite_;
   }
   uint32_t GetChanceOfCopyingObject() { return chance_of_copying_object_; }
+  uint32_t GetChanceOfCreatingIntSynonymsUsingLoops() {
+    return chance_of_creating_int_synonyms_using_loops_;
+  }
   uint32_t GetChanceOfDonatingAdditionalModule() {
     return chance_of_donating_additional_module_;
   }
@@ -218,6 +224,9 @@ class FuzzerContext {
   }
   uint32_t GetChanceOfGoingDeeperWhenMakingAccessChain() {
     return chance_of_going_deeper_when_making_access_chain_;
+  }
+  uint32_t GetChanceOfHavingTwoBlocksInLoopToCreateIntSynonym() {
+    return chance_of_having_two_blocks_in_loop_to_create_int_synonym_;
   }
   uint32_t GetChanceOfInliningFunction() {
     return chance_of_inlining_function_;
@@ -342,6 +351,15 @@ class FuzzerContext {
   uint32_t GetRandomIndexForCompositeInsert(uint32_t number_of_components) {
     return random_generator_->RandomUint32(number_of_components);
   }
+  uint64_t GetRandomLongInteger() {
+    uint64_t result = random_generator_->RandomUint32(INT32_MAX);
+    result <<= 31u;
+    result += random_generator_->RandomUint32(INT32_MAX);
+    if (random_generator_->RandomBool()) {
+      result = -result;
+    }
+    return result;
+  }
   uint32_t GetRandomLoopControlPartialCount() {
     return random_generator_->RandomUint32(max_loop_control_partial_count_);
   }
@@ -350,6 +368,9 @@ class FuzzerContext {
   }
   uint32_t GetRandomLoopLimit() {
     return random_generator_->RandomUint32(max_loop_limit_);
+  }
+  uint32_t GetRandomNumberOfLoopIterations(uint32_t max_num_iterations) {
+    return ChooseBetweenMinAndMax({1, max_num_iterations});
   }
   uint32_t GetRandomNumberOfNewParameters(uint32_t num_of_params) {
     assert(num_of_params < GetMaximumNumberOfFunctionParameters());
@@ -403,6 +424,7 @@ class FuzzerContext {
   uint32_t chance_of_adding_load_;
   uint32_t chance_of_adding_local_variable_;
   uint32_t chance_of_adding_loop_preheader_;
+  uint32_t chance_of_adding_loop_to_create_int_synonym_;
   uint32_t chance_of_adding_matrix_type_;
   uint32_t chance_of_adding_no_contraction_decoration_;
   uint32_t chance_of_adding_opphi_synonym_;
@@ -423,11 +445,13 @@ class FuzzerContext {
   uint32_t chance_of_choosing_workgroup_storage_class_;
   uint32_t chance_of_constructing_composite_;
   uint32_t chance_of_copying_object_;
+  uint32_t chance_of_creating_int_synonyms_using_loops_;
   uint32_t chance_of_donating_additional_module_;
   uint32_t chance_of_duplicating_region_with_selection_;
   uint32_t chance_of_flattening_conditional_branch_;
   uint32_t chance_of_going_deeper_to_insert_in_composite_;
   uint32_t chance_of_going_deeper_when_making_access_chain_;
+  uint32_t chance_of_having_two_blocks_in_loop_to_create_int_synonym_;
   uint32_t chance_of_inlining_function_;
   uint32_t chance_of_interchanging_signedness_of_integer_operands_;
   uint32_t chance_of_interchanging_zero_like_constants_;
--- a/source/fuzz/fuzzer_context.h
+++ b/source/fuzz/fuzzer_context.h
@@ -356,7 +356,7 @@ class FuzzerContext {
     uint64_t result = random_generator_->RandomUint32(INT32_MAX);
     result <<= 31u;
     result += random_generator_->RandomUint32(INT32_MAX);
-    // We choose the sign, so the number returned in the range (-2^62, 2^62)
+    // We choose the sign, so the number returned is in the range (-2^62, 2^62).
     return random_generator_->RandomBool() ? result : (0 - result);
   }
   uint32_t GetRandomLoopControlPartialCount() {
--- a/source/fuzz/fuzzer_context.h
+++ b/source/fuzz/fuzzer_context.h
@@ -349,15 +349,7 @@ class FuzzerContext {
     return random_generator_->RandomUint32(number_of_components);
   }
   int64_t GetRandomValueForStepConstantInLoop() {
-    // INT32_MAX is the maximum value that can be used with RandomUint32. Each
-    // call to RandomUint32 will return a 31-bit number (the 32-th bit - the
-    // sign bit - is always 0). By appending them together, we get a number
-    // between 0 and 2^62-1 (included).
-    uint64_t result = random_generator_->RandomUint32(INT32_MAX);
-    result <<= 31u;
-    result += random_generator_->RandomUint32(INT32_MAX);
-    // We choose the sign, so the number returned is in the range (-2^62, 2^62).
-    return random_generator_->RandomBool() ? result : (0 - result);
+    return random_generator_->RandomUint64(UINT64_MAX);
   }
   uint32_t GetRandomLoopControlPartialCount() {
     return random_generator_->RandomUint32(max_loop_control_partial_count_);
--- a/source/fuzz/fuzzer_context.h
+++ b/source/fuzz/fuzzer_context.h
@@ -348,14 +348,16 @@ class FuzzerContext {
   uint32_t GetRandomIndexForCompositeInsert(uint32_t number_of_components) {
     return random_generator_->RandomUint32(number_of_components);
   }
-  uint64_t GetRandomLongInteger() {
+  int64_t GetRandomLongInteger() {
+    // INT32_MAX is the maximum value that can be used with RandomUint32. Each
+    // call to RandomUint32 will return a 31-bit number (the 32-th bit - the
+    // sign bit - is always 0). By appending them together, we get a number
+    // between 0 and 2^62-1 (included).
     uint64_t result = random_generator_->RandomUint32(INT32_MAX);
     result <<= 31u;
     result += random_generator_->RandomUint32(INT32_MAX);
-    if (random_generator_->RandomBool()) {
-      result = -result;
-    }
-    return result;
+    // We choose the sign, so the number returned in the range (-2^62, 2^62)
+    return random_generator_->RandomBool() ? result : (0 - result);
   }
   uint32_t GetRandomLoopControlPartialCount() {
     return random_generator_->RandomUint32(max_loop_control_partial_count_);
--- a/source/fuzz/fuzzer_context.h
+++ b/source/fuzz/fuzzer_context.h
@@ -154,9 +154,6 @@ class FuzzerContext {
   uint32_t GetChanceOfAddingLoopPreheader() {
     return chance_of_adding_loop_preheader_;
   }
-  uint32_t GetChanceOfAddingLoopToCreateIntSynonym() {
-    return chance_of_adding_loop_to_create_int_synonym_;
-  }
   uint32_t GetChanceOfAddingMatrixType() {
     return chance_of_adding_matrix_type_;
   }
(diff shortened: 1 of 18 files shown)


source/fuzz/fuzzer_pass_add_loops_to_create_int_constant_synonyms.cpp
```diff
@@ -0,0 +1,234 @@
...
+      block_id = block->id();
+    }
+
+    // Randomly decide the values for the number of iterations and the step
+    // value, and compute the initial value accordingly.
+
+    // The maximum number of iterations depends on the loop nesting depth of
+    // the block. It is:
+    // - 1 if the nesting depth is <= 4
+    // - 2^(4 - nesting_depth) otherwise
```
antnkarp, line 114: If nesting_depth is >4 than the expression 2^(4-nesting_depth) is not integral. Did you mean (2^nesting_depth-4)?

source/fuzz/fuzzer_pass_add_loops_to_create_int_constant_synonyms.cpp
```diff
@@ -0,0 +1,234 @@
...
+  }
+}
+
+std::pair<uint32_t, uint32_t> FuzzerPassAddLoopsToCreateIntConstantSynonyms::
+    FindSuitableStepAndInitialValueConstants(uint64_t constant_val,
+                                             uint32_t bit_width, bool is_signed,
+                                             uint32_t num_iterations) {
+  // Choose the step value randomly and compute the initial value accordingly.
+  uint64_t step_value = GetFuzzerContext()->GetRandomLongInteger();
+  uint64_t initial_value = constant_val + step_value * num_iterations;
```
antnkarp, line 220: If |step_value| is a random long integer, this expression could potentially overflow. I think you need to check for this in your code. Also, try to add some test cases for this scenario.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -742,6 +743,88 @@ message TransformationAddLoopPreheader {
...
+  // A transformation that uses a loop to create a synonym for an integer
+  // constant C (scalar or vector) using an initial value I, a step value S and
+  // a number of iterations N such that C = I - N * S. For each iteration, S is
+  // added to the total.
+  // The loop can be made up of one or two blocks, and it is inserted before a
+  // loop with a single predecessor. In the one-block case, it is of the form:
+  //
+  //            %loop_id = OpLabel
+  //             %ctr_id = OpPhi %int %int_0 %pred %incremented_ctr_id %loop_id
+  //            %temp_id = OpPhi %intI %I %pred %eventual_syn_id %loop_id
```
antnkarp, line 756: Nit: use '%int' instead '%intI' to improve readability.

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
```diff
@@ -0,0 +1,418 @@
...
+  auto initial_val = ir_context->get_constant_mgr()->FindDeclaredConstant(
+      message_.initial_val_id());
+  auto step_val = ir_context->get_constant_mgr()->FindDeclaredConstant(
+      message_.step_val_id());
+
+  if (!constant || !initial_val || !step_val) {
+    return false;
+  }
+
+  // Check that |constant| is an integer scalar or vector constant.
```
antnkarp, line 66: I would write: "Check that |constant| is a constant that is an integer scalar or a vector of element type integer.

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
```diff
@@ -0,0 +1,418 @@
...
+  // |message_.num_iterations_id| is an integer constant with bit width 32.
+  auto num_iterations = ir_context->get_constant_mgr()->FindDeclaredConstant(
+      message_.num_iterations_id());
+
+  if (!num_iterations || !num_iterations->AsIntConstant() ||
+      num_iterations->type()->AsInteger()->width() != 32) {
+    return false;
+  }
+
+  // Check that the number of iterations is > 0 and <= 32.
```
antnkarp, line 109: I am not sure but is it possible that the nesting depth of a block is >9 and the transformation won't be applicable?

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -742,6 +743,88 @@ message TransformationAddLoopPreheader {
...
+  //    %eventual_syn_id = OpISub %intI %temp_id %step_val_id
+  // %incremented_ctr_id = OpIAdd %int %ctr_id %int_1
+  //            %cond_id = OpSLessThan %bool %incremented_ctr_id %num_iterations_id
+  //                       OpLoopMerge %block_after_loop_id %loop_id %none
+  //                       OpBranchConditional %cond_id %loop_id %block_after_loop_id
+  //
+  // Which can be translated, assuming that N > 0, to:
+  // for (int ctr = 0; ctr < N; ctr++) S = S - I
```
antnkarp, line 764: I think that you meant: 'I = I - S'

antnkarp (commented):
Very well designed and implemented transformation! I have added a few minor comments, one of which points out a possiblity of unhandled integer overflow.

source/fuzz/fuzzer_context.h
```diff
@@ -327,6 +333,15 @@ class FuzzerContext {
   uint32_t GetRandomIndexForCompositeInsert(uint32_t number_of_components) {
     return random_generator_->RandomUint32(number_of_components);
   }
+  uint64_t GetRandomLongInteger() {
+    uint64_t result = random_generator_->RandomUint32(INT32_MAX);
+    result <<= 31u;
+    result += random_generator_->RandomUint32(INT32_MAX);
+    if (random_generator_->RandomBool()) {
+      result = -result;
```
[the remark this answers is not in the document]
stefanomil: The reason is that I want it to also return negative integers. The fact that the return type is unsigned is indeed misleading, so I changed it and added some comments to explain this.

source/fuzz/fuzzer_pass_add_loops_to_create_int_constant_synonyms.cpp
```diff
@@ -0,0 +1,234 @@
...
+  FindOrCreateIntegerConstant({0}, 32, true, false);
+  FindOrCreateIntegerConstant({1}, 32, true, false);
+
+  // Consider each constant and each block.
+  for (uint32_t constant_id : constants) {
+    // Choose one of the blocks.
+    uint32_t block_id = blocks[GetFuzzerContext()->RandomIndex(blocks)];
+
+    // Adjust the block so that the transformation can be applied.
+
```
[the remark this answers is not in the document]
stefanomil: Done

source/fuzz/fuzzer_pass_add_loops_to_create_int_constant_synonyms.cpp
stefanomil: I meant "1 if the nesting depth is >= 4". Thanks for spotting this, I fixed the comment

source/fuzz/fuzzer_pass_add_loops_to_create_int_constant_synonyms.cpp
stefanomil: Overflows here should not cause any problems, since the transformation also allows them (the equation is satisfied as long as the last |bit_width| bits of the values of C and of (I - S*N) match).

Whether there are overflows or not C == I - S*N iff I == C + S*N (where == only considers the last |bit_width| bits).

I added some comments to explain this.

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
stefanomil: I changed it to "Check that the type of |constant| is integer scalar or vector with integer components", since I already checked that it is a constant.

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
stefanomil: The possible values that the fuzzer pass will give are {1, 2, 4, 8, 16}. I added that limit because it's in general not desirable to have too many iterations.

(If the nesting depth is >=4, the number of iterations will be 1. The comment in the fuzzer pass was wrong)

source/fuzz/protobufs/spvtoolsfuzz.proto
stefanomil: The comment is indeed wrong, thanks for spotting this. I changed it and used an additional variable syn, to make it clear that the value of I does not change:

```
  // int syn = I;
  // for (int ctr = 0; ctr < N; ctr++) syn = syn - S;
```

source/fuzz/protobufs/spvtoolsfuzz.proto
antnkarp: Great. In my opinion now it's much more readable.

source/fuzz/fuzzer_pass_add_loops_to_create_int_constant_synonyms.cpp
```diff
@@ -0,0 +1,236 @@
...
+    // The maximum number of iterations depends on the loop nesting depth of
+    // the block. It is:
+    // - 1 if the nesting depth is >= 4
+    // - 2^(4 - nesting_depth) otherwise
+    uint32_t nesting_depth =
+        GetIRContext()->GetStructuredCFGAnalysis()->LoopNestingDepth(
+            block->id());
+    uint32_t num_iterations =
+        nesting_depth >= 4
```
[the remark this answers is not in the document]
afd: Should we be considering inter-procedural loop nesting here?

afd (changes requested):
This is looking good - only a partial review but thought I'd submit this feedback and suggestions so far.

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -742,6 +743,89 @@ message TransformationAddLoopPreheader {
 
 }
 
+message TransformationAddLoopToCreateIntConstantSynonym {
+  // A transformation that uses a loop to create a synonym for an integer
+  // constant C (scalar or vector) using an initial value I, a step value S and
+  // a number of iterations N such that C = I - N * S. For each iteration, S is
+  // added to the total.
+  // The loop can be made up of one or two blocks, and it is inserted before a
+  // loop with a single predecessor. In the one-block case, it is of the form:
```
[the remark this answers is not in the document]
stefanomil: Yes

source/fuzz/protobufs/spvtoolsfuzz.proto
```diff
@@ -742,6 +743,89 @@ message TransformationAddLoopPreheader {
...
+  //
+  //            %loop_id = OpLabel
+  //             %ctr_id = OpPhi %int %int_0 %pred %incremented_ctr_id %loop_id
+  //            %temp_id = OpPhi %type_of_I %I %pred %eventual_syn_id %loop_id
+  //    %eventual_syn_id = OpISub %type_of_I %temp_id %step_val_id
+  // %incremented_ctr_id = OpIAdd %int %ctr_id %int_1
+  //            %cond_id = OpSLessThan %bool %incremented_ctr_id %num_iterations_id
+  //                       OpLoopMerge %block_after_loop_id %loop_id %none
+  //                       OpBranchConditional %cond_id %loop_id %block_after_loop_id
```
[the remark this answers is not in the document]
stefanomil: I have added a few lines to the comment to show it

source/fuzz/fuzzer_pass_add_loops_to_create_int_constant_synonyms.cpp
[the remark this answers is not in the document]
stefanomil: Yes, I had forgotten about that

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+void TransformationAddLoopToCreateIntConstantSynonym::Apply(
+    opt::IRContext* ir_context,
+    TransformationContext* transformation_context) const {
+  // Find 32-bit signed integer constants 0 and 1.
+  uint32_t const_0_id = fuzzerutil::MaybeGetIntegerConstant(
+      ir_context, *transformation_context, {0}, 32, true, false);
+  auto const_0_def = ir_context->get_def_use_mgr()->GetDef(const_0_id);
+  uint32_t const_1_id = fuzzerutil::MaybeGetIntegerConstant(
+      ir_context, *transformation_context, {0}, 32, true, false);
```
antnkarp, line 221: Shouldn't here be '{1}' instead of '{0}' ?

stefanomil: Yes, thanks for spotting this!

source/fuzz/fuzzer_pass_add_loops_to_create_int_constant_synonyms.cpp
```diff
@@ -0,0 +1,236 @@
...
+    // Randomly decide whether to consider this constant.
+    if (!GetFuzzerContext()->ChoosePercentage(
+            GetFuzzerContext()->GetChanceOfCreatingIntSynonymsUsingLoops())) {
+      continue;
+    }
+
+    auto constant = GetIRContext()->get_constant_mgr()->FindDeclaredConstant(
+        constant_def->result_id());
+
+    // We only consider integer constants (scalar or vector).
```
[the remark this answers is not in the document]
afd: Ah yes, that makes total sense, thanks.

source/fuzz/fuzzer_context.h
```diff
@@ -327,6 +333,17 @@ class FuzzerContext {
...
+  int64_t GetRandomValueForStepConstantInLoop() {
+    // INT32_MAX is the maximum value that can be used with RandomUint32. Each
+    // call to RandomUint32 will return a 31-bit number (the 32-th bit - the
+    // sign bit - is always 0). By appending them together, we get a number
+    // between 0 and 2^62-1 (included).
+    uint64_t result = random_generator_->RandomUint32(INT32_MAX);
+    result <<= 31u;
+    result += random_generator_->RandomUint32(INT32_MAX);
+    // We choose the sign, so the number returned is in the range (-2^62, 2^62).
+    return random_generator_->RandomBool() ? result : (0 - result);
```
afd, line 345: Am I right that what you'd really like is a function, RandomInt64, that could be invoked on random_generator_?  If so then I think it would be better to simply add that function rather than resort to pasting together two integers.

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
```diff
@@ -0,0 +1,419 @@
...
+    return false;
+  }
+
+  // Check that the number of iterations is > 0 and <= 32.
+  int32_t num_iterations_value =
+      num_iterations->AsIntConstant()->GetS32BitValue();
+
+  if (num_iterations_value <= 0 || num_iterations_value > 32) {
+    return false;
+  }
```
afd, line 116: Alternative, and slightly cleaner in my opinion: get the number of iterations as a U32 value, and then check whether it exceeds 32.

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
```diff
@@ -0,0 +1,419 @@
...
+    // Insert the header before the existing block.
+    block_after_loop->GetParent()->InsertBasicBlockBefore(std::move(loop_block),
+                                                          block_after_loop);
+  }
+
+  // Update the branching instructions leading to this block.
+  ir_context->get_def_use_mgr()->ForEachUse(
+      message_.block_after_loop_id(),
+      [this](opt::Instruction* instruction, uint32_t operand_index) {
+        // Replace all uses of the label inside branch instructions.
```
afd, line 367: I believe that by construction the instruction's opcode cannot be OpLoopMerge, OpSelectionMerge or OpSwitch, but it would be good to add assertions to document and check that.

afd (changes requested):
This is looking great!  A few relatively small changes requested.

source/fuzz/fuzzer_context.h
stefanomil: Indeed. It is now used after being introduced by #3809 .

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
stefanomil: Changed

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
stefanomil: I have added an assertion

source/fuzz/transformation_add_loop_to_create_int_constant_synonym.cpp
stefanomil: We actually still need to check that the number of iterations is not 0

afd (approved):
Looks excellent - thanks!

Issue #3723, opened 2020-08-18
spirv-fuzz: Determine if two pointers point to different memory regions


Vasniktel:
Currently, there is no way to tell if two pointers point to the different memory locations. That being said, DataSynonym fact is not enough.

We could implement some interesting transformations if we had a way to determine this. In particular, we could improve support for memory instructions in `TransformationMoveInstructionDown`.

Vasniktel:
I've thrown this here "just in case". Maybe it is better not to implement some general-purpose machinery and instead focus on a particular use case.

PR #3795, opened 2020-09-11
spirv-fuzz: Consider all ids from dead blocks irrelevant

### LINKED ISSUE #3733: spirv-fuzz: All instructions in a dead block are irrelevant
Consider marking all instructions in dead blocks as irrelevant.

stefanomil:
This PR modifies the FactManager methods IdIsIrrelevant and GetIrrelevantIds so
that an id is always considered irrelevant if it comes from a dead block.

Fixes #3733 .

### DIFF
--- a/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
+++ b/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
@@ -56,6 +56,7 @@ void DataSynonymAndIdEquationFacts::AddFact(
     const DeadBlockFacts& dead_block_facts,
     const IrrelevantValueFacts& irrelevant_value_facts,
     opt::IRContext* context) {
+  (void)dead_block_facts;        // Keep release compilers happy.
   (void)irrelevant_value_facts;  // Keep release compilers happy.
   assert(!irrelevant_value_facts.IdIsIrrelevant(fact.data1().object(),
                                                 dead_block_facts, context) &&
@@ -73,6 +74,7 @@ void DataSynonymAndIdEquationFacts::AddFact(
     const DeadBlockFacts& dead_block_facts,
     const IrrelevantValueFacts& irrelevant_value_facts,
     opt::IRContext* context) {
+  (void)dead_block_facts;        // Keep release compilers happy.
   (void)irrelevant_value_facts;  // Keep release compilers happy.
   assert(!irrelevant_value_facts.IdIsIrrelevant(fact.lhs_id(), dead_block_facts,
                                                 context) &&
--- a/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
+++ b/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
@@ -53,11 +53,14 @@ bool DataSynonymAndIdEquationFacts::OperationEquals::operator()(
 
 void DataSynonymAndIdEquationFacts::AddFact(
     const protobufs::FactDataSynonym& fact,
+    const DeadBlockFacts& dead_block_facts,
     const IrrelevantValueFacts& irrelevant_value_facts,
     opt::IRContext* context) {
   (void)irrelevant_value_facts;  // Keep release compilers happy.
-  assert(!irrelevant_value_facts.IdIsIrrelevant(fact.data1().object()) &&
-         !irrelevant_value_facts.IdIsIrrelevant(fact.data2().object()) &&
+  assert(!irrelevant_value_facts.IdIsIrrelevant(fact.data1().object(),
+                                                dead_block_facts, context) &&
+         !irrelevant_value_facts.IdIsIrrelevant(fact.data2().object(),
+                                                dead_block_facts, context) &&
          "Irrelevant ids cannot be synonymous with other ids.");
 
   // Add the fact, including all facts relating sub-components of the data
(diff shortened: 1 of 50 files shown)


Vasniktel (approved):

afd:
@stefanomil Looks like you have an unused parameter - likely one that's used only for assertion checking.

source/fuzz/fact_manager/dead_block_facts.h
```diff
@@ -33,6 +33,9 @@ class DeadBlockFacts {
   // See method in FactManager which delegates to this method.
   bool BlockIsDead(uint32_t block_id) const;
 
+  // Returns an unordered set of all the block ids that have been declared dead.
```
afd, line 36: Nit: I try to avoid comments that reiterate implementation details of the return type.  I'd say here "Returns the set of all block ids that have been declared dead."  The fact that it's an unordered set is apparent from the type.

source/fuzz/fact_manager/irrelevant_value_facts.cpp
```diff
@@ -60,13 +62,39 @@ bool IrrelevantValueFacts::PointeeValueIsIrrelevant(uint32_t pointer_id) const {
...
-  return irrelevant_ids_;
+std::unordered_set<uint32_t> IrrelevantValueFacts::GetIrrelevantIds(
+    const DeadBlockFacts& dead_block_facts, opt::IRContext* context) const {
+  // Get all the ids that have been declared irrelevant.
+  auto irrelevant_ids = irrelevant_ids_;
+
+  // Get all the ids declared in dead blocks.
+  for (uint32_t block_id : dead_block_facts.GetDeadBlocks()) {
+    auto block = fuzzerutil::MaybeFindBlock(context, block_id);
+    if (!block) {
```
afd, line 87: I think it would be good to remark here that it's possible for the block to no longer exist in the module, e.g. if it has been merged with another block.

afd (changes requested):
Thanks for making this big change!

A couple of nits in the comments.

Also: while it's great you modified all the existing tests to reflect the changes in the fact manager interface, I don't think you added any new tests to check whether ids from dead blocks are actually regarded as irrelevant.

Would you be able to do something for this?  There should already be tests that involve the BlockIsDead fact; you could adapt them to also assert that ids declared in those blocks are deemed irrelevant.

Once this is done I will approve and merge.

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
```diff
@@ -53,11 +53,15 @@ bool DataSynonymAndIdEquationFacts::OperationEquals::operator()(
 
 void DataSynonymAndIdEquationFacts::AddFact(
     const protobufs::FactDataSynonym& fact,
+    const DeadBlockFacts& dead_block_facts,
     const IrrelevantValueFacts& irrelevant_value_facts,
```
Vasniktel, line 57: This looks a bit convoluted to me: we are not using `dead_block_facts` anywhere in the method and only passing it down into `irrelevant_value_facts.IdIsIrrelevant` call. Maybe it's better to refactor all submanagers into class fields instead? @afd, @stefanomil, what do you think?

source/fuzz/fact_manager/dead_block_facts.h
stefanomil: Fixed

source/fuzz/fact_manager/irrelevant_value_facts.cpp
stefanomil: Done

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
stefanomil: Summary of the discussion: we will keep it as it is (at least for now).

stefanomil:
I have made the changes and added tests to check that ids from dead blocks are considered irrelevant.

I have also changed it so that if a pointer is in a dead block, the pointee value is considered irrelevant rather than the pointer id. This is to be consistent with the checks made when declaring new irrelevant ids or pointee values.

source/fuzz/fact_manager/irrelevant_value_facts.cpp
```diff
@@ -56,17 +58,79 @@ void IrrelevantValueFacts::AddFact(
   irrelevant_ids_.insert(fact.result_id());
 }
 
-bool IrrelevantValueFacts::PointeeValueIsIrrelevant(uint32_t pointer_id) const {
-  return pointers_to_irrelevant_pointees_ids_.count(pointer_id) != 0;
+bool IrrelevantValueFacts::PointeeValueIsIrrelevant(
```
Vasniktel, line 61: I'm not sure if we can consider pointees irrelevant even if the corresponding instructions are in dead blocks. Consider:
```
%1 = OpFunction %void None %type
%2 = OpLabel
%3 = OpVariable ...
...
     OpBranchConditional %true %4 %dead
%4 = OpLabel
...
%dead = OpLabel
%5 = OpCopyObject %pointer %3 ; can't be marked with PointeeValueIsIrrelevant
...
```

This is related to the fact that we can't determine whether two pointers point to the same memory location right now (#3723).

Vasniktel (changes requested):
Spotted a potential bug regarding irrelevant pointees.

source/fuzz/fact_manager/irrelevant_value_facts.cpp
stefanomil: You are quite right, I have now fixed it so that they are not considered irrelevant

afd (approved):
LGTM.

PR #3796, opened 2020-09-11
spirv-fuzz: Pass submanagers to other submanagers when necessary

### LINKED ISSUE #3550: spirv-fuzz: Unable to create certain types of queries in the fact manager
The fact manager consists of a set of sub-managers for each specific fact (e.g. there is a `DataSynonymAndIdEquationFacts` class  that handles `DataSynonym` facts, `IrrelevantValueFacts` class that handles `PointeeValueIsIrrelevant` and `IdIsIrrelevant` facts etc.)

 It is sometimes useful to make queries about some facts from sub-managers of different facts (e.g. `DataSynonym` might want to assert that neither of two synonymous data descriptors contains an irrelevant id by querying `IrrelevantValueFacts` class). 

However, we can't do that with the current design of the fact manager - neither of the sub-managers has a reference to either the parent fact manager or other sub-managers. A simple solution is to pass a pointer (or a reference) to the parent fact manager to all its components. This approach might not be the best one in terms of readability, though. The other approach is to flatten all methods from the fact manager's components into the manager itself.

stefanomil:
This PR changes the fact manager so that, when calling some of the
functions in submanagers, passes references to other submanagers if
necessary (e.g. to make consistency checks).

In particular:
- DataSynonymAndIdEquationFacts is passed to the AddFactIdIsIrrelevant
  function of IrrelevantValueFacts
- IrrelevantValueFacts is passed to the AddFact functions of
  DataSynonymAndIdEquationFacts
  
The IRContext is also passed when necessary and the calls to the
corresponding functions in FactManager were updated to be valid and
always use an updated context. 

Fixes #3550 .

### DIFF
--- a/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
+++ b/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
@@ -55,7 +55,7 @@ void DataSynonymAndIdEquationFacts::AddFact(
     const protobufs::FactDataSynonym& fact,
     const IrrelevantValueFacts& irrelevant_value_facts,
     opt::IRContext* context) {
-  (void)irrelevant_value_facts;
+  (void)irrelevant_value_facts;  // Keep release compilers happy.
   assert(!irrelevant_value_facts.IdIsIrrelevant(fact.data1().object()) &&
          !irrelevant_value_facts.IdIsIrrelevant(fact.data2().object()) &&
          "Irrelevant ids cannot be synonymous with other ids.");
@@ -69,7 +69,7 @@ void DataSynonymAndIdEquationFacts::AddFact(
     const protobufs::FactIdEquation& fact,
     const IrrelevantValueFacts& irrelevant_value_facts,
     opt::IRContext* context) {
-  (void)irrelevant_value_facts;
+  (void)irrelevant_value_facts;  // Keep release compilers happy.
   assert(!irrelevant_value_facts.IdIsIrrelevant(fact.lhs_id()) &&
          "Irrelevant ids are not allowed.");
 
--- a/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
+++ b/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "source/fuzz/fact_manager/data_synonym_and_id_equation_facts.h"
+
 #include "source/fuzz/fuzzer_util.h"
 
 namespace spvtools {
@@ -54,6 +55,7 @@ void DataSynonymAndIdEquationFacts::AddFact(
     const protobufs::FactDataSynonym& fact,
     const IrrelevantValueFacts& irrelevant_value_facts,
     opt::IRContext* context) {
+  (void)irrelevant_value_facts;
   assert(!irrelevant_value_facts.IdIsIrrelevant(fact.data1().object()) &&
          !irrelevant_value_facts.IdIsIrrelevant(fact.data2().object()) &&
          "Irrelevant ids cannot be synonymous with other ids.");
@@ -67,6 +69,7 @@ void DataSynonymAndIdEquationFacts::AddFact(
     const protobufs::FactIdEquation& fact,
     const IrrelevantValueFacts& irrelevant_value_facts,
     opt::IRContext* context) {
+  (void)irrelevant_value_facts;
   assert(!irrelevant_value_facts.IdIsIrrelevant(fact.lhs_id()) &&
          "Irrelevant ids are not allowed.");
 
--- a/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
+++ b/source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #include "source/fuzz/fact_manager/data_synonym_and_id_equation_facts.h"
-
 #include "source/fuzz/fuzzer_util.h"
 
 namespace spvtools {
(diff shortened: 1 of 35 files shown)


source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #include "source/fuzz/fact_manager/data_synonym_and_id_equation_facts.h"
-
```
afd, line 16: Don't remove this line; the #include for the header file associated with the cpp file comes first, then a space, then any standard headers (none in this case), then a space, then other headers.  So we do want the space.

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
```diff
@@ -52,19 +51,24 @@ bool DataSynonymAndIdEquationFacts::OperationEquals::operator()(
 }
 
 void DataSynonymAndIdEquationFacts::AddFact(
-    const protobufs::FactDataSynonym& fact, opt::IRContext* context) {
-  // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3550)
-  //  Assert that ids are not irrelevant.
+    const protobufs::FactDataSynonym& fact,
+    const IrrelevantValueFacts& irrelevant_value_facts,
+    opt::IRContext* context) {
+  assert(!irrelevant_value_facts.IdIsIrrelevant(fact.data1().object()) &&
```
afd, line 57: Before the assertion you'll need to do:

```
(void) irrelevant_value_facts; // Keep release compilers happy.
```

afd: And similar in other places in the code where you pass in a sub-manager only for assertion-checking purposes.

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
```diff
@@ -828,6 +832,11 @@ bool DataSynonymAndIdEquationFacts::DataDescriptorsAreWellFormedAndComparable(
           get_bit_count_for_numeric_type(*type_b));
 }
 
+std::vector<const protobufs::DataDescriptor*>
+DataSynonymAndIdEquationFacts::GetSynonymsForId(uint32_t id) const {
+  return GetSynonymsForDataDescriptor(MakeDataDescriptor(id, {}));
+}
+
```
afd, line 839: Thanks for this clean-up.

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.h
```diff
@@ -32,10 +33,18 @@ namespace fact_manager {
 class DataSynonymAndIdEquationFacts {
  public:
   // See method in FactManager which delegates to this method.
-  void AddFact(const protobufs::FactDataSynonym& fact, opt::IRContext* context);
```
afd, line 35: Can you update the comment to add the point that |irrelevant_value_facts| is passed in to allow consistency checking?

afd: ... and similarly for other methods that take a new argument.

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.h
```diff
@@ -20,6 +20,7 @@
 
 #include "source/fuzz/data_descriptor.h"
 #include "source/fuzz/equivalence_relation.h"
+#include "source/fuzz/fact_manager/irrelevant_value_facts.h"
```
afd, line 23: I wonder whether a forward reference might be more appropriate here (with the #include happening in the .cpp file).

The #include works because we only do things in one direction, but if in the future we needed consistency checks in the other direction then we would have to fall back to forward referencing.

afd: In fact, we do do it both ways round and you have forward-referenced (necessarily) the other way.  I think it would be better to keep it symmetric and forward reference in both cases.

source/fuzz/fact_manager/irrelevant_value_facts.cpp
```diff
@@ -15,25 +15,30 @@
...
-    const protobufs::FactPointeeValueIsIrrelevant& fact) {
-  // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3550)
-  //  Assert that the id does not participate in DataSynonym facts and is a
-  //  pointer.
+    const protobufs::FactPointeeValueIsIrrelevant& fact,
+    const DataSynonymAndIdEquationFacts& data_synonym_and_id_equation_facts) {
+  assert(data_synonym_and_id_equation_facts.GetSynonymsForId(fact.pointer_id())
+             .empty() &&
+         "The id cannot participate in DataSynonym facts.");
+  // TODO: Assert that the id is a pointer.
```
afd, line 30: We avoid raw TODOs in the project, so if we're keeping this then it would be good to keep the associated issue or open a new issue.  I guess we'd need to pass in IRContext to check whether it's a pointer; I'd be fine with that.

source/fuzz/fact_manager/irrelevant_value_facts.cpp
```diff
@@ -15,25 +15,30 @@
...
-  // TODO(https://github.com/KhronosGroup/SPIRV-Tools/issues/3550)
-  //  Assert that the id does not participate in DataSynonym facts and is not a
-  //  pointer.
+void IrrelevantValueFacts::AddFact(
+    const protobufs::FactIdIsIrrelevant& fact,
+    const DataSynonymAndIdEquationFacts& data_synonym_and_id_equation_facts) {
+  assert(data_synonym_and_id_equation_facts.GetSynonymsForId(fact.result_id())
+             .empty() &&
+         "The id cannot participate in DataSynonym facts.");
+  // TODO: Assert that the id is not a pointer.
```
afd, line 41: Same remark.

afd (changes requested):
Looks very good!  Some changes suggested.

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
stefanomil: I added it back

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.cpp
stefanomil: Done

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.h
stefanomil: Done

source/fuzz/fact_manager/data_synonym_and_id_equation_facts.h
stefanomil: I have now used a forward reference in both cases

source/fuzz/fact_manager/irrelevant_value_facts.cpp
stefanomil: The functions now check whether it's a pointer by using an IRContext (I have also updated all the calls to it and adjusted them so that they don't use an outdated IRContext)

afd (approved):
Thanks for this contribution!

Issue #3698, opened 2020-08-14
spirv-fuzz: Refactor the fact manager


Vasniktel:
This includes:
1. Split the implementation into multiple files
2. #3550
3. Assert that we are adding valid facts (e.g. `PointeeValueIsIrrelevant` is applied to the pointer).
4. Refactor existing transformations to add facts after analyses have been invalidated.
5. AddInitialFacts function.
6. Refactor tests into separate files

PR #3804, opened 2020-09-15
Fix build errors

afd:


paulthomson (approved):

Vasniktel:
There is also a bug in https://github.com/KhronosGroup/SPIRV-Tools/blob/4c239bd81bf79337e4a0cfe1c9da9caf1b25108a/test/fuzz/transformation_move_instruction_down_test.cpp#L610 if I recall correctly.

source/fuzz/transformation_move_instruction_down.cpp
```diff
@@ -14,9 +14,9 @@
 
 #include "source/fuzz/transformation_move_instruction_down.h"
 
-#include "external/spirv-headers/include/spirv/unified1/GLSL.std.450.h"
 #include "source/fuzz/fuzzer_util.h"
 #include "source/fuzz/instruction_descriptor.h"
+#include "spirv/unified1/GLSL.std.450.h"
```
Vasniktel, line 19: Thanks for fixing!

afd: No problem!

Vasniktel:
It seems tests are failing because of the assertion in the fact manager. https://github.com/KhronosGroup/SPIRV-Tools/blob/51b826140a63b74c93cfe8074ef643dec026adb4/source/fuzz/transformation_flatten_conditional_branch.cpp#L581-L582 should be moved after the call to `InvalidateAnalyses`.

Alternatively, you could comment this out as well as disable the corresponding tests and I will resolve this issue when I refactor the fact manager as in #3698.

afd:
Thanks - I am working on a fix.  Seems like these PRs landed in an unfortunate order :)

PR #3810, opened 2020-09-16
spirv-fuzz: Use an irrelevant id for the unused components

### LINKED ISSUE #3808: spirv-fuzz: Should TransformationAddImageSampleUnusedComponents use an irrelevant id for the unused components?
TransformationAddImageSampleUnusedComponents adds unused components to an image sample instruction.  These components can presumably have arbitrary values, so we should use a zero with the id is irrelevant fact.  I'm not sure we do that at present.

@andreperezmaselco for info.

andreperezmaselco:
Fixes #3808.


afd (approved):
Thanks!

PR #3815, opened 2020-09-17
spirv-fuzz: Fix TransformationDuplicateRegionWithSelection

antnkarp:
Introduces two changes:
1. duplicated_exit_region refers to a correct block, regardless of the order of the blocks in the enclosing function.
2. Exclude the case where the continue target is the exit block.

### DIFF
--- a/source/fuzz/transformation_duplicate_region_with_selection.cpp
+++ b/source/fuzz/transformation_duplicate_region_with_selection.cpp
@@ -167,9 +167,11 @@ bool TransformationDuplicateRegionWithSelection::IsApplicable(
           ir_context->cfg()->block(loop_merge->GetSingleWordOperand(1));
       // The continue target is a single-entry, single-exit region. Therefore,
       // if the continue target is the exit block, the region might not contain
-      // the loop header.
-      if (continue_target != exit_block &&
-          region_set.count(&block) != region_set.count(continue_target)) {
+      // the loop header. However, we would like to exclude this situation,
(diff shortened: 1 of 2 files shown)


source/fuzz/transformation_duplicate_region_with_selection.cpp
```diff
@@ -499,8 +503,9 @@ void TransformationDuplicateRegionWithSelection::Apply(
       opt::Instruction::OperandList(
           {{SPV_OPERAND_TYPE_ID, {message_.merge_label_fresh_id()}}}));
   exit_block->AddInstruction(MakeUnique<opt::Instruction>(merge_branch_instr));
+  auto merge_branch_instr_cloned = merge_branch_instr.Clone(ir_context);
   duplicated_exit_block->AddInstruction(
-      MakeUnique<opt::Instruction>(merge_branch_instr));
+      std::unique_ptr<opt::Instruction>(merge_branch_instr_cloned));
```
afd, line 508: Could you avoid the "merge_branch_instr_cloned" temporary and just pass:

```
std::unique_ptr<opt::Instruction>(merge_branch_instr.Clone(ir_context)
```

as the parameter?

afd (changes requested):
Whenever we fix a bug, we try to add a unit test that would have exposed the bug before the fix was applied.  Could you do that in this PR please?

afd (approved):
LGTM.

PR #3818, opened 2020-09-18
spirv-fuzz: Refactor fuzzer, replayer and shrinker

afd:
In preparation for some upcoming work on the shrinker, this PR changes
the interfaces of Fuzzer, Replayer and Shrinker so that all data
relevant to each class is provided on construction, meaning that the
"Run" method can become a zero-argument method that returns a status,
transformed binary and sequence of applied transformations via a
struct.

This makes greater use of fields, so that -- especially in Fuzzer --
there is a lot less parameter passing.

### DIFF
--- a/source/fuzz/fuzzer.h
+++ b/source/fuzz/fuzzer.h
@@ -24,6 +24,8 @@
 #include "source/fuzz/pass_management/repeated_pass_instances.h"
 #include "source/fuzz/pass_management/repeated_pass_recommender.h"
 #include "source/fuzz/protobufs/spirvfuzz_protobufs.h"
+#include "source/fuzz/random_generator.h"
+#include "source/opt/ir_context.h"
 #include "spirv-tools/libspirv.hpp"
 
 namespace spvtools {
@@ -41,6 +43,12 @@ class Fuzzer {
     kInitialBinaryInvalid,
   };
 
+  struct FuzzerResult {
+    FuzzerResultStatus status;
+    std::vector<uint32_t> transformed_binary;
+    protobufs::TransformationSequence applied_transformations;
+  };
+
   // Each field of this enum corresponds to an available repeated pass
   // strategy, and is used to decide which kind of RepeatedPassManager object
   // to create.
@@ -50,6 +58,7 @@ class Fuzzer {
     kLoopedWithRecommendations
   };
 
+  // TODO revise comment
   // Constructs a fuzzer from the given target environment |target_env|.  |seed|
   // is a seed for pseudo-random number generation.  If |enable_all_passes| is
   // true then all fuzzer passes will be enabled, otherwise a random subset of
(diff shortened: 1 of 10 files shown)


afd:
@Vasniktel If you have time to review this then I'd be keen to get feedback on the way I've used std::move.  I'm not yet super-confident on when it's best to use std::move.

source/fuzz/shrinker.cpp
```diff
@@ -237,22 +242,30 @@ Shrinker::ShrinkerResultStatus Shrinker::Run(
...
   // step limit.
+  //
+  // Either way, the output from the shrinker is the best binary we saw, and the
+  // transformations that led to it.
   assert(attempt <= step_limit_);
   if (attempt == step_limit_) {
     std::stringstream strstream;
     strstream << "Shrinking did not complete; step limit " << step_limit_
               << " was reached.";
     consumer_(SPV_MSG_WARNING, nullptr, {}, strstream.str().c_str());
```
Vasniktel, line 255: According to https://en.cppreference.com/w/cpp/io/basic_stringstream/str, `str()` returns a temporary object. Thus, `str().c_str()` results in a dangling pointer.

I've seen this bug somewhere else in the project as well. Maybe it would be better to make a separate PR for fixing it.

source/fuzz/shrinker.cpp
```diff
@@ -237,22 +242,30 @@ Shrinker::ShrinkerResultStatus Shrinker::Run(
...
-    return Shrinker::ShrinkerResultStatus::kStepLimitReached;
+    return {Shrinker::ShrinkerResultStatus::kStepLimitReached,
+            current_best_binary, current_best_transformations};
   }
-  return Shrinker::ShrinkerResultStatus::kComplete;
+  return {Shrinker::ShrinkerResultStatus::kComplete, current_best_binary,
+          current_best_transformations};
+}
```
Vasniktel, line 261: You could've probably used `std::move` here as well.

source/fuzz/fuzzer.h
```diff
@@ -70,73 +75,64 @@ class Fuzzer {
...
-      const std::vector<fuzzerutil::ModuleSupplier>& donor_suppliers,
-      std::vector<uint32_t>* binary_out,
-      protobufs::TransformationSequence* transformation_sequence_out);
+  // Transforms |binary_in_| by running a number of randomized fuzzer passes.
+  // Initial facts about the input binary and the context in which it will
+  // execute are provided via |initial_facts_|.  A source of donor modules to be
+  // used by transformations is provided via |donor_suppliers_|.  On success,
+  // returns the transformed binary and the sequence of transformations that
+  // were applied.
+  FuzzerResult Run();
```
Vasniktel, line 85: Maybe it's better to explicitly mention here that `FuzzerResult::trasformed_binary` and `FuzzerResult::applied_transformations` are empty if `FuzzerResult::status` is not `kComplete`. @afd, what do you think?

If you agree, don't forget to adjust your comments in shrinker and replayer as well.

Vasniktel (commented):
Looks good. Your uses of `std::move` seem OK to me.

source/fuzz/shrinker.cpp
afd: I believe this is OK because the lifetime of a temporary object is dictated by the "full-expression" in which it is nested.

Here, the full-expression is:

```
consumer_(SPV_MSG_WARNING, nullptr, {}, strstream.str().c_str())
```

I believe there would only be UB if we did something like this instead:

```
auto temp = strstr.str().c_str();
consumer_(SPV_MSG_WARNING, nullptr, {}, temp)
```

source/fuzz/shrinker.cpp
afd: Done.

source/fuzz/fuzzer.h
afd: Done.

source/fuzz/shrinker.cpp
Vasniktel: Yes, you are right! I didn't know that. In this case, I'm OK with the current approach.

Vasniktel (approved):

source/fuzz/shrinker.cpp
afd: Thanks.

PR #3834, opened 2020-09-22
Validate SPIRV Version number when parsing binary header

### LINKED ISSUE #3831: spirv-val: version number in header not checked that it is a valid version number
An unknown/invalid value like 0x00010021 will be accepted. 
If further a tool further down the chain, i.e. spirv-cross, is more strict about the version numbers this can cause problems.

In spvBinaryHeaderGet there is a TODO about validation.

zoddicus:
Fixes #3831

### DIFF
--- a/source/binary.cpp
+++ b/source/binary.cpp
@@ -48,6 +48,7 @@ spv_result_t spvBinaryHeaderGet(const spv_const_binary binary,
   // Per 2.3.1 version's high and low bytes are 0
   if ((pHeader->version & 0x000000ff) || pHeader->version & 0xff000000)
     return SPV_ERROR_INVALID_BINARY;
+  // Minimum version was 1.0 and max version is defined by SPV_VERSION.
   if (pHeader->version < SPV_SPIRV_VERSION_WORD(1, 0) ||
       pHeader->version > SPV_VERSION)
     return SPV_ERROR_INVALID_BINARY;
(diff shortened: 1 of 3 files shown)


source/binary.cpp
```diff
@@ -45,6 +45,21 @@ spv_result_t spvBinaryHeaderGet(const spv_const_binary binary,
   // TODO: Validation checking?
   pHeader->magic = spvFixWord(binary->code[SPV_INDEX_MAGIC_NUMBER], endian);
   pHeader->version = spvFixWord(binary->code[SPV_INDEX_VERSION_NUMBER], endian);
+  // Assumes versions are composed of a major and minor number, where major = 1
+  // and minor {0, current spec minor}. If version 2.x or 1.1x is released this
+  // check will need to be updated.
+
+  bool is_valid_version = false;
+  for (uint32_t minor = 0;
```
alan-baker, line 53: Why not just?
```
// Minimum version was 1.0 and max version is defined by SPV_VERSION.
if (pHeader->version < SPV_SPIRV_VERSION_WORD(1, 0) || pHeader->version > SPV_VERSION)
  return SPV_ERROR_INVALID_BINARY;
```

This should handle all future versions too.

alan-baker (changes requested):
You seem to be missing tests that catch invalid versions

source/binary.cpp
zoddicus: That is a much more elegant solution.

zoddicus:
> You seem to be missing tests that catch invalid versions

Added tests now

alan-baker (approved):

PR #3835, opened 2020-09-22
Debug info preservation in convert-local-access-chains pass

### LINKED ISSUE #3820: [OpenCL.DebugInfo.100] support debug info preservation in convert-local-access-chains pass

jaebaek:
Fixes #3820

### DIFF
--- a/source/opt/local_access_chain_convert_pass.cpp
+++ b/source/opt/local_access_chain_convert_pass.cpp
@@ -241,7 +241,7 @@ void LocalAccessChainConvertPass::FindTargetVars(Function* func) {
   }
 }
 
-bool LocalAccessChainConvertPass::HasNoOpenCL100DebugRef(
+bool LocalAccessChainConvertPass::HasOpenCL100DebugRef(
     Instruction* inst) const {
   return !get_def_use_mgr()->WhileEachUser(inst, [](Instruction* user) {
     if (user->IsOpenCL100DebugInstr()) {
@@ -262,11 +262,11 @@ Pass::Status LocalAccessChainConvertPass::ConvertLocalAccessChains(
     for (auto ii = bi->begin(); ii != bi->end(); ++ii) {
       switch (ii->opcode()) {
         case SpvOpLoad: {
-          if (HasNoOpenCL100DebugRef(&*ii)) break;
+          if (HasOpenCL100DebugRef(&*ii)) break;
           uint32_t varId;
           Instruction* ptrInst = GetPtr(&*ii, &varId);
           if (!IsNonPtrAccessChain(ptrInst->opcode())) break;
-          if (HasNoOpenCL100DebugRef(ptrInst)) break;
+          if (HasOpenCL100DebugRef(ptrInst)) break;
           if (!IsTargetVar(varId)) break;
           std::vector<std::unique_ptr<Instruction>> newInsts;
           if (!ReplaceAccessChainLoad(ptrInst, &*ii)) {
@@ -279,7 +279,7 @@ Pass::Status LocalAccessChainConvertPass::ConvertLocalAccessChains(
           Instruction* store = &*ii;
           Instruction* ptrInst = GetPtr(store, &varId);
           if (!IsNonPtrAccessChain(ptrInst->opcode())) break;
-          if (HasNoOpenCL100DebugRef(ptrInst)) break;
+          if (HasOpenCL100DebugRef(ptrInst)) break;
           if (!IsTargetVar(varId)) break;
           std::vector<std::unique_ptr<Instruction>> newInsts;
           uint32_t valId = store->GetSingleWordInOperand(kStoreValIdInIdx);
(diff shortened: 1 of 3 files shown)


source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -236,6 +241,16 @@ void LocalAccessChainConvertPass::FindTargetVars(Function* func) {
   }
 }
 
+bool LocalAccessChainConvertPass::HasNoOpenCL100DebugRef(
```
s-perron, line 244: Naming a function with a negative in it is more awkward to read because you can end up with double negaives:

`if (!HasNoDebugRef)` reads "if it does not have no debug ref."

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -236,6 +241,16 @@ void LocalAccessChainConvertPass::FindTargetVars(Function* func) {
...
+    Instruction* inst) const {
+  return !get_def_use_mgr()->WhileEachUser(inst, [](Instruction* user) {
```
s-perron, line 246: Is this returning what I expect?  You return true if it has a debug reference, but I expect it to return false because the name is "Has no opencl 100 debug ref."

s-perron (changes requested):
I need some clarification on what the function `LocalAccessChainConvertPass::HasNoOpenCL100DebugRef` is suppose to do.  It seems to be implemented and used opposite how it is documented.

source/opt/local_access_chain_convert_pass.cpp
jaebaek: Ah .. the name is wrong. I changed it to `HasOpenCL100DebugRef`.

source/opt/local_access_chain_convert_pass.cpp
jaebaek: Likewise.

jaebaek (commented):
Thank you for your code review.
The name is wrong. I changed it to `HasOpenCL100DebugRef`.
PTAL!

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -247,9 +262,11 @@ Pass::Status LocalAccessChainConvertPass::ConvertLocalAccessChains(
     for (auto ii = bi->begin(); ii != bi->end(); ++ii) {
       switch (ii->opcode()) {
         case SpvOpLoad: {
+          if (HasOpenCL100DebugRef(&*ii)) break;
```
s-perron, line 265: I'm wondering if this restriction might be a bit too much.  If we do not change the access chain, then many other optimizations could be disabled later one, and they will be needed for legalization.  We should talk about which cases you are worried about, and how they can be handled.

For example, if the result of the load is used in a DebugValue, then you change the DebugValue to use the result of toe extract.  I believe that will already happen.

source/opt/local_access_chain_convert_pass.cpp
```diff
@@ -247,9 +262,11 @@ Pass::Status LocalAccessChainConvertPass::ConvertLocalAccessChains(
...
           uint32_t varId;
           Instruction* ptrInst = GetPtr(&*ii, &varId);
           if (!IsNonPtrAccessChain(ptrInst->opcode())) break;
+          if (HasOpenCL100DebugRef(ptrInst)) break;
```
s-perron, line 269: Why should we not change the access chain?  I believe this pass leaves the original pointer around to be cleaned up by dead code elimination (I could be wrong about that).  If it does, then why does replacing the access chain make the debug information incorrect?

Same for the store below.

test/opt/local_access_chain_convert_test.cpp
```diff
@@ -96,6 +96,208 @@ OpFunctionEnd
                                                      true);
 }
 
+TEST_F(LocalAccessChainConvertTest, DebugInfo) {
```
s-perron, line 99: If would be nice if these tests told me what you where actually looking for.  A name like "DebugValueCorrectlyUpdated" or "DontOptimizeDebugDeclare" with a few comment are good.

s-perron (changes requested):
I think this is too pessimistic.  Since I can't tell what the tests are actually looking for, I cannot see an example of why it has to be that pessimistic.

source/opt/local_access_chain_convert_pass.cpp
jaebaek: I agree that this is too much. I just realized that the result of `ReplaceAccessChainLoad()` does not change the value -> the original one is a load of a component of a composite vs. the new one is an extraction of a composite.

I simply remove this if-statement.

source/opt/local_access_chain_convert_pass.cpp
jaebaek: I agree. I thought it will be killed by this pass, but it will not.
I simply remove this if-statement.

test/opt/local_access_chain_convert_test.cpp
jaebaek: updated

jaebaek:
Thank you for your review. I updated it. PTAL!

s-perron (approved):

PR #3836, opened 2020-09-22
Fix compiler error on macOS with XCode12

richard-lunarg:
Fixed a minor language building snafu with Apple's XCode12 and updated compiler tools. XCode12 is needed for the new MoltenVK (which brings in this library) libraries.


richard-lunarg:
@s-perron PTAL
Thanks.

s-perron (approved):

